PR #9, opened 2020-11-29
Resize via dragging corners

mechpen:


johanmalm:
Thank you. Very tidy. Both were on my radar - really appreciate the help.

PR #10, opened 2020-12-01
Fix zombie children

mechpen:


johanmalm:
Thanks 😄

johanmalm:
Not sure why the handling of SIGCHLD caused issues, but this seems to work fine.

PR #19, opened 2021-03-04
Allow downstream to disable manpages even if pandoc is found

jbeich:
See swaywm/sway#3452. To be used by Gentoo and [FreeBSD](https://github.com/freebsd/freebsd-ports/commit/d1251e059251dd28901b71374eb78b5c68514cc3).

```
$ pkg install hs-pandoc
$ meson setup -Dman-pages=disabled _build
[...]
Program pandoc skipped: feature man-pages disabled
[...]
```


johanmalm:
Thanks

PR #35, opened 2021-06-03
Unbreak build with wlroots 0.14.0 (2021-06-03 snapshot)

jbeich:
6744e103014b bumped wlroots to 0.14.0 (unreleased) which recently got a breaking change.


johanmalm:
Thanks

PR #37, opened 2021-06-26
meson: Use wrap files to manage dependencies

ezequielgarcia:
This simplifies building as there's no need to manually
git clone subprojects anymore.


johanmalm:
Thanks.

PR #54, opened 2021-08-22
implement input_inhibit protocol, needed for swaylock

telent:
this is in "it appears to work" state, though I blindly copy-pasted a little more code[*] than I'm happy with, so might benefit from a review

[*] from here: https://github.com/swaywm/wlroots/pull/807

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -41,14 +41,12 @@ handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym)
 	return false;
 }
 
-static void
-keyboard_key_notify(struct wl_listener *listener, void *data)
+static bool
+handle_compositor_keybindings(struct wl_listener *listener,
+			      struct wlr_event_keyboard_key *event)
 {
-	/* This event is raised when a key is pressed or released. */
 	struct seat *seat = wl_container_of(listener, seat, keyboard_key);
 	struct server *server = seat->server;
-	struct wlr_event_keyboard_key *event = data;
-	struct wlr_seat *wlr_seat = server->seat.seat;
 	struct wlr_input_device *device = seat->keyboard_group->input_device;
 
 	/* Translate libinput keycode -> xkbcommon */
@@ -68,13 +66,14 @@ keyboard_key_notify(struct wl_listener *listener, void *data)
 			/* end cycle */
 			desktop_focus_view(&server->seat, server->cycle_view);
 			server->cycle_view = NULL;
+			/* XXX should we handled=true here? */
 		} else if (event->state == WL_KEYBOARD_KEY_STATE_PRESSED) {
 			/* cycle to next */
 			server->cycle_view =
 				desktop_cycle_view(server, server->cycle_view);
 			osd_update(server);
 			damage_all_outputs(server);
-			return;
+			return true;
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


src/keyboard.c
```diff
@@ -68,13 +66,14 @@ keyboard_key_notify(struct wl_listener *listener, void *data)
 			/* end cycle */
 			desktop_focus_view(&server->seat, server->cycle_view);
 			server->cycle_view = NULL;
+			/* XXX should we handled=true here? */
```
johanmalm, line 69: Yes, seems like it. I'll make a note of that and play with that later.

src/server.c
```diff
@@ -78,6 +79,81 @@ drop_permissions(void)
 	}
 }
 
+static void seat_clear_touch_points(struct seat *seat,
```
johanmalm, line 82: We don't support touch (yet), so don't need this function.

src/server.c
```diff
@@ -78,6 +79,81 @@ drop_permissions(void)
...
+static void seat_inhibit_input(struct seat *seat,  struct wl_client *active_client)
+{
+	seat->active_client_while_inhibited = active_client;
+
+	if(seat->focused_layer &&
+	   (wl_resource_get_client(seat->focused_layer->resource) !=
+	    active_client))
+	{
+		wlr_log(WLR_INFO, "defocus layer");
+		seat_set_focus_layer(seat, NULL); /* ? */
```
johanmalm, line 105: I note your question mark. I think is fine.
If the active_client is in a layer, and that layer is the focused one, then leave it - otherwise set focused layer to NULL

src/server.c
```diff
@@ -78,6 +79,81 @@ drop_permissions(void)
...
+		seat_focus_surface(seat, NULL);	  /* keyboard focus */
+	}
+
+	struct wlr_seat_client * previous_ptr_client = seat->seat->pointer_state.focused_client;
+	if (previous_ptr_client &&
+	    (previous_ptr_client->client != active_client)) {
+		wlr_log(WLR_INFO, "defocus ptr");
+		wlr_seat_pointer_clear_focus(seat->seat);
+	}
+	seat_clear_touch_points(seat, active_client);
```
johanmalm, line 120: Again, remove this line.

johanmalm (changes requested):
Looks good.
Are you able to remove the 'touch' function and I'll merge.

src/server.c
telent: Hah, good spot. I put the comments in to keep straight in my head the difference between pointer focus and keyboard focus, will delete

src/server.c
telent: :+1:

telent:
One thing that occurs to me after a couple more days: it might be worth marking some of the compositor bindings as "still works even when input inhibited". For example, how it is right now, VT switching is not possible while the screen is locked

(On the other hand, maybe some people would regard that as a security feature...)

johanmalm:
With thanks :smile:

johanmalm:
> One thing that occurs to me after a couple more days: it might be worth marking some of the compositor bindings as "still works even when input inhibited". For example, how it is right now, VT switching is not possible while the screen is locked
> 
> (On the other hand, maybe some people would regard that as a security feature...)

Interesting point. I guess some display managers let other users login in even when the screen is locked. Don't think I'd have much need for it myself, but happy to consider.

PR #55, opened 2021-08-28
fix cycle_view for bindings other than Alt_L

telent:
Change the code that detects when to stop cycling the view,
so that instead of looking for the left Alt key release it
looks for all modifiers released (thus. will work whether you
have this action bound to Alt-Tab, C-Tab, Meta-Tab, etc)

To do this we move the test from keyboard_key.notify to
the later signal keyboard_modifiers.notify, because the xkb_state
has been updated by then and the modifier state will be
accurate


johanmalm:
Thanks

PR #57, opened 2021-08-29
Mouse rc

apbryan:
First implementation of <mouse> in rc.xml 

Can successfully parse the following XML and and implement the action:

```
<mouse>
    <context name="TitleBar">
        <mousebind button="Left" action="DoubleClick">
            <action name="ToggleMaximize"/>
        </mousebind>
    </context>
</mouse>
```
REQUEST FOR COMMENT:
The XML parsing code for this looks A LOT different than the already
existing XML parsing code. It may have to be reworked.  

Done: update docs/rc.xml and docs/rc.xml.all

### DIFF
--- a/include/config/mousebind.h
+++ b/include/config/mousebind.h
@@ -1,0 +1,41 @@
+#ifndef __LABWC_MOUSEBIND_H
+#define __LABWC_MOUSEBIND_H
+
+#include <wayland-util.h>
+
+enum mouse_context {
+	MOUSE_CONTEXT_TITLEBAR,
+	MOUSE_CONTEXT_NONE
+};
+
+enum mouse_button {
+	/*
+	 * These values match the values returned by button event->button and were
+	 * obtained experimentally
+	 */
+	MOUSE_BUTTON_LEFT = 272,
+	MOUSE_BUTTON_RIGHT = 273,
+	MOUSE_BUTTON_MIDDLE = 274,
+	MOUSE_BUTTON_NONE = -1
+};
+
+enum action_mouse_did {
+	MOUSE_ACTION_DOUBLECLICK,
+	MOUSE_ACTION_NONE
+};
+
+struct mousebind {
+	enum mouse_context context; /* ex: titlebar */
+	enum mouse_button button; /* ex: left, right, middle */
+	enum action_mouse_did mouse_action; /* ex: doubleclick, press, drag, etc */
+	const char* action; /* what to do because mouse did previous action */
+	const char* command;
+	struct wl_list link;
+};
+
+struct mousebind*
+mousebind_create(const char* context_str, const char* mouse_button_str,
+		const char* action_mouse_did_str, const char* action,
+		const char* command);
+
+#endif /* __LABWC_MOUSEBIND_H */
--- a/include/config/mousebind.h
+++ b/include/config/mousebind.h
@@ -1,23 +1,8 @@
 #ifndef __LABWC_MOUSEBIND_H
 #define __LABWC_MOUSEBIND_H
 
+#include "ssd.h"
 #include <wayland-util.h>
-
-enum mouse_context {
-	MOUSE_CONTEXT_TITLEBAR,
-	MOUSE_CONTEXT_NONE
-};
-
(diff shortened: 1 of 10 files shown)


johanmalm:
Hi. Really appreciate the contribution.
There are a few things we should consider changing.
If you’re happy to implement, I can give pointers.
Regarding xml parsing, I’d start with simplifying the double click speed.
In entry(), you can just insert a doubleclickspeed.mouse and do the strtol() there.
https://github.com/johanmalm/labwc/blob/75564d6b8d640c8653e5db82b5a0cba6375194c6/src/config/rcxml.c#L194

johanmalm:
Regarding xml parsing, the parsing of <font place=“”> is a good example for the mouse binds. Keeps the code a bit more consistent and simple.

include/config/mousebind.h
```diff
@@ -0,0 +1,41 @@
...
+#define __LABWC_MOUSEBIND_H
+
+#include <wayland-util.h>
+
+enum mouse_context {
+	MOUSE_CONTEXT_TITLEBAR,
+	MOUSE_CONTEXT_NONE
+};
+
+enum mouse_button {
```
johanmalm, line 11: These are defined in ‘#include <linux/input-event-codes.h>’
E.g. BTN_RIGHT, etc
So, would a uint32_t be enough?
Doing it this way we’d be consistent with wlroots’s ‘ struct wlr_event_pointer_button’ event->button

src/cursor.c
```diff
@@ -342,8 +343,15 @@ cursor_button(struct wl_listener *listener, void *data)
 	desktop_focus_view(&server->seat, view);
 	damage_all_outputs(server);
 
-	if (is_double_click(500) && view_area == LAB_SSD_PART_TITLEBAR) {
-		view_toggle_maximize(view);
+	if (is_double_click(rc.doubleclick_time) && view_area == LAB_SSD_PART_TITLEBAR) {
```
johanmalm, line 346: We can do this in a better, but let cross that bridge after we’ve got this merged

include/config/mousebind.h
```diff
@@ -0,0 +1,41 @@
+#ifndef __LABWC_MOUSEBIND_H
+#define __LABWC_MOUSEBIND_H
+
+#include <wayland-util.h>
+
+enum mouse_context {
```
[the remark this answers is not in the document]
apbryan: done

include/config/mousebind.h
apbryan: done

src/cursor.c
apbryan: Agreed. I just wanted to get one use-case going (doubleclick the title bar) to test the xml parsing. This will need to change a lot to support more mousebinds cleanly

src/config/rcxml.c
```diff
@@ -235,6 +354,13 @@ xml_tree_walk(xmlNode *node)
 			traverse(n);
 			in_keybind = false;
 			continue;
+		} 
+		if(!strcasecmp((char *)n->name, "mousebind")) {
+			in_mousebind = true;
+			traverse(n);
+			in_mousebind = false;
+			add_new_mousebinds();
```
johanmalm, line 362: Great progress. I think we’re very close.
We should not add_new_mousebind here.
I’d need to play with it myself to check, but think that mousebind_create() should be called from fill_mousebind() the first time we get to a new ‘<mousebind>’ element - and after truncation I’m guessing that nodename would then be empty (“”).
If it runs fine, I’m happy to merge now and we can tweak/re-factor a bit when I’m back behind a machine.
I’m not that pedantic on coding style, but if you’re happy to fix the following, that’d be great
- Space between if and opening bracket
- When declaring pointers, but the star with the binding (so do ‘struct a *b’ rather than ‘struct a* b’.
Do you want me to quash the commits when merging?

johanmalm: I try to keep consistent with this https://git.sr.ht/~sircmpwn/cstyle which is pretty close to wlroots/sway.

Have no clang-format.

Just say if you prefer me to do anything. I’m grateful for support, but no pressure.

PR #59, opened 2021-09-01
Specify GPL-2.0-only

notthewave:
If `-or-later` is not specified, it can be generally assumed that it's `-only`, this however makes it clear.


johanmalm:
Thanks.

Consistent with https://spdx.org/licenses/

PR #62, opened 2021-09-19
improve actions

bi4k8:
Implements 3 actions present in openbox but not yet implemented in labwc and changes the default target for keyboard actions to the focused window rather than the topmost mapped window.


johanmalm:
Thanks.

PR #67, opened 2021-10-03
Basic drag and drop support

ARDiDo:
I have tested this using Firefox for Wayland, and it works as expected. The only problem I have encountered was that in creating a new Firefox window, I would need to drag right or down, because dragging left or up wouldn't spawn a new window.


johanmalm:
Thanks. Much appreciated.
And thanks for chasing the wlroots API change too.

PR #68, opened 2021-10-09
partial libinput configuration

### LINKED ISSUE #63: no input configuration
As far as I can tell, there's no way to configure e.g. key repeat rates or mouse acceleration with labwc. This should just be a matter of plumbing settings from the config file into libinput.

ghost:
Adds configuration support for pointer speed, natural scrolling, left-handed mode, tap to click, pointer acceleration profiles, middle button emulation, disable while typing, keyboard repeat rate, and keyboard repeat delay. Closes #63
This only implements partial support because there is no method for device specific configuration.


johanmalm:
Sounds good 👍 
If wlroots upversion, I’ll update the documentation+tag, otherwise no rush.

PR #83, opened 2021-10-17
Implement XWayland size hints

misyltoad:
Also has a fix for janky resizing when reaching a single min extent

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -171,12 +171,15 @@ handle_set_app_id(struct wl_listener *listener, void *data)
 static void
 xdg_toplevel_view_configure(struct view *view, struct wlr_box geo)
 {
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Sec, there is an error in this PR.

Joshua-Ashton:
Fixed.

johanmalm:
Just a thought. Might it be better to implement this directly in the xwayland.c file in the move() and configure() functions.
That way it'll apply to other attempts to configure/move too - for example SnapToEdge :smile:

Joshua-Ashton:
Possibly ye, will look into it.

Joshua-Ashton:
Should handle it there now and also avoided a judder in cursor if you snapped to min extent too quickly

PR #85, opened 2021-10-17
Add basic pointer constraints

ghost:
I have tested these constraints in both Linux-native and Windows-only games that were played through the Proton compatibility layer. I didn't get a chance to test whether or not these constraints work on non-fullscreen windows, or on any games that are Wayland-native.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -23,7 +23,9 @@
 #include <wlr/types/wlr_output_damage.h>
 #include <wlr/types/wlr_output_management_v1.h>
 #include <wlr/types/wlr_output_layout.h>
+#include <wlr/types/wlr_relative_pointer_v1.h>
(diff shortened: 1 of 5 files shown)


johanmalm:
Thanks!
Were you able to re-produce any buggy/erratic behavior before this patch? Does the patch fix it?

I noted this reported issue, but can't reproduce it myself. I think my laptop is too slow to play that game.
https://github.com/puppylinux-woof-CE/woof-CE/issues/2534

I also note a couple of issues in ifreund/river which seem to relate to games require pointer constraints

Joshua-Ashton:
We should probably have somewhere to discuss so we avoid duplicating work and such :b

Joshua-Ashton:
Seems better than what I wrote. The only thing that is missing is support for absolute cursors. I will do this now.

PR #89, opened 2021-10-17
view: Add 'center' option to SnapToEdge

misyltoad:
Similar to fullscreen, but not a separate state, and with gaps.


Joshua-Ashton:
(also fixes the CI)

PR #91, opened 2021-10-18
config: Fix RaiseOnFocus always setting FollowMouse

misyltoad:

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -303,8 +303,10 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "FollowMouse.focus")) {
 		rc.focus_follow_mouse = get_bool(content);
 	} else if (!strcasecmp(nodename, "RaiseOnFocus.focus")) {
-		rc.focus_follow_mouse = true;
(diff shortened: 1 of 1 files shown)


Consolatis:
What was the reasoning behind this change? It would be quite handy to have that option while disabling FollowMouse.

johanmalm:
Guess it was to try to increase consistency with openbox, where raiseOnFocus only applies if followMouse is set.

http://openbox.org/wiki/Help:Configuration

Happy to simplify this to set them independently of each other.

@Joshua-Ashton shout if you have an issue with this.

Joshua-Ashton:
I made this change because originally because even if the option was set to false, it would enable follow mouse if it existed -- and I didn't want that.

Consolatis:
@Joshua-Ashton I completely missed the removed line (which was the whole reason for this PR) and only saw you enabled FollowMouse once RaiseOnFocus was true. Sorry :)

> consistency with openbox, where raiseOnFocus only applies if followMouse is set.  

Well it enables followMouse when raiseOnFocus is set. So its actually the other way round.

johanmalm:
I've de-couple them. c797dbe1

johanmalm:
> So its actually the other way round.

Yes :smile:

Issue #77, opened 2021-10-17
'MoveToEdge' does nothing when the center of the window it outside an output - consider handling this better


misyltoad:
`wlr_output_layout_output_at` returns `NULL` here and causes a crash.

We probably want to clamp the values here to our available space somehow.

PR #92, opened 2021-10-19
Improve view_output

ghost:
If the center of a view isn't in any output, then view_output() uses the first output it can find at one of the view's four corners. This allows 'MoveToEdge' to function even when its center is outside of an output. This may provide better handling for #77.
I haven't had the opportunity to test how well view_output() functions with multiple monitors, or how well it works with 'SnapToEdge'.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -80,23 +80,16 @@ view_minimize(struct view *view, bool minimized)
 }
 
 static struct wlr_output *
-view_available_wlr_output(struct view *view)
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
We should probably calculate the shortest distance from the centre to any output rect and use that instead

ARDiDo:
Good point. Closing this PR until I have that figured out.

johanmalm:
Thanks!

PR #100, opened 2021-10-30
update layer surfaces on output geometry changes

### LINKED ISSUE #99: Background layer geometry doesn't update when running labwc nested under Gnome wayland and changing winow size

ARDiDo:
I tested this using swaybg running under a labwc session nested under another labwc session.
Closes #99.


johanmalm:
Nice one 👌

PR #104, opened 2021-11-02
More mousebinding support

ARDiDo:
This will add support for the iconify, maximize, and close mouse contexts, as well as for the press, click, release, and double click (already supported) mouse events.
I moved the default actions of the iconify, maximize, and close buttons to click mouse bindings. This felt more natural to me.
I also changed up the formatting of labwc-config(5) a bit because `'TitleBar'` wouldn't display with mandoc's or busybox's man commands. I'm not sure if this is the case for man-db's man.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -111,7 +111,16 @@ Configuration must be wrapped in a <labwc_config> root-node.
 
 *<mouse><context name=""><mousebind button="" action="">*
 	Define a mouse binding. Supported context-names include:
-	'TitleBar'.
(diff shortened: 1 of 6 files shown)


johanmalm:
Ta

johanmalm:
@ARDiDo Is there a way we could chat offline? Maybe e-mail?

ARDiDo:
Email works for me @johanmalm. I could email you using my preferred email, or I could change my commit email address to it. Whatever works best for you.

johanmalm:
@ARDiDo that would be good. Maybe if you e-mail me that would be easiest.

PR #111, opened 2021-11-13
Fix HiDPI scaling.

### LINKED ISSUE #105: Menus, titlebar text, and rounded corners are blurred when output scale is >1
For some reason labwc does not scale the output on my laptop. I have 3000x2000 14" display and I see the following:

labwc: 
scaling-bad

sway:
scaling-good

ARDiDo:
Once it's done, this should close #105.
- [X] Scale views that originate on the HiDPI monitor
- [X] Scale views originating from non-HiDPI monitors
~~Scale menu and titlebar text~~

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -363,6 +363,8 @@ void view_for_each_surface(struct view *view,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 void view_for_each_popup_surface(struct view *view,
 	wlr_surface_iterator_func_t iterator, void *data);
+void view_output_enter(struct view *view, struct wlr_output *wlr_output);
+void view_output_leave(struct view *view, struct wlr_output *wlr_output);
 void view_move_to_edge(struct view *view, const char *direction);
 void view_snap_to_edge(struct view *view, const char *direction);
 const char *view_get_string_prop(struct view *view, const char *prop);
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -206,6 +206,7 @@ struct view {
 	enum view_type type;
 	const struct view_impl *impl;
 	struct wl_list link;
+	struct output *output;
 
 	union {
 		struct wlr_xdg_surface *xdg_surface;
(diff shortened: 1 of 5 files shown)


johanmalm:
Should the surface_*_for_each_surface() functions be static void?
Otherwise looks good

johanmalm:
Clang build fails because of an old “unused variables” in layers.c popup_damage()
I’ve left that a bit untidy because I think damage tracking will be taken care of by wlr_scene when we move to that.
Do you want me to sort out, or do you want to do it yourself? Suggest we just go with damage-all-outputs for now.

ARDiDo:
Yeah, the iterators should be static void. That was leftover from before the view_output_*() functions were made. I'll fix it in the next commit.
I'll leave fixing the build failures to you, since you know more about the situation than I do.

johanmalm:
clang-warning fixed.

heroin-moose:
Fixes alacritty for me. However, it seems that the logical screen size is now 4 times bigger than the physical one, because maximized applications are only partially visible.

heroin-moose:
Btw, labwc's menus still looks blurry.

ARDiDo:
Thanks for the feedback @heroin-moose. The latest commit fixed the scaling and placement on my end, and I would be very grateful if you could test it on your machine.
Being honest, I think I should just merge the commits I have, and open another pull request once I'm done scaling the menu and titlebar text. Fixing that involves working with pango and cairo, and I'm not familiar with those at all. I don't want to fix any merge conflicts that will eventually arise, and I want to be able to take my time on it.

johanmalm:
@ARDiDo sounds a good plan.

With the menu/etc, do you think the approach is to just make textures bigger at source by a factor of scale? And then remove line https://github.com/labwc/labwc/blob/aa9720f6273de9117694064479251d7961af7816/src/output.c#L506

Guess it might get more complicated if different outputs gave different scales.

heroin-moose:
OSD yes, maximize yes, however snaptoedge still moves window beyond the screen.

ARDiDo:
That should about do it. Thanks for the help @heroin-moose, and thank you for the advice @johanmalm.

Issue #116, opened 2021-11-23
No video when monitor turned off then back on


droc12345:
I sync'd to latest git of wayland, wayland-protocols, wlroots and labwc this morning and noticed a problem.

If I turn off the monitor and then turn it back on there is no video displayed, the screen turns on (you can see backlight), and the keyboard works, so it's not hung waiting for something. 

When I turn of monitor this is in log
info: wayland.c:647: [main.c:392] Destroying output DP-1 (Goldstar Company Ltd LG HDR 4K 0x0000371F)monitor disabled: DP-1
and when I turn on monitor, this is in log
<mode lines for monitor>
00:00:12.463 [INFO] [backend/drm/drm.c:1422] connector DP-1: Requesting modeset
00:00:12.463 [ERROR] [types/output/output.c:637] Basic output test failed for DP-1

It allows me to shut down labwc, goes back to console and displays properly.

This was fine as of my syncing on Nov 10 so some changes (to wlroots most likely) have happened since.

johanmalm:
Would you be able to track down the offending commit?

droc12345:
I'm trying to chase it down, this is what I'm seeing so far.

I can see where the monitor is disable, output destroyed,
info: wayland.c:647: monitor disabled: DP-1
2021-11-23 11:38:55 - [main.c:392] Destroying output DP-1 (Goldstar Company Ltd LG HDR 4K 0x0000371F)

then I see  udev event when the monitor is turned on, then I see
00:00:11.309 [INFO] [backend/drm/drm.c:1422] connector DP-1: Requesting modeset
00:00:11.309 [DEBUG] [../labwc-9999/src/output.c:953] set preferred mode
00:00:11.309 [DEBUG] [types/output/output.c:603] Tried to modeset a disabled output
00:00:11.309 [DEBUG] [../labwc-9999/src/output.c:965] preferred mode rejected, falling back to another mode
And a series of trying different modes, then it just sits (blank screen), until I use the quit hotkey.

droc12345:
The code needed an output enable true in src/output.c

> 955a956,957
> \> 
> \> 	wlr_output_enable(wlr_output, true);

Seems to work now, at least for that problem.

PR #118, opened 2021-11-26
Enable new outputs (before setting mode)

jlindgren90:
When `labwc` is started in its default configuration, all display outputs are initially disabled.  `sway` does not have the same issue.  In comparison with the `sway` code, it appears the `labwc` is missing a call to `wlr_output_enable()`.

Fixes the following error with `wlroots-git` 0.14.0.r391.g585a908a-1:

    [DEBUG] [types/output/output.c:603] Tried to modeset a disabled output

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -955,6 +955,9 @@ new_output_notify(struct wl_listener *listener, void *data)
 		return;
 	}
 
+	wlr_log(WLR_DEBUG, "enable output");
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. Much appreciated.
It was on my radar, just hadn’t had a chance to action.
@droc12345 this matches your note in issue #116

PR #119, opened 2021-11-26
Basic support for Drag/Move action

jlindgren90:
This makes it possible to move a window by dragging its titlebar.  Tested working with my existing Openbox config.

I'm new to the `labwc` code, so apologies if I missed anything.

Example config snippet:

    <mouse>
      <context name="Titlebar">
        <mousebind button="Left" action="Drag">
          <action name="Move"/>
        </mousebind>
      </context>
    </mouse>

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -430,6 +430,8 @@ struct view *desktop_surface_and_view_at(struct server *server, double lx,
 	double ly, struct wlr_surface **surface, double *sx, double *sy,
 	int *view_area);
 
+struct view *desktop_view_at_cursor(struct server *server);
+
 void cursor_init(struct seat *seat);
 
 void keyboard_init(struct seat *seat);
--- a/src/action.c
+++ b/src/action.c
@@ -72,6 +72,11 @@ action(struct server *server, const char *action, const char *command)
 		if (view) {
 			view_minimize(view, true);
 		}
+	} else if (!strcasecmp(action, "Move")) {
+		struct view *view = desktop_view_at_cursor(server);
+		if (view) {
+			interactive_begin(view, LAB_INPUT_STATE_MOVE, 0);
+		}
 	} else {
 		wlr_log(WLR_ERROR, "action (%s) not supported", action);
 	}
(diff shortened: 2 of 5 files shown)


johanmalm:
Hej

Thanks. Looks good.

I've not compiled + tested.

Would you be happy to take a view on the following:

Should we delete https://github.com/labwc/labwc/blob/687993370faf7b517cada18dc9415c324bc53803/src/cursor.c#L584  ??

Should we add a default mouse binding for this? https://github.com/labwc/labwc/blob/687993370faf7b517cada18dc9415c324bc53803/src/config/rcxml.c#L468

Add the new action to labwc-action(5)
https://raw.githubusercontent.com/labwc/labwc/master/docs/labwc-actions.5.scd

Add 'drag' to 'Supported mouse actions' in labwc-config(5)
https://raw.githubusercontent.com/labwc/labwc/master/docs/labwc-config.5.scd

Actually... just thinking - is 'drag' the right thing here? And if it is, are we implementing it right?
See http://openbox.org/wiki/Help:Bindings#Mouse_bindings

Would press/click be enough?
If we really want drag, should we check for movement after 'press' before we consider it a 'drag'

jlindgren90:
Hi @johanmalm and thanks for the feedback!

I checked Openbox's behavior and it does also work to bind "Move" to the "Press" signal (instead of "Drag").  So for now, I'll leave "Drag" handling out of this merge request and only add the "Move" action.

Adding a default mouse binding in `rcxml.c` is a good idea; then we can also delete the hard-coded fallback in `cursor.c`.  Good call and thanks for making me aware of those parts of the code :)

I'll push a new version shortly, which I've tested in various scenarios:
- No `<mouse>` section in `rc.xml`: dragging the titlebar works by default
- Binding (via `rc.xml`) of "Press" to only "Focus" and "Raise": dragging the titlebar is disabled (expected)
- Binding (via `rc.xml`) of "Press" to "Focus", "Raise", and "Move": dragging the titlebar works again

johanmalm:
Thanks.

johanmalm:
I've updated labwc-action.5.scd and rc.xml.all

jlindgren90:
Thanks, much appreciated! I ran out of time to update the docs yesterday.

PR #123, opened 2021-11-27
Force repaint after window title changes

jlindgren90:
When a window title changed dynamically (for example when running
"cd" within an xfce4-terminal), the titlebar did not immediately
update to show the new title.  (The titlebar would update as soon
as the mouse cursor moved.)

Calling `damage_all_outputs()` might not be the most optimal fix,
but it at least seems to work.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -158,6 +158,7 @@ handle_set_title(struct wl_listener *listener, void *data)
 	struct view *view = wl_container_of(listener, view, set_title);
 	assert(view);
 	view_update_title(view);
+	damage_all_outputs(view->server);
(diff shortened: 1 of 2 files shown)


ARDiDo:
Good catch. You might want to move the `damage_all_outputs()` call to `view_update_title()` so you don't have to repeat that line in both of the client type's set title functions.

jlindgren90:
> You might want to move the `damage_all_outputs()` call to `view_update_title()` so you don't have to repeat that line in both of the client type's set title functions.

I wasn't sure what your preference would be.  Please feel free to make that change. :+1:

PR #125, opened 2021-11-27
Fix invisible cursor at startup

jlindgren90:
The mouse cursor was invisible after starting `labwc` (and only appears once the mouse is physically moved).

The root cause seems to be that `wlroots` tracks cursor state on a per-output basis (see the `wlr_output_cursor` module) and doesn't update it correctly when outputs are changed.  So after adding a new output, the `wlr_output_cursor` on that output is invisible, while a stale cursor image is displayed on a previous output.  As soon as the mouse is moved again, the cursor image "jumps" from the previous output to the new one.

Long-term, it may be better to work on fixing the root cause of the issue in `wlroots`, but that would be a bit too involved for me at the moment.


ARDiDo:
Thanks :)

PR #126, opened 2021-11-27
Fix graphical artifacts after adding/removing outputs

jlindgren90:
When an output is added or removed, the position of the other
outputs can also change, resulting in windows moving between them.
We need to redraw all the outputs (not just the new one) to prevent
stale/corrupted screen contents.


johanmalm:
Thanks

PR #131, opened 2021-12-01
fix signal handlers

bi4k8:
avoid segfaulting sigint/sigterm handlers and drop an unused argument from sighup handler


johanmalm:
Thanks

PR #132, opened 2021-12-01
Support modifiers for mouse bindings

bi4k8:


ARDiDo:
Looks good. We might be able to remove the hardcoded alt-key + drag mousebind after this.

bi4k8:
> Looks good. We might be able to remove the hardcoded alt-key + drag mousebind after this.

That's my plan, but I think we want proper [`Frame`/`Client` mouse binding context](http://openbox.org/wiki/Help:Bindings#Context) support first (the former swallows events, the latter does not) to match the current logic which does not forward `A+Left` clicks that move the window.

bi4k8:
I've implemented the above logic in [my fork](https://github.com/bi4k8/labwc) and will PR it after the [resize PR](https://github.com/labwc/labwc/pull/135) which is likely to conflict otherwise.

PR #133, opened 2021-12-01
Parse menu names from Menu mouse-bindings

bi4k8:
These were being ignored before. The only menu currently supported by labwc (from what I could tell) is "root-menu", but even that cannot be triggered by mouse bindings without this fix.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -98,6 +98,8 @@ fill_mousebind(char *nodename, char *content)
 		current_mousebind->action = strdup(content);
 	} else if (!strcmp(nodename, "command.action")) {
 		current_mousebind->command = strdup(content);
+	} else if (!strcmp(nodename, "menu.action")) {
+		current_mousebind->command = strdup(content);
 	}
 }
 


johanmalm:
Do we need to tweak the code at

https://github.com/labwc/labwc/blob/031ced85efa771518211ff0ea5b500b75753b694/src/cursor.c#L593

or

https://github.com/labwc/labwc/blob/031ced85efa771518211ff0ea5b500b75753b694/src/action.c#L14

to make this work?

bi4k8:
The string "root-menu" should be put in a default binding if we want to leave behavior unchanged here.

bi4k8:
i.e., 
```diff
diff --git a/src/config/rcxml.c b/src/config/rcxml.c
index 1eb93ae..0ec50fd 100644
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -470,6 +470,9 @@ static struct {
        { "Close", "Left", "Click", "Close", NULL },
        { "Iconify", "Left", "Click", "Iconify", NULL},
        { "Maximize", "Left", "Click", "ToggleMaximize", NULL},
+       { "Root", "Left", "Press", "ShowMenu", "root-menu"},
+       { "Root", "Right", "Press", "ShowMenu", "root-menu"},
+       { "Root", "Middle", "Press", "ShowMenu", "root-menu"},
        { NULL, NULL, NULL, NULL, NULL },
 };
 
```

But that requires support for mouse bindings on the "Root" context (which I've prototyped but not yet PR'd).

johanmalm:
Super

PR #136, opened 2021-12-01
Implement Frame, Client, and Root mouse binding contexts

bi4k8:
This changes mouse event handling to better match Openbox (in that mouse bindings on the Frame context swallow the event while others do not) and support more mouse contexts.

After these changes, mouse bindings should be expressive enough to remove some hard-coded logic in favor of default bindings.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -550,6 +550,11 @@ cursor_button(struct wl_listener *listener, void *data)
 			cursor_rebase(&server->seat, event->time_msec);
 			return;
 		}
+
+		/* Handle _release_ on root window */
+		if (!view) {
+			handle_release_mousebinding(server, event->button, modifiers, LAB_SSD_ROOT, 0);
+		}
 		goto mousebindings;
 	}
 
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -421,47 +421,36 @@ cursor_motion_absolute(struct wl_listener *listener, void *data)
 	process_cursor_motion(seat->server, event->time_msec);
 }
 
-static void
-handle_cursor_button_with_meta_key(struct view *view, uint32_t button,
-		double lx, double ly)
-{
-	if (!view) {
-		return;
-	}
-
-	/* move */
-	if (button == BTN_LEFT) {
-		interactive_begin(view, LAB_INPUT_STATE_MOVE, 0);
-		return;
-	}
-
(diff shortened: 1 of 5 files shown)


src/cursor.c
```diff
@@ -421,47 +421,36 @@ cursor_motion_absolute(struct wl_listener *listener, void *data)
...
+				case MOUSE_ACTION_RELEASE:
+					break;
+				case MOUSE_ACTION_CLICK:
+					if (mousebind->pressed_in_context) {
+						mousebind->pressed_in_context = false;
+						break;
+					}
+					continue;
+				default:
+					continue;
```
bi4k8, line 445: This logic requires that `wl_list_for_each_reverse` supports `continue` like a native loop.

bi4k8:
I think this may interfere with buttons in decorations at present; this will need more testing before it's ready for merge.

ARDiDo:
Gotcha. Take your time.

bi4k8:
It turns out that buttons in decorations were broken because of the default mouse binding `{ "TitleBar", "Left", "Press", "Move", NULL },`, which is itself not wrong, but caused the "Move" action to begin, which in turn changes `server->input_mode` to become `LAB_INPUT_STATE_MOVE`. This triggers this conditional in `cursor_button`'s release handling:

```C
		if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
			/* Exit interactive move/resize/menu mode. */
			server->input_mode = LAB_INPUT_STATE_PASSTHROUGH;
			cursor_rebase(&server->seat, event->time_msec);
			return;
		}
```
...which cancels the rest of the handling of the mouse release event, which would otherwise have completed the "click" for the maximize/minimize/close buttons.

A one-line fix is to remove the `return` in the above conditional--control then reaches `goto mousebindings;` and we handle mouse bindings before finally returning. However, this fix would mean that, in theory, other mouse bindings could be activated while in a menu or while moving/resizing windows, which I don't think we want. In practice, though, I couldn't find any scenario where this is really a problem; maybe a contrived config could hit it.

The more correct way to deal with this issue is to implement "drag" mouse action support, but I would rather not open that can of worms during this PR.

Thoughts?

bi4k8:
I've pushed the "remove `return`" fix mentioned above; I think this PR is now good to merge.

PR #138, opened 2021-12-02
[bugfix] forward click events to layer surfaces

bi4k8:
I just noticed this was broken as fallout from my last PR. Sorry about the regression.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -574,6 +574,8 @@ cursor_button(struct wl_listener *listener, void *data)
 		if (layer->current.keyboard_interactive) {
 			seat_set_focus_layer(&server->seat, layer);
 		}
+		wlr_seat_pointer_notify_button(seat->seat, event->time_msec,
+			event->button, event->state);
 		return;
 	}
 


johanmalm:
Thanks for implementing all this. It was a change I had been putting off because I thought it was going to be a bit messy.
I did just pull, and realised the cursor is always left_ptr when over client. Any ideas whats causing it?

bi4k8:
> Thanks for implementing all this. It was a change I had been putting off because I thought it was going to be a bit messy. I did just pull, and realised the cursor is always left_ptr when over client. Any ideas whats causing it?

@johanmalm Thanks for catching that--the issue seems to be that there are now `ssd_part_type`s other than `LAB_SSD_ROOT` that are not part of window decorations (specifically, `LAB_SSD_CLIENT`). This seems to fix things:

```diff
diff --git a/src/cursor.c b/src/cursor.c
index c1c08e8..d2e8678 100644
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -202,7 +202,7 @@ process_cursor_motion(struct server *server, uint32_t time)
                        cursor_name_set_by_server = true;
                        cursor_set(&server->seat,
                                wlr_xcursor_get_resize_name(resize_edges));
-               } else if (view_area != LAB_SSD_NONE) {
+               } else if (view_area != LAB_SSD_NONE && view_area != LAB_SSD_CLIENT) {
                        /* title and buttons */
                        cursor_set(&server->seat, XCURSOR_DEFAULT);
                        cursor_name_set_by_server = true;
```

But it would be cleanest to introduce a "ssd_part_is_decoration" helper, probably.

PR #142, opened 2021-12-03
interactive: Fix crash if a window is destroyed while moving it

Joshua-Ashton:
Need to clean this up if it's our grabbed window!


johanmalm:
Nice.

PR #145, opened 2021-12-03
action: If we have a view that is an activator, use that instead of t…

Joshua-Ashton:
…he current focus window

If we don't switch focus, we want the close button to close the window associated with it, not the current focus window.


johanmalm:
Thanks Joshie.
Yes, agree - we need to be able to pass `struct view` as an argument.
It feels like we should simplify the action() interface... it's getting too many arguments, some of them optional. Any suggestions?

Issue #98, opened 2021-10-26
Stuck alt-tab cycle-window OSD


johanmalm:
It's been reported that someone got stuck in the alt-tab cycler and that has higher priority than switching to a tty so couldn't escape it.

Need to find a way of reproducing the issue before trying to fix it.

Worth considering:
1. Maybe need to move the cycle_view related code from keyboard_modifiers_notify() to handle_compositor_keybindings().
2. After handle_keybinding() is called for WL_KEYBOARD_KEY_STATE_PRESSED, maybe we need to prevent the equivalent WL_KEYBOARD_KEY_STATE_RELEASED being forwarded to clients.

01micko:
This bug affects my productivity almost daily so @ARDiDo , @Joshua-Ashton any ideas? The `sway/input/keyboard.c` line does look like a way forward to me, but I really have no idea.

01micko:
More info as I've been capturing unique logs over past weeks. (only last 10 lines posted each log).
Note that I at least 3 times a week to latest git version and do run a patch to run as root; see https://github.com/puppylinux-woof-CE/woof-CE/issues/2563. The error is quite consistent and persistent!

`labwc -d >> /var/log/labwc${$}.log`

```log

/var/log/labwc10789.log
00:00:00.001 [INFO] [libseat] [libseat/libseat.c:67] Seat opened with backend 'builtin'
00:00:00.001 [INFO] [libseat] [libseat/backend/seatd.c:222] Enabling seat
00:00:00.001 [INFO] [backend/session/session.c:110] Successfully loaded libseat session
00:00:00.003 [INFO] [backend/backend.c:219] Found 1 GPUs
00:00:00.003 [INFO] [backend/drm/backend.c:183] Initializing DRM backend for /dev/dri/card0 (radeon)
00:00:00.003 [DEBUG] [backend/drm/drm.c:81] Atomic modesetting unsupported, using legacy DRM interface
00:00:00.003 [DEBUG] [backend/drm/drm.c:98] ADDFB2 modifiers unsupported
00:00:00.003 [INFO] [backend/drm/drm.c:259] Found 2 DRM CRTCs
00:00:00.003 [INFO] [backend/drm/drm.c:186] Found 2 DRM planes
00:00:00.003 [ERROR] [../src/server.c:80] unable to drop root

/var/log/labwc12570.log
00:00:18.013 [DEBUG] [types/output/cursor.c:395] Falling back to software cursor on output 'DVI-D-1'
[...]

PR #149, opened 2021-12-04
Alt-Tab switcher improvements

jlindgren90:
A couple of improvements to the Alt-Tab window switcher:

- Pressing Alt-Tab a single time should switch to the next window
- Pressing Escape should cancel the window switcher

These changes both make `labwc` work more like OpenBox and other stacking WMs.

Thanks!

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -140,7 +140,6 @@ handle_compositor_keybindings(struct wl_listener *listener,
 					desktop_cycle_view(server, server->cycle_view,
 						backwards ? LAB_CYCLE_DIR_BACKWARD : LAB_CYCLE_DIR_FORWARD);
 				osd_update(server);
-				damage_all_outputs(server);
 			}
 		}
 		/* don't send any key events to clients when osd onscreen */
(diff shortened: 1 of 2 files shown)


src/keyboard.c
```diff
@@ -132,7 +140,6 @@ handle_compositor_keybindings(struct wl_listener *listener,
 					desktop_cycle_view(server, server->cycle_view,
 						backwards ? LAB_CYCLE_DIR_BACKWARD : LAB_CYCLE_DIR_FORWARD);
 				osd_update(server);
-				damage_all_outputs(server);
```
bi4k8, line 135: Why is this call removed? Is there a reason it's actually not necessary, or was it causing misbehavior?

bi4k8:
Looks good--one question inline.

src/keyboard.c
ARDiDo: From what testing I've done, it looks like it isn't necessary.

ARDiDo:
Thanks @jlindgren90. I haven't personally been able to reproduce the stuck alt-tab bug, but this should alleviate some headaches for those who do experience it.
And thanks for the code review @bi4k8.

src/keyboard.c
jlindgren90: It appeared to be unnecessary since `damage_all_outputs()` was already called on line 120.
Thanks @ARDiDo for confirming!

01micko:
So far working well for me. I'll test for a few hours and if good will close #98

PR #154, opened 2021-12-06
[bugfix] Query passed prop, not 'title'

bi4k8:
This was causing (among other issues) windows from foot, chromium, and other programs that change their titles to lose their window icons when minimized and reactivated using waybar.


ARDiDo:
Looks good. Thanks for the fix.

johanmalm:
Whoops :smile:

PR #156, opened 2021-12-11
Implement corner and edge mouse contexts, improve docs, and unhardcode corner/edge resize

bi4k8:


ARDiDo:
Thanks.
I gave it a test run, and the only problem I could find was that when resizing from an edge, I couldn't resize just the width or just the height. It was almost like I was resizing from the nearest corner.

bi4k8:
> Thanks. I gave it a test run, and the only problem I could find was that when resizing from an edge, I couldn't resize just the width or just the height. It was almost like I was resizing from the nearest corner.

The reason for this is that we now use the same resize edges for both resize actions that begin on the window and resize actions that begin on the frame. Resize actions on the window have the following code (in `cursor_button` in `cursor.c`) that picks the nearest corner for in-window clicks that trigger resize actions:
```c
        resize_edges |= server->seat.cursor->x < view->x + view->w / 2 ? WLR_EDGE_LEFT : WLR_EDGE_RIGHT;
        resize_edges |= server->seat.cursor->y < view->y + view->h / 2 ? WLR_EDGE_TOP : WLR_EDGE_BOTTOM;
 ```
And this code now also applies for window border events.

We could try to move cursor location handling into the code that executes actions, but that's tangling up unrelated notions. The right way to fix this is, I think, what Openbox also does: having resize actions specify the edge/corner to use (see [its docs](http://openbox.org/wiki/Help:Actions#Resize)). I can implement that in this PR, too; it would be a cleaner change overall.

**Comment by ARDiDo** [comment]
Sounds good. I'll hold off on merging until you're ready.
Thanks for all you help.

File: src/ssd.c [hunk 1]
```diff
@@ -423,10 +423,22 @@ ssd_part_contains(enum ssd_part_type whole, enum ssd_part_type candidate)
 		return true;
 	}
 	if (whole == LAB_SSD_PART_TITLEBAR) {
-		return candidate >= LAB_SSD_BUTTON_CLOSE && candidate <= LAB_SSD_PART_CORNER_TOP_RIGHT;
+		return candidate >= LAB_SSD_BUTTON_CLOSE && candidate <= LAB_SSD_PART_TITLE;
 	}
 	if (whole == LAB_SSD_FRAME) {
 		return candidate >= LAB_SSD_BUTTON_CLOSE && candidate <= LAB_SSD_CLIENT;
 	}
+	if (whole == LAB_SSD_PART_TOP) {
```
johanmalm, line 431: Is there a cleaner implementation that avoids using sdd_part_contains()?

What is the situation that you can't handle by just iterating over them and get to the right one? When the parts are created, they are carefully ordered so that rendering / pressing should just be able to iterate over the list and stop at the first match.

johanmalm:
Am I right in thinking we ought to hold off releasing `0.4.0` until we've merged this one?

ARDiDo:
I don't think it would hurt to wait for this PR.
We've been making some good progress recently on unhardcoding many configuration options, and it might be nice to be able to cross off this as well before `0.4.0`.

johanmalm:
I've pushed a tweak that I think sorts the setting of `resize_edges`.

@bi4k8 A bit thank you for implementing `<mousebinds>` it's a been a lot of quite intrusive work, but is definitely worth it :smile:

PR #159, opened 2021-12-11
[bugfix] Don't crash when a minimized window closes

bi4k8:
To reproduce, open an editor from a terminal, minimize it, then hit ^C in the terminal.

The crash can be avoided if we simply don't unmap xdg or xwayland views if they are not currently marked as mapped.


ARDiDo:
Haven't been able to reproduce that myself, but I'll take your word for it.
Thanks for finding and fixing this bug.

PR #160, opened 2021-12-12
Add tapButtonMap libinput setting

bi4k8:
This is a standard libinput option that wasn't exposed yet. Docs included.


ARDiDo:
Ah, looks like I forgot that one. Thanks.

PR #164, opened 2021-12-14
Invalidate mouse focus when cycling windows

### LINKED ISSUE #162: Alt-tab does not change which window gets scroll events
This is a surprising one, but probably there's just some kind of focus setting missing in the alt-tab code.

To reproduce:
1. open two windows that overlap but where neither is contained by the other
2. put the cursor in the overlap region; notice that scrolling goes to the window on top
3. without moving the cursor, hit alt-tab to switch which window is on top and then scroll again
4. notice that the window now in the back (without keyboard focus) scrolls instead of the newly-focused, raised window

Moving the cursor even a single pixel makes subsequent scroll events go to the newly-focused, raised window as expected.

bi4k8:
Fixes #162.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -46,6 +46,7 @@ keyboard_modifiers_notify(struct wl_listener *listener, void *data)
 			desktop_focus_and_activate_view(&server->seat,
 				server->cycle_view);
 			desktop_move_to_front(server->cycle_view);
+			process_cursor_motion(server, event->time_msec);
 			server->cycle_view = NULL;
 		}
 	}
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -440,6 +440,14 @@ void cursor_rebase(struct seat *seat, uint32_t time_msec);
  */
 void cursor_set(struct seat *seat, const char *cursor_name);
 void cursor_init(struct seat *seat);
+/**
+ * process_cursor_motion - update cursor icon for and send enter events for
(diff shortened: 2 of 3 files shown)


src/keyboard.c
```diff
@@ -46,6 +46,7 @@ keyboard_modifiers_notify(struct wl_listener *listener, void *data)
 			desktop_focus_and_activate_view(&server->seat,
 				server->cycle_view);
 			desktop_move_to_front(server->cycle_view);
+			process_cursor_motion(server, event->time_msec);
```
johanmalm, line 49: I've not had a chance to think this through, so am asking this as a question rather than a well thought-through statement.

1. Would it feel better to do the function all to update the cursor focus from desktop_move_to_front()? I think that's the function that moves windows around and therefore requires a cursor 'refocus'
2. If we call it from desktop_move_to_front(), we wouldn't have the event time, so would be better creating a wrapper function and use `clock_gettime(CLOCK_MONOTONIC, &now)` or similar. That's probably better anyway because it avoid using the time of the key event, which might be a bit before the windows had a chance to move around.
3. When making functions public, I like to use the filename as a prefix in the function name. So, might a good public function name for this wrapper be cursor_update_focus()?

johanmalm:
Hi - thanks for discovering, suggesting and writing this. I preferred to re-structure a little, so re-wrote. Hope that's okay. Credit it all to you for doing the hard work :smile:
Pushed on master branch.

PR #165, opened 2021-12-16
[bugfix] Fix crash in osd.c when props are NULL

bi4k8:
This happened while running Xwayland *over* waypipe, so is likely not a common case, but is possible.


ARDiDo:
Looks good. Thanks for finding + fixing it.

Issue #166, opened 2021-12-18
Sometimes the clipboard doesn't work


Joshua-Ashton:
Sometimes the clipboard just doesn't work and it hangs whatever application I ctrl + v'ed into for a few seconds before giving up and pasting absolutely nothing.

Seems to happen randomly at some point but persists until restarting labwc. Super annoying.

ARDiDo:
Have you had this problem with sway at all?
From what I can see, wlroots compositors just need to make a `wlr_data_device_manager` using `wlr_data_device_manager_create()`, and they're good to go. If you have problems with sway also, then it's a wlroots problem.

johanmalm:
Maybe it's the released keys associated with keybinds that are playing havoc. I will try to fix when I get some time on the computer.
@Joshua-Ashton can you reproduce it on a new session if you avoid using any compositor-keybinds (i.e. to prove the theory above).

Unless  57cc72b49 fixed it.

Consolatis:
In the wayland clipboard protocol the compositor only relays messages from application A to application B containing file descriptors.

If you paste in application B that application will usually create a pipe and send the write end of that pipe to the compositor which then forwards that FD to application A. If application A then is doing nothing at all with that FD application B might hang because its trying to read from the pipe. In your case your application at least seems to have some kind of timeout and isn't stuck completely.

I don't think issues with the clipboard in wayland have anything to do with the compositor but instead everything to do with applications / toolkits.

Example for Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1731511

johanmalm:
Linked to this?
https://github.com/labwc/labwc/commit/44082ad80a88588b1f11ae8a32310867dea1257a

Consolatis:
This is crazy. Ordering of globals in wl_display.registry shouldn't matter at all. If it does its a bug in the appliction.
I tried to write a simple clipboard manager as a testcase for my own Python implemention of the wayland wire protocol and was fetching *all* offered mime types, just like in the Firefox bug I linked above. All kind of things happen when you do that. Usually just fetching text/plain;charset=utf8 works fine though.

Consolatis:
Does this still happen and if it happens how many fds are open at that time by labwc?
`ls /proc/$(pidof labwc)/fd/ | wc -l`

What does `ulimit -Sn` and `ulimit -Hn` say in the shell before you spawn labwc?
Doing `ulimit -n $(ulimit -Hn)` before spawing labwc may help if its indeed an fd limit.

Issue #146, opened 2021-12-03
Selection in alacritty broken by latest commits


Joshua-Ashton:
Doing selection in alacritty doesn't stop doing it after you release the mouse anymore

bi4k8:
Can you reproduce this behavior with `labwc -c /dev/null` (to use the default built-in configuration)?

if not, you'll need to include your `rc.xml` to help reproduce and fix this.

ARDiDo:
I can replicate this, but I'm not entirely sure this is a labwc problem. I experience the exact same problem on other wayland compositors, as well as on Xorg.

johanmalm:
This is working fine for me now. I'll close for now assuming it's overtaken-by-events, but please do say if you don't think it is.

PR #167, opened 2021-12-21
keyboard: absolve release event if press was bound

johanmalm:
When key press events are handled by compositor keybindings, do not
forward the corresponding release events to clients.


johanmalm:
@ARDiDo would you be happy to review and merge if you agree.

May solve #98 #166 #146 but I cannot reproduce any of them, so need help to test (@01micko @Joshua-Ashton - if you've got time)

ARDiDo:
This looks good. Great work.

01micko:
Sorry to be late, but I did test! All good here.

PR #168, opened 2021-12-23
config/rcxml: Allow multiple <action>s inside of a <mousebind>

Consolatis:
Issue arises when using the default config from docs/rc.xml.all.
Without this patch only the last action defined inside a `<mousebind>` will have an effect.

Without a config or when defining the same `<mousebind>` multiple times with each containing only a single `<action>` the issue does not exist.


johanmalm:
Thanks. I like this one :)

PR #170, opened 2021-12-23
Implement foreign toplevel close

Consolatis:
I've basically copied #140 + the action handler for Close.

While doing that I also wrapped view->impl->close(view) in a view_close(view) function which AFAIU seems to be the desired design. The action handler for Close now uses view_close(view) as well.

I've only tested the PR using mate-panel (including context menu PR) but it should work with any foreign-toplevel client.


ARDiDo:
Looks good. Thanks for your help!

johanmalm:
Thanks.
Yes, view_close() is better.

Issue #174, opened 2021-12-24
Alt-Tab after disabling output causes segfault


jlindgren90:
`labwc` crashes immediately when pressing Alt-Tab after an output has been disabled.  100% reproducible for me.

`labwc` was built locally from commit fd09e26ac896.
`wlroots` from https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/812951f5bc47.

Steps to reproduce:

1. Start `labwc` in a triple-monitor setup (see `wlr_randr` output below)
2. Open a terminal window (`xfce4-terminal`, but I don't expect it matters which)
3. Disable the (logically) center monitor: `wlr_randr --output eDP-1 --off`
4. Press Alt-Tab

Backtrace:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000055930bb8dae9 in render_osd (server=0x7ffe6089f130, damage=0x7ffe6089e7a0, output=<optimized out>)
    at ../src/output.c:530
530				.x = ol_output->x + o->wlr_output->width
(gdb) thread apply all bt full

Thread 1 (Thread 0x7fd2d2ed4e00 (LWP 11839)):
#0  0x000055930bb8dae9 in render_osd (server=0x7ffe6089f130, damage=0x7ffe6089e7a0, output=<optimized out>) at ../src/output.c:530
        box = {x = 5454, y = 704, width = 612, height = 32}
        o = 0x55930c3bc1a0
        layout = 0x55930c2d9370
        ol_output = 0x0
        server = 0x7ffe6089f130
        wlr_output = <optimized out>
        renderer = 0x55930c2a9b10
        color = {0, 0, 0, 1}
        view = <optimized out>
        output_height = 1619650464
        rects = <optimized out>
        frame_damage = {extents = {x1 = 5454, y1 = 704, x2 = 612, y2 = 32}, data = 0x55930c349d10}
        nrects = 1
        output_width = 0
        transform = <optimized out>
        output = 0x55930c349d10
        needs_frame = true
        damage = {extents = {x1 = 0, y1 = 0, x2 = 2560, y2 = 1440}, data = 0x0}
        frame_data = {when = {tv_sec = 140730518071200, tv_nsec = 3553}}
#1  output_render (damage=0x7ffe6089e7a0, output=<optimized out>) at ../src/output.c:802
        server = 0x7ffe6089f130
        wlr_output = <optimized out>
        renderer = 0x55930c2a9b10
        color = {0, 0, 0, 1}
[...]

PR #175, opened 2021-12-24
Check the return value of wlr_output_layout_get() for NULL

jlindgren90:
`wlr_output_layout_get()` seems to return `NULL` for disabled outputs.

Fixes: #174


johanmalm:
Thanks

PR #180, opened 2021-12-28
view:c Do not overwrite unmaximized dimensions on fullscreen

Consolatis:
Before: window -> maximize -> fullscreen -> unfullscreen would reset pre maximized state
Now: window -> maximize -> fullscreen -> unfullscreen -> unmaximize works as expected

This feels like a hack though.
I tried something like this as well:
```
view_move_resize(view, view->unmaximized_geometry);
if (view->maximized) {
    view->maximized = false;
    view_maximize(view, true);
}
```
but view->x and friends were still based on the fullscreen dimensions so it again overwrote the original premaximized state.
That would also cause a double resize which may not look that great for the user.

Maybe somebody has a better idea how to handle that (one could add something like view->unfullscreen_geometry but that feels like overkill).

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -222,10 +222,12 @@ view_set_fullscreen(struct view *view, bool fullscreen,
 			view->toplevel_handle, fullscreen);
 	}
 	if (fullscreen) {
-		view->unmaximized_geometry.x = view->x;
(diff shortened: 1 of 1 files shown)


ARDiDo:
It looks good to me.
Thanks for the fix!

PR #181, opened 2021-12-28
src/foreign.c: verify associated view is still valid

### LINKED ISSUE #172: foreign.c maybe missing checks if view still exists
As far as I understand, all handle_toplevel_handle_request_* callbacks are called once a wayland message for that particular request is received.
The view might have been destroyed in the meantime and thus wl_container_of(listener, ..) would return NULL (I assume) which then gets dereferenced.
I don't know enough about wlroots / labwc to be sure if that is possible but if yes all of those handlers should have a check like
```
if (!view) {
    /* Maybe log */
    return;
}
```

If that is indeed the proper way (and place) to deal with that race condition I'd add those checks and send a PR.

Consolatis:
Closes #172

I switched to `if (view) { .. }` because it saves a return line but I could change the PR if you prefer the check with return directly after `wl_container_of(..)`.

### DIFF
--- a/src/foreign.c
+++ b/src/foreign.c
@@ -7,7 +7,9 @@ handle_toplevel_handle_request_minimize(struct wl_listener *listener, void *data
 	struct view *view = wl_container_of(listener, view,
 		toplevel_handle_request_minimize);
 	struct wlr_foreign_toplevel_handle_v1_minimized_event *event = data;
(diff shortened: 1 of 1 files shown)


ARDiDo:
`if (view) { . . }` works perfectly.
Thanks for adding this check.

PR #184, opened 2021-12-31
Remaximize views when toggling decorations

### LINKED ISSUE #137: toggling decoration on a maximized window puts decoration outside usable screen area
This is an unfortunate interaction between maximize and decoration-toggling that I think just needs to be handled as a corner case.

To reproduce:
- run some taskbar, e.g. waybar
- open a normal client
- disable the client's decoration with a key binding
- maximize the client
- reenable the client's decoration

The decoration will overlap the taskbar instead of resizing the window to the same size an already-decorated window becomes when maximized.

ARDiDo:
This will resize the window to the proper boundary.
Fixes #137.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -191,6 +191,10 @@ view_toggle_decorations(struct view *view)
 {
 	view->ssd.enabled = !view->ssd.enabled;
 	ssd_update_geometry(view, true);
+	if (view->maximized) {
+		view_maximize(view, false);
+		view_maximize(view, true);
+	}
 }
 
 void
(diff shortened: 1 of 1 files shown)


Consolatis:
When #185 is merged this should be changed to  `view_apply_maximized_geometry(view)` so we don't have to double resize

ARDiDo:
Thank for the suggestion @Consolatis. https://github.com/labwc/labwc/commit/0f9374c701e3f11c450cf2d9a087d058960ecc78 implements this.

Issue #177, opened 2021-12-26
"Lost" windows after disabling monitor


jlindgren90:
Steps to reproduce:

1. Start `labwc` in a dual-monitor (side by side) setup
2. Move a window to the right-hand monitor
3. Disable the right-hand monitor using `wlr-randr`

The window becomes "lost" -- that is, there is no way to reach it with the mouse and move it back into the visible area.  It's still visible in the Alt-Tab window list, however.

Ideally, `labwc` would move the window automatically so that at least part of its titlebar is visible.

johanmalm:
Maybe we could handle this in a similar way to issue #137 

How does the compositor know when an output it `turned off` by a screen-locking client (and therefore windows should not be re-arranged) versus when the user actually wants to disable a particular output (and therefore wants the windows re-arranged)?

droc12345:
It would be hard to tell (impossible?) why the monitor is turned off (screen-lock or using wlr-randr) 
The system, even at kernel level will simply tell you the monitor is off or disabled, not who or why.

So, if you turn the monitor back on, does the window still display where it was? 
If so, then it sounds like it's working the way it's supposed to. 
If the window doesn't reappear on the now enabled second monitor then there is a problem, but a different kind.

jlindgren90:
The window displays where it was before if I enable the monitor again.

Under X11 there is a concept of a "disabled" monitor (`xrandr --off`) which is separate from a monitor turned off to save power (`xset dpms force off`).  The same is true in e.g. Microsoft Windows.  Waking up/unlocking the system turns back on monitors which were turned off to save power, but leaves "disabled" monitors turned off.

Does Wayland not make such a distinction?  Or is `wlr-randr` the wrong tool to "disable" a monitor?

droc12345:
Nothing wrong with wlr-randr, you could use wdisplays, but I think it would do the same thing.

From a system level, I can tell dpms is on/off, monitor is enabled/disabled and whether connected or not.
That's the only statuses that are available.

I'm not sure whether wlr-randr's off option does a disconnect or a disable. I'll have to check later.

droc12345:
This was all done in a labwc session.

wlr-randr and wdisplay mess with the enable option.

Normal -- enabled=enabled status=connected
randr off - enabled=disabled status=connected
power off - enabled=enabled status=connected 

Other compositors do different things (no one is more right than anyone else) to the enabled and status states.

I don't have a screen locker installed and don't know what it does to the states mentioned above.
[...]

PR #185, opened 2021-12-31
Adjust views to account for output layout changes

jlindgren90:
labwc currently doesn't handle output layout changes very well:

 - Windows can end up "lost" completely offscreen
 - Maximized/fullscreen windows can end up spanning multiple outputs

Currently, `new_output_notify()` and `output_destroy_notify()` contain logic
to update the cursor and force a repaint when outputs are added or
removed.  This logic in fact needs to run on any output layout change,
so consolidate it into a new function, `output_update_for_layout_change()`.

Then add a second new function, `view_adjust_for_layout_change()`, which
adjusts window placement to account for the new layout.

The behavior is roughly as follows:

 - Normal windows that end up offscreen are centered on the closest
   output (making use of the existing `view_center()` logic)
 - Maximized windows are re-maximized on the closest output.  Logic is
   also added to the unmaximize step to check that the original
   unmaximized position is still on-screen.
 - Fullscreen windows are re-fullscreened on the same output if
   possible; otherwise they are un-fullscreened.

Minimized windows don't require any special handling.  Their placement
is adjusted just the same, but invisible to the user until they are
later unminimized.

There is some positioning glitch still with un-fullscreening a window
whose output has been disconnected/disabled; it can end up in an
unexpected position (but at least has the correct size and decoration).
I don't think this is due to a bug in my change per se, but perhaps the
change has exposed a bug elsewhere.

Fixes: #177

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -370,6 +370,7 @@ void view_toggle_maximize(struct view *view);
 void view_toggle_decorations(struct view *view);
 void view_set_decorations(struct view *view, bool decorations);
 void view_toggle_fullscreen(struct view *view);
+void view_adjust_for_layout_change(struct view *view);
 void view_for_each_surface(struct view *view,
 	wlr_surface_iterator_func_t iterator, void *user_data);
 void view_for_each_popup_surface(struct view *view,
--- a/src/output.c
+++ b/src/output.c
@@ -940,9 +940,6 @@ output_destroy_notify(struct wl_listener *listener, void *data)
 	struct output *output = wl_container_of(listener, output, destroy);
 	wl_list_remove(&output->link);
 	wl_list_remove(&output->destroy.link);
-
-	/* Windows may have moved; redraw all outputs */
-	damage_all_outputs(output->server);
 }
 
 static void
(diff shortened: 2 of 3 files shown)


johanmalm:
Thanks. 
It is getting a bit late for reviewing here, so will have a look tomorrow.
A question in the meantime: If we were to consider output->usable_area (possibly every time arrange_layers() is called to reflect changes in exclusive layer-shell clients), would we approach this differently - or is this patch still the way to go?

Consolatis:
This looks nice. I had a quick test for maximize->fullscreen->unfullscreen-unmaximize and it works as expected :)
For some users it might be the desired behavior to keep some windows on an offscreen monitor which is expected to come back sooner or later. Maybe we'd need some config which disables that moving on output loss.

johanmalm:
Assume screensaver case should be dealt with using https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-output-power-management-unstable-v1.xml

…and a client like https://git.sr.ht/~leon_plickat/wlopm

jlindgren90:
> A question in the meantime: If we were to consider output->usable_area (possibly every time arrange_layers() is called to reflect changes in exclusive layer-shell clients), would we approach this differently - or is this patch still the way to go?

I think this would provide a good start for handling `usable_area` changes. It looks like only maximized windows care about the change, so one could loop over those views and call `view_apply_maximized_geometry()`.

johanmalm:
Looks very neat :smile:
Thanks for doing this.
I'm just starting a `Roadmap.md` list. Will add `wlr-output-power-management-unstable-v1.xml` to it for screensavers support.

PR #186, opened 2021-12-31
Implement pointer gestures

bi4k8:
This is pretty straightforward plumbing but also includes some minor cursor/keyboard cleanups.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -30,6 +30,7 @@
 #include <wlr/types/wlr_relative_pointer_v1.h>
 #include <wlr/types/wlr_pointer.h>
 #include <wlr/types/wlr_pointer_constraints_v1.h>
+#include <wlr/types/wlr_pointer_gestures_v1.h>
 #include <wlr/types/wlr_seat.h>
 #include <wlr/types/wlr_server_decoration.h>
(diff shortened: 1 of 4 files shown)


bi4k8:
CI seems to have failed due to HTTP 504 from gitlab.freedesktop.org.

johanmalm:
Thanks.
Does this all relate to touch-screens? Have you been able to test it?
Do we need to add:

```
case WLR_INPUT_DEVICE_TOUCH:
		new_touch(seat, input);
		break;
```
...here:
https://github.com/labwc/labwc/blob/a837fefc80a4a77e4e0353e4b8a14ece258a642a/src/seat.c#L149

johanmalm:
And this further down...
```
	case WLR_INPUT_DEVICE_TOUCH:
			caps |= WL_SEAT_CAPABILITY_TOUCH;
			break;
```

bi4k8:
This is for pinch/zoom gestures on trackpad pointers, not touch screens. I've tested the pinch gesture and it seems to behave as expected, e.g. in Firefox. I don't know any software that uses other gestures.

johanmalm:
Got it. Thanks.

PR #195, opened 2022-01-04
Allow multiple actions at once

Consolatis:
I converted all actions to a `wl_list` containing `struct action`. That is: keybindings, mousebindings and menus.
Seems to work so far. For example this now works:
```
<keybind key="A-1">
	<action name=ToggleDecorations" />
	<action name=ToggleMaximize" />
</keybind>
```
Struct currently looks like this:
```
struct action {
	uint32_t type;
	char *arg;
	struct wl_list link;
};
```
~~Open questions:~~
~~- Should `action_create(action_str)` get a `wl_list` pointer as second argument and add itself to that list?~~
~~Currently it is the responsibility of the caller to add the created action to its list.~~

~~Do not merge the PR as is, it contains debug statements in action.c and changed the global loglevel to INFO.~~

Is the approach generally usable or should it be changed in design or implementation?

### DIFF
--- a/include/action.h
+++ b/include/action.h
@@ -1,0 +1,13 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_ACTION_H
+#define __LABWC_ACTION_H
+
+struct action {
+	uint32_t type;
+	char *arg;
+	struct wl_list link;
+};
+
+struct action *action_create(const char *action_str);
+
+#endif
--- a/src/action.c
+++ b/src/action.c
@@ -4,6 +4,78 @@
 #include "common/spawn.h"
 #include "labwc.h"
 #include "menu/menu.h"
+#include "action.h"
+
+enum action_type {
+	ACTION_TYPE_NONE = 0,
+	ACTION_TYPE_CLOSE,
+	ACTION_TYPE_DEBUG,
+	ACTION_TYPE_EXECUTE,
+	ACTION_TYPE_EXIT,
+	ACTION_TYPE_MOVE_TO_EDGE,
+	ACTION_TYPE_SNAP_TO_EDGE,
+	ACTION_TYPE_NEXT_WINDOW,
+	ACTION_TYPE_PREVIOUS_WINDOW,
+	ACTION_TYPE_RECONFIGURE,
+	ACTION_TYPE_SHOW_MENU,
+	ACTION_TYPE_TOGGLE_MAXIMIZE,
+	ACTION_TYPE_TOGGLE_FULLSCREEN,
+	ACTION_TYPE_TOGGLE_DECORATIONS,
+	ACTION_TYPE_FOCUS,
+	ACTION_TYPE_ICONIFY,
+	ACTION_TYPE_MOVE,
+	ACTION_TYPE_RAISE,
+	ACTION_TYPE_RESIZE,
+};
+
+const char *action_names[] = {
+	"NoOp",
+	"Close",
+	"Debug",
+	"Execute",
+	"Exit",
+	"MoveToEdge",
+	"SnapToEdge",
+	"NextWindow",
+	"PreviousWindow",
+	"Reconfigure",
+	"ShowMenu",
+	"ToggleMaximize",
+	"ToggleFullscreen",
+	"ToggleDecorations",
+	"Focus",
+	"Iconify",
+	"Move",
+	"Raise",
+	"Resize",
+	NULL
+};
+
+
+static enum action_type
+action_type_from_str(const char *action_name)
+{
+	for (size_t i=1; action_names[i] != NULL; i++) {
+		if (!strcasecmp(action_name, action_names[i])) {
+			return i;
+		}
+	}
+	wlr_log(WLR_ERROR, "Invalid action: %s", action_name);
+	return ACTION_TYPE_NONE;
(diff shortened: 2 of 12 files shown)


Consolatis:
I could use some help here:
In 87608594819bb6e7c0a529f71caaa0a4f4e09097 I simplified the code to map the action string to the enum.
For that I added a `ACTION_NAMES[]` array in `include/action.h` so both the enums and the strings are neatly together and future updates to add new functions are easy to do. But that causes following warnings:
```
In file included from ../src/config/rcxml.c:16:
../include/action.h:26:20: warning: ‘ACTION_NAMES’ defined but not used [-Wunused-variable]
   26 | static const char *ACTION_NAMES[] = {
      |                    ^~~~~~~~~~~~
..
In file included from ../src/menu/menu.c:21:                           
../include/action.h:26:20: warning: ‘ACTION_NAMES’ defined but not used [-Wunused-variable]
   26 | static const char *ACTION_NAMES[] = {                                                                                  
      |                    ^~~~~~~~~~~~
```
That makes sense but rcxml.c and menu.c require the definitions of `struct action` and `action_create()`.
I could just put the `ACTION_NAMES[]` array into action.c but that defeats the point of having them both together.
Now I tried to forward declare `enum action_type` `struct action` and `action_create()` in labwc.h but that resulted in
```
error: field ‘type’ has incomplete type
```

Any hints how to approach this?

src/action.c
```diff
@@ -26,85 +54,119 @@ activator_or_focused_view(struct view *activator, struct server *server)
 }
 
 void
-action(struct view *activator, struct server *server, const char *action, const char *command, uint32_t resize_edges)
+action(struct view *activator, struct server *server, struct wl_list *actions, uint32_t resize_edges)
 {
-	if (!action)
+	wlr_log(WLR_INFO, "action() start");
+
+	if (!actions) {
```
johanmalm, line 61: Should this be `wl_list_empty(...)`?

include/action.h
```diff
@@ -0,0 +1,56 @@
...
+	ACTION_TYPE_TOGGLE_FULLSCREEN,
+	ACTION_TYPE_TOGGLE_DECORATIONS,
+	ACTION_TYPE_FOCUS,
+	ACTION_TYPE_ICONIFY,
+	ACTION_TYPE_MOVE,
+	ACTION_TYPE_RAISE,
+	ACTION_TYPE_RESIZE,
+};
+
+static const char *ACTION_NAMES[] = {
```
johanmalm, line 26: It would be better to put this in `action.c` like this:

```
static struct {
    const char *name;
    enum action_type;
} action_names[] = {
    { "None", ACTION_TYPE_NONE },
    { "Close", ACTION_TYPE_CLOSE },
    ...etc
    { NULL, NULL },
};
```

You could then iterate over it in `action_type_from_str()` like so:

```
for (size_t i=0; action_name[i].name; ++i) {
    ...
}
```
This approach will avoid the warning.

src/action.c
Consolatis: No. That one just guards against invalid internal calls which provide NULL.
If the list is empty it will just not enter the loop body.

include/action.h
Consolatis: That is much better. Will do.

johanmalm:
Great work.
Ref `ACTION_NAMES[]`, I've suggested an approach in the review. It's only the enums that need to sit in the header-file.

Consolatis:
> Ref `ACTION_NAMES[]`, I've suggested an approach in the review. It's only the enums that need to sit in the header-file.

Well.. I was a bit too fast with that reply.
That would mean to have two data structures that must be synced in two different files.
I'd like to prevent that and keep them both together.

Using a struct array instead of char array is much better as it allows to use alternatives (Minimize, Iconify, Hide, Unmap) which all point to ACTION_TYPE_ICONIFY. It's not required right now but it may come in handy in the future.

Consolatis:
I moved the enums and the string array into action.c and used a uint32_t for the type value in `struct action`.
That way there is a single file to update all 3 required parts if adding a new action: `enum action_type`, `action_names[]` and `action()` handler.

As no other part of labwc should interact with the enums or strings anyway it is all nicely incapsulated in a single file.

Consolatis:
I am satisfied with the current state of the PR, I would just need to remove the debug statements and squash the commits.

Still 1 open question:
- Keep `action_create(action_str)` as is or change it to `action_create(action_str, wl_list_ptr)` and let it add itself to the given list?

Anything else to change or improve?

johanmalm:
> Still 1 open question:
> 
>     * Keep `action_create(action_str)` as is or change it to `action_create(action_str, wl_list_ptr)` and let it add itself to the given list?
>       Or a second public function `action_create_and_append(action_str, wl_list_ptr)` which wraps `action_create(action_str)`

My preference is to leave as is (i.e. separate `action_create()` and `wl_list_insert()`) as it's more consistent with the rest of the code base.

> Anything else to change or improve?
I'm not mad keen on adding variable types to names, for example `action_str`.

Great work. Looks good.

Consolatis:
> I'm not mad keen on adding variable types to names, for example `action_str`.

Changed to `action_name` + removed logging + squashed + rebased

johanmalm:
Excellent 👍

Issue #122, opened 2021-11-27
[Feature Request] Mouse window snapping


skyne98:
Hey, there! 

It would be nice to have the possibility to use your mouse to snap windows (or make them full-screen by snapping). I would suggest it should replicate the behavior of the new Windows 11, where you can snap windows not only sides, but also corners, which helps on big screens.

In my head, I see two alternative behaviors working together:
1. When using your mouse to drag the window (either using the decoration or `Alt+LMB`), moving it close to one of the edges of some display will `suggest` releasing the button to snap it there (the same way as `Meta+Arrow Key` would).
2. When doing the same actions as in behavior number 1, but holding `Shift` in the process, the compositor will now be tracking your cursor being in some regions of the screen, not edges. Something like this ![image](https://www.howtogeek.com/wp-content/uploads/2021/06/Windows-11-Snap-Layouts.png?height=200p&trim=2,2,2,2) but not configurable for now (to make it simpler?).

As an additional point, I think it is important to allow users to drag the windows which were double-clicked back into their original forms (and sizes), making them not full-screen again.

I would be glad to help somehow, or get involved, but my experience with C is very limited, I am mostly a Rust/.NET/JS/TS/Dart developer. If there is any way I can help, please tell me!

Consolatis:
By "`suggest` releasing the button" I guess you mean something like a mostly transparent overlay covering the area the window would take if it would be released.

Both `1.` and `2.` should be doable.
@johanmalm is something like `SnapToRegion` within the scope of labwc, either now or later?

As `SnapToEdge` is already implemented I might give it a try.
Sounds like the view should be placed in `interactive_end()` if `view->server->input_mode` == `LAB_INPUT_STATE_MOVE` && view is in X pixel distance of one of the output's edges.
While the overlay should likely be drawn in `process_cursor_move()` using the same condition.
I have zero knowledge about graphic stuff so no idea how to render that overlay but one after the other.

Consolatis:
So whats left is the overlay and possibly regions.

I don't know how one would even configure regions though
- based on raw (output local) pixel coordinates (that would require another config section to define outputs in the first place)?
- just setting a predefined layout like `quarter`, `twobythree`, `threebytwo` which then gets calculated based on current output area?
- what about those mixed region tilings like one half screen on the left and 3 1/3 half-screens on the right
- or how would one express a 2/3 screen on the left and a 1/3 screen on the right

skyne98:
[...]

PR #197, opened 2022-01-05
Unmaximize on Move

Consolatis:
By @skyne98 in #122:
> As an additional point, I think it is important to allow users to drag the windows which were double-clicked back into their original forms (and sizes), making them not full-screen again.

I gave a try at implementing that feature.

The PR uses the fractional position of where the cursor is located related to size and position of the maximized window and scales it down to the unmaximized state.
The window will then be positioned on both axis based on the calculation and the current cursor position so it will not suddenly jump around and (at least for me) it feels natural to drag, both when moving by grabbing the title bar and when moving with `A-BTN1` or a generic `Move` keybind.

I had used `struct border` before to determine the required offsets but then it would need to consider exclusive zones and possibly other offsets in the future. That isn't required at all though because the window is already maximized and thus all those calculations had already be done and we can just use `view->x` / `view->y` to determine the correct offsets.

~~Warning: only tested using a single screen.~~

### DIFF
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -1,16 +1,49 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include "labwc.h"
 
+static int
+max_move_scale(double pos_cursor, double pos_current,
+	double size_current, double size_orig)
+{
+	double anchor_frac = (pos_cursor - pos_current) / size_current;
+	int pos_new = pos_cursor - (size_orig * anchor_frac);
+	if (pos_new < pos_current) {
+		/* Clamp by using the old offsets of the maximized window */
+		pos_new = pos_current;
+	}
+	return pos_new;
+}
+
 void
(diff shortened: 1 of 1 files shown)


johanmalm:
I like this in principle - including the fractional approach.
I suggest though, that we implement the "Drag" mouse-event before pulling this PR, otherwise double-clicking on the titlebar (to ToggleMaximize) is a bit weird.
I think "Drag" is the only mouse-event that we've not yet implemented and I understand now more than ever why it's needed.
Implementing "Drag" would also avoid showing a "grab" cursor when doubling clicking the title bar which doesn't feel right.

ARDiDo:
I'm interested in implementing "Drag". I can get started on it soon-- that is, if no one else has already started.

johanmalm:
> I'm interested in implementing "Drag". I can get started on it soon-- that is, if no one else has already started.

That’d be great. Not aware that anyone is working on it.

PR #198, opened 2022-01-05
Allow mouse movements to trigger SnapToEdge

Consolatis:
Basic support for window snapping by "throwing" the window onto an edge.
See #122 for context.

rc.xml.all got two new options:
```
<snapping>
  <range>1</range>
  <TopMaximize>yes</TopMaximize>
</snapping>
```
Whenever a move ends within `range` of `usable_area`, `view_snap_to_edge()`  for the given edge is called (or `view_maximize()` if we are dealing with the top edge and TopMaximize setting is set to true) .
A range of 0 ~~(or not setting the new config at all)~~ disables the feature.

~~Warning: only tested using a single screen.~~

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -30,6 +30,14 @@
   </focus>
 
   <!--
+    Set range to 0 to disable window snapping completely
+  -->
+  <snapping>
+    <range>1</range>
+    <topMaximize>yes</topMaximize>
+  </snapping>
+
+  <!--
     Keybind actions are specified in labwc-actions(5)
     The following keybind modifiers are supported:
       W - window/super/logo
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -44,6 +44,10 @@ struct rcxml {
 
 	/* resistance */
 	int screen_edge_strength;
+
+	/* window snapping */
+	int snap_edge_range;
+	bool snap_top_maximize;
 };
 
 extern struct rcxml rc;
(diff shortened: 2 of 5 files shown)


johanmalm:
Need to have a think about this one. Will get back to you.

Consolatis:
Feel free to close my PRs if you don't like the given feature. I just enjoy writing code and figuring out code bases.

Personally I don't use this mouse snapping stuff in Mate at all but I know people who can't live without it.
What I sometimes like though is having windows snap to other windows while dragging but I didn't investigate that yet.
And thus didn't ask if it would be in scope for labwc either. It would basically be something like the edge resistance but for windows instead of output edges.

Consolatis:
- Fixed for multiple outputs by translation of cursor coordinates into output local ones
- Added a second setting of what to do when we drag the window to the top (maximize or snap)
- Feature is now enabled by default with a range of 1 and the default action for top edge is maximize

johanmalm:
Let's pull this. It's a good feature.
We could render a box with the "to-be" geometry, but suggest we leave that to later as I'm trying to simplify `output.c` in readiness for `wlr_scene()`
Would you be happy to change `TopMaximize` to `topMaximize`. The rest of the config is lower camel case, so increases consistency.
We ought to add this to the man-page too (which I can easily do later - so not a problem :smile: - up to you)

Consolatis:
Changed to `topMaximize` + rebased.

PR #199, opened 2022-01-06
More robust rc.xml parsing

Consolatis:
- Reset `current_mousebind` when assigning a new `current_mouse_context`
- Reset `current_*_action` when creating a new `{key,mouse}bind` or `menuitem`
- Reset `current_*` in `rcxml_finish()`
- Check for `current_*` before trying to create a new action
- Check for `current_*_action` before trying to set `action->arg`
- Check for `current_mouse_context` before creating a new mousebind
- Complain to the user about most invalid config entries

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -52,19 +52,25 @@ fill_keybind(char *nodename, char *content)
 	string_truncate_at_pattern(nodename, ".keybind.keyboard");
 	if (!strcmp(nodename, "key")) {
 		current_keybind = keybind_create(content);
-	}
-	/*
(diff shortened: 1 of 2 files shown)


johanmalm:
Nice one.
We ought to do some unit-tests for this really.
I started off writing some (https://github.com/labwc/labwc/tree/adf0f70a04fa8e9d24da4f5931618eddef95ec1b/tests), but gave up because the linking got so cumbersome.
I wonder if maintaining some unit-tests in a separate repo would be a good idea.
It's always tricky to test the graphics-heavy stuff, but some of the these types of functions should be okay.

PR #200, opened 2022-01-06
Add support for drag mouse event

ARDiDo:
I wasn't too sure which of our default mousebindings should have been changed to "Drag", so I changed all of the resize and move related mousebindings.


johanmalm:
Looking at openbox’s rc.xml, should Raise and Focus happen on Press, and Move/Resize happen on Drag?

I think that would feel more natural too.

https://github.com/danakj/openbox/blob/master/data/rc.xml

docs/labwc-config.5.scd
```diff
@@ -144,6 +144,7 @@ Configuration must be wrapped in a <labwc_config> root-node.
 	- Release: Releasing the specified button in the context.
 	- Click: Pressing and then releasing inside of the the context.
 	- DoubleClick: Two presses within the doubleClickTime.
+	- Drag: Pressing the button within the contex, then moving the cursor
```
johanmalm, line 147: Missing t in context

ARDiDo:
That should do it. What do you think @johanmalm?

johanmalm:
Looking good. Let's go for it :smile:

johanmalm:
I've pulled the "Unmaximize on move" PR too.
Running with defaults (without config file), there if something odd going on... when double-clicking the title-bar it maximizes, but then maximizes as soon as your move the mouse.
When double-clicking the left/right/top/bottom edges, it's the same (get stuck on resizing).
It's getting a bit late here, so will leave it.
Appreciate anyone investigating.

Consolatis:
Had a quick test of #202 using `build/labwc -c /dev/null` and it seems to fix it.

PR #202, opened 2022-01-08
Fix stuck mouse drags when maximizing a window

jlindgren90:
Fix two separate bugs that predate, but were exposed by, ae43d4b9d1c8fb1a758fa9f212f629a72e6d7c27 ("Unmaximize on Move"):

1. Double-clicking on the titlebar and then moving the mouse (while still held after the second press) would trigger both a DOUBLECLICK and then also a DRAG binding, resulting in the window maximizing and then immediately unmaximizing.
2. Clicking on the maximize button would trigger a CLICK binding but also leave a DRAG binding active, resulting in the window unmaximizing and moving as soon as the mouse moved, even though the mouse button was no longer held.

More details in the commit messages.


johanmalm:
Awesome.

Issue #194, opened 2022-01-03
turning off outputs with `swayidle` crashes wayland


01micko:
I tried the latest commit at the time f28319b and when my screens blanked (using `wlr-randr`) if a `gtk` app is running such as firefox, geany or pcmanfm (or anything else gtk) wayland crashes - and I lose all input and have to hard poweroff.

I'm running dual outputs, also pure wayland.

Here's the last few lines of my log:
``` 
Gdk-Message: 09:21:24.593: Error reading events from display: Broken pipe
Gdk-Message: 09:21:24.593: Error reading events from display: Broken pipe
Gdk-Message: 09:21:24.593: Error reading events from display: Broken pipe
warn: wayland.c:691: disconnected from Wayland
Gdk-Message: 09:21:24.593: Error reading events from display: Broken pipe
Gdk-Message: 09:21:24.594: Error reading events from display: Broken pipe
Gdk-Message: 09:21:24.595: Error reading events from display: Broken pipe
```

The `0.4,0` tag doesn't exhibit this behaviour so I decided to chase down when it started.

|Commit|Behaviour|
|:----------|:-------------|
|f28319b|crash|
|896102d|crash|
|a837fefc|crash|
|12b0560|ok|

So after the `0.4.0` tag the next commit a837fefc shows the crash.

Here's my `~/.config/swayidle/config`
```
timeout 300 'wlr-randr --output DVI-D-1 --off --output VGA-1 --off &' resume 'wlr-randr --output DVI-D-1 --pos 0,0 --on --output VGA-1 --pos 1920,0 --on &'

Thanks for any insight.
```

Consolatis:
I can't test this as I am still using a Mate X11 session as my main desktop and my daily driver is actually a Odroid C4 which only has a single HDMI output. A gdb stacktrace would be useful though.

01micko:
> A gdb stacktrace would be useful though.

@Consolatis I'm certainly no programmer but have used `gdb` to debug my simple programs (see my [repositories](https://github.com/01micko?tab=repositories) ) to find memory leaks etc.

A compositor is a bigger fish; can you give an example of how to?

Consolatis:
One way would be to start normally then connect to your system from another one and attach gdb to the running labwc pid like so: `gdb -p $(pidof labwc)`.
In the gdb shell press c for continue and trigger the bug like you'd normally do.
[...]

PR #204, opened 2022-01-08
view: Handle maximized view on disabled output more simply

jlindgren90:
See discussion on issue #194 .

I tested and verified that:
- Disabling all outputs leaves maximized windows in their original location and no longer crashes `labwc`
- Disabling only one output correctly re-maximizes windows from that output onto a different output

Partially reverts commit ab07b68d7e888211e9ef5a3731f5272912d4b878.


johanmalm:
Thanks

Issue #210, opened 2022-01-13
New windows don't always receive keyboard focus


bi4k8:
If the cursor is over an existing window, new windows do not receive keyboard focus. This makes it frustrating to e.g. open a terminal and begin typing it, because it requires finding the cursor and moving it onto the newly-opened window.

This behavior is relatively recent, perhaps related to focus cleanups around maximize/etc.

johanmalm:
Linked to this?

https://github.com/labwc/labwc/commit/6651d45ed44e2af28e8f99f326cb4ef4190f0bdd

bi4k8:
Yep, reverting 6651d45 fixes this for me.

johanmalm:
I'm hoping this is not fixed. Would appreciate feedback if possible.

johanmalm:
Believe this is fixed

Issue #193, opened 2022-01-03
Applications that start in fullscreen aren't in the right location


ARDiDo:
The application appears to be scaled correctly, but there's, maybe, a distance of 100 pixels between the top of the window and the top of the screen.
I've only been able to reproduce this after https://github.com/labwc/labwc/commit/a837fefc80a4a77e4e0353e4b8a14ece258a642a, but I can't confirm that that commit contained the breaking change.

Something of note is that screenEdgeStrength activated when encountering that clearance area.

johanmalm:
Is this relevant?
https://github.com/labwc/labwc/blob/master/src/xdg.c#L334-L338
https://github.com/labwc/labwc/blob/master/src/xwayland.c#L271

ARDiDo:
There might be a connection. But sadly, just commenting out those lines didn't help : )
I'll try to look into this a bit more.

johanmalm:
Sorry, I meant to do the same for full-screen. I.e. don’t let the application start in full screen.

The alternative is to set the view padding/margin (for csd/ssd respectively) on first unmax/unfullscreen state… which seems messy.

ARDiDo:
Yeah, I get that.
I might try to tackle that problem in the future, but I'll just force the view to be unfullscreened for now.

Consolatis:
Maybe this bug should be kept open and extended to starting in maximized state.
Currently (with dc203a2 applied) starting in either maximized or fullscreen state (`foot --maximized` || `foot --fullscreen`) places the window somewhere around -20 -20.

johanmalm:
I've tried with `mpv -fs ...` and also get an odd placement.
Agree, let's keep this open as a reminder to fix this.

ARDiDo:
https://github.com/labwc/labwc/commit/8e9643a85598d24d76c793099f33b7e99555c0e8 fixes this with mpv, and SDL with the Wayland backend.

I have a working fix for xwayland as well, but I want to do a little bit more testing on it first.

johanmalm:
Does it work for both server and client side decoration?

johanmalm:
And if an xdg application is started in maximized mode, is the right padding/margin set when unmaximized?

ARDiDo:
For your first question, I can't think of any xdg clients that use client side decorations off the top of my head, so I am unable to answer that.

[...]

PR #217, opened 2022-01-21
output: Add option to preview the contents of the current cycle_view

liam-middlebrook:
Add the 'cycleViewPreview.core' option to rc.xml to enable previews of
the selected view when cycling between windows. Default this option to
be disabled to match current behavior.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -31,6 +31,10 @@ Configuration must be wrapped in a <labwc_config> root-node.
 *<core><adaptiveSync>* [yes|no]
 	Enable adaptive sync. Default is no.
 
+*<core><cycleViewPreview>* [yes|no]
+	Preview the contents of the selected window when cycling between windows.
+	Default is no.
+
 # FOCUS
 
 *<focus><followMouse>* [yes|no]
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -11,6 +11,7 @@
     <decoration>server</decoration>
     <gap>0</gap>
     <adaptiveSync>no</adaptiveSync>
+    <cycleViewPreview>no</cycleViewPreview>
   </core>
 
   <!--
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -48,6 +48,9 @@ struct rcxml {
 	/* window snapping */
 	int snap_edge_range;
 	bool snap_top_maximize;
+
+	/* cycle view (alt+tab) */
+	bool cycle_preview_contents;
 };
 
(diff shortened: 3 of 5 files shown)


johanmalm:
I like the look of this. @ARDiDo @Consolatis et al. We should probably do this by default. What do you think?

Consolatis:
I don't know yet if I personally like it. I'd have to use it for some time to make up my mind, especially in combination with the high contrast double borders. But as it can be turned off I am totally fine with the feature as such, regardless of being turned on or off by default.

PR misses rendering server side decoration though, so there is a transparent hole between the top border and the client window which looks pretty weird.

Joshua-Ashton:
I tend to have a lot, ~30 windows open at once. An option to show only that window would be appreciated also.

ARDiDo:
I like it. I think it makes cycling views belonging to the same program much easier. This should probably be enabled by default.

johanmalm:
Right - I've done it :smile:
Guess we just need to agree what the default should be before the 0.5.0 tag.

johanmalm:
Having played with it, I think this is definitely much better. I vote for enabled by default.

@Joshua-Ashton I don't follow your thought. Do you mean to hide all other views and only render `server->cycle_view` during `alt-tab`?

Consolatis:
@johanmalm 
> Guess we just need to agree what the default should be before the 0.5.0 tag.

I strongly suggest either reverting or fixing 6651d45ed44e2af28e8f99f326cb4ef4190f0bdd before tagging a new release. [Context 1](https://github.com/labwc/labwc/issues/193#issuecomment-1019163749), [Context 2](https://github.com/labwc/labwc/issues/210).

Maybe you could open an issue whenever you plan on tagging a release.
That issue could then contain a list of release-blocking issues that should be fixed before doing the actual release.

johanmalm:
> @johanmalm
> 
> > Guess we just need to agree what the default should be before the 0.5.0 tag.
> 
> I strongly suggest either reverting or fixing [6651d45](https://github.com/labwc/labwc/commit/6651d45ed44e2af28e8f99f326cb4ef4190f0bdd) before tagging a new release. [Context 1](https://github.com/labwc/labwc/issues/193#issuecomment-1019163749), [Context 2](https://github.com/labwc/labwc/issues/210).
> 
> Maybe you could open an issue whenever you plan on tagging a release. That issue could then contain a list of release-blocking issues that should be fixed before doing the actual release.

I've reverted 6651d45 https://github.com/labwc/labwc/commit/2652522a70f70d156e76c5ed7070c7f615937a8a

Will open issue to help manage approach to next release. Good idea.

PR #218, opened 2022-01-23
Add a window menu

Consolatis:
I added a very basic window menu which is activated (by default) on `A-Space` or `Right` mouse button on `TitleBar`.
Before I go further and add a window menu icon on the left of the titlebar (and figure out how the themes work) I'd like to get some feedback for the feature and approach.

This is what it currently looks like:
window-menu-2

It reuses all of the existing root-menu implementation and just adds a new menu on the same level as the `root-menu` node. Some of the rendering paths seems a bit hacky to me, they set / react to `menu->visisble`.

It also isn't context aware, so `Maximize` will always be `Maximize` even when the view is currently maximized. When activated it will then unmaximize but the text doesn't match what will be happening if clicked. I am not sure how that could be changed within the existing menu framework.

So before I go further with this, is the feature in scope of labwc and is the implementation somewhat reasonable?


ARDiDo:
1. This is in scope since it's on our TODO list. https://github.com/labwc/labwc-acceptance-criteria/blob/main/TODO.md
2. I've only checked it out for a couple of minutes, and done some basic testing. I don't have any immediate concerns, and using `menu->visible` seems fine to me. You might want to wait for @johanmalm's opinion, since he's often a better judge of these things.

I'm against adding any context awareness, since that would require some amount of control flow, among other complexities, to be added to our menu parsing and rendering. That just seems a bit outside of the scope. Maybe change `Maximize` to `(Un)Maximize`, to reduce confusion when the user sees it with a maximized window.

johanmalm:
This is really good. Definitely on the TODO list, as @ARDiDo pointed out.

I think I'm right in saying that openbox doesn't not allow configuration of the window-menu (i.e. it's all hard-coded). I had in mind doing exactly what you've done though - it just makes so much sense.

It would be nice to position the window-menu at the top-left corner of the view.
In action.c, we could pass `struct view *view` to `show_menu()` and then position the menu using `menu_move()`. I've not tried it, but think that would work.

Regarding your thoughts on `menu->visible`, I agree that it becomes a bit too if-statement heavy, but I guess we're unlikely to have more menus than these two (??) so it's probably not that big a deal. A cleaner way could be to set a `server->current_menu` a bit like we do with `server-cycle_view`.

https://github.com/labwc/labwc/blob/f0a3f365e52741001b29251e0f20244a7275a266/include/labwc.h#L167

@ARDiDo - thanks for the vote of confidnece in being a good judge on this. I don't think it's true though. I vote for having a healthy mix of opinion between a few of us :smile: Seems to work best.

Regarding the context aware submenus, maybe that's why openbox hard-code the window menu (just guessing, haven't had a look). If we pass `view` to `show_menu()` as discussed, it probably wouldn't be too hard to do though.

If you let me get in and do the menu `separator` implementation, that will add functionality for "un-selectable" items, and then we can ponder that one further.
@ARDiDo, sorry I didn't respond to your message on the `HiDPI` issue - are you waiting to merge some scale fixes for the menu implementation?

ARDiDo:
@johanmalm, not really. I just wanted to give a small update.

Consolatis:
> It would be nice to position the window-menu at the top-left corner of the view.

See last patch. I don't know if I prefer that though, right clicking somewhere on the title and having the menu open on the top-left corner feels a bit weird.

At least in Mate when you right click on the title the menu opens where the cursor is. If you use `A-Space` it opens on the top-left corner, same for clicking the window-menu icon on the top-left corner. I don't know if we can somehow replicate that behavior (if that is the agreed on one). 

We'd need to test the cursor position for
- matching window-menu-icon (yet to be implemented): top-left
- matching titlebar: cursor position
- else: top-left

johanmalm:
We ought to call it client-menu actually. See ShowMenu action section:

http://openbox.org/wiki/Help:Actions#ShowMenu

Agree with the logic for position.

Openbox spec accepts a position element when x and y canbe specified. The wiki documentation doesn’t actually say, but I assume that not specifying a position results in launching at pointer (as is normal for root-menu).
It describes that the x and y vales are relative to the top left corner.

Consolatis:
This is what it currently looks like:
window-menu-button

Added the window-menu button by simply reusing the icon from `Maximize` as I have no idea how to work with themes.
The whole new button implementation was bascially yolo'd in one try so please be very careful to check if I missed something.

> Agree with the logic for position.

Implemented like described, `Left`+`Right` on the window menu icon opens top-left, `Right` on the titlebar opens where the cursor is and everything else opens top-left.

> We ought to call it client-menu actually.

We can certainly change that before pulling the change, I think window-menu / WindowMenu sounds more familiar but I don't mind changing that.

src/ssd.c
```diff
@@ -75,14 +75,20 @@ ssd_box(struct view *view, enum ssd_part_type type)
...
+		break;
 	case LAB_SSD_PART_TITLEBAR:
 		box.x = view->x;
 		box.y = view->y - theme->title_height;
 		box.width = view->w;
 		box.height = theme->title_height;
 		break;
 	case LAB_SSD_PART_TITLE:
-		box.x = view->x + corner_square / 2;
+		box.x = view->x + theme->title_height + corner_square / 2;
```
johanmalm, line 91: Shouldn't the offset here by `corner_square` rather than `theme->title_height`. I think they'll pretty close most of the time, but I think we're allowing a width of `corner_square` per icon.

src/output.c
```diff
@@ -602,6 +603,9 @@ render_deco(struct view *view, struct output *output,
 		box = ssd_visible_box(view, LAB_SSD_BUTTON_ICONIFY);
 		render_icon(output, output_damage, &box,
 			theme->xbm_iconify_active_unpressed);
+		box = ssd_visible_box(view, LAB_SSD_BUTTON_WINDOW_MENU);
+		render_icon(output, output_damage, &box,
+			theme->xbm_maximize_active_unpressed);
```
johanmalm, line 608: I'm not sure that the maximized icon looks quite right here. Aesthetically it doesn't feel quite right.
I've had a look at the openbox spec. Interestingly, they don't support a window xbm icon, so none of the themes will contain one. They just use the icon that comes with with application (and I think can be obtain via some X11 call - can't remember). Not sure if the wayland protocol supports this.
One option could be to just have an invisible icon I guess.

johanmalm:
The image looks awesome :+1: 

> > We ought to call it client-menu actually.
> 
> We can certainly change that before pulling the PR, I think window-menu / WindowMenu sounds more familiar but I don't mind changing that.

Yeah, I know what you mean. It just seems more consistent to stick with openbox-speak where we can. It's like `iconify` - I did consider just calling it `minimize`, but openbox spec is like a POSIX C standard for me :smile:

> Figuring out the current view->maximized and friends states is easy, indeed. However, changing the existing menu items to display something else seems not that easy without blowing up the existing menu framework out of proportion. At least I didn't come up with an easy solution. As the whole menu is also completely up to the user to define, it would have to have some context like
> 
> ```
> <item state="maximzed" label="Unmaximize">
>   <action name="ToggleMaximize" />
> </item>
> <item state="unmaximzed" label="Maximize">
>   <action name="ToggleMaximize" />
> </item>
> ..
> ```
> 
> So I tend to agree with @ARDiDo, maybe its just not worth the additional complexity.

Yes... you're probably right.

src/output.c
Consolatis: > I'm not sure that the maximized icon looks quite right here. Aesthetically it doesn't feel quite right.

I agree, it was just the best match of the 3 Icons I had at my disposal :)
Personally I'd prefer just a dot in the middle.

> Interestingly, they don't support a window xbm icon, so none of the themes will contain one.
> They just use the icon that comes with with application [..] Not sure if the wayland protocol supports this.

I think (but may be wrong) it does not. I didn't really investigate any of the xdg_* specs yet. The closest to some kind of icon I came across so far is the `app_id` which is apparently being used with some informed guesses (but still guesses) to match to a `.desktop` file which then contains the name of an icon which may or may not be available in the current GTK (or Qt?) theme. Unless labwc decides to depend on GTK I'd rather not implement that 'specification'. I had to implement it within the panel but its ugly.

> One option could be to just have an invisible icon I guess.

Or maybe hardcode one like in `src/xbm/xbm.c`: `char close_button_normal[]`

src/ssd.c
Consolatis: The other buttons use `theme->title_height` for their `box.x` offsets and for `width` and `height`.
I was confused in the calculation of `box.width = view->w - NR_BUTTONS * corner_square` as well, because that one instead indeed uses `corner_square` for the size of a button.

I'd welcome a second look into this.

johanmalm: Sorry. Yes, theme->title_height is probably better.

src/output.c
johanmalm: There is https://gitlab.freedesktop.org/ddevault/fdicons
so, could implement without gtk, but there is a lot said for keeping it simple and avoiding all that.

Yes, hard coding in xbc.c is a good option.

I think I like a dot too.
Alternatively, hard-code a blank icon and support reading a window.xbm or something, so anyone keen can add it to their theme

src/ssd.c
Consolatis: See #221

src/output.c
Consolatis: Had a quick try at implementing this:
window-menu-rounded

I made 3 different variants which can be easily switched for testing:
- 4x4 square
- 4x4 filled square
- 4x4 filled rounded square
- Color is `window.(in)active.button.unpressed.image.color` from theme + defaults to black

Screenshot obviously uses the filled rounded square (or 'dot')
Patch should try to load a `menu.xpm` from the theme but I didn't test that.
I don't quite understand the `*_button_toggled[]` variants in `src/xbm/xbm.c`, they don't seem to be used anywhere.

johanmalm: Look good. I like the dot.
The `max_button_toggled[]` is the icon to show in the maximized state - and so on.
Need to implement that one soon.

johanmalm:
Just had a look at the code changes.
Shouldn't need menu_button_toggled[]. I think that only applies to `max`
http://openbox.org/wiki/Help:Themes#max_toggled.xbm

Consolatis:
> Just had a look at the code changes. Shouldn't need menu_button_toggled[].

Alright, I'm gonna remove it. It was just an empty placeholder anyway.

These are the items I have on my todo for this PR:
- [ ] Rename the button / mousebinding to `ClientMenu`
- [ ] Rename the menu to `client-menu`
- [ ] Remove the other test xbm bitmaps and just keep the dot one
- [ ] Rebase on master
- [ ] Squash, maybe split into two commits

Is there anything else I forgot, or something else to change?

johanmalm:
> > Just had a look at the code changes. Shouldn't need menu_button_toggled[].
> 
> Alright, I'm gonna remove it. It was just an empty placeholder anyway.
> 
> These are the items I have on my todo list for this PR:
> 
> * [ ]  Rename the button / mousebinding to `ClientMenu`
> * [ ]  Rename the menu to `client-menu`
> * [ ]  Remove the other test xbm bitmaps and just keep the dot one
> * [ ]  Rebase on master
> * [ ]  Squash, maybe split into two commits
> 
> Is there anything else I forgot, or something else to change?

Looks good, except should the action be ShowMenu?

Consolatis:
> Looks good, except should the action be ShowMenu?

Yes, the action is `ShowMenu` but the new trigger button is currently called `WindowMenu`:
```
<context name="WindowMenu">
    <action name="ShowMenu">
      <menu>window-menu</menu>
    </action>
  </mousebind>
</context>
```
If I got it right you'd prefer to have it called `ClientMenu` instead, or were you only referring to the term `window-menu` in `menu.xml` and as argument to `ShowMenu`?

johanmalm:
> > Looks good, except should the action be ShowMenu?
> 
> Yes, the action is `ShowMenu` but the new trigger button is currently called `WindowMenu`:

Got it.

> 
> ```
> <context name="WindowMenu">
>   <mousebind button="Left" action="Click">
>     <action name="ShowMenu">
>       <menu>window-menu</menu>
>     </action>
>   </mousebind>
> </context>
> ```
> 
> If I got it right you'd prefer to have it called `ClientMenu` instead, or were you only referring to the term `window-menu` in `menu.xml` and as argument to `ShowMenu`?

Yes, I think the `<action><menu>` value should be `client-menu`

The context is a more tricky one :smile: I hadn't thought of that one.
I suppose the choice most consistent with Openbox spec would be `Icon`
http://openbox.org/wiki/Help:Bindings#Context
But I'm easy on this one, because it's not really the application icon, it's something new we've created.
I'm happy with `WindowMenu`

Consolatis:
All items from todo list done + split into two commits. One for adding the context menu itself and one for the new button.
I think this PR is ready for a final review.

PR #219, opened 2022-01-23
Centralize freeing of action lists

Consolatis:
Removes some code duplication and makes it easier to extend `struct action` by adding further arguments.


ARDiDo:
That cleans everything up nicely. Good work!

PR #220, opened 2022-01-23
config/rcxml: Prevent overwriting action argument

Consolatis:
\+ Prevent leaking memory in those cases.


ARDiDo:
Good work. Thanks for finding + fixing this.

PR #221, opened 2022-01-25
src/ssd.c: Clarify button dimension calculations

Consolatis:
\+ change width calculation of `LAB_SSD_PART_TITLE` to use `button_width` instead of `corner_square`
I could change `int16_t` into a usual `int` if you prefer.


johanmalm:
Thanks. Much clearer.
int16_t is good. Better than int.

PR #223, opened 2022-01-29
Reduce damage on cursor_move

Consolatis:
Part two of #222:
- Detect button enter/leave events and only damage the current output if happened
- Use proper `enum ssd_part_type` for `desktop_surface_and_view_at()`
- Move `is_button()` from `src/output.c` into `src/ssd.c` and make it public

This could be further optimized by damaging only the old/new button regions on button enter/leave.

By the way, thanks for this comment:
`/* Required for iconify/maximize/close button mouse-over deco */`
This made it really easy to figure out what was going on and - way more important - why it has been going on.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -43,6 +43,7 @@
 #include <xkbcommon/xkbcommon.h>
 #include "config/keybind.h"
 #include "config/rcxml.h"
+#include "ssd.h"
 
 #define XCURSOR_DEFAULT "left_ptr"
 #define XCURSOR_SIZE 24
@@ -448,7 +449,7 @@ bool isfocusable(struct view *view);
  */
 struct view *desktop_surface_and_view_at(struct server *server, double lx,
 	double ly, struct wlr_surface **surface, double *sx, double *sy,
-	int *view_area);
+	enum ssd_part_type *view_area);
 
 struct view *desktop_view_at_cursor(struct server *server);
 
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -69,5 +69,6 @@ void ssd_create(struct view *view);
 void ssd_destroy(struct view *view);
 void ssd_update_geometry(struct view *view, bool force);
 bool ssd_part_contains(enum ssd_part_type whole, enum ssd_part_type candidate);
+bool ssd_is_button(enum ssd_part_type type);
 
 #endif /* __LABWC_SSD_H */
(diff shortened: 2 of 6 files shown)


johanmalm:
Thanks. A couple of further thoughts:
Should we rename `view_area` to `ssd_type`
Should we rename `enum sdd_part_type` to `sdd_type`
Just semantics, I know, but particularly `view_area` seems a bit odd - a hangover from long ago I think.

Consolatis:
> Should we rename view_area to ssd_type

I have no strong opinion about that. Now that it is defined using the enum, reading the code should make it clear what it contains. Took me a while to understand that part which is why I changed the signature of `desktop_surface_and_view_at()`.

> Should we rename enum sdd_part_type to sdd_type

No strong opinion either:
- `ssd_type` is  is shorter
- `ssd_part_type` is a more correct description (as it is not a type of server side decoration but a type of a part of the server side decoration)

If you prefer to rename one or both I could write a patch if you don't currently have the time to do so.
That is for other stuff as well by the way, I you want me to tackle something freel free to point it out.

Totally offtopic: I'm idleing in #labwc @libera.

johanmalm:
I think you're right - it's probably okay as it is :)

Consolatis:
> Hi there. Good idea setting this up. Maybe we ought to make it the official channel.

Registered the channel #labwc thus I am the official "Founder" for now.
If you want to make the channel official it appears this is the way to do it: https://libera.chat/chanreg/#project-registration

Obviously I am perfectly happy to give the Founder status of the channel to whoever you prefer (they must have an nickserv registered account) and become a normal member of the channel.

johanmalm:
> > Hi there. Good idea setting this up. Maybe we ought to make it the official channel.
> 
> Registered the channel #labwc thus I am the official "Founder" for now. If you want to make the channel official it appears this is the way to do it: https://libera.chat/chanreg/#project-registration
> 
> Obviously I am perfectly happy to give the Founder status of the channel to whoever you prefer (they must have an nickserv registered account) and become a normal member of the channel. I think there can be multiple people with Founder role at once but I am not sure about that.

Sorry for slow response. The IRC channel is a great idea.
I'm personally pretty rubbish at staying connected on IRC - mostly because I don't spend that much time on the machine and when I do, I like to turn my hand to coding. I know it's quite an insular approach and not very team friendly, but sort is the best I can offer. I think many work well on IRC though... Maybe we should add it to the README, but guess it relies on people hanging around :smile:

Consolatis:
> but guess it relies on people hanging around

Kinda. Just missed you again on IRC :smile: 

Guess you would profit from using a IRC bouncer that always stays connected to IRC where you log in and get the messages you missed while being disconnected.

PR #228, opened 2022-02-04
build: Relax the version constraint for wlroots to accept patch releases

primeos:
Patch releases only contain backwards compatible changes (mainly bug
fixes) so we want to allow them. This fixes the build with the recently
released wlroots 0.15.1 and uses the same version constraints as other
projects that depend on wlroots (e.g., Sway).


johanmalm:
Thanks

PR #229, opened 2022-02-06
OSD: Scale OSD to the largest output scale

ARDiDo:
I've tested this successfully with a single monitor, and two monitors with different scales. I'm still attempting to fix menu and titlebar text with multiple outputs, but I was hoping to get a little bit of feedback before I continue working on them.

By scaling the OSD using the greatest scale for all the outputs, we end up scaling down the OSD texture for outputs with any other scales. Downscaling a texture is significantly less noticeable than upscaling (which we currently do), but it is noticeable.
With my current patches, menu text would follow a similar design, so I wanted to know if anyone had any opinions on this before I continue working on it.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -154,6 +154,7 @@ struct server {
 	struct wlr_output_manager_v1 *output_manager;
 	struct wl_listener output_manager_apply;
 	struct wlr_output_configuration_v1 *pending_output_config;
+	float greatest_scale;
 
 	struct wlr_foreign_toplevel_manager_v1 *foreign_toplevel_manager;
 
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -144,7 +144,6 @@ struct server {
 	double grab_x, grab_y;
 	struct wlr_box grab_box;
 	uint32_t resize_edges;
-	struct wlr_texture *osd;
 
 	struct wl_list outputs;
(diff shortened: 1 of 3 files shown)


johanmalm:
Hi there. Thanks for this.

I think it would be better to create separate osd textures for each scale. We create new osd textures every time we alt-tab anyway, so creating one per output-scale should be insignificant in terms of overhead.

I wonder if a data structure like the following would be a good starting point?

```
struct osd {
	struct wlr_texture *texture;
	float scale;
	struct wl_list link; /* server::osds */
};
```

In `struct server`, replace `struct wlr_texture *osd` with `struct wl_list osds`

 What are you thoughts?

johanmalm:
Regarding scaling the menu right, if you want, I can take a look at that. The pango bits need a bit of a re-write anyway and I reckon creating one texture for the whole menu might be a better approach.

ARDiDo:
That data structure looks like a good idea, thanks for thinking of it. I should have something for it soon.

Creating one texture for the menu would definitely make scaling easier. 
Thank you for your offer. I've never really handled raster graphics before, so I'll be very grateful if you can take a look at it. If you have any plans or ideas on changing `src/common/font.c`, I've got a patch for scaling font textures that you can check out.
https://github.com/ARDiDo/labwc/tree/pango_scale

Issue #211, opened 2022-01-14
wlroots 0.16.0 compatibility


johanmalm:
https://github.com/swaywm/sway/commit/d6f8820a8b7190a2ce76a0f140e1514e2c32dd50

ARDiDo:
Since our master branch is based off of wlroots 0.15.0, do you want to create an alternative branch to chase the wlroots master branch?

johanmalm:
Can’t make up my mind what’s best. What do you think?

I thought I’d just watch sway for any breaking changes in wlroots, log them here and deal with them later. I guess it’s the lazy, simple approach, but could avoid a headache if we do significant code re-factoring like splitting labwc.h into separate head files.
The advantage of maintaining a branch tracking 16, is that it’s easier to check if odd bugs are fixed through changes in wlroots

ARDiDo:
Yeah, I think that this lazy approach will save us some headaches that we'd get from creating a separate branch.

johanmalm:
Ha ha. Sounds like the beginning of a profound development policy... laziness before complexity :smiley:

droc12345:
I've made a couple of changes, one to meson files, the other to add includes so latest labwc git compiles against latest wlroots git.

https://github.com/droc12345/labwc/commit/66251c6335047c8740c65304dceb7453fbd24de3
This patch adds a system_wlroots option, to allow for disabling using system wlroots and instead using the latest git version.
It also changes the meson.build file to incorporate the new option and pretties up some of the file.
It also changes the pointer in wlroots.wrap to point to HEAD (always latest git)

https://github.com/droc12345/labwc/commit/240340af5cfa92d5b40467612aacd63449ebd582
This patch replaces wlr_surface.h with wlr_{sub}compositor.h (wlr_surface.h is being depreciated) in layers.h
and adds layers.h to subsurface.c and xdg.c (to get them to compile - gcc 11.2)

johanmalm:
Covered by PR #238

PR #238, opened 2022-02-12
wip: Update for wlroots-git (0.16.0)

jlindgren90:
Various small updates needed to build labwc with the `master` branch of wlroots.  I plan on continuing to update this branch if/when wlroots makes additional breaking changes.

This isn't intended to be merged until wlroots 0.16.0 is released, but it may be useful to anyone else wanting to use labwc with a bleeding-edge version of wlroots.

**edit 2022/3/20:**
This is no longer needed for the `master` branch of labwc, but might still be useful if anyone is trying to build `v0.5` with a newer wlroots version.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -33,6 +33,7 @@
 #include <wlr/types/wlr_pointer_gestures_v1.h>
 #include <wlr/types/wlr_seat.h>
 #include <wlr/types/wlr_server_decoration.h>
+#include <wlr/types/wlr_subcompositor.h>
 #include <wlr/types/wlr_xcursor_manager.h>
 #include <wlr/types/wlr_xdg_decoration_v1.h>
 #include <wlr/types/wlr_xdg_shell.h>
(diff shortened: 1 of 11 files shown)


jlindgren90:
See also #211

johanmalm:
Thanks

droc12345:
src/output.c instead of removing the if (box) - line 1203 and following 
it could be changed to if (!wlr_box_empty(&box)) instead 
and this keeps the check for an empty or bad parm box and associated log message.

johanmalm:
Thank you both. 
I've incorporated in the scene-graph branch (https://github.com/johanmalm/labwc/tree/scene) because we needed some new functions added after 0.15.1
Also added wlr_subcompositor_create() in server.c

jlindgren90:
> See [d844f67](https://github.com/labwc/labwc/commit/d844f6798b2de509bef80c9f8148b5a975e98f6a) You might be able to cherry-pick

I was able to, thanks :)

acoolstraw:
cool!

acoolstraw:
why does the CI/build fail?

jlindgren90:
I've been using only `master` lately and haven't updated this branch in quite a while, so I'm going to go ahead and close the merge request. I'll leave the `wlroots-git` branch up for now in case anyone wants to use it as a starting point for a new effort.

Issue #225, opened 2022-01-29
Qutebrowser crashes on touchpad scroll


ahmadraniri1994:
I'm getting  a weird behaviour when I'm opening qutebrowser on LABWC, it opens ordinarily for about 2 minutes and start hang and crash / force closed. I never get this behaviour on other wayland compositors (sway, hikari). I have no clue if it's related to LABWC specifically. Thanks.
OS : Archlinux
Labwc : Latest master (two recent update still same).

johanmalm:
I don't normally use `qutebrowser`, but have just installed/ran it for a while. Seems to work fine here, so I'm unable to reproduce and fault find. I'm on Arch also.

It is just `qutebrowser` that hangs or the whole `labwc` compositor?

If you run `qutebrowser` from the terminal, do you get any error messages when it hangs?
If not, could you try running it with debug log-level (`qutebrowser -l debug`)

ahmadraniri1994:
Well, I'll check it later.

ahmadraniri1994:
I'm sorry for late reply, turns out it's happening on my debian setup too.
Qutebrowser log : https://pastebin.com/Fa3kkwzV

ahmadraniri1994:
For some reason I reinstall my archlinux, I did reinstall labwc too (labwc-git from aur), but the issue still exixt. Dunno why.

ahmadraniri1994:
These are some stuffs I got when I run qutebrowser from commandline : 
> [ahmad@LIDGNULINUX ~]$ qutebrowser 
> 07:35:59 INFO: Run :adblock-update to get adblock lists.
> 07:36:00 WARNING: Wayland does not support QWindow::requestActivate()
> Segmentation fault (core dumped)

ahmadraniri1994:
Turns out every qt apps crash on labwc, soon after the apps are opened, the apps will crash. Will investigate it further.

ahmadraniri1994:
Set QT_QPA_PLATFORM=xcb seems fix the issue, but it means every QT apps use xwayland now.

ahmadraniri1994:
I give up, hope the next release will fix the issue (oops still exist when I compile the last master).

Consolatis:
> These are some stuffs I got when I run qutebrowser from commandline :
> 
> > [ahmad@LIDGNULINUX ~]$ qutebrowser
> > 07:35:59 INFO: Run :adblock-update to get adblock lists.
[...]

PR #256, opened 2022-02-26
cursor: add cursor_update_focus()

### LINKED ISSUE #162: Alt-tab does not change which window gets scroll events

Consolatis:
...and call it from desktop_move_to_front() in order force an enter event
on the surface below the cursor when cycling views.

Fixes #162 and #225
Inspired by PR #164 - just restructured it a bit.

Third time is a charm. See https://github.com/labwc/labwc/issues/162#issuecomment-1051570900 for context.

If the PR is accepted I'll cherry-pick it for scene-graph as well (applies cleanly).

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -467,6 +467,14 @@ void cursor_rebase(struct seat *seat, uint32_t time_msec);
  */
 void cursor_set(struct seat *seat, const char *cursor_name);
 
+/**
+ * cursor_update_focus - update cursor focus
+ * @server - server
+ * Use it to force an update of the cursor icon and to send an enter event
+ * to the surface below the cursor.
+ */
+void cursor_update_focus(struct server *server);
+
 void cursor_init(struct seat *seat);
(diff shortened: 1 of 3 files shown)


johanmalm:
Does it handle the reason for previous reversion?
See https://github.com/labwc/labwc/commit/2652522a70f70d156e76c5ed7070c7f615937a8a

johanmalm:
Nice work.
Works well and doesn't re-create issue #210

PR #258, opened 2022-02-27
Avoid breaking double-click selection in foot on v0.5 branch

bi4k8:
This regressed in ce38d2d; basically, we shouldn't clear and set focus again if the same surface is focused before and after cursor rebase.

`WAYLAND_DEBUG=1` output of double-clicking on foot looks like the below:

master:
```
[41.928] wl_pointer@19.button(7086, 77762627, 272, 1)
[41.947] wl_pointer@19.frame()
```

v0.5:
```
[41.856] wl_pointer@19.leave(7084, wl_surface@3)
[41.902] wl_pointer@19.frame()
[41.908] wl_pointer@19.enter(7085, wl_surface@3, 48.48437500, 11.87500000)
[41.924] wl_pointer@19.frame()
[41.928] wl_pointer@19.button(7086, 77762627, 272, 1)
[41.947] wl_pointer@19.frame()
```

And looks just like master again after this fix.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -23,6 +23,16 @@ cursor_rebase(struct seat *seat, uint32_t time_msec)
 		seat->cursor->y, &surface, &sx, &sy, &view_area);
 
 	if (surface) {
+		struct view *focused_view = desktop_focused_view(seat->server);
+		struct wlr_surface* focused_surface = focused_view
+			? focused_view->surface
+			: NULL;
+
+		/* do not notify unless rebasing changes which surface has focus */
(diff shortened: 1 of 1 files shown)


Consolatis:
Nice catch.
Sorry for not detecting this earlier, apparently my local rc.xml does something different with Focus/Raise.
Should really have tested this with `-c /dev/null` as well.

```
desktop_surface_and_view_at([..]);
if (surface == seat->seat->pointer_state.focused_surface) {
  return;
}
```
should be enough here.

Issue #259, opened 2022-02-27
Cannot use Alt+Left (mouse and touchpad) to move window (v0.5 branch).


ahmadraniri1994:
I'm happy to try v0.5 branch since the scroll bug on qt apps are fixed, but now I can't use Alt+Left to move and drag windows (both touchpad and mouse). Switch to master suddenly the problem is gone. I don't know if this behaviour is already reported before. Thanks.

01micko:
+1

Can still drag by titlebar though.

ahmadraniri1994:
Yeah. So, is it confirmed ? Better switch v0.5 again then. Thanks.

01micko:
I wouldn't. You should still be able to drag windows and resize. I'm sure there'll be a simple fix.

Consolatis:
> Yeah. So, is it confirmed ?

Yes it is confirmed by code as well. Fallout of #241.

Currently working on a fix that hopefully doesn't break anything else.
Would be great if you could test drag and drop of #260

Consolatis:
#260 has been merged into the v0.5 branch so just updating it should solve the issue here (+ double click in foot).
Further testing of v0.5 branch would be greatly appreciated because we are thinking about doing a minor 0.5.1 release soon.

ahmadraniri1994:
Confirmed, drag with Alt+left works again. Thanks.

01micko:
Fixed with no popup regressions, tested firefox and vivaldi. :smile:

01micko:
![20220228_11h46m54s_grim](https://user-images.githubusercontent.com/1019119/155911368-4e61b73a-f182-4734-8349-798f3bb05cfa.png)

Consolatis:
@01micko awesome! thank you :)

PR #260, opened 2022-02-27
Restore Drag mouse bindings and proper double click

Consolatis:
Fixes #258 and #259

Reported-by: @ahmadraniri1994
Reported-by: @01micko
Suggested-by: @bi4k8

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -13,7 +13,7 @@
 #include "config/mousebind.h"
 
 void
-cursor_rebase(struct seat *seat, uint32_t time_msec)
+cursor_rebase(struct seat *seat, uint32_t time_msec, bool force)
 {
 	double sx, sy;
 	struct wlr_surface *surface;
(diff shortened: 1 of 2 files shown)


src/cursor.c
```diff
@@ -296,6 +289,17 @@ process_cursor_motion(struct server *server, uint32_t time)
 		last_button_hover = LAB_SSD_NONE;
 	}
 
+	if (server->seat.active_view && !server->seat.drag_icon) {
```
johanmalm, line 292: Could the logic here just be:

```
if (server->seat.active_view && !server->seat.drag_icon) {
  //do stuff
} else if (surface &&
	    !input_inhibit_blocks_surface(&server->seat, surface->resource)) {
  // more stuff
} else {
  wlr_seat_pointer_clear_focus(wlr_seat);  
}
```

Consolatis: you are right. added a new fixup commit to simplify the logic

PR #266, opened 2022-03-08
Support wlr-output-power-management protocol

bi4k8:
This allows using tools like `https://github.com/lilydjwg/dpms-off/` and the `output-power-management` example from wlroots. In conjunction with `swayidle`, the latter can make idle blanking work the way users generally expect.


johanmalm:
Looks good. Thanks.

PR #268, opened 2022-03-09
README.md: add `wlopm` to "6. Integrate" (recommended programs)

01micko:
- made it first in the array of 'output managers' because since the
`scene_graph` merge it's the only way I have found to reliably
turn off/on outputs.
- it supports "wlr-output-power-management-unstable-v1"
- as does labwc since c23397f
- works with swayidle/swaylock

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Suggested apps to use with labwc:
 - Background image: [swaybg]
 - Panel: [waybar], [yambar], [lavalauncher], [sfwbar]
 - Launchers: [bemenu], [fuzzel], [wofi]
-- Output managers: [kanshi], [wlr-randr]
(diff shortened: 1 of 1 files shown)


johanmalm:
Brilliant. Thanks @01micko

johanmalm:
@01micko 
We ought to update the comment and example in docs/autostart
https://github.com/labwc/labwc/blob/4ee9f351eaba445158aac9aecb638f96e82656af/docs/autostart#L32 with wlopm
Would you be happy to propose something, or do you want me to?

01micko:
> Would you be happy to propose something, or do you want me to?

Maybe; let me think about it and do some testing in different circumstances.

PR #282, opened 2022-03-20
xwayland: Honor size increments from WM_SIZE_HINTS

jlindgren90:
Makes labwc aware of the size increments specified in WM_SIZE_HINTS.  Typically, X11 terminal emulators will use WM_SIZE_HINTS to make sure that the terminal is resized to a width/height even divisible by the cell (character) size.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -170,9 +170,26 @@ handle_set_class(struct wl_listener *listener, void *data)
 	view_update_app_id(view);
 }
 
+static int
+round_to_increment(int val, int base, int inc) {
+	if (base < 0 || inc <= 0)
+		return val;
+	return base + (val - base + inc / 2) / inc * inc;
+}
+
 static void
 configure(struct view *view, struct wlr_box geo)
 {
+	// honor size increments from WM_SIZE_HINTS
+	struct wlr_xwayland_surface_size_hints *hints =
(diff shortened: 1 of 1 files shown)


jlindgren90:
wlroots currently has a bug preventing this from working correctly. See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3495.

johanmalm:
Thanks
Guess we ought to do this on the master branch too?

jlindgren90:
Yep, I think the patch will apply to master as well 👍

johanmalm:
Done https://github.com/labwc/labwc/commit/08c537ebd9011ffdf8b0e8209d6fdec59f8dcf06

johanmalm:
I love the support of the WM_SIZE_HINTS increments.
I note that when resizing from left and/or top edges, the opposite edges of the window jump around. Any thoughts on how we fix this?
I've not given it any thought as yet, but can do.

jlindgren90:
Hmm, probably the logic needs to be moved to `process_cursor_resize()` (cursor.c) and handled in an edge-specific way there like the minimum-size logic (WLR_EDGE_TOP/LEFT/BOTTOM/RIGHT).

jlindgren90:
@johanmalm See #291.

PR #284, opened 2022-03-20
Implement touch support

bi4k8:
This is mostly plumbing; the most interesting logic is in `touch_get_coords`.

This introduces a bit of duplication in `seat.c` where `new_touch` is almost a copy-paste of `new_pointer`. If someone would like this factored differently just let me know.

This doesn't implement anything fancy like pointer emulation, but by my testing makes touch work in GTK3 programs including drop-down menus.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -107,6 +107,11 @@ struct seat {
 	struct wl_listener keyboard_key;
 	struct wl_listener keyboard_modifiers;
 
+	struct wl_listener touch_down;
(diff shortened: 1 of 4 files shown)


johanmalm:
Thanks. Looks good.
I haven't got a touch screen, so can't test :smile:

01micko:
I have an oldish (~2015) HP tablet that I can drag out of mothballs. Maybe on the week end.

However, for said tablet to be useful we'd need virtual keyboard support re #286

PR #290, opened 2022-03-30
cursor.c: Do not segfault on missing drag icon

Consolatis:
Observed by moving tabs in Chromium 98 wayland native
(started with --ozone-platform-hint=wayland).

I am not sure how to deal with this situation properly so this commit just stops segfaulting the whole compositor for now.
Moving tabs with Firefox wayland native (MOZ_ENABLE_WAYLAND=1) works.


johanmalm:
Thanks

PR #291, opened 2022-04-03
cursor: Fix "jumping opposite edges" issue when resizing.

jlindgren90:
Commit ec68806354f2 ("xwayland: Honor size increments from
WM_SIZE_HINTS") adjusted only the window width/height according
to the size hints.  If resizing from the top or left edge of the
window, we also need to adjust the window position to keep the
bottom or right edge from jumping around.


johanmalm:
This is great. Thank you so much for doing it. It's nice to get these technical details right.
I've cherry-picked onto the master too.

PR #294, opened 2022-04-08
Add ToggleAlwaysOnTop action

Consolatis:
See http://openbox.org/wiki/Help:Actions#ToggleAlwaysOnTop

If not already set to always on top, the view will reparent onto a new `server->view_tree_always_on_top` tree or onto the usual `server->view_tree` if already set to always on top.

This also means always-on-top views will show up above fullscreen'd  applications which I think is what the user would expect.
If we were to change this we'd need yet another server view_tree for fullscreen applications.

This commit shouldn't change anything regarding layershell clients like panels or notifications.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -61,6 +61,9 @@ Actions are used in keyboard bindings.
 *<action name="ToggleMaximize">*
 	Toggle maximize state of focused window.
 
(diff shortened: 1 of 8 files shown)


johanmalm:
You make it look so easy :smile:

Consolatis:
I'd say the change to scene-graph is mostly to blame for that. bad scene-graph.

PR #298, opened 2022-04-20
xdg-shell: remove view->padding

johanmalm:
With the wlroots scene-graph API, the padding variable is no longer
needed for windows with Client Side Decoration (CSD).

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -297,15 +297,6 @@ struct view {
 	 */
 	struct border margin;
 
-	/*
-	 * padding refers to the space between the extremities of the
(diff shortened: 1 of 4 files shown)


johanmalm:
@Consolatis 
I’m not sure why we don’t need the padding variable anymore, but removing fixes move-to-edge and snap-to-edge for windows with client side deco.
Would you mind casting your eyes over it?

Consolatis:
LGTM. As we figured out in IRC, `wlroots/types/scene/xdg_shell.c` already adjust geometry with its own commit handler and X11 clients aren't using the padding anyway so it should be fine to remove.

PR #299, opened 2022-04-21
view: Refactor view destruction

Joshua-Ashton:
Consolidates all of the view destruction code for xwl + xdg into one function.

    Fixes several notable bugs along the way:
      - Fixes a crash when alt tabbing when a selected view gets destroyed.
      - Fixes the OSD not updating to reflect a view has been destroyed.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -404,6 +404,8 @@ void view_update_app_id(struct view *view);
 void view_impl_map(struct view *view);
 void view_adjust_size(struct view *view, int *w, int *h);
 
+void view_handle_destroy(struct view *view);
+
 void foreign_toplevel_handle_create(struct view *view);
 
 /*
--- a/src/view.c
+++ b/src/view.c
@@ -651,3 +651,63 @@ view_update_app_id(struct view *view)
 	wlr_foreign_toplevel_handle_v1_set_app_id(
 		view->toplevel_handle, app_id);
 }
+
+void
+view_handle_destroy(struct view *view)
+{
+	if (view->toplevel_handle) {
+		wlr_foreign_toplevel_handle_v1_destroy(view->toplevel_handle);
+	}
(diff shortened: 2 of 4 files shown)


src/server.c
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #define _POSIX_C_SOURCE 200809L
 #include "config.h"
+#include <assert.h>
```
Consolatis, line 4: Please remove

include/labwc.h
```diff
@@ -527,6 +527,8 @@ void server_init(struct server *server);
 void server_start(struct server *server);
 void server_finish(struct server *server);
 
+void server_handle_view_destroy(struct server *server, struct view *view);
```
Consolatis, line 530: Please rename to `view_handle_destroy` and move the implementation to `src/view.c`

src/server.c
```diff
@@ -464,3 +465,30 @@ server_finish(struct server *server)
...
+			server->cycle_view, LAB_CYCLE_DIR_BACKWARD);
+
+		/* If we cycled back to ourselves, then we have no windows.
+		 * just remove it and close the osd for good. */
+		if (server->cycle_view == view) {
+			server->cycle_view = NULL;
+			osd_finish(server);
+		}
+	}
+	else if (server->cycle_view != NULL) {
```
Consolatis, line 489: Please use a single line for `} else if (..) {`

Consolatis (changes requested):
Other then the comments above this looks good to me.
Thanks :)

Consolatis:
Great work!

Not sure we want to keep all the xdg / xwayland specific things (like the xwayland listeners) in there or move them back to `xdg.c` / `xwayland.c` but we can deal with that in a follow up commit.

PR #302, opened 2022-04-22
view: Refactor view destruction some more

Consolatis:
As a follow up to #299:
- Move XDG / XWayland specific cleanups back into their own files
- Update the OSD when the destroyed view was the currently selected one
- Handle NULL return from `desktop_cycle_view()`
- Rename `view_handle_destroy()` to `view_destroy()` (it actually frees the view)

Open questions:
- Maybe I am lost here but it seems we are cleaning up some random subset of event listeners for XDG / XWayland. Shouldn't we either clean up all of them or none of them?
- @johanmalm are you fine with having the generic `view_destroy()` in `src/view.c` or should we move it somewhere else?

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -653,7 +653,7 @@ view_update_app_id(struct view *view)
 }
 
 void
-view_handle_destroy(struct view *view)
+view_destroy(struct view *view)
 {
 	if (view->toplevel_handle) {
 		wlr_foreign_toplevel_handle_v1_destroy(view->toplevel_handle);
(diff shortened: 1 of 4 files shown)


johanmalm:
Thanks. Looks really tidy.

>     * Maybe I am lost here but it seems we are cleaning up some random subset of event listeners for XDG / XWayland. Shouldn't we either clean up all of them or none of them?

I'm not sure what the right approach is, but I agree, it seems inconsistent. Happy to leave as is for the minute, but we ought to check this throughout the whole code-base.

>     * @johanmalm are you fine with having the generic `view_destroy()` in `src/view.c` or should we move it somewhere else?

The flow is odd, but I'm okay with it. If I could think of a really clean way of doing this, then I'd propose that, but I can't at the moment :smile: The idea of `view-impl.c` was to gather common hunks to be called from `xdg.c` and `xwayland.c` (such as `view_destroy()`) 
It would be nice to think of `view.c` as containing 'public' functions that could be called from anywhere (e.g. `cursor.c`, `keyboard.c`, `action.c` and `desktop.c`).

Maybe `shell-common.c` would be a better name than `view-impl.c`.

Or we could merge `xdg.c` and `xwayland.c` and just call it `shell.c` and have those common bits of code as static functions within.

Or we could create a `shell` sub-directory and have a `common.c` in there along with `xdg.c`, `xwayland.c`, etc.
We could also stick the `common.h` in there to keep it more private :smile:
The delights of C!!

From memory, sway keeps jumping between `view.c` and `xdg_shell.c`/etc. which is quite hard to follow at times.

Consolatis:
Updated the PR and simplified the branches a bit. Should be ready to pull.

> The idea of `view-impl.c` was to gather common hunks to be called from `xdg.c` and `xwayland.c` (such as `view_destroy()`)
> Maybe `shell-common.c` would be a better name than view-impl.c

Or maybe something like `view-common.c` as some kind of semi private API?
I could never really figure out what `view-impl.c` is actually supposed to be used for.

> Or we could merge `xdg.c` and `xwayland.c` and just call it `shell.c`

I think having those two separate makes it easier to build without xwayland support and also simplifies the code as we don't have to have everything within `#ifdef HAVE_XWAYLAND` branches.

> Or we could create a shell sub-directory and have a `common.c` in there along with `xdg.c`, `xwayland.c`
I think of those ideas this I like the most, not sure about the semi private API though.

src/view.c
```diff
@@ -671,43 +671,25 @@ view_handle_destroy(struct view *view)
 			view->server->cycle_view, LAB_CYCLE_DIR_BACKWARD);
 
 		/* If we cycled back to ourselves, then we have no windows.
-		 * just remove it and close the osd for good. */
-		if (view->server->cycle_view == view) {
+		 * just remove it and close the OSD for good. */
+		if (view->server->cycle_view == view || !view->server->cycle_view) {
 			view->server->cycle_view = NULL;
```
Joshua-Ashton, line 676: Why? We don't have an OSD.

Consolatis: It is within the inner branch so the view that is being destroyed was the selected one and `desktop_cycle_view()` returned `NULL` (no focusable view found).

So instead of showing an empty OSD we are closing it, even if the modifier is still pressed at this point.

johanmalm:
With thanks.

Issue #305, opened 2022-04-26
Maximised window resizes very small when dragging app across outputs.


01micko:
NOTE: app has to be started maximised. If not, then it's maximised, it reverts to the size it was before maximised.

Seems only affecting pure wayland apps.

wayland: Tried, geany, vivaldi-snapshot, firefox - all show small window

Xwayland: Tried gimp and vivaldi-stable and no problem.

Just guessing, size is about 600x200. My displays are 1920x1080.

Consolatis:
Dragging a maximized window will automatically un-maximize it (you could release it at the top edge of the target and it should automatically maximize again).

So to clarify: Issue is when starting an wayland native application in maximized mode (like `foot -m`) and then un-maximizing it (be it via dragging or double click on the title bar or just using the maximize button), it becomes a very small window?

01micko:
No the drag operation makes it small... and I just tried foot -m

![20220427_05h32m58s_grim](https://user-images.githubusercontent.com/1019119/165378227-9d66ddd9-aff1-4928-afdb-61a88829ff8f.png)

__But wait__, it's early morning here and I just realised I was testing 0.5.1! So the bug has been there for quite a while and I only just discovered it!

johanmalm:
Should be done.  Have tested, but as always... if anyone has time to double-check that is welcome.

01micko:
Fixed for me. Thanks :smile:

PR #306, opened 2022-04-26
xdg-shell: set view position before view_maximize()

johanmalm:
...on first map so that view->unmaximized_geometry is known even if an
xdg-shell application started in maximized mode.

Fixes issue #305
Reported-by: @01micko

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -213,9 +213,33 @@ view_apply_maximized_geometry(struct view *view)
 	view_move_resize(view, box);
 }
 
+#define LAB_FALLBACK_WIDTH (640)
+#define LAB_FALLBACK_HEIGHT (480)
+
+static void
+set_fallback_geometry(struct view *view)
+{
+	view->unmaximized_geometry.width = LAB_FALLBACK_WIDTH;
+	view->unmaximized_geometry.height = LAB_FALLBACK_HEIGHT;
+	view_compute_centered_position(view,
+		view->unmaximized_geometry.width,
+		view->unmaximized_geometry.height,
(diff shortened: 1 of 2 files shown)


johanmalm:
@01micko @Consolatis any chance of checking this one?

Consolatis:
When I start `foot -m` and then un-maximize it, it is still way smaller than when starting foot without `-m`. This might be related to foot as well though.

Left: started normally. Right: started via `-m` and then un-maximized.
foobar

01micko:
With each application, the bug seemed to be reproduced **_but_** I then resized the minimised app to something reasonable and at least for vivaldi-snapshot and firefox the bug is gone after a few tests however foot shows the same problem as @Consolatis ' screencap above. Maybe @dnkl  can shed some light.

dnkl:
Took a quick look, and from what I can see, foot never receives a configure event with width=0 and height=0 (which indicates that foot should choose its own size).

dnkl:
In other words, I don't see foot doing anything wrong here - it just respond to, and respects, the size hints sent from labwc

01micko:
@dnkl makes sense in the context of those 2 constants. Thanks.

johanmalm:
I've pushed another commit to set a fallback geometry if `view->unmaximized_geometry` is empty on first unmaximize.
This with works fine with GTK/Qt apps as well as `foot -m`.

I've tried adding a configure with width=0 and height=0, but can't seem to get sensible results (tried mostly with `foot -m` and `qterminal -platform wayland`. I'm sure it's doable, but I can't work it out.

PR #307, opened 2022-04-26
Feature/tab preview: restore functionality after the move to scene-graph

Consolatis:
~~I am still not 100% satisfied with the implementation.~~
First commit basically just replaces `server->cycle_view` with `server->osd_state.cycle_view` so can be skipped for review.

What I don't really like:
- `view_destroy()` starts to look like a OSD clean up routine, maybe we should have an additional `osd_view_destroyed()` handler or something like that that we can call from `view_destroy()` (similar to `ssd_destroy()`)?

What is missing: 
- [x] some kind of marker for the currently selected window (see #430)
- [x] minimized views wouldn't have any preview because their scene_node is disabled (fixed)
- [x] refactor `view_destroy()` to call some `osd_view_destroy()`
- [x] refactor to `prev_node()` as suggested by @johanmalm
- [x] respect the existing config option to enable / disable the preview
- [x] document in header file what `osd_preview_restore()` does

All in all this is one of the very few things that actually got harder by the move to the scene-graph API.

### DIFF
--- a/include/common/scene-helpers.h
+++ b/include/common/scene-helpers.h
@@ -1,7 +1,17 @@
 /* SPDX-License-Identifier: GPL-2.0-only */
 
-#include <wlr/types/wlr_scene.h>
+struct wlr_scene_node;
+struct wlr_scene_rect;
+struct wlr_scene_tree;
+struct wlr_surface;
 
 struct wlr_scene_rect *lab_wlr_scene_get_rect(struct wlr_scene_node *node);
 struct wlr_scene_tree *lab_scene_tree_from_node(struct wlr_scene_node *node);
 struct wlr_surface *lab_wlr_surface_from_node(struct wlr_scene_node *node);
+
+/**
+ * lab_get_prev_node - return previous (sibling) node
+ * @node: node to find the previous node from
+ * Return NULL if previous link is list-head which means node is bottom-most
+ */
+struct wlr_scene_node *lab_wlr_scene_get_prev_node(struct wlr_scene_node *node);
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -583,6 +583,8 @@ void osd_update(struct server *server);
 void osd_finish(struct server *server);
 /* Moves preview views back into their original stacking order and state */
 void osd_preview_restore(struct server *server);
+/* Notify OSD about a destroying view */
+void osd_on_view_destroy(struct view *view);
 
 /*
  * wlroots "input inhibitor" extension (required for swaylock) blocks
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -202,8 +202,6 @@ struct server {
 	/* Tree for built in menu */
 	struct wlr_scene_tree *menu_tree;
 
-	struct multi_rect *osd_preview_outline;
-
 	/* Workspaces */
 	struct wl_list workspaces;  /* struct workspace.link */
 	struct workspace *workspace_current;
@@ -236,6 +234,7 @@ struct server {
 		bool preview_was_enabled;
 		struct wlr_scene_node *preview_node;
 		struct wlr_scene_node *preview_anchor;
+		struct multi_rect *preview_outline;
 	} osd_state;
 
 	struct theme *theme;
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -233,6 +233,7 @@ struct server {
 	/* Set when in cycle (alt-tab) mode */
 	struct osd_state {
 		struct view *cycle_view;
+		bool preview_was_enabled;
 		struct wlr_scene_node *preview_node;
 		struct wlr_scene_node *preview_anchor;
 	} osd_state;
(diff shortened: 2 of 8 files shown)


johanmalm:
Hey, I really like the look of this. Great stuff.

It works really well - event the view_destroy() even handles disappearing windows whilst alt-tabbinb :smile:

In way of simplification - what do you think of this?

```
diff --git a/src/osd.c b/src/osd.c
index 1c4d3be..157e651 100644
--- a/src/osd.c
+++ b/src/osd.c
@@ -116,6 +116,22 @@ osd_preview_restore(struct server *server)
 	}
 }
 
+/**
+ * prev_node - return node of previous view
+ * @node: node to find the previous node from
+ * Return NULL if previous link is list-head which means node is bottom-most
+ */
+static struct wlr_scene_node *
+prev_node(struct wlr_scene_node *node)
+{
+	struct wlr_scene_node *anchor;
+	anchor = wl_container_of(node->state.link.prev, node, state.link);
+	if (&anchor->state.link == &node->parent->state.children) {
+		return NULL;
+	}
+	return anchor;
+}
+
 static void
 preview_cycled_view(struct view *view)
 {
@@ -128,14 +144,7 @@ preview_cycled_view(struct view *view)
 
 	/* Remember the sibling right before the selected node */
 	osd_state->preview_node = &view->scene_tree->node;
-	osd_state->preview_anchor = wl_container_of(
-		osd_state->preview_node->state.link.prev,
-		osd_state->preview_anchor, state.link);
-	if (&osd_state->preview_anchor->state.link
-			== &osd_state->preview_node->parent->state.children) {
-		/* Sibling was in fact the list head itself so we are the first node */
-		osd_state->preview_anchor = NULL;
-	}
+	osd_state->preview_anchor = prev_node(osd_state->preview_node);
 
 	/* Finally raise selected node to the top */
```

The same function might work for the `wl_container_of()` bit in `view_destroy()`. I didn't try it though.

I get this warning with checkpatch.pl:

```
src/view.c:670: WARNING: Block comments use a trailing */ on a separate line
src/view.c:675: WARNING: Block comments use a trailing */ on a separate line
src/view.c:691: WARNING: Block comments use a trailing */ on a separate line
```

> some kind of marker for the currently selected window

I'm not following you. Why do we need it?

Consolatis:
> In way of simplification - what do you think of this?
> [..]

Much better. Will use it in the next iteration.

> The same function might work for the wl_container_of() bit in view_destroy()

Yes, thought the same. Will try that.

> src/view.c:670: WARNING: Block comments use a trailing */ on a separate line

Will fix.

> some kind of marker for the currently selected window

As discussed at IRC this will likely be some global nodes (could actually be childs of `server->osd_tree`) which will be repositioned / scaled based on the selected view.

Caught one new issue while running this PR: sometimes the wrong preview is shown. This might be caused by trying to preview a minimized application which then just shows whatever was the top window before pressing `Alt-Tab`.
We could likely get away with just force enabling `view->scene_tree->node` and restoring the previous state on switch to the next view. This might not be perfect because AFAIR we actually disconnect the commit handler on minimize as well.

Consolatis:
Had a shot a rebasing this PR and it is completely broken right now.

The reason being: since the workspace PR we are looping over the actual scene_nodes of the current workspace for the `A-Tab` OSD. This PR however changes the ordering of those scene_nodes to show the currently selected window on top of all other windows. That in turn confuses the before mentioned logic to select the next window.

As a result when applying this PR one can switch between two windows only.

Consolatis:
From the logic / design of the feature this PR seems done.
There are 2 more things to do but those are just internal refactorings which should not change any logic.

This PR could need some tests to make sure it doesn't miss any features and feels intuitive to use.

johanmalm:
I'm just pondering the sequencing. In the clip below, when finally cycling from firefox to sakura, should hexchat appear between them?

https://user-images.githubusercontent.com/6955353/185809673-4490b88a-d2a1-488d-9e33-ff8bcfdf84f9.mp4

Also, have updated original post about documenting `osd_preview_restore()`

Consolatis:
> I'm just pondering the sequencing. In the clip below, when finally cycling from firefox to sakura, should hexchat appear between them?

Every time a new application is cycled through (while still being in the OSD) the original layout (before pressing `A-Tab`) is restored and the currently selected application temporarily moved to the front. If only sakura would be on top of firefox it would basically mean to `desktop_move_to_front()` every single application when cycling which IMHO would be more confusing.

I am up for suggestions to change the behavior though to make it more intuitive.

johanmalm:
I've reviewed how Openbox, Ubuntu and Windows handle cycle-preview... and what I've learnt is that.... they don't :smile: (at least not with the [default] settings that I'm using).

I think the way you've implemented it is the right way - and the way alt-tab works on aforesaid systems (albeit they don't do a window-preview).

Consolatis:
This seems ready for a final review / test / merge.

Consolatis:
Thanks everybody for testing

johanmalm:
The big TY goes to you @Consolatis for thinking this through and writing the patch-set :smile:

Issue #310, opened 2022-04-28
Cursor double click error of judgment


sosaxu:
Open two SSD clients, click one client's titlebar, then click another client's titlebar, the second client maximization action will be triggered.

It should be switching focus.

PR #311, opened 2022-04-28
cursor.c: func is_double_click add check view

sosaxu:
add check view when cursor double click, if view is change just return.
[Cursor double click error of judgment](https://github.com/labwc/labwc/issues/310) #310


johanmalm:
Looks really tidy. Thank you and welcome onboard :smile:

Consolatis:
Thanks a lot :)

Issue #312, opened 2022-04-29
First turn off the external screen and then turn it on, the program crashes


fakeczg:
When we use wlr-randr to turn off the external screen first, and then turn it on, the program crashes！

![image](https://user-images.githubusercontent.com/54363314/165873090-f4546855-b3ce-496f-87bc-65961082ddde.png)

Consolatis:
First of all, thanks for your bug report.

Would be great if you could test #315.

fakeczg:
> First of all, thanks for your bug report.
> 
> Would be great if you could test if #315 fixes this issue for you.
I just tested it and it's ok！

![image](https://user-images.githubusercontent.com/54363314/166127563-f3b2ff65-bc89-4b53-af50-59b5f8125ef3.png)

PR #313, opened 2022-04-29
output.c: fix turn off the external screen error

fakeczg:
When we use wlr-randr to turn off the external screen first, and then turn it on, the program crashes！
#312 
Signed-off-by: fakechen <chenzigui@kylinos.cn>
Signed-off-by: sunzhigang <sunzhigang@kylinos.cn>

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -28,15 +28,18 @@ static void
 output_frame_notify(struct wl_listener *listener, void *data)
 {
 	struct output *output = wl_container_of(listener, output, frame);
+	struct wlr_scene *scene = output->server->scene;
+	struct wlr_scene_output *scene_output = wlr_scene_get_scene_output(
+			scene, output->wlr_output);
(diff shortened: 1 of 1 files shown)


johanmalm:
Hi there and thank you for contributing.
Great detective work. It's often not easy to understand what the root-cause is of an assert().
I think it would be better to fix it somewhere other than in output_frame_notify() though as this is a function called a lot. And if `output->scene_output` is incorrectly pointing to NULL, then it might cause problems elsewhere.
I might be wrong here, but should `output->scene_output` be set in output_config_apply() in response to wlr-randr on/off?
I haven't got time to look tonight. If you've got time to investigate that would be great. No pressure though :smile:

fakeczg:
Ok, thank you for your guidance, I'll find out the reason when I get to work

Consolatis:
Replaced by #315

PR #315, opened 2022-04-30
src/output.c: Make sure we are always using the correct scene_node

### LINKED ISSUE #312: First turn off the external screen and then turn it on, the program crashes

Consolatis:
Also make sure we are only applying layout specific changes
once the output specific ones were committed successfully.

Fixes #312
Reported-By: @fakeczg

I only had a quick test with a single screen KMS/DRM setup and this specific issue can't be tested using a nested Wayland / X11 setup (these backends do not actually support disabling outputs at all).

So some more testing (especially with multiple screens) would be really appreciated: just some `wlr-randr --output HDMI-A --off` and `wlr-randr --output HDMI-A --on` cycles should be enough.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -169,15 +169,7 @@ new_output_notify(struct wl_listener *listener, void *data)
 	}
 
 	wlr_output_layout_add_auto(server->output_layout, wlr_output);
-
-	/* TODO: check this makes sense */
-	struct wlr_scene_output *scene_output;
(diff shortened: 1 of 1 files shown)


johanmalm:
Great effort... not an easy one to bottom out.

PR #321, opened 2022-05-09
meson: use proper dependency fallback for wlroots

### LINKED ISSUE #318: uninitialized 'resize_edges' when building with system wlroots
i already have wlroots 0.15.1 and seatd installed why this is downloading and using a subproject?
how do i change the meson.build to use system wlroots and system seatd instead of  using the subproject?
im making the package for slackware.

eli-schwartz:
Make the subproject wrap file declare the dependencies it provides.

Remove the manual subproject invocation and allow Meson to select whichever wlroots it finds which satisfies the version requirement -- either a system one via pkg-config, or the subproject fallback.

Use a dependency-generic method of acquiring wlroots configuration info -- enabled features are present in the pkg-config file and additionally exported as declare_dependency() variables, so there is no need to do C preprocessor checks for it.

This ensures that Meson best practices are followed, and also...

Fixes #318

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -37,22 +37,14 @@ if git.found()
 endif
 add_project_arguments('-DLABWC_VERSION=@0@'.format(version), language: 'c')
 
-wlroots_version = ['>=0.16.0', '<0.17.0']
-wlroots_proj = subproject(
+wlroots = dependency(
   'wlroots',
   default_options: ['default_library=static', 'examples=false'],
-  required: false,
-  version: wlroots_version,
+  version: ['>=0.16.0', '<0.17.0'],
 )
 
-if wlroots_proj.found()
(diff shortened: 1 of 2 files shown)


Consolatis:
> Dependency wlroots found: NO found 0.15.1 but need: '>=0.16.0' ; matched: '<0.17.0'
> Looking for a fallback subproject for the dependency wlroots
> Executing subproject wlroots 

Sweet :)
Thanks a lot for your help.

I am guessing all that is required to backport this PR to the v0.5 branch as well is to change the wlroots version requirements?
https://github.com/labwc/labwc/blob/v0.5/meson.build

eli-schwartz:
Yep, should be.

Consolatis:
.. and backported. Thanks again @eli-schwartz

PR #333, opened 2022-05-19
Add workspaces

Consolatis:
This ~~is just a simple test implementation for workspaces~~ implements workspaces.
I'd welcome some more testing and review.

### Config
2 new actions:
- `SendToDesktop` `target` (move current view to workspace `target`, also available via window menu)
- `GoToDesktop` `target` (change active workspace to `target`)

Workspaces have to be defined in `rc.xml`. Example:
```xml
  <desktops>
    <popupTime>1000</popupTime>
    <names>
      <name>Workspace a</name>
      <name>Workspace b</name>
      <name>Workspace c</name>
    </names>
  </desktops>
```
Actions have to use `to` for their `target` argument, a `target` can be defined as:
- `left` / `right`
- the workspace index, beginning from 1
- the full name of the workspace

Example:
```xml
    <keybind key="W-Left">
      <action name="GoToDesktop" to="left" />
    </keybind>

    <keybind key="W-S-Left">
      <action name="SendToDesktop" to="left" />
      <action name="GoToDesktop" to="left" />
    </keybind>

    <keybind key="W-1">
      <action name="GoToDesktop" to="Workspace a" />
    </keybind>

    <keybind key="W-2">
      <action name="GoToDesktop" to="2" />
    </keybind>
```

### OSD
When switching to a different workspace, a non-interactive workspace OSD pops up.
If the workspace switch had been initiated using a modifier keybinding the OSD will close when all modifiers are released. If the workspace switch instead had been initiated using some other action like using the window menu the OSD will close after `popupTime` ms. A `popupTime` of `0` disables both variants.

### Nested window menu
By default the `client-menu` (also known as window menu) has a new nested menu with id `workspaces` which has two entries `Move left` and `Move right`. That menu can be removed or redefined as usual via a custom `menu.xml`.
If there is only a single workspace, all menu items pointing to a menu with id `workspaces` will be removed automatically.

### Other

As this started as simple test implementation I chose the most simple design:
- A workspace covers all outputs
- The `A-Tab` OSD only shows windows from the active workspace
- A window can only be in a single workspace at any given point in time (no tags)
- External panels will show all windows. When activating a window which is on a different workspace labwc will switch to that workspace automatically

Whatever we decide to do (or not to do), using external panels to switch / manage workspaces or move windows into workspaces seems a looong time away:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/59
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/60

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -40,6 +40,32 @@
   </snapping>
 
   <!--
+    Use GoToDesktop left | right to switch workspaces.
+    Use SendToDesktop left | right to move windows.
+    See man labwc-actions for futher information.
+
+    Workspaces can be configured like this:
+    <desktops>
+      <popupTime>1000</popupTime>
+      <names>
+        <name>Workspace 1</name>
+        <name>Workspace 2</name>
+        <name>Workspace 3</name>
+      </names>
+    </desktops>
+  -->
+  <desktops>
+    <!--
+      popupTime defaults to 1000 so could be left out.
+      Set to 0 to completely disable the workspace OSD.
+    -->
+    <popupTime>1000</popupTime>
+    <names>
+      <name>Default</name>
+    </names>
+  </desktops>
+
+  <!--
     Keybind actions are specified in labwc-actions(5)
     The following keybind modifiers are supported:
       W - window/super/logo
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -55,6 +55,11 @@ struct rcxml {
 
 	/* cycle view (alt+tab) */
 	bool cycle_preview_contents;
+
+	struct {
+		int popuptime;
+		struct wl_list workspaces;  /* struct workspace.link */
+	} workspace_config;
 };
 
 extern struct rcxml rc;
--- a/include/workspaces.h
+++ b/include/workspaces.h
@@ -1,0 +1,28 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_WORKSPACES_H
+#define __LABWC_WORKSPACES_H
+
+struct seat;
+struct view;
+struct server;
+struct wl_list;
+
+/* Double use: as config in config/rcxml.c and as instance in workspaces.c */
+struct workspace {
+	struct wl_list link; /* struct server.workspaces
+	                        struct rcxml.workspace_config.workspaces */
+	struct server *server;
+
+	char *name;
+	struct wlr_scene_tree *tree;
+};
+
+
+void workspaces_init(struct server *server);
+void workspaces_switch_to(struct workspace *target);
+void workspaces_send_to(struct view *view, struct workspace *target);
+void workspaces_destroy(struct server *server);
+void workspaces_osd_hide(struct seat *seat);
+struct workspace * workspaces_find(struct workspace *anchor, const char *name);
+
+#endif /* __LABWC_WORKSPACES_H */
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -1,0 +1,371 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include <cairo.h>
+#include <pango/pangocairo.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include "labwc.h"
+#include "common/font.h"
+#include "common/zfree.h"
+#include "workspaces.h"
+
+/* Internal helpers */
+static size_t
+parse_workspace_index(const char *name)
+{
+	/*
+	 * We only want to get positive numbers which span the whole string.
+	 *
+	 * More detailed requirement:
+	 *  .---------------.--------------.
+	 *  |     Input     | Return value |
+	 *  |---------------+--------------|
+	 *  | "2nd desktop" |      0       |
+	 *  |    "-50"      |      0       |
+	 *  |     "0"       |      0       |
+	 *  |    "124"      |     124      |
+	 *  |    "1.24"     |      0       |
+	 *  `------------------------------´
+	 *
+	 * As atoi() happily parses any numbers until it hits a non-number we
+	 * can't really use it for this case. Instead, we use strtol() combined
+	 * with further checks for the endptr (remaining non-number characters)
+	 * and returned negative numbers.
+	 */
+	long index;
+	char *endptr;
+	errno = 0;
+	index = strtol(name, &endptr, 10);
+	if (errno || *endptr != '\0' || index < 0) {
+		return 0;
+	}
+	return index;
+}
+
+/*
+ * TODO: set_source and draw_border are straight up copies from src/osd.c
+ * find some proper place for them instead of duplicating stuff.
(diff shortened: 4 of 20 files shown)


acoolstraw:
I think it'd be nice to add a different OSD for each workspace, containing only the windows in that workspace

Consolatis:
That last commit is **really** hacky. Guess we could actually loop over `server->workspace_current->tree->node.state.children` and use the node descriptor to get back our view instead.


**If** we want to implement workspaces like this (including hiding other workspace windows from the OSD) I think we should still switch to the correct workspace when an application is activated via foreign toplevel protocol (like from a panel).

I had implemented that  in the second commit via `desktop.c`  but that also caused the switch to happen when the last window on a given workspace had been closed so I disabled it again,

Reverting the last commit should restore showing all the windows via OSD and reenable "switch to correct workspace" when using a panel.

Consolatis:
> Guess we could actually loop over server->workspace_current->tree->node.state.children and use the node descriptor to get back our view instead.
Did that in the latest commit, also added a workspace indicator on top of the OSD.

Somewhat unrelated to this PR, I didn't look into it all that much but I think we could actually get rid of `server->views` completely. Its basically the same as the `server->view_tree->node.state.children` + node descriptors to get the views of those.

Consolatis:
Two minor changes:
- Closing a window should now focus the top-most window on the current workspace
- Switching workspaces now focuses the top-most window on the new workspace

Narrat:
Although I know labwc can be built with wlroots as a subproject and with a working commit, when I tried to built this branch with latest wlroots master it failed to compile. Went back to labwc master and it failed with the same build error.
In case it wasn't known.
Nevertheless I will report back when I tested this branch :) Really excited to see that time and work went into this feature. Thanks for that.

Consolatis:
wlroots master has quite some churn and we might have to chase some more recent changes again.
Until then, building with `meson --force-fallback-for=wlroots build_master && meson compile -C build_master` should get it built with the older commit from the subproject.

docs/rc.xml.all
```diff
@@ -38,6 +38,19 @@
...
+    Use SendToDesktop left | right to move windows.
+    See man labwc-actions for futher information.
+  -->
+  <desktops>
+    <names>
+      <name>Workspace 1</name>
+      <name>Workspace 2</name>
+      <name>Workspace 3</name>
+    </names>
+  </desktop>
```
Narrat, line 52: Small typo I noticed. Should be `</desktops>` :)

Narrat:
Had a short test with this and it worked great :) No issues or odd behaviour. I will test this more (even as a daily driver) and see what comes around.

docs/rc.xml.all
johanmalm: Great pick up. Thanks

Consolatis: fixed, thanks

Consolatis:
After some feedback from @johanmalm and some further discussion via IRC:
- [ ] Goal: on a default config this PR should not change \*anything at all\*
   - [ ] Hide OSD workspace label when there is only one workspace configured
   - [ ] Hide default window-menu entry for moving the view into a different workspace if there is only one workspace configured
   - [ ] Use a comment for the workspace config in `docs/rc.xml.all` as it should default to a single workspace
- [ ] Provide a workspace OSD which shows up on workspace switches and automatically hides after a specified time, configured via `rc.xml` (`0` should disable the workspace OSD completely, default value will likely be 1000 ms / 1 s)
- [ ] Rename the default namespace into "There is only one thing better than tomatoes: tomatoes"

Prototype of the workspace OSD:
workspace_osd_2

johanmalm:
>     * [ ]  Rename the default namespace into "There is only one thing better than tomatoes: tomatoes"

Assume this is to check whether or not I actually pay attention? :smile:

heroin-moose:
Works neatly for me, I have 9 workspaces and so far everything works good.

acoolstraw:
OSD gets stuck with only the workspace name showing if you close a window and then A-Tab. might be some issue with my Void (musl, runit) system, I don't recall it happening on Arch

ahmadraniri1994:
It's great.

Consolatis:
@heroin-moose @acoolstraw @ahmadraniri1994 thanks for testing :)

I started to rebase this branch on top of the latest master which includes massive patches to handle the upstream wlroots refactoring and am currently testing everything again. As soon as that is ready and I don't detect any new issues I will update this PR branch.

Current issues unrelated to the rebasing so far:
- [ ] When having a always-on-top window with keyboard focus and switching workspaces, that window will loose the focus and it will be given to the top-most window of the new workspace instead
- [ ] always-on-top windows will never show up in the A-Tab OSD because they have their own global parent and are technically not part of any workspace. This is an implementation issue and I am currently not sure how to solve it
- [ ] A-Tab OSD gets stuck with only the workspace name showing if you close a window and then A-Tab (this might be a race condition, when a window is destroyed it should make the A-Tab OSD disappear if there are no other windows on the current workspace).

include/workspaces.h
```diff
@@ -0,0 +1,28 @@
...
+
+/* Forward declarations */
+struct seat;
+struct view;
+struct server;
+struct wl_list;
+
+/* Double use: as config in config/rcxml.c and as instance in workspaces.c */
+struct workspace {
+	struct wl_list link; /* server::workspaces */
```
Consolatis, line 13:
```
/* struct server.workspaces
   struct rcxml.workspace_config.workspaces */
```

src/workspaces.c
```diff
@@ -0,0 +1,371 @@
...
+}
+
+void
+workspaces_destroy(struct server *server)
+{
+	struct workspace *workspace, *tmp;
+	wl_list_for_each_safe(workspace, tmp, &server->workspaces, link) {
+		wlr_scene_node_destroy(&workspace->tree->node);
+		free(workspace->name);
+		workspace->name = NULL;
```
Consolatis, line 366: `zfree()`

Consolatis (commented):
Other than the following to minor things I am satisfied with the PR, the always-on-top loosing keyboard-focus issue can be solved later. It seems pretty minor in comparison to the overall scope of this PR.

include/workspaces.h
```diff
@@ -0,0 +1,28 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_WORKSPACES_H
+#define __LABWC_WORKSPACES_H
+
+/* Forward declarations */
```
Consolatis, line 5: remove

include/workspaces.h
Consolatis, line 13: `struct server.workspaces`

include/config/rcxml.h
```diff
@@ -55,6 +55,11 @@ struct rcxml {
 
 	/* cycle view (alt+tab) */
 	bool cycle_preview_contents;
+
+	struct {
+		int popuptime;
+		struct wl_list workspaces;
```
Consolatis, line 61: `struct workspace.link`

acoolstraw:
Awesome!!!!!

PR #341, opened 2022-05-24
src/cursor.c: Fix branch condition for out-of-view selecting/dragging

### LINKED ISSUE #340: Drawing on swappy offsets the brush position
> * [ ]  Drawing on swappy (`grim -g "$(slurp)" - | swappy -f -`) is buggy, as for some reason the brush poiner is offset diagonally, but your actual cursor stays where you originally started the drawing

Caused by 

https://github.com/labwc/labwc/blob/fbcdf38b52bb5a9635962e346c4f82e8c932ec89/src/cursor.c#L296

The offset is because `scene_xdg_surface_update_position()` takes into account the invisible border of the view and positions the node accordingly in https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/xdg_shell.c#L57

_Originally posted by @johanmalm in https://github.com/labwc/labwc/issues/332#issuecomment-1133950545_

Consolatis:
Instead of using the stored view for comparison use the actual surface.

Before this patch, there were situations where the branch intended for
out-of-window text selection / scrollbar dragging was used even though
the cursor never left the surface.

Partly fixes #340

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -91,16 +91,19 @@ struct seat {
 	struct wlr_layer_surface_v1 *focused_layer;
 
 	/**
-	 * active_view will usually be NULL and is only set on button press
-	 * while the mouse is over a view surface and reset to NULL on button
-	 * release.
+	 * pressed view/surface will usually be NULL and is only set on button
+	 * press while the mouse is over a view surface and reset to NULL on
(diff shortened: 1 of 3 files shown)


johanmalm:
:smile:

Issue #278, opened 2022-03-15
When using wapanel some surfaces do not receive cursor-button


johanmalm:
When using https://github.com/Firstbober/wapanel some areas do not receive cursor-button.

Do we need  to create nodes for layer-shell subsurfaces so that `desktop_node_and_view_at()` returns the correct node/surface?

The same issue has been observed with the gkt-layer-shell popup plus/minus buttons

Consolatis:
The gtk-layer-demo issue with the `+` and `-` buttons for margin also happens for v0.5.

Consolatis:
The focus is now shifted to 0.6 / latest master branch, thus: closing again.

PR #342, opened 2022-05-24
desktop_node_and_view_at(): handle layer surfaces

Consolatis:
\+ Replace some tabs with spaces in `include/node.h`

Fixes: #278

Only tested with `gtk-layer-demo` but that turned from not-working into working so I expect `wapanel` to be fixed as well.

This PR uncovered two more issues with the previous cursor fix:
- with this PR we now have a (layer) surface without view which segfaults when doing the out-of-surface-thingie with a layer surface (as we are using the `view` geometry)
- when out-of-surface, we don't actually send the button release event to the correct surface which in case of the gtk-layer-demo causes the respective button to trigger continuously

For this reason I'll put the PR into draft mode until the two new issues are fixed.

### DIFF
--- a/include/node.h
+++ b/include/node.h
@@ -29,10 +29,10 @@ struct node_descriptor {
  * @scene_node: wlr_scene_node to attached node_descriptor to
  * @type: node descriptor type
  * @data: struct to point to as follows:
- *   - LAB_NODE_DESC_VIEW		struct view
- *   - LAB_NODE_DESC_XDG_POPUP		struct view
- *   - LAB_NODE_DESC_LAYER_SURFACE	struct lab_layer_surface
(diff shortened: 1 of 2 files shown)


Consolatis:
> This PR uncovered two more issues with the previous cursor fix:
> 
> * with this PR we now have a (layer) surface without view which segfaults when doing the out-of-surface-thingie with a layer surface (as we are using the `view` geometry)
> * when out-of-surface, we don't actually send the button release event to the correct surface which in case of the gtk-layer-demo causes the respective button to trigger continuously
> 
> For this reason I'll put the PR into draft mode until the two new issues are fixed.

PR #344 contains fixes for both issues. Once that one is merged I'll rebase this PR to get rid of the duplicated commits.

PR #346, opened 2022-05-26
Various smallish fixes

Consolatis:
Various smallish fixes

When using a font size of 0, random stuff starts exploding so I tried to fix that in two commits:
- [[wip] deal with font_buffer creation failures](https://github.com/labwc/labwc/commit/be28373cda704e6f0e70365e3e7016f65a2f8f5e)
  This commit deals in a general way with errors from creating a font_buffer.
- [[wip] font_buffer_create: clamp size](https://github.com/labwc/labwc/commit/16bf2f85e8e1e52d36276b88bf8f48f74cce0a23)
  This commit prevents creating font_buffers with a size < 1x1 so it shouldn't actually trigger above fallback paths.

Before merging these two commits I'd welcome some second opinions.

Also: some memory leak fixes
- [src/osd.c: Free buf.buf as its malloc'd in buf_init()](https://github.com/labwc/labwc/commit/b078bc95911a77f857a3226f047e7a5345efc074)
- [src/theme.c: Clean up corner buffers on finish](https://github.com/labwc/labwc/commit/5d43a2c161337f24dcbafc049dda63cd3cd8c95e) (`wlr_log()` will be removed before merging)

### DIFF
--- a/src/osd.c
+++ b/src/osd.c
@@ -103,6 +103,8 @@ osd_update(struct server *server)
 
 	struct theme *theme = server->theme;
 
+	struct buf buf;
+	buf_init(&buf);
 	struct output *output;
 	wl_list_for_each(output, &server->outputs, link) {
 		destroy_osd_nodes(output);
@@ -168,10 +170,7 @@ osd_update(struct server *server)
 
 		pango_cairo_update_layout(cairo, layout);
 
-		struct buf buf;
-		buf_init(&buf);
 		y = OSD_BORDER_WIDTH;
-
 		y += (OSD_ITEM_HEIGHT - font_height(&font)) / 2;
 
 		wl_list_for_each(view, &server->views, link) {
@@ -221,4 +220,5 @@ osd_update(struct server *server)
 		wlr_scene_node_set_position(&scene_buffer->node, lx, ly);
 		wlr_scene_node_set_enabled(&output->osd_tree->node, true);
 	}
+	free(buf.buf);
 }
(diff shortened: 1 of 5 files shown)


src/osd.c
```diff
@@ -204,6 +204,7 @@ osd_update(struct server *server)
 			pango_cairo_show_layout(cairo, layout);
 			y += OSD_ITEM_HEIGHT;
 		}
+		free(buf.buf);
```
johanmalm, line 207: `buf_init()` is called with the `wl_list_for_each(..., &server->outputs, ...)` which isn't great.
We ought to move these two lines to before the foreach loop

```
struct buf buf;
buf_init(&buf);
```

And then in the loop just set buf.len = 0 to clear it.

That way we know buf.buf can be freed. If we have not outputs we risk freeing a NULL pointer otherwise.

johanmalm (approved):

src/osd.c
Consolatis: Isn't the new `free()` within the same loop as well though?

> We ought to move these two lines to before the foreach loop
> And then in the loop just set buf.len = 0 to clear it.

Yes, that is a good idea, will change

Consolatis:
As discussed at IRC:
- I removed the "clamp size" patch because it may hide bugs in the future
- `buf_init()` and `free(buf.buf)` has been moved outside of the output loop.

From my POV this seems ready to rebase-and-merge unless I missed something.

PR #356, opened 2022-05-30
xwayland: call foreign-toplevel-destroy on unmap

johanmalm:
Some xwayland clients leave unmapped child views around, typically when a
dialog window is closed. Although handle_destroy() is not called for
these, we have to call foreign-toplevel-destroy to avoid panels and the
like incorrecly showing them.


johanmalm:
@Consolatis Were you okay with this one? Without this patch, the whole compositor can be crashed by foreign-toplevel clients.

Consolatis:
Good catch, diff makes sense to me and I don't see it breaking anything.

Also had a quick test and minimizing and un-minimizing for an X11 app seems to work just fine.

Issue #280, opened 2022-03-16
Add more theming / configuration options


Consolatis:
- (max) menu width
possibly combined with calculation of dynamic width based on the widest element within a given menu node
- menu font padding within a single menu item
- submenu bullet
- button width
- button layout
- button hover: #271
- other suggestions?

acoolstraw:
- menu border
- menu corner radius
- alt tab border width and color
- alt tab bg and fg colors

Consolatis:
Related: https://github.com/labwc/labwc-acceptance-criteria/issues/2

acoolstraw:
[ ] osd.hilight.* for workspace and alt tab OSD styling

anohren:
Would you prefer I make a suggestion here or add a separate issue?

I was thinking it'd be nice with three groups of identifiers in <theme><titlebar><layout> and the introduction of a `title` identifier, such that I could write e.g: `close,max,iconify:icon,title:`, and get the middle group perfectly centered.

johanmalm:
Feature requests go in discussions to keep issue noise down.

The one above is out-of-scope. Was discussed extensively when last refactoring title layout.

PR #358, opened 2022-05-31
[wip] theme: implement osd.border.color and osd.border.width

Consolatis:
Related to #280 and the yet-to-be-pushed workspace OSD for #333.

http://openbox.org/wiki/Help:Themes#osd.border.color
http://openbox.org/wiki/Help:Themes#osd.border.width

The debug log in `src/theme.c` will obviously be removed before turning this draft into a mergeable PR.

Open questions:
-  somewhat unrelated to this specific PR, do we want to check for forbidden negative values in post processing of `theme.c`?
   (like negative borders or padding)
- do want to keep the fallback for `osd.border.color` like this?
```c
/*
* As per http://openbox.org/wiki/Help:Themes#osd.border.color
* we should fall back to `window_active_border_color` but
* that is usually the same as `window_active_title_bg_color`
* and thus the fallback for `osd_bg_color`. Which would mean
* they are both the same color and thus the border is invisible.
*
* Instead, we fall back to `osd_label_text_color` which in turn
* falls back to `window_active_label_text_color`.
*/
```
This is also what was previously hardcoded for the OSD border color and I think it makes sense to diverge from the Openbox specs in this case.
Its only the fallback path anyway so if a theme is configuring `osd.border.color` that will get prioritized.


src/osd.c
```diff
@@ -24,6 +24,29 @@ set_source(cairo_t *cairo, float *c)
 	cairo_set_source_rgba(cairo, c[0], c[1], c[2], c[3]);
 }
 
+/* Draws a border with a specified border width */
+static void
+draw_border(cairo_t *cairo, double width, double height, double border_width)
+{
+	double line_width = cairo_get_line_width(cairo);
```
johanmalm, line 31: You can use `cairo_save()` for this

src/osd.c
```diff
@@ -24,6 +24,29 @@ set_source(cairo_t *cairo, float *c)
...
+	double x, y, w, h;
+	/* The anchor point of a line is in the center */
+	x = y = border_width / 2;
+	w = width - border_width;
+	h = height - border_width;
+	cairo_rectangle(cairo, x, y, w, h);
+	cairo_stroke(cairo);
+
+	/* Restore original line width */
+	if (line_width != border_width) {
```
johanmalm, line 45: cairo_restore()

src/osd.c
```diff
@@ -24,6 +24,29 @@ set_source(cairo_t *cairo, float *c)
 	cairo_set_source_rgba(cairo, c[0], c[1], c[2], c[3]);
 }
 
+/* Draws a border with a specified border width */
+static void
+draw_border(cairo_t *cairo, double width, double height, double border_width)
```
johanmalm, line 29: Could do `s/border_width/line_width`... just a matter of taste.

johanmalm:
> Open questions:
> 
>     * somewhat unrelated to this specific PR, do we want to check for forbidden negative values in post processing of `theme.c`?
>       (like negative borders or padding)

Yes, good idea. Maybe a `sanity_check()` after `post_processing()`. Just thinking about the issue the other day with font-size==0. Could be a good place to check things like that too.


>     * do we want to keep the fallback for `osd.border.color` like this?

Agree. Good idea.

> This is also what was previously hardcoded for the OSD border color and I think it makes sense to diverge from the Openbox specs in this case. Its only the fallback path anyway so if a theme is configuring `osd.border.color` that will get prioritized.

That might be the "oddity" I remember :smile: (further to IRC discussion)

src/osd.c
Consolatis: :heavy_check_mark:

src/osd.c
Consolatis: :heavy_check_mark:

src/osd.c
Consolatis: As `border_width` could theoretically be confused with `width of the whole border`: :heavy_check_mark:

src/theme.c
```diff
@@ -28,6 +28,8 @@
 #include "buffer.h"
 #include "ssd.h"
 
+#define CONF_UNSET (-1)
```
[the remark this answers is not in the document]
Consolatis: Good point, `INT_MIN` is much better for this case. :heavy_check_mark:

Consolatis:
> The debug log in src/theme.c will obviously be removed before turning this draft into a mergeable PR.

Removed, unless I missed something: ready to go from my side

PR #373, opened 2022-06-05
Chase wlroots: upstream scene_node refactoring

Consolatis:
These are likely the biggest changes regarding the scene-graph API we had to adjust to since moving to the scene-graph API in the first place.

I split everything into their own commits, each following a upstream commit so we are able to diagnose things better if something goes wrong. I had a quick test on each step along the way but when seeing the amount of changes required I think it makes sense for others to test and/or have a look at the single commits before merging this.

As always: use `meson subprojects update wlroots` to check out the corresponding wlroots upstream commit before building.

See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3509/commits for the upstream commits.

I will push commit by commit into this branch so the CI triggers for all of them.

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -172,10 +172,14 @@ dump_tree(struct server *server, struct wlr_scene_node *node,
 		printf("%*c%s\n", pos + 4 + INDENT_SIZE, ' ', "<skipping children>");
 		return;
 	}
-	struct wlr_scene_node *child;
-	wl_list_for_each(child, &node->children, link) {
-		dump_tree(server, child, pos + INDENT_SIZE,
-			x + child->x, y + child->y);
+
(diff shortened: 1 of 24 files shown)


src/debug.c
```diff
@@ -9,17 +9,18 @@
 #define HEADER_CHARS "------------------------------"
 
 #define INDENT_SIZE 3
-#define IGNORE_SSD true
+#define IGNORE_SSD false
```
Consolatis, line 12: woops.

Consolatis (commented):
`Pull request authors can't request changes on their own pull request`

That is strange. ..I think?

Consolatis:
> The xbm buttons in the corner pixmaps are vertically offset (ca 10 pixels to low)

Thanks for spotting this, should be fixed in the latest version.

PR #374, opened 2022-06-05
xwayland-unmanaged: Fix set_geometry using container_of of wrong member

Joshua-Ashton:
Fixes a segfault with starting Steam.


johanmalm:
Thanks.

PR #378, opened 2022-06-06
Fix/view center

johanmalm:


johanmalm:
I've tested with

- both SSD/CSD
- two physical monitors with different scales

Consolatis:
Also had a quick test on a nested labwc:
- scale 0.5, 1 and 2
- SSD (`foot` wayland native)
- CSD (`d-feet` once with `GDK_BACKEND=x11` and once as wayland native)

All seem starting centered so once the `view->type` check is guarded by an `#if HAVE_XWAYLAND` this seems ready to go.

heroin-moose:
^_^

heroin-moose:
Works for me too.

PR #387, opened 2022-06-10
src/action.c: Convert action->arg to a list of action_arg

Consolatis:
In preperation for Actions that require multiple arguments

### DIFF
--- a/include/action.h
+++ b/include/action.h
@@ -2,19 +2,23 @@
 #ifndef __LABWC_ACTION_H
 #define __LABWC_ACTION_H
 
-struct server;
 struct view;
+struct server;
+struct wl_list;
 
 struct action {
-	uint32_t type;
-	char *arg;
-	struct wl_list link;
+	struct wl_list link;  /* struct keybinding.actions,
+	                       * struct mousebinding.actions,
+	                       * struct menuitem.actions */
+
+	uint32_t type;        /* enum action_type */
+	struct wl_list args;  /* struct action_arg.link */
 };
 
 struct action *action_create(const char *action_name);
-void action_list_free(struct wl_list *action_list);
-
+void action_arg_add_str(struct action *action, char *key, const char *value);
 void actions_run(struct view *activator, struct server *server,
 	struct wl_list *actions, uint32_t resize_edges);
+void action_list_free(struct wl_list *action_list);
 
-#endif
+#endif /* __LABWC_ACTION_H */
--- a/include/config/keybind.h
+++ b/include/config/keybind.h
@@ -11,8 +11,8 @@ struct keybind {
 	uint32_t modifiers;
 	xkb_keysym_t *keysyms;
 	size_t keysyms_len;
-	struct wl_list actions;
-	struct wl_list link;
+	struct wl_list actions;  /* struct action.link */
+	struct wl_list link;     /* struct rcxml.keybinds */
 };
 
 /**
--- a/include/private/action.h
+++ b/include/private/action.h
@@ -1,0 +1,24 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_PRIVATE_ACTION_H
+#define __LABWC_PRIVATE_ACTION_H
+
+/* Don't include ourself as search path starts at current directory */
+#include "../action.h"
+
+enum action_arg_type {
+	LAB_ACTION_ARG_STR = 0,
+};
+
+struct action_arg {
+	struct wl_list link;        /* struct action.args */
+
+	const char *key;            /* May be NULL if there is just one arg */
+	enum action_arg_type type;
+};
+
+struct action_arg_str {
+	struct action_arg base;
+	char *value;
+};
+
+#endif /* __LABWC_PRIVATE_ACTION_H */
--- a/src/action.c
+++ b/src/action.c
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #define _POSIX_C_SOURCE 200809L
+#include <assert.h>
 #include <signal.h>
+#include <string.h>
 #include <strings.h>
 #include <unistd.h>
 #include <wlr/util/log.h>
@@ -9,8 +11,8 @@
 #include "debug.h"
 #include "labwc.h"
 #include "menu/menu.h"
+#include "private/action.h"
 #include "ssd.h"
-#include "action.h"
 #include "workspaces.h"
 
 enum action_type {
@@ -64,6 +66,24 @@ const char *action_names[] = {
 	NULL
 };
 
+static char *
+action_str_from_arg(struct action_arg *arg)
+{
+	assert(arg->type == LAB_ACTION_ARG_STR);
+	return ((struct action_arg_str *)arg)->value;
+}
+
+static struct action_arg *
+action_get_first_arg(struct action *action)
+{
+	struct action_arg *arg;
+	struct wl_list *item = action->args.next;
+	if (item == &action->args) {
+		return NULL;
+	}
+	return wl_container_of(item, arg, link);
+}
+
 static enum action_type
 action_type_from_str(const char *action_name)
 {
@@ -85,15 +105,26 @@ action_create(const char *action_name)
 	}
 	struct action *action = calloc(1, sizeof(struct action));
 	action->type = action_type_from_str(action_name);
+	wl_list_init(&action->args);
 	return action;
 }
 
 void action_list_free(struct wl_list *action_list)
 {
+	struct action_arg *arg, *arg_tmp;
 	struct action *action, *action_tmp;
+	/* Free actions */
 	wl_list_for_each_safe(action, action_tmp, action_list, link) {
 		wl_list_remove(&action->link);
-		zfree(action->arg);
+		/* Free args */
+		wl_list_for_each_safe(arg, arg_tmp, &action->args, link) {
+			wl_list_remove(&arg->link);
+			zfree(arg->key);
+			if (arg->type == LAB_ACTION_ARG_STR) {
+				free(action_str_from_arg(arg));
+			}
+			zfree(arg);
+		}
 		zfree(action);
 	}
 }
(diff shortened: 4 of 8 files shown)


include/action.h
```diff
@@ -2,19 +2,48 @@
...
+	uint32_t type;     /* enum action_arg_type */
+};
+
+
+struct action *action_create(const char *action_name);
+void action_add_str_arg(struct action *action, char *name, const char *value);
+void action_add_int_arg(struct action *action, char *name, int value);
+void action_add_double_arg(struct action *action, char *name, double value);
+void action_add_bool_arg(struct action *action, char *name, bool value);
+struct wlr_box *action_add_box_arg(struct action *action, char *name);
```
Consolatis, line 44: Only `action_add_str_arg()` is used in this PR, the others are examples of how this action_arg framework could be used

include/private/action.h
```diff
@@ -0,0 +1,41 @@
...
+
+struct action_arg_bool {
+	struct action_arg base;
+	bool value;
+};
+
+struct action_arg_box {
+	struct action_arg base;
+	struct wlr_box value;
+};
```
Consolatis, line 39: Only `struct action_arg_str` is used in this PR, the others are examples of how this action_arg framework could be used in the future

src/action.c
```diff
@@ -56,6 +60,52 @@ const char *action_names[] = {
...
+	assert(arg->type == LAB_ACTION_ARG_BOOL);
+	return ((struct action_arg_bool *)arg)->value;
+}
+
+static struct wlr_box *
+action_box_from_arg(struct action_arg *arg)
+{
+	assert(arg->type == LAB_ACTION_ARG_BOX);
+	return &((struct action_arg_box *)arg)->value;
+}
```
Consolatis, line 96: Only `action_str_from_arg()` is used in this PR, the others are examples of how this action_arg framework could be used in the future

src/action.c
```diff
@@ -267,9 +335,70 @@ actions_run(struct view *activator, struct server *server,
...
+action_add_box_arg(struct action *action, char *name)
+{
+	struct action_arg_box *arg = calloc(1, sizeof(*arg));
+	arg->base.type = LAB_ACTION_ARG_BOX;
+	if (name) {
+		arg->base.name = strdup(name);
+	}
+	wl_list_insert(action->args.prev, &arg->base.link);
+	return &arg->value;
+}
```
Consolatis, line 404: Only `action_add_str_arg()` is used in this PR, the others are examples of how this action_arg framework could be used in the future

include/action.h
```diff
@@ -2,19 +2,48 @@
 #ifndef __LABWC_ACTION_H
 #define __LABWC_ACTION_H
 
-struct server;
+/* Forward declare arguments */
```
johanmalm, line 5: Consider "/* Forward declaration of opaque structs for pointer arguments */"
Although I think I prefer leaving it out because we should try to do that wherever we can.

include/action.h
```diff
@@ -2,19 +2,48 @@
...
 
 struct action {
-	uint32_t type;
-	char *arg;
-	struct wl_list link;
+	struct wl_list link;  /* struct keybinding.actions,
+	                       * struct mousebinding.actions,
+	                       * struct menuitem.actions */
+
+	uint32_t type;        /* enum action_type */
```
johanmalm, line 16: Wouldn't it be better to declare the enum in this header-file and just make this `enum action_type type`.
I'm thinking that other translation units will want to do things like `if (action->type == LAB_ACTION_WHATEVER)`

include/action.h
```diff
@@ -2,19 +2,48 @@
...
+	struct wl_list args;  /* struct action_arg.link */
 };
 
-struct action *action_create(const char *action_name);
-void action_list_free(struct wl_list *action_list);
+struct action_arg {
+	struct wl_list link;  /* struct action.args */
 
+	/**
+	 * FIXME: Do we want to replace `name` by a plain uint32_t and have
```
johanmalm, line 24: Don't mind. action_run() isn't a function that gets called that often, so the tiny overhead of a strcmp()s probably pales into insignificance :smile:

include/action.h
```diff
@@ -2,19 +2,48 @@
...
+	 * things like a wlr_box which gets filled while parsing the
+	 * key/mouse/menu binds. `MoveResizeTo` comes to mind :)
+	 */
+	const char *name;  /* May be NULL if there is just one arg */
+	uint32_t type;     /* enum action_arg_type */
+};
+
+
+struct action *action_create(const char *action_name);
+void action_add_str_arg(struct action *action, char *name, const char *value);
```
johanmalm, line 40: Would be more consistent to name these functions `action_add_arg_XX` or even `action_arg_add_XX`

include/action.h
```diff
@@ -2,19 +2,48 @@
...
+void action_add_int_arg(struct action *action, char *name, int value);
+void action_add_double_arg(struct action *action, char *name, double value);
+void action_add_bool_arg(struct action *action, char *name, bool value);
+struct wlr_box *action_add_box_arg(struct action *action, char *name);
 void actions_run(struct view *activator, struct server *server,
 	struct wl_list *actions, uint32_t resize_edges);
+void action_list_free(struct wl_list *action_list);
 
-#endif
+#endif // __LABWC_ACTION_H
```
johanmalm, line 49: Prefer `/* __LABWC_ACTION_H */` style comments

include/config/keybind.h
```diff
@@ -11,8 +11,8 @@ struct keybind {
 	uint32_t modifiers;
 	xkb_keysym_t *keysyms;
 	size_t keysyms_len;
-	struct wl_list actions;
-	struct wl_list link;
+	struct wl_list actions;  /* action::link */
```
johanmalm, line 14: We ought to stick with either `::` or `.`
I think I saw a change in wlroots recently moving from `::` to `.` which I guess more C like. I don't really mind which, but consistency is good. (The `.` was used in action.h above)

johanmalm, line 14: I think that for `wl_list actions` is obvious what it's a list of.... a list of actions.
So I think the only `struct wl_list` entries where it's useful to have a comment is the `struct wl_list link` because it's not obvious what list is using that link... so the line below `/* rcxml::keybinds */` is where it's useful.
Do you agree?

src/action.c
```diff
@@ -200,11 +266,15 @@ actions_run(struct view *activator, struct server *server,
 			osd_update(server);
 			break;
 		case ACTION_TYPE_RECONFIGURE:
-			/* Should be changed to signal() */
+			/* Could likely be changed to kill(0, SIGHUP) */
```
johanmalm, line 269: Oh nice :smile:

src/action.c
```diff
@@ -56,6 +60,52 @@ const char *action_names[] = {
...
+
+static struct wlr_box *
+action_box_from_arg(struct action_arg *arg)
+{
+	assert(arg->type == LAB_ACTION_ARG_BOX);
+	return &((struct action_arg_box *)arg)->value;
+}
+
+static struct action_arg *
+action_get_first_arg(struct action *action)
```
johanmalm, line 99: In terms of the general design, what do you think of the following???

```
struct action {
	char *name;
	struct wl_list arguments;

	struct wl_list link; /* keybind.actions, etc */
}

struct action_arg {
	enum action_arg_type type;
	char *key;	/* e.g. "width", "direction" */
	union {
		char *value;
		int value;
		etc...
	}
	struct wl_list link; /* action.arguments */
}
```

It's possibly a bit simpler and behave more like a dictionary (although implemented as a `wl_list`).

The argument key gives the option of a helper in action.c to do things like `width = arg(action_arg_list, "width")`
It avoids functions like `get_first_argument`

Just a thought :smile:

johanmalm (commented):
I like it in principle. Mostly minor comments, but one change in design proposal (just structuring the structs slightly differently)

include/action.h
Consolatis: Will remove it completely then

include/action.h
Consolatis: will change to `action_arg_add_XX()`

include/action.h
Consolatis: woops :)

include/config/keybind.h
Consolatis: woops. Saw the wlroots change as well but for some reason still used `::` here.

Consolatis: If i think about being new to the codebase and seeing `wl_list actions` I'd also assume there are some actions in that list. But I wouldn't necessarily know that there is a `struct action {` I could grep for in `include`.
As comments are free I'd prefer to have `/* struct action.link */` in there.

src/action.c
Consolatis: I like this better than the inheritance thing. Need to think a bit about this.

Re `width = arg(action_arg_list, "width")`:
How would the function signature for `arg()` look like in this case? At some point it would need to cast the result to one of the union members?

include/action.h
Consolatis: I don't know if we actually need that kind of introspection outside of `action.c`.
The idea was that nobody else should care about the internals of `action.c`. The outside API is either string based (`action_create()`) or `struct action *` based so the enum isn't ever used anywhere other than in `action.c` itself.

The action names are internally translated into the enum in `action_create()` and `actions_run()` gets a *wl_list of action structs as argument.
Maybe we could start using the same notation as wlroots in just putting  a `/* private */` on top of struct members like this?

src/action.c
Consolatis: Re union vs inheritance + casting, gave the union a try and ended up with this:
```c
struct action_arg {
        const char *key;      /* May be NULL if there is just one arg */
        struct wl_list link;  /* struct action.args */

        /* Private */
        uint32_t type;        /* enum action_arg_type */
        union {
                char *val_str;
                int val_int;
                double val_double;
                bool val_bool;
                struct wlr_box val_box;
        };
};
```
It makes the internal `action_x_from_arg()` function a bit easier to read but IMHO also
- exposes internal members to the outside
- make the struct look ugly

Are there technical reasons to prefer one over the other?

I can only think of a positive for the inheritance + casting approach:
- it only needs as much additional storage as required be the relevant argument instead of the max size of any argument

include/config/keybind.h
johanmalm: Yes, good call. Simple thing to help readability.

Consolatis:
I think I addressed everything, if I missed something or there is something else which should be changed / improved please say so. Also removed the not (yet?) required variants for `int`, `bool`, `double` and `wlr_box` for now.

Issue #386, opened 2022-06-09
Provide SnapToEdge with half width/height


heroin-moose:
Currently snap to edge uses all vertical/horizontal space depending on the snap direction.

For example, SnapToEdge (left) takes both `Upper left` and `Bottom left`:

```
+----------------------------+
|  Upper left |  Upper right |
+-------------+--------------+
| Bottom left | Bottom right |
+----------------------------+
```

On a relatively large screen (like 5k 27"), a whole lot of space seems wasted, so it would be nice to have SnapToEdgeHalved (poor naming, just an example) that fits the window in only one quadrant instead of two.

Consolatis:
Somewhat related to
- #122

The question here is if we want to
- add a new specialized command like you proposed (`SnapToCorner`) and thus limit it to the 4 corners
- somehow let the user define a layout (which really gets tricky to get right) and then have something like `SnapToRegion` with a user defined region as argument
- modify `SnapToEdge` to additionally allow `top-left`, `top-right`, `bottom-left` and `bottom-right` as argument + possibly create a new additional alias for that command so it actually matches the feature set

Or maybe [there is already](http://openbox.org/wiki/Help:Actions) something covering one or both of those use cases with a nice API we could reimplement.

The other question is if one of these variants would be in-scope of labwc.
Let's see what @johanmalm  has to say about that.

johanmalm:
To start with, I would like this feature too.

As a general principle, we should try to implement small (atomic?) actions and we ought to focus on Cat A and Cat B items in the [Requirements Document](https://github.com/labwc/labwc-acceptance-criteria/blob/main/Acceptance-criteria.md) as a starting point (e.g. 5.22 - 5.28)

Looking at [MoveResizeTo](http://openbox.org/wiki/Help:Actions#MoveResizeTo) it does take fractions and percentages as `<width>` and `<height>` arguments.

`<SnapToEdge>` is a compond-action (if that's a word). For example `super+right-arrow` sets width=50%, moves to the edge and if already snapped-to that edge, it moves over to the next output :smile: 
Now don't get me wrong, I really, really like this feature (probably because I was used to it from Gnome + Windows), but from a design and development perspective it was the wrong time to add it (in my opinion). We should have implemented simpler actions which could be combined to perform the same thing, because that way we get much more flexibility to turn it into other things without complicating the code. So, the reason for sharing all that is that I much prefer to not extend SnapToEdge, but rather go back to basics and really design the right small actions.

[...]

PR #388, opened 2022-06-10
[wip] Add MoveResizeTo subset

Consolatis:
Loosely based on http://openbox.org/wiki/Help:Actions#MoveResizeTo

Only allows to use percent values for x, y, width and height attributes.
Does not support `monitor` argument.

Example config:
```xml
<!-- Resize via W-Numpad -->
<keybind key="W-KP_1">
    <action name="MoveResizeTo" x="0%" y="50%" height="50%" width="50%" />
</keybind>
<keybind key="W-KP_2">
    <action name="MoveResizeTo" x="0%" y="50%" height="50%" width="100%" />
</keybind>
<keybind key="W-KP_3">
    <action name="MoveResizeTo" x="50%" y="50%" height="50%" width="50%" />
</keybind>
<keybind key="W-KP_4">
    <action name="MoveResizeTo" x="0%" y="0%" height="100%" width="50%" />
</keybind>
<keybind key="W-KP_5">
    <action name="MoveResizeTo" />
</keybind>
<keybind key="W-KP_6">
    <action name="MoveResizeTo" x="50%" y="0%" height="100%" width="50%" />
</keybind>
<keybind key="W-KP_7">
    <action name="MoveResizeTo" x="0%" y="0%" height="50%" width="50%" />
</keybind>
<keybind key="W-KP_8">
    <action name="MoveResizeTo" x="0%" y="0%" height="50%" width="100%" />
</keybind>
<keybind key="W-KP_9">
    <action name="MoveResizeTo" x="50%" y="0%" height="50%" width="50%" />
</keybind>
```

Partly fixes
- #386

PR is based on
- #387


src/view.c
```diff
@@ -611,6 +612,41 @@ view_get_edge_snap_box(struct view *view, struct output *output,
 	return dst;
 }
 
+/* Move and resize view based on percent output local values in rel_box */
+void
+view_rel_move_resize_to(struct view *view, struct wlr_box *rel_box)
```
johanmalm, line 617: `rel` and `rel_box` misled me here because I was thinking relative vs absolute. Other openbox actions using `relative` in that latter sense, but it'd be good to think of another term.

src/view.c
```diff
@@ -611,6 +612,41 @@ view_get_edge_snap_box(struct view *view, struct output *output,
...
+
+	struct output *output = view_output(view);
+	if (!output) {
+		return;
+	}
+	struct wlr_box usable = output_usable_area_in_layout_coords(output);
+
+	/* Calculate new size and position */
+	struct wlr_box new_geo = { view->x, view->y, view->w, view->h };
+	if (rel_box->x != INT_MIN) {
```
johanmalm, line 630: openbox spec doesn't allow for `<x>` and `<y>` expressed as percentages. Is it actually useful to do that??

src/view.c
johanmalm, line 617: How about re-writing this function to be `absolute`, but prior to calling it do a `resolve_size()` to translate width/height expressed as percentages or fractions into actual pixels.

A `resolve_size()` could honour `rc.gap`

```
width_ratio = 0.5;  /* calculate by resolve_size() */
nr_gaps = (1/width_ratio + 1)
width = (output->width - nr_gaps) * width_ratio;
```
(Might have got that slightly wrong... but hopefully you know where I'm going)

src/view.c
```diff
@@ -611,6 +612,41 @@ view_get_edge_snap_box(struct view *view, struct output *output,
...
+		new_geo.x = usable.width * rel_box->x / 100;
+		new_geo.x += usable.x + view->margin.left;
+	}
+	if (rel_box->y != INT_MIN) {
```
johanmalm, line 634: It would be nice honour `rc.gap` here too.

src/view.c
Consolatis: Hm.. right. It \*is\* relative but relative to the primary output of the view.

src/view.c
Consolatis: I think it is an absolute requirement for our use-case. How else would you have some bindings that allow to freely tile independent of any knowledge of the current output or its dimensions?
With this a user could easily split any output into a `2x2`, `3x3` tiling state or instead do something like this:
```
.----------.--------------------.----------.
|          |                    |          |
|          |                    |          |
|          |                    |----------|
|          |                    |          |
|          |                    |          |
|          |                    |----------|
|          |                    |          |
|          |                    |          |
`----------^--------------------^----------´
```

src/view.c
Consolatis: hm.. basically this is already something like that. At the end we call the absolute `view_move_resize()`.
About the gaps I am not sure they would fit here at all.. If a users wants gaps between the tiled windows they could offset the `x` and `y` variants by a percent.

Currently this prototype only supports integers for the percentages (as we internally use the wlr_box for storing) but I guess it could be replaced by `struct wlr_fbox` instead.

heroin-moose: I wanna gaps! Actually, it would be nice to (optionally) respect gaps in maximized mode too (however, with MoveResizeTo I can do it myself of course).

PR #389, opened 2022-06-11
correctly scale window switcher

bi4k8:
We should do our drawing in the "abstract" coordinate space (relative to unscaled w/h), but use output-scaled coordinates when allocating the buffer (`buffer_create_cairo` forwards its width/height directly to `wlr_buffer_init` without adjusting by `scale`). We also call `wlr_scene_buffer_set_dest_size` to inform wlroots that the buffer is scaled.

### DIFF
--- a/src/osd.c
+++ b/src/osd.c
@@ -127,13 +127,15 @@ osd_update(struct server *server)
 	wl_list_for_each(output, &server->outputs, link) {
 		destroy_osd_nodes(output);
 		float scale = output->wlr_output->scale;
-		int w = (OSD_ITEM_WIDTH + (2 * OSD_BORDER_WIDTH)) * scale;
-		int h = get_osd_height(&server->views) * scale;
+		int w = (OSD_ITEM_WIDTH + (2 * OSD_BORDER_WIDTH));
+		int h = get_osd_height(&server->views);
+		int scaled_w = w * scale;
+		int scaled_h = h * scale;
 
 		if (output->osd_buffer) {
 			wlr_buffer_drop(&output->osd_buffer->base);
 		}
-		output->osd_buffer = buffer_create_cairo(w, h, scale, true);
+		output->osd_buffer = buffer_create_cairo(scaled_w, scaled_h, scale, true);
 
 		cairo_t *cairo = output->osd_buffer->cairo;
 		cairo_surface_t *surf = cairo_get_target(cairo);
@@ -225,6 +227,7 @@ osd_update(struct server *server)
 
 		struct wlr_scene_buffer *scene_buffer = wlr_scene_buffer_create(
 			output->osd_tree, &output->osd_buffer->base);
+		wlr_scene_buffer_set_dest_size(scene_buffer, w, h);
 
 		/* Center OSD */
 		struct wlr_box output_box;


Consolatis:
Thanks, looks good to me and testing a scale of 2 and 0.5 was fine as well.
Definitely an improvement over the current status.

To be honest, I am not 100% certain that
- drawing using the unscaled dimensions
- using the scaled dimensions for `cairo_image_surface_create()`
- `cairo_surface_set_device_scale()`
- `wlr_scene_buffer_set_dest_size()` with the unscaled dimensions

is the correct thing to do but the result seems scaled correctly and doesn't look blurry to me.
Sway seems to [add a scale attribute](https://github.com/swaywm/sway/blob/42b61ab27a6077e2089246513d6f66d20f536dd8/common/pango.c#L75) when creating fonts with pango but I didn't yet look further into it.

If we instead just ignore scaling completely (like we do within the SSD), wlroots will take care of it and automatically upscale as necessary but the result may look blurry for cases like fonts.

bi4k8:
- Drawing using unscaled dimensions is correct because Cairo transparently adapts all coordinated by the device scale, which is set by `buffer_create_cairo`.
- `cairo_image_surface_create` has no internal notion of scale, so it needs the scaled dimensions to allocate a big enough buffer.
- Using `cairo_surface_set_device_scale` to set the device scale lets us render in unscaled coordinates. It appears that Sway instead [avoids ever setting device scale](https://github.com/swaywm/sway/search?q=cairo_surface_set_device_scale) and instead manually renders upscaled content in a scaled coordinate space, which is also consistent, but further from what our code was doing, and seems more complex--I don't know of an advantage to this approach. The only possibility that comes to mind is that it could be useful if we want to render in output-space for some reason (custom AA or "as thin as possible" lines).
- Without `wlr_scene_buffer_set_dest_size`, the pixels in the buffer are interpreted in the scene graph as being rendered without scale-awareness, causing OSD to be 2x as large as it should when scale=2, etc. So I think we're using it properly based on its effect.

The only thing I'm curious about is whether it would be clearer for our `buffer_create_cairo` function to accept unscaled width/height, given that it does know the scale via its explicit `scale` argument. This would make consumers that construct a buffer and then draw into it (like `osd.c`) more homogeneous.

Consolatis:
Great explanation!

Regarding this PR I'll wait for @johanmalm to look over the diff because its not really my area of expertise.

> The only thing I'm curious about is whether it would be clearer for our buffer_create_cairo function to accept unscaled width/height, given that it does know the scale via its explicit scale argument. This would make consumers that construct a buffer and then draw into it (like osd.c) more homogeneous.

I like this idea, might be interesting for another PR.

So if I got that right we'd just use unscaled dimensions everywhere and let `buffer_create_cairo()` deal with the scale internally (allocation + set_device_scale). Only thing that must be done manually by the caller would be to provide the proper scale argument when creating the buffer + `wlr_scene_buffer_set_dest_size(unscaled_width, unscaled_height)` after attaching the `wlr_buffer` to a `wlr_scene_buffer`.

johanmalm:
@bi4k8 grateful for help in navigating this.

Just trying to think this through... 

Let's take a case of a buffer at 100x100 which at output->scale==1 we want to render as 50x50 and at output->scale==2 we want to render as 100x100

In osd.c as we iterate over outputs, if we get to an output with scale==2, we still set `wlr_scene_buffer_set_dest_size(scene_buffer, 50, 50);`? so that scene_buffer->dst_height/width will be 50.

At https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/wlr_scene.c#L520
we create dst_box based on buffer->dst_height/width (still 50x50).

At https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/wlr_scene.c#L884
we then use output->scale to scale dst_box, so dst_box->width==100 and dst_box->height==100

So that at https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/wlr_scene.c#L915 dst_box is 100x100.

The important thing to make it 100x100 to start with so that src_box is 100x100, so that we never have to scale up (not relevant to the osd discussion, I suppose, but will be for server-side-deco and the root-menu).

Makes sense to me.

Are fonts clear on bigger output->scale or do we need to set some pango-attributes?
https://github.com/swaywm/sway/blob/1c69d0e72fe868d0fa2f6d7d9f9b4b3d7a3f78b4/common/pango.c#L75-L79

@Consolatis Suggest we proper-merge this one so we have a reference to the discussion.

bi4k8:
Fonts are clear/crisp in my testing with an output of scale 2. Setting the scale pango attribute is an alternative way to communicate scale to pango (vs. setting the surface device scale in Cairo), and Sway has to use it because it never calls `cairo_surface_set_device_scale`.

Consolatis:
> The important thing to make it 100x100 to start with so that src_box is 100x100, so that we never have to scale up (not relevant to the osd discussion, I suppose, but will be for server-side-deco and the root-menu).

`struct wlr_scene_buffer` actually has a output enter / output leave event. I have a prototype somewhere which dynamically regenerates the title label based on the biggest output scale that node is in. But a proper generic solution (so it can be used for menu and SSD title) would likely consists of some `lab_font_buffer_node` that has a pointer to the `wlr_scene_buffer` and subscribes to output enter / leave events and then regenerates the internal `wlr_buffer` when the max scale changes.

@johanmalm completely fine with `merging` this PR instead of `rebase and merge` so we have a reference to this discussion attached the commit.

Issue #348, opened 2022-05-27
Output scaling of SSD buttons / title


Consolatis:
I don't understand the scene-graph scaling:
![scale_5](https://user-images.githubusercontent.com/35009135/170690751-c4d9c182-4e5d-4096-aa64-4c6a90728c5f.png)

`wlr_scene_rect` gets scaled just fine, so the buttons (and their hover overlays) seem correct.
`wlr_scene_buffer` however.. the SSD title doesn't seem to be touched \*at all\* but the icons on the buttons seem as if they were, for example, replicated to the right (especially visible on the window-menu one on the left).
The rounded corners however, are seemingly scaled correctly as well (which are also `wlr_scene_buffer`).

For the record: titlebar scene node layout
```
                                X     Y        Node
                             ----  ----  ----------------
+-- titlebar.active           290    98  [0xaaaaab42a970]
   +-- rect                   316    98  [0xaaaaab43d110]  title bar between the buttons
      +-- buffer              626   102  [0xaaaaab43eca0]      title
   +-- buffer                 289    97  [0xaaaaab43d1e0]  rounded corner left
      +-- buffer              300   107  [0xaaaaab43d330]      window-menu icon
      +-- rect                290    98  [0xaaaaab43d480]      hover overlay
   +-- rect                   912    98  [0xaaaaab43d550]  usual button
      +-- buffer              921   106  [0xaaaaab43d620]      minimize icon
      +-- rect                912    98  [0xaaaaab43d770]      hover overlay
   +-- rect                   938    98  [0xaaaaab43d840]  usual button
      +-- buffer              947   106  [0xaaaaab43d910]      maximize icon
      +-- rect                938    98  [0xaaaaab43da60]      hover overlay
   +-- buffer                 964    97  [0xaaaaab43db30]  rounded corner right
      +-- buffer              973   106  [0xaaaaab43dc80]      close icon
      +-- rect                964    98  [0xaaaaab43ddd0]      hover overlay
```

Original issue: #289

johanmalm:
Not had a chance to look, but this might be a useful source to compare with
https://github.com/Nefsen402/sway/blob/scene-graph/sway/sway_text_buffer.c

The have a `struct text_buffer_output` and `handle_output_enter()`
I think fundamentally our buffer points to a `cairo_t` which surface should be made bigger the greater the output scale

johanmalm:
On phone, so can’t compile/test/experiment but found this:

[...]

PR #390, opened 2022-06-12
Add autoscaling font buffer

Consolatis:
This adds an auto scaling font buffer to labwc and converts menu and ssd_titlebar to use it.
I tried to put everything into small and digestible commits.

First three commits could likely be split out of this PR.

auto_scale_buffer

I am mostly satisfied with the auto scale buffer state for now, I'd just need to remove the debug logs.

Something that might be interesting for the future is to generalize this some more and instead of having an scale adapting font buffer we'd have an abstract scale adapting 'thing' that gets a config struct when initialized containing callbacks that will get called whenever a new buffer needs to be rendered. The 'thing'  would then only be responsible for setting the appropriate scaled buffer on output_enter/leave events and request new buffers if the node had been moved to an unknown (or already evicted) scale. The first user of that 'thing' would then be a `autoscaling_font_buffer`.
Somewhat similar in design to `wlr_buffer` <--> `lab_data_buffer`.

Edit: did that ^. [Abstract](https://github.com/labwc/labwc/blob/dfd33dc721754629e8fb1e123b4421163861e36c/include/common/scaled_scene_buffer.h) & [Font buffer](https://github.com/labwc/labwc/blob/dfd33dc721754629e8fb1e123b4421163861e36c/include/common/scaled_font_buffer.h)

Related to
- #348
- #389

### DIFF
--- a/include/common/scaled_font_buffer.h
+++ b/include/common/scaled_font_buffer.h
@@ -1,0 +1,49 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LAB_COMMON_SCALED_FONT_BUFFER_H
+#define __LAB_COMMON_SCALED_FONT_BUFFER_H
+
+struct font;
+struct wlr_scene_tree;
+struct wlr_scene_buffer;
+struct scaled_scene_buffere;
+
+struct scaled_font_buffer {
+	struct wlr_scene_buffer *scene_buffer;
+	int width;   /* unscaled, read only */
+	int height;  /* unscaled, read only */
+
+	/* Private */
+	char *text;
+	int max_width;
+	float color[4];
+	struct font font;
+	struct scaled_scene_buffer *scaled_buffer;
+};
+
+/**
+ * Create an auto scaling font buffer, providing a wlr_scene_buffer node for
+ * display. It gets destroyed automatically when the backing scaled_scene_buffer
+ * is being destoyed which in turn happens automatically when the backing
+ * wlr_scene_buffer (or one of its parents) is being destroyed.
+ *
+ * To actually show some text, scaled_font_buffer_update() has to be called.
+ *
+ */
+struct scaled_font_buffer *scaled_font_buffer_create(struct wlr_scene_tree *parent);
+
+/**
+ * Update an existing auto scaling font buffer.
+ *
+ * No steps are taken to detect if its actually required to render a new buffer.
+ * This should be done by the caller to prevent useless recreation of the same
+ * buffer in case nothing actually changed.
+ *
+ * Some basic checks could be something like
+ * - truncated = buffer->width == max_width
+ * - text_changed = strcmp(old_text, new_text)
+ * - font and color the same
+ */
+void scaled_font_buffer_update(struct scaled_font_buffer *self, const char *text,
+	int max_width, struct font *font, float *color);
+
+#endif /* __LAB_COMMON_SCALED_FONT_BUFFER_H */
--- a/include/common/scaled_scene_buffer.h
+++ b/include/common/scaled_scene_buffer.h
@@ -1,0 +1,70 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LAB_COMMON_SCALED_SCENE_BUFFER_H
+#define __LAB_COMMON_SCALED_SCENE_BUFFER_H
+
+#define LAB_SCALED_BUFFER_MAX_CACHE 2
+
+struct wl_list;
+struct wlr_buffer;
+struct wl_listener;
+struct wlr_scene_tree;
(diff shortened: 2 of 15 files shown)


src/common/font.c
```diff
@@ -73,7 +73,7 @@ font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 		rect.width = max_width;
 	}
 	/* TODO: scale */
```
Consolatis, line 75: remove

src/common/scaled_buffer.c
```diff
@@ -0,0 +1,197 @@
...
+	if (!self->scene_buffer) {
+		wlr_log(WLR_ERROR, "Failed to create scene buffer");
+		free(self);
+		return NULL;
+	}
+
+	self->active_scale = 1;
+	wl_list_init(&self->cache);
+
+	/* Listen to output enter / leave events so we can update the scale */
```
Consolatis, line 147: get notified about scale changes*

src/common/scaled_buffer.c
```diff
@@ -0,0 +1,197 @@
...
+	if (primary && primary->output->scale != self->active_scale) {
+		update_buffer(self, primary->output->scale);
+	}
+}
+
+/* Public API */
+struct scaled_buffer *
+scaled_buffer_create(struct wlr_scene_tree *parent)
+{
+	struct scaled_buffer *self = calloc(1, sizeof(*self));
```
Consolatis, line 132: `assert(parent)`

src/common/scaled_buffer.c
```diff
@@ -0,0 +1,197 @@
...
+		const char *text, int max_width, struct font *font, float *color)
+{
+	assert(self);
+	assert(text);
+	assert(font);
+	assert(color);
+
+	wlr_log(WLR_ERROR, "Update for '%s'", text);
+
+	/* Clean up old state */
```
Consolatis, line 171: internal state*

Consolatis:
@Nefsen402 I am aware of that. Still thinking about how to support output scale changes without the requirement of syncing a list of all outputs + listening to their scale changes for every single buffer node.

include/common/scaled_font_buffer.h
```diff
@@ -0,0 +1,51 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LAB_COMMON_SCALED_FONT_BUFFER_H
+#define __LAB_COMMON_SCALED_FONT_BUFFER_H
+
+#define LAB_SCALED_BUFFER_MAX_CACHE 2
```
Consolatis, line 5: remove

src/common/scaled_scene_buffer.c
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	/* Limit amount of cached buffers */
+	if (wl_list_length(&self->cache) == LAB_SCALED_BUFFER_MAX_CACHE) {
+		/* Remove last element */
+		cache_entry = wl_container_of(self->cache.prev, cache_entry, link);
+		wlr_log(WLR_ERROR, "[%p] Removing oldest cache entry for scale %f",
+			&self->scene_buffer->node, cache_entry->scale);
+		_cache_entry_destroy(cache_entry);
+	}
```
Consolatis, line 49: ^ reuse cache entry instead of destroying. else `cache_entry = calloc(1, sizeof(*cache_entry));`
+ remove calloc of cache_entry below

src/buffer.c
```diff
@@ -92,10 +92,15 @@ buffer_create_cairo(uint32_t width, uint32_t height, float scale,
...
+	buffer->unscaled_height = height;
+	width *= scale;
+	height *= scale;
+
 	wlr_buffer_init(&buffer->base, &data_buffer_impl, width, height);
 
-	cairo_surface_t *surf =
-		cairo_image_surface_create(CAIRO_FORMAT_ARGB32, width, height);
+	cairo_surface_t *surf = cairo_image_surface_create(
+		CAIRO_FORMAT_ARGB32, width, height);
```
Consolatis, line 103: commit noise, restore

include/common/scaled_scene_buffer.h
```diff
@@ -0,0 +1,69 @@
...
+struct wl_list;
+struct wlr_buffer;
+struct wl_listener;
+struct wlr_scene_tree;
+struct lab_data_buffer;
+struct scaled_scene_buffer;
+
+struct scaled_scene_buffer_impl {
+	/* Return a new buffer optimized for the new scale */
+	struct lab_data_buffer *(*create_buffer)(struct scaled_scene_buffer *, double);
```
johanmalm, line 16: Consider adding the variable names here (i.e. `self` and `scale`). Not a requirement in C, but can make it clearer.

PR #394, opened 2022-06-14
action.c: use kill() instead of spawning killall

### LINKED ISSUE #393: Reconfigure do nothing
basically if i  change theme editing rc file save and hit reconfigure nothing happens. If i exit and restart labwc the theme is updated.
not sure if this is expeted, but i saw a video where Reconfigure basically update this.

johanmalm:
Also update documentation to say `killall -s SIGHUP labwc` to make
it work on Void Linux.

Fix #393


Consolatis:
Nice. We should likely backport this to v0.5 as well

PR #400, opened 2022-06-18
workspaces: Allow switching/sending to the last used workspace

heroin-moose:
Actions GoToDesktop and SendToDesktop now support the new direction
called "last" that corresponds to the last used workspace (see Openbox
help[1] for reference).

[1]: http://openbox.org/wiki/Help:Actions#GoToDesktop


Consolatis:
Thanks!

PR #401, opened 2022-06-18
Restore original window geometry after tiling

### LINKED ISSUE #391: Restore geometry when moving window with the mouse away from window snapping
Similar to what already happens with maximized windows, but also for SnapToEdge and ideally for whatever gets implemented in the future (https://github.com/labwc/labwc/issues/386).

Consolatis:
This could need some testing, especially in combination with (un)maximizing or fullscreen (fullscreen not actually tested \*at all\*)

Fixes #391

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -306,6 +306,7 @@ struct view {
 	bool been_mapped;
 	bool minimized;
 	bool maximized;
+	bool is_tiled;
 	struct wlr_output *fullscreen;
 
 	/* geometry of the wlr_surface contained within the view */
@@ -313,6 +314,9 @@ struct view {
 
 	/* geometry before maximize */
 	struct wlr_box unmaximized_geometry;
+
+	/* geometry before tiling */
+	struct wlr_box untiled_geometry;
 
 	/*
 	 * margin refers to the space between the extremities of the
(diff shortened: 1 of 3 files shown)


Consolatis:
> Actually I did find two things. After maximizing, tiling and then dragging Nautilus the border is gone, it doesn't restore properly from when it hides while maximized.

Hm.. right, it appears we are indeed hiding borders when maximized. Good catch, will fix.

> Another with programs that start out maximized, when tiling and then dragging they are sized very small instead of what seems to be the default size.

Interesting, will investigate.

Consolatis:
Replaced by #427

Consolatis:
> Actually I did find two things. After maximizing, tiling and then dragging Nautilus the border is gone, it doesn't restore properly from when it hides while maximized.

This should be fixed with latest master as we now do a proper un-maximize before tiling.

> Another with programs that start out maximized, when tiling and then dragging they are sized very small instead of what seems to be the default size. [Screenshot](https://user-images.githubusercontent.com/4444593/174479084-b93056a6-0b7f-4ccd-a132-578de3ba7bda.jpg)

Could you verify if this still happens on latest master? It should default to the fallback geometry of (640x480)

Issue #406, opened 2022-06-19
A-Tab OSD - prevent long window titles from overflowing selection box


Flrian:
![window_title](https://user-images.githubusercontent.com/4444593/174493707-fa794c57-73a2-4431-b4c1-df23e50ce6fd.png)

johanmalm:
Suggest set_ellipsie
https://docs.gtk.org/Pango/method.Layout.set_ellipsize.html

johanmalm:
Sorry. We already do set_ellipsize... bad steer from me

I think this fixes it:

```
diff --git a/src/osd.c b/src/osd.c
index 646feb9..496e7e1 100644
--- a/src/osd.c
+++ b/src/osd.c
@@ -189,7 +189,8 @@ osd_update(struct server *server)
 		/* text */
 		set_source(cairo, theme->osd_label_text_color);
 		PangoLayout *layout = pango_cairo_create_layout(cairo);
-		pango_layout_set_width(layout, w * PANGO_SCALE);
+		pango_layout_set_width(layout,
+			(OSD_ITEM_WIDTH - 2 * OSD_ITEM_PADDING) * PANGO_SCALE);
 		pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
 
 		struct font font = {
```

Fancy testing it and submitting a PR?

Consolatis:
Nice. Another thing I see there is the workspace name is not correctly centered on the y axis.
Guess its because it uses bold but the height calculation uses regular text instead.

Consolatis:
heh, how did you manage to do that?

Consolatis:
oh nice one. That might come in handy for testing other stuff as well. I wonder if we should just limit the app_id to 20 chars or something

PR #407, opened 2022-06-19
src/osd.c: fix max text width

Flrian:
Making sure long window titles don't overflow the selection box

Closes issue #406

Written-by: @johanmalm


Consolatis:
LGTM

PR #414, opened 2022-06-22
menu: support <separator />

johanmalm:


johanmalm:
I've updated documentation or added theme variables yet, but just wanted to get some early quick reviews if poss.

src/menu/menu.c
```diff
@@ -148,21 +150,76 @@ item_create(struct menu *menu, const char *text)
...
+		y += item->height;
+	}
 	wlr_scene_node_set_position(&menuitem->tree->node, 0, y);
 
 	/* Hide selected state */
 	wlr_scene_node_set_enabled(&menuitem->selected.tree->node, false);
 
 	/* Update menu extents */
-	menu->size.height = (item_count + 1) * menu->item_height;
+	menu->size.height = y + menuitem->height;
```
Consolatis, line 164: similar, couldn't we just do a `menu->size.height += menu->item_height`

src/menu/menu.c
```diff
@@ -148,21 +150,76 @@ item_create(struct menu *menu, const char *text)
...
 	return menuitem;
 }
 
+static struct menuitem *
+separator_create(struct menu *menu, const char *label)
+{
+	int menu_separator_width = 1;
+	int menu_separator_padding_width = 10;
+	int menu_separator_padding_height = 2;
+	float menu_separator_color[4] = { 0.5, 0.5, 0.5, 1.0 };
```
Consolatis, line 177: Does Openbox provide theme variables for all of these settings?

src/menu/menu.c
```diff
@@ -148,21 +150,76 @@ item_create(struct menu *menu, const char *text)
...
+	wl_list_for_each_reverse(item, &menu->menuitems, link) {
+		y += item->height;
+	}
+	wlr_scene_node_set_position(&menuitem->tree->node, 0, y);
+
+	/* Vertically center-align separator line */
+	wlr_scene_node_set_position(menuitem->normal.text,
+		menu_separator_padding_width, menu_separator_padding_height);
+
+	menu->size.height = y + menuitem->height;
```
Consolatis, line 217: and similar to above just a `menu->size.height += menuitem->height;` ?

src/menu/menu.c
```diff
@@ -645,6 +717,9 @@ menu_process_cursor_motion(struct wlr_scene_node *node)
 	assert(node && node->data);
 	struct menuitem *item = node_menuitem_from_node(node);
 
+	if (!item || !item->selectable) {
+		return;
+	}
```
Consolatis, line 722: Alternative: We could also use a new node descriptor for this instead of `item->selectable`. I think I'd prefer your current solution with the `selectable` bool though.

Consolatis, line 722: IMHO the check for `!item` isn't necessary here because `node_menuitem_from_node()` already asserts if the type is wrong and the assert above asserts that there is in fact some kind of descriptor set. We could turn it into an additional `assert(item)` though.

src/menu/menu.c
johanmalm: Yep, same named, just dot instead of underscore…. two final cat A items

src/menu/menu.c
```diff
@@ -148,21 +150,76 @@ item_create(struct menu *menu, const char *text)
...
 
 	/* Position the item in relation to its menu */
-	int item_count = wl_list_length(&menu->menuitems);
-	y = item_count * menu->item_height;
+	y = 0;
+	struct menuitem *item;
+	wl_list_for_each_reverse(item, &menu->menuitems, link) {
+		y += item->height;
+	}
 	wlr_scene_node_set_position(&menuitem->tree->node, 0, y);
```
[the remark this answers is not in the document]
johanmalm: Yes. That’s cleaner. Will sort tomorrow.

src/menu/menu.c
johanmalm: 👍

src/menu/menu.c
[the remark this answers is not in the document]
Consolatis: Just recognized: `menu_hide_submenu()` needs that loop for recalculating the new offset + menu size though

Consolatis:
Looks like your IRC bouncer provider decided to have some maintenance or something.

> \<malm\> Ready to merge unless you can see any problems with it

Changes look good, your new hide / reposition logic is definitely way easier to grasp than it was before, guess I sometimes tend to unnecessarily over-complicate things.

If you are ok with it I'd like to delay the merge to tomorrow so I can give it a proper test ride.

johanmalm:
Yes, IRC crashed :smile:
No rush, merge when ready... and do just say if you think it needs further work.

PR #421, opened 2022-06-28
Resolve focus change theme color transition error

yuanye100:
When using the server decoration, the focus shifts to the client decorated with the client, and the theme color will not become inactive.

### DIFF
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -135,7 +135,7 @@ struct ssd_hover_state {
 /* Public SSD API */
 void ssd_create(struct view *view);
 void ssd_hide(struct view *view);
-void ssd_set_active(struct view *view);
+void ssd_set_active(struct view *view, bool activated);
 void ssd_update_title(struct view *view);
 void ssd_update_geometry(struct view *view);
 void ssd_reload(struct view *view);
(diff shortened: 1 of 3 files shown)


Consolatis:
This doesn't seem to switch the previous active view (`view->server->ssd_focused_view`) back to inactivate state.

The issue this is trying to solve is when switching between CSD (client side decoration) and SSD (server side decoration) window?

johanmalm:
@yuanye100 
Appreciate the contribution :smile:
The code looks good, but I'd appreciate help understanding the issue.
Could you describe what steps you take to change focus without the server-side-deco color changing.

yuanye100:
Open a CSD client and an SSD client. When the focus is on the SSD client, the decoration color is active. However, when the focus is transferred to the CSD client, the decoration color of the SSD client is not changed to inactive. To solve this problem, I will change the flag recording the SSD status, and directly judge the status of the setting client, instead of judging the SSD client alone.

johanmalm:
@yuanye100 Sorry we've been slow to respond properly to this.
I get the issue now. I think #504 covers it. It's a similar solution to what you've provided, but does some additional re-factoring.

johanmalm:
Assume this is now overtaken by events (#504)?

Issue #297, opened 2022-04-16
Inconsistency between CSD/SSD reported to client, and actually rendered


dnkl:
As mentioned in https://github.com/labwc/labwc/issues/296#issuecomment-1100699231, setting `<decoration>client</decoration>` doesn't actually enable CSDs in foot, but does in Alacritty. What gives?

The difference is how these two applications use the xdg-decoraction protocol. This protocol has two methods: `set_mode()` and `unset_mode()`.

Alacritty uses `unset_mode()`, while foot uses `set_mode()`.

`unset_mode()`:

> Unset the toplevel surface decoration mode. This informs the compositor that the client doesn't prefer a particular decoration mode.

`set_mode()`:

> Set the toplevel surface decoration mode. This informs the compositor that the client prefers the provided decoration mode 
After requesting a decoration mode, the compositor will respond by emitting an xdg_surface.configure event. The client should then update its content, drawing it without decorations if the received mode is server-side decorations. The client must also acknowledge the configure when committing the new content (see xdg_surface.ack_configure).
> 
> **The compositor can decide not to use the client's mode and enforce a different mode instead.**

Key words here are **prefers** and **can enforce a different mode instead**.

What I suspect is happening is that labwc receives foot's `set_mode()` request, where foot says it **prefers** SSDs, and labwc replies with "sure ok, no problem" (i.e. foot receives `ZXDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE` in the xdg toplevel decoration `configure` event - this has been verified, see attached ` WAYLAND_DEBUG` logs).

But when rendering the foot window, labwc just looks at its `<decoration>` setting and decides to not draw any SSDs.

`WAYLAND_DEBUG` logs (`grep decoration`  - full logs are attached):

Alacritty
```
[1604551.033] wl_registry@2.global(15, "zxdg_decoration_manager_v1", 1)
[1604551.049]  -> wl_registry@2.bind(15, "zxdg_decoration_manager_v1", 1, new id [unknown]@8)
[1604551.068] wl_registry@2.global(16, "org_kde_kwin_server_decoration_manager", 1)
[1604607.326]  -> zxdg_decoration_manager_v1@8.get_toplevel_decoration(new id zxdg_toplevel_decoration_v1@22, xdg_toplevel@19)
[1604607.352]  -> zxdg_toplevel_decoration_v1@22.unset_mode()
[1604615.769] zxdg_toplevel_decoration_v1@22.configure(1)
[...]

PR #422, opened 2022-06-28
src/ssd/ssd.c: Keep view->margin in sync when toggling decorations

### LINKED ISSUE #409: SnapToEdge does not consider turned off window decorations
There is just a gap where the titlebar and border would be.
[Screenshot](https://user-images.githubusercontent.com/4444593/174590236-87afde56-8bd4-44b7-81d3-04f6f7b32fa0.png)

Consolatis:
- Fixes #409

I'd welcome some testing to make sure this doesn't break anything else like center positioning or starting maximized.

While testing this I discovered Wayland native applications starting in fullscreen mode like `foot -F` actually keep their SSD for unknown reasons, that appears to be the case even without this PR though so its unrelated.

### DIFF
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -134,7 +134,6 @@ struct ssd_hover_state {
 
 /* Public SSD API */
 void ssd_create(struct view *view);
-void ssd_hide(struct view *view);
 void ssd_set_active(struct view *view);
 void ssd_update_title(struct view *view);
 void ssd_update_geometry(struct view *view);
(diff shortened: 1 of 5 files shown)


johanmalm:
Regarding foot, see #297

johanmalm:
I've tested with CSD + SSD + outputs with different scales. Seems to work fine.

Consolatis:
Didn't detect any issues with it either so far. Feel free to merge if you are ok with it.

PR #423, opened 2022-06-28
Fix usable area calculation

### LINKED ISSUE #420: labwc renders mako notifications over the statusbar
The following screenshot illustrates the issue:

image

Instead of respecting the statusbar height (10px top margin + 25px height + 10px bottom gap from labwc), labwc renders the surface directly in the top-right corner.

For example, sway correctly (more or less) handles this and renders the surface with respect to the statusbar dimensions (but not the compositor gaps):

image

heroin-moose:
Currently if a surface with exclusive zone is created prior to regular
surfaces, the size of the exclusize zone does not affect the usable area
for regular surfaces. This for example results in notifications being
rendered over the statusbar.

This commit fixes the issue by handling the surfaces with exclusive
zones first.

Fixes #420.


Consolatis:
LGTM. Thanks :)

PR #425, opened 2022-06-30
src/layers.c: Adjust views based on usable_area changes

Consolatis:


johanmalm:
This is really good. We should have done this ages ago :smile:

Should we consider re-positioning unmaximized windows too? Quick demo of how sway does it below.

https://user-images.githubusercontent.com/6955353/176762544-62a6cb22-919f-4c42-96cb-2671da54041a.mp4

Could do as separate patch.

Consolatis:
> Should we consider re-positioning unmaximized windows too
> Could do as separate patch.

I think for `top` and `left` offsets it would be a nice feature, even though we couldn't restore the change when removing the panel without storing that information in yet another `view->delta_geometry` or something.

PR #428, opened 2022-07-02
(Partly) fix handling of client-initiated configure requests

### LINKED ISSUE #418: Undecorated XWayland windows can't be moved
Under X11, it's possible for an undecorated window to be moved by dragging it (from within the client area of the window).  The way this works is that the application handles the mouse events and simply sends "move window" requests to the window manager.  These types of windows (running in XWayland mode) currently can't be moved at all in `labwc` -- you can click and drag on them, but the windows just don't move.

Examples:
- Main window of Audacious in Winamp mode
- Detached toolbars of Qt applications (try `kdiff3`)

I don't think this is a recent regression but rather something that has never worked in `labwc` so far.  I would guess that there is some X11/NetWM message that isn't currently being handled.

I don't have time to look into a fix right now, but thought I'd document the issue in case someone else has time/inclination to work on it.

jlindgren90:
- Add missing call to wlr_scene_node_set_position() in
  unmanaged_handle_commit() -- this fixes moving unmanaged XWayland
  windows.

- Update view->pending_move_resize when we receive a configure request
  for a managed XWayland surface -- this fixes moving managed but
  undecorated XWayland windows.

- Also update view->pending_move_resize when we move a surface while a
  configure request is already pending -- this fixes a discrepancy
  between displayed and actual position for XWayland windows that try to
  set their own initial position, but then get overridden by labwc's
  positioning.

Moving undecorated XWayland windows is still really glitchy -- it appears
that an XWayland window gets sent incorrect mouse motion coordinates when
there is a pending configure request moving the window itself.

Fixes #418

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -365,6 +365,7 @@ struct view {
 struct xwayland_unmanaged {
 	struct server *server;
 	struct wlr_xwayland_surface *xwayland_surface;
+	struct wlr_scene_node *node;
 	struct wl_list link;
 	int lx, ly;
 
--- a/src/xwayland-unmanaged.c
+++ b/src/xwayland-unmanaged.c
@@ -20,6 +20,8 @@ unmanaged_handle_commit(struct wl_listener *listener, void *data)
 	struct wlr_xwayland_surface *xsurface = unmanaged->xwayland_surface;
 	unmanaged->lx = xsurface->x;
 	unmanaged->ly = xsurface->y;
+	wlr_scene_node_set_position(unmanaged->node,
+		unmanaged->lx, unmanaged->ly);
 }
 
 static void
@@ -59,10 +61,11 @@ unmanaged_handle_map(struct wl_listener *listener, void *data)
 	}
 
 	/* node will be destroyed automatically once surface is destroyed */
-	struct wlr_scene_node *node = &wlr_scene_surface_create(
+	unmanaged->node = &wlr_scene_surface_create(
 			unmanaged->server->unmanaged_tree,
 			xsurface->surface)->buffer->node;
-	wlr_scene_node_set_position(node, unmanaged->lx, unmanaged->ly);
+	wlr_scene_node_set_position(unmanaged->node,
+		unmanaged->lx, unmanaged->ly);
 }
 
 static void
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -144,6 +144,13 @@ handle_request_configure(struct wl_listener *listener, void *data)
 	int height = event->height;
 	view_adjust_size(view, &width, &height);
 
+	view->pending_move_resize.update_x = event->x != view->x;
+	view->pending_move_resize.update_y = event->y != view->y;
+	view->pending_move_resize.x = event->x;
+	view->pending_move_resize.y = event->y;
+	view->pending_move_resize.width = width;
+	view->pending_move_resize.height = height;
+
 	wlr_xwayland_surface_configure(view->xwayland_surface,
 		event->x, event->y, width, height);
 }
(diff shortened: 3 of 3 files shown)


johanmalm:
Let me start by saying a big thank you. Your commits always seem well thought through, and this looks no different :smile: Also, this bit of the code is certainly not easy to grapple with, so efforts are much appreciated.

The `unmanaged->node` is a great catch, as is the new code in `handle_request_configure()`

One nit-pik (but happy we can tidy up later): Could we consider calling configure() from handle_request_configure() to reduce duplication? Just create a wlr_box based on event.{x,y} and width/height.

Reading this through, I'm thinking that we ought to consider removing:

- `wlr_scene_node_set_position()` and `ssd_update_geometry()` from `view_move()`
- `ssd_update_geometry()` from `view_move_resize()`

...and just let commit-handlers do those things.

I've not tried it or thought it through properly though, so am just offering it up in case your have thoughts on it.

Consolatis:
Small Python QT5 test application:
```python3
#!/usr/bin/env python3

import sys
from PyQt5.QtWidgets import QApplication, QWidget

class Window(QWidget):
	def __init__(self, title, width, height):
		super().__init__()
		self.setWindowTitle(title)
		self.resize(width, height)
		self.mousePressEvent = self.handle_press
		self.mouseMoveEvent = self.handle_move
		self.show()

	def handle_press(self, event):
		self.start_pos_cursor = event.globalPos()
		self.start_pos_window = self.pos()

	def handle_move(self, event):
		delta = event.globalPos() - self.start_pos_cursor

		use_absolute = True
		if use_absolute:
			# Position based on delta to initial PressEvent
			delta += self.start_pos_window
		else:
			# Position based on delta to last MoveEvent
			delta += self.pos()
			self.start_pos_cursor = event.globalPos()

		self.move(delta)

if __name__ == '__main__':
	app = QApplication(sys.argv)
	window = Window('Move me', 640, 480)
	sys.exit(app.exec_())
```

johanmalm:
Does this help?

```
index 37cf4c0..9c51643 100644
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -151,6 +151,7 @@ handle_request_configure(struct wl_listener *listener, void *data)
 	view->pending_move_resize.width = width;
 	view->pending_move_resize.height = height;
 
+	wlr_scene_node_set_position(&view->scene_tree->node, event->x, event->y);
 	wlr_xwayland_surface_configure(view->xwayland_surface,
 		event->x, event->y, width, height);
 }
```

I can reproduce by taking the following steps:

1. Start leafpad
2. Open Help --> About
3. Move the dialog 
4. Close the dialog
5. Open it again
6. Drag it and observe a jump in position

The above patch fixes it.

johanmalm:
Nice script by the way. I've tried with that to. 
Get jumpy move without patch. Fixed with patch :smile:
Be good to understand why though.

Consolatis:
> The above patch fixes it.

Yep, fixes it for me as well.

jlindgren90:
Thanks for merging, and for the "glitchiness" fix. I have not had a chance to test it yet, but I look forward to doing so :)

> One nit-pik (but happy we can tidy up later): Could we consider calling configure() from handle_request_configure() to reduce duplication? Just create a wlr_box based on event.{x,y} and width/height.

Sounds like a nice cleanup 👍 

> Reading this through, I'm thinking that we ought to consider removing:
> - wlr_scene_node_set_position() and ssd_update_geometry() from view_move()
> - ssd_update_geometry() from view_move_resize()
>
> ...and just let commit-handlers do those things.

I noticed some inconsistencies there too. I am not sure of the best solution. Perhaps it would make sense to call `wlr_scene_node_set_position()` earlier (before `wlr_xwayland_surface_configure()`) in all of the different cases? But then I am not sure what is left for the commit-handlers to do -- or if there is even a reason for the `pending_move_resize` stuff to still exist at that point.

PR #429, opened 2022-07-02
src/view.c: Fall back to default geometry when changing state

### LINKED ISSUE #403: Restore geometry issues with windows that start out maximized
In apps that start out maximized (have no geometry to restore?), when dragging the window away with the mouse the window is moved so that the cursor is in the left side of the titlebar [Screenshot](https://user-images.githubusercontent.com/4444593/174476149-670bf1f1-097e-48da-9187-9e9c3e2a243b.jpg) < see terminal window

Firefox with two windows that were maximized before closing Firefox and opening again, [one is positioned over the panel](https://user-images.githubusercontent.com/4444593/174476699-841e48e6-25d8-4163-bd87-db1646ceedcd.jpg)
 and has the issue, one is properly maximized and doesn't. (Should probably be a new issue?)

Consolatis:
This makes sure that applications starting in maximized of fullscreen
mode always have their `natural_geometry` set to sensible values.

Partly fixes #403

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -8,6 +8,8 @@
 #include "menu/menu.h"
 #include "workspaces.h"
 
+#define LAB_FALLBACK_WIDTH 640
+#define LAB_FALLBACK_HEIGHT 480
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 
 /**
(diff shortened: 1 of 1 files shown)


src/view.c
```diff
@@ -268,6 +261,38 @@ view_compute_centered_position(struct view *view, int w, int h, int *x, int *y)
...
+
+static void
+view_store_natural_geometry(struct view *view)
+{
+	/**
+	 * If an application was started maximized or fullscreened, its
+	 * natural_geometry width/height may still be zero in which case we set
+	 * some fallback values. This is the case with foot and Qt applications.
+	 */
+	if (!view->w && !view->h) {
```
Consolatis, line 286: Do we want to keep the `&&` here or instead use `||` ?

johanmalm: Yes, `||` makes more sense.
I don't suspect it'll make a difference in reality, but `||` is good.

Consolatis: changed to `||` and rebased on latest master

PR #430, opened 2022-07-04
src/osd.c: Add preview outlines

Consolatis:
outlines

Missing:
- [x] use some theme defined color for the outlines
- [x] use some theme defined width of the outlines
- [ ] ~~use some config setting to enable / disable the outlines~~
- [x] destroy the multi_rect on `Reconfigure` (so its using the new theme when used the next time)
- [ ] extend the theme to define the 3 colors + width but keep falling back to `osd.bg` + `osd.label.text` + `osd.border.width`
- [ ] do we want this in the first place?

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -202,6 +202,8 @@ struct server {
 	/* Tree for built in menu */
 	struct wlr_scene_tree *menu_tree;
 
+	struct multi_rect *osd_preview_outline;
+
 	/* Workspaces */
 	struct wl_list workspaces;  /* struct workspace.link */
 	struct workspace *workspace_current;
--- a/src/osd.c
+++ b/src/osd.c
@@ -83,6 +83,32 @@ destroy_osd_nodes(struct output *output)
 	}
 }
 
+static void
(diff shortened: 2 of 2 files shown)


heroin-moose:
Awesome!

jlindgren90:
This would be nice :)

Just throwing an idea out there - would it be difficult to temporarily render the whole window on top, rather than just the outline? I think to me that would be more appealing visually, just a personal preference though.

johanmalm:
Agree. Window preview is the way to go.
We supported this before scene-graph and it was nice.
Consolatis drafted a PR with the “current”window-node moved to the front and an anchor-pointer to remember where it should go back if not chosen.
From memory, the edge-cases were a bit troublesome (like windows closing whilst alt-tabbing)

Consolatis:
Windows closing while alt-tabbing was fine (AFAIR). The issue was mainly previously minimized windows.
If anybody wants to take over this one or #307 go ahead, I might not work on these two for the next month or two.

Could rebase #307 to fix the merge conflicts before somebody else takes over, just throw me a message in that case.

PR #433, opened 2022-07-06
[wip] Add SnapToRegion

Consolatis:
regions

Features:
- Regions are defined by the user via `rc.xml` based on output size percentages
- Keyboard based snapping via new action `<action name="SnapToRegion" id="top-right" />`
- Mouse based snapping via holding any modifier while moving a window
- Region snapped windows will update based on output resolution changes and usable_area changes (e.g. panels)
- Region snapped windows restore to original geometry when moved
- Moving a window with any keyboard modifier pressed will show a preview overlay of the area to snap to

Missing:
- On `Reconfigure` region changes are only applied to new outputs.
- The `id` argument for a `<region>` has to come first or funky stuff will happen. Not yet sure how to solve that with the current config parsing. Seems also the case for the `name` argument for `action`.
- `view->tiled_region` isn't reset when we loose an output. This may cause segfaults.
- The half transparent overlay might \*really\* hurt the performance on bigger outputs using software rendering. We should fall back to a simple solid outline when detecting software rendering.

Open questions:
- Should we rename the `id` argument for `SnapToRegion` to `region`?

Configuration example:
```xml
        <!-- Percent based regions based on output->usable_area, the % char is optional -->
        <regions>
                <region id="bottom-left" x="0%"  y="50%" height="50%"  width="50%"  />
                <region id="bottom" x="0%"  y="50%" height="50%"  width="100%" />
                <region id="bottom-right" x="50%" y="50%" height="50%"  width="50%"  />
                <region id="left" x="0%"  y="0%"  height="100%" width="50%"  />
                <!-- following region must be rejected while parsing as the height is bigger than 100% -->
                <region id="5" x="0%"  y="0%"  height="120%" width="100%" />
                <region id="right" x="50%" y="0%"  height="100%" width="50%"  />
                <region id="top-left" x="0%"  y="0%"  height="50%"  width="50%"  />
                <region id="top" x="0%"  y="0%"  height="50%"  width="100%" />
                <region id="top-right" x="50%" y="0%"  height="50%"  width="50%"  />
        </regions>
        <keyboard>
                <!-- Resize via W-Numpad -->
                <keybind key="W-KP_1">
                        <action name="SnapToRegion" id="bottom-left" />
                </keybind>
                <keybind key="W-KP_2">
                        <action name="SnapToRegion" id="bottom" />
                </keybind>
                <keybind key="W-KP_3">
                        <action name="SnapToRegion" id="bottom-right" />
                </keybind>
                <keybind key="W-KP_4">
                        <action name="SnapToRegion" id="left" />
                </keybind>
                <keybind key="W-KP_5">
                        <!-- Invalid id -->
                        <action name="SnapToRegion" id="magic" />
                </keybind>
                <keybind key="W-KP_6">
                        <action name="SnapToRegion" id="right" />
                </keybind>
                <keybind key="W-KP_7">
                        <action name="SnapToRegion" id="top-left" />
                </keybind>
                <keybind key="W-KP_8">
                        <action name="SnapToRegion" id="top" /> 
                </keybind>
                <keybind key="W-KP_9">
                        <action name="SnapToRegion" id="top-right" />
                </keybind>
        </keyboard>
```

Testing welcome

Fixes:
- #386
- #122

Replaces:
- #388


heroin-moose:
One of the things I dislike about tiling window manager is that they modify the window geometry. I like `MoveToEdge` because it just moves the window preserving it's geometry. It would nice to have an option to move a window to the region without touching it's current size.

Consolatis:
> It would nice to have an option to move a window to the region without touching it's current size.

Never though about that.. I mean theoretically we could easily add a very simple `MoveToRegion` action which just moves the window the the x and y position of the region defined via argument without changing the size. It would not be an actual tiled state so no update on output resolution changes or when starting exclusive zone layershell clients.

Not sure of the usefulness of that feature in practice. As that feature would really just ride on top of the already implemented `SnapToRegion` feature its basically just some lines of code in `action.c` + a very small `view_move_to_region()` function, maybe like 25 lines of code total and no change of any state. In this case I would argue its worth the added complexity because there is basically no added complexity.

heroin-moose:
Well, currently MoveToEdege does just that. AFAIR this ticket started as "let's give the user the ability to define where to move/snap", so if that is possible, it would be nice to have.

johanmalm:
@heroin-moose Would the window action MoveResizeTo cover this (not implemented yet)?

`<action name="MoveResizeTo"><x></x><y></y></action>`

http://openbox.org/wiki/Help:Actions#MoveResizeTo

It feels weird to combine this with `<regions>` (but I don't think you meant that anyway).

johanmalm:
> Added a software rendering fallback for the overlay, its just a solid block for now but I think using a 2 to 4px outline might be more visually pleasing.
> Solid
> Outline using `theme->window_active_border_color`
> Outline using the same color as the solid variant
> Not sure which route to take

My thoughts:

- I experienced a similar dilemma when writing the original code for the alt-tab cycle frame. At the time I chose the openbox approach which consisted of a 6 pixel border where the outermost and innermost pixels were white, and the rest black. See: https://github.com/labwc/labwc/blob/12b05604a2bf128d818822288de253908d9d03a8/src/output.c#L443-L452
- The reason for the choice was that using only one color sometimes made the border quite hard to see - if the background/client colors were too close.
- Black/white aren't the prettiest. Maybe a better option is using a combination of `window.active.title.bg.color` and `window.active.title.label.color` or something like that. They're likely to always be quite contrasting.
- Another option could be `osd.bg.color` and `osd.border.color`

A completely different thought: would we want to show all regions (maybe with alpha 0.5 or something) during the interactive move?

Consolatis:
> A completely different thought: would we want to show all regions (maybe with alpha 0.5 or something) during the interactive move?

That would make it hard to get feedback which region we would snap to when releasing the mouse button. Unless we mark that one again in a different color or with some outline. It would also be an issue with overlapping regions like `top` and `top-left` in the example config above.

Regarding the pixman renderer fallback: thought about a variant like that as well, will be pretty convoluted if we use `wlr_scene_rects` for those though. Guess we could use cairo instead and turn the 4 "rects" (a.k.a. lines) into `wlr_scene_buffer`s.

johanmalm:
> That would make it hard to get feedback which region we would snap to when releasing the mouse button. Unless we mark that one again in a different color or with some outline. It would also be an issue with overlapping regions like `top` and `top-left` in the example config above.

Just a wild thought :smile: I'm probably getting a bit carried away.
Yes, the overlapping regions would be quite tricky. Not sure there is a simple generic solution to that one.

Consolatis:
> Another option could be osd.bg.color and osd.border.color

Went with that one:
regions_pixman_double_outline

And the same for Numix:
regions_pixman_double_outline_numix

Consolatis:
Rebased on latest master and cleaned the messy commits.

@Flrian Thanks for testing. As discussed via IRC, quite possible that there is a rounding error causing that 1px offset. Will look into that gap (and the other ones as well) when working on this the next time.

Consolatis:
@Flrian got some gaps for you (ignore the blur in the screenshot, my GPU drivers are doing strange stuff with red color).
regions_gap

Didn't investigate the rounding issue yet so that has to wait for the next time.

src/cursor.c
```diff
@@ -185,6 +186,16 @@ process_cursor_move(struct server *server, uint32_t time)
 	dy += server->grab_box.y;
 	resistance_move_apply(view, &dx, &dy);
 	view_move(view, dx, dy);
+
+	struct wlr_keyboard *keyboard = &server->seat.keyboard_group->keyboard;
+	if (keyboard_any_modifiers_pressed(keyboard)) {
```
Consolatis, line 191: Verify this is still correct after the recent keyboard_group changes

src/interactive.c
```diff
@@ -98,11 +100,26 @@ void
...
 			 && rc.snap_edge_range;
 		view->server->input_mode = LAB_INPUT_STATE_PASSTHROUGH;
 		view->server->grabbed_view = NULL;
+
 		if (should_snap) {
+			/* SnapToRegion */
+			struct wlr_keyboard *keyboard =
+				&view->server->seat.keyboard_group->keyboard;
+
+			if (keyboard_any_modifiers_pressed(keyboard)) {
```
Consolatis, line 114: Verify this is still correct after the recent keyboard_group changes

src/menu/menu.c
```diff
@@ -240,6 +241,10 @@ fill_item(char *nodename, char *content)
 		 */
 		action_arg_add_str(current_item_action, NULL, content);
 	} else if (!strcmp(nodename, "to.action")) {
+		/* GoToDesktop, SendToDesktop */
+		action_arg_add_str(current_item_action, NULL, content);
```
Consolatis, line 245: Split out new comments into own PR to ease rebasing.

heroin-moose:
Please merge this, this would make my life so much easier.

Consolatis:
> Please merge this, this would make my life so much easier.

There is quite a bit of rebasing to do here which will take some time.
I'll try to get it done within the next two weeks but no promises.

Consolatis:
Crash related to region preview overlay vs OSD preview.
Caused by pressing `Alt` + moving window + pressing `Tab` (so OSD is shown) + releasing button on region (thus snapping).
```
labwc: types/scene/wlr_scene.c:836: wlr_scene_node_place_above: Assertion `node->parent == sibling->parent' failed.

Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=281474841849888, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (threadid=281474841849888, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x0000fffff6a109e4 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x0000fffff69ca76c in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x0000fffff69b74bc in __GI_abort () at ./stdlib/abort.c:79
#4  0x0000fffff69c41e4 in __assert_fail_base
    (fmt=0xfffff6adf148 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0xaaaaaae60268 "node->parent == sibling->parent", file=file@entry=0xaaaaaae600e8 "types/scene/wlr_scene.c", line=line@entry=836, function=function@entry=0xaaaaaae60490 <__PRETTY_FUNCTION__.10> "wlr_scene_node_place_above") at ./assert/assert.c:92
#5  0x0000fffff69c424c in __GI___assert_fail (assertion=0xaaaaaae60268 "node->parent == sibling->parent", file=0xaaaaaae600e8 "types/scene/wlr_scene.c", line=836, function=0xaaaaaae60490 <__PRETTY_FUNCTION__.10> "wlr_scene_node_place_above")
    at ./assert/assert.c:101
#6  0x0000aaaaaad3ce70 in wlr_scene_node_place_above (node=0xaaaaab88f820, sibling=0xaaaaab7eee00) at ../subprojects/wlroots/types/scene/wlr_scene.c:836
#7  0x0000aaaaaac46d74 in osd_preview_restore (server=0xffffffffed98) at ../src/osd.c:192
#8  0x0000aaaaaac36790 in desktop_cycle_view (server=0xffffffffed98, start_view=0xaaaaab88f580, dir=LAB_CYCLE_DIR_FORWARD) at ../src/desktop.c:221
#9  0x0000aaaaaac3e348 in handle_compositor_keybindings (keyboard=0xaaaaab7d74c0, event=0xffffffffe880) at ../src/keyboard.c:213
#10 0x0000aaaaaac3e6bc in handle_keybind_repeat (data=0xaaaaab7d74c0) at ../src/keyboard.c:251
#11 0x0000fffff7f7b390 in wl_event_loop_dispatch () at /lib/aarch64-linux-gnu/libwayland-server.so.0
#12 0x0000fffff7f789dc in wl_display_run () at /lib/aarch64-linux-gnu/libwayland-server.so.0
#13 0x0000aaaaaac44870 in main (argc=4, argv=0xfffffffff578) at ../src/main.c:179
```

src/server.c
```diff
@@ -16,6 +16,7 @@
 #include "labwc.h"
 #include "layers.h"
 #include "menu/menu.h"
+#include "regions.h"
```
Consolatis, line 19: remove

src/view.c
```diff
@@ -349,6 +356,79 @@ view_center(struct view *view)
...
+		}
+
+		/* Get new output local region */
+		view->tiled_region = regions_from_id(
+			view->tiled_region_id_evacuate, output);
+		/*
+		 * FIXME:
+		 * This will break once Reconfigure
+		 * support for regions is implemented.
+		 */
```
Consolatis, line 380: It will break even before that for new outputs

src/regions.c
```diff
@@ -0,0 +1,242 @@
...
+{
+	assert(output);
+	struct view *view;
+	struct region *region;
+
+	wl_list_for_each(view, &output->server->views, link) {
+		wl_list_for_each(region, &output->regions, link) {
+			if (view->tiled_region == region) {
+				if (!view->tiled_region_id_evacuate) {
+					view->tiled_region_id_evacuate = xstrdup(region->id);
```
Consolatis, line 218: This may leak memory, we need to ensure its free'd in view_destroy

Consolatis:
I am largely satisfied with the current state.

This could need some final testing and review especially of the added docs but overall I think this is close to be merge ready.

There is also still one open question:
Should we rename the `id` argument for `SnapToRegion` to `region`?

e.g.
```xml
<action name="SnapToRegion" id="bottom-left" />
```
vs
```xml
<action name="SnapToRegion" region="bottom-left" />
```

And similar for the `<region>` definition:
```xml
 <region id="bottom-left" x="0%"  y="50%" height="50%"  width="50%"  />
```
vs
```xml
 <region name="bottom-left" x="0%"  y="50%" height="50%"  width="50%"  />
```
or
```xml
 <region region="bottom-left" x="0%"  y="50%" height="50%"  width="50%"  />
```

Personally I think I'd prefer
```xml
<regions>
	<region name="bottom-left" x="0%"  y="50%" height="50%"  width="50%"  />
</regions>
<keyboard>
	<default />
	<keybind key="W-whatever">
		<action name="SnapToRegion" region="bottom-left" />
	</keybind>
</keyboard>
```

johanmalm:
> So `name` for the region definition and `region` for the action.

Agree with this. Feel clear.

src/config/rcxml.c
```diff
@@ -400,6 +420,12 @@ entry(xmlNode *node, char *nodename, char *content)
 		wl_list_append(&rc.workspace_config.workspaces, &workspace->link);
 	} else if (!strcasecmp(nodename, "popupTime.desktops")) {
 		rc.workspace_config.popuptime = atoi(content);
+	} else if (!strcasecmp(nodename, "id.region.regions")) {
+		current_region = znew(*current_region);
+		wl_list_append(&rc.regions, &current_region->link);
+		fill_region_geo(nodename, content);
```
johanmalm, line 426: `region` elements are parsed differently to keybinds, mousebinds and libinput devices.
Consider setting an `in_region` bool in `xml_tree_walk()` and `zalloc()` on hitting id/name.
I guess either will work fine, but the recursive `xml_tree_walk()` is more consistent.

include/regions.h
```diff
@@ -0,0 +1,44 @@
...
+	union {
+		struct wlr_scene_rect *overlay;
+		struct multi_rect *pixman_overlay;
+	};
+};
+
+void regions_reconfigure(struct server *server);
+void regions_reconfigure_output(struct output *output);
+void regions_update_geometry(struct output *output);
+void regions_evacuate_output(struct output *output);
```
johanmalm, line 36: Consider adding a comment here do describe what `regions_evacuate_output()` does.

Consolatis:
Mostly internal changes:
- `include/regions.h`: add some docs
- `src/config/rcxml.c`: use `fill_region` depending on bool `in_regions`
- `docs/rc.xml.all`: add a more complete (commented out) example
- `src/view.c`: add public `view_is_tiled(view)` function
- `src/view.c`: free `view->tiled_region_evacuate` on `view_set_untiled()`
- prevent dangling `view->tiled_region` pointer by treating `view->tiled_region_evacuate` as tiled
- config: use `name` for the region definition and `region` for the action.
- rename `(struct region)->id` to `name`
- rename `view->tiled_region_id_evacuate` to `view->tiled_region_evacuate`
- rename `regions_from_id()` to `regions_from_name()`
- further compact commits: remove update ones by merging into their original commits

include/regions.h
Consolatis: Added a comment, I hope its somewhat understandable for somebody not inherently used to the codebase.

src/config/rcxml.c
Consolatis: done

johanmalm:
I've just tried with [real] multi-output setup. There is something not right with output-evacuation.
With window set to a region on HDMI-A-1, if I turn HDMI-A-1 the window moves to eDP1 but NOT with co-ordinates that correspond to the region on that new output.
It works well with `WLR_WL_OUTPUTS=2` though!
Need to run, but will be back later and can discuss/think through on IRC

Consolatis:
I think there is a bug with the keybind action when defining region names that are contained within other region names.
E.g. "top", "top-left", "top-right" should work but "top-left", "top-right", "top" will likely use the "top-right" region when actually requesting the "top" one. Because `strcmp()` finds "top" and returns a match.

Consolatis:
Thanks to @johanmalm for debugging / testing via IRC. The last commit should fix the issue, it was related to disabling an output vs. unplugging it.

Also thanks to @Flrian for testing / reporting another issue of firefox wayland native not wanting to switch regions, should be fixed as well.

We intend to go for a merge around Wednesday so more test results / issues are always welcome.

johanmalm:
Geometry attributes in pixels are parsed as percentage values.

Input:

```
  <regions>
    <region name="top-left" x="0" y="0" height="369" width="683"/>
    <region name="top-right" x="683" y="0" height="369" width="683"/>
    <region name="bottom-left" x="0" y="369" height="369" width="683"/>
    <region name="bottom-right" x="683" y="369" height="369" width="683"/>
  </regions>
```

output

```
00:00:00.000 [../src/config/rcxml.c:749] Removing invalid region 'top-left': 683% x 369% @ 0%,0%
00:00:00.000 [../src/config/rcxml.c:749] Removing invalid region 'top-right': 683% x 369% @ 683%,0%
00:00:00.000 [../src/config/rcxml.c:749] Removing invalid region 'bottom-left': 683% x 369% @ 0%,369%
00:00:00.000 [../src/config/rcxml.c:749] Removing invalid region 'bottom-right': 683% x 369% @ 683%,369%

```

johanmalm:
Discovered with latest https://github.com/johanmalm/regions

Consolatis:
Right, this PR only supports percent based regions so they always work, independent of the actual output resolution (which might differ when you have multiple outputs) and existence of exclusive layershell zones.

johanmalm:
> Right, this PR only supports percent based regions so they always work, independent of the actual output resolution (which might differ when you have multiple outputs) and existence of exclusive layershell zones.

Okay. Sorry, ignore my latest.
I misinterpreted the `labwc-config(5)` words `The "%" character is optional`.

Consolatis:
I got some feedback from somebody else about this and their argument about supporting both - absolute pixels + percent - was that there may be applications like video players you want to dock somewhere which just look weird when not defining the exact dimensions in pixels. I think that is a valid point. Maybe we could indeed extend this and add support for absolute pixel values in the future. For now I think the percent based values cover most use cases.

If we agree on possibly extending this in the future to treat numbers without trailing `%` as pixels I would remove the `is optional` part of the docs and reject regions not having a trailing `%`.

johanmalm:
> If we agree on possibly extending this in the future to treat numbers without trailing `%` as pixels I would remove the `is optional` part of the docs and reject regions not having a trailing `%`.

Good idea. Let’s do that. Seems cleaner.

Consolatis:
Squashed all the `[fixup]` commits into their appropriate commits and rebased on latest master.
Further changes:
- Change docs and config parsing to require the `%` char to keep the format open for further extension to also carry pixel values
- If a user uses multiple `name` attributes for a single `<region>` we won't leak memory anymore and just ignore later `name` attributes.

Consolatis:
Thanks everybody involved for testing / review :)

Narrat:
I kinda see me switching the daily driver anytime soon :D
Just a question, but would it be in general possible to enhance the current implementation to define layouts from regions and if invoked it takes whatever open apps there are und place them in the chosen regions? (Or in short: Layouts from [hikari](https://hikari.acmelabs.space/manpage.html))

johanmalm:
That’s cool.
I’m not that familiar with hikaris layout workflow.
How about we try creating a client to prototype the workflow?

Consolatis:
> How about we try creating a client to prototype the workflow?

This is be a bit hacky but could work:
- foreign toplevel to get a list of open windows handles that are not minimized / maximized / fullscreen
- for each window handle
  - foreign toplevel to focus
  - `wtype -M logo -P KP_1` to call some keybind to snap now focused window to region 1

In regards to implementing that kind of logic in labwc itself: I feel a bit two sided on this. On the one hand it wouldn't be that hard to implement (although the config format would need some thought) but on the other hand that would really start to move the line a bit towards a tiling window manager. My fear there is that people start to rely on that feature and then expect all the little things to get implemented / fixed as well. E.g. what about workspaces in relation to layouts, what about persisting the client ordering within those layouts, moving windows to another output to tile it there, ..

So to sum it up: implementing wouldn't be that hard, maintaining for the future might be.

johanmalm:
From a quick read of hikari's man page it looks like the have a register of layouts (a-z).
Guess we could just write a client with a bunch of 'layouts' (possibly editable) and use that to change the `rc.xml` regions.
(A more advanced version of https://github.com/johanmalm/regions)

Narrat:
That it may even possible with a separate client is good to know.
Regarding for the mentioned concerns: Hikari itself is still a stacking compositor although it has some tiling capabilities. And with some I mean a veeeeery basic one. And personally I only want a basic one as I cannot come to terms with tiling wm...
It just helps setting up your workspace. E.g. I have one workspace with three terminals and one with an file explorer plus 2 terminals. Instead of positioning every program in its place with its desired size I can just use a keyboard shortcut.
Done since my days with openbox via autostart and placement rules :D It was more convinient with the way hikari did it and simpler to setup. But sadly it seems the development kinda came to a stop... :/ Therefore I was happy to see regions got added.

In regards to the points raised:
>workspaces in relation to layouts

Dunno if I can follow you on this one. If wished a layout is applied per workspace

> persisting the client ordering within those layouts

Again not sure what is meant with this. In general it is only possible to switch the current active window with the first or previous window. Tabbing through the windows is like regular tabbing on floating windows.

> moving windows to another output to tile it there

Not part of hikari. If moved to another output it isn't part of the layout anymore and is again just a regular floating window.

> tiling newly spawned windows to the existing layout

Not part of hikari. If a layout is applied only the current open windows on the currently active workspace are considered. Any newly spawned window is floating. If it should be part of an layout you need to set again the layout.

> having more windows around than are defined for the currently active layout

They are ignored and will be floating windows.

> moving windows around within the active layout

They change their position but it doesn't change anything in regarding the size of the windows. The tiling is kinda just a case of: Create regions and fit/resize windows in those. And after that.. whatever :D For example if you close a window which is part of a layout it will just close. No resizing or whatever to fill the place where it was.


Regarding of setting those up. As mentioned it is possible to define layouts from a-z. And in the layout config one declares the regions and what happens.
Example from my config:
```
  a = {
    scale = 0.55
    left = single
    right = stack
  }
```
Workspace is divided vertically. Left side 55% of the output and just a single window. On the right side windows get stacked.
 ```
  x = {
    scale = 0.45
    right = single
    #left = stack
    left = {
      scale = 0.45
      top = single
      bottom = single
    }
 ```
Another similar example: In this case the ratio is 45% for the left side. Single window on the right side. Left is also scaled and set up in such a way that it only allows two windows to be tiled. If there are four the fourth gets ignored and stays as is.

I hope that clarifies it somewhat?

Arnaudv6:
@Narrat , I'd like to have rules to create windows with given position/size, based on their [app_]id.
I have the feeling that this would cover much of your need also?
Then you could autostart some terms with predefined ids, that would pop at the right place. Is this so?

Narrat:
@Arnaudv6 This would replicate the openbox behaviour. And yes, it kinda fills the same space, but it is a static setup. Sessions is starting and it will be launched everytime. That may not be my wish at times.
And I still want to setup a layout on the fly which I may not need every day.

Consolatis:
Personally I just use tmux for my (terminal) tiling needs, I usually have just a single terminal window open, maximized and without decorations running tmux. All this automatic positioning feels a bit out of scope for labwc to me, its like opening a can of worms with never-ending requests coming in to support specific workflows. Regarding window policies I think we have some aspects of them categorized as `B` so a subset might be implemented in labwc at some point but I don't remember right now if the positioning will be part of that.

Anyway, here is a quick hack to get some kind of layout working with an external client:
<details><summary>quick 'n dirty layout hack</summary>

Requires https://github.com/Consolatis/wl_framework
Copy the script below into its root directory and make it executable. See the top of the file for configuration.
With its stock config the script requires `W-KP_1` to `W-KP_x` (`x` being `MAX_REGIONS`) to be configured in `rc.xml` with `SnapToRegion`. It could easily be enhanced to react to `window.app_id` or `window.title`.

This will mix up windows from different workspaces because https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40 is still in limbo.

```python3
#!/usr/bin/env python3

from wl_framework.network.connection import WaylandConnection, WaylandDisconnected
from wl_framework.protocols.foreign_toplevel import ForeignTopLevel
from wl_framework.protocols.virtual_keyboard import VirtualKeyboardManager


MAX_REGIONS = 3
KEYBIND_MODIFIER = "MOD_LOGO"
KEYBIND_FORMAT = "KP_{}"


class LabwcLayout(WaylandConnection):

	def on_initial_sync(self, data):
		super().on_initial_sync(data)
		self.manager = VirtualKeyboardManager(self)
		self.keyboard = self.manager.create_virtual_keyboard(self.display.seat)
		self.toplevels = ForeignTopLevel(self)
		self.sync(self.info_done)

	def _get_windows(self):
		filtered = list()
		for window in self.toplevels.windows.values():
			for state in ('minimized', 'maximized', 'fullscreen'):
				if state in window.states:
					break
			else:
				if 'activated' in window.states:
					filtered.insert(0, window)
				else:
					filtered.append(window)
		return filtered

	def info_done(self, data):
		windows = self._get_windows()
		for index, window in enumerate(windows[:MAX_REGIONS]):

			window.activate(self.display.seat)
			with self.keyboard.modifier(getattr(self.keyboard, KEYBIND_MODIFIER)):
				self.keyboard.send_key(KEYBIND_FORMAT.format(index + 1))

		if windows:
			# Ensure previous focused window is still focused
			windows[0].activate(self.display.seat)

		self.sync(self.on_done)

	def on_done(self, data):
		self.shutdown()

if __name__ == '__main__':

	from wl_framework.loop_integrations import PollIntegration
	loop = PollIntegration()

	try:
		app = LabwcLayout(eventloop_integration=loop)
		loop.run()
	except RuntimeError as e:
		print(e)
		exit(1)
	except KeyboardInterrupt:
		print()
	except WaylandDisconnected:
		pass
```

</details>

Narrat:
Gosh.. already half a year... But I finally could come back to this.
First and foremost: I really appreciate it that you took some of your time listening to this question and working something up despite your reservations. Not something which can be taken for granted.
And it's kinda funny comparing workflows. It seems we are on polar opposites, as I don't really like terminal multiplexers :D No wonder I'm having a hard time arguing ;)
The script as a proof of concept works really fine and is promising. It shows it is generally working as a solution outside of a compositor. So no additional burden on something possibly rarely used :D
Interesting note regarding the workspaces. I generally assumed the layout or how apps are tracked by a compositor involves actively the workspace. But that isn't a requirement then? So hikari does something on top to mark/track applications to a specific workspace? At least there is the slight chance of an official protocol in that regard. Will follow the development of that.

Consolatis:
> I generally assumed the layout or how apps are tracked by a compositor involves actively the workspace. But that isn't a requirement then? So hikari does something on top to mark/track applications to a specific workspace?

Right, as far as I am aware there is no notion of workspaces in any wayland protocol at all, similar as how a window doesn't know anything about its global position. All wayland compositors that support some kind of workspaces (or tiles / regions) implement their own variant.

> At least there is the slight chance of an official protocol in that regard.

That one is about describing a window ("toplevel") relation with an abstract workspace to a third party application and allow said third party application to move a window to a different workspace. The window itself still has no clue about it being on some specific workspace. I really hope that we get that protocol merged at some point so cross-compositor compatible panels are able to actually offer that option (instead of having to use compositor keybinds / some compositor defined window menu).

And it would allow a tool like above to get that information as well to filter out all windows that are not on the current workspace.

Narrat:
> By the way we merged the window policy framework in the meantime which could be used for some basic static layout configuration.

That's nice. One of the topics I need to look after then. Plus looking more closely what additionally arrived with the releases between 0.6.1 and 0.6.4

PR #435, opened 2022-07-08
src/output.c: Always react to new output configuration

Consolatis:
Without this fix we will never react to output configuration changes
via wlr-randr and friends.

If `output_config_apply()` commits the new config, `handle_output_layout_change()`
is called but doesn't do anything because the config is still pending.

This patch moves the code into its own function `do_output_layout_change()`
which additionally gets called after all output configs are committed.

The original handler is turned into a wrapper around `do_output_layout_change()`.

Reported-by @heroin-moose


johanmalm:
Thanks. Great patch.

PR #437, opened 2022-07-10
src/view.c: Prevent crash by killing a moving application

Consolatis:
When a view is destroyed labwc calls `interactive_end(view)` which
may reposition the view which is partly destroyed and doesn't own
any surface anymore. To prevent this scenario from happening don't
call `interactive_end()` at all and just reset `server->grabbed_view`
and `server->input_mode` directly.

Before this patch, the bug can be reproduced by:
- xcalc &
- sleep 5; killall xcalc
- move the xcalc window completely to one of the edges

The change in `src/xwayland.c` is not required for this bug
to be fixed but may prevent something similar in the future.


Consolatis:
The build error for Void-musl is related to not finding `libxml-2.0`. However, it installed  the package `libxml2-2.9.13_3`. Might be a temporary issue with that version or something else but In any case unrelated to this PR.

Issue #412, opened 2022-06-21
imv crashes sometimes


acoolstraw:
[imv](https://sr.ht/~exec64/imv/) crashes sometimes, I'm not sure how exactly to reproduce it, but if you A-Tab to another window and back and do some other stuff with the mouse it crashes with this message:
`Floating point exception (core dumped)`

Consolatis:
I think we still leak a stray PRESS event for the modifier (`A` in this case) to the focused application when doing a keybind involving modifiers.
That could be the issue here but I have no idea what has to do with that floating point exception.

bi4k8:
This is at least in part an imv bug; the below patch for imv treats a zero repeat interval as 1, which is wrong but prevents the crash:
```
diff --git a/src/wl_window.c b/src/wl_window.c
index 5efa42f..4b8dd1c 100644
--- a/src/wl_window.c
+++ b/src/wl_window.c
@@ -239,7 +239,7 @@ static void keyboard_repeat(void *data, struct wl_keyboard *keyboard,
   (void)keyboard;
   struct imv_window *window = data;
   window->repeat_delay = delay;
-  window->repeat_interval = 1000 / rate;
+  window->repeat_interval = 1000 / (rate ? rate : 1);
 }
 
 static const struct wl_keyboard_listener keyboard_listener = {
```

On the labwc side, it's concerning that we send a `repeat_info.rate` of 0. This may have something to do with the key repeat problem [mentioned](https://github.com/labwc/labwc/issues/203#issuecomment-1058205955) in #203.

bi4k8:
Apparently we set the key repeat rate to 0 whenever a keybind is activated:
```C
static bool
handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym)
{
	struct keybind *keybind;
	struct wlr_keyboard *kb = &server->seat.keyboard_group->keyboard;
	wl_list_for_each_reverse (keybind, &rc.keybinds, link) {
		if (modifiers ^ keybind->modifiers) {
			continue;
		}
		for (size_t i = 0; i < keybind->keysyms_len; i++) {
			if (xkb_keysym_to_lower(sym) == keybind->keysyms[i]) {
				wlr_keyboard_set_repeat_info(kb, 0, 0);
				actions_run(NULL, server, &keybind->actions, 0);
				return true;
			}
		}
	}
	return false;
}
```

[...]

PR #438, opened 2022-07-11
Revert "keyboard: cancel repeat when handling key-bind"

johanmalm:
This reverts commit e62bb51bfb33ee520e800cf98553d766824fe9cf.

Fix issue #412


johanmalm:
Maybe this is it.
https://github.com/swaywm/sway/commit/2573606b60f0606860634b90749f734cd0e918b7

johanmalm:
Testing appreciated. Works fine for me... apart from dmenu. Solves #510

Consolatis:
```
<Consolatis> > This one happens in master too
<Consolatis> just to be 100% sure: happens for both, PR and master?
<Flrian> yes
<Consolatis> interesting, so its caused by something that wasn't actually changed in that PR
<Consolatis> i might have an idea then.. when we handle a keybinding we have that list of pressed keys
<Consolatis> maybe we just add all currently pressed keys to it
<Consolatis> instead of only the ones defined in the keybind
<Consolatis> that would prevent the release event of the first pressed non-keybinding key to get to the application
<Consolatis> which then does the absolutely correct thing and repeats it until it either gets a release or a looses focus
<Consolatis> so that would totally explain that
<Consolatis> or looses focus*
<Consolatis> so the solution would be: only add keys to the list if they are actually part of the keybindd
```

src/keyboard.c
```diff
@@ -203,17 +243,20 @@ keyboard_key_notify(struct wl_listener *listener, void *data)
...
-	if (!seat->active_client_while_inhibited) {
-		handled = handle_compositor_keybindings(listener, event);
-	}
+	/*
+	 * Keys can be bound by C-A-F?? for TTY change, window cycling
+	 * (typically alt-tab, but could be any key/modifier), and user defind
+	 * keybinds
+	 */
+	wlr_log(WLR_DEBUG, "Number of pressed keys that were bound: %d\n",
+		key_state_nr_keys());
```
Consolatis, line 254: Should we maybe put this within a ifdef to prevent spamming the debug log on keypress?
```c
#ifdef LAB_DEBUG_INPUT
[..]
#endif
```

Consolatis:
@johanmalm Nice work!

I am running this on my DRM main session and only found one minor thing [0] but that should not prevent this PR to be merged. Also I'll be honest here: I don't understand all the details of the changes (nor of our keyboard input handling in general) but didn't spot anything suspicious. So this seems ready to go. Thanks again for your work on this nasty issue :)

[0]
We still leak a `Press` event for a modifier (e.g. `Alt_L`) to the focused surfaces when issuing a keybind containing a modifier.
E.g. when opening  xev / wev, focusing it, pressing `A-Tab` to move away we leak a stray `Press` for the modifier (`Alt_L` in this case). When switching back to it not a single keybinding related keypress is sent which is an amazing improvement.

src/keyboard.c
johanmalm: I've just deleted the commit that added that `wlr_log()`. For keyboard debugging purpose you are likely to need `fprintf()s` is several places, so on reflection, I don't see much point in keeping it.

Consolatis: also works :)

johanmalm:
> We still leak a `Press` event for a modifier (e.g. `Alt_L`) to the focused surfaces when issuing a keybind containing a modifier. E.g. when opening xev / wev, focusing it, pressing `A-Tab` to move away we leak a stray `Press` for the modifier (`Alt_L` in my case). When switching back to it (or executing something via keybind) not a single keybinding related keypress is sent which is an amazing improvement. I also didn't manage to break key-repeat nor have a key repeating without holding it down on purpose, including in a nested labwc.

I'm not 100%, but I think this is how it is meant to work. It's know easy to establish the "correct" behaviour because there is no manual for this :smile:

Consider the steps below:

1. Set `Alt_L` + `d` to launch `dmenu_run`
2. Open `mousepad` or `leafpad`
3. Press `Alt_L` and observe that menu entries (e.g. File, Edit) underline keyboard-shortcut letters
4. Press `d` and observe that `dmenu_run` starts and the underlined letters in `mousepad` / `leafpad` disappear.

I think clients have to operate this way because when `Alt_L` is pressed the client doesn't know if it's going to result in a compositor-keybind or not.

Sway does the same.

Consolatis:
Good points. I just call it "we are doing it the way its supposed to be done" from now on.

Issue #434, opened 2022-07-07
Alpine linux package


keshto:
From the build log it looks like it is not using xwayland and it is not a dependency. However, when trying to start labwc one gets a "Cannot create xwayland server /usr/bin/xwayland"

Consolatis:
The labwc Alpine package seems to depend on [wlroots](https://git.alpinelinux.org/aports/tree/community/wlroots/APKBUILD) which says in its [build log](https://build.alpinelinux.org/buildlogs/build-edge-x86_64/community/wlroots/wlroots-0.15.1-r1.log)
> xwayland        : YES

And the [labwc build](https://git.alpinelinux.org/aports/tree/community/labwc/APKBUILD) doesn't seem to disable xwayland either.
So if its really not a dependency it should likely be added as one.

Adding @ptrcnull as Alpine maintainer.

ptrcnull:
Should be fixed as soon as this is merged: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/36084/diffs

nekopsykose:
fixed, though i don't understand why it always tries to start xwayland just because wlroots supports it, instead of making it optional (read the code)

Consolatis:
We could try to degrade gracefully when labwc was built with xwayland support but it isn't available while executing.

That requires some more investigation on how massive these changes would be and depending on the result this might never be merged into the `v0.5` branch but go into the `master` branch only.

Turning this into a feature request.

keshto:
I think having that feature is a must have. We can close out this issue and create a new one for that feature though.

Hopefully alpine package will have a no xwayland dependant build sometime.

droc12345:
Best way would be to have an xwayland config option.
This would allow for ignoring xwayland even if compiled in.
And if it is seen that the xwayland option is set, but xwayland doesn't exist, 
then the config option could be turned off in the code.

Along with this every place that has HAVE_XWAYLAND needs to be looked at to see about adding a check 
for config->xwayland being set, because xwayland code sometimes gets called along with regular wayland calls
inside routines and those would need a conditional to keep from trying to execute.

Consolatis:
Same issue seems to exists on Fedora: https://www.reddit.com/r/labwc/comments/zo0xob/comment/j0kscgd/

quite:
I would like to revive this, to make it possible for distributions to not force installation of xwayland (the labwc package will depend on xwayland package, just like sway typically does not).

[...]

PR #441, opened 2022-07-11
Don't quit if xwayland not found

keshto:
Should fix: https://github.com/labwc/labwc/issues/434

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -397,36 +397,35 @@ server_init(struct server *server)
 		wlr_xwayland_create(server->wl_display, compositor, true);
 	if (!server->xwayland) {
 		wlr_log(WLR_ERROR, "cannot create xwayland server");
-		exit(EXIT_FAILURE);
-	}
-	server->new_xwayland_surface.notify = xwayland_surface_new;
-	wl_signal_add(&server->xwayland->events.new_surface,
+		// exit(EXIT_FAILURE);
+	} else {
+		server->new_xwayland_surface.notify = xwayland_surface_new;
+		wl_signal_add(&server->xwayland->events.new_surface,
 		      &server->new_xwayland_surface);
 
-	if (setenv("DISPLAY", server->xwayland->display_name, true) < 0) {
+		if (setenv("DISPLAY", server->xwayland->display_name, true) < 0) {
 		wlr_log_errno(WLR_ERROR, "unable to set DISPLAY for xwayland");
-	} else {
+		} else {
 		wlr_log(WLR_DEBUG, "xwayland is running on display %s",
 			server->xwayland->display_name);
+		}
+
+		struct wlr_xcursor *xcursor;
+		xcursor = wlr_xcursor_manager_get_xcursor(server->seat.xcursor_manager,
+						  XCURSOR_DEFAULT, 1);
+		if (xcursor) {
+		struct wlr_xcursor_image *image = xcursor->images[0];
+		wlr_xwayland_set_cursor(server->xwayland, image->buffer,
+					image->width * 4, image->width,
+					image->height, image->hotspot_x,
+					image->hotspot_y);
+		}
 	}
 #endif
 
 	if (!wlr_xcursor_manager_load(server->seat.xcursor_manager, 1)) {
 		wlr_log(WLR_ERROR, "cannot load xcursor theme");
 	}
-
-#if HAVE_XWAYLAND
-	struct wlr_xcursor *xcursor;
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks for fixing the Void-musl runner! Would you mind splitting those three commits into an own PR?

Can be done for example like this:
```shell
git checkout xway
git checkout -b fix/void
git rebase -i master
# just remove the "Don't quit if xwayland not found" commit from the list and exit the editor
```
Alternative:
```shell
git checkout master
git checkout -b fix/void
git cherry-pick 36c2e78c1138cddd2694edc35c2e4b9fe797aa58
git cherry-pick 4fd02b3017bb4f7f7db212aca350eebd63da581c
git cherry-pick ce4ea388bd57c4f69313e395e3d7b88f1e9d63ee
```

keshto:
You can go ahead and close this out. Labwc has features I rely on in a wm not in scope and don't want to spend time on something that I am not going to be able to use. Sorry.

src/server.c
```diff
@@ -397,37 +397,36 @@ server_init(struct server *server)
...
-	} else {
+		} else {
 		wlr_log(WLR_DEBUG, "xwayland is running on display %s",
 			server->xwayland->display_name);
-	}
-#endif
-
-	if (!wlr_xcursor_manager_load(server->seat.xcursor_manager, 1)) {
-		wlr_log(WLR_ERROR, "cannot load xcursor theme");
-	}
```
Consolatis, line 416: I don't think moving this down will work because:

src/server.c
```diff
@@ -397,37 +397,36 @@ server_init(struct server *server)
...
+		}
 
-#if HAVE_XWAYLAND
-	struct wlr_xcursor *xcursor;
-	xcursor = wlr_xcursor_manager_get_xcursor(server->seat.xcursor_manager,
+		struct wlr_xcursor *xcursor;
+		xcursor = wlr_xcursor_manager_get_xcursor(server->seat.xcursor_manager,
 						  XCURSOR_DEFAULT, 1);
```
Consolatis, line 415: it is actually used here.

`wlr_xcursor_manager_load()`:
> Ensures an xcursor theme at the given scale factor is loaded in the manager

src/server.c
```diff
@@ -397,37 +397,36 @@ server_init(struct server *server)
...
 		      &server->new_xwayland_surface);
 
-	if (setenv("DISPLAY", server->xwayland->display_name, true) < 0) {
+		if (setenv("DISPLAY", server->xwayland->display_name, true) < 0) {
 		wlr_log_errno(WLR_ERROR, "unable to set DISPLAY for xwayland");
-	} else {
+		} else {
 		wlr_log(WLR_DEBUG, "xwayland is running on display %s",
 			server->xwayland->display_name);
-	}
```
Consolatis, line 411: Wrong indentation within the `if` and `else` branches

src/server.c
```diff
@@ -397,37 +397,36 @@ server_init(struct server *server)
 		wlr_xwayland_create(server->wl_display, compositor, true);
 	if (!server->xwayland) {
```
Consolatis, line 398: Could be negated, thus the `wlr_log` and `exit` being removed and the `if` changed to
```c
if (server->xwayland) {
  ..
}
```

src/server.c
```diff
@@ -397,37 +397,36 @@ server_init(struct server *server)
...
+		xcursor = wlr_xcursor_manager_get_xcursor(server->seat.xcursor_manager,
 						  XCURSOR_DEFAULT, 1);
-	if (xcursor) {
+		if (xcursor) {
 		struct wlr_xcursor_image *image = xcursor->images[0];
 		wlr_xwayland_set_cursor(server->xwayland, image->buffer,
 					image->width * 4, image->width,
 					image->height, image->hotspot_x,
 					image->hotspot_y);
+		}
```
Consolatis, line 422: Wrong indentation within the `if`

Consolatis (changes requested):
Thanks for working on this. There are a couple of things that could be improved with this patch.

What kind of configurations did you test? I think these ones should be verified to work properly:
- labwc compiled with `-Dxwayland=disabled`
- labwc compiled normally but with xwayland binary available
- labwc compiled normally but without xwayland binary available

keshto:
I only tested the last case.

On July 11, 2022 10:31:35 PM UTC, Consolatis ***@***.***> wrote:
***@***.*** requested changes on this pull request.

PR #453, opened 2022-07-17
Fix a few more move/resize glitches

jlindgren90:
The main improvement here is to process move-only client configure requests more quickly for XWayland windows.
    
For a client configure request that doesn't resize the window, we now treat it similarly to `view_move()` by updating the scene-graph immediately, rather than waiting for `handle_commit()`.  Move-and-resize requests are handled the same as before.
    
This (mostly?) fixes the glitchiness that was noticeable when dragging an undecorated XWayland window (e.g. Audacious in Winamp mode).
    
Also reduce some code duplication in `handle_request_configure()` by simply calling `configure()`, as suggested by @johanmalm in #428.

Plus a couple of smaller fixes related to edge resistance.


src/resistance.c
```diff
@@ -61,6 +61,9 @@ resistance_move_apply(struct view *view, double *x, double *y)
 	}
 
 	wl_list_for_each(output, &server->outputs, link) {
+		if (!output->wlr_output->enabled)
```
Consolatis, line 64: Please always use curly braces like `if (..) {`, even if there is only a single statement.

src/resistance.c
```diff
@@ -129,6 +132,9 @@ resistance_resize_apply(struct view *view, struct wlr_box *new_view_geo)
 		return;
 	}
 	wl_list_for_each(output, &server->outputs, link) {
+		if (!output->wlr_output->enabled)
```
Consolatis, line 135: curly braces

src/xwayland.c
```diff
@@ -15,36 +15,22 @@ handle_commit(struct wl_listener *listener, void *data)
 	struct wlr_surface_state *state = &view->surface->current;
 	struct view_pending_move_resize *pending = &view->pending_move_resize;
 
-	bool move_pending = pending->update_x || pending->update_y;
-	bool size_changed = view->w != state->width || view->h != state->height;
-
-	if (!move_pending && !size_changed) {
+	if (view->w == state->width && view->h == state->height)
```
Consolatis, line 18: curly braces

src/xwayland.c
```diff
@@ -206,38 +206,25 @@ handle_set_class(struct wl_listener *listener, void *data)
...
-	view->pending_move_resize.width = geo.width;
-	view->pending_move_resize.height = geo.height;
-	wlr_xwayland_surface_configure(view->xwayland_surface, (int16_t)geo.x,
-	       (int16_t)geo.y, (uint16_t)geo.width, (uint16_t)geo.height);
-}
-
 static void
 move(struct view *view, double x, double y)
 {
+	assert(view && view->xwayland_surface);
```
Consolatis, line 212: Please use two asserts() here. If it ever triggers we won't be able to tell the difference based on the log otherwise.

src/cursor.c
```diff
@@ -160,6 +160,7 @@ process_cursor_resize(struct server *server, uint32_t time)
 	else if (server->resize_edges & WLR_EDGE_RIGHT)
```
Consolatis, line 160: I know this is not part of your changes but would you mind adding the curly braces here as well?

Consolatis (changes requested):
Thanks a lot for your work on this area of the code!

I didn't verify the logic itself but saw some code style based issues:

Consolatis:
Had a quick test and everything seems to work as expected, didn't see any regressions.

src/xwayland.c
jlindgren90: I think I will just leave out the `view &&` part -- this is called as `view->impl->move(view, ...` so we'd crash before this point anyway if `view == NULL`.

jlindgren90:
@Consolatis Thanks for testing and reviewing :) Code style issues should be all fixed now.

Consolatis (approved):

Consolatis:
LGTM but I want to give @johanmalm a chance to verify the logic so I am not merging this right away.

src/xwayland.c
```diff
@@ -134,6 +121,29 @@ handle_destroy(struct wl_listener *listener, void *data)
...
+	wlr_xwayland_surface_configure(view->xwayland_surface, geo.x, geo.y,
+				       geo.width, geo.height);
+
+	/* If not resizing, process the move immediately */
+	if (view->w == geo.width && view->h == geo.height) {
+		view->x = geo.x;
+		view->y = geo.y;
+
+		wlr_scene_node_set_position(&view->scene_tree->node, geo.x, geo.y);
+		ssd_update_geometry(view);
```
johanmalm, line 143: Do we need `ssd_update_geometry()` here?
If we're only moving, I think I'm right in saying that the ssd-nodes will move with the view-node.

johanmalm:
I like this refactoring + fixes a lot. It feels move obvious what the code is doing.
Just one questions regarding the use of `ssd_update_geometry()` in `configure()` when we're only moving.
I've tried without and it seems to work fine, but thought I'd check with you in case I'm missing something.

src/xwayland.c
jlindgren90: I expect you're right. Perhaps the `ssd_update_geometry()` in `view_move()` is also unnecessary?

jlindgren90: See 5d8eb02af6138a1a7d76229478f6aa1a204496d0

johanmalm: Nice. It'd be good to just get @Consolatis eyes on the `ssd_update_geometry()` bit.

Consolatis: We need to call `ssd_update_geometry()` for move as well because we adjust the invisible resize grab area on move in case they would be covering a layer-shell client. See
https://github.com/labwc/labwc/blob/06cc21664fd66a969012cb5da56a1bf32f1c0a1e/src/ssd/ssd.c#L189-L197

johanmalm: Thanks. Well remembered :smile: I even looked at that code an didn't clock it!

jlindgren90: Okay that simplifies things, the post-processing for move and move+resize is the same then. Thanks for checking!

jlindgren90:
Rebased to tidy up all the revisions. I think I've incorporated all the feedback up to this point.

johanmalm:
Very nice. Big thanks.

PR #454, opened 2022-07-17
main: Increase FD limit

Joshua-Ashton:
This defaults to 1024, which is tiny, but is a requirement
for processes using the deprecated `select` function.

We must reset this back whenever we fork to start a new process,
as this is inherited, and breaks applications using `select` otherwise.

### DIFF
--- a/src/common/fd_util.c
+++ b/src/common/fd_util.c
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+
+#include <sys/resource.h>
+#include <wlr/util/log.h>
+
+#include "common/fd_util.h"
+
+static struct rlimit original_nofile_rlimit = {0};
+
+void
+increase_nofile_limit(void)
+{
+	if (getrlimit(RLIMIT_NOFILE, &original_nofile_rlimit) != 0) {
+		wlr_log_errno(WLR_ERROR, "Failed to bump max open files limit: getrlimit(NOFILE) failed");
+		return;
+	}
+
+	struct rlimit new_rlimit = original_nofile_rlimit;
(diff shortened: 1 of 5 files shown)


src/common/fd_util.c
```diff
@@ -0,0 +1,39 @@
...
+	if (getrlimit(RLIMIT_NOFILE, &original_nofile_rlimit) != 0) {
+		wlr_log_errno(WLR_ERROR, "Failed to bump max open files limit: getrlimit(NOFILE) failed");
+		return;
+	}
+
+	struct rlimit new_rlimit = original_nofile_rlimit;
+	new_rlimit.rlim_cur = new_rlimit.rlim_max;
+	if (setrlimit(RLIMIT_NOFILE, &new_rlimit) != 0) {
+		wlr_log_errno(WLR_ERROR, "Failed to bump max open files limit: setrlimit(NOFILE) failed");
+
```
Consolatis, line 23: Empty line here can be removed

src/common/fd_util.c
```diff
@@ -0,0 +1,39 @@
...
+		wlr_log(WLR_INFO, "Running with %d max open files",
+            (int)original_nofile_rlimit.rlim_cur);
```
Consolatis, line 25: Mixed use of tab and spaces

Consolatis (commented):
Thanks!

I wonder if we need something similar before spawing XWayland but I \*think\* that one doesn't use `spawn_async_no_shell()` and instead happens somewhere within wlroots so it should inherit the upped limit.

Just some small code style based comments:

Joshua-Ashton:
> I wonder if we need to adjust this before spawing XWayland 

We do not.

johanmalm:
Thanks @Joshua-Ashton 

"Note to self" references:
- http://0pointer.net/blog/file-descriptor-limits.html
- https://github.com/swaywm/sway/pull/6629

PR #461, opened 2022-07-23
Changes to get screen cast portals working

Joshua-Ashton:
Sets the desktop name for the session and initializes some variables dbus/systemd need to call the right things.

With these changes, I can successfully screencast with OBS Studio, previously it would ask the global portal interface and there would be none.

It doesn't find portals itself to talk to like some other applications do.

### DIFF
--- a/docs/labwc.desktop
+++ b/docs/labwc.desktop
@@ -3,4 +3,4 @@ Name=labwc
 Comment=A wayland stacking compositor
 Exec=labwc
 Type=Application
-
+DesktopNames=wlroots
--- a/src/main.c
+++ b/src/main.c
@@ -92,6 +92,15 @@ main(int argc, char *argv[])
 	server_init(&server);
 	server_start(&server);
 
+	/* Initialize systemd and dbus with our relevant variables
+	 * which are needed for some desktop portal use.
+	 *
+	 * This fixes things such as OBS screen casting not working.
+	 * In the worst case, on non-systemd/dbus systems, this will just not do anything.
+	 */
+	system("systemctl --user import-environment DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP");
+	system("dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP");
+
 	struct theme theme = { 0 };
 	theme_init(&theme, rc.theme_name);
 	rc.theme = &theme;


kassindornelles:
merge! merge! merge! :tada:

acoolstraw:
the pain of xdg-desktop-portal will at last be lessened

droc12345:
You need to make sure it works for those not running systemd. 
I'm not so sure that using a system call on something that might not be there is a good idea.

If /run/systemd/system/ exists then systemd is in use, just check and bracket with an if.

Joshua-Ashton:
We don't need to check anything, it will just silently fail and continue on.

droc12345:
Why are you not running those commands from autostart?

Joshua-Ashton:
Because making users put stuff in their personal configs to make basic functionality that should work out of the box work is a bad thing.

Consolatis:
The change for `labwc.desktop` is completely fine with me. Could also be added to `docs/environment` (including a short comment stating the reason) so it also works for people not using a login manager.

Regarding the hardcoded (and blocking) `system()` calls in `main.c`: I am really not much a fan of this.
Even if the `system()` calls silently fail we still hard-code user configuration in labwc, the `autostart` file exists for this reason.

I'd be completely fine to add these two commands (and the corresponding comment) to `docs/autostart` and possibly to https://labwc.github.io/integration.html / https://labwc.github.io/getting-started.html though.

01micko:
If we're going yo use `system` why not check if we have `systemctl` available?

```c
  /* use type or command -v ? */
  if(system("type systemctl >/dev/null 2>&1") == 0) { 
    ...
  }
```

Joshua-Ashton:
@Consolatis This is not user configuration, this is completing basic setup of the session needed for things such as portals to work.

Other DEs like KDE and Gnome do essentially the same thing, and the Arch packages for Sway even inject a config to do this.

I am fine with changing it from system to something else that is non-blocking and cleaner.

Consolatis:
> Other DEs like KDE and Gnome do essentially the same thing

Labwc is not a desktop environment though, its a compositor.

For the same reason labwc doesn't depend on (or tries to start if available) `sway-bg`, `mako` or some random panel. You can do screen-recording / screenshots without using dbus / xdg-desktop-portal (e.g. via `wf-recorder` or `grim`) so if you prefer using xdg-desktop-portal (or applications which require it) its your choice as a user and thus in my opinion it is a user configuration and belongs into `autostart` / `environment`.

> and the Arch packages for Sway even inject a config to do this.

I assume the reason they do this is that sway itself doesn't want to hard-code it either.

Joshua-Ashton:
If you "prefer" to use applications that don't use it, you can just not use them. You are just arguing that some applications should just randomly be broken unless a user puts magic incantations to bother to finish setting up the session in their autostart.

Setting up your panels is absolutely a user configuration, but this is (and has already to me and others who use labwc) just going to be a useless footgun otherwise.

I agree the current implementation is slightly ugly but it is also simple. For the dbus-stuff, perhaps call it directly, if preferred, etc and use a non-blocking call for the systemd part.

Consolatis:
> this is (and has already to me and others who use labwc) just going to be a useless footgun otherwise.

I completely agree that this is a pain point and thus both commands / `XDG_CURRENT_DESKTOP=wlroots` should really be included in `docs/autostart` / `docs/environment` / the getting-started page on labwc.github.io but I still don't see a point in setting random stuff up via hard-coded commands in the compositor itself when the user might not even use dbus / systemd.

acoolstraw:
> You can do screen-recording / screenshots without using dbus / xdg-desktop-portal (e.g. via wf-recorder or grim) so if you prefer using xdg-desktop-portal (or applications which require it) its your choice as a user and thus in my opinion it is a user configuration and belongs into autostart / environment.

I disagree. Chromium and Flameshot for example require xdg-desktop-portal for screen sharing and taking screenshots. You could use wf-recorder or grim, but sometimes you need to use xdg-desktop-portal (for low-latency screen sharing)

I think if this doesn't impact anything (eg, if it just turns on something and if that something isn't there it just does nothing) it's alright, and would likely help people not waste hours attempting to get screen sharing/recording/shotting to work, and make Wayland more appealing (for example, my friends cite xdg-desktop-portal as the #1 frustration they have with Wayland)

Consolatis:
hm.. I still think that providing a proper example in `docs/autostart` / website is enough to fix this.

I am not completely against merging this when
- it doesn't cause any startup delay when running on a non-systemd system
- it doesn't cause any startup delay when running without dbus session
- @johanmalm decides it to be fine

kassindornelles:
Does anyone really cares about the one guy that will use labwc in a non-systemd system?

Consolatis:
> Does anyone really cares about the one guy that will use labwc in a non-systemd system?

yes

01micko:
> Does anyone really cares about the one guy that will use labwc in a non-systemd system?

I know at least 3 not including me and there will be a whole bunch more when I release...

kassindornelles:
> > Does anyone really cares about the one guy that will use labwc in a non-systemd system?
> 
> I know at least 3 not including me and there will be a whole bunch more when I release...

okay, so what would be a solution to get desktop-portal working without hurting the 4 guys using non-systemd systems?
is a huge pain in the ass to not have it working since this is the path linux distros are going for years now

Consolatis:
> okay, so what would be a solution to get desktop-portal working without hurting the 4 guys using non-systemd systems?

Add to `~/.config/labwc/environment`:
```shell
XDG_CURRENT_DESKTOP=wlroots
```

Add to `~/.config/labwc/autostart`
```shell
systemctl --user import-environment DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
```

Restart labwc.

kassindornelles:
> > okay, so what would be a solution to get desktop-portal working without hurting the 4 guys using non-systemd systems?
> 
> Add to `~/.config/labwc/environment`:
> 
> ```shell
> XDG_CURRENT_DESKTOP=wlroots
> ```
> 
> Add to `~/.config/labwc/autostart`
> 
> ```shell
> systemctl --user import-environment DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
> dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP
> ```
> 
> Restart labwc.

that can work, what you think @Joshua-Ashton?
If Josh agrees i guess just adding documentation for it would be enough.

droc12345:
A few comments.
1. autostart is where systemd/dbus/<other app> startup/configuration belongs.
2. it really is up to the end user to configure their own system, there is no one size fits all
3. things like screen-sharing/portals are not basic functionality (plenty of people don't need or use them)

Add the lines to autostart AND add a little description to either the generic doc/wiki and/or man pages mentioning this and what it's used for and why.

Joshua-Ashton:
> autostart is where systemd/dbus/ startup/configuration belongs.

Can you give a genuine reason other than inconveniencing users?

> it really is up to the end user to configure their own system, there is no one size fits all

This is not configuration.

> things like screen-sharing/portals are not basic functionality (plenty of people don't need or use them)

Then they wouldn't have them installed! This just makes them work, if they are installed.

Joshua-Ashton:
I am willing to compromise on doing this only when `XDG_CURRENT_DESKTOP` is set, (ie. launched as a session) and only when systemd is available in a non-blocking way.

johanmalm:
Thanks for all the contributions here. Good discussion, although I note a slight harshness in tone and politely ask everyone stay respectful at all times.

I’m okay with the proposed patch - although think spawn_async_no_shell() or similar is preferable.

My initial reaction was similar to that of @Consolatis thinking that this is the role of a session-file or autostart. I did even contemplate the openbox approach of loading both (rather than one or the other) system wide and user autostart, and then shipping a system autostart template with the systemd/dbus calls.

However, on consideration, I think Josh’s patch makes sense.

We want labwc to be easy to run without too much tweaking and reading of man pages. I also think it’s a shame when distros/OSs have to patch software - better to implement upstream as long as it does not create problems for others or result in technical debt which I don’t think it will in this case.

If we later regret the implementation we can move to a system-wide autostart approach without regression (except that a system wide installation will be required)

acoolstraw:
> Does anyone really cares about the one guy that will use labwc in a non-systemd system?

yes, I was the one guy

Consolatis:
Gave it a try:
- #466 

Differences to this PR:
- uses non-blocking `spawn_async_no_shell()` instead of `system()`
- uses `dbus-update-activation-environment` without the `--systemd` flag so it also works for users that use a dbus session but no systemd user session
- sets `XDG_CURRENT_DESKTOP` as environment variable even if started without a login manager

Same to this PR:
- Fails gracefully of there is no systemd / dbus session around.

Consolatis:
Implemented by
- #466

PR #463, opened 2022-07-23
xwayland: Process unmanaged client configure requests more quickly

jlindgren90:
This is a similar fix to 065c37d3f5ee but for unmanaged windows.  The issue
could be seen for example when moving the undocked Search Tool window in
Audacious.  For unmanaged windows, we don't track any pending move/resizes,
so just process all client configure requests immediately.

Video of the issue:
https://user-images.githubusercontent.com/1244737/180612621-461f4503-faa3-4a00-ab4b-411929985478.mp4


Joshua-Ashton (approved):
LGTM!

Consolatis:
Looks good to me as well, thanks.

PR #466, opened 2022-07-27
src/config/session.c: Update dbus / systemd activation environment

Consolatis:
This allows xdg-desktop-portal-wlr to work out of the box for screen-recording.
If systemd or dbus is not available the environment update will fail gracefully.

This patch will set `XDG_CURRENT_DESKTOP=wlroots` but a user may change this by
either having the environment variable set before starting labwc or by having
a different value set in `~/.config/labwc/environment`.

Based on PR #461 by @Joshua-Ashton

### DIFF
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -78,16 +78,51 @@ build_path(const char *dir, const char *filename)
 		return NULL;
 	}
 	int len = strlen(dir) + strlen(filename) + 2;
-	char *buffer = calloc(len, 1);
+	char *buffer = calloc(1, len);
 	strcat(buffer, dir);
 	strcat(buffer, "/");
 	strcat(buffer, filename);
 	return buffer;
 }
 
+static void
+update_activation_env(const char *env_keys)
+{
+	if (!getenv("DBUS_SESSION_BUS_ADDRESS")) {
+		/* Prevent accidentally auto-launching a dbus session */
+		wlr_log(WLR_INFO, "Not updating dbus execution environment: "
+			"DBUS_SESSION_BUS_ADDRESS not set");
+		return;
+	}
+	wlr_log(WLR_INFO, "Updating dbus execution environment");
+
+	char *cmd;
+	const char *dbus = "dbus-update-activation-environment ";
+	const char *systemd = "systemctl --user import-environment ";
+
+	cmd = calloc(1, strlen(dbus) + strlen(env_keys) + 1);
+	strcat(cmd, dbus);
+	strcat(cmd, env_keys);
+	spawn_async_no_shell(cmd);
+	free(cmd);
+
+	cmd = calloc(1, strlen(systemd) + strlen(env_keys) + 1);
+	strcat(cmd, systemd);
+	strcat(cmd, env_keys);
+	spawn_async_no_shell(cmd);
+	free(cmd);
+}
+
 void
 session_environment_init(const char *dir)
 {
+	/*
+	 * Set default for XDG_CURRENT_DESKTOP so xdg-desktop-portal-wlr is happy.
+	 * May be overriden either by already having a value set or by the user
+	 * supplied environment file.
+	 */
+	setenv("XDG_CURRENT_DESKTOP", "wlroots", 0);
+
 	const char *environment = build_path(dir, "environment");
 	if (!environment) {
 		return;
@@ -99,6 +134,9 @@ session_environment_init(const char *dir)
 void
 session_autostart_init(const char *dir)
 {
+	/* Update dbus and systemd user environment, each may fail gracefully */
+	update_activation_env("DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP");
+
 	const char *autostart = build_path(dir, "autostart");
 	if (!autostart) {
 		return;
@@ -109,7 +147,7 @@ session_autostart_init(const char *dir)
 	}
 	wlr_log(WLR_INFO, "run autostart file %s", autostart);
 	int len = strlen(autostart) + 4;
-	char *cmd = calloc(len, 1);
+	char *cmd = calloc(1, len);
 	strcat(cmd, "sh ");
 	strcat(cmd, autostart);
 	spawn_async_no_shell(cmd);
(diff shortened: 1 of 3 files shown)


Consolatis:
hm.. `dbus-update-activation-environment --verbose DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP` seems to actually start a dbus session for me when there is none available. This should not be the case! Maybe we could check for the existence of `DBUS_SESSION_BUS_ADDRESS`?

johanmalm:
Looks really good.

It’s tidy to explicitly set XDG_CURRENT_DESKTOP because I think a lot of user run labwc without a Display Manager.

Normally calloc(1,size) rather than calloc(size,1). Not sure if it makes a difference though.

Good catch regarding DBUS_SESSION_BUS_ADDRESS

Consolatis:
> Normally calloc(1,size) rather than calloc(size,1). Not sure if it makes a difference though.
> void *calloc(size_t nmemb, size_t size);

Uh, nice catch!
I had actually copied from another part of the same file, changed it (and the other occurrences as well).

src/config/session.c
```diff
@@ -78,16 +78,51 @@ build_path(const char *dir, const char *filename)
...
+		return;
+	}
+	wlr_log(WLR_INFO, "Updating dbus execution environment");
+
+	char *cmd;
+	const char *dbus = "dbus-update-activation-environment ";
+	const char *systemd = "systemctl --user import-environment ";
+
+	cmd = calloc(1, strlen(dbus) + strlen(env_keys) + 1);
+	strcat(cmd, dbus);
```
heroin-moose, line 104: Basically you are reimplementing `asprintf()` here.

Consolatis: > Basically you are reimplementing `asprintf()`

Yes, I am aware of this but:

> These functions are GNU extensions, not in C or POSIX.

> Also not checking for error code on calloc() is bad.

Agree'd. will change.

src/config/session.c
```diff
@@ -99,6 +134,9 @@ session_environment_init(const char *dir)
 void
 session_autostart_init(const char *dir)
 {
+	/* Update dbus and systemd user environment, each may fail gracefully */
+	update_activation_env("DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP");
```
heroin-moose, line 138: If you'll move the environment variables inside the function, you can do it with no allocations at all.

Consolatis: That is what I had at the beginning (2 straight copies of the env vars) but for maintainability reasons I think this is actually cleaner.

If we decide in the future to add more variables we just have to modify this single argument and everything else automatically works.

src/config/session.c
Consolatis: Changed for the whole file.

src/config/session.c
```diff
@@ -78,16 +78,51 @@ build_path(const char *dir, const char *filename)
...
+
 void
 session_environment_init(const char *dir)
 {
+	/*
+	 * Set default for XDG_CURRENT_DESKTOP so xdg-desktop-portal-wlr is happy.
+	 * May be overriden either by already having a value set or by the user
+	 * supplied environment file.
+	 */
+	setenv("XDG_CURRENT_DESKTOP", "wlroots", 0);
```
heroin-moose, line 135: Ain't `DesktopNames=wlroots` already handles that?

src/config/session.c
heroin-moose: Yeah, but this function is present on all BSDs as well.

src/config/session.c
Consolatis: Only if you actually use the `labwc.desktop` file with a login manager.

src/config/session.c
```diff
@@ -101,16 +104,24 @@ update_activation_env(const char *env_keys)
 	const char *systemd = "systemctl --user import-environment ";
 
 	cmd = calloc(1, strlen(dbus) + strlen(env_keys) + 1);
-	strcat(cmd, dbus);
-	strcat(cmd, env_keys);
-	spawn_async_no_shell(cmd);
-	free(cmd);
+	if (!cmd) {
+		wlr_log(WLR_ERROR, "Failed to allocate memory for dbus env update");
```
heroin-moose, line 108: IMO it's makes sense do it like this:

```
cmd = calloc(...);
if (!cmd) {
        oom(); // writes to log and exit(1)
}
```

or simply

```
cmd = xcalloc(...); // exits on allocation error
````

src/config/session.c
heroin-moose: Aha, got it.

src/config/session.c
Consolatis: But might not be in musl or other implementations, we try to not use any custom compiler extensions if at all possible.

src/config/session.c
Consolatis: As labwc is a compositor it will take down everything else (at least graphical applications) with it. But honestly if we can't even allocate a few bytes for a string lots of other places will break as well (likely including the `wlr_log()` statement).

heroin-moose: Exactly. It's better to exit early than to start with half-baked labwc.

johanmalm:
Whoos. That’s probably my original calloc(size,1). 😄 

Agree, best to avoid asprintf() unless we want lots of faff with supportability.

Considering we link with glib anyway, we could always use g_strdup_printf()

I’m pretty easy on what we do with handling calloc/strdup returns on such small + one-off allocations. Like Consolatis says, if we run out of memory, lots of this will break.
Just had a look at sway and they seem pretty good at always doing

`foo=calloc(); if (!foo) { wlr_log(); return/exit/goto}`

If we fancy it, we could add xcalloc() like this https://github.com/johanmalm/jgmenu/blob/1533e8def33bd3e1d7f4734206af4a66b51b9960/src/util.c#L91

Consolatis:
> Considering we link with glib anyway, we could always use g_strdup_printf()

Would not offer any benefit vs. `calloc()` -> `strcat()` though, same for `asprintf()`. We could also use the stack as those are all `const char *`. Or use `snprintf()`. Or `memcpy()`. Or do it in 10 other ways.

The whole discussion about quitting on allocation errors is interesting and could be extended to come to a general conclusion about the whole labwc codebase but feels like its a bit besides the point of this specific PR.

If there are no other issues I'd vote to rebase/merge this one so wlr-screencopy via xdg-desktop-portal works out-of-the-box.

PR #467, opened 2022-07-29
CI: prevent FreeBSD breakage via auto-update

jbeich:
Recommended since https://github.com/vmactions/freebsd-vm/compare/v0.2.1...v0.2.2


Consolatis:
Thanks for keeping an eye on the FreeBSD runner :)

PR #471, opened 2022-08-02
Submenu arrows

johanmalm:
@Consolatis I've had a go at rendering submenu-arrows for the root+context menus. Review with a view to merge appreciated if you have time.

### DIFF
--- a/src/common/font.c
+++ b/src/common/font.c
@@ -57,17 +57,6 @@ font_width(struct font *font, const char *string)
 }
 
 void
-font_buffer_update(struct lab_data_buffer **buffer, int max_width,
-	const char *text, struct font *font, float *color, double scale)
-{
-	if (*buffer) {
-		wlr_buffer_drop(&(*buffer)->base);
-		*buffer = NULL;
-	}
-	font_buffer_create(buffer, max_width, text, font, color, scale);
-}
-
-void
 font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	const char *text, struct font *font, float *color, double scale)
 {
--- a/src/common/font.c
+++ b/src/common/font.c
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
 #include <cairo.h>
 #include <drm_fourcc.h>
 #include <pango/pangocairo.h>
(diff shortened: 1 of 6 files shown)


src/common/font.c
```diff
@@ -56,33 +60,30 @@ font_width(struct font *font, const char *string)
...
-	font_buffer_create(buffer, max_width, text, font, color, scale);
-}
-
 void
 font_buffer_create(struct lab_data_buffer **buffer, int max_width,
-	const char *text, struct font *font, float *color, double scale)
+	const char *text, struct font *font, float *color, const char *arrow,
+	double scale)
 {
+	assert(max_width);
```
Consolatis, line 68: Does this still work with setting any of these theme `size` vars to `0`?
- `ActiveWindow`
- `MenuItem`
- `OSD`

src/menu/menu.c
```diff
@@ -137,10 +137,11 @@ item_create(struct menu *menu, const char *text)
 	menuitem->selected.text = &menuitem->selected.buffer->scene_buffer->node;
 
 	/* Font buffers */
+	const char *arrow = show_arrow ? "›" : NULL;
```
Consolatis, line 140: Does the arrow always fit into a single byte? I don't have much experience with UTF-8 const char assignments like this or wide chars / UTF storage in C in general.

Consolatis (commented):
~~Two questions /~~ One potential issue came up during review.

Didn't test but looks good to me otherwise.

src/common/font.c
```diff
@@ -56,33 +60,30 @@ font_width(struct font *font, const char *string)
...
 	}
 
-	PangoRectangle rect = font_extents(font, text);
-	if (max_width && rect.width > max_width) {
-		rect.width = max_width;
+	PangoRectangle text_extents = font_extents(font, text);
+	PangoRectangle arrow_extents = font_extents(font, arrow);
+
+	if (arrow) {
+		text_extents.width = max_width - arrow_extents.width;
```
Consolatis, line 78: This could turn negative if `max_width` is set to `1` or something small like that.
Currently this can't happen because the title doesn't use the arrow and the menu width is hard-coded.
It feels like a bug lingering around waiting to be triggered in the future with dynamic menu-width or similar though.

johanmalm: fixed

johanmalm:
Have force pushed fix

Consolatis:
Nice one, this feels very intuitive to me
menu
menu

Issue #474, opened 2022-08-08
Change default of A-F3 to "bemenu-run -n"


datMaffin:
I would recommend to change the `bemenu-run` run command to  `bemenu-run -n` to avoid the issue of overlapping UI.
In my case bemenu decided to spawn beneath waybar. I therefore feel like adding the `-n` switch will result in better default behavior.

Code: https://github.com/labwc/labwc/blob/32b176e05e0b7898028006123bfe1dad6d9b1082/src/config/rcxml.c#L506

Related: https://github.com/Cloudef/bemenu/issues/134

Consolatis:
Thanks for bringing this up. Personally I don't even have bemenu installed so can't test anything but created [a PR](https://github.com/labwc/labwc/pull/479) for others to test / accept.

johanmalm:
It is strange that bemenu is rendering behind waybar. The layer-surface created last should render on top (within each layer) because of how the nodes are added. Or am I missing something? Do we have to do a manual `wlr_scene_node_raise_to_top()` after creating  a new layer-surface?

Please could you confirm that:
- you start bemenu after waybar
- you run waybar in either `bottom` or `top` layer

I use both waybar and bemenu. With `bemenu-run` I get this:

![20220811_15h56m46s_grim](https://user-images.githubusercontent.com/6955353/184167208-8541b8b3-26df-4c6f-963d-ab4b0b49d6fa.png)

With `bemenu-run -n` I get:

![20220811_15h57m24s_grim](https://user-images.githubusercontent.com/6955353/184167349-71b0cfb8-8b6a-4274-bdab-5a114aa0fd4d.png)

I prefer the former, but am keen to understand if we have a problem with node z-order.

datMaffin:
Äähmm.. 

I just tested it again on the latest version packaged in the Distro and from master.

It looks like this is no longer an issue on the master! Sorry.

Master:
![image](https://user-images.githubusercontent.com/29998721/184180138-8fd0110d-6965-478b-ae94-2e9915557d84.png)

Consolatis:
Thanks for verifying :)

Closing this as requested.

PR #479, opened 2022-08-10
Use "bemenu-run -n" for the A-F3 default keybind to be on top of panels

Consolatis:
Fixes
- #474

I can't test this as I don't even have bemenu installed but the [man page](https://github.com/Cloudef/bemenu/blob/master/man/bemenu.1.scd.in#L103) for `-n` made sense.

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -92,7 +92,7 @@
       <action name="Execute"><command>alacritty</command></action>
     </keybind>
     <keybind key="A-F3">
-      <action name="Execute"><command>bemenu-run</command></action>
+      <action name="Execute"><command>bemenu-run -n</command></action>
     </keybind>
     <keybind key="A-F4">
       <action name="Close" />
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -503,7 +503,7 @@ static struct {
 	{ "A-Tab", "NextWindow", NULL },
 	{ "A-Escape", "Exit", NULL },
 	{ "W-Return", "Execute", "alacritty" },
-	{ "A-F3", "Execute", "bemenu-run" },
+	{ "A-F3", "Execute", "bemenu-run -n" },
 	{ "A-F4", "Close", NULL },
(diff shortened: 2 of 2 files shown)


datMaffin:
I tried to compile and run your branch ([Consolatis:issue/474_bemenu](https://github.com/Consolatis/labwc/tree/issue/474_bemenu)), but I ran into linking errors: 
```
/nix/store/rq6bh3qfrqnyqwik0w3q6w180zg3w2pa-binutils-2.38/bin/ld: labwc.p/src_seat.c.o: in function `new_pointer':
/home/user/Documents/labwc-consolatis/build/../src/seat.c:157: undefined reference to `wlr_pointer_from_input_device'
/nix/store/rq6bh3qfrqnyqwik0w3q6w180zg3w2pa-binutils-2.38/bin/ld: labwc.p/src_seat.c.o: in function `new_keyboard':
/home/user/Documents/labwc-consolatis/build/../src/seat.c:171: undefined reference to `wlr_keyboard_from_input_device'
/nix/store/rq6bh3qfrqnyqwik0w3q6w180zg3w2pa-binutils-2.38/bin/ld: labwc.p/src_seat.c.o: in function `new_touch':
/home/user/Documents/labwc-consolatis/build/../src/seat.c:189: undefined reference to `wlr_touch_from_input_device'
```

I do not really understand why, though. The branch labwc/labwc did compile successfully previously.
I would guess that wlroots-0.16.0-dev changed something like the naming of the now missing functions(?).

Anyway pretty sure that the commit looks good.

Consolatis:
That sounds like your wlroots-dev version is too old. So if updating that one doesn't work you could use something like this:
```shell
meson configure --force-fallback-for=wlroots build
meson subprojects update wlroots
meson compile -C build
```

datMaffin:
Can confirm that this change works as expected for me.

Consolatis:
Not changing the defaults: https://github.com/labwc/labwc/issues/474#issuecomment-1212128204

Issue #472, opened 2022-08-04
menu.xml: Allow root-menu to have no label


devosalain:
Right-mouse-click or Left-mouse-click only shows two things : 
-Reconfigure
-Exit
Although my menu.xml is:
``` 
<?xml version="1.0" encoding="UTF-8"?>

<openbox_menu>

<menu id="root-menu">
  <item label="Web browser">
    <action name="Execute"><command>firefox</command></action>
  </item>
  <item label="thunar">
    <action name="Execute"><command>thunar</command></action>
  </item>
  <item label="nemo">
    <action name="Execute"><command>nemo</command></action>
  </item>
  <item label="gtk-mixer">
    <action name="Execute"><command>gtk-mixer</command></action>
  </item>
  <item label="strawberry">
    <action name="Execute"><command>strawberry</command></action>
  </item>
  <item label="QTerminal">
    <action name="Execute"><command>qterminal</command></action>
  </item>
  <item label="xfce4-terminal">
    <action name="Execute"><command>xfce4-terminal</command></action>
  </item>
  <item label="Terminal">
    <action name="Execute"><command>kitty</command></action>
  </item>
  <item label="Reconfigure">
    <action name="Reconfigure"></action>
  </item>
  <item label="Exit">
    <action name="Exit"></action>
  </item>
  <item label="Poweroff">
    <action name="Execute"><command>systemctl -i poweroff</command></action>
  </item>
</menu>

</openbox_menu>
``` 

My O.S. is FreeBSD 13.1-RELEASE

Consolatis:
Your `menu.xml` looks fine on the first look so it appears labwc is not picking it up correctly.
Is it available as `~/.config/labwc/menu.xml`?

devosalain:
Yes it is.

Consolatis:
Just tested your `menu.xml`, the root-menu misses a label (which can be empty). So something like this should work:
`<menu id="root-menu" label="">`

devosalain:
Indeed, that fixed the problem.

johanmalm:
`(label && id)` creates inline menu

https://github.com/labwc/labwc/blob/ac5c9dae00fab736b81260df0c7247beb6478933/src/menu/menu.c#L317

`id` only creates an submenu entry point to a menu defined elsewhere.

https://github.com/labwc/labwc/blob/ac5c9dae00fab736b81260df0c7247beb6478933/src/menu/menu.c#L339

[...]

PR #481, opened 2022-08-11
menu: do not require label attribute for highest level menu definition

johanmalm:
Allow highest level menu definitions - typically used for root-menu and
client-menu - to be defined like this:

    <openbox_menu>
      <menu id="">
      </menu>
    </openbox>

Previously this required a label attribute (which was not used for
anything and could be an empty string) as show below:

    <openbox_menu>
      <menu id="" label="">
      </menu>
    </openbox>

Closes issue #472

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -51,7 +51,7 @@ menu_create(struct server *server, const char *id, const char *label)
 	nr_menus++;
 	wl_list_init(&menu->menuitems);
 	menu->id = strdup(id);
-	menu->label = strdup(label);
+	menu->label = label ? strdup(label) : strdup(id);
 	menu->parent = current_menu;
 	menu->server = server;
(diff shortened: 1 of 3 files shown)


src/menu/menu.c
```diff
@@ -314,7 +325,23 @@ handle_menu_element(xmlNode *n, struct server *server)
...
+		 * inline menu definition.
+		 *
+		 * (id && nr_parents(n) == 2) refers to:
+		 * <openbox_menu>
+		 *   <menu id="">
+		 *   </menu>
+		 * </openbox>
+		 *
+		 * which is the highest level a menu can be defined at.
+		 * Whist openbox spec stricly speaking wants a label="" defined
```
Consolatis, line 340: spelling

Otherwise: LGTM. Not tested though.

johanmalm: Thanks :smile:

Issue #483, opened 2022-08-12
Partial motion event outside the surface when the pointer is grabbed - No event at all with layer-shell


Tamaranch:
Labwc 0.5.3 on Arch Linux

Here is the code to reproduce the problem:

<p>
<details>
<summary>p.c</summary>

```c
#include <gtk/gtk.h>
#include <gtk-layer-shell/gtk-layer-shell.h>

static gboolean
on_button_press (GtkWidget *widget,
                 GdkEvent  *event,
                 gpointer   user_data)
{
  g_printerr ("%s\n", G_STRLOC);

  return FALSE;
}

static gboolean
on_button_release (GtkWidget *widget,
                   GdkEvent  *event,
                   gpointer   user_data)
{
  g_printerr ("%s\n", G_STRLOC);

  return FALSE;
}

static gboolean
on_motion_notify (GtkWidget      *widget,
                  GdkEventButton *event,
                  gpointer        user_data)
{
  g_printerr ("%s:x: %f\n", G_STRLOC, event->x);
  g_printerr ("%s:y: %f\n", G_STRLOC, event->y);

  return FALSE;
}

gint main (gint argc, gchar **argv)
{
  GtkWidget *window;
  GMainLoop *loop;
  GtkWidget *widget;

  gtk_init (&argc, &argv);

  window = gtk_window_new (GTK_WINDOW_TOPLEVEL);
  widget = gtk_button_new_with_label ("test");
  gtk_container_add (GTK_CONTAINER (window), widget);
  g_signal_connect (widget, "button-press-event", G_CALLBACK (on_button_press), window);
  g_signal_connect (widget, "button-release-event", G_CALLBACK (on_button_release), window);
  g_signal_connect (widget, "motion-notify-event", G_CALLBACK (on_motion_notify), window);

  if (g_strcmp0 (argv[1], "--layer") == 0)
    gtk_layer_init_for_window (GTK_WINDOW (window));

  gtk_widget_show (widget);
  gtk_widget_show (window);

  loop = g_main_loop_new (NULL, FALSE);
  g_signal_connect_swapped (window, "destroy", G_CALLBACK (g_main_loop_quit), loop);
  g_main_loop_run (loop);
  g_main_loop_unref (loop);
}
```

</details>
</p>

It is built by
```sh
gcc p.c -o p $(pkg-config --cflags --libs gtk+-3.0 gtk-layer-shell-0)
```

Run by
```sh
./p [--layer]
```
to activate or not layer-shell.

[...]

PR #485, opened 2022-08-13
cursor: Do not clamp motion coordinates for XWayland surfaces.

jlindgren90:
This is a follow-on to #453 and continues the effort to make
client-side move/resize of XWayland apps work properly.

X11 apps expect to be able to receive motion events outside
the window area; therefore, for XWayland surfaces, we should
not "clamp" the motion coordinates to within the surface.

Before this change, attempting to enlarge an XWayland window
using a client-side resize handle resulted in the window size
lagging behind the mouse cursor quite severely, since each
motion event was in effect allowed to expand the window by
only a few pixels.  The closer the initial button-press was
to the edge of the window, the worse the lag would be.

The discussion in #483 seems related, but centers around
native Wayland clients, with which I am less familiar.  For
XWayland surfaces, I believe the problem and solution are
pretty clear and simple.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -95,8 +95,8 @@ struct seat {
 	struct wlr_layer_surface_v1 *focused_layer;
 
 	/**
-	 * pressed view/surface will usually be NULL and is only set on button
-	 * press while the mouse is over a view surface and reset to NULL on
+	 * pressed view/surface/node will usually be NULL and is only set on
+	 * button press while the mouse is over a surface and reset to NULL on
 	 * button release.
 	 * It is used to send cursor motion events to a surface even though
 	 * the cursor has left the surface in the meantime.
@@ -106,6 +106,7 @@ struct seat {
 	 */
 	struct {
 		struct view *view;
+		struct wlr_scene_node *node;
 		struct wlr_surface *surface;
 	} pressed;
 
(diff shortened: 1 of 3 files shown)


Consolatis:
If this still allows X11 applications to select text / use the scrollbar correctly when the mouse is outside of the surface this is fine with me. I am currently a bit short on time so I didn't test it myself. Thanks for your continued work in this area, much appreciated.

jlindgren90:
Based on a quick test with xfce4-terminal (`GDK_BACKEND=x11`), scrolling still works fine.
https://user-images.githubusercontent.com/1244737/184510910-c200fc19-6b3b-4a47-9595-68b8bbc43615.mp4

johanmalm:
Yes, thanks.
I’ll do some select/scroll testing when I get near a computer (probably not until Monday) and then merge.

Consolatis:
Added a follow-up commit to this PR which adds layershell clients to the out-of-surface motion variants.
There should be no changes in behavior in regards to XDG and XWayland surfaces.

johanmalm:
Thanks both!
I've tested with leafpad (xwayland) and mousepad (xdg). Select text + scroll bar work well for both when cursor moves off surface.
Layer surface fix is great. I tested with #483 p.c test program by @Tamaranch and it seems to work well.

PR #488, opened 2022-08-17
func output_destroy_notify add free

sosaxu:
function output_destroy_notify lack free operation


Consolatis:
Not sure how that could slip through all the time.

Thanks!

PR #489, opened 2022-08-17
seat: Implement virtual pointer and keyboard protocols

Joshua-Ashton:
Tested with [wayvnc](https://github.com/any1/wayvnc).

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -41,6 +41,7 @@
 #include <wlr/types/wlr_xdg_shell.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
 #include <wlr/types/wlr_virtual_pointer_v1.h>
+#include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include <wlr/util/log.h>
 #if HAVE_XWAYLAND
 #include <wlr/xwayland.h>
@@ -149,6 +150,9 @@ struct seat {
 
 	struct wlr_virtual_pointer_manager_v1 *virtual_pointer;
 	struct wl_listener virtual_pointer_new;
+
+	struct wlr_virtual_keyboard_manager_v1 *virtual_keyboard;
+	struct wl_listener virtual_keyboard_new;
 };
 
 struct lab_data_buffer;
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -40,6 +40,7 @@
 #include <wlr/types/wlr_xdg_decoration_v1.h>
 #include <wlr/types/wlr_xdg_shell.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
+#include <wlr/types/wlr_virtual_pointer_v1.h>
 #include <wlr/util/log.h>
 #if HAVE_XWAYLAND
 #include <wlr/xwayland.h>
@@ -145,6 +146,9 @@ struct seat {
 	struct wl_listener destroy_drag;
 	struct wl_listener constraint_commit;
 	struct wl_listener idle_inhibitor_create;
+
+	struct wlr_virtual_pointer_manager_v1 *virtual_pointer;
+	struct wl_listener virtual_pointer_new;
 };
 
 struct lab_data_buffer;
--- a/src/seat.c
+++ b/src/seat.c
@@ -221,13 +221,23 @@ seat_update_capabilities(struct seat *seat)
 }
 
 static void
+seat_add_device(struct seat *seat, struct input *input)
+{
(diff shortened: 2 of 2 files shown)


johanmalm:
Thanks Josh.
I tried [wtype](https://github.com/atx/wtype) which now works well except that I don't get the normal keyboard back. For example if I issue `wtype foo` in the terminal it prints foo, but my 'normal' keyboard then does work.
In sway I get my 'normal' keyboard back.

Joshua-Ashton:
Thanks, can reproduce. Not sure why yet though.

Joshua-Ashton:
It seems merely adding a keyboard to a keyboard group makes the other one not work, even after it's removed :l

Joshua-Ashton:
Oh huh wtf. I do that, then type with the number keys and it gives random letters... May be keymap related.

Joshua-Ashton:
I think wtype may set a keymap for the characters it presses, because with `wtype test`: esc = t  1-> e 2 -> s

Joshua-Ashton:
wlroots does not seem to reset the keymap on the global keyboard of the keyboard group based on the active one.

This smells like a bug and something it should be doing.

Consolatis:
Great stuff! Thanks for working on this.

01micko:
Thanks from me too :-)

Consolatis:
Had a quick nested test via
`WLR_NO_HARDWARE_CURSORS=1 WLR_RENDERER=pixman build_master/labwc -s foot` + `wayvnc` and connecting via `remmina`:
- the keyboard layout seems to be reset immediately (was `XKB_DEFAULT_LAYOUT=de` and once wayvnc starts up `y` and `z` values change)
- mouse scrolling (at least via remmina) doesn't seem work

johanmalm:
I'm not sure what the answer is here, but note the following:

Sway says that virtual devices should not be grouped
https://github.com/swaywm/sway/blob/6e023257e8e9a6673fc3ff69048cdaf8baa70044/sway/input/keyboard.c#L874

The commit when we added keyboard_group
https://github.com/labwc/labwc/commit/25829d122c2fd530df12d7383ad48259717a000e

Before that we had a list of `struct keyboard` in `struct server` with their own respective key/modifiers listeners.
https://github.com/labwc/labwc/blob/a4c22f7c4de1319e1e252d4a1d65d033fcb93ef2/include/labwc.h#L183

Maybe easier to add/remove keyboards that way?
Not sure what keymap should be used for a virtual keyboard, but probably easier to set that way too.
Sorry - didn't really consider all this when moving to `keyboard_group`

johanmalm:
How we used to do wlr_seat_set_keyboard() a looong time ago:
https://github.com/labwc/labwc/blob/1721b339da73c29fdadbc0022e7ec614defb4ac7/src/keyboard.c#L82

How we do it now (passing `&seat->keyboard_group->keyboard`)
https://github.com/labwc/labwc/blob/52d1e32b59a2bc3e5222ff2aa1102dbe910c4127/src/keyboard.c#L212

I might be barking up the wrong tree, but suspect we would need to use the old approach or similar.

Consolatis:
> Not sure what keymap should be used for a virtual keyboard, but probably easier to set that way too.

At least wayvnc has a command line parameter for a keyboard layout, so it might be set by the virtual keyboard protocol (which seems to work, but sets the layout for the whole group as mentioned). Did not verify this at all though. VNC clients also have a keyboard layout setting per connection but I am not sure how that fits into the picture.

Consolatis:
> mouse scrolling (at least via remmina) doesn't seem to work but works fine with sway

Turns out to be a wlroots issue: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3707
Must have tested with my distro maintained sway version using wlroots 0.15.1 which doesn't have that issue, latest sway (using latest wlroots) doesn't scroll either.

Or rather: both, labwc and latest sway scroll just fine when you scroll 120 times. Gets tiring fast though.

johanmalm:
I have drafted a patch that fixes this. I’ll tidy it up and push for review tomorrow. Just to make sure nobody is duplicating effort.

Joshua-Ashton:
Go for it, thanks :)

johanmalm:
https://github.com/johanmalm/labwc/tree/virtual-devices

Works well for me, but would appreciate some feedback before I PR + merge.

Consolatis:
Looks good to me, didn't test. Will do that on the new PR.

johanmalm:
Covered by PR #570

PR #490, opened 2022-08-17
view: view_move() s/double/int/ for x and y

johanmalm:
@Consolatis @jlindgren90 Can you think of any reasons why we should stick with `double`? Everything else related seems to be int. Even `struct wlr_xwayland_surface_configure_event` uses `int16_t` for x and y.


Consolatis:
Cursor coordinates are in double AFAIR (e.g. `process_cursor_move()`).
But automatic downcasting should be fine.

PR #493, opened 2022-08-21
osd+keyboard: Two small Alt-Tab fixes

jlindgren90:
Two separate fixes for the Alt-Tab (window switcher) OSD, short and sweet. Hopefully I didn't miss anything.

- Fix extra Alt-Tab OSD being shown if there is a disabled output (e.g. laptop display when docked)
- Fix Alt-Tab OSD remaining displayed after pressing Alt-Tab then Escape

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -151,6 +151,7 @@ handle_compositor_keybindings(struct wl_listener *listener,
 				if (syms[i] == XKB_KEY_Escape) {
 					/* cancel */
 					server->cycle_view = NULL;
+					osd_finish(server);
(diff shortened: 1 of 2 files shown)


Consolatis:
Both fixes look good to me. Thanks.

> Fix Alt-Tab OSD remaining displayed after pressing Alt-Tab then Escape

That fix also exists within the [`A-Tab` live preview  PR](https://github.com/labwc/labwc/pull/307) (which btw would benefit from some people testing it).
I am completely fine with applying it here though, I'll rebase #307 once this is merged.

PR #498, opened 2022-08-24
Add config option to disable preview outlines

Flrian:
Feel free to close if you guys decide its not quite what you want or prefer something else! :)


Consolatis:
Code looks great to me. Thanks!

@johanmalm up to you if we want to merge this config option, either
- permanently
- only temporarily until we extend the theme + have the theme-override implemented
- not at all

I personally would vote for either permanently or at least temporarily.

PR #504, opened 2022-08-26
view/ssd: Handle view_set_activated(false) case correctly

### LINKED ISSUE #494: Incorrect handling of floating XWayland windows (multiple issues)
### Issue 1 - incorrect title bar highlight

Steps to reproduce:

- From an XWayland terminal window (I used xfce4-terminal, but I don't think it
  matters which), run `audacious` (in Qt 5 mode).
- Activate the Search Tool pane (use the "search" icon in the toolbar) if not
  already active.
- Undock the Search Tool pane (click the "restore" icon in the top-right corner
  of the pane, next to the "close" icon).

Expected result:

- The title bar of the main Audacious window remains highlighted as the active
  window.

Actual result:

- The title bar of some other, unrelated top-level window (sometimes the
  terminal window, sometimes the window of a third, completely unrelated
  application) gets the "active window" highlight color.
- However, key-press events are passed to the floating Search Tool pane (as
  expected, but contrary to what the highlighted title bar on the other window
  would suggest).

### Issue 2 - incorrect mouse event handling

Steps to reproduce:

- Start `kdiff3`.
- Right-click the toolbar, and un-check "Lock Toolbar Positions".
- Drag the toolbar to a floating position in front of the main kdiff3 window.

Expected result:

- Buttons on the toolbar can be clicked.

Actual result:

- Mouse clicks apparently "pass through" the toolbar to the main kdiff3 window
  below it.  Only if the main window is moved away, out from behind the toolbar,
  is it then possible to click buttons on the toolbar.
- Moving another, unrelated top-level window (such as a terminal window)
  underneath the kdiff3 toolbar also results in mouse clicks "passing through"
  the toolbar to that window.

jlindgren90:
`view_set_activated()` takes a Boolean parameter `activated`.
`ssd_set_active()` needs a Boolean parameter as well, otherwise `view_set_activated(false)` results in calling `_ssd_set_active(true)`.

Partially fixes #494.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -188,7 +188,7 @@ struct server {
 	uint32_t resize_edges;
 
 	/* SSD state */
-	struct view *ssd_focused_view;
+	struct view *focused_view;
 	struct ssd_hover_state ssd_hover_state;
 
 	/* Tree for all non-layer xdg/xwayland-shell surfaces */
@@ -408,7 +408,7 @@ struct xwayland_unmanaged *xwayland_unmanaged_create(struct server *server,
 void unmanaged_handle_map(struct wl_listener *listener, void *data);
 #endif
 
-void view_set_activated(struct view *view, bool activated);
+void view_set_activated(struct view *view);
 void view_close(struct view *view);
 
 /**
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -134,7 +134,7 @@ struct ssd_hover_state {
 
 /* Public SSD API */
 void ssd_create(struct view *view);
-void ssd_set_active(struct view *view);
+void ssd_set_active(struct view *view, bool active);
 void ssd_update_title(struct view *view);
 void ssd_update_geometry(struct view *view);
 void ssd_reload(struct view *view);
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -87,18 +87,6 @@ desktop_move_to_back(struct view *view)
 	wl_list_insert_tail(&view->server->views, &view->link);
 }
 
-static void
-deactivate_all_views(struct server *server)
-{
-	struct view *view;
-	wl_list_for_each (view, &server->views, link) {
-		if (!view->mapped) {
-			continue;
-		}
-		view_set_activated(view, false);
-	}
-}
-
 void
 desktop_arrange_all_views(struct server *server)
 {
@@ -151,8 +139,7 @@ desktop_focus_and_activate_view(struct seat *seat, struct view *view)
 		return;
 	}
 
-	deactivate_all_views(view->server);
-	view_set_activated(view, true);
+	view_set_activated(view);
 	seat_focus_surface(seat, view->surface);
 }
 
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -140,21 +140,15 @@ ssd_resize_edges(enum ssd_part_type type)
 	return 0;
 }
 
-static void
-_ssd_set_active(struct ssd *ssd, bool active)
-{
-	wlr_scene_node_set_enabled(&ssd->border.active.tree->node, active);
-	wlr_scene_node_set_enabled(&ssd->titlebar.active.tree->node, active);
-	wlr_scene_node_set_enabled(&ssd->border.inactive.tree->node, !active);
-	wlr_scene_node_set_enabled(&ssd->titlebar.inactive.tree->node, !active);
-}
-
 void
 ssd_create(struct view *view)
 {
+	bool is_active = view->server->focused_view == view;
+
 	if (view->ssd.tree) {
 		/* SSD was hidden. Just enable it */
 		wlr_scene_node_set_enabled(&view->ssd.tree->node, true);
+		ssd_set_active(view, is_active);
 		return;
 	}
 
@@ -164,6 +158,7 @@ ssd_create(struct view *view)
 	ssd_border_create(view);
 	ssd_titlebar_create(view);
 	view->margin = ssd_thickness(view);
+	ssd_set_active(view, is_active);
 }
 
 void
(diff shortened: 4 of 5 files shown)


Consolatis:
I had a quick look at the whole implementation and discovered that we have two different code paths for activating / deactivating views:
- the one being worked on here (`view_set_activated()` in `src/view.c`)
- `deactivate_all_views()` in `src/desktop.c` (which also calls the first one)

Personally I'd like to get rid of `deactivate_all_views()` because there can only be one single activated view at any point in time anyway. Thus we don't have to loop over all views to deactivate them and instead just use the pointer to deactivate the actual active view before switching to the new view.

When removing `deactivate_all_views()` CSD clients are not handled properly though.
If the consensus is to keep the loop we can remove `server->ssd_focused_view` pointer instead.

Putting the PR into a work-in-progress for now until we come to an conclusion of the way forward.

jlindgren90:
Thanks for looking at it. Sounds like there is definitely some room for cleanup. I don't really have an opinion as to the best solution between the options you described.

Given that the current behavior is functionally wrong, would it be worth merging my patch as a stop-gap fix, with a suitable TODO comment that it should be cleaned up further?

jlindgren90:
Or the alternative, minimally-intrusive fix for the backwards logic:
```
index a8c63b0..489c5c4 100644
--- a/src/view.c
+++ b/src/view.c
@@ -98,7 +98,7 @@ view_get_edge_snap_box(struct view *view, struct output *output,
 void
 view_set_activated(struct view *view, bool activated)
 {
-       if (view->ssd.enabled) {
+       if (activated && view->ssd.enabled) {
                ssd_set_active(view);
        }
        if (view->impl->set_activated) {
```

Consolatis:
Would you mind giving https://github.com/Consolatis/labwc/tree/fix/ssd_active a try?

jlindgren90:
> Would you mind giving https://github.com/Consolatis/labwc/tree/fix/ssd_active a try?

It seems to fix the titlebar issue, but I encountered a problem with window ordering in the Alt-Tab switcher:
1. Start Audacious (main Audacious window has focus)
2. Pop out Search Tool (main Audacious title bar turns inactive color)
3. Alt-Tab once back to the main Audacious window (main Audacious title bar turns active color)

At this point, if I press Alt-Tab again, the first window in the list is the main Audacious window even though it's already active.  (In other words, pressing Alt-Tab once and releasing it would do nothing.)

jlindgren90:
The Alt-Tab issue seems to be pre-existing, for what it's worth, so maybe not directly related.

Consolatis:
> The Alt-Tab issue seems to be pre-existing

Yeah, that sounds like a different issue.

Are you fine with the commit of the ssd_active branch as such? I rebased my changes on top of yours so its your original commit with rewritten message + added `Co-Authored-by Consolatis`.

johanmalm:
I like the look of the combined commit here (https://github.com/Consolatis/labwc/tree/fix/ssd_active)

A few of thoughts:

- Should we rename `server->ssd_active_view`? Maybe just drop the ssd prefix.
- I like the idea of removing the true/false in view_set_activated(). Not sure why it was written with a bool argument in the first place :smile:  I checked foreign.c and it seems request_activate is always for 'true'
- In `view_minimize()`, do we not need view_set_activated(view, false); Can't get my head around how it works without that.

johanmalm:
> Personally I'd like to get rid of `deactivate_all_views()` because there can only be one single activated view at any point in time anyway.

Agree. Good spot and good proposed approach.

johanmalm:
I've tested https://github.com/Consolatis/labwc/tree/fix/ssd_active from tty. Seems to work well with both SSD and CSD. No objections to merging.

jlindgren90:
Looks good to me, and feel free to force-push to this branch 👍 Thanks!

Consolatis:
> A few of thoughts:
> 
> * Should we rename `server->ssd_active_view`? Maybe just drop the ssd prefix.

Changed to `server->focused_view`.

> * In `view_minimize()`, do we not need view_set_activated(view, false); Can't get my head around how it works without that.

Had to debug that one myself when writing the patch:
We are doing an unmap on minimize, that in turn causes a  `desktop_focus_topmost_mapped_view()` within the unmap handlers in `xdg.c` and `xwayland.c`. Which in turn calls `desktop_focus_and_activate_view()` which then finally calls `view_set_activated()` with the new view.

johanmalm:
> > A few of thoughts:
> > 
> > * Should we rename `server->ssd_active_view`? Maybe just drop the ssd prefix.
> 
> Changed to `server->focused_view`.
> 
> > * In `view_minimize()`, do we not need view_set_activated(view, false); Can't get my head around how it works without that.
> 
> Had to debug that one myself when writing the patch: We are doing an unmap on minimize, that in turn causes a `desktop_focus_topmost_mapped_view()` within the unmap handlers in `xdg.c` and `xwayland.c`. Which in turn calls `desktop_focus_and_activate_view()` which then finally calls `view_set_activated()` with the new view.

Thanks. Makes sense.

Ready for merge?

Consolatis:
Replaced the whole of `desktop_focused_view()` with a simple `return server->focused_view` while at it.

> Ready for merge?

If the compile tests go though it seems ready to me, yes.

johanmalm:
> Replaced the whole of `desktop_focused_view()` with a simple `return server->focused_view` while at it.

Oh... will `server->focused_view` always be in sync with the associated `keyboard_state.focused_surface`? Wlroots sets `keyboard_state.focused_surface` internally.

Consolatis:
> > Replaced the whole of `desktop_focused_view()` with a simple `return server->focused_view` while at it.
> 
> Oh... will `server->focused_view` always be in sync with the associated `keyboard_state.focused_surface`? Wlroots sets `keyboard_state.focused_surface` internally.

It should be.. I think the only exception could be if we are minimizing the last mapped view. `desktop_focus_and_activate_view()` would then get called with a `NULL` view which resets the keyboard focus to `NULL` but doesn't reset `server->focused_view`.

That function is only used for cursor constrains though so I think its pretty much impossible to see them go out of sync because for the cursor constrains to kick in the mouse button would have to pressed on top of a surface which also gives keyboard focus.

I'll remove that part for now because its not strictly correlated to the actual issue and should have been its own commit in the first place.

PR #505, opened 2022-08-26
view: Do not call ssd_set_active() for inactive views

### LINKED ISSUE #494: Incorrect handling of floating XWayland windows (multiple issues)

jlindgren90:
Partially fixes #494.

A minimally-invasive fix for the obvious logic error in `view_set_active()`.  I would request that this be merged as at least a stop-gap fix until the best long-term solution is identified (see discussion in #504).

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -98,7 +98,7 @@ view_get_edge_snap_box(struct view *view, struct output *output,
 void
 view_set_activated(struct view *view, bool activated)
 {
-	if (view->ssd.enabled) {
+	if (activated && view->ssd.enabled) {
 		ssd_set_active(view);
 	}
 	if (view->impl->set_activated) {


johanmalm:
I sense we're pretty close to a 504 way forward. Happy to hold on for a minute?

jlindgren90:
Yep, #504 looks good with @Consolatis 's changes 👍

johanmalm:
Assume this is now overtaken by events (#504)

PR #514, opened 2022-08-30
src/xwayland-unmanaged.c: Give keyboard focus back to topmost mapped view

Consolatis:
Previously, when unmapping and not finding another unmanaged surface
to give keyboard focus to we were not doing anything. With this patch
we will give focus to the topmost mapped view which is the same
behavior that also happens when an usual X11 surface unmaps.

This fixes dmenu when being exited with ESC.

Reported-by: @Flrian


johanmalm:
Nice one.

PR #515, opened 2022-08-31
[wip] Preliminary support for overwriting subpixel rendering

Consolatis:
Reported-by: @acoolstraw

Missing:
- [ ] docs
- [ ] use actual subpixel rendering names for the config instead of using raw enum based integers
- [ ] use a better config option name, "subpixel" doesn't really say all that much
- [ ] turn log message into debug
- [ ] create a copy of the enum so we don't have to depend on wayland-protocols
- [ ] wire up for `Reconfigure`

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -11,6 +11,7 @@
     <decoration>server</decoration>
     <gap>0</gap>
     <adaptiveSync>no</adaptiveSync>
+    <subpixel>0</subpixel>
     <cycleViewPreview>no</cycleViewPreview>
     <cycleViewOutlines>yes</cycleViewOutlines>
   </core>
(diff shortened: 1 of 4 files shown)


johanmalm:
Should we set this for cairo contexts too?
https://github.com/swaywm/sway/blob/d45623c2db570b7a2e0582943734d18f3d64b126/common/cairo.c#L13

Consolatis:
I think we may have different subpixel settings for each output so usually that would be an issue.
But as this PR actually overwrites all of the output defined ones if set to != 0 we could in fact do that in that case.

Consolatis:
This PR was done in an attempt to help debug some issue.
If somebody else wants to pick up from here feel free.

PR #517, opened 2022-09-01
xwayland: Fix stacking order of unmanaged surfaces

jlindgren90:
1. Implement `unmanaged_handle_request_activate()`.
2. Make sure (somewhat crudely) that unmanaged surfaces remain on top of the X11 window stack. I'm open to suggestions on how to improve this logic, especially to make sure that the X11 stacking order and the scene graph stacking order remain consistent with each other.

This fixes the second issue (mouse clicks passing through the `kdiff3` toolbar) described in #494.


src/xwayland-unmanaged.c
```diff
@@ -119,7 +119,11 @@ unmanaged_handle_override_redirect(struct wl_listener *listener, void *data)
 static void
 unmanaged_handle_request_activate(struct wl_listener *listener, void *data)
 {
-	wlr_log(WLR_DEBUG, "request_activate not handled\n");
+	struct xwayland_unmanaged *unmanaged =
+		wl_container_of(listener, unmanaged, request_activate);
+	wlr_xwayland_surface_activate(unmanaged->xwayland_surface, true);
+	wlr_xwayland_surface_restack(unmanaged->xwayland_surface, NULL,
+				     XCB_STACK_MODE_ABOVE);
```
Consolatis, line 126: Didn't think this through but you may be able to do something like this here:
```c
wl_list_remove(&unmanaged->link);
wl_list_insert(unmanaged->server->unmanaged_surfaces.prev, &unmanaged->link);
```
This should make it synchronized within Xwayland at least.
Regarding scene-graph I'll have a look at this later again.

jlindgren90: I like that; then I think we could just update the scene-graph so (not tested):
```
if (unmanaged->node) {
	wlr_scene_node_raise_to_top(unmanaged->node);
}
```

Consolatis: yes, that sounds like a good solution

jlindgren90: Hmm, in testing, `unmanaged_handle_request_activate()` is called before `unmanaged_handle_map()` and so the `wl_list_remove()` call segfaults.

src/xwayland-unmanaged.c
```diff
@@ -119,7 +119,11 @@ unmanaged_handle_override_redirect(struct wl_listener *listener, void *data)
 static void
 unmanaged_handle_request_activate(struct wl_listener *listener, void *data)
 {
-	wlr_log(WLR_DEBUG, "request_activate not handled\n");
+	struct xwayland_unmanaged *unmanaged =
+		wl_container_of(listener, unmanaged, request_activate);
+	wlr_xwayland_surface_activate(unmanaged->xwayland_surface, true);
```
Consolatis, line 124: Do we have to check for `unmanaged->xwayland_surface` here? Not sure what happens on unmap.

src/xwayland-unmanaged.c
jlindgren90: v2 avoids the `unmanaged_handle_request_activate()` questions by not implementing it for now. Doing the restack in `unmanaged_handle_map()` seems more straightforward and is enough to solve the original issue :shrug:

src/xwayland-unmanaged.c
jlindgren90: Addressed in v2 by doing the restack in `unmanaged_handle_map()`.

src/xwayland.c
```diff
@@ -386,6 +386,13 @@ set_activated(struct view *view, bool activated)
 	wlr_xwayland_surface_activate(surface, activated);
 	if (activated) {
 		wlr_xwayland_surface_restack(surface, NULL, XCB_STACK_MODE_ABOVE);
+		/* Restack unmanaged surfaces on top */
+		struct xwayland_unmanaged *u;
+		struct wl_list *list = &view->server->unmanaged_surfaces;
+		wl_list_for_each (u, list, link) {
+			wlr_xwayland_surface_restack(u->xwayland_surface, NULL,
+						     XCB_STACK_MODE_ABOVE);
```
Consolatis, line 394: I think we need to check for `u->node != NULL` here.

Other than that: LGTM

Consolatis:
I currently can't really test this but the logic seems correct and I don't see it breaking anything.

Thanks once again :)

src/xwayland.c
jlindgren90: Good call

PR #518, opened 2022-09-01
cursor: Fix out-of-surface movement for unmanaged surfaces

jlindgren90:
Fixes an issue when dragging an unmanaged XWayland surface (e.g. `kdiff3` toolbar) -- if the mouse moves outside the toolbar, the toolbar gets "stuck" and no longer follows the mouse.

The handling for unmanaged surfaces is pretty similar to what we do for layer-shell surfaces.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -197,6 +197,42 @@ input_inhibit_blocks_surface(struct seat *seat, struct wl_resource *resource)
 }
 
 static void
+process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
+{
+	struct view *view = server->seat.pressed.view;
+	struct wlr_scene_node *node = server->seat.pressed.node;
+	struct wlr_surface *surface = server->seat.pressed.surface;
+	assert(surface);
+	int lx, ly;
+
+	if (view) {
+		lx = view->x;
+		ly = view->y;
+	} else if (node && wlr_surface_is_layer_surface(surface)) {
+		wlr_scene_node_coords(node, &lx, &ly);
+#if HAVE_XWAYLAND
+	} else if (node && node->parent == server->unmanaged_tree) {
+		wlr_scene_node_coords(node, &lx, &ly);
+#endif
+	} else {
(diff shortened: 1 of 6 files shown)


src/cursor.c
```diff
@@ -296,6 +296,10 @@ process_cursor_motion(struct server *server, uint32_t time)
 			ly = view->y;
 		} else if (wlr_surface_is_layer_surface(server->seat.pressed.surface)) {
 			wlr_scene_node_coords(server->seat.pressed.node, &lx, &ly);
+#if HAVE_XWAYLAND
+		} else if (server->seat.pressed.is_unmanaged) {
+			wlr_scene_node_coords(server->seat.pressed.node, &lx, &ly);
+#endif
```
Consolatis, line 302: I think this could be replaced by `if(node->parent == server->unmanaged_tree)` so we don't need the `is_managed` bool in `seat->pressed`.

Consolatis:
Regarding the other change with using `memset()` to clear out the struct:

I am not sure about the change, it simplifies things a bit but also makes it harder to grep / search for changes of those variables. If we want to keep using it I'd prefer some small function wrapper `lab_reset_pressed_button()` (or a better name) instead so its a bit more self-explaining when reading the code.

Changed the PR into draft for now, feel free to switch it back when ready to merge.

jlindgren90:
Updated with suggestions from @Consolatis. I also split out `process_cursor_motion_out_of_surface()`, mainly to avoid confusion between different `node`/`surface`/`view` variables.

Consolatis:
> I also split out `process_cursor_motion_out_of_surface()`

Nice, that looks much better. I'll give the PR a test run but code LGTM so far.

Consolatis:
Tested with and without xwayland enabled (`meson configure -Dxwayland=disabled build_master`) and everything seems to work fine:
- out-of-surface for usual views
- out-of-surface for layershell clients
- out-of-surface for unmanaged x11 clients

Nice work! Merging.

PR #521, opened 2022-09-02
desktop: Cycle first to topmost view if not already focused

jlindgren90:
The topmost view may not always be the focused view; for example,
when Audacious's main window is focused but the floating Search
Tool window remains on top of it.  In that case the floating window
(not the main window) should be the first view selected in the
window switcher.

Fixes the issue mentioned in this comment:
https://github.com/labwc/labwc/pull/504#issuecomment-1228903654

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -188,11 +188,38 @@ struct view *
 desktop_cycle_view(struct server *server, struct view *start_view,
 		enum lab_cycle_dir dir)
 {
-	struct view *view = start_view ? start_view : first_view(server);
-	if (!view) {
-		return NULL;
+	/*
+	 * Views are listed in stacking order, topmost first.  Usually
+	 * the topmost view is already focused, so we pre-select the
+	 * view second from the top:
+	 *
+	 *   View #1 (on top, currently focused)
+	 *   View #2 (pre-selected)
+	 *   View #3
+	 *   ...
+	 *
+	 * This assumption doesn't always hold with XWayland views,
+	 * where a main application window may be focused but an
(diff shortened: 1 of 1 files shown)


Consolatis:
Will have to think a bit about this, will reply later.

Consolatis:
This made my brain hurt.

> when Audacious's main window is focused but the floating Search Tool window remains on top of it. 

Hm.. right. XWayland surfaces are actually organized in a `wlr_scene tree` which may contain multiple surfaces. That tree in turn is a child of `view->scene_tree`. When we raise a view we raise `view->scene_tree` and thus all connected surfaces. Additionally we use `move_xwayland_sub_views_to_front()` to also raise connected views (and their related surfaces).

Could reproduce the issue with the about dialog of VLC.
When the VLC main window is focused `A-Tab` will first switch to the already focused VLC which is obviously wrong and this PR fixes this while not breaking anything wayland native because there it shouldn't be possible to even have an unfocused view on top of a focused one (with default mousebindings at least).

Nice work!

Only suggestions would be to add something like this to the comment:
> This is especially important for xwayland views as they may consist of multiple views which may be on top of the focused one. Example: VLC "About" dialog, clicking on the main VLC window, "A-Tab"


Also: sorry for the delay, was away from my main system for a few days.

jlindgren90:
Thanks for the review! I expanded the comment with a couple of examples; hopefully it's clearer now.

Consolatis:
> hopefully it's clearer now.

Very much so, nice comment.
Thanks so much for working on all these details, it really makes all the difference on how a compositor feels like for a user.

Merging

Issue #522, opened 2022-09-02
Mouse focus issue (?) after new window is mapped under cursor


jlindgren90:
Steps to reproduce:
1. Start Audacious in Qt 5 mode and display the Search Tool panel (docked inside the main window)
2. Drag the Search Tool panel header away from the main window to undock the panel (so that it becomes a floating window)
3. **Without moving the mouse** press and hold the left mouse button again
4. Attempt to move the Search Tool panel by moving the mouse around

Expected results:
The Search Tool panel moves with the mouse

Actual results:
The Search Tool panel does not move with the mouse

If the mouse is moved even a single pixel between steps 2 and 3, everything works as expected.  It's my guess that the Search Tool panel loses "mouse focus" (this is a thing in `wlroots` right?) due to being unmapped and remapped at the end of step 2.  Then it's necessary to move the mouse a smidge in order to give the Search Tool panel "mouse focus" again.

The same issue doesn't occur under Openbox.

edit: Numbering is hard.

jlindgren90:
This is with `QT_QPA_PLATFORM=xcb` (so using XWayland).

Consolatis:
> Tool panel loses "mouse focus" (this is a thing in wlroots right?)

Yes, it is. wlroots will send application a mouse enter event when newly entering a surface, same for leaving a surface.
Applications may use these events for example to set the cursor icon. See `cursor_rebase()` in `src/cursor.c`.

We are updating mouse focus in `process_cursor_motion)` so I think your theory is solid (unmapped, no mouse movement, click).

It might be enough to add a `cursor_rebase()` to the mouse press handler to fix this, AFAIR cursor_rebase checks if its the same surface and if yes simply returns.

PR #523, opened 2022-09-03
cursor: Update focus when unmanaged surfaces are mapped/unmapped

jlindgren90:
Update cursor focus when unmanaged XWayland surfaces are mapped/unmapped, unless an out-of-surface drag is in progress.

Fixes: #522

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -522,14 +522,6 @@ struct view *desktop_node_and_view_at(struct server *server, double lx,
 struct view *desktop_view_at_cursor(struct server *server);
 
 /**
- * cursor_rebase - set cursor icon for and send motion-event to surface below it
- * @seat - current seat
- * @time_msec - time now
- * @force - disable check for skipping already focused surface
- */
-void cursor_rebase(struct seat *seat, uint32_t time_msec, bool force);
-
-/**
  * cursor_set - set cursor icon
  * @seat - current seat
  * @cursor_name - name of cursor, for example "left_ptr" or "grab"
@@ -537,7 +529,7 @@ void cursor_rebase(struct seat *seat, uint32_t time_msec, bool force);
 void cursor_set(struct seat *seat, const char *cursor_name);
 
 /**
- * cursor_update_focus - update cursor focus
+ * cursor_update_focus - update cursor focus, may update the cursor icon
  * @server - server
  * Use it to force an update of the cursor icon and to send an enter event
  * to the surface below the cursor.
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -24,21 +24,15 @@ is_surface(enum ssd_part_type view_area)
 		;
 }
 
-void
-cursor_rebase(struct seat *seat, uint32_t time_msec, bool force)
+/*
+ * cursor_rebase() for internal use: reuses node, surface, sx and sy
+ * For a public variant use cursor_update_focus()
+ */
+static void
+cursor_rebase(struct seat *seat, struct wlr_scene_node *node,
+		struct wlr_surface *surface, double sx, double sy, uint32_t time_msec,
+		bool force)
 {
-	double sx, sy;
-	struct wlr_scene_node *node;
-	enum ssd_part_type view_area = LAB_SSD_NONE;
-	struct wlr_surface *surface = NULL;
-
-	desktop_node_and_view_at(seat->server, seat->cursor->x,
-		seat->cursor->y, &node, &sx, &sy, &view_area);
-	if (is_surface(view_area)) {
-		surface = lab_wlr_surface_from_node(node);
-	}
-
-	ssd_update_button_hover(node, &seat->server->ssd_hover_state);
 	if (seat->pressed.surface && surface != seat->pressed.surface) {
 		/* Don't leave surface when a button was pressed over another surface */
 		return;
@@ -373,10 +367,26 @@ msec(const struct timespec *t)
 void
 cursor_update_focus(struct server *server)
 {
+	double sx, sy;
+	struct wlr_scene_node *node = NULL;
+	struct wlr_surface *surface = NULL;
+	enum ssd_part_type view_area = LAB_SSD_NONE;
+
 	struct timespec now;
 	clock_gettime(CLOCK_MONOTONIC, &now);
+
+	struct seat *seat = &server->seat;
+	desktop_node_and_view_at(seat->server, seat->cursor->x,
+		seat->cursor->y, &node, &sx, &sy, &view_area);
+
+	if (is_surface(view_area)) {
+		surface = lab_wlr_surface_from_node(node);
+	}
+
+	ssd_update_button_hover(node, &seat->server->ssd_hover_state);
+
 	/* Focus surface under cursor if it isn't already focused */
-	cursor_rebase(&server->seat, msec(&now), false);
+	cursor_rebase(seat, node, surface, sx, sy, msec(&now), false);
 }
 
 void
@@ -701,7 +711,7 @@ cursor_button(struct wl_listener *listener, void *data)
 	wlr_idle_notify_activity(seat->wlr_idle, seat->seat);
 
 	double sx, sy;
-	struct wlr_scene_node *node;
+	struct wlr_scene_node *node = NULL;
 	enum ssd_part_type view_area = LAB_SSD_NONE;
 	uint32_t resize_edges = 0;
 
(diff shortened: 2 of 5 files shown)


Consolatis:
Is there a benefit of doing it on map/unmap instead of cursor button press?
If done on button press it would also cover other surfaces.

jlindgren90:
Waiting until the button press would probably solve #522, but doesn't seem as correct to me.  My thought is that, at any given point in time, the surface visually under the mouse cursor should have the mouse focus.  Doing the rebase on map/unmap makes sure of that.

The rebase on map/unmap probably does need to be done more generally, not just for unmanaged surfaces.

jlindgren90:
I am not really sure of a good general solution here.  It seems like, ultimately, any scene-graph change that affects window/surface arrangement needs to trigger `cursor_update_focus()`.  Which makes me wonder if the "cursor focus" concept is something that should be managed by `wlr_scene` itself.

Consolatis:
I would argue that there is only a finite amount actions that depend on the cursor focus:
- button press/release
- motion
- scrolling

On the other side we have actually removed the `cursor_rebase()` for scrolling here 97846ae45b06600ed5bf4fb54daada583fc48d59 and replaced it by calling it on `desktop_move_to_front()`.

So we have two variants:
- based on surface changes
- based on input

For surface changes we need to have the combination of
- `desktop_move_to_front()`
- map and unmap handler for all surfaces (xwayland, xwayland-unmanaged, xdg, layershell, xdg popups, layershell popups)
- handlers for self-moving surfaces (xwayland + unmanaged) / potentially `view_moved()`
- all tiling / snap to edge / unmaximize cases
- potentially more

For input changes we have
- cursor motion
- cursor button press/release
- axis events (wheel / touchpad scrolling)
- possibly gesture handlers like pinch-to-zoom

> any scene-graph change that affects window/surface arrangement needs to trigger `cursor_update_focus()`

Yes, + potentially a few more.

I think for maintainability reasons going the input based route is more sensible.

jlindgren90:
That's a very good summary 👍 I'm not personally attached to one approach vs. the other.

Consolatis:
@johanmalm thoughts? I would be fine merging this (code LGTM by the way) in preparation for also adding a `cursor_rebase()` / `cursor_update_focus()` to the other map / unmap and axis / button press/release handlers .

johanmalm:
Can’t test, but commit looks good. Happy to merge.

Consolatis, your analysis seems right to me. cursor_rebase() and associated functions are hard to get right! Definitely welcome simplification/overhaul. We just have to step through it carefully. 
Sorry - haven’t got anything more intelligent to say about it :smile:

Consolatis:
The additional test in `cursor_rebase()` breaks the `Raise` action (when, as usual, bound to `Frame` context on `Press` event) which calls `cursor_update_focus()`:
```
00:01:11.316 [types/wlr_compositor.c:712] New wlr_surface 0xffff94d13480 (res 0xffff0fc29c80)
00:01:11.317 [types/wlr_compositor.c:712] New wlr_surface 0xffff94d13080 (res 0xffff0fc29b00)
00:01:11.317 [types/xdg_shell/wlr_xdg_surface.c:391] new xdg_surface 0xffff0da52e40 (res 0xffff0fc29980)
00:01:11.317 [types/wlr_xdg_decoration_v1.c:206] new xdg_toplevel_decoration 0xfffecd424440 (res 0xffff0fc29800)
00:01:11.318 [types/wlr_compositor.c:712] New wlr_surface 0xffff94d12880 (res 0xffff0fc29500)
00:01:11.422 [../src/cursor.c:38] Rebasing cursor:     pressed not set, drag_icon not set
00:01:27.874 [../src/action.c:187] Handling action Focus (15)
00:01:27.874 [../src/action.c:187] Handling action Raise (18)
00:01:27.874 [../src/cursor.c:31] Not rebasing cursor: pressed     set, drag_icon not set
```

It think it should be fine to move the check down a bit (right before `if (surface) {`) and compare it to the newly detected surface (e.g. something like `if (seat->pressed.surface && seat->pressed.surface != surface) { return: }`.

The issue can be tested the same as for #522 just with an xdg / xwayland window. e.g. something like
- open some application that will allow to scroll something and place it somewhere in the middle of the screen
- open another terminal somewhere else
- `sleep 5; foot`
- position the cursor in the middle of the screen
- once the terminal opens it should have keyboard focus so execute something to have a lot of text (/me used `man man`)
- try to scroll with the mouse wheel without moving the mouse at all
- observe the scrolling going to the first window that is behind the new one
- try to click without moving the mouse at all (slightly tricky)
- observe it calls `Raise` (due to the mouse `Frame` binding for it) and fails because the button is pressed

This wouldn't be an issue when we'd have an additional cursor_rebase on click and axis but we don't have that for now.
So to get this forward we have two options:

- waiting for the other parts of the (hopefully final) solution to be implemented
  (I could push those here if nobody else wants to work on it) and replace the `desktop_move_to_front()` call with cursor button handler + map/unmap handlers
- fixing this PR commit so it doesn't break `Raise` for now to fix #522.

I am fine with both variants.

jlindgren90:
> * open some application that will allow to scroll something and place it somewhere in the middle of the screen
> * open another terminal somewhere else
> * `sleep 5; foot`
> * position the cursor in the middle of the screen
> * once the terminal opens it should have keyboard focus so execute something to have a lot of text (/me used `man man`)
> * try to scroll with the mouse wheel without moving the mouse at all
> * observe the scrolling going to the first window that is behind the new one

This particular issue already exists in `master` today.

The problem seems to be that `desktop_node_and_view_at()` does not "see" XDG views until `handle_commit()`, which is called **after** `desktop_move_to_front()`.  Therefore the `cursor_update_position()` in `desktop_move_to_front()` doesn't work for newly-mapped XDG views.

Adding a `cursor_update_position()` to the end of `view_moved()` would fix the issue, since `view_moved()` is already called from `handle_commit()`.

jlindgren90:
@Consolatis I added the `cursor_update_focus()` call to `view_moved()` as well as a couple more in `xwayland-unmanaged.c` for completeness. I believe this fixes the `foot` scrolling issue you described.

I don't think I understand the issue with `Raise`.  In my attempt to repeat your steps, the new `foot` window is already "raised" (on top) at this point:

> * try to click without moving the mouse at all (slightly tricky)
> * observe it calls Raise (due to the mouse Frame binding for it) and fails because the button is pressed

Consolatis:
> the new foot window is already "raised" (on top) at this point:

Right, but the cursor focus is still somewhere else (due to missing mouse movement and missing `cursor_rebase` calls in the `xdg.c` map handler).

The raising-to-top feature when clicking keeps working fine but the internally called `cursor_update_focus()` fails because that mousebinding is called on button `Press` and as such the new check in `cursor_rebase()` triggers and returns early because it assumes this is an out-of-surface movement (which it is not).

jlindgren90:
> Right, but the cursor focus is still somewhere else (due to missing mouse movement and missing cursor_rebase calls in the xdg.c map handler).

Please try with the latest version; this should be fixed since the cursor focus is updated in `view_moved()` (which is also called indirectly after map by `handle_commit()`).

Consolatis:
> This particular issue already exists in `master` today.
> 
> The problem seems to be that `desktop_node_and_view_at()` does not "see" XDG views until `handle_commit()`, which is called **after** `desktop_move_to_front()`. Therefore the `cursor_update_position()` in `desktop_move_to_front()` doesn't work for newly-mapped XDG views.
> 
> Adding a `cursor_update_position()` to the end of `view_moved()` would fix the issue, since `view_moved()` is already called from `handle_commit()`.

Somehow I completely missed this message.
That explains why the first `cursor_rebase()` (log timestamp `00:01:11.422`) didn't do anything.

Will test your latest version in a few hours, need to get some rest first.

jlindgren90:
I added one more loosely-related commit, which fixes a trivial issue, where the "maximize" button would still appear "hovered" after maximizing the window.

Consolatis:
>  the "maximize" button would still appear "hovered" after maximizing the window.

Just to clarify, this is when not moving the mouse, right?

Consolatis:
Pushed some additional commits:
- src/xdg.c: Only call view_moved() when required
- cursor: Allow cursor button events to trigger cursor_rebase()
- cursor: Make cursor_rebase() private / split into public (`cursor_update_focus()`) and private (`cursor_rebase()`) parts
- cursor: Update cursor focus on scroll events

jlindgren90:
I don't see any regressions with the latest commits, +1 from me

johanmalm:
Thank you both :smile:

PR #525, opened 2022-09-03
xwayland: Factor out focus_next_surface() from unmanaged_handle_unmap()

### LINKED ISSUE #520: Unmanaged X11 windows may give focus to a wrong unmanaged window on unmap
I think this should be a `wl_list_for_each_reverse()` instead, didn't test though.
We are adding to the list using `.prev` so the list is ordered correctly.
Thus `last created unmanaged xwayland surface` should be at the end of the list.

https://github.com/labwc/labwc/blob/d405a2b3c61fd5e82fbd4b719b86215d6b25469d/src/xwayland-unmanaged.c#L85-L95

jlindgren90:
- Eliminate multiple `return` paths in `unmanaged_handle_unmap()`, which were a bug waiting to happen.
- Use `wl_list_for_each_reverse()` rather than `wl_list_for_each()` to find the topmost (most-recently-created) unmanaged surface.
- Only call `desktop_focus_topmost_mapped_view()` if the unmapped surface was actually focused.

Fixes #520.

### DIFF
--- a/src/xwayland-unmanaged.c
+++ b/src/xwayland-unmanaged.c
@@ -52,6 +52,38 @@ unmanaged_handle_map(struct wl_listener *listener, void *data)
 }
 
 static void
+focus_next_surface(struct server *server, struct wlr_xwayland_surface *xsurface)
+{
+	/*
+	 * Try to focus on parent surface
+	 * This seems to fix JetBrains/Intellij window focus issues
+	 */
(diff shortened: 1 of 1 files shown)


Consolatis:
You really are on a mission here :)

> Only call desktop_focus_topmost_mapped_view() if the unmapped surface was actually focused.

Nice catch.

Merging, thanks

Issue #317, opened 2022-05-07
GIMP popups are invisible sometimes


acoolstraw:
GIMP running under xwayland sometimes has the popup windows invisible (for example color change). At first, when you do something that creates a popup like changing text color from the overlay or scaling, the window contents are visible. But sometimes, when you do that thing, it just doesn't render the window contents. I'm not sure why, or how that happens.

labwc version: `labwc 0.5.0-129-g9f181de` (compiled from master, did happen on 0.5.0 too)
wlroots version: `0.15.1-3`

Logs of running `labwc -d 2>log.txt` and reproducing it after a lot of attempts: https://i.troplo.com/i/58708fa34170.txt

![image](https://i.troplo.com/i/84e5632b9973.png )

acoolstraw:
[Clip of it happening, starts at around 0:50](https://i.troplo.com/i/ddcb151dfd62.mp4)

Consolatis:
> [01:14:15] <acoolstraw> this is weird
> [01:14:20] <acoolstraw> bemenu just doesn't appear
> [01:14:33] <acoolstraw> it appears to work - if I run bemenu-run and run a command it does it
> [01:14:39] <acoolstraw> I can launch foot and other apps
> [01:14:47] <acoolstraw> but the menu itself just doesn't render
> [01:15:58] <acoolstraw> it started after I ran Veloren (a Vulkan application, maybe that has something to do with it), and I did something that made the SSD disappear except for the window buttons which stayed there but they were to the left of the dot button (???)
> [01:23:13] <acoolstraw> has anyone had a similar issue with window decorations?

SSD disappearing while still showing the window buttons seems very strange to me.
When toggling decorations it will set the whole SSD tree to disabled and thus the children should *never* be visible because one of their parents got disabled. This almost sounds like some buffers wouldn't get allocated properly which could also explain why an external application like bemenu would suddenly stop rendering as well.

While thinking about that (and also the other issues you experienced).. (x)wayland uses FDs for buffers and the more of them there are it might be possible that the FD limit for labwc and/or xwayland has been reached and thus new buffers can't be allocated properly anymore. #317 (gimp windows) might actually be related to that as well. Setting higher FD limits via ulimit before starting labwc could eventually solve it. #249 might help to solve this automatically if it is indeed the root cause.

[...]

PR #526, opened 2022-09-03
xwayland: Keep view->xwayland_surface and view->surface in sync

Consolatis:
This will hopefully fix
- #317

@acoolstraw @BitterlyIronic testing appreciated.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -349,6 +349,7 @@ struct view {
 	struct wl_listener map;
 	struct wl_listener unmap;
 	struct wl_listener destroy;
+	struct wl_listener surface_destroy;
 	struct wl_listener commit;
 	struct wl_listener request_move;
 	struct wl_listener request_resize;
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -96,12 +96,33 @@ handle_unmap(struct wl_listener *listener, void *data)
 }
 
 static void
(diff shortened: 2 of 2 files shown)


BitterlyIronic:
Been running this for the past day and I haven't seen the empty window issue crop up in that time.

johanmalm:
> Been running this for the past day and I haven't seen the empty window issue crop up in that time.

Thanks for feedback. I vote for merge if @Consolatis is content

Consolatis:
Simplified the code a bit:
https://github.com/labwc/labwc/blob/a0afd443f59b6b9cab1563d3b963e9ecb7af98e4/src/xwayland.c#L99-L108

@johanmalm would be great if you could look over it again, there shouldn't actually be any race conditions left.

Thus the new `assert()` should never trigger:
- on map
  - when `view->surface` changed
    - when `view->surface` is set
      - remove the destroy handler
    - assign new `view->surface`
    - add the destroy handler for the new surface
- on `view->xwayland_surface` destroy
  - when `view->surface` set
    - remove the destroy handler
    - set `view->surface` to `NULL`
- on `view->surface` destroy
  - assert `view->surface` == surface being destroyed
  - remove the destroy handler
  - set `view->surface` to `NULL`

johanmalm:
Thanks!

PR #527, opened 2022-09-05
Honor XWayland client window positions

jlindgren90:
Currently all XWayland views are centered, regardless of whether the application requested a specific window position. This breaks applications that save and restore their own window positions, and is particularly bad for usability when the application displays multiple windows (e.g. Audacious in Winamp mode), since they all get stacked on top of each other.

Two changes here:
- Honor the ICCCM window position hints
- Fix a bug where `wlr_scene_node_set_position()` might never be called

v2: Use `view_adjust_for_layout_change()` to make sure the window is still on-screen.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -307,7 +307,7 @@ xdg_toplevel_view_map(struct view *view)
 			view_maximize(view, true);
 		}
 
-		view_discover_output(view);
+		view_moved(view);
 		view->been_mapped = true;
 	}
 
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -277,6 +277,27 @@ handle_override_redirect(struct wl_listener *listener, void *data)
 }
 
 static void
+set_initial_position(struct view *view)
+{
+	/* Don't center views with position explicitly specified */
+	bool has_position = view->xwayland_surface->size_hints &&
+		(view->xwayland_surface->size_hints->flags &
+			(XCB_ICCCM_SIZE_HINT_US_POSITION |
+			 XCB_ICCCM_SIZE_HINT_P_POSITION));
+
+	if (has_position) {
+		/* Just make sure the view is on-screen */
+		view_adjust_for_layout_change(view);
+	} else {
+		struct wlr_box box =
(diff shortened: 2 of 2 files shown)


johanmalm:
Do we need to sanity check the position to ensure the view is visible on an output?
I can’t test atm, but what happens if the output-layout has changed since the client was last run?

jlindgren90:
> Do we need to sanity check the position to ensure the view is visible on an output?

It would be a good idea, yes. Do we have anything like a `view_ensure_visible()` yet?

johanmalm:
> It would be a good idea, yes. Do we have anything like a `view_ensure_visible()` yet?

No. The use of view_centre() has avoided a need for it. 😄 
The only “adjustment” we make is to top/left align if view-size > usable_area. See link below:
https://github.com/labwc/labwc/blob/05b7afd05a32abc6c167dae3d8634853bec78799/src/view.c#L276

jlindgren90:
I found `view_adjust_for_layout_change()` which we are using already when an output is disconnected -- it seems to do the trick here as well.

johanmalm:
> I found `view_adjust_for_layout_change()` which we are using already when an output is disconnected -- it seems to do the trick here as well.

Can’t test at the moment. Do we get sensible behaviour if taking the following steps?

1. Launch client on large screen
2. Position client window bottom/right
3. Shut down client + compositor
4. Re-start compositor with smaller output (e.g. laptop) - where output extents does not cover the clients previous position.
5. Launch client

jlindgren90:
> Can’t test at the moment. Do we get sensible behaviour if taking the following steps?
> 
> 1. Launch client on large screen
> 2. Position client window bottom/right
> 3. Shut down client + compositor
> 4. Re-start compositor with smaller output (e.g. laptop) - where output extents does not cover the clients previous position.
> 5. Launch client

Yes, in that case the client windows end up centered on the smaller output.  (In the case of a client with multiple windows, they get stacked on top of each other again, but I think that's to be expected in the fallback case.)

johanmalm:
Thanks!

PR #529, opened 2022-09-06
action: Allow explicit None action without warning

### LINKED ISSUE #524: add the "None" action, that unsets default behavior
This is the corollary to #416 

In openBox, I had this config XML so left click would not trigger the menu. (this makes for more consistent behavior IMO): 

    <context name="Desktop">                                                            
        <action name="None"/>                                                           
      </mousebind>                                                                      
    </context> 
 
thanks for awesome project

jlindgren90:
Silence the "Invalid action" warning when encountering a config section such as this:
```
  <mouse>
    <context name="Desktop">
        <action name="None"/>
      </mousebind>
    </context>
  </mouse>
```

Besides silencing the warning, there is no behavioral difference, since unrecognized actions are already handled as the no-op `ACTION_TYPE_NONE`.

Fixes #524.


johanmalm:
Aahh… I assumed the purpose of “None” in #524 was to clear any existing actions in the same context+etc (i.e. desktop-left-click in example above).

jlindgren90:
> Aahh… I assumed the purpose of “None” in #524 was to clear any existing actions in the same context+etc (i.e. desktop-left-click in example above).

It does that already, but causes a warning to be logged.

jlindgren90:
Hmm, I had an error in my config.  This is not working as intended yet.

jlindgren90:
It IS working correctly; there was a second error in my config. Sorry for the noise.

This is the corrected config section I have tested:
```
  <mouse>
    <default />
    <context name="Desktop">
      <mousebind button="Left" action="Press">
        <action name="None"/>
      </mousebind>
    </context>
  </mouse>
```

jlindgren90:
> I would prefer to additionally have another type like `ACTION_TYPE_INVALID` at index `0` which then acts the same as `ACTION_TYPE_NONE` did before (e.g. replacing the fallback return value from `action_type_from_str()` with the new type, starting from index `1` and adding another log handler for it like the original one for `ACTION_TYPE_NONE`).
>
> We should likely also document this somewhere.

I don't disagree :) Unfortunately I am a little burned out labwc issues at the moment, so please feel free to take this change as a starting point (if it's useful) and improve/rewrite it.

Consolatis:
> Unfortunately I am a little burned out on labwc issues at the moment

That was quite a run though :)

> please feel free to take this change as a starting point

Will do, turning this into a draft PR for now

Consolatis:
Force pushed a new version, rebased on top of the original patch.
Didn't test but should be fine, the docs part is pretty bad though (help appreciated).

We could also enhance this and don't actually create the action in the first place and return `NULL` instead but that would require some more work in `src/config/rcxml.c` to handle a `NULL` result. When we have that done we could add more post-processing steps to
- deduplicate the keybinds like we are already doing for the mousebinds
- remove all key- and mousebinds which have an empty action list

That would just be some optimization though so we have less bindings to loop through.

@johanmalm up to you if we want to merge it now (maybe with better docs though) or wait for a "proper clear" to be implemented as well.

johanmalm:
I think it looks good. Sensible approach. Keeps it simple. I think docs are okay.

Consolatis:
We are missing an actual clear-keybind functionality. Maybe `actions_run()` could have a `bool` return value which allows the key combination to be delivered to the focused application instead? Or do we need a futher `Clear` action instead? Or do we solve this by implementing the suggestion above with deduplication + clearing of the bindings with empty action list?

Currently its impossible to use `A-Left` or `A-Right` for applications while also using `<default />` for keybinds, even when setting the keybind to the new `None` action.

PR #531, opened 2022-09-08
Titlebar fixes

### LINKED ISSUE #530: Windows occasionally end up in weird half-unmaximized state
Occasionally when attempting to unmaximize a window, it ends up in this weird in-between state:
PXL_20220908_153354493

There are a couple of things going on here:
- If the mouse cursor moves ever so slightly while pressed on the "unmaximize" button, `labwc` thinks we are trying to move the window by dragging on the titlebar (i.e. `interactive_begin()`).
- `interactive_begin()` changes `view->x` and `view->y` directly but is missing a call to `view_moved()`.  This leads to the scene-graph being out-of-date and not matching where `labwc` "thinks" the view is displayed.

I'll follow up with a pull request to fix at least the missing `view_moved()` call.

jlindgren90:
Three different fixes here:

1.

    interactive: Add missing view_moved() call
    
    Without this, the scene-graph isn't updated, and the view ends up
    in a weird half-maximized state.

2.

    config: Add TitlebarNoButtons mousebind context
    
    Use this for the Drag (Move) and DoubleClick (Maximize) titlebar
    actions, which are unexpected when the cursor is over one of the
    window buttons.

3.

    ssd: Make ssd_get_part_type() work for corner buttons
    
    Corner buttons (WINDOW_MENU and CLOSE) are one more level down in
    the scene-tree (see add_scene_button_corner() in ssd_part.c).
    
    This fixes a minor issue where, when right-clicking on the CLOSE
    button, the client-menu would be displayed in the wrong location.

Fixes #530.


Consolatis:
> interactive: Add missing view_moved() call

See #530

> config: Add TitlebarNoButtons mousebind context

Seem a pragmatic way to solve this. /me likes.
Would be great of you could also update `docs/rc.xml.all` with the new defaults so that its always in sync with our hardcoded fallbacks. Other than that: Code LGTM, not tested

> ssd: Make ssd_get_part_type() work for corner buttons

Interesting observation. And appropriate fix as well. Code LGTM, not tested

jlindgren90:
I fixed up the `interactive_begin()` logic to write only to `server->grab_box`, not `view->x/y/w/h`.  Also updated `docs/rc.xml.all` per request.

Consolatis:
Had a quick test:

> interactive: Don't write to view->x/y/w/h directly

Seems to work fine

> config: Add TitlebarNoButtons mousebind context

Seems to work fine, including when using the old config

> ssd: Make ssd_get_part_type() work for corner buttons

Seems to work fine

Ready to merge from my point of view.

However, as we are adding a new Mouse context here I won't merge this myself without giving @johanmalm a chance to approve of the new `TitlebarNoButtons` context.

johanmalm:
What’s the case for still needing Titlebar in this patch-set?
If there is no case, could the new TitlebarNons be called Titlebar (i.e. just redefined to not include area containing buttons)

jlindgren90:
My thought was that right-clicking on one of the buttons should still show the client-menu (`Titlebar` action). And left-click should still invoke `Raise` and `Focus`. But I'm not strongly opinionated about it one way or the other.

jlindgren90:
Alternatively we could redefine `Title` to include the blank space and be the equivalent of the new `TitlebarNoButtons`.

Consolatis:
> My thought was that right-clicking on one of the buttons should still show the client-menu (Titlebar action). And left-click should still invoke Raise and Focus.

Good point

> Alternatively we could redefine Title to include the blank space and be the equivalent of the new TitlebarNoButtons.

I like this

jlindgren90:
Updated to use `Title` in place of `TitlebarNoButtons`

Consolatis:
Had a quick test:
When only using the new `Title` bindings everything appears to work fine.
When additionally using the old bindings for `TitleBar` however, things start to break in regards to maximize / unmaximize.

Personally I think this is a nice improvement and we should just tell people to either use the `<default />` for mousebinds or remove their existing `TitleBar` binding for `Drag` and `DoubleClick` from their configs.

Maybe we could add a note to the commit to remove the old `Move` and `Maximize` binding for `TitleBar` (even though I guess nobody usually reads those unless they are developing for labwc).

We should also add a note to the changelog / release notes when releasing `0.6` at some point in the future.

johanmalm:
Great fixes. Thanks.
Consolatis, I've made a note in my NEWS.md about breaking changes.

PR #532, opened 2022-09-09
rcxml.c: enable tap be default on non-touch devices

johanmalm:
Some trackpads do not honour tap if we only enable it on touch-devices.


johanmalm:
I now have two laptops where this applies. My main laptop changed behavior after the last `pacman -Syyu` so I reckon it makes more sense to change to `DEFAULT_DEVICE`.

jlindgren90:
This makes tap-to-click work on my laptop as well :+1:

Issue #350, opened 2022-05-27
Verify drag and drop


Consolatis:
Currently drag and drop on X11 seems broken for both, master and v0.5.
Drag and drop on wayland seems usually working (as long as the application sets a drag_icon) for master and v0.5.

We are using `server->seat.drag_icon` to figure out if we are actually in a drag and drop operation right now but wlroots actually [says](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_data_device.h#L125) that the icon may be `NULL` so its not a reliable indicator for a drag and drop operation.
Additionally, we are using the `destroy` of the icon to finish the d'n'd operation, we should likely switch to `wlr_drop->events.destroy`.

There is also a `wlr_drop->events.motion` event which could potentially decouple the d'n'd from the other cursor motion stuff.

jlindgren90:
For what it's worth, I just now tried dragging-and-dropping some files from one Thunar window to another (both XWayland) and it seemed to work fine.

jlindgren90:
Dragging from Thunar to Thunar (Wayland) also works but the mouse cursor doesn't change.
Dragging from Thunar (Wayland) to Audacious (XWayland) doesn't work, and the mouse cursor over the Thunar window gets stuck as a hand/move icon thereafter.

Consolatis:
Thanks to @Flrian via IRC we got some more information around this area:

Crashing labwc, simple reproducer:
`sleep 20; killall nautilus` and start a drag within nautilus. Once nautilus is killed labwc crashes. Likely the same with other applications.

```
==128267==ERROR: AddressSanitizer: heap-use-after-free on address 0x61a00000a888 at pc 0x55c943acc5e9 bp 0x7ffee27342a0 sp 0x7ffee2734298
READ of size 8 at 0x61a00000a888 thread T0
    #0 0x55c943acc5e8 in drag_set_focus ../subprojects/wlroots/types/data_device/wlr_drag.c:64
    #1 0x55c943acdc84 in drag_handle_pointer_enter ../subprojects/wlroots/types/data_device/wlr_drag.c:166
    #2 0x55c943b0ff02 in wlr_seat_pointer_notify_enter ../subprojects/wlroots/types/seat/wlr_seat_pointer.c:425
    #3 0x55c94397e54e in cursor_update_common ../src/cursor.c:386
    #4 0x55c94397fa68 in cursor_update_focus ../src/cursor.c:482
[...]

PR #533, opened 2022-09-10
cursor: Fix GTK3 menus in native Wayland mode

jlindgren90:
**Problem description**

Drop-down menus in GTK3 (native Wayland) applications are not working correctly:

1. Press and hold left mouse button over e.g. "File" in the menu bar
2. While holding the button, move the mouse down into the menu that appears
3. Let up the mouse button over the menu item you want to activate (e.g. "Open File")

With GTK3 apps in native Wayland mode, none of the menu items are highlighted while moving the mouse over them, and the menu simply closes when releasing the button (no menu item is activated).

I tested various GTK3 apps (Geany, Thunar, xfce4-terminal): all have the same issue.

**Other observations**

- The same issue occurs under both Sway and KWin.
- The issue does not occur under Weston.
- The issue does not occur with `GDK_BACKEND=X11`.
- Qt 5 menus don't suffer from the same issue.

There was a bug report of a similar issue at https://gitlab.gnome.org/GNOME/gtk/-/issues/3662 but it's supposedly fixed. There was apparently a change needed in Mutter as well.

**Analysis and fix/workaround**

Using `WAYLAND_DEBUG=1`, I found that Weston is sending leave/enter events when the mouse pointer enters the menu (XDG popup).  Labwc currently doesn't send any leave/enter events at this point due to our `seat.pressed` logic in `cursor.c`.  We can't just rip out that logic either, without breaking things in other scenarios.

It's tempting to just call this a bug in GTK3, especially given that it occurs under KWin as well.  But I'm not hopeful about our chances trying to get the GTK developers to share that perspective.

A workaround we could use (and what I've implemented in this pull request) is to add a special case (ugh) and send the leave/enter events when the pointer is moved between XDG surfaces owned by the same client.  There's some precedent for this, looking at how `wlroots` handles popup grabs (see `xdg_pointer_grab_enter()` in `wlr_xdg_popup.c`):
```
	if (wl_resource_get_client(surface->resource) == popup_grab->client) {
		wlr_seat_pointer_enter(grab->seat, surface, sx, sy);
	} else {
		wlr_seat_pointer_clear_focus(grab->seat);
	}
```

I'd appreciate some other ideas and lots of testing on this, to make sure it doesn't break other things.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -45,6 +45,32 @@ is_surface(enum ssd_part_type view_area)
 		;
 }
 
+static struct wlr_surface *
+get_toplevel(struct wlr_surface *surface)
+{
+	while (surface && wlr_surface_is_xdg_surface(surface)) {
+		struct wlr_xdg_surface *xdg_surface =
+			wlr_xdg_surface_from_wlr_surface(surface);
+		if (!xdg_surface) {
+			return NULL;
+		}
+
+		switch (xdg_surface->role) {
+		case WLR_XDG_SURFACE_ROLE_NONE:
+			return NULL;
+		case WLR_XDG_SURFACE_ROLE_TOPLEVEL:
+			return surface;
+		case WLR_XDG_SURFACE_ROLE_POPUP:
+			surface = xdg_surface->popup->parent;
+			continue;
+		}
+	}
+	if (surface && wlr_surface_is_layer_surface(surface)) {
+		return surface;
+	}
+	return NULL;
+}
+
 static void
 request_cursor_notify(struct wl_listener *listener, void *data)
 {
@@ -227,6 +253,28 @@ input_inhibit_blocks_surface(struct seat *seat, struct wl_resource *resource)
 		&& inhibiting_client != wl_resource_get_client(resource);
 }
 
+static bool
+update_pressed_surface(struct seat *seat, struct view *view,
+		struct wlr_scene_node *node, struct wlr_surface *surface)
+{
+	/*
+	 * In most cases, we don't want to leave one surface and enter
+	 * another while a button is pressed.  However, GTK/Wayland
+	 * menus (implemented as XDG popups) do need the leave/enter
+	 * events to work properly.  To cover this case, we allow
+	 * leave/enter events between XDG popups and their toplevel.
+	 */
+	if (seat->pressed.surface && surface != seat->pressed.surface) {
+		struct wlr_surface *toplevel = get_toplevel(surface);
+		if (toplevel && toplevel == seat->pressed.toplevel) {
+			seat_set_pressed(seat, view, node,
+				surface, toplevel);
+			return true;
+		}
+	}
+	return false;
+}
+
 static void
 process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
 {
@@ -267,10 +315,10 @@ process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
  * Common logic shared by cursor_update_focus() and process_cursor_motion()
  */
 static void
-cursor_update_common(struct server *server, struct wlr_scene_node *node,
-		struct wlr_surface *surface, double sx, double sy,
-		enum ssd_part_type view_area, uint32_t time_msec,
-		bool cursor_has_moved)
+cursor_update_common(struct server *server, struct view *view,
+		struct wlr_scene_node *node, struct wlr_surface *surface,
+		double sx, double sy, enum ssd_part_type view_area,
+		uint32_t time_msec, bool cursor_has_moved)
 {
 	struct seat *seat = &server->seat;
 	struct wlr_seat *wlr_seat = seat->seat;
@@ -287,6 +335,7 @@ cursor_update_common(struct server *server, struct wlr_scene_node *node,
 
 	/* TODO: verify drag_icon logic */
 	if (seat->pressed.surface && surface != seat->pressed.surface
+			&& !update_pressed_surface(seat, view, node, surface)
 			&& !seat->drag_icon) {
 		if (cursor_has_moved) {
 			/*
@@ -398,8 +447,8 @@ process_cursor_motion(struct server *server, uint32_t time)
 		}
 	}
 
-	cursor_update_common(server, node, surface, sx, sy, view_area, time,
-		/*cursor_has_moved*/ true);
+	cursor_update_common(server, view, node, surface, sx, sy, view_area,
+		time, /*cursor_has_moved*/ true);
 }
 
 static uint32_t
@@ -420,15 +469,15 @@ cursor_update_focus(struct server *server)
 	clock_gettime(CLOCK_MONOTONIC, &now);
 
 	struct seat *seat = &server->seat;
-	desktop_node_and_view_at(seat->server, seat->cursor->x,
-		seat->cursor->y, &node, &sx, &sy, &view_area);
+	struct view *view = desktop_node_and_view_at(seat->server,
+		seat->cursor->x, seat->cursor->y, &node, &sx, &sy, &view_area);
 
 	if (is_surface(view_area)) {
 		surface = lab_wlr_surface_from_node(node);
 	}
 
 	/* Focus surface under cursor if it isn't already focused */
-	cursor_update_common(server, node, surface, sx, sy, view_area,
+	cursor_update_common(server, view, node, surface, sx, sy, view_area,
 		msec(&now), /*cursor_has_moved*/ false);
 }
 
@@ -798,8 +847,9 @@ cursor_button(struct wl_listener *listener, void *data)
 		if (server->input_mode == LAB_INPUT_STATE_MENU) {
 			if (close_menu) {
 				menu_close_root(server);
-				cursor_update_common(server, node, surface, sx, sy,
-					view_area, event->time_msec, false);
+				cursor_update_common(server, view, node,
+					surface, sx, sy, view_area,
+					event->time_msec, false);
 				close_menu = false;
 			}
 			return;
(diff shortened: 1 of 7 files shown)


Consolatis:
Interesting, thanks for sharing your research and this potential solution. I think it is a solid idea.
We could potentially expand this to layer surfaces (e.g. GTK based panels like sfwbar) as well.

Not sure about interactions with drag-and-drop with this approach but that [really needs work](https://github.com/labwc/labwc/issues/350) anyway because its currently half broken.

johanmalm:
Wow, I had never noticed this behaviour - probably because I don't keep pressing the mouse-button when moving around menus - but I can re-produce it and yeah, that's pretty bad :smile:

Thanks for the analysis.

My initial thoughts are that:

- Clients (including GTK) are quite within their rights to do this and compositors ought to deal with it.
- The edge-case is really out-of-surface scrolling / selection, i.e. the stuff we do in process_cursor_motion_out_of_surface()... and it's amazing how much trouble that creates.

In terms of readability, I suggest (and it's only a suggestions) splitting okay_to_focus_surface() into

is_pressed_surface(struct seat *seat, struct wlr_surface *surface)
{
	assert(seat);
	return seat->pressed_surface && surface == seat->pressed.surface;
}

has_same_client_as_pressed_surface(struct seat *seat, struct wlr_surface *surface)
{
	assert(seat);
	if (!surface) {
		return false;
	}
	// Should we really check that it's an xdg-surface??
	// Presumably any client is entitled to create new (sub-) surfaces if they feel
	// like it.
	return surface->resource->client == pressed->resource->client);
}

(not tested by the way - just thinking aloud).

johanmalm:
There is a problem using `resource->client` for the comparison.

**Problem description**

1. Open `geany` (run as xdg-shell)
2.  Open `about` dialog (help --> about). The `about` window should be a a lot smaller than geany - and positioned within the max extents of the geany-toplevel-window.
3. Select the `licence` tab
4. Scroll with the scroll bar or select text and (with the mouse-button still pressed) move outside the window.
5. Observe that scroll / selection is lost because the geany-toplevel-window is the same client.

Maybe we could compare toplevels instead? E.g. using xdg_surface_from_wlr_surface() and comparing the toplevel members.

Consolatis:
> Maybe we could compare toplevels instead? E.g. using xdg_surface_from_wlr_surface() and comparing the toplevel members.

That sounds like it could work. I think it might also catch subsurfaces, not sure though.
Basically we only want this to happen on xdg / layershell popups, that might be another way to check for it, something like is_popup_of(new_surface, pressed_surface) which checks for `resource->client` && type `{xdg/layershell}_popup`?

johanmalm:
> ...something like `is_popup_of(new_surface, pressed_surface)`

What if a submenu has been opened using a keybind, when then press on that submenu and follow it to its parent? That's why I thought we might need to follow it up to a common toplevel surface... but I'm not sure.

jlindgren90:
> What if a submenu has been opened using a keybind, when then press on that submenu and follow it to its parent?

Do the `is_popup_of` check in both directions?

`is_popup_of(new_surface, pressed_surface) || is_popup_of(pressed_surface, new_surface)`

Consolatis:
> What if a submenu has been opened using a keybind, when then press on that submenu and follow it to its parent? That's why I thought we might need to follow it up to a common toplevel surface... but I'm not sure.

Good point. Didn't think about moving from the popup to the parent.
With that in mind, not restricting this to popups and instead only checking for the toplevel indeed sounds like a good plan.

jlindgren90:
I'd like to get #540 merged first. This will need rebasing to fix a merge conflict.

jlindgren90:
Ready for review, there will still be a merge conflict with #540 though.

Consolatis:
With the following this also works for layershell surfaces (turns out layershell popups are in fact just xdg_popups):

```diff
diff --git a/src/cursor.c b/src/cursor.c
index 29a9e4a..31c2825 100644
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -50,7 +50,7 @@ is_surface(enum ssd_part_type view_area)
 }

 static struct wlr_surface *
-get_xdg_toplevel(struct wlr_surface *surface) {
+get_toplevel(struct wlr_surface *surface) {
        while (surface && wlr_surface_is_xdg_surface(surface)) {
                struct wlr_xdg_surface *xdg_surface =
                        wlr_xdg_surface_from_wlr_surface(surface);
@@ -64,11 +64,14 @@ get_xdg_toplevel(struct wlr_surface *surface) {
                        continue;
                }
        }
+       if (surface && wlr_surface_is_layer_surface(surface)) {
+               return surface;
+       }
        return NULL;
 }

 static bool
-is_same_xdg_toplevel(struct wlr_surface *a, struct wlr_surface *b) {
+is_same_toplevel(struct wlr_surface *a, struct wlr_surface *b) {
        /*
         * In most cases, we don't want to leave one surface and enter
         * another while a button is pressed.  However, GTK/Wayland
@@ -76,8 +79,8 @@ is_same_xdg_toplevel(struct wlr_surface *a, struct wlr_surface *b) {
         * events to work properly.  To cover this case, we allow
         * leave/enter events between an XDG toplevel and its popups.
         */
-       struct wlr_surface *toplevel_a = get_xdg_toplevel(a);
-       return toplevel_a && toplevel_a == get_xdg_toplevel(b);
+       struct wlr_surface *toplevel_a = get_toplevel(a);
+       return toplevel_a && toplevel_a == get_toplevel(b);
 }

 /*
@@ -90,7 +93,7 @@ cursor_rebase(struct seat *seat, struct wlr_scene_node *node,
                bool force)
 {
        if (seat->pressed.surface && surface != seat->pressed.surface
-                       && !is_same_xdg_toplevel(surface, seat->pressed.surface)) {
+                       && !is_same_toplevel(surface, seat->pressed.surface)) {
                /* Don't leave surface when a button was pressed over another surface */
                return;
        }
@@ -424,7 +427,7 @@ process_cursor_motion(struct server *server, uint32_t time)

        struct seat *seat = &server->seat;
        if (seat->pressed.surface && surface != seat->pressed.surface
-                       && !is_same_xdg_toplevel(surface, seat->pressed.surface)
+                       && !is_same_toplevel(surface, seat->pressed.surface)
                        && !server->seat.drag_icon) {
                /*
                 * Button has been pressed while over another surface
```

With that: LGTM

src/cursor.c
```diff
@@ -49,6 +49,37 @@ is_surface(enum ssd_part_type view_area)
...
+}
+
+static bool
+is_same_xdg_toplevel(struct wlr_surface *a, struct wlr_surface *b) {
+	/*
+	 * In most cases, we don't want to leave one surface and enter
+	 * another while a button is pressed.  However, GTK/Wayland
+	 * menus (implemented as XDG popups) do need the leave/enter
+	 * events to work properly.  To cover this case, we allow
+	 * leave/enter events between an XDG toplevel and its popups.
```
Consolatis, line 77: maybe change to `..between XDG popups and their toplevel`, that would also match the layershell toplevel.

jlindgren90:
@Consolatis If you would like to push a change with the layer-shell adjustments, please feel free; I'm less familiar with those but the suggestions sound fine.

Consolatis:
Managed to crash this:
```
#0  0x0000aaaaaac1edfc in get_xdg_toplevel (surface=0xaaaaab9ef580) at ../src/cursor.c:57
        xdg_surface = 0x0
#1  0x0000aaaaaac1ef6c in is_same_xdg_toplevel (a=0xaaaaab9d30e0, b=0xaaaaab9ef580) at ../src/cursor.c:83
        toplevel_a = 0xaaaaab9d30e0
```
Not sure how this can happen though.. `surface` must have been != `NULL` and `wlr_surface_is_xdg_surface(surface)` must have returned `true` and still `wlr_xdg_surface_from_wlr_surface(surface)` returned `NULL`.
So we might need another `if (!xdg_surface) { return NULL; }` check on top of the loop.

I think I opened a menu normally (including releasing the button), then pressed the button on a disabled (grayed out) menu item and moved back to the parent menu bar.

jlindgren90:
@Consolatis Good points, I am working on addressing them all.

Regarding missing `seat_reset_pressed()`, I think I will add a listener for the `destroy` signal of the `wlr_surface` rather than continuing to sprinkle `seat_reset_pressed()` calls all over.

Also, taking a step back, doesn't `seat.pressed.view` tell us (indirectly) the toplevel already?
Maybe we don't need `get_toplevel()` at all and can just compare views instead.

jlindgren90:
> otherwise when starting the press on a menu and then moving up to the parent (which usually closes the menu), and opening another menu entry the cursor won't be allowed to enter there because the current seat->pressed.surface has been destroyed in the meantime and thus the toplevel doesn't match anymore.

Interestingly, this case doesn't work with `GDK_BACKEND=x11` currently either. I've got it working locally with native Wayland though.

jlindgren90:
Updated, I think I've addressed all the feedback so far.

Consolatis:
Depending on the view breaks this for layer surfaces (they don't have any view).

jlindgren90:
> Depending on the view breaks this for layer surfaces (they don't have any view).

Oh, okay, bummer. I'll go back to the `get_toplevel()` approach.

jlindgren90:
Rebased and restored the `get_toplevel()` approach.

Consolatis:
Thanks, code looks good to me. Will test this later and if I don't see anything wrong will merge.

Consolatis:
No new[^0] issues detected while testing, layershell seems fine in both directions `popup -> parent` and `parent -> popup`. Same for usual XDG views. Nice work!

[^0]: Only detected a pre-existing issue: when starting a click on a layershell popup and then moving out of surface the log gets spammed with `[../src/cursor.c:297] Can't detect surface for out-of-surface movement`

Consolatis:
> I'd like to get #540 merged first. This will need rebasing to fix a merge conflict.

Regarding rebasing: If I have multiple PRs that touch the same area I am usually basing the new PR on top of other existing PRs, mention it in the new PR and mark it as `Draft` until the other ones are merged. At that point rebasing the new PR is usually seamless and consists of just dropping the already merged commits (unless the merged PRs had too much changes).

Not saying that everybody has to do it this way, just wanted to mention what usually works best for me.
Reviewing such PRs is a little more difficult because instead of using the `Files changed` tab on top one has to go through the actual new commits themselves but I would be fine doing that.

PR #535, opened 2022-09-10
cursor: Prevent resetting cursor icon during Move or Resize

Consolatis:
Reported-by: @Flrian

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -365,7 +365,7 @@ msec(const struct timespec *t)
 }
 
 void
-cursor_update_focus(struct server *server)
+cursor_update_focus(struct server *server, bool force_reenter)
 {
 	double sx, sy;
 	struct wlr_scene_node *node = NULL;
@@ -386,7 +386,7 @@ cursor_update_focus(struct server *server)
 	ssd_update_button_hover(node, &seat->server->ssd_hover_state);
 
 	/* Focus surface under cursor if it isn't already focused */
-	cursor_rebase(seat, node, surface, sx, sy, msec(&now), false);
+	cursor_rebase(seat, node, surface, sx, sy, msec(&now), force_reenter);
 }
 
 void
@@ -848,7 +848,7 @@ cursor_axis(struct wl_listener *listener, void *data)
 	wlr_idle_notify_activity(seat->wlr_idle, seat->seat);
 
 	/* Make sure we are sending the events to the surface under the cursor */
-	cursor_update_focus(seat->server);
+	cursor_update_focus(seat->server, false);
 
 	/* Notify the client with pointer focus of the axis event. */
 	wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -38,8 +38,16 @@ cursor_rebase(struct seat *seat, struct wlr_scene_node *node,
 		return;
 	}
 
+	if (seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
+		/* Prevent resetting focus / cursor image when moving or resizing */
+		return;
+	}
+
+	struct wlr_surface *focused_surface =
+		seat->seat->pointer_state.focused_surface;
+
 	if (surface) {
-		if (!force && surface == seat->seat->pointer_state.focused_surface) {
+		if (!force && surface == focused_surface) {
 			/*
 			 * Usually we prevent re-entering an already focused surface
 			 * because it sends useless leave and enter events.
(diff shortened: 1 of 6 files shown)


src/cursor.c
```diff
@@ -56,7 +64,7 @@ cursor_rebase(struct seat *seat, struct wlr_scene_node *node,
 		wlr_seat_pointer_notify_clear_focus(seat->seat);
 		wlr_seat_pointer_notify_enter(seat->seat, surface, sx, sy);
 		wlr_seat_pointer_notify_motion(seat->seat, time_msec, sx, sy);
-	} else {
+	} else if(focused_surface) {
```
jlindgren90, line 67: style nitpick: missing space after `if`

Consolatis: Thanks, fixed

jlindgren90:
I believe we also need a similar check in `request_cursor_notify()` just below, to prevent a client request from changing the cursor:
```
 static void
 request_cursor_notify(struct wl_listener *listener, void *data)
 {
        struct seat *seat = wl_container_of(listener, seat, request_cursor);
+
+       if (seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
+               /* Prevent setting cursor image when moving or resizing */
+               return;
+       }
+
        /*
         * This event is raised by the seat when a client provides a cursor
         * image
```

src/interactive.c
```diff
@@ -134,5 +134,6 @@ interactive_end(struct view *view)
 				view_snap_to_edge(view, "down");
 			}
 		}
+		cursor_set(&view->server->seat, "left_ptr");
```
jlindgren90, line 137: But `left_ptr` may not be the cursor we want, for example when `interactive_end()` ends with the cursor positioned over a text-widget.  Maybe material for another issue/pull request, but perhaps we could put a `FIXME` or `TODO` comment here?

src/cursor.c
```diff
@@ -56,7 +64,7 @@ cursor_rebase(struct seat *seat, struct wlr_scene_node *node,
 		wlr_seat_pointer_notify_clear_focus(seat->seat);
 		wlr_seat_pointer_notify_enter(seat->seat, surface, sx, sy);
 		wlr_seat_pointer_notify_motion(seat->seat, time_msec, sx, sy);
-	} else {
+	} else if (focused_surface) {
 		cursor_set(seat, "left_ptr");
```
jlindgren90, line 68: It might be nice to use the `XCURSOR_DEFAULT` macro here, to be consistent with other call sites such as in `process_cursor_motion()`.

src/interactive.c
jlindgren90: Also we could use `XCURSOR_DEFAULT` here too.

Consolatis:
> I believe we also need a similar check in request_cursor_notify() just below, to prevent a client request from changing the cursor

I think you are right, at least when moving / resizing using the `Alt` modifier. We are not actually clearing the mouse focus on move / resize, we just don't send cursor motion events and instead consume them ourselves. Thus the mouse focused application could in fact send random cursor images because it still "owns" the mouse focus. In comparison, when not using the `Alt` modifier for `Move` / `Resize` and instead using the SSD the application doesn't own the focus in the first place and thus isn't allowed to set cursor images.

Will add the check.

src/interactive.c
Consolatis: When closing the menu we are using a `cursor_rebase(.., true)` which forces a cursor_exit + cursor_enter event which in turn should get the application to set a new cursor image. Maybe we could keep setting the cursor to `left_ptr` (or `XCURSOR_DEFAULT`) in case the cursor is not currently on a surface but additionally force a re-enter for the underlying surface (if there is one) so it has a chance to updates the cursor image.

src/cursor.c
Consolatis: changed both occurrences to `XCURSOR_DEFAULT`

src/interactive.c
jlindgren90: Sounds reasonable. I tried something like that, but couldn't convince myself it was working properly, and didn't pursue it further.

p.s. There is a similar incorrectly hard-coded `XCURSOR_DEFAULT` (and `TODO` comment) in `output_update_for_layout_change()` for what it's worth. I think I'm to blame for that one.

Consolatis:
I think I addressed all the comments. Thanks for the review :)

src/interactive.c
Consolatis: > There is a similar incorrectly hard-coded XCURSOR_DEFAULT

I am not sure I understand the movement part of that one but I also didn't really look into the relation of the "global" seat cursor and output cursors yet.

Regarding the cursor image part of that: I think that one is actually fine but I guess you could throw an `cursor_update_focus(server, true)` in there as well. Just be aware that it may cause hard to diagnose bugs when used during touchpad scrolling, double click or similar cursor actions (see the comment in `cursor_rebase()`).

johanmalm:
Thanks. Seems to work well.

PR #536, opened 2022-09-10
cursor: Prevent setting the same cursor image twice

### LINKED ISSUE #512: Mouse cursor stutters/lags when tabbing out of graphically intense game and moving it over the SSD title bar

Consolatis:
Possibly fixes #512

Reported-by: @Flrian


Consolatis:
wlroots `wlr_xcursor_manager_set_cursor_image()` might actually be a better place for this because AFAIR they can just compare a pointer instead of us comparing strings but i don't feel like writing that up right now.

Issue #246, opened 2022-02-18
Cursor does not change when resizing a window


keshto:
When the window is being resized with the mouse brought to the border it does not change to a resize cursor. This is also the case when mouse is brought to the border. Desired outcome would be for the cursor to change to a resize cursor.

Consolatis:
I never had that issue. Do your mouse cursors change at all?

johanmalm:
What type of window is it (xwayland or xdg). If you alt-tab it'll tell you.
Is window decoration showing.
Are you running from tty or under x11?

johanmalm:
Would you be able to try running tinywl.c to see if that works?
https://gitlab.freedesktop.org/wlroots/wlroots/-/tree/master/tinywl

johanmalm:
See
https://github.com/labwc/labwc/issues/231
https://github.com/labwc/labwc/issues/150

johanmalm:
Does it work with `WLR_NO_HARDWARE_CURSORS=1 labwc`

keshto:
> Does it work with `WLR_NO_HARDWARE_CURSORS=1 labwc`

tried that earlier, no

> Would you be able to try running tinywl.c to see if that works?
https://gitlab.freedesktop.org/wlroots/wlroots/-/tree/master/tinywl

tinywl wouldn't compile for me with wlroots-git so tried 0.15.1 and yes it shows the resize cursors. I recompiled labwc with wlroots 0.15.1 instead of git and still no.

01micko:
Have you tried another cursor theme?

keshto:
I see the resize cursor in tinywl etc. So, dont think that is it, but will give it a try and report back if it works.

ryanlath:
I'm also having this problem, eg no resize, handgrab cursors, etc.

Tried:
- labwc 0.5.0-75-g4ee9f35 / labwc 0.5.0-80-g0604233
- wlroots 0.15.1/0.16
- two different cursor themes
- WLR_NO_HARDWARE_CURSORS=1

`0:00:00.692 [ERROR] [types/output/cursor.c:223] Failed to get cursor display formats`

johanmalm:
@keshto @ryanlath would you mind checking if you still experience this issue with the latest master. We've had some changes to `desktop_node_and_view_at()` which might have resolved this.
If not fixed, would you mind posting your `rc.xml`, `autostart` and `environment` files.

keshto:
[...]

PR #539, opened 2022-09-10
cursor: Handle missing cursor theme

Consolatis:
Temporary fix for #246

This should be reverted once wlroots MR !3651 is merged.

Can be tested nested with `XCURSOR_PATH=/tmp/  labwc` and observing the cursor when moving or resizing a window.

~~This PR is based on top of #535, will rebase once that one is merged.~~


jlindgren90:
I tested this (along with #535 and #536) and it's working great here. I never saw any move/resize cursors before, now they work. Thanks!

Issue #538, opened 2022-09-10
gtklock doesn't grab keyboard focus until the mouse cursor is moved


Flrian:
Especially an issue because for some reason tofi and bemenu still receive what you type until then. Try it with `sleep 5; gtklock`
 and then opening bemenu or tofi before the lock screen appears.

On sway you can immediately enter the password without moving the mouse.

PR #540, opened 2022-09-11
cursor: Factor out cursor_update_common() and fix some glitches

jlindgren90:
Fix a couple of glitches seen when exiting interactive move/resize:

 - Cursor briefly set to `left_ptr` rather than the correct cursor image
 - Cursor not updated if the view being moved/resized is destroyed

Also make sure to exit interactive mode if the view is going fullscreen (labwc gets very confused otherwise).

Code changes in detail:

 - Factor out `set_server_cursor()` which will set the correct cursor image for non-client areas (either `XCURSOR_DEFAULT` or one of the resize cursors).
 - Unify the logic from `cursor_rebase()` and `process_cursor_motion()` by factoring out `cursor_update_common()`.  This corrects some logic discrepancies between the two, which should be a good thing(TM).
 - Remove the extra `cursor_set(XCURSOR_DEFAULT)` from `interactive_end()` and instead rely on `cursor_update_focus()` to do the right thing.
 - Simplify `cursor_button()` by just calling `interactive_end()` when we want to exit interactive mode.
 - Call `cursor_update_focus()` from `view_destroy()` if the view had mouse focus or was being interactively moved/resized.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -80,6 +80,12 @@ struct seat {
 	struct wlr_keyboard_group *keyboard_group;
 
 	bool cursor_requires_fallback;
+	/*
+	 * Name of most recent server-side cursor image.  Set by
+	 * cursor_set().  Cleared when a client surface is entered
+	 * (in that case the client is expected to set a cursor image).
+	 */
+	char *cursor_set_by_server;
 	struct wlr_cursor *cursor;
 	struct wlr_xcursor_manager *xcursor_manager;
 
@@ -533,13 +539,12 @@ void cursor_set(struct seat *seat, const char *cursor_name);
 /**
  * cursor_update_focus - update cursor focus, may update the cursor icon
  * @server - server
- * @force_reenter - re-enter a surface if it already owns the cursor focus
  *
  * This can be used to give the mouse focus to the surface under the cursor
  * or to force an update of the cursor icon by sending an exit and enter
- * event to an already focused surface when @force_reenter is true.
+ * event to an already focused surface.
  */
-void cursor_update_focus(struct server *server, bool force_reenter);
+void cursor_update_focus(struct server *server);
 
 void cursor_init(struct seat *seat);
 void cursor_finish(struct seat *seat);
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -14,10 +14,6 @@
 #include "common/scene-helpers.h"
 #include "common/zfree.h"
 
-/* Used to prevent setting the same cursor image twice */
-static char *last_cursor_image = NULL;
-
-
 static const struct cursor_alias {
 	const char *name, *alias;
 } cursor_aliases[] = {
@@ -47,52 +43,6 @@ is_surface(enum ssd_part_type view_area)
 		|| view_area == LAB_SSD_UNMANAGED
 #endif
 		;
-}
-
-/*
- * cursor_rebase() for internal use: reuses node, surface, sx and sy
- * For a public variant use cursor_update_focus()
- */
-static void
-cursor_rebase(struct seat *seat, struct wlr_scene_node *node,
-		struct wlr_surface *surface, double sx, double sy, uint32_t time_msec,
-		bool force)
-{
-	if (seat->pressed.surface && surface != seat->pressed.surface) {
-		/* Don't leave surface when a button was pressed over another surface */
-		return;
-	}
-
-	if (seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
-		/* Prevent resetting focus / cursor image when moving or resizing */
-		return;
-	}
-
-	struct wlr_surface *focused_surface =
-		seat->seat->pointer_state.focused_surface;
-
-	if (surface) {
-		if (!force && surface == focused_surface) {
-			/*
-			 * Usually we prevent re-entering an already focused surface
-			 * because it sends useless leave and enter events.
-			 *
-			 * They may also seriously confuse clients if sent between
-			 * connected events like a double click (#258) or fast scroll
-			 * events caused by a touchpad (#225).
-			 *
-			 * If we just want to update the cursor image though
-			 * the @force argument may be used to allow re-entering.
-			 */
-			return;
-		}
-		wlr_seat_pointer_notify_clear_focus(seat->seat);
-		wlr_seat_pointer_notify_enter(seat->seat, surface, sx, sy);
-		wlr_seat_pointer_notify_motion(seat->seat, time_msec, sx, sy);
-	} else if (focused_surface) {
-		cursor_set(seat, XCURSOR_DEFAULT);
-		wlr_seat_pointer_notify_clear_focus(seat->seat);
-	}
 }
 
 static void
@@ -127,10 +77,6 @@ request_cursor_notify(struct wl_listener *listener, void *data)
 
 		wlr_cursor_set_surface(seat->cursor, event->surface,
 			event->hotspot_x, event->hotspot_y);
-
-		if (last_cursor_image) {
-			zfree(last_cursor_image);
-		}
 	}
 }
 
@@ -250,16 +196,26 @@ cursor_set(struct seat *seat, const char *cursor_name)
 	}
 
 	/* Prevent setting the same cursor image twice */
-	if (last_cursor_image) {
-		if (!strcmp(last_cursor_image, cursor_name)) {
-			return;
-		}
-		free(last_cursor_image);
+	if (seat->cursor_set_by_server && !strcmp(cursor_name,
+			seat->cursor_set_by_server)) {
+		return;
 	}
-	last_cursor_image = strdup(cursor_name);
 
 	wlr_xcursor_manager_set_cursor_image(
 		seat->xcursor_manager, cursor_name, seat->cursor);
+	zfree(seat->cursor_set_by_server);
+	seat->cursor_set_by_server = strdup(cursor_name);
+}
+
+static void
+set_server_cursor(struct seat *seat, enum ssd_part_type view_area)
+{
+	uint32_t resize_edges = ssd_resize_edges(view_area);
+	if (resize_edges) {
+		cursor_set(seat, wlr_xcursor_get_resize_name(resize_edges));
+	} else {
+		cursor_set(seat, XCURSOR_DEFAULT);
+	}
 }
 
 bool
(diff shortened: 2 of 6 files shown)


src/cursor.c
```diff
@@ -344,50 +361,8 @@ process_cursor_motion(struct server *server, uint32_t time)
...
-		 * Clear pointer focus so future button events and such are not
-		 * sent to the last client to have the cursor over it.
-		 *
-		 * Except if we started pressing a button on the last client and
-		 * are not currently in drag and drop mode.
-		 */
-		wlr_seat_pointer_clear_focus(wlr_seat);
-	}
+	cursor_update_common(server, node, surface, sx, sy, view_area, time,
+		/*force_reenter*/ false, /*cursor_has_moved*/ true);
```
Consolatis, line 365: I love these comments. Will do that myself from now on.

src/view.c
```diff
@@ -850,4 +854,8 @@ view_destroy(struct view *view)
 	/* Remove view from server->views */
 	wl_list_remove(&view->link);
 	free(view);
+
+	if (need_cursor_update) {
+		cursor_update_focus(server, true);
```
Consolatis, line 859: We do we need a `force_reenter` here?

Consolatis:
Very nice! I actually tried to factor out the focus handling some months ago but wasn't satisfied with the result but this seems pretty good. Didn't test nor look in detail into possible logic changes yet, will do that later.

src/view.c
jlindgren90: I was thinking we need it when exiting interactive mode but could be wrong. I'll double-check.

Consolatis: `force_reenter` should only be required if we are already over a surface with active mouse focus but did change the cursor behind the back of the surface for some reason.

In this case the usual mouse focus stuff should just detect a different surface below the mouse (not owning cursor focus), thus send the enter event and let the new surface set the cursor image.

jlindgren90: With SSD resize margins, could we be resizing one surface (the one that gets destroyed here) while another is under the cursor and has mouse focus? That's the only case I can think of where it would matter one way or the other. If that's an impossible case, then I agree, the `force_reenter` here is redundant. It should also be harmless, so on the whole I would be happier keeping it, but go with whatever you think is best.

Consolatis: I think once you move the mouse to reach the resize grabs the cursor focus is cleared (because there is no surface under the cursor). All the following movements are then handled by labwc until interactive_end() at which point the usual cursor focus stuff should kick back in, detect the new surface under the cursor and send enter -> application sets cursor -> business as usual.

At least that is how it works in my head, will verify this point when looking through this in more detail.

src/cursor.c
jlindgren90: C's version of named parameters, haha

jlindgren90:
v2: Eliminate force_reenter parameters and figure out automatically when we need to re-enter the surface.

src/view.c
jlindgren90: I think I came up with a way we can avoid this kind of manual analysis and just let `cursor_update_common()` figure it out for us. See v2.

johanmalm:
I like the look of this. Happy to merge but will hold on in case Consolatis wants to comment/review.

Wee nitpik: would wlr_seat be better than wlseat?

jlindgren90:
> Wee nitpik: would wlr_seat be better than wlseat?

lol, I spent far too much time going back and forth on that.  I like `wlseat` for brevity but it really doesn't matter to me.

johanmalm:
I guess it’s just to avoid giving anyone the idea that it’s a pointer to a “struct wl_seat”.

We ought to really change “struct wlr_seat *seat” to “struct wlr_seat *wlr_seat” in labwc.h - but maybe for a different day.
I have pondered before that the “seat->seat” is a bit confusing.

jlindgren90:
v3: Renamed `wlseat` -> `wlr_seat`.

Consolatis:
Had a first look through the new `cursor.c` and so far the logic seems sound. Also didn't detect any regressions while testing.

include/labwc.h
```diff
@@ -80,6 +80,20 @@ struct seat {
...
+	 *	cursor image.  Set when the client surface is entered.
+	 *	Cleared by cursor_set().
+	 *
+	 * server_cursor_name:
+	 *	Name of most recent server-side cursor image.  Set by
+	 *	cursor_set().  Cleared when a client cursor image is
+	 *	received.
+	 */
+	bool client_cursor_requested;
+	char *server_cursor_name;
```
Consolatis, line 95: Could we combine these two? Not 100% sure right now but I think `server_cursor_name == NULL` should be the same as `client_cursor_requested`.

src/cursor.c
```diff
@@ -250,16 +198,26 @@ cursor_set(struct seat *seat, const char *cursor_name)
 	}
 
 	/* Prevent setting the same cursor image twice */
-	if (last_cursor_image) {
-		if (!strcmp(last_cursor_image, cursor_name)) {
-			return;
-		}
-		free(last_cursor_image);
+	if (!g_strcmp0(seat->server_cursor_name, cursor_name)) {
```
Consolatis, line 201: What is your reasoning for `g_strcmp0` vs `strcmp`?

src/cursor.c
```diff
@@ -307,11 +265,86 @@ process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
...
+				sx, sy);
+		}
+	} else {
+		/*
+		 * Cursor is over a server (labwc) surface.  Clear focus
+		 * from the focused client (if any, no-op otherwise) and
+		 * set the cursor image ourselves.
+		 */
+		wlr_seat_pointer_notify_clear_focus(wlr_seat);
+		set_server_cursor(seat, view_area);
```
Consolatis, line 341: If a surface was blocked due to `input_inhibit` this will clear the focus which might or might not be what we should do.
It is unrelated to the PR though as it was the same before AFAIR. We should still verify what should happen in that case as some follow up.

src/cursor.c
johanmalm: Probably to avoid having to check for NULL pointers

src/view.c
Consolatis: Well done, I think that should work fine for us

src/cursor.c
Consolatis: Hm.. Me personally would prefer to keep it base C where ever possible. I know we are already using glib in other parts for splitting strings and similar but in this case I don't really see an advantage vs using an additional `if (seat->server_cursor_name)`.

include/labwc.h
Consolatis: We could rename the bool to `cursor_set_by_server` though

src/cursor.c
jlindgren90: `g_strcmp0` is more readable in my opinion, but I have no problem spelling out the `NULL` check if that's preferred.
`g_strcmp0` is already used in `is_title_different()` (osd.c) as well, for what it's worth.

src/cursor.c
jlindgren90: #538 might (or might not) be related.

include/labwc.h
jlindgren90: I would like to combine them, didn't really think it through before. I think it will work if we do `zfree(server_cursor_name)` when entering the client surface (in place of `client_cursor_requested = true`).

src/cursor.c
Consolatis: Well, its a project decision /me thinks. As such: thoughts @johanmalm?
Just because I prefer one variant doesn't mean that everybody else has to agree. I am fine with whatever the project lead decides.

include/labwc.h
Consolatis: Good point, that should work.

Consolatis:
Thanks. Merging

johanmalm:
Sorry I’m a bit late to the party. The g_strcmp0 query is a great one. I realise you’ve moved on already, but feel I ought to respond anyway.

The simple answer is that I’m quite relaxed about it and don’t mind a few simple glib functions such as g_strcmp0()… and ahemm… I have let a couple of them in before, not because I don’t care or didn’t notice, but because I thought it was okay.

Having said that, let me outline a few arguments from my perspective to help have a discussion about.

- You’re both key devs and it’s important to me that these types of decisions feel right to you too. We’ve currently not made use of many glib functions and could fairly easily get rid of them at this stage.
- With our use of cairo and pango we depend on glib-2.0 anyway so linking with it and making use of some of its helper functions feels like it comes for free and actually keeps to code simpler.
- If glib wasn’t already in the mix, I think the scope of labwc is such that we could sensibly justify writing a few helpers to avoid the dependency, but as it is I feel it’s sensible to make use of them.
- I don’t know glib well but from the limited exposure I’ve had I’m impressed with it. For example writing [https://github.com/labwc/labwc-menu-generator/blob/master/main.c ](https://github.com/labwc/labwc-menu-generator/blob/master/main.c)would have been harder without glib.
- I started using Address Sanitizer when it was quite new in 2014. Back then glib was quite troublesome with ASan which was a good reason to avoid it, but that doesn’t seem to be the case anymore.
- Some functions (e.g. g_utf8_casefold()) would be pretty hard to write from scratch. That’s when I first started using glib - albeit on a different project.
- I think the first glib function I used on labwc was g_shell_parse_argv() in spawn.c, because parsing all those (double)quotes just seemed like too much effort.
- The use of g_strsplit() in a few places also just seems like good practice to me. The alternative is strtok or a bespoke splitter such as [https://github.com/torvalds/linux/blob/master/tools/lib/argv_split.c.](https://github.com/torvalds/linux/blob/master/tools/lib/argv_split.c) Again, I wouldn’t make a project depend on glib for the convenience of this function alone, but given our eco-system I don’t think we should feel awkward about using it.
- In terms of consistency it would feel wrong to me to start using glib iterators considering we live on wl_list et al.

We ought to add a section in CONTRIBUTING.md on this.

jlindgren90:
> I don’t know glib well but from the limited exposure I’ve had I’m impressed with it.
> Some functions (e.g. g_utf8_casefold()) would be pretty hard to write from scratch.

This echoes my experience. I have been working with GLib in Audacious for over a decade, and in my experience the code and documentation quality is very good. GLib != GTK thankfully, and even after switching away from GTK to Qt, we still use GLib, in particular for UTF-8 handling.

Having some guidelines on "what parts of GLib to use or avoid" would be nice.

Consolatis:
I am not against using glib in labwc where it makes sense (including stuff like splitting, utf8). Usually for me that means not having to reinvent more complex one-off functions (which may introduce further bugs) instead of just using an existing one from glib.

I can only speak from my POV here and seeing a `g_strcmp0()` instead of a  `strcmp()` makes it harder for me to quickly understand the code and its potential side effects because I am not 100% aware of all the subtle differences. This may also go in the other direction, when I associate `*strcmp*` doing an internal `NULL` check I might overlook a missing check when using `strcmp` instead.

johanmalm:
Okay, fair point. I’ll have a go at drafting some guidance in CONTRIBUTING.

PR #541, opened 2022-09-13
cursor: Add struct cursor_context and clean up code

jlindgren90:
The `desktop_node_and_view_at()` function is quite cumbersome to use, because of all the "out" parameters.  It seems like a great candidate to return a `struct` containing all the related information.

Since `desktop_node_and_view_at()` is always used to find the view, node, `ssd_part_type` etc. around the mouse cursor, let's call the new struct `cursor_context` and rename the function to `get_cursor_context()`.  The `cursor_context` struct is also very useful for passing around as a single parameter inside `cursor.c`.

No functional change.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -34,17 +34,6 @@ static const struct cursor_alias {
 	{ "w-resize",   "left_side" },
 };
 
-static bool
-is_surface(enum ssd_part_type view_area)
-{
-	return view_area == LAB_SSD_CLIENT
-		|| view_area == LAB_SSD_LAYER_SURFACE
-#if HAVE_XWAYLAND
-		|| view_area == LAB_SSD_UNMANAGED
-#endif
-		;
-}
-
 static struct wlr_surface *
 get_toplevel(struct wlr_surface *surface)
 {
@@ -254,8 +243,7 @@ input_inhibit_blocks_surface(struct seat *seat, struct wl_resource *resource)
 }
 
 static bool
-update_pressed_surface(struct seat *seat, struct view *view,
-		struct wlr_scene_node *node, struct wlr_surface *surface)
+update_pressed_surface(struct seat *seat, struct cursor_context *ctx)
 {
 	/*
 	 * In most cases, we don't want to leave one surface and enter
@@ -264,11 +252,11 @@ update_pressed_surface(struct seat *seat, struct view *view,
 	 * events to work properly.  To cover this case, we allow
 	 * leave/enter events between XDG popups and their toplevel.
 	 */
-	if (seat->pressed.surface && surface != seat->pressed.surface) {
-		struct wlr_surface *toplevel = get_toplevel(surface);
+	if (seat->pressed.surface && ctx->surface != seat->pressed.surface) {
+		struct wlr_surface *toplevel = get_toplevel(ctx->surface);
 		if (toplevel && toplevel == seat->pressed.toplevel) {
-			seat_set_pressed(seat, view, node,
-				surface, toplevel);
+			seat_set_pressed(seat, ctx->view, ctx->node,
+				ctx->surface, toplevel);
 			return true;
 		}
 	}
@@ -315,15 +303,13 @@ process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
  * Common logic shared by cursor_update_focus() and process_cursor_motion()
  */
 static void
-cursor_update_common(struct server *server, struct view *view,
-		struct wlr_scene_node *node, struct wlr_surface *surface,
-		double sx, double sy, enum ssd_part_type view_area,
+cursor_update_common(struct server *server, struct cursor_context *ctx,
 		uint32_t time_msec, bool cursor_has_moved)
 {
 	struct seat *seat = &server->seat;
 	struct wlr_seat *wlr_seat = seat->seat;
 
-	ssd_update_button_hover(node, &server->ssd_hover_state);
+	ssd_update_button_hover(ctx->node, &server->ssd_hover_state);
 
 	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
 		/*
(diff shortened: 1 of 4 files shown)


johanmalm:
In the change to use the scene-graph API I felt that we left some technical debt in desktop_node_and_view_at(). Simplifying this function involves refactoring/simplifying the calling functions too. It’s not an easy job, but one I think we ought to invest some time in to at least explore what the options are. Always returning node, view, surface and SSD-part does not seem elegant.

Extending the use of node descriptors might be part of the solution. Not sure.

Consolatis:
> Always returning node, view, surface and SSD-part does not seem elegant.

I think we are not even returning a surface but just a node + that strange SSD enum and let each caller figure out how to get the surface.

johanmalm:
> That may be, but I'm very much in favor of incremental improvement and I view this change as such. 

I’m good with that. Just wanted to share what was at the back of my mind regarding that function.

src/cursor.c
```diff
@@ -769,14 +746,8 @@ cursor_button(struct wl_listener *listener, void *data)
...
 
-	struct view *view = desktop_node_and_view_at(server,
-		server->seat.cursor->x, server->seat.cursor->y, &node,
-		&sx, &sy, &view_area);
-
-	struct wlr_surface *surface = NULL;
-	if (is_surface(view_area)) {
-		surface = lab_wlr_surface_from_node(node);
-	}
+	struct cursor_context c = get_cursor_context(server);
```
Consolatis, line 749: Usually I am not much a fan of those single char variable names. I can't think of something better that is not way too long right now though.

Consolatis:
> That may be, but I'm very much in favor of incremental improvement and I view this change as such. Given that we have a function returning several different things, let's at least do it in a maintainable fashion.

I think adding the surface to the new struct would actually allow to clean up some more (e.g. `surface_at()` and the additional `surface` arg for `cursor_update_common).

jlindgren90:
> I think adding the surface to the new struct would allow to clean up some more (e.g. `surface_at()` and the additional `surface` arg for `cursor_update_common()`).

I can add that, unless it goes in the opposite direction of what @johanmalm has in mind?

I have some other refactoring for `cursor.c` waiting as well. I can either wait and open a new pull request after this, or dump it all here if that's better.

src/cursor.c
jlindgren90: `ctx` maybe?

Consolatis:
> I have some other refactoring for cursor.c waiting as well. I can either wait and open a new pull request after this, or dump it all here if that's better.

I think that depends on the changes you have in queue, e.g. if its related to this one or not.

jlindgren90:
v2:
- Add `surface` to `cursor_context`, rename `c` -> `ctx`
- Factor out `determine_resize_edges()`

src/cursor.c
```diff
@@ -336,10 +324,25 @@ cursor_update_common(struct server *server, struct wlr_scene_node *node,
 		 * set the cursor image ourselves.
 		 */
 		wlr_seat_pointer_notify_clear_focus(wlr_seat);
-		set_server_cursor(seat, view_area);
+		set_server_cursor(seat, ctx->type);
 	}
 }
 
+static uint32_t determine_resize_edges(struct wlr_cursor *cursor,
```
Consolatis, line 331: Style nitpick:
```c
static uint32_t
determine_resize_edges(..)
```

I personally would prefer `enum wlr_edges` as return value but I am not sure if the compiler likes the binary ORed together variants for that.

src/cursor.c
```diff
@@ -383,23 +377,15 @@ process_cursor_motion(struct server *server, uint32_t time)
...
-			uint32_t resize_edges = ssd_resize_edges(view_area);
-			if (view && !resize_edges) {
-				resize_edges |= server->seat.cursor->x
-					< view->x + view->w / 2 ? WLR_EDGE_LEFT
-					: WLR_EDGE_RIGHT;
-				resize_edges |= server->seat.cursor->y
-					< view->y + view->h / 2 ? WLR_EDGE_TOP
-					: WLR_EDGE_BOTTOM;
-			}
+			uint32_t resize_edges =
```
Consolatis, line 380: same here in regards to `enum wlr_edges`

src/desktop.c
```diff
@@ -293,92 +293,74 @@ desktop_focus_topmost_mapped_view(struct server *server)
 	desktop_move_to_front(view);
 }
 
-struct view *
-desktop_node_and_view_at(struct server *server, double lx, double ly,
-		struct wlr_scene_node **scene_node, double *sx, double *sy,
-		enum ssd_part_type *view_area)
+/* TODO: make this less big and scary */
```
Consolatis, line 296: /me approves of this comment

src/desktop.c
```diff
@@ -293,92 +293,74 @@ desktop_focus_topmost_mapped_view(struct server *server)
...
-	*scene_node = node;
+	ret.node = node;
 	if (!node) {
-		*view_area = LAB_SSD_ROOT;
-		return NULL;
+		ret.type = LAB_SSD_ROOT;
+		return ret;
 	}
 	if (node->type == WLR_SCENE_NODE_BUFFER) {
 		struct wlr_surface *surface = lab_wlr_surface_from_node(node);
```
Consolatis, line 312: Could we already set the surface here? AFAIR that whole SSD enum check of what is a surface is technical debt from before scene-graph.

jlindgren90: I tried to stay as close to the existing logic as possible here just for fear of not fully understanding the implications, and breaking something. So that meant setting `ret.surface` only in the three cases covered by the old `is_surface()` check:

- `LAB_SSD_CLIENT`
- `LAB_SSD_LAYER_SURFACE`
- `LAB_SSD_UNMANAGED`

I expect there are more cleanups possible here, but I'd rather leave them for a future change.

src/cursor.c
jlindgren90: I'll fix the formatting. `ssd_resize_edges()` already returns a `uint32_t` so I guess we should be consistent with that.

jlindgren90:
Rebased and fixed formatting.

Consolatis:
Went through the diff for a final time and had a last test, did not detect any issues whatsoever.
Merging

PR #542, opened 2022-09-13
cursor: Use enum for server set cursor names

Consolatis:
This mainly prevents having to use `strcmp()` on every mouse move.

Just a rough sketchup for now to gather some feedback, especially about the following:
- `static const char *cursor_{xdg,x11}[]` in a header, guarded by an additional `#ifdef`
- `static const char **cursor_names` in `src/cursor.c` which gets set in `cursor_init()`

I don't want to disturb others working on these parts of the code so I am happy to keep rebasing this if required.

Missing:
- [ ] `uint32_t` -> `enum wlr_edges`

### DIFF
--- a/include/cursor.h
+++ b/include/cursor.h
@@ -1,0 +1,100 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_CURSOR_H
+#define __LABWC_CURSOR_H
+
+#include <wlr/util/edges.h>
+#include "ssd.h"
+
+struct view;
+struct seat;
+struct server;
+struct wlr_surface;
+struct wlr_scene_node;
+
+/* Cursors used internally by labwc */
+enum lab_cursors {
+	LAB_CURSOR_CLIENT = 0,
+	LAB_CURSOR_DEFAULT,
+	LAB_CURSOR_GRAB,
+	LAB_CURSOR_RESIZE_NW,
+	LAB_CURSOR_RESIZE_N,
+	LAB_CURSOR_RESIZE_NE,
+	LAB_CURSOR_RESIZE_E,
+	LAB_CURSOR_RESIZE_SE,
+	LAB_CURSOR_RESIZE_S,
+	LAB_CURSOR_RESIZE_SW,
+	LAB_CURSOR_RESIZE_W,
+	LAB_CURSOR_COUNT
+};
+
+struct cursor_context {
+	struct view *view;
+	struct wlr_scene_node *node;
+	struct wlr_surface *surface;
+	enum ssd_part_type type;
+	double sx, sy;
+};
+
+/**
+ * get_cursor_context - find view and scene_node at cursor
+ *
+ * Behavior if node points to a surface:
+ *  - If surface is a layer-surface, type will be
+ *    set to LAB_SSD_LAYER_SURFACE and view will be NULL.
+ *
+ *  - If surface is a 'lost' unmanaged xsurface (one
+ *    with a never-mapped parent view), type will
+ *    be set to LAB_SSD_UNMANAGED and view will be NULL.
+ *
+ *    'Lost' unmanaged xsurfaces are usually caused by
+ *    X11 applications opening popups without setting
+ *    the main window as parent. Example: VLC submenus.
+ *
+ *  - Any other surface will cause type to be set to
+ *    LAB_SSD_CLIENT and return the attached view.
+ *
+ * Behavior if node points to internal elements:
+ *  - type will be set to the appropriate enum value
+ *    and view will be NULL if the node is not part of the SSD.
+ *
+ * If no node is found for the given layout coordinates,
+ * type will be set to LAB_SSD_ROOT and view will be NULL.
+ *
+ */
+struct cursor_context get_cursor_context(struct server *server);
+
+/**
+ * cursor_set - set cursor icon
+ * @seat - current seat
+ * @cursor_name - name of cursor, for example "left_ptr" or "grab"
+ */
+void cursor_set(struct seat *seat, const char *cursor_name);
+
+/**
+ * cursor_get_resize_edges - calculate resize edge based on cursor position
+ * @cursor - the current cursor (usually server->seat.cursor)
+ * @cursor_context - result of get_cursor_context()
+ *
+ * Calculates the resize edge combination that is most appropriate based
+ * on the current view and cursor position in relation to each other.
+ *
+ * This is mostly important when either resizing a window using a
+ * keyboard modifier or when using the Resize action from a keybind.
+ */
+uint32_t cursor_get_resize_edges(struct wlr_cursor *cursor,
+	struct cursor_context *ctx);
+
+/**
+ * cursor_update_focus - update cursor focus, may update the cursor icon
+ * @server - server
+ *
+ * This can be used to give the mouse focus to the surface under the cursor
+ * or to force an update of the cursor icon by sending an exit and enter
+ * event to an already focused surface.
+ */
+void cursor_update_focus(struct server *server);
+
+void cursor_init(struct seat *seat);
+void cursor_finish(struct seat *seat);
+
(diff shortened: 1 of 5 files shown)


src/cursor.c
```diff
@@ -308,7 +314,7 @@ cursor_update_common(struct server *server, struct wlr_scene_node *node,
 		 * in response to the enter event.
 		 */
 		if (surface != wlr_seat->pointer_state.focused_surface
-				|| seat->cursor_set_by_server) {
+				|| seat->server_cursor) {
```
jlindgren90, line 317: I would make this `seat->server_cursor != LAB_CURSOR_CLIENT`; using the `enum` as a `bool` feels dirty.

src/interactive.c
```diff
@@ -15,7 +15,7 @@ max_move_scale(double pos_cursor, double pos_current,
 }
 
 void
-interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
+interactive_begin(struct view *view, enum input_mode mode, enum wlr_edges edges)
```
jlindgren90, line 18: `interactive_begin()` is passed a `uint32_t` or literal `0` in the places I see.  It would be nice to be consistent -- either stick with `uint32_t` everywhere or switch over to `wlr_edges` everywhere.

jlindgren90:
This is nice if for nothing else than getting rid of the `strdup` and `zfree` :+1:

include/cursor.h
```diff
@@ -0,0 +1,85 @@
...
+	LAB_CURSOR_RESIZE_N,
+	LAB_CURSOR_RESIZE_NE,
+	LAB_CURSOR_RESIZE_E,
+	LAB_CURSOR_RESIZE_SE,
+	LAB_CURSOR_RESIZE_S,
+	LAB_CURSOR_RESIZE_SW,
+	LAB_CURSOR_RESIZE_W
+};
+
+#ifdef __LABWC_IN_CURSOR_C
```
jlindgren90, line 25: I'm not a huge fan of the #ifdef here.  Maybe just put these in `cursor.c`, and leave a comment, "If you add to this enum, remember to update cursors_xdg and cursors_x11 in cursor.c"?

src/interactive.c
Consolatis: Yes you have a point there. I think I'll try to use the enum everywhere, including `WLR_EDGE_NONE` instead of `0`.
Will split that uint32_t -> enum stuff up into an own commit.

include/cursor.h
Consolatis: Yeah.. I'd like to have all the 3 in a single place to make it harder for people to shoot themselves in the foot in the future but I am also not much a fan of that `ifdef` include stuff either.

Trade-off seams to be somewhat along the lines of "don't do something unexpected" and "make sure those structs are always kept in sync". 
Guess a comment on both ends would do as well.

jlindgren90: Do we have access to C11 `static_assert`?  If so, then we could add a final `NUM_LAB_CURSORS` value to the enum and then `static_assert` that the length of the arrays are equal to `NUM_LAB_CURSORS`.

johanmalm: I think so
https://github.com/labwc/labwc/blob/89890b6be9bb79b701b5b257eb8efee0204ec744/meson.build#L8

It’s a nice idea. A bit like linux BUG_ON() macro?

Consolatis: Nice idea. This seems to work fine:
```c
#define ARRAY_SIZE(array) (sizeof(array) / sizeof((array)[0]))
static_assert(ARRAY_SIZE(cursors_xdg) == LAB_CURSOR_COUNT, "XDG cursor names are out ot sync");
static_assert(ARRAY_SIZE(cursors_x11) == LAB_CURSOR_COUNT, "X11 cursor names are out ot sync");
#undef ARRAY_SIZE
```

johanmalm:
Looks very neat.

Consider macro naming consistency with linux kernel.

https://github.com/torvalds/linux/blob/3245cb65fd91cd514801bf91f5a3066d562f0ac4/include/linux/kernel.h#L55

Consolatis:
> Consider macro naming consistency with linux kernel.

I had it as `ARRAY_SIZE` before but changed it to `ARR_LEN` after which made more sense to me. But I am not hooked up on that so will change it back.

Consolatis:
Rebased + changed back to `ARRAY_SIZE`. Still misses splitting the `uint32_t` -> `enum ` changes into its own commit.

src/cursor.c
```diff
@@ -14,26 +14,100 @@
...
+static_assert(
+	ARRAY_SIZE(cursors_xdg) == LAB_CURSOR_COUNT,
+	"XDG cursor names are out ot sync");
+static_assert(
+	ARRAY_SIZE(cursors_x11) == LAB_CURSOR_COUNT,
+	"X11 cursor names are out ot sync");
+#undef ARRAY_SIZE
+
+static enum lab_cursors
+cursor_get_from_ssd(enum ssd_part_type view_area)
```
jlindgren90, line 59: In the interest of code-reuse, I suppose `cursor_get_from_edge(ssd_resize_edges(view_area))` would be equivalent, and save writing another function.

src/cursor.c
```diff
@@ -14,26 +14,100 @@
...
+		return LAB_CURSOR_RESIZE_SE;
+	case WLR_EDGE_BOTTOM:
+		return LAB_CURSOR_RESIZE_S;
+	case WLR_EDGE_BOTTOM | WLR_EDGE_LEFT:
+		return LAB_CURSOR_RESIZE_SW;
+	case WLR_EDGE_LEFT:
+		return LAB_CURSOR_RESIZE_W;
+	default:
+		wlr_log(WLR_ERROR,
+			"Failed to resolve wlroots edge %u to cursor name", edge);
```
jlindgren90, line 106: (Trivial) I'd be inclined to omit the log message here, in such a "utility" style function as this.

src/cursor.c
Consolatis: Yes, thought about that but felt its unnecessary CPU work that has to be done which might be silly.
Would you be fine if I merge these two by having cursor_get_from_ssd call `ssd_resize_edges()` -> `cursor_get_from_edge()`?

jlindgren90: I'm fine with whatever you like best here, it was just an idea that popped into my head, and it sounds like you've already considered it :+1:

Consolatis: Kept the wrapper but replaced the switch() with a call to `ssd_resize_edges()` and `cursor_get_from_edge()`

src/cursor.c
Consolatis: I think we should never receive some invalid edge combination here. If we do, its a bug so I added an `assert(false)` after the error. I am open to change it to just return the default cursor and remove the log though.

include/cursor.h
```diff
@@ -84,6 +84,15 @@ void cursor_set(struct seat *seat, const char *cursor_name);
...
 	struct cursor_context *ctx);
 
+/**
+ * cursor_get_from_edge - translate wlroots edge enum to lab_cursor enum
+ * @edge - wlroots edge enum like WLR_EDGE_TOP | WLR_EDGE_RIGHT
+ *
+ * Returns LAB_CURSOR_DEFAULT if edge is neither
+ * one of the 4 corners nor one of the 4 edges.
+ */
+enum lab_cursors cursor_get_from_edge(enum wlr_edges edge);
```
Consolatis, line 94: Missed that one, will change it to `uint32_t`. I have another branch that changes everything to `enum wlr_edge` but I don't want to mix it in inhere.

jlindgren90:
Code looks good to me; I have not tested the latest version

Consolatis:
> Code looks good to me; I have not tested the latest version

Thanks for review.

Changed `ssd_resize_edge()` to return `WLR_EDGE_NONE` instead of `0` in the `default` case to have it at least somewhat consistent. Will open another PR with the full transition to the enum.

With that: ready to merge from my POV.

PR #544, opened 2022-09-14
cursor: Store view + resize edges for move/resize in press event

### LINKED ISSUE #543: Sometimes the wrong window is resized with resize edges
It's a bit fiddly to reproduce, but occasionally when attempting to resize a window by using left-click-and-drag on a resize edge, the window underneath it gets resized instead.

An example is the second resize attempt in this video:
https://user-images.githubusercontent.com/1244737/190072383-9dd3663c-43a6-48bc-a46f-c63c59f907cc.mp4

The Geany window should be resized from its bottom-right corner, but instead the terminal window underneath is resized.

I think this is because internally, the resize actually starts on the first cursor-motion event (not the cursor-press event), and that's when the `view` and `resize_edges` are chosen.  But when the first motion event is reported, the mouse cursor may have already moved off of the resize-edge node.

To fix this, we probably just need to make sure that the `view` and `resize_edges` used for the interactive resize are computed in the cursor-press event, rather than the first cursor-motion event.

jlindgren90:
Fixes #543.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -166,9 +166,23 @@ show_menu(struct server *server, struct view *view, const char *menu_name)
 }
 
 static struct view *
-activator_or_focused_view(struct view *activator, struct server *server)
+view_for_action(struct view *activator, struct server *server,
+	struct action *action)
 {
-	return activator ? activator : desktop_focused_view(server);
+	/* View is explicitly specified for mousebinds */
+	if (activator) {
+		return activator;
+	}
+
+	/* Select view based on action type for keybinds */
+	switch (action->type) {
+	case ACTION_TYPE_FOCUS:
+	case ACTION_TYPE_MOVE:
+	case ACTION_TYPE_RESIZE:
+		return get_cursor_context(server).view;
+	default:
+		return desktop_focused_view(server);
+	}
 }
 
 void
@@ -194,7 +208,7 @@ actions_run(struct view *activator, struct server *server,
 		 * Refetch view because it may have been changed due to the
 		 * previous action
 		 */
-		view = activator_or_focused_view(activator, server);
+		view = view_for_action(activator, server, action);
 
 		switch (action->type) {
 		case ACTION_TYPE_CLOSE:
@@ -275,7 +289,6 @@ actions_run(struct view *activator, struct server *server,
 			}
 			break;
 		case ACTION_TYPE_FOCUS:
-			view = get_cursor_context(server).view;
 			if (view) {
 				desktop_focus_and_activate_view(&server->seat, view);
 			}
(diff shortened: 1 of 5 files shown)


src/action.c
```diff
@@ -286,7 +285,6 @@ actions_run(struct view *activator, struct server *server,
 			}
 			break;
 		case ACTION_TYPE_MOVE:
-			view = desktop_view_at_cursor(server);
```
Consolatis, line 289: This causes a change in behavior when using `Move` from a keybind.
`Resize` from a keybind seems completely broken but that is a pre-existing issue (this PR would change the behavior of that as well though if it was working).

Maybe we could keep doing the `desktop_view_at_cursor()` calls if `view == NULL`?

We should likely also put a comment in there of the purpose of `desktop_view_at_cursor()` being important when using these actions from a keybind.

jlindgren90: Good catch, I'll fix this to match previous behavior for keybinds (`activator == NULL` case).

jlindgren90:
Rebased and restored the previous behavior for keybinds.

src/action.c
```diff
@@ -166,9 +166,23 @@ show_menu(struct server *server, struct view *view, const char *menu_name)
...
+		return activator;
+	}
+
+	/* Select view based on action type for keybinds */
+	switch (action->type) {
+	case ACTION_TYPE_FOCUS:
+	case ACTION_TYPE_MOVE:
+	case ACTION_TYPE_RESIZE:
+		return get_cursor_context(server).view;
+	default:
```
Consolatis, line 183: Do we need `Raise` here as well?

jlindgren90: I don't personally use any non-default `Focus` and `Raise` actions though, so I'm probably not the target user here ... that said, I would vote for keeping the existing behavior (where `Raise` applies to the focused view), or open a new issue/pull request if there's a good reason to change it.

Consolatis: >  I don't personally use any non-default Focus and Raise actions

Same for me. I would actually like to get rid of those two actions completely and let labwc handle it internally.
Even Openbox warns about custom mouse bindings. But that is a discussion for another time.

Consolatis:
Code LGTM, didn't see any new issues while testing: Merging.

While looking through this I recognized we are looping over all mousebinds on every every cursor move in `process_cursor_motion()` to detect `Drag` events. Maybe we could check for `seat->pressed` (or another bool that gets set on button press and reset on release) before doing that. Moving the cursor is somewhat of a hot path so everything we are not doing there could benefit the "feel" of labwc on a slow system (or when using a high refresh rate mouse).

PR #545, opened 2022-09-15
cursor: Split out cursor_button_press() and cursor_button_release()

jlindgren90:
The control flow in `cursor_button()` is a bit hard to follow, especially the `goto mousebindings` part. Looking at it closely, the logic for press events and release events is different enough that I think it makes sense to split out two smaller functions, `cursor_button_press()` and `cursor_button_release()`.

This includes and builds on #544, which should be merged first.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -637,16 +637,18 @@ cursor_motion_absolute(struct wl_listener *listener, void *data)
 }
 
 static bool
-handle_release_mousebinding(struct view *view, struct server *server,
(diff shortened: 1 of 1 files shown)


Consolatis:
Code LGTM, will test later

johanmalm:
Thank you! Looks so much better.

Consolatis:
And testing didn't show any issues either. So nice to actually get a quick grasp on what is going on in those functions now :)

PR #546, opened 2022-09-15
action: Select resize edges for Resize triggered by keybind

jlindgren90:
@Consolatis noticed in https://github.com/labwc/labwc/pull/544#discussion_r971306613 that `Resize` does not work when triggered from a keybind.  This fixes the issue by making `cursor_get_resize_edges()` public and calling from the keybind path also.

An alternative would be always to use a fixed value of `resize_edges` (e.g. `BOTTOM | RIGHT`) for the keybind case, rather than querying the cursor.

There is a very minor merge conflict with #545, due to renaming `determine_resize_edges()`.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -545,6 +545,9 @@ struct cursor_context get_cursor_context(struct server *server);
  */
 void cursor_set(struct seat *seat, const char *cursor_name);
 
+uint32_t cursor_get_resize_edges(struct wlr_cursor *cursor,
+	struct cursor_context *ctx);
+
 /**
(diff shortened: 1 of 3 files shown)


Consolatis:
LGTM, testing shows it indeed restores resize from a keybind.

Want me to merge this first or #545?

jlindgren90:
If you're good with both, then #545 first please. That's the order I have them in locally.

jlindgren90:
Rebased onto `master`

Consolatis:
And merged. thanks :)

PR #547, opened 2022-09-15
config: Add support for font slant and weight

jlindgren90:
Adds support for bold and italic fonts, following the Openbox syntax.

### DIFF
--- a/include/common/font.h
+++ b/include/common/font.h
@@ -4,10 +4,24 @@
 
 struct lab_data_buffer;
 
+enum font_slant {
+	FONT_SLANT_NORMAL,
+	FONT_SLANT_ITALIC
+};
+
+enum font_weight {
+	FONT_WEIGHT_NORMAL,
+	FONT_WEIGHT_BOLD
+};
+
 struct font {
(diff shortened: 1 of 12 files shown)


include/common/font.h
```diff
@@ -4,11 +4,25 @@
...
+};
+
 struct font {
 	char *name;
 	int size;
+	enum font_slant slant;
+	enum font_weight weight;
 };
 
+struct _PangoFontDescription *font_to_pango_desc(struct font *font);
```
johanmalm, line 24: Is the use of struct here to avoid being pango in the includes?

johanmalm:
LGTM

include/common/font.h
jlindgren90: Yes, especially now that `font.h` is pulled in by `rcxml.h` as well

johanmalm: Nice. It took me a few seconds before the penny dropped. Like it.

johanmalm:
I have update the scope document.
https://github.com/labwc/labwc-scope/commits/main

PR #551, opened 2022-09-16
Fix uses-after-free on exit

jlindgren90:
Fix a couple different use-after-free bugs due to unregistering a `wl_listener` after the event sender is already destroyed.

Off-topic: `QObject::connect()` solved this whole class of bugs a long time ago, by breaking the event connection as soon as either the event sender or recipient is destroyed.  Unfortunately a universal fix of that kind would probably require converting both `wlroots` and `labwc` to C++.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -241,8 +241,15 @@ void
 keyboard_finish(struct seat *seat)
 {
 	if (seat->keyboard_group) {
+		/*
+		 * Caution - these event listeners are connected to
+		 * seat->keyboard_group->keyboard and must be
+		 * unregistered before wlr_keyboard_group_destroy(),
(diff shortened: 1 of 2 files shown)


Consolatis:
How did you spot these? Just by manual analysis?
I am usually compiling at least with `b_sanitize=undefined` and don't get any use-after-free aborts on exit.

jlindgren90:
I used valgrind, it's an awesome tool if you've never used it :)

It found https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3720 that same run too.

Consolatis:
I did use valgrind before but it somehow fails on my current ARM aarch64 system. Might just have to get a newer version.

Thanks, merging

PR #552, opened 2022-09-16
common: Add more friendly memory utilities

jlindgren90:
- Add and use BSD-style `xcalloc`/`xmalloc`/`xrealloc`/`xstrdup`, which immediately `exit()` on error, rather than propagating the `NULL` pointer.  Neither labwc nor most of its dependencies have even remotely correct out-of-memory handling, so `exit()` is the best we can do, realistically.

- Add `MAKE_NEW(struct view, view)` macro as shorthand for the common idiom:
```
    struct view *view = calloc(1, sizeof(struct view))
```
- Add `replace_str(dest, src)` macro as shorthand for the common idiom:
```
    free(dest);
    dest = xstrdup(src);
```
- Reimplement `zfree()` more simply, and in a way that avoids silencing legitimate compile warnings, like `zfree()` of a `const char *`.

The goal is to make it easier to deal with memory allocation in a way that's correct, readable, and consistent.


include/common/mem.h
```diff
@@ -0,0 +1,48 @@
...
+char *xstrdup(const char *str);
+
+/*
+ * Frees <str> and replaces it with a copy of <new_str>.
+ * Sets <str> to NULL only if (new_str == NULL).
+ * Calls exit() on error.
+ */
+#define xreplace_str(str, new_str) do { \
+        free(str); (str) = xstrdup(new_str); \
+} while(0)
```
jlindgren90, line 38: I struggled to come up with a good name for `xreplace_str`. Other ideas I had were `xstrdup_to` or `xstrdup_over`.

Tamaranch:
For what it's worth, GLib already covers at least the first two points: https://developer-old.gnome.org/glib/stable/glib-Memory-Allocation.html

johanmalm:
Happy with xalloc approach. I can’t remember why I didn’t do that from the start. Probably - like you say - because if we run out of memory for the sort of small allocations that labwc makes were pretty screwed anyway.

The following can be useful to get line numbers, etc. but maybe it gets a bit academic:

`#define xrealloc(p, s) (die_on_null(realloc(p, s), __FILE__, __LINE__)) `

I like xzalloc() too - same as xcalloc() but without having to provide 1 as the first argument.

All of the above are so common, I can’t see any C code struggling with the… except possible for the fact that we’re designing the compositor to crash - but at least we’re being explicit about it.

Regarding the MAKE_NEW - my initial reaction is that I prefer:

`struct foo *f = xzalloc(sizeof(struct foo))`

…mostly for readability for anyone other than  us (who would soon get used to it).

Yes, happy with zfree() re-write.

Will have a think about replace_str()

johanmalm:
I’ve never used talloc on anything remotely serious, but think it’s worth exploring.

Rusty’s tal also looks like an option in this space. Close to C++ destructors and RAII?

https://github.com/rustyrussell/ccan/blob/master/ccan/tal/_info

jlindgren90:
> For what it's worth, GLib already covers at least the first two points: https://developer-old.gnome.org/glib/stable/glib-Memory-Allocation.html

I would be fine with using the GLib functions too. In that case I would suggest we make use of `g_new0()`:

`struct view *view = g_new0(struct view, 1);`

src/action.c
```diff
@@ -103,7 +103,7 @@ action_create(const char *action_name)
 		wlr_log(WLR_ERROR, "action name not specified");
 		return NULL;
 	}
-	struct action *action = calloc(1, sizeof(struct action));
+	MAKE_NEW(struct action, action);
```
Consolatis, line 106: I am not a fan if the look of this. I very much prefer `struct action *action = whatever()`

jlindgren90:
> I like xzalloc() too - same as xcalloc() but without having to provide 1 as the first argument.

`xzalloc()` sounds good, then we probably don't need `xcalloc()`.

include/common/mem.h
```diff
@@ -0,0 +1,48 @@
...
+void *xrealloc(void *ptr, size_t size);
+
+/* malloc() is a specific case of realloc() */
+#define xmalloc(size) xrealloc(NULL, (size))
+
+/*
+ * Like strdup(), but calls exit() on error.
+ * Returns NULL only if (str == NULL).
+ */
+char *xstrdup(const char *str);
```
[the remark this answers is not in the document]
jlindgren90: Or leave the `NULL` case for `xreplace_str()` (which is supposed to mimic assignment) to handle.

Consolatis:
> Will have a think about replace_str()

Yeah. I totally see the benefit vs. the usual `free()` + `strdup()`. I am just not sure if the implicitly allowed `NULL` may hide bugs.

include/common/mem.h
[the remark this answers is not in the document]
jlindgren90: > > Will have a think about replace_str()
> 
> Yeah. I totally see the benefit vs. the usual `free()` + `strdup()`. I am just not sure if the implicitly allowed `NULL` may hide bugs.

Honestly I don't think it's any different or worse than normal C pointer assignment in that case.

We can always add an explicit  `assert(str);` in cases where non-NULL is important.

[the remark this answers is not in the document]
Consolatis: Maybe it could even assert on `NULL`, I didn't think this trough though in regards to potential use-cases though.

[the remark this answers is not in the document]
jlindgren90: (reply ended up in the wrong place somehow)

Consolatis:
I think it would make sense to only update the header + implementation of the new macros for now so you don't have to keep changing the whole codebase over and over again.

jlindgren90:
> > remove MAKE_NEW, use struct view *view = xzalloc(sizeof(*view)) instead
> 
> Why do we need the `sizeof()` here? Couldn't that be part of the macro?

It ought to be named something else then to avoid confusion. Other projects already have an `xzalloc(size_t)`.

Consolatis:
> It ought to be named something else then to avoid confusion.

That would be totally fine with me. maybe something less techie like `struct view *view = create_struct(*view)`?

jlindgren90:
> That would be totally fine with me. maybe something less techie like `struct view *view = create_struct(*view)`?

I'm concerned that this is too opaque though. `sizeof(*view)` is a common enough idiom, people should be familiar with it.

`create_struct(*view)` to me is less obvious than the `MAKE_NEW` syntax. Does it deference `*view`? Is it supposed to be `(*view)` or `(view)`?

Consolatis:
We could also use `struct view *some_view = create_struct(struct view)`, the macro would be the same anyway.

jlindgren90:
> I'd prefer a less techie name for struct allocations than [xz][xz]alloc though.

Well, that (and not having to type `struct view` twice) was the goal of `MAKE_NEW`. :)

Consolatis:
> Well, that (and not having to type struct view twice) was the goal of MAKE_NEW. :)

To me that one does feel very alien though (e.g. not like a function returning something).
And thus breaks the usual C assignment like [<definition>] <var> = something

jlindgren90:
Suggestion:
`#define znew(arg) xzalloc(sizeof(arg))`
`struct view *view = znew(struct view)`

Consolatis:
I personally would still prefer
`#define create_struct(args) xzalloc(sizeof(arg))`
`struct view *view = create_struct(struct view)`

I know this might not be very C like (no cryptic `*alloc` or `*new` ) but actually readable and also understandable by people not inherently familiar with decades of `C` work-around patterns. I am not hooked on this though. Lets see what other opinions there are :)

johanmalm:
> I'm leaning towards a v2 with:
>     * `xzalloc()`

Agree

>     * remove `MAKE_NEW`, use `struct view *view = xzalloc(sizeof(*view))` instead

Agree. And kernel coding style prefers `*view` to `struct view`
https://www.kernel.org/doc/html/latest/process/coding-style.html#allocating-memory

>     * add `assert(str)` to `xstrdup()`

Yep

>     * keep `xreplace_str()` NULL-safe though

Suggest `free_and_xstrdup()` to keep code clear for new joiners.
Agree, we should NULL-check before free.

johanmalm:
Thoughts on this alternative?

```
static inline void *die_on_null(void *p, char *file, int line)
{
	if (p) {
		return p;
	}
	if (file) {
		fprintf(stderr, "%s:%d: ", file, line);
	}
	fprintf(stderr, "fatal: out of memory\n");
	exit(EXIT_FAILURE);
}

#define xrealloc(p, s) (die_on_null(realloc(p, s), __FILE__, __LINE__))
```

jlindgren90:
> And kernel coding style prefers `*view` to `struct view`

For now, I'll focus on just replacing the `calloc(1, whatever)` with `xzalloc(whatever)`.  That's an easy find-and-replace anyway, and there doesn't seem to be consensus on what the `whatever` should be, so status quo wins.

Consolatis:
> and there doesn't seem to be consensus on what the whatever should be

Guess I am guilty of that myself: Usually I prefer the `*view` variant but I also committed code that used `struct view` instead.
Is it worth to define this in the code guidelines for the future?

jlindgren90:
> Suggest free_and_xstrdup() to keep code clear for new joiners.

One nitpick is that if `xstrdup()` doesn't allow `NULL`, then it sounds as if `free_and_xstrdup()` doesn't either.
Maybe `zfree_or_strdup()` would still hint at the "zero if NULL, else strdup" aspect?
I would leave out the `x` here just for readability; the name is long enough already.

jlindgren90:
> Is it worth to define this in the code guidelines for the future?

I think I'll state this as my preference, for the record, and then let it go:

```
#define znew(type) ((type *)xzalloc(sizeof(type))
#define znew_n(type, n) ((type *)xzalloc((n) * sizeof(type))

struct view *view = znew(struct view);
uint32_t *u32_array = znew_n(uint32_t, len);
```

It's type-safe (doesn't return `void *`) and also works for other types besides `struct`, e.g. `uint32_t`.  Obviously it's influenced by C++, as well as GLib's `g_new0()`, and thus probably goes against the grain for someone more used to Linux kernel style, but that's the way I lean.

jlindgren90:
> Thoughts on this alternative?
> 
> ```
> static inline void *die_on_null(void *p, char *file, int line)
> ```

Logging file and line is a nice touch.  I probably would put `die_on_null()` in `mem.c` rather than making the whole thing inline.

jlindgren90:
v2:
 - Add `xzalloc()`, remove `xcalloc()`.
 - Remove `MAKE_NEW`.
 - Disallow `xstrdup(NULL)`.
 - Rename `xreplace_str()` to `xstrdup_replace()`.
 - Make self-assignment work: `xstrdup_replace(str, str)`.

jlindgren90:
Maybe ready for a final review pass and merge, if the `xstrdup_replace()` name is acceptable?

src/output.c
```diff
@@ -16,13 +15,10 @@
...
 #include <wlr/util/region.h>
 #include <wlr/util/log.h>
-#include "buffer.h"
+#include "common/mem.h"
 #include "labwc.h"
 #include "layers.h"
-#include "menu/menu.h"
 #include "node.h"
-#include "ssd.h"
-#include "theme.h"
```
Consolatis, line 25: nice cleanup of the `#includes` here

src/xwayland-unmanaged.c
```diff
@@ -139,8 +140,8 @@ struct xwayland_unmanaged *
 xwayland_unmanaged_create(struct server *server,
 			  struct wlr_xwayland_surface *xsurface)
 {
-	struct xwayland_unmanaged *unmanaged;
-	unmanaged = calloc(1, sizeof(struct xwayland_unmanaged));
+	struct xwayland_unmanaged *unmanaged
+		= xzalloc(sizeof(struct xwayland_unmanaged));
```
Consolatis, line 144: nit: We are usually having the `=` at the end of the first line

Consolatis:
Code LGTM, didn' test.

> Make self-assignment work: `xstrdup_replace(str, str)`.

What would be a use case for this?

jlindgren90:
> > Make self-assignment work: `xstrdup_replace(str, str)`.
> 
> What would be a use case for this?

Usually it would be accidental / not obvious.

E.g.
```
const char *title = view->title;
... some code that might or might not change title ...
xstrdup_replace(view->title, title);
```

src/xwayland-unmanaged.c
johanmalm: That reminds me.... haven't run `checkpatch.pl` for a while :smile: 
Incoming

johanmalm:
Agree `zfree_and_strdup()` is best. It give a strong sense that it handles the NULL case.
I think think the current implementation (below) would be better as a function for readaiblity, type-safety, and variable clash-safety.

```
#define xstrdup_replace(str, new_str) do { \
        const char *_xtmp1 = (new_str); \
        char *_xtmp2 = _xtmp1 ? xstrdup(_xtmp1) : NULL; \
        free(str); (str) = _xtmp2; \
} while(0)
```

I think it would also benefit from a comment that describes the arguments more clearly. Maybe:

```
/**
 * zfree_and_strdup - free any current allocation and strdup
 * @ptr: pointer to memory block of C string
 * @s: string to be duplicated
 * Safetly frees @ptr and replaces it with a copy of @s.
 * Note: calls exit() on allocation error.
 */
zfree_and_strdup(char *ptr, const char *s);

```

When we discussed NULL-check in the above, I assumed it was for the initial free().
Do we really want it for the strdup() bit?


> ```
> #define znew(type) ((type *)xzalloc(sizeof(type))
> #define znew_n(type, n) ((type *)xzalloc((n) * sizeof(type))
> 
> struct view *view = znew(struct view);
> uint32_t *u32_array = znew_n(uint32_t, len);
> ```


`struct view *view = znew(*view);` is neat. Happy to go with that.

Interestingly [GLib Reference Manual for g_new0](https://developer.gimp.org/api/2.0/glib/glib-Memory-Allocation.html#g-new0) quotes the following which might be a good reason to stick with a malloc() wrapper (such as znew).

```
It's important to match g_malloc() with g_free(), plain malloc() with free(), and (if you're using C++) new with delete and new[] with delete[]. Otherwise bad things can happen, since these allocators may use different memory pools (and new/delete call constructors and destructors). See also g_mem_set_vtable().
```

Regarding sizeof(*foo) vs sizeof(struct foo), it probably doesn't matter that much. I've drafted a note in CONTRIBUTING.md to clarify a project preference. Interestingly wlroots mixes them wildly :smile:

Regarding [xzalloc](https://github.com/wayland-project/wayland/blob/985ab55d59db45ea62795c76dff5949343e86b2f/src/scanner.c#L278) it's following the Wayland tradition

Agree regarding `fail_on_null()` / `die_on_null()` - it's better not to inline. What I particularly had in mind for that function thought was to feed it `__FILE__` and `__LINE__`. Happy to give up on that one though if you prefer to keep xrealloc() et al as functions.

Thank you for persevering with this. I love getting these details right.

jlindgren90:
≥ Agree zfree_and_strdup() is best. It gives a strong sense that it handles the NULL case.

> Do we really want [NULL transparency] for the strdup() bit?

I guess there is no consensus on name nor behavior for this macro. I'll leave it out and focus on the `xzalloc` stuff.

> zfree_and_strdup(char *ptr, const char *s);

This could work but the first arg would have to be `char**`.

> struct view *view = znew(*view); is neat.

... but would cause a syntax error (invalid cast) with the macro I suggested. Removing the cast loses the type-safety.

jlindgren90:
Really `xstrdup_replace()` is attempt to mimic `std::string::operator=()`.

With the `NULL` safety questions (which C±+ avoids through strong typing) maybe it's a lost cause and better just to be explicit everywhere:

```
zfree(str);
if (newstr) {
    str = xstrdup(new_str);
}
```

johanmalm:
> > zfree_and_strdup(char *ptr, const char *s);
> 
> This could work but the first arg would have to be `char**`.

Ah... of course, yes. Unless we realloc() + memcpy()... would that work?

> > struct view *view = znew(*view); is neat.
> 
> ... but would cause a syntax error (invalid cast) with the macro I suggested. Removing the cast loses the type-safety.

Sorry - forgot about the cast. Well `struct view` is fine. The `*view` thing really doesn't matter.

johanmalm:
> Really `xstrdup_replace()` is an attempt to mimic `std::string::operator=()`.

Got you. Yes, I think quite quickly it's easier to just use GString, C++ std::string or whatever.

I think it's still worth simplifying the chunk below into a oneliner. It's a pretty common idiom in labwc but one that probably doesn't warrant any sophisticated string-handling.

```
zfree(str);
if (new_str) {
    str = xstrdup(new_str);
}
```

jlindgren90:
v3:
 - Fix minor formatting issue
 - Omit `xstrdup_replace()` for now

jlindgren90:
I'll make one last attempt at the others:

```
#define znew(type)       ((type *)xzalloc(sizeof(type))
#define znew_n(type, n)  ((type *)xzalloc((n) * sizeof(type))

static inline char *
xstrdup_or_null(const char *str) {
    return str ? xstrdup(str) : NULL;
}

#define xstrdup_set_or_zfree(str, new_str) do { \
        const char *_xtmp = xstrdup_or_null(new_str); \
        free(str); (str) = _xtmp; \
} while(0)
```

@johanmalm @Consolatis Please give me a yes/no vote on any/all of the above (as-is), and I'll include any that you both want.
Otherwise let's just merge the `xzalloc()` stuff and leave out the rest for now.

johanmalm:
> I'll make one last attempt at the others:
> 
> ```
> #define znew(type)       ((type *)xzalloc(sizeof(type))
> #define znew_n(type, n)  ((type *)xzalloc((n) * sizeof(type))
```

Vote yes.

> static inline char *
> xstrdup_or_null(const char *str) {
>     return str ? xstrdup(str) : NULL;
> }
> 
> #define xstrdup_set_or_zfree(str, new_str) do { \
>         const char *_xtmp = xstrdup_or_null(new_str); \
>         free(str); (str) = _xtmp; \
> } while(0)
> ```

Vote no.

A side note - `zfree(x)` was previously pseudo-ish `if (x) free(x) ; x = NULL`
Throughout the code-base we use zfree() without any `if (whatever) { }` around it, so we really need the `if(x)` bit.

**Comment by jlindgren90** [comment]
> A side note - `zfree(x)` was previously pseudo-ish `if (x) free(x) ; x = NULL` Throughout the code-base we use zfree() without any `if (whatever) { }` around it, so we really need the `if(x)` bit.

`free(NULL)` is already a no-op.

**Comment by johanmalm** [comment]
> `free(NULL)` is already a no-op.

Apologies. Left brain behind.

**Comment by Consolatis** [comment]
```c
#define znew(type) ((type *)xzalloc(sizeof(type))
#define znew_n(type, n) ((type *)xzalloc((n) * sizeof(type))
```
Why the type cast though? It already happens on assignment automatically if the right side is a `void*`.

jlindgren90:
> Why the type cast though? It already happens on assignment automatically if the right side is a `void*`.

It allows the compiler to emit a warning if you do:

    struct a *a = znew(struct b);

Consolatis:
> It allows the compiler to emit a warning if you do:
> ```c
> struct a *a = znew(struct b);
> ```

> Removing the cast loses the type-safety.

Hm.. When just allowing `*view` (without the cast) it would work the same and in fact automatically update with whatever changes of the struct members.

jlindgren90:
> Hm.. When just allowing `*view` (without the cast) it would work the same and in fact automatically update with whatever changes of the struct members.

Echoing back to your `create_struct` idea, we could have a slightly different variant:
```
#define znew_for(ptr) ((type *)xzalloc(sizeof(*(var))))
```

Then we would have (not tested):
```
struct view *view = znew(struct view);  /* works & allows compiler to check type */
struct view *view = znew_for(view);     /* works & infers type */

struct view *view = znew(view);             /* syntax error */
struct view *view = znew_for(struct view);  /* syntax error */
struct view *view = znew_for(*view);        /* syntax error */
```

Consolatis:
You really don't like the `*view` syntax if its not inside a `sizeof()`, do you? :)

> `create_struct(*view)` to me is less obvious than the `MAKE_NEW` syntax. Does it dereference `*view`? Is it supposed to be `(*view)` or `(view)`?

```c
struct foo *bar = znew(*bar);
```
Well it indeed 'dereferences'  the `bar` variable which is defined as a pointer to a `struct foo` so the result would be the same as for `xzalloc(sizeof(struct foo))`. Or `__typeof__(struct foo) == __typeof__(*bar)`.

I actually find the `znew_for(bar)` variant a bit counter intuitive because I would now question if I am getting an allocation with the size of a single pointer back. But if that is defined in the code guidelines I would be ok with this variant as well.

jlindgren90:
> You really don't like the *view syntax if its not inside a sizeof(), do you? :)

I am just afraid someone will forget the `*` and allocate the size of the pointer itself. Maybe compilers nowadays are good enough to catch that but it still makes me nervous. With the `*` inside the macro, in most cases you can't get it wrong.

jlindgren90:
If you and @johanmalm both prefer the `znew_for(*view)` style I am okay with that instead.

Consolatis:
> I am just afraid someone will forget the * and allocate the size of the pointer itself. 

That is a good point.

jlindgren90:
Added `znew/znew_n/znew_for()` in a new commit.

I hope the `__typeof__` is okay; it saved me from one accident already:

```
struct keybind *
keybind_create(const char *keybind)
{
	struct keybind *k = znew_for(keybind); // oops!
```

include/common/mem.h
```diff
@@ -0,0 +1,58 @@
...
+ * Preferred over plain znew() in pointer declarations since it
+ * does not require repeating the type name.
+ *
+ * Example:
+ *   struct wlr_box *box = znew_for(box);
+ *
+ * Note:
+ *   In C23 "typeof" will be standard. Until then, __typeof__ works
+ *   with both GCC and clang (do we care about other compilers?)
+ */
```
Consolatis, line 29: `__typeof__` is already used in `wl_list_for_each` and `FOR_EACH()` in `ssd.h` so its usage is fine IMO.

With the added type check we could go back to `znew_for(*ptr)` though, couldn't we?
E.g.
```c
define znew_for(ptr)  ((__typeof__(&ptr))xzalloc(sizeof(ptr)))
```

Did the conversion for that here to get a feel of the API: https://github.com/Consolatis/labwc/tree/wip/znew_for

jlindgren90: If we could unify the two safely with `__typeof__`, that would be very slick.

jlindgren90:
Unified `znew()` using `__typeof__` according to the excellent suggestion of @Consolatis (thanks!)

include/common/mem.h
```diff
@@ -0,0 +1,51 @@
...
+void *xzalloc(size_t size);
+
+/*
+ * Type-safe macros in the style of C++ new/new[].
+ * <expr> may be either a type name or value expression.
+ *
+ * Examples:
+ *   struct wlr_box *box = znew(*box);
+ *   char *buf = znew_n(char, 80);
```
Consolatis, line 20: another example for `struct wlr_box *box = znew(struct wlr_box);` would be nice here

include/common/mem.h
```diff
@@ -0,0 +1,51 @@
...
+
+/*
+ * As defined in busybox, weston, etc.
+ * Allocates zero-filled memory; calls exit() on error.
+ * Returns NULL only if (size == 0).
+ */
+void *xzalloc(size_t size);
+
+/*
+ * Type-safe macros in the style of C++ new/new[].
```
Consolatis, line 15: IMO we should not refer to another existing standard from another language here and instead explain what we are using it for and why.

johanmalm:
Looks great. Very happy with __typeof__

wayland-util.h uses it so we’re already in the eco-system of GNU-extensions.

https://github.com/wayland-project/wayland/blob/985ab55d59db45ea62795c76dff5949343e86b2f/src/wayland-util.h#L408

Happy to merge.

Will add some notes to CONTR… with our new API when merged.

include/common/mem.h
jlindgren90: I thought our coding style was going to recommend against that style though.

Consolatis:
> Happy to merge.

With those last two comments addressed: +1

include/common/mem.h
jlindgren90: I don't see the harm here, `new` is well-known and what `znew` is modeled after.

include/common/mem.h
Consolatis: With the type check in place I personally don't really care all that much if we 'allow' both variants or just the `znew(*box)` one. We can always update this later on again once we merge #556.

So happy to drop this point for now.

include/common/mem.h
Consolatis: `new` might not be well-known for everybody. And I'd like to keep the code base reasonably understandable by people completely new to C as well. You can't do a `man new` like you can do a `man calloc`.

jlindgren90: That's fair. I'll add some more info to the comment.

jlindgren90: I pushed a new commit, hopefully there's enough info there now for someone not familiar with C++.

jlindgren90:
I updated the comment, hopefully it's good now

Consolatis:
I waited for @johanmalm to merge this but
> Happy to merge

is good enough for me.

Thanks a bunch for working on this, including stating your opinion on the pros and cons of the various options.
I think this turned out pretty good.

I'd be fine discussing some `free_and_strdup()` variant as well in a future Issue / PR.

PR #556, opened 2022-09-17
CONTRIBUTING.md: add some notes on style

johanmalm:


johanmalm:
Whilst we're at it there is an inconsistency in the code-base that would be good to sort out, namely the signal-handler-function naming convention. I'd appreciate thoughts/votes.

Firstly, there is `surface->surface_commit.notify = surface_commit_notify;` where the name of the function is an under-scorified version of the variable.
https://github.com/labwc/labwc/blob/master/src/layers.c

Secondly, there is `view->request_resize.notify = handle_request_resize` where we just use `handle_` as a prefix.
https://github.com/labwc/labwc/blob/master/src/xwayland.c

I don't really mind which, but don't like the inconsistency.

CONTRIBUTING.md
```diff
@@ -52,6 +52,40 @@ If you are maintaining a labwc package for another distro feel free to open an i
...
+
+- We prefer `sizeof(*foo)` to `sizeof(struct foo)`
+
+### The use of glib
+
+We try to keep the use of glib pretty minimal for the following reasons:
+
+- The use of glib has been known to make AddressSanitiser diagnose false
+  positives and negatives.
+- Log messages come glib functions look inconsistent.
```
jlindgren90, line 65: Was this meant to say "Log messages coming from glib functions"?

jlindgren90:
I'm fine with all of this.

Other options for event handler naming:
- `surface_commit_cb`
- `on_surface_commit`

CONTRIBUTING.md
johanmalm: Yes. Force-pushed fix.

Consolatis:
Personally I am used to `on_surface_commit` but I don't have a strong preference. I agree that we should use the same pattern for all of the handlers though, not really invested in which variant.

CONTRIBUTING.md
```diff
@@ -52,6 +52,40 @@ If you are maintaining a labwc package for another distro feel free to open an i
...
+`g_utf8_casefold()` would be pretty hard to write from scratch and are
+fine to use. Having said that, labwc does not do much string-work and thus
+far none of these have been needed.
+
+The following functions are used today and are deemed acceptable by the core
+devs:
+
+- `g_shell_parse_argv()`
+- `g_strsplit()`
+- `g_strcmp0`
```
Consolatis, line 87: > are deemed acceptable by the core devs:

[...]
> `g_strcmp0`

hrm.. by `most of the core devs`? :p

I would still argue against this specific function but if the project lead decides that that specific functions is fine I am fine with it as well and thus `the core devs` would include myself.

CONTRIBUTING.md
```diff
@@ -52,6 +52,40 @@ If you are maintaining a labwc package for another distro feel free to open an i
 Let's try to stick to sircmpwn's [coding style]. If you're not used to it, you
 can use [checkpatch.pl] to run a few simple formatting checks.
 
+Specific nuances that apply to labwc:
+
+- We prefer `sizeof(*foo)` to `sizeof(struct foo)`
```
Consolatis, line 57: We prefer `sizeof(*bar)` to `sizeof(struct foo)` for `struct foo *bar;`

We could also add a reasoning here. Badly written suggestions:
- shorter (especially with very long struct names)
- automatically adapts to struct renaming / change of the type of `bar`

jlindgren90: Probably worth mentioning `znew/znew_for` here also if we decide to add those (related to #552).

Consolatis:
Are there opinions regarding the use of `self`?

johanmalm:
> Are there opinions regarding the use of `self`?

Depends on what you mean by it. As in python-class-self? How would it change current design patterns?
Is this one for IRC?

Consolatis:
> Depends on what you mean by it. As in python-class-self?

Somewhat based on python yes, basically as a way to refer to the "own" object in encapsulated areas. Examples:
- https://github.com/labwc/labwc/blob/b7955ef9a858c0c9d03471fae0f2e5d6a7911919/src/drag_icon.c
- https://github.com/labwc/labwc/blob/master/src/common/scaled_scene_buffer.c

Personally I feel its a bit easier to read than having a `struct scaled_scene_buffer *scaled_scene_buffer = ..` everywhere.

johanmalm:
> > Depends on what you mean by it. As in python-class-self?
> 
> Somewhat based on python yes, basically as a way to refer to the "own" object in encapsulated areas. Examples:
> 
>     * https://github.com/labwc/labwc/blob/b7955ef9a858c0c9d03471fae0f2e5d6a7911919/src/drag_icon.c
> 
>     * https://github.com/labwc/labwc/blob/master/src/common/scaled_scene_buffer.c
> 
> 
> Personally I feel its a bit easier to read than having a `struct scaled_scene_buffer *scaled_scene_buffer = ..` everywhere. But I am fine to stop using that pattern and change the existing occurrences if its deemed to alien for the codebase.

Looks fine to me.

johanmalm:
@jlindgren90 @Consolatis I've added quite a lot to `CONTRIBUTING.md` Appreciate review and opinions.

jlindgren90:
Looks fine to me. I noticed one spelling typo (immitate/imitate) but otherwise +1.

CONTRIBUTING.md
```diff
@@ -1,39 +1,65 @@
-# Contributing
+# How to Contribute
 
-## How to Contribute
+1. Report bugs as github issues. We don't use a template, but try to provide
+   some sensible information such as what what happened, what you expected to
```
Consolatis, line 4: > what what

CONTRIBUTING.md
```diff
@@ -1,39 +1,65 @@
...
+   happen and steps to reproduce. If applicable try with default configuration.
+   If you are able to, try to do some debugging (guidelines below).
 
-## Debugging
+2. Submit patches as github pull-requests. If the patch introduces significant
+   changes or new features, discuss on IRC or via gh issue first to avoid
+   wasting time. Also consult the scope document to ensure it's in keeping.
```
Consolatis, line 10: "wasting time" -> "wasting your time"? Not sure if the first implies the second. Not a native english speaker so this might be totally fine.

> Also consult the scope document to ensure it's in keeping.

Maybe we could link "scope document"?

> to ensure it's in keeping.

"keeping" == Category `A`, `B+` or `B`?

Could

CONTRIBUTING.md
```diff
@@ -1,39 +1,65 @@
...
+[wlhax](https://git.sr.ht/~kennylevinsen/wlhax).
 
-If you think you've got a damage issue, you can run labwc like this: `WLR_SCENE_DEBUG_DAMAGE=highlight labwc` to get a visual indication of damage regions.
+If you think you've got a damage issue, you can run labwc like this:
+`WLR_SCENE_DEBUG_DAMAGE=highlight labwc` to get a visual indication of damage
+regions.
 
-To emulate multiple outputs (even if you only have one physical monitor), run with `WLR_WL_OUTPUTS=2 labwc` are similar. See [wlroots/docs/env_vars.md](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/docs/env_vars.md) for more options.
+To emulate multiple outputs (even if you only have one physical monitor), run
+with `WLR_WL_OUTPUTS=2 labwc` are similar. See
```
Consolatis, line 46: "are" -> "or"

CONTRIBUTING.md
```diff
@@ -1,39 +1,65 @@
...
+[wlroots/docs/env_vars.md](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/docs/env_vars.md)
+for more options.
 
-For some types of bugs, it might be useful to find out which mesa driver (.so) you are using. This can be done with `EGL_LOG_LEVEL=debug labwc 2>&1 | grep MESA-LOADER`
+For some types of bugs, it might be useful to find out which mesa driver (.so)
+you are using. This can be done with `EGL_LOG_LEVEL=debug labwc 2>&1 | grep
+MESA-LOADER`
```
Consolatis, line 52: Should we mention running with pixman as well? E.g. `WLR_RENDERER=pixman`.

CONTRIBUTING.md
```diff
@@ -44,19 +70,97 @@ Some distributions carry labwc in their repositories or user repositories.
...
+	 * longer.
+	 */
+```
+
+`#include` statements at the top of the file are organized by locality (`<>`
+first, then `""`), then alphabetized.
+
+The line length limit is 80 columns, although it is a bit soft.
+
+Tabs are 8 columns wide. Identation with spaces is not used.
```
Consolatis, line 129: Woops, I am actually using 4 columns, should likely adopt to 8 as well to not accidentally overflow the ~80 cols.

CONTRIBUTING.md
```diff
@@ -44,19 +70,97 @@ Some distributions carry labwc in their repositories or user repositories.
...
 
-Let's try to stick to sircmpwn's [coding style]. If you're not used to it, you
-can use [checkpatch.pl] to run a few simple formatting checks.
+## labwc specifics
 
-Specific nuances that apply to labwc:
+We prefer `sizeof(*foo)` over `sizeof(struct foo)`
 
-- We prefer `sizeof(*foo)` to `sizeof(struct foo)`
+We use `if (!p)` instead of `if (p == 0)`
```
Consolatis, line 138: and instead of `if (p == NULL)`

CONTRIBUTING.md
```diff
@@ -86,30 +190,80 @@ devs:
...
+We avoid [GNU C extensions] because we want to fit into the eco-system
+(wayland and wlroots) we live in.
+
+We do use `__typeof__` which strictly speaking is a GNU C extension (`typeof`)
+but through the use of `__` is supported by gcc and clang without defining
+`_GNU_SOURCE`. The justification for this is that Wayland uses it, for example
+in the [`wl_container_of()`] macro which is needed in `wl_list*` and it
+does provide pretty big benefits in terms of type safety.
+
+We compile with `-std=c11` because that's what 'wlroots' and we do not want
```
Consolatis, line 204: missing "uses" or something similar.

CONTRIBUTING.md
```diff
@@ -86,30 +190,80 @@ devs:
...
+
+- Be a short description
+- In most cases be prefixed with "area: " where area refers to a filename
+  or identifier for the general area of the code being modified.
+- Not capitalize the first word following the "area: " prefix, unless
+  it's a name, acronym or similar.
+- Skip the full stop
 
-## Naming Convention
+And please wrap the main commit message at max 74 characters, otherwise
```
Consolatis, line 228: Just for me to understand, "main commit message" == "first line"?

CONTRIBUTING.md
johanmalm: Well... :) we can have a template if you want

CONTRIBUTING.md
jlindgren90: "wasting time" could probably mean "your time", "our time", or both. Which might all be true in some cases :)

CONTRIBUTING.md
Consolatis: No i mean its literally in the changed text: "as what what happened" :D

CONTRIBUTING.md
johanmalm: "wasting time" relates to everyone's time. I've toned it down a bit though to sound more friendly (which is how it is intended)

Have added link to scope document.

"In keeping" = "consistent" / "doesn't deviate". I have refactored the sentence to remove this ambiguous term.

Consolatis: Right but "wasting our time" sounds a bit arrogant, "wasting your time" is sensible. So what I was referring to was to not give the wrong impression of "RTFM to not waste our time".

CONTRIBUTING.md
johanmalm: Have added a note to say "To rule out driver issues you can run with `WLR_RENDERER=pixman labwc`"

CONTRIBUTING.md
johanmalm: I can't believe it :smiley:

CONTRIBUTING.md
johanmalm: Yes, I guess so if p is a pointer. Could be an int. I'll add both.

CONTRIBUTING.md
johanmalm: Thanks

CONTRIBUTING.md
johanmalm: I mean the whole thing. I'll remove the word "main" to make it clearer.
Some projects seem to have a soft limit on the first line of r50-52... but I don't really understand the rationale for this and often find it can be a challenge to write a sensible first line within that tight constraint.

CONTRIBUTING.md
```diff
@@ -44,38 +69,204 @@ Some distributions carry labwc in their repositories or user repositories.
...
+If the statement declares a `{}` block, indent twice instead.
+
+## labwc specifics
+
+We prefer `sizeof(*foo)` over `sizeof(struct foo)`
+
+We use `if (!p)` instead of `if (p == 0)` or `if (p == NULL)`
 
-## Coding Style
+When documenting functions in header files we use the linux Kernel approach.
```
Consolatis, line 139: Capitalized `Linux`? No strong feeling about this though.

Consolatis:
Looks really nice. So I guess whats left is
- handler naming
- `g_strcmp0`

CONTRIBUTING.md
johanmalm: Yes - I'll leave it out to avoid being offensive to anyone... however true it might all be :smile:

johanmalm:
>     * handler naming

I suggest using `handle_` as prefix to be consistent with sway and rootston.

>     * `g_strcmp0`

I'm fine with it, but will remove if you think it's too out of place. Not a big deal for me either way.

Consolatis:
> I suggest using handle_ as prefix to be consistent with sway and rootston.

Fine with me.

> I'm fine with it, but will remove if you think it's too out of place. Not a big deal for me either way.

My issue is mostly about the mixed use of `strcmp` / `strcasecmp` and `g_strcmp0`.
IMHO we should use one variant only so either `g_strmp0` everywhere or `strcmp` everywhere but not use both.

johanmalm:
> My issue is mostly about the mixed use of `strcmp` and `g_strcmp0`. IMHO we should use one variant only so either `g_strmp0` everywhere or `strcmp` everywhere but not use both.

Yes, it does feel inconsistent. Also, to most C coders `!strcmp()` is quite plain but other wrappers, etc just make it hard to read.
I'll remove `g_strcmp0` from the document.

Consolatis:
Sweet, squash / fixup and ready to go?

PR #558, opened 2022-09-18
cursor: Restore drag icon after the move to scene-graph

### LINKED ISSUE #549: Cursor changes to resize image when going over labwc ssd border during drag operation
Using the latest master build.
1. Open for example Nautilus and start dragging something with the mouse cursor
2. When holding it and dragging it over the labwc window decoration border the cursor changes to the resize image and doesn't change back.

https://user-images.githubusercontent.com/4444593/190471485-e0eae687-3e73-4892-90c5-f8d39a76de39.mp4

Consolatis:
This is still very much work-in-progress but at least shows drag icons again.

drag_icon

Missing:
- [x] ~~remove debug logs~~
- [x] ~~figure out why `thunar` doesn't accept its own drag events (e.g. move file into directory). works in sway~~
- [x] ~~figure out why `thunar` doesn't appear to set the cursor icon but only a drag icon. works in sway AFAIR~~
- [x] ~~potentially refactor some stuff~~

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -9,6 +9,7 @@
 #include "common/mem.h"
 #include "common/scene-helpers.h"
 #include "config/mousebind.h"
+#include "dnd.h"
 #include "labwc.h"
 #include "menu/menu.h"
 #include "resistance.h"
@@ -173,21 +174,6 @@ request_set_primary_selection_notify(struct wl_listener *listener, void *data)
 }
 
 static void
-request_start_drag_notify(struct wl_listener *listener, void *data)
-{
-	struct seat *seat = wl_container_of(
-		listener, seat, request_start_drag);
-	struct wlr_seat_request_start_drag_event *event = data;
-	if (wlr_seat_validate_pointer_grab_serial(seat->seat, event->origin,
-			event->serial)) {
-		wlr_seat_start_pointer_drag(seat->seat, event->drag,
-			event->serial);
-	} else {
-		wlr_data_source_destroy(event->drag->source);
-	}
-}
-
-static void
 process_cursor_move(struct server *server, uint32_t time)
 {
 	double dx = server->seat.cursor->x - server->grab_x;
@@ -348,7 +334,7 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 	/* TODO: verify drag_icon logic */
 	if (seat->pressed.surface && ctx->surface != seat->pressed.surface
 			&& !update_pressed_surface(seat, ctx)
-			&& !seat->drag_icon) {
+			&& !seat->drag.active) {
 		if (cursor_has_moved) {
 			/*
 			 * Button has been pressed while over another
@@ -403,7 +389,7 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 		 * a drag operation.
 		 */
 		wlr_seat_pointer_notify_clear_focus(wlr_seat);
-		if (!seat->drag_icon) {
+		if (!seat->drag.active) {
 			cursor_set(seat, cursor_get_from_ssd(ctx->type));
 		}
 	}
@@ -444,6 +430,10 @@ process_cursor_motion(struct server *server, uint32_t time)
 		menu_process_cursor_motion(ctx.node);
 		cursor_set(&server->seat, LAB_CURSOR_DEFAULT);
 		return;
+	}
+
+	if (seat->drag.active) {
+		dnd_icons_move(seat, seat->cursor->x, seat->cursor->y);
 	}
 
 	if (ctx.view && rc.focus_follow_mouse) {
@@ -488,21 +478,6 @@ cursor_update_focus(struct server *server)
 	struct cursor_context ctx = get_cursor_context(server);
 	cursor_update_common(server, &ctx, msec(&now),
 		/*cursor_has_moved*/ false);
-}
-
-void
-start_drag(struct wl_listener *listener, void *data)
-{
-	struct seat *seat = wl_container_of(listener, seat, start_drag);
-	struct wlr_drag *wlr_drag = data;
-	seat_reset_pressed(seat);
-	seat->drag_icon = wlr_drag->icon;
-	if (!seat->drag_icon) {
-		wlr_log(WLR_ERROR,
-			"Started drag but application did not set a drag icon");
-		return;
-	}
-	wl_signal_add(&seat->drag_icon->events.destroy, &seat->destroy_drag);
 }
 
 void
@@ -609,18 +584,6 @@ cursor_motion(struct wl_listener *listener, void *data)
 			event->delta_x, event->delta_y);
 	}
 	process_cursor_motion(seat->server, event->time_msec);
-}
-
-void
-destroy_drag(struct wl_listener *listener, void *data)
-{
-	struct seat *seat = wl_container_of(listener, seat, destroy_drag);
-
-	if (!seat->drag_icon) {
-		return;
-	}
-	seat->drag_icon = NULL;
-	desktop_focus_topmost_mapped_view(seat->server);
 }
 
 void
@@ -1022,6 +985,8 @@ cursor_init(struct seat *seat)
 		cursor_names = cursors_x11;
 	}
 
+	dnd_init(seat);
+
 	seat->cursor_motion.notify = cursor_motion;
 	wl_signal_add(&seat->cursor->events.motion, &seat->cursor_motion);
 	seat->cursor_motion_absolute.notify = cursor_motion_absolute;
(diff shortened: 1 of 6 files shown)


src/desktop.c
```diff
@@ -293,16 +294,34 @@ desktop_focus_topmost_mapped_view(struct server *server)
 	desktop_move_to_front(view);
 }
 
+void
+drag_icons_show(struct seat *seat, bool show)
```
Consolatis, line 298: Not sure if this should be moved to `src/drag_icon.h`

src/cursor.c
```diff
@@ -482,19 +484,43 @@ cursor_update_focus(struct server *server)
...
+	struct wlr_drag *event = data;
 	seat_reset_pressed(seat);
-	seat->drag_icon = wlr_drag->icon;
-	if (!seat->drag_icon) {
-		wlr_log(WLR_ERROR,
-			"Started drag but application did not set a drag icon");
-		return;
+	seat->drag.active = true;
+	if (event->icon) {
+		drag_icon_create(seat, event->icon);
```
Consolatis, line 520: Not sure if this should keep being a `void` function which adds itself to `seat->drag.icons` (and removes itself from there when being destroyed). It seems much less boiler plate in comparison to have `cursor.c` add the returned struct to the list and then listen to the `destroy` signal as well to be able to remove it from the list again.

src/cursor.c
```diff
@@ -432,6 +425,15 @@ process_cursor_motion(struct server *server, uint32_t time)
...
 	struct seat *seat = &server->seat;
 
+	if (seat->drag.active) {
+		struct drag_icon *icon;
+		double lx = seat->cursor->x;
+		double ly = seat->cursor->y;
+		wl_list_for_each(icon, &seat->drag.icons, link) {
+			wlr_scene_node_set_position(&icon->tree->node, lx, ly);
+		}
+	}
```
Consolatis, line 435: Not sure if this should be put into a `drag_icons_update()` or similar and thrown into `src/drag_icon.c` instead.

src/drag_icon.c
```diff
@@ -0,0 +1,80 @@
...
+#include <wlr/util/log.h>
+#include "drag_icon.h"
+#include "labwc.h"  /* for struct seat */
+
+static void
+on_icon_map(struct wl_listener *listener, void *data)
+{
+	struct drag_icon *self = wl_container_of(listener, self, events.map);
+	struct wlr_drag_icon *icon = data;
+	icon->data = wlr_scene_subsurface_tree_create(self->tree, icon->surface);
```
Consolatis, line 14: Should check for `icon->data` already being set..Otherwise map -> unmap -> map feels a bit clunky (won't leak memory though because we are destroying the shared parent).

src/drag_icon.c
```diff
@@ -0,0 +1,80 @@
...
+	struct wlr_drag_icon *icon = data;
+	wlr_scene_node_set_enabled(icon->data, false);
+	wlr_log(WLR_ERROR, "Drag icon unmapped");
+}
+
+static void
+on_icon_destroy(struct wl_listener *listener, void *data)
+{
+	struct drag_icon *self = wl_container_of(listener, self, events.destroy);
+	wlr_scene_node_destroy(&self->tree->node);
```
Consolatis, line 41: Should be moved down so we destroy after signal unsubscribe.

src/drag_icon.c
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void
+drag_icon_create(struct seat *seat, struct wlr_drag_icon *wlr_icon)
+{
+	struct drag_icon *self = calloc(1, sizeof(*self));
+
+	/* Set up parent tree */
+	self->icon = wlr_icon;
+	self->tree = wlr_scene_tree_create(&seat->server->scene->tree);
```
Consolatis, line 59: Potentially create a shared sub tree for all icons instead so we don't have to loop through all of them and instead just position the parent tree based on the cursor coordinates. Same for showing / hiding to prevent being on top when doing hitbox detection.

src/cursor.c
```diff
@@ -482,19 +484,43 @@ cursor_update_focus(struct server *server)
...
+			event->serial)) {
+		wlr_seat_start_pointer_drag(seat->seat, event->drag,
+			event->serial);
+	} else {
+		wlr_data_source_destroy(event->drag->source);
+	}
+}
+
+void
+destroy_drag(struct wl_listener *listener, void *data)
```
Consolatis, line 503: naming + this should be `static`

src/cursor.c
```diff
@@ -1047,18 +1061,18 @@ cursor_init(struct seat *seat)
...
-		&seat->start_drag);
-	seat->destroy_drag.notify = destroy_drag;
 
 	seat->request_set_primary_selection.notify =
 		request_set_primary_selection_notify;
 	wl_signal_add(&seat->seat->events.request_set_primary_selection,
 		&seat->request_set_primary_selection);
+
+	seat->request_start_drag.notify = request_start_drag_notify;
+	seat->start_drag.notify = start_drag;
```
Consolatis, line 1071: move to `seat->drag->events`

src/desktop.c
Consolatis: moved to `src/drag_icon.c`

src/cursor.c
Consolatis: removed completely and replaced by setting the single shared drag icon parent

src/drag_icon.c
Consolatis: `seat->drag.icons` is now the shared parent scene tree for all drag icons.

Consolatis:
Rebased on top of
- #565

01micko:
Seems I can't drag an image to github from `pcmanfm` (xdg-shell) to `vivaldi` either xdg-shell or Xwayland. I'll try Chrome.

01micko:
Just tried with Firefox [xdg-shell] from pcmanfm [xdg-shell] and I can drag an image to github successfully.

Also tested the same with google-chrome [xdg-shell **and** Xwayland] form pcmanfm [xdg-shell] successfully.
spot

And that one is from Vivaldi [xdg-shell]

Drag mimetype icon appears in all cases.

:+1:

Consolatis:
Sweet, thanks for testing.

01micko:
Be nice when you can re-base this one.

Consolatis:
Rebased on latest master.

While testing the rebase I discovered another issue which I am not sure how to react to:
If starting a DnD and releasing the mouse button on the "desktop", the drag operation is still active and thus
- still shows the attached drag image
- when then clicking on another window the drag operation is resumed as if the button was never released in the first place

Based on my intuition I'd assume the proper way to deal with this is to simply abort the DnD operation when the release is on a labwc managed surface like the "desktop" or some part of the SSD.

johanmalm:
> Based on my intuition I'd assume the proper way to deal with this is to simply abort the DnD operation when the release is on a labwc managed surface like the "desktop" or some part of the SSD.

Yes, probably. What about layer-shell surfaces though?

Consolatis:
> Yes, probably. What about layer-shell surfaces though?

IMHO they are perfectly fine DnD sources / targets. Example: moving a `.desktop` file to the panel to create a launcher.
Same for DnD operations between two instances of a panel (like a top and bottom one) or within a single panel (reorder window lists).

src/cursor.c
```diff
@@ -446,6 +432,11 @@ process_cursor_motion(struct server *server, uint32_t time)
 		return;
 	}
 
+	if (seat->drag.active) {
+		wlr_scene_node_set_position(&seat->drag.icons->node,
+			seat->cursor->x, seat->cursor->y);
```
Consolatis, line 437: After some comments from @johanmalm via IRC I moved everything DnD related to `src/dnd.c` so we don't collect random stuff in `src/cursor.c`. This part is directly accessing a `src/dnd.c` "owned" scene node to update its position.

Should I add another wrapper for this in `src/dnd.c` instead? Something like on of these?
```c
dnd_icons_update(struct seat *seat);
dnd_icons_move(struct seat *seat);
dnd_icons_reposition(struct seat *seat);

```

Consolatis: After further discussion via IRC it was decided to use
```c
dnd_icons_move(struct seat *seat, double x, double y);
```

Consolatis:
- Added `dnd_icons_move()`
- Squashed

johanmalm:
> * Added `dnd_icons_move()`
> 
>     * Squashed

Beautiful. Looks good. Let's go for the big button.

01micko:
Thanks for restoring this. Works well in thunar and pcmanfm.
Also works thunar > chrome browsers (several)  that I've tested in Xwayland and native. Thunar is running native.

Issue #561, opened 2022-09-20
Issues with cursor after disabling and re-enabling of output


Flrian:
1. Cursor is always positioned at 0, 0
2. Cursor is hidden until moving
3. If the cursor was on the blank background before on/off its invisible until moving over a window, even while opening the root menu.

Consolatis:
3. sounds partly related to not setting the server cursor to the same it is already set. So we it seems we are missing some reset of `seat->server_cursor` on output loss / restore.

jlindgren90:
The `0, 0` passed to `wlr_cursor_move()` there are delta coordinates -- that should not be moving the cursor at all.

>    Cursor is always positioned at 0, 0
>    Cursor is hidden until moving
>    If the cursor was on the blank background before on/off its invisible until moving over a window, even while opening the root menu.

That sounds like exactly what used to happened before I added the `wlr_cursor_move() + `set_cursor_image()` workaround. Maybe `output_update_for_layout_change()` isn't being called for some reason, or something changed in `wlroots` that broke the workaround?

jlindgren90:
For background, `wlroots` internally has a separate cursor (`wlr_output_cursor` if the comment is still correct) per output. So if you add an output, the cursor on it starts out essentially uninitialized -- at (0,0) and invisible. That's what the `wlr_cursor_move() + set_cursor_image()` workaround was supposed to solve.

jlindgren90:
@Flrian Can you try #562 and see if it helps?

jlindgren90:
@Flrian thanks for testing. Is this a single-monitor setup, or what type of layout do you have? How are you enabling/disabling outputs?

jlindgren90:
I can reproduce the (0,0) issue now that I know it's a single-monitor setup.

PR #562, opened 2022-09-20
output: Call do_output_layout_change() at end of new_output_notify()

jlindgren90:
This fixes an issue with the `wlr_output_cursor` not being properly initialized on new outputs, because `wlr_output_layout.events.change` is triggered too soon, before the `wlr_output_cursor` exists.

May possibly help with #561.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -225,7 +225,14 @@ struct server {
 	struct wl_listener output_layout_change;
 	struct wlr_output_manager_v1 *output_manager;
 	struct wl_listener output_manager_apply;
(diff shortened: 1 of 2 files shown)


Consolatis:
Thanks, this likely also solves an issue for the region PR that I worked around earlier today (output create -> add to layout -> layout_add event ->  apply_region_geometry -> output local regions for the new output not yet initialized).
Will look into this PR in more details later today.

Consolatis:
Code LGTM, Testing didn't show anything suspicious either. Thanks.

PR #564, opened 2022-09-21
Chase wlroots: output cursor

### LINKED ISSUE #561: Issues with cursor after disabling and re-enabling of output

Consolatis:
To update the wlroots subproject use
meson subprojects update wlroots

Chases wlroots 68c8cef38edafa5e52532e0b43a4888554168361
cursor: Don't warp to (0,0) when last output is disconnected

Fixes #561

Thanks @jlindgren90


Consolatis:
CI fails for FeeBSD because:
> wlroots| Dependency libdrm found: NO found 2.4.112 but need: '>=2.4.113'

CI fails for Void because:
> wlroots| Dependency libdrm found: NO found 2.4.112 but need: '>=2.4.113'

Will wait a few days before running the CI pipeline again.

Consolatis:
I had to change the CI FreeBSD runner to use the `latest` port branch instead of the default `quarterly` one to have it provide `libdrm-2.4.113` as required by the latest wlroots `master`.

When we merge this it would mean that labwc master will not work anymore on the default `quarterly` ports branch of FreeBSD. I couldn't figure out when the next `quarterly` update is being done on a quick search. Maybe @jbeich could shed a light on this and advice on the severity of this decision? We are only talking about the labwc master branch + the next release here though (which might take some more time).

.github/workflows/build.yml
```diff
@@ -58,6 +58,11 @@ jobs:
         with:
           usesh: true
           prepare: |
+            mkdir -p /usr/local/etc/pkg/repos
+            sed 's/quarterly/latest/' /etc/pkg/FreeBSD.conf \
+              > /usr/local/etc/pkg/repos/FreeBSD.conf
+            pkg update -f
```
jbeich, line 64: Not necessary as it's automatic.

.github/workflows/build.yml
```diff
@@ -58,6 +58,11 @@ jobs:
...
+            pkg upgrade -y
```
jbeich, line 65: Not necessary as nothing is installed by default. FreeBSD base itself isn't tracked by pkg [yet](https://wiki.freebsd.org/PkgBase).

.github/workflows/build.yml
```diff
@@ -58,6 +58,11 @@ jobs:
         with:
           usesh: true
           prepare: |
+            mkdir -p /usr/local/etc/pkg/repos
+            sed 's/quarterly/latest/' /etc/pkg/FreeBSD.conf \
```
jbeich, line 62: Can be simplified for CI as changes will be discarded on shutdown.

See also https://github.com/Alexays/Waybar/blob/710f89599e87/.github/workflows/freebsd.yml#L20

jbeich (commented):
`/quarterly` changes branches after each quarter (Jan 1, Apr 1, Jul 1, Oct 1), offset by human factor, the scheduled time to build (Tue/Thu/Sat/Sun at 01:00 UTC), whether previous build is still running, the number of changed packages (up to 32k packages), time to build (up to 3 days) and sync to mirrors (up to 30 hours). 2022Q4 will branch around 2022-10-01 with binary packages catching up around 2022-10-07.

jbeich:
For severity, `/quarterly` is very unstable during branch switches while `/latest` is always slightly unstable. The stability mainly impacts big DEs like GNOME and KDE where meta-packages may go missing for a short time due to failed dependencies. Given sway CI and wlroots CI use `/latest` it's unlikely labwc would bump into unique issues.

Note, wlroots supports building libdrm as subproject.

.github/workflows/build.yml
Consolatis: Thank you for your review and advice. We will change to FreeBSD `latest` ports branch for the CI as well then.

PR #565, opened 2022-09-22
[wip] cursor: Refactor cursor_update_common()

Consolatis:
Wlroots resets the focused surface on Drag and Drop.

Before this patch we tried to enforce a client side cursor
image update by forcefully clearing the focus which, even though
`wlr_seat->pointer_state.focused_surface` was `NULL`, was sent to the
drag source. After the clear we are sending an enter event for the
surface under the cursor and if this surface was the drag source itself
it was heavily confused. This caused all kinds of weird behavior from
spawning new drag icons constantly to outright crashing the application.

Reported-by: @Flrian
Bisected-by: @Flrian

The main thing this changes is only doing `wlr_seat_pointer_notify_clear_focus()` when we are on the already focused surface.

I would very much welcome testing of this PR for
- regressions regarding cursor icon changes
- drag and drop

Tested with `WAYLAND_DEBUG=1 foot` to make sure we are not sending unnecessary enter / leave sequences and I didn't detect any. Validation of this point would be appreciated as well.

Would also appreciate a review of @jlindgren90 to make sure this function is still understandable and the logic is sound.

Leaving this as Draft until properly tested for regressions / reviewed.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -345,7 +345,6 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 		return;
 	}
 
-	/* TODO: verify drag_icon logic */
 	if (seat->pressed.surface && ctx->surface != seat->pressed.surface
 			&& !update_pressed_surface(seat, ctx)
 			&& !seat->drag_icon) {
@@ -362,36 +361,7 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 		return;
 	}
 
-	if (ctx->surface && !input_inhibit_blocks_surface(seat,
-			ctx->surface->resource)) {
-		/*
-		 * Cursor is over an input-enabled client surface.  The
-		 * cursor image will be set by request_cursor_notify()
-		 * in response to the enter event.
-		 */
-		if (ctx->surface != wlr_seat->pointer_state.focused_surface
-				|| seat->server_cursor != LAB_CURSOR_CLIENT) {
-			/*
-			 * Enter the surface if necessary.  Usually we
-			 * prevent re-entering an already focused
-			 * surface, because the extra leave and enter
-			 * events can confuse clients (e.g. break
-			 * double-click detection).
-			 *
-			 * We do however send a leave/enter event pair
-			 * if a server-side cursor was set and we need
-			 * to trigger a cursor image update.
-			 */
-			wlr_seat_pointer_notify_clear_focus(wlr_seat);
-			wlr_seat_pointer_notify_enter(wlr_seat, ctx->surface,
-				ctx->sx, ctx->sy);
-			seat->server_cursor = LAB_CURSOR_CLIENT;
-		}
-		if (cursor_has_moved) {
-			wlr_seat_pointer_notify_motion(wlr_seat, time_msec,
-				ctx->sx, ctx->sy);
-		}
-	} else {
+	if (!ctx->surface) {
 		/*
 		 * Cursor is over a server (labwc) surface.  Clear focus
 		 * from the focused client (if any, no-op otherwise) and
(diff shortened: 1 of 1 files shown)


src/cursor.c
```diff
@@ -362,45 +361,56 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 		return;
 	}
 
-	if (ctx->surface && !input_inhibit_blocks_surface(seat,
-			ctx->surface->resource)) {
+	if (!ctx->surface) {
+		/*
+		 * Cursor is over a server (labwc) surface.  Clear focus
```
johanmalm, line 366: Suggest changing `server (labwc) surface` to something more descriptive. I'm not sure what.
Is it server-side-deco only?

jlindgren90: Could also be the root/desktop, menus, etc.

jlindgren90: Well maybe not menus -- I guess those are an interactive mode and handler earlier.

Consolatis:
> So, to make sure I understand the issue, would this also fix it?

Yes, I think so, but didn't test. This was yet another alternative I tested before: https://github.com/Consolatis/labwc/commit/6a9e84f47e70d270ea0d861369f103e066c14025

The main issue with the existing if condition is that it mixes two somewhat unrelated things together
- new surface under cursor
- having a server cursor set

And in both cases uses a clear_focus that seems only required when we already have focus on the current surface but set a server side cursor for some reason. This is also a regression by 6c6e406507bfb0e9a809b74029b5b05b5a095768 for cursor movement in this case.

I was testing this with `thunar` in wayland native mode running on a nested `labwc -d`. With the clear_focus in place it didn't allow to move files into folders via DnD and when starting a DnD operation it was spamming new xdg_surfaces on every mouse move. Nautilus in wayland native mode in comparison appears to even crash at some point (test reports from @Flrian via IRC). Removing clear_focus from the equation apparently also fixed that issue.

When implementing this I also used the chance to simplify the conditions (in my opinion) by reordering. Personally I think the logic is easier to follow now but that is totally up for discussion.

src/cursor.c
Consolatis: "server (labwc) owned surface"

jlindgren90:
> The main issue with the existing if condition is that it mixes two somewhat unrelated things together
>
> - new surface under cursor
> - having a server cursor set

They do often occur together though, and in that case it's nice to send the enter event only once. At least it made sense to have them combined in my mind :)

I guess my preference would be to fix the drag issue minimally, and then do any refactoring in other commits. Mainly in case we have to bisect this later.

There were a couple of additional logic differences, I think, that I wasn't sure about:

- Doing nothing in the inhibit case vs. setting a server/default cursor. Not sure which is right here, but I'm inclined toward setting a default cursor. Otherwise I think we could end up displaying a random one from whatever surface the cursor was over previously.
- I think in the refactored logic we end up calling `notify_enter()` on every cursor movement. This seems unnecessary even if `wlroots` makes it a no-op (which I'm not sure of).

Consolatis:
> I guess my preference would be to fix the drag issue minimally, and then do any refactoring in other commits. Mainly in case we have to bisect this later.

Very good point. Will do that.

> Doing nothing in the inhibit case vs. setting a server/default cursor.

We are also not clearing the focus anymore which I think is the right thing to do. AFAIU inhibited / lockscreen should not allow \*any\* state change whatsoever for surfaces not owning the inhibition / lock, including taking their mouse focus away or changing cursors. But I didn't really investigate this so this might have been a bit premature to change. Another point for just fixing the DnD issue for now.

> I think in the refactored logic we end up calling notify_enter() on every cursor movement. This seems unnecessary even if wlroots makes it a no-op (which I'm not sure of).

At least when debugging with `WAYLAND_DEBUG=1 foot` I didn't spot any additional enter events, and its the same sway does so I just went with it.

https://github.com/swaywm/sway/blob/cffb006feba52c318e66f73c3463032fa76782dc/sway/input/seatop_default.c#L591-L599

src/cursor.c
johanmalm: Reading it through I just wasn't clear on what it means, so I thought that's a good reason for being quite explicit in the comment - maybe be giving an example. Having said that... I don't feel strongly - so go with what you think.

It's slightly non-intuitive that when `ctx->surface==NULL` it might still be a surface just one that is owned by the server.
I haven't looked, but guess this is relating to the clash of surface/node/SDD-enums in the ...node_at() function?

Consolatis: Agree.. that is indeed a bit strange. Not sure what to do to make it more clear though.

Consolatis:
Somewhat superseded by #585 (thanks again).

Will reuse this PR for the refactor to keep the discussion in a single place.

Consolatis:
Closing this for now, it just bitrots anyway.

PR #569, opened 2022-09-26
src/osd.c: Prevent showing invalid window

Consolatis:
This might happen when closing the last
application "to tray" like VLC, Discord or Steam.

Reported-by: @Flrian

@jlindgren90 testing for regressions of the original use-case of detached search window appreciated.


jlindgren90:
@Consolatis I don't see any regressions with this. LGTM.

johanmalm:
Many thanks.

PR #574, opened 2022-09-29
NEWS.md: update notes on 0.6.0

johanmalm:


Consolatis:
This looks like it has been a lot of work. Thanks!

>  Implement virtual keyboard and pointer protocols, enabling the use of
  clients such as wtype.

Maybe add wayvnc as additional example. Allowing to remote in seems like nice feature to have in the changelog :)

PR #580, opened 2022-10-04
Set LABWC_PID

johanmalm:


Consolatis:
LGTM.

Do we want to add an argument for labwc itself to trigger a reload based on the new env var?
Something like `labwc --reload` or `labwc --reconfigure`.

docs/labwc.1.scd
```diff
@@ -16,6 +16,14 @@ It is light-weight and independent with a focus on simply stacking windows
 well and rendering some window decorations. Where practicable it uses clients
 for wall-paper, panels, screenshots and so on.
 
+Exit and reconfigure the compositor on SIGTERM and SIGHUP respectively.
```
jlindgren90, line 19: I get what this means, but I stumble a bit over the wording. A couple of alternatives:

- The compositor will exit or reload its configuration upon receiving SIGTERM or SIGHUP respectively.
- One can exit or reconfigure the compositor by sending it SIGTERM or SIGHUP respectively.

jlindgren90:
LGTM, my only suggestion is re: the documentation (which is probably also fine as-is).

docs/labwc.1.scd
johanmalm: Thanks. I've used the first suggestions. That's much better English.

johanmalm:
> Do we want to add an argument for labwc itself to trigger a reload based on the new env var? Something like `labwc --reload` or `labwc --reconfigure`.

Well, that's a thought. Up until now I've resisted it as I do not see the point in having two ways of doing the same thing.

BUT I suppose the advantage of `labwc --reconfigure` compared with `kill -s SIGHUP $LABWC_PID` is that it can be called without a `sh -c` wrapper to expand the environment variable, which potentially could be quite useful.

And `openbox` does indeed have the `--reconfigure` and `--exit` options, so users do ask for it from time to time :smile:

johanmalm:
I think it's ready, but review welcome

src/main.c
```diff
@@ -31,6 +47,17 @@ usage(void)
...
 
+static void
+send_signal_to_labwc_pid(int signal)
+{
+	char *pid = getenv("LABWC_PID");
+	if (!pid) {
+		wlr_log(WLR_ERROR, "LABWC_PID not set");
+		exit(1);
+	}
+	kill(atoi(pid), signal);
```
Consolatis, line 58: This might call `kill(0, signal)` when `atoi()` fails to parse the content which is a special case:

> If pid equals 0, then sig is sent to every process in the process group of the calling process.

johanmalm:
Updated last commit to not send signal to PID 0

Consolatis:
Sweet, Had a quick test and everything seem to work fine.
Thanks

Issue #572, opened 2022-09-26
Usage of `wl_list_insert(list.prev, element)`


johanmalm:
# Background

[`wayland-util.h`](https://github.com/wayland-project/wayland/blob/master/src/wayland-util.h) provides macros and functions for doubly-linked lists. They are very similar to the Linux kernel [`list.h`](https://github.com/torvalds/linux/blob/master/include/linux/list.h) implementation and are used as the defacto way of doing lists in wlroots and therefore by us.

# The Inconsistency

Consider the following commonly used list functions/macros.

```
wl_list_insert(list, element);

wl_list_for_each(element, list, link) {
	foo(element);
}
```

Anyone not used to this interface could be forgiven for thinking that `wl_list_insert()` adds the element to the end of the list, so that when iterating over the list the oldest element comes first.

Well, it doesn't. `wl_list_insert()` adds the element at the head/front :smile: This is good for implenting stacks (last in first out).

So, if you want a queue (first in first out) - like `.append()` in python - then you've got three options:

1. Use `wl_list_insert(list.prev, element)` (i.e. use `list.prev` instead of `list`). This adds the element to the tail of the list. We do this quite a lot in `labwc`.
2. Use `wl_list_insert()` and then iterate with `wl_list_for_each_reverse()` - basically what most people would probably think of as a queue in reverse.
3. Craft your own `wl_list_insert_tail()` which we did [here](https://github.com/labwc/labwc/blob/d424514e24bbee1c32b19c9192a065c18c9e7fe5/src/desktop.c#L72) and then iterate using `wl_list_for_each()`

Interesting, the Linux kernel calls the equivalent `list_add()` (stack) and `list_add_tail()` (queue).

I did some quick analysis on our `labwc`:

- `ack wl_list_insert src/ | grep -v prev | wc -l` gives 12
- `ack wl_list_insert src/ | grep prev | wc -l` gives 14
- `ack wl_list_for_each_reverse src | wc -l` gives 14

Interestingly wlroots uses wl_list_insert() without `prev` 157 times and with `prev` 17 times. I haven't analysed, but suspect the 157 times is when the order does not matter and the 17 is when they really want a queue.

# Options

1. Ignore this.
2. Make a note in `CONTRIBUTING.md` to say that we use the `link.prev` approach when we desire a queue-like list. Consistent with wlroots.
[...]

PR #581, opened 2022-10-04
Add wl_list_append()

johanmalm:
See #572

### DIFF
--- a/include/common/list.h
+++ b/include/common/list.h
@@ -1,0 +1,19 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#include <wayland-server-core.h>
+
+/**
+ * wl_list_append() - add a new element to the end of a list
+ * @list: list head to add it before
+ * @elm: new element to be added (link of the containing struct to be precise)
+ *
+ * Note: In labwc, most lists are queues where we want to add new elements to
+ * the end of the list. As wl_list_insert() adds elements at the front of the
+ * list (like a stack) - without this helper-function - we have to use
+ * wl_list_insert(list.prev, element) which is verbose and not intuitive to
+ * anyone new to this API.
+ */
+static inline void
+wl_list_append(struct wl_list *list, struct wl_list *elm)
+{
+	wl_list_insert(list->prev, elm);
+}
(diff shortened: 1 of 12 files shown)


include/common/list.h
```diff
@@ -0,0 +1,19 @@
...
+#include <wayland-server-core.h>
+
+/**
+ * wl_list_append() - add a new element to the end of a list
+ * @list: list head to add it before
+ * @elm: new element to be added (link of the containing struct to be precise)
+ *
+ * Note: In labwc, most lists are queues where we want to add new elements to
+ * the end of the list. As wl_list_insert() adds elements at the front of the
+ * list (like a stack) - without this helper-functions - users have to add call
```
Consolatis, line 11: `s/add//`

nit: `s/users/we/` or `s/users have/one has/` ?

src/ssd/ssd_part.c
```diff
@@ -41,7 +42,7 @@ add_scene_part(struct wl_list *part_list, enum ssd_part_type type)
 {
 	struct ssd_part *part = znew(*part);
 	part->type = type;
-	wl_list_insert(part_list->prev, &part->link);
+	wl_list_append(part_list, &part->link);
```
Consolatis, line 45: I like this new wrapper, so much clearer.

Consolatis:
Do you want to get over the existing code and do a mass replace? Otherwise I could do that as well if you prefer.

johanmalm:
> Do you want to go over the existing code and do a mass replace? Otherwise I could do that as well if you prefer.

Feel free :)
I ran out of of energy.

Consolatis:
I think i got everything changed, also reordered some includes while at it.

include/common/list.h
johanmalm: Done

johanmalm:
I'll merge this for now.
I did a search for `wl_list_for_each_reverse` and suspect most of them should drop the reverse (and suspect the ones in `scaled_scene_buffer.c` are the exception).
I'd rather do that quite slowly and testing thoroughly thought :smile:

Issue #339, opened 2022-05-24
Cursor disappears in Wine games


acoolstraw:
Cursor disappears in (seemingly) only Wine games when you click somewhere. The cursor stays where you clicked, but it doesn't move anymore, it's still functional though and reappears when alt tab out and back, then disappears again once you click on something.

Bug happens in Wreckfest, CS 1.6 and other Wine/Proton games. Interestingly, this bug doesn't happen if you have multiple displays/outputs. `WLR_NO_HARDWARE_CURSORS` doesn't make a difference.

Relwi:
It happens to me too, I only test with Steam games, in most games there is no issues with the cursor, but in some other games (Path of Exile, Guild Wars 2, Apex Legends) the cursor become invisible, the mouse actions works, but the cursor is invisible (only when playing at Fullscreen and Windowed Fullscreen).

acoolstraw:
Looking around in the ArchWiki, I found this. Could be a Wayland thing, not exclusive to labwc

https://wiki.archlinux.org/title/sway#Mouse_not_working_in_WINE_applications

Ph42oN:
I got this issue in Planetside 2. It only happens when i run it in fullscreen mode, not in windowed or windowed fullscreen. I tried looking at the code to find the cause. Commenting out this if in cursor.c in cursor_motion function makes mouse cursor move as expected:
`if (!seat->current_constraint) {`
So calling wlr_cursor_move even with constraints makes cursor work normally.

FuzzyQuills:
In my case, it seems the cursor doesn't vanish but gets stuck, yet moving the mouse itself around reveals that the actual cursor is still moving. (Overwatch 2)

Initially thought it was a GPU driver bug until I found this. Gamescope it is then.

Consolatis:
This should be fixed by #586.
If it isn't feel free to comment and we'll reopen this issue.

Issue #534, opened 2022-09-10
Grab mouse in xwayland (games) with multi-monitor setup


KaitoKariheddo:
Playing games on multi-monitor setups result in mouse moving from one ot other monitor, e.g. if you look around ingame. 
According to the Arch Wiki wlroots already implement a catching extension. https://wiki.archlinux.org/title/wayland#Input_grabbing_in_games,_remote_desktop_and_VM_windows
Is it possible to activate this extension in labwc ?

Consolatis:
Cursor constrains should be activated but its possible we are doing something incorrectly.

You can try to run your games via gamescope compositor on top of labwc as a work-around for now.

KaitoKariheddo:
Thank you, gamescope works fine! Would be nice if you find the bug.

Ph42oN:
I would like to see this feature too, as sometimes when i click outside game, my mouse still moves camera in game but keyboard goes into something else. I tried gamescope but it is not an option for me because adaptive sync does not work with it.

Consolatis:
This should be fixed by #586.
If it isn't feel free to comment and we'll reopen this issue.

Issue #475, opened 2022-08-09
Pointer constraints seem to be broken, especially when using wlr layer shell protocol


Tamaranch:
Labwc 0.5.3 on Arch Linux

I use a minimal example with a GtkWindow and a GtkButton that triggers the constraint when pressed, and removes it when released. I can provide the code if needed.

As is, the confinement constraint has the same effect as the lock constraint: the pointer is effectively locked, and the "motion" signal is no longer emitted.

If one activates the wlr layer shell protocol (in my case via the Gtk Layer Shell library, by a call to `gtk_layer_init_for_window()`), neither constraint works, the "confined" and "locked" signals are not emitted.

All this works normally on Wayfire for example.

Consolatis:
> I can provide the code if needed.

That would be welcome. Having a minimal example should speed things up.

Tamaranch:
Here is a test code that should do the trick I think.

<p>
<details>
<summary>p.c</summary>

```c
#include <gtk/gtk.h>
#include <gdk/gdkwayland.h>
#include <gtk-layer-shell/gtk-layer-shell.h>
#include "pointer-constraints-unstable-v1-client.h"

static struct wl_registry *wl_registry;
static struct zwp_pointer_constraints_v1 *zwp_pointer_constraints = NULL;
static struct zwp_confined_pointer_v1 *zwp_confined_pointer = NULL;
static struct zwp_locked_pointer_v1 *zwp_locked_pointer = NULL;
static gboolean surface_is_layer = FALSE;
static gboolean pointer_is_locked = FALSE;
static gboolean motion_emitted = FALSE;

static void
panel_window_wl_registry_global (void               *data,
                                 struct wl_registry *registry,
                                 uint32_t            id,
                                 const char         *interface,
                                 uint32_t            version)
{
  if (g_strcmp0 (interface, "zwp_pointer_constraints_v1") == 0)
    zwp_pointer_constraints =
      wl_registry_bind (registry, id, &zwp_pointer_constraints_v1_interface,
                        MIN ((uint32_t) zwp_pointer_constraints_v1_interface.version, version));
}

static void
panel_window_wl_registry_global_remove (void               *data,
                                        struct wl_registry *registry,
                                        uint32_t            id)
{
}

static const struct wl_registry_listener wl_registry_listener =
{
  panel_window_wl_registry_global,
  panel_window_wl_registry_global_remove
};

static void
panel_window_zwp_confined_pointer_confined (void                           *data,
[...]

PR #586, opened 2022-10-12
Mouse constraints implementation

Ph42oN:
I looked how this was done in sway and did it in labwc. This fixes issues #339 and #534 for me.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -595,6 +595,7 @@ apply_constraint(struct seat *seat, struct wlr_pointer *pointer, double *x, doub
 	if (!seat->current_constraint || pointer->base.type != WLR_INPUT_DEVICE_POINTER) {
 		return;
 	}
+	assert(seat->current_constraint->type == WLR_POINTER_CONSTRAINT_V1_CONFINED);
 
 	double sx = seat->cursor->x;
 	double sy = seat->cursor->y;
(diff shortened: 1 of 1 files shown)


Consolatis:
First of all: thanks for working on this.

I had a quick test and it doesn't seem to confine the cursor correctly, it can be tested via the [example application here](https://github.com/labwc/labwc/issues/475#issuecomment-1277924728). It also breaks the same test with `--locked`.

I had a quick try in figuring out how all this is supposed to work but failed. I might look into this some more during the next days, if somebody else wants to work on this feel free.

johanmalm:
Thanks both. I’m just having a busy few days, so not able to look at it just now.

Ph42oN:
Looks like it actually only makes cursor stay within window area for fullscreen applications. I did this in simpler way than it was done in sway, so i think there is something missing.

Ph42oN:
I made constraints work for windowed.

Consolatis:
This looks very promising!

It is a bit hard to test on a nested instance as that one uses absolute coordinates but this seems to completely fix the pointer constraints for windows. Nice work.

I added back the lock confinement and will test later on an actual DRM session.
More testers would be welcome.

We may need to restore the pixman dependency that we dropped some time ago but that can be handled once this is merged.

johanmalm:
LGTM
Have tested with @Consolatis 's test application https://github.com/labwc/labwc/issues/475#issuecomment-1277924728
Agree we should squash.

Consolatis:
Cleaned up the history, this seems ready to go.

Thanks a bunch @Ph42oN!

johanmalm:
Thank you both. Really nice :smile:

PR #592, opened 2022-10-14
Refuse to start when SUID is detected

johanmalm:


Consolatis:
Code looks good to me, didn't test.

Regarding the feature itself I think it should be fine to remove `suid` shenanigans.
Also it was telling the user of the deprecation already.

We could actually add a warning when we detect that labwc is running as root as well.

PR #596, opened 2022-10-26
add support for mouse buttons UP and DOWN (scroll)

Arnaudv6:
Hi, C junior here: bear with me, thanks for awesome project
I'll update the documentation after review.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -219,6 +219,8 @@ The rest of this man page describes configuration options.
 	- Left
 	- Middle
 	- Right
+	- Up (scroll up)
+	- Down (scroll down)
 
 	Supported mouse actions include:
 	- Press: Pressing the specified button down in the context.
@@ -226,6 +228,7 @@ The rest of this man page describes configuration options.
 	- Click: Pressing and then releasing inside of the the context.
 	- DoubleClick: Two presses within the doubleClickTime.
 	- Drag: Pressing the button within the context, then moving the cursor
+	- Scroll: Scrolling up or down in the context.
 
 *<mouse><default />*
 	Load default mousebinds. This is an addition to the openbox
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -301,6 +301,12 @@
       <mousebind button="Middle" action="Press">
         <action name="ShowMenu"><menu>root-menu</menu></action>
       </mousebind>
+      <mousebind button="Up" action="Scroll">
+        <action name="GoToDesktop" to="left"/>
+      </mousebind>
+      <mousebind button="Down" action="Scroll">
+        <action name="GoToDesktop" to="right"/>
+      </mousebind>
     </context>
 
   </mouse>
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -882,6 +882,42 @@ cursor_button(struct wl_listener *listener, void *data)
 	}
 }
 
+bool
+handle_cursor_axis(struct server *server, struct cursor_context *ctx,
+		struct wlr_pointer_axis_event *event)
+{
+	struct mousebind *mousebind;
+	bool activated_any = false;
+
+	uint32_t modifiers = wlr_keyboard_get_modifiers(
+			&server->seat.keyboard_group->keyboard);
+
+	uint32_t button = 0;
+	if (event->delta_discrete < 0) {
+		button = BTN_GEAR_UP;
+	} else if (event->delta_discrete > 0) {
+		button = BTN_GEAR_DOWN;
+	}
+
+	if (!button || event->source != WLR_AXIS_SOURCE_WHEEL
+			|| event->orientation != WLR_AXIS_ORIENTATION_VERTICAL) {
+		wlr_log(WLR_DEBUG, "Failed to handle cursor axis event");
+		return false;
+	}
+
+	wl_list_for_each(mousebind, &rc.mousebinds, link) {
+		if (ssd_part_contains(mousebind->context, ctx->type)
+				&& mousebind->button == button
+				&& modifiers == mousebind->modifiers
+				&& mousebind->mouse_event == MOUSE_ACTION_SCROLL) {
+			activated_any = true;
+			actions_run(ctx->view, server, &mousebind->actions, /*resize_edges*/ 0);
+		}
+	}
+
+	return activated_any;
+}
+
 void
 cursor_axis(struct wl_listener *listener, void *data)
 {
@@ -891,15 +927,22 @@ cursor_axis(struct wl_listener *listener, void *data)
 	 */
 	struct seat *seat = wl_container_of(listener, seat, cursor_axis);
 	struct wlr_pointer_axis_event *event = data;
+	struct server *server = seat->server;
+	struct cursor_context ctx = get_cursor_context(server);
 	wlr_idle_notify_activity(seat->wlr_idle, seat->seat);
 
-	/* Make sure we are sending the events to the surface under the cursor */
-	cursor_update_focus(seat->server);
+	/* Bindings swallow mouse events if activated */
+	bool triggered_axis_binding = handle_cursor_axis(server, &ctx, event);
 
-	/* Notify the client with pointer focus of the axis event. */
-	wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
-		event->orientation, event->delta, event->delta_discrete,
-		event->source);
+	if (ctx.surface && !triggered_axis_binding) {
+		/* Make sure we are sending the events to the surface under the cursor */
+		cursor_update_common(server, &ctx, event->time_msec, false);
+
+		/* Notify the client with pointer focus of the axis event. */
+		wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
+			event->orientation, event->delta, event->delta_discrete,
+			event->source);
+	}
 }
 
 void
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -634,8 +634,8 @@ handle_release_mousebinding(struct server *server,
 		struct cursor_context *ctx, uint32_t button)
 {
 	struct mousebind *mousebind;
-	bool activated_any = false;
-	bool activated_any_frame = false;
+	bool handled = false;
+	bool handled_in_compositor = false;
 
 	uint32_t modifiers = wlr_keyboard_get_modifiers(
 			&server->seat.keyboard_group->keyboard);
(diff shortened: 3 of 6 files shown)


src/cursor.c
```diff
@@ -891,15 +935,24 @@ cursor_axis(struct wl_listener *listener, void *data)
...
+
+	if (ctx.surface && !triggered_axis_binding) {
+		/* Notify client of axis event */
 
-	/* Notify the client with pointer focus of the axis event. */
-	wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
-		event->orientation, event->delta, event->delta_discrete,
-		event->source);
+		/* Make sure we are sending the events to the surface under the cursor */
+		cursor_update_focus(seat->server);
```
Arnaudv6, line 949: Should use server here, now

johanmalm:
Thanks for working on this. Looks great.
I’m having a busy few day, so unless someone has got time to review, I probably won't get to it until the weekend.

src/cursor.c
```diff
@@ -882,6 +882,50 @@ cursor_button(struct wl_listener *listener, void *data)
...
+{
+	struct mousebind *mousebind;
+	bool activated_any = false;
+	bool activated_any_frame = false;
+
+	uint32_t modifiers = wlr_keyboard_get_modifiers(
+			&server->seat.keyboard_group->keyboard);
+
+	uint32_t button = 0;
+	if (event->delta_discrete<0) {
```
Consolatis, line 897: Please use additional spaces around `<`, same for the next branch

src/cursor.c
```diff
@@ -882,6 +882,50 @@ cursor_button(struct wl_listener *listener, void *data)
...
+		button = BTN_GEAR_UP;
+	} else if (event->delta_discrete>0) {
+		button = BTN_GEAR_DOWN;
+	}
+
+	if (!button
+			|| event->source != WLR_AXIS_SOURCE_WHEEL
```
Consolatis, line 904: First two conditions can be combined, like
```c
if (!button || event->source != WLR_AXIS_SOURCE_WHEEL
		|| event->orientation [..]
```

src/cursor.c
```diff
@@ -882,6 +882,50 @@ cursor_button(struct wl_listener *listener, void *data)
...
+			|| event->orientation != WLR_AXIS_ORIENTATION_VERTICAL) {
+		wlr_log(WLR_ERROR, "Failed to handle cursor axis event");
```
Consolatis, line 906: Not sure we need the `WLR_ERROR` here, `WLR_DEBUG` might be enough (if the log is required at all).
In what cases can this go wrong?

src/cursor.c
```diff
@@ -882,6 +882,50 @@ cursor_button(struct wl_listener *listener, void *data)
...
+				activated_any_frame |= mousebind->context == LAB_SSD_FRAME;
+				actions_run(ctx->view, server, &mousebind->actions,
+					/*resize_edges*/ 0);
+			default:
+				continue;
+			}
+		}
+	}
+
+	return activated_any || activated_any_frame;
```
Consolatis, line 926: `activated_any_frame` seems redundant, all cases which set it also set `activated_any`.

Also not a fan of the naming in general, not specific to this PR but `activated_something` doesn't exactly say what that variable does. Any suggestions for a more intuitive name? Maybe something like `handled` ?

src/cursor.c
```diff
@@ -891,15 +935,24 @@ cursor_axis(struct wl_listener *listener, void *data)
...
+
+	if (ctx.surface && !triggered_axis_binding) {
+		/* Notify client of axis event */
 
-	/* Notify the client with pointer focus of the axis event. */
-	wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
-		event->orientation, event->delta, event->delta_discrete,
-		event->source);
+		/* Make sure we are sending the events to the surface under the cursor */
+		cursor_update_focus(server);
```
Consolatis, line 949: We could likely use the `cursor_update_common()` variant here instead as we already have the cursor context and thus don't have to do the hitbox detection again.

Consolatis (changes requested):
Thanks for working on this. Just a quick code review, I am a bit short on time as well currently so I didn't spent too much time to verify the logic itself nor did I test the PR. Overall I think this looks pretty good :)

Also didn't verify that `Up` and `Down` are the correct Openbox names for mouse wheel events.

Just curious, what are you using the bindings for?

src/cursor.c
Arnaudv6: Right!
I guess one not supported case would be wlr_pointer_axis_event containing
delta rather than delta_discrete (continuous mouse wheel?)?
Changed log to WLR_DEBUG, but will gladly remove it completely, your call.

Consolatis:
I had a quick test (using your waypiedock + commenting out the swaymsg calls) and this indeed seems to work:

```xml
	<mouse>
		<default />
		<context name="Root">
			<mousebind button="Right" action="Press">
				<action name="ShowMenu" menu="desktop-menu" />
			</mousebind>
			<mousebind button="W-Up" action="Click">
				<action name="Execute" command="waypiedock UP" />
			</mousebind>
			<mousebind button="W-Down" action="Click">
				<action name="Execute" command="waypiedock DOWN" />
			</mousebind>
		</context>

		<context name="Frame">
			<mousebind button="W-Left" action="Press">
				<action name="Focus" />
				<action name="Raise" />
			</mousebind>
			<mousebind button="W-Left" action="Press">
				<action name="Move"/>
			</mousebind>
			<mousebind button="W-Right" action="Press">
				<action name="Resize"/>
			</mousebind>
			<mousebind button="W-Up" action="Click">
				<action name="Execute" command="waypiedock UP" />
			</mousebind>
			<mousebind button="W-Down" action="Click">
				<action name="Execute" command="waypiedock DOWN" />
			</mousebind>
		</context>
	</mouse>
```

To me it feels a bit strange to have to use `Click` for wheel steps.
Not sure what to do about that though or what would be a better alternative. Thoughts @johanmalm ?

src/cursor.c
```diff
@@ -882,6 +882,47 @@ cursor_button(struct wl_listener *listener, void *data)
...
+
+	wl_list_for_each(mousebind, &rc.mousebinds, link) {
+		if (ssd_part_contains(mousebind->context, ctx->type)
+				&& mousebind->button == button
+				&& modifiers == mousebind->modifiers) {
+			switch (mousebind->mouse_event) {
+			case MOUSE_ACTION_CLICK:
+				activated_any = true;
+				actions_run(ctx->view, server, &mousebind->actions,
+					/*resize_edges*/ 0);
```
Consolatis, line 916: If we only want to support the `Up` / `Down` + `Click` combination we could also replace the switch with a simple
```c
if (mousebind->mouse_event == MOUSE_ACTION_CLICK) {
	actions_run(..)
}
```

src/cursor.c
Consolatis: AFAIR continuous mouse wheel scrolling is internally translated in wlroots into discrete steps of 120 (or the other way round), not completely sure right now.

I think the debug log is fine for now, we can always change it later when we figure out that it unnecessarily spams the log.

src/cursor.c
Arnaudv6: >Also not a fan of the naming in general, not specific to this PR but activated_something doesn't exactly say what that variable does. Any suggestions for a more intuitive name? Maybe something like handled ?

"consumed" maybe? this way we could have "consumed_by_frame".

Arnaudv6:
> To me it feels a bit strange to have to use Click for wheel steps.

Granted this openbox jargon 'Click' does not convey the "Step" nature.
In fact it can never be broken down into Press and Release sub-events...
I do hear a _click_ when pushing that mouse wheel UP one step though :smiley:

src/cursor.c
```diff
@@ -891,15 +927,24 @@ cursor_axis(struct wl_listener *listener, void *data)
...
-	cursor_update_focus(seat->server);
+	/* Bindings swallow mouse events if activated */
+	bool triggered_axis_binding = handle_cursor_axis(server, &ctx, event);
 
-	/* Notify the client with pointer focus of the axis event. */
-	wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
-		event->orientation, event->delta, event->delta_discrete,
-		event->source);
+	if (ctx.surface && !triggered_axis_binding) {
+		/* Notify client of axis event */
```
johanmalm, line 938: This comment is in effect the same as 5 lines down. Suggest delete.

src/cursor.c
```diff
@@ -891,15 +927,24 @@ cursor_axis(struct wl_listener *listener, void *data)
...
+
+		/* Make sure we are sending the events to the surface under the cursor */
+		cursor_update_common(server, &ctx, 0, false);
```
johanmalm, line 941: Do we need `cursor_update_common()` here? 

The function `cursor_button_press()` does not have it at the equivalent place.
The `cursor_update_*()` functions took a long time to get right, so I'm a bit sensitive to adding additional calls unless we're very clear why.

Consolatis: We need that to ensure that the mouse focus shifts to the the window under the cursor. It isn't a change in this PR, it was already in there before in form of `cursor_update_focus()`.

Arnaudv6:
Thanks for review! I'm away from dev. machine for a week. Will take on when I'm back.

src/cursor.c
johanmalm: @Consolatis thanks. Sorry I didn't think through properly.
It should be `cursor_update_common(server, &ctx, event->time_msec, false);`

johanmalm: Post below updated

src/cursor.c
Consolatis: Just `bool consumed` or `bool handled` would be easier to understand I think. As this will always set `activated_any` anyway there is not really a point in differentiating between `*_by_frame` and `*`.

johanmalm: I vote for `handled` - it's consistent with `keyboard.c` that way.

https://github.com/labwc/labwc/blob/cce2fffa1dc306b5b95b00a5b3b9fc0f5b7b67c1/src/keyboard.c#L245
https://github.com/labwc/labwc/blob/cce2fffa1dc306b5b95b00a5b3b9fc0f5b7b67c1/src/keyboard.c#L122

Assume we talking both `s/activated_any/handled/` and `s/triggered_axis_binding/handled/`?

Consolatis: > Assume we talking both s/activated_any/handled/ and s/triggered_axis_binding/handled/?

Was only talking about the new function (`activated_any`) but you are right, we should likely use the same for both.
Fine with `handled`.

Arnaudv6: OK done

Arnaudv6:
I think we're ready for merge once FreeBSD check passes, if the many commits are OK for you.
void-musl check failure seems related to wl-roots build dependency not up-to-date.

Consolatis:
Thanks for fixing the remaining issues. This indeed looks about ready to go, I'll clean up the commits a bit (e.g. get rid of the merge commit and possibly combine some of them), give it a final test and then merge.

Thanks again for the whole PR :)

docs/labwc-config.5.scd
```diff
@@ -219,13 +219,16 @@ The rest of this man page describes configuration options.
 	- Left
 	- Middle
 	- Right
+	- UP (scroll up)
+	- DOWN (scroll down)
```
Consolatis, line 223: Note to self: CamelCase, will change during rebase

docs/rc.xml.all
```diff
@@ -303,6 +303,15 @@
...
 
+    <context name="Desktop">
+      <mousebind button="Up" action="Scroll">
+        <action name="GoToDesktop" to="left"/>
+      </mousebind>
+      <mousebind button="Down" action="Scroll">
+        <action name="GoToDesktop" to="right"/>
+      </mousebind>
+    </context>
+
```
Consolatis, line 314: Note to self: maybe put into existing `Desktop` context

src/cursor.c
```diff
@@ -634,8 +634,8 @@ handle_release_mousebinding(struct server *server,
 		struct cursor_context *ctx, uint32_t button)
 {
 	struct mousebind *mousebind;
-	bool activated_any = false;
-	bool activated_any_frame = false;
+	bool handled = false;
+	bool handled_in_compositor = false;
```
Consolatis, line 638: Not to self: Use own commit for `activated` -> `handled` change.

Consolatis (approved):

Arnaudv6:
> Thanks ...

Well thank you too, for kind support and review :)

Issue #597, opened 2022-10-27
1-second lag in closing XWayland menus (XWayland issue?)


jlindgren90:
Since updating xwayland from 22.1.3 to 22.1.4, there is approximately a one-second lag from the time an XWayland menu item is selected until the menu closes.  This leads to confusion as to whether the menu selection took effect or not, since there is no visual response until one second later.

With xwayland 22.1.3, menus close more-or-less instantly.

I reported this upstream as https://gitlab.freedesktop.org/xorg/xserver/-/issues/1399, but I'm not sure yet if there is something we can do on the compositor side to fix/mitigate the issue.

Consolatis:
Not sure why this relies on the destroy signal, shouldn't the menu window send an unmap signal first anyway?
After a quick look at `src/xwayland-unmanaged.c` it seems we are not actually handling the unmap signal by calling `wlr_scene_node_set_enabled(node, false)` so this very much seems like a labwc bug.

jlindgren90:
Adding `wlr_scene_node_set_enabled(unmanaged->node, false)` seems to fix it. I'll create a PR. Thanks for the suggestion!

PR #598, opened 2022-10-27
xwayland: Hide unmanaged nodes on unmap()

jlindgren90:
Since xwayland commit 9e4b7948df5a9a67632bb606150ec171b5519672 there is a one-second delay between XWayland surfaces being unmapped and destroyed.  This revealed that we were not actually handling the unmap() signal correctly for unmanaged surfaces.

Fixes: #597

Thanks @Consolatis for the suggestion.

### DIFF
--- a/src/xwayland-unmanaged.c
+++ b/src/xwayland-unmanaged.c
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
 #include "common/list.h"
 #include "common/mem.h"
 #include "labwc.h"
@@ -35,6 +36,7 @@ unmanaged_handle_map(struct wl_listener *listener, void *data)
 	struct xwayland_unmanaged *unmanaged =
 		wl_container_of(listener, unmanaged, map);
 	struct wlr_xwayland_surface *xsurface = unmanaged->xwayland_surface;
+	assert(!unmanaged->node);
 
 	/* Stack new surface on top */
 	wlr_xwayland_surface_restack(xsurface, NULL, XCB_STACK_MODE_ABOVE);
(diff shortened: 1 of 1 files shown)


src/xwayland-unmanaged.c
```diff
@@ -35,6 +36,7 @@ unmanaged_handle_map(struct wl_listener *listener, void *data)
 	struct xwayland_unmanaged *unmanaged =
 		wl_container_of(listener, unmanaged, map);
 	struct wlr_xwayland_surface *xsurface = unmanaged->xwayland_surface;
+	assert(!unmanaged->node);
```
Consolatis, line 39: I think these asserts() will allow a client application to crash the compositor by just sending two maps or unmaps in a row. Simply returning might be better instead.

We are also recreating the complete scene node based on the surface on an unmap / map cycle but even though that is not the best in case of latency I expect it not to be an issue in practice.

jlindgren90: It looks like wlroots already filters out duplicate map/unmap events; see:
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c#L836
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c#L853
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c#L913

But I can add `return` statements instead of the `assert`s if we don't want to rely on the wlroots event filtering.

Consolatis: Thanks for looking this up.

With that: Code LGTM, didn't test but feel free to merge.

jlindgren90: Thanks for the review!

PR #599, opened 2022-10-30
Support <action name="" to=""> and combine common code in the fill_{key,mouse}bind() functions

johanmalm:


Consolatis:
I wonder why we didn't do that already. Looks pretty straight forward in hindsight :)
Didn't test but code LGTM.

johanmalm:
> I wonder why we didn't do that already. Looks pretty straight forward in hindsight :) Didn't test but code LGTM.

Yes, it made me smile what I saw it - particularly when the functions were right next to each other.
Nice to simplify.

PR #603, opened 2022-11-01
checkpatch.pl

johanmalm:


scripts/check
```diff
@@ -0,0 +1,52 @@
...
+	fi
+}
+
+run_checks () {
+	if [ ! -z "$file" ]; then
+		run_checkpatch "${file}"
+		return
+	fi
+
+	for f in `find src/ -name "*.c" -type f`; do
```
Consolatis, line 27: I think there is some kind of limit on the amount of arguments like this.
So piping the result to a `while read -r file` loop might be more robust, possibly combined with having `find` use `\0` as separator and the same for the `while` loop.

Or alternatively we could use the execute option of find. However, I am not sure if those then run in parallel and if so, how many instances of those would be running concurrently.

Consolatis (commented):
Nice one.

Do we want to wrap these into a `nice` call to only use spare CPU time?

Consolatis:
Added a way to accumulate errors when checking all files. This allows to fix multiple issues at once before force pushing the next revision. Otherwise the CI would spit out an error for a file, it would get fixed, force pushed and then the CI would spit out another error for a different file.

```diff
diff --git a/scripts/check b/scripts/check
index cd6c7dd..39eb4e5 100755
--- a/scripts/check
+++ b/scripts/check
@@ -12,21 +12,25 @@ OPTIONS:
 
 run_checkpatch() {
 	if [ "$ci" = "t" ]; then
-		scripts/checkpatch.pl --terse --no-tree --file "$1" || exit 1
+		scripts/checkpatch.pl --terse --no-tree --file "$1"
 	else
 		scripts/checkpatch.pl --terse --no-tree --strict --file "$1"
 	fi
+	return $?
 }
 
 run_checks () {
 	if [ ! -z "$file" ]; then
 		run_checkpatch "${file}"
-		return
+		return $?
 	fi
 
+	errors=0
 	for f in `find src/ -name "*.c" -type f`; do
-		run_checkpatch "${f}"
+		run_checkpatch "${f}" || errors=1
 	done
+
+	return $errors
 }
 
 main () {
```

Consolatis:
- Do we also want to scan the headers in `include/` ?
- Do we want to scan for `calloc()`, `malloc()` or other "forbidden" function calls / patterns?

johanmalm:
Thanks for great suggestions. Have updated `scripts/check` with the argument-limit-fix and better error handling.

> * Do we also want to scan the headers in `include/` ?
Also done. Don't know how I missed that :smile:


>     * Do we want to scan for `calloc()`, `malloc()`, `realloc()`, `g_strcmp0` or other "forbidden" function calls / patterns?
Not sure how to best do that.
Could do a `git grep -n "malloc" src/` but unfortunately that also picks up `xmalloc`.
Maybe a python tokenizer?

>     * Do we want to scan for uninitialized static variables? (Basically turning the current check around)
Suggest we either stick with warning as is - or turn it off. I think reverse would be a step to far.

>     * Do we want to wrap checkpatch.pl calls into `nice` to only use spare CPU time?
I'm not bothered. Happy to do it though if helpful. Not sure how to, so grateful for pointer.

johanmalm:
> ```
> * Do we want to scan for `calloc()`, `malloc()`, `realloc()`, `g_strcmp0` or other "forbidden" function calls / patterns?
> ```

Suggest we do this in a different PR

> ```
> * Do we want to scan for uninitialized static variables? (Basically turning the current check around)
> ```

I've turned off `INITIALISED_STATICS`

> * Do we want to wrap checkpatch.pl calls into `nice` to only use spare CPU time?

Suggest we go without this for now


@Consolatis If you're content I'll re-base and it'll be ready.

Consolatis:
LGTM

PR #610, opened 2022-11-07
cursor: combine two variables relating to frame-context

johanmalm:
Mouse bindings, unlike key bindings, are made within contexts which represent what was clicked/dragged.  The context 'Frame' refers to the entire window frame including both the window decorations (if any) and the client window itself. It is typically used for alti + left/right click to move/resize the window.

'Frame' is a special case in that when a button is bound in this context, the action will not be forwarded to the client, which is what we describe with the 'consumed_by_frame_context' variable.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -634,8 +634,7 @@ handle_release_mousebinding(struct server *server,
 		struct cursor_context *ctx, uint32_t button)
 {
 	struct mousebind *mousebind;
-	bool handled = false;
-	bool handled_in_compositor = false;
+	bool consumed_by_frame_context = false;
 
 	uint32_t modifiers = wlr_keyboard_get_modifiers(
 			&server->seat.keyboard_group->keyboard);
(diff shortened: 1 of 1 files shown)


johanmalm:
@Consolatis would appreciate a sanity check of this

Consolatis:
LGTM and is indeed much easier to grasp when first reading the code.

The `switch()` within the loop doing `break` and `continue` is not exactly easy to grasp though (`break` corresponding to the `switch()` while `continue` corresponds to the outer loop), this might deserve another Issue / PR.

I wonder though.. don't we leak stray press / release events?
E.g. when an action is bound to `Frame` + `Press` we would leak the corresponding release event (if there is no `Frame` handler for that as well). And similar when an action is bound to `Frame` + `Release` we would leak the corresponding press event.

That means that we can't actually properly implement `Release` mousebindings for `Frame` without also leaking the press event. And we should set `mousebind->pressed_in_context` when we execute a `Frame` + `Press` mousebind which gets cleared on the release handler (+ consumes the release event).

`Click` events are handled properly though because we mark them as pressed on the `Press` handler and execute them later within the `Release` handler. `Drag` bindings look like they leak a stray release event as well because we reset `mousebind->pressed_in_context` in `process_cursor_motion()`. I didn't test any of this though, probably something for a separate issue. 

So to sum up:
- This PR looks good to me (didn't test)
- The `switch()` within the loop using `break` and `continue` is hard to grasp IMO and could deserves a refactor
- I think we are leaking some mouse events for the `Frame` context.

PR #611, opened 2022-11-09
[RFC] Implement Kill action

bi4k8:
This action forcibly disconnects clients, rather than requesting them to close. It's useful for when a client is misbehaving or frozen, but I'm not sure that it's necessarily the Right Thing. Specifically, it doesn't kill the process itself, which may be hung or unresponsive to a Wayland disconnection. Or, if the client supports "wayland robustness" (cf. David Edmundson), it might simply reconnect, which might not be the user's intention.

So, perhaps this action should instead be called "Disconnect". Perhaps I should instead (or additionally) implement a proper `Kill` action using `wl_client_get_credentials` to obtain the client pid. The [docs](https://wayland.freedesktop.org/docs/html/apc.html#Server-structwl__client_1a82a97cb3a66c1c56826a09a7b42453d9) for that function warn of potential race conditions, so I'm not sure the best (or any non-racy) way to do so, but that may not matter in practice.

Feedback would be appreciated.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -11,6 +11,10 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Close">*
 	Close top-most window.
 
+*<action name="Kill">*
+	Kill the process associated with the current window by sending it the
+	SIGTERM signal.
+
 *<action name="Execute"><command>*
 	Execute command.  Note that in the interest of backward compatibility,
 	labwc supports <execute> as an alternative to <command> even though
--- a/src/action.c
+++ b/src/action.c
@@ -37,6 +37,7 @@ enum action_type {
 	ACTION_TYPE_INVALID = 0,
 	ACTION_TYPE_NONE,
 	ACTION_TYPE_CLOSE,
+	ACTION_TYPE_KILL,
 	ACTION_TYPE_DEBUG,
 	ACTION_TYPE_EXECUTE,
 	ACTION_TYPE_EXIT,
@@ -63,6 +64,7 @@ const char *action_names[] = {
 	"INVALID",
 	"None",
 	"Close",
+	"Kill",
 	"Debug",
 	"Execute",
 	"Exit",
@@ -242,6 +244,17 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_CLOSE:
 			if (view) {
 				view_close(view);
+			}
+			break;
+		case ACTION_TYPE_KILL:
+			if (view && view->surface) {
+				/* Send SIGTERM to the process associated with the surface */
+				pid_t pid = -1;
+				struct wl_client *client = view->surface->resource->client;
+				wl_client_get_credentials(client, &pid, NULL, NULL);
+				if (pid != -1) {
+					kill(pid, SIGTERM);
+				}
 			}
 			break;
 		case ACTION_TYPE_DEBUG:


Consolatis:
While I am aware that most desktop environments have a "force kill" feature when an application is unresponsive I am not sure if the compositor is the right place for it (in comparison to some system monitor or `ps + kill`). Additional opinions from e.g. @johanmalm and @jlindgren90 regarding the feature and it being in scope for labwc or not would be welcome.

bi4k8:
When the misbehavior is a frozen window, connecting that window to a PID must necessarily be done by the compositor. If there were a simple protocol to go from a foreign toplevel handle to a PID, we could simply implement that instead of doing any process killing in the compositor, but I don't think such a thing exists.

Consolatis:
> When the misbehavior is a frozen window, connecting that window to a PID must necessarily be done by the compositor.

On the technical side I agree but usually the user is also aware of what application an unresponsive window belongs to. I am not completely against implementing this feature in labwc I just wonder if its in scope or not.

bi4k8:
> > When the misbehavior is a frozen window, connecting that window to a PID must necessarily be done by the compositor.
> 
> On the technical side I agree but usually the user is also aware of what application an unresponsive window belongs to. I am not completely against implementing this feature in labwc I just wonder if its in scope or not.

The user does know which application is involved (either subjectively or programmatically via querying the active toplevel's `app_id` with the foreign toplevel protocol), but not which instance of the application, which means practically having to play process Russian roulette.

heroin-moose:
Also trying to figure out which electron instance you need is a PITA.

jlindgren90:
My two cents:

I like the way Openbox handles unresponsive windows, by presenting a prompt after a timeout ("The window N does not seem to be responding. Do you want to force it to exit by sending the terminate signal?") Hopefully, it is a rare situation where the Kill action is needed, and when it occurs, I don't want to have to remember or look up a keybinding that I rarely use. A separate action and keybinding is also not very "discoverable" for new users.

Sending SIGTERM (possibly followed by SIGKILL after a delay) is, in my opinion, a better way to kill an unresponsive application. SIGTERM gives a well-coded application a "last chance" to save critical data before dying, and SIGKILL works for a completely deadlocked application that doesn't respond to SIGTERM. Disconnecting the Wayland connection does neither.

bi4k8:
I've changed the action to send `SIGTERM`.

Consolatis:
Maybe we could do something like this?
```
Close action
    -> send ping
    -> notify application close
    -> start timer 1

timer 1
    -> show dialog force abort
    -> if accepted by user:
        -> get pid
        -> sigterm pid
        -> create new timer 2
            -> on execute
                sigkill pid

application closed
    -> kill timer 1
    -> kill timer 2 if set

receive pong:
    -> kill timer 1
    -> kill timer 2 if set
```

bi4k8:
@Consolatis I think we could, but that's a solution for a somewhat different problem. For example, an application that responds to "Close" with a save prompt will never exit based on the compositor's Close request alone. Is a `Kill` action that applications cannot ignore decidedly out of scope?

Consolatis:
> For example, an application that responds to "Close" with a save prompt will never exit based on the compositor's Close request alone.

Yes, that would be the intended behavior with the logic above. That one should only allow users to kill actually unresponsive windows (that not even react to a wayland ping signal when being asked to close).

> Is a Kill action that applications cannot ignore decidedly out of scope?

Personally I don't really see the point for that feature in practice but that doesn't mean that it isn't in scope for labwc.

Consolatis:
@johanmalm @jlindgren90 opinions regarding the `Kill` action?

Code LGTM, just misses a docs update (`docs/labwc-actions.5.scd`)

johanmalm:
Happy to merge when docs done. I'll update scope document when merged.

I like the idea of having a timer for unresponsive apps - but suggest we leave that for another PR/day when/if we feel like it. I can't remember when I last had an unresponsive app, so it's possibly quite a low priority feature.

jlindgren90:
I don't have a problem with the addition of `Kill`. I probably wouldn't use it personally.

jlindgren90:
Looks good to me.

johanmalm:
Thanks.
Scope document updated https://github.com/labwc/labwc-scope/commit/3ecf586d84a060d1bfea629b70f334bd2428d758

PR #612, opened 2022-11-09
Support smooth scroll and horizontal scroll

bi4k8:
Note that this changes `Scroll` mousebinds from taking a "button" attribute to taking a "direction" attribute. ~Also note that while this enables trackpads to activate scroll bindings, they are currently hyper-sensitive; we should probably compare the accumulated scroll offset against some (libinput-based?) measurement.~

I've now implemented thresholding that makes smooth scrolling emit events at a reasonable speed.

### DIFF
--- a/include/config/mousebind.h
+++ b/include/config/mousebind.h
@@ -16,11 +16,22 @@ enum mouse_event {
 	MOUSE_ACTION_SCROLL,
 };
 
+enum direction {
+	LAB_DIRECTION_INVALID = 0,
+	LAB_DIRECTION_LEFT,
+	LAB_DIRECTION_RIGHT,
+	LAB_DIRECTION_UP,
+	LAB_DIRECTION_DOWN,
+};
+
 struct mousebind {
 	enum ssd_part_type context;
 
 	/* ex: BTN_LEFT, BTN_RIGHT from linux/input_event_codes.h */
 	uint32_t button;
+
+	/* scroll direction; considered instead of button for scroll events */
+	enum direction direction;
 
 	/* ex: WLR_MODIFIER_SHIFT | WLR_MODIFIER_LOGO */
 	uint32_t modifiers;
@@ -35,6 +46,7 @@ struct mousebind {
 
 enum mouse_event mousebind_event_from_str(const char *str);
 uint32_t mousebind_button_from_str(const char *str, uint32_t *modifiers);
+enum direction mousebind_direction_from_str(const char *str, uint32_t *modifiers);
 struct mousebind *mousebind_create(const char *context);
 
 #endif /* __LABWC_MOUSEBIND_H */
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -107,6 +107,9 @@ struct seat {
 	enum lab_cursors server_cursor;
 	struct wlr_cursor *cursor;
 	struct wlr_xcursor_manager *xcursor_manager;
+	struct {
+		double x, y;
+	} smooth_scroll_offset;
 
 	struct wlr_pointer_constraint_v1 *current_constraint;
 	struct wlr_idle *wlr_idle;
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -881,6 +881,38 @@ cursor_button(struct wl_listener *listener, void *data)
 	}
 }
 
+static int
+compare_delta(const struct wlr_pointer_axis_event *event, double *accum)
+{
+	/*
+	 * Smooth scroll deltas are in surface space, so treating each unit as a
+	 * scroll event would result in too-fast scrolling.
+	 *
+	 * This fudge factor (inherited from various historic projects, incl. Weston)
+	 * produces events at a more reasonable rate.
+	 *
+	 * For historic context, see:
+	 * https://lists.freedesktop.org/archives/wayland-devel/2019-April/040377.html
+	 */
+	const double SCROLL_THRESHOLD = 10.0;
+
+	if (event->delta == 0.0) {
+		/* Delta 0 marks the end of a scroll */
+		*accum = 0.0;
+	} else {
+		/* Accumulate smooth scrolling until we hit threshold */
+		*accum += event->delta;
+	}
+	if (event->delta_discrete < 0 || *accum < -SCROLL_THRESHOLD) {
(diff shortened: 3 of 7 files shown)


include/config/mousebind.h
```diff
@@ -22,6 +23,9 @@ struct mousebind {
 	/* ex: BTN_LEFT, BTN_RIGHT from linux/input_event_codes.h */
 	uint32_t button;
 
+	/* scroll direction; considered iff button is UINT32_MAX */
```
Consolatis, line 26: `s/iff/if/`

Not sure about the `UINT32_MAX` here, at the very least we should use a define (e.g. something along the lines of `LAB_BTN_SCROLL` or some better name).

include/config/mousebind.h
```diff
@@ -3,6 +3,7 @@
 #define __LABWC_MOUSEBIND_H
 
 #include <wayland-util.h>
+#include <wlr/types/wlr_output_layout.h>
```
Consolatis, line 6: I think it makes sense to create our own enum in this case which could then also contain an `LAB_DIRECTION_INVALID` as index 0.

src/cursor.c
```diff
@@ -891,22 +891,34 @@ handle_cursor_axis(struct server *server, struct cursor_context *ctx,
 	uint32_t modifiers = wlr_keyboard_get_modifiers(
 			&server->seat.keyboard_group->keyboard);
 
-	uint32_t button = 0;
-	if (event->delta_discrete < 0) {
-		button = BTN_GEAR_UP;
-	} else if (event->delta_discrete > 0) {
-		button = BTN_GEAR_DOWN;
+	bool found_direction = false;
```
Consolatis, line 894: If we use our own enum here the bool isn't required anymore and we could just initialize the enum with `LAB_DIRECTION_INVALID` + check for it below.

Consolatis:
Thanks for working on this.

> Also note that while this enables trackpads to activate scroll bindings, they are currently hyper-sensitive; we should probably compare the accumulated scroll offset against some (libinput-based?) measurement.

As written in https://github.com/labwc/labwc/pull/596#discussion_r1008395295 I think we have to restrict the actions to only trigger on steps of 120 but I am not sure on the relevant struct member from the event for this right now.

bi4k8:
Updated with the suggested `enum` design, which is significantly cleaner.

> Thanks for working on this.
> 
> > Also note that while this enables trackpads to activate scroll bindings, they are currently hyper-sensitive; we should probably compare the accumulated scroll offset against some (libinput-based?) measurement.
> 
> As written in [#596 (comment)](https://github.com/labwc/labwc/pull/596#discussion_r1008395295) I think we have to restrict the actions to only trigger on steps of 120 but I am not sure on the relevant struct member from the event for this right now.

I don't see any logic relevant to this in wlroots ([search results](https://gitlab.freedesktop.org/search?search=delta_discrete&project_id=12103&group_id=9759&search_code=true&repository_ref=master))--I think we may have to accumulate continuous events ourselves.

gdk has code for such accumulation [here](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3839/diffs?commit_id=5a1fcad79eb18ea5577486d92da018abd5f07788) which makes it look like sane behavior might just be a matter of checking the accumulator's absolute value against 1. The hardest part of such a patch is figuring out a pointer-specific place to store the accumulator and when to reset it.

bi4k8:
Looking at [wlroots code](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/eec95e3d5e1a4f2e13b1f6b34cc287475ca57daf/types/seat/wlr_seat_pointer.c#L375-L377), it seems we get an axis event with `delta == 0.0` when a scroll ends, so we can simply accumulate until then.

include/labwc.h
```diff
@@ -107,6 +107,9 @@ struct seat {
 	enum lab_cursors server_cursor;
 	struct wlr_cursor *cursor;
 	struct wlr_xcursor_manager *xcursor_manager;
+	struct {
+		double x, y;
```
Consolatis, line 111: Do we really need both here? Or, in other words, could we receive / should we handle two scrolling actions on different axis at once?

src/cursor.c
```diff
@@ -891,22 +913,32 @@ handle_cursor_axis(struct server *server, struct cursor_context *ctx,
...
 		return false;
 	}
 
 	wl_list_for_each(mousebind, &rc.mousebinds, link) {
 		if (ssd_part_contains(mousebind->context, ctx->type)
-				&& mousebind->button == button
+				&& mousebind->direction == direction
 				&& modifiers == mousebind->modifiers
 				&& mousebind->mouse_event == MOUSE_ACTION_SCROLL) {
 			handled = true;
```
Consolatis, line 944: Do we want to reset the accumulator after executing a binding?

src/cursor.c
```diff
@@ -881,6 +881,28 @@ cursor_button(struct wl_listener *listener, void *data)
 	}
 }
 
+static int
+compare_delta(const struct wlr_pointer_axis_event *event, double *accum)
+{
+	const double SCROLL_THRESHOLD = 10.0;
```
Consolatis, line 887: A comment explaining this value would be good, e.g. where does it come from / why exactly `10.0`.

include/labwc.h
bi4k8: Yes; when a user scrolls "diagonally" both axes have nonzero deltas interleaved.

src/cursor.c
bi4k8: I don't think we do--when the user stops scrolling, the accumulator is reset, and if the user continues scrolling, we don't want to drop the additional partial progress they may have made towards the next quantized activation.

src/cursor.c
bi4k8: Sadly this value was invented whole cloth on the basis of what "feels right". I'll add a comment to that effect.

I have a vague memory of there being a similar 10 hard-coded somewhere in the Gdk/Gtk stack, but wasn't able to find it on a cursory search. I'll look again when I get a chance.

bi4k8: I've tracked down a place explaining the history of the 10 constant (which is in Mutter, libinput, and Weston, not GTK) and added a comment explaining its origin.

johanmalm:
LGTM

PR #613, opened 2022-11-09
Use 1000/rate msecs for keybind repeat event source delay

bi4k8:
This makes the rate of labwc keybind repeats match that of regular key repeats.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -235,6 +235,7 @@ handle_keybind_repeat(void *data)
 {
 	struct keyboard *keyboard = data;
 	assert(keyboard->keybind_repeat);
+	assert(keyboard->keybind_repeat_rate > 0);
 
 	/* synthesize event */
 	struct wlr_keyboard_key_event event = {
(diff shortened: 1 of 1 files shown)


src/keyboard.c
```diff
@@ -243,8 +243,9 @@ handle_keybind_repeat(void *data)
 	};
 
 	handle_compositor_keybindings(keyboard, &event);
+	int next_repeat_ms = 1000 / keyboard->keybind_repeat_rate;
```
Consolatis, line 246: Is it possible that `keyboard->keybind_repeat_rate` is set to `0`? If yes this would be a division by zero.

jlindgren90:
Huh. I misunderstood the units of the "rate" value evidently. Good catch.

src/keyboard.c
jlindgren90: It shouldn't be zero here because we checked it in `start_keybind_repeat()`. Wouldn't hurt to add another `assert()` at the start of this function though.

bi4k8: Assert added.

Consolatis:
Thanks

PR #614, opened 2022-11-09
Correctly update kbd repeat rate by filtering input devices

bi4k8:
Without this patch, reconfiguring keyboard repeat rate/delay at runtime does not work. However, I've left this as a draft because I don't understand the role of the `keyboard_group` here. Is `seat->keyboard_group->keyboard` guaranteed to be present in the `seat->inputs` list, or should it be handled separately (as commented-out)?

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -418,16 +418,17 @@ seat_reconfigure(struct server *server)
 {
 	struct seat *seat = &server->seat;
 	struct input *input;
-	struct wlr_keyboard *kb = &seat->keyboard_group->keyboard;
 	wl_list_for_each(input, &seat->inputs, link) {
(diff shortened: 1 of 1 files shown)


Consolatis:
I am not that familiar with this part of the code. I think @johanmalm might have some further input for this but AFAIR he is currently a bit busy so it might take a few days for reply / review.

johanmalm:
> Is `seat->keyboard_group->keyboard` guaranteed to be present in the `seat->inputs` list, or should it be handled separately (as commented-out)?

I've not studied `wlr_keyboard_group` in detail, but am pretty sure that `seat->inputs` will contain all keyboards.
If you remove the commented out lines, I'm happy to merge.

PR #616, opened 2022-11-09
Distinguish no and unknown font places

bi4k8:
Currently, the `rc.xml` parser applies font settings in a `<font>` tag with an unknown value for its `place` attribute to all fonts. This means that whatever the final unknown-`place` `<font>` tag is in a user's `rc.xml` applies to all text drawn by labwc.

Instead, only treat `<font>` tags with an empty or missing `place` attribute as applying globally, and warn when encountering unknown `place` attribute values (which will help us find font places to support).

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -35,7 +35,8 @@ static struct action *current_keybind_action;
 static struct action *current_mousebind_action;
 
 enum font_place {
-	FONT_PLACE_UNKNOWN = 0,
+	FONT_PLACE_NONE = 0,
+	FONT_PLACE_UNKNOWN,
 	FONT_PLACE_ACTIVEWINDOW,
 	FONT_PLACE_MENUITEM,
 	FONT_PLACE_OSD,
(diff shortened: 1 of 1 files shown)


Consolatis:
Didn't test but this looks good to me.
Personally I'd prefer the commit message to start with the file, e.g.
`src/config/rc.xml: Distinguish no and unknown font places`.

Plus the body of the commit message should contain the text written above because it sums up the point of the change pretty nicely.

On a related note: We likely don't want to overwrite already set font settings with the global one if it comes later in the config.
So maybe we should check for it being set to something != default first before overwriting it with an empty place attribute.
This doesn't have to be done in this PR though (and if it is with a separate commit). An example for this would be `rc.workspace_config.popuptime` which gets set to `INT_MIN` in `rcxml_init()` and then corrected in `post_processing()` if still set to it after parsing. We should use a define like `LAB_CONF_UNSPEC_INT` or similar for this though.

jlindgren90:
LGTM. The existing behavior was confusing to me too.

bi4k8:
Commit message updated. I agree we should further improve the semantics to be a proper fallback/default. Are there docs describing the intended behavior?

Consolatis:
Thanks.

> Are there docs describing the intended behavior?

There is this statement in `docs/labwc-config.5.scd`
> If no place attribute is provided, the setting will be applied to all places.

So if we change that to only overwrite default settings this part should be updated as well.
Maybe we could have an additional static var like `struct font *default_font` which gets initialized in `rcxml_init()` and updated on empty `place` attribute.
Then in `post_processing()` we check for `LAB_CONF_UNSPEC_*` for the various font node arguments and if they weren't set specifically to something else we then set their arguments to the same as in the `default_font` struct.

That way we will never overwrite specifically set font attributes and instead only use it to overwrite the defaults.

PR #618, opened 2022-11-11
xwayland: Better document/assert view/surface association

jlindgren90:
Each XWayland view is paired with a particular `wlr_xwayland_surface` and its lifetime is tied to that surface.  This condition in `handle_map()`:

    if (xsurface != view->xwayland_surface)

could never be true since the view is only registered to receive the "map" signal from `view->xwayland_surface`, and no other.  So the code updating `view->xwayland_surface` in `handle_map()` was dead.

So let's clean things up a little:

- Remove the dead code
- Add some comments, and slightly rearrange code to match
- Add/update `assert()`s in signal handlers for consistency
- Pass `xsurface` as `data` when calling `handle_unmap()` and `handle_destroy()` explicitly, to be consistent

Lightly tested; I didn't see any new crashers or `assert()` failures.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -10,7 +10,7 @@ static void
 handle_commit(struct wl_listener *listener, void *data)
 {
 	struct view *view = wl_container_of(listener, view, commit);
-	assert(view->surface);
+	assert(data && data == view->surface);
 
 	/* Must receive commit signal before accessing surface->current* */
 	struct wlr_surface_state *state = &view->surface->current;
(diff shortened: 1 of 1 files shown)


Consolatis:
This was added in
- #526

AFAIR the issue was that some X11 application like VLC and Discord would create completely new surfaces after being minimized to tray and then restored. The same happened randomly for Gimp color picker windows. On a short look I agree that that should only matter for `view->surface` / `view->xwayland_surface->surface` and the check in `handle_map()` doesn't really make sense.

Will look trough the diff later in more detail.

Consolatis:
Had a quick test with minimizing VLC to tray and then opening it again, no issue detected.
I didn't verify the whole logic here but feel free to merge if you are fine with the PR.

johanmalm:
Sorry for being a bit slow to respond. Let me try this with the gimp color picker as a regression test.

johanmalm:
It works fine with gimp, although I can't actually reproduce the black window even with older commits now, so can't vouch for it as a regression test pass.

Anyway, it looks like an improvement in code correctness, so I'll go for the merge.

Looking back, I suspect it was https://github.com/labwc/labwc/commit/a0afd443f59b6b9cab1563d3b963e9ecb7af98e4 that fixed the gimp-color-picker issue last time rather than https://github.com/labwc/labwc/commit/01e832b4a88a825700bc8956e2b67dedad100288

PR #622, opened 2022-11-15
Release layer tree when releasing output

yuanye100:
When the output is connected, the layer shell related trees will be created, but when the output exits, these trees will not be released. Add this code to release the layer shell related tree created when the output is accessed before the output exits.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -43,6 +43,19 @@ output_destroy_notify(struct wl_listener *listener, void *data)
 	wl_list_remove(&output->frame.link);
 	wl_list_remove(&output->destroy.link);
 
+	struct lab_layer_surface *layer, *tmp_layer;
+	int nr_layers = sizeof(output->layers) / sizeof(output->layers[0]);
+	for (int i = 0;i < nr_layers;i++) {
+		wl_list_for_each_safe (layer, tmp_layer, &output->layers[i], link) {
+			wlr_layer_surface_v1_destroy(layer->scene_layer_surface->layer_surface);
+		}
+	}
+
+	for (int i = 0; i < nr_layers; i++) {
+		wlr_scene_node_destroy(&output->layer_tree[i]->node);
+	}
+	wlr_scene_node_destroy(&output->layer_popup_tree->node);
+
 	struct view *view;
 	struct server *server = output->server;
 	wl_list_for_each(view, &server->views, link) {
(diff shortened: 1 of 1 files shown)


src/output.c
```diff
@@ -43,6 +43,19 @@ output_destroy_notify(struct wl_listener *listener, void *data)
 	wl_list_remove(&output->frame.link);
 	wl_list_remove(&output->destroy.link);
 
+	struct lab_layer_surface *layer, *tmp_layer;
+	int nr_layers = sizeof(output->layers) / sizeof(output->layers[0]);
+	for (int i = 0;i < nr_layers;i++) {
+		wl_list_for_each_safe (layer, tmp_layer, &output->layers[i], link) {
+			wlr_layer_surface_v1_destroy(layer->scene_layer_surface->layer_surface);
+		}
+	}
```
Consolatis, line 52: Thanks for your PR.

I am not sure we need to destroy the layer surfaces here, they might be destroyed when their scene node gets destroyed. Need to verify this though. At the very least we should also remove the items from the list.

src/output.c
```diff
@@ -43,6 +43,19 @@ output_destroy_notify(struct wl_listener *listener, void *data)
...
+
+	for (int i = 0; i < nr_layers; i++) {
+		wlr_scene_node_destroy(&output->layer_tree[i]->node);
+	}
+	wlr_scene_node_destroy(&output->layer_popup_tree->node);
```
Consolatis, line 57: This looks sensible. We may also want to destroy the OSD tree node here.

src/output.c
```diff
@@ -230,7 +243,7 @@ output_update_for_layout_change(struct server *server)
 
 	/*
 	 * "Move" each wlr_output_cursor (in per-output coordinates) to
-	 * align with the seat cursor.  Set a default cursor image so
+	 * align with the seat cursor.	Set a default cursor image so
```
Consolatis, line 246: This seems to replace a double space with a tab, should be replaced by a single space instead.

johanmalm:
Thanks for working on this. I've converted to draft to reflect that it's work in progress (and to eliminate the risk of me merging it).
Shout if you want help thinking any of this through or discussing it :smile:

src/output.c
Consolatis: They are destroyed in layers.c when the corresponding output is being destroyed.

src/output.c
Consolatis: Added the the OSD tree node

src/output.c
Consolatis: Fixed by replacing with a single space.

Consolatis:
Added a slightly modified version of this commit in PR #700.

Thanks again.

PR #625, opened 2022-11-16
output,xwayland: Add support for _NET_WM_STRUT_PARTIAL

jlindgren90:
Account for space taken up by XWayland panels (as indicated by the
_NET_WM_STRUT_PARTIAL property) in the `usable_area` calculation.

This makes it possible to use labwc in a "transitional" setup, where it
replaces the X11 window manager and compositor, but most other parts of
a existing X11 desktop environment can still be used via XWayland.

(Some remaining drawbacks of such a setup would be the lack of desktop
icons, and native Wayland clients not showing up in X11-based taskbars.)

This is a draft because:
- it depends on a change to wlroots, something like https://github.com/jlindgren90/wlroots/commit/b8055330159b86d5e88d01596a16fc1a44361c97.
- I'm looking for feedback on whether this is useful to anyone else besides me.

The first commit, however, could be merged on its own (without any wlroots changes).

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -112,6 +112,8 @@ struct view_impl {
 	struct view_size_hints (*get_size_hints)(struct view *self);
 	/* if not implemented, VIEW_WANTS_FOCUS_ALWAYS is assumed */
 	enum view_wants_focus (*wants_focus)(struct view *self);
+	/* returns true if view reserves space at screen edge */
+	bool (*has_strut_partial)(struct view *self);
 };
 
 struct view {
@@ -427,6 +429,14 @@ void view_append_children(struct view *view, struct wl_array *children);
  * views/surfaces.
  */
 bool view_is_related(struct view *view, struct wlr_surface *surface);
+
+/**
+ * view_has_strut_partial() - returns true for views that reserve space
+ * at a screen edge (e.g. panels). These views are treated as if they
+ * have the fixedPosition window rule: i.e. they are not restricted to
+ * the usable area and cannot be moved/resized interactively.
+ */
+bool view_has_strut_partial(struct view *view);
 
 const char *view_get_string_prop(struct view *view, const char *prop);
 void view_update_title(struct view *view);
--- a/src/output.c
+++ b/src/output.c
@@ -603,6 +603,16 @@ update_usable_area(struct output *output)
 	struct wlr_box old = output->usable_area;
 	layers_arrange(output);
 
+#if HAVE_XWAYLAND
+	struct view *view;
+	wl_list_for_each(view, &output->server->views, link) {
+		if (view->mapped && view->type == LAB_XWAYLAND_VIEW) {
+			xwayland_adjust_usable_area(view,
+				output->server->output_layout,
+				output->wlr_output, &output->usable_area);
+		}
+	}
+#endif
 	return !wlr_box_equal(&old, &output->usable_area);
 }
 
--- a/src/view.c
+++ b/src/view.c
@@ -593,7 +593,9 @@ view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
 		return false;
 	}
 
-	if (window_rules_get_property(view, "fixedPosition") == LAB_PROP_TRUE) {
+	/* Avoid moving panels out of their own reserved area ("strut") */
+	if (window_rules_get_property(view, "fixedPosition") == LAB_PROP_TRUE
+			|| view_has_strut_partial(view)) {
 		return false;
 	}
 
@@ -1541,6 +1543,14 @@ view_is_related(struct view *view, struct wlr_surface *surface)
 		return view->impl->is_related(view, surface);
 	}
 	return false;
+}
+
+bool
+view_has_strut_partial(struct view *view)
+{
+	assert(view);
+	return view->impl->has_strut_partial &&
+		view->impl->has_strut_partial(view);
 }
 
 const char *
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -6,6 +6,8 @@
 #include "view.h"
 
 struct wlr_compositor;
+struct wlr_output;
+struct wlr_output_layout;
 
 struct xwayland_unmanaged {
 	struct server *server;
@@ -37,6 +39,7 @@ struct xwayland_view {
 	struct wl_listener set_class;
 	struct wl_listener set_decorations;
 	struct wl_listener set_override_redirect;
+	struct wl_listener set_strut_partial;
 
 	/* Not (yet) implemented */
 /*	struct wl_listener set_role; */
@@ -57,6 +60,10 @@ struct wlr_xwayland_surface *xwayland_surface_from_view(struct view *view);
 void xwayland_server_init(struct server *server,
 	struct wlr_compositor *compositor);
 void xwayland_server_finish(struct server *server);
+
+void xwayland_adjust_usable_area(struct view *view,
+	struct wlr_output_layout *layout, struct wlr_output *output,
+	struct wlr_box *usable);
 
 void xwayland_update_workarea(struct server *server);
 
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -35,7 +35,9 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 		return;
 	}
 
-	if (window_rules_get_property(view, "fixedPosition") == LAB_PROP_TRUE) {
+	/* Prevent moving/resizing fixed-position and panel-like views */
+	if (window_rules_get_property(view, "fixedPosition") == LAB_PROP_TRUE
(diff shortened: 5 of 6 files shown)


Consolatis:
> I'm looking for feedback on whether this is useful to anyone else besides me.

For me personally, I don't have much use for this feature (using my own wayland native panel) but I am all for better interoperability between X11 and Wayland. Maybe we could somehow tunnel the xdg toplevel information to xwayland so they would show up on an X11 panel as well?

Is there a PR for wlroots to implement that feature?

jlindgren90:
> Is there a PR for wlroots to implement https://github.com/jlindgren90/wlroots/commit/b8055330159b86d5e88d01596a16fc1a44361c97?

I plan to create one shortly, just wanted to get initial feedback here first.

jlindgren90:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3866

johanmalm:
I'm fine with supporting this.
I've not reviewed the code in detail, but looks okay in principle.
Just a thought (and it might be misguided and an ill through one) could we check the `_NET_WM_STRUT_PARTIAL` property on map? Possible inspiration here: https://github.com/swaywm/sway/blob/2dace6b8242189332c68d47467f59feb54e6a56a/sway/desktop/xwayland.c#L305
There might be good reasons for adding event/listener - for example if you want to drag the panel around - but just thought I'd throw it in.

jlindgren90:
> Just a thought (and it might be misguided and an ill through one) could we check the `_NET_WM_STRUT_PARTIAL` property on map?

I'm not sure exactly what you had in mind, but I do have this check at the end of `map()` in `xwayland.c`:
```
	/* Update usable area to account for XWayland "struts" (panels) */
	if (view->xwayland_surface->strut_partial) {
		output_update_all_usable_areas(view->server, false);
	}
```

jlindgren90:
> There might be good reasons for adding event/listener - for example if you want to drag the panel around

Yes, we need the event listener for when the panel moves around to different screen edges, or (with xfce4-panel for example) when someone toggles "Span Multiple Outputs".

johanmalm:
> I'm not sure exactly what you had in mind, but I do have this check at the end of `map()` in `xwayland.c`:
> 

I was just thinking of ways to deal with it in labwc if wlroots won’t accept the patch.

jlindgren90:
> I was just thinking of ways to deal with it in labwc if wlroots won’t accept the patch.

I see, makes sense. It looks a bit difficult to implement this without support in wlroots though, for a couple of reasons (e.g. we need access to the `xcb_screen_t` to get the X11 screen size) -- I'm not sure if I would pursue it further in that case.

jlindgren90:
Rebased and reorganized a bit.

src/output.c
```diff
@@ -434,6 +430,51 @@ output_from_wlr_output(struct server *server, struct wlr_output *wlr_output)
...
+	struct view *view;
+	wl_list_for_each(view, &output->server->views, link) {
+		if (view->mapped && view->type == LAB_XWAYLAND_VIEW) {
+			view_subtract_strut_from_usable_area(
+				view, output->server->output_layout,
+				output->wlr_output, &output->usable_area);
+		}
+	}
+#endif
+	return old.x != output->usable_area.x
```
Consolatis, line 450: `return !wlr_box_equal(&old, &output->usable_area)`

src/server.c
```diff
@@ -100,10 +100,7 @@ seat_disinhibit_input(struct seat *seat)
 	 * Triggers a refocus of the topmost surface layer if necessary
 	 * TODO: Make layer surface focus per-output based on cursor position
 	 */
-	struct output *output;
-	wl_list_for_each(output, &seat->server->outputs, link) {
-		layers_arrange(output);
-	}
+	output_update_all_usable_areas(seat->server, false);
```
Consolatis, line 103: Inline comment for the `bool` would be nice here

Consolatis:
I really like the refactoring here regarding only output.c being responsible for the layout / usable_area change.
That would also clean up the region PR which added yet another call in `layers_arrange()` and calls `desktop_arrange_all_views()` more often than necessary.

Would you mind splitting that refactoring from this PR and into a new one?
Could also do it myself if you are busy with something else currently.

jlindgren90:
> I really like the refactoring regarding only output.c being responsible for the layout / usable_area change. That would also clean up the region PR which added yet another call in `layers_arrange()` and calls `desktop_arrange_all_views()` more often than necessary.
> 
> Would you mind splitting that refactoring from this PR into a new one? Could also do it myself if you are busy with something else currently.

Busy with full-time work and taking care of a baby currently :) Please go ahead and pull the refactoring in yourself, if you would like.

Consolatis:
> Busy with full-time work and taking care of a baby currently :)

Priorities done right :)

> Please go ahead and pull the refactoring in yourself, if you would like.

Did that: #713
I amended your commit, I hope thats fine by you.

jlindgren90:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3866 has been merged into wlroots master so this could be rebased and ready for review once we move up to wlroots 0.17.x (see also: #626)

Consolatis:
Awesome!

I am currently tinkering with some wlr-foreign-toplevel-management to X11 bridge (in form of an external application for now). Theoretically we should be able to have both states mostly synced at some point. When I figure out how the X11 lookup works in practice (it seems its just root window -> `_NET_CLIENT_LIST` -> window lookup -> `WM_NAME`) + how the control of those windows from a panel works we could potentially even think about the possibility to include a bridge like that in labwc itself.

Consolatis:
Small status report regarding mapping xdg toplevel views to (unmapped) X11 window:

The idea was to create a new X11 window for every wayland native window and keep its name and app_id in sync.
Further, requests to activate / minimize / maximize / close that window should be relayed back to the original wayland native window.

Using an external application (wlr-foreign-toplevel + xcb) works but is a bad idea: we don't know if a new toplevel we receive is in fact a xwayland window and thus end up creating unmapped X11 windows for existing X11 windows.

Thus I tinkered a bit with wlroots / labwc directly and can now use tint2 to list and control (activate, minimize, maximize, close) wayland windows. So this doesn't just work in theory but also in practice. I am currently trying to figure out what part of the work must be done by the wlroots X11 window manager (`wlroots/xwayland/xwm.c`) and what can be split off into the compositor.

The reason for also keeping the app_id in sync for those fake X11 windows is so that an external application can react to those fake windows, read the app_id, figure out the appropriate icon (e.g. via `.desktop` files or other means) and then set the correct icon on the fake X11 window. A taskbar should then pick that up and have something to show other than a generic "not-found" icon. I didn't yet start on that part of the work but it should be relatively trivial, I already have the logic for the icon lookup available (its used in my own panel) and using xlib / xcb to react to new X11 windows + fetch the app_id + set properties was something I already implemented when testing the original external application idea.

jlindgren90:
That's very impressive @Consolatis !

jlindgren90:
This will need a small update if https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4392 is merged.

jlindgren90:
Rearranged code a little, no functional change.

jlindgren90:
This might need some additional work to account for #1225 (to ensure we don't move XWayland panels out of their own exclusion area).

jlindgren90:
This one should be ready to go now. I added a second commit to treat XWayland panels the same as if the `fixedPosition` rule is set for them.

src/view.c
```diff
@@ -1543,6 +1545,13 @@ view_is_related(struct view *view, struct wlr_surface *surface)
 	return false;
 }
 
+bool view_has_panel_strut(struct view *view)
+{
```
Consolatis, line 1549: nitpick: `bool` in its own line

include/view.h
```diff
@@ -68,6 +68,8 @@ enum view_wants_focus {
 };
 
 struct view;
+struct wlr_output;
+struct wlr_output_layout;
```
Consolatis, line 72: Leftovers from a rebase or similar?
I didn't spot any `wlr_output` nor `wlr_output_layout` in `view.h` at least.

jlindgren90: Hmm yes, these were probably for `xwayland_adjust_usable_area` which is in xwayland.h now.

Consolatis:
LGTM but didn't test.

johanmalm:
LGTM

Just one suggestion in terms of naming:

`s/has_panel_strut/has_net_wm_strut_partial/`
`s/view_has_panel_strut/view_has_net_wm_strut_partial/`

...because it might not be a panel and `_NET_WM_STRUT_PARTIAL` is fairly well-known so the reader has a better chance of making the connection.

Happy to merge without this change - just a thought.

jlindgren90:
How about `has_strut_partial`?

johanmalm:
Great stuff.

PR #627, opened 2022-11-16
Fix positioning of initially-maximized XWayland views

jlindgren90:
`map()` in `xwayland.c` called `ssd_create()` but did not call `view_apply_maximized_geometry()` afterward, resulting in the decorations being displayed off-screen.

Rather than calling `view_apply_maximized_geometry()` in more places, let's reuse the existing call in `view_set_decorations()`, and extend `ssd_update_geometry()` to call `ssd_create()` when needed.

I also included a second commit which simply deduplicates some code, by having `view_toggle_decorations()` call `view_set_decorations()`.

### DIFF
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -169,16 +169,18 @@ ssd_create(struct view *view)
 void
 ssd_update_geometry(struct view *view)
 {
-	if (!view->ssd.tree || !view->scene_node) {
+	if (!view->scene_node) {
 		return;
 	}
 
 	if (!view->ssd.enabled) {
-		if (view->ssd.tree->node.enabled) {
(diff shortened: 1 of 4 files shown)


Consolatis:
Code LGTM but didn't test. Feel free to merge if you tested the changes.

johanmalm:
I can't can't find a way to test this, nor reproduce the issue.

I've tried with xterm and `GDK_BACKEND=x11 mousepad` etc. What application do you use to push the ssd off-screen?

jlindgren90:
I was seeing the issue with claws-mail (GTK2 version)

johanmalm:
I can re-produce with claws-mail.
Big thanks for thinking through the fix - it does indeed tidy things up :smile:

Issue #633, opened 2022-11-19
Natural geometry not saved correctly after SnapToEdge+Maximize


jlindgren90:
Steps to reproduce:
- Start with a normal (untiled, unmaximized) window in a given position
- Drag the window via titlebar to a screen edge until it snaps
- Drag the window via titlebar to a screen top until it maximizes
- Unmaximize the window

Expected results:
The window should return to the original position.

Actual results:
The window ends up in the top-left corner of the screen.

This has to do with the (somewhat sketchy) way `view->natural_geometry` is manipulated in `interactive_begin()/interactive_end()`. I have some ideas for a fix, and will submit a merge request when it's ready.

Consolatis:
Hm.. That UX pattern is a bit strange because it differs from using doubleclick / a keybind for maximize when already in tiled mode. What should happen is that un-maximizing the window returns to the tiled state and if then moving the tiled window it should return to the original state (before tiling).

I am not sure if it makes sense to check for that in interactive begin to have the same behavior when dragging the window to maximize and then in interactive_end reset the tiling state. That adds yet more complexity but would be more coherent with the doubleclick / keybind method of maximizing.

jlindgren90:
> That adds yet more complexity but would be more coherent with the doubleclick / keybind method of maximizing.

Hmm, yes, that would be more consistent. Currently `interactive_begin()` just clears the tiled state regardless. Maybe we could clear it in `interactive_end()` instead.

I think the changes in #634 would still be applicable whatever we do with the tiled state.

jlindgren90:
I updated #634 so that un-maximizing the window returns to the tiled state.

PR #634, opened 2022-11-19
interactive: Clean up natural_geometry hacks

jlindgren90:
There are a couple of sketchy things going on here:

- interactive_begin() pokes its own values into view->natural_geometry to force view_maximize() to set a particular geometry.
- interactive_end() "fixes" view->natural_geometry after calling view_maximize() to save the original geometry from the start of the interactive move/resize.

Let's adjust/expand the API of view.c so that the interactive.c can avoid this "back door" of overwriting view->natural_geometry directly.

This also addresses an issue where the natural_geometry was not saved correctly when dragging and snapping a window to a screen edge.

Fixes: #633

### DIFF
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -17,123 +17,153 @@ max_move_scale(double pos_cursor, double pos_current,
 void
 interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 {
-	if (mode == LAB_INPUT_STATE_MOVE && view->fullscreen) {
-		/**
-		 * We don't allow moving fullscreen windows.
-		 *
-		 * If you think there is a good reason to allow it
-		 * feel free to open an issue explaining your use-case.
-		 */
-		return;
-	}
-	if (mode == LAB_INPUT_STATE_RESIZE
-			&& (view->fullscreen || view->maximized)) {
-		/* We don't allow resizing while in maximized or fullscreen state */
-		return;
-	}
-
 	/*
 	 * This function sets up an interactive move or resize operation, where
 	 * the compositor stops propagating pointer events to clients and
 	 * instead consumes them itself, to move or resize windows.
 	 */
-	struct seat *seat = &view->server->seat;
 	struct server *server = view->server;
-	server->grabbed_view = view;
-	server->input_mode = mode;
-
-	/* Remember view and cursor positions at start of move/resize */
-	server->grab_x = seat->cursor->x;
-	server->grab_y = seat->cursor->y;
-	server->grab_box.x = view->x;
-	server->grab_box.y = view->y;
-	server->grab_box.width = view->w;
-	server->grab_box.height = view->h;
-	server->resize_edges = edges;
-
-	if (view->maximized || view->tiled) {
-		if (mode == LAB_INPUT_STATE_MOVE) {
-			/* Exit maximized or tiled mode */
-			int new_x = max_move_scale(view->server->seat.cursor->x,
-				view->x, view->w, view->natural_geometry.width);
-			int new_y = max_move_scale(view->server->seat.cursor->y,
-				view->y, view->h, view->natural_geometry.height);
-			view->natural_geometry.x = new_x;
-			view->natural_geometry.y = new_y;
-			if (view->maximized) {
-				view_maximize(view, false);
-			}
-			if (view->tiled) {
-				view_move_resize(view, view->natural_geometry);
-			}
-			/**
-			 * view_maximize() / view_move_resize() indirectly calls
-			 * view->impl->configure which is async but we are using
-			 * the current values in server->grab_box. We pretend the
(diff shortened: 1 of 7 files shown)


src/interactive.c
```diff
@@ -112,26 +106,22 @@ interactive_end(struct view *view)
 			wlr_output_layout_output_coords(view->server->output_layout,
 				view->output->wlr_output, &cursor_x, &cursor_y);
 
+			/*
+			 * Don't store natural geometry here (it was
+			 * stored already in interative_begin())
```
Consolatis, line 111: s/interative/interactive/

src/interactive.c
```diff
@@ -112,26 +106,22 @@ interactive_end(struct view *view)
...
+			 */
 			if (cursor_x <= area->x + snap_range) {
-				view_snap_to_edge(view, "left");
+				view_snap_to_edge(view, "left", false);
 			} else if (cursor_x >= area->x + area->width - snap_range) {
-				view_snap_to_edge(view, "right");
+				view_snap_to_edge(view, "right", false);
```
Consolatis, line 116: Inline comment about the new `false` would nice in this function as well, same for `view_maximize()` further down.

src/interactive.c
```diff
@@ -54,29 +54,23 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
...
 			 */
-			server->grab_box.x = new_x;
-			server->grab_box.y = new_y;
-			server->grab_box.width = view->natural_geometry.width;
-			server->grab_box.height = view->natural_geometry.height;
+			server->grab_box = geometry;
 		}
+	} else {
+		/* Store natural geometry at start of move/resize */
+		view_store_natural_geometry(view);
```
Consolatis, line 73: I think we only have to store the natural geometry when `mode == LAB_INPUT_STATE_MOVE`. Maybe we could switch the `mode == MOVE` check and the `view->maximized || view->tiled check` here so the `mode == MOVE` check is the outer one and the else branch here is for the inner check (`!maximized && !tiled`)

src/view.c
```diff
@@ -423,44 +423,64 @@ view_apply_unmaximized_geometry(struct view *view)
...
-	}
-	if (view->toplevel_handle) {
-		wlr_foreign_toplevel_handle_v1_set_maximized(
-			view->toplevel_handle, maximize);
-	}
+	set_maximized(view, maximize);
 	if (maximize) {
-		interactive_end(view);
-		if (!view->tiled) {
+		interactive_end(view, /*allow_snap*/ false);
```
Consolatis, line 464: I don't remember why we have to call `interactive_end()` here in the first place, if you do would you mind adding a comment here?

src/view.c
```diff
@@ -534,7 +554,7 @@ view_set_fullscreen(struct view *view, bool fullscreen,
 		wlr_output = view_wlr_output(view);
 	}
 	if (fullscreen) {
-		interactive_end(view);
+		interactive_end(view, /*allow_snap*/ false);
```
Consolatis, line 557: Same here

src/view.c
```diff
@@ -762,8 +783,8 @@ view_snap_to_edge(struct view *view, const char *direction)
 
 	if (view->maximized) {
 		/* Unmaximize + keep using existing natural_geometry */
-		view_maximize(view, false);
-	} else if (!view->tiled) {
+		view_maximize(view, false, false);
```
Consolatis, line 786: inline comment would be nice

jlindgren90:
v2:

- Save the natural geometry and the tiled state of the view in
  interactive_begin() when starting to move the view.  When done,
  interactive_end() will update the tiled state if appropriate but
  *not* overwrite the natural geometry.
- Refactor interactive_begin() a bit.
- Split interactive_end() into interactive_finish() and
  interactive_cancel(). Split out snap_to_edge().

src/interactive.c
jlindgren90: Oops! Fixed.

src/interactive.c
jlindgren90: Added inline comments.

src/interactive.c
jlindgren90: Good call. I reorganized `interactive_begin()` so that there is now one `switch(mode)` rather than the various `if` statements.

src/view.c
jlindgren90: Added comments.

src/view.c
jlindgren90: Added comments.

src/view.c
jlindgren90: Added inline comment.

johanmalm:
Works well. Another great contribution :smile: Thanks.
And thanks for review @Consolatis

Consolatis:
Nice changes. Thanks :)

PR #638, opened 2022-11-21
include: Break out view.h from labwc.h

jlindgren90:
IMHO it encourages better design (by making dependencies more obvious) to have source file/header file pairs like view.c/view.h, rather than a monolithic header like labwc.h with everything in it.

I don't think we need to break up all of labwc.h at once, but maybe we can start pulling it apart bit by bit as it's convenient.

Also:

- Move `struct border` to ssd.h so that view.h can use it without pulling in all of labwc.h.
- Add a missing required `#include` within scaled_font_buffer.h (forward declaration of `struct font` is not enough).


johanmalm:
LGTM.
I like this a lot and we should have probably done this ages ago :smile:

PR #640, opened 2022-11-22
ssd: Hide implementation details from the outside world

jlindgren90:
The basic idea is as the title says, to hide as much of the SSD logic so that it's only visible to the source files starting in `ssd_`.

Changes in detail:

- Add new ssd-internal.h header
- Move all ssd structs and internal functions to ssd-internal.h
- Allocate `struct ssd` and `struct ssd_hover_state` separately
- Add a few accessors to replace direct access into ssd structs
- Store a pointer to the `struct view` in `struct ssd`
- Pass `struct ssd *` instead of `struct view *` to ssd functions
- Add `view_reload_ssd()`, replacing `ssd_reload()` and surrounding logic
- Move some parts of debug functions to ssd.c

The fact that we now have both `ssd_new()` and `ssd_create()` isn't ideal (though the functions do have different purposes). Perhaps at some point we can revisit those and clarify the naming at least.

There shouldn't be any functional change here, but more testing would definitely be appreciated to be sure my fingers didn't slip somewhere ;)

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -1,0 +1,141 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_SSD_INTERNAL_H
+#define __LABWC_SSD_INTERNAL_H
+
+#include "ssd.h"
+
+#define FOR_EACH(tmp, ...) \
+{ \
+	__typeof__(tmp) _x[] = { __VA_ARGS__, NULL }; \
+	size_t _i = 0; \
+	for ((tmp) = _x[_i]; _i < sizeof(_x) / sizeof(_x[0]) - 1; (tmp) = _x[++_i])
+
+#define FOR_EACH_END }
+
+struct ssd_button {
+	struct view *view;
+	enum ssd_part_type type;
+	struct wlr_scene_node *hover;
+
+	struct wl_listener destroy;
+};
+
+struct ssd_sub_tree {
+	struct wlr_scene_tree *tree;
+	struct wl_list parts; /* ssd_part::link */
+};
+
+struct ssd_state_title_width {
+	int width;
+	bool truncated;
+};
+
+struct ssd {
+	struct view *view;
+	struct wlr_scene_tree *tree;
+
+	/*
+	 * Cache for current values.
+	 * Used to detect actual changes so we
+	 * don't update things we don't have to.
+	 */
+	struct {
+		int x;
+		int y;
+		int width;
+		int height;
+		struct ssd_state_title {
+			char *text;
+			struct ssd_state_title_width active;
+			struct ssd_state_title_width inactive;
+		} title;
+	} state;
+
+	/* An invisble area around the view which allows resizing */
+	struct ssd_sub_tree extents;
+
+	/* The top of the view, containing buttons, title, .. */
+	struct {
+		/* struct wlr_scene_tree *tree;      unused for now */
+		struct ssd_sub_tree active;
+		struct ssd_sub_tree inactive;
+	} titlebar;
+
+	/* Borders allow resizing as well */
+	struct {
+		/* struct wlr_scene_tree *tree;      unused for now */
+		struct ssd_sub_tree active;
+		struct ssd_sub_tree inactive;
+	} border;
+
+	/*
+	 * Space between the extremities of the view's wlr_surface
+	 * and the max extents of the server-side decorations.
+	 * For xdg-shell views with CSD, this margin is zero.
+	 */
+	struct border margin;
+};
+
+struct ssd_part {
+	enum ssd_part_type type;
+
+	/* Buffer pointer. May be NULL */
+	struct scaled_font_buffer *buffer;
+
+	/* This part represented in scene graph */
+	struct wlr_scene_node *node;
+
+	/* Targeted geometry. May be NULL */
+	struct wlr_box *geometry;
+
+	struct wl_list link;
+};
+
+struct ssd_hover_state {
+	struct view *view;
+	struct wlr_scene_node *node;
+};
+
+struct wlr_buffer;
+struct wlr_scene_tree;
+
+/* SSD internal helpers to create various SSD elements */
+/* TODO: Replace some common args with a struct */
+struct ssd_part *add_scene_part(
+	struct wl_list *part_list, enum ssd_part_type type);
+struct ssd_part *add_scene_rect(
+	struct wl_list *list, enum ssd_part_type type,
+	struct wlr_scene_tree *parent, int width, int height, int x, int y,
+	float color[4]);
+struct ssd_part *add_scene_buffer(
+	struct wl_list *list, enum ssd_part_type type,
+	struct wlr_scene_tree *parent, struct wlr_buffer *buffer, int x, int y);
+struct ssd_part *add_scene_button(
+	struct wl_list *part_list, enum ssd_part_type type,
+	struct wlr_scene_tree *parent, float *bg_color,
+	struct wlr_buffer *icon_buffer, int x, struct view *view);
+struct ssd_part *add_scene_button_corner(
+	struct wl_list *part_list, enum ssd_part_type type,
+	enum ssd_part_type corner_type, struct wlr_scene_tree *parent,
+	struct wlr_buffer *corner_buffer, struct wlr_buffer *icon_buffer,
+	int x, struct view *view);
+
+/* SSD internal helpers */
+struct ssd_part *ssd_get_part(
+	struct wl_list *part_list, enum ssd_part_type type);
+void ssd_destroy_parts(struct wl_list *list);
+
+/* SSD internal */
+void ssd_titlebar_create(struct ssd *ssd);
+void ssd_titlebar_update(struct ssd *ssd);
+void ssd_titlebar_destroy(struct ssd *ssd);
+
+void ssd_border_create(struct ssd *ssd);
+void ssd_border_update(struct ssd *ssd);
+void ssd_border_destroy(struct ssd *ssd);
+
+void ssd_extents_create(struct ssd *ssd);
+void ssd_extents_update(struct ssd *ssd);
+void ssd_extents_destroy(struct ssd *ssd);
+
+#endif /* __LABWC_SSD_INTERNAL_H */
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -10,7 +10,7 @@
 #include "common/mem.h"
 #include "common/scene-helpers.h"
 #include "labwc.h"
-#include "ssd.h"
+#include "ssd-internal.h"
 #include "theme.h"
 #include "view.h"
 
@@ -284,6 +284,18 @@ struct ssd_hover_state *
 ssd_hover_state_new(void)
 {
 	return znew(struct ssd_hover_state);
+}
+
+enum ssd_part_type
+ssd_button_get_type(const struct ssd_button *button)
+{
+	return button ? button->type : LAB_SSD_NONE;
+}
+
+struct view *
+ssd_button_get_view(const struct ssd_button *button)
+{
+	return button ? button->view : NULL;
 }
 
 bool
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -8,14 +8,6 @@
 #define BUTTON_WIDTH 26
 #define EXTENDED_AREA 8
 
-#define FOR_EACH(tmp, ...) \
-{ \
-	__typeof__(tmp) _x[] = { __VA_ARGS__, NULL }; \
-	size_t _i = 0; \
-	for ((tmp) = _x[_i]; _i < sizeof(_x) / sizeof(_x[0]) - 1; (tmp) = _x[++_i])
-
-#define FOR_EACH_END }
-
 /*
  * Sequence these according to the order they should be processed for
  * press and hover events. Bear in mind that some of their respective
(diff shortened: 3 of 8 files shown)


src/debug.c
```diff
@@ -175,8 +158,8 @@ dump_tree(struct server *server, struct wlr_scene_node *node,
 	printf("%s %*c %4d  %4d  [%p]\n", type, padding, ' ', x, y, node);
 
 	if ((IGNORE_MENU && node == &server->menu_tree->node)
-			|| (IGNORE_SSD && view && view->ssd.tree
-			&& node == &view->ssd.tree->node)) {
+			|| (IGNORE_SSD && view
+			&& node == ssd_get_tree_node(view->ssd))) {
```
Consolatis, line 162: Maybe this could be `ssd_is_treenode(node)` instead?

src/ssd/ssd.c
```diff
@@ -278,13 +304,48 @@ ssd_part_contains(enum ssd_part_type whole, enum ssd_part_type candidate)
...
+/* debug */
+struct wlr_scene_node *
+ssd_get_tree_node(const struct ssd *ssd)
+{
+	return ssd->tree ? &ssd->tree->node : NULL;
+}
+
+/* debug */
+const char *
+ssd_get_view_part(const struct ssd *ssd, struct wlr_scene_node *node)
```
Consolatis, line 327: I am not sure about this, It accesses internal state of the SSD so having it in `src/ssd/` kind of makes sense.
On the other side this is only relevant for the debug action. Maybe we could move these two functions to `src/ssd/ssd_debug.c` and maybe even create a separate `include/ssd_debug.h`? It's really hard for me to figure out the right granularity here.

Also: something like `ssd_debug_name(const struct ssd *ssd, struct wlr_scene_node *node)` would be more self explaining I think and could be expanded in the future so we also have a `view_debug_name()` or similar. So each "module" would describe itself instead of having the external debug.c caller poke around in "module internal" states.

src/workspaces.c
```diff
@@ -7,12 +7,12 @@
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
+#include "buffer.h"
```
Consolatis, line 10: This seems unrelated to the PR

src/workspaces.c
```diff
@@ -7,12 +7,12 @@
...
 #include "common/font.h"
 #include "common/graphic-helpers.h"
 #include "common/list.h"
 #include "common/mem.h"
 #include "labwc.h"
-#include "view.h"
```
Consolatis, line 15: This seems unrelated to the PR

src/xdg.c
```diff
@@ -276,8 +276,10 @@ position_xdg_toplevel_view(struct view *view)
 		view->x = center_x - view->xdg_surface->current.geometry.width / 2;
 		view->y = center_y - view->xdg_surface->current.geometry.height / 2;
 	}
-	view->x += view->ssd.margin.left;
-	view->y += view->ssd.margin.top;
+
+	struct border margin = ssd_get_margin(view->ssd);
```
Consolatis, line 280: I think this is a hot path in which case the additional function call might actually impact the perceived latency on low power systems. Not sure how to deal with this though, maybe we could cache it somewhere?

src/view.c
```diff
@@ -921,6 +934,7 @@ view_destroy(struct view *view)
 
 	/* Remove view from server->views */
 	wl_list_remove(&view->link);
+	free(view->ssd);
```
Consolatis, line 937: Having `ssd_destroy()` not actually destroy the SSD seems a bit strange on the first look. But I think we have a similar pattern with `xxx_finish()` and `xxx_destroy()` somewhere else already. Not sure right now if its `theme.c` or `rcxml.c`. We should likely decide on one naming convention, add it to Contributing.md and stick with it.

The straight `free()` call seems wrong to me though, `ssd_free()` (or a better named variant) would be more intuitive IMHO because we didn't `calloc()` / `znew()` it ourselves but instead called `ssd_create()` in `xdg.c` / `xwayland.c`.

Consolatis:
Had a quick look through the diff and added some comments about some things that stood out to me. I didn't verify the logic but it indeed seems to not change anything. Will look through the diff again later.

include/ssd-internal.h
```diff
@@ -0,0 +1,139 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_SSD_INTERNAL_H
```
Consolatis, line 2: For `src/action.c` I used `include/private/action.h`, but I actually tend to like the `-internal` variant better (e.g. `include/action-internal.h`).

We should likely discuss having the same place and name + `-internal.h` or having it sub-directory based (e.g. `include/internal/ssd.h`), write that down in Contributing.md and then migrate both cases to whatever we decide on. Pinging @johanmalm as well.

jlindgren90: I like to avoid having two headers with the same name -- it tends to confuse IDEs. I can go ahead and rename `include/private/action.h` to `include/action-internal.h` if you're good with that.

src/workspaces.c
jlindgren90: Previously `buffer.h` was included indirectly via `ssd.h`, now that's no longer the case so we need to include it here to avoid compiler errors.

src/workspaces.c
jlindgren90: Right, in hindsight it should have been part of #639, but I missed it then. Didn't seem worth opening a new MR for since I needed to fix up the includes in this file anyway.

src/view.c
jlindgren90: I'll add `ssd_free()`, it seems like a little overkill now, but I agree it's consistent and is good future-proofing.

src/debug.c
jlindgren90: Changed to `ssd_debug_is_root_node()`

src/ssd/ssd.c
jlindgren90: Renamed to `ssd_debug_get_node_name()`. Since there are only two `ssd_debug` functions, I don't think we necessarily need to add a new source file + header at this point.

src/view.c
jlindgren90: Added `ssd_free()`

include/ssd-internal.h
jlindgren90: Looking at it again, `include/private/action.h` was included only in `action.c`, so I just inlined it there,

jlindgren90:
v2:
- Add `ssd_free()`
- Prefix debug functions with `ssd_debug_`
- Inline `private/action.h`

include/ssd-internal.h
johanmalm: Looks like there are no more `include/private/` then. I thought we had a 'private' header file common to the files in `src/ssd`, but can't find it now. Maybe my memory is playing tricks on me.

I don't feel strongly on this, but my preference is as follows in terms of location/naming of 'private' header files (starting with highest preference):

1. Put 'private' header files in `src/<subdir-that-wants-common-header>` using the format `foo_p.h`. This seems clean and self-contained, but we lose the ability to grep `include/` for function declarations (which isn't a big deal for me).
2. Put 'private' header files in 'include/' and calling them `foo_p.h`
3. Same as (3) but call them `foo-internal.h`

johanmalm:
Look good at a glance. Let me know if you want me to do a thorough review.

include/ssd-internal.h
jlindgren90: I am fine with any of these, though (1) is a little bit unclear when there isn't a subdirectory; for example I am considering adding a `view-internal.h` that would be used from `view.c`, `view-impl.c`, `xdg.c`, and `xwayland.c`.

jlindgren90:
v3: Simplify `ssd_new`/`ssd_create`/`ssd_enabled` logic

Now the view either has SSD enabled, or `view->ssd == NULL`; there are no in-between states. This should avoid the `ssd_new`/`ssd_create` confusion.

src/view.c
jlindgren90: Removed `ssd_free()` again and made `ssd_destroy()` actually free the `struct ssd`

include/ssd-internal.h
johanmalm: @Consolatis and I pondered these private files before and not really a clean approach that fits all.

Sway often has larger .c files and thus less need for private header files.

At the end of the day, it’s not a library so we don’t have to be super strict about public/private. So it comes down to readability/maintainability.


The xbm/ dir is a good example. We should only really ‘expose’ one function in the include/ header. The rest is local implementation detail.
Sway probably would have put it all in one .c file. 

I think that when it’s starts getting a bit grey it’s probably a good time to stop splitting files 😄 

Whats the common view/xdg/xwayland implementation detail? Are you thinking of the structs?

jlindgren90: > Whats the common view/xdg/xwayland implementation detail? Are you thinking of the structs?

What I am hoping to do is to make `struct view` effectively read-only outside `view.c` and friends, so as to enforce that e.g. `view->x` can't be overwritten from `desktop.c` directly; it has to call `view_move()` or similar.

So I was imagining something along the lines of:
- Make `struct view` itself opaque (outside `view-internal.h`).
- Put most of the flags/geometry/etc. in a nested struct (`view_state` or such).
- Then have an accessor `const view_state *view_get_state(view *)`. This would allow broad read-only access to e.g. `view->state->x` or `view->state->maximized` without allowing those fields to be overwritten directly.

I haven't fully fleshed out the idea, and it may or may not still make sense once #641 lands.

src/view.c
```diff
@@ -554,9 +555,21 @@ void
...
+			view == view->server->focused_view);
+	} else {
+		if (!view->ssd) {
+			return;
+		}
+		ssd_destroy(view->ssd);
+		view->ssd = NULL;
+	}
+
+	if (!view->fullscreen) {
```
jlindgren90, line 572: The existing logic made `view_set_decorations()` a no-op for a fullscreen view, and I am not sure why.
Also, it looks like the decorations are not actually hidden for a fullscreen view (are they just off-screen?)
So it seems like it's okay to do the `ssd_create/destroy()` part for a fullscreen view, which will result in the decorations being visually enabled/disabled once the view is no longer fullscreen. But I'd appreciate a second pair of eyes on this.

src/view.c
```diff
@@ -866,6 +881,17 @@ view_update_app_id(struct view *view)
 		view->toplevel_handle, app_id);
 }
 
+void
+view_reload_ssd(struct view *view)
+{
+	assert(view);
+	if (view->ssd) {
```
jlindgren90, line 888: The existing logic (that `view_reload_ssd()` replaces) would skip over views that aren't mapped, and not recreate the SSD nodes for them. My understanding is that when we unmap a view, the entire scene tree (including the SSD nodes) is disabled, and so calling `ssd_destroy()/ssd_create()` is harmless, the nodes are invisible anyway. But I'd appreciate a second pair of eyes on this logic.

include/ssd-internal.h
Consolatis: > Put 'private' header files in src/<subdir-that-wants-common-header> using the format foo_p.h. This seems clean and self-contained, but we lose the ability to grep include/ for function declarations (which isn't a big deal for me).

It would be for my workflow, I'd really prefer all the headers being in `include/`.

> for example I am considering adding an include/view-internal.h that would be used from view.c, view-impl.c, xdg.c, and xwayland.c.

Which raises the question if we want a "module" directory for those (+ `xwayland-unmanaged.c`), similar to `src/ssd/`. I may be overshooting with this though.

src/ssd/ssd.c
Consolatis: Sounds good

src/view.c
Consolatis: The reason for skipping those and the reason for that 3rd allocated-but-disabled state was performance improvements.
That way we don't have to create everything from scratch (including uploading textures to the GPU by wlroots) when leaving fullscreen or toggling the SSD for another reason.

Due to the use of for example the autoscaling title buffer we may also have to wait for an enter event until pango renders the title in the new output scale.

jlindgren90: Re: reloading, if we skip unmapped views then we would need to set a flag to reload them next time they are mapped, right? Seems more straightforward to just do the reload immediately for all views that have SSD (mapped or not).

Re: toggling SSD, is it something that happens frequently enough to be worth caching those textures? I guess I'd vote for simplicity of implementation over performance here unless it's having a noticeable impact.

include/ssd-internal.h
jlindgren90: A `view` subdirectory might be nice in future.

For this MR, is it okay to stick with `include/ssd-internal.h` as it stands currently?

johanmalm: Yes, fine.

jlindgren90:
Rebased on top of #641.

@Consolatis @johanmalm From my point of view this is ready to merge; what do you think?

johanmalm:
It all looks good with the following exceptions:

1. Positioning of views appears to no longer take into account SDD. For example, when do a snap-to-edge, the view is overlaps the panel (see scrot below).
2. I have confirmed the full-screen bug (which exists both with and without this patch-set). In full screen, the SSD spills over to adjacent outputs. So we ought to disable (or destroy with this patch-set) on fullscreen.

A couple of side-notes:

a. I'm fine with `*_destroy/free/etc` in this patch-set, but would appreciate thinking through consistent naming as a separate exercise. `wlroots` have defined this in their [CONTRIBUTING.md](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/CONTRIBUTING.md#constructiondestruction-functions) but I don't find it easy to apply consistently, no do I think `wlroots` have.
b. I'm fine with this patch-set, but generally find it easier to review with smaller, atomic commits.

20221125_21h05m23s_grim

jlindgren90:
I'll work on addressing those two issues, and split this into smaller MRs.
@johanmalm For the sake of reproducing, which panel is in your screenshot?

johanmalm:
> I'll work on addressing those two issues, and split this into smaller MRs. @johanmalm For the sake of reproducing, which panel is in your screenshot?

waybar

src/view.c
Consolatis: > I guess I'd vote for simplicity of implementation over performance here unless it's having a noticeable impact.

I am fine with trying the destruct / recreate logic for now. If we detect that this will introduce lags on low power systems especially when un-fullscreening (e.g. video in a browser) we should be able to reintroduce the disable / enable flow again relatively easy.

We had issues before like for example the mouse cursor was starting to lag really bad when alt-tabbing out of a GPU intensive game which was caused by higher latency to upload the cursor texture to the GPU (and us not preventing to set the new cursor image to the one already set). At some points wlroots will likely add a feature to reuse textures for wlr_buffers that were bound to multiple wlr_scene_buffer nodes (e.g. all SSD icons and rounded corners) but as of right now each one of them gets a new texture uploaded for every SSD instance AFAIR.

src/view.c
jlindgren90: Addressed in #650.

jlindgren90:
Rebased on top of #654. I believe I've addressed both of the issues that @johanmalm noticed as well.

johanmalm:
Thanks!

PR #641, opened 2022-11-22
view: create derived structs

johanmalm:
...for XDG toplevels and XWayland views to only include applicable wl_listeners for each type of view.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -9,6 +9,11 @@
 #include <wlr/util/box.h>
 #include "ssd.h"
 
+/*
+ * In labwc, a view is a container for surfaces which can be moved around by
+ * the user. In practice this means XDG toplevel and XWayland windows.
+ */
+
 enum view_type {
 	LAB_XDG_SHELL_VIEW,
 #if HAVE_XWAYLAND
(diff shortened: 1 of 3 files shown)


johanmalm:
@Consolatis @jlindgren90 I'd appreciate  your views on this approach. I think it makes it a bit purer/cleaner. Not sure if it makes it easier or harder to understand.

jlindgren90:
Based on a quick look, I like the approach.

Would it make sense to also put the `xdg_surface` and `xwayland_surface` in the respective derived structs? I'm not really a fan of the current union-of-pointers approach (it seems dangerous if someone forgets a `view->type` check).

johanmalm:
> Would it make sense to also put the `xdg_surface` and `xwayland_surface` in the respective derived structs? I'm not really a fan of the current union-of-pointers approach (it seems dangerous if someone forgets a `view->type` check).

Oh, that's a thought. Will have a go tomorrow. Getting late here.

Consolatis:
I like this. Do we want to use a function or macro for the casting from `struct view` to the extended ones? That one could then also assert `view->view_type`, similar to action args and node descriptors.

johanmalm:
I've implemented Consolatis's suggestion.
The xwayland_surface / xdg_surface change ends up being a bit more intrusive, so can't make up my mind if it's the right thing to do.
Suggest we interactive-rebase-fixup this and merge.

PR #645, opened 2022-11-25
docs: add default mousebind example

Consolatis:
Should make it easier for users to discover the requirement of `<default />` when having custom mousebinds in place.

~~Not actually tested yet, thus draft for now.~~

### DIFF
--- a/docs/menu.xml
+++ b/docs/menu.xml
@@ -18,6 +18,10 @@
   <item label="AlwaysOnTop">
     <action name="ToggleAlwaysOnTop" />
   </item>
+  <!--
+    Any menu with the id "workspaces" will be hidden
+    if there is only a single workspace available.
+  -->
   <menu id="workspaces" label="Workspace">
     <item label="Move left">
       <action name="SendToDesktop" to="left" />
--- a/docs/menu.xml
+++ b/docs/menu.xml
@@ -3,51 +3,51 @@
 <openbox_menu>
 
 <menu id="client-menu">
-	<item label="Minimize">
-		<action name="Iconify" />
-	</item>
-	<item label="Maximize">
-		<action name="ToggleMaximize" />
-	</item>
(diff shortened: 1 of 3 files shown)


johanmalm:
LGTM

Consolatis:
Reformatted `docs/rc.xml` and `docs/menu.xml` to
- use tabs instead of spaces
- use the more compact form (e.g. `<action name="foo" />` instead of `<action><name>foo</name></action>`)

The mousebind example now shows `some-custom-menu` on right click on the desktop so we now also show how to create a custom menu.

johanmalm:
Looks good except that we're change the indentation standard for XML files.

https://github.com/labwc/labwc/blob/7ad5490ff00696f898d57fd78dbf7c28b58bc78c/.editorconfig#L13-L14

I'm not particularly bothered either way to be honest.

Consolatis:
>  except that we're changing the indentation standard for XML files.

Oh, we have an `.editorconfig`?

I can revert that change or update `rc.xml.all` + `.editorconfig` as well.
Personally I prefer tabs for xml files but I don't feel strongly about it either.

johanmalm:
> I can revert that change or update `rc.xml.all` + `.editorconfig` as well. Personally I prefer tabs for xml files but I don't feel strongly about it either.

Wayland and wlroots protocols use 2-space-indentation, so I think we ought to go with that. It also helps keep line length shorter which can be tricky with `rc.xml` and `menu.xml`.

I like all the other changes.

Consolatis:
Changes:

- Reverted the change to tabs
- Converted some tabs in `menu.xml` `client-menu` to spaces
- gave `rc.xml.all` the same compact form treatment (e.g. `<action name="foo" />` instead of `<action><name>foo</name></action>`)
- Ensured all configs have a space before the closing `"/>` part.

johanmalm:
Thanks!

PR #648, opened 2022-11-25
view: Move xdg_surface + xwayland_surface to derived structs

jlindgren90:
Following up on a suggestion in #641.

- Add `view_get_xdg_surface()` + `view_get_xwayland_surface()` accessors that `assert()` the view is of the expected type before returning.
- Fix a real bug in `xdg.c` `parent_of()` that dereferenced `view->xdg_surface->toplevel` without first checking `view->type`.

The goal of the new accessors is to catch similar bugs in future.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -7,6 +7,25 @@
 #include "view.h"
 #include "workspaces.h"
 
+struct wlr_xdg_surface *
+xdg_surface_from_view(struct view *view)
+{
+	assert(view->type == LAB_XDG_SHELL_VIEW);
+	struct xdg_toplevel_view *xdg_toplevel_view =
+		(struct xdg_toplevel_view *)view;
+	assert(xdg_toplevel_view->xdg_surface);
+	return xdg_toplevel_view->xdg_surface;
+}
+
+static struct wlr_xdg_toplevel *
+xdg_toplevel_from_view(struct view *view)
+{
+	struct wlr_xdg_surface *xdg_surface = xdg_surface_from_view(view);
+	assert(xdg_surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL);
+	assert(xdg_surface->toplevel);
+	return xdg_surface->toplevel;
+}
+
 static void
 handle_new_xdg_popup(struct wl_listener *listener, void *data)
 {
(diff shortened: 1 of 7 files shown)


src/xdg.c
```diff
@@ -7,6 +7,25 @@
 #include "view.h"
 #include "workspaces.h"
 
+struct wlr_xdg_surface *
+view_get_xdg_surface(struct view *view)
```
johanmalm, line 11: I like to prefix extern functions with the name of the translation unit file (e.g. in `xdg.c` extern functions start with `xdg_`).
So, either put this function in `view.c` or call is something like `xdg_surface_from_view()`. Before I got too tired last night I started with the latter approach, but really don't mind which.

src/xdg.c
```diff
@@ -212,44 +234,38 @@ xdg_toplevel_view_move(struct view *view, int x, int y)
...
 }
 
 static void
 xdg_toplevel_view_set_activated(struct view *view, bool activated)
 {
-	struct wlr_xdg_surface *surface = view->xdg_surface;
-	if (surface->role == WLR_XDG_SURFACE_ROLE_TOPLEVEL) {
-		wlr_xdg_toplevel_set_activated(surface->toplevel, activated);
-	}
+	wlr_xdg_toplevel_set_activated(get_xdg_toplevel(view), activated);
```
johanmalm, line 249: We're changing the logic slightly here. I can't think that it'll be a problem, but have not tested.

johanmalm:
Thanks for taking this the next step. A couple of review comments - otherwise LGTM.

src/xdg.c
jlindgren90: The `wlr_xdg_surface::role` should never be anything other than `WLR_XDG_SURFACE_ROLE_TOPLEVEL` or we would not have created an `xdg_toplevel_view` in the first case. So I believe it is safe to remove the `if` here.

johanmalm: Yes, let's go for it. I can't remember why it's there, but have tried running with a couple of GTK (xdg-shell) applications and it seems to behave as expected.

src/xdg.c
jlindgren90: Fair enough, I'll rename the new functions to `xdg_surface_from_view()` + `xwayland_surface_from_view()` etc.

jlindgren90:
v2: Updated naming per suggestion from @johanmalm

PR #650, opened 2022-11-26
view: Hide decorations for fullscreen views

jlindgren90:
Fixes the pre-existing issue that @johanmalm noticed in #640, where decorations are shown around fullscreen views and can be visible on other outputs.

- Update `view_set_decorations()` and `view_set_fullscreen()` to disable decorations for fullscreen views
- Remove enable/disable logic from `ssd_update_geometry()` since it is now handled by `view_set_decorations()` and `view_set_fullscreen()`

Some minor fixes/cleanups as well:

- Use C99-style struct literals in `return` statements to eliminate unnecessary temporary variables
- Add an `active` argument to `ssd_create()` for consistency with `ssd_set_active()`
- Update `view->ssd.state` at the end of `ssd_create()` so that the next call to `ssd_update_geometry()` doesn't do unnecessary work
- Move `ssd_reload()` to `view_reload_ssd()` since it fits better in `view.c`
- Reset `view->ssd.margin` to zero at the end of `ssd_destroy()` to prevent accidentally using stale values

Seems to work in my testing, but additional testing would be appreciated. When leaving fullscreen, you can still see the decorations flicker for a frame or so around the fullscreen geometry -- I think this is because move/resize is an async operation, whereas redisplaying the decorations is immediate.

### DIFF
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -146,7 +146,7 @@ struct ssd_hover_state {
 };
 
 /* Public SSD API */
-void ssd_create(struct view *view);
+void ssd_create(struct view *view, bool active);
 void ssd_set_active(struct view *view, bool active);
 void ssd_update_title(struct view *view);
 void ssd_update_geometry(struct view *view);
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -150,7 +150,6 @@ void ssd_create(struct view *view);
 void ssd_set_active(struct view *view, bool active);
 void ssd_update_title(struct view *view);
 void ssd_update_geometry(struct view *view);
-void ssd_reload(struct view *view);
 void ssd_destroy(struct view *view);
 void ssd_update_button_hover(struct wlr_scene_node *node,
 	struct ssd_hover_state *hover_state);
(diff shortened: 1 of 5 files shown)


jlindgren90:
(intended to be merged before #640 (which will need to be rebased))

Consolatis:
I would strongly prefer to split this up into two commits, one changing the reload logic to actually destroy and recreate the SSD instead of just toggling the scene_tree and the other for the actual change to disable the SSD when fullscreening.

jlindgren90:
Sure, I can split it up.

jlindgren90:
v2: Split up into more granular commits. Overall diff is identical.

johanmalm:
LGTM. Tested and works fine.

johanmalm:
Suggestion for a different day (for when one of us has time/motivation - if indeed you agree): should we replace `view->{x,y,w,h}` with a `struct wlr_box`?

jlindgren90:
> Suggestion for a different day (for when one of us has time/motivation - if indeed you agree): should we replace `view->{x,y,w,h}` with a `struct wlr_box`?

That has crossed my mind too :)

Consolatis:
Yeah, mine as well.

There is also a `wlr_box_equal()` which might come in handy eventually.

jlindgren90:
> Sorry if I came across a bit salty here btw, RL currently takes some energy away.

No worries, and sending best wishes for restored energy :)

PR #651, opened 2022-11-26
ssd: Cosmetic cleanups & add debug helpers

jlindgren90:
Split off from #640.

- Minimize includes in `ssd.h`
- Avoid repetitive `view->ssd.margin` pattern
- Use `struct ssd *` or `const struct ssd *` rather than `struct view *`
  where convenient
- Add `ssd_debug` helpers

Part of the motivation is to make it easier to separate `struct ssd`
from `struct view` in a future commit (i.e. the rest of #640).

### DIFF
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -2,8 +2,7 @@
 #ifndef __LABWC_SSD_H
 #define __LABWC_SSD_H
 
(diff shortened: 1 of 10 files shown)


johanmalm:
LGTM

PR #652, opened 2022-11-26
ci: run checkpatch.pl on all .c and .h files

johanmalm:


Consolatis:
In the future we might want to only run the check on the Ubuntu host before starting the builds for the containers / FreeBSD VM.
That way the build is not even tried when the code check failed. But we can adjust this later on so I'll merge this for now.

Thanks for all your work on the checkpatch.pl stuff :)

PR #654, opened 2022-11-26
ssd: Allocate `struct ssd` and `struct ssd_hover_state` separately

jlindgren90:
Allocating these structs separately allows them to be made opaque in a later commit.

- Store a pointer to the `struct view` in `struct ssd`
- Pass `struct ssd *` instead of `struct view *` to ssd functions
- Add `ssd_get_margin()` convenience function

Split off from #640.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -50,7 +50,6 @@
 #include "cursor.h"
 #include "config/keybind.h"
 #include "config/rcxml.h"
-#include "ssd.h"
 #if HAVE_NLS
 #include <libintl.h>
 #include <locale.h>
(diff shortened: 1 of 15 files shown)


Consolatis:
Didn't know GH allows markdown in titles, learning something new every day :)

I did not test this but on a quick look didn't spot anything suspicious.

jlindgren90:
> Didn't know GH allows markdown in titles, learning something new every day :)

Me too, I wrote the backticks without thinking and was surprised to see that afterward.

johanmalm:
Looks good

PR #655, opened 2022-11-27
view: Fix SSD margin computation

### LINKED ISSUE #653: 2 issues after PR #650
imv now opens with ssd off screen. Same if you start labwc nested. Both windows open in the top left, might be related?

foot ssd gets moved partially off screen when snapping to edge. If I reconfigure labwc while foot is open and then snap it, it works fine.

[Screenshot showing both bugs](https://user-images.githubusercontent.com/4444593/204109708-3b8fe729-f901-4bf5-8084-90f321c49b7f.png)

jlindgren90:
If `view->ssd_enabled` is false when calling `ssd_create()` then `ssd_thickness()` returns zeroes which are stored in `ssd->margins`.

The quick fix is just to ensure we set `view->ssd_enabled` before calling `ssd_create()`. At some point, it might be nice to refactor so that `ssd_create()` does not reference `view->ssd_enabled`.

Fixes #653 which was a regression caused by #650 (oops).

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -559,12 +559,16 @@ view_set_decorations(struct view *view, bool decorations)
 {
 	assert(view);
 	if (view->ssd_enabled != decorations && !view->fullscreen) {
+		/*
+		 * Set view->ssd_enabled first since it is referenced
+		 * within the call tree of ssd_create()
+		 */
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks

PR #657, opened 2022-11-28
CI: check codestyle only once

Consolatis:


Consolatis:
This checks for codestyle first and only runs the build tests when that one has been successful. I think we could also configure the CI to ignore the success status of the codestyle and thus still run the build tests but I think I'll prefer the early exist on codestyle errors.

PR #660, opened 2022-11-29
README: restructure and amend project description

johanmalm:
- Describe why the projects exists and state some core-dev values
- Explain that labwc does not depend on any sessions or UI toolkits
- Merge the two scope sections
- Put video links before screenshot

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,22 @@ href="https://github.com/labwc/labwc-scope#readme">Scope</a>] [<a
 href="https://web.libera.chat/gamja/?channels=#labwc">IRC&nbsp;Channel</a>] [<a
 href="NEWS.md">Release&nbsp;Notes</a>]</h3>
 
-- [1. What is this?](#1-what-is-this)
-  - [1.1 Screenshot](#11-screenshot)
-  - [1.2 Videos](#12-videos)
+- [1. Project Description](#1-project-description)
+  - [1.1 What Is This?](#11-what-is-this)
+  - [1.2 Why](#12-why)
+  - [1.3 Why The Openbox Theme Specification?](#13-why-the-openbox-specification)
+  - [1.4 Very High Level Scope](#14-very-high-level-scope)
+  - [1.5 Videos](#15-videos)
+  - [1.6 Screenshot](#16-screenshot)
 - [2. Build and Installation](#2-build-and-installation)
 - [3. Configuration](#3-configuration)
 - [4. Theming](#4-theming)
 - [5. Usage](#5-usage)
 - [6. Integration](#6-integration)
-- [7. Scope](#7-scope)
 
-## 1. What is this?
+## 1. Project Description
+
+### 1.1 What Is This?
 
 Labwc stands for Lab Wayland Compositor, where lab can mean any of the
 following:
@@ -25,8 +30,8 @@ following:
 - inspired by BunsenLabs and ArchLabs
 - your favorite pet
 
-Labwc is a [wlroots]-based window-stacking compositor for [wayland], inspired by
-[openbox].
+Labwc is a [wlroots]-based window-stacking compositor for [wayland], inspired
+by [openbox].
 
 It is light-weight and independent with a focus on simply stacking windows well
 and rendering some window decorations. It takes a no-bling/frills approach and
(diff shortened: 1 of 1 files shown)


johanmalm:
@Consolatis @jlindgren90 
This one is a bit more existential, so would appreciate your views before I press "Ready for Review"
I'm not hard over on any of this, so happy to take things out or modify until we're all happy.

README.md
```diff
@@ -38,41 +43,99 @@ on to create a full desktop environment.
...
+balance between minimalism and bloat approximately at the level where openbox
+resides in the X11 domain. This does not mean that it should look square and
+pixelated like something out of the 1990s or indeed that the improved
+performance offered by Wayland should be compromised.
+
+It is important to us that scope is tightly controlled so that the compositor
+matures into production quality. On the whole, we value robustness, reliability
+stability and simplicity over new features. This is because we believe that
+coming up with and implementing new ideas is easy, but making them stick and
+work for a long time hard. We want the latter.
```
jlindgren90, line 61: "long time is hard" (missing word)

README.md
```diff
@@ -38,41 +43,99 @@ on to create a full desktop environment.
...
 
-See [scope] for full details on implemented features.
-
-## 1.1 Screenshot
+Items which are *not* intended to be implemented:
 
-<a href="https://i.imgur.com/vOelinT.png">
-  <img src="https://i.imgur.com/vOelinTl.png">
-</a>
+- Icons (except window buttons)
```
jlindgren90, line 117: This line isn't new, I know, but I've wondered whether it means that application icons within title bars are in or out of scope. It's something I miss from Openbox.

README.md
```diff
@@ -38,41 +43,99 @@ on to create a full desktop environment.
...
+coming up with and implementing new ideas is easy, but making them stick and
+work for a long time hard. We want the latter.
+
+[KDE/kwin] and [GNOME/mutter] are great compositors offering brilliant
+experiences, but at ca 200k and 400k lines-of-code respectively and with heavy
+reliance on their respective eco-systems, they cannot be an elegant solution to
+the need for a frugal, independent, Wayland stacking compositor which is the
+space that labwc occupies.
+
+The Wayland community has achieved an amazing amount so far, and we want help
```
jlindgren90, line 69: "and we want to help" (missing word)

jlindgren90:
Looks good to me. I noticed a couple typos and had one question, but I'm happy with all of this.

README.md
Consolatis: Application icons are really tricky with wayland (coming from somebody who implemented application icon lookup for a panel). Wayland doesn't provide any notion of icons but has the app-id. By convention (!) the app-id is mapped to a `.desktop` file (e.g. sharing the same name) which then usually contains an icon name.

This may break for some xwayland applications, applications using CamelCase for the app-id but for their `.desktop` its all lowercase, filesystem namespaces (e.g. flatpack) which don't install their `.desktop` files in the compositor filesystem namespace and various other possible issues.

To sum up: figuring out the correct desktop icon is hard. We \*could\* implement it as we are depending on glib anyway but personally I don't think the complexity outweighs the benefits.

jlindgren90: Hmm, that's unfortunate but makes sense. I was hoping it worked like X11 where the application provides the image data.

README.md
```diff
@@ -38,41 +43,99 @@ on to create a full desktop environment.
...
+matures into production quality. On the whole, we value robustness, reliability
+stability and simplicity over new features. This is because we believe that
+coming up with and implementing new ideas is easy, but making them stick and
+work for a long time hard. We want the latter.
+
+[KDE/kwin] and [GNOME/mutter] are great compositors offering brilliant
+experiences, but at ca 200k and 400k lines-of-code respectively and with heavy
+reliance on their respective eco-systems, they cannot be an elegant solution to
+the need for a frugal, independent, Wayland stacking compositor which is the
+space that labwc occupies.
```
Consolatis, line 67: Non-native english speaker here but to me "occupies" somewhat sounds like we'd be saying that labwc is the only one in that space. Maybe "resides in" would fit better?

I am also not sure if we need to point out the SLOC for kwin and mutter here. Or even need to mention them in the first place. In general I am more of a fan of pointing out our strong points (and weaknesses) than to say why we are better than $xyz. I still like "the boring compositor" btw :p

README.md
01micko: ~hard~ difficult

01micko: ~This is because we believe that
coming up with and implementing new ideas is easy, but making them stick and
work for a long time hard. We want the latter.~

Coming up with new ideas and features is easy. Maintaining and stabilizing them is not. 

_Just my 2 cents to keep it simple._

README.md
01micko: > Non-native english speaker here but to me "occupies" somewhat sounds like we'd be saying that labwc is the only one in that space. Maybe "resides in" or "aims for" would fit better?

Ending a sentence with a [preposition](https://academicguides.waldenu.edu/writingcenter/grammar/prepositions#:~:text=A%20preposition%20is%20a%20word,,%22%20and%20%22to.%22) ("for" or "in" in this case) is poor English.

"occupies" is fine by me. More than one entity can occupy the same space.

01micko: On further thought maybe something like :

_which is the space that labwc strives to occupy._

jlindgren90: >  In general I am more of a fan of pointing out our strong points (and weaknesses) than to say why we are better than $xyz.

This is a good point. Maybe we could just leave out the paragraph that mentions Mutter and KWin. Saying "we believe there is a need for" labwc is enough without mentioning other compositors by name.

Consolatis: We could also spin this a bit more in explaining our own backgrounds, coming from low resource DEs like Mate / XFCE / Openbox and searching for a simple wayland replacement compositor and then finding and enhancing labwc / creating it in the first place (johanmalm).

I am not sure how to properly formulate this but instead of saying "We believe there is a need for labwc because others are bad" we could explain our personal need for it.

Sometimes I suck at putting my thoughts into words but I hope I got my point across.

johanmalm:
Thanks! All suggestions incorporated. It turned into a complete section re-hash, so if you've got time to re-read that would be awesome.

Consolatis:
LGTM.

> Fourthly, we are of the view that a compositor should not be surrounded by hype
and eye-candy, but rather be pretty boring. In this regard we follow in the
footsteps of [metacity] which describes itself as a "Boring window manager for
the adult in you. Many window managers are like Marshmallow Froot Loops;
Metacity is like Cheerios."

Nice metacity reference regarding "boring" there. That was new to me :)

README.md
```diff
@@ -38,41 +43,102 @@ on to create a full desktop environment.
...
+
+Firstly, we believe that there is a need for a simple Wayland window-stacking
+compositor which strikes a balance between minimalism and bloat approximately
+at the level where Window Managers like Openbox reside in the X11 domain.
+This belief has been borne out of the X11 computing background of most of the
+core-devs which is one of low resource Desktop Environments such Mate/XFCE or
+standalone Window Managers like Openbox with no reliance on heavy eco-systems.
+
+Secondly, the Wayland community has achieved an amazing amount so far, and we
+want help to solve the unsolved problems to make Wayland viable for more
```
jlindgren90, line 59: Was this supposed to be "we want to help solve"?

"We want help to solve" could also be correct but changes the meaning a bit (we are looking for someone to help us, rather than trying ourselves to offer help).

README.md
```diff
@@ -38,41 +43,102 @@ on to create a full desktop environment.
...
+session. Nor does it depend on any UI toolkits such as Qt or GTK.
+
+### 1.1 Why?
+
+Firstly, we believe that there is a need for a simple Wayland window-stacking
+compositor which strikes a balance between minimalism and bloat approximately
+at the level where Window Managers like Openbox reside in the X11 domain.
+This belief has been borne out of the X11 computing background of most of the
+core-devs which is one of low resource Desktop Environments such Mate/XFCE or
+standalone Window Managers like Openbox with no reliance on heavy eco-systems.
```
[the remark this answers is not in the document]
Consolatis: Sounds nice.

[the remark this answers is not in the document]
johanmalm: Let's go for this.

README.md
johanmalm: Thanks good spot. Should be "to help" (we want to help...)

PR #662, opened 2022-12-05
menu: Dynamically adjust menu width

Consolatis:
This PR introduces four new theme variables:
- `menu.items.padding.x`
- `menu.items.padding.y`
- `menu.width.min` (menus will never be smaller than this)
- `menu.width.max` (menus will never be wider than this + padding)

Missing:
- [x] parsing theme config, uses a default of 20 / 200
- [x] docs
- [x] rename `scaled_font_buffer_resize` to `scaled_font_buffer_set_max_width`
- [x] squash last commits

~~The way I wired this up is not optimal: When creating a new menu item the width will be compared to the stored width of the whole menu and if the new item width is wider, all existing items will be resized and the new width will be set for the menu.
This means that we may end up unnecessarily resizing items multiple times. However, this only happens while parsing `menu.xml` so it shouldn't be an issue in practice. Feels a bit hackish though.~~

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -17,13 +17,12 @@
 #include "common/mem.h"
 #include "common/nodename.h"
 #include "common/scaled_font_buffer.h"
+#include "common/scene-helpers.h"
 #include "common/string-helpers.h"
 #include "labwc.h"
 #include "menu/menu.h"
 #include "node.h"
 #include "theme.h"
-
-#define MENUWIDTH (110)
 
 /* state-machine variables for processing <item></item> */
 static bool in_item;
@@ -52,7 +51,7 @@ menu_create(struct server *server, const char *id, const char *label)
 	menu->label = xstrdup(label ? label : id);
 	menu->parent = current_menu;
 	menu->server = server;
-	menu->size.width = MENUWIDTH;
+	menu->size.width = server->theme->menu_min_width;
 	/* menu->size.height will be kept up to date by adding items */
 	menu->scene_tree = wlr_scene_tree_create(server->menu_tree);
 	wlr_scene_node_set_enabled(&menu->scene_tree->node, false);
@@ -75,6 +74,59 @@ menu_get_by_id(const char *id)
 	return NULL;
 }
 
+static void
+menu_update_width(struct menu *menu)
+{
+	struct menuitem *item;
+	struct theme *theme = menu->server->theme;
+	int max_width = theme->menu_min_width;
+
+	/* Get widest menu item, clamped by menu_max_width */
+	wl_list_for_each(item, &menu->menuitems, link) {
+		if (item->native_width > max_width) {
+			max_width = item->native_width < theme->menu_max_width
+				? item->native_width : theme->menu_max_width;
+		}
+	}
+	menu->size.width = max_width + 2 * theme->menu_item_padding_x;
+
+	/* Update all items for the new size */
+	wl_list_for_each(item, &menu->menuitems, link) {
+		wlr_scene_rect_set_size(
+			lab_wlr_scene_get_rect(item->normal.background),
+			menu->size.width, item->height);
+
+		if (!item->selected.background) {
+			/* This is a separator. They don't have a selected background. */
+			wlr_scene_rect_set_size(
+				lab_wlr_scene_get_rect(item->normal.text),
+				menu->size.width - 2 * theme->menu_separator_padding_width,
+				theme->menu_separator_line_thickness);
+		} else {
+			/* Usual menu item */
+			wlr_scene_rect_set_size(
+				lab_wlr_scene_get_rect(item->selected.background),
+				menu->size.width, item->height);
+			if (item->native_width > max_width || item->submenu) {
+				scaled_font_buffer_set_max_width(item->normal.buffer,
+					max_width);
+				scaled_font_buffer_set_max_width(item->selected.buffer,
+					max_width);
+			}
+		}
+	}
+}
+
+static void
+post_processing(struct server *server)
+{
+	struct menu *menu;
+	for (int i = 0; i < nr_menus; ++i) {
+		menu = menus + i;
+		menu_update_width(menu);
+	}
+}
+
 static struct menuitem *
 item_create(struct menu *menu, const char *text, bool show_arrow)
 {
@@ -84,6 +136,8 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
 	struct server *server = menu->server;
 	struct theme *theme = server->theme;
 
+	const char *arrow = show_arrow ? "›" : NULL;
+
 	if (!menu->item_height) {
 		menu->item_height = font_height(&rc.font_menuitem)
 			+ 2 * theme->menu_item_padding_y;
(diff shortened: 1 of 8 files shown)


src/menu/menu.c
```diff
@@ -17,13 +17,13 @@
...
+#include "common/scene-helpers.h"
 #include "common/string-helpers.h"
 #include "labwc.h"
 #include "menu/menu.h"
 #include "node.h"
 #include "theme.h"
 
-#define MENUWIDTH (110)
 #define MENU_ITEM_PADDING_Y (4)
 #define MENU_ITEM_PADDING_X (7)
```
Consolatis, line 28: item padding should also be moved to theme

src/theme.c
```diff
@@ -128,6 +128,9 @@ theme_builtin(struct theme *theme)
 	parse_hexstr("#dddad6", theme->menu_items_active_bg_color);
 	parse_hexstr("#000000", theme->menu_items_active_text_color);
 
+	theme->menu_min_width = 20;
+	theme->menu_max_width = 200;
```
Consolatis, line 132: Opinions for default values? I chose 20 for min here because that allows to have a menu consisting of only single emojis as menu entries. Although that likely depends on the font size as well..

src/menu/menu.c
```diff
@@ -86,14 +102,59 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
...
+	if (arrow) {
+		text_width += font_width(&rc.font_menuitem, arrow);
+	}
+	text_width = clamp(text_width, theme->menu_min_width, theme->menu_max_width);
+
+	if (menu->size.width >= text_width + 2 * MENU_ITEM_PADDING_X) {
+		/* Reuse existing menu width */
+		text_width = menu->size.width - 2 * MENU_ITEM_PADDING_X;
+	} else {
+		/* Store new menu width and update existing items */
```
Consolatis, line 124: Guess this could actually be split into its own static function.
Something like `menu_update_width(struct menu *menu, int new_width)`.

src/menu/menu.c
johanmalm: Agree

src/menu/menu.c
```diff
@@ -175,11 +235,12 @@ separator_create(struct menu *menu, const char *label)
 	menuitem->normal.tree = wlr_scene_tree_create(menuitem->tree);
 	menuitem->normal.background = &wlr_scene_rect_create(
 		menuitem->normal.tree,
-		MENUWIDTH, menuitem->height,
+		menu->size.width, menuitem->height,
 		theme->menu_items_bg_color)->node;
 
 	/* theme->menu_separator_width is the line-thickness (so height here) */
-	int width = MENUWIDTH - 2 * theme->menu_separator_padding_width;
+	/* TODO: Rename to theme->menu_separator_thickness. */
```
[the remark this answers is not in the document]
johanmalm: I prefer 'thickness' for line-thickness because it avoid confusion with width/height.
`menu_separator_line_thickness` would be super clear :smile:

src/theme.c
johanmalm: Fine.

johanmalm:
You broke code-check :smile: Proves it works which is good.

> src/menu/menu.c:80: WARNING: Consider removing the code enclosed by this #if 0 and its #endif


Consider `s/scaled_font_buffer_resize/scaled_font_buffer_set_max_width/`

src/menu/menu.c
johanmalm: Are we updating the width of the menu and all its items here on every `item_create()`?
Would it make more sense to calculate and set/update the menu width once on hitting `</menu>`?
I think that would mean calling such a function here: https://github.com/labwc/labwc/blob/aa4ef94af02c9488f0f8e2be25912ee4b358c4ff/src/menu/menu.c#L402

Consolatis: I thought about that but the issue was that to get the max font width of all menu items within a menu we need the actual text (to call `font_width()`). But we only store the `char *` within the `scaled_font_buffer` and its not really intended to access that from the outside. And storing another copy of the `char *` within each item feels redundant.

Maybe `item_create()` could store the calculated font_width within the `struct menuitem` though, that would just be an additional `int` per item. Then the resizing could indeed be done just once when encountering a `</menu>`.

johanmalm: Sorry - the rightful place would probably be here  after traverse()...
https://github.com/labwc/labwc/blob/aa4ef94af02c9488f0f8e2be25912ee4b358c4ff/src/menu/menu.c#L360
Not tested - just reading the code.

src/menu/menu.c
```diff
@@ -77,6 +77,22 @@ menu_get_by_id(const char *id)
...
+		return low;
+	}
+	if (val > high) {
+		return high;
+	}
+	return val;
+}
+#endif
+
+#define clamp(val, low, high) ((val) < (low) ? (low) : (val) > (high) ? (high) : (val))
```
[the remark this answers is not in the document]
Consolatis: `high` could be the same as `low` which would mean its a fixed menu width.
So something like `assert(high >= low)`?

src/menu/menu.c
johanmalm: > I thought about that but the issue was that to get the max font width of all menu items within a menu we need the actual text (to call `font_width()`). But we only store the `char *` within the `scaled_font_buffer` and its not really intended to get accessed from the outside. Storing another copy of the `char *` within each item feels redundant.

Sorry - yes, have just seen your comment in OP :smile:

> Maybe `item_create()` could store the calculated font_width within the `struct menuitem` though, that would just be an additional `int` per item. Then the resizing could indeed be done just once when encountering a `</menu>`.

Good idea. That seems the cleanest.

Consolatis:
Thanks for review.

> Consider `s/scaled_font_buffer_resize/scaled_font_buffer_set_max_width/`

Good call, will do.

src/menu/menu.c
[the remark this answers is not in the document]
johanmalm: I prefer to only use assert for errors caused by the programmer.

What happens if user sets a greater low than high in a themerc.
Maybe it’s better to just check high and low in theme_post_processing()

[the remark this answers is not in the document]
Consolatis: Removed the function completely. Instead, I added a post_processing() hook which just does the clamping internally.
We might stall want to check high >= low in a post_processing step within `theme.c` so leaving this unresolved for now.

src/menu/menu.c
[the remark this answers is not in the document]
Consolatis: Changed to `menu_separator_line_thickness`.

src/menu/menu.c
Consolatis: Did that and did some minor refactoring to only expose a `menu_init()` which then calls `init_rootmenu()` and `init_windowmenu()` internally (which were changed to being `static`). Then added a `post_processing()` step there which triggers the menu width calculations / resizing for all menus.

src/menu/menu.c
```diff
@@ -640,6 +696,14 @@ menu_init_windowmenu(struct server *server)
 	}
 }
 
+void
+menu_init(struct server *server)
+{
+	menu_init_rootmenu(server);
+	menu_init_windowmenu(server);
```
Consolatis, line 703: Should we drop the `menu_` prefix now that these two are `static`?

src/menu/menu.c
Consolatis: Moved. New theme vars:
- `menu.items.padding.x`
- `menu.items.padding.y`

johanmalm:
Nice. LGTM.
Use `post_processing()` in menu/menu.c is an even better solution. Nice and clean.

PR #663, opened 2022-12-05
theme: read ~/.config/labwc/themerc-override

johanmalm:


johanmalm:
Just a quick play.

TODO

- [ ] Docs

PR #666, opened 2022-12-08
Menu tidy up

johanmalm:


docs/themerc
```diff
@@ -41,3 +45,4 @@ osd.bg.color: #dddda6
 osd.border.color: #000000
 osd.border.width: 1
 osd.label.text.color: #000000
+
```
Consolatis, line 48: New empty line. intended?

Consolatis:
Commit message misses the `padding` part in `menu.items.{x,y}`.
Suggest: docs/themerc: add new menu settings because Consolatis didn't pay attention.

Nit: feels a bit strange to have `menu.overlap` on top and then a block for the other menu entries below.

Other than that: LGTM

docs/themerc
johanmalm: Not intended. Removed and force-pushed.

johanmalm:
> Commit message misses the `padding` part in `menu.items.{x,y}`.

Sorted.
 
> Nit: feels a bit strange to have `menu.overlap` on top and then a block for the other menu entries below.

Agree. I've moved `menu.overlap.*` down to the menu section.

johanmalm:
The ordering of the `themerc` was originally in accordance with http://openbox.org/wiki/Help:Themes#Theme_file_structure - but it does seems a bit odd having menu.overlap at the top :smile:

PR #669, opened 2022-12-10
menu: Reuse action argument parsing from rcxml.c

Consolatis:
This fixes being unable to use the `direction` argument in menu entries.

Reported-by: mahk via IRC

### DIFF
--- a/include/action.h
+++ b/include/action.h
@@ -20,6 +20,8 @@ struct action {
 
 struct action *action_create(const char *action_name);
 void action_arg_add_str(struct action *action, char *key, const char *value);
+void action_arg_from_xml_node(struct action *action, char *nodename, char *content);
+
 void actions_run(struct view *activator, struct server *server,
 	struct wl_list *actions, uint32_t resize_edges);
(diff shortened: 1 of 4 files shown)


johanmalm:
Very nice indeed. Great to combined the action-filling code of menu.c and rcxml.c

How would you feel about moving `rcxml_fill_action()` to `action.c` and calling it something like `action_parse_one_entry()` or similar. That way we remove the coupling between `menu.c' and `rcxml.h`

Consolatis:
`action_parse_xml_arg()`?

johanmalm:
> > and calling it something like `action_parse_one_entry()` or similar.
> 
> What about `action_parse_xml_arg()` or `action_parse_config_arg()`?

I like the first one better. How about `action_parse_xml_node()` considering we pass a nodename to it.

Consolatis:
- Changed to `action_arg_from_xml_node()` but can change if you prefer.
- Moved the function to `src/action.c` and updated the commit message.

PR #670, opened 2022-12-10
src/view.c: dont switch output on SnapToEdge if view is maximized

Flrian:
Before this on my setup with a single monitor the following would happen:
1. Snap to Edge
2. Maximize
3. Snapping to the original edge would not work


Consolatis:
Thanks :)

Issue #672, opened 2022-12-12
Hide Alt+Tab Dialog


wesbluemarine:
Is there a way to remove the dialog?

johanmalm:
No. 

I have no issue with adding - maybe as a core.cycleViewOSD

https://github.com/labwc/labwc/blob/e4aa7cef765a81a6911839de16644eceeb2fb5e9/docs/rc.xml.all#L15

Consolatis:
Should be relatively easy to implement (either by just disabling the OSD scene tree or actually preventing the rendering, 2nd option preferred but also a bit more involved as we still want to keep showing window previews / preview outlines).

PR welcome.

PR #674, opened 2022-12-12
osd: factor drawing out of `osd_update`

### LINKED ISSUE #647: Cursor does not update when a window opens or within the closing alt-tab dialog
Specifically:

- If a window opens under the cursor, it does not see a `wl_pointer.frame` event until the pointer is moved. This means that the cursor image may be wrong until the user moves the mouse, which is particularly annoying in cases of screen lockers or video players that normally hide the cursor.

- When alt-tabbing, the cursor does not change to the appropriate one for the window under it if the cursor's position is inside the bounds of the alt-tab dialog when the alt-tab dialog closes. This is easy to observe with a terminal or text editor, as moving the cursor onto the alt-tab dialog sets a "pointer" cursor image, and the "I-beam" cursor image is not restored after alt-tabbing into such a client if the cursor lies inside the bounds of the alt-tab dialog.

These incorrect behaviors have been around for quite some time--I don't think this is a recent regression. It's possible that these issues could be related to the same underlying issue as #646.

bi4k8:
This gives a clean separation of `osd_update` for changing the state of labwc itself and `osd_draw` for putting pixels in the surface buffer. I didn't want to add anything new to `osd_update` (as may be required by #647 and possibly #672) without simplifying it first.

### DIFF
--- a/src/osd.c
+++ b/src/osd.c
@@ -223,6 +223,156 @@ preview_cycled_view(struct view *view)
 	wlr_scene_node_raise_to_top(osd_state->preview_node);
 }
 
+static void
+render_osd(cairo_t *cairo, int w, int h, struct wl_list *node_list,
+		struct view *cycle_view, struct theme *theme, bool show_workspace,
+		const char *workspace_name)
+{
+	struct wlr_scene_node *node;
+	cairo_surface_t *surf = cairo_get_target(cairo);
+
+	/* Draw background */
+	set_cairo_color(cairo, theme->osd_bg_color);
+	cairo_rectangle(cairo, 0, 0, w, h);
+	cairo_fill(cairo);
+
+	/* Draw border */
+	set_cairo_color(cairo, theme->osd_border_color);
+	draw_cairo_border(cairo, w, h, theme->osd_border_width);
+
+	/* Set up text rendering */
+	set_cairo_color(cairo, theme->osd_label_text_color);
+	PangoLayout *layout = pango_cairo_create_layout(cairo);
+	pango_layout_set_width(layout,
+		(OSD_ITEM_WIDTH - 2 * OSD_ITEM_PADDING) * PANGO_SCALE);
+	pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
+
+	PangoFontDescription *desc = font_to_pango_desc(&rc.font_osd);
+	pango_layout_set_font_description(layout, desc);
+
+	PangoTabArray *tabs = pango_tab_array_new_with_positions(2, TRUE,
+		PANGO_TAB_LEFT, OSD_TAB1, PANGO_TAB_LEFT, OSD_TAB2);
+	pango_layout_set_tabs(layout, tabs);
+	pango_tab_array_free(tabs);
+
+	pango_cairo_update_layout(cairo, layout);
+
+	int y = OSD_BORDER_WIDTH;
+
+	/* Center text entries on the y axis */
+	int y_offset = (OSD_ITEM_HEIGHT - font_height(&rc.font_osd)) / 2;
+	y += y_offset;
+
+	/* Draw workspace indicator */
+	if (show_workspace) {
+		/* Center workspace indicator on the x axis */
+		int x = font_width(&rc.font_osd, workspace_name);
+		x = (OSD_ITEM_WIDTH - x) / 2;
+		cairo_move_to(cairo, x, y);
+		PangoWeight weight = pango_font_description_get_weight(desc);
+		pango_font_description_set_weight(desc, PANGO_WEIGHT_BOLD);
+		pango_layout_set_font_description(layout, desc);
+		pango_layout_set_text(layout, workspace_name, -1);
+		pango_cairo_show_layout(cairo, layout);
+		pango_font_description_set_weight(desc, weight);
+		pango_layout_set_font_description(layout, desc);
+		y += OSD_ITEM_HEIGHT;
+	}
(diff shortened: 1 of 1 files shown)


src/osd.c
```diff
@@ -223,6 +223,138 @@ preview_cycled_view(struct view *view)
 	wlr_scene_node_raise_to_top(osd_state->preview_node);
 }
 
+static void
+osd_draw(cairo_t *cairo, int w, int h, struct wl_list *node_list,
+	int current_view_index, struct theme *theme, bool show_workspace,
+	const char *workspace_name)
```
Consolatis, line 229: Suggest changing to `render_osd(struct output *output)` and moving everything rendering related from `osd_update()` to the top / bottom of this new function. The `struct server *server` pointer can be retrieved via `output->server`.

src/osd.c
```diff
@@ -223,6 +223,138 @@ preview_cycled_view(struct view *view)
...
+	set_cairo_color(cairo, theme->osd_bg_color);
+	cairo_rectangle(cairo, 0, 0, w, h);
+	cairo_fill(cairo);
+
+	/* Draw border */
+	set_cairo_color(cairo, theme->osd_border_color);
+	draw_cairo_border(cairo, w, h, theme->osd_border_width);
+
+	/* Highlight current window */
+	if (current_view_index != -1) {
```
Consolatis, line 244: This can be removed completely and replaced by a simpler check further down below. See next comment.

src/osd.c
```diff
@@ -223,6 +223,138 @@ preview_cycled_view(struct view *view)
...
+#endif
+		}
+
+		if (is_title_different(view)) {
+			buf_add(&buf, view_get_string_prop(view, "title"));
+		}
+
+		pango_layout_set_text(layout, buf.buf, -1);
+		pango_cairo_show_layout(cairo, layout);
+		y += OSD_ITEM_HEIGHT;
```
Consolatis, line 331: between `pango_cairo_show_layout()` and `y += OSD_ITEM_HEIGHT` this can be added:
```c
		if (view == server->osd_state.cycle_view) {
			/* Highlight current window */
			cairo_rectangle(cairo, OSD_BORDER_WIDTH, y - y_offset,
				OSD_ITEM_WIDTH, OSD_ITEM_HEIGHT);
			cairo_stroke(cairo);
		}
```

src/osd.c
```diff
@@ -223,6 +223,138 @@ preview_cycled_view(struct view *view)
...
+	}
+	free(buf.buf);
+	g_object_unref(layout);
+
+	cairo_surface_flush(surf);
+}
+
+/* returns index of node with the given view, or -1 if not found */
+static int
+find_view_index(struct wl_list *node_list, struct view *view_to_find)
```
Consolatis, line 341: This function can be removed, see previous comment

src/osd.c
```diff
@@ -223,6 +223,138 @@ preview_cycled_view(struct view *view)
...
+		if (view == view_to_find) {
+			return i;
+		}
+		i += 1;
+	}
+	return -1;
+}
+
 void
 osd_update(struct server *server)
```
Consolatis, line 359: This function can be cleared up completely by moving the parts relevant to the OSD to the new function.
All that is left then looks something like this:
```c
void
osd_update(struct server *server)
{
	if (wl_list_empty(&server->workspace_current->tree->children)
			|| !server->osd_state.cycle_view) {
		osd_finish(server);
		return;
	}

	/* Render the actual OSD */
	struct output *output;
		destroy_osd_nodes(output);
		if (!output->wlr_output->enabled) {
			continue;
		}
		render_osd(output);
	}

	if (rc.cycle_preview_outlines) {
		osd_update_preview_outlines(server->osd_state.cycle_view);
	}

	if (rc.cycle_preview_contents) {
		preview_cycled_view(server->osd_state.cycle_view);
	}
}
```

The whole output loop can then be further enclosed in a new condition like `if (rc.cycle_osd) { .. }` (in a separate commit / PR).

Consolatis (changes requested):
Thanks for working on this. A few comments below.

I'll change the PR to Draft for now to prevent accidental merging, feel free to change it back whenever you feel this is ready.

src/osd.c
bi4k8: I've done something along these lines, but still have a separation between `display_osd` (which does the work to get OSD onscreen) and `render_osd` (which does not rely on the rest of the compositor for the most part, and simply fills in pixels, because there's plenty of complexity in just that). See how you like it.

src/osd.c
```diff
@@ -223,173 +223,183 @@ preview_cycled_view(struct view *view)
 	wlr_scene_node_raise_to_top(osd_state->preview_node);
 }
 
-void
-osd_update(struct server *server)
+static void
+render_osd(cairo_t *cairo, int w, int h, struct wl_list *node_list,
+	struct view *cycle_view, struct theme *theme, bool show_workspace,
+	const char *workspace_name)
```
Consolatis, line 229: Style nit: In the rest of the codebase we are using two tab indentation for continued argument lines. E.g.
```c
static void
render_osd(cairo_t *cairo, int w, int h, struct wl_list *node_list,
		struct view *cycle_view, struct theme *theme, bool show_workspace,
		const char *workspace_name)
{
```

Consolatis:
Other than that minor style nit this looks very good to me, thanks :)

src/osd.c
bi4k8: Will fix.

Consolatis (approved):

PR #676, opened 2022-12-14
Update cursor on next commit after xdg surface map

bi4k8:
Fixes the "new window opening" portion of #647.

I can put this in the same PR as #675 if desired--I just didn't know how long it would be until I found a decent fix, and the two really are, despite looking similar, pretty much independent bugs+fixes.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -46,6 +46,7 @@ struct view {
 
 	bool mapped;
 	bool been_mapped;
+	bool committed_since_mapped;
 	bool ssd_enabled;
 	bool minimized;
 	bool maximized;
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -66,7 +66,8 @@ handle_commit(struct wl_listener *listener, void *data)
 	struct wlr_box size;
 	wlr_xdg_surface_get_geometry(xdg_surface, &size);
 
-	bool update_required = false;
+	bool update_required = !view->committed_since_mapped;
+	view->committed_since_mapped = true;
 
 	if (view->w != size.width || view->h != size.height) {
 		update_required = true;
@@ -350,6 +351,7 @@ xdg_toplevel_view_map(struct view *view)
 		view_moved(view);
 		view->been_mapped = true;
 	}
+	view->committed_since_mapped = false;
 
 	view->commit.notify = handle_commit;
 	wl_signal_add(&xdg_surface->surface->events.commit, &view->commit);


Consolatis:
Not much time right now, so just two short questions:

Do we need to reset the bool on unmap?
What about xwayland windows, do we need the same fix there?

jlindgren90:
I don't have time to investigate more right now but I'd also like to understand a bit deeper what's going on here:

1. Why does the existing view_moved() call in map() not work?
2. As Consolatis wondered, are we setting view->x/y too early and should we delay that until commit()?

johanmalm:
Thanks @bi4k8 Nice detective work.
I think you've got us pretty close, but it feels like there is a simpler solution without the additional variable.
Suggest we wait until one of us has time to explore further.

bi4k8:
Thanks for the careful review.

> Nice catch.
> 
> Not much time right now, so just two short questions:
> 
> Do we need to reset the bool on unmap? What about xwayland windows, do we need the same fix there?
> 
> Edit: So what might have caused this regressions is that at some point we started to set `view->w` / `view->h` to the desired values before the commit actually happened. Otherwise the check below would have detected the difference and would have set `update_required` to `true`.
> 
> Edit 2: I am not sure if this band-aid is the proper solution or if we should instead prevent setting `view->w` and `view->h` too early. This might have had something to do with starting windows in maximized mode, but I may be wrong on this. Thoughts @jlindgren90?

I believe you're right. 3034da776bdba9ce60f7356ea01239f372d611bf introduced the early writes to `view->w` and `view->h`, and removing these seems to fix the behavior. Unconditionally calling `view_moved` on commit (effectively reverting bc4ce1bd5d2f82b1727086a938e29686a96001bc) also works, but we shouldn't need to remove the optimization.

Because these early writes were only used in order to be able to call `view_center`, we can simply center the view without writing to `view->w` and `view->h` and solve this bug more cleanly without additional state. I'll file a PR doing that, superseding this PR.

Consolatis:
Really nice detective work. Again. :)

Just had a quick look at your new PR and so far I think its very sensible. I'd like investigate this a bit further and run some tests to see if the missing `view->w` and `view->h` values before the first commit change anything regarding applications starting in maximized / fullscreen mode, will comment on that PR in the coming days.

> Unconditionally calling view_moved on commit (effectively reverting https://github.com/labwc/labwc/commit/bc4ce1bd5d2f82b1727086a938e29686a96001bc) also works, but we shouldn't need to remove the optimization.

That would be really bad, that commit handler is triggered every single time an application commits its new buffer contents AFAIR, e.g. all the time. We'd do non-stop hitbox detection to figure out if we'd need to change the cursor in that case.

PR #680, opened 2022-12-17
Avoid setting `view->w` and `view->h` before xdg view commit

bi4k8:
A cleaner fix for the second half of #647. Supersedes #676.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -159,6 +159,9 @@ void view_reload_ssd(struct view *view);
 void view_impl_map(struct view *view);
 void view_adjust_size(struct view *view, int *w, int *h);
 
+bool view_compute_centered_position(struct view *view, int w, int h,
+	int *x, int *y);
+
 void view_on_output_destroy(struct view *view);
 void view_destroy(struct view *view);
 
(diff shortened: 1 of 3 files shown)


jlindgren90:
Thanks! Looks reasonable to me. I wonder if adding some kind of a comment to the code would be good so we don't inadvertantly introduce a similar bug again in future?

Consolatis:
Looked a bit more into this and had a quick test nested: everything seems fine (only tested with `foot -m` / `foot -F` though).
Applications starting maximized or fullscreen (and thus may not have the original geometry set) are dealt with by setting the hardcoded fallback geometry:
https://github.com/labwc/labwc/blob/3dec9cf40cebe0e231e42722caa3c8fbdc2d5f83/src/view.c#L323-L340

> I wonder if adding some kind of a comment to the code would be good so we don't inadvertently introduce a similar bug again in future?

Agree, a comment to explain why we don't set them manually on map would be appreciated, (e.g. just a copy the first part of the commit message should be enough).

bi4k8:
Comment added.

Consolatis:
Merging. Thanks again for the extensive investigation and the two fixes regarding OSD and newly mapped windows.

PR #682, opened 2022-12-19
Unbreak build without Xwayland (in wlroots)

jbeich:
Found downstream when building everything with X11 disabled (not just labwc). See also [error log](https://github.com/labwc/labwc/files/10258590/labwc-0.6.0.log).

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2,12 +2,15 @@
 #include <assert.h>
 #include <stdio.h>
 #include <strings.h>
-#include <xcb/xcb_icccm.h>
 #include "common/scene-helpers.h"
 #include "labwc.h"
 #include "ssd.h"
 #include "menu/menu.h"
 #include "workspaces.h"
+
+#if HAVE_XWAYLAND
+#include <xcb/xcb_icccm.h>
+#endif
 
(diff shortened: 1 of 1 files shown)


jbeich:
Alternatively, the header can be dropped as wlr/xwayland/xwayland.h (from wlroots) already includes xcb/xcb_icccm.h

jbeich:
Nevermind, if sway doesn't rely on header bootlegging then labwc probably shouldn't as well.

jbeich:
Moved the include after `"labwc.h"` to avoid the following:
```c
../src/view.c:5:5: error: "HAVE_XWAYLAND" is not defined, evaluates to 0 [-Werror=undef]
    5 | #if HAVE_XWAYLAND
      |     ^~~~~~~~~~~~~
```

Consolatis:
D'oh. Thanks :)

Consolatis:
This makes me wonder if we should refactor things a little so that everything that requires external X11 headers is confined to `xwayland.c` / `xwayland-unmanaged.c` and then exposed as functions to the rest of the codebase, guarded by the usual define.

PR #684, opened 2022-12-20
subprojects/wlroots.git: use commit fe7e7a14 on 0.16 branch

johanmalm:
...in order to use recent fixes including setting exclusive zone correctly when only one edge is anchored (wlroots@48920828)


Consolatis:
LGTM

Issue #667, opened 2022-12-09
No enter event sent when a new layer surface appears under the pointer


Tamaranch:
Labwc 0.6.0 (this was already the case with 0.5.3)

The event is only sent if the pointer is moved manually after the window has appeared. This is a problem when using functions like [`gdk_device_get_window_at_position()`](https://developer-old.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-get-device-position), which return NULL even though an app window is under the pointer.

Consolatis:
Very likely the same underlying issue as
- #647

In short:  We should send an enter event but for some reason the wlroots scene graph doesn't seem to detect the new surface right away in the `map()` handler.
This needs further debugging to figure out why this is happening again and how to fix it.

Closing this as duplicate.

Consolatis:
Should be fixed in latest master branch, `WAYLAND_DEBUG=1 foot 2>&1 | grep pointer` now shows a `wl_pointer@19.enter()` event without moving the cursor.

Tamaranch:
It does fix the problem for a normal window, but not for a layer-shell window (which was my initial test case).

Consolatis:
Thanks for clarifying.

johanmalm:
I’ll try to do some patches for this tomorrow.

johanmalm:
@Tamaranch Any chance of testing with PR #685 ? Or give me some pointers for testing.

I've tried the code below, but can't make sense of it.

```
#include <stdio.h>
#include <stdlib.h>
#include <gtk/gtk.h>

static void update(GtkWidget *widget, gpointer data)
{
	usleep(1000000);
	GdkDevice *device = gdk_seat_get_pointer(gdk_display_get_default_seat(gdk_display_get_default()));
	GdkWindow *window = gdk_device_get_window_at_position(device, NULL, NULL);
	fprintf(stderr, "window=%p\n", window);
}

static void activate(GtkApplication *app, gpointer user_data)
{
	GtkWidget *window = gtk_application_window_new(app);
	gtk_window_set_title(GTK_WINDOW(window), "test");
	gtk_window_set_default_size(GTK_WINDOW(window), 320, 240);

	GtkWidget *widget = gtk_button_new_with_label("Update");
	g_signal_connect(widget, "clicked", G_CALLBACK(update), NULL);
	gtk_container_add(GTK_CONTAINER(window), widget);

	gtk_widget_show_all(window);
}


int main(int argc, char **argv)
{
[...]

PR #685, opened 2022-12-20
[wip] tidy up layer.c

johanmalm:
--edited--
First commit sends pointer-enter-event when layer-surface is under cursor.
Fixes issue #667

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -138,6 +138,15 @@ surface_commit_notify(struct wl_listener *listener, void *data)
 			output_from_wlr_output(layer->server, wlr_output);
 		layers_arrange(output);
 	}
+
+	/*
+	 * We need to call cursor_update_focus() to get a pointer-enter event
+	 * when a new layer surface appears under the cursor.  This enables
+	 * gdk_device_get_window_at_position() to work without moving pointer.
+	 * See issue #667
+	 */
+	struct output *output = (struct output *)layer_surface->output->data;
+	cursor_update_focus(output->server);
 }
 
 static void
@@ -182,9 +191,12 @@ unmap_notify(struct wl_listener *listener, void *data)
 static void
 map_notify(struct wl_listener *listener, void *data)
 {
-	return;
-	struct wlr_layer_surface_v1 *layer_surface = data;
-	wlr_surface_send_enter(layer_surface->surface, layer_surface->output);
+	/*
+	 * Since moving to the wlroots scene-graph API, there is no need to
+	 * call wlr_surface_send_enter() from here since that will be done
+	 * automatically based on the position of the surface and outputs in
+	 * the scene. See wlr_scene_surface_create() documentation.
+	 */
 }
 
 static void
--- a/src/layers.c
+++ b/src/layers.c
@@ -131,11 +131,10 @@ surface_commit_notify(struct wl_listener *listener, void *data)
 		return;
 	}
 
+	struct output *output = (struct output *)layer_surface->output->data;
 	if (layer_surface->current.committed
 			|| layer->mapped != layer_surface->mapped) {
 		layer->mapped = layer_surface->mapped;
-		struct output *output =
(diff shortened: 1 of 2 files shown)


Consolatis:
I am not sure if the layer scene API already sends that output enter event:
```c
 /**
 * Add a node displaying a single surface to the scene-graph.
 *
 * The child sub-surfaces are ignored.
 *
 * wlr_surface_send_enter() and wlr_surface_send_leave() will be called
 * automatically based on the position of the surface and outputs in
 * the scene.
 */
struct wlr_scene_surface *wlr_scene_surface_create(struct wlr_scene_tree *parent,
	struct wlr_surface *surface);
```

johanmalm:
Tested with https://github.com/labwc/labwc/issues/667#issuecomment-1360457763 I now get a pointer-enter-event.
There is more refactoring to be done, I just want to get this bit right first.

Consolatis:
> Tested with https://github.com/labwc/labwc/issues/667#issuecomment-1360457763 I now get a pointer-enter-event.

Strange, I do not. How I tested: positioning the cursor on bottom center (where the layer surface will end up) and then starting the test application I only see the `wl_surface@xx.enter()`. For the `wl_pointer@xx.enter()` event I have to move the cursor.

Tamaranch:
It doesn't work for me either, even with the third commit.

src/layers.c
```diff
@@ -131,22 +145,36 @@ surface_commit_notify(struct wl_listener *listener, void *data)
...
 		layers_arrange(output);
 	}
+
+	/*
+	 * We need to call cursor_update_focus() to get a pointer-enter event
+	 * when a new layer surface appears under the cursor.  This enables
+	 * gdk_device_get_window_at_position() to work without moving pointer.
+	 * See issue #667
+	 */
+	cursor_update_focus(output->server);
```
Consolatis, line 161: This will get called every time the client updates the buffer. Maybe it could be moved within the above condition so its only called on the first map?

src/layers.c
```diff
@@ -131,22 +145,36 @@ surface_commit_notify(struct wl_listener *listener, void *data)
...
+	struct wlr_layer_surface_v1 *layer_surface =
+		layer->scene_layer_surface->layer_surface;
+
+	if (seat->focused_layer == layer_surface) {
 		seat_set_focus_layer(seat, NULL);
 	}
+	if (seat->pressed.surface == layer_surface->surface) {
+		seat->pressed.node = NULL;
+		seat->pressed.surface = NULL;
+	}
```
Consolatis, line 177: We have a `seat_reset_pressed(struct seat *seat)` now, that one also handles removing the additional destroy handler set up there.

johanmalm:
[recording.webm](https://user-images.githubusercontent.com/6955353/208995560-9cbd1d22-63f2-4719-a25b-ec5be829fe01.webm)

Evidence by demo :smile:
I'm not sure what's going on then. I tried putting the cursor_update_focus() in the map handler to start with and then `layers_arrange()`, but could only get it to work in the commit-handler.
I agree though - we don't really want to update on every commit - but let's get it working for your two first.

src/layers.c
johanmalm: Sorry - forgot. Okay will drop this commit.

Consolatis: I think the commit is very valid, would just need to call the `seat_reset_pressed()` wrapper instead of manually resets.

Consolatis:
Hm.. that sounds like
a) a race condition
b) the same issue we just (re)solved for xdg windows (`cursor_update_focus()` was called too early, before the scene-graph had updated its internal state so the new window was simply not found during hitbox detection).

For the XDG case we solved that by comparing the actual surface size (and position) with the cached size in the commit handler and if they differ we call `view_moved()` which internally calls `cursor_update_focus()`.
We might need to do something similar here.

johanmalm:
Superseded by #697

Issue #681, opened 2022-12-18
Cursor initial size/position wrong


bi4k8:
My `~/.config/labwc/autostart` executes `kanshi`, which sets output scale to 2. When I start labwc on a TTY, the cursor shows up near the top-left of the screen, but is half the size it should be (i.e., it is rendered at scale 1). When I move the mouse, the cursor jumps to a position twice as far down and right as it was, and renders at the expected (2x) size.

I believe there are two bugs here, as if I change the call to `wlr_cursor_move(server->seat.cursor, NULL, 0, 0);` to pass any nonzero number for either argument, then the cursor no longer jumps (but is still the wrong size until moved).

bi4k8:
Furthermore, if I comment out this early-return from `wlr_output_cursor_move` in wlroots' `src/output/cursor.c`, the cursor also no longer jumps (but is still the wrong size until moved):
```C
	if (cursor->x == x && cursor->y == y) {
		return true;
	}
```

So I think this part of the bug is a case of wlroots being over-eager in this early-return optimization.

bi4k8:
Hmm, it seems like the cursor size part of this bug may also have some (possibly confounding) interaction with swaybg, as when I run labwc with no autostart programs (whereas prior I was running swaybg) and later manually execute kanshi or wlr-randr to change output scale, the cursor never changes size except when over a client, and reverts to scale=1 size whenever over the (blank) desktop.

Consolatis:
hm.. The cursors are loaded with a scale argument via `wlr_xcursor_manager_load()` in two places:
- `src/cursor.c` `cursor_init()` (initialization, hardcoded to `1`)
- `src/output.c` `handle_output_manager_apply()` (output config change, one call per output, using that output's scale)

> the cursor never changes size except when over a client

In wayland, clients always render their own cursor so this is somewhat expected.

> and reverts to scale=1 size whenever over the (blank) desktop

When the cursor is over the desktop (or SSD or some labwc menu) its labwc that sets the cursor, so that sounds like we are somehow missing to update the cursor scale. AFAIR wlroots caches the cursor images for different scales, we might miss something to tell the cursor manager to use the new scale.

`WLR_NO_HARDWARE_CURSORS=1` might also play a role in this.

bi4k8:
[...]

PR #686, opened 2022-12-20
Ensure cursor is correctly scaled at startup

bi4k8:
Fixes (the size portion of) #681. The other half has an upstream fix as noted in the issue thread.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -335,6 +335,9 @@ handle_output_manager_apply(struct wl_listener *listener, void *data)
 		wlr_xcursor_manager_load(server->seat.xcursor_manager,
 			output->wlr_output->scale);
 	}
+
+	/* Re-set cursor image in case scale changed */
+	cursor_update_focus(server);
 }
 
 /*


Consolatis:
Nice one, does this also fix the runtime scale change via wlr-randr?

bi4k8:
I'll have to do some more testing to see exactly which cases (AFAIK, just ones around compositor surfaces like the root and menus) are showing the wrong cursor sizes even after this patch.

johanmalm:
I've tried with/without this patch, but don't see a difference (run-time change or otherwise)

bi4k8:
> I've tried with/without this patch, but don't see a difference (run-time change or otherwise)

To reproduce the behavior that this patch fixes:
```
$ mkdir -p /tmp/labwc-homedir/.config/labwc/
$ > /tmp/labwc-homedir/.config/labwc/autostart (echo '#!/bin/sh'; echo "kanshi -c <(echo -e 'profile { output * scale 2\\\\n}')")
$ XCURSOR_THEME=Adwaita HOME=/tmp/labwc-homedir labwc
```

The cursor will be small, and will immediately jump to 2x scale when first moved.

This patch fixes this scale jump on the first movement of the cursor.

The runtime scale change problem seems to be triggered when no valid cursor theme is found; see [this comment](https://github.com/labwc/labwc/issues/681#issuecomment-1369228967).

Consolatis:
I assume the actual issue here is
https://github.com/labwc/labwc/blob/8bde7bcee0deac8b5d6ba51adaa4f524091bde1d/src/cursor.c#L237-L240
which prevents setting the same cursor that is already set.

This does solve the image not being updated to the new scale on the first scale change for me (nested):
```diff
diff --git a/include/cursor.h b/include/cursor.h
index 757a14d..0f42cfb 100644
--- a/include/cursor.h
+++ b/include/cursor.h
@@ -106,6 +106,16 @@ enum lab_cursors cursor_get_from_edge(uint32_t resize_edges);
  */
 void cursor_update_focus(struct server *server);
 
+/**
+ * cursor_update_image - re-set the labwc cursor image
+ * @seat - seat
+ *
+ * This can be used to update the cursor image on output scale changes.
+ * If the current cursor image was not set by labwc but some client
+ * this is a no-op.
+ */
+void cursor_update_image(struct seat *seat);
+
 void cursor_init(struct seat *seat);
 void cursor_finish(struct seat *seat);
 
diff --git a/src/cursor.c b/src/cursor.c
index 3cfc6b1..c6f2fc4 100644
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -244,6 +244,17 @@ cursor_set(struct seat *seat, enum lab_cursors cursor)
 	seat->server_cursor = cursor;
 }
 
+void
+cursor_update_image(struct seat *seat)
+{
+	enum lab_cursors cursor = seat->server_cursor;
+	if (cursor == LAB_CURSOR_CLIENT) {
+		return;
+	}
+	wlr_xcursor_manager_set_cursor_image(
+		seat->xcursor_manager, cursor_names[cursor], seat->cursor);
+}
+
 bool
 input_inhibit_blocks_surface(struct seat *seat, struct wl_resource *resource)
 {
diff --git a/src/output.c b/src/output.c
index da4bba7..0eb748f 100644
--- a/src/output.c
+++ b/src/output.c
@@ -240,13 +240,9 @@ output_update_for_layout_change(struct server *server)
 	 * "Move" each wlr_output_cursor (in per-output coordinates) to
 	 * align with the seat cursor.  Set a default cursor image so
 	 * that the cursor isn't invisible on new outputs.
-	 *
-	 * TODO: remember the most recent cursor image (see cursor.c)
-	 * and set that rather than XCURSOR_DEFAULT
 	 */
 	wlr_cursor_move(server->seat.cursor, NULL, 0, 0);
-	wlr_xcursor_manager_set_cursor_image(server->seat.xcursor_manager,
-		XCURSOR_DEFAULT, server->seat.cursor);
+	cursor_update_image(&server->seat);
 }
 
 static void
@@ -336,7 +332,7 @@ handle_output_manager_apply(struct wl_listener *listener, void *data)
 	}
 
 	/* Re-set cursor image in case scale changed */
-	cursor_update_focus(server);
+	cursor_update_image(&server->seat);
 }
 
 /*
```

bi4k8:
@Consolatis Your patch (applied atop this PR) actually causes the cursor to be completely invisible until moved with the reproduction steps [here](https://github.com/labwc/labwc/pull/686#issuecomment-1369229245).

Consolatis:
If you re-add the `cursor_update_focus(server)` in `handle_output_manager_apply()` above the new `cursor_update_image()` (e.g. restore your original patch), does that fix the cursor image in all cases (including runtime changes via `wlr-randr --output whatever --scale=2`)?

bi4k8:
> If you re-add the `cursor_update_focus(server)` in `handle_output_manager_apply()` above the new `cursor_update_image()` (e.g. restore your original patch), does that fix the cursor image in all cases (including runtime changes via `wlr-randr --output whatever --scale=2`)?

Yes, it does.

Consolatis:
Alright, lets merge this PR then. I'll follow up with a new one that should handle the runtime change for the cursor image.

PR #696, opened 2022-12-22
rcxml: Make sure a default libinput category always exists

jlindgren90:
Previously, if `rc.xml` defined only non-default libinput categories, no default category was created. This meant that `configure_libinput()` might totally skip configuring some devices even with default settings, like tap-to-click.

Fix this by making sure that a default category is always created.


Consolatis:
Code itself LGTM, not tested.

I can't really say anything about the logic changes, never really worked on that part of the codebase in depth (and neither with libinput itself).

johanmalm:
Thanks.

PR #697, opened 2022-12-22
Use scene-node approach for layers and simplify usage of wlr_scene_node_at()

johanmalm:
Related to issue #667

This will supersede PR #685 

TODO

- [x] Use nodes for popups too
- [x] In commit-handler deal with changed layer
- [x] Ensure pointer-enter-event is sent when new layer-surface appears under pointer (without the pointer moving)
- [x] Rename all handlers to `handle_*` rather than `*_notify`
- [x] Remove redundant `struct wlr_box geo` in `layers.h` 

Later
- Use ARRAY_SIZE macro

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -310,6 +310,21 @@ get_surface_from_layer_node(struct wlr_scene_node *node)
 	return NULL;
 }
 
+static bool
+is_layer_descendant(struct wlr_scene_node *node)
+{
+	goto start;
+	while (node) {
+		struct node_descriptor *desc = node->data;
+		if (desc && desc->type == LAB_NODE_DESC_LAYER_SURFACE) {
+			return true;
+		}
+start:
+		node = node->parent ? &node->parent->node : NULL;
+	}
+	return false;
+}
+
 /* TODO: make this less big and scary */
 struct cursor_context
 get_cursor_context(struct server *server)
@@ -389,6 +404,27 @@ get_cursor_context(struct server *server)
 				break;
 			}
 		}
+
+		/* Edge-case nodes without node-descriptors */
+		if (node->type == WLR_SCENE_NODE_BUFFER) {
+			struct wlr_surface *surface = lab_wlr_surface_from_node(node);
+			if (surface) {
+				if (is_layer_descendant(node)) {
+					/*
+					 * layer-shell subsurfaces need to be
+					 * able to receive pointer actions.
+					 *
+					 * Test by running
+					 * `gtk-layer-demo -k exclusive`, then
+					 * open the 'set margin' dialog and try
+					 * setting the margin with the pointer.
+					 */
+					ret.surface = surface;
+					return ret;
+				}
+			}
+		}
+
 		/* node->parent is always a *wlr_scene_tree */
 		node = node->parent ? &node->parent->node : NULL;
 	}
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -5,6 +5,7 @@
 #include "common/scene-helpers.h"
 #include "dnd.h"
 #include "labwc.h"
+#include "layers.h"
 #include "node.h"
 #include "ssd.h"
 #include "view.h"
@@ -292,6 +293,23 @@ desktop_focus_topmost_mapped_view(struct server *server)
 	desktop_move_to_front(view);
 }
 
+static struct wlr_surface *
+get_surface_from_layer_node(struct wlr_scene_node *node)
+{
+	assert(node->data);
+	struct node_descriptor *desc = (struct node_descriptor *)node->data;
+	if (desc->type == LAB_NODE_DESC_LAYER_SURFACE) {
+		struct lab_layer_surface *surface;
+		surface = node_layer_surface_from_node(node);
+		return surface->scene_layer_surface->layer_surface->surface;
+	} else if (desc->type == LAB_NODE_DESC_LAYER_POPUP) {
+		struct lab_layer_popup *popup;
+		popup = node_layer_popup_from_node(node);
+		return popup->wlr_popup->base->surface;
+	}
+	return NULL;
+}
+
 /* TODO: make this less big and scary */
 struct cursor_context
 get_cursor_context(struct server *server)
@@ -317,21 +335,16 @@ get_cursor_context(struct server *server)
 		ret.type = LAB_SSD_ROOT;
 		return ret;
 	}
+#if HAVE_XWAYLAND
 	if (node->type == WLR_SCENE_NODE_BUFFER) {
 		struct wlr_surface *surface = lab_wlr_surface_from_node(node);
-		if (surface && wlr_surface_is_layer_surface(surface)) {
-			ret.type = LAB_SSD_LAYER_SURFACE;
-			ret.surface = surface;
-			return ret;
-		}
-#if HAVE_XWAYLAND
 		if (node->parent == server->unmanaged_tree) {
 			ret.type = LAB_SSD_UNMANAGED;
 			ret.surface = surface;
 			return ret;
 		}
-#endif
 	}
+#endif
 	while (node) {
 		struct node_descriptor *desc = node->data;
 		if (desc) {
@@ -357,9 +370,14 @@ get_cursor_context(struct server *server)
 				return ret;
 			}
 			case LAB_NODE_DESC_LAYER_SURFACE:
+				ret.node = node;
+				ret.type = LAB_SSD_LAYER_SURFACE;
+				ret.surface = get_surface_from_layer_node(node);
+				return ret;
 			case LAB_NODE_DESC_LAYER_POPUP:
+				ret.node = node;
 				ret.type = LAB_SSD_CLIENT;
-				ret.surface = lab_wlr_surface_from_node(ret.node);
+				ret.surface = get_surface_from_layer_node(node);
 				return ret;
 			case LAB_NODE_DESC_MENUITEM:
 				/* Always return the top scene node for menu items */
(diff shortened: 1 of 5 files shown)


src/desktop.c
```diff
@@ -317,14 +325,9 @@ get_cursor_context(struct server *server)
 		ret.type = LAB_SSD_ROOT;
 		return ret;
 	}
+#if HAVE_XWAYLAND
```
Consolatis, line 328: This was guarding the unmanaged tree access which doesn't exist for non-xwayland compiles

src/layers.c
```diff
@@ -29,44 +45,9 @@ layers_arrange(struct output *output)
...
-
 	int nr_layers = sizeof(output->layers) / sizeof(output->layers[0]);
 	for (int i = 0; i < nr_layers; i++) {
-		struct lab_layer_surface *lab_layer_surface;
-
-		/*
-		 * First we go over the list of surfaces that have
-		 * exclusive_zone set (e.g. statusbars) because we have to
-		 * determine the usable area before processing regular layouts.
-		 */
```
Consolatis, line 48: AFAIR the comment was about an important fix. `arrange_one_layer()` doesn't seem to respect this anymore.

src/layers.c
```diff
@@ -29,44 +45,9 @@ layers_arrange(struct output *output)
...
-			struct wlr_scene_layer_surface_v1 *scene_layer_surface =
-				lab_layer_surface->scene_layer_surface;
-			if (!scene_layer_surface->layer_surface->current.exclusive_zone) {
-				wlr_scene_layer_surface_v1_configure(
-					scene_layer_surface, &full_area, &usable_area);
-			}
-		}
-
-		wlr_scene_node_set_position(&output->layer_tree[i]->node,
-			scene_output->x, scene_output->y);
```
Consolatis, line 69: I think we still need to adjust the whole layer tree here to react to output resolution changes

Consolatis (commented):
Not yet tested, just some comments of things that stood out to me when reading the diff

src/layers.c
johanmalm: `wlroots` does the magic now. Have tested - works fine.

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/layer_shell_v1.c#L133
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/layer_shell_v1.c#L86

src/layers.c
johanmalm: Oh. Good spot. Shouldn't have taken that line out.

src/desktop.c
johanmalm: Separate fix?

src/layers.c
johanmalm: I've put `wlr_scene_node_set_position()` back :smile:

src/layers.c
Consolatis: I am not 100% certain right now but I think the issue this was fixing was:
- docking a layershell client to the top, no exlcusive zone
- docking another layershell client to the top, exclusive zone

The result before the fix was that the exclusive zone client was overlapping the 1st non-exlcusive-zone client.
By first adjusting the usable area by all clients which use the exclusive zone and then all the other ones the 1st layershell client would then be positioned below the 2nd one using the exlcusive zone.

src/layers.c
Consolatis: Might deserve a comment about resolution changes while at it :)

src/desktop.c
johanmalm: Done

johanmalm:
I've pushed subsequent changes (since first commit 1 week ago) in separate commits. It's hopefully easier to read that way.
I just changed the order somewhat which is why I had to force-push.

src/desktop.c
```diff
@@ -292,6 +293,22 @@ desktop_focus_topmost_mapped_view(struct server *server)
 	desktop_move_to_front(view);
 }
 
+static struct wlr_surface *
+get_surface_from_layer_node(struct wlr_scene_node *node)
+{
```
Consolatis, line 298: Do we want to add an `assert(node->data)` here?

src/layers.c
```diff
@@ -39,34 +55,13 @@ layers_arrange(struct output *output)
...
-				wlr_scene_layer_surface_v1_configure(
-					scene_layer_surface, &full_area, &usable_area);
-			}
-		}
+		struct wlr_scene_tree *layer = output->layer_tree[i];
+		arrange_one_layer(output, &full_area, &usable_area, layer);
 
-		wlr_scene_node_set_position(&output->layer_tree[i]->node,
-			scene_output->x, scene_output->y);
+		/* TODO: Is this really needed? */
```
Consolatis, line 61: Guess the `TODO` can be removed now, it is needed in case the output position in the server layout changes.
`wlr_scene_node_set_position()` has an internal check which compares the old values vs the new ones and early returns in case they are the same so this is pretty cheap.

The alternative would be to listen to layout changes here as well and then act accordingly.

Consolatis:
> I've pushed subsequent changes (since first commit 1 week ago) in separate commits. It's hopefully easier to read that way.
I just changed the order somewhat which is why I had to force-push.

I am usually using the "Files changed" view on top thus I look into all the combined changes and ignore single commits. So for me personally it really doesn't matter how the WIP commits are structured. Feel free to use whatever is easiest for you.

src/layers.c
johanmalm: You're right. Have implemented are you outlined.

src/layers.c
johanmalm: Done

src/desktop.c
johanmalm: Missed this one. Have just added and re-force-pushed.

src/layers.c
```diff
@@ -20,6 +19,26 @@
...
 
+static void
+arrange_one_layer(struct output *output, const struct wlr_box *full_area,
+		struct wlr_box *usable_area, struct wlr_scene_tree *tree,
+		bool exclusive)
+{
+	struct wlr_scene_node *node;
+	wl_list_for_each(node, &tree->children, link) {
+		/* ignore nodes in the process of being destroyed */
+		if (!node || !node->data) {
```
Consolatis, line 30: I don't think `!node` can happen here.

The `!node->data` shouldn't happen either but better safe than sorry.

src/layers.c
```diff
@@ -37,55 +56,41 @@ layers_arrange(struct output *output)
...
 	for (size_t i = 0; i < nlayers; ++i) {
-		wl_list_for_each_reverse(layer,
-				&output->layers[layers_above_shell[i]], link) {
-			struct wlr_layer_surface_v1 *layer_surface =
-				layer->scene_layer_surface->layer_surface;
-			if (layer_surface->current.keyboard_interactive) {
-				topmost = layer;
+		struct wlr_scene_tree *tree = output->layer_tree[layers_above_views[i]];
+		/* Iterate in reverse to give most recent node preference */
+		wl_list_for_each_reverse(node, &tree->children, link) {
```
Consolatis, line 89: This one doesn't check for `!node->data` like the loop above, so whatever route we take we should likely take the same route in both cases.

src/layers.c
```diff
@@ -131,60 +137,62 @@ surface_commit_notify(struct wl_listener *listener, void *data)
...
-	if (layer_surface->current.committed
-			|| layer->mapped != layer_surface->mapped) {
-		layer->mapped = layer_surface->mapped;
-		struct output *output =
-			output_from_wlr_output(layer->server, wlr_output);
-		layers_arrange(output);
+	/*
+	 * 'committed' here refers to change in layer-surface state including
+	 * desired-size, anchor, margin, exlusive-zone, keyboard-interactivity
+	 * and/or change-layer, rather than frame-commit.
```
Consolatis, line 143: I don't think this comment is correct, I added a `wlr_log(WLR_ERROR, "layer commit");` after the comment and it gets triggered on frame updates as well. Can be seen by just having some panel running and then switching focus between two windows.

Consolatis:
The mouse enter event now seems to work correctly for me using the example for the original issue :)

Other than the minor stuff I posted above this seems ready to go. Although I didn't actually test the PR in depth, there aren't really that many logic changes in here so I'm pretty positive that this won't break anything major (famous last words).

src/layers.c
```diff
@@ -37,55 +56,41 @@ layers_arrange(struct output *output)
...
-		}
-
-		wlr_scene_node_set_position(&output->layer_tree[i]->node,
-			scene_output->x, scene_output->y);
+		/* Set node position to account for output layout change */
+		wlr_scene_node_set_position(&layer->node, scene_output->x,
+			scene_output->y);
 	}
 
 	memcpy(&output->usable_area, &usable_area, sizeof(struct wlr_box));
```
Consolatis, line 76: Although not changed in this PR, IMHO we should use `sizeof(output->usable_area)` here to make sure we don't have an out-of-bounds write whenever we decide to change from `struct wlr_box` to something else in the future.

johanmalm: Done

src/layers.c
johanmalm: Comment removed.

src/layers.c
johanmalm: Agree, it shouldn't happen. I can't force any errors, so have removed. Was probably being over-cautious.

src/layers.c
johanmalm: Same.

johanmalm:
Review of final commit (relating to layer-subsurfaces) would be welcome

Consolatis:
Nice clean up here. Thanks :)

Issue #369, opened 2022-06-03
Disabling and re-enabling of an output makes the panel disappear


Consolatis:
Only a single output available, thus forced movement of layer shell surfaces.

Reported-by @Flrian via IRC (thanks!)

```
==27600==Hint: address points to the zero page.
    #0 0x5623e8d732df in new_popup_notify ../src/layers.c:248
    #1 0x5623e8fe1a6d in wlr_signal_emit_safe ../subprojects/wlroots/util/signal.c:29
    #2 0x5623e8f7c423 in layer_surface_handle_get_popup ../subprojects/wlroots/types/wlr_layer_shell_v1.c:226
    #3 0x7f55cbe39391  (/lib64/libffi.so.8+0x7391)
    #4 0x7f55cbe361d1  (/lib64/libffi.so.8+0x41d1)
    #5 0x7f55cd1a8077  (/lib64/libwayland-server.so.0+0x9077)
    #6 0x7f55cd1ac1b3  (/lib64/libwayland-server.so.0+0xd1b3)
    #7 0x7f55cd1aac71 in wl_event_loop_dispatch (/lib64/libwayland-server.so.0+0xbc71)
    #8 0x7f55cd1ab3b4 in wl_display_run (/lib64/libwayland-server.so.0+0xc3b4)
    #9 0x5623e8d75e73 in main ../src/main.c:105
```

which is
```c
struct output *output = toplevel->scene_layer_surface->layer_surface->output->data;
```

Consolatis:
After some debugging I found the issue for latest master and.. its complicated:

We are using the layershell protocol in a way that its not supposed to be used regarding output destruction (at least as far as I understand). Usually the [protocol](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-layer-shell-unstable-v1.xml#L355) expects to just destroy a layershell surface when the corresponding output is gone but we can't do that right now because of an issue in wlroots. I am not 100% certain this also should be done when the layershell client didn't specify any output to begin with though.

Upstream is working on a fix for the underlying issue
- https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3513
- https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3530

In the meantime, I will throw a work-around patch together to fix our current usage of moving the layer surface to a new output.

Debugging this also uncovered two more issues:
- [ ] we are not currently updating the outputs of views when they are moved to a new output because the old one is gone
[...]

PR #700, opened 2022-12-26
Fix various crashes related to no outputs available

Consolatis:
- [x] `output_usable_area_from_cursor_coords()`, called for example when spawning a new window
- [x] `view_set_fullscreen()`
- [x] `view_apply_fullscreen_geometry()`
- [x] `new_layer_surface_notify()`
- [x] potentially `src/layers.c` `new_popup_notify()`
- [x] destroy layer surfaces on assigned output destruction, thus finally get https://github.com/labwc/labwc/issues/369#issuecomment-1254189860 done.
- [x] verify layer popups are destroyed when loosing the output
- [x] ensure fullscreen'd windows get properly un-fullscreen'd (including being restored to their natural geometry) if their output goes away

There might be more issues, most of this has been tested by running nested:
- `sleep 5; foot`
- `sleep 5; foot -F`
- `sleep 5; wtype -M logo f` (which is bound to `ToggleFullscreen` in my nested instance)
- `sleep 5; wl_framework/run_example wl_framework/examples/wl_example_panel.py`

and then closing the nested labwc window.

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -316,6 +316,12 @@ handle_new_layer_surface(struct wl_listener *listener, void *data)
 		struct wlr_output *output = wlr_output_layout_output_at(
 			server->output_layout, server->seat.cursor->x,
 			server->seat.cursor->y);
+		if (!output) {
+			wlr_log(WLR_INFO,
+				"No output available to assign layer surface");
+			wlr_layer_surface_v1_destroy(layer_surface);
+			return;
+		}
 		layer_surface->output = output;
 	}
 
@@ -360,11 +366,6 @@ handle_new_layer_surface(struct wl_listener *listener, void *data)
 	surface->node_destroy.notify = handle_node_destroy;
 	wl_signal_add(&surface->scene_layer_surface->tree->node.events.destroy,
 		&surface->node_destroy);
-
-	if (!output) {
-		wlr_log(WLR_ERROR, "no output for layer");
-		return;
-	}
 
 	/*
 	 * Temporarily set the layer's current state to pending so that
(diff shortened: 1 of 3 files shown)


src/output.c
```diff
@@ -435,6 +435,7 @@ output_from_wlr_output(struct server *server, struct wlr_output *wlr_output)
 struct wlr_box
 output_usable_area_in_layout_coords(struct output *output)
 {
+	assert(output);
```
Consolatis, line 438: Alternative:
```c
if (!output) {
    return (struct wlr_box){0};
```

And then remove the new check from `output_usable_area_from_cursor_coords()`

src/layers.c
```diff
@@ -318,6 +318,9 @@ new_layer_surface_notify(struct wl_listener *listener, void *data)
 		struct wlr_output *output = wlr_output_layout_output_at(
 			server->output_layout, server->seat.cursor->x,
 			server->seat.cursor->y);
+		if (!output) {
+			/* FIXME: what to do here? */
```
Consolatis, line 322: I am not sure how to deal with the new layer_surface in particular, should we just send back some wayland error in that case?

Consolatis: The scenegraph for sway seems to simply destroy the surface in that case:

https://github.com/swaywm/sway/blob/993deff56993dd5565061a73b6ca2d65cfa6e135/sway/desktop/layer_shell.c#L323-L332

johanmalm: Yes, I think this is it. `wlr_layer_surface_v1_destroy()` should sort the associated scene-node

src/output.c
johanmalm: Looks good.

Consolatis:
I think this is ready for review.

johanmalm:
Super effort.
LGTM.

PR #703, opened 2022-12-29
src/workspaces.c: prevent re-focus for always-on-top views

Consolatis:
Situation: Switching workspaces while having the keyboard focus on an always-on-top window

Options:

1. re-focus to the topmost window on the new workspace
2. re-focus to the topmost always-on-top window
3. if the focus is on an always-on-top window do not change focus at all, otherwise re-focus to the topmost window

`1.` is the current state which is kind of annoying because you'll loose the keyboard focus.
`2.` doesn't sound right either, imagine having some kind of chat or log-output set to always-on-top and giving it focus whenever switching workspaces
`3.` is the best UX option in my opinion thus what I implemented in this PR.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -143,6 +143,7 @@ void view_set_fullscreen(struct view *view, bool fullscreen,
 void view_toggle_maximize(struct view *view);
 void view_toggle_decorations(struct view *view);
 void view_toggle_always_on_top(struct view *view);
+bool view_is_always_on_top(struct view *view);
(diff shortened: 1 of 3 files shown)


johanmalm:
LGTM
Agree with approach 3
Not tested
Code looks ready for merge

Consolatis:
Made `bool view_is_always_on_top(struct view *view)` public so we don't have to touch view internal stuff within `src/workspaces.c`.

PR #705, opened 2022-12-31
src/view.c: on un-fullscreen restore SSD before applying previous geometry

Consolatis:
Before this patch following would cause the SSD to be rendered offscreen:
- snap a window to left or right edge
- toggle fullscreen to fullscreen
- toggle fullscreen to un-fullscreen

To fix that restore the SSD before calculating the new geometry.


johanmalm:
LGTM

PR #706, opened 2023-01-01
Fix/small refactor

Consolatis:
Some small refactoring which was split from #433 and makes rebasing / review easier.


johanmalm:
LGTM

PR #712, opened 2023-01-04
src/view.c: move xwayland specific function to xwayland.c

Consolatis:
Also remove the `<xcb/xcb_icccm.h>` include as its already included by
`"labwc.h"` -> `<wlr/xwayland.h>` -> `<wlr/xwayland/xwayland.h>`.

Addresses https://github.com/labwc/labwc/pull/682#issuecomment-1357614626

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -9,12 +9,11 @@
 #include "view.h"
 #include "workspaces.h"
 
-#if HAVE_XWAYLAND
-#include <xcb/xcb_icccm.h>
-#endif
-
-#define LAB_FALLBACK_WIDTH 640
+#define LAB_MIN_VIEW_WIDTH  100
+#define LAB_MIN_VIEW_HEIGHT  60
+#define LAB_FALLBACK_WIDTH  640
 #define LAB_FALLBACK_HEIGHT 480
+
 #define MAX(a, b) (((a) > (b)) ? (a) : (b))
 
 /**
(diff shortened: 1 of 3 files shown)


src/view.c
```diff
@@ -168,49 +167,20 @@ view_move_resize(struct view *view, struct wlr_box geo)
...
-				hints->width_inc);
-			*h = round_to_increment(*h, hints->base_height,
-				hints->height_inc);
-
-			min_width = MAX(1, hints->min_width);
-			min_height = MAX(1, hints->min_height);
-		}
+#if HAVE_XWAYLAND
+	if (xwayland_apply_size_hints(view, w, h)) {
+		/* We don't want to cap the size to keep the aspect ratio */
```
Consolatis, line 177: Do we really want this though? At least `GDK_BACKEND=x11 mate-terminal` can turn out smaller than the SSD which is kind of weird IMO.

src/xwayland.c
```diff
@@ -7,6 +7,43 @@
 #include "view.h"
 #include "workspaces.h"
 
+#define MAX(a, b) (((a) > (b)) ? (a) : (b))
```
Consolatis, line 10: I think we really need a generic `MIN(a, b)` and `MAX(a, b)` macro somewhere, its only a matter of time unless we get a redefinition error somewhere. I usually tend to define it before using and then undefine it right after usage but that produces some pretty ugly code.

src/view.c
jlindgren90: XWayland windows can set their own min size, and some are not decorated. For example Audacious can have windows "rolled up" to less than 60px high.

src/xwayland.c
jlindgren90: That would be nice. I would say just define MIN and MAX in labwc.h and be done with it.

johanmalm: Have been thinking the same.
MAX/MIM are defined by cairo/pango, so we’d have to call it something else. 
Lower-case?

Consolatis: This works fine (labwc.h):
```
#define MIN(a, b) (((a) < (b)) ? (a) : (b))
#define MAX(a, b) (((a) > (b)) ? (a) : (b))
```

However, this breaks (labwc.h, wrapped in unnecessary `(` and `)`):
```
#define MIN(a, b) ((((a) < (b)) ? (a) : (b)))
#define MAX(a, b) ((((a) > (b)) ? (a) : (b)))
```
```
In file included from ../src/osd.c:14:
../include/labwc.h:64: warning: "MIN" redefined
   64 | #define MIN(a, b) ((((a) < (b)) ? (a) : (b)))
      |
In file included from /usr/lib/aarch64-linux-gnu/glib-2.0/include/glibconfig.h:9,
                 from /usr/include/glib-2.0/glib/gtypes.h:32,
                 from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from /usr/include/glib-2.0/gobject/gbinding.h:28,
                 from /usr/include/glib-2.0/glib-object.h:22,
                 from /usr/include/pango-1.0/pango/pango-coverage.h:25,
                 from /usr/include/pango-1.0/pango/pango-font.h:25,
                 from /usr/include/pango-1.0/pango/pango-attributes.h:25,
                 from /usr/include/pango-1.0/pango/pango.h:25,
                 from /usr/include/pango-1.0/pango/pangocairo.h:25,
                 from ../src/osd.c:6:
/usr/include/glib-2.0/glib/gmacros.h:887: note: this is the location of the previous definition
```

So it seems to work when defining the macros exactly like in `gmacros.h`.
I don't know if we should rely on that though. We could also just include the glib macros instead.

src/view.c
jlindgren90: That could work, no objections here.

src/xwayland.c
Consolatis: `#ifndef MIN/MAX` would also be fine by me. No strong preferences for either `#undef` or `#ifndef`.

jlindgren90:
LGTM

src/view.c
johanmalm: Agree with 4*BUTTON_WIDTH approach

PR #713, opened 2023-01-04
output: Add output_update_usable_area/all_usable_areas()

Consolatis:
Move the desktop_arrange_all_views() call outside layers_arrange() into a new function, output_update_usable_area().  The new function currently does exactly what layers_arrange() used to, but will be expanded in a later commit.

Add output_update_all_usable_areas(), which is the same as calling output_update_usable_area() for each output, but only calls desktop_arrange_all_views() once.

Rebased and slightly modified by @Consolatis

---

Split off from #625 and addressed the review commits.
Further change:
- call `cursor_update_focus()` on layer surface commits which change something other than just the buffer content itself to keep the `pointer.enter` event working for new layer surfaces spawning under the cursor (and when moving / resizing layer surfaces).

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -423,6 +423,8 @@ void output_init(struct server *server);
 void output_manager_init(struct server *server);
 struct output *output_from_wlr_output(struct server *server,
 	struct wlr_output *wlr_output);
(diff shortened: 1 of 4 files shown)


jlindgren90:
LGTM

PR #715, opened 2023-01-06
include/labwc.h: Provide MIN/MAX macros

Consolatis:
As discussed in https://github.com/labwc/labwc/pull/712#discussion_r1061114212

We could also move the macros from `labwc.h` to some other header, I don't particularly mind either variant.


jlindgren90:
LGTM

PR #720, opened 2023-01-07
xwayland: Add xwayland.h and move more things to xwayland.c

jlindgren90:
- Move xwayland-specific struct definitions to new `xwayland.h` header
- Move `xwayland_move_sub_views_to_front()` from `desktop.c` to `xwayland.c`
- Split out `xwayland_server_init/finish()` from `server_init/finish()`
- Rename `new_xwayland_surface` -> `xwayland_new_surface` and `xwayland_surface_new()` -> `handle_new_surface()` for consistency
- Add `mapped` argument to `xwayland_unmanaged_create()` so that we can make `unmanaged_handle_map()` private to `xwayland-unmanaged.c`

### DIFF
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -1,0 +1,56 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_XWAYLAND_H
+#define __LABWC_XWAYLAND_H
+#include "config.h"
+#if HAVE_XWAYLAND
+#include "view.h"
+
+struct wlr_compositor;
+
+struct xwayland_unmanaged {
+	struct server *server;
+	struct wlr_xwayland_surface *xwayland_surface;
+	struct wlr_scene_node *node;
+	struct wl_list link;
+
+	struct wl_listener request_activate;
+	struct wl_listener request_configure;
+/*	struct wl_listener request_fullscreen; */
+	struct wl_listener set_geometry;
+	struct wl_listener map;
+	struct wl_listener unmap;
+	struct wl_listener destroy;
+	struct wl_listener override_redirect;
+};
+
+struct xwayland_view {
+	struct view base;
+	struct wlr_xwayland_surface *xwayland_surface;
+
+	/* Events unique to XWayland views */
+	struct wl_listener request_configure;
+	struct wl_listener set_app_id;		/* TODO: s/set_app_id/class/ */
+	struct wl_listener set_decorations;
+	struct wl_listener override_redirect;
+
+	/* Not (yet) implemented */
+/*	struct wl_listener set_role; */
+/*	struct wl_listener set_window_type; */
+/*	struct wl_listener set_hints; */
+};
+
+void xwayland_unmanaged_create(struct server *server,
+	struct wlr_xwayland_surface *xsurface, bool mapped);
+
+struct wlr_xwayland_surface *xwayland_surface_from_view(struct view *view);
+
+bool xwayland_apply_size_hints(struct view *view, int *w, int *h);
+void xwayland_move_sub_views_to_front(struct view *parent,
+	void (*move_to_front)(struct view *view));
+
+void xwayland_server_init(struct server *server,
+	struct wlr_compositor *compositor);
+void xwayland_server_finish(struct server *server);
+
+#endif /* HAVE_XWAYLAND */
+#endif /* __LABWC_XWAYLAND_H */
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -10,6 +10,7 @@
 #include "ssd.h"
 #include "view.h"
 #include "workspaces.h"
+#include "xwayland.h"
 
 static void
 move_to_front(struct view *view)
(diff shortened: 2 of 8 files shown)


include/xwayland.h
```diff
@@ -0,0 +1,56 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_XWAYLAND_H
+#define __LABWC_XWAYLAND_H
+#include "config.h"
+#if HAVE_XWAYLAND
```
Consolatis, line 5: I like this, it makes the `#include "xwayland.h"` statements much cleaner.

src/desktop.c
```diff
@@ -68,7 +28,7 @@ desktop_move_to_front(struct view *view)
 	}
 	move_to_front(view);
 #if HAVE_XWAYLAND
-	move_xwayland_sub_views_to_front(view);
+	xwayland_move_sub_views_to_front(view, move_to_front);
```
Consolatis, line 31: Usually I am not that much a fan of function pointers for cases like this but this still seems cleaner than to make the `move_to_front()` helper public or to repeat its code in xwayland.c.

Consolatis (commented):
Nice refactor.

src/server.c
```diff
@@ -11,6 +11,9 @@
 #include <wlr/types/wlr_primary_selection_v1.h>
 #include <wlr/types/wlr_screencopy_v1.h>
 #include <wlr/types/wlr_viewporter.h>
+#if HAVE_XWAYLAND
+#include <wlr/xwayland.h>
+#endif
```
Consolatis, line 16: We could remove the last `<wlr/xwayland.h>` include completely by adding a filter function to `xwayland.c`. Something like
```c
bool
xwayland_allow_global(struct server *server, const struct wl_client *client,
		const struct wl_interface *interface)
{
	struct wl_client *xwayland_client =
		server->xwayland ? server->xwayland->server->client : NULL;
	if (!xwayland_client || client != xwayland_client) {
		return true;
	}

	// Comment from original function
	if (!strcmp(iface->name, wp_drm_lease_device_v1_interface.name)) {
		return false;
	}
	return true;
}
```

and then call it from `server.c` like
```c
	[..]
	// Potentially further checks in the future
#if HAVE_XWAYLAND
	if (!xwayland_allow_global(server, client, iface)) {
		return false;
	}
#endif
	// Potentially further checks in the future
	return true;
```

That would also require to move some more includes to `xwayland.c` though, like `"drm-lease-v1-protocol.h"` and potentially `<wayland-server-core.h>` (not sure about the last one).

jlindgren90: I wasn't sure what to do with the drm-lease stuff, mainly since I don't really understand what it's doing (some kind of a workaround from what I could see?)

Is it all right to leave that for a future MR?

src/desktop.c
jlindgren90: Yeah it's a little bit awkward, but I was struggling to come up with a better alternative.

src/server.c
Consolatis: The DRM lease protocol allows applications to exclusively take over the whole output, it is most commonly used for VR headsets but there is also a branch of gamescope which can take advantage of it to render games directly to the output instead of going through labwc.

Sadly the xwayland implementation for leased outputs had (or still has) a bug which causes some applications (mainly chromium based ones running in X11 mode) to lag over time. To prevent that from happening we just stop announcing lease support to xwayland (as its itself a wayland client) until there is a better solution available.

> Is it all right to leave that for a future MR?

Sure.

src/desktop.c
Consolatis: I can't think of a better solution either so I guess this is it for now.

Consolatis:
The code itself LGTM, didn't test in depth but it still compiles fine (with and without xwayland enabled) and still works nested in both cases.

Feel free to "Rebase and merge" if you are satisfied with the PR.

johanmalm:
Thanks for this. Looks good. 

I like having xwayland's handle_new_surface stuff all within one translation unit (i.e. both the wl_signal_add() and the callback function). That is how we have gone for most areas now. Possibly just xdg-shell left to tackle :smile:

Please consider the points below for further improvements.

## Point 1

Passing the function pointer for `move_to_front()` is a neat trick, but inconsistent with labwc general design patterns. As an alternative, we could simply put `move_to_front()` in view.c and make it public as `view_move_to_front()`, however I suggest considering the following:

In `struct view_impl` (view.h:24) , add a `void (*move_to_front)(struct view *view, struct wl_list *views);`

That way we can handle the xwayland and xdg-shell methods in what feels to me like a more encapsulated way.

My preferred approach is to wrap the implementation function pointer with:

```
void view_move_to_front(struct view *view, struct wl_list *views)
{
	assert(view);
	if (view->impl->move_to_front) {
		view->impl->move_to_front(view, views);
	}
}
```

This leaves `desktop_move_to_front()` to merely call `view_move_to_front()` and `cursor_update_focus()`.

If we adopt this approach, I suppose we would end up with small duplication in the xdg-shell and xwayland implementations:

```
	wl_list_remove(&view->link);
	wl_list_insert(&view->server->views, &view->link);
	wlr_scene_node_raise_to_top(&view->scene_tree->node);
```

This is what `view-impl.c` was inteded for. Maybe we should rename it `view-impl-common.c`. The code snippet above could of course exist in both `xwayland.c` and `xdg.c`, or as a standalone function in `view.c`, but in my opinion we reduce coupling if the code the is common to `xwayland.c` and `xdg.c` (and is not called from anywhere else) lives in `view-impl.c`.

## Point 2

Consider the same approach (method dispatch) for xwayland_apply_size_hints(). It does not matter if xdg-shell doesn't have one.

We could even include the `LAB_MIN_VIEW_{WIDTH,HEIGHT}` in this method. Maybe call it `apply_size_constraints()` or something like that. I'm just looking at `view_adjust_size()` and reflect that we actually treat xwayland and xdg-shell differently in terms of min height/width, so a more object oriented approach feels right.

## Point 3

Consider 's/xwayland_server_init/xwayland_init/'. Just a suggestion, don't feel strongly.

Consolatis:
> In `struct view_impl` (view.h:24) , add a `void (*move_to_front)(struct view *view, struct wl_list *views);`

\+ impl living in `view-imp.c` / `view-impl-common.c`
I like this, this is much cleaner.

Re size-hints: https://github.com/labwc/labwc/pull/712#discussion_r1061461121

If we decide to implement that we'd have
- a generic `adjust_size()` which makes sure that geo is >= SSD_MIN_WIDTH if `view->ssd_enabled`
- additionally `xwayland_apply_size_constrains()` for xwayland views

So we could throw the generic part into `view-impl-common.c` as well and then assign that to `struct view_impl` in `xdg.c` and have an own implementation in `xwayland.c` which applies the size hints and then calls the generic impl one.

That caller would then always just call `view->impl->adjust_size(view)`.

PR #733, opened 2023-01-27
README.md: suggest using nested gamescope for gaming

### LINKED ISSUE #690: Mention gamescope on website / readme
Debugging X11 wine game issues is basically impossible for me because of my aarch64 system and the other core devs don't seem to game all day either. Most issues apparently can be "solved" by running the games via a nested gamescope instance so we should mention that prominently in the readme / website to make sure users are able to game as easy as possible on labwc.

Consolatis:
Fixes #690

@Joshua-Ashton @Flrian

As I can't test it myself, is this the right way to use gamescope for steam games?
Steam doesn't seem to provide any generic command prefix so I guess users have to modify the launch option of every single game they want to play unless I missed something.

Also not sure if we should mention `sudo setcap 'CAP_SYS_NICE=eip' $(which gamescope)` in there (and if labwc should try to renice to `-10` or something as well).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ href="NEWS.md">Release&nbsp;Notes</a>]</h3>
 - [3. Configuration](#3-configuration)
 - [4. Theming](#4-theming)
 - [5. Usage](#5-usage)
+  - [5.1 Gaming](#51-gaming)
 - [6. Integration](#6-integration)
 
 ## 1. Project Description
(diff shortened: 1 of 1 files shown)


Consolatis:
I'd suggest merging this for now, we can always change the exact text later again.

johanmalm:
Good idea. I was thinking the same before you said it :smile:

Issue #704, opened 2022-12-30
Layer-shell keyboard interactivity


johanmalm:
This covers a new issue discovered by @Tamaranch at the end of #667.

> @johanmalm Yes I can confirm that this issue is fixed +1 However, keyboard interactivity does not work properly anymore for me: the Esc key does not close the panel menus (e.g. apps menu). It worked until [d7eecca](https://github.com/labwc/labwc/commit/d7eecca34db0b65708713262cb4951bf39de62af) (at least).

I have narrowed the 'breakage' down to this commit https://github.com/labwc/labwc/commit/86fa0c0a6558eb85542e78eb050aa8b8a58b42eb

I'm not sure what the correct approach is here. It appears some clients want `get_cursor_context()` to return layer-shell subsurface surfaces, and others don't. Grateful for advice if anyone has knowledge or experience in this space.

Consolatis:
Maybe we could get away with storing both surfaces, the main one used for keyboard focus and the other one for cursor focus?

Consolatis:
As stated in https://github.com/labwc/labwc/pull/734#issuecomment-1407637384 this seems fixed.
Feel free to comment / reopen if that is not the case.

PR #734, opened 2023-01-28
layer: improve keyboard-interactivity

### LINKED ISSUE #725: Focus lost when a notification occurs
Hello

When tools such as: 
- https://github.com/ErikReider/SwayNotificationCenter 
- https://github.com/francma/wob
- https://github.com/dunst-project/dunst
 displays a popup, the focus on the currently active window is lost:

cropped2

I tried to enable the ` followMouse` option, which does not really solve the issue as I still have to move the mouse to get back the focus on the previously focused window.

Is there an option that allows to keep the focus when a notification is displayed on the screen?

johanmalm:
- Process layer-shell keyboard interactivity in the map and commit handlers only, rather than in layers_arrange(). This allows handling of the layer-surface that emitted the event rather than iterating over all surfaces in the output layer-tree, and therefore avoids having to guess which surface should have keyboard preference (and it might not be the last one added to the list which was the assumption previously).

- Prevent seat_focus_surface() from setting keyboard focus if a layer-shell surface with exclusive keyboard-interactivity has the focus.

- Set cursor_context type for layer-surfaces without node-descriptors in order to set keyboard focus correctly in cursor_button_press().

Tested satisfactorily with xfce4-panel and gtk-layer-demo.

Fixes #725 and #704

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -75,38 +75,6 @@ layers_arrange(struct output *output)
 	}
 
 	output->usable_area = usable_area;
-
-	/* Find topmost keyboard interactive layer, if such a layer exists */
-	uint32_t layers_above_views[] = {
-		ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY,
-		ZWLR_LAYER_SHELL_V1_LAYER_TOP,
-	};
-	size_t nlayers = sizeof(layers_above_views) / sizeof(layers_above_views[0]);
-	struct lab_layer_surface *topmost = NULL;
-	struct wlr_scene_node *node;
-	for (size_t i = 0; i < nlayers; ++i) {
-		struct wlr_scene_tree *tree = output->layer_tree[layers_above_views[i]];
-		/* Iterate in reverse to give most recent node preference */
-		wl_list_for_each_reverse(node, &tree->children, link) {
-			struct lab_layer_surface *surface = node_layer_surface_from_node(node);
-			struct wlr_scene_layer_surface_v1 *scene = surface->scene_layer_surface;
-			if (scene->layer_surface->current.keyboard_interactive) {
-				topmost = surface;
-				break;
-			}
-		}
-		if (topmost) {
-			break;
-		}
-	}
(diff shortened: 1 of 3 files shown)


johanmalm:
@Hazegard and @Tamaranch - testing would be gratefully received if you are able to.

src/layers.c
```diff
@@ -118,6 +86,34 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+	if (layer_surface->current.keyboard_interactive
+			&& layer_surface->current.layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP) {
+		/*
+		 * Give keyboard focus to surface if
+		 * - keyboard-interactivity is 'exclusive' or 'on-demand'; and
+		 * - surface is in top/overlay layers; and
+		 * - currently focused layer has a lower precedence
+		 */
+		if (!seat->focused_layer || seat->focused_layer->current.layer
+				>= layer_surface->current.layer) {
```
Consolatis, line 104: This seems to contradict the comment above, shouldn't it be `seat->focused_layer->current.layer < layer_surface->current.layer` to match `currently focused layer has a lower precedence`?

Tamaranch:
@johanmalm Yes according to my tests it fixes both issues :+1:
Can we expect to have this in 0.6.1? (would be nice :))

Hazegard:
I also tested and I confirm that is fixes #725 !

src/layers.c
johanmalm: You're 100% right. Thanks you. The power of peer review!
I've changed this + extended the comment to make it super clear.

johanmalm:
> @johanmalm Yes according to my tests it fixes both issues +1 Can we expect to have this in 0.6.1? (would be nice :))

Thanks.
Yes - we are indeed 'delaying' `0.6.1` to get this patch in :smile:

johanmalm:
> I also tested and I confirm that is fixes #725 !

Thanks.

PR #735, opened 2023-01-29
Allow clearing key/mouse bindings

### LINKED ISSUE #567: Add clear-keybind functionality
We are missing an actual clear-keybind functionality. Maybe `actions_run()` could have a `bool` return value which allows the key combination to be delivered to the focused application instead? Or do we need a futher `Clear` action instead? Or do we solve this by implementing the suggestion above with deduplication + clearing of the bindings with empty action list? We could also add a `SendKeys` action which would possibly allow for some macro magic as well.

Currently its impossible to use `A-Left` or `A-Right` for applications while also using `<default />` for keybinds, even when setting the keybind to the new `None` action.

_Originally posted by @Consolatis in https://github.com/labwc/labwc/issues/529#issuecomment-1253201447_

Consolatis:
Fixes #567

I did not really test this in depth so draft for now until it received some proper testing.

Changes this PR is doing:
- when creating a `None` action we don't actually create an action at all and just return `NULL` for the action
- we deduplicate keybinds the same as mousebinds (so we also get rid of the `_reverse()` when checking for keybinds + loop over slightly less entries in the keybind list for every key event)
- after the deduplication, all bindings with an empty action list are removed

This means that both of these variants now cause the creation of a keybind without any action which then first replaces all previous binding for the same key and finally gets removed because it has an empty action list, effectively removing an earlier binding set for `W-F4`:
```xml
<keybind key="W-F4" />
<keybind key="W-F4">
	<action name="None" />
</keybind>
```

TODO:
- [x] rename `merge_xxx_bindings()` to `deduplicate_xxx_bindings()`
- [ ] Minor: `merge_mouse_bindings()` and `merge_key_bindings()` are straight up copies of each other but I can't think of an easy way to reuse just one function for both as they operate on different structs, even though we only access the same members. Any hints on how to deduplicate the deduplication functions appreciated. We could use a base `bind` struct as first member of both structs which only carries the `actions` list and `struct wl_list link` + supplying a function pointer for `xxx_the_same()` which would then do the upcast again but I am not sure if its worth it. It would also mean that we have to update all other loops over `rc.xxx_bindings` in labwc as the `link` member changed to `base.link`, same for all users of the bind actions. Another option could be to have `link` and `actions` being the first members in both structs, define a new struct with just these 2 members and then just cast the mouse/keybind structs to that new struct. So something like an overlay struct. That feels really hacky and pretty dirty though.

### DIFF
--- a/docs/rc.xml
+++ b/docs/rc.xml
@@ -23,6 +23,13 @@
     <keybind key="W-Return">
       <action name="Execute" command="foot" />
     </keybind>
+    <!--
+      Remove a previously defined keybind
+      A shorter alternative is <keybind key="W-F4" />
+    -->
+    <keybind key="W-F4">
+      <action name="None" />
+    </keybind>
   </keyboard>
 
   <mouse>
--- a/include/config/keybind.h
+++ b/include/config/keybind.h
@@ -28,4 +28,6 @@ struct keybind *keybind_create(const char *keybind);
  */
 uint32_t parse_modifier(const char *symname);
 
+bool keybind_the_same(struct keybind *a, struct keybind *b);
+
 #endif /* __LABWC_KEYBIND_H */
--- a/include/config/mousebind.h
+++ b/include/config/mousebind.h
@@ -48,5 +48,6 @@ enum mouse_event mousebind_event_from_str(const char *str);
 uint32_t mousebind_button_from_str(const char *str, uint32_t *modifiers);
 enum direction mousebind_direction_from_str(const char *str, uint32_t *modifiers);
 struct mousebind *mousebind_create(const char *context);
+bool mousebind_the_same(struct mousebind *a, struct mousebind *b);
 
 #endif /* __LABWC_MOUSEBIND_H */
--- a/src/action.c
+++ b/src/action.c
@@ -157,8 +157,14 @@ action_create(const char *action_name)
 		wlr_log(WLR_ERROR, "action name not specified");
 		return NULL;
 	}
+
+	enum action_type action_type = action_type_from_str(action_name);
+	if (action_type == ACTION_TYPE_NONE) {
+		return NULL;
+	}
+
 	struct action *action = znew(*action);
-	action->type = action_type_from_str(action_name);
+	action->type = action_type;
 	wl_list_init(&action->args);
 	return action;
 }
@@ -444,8 +450,6 @@ actions_run(struct view *activator, struct server *server,
 			} else {
 				wlr_log(WLR_ERROR, "Invalid SnapToRegion id: '%s'", region_name);
 			}
-			break;
-		case ACTION_TYPE_NONE:
 			break;
 		case ACTION_TYPE_INVALID:
 			wlr_log(WLR_ERROR, "Not executing unknown action");
--- a/src/config/keybind.c
+++ b/src/config/keybind.c
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #define _POSIX_C_SOURCE 200809L
+#include <assert.h>
 #include <glib.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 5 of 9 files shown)


Consolatis:
I am now running this on my main DRM session and it works fine and allowed me to drastically reduce the size of my `rc.xml`. This feels ready to merge.

<details><summary>rc.xml</summary>

```xml
<?xml version="1.0"?>
<labwc_config>
	<core cycleViewPreview="yes" />
	<resistance screenEdgeStrength="0" />
	<theme name="Numix" />>

	<desktops popupTime="0">
		<names>
			<name>Workspace 1</name>
			<name>Workspace 2</name>
			<name>Workspace 3</name>
		</names>
	</desktops>

	<keyboard>
		<default />

		<!-- Disable A-Left/Right keybinds -->
		<keybind key="A-Left" />
		<keybind key="A-Right" />

		<!-- Custom keybinds -->
		<keybind key="W-1">
			<action name="ToggleDecorations" />
			<action name="ToggleMaximize" />
		</keybind>
		<keybind key="W-2">
			<action name="Debug" />
		</keybind>
		<keybind key="W-Return">
		</keybind>
		<keybind key="W-e">
			<action name="Execute" command="runraise.sh Thunar thunar" />
		</keybind>
		<keybind key="W-Up">
			<action name="Execute" command="change_volume 3" />
		</keybind>
		<keybind key="W-Down">
			<action name="Execute" command="change_volume -3" />
		</keybind>
		<keybind key="Print">
			<action name="Execute">
				<command>sh -c 'grim -g "`slurp`"'</command>
			</action>
		</keybind>

		<!-- Workspaces -->
		<keybind key="W-Left">
			<action name="GoToDesktop" to="left" />
		</keybind>
		<keybind key="W-Right">
			<action name="GoToDesktop" to="right" />
		</keybind>
		<keybind key="W-S-Left">
			<action name="SendToDesktop" to="left" />
			<action name="GoToDesktop" to="left" />
		</keybind>
		<keybind key="W-S-Right">
			<action name="SendToDesktop" to="right" />
			<action name="GoToDesktop" to="right" />
		</keybind>
	</keyboard>

	<mouse>
		<default />

		<context name="Root">
			<mousebind button="Right" action="Press">
				<action name="ShowMenu" menu="desktop-menu" />
			</mousebind>
		</context>

	</mouse>

</labwc_config>
```

</details>

johanmalm:
> Side note: I just discovered that our very own `magic-malm-parser` actually supports this as well
> 
> ```
> <keybind key="W-3" name.action="Execute" command.action="foot" />
> ```

Oh, I like that very much. Great discovery! 
Maybe we ought to describe this in the man pages or on the website.
Getting a bit tired this evening. Will review PR tomorrow.

johanmalm:
> Side note: I just discovered that our very own `magic-malm-parser` actually supports this as well
> ```
> <keybind key="W-3" name.action="Execute" command.action="foot" />
> ```

Should we reverse the order of the nodename parsing?

Currently we parse `<a><b><c>` to `c.b.a`. I might be a bit more user-friendly to go `a.b.c`.
The `labwc-tweaks` implementation actually does it that way around :smile:
https://github.com/labwc/labwc-tweaks/blob/253667cb776cd74d9f39067721b1bf5f729c7b45/xml.c#L49

The example above would then turn into:

`<keybind key="W-3" action.name="Execute" action.command="foot" />` which reads better for humans

Consolatis:
> Should we reverse the order of the nodename parsing?

No issues with that, might be a bit of work to get it right though (e.g. the `truncate_at` function).

> Maybe we ought to describe this in the man pages or on the website.

I think we should at least describe these two being equal
```xml
<action>
	<name>Execute</name>
	<command>foot</command>
</action>
```
```xml
```
the `<keybind key="W-3" name.action="Execute" command.action="foot" />` variant I would not document as of right now, it seems a bit hacky and unintuitive. If we indeed end up changing the order of nodename parsing that would be valid option as well though.

johanmalm:
LGTM. 
Big thanks for thinking this one through and pushing it over the line.
Agree, it's not worth doing OO inheritance, base struct, etc. for the sake of those two `deduplicate` functions. I suppose, some people might consider a macro, but I personally think we'd lose more than we gain.
Will merge.

PR #736, opened 2023-01-29
ToggleShowDesktop

### LINKED ISSUE #455: [Feature Request] Keyboard shortcut for "show desktop"
In Openbox there is the shortcut W-d to hide all windows and shows the desktop.

Consolatis:
Fixes #455 

The main thing to watch out for while testing this is if after pressing `W-d`, random actions correctly restore all windows, namely
- [x] pressing `W-d` again
- [x] opening a new application
- [ ] switching workspaces (only works if there are window on that target workspace)
- [ ] `A-Tab` OSD (works once switched but fails while still in the OSD)
- [x] activating a window via panel
- other?

TODO:
- We should restore all windows when we start the `A-Tab` selection
- We should restore all windows when switching workspaces
- A panel will still show a window being active even though its not even visible anymore. The reason is that we don't have any public way to "disable" a view other than via minimizing or activating a different view. This is an issue:
  - when switching to an empty workspace
  - when toggling show desktop
- There are multiple options to fix that (which should be done in a separate PR):
  - `view_set_activated()` gets a `bool` argument 
  - we have to expose the view internal `_view_set_activated(struct view *view, bool activated)` function
  - we add a new public `view_unset_activated()` function, this would be my favorite as there is only a very limited amount of callers that need this so we don't have to complicate the API for other callers.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -83,6 +83,9 @@ Actions are used in menus and keyboard/mouse bindings.
 	Send active window to workspace.
 	Supported values are the same as for GoToDesktop.
 
+*<action name="ToggleShowDesktop">*
+	Hide / unhide all windows
+
 *<action name="None">*
 	If used as the only action for a binding: clear an earlier defined binding.
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -129,6 +129,9 @@
     <keybind key="W-a">
       <action name="ToggleMaximize" />
     </keybind>
+    <keybind key="W-d">
+      <action name="ToggleShowDesktop" />
+    </keybind>
     <keybind key="A-Left">
       <action name="MoveToEdge" direction="left" />
     </keybind>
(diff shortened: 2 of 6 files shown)


gevera:
Looking forward for this

Faugus:
This is a great function. I'm waiting for it.

Consolatis:
If somebody wants to take this over feel free.

napcok:
In Openbox it works differently... and probably is much easier to implement the same behavior in Labwc.

>  ToggleShowDesktop
> Hides all windows so that the desktop is visible, and gives focus to the desktop window if one exists (such as in GNOME and KDE). You can also use the action again to show the windows again, **if no windows have become visible yet**. 

So you can restore minimized windows only by invoking ToggleShowDesktop again right after the first run.
`W-d` right after `W-d`
So only:

- [x] pressing `W-d` again 

is important. 

Any other actions do not care about ToggleShowDesktop "state" at all.

Consolatis:
In comparison to Openbox, labwc uses a scene tree for everything. What this PR does is to disable the sub tree containing all windows. So the difficulty here is to to ensure that

> if no windows have become visible yet.

actually becomes possible again when that sub tree is currently disabled. Otherwise one would be stuck with no windows visible at all.

napcok:
This isn't an essential action, from what I've read, it would be easy to replace with a wlrctl script. I'll try that.

PR #741, opened 2023-01-30
Add scripts/find-banned.sh

johanmalm:


Consolatis:
Very nice tokenizer!

It seems to miss treating `//` as `SPECIAL_COMMENT_BEGIN`, not sure how to then end the comment though, should be on the next linebreak I think.

Would it make sense to have it read multiple files paths from stdin so we can search across a whole directory tree (helped by `find`)?

Consolatis:
Re single line comments with `//`:

```diff
diff --git a/scripts/helper/find-idents.c b/scripts/helper/find-idents.c
index 69556ae..0531126 100644
--- a/scripts/helper/find-idents.c
+++ b/scripts/helper/find-idents.c
@@ -54,6 +54,7 @@ enum {
 	SPECIAL_COMPARISON_OP,
 	SPECIAL_COMMENT_BEGIN,
 	SPECIAL_COMMENT_END,
+	SPECIAL_COMMENT_LINE,
 };
 
 static char *current_buffer_position;
@@ -198,6 +199,7 @@ struct {
 	{ "!=", SPECIAL_COMPARISON_OP },
 	{ "/*", SPECIAL_COMMENT_BEGIN },
 	{ "*/", SPECIAL_COMMENT_END },
+	{ "//", SPECIAL_COMMENT_LINE },
 	{ ";", ';' },
 	{ "{", '{' },
 	{ "}", '}' },
@@ -276,6 +278,21 @@ handle_preprocessor_directive(void)
 	}
 }
 
+static void
+ignore_single_line(void)
+{
+	for (;;) {
+		++current_buffer_position;
+		if (current_buffer_position[0] == '\0') {
+			return;
+		}
+		if (current_buffer_position[0] == '\n') {
+			++current_line;
+			return;
+		}
+	}
+}
+
 struct token *
 lex(char *buffer)
 {
@@ -309,6 +326,9 @@ lex(char *buffer)
 			token = add_token();
 			get_special_token(token);
 			token->kind = TOKEN_SPECIAL;
+			if (token->special == SPECIAL_COMMENT_LINE) {
+				ignore_single_line();
+			}
 			continue;
 		case '#':
 			handle_preprocessor_directive();
```

johanmalm:
> Would it make sense to have it read multiple (single) files paths from stdin so we can search across multiple files (or whole directory trees, helped by `find`)?

How about this usage:

```
find-idents [OPTIONS] FILES...

-f, --find <comma-separated-string of tokens>     Tokens to find
```

johanmalm:
Thanks @Consolatis  Went for option 2 for C99 `//` style comments.

Consolatis:
> > Would it make sense to have it read multiple (single) files paths from stdin so we can search across multiple files (or whole directory trees, helped by `find`)?
> 
> How about this usage:
> 
> ```
> find-idents [OPTIONS] FILES...
> 
> -f, --find <comma-separated-string of tokens>     Tokens to find
> ```

I am not sure about the allowed count and size of arguments.
For single files, that format seems nicer than reading from stdin.
We could also say that a file argument being `-` causes the tokenizer to read the paths from stdin, additionally to the arguments.

johanmalm:
Steps outstanding:

1. Agree if we should use multi-file approach. On reflection, I can't see the reason for mangling multiple files in each call. Multiple tokes, of course, but files? 
2. Agree what functions that should be banned
3. Clean up code to remove banned functions
4. Incorporate `scripts/find-banned.sh` in CI

I'd be happy to merge after (1) + (2). I think (3) + (4) can come later.

Consolatis:
1. multiple files would be usable to detect functions that should be static (e.g. files using the function in `src/` == 1)
2. I don't think there are really that many:
     - `g_strcmp0`
     - `calloc` / `malloc`
     - others?

johanmalm:
Multiple files supported through stdin.

Have left strcat in for a positive read-out. Will remove before merge.

johanmalm:

scripts/helper/find-idents.c
```diff
@@ -0,0 +1,497 @@
...
+{
+	return;
+	if (current_buffer_position[0] == '\n') {
+		++current_line;
+	}
+	current_buffer_position++;
+	if (isspace(current_buffer_position[0])) {
+		handle_whitespace(token);
+	}
+}
```
Consolatis, line 171: This function is stubbed out. I think this came from my patch. We should likely either remove the function completely or fix it some other way. See top message of https://github.com/johanmalm/labwc/pull/8#issue-1577310757

PR #742, opened 2023-01-31
Enable more compiler warnings

Consolatis:
Basically just copied over the compiler warning settings from wlroots and fixed those that came up for me.

We might want to keep `-Wmissing-field-initializers` and `-Wmissing-braces` though, wlroots disables these (which is what this PR is doing as well). See https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html for a description of both.

Another candidate to enable: `-Wunused-macros`, it only detects unused defines in `.c` files though.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -19,13 +19,24 @@ add_project_arguments(
 
 cc = meson.get_compiler('c')
 
-add_project_arguments(cc.get_supported_arguments(
-  [
-    '-Wno-unused-parameter',
-    '-Wundef',
-  ]),
-  language: 'c',
-)
+add_project_arguments(cc.get_supported_arguments([
+  '-Wundef',
+  '-Wlogical-op',
+  '-Wmissing-include-dirs',
+  '-Wold-style-definition',
+  '-Wpointer-arith',
+  '-Winit-self',
+  '-Wstrict-prototypes',
+  '-Wimplicit-fallthrough=2',
+  '-Wendif-labels',
+  '-Wstrict-aliasing=2',
+  '-Woverflow',
+  '-Wmissing-prototypes',
+  '-Walloca',
+  '-Wunused-macros',
+
+  '-Wno-unused-parameter',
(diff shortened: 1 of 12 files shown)


johanmalm:
LGTM

johanmalm:
Happy to include `-Wmissing-field-initializers` and `-Wmissing-braces`
I'm not sure what's included with `warning_level=2`, but can't see any harm in adding them.

johanmalm:
Setting `warning_level=everything` in `meson.build` makes for interesting reading :smile:

Consolatis:
Apparently this is what that setting actually does:

> * warning_level 1 adds `-Wall`,
> * warning_level 2 adds warning_level 1 + `-Wextra`,
> * warning_level 3 adds warning_level 2 + `-Wpedantic`

So `everything` or `3` isn't really everything but just pedantic which doesn't like something like `wlr_log(WLR_ERROR, "foobar");` because `__VA_ARGS__` seems to require at least one argument so its likely not something we really want to enable. `-Wall` + `-Wextra` + the wlroots collection seems like a nice compromise.

eli-schwartz:
> So `everything` or `3` isn't really everything but just pedantic

warning_level "everything" is the next step up from "3", and is explicitly designed to either pass -Weverything on clang, or emulate it on GCC by passing ~~several hundred~~ a whole lot of warning options by basically going down a semi-exhaustive list of every warning option someone ever added to GCC.

https://mesonbuild.com/Release-notes-for-1-0-0.html#warningleveleverything-option

The intended purpose of this option is to:
- provide a high-level wrapper over clang -Weverything that isn't clang-specific
- enable developers to experiment with highly unstable sets of warnings, see what gets triggered, and discover the names of interesting warning flags

Consolatis:
Thanks for the clarification about `3` != `everything`.

If you have a spare minute, we could need some further advice / confirmation in
- #752

Issue #744, opened 2023-01-31
Show corresponding tasks for each monitor on task bar with Dual Monitors


boris3812:
labwc version 0.6.1 (thanks guys for adding last version in AUR, and it works finally with wlroots 0.16.1 from community )
so of course we have Archlinux 😉

Verified on next panels:
- sfwbar
- wf-panel (from wf-shell)
- waybar (wlr/taskbar module)
(because we need some tint2 for our desktop 😁 )

So what we have now, when we work with 2 monitors - some tasks are showing on the taskbar from the second monitor.
How to reproduce and understand it:
When we are opening  some app, for example  from menu LabWC open the thunar or nemo or firefox or xed or tilix or gmrun or telegram or any over ->  yes , we see that the task for this window is on the correct monitor , so where it was discovered,
**but**,
when we are moving this window to the second monitor - the task still remains in the first monitor on our panel where it was opened !
**and**,
only when 
**1.** we are maximizing this window in our first monitor, after that 
**2.** we are moving it to the second monitor, and at last 
**3.** on the second monitor we also are maximizing this window 
- just only in this case the task will be moved to our second monitor .

johanmalm:
Thanks for logging this issue.

Does this patch solve it? https://github.com/labwc/labwc/pull/745

boris3812:
so I have maked build , and i have :
labwc 0.6.1-1-g1995a33   ,  I understand this is with you lasst patch ...

the problem is also present , only difference what i see - is only with terminal VTE (tilix) when i'm writing something , only after that it goes to other output, for else apps windows are still on first output

wroots version in comunity is wlroots 0.16.1-2, if this is important

johanmalm:
When I opened a new application it sometimes showed in `waybar` on both outputs. I had never realised that this was wrong - but do now and the patch fixes it.

> **but**, when we are moving this window to the second monitor - the task still remains in the first monitor on our panel where it was opened ! 

I can't reproduce this. When I move a window, it always just appears in waybar on the output that the window is mostly on.

Would you be happy to do some fault finding and compile + test?
If so, I'd start by adding to fprintf()s  in view_discover_output() to get a feel for what's going on.
https://github.com/labwc/labwc/blob/1995a33df96b907a9a4ecd6580924d76e8720109/src/view.c#L783

You can use `%p` for pointers like so `fprintf(stderr, "%p\n", view->output);`
[...]

PR #745, opened 2023-01-31
foreign: fix initial leave/enter event bug

johanmalm:
Create foreign-toplevel-handle after initial view positioning to ensure leave/enter events are sent to the correct output.

Fixes: #744


Consolatis:
With #747 this shouldn't really matter anymore as we don't send any immediate `output_enter` event when creating the toplevel.

PR #746, opened 2023-02-01
src/foreign.c: refactor

Consolatis:
I have a another branch waiting that depends on this PR which moves the foreign-toplevel output notifications over to the scene graph emitted events.

Only actual change besides some refactoring:
- implement the `destroy` handler (which just removes the event listeners, required for the follow up PR).


johanmalm:
LGTM.
Nice refactoring.

PR #747, opened 2023-02-02
src/foreign.c: track outputs

Consolatis:
This is the 2nd iteration, the 1st one (using the `wlr_scene_buffer` `output_enter` / `output_leave` events) doesn't work because it will send an `output_leave` event whenever the window is completely occluded. It also required setting up a "click-through" `wlr_scene_buffer` on top of the view which required that its size kept getting synchronized with the actual view size.

This version now simply loops over all outputs and notifies the wlroots foreign_toplevel implementation about the current state.

The wlroots implementation takes care of
- de-duplicating the events (e.g. not send `output_enter` for the same output twice)
- listening to output destroy events, sending `output_leave` if the output was entered before

Benefits over our current implementation:
- will detect all outputs the view is on, not just the primary one
- will be called when
  - the view changed its position or size (including for maximize, fullscreen, tiling)
  - the output layout changes (e.g. the view position is still the same but now rendered on a different output)
  - completely separate from our window positioning / evacuation logic, that one tries to find the closest output (often the same as the one the window is mainly on) while we want to find all outputs intersecting the current window position / size here

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -345,6 +345,7 @@ void xdg_toplevel_decoration(struct wl_listener *listener, void *data);
 void xdg_surface_new(struct wl_listener *listener, void *data);
 
 void foreign_toplevel_handle_create(struct view *view);
+void foreign_toplevel_update_outputs(struct view *view);
 
 /*
  * desktop.c routines deal with a collection of views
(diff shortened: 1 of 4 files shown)


Consolatis:
This is more a `draft` -> `ready for review` dance than anything else :)
When thinking this through and looking at the emitted events I still think this PR fixes all foreign toplevel output tracking issues (from our side at least).

johanmalm:
LGTM

PR #748, opened 2023-02-05
Refactor view implementations following PR #720

johanmalm:
Ref: https://github.com/labwc/labwc/pull/720#issuecomment-1374831834

- [x] Add view `move_to_front` method to avoid passing a function pointer as a function argument and to keep xwayland specific code in one place.
- [x] Prefix xwayland view_impl functions with `xwayland_view_`

Do later:

- Add view_impl apply_size_constraints to replace `xwayland.c`'s `xwayland_apply_size_hints()`
- `s/xwayland_server_init/xwayland_init/`
- `{xwayland,xdg_toplevel}_view_unmap()` code is the same. Move to `view-impl-common.c`
- ~~Move desktop_move_to_back() to view.c for consistency~~

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -4,6 +4,7 @@
 #include "labwc.h"
 #include "node.h"
 #include "view.h"
+#include "view-impl-common.h"
 #include "workspaces.h"
 
 struct wlr_xdg_surface *
@@ -389,6 +390,7 @@ static const struct view_impl xdg_toplevel_view_impl = {
 	.set_fullscreen = xdg_toplevel_view_set_fullscreen,
 	.unmap = xdg_toplevel_view_unmap,
 	.maximize = xdg_toplevel_view_maximize,
+	.move_to_front = view_impl_move_to_front,
 };
 
 void
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -365,7 +365,6 @@ void foreign_toplevel_update_outputs(struct view *view);
  *              cannot assume this means that the window actually has keyboard
  *              or pointer focus, in this compositor are they called together.
  */
-
 void desktop_move_to_front(struct view *view);
 void desktop_move_to_back(struct view *view);
 void desktop_focus_and_activate_view(struct seat *seat, struct view *view);
--- a/include/view-impl-common.h
+++ b/include/view-impl-common.h
@@ -1,0 +1,14 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_VIEW_IMPL_COMMON_H
+#define __LABWC_VIEW_IMPL_COMMON_H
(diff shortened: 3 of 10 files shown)


src/xdg.c
```diff
@@ -244,6 +245,12 @@ xdg_toplevel_view_maximize(struct view *view, bool maximized)
 	wlr_xdg_toplevel_set_maximized(xdg_toplevel_from_view(view), maximized);
 }
 
+static void
+xdg_toplevel_view_move_to_front(struct view *view)
+{
+	view_impl_move_to_front(view);
+}
+
```
Consolatis, line 253: Why the extra wrap function here?

src/view.c
```diff
@@ -624,6 +624,16 @@ view_toggle_always_on_top(struct view *view)
...
 
+void
+view_move_to_front(struct view *view)
+{
+	assert(view);
+	if (view->impl->move_to_front) {
+		view->impl->move_to_front(view);
+		cursor_update_focus(view->server);
+	}
+}
```
Consolatis, line 635: The assert here changes the requirements of calling this function, I think all callers do check for `if (view)` so it should be fine but maybe we should document that requirement in the header file.

Side note:
This blurs the line between `desktop.c` and `view.c` a bit, I think usually `desktop.c` handles relations of views with each other, as in this case the stacking order. So I think I'd slightly prefer this kept as `desktop_move_to_front()` or maybe `desktop_move_view_to_front()` or just `desktop_raise_view()`.

include/view-impl.h
```diff
@@ -0,0 +1,14 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_VIEW_IMPL_INTERNAL_H
```
Consolatis, line 2: Should we use this opportunity to rename this to `view-common.h` or `view-impl-common.h` ?

Consolatis:
The CI Debian build error should be solved when rebasing this against latest master.

johanmalm:
All great points. Agree and have implemented.

src/xdg.c
johanmalm: Have removed - just calling the function in `view-impl-common.c` directly

src/view.c
johanmalm: Have restored `if (!view) { return }`
Have re-instated `desktop_move_to_front()`. Agree - it's more consistent to keep it in `desktop.c`

include/view-impl.h
johanmalm: `git mv view-impl.c view-impl-common.c` done.

johanmalm:
Just rebased. No other change.

Consolatis:
Sweet, one less `HAVE_XWAYLAND` around :)

PR #754, opened 2023-02-08
xwayland: Save natural geometry from initially maximized views

jlindgren90:
Revises the logic for setting the initial geometry of XWayland views, so that we can save the natural geometry correctly even if `request_maximize` or `request_fullscreen` events are received before `map()`.

This way, an application that sets a recommended "natural" size at startup, while also requesting maximized/fullscreen, will restore back down to the recommended natural size after being unmaximized/leaving fullscreen.

Tested working with Audacious (Qt5), VLC (Qt5), and Claws Mail (GTK2).
VLC also needs https://github.com/labwc/labwc/pull/753 to work correctly when starting in fullscreen mode.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -107,6 +107,38 @@ xwayland_surface_from_view(struct view *view)
 }
 
 static void
+ensure_initial_geometry(struct view *view)
+{
+	if (!view->w || !view->h) {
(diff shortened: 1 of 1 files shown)


johanmalm:
Again looks good. Thanks.

PR #756, opened 2023-02-09
menu: Compute usable area for the correct output

jlindgren90:
If a menu was opened via keybind while the cursor was on another, differently-sized output, the wrong usable area was used to position the menu.


Consolatis:
Didn't test but changes LGTM. Feel free to "Rebase and merge".

Nice catch!

PR #757, opened 2023-02-09
view: Use pending geometry for positioning logic

### LINKED ISSUE #671: Race condition while maximizing/unmaximizing windows
To reproduce:
Use this config file:
```
<?xml version="1.0"?>
<labwc_config>
  <keyboard>
    <default />
    <keybind key="A-Return">
      <action name="ToggleMaximize"></action>
    </keybind>

    <repeatRate>75</repeatRate>
    <repeatDelay>240</repeatDelay>
  </keyboard>
</labwc_config>
```

Open any window (at least, foot, various GTK programs, imv, and weston-terminal). Press and hold A-Enter, so the `ToggleMaximize` keybind repeats. Amidst maximizing and unmaximizing, the program will at some point forget its original unmaximized size, and begin toggling only between its original position and the top-left of the screen while always remaining the full size of the screen.

You may have to try pressing A-Enter several times--shorter, varied lengths tend to help. There also seems to be a repeat-rate dependence in this race condition--it's possible (but harder) to trigger it with a `repeatRate` of 60, and (so far in my tests) impossible to trigger with the default repeat rate. For what it's worth, my monitor runs at 60Hz, and I don't know if these repeat rates interact differently when the monitor runs at a different refresh rate.

While the window is maximizing/unmaximizing, various frames other than the expected two are also visible for fleeting moments, as well, such as titlebars that stick off the right edge of the screen (from a full-size but unmoved window).

jlindgren90:
- Make sure that `view->pending_move_resize` is always up-to-date for mapped views
- Use `view->pending_move_resize` rather than `view->x/y/w/h` for view positioning logic

By using the pending (most recently requested) dimensions instead of the current (displayed) dimensions of the view for positioning logic, we can avoid race conditions that might occur if we try to reposition a view while there is a pending move/resize that has not been processed yet.

This *should* fix #671 (I have not verified the fix since reproducing the issue looks quite involved).

It also greatly simplifies the logic that was committed in #680 since we can now use `view_center()` again, still without touching `view->w/h` before commit. I did test that there is no regression of that fix.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -243,7 +243,7 @@ view_output(struct view *view)
 	return output_from_wlr_output(view->server, wlr_output);
 }
 
-bool
+static bool
 view_compute_centered_position(struct view *view, int w, int h, int *x, int *y)
 {
 	struct output *output = view_output(view);
--- a/src/view.c
+++ b/src/view.c
@@ -310,10 +310,10 @@ view_store_natural_geometry(struct view *view)
 	 * natural_geometry width/height may still be zero in which case we set
 	 * some fallback values. This is the case with foot and Qt applications.
 	 */
-	if (wlr_box_empty(&view->current)) {
+	if (wlr_box_empty(&view->pending)) {
 		set_fallback_geometry(view);
 	} else {
-		view->natural_geometry = view->current;
+		view->natural_geometry = view->pending;
 	}
 }
 
@@ -322,8 +322,8 @@ view_center(struct view *view)
 {
 	assert(view);
 	int x, y;
-	if (view_compute_centered_position(view, view->current.width,
-			view->current.height, &x, &y)) {
+	if (view_compute_centered_position(view, view->pending.width,
+			view->pending.height, &x, &y)) {
 		view_move(view, x, y);
 	}
 }
@@ -417,8 +417,8 @@ view_apply_region_geometry(struct view *view)
 	geo.width -= margin.left + margin.right;
 	geo.height -= margin.top + margin.bottom;
 
-	if (view->current.width == geo.width
-			&& view->current.height == geo.height) {
+	if (view->pending.width == geo.width
+			&& view->pending.height == geo.height) {
 		/* move horizontally/vertically without changing size */
 		view_move(view, geo.x, geo.y);
 	} else {
@@ -439,8 +439,8 @@ view_apply_tiled_geometry(struct view *view, struct output *output)
 	}
 
 	struct wlr_box dst = view_get_edge_snap_box(view, output, view->tiled);
-	if (view->current.width == dst.width
-			&& view->current.height == dst.height) {
+	if (view->pending.width == dst.width
+			&& view->pending.height == dst.height) {
 		/* move horizontally/vertically without changing size */
 		view_move(view, dst.x, dst.y);
 	} else {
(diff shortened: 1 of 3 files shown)


jlindgren90:
Separately, I'd also suggest renaming:
- `view.x/y/w/h` to `view.geo.current`
- `view.pending_move_resize` to `view.geo.pending`
- `view.natural_geometry` to `view.geo.natural`

And use `struct wlr_box` for all three (I think this has been mentioned in passing before).
That would be a different pull request, though.

src/view.c
```diff
@@ -417,7 +418,8 @@ view_apply_region_geometry(struct view *view)
 	geo.width -= margin.left + margin.right;
 	geo.height -= margin.top + margin.bottom;
 
-	if (view->w == geo.width && view->h == geo.height) {
+	if ((int)view->pending_move_resize.width == geo.width
```
Consolatis, line 421: Could we change `pending_move_resize.{width,height}` to int instead? It doesn't seem to make sense to have them as unsigned while everything else is signed. It also doesn't seem to be used all that much (on latest master, didn't check for this PR):
```
$ egrep -R 'pending_move_resize[.->]*(height|width)' src/
src/xwayland.c: view->pending_move_resize.width = geo.width;
src/xwayland.c: view->pending_move_resize.height = geo.height;
src/xwayland.c:         view->w = view->pending_move_resize.width;
src/xwayland.c:         view->h = view->pending_move_resize.height;
src/xdg.c:                              view->pending_move_resize.width - size.width;
src/xdg.c:                              view->pending_move_resize.height - size.height;
src/xdg.c:      view->pending_move_resize.width = geo.width;
src/xdg.c:      view->pending_move_resize.height = geo.height
```

src/view.c
```diff
@@ -752,7 +755,12 @@ view_adjust_for_layout_change(struct view *view)
 		}
 	} else if (!view_apply_special_geometry(view)) {
 		/* reposition view if it's offscreen */
-		struct wlr_box box = { view->x, view->y, view->w, view->h };
+		struct wlr_box box = {
+			.x = view->pending_move_resize.x,
+			.y = view->pending_move_resize.y,
+			.width = view->pending_move_resize.width,
+			.height = view->pending_move_resize.height
+		};
```
Consolatis, line 763: This could just be a `struct wlr_box box = view->geo.pending` so another point for the refactor

src/view.c
jlindgren90: `wlr_box.{width,height}` are `int` so we'd get that change for free if we go that route.

src/view.c
jlindgren90: Yes indeed :)

If you prefer, I can submit that pull request (for the rename/refactor) and rebase this one on top.

src/view.c
Consolatis: Right, I'd prefer to not have the casting everywhere but I don't really mind if we first do the `wlr_box` refactoring or switch the struct member to `int` in the meantime.

src/view.c
Consolatis: Totally up to you :)

I'll put this PR into draft mode for now, but feel free to undraft it at any time.

Consolatis:
The code change itself looks good to me but I didn't test.

jlindgren90:
Rebased. Diff is a lot cleaner now after the `wlr_box` refactor.

Consolatis:
> Diff is a lot cleaner now after the `wlr_box` refactor.

Indeed, very much so. Thanks

Consolatis:
Btw, in my opinion the separation of view.c `apply_xxx_geometry()` and the actions which merely compute the sizes + deal with the global state and now the clear separation of `current` `pending` and `natural` geometry are two of the most important changes for the labwc codebase from a design and maintainability perspective. Had to be said somewhere so why not here.

PR #758, opened 2023-02-09
Add xdg-activation protocol

Consolatis:
This PR allows applications to activate themselves <b>if they provide a valid xdg_activation token</b> (e.g. raise to the top and get keyboard focus).

These tokens are given out by the xdg_activation protocol implemented in wlroots and can be configured by the client requesting the token in three ways:
- an "empty" token (apparently used to tell the compositor about "urgency")
- seat / input serial attached
- surface attached

Wlroots makes sure that
- <b>If</b> the client attached the seat / input serial: those two are valid.
- <b>If</b> the client attached a surface: that it has keyboard focus at the point where the request is finalized. There is a [patch](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/f6008ffff41f67e3c20bd8b3be8f216da6a4bb30) pending for backport to wlroots 0.16 that also allows valid tokens when the supplied surface had cursor focus.
- a token is only valid for 30 seconds after being given out

The token can then be used by the client or given to other clients by unspecified means (e.g. via environment variable or dbus) which then may use the token on their own surface and request activation.

We only handle the actual request activation part:
- If the seat is set on the token we know wlroots validated seat and input serial
- Thus, if no seat is set we deny the activation request so we don't have windows suddenly popping up and taking over the keyboard focus (focus stealing prevention)
- We should also check for the surface being set but we can't do that with wlroots 0.16 as it will reset the surface to `NULL` when it is destroyed (which is something that usually happens for notifications). Once we move to wlroots 0.17.x we can add the missing surface check because it provides a `new_token` signal. We can use it to attach further details to the token which are then verified later when we decide if we allow the activate request or not.

With this PR in place the following setup should activate windows:
- launching an URL in foot should activate the target application if it is already running (foot requests a proper token and then sets it as `XDG_ACTIVATION_TOKEN` environment var before spawning `xdg-open`)
- clicking on a `mako` notification with a `default` action defined (and when using the wlroots patch linked above) should request a proper token which is then given to the application starting the notification and can thus be used to activate itself

This protocol is still very much in the process of being implemented / finalized all over the place (e.g. GTK / QT / Firefox / notification daemons, ..) but we should do our part and remove labwc from the puzzle of potential issues causing this not to work.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -37,6 +37,7 @@
 #include <wlr/types/wlr_server_decoration.h>
 #include <wlr/types/wlr_subcompositor.h>
 #include <wlr/types/wlr_xcursor_manager.h>
+#include <wlr/types/wlr_xdg_activation_v1.h>
 #include <wlr/types/wlr_xdg_decoration_v1.h>
 #include <wlr/types/wlr_xdg_shell.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
@@ -226,6 +227,9 @@ struct server {
 	struct wl_listener input_inhibit_activate;
 	struct wl_listener input_inhibit_deactivate;
 
+	struct wlr_xdg_activation_v1 *xdg_activation;
+	struct wl_listener xdg_activation_request;
+
 	struct wl_list views;
 	struct wl_list unmanaged_surfaces;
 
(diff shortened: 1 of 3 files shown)


heroin-moose:
Nah, it does not work.

heroin-moose:
Clicking on mako does not swith focus.

heroin-moose:
Ah, ok.

jlindgren90:
I have not tested, but code LGTM.

Consolatis:
> Click on mako does not switch focus.

The wlroots patch has been backported to the 0.16 branch and thus also queued for a future wlroots 0.16.2 release.
`meson subprojects update wlroots` and rebuilding this branch should now make it work (if the application supports it as well).

PR #759, opened 2023-02-09
view: Use wlr_box for current/pending geometry

jlindgren90:
Wrap up separate `x/y/width/height` fields into `struct wlr_box` for convenience.
No functional change.

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -2,6 +2,7 @@
 #ifndef __LABWC_SSD_INTERNAL_H
 #define __LABWC_SSD_INTERNAL_H
 
+#include <wlr/util/box.h>
 #include "ssd.h"
 
 #define FOR_EACH(tmp, ...) \
@@ -40,10 +41,7 @@ struct ssd {
 	 * don't update things we don't have to.
 	 */
 	struct {
-		int x;
-		int y;
-		int width;
-		int height;
+		struct wlr_box geometry;
 		struct ssd_state_title {
 			char *text;
 			struct ssd_state_title_width active;
@@ -51,7 +49,7 @@ struct ssd {
 		} title;
 	} state;
 
-	/* An invisble area around the view which allows resizing */
+	/* An invisible area around the view which allows resizing */
 	struct ssd_sub_tree extents;
 
 	/* The top of the view, containing buttons, title, .. */
--- a/src/resistance.c
+++ b/src/resistance.c
@@ -37,10 +37,9 @@ resistance_move_apply(struct view *view, double *x, double *y)
 {
 	struct server *server = view->server;
 	struct wlr_box mgeom, intersection;
-	struct wlr_box vgeom = {.x = view->x, .y = view->y, .width = view->w,
-		.height = view->h};
-	struct wlr_box tgeom = {.x = *x, .y = *y, .width = view->w,
-		.height = view->h};
+	struct wlr_box vgeom = view->current;
+	struct wlr_box tgeom = {.x = *x, .y = *y, .width = vgeom.width,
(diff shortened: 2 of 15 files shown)


src/resistance.c
```diff
@@ -37,26 +37,25 @@ resistance_move_apply(struct view *view, double *x, double *y)
 {
 	struct server *server = view->server;
 	struct wlr_box mgeom, intersection;
-	struct wlr_box vgeom = {.x = view->x, .y = view->y, .width = view->w,
-		.height = view->h};
-	struct wlr_box tgeom = {.x = *x, .y = *y, .width = view->w,
-		.height = view->h};
+	struct wlr_box vgeom = view->current;
+	struct wlr_box tgeom = {.x = *x, .y = *y, .width = vgeom.width,
+		.height = vgeom.height};
```
Consolatis, line 42: Interesting variable names, I think I never looked into this file even once.

src/view.c
```diff
@@ -308,13 +310,10 @@ view_store_natural_geometry(struct view *view)
 	 * natural_geometry width/height may still be zero in which case we set
 	 * some fallback values. This is the case with foot and Qt applications.
 	 */
-	if (!view->w || !view->h) {
+	if (wlr_box_empty(&view->current)) {
```
Consolatis, line 313: This does change the logic, it was ignoring `.x` and `.y` before.
Not sure if that matters though.

src/xwayland.c
```diff
@@ -109,24 +109,19 @@ xwayland_surface_from_view(struct view *view)
 static void
 ensure_initial_geometry(struct view *view)
 {
-	if (!view->w || !view->h) {
+	if (wlr_box_empty(&view->current)) {
```
Consolatis, line 112: Same here

include/ssd-internal.h
```diff
@@ -2,6 +2,7 @@
 #ifndef __LABWC_SSD_INTERNAL_H
 #define __LABWC_SSD_INTERNAL_H
 
+#include <wlr/util/box.h>
```
Consolatis, line 5: Do we actually need this?
A forward declaration should be enough if all users of this header include box.h themselves.

Consolatis:
Nice job! Didn't spot a single x <-> y conversion or similar.
Besides the `wlr_box_empty()` calls now not ignoring x and y anymore this seems fine to me.

src/view.c
jlindgren90: `wlr_box_empty` ignores x and y too:
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/util/box.c#L37

include/ssd-internal.h
jlindgren90: I think the `struct wlr_box geometry;` line is a compile error (if the header were to be compiled on its own) without this.

src/view.c
Consolatis: Ohh. I am sorry, should have checked that. Thanks for the pointer.

include/ssd-internal.h
Consolatis: Not sure what you mean by "compiled on its on" but in any case the same would be true for e.g. `struct view` or `struct wlr_scene_node` then as they are only forward-declared in `ssd.h` as well (which is included by `ssd-internal.h`).

Consolatis:
All issues resolved, LGTM but didn't test.
Feel free to "Rebase and merge" when you are satisfied with the PR.

Thanks for the work you put into this.

jlindgren90:
> All issues resolved, LGTM but didn't test. Feel free to "Rebase and merge" when you are satisfied with the PR.

Will do. Thanks for the review!

Issue #760, opened 2023-02-09
`foot --maximized` starts with ssd off screen and gap the size of the ssd at the bottom


Flrian:
Started with one of the commits from the last 24h I think.

Consolatis:
Can confirm.
@jlindgren90 seems we missed something somewhere. Thoughts?

jlindgren90:
I'll take a look. I expect it's probably related to one of my changes, not sure which one.

jlindgren90:
Bisecting points to 103a2a8f6c38a8eb664160e369477cad98003598 as the culprit. Still working to understand why.

Consolatis:
```
00:00:06.526 [../src/view.c:504] apply_special
00:00:06.527 [../src/view.c:513] apply_special non-special
00:00:06.527 [../src/view.c:504] apply_special
00:00:06.527 [../src/view.c:506] apply_special maximize
00:00:06.630 [../src/xdg.c:337] First time xdg map, enabling decorations: True
00:00:06.639 [../src/xdg.c:305] Adjusting view by 1,2
```
It seems we are maximizing before the map is actually called. Not sure why

Consolatis:
Right.. `handle_request_maximize()` in xdg.c
```
00:00:02.595 [../src/view.c:504] [0xaaaaab92c300] apply_special
00:00:02.595 [../src/view.c:513] [0xaaaaab92c300] apply_special non-special
00:00:02.595 [../src/xdg.c:172] [0xaaaaab92c300] maximize request
00:00:02.595 [../src/view.c:504] [0xaaaaab92c300] apply_special
00:00:02.595 [../src/view.c:506] [0xaaaaab92c300] apply_special maximize
00:00:02.668 [../src/xdg.c:338] [0xaaaaab92c300] First time xdg map, enabling decorations: True
00:00:02.668 [../src/xdg.c:306] Adjusting view by 1,26
```
Does it need a `view->mapped` check? Or should it create the SSD?

jlindgren90:
> It seems we are maximizing before the map is actually called. Not sure why

Yes. It seems we would maximize first, then center the view in `map()`. 103a2a8f6c38a8eb664160e369477cad98003598 isn't really at fault because it only made the center *actually work*.

The easy fix seems to be add an `if (view->been_mapped)` to `handle_request_maximize()` and `handle_request_fullscreen()`. But in my testing, this causes a visual glitch: the view appears briefly in a non-maximized geometry and is then maximized afterward.

[...]

PR #765, opened 2023-02-09
xdg: Fix positioning of initially-maximized views

jlindgren90:
This is a minimal fix for the regression described in #760. It basically restores the status quo prior to 103a2a8f6c38a8eb664160e369477cad98003598.

- Don't overwrite pending size in `map()` if it was already set
- Don't reposition view in `map()` if maximized/fullscreen

Also, as future-proofing in case we one day allow initially-tiled views, replace explicit maximized/fullscreen checks with `view_is_floating()`.

I haven't attempted to address the separate (pre-existing) issue of `foot --maximized` always appearing on the top-left output (regardless of cursor position). There may also be issues with saving/restoring natural geometry for initially-maximized xdg-shell views -- I suspect that has never worked correctly (hence the `set_fallback_geometry()` workaround in `view.c`).

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -340,16 +340,21 @@ xdg_toplevel_view_map(struct view *view)
 		 * view_set_fullscreen/view_maximize() below). "Current"
 		 * dimensions remain zero until handle_commit().
 		 */
-		view->pending.width = xdg_surface->current.geometry.width;
-		view->pending.height = xdg_surface->current.geometry.height;
+		if (wlr_box_empty(&view->pending)) {
+			view->pending.width =
+				xdg_surface->current.geometry.width;
+			view->pending.height =
+				xdg_surface->current.geometry.height;
+		}
 
-		position_xdg_toplevel_view(view);
 		if (!view->fullscreen && requested->fullscreen) {
 			view_set_fullscreen(view, true,
 				requested->fullscreen_output);
 		} else if (!view->maximized && requested->maximized) {
 			view_maximize(view, true,
 				/*store_natural_geometry*/ true);
+		} else if (view_is_floating(view)) {
+			position_xdg_toplevel_view(view);
 		}
 
 		view_moved(view);
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -534,7 +534,7 @@ map(struct view *view)
 	if (!view->been_mapped) {
 		view_set_decorations(view, want_deco(xwayland_surface));
 
-		if (!view->maximized && !view->fullscreen) {
+		if (view_is_floating(view)) {
 			set_initial_position(view, xwayland_surface);
 		}
 
@@ -549,7 +549,7 @@ map(struct view *view)
 		view->been_mapped = true;
 	}
 
-	if (view->ssd_enabled && !view->fullscreen && !view->maximized) {
+	if (view->ssd_enabled && view_is_floating(view)) {
 		top_left_edge_boundary_check(view);
 	}
 


src/xdg.c
```diff
@@ -340,10 +340,17 @@ xdg_toplevel_view_map(struct view *view)
...
-		view->pending.height = xdg_surface->current.geometry.height;
+		if (wlr_box_empty(&view->pending)) {
+			view->pending.width =
+				xdg_surface->current.geometry.width;
+			view->pending.height =
+				xdg_surface->current.geometry.height;
+		}
+
+		if (view_is_floating(view)) {
+			position_xdg_toplevel_view(view);
```
Consolatis, line 351: I think the `view_center()` call in `position_xdg_toplevel_view()` will overwrite `view->pending`

jlindgren90: That's intentional in the case of floating views, because we really do want to change the geometry in that case.

jlindgren90: The earlier `wlr_box_empty()` is just a safeguard to ensure we don't undo some previous, possibly still pending, geometry change and overwrite it with the current surface geometry (which may be out of date).

jlindgren90:
@Consolatis Did you have remaining concerns with this change, or can I mark it ready again?

Consolatis:
Lets dial back a bit (mostly me :)).

So.. the issue is that maximize is called before the map.
My diff fixed that part but also moves view_center() (which just updates the x and y position of `pending`) into an else branch which seems wrong as `pending` will be used by the following `view_maximize()` call in map() to set the natural geometry.

I can't right follow why this PR fixes #760

src/xdg.c
Consolatis: Right, but as this is the first map call there should be no pending changes?

Consolatis: And position_xdg_toplevel only sets x and y of pending, based on the width and height of the surface.
Which should then be used as natural geometry once we call view_maximize() (within the first map branch).

Consolatis: view_set_decoration() however, will call `apply_xxx_geometry()` and in this case the `apply_maximized_geometry()` variant (due to the earlier call to request_maximize before the map). I am missing a link on how any of this relates to `view->pending`.

jlindgren90: `handle_request_maximize()` or `handle_request_fullscreen()` could be called before first map and in that case would have updated `view->pending` already.

jlindgren90:
This fixes #760 by not calling `view_center()` if `handle_request_maximize()` is called before first map. In that case `view->maximized` is already `true` when the view is mapped. The `if (!view->maximized && requested->maximized)` branch is not entered in that case.

On second look, I think I would revise this to move the `view_is_floating()` check to an `else if` branch, for the sake of correctness:
```
		if (!view->fullscreen && requested->fullscreen) {
			view_set_fullscreen(view, true,
				requested->fullscreen_output);
		} else if (!view->maximized && requested->maximized) {
			view_maximize(view, true,
				/*store_natural_geometry*/ true);
		} else if (!view_is_floating(view)) {
			position_xdg_toplevel_view(view);
		}
```

If we went with your alternative fix of delaying `view_maximize()` until the first map, then the `if (view_is_floating())` check I've added here might not be necessary, and we could go with the simple `else` as you did. But that introduces a visual glitch as I mentioned in #760.

There are definitely still some issues with initiall-maximized/fullscreen views that this PR doesn't address, but it's currently my best attempt at fixing the particular regression mentioned in #760, while avoiding any other new side effects.

src/xdg.c
Consolatis: Alright, after some much needed sleep and some more debugging this all starts to make sense now:
- `wlr_box_empty()` does nothing to solve the actual issue here because the sizes are the same in this case, e.g. whatever the pre-map maxmize_request sent to the client via configure.
- `view_center()` calculates the wrong result, it indeed uses the ssd border thickness but then does not add it afterwards (which is required because `view_move()` addresses the actual surface, not the SSD (which is using negative positions in relation to the view tree)
- `position_xdg_toplevel_view()` "fixes" that by manually modifying `view->current.{x,y}` but does <b>not</b> call `view_move()` which causes the coordinates to be out of sync.
- Using `view_move()` in `position_xdg_toplevel_view()` instead also fixes the issue but is a bad solution as we don't have to call it in the first place (so this PR does the absolutely correct thing in that regard).
- This btw also results in usual views not actually being quite centered when spawned, but that is something for another PR

Consolatis:
> If we went with your alternative fix of delaying view_maximize() until the first map, then the if (view_is_floating()) check I've added here might not be necessary, and we could go with the simple else as you did. But that introduces a visual glitch as I mentioned in https://github.com/labwc/labwc/issues/760.

I am very much convinced now that delaying `view_maximize()` is a bad idea because it will indeed cause visual glitches and solves a  real issue when called before the first map. I'll add a follow-up PR which should fix the initial output selection for xdg views starting maximized and adds a comment on there explaining it in detail so somebody new to the codebase (or me in a few months) understands how this works.

Consolatis:
Thanks for the fix and sorry for the behavior earlier. It was, frankly speaking, simply unprofessional and annoying. I'll try to not interact in issues / PRs when being sleep deprived in the future :)

src/xdg.c
jlindgren90: Nice analysis. Writing directly to view->current did look "sketchy" to me but I didn't look at it further. I think xwayland.c has some similar code which might also be wrong in that case.

jlindgren90:
No worries. Thanks for the review!

Issue #456, opened 2022-07-18
Name of possible keyboard keys


dglava:
I wanted to map some keys on my keyboard to certain actions, but I couldn't find a list of all the keyboard keys to use. For example: I wanted to bind something to Page Up and I didn't know what the key's name was (inside labwc). I quickly checked my old Openbox config and found out it's `Page_Up` and it appears to work in labwc as well.

The [Openbox documentation](http://openbox.org/wiki/Help:Bindings#Keys) recommends to use `xev` to find out, but that's an X tool. Is there a Wayland tool which could be used? Alternatively, is it possible to include a list of all possible key names with labwc? I reckon that would be more user friendly in terms of documentation.

Consolatis:
labwc is using libxkbcommon which has all the keysyms [defined in a header](https://github.com/xkbcommon/libxkbcommon/blob/master/include/xkbcommon/xkbcommon-keysyms.h).

`xev` (which runs via xwayland) or `wev` (wayland native) should both do the trick.
Maybe we should mention `xev` and `wev` somewhere in the man pages.

johanmalm:
Or on the 'getting-started' page on the website. Appreciate suggestions.

Consolatis:
https://labwc.github.io/getting-started.html#step-3---set-a-shortcut-to-a-launcher

Maybe something like "You can use either `xev` (widely available, runs via xwayland) or `[wev](https://some.link.to/sources)` to figure out the name of a key."

johanmalm:
Solved by #766

PR #766, opened 2023-02-10
docs: add hint for using wev

dglava:
This probably solves this issue: #456 

I don't know what the policy about linking in man-pages is, but perhaps it would also be okay to link to the wev source page? Perhaps it would even be useful to mention the `xkbcommon-keysyms.h` header file.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -168,7 +168,8 @@ The rest of this man page describes configuration options.
 	Define a key binding in the format *modifier-key*, where supported
 	modifiers include S (shift); C (control); A (alt); W (super). Unlike
 	Openbox, multiple space-separated key combinations and key-chains are
-	not supported.
+	not supported. The application "wev" (wayland event viewer) is packaged
+	in a lot of distributions and can be used to view all available keynames.
 
 *<keyboard><keybind key=""><action name="">*
 	Keybind action. See labwc-action(5)


docs/labwc-config.5.scd
```diff
@@ -169,6 +169,7 @@ The rest of this man page describes configuration options.
 	modifiers include S (shift); C (control); A (alt); W (super). Unlike
 	Openbox, multiple space-separated key combinations and key-chains are
 	not supported.
+	Use wev (wayland event viewer) to view all available keynames.
```
Consolatis, line 172: Suggest moving the start of the sentence to the previous line, potentially breaking to the next one.
E.g. something like this:

	Define a key binding in the format *modifier-key*, where supported
	modifiers include S (shift); C (control); A (alt); W (super). Unlike
	Openbox, multiple space-separated key combinations and key-chains are
	not supported. The application "wev" (wayland event viewer) can be
	used to view all available keynames.

Consolatis (commented):
Thanks for the PR, improving user documentation is always a good thing.

> I don't know what the policy about linking in man-pages is, but perhaps it would also be okay to link to the wev source page?

I am not quite sure either.
Maybe there is special syntax available for cross references to other man pages. Or we could mention searching in the package repositories, wev is [widely available](https://repology.org/project/wev/versions).
Thoughts @johanmalm ?

Some further feedback below:

docs/labwc-config.5.scd
dglava: I have moved it and added a hint to look for wev in the distribution's repositories.

johanmalm:
Thanks! Looks good.
I've also updated the website https://github.com/labwc/labwc.github.io/commit/37946271a9aeecdf51bb9264ae38107a4ba4472f
With man pages, I prefer to only refer to other man pages with the syntax `xkeyboard-config(7)`
For further description of other clients + tips&tricks + weblinks I think it's cleaner to use the website.

PR #767, opened 2023-02-10
src/xdg.c: spawn initially maximized views on cursor output

Consolatis:
\+ add a small comment whats going on there.

@jlindgren90 pointing out errors and thoughts about the "Open question" part would be much appreciated.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -146,6 +146,8 @@ bool view_is_always_on_top(struct view *view);
 bool view_is_tiled(struct view *view);
 bool view_is_floating(struct view *view);
 void view_move_to_workspace(struct view *view, struct workspace *workspace);
+void view_move_to_output(struct view *view, struct output *output, bool apply_layout);
+
 void view_set_decorations(struct view *view, bool decorations);
 void view_toggle_fullscreen(struct view *view);
 void view_adjust_for_layout_change(struct view *view);
--- a/src/action.c
+++ b/src/action.c
@@ -303,6 +303,10 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_DEBUG:
 			debug_dump_scene(server);
+			if (view) {
+				wlr_log(WLR_ERROR, "moving to cursor output");
+				view_move_to_output(view, NULL, /* apply_layout */ true);
+			}
 			break;
 		case ACTION_TYPE_EXECUTE:
 			if (!arg) {
--- a/src/view.c
+++ b/src/view.c
@@ -630,6 +630,67 @@ view_move_to_workspace(struct view *view, struct workspace *workspace)
 	}
 }
 
+void
+view_move_to_output(struct view *view, struct output *output, bool apply_layout)
+{
+	assert(view);
+	if (!output) {
+		/* Get the output where the cursor is currently on */
+		struct wlr_output *wlr_output = wlr_output_layout_output_at(
+			view->server->output_layout,
+			view->server->seat.cursor->x,
+			view->server->seat.cursor->y);
+		output = output_from_wlr_output(view->server, wlr_output);
+	}
+	if (!output) {
+		wlr_log(WLR_ERROR, "Can't move view to output: No outputs available");
+		return;
+	} else if (output == view->output || output->wlr_output == view->fullscreen) {
+		wlr_log(WLR_ERROR, "Refusing to move view to same output");
+		return;
+	}
+
+	/*
+	 * TODO:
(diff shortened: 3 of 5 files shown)


jlindgren90:
I can't discuss in too much depth right now but initial thoughts are:
1. This probably works but does seem a bit roundabout. In my own master branch, I went in a different direction of setting view->output directly before calling view_maximize(), and modifying view_maximize() to use view->output.
2. We probably need a similar fix for fullscreen as well.

Consolatis:
This pattern of modifying `view->current.{x,y}` is used in `xwayland.c` and `xdg.c` and basically just forces `view_discover()` to use a defined output.

I completely agree that this is a complete hack and abuse of implementation internals.
In the meantime, due to our current use of this pattern and the missing alternative, what do you think about creating a function like `view_move_to_output(struct view *view, struct ouput *output)` so we confine that pattern into a single place?

Then the current users could use that function instead and all the hacky details are hidden from the outside until we have a better strategy to deal with it.

> We probably need a similar fix for fullscreen as well.

Yes, indeed. If everbody agrees with the `view_move_to_output()` idea I could add that to `request_fullscreen()` as well and also replace the existing uses with the new function call.

jlindgren90:
Sounds fine to me as a short-term fix.

johanmalm:
Just looking at this fairly quickly, so might be barking up the wrong tree, but would something like this work:

In `handle_request_maximize()` if `(!view->been_mapped)` then pass store_natural_geometry=false to `view_maximize()`.

Then in `xdg_toplevel_view_map()` in the `!view->been_mapped` section check for view->maximized and set the natural geometry there based on `output_usable_area_from_cursor_coords()` and `xdg_surface->current.geometry.{width,height}` (because in the map handler we are allowed to read those variables).

jlindgren90:
For storing the natural geometry, another idea would be to set view.pending to the surface width/height prior to calling view_maximize(). That's what we do for XWayland views (see ensure_initial_geometry()). (XWayland also sets view.current but I don't think that's necessary anymore.)

Consolatis:
> In `handle_request_maximize()` if `(!view->been_mapped)` then pass store_natural_geometry=false to `view_maximize()`.
> Then in `xdg_toplevel_view_map()` in the `!view->been_mapped` section check for view->maximized and set the natural geometry there based on `output_usable_area_from_cursor_coords()` and `xdg_surface->current.geometry.{width,height}`

That would set the natural geometry to the maximized width / height, e.g. when the user would then un-maxmize foot it would basically not change position or size but just not be maximized anymore.

> another idea would be to set view.pending to the surface width/height prior to calling view_maximize().

Just tried that: `[../src/xdg.c:201] handle_request_maximize: Set pending to 0x0`
So it seems there are no previous commit calls (by foot at least) which we could use to define the natural geometry.

jlindgren90:
> Just tried that: `[../src/xdg.c:201] handle_request_maximize: Set pending to 0x0` So it seems there are no previous commit calls (by foot at least) which we could use to define the natural geometry.

Okay, thanks for checking. I guess I am thinking in X11 terms and relatively ignorant of how xdg-shell works in this area.

Consolatis:
I've added a new `view_move_to_output(view, output, /* apply_geometry */)` function and replaced the usage of the position hack in xwayland.c and xdg.c. Ignore the build warning for Void, its unrelated to the PR.

Turns out that function was actually pretty easy to enhance to allow moving usual views to different outputs as well (which is what the `apply_geometry` arg is for). In case somebody wants to test that: I've bound the new function to the `Debug` action.
It will move the active view to the output where the cursor is currently on while preserving the state (e.g. maximized / snapped to edge / fullscreen). This might become pretty handy when using VNC clients or similar which only have access to a single screen. I recommend not trying that with region tiled views right now though as the region they are using is attached to a single output and thus need to be replaced with the same region on the target output. Will implement that later.

This branch is based on two more commits which fix issues I detected along the way, I've split those into their own PRs and will rebase this branch once they are merged. Only the last two commits are related to this PR.

Consolatis:
Suggestions on how to move forward?
Just adding `struct output` to `view_maximize()` and using output_from_cursor in the request_maximize `!view->been_mapped` handler?

jlindgren90:
> Suggestions on how to move forward? Just adding `struct output` to `view_maximize()` and using output_from_cursor in the request_maximize `!view->been_mapped` handler?

Sounds good to me, at least short-term.

Consolatis:
Replaced by
- #798

PR #770, opened 2023-02-11
src/view.c: view_is_floating() add missing tiled_region_evacuate check

Consolatis:
Also
- remove static inline (e.g. trust the compiler to optimize)
- move from `include/view.h` to `src/view.c`

As mentioned in https://github.com/labwc/labwc/pull/753#discussion_r1102841140 there is one other case where we do `static inline` + function implementation in a header file: https://github.com/labwc/labwc/blob/master/include/common/list.h

But in that case I would argue that we don't have a proper place to put it, creating a new `.c` file just to define that single function seems a bit excessive.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -116,13 +116,6 @@ struct xdg_toplevel_view {
 	struct wl_listener new_popup;
 };
 
-static inline bool
-view_is_floating(struct view *view)
-{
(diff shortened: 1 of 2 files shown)


johanmalm:
Thanks!

Yes, agree that `wl_list_append()` is a bit special :smile:

PR #771, opened 2023-02-11
src/cursor.c: cursor names: use absence of "grab" to trigger fallback

Consolatis:
This is necessary because wlroots backported a patch containing
cursor name aliases to the 0.16 branch. That internal mapping
does not alias "grab" to "grabbing" because it is not technically
part of the cursor-names spec.

To keep the "grab" cursor working on systems with no cursor theme
available (or in an unusual location) change the trigger for our
internal cursor name aliasing to "grab".

See wlroots commit [1e345c74c3ca5d2575aa975d4d9af87cd0c7ba13](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/1e345c74c3ca5d2575aa975d4d9af87cd0c7ba13)
Alias existing cursor defaults to cursor-spec cursor names

for further details.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -1090,7 +1090,7 @@ cursor_init(struct seat *seat)
 	wlr_xcursor_manager_load(seat->xcursor_manager, 1);
 
 	if (wlr_xcursor_manager_get_xcursor(seat->xcursor_manager,
-			cursors_xdg[LAB_CURSOR_DEFAULT], 1)) {
+			cursors_xdg[LAB_CURSOR_GRAB], 1)) {
 		cursor_names = cursors_xdg;
 	} else {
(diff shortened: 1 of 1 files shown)


johanmalm:
Is it worth adding some comments in `cursor.c` to explain why we test with "grab" etc. to help future devs?

Consolatis:
Yes, that sounds like a good idea. We should likely have added that comment right from the start when adding our internal alias tables. I'll look into it later and do another PR for that.

johanmalm:
Yes, when I started reviewing it I was like "what's going on here"!
Agree, relates to original code so didn't think it was worth delaying merging the PR.

Consolatis:
- #773

PR #774, opened 2023-02-12
view: make view_move_resize always work

Consolatis:
It was not working before because in the case of wayland we are only dealing with sizes as wayland has no notion of a global position. A wayland client would thus not necessarily respond to a configure request which sets the same size again. This causes us to also not apply a new position set in view->pending because there may be no commit from the client in those cases.

We previously worked around this issue in some parts of the code to check our new sizes against the pending ones and if they were the same we would call view_move instead. That had two issues:

- Not all parts of the code did that which could end up delaying the positioning either to the next completely unrelated xdg commit event or to the next view_move call

- The code started to repeat itself, e.g. the same condition with calls to either view_move or view_move_resize based on the result

This patch fixes it by doing the check in the xdg configure handler instead. Xwayland is unaffected by this issue as we are always configuring a xwayland client with both, position and size.

---
Void build error is unrelated to this PR

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -162,7 +162,6 @@ view_moved(struct view *view)
 	}
 }
 
-/* N.B. Use view_move() if not resizing. */
 void
 view_move_resize(struct view *view, struct wlr_box geo)
 {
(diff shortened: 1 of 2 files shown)


jlindgren90:
LGTM. I have not tested but the logic looks sound. I wanted to submit a similar PR eventually, so 👍

johanmalm:
Great work.

PR #775, opened 2023-02-12
src/xdg.c: sync move with pending resize 

Consolatis:
Before this patch, configuring a surface with a new size,
immediately followed up by one or more view_move() calls
would move the surface to the new coordinates immediately
without waiting for the resize. This caused visual glitches
when for example dragging a maximized window: the position
would change but the size was still that of a maximized
window.

This patch fixes that by just ignoring view_move() requests
(but still updating view->pending) if there is a configure
request pending. Once the client commit comes in the new
size will be applied as usual.

---
This PR is depends on
- #774

Will rebase and "undraft" once merged.

---
Void build error is unrelated to this PR

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -203,13 +203,17 @@ handle_set_app_id(struct wl_listener *listener, void *data)
 static void
 xdg_toplevel_view_move(struct view *view, int x, int y)
 {
-	view->current.x = x;
-	view->current.y = y;
-
 	/* override any previous pending move */
 	view->pending.x = x;
 	view->pending.y = y;
 
+	/* Syncs moves with resizes  */
+	if (view->pending_configure_serial > 0) {
+		return;
+	}
+
+	view->current.x = x;
+	view->current.y = y;
 	view_moved(view);
 }
 


Consolatis:
Same issue seems to exists for xwayland views:
- Maximize view
- Use the mouse to move the view
- Can be seen best when running on two screens, maximizing on the left screen, then starting the move somewhere on the right on the titlebar. It should pop up on the second screen for a moment (e.g. move while keeping the old size) until the configure went through and the new size is used

jlindgren90:
> Edit: The xwayland `move()` function also seems to use the `xwayland_surface` sizes for the configure request, shouldn't it use the sizes from `view->pending` instead?

I agree it should use `view->pending`. Seems like there would be a race condition otherwise.

jlindgren90:
Just an idea, what if we were to eliminate `view->impl->move` entirely and implement view_move in terms of view_move_resize?

```
void
view_move(struct view *view, int x, int y)
{
	assert(view);
	view_move_resize(view, (struct wlr_box){
		.x = x, .y = y,
		.width = view->pending.width,
		.height = view->pending.height
	});
}
```

Consolatis:
> Just an idea, what if we were to eliminate view->impl->move entirely and implement view_move in terms of view_move_resize?

Would be fine by me, I don't particularly mind either way. I guess we could potentially remove the move implementation in xwayland then as well (if it isn't also used by some x11 callback that is).

jlindgren90:
Anyway, this LGTM. The view_move/view_move_resize refactor can happen later if we want it.

PR #778, opened 2023-02-15
view: Unify view_move()/view_move_resize()

jlindgren90:
`view->impl->move()` is a specific case of `view->impl->configure()`. To reduce code duplication, we can use `view->impl->configure()` for pure moves (without resize) as well.

xwayland's `move()` function also possibly contained a race condition when there was a pending resize, as it used the current surface width/height rather than the pending width/height. Now we use the pending width/height so there is no race condition.

This may resolve the visual glitch @Consolatis mentioned in https://github.com/labwc/labwc/pull/775#issuecomment-1426973649. But I wasn't able to reproduce the glitch myself, so I can't verify.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -26,7 +26,6 @@ struct view_impl {
 	void (*close)(struct view *view);
 	const char *(*get_string_prop)(struct view *view, const char *prop);
 	void (*map)(struct view *view);
-	void (*move)(struct view *view, int x, int y);
 	void (*set_activated)(struct view *view, bool activated);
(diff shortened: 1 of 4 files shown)


Consolatis:
Looking good, thanks.
I'll test this later and then merge if everything works as expected.

Consolatis:
Worked perfectly fine when tested and now the xwayland move is delayed as well until the size has been updated which indeed fixes the xwayland issue I was seeing before. It might only show up on low-end hardware + I've only seen it before when testing nested.

Issue #14, opened 2021-01-15
Grab openbox rc.xml file if it exists.


FlaviusHouk:
If this compositor is influenced by openbox and even migh works with it's config, should labwc take openbox config if there is one? For example there might be additional entry in config_dirs array (dir.c, line 21) after labwc entry. If there is special entry, for labwc take it, if no - al least check available openbox one.

johanmalm:
I prefer not to. 

Early commits did indeed respect the .config/openbox/rc.xml and XDG config dir equivalents, but I found it got confusing and prefer to keep it simple and explicit. 

Also, autostart and environment are likely to be pretty different from those used under X11 and will need to be handcrafted, so to use an rc.xml in a different directory is probably not all that helpful.

Hope that’s okay.

FlaviusHouk:
Just would like to verify that this is an intended behavior.

johanmalm:
Thanks

Issue #12, opened 2021-01-01
fluxbox


xenoterracide:
I used openbox, blackbox, and fluxbox, way back >15 years ago. I switched largely because I thought mounting removable media was a pain. Of the 3, I thought fluxbox was the best, I don't remember if it's how the right click menus worked, I think it had tabbed windows, and you could easily trigger a run prompt. I know it had a taskbar. It was still of course very minimal. Do you have any plans to investigate its features to make certain they're available?

johanmalm:
Sorry, missed this one.
I have used both openbox and fluxbox.
I decided to use openbox spec (a) in order to avoid re-inventing a spec; (b) because it's pretty 'no-frills' and (c) because I think it's widely adopted (e.g. by LXDE, raspberry pi, Bunsenlabs, etc).
I do quite like the fluxbox menu syntax, but think it's simpler to just stick with the openbox xml format.
Regarding run-prompts, taskbars and the like, I intend to use other applications for this.
The aim is to keep the code base pretty small+simple and just focus on being a good stacking window compositor (albeit with a menu :smile:)

Issue #11, opened 2020-12-02
Click dragging top/left/top-left corner causes jiggles


kelppo:
As of https://github.com/johanmalm/labwc/commit/ca24c215eea64430a69fd870f4e9abc47d9f671e, click dragging top or left corner of a window causes jiggles on the opposite corner, click dragging top-left corner causes the entire window to jiggle.

johanmalm:
Does it work if you build with the master branch of wlroots (in a 'subprojects' directory)

kelppo:
Nope.

$ ldd /home/ketsui/labwc/build/labwc | grep wlroots  
libwlroots.so.7 => /home/ketsui/labwc/build/subprojects/wlroots/libwlroots.so.7 (0x00007f22aad75000)

johanmalm:
Which commit actually changes the behaviour? 
The one labelled “Resize via dragging corners”?
Or has it always been like it?

kelppo:
I tested 0347e5d9032be79c6318a0efff5c14be5e146995 (the one before "Resize via dragging corners") and click dragging the top or left border causes jiggling on the opposite side.

johanmalm:
Thanks. I’ll take a look to see if I can figure it out.

johanmalm:
That should be it

PR #779, opened 2023-02-15
ssd: Fix segfault when disabling output with fullscreen view

jlindgren90:
`wlr_output_layout_get()` returns `NULL` for disabled outputs.

Backtrace (abbreviated):

     #0  lab_wlr_output_layout_layout_coords at ../src/ssd/ssd_extents.c:33
     #1  ssd_extents_update at ../src/ssd/ssd_extents.c:133
     #2  ssd_extents_create at ../src/ssd/ssd_extents.c:95
     #3  ssd_create at ../src/ssd/ssd.c:161
     #4  decorate.part.0.lto_priv.0 at ../src/view.c:637
     #5  decorate at ../src/view.c:636
     #6  view_set_fullscreen at ../src/view.c:716
     #7  view_adjust_for_layout_change ../src/view.c:745
     #8  desktop_arrange_all_views ../src/desktop.c:52
     #9  output_update_all_usable_areas at ../src/output.c:495
     #10 output_update_for_layout_change at ../src/output.c:263
     #11 do_output_layout_change at ../src/output.c:423
     #12 do_output_layout_change at ../src/cursor.c:267
     #13 output_config_apply at ../src/output.c:334
     #14 handle_output_manager_apply at ../src/output.c:354

### DIFF
--- a/src/ssd/ssd_extents.c
+++ b/src/ssd/ssd_extents.c
@@ -24,16 +24,6 @@ add_extent(struct wl_list *part_list, enum ssd_part_type type,
 	return part;
 }
 
-static void
-lab_wlr_output_layout_layout_coords(struct wlr_output_layout *layout,
-		struct wlr_output *output, int *x, int *y)
-{
-	struct wlr_output_layout_output *l_output;
-	l_output = wlr_output_layout_get(layout, output);
-	*x += l_output->x;
-	*y += l_output->y;
-}
-
 void
 ssd_extents_create(struct ssd *ssd)
 {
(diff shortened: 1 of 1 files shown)


Consolatis:
Right.. So view in fullscreen thus no SSD, disable output -> disables fullscreen, re-enables SSD, output not part of the layout anymore: crash. Not sure why this didn't show up sooner, I guess not many people disable outputs. Could also be related to us now recreating the SSD instead of just enabling it again.
In either case: Nice catch!

I didn't test but it looks good to me, if it fixes the issue for you feel free to "Rebase and merge".
And thanks for the further code improvements as well.

---
The codecheck complained earlier:
> src/ssd/ssd_extents.c:101: ERROR: do not use assignment in if condition

Although I use assignments within `if` or `while` myself at times I think its easy to miss a `=` instead of `==` or some more `()` around the condition when just reading through the code so I think we should keep the check.

jlindgren90:
Thanks for the review! And I agree with keeping the code check; it enforces a good practice that I'm sometimes too lazy or "clever" to follow :sweat_smile:

PR #780, opened 2023-02-15
view: Tidy up view->output/view->fullscreen redundancy

jlindgren90:
A fullscreen view currently has its output specified twice by:
  - `struct output *output`
  - `struct wlr_output *fullscreen`

`view->fullscreen` may also become a dangling pointer if the output is disconnected, because `view_on_output_destroy()` clears `view->output` but not `view->fullscreen`.

To eliminate the redundancy and the dangling pointer, let's change `view->fullscreen` to a Boolean and rely on `view->output` to specify the output.

Along the way, change a few related usages of `struct wlr_output` to `struct output` as well.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -221,9 +221,9 @@ view_minimize(struct view *view, bool minimized)
 	}
 }
 
-/* view_wlr_output - return the output that a view is mostly on */
-static struct wlr_output *
-view_wlr_output(struct view *view)
+/* view_output - return the output that a view is mostly on */
+static struct output *
+view_output(struct view *view)
 {
 	assert(view);
 	double closest_x, closest_y;
@@ -234,13 +234,6 @@ view_wlr_output(struct view *view)
 		&closest_x, &closest_y);
 	wlr_output = wlr_output_layout_output_at(view->server->output_layout,
 		closest_x, closest_y);
-	return wlr_output;
-}
-
-static struct output *
-view_output(struct view *view)
-{
-	struct wlr_output *wlr_output = view_wlr_output(view);
 	return output_from_wlr_output(view->server, wlr_output);
 }
 
@@ -442,16 +435,17 @@ view_apply_tiled_geometry(struct view *view, struct output *output)
 }
 
 static void
-view_apply_fullscreen_geometry(struct view *view, struct wlr_output *wlr_output)
+view_apply_fullscreen_geometry(struct view *view)
 {
-	assert(wlr_output);
-	struct output *output =
-		output_from_wlr_output(view->server, wlr_output);
+	assert(view->fullscreen);
+	assert(output_is_usable(view->output));
+
 	struct wlr_box box = { 0 };
-	wlr_output_effective_resolution(wlr_output, &box.width, &box.height);
+	wlr_output_effective_resolution(view->output->wlr_output,
+		&box.width, &box.height);
 	double ox = 0, oy = 0;
-	wlr_output_layout_output_coords(output->server->output_layout,
-		output->wlr_output, &ox, &oy);
+	wlr_output_layout_output_coords(view->server->output_layout,
+		view->output->wlr_output, &ox, &oy);
 	box.x -= ox;
 	box.y -= oy;
 	view_move_resize(view, box);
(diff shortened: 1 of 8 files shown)


Consolatis:
> `view->fullscreen` may also become a dangling pointer if the output is disconnected, because `view_on_output_destroy()` clears `view->output` but not `view->fullscreen`.

We are resetting the fullscreen pointer to `false` (which seems a bit strange) in `view_adjust_for_layout_change()` which gets triggered by (among others) the output_destroy handler in `output.c`.

Will look into the diff in more detail later.

jlindgren90:
> We are resetting the fullscreen pointer to `false` (which seems a bit strange, should be `NULL` instead) in `view_adjust_for_layout_change()` which gets triggered by (among others) the output_destroy handler in `output.c`.

Okay, yes, `view_adjust_for_layout_change()` does clear `view->fullscreen`, but only after calling `wlr_output_layout_get(layout, view->fullscreen)` with a potentially invalid pointer. This seems bad since `wlr_output_layout_get()` does dereference the point, and in fact I'm surprised we haven't seen any crash from it yet.

jlindgren90:
v2: Don't allow entering fullscreen on disabled output (makes conditions for entering/leaving fullscreen symmetric)

Consolatis:
> Okay, yes, `view_adjust_for_layout_change()` does clear `view->fullscreen`, but only after calling `wlr_output_layout_get(layout, view->fullscreen)` with a potentially invalid pointer. This seems bad since `wlr_output_layout_get()` does dereference the pointer, and in fact I'm surprised we haven't seen any crash from it yet.

The output is not destroyed at that point: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.16/types/output/output.c#L400
E.g. the destroy signal is emitted before the free of the output. I agree that this feels racy though. That check was mainly intended for disabled outputs as far as a I remember, e.g. the output is still around but not part of the layout anymore.

src/view.c
```diff
@@ -34,6 +34,12 @@ enum view_edge {
 	VIEW_EDGE_CENTER,
 };
 
+static bool
+is_enabled_output(struct output *output)
+{
+	return output && output->wlr_output->enabled;
```
Consolatis, line 40: I think this should also check for `!output->leased`. See output.c for reference.
Should we move this function to `output.c` and call it `output_available()` or something instead?

src/view.c
```diff
@@ -436,16 +435,18 @@ view_apply_tiled_geometry(struct view *view, struct output *output)
...
 static void
-view_apply_fullscreen_geometry(struct view *view, struct wlr_output *wlr_output)
+view_apply_fullscreen_geometry(struct view *view)
 {
-	assert(wlr_output);
-	struct output *output =
-		output_from_wlr_output(view->server, wlr_output);
+	assert(view->fullscreen);
+	if (!is_enabled_output(view->output)) {
+		return;
```
Consolatis, line 442: We may want to prevent ever calling `apply_fullscreen_geometry()` when the output is not available, e.g. we could use an `assert()` here instead. If the output for a fullscreen view is gone, I think we should simply disable the fullscreen and then let the usual logic to move views that are not within the layout anymore take over.

src/view.c
```diff
@@ -673,16 +674,15 @@ view_toggle_fullscreen(struct view *view)
 }
 
 void
-view_set_fullscreen(struct view *view, bool fullscreen,
-		struct wlr_output *wlr_output)
+view_set_fullscreen(struct view *view, bool fullscreen, struct output *output)
 {
 	assert(view);
-	if (fullscreen != !view->fullscreen) {
+	if (fullscreen == view->fullscreen) {
```
Consolatis, line 680: That is so much better, can't count how many times i've had to stop here and verify that it was actually correct before.

jlindgren90:
> The output is not destroyed at that point: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.16/types/output/output.c#L400 E.g. the destroy signal is emitted before the free of the output. I agree that this feels racy though. That check was mainly intended for disabled outputs as far as a I remember, e.g. the output is still around but not part of the layout anymore.

I see, makes sense, thanks.

src/view.c
jlindgren90: Sounds good. Would `output_is_usable` be a good name or do you prefer `output_available`?

src/view.c
jlindgren90: We do have logic to do that (disable fullscreen) if needed in response to a layout change. However I'm struggling to convince myself that an `assert` here would never be hit. As a compromise, how about just logging an error before the `return`?

src/view.c
Consolatis: `output_is_usable` sounds usable to me

src/view.c
jlindgren90: Looking at it again, the `assert` is fine. Added.

jlindgren90:
v3: Add `output_is_usable()` helper

Consolatis:
LGTM, but didn't test.
Feel free to "Rebase and merge" if you are satisfied with the PR.

jlindgren90:
Thanks for the review!

PR #781, opened 2023-02-15
desktop: Avoid centering views without initial geometry

jlindgren90:
While working on a different change, I noticed that in some cases (generally at labwc startup, when several windows are created at once) we were calling `view_compute_centered_position()` with an empty width/height. This occurs because `desktop_arrange_all_views()` has no regard for whether views are mapped or have an initial width/height set yet.

I don't think there were any user-observable side effects, but even so it seems best to avoid calling `view_compute_centered_position()` for views that have no initial width/height, since it's useless and (to me at least) unexpected, which could lead to surprises in future development.

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -46,10 +46,21 @@ desktop_move_to_back(struct view *view)
 void
 desktop_arrange_all_views(struct server *server)
 {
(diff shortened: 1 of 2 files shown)


johanmalm:
Thanks.

PR #782, opened 2023-02-15
view: Add view_evacuate_regions()

jlindgren90:
Just a slight refactoring to follow the existing precedent that `struct view` fields are typically only modified within `view.c`.

### DIFF
--- a/src/regions.c
+++ b/src/regions.c
@@ -164,6 +164,7 @@ regions_reconfigure_output(struct output *output)
 	wl_list_for_each(region, &rc.regions, link) {
 		struct region *region_new = znew(*region_new);
 		/* Create a copy */
+		region_new->output = output;
 		region_new->name = xstrdup(region->name);
 		region_new->percentage = region->percentage;
 		wl_list_append(&output->regions, &region_new->link);
@@ -214,14 +215,9 @@ regions_evacuate_output(struct output *output)
 {
 	assert(output);
 	struct view *view;
-	struct region *region;
-
 	wl_list_for_each(view, &output->server->views, link) {
-		wl_list_for_each(region, &output->regions, link) {
-			if (view->tiled_region == region) {
-				view_evacuate_region(view);
-				break;
-			}
+		if (view->tiled_region && view->tiled_region->output == output) {
+			view_evacuate_region(view);
 		}
 	}
 }
--- a/src/regions.c
+++ b/src/regions.c
@@ -219,12 +219,7 @@ regions_evacuate_output(struct output *output)
 	wl_list_for_each(view, &output->server->views, link) {
 		wl_list_for_each(region, &output->regions, link) {
 			if (view->tiled_region == region) {
-				if (!view->tiled_region_evacuate) {
-					view->tiled_region_evacuate =
-						xstrdup(region->name);
-				}
-				/* Prevent carrying around a dangling pointer */
-				view->tiled_region = NULL;
+				view_evacuate_region(view);
 				break;
 			}
 		}
--- a/src/view.c
+++ b/src/view.c
@@ -775,6 +775,17 @@ view_discover_output(struct view *view)
 }
 
 void
+view_evacuate_region(struct view *view)
+{
+	assert(view);
+	assert(view->tiled_region);
+	if (!view->tiled_region_evacuate) {
+		view->tiled_region_evacuate = xstrdup(view->tiled_region->name);
+	}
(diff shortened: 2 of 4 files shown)


src/regions.c
```diff
@@ -214,20 +214,8 @@ regions_evacuate_output(struct output *output)
...
-				if (!view->tiled_region_evacuate) {
-					view->tiled_region_evacuate =
-						xstrdup(region->name);
-				}
-				/* Prevent carrying around a dangling pointer */
-				view->tiled_region = NULL;
-				break;
-			}
-		}
+		view_evacuate_regions(view, &output->regions);
```
Consolatis, line 218: This could actually have an early out, e.g. something like
```c
if (!view->tiled_region) {
  continue;
}
```
before calling `view_evacuate_regions()`

Nit: I think `evacuate_region()` (singular) would better describe the new function, possibly even `try_evacuate_region()`?

An alternative would be to add a `struct output` pointer to `struct region`, then we wouldn't have to loop over all regions and could just do something like
```c
wl_list_for_each(view, &output->server->views, link) {
	if (view->tiled_region && view->tiled_region->output == output) {
		view->tiled_region_evacuate = xstrdup(view->tiled_region->name);
		view->tiled_region = NULL;
	}
}
```

jlindgren90: Good ideas; I will mull over these and come back to this another day (getting late here).

jlindgren90: How about `view_evacuate_any_region_of(struct view *view, struct output *output)`?

And I will add the "early out" as suggested.

jlindgren90: Or alternatively leave the loop+conditional where it is, and then `view_evacuate_region(struct view *view)` unconditionally evacuates `view->tiled_region`.

Consolatis: > How about `view_evacuate_any_region_of(struct view *view, struct output *output)`?

A view can only have a single region so the `any_region_of` part seems a bit misleading to me.

You don't like the  "add `struct output` to `struct region`" idea? I think it would improve the readability of that function immensely and by making the complete inner loop obsolete.

jlindgren90: > You don't like the "add struct output to struct region" idea?
I do like the idea, it's a nice optimization and has precedent in labwc (e.g. `struct view` pointing back to `struct server`).
I was just hesitating as I'm not too familiar with the region code yet. I'm happy to make the change if you can give it some good testing and make sure I haven't inadvertently broken something.

jlindgren90: Just pushed a new version that (I think) incorporates all the suggestions.

src/view.c
```diff
@@ -772,6 +772,17 @@ view_discover_output(struct view *view)
 	view->output = view_output(view);
 }
 
+void
+view_evacuate_region(struct view *view)
+{
+	assert(view);
+	if (view->tiled_region) {
+		assert(!view->tiled_region_evacuate);
```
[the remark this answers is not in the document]
jlindgren90: Hmm, I found two places where `tiled_region` is set (other than to `NULL`):
- https://github.com/labwc/labwc/blob/0ca6c4c763eaee091a3abeddd135eb6272fcfd26/src/view.c#L369
- https://github.com/labwc/labwc/blob/0ca6c4c763eaee091a3abeddd135eb6272fcfd26/src/view.c#L950

In both cases it is accompanied by `zfree(view->tiled_region_evacuate)`. So it appeared to me impossible for both `tiled_region` and `tiled_region_evacuate` to be non-null at the same time. Thus `assert(!view->tiled_region_evacuate)` seemed safe.

However maybe it would be best to be safe/flexible here. What about this?
```
	assert(view);
	if (view->tiled_region && !view->tiled_region_evacuate) {
		view->tiled_region_evacuate = xstrdup(view->tiled_region->name);
	}
	view->tiled_region = NULL;
```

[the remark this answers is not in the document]
Consolatis: IMHO this function should really *never* be called if there is no region to evacuate.
So not using an assert for `view->tiled_region` allows hiding future bugs in the call path.

Regarding `tiled_region_evacuate`: It might be true that the check is redundant, however, its a logic change from the state before and I am too tired right now to look through all possible call chains and conditions to verify that it will never be set along with `tiled_region` itself.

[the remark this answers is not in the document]
jlindgren90: All right, I updated the `assert`s as you suggested.

Consolatis:
Thanks

PR #786, opened 2023-02-17
xwayland: Be more selective about updating natural geometry

jlindgren90:
For maximized/fullscreen/tiled views, interpret ConfigureRequest as updating the natural geometry only if it's still zero. Wait as long as possible before setting the fallback geometry in order to prefer any geometry we might get from ConfigureRequests.

Fixes an issue with xfce4-terminal (XWayland mode), where after tiling, the natural geometry was incorrectly overwritten with the tiled geometry.

Fixes: 09599861ac726b142a79ed60035809f04c151457
("xwayland: Fix size issue when starting VLC fullscreen")

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -279,21 +279,6 @@ view_compute_centered_position(struct view *view, int w, int h, int *x, int *y)
 	return true;
 }
 
-static void
-set_fallback_geometry(struct view *view)
-{
-	view->natural_geometry.width = LAB_FALLBACK_WIDTH;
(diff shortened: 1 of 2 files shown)


jlindgren90:
Marked as draft for now since I would like to do some more testing still.

jlindgren90:
This needs more thought. I'm not certain that we shouldn't just ignore the weird VLC configure-request (after entering fullscreen) and consider it a bug in VLC. The VLC window would just use fallback geometry (640x480) when leaving fullscreen, but that's not the end of the world.

Issue #768, opened 2023-02-10
use with lxqt


daktras420:
i am working on my system and really hate kwin and even kwinft.. and i would love to use the full lxqt list of applications using your de but the lxqt-panel doesn't load the way sfwbar waybar or any of the gtk based wayland bar apps run on your wm project. the following link will give you info why the lxqt team is having issues with using you wm as a wayland replacement for openbox on x11.
https://github.com/lxqt/lxqt/issues/10#issue-comment-box

Consolatis:
For panels all you need is usually
- the wlr-layershell protocol
- wlr-foreign-toplevel-management protocol
- hopefully at some point in the future the ext-workspace protocol

I am not sure about the current state of the wayland support for lxqt-panel.
If there are specific issues in regards to labwc feel free to mention them.

daktras420:
Lxqt panel works on kwin kwinft they claim because those desktops give
special support and the need for a qt-layer-shell is needed please refer to
the discussion previously posted to see what the developers are saying
about your wm and what is needed to make it usable for lxqt.

Consolatis:
I am sorry but I will not read through a general wayland support issue with > 400 comments and search for (possibly long fixed) issues with labwc. But again, feel free to mention specific things that are still broken on the latest master branch or labwc 0.6.1.

daktras420:
okay i will do my best to try to summarize the info for you for its
actually a big issue.
according to the following work and info of using lxqt de on wayland and
working with your labwc wm project.
https://github.com/stefonarch/LXQt-Wayland-files
with qt apps the wlr-layershell does not allow auto placement and since
theres no window rules it makes qt placement and access of layershell not
possible. and like sfwbar go past the screen and make the ability to use a
menu not functional as well as the no automatic placement of app windows. i
do apologize but i am not a programmer and so trying to help you by
facilitate information on the problems when i dont understand all the terms
my self is difficult. and thought providing the links to discussions why
developers are saying your wm isnt usable yet would be helpful if the
information i am trying to provide isn't useful i am sorry for wasting your
time. but in x11 my only use for openbox was to integrate it onto lxqt and
to be honest that is the only use i would have for your project as well i
get hitting a button to go to the bottom of the page and read the current
state of things and staying up-to-date on the discussion may be to much for
you but feel very insulted by your reluctance to work on obviously blatant
issues

Consolatis:
> with qt apps the wlr-layershell does not allow auto placement

[...]

PR #787, opened 2023-02-17
[very wip] window-rules in support of lxqt-panel

johanmalm:
Related-to: #768

I'm not sure if this is a good idea, but would appreciate high-level thoughts on it.
Please ignore ugly code, etc - it's the general idea I'm wanting to test.

I can't get `lxqt-panel` to work on my machine (just get a completely transparent panel - both on labwc and openbox).
Have tested with this little script:

```python3
#!/usr/bin/env python3

from PyQt5.QtWidgets import *
from PyQt5.QtCore import *
import sys

class Panel(QWidget):
    def __init__(self):
        super().__init__()
        self.setGeometry(0, 0, 1362, 24)
        self.setWindowTitle('lxqt-panel')

        self.cb = QCheckBox('toggle title', self)
        self.cb.move(20, 0)
        self.cb.toggle()
        self.cb.stateChanged.connect(self.changeTitle)

        self.button = QPushButton("Close", self)
        self.cb.move(100, 0)
        self.button.clicked.connect(lambda:self.close())

        self.show()

    def changeTitle(self, state):
        if state == Qt.Checked:
            self.setWindowTitle('foo')
        else:
            self.setWindowTitle('lxqt-panel')

def main():
    app = QApplication(sys.argv)
    ex = Panel()
    sys.exit(app.exec_())

if __name__ == '__main__':
    main()
```

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -6,6 +6,7 @@
 #include "view.h"
 #include "view-impl-common.h"
 #include "workspaces.h"
+#include "xdg-unmanaged.h"
 
 static struct xdg_toplevel_view *
 xdg_toplevel_view_from_view(struct view *view)
@@ -132,6 +133,17 @@ handle_destroy(struct wl_listener *listener, void *data)
 	wl_list_remove(&xdg_toplevel_view->new_popup.link);
 
 	view_destroy(view);
+}
+
+void
+xdg_toplevel_view_convert_to_unmanaged(struct view *view)
+{
+	struct wlr_xdg_surface *xdg_surface = xdg_surface_from_view(view);
+	struct server *server = view->server;
+
+	handle_unmap(&view->unmap, xdg_surface);
+	handle_destroy(&view->destroy, xdg_surface);
+	xdg_unmanaged_create(server, xdg_surface);
 }
 
 static void
@@ -375,6 +387,7 @@ xdg_toplevel_view_map(struct view *view)
 	wl_signal_add(&xdg_surface->surface->events.commit, &view->commit);
 
 	view_impl_map(view);
+	view_apply_rules(view);
 }
 
 static void
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -348,6 +348,7 @@ void xdg_toplevel_decoration(struct wl_listener *listener, void *data);
 
 void xdg_activation_handle_request(struct wl_listener *listener, void *data);
 
+void xdg_toplevel_view_convert_to_unmanaged(struct view *view);
 void xdg_surface_new(struct wl_listener *listener, void *data);
 
 void foreign_toplevel_handle_create(struct view *view);
--- a/include/node.h
+++ b/include/node.h
@@ -13,6 +13,7 @@ enum node_descriptor_type {
 	LAB_NODE_DESC_NODE = 0,
 	LAB_NODE_DESC_VIEW,
 	LAB_NODE_DESC_XDG_POPUP,
+	LAB_NODE_DESC_XDG_UNMANAGED,
 	LAB_NODE_DESC_LAYER_SURFACE,
 	LAB_NODE_DESC_LAYER_POPUP,
 	LAB_NODE_DESC_MENUITEM,
@@ -37,6 +38,7 @@ struct node_descriptor {
  * @data: struct to point to as follows:
  *   - LAB_NODE_DESC_VIEW           struct view
  *   - LAB_NODE_DESC_XDG_POPUP      struct view
+ *   - LAB_NODE_DESC_XDG_UNMANAGED  struct xdg_unmanaged
  *   - LAB_NODE_DESC_LAYER_SURFACE  struct lab_layer_surface
  *   - LAB_NODE_DESC_LAYER_POPUP    struct lab_layer_popup
  *   - LAB_NODE_DESC_MENUITEM       struct menuitem
--- a/include/view.h
+++ b/include/view.h
@@ -167,6 +167,7 @@ void view_adjust_size(struct view *view, int *w, int *h);
 void view_evacuate_region(struct view *view);
 void view_on_output_destroy(struct view *view);
 void view_destroy(struct view *view);
+void view_apply_rules(struct view *view);
 
 /* xdg.c */
 struct wlr_xdg_surface *xdg_surface_from_view(struct view *view);
--- a/include/xdg-unmanaged.h
+++ b/include/xdg-unmanaged.h
@@ -1,0 +1,10 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_XDG_UNMANAGED
+#define __LABWC_XDG_UNMANAGED
+
+struct server;
+struct wlr_xdg_surface;
+
+void xdg_unmanaged_create(struct server *server, struct wlr_xdg_surface *wlr_xdg_surface);
+
+#endif /* __LABWC_XDG_UNMANAGED */
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -348,6 +348,10 @@ get_cursor_context(struct server *server)
 					ret.surface = lab_wlr_surface_from_node(ret.node);
 				}
 				return ret;
+			case LAB_NODE_DESC_XDG_UNMANAGED:
+				ret.type = LAB_SSD_CLIENT;
+				ret.surface = lab_wlr_surface_from_node(ret.node);
+				return ret;
 			case LAB_NODE_DESC_SSD_BUTTON: {
 				/*
 				 * Always return the top scene node for SSD
--- a/src/meson.build
+++ b/src/meson.build
@@ -26,6 +26,7 @@ labwc_sources = files(
   'xdg.c',
   'xdg-deco.c',
   'xdg-popup.c',
+  'xdg-unmanaged.c',
 )
 
 if have_xwayland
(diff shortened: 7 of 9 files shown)


Consolatis:
What is the reason for the new unmanaged_xdg struct?

Couldn't we just parent to the always_on_top_tree directly based on the rule match, same for the SSD?

src/xdg.c
```diff
@@ -114,17 +121,73 @@ handle_destroy(struct wl_listener *listener, void *data)
...
-	/* Reset XDG specific surface for good measure */
-	((struct xdg_toplevel_view *)view)->xdg_surface = NULL;
+	wl_list_remove(&view->map.link);
+	wl_list_remove(&view->unmap.link);
+	wl_list_remove(&view->request_move.link);
+	wl_list_remove(&view->request_resize.link);
+	wl_list_remove(&view->request_minimize.link);
+	wl_list_remove(&view->request_maximize.link);
+	wl_list_remove(&view->request_fullscreen.link);
+	wl_list_remove(&view->set_title.link);
```
Consolatis, line 137: We could move the listener cleanup for the common `struct view` listeners to `view_destroy()` and only clean up the xdg specific ones here. This should likely be done in a separate PR though.

I like the `xdg_toplevel_view_from_view()` addition, inline casting like that always feels a bit meh.

Just a nit: I ended up using `struct xdg_toplevel_view *xdg_view = ..` so we don't have to spell out the variable name each time.

jlindgren90:
In the way of high-level thoughts, I have a question: is the end-goal that lxqt-panel (or any other panel for that matter) uses layer-shell? Or are these rules an alternative method to run a (Qt) panel, without layer-shell? What lxqt-panel ends up doing is particularly of interest to me since I would likely follow the same approach with qmpanel.

> I can't get lxqt-panel to work on my machine (just get a completely transparent panel - both on labwc and openbox).

Is it possible that the default configuration is in fact an empty, transparent panel? It's been a while since I ran lxqt-panel, but this sounds vaguely familiar.

Consolatis:
My 2 cents:

From the labwc side I think we can support 2 ways for panels:
- wayland native via wlr-layershell (and hopefully at some point ext-layershell if the protocol ends up being accepted and implemented in wlroots)
- X11 via your work on `STRUT_PARTIAL` + my work for the xdg <-> x11 bridge (this won't work until at least wlroots 0.17)

The window rule thing I personally see with a different goal, not specific to panels: Allow users to do some automation for views based on app_id or title. They can also be used for panels obviously but panels should not rely on them.

stefonarch:
I've no problems to have the panel visible on labwc.

> Is it possible that the default configuration is in fact an empty, transparent panel? It's been a while since I ran lxqt-panel, but this sounds vaguely familiar.
 
A second panel added is empty, the first one should take it's default config which segfaults afaik due to the desktop switch plugin.  [Here](https://github.com/stefonarch/LXQt-Wayland-files/blob/main/lxqt-wayland/panel.conf)'s a working config, `lxqt-panel -c /path/to/panel.conf`

Will test the draft as soon as I find time. 

>Allow users to do some automation for views based on app_id or title. They can also be used for panels obviously but panels should not rely on them.

It's not only the panel in fact: some rules could be very useful for `pcmanfm-qt --desktop`: always_at_bottom. not_closable, on_all_desktops, fullscreen, exclude from switcher. Before https://github.com/lxqt/pcmanfm-qt/pull/1729  it couldn't be distinguished by title fdrom other windows.

jlindgren90:
> It's not only the panel in fact: some rules could be very useful for pcmanfm-qt --desktop: always_at_bottom. not_closable, on_all_desktops, fullscreen, exclude from switcher.

Isn't that (desktop icons/folder view) also a use-case that layer-shell is meant to serve?

stefonarch:
Sure, but porting pcmanfm-qt to wayland is in the future.

stefonarch:
Wow, started from build folder and started my panel, just perfect!
screen_area_sab_14:22:10_

stefonarch:
BTW this will  work really fine per `pcmanfm-qt --desktop`  (`title=pcmanfm-desktop0`, `always_on_bottom`) too, except the right click not working

johanmalm:
Background:

- https://github.com/labwc/labwc/issues/768#issuecomment-1427162219

@Consolatis, @jlindgren90: So, I'm exploring ways to run q5/6/gtk4 applications
in non-standard ways because it seems like their 'full' adoption of layer-shell
is a long way off. I just learnt a bit more about QtLayerShell with qt5 through
@Consolatis's example (https://github.com/labwc/labwc/issues/768#issuecomment-1435736959),
but I think (to be confirmed though!) that popup's will be hard.

Yes, of course layer-shell it the preferred method, but if it is going to be
years until that's a reality, then I'm keen to explore options to make it work
now. To get a panel, desktop-client or similar working is technically not that
hard, but we obviously want the solution to stand the test of time, not add
disproportionate complexity, etc etc.

What I hadn't realised until a couple of weeks ago was that
sway/wayfire/hyprland implement window rules (defined in config file and based
on app_id/title) and that kwin assign window roles through a protocol.

So I started thinking about how we could do this without creating a monster
and disappearing down a hole. I'm quite a slow adopter and conceptually I
envisage that window-roles/rules would add a lot of code complexity with
if-statements all over the place (or possibly xdg_toplevel_view inheritance
to abstract some of that away) but nonetheless adding significant complexity.

This PR is an experiment to test a theory that for a panel, desktop-client
(e.g. `pcmanfm-qt --desktop`) or similar, pretty much all the `view`
functionality should be turned off for example:

- Ability for user to move or resize window
- Server side decoration
- Foreign top-level handling
- Window stacking and all associated `desktop.c` stuff
- Window cycling (like alt-tab)

...which led me to think that if we turned traditional window-rules concept on
its head and created a super simple struct to hold the node and node-descriptor
(i.e. `xdg_unmanaged`) then maybe that could be an `<applications>` rule as
follows:

```xml
<applications>
  <application app_id="lxqt-panel">
    <unmanaged layer="top" />
    <position x="0" y="0" />
  </application>
</applications>
```

```xml
<margins top="30" bottom="0" right="0" left="0" output="WL-1" />
```

With 'traditional' window-rules we'd have to do some sort of implementation
of position, decor, pin, sticky, skip_taskbar, skip_pager, layer which seems
quite complex and deep - which is why I'm entertaining the idea of a complete
opt out (i.e. position the surface, and that's it).

And, yes - of course it's a manual mini-layer-shell implementation but with
based on configuration rather than a protocol and with a lot let flexibility.

From a user perspective, the main disadvantages is that `rc.xml` need to be
edited in order to get the panel to work as desired - including setting
`<margin>` to adjust usable_area and so on.

So, yes, it's a compromise in making labwc usable to communities such as
lxqt without waiting for wider adoption of the layer-shell.

# References

`sway` has the for_window command to implement window rules. See: https://man.archlinux.org/man/sway.5#CRITERIA

It looks like wayfire can also associated rules with events - but I suspect
applying rules on `map` will be quite sufficient for most use-cases.

Examples of rules include:

- `decor` (openbox)
- `shade` - not really relevant because we don't do that anyway :smile:
- `position` (openbox)
  - `x`
  - `y`
  - `monitor`
  - `region` (did not find this anywhere else, but could be of interest to us)
- `size`
  - `width`
  - `height`
- `focus`
- `desktop`
- `layer` (e.g. above/below). Some call it `stay_on_top` or `stay_on_bottom` or similar.
- `iconic`
- `skip_pager`
- `skip_taskbar` (plasma-shell)
- `maximized`
- `fullscreen`

X11 Window types:

- `_NET_WM_WINDOW_TYPE_NORMAL`
- `_NET_WM_WINDOW_TYPE_DIALOG`
- `_NET_WM_WINDOW_TYPE_UTILITY`
- `_NET_WM_WINDOW_TYPE_TOOLBAR`
- `_NET_WM_WINDOW_TYPE_SPLASH`
- `_NET_WM_WINDOW_TYPE_MENU`
- `_NET_WM_WINDOW_TYPE_DROPDOWN_MENU`
- `_NET_WM_WINDOW_TYPE_POPUP_MENU`
- `_NET_WM_WINDOW_TYPE_TOOLTIP`
- `_NET_WM_WINDOW_TYPE_NOTIFICATION`

Kwin plasma-shell roles include:

- `normal`
- `desktop`
- `panel`
- `onscreendisplay`
- `notification`
- `tooltip`
- `criticalnotification`
- `appletpopup`

stefonarch:
@johanmalm Nice writeup and thanks for your efforts! 

 `lxqt-panel` is working fine now: sliders, calendar and menu at correct position, only the "configure panel" windows are at 0 0 (this happens on every compositor). As poups are working now the desktop in background layer should be easy.

FYI  `pcmanfm-qt  --desktop` works perfectly (except that it's now outside `lxqt-session`) with the [background_view plugin](https://github.com/WayfireWM/wayfire-plugins-extra/blob/master/src/background-view.cpp) in wayfire.
```
[background-view]
command = pcmanfm-qt --desktop
app_id = pcmanfm-qt
inhibit_input = false

```

stefonarch:
> From a user perspective, the main disadvantages is that `rc.xml` need to be
> edited in order to get the panel to work as desired 

This will be probably the situation with any compositor, some changes to the default config files. My idea is that in `lxqt-session` the wayland compositor can be set as now the x11 WM and those configs have to be shipped by LXQt, mostly the `window_rules` part.

johanmalm:
Leaving this one open for now, but see #933 for latest development.

Consolatis:
Is this PR still relevant now that we have window rules?

johanmalm:
Agree. Will close

Issue #785, opened 2023-02-17
#753 caused issues with tiled xfce4-terminal (xwayland)


jlindgren90:
#753 caused a couple of issues that show up when tiling xfce4-terminal in xwayland mode:

1. When tiling the window (SnapToEdge), the window contents don't fully repaint
2. When untiling the window again, the natural geometry is not restored (the window keeps roughly the tiled geometry)

After some investigation, both issues stem from the fact that the tiled geometry does not exactly match the requested size increments of the terminal window, and the application is trying to adjust or "fix" the tiled geometry by sending a ConfigureRequest.

Before #753, we would honor the ConfigureRequest and set the requested size, resulting in a small gap between the terminal window and the edge of the screen. Now we ignore the ConfigureRequest but (in this case incorrectly) update the natural geometry.

The first issue needs to be fixed on the wlroots side. X11 window managers are free to ignore a client-initiated ConfigureRequest, but are supposed to send a synthetic ConfigureNotify with the unchanged window geometry in that case. wlroots doesn't currently send the synthetic event, leaving xfce4-terminal waiting (which is why it doesn't repaint). I'll submit a wlroots pull request shortly.

The second issue can be fixed on the labwc side, with some heuristics to decide whether or not to update the natural geometry from the ConfigureRequest. I'll submit a pull request for this too.

jlindgren90:
Fix for first issue: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4008

Consolatis:
Regarding the second issue, as SnapToEdge (and SnapToRegion for that matter) are labwc internal states and not visible to the application, would it make sense to only use the change behavior introduced in #753 for maximized and fullscreen views?

jlindgren90:
> would it make sense to only use the changed behavior introduced in #753 for maximized and fullscreen views?

I'm not sure, I saw the issue with a tiled view but I think it could also happen with a maximized or fullscreen view. My plan was basically to update the natural geometry only if it was empty/zero.

Consolatis:
This might also have uncovered a third issue: I assume the terminal uses size hints so it seems we are not applying those for labwc initiated configure events (e.g. `view_move_resize()`). Otherwise xfce4-terminal wouldn't have to correct the sizes.

jlindgren90:
[...]

PR #788, opened 2023-02-17
Revert "xwayland: Fix size issue when starting VLC fullscreen"

jlindgren90:
The fix caused a couple of issues:

1. Ignoring client configure requests caused some clients to hang and not repaint correctly. We are supposed to synthesize a ConfigureNotify event when ignore/override a client configure request, but this isn't possible with current wlroots.
2. Setting `view->natural_geometry` from client configure requests resulted in overwriting good values with bad in some cases (e.g. with tiled xfce4-terminal in xwayland mode).

For now, revert the fix. This does allow clients to mess with view positioning for maximized/fullscreen/tiled views, but right now the alternatives seem worse.

The original specific issue (VLC undoing its fullscreen geometry) is arguably a bug in VLC anyway.

This reverts commit 09599861ac726b142a79ed60035809f04c151457.

Fixes: #785

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -305,20 +305,6 @@ handle_request_configure(struct wl_listener *listener, void *data)
 	int height = event->height;
 	view_adjust_size(view, &width, &height);
 
-	/*
-	 * If a configure request is received while maximized/
-	 * fullscreen/tiled, update the natural geometry only. This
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks for investigating this issue.

Issue #721, opened 2023-01-08
Application's dialog box placement problem


milisarge:
Some of applications dialog boxes (e.g geany's open file) place on panel area, not properly placed at the top of the panel. I am doing maximize then it places to right positions.

Consolatis:
The original issue should be solved, the issue of bigger-than-screen dialog windows is a generic one and not strictly related to labwc (or even wayland). Just automatically maximizing a dialog window when its bigger than the output sounds a bit hacky so closing this for now.

PR #789, opened 2023-02-18
xdg: Save natural_geometry.x/y with initially maximized view

jlindgren90:
Fixes an issue where, if Thunar was started maximized, it would un-maximize to the top-left corner rather than the center. This occurs because we are storing a valid `natural_geometry.width/height` but leaving `natural_geometry.x/y` as 0,0.

In order to make `position_xdg_toplevel_view()` work without writing to `view->current` or calling `view_move()`, I made `view_compute_centered_position()` public again and added an `output` parameter.

Background/off-topic/related to #767:
Unlike `foot --maximized`, Thunar maximizes *at* map rather than sending a separate request *before* map. As a result, we are able to save the natural geometry, but also get a flicker at startup while the view adjusts to the maximized size. So currently it seems that either approach (Thunar vs. `foot` behavior) has its own disadvantage.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -332,6 +332,19 @@ xdg_toplevel_view_map(struct view *view)
 				xdg_surface->current.geometry.height;
 		}
 
+		/*
+		 * Set initial "pending" position for floating views.
+		 * Do this before view_set_fullscreen/view_maximize() so
+		 * that the position is saved with the natural geometry.
+		 *
+		 * FIXME: the natural geometry is not saved if either
+		 * handle_request_fullscreen/handle_request_maximize()
+		 * is called before map (try "foot --maximized").
+		 */
+		if (view_is_floating(view)) {
+			position_xdg_toplevel_view(view);
+		}
+
 		if (!view->fullscreen && requested->fullscreen) {
 			struct output *output = output_from_wlr_output(
 				view->server, requested->fullscreen_output);
@@ -339,9 +352,14 @@ xdg_toplevel_view_map(struct view *view)
 		} else if (!view->maximized && requested->maximized) {
 			view_maximize(view, true,
 				/*store_natural_geometry*/ true);
-		} else if (view_is_floating(view)) {
-			position_xdg_toplevel_view(view);
 		}
+
+		/*
+		 * Set initial "current" position directly before
+		 * calling view_moved() to reduce flicker
+		 */
+		view->current.x = view->pending.x;
+		view->current.y = view->pending.y;
 
 		view_moved(view);
 		view->been_mapped = true;
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -273,12 +273,12 @@ lookup_view_by_xdg_toplevel(struct server *server,
 static void
 position_xdg_toplevel_view(struct view *view)
 {
-	struct wlr_xdg_surface *xdg_surface = xdg_surface_from_view(view);
 	struct wlr_xdg_toplevel *parent_xdg_toplevel =
 		xdg_toplevel_from_view(view)->parent;
 
 	if (!parent_xdg_toplevel) {
-		view_center(view, output_from_cursor_coords(view->server));
+		view_center(view, output_from_cursor_coords(view->server),
+			NULL);
 	} else {
 		/*
 		 * If child-toplevel-views, we center-align relative to their
@@ -287,17 +287,8 @@ position_xdg_toplevel_view(struct view *view)
 		struct view *parent = lookup_view_by_xdg_toplevel(
 			view->server, parent_xdg_toplevel);
 		assert(parent);
-		int center_x = parent->current.x + parent->current.width / 2;
-		int center_y = parent->current.y + parent->current.height / 2;
-		view->current.x = center_x
-			- xdg_surface->current.geometry.width / 2;
-		view->current.y = center_y
-			- xdg_surface->current.geometry.height / 2;
+		view_center(view, view_output(parent), &parent->pending);
 	}
-
-	struct border margin = ssd_get_margin(view->ssd);
-	view->current.x += margin.left;
-	view->current.y += margin.top;
 }
 
 static const char *
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -278,11 +278,7 @@ position_xdg_toplevel_view(struct view *view)
 		xdg_toplevel_from_view(view)->parent;
 
 	if (!parent_xdg_toplevel) {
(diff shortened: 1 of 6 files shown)


src/xdg.c
```diff
@@ -271,18 +271,16 @@ lookup_view_by_xdg_toplevel(struct server *server,
...
 
 	if (!parent_xdg_toplevel) {
-		struct wlr_box box =
-			output_usable_area_from_cursor_coords(view->server);
-		view->current.x = box.x;
-		view->current.y = box.y;
-		view_center(view);
+		view_compute_centered_position(view,
+			output_from_cursor_coords(view->server),
+			view->pending.width, view->pending.height, x, y);
```
Consolatis, line 283: Not completely thought though but anyway:

Could we keep using `view_center(view, output_from_cursor_coords(view->server))` here and move the `wlr_scene_node_set_enabled(&view->scene_tree->node, true)` after the `if (!view->been_mapped)` branch?

Then we could keep `view_compute_centerd_position()` private.

jlindgren90: Looking at it again, I think `view_center` could in fact work. It would call through to `view->impl->configure` unnecessarily, but since only the position (not size) has changed, that should be a no-op.

Not sure about moving the `wlr_scene_node_set_enabled` call. I haven't really looked at that as it didn't seem necessary to touch it for this fix.

jlindgren90:
v2: Use view_center() + pending coords everywhere

src/xdg.c
Consolatis: Delaying the scene node enable call would just mean that we only make the view visible after all in-between steps (e.g. calls to `view_moved()`) have been done. So we shouldn't see any in-between states on the screen that way.

jlindgren90: I see, makes sense. Would it be okay with you to defer that change to a future PR?

Consolatis:
This looks pretty good to me so far, didn't test.

There is a more generic question that came up while reading through the diff:
Should we allow `view_center(view, NULL)` which would then fall back to the cursor output?

jlindgren90:
> There is a more generic question that came up while reading through the diff: Should we allow `view_center(view, NULL)` which would then fall back to the cursor output?

One idea I had (which I've been playing with in my own `master` branch) is to make `view_center` (and `view_maximize` and the others) simply use `view->output`. So the extra `output` argument could go away again.

We would need something like a `view_impl_set_initial_output()` helper which would set the initial `view->output` based on the cursor location. After that we would need to ensure that `view->output` is consistently updated as the view moves or the output layout changes.

jlindgren90:
We could also rename `view_center()` to something like `view_impl_set_initial_centered_position()` and make it always use the cursor output. We never use the function for any other purpose.

Consolatis:
I think we have 3 use-cases
- use cursor output (initial mapping only?)
- use closest output to current layout position of the view (`apply_xxx_geometry()`)
- use specific output (fullscreen)

So whatever we do it should still allow for all the 3 to work correctly. Its especially important that `apply_xxx_geometry()` doesn't break. Until we figure out a nice design I'd be fine with adding the output as argument to `view_center()` and thus being more explicit on the caller side.

jlindgren90:
More thoughts on this:

> Should we allow view_center(view, NULL) which would then fall back to the cursor output?

Perhaps for now we make it fall back to the current behavior i.e. using `view_output(view)`. This would also be consistent with what `view_set_fullscreen` does.

I would also like to add an optional `struct wlr_box *ref` argument. Then we can also use `view_center` for the "child" xdg views that are centered relative to their parents.

Consolatis:
> Perhaps for now we make it fall back to the current behavior i.e. using `view_output(view)`. This would also be consistent with what `view_set_fullscreen` does.

That is a good point. In either case, we should likely document the fallback behavior in the header file.

> I would also like to add an optional `struct wlr_box *ref` argument. Then we can also use `view_center` for the "child" xdg views that are centered relative to their parents.

That would make sense as well, do we have other code paths that also position themselves relative to some other box? I am not sure about xwayland right now, there the clients possibly handle the positioning in relation to their parent themselves.

jlindgren90:
> do we have other code paths that position views relative to some other box? I am not sure about xwayland right now, there the clients possibly handle the positioning in relation to their parent themselves.

X11/xwayland has so many different options/combinations for window positioning. Right now we rely on the clients to handle any parent-relative positioning themselves, but there are possibly cases where we ought to do it on our side. Not 100% sure. I'd leave any xwayland positioning changes for a different issue.

jlindgren90:
Updated to use `view_center()` for parent-relative positioning; in detail:

- Add optional struct wlr_box* ref argument to view_center()
- Perform margin adjustment within view_center()
- Expose view_output() to fetch parent's output

I also included a small change as suggested in https://github.com/labwc/labwc/issues/721#issuecomment-1375893132 to align large views to the usable area rather than (0,0). So this changeset should also fix #721 now.

Consolatis:
The force-push diff is looking good, will look through the whole diff in more detail later again.

Consolatis:
LGTM, didn't test.
Feel free to "Rebase and merge" when you are satisfied with the current state.

I am really not the greatest fan of calls like `view_center(view, NULL, NULL);` because you'd have no clue what this actually means when just reading it. But I also don't see a better option currently.
This is an area of the code-base where we should really figure out a better API in the future.
`view_compute_centered_position()` gets really massive regarding the argument count as well.

jlindgren90:
> LGTM, didn't test. Feel free to "Rebase and merge" when you are satisfied with the current state.

Thanks for the review!

> I am really not the greatest fan of calls like `view_center(view, NULL, NULL);` because you'd have no clue what this actually means when just reading it. But I also don't see a better option currently. Renaming it to `view_center_on_output()` would explain the first `NULL` but for this area of the code-base we should really figure out a better API in the future.

Definitely agree there is room for improvement here. I have some ideas but will save them for a future PR/discussion.

> `view_compute_centered_position()` also gets pretty massive regarding the argument count, maybe we could supply and return `struct wlr_box` to reduce them a bit. Anyway, that is something for another PR, this is about saving natural geometry.

I've thought of passing a `struct wlr_box *` here too. It then becomes an "in-out" argument which I'm not a huge fan of normally, but it might make sense in this case.

Consolatis:
Alternative would be that its just an in argument (or we keep the width and height int ones. The function could return a struct wlr_box containing the the new x and y coordinates (and possibly a copy of the width and height arguments. Not sure if that would make things worse on the caller side though.

PR #790, opened 2023-02-18
xwayland: Prevent overriding maximized/fullscreen/tiled geometry

jlindgren90:
Prevents xwayland clients from messing with the geometry of maximized/fullscreen/tiled views.

This should fix the main issue of #753 (VLC having incorrect geometry when started fullscreen) more properly, without the side effects documented in #785. Unfortunately it depends on a wlroots change (https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4008) to work properly, and shouldn't be merged until we can require a wlroots version containing that change (otherwise it will trigger the same repaint issues that #753 suffered from).

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -319,12 +319,21 @@ handle_request_configure(struct wl_listener *listener, void *data)
 	struct view *view = &xwayland_view->base;
 	struct wlr_xwayland_surface_configure_event *event = data;
 
-	int width = event->width;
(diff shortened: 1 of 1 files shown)


jlindgren90:
This one should be ready to go now that we're targeting wlroots 0.17.

Consolatis:
LGTM but didn't test.

johanmalm:
Thanks!

PR #794, opened 2023-02-20
view: Rediscover output in view_adjust_for_layout_change()

jlindgren90:
Currently, `view_discover_output()` is only called by `view_moved()`. To be
consistent, we should also call `view_discover_output()` when the output
layout changes, since this can effectively "move" views to a different
output without actually calling `view_moved()`.

Letting `view->output` get out-of-date probably had minimal/obscure
effects in the current code, since in many cases we ignore `view->output`
altogether and find the nearest output "on the fly" with `view_output()`.
Nevertheless, I think it's still a bug and should be fixed.

Includes some slight refactoring in `view.c`:

 - Add fullscreen case to `view_apply_special_geometry()`
 - Split out private `set_fullscreen()` helper from `view_set_fullscreen()`
   (similar to existing `set_maximized/view_set_maximized()` split)

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -487,7 +487,9 @@ view_apply_maximized_geometry(struct view *view)
 static bool
 view_apply_special_geometry(struct view *view)
 {
-	if (view->maximized) {
+	if (view->fullscreen) {
+		view_apply_fullscreen_geometry(view);
+	} else if (view->maximized) {
 		view_apply_maximized_geometry(view);
 	} else if (view->tiled) {
 		view_apply_tiled_geometry(view, NULL);
@@ -729,13 +731,8 @@ view_set_fullscreen(struct view *view, bool fullscreen, struct output *output)
 	}
 
 	set_fullscreen(view, fullscreen);
-	if (fullscreen) {
-		view_apply_fullscreen_geometry(view);
-	} else {
-		/* Restore non-fullscreen geometry */
-		if (!view_apply_special_geometry(view)) {
-			view_apply_natural_geometry(view);
-		}
+	if (!view_apply_special_geometry(view)) {
+		view_apply_natural_geometry(view);
 	}
 }
 
--- a/src/view.c
+++ b/src/view.c
@@ -740,19 +740,26 @@ void
 view_adjust_for_layout_change(struct view *view)
 {
 	assert(view);
-	if (view->fullscreen) {
-		if (output_is_usable(view->output)) {
-			/* recompute fullscreen geometry */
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM but didn't test.

Feel free to merge, I'll look into the whole of `view.c` later again to see if I spot any accidentally introduced regressions during the last days. We got a report via IRC about the [havoc](https://github.com/ii8/havoc) terminal emulator (was new to me, actually looks pretty nice) doing strange stuff during/after the initial positioning. Debugging that might uncover something.

jlindgren90:
Thanks. If someone is willing to write up an issue report for the havoc issue (or even just copy & paste the relevant part of the IRC log), I'd be happy to look into that as well.

Consolatis:
I could reproduce it somewhat (but am currently running on a not quite recent master checkout): just starting havoc positions on 0,0.

I've also tried it in a recently compiled (although some days old) checkout of the master branch nested and there its the same but additionally when moving the terminal gets really small (< our fallback geometry). But I did not really start to debug at all so far (or even verified its still an issue on latest master).

jlindgren90:
Okay I can reproduce the havoc issue too. We hit this error:
https://github.com/labwc/labwc/blob/397a8f478565cd9a00d23ebc68742d1f5461edc2/src/view.c#L245

The view isn't centered properly on labwc 0.61 either (though in that version at least the SSD gets positioned at 0,0 rather than offscreen).

Consolatis:
Regarding the SSD offset, it might be somewhat related to us having two different ways to query the ssd margin:`ssd_thickness()` and `ssd_get_margin()`.

One just reacts to `view->ssd_enabled` and the other one checks for `view->ssd`.
The first one is for example important for the `foot -m` case which doesn't yet have the SSD initialized (as its called before map) but should still return the dimensions as if would be enabled.

jlindgren90:
> Regarding the SSD offset, it might be somewhat related to us having two different ways to query the ssd margin:`ssd_thickness()` and `ssd_get_margin()`.

Another theory to explain that difference:
Until recently, we had the following code in `position_xdg_toplevel_view()`:
```
struct border margin = ssd_get_margin(view->ssd);
view->current.x += margin.left;
view->current.y += margin.top;
```

#789 moved that code inside `view_compute_centered_position()` so now it doesn't run if we can't center the view. And we are failing to center havoc's view at map because the surface geometry is still 0x0 for some reason.

jlindgren90:
@Consolatis See #795 for the havoc issue.

PR #795, opened 2023-02-20
xdg: Use wlr_xdg_surface_get_geometry() to get size

jlindgren90:
Fixes the positioning issue with havoc mentioned in https://github.com/labwc/labwc/pull/794#issuecomment-1437427670.

Calling `position_xdg_toplevel_view()` from inside `handle_commit()` makes me a little nervous honestly, because it seems unexpected to be setting `view->pending` there. But it seems to work, and right now I can't think of any specific reasons not to do it. I'd appreciate someone else giving me a sanity check here, to make sure this doesn't break anything.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -325,10 +325,10 @@ xdg_toplevel_view_map(struct view *view)
 		 * dimensions remain zero until handle_commit().
 		 */
 		if (wlr_box_empty(&view->pending)) {
-			view->pending.width =
-				xdg_surface->current.geometry.width;
-			view->pending.height =
-				xdg_surface->current.geometry.height;
+			struct wlr_box size;
+			wlr_xdg_surface_get_geometry(xdg_surface, &size);
+			view->pending.width = size.width;
+			view->pending.height = size.height;
 		}
 
 		/*
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -3,7 +3,6 @@
 #include "common/mem.h"
 #include "labwc.h"
 #include "node.h"
-#include "ssd.h"
 #include "view.h"
 #include "workspaces.h"
 


Consolatis:
> Calling position_xdg_toplevel_view() from inside handle_commit() makes me a little nervous honestly

Yah. this seems more like a bandaid honestly.
I think we should look for a more generic solution, I can look into the issue in more details tomorrow.

jlindgren90:
> I am not sure if its even allowed to map a 0x0 surface or if that is a protocol violation. In that case we should likely just disconnect the client.

I noticed that havoc also doesn't respect the size we pass to `wlr_xdg_toplevel_set_size()` exactly, but adjusts it to match its desired size increment (i.e. multiple of terminal character/cell size). So when maximizing the window, it leaves a gap. Not sure if that's allowed by the protocol either.

jlindgren90:
I tried havoc under weston, just for kicks. The window is a reasonable size, and gets centered correctly. However it's impossible to move/resize due to weston's lack of server-side decorations.

jlindgren90:
It looks like weston does most of its `map()` processing from within it's commit-handler (which is the only place `map()` is called):
https://github.com/wayland-project/weston/blob/26678297466fd7c0bfafa9da5490f3e35d85079a/desktop-shell/shell.c#L2067

jlindgren90:
Okay, I was completely barking up the wrong tree with this one, and the real fix is much simpler. Oops.

The issue is that `xdg_surface->current.geometry` is set only by the `xdg_surface::set_geometry` message, which is optional.[1] If it's unset, then we are supposed to use the surface geometry. `wlr_xdg_surface_get_geometry()` already accounts for this, so we just need to use `wlr_xdg_surface_get_geometry()` instead of reading `xdg_surface->current.geometry` directly.

I'll force-push a fix shortly.

[1] https://wayland.app/protocols/xdg-shell#xdg_surface:request:set_window_geometry

johanmalm:
Nice work with the set_window_geometry.

Regarding map/commit on weston… that quite different. Looks like they don’t have a map handler in their “impl” vtable.
We don’t add the commit listener until we map :smile:

Consolatis:
Nice investigation and fix!
Can confirm that this fixes the initial positioning and move issue I saw before.

Consolatis:
> > I am not sure if its even allowed to map a 0x0 surface or if that is a protocol violation. In that case we should likely just disconnect the client.
> 
> I noticed that havoc also doesn't respect the size we pass to `wlr_xdg_toplevel_set_size()` exactly, but adjusts it to match its desired size increment (i.e. multiple of terminal character/cell size). So when maximizing the window, it leaves a gap. Not sure if that's allowed by the protocol either.

If you set `margin=yes` in `havoc.cfg` and start it from the same directory it stops doing that.
IMHO it should set that one internally when its maximized. The question remains if we should actually prevent that or not. And if yes, how exactly we should do that.

PR #798, opened 2023-02-21
view: Refactor view->output usage

jlindgren90:
Needs more thorough testing but I'm looking for initial feedback on the approach.

This comes with two (intentional) behavior changes:
- Maximized/tiled views now try harder to remain on the same output if the output layout changes
- Initially maximized xdg-shell views now appear on the cursor output (resolves #767)

`view->output` is now used as follows (copy & pasting from `view.h`):
```
/*
 * The output that the view is displayed on. Specifically:
 *
 *  - For floating views, this is the output nearest to the
 *    center of the view. It is computed automatically when the
 *    view is moved or the output layout changes.
 *
 *  - For fullscreen/maximized/tiled views, this is the output
 *    used to compute the view's geometry. The view remains on
 *    the same output unless it is disabled or disconnected.
 *
 * Many view functions (e.g. view_center(), view_fullscreen(),
 * view_maximize(), etc.) allow specifying a particular output
 * by setting view->output explicitly before calling them.
 */
```

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -138,9 +138,7 @@ void view_restore_to(struct view *view, struct wlr_box geometry);
 void view_set_untiled(struct view *view);
 void view_maximize(struct view *view, bool maximize,
 	bool store_natural_geometry);
-/* output is optional, defaults to current nearest output */
-void view_set_fullscreen(struct view *view, bool fullscreen,
-	struct output *output);
+void view_set_fullscreen(struct view *view, bool fullscreen);
 void view_toggle_maximize(struct view *view);
 void view_toggle_decorations(struct view *view);
 void view_toggle_always_on_top(struct view *view);
--- a/include/view.h
+++ b/include/view.h
@@ -131,9 +131,7 @@ void view_moved(struct view *view);
 void view_minimize(struct view *view, bool minimized);
 struct output *view_output(struct view *view);
 void view_store_natural_geometry(struct view *view);
-/* output is optional, defaults to current nearest output */
-void view_center(struct view *view, struct output *output,
-	const struct wlr_box *ref);
+void view_center(struct view *view, const struct wlr_box *ref);
 void view_restore_to(struct view *view, struct wlr_box geometry);
 void view_set_untiled(struct view *view);
 void view_maximize(struct view *view, bool maximize,
--- a/include/view.h
+++ b/include/view.h
@@ -151,7 +151,6 @@ void view_move_to_workspace(struct view *view, struct workspace *workspace);
 void view_set_decorations(struct view *view, bool decorations);
 void view_toggle_fullscreen(struct view *view);
 void view_adjust_for_layout_change(struct view *view);
-void view_discover_output(struct view *view);
 void view_move_to_edge(struct view *view, const char *direction);
 void view_snap_to_edge(struct view *view, const char *direction,
 	bool store_natural_geometry);
--- a/include/view.h
+++ b/include/view.h
@@ -129,7 +129,6 @@ void view_move_resize(struct view *view, struct wlr_box geo);
 void view_move(struct view *view, int x, int y);
 void view_moved(struct view *view);
 void view_minimize(struct view *view, bool minimized);
-struct output *view_output(struct view *view);
 void view_store_natural_geometry(struct view *view);
 void view_center(struct view *view, const struct wlr_box *ref);
 void view_restore_to(struct view *view, struct wlr_box geometry);
--- a/include/view.h
+++ b/include/view.h
@@ -147,6 +147,13 @@ void view_move(struct view *view, int x, int y);
 void view_moved(struct view *view);
 void view_minimize(struct view *view, bool minimized);
 void view_store_natural_geometry(struct view *view);
+
+/**
+ * view_center - center view within some region
+ * @view: view to be centered
+ * @ref: optional reference region (in layout coordinates) to center
+ * within; if NULL, view is centered within usable area of its output
+ */
 void view_center(struct view *view, const struct wlr_box *ref);
 void view_restore_to(struct view *view, struct wlr_box geometry);
 void view_set_untiled(struct view *view);
--- a/include/view.h
+++ b/include/view.h
@@ -38,6 +38,22 @@ struct view {
 	enum view_type type;
 	const struct view_impl *impl;
 	struct wl_list link;
+
+	/*
+	 * The output that the view is displayed on. Specifically:
+	 *
+	 *  - For floating views, this is the output nearest to the
+	 *    center of the view. It is computed automatically when the
+	 *    view is moved or the output layout changes.
+	 *
+	 *  - For fullscreen/maximized/tiled views, this is the output
+	 *    used to compute the view's geometry. The view remains on
+	 *    the same output unless it is disabled or disconnected.
+	 *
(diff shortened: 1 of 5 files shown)


src/xdg.c
```diff
@@ -175,17 +175,33 @@ handle_request_maximize(struct wl_listener *listener, void *data)
...
+	if (!view->fullscreen && requested->fullscreen
+			&& requested->fullscreen_output) {
+		struct output *output = output_from_wlr_output(view->server,
+			requested->fullscreen_output);
+		if (output_is_usable(output)) {
+			view->output = output;
+		} else {
+			wlr_log(WLR_ERROR,
+				"invalid output in fullscreen request");
+		}
```
Consolatis, line 191: I am not sure right now but I think clients can ask for fullscreen without supplying an output.
In the case of `!view->been_mapped` that should likely be the cursor output, otherwise the output the view is mainly on.

src/xdg.c
```diff
@@ -175,17 +175,33 @@ handle_request_maximize(struct wl_listener *listener, void *data)
...
+static void
+set_fullscreen_from_request(struct view *view,
+		struct wlr_xdg_toplevel_requested *requested)
+{
+	if (!view->fullscreen && requested->fullscreen
+			&& requested->fullscreen_output) {
+		struct output *output = output_from_wlr_output(view->server,
+			requested->fullscreen_output);
+		if (output_is_usable(output)) {
+			view->output = output;
```
Consolatis, line 187: This directly modifies the output for the view, what do you think about something like `view_set_output(view, output)` instead?

src/xdg.c
jlindgren90: That should be handled all right; view_set_fullscreen() is called regardless and will use the current view->output in that case.

src/view.c
```diff
@@ -342,15 +354,11 @@ view_apply_region_geometry(struct view *view)
 {
 	assert(view);
 	assert(view->tiled_region || view->tiled_region_evacuate);
+	struct output *output = view->output;
+	assert(output_is_usable(output));
```
Consolatis, line 358: Couldn't this assert trigger when there are not outputs left?

Consolatis:
In theory I like this approach. I am a bit worried how this reacts in the case of no outputs being available at all.
Including for fullscreen / maximize / tiling requests (from the client or the user via keybind), newly mapped windows and simliar.

src/xdg.c
jlindgren90: `view_set_output(view, output)` might be nicer.

Inside `view_set_output`, we could do `assert(!view->fullscreen)` (or maybe just return) since modifying `view->output` for a fullscreen view isn't allowed.

src/view.c
jlindgren90: The calling paths should prevent that now, see the new checks in `view_apply_special_geometry()` and `view_snap_to_region()`.

jlindgren90:
> In theory I like this approach. I am a bit worried how this reacts in the case of no outputs being available at all. Including for fullscreen / maximize / tiling requests (from the client or the user via keybind), newly mapped windows and simlar.
> 
> Edit: I guess if we prevent any `apply_xxx_geometry()` to be called in those cases it would be fine.

There is a top-level check for a usable output in `view_apply_special_geometry()` which is where (with a few exceptions) most of the other `apply` functions are called from. I think I checked the "few exceptions" too and made sure we have a usable output in those cases (hence the new checks to `view_move_to_edge` + `view_snap_to_edge` + `view_snap_to_region`) but definitely would appreciate a second pair of eyes (and testing!) to make sure I didn't miss any.

johanmalm:
Other than Consolatis's comments, it looks fine to me. I like the approach - seems to simplify things :smile:

jlindgren90:
Rebased to fix merge conflicts. I still need to add `view_set_output()`.

jlindgren90:
I added `view_set_output()` (new commit on top; probably not much value in rewriting the previous ones). I think this should be all set now, from my point-of-view.

Consolatis:
I'll try to test this later today.

include/view.h
```diff
@@ -129,18 +146,13 @@ void view_move_resize(struct view *view, struct wlr_box geo);
 void view_move(struct view *view, int x, int y);
 void view_moved(struct view *view);
 void view_minimize(struct view *view, bool minimized);
-struct output *view_output(struct view *view);
 void view_store_natural_geometry(struct view *view);
-/* output is optional, defaults to current nearest output */
-void view_center(struct view *view, struct output *output,
-	const struct wlr_box *ref);
+void view_center(struct view *view, const struct wlr_box *ref);
```
Consolatis, line 150: When reading the function definition it doesn't immediately become clear to me what `*ref` means.
So maybe we could add either a comment above explaining it or rename the argument. Maybe something like `container` or `parent` would make it more clear?

src/view.c
```diff
@@ -102,6 +102,21 @@ view_get_edge_snap_box(struct view *view, struct output *output,
 	return dst;
 }
 
+/*
+ * At present, a view can only 'enter' one output at a time, although the view
+ * may span multiple outputs. Ideally we would handle multiple outputs, but
+ * this method is the simplest form of what we want.
+ */
```
Consolatis, line 109: Is this comment still valid, e.g. do we actually need to know if a view is on multiple outputs?

src/view.c
```diff
@@ -102,6 +102,21 @@ view_get_edge_snap_box(struct view *view, struct output *output,
...
+static void
+view_discover_output(struct view *view)
```
Consolatis, line 111: Is this comment still valid, e.g. do we actually need to know if a view is on multiple outputs?

src/view.c
```diff
@@ -156,7 +183,14 @@ view_moved(struct view *view)
...
 		view->current.x, view->current.y);
-	view_discover_output(view);
+	/*
+	 * Only floating views change output when moved. Non-floating
+	 * views (maximized/tiled/fullscreen) are tied to a particular
+	 * output when they enter that state.
+	 */
+	if (view_is_floating(view)) {
+		view_discover_output(view);
+	}
```
Consolatis, line 193: I didn't investigate this in detail yet, just wanted to point out that this *might* be an issue for interactive move from tiled state because we delay `view_set_untiled(view)`, I am not quite sure right now if `view_moved()` is called before the tiling state is reset.

Consolatis:
I had a quick test nested and views in various "special" states migrated fine when closing one of the virtual outputs.
Also tried opening new windows with no outputs left and apart from some error logs everything seems fine. This whole PR looks like a really nice improvement and we finally get rid of those `view->current.x/y` hacks to define the initial output.

include/view.h
jlindgren90: I'll add a comment, seems worth explaining.

src/view.c
jlindgren90: I'm good with removing this comment. It seems less relevant now.

johanmalm:
LGTM.

I note Consolatis's comments from yesterday, but otherwise seems ready to merge.

jlindgren90:
I added a comment for `view_center` and deleted the obsolete one for `view_discover_output`. I haven't had a chance yet to test moving a view from the tiled state (currently spending time debugging an amdgpu driver regression).

src/view.c
johanmalm: I've tested to/from tiled state with mouse/keyboard. Works fine.

PR #802, opened 2023-02-24
Tidy up xdg view destroy & make request_activate listener xwayland only

johanmalm:


johanmalm:
Just some minor tidying up. No functional change intended.

src/xdg.c
```diff
@@ -114,14 +121,26 @@ handle_destroy(struct wl_listener *listener, void *data)
...
+	xdg_toplevel_view->xdg_surface = NULL;
+
+	wl_list_remove(&view->map.link);
+	wl_list_remove(&view->unmap.link);
+	wl_list_remove(&view->request_move.link);
+	wl_list_remove(&view->request_resize.link);
+	wl_list_remove(&view->request_minimize.link);
+	wl_list_remove(&view->request_maximize.link);
+	wl_list_remove(&view->request_fullscreen.link);
+	wl_list_remove(&view->set_title.link);
```
Consolatis, line 137: We could also remove the common listeners in `view_destroy()` instead.

johanmalm:
Did you also want me to `s/xdg_toplevel_view/xdg_view/` throughout the code-base?
I think you mentioned it the other day, and I agree. There are only two types of views (that we care about anyway - in the current Wayland eco-system) and they are `xwayland` and `xdg`.

src/xwayland.c
```diff
@@ -279,8 +279,10 @@ handle_request_configure(struct wl_listener *listener, void *data)
 static void
 handle_request_activate(struct wl_listener *listener, void *data)
 {
-	struct view *view = wl_container_of(listener, view, request_activate);
-	assert(view);
+	struct xwayland_view *xwayland_view =
+		wl_container_of(listener, xwayland_view, request_activate);
+	assert(xwayland_view);
```
Consolatis, line 284: Kind of unrelated to this specific PR but all these asserts after `wl_container_of()` seem superfluous. I actually might have been the one introducing them a while ago.

Consolatis:
> Did you also want me to s/xdg_toplevel_view/xdg_view/ throughout the code-base?
I think you mentioned it the other day, and I agree. There are only two types of views (that we care about anyway - in the current Wayland eco-system) and they are xwayland and xdg.

I don't think that is necessary but it wouldn't harm either. Just wanted to point that out the other day mainly for new code because its just less to write.

src/xdg.c
johanmalm: Done

src/xwayland.c
johanmalm: :smile: Removed.

johanmalm:
> > Did you also want me to s/xdg_toplevel_view/xdg_view/ throughout the code-base?
> > I think you mentioned it the other day, and I agree. There are only two types of views (that we care about anyway - in the current Wayland eco-system) and they are xwayland and xdg.
> 
> I don't think that is necessary but it wouldn't harm either. Just wanted to point that out the other day mainly for new code because its just easier to write.

Okay - I'll leave it then. Keep churn low. Might also simplify rebasing John's WIP PR.

johanmalm:
This is ready to go then.

Consolatis:
Nice refactor :)

PR #803, opened 2023-02-25
view: Anchor right/bottom edge only when resizing via top/left edge

jlindgren90:
Currently, we anchor the right/bottom edge of the view whenever the top/left edge is moving (`current.x/y != pending.x/y`). Doing so doesn't make much sense when the right/bottom edge is also moving. In that case it's probably best to move the view (or at least its top/left corner) directly to its final position.

The most noticeable effect of this change is with views that don't accept their requested size exactly when tiled or maximized (examples: havoc, xfce4-terminal). Previously, their right-bottom corner would be aligned with the screen edge, leaving gaps on the left and top. Now the top-left corner will be aligned and the gaps will be on the right and bottom. This is still not ideal, but IMHO less surprising to the user.

I don't fully understand all of the `update_required` logic in xdg's `handle_commit()` so would appreciate a second pair of eyes on that logic.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -7,6 +7,8 @@
 #include "view-impl-common.h"
 #include "workspaces.h"
 
+#define CONFIGURE_TIMEOUT_MS 100
+
 static struct xdg_toplevel_view *
 xdg_toplevel_view_from_view(struct view *view)
 {
@@ -78,15 +80,51 @@ handle_commit(struct wl_listener *listener, void *data)
 		|| current->height != size.height;
 
 	uint32_t serial = view->pending_configure_serial;
-	if (serial > 0 && serial >= xdg_surface->current.configure_serial) {
+	if (serial > 0 && serial == xdg_surface->current.configure_serial) {
+		assert(view->pending_configure_timeout);
+		wl_event_source_remove(view->pending_configure_timeout);
+		view->pending_configure_serial = 0;
+		view->pending_configure_timeout = NULL;
 		update_required = true;
-		if (serial == xdg_surface->current.configure_serial) {
-			view->pending_configure_serial = 0;
-		}
 	}
+
 	if (update_required) {
 		view_impl_apply_geometry(view, size.width, size.height);
 	}
+}
+
+static int
+handle_configure_timeout(void *data)
+{
+	struct view *view = data;
+	assert(view->pending_configure_serial > 0);
+	assert(view->pending_configure_timeout);
+
+	const char *app_id = view_get_string_prop(view, "app_id");
+	wlr_log(WLR_ERROR, "client (%s) did not respond to configure request "
+		"in %d ms", app_id, CONFIGURE_TIMEOUT_MS);
+
+	wl_event_source_remove(view->pending_configure_timeout);
+	view->pending_configure_serial = 0;
+	view->pending_configure_timeout = NULL;
+
+	view_impl_apply_geometry(view, view->current.width,
+		view->current.height);
+
+	return 0; /* ignored per wl_event_loop docs */
+}
+
+static void
+set_pending_configure_serial(struct view *view, uint32_t serial)
+{
+	view->pending_configure_serial = serial;
+	if (!view->pending_configure_timeout) {
+		view->pending_configure_timeout =
+			wl_event_loop_add_timer(view->server->wl_event_loop,
+				handle_configure_timeout, view);
+	}
+	wl_event_source_timer_update(view->pending_configure_timeout,
+		CONFIGURE_TIMEOUT_MS);
 }
 
 static void
@@ -117,6 +155,11 @@ handle_destroy(struct wl_listener *listener, void *data)
 	/* Remove xdg-shell view specific listeners */
 	wl_list_remove(&xdg_toplevel_view->set_app_id.link);
 	wl_list_remove(&xdg_toplevel_view->new_popup.link);
+
+	if (view->pending_configure_timeout) {
+		wl_event_source_remove(view->pending_configure_timeout);
+		view->pending_configure_timeout = NULL;
+	}
 
 	view_destroy(view);
 }
@@ -215,7 +258,7 @@ xdg_toplevel_view_configure(struct view *view, struct wlr_box geo)
 
 	view->pending = geo;
 	if (serial > 0) {
-		view->pending_configure_serial = serial;
+		set_pending_configure_serial(view, serial);
 	} else if (view->pending_configure_serial == 0) {
 		/*
 		 * We can't assume here that view->current is equal to
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -74,31 +74,18 @@ handle_commit(struct wl_listener *listener, void *data)
 	wlr_xdg_surface_get_geometry(xdg_surface, &size);
 
 	struct wlr_box *current = &view->current;
-	struct wlr_box *pending = &view->pending;
-	bool update_required = false;
-
-	if (current->width != size.width || current->height != size.height) {
-		update_required = true;
-		current->width = size.width;
-		current->height = size.height;
-	}
+	bool update_required = current->width != size.width
+		|| current->height != size.height;
 
 	uint32_t serial = view->pending_configure_serial;
 	if (serial > 0 && serial >= xdg_surface->current.configure_serial) {
-		if (current->x != pending->x) {
-			update_required = true;
-			current->x = pending->x + pending->width - size.width;
-		}
-		if (current->y != pending->y) {
-			update_required = true;
-			current->y = pending->y + pending->height - size.height;
-		}
+		update_required = true;
 		if (serial == xdg_surface->current.configure_serial) {
 			view->pending_configure_serial = 0;
 		}
 	}
 	if (update_required) {
-		view_moved(view);
+		view_impl_apply_geometry(view, size.width, size.height);
 	}
 }
 
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -101,7 +101,7 @@ handle_configure_timeout(void *data)
 	assert(view->pending_configure_timeout);
 
 	const char *app_id = view_get_string_prop(view, "app_id");
-	wlr_log(WLR_ERROR, "client (%s) did not respond to configure request "
+	wlr_log(WLR_INFO, "client (%s) did not respond to configure request "
 		"in %d ms", app_id, CONFIGURE_TIMEOUT_MS);
 
 	wl_event_source_remove(view->pending_configure_timeout);
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -217,9 +217,16 @@ xdg_toplevel_view_configure(struct view *view, struct wlr_box geo)
 	if (serial > 0) {
 		view->pending_configure_serial = serial;
 	} else if (view->pending_configure_serial == 0) {
-		view->current.x = geo.x;
-		view->current.y = geo.y;
-		view_moved(view);
+		/*
+		 * We can't assume here that view->current is equal to
+		 * view->pending because some clients (e.g. terminals)
+		 * refuse to accept the exact size we requested. To
+		 * account for the size difference and avoid visual
+		 * glitches during resize, we apply the same position
+		 * adjustments here as in handle_commit().
+		 */
+		view_impl_apply_geometry(view, view->current.width,
+			view->current.height);
 	}
 }
 
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -109,24 +109,16 @@ handle_commit(struct wl_listener *listener, void *data)
 	/* Must receive commit signal before accessing surface->current* */
 	struct wlr_surface_state *state = &view->surface->current;
 	struct wlr_box *current = &view->current;
-	struct wlr_box *pending = &view->pending;
 
-	if (current->width == state->width && current->height == state->height) {
-		return;
+	/*
+	 * If there is a pending move/resize, wait until the surface
+	 * size changes to update geometry. The hope is to update both
+	 * the position and the size of the view at the same time,
+	 * reducing visual glitches.
+	 */
+	if (current->width != state->width || current->height != state->height) {
+		view_impl_apply_geometry(view, state->width, state->height);
 	}
-
-	current->width = state->width;
-	current->height = state->height;
-
-	if (current->x != pending->x) {
-		/* Adjust x for queued up configure events */
-		current->x = pending->x + pending->width - current->width;
-	}
-	if (current->y != pending->y) {
-		/* Adjust y for queued up configure events */
-		current->y = pending->y + pending->height - current->height;
-	}
-	view_moved(view);
 }
 
 static void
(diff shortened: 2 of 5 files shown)


johanmalm:
I think this looks fine.

I wish I could find a simpler way to organise that part of the code - to make it all really obvious and intuitive - but I can't think of one.

A couple of minor thoughts:

- Consider moving `view_moved()` to `view-impl-common.h` also.

- Consider moving the comment above `view_impl_apply_geometry()` to header file to increase consisteny in documenting functions.

Suggestion for comment fix:

```
index 845812b..5862fe4 100644
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -82,8 +82,15 @@ handle_commit(struct wl_listener *listener, void *data)
 	}
 
 	/*
-	 * FIXME: please add a comment to explain what's going on with
-	 * the configure serials here.
+	 * We check the serials here to ensure that we only change the view's
+	 * x/y position when the surface is ready to commit with pending
+	 * width/height (rather than at some other commit).
+	 *
+	 * The serial check also ensures that we only resize to width/height
+	 * of the last call to wlr_xdg_toplevel_set_size().
+	 *
+	 * Moves without corresponding resize are processed directly in
+	 * xdg_toplevel_view_configure()
 	 */
 	uint32_t serial = view->pending_configure_serial;
 	if (serial > 0 && serial >= xdg_surface->current.configure_serial) {
```

johanmalm:
Would this be valid following your re-factoring - or does it make a functional change?

```
index 845812b..70b9c43 100644
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -87,9 +87,7 @@ handle_commit(struct wl_listener *listener, void *data)
 	 */
 	uint32_t serial = view->pending_configure_serial;
 	if (serial > 0 && serial >= xdg_surface->current.configure_serial) {
-		if (current->x != pending->x || current->y != pending->y) {
-			update_required = true;
-		}
+		update_required = true;
 		if (serial == xdg_surface->current.configure_serial) {
 			view->pending_configure_serial = 0;
 		}
```

src/xwayland.c
```diff
@@ -109,24 +109,16 @@ handle_commit(struct wl_listener *listener, void *data)
...
-		current->x = pending->x + pending->width - current->width;
-	}
-	if (current->y != pending->y) {
-		/* Adjust y for queued up configure events */
-		current->y = pending->y + pending->height - current->height;
+	/*
+	 * If there is a pending move/resize, wait until the surface
+	 * size changes to update geometry. The hope is to update both
+	 * the position and the size of the view at the same time,
+	 * reducing visual glitches.
```
Consolatis, line 117: > FIXME: please add a comment to explain what's going on with the configure serials here

It is basically doing the same thing as this comment explains for xwayland. The difference is that the wayland protocol with its configure serial is deterministic while the X11 variant requires guesswork based on size / position.

As far as I understand `view->pending_configure_serial` can have 3 states:
- `0`: no configure request pending
- `< xdg_surface->current.configure_serial`: some configure request still pending
- `== xdg_surface->current.configure_serial`: some configure request was pending but we now catched up and thus reset `view->pending_configure_serial` to `0`

Feel free to correct me when I got something wrong.

Consolatis:
Regarding the `update_required` part in the xdg commit handler: That handler is called for completely unrelated changes (other than size changes) as well (try adding a wlr_log to that function and just type in foot for example).

There might be a flag we could check for which tells us that the current commit confirms a new geometry but I didn't look into it at all. That way we could just return early and wouldn't have to deal with any of the further checks.

jlindgren90:
> Would this be valid following your re-factoring - or does it make a functional change?
> 
> ```
> diff --git a/src/xdg.c b/src/xdg.c
> index 845812b..70b9c43 100644
> --- a/src/xdg.c
> +++ b/src/xdg.c
> @@ -87,9 +87,7 @@ handle_commit(struct wl_listener *listener, void *data)
>  	 */
>  	uint32_t serial = view->pending_configure_serial;
>  	if (serial > 0 && serial >= xdg_surface->current.configure_serial) {
> -		if (current->x != pending->x || current->y != pending->y) {
> -			update_required = true;
> -		}
> +		update_required = true;
>  		if (serial == xdg_surface->current.configure_serial) {
>  			view->pending_configure_serial = 0;
>  		}
> ```

I think that works; `view_impl_apply_geometry()` should be a no-op if nothing changed, so no need to duplicate the checks it already does. (Nitpick: the name `update_required` might no longer be accurate in that case.)

jlindgren90:
I'm still pondering over the configure serials & will respond back when I have more time. The other suggestions sound good; I will implement those later on as well.

jlindgren90:
So reading over the comments, I think I am still confused by one of the 3 states that @Consolatis mentioned:

- `view->pending_configure_serial > xdg_surface->current.configure_serial`

This should mean that there is still a pending configure request.
Now, @johanmalm's comment says:

> The serial check also ensures that we only resize to width/height of the last call to wlr_xdg_toplevel_set_size().

That makes it sound like we should wait to update the geometry if there is still a pending configure request. But in the current code, this state will meet the `>=` condition and set `update_required = true`.

I guess I don't understand why we have the outer `>=` condition at all; why not just a single `==` condition?

```
uint32_t serial = view->pending_configure_serial;
if (serial == xdg_surface->current.configure_serial) {
	update_required = true;
	view->pending_configure_serial = 0;
}
```

What am I missing here?

jlindgren90:
> Consider moving view_moved() to view-impl-common.h also.

I ran into a problem with this: `view_moved()` calls `view_discover_output()`, which will become private to `view.c` once #798 is merged. I'm not sure of the best way to proceed here:

- Leave `view_moved()` where it is?
- Make `view_discover_output()` public again?
- Move `view_discover_output()` to `view-impl-common.c` also?

johanmalm:
> This should mean that there is still a pending configure request. Now, @johanmalm's comment says:
> 
> > The serial check also ensures that we only resize to width/height of the last call to wlr_xdg_toplevel_set_size().
> 

I think I was wrong with that comment :smile: Whoops.

Let me try to analyze the two situations.

1. `pending-serial == configure-serial`

This means that the commit relates to the latest size request (by `xdg_toplevel_view_configure()`. In this case, life is simple and current code works fine.

2. `pending-serial > configure-serial`

This means the surface commit relates to an old size request. In other words, there has been a more recent call to `xdg_toplevel_view_configure()` but the surface isn't yet ready to to commit to that size. I suspect this is the source of some glitches/flickering, because in this situation we shouldn't use `view->pending` in `view_impl_apply_geometry()` but rather the geometry associated with the commit.

Not sure if it's best to get it from `wlr_xdg_surface->current.geometry` or `wlr_xdg_surface_get_geometry()`.

johanmalm:
Just another thought...

In the commit-handler, could we just use a `static uint32_t last_configure_serial` and use that to return early if `xdg_surface->current.configure_serial <= last_configure_serial`

We'd have to do `<=` rather than `==` because a configure event relating to another surface might have bumped the serial.

I think this approach could avoid all the `view->pending_configure_serial` stuff.

I hope I'm thinking about all this straight... haven't hacked on this part of the code for a while. Apologies if it's a bad steer.

jlindgren90:
> This means the surface commit relates to an old size request. In other words, there has been a more recent call to xdg_toplevel_view_configure() but the surface isn't yet ready to to commit to that size. I suspect this is the source of some glitches/flickering, because in this situation we shouldn't use view->pending in view_impl_apply_geometry() but rather the geometry associated with the commit.

We do have the separate width/height inequality check which should update the geometry appropriately for the old size request/commit in that case.

It seems to me that with the width/height inequality check and a simple equality check for the most recent serial, we should have all our bases covered. How about if I were to push a new commit with that change (the code snippet from a few comments back)?

johanmalm:
See video below. Watch right-hand edge of Firefox window during resize. Occasionally the surface goes outside the window. I suspect it's because we change view->current.x before the surface is ready (i.e. before it's committed to that width)

We set `view->pending` here https://github.com/labwc/labwc/blob/d571b0b28aab8fc117756cf03e83868d306ee7a3/src/xdg.c#L229

...but on the next commit may be processing an older geo.

[recording.webm](https://user-images.githubusercontent.com/6955353/221420573-111a63ef-cc2a-4026-b087-2d84a820d3a0.webm)

jlindgren90:
I can't load the video right now, but is that behavior with or without 4b3803d8f706? Does 4b3803d8f706 make any difference (good or bad)?

johanmalm:
No, it makes no difference. I just checkout out `0.6.0` and it's the same there, so looks like an old problem.

Consolatis:
> I can't load the video right now,

Screenshot of the video to show the issue:
20230226_19h59m43s_grim

So this seems to happen when rapidly resizing a XDG window from the left edge and the application being resized being slow for whatever reason.

I didn't investigate this in more depth but I think overwriting `view->pending` is the correct thing to do. So it always has the latest state of what should be the geometry once all queued configure events are synced up.

Maybe we should just not move at all (e.g. not set `update_required`) as long as a configure request is still pending instead of trying to adjust for the difference between wanted (pending) and current (`wlr_xdg_surface_get_geometry()`) state.

johanmalm:
I think we're onto a detail that would be really good to get right. I suspect we've never handled this correctly.
Suggest we dwell a pause with this PR until we understand how we want to handle xdg-commits (in case we don't want to share the code with xwayland).
Maybe we should take a leaf out of sway's transaction approach: https://github.com/swaywm/sway/blob/master/include/sway/desktop/transaction.h

I tried overwriting `view->pending` at commit time, but that creates chaos :smile:

Also tried not moving until whilst configure events are pending (we must think alike) - but on continuous resizing that creates a real delay in updating the SSD.

I can't quite see my way through this one at the moment. Maybe we need a `wl_list` of pending states - with associated serials or something.

jlindgren90:
The Firefox resize glitch appears to be a bug in Firefox. It happens under Weston also. (Under Weston, it's particularly strange because the CSD titlebar - drawn by Firefox! - resizes correctly but the window contents below the titlebar "jump" outside the expected bounds, just like in the video above.)

As background, xdg-shell surfaces are allowed to extend outside their window geometry. In other words it's perfectly legal for an xdg-shell application to paint outside its window. (That seems like a really bad idea to me, but I guess the use-case is for client-side decorations to be able to include shadows outside the window geometry. Whatever.)

Try adding the following lines to `handle_commit()`:
```
        struct wlr_box size;
        wlr_xdg_surface_get_geometry(xdg_surface, &size);
 
+       struct wlr_box extends;
+       wlr_surface_get_extends(xdg_surface->surface, &extends);
```

If you then compare `size` and `extends` you'll see that they don't match during resize. It appears that Firefox commits the new window geometry first and then updates the surface shortly afterward, which is why we see the window contents extending beyond the frame for a brief moment.

I really don't think there's a good way to solve/work around the Firefox resize issue in labwc. (I tried using the width/height from `wlr_surface_get_extends()` in `handle_commit()` but that just creates worse issues.) Also, the issue seems only tangentially related to this pull request, and existed before it. So I'd suggest that we don't let it hold up fixing the separate window anchoring issue that this changeset does fix.

I did find a remaining resize issue with xfce4-terminal (unrelated to the Firefox issue, but similar symptoms) that I'll add a fix for.

johanmalm:
John, sorry having a crazy week. Do you mind if I leave it to the weekend to play with Firefox a bit further?

jlindgren90:
That's fine, I need to rework this a bit still anyway.

jlindgren90:
There is another problem (in the current code) which can be seen when attempting to tile an xdg-shell view that doesn't allow resizing. Example: the "about" window of xfce4-terminal. In that case, we never get a commit acknowledging the configure-request, and it remains "pending" indefinitely until an unrelated commit. As a result, we delay moving the "about" window to its tiled location, until the mouse cursor is moved over it -- then it suddenly jumps.

This presumably started occurring after ecd03b70ef93616b0de73b795a5c05180dadcc3c (haven't verified). To fix it, I would suggest a timeout (100 ms?) after which, if we haven't received the acknowledgement commit, we move the view anyway.

src/xdg.c
```diff
@@ -78,17 +80,53 @@ handle_commit(struct wl_listener *listener, void *data)
...
+static int
+handle_configure_timeout(void *data)
+{
+	struct view *view = data;
+	assert(view->pending_configure_serial > 0);
+	assert(view->pending_configure_timeout);
+
+	const char *app_id = view_get_string_prop(view, "app_id");
+	wlr_log(WLR_ERROR, "client (%s) did not respond to configure request "
+		"in %d ms", app_id, CONFIGURE_TIMEOUT_MS);
```
Consolatis, line 105: I think this could be a `WLR_INFO` instead.

src/xdg.c
```diff
@@ -118,6 +156,11 @@ handle_destroy(struct wl_listener *listener, void *data)
 	wl_list_remove(&xdg_toplevel_view->set_app_id.link);
 	wl_list_remove(&xdg_toplevel_view->new_popup.link);
 
+	if (view->pending_configure_timeout) {
+		wl_event_source_remove(view->pending_configure_timeout);
+		view->pending_configure_timeout = NULL;
```
Consolatis, line 161: Should we reset the serial to `0` as well here?

Consolatis:
> John, sorry having a crazy week. Do you mind if I leave it to the weekend to play with Firefox a bit further?

If the surface is really bigger than the geometry, we could also temporarily clip the surface once wlr_scene supports clipping.
I am not sure how that would affect drop shadows and friends in the CSD case though..

src/xdg.c
jlindgren90: It shouldn't matter since the view is being destroyed (even setting `view->pending_configure_timeout = NULL` should be technically unnecessary, and is only there as a safeguard).

Consolatis: Right.. its the destroy handler. Somehow missed that earlier.

jlindgren90:
> If the surface is really bigger than the geometry, we could also temporarily clip the surface once wlr_scene supports clipping. I am not sure how that would affect drop shadows and friends in the CSD case though..

There might even be legitimate use-cases for allowing subsurfaces to extend outside SSD views in some cases. There was talk e.g. of using subsurfaces to implement QDockWidget (see https://bugreports.qt.io/browse/QTBUG-87332) though I don't think anything has been implemented yet.

jlindgren90:
> The difference is that the wayland protocol with its configure serial is deterministic while the X11 variant requires guesswork based on size / position.

Off-topic: there is actually a very similar "configure serial" protocol for X11, _NET_WM_SYNC_REQUEST:
https://specifications.freedesktop.org/wm-spec/1.3/ar01s06.html

AFAICT all the major toolkits and WMs support it. It doesn't guarantee "frame perfect" resize, but it addresses the classic X11 problem where the WM queues up a whole bunch of configure requests faster than the application can redraw. I've looked at the Openbox implementation and it doesn't look like it would really be that hard to add support for the protocol to wlroots + labwc.

src/xwayland.c
jlindgren90: I removed the FIXME; I think I understand how the configure serials work now. Thanks!

johanmalm:
LGTM. Happy to merge unless there's anything else you want to do.

In terms of the Firefox issue, I've tried creating a list of serials and using those `wlr_box`s. Have also tried reading `wlr_surface->current.width` and `wlr_surface->current.buffer_width`directly on commit, but nothing seems to know Firefox's width/height.

Interestingly `Firefox` behaves the same way on `sway`.

jlindgren90:
Thanks. I pushed one more commit to change `WLR_ERROR` to `WLR_INFO` as @Consolatis suggested. With that, I think I'm happy with where this stands, so marking it ready.

johanmalm:
Thanks!

PR #811, opened 2023-03-03
Add ToggleKeybinds action

### LINKED ISSUE #738: Temporarily disable all keybinds to control remote systems or VMs
River seems to implement "input-modes" e.g. one defines a keybind to enter that "input-mode" and then have a set of keybinds defined within that input-mode including one to exit the mode again. One use-case for that is to temporarily disable all keybinds like `A-Tab` when controlling a VM / remote system via VNC or similar.

We are basically doing something similar for the `A-Tab` OSD already which handles all key events as long as its open.
I don't think having a fully fledged implementation of "input-modes" is something that really fits labwc but for this specific use-case of temporarily disabling all keybinds (other than an "exit-mode" keybind) we could use the same design.

The end goal here would be to be able to use for example `A-F4` or `A-Tab` within a VNC client / VM window or nested compositor.

This would also remove the need to support the [keyboard-shortcuts-inhibit](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/unstable/keyboard-shortcuts-inhibit/keyboard-shortcuts-inhibit-unstable-v1.xml) protocol which allows wayland applications to enable a mode like that and thus [prevent normal usage of the system](https://bugs.chromium.org/p/chromium/issues/detail?id=1338554). The difference here is that with this implementation it is the user who decides when to disable / re-enable the keybinds and it works for all applications.

We could even bind the state to a view/surface so that when the user enables this mode while the keyboard focus is on a given window and then uses the mouse to switch the focus to another window the mode is automatically reset (e.g. all keybinds work again) and when then switching back to the original window the mode is back in action.

### LINKED ISSUE #810: How to allow app to capture keyboard shortcuts?
I wonder if it is possible to use keyboard shortcuts in a virtual machine. For example, with a Windows VM, I cannot Alt+Tab, or some keybindings with Win+...

With gnome, this can be done easily. It shows a dialog to ask if i want to capture shortcut. I understand that labwc doesn't have any kind of dialog to confirm like that, so I think this feature can be set with app-specific configuration. I've read every documents available for labwc, as well as a lot of thread about using virt-manager or remote-viewer under Wayland, but have no luck...

I want to share that I have begin with gnome, then switch to i3, then sway, finally here. Now labwc is a thing that I will live with as long as I can still sit in front of a computer. That explains how awesome this project is to me.

I'm looking forward to your response. Thank you guys a lot.

Consolatis:
This can be used to better control Virtual Machines, VNC clients, nested compositors or similar. All keybinds other than `ToggleKeybinds` itself are disabled when first called, a 2nd call will restore handling of all keybinds.

Fixes #738
Fixes #810

---
I only had a quick test nested with this so a proper test would be appreciated as I don't really have that much use-cases for this feature myself. To test this it should be enough to add something like this to `rc.xml`:
```xml
<keybind key="Scroll_Lock">
	<action name="ToggleKeybinds" />
</keybind>
```

This PR does not add an always-on-top status indicator about the keybinds currently being disabled, that could be done as well but would require a bit more code to properly react to output layout changes. Thus I am not sure if such an indicator would justify the additional maintenance burden.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -74,6 +74,12 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="ToggleAlwaysOnTop">*
 	Toggle always-on-top of focused window.
 
+*<action name="ToggleKeybinds">*
+	Stop handling keybinds other than ToggleKeybinds itself.
+	This can be used to allow A-Tab and similar keybinds to be delivered
+	to Virtual Machines, VNC clients or nested compositors.
+	A second call will restore all original keybinds.
+
 *<action name="GoToDesktop" to="value" />*
 	Switch to workspace. Supported values are "last", "left", "right" or the
 	full name of a workspace or its index (starting at 1) as configured in
--- a/include/action.h
+++ b/include/action.h
@@ -22,6 +22,8 @@ struct action *action_create(const char *action_name);
 void action_arg_add_str(struct action *action, char *key, const char *value);
 void action_arg_from_xml_node(struct action *action, char *nodename, char *content);
 
+bool actions_contain_toggle_keybinds(struct wl_list *action_list);
+
 void actions_run(struct view *activator, struct server *server,
 	struct wl_list *actions, uint32_t resize_edges);
 void action_list_free(struct wl_list *action_list);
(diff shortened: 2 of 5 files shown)


johanmalm:
LGTM

PR #812, opened 2023-03-05
ToggleKeybinds per window

Consolatis:
Did some more work on the new `ToggleKeybinds` action so that it works per window (view).
I started with storing a list of surfaces in `struct seat` and only inhibiting keybinds when one of those surfaces had focus.
Then switched to storing a list of views instead of surfaces and finally ended up on just a simple `bool` for `struct view` + a counter for `struct seat`. The counter is just a performance optimization so we don't have to check further unless that counter is `> 0`.

I've also added a (pretty strange) SSD status indicator as a proof of concept:
20230305_10h38m04s_grim

I am not sure what the best approach for the status indicator would be, turning the whole border to another color isn't that easy because we'd actually have to redraw the rounded corners as well (as they are also part of the border).
Alternatives I can think of (but different options very welcome):
- we could render some text (like just a `!` or `Keybinds inhibited`) and put it after the title.
- we could load an additional xbm icon and throw it somewhere on the titlebar
- we could render some icon (xbm or cairo, for example some small red dot) on the top right of the actual surface

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -105,6 +105,11 @@ Actions are used in menus and keyboard/mouse bindings.
 	to Virtual Machines, VNC clients or nested compositors.
 	A second call will restore all original keybinds.
 
+	This action will only affect the window that had keyboard focus when
+	the binding was executed. Thus when switching to another window, all
+	the usual keybinds will function again until switching back to the
+	original window. There can be multiple windows with this mode set.
+
 *<action name="FocusOutput" output="HDMI-A-1" />*
 	Give focus to topmost window on given output and warp the cursor
 	to the center of the window. If the given output does not contain
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -78,6 +78,10 @@ labwc-config(5).
 *window.inactive.border.color*
 	Border color of inactive window
 
+*window.active.indicator.toggled-keybind.color*
+	Status indicator for the ToggleKeybinds action. Can be set to the same
+	value as set for window.active.border.color to disable the status indicator.
+
 *window.active.title.bg.color*
 	Background color for the focused window's titlebar
 
--- a/docs/themerc
+++ b/docs/themerc
@@ -13,6 +13,9 @@ padding.height: 3
 window.active.border.color: #dddad6
 window.inactive.border.color: #f6f5f4
 
+# ToggleKeybinds status indicator
+window.active.indicator.toggled-keybind.color: #ff0000
+
 # window titlebar background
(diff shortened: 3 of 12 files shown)


johanmalm:
I like the look of this.

I vote for adding some text to the title as it keeps it simple. The use-case is fairly specialist, so I don't feel we need to super-user friendly on it. If we add a string in English, we ought to considering i18n, so maybe a simple `!` is enough.

Consider `s/inhibited_keybind_views/nr_inhibited_keybind_views/` or `s/inhibited_keybind_views/nr_views_with_keybinds_inhibited/`. I find `nr_` or `num_` prefix helpful for these types of variables.

As a sidenote:
`ack nr_ src/ | wc -l` gives 42
`ack num_ src/ | wl -l` gives 3

Consolatis:
Actually when looking at the screenshot again, I kind of like that strange status indicator on the top. This only works for SSD windows but that is the same for the title change approach as well. We could keep it that way and add a theme config option for the color.

johanmalm:
> Actually when looking at the screenshot again, I kind of like that strange status indicator on the top. This only works for SSD windows but that is the same for the title change approach as well. We could keep it that way and add a theme config option for the color.

I’m good with that.

Consolatis:
Added `window.active.toggled_keybinds.color` as theme var defaulting to `#ff0000`
I am fine with changing it to something else (name and/or color) but not sure to what.

johanmalm:
Suggest using hyphen in theme variables (e.g. `s/toggled_keybinds/toggled-keybinds/`) to be consistent with `window-switcher`

Maybe we could use a simpler name like `window.keybind-inhibit.color`, `window.keygrab.color` or something

Consolatis:
> Suggest using hyphen in theme variables (e.g. `s/toggled_keybinds/toggled-keybinds/`) to be consistent with `window-switcher`

ACK

> Maybe we could use a simpler name like `window.keybind-inhibit.color`, `window.keygrab.color` or something

They are only shown when the windows has focus so I think we should include `window.active.`. Also the action is called `ToggleKeybinds` and we don't seem to mention `inhibiting keybinds` so its meaning might not be obvious to new users (especially non native english speaking ones). But feel free to overrule me here and I'll change it to `keybind-inhibitor`.

Personally I'd prefer something like `window.active.indicator.toggled-keybind.color`

johanmalm:
> Personally I'd prefer something like `window.active.indicator.toggled-keybind.color`

I'm good with that.

Consolatis:
Changed the theme var name and added a small section to the `ToggleKeybinds` action documentation, describing the per-window behavior. Also rebased and squashed. Once the CI has gone through this is ready to go from my perspective.

PR #814, opened 2023-03-05
Refactor XDG and Xwayland view creation and signal cleanup

Consolatis:
This is a rough sketch for now to gather feedback.

What this does is first and foremost establishing a `view_init(struct view *view)` function to reduce some code duplication when creating new views in `src/xdg.c` and `src/xwayland.c`.
It will also be required for a future xdg <--> x11 bridge which uses it to set up new view signals.

This PR additionally adds 4 new functions to `view->impl`:
- `setup_common_listeners()`
- `setup_specific_listeners()`
- `remove_common_listeners()`
- `remove_specific_listeners()`

`remove_common_listeners()` in both, the xdg and xwayland case, is implemented in `view-impl-common.c` as `view_impl_remove_common_listeners(struct view *view)`.
Currently I don't see how we could further deduplicate the setup of the common signal handlers as they are using different event sources (e.g. `wlr_xwayland_surface`, `wlr_xdg_surface`, `wlr_xdg_toplevel`).

On a related note, I experimented with using a macro to connect a signal to a function a while ago but it didn't really improve the readability of the code that much.


johanmalm:
I like this. It's strikes a good balance between method dispatch (the view-impl) and calls from {xdg,xwayland}.c to view(-impl-common).c. The most obvious alternative approach would be to keep a stronger sense of {con,de}structors within {xdg,xwayland}.c but I think it's harder to avoid code-duplication that way (whilst maintaining readability).

Happy to merge proposed design, but nonetheless, here follow some considerations:

- Move `view_init()` to `view-impl-common.c` if the only calls to it are ever going to come from `xdg.c` and `xwayland.c`. That way we keep `view.c` a more public API that actions, etc. can use. (Bear in mind that I don't have knowledge of the bridge-implementation, so I might be missing something).
- Same for `view_destroy()`. Should that live in `view-impl-common.c`? Having `view_init()` and `view_destroy()` in `view.c` feels like bits of the {con,de}structors are spreading into public API space.

jlindgren90:
Looks like a nice refactor. I was a tiny bit disappointed to see that the line count went up and not down, but I still like how it adds structure and breaks up some long functions. :+1:

Consolatis:
I actually may have gone a bit overboard with the 4 new function pointers.
Maybe just using 2 would be better, e.g.
- `view->impl->setup_listeners()`
- `view->impl->remove_listeners()`

And then have the handlers for these functions call the internal `xxx_{setup,remove}_{common,specific}_listeners()` functions instead.

Re `view_init()` and `view_destroy()` in `view-impl-common.c`:

I think it would theoretically make sense but it deals with cleanups for random parts of the codebase. I've thought about adding a `view->events.destroy` signal for a while, so other parts of the code (like e.g. `seat.c`, `osd.c` or `workspaces.c` could have their own clean-up handlers instead of us throwing everything into the the generic project wide `view_destroy()` function. But that will have its own drawbacks (namely keeping track of all the views and listeners everywhere).

So currently I am not really convinced that `view-impl-common.c` is the right place for `view_destroy()`. But neither am I convinced that `view.c` is the right place. Moving `view_init()` into `view-impl-common.c` sounds fine but then having `_init()` in one place and `_destroy()` in another is also a bit strange.

johanmalm:
> I actually may have gone a bit overboard with the 4 new function pointers. Maybe just using 2 would be better, e.g.
> 
>     * `view->impl->setup_listeners()`
>     * `view->impl->remove_listeners()`
> 

Good idea. Feels cleaner.

Consider calling them `listeners_{init,finish}`
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/CONTRIBUTING.md#constructiondestruction-functions

> I think it would theoretically make sense but it deals with cleanups for random parts of the codebase.

It's good to minimize coupling (between objects, translation units and so on) - but in this case the amount of coupling remains the same regardless of the location of `view_init()` and `view_destroy()`.

Also, `view-impl-common.c` already contains calls to other translation units, so the dream of complete encapsulation is already shattered :smile:

So, in my mind the only thing to be gained by shifting some stuff to `view-impl-common.c` is a cleaner `view.c`

Anyway, not a big deal for me, but I enjoy the discussion.

Consolatis:
> the amount of coupling remains the same regardless of the location of view_init() and view_destroy().

true

> Anyway, not a big deal for me, but I enjoy the discussion.

same :)

include/view-impl-common.h
```diff
@@ -20,4 +20,9 @@ void view_impl_map(struct view *view);
  */
 void view_impl_apply_geometry(struct view *view, int w, int h);
 
+void view_init(struct view *view);
+void view_destroy(struct view *view);
```
Consolatis, line 24: - rebased
- moved `view_init()` and `view_destroy()` to `src/view-impl-common.c` (which required to add a bunch of new includes there)
- reduced the additional `struct view_impl` function pointers to 2: `listeners_init()` and `listeners_remove()`. I decided against using `listeners_finish()` for now because the term `finished` really can mean a buch of things and `remove` is pretty clear in this case. I'd could change it to `listeners_finish()` though if it is really necessary (but would prefer not to).

Should we rename `view_init()` / `view_destroy()` to `view_impl_init()` / `view_impl_destroy()` now that they have been moved? They are only called from `src/xdg.c` and `src/xwayland.c`.

johanmalm: Looks good. 

Yes, `view_impl_{init,destroy}` is better.

Consolatis:
This is pretty much outdated and could be implemented faster by rewriting it. If somebody feels strongly about this feel free to do so.

PR #815, opened 2023-03-05
src/workspaces.c: update cursor focus

### LINKED ISSUE #807: Apparently missing cursor_update_focus() when switching workspace
> `<tprepper>` found the most minor bug ever. if you hover over a place where you can resize a window, it changes the cursor to a resizing cursor icon. however, if you move to another workspace it will not update the cursor icon back to normal until you move the cursor

> `<Consolatis>` tprepper: right.. guess we are missing a single cursor_update_focus() call when switching workspaces. Should be easy to fix.

Consolatis:
Fixes #807

### DIFF
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -271,6 +271,12 @@ workspaces_switch_to(struct workspace *target)
 
 	/* And finally show the OSD */
 	_osd_show(server);
+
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks!

PR #816, opened 2023-03-05
Add `FocusOutput` action

### LINKED ISSUE #806: Enhancement: bind to focus monitor
Hi! So I am trying to rice this stuff, as I really really like it, but with my triple monitor setup, keybinds to focus monitors (and have the cursor spawn there) are absolutely crucial to my workflow. Would you like to add this, by any chance?

Consolatis:
Fixes #806

---
Gave this a try but can't properly test it nested, also didn't check if there is a Openbox action relating to this so this is a Draft for now until somebody verifies this actually works as intended.
If it is deemed out-of-scope for labwc feel free to just close the PR.

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -268,6 +268,46 @@ desktop_focus_topmost_mapped_view(struct server *server)
 	desktop_move_to_front(view);
 }
 
+void
+desktop_focus_output(struct output *output)
+{
+	if (!output_is_usable(output) || output->server->input_mode
+			!= LAB_INPUT_STATE_PASSTHROUGH) {
+		return;
+	}
+	struct view *view;
+	struct wlr_scene_node *node;
+	struct wlr_output_layout *layout = output->server->output_layout;
+	struct wl_list *list_head =
+		&output->server->workspace_current->tree->children;
+	wl_list_for_each_reverse(node, list_head, link) {
+		if (!node->data) {
+			continue;
+		}
+		view = node_view_from_node(node);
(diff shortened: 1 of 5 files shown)


johanmalm:
Happy to add this to scope. Whilst the demand is probably low, it looks like pretty simple and self-contained code which is highly unlikely to cause a maintainability issue.

I'm not aware of Openbox supporting this sort of thing.

I've not tested yet. Will have to be another day, unless someone else can verify.

src/desktop.c
```diff
@@ -268,6 +268,43 @@ desktop_focus_topmost_mapped_view(struct server *server)
 	desktop_move_to_front(view);
 }
 
+void
+desktop_focus_output(struct output *output)
+{
+	if (!output || output->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
```
jlindgren90, line 274: Maybe replace `!output` with `!output_is_usable(output)` to prevent trying to warp the cursor to a disabled output?

jlindgren90: Hmm, I missed the `output_is_usable()` check in `output_from_name()`. Still it might not hurt to check again here, to make `desktop_focus_output()` more independent.

Consolatis: Added the additional check.

Consolatis:
Added the additional check suggested by @jlindgren90
@EysseW [confirmed](https://github.com/labwc/labwc/issues/806#issuecomment-1456554099) it working as intended so this is ready to go from my POV.

PR #819, opened 2023-03-07
Prevent rare crash due to layering move / resize / menu operations

### LINKED ISSUE #817: Crash due to overwriting server->input_mode
```
<kode54> please test this game demo on Steam
<kode54> https://store.steampowered.com/app/1522810/Shield_Cat/
<kode54> I was running that by setting to use proton for everything so it would let me install it
<kode54> then ran it with ge proton 49
<kode54> I ran it in windowed mode and used a bluetooth DualShock 4 controller
<kode54> when I quit the game, labwc crashed
<kode54> and then the uwsm setup didn't tear down correctly as a result, so I had to reboot to log in again
<kode54> Mar 05 20:28:32 mrgency labwc[54233]: labwc: ../labwc/src/menu/menu.c:843: menu_close_root: Assertion `server->input_mode == LAB_INPUT_STATE_MENU' failed.
<Consolatis> uh? that is a new one
<Consolatis> we didn't change anything around there for ages, will have a look later today
<Consolatis> thanks for the pointer
<Consolatis> oh, i see. could reproduce. same happens when opening a client menu, then calling for example a Move keybind and finally closing that window
<Consolatis> the reason that it didn't show up before is likely that nobody uses move or resize keybinds in the first place and instead uses the mouse
<Consolatis> so the controller input might have be interpreted as some keybind, that is just guesswork though
```

So to reproduce (assumingly) the same issue:
- bind `Move` to some keybind
- open a window/client-menu (e.g. `A-Space` or the left SSD icon)
- move the mouse over some window
- press the `Move` keybind
- server->input_mode gets overwritten due to the move
- close the window owning the client-menu opened earlier
- labwc crashes on the aforementioned assert() because we now have the client menu open but input_mode being something else

Now the question is how we should fix this:
- Just remove the assert and live with `server->input_mode` potentially getting overwritten
- Prevent modifications of server->input when its not `_PASSTHROUGH`, e.g. return early in `show_menu()` and `interactive_begin()`
- Stop setting `server->input_mode` directly and instead use something like `server_set_state(server, LAB_INPUT_STATE_MOVE)`
  - Have the new function check the previous state and clean up, e.g.
    - for _MOVE and _RESIZE call `interactive_cancel()` (or `interactive_end()`?)
    - for _MENU call `menu_close_root()`

I didn't investigate if similar issues also happen in other combinations, e.g. using a `Move` keybind while currently resizing or spawning a menu via keybind while resizing or moving.

Personally I think I'd prefer the early-out variant but completely open to suggestions.

Consolatis:
Fixes #817

Reported-by: @kode54

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -211,6 +211,12 @@ void action_list_free(struct wl_list *action_list)
 static void
 show_menu(struct server *server, struct view *view, const char *menu_name)
 {
+	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH
+			&& server->input_mode != LAB_INPUT_STATE_MENU) {
+		/* Prevent opening a menu while resizing / moving a view */
+		return;
+	}
+
 	bool force_menu_top_left = false;
 	struct menu *menu = menu_get_by_id(menu_name);
 	if (!menu) {
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -28,6 +28,10 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 	struct seat *seat = &server->seat;
 	struct wlr_box geometry = view->current;
 
+	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
+		return;
+	}
+
 	switch (mode) {
 	case LAB_INPUT_STATE_MOVE:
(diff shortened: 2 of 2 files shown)


johanmalm:
Thanks.

PR #821, opened 2023-03-08
osd: add config option to disable osd

### LINKED ISSUE #672: Hide Alt+Tab Dialog

Flrian:
Fixes #672

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -77,6 +77,10 @@ The rest of this man page describes configuration options.
 	be used with labwc the preferred mode of the monitor is used instead.
 	Default is no.
 
+*<core><cycleViewOSD>* [yes|no]
+	Draw the OSD when cycling between windows.
+	Default is yes.
+
 *<core><cycleViewPreview>* [yes|no]
 	Preview the contents of the selected window when cycling between windows.
 	Default is no.
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -12,6 +12,7 @@
     <gap>0</gap>
     <adaptiveSync>no</adaptiveSync>
     <reuseOutputMode>no</reuseOutputMode>
+    <cycleViewOSD>yes</cycleViewOSD>
     <cycleViewPreview>no</cycleViewPreview>
     <cycleViewOutlines>yes</cycleViewOutlines>
   </core>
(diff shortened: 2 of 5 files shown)


docs/rc.xml.all
```diff
@@ -12,6 +12,7 @@
     <gap>0</gap>
     <adaptiveSync>no</adaptiveSync>
     <reuseOutputMode>no</reuseOutputMode>
+    <cycleViewOSD>yes</cycleViewOSD>
     <cycleViewPreview>no</cycleViewPreview>
     <cycleViewOutlines>yes</cycleViewOutlines>
```
Consolatis, line 17: Honestly I am not a big fan of the naming of these 3 config options but I can't  up with a better name either and breaking user config is also really bad. So even if we come up with a better name we should likely still accept the old ones as alternatives.

For the new config option, maybe we could use `<cycleView>` for now?
I'd say not every user can be assumed to know what `OSD` means.

But that could then potentially be thought of disabling all of the `A-Tab` switching, including previews and outlines.
Soo... not sure what to do here. Thoughts?

CC @johanmalm @jlindgren90

docs/labwc-config.5.scd
```diff
@@ -77,6 +77,10 @@ The rest of this man page describes configuration options.
 	be used with labwc the preferred mode of the monitor is used instead.
 	Default is no.
 
+*<core><cycleViewOSD>* [yes|no]
+	Draw the OSD when cycling between windows.
```
Consolatis, line 81: Maybe s/OSD/window menu/ or "window dialog" or something similar?

Consolatis:
Thanks for the PR! I've tested it nested and it works perfectly for me. Other than the user interfacing naming this LGTM.

I wouldn't have thought I'd like disabling the OSD for my personal setup but will do that once this is merged while still keeping preview / outlines enabled.

docs/labwc-config.5.scd
Consolatis: I'd vote for:
`Draw the window switcher overlay when cycling between windows.`

Consolatis: Hm.. you are right. Just had a look myself and especially `labwc-config.5.scd` has some more references to `OSD`.
Still not really happy with that but it can be changed for all the occurrences via a separate PR.

Consolatis:
This looks ready to go from my perspective.

johanmalm:
Thanks

PR #822, opened 2023-03-08
src/config/rcxml.c: reset default values on Reconfigure

Consolatis:
Reported-by: \<umk3\> via IRC

---
Only slightly tested but confirmed working via IRC.

Original issue was that setting something to a non-default value, `Reconfigure`, then deleting that setting and `Reconfigure` again wouldn't reset the state back to the default. Thus it would differ vs. restarting labwc.

### DIFF
--- a/include/common/font.h
+++ b/include/common/font.h
@@ -5,12 +5,12 @@
 struct lab_data_buffer;
 
 enum font_slant {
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM

PR #824, opened 2023-03-11
protocols: remove redundant wlr-output-management

johanmalm:


Consolatis:
Nice catch. Seems it was not even compiled in `protocols/meson.build` either

johanmalm:
> Nice catch. Seems it was not even compiled in `protocols/meson.build` either

Yes, that's how I spotted it. I don't think it's ever been needed :smile:

Sunderland93:
This protocol is needed for tools like kanshi or wdisplays. This utilities will not work anymore?

johanmalm:
Yes. As Consolatis said. No functionality has changed. I use kanshi every day and it still works fine.
Sorry for poorly worded PR title - should have made it clearer.

PR #826, opened 2023-03-11
view: close popups on focus-change, SSD mouse-event and window switching

johanmalm:
Just submitting for early thoughts/discussion.

I'm sort of puzzled that nobody has reported this before. Try opening a menu and then dragging a window, alt-tab or similar.

- [ ] close xwayland popups
- [ ] close layer-shell popups

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -296,6 +296,16 @@ xdg_toplevel_view_close(struct view *view)
 }
 
 static void
+xdg_toplevel_view_close_popups(struct view *view)
+{
+	struct wlr_xdg_toplevel *toplevel = xdg_toplevel_from_view(view);
+	struct wlr_xdg_popup *popup, *next;
+	wl_list_for_each_safe(popup, next, &toplevel->base->popups, link) {
+		wlr_xdg_popup_destroy(popup);
+	}
+}
+
+static void
 xdg_toplevel_view_maximize(struct view *view, bool maximized)
 {
 	wlr_xdg_toplevel_set_maximized(xdg_toplevel_from_view(view), maximized);
@@ -446,6 +456,7 @@ xdg_toplevel_view_unmap(struct view *view)
 static const struct view_impl xdg_toplevel_view_impl = {
 	.configure = xdg_toplevel_view_configure,
 	.close = xdg_toplevel_view_close,
+	.close_popups = xdg_toplevel_view_close_popups,
 	.get_string_prop = xdg_toplevel_view_get_string_prop,
 	.map = xdg_toplevel_view_map,
 	.set_activated = xdg_toplevel_view_set_activated,
--- a/include/view.h
+++ b/include/view.h
@@ -24,6 +24,7 @@ struct view;
 struct view_impl {
 	void (*configure)(struct view *view, struct wlr_box geo);
 	void (*close)(struct view *view);
+	void (*close_popups)(struct view *view);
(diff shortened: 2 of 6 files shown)


src/xdg.c
```diff
@@ -295,6 +295,16 @@ xdg_toplevel_view_close(struct view *view)
...
 }
 
+static void
+xdg_toplevel_view_close_popups(struct view *view)
+{
+	struct wlr_xdg_toplevel *toplevel = xdg_toplevel_from_view(view);
+	struct wlr_xdg_popup *popup, *next;
+	wl_list_for_each_safe(popup, next, &toplevel->base->popups, link) {
+		wlr_xdg_popup_destroy(popup);
+	}
```
Consolatis, line 305: I am not that familiar with our implementation for popups, does this also handle layered popups?
E.g. a menu spawning a submenu?

johanmalm: Yes, indeed.

jlindgren90:
I've noticed popups not closing also (including labwc's own menus).

> close xwayland popups

I've wondered how we would even do this. X11 popups usually close voluntarily when there is a mouse-event on another window, I believe (not 100% sure). So do we send some kind of a synthetic mouse event to the X server where we know there isn't a window?

Consolatis:
I guess we could also loop over the unmanged window tree.. but I guess there are other unmanged windows that are not popups as well. Maybe we could filter them for actual menu surfaces / tooltips?

johanmalm:
I did try iterating over the unmanaged_surfaces wl_list to close them all using wlr_xwayland_surface_close()

They did close, but the top level window didn’t seem to register it (e.g. style for open menu still showing).

Could try unmap but it’s harder to achieve and I suspect the problem will remain.

I tried with sway, and they have the same problem. Not look at any others yet.

01micko:
7a19184 should solve at least part of this.

johanmalm:
> [7a19184](https://github.com/labwc/labwc/commit/7a19184e5ecba88c330f04150f1d9b0896c959f9) should solve at least part of this.

We've got a choice here. We either (a) close popup-menus on alt-tab/titlebar-drag/etc; or (b) don't allow operations whilst popup menus are open.

Mutter/openbox work like (b).

In openbox, the menu's close on titlebar click - so if you want to drag, you first have to click, then press+drag.

I think I prefer (b)

Suggest we fine a solution to xwayland unmanaged views outside of this PR because I'm pretty short on ideas for that.

Consolatis:
> In openbox, the menu's close on titlebar click - so if you want to drag, you first have to click, then press+drag.

That should be possible for wayland native windows by adding the appropriate view_enums to 7a19184e5ecba88c330f04150f1d9b0896c959f9. Or we could reverse that condition and check for `ctx->type != LAB_SSD_CLIENT`.

For X11 popups.. I don't know what to do either.
- Simply destroying all unmanaged surfaces feels wrong (as there may be other ones than menus).
- AFAIR we also discovered in the past that some X11 clients use actual undecorated toplevel surfaces for submenus, not sure right now. I think it was about (older versions?) of VLC and also tcl/tk based applications like gitk.
- Destroying only popup ones seems to confuse X11 clients.
- Fake injecting a mouse event somewhere as mentioned by @jlindgren90 could work I guess but feels pretty hacky.

Consolatis:
Is this still relevant?

johanmalm:
> Is this still relevant?

Yes. Will pick this up again at some point.

PR #827, opened 2023-03-11
Config window switcher

johanmalm:
@Flrian - review appreciated
...and @Consolatis of course :smile:

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -588,7 +588,7 @@ rcxml_init(void)
 	rc.snap_edge_range = 1;
 	rc.snap_top_maximize = true;
 	rc.cycle_view_osd = true;
-	rc.cycle_preview_contents = false;
+	rc.cycle_preview_contents = true;
 	rc.cycle_preview_outlines = true;
 	rc.workspace_config.popuptime = INT_MIN;
 }
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -442,12 +442,29 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.snap_edge_range = atoi(content);
 	} else if (!strcasecmp(nodename, "topMaximize.snapping")) {
 		rc.snap_top_maximize = get_bool(content);
+
+	/* <windowSwitcher show="" preview="" outlines="" /> */
+	} else if (!strcasecmp(nodename, "show.windowSwitcher.core")) {
+		rc.cycle_view_osd = get_bool(content);
+	} else if (!strcasecmp(nodename, "preview.windowSwitcher.core")) {
+		rc.cycle_preview_contents = get_bool(content);
+	} else if (!strcasecmp(nodename, "outlines.windowSwitcher.core")) {
+		rc.cycle_preview_outlines = get_bool(content);
+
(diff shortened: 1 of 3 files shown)


Consolatis:
Other than ending the first line of the deprecation logs with `[dot][space]` this looks good to me.
I assume that is also what the CodeStyleCheck error is about, that thing is amazingly smart sometimes.

johanmalm:
^ But I want the space there!!

Consolatis:
I think that is what the message is about, the last 2 are missing the dot space. The first message uses it.

johanmalm:
> I think that is what the error is about, the last 2 are missing the dot-space. The first message uses it (and doesn't raise a codecheck error).

Yep. I'll add a space on the next line instead. `checkpatch.pl` is too clever :smile:

Consolatis:
> to be more consistent with the other options already there?

In regard to `rc.xml.all` I agree. I think the docs are fine though.
The parser implementation supports both variants so whatever the user decides to use is fine.

johanmalm:
I've stripped out `window` but left the rest as it is. Ready to merge from my perspective.

Consolatis:
You want to keep the more compact variant in `rc.xml.all` vs the long form with own tags instead of attributes shown above?

src/config/rcxml.c
```diff
@@ -442,12 +442,29 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.snap_edge_range = atoi(content);
 	} else if (!strcasecmp(nodename, "topMaximize.snapping")) {
 		rc.snap_top_maximize = get_bool(content);
+
+	/* <windowSwitcher show="" windowPreview="" windowOutlines="" /> */
+	} else if (!strcasecmp(nodename, "show.windowSwitcher.core")) {
+		rc.cycle_view_osd = get_bool(content);
+	} else if (!strcasecmp(nodename, "windowPreview.windowSwitcher.core")) {
+		rc.cycle_preview_contents = get_bool(content);
+	} else if (!strcasecmp(nodename, "windowOutlines.windowSwitcher.core")) {
```
Consolatis, line 451: These checks include the `window` prefix.

johanmalm: Whooa. Thanks for catching.

johanmalm:
> You prefer the more compact variant in `rc.xml.all` vs the long form with own tags instead of attributes shown above?

Yes, I think it's neater. Keeps it more compact.

The rest of `rc.xml.all` contains a mix anyway, so I think decision criteria ought to be (in this order):

1. Keep consistent with Openbox 3.6
2. If there is a reasonable chance of wanting to expand with further nested options in future, then use nested `<elements>`
3. Keep succinct - by using `attributes=""`

In this case (1) doesn't apply and I think (2) is pretty unlikely.

johanmalm:
Okay. Ready to go from my angle, but will hang on merging for a day or so in case anyone has any further thoughts as it's been quite a lengthy perspective already :smile:

Will also be a good check of the new website manpage workflow.

Consolatis:
PR looks good. I'll go for the merge.

> Will also be a good check of the new website manpage workflow.

Indeed.

PR #829, opened 2023-03-20
Implement action "Lower"

### LINKED ISSUE #578: Please implement <action name="Lower">
I'm currently exploring the possibility of switching from X11+OpenBox to labwc, and I cannot find a way to reproduce the following OpenBox behaviour:
```
    <keybind key="A-Down">
      <action name="Lower"/>
      <action name="FocusToBottom"/>
      <action name="Unfocus"/>
    </keybind>
```
I'm not sure about the focusing behaviour, but I'm fairly positive that the "Lower" action is not implemented in labwc.  Since it seems to my untrained eyes like a three line modification, I'd be grateful if it could be implemented.

jech:
Fixes #578.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -29,6 +29,9 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Raise">*
 	Restack the current window above other open windows.
 
+*<action name="Lower">*
+	Restack the current window below other open windows.
+
 *<action name="Iconify">*
 	Iconify (minimize) focused window.
 
--- a/src/action.c
+++ b/src/action.c
@@ -63,6 +63,7 @@ enum action_type {
 	ACTION_TYPE_ICONIFY,
 	ACTION_TYPE_MOVE,
 	ACTION_TYPE_RAISE,
+	ACTION_TYPE_LOWER,
 	ACTION_TYPE_RESIZE,
 	ACTION_TYPE_GO_TO_DESKTOP,
 	ACTION_TYPE_SEND_TO_DESKTOP,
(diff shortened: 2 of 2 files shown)


johanmalm:
Looks fine in itself, but I’m not clear if it overlaps with Unfocus and FocusToBottom.
Maybe there is a good case to combine these actions. Would appreciated some analysis if you’re able to.

http://openbox.org/wiki/Help:Actions#Unfocus

jech:
Sorry, I don't understand it either.

In any case, this PR is about action Lower, it has nothing to do with focus.

johanmalm:
You're right. I assumed `desktop_move_to_back()` did stuff to focus but it looks like it doesn't.

I've found an inconsistency though (not with this PR, but with existing code).

`desktop_move_to_back()` does not call `cursor_update_focus()`, nor `wlr_scene_node_lower_to_bottom()`, nor does it move xwayland 'sub-views' with it. It looks like we ought to model it on the `move_to_front` equivalent. I suspect the reason we have not seen this before is that `desktop_move_to_back()` is only called from `view_minimize()`
I can have a go at patching that first - but too tired tonight.

Consolatis:
> not seen this before is that `desktop_move_to_back()` is only called from `view_minimize()`

I am not sure if that is the correct thing to do though. I think minimizing should hide the views but not change their stacking order.

johanmalm:
We will get to this, but this patch has uncovered some dead(ish) code in `desktop_move_to_back()` so I want to sort that first. Don't worry about rebasing anymore. We'll sort that out on merge.

jech:
That's okay, I'm running with this patch, so I need to rebase whenever I update labwc.

Consolatis:
This is some pretty clean patch now :)
LGTM once #848 is merged and this is rebased.

johanmalm:
Thanks.
And appreciate the patience 😃

PR #838, opened 2023-03-20
Decorations: respect earlier decoration negotiation results

### LINKED ISSUE #297: Inconsistency between CSD/SSD reported to client, and actually rendered

### LINKED ISSUE #831: Element (Electron based) not showing decorations when CSD is enabled in rc.xml
Certain programs like Element, MPV and PPSSPP (all of them wayland native) don't show any decorations when CSD is enabled. (`<core><decoration>client</decoration></core>`)

I managed to "fix" it with a -really- nasty hack in `xdg-deco.c` ; it works but could bring other problems I believe. Here's the patch: https://github.com/umksan/labwc/commit/47f4b67f2c0b01d921e39e6ddd07d5b10b128fd9

Since Element uses Electron I have tried using `--enable-features=WaylandWindowDecorations` but decorations didn't appear either.

On the other hand, a retro gaming client Fightcade (which uses Electron but runs on Xwayland) is able to show SSD while CSD is enabled in rc.xml.

----

Thanks to @Consolatis for helping me to debug behaviour correctly. I tested multiple programs. All of them wayland native.

MPV and Element return this:

```
00:01:56.256 [../src/xdg-deco.c:31] Client requested server decorations
00:01:56.256 [../src/xdg-deco.c:39] We set decorations to server
```

PPSSPP:

```
00:00:01.932 [../src/xdg-deco.c:31] Client requested None decorations
00:00:01.932 [../src/xdg-deco.c:39] We set decorations to client
00:00:01.932 [../src/xdg-deco.c:31] Client requested server decorations
00:00:01.932 [../src/xdg-deco.c:39] We set decorations to server
```

Consolatis:
Before this patch, setting `<decoration>` to `client` would cause applications
which prefer server side decorations to not have any decorations at all. This
patch fixes it by respecting the result of earlier negotiations via the
xdg-decoration-unstable-v1 protocol.

This patch also slightly changes the behavior of the decoration negotiation
depending on the user configured decoration settings in rc.xml:
- If set to `server`: application preferences are ignored
- If set to `client`: application settings are honored. This makes sure clients
  that don't actually implement any kind of client side decorations (and thus
  request server side decorations) will not end up without any decorations.

Fixes #297
Fixes #831

---
This is a draft for now until it received some testing to prevent regressions.


Consolatis:
```
<Consolatis> hm... chromium plainly seems to ignore when we tell it to use server side decorations
<Consolatis> it asks for client side, gets told to use server side and yet still uses client side ones
<Consolatis> it can be switched via config to use "system titlebar and borders" though
<Consolatis> i guess we are back to the beginning of the question what should happen when a users sets rc.xml to "server": overwrite the preferences of clients or accept them
```

Consolatis:
Squashed the commits, this PR now only fixes the "no decorations at all when configured `client` decorations" case and moves `xdg-deco.c` to `src/decorations/`. This seems ready to go.

src/server.c
```diff
@@ -310,26 +312,8 @@ server_init(struct server *server)
 		&server->new_xdg_surface);
 
 	/* Disable CSD */
```
johanmalm, line 314: Remove comment 'Disable CSD'??

src/xdg.c
```diff
@@ -48,21 +48,19 @@ handle_new_xdg_popup(struct wl_listener *listener, void *data)
...
-
-	/*
-	 * Some XDG shells refuse to disable CSD in which case their
-	 * geometry.{x,y} seems to be greater than zero. We filter on that
-	 * on the assumption that this will remain true.
-	 */
-	struct wlr_xdg_surface_state *current =
-		&xdg_surface_from_view(view)->current;
-	if (current->geometry.x || current->geometry.y) {
+	/* view->ssd_preference is set by the kde or xdg decoration variant */
```
johanmalm, line 51: Spell out `xdg-decoration` or `kde-server-decoration`

src/xdg.c
```diff
@@ -526,6 +524,7 @@ xdg_surface_new(struct wl_listener *listener, void *data)
 	view->type = LAB_XDG_SHELL_VIEW;
 	view->impl = &xdg_toplevel_view_impl;
 	xdg_toplevel_view->xdg_surface = xdg_surface;
+	view->ssd_preference = LAB_SSD_PREF_CLIENT;
```
johanmalm, line 527: Should we explain why we set `ssd_preference` to client here? I get it following IRC discussion, but it's pretty non-intuitive.

src/server.c
Consolatis: removed

src/xdg.c
Consolatis: I've used the `.c` files instead but happy to change to the protocol names if desired.

src/xdg.c
Consolatis: Added some more comments

johanmalm:
LGTM.
…and looks a well crafted/encapsulated solution. Really nice.
Happy to merge now.
Or I can do some testing in the morning and merge then.

Consolatis:
> Or I can do some testing in the morning and merge then.

That would be great. It also provides for some more time to detect regressions.

johanmalm:
Have tested with `<decoration>` set to both client|server with a range of applications. Seems to work well. Will go for merge.

PR #844, opened 2023-03-23
cursor: fix focus.followMouse glitches

johanmalm:


src/cursor.c
```diff
@@ -508,6 +508,19 @@ cursor_update_focus(struct server *server)
...
 	struct cursor_context ctx = get_cursor_context(server);
+
+	if (ctx.view && rc.focus_follow_mouse) {
+		desktop_focus_and_activate_view(&server->seat, ctx.view);
+		if (rc.raise_on_focus) {
+			/*
+			 * Call view method directly as desktop_move_to_front()
+			 * contains a call to cursor_update_focus() and thus
+			 * loops inifinitely
+			 */
```
Consolatis, line 519: Heh, nice catch :)
As said, it was only compile tested.

Consolatis:
@jech another round of testing would be greatly appreciated.

Consolatis:
Alright then :)

jech:
It seems to work fine for switching desktops and closing the topmost window, thanks to both of you.

Testing together with https://github.com/labwc/labwc/pull/829, I still need action "Focus" after I do action "Lower".  This is consistent with Openbox, but not what I'd expect from FFM.  (I can live with either behaviour, it's just that I find the current behaviour counter-intuitive; so I guess a decision from a wise person is needed.)

Issue #846, opened 2023-03-25
No obvious way to tweak touchpad scrolling speed


jech:
Hi again,

The default scroll speed of the touchpad of my laptop is way faster under lawbc than under X.  I couldn't find any way to tweak it in the configuration file.

Please provide a way to tweak the scroll speed in the labwc configuration, or document how to do it outside of labwc.

On a related note, it's not obvious from the documentation whether it is possible to enable right edge scrolling.

johanmalm:
We don't currently support scroll-factor.
I think this patch adds the support - but would appreciate testing.
https://github.com/labwc/labwc/pull/847

Not sure how to achieve right edge scroll.

jech:
> Not sure how to achieve right edge scroll.

I believe you need to call `libinput_device_config_scroll_set_method`.  My understanding is that it takes a bitfield of scrolling methods (I, for one, want to have both two-finger scroll and right-edge scrolle enabled), so it's not immediately obvious to me how to map it to an XML.

jech:
The scroll_factor solves my immediate problem quite nicely, thanks.  (I cannot help wondering, though, whether the scroll factor should be the same on the vertical and horizontal axis.  I guess we'll worry about it if the issue arises.)

Consolatis:
From a quick look at the PR, it should use the same scroll factor for both values. Do you think it should use different ones, depending on the axis?

jech:
All I can say right now is that I've applied Johan's patch, set the scaling factor to 0.5, and the scrolling behavior feels similar to X.org's Synaptic driver.  My perhaps slightly obscure parenthetical remark was meant to say that I'm a little worried that linking the two is not the right behaviour, but that I suggest we worry about it later.

As far as I'm concerned, feel free to merge.

johanmalm:
Interestingly libinput 1.23 has a new pointer acceleration profile: the "custom" profile.

https://mastodon.social/@jose_exposito@fosstodon.org/110085325061609318
https://gitlab.freedesktop.org/libinput/libinput/-/releases/1.23.0
http://who-t.blogspot.com/2023/01/libinput-and-custom-pointer.html

This might negate the requirement for a dedicated `rc.xml` `scroll_factor`

johanmalm:
[...]

PR #847, opened 2023-03-25
config: support scroll-factor

johanmalm:
Fixes: issue #846


docs/rc.xml.all
jech, line 1: Typo, should say `</scrollFactor>`

Consolatis:
If the review comment from @jech is addressed: LGTM.

docs/rc.xml.all
johanmalm: Thanks. Done.

johanmalm:
Awaiting evaluation of libinput 1.23 pointer acceleration "custom" profile.

johanmalm:
That “custom” acceleration profile is quite different. Will go for merge.

PR #848, opened 2023-03-25
desktop: move scene-tree node in move-to-back

johanmalm:
@Consolatis @jech How about this as a pre-commit for PR #829 

`view_minimize()` does not need to call `desktop_move_to_back()` because the stacking order is not changed and the windowSwitcher uses the scene-tree nodes anyway.

I thought of not keeping `server->views` in sync with view z-order, but xwayland sub-views still relies on it.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -497,8 +497,13 @@ xwayland_view_maximize(struct view *view, bool maximized)
 		maximized);
 }
 
+enum z_direction {
+	LAB_TO_FRONT,
+	LAB_TO_BACK,
+};
+
 static void
-move_sub_views_to_front(struct view *parent)
+move_sub_views(struct view *parent, enum z_direction z_direction)
 {
 	assert(parent);
 
@@ -524,7 +529,11 @@ move_sub_views_to_front(struct view *parent)
 		if (top_parent_of(view) != parent_xwayland_surface) {
 			continue;
 		}
-		view_impl_move_to_front(view);
+		if (z_direction == LAB_TO_FRONT) {
+			view_impl_move_to_front(view);
+		} else if (z_direction == LAB_TO_BACK) {
+			view_impl_move_to_back(view);
+		}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


jech:
I've rebased #829 over this, and I'm running it right now.  It seems to do the right thing in the last five minutes.

There's one change in behaviour, which is that now *action Lower* has the side effect of switching keyboard focus to the newly toplevel window.  This is clearly the right behaviour in FFM mode, but it might not be the expected behaviour in click-to-focus mode.  (I wouldn't know, I only use click-to-focus under duress.)

src/xwayland.c
```diff
@@ -497,8 +497,13 @@ xwayland_view_maximize(struct view *view, bool maximized)
 		maximized);
 }
 
+enum direction {
+	LAB_TO_FRONT,
+	LAB_TO_BACK,
+};
```
Consolatis, line 503: Even though this is an xwayland.c internal enum I'd propose something more specific so we don't end up clashing with another `direction` enum in the future, bad examples: `z_order` or `stack_direction`.

Consolatis:
Other than the enum naming nitpick this looks good to me. Didn't test but I don't expect it to change behavior in the master branch.

Consolatis:
> I've rebased #829 over this, and I'm running it right now. It seems to do the right thing in the last five minutes.
> 
> There's one change in behaviour, which is that now _action Lower_ has the side effect of switching keyboard focus to the window that's now under the moust. This is clearly the right behaviour in FFM mode, but it's different from what OpenBox does, so you may want to consider documenting the discrepancy.

I'd argue this feels more like a bug in OpenBox though if the idea of followMouse is to always give focus to the surface under the cursor.

jech:
Agreed.  (There's a typo in `rc.xml.all` file, by the way, I've marked it in the PR.)

src/xwayland.c
johanmalm: Good point. Done.

Issue #849, opened 2023-03-26
focus.followMouse regression when manually switching keyboard focus via A-Tab


Consolatis:
#844 changed the focus.followMouse behavior by integrating it into `cursor_update_focus()`.
This means that its now impossible to switch the keyboard focus away from the window the cursor is currently on.

I am not sure how we can properly solve this.
We could move the follow-mouse logic from `cursor_update_focus()` to the underlying callers:
- when switching workspaces
- when closing a window
- when opening a window
- when moving a window (tiling, maximizing)
- when minimizing, un-minimizing
- ...

But that seems unmaintainable and is 98% of the same use-cases as for `cursor_update_focus()`. Other ideas?

Edit:
Maybe we could simply check for `&& !server->osd_state.cycle_view` in the followMouse case of `cursor_update_focus()`.

Edit 2:
Just tested, seems to work fine if combined with delaying the reset of `cycle_view` until after calling `cursor_update_focus()` in `osd_finish()`. Will put together a PR.

PR #850, opened 2023-03-26
followMouse: allow focus switching via A-Tab

Consolatis:
Fixes a regression caused by the recent enhancement of followMouse behavior.

Fixes: #849
Reported-by: mitigate via IRC


johanmalm:
Thanks! LGTM.

PR #851, opened 2023-03-26
Workspace follow

johanmalm:


johanmalm:
Have tested. Seems to work fine.
Need to amend documentation.
Would appreciate a quick review.

src/config/rcxml.c
```diff
@@ -119,6 +120,11 @@ fill_keybind(char *nodename, char *content)
 		wlr_log(WLR_ERROR, "expect <action name=\"\"> element first. "
 			"nodename: '%s' content: '%s'", nodename, content);
 	} else {
+		/*
+		 * Here we deal with action sub-elements such as <to>, <output>,
+		 * <region>, <direction> and so on. This is is common to key-
```
Consolatis, line 125: duplicate "is"

src/menu/menu.c
```diff
@@ -664,15 +664,15 @@ init_windowmenu(struct server *server)
 		/* Workspace sub-menu */
 		struct menu *workspace_menu = menu_create(server, "workspaces", "");
 		current_item = item_create(workspace_menu, _("Move left"), false);
+		/*
+		 * <action name="SendToDesktop"><follow> is true by default so
+		 * GoToDesktop will be called part of this action.
```
Consolatis, line 669: "as part of the action"

Consolatis:
As requested: just a quick review: new bool argument logic looks fine. Same for the handling of `SendToDesktop`.
I am not sure about using `follow="yes"` as default, if openbox defaults to that as well I guess its fine.

We could add something like `struct action_arg *get_arg_by_name(const char *name)` but the loop as currently implemented looks good to me as well.

src/config/rcxml.c
johanmalm: Done

johanmalm:
> As requested: just a quick review: new bool argument logic looks fine. Same for the handling of `SendToDesktop`.

Thanks.

> I am not sure about using `follow="yes"` as default, if openbox defaults to that as well I guess its fine.

It feels 50/50, but I'm good with taking the Openbox line if you are.

> We could add something like `struct action_arg *get_arg_by_name(const char *name)` but the loop as currently implemented looks good to me as well.

Shall we ponder that further and consider it as a separate patch? Maybe:

```
bool action_get_bool_from_key(struct wl_list *action_args, const char *key);
char *action_get_str_from_key(struct wl_list *action_args, const char *key);
```

johanmalm:
Ready for merge from my perspective

Consolatis:
> Shall we ponder that further and consider it as a separate patch? Maybe:
> 
> ```
> bool action_get_bool_from_key(struct wl_list *action_args, const char *key);
> char *action_get_str_from_key(struct wl_list *action_args, const char *key);
> ```

The issue I see with especially the bool variant is that we can't figure out if its `false` because the value is `false` or because the there is no argument with that name. For the str variant we can just return `NULL` but for the bool one I don't see a way to allow for distinction. We could have them return some struct like this instead
```c
struct action_arg_bool_value {
	bool found;
	bool value;
}
struct action_arg_str_value {
	bool found;
	const char *value;
}
```

But then we'd basically have the same as when requesting `struct action_arg` itself.

Or we do it the python way and supply a default value as part of the arguments, that way the caller doesn't have to care if the argument was found or not.

Consolatis:
Nice PR :)

johanmalm:
Have updated scope document
https://github.com/labwc/labwc-scope/commit/3d4f2a227045e8efadff7808fd0437e857b01e05

johanmalm:
> Or we do it the python way and supply a default value as part of the arguments, that way the caller doesn't have to care if the argument was found or not.

I like the python way. It's succinct.

I guess the other 'C' ways are

(a) to return an int and set it to `-1` on error. But that requires checking `if (foo < 0)` before using it and then we'd have to `bool` cast or just using `int` variables instead.

(b) to pass an `int err = 0` variable... `get_bool(foo, bar, &err)`, but it's just as much hassle :smile:

PR #853, opened 2023-03-26
Action bool args

johanmalm:
@Consolatis 
Just playing with ways to clean up the process of getting str/bool from actions by key.
Appreciate thoughts.


src/action.c
```diff
@@ -460,19 +505,9 @@ actions_run(struct view *activator, struct server *server,
...
-
-				wl_list_for_each(arg, &action->args, link) {
-					if (!strcmp("to", arg->key)) {
-						target_name = action_str_from_arg(arg);
-					} else if (!strcmp("follow", arg->key)) {
-						follow = action_bool_from_arg(arg);
-					}
-				}
-				target = workspaces_find(view->workspace, target_name);
+				char *to = str_from_key(action, "to", "");
```
Consolatis, line 508: Instead of `""` I guess we could also use `NULL` here. I am not sure how `workspaces_find()` handles it though.

src/action.c
```diff
@@ -144,11 +169,31 @@ action_str_from_arg(struct action_arg *arg)
 	return ((struct action_arg_str *)arg)->value;
 }
 
+static char *
+str_from_key(struct action *action, const char *key, char *default_value)
```
Consolatis, line 173: `const char *` for both the argument and return type?

src/action.c
```diff
@@ -144,11 +169,31 @@ action_str_from_arg(struct action_arg *arg)
...
+{
+	assert(key);
+	struct action_arg *arg;
+	wl_list_for_each(arg, &action->args, link) {
+		if (!strcmp(key, arg->key)) {
+			return action_str_from_arg(arg);
+		}
```
Consolatis, line 180: `strcasecmp()` to handle silly dev mistakes in the future? By adding the default value those mistakes might be hidden during testing so I feel like leaving a bit more space for error correction in this case could be appropriate.

As alternative we could also define that all internal action arg keys have to be lowercase, that should likely be mentioned in the code style and function docs for `action_arg_add_xxx()` / `xxx_from_key()`.

src/action.c
```diff
@@ -144,11 +169,31 @@ action_str_from_arg(struct action_arg *arg)
...
+	}
+	return default_value;
+}
+
 static bool
-action_bool_from_arg(struct action_arg *arg)
+bool_from_key(struct action *action, const char *key, bool default_value)
```
Consolatis, line 186: I am not sure about the naming here, I kind of miss the `arg` part in `bool_from_key` / `str_from_key`.
Suggestion: `get_arg_value_bool()` / `get_arg_value_str()`. Open to different naming though.

src/action.c
johanmalm: Done. workspaces_find() is good with it.

src/action.c
johanmalm: Have added const.

src/action.c
johanmalm: Good point. Have gone for strcasecmp

src/action.c
johanmalm: Renamed to get_arg_value_{bool,str}.
Agree, it seems better

johanmalm:
Have now got three variants of get_bool()!

include/common/get-bool.h
```diff
@@ -4,12 +4,23 @@
...
 /**
- * get_bool - interpret string and return boolean
- * @s: string to interpret
+ * get_bool - return boolean value of string
+ * @string: string to interpret
+ * @ret: boolean variable to set
  *
- * Note: This merely performs a case-insensitive check for 'yes' and 'true'.
- * Returns false by default.
+ * Note: @var will not be touched if the string is not one of 'yes', 'no',
```
Consolatis, line 11: s/@var/@ret/

Suggest renaming from
- `get_bool()` to `parse_bool()`
- `get_bool_fallback()` to `parse_bool_with_default()`
- `get_bool2()` to `parse_bool_as_int()`.

The last one isn't immediately clear to me though, it could also mean the input is a bool and we cast it to an int. Can't come up with a better name right now.

I am also not sure about the default/fallback variant returning the result vs the other ones supplying a pointer to the result. I have a feeling that all 3 should use the same variant. The docs for the first two also mentions returning the bool.

johanmalm: Thank you. Very helpful feedback. 
We now just have one `parse_bool()` to cover all. It now seems a whole lot simpler.
I must have been a bit tired when I came up with all those versions :smile:

johanmalm:
Ready from my perspective. Would a appreciate a scan-over the last commit if anyone has time.

johanmalm:
Have added a fix for #894 and rebased.

johanmalm:
Okay. *Now* it's ready :smile:

Consolatis:
> Ready from my perspective. Would a appreciate a scan-over the last commit if anyone has time.

Uh, completely missed this comment. Sorry.
LGTM, didn't test.

The `if (ret < 0)` branching is a bit repetitive and could potentially be reworked later on but for now I think its fine.
Having a single function to set/return: an error code, the value, a default is a bit tricky. Can't currently come up with a nice function signature either that supports all 3 and doesn't require branching by the caller.

Consolatis:
For future reference, what about something like this? (completely untested pseudo code)
```c

/*
 * Always dereferences @result and sets either the parsed value or the @default_value.
 *
 * Returns success of parsing: false of we couldn't parse @str, true otherwise.
 */
bool
parse_bool(const char *str, bool *result, bool default_value)
{
	if (!str) {
		goto error_not_a_boolean;
	} else if (!strcasecmp(str, "yes") || !strcasecmp(str, "true")) {
		*result = true;
		return true;
	} else if (!strcasecmp(str, "no") || !strcasecmp(str, "false")) {
		*result = false;
		return true;
	}
error_not_a_boolean:
	wlr_log(WLR_ERROR, "(%s) is not a boolean value", str);
	*result = default_value;
	return false;
}

/* Which could then in most cases just be called like this, no if() required */
parse_bool(content, &rc.adaptive_sync, rc.adaptive_sync);

/* Or, if we can't assign the value directly */
bool ret = false;

if (parse_bool(content, &ret, ret)) {
	current_libinput_category->tap = ret ? LIBINPUT_CONFIG_TAP_ENABLED
		: LIBINPUT_CONFIG_TAP_DISABLED;
}
```

johanmalm:
Proposal in #896

It keeps `parse_bool()` clean/simple, and just adds a couple of helpers to avoid the repeated `if (ret < 0)...`

Prefer not to have `bool ret = false;` in `rcxml.c`'s `entry()`. It's tidier to set all default values in the same place.

PR #854, opened 2023-03-27
src/cursor.c: fix invisible cursor on output loss / restore

### LINKED ISSUE #820: Cursor is invisible when hovering over window after turning monitor off and on again
In some applications just moving it is enough to make it reappear (Firefox), for others I have to move the cursor over the window titlebar / border (foot).

Consolatis:
Previously, the cursor image was only updated on output loss when the cursor was on a labwc owned surface. This patch forces a re-enter of a client surface in the remaining case of cursor being over a non-labwc surface which causes the client to re-set its own cursor image.

This fixes a regression caused by 4dc99e2f3856650f3b61778f015a823cfae978d2.
Thanks to @Flrian for finding the root cause of the issue.

Fixes #820

Reported-by: @Flrian
Tested-by: @Flrian

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -260,6 +260,16 @@ cursor_update_image(struct seat *seat)
 {
 	enum lab_cursors cursor = seat->server_cursor;
 	if (cursor == LAB_CURSOR_CLIENT) {
+		/*
+		 * When we loose the output cursor while over a client
+		 * surface (e.g. output was destroyed and we now deal with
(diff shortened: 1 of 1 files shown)


Consolatis:
Latest commits confirmed (via IRC) to fix the cursor being invisible after startup case.
I've squashed the commits, this is ready to go from my point of view.

johanmalm:
Thanks!

PR #861, opened 2023-03-30
Support ext-session-lock protocol

johanmalm:
…because the next version of swaylock will not run without it.

### DIFF
--- a/src/session-lock.c
+++ b/src/session-lock.c
@@ -1,0 +1,314 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include "common/mem.h"
+#include "labwc.h"
+
+static struct wl_listener new_lock;
+static struct wl_listener manager_destroy;
+static struct wlr_session_lock_manager_v1 *wlr_session_lock_manager;
+
+static struct server *g_server;
+
+struct session_lock_output {
+	struct wlr_scene_tree *tree;
+	struct wlr_scene_rect *background;
+	struct session_lock *lock;
+	struct output *output;
+	struct wlr_session_lock_surface_v1 *surface;
+
+	struct wl_list link; /* session_lock.outputs */
+
+	struct wl_listener destroy;
+	struct wl_listener commit;
+	struct wl_listener surface_destroy;
+	struct wl_listener surface_map;
+};
+
+static void
+focus_surface(struct session_lock *lock, struct wlr_surface *focused)
+{
+	lock->focused = focused;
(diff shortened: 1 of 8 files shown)


src/session-lock.c
```diff
@@ -0,0 +1,309 @@
...
+	struct wlr_session_lock_surface_v1 *lock_surface = data;
+	struct output *output = lock_surface->output->data;
+	struct session_lock_output *lock_output;
+	wl_list_for_each(lock_output, &lock->session_lock_outputs, link) {
+		if (lock_output->output == output) {
+			goto found_lock_output;
+		}
+	}
+	wlr_log(WLR_ERROR, "fatal: new lock surface, but no output");
+	exit(EXIT_FAILURE);
```
Consolatis, line 102: Just had a quick look through the diff, this is the only place I saw a potential issue.

On some systems, turning off the monitors will cause them to actually be disconnected e.g. we don't have any outputs left. Could this trigger a complete exit of labwc when being locked and then turning off the monitors?

johanmalm: Good point. Have changed to just returning.

johanmalm:
Background info: https://github.com/swaywm/swaylock/commit/ac3b49b6571ceda3f8db11a98bfe320106996280

src/session-lock.c
Consolatis: Hm.. now I wonder if that could cause a lockscreen to not render a surface so there is no way to unlock the session once the output is back though. I guess some tests regarding that would be useful.

johanmalm: It works fine.

The surface is created when the output comes back (in `session_lock_output_create()`).

I tested with the following:

```bash
#!/bin/sh

printf '%b\n' "starting..."
sleep 1

printf '%b\n' "eDP-1 off"
wlr-randr --output eDP-1 --off

sleep 2

printf '%b\n' "swaylock before"
./build/swaylock -c 222222 -d >$HOME/log.txt 2>&1 & 
printf '%b\n' "swaylock after"

sleep 2

printf '%b\n' "eDP-1 on"
wlr-randr --output eDP-1 --on

sleep 1
printf '%b\n' "complete"
```

Consolatis: Thanks for confirming.

Consolatis:
LGTM, didn't test.

johanmalm:
Thanks for review.

Have just discovered a problem on multi-output setup. Need to investigate, so have put back to draft.

johanmalm:
Helps if we set the node position :smile:

Ready now from my perspective. Will go for merge shortly.

Issue #856, opened 2023-03-29
AlwaysOnTop windows don't appear in the OSD


ghost:
Easy to replicate, toggle any window AlwaysOnTop via the graphical client-menu and then Alt+Tab.
It happens with Wayland native programs and Xwayland as well. I only tested this on the lastest git master.
I believe this is a bug because the user cannot gain control back on the "OnTop" window.

Consolatis:
Note about implementing a fix:
We might need to return to looping over `server->views` with a filter for node->parent being either the current workspace tree or the always-on-top tree. That should fix it.

Consolatis:
> > I actually intend to revert to using server->views + filter for current workspace (at least for `A-Tab`) because otherwise always-on-top views (and other non direct children of the workspace nodes) will not show up there (#856).
> 
> I had been thinking similar thoughts :smile:
> +possibly showing all workspaces in windowSwitcher? I know we said originally that we wouldn't but...
> Preview would be hard I guess - but possibly not needed for other workspaces.

_Originally posted by @johanmalm in https://github.com/labwc/labwc/issues/866#issuecomment-1493115601_

Good point regarding preview. Guess there is still some stuff to work out if we ever to decide to add that config option.

johanmalm:
COMPLETE
https://github.com/labwc/labwc/commit/c71fccbcec7749635bac1967942fe18d1873f8f8

PR #866, opened 2023-04-01
view: Add view_move_to_front/back() with _no_focus variant

jlindgren90:
This avoids calling `view->impl` functions from `cursor.c` and `desktop.c`.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -184,6 +184,10 @@ void view_snap_to_edge(struct view *view, const char *direction,
 	bool store_natural_geometry);
 void view_snap_to_region(struct view *view, struct region *region,
 	bool store_natural_geometry);
+
+void view_move_to_front(struct view *view);
+void view_move_to_back(struct view *view);
+
 const char *view_get_string_prop(struct view *view, const char *prop);
 void view_update_title(struct view *view);
 void view_update_app_id(struct view *view);
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -40,7 +40,9 @@ static void
 end_cycling(struct server *server)
 {
 	desktop_focus_and_activate_view(&server->seat, server->osd_state.cycle_view);
-	desktop_move_to_front(server->osd_state.cycle_view);
+	if (server->osd_state.cycle_view) {
+		view_move_to_front(server->osd_state.cycle_view);
+	}
 
 	/* osd_finish() additionally resets cycle_view to NULL */
 	osd_finish(server);
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -367,8 +367,6 @@ void foreign_toplevel_update_outputs(struct view *view);
  *              cannot assume this means that the window actually has keyboard
  *              or pointer focus, in this compositor are they called together.
  */
-void desktop_move_to_front(struct view *view);
-void desktop_move_to_back(struct view *view);
 void desktop_focus_and_activate_view(struct seat *seat, struct view *view);
 void desktop_arrange_all_views(struct server *server);
 void desktop_focus_output(struct output *output);
--- a/src/action.c
+++ b/src/action.c
@@ -439,12 +439,12 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_RAISE:
 			if (view) {
-				desktop_move_to_front(view);
+				view_move_to_front(view);
 			}
 			break;
 		case ACTION_TYPE_LOWER:
 			if (view) {
-				desktop_move_to_back(view);
+				view_move_to_back(view);
 			}
 			break;
 		case ACTION_TYPE_RESIZE:
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -481,7 +481,7 @@ process_cursor_motion(struct server *server, uint32_t time)
 	if (ctx.view && rc.focus_follow_mouse) {
 		desktop_focus_and_activate_view(seat, ctx.view);
 		if (rc.raise_on_focus) {
-			desktop_move_to_front(ctx.view);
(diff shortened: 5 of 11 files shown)


Consolatis:
I am not sure about moving `_to_front/back` to `view.c`.
Usually we do all actions relating to interactions of different views in `desktop.c` and actions relating to views themselves in `view.c`.

An alternative solution to the PR could be to return early if the view is already the top view, so we wouldn't call `cursor_update_focus()` if the few is already at the top.

jlindgren90:
> Usually we do all actions relating to interactions of different views in desktop.c and actions relating to views themselves in view.c.

Yes, this is an awkward case though since the actual multi-view interactions are done in the `view->impl` functions, not in desktop.c. So the conceptual separation is already broken in that sense.

johanmalm:
I'm good with moving these to `view.c` and on reflection, do think it's actually preferable.

If we wanted to be really hard on the desktop argument, we should have left all the xwayland-move-sub-view stuff in desktop.c --- but I think it is better the way we've gone (even if it's adds a bit of grey to the desktop/view division).

The way John has proposed keeps the calling of `view->impl->*` more consistent.

> An alternative solution to the PR could be to return early if the view is already the top view, so we wouldn't call 

Another problem with this is that we want to be workspace specific (i.e. scene_tree orientated rather than iterating over `server->views`).

On that note, I don't think there are any places left in the code that cares about the order the `server->views` other than xwayland's `move_sub_views()`.
So we could nearly remove the 

```
wl_list_remove(&view->link);
wl_list_insert(&view->server->views, &view->link);
```

...which helps it feel even less 'desktopy'

Whilst on the topic - do we actually need the `move_sub_views()` thing? I tried with openbox/mutter and they don't seems to let focus change to other toplevel with popups/sub-windows open at all. I found this when experimenting the `close_all_popups` PR. We could do with finding an xwayland application where the whole `move_sub_views()` actually makes a difference. I must have put it in there for a reason, but I can't remember what it is.

Consolatis:
I think the `move_sub_views()` thingie was about `A-Tab` so it would focus dialog windows and similar that are on top of the main window.

> On that note, I don't think there are any places left in the code that cares about the order the `server->views` other than xwayland's `move_sub_views()`. So we could nearly remove the
> 
> ```
> wl_list_remove(&view->link);
> wl_list_insert(&view->server->views, &view->link);
> ```

I actually intend to revert to using server->view (at least for `A-Tab`) because otherwise always-on-top views (and other non direct children of the workspace nodes) will not show up there (#856).

src/keyboard.c
```diff
@@ -40,7 +40,9 @@ static void
 end_cycling(struct server *server)
 {
 	desktop_focus_and_activate_view(&server->seat, server->osd_state.cycle_view);
-	desktop_move_to_front(server->osd_state.cycle_view);
+	if (server->osd_state.cycle_view) {
+		view_move_to_front(server->osd_state.cycle_view);
```
Consolatis, line 44: This could use the _no_focus variant as `osd_finish()` also calls `cursor_update_focus()`.

jlindgren90:
> Whilst on the topic - do we actually need the move_sub_views() thing? I tried with openbox/mutter and they don't seems to let focus change to other toplevel with popups/sub-windows open at all.

`move_sub_views()` is needed with any X11 application that uses floating toolbars or the like (a UI pattern that incidentally does not seem to really be allowed for native Wayland applications). Try Audacious in Winamp mode - the playlist & equalizer windows should rise to the top along with the main window.

include/view.h
```diff
@@ -184,6 +184,12 @@ void view_snap_to_edge(struct view *view, const char *direction,
 	bool store_natural_geometry);
 void view_snap_to_region(struct view *view, struct region *region,
 	bool store_natural_geometry);
+
+void view_move_to_front(struct view *view);
+/* This variant does not call cursor_update_focus() */
+void view_move_to_front_no_focus(struct view *view);
```
Consolatis, line 190: Thoughts about using `view_move_to_front(struct view *view, bool focus)` vs. the two function variant?

jlindgren90: My thought was that `view_move_to_front()` was the normal case and that the only purpose of the `_no_focus` variant was to prevent recursion into `cursor_update_focus()`. So I didn't want to add a second parameter to all of the "normal" call sites.

Another possibility is to put a recursion guard in `cursor_update_focus()`; then `view_move_to_front()` doesn't need to care.

Consolatis: Personally I'd slightly prefer the bool variant but I understand the argument for having two functions instead.
I am fine with both variants.

> Another possibility is to put a recursion guard in `cursor_update_focus()`; then `view_move_to_front()` doesn't need to care.

I personally would prefer that one to the current solution.
It would also reduce the chances of introducing something similar in the future.

jlindgren90:
v2: Add an explicit recursion guard in `cursor_update_focus()`.

src/keyboard.c
jlindgren90: `_no_focus` variant removed now, so marking this resolved.

Consolatis:
Sweet, just looked through the diff again to verify that `view` is always `!= NULL` when calling due to the new `assert(view)`.
This PR looks good to me.

johanmalm:
LGTM. Ready for merge?

johanmalm:
> I actually intend to revert to using server->views + filter for current workspace (at least for `A-Tab`) because otherwise always-on-top views (and other non direct children of the workspace nodes) will not show up there (#856).

I had been thinking similar thoughts :smile:
+possibly showing all workspaces in windowSwitcher? I know we said originally that we wouldn't but...
Preview would be hard I guess - but possibly not needed for other workspaces.

Issue #860, opened 2023-03-30
Add policykit and software rendering fallback to README.md


Consolatis:
We should likely add the policykit dependency (and the WLR_RENDERER=pixman fallback) somewhere to the readme.
I think it is usually available on systems that have at least one desktop environment running but for new systems without any desktop environment that would reduce the time required to get labwc running and prevent some troubleshooting steps.

See #859 for reference.

Consolatis:
That is good to know.
So maybe we can add something generic like "When experiencing permission errors during labwc startup, installing policykit might help".

FuzzyQuills:
> We should likely add the policykit dependency (and the `WLR_RENDERER=pixman` software rendering fallback) somewhere to the readme. I think it is usually available on systems that have at least one desktop environment running but for new systems without any desktop environment it would reduce the time required to get labwc running and prevent some troubleshooting steps.
> 
> See #859 for reference.

Would it be worth also listing GUI polkit software packages someone could use in autostart to get a GUI popup? I recently had to figure this out and my solution was installing lxqt's polkit agent and autostarting it in the background. (It was otherwise trying to use systemctl's prompt and wasn't spawning a terminal window to make it work)

johanmalm:
> Would it be worth also listing GUI polkit software packages someone could use in autostart to get a GUI popup? I recently had to figure this out and my solution was installing lxqt's polkit agent and autostarting it in the background. (It was otherwise trying to use systemctl's prompt and wasn't spawning a terminal window to make it work)

Yes. I'll try to get around to this soon. 

I don't know anything about GUI polkit software. If you're happy to provide some words/sentences or a few pointers, that would be good.

ludg1e:
Maybe https://wiki.archlinux.org/title/Polkit#Authentication_agents is useful

PR #867, opened 2023-04-01
server: add wlr_backend_autocreate() trouble-shooting message

johanmalm:


johanmalm:
Something like this in support of #860 

If we agree the message here, I can add the same to the website getting-started page.

src/server.c
```diff
@@ -183,6 +183,27 @@ server_global_filter(const struct wl_client *client, const struct wl_global *glo
...
+ * exist, in the hope that it will reduce the time required to get labwc running
+ * and prevent some troubleshooting steps.
+ */
+static const char helpful_seat_error_message[] =
+"\n"
+"Some friendly trouble-shooting help\n"
+"===================================\n"
+"\n"
+"If a seat could not be created, this may be caused by lack of permission to the\n"
+"seat and video groups. If you are using a systemd setup, try installing polkit\n"
```
Consolatis, line 199: `seat, input and video groups`. (Strictly speaking potentially also the `tty` group but mentioning just the 3 seems enough).

Consolatis:
Other than the small nitpick above this looks good to me.
Nice idea to add that message!

johanmalm:
Fixed.
Okay. I'll go for the merge then.
I uninstalled `polkit` to try it out for myself... took longer than I thought it would as so much depended on it :smile:

Issue #830, opened 2023-03-20
Focus follows mouse is incomplete


jech:
When using focus follows mouse, the focus does not always follow the mouse.  Specific cases that I've noticed:

   * if the focused window is closed, the window under the cursor is not selected until I move the mouse;
   * after switching desktops using action `GoToDesktop`, the window under the cursor does not get selected.
 
There are certainly others, these are just the cases that bother me in everyday usage.

Consolatis:
Hm.. right. Thanks for bringing this up.
We are only updating the mouse focus in those cases but do not update the keyboard focus.

I wonder if we could get away with just doing the `desktop_focus_and_activate_view()` / `desktop_move_to_front()` dance in `cursor_update_focus()` if followMouse / raiseOnFocus is set by the user.

jech:
Thanks, that seems to work fine.  It brings labwc to a level at which I can attempt to do some serious work under it :-)

Consolatis:
Should be fixed by #844, feel free to comment when that is not the case.

PR #869, opened 2023-04-02
followMouse: add followMouseRequiresMovement

### LINKED ISSUE #862: With `focus.followMouse` enabled, new windows no longer receive focus
Bisection indicates that this regressed in f5c4651c67236191caa30b09c5e5f3774fa98864. To reproduce, simply enable `focus.followMouse`, open a window, maximize it, and open a second window. The second, new window does not receive focus.

Openbox has a `focus.focusNew` [setting](http://openbox.org/wiki/Help:Configuration#Focus), which is independent of `focus.followMouse`. labwc doesn't implement this setting yet, but I don't think the focus-new behavior should depend on `focus.followMouse` the way they now do. In general it is surprising that all of a sudden new windows no longer receive focus.

I believe the changes in f5c4651c67236191caa30b09c5e5f3774fa98864 may have made `focus.followMouse` behave more like `focus.underMouse`, another Openbox setting that labwc doesn't yet support.

Consolatis:
This implements the same config option as `underMouse` in Openbox.

Fixes #862

---
The implementation uses a different name for the config than Openbox (`followMouseRequiresMovement` vs `underMouse` which I hope is easier to understand and doesn't require looking it up in the docs). The default of the new config is set to `yes` to restore the previous behavior of `followMouse`.

If set to `no`, this PR currently prevents focusing new windows even if the cursor is at the place the new window spawns, I am not quite sure why that is. It also wasn't really tested in depth so draft for now. Practical testing (with reports including the tested variant of `yes`/`no`) welcome.

CC @bi4k8 @jech

Ref:
- #830
- #862

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -116,6 +116,14 @@ The rest of this man page describes configuration options.
 	Make focus follow mouse, i.e. focus is given to window under mouse
 	cursor. Default is no.
 
+*<focus><followMouseRequiresMovement>* [yes|no]
+	Requires cursor movement if followMouse is enabled. It is the same
+	as the "underMouse" setting in Openbox. If set to "no", labwc will
+	additionally focus the window under the cursor in all situations
+	which change the position of a window (e.g. switching workspaces,
+	opening/closing windows). Focusing a different window via A-Tab is
+	still possible, even with this setting set to "no". Default is yes.
+
 *<focus><raiseOnFocus>* [yes|no]
 	Raise window to top when focused. Default is no.
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -53,6 +53,7 @@
 
   <focus>
     <followMouse>no</followMouse>
+    <followMouseRequiresMovement>yes</followMouseRequiresMovement>
     <raiseOnFocus>no</raiseOnFocus>
   </focus>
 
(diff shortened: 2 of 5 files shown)


jech:
I've always considered this a bug in OpenBox.

The way FFM used to behave (with twm and friends) is that the window under the mouse always has focus, no matter what.  This implies that a new window only has focus if it happens to be opened under the mouse, and that Alt-tab switching only raises the selected window, and does not actually focus it unless the mouse cursor happens to be above it.

This is a simple, consistent model.  I realise that some people want different behaviour; that's okay, but I think it should be possible to achieve the behaviour described above with minimal configuration, and the docs should document clearly how to achieve it.

johanmalm:
Thanks. Looks good.

> The implementation uses a different name for the config than Openbox (`followMouseRequiresMovement` vs `underMouse` 

Happy with deviation. Will record in scope-document when merged.

> The default of the new config is set to `yes` to restore the previous behavior of `followMouse`.

Good plan.

> If set to `no`, this PR currently prevents focusing new windows even if the cursor is at the place the new window spawns, I am not quite sure why that is.

I can replicate this, but can't figure out why either (although have only looked for 10 min).

Consolatis:
> If set to `no`, this PR currently prevents focusing new windows even if the cursor is at the place the new window spawns, I am not quite sure why that is.

I think I've figured it out, it happens when `raiseOnFocus` is set as well and the cursor is on top of another surface while spawning a new one that would then be under the cursor. But we will actually raise the old surface on top of the new one and thus the focus never changes. It does not happen when the cursor is over the empty desktop when spawning a new surface.

Issue #872, opened 2023-04-12
Pointer warp doesn't seem to be implemented(?), breaking certain games.


FuzzyQuills:
This bug shows up in the most obvious way in Overwatch 2; normally when opening the communication wheel in-game by holding down the C key, the cursor gets recentered to the middle of the wheel by the game. 

In Labwc however, this doesn't occur and as a result, the cursor is still sitting where it was last moved. (Let's say I said "thank you" before closing the wheel, next time it's opened, it's still sitting on "thank you" instead of being in the middle like it's supposed to be)

Workaround: Play game through gamescope or use a different WM.

Given Sway also has this issue, could this be a wlroots bug?
KDE doesn't have this issue.

FuzzyQuills:
Just made an attempt to fix this issue myself; feedback would be appreciated!

Consolatis:
Thanks a lot. I am a bit short on time currently but will review the PR later today.

PR #873, opened 2023-04-12
Mouse constraints improvements.

FuzzyQuills:
After quite a few awkward and frustrating moments where I couldn't properly use the communication wheel in Overwatch 2 (#872 ) I decided to try implementing pointer warp myself by looking at how Gamescope does it.

This is the result, the game is now able to re-center it's pointer properly. Works for both windowed and fullscreen. Multiple monitors not tested.

Please let me know if there's any silly mistakes in the code.

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -544,6 +544,27 @@ cursor_update_focus(struct server *server)
 }
 
 static void
+warp_cursor_to_constraint_hint(struct seat *seat,
+		struct wlr_pointer_constraint_v1 *constraint)
+{
+	if (!seat->server->focused_view) {
+		return;
+	}
+
+	if (constraint->current.committed
+			& WLR_POINTER_CONSTRAINT_V1_STATE_CURSOR_HINT) {
+		double sx = constraint->current.cursor_hint.x;
+		double sy = constraint->current.cursor_hint.y;
+		wlr_cursor_warp(seat->cursor, NULL,
+			seat->server->focused_view->current.x + sx,
+			seat->server->focused_view->current.y + sy);
+
+		/* Make sure we are not sending unnecessary surface movements */
+		wlr_seat_pointer_warp(seat->seat, sx, sy);
+	}
+}
+
+static void
 handle_constraint_commit(struct wl_listener *listener, void *data)
 {
 	struct seat *seat = wl_container_of(listener, seat, constraint_commit);
@@ -561,6 +582,8 @@ destroy_constraint(struct wl_listener *listener, void *data)
 
 	wl_list_remove(&constraint->destroy.link);
 	if (seat->current_constraint == wlr_constraint) {
+		warp_cursor_to_constraint_hint(seat, wlr_constraint);
+
 		if (seat->constraint_commit.link.next) {
 			wl_list_remove(&seat->constraint_commit.link);
 		}
@@ -600,6 +623,10 @@ constrain_cursor(struct server *server, struct wlr_pointer_constraint_v1
 	}
 	wl_list_remove(&seat->constraint_commit.link);
 	if (seat->current_constraint) {
+		if (!constraint) {
+			warp_cursor_to_constraint_hint(seat, seat->current_constraint);
+		}
+
 		wlr_pointer_constraint_v1_send_deactivated(
 			seat->current_constraint);
 	}


Consolatis:
The commit before was fine, I think it was just the line length > 100 that the CI complained about.
And no worries about the commits, we can squash them later before merging.

FuzzyQuills:
> The commit before was fine, I think it was just the line length > 100 that the CI complained about. And no worries about the commits, we can squash them later before merging.

No worries, thank you.

src/cursor.c
```diff
@@ -543,6 +543,20 @@ cursor_update_focus(struct server *server)
...
+warp_cursor_to_constraint_hint (struct seat *seat, struct wlr_pointer_constraint_v1
+*constraint)
+{
+	// TODO: Should the device pointer be passed in?
+	if (constraint->current.committed & WLR_POINTER_CONSTRAINT_V1_STATE_CURSOR_HINT) {
+		double cursor_hint_x = 	constraint->current.cursor_hint.x +
+					seat->server->focused_view->current.x;
+		double cursor_hint_y = 	constraint->current.cursor_hint.y +
+					seat->server->focused_view->current.y;
+		wlr_cursor_warp_closest(seat->cursor, NULL, cursor_hint_x, cursor_hint_y);
```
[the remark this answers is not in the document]
johanmalm: Convert to layout using view->current->x/y and wlr_output_layout_output_coords()

https://github.com/labwc/labwc/blob/d4288415c942bb3f8577f5fcddcaf49f199838f9/src/view.c#L450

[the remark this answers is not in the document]
Consolatis: They seem to be layout coordinates already? e.g. `seat->focused_view->current.x + constraint->current.cursor_hint.x`

src/cursor.c
```diff
@@ -543,6 +543,20 @@ cursor_update_focus(struct server *server)
...
+	}
```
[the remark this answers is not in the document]
Consolatis: Yes, this seems appropriate. It seems to basically just synchronize the wlroots internal seat cursor position on the focused surface with the new coordinates.

src/cursor.c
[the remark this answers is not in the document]
Consolatis: For the `FocusOutput` action we are using `wlr_cursor_warp()` as well, so unless we suddenly use an output or similar I don't see a reason to use the `_closest()` variant here.

Consolatis:
Did a slight code refactor and added a `wlr_seat_pointer_warp()` call.
Testing greatly appreciated. If it still works as intended we can squash the commits and get this merged.

FuzzyQuills:
> Did a slight code refactor and added a `wlr_seat_pointer_warp()` call. Testing greatly appreciated. If it still works as intended we can squash the commits and get this merged.

Hello, will test soon as I've got my coffee. 

Just read the review comments, I'll admit I did this after probably a good 2-3 hours staring at the code with no prior experience developing wlroots code, really appreciate the review.

Consolatis:
> I did this after probably a good 2-3 hours staring at the code with no prior experience developing wlroots code

This is exactly what I did as well when writing my first PR for labwc. Everybody has to start somewhere.
Finding an issue, figuring out the reason for it, figuring out how to fix it, sending a PR and finally getting that PR merged and thus the issue fixed for everybody is a nice feeling though. You did great.

FuzzyQuills:
The refractor does something strange; if I enter Practice Range in Overwatch 2 and don't move my mouse, opening the communication wheel fails to center the pointer. It also fails to center the pointer if I press Esc and move the pointer elsewhere.

Mouse movement before opening the communication wheel and every instance thereafter corrects this.

The original version always centered, so I'm not sure what could be wrong here.

I also witnessed similar behaviour in KDE when I was tracing it's wayland calls.

FuzzyQuills:
Okay, after further testing, I've been able to reproduce the following behaviour:
- If the mouse isn't moved at all after lock, the cursor hint isn't honoured.
- If the mouse *is* moved, even by 1 pixel, after the lock, the hint is honoured.

Given this is how KDE behaves, I'm going to assume this is wayland<->xwayland shenanigans, LTGM. It's ready. I might try improving it further at some point anyway as I tend to be a bit of a perfectionist with this stuff.

Consolatis:
But just commenting out the new `wlr_seat_pointer_warp` makes it work in all cases?

FuzzyQuills:
> But just commenting out the new `wlr_seat_pointer_warp` makes it work in all cases?

This was my initial thought but both behave the same as it turns out.

Consolatis:
Got it.

> Okay, after further testing, I've been able to reproduce the following behaviour:
> 
> * If the mouse isn't moved at all after lock, the cursor hint isn't honoured.
> * If the mouse _is_ moved, even by 1 pixel, after the lock, the hint is honoured.

Hm.. right. We only adjust the cursor on movement. I guess we could actually fix that as well in this PR.

src/cursor.c
```diff
@@ -600,6 +623,10 @@ constrain_cursor(struct server *server, struct wlr_pointer_constraint_v1
 	}
 	wl_list_remove(&seat->constraint_commit.link);
 	if (seat->current_constraint) {
+		if (!constraint) {
+			warp_cursor_to_constraint_hint(seat, seat->current_constraint);
+		}
```
Consolatis, line 628: Does removing the condition and always executing `warp_cursor_to_constraint_hint()` fix it?
Currently we only warp the pointer if we reset the lock but not when we replace it with a new one.

FuzzyQuills: Nope, same behaviour.
Could it be the early bail for cursor_locked doing it?

FuzzyQuills: Okay, this may actually be a game bug; I've just done the same test on my game engine which warps the pointer to the center on unlock and it works fine in all cases there. Might test on Windows.

Consolatis: Well, worth a shot.

> Could it be the early bail for cursor_locked doing it?
That is unlikely, that one is only triggered when actually moving the mouse and this PR should in fact update the cursor position (and image) even without movements.

If you are fine with the squashed commit and the current state I'd go for a merge of this PR.
We can always improve this in a new PR.

FuzzyQuills: > Well, worth a shot.
> 
> > Could it be the early bail for cursor_locked doing it?
> 
> That is unlikely, that one is only triggered when actually moving the mouse and this PR should in fact update the cursor position (and image) even without movements.
> 
> If you are fine with the squashed commit and the current state I'd go for a merge of this PR. We can always improve this in a new PR.

I just tested on X11 and it does center there even without movement. I'll look into this further but in it's current state I'm happy to have it merged.

Consolatis: While thinking a bit about this, two potential reasons:
- it might be possible that the hint is only set when xwayland receives relative motion events which we only send if the cursor is actually moved. Thus no movement after creating the lock: no update to the cursor hint.
- it might be possible that it isn't committed at that point, e.g. the hint only being available via `constraint->pending.cursor_hint`.

FuzzyQuills: > While thinking a bit about this, two potential reasons:
> 
> * it might be possible that the hint is only set when xwayland receives relative motion events which we only send if the cursor is actually moved. Thus no movement after creating the lock: no update to the cursor hint.
> * it might be possible that it isn't committed at that point, e.g. the hint only being available via `constraint->pending.cursor_hint`.

These were also my thoughts; it's something tied to an event that isn't sent yet. It sure looks like it.

Consolatis:
> I'll look into this further but in it's current state I'm happy to have it merged.

Merged. Thanks a lot :)

PR #877, opened 2023-04-14
Add Russian translation

Sunderland93:

PR #878, opened 2023-04-14
Add Indonesia translation 

mantarimay:


Consolatis:
Thanks for the translation.

CC @01micko 

Once approved I will squash the commits before merging.

johanmalm:
Thanks.

01micko:
Apologies for tardy response but this and #877 LGTM without a thorough analysis.

johanmalm:
> Apologies for tardy response but this and #877 LGTM without a thorough analysis.

I did a google translate on a few and just went for it :smile:

johanmalm:
> Shouldn't it be `Indonesian` instead of `Indonesia` for `Language:`? Edit: Also the gettext example files use the short form id, es, de etc, does that matter?

Yes, ought to be `Indonesian`. 

I don't know much about PO files, but it looks like the first entry with an empty `msgid` string is considered as a header and that it contains meta-information about the content in the file. For the purposes of what we do, it doesn't seem like it makes a difference in practice, so I'm not overly bothered. Happy to merge if anyone fancies tidying up though.

Yes, looks like GNU guidance is `ll` notation (e.g. `id` and `de`). https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html

PR #886, opened 2023-04-19
Support window switcher field configuration

johanmalm:


docs/labwc-config.5.scd
```diff
@@ -87,6 +89,15 @@ The rest of this man page describes configuration options.
...
 	switching between windows. Default is yes.
 
+*<windowSwitcher><fields><field content="" width="%">*
+	Define window switcher fields which can be any of:
+
+	*type* Show view type ("xdg-shell" or "xwayland")
+
+	*app_id* Show app_id (class for XWayland)
+
+	*title* Show window title if different to app_id
```
Consolatis, line 99: Suggest adding a line like `*content* values can be one of:` and then the type, app_id and title explanations.

Do we want to document `width` being a percentage of whatever is the width of the OSD for completeness?

src/config/rcxml.c
```diff
@@ -51,6 +55,38 @@ enum font_place {
...
+	} else if (!strcmp(nodename, "content")) {
+		if (!current_field->content) {
+			current_field->content = xstrdup(content);
+		}
+	} else if (!strcmp(nodename, "width") && !strchr(content, '%')) {
+		wlr_log(WLR_ERROR, "Removing invalid field '%s': %s='%s' misses"
+			" a trailing %%", current_field->content, nodename, content);
+		wl_list_remove(&current_field->link);
+		zfree(current_field->content);
+		zfree(current_field);
```
Consolatis, line 81: `current_field = NULL`? Otherwise we'd have a dangling pointer around

docs/labwc-config.5.scd
johanmalm: Good point. Done.

src/config/rcxml.c
johanmalm: zfree() does that

Consolatis: Oh, wow. That is what reviewing code while tired does.

include/config/windowswitcher-field.h
```diff
@@ -0,0 +1,11 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef __LABWC_WINDOWSWITCHER_FIELD_H
+#define __LABWC_WINDOWSWITCHER_FIELD_H
+
+struct windowswitcher_field {
+	char *content;
+	int width;
+	struct wl_list link; /* struct rcxml.windowswitcher.fields */
+};
```
Consolatis, line 9: I am not sure if we need the new header file here, we could also define this in the generic `rcxml.h` header.

src/config/rcxml.c
```diff
@@ -437,13 +477,17 @@ entry(xmlNode *node, char *nodename, char *content)
...
+	} else if (!strcasecmp(nodename, "preview.windowSwitcher")) {
 		rc.cycle_preview_contents = get_bool(content);
-	} else if (!strcasecmp(nodename, "outlines.windowSwitcher.core")) {
+	} else if (!strcasecmp(nodename, "outlines.windowSwitcher")) {
 		rc.cycle_preview_outlines = get_bool(content);
 
+	/* Remove this long term - just a friendly warning for now */
+	} else if (strstr(nodename, "windowswitcher.core")) {
+		wlr_log(WLR_ERROR, "<windowSwitcher> should not be child of <core>");
+
```
Consolatis, line 490: Should we also parse the old variants? e.g. something like
```c
if (strcasestr(nodename, "windowSwitcher.core")) {
	string_truncate_at_pattern(nodename, ".core");
	wlr_log(WLR_ERROR, "<windowSwitcher> should not be child of <core>");
}

..
} else if (!strcasecmp(nodename, "show.windowSwitcher") {
	rc.cycle_view_osd = get_bool(content);
} ..

```

include/config/windowswitcher-field.h
Consolatis, line 9: We could use an enum instead of `char *content` here, e.g. parsing the string in `rcxml.c` and then just setting the enum, that way the OSD doesn't have to do the string comparisons every time it renders.

And we can also check the `content` value while parsing the config and simply reject invalid ones. Currently it seems the OSD would spam the log in case a value is invalid.

include/config/rcxml.h
```diff
@@ -65,6 +65,10 @@ struct rcxml {
 
 	/* Regions */
 	struct wl_list regions;  /* struct region.link */
+
+	struct {
+		struct wl_list fields;  /* struct windowswitcher_field.link */
+	} windowswitcher;
```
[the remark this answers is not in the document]
FuzzyQuills: I agree with the naming of the windowswitcher struct, kind of looks like windows witcher if you misread it haha.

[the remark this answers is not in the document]
johanmalm: Done.
Have also put the `cycle_*` bools in the new struct as a separate commit

include/config/windowswitcher-field.h
johanmalm: Have removed new header file and moved struct to `rcxml.h`

src/config/rcxml.c
johanmalm: Okay, good point. Now supporting old variants.
Spelt out the full nodenames to make it super-clear what's happening rather than using the string_truncate... helper.

include/config/windowswitcher-field.h
johanmalm: Very good point. Done.

johanmalm:
All done. Should be ready to go except I just need to `rebase -i` all the fixups before merge.

Consolatis:
LGTM, didn't test.

I assume we should start to restrict the width of the text in the single fields at some point so we don't end up overwriting other fields due to overlong app_ids or small field widths. E.g. something like `pango_layout_set_width()` + `pango_layout_set_ellipsize()`. I am not sure how that works with the single fields though, maybe pango supports layered layouts?

johanmalm:
> LGTM, didn't test.

Thanks. Will go for the merge.

> I assume we should start to restrict the width of the text in the single fields at some point so we don't end up overwriting other fields due to overlong app_ids or small field widths. E.g. something like `pango_layout_set_width()` + `pango_layout_set_ellipsize()`.

Will do.

Issue #874, opened 2023-04-12
League of Legends window breakage.


FuzzyQuills:
When League of Legends launches the main game  after the client finalises champ select, the game window has several issues:
- Any toplevel windows owned by panels (aka. waybar) still show on top and are clickable.
- All other windows are always below the game window. (Except wofi, I'm guessing this is because wofi likely uses the overlay?)
- The game window completely lacks decorations when using windowed mode, making it impossible to exit the game in some cases.
- The window itself in all cases is completely missing in the window switcher, making it possible to interact with a window behind the game window using a keyboard.

This also has the potential to break the game's keyboard input as in one test I managed to jam the Ctrl key which meant I kept pinging with left click.

Current workarounds I've found to this issue:
- Run the game windowed, but only after unticking "let the window manager decorate the windows" in winecfg. This fixes the window switcher and is far less buggy in general.
- Use Gamescope. I haven't tested this.

johanmalm:
If you rum with debug log (-d) do you get anything relating to this?? https://github.com/labwc/labwc/blob/d4288415c942bb3f8577f5fcddcaf49f199838f9/src/xwayland-unmanaged.c#L134

Also, we’ve not yet implemented: set_role, set_window_type and set_hints
https://github.com/labwc/labwc/blob/d4288415c942bb3f8577f5fcddcaf49f199838f9/include/xwayland.h#L38

Does it work with sway? If so, would you be able to do some printfs in the callbacks associated with those listeners to try to build a picture of what we need to do?

From what you’ve described it seems like we need to convert some unmanaged surface to a toplevel window.

Consolatis:
> The window itself in all cases is completely missing in the window switcher, making it impossible to alt-tab back without clicking in the window.

This indeed sounds like an unmanaged window. I can't think of another reason why a window wouldn't show up in the window switcher. Either wayland native and layer-shell or xwayland and unmanaged.

FuzzyQuills:
> If you rum with debug log (-d) do you get anything relating to this??
> 
> https://github.com/labwc/labwc/blob/d4288415c942bb3f8577f5fcddcaf49f199838f9/src/xwayland-unmanaged.c#L134
> 
> Also, we’ve not yet implemented: set_role, set_window_type and set_hints
> 
> https://github.com/labwc/labwc/blob/d4288415c942bb3f8577f5fcddcaf49f199838f9/include/xwayland.h#L38
> 
[...]

PR #887, opened 2023-04-19
xwayland: validate PID before activating unmanaged surface

johanmalm:
Check that an unmanaged surface trying to grab focus is actually a child of the topmost mapped view.  Without this patch League of Legends is reported to re-grab focus within milli-seconds of alt-tabbing away from the game.

Related-to: #874

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -374,6 +374,7 @@ void foreign_toplevel_update_outputs(struct view *view);
 void desktop_focus_and_activate_view(struct seat *seat, struct view *view);
 void desktop_arrange_all_views(struct server *server);
 void desktop_focus_output(struct output *output);
+struct view *desktop_topmost_mapped_view(struct server *server);
 
 enum lab_cycle_dir {
 	LAB_CYCLE_DIR_NONE,
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -200,8 +200,8 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 	return NULL;
 }
 
-static struct view *
-topmost_mapped_view(struct server *server)
+struct view *
+desktop_topmost_mapped_view(struct server *server)
 {
 	struct view *view;
 	struct wl_list *node_list;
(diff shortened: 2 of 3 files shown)


johanmalm:
Further to #874

> Only just now managed to try it and I actually still have issues with the game. I can't switch focus with alt tab, it only flickers to the other window for a millisecond and then immediately switches back to the game client. It did fix the game window following me when switching workspaces, but if I switch and click on a window to focus it, it's active for a millisecond before its inactive again.

@FuzzyQuills @Flrian Does this patch help?
Cc @Consolatis

FuzzyQuills:
> Further to #874
> 
> > Only just now managed to try it and I actually still have issues with the game. I can't switch focus with alt tab, it only flickers to the other window for a millisecond and then immediately switches back to the game client. It did fix the game window following me when switching workspaces, but if I switch and click on a window to focus it, it's active for a millisecond before its inactive again.
> 
> @FuzzyQuills @Flrian Does this patch help? Cc @Consolatis

Unfortunately no; it still steals focus away from the user when trying to alt-tab to another window.
I honestly think it might be a game bug, but I'd have to test on Windows.

FuzzyQuills:
> As far as I understood in case of LoL it starts up as unmanaged and then switches to managed. So this seems to be about `handle_request_activate()` in `src/xwayland.c` rather than `src/xwayland-unmanaged.c`. However, this PR might still make sense regardless of the strange LoL behavior. E.g. should we keep allowing any random unmanaged window to steal focus? On the other side, the whole X11 concept is inherently requiring application to behave nice.

I was not aware LoL was doing this, that's quite strange. In that case, it might be worth looking at managed request_activate too.

Consolatis:
For the managed part its not that easy to add a filter. E.g. compare firefox requesting activation because you clicked on some link in a terminal or email client vs LoL requesting activation. How should we differentiate between these two usecases?

(Other than starting to check app_ids / titles, that doesn't sound like something labwc should start doing). We could potentially add a setting for that once the window policy framework becomes a thing though, so users could configure their system to reject request_activate for specific applications.

johanmalm:
> ...this PR might still make sense regardless of the strange LoL behavior. E.g. should we keep allowing any random unmanaged window to steal focus?

Yes, I think we should check PIDs for the unmanaged surfaces. Have removed the mention of LoL from the comment though.

Consolatis:
I am not sure about keeping the name of `topmost_mapped_view()` in the "global" namespace, usually it starts with the compilation unit name. Other than that this looks good to me, didn't test.

johanmalm:
> I am not sure about keeping the name of `topmost_mapped_view()` in the "global" namespace, usually it starts with the compilation unit name. 

Good catch. Done.

PR #889, opened 2023-04-21
osd: set field width to ellipsize

johanmalm:


Consolatis:
Tested nested and seems to work as it should. The whole implementation also handles a field width of `120%` and with this PR it now cuts off correctly when using something like `10%`. Nice work!

johanmalm:
Thanks. Will fix as part of https://github.com/labwc/labwc/pull/893

PR #896, opened 2023-04-26
parse-bool.c: add helpers to set bool/int iff valid boolean

johanmalm:


Consolatis:
This is looking good to me.
Suggestion: `s/get_bool/set_bool/`

johanmalm:
I have:

- Changed `s/get_bool/set_bool/`. Good point :smile:
- Tweaked the documentation in the header file a bit
- Changed the function variable name from `value` to `variable`

Any thoughts on

1. `s/set_bool_as_int/set_int/` - too simple??
2. Or something like `parse_bool_text_and_set_int()` - too wordy?

Consolatis:
For me, `set_int()` would imply a `atoi()` or `strtol()` parser.
`set_int_from_bool()`? Although it misses the parsing bit as well. I am not sure.

With `set_bool_as_int()` we have a least a consistent family of `set_bool()` functions that all parse a string input as boolean and if successful, set it.

Proper naming is hard :)

johanmalm:
^ Thanks for help thinking through. I'll just go as is then. Probably simplest. We can always improve later if we have a flash of inspiration.

PR #897, opened 2023-04-27
Ignore SIGPIPE

### LINKED ISSUE #890: SIGPIPE when taking a screenshot in Overwatch 2.
Apparently happens in main.c at line 179, somewhere inside wlroots:
```
#0  __GI___libc_write (nbytes=2490422, buf=0x7fffb3cca030, fd=66) at ../sysdeps/unix/sysv/linux/write.c:26
#1  __GI___libc_write (fd=fd@entry=66, buf=0x7fffb3cca030, nbytes=nbytes@entry=2490422) at ../sysdeps/unix/sysv/linux/write.c:24
#2  0x00007ffff7f017f7 in write_selection_property_to_wl_client (fd=66, mask=<optimized out>, data=0x5555561bc670) at ../wlroots-0.16.2/xwayland/selection/incoming.c:111
#3  0x00007ffff7f7cae2 in wl_event_loop_dispatch (loop=0x5555555bdd00, timeout=timeout@entry=-1) at ../wayland-1.22.0/src/event-loop.c:1027
#4  0x00007ffff7f7d2d7 in wl_display_run (display=0x5555555bdc10) at ../wayland-1.22.0/src/wayland-server.c:1493
#5  0x0000555555568313 in main (argc=1, argv=0x7fffffffdf18) at ../src/main.c:179
```
Not sure if this is even a wlroots or a labwc bug; I only started getting this SIGPIPE crash a month ago. It's also embarassingly cost me a few competitive matches due to leaving early. (because labwc had terminated, taking everything with it)

Could it also be possible glibc broke something again?

To reproduce:
- Boot Overwatch 2 inside a labwc session.
- At login, in the menus, in match, practically anywhere else... press PrntScr to take a screenshot. Sometimes the first screenshot doesn't kill the compositor however, so just keep taking screenshots.
- Observe the session instantly quit after a few seconds.
- If GDB is open and attached to labwc, it reports a SIGPIPE.

Consolatis:
Prevent wayland clients that request the X11 clipboard but closing their read fd prematurely to crash labwc because of the unhandled SIGPIPE signal. It is caused by wlroots trying to write the X11 clipboard data to the closed fd of the wayland client.

See https://github.com/labwc/labwc/issues/890#issuecomment-1524962995 for a reproducer involving xclip and wl-paste | head -c 1.

Fixes #890
Reported-by: @FuzzyQuills

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -227,6 +227,16 @@ server_init(struct server *server)
 	server->wl_event_loop = event_loop;
 
 	/*
+	 * Prevent wayland clients that request the X11 clipboard but closing
+	 * their read fd prematurely to crash labwc because of the unhandled
+	 * SIGPIPE signal. It is caused by wlroots trying to write the X11
+	 * clipboard data to the closed fd of the wayland client.
+	 * See https://github.com/labwc/labwc/issues/890#issuecomment-1524962995
+	 * for a reproducer involving xclip and wl-paste | head -c 1.
+	 */
+	signal(SIGPIPE, SIG_IGN);
+
+	/*
 	 * The backend is a feature which abstracts the underlying input and
 	 * output hardware. The autocreate option will choose the most suitable
 	 * backend based on the current environment, such as opening an x11


johanmalm:
LGTM

PR #899, opened 2023-04-27
output: Make the session lock layer the topmost

heroin-moose:
Make sure that only the session lock layer is visible to a user. Fixes the issue with mako notifications being rendered above the swaylock window.


Consolatis:
Thanks :)

johanmalm:
Indeed. Thank you :smile:

PR #901, opened 2023-04-29
[wip] Support window-rules

johanmalm:
Add config option

    <applications>
      <application app_id="">
        <action name=""/>
      </application>
    </applications>


src/config/rcxml.c
```diff
@@ -1059,6 +1106,14 @@ rcxml_finish(void)
...
 	}
 
+	struct window_rule *rule, *rule_tmp;
+	wl_list_for_each_safe(rule, rule_tmp, &rc.window_rules, link) {
+		wl_list_remove(&rule->link);
+		zfree(rule->app_id);
+		action_list_free(&rule->actions);
+		zfree(rule);
+	}
+
```
Consolatis, line 1116: I think the current globals aren't reset, see the section below.

johanmalm: Thanks. Sorted.

johanmalm:
Ready for sanity checking and critical review. Have to consider other approaches, but I think it's landed quite well.

See further comments in commit messages + documentation added.

m1027:
I've tried labwc from your branch with the above config snippet for Alacritty, however, it doesn't toggle to maximized state initially as expected. Currently I don't see what could be missing...

docs/labwc-config.5.scd
```diff
@@ -330,6 +330,50 @@ The rest of this man page describes configuration options.
 	any motion events while a keyboard is typing, and for a short while
 	after as well.
 
+## WINDOW RULES
+
+Two types of window rules are supported, actions and properties. They are
+defined as shown below. Property values can be *yes*, *no* or *unset*.
```
Consolatis, line 336: I am not sure we need to expose the `unset` argument to the user.
We need it internally to differentiate between a user supplying `no` vs not setting the property at all though.

docs/labwc-config.5.scd
```diff
@@ -330,6 +330,50 @@ The rest of this man page describes configuration options.
...
+	case-insensitive.
+
+	*identifier* relates to app_id for native Wayland windows and WM_CLASS
+	for XWayland clients.
+
+	*event* is only required for window-rules with *<action>* and defines
+	the event that the rule relates to.  The following events are
+	supported:
+
+		- *onMap*
```
Consolatis, line 371: Could there be any more events we want to react to?
If there are none I think we could simply drop the `event` attribute and always execute the actions (if defined) on map.

docs/labwc-config.5.scd
```diff
@@ -330,6 +330,50 @@ The rest of this man page describes configuration options.
...
+
+*<windowRules><windowRule identifier=""><serverDecoration>* [yes|no|unset]
+	*serverDecoration* over-rules any other setting for server-side window
+	decoration
```
Consolatis, line 375: I'd suggest having a parent section about properties which then describes all the properties. Currently this seems on the same level as the generic description of the window rule.

docs/rc.xml.all
```diff
@@ -374,4 +374,15 @@
...
 
+  <!--
+  <windowRules>
+    <windowRule identifier="" event="">
+      <action name="">
+    </windowRules>
+    <windowRule identifier="">
+      <serverDecoration></serverDecoration>
+    </windowRules>
+  </windowRules>
```
Consolatis, line 385: Having this example completely empty feels a bit strange, I'd suggest filling it with something which makes it easier to understand by the context on the first look, especially if somebody didn't read the man page.

docs/labwc-config.5.scd
```diff
@@ -330,6 +330,50 @@ The rest of this man page describes configuration options.
...
+
+  <!-- Action -->
+  <windowRule identifier="" event="">
+    <action name="">
+  </windowRules>
+
+  <!-- Property -->
+  <windowRule identifier="">
+    <serverDecoration></serverDecoration>
+  </windowRules>
```
Consolatis, line 349: `s/windowRules/windowRule/` for the closing ones.

Even though `magic malm parser` handles both, I'd suggest using the compact syntax in all examples:
```xml
  <!-- Property -->
  <windowRule identifier="some.application" serverDecorations="yes" />
```

include/common/match.h
```diff
@@ -0,0 +1,12 @@
...
+#define __LABWC_MATCH_H
+
+/**
+ * match() - Pattern match case-insensitively using '*' wildcards and '?' jokers
+ * @pattern: Pattern to match against.
+ * @string: String to search.
+ */
+bool match(const gchar *pattern, const gchar *string);
+
+#endif /* __LABWC_MATCH_H */
```
Consolatis, line 12: This seems to miss an glib include due to the use of `gchar`. Otherwise it would always have to be included after something else which includes glib.

include/config/rcxml.h
```diff
@@ -80,6 +80,8 @@ struct rcxml {
 		bool outlines;
 		struct wl_list fields;  /* struct window_switcher_field.link */
 	} window_switcher;
+
+	struct wl_list window_rules;
```
Consolatis, line 84: Missing comment regarding the struct and link member

src/common/match.c
```diff
@@ -0,0 +1,19 @@
...
+#include <stdbool.h>
+#include "common/match.h"
+
+bool
+match(const gchar *pattern, const gchar *string)
+{
+	GString *p = g_string_new(pattern);
+	GString *s = g_string_new(string);
+	g_string_ascii_down(p);
+	g_string_ascii_down(s);
```
Consolatis, line 13: Do we need a unicode down variant here? I am not sure if there are any restrictions to only use ASCII for class or app_id.

src/common/match.c
```diff
@@ -0,0 +1,19 @@
...
+	bool ret = (bool)g_pattern_match_simple(p->str, s->str);
```
Consolatis, line 14: Do we need the additional manual cast here?

src/config/rcxml.c
```diff
@@ -54,6 +58,59 @@ enum font_place {
...
+	} else if (!current_window_rule) {
+		wlr_log(WLR_ERROR, "no window-rule");
+	} else if (!strcmp(nodename, "identifier")) {
+		free(current_window_rule->identifier);
+		current_window_rule->identifier = xstrdup(content);
+	} else if (!strcmp(nodename, "event")) {
+		if (!strcasecmp(content, "onMap")) {
+			current_window_rule->event = LAB_WINDOW_RULE_EVENT_ON_MAP;
+		}
+	} else if (!strcmp(nodename, "serverdecoration")) {
```
Consolatis, line 98: Mixed use of `strcmp` and `strcasecmp` in this function, especially for the term `serverDecorations` I think we should use `strcasecmp` instead.

src/view.c
```diff
@@ -658,6 +659,21 @@ void
...
+		decorate(view);
+		return;
+	case LAB_PROP_FALSE:
+		view->ssd_enabled = false;
+		undecorate(view);
+		return;
+	default:
+		break;
+	}
+
```
Consolatis, line 676: I am not sure if we should use `view_set_decorations` to implement this. I didn't test it but I'd assume this will stop `ToggleDecorations` to work. If that is the case we should likely just set the initial `view->ssd_enabled` bool on map.
Although I am not sure right now how that interacts with the 2 decoration protocols.

src/window-rules.c
```diff
@@ -0,0 +1,83 @@
...
+{
+	switch (view->type) {
+	case LAB_XDG_SHELL_VIEW:
+		return view_get_string_prop(view, "app_id");
+#if HAVE_XWAYLAND
+	case LAB_XWAYLAND_VIEW:
+		return view_get_string_prop(view, "class");
+#endif
+	}
+	return "";
```
Consolatis, line 27: `src/xwayland.c` responds to a `app_id` request as well so I assume we can just always ask for `app_id`.

src/window-rules.c
```diff
@@ -0,0 +1,83 @@
...
+		if (!rule->identifier) {
+			continue;
+		}
+		if (match(rule->identifier, identifier)) {
+			if (!strcasecmp(property, "serverDecoration")) {
+				return rule->server_decoration;
+			}
+			wlr_log(WLR_ERROR, "unknown property '%s'", property);
+		}
+	}
```
Consolatis, line 80: I think we should `continue` when the requested value is set to `LAB_PROP_UNSPECIFIED`. Otherwise this would return `LAB_PROP_UNSPECIFIED` if asked for the `serverDecorations` of `foo.bar`:
```xml
<windowRules>
    <windowRule identifier="*" serverDecorations="no" />
    <windowRule identifier="foo.bar" />
```

docs/labwc-config.5.scd
johanmalm: The reason for 'unset' is that if a property has already been set to yes/no using for example `identifier="*"`, we might then want to unset the property for a particular window in order for it to fall back onto default values (or in the example server decoration, the two related protocol implementations).

docs/labwc-config.5.scd
johanmalm: Not sure.

I just wanted to avoid breaking changes later, but I suppose we could just default to 'onMap'.

Sway only apply window-rules when windows 'appear' - so on the basis that sway users have not required anything different, we're probably safe :smile:

Wayfire supports the following events: created, unmaximized, maximized, minimized and fullscreened
I currently can't think of use-cases for these.

docs/labwc-config.5.scd
johanmalm: Thanks. Done.

include/common/match.h
johanmalm: Added `#include <glib.h>`

Regarding naming... :smile: I did actually call it `match-glob.c` and `match_glob()` - but changed when I realized that we'll have a clash with the osd-PR and I was keen to avoid a merge conflict (sorry for being so lazy)

include/config/rcxml.h
johanmalm: Whoops. Thanks for catching. Done.

src/common/match.c
johanmalm: Good shout. I think I'd like to add `title` at some point, so converting strings to form independent of case seems very sensible

It actually looks a lot simpler anyway :smile:

```
diff --git a/src/common/match.c b/src/common/match.c
index c4992e4..46de22c 100644
--- a/src/common/match.c
+++ b/src/common/match.c
@@ -7,13 +7,7 @@
 bool
 match(const gchar *pattern, const gchar *string)
 {
-	GString *p = g_string_new(pattern);
-	GString *s = g_string_new(string);
-	g_string_ascii_down(p);
-	g_string_ascii_down(s);
-	bool ret = (bool)g_pattern_match_simple(p->str, s->str);
-	g_string_free(p, true);
-	g_string_free(s, true);
-	return ret;
+	return (bool)g_pattern_match_simple(g_utf8_casefold(pattern, -1),
+		g_utf8_casefold(string, -1));
 }
 ```

Comment (johanmalm) [reply src/common/match.c]: Change pushed.

File: src/common/match.c (hunk 10 above)
Comment (johanmalm) [reply src/common/match.c]: It appears not. Done.

File: src/config/rcxml.c (hunk 11 above)
Comment (johanmalm) [reply src/config/rcxml.c]: Have changed to `strcasecmp` and `serverDecoration` because it's easier to read.

`nodename` is always lower-case so not change in functionality - just easier to read (at the expense of a few CPU cycles).

File: src/view.c (hunk 12 above)
Comment (johanmalm) [reply src/view.c:676]: It depends on if we want the property to *always* apply - or only apply when the view is first mapped.

I thought that the always (persistent) approach was best - but don't mind.

File: src/window-rules.c (hunk 13 above)
Comment (johanmalm) [reply src/window-rules.c:27]: Yes, although it's clearer what's going on the way it is written, don't you think?

File: src/window-rules.c (hunk 14 above)
Comment (johanmalm) [reply src/window-rules.c]: Good catch indeed.
I've pushed a fix which differentiates between unspecified and unset.

**Comment by johanmalm** [comment]
Thanks for detailed review :smile:

File: docs/labwc-config.5.scd (hunk 6 above)
Comment (Consolatis) [reply docs/labwc-config.5.scd:349]: Lost a commit? :D

File: docs/labwc-config.5.scd [hunk 15]
```diff
@@ -330,6 +330,56 @@ The rest of this man page describes configuration options.
...
+
+		- *onMap*
+
+*Properties*
+
+Property values can be *yes*, *no* or *unset*.
+
+If a window matches criteria for multiple rules which set the same property,
+later config entires have higher priority. *unset* can be useful in this
+situation.
```
Consolatis, line 377: `s/entires/entries/`

Potentially `s/unset/default/`, see other thread

src/common/match.c
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include <glib.h>
```
Consolatis, line 3: The glib include could be removed here as its now included in the header

src/view.c
Consolatis: I'd argue that we shouldn't break manual `ToggleDecorations` actions by the user, even if some rule had been set.

src/window-rules.c
Consolatis: We'd get rid of the `#if HAVE_XWAYLAND` though. Maybe we could add a handler for the string `identifier` instead? And then have either `src/view.c` or `src/xdg.c` / `src/xwayland.c` handle it appropriately.

docs/labwc-config.5.scd
Consolatis: What do you think about renaming it to `default` in that case?
`unspec` feels a bit "codish" to me for a user facing config file.

Consolatis:
Just a short re-review, will take a more detailed look tomorrow again and respond to the remaining open threads.

johanmalm:
> I've tried labwc from your branch with the above config snippet for Alacritty, however, it doesn't toggle to maximized state initially as expected. Currently I don't see what could be missing...

Not sure. Works for me. Could you try with another application?? And show your `<windowRules>` config.

docs/labwc-config.5.scd

docs/labwc-config.5.scd
johanmalm: Done now. Had done it in the rc.xml.all file but not the man page :smile:

src/view.c
johanmalm: On second thought, agree. Have moved that code to has_ssd() and want_deco(). Seems to work fine.

src/window-rules.c
johanmalm: Okay. Done. Have got for a simple call for `prop("app_id")`.

docs/labwc-config.5.scd
johanmalm: Done.

src/common/match.c
johanmalm: Done.

johanmalm:
All done except for:

- Getting rid of event attribute. Happy to get rid, just wanted to await response to comments higher up. Have also moved the event to be onFirstMap (but not yet updated documentation), so either way we need a tidy up here.
- s/match/match_bool/

docs/rc.xml.all
```diff
@@ -374,4 +374,13 @@
     </device>
   </libinput>
 
+  <!--
+  <windowRules>
+    <windowRule identifier="*" event="onMap">
+      <action name="Maximize">
```
Consolatis, line 380: `ToggleMaximize`?

Consolatis:
Other than the open question regarding the `event` attribute and the `rc.xml.all` nitpick above this LGTM. Didn't test though.

m1027:
> > I've tried labwc from your branch with the above config snippet for Alacritty, however, it doesn't toggle to maximized state initially as expected. Currently I don't see what could be missing...
> 
> Not sure. Works for me. Could you try with another application?? And show your `<windowRules>` config.

I have just tried again with your most recent commit: `[fixup] apply window-rules on first map only`. However, alacritty just doesn't appear maximized.

From `rc.xml`:
```
<windowRules>
  <windowRule identifier="Alacritty" event="onMap">
    <action name="ToggleMaximized"/>
  </windowRule>
</windowRules>
```
If it works for you I assume it is something trivial... Anyway, no hurry here.

Consolatis:
Try `ToggleMaximize` instead (note the additional `d` at the end).

johanmalm:
Right. Have added a `Maximize` action. :smile:
Ready to go!
@m1027 could you try with the latest force-push and the following in your config:

```XML
<windowRules>
  <windowRule identifier="*">
    <action name="Maximize"/>
  </windowRule>
</windowRules>
```

PR #913, opened 2023-05-09
window-rules: support title criteria

johanmalm:
Example config:

```
<windowRules>
  <windowRule identifier="foot" title="foo"><action name="Maximize"/></windowRule>
</windowRules>
```

Observe that:

- `foot -T foo` starts maximized
- `xterm -T foo` starts normal

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -376,11 +376,19 @@
   </libinput>
 
   <!--
+    Window Rules
+      - Criteria can consist of 'identifier' or 'title' or both (in which case
+        AND logic is used).
+      - 'identifier' relates to app_id for native Wayland windows and WM_CLASS
+        for XWayland clients.
+      - Matching against patterns with '*' (wildcard) and '?' (joker) is
+        supported. Pattern matching is case-insensitive.
+
   <windowRules>
-    <windowRule identifier="*">
-      <action name="Maximize" />
-    </windowRule>
-    <windowRule identifier="some-application" serverDecoration="yes" />
+    <windowRule identifier="*"><action name="Maximize"/></windowRule>
+    <windowRule identifier="foo" serverDecoration="yes"/>
+    <windowRule title="bar" serverDecoration="yes"/>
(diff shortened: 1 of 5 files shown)


src/window-rules.c
```diff
@@ -13,20 +14,44 @@
...
+			return false;
+		}
+		return match_glob(rule->identifier, id);
+	} else if (rule->title) {
+		if (!title) {
+			return false;
+		}
+		return match_glob(rule->title, title);
+	} else {
+		wlr_log(WLR_ERROR, "rule has no identifier or title\n");
```
Consolatis, line 41: Another candidate for a `validate()` check in `src/rcxml.c`

docs/rc.xml.all
```diff
@@ -380,7 +380,7 @@
     <windowRule identifier="*">
       <action name="Maximize" />
     </windowRule>
-    <windowRule identifier="some-application" serverDecoration="yes" />
+    <windowRule title="some-title" serverDecoration="yes" />
   </windowRules>
```
Consolatis, line 384: Should we add examples for all combinations? E.g
- only identifier (set to `*`, so the first example rule)
- identifier and title set
- only title set

Consolatis:
Only nitpicking. LGTM, didn't test.

Consolatis:
> Example config:
> 
> Observe that:
> 
> * `foot -T foo` starts maximized
> * `xterm -T foo` starts normal

I don't quite understand what this means though.

src/window-rules.c
johanmalm: Have added `validate()` which removes window-rules without criteria.

docs/rc.xml.all
johanmalm: Have added one of each + both.

johanmalm:
> > Example config:
> > Observe that:
> > 
> > * `foot -T foo` starts maximized
> > * `xterm -T foo` starts normal
> 
> I don't quite understand what this means though.

There was a config snippet there which I hadn't indented properly, so it didn't show. Same as the commit message.

PR #915, opened 2023-05-10
Do not render workspace OSD on disabled outputs

### LINKED ISSUE #914: Popup when switching a workspace is shown two times
The popup when switching a workspace is shown two times on a wide display (30inch 2560x1080 px, 180.000000 Hz).
One popup is set perfectly in the middle.
The other popup appears ca 2 cm to the left and 0,5 cm up.
Otherwise the popup work fine, it just looks weird.

Consolatis:
Fixes #914


johanmalm:
LGTM
Didn’t test

Issue #894, opened 2023-04-25
`SendToDesktop` without `to` parameter: triggering the action crashes labwc


jacopo-salamina:
Whenever I send a window to another desktop, labwc immediately crashes.

Minimal reproducible example:

- edit `rc.xml` like this (I chose `Alt-Left` and `Alt-Right` as shortcuts, but other ones are fine):
```xml
<?xml version="1.0"?>
<labwc_config>
  <desktops>
    <popupTime>0</popupTime>
    <names>
      <name>1</name>
      <name>2</name>
    </names>
  </desktops>
  <keyboard>
    <keybind key="A-l">
      <action name="SendToDesktop" direction="left" />
    </keybind>
    <keybind key="A-r">
      <action name="SendToDesktop" direction="right" />
    </keybind>
  </keyboard>
</labwc_config>
```
- edit `autostart` so it opens a single GUI application, and nothing else (`pcmanfm-qt &` in my case);
- delete both `menu.xml` and `environment`;
- start labwc (either manually or using a display manager);
- using the chosen keyboard shortcut, send the GUI application to the desktop on the left/right.

I also tried to inspect the generated coredump, and I found this backtrace (full gdb log below):

```
#0  __strcmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:287
#1  0x00005606dcbea785 in actions_run (activator=0x0, server=0x7fff9b810200, actions=0x5606dd019f58, resize_edges=0) at ../src/action.c:476
#2  0x00005606dcbf04f8 in handle_keybinding (server=0x7fff9b810200, modifiers=8, sym=108) at ../src/keyboard.c:101
#3  0x00005606dcbf0896 in handle_compositor_keybindings (keyboard=0x5606dd9c84e0, event=0x7fff9b80ff20) at ../src/keyboard.c:238
...
```

Additional information:

- OS: Arch Linux;
- installed dependencies:

| Package | Version |
| - | - |
| wlroots | 0.16.2-1 |
| wayland | 1.22.0-1 |
| libinput | 1.23.0-1 |
| libxkbcommon | 1.5.0-1 |
| libxml2 | 2.10.4-2 |
| cairo | 1.17.8-2 |
| pango | 1:1.50.14-1 |
| glib2 | 2.76.1-1 |
| xorg-xwayland | 23.1.1-1 |
| libxcb | 1.15-2 |
| meson | 1.0.1-1 |
| ninja | 1.11.1-2 |
[...]

PR #919, opened 2023-05-11
More robust action argument parsing

Consolatis:
And assign argument names.

Missing:
- [x] add asserts() for the argument name in `action_arg_add_*()`
- [x] remove `if (!arg->key)` check in `get_arg_value_*()`
- [x] potentially use `string_truncate_at_pattern()` on the original `nodename` argument
- [x] verify that all action and argument combinations have been properly defined here
- [x] add `bool action_valid()` to verify that all required arguments exist for the given action and call it from the `validate()` hook in `rcxml.c`
- [x] same for `menu.c`
- [x] same for window rules
- [x] remove checks for required arguments when running actions as we now do that at config load time

Fixes:
- #894

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -122,38 +122,38 @@ const char *action_names[] = {
 };
 
 void
-action_arg_add_str(struct action *action, char *key, const char *value)
+action_arg_add_str(struct action *action, const char *key, const char *value)
 {
+	assert(action);
+	assert(key);
 	assert(value && "Tried to add NULL action string argument");
 	struct action_arg_str *arg = znew(*arg);
 	arg->base.type = LAB_ACTION_ARG_STR;
-	if (key) {
-		arg->base.key = xstrdup(key);
-	}
+	arg->base.key = xstrdup(key);
 	arg->value = xstrdup(value);
 	wl_list_append(&action->args, &arg->base.link);
 }
 
 static void
-action_arg_add_bool(struct action *action, char *key, bool value)
+action_arg_add_bool(struct action *action, const char *key, bool value)
 {
+	assert(action);
+	assert(key);
 	struct action_arg_bool *arg = znew(*arg);
 	arg->base.type = LAB_ACTION_ARG_BOOL;
-	if (key) {
-		arg->base.key = xstrdup(key);
-	}
+	arg->base.key = xstrdup(key);
 	arg->value = value;
 	wl_list_append(&action->args, &arg->base.link);
 }
 
 static void
-action_arg_add_int(struct action *action, char *key, int value)
+action_arg_add_int(struct action *action, const char *key, int value)
 {
+	assert(action);
+	assert(key);
 	struct action_arg_int *arg = znew(*arg);
 	arg->base.type = LAB_ACTION_ARG_INT;
-	if (key) {
-		arg->base.key = xstrdup(key);
-	}
+	arg->base.key = xstrdup(key);
 	arg->value = value;
 	wl_list_append(&action->args, &arg->base.link);
 }
@@ -252,12 +252,10 @@ action_str_from_arg(struct action_arg *arg)
 static const char *
 get_arg_value_str(struct action *action, const char *key, const char *default_value)
 {
+	assert(action);
 	assert(key);
 	struct action_arg *arg;
 	wl_list_for_each(arg, &action->args, link) {
-		if (!arg->key) {
-			continue;
-		}
 		if (!strcasecmp(key, arg->key)) {
 			return action_str_from_arg(arg);
 		}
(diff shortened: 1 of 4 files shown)


Consolatis:
The `enforce action arg->key` commit is pretty verbose as it required additional changes to the default bindings.

Once I verified that the rewrite of `action_arg_from_xml_node()`  properly defines all action and argument combinations I will split the first two commits off into a separate PR to ease reviewing.

Consolatis:
I could split the first commit off into another PR to ease reviewing if it makes live easier.
Otherwise this seems about ready to review.

src/action.c
```diff
@@ -547,40 +588,32 @@ actions_run(struct view *activator, struct server *server,
...
-				if (target) {
-					view_move_to_workspace(view, target);
-					if (follow) {
-						workspaces_switch_to(target);
-					}
+			if (!view) {
+				break;
+			}
+			const char *to = get_arg_value_str(action, "to", NULL);
+			bool follow = get_arg_value_bool(action, "follow", true);
```
Consolatis, line 605: Not sure if we should add an `assert(to)` here due to the default value of `get_arg_value_str()`. Its safe to do as the action would otherwise be deleted in the first place but we don't do it for the other required arguments either for the same reason.

src/action.c
```diff
@@ -293,6 +306,48 @@ actions_contain_toggle_keybinds(struct wl_list *action_list)
 	return false;
 }
 
+/* Checks for *required* arguments */
+bool
+action_is_valid(struct action *action)
+{
+	char *arg_name = NULL;
+	switch (action->type) {
```
johanmalm, line 314: We could just create a separate `get_arg_name()` function here like with:

```
case ACTION_TYPE_EXECUTE:
    return "command";
...
```

src/action.c
```diff
@@ -293,6 +306,48 @@ actions_contain_toggle_keybinds(struct wl_list *action_list)
...
+	case ACTION_TYPE_EXECUTE:
+		arg_name = xstrdup("command");
```
johanmalm, line 316: There is no need to heap-allocate here.
Within a frame I think it's simpler to just do `arg_name="command";` or whatever - or use a helper like `get_arg_name()` as suggested above.

src/action.c
```diff
@@ -348,23 +348,29 @@ action_is_valid(struct action *action)
...
+void
+action_free(struct action *action)
 {
+	/* Free args */
 	struct action_arg *arg, *arg_tmp;
+	wl_list_for_each_safe(arg, arg_tmp, &action->args, link) {
+		wl_list_remove(&arg->link);
+		zfree(arg->key);
+		if (arg->type == LAB_ACTION_ARG_STR) {
+			free((void *)action_str_from_arg(arg));
```
johanmalm, line 360: Is this right? Looks odd.
Shouldn't it be `free(arg->value);`

src/action.c
johanmalm: > Or maybe just add a comment here to say that its always available due to the config time check and potential removal of the action?

Yes, good idea. Agree.

johanmalm:
Looks good other than comments above.

src/action.c
Consolatis: This PR just refactored that bit. I agree though that we could cast to `struct action_arg_str` and then free that value instead of "unconst" the return value from `action_str_from_arg()`.

src/action.c
Consolatis: The issue with a generalized function like that is that there may be multiple arguments per action.
In this specific case we only care for the required arguments which is at most one per action right now but that may change in the future. And as its only required in this new validation function I didn't think its worth adding another function for it.

src/action.c
Consolatis: Will look into it when I'm back.

src/action.c
```diff
@@ -356,23 +364,75 @@ actions_contain_toggle_keybinds(struct wl_list *action_list)
...
+	wl_list_for_each_safe(arg, arg_tmp, &action->args, link) {
+		wl_list_remove(&arg->link);
+		zfree(arg->key);
+		if (arg->type == LAB_ACTION_ARG_STR) {
+			/*
+			 * struct action_arg_str *str_arg = (struct action_arg_str *)arg;
+			 * zfree(str_arg->value);
+			 */
+			zfree(((struct action_arg_str *)arg)->value);
+		}
```
Consolatis, line 423: @johanmalm which of the versions do you prefer, the commented or the other one?
I actually slightly prefer the commented 2 line variant as its easier to read.

src/action.c
Consolatis: Added two variants and opened another review comment to select one of the variants. Resolving this one.

src/action.c
johanmalm: Yes, agree.
I prefer the commented out version too.

Consolatis: done

src/action.c
Consolatis: added: 455c6fc3095597adc34a3c47c42676ac31b4fdad

johanmalm:
LGTM

Consolatis:
Alright, will squash and merge then. Thanks for your review.

PR #921, opened 2023-05-13
includes: identifier consistency in include guards

domo141:
Made all header files to have LABWC_ prefix in include guard identifers.

Converted from __LABWC_ in 35 include/ files.
Converted from __LAB_ in 5 include/ files.
Added LABWC prefix to 3 include/ files.
Added include guards to 3 include/ files.

### DIFF
--- a/include/common/graphic-helpers.h
+++ b/include/common/graphic-helpers.h
@@ -1,4 +1,6 @@
 /* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_GRAPHIC_HELPERS_H
+#define LABWC_GRAPHIC_HELPERS_H
 
 #include <cairo.h>
 #include <wayland-server-core.h>
@@ -44,3 +46,5 @@ void set_cairo_color(cairo_t *cairo, float *color);
 /* Draws a border with a specified line width */
 void draw_cairo_border(cairo_t *cairo, double width, double height,
 		double line_width);
+
+#endif /* LABWC_GRAPHIC_HELPERS_H */
--- a/include/action.h
+++ b/include/action.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: GPL-2.0-only */
-#ifndef __LABWC_ACTION_H
-#define __LABWC_ACTION_H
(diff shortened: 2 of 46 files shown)


include/common/graphic-helpers.h
```diff
@@ -1,4 +1,6 @@
 /* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_COMMON_GRAPHIC_HELPERS_H
+#define LABWC_COMMON_GRAPHIC_HELPERS_H
```
Consolatis, line 3: First of all: thanks for working on this.

The newly added include guards are slightly inconsistent with the other ones, the existing ones don't include the `COMMON` part whereas the new ones do. Personally I'd slightly prefer including the `COMMON` part but then we should do it consistently, e.g. the same for all subdirectories of `include/`. Or we leave them out everywhere and accept the potential (although minimal) risk of include guard clashes in the future.

Thoughts @johanmalm?

johanmalm: I think we should leave them out (i.e. just base it on filename) and sort it case-by-case in future if there is an identifier clash.

johanmalm:
Agree. Looks good.

Suggest adding a sentence in the commit message about the C standard reserving underscore + capital letter at start of identifier as a justification for removing the double underscores.

domo141:
FYI: also dropped COMMON_ parts that were in scaled_font_buffer.h and scaled_scene_buffer.h (i.e. not added by me in those 3 files that did not have include guards)

domo141:
Compare seems not to be smart enough to notice difference in commit message -- added that "justification" there.

Consolatis:
Thank you for the update. Looks good to me.
I'll leave it to @johanmalm to have a final look and accept the PR.

> FYI: also dropped COMMON_ parts that were in scaled_font_buffer.h and scaled_scene_buffer.h (i.e. not added by me in those 3 files that did not have include guards)

I guess somebody was using personal preferences here instead of following the existing pattern in other headers. I wonder who that was. :)

johanmalm:
Thanks. Great stuff.

PR #923, opened 2023-05-15
src/action.c: validate action type when assigning arguments

Consolatis:
Before this patch it was possible to assign an argument with an
existing name to an action that didn't support the given argument.
An example of this is using `direction` for `GoToDesktop`.

This patch now only creates action arguments that are actually
defined for the given action type and logs an error for
unsupported arguments.

The commit also makes sure to always supply the argument name.
This will reduce required checks in other parts of the codebase
in future commits.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -11,6 +11,7 @@
 #include "common/mem.h"
 #include "common/parse-bool.h"
 #include "common/spawn.h"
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM-DNT

PR #933, opened 2023-05-20
Add window rules and actions for `pcmanfm-qt --desktop`

johanmalm:
Add
- ~~ToggleAlwaysBelow~~ `ToggleAlwaysOnBottom`
- `skipTaskar`
- `skipWindowSwitcher` (which isn't really needed for this because window-switcher doesn't show views in always-below-tree, but it seems more complete to include it here anyway in case we change the window-switcher behavior later)

Try it with (--EDITED--):

```
  <windowRules>
    <windowRule title="pcmanfm-desktop*" skipTaskbar="yes" skipWindowSwitcher="yes">
      <action name="ToggleAlwaysOnBottom"/>
    </windowRule>
  </windowRules>
```

This example is not output-centered like a layer-shell implementation would be, but rather works across the whole work area.

In `pcmanfm-qt` desktop-preferences the following settings are useful:
- `Background` - individual wallpapers for each monitor. This looks better if you have outputs of different size.
- `General` - margins of work area (use this to allow for layer-shell panels and the like)

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -80,6 +80,13 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="ToggleAlwaysOnTop">*
 	Toggle always-on-top of focused window.
 
+*<action name="ToggleAlwaysOnBottom">*
+	Toggle bewteen layers 'always-on-bottom' and 'normal'. When a window is
+	in the 'always-on-bottom' layer, it is rendered below all other
+	top-level windows. It is anticipated that this action will be useful
+	when defining window-rules for desktop-management tools that do not
+	support the wlr-layer-shell protocol.
+
 *<action name="ToggleKeybinds">*
 	Stop handling keybinds other than ToggleKeybinds itself.
 	This can be used to allow A-Tab and similar keybinds to be delivered
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -430,6 +430,15 @@ situation.
 	*serverDecoration* over-rules any other setting for server-side window
 	decoration on first map.
 
+*<windowRules><windowRule skipTaskbar="">* [yes|no|default]
+	*skipTaskbar* removes window foreign-toplevel protocol handle so that
+	it does not appear in clients such as panels and taskbars using that
+	protocol.
+
+*<windowRules><windowRule skipWindowSwitcher="">* [yes|no|default]
+	*skipWindowSwitcher* removes window from the Window Switcher (alt-tab
+	on-screen-display)
+
 ## ENVIRONMENT VARIABLES
 
 *XCURSOR_THEME* and *XCURSOR_SIZE* are supported to set cursor theme
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -263,6 +263,7 @@ struct server {
 
 	/* Tree for all non-layer xdg/xwayland-shell surfaces with always-on-top/below */
 	struct wlr_scene_tree *view_tree_always_on_top;
+	struct wlr_scene_tree *view_tree_always_on_bottom;
 #if HAVE_XWAYLAND
 	/* Tree for unmanaged xsurfaces without initialized view (usually popups) */
 	struct wlr_scene_tree *unmanaged_tree;
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -254,7 +254,14 @@ struct server {
 
 	/* Tree for all non-layer xdg/xwayland-shell surfaces */
 	struct wlr_scene_tree *view_tree;
-	/* Tree for all non-layer xdg/xwayland-shell surfaces with always-on-top */
+
+	/*
+	 * Popups need to be rendered above always-on-top views, so we reparent
+	 * them to this dedicated tree
+	 */
+	struct wlr_scene_tree *xdg_popup_tree;
+
+	/* Tree for all non-layer xdg/xwayland-shell surfaces with always-on-top/below */
 	struct wlr_scene_tree *view_tree_always_on_top;
 #if HAVE_XWAYLAND
 	/* Tree for unmanaged xsurfaces without initialized view (usually popups) */
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -34,6 +34,22 @@ view_impl_map(struct view *view)
 	if (!view->been_mapped) {
 		window_rules_apply(view, LAB_WINDOW_RULE_EVENT_ON_FIRST_MAP);
 	}
+
+	/*
+	 * It's tempting to just never create the foreign-toplevel handle in the
+	 * map handlers, but the app_id/title might not have been set at that
+	 * point, so it's safer to process the property here
+	 */
+	enum property ret = window_rules_get_property(view, "skipTaskbar");
+	if (ret == LAB_PROP_TRUE) {
+		if (view->toplevel.handle) {
+			wlr_foreign_toplevel_handle_v1_destroy(view->toplevel.handle);
+		}
+	}
+
+	wlr_log(WLR_DEBUG, "[map] identifier=%s, title=%s\n",
+		view_get_string_prop(view, "app_id"),
+		view_get_string_prop(view, "title"));
 }
 
 static bool
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -27,13 +27,13 @@ view_impl_move_to_back(struct view *view)
 void
 view_impl_map(struct view *view)
 {
-	if (!view->been_mapped) {
-		window_rules_apply(view, LAB_WINDOW_RULE_EVENT_ON_FIRST_MAP);
-	}
 	desktop_focus_and_activate_view(&view->server->seat, view);
 	view_move_to_front(view);
 	view_update_title(view);
 	view_update_app_id(view);
+	if (!view->been_mapped) {
+		window_rules_apply(view, LAB_WINDOW_RULE_EVENT_ON_FIRST_MAP);
+	}
 }
 
 static bool
--- a/include/view.h
+++ b/include/view.h
@@ -171,8 +171,11 @@ void view_maximize(struct view *view, bool maximize,
 void view_set_fullscreen(struct view *view, bool fullscreen);
 void view_toggle_maximize(struct view *view);
 void view_toggle_decorations(struct view *view);
-void view_toggle_always_on_top(struct view *view);
+
 bool view_is_always_on_top(struct view *view);
+void view_toggle_always_on_top(struct view *view);
+void view_toggle_always_on_bottom(struct view *view);
+
 bool view_is_tiled(struct view *view);
 bool view_is_floating(struct view *view);
 void view_move_to_workspace(struct view *view, struct workspace *workspace);
--- a/include/window-rules.h
+++ b/include/window-rules.h
@@ -26,6 +26,8 @@ struct window_rule {
 	struct wl_list actions;
 
 	enum property server_decoration;
+	enum property skip_taskbar;
+	enum property skip_window_switcher;
 
 	struct wl_list link; /* struct rcxml.window_rules */
 };
--- a/src/action.c
+++ b/src/action.c
@@ -61,6 +61,7 @@ enum action_type {
 	ACTION_TYPE_TOGGLE_FULLSCREEN,
 	ACTION_TYPE_TOGGLE_DECORATIONS,
 	ACTION_TYPE_TOGGLE_ALWAYS_ON_TOP,
+	ACTION_TYPE_TOGGLE_ALWAYS_ON_BOTTOM,
 	ACTION_TYPE_FOCUS,
 	ACTION_TYPE_ICONIFY,
 	ACTION_TYPE_MOVE,
@@ -93,6 +94,7 @@ const char *action_names[] = {
 	"ToggleFullscreen",
 	"ToggleDecorations",
 	"ToggleAlwaysOnTop",
+	"ToggleAlwaysOnBottom",
 	"Focus",
 	"Iconify",
 	"Move",
@@ -513,6 +515,11 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_TOGGLE_ALWAYS_ON_TOP:
 			if (view) {
 				view_toggle_always_on_top(view);
+			}
+			break;
+		case ACTION_TYPE_TOGGLE_ALWAYS_ON_BOTTOM:
+			if (view) {
+				view_toggle_always_on_bottom(view);
 			}
 			break;
 		case ACTION_TYPE_FOCUS:
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -110,6 +110,10 @@ fill_window_rule(char *nodename, char *content)
 	/* Properties */
 	} else if (!strcasecmp(nodename, "serverDecoration")) {
 		set_property(content, &current_window_rule->server_decoration);
+	} else if (!strcasecmp(nodename, "skipTaskbar")) {
+		set_property(content, &current_window_rule->skip_taskbar);
+	} else if (!strcasecmp(nodename, "skipWindowSwitcher")) {
+		set_property(content, &current_window_rule->skip_window_switcher);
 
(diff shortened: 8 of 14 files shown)


johanmalm:
[recording.webm](https://github.com/labwc/labwc/assets/6955353/8b4be3a1-1014-4140-a335-0911f6737cf6)

docs/labwc-actions.5.scd
```diff
@@ -80,6 +80,9 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="ToggleAlwaysOnTop">*
 	Toggle always-on-top of focused window.
 
+*<action name="ToggleAlwaysBelow">*
+	Toggle always-below of focused window.
```
Consolatis, line 84: Outside of the things mentioned below I don't really see any use-case for the action. Maybe some examples would be good here?
Something like "desktop" implementations or components (like resource monitors or "widgets") that don't support the wlr-layershell protocol. Or mentioning that the main use-case is for use in windowRules for those desktop components.

Not sure if it would make it more confusing though so maybe its fine as it is right now.

docs/labwc-config.5.scd
```diff
@@ -430,6 +430,12 @@ situation.
 	*serverDecoration* over-rules any other setting for server-side window
 	decoration on first map.
 
+*<windowRules><windowRule skipTaskbar="">* [yes|no|default]
+	*skipTaskbar* do not create a foreign-toplevel handle
+
+*<windowRules><windowRule skipTaskbar="">* [yes|no|default]
+	*skipWindowSwitcher* do not appear in Window Switcher (alt-tab OSD)
+
```
Consolatis, line 438: Two times `skipTaskar`

include/labwc.h
```diff
@@ -254,8 +254,16 @@ struct server {
 
 	/* Tree for all non-layer xdg/xwayland-shell surfaces */
 	struct wlr_scene_tree *view_tree;
-	/* Tree for all non-layer xdg/xwayland-shell surfaces with always-on-top */
+
+	/*
+	 * Popups need to be rendered above always-on-top views, so we reparent
+	 * them to this dedicated tree
+	 */
+	struct wlr_scene_tree *popup_tree;
```
Consolatis, line 262: Should we name this `server->xdg_popup_tree`?

stefonarch:
Testing this "lxqt" branch, but somehow I get nothing of the 3 to work: waybar and switcher show still `pcmanfm--desktop0` and clicking on  the desktop moves it in front ("closing" smaller windows. Am I missing something?
```
  <windowRules>
      <windowRule identifier="lxqt-panel">
        <action name="MoveToEdge" direction="up" />
      </windowRule>

    <windowRule title="pcmanfm-desktop*" skipTaskbar="yes" skipWindowSwitcher="yes">
      <action name="ToggleAlwaysBelow"/>
      <action name="Maximize"/>
    </windowRule>


  </windowRules>

```
The lxqt branch is `0.4.0.770.g1424542` against` 0.6.1.60.g93aff4e` the master branch.

johanmalm:
> Testing this "lxqt" branch, but somehow I get nothing of the 3 to work: waybar and switcher show still `pcmanfm--desktop0` and clicking on the desktop moves it in front "closing" smaller windows. Am I missing something?
> 

Weird. Not sure why.

You've got two hyphens in the title (`pcmanfm--desktop0`). Is that a typo? If not that might be why the window rule doesn't pick it up. Guess you could try `pcmanfm*desktop*`.

If that doesn't work, try applying this patch to see what the app_id + title actually are on map.

```
diff --git a/src/view-impl-common.c b/src/view-impl-common.c
index 32f1e39..6f17813 100644
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -46,6 +46,9 @@ view_impl_map(struct view *view)
 			wlr_foreign_toplevel_handle_v1_destroy(view->toplevel.handle);
 		}
 	}
+	fprintf(stderr, "XXX app_id=%s, title=%s\n",
+			view_get_string_prop(view, "app_id"),
+			view_get_string_prop(view, "title"));
 }
 
 static bool
```

stefonarch:
>You've got two hyphens in the title

 that was a typo, the rule (c&p up here) should be ok. 

>...to see what the app_id + title actually are on map.

`XXX app_id=pcmanfm-qt, title=pcmanfm-desktop0
`
But app-id and title should be ok too, they were always like that on every compositor, and skip rules do not work for `lxqt-panel` too while `MovetoEdge` does. 

Install is correct?
```

git clone https://github.com/johanmalm/labwc.git
cd labwc
git checkout lxqt
meson setup build
ninja -C build install
```

johanmalm:
> Install is correct?
> 
> ```
> 
> git clone https://github.com/johanmalm/labwc.git
> cd labwc
> git checkout lxqt
> meson setup build
> ninja -C build install
> ```

Try running it from the build directory. That rules out that it's the wrong binary that you're running.
(if you run `type labwc` it tells you which binary it would run - but it's simpler to just run the build dir).

```bash
git clone --branch=lxqt https://github.com/johanmalm/labwc.git
cd labwc
meson setup build
meson compile -C build
./build/labwc -s 'pcmanfm-qt --desktop'
```

stefonarch:
Okeey, it was my startup script. I didn't test labwc for some time and forgot that it started it directly from the build folder without installing it and not from `/usr/local/bin` - that was a workaround months ago for some wlroots conflict. Anyway, everything works perfectly, even 
<windowRule identifier="lxqt-panel" skipTaskbar="yes" skipWindowSwitcher="yes">

johanmalm:
> Okeey, it was my startup script. I didn't test labwc for some time and forgot that it started it directly from the build folder without installing it and not from `/usr/local/bin` - that was a workaround months ago for some wlroots conflict. Anyway, everything works perfectly, even

Hey - we must have typed at the same time :smile:
...same conclusion.

stefonarch:
Yeah, we were.  Is it possible to set layers e/o roles in theory with labwc? Anyway, for pcmanfm this approach works fine, thanks!

Consolatis:
> "maximize" isn't needed IMO.

It might be required if you have another panel that uses the proper layershell protocol, it basically should make sure that it only uses the usable space that is not exclusively used by a panel. If pcmanfm-qt request maximized state itself that should also be fine but request the output dimensions would cause issues.

The same will be true once some `Margin` property is implemented as it will work basically the same way as having a panel with exclusive zone set.

johanmalm:
> "maximize" isn't needed IMO.

Well, if you have a panel in the bottom layer (which I do) then the desktop hides it.

There is another advantage of `maximize` which is that it takes into account the usable-area (i.e. the area not taken up by exclusive layer-shell clients). For this to be beneficial with pcmanfm-qt we'd have to explore changing its source code to work with the window size rather than the screen size.

We can make it work without `maximize`, but it relies on the user setting the pcmanfm-qt margins. Say for example that the user open a new panel (like waybar)... it would be nice if the pcmanfm-qt desktop icons just automatically rearranged themselves.

I can have a look at the pcmanfm-qt code to see if I can figure something out - but will have to be another day.

stefonarch:
Sorry, maximize for the desktop I meant.

johanmalm:
> The same will be true once some `Margin` property is implemented as it will work basically the same way as having a panel with exclusive zone set.

Exactly. Without the `maximize` approach this will get increasingly hacky. Just building on what @Consolatis said - imagine if we have a top-aligned `lxqt-panel` and implement `<margin>` to make sure windows take into account the panel, then the `pcmanfm-qt --desktop` window would be moved vertically down, so the user would have to set the pcmanfm-qt bottom margin to allow for a panel at the top..... it'll be too confusing for people.

johanmalm:
It might be this line https://github.com/lxqt/pcmanfm-qt/blob/8a7e4b6d6b37e5db619356a3bacd476e3683312c/pcmanfm/desktopwindow.cpp#LL1332 ... but could lead to regressions

src/view-impl-common.c
```diff
@@ -27,13 +27,25 @@ view_impl_move_to_back(struct view *view)
...
+	 * It's tempting to just never create the foreign-toplevel handle in the
+	 * map handlers, but the app_id/title might not have been set at that
+	 * point, so it's safer to process the property here
+	 */
+	enum property ret = window_rules_get_property(view, "skipTaskbar");
+	if (ret == LAB_PROP_TRUE) {
+		if (view->toplevel.handle) {
+			wlr_foreign_toplevel_handle_v1_destroy(view->toplevel.handle);
+		}
+	}
```
Consolatis, line 48: I think this needs a `view->toplevel.handle = NULL;`, otherwise we will end up calling function on the destroyed object.

johanmalm:
@stefonarch 
Any idea how we avoid using a virtual desktop so that we get a `pcmanfm-qt --desktop` window for each screen?
https://github.com/lxqt/pcmanfm-qt/blob/8a7e4b6d6b37e5db619356a3bacd476e3683312c/pcmanfm/application.cpp#L460-L473

stefonarch:
No idea about that, you may ask @tsujan, I've never had any dual monitor setup.

Just noticed that 
```
      <windowRule identifier="lxqt-panel" skipTaskbar="yes" skipWindowSwitcher="yes">
        <action name="MoveToEdge" direction="up" />
        <action name="ToggleAlwaysOnTop"/>
      </windowRule>
```
makes the panel sticky also on all desktops (thought there is no such action atm). Probably also "region" can be used well for the notifications

tsujan:
> Any idea how we avoid using a virtual desktop so that we get a pcmanfm-qt --desktop window for each screen?

If by desktop window you mean the wallpaper, check this box:

desktop_preferences

Otherwise, the desktop window is treated like any other window, i.e., as a single window.

johanmalm:
> Otherwise, the desktop window is treated like any other window, i.e., as a single window.

@tsujan I'm not really pushing in any particular direction, I'm just experimenting and learning at the moment.

I'm more interested in the icons. Let me share an observation :smile:

If I apply the patch below and implement a rule which moves `pcmanfm-desktop1` (the second instance) to the co-ordinates of my second output I get the first screenshot below. We could easily do a per-title window-rule to send instances to a particular output. 

```
diff --git a/pcmanfm/application.cpp b/pcmanfm/application.cpp
index 1a034fc..bfbc86c 100644
--- a/pcmanfm/application.cpp
+++ b/pcmanfm/application.cpp
@@ -460,7 +460,8 @@ void Application::desktopManager(bool enabled) {
             // NOTE: there are two modes
             // When virtual desktop is used (all screens are combined to form a large virtual desktop),
             // we only create one DesktopWindow. Otherwise, we create one for each screen.
-            if(primaryScreen() && primaryScreen()->virtualSiblings().size() > 1) {
+	    if (0) {
+//            if(primaryScreen() && primaryScreen()->virtualSiblings().size() > 1) {
                 DesktopWindow* window = createDesktopWindow(-1);
                 desktopWindows_.push_back(window);
             }
@@ -468,6 +469,7 @@ void Application::desktopManager(bool enabled) {
                 int n = qMax(allScreens.size(), 1);
                 desktopWindows_.reserve(n);
                 for(int i = 0; i < n; ++i) {
+		    qDebug("createDesktopWindow()");
                     DesktopWindow* window = createDesktopWindow(i);
                     desktopWindows_.push_back(window);
                 }
```

multiple-instances


If running `pcmanfm-qt --desktop` with no patches applied, I get this:

one-instance

So, I'm just thinking that if `pcmanfm-qt --desktop` worked off the window geometry - rather than the screen geometry - for things like icon positions, then we have a pretty flexible solution.

I'm not really considering any regressions, etc. (just playing)

Also, please bear in mind that I've spent a lot of time thinking about layer-shell implementations which are output based (not designed to have one big surface spanning multiple outputs), so my head is rightly or wrongly stuck in that space at the moment.

tsujan:
> I'm more interested in the icons.

People use multiple screens to *extend* their work area.

Duplicating items inside the same widget would be a bug. Adding multiple widgets for multiple screens and making multiple copies of each item would be counter-intuitive — to say nothing of the complex code, which also contradicts lightweightness.

BTW, If `layer-shell-qt`becomes what is advertised at https://invent.kde.org/plasma/layer-shell-qt/-/merge_requests/25, the desktop of the Qt6 version of pcmanfm-qt should logically be able to put itself on the background layer, anchored to the four screen edges.

johanmalm:
> > I'm more interested in the icons.
> 
> People use multiple screens to _extend_ their work area.
> 
> Duplicating items inside the same widget would be a bug. Adding multiple widgets for multiple screens and making multiple copies of each item would be counter-intuitive — to say nothing of the complex code, which also contradicts lightweightness.
> 
> BTW, If `layer-shell-qt`becomes what is advertised at https://invent.kde.org/plasma/layer-shell-qt/-/merge_requests/25, the desktop of the Qt6 version of pcmanfm-qt should logically be able to put itself on the background layer, anchored to the four screen edges.

Understood.

Yes, the `layer-shell-qt` MR looks promising.

tsujan:
> Yes, the layer-shell-qt MR looks promising.

I hope it'll work with all wlroots-based compositors. Its Qt5 version "worked" with them (I'd tested it) but terribly ;) because it put the whole app on a layer —  its menus and tooltips too! Perhaps the reason was Qt5 itself.

src/view-impl-common.c
johanmalm: It'd done in the destroy-handler

https://github.com/labwc/labwc/blob/f2c3f83041fce356165bd4124fd4f388bf06570d/src/foreign.c#L62

Consolatis: good point.

docs/labwc-actions.5.scd
johanmalm: Done.

docs/labwc-config.5.scd

include/labwc.h
johanmalm: Done

johanmalm:
All review comments processed.

Suggest dealing with 'sticky' in a separate PR.

Original Post updated. Note:
- `s/ToggleAlwaysBelow/ToggleAlwaysOnBottom/`
- I think this will work best without `Maximize` and just let `pcmanfm-qt --desktop` use the whole work area.
-

tsujan:
> I think this will work best without Maximize and just let pcmanfm-qt --desktop use the whole work area.

Right, there's no need to maximization.

I'm not familiar with the code of labwc but, months ago, I did it for the stable version of Wayfire by making personal patches for (1) setting the layer, (2) setting stickiness, and (3) preventing the desktop from being raised when focused (by clicking).

Consolatis:
LGTM, did not have time to test.

johanmalm:
Thanks all for the help. Another step forward :smile:

stefonarch:
Thanks to you! Using now labwc as daily driver.
Updated https://github.com/stefonarch/LXQt-Wayland-files#labwc-stacking

Wondering if  there is some irc or matrix or whatever channel to discuss, ask questions on the fly etc?

stefonarch:
Only noticed now: copy|move popup and file tooltips info do not appear on the desktop (moving a file into a folder for example should be like that:
screen_area_mer_17:32:47_

johanmalm:
> Only noticed now: copy|move popup ~and file tooltips info~ do not appear on the desktop (moving a file into a folder for example should be like that)
> It's impossible without using a shortcut (ctrl or shift) to move files into a folder.
> Tooltips appear when clicking the empty desktop first.

Those drag-and-drop popups work for me when dragging to `pcmanfm-qt --desktop` from `nautilus`, `thunar` and `pcmanfm-qt` - but interestingly it does not work from `pcmanfm` (without `-qt`).

It looks like it's not related to the desktop implementation or our window-rules because I can't drag-and-drop between `pcmanfm-qt` and `pcmanfm` either (both without `--desktop`).

Which application did you drag from?

johanmalm:
Oh - I see. Got it. That's weird!

johanmalm:
> Only noticed now: copy|move popup ~and file tooltips info~ do not appear on the desktop (moving a file into a folder for example should be like that): screen_area_mer_17:32:47_
> 
> It's impossible without using a shortcut (ctrl or shift) to move files into a folder.
> 
> Tooltips appear when clicking the empty desktop first.

Have raised https://github.com/labwc/labwc/issues/939 so that we don't lose this one.

PR #934, opened 2023-05-21
action: support {Go,Send}ToDesktop 'wrap' option

domo141:
Make wrap 'true' by default for both GoToDesktop and SendToDesktop, in order to default the current behaviour.

For GoToDesktop OpenBox defaults to wrap="yes", but I don't know whether OpenBox SendToDesktop does the same.

Note: this is an interface change.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -91,20 +91,26 @@ Actions are used in menus and keyboard/mouse bindings.
 	to the center of the window. If the given output does not contain
 	any windows, the cursor is centered on the given output.
 
-*<action name="GoToDesktop" to="value" />*
+*<action name="GoToDesktop" to="value" wrap="yes" />*
 	Switch to workspace.
 
 	*to* The workspace to switch to. Supported values are "last", "left",
 	"right" or the full name of a workspace or its index (starting at 1)
 	as configured in rc.xml.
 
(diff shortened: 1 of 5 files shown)


johanmalm:
Super effort.

Code looks good and it works as expected.

Would you mind modifying the commit message before we merge based on feedback below:

1. Openbox defaults to wrap="yes" for both the `Go` and `Send` variants. I think we should therefore remove: `For GoToDesktop OpenBox defaults to wrap="yes", but I don't know whether OpenBox SendToDesktop does the same.`

    http://openbox.org/wiki/Help:Actions#GoToDesktop
    http://openbox.org/wiki/Help:Actions#GoToDesktop

2. I think your statement `Make wrap 'true' by default for both GoToDesktop and SendToDesktop, in order to default to the current behaviour` is perfect. We could add `and to stay consistent with Openbox behaviour` if you want.

3. I suggest removing `Note: this is an interface change.` because it isn't  :smile: unless we make `wrap` a mandatory attribute and that would be unnecessary. I argue that an interface change is when a user's current config won't work or will result in a different behaviour. All we have done here is to add a config option to 'disable' wrap.

I assume - based on your coding skills - that you're familiar with the process, but adding the lines below just in case.

```
git commit --amend
git push -f origin todesk-wrap
```

domo141:
pushed commit a6c4353

Consolatis:
Thanks :)

Issue #939, opened 2023-05-27
Set surface on completion of dnd


johanmalm:
Continuation of: https://github.com/labwc/labwc/pull/933#issuecomment-1561388921

Relating to `pcmanfm-qt --desktop`

> copy|move popup ~and file tooltips info~ do not appear on the desktop (moving a file into a folder for example should be like that):
![screen_area_mer_17:32:47_](https://github.com/labwc/labwc/assets/10681413/546be158-efa0-4037-8fd3-78d5e60de1f1)

PR #940, opened 2023-05-27
dnd: refocus surface under cursor on drag-destroy

johanmalm:
Keyboard focus is not changed during drag, so we need to refocus the surface under the cursor and the end of a drag-and-drop operation.

This fixes the issue described in #939, but I'd appreciate if anyone has time to do some testing to make sure it doesn't break something else :smile:

### DIFF
--- a/include/cursor.h
+++ b/include/cursor.h
@@ -2,6 +2,7 @@
 #ifndef LABWC_CURSOR_H
 #define LABWC_CURSOR_H
 
+#include <wlr/types/wlr_cursor.h>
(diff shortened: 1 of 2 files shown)


stefonarch:
A first test is 100% positive, but I'll invest some more time now for testing.

stefonarch:
I tried every possible action and didn't found any issue (except that after lowering all windows the desktop doesn't get focus automatically to show the tooltips on hovering items, but that isn't related to this).

johanmalm:
Excellent. Let's go for it then.

Issue #941, opened 2023-05-27
lxqt


johanmalm:
Things to fix in general:

- [x] Window-rule property ~~`skipInteractiveMoveResize`~~ `fixedPosition`
- [x] #940 
- ~~Remove 60px hard-coded min-height~~
- [x] Hide always-on-top layer when running an application in full-screen. See conversation here: https://github.com/labwc/labwc/pull/947#issuecomment-1606256482

In support of `lxqt-panel`
- [x] `MoveTo` action (#947)
- [x] `<margin>` (#947)
- [x] `matchOnce` property to only apply rules to panel rather than configuration windows with the same app_id (#947)

In support of `lxqt-notificationd`
- ~~Window-rule property `allowFocusRequest` in support of `<windowRule identifier="lxqt-notificationd" skipTaskbar="yes" skipWindowSwitcher="yes" allowFocusRequst="no">`~~

stefonarch:
For notifications some kind of  ` <windowRule identifier="lxqt-notificationd" skipTaskbar="yes" skipWindowSwitcher="yes" nofocus="yes">` rule would be  needed.

stefonarch:
> In support of `lxqt-panel`

Just started using gaps and the panel is affected by it, so a `gap="no"` maybe needed.

Consolatis:
That might somewhat complicate the codebase. Just thinking out loud, maybe we could instead add an optional argument to `MoveToEdge` which ignores the gap. I am not sure which variant is less disruptive.

johanmalm:
> > In support of `lxqt-panel`
> 
> Just started using gaps and the panel is affected by it, so a `gap="no"` maybe needed.

Yes, I had noticed the regions/gap interaction. Also, regions currently only take a % argument, so not quite right atm.
I envisaged a simple `MoveTo` (which doesn't take into account gaps) as suggested in Original Post. Thoughts?

stefonarch:
This sounds good. Using "regions" has also the downside  that it affects the config window.

stefonarch:
Some checkboxes can be ticked. The 60px rule could stay as it doesn't matter anymore, a bottom panel can be placed with `<action name="MoveTo" x="0" y="x"/>`  with x = screen height - panel height, although all popups open at screen border so it's anyway not really usable.

stefonarch:
>Hide always-on-top layer when running an application in full-screen.

[...]

PR #947, opened 2023-06-06
Add `<margin>` and `MoveTo` in support of `lxqt-panel`

johanmalm:
Try with

```
  <margin top="32" />

  <windowRules>
    <!-- set matchOnce to avoid applying rule to panel configuration window -->
    <windowRule identifier="lxqt-panel" matchOnce="true">
      <skipTaskbar>yes</skipTaskbar>
      <action name="MoveTo" x="0" y="0" />
      <action name="ToggleAlwaysOnTop"/>
    </windowRule>
    <windowRule title="pcmanfm-desktop*">
      <skipTaskbar>yes</skipTaskbar>
      <skipWindowSwitcher>yes</skipWindowSwitcher>
      <action name="MoveTo" x="0" y="0" />
      <action name="ToggleAlwaysOnBottom"/>
    </windowRule>
  </windowRules>

```

- ~~Hide always-on-top layer when running an application in full-screen~~ Moved to first post of #941

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -411,9 +411,9 @@ defined as shown below.
 </windowRules>
 ```
 
-*Actions*
+*Criteria*
 
-*<windowRules><windowRule identifier="" title="">*
+*<windowRules><windowRule identifier="" title="" matchOnce="">*
 	Define a window rule for any window which matches the criteria defined
 	by the attributes *identifier* or *title*. If both are defined, AND
 	logic is used, so both have to match.
@@ -424,6 +424,9 @@ defined as shown below.
 	for XWayland clients.
 
 	*title* is the title of the window.
+
+	*matchOnce* can be true|false. If true, the rule will only apply to the
+	first instance of the window with the specified identifier or title.
 
 *Properties*
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -235,6 +235,12 @@ Therefore, where multiple objects of the same kind are required (for example
 *<theme><font place=""><weight>*
 	Font weight (normal or bold). Default is normal.
 
+## MARGIN
+
+*<margin top="" bottom="" left="" right="" output="" />*
+	Specify the number of pixels to reserve at the edges of an output.
+	New, maximized and tiled windows will not be placed in these areas.
+
 ## KEYBOARD
 
 *<keyboard><keybind key="">*
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -89,6 +89,17 @@
     </names>
   </desktops>
 
+  <!--
+    # <margin> can be used to reserve space where new/maximized/tiled
+    # windows will not be placed. Clients using layer-shell protocol reserve
+    # space automatically, so <margin> is only intended for other, specialist
+    # cases.
+    #
+    # If output is left empty, the margin will be applied to all outputs.
+
+    <margin top="" bottom="" left="" right="" output="" />
+  -->
+
   <!-- Percent based regions based on output usable area, % char is required -->
   <!--
     <regions>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -387,20 +387,41 @@
   </libinput>
 
   <!--
-    Window Rules
-      - Criteria can consist of 'identifier' or 'title' or both (in which case
-        AND logic is used).
-      - 'identifier' relates to app_id for native Wayland windows and WM_CLASS
-        for XWayland clients.
-      - Matching against patterns with '*' (wildcard) and '?' (joker) is
-        supported. Pattern matching is case-insensitive.
+    # Window Rules
+    #   - Criteria can consist of 'identifier' or 'title' or both (in which case
+    #     AND logic is used).
+    #   - 'identifier' relates to app_id for native Wayland windows and WM_CLASS
+    #     for XWayland clients.
+    #   - Criteria can also contain `matchOnce="true"` meaning that the rule
+    #     must only apply to the first instance of the window with that
+    #     particular 'identifier' or 'title'.
+    #   - Matching against patterns with '*' (wildcard) and '?' (joker) is
+    #     supported. Pattern matching is case-insensitive.
 
-  <windowRules>
-    <windowRule identifier="*"><action name="Maximize"/></windowRule>
-    <windowRule identifier="foo" serverDecoration="yes"/>
-    <windowRule title="bar" serverDecoration="yes"/>
-    <windowRule identifier="baz" title="quax" serverDecoration="yes"/>
-  </windowRules>
+    <windowRules>
+      <windowRule identifier="*"><action name="Maximize"/></windowRule>
+      <windowRule identifier="foo" serverDecoration="yes"/>
+      <windowRule title="bar" serverDecoration="yes"/>
+      <windowRule identifier="baz" title="quax" serverDecoration="yes"/>
+    </windowRules>
+
+    # Example below for `lxqt-panel` and `pcmanfm-qt \-\-desktop`
+    # where 'matchOnce' is used to avoid applying rule to the panel configuration
+    # window with the same 'app_id'
+
+    <windowRules>
+      <windowRule identifier="lxqt-panel" matchOnce="true">
+        <skipTaskbar>yes</skipTaskbar>
+        <action name="MoveTo" x="0" y="0" />
+        <action name="ToggleAlwaysOnTop"/>
+      </windowRule>
+      <windowRule title="pcmanfm-desktop*">
+        <skipTaskbar>yes</skipTaskbar>
+        <skipWindowSwitcher>yes</skipWindowSwitcher>
+        <action name="MoveTo" x="0" y="0" />
+        <action name="ToggleAlwaysOnBottom"/>
+      </windowRule>
+    </windowRules>
   -->
 
 </labwc_config>
--- a/src/action.c
+++ b/src/action.c
@@ -74,6 +74,7 @@ enum action_type {
 	ACTION_TYPE_RAISE,
 	ACTION_TYPE_LOWER,
 	ACTION_TYPE_RESIZE,
+	ACTION_TYPE_MOVETO,
 	ACTION_TYPE_GO_TO_DESKTOP,
 	ACTION_TYPE_SEND_TO_DESKTOP,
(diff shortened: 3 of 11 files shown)


johanmalm:
@stefonarch Appreciate review/testing if you have time.

stefonarch:
Nice, `margin` works fine, I had a hack for the exclusive zone first. As you mentioned the config and confirmation windows are partially out of screen now (title bar) and all open at 0 0. Popups for volume etc are all fine.

Panel doesn't still hide in fullscreen views, but this should be solved by `toplevel` too afaik.

johanmalm:
[recording.webm](https://github.com/labwc/labwc/assets/6955353/b86dd922-8322-4614-865f-df5c8e996fcb)

Starting to look okay. Config in OP updated.

stefonarch:
By chance I saw the checks running and compiled: works  perfectly! Thanks!

stefonarch:
I noticed  the `MoveToLayer=' in "Scope", will this be implemented?

docs/labwc-config.5.scd
stefonarch, line 1: Wouldn't "margin" be better under <core> like `gap`?

stefonarch:
One issue is https://github.com/lxqt/lxqt-notificationd/issues/344

stefonarch:
FYI: `lxqt-notificationd` with rudimental wayland support. The menu (saved notrifications) is fullscreen and at 0 0 though.

stefonarch:
I've noticed  2 issues: 

* for some reason a left panel starts in the middle of the screen downwards, and if labwc is restarted with 2 panels they are crossed in the middle (=placement not working)
* the shortcut for "minimize" works also for the panel, sending it to nowhere visible, only way is to restart it from session settings→modules.

Using lxqt-panel built with layer-shell-qt (layerBottom) those issues don't exist (but it's not usable at all as it doesn't detach qt processes, opening them fullscreen on all workspaces like it's menus). An option "SendToLayer" would probably solve much.

stefonarch:
@johanmalm  Any news on this?

johanmalm:
> @johanmalm Any news on this?

Hi there. No. I'm keen to get a second pair of eyes (ideally from one of the other core devs) on the stuff in this PR. Some PRs quite quite low risk and easy to tidy-up afterwards if things go wrong, but I'd rather wait a bit to reduce the risk of realizing later that something was a stupid idea and then having to break backward compatibility by removing/changes a window rule.

> I noticed the `MoveToLayer=' in "Scope", will this be implemented?

If there is a good reason to. In the Openbox sense, this is the non-toggle equivalent to ToggleAlwaysOnTop, etc. The layer shell layers are quite different in a number of ways, for example they are output-specific. Creating a rule to cover xdg-shell clients to `unmanaged` as some kind of pseudo layer-shell clients is what I started in #787 but that is quite a different topic and one that fundamentally requires window rules anyway.

> Wouldn't "margin" be better under <core> like gap?

We try to stick to Openbox syntax where we can, and `<margin>` falls into that category. `<core>` non-Openbox stuff and particularly Wayland specific stuff.

> One issue is https://github.com/lxqt/lxqt-notificationd/issues/344

I've not even started thinking about `lxqt-notificationd` yet. I prefer the one-step-at-a-time approach and prefer to get this PR through.

> for some reason a left panel starts in the middle of the screen downwards,

I welcome help on this. I think the starting point here is to understand what the `lxqt-panel` does in response to its own configuration. A `MoveTo` action might be the most sensible/generic thing we can do here.

> and if labwc is restarted with 2 panels they are crossed in the middle (=placement not working)

Well :smile: Do you have any suggestions? We could try to create some more sophisticated window rules, but if they have the same app_id and title it's unlikely to be pretty (and we prefer it pretty).

> Using lxqt-panel built with layer-shell-qt (layerBottom) those issues don't exist (but it's not usable at all as it doesn't detach qt processes, opening them fullscreen on all workspaces like it's menus).

Yep, I know. Bring on qt 6.5!!

> An option "SendToLayer" would probably solve much.

See above. I don't think that the `xdg-unmanaged` solves anything really. It still requires window rules. The advantage is that we could use one big rule rather than lots of small ones, but it's likely to get more complicated than that. The disadvantage of course is that you end up with lots of parallel code paths (for example for treating xdg-popups) and the user would still need to change labwc configuration when panel config is changed, so all in all, it's unlikely to be a sensible answer.

docs/labwc-config.5.scd
```diff
@@ -405,9 +411,9 @@ defined as shown below.
 </windowRules>
 ```
 
-*Actions*
+*Criteria*
 
-*<windowRules><windowRule identifier="" title="">*
+*<windowRules><windowRule identifier="" title="" singleton="">*
```
Consolatis, line 416: `singleton` sounds very devel-centric. I am not sure what name to use instead though to make it understandable for non-developers. The feature itself sounds pretty useful.

src/action.c
```diff
@@ -557,6 +565,15 @@ actions_run(struct view *activator, struct server *server,
 					resize_edges);
 			}
 			break;
+		case ACTION_TYPE_MOVETO:
+			if (view) {
+				const char *x = get_arg_value_str(action, "x", NULL);
+				const char *y = get_arg_value_str(action, "y", NULL);
+				if (x && y) {
+					view_move(view, atoi(x), atoi(y));
+				}
```
Consolatis, line 574: I think we we should add them as `int` args while parsing instead of parsing the string values when executing.

docs/rc.xml.all
```diff
@@ -89,6 +89,17 @@
     </names>
   </desktops>
 
+  <!--
+    You can reserve portions of your screens where windows will not cover when
+    they are maximized/tiled/etc or when they are initially placed. Clients
```
Consolatis, line 94: You can reserve portions of your screens which will not be used by windows in maximized or tiled states.

docs/labwc-config.5.scd
johanmalm: True.

I couldn't think of anything better. We could just call it `unique` or `singleInstance` or something.

src/action.c
johanmalm: So, add a `get_arg_value_int()` function?

docs/rc.xml.all
johanmalm: Looking at the man page, I opted for "Specify the number of pixels to reserve at the edges of an output. New, maximized and tiled windows will not be placed in these areas." I'll patch it with something along those lines.

Consolatis: sounds good.

src/action.c
Consolatis: yes, sounds good

docs/labwc-config.5.scd
Consolatis: Just thinking out loud. `firstOnly` or `matchOnce`?

johanmalm: Genius. `matchOnce` is very good.

src/action.c
johanmalm: Done

docs/rc.xml.all
johanmalm: Done

johanmalm:
All done. Ready to go from my perspective.

Consolatis:
Just making sure it doesn't go unnoticed: top post has a unsolved TODO:
> Hide always-on-top layer when running an application in full-screen

Which I am actually not sure about: I understand the reasoning for it in the panel at always-on-top case.
But in other situations, always-on-top windows might be intended by the user to be on top of a fullscreen application, e.g. some small chat window or similar on top of a youtube video. Same for the notify daemon (assuming its using always-on-top as well).

stefonarch:
>>  for some reason a left panel starts in the middle of the screen downwards,
>
>I welcome help on this. I think the starting point here is to understand what the lxqt-panel does in response to its own configuration.

I found the issue just now, thinking about why on all other compositors a rule `position 0 0` works fine for 2 panels. With `singleton="false">`  the left panel is working fine, which has it's logic, as it is affecting also a second panel. Panels can also be restarted perfectly. So basically now there is a choice  between the config windows at 0 0 and up to 2 panels or just one panel with good config popups. I think that's acceptable for how things are atm, as there is nowhere a panel with a GUI for wayland I know about. I managed to add also [workspace switchers](https://github.com/stefonarch/LXQt-Wayland-files/blob/main/lxqt-panel.md#desktop-cyclerswitcher).

>>   An option "SendToLayer" would probably solve much.
>
> I don't think that the xdg-unmanaged solves anything really

I was thinking about "bottom" layer as in layer shell, the wayfire plugin solves most issues with assigning layers, but yes, on that branch the issue was the same.

>I've not even started thinking about lxqt-notificationd yet.

The only thing missing would be "nofocus" rule as regions can be already  set, but the wip branch of lxqt-notificationd will solve this anyway and also all settings and preview.

Anyway, thanks for your work! I'm preparing a `lxqt-wayland-sessions` repository which installs all necessary files for a LXQt Labwc session working out of the box.

stefonarch:
>Edit:
Maybe we need another property for this like hideOnFullscreen or similar?

This would solve it IMO.
>
>Edit 2:
Do we actually need to use always-on-top for the panel when we also configure the margin?

Always-on-top is also "sticky" or "pin" on all workspaces so  yes if there is more than one. But with only one workspace without that rule panel is hiding, tested just now. An option could be to create  "always on top" and "on all workspaces" rules.

>E.g. the same situation we have with a layershell bottom layer + exclusive zone.

Is that so? My waybar has it's exclusive zone and is on bottom layer and hides perfectly with fullscreen video or F11 windows.

As the margin rule affects also the desktop the lxqt-panel can have an opacity which is seen only with windows in fullscreen.

screen_area_dom_23:26:13_

Consolatis:
> > E.g. the same situation we have with a layershell bottom layer + exclusive zone.
> 
> Is that so? My waybar has it's exclusive zone and is on bottom layer and hides perfectly with fullscreen video or F11 windows.

That is what I meant. When using just the margin and no always-on-top setting for the panel it should result in similar behavior to layershell `bottom` + exclusive zone and thus should have no issues with being on top of fullscreen windows.



> Always-on-top is also "sticky" or "pin" on all workspaces so yes if there is more than one. But with only one workspace without that rule panel is hiding, tested just now. An option could be to create "always on top" and "on all workspaces" rules.

Good point. We really need to decouple these two settings.

johanmalm:
Thanks for review. I've moved the reminder about 'always on top' to the first post of #941 so that it's all in one place.

PR #954, opened 2023-06-14
fix typo

stefonarch:


stefonarch:
It just made me think if it could matter or not, was for consistency.

Issue #763, opened 2023-02-09
Make it possible to configure the number of Workspaces without naming them


Flrian:
To make it work more similar to [how it does in Openbox, at least according to the wiki?](http://openbox.org/wiki/Help:Configuration#Desktops)

```
<desktops>
  <number>4</number>
</desktops>
```

>number The number of virtual desktops to use.

>names Each name tag names your desktops, in ascending order. Unnamed desktops will be named automatically depending on the locale. You can name more desktops than specified in number if you want.

Very very very low priority, just thought it would be nice to have sometime in the future?

Consolatis:
The feature should be designed in a way that all of this variants work correctly, e.g. end up with "Workspace 1" to "Workspace 4":
```xml
<desktops number="4" />

<desktops>
  <number>4</number>
</desktops>

<desktops>
  <number>4</number>
  <names>
    <name>Workspace 1</name>
    <name>Workspace 2</name>
  </names>
</desktops>

<desktops number="4">
  <names>
    <name>Workspace 1</name>
    <name>Workspace 2</name>
  </names>
</desktops>
```

Sachin-Bhat:
Hello folks, I tried my hand at this issue. I am still fairly new to C. So, if possible could someone kindly direct me on building and testing my code. To give some context I am running Arch Linux with labwc. I just want to ensure that I don't brick what I currently have set up!

johanmalm:
To build:

```
meson setup build
ninja -C build/
```

Run with `./build/labwc`

Carry out some some basics checks on a file, by running: `./scripts/check --file=src/workspaces.c`

We don't have any unit tests, so rely on peer review and manually testing features.

In terms of sharing your patches, I suggest the following (assuming that you've cloned `https://github.com/Sachin-Bhat/labwc.git` and that you're working from the toplevel directory of that repo):

```
git checkout -b config-workspaces-by-number
# make some changes to src//workspaces.c etc
git add src/workspaces.c
git commit -m 'config: support <desktops number="">'
git push origin config-workspaces-by-number
```

Then, either go to the github website and create a draft pull request against the `labwc/labwc.git` master branch or just let us know that you've pushed some commits against a branch on your fork and we can give you some feedback.

You might have written the perfect patch-set, but don't worry if you haven't. Most people don't :smile: including me. Being able to discuss and take feedback is an important part of development.
[...]

PR #956, opened 2023-06-19
config: support <desktops number="">

Sachin-Bhat:
This are the set of changes that I have been working on with reference to #763. I would really appreciate some feedback and suggestions. Currently the documentation of the changes are still pending but I shall write them once the updated changes are finalised.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -196,7 +196,7 @@ Therefore, where multiple objects of the same kind are required (for example
 
 ## WORKSPACES
 
-*<desktops><names><name>*
+*<desktops number=""><names><name>*
 	Define workspaces. A workspace covers all outputs. The OSD only shows
 	windows on the current workspace. Workspaces can be switched to with
 	GoToDesktop and windows can be moved with SendToDesktop. See
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -202,6 +202,10 @@ Therefore, where multiple objects of the same kind are required (for example
 	GoToDesktop and windows can be moved with SendToDesktop. See
 	labwc-actions(5) for more information about their arguments.
 
+	The number attribute defines the minimum number of workspaces. Default is 1. 
+	The number attribute is optional. If the number attribute is specified, names.name
+	is not required.
+
 *<desktops><popupTime>*
 	Define the timeout after which to hide the workspace OSD.
 	A setting of 0 disables the OSD. Default is 1000 ms.
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -69,16 +69,17 @@
     See man labwc-actions for further information.
 
     Workspaces can be configured like this:
-    <desktops>
+    <desktops number="4">
       <popupTime>1000</popupTime>
       <names>
         <name>Workspace 1</name>
         <name>Workspace 2</name>
         <name>Workspace 3</name>
+        <name>Workspace 4</name>
       </names>
     </desktops>
   -->
-  <desktops>
+  <desktops number="4">
     <!--
       popupTime defaults to 1000 so could be left out.
       Set to 0 to completely disable the workspace OSD.
@@ -86,6 +87,9 @@
     <popupTime>1000</popupTime>
     <names>
       <name>Default</name>
+      <name>Browsing</name>
+      <name>Music</name>
+      <name>Chat</name>
     </names>
   </desktops>
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -69,7 +69,7 @@
     See man labwc-actions for further information.
 
     Workspaces can be configured like this:
-    <desktops number="4">
+    <desktops>
       <popupTime>1000</popupTime>
       <names>
         <name>Workspace 1</name>
@@ -78,20 +78,15 @@
         <name>Workspace 4</name>
       </names>
     </desktops>
+
+    Or it can also be configured like this:
+    <desktops number="4" />
   -->
-  <desktops number="4">
     <!--
       popupTime defaults to 1000 so could be left out.
       Set to 0 to completely disable the workspace OSD.
     -->
-    <popupTime>1000</popupTime>
-    <names>
-      <name>Default</name>
-      <name>Browsing</name>
-      <name>Music</name>
-      <name>Chat</name>
-    </names>
-  </desktops>
+
 
   <!-- Percent based regions based on output usable area, % char is required -->
   <!--
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -86,7 +86,12 @@
       popupTime defaults to 1000 so could be left out.
       Set to 0 to completely disable the workspace OSD.
     -->
-
+  <desktops>
+    <popupTime>1000</popupTime>
+    <names>
+      <name>Default</name>
+    </names>
+  </desktops>
 
   <!-- Percent based regions based on output usable area, % char is required -->
   <!--
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -69,7 +69,7 @@ struct rcxml {
 
 	struct {
 		int popuptime;
-		int count;
+		int min_nr_workspaces;
 		struct wl_list workspaces;  /* struct workspace.link */
 	} workspace_config;
 
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -69,6 +69,7 @@ struct rcxml {
 
 	struct {
 		int popuptime;
+		int count;
 		struct wl_list workspaces;  /* struct workspace.link */
 	} workspace_config;
 
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1051,11 +1051,11 @@ post_processing(void)
 	int nr_workspaces = wl_list_length(&rc.workspace_config.workspaces);
 	if (nr_workspaces < rc.workspace_config.min_nr_workspaces) {
 		struct workspace *workspace;
+		char workspace_name[32];
 		for (int i = nr_workspaces; i < rc.workspace_config.min_nr_workspaces; i++) {
 			workspace = znew(*workspace);
-			char *workspace_name = malloc(sizeof(char) * 32);
-			snprintf(workspace_name, 32, "Workspace %d", i + 1);
-			workspace->name = workspace_name;
+			snprintf(workspace_name, sizeof(workspace_name), "Workspace %d", i + 1);
+			workspace->name = xstrdup(workspace_name);
 			wl_list_append(&rc.workspace_config.workspaces, &workspace->link);
 		}
 	}
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -616,7 +616,7 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "popupTime.desktops")) {
 		rc.workspace_config.popuptime = atoi(content);
 	} else if (!strcasecmp(nodename, "number.desktops")) {
-		rc.workspace_config.count = MAX(1, atoi(content));
+		rc.workspace_config.min_nr_workspaces = MAX(1, atoi(content));
 	}
 }
 
@@ -759,7 +759,7 @@ rcxml_init(void)
 	rc.window_switcher.outlines = true;
 
 	rc.workspace_config.popuptime = INT_MIN;
-	rc.workspace_config.count = 1;
+	rc.workspace_config.min_nr_workspaces = 1;
 }
 
 static struct {
@@ -1048,14 +1048,14 @@ post_processing(void)
 		assert(l && libinput_category_get_default() == l);
 	}
 
-	int workspaces_configured = wl_list_length(&rc.workspace_config.workspaces);
-	if (workspaces_configured < rc.workspace_config.count) {
+	int nr_workspaces = wl_list_length(&rc.workspace_config.workspaces);
+	if (nr_workspaces < rc.workspace_config.min_nr_workspaces) {
 		struct workspace *workspace;
-		char workspace_name[32]; // Maximum length of workspace name "Workspace X"
-		for (int i = workspaces_configured; i < rc.workspace_config.count; i++) {
+		for (int i = nr_workspaces; i < rc.workspace_config.min_nr_workspaces; i++) {
 			workspace = znew(*workspace);
-			snprintf(workspace_name, sizeof(workspace_name), "Workspace %d", i + 1);
-			workspace->name = xstrdup(workspace_name);
+			char *workspace_name = malloc(sizeof(char) * 32);
+			snprintf(workspace_name, 32, "Workspace %d", i + 1);
+			workspace->name = workspace_name;
 			wl_list_append(&rc.workspace_config.workspaces, &workspace->link);
 		}
 	}
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -8,6 +8,7 @@
 #include <libxml/tree.h>
 #include <stdbool.h>
 #include <stdio.h>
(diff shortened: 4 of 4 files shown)


Consolatis (commented):
Thanks for working on this!
Some review comments below. Please don't take the comments as diminishing your work in any way. The review process is an important part of any bigger project.

The changes in `src/workspaces.c` are not necessary and can be removed. The documentation regarding the double use of `struct workspace` as config entries (e.g. "what should exist", created in `rcxml.c`) and actual structures that hold the workspace (created by `add_workspace()`) is not the best, sorry about that.

What works best for me in cases like this is something like the following:
```sh
# move to the master branch
git checkout master
# create a copy of the file that should be reverted
cp src/workspaces.c tmp.c
# back to the development branch
git checkout config-workspaces-by-number
# overwrite the file with the copy from the master branch
mv tmp.c src/workspaces.c
# and finally add and commit the changes
git add src/workspaces.c
git commit -m "revert changes to workspaces.c"
```
There might be better ways to do that but this is what works best for me.
Later on we can merge the commits together to have a single clean commit.

Sachin-Bhat:
Thanks a lot for the feedback! This is extremely valuable in my learning process and helps me a great deal! I shall make the changes and push them as per your suggestions. Do let me know if there is anything else that I could improve :)

src/config/rcxml.c
```diff
@@ -1010,10 +1015,19 @@ post_processing(void)
...
+
+	int workspaces_configured = wl_list_length(&rc.workspace_config.workspaces);
+
+	if (workspaces_configured < rc.workspace_config.count) {
+		struct workspace *workspace;
+		char workspace_name[32]; // Maximum length of workspace name "Workspace X"
+
+		for (int i = workspaces_configured; i < rc.workspace_config.count; i++) {
+			workspace = znew(*workspace);
+			snprintf(workspace_name, sizeof(workspace_name), "Workspace %d", i);
```
Consolatis, line 1027: The `snprintf()` call misses a `+ 1` for the `i` value.
Currently it would create something like this with the example from above:
```
- 1st workspace
- 2nd workspace
- Workspace 2
- Workspace 3
```

The two empty lines (before `if()` and `for()`) can be removed.

Other than the missing `+ 1` and the empty line nitpick this seems to only miss the documentation update :)
Feel free to `Resolve` the comments that you took care of.

Sachin-Bhat:
I have resolved the changes that you had suggested. Kindly have a look. If I have to test it, do I have to add either:
```xml
  <desktops number="4">
    <names>
      <name>1st workspace</name>
      <name>2nd workspace</name>
    </names>
  </desktops>
```
or
```xml
  <desktops number="4"/>
```
to `labwc/rc.xml`?

Consolatis:
Both variants would be good to test.
In the first case it should provide the following workspaces:
```
- 1st workspace
- 2nd workspace
- Workspace 3
- Workspace 4
```

And in the second case
```
- Workspace 1
- Workspace 2
- Workspace 3
- Workspace 4
```

Another thing that could be tested is
```xml
  <desktops number="0" />
```
```xml
  <desktops number="-100" />
```
```xml
  <desktops number="invalid_value" />
```
All those cases should result in a single workspace.

If you feel like even more testing:
```xml
  <desktops number="1">
    <names>
      <name>1st workspace</name>
      <name>2nd workspace</name>
    </names>
  </desktops>
```
Should result in
```
- 1st workspace
- 2nd workspace
```

You can test nested, e.g. just starting labwc on top of labwc (or any other random compositor / X session of your choice) and then ctrl-c'ing it when the workspaces show up as expected.

Sachin-Bhat:
> Both variants would be good to test. 

I did try testing and was unable to get any output on the workspaces itself. I tried both `sfwbar` and `waybar` but to no avail. I have edited the `rc.xml` according to the suggestions but once it gets built, the rc.xml file isn't within the build directory. Can you please help me out? Do I need to do some edits to the `meson.build`? If not, I would be happy to know where I made a mistake.

Consolatis:
>  the rc.xml file isn't within the build directory

You have to manually put `rc.xml` into `~/.config/labwc/` (or some other directory and then use `-C /that/other/directory/` when starting labwc).

> I did try testing and was unable to get any output on the workspaces itself.

Usually when you switch the workspaces you'll see them listed in the on screen popup.
When you only got a single workspace the popup will not be created as we didn't actually switch workspaces in that case.
So if you expect a single workspace and you can't switch workspaces everything works as intended.

I've added a debug line in `src/workspaces.c add_workspace()` and verified all the scenarios written above: everything works perfect (including negative numbers, 0 and invalid numbers).

Usually you can switch workspaces by scrolling on the desktop (in the default configuration) but there was a bug in the master branch that has been fixed in the meantime. This PR does not contain the fix though.

If you want to you could rebase your feature branch on top of the latest changes in the master branch. This would be a little exercise in git. Otherwise I could do that for you before merging the PR, just say what works best for you.

```sh
# add the upstream remote to the main repo
git remote add upstream https://github.com/labwc/labwc.git
git fetch --all
# move to your feature branch
git checkout config-workspaces-by-number
# create a backup of your branch just to be sure you can revert when something goes horribly wrong
git branch backup/workspace-by-number config-workspaces-by-number
# rebase your branch on top of the latest state in the master branch
git rebase upstream/master
```

At this point you should be able to test your implementation as the bug described above should now be fixed and you can scroll on the desktop to switch branches.

The next point in this adventure would then be to "squash" your commits into one:
```sh
# start an interactive rebase
git rebase -i upstream/master
```

This should now show you something like this:
```
pick ff2fff4 config: support <desktops number="">
pick e3477af config: support <desktops number="">
pick c7bead2 revert changes to workspaces.c
pick 4c2693b config: support <desktops number="">
pick a72f747 config: support <desktops number=>
pick a1b64ae config: support <desktops number=>

# Rebase e82e557..a1b64ae onto e82e557 (6 commands)
#
# Commands:
# p, pick <commit> = use commit
# r, reword <commit> = use commit, but edit the commit message
# e, edit <commit> = use commit, but stop for amending
# s, squash <commit> = use commit, but meld into previous commit
# f, fixup <commit> = like "squash", but discard this commit's log message
# x, exec <command> = run command (the rest of the line) using shell
# b, break = stop here (continue rebase later with 'git rebase --continue')
# d, drop <commit> = remove commit
# l, label <label> = label current HEAD with a name
# t, reset <label> = reset HEAD to a label
# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
# .       create a merge commit using the original merge commit's
# .       message (or the oneline, if no original merge commit was
# .       specified). Use -c <commit> to reword the commit message.
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
```

You can edit the first word before each of your commits to say `f` or `fixup` but make sure to keep the first line as it is. This would result in something like this:
```
pick ff2fff4 config: support <desktops number="">
f e3477af config: support <desktops number="">
f c7bead2 revert changes to workspaces.c
f 4c2693b config: support <desktops number="">
f a72f747 config: support <desktops number=>
f a1b64ae config: support <desktops number=>
```
Now save and exit the editor. If you now do a `git log -n 1 --patch` you should see a single commit containing all your changes.
You will notice that you can't just push the branch anymore because you rewrote the history (two times actually, first time due to the rebase and the 2nd time due to the squashing).
To solve that you can add the `--force` argument so the command becomes something like
```sh
git push --force origin config-workspaces-by-number:config-workspaces-by-number
```

If something breaks you can restore your original branch by
```sh
git reset --hard backup/workspace-by-number 
```

Sachin-Bhat:
Documentation has been updated at `docs/labwc-config.5.scd` and `docs/rc.xml.all` to the best of my knowledge. Please let me know if I am missing anything. Thanks again for your valuable assistance!

Sachin-Bhat:
On a related note, while testing I did see the various screen popups while switch workspaces, my config files for workspaces for either `sfwbar` or `waybar` were not visible. May I know why this is the case? I am using the experimental build of `waybar` and the git package of `sfwbar`. Is there anything else that must be done to support the workspace pager on these bars? I can't say for the other bars but at least these two.

johanmalm:
> May I know why this is the case? I am using the experimental build of `waybar` and the git package of `sfwbar`. Is there anything else that must be done to support the workspace pager on these bars?

There is no official workspace protocol for Wayland yet, and we’ve decided not to implement any until proposed ones are accepted.

docs/labwc-config.5.scd
```diff
@@ -196,7 +196,7 @@ Therefore, where multiple objects of the same kind are required (for example
 
 ## WORKSPACES
 
-*<desktops><names><name>*
+*<desktops number=""><names><name>*
```
johanmalm, line 199: We ought to describe here that the number attribute specifies the minimum number of workspaces.
We should also describe that number is an optional attribute.
I’d be good to also say that if number is specified, names.name is not required.

docs/rc.xml.all
```diff
@@ -69,23 +69,27 @@
     See man labwc-actions for further information.
 
     Workspaces can be configured like this:
-    <desktops>
```
johanmalm, line 72: I think it would be better to provide two examples here. One using number and the other (as-is) using names.name 

…because using both together doesn’t really make sense. It should be one or the other.

docs/rc.xml.all
```diff
@@ -69,23 +69,27 @@
...
-  <desktops>
+  <desktops number="4">
     <!--
       popupTime defaults to 1000 so could be left out.
       Set to 0 to completely disable the workspace OSD.
     -->
     <popupTime>1000</popupTime>
     <names>
       <name>Default</name>
+      <name>Browsing</name>
```
johanmalm, line 90: We should remove these. The uncommented sections of rc.xml contain the default values, so this feels inconsistent

include/config/rcxml.h
```diff
@@ -69,6 +69,7 @@ struct rcxml {
 
 	struct {
 		int popuptime;
+		int count;
```
johanmalm, line 72: I think we should find a more descriptive variable name here. “count” is a bit like “i” which is fine in a loop where it’s obvious what it does.

I’m on a phone so can’t easily navigate around the patch, but think I’m right in thinking that “count” is only used in post_processing().

So, does the variable specify the minimum number of workspaces desired by user?

Could we call it ‘min_nr_workspaces’ ? I think that would be helpful to people looking at this in the future. And easier to grep for.

We have tended to use ‘nr_’ as a “number of” prefix throughout the code base rather than “num_” by the way.

src/config/rcxml.c
```diff
@@ -1042,10 +1047,17 @@ post_processing(void)
 		struct libinput_category *l = libinput_category_create();
 		assert(l && libinput_category_get_default() == l);
 	}
-	if (!wl_list_length(&rc.workspace_config.workspaces)) {
-		struct workspace *workspace = znew(*workspace);
-		workspace->name = xstrdup("Default");
-		wl_list_append(&rc.workspace_config.workspaces, &workspace->link);
+
+	int workspaces_configured = wl_list_length(&rc.workspace_config.workspaces);
```
johanmalm, line 1051: Suggest nr_workspaces instead of workspaces_configured

src/config/rcxml.c
```diff
@@ -1042,10 +1047,17 @@ post_processing(void)
...
+	if (workspaces_configured < rc.workspace_config.count) {
+		struct workspace *workspace;
+		char workspace_name[32]; // Maximum length of workspace name "Workspace X"
```
johanmalm, line 1054: I’d remove comment because it should be self-explanatory. Also, we ought to allow for translation in future in which case 32 might not be enough. But have to sort translation as a different patch.

We try to use /* */ comments instead of //.
In fact I’m surprised that the checkpatch.pl script didn’t catchy that.

johanmalm (commented):
Thanks. A fee fairly trivial comments. Looks good on the whole.

include/config/rcxml.h
Sachin-Bhat: > I’m on a phone so can’t easily navigate around the patch, but think I’m right in thinking that “count” is only used in post_processing().

It is used in entry() as well.

> So, does the variable specify the minimum number of workspaces desired by user?

Yes that is correct!

> Could we call it ‘min_nr_workspaces’ ? I think that would be helpful to people looking at this in the future. And easier to grep for.
> 
> We have tended to use ‘nr_’ as a “number of” prefix throughout the code base rather than “num_” by the way.

Noted.

src/config/rcxml.c
Sachin-Bhat: > I’d remove comment because it should be self-explanatory. Also, we ought to allow for translation in future in which case 32 might not be enough. But have to sort translation as a different patch.
> 
> We try to use /* */ comments instead of //. In fact I’m surprised that the checkpatch.pl script didn’t catchy that.

Do you want me to dynamically declare this? For example, using a linked list may help?

Sachin-Bhat: I could try something like this maybe:
Replace lines 1054 to 1056 with
```c
        char *workspace_name = malloc(sizeof(char) * 32); /* Dynamically allocate memory */
        snprintf(workspace_name, 32, "Workspace %d", i + 1);
        workspace->name = workspace_name;
```

Would love to hear your thoughts!

Sachin-Bhat: This should fix the fact that
> in which case 32 might not be enough.

Consolatis: It could be done this way to make it easier to adjust the size later on when we add the translation. I am not sure if we should do that within this PR though but maybe rather change it when adding the translation. And then also use `xmalloc()` instead of `malloc()` (our wrapper for it, see `include/common/mem.h`).

Sachin-Bhat: Will revert back with the necessary changes.

Sachin-Bhat:
I have resolved all the comments given by you. Please let me know if you would like me to address anything else. Until then I shall mark it `Ready for review` to allow for merging.
Thank you so much @johanmalm and @Consolatis for your valuable assistance! It was a rewarding experience for me!

docs/rc.xml.all
```diff
@@ -75,19 +75,18 @@
...
         <name>Workspace 2</name>
         <name>Workspace 3</name>
+        <name>Workspace 4</name>
       </names>
     </desktops>
+
+    Or it can also be configured like this:
+    <desktops number="4" />
   -->
-  <desktops>
```
johanmalm, line 81: You've deleted the a section of the man page. Would you mind putting it back.

```
<desktops>
<popupTime>1000</popupTime>
    <names>
      <name>Default</name>
    </names>
  </desktops>
```

johanmalm:
A few lines of the man page were deleted - by mistake by the looks of it.
Other than that it looks good to me.
If you wouldn't mind putting those lines back, I'll take it for a test drive and we're good.

src/config/rcxml.c
johanmalm: > I could try something like this maybe:
> 
> ```c
>         char *workspace_name = malloc(sizeof(char) * 32); /* Dynamically allocate memory */
>         snprintf(workspace_name, 32, "Workspace %d", i + 1);
>         workspace->name = workspace_name;
> ```
> 
> Would love to hear your thoughts!

What you'd normally do here is to use a string library of some sort :smile:
There is glib's `g_strdup_printf()` and with `DEFINE_GNU_SOURCE` you get various heap-alloc varieties of sprintf family such as `asprintf()`. We could just use `asprintf()` but not everyone likes using compilers / C-libraries that support these so it always stirs up debate which we try to stay clear of.

The problem with the code snippet above it that you're still defining the size at 32.

Don't worry about it for now. There are other areas of the code-base that would benefit from this anyway. The simplest way is to do a wrapper around `vsnprintf()` to first get the size, then allocate the buffer, then populate it.
I'll do a PR :smile:

johanmalm:
Have tested. Works well.

johanmalm:
Thanks

PR #959, opened 2023-06-25
Add common/strdup-printf.c

johanmalm:


johanmalm:
@Sachin-Bhat for info.

Sachin-Bhat:
Noted.

johanmalm:
Ready from my perspective.

Consolatis:
PR itself looks good to me.

Unrelated to this PR I stumbled on this part that seems a bit strange:

https://github.com/labwc/labwc/pull/959/files#diff-77c463bc6f74f9247b2dea0b5e282b23706a577c5127c69fe0507ef112168135R129-R142

We are returning a `const char *`  with heap allocated memory. And then casting it to a void pointer to free it.
Shouldn't we return a `char *` for `build_path()` instead?

johanmalm:
Thanks for review.
Good pick up ref the `const char`. I've taken the liberty of adding a commit to un-constify.

PR #961, opened 2023-06-25
config: support libinput sendEventsMode

### LINKED ISSUE #559: [feature request] support libinput send events mode
...to enable/disable input devices

https://wayland.freedesktop.org/libinput/doc/latest/configuration.html#send-events-mode

Sachin-Bhat:
Fixes #559. 

I have changed the following files:
- `include/config/libinput.h`
- `src/config/libinput.c`
- `src/config/rcxml.c`

I shall update the documentation once the code review is complete:)

### DIFF
--- a/src/config/libinput.c
+++ b/src/config/libinput.c
@@ -24,6 +24,7 @@ libinput_category_init(struct libinput_category *l)
 	l->middle_emu = -1;
 	l->dwt = -1;
 	l->click_method = -1;
+	l->send_events_mode = -1;
 }
 
 enum lab_libinput_device_type
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -461,6 +461,29 @@ get_accel_profile(const char *s)
 	return -1;
 }
 
+static int
+get_send_events_mode(const char *s)
+{
+	if (!s) {
+		goto err;
+	}
+
+	int ret = parse_bool(s, -1);
+	if (ret >= 0) {
+		return ret
+			? LIBINPUT_CONFIG_SEND_EVENTS_ENABLED
+			: LIBINPUT_CONFIG_SEND_EVENTS_DISABLED;
+	}
+
+	if (!strcasecmp(s, "disabledOnExternalMouse")) {
+		return LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE;
+	}
+
+err:
+	wlr_log(WLR_INFO, "Not a recognised send events mode");
+	return -1;
+}
+
 static void
 fill_libinput_category(char *nodename, char *content)
 {
@@ -576,6 +599,9 @@ fill_libinput_category(char *nodename, char *content)
 		} else {
 			wlr_log(WLR_ERROR, "invalid clickMethod");
 		}
+	} else if (!strcasecmp(nodename, "sendEventsMode")) {
+		current_libinput_category->send_events_mode =
+			get_send_events_mode(content);
 	}
 }
 
--- a/src/seat.c
+++ b/src/seat.c
@@ -91,6 +91,24 @@ get_category(struct wlr_input_device *device)
 static void
 configure_libinput(struct wlr_input_device *wlr_input_device)
 {
+	/*
+	 * TODO: We do not check any return values for the various
+	 *       libinput_device_config_*_set_*() calls. It would
+	 *       be nice if we could inform the users via log file
+	 *       that some libinput setting could not be applied.
+	 *
+	 * TODO: We are currently using int32_t with -1 as default
+	 *       to desribe the not-configured state. This is not
+	 *       really optimal as we can't properly deal with
+	 *       enum values that are 0. After some discussion via
+	 *       IRC the best way forward seem to be to use a
+	 *       uint32_t instead and UINT32_MAX as indicator for
+	 *       a not-configured state. This allows to properly
+	 *       test the enum being a member of a bitset via
+	 *       mask & value == value. All libinput enums are
+	 *       way below UINT32_MAX.
+	 */
+
 	if (!wlr_input_device) {
 		wlr_log(WLR_ERROR, "no wlr_input_device");
 		return;
@@ -208,6 +226,16 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		 */
 
 		libinput_device_config_click_set_method(libinput_dev, dc->click_method);
+	}
+
+	if ((dc->send_events_mode != LIBINPUT_CONFIG_SEND_EVENTS_ENABLED
+			&& (libinput_device_config_send_events_get_modes(libinput_dev)
+				& dc->send_events_mode) == 0)
+			|| dc->send_events_mode < 0) {
+		wlr_log(WLR_INFO, "send events mode not configured");
+	} else {
+		wlr_log(WLR_INFO, "send events mode configured");
+		libinput_device_config_send_events_set_mode(libinput_dev, dc->send_events_mode);
 	}
 }
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -684,6 +684,20 @@ extending outward from the snapped edge.
 
 	The default method depends on the touchpad hardware.
 
+*<libinput><device><sendEventsMode>* [yes|no|disabledOnExternalMouse]
+	Optionally enable or disable sending any device events.
+
+	The options available are:
+	- *yes* - Events are sent as usual
+	- *no* - No events are sent from this device
+	- *disabledOnExternalMouse* - This device does not send events if an
+	  external mouse has been detected.
+
+	It is possible to prevent events from a device in the config and then do
+	a Reconfigure to temporarily enable / disable specific devices.
+
+	By default, this setting is not configured.
+
 ## WINDOW RULES
 
 Two types of window rules are supported, actions and properties. They are
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -473,6 +473,7 @@
       - accelProfile [flat|adaptive]
       - tapButtonMap [lrm|lmr]
       - clickMethod [none|buttonAreas|clickfinger]
+      - sendEventsMode [yes|no|disabledOnExternalMouse]
   -->
   <libinput>
     <device category="default">
@@ -487,6 +488,7 @@
       <middleEmulation></middleEmulation>
       <disableWhileTyping></disableWhileTyping>
       <clickMethod></clickMethod>
+      <sendEventsMode></sendEventsMode>
     </device>
   </libinput>
 
--- a/include/config/libinput.h
+++ b/include/config/libinput.h
@@ -23,12 +23,13 @@ struct libinput_category {
 	int left_handed;
 	enum libinput_config_tap_state tap;
 	enum libinput_config_tap_button_map tap_button_map;
-	int tap_and_drag; /* -1 or libinput_config_drag_state */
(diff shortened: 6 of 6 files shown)


src/config/libinput.c
```diff
@@ -41,6 +42,7 @@ libinput_category_create(void)
 {
 	struct libinput_category *l = znew(*l);
 	libinput_category_init(l);
+	l->send_events_mode = LIBINPUT_CONFIG_SEND_EVENTS_ENABLED;
```
Consolatis, line 45: This seems redundant as we call `libinput_category_init()` before this line which also sets the flag.

Consolatis:
Parsing looks good on the first glance. Nice job!

This is not exactly my area of expertise but I assume the new flag has to be used in `src/seat.c` `configure_libinput()` so it does something with the input devices.

src/config/libinput.c
Sachin-Bhat: Yeah, my bad :sweat_smile:. I just realised. Will update with relevant changes!

Sachin-Bhat:
> Parsing looks good on the first glance. Nice job!

Thank you!
> This is not exactly my area of expertise but I assume the new flag has to be used in `src/seat.c` `configure_libinput()` so it does something with the input devices.

I have updated `src/seat.c` `configure_libinput()` to do the relevant job with the input devices. Do let me know if I missed out anything.

Sachin-Bhat:
@johanmalm just wanted your opinion on `src/seat.c` as well. I did read the documentation link provided in #559 and have created an implementation based on my understanding. Please let me know if I may be missing something.

src/config/rcxml.c
```diff
@@ -317,6 +317,24 @@ get_accel_profile(const char *s)
 	return LIBINPUT_CONFIG_ACCEL_PROFILE_FLAT;
 }
 
+static enum libinput_config_send_events_mode
+get_events_mode(const char *s)
+{
+	if (!s) {
+		return LIBINPUT_CONFIG_SEND_EVENTS_ENABLED;
```
johanmalm, line 324: We could just return `-1` here and warn that it is not a recognised mode.

src/config/rcxml.c
```diff
@@ -317,6 +317,24 @@ get_accel_profile(const char *s)
...
+	if (!strcasecmp(s, "enabled")) {
+		return LIBINPUT_CONFIG_SEND_EVENTS_ENABLED;
+	}
+	if (!strcasecmp(s, "disabled")) {
+		return LIBINPUT_CONFIG_SEND_EVENTS_DISABLED;
+	}
+	if (!strcasecmp(s, "disabled_on_external_mouse")) {
+		return LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE;
+	}
+	return LIBINPUT_CONFIG_SEND_EVENTS_ENABLED;
```
johanmalm, line 335: Same here.

src/config/rcxml.c
```diff
@@ -390,6 +408,9 @@ fill_libinput_category(char *nodename, char *content)
 		current_libinput_category->dwt = ret
 			? LIBINPUT_CONFIG_DWT_ENABLED
 			: LIBINPUT_CONFIG_DWT_DISABLED;
+	} else if (!strcasecmp(nodename, "sendEventsMode")) {
+		current_libinput_category->send_events_mode =
+			get_events_mode(content);
```
johanmalm, line 413: Should it be `get_send_events_mode()`? (with the _send_)

src/seat.c
```diff
@@ -143,6 +143,19 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		wlr_log(WLR_INFO, "dwt configured");
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
+
+	if (libinput_device_config_send_events_get_modes(libinput_dev) !=
```
johanmalm, line 147: I think this should be a bitwise operation, so something like

```c
uint32_t possible_modes = libinput_device_config_send_events_get_modes(libinput_dev);
if (possible_modes & LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE) {
    fprintf(stderr, "info: we support send-events-disabled-on-external-mouse\n");
}
```

Sachin-Bhat: I personally do not understand why a bitwise operation may be better in this case. From what I understand the two implementations serve slightly different purposes and which one is serves our purpose may depend on the specific requirements of our program.

My implementation is trying to set the send events mode to `LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE` if the current mode is not `LIBINPUT_CONFIG_SEND_EVENTS_DISABLED`. It checks if the operation is successful and logs the result. This is a direct approach to set a specific mode and handle the result.

Your suggestion from my understanding, on the other hand, is checking if the device supports the `LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE` mode. It uses bitwise AND operation to check if the mode is present in the possible modes. This is a good way to check for support of a specific mode, but it doesn't actually set the mode.

I am still a bit muddled with the requirements. So, if our goal is to set a specific mode and handle the result, my implementation should work well. If we just want to check if a specific mode is supported, the second implementation is more suitable.

However, in a more comprehensive implementation, I suggest a combination of the two approaches, as in we might want to first check if a mode is supported (like in your implementation) and then try to set it (like in my implementation). This would ensure that we only try to set modes that are actually supported by the device, which could help to avoid unnecessary errors.

Please do correct me if you find some error in my understanding:)

Consolatis: I'd suggest something like this instead:
```c
if (dc->send_events_mode >= LIBINPUT_CONFIG_SEND_EVENTS_ENABLED) {
	uint32_t possible_modes = libinput_device_config_send_events_get_modes(libinput_dev);
	if (dc-send_events_mode == LIBINPUT_CONFIG_SEND_EVENTS_ENABLED
			|| possible_modes & dc->send_events_mode) {
		libinput_device_config_send_events_set_mode(libinput_dev, dc->send_events_mode);
	}
}
```
With `dc->send_events_mode` defaulting to `-1`.
The reason for the weird `== LIBINPUT_CONFIG_SEND_EVENTS_ENABLED || possible_modes & mode` check is that libinput uses `0` for the enabled mode which obviously can't be checked within a bitset and is always supported: https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/src/libinput.h#L5012

Kudos to malm for finding the enum definition.

Sachin-Bhat: Thanks for the suggestions, I am going to push a commit that will contain your suggestions along with @johanmalm 's approach of 
> ```
> if (l->send_events_mode < 0) {
>     /* let's not configure this because the user didn't ask us to */
> } else {
>     /* do some stuff with libinput_device_config_send_events_set_mode() */
> }
> ```
putting your suggestions in the else block along with relevant error and info messages.

Sachin-Bhat:
Do let me know if you have further suggestions with respect to the code. If not I can get started with the documentation.

src/config/rcxml.c
```diff
@@ -317,6 +317,26 @@ get_accel_profile(const char *s)
 	return LIBINPUT_CONFIG_ACCEL_PROFILE_FLAT;
 }
 
+static enum libinput_config_send_events_mode
+get_send_events_mode(const char *s)
+{
+	if (!s) {
+		wlr_log(WLR_INFO, "Not a recognised node.");
```
johanmalm, line 324: s/node/mode/

Also, I suggest using ERROR instead of INFO for this to alert the user without them having to run in info/debug mode

src/seat.c
```diff
@@ -143,6 +143,25 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		wlr_log(WLR_INFO, "dwt configured");
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
+
+	if (dc->send_events_mode < 0) {
```
johanmalm, line 147: No need to do this anymore. I can’t quite remember what the construct was before that made me suggest it.
The next line now covers it.

src/seat.c
```diff
@@ -143,6 +143,25 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
...
+		enum libinput_config_status status;
+		if (dc->send_events_mode == LIBINPUT_CONFIG_SEND_EVENTS_ENABLED
+			|| possible_modes & dc->send_events_mode) {
+			status = libinput_device_config_send_events_set_mode(libinput_dev,
+				dc->send_events_mode);
+			if (status != LIBINPUT_CONFIG_STATUS_SUCCESS) {
+				wlr_log(WLR_ERROR, "Failed to set events mode.");
+			} else {
+				wlr_log(WLR_INFO, "Send events mode set successfully!");
+			}
```
johanmalm, line 160: You can’t have two else :smile:

I think it would be tidier here to do a switch (status)… and handle the three enums (success, unsupported, invalid)

https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#ga33ff306ed8505129adce4c092a3ed175

src/config/rcxml.c
Consolatis: and we might want to change the message to something like "Not a recognized mode for libinput sendEventsMode".
Otherwise a user seeing that error message in the logs won't necessarily know what that error is about.

src/seat.c
```diff
@@ -143,6 +143,25 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
...
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
+
+	if (dc->send_events_mode < 0) {
+	} else if (dc->send_events_mode >= LIBINPUT_CONFIG_SEND_EVENTS_ENABLED) {
+		uint32_t possible_modes =
+			libinput_device_config_send_events_get_modes(libinput_dev);
+		enum libinput_config_status status;
+		if (dc->send_events_mode == LIBINPUT_CONFIG_SEND_EVENTS_ENABLED
+			|| possible_modes & dc->send_events_mode) {
```
Consolatis, line 153: Our code style for this project requests that all continuation lines in a multi-line if statement should be indented two times to differentiate between the condition and the body. See the snippet I posted above for an example.

Consolatis:
Rebased and changed slightly:
- moved from `[enabled|disabled|disabled_on_external_mouse]` to our usual bool `[yes|no|disabled_on_external_mouse]`
- added docs
- slightly restructured the code in seat.c when applying the config

So far I did not test this at all and thus also don't know if changing the setting and doing a `Reconfigure` actually works. Also not sure if the `disabled_on_external_mouse` setting actually works.

jp7677:
I just pull this PR locally to disable the touchpad on my dual sense controller to not interfere with in-game controller behavior, it worked perfectly. Looking forward for this to land.

Consolatis:
Thanks for testing, If you have the time, could you toggle the setting and then do a `Reconfigure` (or `labwc --reconfigure`) to verify that it actually works to enable / disable a device during runtime?

jp7677:
> Reconfigure also does work, as long as not setting the value back to `on`. This is the same issue as in #1556 because `LIBINPUT_CONFIG_SEND_EVENTS_ENABLED` is zero and thus fails the bitmask test.

Yes, applying the same fix fixes this and reconfigure works correctly now.

Consolatis:
Squashed, ready to go from my POV.

jp7677:
I re-tested again after the last commits, reconfigure works now nicely, also for switching back to the default.

ahesford (approved):

PR #963, opened 2023-06-26
add MoveRelative

Ph42oN:
Adding MoveTo showed better way to implement this than i have been using in my own fork, now this should work like openbox specification says.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -75,6 +75,7 @@ enum action_type {
 	ACTION_TYPE_LOWER,
 	ACTION_TYPE_RESIZE,
 	ACTION_TYPE_MOVETO,
+	ACTION_TYPE_MOVE_RELATIVE,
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks.

johanmalm:
Ah, we should have added an entry in the man page. Do you want to or shall I?

Ph42oN:
I did another commit where i add that. Would it be good like that?

Consolatis:
Commit looks good. But I don't think it will show up here as this PR is already merged.
You might want to open a new PR for it. (Just in case: your branch can be rebased on latest master or the commit can be cherry-picked into a new branch).

Ph42oN:
Actually im planning to make PR for ResizeRelative. Maybe i just include it there. But i have some questions about how should i do it. I'll make draft.

PR #966, opened 2023-06-27
add ResizeRelative

Ph42oN:
I added ResizeRelative, but is it good like this or would it be better if i moved code to calculate new window geometry to action.c instead creating new function in view.c?

Also included adding MoveRelative to man page that was missing from my last PR.


Ph42oN:
I don't know why this freebsd build fails.

Also i have question unrelated to this PR, would you want to add option like max_render_time in sway if i make it?

johanmalm:
I've not tested PR, but it looks good to me with one nitpick: In way of coding style, we prefer `if (left)` over `if (left != 0)`. I suspect `./scripts/check` would pick that up.

My personal preference would be to merely do:

```
	newgeo.x -= left;
	newgeo.width += left + right;
	newgeo.y -= top;
	newgeo.height += top + bottom;
```

I think that the separate `view_resize_relative()` is tidy (as you've done it).

Regarding your 'aside' question, I think `max_render_time` would be good. I've never tested it with sway. Do you have a use-case where you can see a difference?

There is a summer project on the topic (https://blog.krx.sh/) so would be good to reach out to see where that might take us in terms of wlroots scene-graph API in future - just to avoid having to revert in future. I think Rose operates under `rosefromthedead` on IRC. Or I can ask if you prefer.

Consolatis:
I am not sure if we should use `view->current` rather than `view->pending` here. Same for the previous PR.
When the new actions are bound to a key and that key is kept being pressed I assume the current one feels a bit clunky while the `->pending` variant should be smoother.

Ph42oN:
Oh right you can do the same without ifs, i guess thats better. I will change that and use view->pending instead.

That max_render_time is for reducing latency, and to me wayland has been worse than xorg for some games due to increased latency.

johanmalm:
LGTM

Consolatis:
Had a quick test nested and it seems to work fine. Thanks!

PR #626, opened 2022-11-16
Update for wlroots 0.17

jlindgren90:
This is a set of changes required to build with wlroots `master` / 0.17.0-dev (labwc `master` currently uses wlroots 0.16.x).

Perhaps it will be useful to other wanting to use wlroots-git. I plan to update/rebase it as further breaking changes are made on the wlroots side.

PR #971, opened 2023-07-01
view: add client_request flag to view->impl->unmap()

jlindgren90:
This makes explicit the subtle behavioral difference between `xwayland_view_unmap()` and `handle_unmap()`.

With this change, the XDG and XWayland versions of `handle_map/unmap()` are now identical, which will make further refactoring possible.


jlindgren90:
Split out from https://github.com/labwc/labwc/pull/626 as suggested by @Consolatis.

johanmalm:
LGTM

Consolatis:
Thanks

PR #973, opened 2023-07-02
Set untiled in ResizeRelative and maybe other fixes

Ph42oN:
I noticed that if window is snapped to edge, resized with ResizeRelative, then moved with mouse it returns to size it was before snap to edge. Can be fixed by adding view_set_untiled.

Also both MoveRelative and ResizeRelative don't check if view is fullscreen or maximized, should we change that? I think they should be disabled at least for fullscreen.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -629,9 +629,10 @@ actions_run(struct view *activator, struct server *server,
 			}
 			break;
 		case ACTION_TYPE_MOVE_RELATIVE:
-			if (view) {
+			if (view && !view->fullscreen) {
 				int x = get_arg_value_int(action, "x", 0);
 				int y = get_arg_value_int(action, "y", 0);
+				view_maximize(view, false, false);
 				view_move(view, view->pending.x + x, view->pending.y + y);
 			}
 			break;
--- a/src/view.c
+++ b/src/view.c
@@ -206,12 +206,15 @@ view_move_resize(struct view *view, struct wlr_box geo)
 void
 view_resize_relative(struct view *view, int left, int right, int top, int bottom)
 {
+	if (view->fullscreen || view->maximized)
+		return;
 	struct wlr_box newgeo = view->pending;
 	newgeo.x -= left;
 	newgeo.width += left + right;
 	newgeo.y -= top;
 	newgeo.height += top + bottom;
 	view_move_resize(view, newgeo);
+	view_set_untiled(view);
 }
 
 void


Consolatis:
> I noticed that if window is snapped to edge, resized with ResizeRelative, then moved with mouse it returns to size it was before snap to edge. Can be fixed by adding `view_set_untiled`.

PR LGTM

> Also both MoveRelative and ResizeRelative don't check if view is fullscreen or maximized, should we change that? I think they should be disabled at least for fullscreen.

Agree on the fullscreen part. For the maximized case we may either exit maximized state and adjust the geometry based on the current one (e.g. whats being done in the snap_to_edge / region tiled case) or we can simply ignore requests while in maximized state. I am not sure which one is more intuitive for the user.

Ph42oN:
On fluxbox there is options, user can choose if window can be moved or resized when maximized. Didn't find anything about this in openbox documentation.

Consolatis:
> If we don't want to add option i think allow resize and disable moving would make more sense.

Agree on disabling moving while maximized.
For resize I am still not sure what is best. Thoughts @johanmalm @jlindgren90 ?

Ph42oN:
If i use set_maximized, the function needs to be moved below set_maximized or it gives compile error.

jlindgren90:
I think (not 100% sure) that moving/resizing with the mouse currently unmaximizes the view, so for consistency I would vote for both MoveRelative and ResizeRelative to do the same.

Ph42oN:
Looks like moving with mouse restores natural_geometry and resizing is not allowed on maximized.

Ph42oN:
Looks like thats true for moving with mouse, but trying to resize with mouse does nothing.

jlindgren90:
Okay, I may be remembering wrong. Can't check right now. I don't care strongly as long as it's consistent.

Ph42oN:
Now this commit should make it consistent with how they work using mouse.

johanmalm:
I’ll do some testing tomorrow if that’s okay.

johanmalm:
Having said that, the patch looks good to me. No issue with merging

johanmalm:
Thanks. LGTM. Have tested and seems fine.

As a side-note, `MoveRelative` on a snapped-to-edge window ought to revert it to natural-geometry just like moving with mouse does.

Ph42oN:
Ok, i didn't think of that. I can change that too sometime soon. Im starting to think, should i make view_move_relative to keep it more simple in action.c?

Consolatis:
> should i make view_move_relative to keep it more simple in action.c?

Yes, that sounds good.

PR #974, opened 2023-07-02
Support CDATA for <menu><item><action name="Execute"><command>

johanmalm:


johanmalm:
Okay. Ready from my perspective now.

Issue #980, opened 2023-07-08
Add support for `ext_idle_notifier_v1` as swayidle removed support for `org_kde_kwin_idle`


CaptainBloodz:
user@amd64 ~ $ labwc -v
labwc 0.6.3-57-ge06a8cf

user@amd64 ~ $ swayidle -d
2023-07-08 15:35:47 - [Line 708] Register idle timeout at 600000 ms
2023-07-08 15:35:47 - [Line 709] Setup idle
2023-07-08 15:35:47 - [Line 671] Command: grim  /tmp/screen_locked.png && convert /tmp/screen_locked.png -scale 5% -scale 2000% /tmp/screen_locked.png && swaylock -f -i /tmp/screen_locked.png
2023-07-08 15:35:47 - [Line 1041] Loaded config at /home/luc/.config/swayidle/config
2023-07-08 15:35:47 - [Line 1074] Compositor doesn't support idle protocol

swayidle arguments launch fine from foot terminal.

Working fine in sway.

What can I be missing?

Consolatis:
Thanks for reporting this.

There are two protocols for idle notifications. Maybe swayidle switched to the `ext-` variant and dropped support for the `kde-`  variant. Labwc only supports the `kde-` one at the moment.

Just checked wlroots 0.16 and it provides support for both so we should simply add it to labwc as well.
As a short term measure it might be enough to use an older version of swayidle (or just revert https://github.com/swaywm/swayidle/commit/61d653fb7a752910f71efb3e7d08d75b266cf3ea).

See also: https://github.com/swaywm/swayidle/pull/137

CaptainBloodz:
> Thanks for reporting this.
> 
> There are two protocols for idle notifications. Maybe swayidle switched to the `ext-` variant and dropped support for the `kde-` variant. Labwc only supports the `kde-` one at the moment.
> 
> Just checked wlroots 0.16 and it provides support for both so we should simply add it to labwc as well. As a short term measure it might be enough to use an older version of swayidle (or just revert [swaywm/swayidle@61d653f](https://github.com/swaywm/swayidle/commit/61d653fb7a752910f71efb3e7d08d75b266cf3ea)).

Previously running git-master,, switching back to  1.8.0 (2022-12-04) fixed the error indeed.

CaptainBloodz:
[...]

PR #981, opened 2023-07-08
Add support for ext_idle_notify_v1 and combine both variants into a standalone manager

Consolatis:
Should fix
- #980

Potential issues:
- It seems our implementation for the idle inhibitor is not complete, it looks like it should check if one of the active inhibitor surfaces is currently visible. I would have to look up the scene API again if there is an easy way for us to figure that out.
- We don't properly clean up the manager when shutting down labwc. This is arguably  a pretty minor issue as the memory is released on exit anyway. We could add that here by listening to the `wl_display` destroy event.

### DIFF
--- a/include/idle.h
+++ b/include/idle.h
@@ -1,0 +1,11 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_IDLE_H
+#define LABWC_IDLE_H
+
+struct wl_display;
+struct wlr_seat;
+
+void idle_manager_create(struct wl_display *display, struct wlr_seat *wlr_seat);
+void idle_manager_notify_activity(struct wlr_seat *seat);
+
+#endif /* LABWC_IDLE_H */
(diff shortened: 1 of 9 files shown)


johanmalm:
Great stuff. This looks a lot tidier... as well as supporting `ext-idle-notifier`.
LGTM from me + vote for merge.

> Should fix
> 
>     * [Add support for `ext_idle_notifier_v1` as swayidle removed support for `org_kde_kwin_idle` #980](https://github.com/labwc/labwc/issues/980)

I've tested it. It does :smile:

> It seems our implementation for the idle inhibitor is not complete, it looks like it should check if one of the active inhibitor surfaces is currently visible. I would have to look up the scene API again if there is an easy way for us to figure that out.

Interesting. Yes, I've just looked at the sway implementation. It seems a bit more sophisticated.

Consolatis:
So ignore the idle-inhibited-surface-is-visible part for now and create an issue about it instead?

I think I'll work on this PR later again and
- [ ] move the `idle_manager_create()` call to `server.c` as its not actually seat specific (although we only support a single one anyway and thus its technically not wrong what we are doing now)
- [ ] add a `idle_manager_destroy()`

johanmalm:
> So ignore the idle-inhibited-surface-is-visible part for now and create an issue about it instead?

Yes. Good plan.

johanmalm:
Really nice work!

Issue #976, opened 2023-07-07
Drag 'n drop: destination window is not raised, but gets focus


Arnaudv6:
Bug reproduction:
labvc from clonned repo (today)
Drag-n-drop between 2 Thunar (pure wayland) windows.

Expected result:
I like it that dest window would not be raised: makes it easier to perform another drag from src to dest...
But then press [del] on keyboard: this will remove file on dest...bad things will happen!

I'd say this behavior is fairly new (less than a month)

Consolatis:
Hm.. so I guess we have two options here:
- raise the target window
- restore old focus after the DND operation

Arnaudv6:
To me it's the second one, no brainer.
But in the end, this could be a FocusSrcAfterDrag rc.xml option.

Consolatis:
Thoughts @johanmalm ?

johanmalm:
> Thoughts @johanmalm ?

Let's me sort out a patch for this one. I should have researched it properly last time :smile:

I've tried with Openbox + mutter and they both restore the old focus after the DND operation finished. Removing the lines below achieves this (and `pcmanfm-qt --desktop` dnd still works fine which was the reason we changed it last time):
https://github.com/labwc/labwc/blob/40aba9e5427b4c1ca6c5f623d1231501c1fa68f5/src/dnd.c#L140-L149

Interestingly though sway behaves differently - because it uses a focus-follows-mouse type policy.
So before we remove the lines above, I need to do some playing with `followMouse` and `raiseOnFocus` (which I suspect the lines above are perfect for).

Arnaudv6:
Floating is 2nd class in sway, and so mouse mouse might be 😅

Arnaudv6:
Thank you both for awesome soft and healthy community 🙂

Consolatis:
Should be fixed by
- #984

Feel free to comment if that is not the case.

Arnaudv6:
Fixed with #984, thanks!

PR #984, opened 2023-07-10
dnd: fix focus at end of dnd operation

johanmalm:
By default, leave keyboard-focus with the surface that originally had it at the end of a dnd operation.  This is consistent with the default behaviour of openbox and mutter.

If the 'focus/followMouse' option is enabled, then focus on the surface under the cursor at the end of the dnd operation.

Fixes: issue #976


Consolatis:
LGTM, didn't test.

Issue #988, opened 2023-07-14
Top left and top right borders render incorrectly.


burein-ita:
Using 6b83db34afca59686d36a458b527b71355680d48
Arch-Linux

Setting `cornerRadius` to `0` causes the top left and to right borders to render incorrectly.
It is more noticeable if the border thickness is set to some stupid value.

![wack](https://github.com/labwc/labwc/assets/131406764/f3fa6523-5b2b-448e-a308-d3a51ab6ff02)

Top picture border is set to 1px.
Bottom one is set to 25px.

Consolatis:
Thanks for bringing this up. It also looks very wrong at other `cornerRadius` values and a 25px border width.
Something is definitely off here!

Consolatis:
Should be fixed by
- #990

Thanks again for bringing this up.

PR #990, opened 2023-07-15
Fix rounded corners

johanmalm:


johanmalm:
@burein-ita Thanks for reporting the bug in #988 
This patch fixes it for me, but if you're able to test and confirm that would be great.

burein-ita:
Just tested it. I can confirm that this patch indeed solves the issue.

Consolatis:
Tested as well, and it mostly LGTM (some debug stuff in there, e.g. giant font size + double button width + different color for the corners + corners attached on the left without any further post processing like hover overlays, icons or wlroots scaling):

20230716_03h59m12s_grim

What I am not sure about is if the inner corners should be rounded as well or not. In some `cornerRadius` settings they seem not-rounded (20 in my case) and in some its rounded (40) in my case (both with `border.width: 25`):
20230716_04h02m46s_grim

Personally I slightly prefer the inner corner being rounded as well, e.g. what is seen on the second screenshot. But whatever we decide to go with it should be consistent IMHO.

johanmalm:
@burein-ita Thanks for testing

@Consolatis Thanks for the review/analysis. I wish there was a of including those images in the comments in the code. I'm not sure I could convey the same with ascii art. Thoughts/answers below:

The first commit actually fixes a number of things but I couldn't sensibly split the commit down any further:

1. The reason for the original weird angles was that the starting position was wrong following the drawing of the arc, so I added another `cairo_move_to()`.
2. With the previous `CAIRO_LINE_CAP_ROUNDED` there were actually 'caps' hiding some sins, so using `BUTT` is cleaner and requires pixel perfection.
3. Drawing arcs doesn't make sense (for us) when line thickness is greater than the radius. I don't think I had ever considered/tested that before.  I haven't analyzed the cairo-arc intended behavior, but for us it resulted in square corner rather than a radius. Your two images in fact illustrated it perfectly. In the first one the line-thickness is indeed greater than the radius, so when drawing the red corner pieces the patch reduces the line-width of the arc to `radius` :smile:. It is not obvious to me how an inner radius could be achieved in this situation. I'm content with the solution above.

Consolatis:
Well, I guess the usual case is one of these two
- `border.width` set to `0`, `cornerRadius` > `0`: works
- `border.width` set to somewhere between `1` and `4`, `cornerRadius` somewhere between `1` to `10`: mostly works

In the second case, the inner radius might be square on low `cornerRadius` values but when increasing it above the `border.width` setting it will get rounded as well. Our defaults set `cornerRadius` to `8` and `border.width` to `1`.

So all in all this seems fine. Ready to merge from my point of view. Nice fix :)

PR #991, opened 2023-07-16
Added missing closing tags

stefonarch:
Noticed when copy&pasted `<action name="ToggleKeybinds">` into `rc.xml` and reloaded config.
Btw nice improvement: this action fixes an accidental closing of lxqt-panel by shortcut.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -8,10 +8,10 @@ labwc - actions
 
 Actions are used in menus and keyboard/mouse bindings.
 
-*<action name="Close">*
+*<action name="Close" />*
 	Close top-most window.
 
-*<action name="Kill">*
+*<action name="Kill" />*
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks!

> Btw nice improvement: this [`ToggleKeybinds`] action fixes an accidental closing of lxqt-panel by shortcut.

Oh, that's a stroke of genius. I hadn't thought of that. The view specific `ToggleKeybinds`as a window-rule indeed solves that problem :smile:

We ought to do the following:

1. Update `rc.xml.all` and the `integration` page on the website with this.
2. Add a section on the `integration` page on lxqt-panel

I'll have a look at this tonight.

stefonarch:
https://aur.archlinux.org/packages/lxqt-labwc-session-git 
I've added  ti to my repos, but not documented.

PR #994, opened 2023-07-17
Trivial wlroots 0.17 compatibility fixes

jlindgren90:
Background: @Consolatis previously mentioned wanting to keep #626 as small as we can (and I agree). These are a couple of the smaller commits that can be merged without breaking compatibility with wlroots 0.16.

- Add a missing header (pulled in indirectly with wlroots 0.16 but no longer with 0.17)
- Remove some asserts that will no longer hold true with wlroots 0.17


Consolatis:
Thanks, LGTM.

PR #995, opened 2023-07-17
xwayland: Fix segfault at exit (seen with wlroots master)

jlindgren90:
Backtrace: https://github.com/labwc/labwc/pull/626#issuecomment-1637515904

I don't see the segfault with wlroots 0.16.2, but it doesn't hurt to cover our bases.


Consolatis:
Thanks

PR #997, opened 2023-07-17
Fall back on raw keysyms when translated ones do not match keybinds

johanmalm:
This enables keybinds like `W-S-1` and `W-S-Tab`

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -141,6 +141,17 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 	translated.nr_syms = xkb_state_key_get_syms(wlr_keyboard->xkb_state,
 		keycode, &translated.syms);
 
+	/*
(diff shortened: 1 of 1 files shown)


Consolatis:
Tested it a bit and it seems to work just fine. Nice job!

What I discovered while testing (which is not a serious issue at all) is that having an existing workaround keybind like `W-S-0x21` (or `W-S-exclam`) configured prevents triggering `W-S-1` because the first one will get hit in the translated path even if it is defined before the `W-S-1` one. 

Usually all previous keybinds get replaced by the last one but I don't really see us adding an unproportional amount of complexity to `keybind_the_same()` just to figure out if two keybinds are actually the same if `WLR_MODIFIER_SHIFT` is set but the keysyms differ (e.g. `XKB_KEY_exclam` vs `XKB_KEY_1`).

johanmalm:
Thanks for review.

Issue #823, opened 2023-03-11
Prevent parent windows from displaying above modal dialog


Flrian:
See https://github.com/labwc/labwc/issues/688#issuecomment-1361972690

For example when opening the about dialog in nautilus you can switch back to the main window and if you didn't pay attention it isn't immediately obvious why you can't interact with it.

I'm also having a bug in calibre where a dialog sometimes (not sure why) spawns below the main application, very noticeable there.

stefonarch:
Same here, some filedialog maybe open and waybar/taskbar shows it with the app_id while the main window is blocked. Tried using window rules using `"title"="Open*"` and `ToggleAlwaysOntop` Tbut this workaround doesn't work unfortunately.

stefonarch:
Fixed by  https://github.com/labwc/labwc/pull/998

johanmalm:
Thanks for the reminder.

PR #998, opened 2023-07-17
Show modal dialog above parent window

johanmalm:
This behaviour is consistent with that of mutter and openbox.

Tested with mousepad's 'about' dialog

Fixes: issue #823

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -233,8 +233,8 @@ view_adjust_size(struct view *view, int *w, int *h)
 	*h = MAX(*h, LAB_MIN_VIEW_HEIGHT);
 }
 
-void
-view_minimize(struct view *view, bool minimized)
+static void
+_minimize(struct view *view, bool minimized)
 {
 	assert(view);
 	if (view->minimized == minimized) {
@@ -265,6 +265,40 @@ view_minimize(struct view *view, bool minimized)
 	} else {
 		view->impl->map(view);
 	}
+}
+
+static void
+minimize_sub_views(struct view *view, bool minimized)
+{
+	struct view **child;
+	struct wl_array children;
+
+	wl_array_init(&children);
+	view_append_children(view, &children);
+	wl_array_for_each(child, &children) {
+		_minimize(*child, minimized);
+		minimize_sub_views(*child, minimized);
+	}
+	wl_array_release(&children);
+}
+
+/*
+ * Minimize the whole view-hierarchy from top to bottom regardless of which one
+ * in the hierarchy requested the minimize. For example, if an 'About' or
+ * 'Open File' dialog is minimized, its toplevel is minimized also. And vice
+ * versa.
+ */
+void
+view_minimize(struct view *view, bool minimized)
+{
+	/*
+	 * Minimize the root window first because some xwayland clients send a
+	 * request-unmap to sub-windows at this point (for example gimp and its
+	 * 'open file' dialog), so it saves trying to unmap them twice
+	 */
+	struct view *root = view_get_root(view);
+	_minimize(root, minimized);
+	minimize_sub_views(root, minimized);
 }
 
 static bool
@@ -1040,6 +1074,25 @@ view_move_to_back(struct view *view)
 	if (view->impl->move_to_back) {
 		view->impl->move_to_back(view);
 		cursor_update_focus(view->server);
+	}
+}
+
+struct view *
+view_get_root(struct view *view)
+{
+	assert(view);
+	if (view->impl->get_root) {
+		return view->impl->get_root(view);
+	}
+	return view;
+}
+
+void
+view_append_children(struct view *view, struct wl_array *children)
+{
+	assert(view);
+	if (view->impl->append_children) {
+		view->impl->append_children(view, children);
 	}
 }
 
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -320,6 +320,71 @@ xdg_toplevel_view_minimize(struct view *view, bool minimized)
 	/* noop */
 }
 
+static struct wlr_xdg_toplevel *
+top_parent_of(struct view *view)
+{
+	struct wlr_xdg_toplevel *toplevel = xdg_toplevel_from_view(view);
+	while (toplevel->parent) {
+		toplevel = toplevel->parent;
+	}
+	return toplevel;
+}
+
+enum z_direction {
+	LAB_TO_FRONT,
+	LAB_TO_BACK,
+};
+
+static void
+move_sub_views(struct view *parent, enum z_direction z_direction)
+{
+	assert(parent);
+
+	if (parent->type != LAB_XDG_SHELL_VIEW) {
+		return;
+	}
+
+	struct wlr_xdg_toplevel *parent_toplevel = xdg_toplevel_from_view(parent);
+	struct view *view, *next;
+	wl_list_for_each_reverse_safe(view, next, &parent->server->views, link)
+	{
+		/* need to stop here, otherwise loops keeps going forever */
+		if (view == parent) {
+			break;
+		}
+		if (view->type != LAB_XDG_SHELL_VIEW) {
+			continue;
+		}
+		if (!view->mapped && !view->minimized) {
+			continue;
+		}
+		if (top_parent_of(view) != parent_toplevel) {
+			continue;
+		}
+		if (z_direction == LAB_TO_FRONT) {
+			view_impl_move_to_front(view);
+		} else if (z_direction == LAB_TO_BACK) {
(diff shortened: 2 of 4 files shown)


Consolatis:
Commit text uses the wrong issue id (303 vs 823).
Will test later but code looks good.

stefonarch:
Nice, it works fine and is great step, but the behavior is a little different than under x11 or in wayfire. If there are 3 windows: firefox, firefox's "open file" dialog and thunderbird for example it is possible with alt-tab to have the dialog on top and firefox at the bottom, closing the dialog thunderbird gets focus. 

I just noticed that under x11 the dialog isn't shown in the taskbar, maybe excluding it from the appswitcher too would exclude the above behavior.

johanmalm:
> IMHO we should make sure that if a subwindow is raised that it instead just raises its parent (which will then due to this PR also raise the subwindow). 

Great catch. Implemented.

Have checked with `openbox` and `mutter` - and that's indeed how they do it with `Gtk3` + `Qt5`.

Consolatis:
Do we need a keyboard focus update as well here?

I didn't test this yet but if consider the following scenario:
- we have a main application and a modal dialog open
- we get some other window on top
- we then switch to the main application

Due to this PR the subwindow would now be raised on top of the main application but wouldn't the keyboard focus still be on the main window instead of the topmost subwindow?

johanmalm:
> ...but wouldn't the keyboard focus still be on the main window instead of the topmost subwindow?

Good check.

I've tested with mousepad (Gtk3) and QTerminal (Qt5). Both refuse to give the toplevel window keyboard-focus when a modal dialog is open. Sometimes the keyboard focus is lots from the modal dialog - sometimes not. Seems to be client dependent.

leafpad (Gtk2) lets the parent window have keyboard focus whilst the modal dialog is open... which is how it works on Openbox too.

I'm slightly nervous about playing with keyboard-focus in case we introduce some unintended, hard-to-spot consequence.

johanmalm:
I've just had a play on Openbox...

With geany and leafpad's "open file" dialogs the keyboard focus stays with that dialog even if the parent window is moved. With this PR the modal dialog loses keyboard focus in the same situation. Hmm....

Consolatis:
As discussed via IRC there doesn't seem to be a way to detect if a wayland surface with a parent relation is a modal or non modal dialog. Examples: A `Ctrl-f` search box in some editor should not prevent the editor to get keyboard focus while `Open file` should.

As we can't differentiate between the two (without implementing some private GTK protocol) we maybe should simply not deal with keyboard focus at all but still make sure that dialogs are always raised on top of the main window. So exactly what this PR is doing right now.

> I'm slightly nervous about playing with keyboard-focus in case we introduce some unintended, hard-to-spot consequence.

At this point I completely agree.

johanmalm:
I think we need to add a `wlr_foreign_toplevel_handle_v1_set_parent()` somewhere to avoid adding child-toplevels to taskbars. Not sure where though. Could do as separate PR.

Other than that, ready to go from my perspective.

stefonarch:
>Other than that, ready to go from my perspective.

It works nicely, thanks!

src/view.c
```diff
@@ -267,6 +267,34 @@ view_minimize(struct view *view, bool minimized)
...
+ * in the hierarchy requested the minimize. For example, if an 'About' or
+ * 'Open File' dialog is minimized, its toplevel is minimized also. And vice
+ * versa.
+ */
+void
+view_minimize(struct view *view, bool minimized)
+{
+	struct view *root = view_get_root(view);
+	_minimize(root, minimized);
+	minimize_sub_views(root, minimized);
```
Consolatis, line 295: `view_get_root()` seems to return `NULL` if `view->impl->get_root` is not set.
It would then call `_minimize()` which asserts `view`. Maybe we should return `view` instead of `NULL` in that case?

src/view.c
```diff
@@ -267,6 +267,34 @@ view_minimize(struct view *view, bool minimized)
 	}
 }
 
+static void
+minimize_sub_views(struct view *view, bool minimized)
+{
+	struct view **child;
+	struct wl_array children = { 0 };
+
+	view_get_children(view, &children);
```
Consolatis, line 276: I am not really familiar with wl_array, no `wl_array_init()` required here?

src/xdg.c
```diff
@@ -320,6 +320,111 @@ xdg_toplevel_view_minimize(struct view *view, bool minimized)
...
+	assert(parent);
+
+	if (parent->type != LAB_XDG_SHELL_VIEW) {
+		return;
+	}
+
+	struct wlr_xdg_toplevel *parent_toplevel = xdg_toplevel_from_view(parent);
+	struct view *view, *next;
+	wl_list_for_each_reverse_safe(view, next, &parent->server->views, link)
+	{
```
Consolatis, line 350: This seems like a copy of `xdg_toplevel_view_get_children()`.
Is it worth combining the two?

Pros:
- less code duplication

Cons:
- additional heap allocations by `wl_array`.

src/view.c
johanmalm: No because we set to zero with `= { 0 }`. All `wl_array_init()` does is a memset to zero.

https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/src/wayland-util.c#L96-100

src/xdg.c
johanmalm: No. `move_sub_views()` is quite subtle in that it uses the `_safe` version of the `for_each` macro because at the bottom of the loop we change the stacking order!

I think there are other areas of the code which could benefit from this type of approach though :smile:
When I get a minute I will experiment with `osd.c` to see if we could get the views we want to iterate over as an array (possibly with options to include/exclude different things). The advantage of wl_array over wl_list in this type of short-lived situation is that we wouldn't need any extra `wl_list link_foo` entries in the structs.

src/view.c
johanmalm: Good catch. Done.

johanmalm:
Need to do some more testing. Get a crash with gimp's 'open file' dialog when (un)minimizing from waybar.

src/view.c
Consolatis: Interesting.
So technically its not necessary at the moment.

I'd still argue that we should always use it to initialize a wl_array as the way it works is an implementation detail of wl_array. They may change it in the future, (e.g. they could add an initial `malloc()` in there and initialize some counters or similar while keeping the API the same).

src/xdg.c
Consolatis: But wouldn't the result be the same if we do something like this?
```c
static void
move_sub_views(struct view *parent, enum z_direction z_direction)
	assert(parent);

	if (parent->type != LAB_XDG_SHELL_VIEW) {
		return;
	}

	struct wl_array subviews;
	wl_array_init(&subviews);
	xdg_toplevel_view_get_children(parent, &subviews);

	struct view *view;
	wl_array_for_each(view, &subviews) {
		if (z_direction == LAB_TO_FRONT) {
			view_impl_move_to_front(view);
		} else if (z_direction == LAB_TO_BACK) {
			view_impl_move_to_back(view);
		}
	}
	wl_array_release(&subiews);
}
```

src/view.c
johanmalm: Okay. Have reverted to using `wl_array_init()`.

src/xdg.c
johanmalm: This is a very good point. Should we do this in a separate PR?
Either way, I'm too tired tonight.

johanmalm:
Have added two commits to fix the gimp open-dialog bug.
Already to go from my perspective.

Consolatis:
> Have added two commits to fix the gimp open-dialog bug. Already to go from my perspective.

Had a quick test with xwayland VLC + about dialog and wayland native pluma + search-and-replace dialog: Seems to work fine. Also had a quick look over the diff again and didn't see anything obvious either.

Regarding combining `move_sub_views()` with `_get_children()` and / or moving `move_sub_views()` to `view-impl.c` (or `view.c` and remove from `struct view_impl`?): I don't mind if we are doing it now or in a future PR. If we decide to 1) do that in the first palce and 2) do it later I'd suggest adding a todo entry on top of both functions so its at least documented in the code.

One last completely nitpicky thing I noticed, up to you if you want to change it or not:
`_get_children()` could be renamed to `_fill_children()` or even `_append_children()` as we supply the array as argument and return nothing.

In any case: nice PR :)

johanmalm:
Thanks for review. Have gone for `append` and made a TODO note about sharing code.

Arnaudv6:
great: thanks :smile:

PR #999, opened 2023-07-20
view: move MoveRelative code from action.c and restore natural geometry

Ph42oN:


src/view.c
```diff
@@ -214,7 +214,22 @@ view_resize_relative(struct view *view, int left, int right, int top, int bottom
 	newgeo.y -= top;
 	newgeo.height += top + bottom;
 	view_move_resize(view, newgeo);
-	view_set_untiled(view);
+	if (view_is_tiled(view))
+		view_set_untiled(view);
+}
```
Consolatis, line 219: Why the additional condition here? `view_set_untiled()` just resets some enum / pointers. I don't really mind the additional condition but I'd be interested in the reasoning for the change.

Small codestyle issue with the whole PR:

Please always use curly braces for all if conditions in the labwc sources, even if there is only a single statement in the body. It's way too easy to add something to the body in the future (for example during debugging or similar) which will then completely mess up things.

src/view.c
```diff
@@ -214,7 +214,22 @@ view_resize_relative(struct view *view, int left, int right, int top, int bottom
...
+
+void
+view_move_relative(struct view *view, int x, int y)
+{
+	if (view->fullscreen)
+		return;
+	if (view->maximized)
+		view_maximize(view, false, false);
```
Consolatis, line 227: An inline comment for the 2nd bool argument would be nice: `view_maximize(view, false,  /*store_natural_geometry*/ false)`

src/view.c
Ph42oN: I was thinking that additional if would be more consistent with how im doing it on view_move_relative. If we dont want that then i think view_maximize also doesn't need it.

Consolatis: So just for consistency, got it.
I am fine with both, e.g. either ifs for both or calling them unconditionally.

Thanks for the update.

Consolatis:
Thanks :)

PR #1003, opened 2023-07-26
[wip] feature: Allow blocking wayland protocols

Consolatis:
Missing:
- [ ] docs (including it only applying to new windows after --reconfigure)
- [ ] discuss config naming and format
- [ ] remove "not blocked: $interface_name" debug log

Test config entry:
```xml
  <blockedProtocols>
        <interface name="zwlr_virtual_pointer_manager_v1" />
        <interface name="zwp_virtual_keyboard_manager_v1" />
        <interface name="zwlr_export_dmabuf_manager_v1" />
        <interface name="zwlr_screencopy_manager_v1" />
        <interface name="zwlr_data_control_manager_v1" />
        <interface name="zwlr_input_inhibit_manager_v1" />
        <interface name="zwlr_foreign_toplevel_manager_v1" />
        <interface name="zwlr_layer_shell_v1" />
  </blockedProtocols>
```

Related:
- https://github.com/labwc/labwc/discussions/1002

PR #1004, opened 2023-07-27
[wip] Only block wayland protocols on a dedicated wayland socket

Consolatis:
Follow-up from
- https://github.com/labwc/labwc/pull/1003

This PR creates a second wayland listening socket
(`$XDG_RUNTIME_DIR/wayland-unpriv`) when there is at
least one protocol blocked in rc.xml.

The blocked protocols are only blocked for the new
socket, the usual wayland socket allows all protocols.

The idea is to use the unprivileged socket for
bind-mounts in sandboxes and similar. Trusted
applications are running against the usual wayland
socket and thus are able to use privileged protocols
(e.g. layershell, screen recording, foreign-toplevel),
whereas clients within a sandbox are prevented to use
those protocols.

Related:
- https://github.com/labwc/labwc/discussions/1002

This PR is very much work in progress.
- [ ] Remove TODO / FIXUP comments:
  - [ ] Add close-on-exec fallbacks for the wayland socket
  - [ ] Add lockfile for the wayland socket
  - [ ] Add close-on-exec for client connections


Consolatis:
Not yet pushed the latest dev work to this PR but this is the current layout for the config format with an example:

```xml
<blockedProtocols>
	<!-- Global fallback block for all sockets if not accepted / blocked by another socket -->
	<interface name="zwlr_virtual_pointer_manager_v1" />
	<interface name="zwp_virtual_keyboard_manager_v1" />
	<interface name="zwlr_export_dmabuf_manager_v1" />
	<interface name="zwlr_screencopy_manager_v1" />
	<interface name="zwlr_data_control_manager_v1" />
</blockedProtocols>

<blockedProtocols socket="wayland-minimal" action="accept">
	<!-- Blocks *all* protocols other than those listed below -->
	<interface name="wl_shm" />
	<interface name="wl_compositor" />
	<interface name="wl_subcompositor" />
	<interface name="wl_seat" />
	<interface name="xdg_wm_base" />
</blockedProtocols>

<blockedProtocols socket="wayland-base" action="reject">
	<!-- Blocks global ones and those listed below -->
	<!-- The "action" argument is optional in this case as it defaults to "reject" -->
	<interface name="zwlr_input_inhibit_manager_v1" />
	<interface name="zwlr_foreign_toplevel_manager_v1" />
	<interface name="zwlr_layer_shell_v1" />
</blockedProtocols>

<blockedProtocols socket="wayland-screenshare" action="accept">
	<!--
		Should block global ones other than those listed below.
	-->
	<!--
		<interface name="zwlr_export_dmabuf_manager_v1" />
		<interface name="zwlr_screencopy_manager_v1" />
	-->
</blockedProtocols>
```

The 4th section (`wayland-screenshare`) is not possible with the current config format, I am currently thinking about how to design the config in a way that supports all use-cases.

Maybe this would be a better option?

| Argument | Description | use-case |
| --- | --- | --- |
| `action="accept"` | directly allow following protocols but fall back to globals filter for further blocks | `wayland-screenshare` |
| `action="reject"` | directly block following protocols but fall back to globals filter for further blocks | `wayland-base` |
| `action="exclusive"` | directly allow the following protocols, block everything else | `wayland-minimal` |

This would keep the config pretty self explaining (I think) and still allow for all use-cases. Or are there any that I've not thought about? Suggestions for better names also appreciated (for example: `mode` vs. `action`?). The hardest part about dev stuff is.. naming things. (Especially true in cases of user facing config files that should be as stable as possible for the foreseeable future).

@jonleivent (and others): Feedback about the config format welcome.

johanmalm:
^ yes I think s/action/mode/ is a good idea. I was thinking that before seeing your question. Helps avoid clashes/confusion with action elements.

Is it wise to use the _V1 suffix on protocol names?

I haven’t got my head around the logic yet, particularly reject, but will browse through code later.

Consolatis:
> ^ yes I think s/action/mode/ is a good idea. I was thinking that before seeing your question. Helps avoid clashes/confusion with action elements.

ACK, will change.

> Is it wise to use the _V1 suffix on protocol names?

Not sure, currently its just a plain `!strcmp()` with the interface name as supplied by wlroots.

> I haven’t got my head around the logic yet, particularly reject, but will browse through code later.

Suggest waiting with that until I push a more recent version, this one is rather outdated and has changed a lot in the meantime.
The current version added the option to have multiple "unprivileged" sockets with their own block/accept list rather than just a single one.

jonleivent:
In the example you show, just to see if I have understood, labwc would do the following?:
- create socket wayland-0 with nothing blocked
- create socket wayland-unpriv with the first blockedProtocol section blocked
- create socket wayland-base with the first and second sections blocked
- create socket wayland-minimal that only accepts the third section

Also, I don't understand why wayland-screenshare is not possible using an accept block.

jonleivent:
I'm still confused about my points above.  However, in regards to:

> This would keep the config pretty self explaining (I think) and still allow for all use-cases. Or are there any that I've not thought about?

I think the ability to block protocols is an example of what can be done with multiple sockets, but not the only thing.  If you want to leave room for expansion later, perhaps have a section introducing the name of the socket, then within that give the blockedProtocol block.  An example of another feature that would fit into this socket-section config syntax: adding the ability to block wl_data_offers in certain directions between sockets (in my ongoing attempts to get a really secure sandboxed clipboard).  Others might be: a prefix tag for the title or app id of windows on that socket.  Maybe a separate debug or verbosity flag per socket.

Another question is how to avoid name clashes between multiple compositors for the same user with similar socket names.  Maybe use the `-#` suffix that gets attached to the standard socket (`wayland-0`) for this purpose.  That would mean the sockets above would get named `wayland-0`, `wayland-unpriv-0`, etc., with all sockets from the same compositor sharing the same number suffix.

Consolatis:
> In the example you show, just to see if I have understood, labwc would do the following?:
> 
> * create socket wayland-0 with nothing blocked

No, the usual wayland socket would block all protocols from the first section

> * create socket wayland-unpriv with the first blockedProtocol section blocked

No, no wayland-unpriv socket will be created unless one of the section uses that as `socket` attribute

> * create socket wayland-base with the first and second sections blocked

Correct

> * create socket wayland-minimal that only accepts the third section

Correct

> Also, I don't understand why wayland-screenshare is not possible using an accept block.

Because "accept" would serve two purposes at once, e.g. the `wayland-minimal` and `wayland-screenshare` examples.
Only one of them works so having a third option like "exclusive_accept" that could be used for the `wayland-minimal` example instead would solve that.

> I think the ability to block protocols is an example of what can be done with multiple sockets, but not the only thing. If you want to leave room for expansion later, perhaps have a section introducing the name of the socket, then within that give the blockedProtocol block.

That is a good point, will change. It would theoretically also allow for other stuff like having different existing option values per socket that overwrite the global ones for clients connected through that socket. Adding a feature like that in the future would be really hacky if we only have the toplevel `blockedProtocol` sections rather than proper `<socket name="foobar">` ones.

> Another question is how to avoid name clashes between multiple compositors for the same user with similar socket names. Maybe use the `-#` suffix that gets attached to the standard socket (`wayland-0`) for this purpose. That would mean the sockets above would get named `wayland-0`, `wayland-unpriv-0`, etc., with all sockets from the same compositor sharing the same number suffix.

That sounds good on the first tought, would likely be more like `wayland-0-socketname_from_config` though, e.g. simply using the existing main socket name as prefix for the custom ones.

jonleivent:
> Because "accept" would serve two purposes at once, e.g. the wayland-minimal and wayland-screenshare examples.
Only one of them works so having a third option like "exclusive_accept" that could be used for the wayland-minimal example instead would solve that.

Got it.  This sounds like single inheritance.  Right now, your scheme has some implicit single inheritance, but maybe making it explicit would help.

With this, you can drop exclusive_accept as well.  The semantics would be that a block does what its parent does with either a few more accepts or rejects.  Maybe then move the accept/reject to the interface instead of the blockedProtocols, so you could mix accepts and rejects in the same block.  Maybe rename blockedProtocols to just protocols:

```
       <socket name="wayland">
          <protocols parent=all>
                <interface name="zwlr_virtual_pointer_manager_v1" mode="reject" />
                <interface name="zwp_virtual_keyboard_manager_v1" mode="reject" />
                <interface name="zwlr_export_dmabuf_manager_v1" mode="reject" />
                <interface name="zwlr_screencopy_manager_v1" mode="reject" />
                <interface name="zwlr_data_control_manager_v1" mode="reject" />
          </protocols>
      </socket>

        <socket name="wayland-base">
                <protocols parent="wayland">
                        <interface name="zwlr_input_inhibit_manager_v1" mode="reject" />
                        <interface name="zwlr_foreign_toplevel_manager_v1" mode="reject" />
                        <interface name="zwlr_layer_shell_v1"mode="reject"  />
                </protocols>
        </socket>

        <socket name="wayland-minimal">
                <protocols parent=none>
                        <interface name="wl_shm" mode="accept" />
                        <interface name="wl_compositor" mode="accept" />
                        <interface name="wl_subcompositor" mode="accept" />
                        <interface name="wl_seat" mode="accept" />
                        <interface name="xdg_wm_base" mode="accept" />

                        <interface name="wl_data_device_manager" mode="accept" />
                </protocols>
        </socket>

        <socket name="wayland-screenshare">
                <protocols parent="wayland">
                        <interface name="zwlr_export_dmabuf_manager_v1" mode="accept" />
                        <interface name="zwlr_screencopy_manager_v1" mode="accept" />
                </protocols>
        </socket>

PR #1005, opened 2023-07-30
Allow snap to window in MoveToEdge; Add GrowToEdge, ShrinkToEdge

digint:
I have been following your work for quite a while now (nice project, thanks a lot!), as I'm still stuck in X11 with openbox. For me there's just some small features which I'm painfully missing in order to control my window manager using the keyboard only.

A very common use case is to open a terminal "W-t", then align it to some other terminal using "W-arrow".

My rc.xml for this:
```
<!-- keyboard window placement -->
<keybind key="W-Left">
  <action name="MoveToEdge" direction="left" snap="window"/>
</keybind>
<keybind key="W-Right">
  <action name="MoveToEdge" direction="right" snap="window"/>
</keybind>
<keybind key="W-Up">
  <action name="MoveToEdge" direction="up" snap="window"/>
</keybind>
<keybind key="W-Down">
  <action name="MoveToEdge" direction="down" snap="window"/>
</keybind>
```

In addition to that, I want to align them (grow, shrink) to fill available space:
```
<keybind key="W-A-Left">
  <action name="ShrinkToEdge" direction="left"/>
</keybind>
<keybind key="W-A-Right">
  <action name="GrowToEdge" direction="right"/>
</keybind>
<keybind key="W-A-Up">
  <action name="ShrinkToEdge" direction="up"/>
</keybind>
<keybind key="W-A-Down">
  <action name="GrowToEdge" direction="down"/>
</keybind>
```

**This MR implements "GrowToEdge","ShrinkToEdge", and adds the ability to snap to next window for "MoveToEdge".**

Works nicely for me with the above config.

I will add some review comments in the code for some parts which I'm not sure about; especially we might consider adding a new compile unit for parts of the code.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -39,9 +39,14 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Move" />*
 	Begin interactive move of window under cursor
 
-*<action name="MoveToEdge" direction="value" />*
-	Move window to edge of outputs. Understands directions "left", "up",
-	"right" and "down".
+*<action name="MoveToEdge" direction="value" snapWindows="value" />*
+	Move window until it hits the next edge.
+
+	*direction* [left|up|right|down] Direction in which to move.
+
+	*snapWindows* [yes|no] Move window until it hits an edge of
+	another window or screen edge. If set to "no", only move to
+	the next screen edge. Default is yes.
 
 *<action name="Resize" />*
 	Begin interactive resize of window under cursor
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -56,6 +56,17 @@ Actions are used in menus and keyboard/mouse bindings.
 	top or bottom tell how much to resize on that edge of window,
 	positive values grow window, negative shrink window.
 
+*<action name="GrowToEdge" direction="value" />*
+	Resize window to fill the space between its edge and any other
+	window edge.
+
+	*direction* [left|up|right|down] Direction in which to grow.
+
+*<action name="ShrinkToEdge" direction="value" />*
+	Reverse of GrowToEdge. Shrinks by a maximum of 50%.
+
+	*direction* [left|up|right|down] Direction in which to shrink.
+
 *<action name="MoveTo" x="" y="" />*
 	Move to position (x, y)
 
--- a/include/snap.h
+++ b/include/snap.h
@@ -1,0 +1,17 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_SNAP_H
+#define LABWC_SNAP_H
+
+#include "common/border.h"
+#include "view.h"
+
+struct wlr_box;
+
+struct border snap_get_max_distance(struct view *view);
+
+void snap_vector_to_next_edge(struct view *view, enum view_edge direction, int *dx, int *dy);
+int snap_distance_to_next_edge(struct view *view, enum view_edge direction);
+void snap_grow_to_next_edge(struct view *view, enum view_edge direction, struct wlr_box *geo);
+void snap_shrink_to_next_edge(struct view *view, enum view_edge direction, struct wlr_box *geo);
+
+#endif /* LABWC_SNAP_H */
--- a/src/action.c
+++ b/src/action.c
@@ -266,6 +266,11 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 		}
 		break;
 	case ACTION_TYPE_MOVE_TO_EDGE:
+		if (!strcasecmp(argument, "snapWindows")) {
+			action_arg_add_bool(action, argument, parse_bool(content, true));
+			goto cleanup;
+		}
+		/* Falls through */
 	case ACTION_TYPE_SNAP_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
 			enum view_edge edge = view_edge_parse(content);
@@ -638,7 +643,8 @@ actions_run(struct view *activator, struct server *server,
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
-				view_move_to_edge(view, edge);
+				bool snap_to_windows = action_get_bool(action, "snapWindows", true);
+				view_move_to_edge(view, edge, snap_to_windows);
 			}
 			break;
 		case ACTION_TYPE_SNAP_TO_EDGE:
--- a/src/action.c
+++ b/src/action.c
@@ -66,6 +66,8 @@ enum action_type {
 	ACTION_TYPE_EXIT,
 	ACTION_TYPE_MOVE_TO_EDGE,
 	ACTION_TYPE_SNAP_TO_EDGE,
+	ACTION_TYPE_GROW_TO_EDGE,
+	ACTION_TYPE_SHRINK_TO_EDGE,
 	ACTION_TYPE_NEXT_WINDOW,
 	ACTION_TYPE_PREVIOUS_WINDOW,
 	ACTION_TYPE_RECONFIGURE,
@@ -105,6 +107,8 @@ const char *action_names[] = {
 	"Exit",
 	"MoveToEdge",
 	"SnapToEdge",
+	"GrowToEdge",
+	"ShrinkToEdge",
 	"NextWindow",
 	"PreviousWindow",
 	"Reconfigure",
@@ -272,6 +276,8 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 		}
 		/* Falls through */
 	case ACTION_TYPE_SNAP_TO_EDGE:
+	case ACTION_TYPE_GROW_TO_EDGE:
+	case ACTION_TYPE_SHRINK_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
 			enum view_edge edge = view_edge_parse(content);
 			if ((edge == VIEW_EDGE_CENTER && action->type != ACTION_TYPE_SNAP_TO_EDGE)
@@ -411,6 +417,8 @@ action_is_valid(struct action *action)
 		break;
 	case ACTION_TYPE_MOVE_TO_EDGE:
 	case ACTION_TYPE_SNAP_TO_EDGE:
+	case ACTION_TYPE_GROW_TO_EDGE:
+	case ACTION_TYPE_SHRINK_TO_EDGE:
 		arg_name = "direction";
 		arg_type = LAB_ACTION_ARG_INT;
 		break;
@@ -652,6 +660,20 @@ actions_run(struct view *activator, struct server *server,
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
 				view_snap_to_edge(view, edge, /*store_natural_geometry*/ true);
+			}
+			break;
+		case ACTION_TYPE_GROW_TO_EDGE:
+			if (view) {
+				/* Config parsing makes sure that direction is a valid direction */
+				enum view_edge edge = action_get_int(action, "direction", 0);
+				view_grow_to_edge(view, edge);
+			}
+			break;
+		case ACTION_TYPE_SHRINK_TO_EDGE:
+			if (view) {
+				/* Config parsing makes sure that direction is a valid direction */
+				enum view_edge edge = action_get_int(action, "direction", 0);
+				view_shrink_to_edge(view, edge);
 			}
 			break;
 		case ACTION_TYPE_NEXT_WINDOW:
--- a/src/snap.c
+++ b/src/snap.c
@@ -1,0 +1,279 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <strings.h>
+#include <wlr/util/box.h>
+#include "labwc.h"
+#include "snap.h"
+#include "view.h"
+#include "workspaces.h"
+
+/* We cannot use MIN/MAX macros, as they may call functions twice, and
+ * can be overridden by previous #define.
+ */
+static inline int
+min(int a, int b) {
+	return a < b ? a : b;
+}
+
+static inline int
+max(int a, int b) {
+	return a > b ? a : b;
+}
+
+static inline int
+min3(int a, int b, int c) {
+	return min(min(a, b), c);
+}
+
+enum snap_mode {
+	SNAP_MODE_MOVE = 0,
+	SNAP_MODE_GROW,
+	SNAP_MODE_SHRINK,
+};
+
+static struct border
+snap_get_view_edge(struct view *view)
+{
+	struct border margin = ssd_get_margin(view->ssd);
+	struct border edge = {
+		.left   = view->pending.x - margin.left,
+		.top    = view->pending.y - margin.top,
+		.right  = view->pending.x + view->pending.width  + margin.right,
+		.bottom = view->pending.y + view->pending.height + margin.bottom
+	};
+	return edge;
+}
+
+struct border
+snap_get_max_distance(struct view *view)
+{
+	struct output *output = view->output;
+	struct border margin = ssd_get_margin(view->ssd);
+	struct wlr_box usable = output_usable_area_scaled(output);
+	struct border distance = {
+		.left   = usable.x + margin.left + rc.gap - view->pending.x,
+		.top    = usable.y + margin.top  + rc.gap - view->pending.y,
+		.right  = usable.x + usable.width  - view->pending.width
+			- margin.right  - rc.gap - view->pending.x,
+		.bottom = usable.y + usable.height - view->pending.height
+			- margin.bottom - rc.gap - view->pending.y
+	};
+	return distance;
+}
+
+struct snap_search {
+	const int search_dir; /* -1: left/up, 1: right/down */
+
+	const int add_view_x;
+	const int add_view_y;
+	const int add_view_width;
+	const int add_view_height;
+
+	const int add_margin_left;
+	const int add_margin_top;
+	const int add_margin_right;
+	const int add_margin_bottom;
+};
+
+/* near/far is the left, right, top or bottom border of a window,
+ * depending on the search direction:
+ *  - near_right: search to the right, snap to left  (near) border of a window.
+ *  - far_right:  search to the right, snap to right (far)  border of a window.
+ *  - near_left:  search to the left,  snap to right (near) border of a window.
+ *  - far_left:   search to the left,  snap to left  (far)  border of a window.
+ *
+ * structs below define what coordinates and margins to take into
+ * account depending near/far, and direction.
+ */
+static const struct snap_search near_left  = { -1,   1, 0, 1, 0,    0,  0,  1,  0 };
+static const struct snap_search near_up    = { -1,   0, 1, 0, 1,    0,  0,  0,  1 };
+static const struct snap_search near_right = {  1,   1, 0, 0, 0,   -1,  0,  0,  0 };
+static const struct snap_search near_down  = {  1,   0, 1, 0, 0,    0, -1,  0,  0 };
+static const struct snap_search far_left   = { -1,   1, 0, 0, 0,   -1,  0,  0,  0 };
+static const struct snap_search far_up     = { -1,   0, 1, 0, 0,    0, -1,  0,  0 };
+static const struct snap_search far_right  = {  1,   1, 0, 1, 0,    0,  0,  1,  0 };
+static const struct snap_search far_down   = {  1,   0, 1, 0, 1,    0,  0,  1,  0 };
+
+static inline int
+_snap_next_edge(struct view *view, int start_pos, const struct snap_search def, int max, int gap)
+{
+	struct output *output = view->output;
+	struct server *server = output->server;
+	struct view *v;
+	int p = max;
+	for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+		if (v == view || v->output != output || v->minimized || v->maximized) {
+			continue;
+		}
+
+		struct border margin = ssd_get_margin(v->ssd);
+		int vp = -start_pos;
+		vp += def.add_margin_left   * margin.left;
+		vp += def.add_margin_top    * margin.top;
+		vp += def.add_margin_right  * margin.right;
+		vp += def.add_margin_bottom * margin.bottom;
+		vp += def.add_view_x        * v->pending.x;
+		vp += def.add_view_y        * v->pending.y;
+		vp += def.add_view_width    * v->pending.width;
+		vp += def.add_view_height   * v->pending.height;
+		vp += gap;
+
+		if (def.search_dir * vp > 0 && def.search_dir * (vp - p) < 0) {
+			p = vp;
+		}
+	}
+	return p;
+}
+
+static void
+_snap_move_resize_to_edge(struct view *view, enum view_edge direction, enum snap_mode mode,
+		struct wlr_box *delta)
+{
+	struct border edge = snap_get_view_edge(view);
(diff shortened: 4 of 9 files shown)


src/view.c
```diff
@@ -54,6 +54,55 @@ view_edge_invert(enum view_edge edge)
...
+	}
+	if (usable.width == output->wlr_output->width
+			&& output->wlr_output->scale != 1) {
+		usable.width /= output->wlr_output->scale;
+	}
+
+	struct edge_manip_hints hints = {
+		.left_edge    = view->current.x - margin.left,
+		.top_edge     = view->current.y - margin.top,
+		.right_edge   = view->current.x + view->pending.width  + margin.right,
```
digint, line 89: I'm not 100% sure if I got the semantics of "pending" vs. "current" right. At what point are those values updated?

src/view.c
```diff
@@ -857,51 +906,250 @@ view_on_output_destroy(struct view *view)
...
+	 * Returns near and far edges from all views, looking from
+	 * point towards direction. Stops searching at max.
+	 *
+	 * Note that the width and height are ignored when looking for
+	 * edges (left/right only considers x, while up/down only
+	 * considers y). This is intentional as it allows to easily
+	 * align your windows to a grid, defined by all other windows)
+	 */
+	int n = max, f = max;
+	if (!strcasecmp(direction, "left")) {
```
digint, line 947: I should probably use something like `enum view_edge` and `view_edge_parse()` for this, but sadly this includes `VIEW_EDGE_CENTER`. I think this should be renamed (there is no center edge), and/or combined somehow.

src/view.c
```diff
@@ -857,51 +906,250 @@ view_on_output_destroy(struct view *view)
 	view->output = NULL;
 }
 
+#define VIEW_GET_NEXT_EDGES(_view_near_def, _view_far_def, _backward)	\
+	struct output *output = view->output;				\
+	struct server *server = output->server;				\
+	struct view *v;							\
+	wl_list_for_each(v, &server->views, link) {			\
+		if (v == view || v->output != output)			\
+			continue;					\
```
digint, line 915: There might be some more views to exclude here?

johanmalm:
Nice patch-set. Thanks!

I like this in principle, but it deserves a proper code review. I probably won't get a chance until mid-week, although @Consolatis  might earlier.

Interestingly, Openbox's MoveToEdge action does always respect other windows by the looks of it (not tested). http://openbox.org/wiki/Help:Actions#Introduction. I think it's a good idea to make this optional (like you've done). @Consolatis and I tend to discuss at length what we call things and what default values ought to be, so hope you don't mind if we end up doing that on this point :smile: Initial thought on `snap="window"` is that it would feel cleaner to do a boolean like 'ignoreWindows="no"`.

Regarding pending/current - it's a bit mind boggling at times, but I think I'm right in saying that pending is the one to use here.

A couple of quick high-level points:

- Suggest better name for `view_get_edge_manip_hints()`. Sorry my brain is a bit tired right now - that edge_manip struct is quite hard to follow. Is there a simpler way of cutting this up? Could the logic be simplified if we used `include/common/border.h` (as a better alternative than `wlr_box` for describing edges). Would `ssd_max_extents()` help reduce noise here?
- Would be help readability if we replace `right_max_dx`, `width_max_dx`, etc by function calls whenever they are needed. Yes, slightly more expensive, but if it simplifies code, I'd gladly accept that.

- I prefer to avoid the `VIEW_GET_NEXT_EDGES` macro if we can (because it makes AddressSanitizer harder to use). The the looks of it, it easily be replaced by a function.

> There might be some more views to exclude here?

Yes, @Consolatis and I were just talking about this last week relating the windows in the `window-switcher` (`osd.c`). We ought to generate window lists centrally to avoid duplication. For example you probably wouldn't want to take into account minimized windows, those that have never been mapped, and so on. We can do that as a separate PR though.

johanmalm:
Just one more thought... could the logic be simplified with a `wl_array *get_snap_points(enum foo direction)` (which just iterates over windows and get the right edges depending on direction). Could also include the window edges.
If we then simply sort the array, it'd be simple to just pick the right point. That way we remove the near/far thing.

digint:
> Interestingly, Openbox's MoveToEdge action does always respect other windows by the looks of it

It does so, that's why I changed the default to also snap to windows.

> Initial thought on snap="window" is that it would feel cleaner to do a boolean like 'ignoreWindows="no"`.

Makes sense, unless you might want to snap to something different in the future.

> Suggest better name for view_get_edge_manip_hints().

Yes. Would that make sense? Split into:
 - `struct border view_get_edges()`, holding left_edge, top_edge, right_edge, bottom_edge
 - `struct border_resize_usable view_get_resize_usable()` for the other 6 vars (still not very happy with this naming though)

> Would be help readability if we replace right_max_dx, width_max_dx, etc by function calls whenever they are needed. Yes, slightly more expensive, but if it simplifies code, I'd gladly accept that.

Maybe, yes. Probably not even more expensive if we hint the compiler with "inline".
Although we should make some function which wraps this into some helper function. Something like: `output_usable_area_in_layout_coords_scaled()` or just `output_usable_area_scaled()`:
```
	struct wlr_box usable = output_usable_area_in_layout_coords(output);
	if (usable.height == output->wlr_output->height
			&& output->wlr_output->scale != 1) {
		usable.height /= output->wlr_output->scale;
	}
	if (usable.width == output->wlr_output->width
			&& output->wlr_output->scale != 1) {
		usable.width /= output->wlr_output->scale;
	}
```

> I prefer to avoid the VIEW_GET_NEXT_EDGES macro if we can (because it makes AddressSanitizer harder to use). The the looks of it, it easily be replaced by a function.

Hmm there was a real reason for a macro at some point, but I can't find / recall it any more (I've had a hard time getting those gaps right...). I will try to replace this by function if possible.

> >  There might be some more views to exclude here?
>
> Yes, @Consolatis and I were just talking about this last week relating the windows in the window-switcher (osd.c). We ought to generate window lists centrally to avoid duplication. For example you probably wouldn't want to take into account minimized windows, those that have never been mapped, and so on. We can do that as a separate PR though.

I think a separate PR makes sense here. Probably  (when growing or shrinking)add wrappers around `wl_list_for_each` ?

> Just one more thought... could the logic be simplified with a wl_array *get_snap_points(enum foo direction) (which just iterates over windows and get the right edges depending on direction). Could also include the window edges. If we then simply sort the array, it'd be simple to just pick the right point. That way we remove the near/far thing.

Not sure if this is possible: the problem is that for growing or shrinking the rc.gap needs to be added to different edges. Your proposal would then probably end up in two functions `get_snap_points_for growing()` and `get_snap_points_for_shrinking()`.

digint:
In the meantime I have been playing around with replacing the macro. This is work-in-progress, but illustrates the problem: we need to specify what to add from the the views INSIDE the iterator:

https://github.com/digint/labwc/pull/new/snap-to-edge-replace-macro

Let me know if you would like me to rebase the code-style fixes (found out about scripts/check ;-)
Usually I try not to rebase things while conversation/review is ongoing.

src/view.c
Consolatis: Pending: whenever we change something
Current:
- when we apply the change ourselves (e.g. wayland native windows for a position only change)
- we get the confirmation of the window (e.g. wayland commit handler for new sizes).

Basically pending is always what the window "should" be and current is what labwc thinks it is right now (which may be outdated).

src/view.c
Consolatis: > I should probably use something like `enum view_edge` and `view_edge_parse()` for this, but sadly this includes `VIEW_EDGE_CENTER`. I think this should be renamed (there is no center edge), and/or combined somehow.

While I logically agree on the `VIEW_EDGE_CENTER` part I don't see that as being an issue in practice here.
It would be the same as having a direction "foobar" in the config.

Thus, suggesting we indeed use `enum view_edge` when parsing the config so we can use a simple `switch()` here.
I am not sure if we should add a new `action_arg` type though or if we just use the existing `int` one.

src/view.c
```diff
@@ -857,51 +907,252 @@ view_on_output_destroy(struct view *view)
...
+			v->current.y - margin.top,
+			v->current.y + v->current.height + margin.bottom,
+			false);
+	}
+	*near = n; *far = f;
+}
+
 void
-view_move_to_edge(struct view *view, const char *direction)
+view_move_to_edge(struct view *view, const char *direction, const char *snap)
```
Consolatis, line 972: At the current point in time I'd suggesting just using a `bool snap_to_window` argument here, if we ever end up supporting more (?) snap targets we can always change the internal API as long as the config format keeps being the same.

src/view.c
```diff
@@ -857,51 +907,252 @@ view_on_output_destroy(struct view *view)
...
-	int x = 0, y = 0;
+	view_move_resize(view, geo);
+}
+
+void
+view_shrink_to_edge(struct view *view, const char *direction)
+{
+	assert(view);
+	if (view->fullscreen || view->maximized)
+		return;
```
Consolatis, line 1105: Please always use curly braces for if statements (and for loops, ..): It is way too easy to add another line to the body during debugging or similar which then can blow up everything.

@johanmalm 
We should really try to get this into checkpatch.pl as well so the CI would warn about it.

src/action.c
```diff
@@ -179,7 +183,14 @@ action_arg_from_xml_node(struct action *action, char *nodename, char *content)
 		}
 		break;
 	case ACTION_TYPE_MOVE_TO_EDGE:
+		if (!strcmp(argument, "snap")) {
+			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
+		/* Falls through */
```
Consolatis, line 190: Similar here to the `view_edge` enum, I think we should use a bool argument here. Thus parsing it here instead of all over again later in the code.

Consolatis:
Thanks for this PR!
Having a snap-to-window functionality was always at the back of my mind.

I just had a quick look trough the code and added some comments but didn't yet look into the actual calculations nor tested it but will do so in the near future. We might also want to think about allowing `resistance.c` to reuse parts of the window detection logic based on a future config option. That doesn't mean that this PR has to already implement it but I think its worth keeping that use-case in the back of the mind.

src/view.c
Consolatis: Yes, at least
- `view->minimized`
- `view->scene_tree->node.parent != view->server->workspace_current->tree && !view_is_always_on_top(view)`

digint: I also exclude `view->minimized || view->maximized` in 0e27cb510f100fd0943ba9a40f8ae0a64c699d2a.

Why would we want to ignore the always-on-top windows? I would rather exclude always-on-bottom, no?

```
view->scene_tree->node.parent != view->server->workspace_current->tree && !view_is_always_on_top(view)
```
^^^ This statement does not work (and I don't understand what  it is doing, leaving it out for now :-/

digint:
I have quite massively reworked the code, especially:
  - Move edge calculation framework to separate compile unit
  - Ignore more views
  - Improve code quality and naming
  - Remove macros

I will wait before doing any changes regarding `view_edge_parse` and `bool snap_to_window` proposal, as @Consolatis will do some changes for the existing MoveToEdge. This is a good thing, I think we should not need to parse strings in view.c (outside of parsing xml).

src/view.c
digint: Fixed in: 75c2d4c43bac1eaa07c96e037118d170e64f2a52 (hoping I caught all of them)

src/view.c
digint: This has moved to `snap_get_view_edge` and `snap_get_max_distance`:

https://github.com/digint/labwc/blob/441eb9031d9541beff60d740d6261c088dde0bd8/src/snap.c#L32

> So in this case for calculating the "to-be-set" state we should use ->pending for everything

Thanks for the explanation.

So you are saying I should change `view->current.x / y` as well? Will it not be confusing if we snap from something which is not "current == what labwc thinks it is right now" rather than some (virtual, to-be-changed in the future) window position? Does that mean that "current" is not always what I see on my screen?

I will gladly change it if you insist (as I'm still a bit puzzeled about this and don't fully understand how things are working under the hood)

src/view.c
Consolatis: > Why would we want to ignore the always-on-top windows? I would rather exclude always-on-bottom, no?

Windows should be ignored that are *neither*
- part of the current workspace
- part of the always-on-top tree

If one of these conditions is true (and the other criteria matches, e.g. not minimized, ..) the windows should be used.
Basically this is for ignoring windows that are on another workspace.

src/view.c
Consolatis: Done:
- #1007

src/view.c
Consolatis: I'd say that in the unlikely case that there is indeed some pending movement / resize for a destination window (e.g. a window we want to snap to) that its nicer to see the new window go to the place where it should be and then having the target window adjust afterwards due to the pending commit than to position the window incorrectly once the pending commit comes through.

johanmalm:
I like the look of this now. Happy to go with it except for a few nitpicks below.

I tried the config in the original post and it makes for a really nice workflow :smile: I think I might adopt it.

Review feedback:

- [ ] Use `pending` throughout
- [ ] Change `char *direction` to new enum (just merged PR from @Consolatis)
- [ ] In `snap.h`, only include prototypes, type/enum declarations and documentation. So change `static inline foo` to `static foo` and move to snap.c
- [ ] Also remove `inline` in snap.c. Not needed.
- [ ] Change // style comments to /* */
- [ ] We generally don't use more than one space between identifiers and +/-/= but happy to ignore that for the benefits of easier reading (through alignment).
- [ ] Just picking up on comments from @Consolatis, we should not snap to windows in other workspaces. I've just tried moving a few windows to a different workspace, and it makes for a weird experience. I think that `view->scene_tree->node.parent != view->server->workspace_current->tree` is the test to use to avoid this. Do should if it doesn't work and I'll have a look.

digint:
Thanks, I will do the changes in a couple of days, very busy at the moment.

> I tried the config in the original post and it makes for a really nice workflow smile I think I might adopt it.

Heh yes, I can't live any more without it ;-)

> Also remove inline in snap.c. Not needed

Even if compilers will probably optimize the code anyways, It is usually a good idea to give it some hints (e.g. many of the `+=` statements should get optimized out, depending on `snap_search`). But I don't mind to remove it.

src/snap.c
```diff
@@ -0,0 +1,284 @@
...
+	struct wlr_scene_tree *current_workspace_tree = view->server->workspace_current->tree;
+	wl_list_for_each(v, &server->views, link) {
+		if (v == view || v->output != output || v->minimized || v->maximized) {
+			continue;
+		}
+		/* not part of the current workspace, and not part of the always-on-top tree */
+		if (v->scene_tree->node.parent != current_workspace_tree &&
+			!view_is_always_on_top(v)) {
+			continue;
+		}
```
digint, line 112: Checking workspaces, as suggested @Consolatis here: https://github.com/labwc/labwc/pull/1005#discussion_r1281295788

Works for me, please double-check.

src/view.c
digint: Fixed, see: https://github.com/labwc/labwc/pull/1005/files#r1285129442

digint:
I have rebased my work, along with requested changes.

Open points: 

- `snap="window/screen"` vs. `ignoreWindows="no"`

  Not sure if we have a consensus on that. I personally prefer
  the first, as it is more descriptive (one can argue about
  that), and does not use negative logic (in fact this would be
  the first attribute in labwc named "ignoreSomething".

- [ ] Also remove inline in snap.c. Not needed.
   
   Not changed yet, see comment above. At a minimum we should keep
inline on min/max functions.

> Just picking up on comments from @Consolatis, we should not
  snap to windows in other workspaces. I've just tried moving a
  few windows to a different workspace, and it makes for a weird
  experience.

I have added the workspace checking, please double-check:

https://github.com/labwc/labwc/pull/1005/files#r1285129442

src/snap.c
Consolatis: Yep, seems to work fine. We should provide a helper function like `view_is_in_viewport()` or something along that line later on but we don't have to do that in this PR.

Consolatis:
> edit: oh and on top of that, I added my config to rc.xml.all in [52314a7](https://github.com/labwc/labwc/commit/52314a7faedc5e8cd6f96608dc42f26f68e74ee5)

`rc.xml.all` should represent the exact same config as what is the hardcoded fallback when no config file has been found.
In this case this would be out of sync. I am also not sure if we want to have those as default bindings either. The existing `MoveToEdge` ones are bad enough already as it prevents using `A-Left/Right` for browsers or file managers. Adding more like that to the default config doesn't seem like a good idea to me. But the last say on that belongs to @johanmalm.

In either case, we have to either remove them from `rc.xml.all` or also add them to `load_default_key_bindings()` (which I btw broke for `MoveToEdge` and `SnapToEdge` with the enum change, just recognized that. Will fix in a new PR!) .

digint:
> rc.xml.all should represent the exact same config as what is the hardcoded fallback when no config file has been found.

I was not aware of that, in this case it should not be part of the MR in my opinion (we can still change defaults later) -> Dropped commit.

In my opinion the defaults should be kept as small-but-still-useful as possible (e.g. many users like myself instantly disable defaults when seeing "A-F4" in the defaults, this just reminds me of painful days :laughing: )

include/snap.h
```diff
@@ -0,0 +1,18 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_SNAP_H
+#define LABWC_SNAP_H
+
+#include "config.h"
+#include <stdint.h>
+#include <wayland-util.h>
+#include <wlr/util/box.h>
```
Consolatis, line 8: `<wayland-util.h>` and `<wlr/util/box.h>` can be removed.
To make it consistent and not rely on `view.h` including it, we can instead forward declare `struct wlr_box`.
We should however additionally include `common/border.h` here.

src/snap.c
```diff
@@ -0,0 +1,284 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <strings.h>
+#include "labwc.h"
+#include "workspaces.h"
+#include "view.h"
+#include "snap.h"
```
Consolatis, line 7: Nitpick: alphabetic ordering for our includes would be nice.

src/snap.c
Consolatis, line 112: Please use double indents for multi line ifs, example:
```c
		if (v->scene_tree->node.parent != current_workspace_tree
				&& !view_is_always_on_top(v)) {
			continue;
		}
```

src/view.c
```diff
@@ -8,13 +8,12 @@
 #include "menu/menu.h"
 #include "regions.h"
 #include "ssd.h"
+#include "snap.h"
```
Consolatis, line 11: Nitpick: alphabetic ordering

Consolatis:
> Open points:
> 

>> * [ ]  Also remove inline in snap.c. Not needed.
>   Not changed yet, see comment above. At a minimum we should keep
>   inline on min/max functions.

Usually my stance is: interfere with compiler optimizations as less as possible and let the compiler do its job.
The best help for that might be to have clear code that can easily be detected as optimization target (which seems to be the case here). However, in this case I personally don't really mind either variant as its pretty much contained within a single file.

I'd still argue that at least `GrowToEdge` should additionally include the optional `snap` / `ignoreWindow` argument so users are able to grow a view until it hits the next screen edge, regardless of any other windows in between. I am not sure right now if the same would make sense for `ShrinkToEdge`. However I'd be fine to delay this to the future as an existing config wouldn't break by adding that argument.

docs/labwc-actions.5.scd
```diff
@@ -50,6 +54,17 @@ Actions are used in menus and keyboard/mouse bindings.
 	top or bottom tell how much to resize on that edge of window,
 	positive values grow window, negative shrink window.
 
+*<action name="GrowToEdge" direction="value" />*
+	Resize window to fill the space between its edge and any other
+	window edge.
+
+	*direction* [left|up|right|down] Direction in which to grow.
```
Consolatis, line 61: I'd still argue that at least `GrowToEdge` should additionally include the optional `snap` / `ignoreWindow` argument so users are able to grow a view until it hits the next screen edge, regardless of any other windows in between. I am not sure right now if the same would make sense for `ShrinkToEdge`. However I'd be fine to delay this to the future as an existing config wouldn't break by adding that argument.

johanmalm:
I know we're polishing now, but if you're happy to bear with us for a bit longer :smile:

- [ ] My only reservation before merge is the `snap` attribute.

Unless I'm misreading it, the logic goes like this:

a. `snap="window"` snaps to window + screen
b. `snap="screen"` snaps to screen
c. `snap="bad"` snaps to screen

So if we're always snapping to screen, it seems counter-intuitive that the user has to set `snap="screen"` to ignore window edges.

It is conceivable that we might want to extend this in future to include all outputs in the output-layout (=desktop) (like Openbox does IIRC). At that point we'd probably want a yes/no logic for each of at least `window` and `desktop`. I do therefore think it's clearer and more extensible to have boolean attributes/elements for these. If we keep the `snap="window|desktop"` approach we'd end up doing a string-split on the content anyway.

I suggest: `<action name="MoveToEdge" direction="" snapWindow="yes|no" />`

Assuming that output edges will always be included (can't think of a use-case for excluding) - a future state which allows the whole desktop (multiple outputs) could be covered by a simple addition of `wholeDesktop="yes"` or something like that.

- Happy to accept `GrowToEdge` without `snap` / `snapWindow` attribute. Agree with @Consolatis though that it would be a future extension.

- Regarding `inline` - I'm fine as you've done it.

- Agree with your (both of you) stance on keybind-defaults. Let's not add anymore :smile:

- Agree with remainder of @Consolatis's nitpiks from last night.

Consolatis:
> I suggest: `<action name="MoveToEdge" direction="" snapWindow="yes|no" />`

Not sure about using singular or plural, e.g. `snapWindow` vs `snapWindows` but I'd be fine with both.

johanmalm:
> > I suggest: `<action name="MoveToEdge" direction="" snapWindow="yes|no" />`
> 
> +1. Not sure about using singular or plural, e.g. `snapWindow` vs `snapWindows` but I'd be fine with both.

Yes, good point. Plural better.

johanmalm:
If you rebase against master, you can now use the new `for_each_view()` API.

So, instead of:

```
wl_list_for_each(v, &server->views, link) {
	if (v == view || v->output != output || v->minimized || v->maximized) {
		continue;
	}
	/* not part of the current workspace, and not part of the always-on-top tree */
	if (v->scene_tree->node.parent != current_workspace_tree &&
		!view_is_always_on_top(v)) {
		continue;
	}
...
```

We can use:

```
for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
	if (v == view || v->output != output || v->minimized || v->maximized) {
		continue;
	}
...
```

Consolatis:
In https://github.com/Consolatis/labwc/commits/pr/1005_grow_to_edge I've added fixup commits for the outstanding issues raised in this thread, I think I caught all the review comments. @digint could either use the fixes or I could do the squashing and force push in this PR.

I didn't add the additional `snapWindows` argument to `GrowToEdge`, I did use the new `for_each_view()` macro though.

docs/labwc-actions.5.scd
johanmalm: Later.

src/view.c
johanmalm: Done (your [fixup])

src/snap.c
johanmalm: Done (your [fixup])

src/snap.c
johanmalm: Done (your [fixup])

include/snap.h
johanmalm: Done (your [fixup])

Consolatis:
LGTM. Thanks @digint for all the work and @johanmalm for the rebase :)

johanmalm:
Thanks

PR #1006, opened 2023-08-01
rc, layers: Add ignoreExclusiveArea

Joshua-Ashton:
Adds a config option to ignore the exclusive area for wl_layer clients such as waybar, allowing windows and their titlebars to appear in the same space.


Consolatis:
I don't really understand the purpose of this PR: waybar could just be configured to not use the exclusive zone?

Joshua-Ashton:
I did not know it had that option :frog: oops

PR #1007, opened 2023-08-02
actions: use enum for _ToEdge actions

Consolatis:
This also improves the config robustness as invalid edge names will now prevent the action to be created in the first place and the user gets notified about the issue.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -181,7 +181,14 @@ action_arg_from_xml_node(struct action *action, char *nodename, char *content)
 	case ACTION_TYPE_MOVE_TO_EDGE:
 	case ACTION_TYPE_SNAP_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
-			action_arg_add_str(action, argument, content);
+			enum view_edge edge = view_edge_parse(content);
+			if ((edge == VIEW_EDGE_CENTER && action->type != ACTION_TYPE_SNAP_TO_EDGE)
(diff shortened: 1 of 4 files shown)


src/action.c
```diff
@@ -181,7 +181,14 @@ action_arg_from_xml_node(struct action *action, char *nodename, char *content)
...
 	case ACTION_TYPE_SNAP_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
-			action_arg_add_str(action, argument, content);
+			enum view_edge edge = view_edge_parse(content);
+			if ((edge == VIEW_EDGE_CENTER && action->type != ACTION_TYPE_SNAP_TO_EDGE)
+					|| edge == VIEW_EDGE_INVALID) {
+				wlr_log(WLR_ERROR, "Invalid argument for action %s: '%s' (%s)",
+					action_names[action->type], argument, content);
+			} else {
+				action_arg_add_int(action, argument, edge);
```
Consolatis, line 190: We are simply reusing the `int` argument storage type here for the enum. The first enum value is initialized with `0` which, when my memory serves me right, makes all following enum entries `+= 1` and thus using `int` for those 6 enum entries should be fine (rather than using a new `uint32_t` or even `enum view_edge` arg storage type).

I am open for changing this though if @johanmalm prefers.

johanmalm: Looks good.

PR #1012, opened 2023-08-05
view: only add foreign-handle for root toplevel

johanmalm:
Welcome thoughts / review.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -384,6 +384,22 @@ xdg_toplevel_view_get_string_prop(struct view *view, const char *prop)
 }
 
 static void
+init_foreign_toplevel(struct view *view)
+{
+	foreign_toplevel_handle_create(view);
+	struct wlr_xdg_toplevel *toplevel = xdg_toplevel_from_view(view);
+	if (!toplevel->parent) {
+		return;
+	}
+	struct wlr_xdg_surface *surface = toplevel->parent->base;
+	struct view *parent = surface->data;
+	if (!parent->toplevel.handle) {
+		return;
+	}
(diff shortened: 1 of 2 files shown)


Consolatis:
I think we should set the parent to the toplevel handle of the parent instead. Then it would be panel policy if and how they show handles that have a parent (e.g. grouping them together, hiding them, showing them as every other handle).

johanmalm:
> I think we should set the parent to the toplevel handle of the parent instead. Then it would be panel policy if and how they show handles that have a parent (e.g. grouping them together, hiding them, showing them like any other handle).

I couldn't work out how to do it. :blush:

Consolatis:
Shouldn't it just be something like this? Not tested at all.
```c
/* Copied from xdg_toplevel_view_get_root() */

struct wlr_xdg_toplevel *toplevel = xdg_toplevel_from_view(view);
if (toplevel->parent) {
	struct wlr_xdg_surface *surface = toplevel->parent->base;
	struct view *parent = surface->data;
	if (parent->toplevel.handle) {
		wlr_foreign_toplevel_handle_v1_set_parent(
			view->toplevel.handle, parent->toplevel.handle);
	}
}
```

johanmalm:
@Consolatis Thanks for that tip. Worked a treat.
Ready to go.

Consolatis:
I'd like to look into the lifecycle management before merging this to make sure we don't accidentally leave a dangling pointer hanging somewhere in wlroots. Will likely have done that until tomorrow.

Consolatis:
When a toplevel is destroyed, wlroots loops over all toplevels and resets the parent to `NULL` if it is currently set to the destroying toplevel (and informs the clients about it). So this should be safe. Thanks for the PR :)

PR #1014, opened 2023-08-05
view: make move_sub_views() use append_children method

johanmalm:
...to share common code with minimize_sub_views()

Helped-by: @Consolatis

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -330,49 +330,6 @@ top_parent_of(struct view *view)
 	return toplevel;
 }
 
-enum z_direction {
-	LAB_TO_FRONT,
-	LAB_TO_BACK,
-};
-
-/*
- * TODO: Combine append_children() and move_sub_views() as much as possible.
- * https://github.com/labwc/labwc/pull/998#discussion_r1284085575
- */
-static void
-move_sub_views(struct view *parent, enum z_direction z_direction)
-{
-	assert(parent);
-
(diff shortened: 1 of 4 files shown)


src/xdg.c
```diff
@@ -393,15 +350,15 @@ xdg_toplevel_view_move_to_front(struct view *view)
...
+	view_impl_move_sub_views(root, LAB_TO_FRONT);
 }
 
 static void
 xdg_toplevel_view_move_to_back(struct view *view)
 {
 	struct view *root = xdg_toplevel_view_get_root(view);
 	view_impl_move_to_back(root);
-	move_sub_views(root, LAB_TO_BACK);
+	view_impl_move_sub_views(root, LAB_TO_BACK);
```
Consolatis, line 361: Should we switch these two around for the `_to_back()` case, e.g. first `move_sub_views()` and then `move_to_back(root)`?
Situation: dialogs on top of toplevel.

Currently:
- toplevel to the back
- dialog to the back
- result: dialog behind toplevel

It will fix itself on unminimize but might mess up A-Tab ordering and similar.

johanmalm:
Good catch. I have tested with the `Lower` action. It in fact only works properly after the fix you proposed!! Nice one :smile:

I have also moved `view_impl_move_sub_views` a bit higher up in `view-impl-common.c` because I think it looks better with it being close to `...move_to_{back,front}`

Ready to go.

PR #1018, opened 2023-08-08
[incomplete] Add ToggleShade

Consolatis:
Just an initial implementation, I am not sure when I'll work on this again. The alternative would have been to have it bitrot in a local dev branch so I thought I may just as well post it as [incomplete] PR on here.

For anybody interested, feel free to use this as a base or open an alternative PR.

Multiple issues that have to be fixed or verified:
- [ ] Interaction with various actions that resize the window or return it into its natural geometry (e.g. unmaximize / untile)
  - [ ] Maximize
  - [ ] SnapToEdge
  - [ ] SnapToRegion
  - [ ] ResizeRelative
  - [ ] GrowToEdge (once merged)
- [ ] Interaction with ToggleDecorations (and Reconfigure) is missing, we should check `view->shaded` on ssd creation
- [ ] Not verified with xwayland windows at all
- [ ] The titlebar misses its top border as we disable the whole border tree

And the usual stuff missing:
- [ ] docs
- [ ] default mouse bindings?

Interactive resizing is completely disabled while in the shaded state. Reason being that the actual client doesn't commit anything if it's scene node is disabled. Moving the title bar around is fine though as we don't need any confirmation from the client for that.


Consolatis:
Replaced by
- #1430

PR #1020, opened 2023-08-09
cursor: cancel popup grab on mouse-press outside client itself

johanmalm:
...for example on any part of the server side decoration

Supercedes #826 for xdg-shell popups

Helped-by: @Consolatis

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -923,8 +923,15 @@ cursor_button_press(struct seat *seat, struct wlr_pointer_button_event *event)
 		}
 	}
 
-	if (ctx.type == LAB_SSD_ROOT && wlr_seat_pointer_has_grab(seat->seat)) {
-		/* Let a click on the desktop cancel an active popup grab */
+	if (ctx.type != LAB_SSD_CLIENT && wlr_seat_pointer_has_grab(seat->seat)) {
+		/*
+		 * If we have an active popup grab (an open popup) we want to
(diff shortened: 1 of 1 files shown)


01micko:
Seems to work fine.

Consolatis:
I am not sure if we have to also check for `LAB_LAYER_SURFACE`, @01micko any different behavior when clicking somewhere on sfwbar while it has a popup open using this PR vs not using it?

01micko:
> I am not sure if we have to also check for `LAB_LAYER_SURFACE`, @01micko any different behavior when clicking somewhere on sfwbar while it has a popup open using this PR vs not using it?

Not at all, but one weird thing running nested, Thunar would only open on my real desktop, not in the nested session. Tried a few other apps and no issue.

Running this PR right now on bare metal.

Consolatis:
> one weird thing running nested, Thunar would only open on my real desktop, not in the nested session. Tried a few other apps and no issue.

That might be related to us telling dbus about `DISPLAY` and `WAYLAND_DISPLAY`. If the nested labwc uses the same dbus session that could cause an issue like this. To work around it you could try starting the nested instance via `dbus-run-session labwc` so it has its own separate dbus session from your main one.

Thanks for testing, PR LGTM.

01micko:
> `dbus-run-session labwc `

Thanks, that fixed it.

Issue #1013, opened 2023-08-05
Feature Request: OSD showing position when moving, size when resizing a window


Gusar321:
Openbox has this. Didn't occur to me how much I look at it when placing/resizing windows until I was checking out labwc for a potential switch to wayland. So I would really welcome seeing this feature in labwc too.

Consolatis:
Personally I've never been an excessive Openbox user, are there some screenshots of how that looks?

johanmalm:
Yes, I know what that looks like. It's a tiny OSD with `1024x768` (or whatever) at the center of the window. It changes with resize.

Consolatis:
> Yes, I know what that looks like. It's a tiny OSD with `1024x768` (or whatever) at the center of the window. It changes with resize.

I see, is the OSD hiding again based on some timeout after the last change or this is only for interactive resize with the mouse, e.g. rendered on start of the resize until releasing the mouse button?

If its only with the mouse it should be relatively trivial to implement in `src/interactive.c`. The timeout based variant that also triggers on keybind resizes would involve some more changes I guess.

Personally I don't mind the feature as long as it is either disabled by default or at least can be disabled via config.

Consolatis:
Thanks for the screenshots :)

@johanmalm is this in scope of labwc?
And if yes, do you want to implement it? Otherwise I could give it a try, sounds fun to implement.

johanmalm:
Go for it. Agree it would be a nice feature to have.

http://openbox.org/wiki/Help:Configuration#Resize_.28and_move.29

labwc-scope 2.6.1
Have changed it to a B

Consolatis:
- #1021

Testing / feedback welcome.

Gusar321:
That was fast. Thank you for working on this! It works. I agree thought that bold font would look better.

About current limitations, supporting popupPosition is not needed IMO, but size hints would be nice to have, if it's not too difficult to implement

Consolatis:
Have implemented size hints for X11 applications but for wayland native ones its might not be possible (but I'd have to look further into it).

[...]

PR #1021, opened 2023-08-10
Add optional resize / move indicator

Consolatis:
20230810_15h31m30s_grim

Partly fixes
- #1013

Current limitations:
- ~~only supports `<resize popupShow="Always / Never" />`~~
- ~~size hints are ignored, e.g. its either the position for a move operation or the size for a resize operation~~
- `popupPosition` is not supported and always set to `Center`, e.g. being centered within the window
- the OSD is using the same font as for the `A-Tab` window switcher, thus configuring the font weight affects both
  - should we add another font place for this specific OSD? Having it in bold looks better IMHO but would cause `A-Tab` to be in bold as well

Deviations from Openbox:
- the position is shown as `100 , 250` rather than `100 x 250` (as that is a size and not a position).

The resize indicator is disabled by default.

Missing:
- [x] Decide what to do with the font:
  - does this OSD deserves its own font place
  - should we copy the generic OSD one and just add a bold weight to it
  - just keep it as it is and use the generic OSD one
- [x] docs
- [x] Split off `fill_size_hints()` impl to its own commit, squash the rest

Testing:
```xml
<resize popupShow="Always" />
```

### DIFF
--- a/src/ssd/resize_indicator.c
+++ b/src/ssd/resize_indicator.c
@@ -1,0 +1,222 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <wlr/types/wlr_scene.h>
+#include <wlr/util/box.h>
+#include <wlr/util/log.h>
+#include "common/scaled_font_buffer.h"
+#include "labwc.h"
+#include "resize_indicator.h"
+#include "view.h"
+
+static void
+resize_indicator_reconfigure_view(struct resize_indicator *indicator)
+{
+	assert(indicator->tree);
+
+	struct theme *theme = rc.theme;
+	indicator->height = font_height(&rc.font_osd)
+		+ 2 * theme->osd_window_switcher_padding
+		+ 2 * theme->osd_border_width;
+
+	/* Static positions */
+	wlr_scene_node_set_position(&indicator->background->node,
+		theme->osd_border_width, theme->osd_border_width);
+
+	wlr_scene_node_set_position(&indicator->text->scene_buffer->node,
+		theme->osd_border_width + theme->osd_window_switcher_padding,
+		theme->osd_border_width + theme->osd_window_switcher_padding);
+
+	/* Colors */
+	wlr_scene_rect_set_color(indicator->border, theme->osd_border_color);
+	wlr_scene_rect_set_color(indicator->background, theme->osd_bg_color);
+}
+
+static void
+resize_indicator_init(struct view *view)
+{
+	assert(view);
+	struct resize_indicator *indicator = &view->resize_indicator;
+	assert(!indicator->tree);
+
+	indicator->tree = wlr_scene_tree_create(view->scene_tree);
+	indicator->border = wlr_scene_rect_create(
+		indicator->tree, 0, 0, rc.theme->osd_border_color);
+	indicator->background = wlr_scene_rect_create(
+		indicator->tree, 0, 0, rc.theme->osd_bg_color);
+	indicator->text = scaled_font_buffer_create(indicator->tree);
+
+	wlr_scene_node_set_enabled(&indicator->tree->node, false);
+	resize_indicator_reconfigure_view(indicator);
+}
+
+static struct wlr_box
+get_size_hints(struct view *view)
+{
+	assert(view);
+
+	struct wlr_box hints = { 0 };
+	if (view->impl->fill_size_hints) {
+		view->impl->fill_size_hints(view, &hints);
+	}
+	return hints;
+}
+
+static bool
+wants_indicator(struct view *view)
+{
+	assert(view);
+
+	if (rc.resize_indicator == LAB_RESIZE_INDICATOR_NON_PIXEL) {
+		if (view->server->input_mode != LAB_INPUT_STATE_RESIZE) {
+			return false;
+		}
+		struct wlr_box size_hints = get_size_hints(view);
+		if (size_hints.width && size_hints.height) {
+			return true;
+		}
+	}
+	return rc.resize_indicator == LAB_RESIZE_INDICATOR_ALWAYS;
+}
+
+void
+resize_indicator_reconfigure(struct server *server)
+{
+	struct view *view;
+	wl_list_for_each(view, &server->views, link) {
+		struct resize_indicator *indicator = &view->resize_indicator;
+		if (indicator->tree) {
+			resize_indicator_reconfigure_view(indicator);
+		}
+		if (view != server->grabbed_view) {
+			continue;
+		}
+
+		/* This view is currently in an interactive move/resize operation */
+		if (indicator->tree && indicator->tree->node.enabled) {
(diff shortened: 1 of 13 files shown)


johanmalm:
Really snazzy :smile:
Have tested. Works well.
Review comments below:

- With `snprintf` use`%d` as it's integer rather than unsigned integer.
- `text[14]` seems a bit tight. `snprintf` writes a maximum of `size` characters, but that includedes the null-terminator, so you'd only have 13. Also, remember that when moving the windows you might have a minus sign there too. And I guess we could future proof it by adding a few more bytes for really big screens. I'd go for 32 or something like that.
- Regarding the font - I suggest doing that as a separate PR because I'm keen to think through the approach (see below).

Font considerations:
- Do we move font from `rc.xml` to `themerc`. It feels like it belongs there and we have `themerc-override`.
- The OSD issue had a proposal to add separate font specification for workspace-switcher and window-switcher.
- We have not implemented the Openbox MenuHeader and InactiveWindow entries.

Consolatis:
> * With `snprintf` use`%d` as it's integer rather than unsigned integer.

On my system `man snprintf` lists `%i` and `%d` as doing the same. Some [searches](https://stackoverflow.com/questions/17329647/should-i-use-i-or-d-to-print-an-integer-in-c-using-printf) seem to agree to that point in the `*printf()` case. `%i` seems to work differently in the `scanf()` case though (e.g. dealing with ordinal or hex formats). Still, apparently `%d` is what most people suggest to use for `*printf()`. I'll change.

> * `text[14]` seems a bit tight. `snprintf` writes a maximum of `size` characters, but that includes the null-terminator, so you'd only have 13. Also, remember that when moving the windows you might have a minus sign there too. And I guess we could future proof it by adding a few more bytes for really big screens. I'd go for 32 or something like that.

Right, I've calculated with 13 + null terminator, see inline comment: `12345 x 12345`. Thus this allows values from
`-9999,-9999 to 99999,99999`. I'd be amazed if we'd see a resolution like that any time soon. However adding a few bytes also doesn't really matter either. I'll change to 32 as suggested.

> * Regarding the font - I suggest doing that as a separate PR because I'm keen to think through the approach (see below).

Alright, I'll keep the font handling as it is right now.

> Font considerations:
> 
> * Do we move font from `rc.xml` to `themerc`. It feels like it belongs there and we have `themerc-override`.
> * The OSD issue had a proposal to add separate font specification for workspace-switcher and window-switcher.
> * We have not implemented the Openbox MenuHeader and InactiveWindow entries.

I think we've talked about this before, theoretically everything that we can remove from the config `<theme>` section and move into the actual theme sounds good to me. I think Openbox used to support setting fonts via theme files:
http://openbox.org/wiki/Help:Themes#Text_shadows but changed it (didn't find a reason why though): http://openbox.org/wiki/Help:Upgrading_to_3.4#Themes

So moving the font to the themes again would be a deviation vs the latest Openbox specs. And might also be a bit difficult to parse because they seem to have reused the existing `.font` names for.. shadows?

src/ssd/resize_indicator.c
```diff
@@ -114,8 +115,17 @@ resize_indicator_update(struct view *view)
 
 	char text[32]; /* 12345 x 12345 would be 13 chars + 1 null byte */
 	if (view->server->input_mode == LAB_INPUT_STATE_RESIZE) {
-		snprintf(text, sizeof(text), "%d x %d",
-			view->current.width, view->current.height);
+		int width = view->current.width;
+		int height = view->current.height;
+		if (view->impl->fill_size_hints) {
+			struct wlr_box hints;
```
johanmalm, line 121: Should we initialise hints to zero here (={0})

johanmalm (commented):
Looks good. Just a pick up zeroing hints for the xdg-shell case.

src/ssd/resize_indicator.c
Consolatis: Not strictly necessary as `xwayland_view_fill_size_hints()` always sets `width` and `height`.
However, due to the addition of the `Nonpixel` variant I had to restructure some stuff and its now done with a helper which initializes the box with `{ 0 }`.

Consolatis:
Added the `nonPixel` mode so this now supports all 3 states as mentioned in the [Openbox wiki](http://openbox.org/wiki/Help:Configuration#Resize_.28and_move.29).

Will clean up the commits later and add docs. Final review / test would be welcome.

src/ssd/resize_indicator.c
```diff
@@ -111,27 +144,40 @@ void
...
 {
 	assert(view);
+	assert(view == view->server->grabbed_view);
+
+	if (!wants_indicator(view)) {
+		return;
+	}
+
+	/*
+	 * Following asserts ensure that we'll never get out-of-sync
```
johanmalm, line 152: If there is a chance that the arguments passed to the asserts may be false because of something the user does (like re-loading the config) then we should re-think approach.
Sorry if I misunderstood logic :smile:

johanmalm:
LGTM except on comment on asserts (which I might have misunderstood)

Have tested. Works well (incl new nonpixel mode)

src/ssd/resize_indicator.c
Consolatis: Comment is badly formulated, will change.

Those asserts() should obviously never trigger by anything the users does, they will however trigger if we do something bad in the future :)

johanmalm: Thought it was just the comment.
:smile:

Consolatis: I have reconsidered the `assert()` approach for this case: it now works around it and logs an `WLR_INFO` message that states that resize_indicator has to use a fallback path. I imagine changes in the future might not get necessarily tested with all variations of the `rc.resize_indicator` setting and thus the chances of a bug slipping through is non-zero in this case. This change prevents crashing labwc for users in that case and should still allow devs to detect the issue after the fact via the log message.

Consolatis:
I have changed the internal design of this feature: instead of using a shared global instance of a resize indicator each view now has their own indicator (if its enabled).

The global variant had a potential issue in that if a view was destroyed while the global instance scene node was attached to it, the global indicator was destroyed as well which would cause labwc to crash on the next usage of the indicator (if lucky). That was guarded against in the earlier implementation but it didn't feel very robust in the long term, especially when also thinking about future changes in the code base.

The new variant prevents this issue completely as each view now has its own indicator that is intended to be destroyed along with the view itself. It adds 40 bytes of memory to each view which I don't think is a bad trade-off vs having a more robust implementation. It could also be reduced to 8 bytes by having the struct being allocated dynamically but I didn't feel like that is actually necessary.

01micko:
> Really snazzy 😄 Have tested. Works well. 

I concurr!

johanmalm:
LGTM. Have tested. Works well.

Consolatis:
> LGTM. Have tested. Works well.

Sweet, thanks for review and testing everyone.

PR #1022, opened 2023-08-10
Add If and ForEach actions

heroin-moose:
This patch series introduces support for [If](http://openbox.org/wiki/Help:Actions#If) and [ForEach](http://openbox.org/wiki/Help:Actions#ForEach) actions as referenced in OpenBox Specification.

Missing parts:
* No action nesting
* Most conditions are not there yet (only identifier and title are supported)
* If's query lacks target argument

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -148,6 +148,56 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="None" />*
 	If used as the only action for a binding: clear an earlier defined binding.
 
+# CONDITIONAL ACTIONS
+
+Actions that execute other actions. Used in keyboard/mouse bindings.
+
+*<action name="If">*
+	This action will execute one set of actions if the focused window
+	matches the criteria, or another if it does not.
+
+	The arguments are as follows:
+
+	```
+	<action name="If">
+	  <query/>
+	  <then><action/></then>
+	  <else><action/></else>
+	</action>
+	```
+
+	*query*
+		Define a query with zero or more conditions. All conditions must
+		be evaluated as true in order for the window to match this
+		query. Multiple queries can be defined.
+
+		Pattern matching is done according to glob(7) and is
+		case-insensitive.
+
+		Conditions are as follows:
+
+		*identifier*
+			XDG shell app_id for Wayland clients, WM_CLASS for
+			XWayland clients.
+
+		*title*
+			XDG shell title for Wayland clients, WM_NAME for
+			XWayland clients.
+
+		This argument is optional.
+	
+	*then*
+		A list of actions to be executed if the window matches any
+		query. This argument is optional.
+
+	*else*
+		A list of actions to be executed if the window does not match
+		any query. This argument is optional.
+
+*<action name="ForEach">*
+	Identical to "If" action, but applies to all windows, not just the
+	focused one.
+
 # SEE ALSO
 
-labwc(1), labwc-config(5), labwc-theme(5)
+labwc(1), labwc-config(5), labwc-theme(5), glob(7)
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -127,9 +127,9 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="GoToDesktop" to="value" wrap="yes" />*
 	Switch to workspace.
 
-	*to* The workspace to switch to. Supported values are "last", "left",
-	"right" or the full name of a workspace or its index (starting at 1)
-	as configured in rc.xml.
+	*to* The workspace to switch to. Supported values are "current", "last",
+	"left", "right" or the full name of a workspace or its index (starting
+	at 1) as configured in rc.xml.
 
 	*wrap* [yes|no] Wrap around from last desktop to first, and vice
 	versa. Default yes.
--- a/include/action.h
+++ b/include/action.h
@@ -23,6 +23,11 @@ struct action *action_create(const char *action_name);
 bool action_is_valid(struct action *action);
 
 void action_arg_add_str(struct action *action, const char *key, const char *value);
+void action_arg_add_actionlist(struct action *action, const char *key);
+void action_arg_add_querylist(struct action *action, const char *key);
+
+struct wl_list *action_get_actionlist(struct action *action, const char *key);
+struct wl_list *action_get_querylist(struct action *action, const char *key);
 
 void action_arg_from_xml_node(struct action *action, const char *nodename, const char *content);
 
--- a/src/action.c
+++ b/src/action.c
@@ -158,6 +158,41 @@ action_arg_add_int(struct action *action, const char *key, int value)
 	wl_list_append(&action->args, &arg->base.link);
 }
 
+static void *
+action_get_arg(struct action *action, const char *key, enum action_arg_type type)
+{
+	assert(action);
+	assert(key);
+	struct action_arg *arg;
+	wl_list_for_each(arg, &action->args, link) {
+		if (!strcasecmp(key, arg->key) && arg->type == type) {
+			return arg;
+		}
+	}
+	return NULL;
+}
+
+static const char *
+action_get_str(struct action *action, const char *key, const char *default_value)
+{
+	struct action_arg_str *arg = action_get_arg(action, key, LAB_ACTION_ARG_STR);
+	return arg ? arg->value : default_value;
+}
+
+static bool
+action_get_bool(struct action *action, const char *key, bool default_value)
+{
+	struct action_arg_bool *arg = action_get_arg(action, key, LAB_ACTION_ARG_BOOL);
+	return arg ? arg->value : default_value;
+}
+
+static int
+action_get_int(struct action *action, const char *key, int default_value)
+{
+	struct action_arg_int *arg = action_get_arg(action, key, LAB_ACTION_ARG_INT);
+	return arg ? arg->value : default_value;
+}
+
 void
 action_arg_from_xml_node(struct action *action, const char *nodename, const char *content)
 {
@@ -256,63 +291,6 @@ action_str_from_arg(struct action_arg *arg)
 	return ((struct action_arg_str *)arg)->value;
 }
 
-static bool
-arg_value_exists(struct action *action, const char *key, enum action_arg_type type)
-{
-	assert(action);
-	assert(key);
-	struct action_arg *arg;
-	wl_list_for_each(arg, &action->args, link) {
-		if (!strcasecmp(key, arg->key) && arg->type == type) {
-			return true;
-		}
-	}
-	return false;
-}
-
-static const char *
-get_arg_value_str(struct action *action, const char *key, const char *default_value)
-{
-	assert(action);
-	assert(key);
-	struct action_arg *arg;
-	wl_list_for_each(arg, &action->args, link) {
-		if (!strcasecmp(key, arg->key)) {
-			return action_str_from_arg(arg);
-		}
-	}
-	return default_value;
-}
-
-static bool
-get_arg_value_bool(struct action *action, const char *key, bool default_value)
-{
-	assert(action);
-	assert(key);
-	struct action_arg *arg;
-	wl_list_for_each(arg, &action->args, link) {
-		if (!strcasecmp(key, arg->key)) {
-			assert(arg->type == LAB_ACTION_ARG_BOOL);
-			return ((struct action_arg_bool *)arg)->value;
-		}
-	}
-	return default_value;
-}
-
-static int
-get_arg_value_int(struct action *action, const char *key, int default_value)
-{
-	assert(action);
-	assert(key);
-	struct action_arg *arg;
-	wl_list_for_each(arg, &action->args, link) {
-		if (!strcasecmp(key, arg->key)) {
-			assert(arg->type == LAB_ACTION_ARG_INT);
-			return ((struct action_arg_int *)arg)->value;
-		}
-	}
-	return default_value;
-}
 
 static struct action_arg *
 action_get_first_arg(struct action *action)
@@ -402,7 +380,7 @@ action_is_valid(struct action *action)
 		return true;
 	}
 
-	if (arg_value_exists(action, arg_name, arg_type)) {
+	if (action_get_arg(action, arg_name, arg_type)) {
 		return true;
 	}
 
@@ -575,14 +553,14 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_MOVE_TO_EDGE:
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
-				enum view_edge edge = get_arg_value_int(action, "direction", 0);
+				enum view_edge edge = action_get_int(action, "direction", 0);
 				view_move_to_edge(view, edge);
 			}
 			break;
 		case ACTION_TYPE_SNAP_TO_EDGE:
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
-				enum view_edge edge = get_arg_value_int(action, "direction", 0);
+				enum view_edge edge = action_get_int(action, "direction", 0);
 				view_snap_to_edge(view, edge, /*store_natural_geometry*/ true);
 			}
 			break;
@@ -665,24 +643,24 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_RESIZE_RELATIVE:
 			if (view) {
-				int left = get_arg_value_int(action, "left", 0);
-				int right = get_arg_value_int(action, "right", 0);
-				int top = get_arg_value_int(action, "top", 0);
-				int bottom = get_arg_value_int(action, "bottom", 0);
+				int left = action_get_int(action, "left", 0);
+				int right = action_get_int(action, "right", 0);
+				int top = action_get_int(action, "top", 0);
+				int bottom = action_get_int(action, "bottom", 0);
 				view_resize_relative(view, left, right, top, bottom);
 			}
 			break;
 		case ACTION_TYPE_MOVETO:
 			if (view) {
-				int x = get_arg_value_int(action, "x", 0);
-				int y = get_arg_value_int(action, "y", 0);
+				int x = action_get_int(action, "x", 0);
+				int y = action_get_int(action, "y", 0);
 				view_move(view, x, y);
 			}
 			break;
 		case ACTION_TYPE_MOVE_RELATIVE:
 			if (view) {
-				int x = get_arg_value_int(action, "x", 0);
-				int y = get_arg_value_int(action, "y", 0);
+				int x = action_get_int(action, "x", 0);
+				int y = action_get_int(action, "y", 0);
 				view_move_relative(view, x, y);
 			}
 			break;
@@ -694,8 +672,8 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_GO_TO_DESKTOP:
 			{
 				bool follow = true;
-				bool wrap = get_arg_value_bool(action, "wrap", true);
-				const char *to = get_arg_value_str(action, "to", NULL);
+				bool wrap = action_get_bool(action, "wrap", true);
+				const char *to = action_get_str(action, "to", NULL);
 				/*
 				 * `to` is always != NULL here because otherwise we would have
 				 * removed the action during the initial parsing step as it is
@@ -708,7 +686,7 @@ actions_run(struct view *activator, struct server *server,
 				}
 				if (action->type == ACTION_TYPE_SEND_TO_DESKTOP) {
 					view_move_to_workspace(view, target);
-					follow = get_arg_value_bool(action, "follow", true);
+					follow = action_get_bool(action, "follow", true);
 				}
 				if (follow) {
 					workspaces_switch_to(target,
--- a/src/action.c
+++ b/src/action.c
@@ -284,25 +284,6 @@ cleanup:
 	free(argument);
 }
 
-static const char *
-action_str_from_arg(struct action_arg *arg)
-{
-	assert(arg->type == LAB_ACTION_ARG_STR);
-	return ((struct action_arg_str *)arg)->value;
-}
-
-
-static struct action_arg *
-action_get_first_arg(struct action *action)
-{
-	struct action_arg *arg;
-	struct wl_list *item = action->args.next;
-	if (item == &action->args) {
-		return NULL;
-	}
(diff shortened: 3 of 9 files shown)


johanmalm:
Wow. Can’t believe we’re doing this. Very cool.
Will go dragon bashing when I get a minute.
Did you have a particular use-case to help us think through the code?

heroin-moose:
I have a bunch of apps requiring my attention, so I press the button, spawn them, drop a line and send them to iconified limbo. For example:

```
    <keybind key="W-apostrophe">
      <action name="ForEach">
        <query title="weechat"/>
        <then>
          <action name="SendToDesktop" to="current"/>
          <action name="SnapToRegion" region="center"/>
          <action name="Raise"/>
          <action name="Focus"/>
        </then>
      </action>
    </keybind>
```

src/workspaces.c
```diff
@@ -333,6 +333,8 @@ workspaces_find(struct workspace *anchor, const char *name, bool wrap)
 				return target;
 			}
 		}
+	} else if (!strcasecmp(name, "current")) {
+		return anchor;
```
Consolatis, line 337: This seems like accidentally committed. If you send a new PR with only this change (and updated docs for `SendToDesktop`) I'd be fine merging this if @johanmalm agrees.

It's listed in the scope repo as `4.7.1.4` for `GoToDesktop` and cat `C` and it really doesn't make sense for that.
For `SendToDesktop` in combination with this PR however, it pretty much makes sense IMHO.

heroin-moose: This is basically to test my use case. I'll drop it after WIP.

include/action.h
```diff
@@ -18,11 +18,20 @@ struct action {
 	struct wl_list args;  /* struct action_arg.link */
 };
 
+struct action_query {
+	struct wl_list link;
+	const char *app_id;
```
johanmalm, line 23: For window-rules we use “identifier” to cover xdg-shell’s app_id and xwayland’s WM_CLASS.

https://labwc.github.io/labwc-config.5.html#window_rules

https://github.com/labwc/labwc/blob/b149526d037b2de4809319cacb08c62dd3954b34/src/window-rules.c#L46

src/action.c
```diff
@@ -743,6 +826,12 @@ actions_run(struct view *activator, struct server *server,
 				desktop_focus_output(output_from_name(server, output_name));
 			}
 			break;
+		case ACTION_TYPE_FOR_EACH:
+			struct view *v, *next;
+			wl_list_for_each_safe(v, next, &server->views, link) {
```
johanmalm, line 831: Iterating over all views for the “else” actions is likely to cause problems with some xwayland windows.

Suggest using the new “desktop_views_append” API (PR close to merge.

Or at least check that view has surface and has been mapped at some point.

src/action.c
```diff
@@ -507,6 +548,48 @@ view_for_action(struct view *activator, struct server *server,
 	}
 }
 
+static bool
+view_matches_query(struct view *view, struct action_query *query)
+{
+	bool result = true;
+
+	const char *app_id = view_get_string_prop(view, "app_id");
```
johanmalm, line 556: See discussion above re identifier

src/action.c
```diff
@@ -118,6 +125,7 @@ const char *action_names[] = {
 	"SnapToRegion",
 	"ToggleKeybinds",
 	"FocusOutput",
+	"ForEach",
```
johanmalm, line 128: Do we need an if action too?

johanmalm:
Looks good so far. A few minor review comments submitted.

include/action.h
heroin-moose: Huh, totally forgot about XWayland TBH. Will update.

src/action.c
heroin-moose: Yeah, I will add it.

src/action.c
heroin-moose: I'll check the desktop_views_append() API, thanks.

heroin-moose: Seems like desktop_views_append() only works for surfaces on the current desktop, while ForEach should work on all desktops. Checking for mapped surfaces maybe an option. Could you elaborate on how to do it and what happens if I don't?

johanmalm: You can use:

```
struct view *view;
wl_list_for_each(view, &server->views, link)
{
	if (!isfocusable(view)) {
		continue;
	}
	// ...then do whatever with view
}
```

If you don't call `isfocusable()` you run the risk of segfault because some xwayland clients create views which are never ever mapped and can set view->surface to NULL at any time.

The new API (https://github.com/labwc/labwc/pull/1024/files) is not just for current workspace. It can be used like this:

```
struct wl_array views;
wl_array_init(&views);
desktop_views_append(server, &views, 0 /* no criteria */);

struct view **view;
wl_array_for_each(view, &views) {
	// Do something with *view
}
wl_array_release(&views);
```

heroin-moose: Okay, the `desktop_` part completely threw me off.

include/action.h
heroin-moose: Actually... do we want  to go forward with it? The rest of the code base uses app_id (e.g. window switcher) for both cases.

Consolatis: The user-facing API (e.g. config file) is `identifier` and we should stick to that, what we do internally doesn't matter all that much.

heroin-moose: Are you sure? Because man clearly says app_id:

```
  <windowSwitcher show="yes" preview="yes" outlines="yes">
    <fields>
      <field content="type" width="25%" />
      <field content="app_id" width="25%" />
    </fields>
  </windowSwitcher>
```

Consolatis: Hm.. good point. I was thinking about [window rules](https://labwc.github.io/labwc-config.5.html#window_rules).

Maybe we should change all window switcher examples to use "identifier" as well here (while still allowing "app_id" to not break older configs). Thoughts @johanmalm ?

heroin-moose:
Updated.

include/action.h
johanmalm: Yes, looks like my mistake using different currencies for window rules and window switcher. 

Agree with @Consolatis. Let’s use identifier, and patch window-switcher with backward compatible support.

heroin-moose: Yup, updated.

src/config/rcxml.c
```diff
@@ -1343,6 +1446,8 @@ rcxml_finish(void)
 	current_mouse_context = NULL;
 	current_keybind_action = NULL;
 	current_mousebind_action = NULL;
+	current_child_action = NULL;
```
johanmalm, line 1449: We must free the action-queries + child-actions here.

src/view.c
```diff
@@ -19,6 +19,24 @@
 #define LAB_FALLBACK_WIDTH  640
 #define LAB_FALLBACK_HEIGHT 480
 
+bool
+view_matches_query(struct view *view, struct view_query *query)
```
johanmalm, line 23: Very neat. This is nice AND logic for multiple queries.

We ought to consider combining this with `view_matches_criteria()` in `window-rules.c`, but can do that later. The way you've written it is better. https://github.com/labwc/labwc/blob/206db43786e066f7de3755a11263a971b4ed77be/src/window-rules.c#L46

Two considerations:

- Use `strcasecmp()`
- Use `match_glob()`

src/config/rcxml.c
```diff
@@ -748,6 +833,24 @@ xml_tree_walk(xmlNode *node)
 			in_window_rules = false;
 			continue;
 		}
+		if (!strcasecmp((char *)n->name, "query")) {
```
johanmalm, line 836: Considering protecting against nested if-actions and foreach-actions.
For example if in an `<else>` branch, should we accept another `<query>` (or `if`/`foreach` action).

src/config/rcxml.c
```diff
@@ -241,6 +247,74 @@ fill_region(char *nodename, char *content)
...
+	if (!strcmp(nodename, "identifier")) {
+		current_action_query->identifier = xstrdup(content);
+	} else if (!strcmp(nodename, "title")) {
+		current_action_query->title = xstrdup(content);
+	}
+}
+
+static void
+fill_child_action(char *nodename, char *content, struct action *parent,
+		  const char *arg_name)
```
johanmalm, line 285: Consider `s/arg_name/branch_name/`

src/config/rcxml.c
```diff
@@ -516,7 +590,18 @@ entry(xmlNode *node, char *nodename, char *content)
 		fill_usable_area_override(nodename, content);
 	}
 	if (in_keybind) {
-		fill_keybind(nodename, content);
+		if (in_action_query) {
+			fill_action_query(nodename, content,
+				current_keybind_action);
```
johanmalm, line 595: Could we have `if (in_action_{query,then_branch,else_branch})` at the same level as `in_{key,mouse}bind` if we use of a global (within translation unit) `parent_action` variable?

src/config/rcxml.c
heroin-moose: They are added to the parent action's list now. Or did I miss some error case?

Consolatis: I don't see the new actions being freed nor the queries being free'd in `action_free()` / `action_list_free()`, similar to the `if (arg->type == LAB_ACTION_ARG_STR) {` condition in `action_free()`.

src/action.c
```diff
@@ -134,6 +144,18 @@ action_arg_add_str(struct action *action, const char *key, const char *value)
 	wl_list_append(&action->args, &arg->base.link);
 }
 
+void
+action_arg_add_list(struct action *action, const char *key)
```
heroin-moose, line 148: I totally forgot about actually freeing that list, LOL.

src/config/rcxml.c
heroin-moose: Hivemind! Just mentioned it further.

Consolatis: After discussion on IRC my question regarding the list for `struct view_query` is answered: To be able to implement OR logic when multiple queries share the same parent.

src/view.c
```diff
@@ -19,6 +20,24 @@
...
+	if (match && query->identifier && identifier) {
+		match &= match_glob(query->identifier, identifier);
+	}
+
+	const char *title = view_get_string_prop(view, "title");
+	if (match && query->title && title) {
+		match &= match_glob(query->title, title);
+	}
+
+	return match;
```
Consolatis, line 38: - I don't understand the first `match` condition (and the `&=` in there), it is always true. I don't particularly dislike it due to the provided symmetry between the two conditions though. So consider this a nitpick.
- I am not sure if we could somehow end up getting `view_get_string_prop()` return `NULL`. In that case we could wrongly match the query.

`view_get_string_prop()` may return:
- `xdg_toplevel->title`
- `xdg_toplevel->app_id`
- `xwayland_surface->title`
- `xwayland_surface->class`

I think we should either verify (and document) that wlroots will never have any of these being `NULL` or we should check against `NULL` and return `""` in `view_get_string_prop()`.

include/action.h
```diff
@@ -23,6 +23,11 @@ struct action *action_create(const char *action_name);
 bool action_is_valid(struct action *action);
 
 void action_arg_add_str(struct action *action, const char *key, const char *value);
+void action_arg_add_alist(struct action *action, const char *key);
+void action_arg_add_qlist(struct action *action, const char *key);
+
+struct wl_list *action_get_alist(struct action *acton, const char *key);
+struct wl_list *action_get_qlist(struct action *acton, const char *key);
```
Consolatis, line 30: I'd prefer using more descriptive naming here, somebody new to the code base (or me in a couple of weeks) won't know what `a` and `q` stand for.

src/action.c
```diff
@@ -158,6 +169,30 @@ action_arg_add_int(struct action *action, const char *key, int value)
...
+action_arg_add_alist(struct action *action, const char *key)
+{
+	assert(action);
+	assert(key);
+	struct action_arg_list *arg = znew(*arg);
+	arg->base.type = LAB_ACTION_ARG_ACTION_LIST;
+	arg->base.key = xstrdup(key);
+	wl_list_init(&arg->value);
+	wl_list_append(&action->args, &arg->base.link);
+}
```
Consolatis, line 194: Suggest combining these two into 3 functions:
Internal `static void action_arg_add_list(struct action *action, const char *key, enum action_arg_type)` doing most of the work and two wrappers for the new types.

src/action.c
```diff
@@ -466,6 +515,30 @@ view_for_action(struct view *activator, struct server *server,
...
+	queries = action_get_qlist(action, "query");
+	if (queries) {
+		branch = "else";
+		wl_list_for_each(query, queries, link) {
+			if (view_matches_query(view, query)) {
+				branch = "then";
+				break;
+			}
+		}
+	}
```
Consolatis, line 534: Suggest documenting the implicit `OR` logic between queries here, simple 1-line comment should be enough.

src/action.c
```diff
@@ -692,6 +765,21 @@ actions_run(struct view *activator, struct server *server,
 				desktop_focus_output(output_from_name(server, output_name));
 			}
 			break;
+		case ACTION_TYPE_IF:
+			if (view) {
+				run_if_action(view, server, action);
+			}
+			break;
+		case ACTION_TYPE_FOR_EACH:
+			struct wl_array views;
```
Consolatis, line 774: `;` or `{` as first statement here please (older compilers and stuff).

src/config/rcxml.c
```diff
@@ -241,6 +247,74 @@ fill_region(char *nodename, char *content)
 	}
 }
 
+static void
+fill_action_query(char *nodename, char *content, struct action *action)
```
Consolatis, line 251: Maybe and `assert(action)` would be good here?

src/config/rcxml.c
```diff
@@ -241,6 +247,74 @@ fill_region(char *nodename, char *content)
...
+	if (!siblings) {
+		action_arg_add_alist(parent, arg_name);
+		siblings = action_get_alist(parent, arg_name);
+	}
+
+	if (!strcmp(nodename, "name.action")) {
+		current_child_action = action_create(content);
+		wl_list_append(siblings, &current_child_action->link);
+	} else if (!current_child_action) {
+		wlr_log(WLR_ERROR, "TODO hurr durr wrong order");
```
Consolatis, line 312: This log message should maybe be a bit more formal :)
We could also look into having the same behavior everywhere when this happens.

include/action.h
heroin-moose: Naming: discussed on IRC, for now we leave it as is.
Typo: Fixed.

src/action.c
heroin-moose: Agree, done.

src/config/rcxml.c
heroin-moose: Why not, done.

src/action.c
heroin-moose: Agree, done.

src/config/rcxml.c
heroin-moose: Yeah, done :)

src/action.c
heroin-moose: Why not, done.

src/config/rcxml.c
heroin-moose: I've added the logic to ignore child If and ForEach. As for query, there is no easy way with current parser to prevent such things.

src/workspaces.c
heroin-moose: Dropped.

src/config/rcxml.c
heroin-moose: I'm not sure I'm following.

heroin-moose: Ah, I get it. In order to get rid of code duplication between mousebind and keybind?

heroin-moose: Sounds good, but let's do it afterwards.

src/config/rcxml.c
```diff
@@ -241,6 +247,74 @@ fill_region(char *nodename, char *content)
...
+		return;
+	}
+
+	struct wl_list *siblings = action_get_alist(parent, arg_name);
+	if (!siblings) {
+		action_arg_add_alist(parent, arg_name);
+		siblings = action_get_alist(parent, arg_name);
+	}
+
+	if (!strcmp(nodename, "name.action")) {
```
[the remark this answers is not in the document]
heroin-moose: Agree, done.

src/config/rcxml.c
johanmalm: Yes, exactly. Fine with one step at a t ime.

src/config/rcxml.c
```diff
@@ -241,6 +247,77 @@ fill_region(char *nodename, char *content)
...
+		siblings = action_get_actionlist(parent, branch_name);
+	}
+
+	if (!strcasecmp(nodename, "name.action")) {
+		if (!strcasecmp(content, "If") || !strcasecmp(content, "ForEach")) {
+			wlr_log(WLR_ERROR, "action '%s' cannot be a child action", content);
+			return;
+		}
+		current_child_action = action_create(content);
+		wl_list_append(siblings, &current_child_action->link);
```
Consolatis, line 312: Just so we don't forget about this: the return value from `action_create()` can be `NULL` so we need to check for that before calling `wl_list_append()`.

src/view.c
```diff
@@ -19,6 +20,33 @@
...
+}
```
Consolatis, line 48: 2nd point already discussed via IRC, comment here just to ensure we don't forget about it: empty queries (caused by for example the wrong arguments) should not match, missing queries should (at least for `ForEach`).

So this function should return `false` when none of the `query` arguments are set.
The missing query part is already handled correctly in `run_if_actions()`

kyak:
I'm trying to use it like this:

```
    <keybind key="W-t">
      <action name="If">
        <query identifier="org.telegram.desktop" />
        <then>
          <action name="Raise" />
        </then>
      </action>
    </keybind>
```

But it doesn't raise Telegram.

I'm pretty sure that the identifier is correct, since this works:

```
    <windowRule identifier="org.telegram.desktop" serverDecoration="no">
    </windowRule>
```

Consolatis:
The `If` action only acts on the currently focused window.
It seems like the `ForEach` action would be more what is intended with your example.

kyak:
> The `If` action only acts on the currently focused window. It seems like the `ForEach` action would be more what is intended with your example.

Yep, this works, thanks!

Seems that `Reconfigure` doesn't reload these nested actions, had to restart labwc.

src/config/rcxml.c
heroin-moose: Fixed.

src/view.c
heroin-moose: Fixed.

heroin-moose: This is ugly though :( It would be better not to create such query in the first place.

docs/labwc-actions.5.scd
```diff
@@ -147,6 +147,56 @@ Actions are used in menus and keyboard/mouse bindings.
...
+
+		*identifier*
+			XDG shell app_id for Wayland clients, WM_CLASS for
+			XWayland clients.
+
+		*title*
+			XDG shell title for Wayland clients, WM_NAME for
+			XWayland clients.
+
+		This argument is optional.
```
johanmalm, line 186: Does it make sense of query to be optional for än if-action?

docs/labwc-actions.5.scd
```diff
@@ -147,6 +147,56 @@ Actions are used in menus and keyboard/mouse bindings.
...
+	  <else><action/></else>
+	</action>
+	```
+
+	*query*
+		Define a query with zero or more conditions. All conditions must
+		be evaluated as true in order for the window to match this
+		query. Multiple queries can be defined.
+
+		Pattern matching is done according to glob(7) and is
```
johanmalm, line 173: Have we used glob(7) or g_pattern_match_simple()? 
On phone, so can't easily scroll down and check.
match_glob() uses the latter and I'm not sure that it's "glob(7)" compliant

docs/labwc-actions.5.scd
heroin-moose: Since we have no option to actually show user a parser error on reconfigure, it seems easier to do what he said.

docs/labwc-actions.5.scd
heroin-moose: match_glob(), but I remember we talked about switch it from non-standard glib function to glob.

johanmalm: Can't remember `fnmatch()` discussion, but happy to move that way.
We have to keep the pattern match find `foo.bar.baz` given `foo.*.baz` (so glob rather than regex). Other that, I don't think we have any backward compatibility considerations.

Did we consider `<title type="regex">` for `<action name="If">` at one point? Have slept since.
http://openbox.org/wiki/Help:Actions#If

heroin-moose: > Can't remember fnmatch() discussion, but happy to move that way.

During this PR?

> Did we consider <title type="regex"> for <action name="If"> at one point? Have slept since.

Not yet. IMO, basic pattern matching is fine for now.

johanmalm: Happy to move to fnmatch as part of this PR or one of could change it later. I’m easy 😊

heroin-moose: Let's get over this, then. Updated.

heroin-moose:
> Seems that `Reconfigure` doesn't reload these nested actions, had to restart labwc.

Can you try on the latest master? Reloading works for me.

johanmalm:
`src/action.c:426: WARNING: Prefer ARRAY_SIZE(branches)` :smile:

johanmalm:
Just applied this to the HEAD~3 commit

```
index a556dd1..414c849 100644
--- a/src/action.c
+++ b/src/action.c
@@ -7,6 +7,7 @@
 #include <unistd.h>
 #include <wlr/util/log.h>
 #include "action.h"
+#include "common/array-size.h"
 #include "common/list.h"
 #include "common/mem.h"
 #include "common/parse-bool.h"
@@ -423,7 +424,7 @@ action_is_valid(struct action *action)
 	case ACTION_TYPE_FOR_EACH:
 		; /* works around "a label can only be part of a statement" */
 		static const char * const branches[] = { "then", "else" };
-		for (size_t i = 0; i < sizeof(branches) / sizeof(branches[0]); i++) {
+		for (size_t i = 0; i < ARRAY_SIZE(branches); i++) {
 			struct wl_list *children = action_get_actionlist(action, branches[i]);
 			if (children && !action_list_is_valid(children)) {
 				wlr_log(WLR_ERROR, "Invalid action in %s '%s' branch",
```

johanmalm:
Thanks @heroin-moose !

PR #1023, opened 2023-08-10
checkpatch.pl: check single statement brace

johanmalm:
Because the @Consolatis asketh for this

### DIFF
--- a/scripts/checkpatch.pl
+++ b/scripts/checkpatch.pl
@@ -5508,11 +5508,32 @@ sub process {
 				}
 			}
 
+
+			# We need this for the labwc-custom check below. It
+			# avoids false positives with do { } while (); etc.
+			my $starts_with_if_while_etc = 0;
+			if ($s =~ /^\+\s*(if|while|for|switch).*/) {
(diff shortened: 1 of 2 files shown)


scripts/checkpatch.pl
```diff
@@ -5508,11 +5508,30 @@ sub process {
...
 			$s =~ s/$;//g;	# Remove any comments
+
+			# CUSTOM: labwc check to find if-condiationals without opening braces because
+			# (as opposed to Linux coding style) we use braces for single statement blocks
+			#
+			# include/ssd-internal.h contains a macro that we can't deal with, so ignore that
+			#
+			if ($starts_with_if_while_etc && !length($s)
+					&& $filename ne "include/ssd-internal.h") {
+				CHK("BRACES", "open brace { expected after if confidional");
```
Consolatis, line 5532: `conditional`, also in comment above

Also: maybe we should use if-conditionals without braces here. Just because :)

johanmalm: Sorted

PR #1024, opened 2023-08-10
Add view_array_append() and for_each_view()

johanmalm:
...to reduce code duplication.

The function populates an array with views which meet any set of critera from:

  - current-workspace
  - no-always-on-top
  - no-skipWindowSwitcher (window-rule)

Make src/osd.c use this new interface. Note that always-on-top views are still filtered out from the window-switcher and that desktop_cycle_view() needs to be re-worked before always-on-top views can be opted in.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -160,6 +160,36 @@ struct xdg_toplevel_view {
 	struct wl_listener new_popup;
 };
 
+enum lab_view_criteria {
+	LAB_VIEW_CRITERIA_ALL = 0,
+	LAB_VIEW_CRITERIA_CURRENT_WORKSPACE = 1 << 0,
+	LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP = 1 << 1,
+	LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER = 1 << 2,
+};
+
+/**
+ * view_array_append - append views that match criteria to array
+ * @server: server context
+ * @views: arrays to append to
+ * @criteria: criteria to match against
+ *
+ * Note: This array has a very short shelf-life so it is intended to be used
+ *       with a single-use-throw-away approach.
+ *
+ * Example usage:
+ *
+ *	struct view **view;
+ *	struct wl_array views;
+ *	wl_array_init(&views);
+ *	view_array_append(server, &views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE);
+ *	wl_array_for_each(view, &views) {
+ *		// Do something with *view
+ *	}
+ *	wl_array_release(&views);
+ */
+void view_array_append(struct server *server, struct wl_array *views,
+	enum lab_view_criteria criteria);
+
 bool view_inhibits_keybinds(struct view *view);
 void view_toggle_keybinds(struct view *view);
 
--- a/include/view.h
+++ b/include/view.h
@@ -161,23 +161,54 @@ struct xdg_toplevel_view {
 };
 
 enum lab_view_criteria {
-	LAB_VIEW_CRITERIA_ALL = 0,
+	LAB_VIEW_CRITERIA_NONE = 0,
 	LAB_VIEW_CRITERIA_CURRENT_WORKSPACE = 1 << 0,
 	LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP = 1 << 1,
 	LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER = 1 << 2,
 };
 
 /**
- * view_array_append - append views that match criteria to array
+ * for_each_view() - iterate over all views which match criteria
+ * @view: Iterator.
+ * @head: Head of list to iterate over.
+ * @criteria: Criteria to match against.
+ * Example:
+ *	struct view *view;
+ *	for_each_view(view, &server->views, LAB_VIEW_CRITERIA_NONE) {
+ *		printf("%s\n", view_get_string_prop(view, "app_id"));
+ *	}
+ */
+#define for_each_view(view, head, criteria)		\
+	for (view = view_next(head, NULL, criteria);	\
+	     view;					\
+	     view = view_next(head, view, criteria))
+
+/**
+ * view_next() - Get next view which matches criteria.
+ * @head: Head of list to iterate over.
+ * @view: Current view from which to find the next one. If NULL is provided as
+ *	  the view argument, the start of the list will be used.
+ * @criteria: Criteria to match against.
+ *
+ * Returns NULL if there are no views matching the criteria.
+ */
+struct view *view_next(struct wl_list *head, struct view *view,
+	enum lab_view_criteria criteria);
+
+/**
+ * view_array_append() - Append views that match criteria to array
  * @server: server context
  * @views: arrays to append to
  * @criteria: criteria to match against
+ *
+ * This function is useful in cases where the calling function may change the
+ * stacking order or where it needs to iterate over the views multiple times,
+ * for example to get the number of views before processing them.
  *
  * Note: This array has a very short shelf-life so it is intended to be used
  *       with a single-use-throw-away approach.
  *
  * Example usage:
- *
  *	struct view **view;
  *	struct wl_array views;
  *	wl_array_init(&views);
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -435,3 +435,4 @@ get_cursor_context(struct server *server)
 	wlr_log(WLR_ERROR, "Unknown node detected");
 	return ret;
 }
+
--- a/src/view.c
+++ b/src/view.c
@@ -18,44 +18,63 @@
 #define LAB_FALLBACK_WIDTH  640
 #define LAB_FALLBACK_HEIGHT 480
 
+static bool
+matches_criteria(struct view *view, enum lab_view_criteria criteria)
+{
+	if (!isfocusable(view)) {
+		return false;
+	}
+	if (criteria & LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+		/*
+		 * Always-on-top views are always on the current desktop and are
+		 * special in that they live in a different tree.
+		 */
+		struct server *server = view->server;
+		if (view->scene_tree->node.parent != server->workspace_current->tree
+				&& !view_is_always_on_top(view)) {
+			return false;
+		}
+	}
+	if (criteria & LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP) {
+		if (view_is_always_on_top(view)) {
+			return false;
+		}
+	}
+	if (criteria & LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER) {
+		if (window_rules_get_property(view, "skipWindowSwitcher") == LAB_PROP_TRUE) {
+			return false;
+		}
+	}
+	return true;
+}
+
+struct view *
+view_next(struct wl_list *head, struct view *view, enum lab_view_criteria criteria)
+{
+	assert(head);
+
+	struct wl_list *elm = view ? &view->link : head;
+
+	for (elm = elm->next; elm != head; elm = elm->next) {
+		view = wl_container_of(elm, view, link);
+		if (matches_criteria(view, criteria)) {
+			return view;
+		}
+	}
+	return NULL;
+}
+
 void
 view_array_append(struct server *server, struct wl_array *views,
 		enum lab_view_criteria criteria)
 {
 	struct view *view;
-	wl_list_for_each(view, &server->views, link)
-	{
-		if (!isfocusable(view)) {
+	for_each_view(view, &server->views, criteria) {
+		struct view **entry = wl_array_add(views, sizeof(*entry));
+		if (!entry) {
+			wlr_log(WLR_ERROR, "wl_array_add(): out of memory");
 			continue;
 		}
-
-		if (criteria & LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
-			/*
-			 * Always-on-top views are always on the current
-			 * desktop and are special in that they live in a
-			 * different tree.
-			 */
-			if (view_is_always_on_top(view)) {
-				goto next;
-			}
-			if (view->scene_tree->node.parent != server->workspace_current->tree) {
-				continue;
-			}
-		}
-next:
-		if (criteria & LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP) {
-			if (view_is_always_on_top(view)) {
-				continue;
-			}
-		}
-		if (criteria & LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER) {
-			if (window_rules_get_property(view, "skipWindowSwitcher")
-					== LAB_PROP_TRUE) {
-				continue;
-			}
-		}
-
-		struct view **entry = wl_array_add(views, sizeof(*entry));
 		*entry = view;
 	}
 }
--- a/src/view.c
+++ b/src/view.c
@@ -18,6 +18,48 @@
 #define LAB_FALLBACK_WIDTH  640
 #define LAB_FALLBACK_HEIGHT 480
 
+void
+view_array_append(struct server *server, struct wl_array *views,
+		enum lab_view_criteria criteria)
+{
+	struct view *view;
+	wl_list_for_each(view, &server->views, link)
+	{
+		if (!isfocusable(view)) {
+			continue;
+		}
+
+		if (criteria & LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+			/*
+			 * Always-on-top views are always on the current
+			 * desktop and are special in that they live in a
+			 * different tree.
+			 */
+			if (view_is_always_on_top(view)) {
+				goto next;
+			}
+			if (view->scene_tree->node.parent != server->workspace_current->tree) {
+				continue;
+			}
+		}
+next:
+		if (criteria & LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP) {
+			if (view_is_always_on_top(view)) {
+				continue;
+			}
+		}
+		if (criteria & LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER) {
+			if (window_rules_get_property(view, "skipWindowSwitcher")
+					== LAB_PROP_TRUE) {
+				continue;
+			}
+		}
+
+		struct view **entry = wl_array_add(views, sizeof(*entry));
+		*entry = view;
+	}
+}
+
 /**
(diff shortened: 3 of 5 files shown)


Consolatis:
The new filter API looks great!
It also seems very easy to extend with further criteria in the future. Me likes.

What do you think about having `osd_update()` do the `desktop_views_append()` call once in the `if (rc.window_switcher.show) {` condition and then supply the array to `display_osd`? That could then in turn also supply the array to `get_osd_height()` and `render_osd()`. So we'd only have to fetch a single consistent state which is then used everywhere. Currently I think we are filling a new array two times for every output.

With this we could also think about integrating `get_osd_height()` into `display_osd()` directly as its just a single multiplication and `+= 2 * border_width + 2 * padding` now.

johanmalm:
Very good suggestion. Makes for much cleaner code. The power of reviews!
Will try to sort over the next couple of days

johanmalm:
@Consolatis Have included your suggestions. Although did the alloc/free in display_osd()

Could consider doing a wl_array_len() helper, but suggest we do that as a separate PR

src/desktop.c
```diff
@@ -434,3 +434,45 @@ get_cursor_context(struct server *server)
 	wlr_log(WLR_ERROR, "Unknown node detected");
 	return ret;
 }
+
+void
+desktop_views_append(struct server *server, struct wl_array *views,
```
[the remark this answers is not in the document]
Consolatis: Within the codebase, usually `desktop` refers to actions involving multiple views. `workspace` however refers to what most people know as "virtual desktop".

I agree though that the `_views_append()` part might be a bit too specific although it matches what the function does: adding view pointers to a supplied array. `desktop_append_views_to_array()` would be even more specific but should make clear what the function does. I think having the `append` part in there makes sense though, otherwise a dev might expect the function to initialize or clean the array before filling it (which we could actually do but it would restrict its uses in the future as well).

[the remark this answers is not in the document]
heroin-moose: The more I look at the code the more I suspect that it perhaps would be more natural do something like this:

```
view_for_each(view, &server->views, criteria) {
    [...]
}
```

That's a common pattern in the Linux kernel:

```
kvm_for_each_vcpu(i, vcpu, kvm) {}
for_each_online_cpu(cpu) {}
for_each_lpi_its {}
for_each_pcm_streams(stream) {}
```

Dunno, just my five cents :)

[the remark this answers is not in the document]
Consolatis: That would indeed be a valid alternative approach, would involve macros though.

src/desktop.c
```diff
@@ -434,3 +434,45 @@ get_cursor_context(struct server *server)
...
+			}
+		}
+		if (criteria & LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER) {
+			if (window_rules_get_property(view, "skipWindowSwitcher")
+					== LAB_PROP_TRUE) {
+				continue;
+			}
+		}
+
+		struct view **entry = wl_array_add(views, sizeof(*entry));
```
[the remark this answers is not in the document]
Consolatis: The `wl_list` type requires the structs to have a `struct wl_list` entry for every list they are a part of. `wl_array` does not.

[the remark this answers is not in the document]
johanmalm: What @Consolatis said + with `wl_list_for_each` we have an edge case where a view operation changes the stacking order - whereas with an allocated list, that edge-case disappears.

src/desktop.c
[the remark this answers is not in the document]
johanmalm: Looks like they invoke `for_each_set_bit` in https://elixir.bootlin.com/linux/latest/source/tools/include/linux/bitops.h#L52
(although didn't look at all).

I prefer to approach we've taken. Avoids macro-bashing which I'm scared of.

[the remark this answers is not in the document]
Consolatis: Alright, lets keep it at `wl_array` then.

Should we change the name of the function though?
I totally get the point that `desktop_views_append()` may sound like appending views to a desktop which is obviously not the case here. So far I still prefer `desktop_append_views_to_array()` the most but maybe there is a better self-explaining name.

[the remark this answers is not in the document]
johanmalm: Yes, agree. `desktop_append_views_to_array()` is better.

We could consider putting the function in a dedicated translation unit and thereby drop the `desktop_`. How about `views.c` with `views_add()`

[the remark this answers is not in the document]
heroin-moose: Why not view.c?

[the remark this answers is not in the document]
heroin-moose: Also in our case no complex macro bashing required:

```
#define server_for_each_view(server, ptr, criteria) \
for (ptr = view_iterate(&server->views, ptr, criteria); ptr != NULL; ptr = view_iterate(ptr, ptr, criteria))
```

But I guess view rearrangement is really a strong argument for keeping things in array.

src/desktop.c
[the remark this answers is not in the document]
heroin-moose: > The wl_list type requires the structs to have a struct wl_list entry for every list they are a part of. wl_array does not.

Right, we are not taking the elements off the view list.

johanmalm:
So how about we do something like this:

1. Add a `view_next()` and move all the criteria stuff to it as well as the `isfocusable()` check.

2. Add the `view_for_each()` macro.

3. `s/desktop_views_append/view_array_append/` and move it to `view.c`

The array approach will be useful when stacking order changes or the user needs to iterate multiple times for example to get the number of items first and then processing them separately. But in other cases a simple `view_for_each(server, view, criteria)` is simpler and avoids the allocation.

`view_array_append()` can then just call the `view_for_each` macro.

Consolatis:
> Add a view_next() and move all the criteria stuff to it as well as the isfocusable() check.

That sounds a bit like what we have for `desktop_cycle_view()`. If we could abstract that away completely that would actually be awesome as we then could also get rid of the always-on-top and on-different-workspace limitation.

In any case (even without merging with `desktop_cycle_view()`): /me likes.

johanmalm:
Have added for_each_view() with a criteria argument.
Welcome review if you have time.
I couldn't think of an elegant way of combining `view_first()` and `view_next()` so have kept them separate.
It'd be good to add direction (as in `desktop_cycle_view()` as @Consolatis suggested - but could do that as a separate PR.

src/view.c
```diff
@@ -18,43 +18,53 @@
...
-			}
-			if (view->scene_tree->node.parent != server->workspace_current->tree) {
-				continue;
-			}
+	view = wl_container_of(view->link.next, view, link);
+	if (&view->link == &server->views) {
+		return view;
+	}
+	if (!isfocusable(view)) {
+		return view_next(view, criteria);
```
heroin-moose: Being a paranoid here, but given enough !isfocusable() views a malicious user can overflow the stack. Can we do something about it?

include/view.h
```diff
@@ -167,6 +167,14 @@ enum lab_view_criteria {
 	LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER = 1 << 2,
 };
 
+
+#define for_each_view(head, view, criteria)				\
+	for (view = 0, view = wl_container_of(head.next, view, link);	\
```
heroin-moose: view = NULL?

src/view.c
```diff
@@ -18,43 +18,53 @@
...
-			 * desktop and are special in that they live in a
-			 * different tree.
-			 */
-			if (view_is_always_on_top(view)) {
-				goto next;
-			}
-			if (view->scene_tree->node.parent != server->workspace_current->tree) {
-				continue;
-			}
+	view = wl_container_of(view->link.next, view, link);
```
heroin-moose: Honestly I think a do { } while () thing here would be better than recursion dependable on controllable global state.

src/view.c
```diff
@@ -18,6 +18,80 @@
...
+			return false;
+		}
+	}
+	return true;
+}
+
+struct view *
+view_first(struct wl_list *head, enum lab_view_criteria criteria)
+{
+	struct view *view = wl_container_of(head->next, view, link);
```
Consolatis, line 54: AFAIR in an empty list, `head->next` would be `head` itself. Thus its not actually a view but `server->views` in this case. So I think we need something like `if (head->next == head) { return NULL; }` as first statement here.

The reason this didn't actually bit us here was that we currently still return early in `osd_update()` on `wl_list_empty(node_list))` with `node_list` being the current workspace scene_graph child nodes (or on `!server->cycle_view`). Thus this function was never called with an empty view list.

include/view.h
```diff
@@ -160,6 +160,66 @@ struct xdg_toplevel_view {
...
+/**
+ * for_each_view() - iterate over all views which match criteria
+ * @head: list head to iterate over
+ * @view: iterator
+ * @criteria: criteria to match against
+ */
+#define for_each_view(head, view, criteria)		\
+	for (view = view_first(head, criteria);		\
+		&view->link != (head);			\
+		view = view_next(view, criteria))
```
Consolatis, line 179: Both of the view_ variants may return `NULL` based on the docs so the condition must contain a view check. I am not sure if the wrongly casted `* server->views` to a `struct view *` works correctly either. Maybe the condition could just be `view` here? Then we'd need to supply `head` as another argument to `view_next()` as well and that one could then do the check internally and return `NULL` if we reached the list head.

src/view.c
```diff
@@ -18,6 +18,80 @@
...
+		}
+	}
+	return view;
+}
+
+struct view *
+view_next(struct view *view, enum lab_view_criteria criteria)
+{
+	assert(view);
+	struct server *server = view->server;
```
Consolatis, line 72: And similar here when we add the additional `head` argument, something like
`if (view->link.next == head) { return NULL; }` as first statement (after the `assert()`).

Alternatively we could also not add the `head` argument but instead do the check like
`view->link.next == &view->server->views`.

Same check would have to be done within the loop body below in this function, e.g. before using `wl_container_of()` rather than after it.

Consolatis, line 72: And similar here when we add the additional `head` argument, something like
`if (view->link.next == head) { return NULL; }` as first statement (after the `assert()`).

Alternatively we could also not add the `head` argument but instead do the check like
`view->link.next == &view->server->views`.

Same check would have to be done within the loop body below in this function, e.g. before using `wl_container_of()` rather than after it.

src/view.c
Consolatis: Although this seems to reference and outdated state, the same comment still applies to the current state. Maybe we could just throw it into a `for(;;) {` loop, similar to `view_first()`?

include/view.h
Consolatis: Seems to refer an outdated state.

Consolatis:
Sorry for the slightly unorganized review which repeats the same issue multiple times, should have done the proper review variant instead of using single comments.

src/view.c
```diff
@@ -18,6 +18,80 @@
...
+	}
+	return view;
+}
+
+void
+view_array_append(struct server *server, struct wl_array *views,
+		enum lab_view_criteria criteria)
+{
+	struct view *view;
+	for_each_view(&server->views, view, criteria) {
```
Consolatis, line 89: I think `wl_list_for_each()` has its arguments ordered different, first the assignment arg (`view` in this case), then the list itself. Suggest ordering the arguments the same for this API to reduce confusion when using it.

src/view.c
johanmalm: Good feedback. Loop is much better than recursion here. Have implemented.

src/view.c
johanmalm: Agree. Have implemented, albeit with a `for(;;)`

src/view.c
johanmalm: Good point. Returning a view didn't feel right! Have implemented.

Have tested with the code below to prove that we deal with an empty list properly.

```
index c04e1c5..cd9a35d 100644
--- a/src/action.c
+++ b/src/action.c
@@ -557,7 +557,10 @@ actions_run(struct view *activator, struct server *server,
 			}
 			break;
 		case ACTION_TYPE_DEBUG:
-			debug_dump_scene(server);
+			for_each_view(view, &server->views, 0) {
+				wlr_log(WLR_ERROR, "id=%s", view_get_string_prop(view, "app_id"));
+			}
+			// debug_dump_scene(server);
 			break;
 		case ACTION_TYPE_EXECUTE:
 			{
```

include/view.h
johanmalm: I've combined `view_next()` + `view_first()` now. Getting rid of that recursion helped me see the similarity and how to deal with the first item in the list.

I think I've covered what you've said here.

The middle condition of the for loop needs to be `view && &view->link != (head)` otherwise we don't handle the case of an empty list properly. We could probably add a `wl_list_length()` check in `view_next()` instead, but I think it's simpler like this.

src/view.c
johanmalm: Done

johanmalm:
> > I couldn't think of an elegant way of combining `view_first()` and `view_next()` so have kept them separate.
> 
> I think in `desktop_cycle_view()` we are doing that by only using the `view_next()` variant and supplying a `NULL` for the reference view (+ adding an additional `head` pointer). Then in the beginning of the function - if the supplied `view` argument is `NULL` - we could return the first item (unless its the list head itself, in that case `NULL`).

Thanks, yes it became clear when I got rid of the recursion.

src/view.c
johanmalm: Done. Good point.

johanmalm:
Thanks for great feedback. I was obviously not quite on form last night.

Would appreciate a 'final' scan through before I `git rebase -i ...`. I feels like it's come together quite nicely.

The following associated changes remain, but I suggest we do them as separate PRs

- [ ] Move `isfocusable()` from `desktop.c` to `view.c`
- [ ] Add direction enum to `view_next()` and add a for_each_view_reverse()
- [ ] Add wl_array_len() macro
- [ ] Review all instances of `wl_list_for_each(view, &server->views, link)` and take a view on using for_each_view()

src/view.c
```diff
@@ -18,6 +18,67 @@
...
+		if (&view->link == head) {
+			return NULL;
+		}
+inside:
+		if (matches_criteria(view, criteria)) {
+			break;
+		}
+	}
+	return view;
+}
```
Consolatis, line 69: In the `!view` case and an empty list we still cast `&server->views` to `struct view *` here. While the access to the `->link` member might technically be fine, accessing any other part of the result will likely read random memory.

I'd suggest something like this instead:
```c
struct view *
view_next(struct wl_list *head, struct view *view, enum lab_view_criteria criteria)
{
	struct wl_list *ref_item = view ? &view->link : head;
	for (;;) {
		ref_item = ref_item->next;
		if (ref_item == head) {
			return NULL;
		}
		view = wl_container_of(ref_item, view, link);
		if (matches_criteria(view, criteria)) {
			return view;
		}
	}
	return NULL; /* Should never be reached */
}
```

include/view.h
Consolatis: > The middle condition of the for loop needs to be `view && &view->link != (head)` otherwise we don't handle the case of an empty list properly. We could probably add a `wl_list_length()` check in `view_next()` instead, but I think it's simpler like this.

I don't see why `view != NULL` wouldn't work here?

src/view.c
johanmalm: Well... you are a complete genius. I'm not sure why I couldn't get to that.
This let's us reduce the middle bit of the `for()` to a mere `view` too which is tidy.
Big thanks.

include/view.h
johanmalm: Done

johanmalm:
Okay. Rebased. Ready for final check.

src/view.c
Consolatis: Heh, just had experienced the same thing (working with the list items directly rather than the struct they are contained in) the hard way before :)

src/view.c
```diff
@@ -18,6 +18,66 @@
...
+	return NULL; /* Should never be reached */
+}
+
+void
+view_array_append(struct server *server, struct wl_array *views,
+		enum lab_view_criteria criteria)
+{
+	struct view *view;
+	for_each_view(view, &server->views, criteria) {
+		struct view **entry = wl_array_add(views, sizeof(*entry));
```
heroin-moose, line 76: wl_array_add can fail. Dunno if we wanna recover or not, but explicit assert IMO would be better.

src/view.c
```diff
@@ -18,6 +18,66 @@
...
+	return true;
+}
+
+struct view *
+view_next(struct wl_list *head, struct view *view, enum lab_view_criteria criteria)
+{
+	assert(head);
+
+	struct wl_list *ref_item = view ? &view->link : head;
+	for (;;) {
```
heroin-moose, line 57: Given the "Should never be reached" I think it would be better to actually reach it with this:

while (ref_item != head) {
 ...
}

heroin-moose, line 57: It would be better to actually exit the cycle and return:

```
struct view *
view_next(struct wl_list *head, struct view *view, enum lab_view_criteria criteria)
{
	assert(head);

	struct wl_list *ptr = view ? &view->link : head;

	for (ptr = ptr->next; ptr != head; ptr = ptr->next) {
		view = wl_container_of(ptr, view, link);
		if (matches_criteria(view, criteria)) {
			return view;
		}
	}
	return NULL;
}
```

include/view.h
```diff
@@ -160,6 +160,66 @@ struct xdg_toplevel_view {
...
+ * @criteria: Criteria to match against.
+ * Example:
+ *	struct view *view;
+ *	for_each_view(view, &server->views, LAB_VIEW_CRITERIA_NONE) {
+ *		printf("%s\n", view_get_string_prop(view, "app_id"));
+ *	}
+ */
+#define for_each_view(view, head, criteria)			\
+	for (view = view_next(head, NULL, criteria); view;	\
+		view = view_next(head, view, criteria))
```
heroin-moose, line 183: My OCD kicks in with "indent with tabs align with spaces" and

```C
	for (view = view_next(head, NULL, criteria);
	     view;
	     view = view_next(head, view, criteria))
```

IMO it would be better, but this is purely my style suggestions.

src/view.c
Consolatis: Not sure how it would be better but the code definitely looks better. +1

src/view.c
johanmalm: Done

src/view.c
johanmalm: Elegant! Implemented.

include/view.h
johanmalm: Well you know... we don't normally align with spaces (Devault coding style) - but checkpatch.pl does not complain and it does look nicer :smile:
Implemented.

Consolatis:
LGTM, testing (before the last refactor) didn't show any issues and I don't suspect the last changes to break something.
Nice PR, thanks for working on this @johanmalm :)

PR #1028, opened 2023-08-12
Add `wl_array_len()`

johanmalm:
Thoughts?


Consolatis:
No issues with the helper. Depends on #1024 though. So the question of keep using the array or using a for_each macro instead.

Issue #1030, opened 2023-08-14
Menu items can no longer be activated on xfce4-panel


Tamaranch:
Clicking on a menu item closes the menu but nothing happens (any item in any menu, e.g. opening panel preferences from a context menu). I've identified this commit as the culprit: 3111c57d6f18d07e1e8b64e2ce396d15efc73b5c.

Consolatis:
Ref:
- #1020

If you are up for testing, does adding a check for the layer surface to the condition fix it?
```c
	if (ctx.type != LAB_SSD_CLIENT && ctx.type != LAB_SSD_LAYER_SURFACE
			&& wlr_seat_pointer_has_grab(seat->seat)) {
		[..]
	}
```

Tamaranch:
> does adding a check for the layer surface to the condition fix it?

No, it doesn't.

> does xfce4-panel happen to have custom context menus rather than the usual GTK ones?

No it uses GtkMenu but it has keyboard interactivity set to on-demand, maybe that matters?

Consolatis:
> > does adding a check for the layer surface to the condition fix it?
>
> No, it doesn't.

Hm.. I don't have a recent xfce4-panel installation around (still on Debian oldstable).
We'd need to know what `ctx.type` is triggering the `wlr_seat_pointer_end_grab()` call and the proper solution would then likely be to figure out why that type isn't `LAB_SSD_CLIENT` or `LAB_SSD_LAYER_SURFACE`.
Maybe @johanmalm could help with some debugging here?

For a quick manual fix it is likely enough to revert 3111c57d6f18d07e1e8b64e2ce396d15efc73b5c for now so popup grabs are only killed when clicking on the empty desktop (or other application surfaces).

Tamaranch:
> For a quick manual fix it is likely enough to revert https://github.com/labwc/labwc/commit/3111c57d6f18d07e1e8b64e2ce396d15efc73b5c

I can confirm it fixes the issue.

johanmalm:
Thanks for reporting.

> Hm.. I don't have a recent xfce4-panel installation around (still on Debian oldstable). We'd need to know what `ctx.type` is triggering the `wlr_seat_pointer_end_grab()` call and the proper solution would then likely be to figure out why that type isn't `LAB_SSD_CLIENT` or `LAB_SSD_LAYER_SURFACE`. Maybe @johanmalm could help with some debugging here?

[...]

PR #1031, opened 2023-08-14
cursor: prevent press on layer-subsurface from cancelling popup grab

johanmalm:
Hopefully fixes: #1030

Would appreciate some testing for unintended consequences.


Tamaranch:
I can confirm it fixes the issue, apparently without side effects.

01micko:
No apparent side effects running `labwc/sfwbar` either.

Consolatis:
Didn't see any issues either. LGTM, nice PR :)

johanmalm:
> Edit: One thing I discovered though: Even without requesting keyboard focus, panels now react to `ESC` to close context menus. Which may or may not be a good thing. I am not sure. To fix that we'd have to also supply (or get) the underlying layershell surface and then check that for its keyboard interactivity setting. I kind of like this though for panels.. it works similar to `ON_DEMAND` now.

Yes, great observation/analysis. I did ponder this. Agree with you though - let's go as is. If it becomes a problem we can find top-parent and check on-demand.

johanmalm:
And thanks everyone for testing :smile:

PR #1039, opened 2023-08-20
config: s/app_id/identifier/ for window-switcher field

johanmalm:
...to be consistent with window rules.

The use of 'app_id' for window-switcher fields is hereby deprecated.


Consolatis:
LGTM

PR #1042, opened 2023-08-21
Support svg buttons

johanmalm:


src/button/common.c
```diff
@@ -0,0 +1,13 @@
...
+#include "common/dir.h"
+#include "config/rcxml.h"
+
+char *
+button_filename(const char *name)
+{
+	static char buf[4096] = { 0 };
+	snprintf(buf, sizeof(buf), "%s/%s", theme_dir(rc.theme_name), name);
+	return buf;
+}
```
Consolatis, line 13: I am not sure about returning statically allocated memory here. If we want to keep doing that it should be made very clear in the header doc that this is the case (e.g. `xstrdup()` when wanting to store / never `free()`).

Nitpick: the header calls the argument `filename` which doesn't match the implementation.

Will look into the PR in more detail tomorrow.

Consolatis:
Regarding the CI: Based on a [quick research](https://repology.org/project/librsvg/versions) and some [manual digging](https://tracker.debian.org/pkg/librsvg) for Debian it seems these are the correct dep names:
```
Void: librsvg-devel
FreeBSD: librsvg2
Arch: librsvg
Debian: librsvg2-dev
```

src/button/button-svg.c
```diff
@@ -0,0 +1,67 @@
...
+
+	cairo_surface_t *image = cairo_image_surface_create(CAIRO_FORMAT_ARGB32, size, size);
+
+	cairo_t *cr = cairo_create(image);
+	rsvg_handle_render_document(svg, cr, &viewport, &err);
+	if (err) {
+		wlr_log(WLR_ERROR, "error rendering svg %s-%s\n", filename, err->message);
+		g_error_free(err);
+		return;
+	}
```
Consolatis, line 50: No need to free cairo stuff in the err condition?

src/button/button-svg.c
```diff
@@ -0,0 +1,67 @@
...
+	cairo_surface_flush(image);
+
+	double w = cairo_image_surface_get_width(image);
+	double h = cairo_image_surface_get_height(image);
+	*buffer = buffer_create_cairo((int)w, (int)h, 1.0, true);
+	cairo_t *cairo = (*buffer)->cairo;
+	cairo_set_source_surface(cairo, image, 0, 0);
+	cairo_paint_with_alpha(cairo, 1.0);
+}
```
Consolatis, line 67: Similar, no need to free the temporary cairo image here?

src/button/common.c
johanmalm: I have `s/filename/name`. Good catch.

I don't see the problem with returning a static variable. I've moved it to translation unit level - but I think either way is fine.

It's like doing `return "foo"`. The calling function can use it but must not try to free() it.

I'm more worried about any function that allocates memory on the heap - because the caller much free it.

We have a few more of these in other places. If memory was super tight I guess we could declare a buf[4096] in the calling function and just pass the pointer to `button_filename()`

johanmalm:
> Regarding the CI: Based on a [quick research](https://repology.org/project/librsvg/versions) and some [manual digging for Debian](https://tracker.debian.org/pkg/librsvg) it seems these are the correct dep names:
> 
> ```
> Void: librsvg-devel
> FreeBSD: librsvg2
> Arch: librsvg
> Debian: librsvg2-dev
> ```

Awesome. Updated.

src/button/button-svg.c
johanmalm: Good catch. Done.

src/button/button-svg.c
johanmalm: Good catch. Done.

src/button/common.c
Consolatis: My issue is more about not necessarily knowing what happens when calling some function that returns a `char *`, is it returning static memory or heap memory? Maybe we should go over all the public function that return either and specify it in the header files.

Returning statically allocated memory here seems fine, for me its really more about the question that come with it when using:
- "Is it leaking memory if I don't free it?"
- "Can is store a pointer to this somewhere or will the content get overwritten by future invocations?"

For me this then usually means that I usually look into the implementation to answer these questions but when somebody using a function like that the first time, that person might expect a specific behavior which may not be the case.

01micko:
I get the `warning: "HAVE_RSVG" is not defined, evaluates to 0` warning but build succeeds so will test shortly.

src/theme.c
```diff
@@ -25,7 +25,11 @@
 #include "common/string-helpers.h"
 #include "config/rcxml.h"
 #include "button/button-png.h"
+
```
01micko, line 28: We need here `#include "config.h"`

01micko: ..or probably the line above.

src/theme.c
```diff
@@ -94,6 +95,7 @@ load_buttons(struct theme *theme)
 	};
 
 	char filename[4096] = {0};
```
01micko, line 101: `#if HAVE_RSVG` needed here

01micko: Not done in latest revision

01micko (commented):
'auto' doesn't seem to work though I'm sure with some more tests in `meson.build` it can. ATM I have to explicitly enable `-Dsvg=enabled`. I don't care but is that intentional?

Once I got things building correctly it works quite well. Nice :)

01micko:
Fancy maccish buttons from [here](https://github.com/puppylinux-woof-CE/woof-CE/tree/testing/woof-code/rootfs-packages/jwm_config/usr/share/jwm/themes_window_buttons/Maccish).

fancy_maccish-buttons

johanmalm:
Great work @01micko 
I didn’t understand where I was going wrong. 
Didn’t think of that config.h
Will try to fix tonight.

01micko:
> Didn’t think of that config.h

Again, the value of peer review :)

01micko:
> 'auto' doesn't seem to work though I'm sure with some more tests in `meson.build` it can. ATM I have to explicitly enable `-Dsvg=enabled`. I don't care but is that intentional?
> 
> Once I got things building correctly it works quite well. Nice :)

I see it is intentional :face_palm:

```meson
svg = dependency('librsvg-2.0', version: '>=2.46', required: get_option('svg'))
```

(the 3rd arg)

src/theme.c
01micko: ..maybe it should go in `include/theme.h` ? Perhaps future proofing?

src/button/common.c
heroin-moose: This is a really bad practice. Somewhere in the future we will call this function twice and overwrite the previous result.

Consolatis: I don't actually think its bad practice at all. Callers just need to know if they have to `strdup()` themselves or not. Same for `free()` when not using statically allocated memory.

heroin-moose: We had this issue with multiple functions in libc over the last 20 years. This is really, really, really bad idea.

johanmalm: Me guilty here. Will fix. Will probably put buf at highest frame. Need to have dinner first though :smile:

johanmalm: `static char buf[4096] = { 0 };` fixed.

src/theme.c
johanmalm: Fixed. Thanks for finding that.

01micko:
Works OOTB :+1:

src/button/button-xbm.c
```diff
@@ -278,7 +271,9 @@ button_xbm_load(const char *filename, struct lab_data_buffer **buffer,
 	color = u32(rgba);
 
 	/* Read file into memory as it's easier to tokenzie that way */
-	char *token_buffer = grab_file(xbm_path(filename));
+	char filename[4096] = { 0 };
+	button_filename(button_name, filename, sizeof(filename));
```
heroin-moose, line 275: Just a suggestion: can we start using `ARRAY_SIZE()` instead of `sizeof()` when array len is involved?

include/button/common.h
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * button_filename() - Get full filename for button.
+ * @name: The name of the button (for example 'iconify.xbm').
+ * @buf: Buffer to fill with the full filename
+ * @len: Length of buffer
+ *
+ * Example return value: /usr/share/themes/Numix/openbox-3/iconfify.xbm
+ */
+void button_filename(const char *name, char *buf, int len);
```
heroin-moose, line 13: size_t is a better type for that.

src/common/file-helpers.c
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <sys/stat.h>
+#include "common/file-helpers.h"
+
+bool
+file_exists(const char *filename)
```
heroin-moose, line 6: Can't we open and check the return code? Using stat() always leads to TOCTOU race. The severity of the race ranges from "minor" to "hurr durr I've rooted your PC".

johanmalm: Yes I think that would work. Just have to make the button_*_load function return a bool.

This thing is a new one on me. Do you get that toctou problem even if you stat and then just open for read as a normal user?

heroin-moose: That's the inherent problem of having two separate calls, you can't work around it. It's not gonna cause any bugs here, but somewhat later it would be tempting use this functions somewhere where it should not be used.

johanmalm: Got it

src/button/button-xbm.c
johanmalm: Yes, I'm up for that.
IIRC we started off like that and then read the Devault coding style and decided to follow.

I think we even disabled it in checkpatch.pl so if we turn it back on we should find the places to tweak.

Can't grep right now, but think we even define it locally somewhere.

Question is, do we go with a simple version or one that does type checking (which I think Linux does). Not sure if the latter need GNU_SOURCE which we've so far avoided.

heroin-moose: That's the kernel one:
```
#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]) + __must_be_array(arr))
```

Magic happens here:
```
#define __must_be_array(a)	BUILD_BUG_ON_ZERO(__same_type((a), &(a)[0]))
```

And it relies on `__builtin_types_compatible_p()`:
```

#define __same_type(a, b) __builtin_types_compatible_p(typeof(a), typeof(b))
```

Both GCC and clang support this builtin. If you have any suspicions about some system that does not have this builtin, we can introduce #ifdef guard.

johanmalm: See #1086

include/button/common.h
johanmalm: Good point. Done.

src/common/file-helpers.c
johanmalm: Done

johanmalm:
Ready for final review + merge from my perspective.

01micko:
Works well.
A suggestion [here](https://github.com/labwc/labwc/commit/486365a2a43831e5ce74d9a3aa60926a21909191#r127550075)

01micko:
> Works well. A suggestion [here](https://github.com/labwc/labwc/commit/486365a2a43831e5ce74d9a3aa60926a21909191#r127550075)

In addition for the 'feature' to work correctly if `-Dsvg=diabled`: 

Line 116 `meson.build` _and_ `src/button/meson.build`

```
s/if svg.found()/if svg.found() and not get_option('svg').disabled()/
```

johanmalm:
> > Works well. A suggestion [here](https://github.com/labwc/labwc/commit/486365a2a43831e5ce74d9a3aa60926a21909191#r127550075)
> 
> In addition for the 'feature' to work correctly if `-Dsvg=disabled`:
> 
> Line 116 `meson.build` _and_ `src/button/meson.build`
> 
> ```
> s/if svg.found()/if svg.found() and not get_option('svg').disabled()/
> ```

@01micko Thanks for working out the logic here. My mind was just drawing a blank yesterday.
Have added patch to allow user to disable svg even if the right version of `librsvg` exists.
I set value=enabled because auto falls back to false and I prefer maintainers not to have to add `-Dsvg=enabled` because some are likely to forget :smile:

01micko:
Works great! Tested all scenarios I could think of.

src/button/button-svg.c
```diff
@@ -0,0 +1,68 @@
...
+	char filename[4096] = { 0 };
+	button_filename(button_name, filename, sizeof(filename));
+
+	GError *err = NULL;
+	RsvgRectangle viewport = { .width = size, .height = size };
+	RsvgHandle *svg = rsvg_handle_new_from_file(filename, &err);
+	if (err) {
+		wlr_log(WLR_DEBUG, "error reading svg %s-%s\n", filename, err->message);
+		g_error_free(err);
+		/* No need to free svg here */
```
Consolatis, line 37: Maybe add something like `rsvg_handle_new_from_file() returns NULL on error` here.

src/button/button-svg.c
```diff
@@ -0,0 +1,68 @@
...
+
+	if (cairo_surface_status(image)) {
+		wlr_log(WLR_ERROR, "error reading svg button '%s'", filename);
+		goto error;
+	}
+	cairo_surface_flush(image);
+
+	double w = cairo_image_surface_get_width(image);
+	double h = cairo_image_surface_get_height(image);
+	*buffer = buffer_create_cairo((int)w, (int)h, 1.0, true);
```
Consolatis, line 59: Suggest adding an inline comment `/* free_on_destroy */ true)` here.

src/theme.c
```diff
@@ -94,16 +109,34 @@ load_buttons(struct theme *theme)
...
+#if HAVE_RSVG
+	int size = theme->title_height - 2 * theme->padding_height;
+#endif
```
Consolatis, line 114: Should we move the size down into the 2nd `HAVE_RSVG` block?

meson.build
```diff
@@ -106,6 +114,11 @@ labwc_deps = [
   math,
   png,
 ]
+if svg.found()
+  labwc_deps += [
+    svg,
+  ]
+endif
```
Consolatis, line 121: Should this be `if have_rsvg` rather than `svg.found()`? Otherwise it appears we are linking librsvg if its found but disabled by the user.

src/button/button-xbm.c
Consolatis: Using `ARRAY_SIZE` for char arrays feels wrong to me.

Consolatis:
Just some minor nitpicks. Besides those: LGTM.

johanmalm:
> Works great! Tested all scenarios I could think of.

Well, thank you for seeing a sensible solution and letting me give up too easily :smile:

src/button/button-svg.c
johanmalm: Good points. Makes it much clearer. Done.

src/theme.c
johanmalm: Have done. 0.000001% less efficient, but much clearer and cleaner.

meson.build
johanmalm: Oh. Good catch indeed. Done.

src/button/button-xbm.c
johanmalm: Agree. @heroin-moose - is ARRAY_SIZE used for char[123] in the Linux Kernel?

stefonarch:
Tried to test this, but didn't manage. Copied [those](https://github.com/labwc/labwc/pull/1042#issuecomment-1691023932) buttons in some theme folder but `.xbm` buttons are preferred still. No idea what I'm missing.
`Run-time dependency librsvg-2.0 found: YES 2.57.0`

johanmalm:
> Tried to test this, but didn't manage. Copied [those](https://github.com/labwc/labwc/pull/1042#issuecomment-1691023932) buttons in some theme folder but `.xbm` buttons are preferred still. No idea what I'm missing. `Run-time dependency librsvg-2.0 found: YES 2.57.0`

You have to rename them to `close-active.svg` `iconify-active.svg` `max-active.svg`
```

**Comment by stefonarch** [comment]
Thanks, that worked. How could I have known that? The `.xbm` buttons have just max|iconify|close syntax.

**Comment by Consolatis** [comment]
https://labwc.github.io/labwc-theme.5.html#buttons

**Comment by stefonarch** [comment]
I see the "inactive" are needed too. No luck with "bullet{-active} in all formats though.

screen_area_lun_09:39:26_

[macish.buttons.svg.tar.gz](https://github.com/labwc/labwc/files/12645573/macish.buttons.svg.tar.gz)

File: src/button/button-xbm.c (hunk 6 above)
Comment (heroin-moose) [reply src/button/button-xbm.c:275]: For any array.

Comment (heroin-moose) [reply src/button/button-xbm.c:275]:
```
	u8 data[4];
[...]
	ret = i2c_smbus_write_i2c_block_data(chip->client,
					BH1770_ALS_TH_UP_0,
					ARRAY_SIZE(data),
					data);
```

Comment (johanmalm) [reply src/button/button-xbm.c:275]: So in the kernel would you not use `sizeof()` in this example:

```
char filename[4096] = { 0 };
button_filename(button_name, filename, sizeof(filename));
```

01micko:
@stefonarch @johanmalm I've uploaded some themes to [labwc-artwork](https://github.com/labwc/labwc-artwork/tree/main/button-themes/svg)

johanmalm:
@01micko Great stuff. Do we have licences + a note on where they come from to go with these files?

01micko:
> @01micko Great stuff. Do we have licences + a note on where they come from to go with these files?

In the main README it says:

> Artwork here is licensed under [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/).

These icons are part of puppy linux and were created by the team there.

johanmalm:
Thanks. Sounds good.

Issue #1044, opened 2023-08-22
Feature Request: Option to hide SSD borders / disable resize borders on maximized windows


gurrgur:
Currently maximized windows are resizable at the screen corners. The problem with this is that at the screen edges other UI elements such as the SSD buttons or scroll bars are not clickable.

I usually flick the mouse to the top right corner to close a maximized window, so this gets in my way often. Likewise, I often flick the mouse towards the right edge of the screen to grab the scrollbar. This is standard behaviour e.g. on kwin/plasma or windows and its a must have feature in my book.
 
I am aware that using a theme without borders yields the desired behaviour, however borders on unmaximized windows are essential to create contrast between windows that share a similar backgroud color.

Consolatis:
This sounds sensible.

I am not sure if we'd need some setting for this behavior (and what the default should be) as I don't know what the usual expectation of other users would be.

Did some initial work in
- #1045

Feel free to test if you are fine with compiling labwc yourself. When using a single screen the PR should allow for the `flick-mouse-to-corner` and `flick-mouse-to-edge` use-cases.

gurrgur:
Thanks for the quick response. I tested the WIP branch and it works great for both usecases!

IMO this behaviour should be the default behaviour. Having close button and scroll bar more easily accessible far outweights the ability to resize maximized windows via mouse. I mean, we already have the restore button for this. Also if someone really wanted to resize a maximized window, it would be possible to use either a key combination or key-down + mouse movement.

Personally I wouldn't add a setting for this, but in the end it probably depends on whether there is demand for specific functionality.

To get some more context I quickly checked how some other projects are doing it:

- **kwin** uses plugins for drawing SSDs. It is up to the decoration plugin to implement this / expose a setting. The default decoration breeze has a setting for enabling borders on maximized windows and also allows resizing. Resizing from the windows buttons is also possible, however when not moving the mouse, the buttons still respond (despite the mouse showing as a resize cursor).
- **mutter**: Unsurprisingly, no user exposed setting to be found here. The default is no border / resize borders on ssd decorated windows (though it might be possible for a user to add borders through gtk css overrides).
- **xfwm4**: There is a setting that allows drawing borders for maximized windows. It is disabled by default. When enabled, windows can be resized from the edges, but not from where the ssd window buttons are.
- **openbox**: Some time ago I verified that openbox behaves exactly like labwc does currently on master.

Consolatis:
- #1045

[...]

PR #1045, opened 2023-08-22
src/ssd: disable border and un-round corners on maximize

Consolatis:
Partly addresses
- #1044

Main issues with this PR:
- [x] the borders of the rounded corners leak into neighboring screens
- [ ] no config setting to enable / disable the new feature

Regarding the corner button leakage.. I am not sure how to tackle this.
If wlroots 0.16 would already provide the scissor support for scene nodes we could just cut the border away from the buttons. There are two other options I see:
- create an additional no-border variant for every corner button (active / inactive, left / right, border / no-border)
- combine this with the long standing `unround-corners-on-maximize` issue and use a scene graph rect for the background of the corner buttons when maximized

Regarding the config setting: not sure if we need / want one or if we just disable borders / unround corner buttons on maximize

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -62,7 +62,7 @@ struct ssd {
 
 	/* Borders allow resizing as well */
 	struct {
-		/* struct wlr_scene_tree *tree;      unused for now */
+		struct wlr_scene_tree *tree;
 		struct ssd_sub_tree active;
 		struct ssd_sub_tree inactive;
 	} border;
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -17,6 +17,7 @@ struct ssd_button {
 	struct view *view;
 	enum ssd_part_type type;
 	struct wlr_scene_node *hover;
+	struct wlr_scene_node *background;
 
 	struct wl_listener destroy;
 };
@@ -41,6 +42,7 @@ struct ssd {
 	 * don't update things we don't have to.
 	 */
 	struct {
+		bool squared_corners;   /* To un-round corner buttons on maximize */
 		struct wlr_box geometry;
 		struct ssd_state_title {
 			char *text;
--- a/src/ssd/ssd_border.c
+++ b/src/ssd/ssd_border.c
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0-only
 
+#include <assert.h>
 #include "common/scene-helpers.h"
 #include "labwc.h"
 #include "ssd-internal.h"
@@ -13,6 +14,9 @@
 void
 ssd_border_create(struct ssd *ssd)
 {
+	assert(ssd);
+	assert(!ssd->border.tree);
+
 	struct view *view = ssd->view;
 	struct theme *theme = view->server->theme;
 	int width = view->current.width;
@@ -23,10 +27,12 @@ ssd_border_create(struct ssd *ssd)
 	struct wlr_scene_tree *parent;
 	struct ssd_sub_tree *subtree;
 
+	ssd->border.tree = wlr_scene_tree_create(ssd->tree);
+	wlr_scene_node_set_position(&ssd->border.tree->node, -theme->border_width, 0);
+
 	FOR_EACH_STATE(ssd, subtree) {
-		subtree->tree = wlr_scene_tree_create(ssd->tree);
+		subtree->tree = wlr_scene_tree_create(ssd->border.tree);
 		parent = subtree->tree;
-		wlr_scene_node_set_position(&parent->node, -theme->border_width, 0);
 		if (subtree == &ssd->border.active) {
 			color = theme->window_active_border_color;
 		} else {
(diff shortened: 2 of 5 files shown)


Consolatis:
> * combine this with the long standing `unround-corners-on-maximize` issue and use a scene graph rect for the background of the corner buttons when maximized

Went with this variant. Currently feels a bit hacky though because the button scene graph hierarchy leaks very much into `ssd_titlebar.c`. Maybe we need a `src/ssd/ssd_button.c` that combines everything ssd button related into a neat modular API.

include/ssd-internal.h
```diff
@@ -62,6 +63,7 @@ struct ssd {
 
 	/* Borders allow resizing as well */
 	struct {
+		bool is_unround;
```
Consolatis, line 66: Woops, that should end up in `ssd->titlebar` or even better in `ssd->state`.

johanmalm:
Nice work.

I can’t remember - does the corner piece include the menu/close buttons, or are they separate nodes?

Two related issues (just throwing them in so that they’re part of the solutioneeeing):

- Themes should have hover buttons rather than the alpha surface we currently use.
- The user should be able to specify the position/order of buttons.

If buttons/corners are too closely linked in an API this might become more difficult to achieve

Consolatis:
> I can’t remember - does the corner piece include the menu/close buttons, or are they separate nodes?

Separate nodes, the corner pieces only consist of a tree and the rounded corner buffer. On top of that (as 2nd child of the tree) a usual button is used, exactly the same as for all the other buttons, e.g. containing the centered icon and default-disabled hover overlay. The only difference is the background being transparent rather than titlebar color. This PR changes the invisible background to the titlebar color and disables the rounded corner buffer (as it also contains the border itself, otherwise we could just have the background overlay the buffer).

> Themes should have hover buttons rather than the alpha surface we currently use.

That should be doable but also speaks for having a `ssd_button.c` which includes the logic for enabling / disabling the hover effect.

> The user should be able to specify the position/order of buttons.

Should be doable as the whole thing is then just a struct with a root tree that we could position wherever we want.

include/ssd-internal.h
Consolatis: changed to `ssd->state`

Consolatis:
If we decide to unconditionally remove borders and unround corners on maximize this seems ready.

I have a follow-up branch that refactors everything button related to `src/ssd/ssd_button.c` with no functional changes intended:
- #1047

johanmalm:
LGTM

src/ssd/ssd_border.c
```diff
@@ -51,7 +57,20 @@ ssd_border_create(struct ssd *ssd)
...
 {
+	assert(ssd);
+	assert(ssd->border.tree);
+
 	struct view *view = ssd->view;
+	if (view->maximized) {
+		wlr_scene_node_set_enabled(&ssd->border.tree->node, false);
+		ssd->margin = ssd_thickness(ssd->view);
+		return;
+	}
```
Consolatis, line 68:
```c
if (view->maximized && ssd->border.tree->node.enabled) {
	..
} else if (view->maximized) {
	return;
} else if(!ssd->border.tree->node.enabled) {
	..
}
```

include/ssd-internal.h
```diff
@@ -41,6 +42,7 @@ struct ssd {
 	 * don't update things we don't have to.
 	 */
 	struct {
+		bool is_unround;
```
Consolatis, line 45: `bool squared_corners`

src/ssd/ssd_titlebar.c
```diff
@@ -90,15 +90,48 @@ is_direct_child(struct wlr_scene_node *node, struct ssd_sub_tree *subtree)
 	return node->parent == subtree->tree;
 }
 
+static void
+enable_rounded_corners(struct ssd *ssd, bool enable)
```
Consolatis, line 94: `set_squared_corners()`

src/ssd/ssd_titlebar.c
```diff
@@ -90,15 +90,48 @@ is_direct_child(struct wlr_scene_node *node, struct ssd_sub_tree *subtree)
...
+		for (size_t i = 0; i < sizeof(ssd_type) / sizeof(ssd_type[0]); i++) {
+			part = ssd_get_part(&subtree->parts, ssd_type[i]);
+			struct ssd_button *button = node_ssd_button_from_node(part->node);
+
+			/* Toggle background between invisible and titlebar background color */
+			struct wlr_scene_rect *rect = lab_wlr_scene_get_rect(button->background);
+			wlr_scene_rect_set_color(rect, enable ? (float[4]) {0, 0, 0, 0} : (
+				subtree == &ssd->titlebar.active
+				? rc.theme->window_active_title_bg_color
+				: rc.theme->window_inactive_title_bg_color));
```
Consolatis, line 110: indent
```c
			wlr_scene_rect_set_color(rect, enable ? (float[4]) {0, 0, 0, 0} : (
				subtree == &ssd->titlebar.active
					? rc.theme->window_active_title_bg_color
					: rc.theme->window_inactive_title_bg_color));
```

Consolatis:
Left a few review comments from malm via IRC on here so I don't forget.

PR #1048, opened 2023-08-25
src/ssd: fix tiling via keybind when maximized

Consolatis:
The previous PR introduced an issue with tiling based actions like SnapToEdge and SnapToRegion using outdated SSD margin values when called via keybind while maximized. That resulted in wrong offsets for the tiled windows.

This commit restores the functionality by forcing a re-calculation of the SSD margin when changing the maximized state.

Thanks to @Flrian for reporting the issue via IRC.


Consolatis:
Merging. Thanks again @Flrian for you invaluable feedback and bug reporting.

Issue #1053, opened 2023-08-27
Stuck dragging window with CSD feat. waybar


DynamoFox:
How to reproduce:
 - Set "layer": "top" in Waybar configuration file and ensure the panel is located at the top of the screen
 - Launch waybar with that configuration
 - Pick a window with client-side decorations (tested: Firefox, Thunerbird, adwaita-1-demo)
 - Drag it to the top of the screen as if you were to maximize it
 - Observe the window did not maximize and is still being dragged even if the left mouse button has been released

This doesn't occur with windows using SSD.

Note that In Firefox&Thunderbird specific cases left clicking the decoration to get rid of the dragging will not work.
Left clicking on adwaita-1-demo will instead stop the drag.

Reproduced as well by Consolatis on IRC.

Consolatis:
Reproduced with wl_example_panel.py in "top" layer + exclusive zone + Firefox with CSD.

DynamoFox:
I can confirm that moving the first condition beneath the other two in the snippet above is enough to solve the issue.

Now the Firefox window correctly maximizes and isn't dragged by the pointer anymore.

Consolatis:
Pushed the PR:
- #1056

Feel free to give it a test ride, in my local tests it fixes the issue and I didn't spot any side effects.

DynamoFox:
I've have rebuilt labwc with that PR and I can confirm that it's working as intended without manifesting any other annoyance so far. Thank you!

PR #1056, opened 2023-08-29
src/cursor: ensure interactive move/resize ends correctly for CSD clients

Consolatis:
Before this patch, when moving a CSD client below a layershell surface - like a panel that was configured with either the "top" or "overlay" layers - we'd only send a matching release button event to the client but not actually end the interactive move operation. That caused the drag to continue even though the user already released the mouse button.

In comparison, SSD clients were not suffering from the same issue because the initial mouse "down" event was not attached to any client surface and thus it would not take the first early return because there was no surface attached to the release event.

This patch fixes the issue by reordering the conditions where we return early. It also ensures that when we finish the move, we still send the release event to CSD clients.

Fixes:
- #1053

Reported-by: @DynamoFox (thanks)

### DIFF
--- a/src/cursor.c
+++ b/src/cursor.c
@@ -300,8 +300,7 @@ update_pressed_surface(struct seat *seat, struct cursor_context *ctx)
 	 * context menus (in contrast) do not use an XDG popup grab and
 	 * do not work properly if we send leave/enter events.
 	 */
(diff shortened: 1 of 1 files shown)


johanmalm:
Nice fix. LGTM. 
Have tested. Can't see any side-effects.

Issue #1064, opened 2023-09-01
Add libinput Tap drag lock feature


burein-ita:
Using a49e12e11238771ab9b7195cfa5f4ce03ec10742
Arch-Linux

Right now is not possible to configure libinput `Tap drag lock` feature from rc.xml.
This makes using labwc with a touch-pad mildly annoying.

According to the [libinput docs:](https://wayland.freedesktop.org/libinput/doc/latest/tapping.html#tap-and-drag)
>When enabled, libinput >will ignore a finger up event during a drag process, provided the finger is set down again within >a implementation-specific timeout.

Consolatis:
I didn't look into the specific config option but I start to wonder if we should try to implement "all the settings" for libinput rather than adding new ones bit by bit when requested.

tokyo4j:
I was working on this from before, so I made a PR for reference.

droc12345:
Touchpad settings you might want look at

disable while typing    
natural scroll  
scroll factor   
middle button emulation 
tap button map  
clickfinger behavior    
tap-to-click    
drag lock       
tap-and-drag
along with acell profile and sensitivity

alaricljs:
https://wayland.freedesktop.org/libinput/doc/latest/scrolling.html

- scroll_method
- scroll_button

Please

PR #1066, opened 2023-09-02
config: support <libinput><device category=""><tapAndDrag> & <dragLock>

tokyo4j:
Addresses #1064 with tap-and-drag support.

This PR also fixes a problem that enums initialized with -1 are bigger than 0 because gcc uses unsigned int as default backing type of enum.

This is my first time sending a PR to OSS, so excuse me if I made silly mistakes.

### DIFF
--- a/include/config/libinput.h
+++ b/include/config/libinput.h
@@ -21,6 +21,8 @@ struct libinput_category {
 	int left_handed;
 	enum libinput_config_tap_state tap;
 	enum libinput_config_tap_button_map tap_button_map;
+	enum libinput_config_drag_state tap_and_drag;
+	enum libinput_config_drag_lock_state drag_lock;
 	enum libinput_config_accel_profile accel_profile;
 	enum libinput_config_middle_emulation_state middle_emu;
 	enum libinput_config_dwt_state dwt;
--- a/include/config/libinput.h
+++ b/include/config/libinput.h
@@ -21,11 +21,11 @@ struct libinput_category {
 	int left_handed;
 	enum libinput_config_tap_state tap;
 	enum libinput_config_tap_button_map tap_button_map;
-	enum libinput_config_drag_state tap_and_drag;
-	enum libinput_config_drag_lock_state drag_lock;
-	enum libinput_config_accel_profile accel_profile;
-	enum libinput_config_middle_emulation_state middle_emu;
-	enum libinput_config_dwt_state dwt;
+	int tap_and_drag; /* -1 or libinput_config_drag_state */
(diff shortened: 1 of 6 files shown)


Consolatis:
This is looking great. Thanks for working on this. :)

> This PR also fixes a problem that enums initialized with -1 are bigger than 0 because gcc uses unsigned int as default backing type of enum.

Wow, not sure how we managed to miss this.

LGTM but I'll leave it to @johanmalm to give the PR a final look through and merge.

include/config/libinput.h
```diff
@@ -21,6 +21,8 @@ struct libinput_category {
 	int left_handed;
 	enum libinput_config_tap_state tap;
 	enum libinput_config_tap_button_map tap_button_map;
+	enum libinput_config_drag_state tap_and_drag;
```
johanmalm, line 24: The less-than-zero check with an enum is a great catch.

I suggest that we change these enums to 'int' instead though, rather than doing an `(int)` cast later.

src/seat.c
```diff
@@ -95,8 +95,26 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 			dc->tap_button_map);
 	}
 
+	if (!libinput_device_config_tap_get_enabled(libinput_dev)
```
johanmalm, line 98: Consider adding a finger-count check.

I had a look at the sway code. They only apply `tap_drag` and `drag_lock` if `finger-count > 0` and 

https://github.com/swaywm/sway/blob/89f85312687b7fd1777c936e169e6400cee0a4b4/sway/input/libinput.c#L50-L70

johanmalm (commented):
Great contribution. Thank you. 

I've added two comments for consideration, but am happy to merge without addressing these.
I think we ought to change `enum` to `int` in the header file, but can do that in a separate PR.

src/seat.c
tokyo4j: That makes sence.

In my previous code, configuring `tap-and-drag` required `tap` to be enabled.
And [`libinput_device_config_tap_get_enabled` checked finger count internally](https://gitlab.freedesktop.org/libinput/libinput/-/blob/99c4721f95ccf50bc8153b53938cb1f3160ad450/src/libinput.c#L3894).
Same goes for `drag lock`, which required `tap-and-drag` to be enabled.

But sway's approach looks simpler and it doesn't make any practical difference.

include/config/libinput.h
tokyo4j: I agree. I fixed my code.

PR #1067, opened 2023-09-03
window-rules: add ignoreFocusRequest property

Consolatis:
This PR allows to use window rules to block self-activation requests from xwayland and wayland native windows.
It is different from preventing focus completely, e.g. manually selecting the window via `A-Tab` or similar still allows interaction with the window.

Example config:
```xml
<windowRules>
	<windowRule identifier="*firefox*" ignoreFocusRequest="yes" />
</windowRule>
```

I have not tested this PR at all as I'd need to set up some testing environment which supports the xdg_activation protocol. It would take much longer than creating this patch in the first place, so testing would be much appreciated.

Outstanding questions / issues:
- [ ] unmanaged xwayland surfaces are not handled as it might break context menus, I am not sure about this.
  Support for them could be added pretty easily though because we already fetch the topmost mapped view in `unmanaged_handle_request_activate()` due to `->pid` checks.
- [x] I didn't check if Openbox supports something similar which we can implement instead of this PR

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -482,6 +482,9 @@ situation.
 	*skipWindowSwitcher* removes window from the Window Switcher (alt-tab
 	on-screen-display)
 
+*<windowRules><windowRule ignoreFocusRequest="">* [yes|no|default]
+	*ignoreFocusRequest* prevent window to activate itself.
+
 ## ENVIRONMENT VARIABLES
 
 *XCURSOR_THEME* and *XCURSOR_SIZE* are supported to set cursor theme
--- a/include/window-rules.h
+++ b/include/window-rules.h
@@ -29,6 +29,7 @@ struct window_rule {
 	enum property server_decoration;
 	enum property skip_taskbar;
 	enum property skip_window_switcher;
+	enum property ignore_focus_request;
 
 	struct wl_list link; /* struct rcxml.window_rules */
 };
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -150,6 +150,8 @@ fill_window_rule(char *nodename, char *content)
 		set_property(content, &current_window_rule->skip_taskbar);
 	} else if (!strcasecmp(nodename, "skipWindowSwitcher")) {
 		set_property(content, &current_window_rule->skip_window_switcher);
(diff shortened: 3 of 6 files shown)


vain:
I gave it a quick shot and it works as expected (in my Firefox scenario from #1065). 👍

I'll keep using that branch for a while.

stefonarch:
I can confirm as for firefox.
Testing with `<windowRule identifier="lxqt-notificationd" ignoreFocusRequest="yes" />` it doesn't work, so focus is still stolen.

Consolatis:
@vain @stefonarch thanks for testing :)

> Testing with `<windowRule identifier="lxqt-notificationd" ignoreFocusRequest="yes" />` it doesn't work, so focus is still stolen.

That sounds like a different thing, I assume the notification daemon creates a new surface which always gets focus.
This PR only prevents self-activation, e.g. it doesn't touch getting focus on new surface / A-Tab selection / mouse click on the surface / panel activation via foreign toplevel protocol.

To make focus-on-new-surface behavior user controllable via window rules we need another property. That one might be more difficult to implement in a generic way though as another use-case for such a prevent-all-focus property is also that mouse clicks / A-Tab / foreign toplevel protocol  shouldn't focus the surface. This then also interacts with the `followMouse` setting thus it involves much more thought to get it right and not cause further issues.

stefonarch:
> The alternative to that whole thing is using layer-shell for client applications like notification daemons / background system monitors (e.g. conky) as the layer-shell protocol includes a keyboard_interactivity setting which labwc respects.

in fact, I've a wip branch with layershell-qt somehow buggy but it does it's job.

vain:
FWIW, I’ve been running this for the past week and haven’t seen any new bugs. 👍

johanmalm:
Code looks good.

Consolatis:
> I didn't check if Openbox supports something similar which we can implement instead of this PR

I've looked into the Openbox docs and the only thing they have regarding focus behavior for their window rules is
```
    <focus>yes</focus>
    # if the window should try be given focus when it appears. if this is set
    # to yes it doesn't guarantee the window will be given focus. some
    # restrictions may apply, but Openbox will try to
```

That seems to be the generic no-focus variant which this PR does not implement.

Issue #1069, opened 2023-09-04
keybind to code


kyak:
It would be great to have keybinds support keycodes.

Currently, I would have to create two keybinds for two keyboard layouts. This is because keybind like e.g. `W-l` will not work when another keyboard layout is active - I would have to create the second keybind like `W-д`.

This used to work in openbox, i.e. I only had to specify keybind for `W-l`. It also works for sway by using `--to-code` with `bindsym`.

Can we support it in labwc?

Consolatis:
We should support that already, we default to checking the translated keysym but fall back to raw keysyms if the translated one didn't match. Thus a keybind like `W-1` should actually work in a somewhat recent version.

Consolatis:
Right.. for the fallback we only look at keys and ignore their modifiers (thus `W-S-2` should work on all layouts). But for the letters the modifier don't play a role. Hrm.

Consolatis:
Implemented by
- #1072

Feel free to point out any issues that may arise.

Issue #446, opened 2022-07-13
Binding by keycode instead of keysym


alandiwix:
Is there any way to bind actions to keycode instead of keysym? If not, it would be really nice to have.

Consolatis:
Related:
- #351
- #365
- #370

This is not exactly my area of expertise so take the following with a grain of salt.

Especially #365 might provide a workaround (Using e.g. `0x21`).
We should likely clean up those issues a bit and figure out a way to go forward, either by documenting the feature or by providing a proper way to create keybinds like `S-!`.

Arnaudv6:
Title should be 'Binding by keycode as an alternative to keysym'.
Stating the obvious, but _most_people®_ do the opposite and want bindings to adapt with layouts.

Consolatis:
Should be fixed by
- #997

Note: the workaround keybinds will have preference over raw ones, e.g. `W-S-0x21` would get used if defined and `W-S-1` would not be used at all. Thus suggesting to replace all workaround keybinds with proper ones. Feel free to open a new issue if there are any further issues with the keybinds.

@Arnaudv6
> Title should be 'Binding by keycode as an alternative to keysym'.
> Stating the obvious, but _most_people®_ do the opposite and want bindings to adapt with layouts.

This should still work using the `W-S-0x21` / `W-S-exclam` variant instead of `W-S-1` for those specific keys which include the shift modifier. Labwc still prefers the layout ones and only falls back to raw ones if no layout based binding had been found.

PR #1072, opened 2023-09-06
[wip] Keybinds: keyboard layout agnostic keybinds

Consolatis:
This is an initial implementation of keyboard layout agnostic keybinds.
Background:
- #1069
- various other issues in the past

The gist of the PR is:
- we add keycodes (of physical keys) to the keybinds
- we prefer keycodes over keysyms
- if no keycode matches, we fall back to keysyms, e.g. exactly the same as what we are doing without the PR

To decide on our way forward with this, these are the goals the implementation should solve from a user perspective IMO:
- in a single layout situation nothing should change at all
- in a multi layout situation, by default keybinds should work independent of the currently active layout
- the ordering of the layout config should not matter, e.g. `us,de` vs `de,us`
- we should allow binding keys that are not available in the English layout like `W-ö` or `W-я`
- the user should still be able to optionally define specific keybinds that only work if that keysym is available in the currently active layout

The initial implementation should mostly [1] solve all but the last goal. However, the details are important here:

A single physical key usually means two different things in the two layouts (examples compared to the "us" layout: in the German layout `ö` is mapped to `;`, in the Russian layout `я` is mapped to `z`). To account for that we get the actual keycode for the physical key of a keybind for all configured layouts when initializing. Problems start to arise when a key is available in both layouts but the keycode differs (e.g. it is on a different physical key). An example for that is `z` and `y` in the German layout which are interchanged. If we now get the keycodes for `z` in a "us,de" situation it results in both the physical keys for `z` and `y` being stored for `<keybind key="W-z" />` (one variant for each layout). Thus both physical keys are now suddenly bound to the action.

We could allow users to work around that issue by adding a further optional flag for keybinds which causes us to simply not store any keycodes. That would then make a `<keybind key="W-z" sym_only="yes" />` work only for the physical key that aligns with the currently active layout (which is the current situation in master). So the physical key to trigger the keybind would change along with the layout.

Another option would be to search for an "English" layout and only search for keycodes in that layout. But that would prevent the possibility to use keys that are not available in the "us" layout like `W-ö` or `W-я` if the "us" layout is currently active. They would still work via the keysym fallback in the other layout though.

So I am not sure how to proceed here, I prefer the optional `sym_only` (name itself up for discussion as usual) argument for keybinds given the available options. As that one allows keys like `W-ö` to work independently of the layout but allows manually marking keys like `W-z` as `sym_only`. But maybe there is another option that I missed? Or are there other goals I missed or should the goals be different?

[1]
There is still an issue that keypad keys only work via the fallback keysym path as we don't go over all the possible modifiers (which includes `NumLock`) when trying to get the keycodes for a given keysym. This should be fine though because the keycode list would then just be empty and thus we fall back to the keysym one.

Outstanding work items:
- depending on further discussion outcome: optional `sym_only` keybind argument
- doc update
- silence debug logging

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -86,7 +86,7 @@ keyboard_modifiers_notify(struct wl_listener *listener, void *data)
 }
 
 static bool
-handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym)
+handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym, xkb_keycode_t code)
 {
 	struct keybind *keybind;
 	wl_list_for_each(keybind, &rc.keybinds, link) {
@@ -98,11 +98,23 @@ handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym)
 				&& !actions_contain_toggle_keybinds(&keybind->actions)) {
 			continue;
 		}
-		for (size_t i = 0; i < keybind->keysyms_len; i++) {
-			if (xkb_keysym_to_lower(sym) == keybind->keysyms[i]) {
-				key_state_store_pressed_keys_as_bound();
-				actions_run(NULL, server, &keybind->actions, 0);
-				return true;
+		if (sym == XKB_KEY_NoSymbol) {
+			/* Use keycodes */
+			for (size_t i = 0; i < keybind->keycodes_len; i++) {
+				if (keybind->keycodes[i] == code) {
+					key_state_store_pressed_keys_as_bound();
+					actions_run(NULL, server, &keybind->actions, 0);
+					return true;
+				}
+			}
+		} else {
+			/* Use syms */
+			for (size_t i = 0; i < keybind->keysyms_len; i++) {
+				if (xkb_keysym_to_lower(sym) == keybind->keysyms[i]) {
+					key_state_store_pressed_keys_as_bound();
+					actions_run(NULL, server, &keybind->actions, 0);
+					return true;
+				}
 			}
 		}
 	}
@@ -303,14 +315,29 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 
 	/* Handle compositor key bindings */
 	if (event->state == WL_KEYBOARD_KEY_STATE_PRESSED) {
-		for (int i = 0; i < translated.nr_syms; i++) {
-			handled |= handle_keybinding(server, modifiers, translated.syms[i]);
-		}
+
+		/* First try keycodes */
+		handled |= handle_keybinding(server, modifiers, XKB_KEY_NoSymbol, keycode);
 		if (handled) {
+			wlr_log(WLR_ERROR, "keycodes matched");
 			goto out;
 		}
+
+		/* Then fall back to keysyms */
+		for (int i = 0; i < translated.nr_syms; i++) {
+			handled |= handle_keybinding(server, modifiers, translated.syms[i], keycode);
+		}
+		if (handled) {
+			wlr_log(WLR_ERROR, "translated keysyms matched");
+			goto out;
+		}
+
+		/* And finally test for keysyms without modifier */
 		for (int i = 0; i < raw.nr_syms; i++) {
-			handled |= handle_keybinding(server, modifiers, raw.syms[i]);
+			handled |= handle_keybinding(server, modifiers, raw.syms[i], keycode);
+		}
+		if (handled) {
+			wlr_log(WLR_ERROR, "raw keysyms matched");
 		}
 	}
 
@@ -413,6 +440,8 @@ keyboard_init(struct seat *seat)
 	}
 	xkb_context_unref(context);
 	wlr_keyboard_set_repeat_info(kb, rc.repeat_rate, rc.repeat_delay);
+
+	keybind_update_keycodes(seat->server);
 }
 
 void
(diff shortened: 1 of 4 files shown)


kyak:
> That would then make a <keybind key="W-z" sym_only="yes" /> work only for the physical key that aligns with the currently active layout

From the user's perspective, the `sym_only="yes"` will only work if they specify their keyboard layout to be the default one, i.e. `de,us`. This is a hidden knowledge that somehow needs to be communicated to the user.

Moreover, the user might have three keyboard layouts like that: `de,us,ru`. The `sym_only` doesn't let them specify the keyboard layout to search for this particular keybind - it will always be searched in the default layout. Renaming `sym_only` to `layout` and letting user specify the layout name (like `layout="de"`) should work, though.

> An example for that is z and y in the German layout which are interchanged. If we now get the keycodes for z in a "us,de" situation it results in both the physical keys for z and y being stored for <keybind key="W-z" /> (one variant for each layout). Thus both physical keys are now suddenly bound to the action.

What if only the first keycode will be stored? This should perfectly match user's expectations without additional `sym_only`. For example, the user has configured `de,us` layout and `W-z` now kicks in for the physical key `z` located to the right of the `t` key. At the same time, after the user switches to the `us` layout, this keybinding is still on the same physical key - without the user having to think about it.

Consolatis:
> > That would then make a  work only for the physical key that aligns with the currently active layout
> 
> From the user's perspective, the `sym_only="yes"` will only work if they specify their keyboard layout to be the default one, i.e. `de,us`. This is a hidden knowledge that somehow needs to be communicated to the user.

No, it would work slightly different: with `sym_only="yes"` and the keybind being `W-z` and a `de,us` layout its either the physical key with `z` on it (in the `de` layout with a german keyboard) or the key with `y` on it (in the `us` layout with a german keyboard).
The situation is different for layouts which do not share keys though, for example `us,ru`. In that case the same keybind would only work in the `us` layout because the `ru` doesn't have a `z` (to my knowledge).
`sym_only` basically falls back to the same mechanism, we currently have in the master branch.

> Moreover, the user might have three keyboard layouts like that: `de,us,ru`. The `sym_only` doesn't let them specify the keyboard layout to search for this particular keybind - it will always be searched in the default layout. Renaming `sym_only` to `layout` and letting user specify the layout name (like `layout="de"`) should work, though.

With `sym_only` It would only be searched in the currently active layout rather than in the default one.

> What if only the first keycode will be stored? This should perfectly match user's expectations without additional sym_only. For example, the user has configured de,us layout and W-z now kicks in for the physical key z located to the right of the t key. At the same time, after the user switches to the us layout, this keybinding is still on the same physical key - without the user having to think about it.

That could actually work, I'll think about this variant a bit more.

Consolatis:
Changed to only store the first set of keycodes for a matching layout. That should solve the `z` vs `y` testcase on `de,us` without requiring user configuration. Thanks @kyak for the suggestion.

E.g. a config like
```xml
<keybind key="W-z">
	<action name="ToggleMaximize" />
</keybind>
<keybind key="W-y">
	<action name="ToggleDecorations />
</keybind>
```
and `XKB_DEFAULT_LAYOUT=de,us` will now maximize on the physical `z` key and toggle decorations with the physical `y` key independed of the currently active layout (on a German keyboard). With `XKB_DEFAULT_LAYOUT=us,de` the physical key `z` toggles decorations and `y` toggles maximize.

We still search in other layouts for keycodes if the previous layouts didn't have them defined at all.

src/keyboard.c
```diff
@@ -303,14 +315,29 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 
 	/* Handle compositor key bindings */
```
johanmalm, line 316: Suggest expanding the comment here to explain why. Maybe something like this:

```
	/*
	 * Handle compositor keybinds
	 * 
	 * When matching against keybinds, we process the input keys in the
	 * following order of precedence:
	 *   a. Keycodes (of physical keys)
	 *   b. Translated keysyms (taking into account modifiers, so if Shift+1
	 *      were pressed on a us keyboard, the keysym would be !)
	 *   c. Raw keysyms (ignoring modifiers such as shift, so in the above
	 *      example the keysym would just be 1)
	 *
	 * The reasons for this approach are:
	 *   1. To make keybinds keyboard-layout agnostic (by checking keycodes
	 *      before keysyms). This means that in a multi-layout situation,
	 *      keybinds work regardless of which layout is active at the time
	 *      of the key-press.
	 *   2. To support keybinds relating to keysyms that are only available
	 *      in a particular layout, for example å, ä and ö
	 *   3. To differentiate between keybinds `S-ö` and `S-Ö` (by checking
	 *      translated before raw)
	 *   4. To support keybinds such as `S-1` (by checking raw keysyms)
	 */
```

Consolatis: Slightly modified and updated.

Consolatis:
Found the request for layout dependent keybinds: https://github.com/labwc/labwc/issues/446#issuecomment-1268119020

This seems a minority of users though, so I still think (but more opinions welcome) having the keybinds be layout agnostic by default and requiring to make them layout dependent via optional argument per keybind is the better variant for an initial `just-works-and-does-what-i-want` impression.

The argument name `sym_only` kind of sucks though, it requires knowledge of the technical implementation to even make any kind of sense to a casual user. `<keybind key="W-z" layout_dependent="yes">` might be an alternative. But I am completely open to other suggestions that makes sense for a casual user.

Consolatis:
Used `layoutDependent` for now and updated docs.
This PR seems ready if @johanmalm agrees on the name of the optional keybind argument, completely fine to change it to something else.

johanmalm:
Happy with `layoutDependent`

LGTM

Issue #1076, opened 2023-09-07
Remember and restore keyboard layout per window


kyak:
Users that have multiple keyboard layouts often prefer that the current keyboard layout is remembered and restored per window. `labwc` only allows setting keyboard layout globally for all windows.

The scenario is usually the following. The user types something in a window (say, chat) using one keyboard layout, then switches to another window (say, browser or IDE) and changes keyboard layout there. Then the user switches back to his chat window and continues typing in whatever keyboard layout was active in that window.

In X, there are a bunch of daemon-like programs (i've used many of them throughout the years) that allow for this functionality.

In wayland, it seems like compositor's job. Please refer to https://github.com/swaywm/sway/issues/2361 for some discussion.

Please consider implementing this functionality in `labwc`.

stefonarch:
I agree,  kind of must-have for many, it's present as a [plugin](https://github.com/AlexJakeGreen/wayfire-kbdd-plugin) in wayfire too.

johanmalm:
Happy to merge.
Does anyone fancy drafting a patch?

Consolatis:
Working on it with the help of @kyak via IRC.

Consolatis:
- #1077

Testing appreciated, especially if LED status indicators likes `grp_led:scroll` still works as expected when switching focus to a window with a different layout. I've only tested it nested with a hacked wlroots that allows toggling the layout via the wayland backend.

xxhh1112:
qq

PR #1077, opened 2023-09-09
keyboard: allow applying keyboard layout per window

Consolatis:
Fixes
- #1076

Windows remember their last configured keyboard layout and restore that layout when getting focus.
New windows will use the default layout. Can be tested via something like this in `~/.config/labwc/environment` and pressing the "menu" key:
```
XKB_DEFAULT_LAYOUT=de,us
XKB_DEFAULT_OPTIONS=grp:menu_toggle,grp_led:scroll
```
Also requires enabling the feature in the first place:
```xml
<keyboard layoutScope="window">
..
</keyboard>
```

Missing:
- [x] some setting to turn this on / off
  - [ ] decide on the setting category and name, maybe Openbox has something similar?
  - [x] decide on the default setting, default-off sounds good
  - [x] docs
- [x] directly include xkb header in `view.h` and `labwc.h` due to the type alias
- [x] add example to commit message

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -283,6 +283,10 @@ windows using the mouse.
 	When recognizing a new keyboard enable or disable Num Lock.
 	Default is on.
 
+*<keyboard layoutScope="">* [global|window]
+	Stores the keyboard layout either globally or per window and restores
+	it when switching back to the window. Default is global.
+
 *<keyboard><keybind key="" layoutDependent="">*
 	Define a *key* binding in the format *modifier-key*, where supported
 	modifiers are:
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -142,6 +142,7 @@
   -->
   <keyboard>
     <numlock>on</numlock>
+    <layoutScope>global</layoutScope>
     <repeatRate>25</repeatRate>
     <repeatDelay>600</repeatDelay>
     <keybind key="A-Tab">
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -63,6 +63,7 @@ struct rcxml {
 	int repeat_rate;
 	int repeat_delay;
 	bool kb_numlock_enable;
+	bool kb_layout_per_window;
 	struct wl_list keybinds;   /* struct keybind.link */
 
 	/* mouse */
--- a/include/input/keyboard.h
+++ b/include/input/keyboard.h
@@ -3,6 +3,7 @@
 #define LABWC_KEYBOARD_H
 
 #include <stdbool.h>
+#include <xkbcommon/xkbcommon.h>
 
 struct seat;
 struct keyboard;
@@ -13,6 +14,7 @@ void keyboard_finish(struct seat *seat);
 
 void keyboard_setup_handlers(struct keyboard *keyboard);
 void keyboard_set_numlock(struct wlr_keyboard *keyboard);
+void keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout);
 void keyboard_cancel_keybind_repeat(struct keyboard *keyboard);
 bool keyboard_any_modifiers_pressed(struct wlr_keyboard *keyboard);
 
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -39,7 +39,6 @@
 #include <wlr/types/wlr_virtual_pointer_v1.h>
 #include <wlr/types/wlr_virtual_keyboard_v1.h>
 #include <wlr/util/log.h>
-#include <xkbcommon/xkbcommon.h>
 #include "config/keybind.h"
 #include "config/rcxml.h"
 #include "input/cursor.h"
--- a/include/view.h
+++ b/include/view.h
@@ -7,6 +7,7 @@
 #include <stdint.h>
 #include <wayland-util.h>
 #include <wlr/util/box.h>
+#include <xkbcommon/xkbcommon.h>
 
 #define LAB_MIN_VIEW_WIDTH  100
 #define LAB_MIN_VIEW_HEIGHT  60
@@ -143,6 +144,7 @@ struct view {
 	bool fullscreen;
 	enum view_edge tiled;
 	bool inhibits_keybinds;
+	xkb_layout_index_t keyboard_layout;
 
 	/* Pointer to an output owned struct region, may be NULL */
 	struct region *tiled_region;
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -735,6 +735,12 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.repeat_delay = atoi(content);
 	} else if (!strcasecmp(nodename, "numlock.keyboard")) {
 		set_bool(content, &rc.kb_numlock_enable);
+	} else if (!strcasecmp(nodename, "layoutScope.keyboard")) {
+		/*
+		 * This can be changed to an enum later on
+		 * if we decide to also support "application".
+		 */
+		rc.kb_layout_per_window = !strcasecmp(content, "window");
 	} else if (!strcasecmp(nodename, "screenEdgeStrength.resistance")) {
 		rc.screen_edge_strength = atoi(content);
 	} else if (!strcasecmp(nodename, "range.snapping")) {
@@ -953,6 +959,7 @@ rcxml_init(void)
 	rc.repeat_rate = 25;
 	rc.repeat_delay = 600;
 	rc.kb_numlock_enable = true;
+	rc.kb_layout_per_window = false;
 	rc.screen_edge_strength = 20;
 
 	rc.snap_edge_range = 1;
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -84,6 +84,7 @@ keyboard_modifiers_notify(struct wl_listener *listener, void *data)
 			}
 		}
 	}
+	wlr_seat_set_keyboard(seat->seat, wlr_keyboard);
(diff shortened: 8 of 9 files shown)


stefonarch:
A first test it looks working fine including the led indicators and it's really "per window" while it could be also "per application" as setting which was the one  I used. I don't think anybody will turn this off, I never did and can't see a reason to do so. Default=off is normal.

Thanks!

screen_area_sab_16:21:00_

Consolatis:
> I don't think anybody will turn this off, I never did and can't see a reason to do so. Default=off is normal.

Personally I never use multi layout so feedback and expectations from users who actually use it regularly is pretty much required. I can't quite parse your statement here: by default, should the layout be remembered per window or be a global state (like it is in current master)?

> while it could be also "per application" as setting which was the one I used

We don't really have an "application" abstraction. Per window is easier to implement but if that is really something that is expected by multi layout users I could look into it and maybe we can have a config that reflects the 3 states "Global" / "Window" and "Application" which then defaults to one of the 3 states if not set.

stefonarch:
> I can't quite parse your statement here: by default, should the layout be remembered per window or be a global state (like it is in current master)?

Default=global (or _off_ as you put it).

I was used to the "per application" setting  but as I use it mostly  in messengers (where basically window=application)  and when I open a  compose window  in Thunderbird it isn't a big deal if there is only a "per window" setting.

Probably I asked already, but there is still no way to have a command to switch layout and no way to expose current layout other than with the LED workaround?

Consolatis:
> there is still no way to have a command to switch layout?

Actually I implemented one for testing and was thinking about a new action like `<action name="SetKeyboardLayout" index="0" />`. That could turn out pretty useful for window rules. Not sure if Openbox supports something similar, if yes we should try to use the same action name.

> no way to expose current layout other than with the LED workaround?

Well, all wayland applications get the current layout via their keyboard modifiers when they are newly focused or when a modifier is pressed on a keyboard. But there is currently no protocol that I am aware of that would allow a passive listener without keyboard focus to get the current layout. I didn't look at how it works for xwayland applications though, maybe one could use xprop to read some X11 property somewhere.

stefonarch:
That's sounds interesting as it could also be read and translated to the active layout even with more than 2, a pity it doesn't work with "per window" as I can confirm.

Consolatis:
Added the config option, its an argument to the main keyboard tag for now:
```xml
<keyboard layoutPerWindow="yes">
..
</keyboard>
```

Default is the global layout.

kyak:
> Edit: Actually we could add another (ugly) workaround to our portfolio, at least when using "global":
> 
> ```
> <keybind key="F8">
>   <action name="SetKeyboardLayout" index="0" />
>   <action name="Execute" command="sh -c 'echo us &gt; $XDG_RUNTIME_DIR/labwc_keyboard_layout'" />
> </keybind>
> <keybind key="F9">
>   <action name="SetKeyboardLayout" index="1" />
>   <action name="Execute" command="sh -c 'echo it &gt; $XDG_RUNTIME_DIR/labwc_keyboard_layout'" />
> </keybind>
> ```

As crazy as it seems, let me expand on your idea. This is out of topic for this PR, but still.

How about dumping current keyboard layout to some file in `/tmp` (this will not cause filesystem write) exactly where you call `keyboard_update_layout`?

This way, it will be trivial to write some "client" that will read from this file (or make it a fifo).

Sounds like rudimental IPC which is out of scope of `labwc`, but I'm just throwing an idea.

kyak:
> Added the config option, its an argument to the main keyboard tag for now:
> 
> ```
> <keyboard layoutPerWindow="yes">
>   ..
> </keyboard>
> ```
> 
> or
> 
> ```
> <keyboard>
>   <layoutPerWindow>yes</layoutPerWindow>
>   ..
> </keyboard>
> ```
> 
> Default is `no`, e.g. a single global layout.

If we decide to implement "per-application" some day, maybe it is better to call this option `layoutScope` with default being `global`, configurable to `window` and with possibility to further implement `application`?

johanmalm:
Looks good.

I don't use this either, so don't really have a view on workflow.

With this PR in place I think it would be fairly simple to extend with `app_id` or `identifier` window rules if anyone needs that.

Regarding storing the layout in files, I do have some reservations around that. Whilst it might seem a simple and you can use `inotify_add_watch()` from `#include "sys/inotify.h"` etc, it does have a few pitfalls and should have filelocks and so on in place. Also, it is a poor form of IPC and if we really want to go that way, a protocol or unix-socket solution is more elegant.

I am open to doing thing here though as displaying current keyboard layout in panels is one of the few things (workspaces being the other) we cannot do in the current eco-system and we have to strike a balance between doing what users need vs promoting commonality in the Wayland landscape. My preference would be to develop a protocol which we use in the short term, but have a view to upstream in the medium/long term.

Super quick example:

`keyboard-control-v1.xml`

```
<protocol name="keyboard_control">
  <copyright \>
  <interface name="keyboard_control_v1" version="1">
    <description summary="control keyboard">
      Control interface for keyboard.
    </description>
    <request name="get_keyboard">
      <arg name="id" type="new_id" interface="wl_keyboard" />
    </request>
  </interface>
</protocol>
```

Consolatis:
> If we decide to implement "per-application" some day, maybe it is better to call this option `layoutScope` with default being `global`, configurable to `window` and with possibility to further implement `application`?

Good point, internally we can still use a boolean up to that point but internal changes to turn it into an enum are always fine as long as the user-facing config keeps being compatible. Will change to `layoutScope="global"`.

stefonarch:
Nice, thanks again!

johanmalm:
I'm happy with `<keyboard layoutScope=""> [global|window]`.

It seems to work well, except a glitch as described below.

With the following settings:

```
XKB_DEFAULT_LAYOUT=gb,se
XKB_DEFAULT_OPTIONS=grp:shift_caps_toggle,grp_led:scroll
```

```
  <keyboard>
    <default/>
    <layoutScope>window</layoutScope>
  </keyboard>
```

1. Open `sakura` and `xfce4-terminal`
2. In `sakura` press shift-caps to change to Swedish layout. Type åäö
3. Alt-tab to `xfce4-terminal`. Press the same physical keys. See [';
4. Alt-tab back to `sakura`. Press the same physical keys. Expect åäö but see [';

20230912_20h43m33s_grim

johanmalm:
PS ran from tty when testing

stefonarch:
I can reproduce a similar thing with qterminal, 
```
XKB_DEFAULT_LAYOUT=it,ch
XKB_DEFAULT_OPTIONS=grp:shift_caps_toggle,grp_led:scroll
```

A identical behavior can be seen with 2 compose windows in thunderbird using `alt-tab`. But if using waybar's taskbar buttons to switch the windows the issue cannot be reproduced, same with tiled windows.

Consolatis:
Just to ensure its not related to caps: does this also happen with another toggle key like `alt_shift_toggle`?

johanmalm:
Have tried with alt_shift_toggle
Same result

As an aside - I normally do not use alt_shift_toggle because it messes up shift-alt-tab 😃

kyak:
I don't get it. It doesn't work for specific apps (qterminal, xfce-terminal, sakura etc)?

I'm not able to reproduce it with the other apps.

stefonarch:
window+next_layout gets mixed up  only if switching with alt-tab. I can also reproduce it with this firefox window and another (text editor, feedreader search bar or whatever), here it's not app dependent.

kyak:
I'm using single key (caps) to switch layout. I thought this is the reason I can't reproduce this behavior. I've changed to alt-shift and layout is still correctly preserved for window when using alt-tab to switch windows.

Still not clear what's so special about your setup and how to reproduce this problem.

stefonarch:
Just tested now also on the main PC with debian: same results.

kyak:
> Just tested now also on the main PC with debian: same results.

It doesn't work at all when using alt-tab, or it doesn't work sometimes?

Consolatis:
Short on time currently, I'll try to look into `Alt-Tab` behavior soon (tm).

stefonarch:
The first switch with alt-tab works, then not anymore here.
But if changing again the layout it works sometimes straight on with only 2 windows (thunderbird and tb "compose"). With 3 windows it's more messy, it looks like a not consistent behavior.

kyak:
What can we do to push this forward?

Maybe rebase on latest master @Consolatis  and then we can give it another try?

There were changes regarding key combos recently. Maybe it will work better this time?

Consolatis:
rebased

johanmalm:
> rebased

Well, well, well... the bug is gone.

I can't reproduce the stuff I described in https://github.com/labwc/labwc/pull/1077#issuecomment-1716324181 anymore.

Excellent!

kyak:
Works here as well.

stefonarch:
Confirming that the appswitcher can be used now without messing up layouts.

Consolatis:
Well, I am not sure what fixed it. Which is usually not a good sign. However, I don't really see an issue in this PR itself.
So up to @johanmalm or @jlindgren90 to decide if we want to merge this.

jlindgren90:
I don't use multiple keyboard layouts and so I don't have much insight to offer, but no objections from my end.

johanmalm:
> Well, I am not sure what fixed it. Which is usually not a good sign. However, I don't really see an issue in this PR itself. So up to @johanmalm or @jlindgren90 to decide if we want to merge this.

Looks good. Let's go for it.

I suspect the increasing correctness brought by recent keyboard commits has helped. For example, only passing non-modifier keycodes to `wlr_seat_keyboard_notify_enter()` in `seat_focus()` could have been the cause of the problem.

johanmalm:
And we're still just within the merge window :smile:

Issue #1075, opened 2023-09-07
mouse section in labw-config manpage incomplete


stefonarch:
I see in the manpage:
```

           Supported mouse buttons are:
           •   Left
           •   Middle
           •   Right


           Supported mouse actions include:
           •   Press: Pressing the specified button down in the context.
           •   Release: Releasing the specified button in the context.
           •   Click: Pressing and then releasing inside of the the context.
           •   DoubleClick: Two presses within the doubleClickTime.
           •   Drag: Pressing the button within the context, then moving the cursor.
           •   Scroll: Scrolling in specified direction in the context.
```
but in rc.xml.all there are also modifiers allowed:

```
<mousebind button="A-Left" action="Drag">
...
 <mousebind button="W-Right" action="Drag">
        <action name="Resize"/>
      </mousebind>
```

Consolatis:
Nice find.

The sample XML snippet in the header of that section also misses the `action` argument: https://labwc.github.io/labwc-config.5.html#mouse

Patches welcome

PR #1080, opened 2023-09-10
docs: clarify mouse binds

stefonarch:
Fixes https://github.com/labwc/labwc/issues/1075
Not sure if also scroll action can be combined, didn't test.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -316,7 +316,7 @@ Therefore, where multiple objects of the same kind are required (for example
 *<mouse><scrollFactor>*
 	Set scroll factor. Default is 1.0.
 
-*<mouse><context name=""><mousebind button="" direction=""><action>*
+*<mouse><context name=""><mousebind button="" direction="" action=""><action>*
 	Multiple *<mousebind>* can exist within one *<context>*; and multiple
 	*<action>* can exist within one *<mousebind>*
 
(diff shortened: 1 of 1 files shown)


docs/labwc-config.5.scd
```diff
@@ -344,7 +344,7 @@ Therefore, where multiple objects of the same kind are required (for example
 	- Desktop: The desktop background, where no windows are present.
 	- Root: A synonym for Desktop (for compatibility).
 
-	Supported mouse *buttons* are:
+	Supported mouse *buttons* can be combined with modifiers and are:
```
Consolatis, line 347: Should we say `keyboard modifiers` here to make it clear what 'modifiers' means in this context?

Consolatis:
Thanks. While looking through the diff I also recognized in the section below which list the supported actions, the `action` word is not enclosed in `*` like `*button*` and `*direction*` above. Would you mind changing that as well?

Short comment about the commit title: Maybe consider something like `docs: clarify mouse binds`.

docs/labwc-config.5.scd
johanmalm: Considering mentioning modifier-keys in a separate paragraph for simplicity and clarity.

For example:

```
Mouse buttons can be combined with modifier-keys (shift, super, control and alt), for example:

    <mousebind button="A-Right" action="Press">

```

johanmalm:
Ready for merge?

Consolatis:
> Ready for merge?

I think your statement about the new section isn't yet addressed, we might also want to rename and squash the commits.

Consolatis:
Squashed the commits and change the commit title.

Thanks for the PR, merging once the CI ran through :)

Issue #32, opened 2021-05-24
<command> expand tilde and variables


ekickx:
Os: Fedora 34.20210423.n.0 (Silverblue)
labwc version: 0.2.0

It seems labwc can't execute bash script. Here is my [rc.xml](https://gist.github.com/ekickx/e8af98dd346087b6a4a8ca5a1f2a8fa3#file-rc-xml) and my [script](https://gist.github.com/ekickx/e8af98dd346087b6a4a8ca5a1f2a8fa3#file-volume). My script works well when I run it from terminal but somehow it can't run when I put in `<command>`

johanmalm:
We use execvp()
This is consistent with openbox's approach and is safer than shell invocations.
In order to be consistent with openbox, we should expand tilde and variables for execvp() though. To be done :smile:

ekickx:
Ohh ok. Thanks for the answer

johanmalm:
I think this does it. Shout if it doesn't.

ekickx:
It's work fine. Thank you

PR #1085, opened 2023-09-13
actions: Prevent env var expansion for subshells like `$()`

Consolatis:
We currently expand Execute action commands like
`sh -c 'echo $(date) >> /tmp/foo'` to
`sh -c 'echo (date) >> /tmp/foo'`

This obviously breaks the subshell. This patch fixes it by preventing single `$` being replaced with nothing.

The more general question is if we actually want to run the env var expansion for Execute command arguments in the first place: It allows for things like `nautilus $HOME` or `nautilus ~` but causes real issues for `sh -c` constructs by replacing all $variables showing up there with (potentially empty) env vars.

An example would be `sh -c 'foo=test; echo $foo'`

### DIFF
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -33,6 +33,13 @@ buf_expand_shell_variables(struct buf *s)
 				++p;
 			}
 			*p = '\0';
+
+			/* Don't expand single $ (as used by subshells) */
+			if (!strlen(environment_variable.buf)) {
+				buf_add(&new, "$");
+				continue;
+			}
+
 			i += strlen(environment_variable.buf);
 			strip_curly_braces(environment_variable.buf);
 			p = getenv(environment_variable.buf);


src/common/buf.c
```diff
@@ -33,6 +33,13 @@ buf_expand_shell_variables(struct buf *s)
 				++p;
 			}
 			*p = '\0';
+
+			/* Don't expand single $ (as used by subshells) */
+			if (!strlen(environment_variable.buf)) {
```
johanmalm, line 38: Could just do !foo.len here

johanmalm:
Fine in principle, just one minor comment + we ought to update the documentation to describe this.

We could ignore $ variable expansion when within single quotes (like shell does).

This function really warrants a unit test. A bit scary to touch it!

johanmalm:
Just had a look a documentation for the environment file. All we have is this: https://github.com/labwc/labwc/blob/143714f1c9c2821ca70ef26949395557b15e2171/docs/labwc-config.5.scd?plain=1#L30

So could sort that as separate patch. Ought to describe what we expand.

Consolatis:
No issue about the environment parsing. And that one doesn't have any single quotes or subshells.
I am not sure if we should parse the `command` argument for action at all though, see the example in the top post.

src/common/buf.c
Consolatis: will change

johanmalm:
I think we're saying the same.

Expand:

$foo
${foo}
"$foo"

Ignore:

$(foo)
'$foo'
'anything'

Consolatis:
What about something like `<action name="Execute"><command>sh -c "foo=test; echo $foo"</command></action>` then?

I also fail to see the actual benefit of action command expansions, I'd expect application like nautilus to deal with things like `~` themselves. And in other cases it seems even harmful like in the `sh -c` case.

johanmalm:
Sorry @Consolatis I didn't properly take it what you wrote (multiple times). It's been a long week!

I kept think "environment" file and had forgotten that we do this for the execute-actions too. Agree we should definitely include your patch to avoid the issue with $(foo)

Removing env var expansion completely feels a bigger step though cos of backward + openbox compatibilty.

https://github.com/labwc/labwc/commit/afe666fd6ed3232aee0bd8bf5fa062b980190ee8

https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/openbox/actions/execute.c#L135

Consolatis:
> [action: expand shell variables before execvp()](https://github.com/labwc/labwc/commit/afe666fd6ed3232aee0bd8bf5fa062b980190ee8)
Expanding shell variables, including tilde, enables the following type
of keybind: <keyboard> <keybind key="XF86AudioMute"> <action name="Execute"> <command>bash ~/mute-script.sh</command> </action> </keybind>
</keyboard> Fixes issue https://github.com/labwc/labwc/issues/32

> Removing env var expansion completely for execute-action feels a bigger step though because of backward + openbox compatibilty.

True, you very much have a point here. Thanks for looking up the original commit and the reasoning for it. It seems neither dash nor bash can handle `~` by itself if given as argument. Pluma also failed to handle it. So I am not sure how to proceed in general with this. I don't think the sub-shell issue that this PR is about would cause any further issues but its just one part of the story.

Do we need to actually replace `$random_var` or `${random_var}` for the `command` argument, or just `~`?
If its just `~` we could have two functions (or an additional bool argument for the existing one): one for `environment` and the other for the `command` argument.

src/common/buf.c
Consolatis: Hm.. are you sure about that?
It seems we are adding the remaining string to `environment_variable` and then just throw a `\0` in there.
Thus I think `.len` is actually out of sync with `strlen()` here.

johanmalm: Whoops. Yes, sorry. Ignore my comment.

johanmalm:
Might this be an easier to understand patch?

```
diff --git a/src/common/buf.c b/src/common/buf.c
index 740489d..386a68f 100644
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -15,6 +15,12 @@ strip_curly_braces(char *s)
 	s[len] = 0;
 }
 
+static bool
+isvalid(char p)
+{
+	return isalnum(p) || p == '_' || p == '{' || p == '}';
+}
+
 void
 buf_expand_shell_variables(struct buf *s)
 {
@@ -24,12 +30,12 @@ buf_expand_shell_variables(struct buf *s)
 	buf_init(&environment_variable);
 
 	for (int i = 0 ; i < s->len ; i++) {
-		if (s->buf[i] == '$') {
+		if (s->buf[i] == '$' && isvalid(s->buf[i+1])) {
 			/* expand environment variable */
 			environment_variable.len = 0;
 			buf_add(&environment_variable, s->buf + i + 1);
 			char *p = environment_variable.buf;
-			while (isalnum(*p) || *p == '_' || *p == '{' || *p == '}') {
+			while (isvalid(*p)) {
 				++p;
 			}
 			*p = '\0';
```

Consolatis:
Agree, your variant seems cleaner.
Mind to throw it into a PR?

johanmalm:
> Do we need to actually replace `$random_var` or `${random_var}` for the `command` argument, or just `~`? If its just `~` we could have two functions (or an additional bool argument for the existing one): one for `environment` and the other for the `command` argument.
> 
> Edit: Just had a look into the linked openbox code: that one only seems to replace `$pid`, `$wid` and `$pointer`.

Good point.

Agree, let's just resolve tilde for the command argument. I suggest we justify this [minor] regression with something like: "...to increase safety and make it easier to write `sh -c ''` constructs without turning labwc into a shell parser in order to expand environment variables, whilst respecting single quotes, escaped characters and not invoke subshells for `$(foo)`.

johanmalm:
> Agree, your variant seems cleaner. Mind to throw it into a PR?

See #1095

Consolatis:
Replaced by
- #1095

PR #1086, opened 2023-09-16
Add ARRAY_SIZE() macro

johanmalm:


include/common/array-size.h
```diff
@@ -0,0 +1,28 @@
...
+#define BUILD_BUG_ON_ZERO(e) (sizeof(struct { int: -!!(e); }))
+
+/* &a[0] degrades to a pointer: a different type from an array */
+#define __must_be_array(a) BUILD_BUG_ON_ZERO(__same_type((a), &(a)[0]))
+
+/**
+ * ARRAY_SIZE - get the number of elements in array @arr
+ * @arr: array to be sized
+ */
+#define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]) + __must_be_array(arr))
```
Consolatis, line 26: This is a bit much undocumented dark magic for my taste:

So `__builtin_types_compatible_p()` appears to return `1` if the types are the same: https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005ftypes_005fcompatible_005fp

I have absolutely no idea what happens in `BUILD_BUG_ON_ZERO`. Couldn't we just do an `static_assert(__same_type((a), &(a)[0]) == 1, "Tried to use ARRAY_SIZE on a non-array")` or something?

Finally `ARRAY_SIZE()`  itself, I don't understand why it adds the result of `__must_be_array()`, if the naming of the macros (e.g. "build-bug on zero") is correct, shouldn't it always return something `> 0` here and thus return the wrong array size?

I assume somehow those macros make sense and if that is the case I think it's fine using them but we should at least document the way they work properly. Otherwise stupid people like me end up reading the code and just thinking.. "no clue what I've just read, guess its fine".

Consolatis (commented):
No issue in principle with the addition of `ARRAY_SIZE()`.
However, personally I'd prefer not having to be a magician to understand what is going on there :)

include/common/array-size.h
heroin-moose: BUILD_BUG_ON_ZERO is basically static_assert().

heroin-moose:
```
#define ARRAY_SIZE(a) { \
	static_assert(__builtin_types_compatible_p(typeof(a), typeof(&(a)[0])) == 0); \
	(sizeof(a) / sizeof(a[0])); \
}
```

heroin-moose: I'm really sorry for OCD in the comments, I know you receive an email for every change I make >_<

Consolatis: > ```
> #define same_type(a, b) (__builtin_types_compatible_p(typeof(a), typeof(b)) == 0)
> #define ARRAY_SIZE(a) ({ static_assert(same_type(a, &(a)[0])); sizeof(a) / sizeof(a[0]); })
> ```

`same_type()` in your example is either misnamed  or wrong as it returns the types being different rather than being same / compatible.

I'd also use the 2nd variant of `static_assert(expr, message)` and thus throw in a friendly error message that shows up during compile if the `static_assert()` fails.

But other than that, this two line variant indeed looks way easier to understand to me.

Consolatis: Btw, I am not sure if we actually need the extra check here, at least on my setup we already get an error (with `--werror`) without the check:
```
../src/main.c:82:34: error: division ‘sizeof (char *) / sizeof (char)’ does not compute the number of array elements [-Werror=sizeof-pointer-div]
   82 | #define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))
      |                                  ^
../src/main.c:90:12: note: in expansion of macro ‘ARRAY_SIZE’
   90 |  int len = ARRAY_SIZE(bar);
      |            ^~~~~~~~~~
../src/main.c:89:8: note: first ‘sizeof’ operand was declared here
   89 |  char *bar = malloc(100)
```

heroin-moose: Nope, it returns zero on a match.

Consolatis:
Rebase required but otherwise LGTM.

include/common/array-size.h
heroin-moose:
```
#define same_type(a, b) (__builtin_types_compatible_p(typeof(a), typeof(b)) == 0)
````
The name is off of course. `static_assert()` should fail exactly when `typeof(a) == typeof(&a[0])`. I should really start to write code in the night.

**Comment by heroin-moose** [comment]
GCC 8 supports -Werror=sizeof-pointer-div. Old enough.
CLANG 7 does not support -Werror=sizeof-pointer-div. Not old enough?

**Review by heroin-moose — CHANGES_REQUESTED** [review]
Really wrong name for == 0. I did a better version:

```
#define same_type(a, b) (__builtin_types_compatible_p(typeof(a), typeof(b)) == 1)
#define ARRAY_SIZE(a) ({ static_assert(!same_type(a, &(a)[0])); sizeof(a) / sizeof(a[0]); })
```

**Comment by johanmalm** [comment]
> GCC 8 supports -Werror=sizeof-pointer-div. Old enough. CLANG 7 does not support -Werror=sizeof-pointer-div. Not old enough?

Excellent. I'm on clang 16 and gcc 13, so seems pretty safe :smile:

**Comment by johanmalm** [comment]
Have updated the comments. Will leave it open for a day or so before merge in case anyone has any further thoughts.
Enjoyed this PR. Good learning :smile:

File: include/common/array-size.h [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+#ifndef LABWC_ARRAY_SIZE_H
+#define LABWC_ARRAY_SIZE_H
+
+/**
+ * ARRAY_SIZE() - Get the number of elements in array.
+ * @arr: array to be sized
+ *
+ * This does not work on pointers.
+ *
+ * Recent versions of GCC and clang support -Werror=sizeof-pointer-div
```
heroin-moose, line 11: We still need to add this Werror to meson.

heroin-moose: Noice.

PR #1088, opened 2023-09-20
Added note for localization support to menu.xml

stefonarch:
Maybe mention it in the man page too?


Consolatis:
Thanks.

> Maybe mention it in the man page too?

Sure, why not.

johanmalm:
@stefonarch thanks.

Just looked at the man page and it doesn't actually mention the identifiers "root-menu" and "client-menu". Whoops!

Issue #1083, opened 2023-09-12
Reconfigure with border-only-window snapped to left or right edge causes gap on top


Consolatis:
Seems like we reserve space for the SSD in those cases even though we only render the borders. `ToggleDecorations` 3 times fixes it.
Reported by @heroin-moose via IRC.

Visone-Selektah:
Also with windows snapped to region, same issue
Also toggle maximize fix it

PR #1093, opened 2023-09-21
src/ssd: use view->ssd_titlebar_hidden for ssd_thickness calculations

Consolatis:
Before this patch we were using the internal .enabled flag of the titlebar tree node. This failed due to ssd_thickness() not having view->ssd assigned when initially called. Instead of assigning view->ssd within ssd_create() we just always use the view boolean flag directly. This fixes an issue where a border-only view that has been snapped to an edge or region would have a gap in the size of the titlebar on top after a Reconfigure.

Fixes
- #1083


johanmalm:
Excellent.

Issue #1082, opened 2023-09-12
labwc crashes when switching between xwayland and wayland applications using Alt-Tab


kyak:
labwc crashes when switching between xwayland and wayland applications using Alt-Tab. This probably has to do with tooltips and popups shown in the xwayland app while switching.

Below are a couple of backtraces.

bt1

```
(gdb) bt
#0  0x00007faca748e83c in  () at /usr/lib/libc.so.6
#1  0x00007faca743e668 in raise () at /usr/lib/libc.so.6
#2  0x00007faca74264b8 in abort () at /usr/lib/libc.so.6
#3  0x00007faca74263dc in  () at /usr/lib/libc.so.6
#4  0x00007faca7436d26 in  () at /usr/lib/libc.so.6
#5  0x000055e0d7771a76 in xwayland_surface_from_view (view=<optimized out>) at ../labwc/src/xwayland.c:89
#6  0x000055e0d7773390 in xwayland_surface_from_view (view=<optimized out>) at ../labwc/src/xwayland.c:371
#7  xwayland_view_get_string_prop (view=<optimized out>, prop=<optimized out>) at ../labwc/src/xwayland.c:359
#8  0x000055e0d777310c in view_get_string_prop (prop=0x55e0d7785788 "app_id", view=0x55e0d98339f0) at ../labwc/src/view.c:1206
#9  other_instances_exist (title=0x55e0d9977570 "Telegram", id=0x55e0d91faae0 "org.telegram.desktop", self=0x55e0d97adab0) at ../labwc/src/window-rules.c:29
#10 view_matches_criteria (rule=rule@entry=0x55e0d8839fe0, view=view@entry=0x55e0d97adab0) at ../labwc/src/window-rules.c:51
#11 0x000055e0d77752db in window_rules_get_property (view=0x55e0d97adab0, property=0x55e0d7785737 "skipWindowSwitcher") at ../labwc/src/window-rules.c:114
#12 0x000055e0d776a7bc in matches_criteria (criteria=(LAB_VIEW_CRITERIA_CURRENT_WORKSPACE | LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP | LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER), view=0x55e0d97adab0)
    at ../labwc/src/view.c:45
[...]

PR #1094, opened 2023-09-21
Ensure xdg and xwayland string_prop() handlers deal with destroying views

Consolatis:
When a view is destroyed (including override_redirect in the xwayland case), the view_destroy() handler is called which checks for a currently open A-Tab window switcher and causes an update there to remove the destroying view from the list. Before view_destroy() is called, both xwayland and xdg handlers reset the xdg_surface / xwayland_surface.

The window switcher update then creates a list of all windows which do not have the 'skipWindowSwitcher' window rule property set. If there is at least one 'matchOnce' window rule configured, this also tries to get string properties of the destroying view which already had their xdg_surface / xwayland_surface reset and thus run into an assert.

This patch fixes that so that the string_prop() handlers always return an empty string in those cases rather than running into the assert.

For a more in-depth analyses and alternative solutions see the linked issue.

Fixes
- #1082

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -439,7 +439,19 @@ position_xdg_toplevel_view(struct view *view)
 static const char *
 xdg_toplevel_view_get_string_prop(struct view *view, const char *prop)
 {
-	struct wlr_xdg_toplevel *xdg_toplevel = xdg_toplevel_from_view(view);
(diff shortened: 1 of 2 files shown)


johanmalm:
Very nice. Thanks.

PR #1095, opened 2023-09-21
buf: do not expand $() in buf_expand_shell_variables()

johanmalm:


Consolatis:
LGTM, tested and seems to work fine.

src/common/buf.c
```diff
@@ -15,6 +16,39 @@ strip_curly_braces(char *s)
...
+	buf_init(&new);
+	for (int i = 0 ; i < s->len ; i++) {
+		if (s->buf[i] == '~') {
+			buf_add(&new, getenv("HOME"));
+		} else {
+			buf_add_one_char(&new, s->buf[i]);
+		}
+	}
+	free(s->buf);
+	s->buf = new.buf;
```
Consolatis, line 43: Should we also copy over new.len` here?

Consolatis:
> This increases safety

I don't really see this increasing safety.

src/common/buf.c
johanmalm: Great catch. Done.

johanmalm:
> > This increases safety
> 
> I don't really see this increasing safety.

Commit message amended to remove the safety bit.

src/common/buf.c
johanmalm: Wrote a unit test to build confidence around this.
Realized that we need to do the same (copy `new.len` + `new.alloc`) for `buf_expand_shell_variables()`.
It's obvious in hindsight and I don't know why I didn't see it. Amazing what a unit test can do :smile:

Consolatis:
LGTM, not tested.

PR #1096, opened 2023-09-22
view: account for base size in resize indicator

jlindgren90:
For views with a non-pixel size increment (e.g. X11 terminals), it's helpful to subtract the base size of the window (typically including menu bar, scrollbars, etc.) before computing the number of size increments (e.g. cells/characters). This way, the displayed size will exactly match the terminal grid (e.g. 80x25 or whatever).

`wlr_box` isn't really the best fit for size hints, so let's define a `struct view_size_hints` and a nice `view_get_size_hints()` function, wrapping `view->impl->get_size_hints()`.

This also seems like a great opportunity to make `view_adjust_size()` window-system-agnostic and eliminate `xwayland_apply_size_hints()`.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -37,6 +37,17 @@ enum view_edge {
 };
 
 struct view;
+
+/* Basic size hints (subset of XSizeHints from X11) */
+struct view_size_hints {
+	int min_width;
+	int min_height;
+	int width_inc;
+	int height_inc;
+	int base_width;
+	int base_height;
+};
+
 struct view_impl {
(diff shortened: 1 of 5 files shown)


Consolatis:
LGTM but didn't test.
Suggest delaying this PR to after the 0.6.5 release though.

jlindgren90:
Cheers to 0.6.5!
Okay to go ahead and merge this now?

Consolatis:
Yep, no objections.

johanmalm:
Thanks :smile:

PR #1099, opened 2023-09-23
Add unit test for common/buf.c

johanmalm:
For discussion post `0.6.5` bump.

For example:

1. Do we want to include tests in this repo?
2. What type of test suite? 
3. Are we happy with `tap.c`?
4. Are we happy with `Makefile`?
5. Are we happy we `prove`?

Interestingly, when I added ASAN it also highlighted that we should use `memmove()` instead of `memcpy()`. Whoops.

```
diff --git a/src/common/buf.c b/src/common/buf.c
index 6311def..84ae957 100644
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -41,7 +41,7 @@ strip_curly_braces(char *s)
 		return;
 	}
 	len -= 2;
-	memcpy(s, s + 1, len);
+	memmove(s, s + 1, len);
 	s[len] = 0;
 }
 ```

### CODE REVIEW

**Comment by jlindgren90** [comment]
No objections here. I imagine that many parts of labwc will be difficult to test in an automated fashion, but that shouldn't stop us from adding tests where we can.

I didn't review everything line-by-line, but at a quick glance it all looks good to me, nice job.

**Comment by johanmalm** [comment]
v2. Just tidied up some coding style.

**Comment by Consolatis** [comment]
> All tests successful.

So I assume that means that there aren't any more bugs in labwc. Yay :)

In all seriousness.. I don't really see a test framework to drastically improve things (and actually I also see a potential for sinking time into writing tests for mostly dubious benefits). But I am also not against adding it. Code itself seems to make sense.

I am a bit curious how it states which tests are wrong though, I've changed
- `snprintf(expect, sizeof(expect), "foo %s/bar", getenv("HOME"));` to
- `snprintf(expect, sizeof(expect), "afoo %s/bar", getenv("HOME"));`

and the only message I got was:
```
make: Entering directory '/home/user/dev/labwc/t'
cc -g -Wall -O0 -I../include -o t1000-buf.t t1000-buf.c tap.c ../src/common/buf.c ../src/common/mem.c -fsanitize=address
make: Leaving directory '/home/user/dev/labwc/t'
t/t1000-buf.t .. Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/10 subtests

Test Summary Report
-------------------
t/t1000-buf.t (Wstat: 512 Tests: 10 Failed: 2)
  Failed tests:  3-4
  Non-zero exit status: 2
Files=1, Tests=10,  4 wallclock secs ( 0.08 usr  0.02 sys +  4.50 cusr  0.01 csys =  4.61 CPU)
Result: FAIL
```
which doesn't seem all that helpful.

**Comment by johanmalm** [comment]
If a test fails, just run it directly ./t1000-but.c to see what triggered the failure

We could simply just create a separate repo to keep it clean and reduce commit noice.

**Comment by heroin-moose** [comment]
Actually, it makes more sense to me to do functional testing first. Complex actions, different config layouts, maybe some kind of moving windows back and forth, that sort of things.

**Comment by johanmalm** [comment]
> Actually, it makes more sense to me to do functional testing first. Complex actions, different config layouts, maybe some kind of moving windows back and forth, that sort of things.

Appreciate thoughts on how to achieve automatic testing beyond unit-tests. Maybe with pixman backed? Will try to get on IRC later… might be easier to discuss there.

**Comment by johanmalm** [comment]
Have added a couple of paragraphs in `t/README.md`:

https://github.com/labwc/labwc/pull/1099/files#diff-ea1cf576d6af5497c99c07e86980795b200126085e5bbb0a3e30cfc95a9d512f

**Comment by Consolatis** [comment]
> If a test fails, just run it directly ./t1000-but.c to see what triggered the failure

Can we somehow make that work automatically? Either via C code or some shell script?
That log output in case of a failed test doesn't really help at all IMHO.

**Comment by johanmalm** [comment]
Will re-think this. Might use cmocka instead.

**Comment by johanmalm** [comment]
Ready for review.
Not yet integrated with CI.
It'd be nice to get this over the line + also potentially useful for the YAML patch. Cc @tokyo4j

File: t/README.md [hunk 1]
```diff
@@ -0,0 +1,22 @@
+# UNIT TESTS
+
+## Introduction
```
Consolatis, line 3: I think this fits better in `CONTRIBUTING.md`.

johanmalm: Done

Consolatis:
Tested and works nice, thanks :)

PR #1100, opened 2023-09-23
common/buf.c: do not use memcpy for overlapping regions

johanmalm:
Post `0.6.5`


Consolatis:
I'd say lets get this in for 0.6.5.

Didn't test but LGTM and totally makes sense.

PR #1102, opened 2023-09-23
Update CONTRIBUTING.md API + Upversion sections

johanmalm:


Consolatis:
LGTM

Consolatis:
Maybe we could also add something like `Mention new dependencies in changelog`?

johanmalm:
> Maybe we could also add something like `Mention new dependencies in changelog`?

Done

PR #1103, opened 2023-09-23
view/xwayland: avoid focusing views that don't want focus

jlindgren90:
XWayland views can self-declare that they don't want keyboard focus via
the ICCCM WM_HINTS property. Most of the logic is already in place to
avoid giving focus to such views (e.g. taskbars).

Add a couple of missing pieces to make this work:

- Hook up `view_isfocusable()` to look at WM_HINTS for XWayland views
- Adjust `desktop_focus_topmost_mapped_view()` to skip unfocusable views

I also noticed that in several places, we will still focus and
raise a view regardless of `view_isfocusable()`. To fix this, make
`desktop_focus_and_activate_view()` return false if the view isn't
focusable, and don't raise the view to the front in that case.
    
Possible FIXME: a couple of places still call `workspaces_switch_to()`
before `desktop_focus_and_activate_view()`. Maybe switching workspaces
should also be prevented if the view isn't focusable?

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -30,6 +30,16 @@ xwayland_view_get_size_hints(struct view *view)
 	};
 }
 
+static bool
+xwayland_view_wants_focus(struct view *view)
+{
+	xcb_icccm_wm_hints_t *hints = xwayland_surface_from_view(view)->hints;
+	if (!hints) {
+		return true;
+	}
+	return (bool)hints->input;
+}
+
 static struct wlr_xwayland_surface *
 top_parent_of(struct view *view)
 {
@@ -505,7 +515,7 @@ xwayland_view_unmap(struct view *view, bool client_request)
 	view->mapped = false;
 	wl_list_remove(&view->commit.link);
 	wlr_scene_node_set_enabled(&view->scene_tree->node, false);
-	desktop_focus_topmost_mapped_view(view->server);
+	desktop_focus_topmost_view(view->server);
 
 	/*
 	 * If the view was explicitly unmapped by the client (rather
@@ -634,6 +644,7 @@ static const struct view_impl xwayland_view_impl = {
 	.get_root = xwayland_view_get_root,
 	.append_children = xwayland_view_append_children,
 	.get_size_hints = xwayland_view_get_size_hints,
+	.wants_focus = xwayland_view_wants_focus,
 };
 
 void
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -37,6 +37,20 @@ xwayland_view_wants_focus(struct view *view)
 	if (!hints) {
 		return true;
 	}
+	/*
+	 * Paraphrased from ICCCM section 4.1.7 (Input Focus):
+	 *
+	 * Clients set the input field of WM_HINTS to True to indicate
+	 * that they require window manager assistance in acquiring the
+	 * input focus. Clients set the input field to False to request
+	 * that the window manager not set the input focus to their
+	 * top-level window.
+	 *
+	 * Clients that use XSetInputFocus() to explicitly set the input
+	 * focus should set the WM_TAKE_FOCUS atom in WM_PROTOCOLS.
+	 * Currently, labwc does not support this method of taking focus
+	 * and thus makes no use of WM_TAKE_FOCUS.
+	 */
 	return (bool)hints->input;
 }
 
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -378,7 +378,7 @@ void foreign_toplevel_update_outputs(struct view *view);
 void desktop_focus_view(struct view *view, bool raise);
 void desktop_arrange_all_views(struct server *server);
 void desktop_focus_output(struct output *output);
-struct view *desktop_topmost_mapped_view(struct server *server);
+struct view *desktop_topmost_focusable_view(struct server *server);
 
 enum lab_cycle_dir {
 	LAB_CYCLE_DIR_NONE,
@@ -393,7 +393,7 @@ enum lab_cycle_dir {
  */
 struct view *desktop_cycle_view(struct server *server, struct view *start_view,
 	enum lab_cycle_dir dir);
-void desktop_focus_topmost_mapped_view(struct server *server);
+void desktop_focus_topmost_view(struct server *server);
 
 void keyboard_cancel_keybind_repeat(struct keyboard *keyboard);
 void keyboard_key_notify(struct wl_listener *listener, void *data);
--- a/include/view.h
+++ b/include/view.h
@@ -69,6 +69,8 @@ struct view_impl {
 	struct view *(*get_root)(struct view *self);
 	void (*append_children)(struct view *self, struct wl_array *children);
 	struct view_size_hints (*get_size_hints)(struct view *self);
+	/* if not implemented, view is assumed to want focus */
+	bool (*wants_focus)(struct view *self);
 };
 
 struct view {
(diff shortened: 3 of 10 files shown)


Consolatis:
Didn't test but code and idea looks good to me.
On a related note: there were also requests to add a window rule property which prevents keyboard focus completely, this PR looks like it would make it pretty easy to implement.

> Possible FIXME: a couple of places still call `workspaces_switch_to()`
before `desktop_focus_and_activate_view()`. Maybe switching workspaces
should also be prevented if the view isn't focusable?

Yes, absolutely.
Maybe we could refactor that a bit to just do the workspace check and switch in `desktop_focus_and_activate_view()` directly rather than having multiple places which call `desktop_focus_and_activate_view()` do it themselves.

jlindgren90:
> Yes, absolutely. Maybe we could refactor that a bit to just do the workspace check and switch in `desktop_focus_and_activate_view()` directly rather than having multiple places which call `desktop_focus_and_activate_view()` do it themselves.

Should `desktop_focus_and_activate_view()` always switch to the workspace of the view being activated? Right now it's inconsistent, e.g. `xdg_activation_handle_request()` does switch but xwayland's `handle_request_activate()` doesn't.

I never got used to using more than one workspace myself, so I'm not sure what's the expected behavior here.

Consolatis:
I can't imagine a situation where one would want to have keyboard focus on a window that is on a different workspace and thus not visible. Activation via foreign toplevel protocol (e.g. click on a panel) also switches workspaces manually currently.

jlindgren90:
Makes sense. In some cases (like un-minimizing) I could imagine wanting the activated window moved to the current workspace, rather than switching workspaces. But labwc doesn't have that option currently, so probably a moot point.

jlindgren90:
Marking this as draft for now. There's another focus issue that ought to be fixed first (pull request coming soon).

jlindgren90:
v2:
- Rebase on top of #1103
- Rename `desktop_focus_and_activate_view()` to simply `desktop_focus_view()`.
- Move switching workspaces and (optionally) raising the view inside `desktop_focus_view()`, so these steps are also skipped for unfocusable views.
- Move the "unfocus if no focusable view" logic from `desktop_focus_view()` to `desktop_focus_topmost_mapped_view()` (now `desktop_focus_topmost_view()`) as we don't want clicking on an unfocusable view (like a taskbar) to unfocus the current view, but do want it unfocused if we switch to an empty workspace.

jlindgren90:
v3: Rebased on top of #1109

jlindgren90:
Should be ready for review now.

Consolatis:
Code LGTM, I'd like to give it a test ride tonight though to ensure we don't regress on workspace related stuff.

src/xwayland.c
```diff
@@ -30,6 +30,16 @@ xwayland_view_get_size_hints(struct view *view)
...
 }
 
+static bool
+xwayland_view_wants_focus(struct view *view)
+{
+	xcb_icccm_wm_hints_t *hints = xwayland_surface_from_view(view)->hints;
+	if (!hints) {
+		return true;
+	}
+	return (bool)hints->input;
```
johanmalm, line 40: How confident are we on this logic (`hints->input`)?
Having a quick read here (https://tronche.com/gui/x/icccm/sec-4.html) makes my head spin.

johanmalm:
This was a new one one me (xwayland views being able to not want keyboard focus)
Is there a particular use-case for this where the surface isn't an override-redirect (OR) surface?

For reference (anyone else reading/reviewing) - this is how wlroots deals with OR surfaces.
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c?ref_type=heads#L2284

include/workspaces.h
```diff
@@ -2,6 +2,9 @@
 #ifndef LABWC_WORKSPACES_H
 #define LABWC_WORKSPACES_H
 
+#include <stdbool.h>
+#include <wayland-util.h>
```
johanmalm, line 6: Why `wayland-util.h` here? 
The `struct wl_list` is already forward declared.

johanmalm:
That second commit is hard to follow (for me reading this late in the evening anyway :smile:)
Is it worth splitting it out into smaller commits doing fewer things?

src/xwayland.c
jlindgren90: Speaking for myself, no more or less confident than anything else X11... it's all trial-and-error.
This is how `xprop` describes that flag, for what it's worth:
```
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
```

jlindgren90: There is some more complexity with a WM_TAKE_FOCUS atom that I don't fully understand either; however, I think we are using the "input" flag (which is just a Boolean, though 32 bits) correctly so far as it goes:

> Passive and Locally Active clients set the input field of WM_HINTS to True, which indicates that they require window manager assistance in acquiring the input focus. No Input and Globally Active clients set the input field to False, which requests that the window manager not set the input focus to their top-level window.

johanmalm: Happy to try it :smile:
Most of the code I've written for labwc has been trial-and-error tbh. Certain things are clear in protocols, etc but there is a lot of things relating to window management that aren't exactly well documented nor obvious.

jlindgren90:
> This was a new one one me (xwayland views being able to not want keyboard focus) Is there a particular use-case for this where the surface isn't an override-redirect (OR) surface?

Some X11 panels, for example `lxqt-panel` or my fork of it.
Oddly, `xfce4-panel` sets `WM_HINTS.input = true` but still does not get keyboard focus under Openbox (not sure why).

include/workspaces.h
jlindgren90: Forward declaration is not enough for the `struct wl_list link;` declaration below; the size of the struct needs to be known.

src/xwayland.c
johanmalm: > There is some more complexity with a WM_TAKE_FOCUS atom that I don't fully understand either; however, I think we are using the "input" flag (which is just a Boolean, though 32 bits) correctly so far as it goes:
> 
> > Passive and Locally Active clients set the input field of WM_HINTS to True, which indicates that they require window manager assistance in acquiring the input focus. No Input and Globally Active clients set the input field to False, which requests that the window manager not set the input focus to their top-level window.

Maybe worth adding the above (ish) as a comment in the code ?? to help us/others to maintain.

include/workspaces.h
jlindgren90: We presumably don't see a compile error because in each case `wayland-util.h` is already included by some other header before `workspaces.h`.

src/xwayland.c
jlindgren90: Sure, I can do that.

jlindgren90:
> That second commit is hard to follow (for me reading this late in the evening anyway 😄) Is it worth splitting it out into smaller commits doing fewer things?
> 
> EDIT: If possible - it would be easier for my brain to cope if the 'no functional change' (just pure refactoring) commits were separate.

Both these commit are functional changes (I think) -- I tried to put all the pure refactoring in #1109.
Renaming `topmost_mapped` to `topmost_focusable` (or just `topmost`) was meant to reflect the functional change, but I can move the renaming to its own commit if you prefer.

include/workspaces.h
johanmalm: Sorry. Forward declaring is only enough for pointers. Ignore me :blush:
As I said... long day.

jlindgren90:
Rebased again, diff should be smaller now that #1111 is merged.

johanmalm:
LGTM. 
Looks like a merge conflict - guess it's with #1112

Issue #1091, opened 2023-09-21
after alt-tab: endless phantom keypress eventtttttt


m1027:
Hi,

the issue I see is:

- First alt-tab to a console window (I am using alacritty in tmux)
- then the console receives keypress events (like mmmmmmm....) continously until I ctrl-c

More details:
 - The cursor glitches happen in a pure bash shall as well in a open vim file.
 - The phantom keyevent is not always the same. Maybe the last key I used in the terminal.
 - This happens apporx. once or twice a day.
 - This issue is present since approx. 3 months now.
 - My labwc version is mostly uptodate with git.

Do you have a hint what component could be responsible of this? Is it likely that labwc is concerned?

Thanks for awesome labwc.

stefonarch:
It should happen with every application that accepts keys, what about other terminal emulators or text editors?

heroin-moose:
It happens to me all the time :D

Electron apps, alacritty, firefox. You name it.

Consolatis:
Hm.. This sounds like something that was fixed ages ago. Personally I did't experience any kind of wrong key-repeat since then.
Does it only happen in xwayland or wayland native applications or does it show up for both?

There is also the https://labwc.github.io/labwc-config.5.html#entry_keyboard_repeatdelay setting which defaults to 600 ms which might eventually be related.

m1027:
@Consolatis: I am not using xwayland, just pure wayland. And I am using `<repeatDelay>300</repeatDelay>` and `<repeatRate>50</repeatRate>`.

heroin-moose:
Native windows:
```
    <repeatRate>40</repeatRate>
    <repeatDelay>200</repeatDelay>
```

kyak:
It happens to me as well. For some reason, the `l` key usually sticks.

This has something to do with swaylock, maybe. Because this happens sometimes after I unlock desktop, and I lock it by using `Super-l` combo (same `l` key that gets stuck).

Really hard to reproduce, it just happens.

heroin-moose:
To me it mostly happens on layout switches.

heroin-moose:
Yeah, I a way to reproduce!

1. Open a terminal
2. Press `s`
3. Without releasing `s` press right Alt to switch the layout
4. Release `s`

Works 10/10.

johanmalm:
> To me it mostly happens on layout switches.

Layout? As in keyboard-layout.... or window-layout?

heroin-moose:
[...]

PR #1108, opened 2023-09-26
keyboard: make keybind match stricter

johanmalm:
...and avoid failing to send release events to clients for any keys that were not absorbed by a keybind.

Do not match keybinds if there are other non-modifier keys (not part of any defined bind) pressed at the same time.

Only store non-modifier keycodes in the key-state.c 'pressed' array. This makes the call to wlr_seat_keyboard_notify_enter() in seat_focus() consistent with the equivalent in sway (in seat_keyboard_notify_enter()).

Fixes: issue #1091


Consolatis:
LGTM, nice fix.

PR #1109, opened 2023-09-27
view: try to reduce confusion in focused_view tracking

jlindgren90:
Our current approach to handling the focused/active view is a bit confusing. In particular, it's hard to be sure when `server->focused_view` is or isn't in sync with the real wlroots keyboard focus.

Try to clean things up a bit. In particular:

- Add comments to `server->focused_view` and `desktop_focused_view()` to clarify that they should match, but it's not guaranteed.
- `desktop_focused_view()` now prints a warning if it detects that `server->focused_view` is out of sync. We should keep an eye out for this warning, and if we see it, try to figure out why it happened.
- For consistency, use only "focus/defocus" as the verbs in function names rather than "activate". This is a bit arbitrary, but the idea is that focus is the primary action while the active/inactive state is a side effect.
- `view_focus/defocus()` replace `view_set_activated()` and now update both focus and active/inactive state, to try to keep them in sync.
- Add comments at `view_focus/defocus()` to warn against calling them directly (we should generally call the desktop.c functions).
- `desktop_focus_view(NULL)` is now forbidden and is no longer handled as a special case to clear the focus. This was (at least to me) a surprising behavior and caused trouble when working on another change.
- To maintain existing behavior, `desktop_focus_topmost_mapped_view()` now explicitly clears the focus if there are no mapped views.

There should be no behavioral change here.

Supersedes #1107.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -246,6 +246,13 @@ struct server {
 	uint32_t resize_edges;
 
 	/* SSD state */
+	/*
+	 * Currently focused view (cached value). This pointer is used
+	 * primarily to track which view is being drawn with "active"
+	 * SSD coloring. We consider it a bug if this gets out of sync
+	 * with the actual keyboard focus (according to wlroots). See
+	 * also desktop_focused_view().
+	 */
 	struct view *focused_view;
 	struct ssd_hover_state *ssd_hover_state;
 
@@ -370,7 +377,7 @@ void foreign_toplevel_update_outputs(struct view *view);
  *              cannot assume this means that the window actually has keyboard
  *              or pointer focus, in this compositor are they called together.
  */
-void desktop_focus_and_activate_view(struct seat *seat, struct view *view);
+void desktop_focus_view(struct view *view);
 void desktop_arrange_all_views(struct server *server);
 void desktop_focus_output(struct output *output);
 struct view *desktop_topmost_mapped_view(struct server *server);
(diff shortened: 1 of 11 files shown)


jlindgren90:
I guess this ended up with a lot of miscellaneous changes. Let me know if it would be better to split it up.

Consolatis:
`struct wlr_seat_keyboard_state` also has a `focus_change` signal with `old_surface` and `new_surface`.
That would be an alternative to our homegrown solution, e.g. something like get_root_surface + get_view_from_surface.

jlindgren90:
That could be a nice solution. If we watch the focus_change signal, then we could update server->focused_view in the handler (along with doing the SSD updates) and be more confident it's correct/up to date.

johanmalm:
LGTM

johanmalm:
Assume focus_change signal comment is intended for future enhancement rather than re-working this PR?

jlindgren90:
I'm good with leaving that for a later enhancement (even if it ends up reworking a lot of the same code). I think this PR is still a good incremental improvement as is.

johanmalm:
> I'm good with leaving that for a later enhancement (even if it ends up reworking a lot of the same code). I think this PR is still a good incremental improvement as is.

Agree, I’ll go for the merge shortly.

johanmalm:
Whilst we’re on the subject…
wlr_{xdg,xwayland}_surface->data points to view, so we could simplify the places where we iterate over all views just to find the one with current surface-focus

johanmalm:
https://github.com/labwc/labwc/blob/003d1fd494eed927ca2205a6dd3776c12ea4c9e9/src/xdg.c#L685

Consolatis:
> Assume focus_change signal comment is intended for future enhancement rather than re-working this PR?

Yep, was more of a general comment.

> `wlr_{xdg,xwayland}_surface->data` points to view, so we could simplify the places where we iterate over all views just to find the one with current surface-focus

Yes, maybe we should have some generic (maybe in view_impl?) `struct view *view_from_wlr_surface(struct wlr_surface *surface)` function around. Which then calls the xdg / xwayland ones for the internal casting to `wlr_xdg_surface` / `wlr_xwayland_surface` (+ eventual ->parent loop).

johanmalm:
> Yes, maybe we should have some generic (maybe in view_impl?) `struct view *view_from_wlr_surface(struct wlr_surface *surface)` function around. Which then calls the xdg / xwayland ones for the internal casting to `wlr_xdg_surface` / `wlr_xwayland_surface` (+ eventual `->parent` loop).

Oh, yes… nice!

PR #1111, opened 2023-09-27
desktop: switch workspaces and optionally raise in desktop_focus_view()

jlindgren90:
Make `desktop_focus_view()` always switch to the workspace containing the view being focused. It doesn't make much sense for an invisible view to have the keyboard focus.

Also add an optional "raise" parameter to `desktop_focus_view()`. This allows the common pattern of `desktop_focus_view()` + `view_move_to_front()` to be reduced to a single function call.

Split off from #1103.


include/workspaces.h
```diff
@@ -2,6 +2,9 @@
 #ifndef LABWC_WORKSPACES_H
 #define LABWC_WORKSPACES_H
 
+#include <stdbool.h>
+#include <wayland-util.h>
+
```
Consolatis, line 7: Nicely spotted the missing `<wayland-util.h>` include.

`struct view` and `struct wl_list` forward declares can be removed.
`struct wlr_scene_tree` forward declare is missing.

src/desktop.c
```diff
@@ -67,6 +67,18 @@ desktop_focus_view(struct view *view)
...
+		view_move_to_front(view);
+	}
+
+	/*
+	 * Switch workspace if necessary to make the view visible
+	 * (unnecessary for "always on top" views).
+	 */
+	if (!view_is_always_on_top(view)) {
+		workspaces_switch_to(view->workspace, /*update_focus*/ false);
+	}
```
Consolatis, line 81: Can we move this above the `view_focus()` line?
At least for me the mental model of
- switch to workspace where the view belongs to (e.g. "make it visible")
- focus the view
- maybe raise the view

is easier to follow than to focus / raise a potentially invisible view if its on a currently hidden workspace.

Consolatis:
Two minor comments. Tested and didn't spot anything obvious, including with `followMouse`.
Overall LGTM. I think this cleans up some stuff well.

include/workspaces.h
jlindgren90: Fixed

src/desktop.c
jlindgren90: Sure, fine by me

Consolatis:
Thanks :)

I have no clue what is going on with our FreeBSD CI abomination today, I've just canceled it.

PR #1112, opened 2023-09-27
seat: use focus_change event to update focused/active view

jlindgren90:
- Connect to `wlr_seat_keyboard_state`'s `focus_change` event.
- Add `view_from_wlr_surface()`, which does what the name says.
- Use `focus_change` event along with `view_from_wlr_surface()` to update `server->focused_view` and set SSD states via `view_set_activated()`.
- Eliminate `desktop_focused_view()` since `server->focused_view` should be reliably up-to-date now.
- Eliminate `view_focus/defocus()` since we can now just call seat_focus_surface() directly.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2,6 +2,7 @@
 #include <assert.h>
 #include <stdio.h>
 #include <strings.h>
+#include <wlr/xwayland.h>
 #include "common/mem.h"
 #include "common/scene-helpers.h"
 #include "labwc.h"
@@ -18,6 +19,34 @@
 #define LAB_MIN_VIEW_HEIGHT  60
 #define LAB_FALLBACK_WIDTH  640
 #define LAB_FALLBACK_HEIGHT 480
+
+struct view *
+view_from_wlr_surface(struct wlr_surface *surface)
+{
+	assert(surface);
+	/*
+	 * TODO:
+	 * - find a way to get rid of xdg/xwayland-specific stuff
+	 * - look up root/toplevel surface if passed a subsurface?
+	 */
+	if (wlr_surface_is_xdg_surface(surface)) {
+		struct wlr_xdg_surface *xdg_surface =
+			wlr_xdg_surface_from_wlr_surface(surface);
+		if (xdg_surface) {
+			return xdg_surface->data;
+		}
+	}
+#if HAVE_XWAYLAND
+	if (wlr_surface_is_xwayland_surface(surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(surface);
+		if (xsurface) {
+			return xsurface->data;
+		}
+	}
+#endif
+	return NULL;
+}
 
 static bool
 matches_criteria(struct view *view, enum lab_view_criteria criteria)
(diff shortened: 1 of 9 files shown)


jlindgren90:
Here's a try at implementing some of the cleanups we chatted about on #1109. I gave it a very brief test and didn't see anything amiss, but it could use more proving-out.

src/view.c
```diff
@@ -408,7 +383,6 @@ _minimize(struct view *view, bool minimized)
 	view->minimized = minimized;
 	if (minimized) {
 		view->impl->unmap(view, /* client_request */ false);
-		view_defocus(view);
```
jlindgren90, line 411: I checked and both of the `unmap()` implementations already call `desktop_focus_topmost_mapped_view()`, so the `view_defocus()` call was redundant and is safe to remove.

src/view.c
```diff
@@ -19,6 +20,25 @@
...
+	}
+#if HAVE_XWAYLAND
+	if (wlr_surface_is_xwayland_surface(surface)) {
+		struct wlr_xwayland_surface *xsurface =
+			wlr_xwayland_surface_from_wlr_surface(surface);
+		return xsurface->data;
+	}
+#endif
+	return NULL;
+}
```
Consolatis, line 40: This feels like it should be in `view-impl-common.c` instead.
IMHO we should strive to keep `view.c` clean of xdg and xwayland related stuff.

AFAIR `wlr_{xdg,xwayland}_surface_from_wlr_surface()` can return `NULL` as well so we should handle that.

jlindgren90: view-impl-common.c is "common code for shell `view->impl` functions" according to its own comment. This isn't called by (nor calls) any `view->impl` functions, so I'm not sure it belongs there.

It's unfortunate that this needs to do anything xdg- or xwayland-specific in the first place, since neither `view` nor `wlr_surface` are xdg- or xwayland-specific. Not sure of how to improve that at the moment though.

jlindgren90: Regarding returning `NULL` - you may be thinking of the `try_from` versions in wlroots 0.17. The `from` versions (without the `try`) in wlroots 0.16 never return `NULL` from what I can see, but will `assert(false)` if passed a `wlr_surface` with the wrong role. This function will look different with wlroots 0.17.

jlindgren90: One option to avoid all the `xdg` and `xwayland` stuff would be to loop over all the views like `desktop_focused_view()` did before. Slightly less efficient, but more generic.

Consolatis: AFAIR they only changed the name of the functions, they could always return `NULL`:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.16/include/wlr/types/wlr_xdg_shell.h#L488
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/0.16/include/wlr/xwayland/xwayland.h?ref_type=heads#L231

jlindgren90: > I don't remember if we have to loop up the wlr_surface tree until we find the main toplevel surface.

I am not sure either. The old `desktop_focused_view()` didn't loop to find the toplevel, but that doesn't mean it was correct.

jlindgren90: > AFAIR they only changed the name of the functions, they could always return `NULL`

Okay, I'll add a `NULL` check

jlindgren90: I added `NULL` checks and TODOs regarding the other comments.

jlindgren90:
Rebased to fix merge conflicts.

johanmalm:
Really nice. Thanks.

Issue #1114, opened 2023-09-28
after unlocking from swaylock: phantom keypress events


m1027:
Hi,

this is probably closely related to #1091. I am using labwc from git where the according PR has been integrated.

I am using `
    <keybind key="W-l">
		<action name="Execute"><command>swaylock</command></action>
	</keybind>
`.

I receive phantom `l` (small "L") keypress events in the following situation. This is *always* reproducable:
- Lock screen: `W-l`
- Unlock screen: `PW` + `Enter` to return into bash
- Press `l` once

Result: Then I receive `lllllllllllllllllll....` until I do `ctrl-c`.

Note: This does not happen when I press another key after unlocking. Only `l` tiggers that. Which may be related to my `W-l` mapping.

stefonarch:
I can't reproduce this here, but I could hardly reproduce this phantom keys anyway.

Consolatis:
@johanmalm I guess we don't reset the pressed state of the keybind here (via `key_state_bound_key_remove()`) because we return before that when `seat->server->session_lock != NULL`.

johanmalm:
I can reproduce the bug and confirm that #1115 fixes it.

Thanks @Consolatis for analysis + tell me how to do it.
I'm sure I would have got there, but having the solution spelt out is pretty handy :smile:

> Maybe its enough to just move the two `Ignore labwc keybindings if input is inhibited` blocks down before `uint32_t modifiers = wlr_keyboard_get_modifiers(wlr_keyboard);`?

Consolatis:
Should be fixed by latest master thanks to @johanmalm :)
Feel free to comment / reopen if that should not be the case.

m1027:
The above test succeeds now. Thank you!

PR #1115, opened 2023-09-28
keyboard: fix release event bug after session lock

johanmalm:
Do not withhold the release event associated with a keybind which executes a keyboard-input-inhibiting client like swaylock.

In other words, make release-events of absorbed keys (those used up by a keybind) de-register as such even though session-lock or input-inhibit is in-force.

Fixes: issue #1114

Helped-by: @Consolatis


Consolatis:
:)

PR #1116, opened 2023-09-28
Add RaiseLower action

johanmalm:


johanmalm:
Would appreciate some testing. I can prove the 'Lower' bit but can't think of how to invoke the action on a window that does not currently have the keyboard focus.

johanmalm:
Building on idea from @Consolatis 

`s/Click/Press/`
`s/Lower/RaiseLower/`

```
<mouse>
    <default />
    <context name="Close">
      <mousebind button="Left" action="Press">
        <action name="RaiseLower" />
      </mousebind>
    </context>
  </mouse>
```

Consolatis:
Code LGTM.
Why do we need the `s/Click/Press/` for the testcase?

johanmalm:
> Why do we need the `s/Click/Press/` for the testcase?

Not sure - just experimenting.

stefonarch:
I can't see an use case for this. Alt-tab, iconify, toggletobottom or else does it. Maybe I'm missing something.
Anyway, wiki: https://redtide.github.io/openbox-wiki/Help/Actions/#raiselower

johanmalm:
> I can't see an use case for this. 

On reflection, I agree.

It relies on having keyboard-focus which (for me) means that it's already raised. I'll delete it.

wesbluemarine:
The use case is to avoid using the keyboard when there is no need. Furthermore, having the dedicated button on the windows is not exactly the same thing. [Here](https://m.youtube.com/watch?v=9F81nFwZVCc&feature=youtu.be) 's an example.

Consolatis:
That same could be done using
- the `Lower` action to move the window to the back
- just clicking on any part of a window to `Raise` it to the top and give `Focus`.

What is missing for that workflow to work is some `Unfocus` action that ensures that the new window on top also gets keyboard focus. This PR would not have solved that though.

johanmalm:
@wesbluemarine What Consolatis said. I just don't think a specific RaiseLower action will help achieve the use-case you describe. I think we're talking apples and pears.

Issue #752, opened 2023-02-06
Prevent installing wlroots related files when using the subproject


Consolatis:
@droc12345
> Seeing you mention the wlroots wrap/subproject, I though I'd mention that I have patched meson.build and meson_options.txt to add a system-wlroots option, so that if it's toggled off the wlroots subproject is used. I've also added the option to make the wlroots wrap produce a static lib not shared.
> 
> The reason for this is I have several compositors that I look at and they all want different levels of wlroots. This make it kind of hard to have a system level wlroots for everyone (at this stage of development)

@joshuataylor
> That's awesome, as you mentioned other wlroots projects have different dependencies, and sometimes you might even want to run against the latest git with pending merge requests for example.

You can use
```
meson setup build_test --force-fallback-for=wlroots
meson configure build_test -Dprefix=/tmp/labwc_build_test
meson compile -C build_test
meson install -C build_test
```

The issue with that is that it still installs some parts of wlroots (like its headers, wlroots.pc, wlroots.a). We should figure out how to prevent that. PRs welcome :)

<details><summary>list of installed files</summary>

```
./share/man/man1/labwc.1
./share/man/man5/labwc-menu.5
./share/man/man5/labwc-actions.5
./share/man/man5/labwc-config.5
./share/man/man5/labwc-theme.5
./share/locale/it/LC_MESSAGES/labwc.mo
./share/locale/tr/LC_MESSAGES/labwc.mo
./share/locale/de/LC_MESSAGES/labwc.mo
./share/locale/sv/LC_MESSAGES/labwc.mo
./share/locale/es/LC_MESSAGES/labwc.mo
./share/doc/labwc/rc.xml
./share/doc/labwc/environment
./share/doc/labwc/themerc
./share/doc/labwc/menu.xml
./share/doc/labwc/rc.xml.all
./share/doc/labwc/README
./share/doc/labwc/autostart
./share/wayland-sessions/labwc.desktop
./lib/aarch64-linux-gnu/libwlroots.a
./lib/aarch64-linux-gnu/pkgconfig/wlroots.pc
./bin/labwc
./include/wlr/config.h
./include/wlr/backend/headless.h
./include/wlr/backend/interface.h
./include/wlr/backend/wayland.h
./include/wlr/backend/session.h
./include/wlr/backend/multi.h
./include/wlr/backend/libinput.h
./include/wlr/backend/x11.h
./include/wlr/backend/drm.h
./include/wlr/render/interface.h
[...]

PR #1117, opened 2023-09-28
README.md: describe --force-fallback-for option

johanmalm:


johanmalm:
@Consolatis I think we've discussed this addition before.
Useful?

README.md
```diff
@@ -163,6 +163,10 @@ Disable xwayland with `meson -Dxwayland=disabled build/`
 
 For OS/distribution specific details see see [wiki].
 
+To use the supplied wlroots.wrap file, run:
+
+    meson setup build --force-fallback-for=wlroots
```
Consolatis, line 168: Suggest `s/use/enforce/` or something like that to get the point across that this will ignore whatever wlroots is installed on the system.

Maybe we should also mention using the `meson install --skip-subprojects` flag if installing after using the `wlroots.wrap` file to prevent installing the wlroots headers.

Ref
- #752

johanmalm: Thanks.
Both done.

Consolatis:
LGTM

Issue #1118, opened 2023-09-29
failed to build commit: 4028a94


CaptainBloodz:
repository:               https://github.com/labwc/labwc
   updating from commit:     e5aef0331996aa0c77f7ea2f0af390b40704f626
   to commit:                4028a9482f9399e3c3587f5c6eca6c0b128c9afc

ends as

FAILED: labwc.p/src_view.c.o 
x86_64-pc-linux-gnu-gcc -Ilabwc.p -I. -I../labwc-9999 -Iinclude -I../labwc-9999/include -I/usr/lib64/libffi/include -I/usr/include/libdrm -I/usr/include/pixman-1 -I/usr/include/elogind -I/usr/include/libxml2 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/cairo -I/usr/include/libpng16 -I/usr/include/freetype2 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/fribidi -I/usr/include/librsvg-2.0 -I/usr/include/gdk-pixbuf-2.0 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=c11 -DWLR_USE_UNSTABLE -Wundef -Wlogical-op -Wmissing-include-dirs -Wold-style-definition -Wpointer-arith -Winit-self -Wstrict-prototypes -Wimplicit-fallthrough=2 -Wendif-labels -Wstrict-aliasing=2 -Woverflow -Wmissing-prototypes -Walloca -Wunused-macros -Wno-unused-parameter '-DLABWC_VERSION="0.6.5-10-g4028a94"' '-DGETTEXT_PACKAGE="labwc"' '-DLOCALEDIR="/usr/share/locale"' -march=native -mtune=native -O2 -flto -fuse-linker-plugin -fno-fat-lto-objects -pipe -Wa,-mbranches-within-32B-boundaries -pthread -MD -MQ labwc.p/src_view.c.o -MF labwc.p/src_view.c.o.d -o labwc.p/src_view.c.o -c ../labwc-9999/src/view.c
../labwc-9999/src/view.c:5:10: fatal error: wlr/xwayland.h: No such file or directory
    5 | #include <wlr/xwayland.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.

where [I] gui-libs/wlroots
[...]

PR #1120, opened 2023-09-29
src/view.c: guard xwayland include

Consolatis:
Fixes:
- #1118

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2,7 +2,6 @@
 #include <assert.h>
 #include <stdio.h>
 #include <strings.h>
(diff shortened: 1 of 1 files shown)


jlindgren90:
Yep, my fault, sorry about that. LGTM.

Consolatis:
Well, malm and me didn't see it either :)

jlindgren90:
Maybe we could add a CI build variant that builds wlroots without X11 support...

Consolatis:
Yep, my thought as well:
- #1121

PR #1122, opened 2023-09-29
various: stop using data argument of wlroots event callbacks

jlindgren90:
These are a couple of commits authored by @Consolatis that are cherry-picked from #626. They are for wlroots 0.17 compatibility, but have no negative side effects with wlroots 0.16 either. So I think we might as well merge them in now.

### DIFF
--- a/src/dnd.c
+++ b/src/dnd.c
@@ -14,7 +14,7 @@ static void
 handle_icon_map(struct wl_listener *listener, void *data)
 {
 	struct drag_icon *self = wl_container_of(listener, self, events.map);
-	struct wlr_drag_icon *icon = data;
+	struct wlr_drag_icon *icon = self->icon;
 	if (icon->data) {
 		struct wlr_scene_tree *surface_tree = icon->data;
 		wlr_scene_node_set_enabled(&surface_tree->node, true);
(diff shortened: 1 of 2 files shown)


Consolatis:
No objections in principle but those commits need a reword, the subjects start with `[something]` and the subject of the first one only refers to a change in the 0.17 branch.

jlindgren90:
I updated the commit message to something more useful.

The changes are both similar to b816c1670150e265719b2fc2ff2a149b50181298, which was also for wlroots 0.17 compatibility but merged to master. I don't know that they're really related to the map-unify stuff (though the original commit message said so).

Consolatis:
>  I don't know that they're really related to the map-unify stuff (though the original commit message said so).

They are somewhat related: https://github.com/labwc/labwc/pull/626/commits/3849761f7373904b913773d55cee0e7e1ab5c173#diff-cbf50fbf541217d1e538e1d4f286a6f65b09ab5eb076a822b916b9df70c38177 moved the map event over from `wlr_drag_icon` which supplied itself in the map handler `*data` pointer to `wlr_surface` which doesn't.

I slightly modified the commit messages so they contain the upstream commits.

PR #1125, opened 2023-09-30
Update README with build details and a toning down of the no-frills approach

johanmalm:
@Consolatis @jlindgren90 @heroin-moose If you have time, any feedback on the third commit (on icons/gradients/dropshadows) would be appreciated - particularly if you disagree with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,9 +164,15 @@ Disable xwayland with `meson -Dxwayland=disabled build/`
 
 For OS/distribution specific details see see [wiki].
 
+If the right version of `wlroots` is not found on the system, the build setup
+will automatically download the wlroots repo. If this fallback is not desired
+please use:
+
+    meson setup --wrap-mode=nodownload build/
+
 To enforce the supplied wlroots.wrap file, run:
 
-    meson setup build/ --force-fallback-for=wlroots
+    meson setup --force-fallback-for=wlroots build/
 
 If installing after using the wlroots.wrap file, use the following to
 prevent installing the wlroots headers:
--- a/README.md
+++ b/README.md
@@ -151,7 +151,8 @@ Run-time dependencies include:
 
 - wlroots, wayland, libinput, xkbcommon
 - libxml2, cairo, pango, glib-2.0
-- libpng, librsvg-2.0
+- libpng
+- librsvg >=2.46 (optional)
 - xwayland, xcb (optional)
 
 Build dependencies include:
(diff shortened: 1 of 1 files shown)


heroin-moose:
I have only one stupid style complaint about `meson setup <dir> --flag=1` instead of `meson setup --flag=1 <dir>`, but that's how the rest of the file is done.

README.md
```diff
@@ -164,6 +164,12 @@ Disable xwayland with `meson -Dxwayland=disabled build/`
 
 For OS/distribution specific details see see [wiki].
 
+If the right version is `wlroots` is not found on the system, the build setup
```
heroin-moose, line 165: of `wlroots`?

README.md
```diff
@@ -151,7 +151,8 @@ Run-time dependencies include:
 
 - wlroots, wayland, libinput, xkbcommon
 - libxml2, cairo, pango, glib-2.0
-- libpng, librsvg-2.0
+- libpng
+- librsvg-2.0 >=2.46 (optional)
```
heroin-moose, line 153: Maybe `libvrsvg >= 2.46`?

README.md
01micko: yeah `s/is/of/`

Consolatis:
> any feedback on the third commit (on icons/gradients/dropshadows)

Fine with me. No issues with icons and gradients.
If we ever end up implementing drop shadows we should make sure that
- they either work fine on really low resource systems and software rendering (stuff like alpha and blur can be costly on those systems)
- they can be disabled and default to be disabled on software rendering (similar as to how the region overlays react to software rendering)

jlindgren90:
I do miss window icons from Openbox so I wouldn't mind trying to implement those at some point. Maybe gradients too (to a lesser extent).

I don't use drop shadows personally and would just be concerned with how much complexity it adds. If it's easy to implement with scene-graph and doesn't add a bunch of corner cases, then I'm fine with it.

johanmalm:
Thanks all. Can't believe I got so many things wrong with a couple of minor changes to the README :smile: Just goes to show the power of review.

I changed the format to `meson setup --flag <dir>`. Agree that feels better.

Agree with thoughts on alpha/drop-shadows.

README.md
johanmalm: Done

README.md
johanmalm: Thanks.

jlindgren90:
I noticed that icons and gradients are still listed under "Items which are not intended to be implemented" in README.md (section 1.4). Maybe that section should be updated too?

johanmalm:
> I noticed that icons and gradients are still listed under "Items which are not intended to be implemented" in README.md (section 1.4). Maybe that section should be updated too?

Good point. https://github.com/labwc/labwc/pull/1127

PR #1126, opened 2023-09-30
key-state: rename function for consistency

johanmalm:
...with the equivalent function that returns the number of pressed keys.

s/key_state_nr_keys/key_state_nr_bound_keys/


Consolatis:
:)

PR #1128, opened 2023-10-01
Refactor title height settings to use values from themerc

mozlima:
**Description:**
Openbox and Labwc currently calculate the title size based on the font size. This can be limiting for users who want to customize the window title's appearance. For instance, if someone desires a larger title height without increasing the font size, the current system doesn't provide an solution.

**Proposed Change:**
This pull request introduces a new configuration setting in the 'themerc' configuration file: `title.height`. With this setting, users can now specify the desired title height independently of the font size. If the font size specified is larger than the `title.height` setting, then will use the font size.

**How to Use:**
1. Edit your 'themerc' configuration file.
2. Add or modify the `title.height` setting, specifying your desired title height.
3. Save the configuration file.

**Example 'themerc' Configuration:**
```ini
# Set the title height to 24 pixels
title.height = 24
```


Consolatis:
Thanks. The code itself looks mostly good to me, two small issues:
- its missing documentation in `docs/labwc-theme.5.scd`
- it must reset `theme->title_height` to some sane value (for example `0`) in `theme_builtin()`.
  Otherwise, a deleted `title.height` theme entry is still set after  `Reconfigure`.

Regarding the feature itself:
Although I like the clarity of the name, I wonder what other benefits this brings. Can't we do something similar with the combination of font size and `padding.height`?

In any case, @johanmalm has to decide on this as its a deviation of the Openbox theme specs so I suggest waiting with updates for the above mentioned issues to not waste time on a feature that might not end up getting accepted.

johanmalm:
I'm up for being able to set the titlebar height independently of the font size (provided the font fits). I thought about this right at the start of developing `labwc`, but opted for the `openbox` approach at the time.

I suggest a `titlebar.height` theme option which if (and only if) set overrides the font-height+2*padding.height calculation. That way we would maintain backward compatibility with openbox themes, but users could override it by setting `titlebar.height` if they choose to. Welcome thoughts on this alternative.

mozlima:
Hello,
Thanks for the review! 

@Consolatis
> Regarding the feature itself: Although I like the clarity of the name, I wonder what other benefits this brings. Can't we do something similar with the combination of font size and `padding.height`?

I noticed that when I set `padding.height` in `themerc`, it affects the Openbox title text line-height (text not vertical centralized).

@johanmalm 
 > suggest a titlebar.height theme option which if (and only if) set overrides the font-height+2*padding.height calculation. That way we would maintain backward compatibility with openbox themes, but users could override it by setting titlebar.height if they choose to. Welcome thoughts on this alternative.

Are you suggesting that we simply exclude the padding from the `titlebar.height` calculation, like this:
```
//reset
theme->title_height = 0;
//parse theme...

//post_processing...
theme->title_height = MAX( theme->title_height , font_height(&rc.font_activewindow) + 2 * theme->padding_height)
```

The notion that 'padding is internal to box size calculation like [margin:size:padding]' indeed makes sense.  
However, as you mentioned earlier with Openbox's themerc,  can lead to issues.


What should happen if font size is bigger than `titlebar.height` ?
1. Adapt titlebar to font
2. Adapt font to titlebar
3.  Font overflow..?

I think the toolbar should grow with the font size because the user may configure the font without needing to edit the theme.

English is not my native language, so if I make any mistakes or sound unclear, please feel free to let me know.

PR #1132, opened 2023-10-02
Refactor title height to use 'titlebar.height' from themerc

mozlima:
**Description:**
Openbox and Labwc currently calculate the title size based on the font size. This can be limiting for users who want to customize the window title's appearance. For instance, if someone desires a larger title height without increasing the font size, the current system doesn't provide an solution.

**Proposed Change:**
This pull request introduces a new configuration setting in the 'themerc' configuration file: `titlebar.height`. With this setting, users can now specify the desired title height independently of the font size. If the font size specified is larger than the `titlebar.height` setting, then will use the font size.

**How to Use:**
1. Edit your 'themerc' configuration file.
2. Add or modify the `titlebar.height` setting, specifying your desired title height.
3. Save the configuration file.

**Example 'themerc' Configuration:**
```ini
# Set the title height to 24 pixels
titlebar.height = 24
```

Continuing from #1128 , the previous pull got messed up due to an editor issue. I'll now send a new pull request to get the latest changes

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -47,6 +47,10 @@ labwc-config(5).
 	Vertical padding size, used for spacing out elements in the window decorations.
 	Default is 3.
 
+*titlebar.height*
+	Window title bar height.
+	Default equals the vertical font extents of the title plus 2x padding.height.
+
 *menu.items.padding.x*
 	Horizontal padding of menu text entries in pixels.
 	Default is 7.
@@ -201,11 +205,6 @@ with the respective titlebar colors. For example: "close-active.png"
 # DEFINITIONS
 
 The handle is the window edge decoration at the bottom of the window.
-
-# DERIVED DIMENSIONS
-
-The window title bar height is equal to the vertical font extents of the title.
-Padding will be added to this later.
 
 # SEE ALSO
 
--- a/src/theme.c
+++ b/src/theme.c
@@ -213,6 +213,7 @@ theme_builtin(struct theme *theme)
 {
 	theme->border_width = 1;
 	theme->padding_height = 3;
+	theme->title_height = INT_MIN;
 	theme->menu_overlap_x = 0;
 	theme->menu_overlap_y = 0;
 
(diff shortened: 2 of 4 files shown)


src/theme.c
```diff
@@ -213,6 +213,7 @@ theme_builtin(struct theme *theme)
 {
 	theme->border_width = 1;
 	theme->padding_height = 3;
+	theme->title_height = 0;
```
johanmalm, line 216: Suggest setting this to `INT_MIN` (representing unset) and then fallback to the default in `post_processing()` if not set.

src/theme.c
```diff
@@ -750,8 +755,9 @@ create_corners(struct theme *theme)
 static void
 post_processing(struct theme *theme)
 {
-	theme->title_height = font_height(&rc.font_activewindow)
-		+ 2 * theme->padding_height;
+	int h = font_height(&rc.font_activewindow) + 2 * theme->padding_height;
+	theme->title_height = MAX(theme->title_height, h);
```
johanmalm, line 759: Suggest changing this to the code below so that `theme.titlebar_height` doesn't just apply when it's greater than font-height + 2 * padding_height.

```
/* Fallback on font-height + 2x padding.height if not set */
if (theme->title_height > 0) {
	theme->title_height = font_height(&rc.font_activewindow)
		+ 2 * theme->padding_height;
}
```

docs/labwc-theme.5.scd
```diff
@@ -47,6 +47,10 @@ labwc-config(5).
 	Vertical padding size, used for spacing out elements in the window decorations.
 	Default is 3.
 
+*titlebar.height*
+	Window title bar height.
+	Default equals the vertical font extents of the title plus Padding.
```
johanmalm, line 52: Suggest spelling out 2x `padding.height` instead of `Padding`

johanmalm (commented):
Thanks!

I have made a few inline comments. Also suggest the following:

- In `include/theme.h`, move `int title_height` up to just below `int padding_height`
- Add an entry in `docs/themerc`. We like to have all theme options with default values in this file, so wouldn't want to miss out on this one. I think it'll make most sense to enter it as a commented out option as the default-fallback is a formula. Something like:


```
# The following options has no default, but fallbacks back to
# font-height + 2x padding.height if not set.
#
# titlebar.height:
```

It terms of doing `fixup`s just keep adding commits and we can squash later. Or if you want just do `git commit --amend` and then `git push -f origin title_height`

src/theme.c
johanmalm: Sorry, should of course be:

```
int h = font_height(&rc.font_activewindow);
if (theme->title_height < h) {
	theme->title_height = h + 2 * theme->padding_height;
}
```

...to avoid titlebar being smaller than than font-height which would be no good.

johanmalm:
Thanks!

Issue #1131, opened 2023-10-02
minimize-raise in waybar's taskbar stopped working 


kyak:
The "minimize-raise" has stopped working as intended in waybar's taskbar. It only does "raise", but not "minimize".

Waybar hasn't updated recently, but labwc has (i'm using the latest master).

Consolatis:
Hmmm, can confirm. This is kind of bad. Will investigate later.

Consolatis:
The reason this happens is actually around for a while already but with the recent change in focus notifications it become an issue:

https://github.com/labwc/labwc/blob/3e5b988d38a55f40ad90e56d1f3e9b084c3f792e/src/cursor.c#L917-L919

We are giving layershell subsurfaces keyboard focus without checking for the keyboard interactivity setting. Commenting that branch out fixed it for me. Basically a click on a tasklist button gives keyboard focus to that button which removes it from the active window and thus confuses the panel logic because it now thinks the window doesn't have focus and in turn only requests focus for the window rather than requesting minimize.

However, I am not sure how to proceed, just removing that branch completely? Or getting the root layershell surface and check the keyboard interactivity?

Thoughts @johanmalm ?

johanmalm:
@Consolatis Great detective work.
I can't think of a way to get node-descriptors for these layer-subsurfaces, so think we have to find the top-parent.

johanmalm:
I've submitted to different PRs which solves the problems in slightly different ways (same general approach though).
Need to get off to bed, but appreciate feedback + help on deciding.

Consolatis:
Small update: after some discussion via IRC we decided to go with a 2 stage deployment:
- fix the panel regression (already merged: #1135)
- develop a proper fix in the next days that handles all the nuances regarding sub-surfaces (layershell or not) correctly

Feel free to reopen / comment if latest master still shows the issue.

stefonarch:
DND of buttons in waybar seems fixed now too. I don't use minimize/raise.

johanmalm:
> DND of buttons in waybar seems fixed now too. I don't use minimize/raise.

Great observation + feedback. This is thanks to @jlindgren90's changes relating to seat->keyboard_state `focus_change` event.

PR #1135, opened 2023-10-02
cursor: fix layer subsurface focus bug

johanmalm:
...by simply not focusing layer-shell subsurfaces on button press.

This is a temporary patch to fix a regression.

Fixes: issue #1131


Consolatis:
Thanks :)

PR #1137, opened 2023-10-03
Allow XWayland views with WM_HINTS.input = false to be focused again

### LINKED ISSUE #1130: Xwayland application loose focus
Something's wrong when trying to focus Xwayland apps.

For example, run xeyes. Then switch to another window. Then try to switch back to xeyes.

First of all, xeyes is not even in the Alt+Tab list. Secondly, it is shown in waybar's taskbar, but clicking it doesn't raise the window, nothing happens.

jlindgren90:
Revert "desktop: try harder to avoid focusing unfocusable views"
    
Some X11 applications (MATLAB is known to be one) apparently still use
the outdated "globally active" input focus model, in which they declare
they don't want the window manager to give them input focus, but expect
to be able to take it explicitly themselves via XSetInputFocus().

Such applications are not a good fit for the Wayland world, and may have
issues even with remotely modern X11 window managers that prevent such
"focus stealing". Labwc certainly doesn't (and won't) allow it. However,
to avoid breaking such applications entirely, let's still allow the user
to give focus by clicking in the window.

For the sake of applications that legitimately don't want to be given
input focus (such as taskbars or other "panels"), we still don't give
focus to them automatically when another view is closed, and they aren't
shown in Alt-Tab.

This reverts commit cae96b0cce441dd9f1981c312ae0108b3b4e42c9.

Along with the revert (1st commit), add some comments to try to clarify
this whole mess (2nd commit).

This undoes half of #1103 and should half-fix #1130. As mentioned in
#1130, it's a compromise that probably won't make everyone 100% happy.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -375,6 +375,18 @@ void foreign_toplevel_update_outputs(struct view *view);
  *              cannot assume this means that the window actually has keyboard
  *              or pointer focus, in this compositor are they called together.
  */
+
+/**
+ * desktop_focus_view() - do multiple things to make a view "active" and
+ * ready to use:
+ *  - unminimize
+ *  - switch to the workspace it's on
+ *  - give input (keyboard) focus
+ *  - optionally raise above other views
+ *
+ * It's okay to call this function even if the view isn't mapped or the
+ * session is locked/input is inhibited; it will simply do nothing.
+ */
 void desktop_focus_view(struct view *view, bool raise);
 void desktop_arrange_all_views(struct server *server);
 void desktop_focus_output(struct output *output);
@@ -393,6 +405,15 @@ enum lab_cycle_dir {
  */
 struct view *desktop_cycle_view(struct server *server, struct view *start_view,
 	enum lab_cycle_dir dir);
+
+/**
+ * desktop_focus_topmost_view() - focus the topmost view on the current
+ * workspace, skipping views that claim not to want focus (those can
+ * still be focused by explicit request, e.g. by clicking in them).
+ *
+ * This function is typically called when the focused view is hidden
+ * (closes, is minimized, etc.) to focus the "next" view underneath.
+ */
 void desktop_focus_topmost_view(struct server *server);
 
 void keyboard_cancel_keybind_repeat(struct keyboard *keyboard);
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -73,6 +73,13 @@ desktop_focus_view(struct view *view, bool raise)
 		workspaces_switch_to(view->workspace, /*update_focus*/ false);
 	}
 
+	/*
+	 * Give input focus, even if the view claims not to want it (see
+	 * view->impl->wants_focus). This is a workaround for so-called
+	 * "globally active" X11 views (MATLAB known to be one such)
+	 * that expect to be able to control focus themselves, but can't
+	 * under labwc since it's disallowed at the wlroots level.
+	 */
 	struct seat *seat = &server->seat;
 	if (view->surface != seat->seat->keyboard_state.focused_surface) {
 		seat_focus_surface(seat, view->surface);
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -38,16 +38,13 @@ void
 desktop_focus_view(struct view *view, bool raise)
 {
 	assert(view);
-	if (!view_isfocusable(view)) {
-		return;
-	}
-
 	/*
 	 * Guard against views with no mapped surfaces when handling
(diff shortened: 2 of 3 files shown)


kyak:
@jlindgren90 i've tried this branch and it works for me. Thank you!

As for the Alt-Tab, only the main MATLAB window is not listed in the Alt-Tab, but other MATLAB windows do show up. As long as taskbar buttons work, this is something that we can live with.

kyak:
P.S. Dropdown menus (like those in the main menu of MATLAB) don't work (they briefly appear and disappear) - but they also didn't work before in labwc.

It's unfortunate that I can't provide you with an example of another app that exhibits this behaviour..

dirk-hub:
I think IntelliJ IDEA is also affected, at least it isn't in the tab-list, does not respond to menu actions but can be closed when clicking the SSD close icon.
Previously it worked fine.
Can not test with the branch, sorry.

johanmalm:
LGTM

Consolatis:
> and they aren't shown in Alt-Tab.

Is this the correct thing to do? I was under the impression that there is some kind of skip_pager property for that.

jlindgren90:
> Is this the correct thing to do? I'd think a window that doesn't want input focus may still want to get selected / raised.

I'm not sure there's a 100% "correct" answer here. I could see an argument either way here. Okay to merge the revert and leave the Alt-Tab discussion for later?

jlindgren90:
If we add support for _NET_WM_STATE_SKIP_TASKBAR in the future, I'd be fine relying only on the that and not WM_HINTS for the Alt-Tab logic.

X11 often has older and newer protocols that have some overlap. I think that's the case here.

Consolatis:
I'd prefer if we could reduce the surprise effect as much as possible, e.g. I am ok with the compromise of preventing auto focus for `input=false` windows but allowing users to select them due to the reasons you stated above. However, that might be an issue for users that do not use a taskbar as this seems difficult to do without them showing up in `A-Tab` (e.g. if they are hidden behind other windows).

Fine with merging this for now (not tested) but I think we should try to restore the window being visible via `A-Tab` relatively soon as from what I gathered so far - as an observer rather than somebody being affected by it - the no-focus hint seems to cause more issues / regressions than it provides benefits. I may be wrong on this though.

jlindgren90:
> Random idea: could we somehow translate `XSetInputFocus()` to `request_activate()` and check for a common ancestor window with the currently focused one?

`xwm_handle_focus_in()` in xwm.c (wlroots) already has some logic along these lines. From the comment: "we allow changing focus between surfaces belonging to the same application". Other `XSetInputFocus()` requests are ignored (and hidden from the compositor). I've wondered about adding a signal that labwc could listen to, and use those requests as a hint that "this view wants focus" (overriding the `WM_HINTS.input` logic). But it seems a bit complicated, and I don't have a clear idea how it would work.

jlindgren90:
Thanks for the reviews, I'll go ahead and merge this. I've created #1139 to capture the _NET_WM_SKIP_PAGER suggestion from @Consolatis.

jlindgren90:
> P.S. Dropdown menus (like those in the main menu of MATLAB) don't work (they briefly appear and disappear) - but they also didn't work before in labwc.

@kyak this particular issue may be solved by #1163 - give it a try.

kyak:
> > P.S. Dropdown menus (like those in the main menu of MATLAB) don't work (they briefly appear and disappear) - but they also didn't work before in labwc.
> 
> @kyak this particular issue may be solved by #1163 - give it a try.

@jlindgren90 well, these dropdown menus don't appear at all with the latest master. Instead, focus from MATLAB is completely lost and another application gets focused. So experience is like this: click a dropdown menu in MATLAB, the main window disappears and another application pops up.

Maybe, if you rebase https://github.com/labwc/labwc/pull/1142 on the latest master, the behaviour will be different. Because I can at least see these dropdown menus with this PR (although can't click it).

jlindgren90:
> So experience is like this: click a dropdown menu in MATLAB, the main window disappears and another application pops up.

Sorry to hear that. Do you know when that started happening? Is it a new regression since this PR (#1137) was merged?

kyak:
> > So experience is like this: click a dropdown menu in MATLAB, the main window disappears and another application pops up.
> 
> Sorry to hear that. Do you know when that started happening? Is it a new regression since this PR (#1137) was merged?

This exact behaviour started happening since #1103.

Dropdown menus still didn't work before #1103, but they also didn't switch focus to another app.

jlindgren90:
All right, we'll continue working through the issues. Thanks for the info.

PR #1140, opened 2023-10-04
keyboard: allow switching VT when locked

jlindgren90:
I noticed that one currently can't switch to a different VT if the labwc session is locked. It seems undesirable to have to unlock one VT session just to switch to another one (which potentially belongs to a different user).

Assuming this was unintentional, can we allow switching VTs while locked? It appears to be as simple as reordering the two logic steps in `handle_compositor_keybindings()`.

### DIFF
--- a/src/keyboard.c
+++ b/src/keyboard.c
@@ -257,20 +257,6 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 		return true;
 	}
 
-	/*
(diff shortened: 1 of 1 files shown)


johanmalm:
I hadn't thought about this before, but VT switching seems fine even with a session-lock in place (so second bullet in @Consolatis list).

johanmalm:
With thanks :smile:

Issue #885, opened 2023-04-18
swaylock failure on display hotplug


kode54:
I had a situation where swaylock was left on my displays, and swayidle was also set to wlopm --off them both, and when I returned, it turned them on, and somehow swaylock had been removed from the one that doesn't usually hotplug itself when DPMS'd. I could not reproduce it though.

Hmm, I wonder if this is because of a problem with #626 being applied and in use? I'm definitely making use of the hidpi-xprop patch, which I verified doesn't apply properly to 0.16.

johanmalm:
We added support for ext-session-lock protocol a couple of weeks ago, but if the `0.17.0` branch has not been rebased since (so does not include this).

Could you confirm that you are using the branch tracking `wlroots-0.17.0` (I think that's what you've said in OP but just want to check).

Are you able to reproduce the issue with the `master` branch (i.e. not the one tracking `0.17.0`)

kode54:
I am using the `wlroots-0.17.0` patch, applied to `master`.

kode54:
I have reproduced an issue, but it isn't exactly as I described. I wasn't able to make hotplugging reset the display lock, but I was able to make a display lose its display lock and also lose its swaylock surfaces, by commanding that display to change its scaling using an SSH session and wlr-randr. This is similar to my previous swayidle that commanded kanshi to restore the display modes after turning the monitors back on.

https://gist.github.com/kode54/0af4fc82ff1957cd08604aab5bef5684

Swaylock issue I filed, though it doesn't appear to be their fault:

https://github.com/swaywm/swaylock/issues/313

Consolatis:
Short update on the situation, it seems this can not be reproduced on latest master + wlroots 0.16.x and only happens when running with the wip wlroots 0.17.x branch at #626. There might be a subtle wlroots change for 0.17 that we don't react to properly.

Behind the lock surfaces there should always be a black rectangle with the same dimension as the output itself, positioned at `0,0` of that output. We will do some more debugging with that wlroots 0.17 wip branch applied.

johanmalm:
Can't get to an extra screen for a couple of days, but when I do I will try connecting/disconnecting to see if I can reproduce it.

johanmalm:
I've been trying with the two scripts below be cannot reproduce this with the `master` branch.

swaylock-test

```
#!/bin/sh
swaylock-logging &
sleep 2
echo "turning off with wlopm"
wlopm --off \*
sleep 10
# (un)plug hdmi cable to second screen
[...]

PR #1141, opened 2023-10-05
session-lock: reconfigure for output layout changes

jlindgren90:
Currently, if the output layout changes while the session is locked, the lock surfaces may end up wrongly positioned, which looks bad and may reveal some of the user's workspace underneath.

As an example, this is what my desktop looks like after docking my laptop:
image

To prevent this, reposition the session-lock scene trees and reconfigure the lock surfaces when the output layout changes.

P.S. It may also make sense to disable/destroy the lock surfaces for disabled outputs. I think the black rectangle in the picture was originally meant for the laptop display, which turns off when docked. (It's not visible after this change - maybe just hidden behind one of the other lock surfaces.)

### DIFF
--- a/include/session-lock.h
+++ b/include/session-lock.h
@@ -4,6 +4,9 @@
 
 #include <wlr/types/wlr_session_lock_v1.h>
 
+struct output;
+struct server;
+
 struct session_lock {
 	struct wlr_session_lock_v1 *lock;
 	struct wlr_surface *focused;
@@ -18,5 +21,6 @@ struct session_lock {
 
 void session_lock_init(struct server *server);
 void session_lock_output_create(struct session_lock *lock, struct output *output);
+void session_lock_update_for_layout_change(void);
 
 #endif /* LABWC_SESSION_LOCK_H */
--- a/src/output.c
+++ b/src/output.c
@@ -272,6 +272,7 @@ static void
 output_update_for_layout_change(struct server *server)
 {
 	output_update_all_usable_areas(server, /*layout_changed*/ true);
+	session_lock_update_for_layout_change();
 
 	/*
 	 * "Move" each wlr_output_cursor (in per-output coordinates) to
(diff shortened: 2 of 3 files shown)


Consolatis:
Likely related:
- #885

Code LGTM, didn't test (only have a single HDMI output on my daily driver).

> P.S. It may also make sense to disable/destroy the lock surfaces for disabled outputs. I think the black rectangle in the picture was originally meant for the laptop display, which turns off when docked. (It's not visible after this change - maybe just hidden behind one of the other lock surfaces.)

I assume the output is merely disabled but not disconnected and we only destroy `output->session_lock_tree` on output disconnect. Maybe we should react to the output being enabled/disabled as well and enable/disable `lock_output->tree`?

CC @johanmalm

johanmalm:
LGTM 

I had been thinking about moving the output related `wlr_scene_node_set_position()s` to `output_update_for_layout_change()`, but as there is more to it, what you've done looks better.

As a side note, I think the only other one is in `layers.c` https://github.com/labwc/labwc/blob/b7bd5a52e44283a0f551ede833bff5197ed53e65/src/layers.c#L92

Funnily enough I can now reproduce #885 - with a kanshi that makes my eDP-1 move when I plug in the HDMI. It only ever happens if the cable is disconnected when I start labwc and if I then connect the HDMI. I get no related error logs from either labwc or swaylock, and the swaylock surface remains (so not dead) - it just doesn't receive keyboard input. This patch doesn't fix that, but I can confirm that it now positions the surfaces correctly, which it does not with out it.

@Consolatis considering I can now reproduce it, I'll try to find something with what I have rather pursuing the `wlr_output` `enable` signal.

jlindgren90:
I saw an issue with swaylock not receiving keyboard input after hotplug a couple of times also. I couldn't reproduce it reliably though.

Incidentally, swaylock losing keyboard input is a good argument for being able to switch VTs as a last resort (#1140).

Thanks for reviews and merging :)

Issue #637, opened 2022-11-21
Document how to enable NumLock automatically


jlindgren90:
I couldn't find in the documentation how to enable keyboard NumLock automatically.
Do we have a method for doing this?

For Sway, you can apparently enable NumLock by adding to `~/.config/sway/config`:
```
input type:keyboard xkb_numlock enabled
```

johanmalm:
I don’t know how that’s done. We’ve not implemented anything to support numlock enable/disable.
It’s possible that some environment variable (XKB_DEFAULT_OPTIONS or similar) can do this, but I’ve not investigated.

01micko:
This is what I am using right now - _ugly_

```c
#include <fcntl.h>
#include <sys/ioctl.h>
#include <unistd.h>
#include <linux/uinput.h>
#include <stdlib.h>

int main(int argc, char *argv[]) {
	struct uinput_setup setup = {.name = "numlock"};
	struct input_event ev;
	int uinput, i;

	uinput = open("/dev/uinput", O_WRONLY | O_NONBLOCK);
	if (uinput < 0) {
		return EXIT_FAILURE;
	}

	if ((ioctl(uinput, UI_SET_EVBIT, EV_KEY) < 0) ||
		(ioctl(uinput, UI_SET_EVBIT, EV_SYN) < 0) ||
		(ioctl(uinput, UI_SET_KEYBIT, KEY_NUMLOCK) < 0) ||
		(ioctl(uinput, UI_DEV_SETUP, &setup) < 0) ||
		(ioctl(uinput, UI_DEV_CREATE) < 0)) {
		close(uinput);
		return EXIT_FAILURE;
	}

	sleep(5); /* time for device to be created - could be slow */
	for(i=1; i>=0; i--) {
		ev.type = EV_KEY;
		ev.code = KEY_NUMLOCK;
		ev.value = i;

		if (write(uinput, &ev, sizeof(ev)) != sizeof(ev)) {
			return EXIT_FAILURE;
		}
	}

	ev.type = EV_SYN;
	ev.code = 0;
	ev.value = 1;

	if (write(uinput, &ev, sizeof(ev)) != sizeof(ev)) {
		return EXIT_FAILURE;
	}

	ioctl(uinput, UI_DEV_DESTROY);
	close(uinput);

	return EXIT_SUCCESS;

}
```

[...]

PR #1143, opened 2023-10-06
keyboard: set numlock automatically

jlindgren90:
This should get us most of the way to solving #637. I haven't coded up the bits to hook it up to rc.xml; help with that would be welcome.

A minor surprise is that the initial NumLock state is re-applied (to all keyboards) if a new keyboard is hot-plugged. It's wlroots doing this behind our back, though with a little more code we could prevent it. In my opinion, it's a pretty minor issue (but then, I basically never turn off NumLock).

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -276,6 +276,10 @@ this is for compatibility with Openbox.
 
 ## KEYBOARD
 
+*<keyboard><numlock>* [on|off]
+	When recognizing a new keyboard enable or disable Num Lock.
+	Default is on.
+
 *<keyboard><keybind key="" layoutDependent="">*
 	Define a *key* binding in the format *modifier-key*, where supported
 	modifiers include S (shift); C (control); A (alt); W (super). Unlike
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -141,6 +141,7 @@
     your favourite terminal or application launcher. See rc.xml for an example.
   -->
   <keyboard>
+    <numlock>on</numlock>
     <repeatRate>25</repeatRate>
     <repeatDelay>600</repeatDelay>
     <keybind key="A-Tab">
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -62,6 +62,7 @@ struct rcxml {
 	/* keyboard */
 	int repeat_rate;
 	int repeat_delay;
+	bool kb_numlock_enable;
 	struct wl_list keybinds;   /* struct keybind.link */
 
 	/* mouse */
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -419,6 +419,7 @@ void desktop_focus_topmost_view(struct server *server);
 void keyboard_cancel_keybind_repeat(struct keyboard *keyboard);
 void keyboard_key_notify(struct wl_listener *listener, void *data);
 void keyboard_modifiers_notify(struct wl_listener *listener, void *data);
+void keyboard_set_numlock(struct wlr_keyboard *keyboard);
 void keyboard_init(struct seat *seat);
 bool keyboard_any_modifiers_pressed(struct wlr_keyboard *keyboard);
 void keyboard_finish(struct seat *seat);
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -732,6 +732,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.repeat_rate = atoi(content);
 	} else if (!strcasecmp(nodename, "repeatDelay.keyboard")) {
 		rc.repeat_delay = atoi(content);
+	} else if (!strcasecmp(nodename, "numlock.keyboard")) {
+		set_bool(content, &rc.kb_numlock_enable);
 	} else if (!strcasecmp(nodename, "screenEdgeStrength.resistance")) {
 		rc.screen_edge_strength = atoi(content);
 	} else if (!strcasecmp(nodename, "range.snapping")) {
@@ -949,6 +951,7 @@ rcxml_init(void)
 	rc.scroll_factor = 1.0;
 	rc.repeat_rate = 25;
 	rc.repeat_delay = 600;
+	rc.kb_numlock_enable = true;
 	rc.screen_edge_strength = 20;
 
 	rc.snap_edge_range = 1;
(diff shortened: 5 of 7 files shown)


jlindgren90:
@johanmalm or @Consolatis - any suggestions on what the config syntax should look like for numlock? There's no Openbox equivalent to my knowledge (standalone numlockx utility was typically used instead).

stefonarch:
Shouldn't this be in environment instead in rc.xml? Other XKB options are there in labwc.
Sway:
```
input <identifier> xkb_numlock enabled|disabled
           Initially enables or disables NumLock on startup, the default is disabled.
```
`kb_numlock_default_state = true` in wayfire.

jlindgren90:
> Shouldn't this be in environment instead in rc.xml? Other XKB options are there in labwc. Sway:
> 
> ```
> input <identifier> xkb_numlock enabled|disabled
>            Initially enables or disables NumLock on startup, the default is disabled.
> ```
> 
> `kb_numlock_default_state = true` in wayfire.

Those examples are both config file lines, not environment variables.

Consolatis:
If we go the `rc.xml` route some argument for `<keyboard>` seems like the most obvious choice.
So maybe something like
```xml
<keyboard numlock="enabled">
  <default />
</keyboard>

```
or with its alternative config format
```xml
<keyboard>
  <numlock>enabled</numlock>
  <default />
</keyboard>
```

stefonarch:
> Those examples are both config file lines, not environment variables.

There are only  config files in sway/wayfire/hyprland and they contain all  XKB_* settings like keyboard layout and keys to switch it ecc.

My idea was only to keep all XKB_ settings in the same place.

Consolatis:
> My idea was only to keep all XKB_ settings in the same place.

Well, is there some XKB env var that allows enabling numlock?
We don't do anything with the env vars ourselves.

jlindgren90:
@Consolatis if you would like to add the config bits to this branch/PR, please feel free 👍

jlindgren90:
A quick web search didn't turn up any existing XKB environment variable for numlock. If it doesn't exist already, it's not our namespace to define a new one in (it would have to start with LABWC_, not XKB_).

Consolatis:
> @Consolatis if you would like to add the config bits to this branch/PR, please feel free 👍

Alright, will do later.

Some generic notes:
- Should we have things like that (that modify the xkb state) in `src/keyboard.c` rather than in `seat.c`?
- When reading through the diff again it reminds me of an earlier version of #1077 (including the need to update the leds and similar). `wlr_keyboard_notify_modifiers()` should be enough as it updates the xkb state and led stuff itself as far as I remember.

jlindgren90:
I'm not sure if the exact version/commit, but `wlr_keyboard_notify_modifiers()` appears to have been made private in recent wlroots. To use it was my initial thought too, based on the sway code.

I'm fine with moving stuff from keyboard.c to seat.c, please feel free to do that if you like as I may not get back to this until after the weekend.

Consolatis:
> I'm not sure of the exact version/commit, but wlr_keyboard_notify_modifiers() appears to have been made private in recent wlroots. To use it was my initial thought too, based on the sway code.

I didn't find it either at first, its hidden in

> user@odroidc4:~/dev/labwc/subprojects/wlroots$ grep -R -A 2 wlr_keyboard_notify_modifiers include/wlr/
include/wlr/interfaces/wlr_keyboard.h:void wlr_keyboard_notify_modifiers(struct wlr_keyboard *keyboard,
include/wlr/interfaces/wlr_keyboard.h-          uint32_t mods_depressed, uint32_t mods_latched, uint32_t mods_locked,
include/wlr/interfaces/wlr_keyboard.h-          uint32_t group);

stefonarch:
> A quick web search didn't turn up any existing XKB environment variable for numlock

Yes, it looks like there isn't any existing.

jlindgren90:
> I didn't find it either at first, its hidden in
> 
> > user@odroidc4:~/dev/labwc/subprojects/wlroots$ grep -R wlr_keyboard_notify_modifiers include/wlr/
> > include/wlr/interfaces/wlr_keyboard.h:void wlr_keyboard_notify_modifiers(struct wlr_keyboard *keyboard,

Okay, if it's available, let's use it 👍

Consolatis:
Pushed the commit, feel free to squash and change as you see fit (e.g. remove the commit from the PR). I also added some docs and added an early return if the numlock index couldn't be found in the given keymap.
I could not really test it as it doesn't seem to work with a nested instance and I hardly ever restart my DRM session.

Just remembered another issue I had with the keyboard-layout-per-window PR in the beginning: when manually updating the xkb state and `kb->modifiers` it wouldn't trigger the `keyboard->events.modifiers` signal as far as I remember. Thus the keyboard group wouldn't react to it via `handle_keyboard_modifiers()` which usually syncs the modifier states across all members of the group. So that was another reason to use `wlr_keyboard_notify_modifiers()`.

johanmalm:
Great effort. Happy with all.

`<keyboard><numlock>` is good choice. And, yeah I increasingly prefer nested elements rather than attributes (like you've done).

Agree. Better to move to keyboard.c.

Thumbs up for `wlr_keyboard_notify_modifiers()`. Suggest adding comment by it to remind us in future. Maybe something like:

```
/*
 * This updates the xkb-state + kb->modifiers and also triggers the
 * keyboard->events.modifiers signal to sync the modifier states across
 * the keyboard-group.
 */
```

jlindgren90:
v2:
- Squash commit by @Consolatis to add config option; add Co-authored-by
- Use `wlr_keyboard_notify_modifiers()`
- Move `set_numlock()` to keyboard.c as `keyboard_set_numlock()`

I think triggering the `keyboard->events.modifiers` signal doesn't actually matter in this case, since we are modifying the modifiers (haha) before calling `wlr_keyboard_group_add_keyboard()` and as a result, nothing is listening to the signal yet. But it's still good for code re-use and future-proofing to use the wlroots function rather than re-implementing it as I did originally. Thanks for all the help and feedback!

Consolatis:
Comment as suggest by malm is missing (even if the signal doesn't matter in this case). I think its still a good comment so people reading the sources in the future know what the function does and that it can be used for similar things. Other than that this LVGTM.

jlindgren90:
How about this?

"This updates the xkb-state + kb->modifiers and also triggers the keyboard->events.modifiers signal (the signal has no effect in current labwc usage since the keyboard is not part of a keyboard-group yet)."

Consolatis:
Sounds fine.

johanmalm:
Thanks.
Have just added the 'agreed comment'.

johanmalm:
Will merge shortly.

Issue #1129, opened 2023-10-01
Keybinds lost and not recoverable


stefonarch:
It happened now for the third time but unfortunately I have no idea what triggers it. I've only lxqt-panel and virtualbox with `toggleKeyBinds` set by window rules and no shortcut for it. 

It happened never before until maybe a week ago and I've to restart the session because reconfigure won't help. Log is [here](https://pastebin.com/u1wSdwFY).

Consolatis:
I don't completely understand the issue. You have a window rules which executes `ToggleKeybinds` for lxqt-panel and virtualbox and can't escape it? Not even when focusing or opening another window?

In any case if you configure a keybind for `ToggleKeybinds` you should always be able to disable it for the currently focused window.

stefonarch:
> You have window rules which execute ToggleKeybinds for lxqt-panel and virtualbox and can't escape it?

Those rules work fine if focus is on there. I mentioned it because it's the only setting where toggleKeybinds is involved. I'm not even sure if it's about toggleKeybinds - they just stopped to work altogether and I've no idea why. Virtualbox was never involved.

johanmalm:
@stefonarch Are you able to consistently re-produce it? If so, can you give me the steps to re-produce.
Does it happen if you run with `labwc -C /dev/null'?

stefonarch:
As I said: I've no idea how to trigger it, can only wait and look if it happens again and maybe then I'll find the steps.

johanmalm:
> As I said: I've no idea how to trigger it, can only wait and look if it happens again and maybe then I'll find the steps.

Okay. Do you restart your computer every day or are you on a long-running (old) tty session? I'm just thinking that we've made a few changes to keyboard handling recently, so if you're on a long running instance, it might be worth re-starting.

stefonarch:
Usually I've long uptimes, until a week or more, only suspending to RAM. After updating labwc I restart the session though.

stefonarch:
It happened again, switching layout, and I can it reproduce always now which is real nasty.
Can reproduce it with default rc.xml and
```
cat .config/labwc/environment 
XKB_DEFAULT_LAYOUT=it,ch
XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle,grp_led:scroll
XCURSOR_THEME=Breeze_Snow
XCURSOR_SIZE=28
MOZ_ENABLE_WAYLAND=1
```

stefonarch:
> Downgrade didn't resolve too.

[...]

Issue #1061, opened 2023-08-30
Limited modifiers compared to X?


alaricljs:
I've been waiting for an Openbox-adjacent to arrive for Wayland so I could be at least somewhat motivated to give the new thing a try.  Unfortunately I tripped up right at the beginning.

I have 5 modifiers in common use under Openbox - Shift, Ctrl, Alt, Super, Meta.  There's still 2 more I could add in, but only 1 cool name (Hyper) and I haven't quite found a good reason to.  I really do not want to lose a modifier though.

Is the modifier limit intrinsic to Wayland or is it something that labwc could handle?

Consolatis:
> Is the modifier limit intrinsic to Wayland or is it something that labwc could handle?

I am not 100% sure but I think its a labwc limitation. Wayland applications (and labwc itself) use libxkb for the keymaps so having more modifiers should be possible. Its not exactly my area of expertise though. You could give `xev` (xwayland) and `wev` (wayland native) a try to debug the modifiers that applications receive.

These are the modifiers that wlroots (which we depend on) supports:
```
 WLR_MODIFIER_SHIFT = 1 << 0,
 WLR_MODIFIER_CAPS = 1 << 1,
 WLR_MODIFIER_CTRL = 1 << 2,
 WLR_MODIFIER_ALT = 1 << 3,
 WLR_MODIFIER_MOD2 = 1 << 4,
 WLR_MODIFIER_MOD3 = 1 << 5,
 WLR_MODIFIER_LOGO = 1 << 6,
 WLR_MODIFIER_MOD5 = 1 << 7
```

Labwc however only supports Shift, Control, Alt, and W / logo for keybinds.

jonhiggs:
I'm also short a modifier. I would love to see support for mod3 and/or mod5.

Consolatis:
Just had a look at [Openbox](http://openbox.org/wiki/Help:Bindings#Syntax) and they seem to support `M` and `H` for Meta (`Mod5` apparently) / Hyper (`Mod3` apparently). Additionally they support `Mod[1-5]` so `Mod1-f` would be the same as `A-f` and `Mod4-f` the same as `W-f`.

Consolatis:
`M` and `H` should now work in latest master. Feel free to comment if there are any issues with it.

jonhiggs:
[...]

PR #1145, opened 2023-10-08
keybinds: add support for Meta and Hyper modifiers

Consolatis:
<b>keybinds: add support for Meta and Hyper modifiers</b>

The modifiers can be used in keybinds via `M-key` and `H-key`

Additionally adds support for:
- `Mod1` (same as `A`)
- `Mod3` (same as `H`)
- `Mod4` (same as `W`)
- `Mod5` (same as `M`)

This is compatible with the format used by Openbox.
(http://openbox.org/wiki/Help:Bindings#Syntax)

`Mod2` (`NumLock`) and `Caps` are still not supported due
to their locking behavior but could theoretically be added.

Fixes:
- #1061

This patch does not support the same changes for mousebinds.

----

<b>keyboard: prevent storing synthetic layout change key event</b>

Before this patch, we were storing the key in our pressed set
and didn't remove it which in turn caused all further keybinds
to fail. The behavior was dependent on the exact flow of press
and release events and was most obvious when using
`XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle`

We now simply treat it as a modifier and thus do not store it
in the pressed set.

Fixes:
- #1129

### DIFF
--- a/src/config/keybind.c
+++ b/src/config/keybind.c
@@ -15,14 +15,19 @@
 uint32_t
 parse_modifier(const char *symname)
 {
+	/* Mod2 == NumLock */
 	if (!strcmp(symname, "S")) {
 		return WLR_MODIFIER_SHIFT;
 	} else if (!strcmp(symname, "C")) {
 		return WLR_MODIFIER_CTRL;
-	} else if (!strcmp(symname, "A")) {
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM.
Simply excellent.
Not tested.

PR #1146, opened 2023-10-09
view: fix some inconsistencies in view_ functions

jlindgren90:
... especially regarding whether a `(view *)` parameter may be `NULL`. It's confusing when some functions accept `NULL` and others don't, and could trip someone up.

I'm partly to blame for the inconsistency, since (if memory serves) I added `view_is_tiled()` and `view_is_floating()`, which do accept `NULL`.

In detail:

- Make `view_is_tiled()` and `view_is_floating()` no longer accept `NULL`.
- Rename `view_isfocusable` -> `view_is_focusable` for consistency with other `view_is_` functions.
- Eliminate `view_inhibits_keybinds()` as it only existed to safely accept `NULL` and check a single flag, which can be checked directly.
- Add `assert(view)` to remaining public `view_` functions to catch accidentally passing `NULL`.
- Inline `inhibit_keybinds()` into `view_toggle_keybinds()`. It is closely related and not called from anywhere else; inlining it allows eliminating an extra `assert()` which is now impossible.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -271,7 +271,7 @@ void view_array_append(struct server *server, struct wl_array *views,
 	enum lab_view_criteria criteria);
 
 /**
- * view_isfocusable() - Check whether or not a view can be focused
+ * view_is_focusable() - Check whether or not a view can be focused
  * @view: view to be checked
  *
  * The purpose of this test is to filter out views (generally Xwayland) which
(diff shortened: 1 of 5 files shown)


Consolatis:
LGTM

johanmalm:
Thanks.

PR #1149, opened 2023-10-09
menu: use wl_list instead of array

johanmalm:
...to make it easier to split menu.c into smaller chunks without exposing nr_menus variable.


Consolatis:
LGTM

PR #1152, opened 2023-10-10
keyboard: add missing Hyper_ and Meta_ syms to modifier detection

Consolatis:
This was forgotten in 65bd32d62500389a1d6affafac6d1989ba28042f
Reported-by: @jonhiggs (thanks)

Also stop treating the synthetic layout change sym as modifier
but still prevent it from being added to the set of pressed keys.

Additionally slightly reformat the code.


johanmalm:
Thanks.

Issue #1153, opened 2023-10-10
wshowkeys utility crashes on labwc


tranzystorekk:
https://github.com/ammgws/wshowkeys experiences a crash after a short time of not typing anything, seemingly only on labwc (doesn't reproduce on river, sway)

repro demonstration:

![H48_](https://github.com/labwc/labwc/assets/5671049/0b9fe0b0-766e-4e76-b2cb-f25e6c443076)
 
environment info:

wshowkeys 0.1
Void Linux
labwc 0.6.5

some debug logs collected from IRC:

- https://dpaste.com/HRTGZDSNB
- https://dpaste.com/2RMCFQJKU
- https://dpaste.com/CDVGA4W2B

Consolatis:
So far it looks like something related to our unmap handler for layershell surfaces.

<details><summary>Complete log as dpaste stuff expires after some time</summary>

```
[1727282.411]  -> wl_display@1.get_registry(new id wl_registry@2)
[1727282.459]  -> wl_display@1.sync(new id wl_callback@3)
[1727282.636] wl_display@1.delete_id(3)
[1727282.657] wl_registry@2.global(1, "wl_shm", 1)
[1727282.665]  -> wl_registry@2.bind(1, "wl_shm", 1, new id [unknown]@4)
[1727282.671] wl_registry@2.global(2, "wl_drm", 2)
[1727282.676] wl_registry@2.global(3, "zwp_linux_dmabuf_v1", 4)
[1727282.681] wl_registry@2.global(4, "zxdg_output_manager_v1", 3)
[1727282.686]  -> wl_registry@2.bind(4, "zxdg_output_manager_v1", 1, new id [unknown]@5)
[1727282.691] wl_registry@2.global(5, "zwlr_output_manager_v1", 4)
[1727282.696] wl_registry@2.global(6, "wl_compositor", 5)
[1727282.701]  -> wl_registry@2.bind(6, "wl_compositor", 4, new id [unknown]@6)
[1727282.705] wl_registry@2.global(7, "wl_subcompositor", 1)
[1727282.710] wl_registry@2.global(8, "wl_data_device_manager", 3)
[1727282.714] wl_registry@2.global(9, "zwp_primary_selection_device_manager_v1", 1)
[1727282.718] wl_registry@2.global(10, "wl_seat", 8)
[1727282.723]  -> wl_registry@2.bind(10, "wl_seat", 5, new id [unknown]@7)
[...]

PR #1154, opened 2023-10-10
layers: arrange layers on destroy

johanmalm:
...rather than unmap because that is the approach of sway and it also avoids wshowkeys crashing.

Fixes: #1153

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -43,6 +43,13 @@ arrange_one_layer(const struct wlr_box *full_area, struct wlr_box *usable_area,
 {
 	struct wlr_scene_node *node;
 	wl_list_for_each(node, &tree->children, link) {
+		/*
+		 * Could be null during destruction when called from
+		 * handle_node_destroy()
+		 */
+		if (!node->data) {
+			continue;
+		}
 		struct lab_layer_surface *surface = node_layer_surface_from_node(node);
 		struct wlr_scene_layer_surface_v1 *scene = surface->scene_layer_surface;
 		if (!!scene->layer_surface->current.exclusive_zone != exclusive) {
@@ -101,6 +108,7 @@ handle_output_destroy(struct wl_listener *listener, void *data)
 	struct lab_layer_surface *layer =
 		wl_container_of(listener, layer, output_destroy);
 	layer->scene_layer_surface->layer_surface->output = NULL;
+	layer->output = NULL;
 	wlr_layer_surface_v1_destroy(layer->scene_layer_surface->layer_surface);
 }
 
@@ -196,6 +204,16 @@ handle_node_destroy(struct wl_listener *listener, void *data)
 	 * focus to.
 	 */
 
+	/*
+	 * Nullify node descriptor early to avoid functions such as
+	 * arrange_one_layer() trying to dereference it.
+	 */
+	layer->tree->node.data = NULL;
+
+	if (layer->output) {
+		output_update_usable_area(layer->output);
+	}
+
 	wl_list_remove(&layer->map.link);
 	wl_list_remove(&layer->unmap.link);
 	wl_list_remove(&layer->surface_commit.link);
@@ -211,9 +229,6 @@ handle_unmap(struct wl_listener *listener, void *data)
 	struct lab_layer_surface *layer = wl_container_of(listener, layer, unmap);
 	struct wlr_layer_surface_v1 *layer_surface =
 		layer->scene_layer_surface->layer_surface;
-	if (layer_surface->output) {
-		output_update_usable_area(layer_surface->output->data);
-	}
 	struct seat *seat = &layer->server->seat;
 	if (seat->focused_layer == layer_surface) {
 		seat_set_focus_layer(seat, NULL);
@@ -376,6 +391,7 @@ handle_new_layer_surface(struct wl_listener *listener, void *data)
 	struct lab_layer_surface *surface = znew(*surface);
 
 	struct output *output = layer_surface->output->data;
+	surface->output = output;
 
 	struct wlr_scene_tree *selected_layer =
 		output->layer_tree[layer_surface->current.layer];
@@ -387,6 +403,7 @@ handle_new_layer_surface(struct wl_listener *listener, void *data)
 		wlr_log(WLR_ERROR, "could not create layer surface");
 		return;
 	}
+	surface->tree = surface->scene_layer_surface->tree;
 
 	node_descriptor_create(&surface->scene_layer_surface->tree->node,
 		LAB_NODE_DESC_LAYER_SURFACE, surface);
--- a/include/layers.h
+++ b/include/layers.h
@@ -10,6 +10,8 @@ struct output;
 struct lab_layer_surface {
 	struct wlr_scene_layer_surface_v1 *scene_layer_surface;
 	struct server *server;
+	struct output *output;
+	struct wlr_scene_tree *tree;
 
 	bool mapped;
 
--- a/src/node.c
+++ b/src/node.c
@@ -11,7 +11,7 @@ descriptor_destroy(struct node_descriptor *node_descriptor)
 		return;
 	}
 	wl_list_remove(&node_descriptor->destroy.link);
-	free(node_descriptor);
+	zfree(node_descriptor);
 }
 
 static void


johanmalm:
Had to add 'tree' and 'output' to 'lab_layer_surface' because `wlr_scene_layer_surface_v1` is gone by the time we get to the node destroy-handler.
Might be able to refactor a few other lines in `layers.c` as a result, but suggest that's for a different PR.

johanmalm:
I could reproduce the describe issue and this patch fixes it for me, but @tranzystorek-io any testing you can do would be good.

I haven't looked at the `wshowkeys` and don't actually understand why it happened.

Consolatis:
I am not convinced that doing `output_update_usable_area()` on node destroy is the right thing to do. If the node itself is destroyed or not shouldn't matter for the usable area at all. However, being unmapped / disabled matters.

Take the following with a grain of salt as I only looked at the code but didn't actually verify:
I think the issue here is that we call `output_update_usable_area()` twice on unmap:
- in the last branch of `handle_surface_commit()`
- and in the unmap handler itself

And both then consider the freshly unmapped surface for the calculations and send a `wlr_scene_layer_surface_v1_configure()` to the layer surface.

So what I would propose instead is:
- keep one of them do the `output_update_usable_area()` call
  - Variant 1: the unmap handler
    - guard against doing the second call in the commit handler by additionally checking for `&& layer->mapped` (so it still works for freshly mapped surfaces)
    - ensure the unmap handler sets `layer->mapped = false` before calling `output_update_usable_area()`
  - Variant 2: the commit handler
    - just remove the `output_update_usable_area()` call
- in `arrange_one_layer()` instead of the new `!node->data` check, check against `if (!layer->mapped) { continue; }`

johanmalm:
Thanks for feedback/review.

I've had a play with `wshowkeys` and can patch it to avoid the protocol error (but it causes a few other problems, so can't got anything I can submit as a fix).

>         * guard against doing the second call in the commit handler by additionally checking for `layer->mapped` before calling `output_update_usable_area()` (so it still works for freshly mapped surfaces)

This results in nothing visible on the screen when typing with `wshowkeys`.

```
diff --git a/src/layers.c b/src/layers.c
index b6ae4bb..1387af5 100644
--- a/src/layers.c
+++ b/src/layers.c
@@ -175,8 +175,10 @@ handle_surface_commit(struct wl_listener *listener, void *data)
 	}
 
 	if (committed || layer->mapped != layer_surface->mapped) {
+		if (layer->mapped) {
+			output_update_usable_area(output);
+		}
 		layer->mapped = layer_surface->mapped;
-		output_update_usable_area(output);
 		/*
 		 * Update cursor focus here to ensure we
 		 * enter a new/moved/resized layer surface.
```

>         * just remove the `output_update_usable_area()` call from the unmap handler (+ maybe add a comment referring to the commit handler)

`wshowkeys` works fine with this, but if I run something like `waybar` it doesn't arrange views on unmap/destroy, so for example if there is a maximized window, it doesn't resize to fill the gap when `waybar` disappears.

>     * while at it, the `output` arg for `arrange_one_layer()` isn't used at all and could be removed. Likely a left-over from an earlier refactor

Oh, nice spot! Will fix in separate PR.

johanmalm:
This commit it fixes it in `wshowkeys`: https://github.com/johanmalm/wshowkeys/commit/a4873fe79a40fb5a1d81b7130270b3c9991095b7

I don't understand why. Any theories / ideas?

Consolatis:
Hm.. When looking at the whole thing it appears that if height or width is `0` (e.g. no `state->keys` a.k.a. please-hide-me-now) wshowkeys attaches a `NULL` buffer and commits the surface which triggers a layershell configure event from our side which supplies the new sizes. Those are then stored in `state->width` and `state->height`. The size we return here is not `0` though but the previous value. That in turn causes the following `set_dirty()` call to end up in the same branch (via `width / scale != state->width`) and basically looping endlessly between these two states. The TODO comment about "infinite loop" above the commit() call might be very related here.

Now, if we should return `0` sizes here instead or not when the buffer was set to `NULL`: I am not sure. We should likely look that up in the layershell protocol. Why we complain about the configure serial being wrong.. I am not sure either. Based on the logs linked in the issue the serials used for the ACKs are correct and match our `configure()` ones.

This might actually hide some other bug and could be a pure coincidence the we ended up there with wlroots complaining about a factually correct serial (at least based on the logs).

alex-huff:
Also fixes #2748.

johanmalm:
Does anyone fancy doing some thinking on this one? I spent quite a bit of time on it over a year ago, but got stuck. I now feel a slight allergic reaction to it :monkey:

alex-huff:
I'll do some digging when I get a chance.

alex-huff:
See: https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1

Specifically:
```
Attaching a null buffer to a layer surface unmaps it.

Unmapping a layer_surface means that the surface cannot be shown by the compositor until it is explicitly mapped again. The layer_surface returns to the state it had right after layer_shell.get_layer_surface. The client can re-map the surface by performing a commit without any buffer attached, waiting for a configure event and handling it as usual.
```

When hiding (unmapping) a kitty panel it crashes with this error:
```
zwlr_layer_surface_v1#31: error 0: wrong configure serial: 2189
[2.405] [glfw error 65544]: Wayland: fatal display error: Protocol error
```

> Based on the logs linked in the issue the serials used for the ACKs are correct and match our `configure()` ones.

wlroots posts a display error for invalid serial even though looking at the logs wlshowkeys/kitty appear to use the correct serial.

See wlroots handling of ack configure:
```c
static void layer_surface_handle_ack_configure(struct wl_client *client,
		struct wl_resource *resource, uint32_t serial) {
        // beginning of function omitted

	// First find the ack'ed configure
	bool found = false;
	struct wlr_layer_surface_v1_configure *configure, *tmp;
	wl_list_for_each(configure, &surface->configure_list, link) {
		if (configure->serial == serial) {
			found = true;
			break;
		}
	}
	if (!found) {
		wl_resource_post_error(resource,
			ZWLR_LAYER_SURFACE_V1_ERROR_INVALID_SURFACE_STATE,
			"wrong configure serial: %" PRIu32, serial);
		return;
	}
        // rest of function omitted
}
```

Notice the serial is checking against the configures in `configure_list`. The problem is `configure_list` gets cleared in `layer_surface_reset` which gets called in `layer_surface_role_commit` when the surface is unmapped.

> Now, if we should return `0` sizes here instead or not when the buffer was set to `NULL`: I am not sure. We should likely look that up in the layershell protocol.

Given the above information, labwc should never send a configure after a client unmaps itself (commits with a NULL buffer), until the client makes another commit signifying it wants to be shown again.

Luckily wlroots provides labwc with the information needed to determine if a surface is ready to receive configure events. This is what the `initialized` field of the `wlr_layer_surface_v1` struct is for. See `layer_surface_role_commit` for when it is updated.

However, checking `initialized` is not valid in `handle_unmap` because the unmap event is signaled before `layer_surface_role_commit` actually updates `initialized`. :(

> I think the issue here is that we call `output_update_usable_area()` twice on unmap:
> 
> * in the last branch of `handle_surface_commit()`
> * and in the unmap handler itself

You are right in that `output_update_usable_area` gets called twice making `output_update_usable_area` unnecessary in the unmap handler. The current logic in `handle_surface_commit` is sufficient for handling the rearranging of surfaces after they are unmapped and since `handle_surface_commit` is called after `layer_surface_role_commit`, `initialized` will be properly set.

However... When a layer surface is destroyed the commit handler doesn't get called and so the surfaces are never rearranged. I believe this is why sway also rearranges everything in the node destroy handler.

src/layers.c
```diff
@@ -43,6 +43,9 @@ arrange_one_layer(const struct wlr_box *full_area, struct wlr_box *usable_area,
 {
 	struct wlr_scene_node *node;
 	wl_list_for_each(node, &tree->children, link) {
+		if (!node->data) {
```
tokyo4j, line 50: nit: I suggest adding a comment like `/* when called form handle_node_destroy() */`.

tokyo4j:
LGTM other than my nitpicking suggestion above.

I understood that there's no clean and reliable way to tell whether a layer surface is being unmapped. The following code will work, but it's just a hack and I prefer the approach in this PR.

```diff
diff --git a/src/layers.c b/src/layers.c
index 9583b7f7..71516cc7 100644
--- a/src/layers.c
+++ b/src/layers.c
@@ -50,6 +50,9 @@ arrange_one_layer(const struct wlr_box *full_area, struct wlr_box *usable_area,
 		if (!scene->layer_surface->initialized) {
 			continue;
 		}
+		if (scene->layer_surface->surface->WLR_PRIVATE.unmap_commit) {
+			continue;
+		}
 		if (!!scene->layer_surface->current.exclusive_zone != exclusive) {
 			continue;
 		}
```

johanmalm:
Thanks for review.

Good idea to add comment about "null during destruction". Done.

Consolatis:
@alex-huff Thanks for your in-depth investigation!

alex-huff:
> @alex-huff Thanks for your in-depth investigation!

You are welcome.

PR #1155, opened 2023-10-11
output: fix invisible cursor after wlopm --off && wlopm --on

jlindgren90:
Use the same fix/workaround as in `output_update_for_layout_change()` to make sure that the cursor is also visible after (re-)enabling an output in `handle_output_power_manager_set_mode()`.


jlindgren90:
Hmm, this only seems to help if the cursor is over an XWayland window. If it's over the desktop, it still remains invisible. Not sure why.

Consolatis:
Thanks

PR #1163, opened 2023-10-14
seat: ignore focus change to unmanaged surface belonging to same PID

### LINKED ISSUE #1161: Context menus of X-applications immediately close upon opening
An observation I made using latest master of labwc:
all the context menus of X applications (LibreOffice, IntelliJ) immediately close upon opening, barely showing for a single frame if at all.
Reverting back to labwc 0.6.5 does not exhibit this effect, so this is the version I am currently using.

jlindgren90:
If an xwayland-unmanaged surface was focused belonging to the same application as the focused view, allow the view to remain active. This fixes an issue with menus immediately closing in some X11 apps (try LibreOffice with SAL_USE_VCLPLUGIN=gen).

Fixes: 4028a9482f9399e3c3587f5c6eca6c0b128c9afc
("seat: use focus_change event to update focused/active view")

Fixes #1161.

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -13,6 +13,10 @@
 #include "labwc.h"
 #include "view.h"
 
+#if HAVE_XWAYLAND
+#include <wlr/xwayland.h>
+#endif
+
 static void
 input_device_destroy(struct wl_listener *listener, void *data)
 {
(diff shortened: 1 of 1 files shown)


Consolatis:
Very nice, thanks!
Due to fixing the regression (not tested myself but looks sensible and makes sense) we should merge this.

However, I really don't like us sprinkling xwayland specific code ( `HAVE_XWAYLAND`) everywhere so ideas how to better encapsulate that are very welcome,

johanmalm:
LGTM.

IIRC sway keep track of the PID in the view struct, but I think your approach is more self-contained.

jlindgren90:
> However, I really don't like us sprinkling xwayland specific code ( HAVE_XWAYLAND) everywhere so ideas how to better encapsulate that are very welcome

Unfortunately it seems that XWayland requires specific behavioral logic in multiple areas (focus, window positioning, etc.) Maybe we could move portions of that logic to xwayland.c but then we have the opposite problem where that one source file touches all those different logical areas.

jlindgren90:
Going ahead with merging this based on the above reviews, thanks!

jlindgren90:
@Flrian Can you open a new issue for that please?

jlindgren90:
Just an (educated) guess but the JDownloader issue is probably related to #1103 rather than this PR.

Consolatis:
> > However, I really don't like us sprinkling xwayland specific code ( HAVE_XWAYLAND) everywhere so ideas how to better encapsulate that are very welcome
> 
> Unfortunately it seems that XWayland requires specific behavioral logic in multiple areas (focus, window positioning, etc.) Maybe we could move portions of that logic to xwayland.c but then we have the opposite problem where that one source file touches all those different logical areas.

I'd prefer that variant. In some cases it might also be possible to add more generic callbacks to `view->impl`.
Something like
```c
[..]
struct view *view = event->new_surface ?
		view_from_wlr_surface(event->new_surface) : NULL;
if (view && view->type == LAB_XWAYLAND_VIEW) {
	if (view->impl->is_related(event->new_surface)) {
		return;
	}
}
```

johanmalm:
@jlindgren90 Really appreciate you working/responding so quickly on this. And I think that the whole move to processing the 'focus_change' signal is the right one and has flushed out some hard-to-find inconsistencies.

Have just made it to the computer... and just an after thought on this:

In `focus_change_notify()` - should we actually respond to any focus_change associated with an `xwayland-unmanaged` surface?
This is all quite hard stuff to analyze but I think I've concluded before that what `wlroots` does in that respect is enough.
Also, if we get a `focus_change` signal from an `unmanaged` surface with a different PID, would we actually want to honor that?

Sorry to throw in an after-thought.

jlindgren90:
Good questions. I think the `focus_change` signal handler is too late to decide whether to allow/deny a focus/activate request since the focus has already changed at this point. And wlroots does have some checks to make sure that spontaneous focus-in events from another PID are denied.

So maybe the PID check in focus_change is redundant and it will always be the same PID? I'm not sure. I don't think it does any harm even if it is redundant.

jlindgren90:
@Consolatis I like that `view->impl->is_related` idea btw.

PR #1164, opened 2023-10-14
Add MoveToCursor action.

Arnaudv6:
Not tested on multiple screens setup, nor with virtual workspaces.  
(Though this should just work (TM)).

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -54,6 +54,10 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="MoveTo" x="" y="" />*
 	Move to position (x, y)
 
+*<action name="MoveToCursor" />*
+	Move to be centered on cursor.
+	Tries to prevent any part of the window from going off-screen.
+
 *<action name="MoveRelative" x="" y="" />*
 	Move window relative to its current position. Positive value of x moves
 	it right, negative left. Positive value of y moves it down, negative up.
--- a/src/action.c
+++ b/src/action.c
@@ -84,6 +84,7 @@ enum action_type {
 	ACTION_TYPE_RESIZE,
 	ACTION_TYPE_RESIZE_RELATIVE,
 	ACTION_TYPE_MOVETO,
+	ACTION_TYPE_MOVETO_CURSOR,
 	ACTION_TYPE_MOVE_RELATIVE,
 	ACTION_TYPE_SEND_TO_DESKTOP,
 	ACTION_TYPE_GO_TO_DESKTOP,
@@ -122,6 +123,7 @@ const char *action_names[] = {
 	"Resize",
 	"ResizeRelative",
 	"MoveTo",
+	"MoveToCursor",
 	"MoveRelative",
 	"SendToDesktop",
 	"GoToDesktop",
@@ -744,6 +746,11 @@ actions_run(struct view *activator, struct server *server,
 				int x = action_get_int(action, "x", 0);
 				int y = action_get_int(action, "y", 0);
 				view_move_relative(view, x, y);
+			}
+			break;
+		case ACTION_TYPE_MOVETO_CURSOR:
+			if (view) {
+				view_move_to_cursor(view);
 			}
 			break;
 		case ACTION_TYPE_SEND_TO_DESKTOP:
--- a/src/view.c
+++ b/src/view.c
@@ -351,6 +351,50 @@ view_move_relative(struct view *view, int x, int y)
 	view_move(view, view->pending.x + x, view->pending.y + y);
 }
 
+void
+view_move_to_cursor(struct view *view)
+{
+	assert(view);
+
+	struct output *pending_output = output_nearest_to_cursor(view->server);
+	if (!output_is_usable(pending_output)) {
+		return;
+	}
+	if (view->fullscreen) {
+		view_set_fullscreen(view, false);
+	}
+	if (view->maximized) {
+		view_maximize(view, false, /*store_natural_geometry*/ false);
+	}
+	if (view_is_tiled(view)) {
+		view_set_untiled(view);
+		view_restore_to(view, view->natural_geometry);
+	}
+
+	struct border margin = ssd_thickness(view);
+	struct wlr_box geo = view->pending;
+	geo.width += margin.left + margin.right;
+	geo.height += margin.top + margin.bottom;
+
+	int x = view->server->seat.cursor->x - (geo.width / 2);
+	int y = view->server->seat.cursor->y - (geo.height / 2);
+
+	struct wlr_box usable = output_usable_area_in_layout_coords(pending_output);
+	if (x + geo.width > usable.x + usable.width) {
+		x = usable.x + usable.width - geo.width;
+	}
+	x = MAX(x, usable.x);
+
+	if (y + geo.height > usable.y + usable.height) {
+		y = usable.y + usable.height - geo.height;
+	}
+	y = MAX(y, usable.y);
+
+	x += margin.left;
+	y += margin.top;
+	view_move(view, x, y);
+}
+
 struct view_size_hints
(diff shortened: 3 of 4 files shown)


src/action.c
```diff
@@ -746,6 +748,33 @@ actions_run(struct view *activator, struct server *server,
...
+					> pending_output->usable_area.y
+					+ pending_output->usable_area.height) {
+				y = pending_output->usable_area.y
+					+ pending_output->usable_area.height - view->pending.height;
+			}
+			if (y < 0) {
+				y = 0;
+			}
+			view_move(view, x, y);
+			/* Falls through to GoToDesktop */
```
Consolatis, line 777: This should end in a `break`

src/action.c
```diff
@@ -746,6 +748,33 @@ actions_run(struct view *activator, struct server *server,
...
+			struct output *pending_output = output_nearest_to_cursor(server);
+			if (x + view->pending.width
+					> pending_output->usable_area.x
+					+ pending_output->usable_area.width) {
+				x = pending_output->usable_area.x
+					+ pending_output->usable_area.width - view->pending.width;
+			}
+			if (x < 0) {
+				x = 0;
+			}
```
Consolatis, line 766: This could be `x = MAX(x, 0)`, same for the `y` case below.
Both cases should also use the x/y coordinates of the usable area instead of `0` (in case there is some panel docked). E.g.

`x = MAX(x, pending->output->usable_area.x);` in this case.

docs/labwc-actions.5.scd
```diff
@@ -54,6 +54,10 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="MoveTo" x="" y="" />*
 	Move to position (x, y)
 
+*<action name="MoveToCursor" />*
+	Move to be centered on cursor.
+	While preventing any part off the window from going off-screen.
```
Consolatis, line 59: If the window is larger than the output in either dimension it will still be partly off screen,
so I'd suggest something like `Tries to prevent any part of the the window to go off-screen`.

Consolatis (changes requested):
Thanks, overall this looks good, some small remarks below.

A more generic remark: usually we have functions like this in `src/view.c` and then call it from `src/action.c` to keep that giant switch statement as small as possible. So I'd suggest to move the new code into a function like `view_move_to_cursor()` in `src/view.c`, add the signature to the `include/view.h` header and then call it from `action.c`.

Regarding the feature itself: I understand the benefit of an action like this. As it is not listed at http://openbox.org/wiki/Help:Actions and thus also not in https://github.com/labwc/labwc-scope#readme this requires approval of @johanmalm.

Arnaudv6:
Right, thanks, can't do that now, but will do ASAP.

Arnaudv6:
Thanks for you detailed review, @Consolatis. All done. I can rebase if you like so.
Do boundaries-checks belong to a separate function, to reuse when we implement [placement](https://github.com/labwc/labwc-scope#23-configuration---placement)?

> Regarding the feature itself: I understand the benefit of an action like this. As it is not listed at http://openbox.org/wiki/Help:Actions and thus also not in https://github.com/labwc/labwc-scope#readme this requires approval of @johanmalm.

Agreed. I think that within Xorg, applications themselves made such requests. So openbox needed not do it.

SideNote: BSD CI is stuck in bootloop.

Arnaudv6:
I can't seem to block CI

johanmalm:
I’m happy to include the feature. Will update scope doc when merged.
I’ve not reviewed code. Let me know if you want me to.

Arnaudv6:
@johanmalm, @Consolatis : help, CI runs in loop, and might consume all free tier CPU I'm afraid ?!

Consolatis:
> @johanmalm, @Consolatis : help, CI runs in loop, and might consume all free tier CPU I'm afraid ?!

No worries, AFAIR public orgas basically have unlimited CI.
(Also the FreeBSD CI is really hacky, it uses a mac runner to run virtualbox to run FreeBSD. Wouldn't be needed when GH would just provide FreeBSD jails.)

I've [limited](https://github.com/labwc/labwc/pull/1176) the job runtime to 20 minutes for now.

src/view.c
```diff
@@ -351,6 +351,29 @@ view_move_relative(struct view *view, int x, int y)
 	view_move(view, view->pending.x + x, view->pending.y + y);
 }
 
+void
+view_move_to_cursor(struct view *view)
+{
+	struct output *pending_output = output_nearest_to_cursor(view->server);
```
Consolatis, line 357: `output_nearest_to_cursor()` can return `NULL`.
So I suggest a simple
```c
if (!pending_output) {
	return;
}
```

Arnaudv6: Nice catch, again

Consolatis:
Just tested the PR and saw that we don't adjust for the decorations. E.g. the top decorations are hidden when we run into the `MAX(y, usable_area.y)` case. We also don't handle maximized / snapped or tiled windows well. If you are fine with that I'll adjust the patch a bit and force push to this PR.

Arnaudv6:
I don't mind: thanks for your dedication on this.

Consolatis:
> Not tested on multiple screens setup,

Same. Anybody up for testing this new action on a multi screen setup?
I tried to test this with two virtual outputs on a nested instance but that didn't really work and is likely related to how the cursor input is mapped there.

Consolatis (approved):

Arnaudv6:
Thanks Consolatis!

johanmalm:
> > Not tested on multiple screens setup,
> 
> Same. Anybody up for testing this new action on a multi screen setup?

It doesn't work on multiple outputs.

With the patch below it does:

```
index 57b70a7..6aca923 100644
--- a/src/view.c
+++ b/src/view.c
@@ -379,17 +379,16 @@ view_move_to_cursor(struct view *view)
 	int x = view->server->seat.cursor->x - (geo.width / 2);
 	int y = view->server->seat.cursor->y - (geo.height / 2);
 
-	if (x + geo.width > pending_output->usable_area.x + pending_output->usable_area.width) {
-		x = pending_output->usable_area.x
-			+ pending_output->usable_area.width - geo.width;
+	struct wlr_box usable = output_usable_area_in_layout_coords(pending_output);
+	if (x + geo.width > usable.x + usable.width) {
+		x = usable.x + usable.width - geo.width;
 	}
-	x = MAX(x, pending_output->usable_area.x);
+	x = MAX(x, usable.x);
 
-	if (y + geo.height > pending_output->usable_area.y + pending_output->usable_area.height) {
-		y = pending_output->usable_area.y
-			+ pending_output->usable_area.height - geo.height;
+	if (y + geo.height > usable.y + usable.height) {
+		y = usable.y + usable.height - geo.height;
 	}
-	y = MAX(y, pending_output->usable_area.y);
+	y = MAX(y, usable.y);
 
 	x += margin.left;
 	y += margin.top;
```

Consolatis:
Whoops, nice catch :)
You want to add a commit to this branch as well or do the change and force push?

johanmalm:
Thanks!
Good teamwork.
Have amended scope document with 5.38
https://github.com/labwc/labwc-scope#5-window-actions

Consolatis:
> Good teamwork.

Agree. And nice PR in the first place.

Arnaudv6:
Well thanks for nice welcome and doing the heavy-lifting :smile:

stefonarch:
Testing  now
 ```
  <windowRule identifier="*">
     <action name="MoveToCursor" />
  </windowRule>
```
is  quite the  openbox setting "Place new window under the mouse pointer". Nice option!

johanmalm:
> Testing now
> 
> ```
>  <windowRule identifier="*">
>     <action name="MoveToCursor" />
>  </windowRule>
> ```
> 
> is quite the openbox setting "Place new windows under the mouse pointer". Nice option!

Good thinking. I like that.

PR #1165, opened 2023-10-14
cursor: backport null check from wlroots-0.17 branch

jlindgren90:
Check that `wlr_layer_surface_v1_from_wlr_surface()` doesn't return NULL. This may be unnecessary with wlroots 0.16 (not sure) but doesn't hurt and reduces the delta to the wlroots-0.17 branch.


Consolatis:
:)

Issue #1167, opened 2023-10-14
Focus loss issues with xwayland app + doesn't show up in window switcher


Flrian:
Sending (+follow) JDownloader to another desktop, or closing a child window makes it lose (or not regain?) focus. Also its not showing up in the window switcher osd. Started after  #1163.

https://0x0.st/H46r.mkv

jlindgren90:
I'm not sure why this would have shown up only after #1163 but it's probably at least somewhat related to #1103.

xprop says:
```
WM_HINTS(WM_HINTS):
		Client accepts input or input focus: False
		Initial state is Normal State.
WM_PROTOCOLS(ATOM): protocols  WM_TAKE_FOCUS, WM_DELETE_WINDOW
```

However the JDownloader window does not paint correctly for me so I'm unable to debug this further.

jlindgren90:
Just found this on the Arch Wiki (https://wiki.archlinux.org/title/wayland):

> The _JAVA_AWT_WM_NONREPARENTING [environment variable](https://wiki.archlinux.org/title/Environment_variable) can be set to "1" to fix misbehavior where the application starts with a blank screen.

jlindgren90:
Related: #1130 and #1137.

johanmalm:
> Just found this on the Arch Wiki (https://wiki.archlinux.org/title/wayland):
> 
> > The _JAVA_AWT_WM_NONREPARENTING [environment variable](https://wiki.archlinux.org/title/Environment_variable) can be set to "1" to fix misbehavior where the application starts with a blank screen.

See this too:

https://github.com/labwc/labwc/blob/50ceae5fb4eb033dbffc04cf41eabc920727dd11/docs/environment#L25-L28

jlindgren90:
> closing a child window makes it lose (or not regain?) focus

I have an idea how to fix this part, will submit a PR later on.
The window not showing up in Alt-Tab would be addressed by #1139.
Not sure about the loss of focus when sending to another desktop.

PR #1142, opened 2023-10-06
xwayland: add/use wlr_xwayland_surface_offer_focus()

jlindgren90:
For testing/discussion. Depends on a wlroots change ~~that I haven't submitted yet~~ submitted here:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4384

Offer focus by sending WM_TAKE_FOCUS to a client window supporting it. The client may accept or ignore the offer. If it accepts, the surface will emit a focus_in signal notifying the compositor that it has received focus. The compositor should then call `wlr_xwayland_surface_activate(surface, true)`.

This is a more compatible method of giving focus to windows using the Globally Active input model (see `wlr_xwayland_icccm_input_model()`) than calling `wlr_xwayland_surface_activate()` unconditionally, since there is no reliable way to know in advance whether these windows want to be focused.

Previous experimental wlroots branch, with lots of extra debugging output for testing purposes:
https://github.com/jlindgren90/wlroots/commit/offer-focus

PR #1172, opened 2023-10-15
desktop: allow re-focus between "globally active" views of the same PID

jlindgren90:
Commit 7e72bf975fb6 changed behavior to not automatically focus xwayland views using the "Globally Active" input model (WM_HINTS.inputs = false but WM_TAKE_FOCUS listed in WM_PROTOCOLS).

One undesired side effect of this change is that when a dialog is closed, the parent window is not re-focused if "Globally Active". This issue is seen for example with JDownloader. It can be solved taking a similar approach to what is done for unmanaged xwayland views: allow automatic re-focus between views sharing the same PID.

Note that it's difficult to completely solve all of the focus issues with Globally Active views without proper WM_TAKE_FOCUS support. Implementing proper support is difficult since it requires wlroots changes and would also mean waiting for a message round-trip in `desktop_focus_topmost_view()`.

Fixes (partially): 7e72bf975fb65c8290b398d21b2ad9d87a22880f
("view/xwayland: avoid focusing views that don't want focus")

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -36,6 +36,21 @@ enum view_edge {
 	VIEW_EDGE_CENTER,
 };
 
+enum view_wants_focus {
+	/* View does not want focus */
+	VIEW_WANTS_FOCUS_NEVER = 0,
+	/* View wants focus */
+	VIEW_WANTS_FOCUS_ALWAYS,
+	/*
+	 * View should be offered focus and may accept or decline
+	 * (a.k.a. ICCCM Globally Active input model). Labwc generally
+	 * avoids focusing these views automatically (e.g. when another
+	 * view on top is closed) but they may be focused by user action
+	 * (e.g. mouse click).
+	 */
+	VIEW_WANTS_FOCUS_OFFER,
+};
+
 struct view;
 struct wlr_surface;
 
@@ -72,8 +87,8 @@ struct view_impl {
 	/* determines if view and surface are owned by the same process */
 	bool (*is_related)(struct view *self, struct wlr_surface *surface);
 	struct view_size_hints (*get_size_hints)(struct view *self);
-	/* if not implemented, view is assumed to want focus */
-	bool (*wants_focus)(struct view *self);
+	/* if not implemented, VIEW_WANTS_FOCUS_ALWAYS is assumed */
+	enum view_wants_focus (*wants_focus)(struct view *self);
 };
 
 struct view {
@@ -272,6 +287,8 @@ struct view *view_next(struct wl_list *head, struct view *view,
  */
 void view_array_append(struct server *server, struct wl_array *views,
 	enum lab_view_criteria criteria);
+
+enum view_wants_focus view_wants_focus(struct view *view);
 
 /**
  * view_is_focusable() - Check whether or not a view can be focused
--- a/include/view.h
+++ b/include/view.h
@@ -291,6 +291,14 @@ void view_array_append(struct server *server, struct wl_array *views,
 enum view_wants_focus view_wants_focus(struct view *view);
 
 /**
+ * view_is_focusable_from() - variant of view_is_focusable()
+ * that takes into account the previously focused surface
+ * @view: view to be checked
+ * @prev_surface: previously focused surface
+ */
+bool view_is_focusable_from(struct view *view, struct wlr_surface *prev);
+
+/**
  * view_is_focusable() - Check whether or not a view can be focused
  * @view: view to be checked
  *
@@ -302,7 +310,10 @@ enum view_wants_focus view_wants_focus(struct view *view);
  * The only views that are allowed to be focusd are those that have a surface
  * and have been mapped at some point since creation.
  */
-bool view_is_focusable(struct view *view);
+static inline bool
+view_is_focusable(struct view *view) {
+	return view_is_focusable_from(view, NULL);
+}
 
 void view_toggle_keybinds(struct view *view);
 
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -192,6 +192,8 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 struct view *
 desktop_topmost_focusable_view(struct server *server)
 {
+	struct wlr_surface *prev =
+		server->seat.seat->keyboard_state.focused_surface;
 	struct view *view;
 	struct wl_list *node_list;
 	struct wlr_scene_node *node;
@@ -202,7 +204,7 @@ desktop_topmost_focusable_view(struct server *server)
 			continue;
 		}
 		view = node_view_from_node(node);
-		if (view->mapped && view_is_focusable(view)) {
+		if (view->mapped && view_is_focusable_from(view, prev)) {
 			return view;
 		}
 	}
--- a/src/view.c
+++ b/src/view.c
@@ -143,6 +143,16 @@ view_array_append(struct server *server, struct wl_array *views,
 	}
 }
 
+enum view_wants_focus
+view_wants_focus(struct view *view)
+{
+	assert(view);
+	if (view->impl->wants_focus) {
+		return view->impl->wants_focus(view);
+	}
+	return VIEW_WANTS_FOCUS_ALWAYS;
+}
+
 bool
 view_is_focusable(struct view *view)
 {
@@ -150,7 +160,7 @@ view_is_focusable(struct view *view)
 	if (!view->surface) {
 		return false;
 	}
-	if (view->impl->wants_focus && !view->impl->wants_focus(view)) {
+	if (view_wants_focus(view) != VIEW_WANTS_FOCUS_ALWAYS) {
 		return false;
 	}
 	return (view->mapped || view->minimized);
--- a/src/view.c
+++ b/src/view.c
@@ -154,16 +154,28 @@ view_wants_focus(struct view *view)
 }
 
 bool
-view_is_focusable(struct view *view)
+view_is_focusable_from(struct view *view, struct wlr_surface *prev)
 {
 	assert(view);
 	if (!view->surface) {
 		return false;
 	}
-	if (view_wants_focus(view) != VIEW_WANTS_FOCUS_ALWAYS) {
+	if (!view->mapped && !view->minimized) {
 		return false;
 	}
-	return (view->mapped || view->minimized);
+	enum view_wants_focus wants_focus = view_wants_focus(view);
+	/*
+	 * Consider "offer focus" (Globally Active) views as focusable
+	 * only if another surface from the same application already had
+	 * focus. The goal is to allow focusing a parent window when a
+	 * dialog/popup is closed, but still avoid focusing standalone
+	 * panels/toolbars/notifications. Note that we are basically
+	 * guessing whether Globally Active views want focus, and will
+	 * probably be wrong some of the time.
+	 */
+	return (wants_focus == VIEW_WANTS_FOCUS_ALWAYS
+		|| (wants_focus == VIEW_WANTS_FOCUS_OFFER
+			&& prev && view_is_related(view, prev)));
 }
 
 /**
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -30,29 +30,61 @@ xwayland_view_get_size_hints(struct view *view)
 	};
 }
 
-static bool
+static enum view_wants_focus
 xwayland_view_wants_focus(struct view *view)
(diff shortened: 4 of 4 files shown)


jlindgren90:
This would fix the particular issue of "closing a child window makes it lose (or not regain?) focus" in #1167.

Dealing with these Globally Active views is quite a pain and I'm struggling to come up with a good generic solution. Possibly using _NET_WM_WINDOW_TYPE as the criteria for focus would solve some issues, but that property isn't currently available through the wlroots API.

I would also understand if just reverting 7e72bf975fb65c8290b398d21b2ad9d87a22880f was seen as the safest solution for now. @johanmalm @Consolatis thoughts?

jlindgren90:
For comparison, Openbox's logic (for deciding whether a view is focusable) seems to live in this `focus_valid_target()` function, which is quite intricate:
http://git.openbox.org/?p=mikachu/openbox.git;a=blob;f=openbox/focus.c;h=a4626bf83af64c900efafb1d45c3695fca832cf3;hb=HEAD#l304

It seems to allow focusing Globally Active views (indicated by the `focus_notify` flag) with some exclusions based on _NET_WM_WINDOW_TYPE.

Consolatis:
> I would also understand if just reverting [7e72bf9](https://github.com/labwc/labwc/commit/7e72bf975fb65c8290b398d21b2ad9d87a22880f) was seen as the safest solution for now.

I don't think we have to revert the whole thing, the infrastructure looks pretty useful for implementing a no-focus window rule.

Malm and I were having a discussion at IRC earlier about this topic (you are very welcome to join as well if you feel like it btw) and he came up with the idea of implementing a window rule [property](https://labwc.github.io/labwc-config.5.html#entry_properties) with a value something like `allow_focus = [yes,never,hints]`. By default we could use `yes` for now which would restore the X11 focus behavior as it was before 7e72bf975fb65c8290b398d21b2ad9d87a22880f while also
- allowing users to configure a manual no-focus rule (with `never`), useful for panels (lxqt in wayland mode but without layershell support or X11 panels with the STRUT_PARTIAL PR for example) or "desktop widget" like system monitors (e.g. conky)
- allowing to keep working on the detailed X11 focus semantics and maybe when switching to wlroots 0.17 we can change the default to `hints` and properly support the globally active model

jlindgren90:
Windows rules could work. I guess I'm resistant to them mainly because this all "just works" in Openbox (or other WMs like xfwm4) without any manual config necessary. I tend to think matching on _NET_WM_WINDOW_TYPE would be better.

I could work on a wlroots merge request to expose _NET_WM_WINDOW_TYPE. It's already being used internally e.g. by `wlr_xwayland_or_surface_wants_focus()` -- which itself is almost what we want, except for omitting the important _NET_WM_WINDOW_TYPE_DOCK.

IRC is awkward for me because the times I have available to work on labwc are very sporadic, but maybe I'll try to check in one of these days.

src/xdg.c
```diff
@@ -554,7 +554,7 @@ xdg_toplevel_view_unmap(struct view *view, bool client_request)
 		view->mapped = false;
 		wlr_scene_node_set_enabled(&view->scene_tree->node, false);
 		wl_list_remove(&view->commit.link);
-		desktop_focus_topmost_view(view->server);
+		desktop_focus_topmost_view(view->server, view->surface);
```
jlindgren90, line 557: This should probably be made conditional on the unmapped view actually having been focused.

jlindgren90: See #1173.

jlindgren90:
Marking this as draft, just to discuss the general idea for now. The implementation could probably be cleaned up a bit.

jlindgren90:
Rebased onto #1174 for `view_is_related()` which is useful here, and cleaned up quite a bit.

jlindgren90:
New, maybe bad, idea: allow any view that has/wants SSD to show up in Alt-Tab and be automatically focusable. It's crude and not perfect but might just fix the most common apps, without causing unwanted stuff like panels and notifications to get focus.

Consolatis:
There are also CSD applications (e.g. gnome stuff).

jlindgren90:
> There are also CSD applications (e.g. gnome stuff).

True, but so far I have not heard that any GTK apps set WM_HINTS.input to false or have issues with focus. So hopefully those are not affected to begin with. It seems like it's mostly Java or less common X11 toolkits that have issues. Matching on title bar might be good enough to fix most of those.

I'm not saying this should be a complete/final solution (I still think matching on _NET_WM_WINDOW_TYPE would be better) but maybe it would help.

johanmalm:
It's pretty hard to know where to go with this, but I back the approach. It seems like a step forward from the simple hints->input approach. I think it'll take a while to grind this out (with for example _NET_WM_WINDOW_TYPE and openbox logic). I suspect feedback from users will important on the journey.

Agree that ideally we want this to work out of the box as much as possible. Reverting 7e72bf9 would be the 'lazy' it-is-what-it-is approach, but I'm all for trying to raise the bar on this. If we get a lot of user-kickback we could consider config options/window-rules (but settling for work-arounds like that should always be plan B where possible - as long as we have energy and enthusiams :smile: )

So, I'm good with this one except that I'd ask for consideration for the points below (but feel free to merge if you don't think relevant).

- [ ] Consider renaming `enum view_focus_type` to include a noun like `desire`, `exepectation` or `want`. Maybe `view_focus_expectation_type`. Expectation seems to be the word icccm use.
- [ ] Does the patch-set need rebasing against master to get rid of the `.is_related` method (which is now already in master).
- [ ] The description of `WLR_ICCCM_INPUT_MODEL_LOCAL` indicates that it should only take focus if same PID. So, `s/NORMAL/MANUAL/` ??
- [ ] Consider a different name than `MANUAL` because it might make the reader think that there is a manual step (requiring user-interaction) which is not the case. Maybe `VIEW_FOCUS_EXPECTATION_IF_SAME_PID` or `VIEW_FOCUS_EXPECTATION_IF_RELATED`?


Also, _NET_WM_WINDOW_TYPE is used by `sway` in `sway/desktop/xwayland.c` to I think we should be able to use that without any changes to `wlroots`.

jlindgren90:
It'll need to be rebased but that should be trivial to do. I'll sleep on the naming questions, naming is perhaps one of the hardest parts of development for me :)

jlindgren90:
Re: _NET_WM_WINDOW_TYPE, it looks like sway opens a new xcb connection to the X server just to fetch the atoms. I would really like to avoid doing that especially since it's duplicating code that wlroots already has.

Maybe worth its own discussion thread -- how would it be to depend on a soft-fork of wlroots with a select few additional changes (with the intent to get them into mainline wlroots in time)? It could unblock several existing PRs as well as make it possible to add stuff like this _NET_WM_WINDOW_TYPE support. The technical issues seem already mostly solved via meson subproject.

Consolatis:
> Maybe worth its own discussion thread -- how would it be to depend on a soft-fork of wlroots with a select few additional changes (with the intent to get them into mainline wlroots in time)? It could unblock several existing PRs as well as make it possible to add stuff like this _NET_WM_WINDOW_TYPE support. The technical issues seem already mostly solved via meson subproject.

That would break compiling with a system wide wlroots installation. Even though one could work around it with `--force-fallback-for=wlroots` that would still completely break all packaging. As such that would be a clear "no" from my side.

[Packaging status](https://repology.org/project/labwc/versions)

jlindgren90:
One naming idea:

```
enum view_wants_focus_t {
    VIEW_WANTS_FOCUS_NEVER,
    VIEW_WANTS_FOCUS_ALWAYS,
    VIEW_WANTS_FOCUS_OFFER,
};
```

"Offer" fits well with the addition of `wlr_xwayland_surface_offer_focus()` and is what I originally used in #1142. Maybe it's technical enough not to be confusing to users?

P.S. We don't need to do any same-PID check for "Locally Active" views. The ICCCM docs are just saying that those clients won't steal focus from another app (pinky promise) but can switch focus between multiple views owned by that same client. The WM can still focus the client's views whenever. IOW it's a restriction on the client, not the WM.

jlindgren90:
> That would break compiling with a system wide wlroots installation.

I was thinking, build wlroots statically and avoid linking to the system-wide copy at all. It works fine from a technical standpoint, I've done it locally. But maybe it would go against some distro's policies on bundled libraries, I don't know.

I definitely understand the desire not to soft-fork just for the sake of avoiding fragmentation. It just irks me personally to have to maintain a personal fork/patchset of labwc continually and not be able to upstream certain key (for me) features/fixes.

Consolatis:
> > That would break compiling with a system wide wlroots installation.
> 
> I was thinking, build wlroots statically and avoid linking to the system-wide copy at all. It works fine from a technical standpoint, I've done it locally. But maybe it would go against some distro's policies on bundled libraries, I don't know.
> 
> I definitely understand the desire not to soft-fork just for the sake of avoiding fragmentation. It just irks me personally to have to maintain a personal fork/patchset of labwc continually and not be able to upstream certain key (for me) features/fixes.

Yes, I completely understand that. It should get a bit better when wlroots does the 0.17 release and we can start tracking it with our master branch. I am also still up to have something like a labwc/0.7-dev branch that contains your 0.17 PR patchset and gets rebased on master regularly. It could contain further labwc features and would allow PRs against it.

jlindgren90:
Rebased onto master, no other changes. If @johanmalm is good with the `VIEW_WANTS_FOCUS_NEVER/ALWAYS/OFFER` naming I'll make that change as well.

jlindgren90:
Possible stop-gap fix based on _MOTIF_WM_HINTS:
https://github.com/jlindgren90/labwc/commit/cf9bcc4ee0a5772e6af7791a4eadb480d7c067ab

This is enough to make IntelliJ IDEA and JDownloader (two apps that have been mentioned as having focus issues lately) work normally. Obviously not a complete/permanent fix, but might be worthwhile considering until we figure out how to access _NET_WM_WINDOW_TYPE.

johanmalm:
> Rebased onto master, no other changes. If @johanmalm is good with the `VIEW_WANTS_FOCUS_NEVER/ALWAYS/OFFER` naming I'll make that change as well.

Yes happy with VIEW_WANTS_FOCUS_OFFER et al

johanmalm:
Thanks! Looks good.

johanmalm:
> Re: _NET_WM_WINDOW_TYPE, it looks like sway opens a new xcb connection to the X server just to fetch the atoms. I would really like to avoid doing that especially since it's duplicating code that wlroots already has.

Have opened a dedicated discussions about how we track `wlroots` to have more of a wholesome think through that.

My opinion on the additional xcb connection for atoms vs sorting out the upstream support/API is that we'll have a much easier (and probably more productive) ride with the former even if it's technically not as neat (whilst on the journey). The hard bit is likely to be the trial & error with lots of user feedback to evaluate and conclude the right blend of X window-types/properties/etc for focus to just work out of the box.

I might just be a bit "burnt" from tracking `master` previously, but that's my intuition anyway.

Consolatis:
There is also a obscure `user_event_handler()` callback which I think I made use of in combination with an own xcb connection in my earlier experiments:
```
/**
* Add a custom event handler to xwayland. Return 1 if the event was
* handled or 0 to use the default wlr-xwayland handler. wlr-xwayland will
* free the event.
*/
int (*user_event_handler)(struct wlr_xwm *xwm, xcb_generic_event_t *event)
```

It supplies a pointer to a private `struct wlr_xwm` which theoretically also has  a pointer item but again, its private.

Consolatis:
See https://github.com/Consolatis/labwc/commits/testing/x11_extend_window_types for an example to detect the DOCK type and do more xwayland shenanigans that doesn't require upstream changes.

Consolatis:
Added some more stuff to play around with: `_ABOVE` `_BELOW` `STRUT_PARTIAL`.
It now also prints out a big error when mapping tint2:
```00:00:18.342 [../src/xwayland.c:619] is dock: yes```

Syncing X11 virtual desktops with our desktop impl also looks feasible with this approach but I didn't try it yet.

jlindgren90:
Nice, maybe worth opening a new draft PR?

Consolatis:
> Nice, maybe worth opening a new draft PR?

I can add one for syncing the atoms + parsing xsurface->window_types later, that is the easy part.
The other part that involves the `user_event_handler` callback is a bit more tricky and I don't know how much time I can spend on that, too many things at once currently.

Consolatis:
- #1185

Issue #1166, opened 2023-10-14
swaylock surface loses focus, can't unlock session


jlindgren90:
About 1 in 10 times (rough guess), I return to a labwc session locked using swaylock and can't unlock it; key presses don't seem to register with swaylock at all. (swaylock is displayed correctly though.)

So far I've determined that when this occurs, `keyboard_key_notify()` is still being called and `handle_compositor_keybindings()` returns false (as it should) but there is no focused surface to send the keypresses to (`seat->seat->keyboard_state.focused_surface` is NULL).

I'm continuing to try to find the root cause, but it's slow progress since the issue only occurs once every few days. Just wanted to record that the issue exists, and progress so far in debugging it.

(Not sure if related to #885.)

johanmalm:
1. Does it happen independently of any output-layout changes?
2. `swaylock` has pushed a few commits which look like potential bug fixes since `1.7.2`. Are you running with their `master`?

jlindgren90:
I am not sure yet whether it's related to output changes or not. I'll keep in mind to try the newer swaylock, thanks for the heads-up.

jlindgren90:
It happened once more this evening and I captured a little more log output. The swaylock surface was not unmapped, it just lost focus. I'm currently thinking this is not a swaylock bug but something on our side.

I'm looking suspiciously at all the `seat_focus_surface()` call sites that don't have a `!server->session_lock` check. For example, there is one in `desktop_focus_view()` but not in the `else` branch of `desktop_focus_topmost_view()`.

jlindgren90:
Found it. I can reproduce the issue by locking the screen while an xfce4-notifyd notification is onscreen (running with GDK_BACKEND=x11), but no focusable views. When the notification is destroyed, the `else` branch of `desktop_focus_topmost_view()` is entered and calls `seat_focus_surface(seat, NULL)`.

So technically this issue was introduced by 3022985ba797a061df8c8ac33b2dfb77bf20dcb3 (which refactored `desktop_focus_topmost_view()` and `desktop_focus_view()`) and would be fixed (accidentally) by #1173.

I'll work on a fix.

Consolatis:
Nice debugging!

PR #1173, opened 2023-10-15
view: only focus topmost view if unmapped view was focused

jlindgren90:
The `unmap()` handlers should only call `desktop_focus_topmost_view()` if the unmapped view was the focused view. Unmapping a view that was not focused should not change the focus.

I expect this rarely had any effect in practice; it would only matter in a focus-follows-mouse config where some view other than the one on top was focused. But it still seems better to fix.

Rather than repeating the logic in two places, create a small `view_impl_unmap()` helper. Perhaps more common "unmap" logic could be moved there in future.

### DIFF
--- a/include/view-impl-common.h
+++ b/include/view-impl-common.h
@@ -19,6 +19,7 @@ void view_impl_move_to_front(struct view *view);
 void view_impl_move_to_back(struct view *view);
 void view_impl_move_sub_views(struct view *parent, enum z_direction z_direction);
 void view_impl_map(struct view *view);
+void view_impl_unmap(struct view *view);
 
 /*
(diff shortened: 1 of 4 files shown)


jlindgren90:
> I expect this rarely had any effect in practice

"Famous last words" since this was actually a contributing cause to #1166,

Consolatis:
> > I expect this rarely had any effect in practice
> 
> "Famous last words" since this was actually a contributing cause to #1166,

:)

LGTM, didn't test.

johanmalm:
Thanks!
Great catch. Seems to make so much sense. Can't believe we didn't already have that logic in there.

PR #1174, opened 2023-10-15
view: add view_is_related()

jlindgren90:
Allows removing xwayland-specific stuff from seat.c.

Based on a suggestion from @Consolatis.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -37,6 +37,7 @@ enum view_edge {
 };
 
 struct view;
+struct wlr_surface;
 
 /* Basic size hints (subset of XSizeHints from X11) */
 struct view_size_hints {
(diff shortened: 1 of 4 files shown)


include/view.h
```diff
@@ -68,6 +69,7 @@ struct view_impl {
 	void (*move_to_back)(struct view *view);
 	struct view *(*get_root)(struct view *self);
 	void (*append_children)(struct view *self, struct wl_array *children);
+	bool (*is_related)(struct view *self, struct wlr_surface *surface);
```
Consolatis, line 72: Suggest adding a short comment what exactly "is_related" means, same for the view_ wrapper function below.

Other than that: LGTM

jlindgren90: I added comments, thanks for the suggestion :+1:

Consolatis:
:)

PR #1175, opened 2023-10-15
seat: move session-lock check down to seat_focus() level

### LINKED ISSUE #1166: swaylock surface loses focus, can't unlock session

jlindgren90:
We were checking for a locked session in `desktop_focus_view()`, but there are several other call sites of `seat_focus_surface()` which were missing such a check. Any one of those could cause the lock screen to lose focus (making the session impossible to unlock) or another surface to gain it (breaching the session lock).

To fix the issue, make any call to `seat_focus_surface()` no-op when the session is locked. Add a specific `seat_focus_lock_surface()` function which is the only way to bypass the check and is called only from session-lock.c.

This should fix #1166, but more eyes on this code are very much appreciated.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -432,6 +432,13 @@ void seat_init(struct server *server);
 void seat_finish(struct server *server);
 void seat_reconfigure(struct server *server);
 void seat_focus_surface(struct seat *seat, struct wlr_surface *surface);
+
+/**
+ * seat_focus_lock_surface() - ONLY to be called from session-lock.c to
+ * focus lock screen surfaces. Use seat_focus_surface() otherwise.
+ */
+void seat_focus_lock_surface(struct seat *seat, struct wlr_surface *surface);
+
 void seat_set_focus_layer(struct seat *seat, struct wlr_layer_surface_v1 *layer);
 void seat_set_pressed(struct seat *seat, struct view *view,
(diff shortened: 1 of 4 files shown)


Consolatis:
We should really test if this still works without the session lock protocol (I think the old variant used layershell + input inhibit).
E.g. maybe on an old versions of swaylock before they implemented the session lock protocol.

Overall this looks good. I like the wrappers around it and thus that the outside API is kept the same.

jlindgren90:
> We should really test if this still works without the session lock protocol (I think the old variant used layershell + input inhibit).

Agree, would be good to test that

johanmalm:
LGTM.
Have tested with swaylock `1.6` (pre-session-lock) and confirm it works well.
Nice fix + refactoring. Thanks!

PR #1177, opened 2023-10-16
xwayland: assume views wanting decorations also want focus

### LINKED ISSUE #1167: Focus loss issues with xwayland app + doesn't show up in window switcher

jlindgren90:
Assume that Globally Active xwayland views do want focus if they want window decorations (according to `_MOTIF_WM_HINTS`). This is a stop-gap fix to ensure that various applications (mainly Java-based ones such as IntelliJ IDEA) get focus normally and appear in the window switcher. It would be better to match based on `_NET_WM_WINDOW_TYPE` instead, but that property isn't currently available through wlroots API.

Fixes (partially): 7e72bf975fb65c8290b398d21b2ad9d87a22880f
("view/xwayland: avoid focusing views that don't want focus")

Fixes #1167 for the application mentioned (JDownloader) as well as IntelliJ IDEA. If there are X11 applications that use the Globally Active model + have client-side decorations + want normal focus (e.g. to appear in Alt-Tab), this won't help with those.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -66,12 +66,24 @@ xwayland_view_wants_focus(struct view *view)
 	 * should respond with a SetInputFocus request.
 	 *
 	 * [Currently, labwc does not fully support clients voluntarily
-	 * taking focus via the WM_TAKE_FOCUS + SetInputFocus mechanism
-	 * and avoids automatically focusing Globally Active windows.
-	 * This may change in future.]
+	 * taking focus via the WM_TAKE_FOCUS + SetInputFocus mechanism.
+	 * Instead, we try to guess whether the window wants focus based
+	 * on some heuristics -- see below.]
 	 */
 	case WLR_ICCCM_INPUT_MODEL_GLOBAL:
-		return VIEW_WANTS_FOCUS_OFFER;
+		/*
+		 * Assume the window does want focus if it wants window
+		 * decorations (according to _MOTIF_WM_HINTS). This is
+		 * a stop-gap fix to ensure that various applications
+		 * (mainly Java-based ones such as IntelliJ IDEA) get
+		 * focus normally and appear in the window switcher. It
+		 * would be better to match based on _NET_WM_WINDOW_TYPE
+		 * instead, but that property isn't currently available
+		 * through wlroots API.
+		 */
+		return (xsurface->decorations ==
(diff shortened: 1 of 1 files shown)


jlindgren90:
Related to discussion on #1172.

jlindgren90:
@Flrian @kyak testing appreciated.

src/xwayland.c
```diff
@@ -33,26 +33,67 @@ xwayland_view_get_size_hints(struct view *view)
...
+	/*
+	 * No Input - The client never expects keyboard input.
+	 *
+	 * No Input and Globally Active clients set the input field to
+	 * False, which requests that the window manager not set the
+	 * input focus to their top-level window.
+	 */
+	default: /* WLR_ICCCM_INPUT_MODEL_NONE */
+		return false;
+	}
```
Consolatis, line 96: Based on the comment: why is `WLR_ICCCM_INPUT_MODEL_NONE` not a `case` here?
Also, no need to check for the input hint here anymore?

jlindgren90: Just a style idiosyncrasy of mine, I guess. Some compilers and/or style guidelines complain if you have a `default` case but all enum values are covered. Others complain if you don't have a `default` case. This makes both happy. I don't care strongly about it.

The return value of `wlr_xwayland_icccm_input_model()` incorporates `WM_HINTS.input` already. See:
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c?ref_type=heads#L2318

Consolatis: I think using the case and no default here is the safest variant. If wlroots ends up adding a new member to the enum we should run into a compile error as not all enums are covered and there is not default. That prevents accidentally treating a newly added enum as `WLR_ICCCM_INPUT_MODEL_NONE`.

> The return value of wlr_xwayland_icccm_input_model() incorporates WM_HINTS.input already. 

Got it, thanks for the explanation.

jlindgren90:
Updated to address comments; this now shares a common first commit (non-functional change) with #1172.

johanmalm:
Second commit LGTM.

jlindgren90:
Thanks, rebased and merged (only the second commit).

PR #1181, opened 2023-10-17
view: raise/lower sub-view correctly relative to other sub-views

jlindgren90:
When a parent view has multiple sub-views (dialogs) visible, focusing one sub-view ought to raise it above the others. This doesn't currently happen -- focusing a sub-view raises the whole group of views together, but has no effect on the relative stacking order between them.

I thought I saw an issue ticket for this recently, but I can't find it now.

Anyway, this seems like a simple oversight in the `view->impl->move_to_front()` functions that's pretty easy to fix. `move_to_back()` appears to need similar logic, so add it there too.

Tested with HomeBank: the Import dialog pops up an additional Open File dialog, which before this change appears behind the Import dialog (and clicking on it does not raise it to the front). After this change, the Open File dialog appears in front as expected.

Only after implementing the fix for xwayland did I notice that xdg has identical logic in its `move_to_front/back()` impl functions. I don't know why we even have `view->impl->move_to_front/back()`, since the logic is identical (except for calling different `get_root()` functions, which is easily abstracted by just calling `view_get_root()`).

So the second commit deduplicates the logic by putting it directly in `view_move_to_front/back()`. This way, we can fix the bug in one place.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -349,14 +349,20 @@ static void
 xdg_toplevel_view_move_to_front(struct view *view)
 {
 	struct view *root = xdg_toplevel_view_get_root(view);
+	/* FIXME: this exact code is repeated in xwayland.c */
 	view_impl_move_to_front(root);
 	view_impl_move_sub_views(root, LAB_TO_FRONT);
+	/* make sure view is in front of other sub-views */
+	if (view != root) {
+		view_impl_move_to_front(view);
+	}
 }
 
 static void
 xdg_toplevel_view_move_to_back(struct view *view)
 {
 	struct view *root = xdg_toplevel_view_get_root(view);
+	/* FIXME: this exact code is repeated in xwayland.c */
 	view_impl_move_sub_views(root, LAB_TO_BACK);
 	view_impl_move_to_back(root);
 }
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -614,14 +614,20 @@ static void
 xwayland_view_move_to_front(struct view *view)
 {
 	struct view *root = xwayland_view_get_root(view);
+	/* FIXME: this exact code is repeated in xdg.c */
 	view_impl_move_to_front(root);
 	view_impl_move_sub_views(root, LAB_TO_FRONT);
+	/* make sure view is in front of other sub-views */
+	if (view != root) {
+		view_impl_move_to_front(view);
+	}
 }
 
 static void
 xwayland_view_move_to_back(struct view *view)
 {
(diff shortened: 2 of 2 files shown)


Consolatis:
Nice catch and refactor.

`impl->get_root()` maybe have been added after `impl->move_to{front,back}()`

I think the base idea of view-impl was to not call it from the outside of xdg.c / xwayland.c.
So maybe it would be cleaner to keep the impl pointers for `move_to_{front,back}` but set them to `view_impl_move_to_{front,back}` in both implementations?

Just looked at the xwayland implementation and its a bit strange that we don't use `wlr_xwayland_surface_restack()` there, we only do that in `xwayland_view_set_activated()`.

jlindgren90:
I'm not familiar with wlr_xwayland_surface_restack() or when it needs to be called. Would have to do some research.

Maybe the code for moving related views rather belongs in desktop.c? That file I think is meant to hold routines dealing with multiple views.

src/xdg.c
```diff
@@ -349,15 +349,25 @@ static void
...
 static void
 xdg_toplevel_view_move_to_back(struct view *view)
 {
 	struct view *root = xdg_toplevel_view_get_root(view);
+	/* FIXME: this exact code is repeated in xwayland.c */
 	view_impl_move_sub_views(root, LAB_TO_BACK);
+	/* make sure view is behind other sub-views */
+	if (view != root) {
+		view_impl_move_to_back(view);
+	}
```
Consolatis, line 370: The `move_to_front` LGTM.

The `move_to_back` part I am not so sure: when using the `Lower` action to move something to the back it causes a loss of the focused subview that was previously in front of all other (potentially rejecting input) subviews. When then clicking on the then-on-top subview (same for `A-Tab` or some panel click), we will raise the whole stack to the front again but may not be able to interact with it at all due to the application rejecting all input for other subviews and the toplevel itself. Can be tested with a text editor for example (I've used `GDK_BACKEND=x11 pluma`), opening the open-file dialog and then about dialog (which block input for everything else) then pressing a keybind that is set to run the `Lower` action. Result: the about dialog is between the editor and the open-file window and nothing can be interacted with unless manually focusing the about dialog again (which now works thanks to this PR). Doesn't work that easy when its hidden behind the new subview though.

So I don't think we need the `move_to_back` adjustment here as we just move the whole stack as it is to the back.

jlindgren90: That makes sense to me, thanks for thinking it through. I'll omit the `move_to_back` part.

johanmalm:
Yes, thanks indeed!

Issue #1011, opened 2023-08-05
Comment out most of docs/environment


Consolatis:
Sometimes people wanting to test labwc throw the example environment file into their config dir and wonder why their keyboard layout is suddenly changed. Same for the breeze cursor theme, it might not be available on the given system.

Maybe we should simply comment those settings out so they still serve documentation purposes but do not actually change the keyboard layout.

johanmalm:
Good idea.

Suggest commenting out all.

XDG_CURRENT_DESKTOP is set in session.c anyway.

PR #1184, opened 2023-10-17
docs/environment: comment out variables

johanmalm:
...to avoid users copying this file to ~/.config/labwc/ without editing it and ending up with Swedish keyboard layout (which has been reported by users more than once!)

### DIFF
--- a/docs/environment
+++ b/docs/environment
@@ -1,28 +1,67 @@
-# Example environment file
+##
+## Example ~/.config/labwc/environment file.
+## Uncomment lines starting with one '#' to suit your needs.
+##
 
-# This allows xdg-desktop-portal-wlr to function (e.g. for screen-recording)
-XDG_CURRENT_DESKTOP=wlroots
+##
+## Use the XKB_DEFAULT_LAYOUT variable to set the keyboard layout. For example
+## to start with Swedish keyboard layout set it to 'se'. If you are unsure what
+## your country code is, refer to the layout section of:
+## /usr/share/X11/xkb/rules/evdev.lst
+##
+## Multiple keyboard layouts can be set by comma-separating the country codes.
+## If multiple layouts are used, specify the toggle-keybind using
+## XKB_DEFAULT_OPTIONS as show below.
+##
+## For further details, see xkeyboard-config(7)
+##
 
-# Set keyboard layout to Swedish
-XKB_DEFAULT_LAYOUT=se
+# XKB_DEFAULT_LAYOUT=se
+# XKB_DEFAULT_LAYOUT=se,de
+# XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle
+# XKB_DEFAULT_OPTIONS=grp:shift_caps_toggle
 
-# Set two keyboard layouts and toggle between them using alt+shift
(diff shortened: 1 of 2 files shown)


docs/environment
```diff
@@ -1,28 +1,61 @@
...
+
+# WLR_NO_HARDWARE_CURSORS=1
+
+##
+## For Java applications such as JetBrains/Intellij Idea, set this variable
+## to avoid menus with incorrect offset and blank windows.
+## See https://github.com/swaywm/sway/issues/595
+##
+
+# _JAVA_AWT_WM_NONREPARENTING=1
```
Consolatis, line 52: I don't think that this makes anything worse so I'd vote for leaving this active.

docs/environment
```diff
@@ -1,28 +1,61 @@
...
+
+##
+## This allows xdg-desktop-portal-wlr to function (e.g. for screen-recording).
+## It is set automatically by labwc though, so it is only includeded here for
+## completeness.
+##
+
+# XDG_CURRENT_DESKTOP=wlroots
```
Consolatis, line 60: AFAIR setting this prevents labwc to overwrite it. So it can also be used to force something else in here (this is just from memory though, didn't verify it).

jlindgren90:
> and ending up with Swedish keyboard layout

Well, what's wrong with that? :)

LGTM.

Consolatis:
Other than the two points above: LGTM (and thanks for sending a PR for it!).

Also fixes
- #1011

johanmalm:
> > and ending up with Swedish keyboard layout
> 
> Well, what's wrong with that? :)
> 
> LGTM.

Ha ha. Well quite.
Although it even catches me out sometimes when trying to :q in vim and it starts to record a macro.

Thanks for review. Have logged out, so will have to fix the two from @Consolatis tomorrow

docs/environment
jlindgren90: I'd argue that having the example config file differ from the default config is still going to cause confusion. If we are going to recommend this in the example, maybe we should just enable it by default?

jlindgren90:
(Lindgren is Swedish, hence why I spoke up. I don't actually use or know Swedish, regrettably...)

docs/environment
Consolatis: Hm.. what about we set it ourselves before loading the environment file, stating that in the comment and changing it to `=0` int he example file.

The comment could then read something like
```
To avoid menus with incorrect offsets and blank windows with some Java
applications like JetBrains / Intellij Idea or JDownloader  we set the following
to `1` internally. It can be overwritten by uncommenting the following line.
```

(Assuming that JDownloader has the same issue and fix).

jlindgren90: Could work. Not a big deal to me, anyway.

johanmalm: Have gone for setting this in `src/config/session.c` and adjusting `docs/environment` as @Consolatis suggested (ish).

docs/environment
johanmalm: Correct.
Have amended comment to clarify this.

johanmalm:

Consolatis:
:)

PR #1185, opened 2023-10-18
xwayland: support querying more window types

Consolatis:
This should allow for follow up work to deal with all window types mentioned at https://specifications.freedesktop.org/wm-spec/wm-spec-1.4.html#idm45649101374512

The PR contains a new function exposed via `xwayland.h` (so it can be used for `-unmanaged` as well):
```c
bool xwayland_surface_contains_window_type(
		struct wlr_xwayland_surface *surface, enum atom window_type);
```

It also adds a (`xwayland.c`) listener for `set_window_type` so there is an entry point to not just check at map time but also get notified about dynamic changes of the window types.

The `atom_names` array could be moved to `xwayland.c` rather than living within `xwayland.h`. If we want to keep the `enum atom` argument for the new function the enum would have to stay in `xwayland.h`. An alternative to that would be to create various wrapper functions like `bool xwayland_surface_is_dock(struct wlr_xwayland_surface *surface);` which then just call into a static `xwayland_surface_contains_window_type()`.

The decision of where to place the array and enum kind of depends on how far we want to extend the atom list and if `xwayland-unmanaged.c` should be able to use them directly or only call `xwayland_` wrapping functions.

Other than that this PR is not really doing much, it syncs some atoms whenever the xwayland server signals being ready (e.g. also when restarting due to idle timeout). The list of atoms can be easily extended in the future. The example uses of the new function are just useless logs with `WLR_ERROR`.

Feel free to cannibalize this PR (and the commit) as required.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -15,7 +15,22 @@
 #include "workspaces.h"
 #include "xwayland.h"
 
+xcb_atom_t atoms[ATOM_LEN] = {0};
+
 static void xwayland_view_unmap(struct view *view, bool client_request);
+
+bool
+xwayland_surface_contains_window_type(
+		struct wlr_xwayland_surface *surface, enum atom window_type)
+{
+	assert(surface);
+	for (size_t i = 0; i < surface->window_type_len; i++) {
+		if (surface->window_type[i] == atoms[window_type]) {
+			return true;
+		}
+	}
+	return false;
+}
 
 static struct view_size_hints
 xwayland_view_get_size_hints(struct view *view)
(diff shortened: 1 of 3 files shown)


Consolatis:
Rebased on master and split off the useless example into its own commit.

johanmalm:
I think this looks good (except the example of course).

src/xwayland.c
```diff
@@ -478,6 +494,12 @@ handle_set_decorations(struct wl_listener *listener, void *data)
 	view_set_decorations(view, want_deco(xwayland_view->xwayland_surface));
 }
 
+static void
+handle_set_window_type(struct wl_listener *listener, void *data)
```
johanmalm, line 498: What's this one for?

johanmalm: Should we connect the signal in readiness?

Consolatis: It was originally connected as part of the example but I left it in there to make it easier to actually use the feature.
I can remove the signal handler though if preferred, technically it isn't necessary for `xwayland_surface_contains_window_type()`.

johanmalm:
LGTM

jlindgren90:
Nice work :)

FYI, I created a wlroots merge request to add a similar function there:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4553

I think that will be better in the long run.

PR #1191, opened 2023-10-20
view: macro-ize signal/listener connection

jlindgren90:
@Consolatis mentioned this as an idea in #814. It seems worth pursuing to me, since it makes the code a bit more readable IMHO (and forces us to be consistent with event handler function names).

No functional change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -228,12 +228,12 @@ We have a very small, modest API and encourage you to use it.
 4. `wl_array_len()` to get number of elements in a `wl_array` [common/array.h]
 
 5. `ARRAY_SIZE()` to get number of elements in visible array
-   [common/array-size.h]
+   [common/macros.h]
 
 [common/mem.h]: https://github.com/labwc/labwc/blob/master/include/common/mem.h
 [common/list.h]: https://github.com/labwc/labwc/blob/master/include/common/list.h
 [common/array.h]: https://github.com/labwc/labwc/blob/master/include/common/array.h
-[common/array-size.h]: https://github.com/labwc/labwc/blob/master/include/common/array-size.h
+[common/macros.h]: https://github.com/labwc/labwc/blob/master/include/common/macros.h
 
 ### The Use of glib
 
--- a/include/common/macros.h
+++ b/include/common/macros.h
@@ -32,4 +32,22 @@
 	(dest)->name.notify = handle_##name; \
 	wl_signal_add(&(src)->events.name, &(dest)->name)
 
+/**
+ * MIN() - Minimum of two values.
+ *
+ * @note Arguments may be evaluated twice.
+ */
+#ifndef MIN
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+#endif
+
+/**
+ * MAX() - Maximum of two values.
+ *
+ * @note Arguments may be evaluated twice.
+ */
+#ifndef MAX
+#define MAX(a, b) (((a) > (b)) ? (a) : (b))
+#endif
+
 #endif /* LABWC_MACROS_H */
--- a/include/common/macros.h
+++ b/include/common/macros.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: GPL-2.0-only */
-#ifndef LABWC_ARRAY_SIZE_H
-#define LABWC_ARRAY_SIZE_H
+#ifndef LABWC_MACROS_H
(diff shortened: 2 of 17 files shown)


jlindgren90:
```
src/xwayland.c:785: CHECK: Macro argument 'name' may be better as '(name)' to avoid precedence issues
```
This seems to be a false-positive in `scripts/check` (parentheses cannot be added around struct field names). Not sure of how to best address this.

Consolatis:
Could we have that defined in some header instead?
Something like `CONNECT_SIGNAL()`?

I don't think there is a reason to have xwayland-unmanaged.c prefix the handlers with `unmanaged_`, those are static functions anyway.

> This seems to be a false-positive in scripts/check (parentheses cannot be added around struct field names). Not sure of how to best address this.

I guess we could roll out the dark magic (e.g. void pointer of dest + offsetof() from name) to have it work for the `dest` part. But the function name part for the handler would still break and type casting is also missing. Easier solution: we just remove that check from checkpatch.pl. CC @johanmalm

Consolatis:
Btw, didn't you add a function doing basically the same for the 0.17 surface associate events?
That would be an alternative as all handlers share the same signature.

johanmalm:
I like this in principle.

Agree with putting macros in header and force commonality (i.e. remove `unmanaged_`).
Other than that LGTM.


Suggest disabling check with:

```
diff --git a/scripts/checkpatch.pl b/scripts/checkpatch.pl
index e6b701f..53d6d26 100755
--- a/scripts/checkpatch.pl
+++ b/scripts/checkpatch.pl
@@ -64,6 +64,7 @@ my @ignore = (
 	"PREFER_FALLTHROUGH",
 	"INITIALISED_STATIC",
 	"UNNECESSARY_ELSE",
+	"MACRO_ARG_PRECEDENCE",
 );
 my $help = 0;
 my $configuration_file = ".checkpatch.conf";
```

johanmalm:
Reminds me a bit of `QObject::connect()`

jlindgren90:
> Btw, didn't you add a function doing basically the same for the 0.17 surface associate events? That would be an alternative as all handlers share the same signature.

Not really - only map/unmap are handled in a separate function. The macro is still useful for all the other signals, including the new associate/dissociate.

jlindgren90:
> Reminds me a bit of `QObject::connect()`

Yes, definitely inspired by that :)

jlindgren90:
Partially updated - I will return to this later.

OK to rename `common/array-size.h` to `common/macros.h`? Having one header per macro seems excessive.

Consolatis:
> OK to rename `common/array-size.h` to `common/macros.h`? Having one header per macro seems excessive.

No issues with that from my side.

jlindgren90:
Should be all set now.

Consolatis:
> > OK to rename `common/array-size.h` to `common/macros.h`? Having one header per macro seems excessive.
> 
> No issues with that from my side.

Actually we could also move the `MIN()` and `MAX()` macros there. Everything we can out of `labwc.h` is a good thing in my book. We don't have to do that in this PR though.

jlindgren90:
Style check was failing due to an indentation issue, fixed now.

jlindgren90:
Moved MIN and MAX as well.

johanmalm:
Be careful with min-max, they clash with the same macros names in cairo.h
We could make max/min lower case to avoid the clash in future

jlindgren90:
I have `MIN/MAX` in glib/gmacros.h rather than cairo.h. But I think we discussed before (when adding them to labwc.h) and decided that the `#ifndef` guard around ours was safe enough. The glib macros are identical to ours and will always take precedence if included (before or after ours) because gmacros.h does `#undef MIN/MAX` beforehand.

jlindgren90:
Incidentally, #1005 recently added static `min/max` functions in snap.c. Probably fine as long as we stick to UNIX-like platforms, but `min/max` (lowercase) are macros defined in windows.h and so naming functions `min/max` is non-portable. That bit us in Audacious years ago.

johanmalm:
> I have `MIN/MAX` in glib/gmacros.h rather than cairo.h. But I think we discussed before (when adding them to labwc.h) and decided that the `#ifndef` guard around ours was safe enough. The glib macros are identical to ours and will always take precedence if included (before or after ours) because gmacros.h does `#undef MIN/MAX` beforehand.

You’re right. Ignore my last :smile:

capital MIN MAX and ifndef makes sense

PR #1192, opened 2023-10-20
xdg: add xdg_shell_init()

johanmalm:
...to have the event-handler functions in the same translation unit as the signal connector.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -356,10 +356,7 @@ struct constraint {
 };
 
 void xdg_popup_create(struct view *view, struct wlr_xdg_popup *wlr_popup);
-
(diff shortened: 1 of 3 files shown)


johanmalm:
@jlindgren90 seeing your signal macro, made me remember this bit of outstanding refactoring.

Consolatis:
Nice, LGTM but didn't test.

I stumbled over a similar issue in `keyboard.c` recently. 
E.g. the event handlers being exposed in `labwc.h` because the setup happens in `seat.c` rather than in some `keyboard_setup()` function in `keyboard.c` but it didn't bother me enough to send a PR.

jlindgren90:
LGTM too.

PR #1193, opened 2023-10-21
view: commonize sub-view logic in view_move_to_front/back()

jlindgren90:
The logic was the same for xdg-shell and xwayland views, so move it from the `view->impl` layer out to the `view_move_to_front/back()` functions.

`view->impl->move_to_front/back()` still exist for now, in case we want to add xdg/xwayland-specific logic in future, but they now move only one view and not sub-views.

Follow-up to #1181.

### DIFF
--- a/include/view-impl-common.h
+++ b/include/view-impl-common.h
@@ -8,16 +8,10 @@
  * functions should call these functions.
  */
 
-enum z_direction {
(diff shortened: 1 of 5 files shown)


johanmalm:
LGTM. Nice refactoring. Thanks!

Issue #1194, opened 2023-10-21
Rarely, title-bar may not be resized on snapping to bottom/top by drag-and-drop


tsujan:
It happened here very randomly and rarely when I snapped a window to the bottom of the screen by dragging its title-bar and dropping it to the bottom edge. Fortunately, I took a screenshot:

![bottom](https://github.com/labwc/labwc/assets/981076/0109ad9f-85ab-4d08-aefa-bd7fd7a24685)

I encountered it only once with the top edge and haven't seen it without drag-and-drop. However, I can't be sure that it's really related to drag-and-drop.

Sorry if it's already known or fixed in git; I use 0.6.5 and couldn't find it after a simple search.

johanmalm:
Thanks for reporting. Have not seen that before!!
What's the stuff in the top left corner? Could you send config for that - to help me reproduce.

> dragging its title-bar and dropping it to the bottom edge

Are you sure? IIRC we only support dropping at top (for `Maximize`) and sides (for `SnapToEdge`)

stefonarch:
> Are you sure? IIRC we only support dropping at top (for `Maximize`) and sides (for `SnapToEdge`)

DND to bottom is similar to sides, half screen here.

johanmalm:
> > Are you sure? IIRC we only support dropping at top (for `Maximize`) and sides (for `SnapToEdge`)
> 
> DND to bottom is similar to sides, half screen here.

Yes, you're right. I've just read the manual to remind myself. Ignore my comment.

tsujan:
Attached my config folder:

[labwc.zip](https://github.com/labwc/labwc/files/13060658/labwc.zip)

tsujan:
> Have not seen that before

Because it happens very randomly and rarely. However, I might be able to make it happen for the bottom edge if I try hard. So, if you make a fixing patch that can be applied or adapted to 0.6.5, I could test it.

> What's the stuff in the top left corner?

Nothing. I sent the bottom part of the screenshot. Waybar is at the top edge of the screen, without any window rule. And PCManFM-Qt draws the desktop.

It *seems* that the non-resized title-bar always fills the left side (even when the DND is made slightly to the the right of the bottom edge), and that its size is the width of the dragged window.

Consolatis:
This may happen when the clients doesn't respond fast enough, do you have anything like
`client (some_client_app_id) did not respond to configure request in 100 ms` in your logs?

tsujan:
[...]

PR #1195, opened 2023-10-21
ssd: use `view->pending` for size calculations to avoid the titlebar being out-of-sync with the view size

johanmalm:
@tsujan - testing welcome (I can't reproduce the bug, so can't verify that this fixes it)

### DIFF
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -62,10 +62,10 @@ ssd_max_extents(struct view *view)
 	assert(view);
 	struct border border = ssd_thickness(view);
 	return (struct wlr_box){
-		.x = view->current.x - border.left,
-		.y = view->current.y - border.top,
-		.width = view->current.width + border.left + border.right,
-		.height = view->current.height + border.top + border.bottom,
+		.x = view->pending.x - border.left,
+		.y = view->pending.y - border.top,
+		.width = view->pending.width + border.left + border.right,
+		.height = view->pending.height + border.top + border.bottom,
 	};
 }
 
@@ -191,7 +191,7 @@ ssd_create(struct view *view, bool active)
 	ssd->margin = ssd_thickness(view);
 	ssd_set_active(ssd, active);
 	ssd_enable_keybind_inhibit_indicator(ssd, view->inhibits_keybinds);
-	ssd->state.geometry = view->current;
+	ssd->state.geometry = view->pending;
 
 	return ssd;
 }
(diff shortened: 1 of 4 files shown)


tsujan:
Unfortunately the issue happened again after applying the patch and restarting LabWC (exiting and logging in) — or one may say "fortunately", because there's no need to wait. It happened when I snapped to the bottom quickly for the 7th or 8th time:

qterminal

Can it be related to QtWayland, or even wlroots itself?

Consolatis:
I think using `view->current` in ssd calculations is correct:

In the #1194 case what should happen is a callchain like this
- `interactive_finish()`
  - `view_snap_to_edge()` 
    - `view_apply_tiled_geometry()`
      - `view_move_resize()`
        - `view->impl->configure()`
        - Once we get the commit from the surface via `handle_commit()`:
          - `view_impl_apply_geometry()` (which updates `current->width` and `current->height`)
            - `view_moved()`
              - `ssd_update_geometry()`

That should mean that the SSD should only use the `->current` values for its calculations to have it synced with what was commited / acknowledged from the client.

@tsujan are we talking about a wayland native application or a xwayland one? `A-Tab` should show it either as `[xdg-shell]` or `[xwayland]`.

tsujan:
> are we talking about a wayland native application or a xwayland one?

Two native Wayland apps (xdg-shell).

Consolatis:
This may happen when the clients doesn't respond fast enough, do you have anything like
`client (some_client_app_id) did not respond to configure request in 100 ms` in your logs?

tsujan:
The system log is clean. `~/.local/share/sddm/wayland-session.log` only shows lines like this about LabWC:

```
00:25:27.484 [ERROR] [../labwc-0.6.5/src/desktop.c:422] Unknown node detected
```

Consolatis:
Sorry, deleted my earlier question here and moved it to the issue itself. I thought I was fast enough before you saw it, seems that was not the case :)

johanmalm:
> I think using `view->current` in ssd calculations is correct:
> 
> In the #1194 case what should happen is a callchain like this
> 
>     * `interactive_finish()`
>       
>       * `view_snap_to_edge()`
>         
>         * `view_apply_tiled_geometry()`
>           
>           * `view_move_resize()`
>             
>             * `view->impl->configure()`
> 
>     * Once we get the commit from the surface via `handle_commit()`:
>       
>       * `view_impl_apply_geometry()` (which updates `current->width` and `current->height`)
>         
>         * `view_moved()`
>           
>           * `ssd_update_geometry()`
> 
> 
> That should mean that the SSD should only use the `->current` values for its calculations to have it synced with what was commited / acknowledged from the client.
> 
> @tsujan are we talking about a wayland native application or a xwayland one? `A-Tab` should show it either as `[xdg-shell]` or `[xwayland]`.

Thanks for analyzing @Consolatis 
I just went on a trial-n-error hunch. Will delete that branch :blush:

PR #1196, opened 2023-10-21
xwayland: update stacking order in move_to_front/back()

jlindgren90:
Currently xwayland views are restacked on top of the XWayland server stacking order when activated (i.e. focused). This is wrong because focus/raise are independent concepts (though often occurring together). The stacking order should be updated when the view is raised/lowered, not when the view is focused.
    
Work is in progress elsewhere (#1142) that will result in views more often being raised without being focused. Without this fix, those views don't get restacked properly, resulting in clicks "passing through" to views underneath.

The first commit, on its own, fixes the issue but has a side effect of increasing our traffic to the X server a bit. The second commit ("view: avoid raising same view over and over") is an optimization to reduce that traffic a bit.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -238,12 +238,17 @@ struct server {
 	struct wlr_box grab_box;
 	uint32_t resize_edges;
 
-	/* SSD state */
 	/*
 	 * Currently focused view. Updated with each "focus change"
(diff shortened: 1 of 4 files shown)


Consolatis:
Code LGTM, didn't test.

johanmalm:
Really nice. Thanks for all the work on this.

PR #1197, opened 2023-10-21
regions: perform rounding more carefully

### LINKED ISSUE #1169: small gap between regions
With my regions there is always a small gap when trying to divide the monitor into thirds. You can see it between the 2nd and 3rd region, and at the right edge of the monitor.
2023-10-14T21:59:50_grim

```
    <region name="vertical-one"     x="0%"  y="0%"  height="100%" width="33%"  />
    <region name="vertical-two"     x="33%" y="0%"  height="100%" width="34%"  />
    <region name="vertical-three"   x="67%" y="0%"  height="100%" width="33%"  />
```

jlindgren90:
Add percentages (x + width, y + height) before scaling so that there is no gap between regions due to rounding variations.

I have a strong hunch this will fix #1169. Not tested though.

### DIFF
--- a/src/regions.c
+++ b/src/regions.c
@@ -208,10 +208,19 @@ regions_update_geometry(struct output *output)
 	wl_list_for_each(region, &output->regions, link) {
 		geo = &region->geo;
 		perc = &region->percentage;
-		geo->x = usable.x + usable.width * perc->x / 100;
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks!

Issue #1060, opened 2023-08-30
Graphics tablet support


jp7677:
Hi there,

I'm using a Wacom tablet (Intous M) most of the time as a mouse replacement. The tablet works perfectly when using Sway, but nothing happens in labwc when moving the stylus, tapping the surface or clicking a button (a normal mouse works though). Since it works just fine in Sway, I guess `wlroots` has tablet support, but it is just not (yet) implemented in labwc? If that is the case, could you please consider adding support for tablets in labwc?

Kind regards and thanks a lot for your cool project here!

Consolatis:
> I guess `wlroots` has tablet support, but it is just not (yet) implemented in labwc?

This is correct.

I've played around with it quite a while ago to figure out if we could theoretically support drawing tablets and how much work it would be but stopped once seeing the events come in. Integrating the drawing tablet input into our cursor handling might be a bit more complex. I'll try to find my old test branch and look into it again but can't give any promises.

jp7677:
Hi @Consolatis , out of curiosity, did you found your old branch and is it somewhere publicly available?

jp7677:
Thanks a lot for your response and for your work in this direction. Fingers crossed for finding your power supply!

>  I'll prioritize this over other enhancements for now. But fixing bugs is still the main priority. So I can't provide any kind of time frame for this.

Very cool and yes, I totally understand that this is a niche feature and finding and fixing bugs has priority.

Consolatis:
> Fingers crossed for finding your power supply!

Well, on the one side that does appear to have helped, found that power supply.
On the other side the branch wasn't there either. So I started fresh and pushed a first very basic initial prototype, feel free to test: https://github.com/Consolatis/labwc/tree/feature/drawing_tablet_support

Some caveats:
- only supports mouse emulation for "tools" / pens
- buttons are not supported, "tip down" is treated as left mouse button press, "tip up" as release
- does not support the wayland tablet protocol. So applications that interact with things like pressure or tilting won't work and will just see a usual mouse
- the pad itself is not supported, so if it contains further things like buttons or wheels or anything fancy like that those won't work
- it will spam the log, unless you want to debug the events I recommend not running labwc with the `-V` or `-d` flags while using this branch
- no guarantees are made about the stability. I tested it a bit and didn't spot anything in that regard but it really is just a first prototype to figure out how to best hook that up into the rest of labwc

jp7677:
[...]

PR #1203, opened 2023-10-24
Refactor/input handling

Consolatis:
Nothing urgent but we accumulated quite a bit of input handling all over the place.

So this PR tries to slightly refactor the input handling parts of labwc into `src/input/`.
It also splits off the gesture handling from `src/cursor.c` into `src/input/gestures.c` and makes the keyboard key / modifier listeners private.

The implicit UAF condition when finalizing the touch (and now gesture) handlers before the cursor one is dealt with by just having `cursor_finalize()` finalize those itself (and `cursor_init()` init them).

I've tested a rebase of #626 on top of this and it went through without issues.

This is a Draft for now (and `[wip]` commits) to ask for opinions about the approach first.
It will come in handy when restarting work on #1060 as that adds yet another input method.

@johanmalm @jlindgren90


include/input/keyboard.h
```diff
@@ -0,0 +1,19 @@
...
+
+struct seat;
+struct keyboard;
+struct wlr_keyboard;
+
+void keyboard_init(struct seat *seat);
+void keyboard_finish(struct seat *seat);
+
+void keyboard_set_numlock(struct wlr_keyboard *keyboard);
+void keyboard_add_keyboard_handler(struct keyboard *keyboard, struct wlr_keyboard *wlr_kb);
```
Consolatis, line 15: It should likely be enough to just use the first arg to get the `wlr_keyboard`. Verify.

include/labwc.h
```diff
@@ -30,7 +30,6 @@
 #include <wlr/types/wlr_relative_pointer_v1.h>
 #include <wlr/types/wlr_pointer.h>
 #include <wlr/types/wlr_pointer_constraints_v1.h>
-#include <wlr/types/wlr_pointer_gestures_v1.h>
```
Consolatis, line 33: All of these could likely be moved to `include/input/cursor.h`. Verify.

jlindgren90:
Very nice, LGTM but didn't test.

An idea for some follow-up refactoring: could we split up `struct seat` into some smaller structs, maybe following similar divisions to what you have here (cursor/gestures/keyboard/etc.)? Doesn't need to be done immediately or in this PR, just something to consider.

Consolatis:
> An idea for some follow-up refactoring: could we split up `struct seat` into some smaller structs

I think we could start with moving `struct seat` from `include/labwc.h` into `include/seat.h` (same for the public `seat_` functions).
Same for things like `enum input_mode` and `struct input`.

We could also consider a `include/output.h`. Basically trying to get rid of that giant `labwc.h` include bit by bit.

Consolatis:
Rebased on master + renamed new keyboard function into  `keyboard_setup_listeners()` + cleaned up the commits .
#626 can still be rebased fine on top of this branch.

As we are potentially doing some changes to the input handling soon, getting this PR merged first seems easier.
I am not sure how exactly `git mv` reacts to being rebased on top of something that changed the contents of the file being moved.

jlindgren90:
I haven't looked over the latest changes but I agree with getting this merged. I don't think we should delay changes to master for the sake of the wlroots 0.17 PR/branch.

johanmalm:
Thanks. Nice!

PR #1204, opened 2023-10-25
xwayland: set _NET_WORKAREA property based on usable area

jlindgren90:
XWayland clients use the _NET_WORKAREA root window property to determine how much of the screen is not covered by panels/docks. The property is used for example by Qt to determine areas of the screen that popup menus should not overlap (see `QScreen::availableVirtualGeometry`).

Depends on wlroots MR (merged for 0.17):
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4406

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -830,6 +830,7 @@ handle_ready(struct wl_listener *listener, void *data)
 	struct server *server =
 		wl_container_of(listener, server, xwayland_ready);
 	wlr_xwayland_set_seat(server->xwayland, server->seat.seat);
+	xwayland_update_workarea(server);
 }
 
 void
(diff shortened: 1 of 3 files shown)


src/xwayland.c
```diff
@@ -880,3 +880,74 @@ xwayland_server_finish(struct server *server)
...
+		}
+		if (output_bottom == layout_bottom) {
+			workarea_bottom = MIN(workarea_bottom, usable_bottom);
+		}
+	}
+
+	/*
+	 * Set _NET_WORKAREA property. We don't report virtual desktops
+	 * to XWayland, so we set only one workarea.
+	 */
```
Consolatis, line 945: This is something that I'd actually like to change at some point.

jlindgren90: That would be cool. I'm not certain exactly how that would affect this code -- maybe it would just send multiple copies of the same wlr_box, maybe no change at all. Usable area is the same for all workspaces in labwc, I think?

(Openbox may also make a distinction between "virtual desktops" and "workspaces". I recall this vaguely from a discussion several years back, so I would have to refresh myself on the details.)

Consolatis: > (Openbox may also make a distinction between "virtual desktops" and "workspaces". I recall this vaguely from a discussion several years back, so I would have to refresh myself on the details.)

Maybe there is indeed some difference in the X11 terminology, I've always considered them to be synonyms but that might actually be wrong and one of them is what we call "output"?

> Usable area is the same for all workspaces in labwc, I think?

Yes, a workspace spans all outputs.

However, as layershell exclusive zones are per output (same as X11 ones I think? e.g. a panel may only be on the primary monitor) this calculation does't seem always correct. Couldn't we just use `output_usable_area_in_layout_coords()` for each output and throw the returned wlr_box directly into the X11 function?

jlindgren90: Looking through the EMWH spec, I think "virtual desktops" does mean essentially "workspaces". I don't think putting per-output geometries in _NET_WORKAREA is correct though. It seems EWMH doesn't really consider multiple physical outputs.

jlindgren90:
v2: prevent calling `wlr_xwayland_set_workareas()` too early

wlroots MR was merged with slight changes.

jlindgren90:
v3: fix segfault at exit (`server->xwayland == NULL`)

jlindgren90:
This one should be good to go now that we're targeting wlroots 0.17.

johanmalm:
LGTM

PR #1207, opened 2023-10-26
view: implement separate horizontal/vertical maximize

jlindgren90:
This is a useful (if lesser-known) feature of at least a few popular X11 window managers, for example Openbox and XFWM4. Typically right-click on the maximize button toggles horizontal maximize, while middle-click toggles vertical maximize.

Support in labwc uses the same configuration syntax as Openbox, where the Maximize/ToggleMaximize actions have an optional "direction" argument: horizontal, vertical, or both (default). The default mouse bindings match the XFWM4 defaults (not sure what Openbox has by default).

Most of the external protocols still assume "maximized" is a Boolean, which is no longer true internally. For the sake of the outside world, a view is only "maximized" if maximized in both directions.

Internally, I've taken the following approach:

- SSD code decorates the view as "maximized" (i.e. hiding borders) only if maximized in both directions.
- Layout code (interactive move/resize, tiling, etc.) generally treats the view as "maximized" (with the restrictions that entails) if maximized in either direction. For example, moving a vertically- maximized view first restores the natural geometry (this differs from Openbox, which instead allows the view to move only horizontally.)

Documentation still needs to be updated, and further testing is most definitely appreciated.


jlindgren90:
I am not sure I understand the CodeStyleCheck failure.
```
src/view.c:922: WARNING: switch default: should use break
```
There is a `break` in the `default` case. Is it saying that the `default` case should only `break` and do nothing else?

src/config/rcxml.c
```diff
@@ -1064,11 +1064,14 @@ static struct mouse_combos {
 	{ "Title", "Left", "DoubleClick", "ToggleMaximize", NULL, NULL },
 	{ "TitleBar", "Right", "Click", "Focus", NULL, NULL},
 	{ "TitleBar", "Right", "Click", "Raise", NULL, NULL},
-	{ "TitleBar", "Right", "Click", "ShowMenu", "menu", "client-menu"},
+	{ "Title", "Right", "Click", "ShowMenu", "menu", "client-menu"},
```
Consolatis, line 1067: This needs to be reflected in `docs/rc.xml.all` to have it synced with our built-in defaults.

It is a subtle change that we should mention this in the changelog. Which brings me to the next point, we should likely have some pinned issue to note changes like that so we don't forget to mention it.

include/view.h
```diff
@@ -127,7 +135,8 @@ struct view {
 	bool ssd_titlebar_hidden;
 	enum ssd_preference ssd_preference;
 	bool minimized;
-	bool maximized;
+	bool hmaximized; /* horizontally maximized */
+	bool vmaximized; /* vertically maximized */
```
Consolatis, line 139: Just a random thought: maybe we could change `view->maximized` to `enum view_axis` and add a `VIEW_AXIS_NONE` as `0`? Then `if (!view->maximized)` would still work. Didn't completely think this through though.

Consolatis:
> I am not sure I understand the CodeStyleCheck failure.
> 
> ```
> src/view.c:922: WARNING: switch default: should use break
> ```
> 
> There is a `break` in the `default` case. Is it saying that the `default` case should only `break` and do nothing else?

This seems like a bug in `checkpatch.pl`.
May it is confused about the fall-through case above?
```c
	case VIEW_AXIS_BOTH:
	default:
```
Does the error still happen with that commented out?

Consolatis:
Nice work. I didn't look into the logic changes in full detail yet nor tested it. Will do that later.

include/view.h
jlindgren90: I like this. We could also allow a `direction` of `none` in rc.xml (different from Openbox, but that's fine).

I would suggest that we still explicitly check `view->maximized != VIEW_AXIS_NONE` or `view->maximized == VIEW_AXIS_BOTH` though (depending on the desired semantics).

jlindgren90:
v2: use enum view_axis for view->maximized

CodeStyleCheck failure should also be fixed since the `default` case is only `break` now.

jlindgren90:
Pushed a new commit to update docs. I think this is ready to go from my perspective.

include/view.h
```diff
@@ -29,6 +29,18 @@ enum ssd_preference {
...
+ * Directions in which a view can be maximized. "None" is used
+ * internally to mean "not maximized" but is not valid in rc.xml.
+ * Therefore when parsing rc.xml, "None" means "Invalid".
+ */
+enum view_axis {
+	VIEW_AXIS_NONE = 0,
+	VIEW_AXIS_HORIZONTAL = (1 << 0),
+	VIEW_AXIS_VERTICAL = (1 << 1),
+	VIEW_AXIS_BOTH = (VIEW_AXIS_HORIZONTAL | VIEW_AXIS_VERTICAL),
+};
```
Consolatis, line 42: I like this `VIEW_AXIS_BOTH` variant. This fixes the odd "enum consists of bits and now we have some kind of var which contains a combination of the bits, what type should that var be" issue. Also I didn't know that this actually works (e.g. using enum entries from the same enum). Nice solution.

src/interactive.c
```diff
@@ -71,7 +71,7 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 		cursor_set(seat, LAB_CURSOR_GRAB);
 		break;
 	case LAB_INPUT_STATE_RESIZE:
-		if (view->maximized || view->fullscreen) {
+		if (view->fullscreen || view->maximized != VIEW_AXIS_NONE) {
 			/*
 			 * We don't allow resizing while maximized or
 			 * fullscreen.
```
Consolatis, line 77: I am not sure about this. I'd assume people would actually want to resize in a `maximized != VIEW_AXIS_BOTH` situation.

src/view.c
```diff
@@ -1231,7 +1288,7 @@ view_snap_to_edge(struct view *view, enum view_edge edge, bool store_natural_geo
 		return;
 	}
 
-	if (view->tiled == edge && !view->maximized) {
+	if (view->tiled == edge && view->maximized == VIEW_AXIS_NONE) {
```
Consolatis, line 1291: Personally I'd prefer keeping the `!view->maximized`, makes it easier to read IMHO. But that is just a personal preference and we could keep it like what you have done here.

src/view.c
```diff
@@ -1285,9 +1342,10 @@ view_snap_to_edge(struct view *view, enum view_edge edge, bool store_natural_geo
 		}
 	}
 
-	if (view->maximized) {
+	if (view->maximized != VIEW_AXIS_NONE) {
```
Consolatis, line 1345: Similar here, e.g. we could just keep using `if (view->maximized)` (to catch any kind of maximize). But again, personal preference.

src/snap.c
```diff
@@ -102,7 +102,8 @@ _snap_next_edge(struct view *view, int start_pos, const struct snap_search def,
 	struct view *v;
 	int p = max;
 	for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
-		if (v == view || v->output != output || v->minimized || v->maximized) {
+		if (v == view || v->output != output || v->minimized
+				|| view->maximized != VIEW_AXIS_NONE) {
```
Consolatis, line 106: I am not sure if we want to ignore windows as snap target to that are not fully maximized.
Maybe here a `view->maximized == VIEW_AXIS_BOTH` is more intuitive for a user wanting to align the windows with each other.

src/xdg.c
```diff
@@ -503,12 +504,10 @@ xdg_toplevel_view_map(struct view *view)
...
-		if (!view->fullscreen && requested->fullscreen) {
-			set_fullscreen_from_request(view, requested);
-		} else if (!view->maximized && requested->maximized) {
-			view_maximize(view, true,
-				/*store_natural_geometry*/ true);
-		}
+		set_fullscreen_from_request(view, requested);
+		view_maximize(view, requested->maximized ?
+			VIEW_AXIS_BOTH : VIEW_AXIS_NONE,
+			/*store_natural_geometry*/ true);
```
Consolatis, line 510: This is a logic change, previously we'd do one or the other (or none) and now we do both. I didn't look into the implications of that but thought I should point it out.

src/interactive.c
jlindgren90: For now we could do the same as resizing a tiled window: you can resize it but then it loses the tiled/maximized state.

src/xdg.c
jlindgren90: view_maximize() currently no-ops if fullscreens, I believe.

src/view.c
```diff
@@ -1167,7 +1207,7 @@ void
 view_grow_to_edge(struct view *view, enum view_edge direction)
 {
 	assert(view);
-	if (view->fullscreen || view->maximized) {
+	if (view->fullscreen || view->maximized != VIEW_AXIS_NONE) {
```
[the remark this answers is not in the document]
jlindgren90: I could add a TODO for this, not sure I want to get into implementing one-axis resize logic right now.

src/snap.c
jlindgren90: Sounds good, I'm not really familiar with the snap code.

Consolatis: Oh, also just detected by testing: it should be `v->maximized` rather than `view->maximized`.

jlindgren90: Oops!

src/view.c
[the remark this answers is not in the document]
Consolatis: Well, they already exist. I think we just need to make sure to compare `view->maximized` vs `enum view_edge direction`.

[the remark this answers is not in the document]
Consolatis: I thought of something lilke this:

<details><summary>axis_aligned.diff</summary>

```diff
diff --git a/src/snap.c b/src/snap.c
index 895dc343..698cbf3b 100644
--- a/src/snap.c
+++ b/src/snap.c
@@ -103,7 +103,7 @@ _snap_next_edge(struct view *view, int start_pos, const struct snap_search def,
 	int p = max;
 	for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
 		if (v == view || v->output != output || v->minimized
-				|| view->maximized != VIEW_AXIS_NONE) {
+				|| v->maximized == VIEW_AXIS_BOTH) {
 			continue;
 		}
 
diff --git a/src/view.c b/src/view.c
index e031b574..67a186cc 100644
--- a/src/view.c
+++ b/src/view.c
@@ -1203,11 +1203,30 @@ view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_wind
 	view_move(view, view->pending.x + dx, view->pending.y + dy);
 }
 
+static bool
+axis_aligned(enum view_axis axis, enum view_edge direction)
+{
+	switch(axis) {
+	case VIEW_AXIS_HORIZONTAL:
+		return direction == VIEW_EDGE_LEFT
+			|| direction == VIEW_EDGE_RIGHT;
+	case VIEW_AXIS_VERTICAL:
+		return direction == VIEW_EDGE_UP
+			|| direction == VIEW_EDGE_DOWN;
+	case VIEW_AXIS_BOTH:
+		return true;
+	case VIEW_AXIS_NONE:
+		break;
+	}
+	return false;
+}
+
 void
 view_grow_to_edge(struct view *view, enum view_edge direction)
 {
 	assert(view);
-	if (view->fullscreen || view->maximized != VIEW_AXIS_NONE) {
+	if (view->fullscreen || axis_aligned(view->maximized, direction)) {
+		wlr_log(WLR_ERROR, "not growing");
 		return;
 	}
 	if (!output_is_usable(view->output)) {
@@ -1224,7 +1243,7 @@ void
 view_shrink_to_edge(struct view *view, enum view_edge direction)
 {
 	assert(view);
-	if (view->fullscreen || view->maximized != VIEW_AXIS_NONE) {
+	if (view->fullscreen || axis_aligned(view->maximized, direction)) {
 		return;
 	}
 	if (!output_is_usable(view->output)) {
```

</details>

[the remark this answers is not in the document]
jlindgren90: We would probably also need to update view->natural_geometry for that axis.

Could this be saved for a follow-up PR?

src/interactive.c
jlindgren90: Pushed a [fixup] commit to make resizing work similar to how it does in tiled states.

src/view.c
jlindgren90: I think I would prefer to keep the comparison with NONE or BOTH explicit. To me the implicit bool cast is a bit less clear, since the half-maximized states are treated as "maximized" in some cases but "not maximized" in order.

src/snap.c
jlindgren90: Pushed a [fixup] commit for this.

src/view.c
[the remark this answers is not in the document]
jlindgren90: Added TODOs for now.

jlindgren90:
I pushed a couple of [fixup] commits to address comments (they should be squashed before merge).

johanmalm:
LGTM.
Have tested. Works well with both `xdg-shell` and `XWayland`.
Have compared with openbox. The only difference I noted is that when maximized on one direction only, openbox does not revert to natural geometry on titlebar drag. But I prefer the way you've done it. Seems more intuitive and useful.
Didn't merge because of two outstanding [fixup] commits.

johanmalm:
...and just to say, nicely implemented!

jlindgren90:
Everything is squashed into one commit now. Thanks for the reviews!

src/view.c
Consolatis: > To me the implicit bool cast is a bit less clear, since the half-maximized states are treated as "maximized" in some cases but "not maximized" in others.

Understood. What about `!view->maximized`?

jlindgren90: I personally don't like using a Boolean operator on an enum value. The equality check is clearer to me.

Consolatis: Alright, guess we have different preferences here which is totally fine :)

Consolatis:
Had a look through the whole diff again and didn't spot anything suspicious. As such: LGTM.
Thanks for your work on this feature :)

Issue #1208, opened 2023-10-28
Keybindings stop working after laptop lid close/open


jlindgren90:
A few times, I've seen all keybindings in labwc stop working. I've finally figured out that it is triggered after closing and then reopening the laptop lid. With some extra logging, I've determined that the laptop is generating a keypress with keycode 0x8f / keysym 0x1008ff2b (XF86XK_WakeUp) but no matching key release. This causes the `key_state_nr_pressed_keys() > 1` check in `handle_compositor_keybindings` to trigger and no keybindings to be handled from that point on.

I'm not sure of the best way to solve this. The "no other keys pressed" check seems a bit prone to situations where odd keycodes like this get stuck. I'm wondering if that check is really necessary. If it is, maybe a timeout could be added to check for keys that are stuck for an unreasonably long time?

Consolatis:
We've talked about a timer based safety net before, it might be an option. But we should ensure that it won't just delete whatever is in there at the point it triggers (maybe have the _set_pressed()` function automatically add a timestamp?). Another idea was to maybe empty the set on `Reconfigure`.

The check is necessary so we don't trigger for example a `A-z' keybind while pressing `A-z-q`.
There is already a `is_layout_switch` check in there that prevents (almost) the same situation from happening when switching keyboard layouts.

For your specific case we could maybe move the `is_layout_switch` check to its own function where it then then also checks for
- `XKB_KEY_XF86WakeUp` and possibly also for
- `XKB_KEY_XF86Suspend`
- `XKB_KEY_XF86Hibernate`

johanmalm:
In terms of context, it's pretty important not to send release events to clients if the press event was absorbed (which the `key-state.c` logic is primarily there to prevent).

Maybe we ought to soften the `key_state_nr_pressed_keys() > 1` check. It's there to make sure we don't press more than one 'normal' key (currently considered anything that isn't a modifier + `XKB_KEY_ISO_Group_Next`). Could we instead consider a 'normal' key to be a `Latin 1` key (range `0020`..`00FF`) and thus filter out `WakeUp`, `Suspend`, `Hibernate`, `ISO_Next_Group`, and anything else we haven't come across yet. Or we could make it a bit more sophisticated to include KP (keypad) keys, Home, End, etc.

https://github.com/xkbcommon/libxkbcommon/blob/6073565903488cb5b9a8d37fdc4a7c2f9d7ad04d/include/xkbcommon/xkbcommon-keysyms.h#L559

[...]

PR #1212, opened 2023-10-29
key-state: make nr-pressed-keys check more accurate

johanmalm:
@jlindgren90 @Consolatis Maybe not the tidiest, but just a quick proof of concept to go with our discussion in issue #1208

### DIFF
--- a/src/key-state.c
+++ b/src/key-state.c
@@ -103,8 +103,38 @@ key_state_nr_bound_keys(void)
 	return bound.nr_keys;
 }
 
-int
-key_state_nr_pressed_keys(void)
+/*
+ * Reference:
+ * https://github.com/xkbcommon/libxkbcommon/blob/HEAD/include/xkbcommon/xkbcommon-keysyms.h
+ */
+bool
+key_state_multiple_normal_keys_pressed(void)
 {
-	return pressed.nr_keys;
+	if (pressed.nr_keys < 2) {
+		return false;
+	}
+
+	/*
+	 * If the pressed array contains multiple keys, we have to analyze a bit
+	 * further because some keys like those listed below do not always
+	 * receive release events.
+	 *   - XKB_KEY_ISO_Group_Next
+	 *   - XKB_KEY_XF86WakeUp
+	 *   - XKB_KEY_XF86Suspend
+	 *   - XKB_KEY_XF86Hibernate
+	 */
+	size_t count = 0;
+	for (int i = 0; i < pressed.nr_keys; ++i) {
+		/* Convert from udev event to xkbcommon code */
+		uint32_t keycode = pressed.keys[i] + 8;
+		if (keycode >= 0x1000000) {
+			/* XFree86 vendor specific keysyms */
+			continue;
+		} else if (keycode >= 0xFE00 && keycode <= 0xFEFF) {
+			/* Extension function and modifier keys */
+			continue;
+		}
+		++count;
+	}
+	return count > 1;
 }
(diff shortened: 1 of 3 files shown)


src/key-state.c
```diff
@@ -103,8 +103,38 @@ key_state_nr_bound_keys(void)
...
-	return pressed.nr_keys;
+	if (pressed.nr_keys < 2) {
+		return false;
+	}
+
+	/*
+	 * If the pressed array contains multiple keys, we have to analyze a bit
+	 * further because some keys like those listed below do not always
+	 * receive release events.
+	 *   - XKB_KEY_ISO_Group_Next
```
Consolatis, line 121: The key that sometimes is identified as `XKB_KEY_ISO_Group_Next` should always receive a release event.
This should be fixed by always removing the key on release events.

src/key-state.c
```diff
@@ -103,8 +103,38 @@ key_state_nr_bound_keys(void)
...
+	 *   - XKB_KEY_XF86WakeUp
+	 *   - XKB_KEY_XF86Suspend
+	 *   - XKB_KEY_XF86Hibernate
+	 */
+	size_t count = 0;
+	for (int i = 0; i < pressed.nr_keys; ++i) {
+		/* Convert from udev event to xkbcommon code */
+		uint32_t keycode = pressed.keys[i] + 8;
+		if (keycode >= 0x1000000) {
```
Consolatis, line 130: AFAIR we are storing the actual keycode here, not the keysyms.
So if we'd want to decide based of the keysyms here we'd have to translate them first.

Consolatis:
While this approach could fix the `keybinds-stuck` issue (for the sym ranges used) wouldn't it be an issue for the applications as well? AFAIR we are using the array (unless bound later) to send the initial state to clients when they get focus. Thus they see a `press` but never a release. I am not entirely sure here though.

jlindgren90:
I still think it's worth at least considering dropping the check entirely. Having one (non-modifier) key still held doesn't prevent further key presses from being processed in most applications I'm familiar with. E.g. I tested typing "qw" into a terminal earlier, pressing "w" before releasing "q", and both letters appeared. I wouldn't personally expect the "w" to be ignored in that case. Not sure why labwc keybindings should be any different.

jlindgren90:
Keeping the general "ignore multiple keys" rule while making exceptions for particular keys, sounds like it could turn into a game of whack-a-mole. My laptop has a WakeUp "key" that gets stuck. Some other system may have a different quirk and a different key gets stuck. Maybe someone spilled soda on their keyboard and one of the keys really does stick for a couple seconds - do we drop other key presses in that case?

jlindgren90:
>  I guess the 2nd issue could be fixed by just storing the current (event->keycode) as bound (and removing it from 'pressed') rather than all the keys in the pressed set.

This makes sense to me -- we should not treat an unrelated pressed/stuck key as bound if it was not actually part of the keybinding.

> But that would make the first issue worse by sending the 'wakeup key pressed' (or any other stuck key) to all applications getting focus until restarting labwc.

I would expect applications to just ignore that key. Do we know of applications that don't?

I agree that everything we do here is managing symptoms -- the root cause in this case is probably a weird/buggy ACPI implementation resulting in that keypress without release.

johanmalm:
Hmm.... some thinking time needed :smile:

Consolatis:
Here might be another one:
- #1223

jlindgren90:
> I guess the 2nd issue could be fixed by just storing the current (event->keycode) as bound (and removing it from 'pressed') rather than all the keys in the pressed set.

This part implemented in #1226.

PR #1215, opened 2023-10-31
view.h: changed the type of tiled in struct view to enum view_edge

domo141:
The type enum view_edge used to be defined in a .c file, so a structure member 'tiled' in struct view had to be defined to use another type.

Later (2023-08-02, commit 1ee8715) the definition of enum view_edge was moved to view.h, so now 'tiled' can be defined to use that type.


johanmalm:
Thanks!

PR #1216, opened 2023-10-31
Support pipemenus

johanmalm:
http://openbox.org/wiki/Help:Menus#Pipe_menus

Before merge
- [x] Remove logging marked `XXX`
- [x] Decide whether or not to insist on an `id` attribute in `<menu label="" id="" execute=""/>`. ~~Openbox requires it, but it isn't actually needed in my opinion.~~ Decided to require `id` to stay consistent with Openbox and to make pipemenu-generators responsible for unique id (rather than labwc having to uniquify them which might become quite hard).

There is lots of stuff that needs to be done with menus in general (e.g. making sure we don't have duplicate menu ids, positioning big menus better within usable areas, handling `nr_items * item_height > usable_area->height`)... but let's leave that for separate PRs.

### DIFF
--- a/src/common/spawn.c
+++ b/src/common/spawn.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #define _POSIX_C_SOURCE 200809L
 #include <assert.h>
+#include <fcntl.h>
 #include <glib.h>
 #include <signal.h>
 #include <stdio.h>
@@ -10,6 +11,26 @@
 #include <wlr/util/log.h>
 #include "common/spawn.h"
 #include "common/fd_util.h"
+
+static void
+reset_signals_and_limits(void)
+{
+	restore_nofile_limit();
+
+	sigset_t set;
+	sigemptyset(&set);
+	sigprocmask(SIG_SETMASK, &set, NULL);
+
+	/* Restore ignored signals */
+	signal(SIGPIPE, SIG_DFL);
+}
+
+static void
+set_cloexec(int fd)
+{
+	int flags = fcntl(fd, F_GETFD);
+	fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
+}
 
 void
 spawn_async_no_shell(char const *command)
@@ -39,14 +60,9 @@ spawn_async_no_shell(char const *command)
 		wlr_log(WLR_ERROR, "unable to fork()");
 		goto out;
 	case 0:
-		restore_nofile_limit();
+		reset_signals_and_limits();
 
 		setsid();
-		sigset_t set;
-		sigemptyset(&set);
-		sigprocmask(SIG_SETMASK, &set, NULL);
-		/* Restore ignored signals */
-		signal(SIGPIPE, SIG_DFL);
 		grandchild = fork();
 		if (grandchild == 0) {
 			execvp(argv[0], argv);
@@ -63,3 +79,76 @@ out:
 	g_strfreev(argv);
 }
 
+pid_t
+spawn_piped(const char *command, int *pipe_fd)
+{
+	assert(command);
+
+	int pipe_rw[2];
+	if (pipe(pipe_rw) != 0) {
+		wlr_log(WLR_ERROR, "unable to pipe()");
+		return -1;
+	}
+
+	pid_t pid = fork();
+	if (pid < 0) {
+		close(pipe_rw[0]);
+		close(pipe_rw[1]);
+		wlr_log(WLR_ERROR, "unable to fork()");
+		return pid;
+	}
+
+	if (pid == 0) {
+		/* child */
+		reset_signals_and_limits();
+
+		/*
+		 * replace stdin and stderr with /dev/null
+		 * and stdout with the write end of the pipe
+		 */
+		dup2(pipe_rw[1], STDOUT_FILENO);
+		close(pipe_rw[0]);
+		close(pipe_rw[1]);
+
+		int dev_null = open("/dev/null", O_RDWR);
+		if (dev_null < 0) {
+			wlr_log_errno(WLR_ERROR, "opening /dev/null failed");
+			/*
+			 * Just close stdin and stderr and
+			 * hope $command can deal with that.
+			 */
+			close(STDIN_FILENO);
+			close(STDERR_FILENO);
+		} else {
+			dup2(dev_null, STDIN_FILENO);
+			dup2(dev_null, STDERR_FILENO);
+			close(dev_null);
+		}
+
+		execl("/bin/sh", "sh", "-c", command, NULL);
+		/*
+		 * Our stderr points to /dev/null or is closed
+		 * at this point so logging is pretty useless.
+		 */
+		_exit(1);
+	}
+
+	/* labwc */
+	close(pipe_rw[1]);
+
+	/*
+	 * Prevent leaking the read end of the pipe to further
+	 * children forked during the lifetime of the descriptor.
+	 */
+	set_cloexec(pipe_rw[0]);
+
+	*pipe_fd = pipe_rw[0];
+	return pid;
+}
+
+void
+spawn_piped_close(pid_t pid, int pipe_fd)
+{
+	close(pipe_fd);
+	/* waitpid() is done in a generic SIGCHLD handler in src/server.c */
+}
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -593,6 +593,25 @@ menu_get_full_width(struct menu *menu)
 	return width + max_child_width;
 }
 
+static struct wlr_box
+get_submenu_position(struct menuitem *item, enum menu_align align)
+{
+	struct wlr_box pos = { 0 };
+	struct menu *menu = item->parent;
+	struct theme *theme = menu->server->theme;
+	int lx = menu->scene_tree->node.x;
+	int ly = menu->scene_tree->node.y;
+
+	if (align & LAB_MENU_OPEN_RIGHT) {
+		pos.x = lx + menu->size.width - theme->menu_overlap_x;
+	} else {
+		pos.x = lx;
+	}
+	int rel_y = item->tree->node.y;
+	pos.y = ly + rel_y - theme->menu_overlap_y;
+	return pos;
+}
+
 static void
 menu_configure(struct menu *menu, int lx, int ly, enum menu_align align)
 {
@@ -650,14 +669,8 @@ menu_configure(struct menu *menu, int lx, int ly, enum menu_align align)
 		if (!item->submenu) {
 			continue;
 		}
-		if (align & LAB_MENU_OPEN_RIGHT) {
-			new_lx = lx + menu->size.width - theme->menu_overlap_x;
-		} else {
-			new_lx = lx;
-		}
-		rel_y = item->tree->node.y;
-		new_ly = ly + rel_y - theme->menu_overlap_y;
-		menu_configure(item->submenu, new_lx, new_ly, align);
+		struct wlr_box pos = get_submenu_position(item, align);
+		menu_configure(item->submenu, pos.x, pos.y, align);
 	}
 }
 
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -874,6 +874,13 @@ menu_process_item_selection(struct menuitem *item)
 {
 	assert(item);
 
+	/* Do not keep selecting the same item */
+	static struct menuitem *last;
+	if (item == last) {
+		return;
+	}
+	last = item;
+
 	if (!item->selectable) {
 		return;
 	}
@@ -1042,12 +1049,6 @@ menu_process_cursor_motion(struct wlr_scene_node *node)
 {
 	assert(node && node->data);
 	struct menuitem *item = node_menuitem_from_node(node);
-
-	if (item->selectable && node == &item->selected.tree->node) {
-		/* We are on an already selected item */
-		return;
-	}
-
 	menu_process_item_selection(item);
 }
 
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -842,7 +842,7 @@ menu_close(struct menu *menu)
 }
 
 void
-menu_open(struct menu *menu, int x, int y)
+menu_open_root(struct menu *menu, int x, int y)
 {
 	assert(menu);
 	if (menu->server->menu_current) {
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <libxml/parser.h>
 #include <libxml/tree.h>
+#include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -19,11 +20,15 @@
 #include "common/nodename.h"
 #include "common/scaled_font_buffer.h"
 #include "common/scene-helpers.h"
+#include "common/spawn.h"
 #include "common/string-helpers.h"
 #include "labwc.h"
 #include "menu/menu.h"
 #include "node.h"
 #include "theme.h"
+
+#define PIPEMENU_MAX_BUF_SIZE 1048576  /* 1 MiB */
+#define PIPEMENU_TIMEOUT_IN_MS 4000    /* 4 seconds */
 
 /* state-machine variables for processing <item></item> */
 static bool in_item;
@@ -33,11 +38,30 @@ static struct action *current_item_action;
 static int menu_level;
 static struct menu *current_menu;
 
+static bool waiting_for_pipe_menu;
+static struct menuitem *selected_item;
+
 /* TODO: split this whole file into parser.c and actions.c*/
+
+static bool
+is_unique_id(struct server *server, const char *id)
+{
+	struct menu *menu;
+	wl_list_for_each(menu, &server->menus, link) {
+		if (!strcmp(menu->id, id)) {
+			return false;
+		}
+	}
+	return true;
+}
 
 static struct menu *
 menu_create(struct server *server, const char *id, const char *label)
 {
+	if (!is_unique_id(server, id)) {
+		wlr_log(WLR_ERROR, "menu id %s already exists", id);
+	}
+
 	struct menu *menu = znew(*menu);
 	wl_list_append(&server->menus, &menu->link);
 
@@ -46,6 +70,7 @@ menu_create(struct server *server, const char *id, const char *label)
 	menu->label = xstrdup(label ? label : id);
 	menu->parent = current_menu;
 	menu->server = server;
+	menu->is_pipemenu = waiting_for_pipe_menu;
 	menu->size.width = server->theme->menu_min_width;
 	/* menu->size.height will be kept up to date by adding items */
 	menu->scene_tree = wlr_scene_tree_create(server->menu_tree);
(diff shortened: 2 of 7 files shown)


johanmalm:
Try it with something like this:

```
<menu id="root-menu" label="Applications">
  <item label="File Manager"><action name="Execute"><command><![CDATA[xdg-open .]]></command></action></item>
  <item label="Terminal"><action name="Execute"><command><![CDATA[xterm]]></command></action></item>
  <item label="Web Browser"><action name="Execute"><command><![CDATA[xdg-open http://]]></command></action></item>
  <item label="Run command"><action name="Execute"><command><![CDATA[gmrun]]></command></action></item>
  <separator/>
  <menu label="labwc-menu-generator -p" execute="labwc-menu-generator"/>
  <menu label="obmenu-generator" execute="obmenu-generator"/>
  <menu label="bl-places-pipemenu" execute="bl-places-pipemenu"/>
</menu>
```

https://github.com/BunsenLabs/bunsen-pipemenus/blob/beryllium/bin/bl-places-pipemenu
https://github.com/labwc/labwc-menu-generator/blob/d7c81071f8b121ef83da32ae3fa16155d1a2ced9/main.c#L36
https://github.com/trizen/obmenu-generator

src/menu/menu.c
```diff
@@ -341,7 +347,11 @@ static bool
 nodename_supports_cdata(char *nodename)
 {
 	return match_glob("command.action.item.*menu.openbox_menu", nodename)
-		|| match_glob("execute.action.item.*menu.openbox_menu", nodename);
+		|| match_glob("execute.action.item.*menu.openbox_menu", nodename)
+		|| match_glob("command.action.item.openbox_pipe_menu", nodename)
+		|| match_glob("execute.action.item.openbox_pipe_menu", nodename)
+		|| match_glob("command.action.item.*menu.openbox_pipe_menu", nodename)
+		|| match_glob("execute.action.item.*menu.openbox_pipe_menu", nodename);
```
Consolatis, line 354: Could we use `!strncmp(nodename, "execute.action", strlen("execute.action"))` here and only check if nodename starts with `execute.action` and `command.action`?

src/menu/menu.c
```diff
@@ -761,16 +789,47 @@ menu_init(struct server *server)
...
+	wl_list_for_each(iter, &menu->server->menus, link) {
+		struct menuitem *item;
+		wl_list_for_each(item, &iter->menuitems, link) {
+			if (item->submenu == menu) {
+				item->submenu = NULL;
+				return;
+			}
+		}
+	}
+}
```
Consolatis, line 805: Can we have multiple items point to the same submenu? In that case we should likely not early return here.

src/menu/menu.c
```diff
@@ -843,22 +920,89 @@ menu_open(struct menu *menu, int x, int y)
...
 menu_process_item_selection(struct menuitem *item)
 {
 	assert(item);
 
+	/* Do not keep selecting the same item */
+	static struct menuitem *last;
+	if (item == last) {
+		return;
+	}
+	last = item;
```
Consolatis, line 961: What if we select an item, close the menu, open the menu and try to select the same item again here?
Wouldn't the static `last` cause an early return in that situation?

src/menu/menu.c
johanmalm: Very good point. Done.

src/menu/menu.c
```diff
@@ -813,23 +878,35 @@ close_all_submenus(struct menu *menu)
...
-	menu_set_selection(menu, NULL);
-	if (menu->selection.menu) {
-		menu_close(menu->selection.menu);
-		menu->selection.menu = NULL;
+	struct server *server = menu->server;
+	wlr_log(WLR_DEBUG, "number of menus before close=%d", wl_list_length(&server->menus));
+	_close(menu);
+	if (menu->is_pipemenu) {
+		menu_free_from(menu->server, menu);
 	}
```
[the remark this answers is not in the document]
johanmalm: I'll add a check on creation to make sure we don't add duplicates. One for tomorrow.

src/menu/menu.c
johanmalm: No. Don't think so.

src/menu/menu.c
```diff
@@ -1008,12 +1153,6 @@ menu_process_cursor_motion(struct wlr_scene_node *node)
 {
 	assert(node && node->data);
 	struct menuitem *item = node_menuitem_from_node(node);
-
-	if (item->selectable && node == &item->selected.tree->node) {
-		/* We are on an already selected item */
-		return;
-	}
```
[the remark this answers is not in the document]
johanmalm: This check didn't work.
I replaced it with a new check in a dedicated pre-pipemenu commit.
https://github.com/labwc/labwc/pull/1216/commits/8b96acce260d1e244d19e7a4353244678b373292

Conan-Kudo:
Any chance this will make it for version 0.7.0?

johanmalm:
> Any chance this will make it for version 0.7.0?

Sorry. It didn't quite fit in.

It's 90% there an works well, but needs some more testing + reviews.
If you're happy to cherry-pick the commits, you can run it as is :smile:

src/menu/menu.c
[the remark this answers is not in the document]
johanmalm: Have made `id` a requirement. And also check for unique `id`s on creation. So... done.

johanmalm:
> I've just given it a try and with `labwc-menu-generator` it works great. Plain `obmenu-generator` does too, but `obmenu-generator -p` causes labwc to completely freeze when I try to open the menu and I have to manually kill it.

Thanks for testing. Regarding `obmenu-generator -p`, we might need gracefully handle empty buffers, etc. Not really tried to feed it bad data.

src/common/spawn.c
```diff
@@ -63,3 +79,87 @@ spawn_async_no_shell(char const *command)
...
+
+	/* labwc */
+
+	/*
+	 * Prevent leaking the read end of the pipe to further
+	 * children forked during the lifetime of the descriptor.
+	 */
+	set_cloexec(pipe_rw[0]);
+
+	close(pipe_rw[1]);
```
Consolatis, line 143: move after `/* labwc */ comment

src/common/spawn.c
```diff
@@ -63,3 +79,87 @@ spawn_async_no_shell(char const *command)
...
+			close(STDERR_FILENO);
+		} else {
+			dup2(dev_null, STDIN_FILENO);
+			dup2(dev_null, STDERR_FILENO);
+			close(dev_null);
+		}
+
+		/* close all remaining fds */
+		close(pipe_rw[0]);
+		close(pipe_rw[1]);
```
Consolatis, line 128: Move after `dup2(, STDOUT_FILENO)`

src/common/spawn.c
```diff
@@ -63,3 +79,87 @@ spawn_async_no_shell(char const *command)
...
+	if (waitpid(pid, &details, options) == -1) {
+		perror("waitpid failed");
+		return;
+	}
+
+	if (WIFEXITED(details)) {
+		wlr_log(WLR_DEBUG, "child exited with %d", WEXITSTATUS(details));
+	} else {
+		wlr_log(WLR_ERROR, "spawned child didn't terminate normally");
+	}
```
Consolatis, line 164: Read signal

src/common/spawn.c
```diff
@@ -63,3 +79,87 @@ spawn_async_no_shell(char const *command)
...
+spawn_piped_close(pid_t pid, int pipe_fd)
+{
+	close(pipe_fd);
+
+	int details = 0;
+	int options = 0;
+	if (waitpid(pid, &details, options) == -1) {
+		perror("waitpid failed");
+		return;
+	}
```
Consolatis, line 158: evaluate using non-blocking waitpid and a fallback labwc wide SIGCHLD handler within the wl event loop

src/common/spawn.c
```diff
@@ -63,3 +79,87 @@ spawn_async_no_shell(char const *command)
...
+			dup2(dev_null, STDIN_FILENO);
+			dup2(dev_null, STDERR_FILENO);
+			close(dev_null);
+		}
+
+		/* close all remaining fds */
+		close(pipe_rw[0]);
+		close(pipe_rw[1]);
+
+		execl("/bin/sh", "sh", "-c", command, NULL);
```
Consolatis, line 130: Should we use `execl` with `sh -c` here? Or split args manually like in `spawn_async_no_shell()`?

src/menu/menu.c
```diff
@@ -19,12 +20,15 @@
...
 #include "common/scaled_font_buffer.h"
 #include "common/scene-helpers.h"
+#include "common/spawn.h"
 #include "common/string-helpers.h"
 #include "labwc.h"
 #include "menu/menu.h"
 #include "node.h"
 #include "theme.h"
 
+#define PIPEMENU_TIMEOUT_IN_MS 4000
```
Consolatis, line 30: Should we make this user controllable, e.g. via some kind of config (or theme) option?

src/menu/menu.c
```diff
@@ -856,22 +970,200 @@ menu_open(struct menu *menu, int x, int y)
...
+	struct pipe_context *ctx = _ctx;
+	kill(ctx->pid, SIGTERM);
+	return 0;
+}
+
+static int
+handle_pipemenu_readable(int fd, uint32_t mask, void *_ctx)
+{
+	struct pipe_context *ctx = _ctx;
+	char data[4096];
```
Consolatis, line 995: Use `4097` to read a complete page + terminating NULL byte

src/menu/menu.c
```diff
@@ -856,22 +970,200 @@ menu_open(struct menu *menu, int x, int y)
...
+
+	/*
+	 * Pipemenus do not contain a toplevel <menu> element so we have to
+	 * create that first `struct menu`.
+	 */
+	menu_level++;
+	current_menu = menu_create(ctx->server, ctx->item->id, NULL);
+	current_menu->is_pipemenu = true;
+	wlr_log(WLR_INFO, "parse pipemenu '%s'", ctx->item->execute);
+	parse_buf(ctx->server, &ctx->buf);
```
Consolatis, line 1030: handle parse_buf returning `false`

src/menu/menu.c
```diff
@@ -856,22 +970,200 @@ menu_open(struct menu *menu, int x, int y)
...
+		perror("Failed to read from pipemenu");
+		wlr_log(WLR_ERROR, "Failed to read data for pipe menu");
+		goto clean_up;
+	}
+	wlr_log(WLR_INFO, "read %li bytes of data", size);
+	if (size) {
+		data[size] = '\0';
+		buf_add(&ctx->buf, data);
+		return 0;
+	}
```
Consolatis, line 1013: Should we limit the buffer size (`ctx->buf.len`) to something like 1 MB and when reached kill the script? Should help against pipemenu scripts accidentally filling up all the ram.

src/common/spawn.c
Consolatis: Uses non-blocking now but labwc wide SIGCHLD handler still missing.

Consolatis:
Here is something fun to test with: `browse.sh`
```bash
#!/bin/bash

MAX_LEVEL=2

ICON_FILE='🗋'
ICON_DIR='🖿'
ICON_PARENT='🗁'
self=$(readlink -f "$0")

list_dirs() {
	local dir=$1
	local level=$2

	#dirs
	for file in "$dir"/*; do
		_basename=$(basename "$file")
		test "$_basename" == "*" && continue
		test -d "$file" || continue

		execute=""
		label="$ICON_DIR $_basename"
		if test $level -eq $MAX_LEVEL; then
			execute=" execute=\"$self '$file' ${MAX_LEVEL}\""
		fi
		printf "%${level}s" ""
		echo "<menu id=\"$file\" label=\"$ICON_DIR $_basename\"$execute>"
		test $level -lt $MAX_LEVEL && list_dirs "$file" $((level + 1))
		echo "</menu>"
	done

	#files
	for file in "$dir"/*; do
		_basename=$(basename "$file")
		test "$_basename" == "*" && continue
		test -d "$file" && continue
		printf "%$(( level * 2 ))s" ""
		echo "<item label=\"$ICON_FILE $_basename\" />"
		# echo could add a <action name="Execute" command="xdg-open '$file'" /> here
	done
}

browse() {
	echo "<openbox_pipe_menu>"
	list_dirs "$1" 0
	echo "</openbox_pipe_menu>"
}

if test $# -lt 1; then
	exit 1
fi

test -d "$1" || exit 2
test -n "$2" && MAX_LEVEL=$2
#browse "$1"
bar=$(browse "$1")
echo "$bar"
```
Can be called like `browse.sh directory [max_depths]`, e.g. `browse.sh ~` for a default max_depths of 2.
It is basically a file browser via pipemenus, by default it fetches up to two levels deep and for directories deeper than that it uses another pipe menu entry to call itself again. Do \*not\* call `browse.sh /` though, it will take \*forever\* with a depth of 2.

Seems to be pretty good to produce random issues with our current menu implementation. We should also think about some caching behavior at some point but that is not important for the initial implementation. (e.g. rather than calling the script every time we select the entry, cache the old menu contents up to $timeout, potentially allow clients via a `<openbox_pipe_menu>` attribute to decide on that timespan) .

src/menu/menu.c
```diff
@@ -774,16 +836,50 @@ menu_init(struct server *server)
...
 {
+	bool destroying = !from;
 	struct menu *menu, *tmp_menu;
 	wl_list_for_each_safe(menu, tmp_menu, &server->menus, link) {
+		if (menu == from) {
+			destroying = true;
+		}
+		if (!destroying) {
+			continue;
+		}
```
Consolatis, line 872:
```
if (from && menu != from) {
	continue;
}
```

src/menu/menu.c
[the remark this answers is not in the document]
Consolatis: I got that wrong initially, `menu_free_from()` only frees that specific menu entry, I thought it would free the whole chain. Slightly confusing naming of the function.

src/menu/menu.c
```diff
@@ -856,22 +970,203 @@ menu_open(struct menu *menu, int x, int y)
...
+	/* We are on an item that has new focus */
 	menu_set_selection(item->parent, item);
 	if (item->parent->selection.menu) {
 		/* Close old submenu tree */
 		menu_close(item->parent->selection.menu);
 	}
 
+	/* Pipemenu */
+	if (item->execute) {
+		parse_pipemenu(item);
```
Consolatis, line 1140: I don't think the following code should be here as we didn't do anything at this point. E.g. we only spawned the subprocess and waiting for its stdout to return something later.

johanmalm:
This seems to fix the occasional duplicate menus / ids.

```
index 2069b654..6eb8cf81 100644
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -988,6 +988,8 @@ handle_pipemenu_timeout(void *_ctx)
 	return 0;
 }
 
+static bool pipemenu_locked_enabled;
+
 static int
 handle_pipemenu_readable(int fd, uint32_t mask, void *_ctx)
 {
@@ -1079,6 +1081,7 @@ clean_up:
 	spawn_piped_close(ctx->pid, ctx->pipe_fd);
 	free(ctx->buf.buf);
 	free(ctx);
+	pipemenu_locked_enabled = false;
 	return 0;
 }
 
@@ -1090,10 +1093,13 @@ parse_pipemenu(struct menuitem *item)
 		return;
 	}
 
+	pipemenu_locked_enabled = true;
+
 	int pipe_fd = 0;
 	pid_t pid = spawn_piped(item->execute, &pipe_fd);
 	if (pid <= 0) {
 		wlr_log(WLR_ERROR, "Failed to spawn pipe menu process %s", item->execute);
+		pipemenu_locked_enabled = false;
 		return;
 	}
 
@@ -1117,6 +1123,10 @@ menu_process_item_selection(struct menuitem *item)
 {
 	assert(item);
 
+	if (pipemenu_locked_enabled) {
+		wlr_log(WLR_DEBUG, "cannot process items as pipemenu lock is enabled");
+		return;
+	}
 	/* Do not keep selecting the same item */
 	static struct menuitem *last;
 	if (item == last) {
```

Consolatis:
As discussed via IRC
- added the locking suggestion above.
- Also slightly refactored to use a dedicated function for the menu specific parts of pipe menus (`create_pipe_menu()`)
- hopefully made it easier to read
- we should now also handle invalid XML gracefully but I didn't actually test that yet
- we now handle the case where the parent menu of a pipe menu is already closed (by not creating the pipe menu)

Issues seen while testing:
- [ ] `menu_process_item_selection()` fails when closing all menus and then reopening them, the last selected item can not be selected anymore until something else gets selected first
- [ ] we still get duplicated-id warnings from time to time, didn't yet investigate why but I expect what is happening is that there is some race between showing a pipemenu and closing its parent, we may not destroy the new pipemenu in that case and thus when showing the selecting the same item for the pipemenu later we get the duplicate id. `browse.sh / 0` is a good test for this.
- [ ] arrows for pipe menu items are not adjusted correctly until the item is actually selected and triggered the pipe menu

src/menu/menu.c
Consolatis: removed via the latest patch.

src/menu/menu.c
Consolatis: I've now changes this to 2 4k pages + NULL byte.
Basically this controls how much data we read at once from the pipe, it is a trade-off with (temporary) stack memory usage vs minimal increase in latency when we have to wait for the next fd readable event. Especially with app menu pipe menus in mind, we could technically increase this up to 64k which is what a pipe can hold internally by default as far as I remember.

I would be interesting to know how big usual application menu pipe menus are and if they write everything at once or in shorter bursts. Can be seen from the log entries like `00:38:19.699 [../src/menu/menu.c:1103] read 1073 bytes of data`.

src/common/spawn.c
johanmalm: I think I'm fine with `/bin/sh`. It's what `popen()` does anyway IIRC.

src/menu/menu.c
johanmalm: Suggest start hard-coded for now and add if there is a request for it.

src/menu/menu.c
johanmalm: I'm fine with two pages. There is the BUFSIZ macro - which I think it meant to give an optimum size (4096 generally I believe), but it probably wasn't tuned for our use-case!!

I suspect most pipemenus produce way less than 4096 anyway.

The biggest ones I've seen are similar to `browse.sh` and one that displays all keybinds (by parsing `rc.xml`). But again, even the keybind one shouldn't be that big.

src/menu/menu.c
johanmalm: Yes. Agree. 
Suspect 100k would be plenty!

Consolatis:
> It actually crashes for me atm when trying to open an app via pipe menu.

Whoops, should have tested with actual actions as well. Should be fixed by the latest commit.

src/common/spawn.c
Consolatis: added the labwc wide SIGCHLD handler.

johanmalm:
Intend to merge in a day or so.

Testers - if you want to try to break it, now is a good time :smile:

johanmalm:
Just did an interactive-rebase HEAD~3 to fixup the last two.

PR #1218, opened 2023-11-02
cursor: allow re-focusing xwayland-unmanaged surfaces

### LINKED ISSUE #1217: dmenu dialog loses keyboard focus
When I launch dmenu (for example, by running `ls /home/bruno | dmenu` in a terminal emulator), the dmenu dialog appears but I need to be careful not to click anywhere away from the dialog. If I accidentally click off of the dialog, it becomes impossible to interact with the dmenu dialog again, even though it is still visible. I think the dmenu dialog is escaping labwc's detection, because it has neither window decorations nor does is appear in the alt+tab menu.
 
dmenu seems popular, which makes me think that this strange behavior is because of me (doing something wrong--e.g., not understanding how to make it detectable by labwc). Do you have any idea how to select dmenu dialog again if I accidentally click somewhere else on my desktop before making a selection?

P.S. If this is a dmenu issue and not a labwc issue, no problem--just let me know and I'll ask about this elsewhere.

jlindgren90:
We already allow some xwayland-unmanaged surfaces to take focus on map, if indicated by `wlr_xwayland_or_surface_wants_focus()`. But once these surfaces lose focus, they never regain it again.

Add `desktop_focus_view_or_surface()` and call it in the appropriate places to allow these views to regain focus in the usual ways (e.g. clicking on them or focus-follows-mouse).

Fixes #1217.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -479,8 +479,9 @@ process_cursor_motion(struct server *server, uint32_t time)
 		dnd_icons_move(seat, seat->cursor->x, seat->cursor->y);
 	}
 
-	if (ctx.view && rc.focus_follow_mouse) {
-		desktop_focus_view(ctx.view, rc.raise_on_focus);
+	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse) {
+		desktop_focus_view_or_surface(seat, ctx.view, ctx.surface,
+			rc.raise_on_focus);
 	}
 
 	struct mousebind *mousebind;
@@ -517,11 +518,12 @@ _cursor_update_focus(struct server *server)
 	/* Focus surface under cursor if it isn't already focused */
 	struct cursor_context ctx = get_cursor_context(server);
 
-	if (ctx.view && rc.focus_follow_mouse
+	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse
 			&& !rc.focus_follow_mouse_requires_movement
 			&& !server->osd_state.cycle_view) {
 		/* Prevents changing keyboard focus during A-Tab */
-		desktop_focus_view(ctx.view, rc.raise_on_focus);
+		desktop_focus_view_or_surface(&server->seat, ctx.view,
+			ctx.surface, rc.raise_on_focus);
 	}
 
 	cursor_update_common(server, &ctx, msec(&now),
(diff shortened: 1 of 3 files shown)


jlindgren90:
Side note: it looks like focus-follows-mouse probably does not work for layer-shell surfaces. Moving the layer-shell focus logic to `desktop_focus_view_or_surface()` also should make that work. I didn't make that change since I don't currently run anything using layer-shell and didn't want to accidentally break anything.

Consolatis:
Hm, it indeed seems like we completely missed refocusing non-view surfaces on click, nice spot.

We had issues with focusing subsurfaces breaking drag and drop before (although I don't think this change will cause that as we only (re)focus the view itself if supplied as argument), might still be worth to verify that though.

src/input/cursor.c
```diff
@@ -908,13 +910,22 @@ cursor_button_press(struct seat *seat, struct wlr_pointer_button_event *event)
...
 	if (ctx.type == LAB_SSD_LAYER_SURFACE) {
 		struct wlr_layer_surface_v1 *layer =
 			wlr_layer_surface_v1_from_wlr_surface(ctx.surface);
 		if (layer && layer->current.keyboard_interactive) {
 			seat_set_focus_layer(seat, layer);
 		}
+#ifdef HAVE_XWAYLAND
+	} else if (ctx.type == LAB_SSD_UNMANAGED) {
+		desktop_focus_view_or_surface(seat, NULL, ctx.surface,
+			/*raise*/ false);
```
Consolatis, line 927: Should we raise here as well?

jlindgren90: I'm not sure. Unmanaged surfaces are already on top of everything else (except possibly other unmanaged surfaces). Maybe it makes sense to restack the clicked one on top.

desktop_focus_view_or_surface() doesn't currently have code to raise an unmanaged surface, but it could be done.

Consolatis: > Unmanaged surfaces are already on top of everything else (except possibly other unmanaged surfaces).

Only on the labwc scene graph side of things though, I am not sure if X11 internally has a unmanaged-always-on- top stacking order.

As we are already in the in the X11 ifdef + unmanaged check, maybe we could just restack it for the X11 side then do a `wlr_scene_node_raise_to_top()` for the scene graph side?

johanmalm:
> We had issues with focusing subsurfaces breaking drag and drop before (although I don't think this change will cause that as we only (re)focus the view itself if supplied as argument), might still be worth to verify that though.

Have confirmed `waybar` dnd still works.

src/input/cursor.c
johanmalm: I'd vote for _not_ adding any `wlr_xwayland_surface_restack()` calls for now. I'm saying that mostly out of fear of messing with OR surfaces, rather than having done any rational analysis (although I did check `sway` + `wlroots` and they only call the restack on map).

johanmalm:
LGTM

src/input/cursor.c
jlindgren90: > I am not sure if X11 internally has a unmanaged-always-on- top stacking order.

We actually restack unmanaged surfaces on top right after raising another (non-OR) XWayland surface. At that moment the labwc and XWayland stacking order should be in sync. Not sure if/how long it stays that way.

I'm good with leaving the current stacking approach as-is unless/until we run into specific issues with it.

johanmalm:
> > We had issues with focusing subsurfaces breaking drag and drop before (although I don't think this change will cause that as we only (re)focus the view itself if supplied as argument), might still be worth to verify that though.
> 
> Have confirmed `waybar` dnd still works.

Well... you obviously meant to try dnd with an XWayland client :smile: I tried with pcmanfm (with `GDK_BACKEND=X11`). Works fine. Couldn't think of anything else sensible to try with.

Consolatis:
I am mostly concerned about situations like X11 DnD and hovering over a tooltip or similar. It might be fine though and I am just overly cautions due to our history with the implicit wlroots DnD grab reacting strange to focus switches.

src/input/cursor.c
Consolatis: > I'm good with leaving the current stacking approach as-is unless/until we run into specific issues with it.

Me as well.

PR #1225, opened 2023-11-03
view: ensure that floating views don't overlap top panels

### LINKED ISSUE #1221: some windows open in sfwbar's exclusive zone
Some windows (mostly of non-gtk X apps--e.g., `feh`) open at the very top of the screen. I use `sfwbar` and prefer to have the bar at the top of the screen. So current options are a) to have the bar at top layer, which results in part of window being obscured (functionally undesirable), or b) to have the bar at bottom layer, which results in part of the bar being obscured (cosmetically undesirable).

For now I'm going with option "b" which seems less bad. This is what it looks like: [screenshot](https://gnuser.ddns.net/public/screenshot001.jpg).

Is there anything you guys can do for `labwc` to interpret layer shell exclusive zones more strictly? Ideally the user wouldn't have to care whether the bar is bottom layer or top layer because windows would stay out of the bar's exclusive zone.

jlindgren90:
The `top_left_edge_boundary_check()` function in xwayland.c ensures that views trying to position themselves at 0,0 don't end up with a titlebar offscreen. However, it doesn't take into account the usable area and thus these views can still end up overlapping a top panel.

Also, there is no good reason for `top_left_edge_boundary_check()` to be xwayland-specific. This logic should really be part of `view_adjust_for_layout_change()`.

To fix all this, add a new `view_adjust_floating_geometry()` function, which replaces the existing similar (and duplicated) logic in `view_apply_natural_geometry()` and `view_adjust_for_layout_change()`.

`view_adjust_for_layout_change()` is already being called from xwayland's `set_initial_position()`, so `top_left_edge_boundary_check()` is now redundant and can just be deleted.

Lightly tested with `waybar` and `feh --geometry 640x480+0+0`. The feh window is now correctly positioned below waybar, even if started before waybar (in that case, the feh window is moved when waybar starts).

Fixes #1221.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -354,6 +354,7 @@ void view_moved(struct view *view);
 void view_minimize(struct view *view, bool minimized);
 bool view_compute_centered_position(struct view *view,
 	const struct wlr_box *ref, int w, int h, int *x, int *y);
+bool view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry);
 void view_store_natural_geometry(struct view *view);
 
 /**
--- a/src/view.c
+++ b/src/view.c
@@ -530,6 +530,7 @@ bool
 view_compute_centered_position(struct view *view, const struct wlr_box *ref,
 		int w, int h, int *x, int *y)
 {
+	assert(view);
 	if (w <= 0 || h <= 0) {
 		wlr_log(WLR_ERROR, "view has empty geometry, not centering");
 		return false;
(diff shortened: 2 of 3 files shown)


Consolatis:
LGTM, didn't test.

johanmalm:
Thanks. I'll do some testing with this later.

johanmalm:
LGTM and have tested (with `feh`, `xterm` as well as disabling all outputs to make sure we don't crash).

Happy to merge, but just wanted to check the objective.

Observation 1: Openbox positions `feh`/`xterm` top/left taking into account _WM_NET_STRUT_PARTIAL when X/Y values of `0` are provided to the `-geometry` argument. However, when positive or negative non-zero values are given, _WM_NET_STRUCT_PARTIAL is not taken into account. So for example, I can position a `feh` image behind `tint2` by passing `--geometry 1366x768+1+1`. So Openbox counts X/Y from screen edge except when the value is zero in which case they count from panel edges.

Observation 2: On mutter, X/Y geometry values are counted from screen edges - but the client is never allowed to overlap panels (so just like you've implemented it in labwc).

Observation 3: Reading man X(7) is seems that it's up to Window Managers to respect desired geometry/position - i.e. it's optional.

I think I prefer it the way you've implemented it, but here are two considerations nonetheless:

1. Should we allow positioning outside of the usable-area? For most clients, it wouldn't make sense to, but as `feh` was used in the issue OP I immediately thought the use-case related to setting a background image (because that's what I'm used to on X). 
2. Should X/Y offset be relative to usable-area or output? On both openbox/mutter it is from the screen (output) edge except for the case where the outset is 0 on Openbox (as described above).

jlindgren90:
> Should X/Y offset be relative to usable-area or output? On both openbox/mutter it is from the screen (output) edge except for the case where the outset is 0 on Openbox (as described above).

Neither, I believe - global X/Y values for X11/XWayland applications are counted from the top-left corner of the root window. As far as I can tell, that should match up with the 0,0 point in the `wlr_output_layout`. In a multi-monitor setup, there might be a screen edge at 0,0 or there might not be.

> Should we allow positioning outside of the usable-area? For most clients, it wouldn't make sense to, but as feh was used in the issue OP I immediately thought the use-case related to setting a background image (because that's what I'm used to on X).

I'm inclined to say we should not allow positioning outside the usable area. The main concern here to me is XWayland applications that save/restore window position but don't check for layout/usable area changes on startup. It's not a great user experience for those applications to end up partially offscreen (or overlapping a panel).

I don't think using `feh` to set a background image will work anyway. In that case, I think it works by setting the backing pixmap of the root window (which is invisible in labwc). So I would vote for not worrying about that case. (XWayland windows of _NET_WM_TYPE_DESKTOP might be possible to support, but that's another discussion with its own concerns.)

johanmalm:
Agree. All good points. Will go for the merge.

Issue #1223, opened 2023-11-03
Keybindings for multimedia keys are no longer active


jp7677:
Hi there,

I'm using keybindings like this for the multimedia keys (the ones on the function keys) of my keyboard in my configuration: 

```
<keybind key="XF86LaunchA"><action name="ToggleFullscreen" /></keybind>
<keybind key="XF86LaunchB"><action name="Execute" command="fuzzel"/></keybind>
```

Unfortunately those bindings, this also includes volume and playback keys, do no longer work on master, but are still fine with 0.6.5.

To trigger the first action I need to press `fn`+`f1` for `XF86LaunchA`. I still see events like `XF86LaunchA` in `wev`, but labwc doesn't react anymore. Do I need to change my keybindings somehow to reactive them?

In case this is a regression,  running `git bisect` I could isolate this to this commit (assuming that I didn't screw up when bisecting and this isn't a false positive):

```
git bisect log
...
# first bad commit: [e77330bc3fe733412e92f4e90d797d31ee6c5652] keyboard: make keybind match stricter

commit e77330bc3fe733412e92f4e90d797d31ee6c5652
Author: Johan Malm <jgm323@gmail.com>
Date:   Tue Sep 26 17:51:54 2023 +0100

    keyboard: make keybind match stricter
    
    ...and avoid failing to send release events to clients for any keys that were
    not absorbed by a keybind.
    
    Do not match keybinds if there are other non-modifier keys (not part of any
    defined bind) pressed at the same time.
    
    Only store non-modifier keycodes in the key-state.c 'pressed' array.
    This makes the call to wlr_seat_keyboard_notify_enter() in seat_focus()
    consistent with the equivalent in sway (in seat_keyboard_notify_enter()).
    
    Fixes: issue #1091

```

Thanks for all your work, I really like labwc!

Consolatis:
Thank you for pointing out this regression!
A fix for this issue need some more discussions on the way forward in regards to our keybind handling.

Consolatis:
Should be fixed on latest master. If there are still keybind related issues for multimedia keys feel free to reopen / comment.

Issue #1220, opened 2023-11-02
Release `0.6.6`


johanmalm:
Intend to release `0.6.6` on 2-Dec-2023

Using https://github.com/labwc/labwc/milestone/2

Feel free to point out any critical issues that should be addressed before the release here.

The core devs will try to honor a two-week cool down period prior to the release where we don't refactor + add, but merely fix issues.

johanmalm:
> Intend to release `0.6.6` on 2-Dec-2023
> 
> Using https://github.com/labwc/labwc/milestone/2
> 
> Feel free to point out any critical issues that should be addressed before the release here.
> 
> The core devs will try to honor a 2-3 week cool down period prior to the release where we don't refactor + add, but merely fix issues.

PR #1202, opened 2023-10-23
src/keyboard.c: always try to remove keys from the pressed set on release

Consolatis:
Fixes:
- #1200

Reported-by: @kyak

<details><summary>Technical nitty gritty details</summary>
The following logs are based on a work-in-progress patch but are good enough to explain the earlier issues:

- #1129
`XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle`
```
00:15:07.138 [../src/keyboard.c:242] Got new key event: 50 (press) modifier: yes
00:15:07.138 [../src/keyboard.c:264]    Translated:
00:15:07.138 [../src/keyboard.c:269]            0xffe1 (Shift_L)
00:15:07.138 [../src/keyboard.c:271]    Raw:
00:15:07.138 [../src/keyboard.c:276]            0xffe1 (Shift_L)
00:15:07.874 [../src/keyboard.c:242] Got new key event: 64 (press) modifier: no
00:15:07.874 [../src/keyboard.c:264]    Translated:
00:15:07.874 [../src/keyboard.c:269]            0xfe08 (ISO_Next_Group)
00:15:07.874 [../src/keyboard.c:271]    Raw:
00:15:07.874 [../src/keyboard.c:276]            0xffe9 (Alt_L)
00:15:08.266 [../src/keyboard.c:242] Got new key event: 50 (release) modifier: no
00:15:08.266 [../src/keyboard.c:264]    Translated:
00:15:08.266 [../src/keyboard.c:269]            0xfe08 (ISO_Next_Group)
00:15:08.266 [../src/keyboard.c:271]    Raw:
00:15:08.266 [../src/keyboard.c:276]            0xffe1 (Shift_L)
00:15:08.754 [../src/keyboard.c:242] Got new key event: 64 (release) modifier: yes
00:15:08.755 [../src/keyboard.c:264]    Translated:
00:15:08.755 [../src/keyboard.c:269]            0xffe9 (Alt_L)
00:15:08.755 [../src/keyboard.c:271]    Raw:
00:15:08.755 [../src/keyboard.c:276]            0xffe9 (Alt_L)
```

In this case
- we were adding key `64` to key_state because it was not detected as a modifier (via translated sym)
- we were *not* removing key `64` from key_state because now it had been detected as a modifier
- the result was that all keybinds stopped working because there was still a `Alt_L` key stuck in the key_state

[...]

PR #1226, opened 2023-11-04
keyboard: avoid stuck keys due to keybindings (alternate approach)

jlindgren90:
For discussion/consideration along with #1212 (not mutually exclusive) for solving #1208.

Before commit e77330bc3fe7, there were issues with keys becoming "stuck" if other keys were pressed at the time a keybinding was matched, because those other keys were included in the "bound" set and the release events were incorrectly eaten by labwc.

Commit e77330bc3fe7 solved that issue with the "big hammer" approach of preventing keybindings from working at all if other keys were pressed:

        if (key_state_nr_pressed_keys() > 1) {
                return false;
        }

This is an alternate approach to solving the original problem, by (1) not including those other keys in the "bound" set and (2) making sure we always forward release events for un-bound keys to clients (even if a menu or OSD is displayed).

Details:

- Since we only ever want to store the single matched keycode as bound, `key_state_store_pressed_keys_as_bound()` doesn't really make sense in the plural, so rename it to `key_state_store_pressed_key_as_bound()` and pass in the keycode.

- The calls to `key_state_store_pressed_keys_as_bound()` within `handle_keybinding()` appear to be redundant since it is also called from the parent function (`handle_compositor_keybindings()`). So remove these calls.

- Finally, rework the logic for handling key-release events so that we always forward release events for keys not in the "bound" set.

This PR does not remove the `key_state_nr_pressed_keys() > 1` check, and because of that should not result in any functional change. It should however make it possible to relax or remove that check in future.

### DIFF
--- a/src/input/key-state.c
+++ b/src/input/key-state.c
@@ -11,7 +11,7 @@ struct key_array {
 	int nr_keys;
 };
 
-static struct key_array pressed, bound, pressed_sent;
+static struct key_array pressed, pressed_mods, bound, pressed_sent;
 
 static bool
 key_present(struct key_array *array, uint32_t keycode)
@@ -69,12 +69,16 @@ key_state_nr_pressed_sent_keycodes(void)
 }
 
 void
-key_state_set_pressed(uint32_t keycode, bool ispressed)
+key_state_set_pressed(uint32_t keycode, bool is_pressed, bool is_modifier)
 {
-	if (ispressed) {
+	if (is_pressed) {
 		add_key(&pressed, keycode);
+		if (is_modifier) {
+			add_key(&pressed_mods, keycode);
+		}
 	} else {
 		remove_key(&pressed, keycode);
+		remove_key(&pressed_mods, keycode);
 	}
 }
 
@@ -82,6 +86,15 @@ void
 key_state_store_pressed_key_as_bound(uint32_t keycode)
 {
 	add_key(&bound, keycode);
+	/*
+	 * Also store any pressed modifiers as bound. This prevents
+	 * applications from seeing and handling the release event for
+	 * a modifier key that was part of a keybinding (e.g. Firefox
+	 * displays its menu bar for a lone Alt press + release).
+	 */
+	for (int i = 0; i < pressed_mods.nr_keys; ++i) {
+		add_key(&bound, pressed_mods.keys[i]);
+	}
 }
 
 bool
--- a/src/input/key-state.c
+++ b/src/input/key-state.c
@@ -79,10 +79,9 @@ key_state_set_pressed(uint32_t keycode, bool ispressed)
 }
 
 void
-key_state_store_pressed_keys_as_bound(void)
+key_state_store_pressed_key_as_bound(uint32_t keycode)
 {
-	memcpy(bound.keys, pressed.keys, MAX_PRESSED_KEYS * sizeof(uint32_t));
-	bound.nr_keys = pressed.nr_keys;
+	add_key(&bound, keycode);
 }
 
 bool
--- a/include/input/key-state.h
+++ b/include/input/key-state.h
@@ -19,7 +19,7 @@
 uint32_t *key_state_pressed_sent_keycodes(void);
 int key_state_nr_pressed_sent_keycodes(void);
 
-void key_state_set_pressed(uint32_t keycode, bool ispressed);
+void key_state_set_pressed(uint32_t keycode, bool is_pressed, bool is_modifier);
 void key_state_store_pressed_key_as_bound(uint32_t keycode);
 bool key_state_corresponding_press_event_was_bound(uint32_t keycode);
 void key_state_bound_key_remove(uint32_t keycode);
--- a/include/input/key-state.h
+++ b/include/input/key-state.h
@@ -2,6 +2,9 @@
 #ifndef LABWC_KEY_STATE_H
 #define LABWC_KEY_STATE_H
 
+#include <stdbool.h>
+#include <stdint.h>
+
 /*
  * All keycodes in these functions are (Linux) libinput evdev scancodes which is
  * what 'wlr_keyboard' uses (e.g. 'seat->keyboard_group->keyboard->keycodes').
@@ -17,7 +20,7 @@ uint32_t *key_state_pressed_sent_keycodes(void);
 int key_state_nr_pressed_sent_keycodes(void);
 
 void key_state_set_pressed(uint32_t keycode, bool ispressed);
-void key_state_store_pressed_keys_as_bound(void);
+void key_state_store_pressed_key_as_bound(uint32_t keycode);
 bool key_state_corresponding_press_event_was_bound(uint32_t keycode);
 void key_state_bound_key_remove(uint32_t keycode);
 int key_state_nr_bound_keys(void);
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -360,19 +360,6 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 	}
 
 	/*
-	 * A keybind is not considered valid if other keys are pressed at the
-	 * same time.
-	 *
-	 * In labwc, a keybind is defined by one/many modifier keys + _one_
-	 * non-modifier key. Returning early on >1 pressed non-modifier keys
-	 * avoids false positive matches where 'other' keys were pressed at the
-	 * same time.
-	 */
-	if (key_state_nr_pressed_keys() > 1) {
-		return false;
-	}
-
-	/*
 	 * Handle compositor keybinds
 	 *
 	 * When matching against keybinds, we process the input keys in the
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -230,25 +230,14 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 	 */
 
 	bool is_modifier = false;
-	bool is_layout_switch = false;
 	uint32_t modifiers = wlr_keyboard_get_modifiers(wlr_keyboard);
 
 	for (int i = 0; i < translated.nr_syms; i++) {
 		is_modifier |= is_modifier_key(translated.syms[i]);
-		is_layout_switch |= translated.syms[i] == XKB_KEY_ISO_Next_Group;
 	}
 
-	/*
-	 * An earlier press event from a key that causes a layout change event
-	 * might have been added already without us knowing that it actually was
-	 * a XKB_KEY_ISO_Next_Group sym. Thus we always try to remove the current
-	 * key from the set of pressed keys on release.
-	 */
-	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-		key_state_set_pressed(event->keycode, false);
-	} else if (!is_modifier && !is_layout_switch) {
-		key_state_set_pressed(event->keycode, true);
-	}
+	key_state_set_pressed(event->keycode,
+		event->state == WL_KEYBOARD_KEY_STATE_PRESSED);
 
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
 		/*
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -105,7 +105,6 @@ handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym, x
 			/* Use keycodes */
 			for (size_t i = 0; i < keybind->keycodes_len; i++) {
 				if (keybind->keycodes[i] == code) {
-					key_state_store_pressed_keys_as_bound();
 					actions_run(NULL, server, &keybind->actions, 0);
 					return true;
 				}
@@ -114,7 +113,6 @@ handle_keybinding(struct server *server, uint32_t modifiers, xkb_keysym_t sym, x
 			/* Use syms */
 			for (size_t i = 0; i < keybind->keysyms_len; i++) {
 				if (xkb_keysym_to_lower(sym) == keybind->keysyms[i]) {
-					key_state_store_pressed_keys_as_bound();
 					actions_run(NULL, server, &keybind->actions, 0);
 					return true;
 				}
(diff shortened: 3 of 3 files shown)


johanmalm:
I'll have to look at this tomorrow.

At a glance, this refactoring makes me think we (I) made a mistake in the "big hammer" commit in that we no longer store modifiers in the "pressed" array, but it's important that we don't pass release events for bound modifiers too (consider Alt in particular).

Consolatis:
> At a glance, this refactoring makes me think we (I) made a mistake in the "big hammer" commit

IMHO at this point in the dev cycle its a shared responsibility, especially in regards to our PR merge rules.
E.g. no point in pointing out any single person to blame and thus the `we` sounds about right :)

> in that we no longer store modifiers in the "pressed" array, but it's important that we don't pass release events for bound modifiers too (consider Alt in particular).

Yeah.. thought the same.

Add to the mix that we also receive key press/release events for modifiers as well (which are sometimes not even detectable as modifiers in which case we can't really detect them being part of a keybind we just executed).

Another thing is that we execute keybinds which may cause focus switching (and thus the seat_enter with the set of pressed keys) on `press`, that might also play a role here.

The whole keybind handling is quite of a mess but I am not sure how we can properly design it differently. There might be a good long term solution but I couldn't think of one so far.

Personally I'll take some keybinds not working over repeating keys all the time. When you had a wrong `enter`, `tab` or `del` key fake repeat on a client before you know what I mean. Pretty funny in a file browser for example. Similar but not as bad when switching to a code editor or chat client.

jlindgren90:
> it's important that we don't pass release events for bound modifiers too (consider Alt in particular).

Good point. I assumed we wanted to pass those through, but in testing with Openbox, it appears not to (i.e. Firefox menus don't appear if you Alt-Tab and cycle back to Firefox).

> Personally I'll take some keybinds not working over wrongly repeating keys all the time.

For me, the issue is that all keybindings will stop working at random times. That's a pretty big usability issue to me, and I think we need to find a way to solve both that and the wrongly repeating keys.

jlindgren90:
> which are sometimes not even detectable as modifiers in which case we can't really detect them being part of a keybind we just executed

Could you explain this part a little more? Is it talking about odd modifiers like `Fn`?

johanmalm:
So, maybe this patch is good if we just add currently pressed modifiers in `key_state_store_pressed_key_as_bound()`.

And... the modifiers are held in `seat->keyboard_group->keyboard->modifiers` (see `seat.c+498`)

johanmalm:
...and we remove the `key_state_nr_pressed_keys() > 1` hammer

jlindgren90:
> So, maybe this patch is good if we just add currently pressed modifiers in key_state_store_pressed_key_as_bound().

All currently pressed modifiers, or just those that are part of the keybinding?

Consolatis:
> > which are sometimes not even detectable as modifiers in which case we can't really detect them being part of a keybind we just executed
> 
> Could you explain this part a little more? Is it talking about odd modifiers like `Fn`?

See #1202 "nitty gritty details" [0]
It might be similar for the `fn` key, that could explain #1223 but I didn't test that.

We may also get different keysyms (and thus know if something is modifier or not) depending on the ordering the keys are pressed and released.

johanmalm:
> All currently pressed modifiers, or just those that are part of the keybinding?

I think those two things have to be the same don't they. 

If "other" modifiers are pressed a keybind shouldn't be valid.

For example, if we have defined C+a as a keybind, then C+W+a should NOT trigger it.

It's only the non-modifier bit that has to be "only one" (and the presence of others such non-modifiers should ideally disqualify the bind, but it looks like reality is messy and that we should probably give up on that)

jlindgren90:
Okay that makes sense. I was thinking about modifiers like Num Lock but those aren't really "pressed", so probably irrelevant in this case.

jlindgren90:
For the ISO_Next_Group thing (translated Alt or Shift), as long as we clear both pressed+bound for the correct (raw) key-code we should be fine, I think? Assuming that ISO_Next_Group is not itself part of a keybinding, we will forward the release event to clients, which I think is no different than what happens today.

Consolatis:
> Assuming that ISO_Next_Group is not itself part of a keybinding, we will forward the release event to clients, which I think is no different than what happens today.

Well, it might be. A user could configure `Super_L` to be the trigger for switching the keyboard layout. That would then cause it to be recognized as `ISO_Next_Group`. The user could then define any keybind that involves `W-`  like `W-t` or `W-C-t`.

jlindgren90:
In that case the concern would be that we forward an unexpected key release event (Super_L) after the keybinding because we didn't originally recognize the key press as a modifier -- do I understand correctly?

What should we (ideally) do in that case?

johanmalm:
> In that case the concern would be that we forward an unexpected key release event (Super_L) after the keybinding because we didn't originally recognize the key press as a modifier -- do I understand correctly?
> 
> What should we (ideally) do in that case?

I’ve not tested, but suspect we have to accept that that whatever key or key-combo is used to switch keyboard layout will not work well (or at all) for the compositor keybinds. I’m okay to accept that.

Whilst of course related, I’d be okay with thinking about that separately.
It’s already an issue today… for example if alt+shift is used for layout switch, then alt+shift+tab doesn’t work.

Consolatis:
I currently do not have the brain capacity to wrap my head around the input handling.
Instead, I'd want to point out that we are planning to release [0.6.6](https://github.com/labwc/labwc/issues/1220) in about a month.

So some generic thoughts instead:
- I think there may be long term improvements to make our key handling more robust
- We have a current issue in that some 'stuck' key prevent all further keybinds to work
- Personally, I really don't want fake repeating keys to return (e.g. sending a `press` to a client but eat up the `release`)
- Sending `release` to a client without having sent a `press` first also caused issues in the past
- IMHO for now we should focus on either working around (safety net) or fixing the current issue regarding restoring the keybinds on 'stuck' keys
- We should try not to do too much generic changes regarding input handling until after the release. This part of the codebase is really finicky in that one may change one thing which improves 90% of the usual usage cases but also introduces hard to debug bugs in another area.

If possible I'd like to leave this issue to the both of you, feel free to ping me for specific input though.

johanmalm:
^ Thanks @Consolatis I'll try to find a clear slot (rare commodity!) over the next week to try to think this through and propose a patch to get us to a 'safe' state. Having had a finger in the keyboard-pie for a long time, I think I can remember most of the pitfalls. But yeah, I agree - this really is a tough area to get right.

jlindgren90:
I'm all for a targeted/conservative fix if that exists. Not sure if it does but looking forward to what @johanmalm comes up with.

Currently my free time is getting sucked up by home improvement projects, so it might be a little longer before I can get back to looking at this.

johanmalm:
Hey, I'll have a go, but don't get hopes up too much 😄 
The last couple of patches I did on keyboard.c turned out not to be too clever when they got some eyeballs on them!

jlindgren90:
I had a little more time to work on this, and I think I addressed the comments so far. Firefox no longer displays its menu bar after Alt-Tab. I'll continue to test and see if any other issues crop up.

src/input/key-state.c
```diff
@@ -69,20 +69,32 @@ key_state_nr_pressed_sent_keycodes(void)
...
 	}
 }
 
 void
-key_state_store_pressed_keys_as_bound(void)
+key_state_store_pressed_key_as_bound(uint32_t keycode)
 {
-	memcpy(bound.keys, pressed.keys, MAX_PRESSED_KEYS * sizeof(uint32_t));
-	bound.nr_keys = pressed.nr_keys;
+	add_key(&bound, keycode);
```
johanmalm, line 88: In the …bound() function do we need to nullify the array before adding?

bound.nr_keys = 0

johanmalm (commented):
I think this is the right direction, i.e. make …bound() function singular and add modifiers when called (like you’ve done).

Have just had look on phone. Just one comment re nullifying array.

Couldn’t follow logic change in first commit (but probably just me)

Will review on computer and test later.

src/input/key-state.c
jlindgren90: I think it doesn't matter one way or the other. If there are still pressed/held modifiers from a previous keybinding (that haven't been released yet) then we will add them a second time, but IIRC add_key() has a check to avoid adding a duplicate key-code.

jlindgren90:
The first commit is probably less readable than it could be due to the indentation change. I could rework it a little to make the diff smaller if that's helpful.

src/input/key-state.c
jlindgren90: Actually, thinking about this more, I think we do not want to clear/nullify the bound array before adding. There may be a non-modifier key still pressed from a previous keybinding (if you pressed two in quick succession) and that should still be considered "bound".

johanmalm: When I made the comment, I had forgotten that there is a `key_state_bound_key_remove()`, so thought we just kept adding :smile:

You make a good point. Agree.

src/input/key-state.c
```diff
@@ -69,20 +69,32 @@ key_state_nr_pressed_sent_keycodes(void)
...
+	/*
+	 * Also store any pressed modifiers as bound. This prevents
+	 * applications from seeing and handling the release event for
+	 * a modifier key that was part of a keybinding (e.g. Firefox
+	 * displays its menu bar for a lone Alt press + release).
+	 */
+	for (int i = 0; i < pressed_mods.nr_keys; ++i) {
+		add_key(&bound, pressed_mods.keys[i]);
+	}
```
Consolatis, line 97: Just wanted to point out something: In a `W-Left` example,
- We'd send a press event for the `Super` key to the focused surfaces (`Super_L` or `Super_R`).
- Once the `m` comes in we'd then handle they keybind
- Store the currently set of pressed keys and modifiers as bound
- And thus don't send the release for the earlier `Super_L` / `Super_R`, resulting in the client getting a stray `Press` for the modifier key itself

We can't really prevent the `Super_L` / `Super_R` key to be sent to the application because at that point we don't yet know if it is actually a part of a keybind or not. But I guess we should try to make sure not the eat its release *if* it has already been sent. Maybe applications are also fine with `Press` events for modifiers, I don't know. It feels wrong though unless I miss something.

As we currently do not store the modifiers as pressed this would not happen and the application would see the corresponding `Release` event.

Another thing: the `is_modifier` state may be wrong in case of layout switching, e.g. it might show up as that group_next thing and thus end up in the usual pressed array, depending on the order of Press and Release events for the layout switch.

johanmalm: > We can't really prevent the `Super_L` / `Super_R` key to be sent to the application because at that point we don't yet know if it is actually a part of a keybind or not. But I guess we should try to make sure not the eat its release _if_ it has already been sent. Maybe applications are also fine with `Press` events for modifiers, I don't know. It feels wrong though unless I miss something.

I've never seen this documented anywhere, but believe applications cope fine with seeing modifier press without associated release. And we've done it like that for a long time. I think `alt` is the example the convinced me... (press+release triggers menu). IIRC the thing the makes applications misbehave is sending a release without having first sent a press.
 
> As we currently do not store the modifiers as pressed this would not happen and the application would see the corresponding `Release` event.

Well - until recently they used all (modifiers + non-modifiers) be stored in the pressed array :smile:

> Another thing: the `is_modifier` state may be wrong in case of layout switching, e.g. it might show up as that group_next thing and thus end up in the usual pressed array, depending on the order of Press and Release events for the layout switch.

I tried to reproduce that, but seems to have gone away (i.e. press shift + press caps + release shift + release caps).

I can't test the keybind with `XF86LaunchA` thing (issue #1223). @jp7677 any chance of testing this branch against that?

johanmalm:
> The first commit is probably less readable than it could be due to the indentation change. I could rework it a little to make the diff smaller if that's helpful.

I've got my head around it now - and am fine with it.

johanmalm:
I've got a good feeling about this patch set. I can't test every case, but I can't reproduce some of the repeat issues we've seen before (e.g. 'other' non-modifier pressed when keybind is processed and getting `...llllllll` after using `W-l` for swaylock).

The logic looks clean and simplifies things in my mind.

I suggest we go with it and see if @kyak + @consus + @jp7677 could try to re-produce recent keyboard issues.

jp7677:
I've tested this branch briefly and the `fn`+`f` issues I reported earlier are solved with this PR, all key bindings with multimedia keys do work.

src/input/key-state.c
jlindgren90: > We can't really prevent the Super_L / Super_R key to be sent to the application because at that point we don't yet know if it is actually a part of a keybind or not. But I guess we should try to make sure not to eat its release if it has already been sent. Maybe applications are also fine with stray Press events for modifiers, I don't know. It feels wrong though unless I miss something.

This initially felt wrong to me too. But it's existing behavior (in both 0.6.5 and master) at least in the Alt-Tab case -- we eat the Alt release-event. And it seems to be the only way to prevent e.g. Firefox from seeing a lone Alt press + release and displaying its menu bar.

> Another thing: the is_modifier state may be wrong in case of layout switching, e.g. it might show up as that group_next thing and thus end up in the usual pressed array, depending on the order of Press and Release events for the layout switch. The corresponding Release may then show up correctly as a modifier. Or the other way around.

This should be handled reasonably well. Both modifiers and non-modifiers are stored in the "pressed" array unconditionally (as raw keycodes). And on release, keycodes are removed from both "pressed" and "pressed_mods" unconditionally. So the only consequence would be in the case where the press event is not detected as a modifier -- then it will not be stored in the "bound" array. I don't think there is much we can do about that.

jlindgren90:
I'd like to figure out what's causing #1238 before we merge this. It may be completely unrelated -- I'm not sure yet.

jlindgren90:
Rebased onto master (no other changes)

kyak:
> I've got a good feeling about this patch set. I can't test every case, but I can't reproduce some of the repeat issues we've seen before (e.g. 'other' non-modifier pressed when keybind is processed and getting `...llllllll` after using `W-l` for swaylock).
> 
> The logic looks clean and simplifies things in my mind.
> 
> I suggest we go with it and see if @kyak + @Consus + @jp7677 could try to re-produce recent keyboard issues.

I haven't seen stuck keys in a while; and I also don't reproduce stuck keys with this branch.

johanmalm:
Thanks @kyak and @jp7677 for extra testing.
And to @jlindgren90 for persevering with this to get us to a position of cleaner, simpler code and (hopefully) a solid key-handling base.

PR #1229, opened 2023-11-06
xdg: try to handle missing set_window_geometry with Qt apps

### LINKED ISSUE #1194: Rarely, title-bar may not be resized on snapping to bottom/top by drag-and-drop

jlindgren90:
Qt applications occasionally fail to call set_window_geometry after a configure request, but do correctly update the actual surface extent. This results in a mismatch between the window decorations (which follow the logical geometry) and the visual size of the client area. As a workaround, try to detect this case and ignore the out-of-date window geometry.

Should fix #1194.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -90,6 +90,27 @@ handle_commit(struct wl_listener *listener, void *data)
 	struct wlr_box size;
 	wlr_xdg_surface_get_geometry(xdg_surface, &size);
 
+	/*
+	 * Qt applications occasionally fail to call set_window_geometry
+	 * after a configure request, but do correctly update the actual
+	 * surface extent. This results in a mismatch between the window
+	 * decorations (which follow the logical geometry) and the visual
+	 * size of the client area. As a workaround, we try to detect
+	 * this case and ignore the out-of-date window geometry.
+	 */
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, didn't test.

Consolatis:
Could you add `Fixes: #1194` to the commit body? That makes it easier in the future to figure out the details of the change (including your debugging journey :)

jlindgren90:
> Could you add `Fixes: #1194` to the commit body? That makes it easier in the future to figure out the details of the change (including your debugging journey :)

Sure, can do it later, or feel free to amend the commit and add that if you like.

Consolatis:
added and force pushed

Consolatis:
Merging due to confirmations that it fixes the issue in #1194.

PR #1232, opened 2023-11-08
interactive: Make window snapping with mouse more intuitive

tokyo4j:
In current implementation, window snapping triggered with mouse has some counter-intuitive behaviors in my (vertically aligned) dual monitor setup:

(a) Cannot snap a tiled window into another output

(b) Dragging a tiled window into another output forces snapping (related to (a))

(c) Snapping is triggered when the center of the window is in another output, even if the cursor is not on the edge.

(d) Snapping a tiled window to the same direction moves the window into the adjacent output and inverts its direction (mentioned in comment)

(a)-(c) occur because the cursor position relative to the output **the view belongs to** is used to determine if and in which direction the snapping is triggered. So this PR changes it to use the output that **contains the cursor** instead.

(d) is addressed by simply adding `across_outputs` parameter to `view_snap_to_edge()`.

I'm testing my commit and it's working fine so far.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -674,7 +674,9 @@ actions_run(struct view *activator, struct server *server,
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
-				view_snap_to_edge(view, edge, /*store_natural_geometry*/ true);
+				view_snap_to_edge(view, edge,
+					/*across_outputs*/ true,
+					/*store_natural_geometry*/ true);
 			}
 			break;
 		case ACTION_TYPE_GROW_TO_EDGE:
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -114,37 +114,45 @@ snap_to_edge(struct view *view)
 		return false;
 	}
 
+	struct output *output = output_nearest_to_cursor(view->server);
+	if (!output_is_usable(output)) {
+		wlr_log(WLR_ERROR, "output at cursor is unusable");
+		return false;
+	}
+
 	/* Translate into output local coordinates */
 	double cursor_x = view->server->seat.cursor->x;
 	double cursor_y = view->server->seat.cursor->y;
 	wlr_output_layout_output_coords(view->server->output_layout,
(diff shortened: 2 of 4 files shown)


src/action.c
```diff
@@ -674,7 +674,8 @@ actions_run(struct view *activator, struct server *server,
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
-				view_snap_to_edge(view, edge, /*store_natural_geometry*/ true);
+				view_snap_to_edge(view, edge, true,
+					/*store_natural_geometry*/ true);
```
Consolatis, line 678: Could you add a inline comment for the new bool here? It makes it way easier to read the code rather than having a random assortment of `true` / `false` arguments.

Same for the call in `interactive.c`.

src/interactive.c
```diff
@@ -114,39 +114,49 @@ snap_to_edge(struct view *view)
 		return false;
 	}
 
-	/* Translate into output local coordinates */
 	double cursor_x = view->server->seat.cursor->x;
 	double cursor_y = view->server->seat.cursor->y;
+
+	/* Window snapping is performed in the output the cursor is at  */
+	struct output *output = output_from_wlr_output(view->server,
+		wlr_output_layout_output_at(view->server->output_layout, cursor_x, cursor_y));
```
Consolatis, line 122: We have a helper for this: `output_nearest_to_cursor(struct server *server)`

Consolatis (commented):
Very nice PR! Two small comments below.
I didn't yet test it but it looks very promising.

tokyo4j:
Thank you for the feedback!

jlindgren90:
I have not tested the changes but they seem to make sense. Thank you and nice job!

tokyo4j:
Sorry I didn't understand what Draft is and just set it as draft because all other open PRs were Draft.

Consolatis:
Draft is fine as well, I would have switched it over after testing. I'll try to do that until the end of the weekend.

johanmalm:
Great analysis and patch. Much appreciated.
LGTM. Have tested and it works fine.

PR #1236, opened 2023-11-09
window-rules: add `fixedPosition` property

johanmalm:


docs/labwc-config.5.scd
```diff
@@ -526,6 +526,10 @@ situation.
 *<windowRules><windowRule ignoreFocusRequest="">* [yes|no|default]
 	*ignoreFocusRequest* prevent window to activate itself.
 
+*<windowRules><windowRule ignoreUsableArea="">* [yes|no|default]
+	*ignoreUsableArea* do not re-position on changing usable area (which
+	takes into account exclusive layer-shell clients)
```
Consolatis, line 531: Suggest adding something like "and `<margin>` settings"

Consolatis:
Other than the small docs nitpick this LGTM. Didn't test.

johanmalm:
Thanks for review. Will incorporated.

How about the following two changes too:

1. `s/ignoreUsableArea/fixPosition/`
2. Disallow alt+mouse-press to move/resize when this property == true

stefonarch:
> Disallow alt+mouse-press to move/resize when this property == true

Good idea!

Consolatis:
> Thanks for review. Will incorporated.
> 
> How about the following two changes too:
> 
> 1. `s/ignoreUsableArea/fixPosition/`
> 2. Disallow alt+mouse-press to move/resize when this property == true

Sounds good to me.

Consolatis:
:)

Issue #1158, opened 2023-10-13
Fullscreen window hides Waybar on all workspaces


tsujan:
The first time I encountered this problem, I thought Waybar had crashed: Skype's window was fullscreen, and I changed the workspace by pressing a `GoToDesktop` shortcut, but there was no trace of Waybar anywhere. However, as soon as I unfullscreened Skype, Waybar was shown on all workspaces.

Then I saw the same thing happened when I fullscreened FeatherPad (Qt6), Falkon (Qt5), Firefox or any other window. Apparently, the toolkit wasn't relevant.

I don't know whether this behavior is specific to Waybar or it happens with other panels too; hence the title of this report.

I use 0.6.5, the latest release for now.

EDIT: I have no window rule for Waybar in `rc.xml`.

Consolatis:
This is caused by us being lazy and not resetting the "top" layer when switching workspaces.
That one is a bit special as we hide it whenever we go into fullscreen, but obviously we should show it again when we are leaving the workspace with the active fullscreen window.

In the meantime you can configure waybar to use the "bottom" layer (+ exclusive zone if not already using that).
That should make stuff work, regardless of fullscreen windows.

heroin-moose:
Was too lazy to report, thank you.

Consolatis:
Should be fixed in latest master. Feel free to comment if there are further issues with top layer layershell surfaces and fullscreen.

tsujan:
Thanks!

I'll wait for 0.6.6 to have recent fixes and changes (0.6.5 has been rock solid here).

johanmalm:
> Thanks!
> 
> I'll wait for 0.6.6 to have recent fixes and changes (0.6.5 has been rock solid here).

Will have a three week cool down period. Aiming for release 2-Dec.

PR #1237, opened 2023-11-10
Update toplayer visiblity on workspace switch

Consolatis:
Fixes:
- #1158

Only had a very short test nested.

I am also not sure if `view.c` is actually the correct place for the new `view_update_top_layer()` function:
- As it depends on a set of views it could also belong to `desktop.c`.
- As it toggles the visibility of a specific layershell layer it could also belong to `layers.c`.


src/view.c
```diff
@@ -1097,6 +1102,33 @@ view_toggle_fullscreen(struct view *view)
 	view_set_fullscreen(view, !view->fullscreen);
 }
 
+void
+view_update_top_layer(struct server *server)
+{
+	struct view *view;
+	struct output *output;
+
+	uint32_t top = ZWLR_LAYER_SHELL_V1_LAYER_TOP;
```
Consolatis, line 1111: remove empty line

jlindgren90:
Didn't test but LGTM. It looks like this would also correctly handle the case where you have two fullscreen windows and close one of them. :+1:

johanmalm:
LGTM. Have tested and seems to work great.

Nice to increase usage of `for_each_view()`

Agree, `desktop.c` feel like a better home. Even if we're dragging some layer-stuff it's all a function of the views.

Suggest `s/view_update_top_layer/desktop_update_top_layer_visibility/`

Consolatis:
> Agree, `desktop.c` feel like a better home. Even if we're dragging some layer-stuff it's all a function of the views.

Moved to `desktop.c`

> Suggest `s/view_update_top_layer/desktop_update_top_layer_visibility/`

Changed

johanmalm:
Really nice patch. Thanks!

Issue #1238, opened 2023-11-11
Shifted keystrokes are delayed when typing quickly (regression 0.6.5 -> master)


jlindgren90:
In master (368ede74602e as of writing), Shift+ or Ctrl+ keystrokes are noticeable delayed when typing quickly. I'm not talking about keybindings, but basic keystrokes like Shift-A (for a capital A) or Ctrl+V (paste). It's a little hard to quantify but very noticeable when doing a significant amount of typing, because there is a small delay with every capital letter and most punctuation.

I think this started happening in the last week or two. The issue is definitely not present in 0.6.5.

jlindgren90:
`git bisect` points to 984aeb0b0bf8858217528342baab0b28889fa699 as the first bad commit, from:

- #1077

Consolatis:
*Might* be related to
```c
wlr_seat_set_keyboard(seat->seat, wlr_keyboard);
```

Just guessing though.

johanmalm:
I can reproduce.

Have rebased (locally) `unbound-key-release` onto master, but no difference.

jlindgren90:
I can confirm it's related to the `wlr_seat_set_keyboard` call in `keyboard_modifiers_notify`. Commenting out that line makes the delay disappear.

Also, pressing even Shift on its own causes the delay -- it doesn't have to be a combination.

johanmalm:
Have just confirmed the same.
Shall I do a patch?

PR #1239, opened 2023-11-11
keyboard: fix shifted keystroke delay

johanmalm:
Remove wlr_seat_set_keyboard() from `keyboard_modifiers_notify()` but leave it in `keyboard_key_notify()`

Fixes regression introduced in 984aeb0

Reported-by: @jlindgren90

Fixes: #1238

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -84,7 +84,6 @@ keyboard_modifiers_notify(struct wl_listener *listener, void *data)
 			}
 		}
 	}
-	wlr_seat_set_keyboard(seat->seat, wlr_keyboard);
 	wlr_seat_keyboard_notify_modifiers(seat->seat, &wlr_keyboard->modifiers);
 }
 


jlindgren90:
I don't know the original reason that call was added, nor understand why it causes the issue, so I'll defer to @Consolatis on this one.

johanmalm:
> I don't know the original reason that call was added, nor understand why it causes the issue, so I'll defer to @Consolatis on this one.

Yep.

I have tested though. Still works as expected.

Consolatis:
> I don't know the original reason that call was added

Usually it should not make a difference unless virtual keyboards and physical keyboards are involved at the same time.
In that case we want to update the modifiers of the keyboard that actually sent the modifiers rather than always using the ones from the physical keyboard group (or whatever handled the last key). We are doing the same in `keyboard_key_notify()`.

> nor understand why it causes the issue

That one I don't understand either.

I am fine with simply removing it for now though, modifiers for virtual keyboards are way less important than fixing slow typing which btw I can reproduce as well. I thought that was related to some mesa update on my system which free's buffers too late or something like that.

jlindgren90:
Is it possible that keyboard_modifiers_notify() is being called for all the individual keyboards in the keyboard group, and wlr_seat_set_keyboard() is cycling between them? Just speculating.

Anyway, it sounds like we are all good with merging this, I'll go for it. Maybe something to revisit later.

Consolatis:
> Is it possible that keyboard_modifiers_notify() is being called for all the individual keyboards in the keyboard group, and wlr_seat_set_keyboard() is cycling between them? Just speculating.

Theoretically yes but if there is just a single keyboard (e.g. the usual case) its always the same anyway, so the early out should have been used. And as we are also setting the keyboard in the key handler itself that one should also lag - but it doesn't. I really have no idea what is going on here.

jlindgren90:
> but if there is just a single keyboard (e.g. the usual case) its always the same anyway

I think I checked on my laptop once (when working on the Num Lock stuff) and there are like 6-7 different "keyboards". Random little devices like lid switch, power button, etc. seem to show up as extra "keyboards". Plus the keyboard group itself is also a wlr_keyboard, so that's at least 2 even if there is only one actual physical keyboard.

johanmalm:
When modifiers are pressed/released, both `keyboard_modifiers_notify()` AND `keyboard_key_notify()` are called.
I'm not sure if/how that explains the delay, but calling `wlr_seat_set_keyboard()` from `...key_notify()` only seems enough.

PR #1240, opened 2023-11-11
CI: run FreeBSD job on Ubuntu (to improve stability)

jbeich:
https://github.com/vmactions/freebsd-vm/issues/74#issuecomment-1798014978

Also drop `llvm15` (via `mesa-dri`) to shave CI time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,6 +89,7 @@ jobs:
           usesh: true
           prepare: |
             sed -i '' 's/quarterly/latest/' /etc/pkg/FreeBSD.conf
+            pkg set -yn pkg:mesa-dri # hack to skip llvm dependency
             pkg install -y git meson gcc pkgconf cairo pango evdev-proto \
               hwdata wayland-protocols wlroots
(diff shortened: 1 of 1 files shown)


Consolatis:
First of all: thanks as always :)

Looks like it would have went through in about 25 minutes rather than 20. I assume this might play a role here:
>  WARNING  KVM acceleration not available, using 'qemu'

Maybe we should just increase the job timeout to 30 for now.

jbeich:
Dropping `mesa-dri` somehow shaved 10 minutes instead of 2 minutes I've estimated based on log timestamps.

Background: FreeBSD < 14 packages use XZ while FreeBSD >= 14 use ZSTD. [Subpackaging](https://reviews.freebsd.org/D40549) isn't supported yet, so all LLVM packages are currently big and a bit slow to extract/install.

Consolatis:
This is still a bit flaky depending on which runner it is executed on. Apparently some have support for nested virtualization while others do not. Thus the runtime is either about 4 minutes (which is great!) and 16 minutes (which still works).

Merging.

PR #1245, opened 2023-11-14
Fix small key-repeat regression

johanmalm:
@jlindgren90  - you're review would be much appreciated if you've got time. Full explanation in commit message.

### DIFF
--- a/src/input/key-state.c
+++ b/src/input/key-state.c
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+#include <wlr/util/log.h>
 #include "input/key-state.h"
 
 #define MAX_PRESSED_KEYS (16)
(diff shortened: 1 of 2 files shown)


jlindgren90:
The logic makes sense, I'll look closer at the code later when not on mobile.

jlindgren90:
The 2nd commit LGTM. I feel like `handle_compositor_keybindings()` could use to be broken up a bit -- all the various `return` and `goto` are getting hard to keep track of -- but no need to hold up the regression fix for that.

No opinion on the 1st (logging) commit (were you planning on merging that?)

jlindgren90:
Possible refactor:
https://github.com/jlindgren90/labwc/compare/1ffca818f86f5bb9b256ba0e3ff7e49bc03b6ef2...f7ea02fd00dec5845337d4aa8a2fd53d2d412c71

I could create a pull request later, maybe it should wait till after 0.6.6 though.

jlindgren90:
Question: should we also call `key_state_store_pressed_key_as_bound()` before `handle_menu_keys()`? That can also result in actions being run, I believe.

johanmalm:
> Question: should we also call `key_state_store_pressed_key_as_bound()` before `handle_menu_keys()`? That can also result in actions being run, I believe.

Great catch. Have added.

johanmalm:
> Possible refactor: [jlindgren90/labwc@1ffca81...f7ea02f](https://github.com/jlindgren90/labwc/compare/1ffca818f86f5bb9b256ba0e3ff7e49bc03b6ef2...f7ea02fd00dec5845337d4aa8a2fd53d2d412c71)
> 
> I could create a pull request later, maybe it should wait till after 0.6.6 though.

The new PR looks great at a glance. Will review properly later. But, yeah I prefer to just honor the cool-down and make minimal/safe changes. It's so easy to overlook something simple in the refactoring. Let's pull that one in two weeks.

PR #1247, opened 2023-11-15
keyboard: break up handle_compositor_keybindings()

jlindgren90:
This function has grown quite large over time. Breaking out various smaller functions makes the logic easier to follow.

Includes the regression fix from #1245 (though it is reworked a bit).
    
Beyond that fix, there is no functional change intended. Minor logical changes do include:
    
- `key_state_store_pressed_key_as_bound()` is now called before executing any menu actions. This ensures the key states are consistent beforehand, similar to the fix for keybinding actions.
    
- Due to factoring out `match_keybinding()`, each keypress can only match a single keybinding now. Previously, it was theoretically possible for a single keypress to map to multiple keysyms which could each match a different keybinding.


jlindgren90:
@johanmalm I like the end result of this better than the minimal fix in #1245, but I'm not sure what our tolerance for further changes is at this point with 0.6.6 coming up. I'm fine with just merging #1245 as-is and leaving this for later if you would prefer.

jlindgren90:
This should be ready for review now.

johanmalm:
Thanks. I'll review tomorrow.

johanmalm:
Thanks. Much cleaner.
I really don't like long functions :smile:

PR #1251, opened 2023-11-21
Keep xwayland stacking order in sync when switching workspaces

Consolatis:
Until we expose the workspaces to xwayland we need a way to
ensure that xwayland views on the current workspace are always
stacked above xwayland views on other workspaces.

If we fail to do so, issues arise in scenarios where we change
the mouse focus but do not change the (xwayland) stacking order.

Reproducer:
- If followMouse is enabled, raiseOnFocus must be disabled
- Open at least two xwayland windows which allow scrolling
  (some X11 terminal with `man man` for example)
- Switch to another workspace, open another xwayland window
  which allows scrolling and maximize it
- Switch back to the previous workspace with the two windows
- Move the mouse to the xwayland window that does <b>not</b> have
  focus
- Start scrolling
- All scroll events should end up on the maximized window on
  the other workspace

This patch fixes the issue by simply raising all windows from
the current workspace again in their original stacking order
when switching workspaces.

Reported-by: Domo via IRC (thanks!)

### DIFF
--- a/include/common/array.h
+++ b/include/common/array.h
@@ -30,4 +30,15 @@ wl_array_len(struct wl_array *array)
 	return array->size / sizeof(const char *);
 }
 
+/**
+ * Iterates in reverse over an array.
+ * @pos: pointer that each array element will be assigned to
+ * @array: wl_array to iterate over
+ */
+#define wl_array_for_each_reverse(pos, array)                                          \
+	for (pos = !(array)->data ? NULL                                               \
+		: (void *)((const char *)(array)->data + (array)->size - sizeof(pos)); \
+		pos && (const char *)pos >= (const char *)(array)->data;               \
+		(pos)--)
+
 #endif /* LABWC_ARRAY_H */
--- a/include/view.h
+++ b/include/view.h
@@ -222,12 +222,24 @@ struct xdg_toplevel_view {
 	struct wl_listener new_popup;
 };
 
+/* All criteria is applied in AND logic */
 enum lab_view_criteria {
+	/* No filter -> all focusable views */
 	LAB_VIEW_CRITERIA_NONE = 0,
-	LAB_VIEW_CRITERIA_CURRENT_WORKSPACE = 1 << 0,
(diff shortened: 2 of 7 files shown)


include/common/array.h
```diff
@@ -30,4 +30,15 @@ wl_array_len(struct wl_array *array)
...
+/**
+ * Iterates in reverse over an array.
+ * @pos: pointer that each array element will be assigned to
+ * @array: wl_array to iterate over
+ */
+#define wl_array_for_each_reverse(pos, array)                                          \
+	for (pos = !(array)->data ? NULL                                               \
+		: (void *)((const char *)(array)->data + (array)->size - sizeof(pos)); \
+		pos && (const char *)pos >= (const char *)(array)->data;               \
+		pos--)
```
Consolatis, line 42: A few more eyes on this would be great

Consolatis:
An alternative approach would be to only use `wlr_xwayland_surface_restack()` for xwayland views but I fear that may introduce hard to debug stacking-order-gets-out-of-sync issues in the future so I opted for just raising all views from the current workspace instead. Would be fine changing to that approach though.

include/common/array.h
johanmalm: Macro looks good. I've played with it, and it seems to work well.

It's unfortunate that we have to deal with the `!data` case and also that we have to involve `sizeof(pos)`, but I think you've come up with the simplest way.

We should enclose the final pos in brackets (`(pos)--`). Partly because it's good practice to do that for parameters in macros but also to be consistent with the `wl_array_for_each()` macro.

johanmalm:
LGTM
Thanks for this patch - it's looks very tidy.
I agree with approach over `wlr_xwayland_surface_restack()`. Messing with the internal `XServer` stacking would worry me :smile:

domo141:
works for me -- i.e. cannot reproduce the problem w/ labc built from the tree of this commit

include/common/array.h
Consolatis: changed

Consolatis:
@domo141 thanks for confirming that this PR fixes the issue.

PR #1252, opened 2023-11-21
xwayland: do not re-set deocrations on unminimize

johanmalm:
Reproduce issue by:

1. Run an XWayland client
2. ToggleDecorations twice to hide deco
3. Minimize/unminimize from panel
4. Observe that deco is back

Reported-by: @Consolatis


Consolatis:
Didn't test but makes sense. LGTM

PR #1253, opened 2023-11-25
keyboard: allow back by arrow-up or arrow-left in cycle view OSD

jp7677:
All non-modifier keys cycle forward in the cycle view OSD which makes sense for e.g. tab but is not very intuitive for arrow-up or arrow-left. Handle those keys separately to provide a feel of navigation by arrow keys in the cycle view.

Sometimes I need a while to find the window in the OSD I want to focus to. Being used to navigate by arrow keys, I got slightly confused that both arrow-down and arrow-up cycle forward. This patch proposes to allow cycling backwards also by arrow-up or arrow-left, next to shift + any-key.

PS: This is no 0.6.6 material imho.

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -354,8 +354,17 @@ handle_cycle_view_key(struct server *server, struct keyinfo *keyinfo)
 	}
 
 	/* cycle to next */
-	bool backwards = keyinfo->modifiers & WLR_MODIFIER_SHIFT;
 	if (!keyinfo->is_modifier) {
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks for working on this. Will have a look after `0.6.6` (and possibly the beginning of the 0.17 port) if that's okay.

jp7677:
Thanks! Rebased and adjusted to keyboard refactor.

johanmalm:
Thanks

Issue #661, opened 2022-12-03
xfce4-panel


johanmalm:
This is just to share that there is a version of `xfce4-panel` which works on `labwc`. It was shared on the #labwc IRC channel last night and I thought some build instruction and references were worthy of some wider publicity :smile:

https://gitlab.xfce.org/xfce/xfce4-panel/-/merge_requests/103

@Tamaranch - copied for info as the main dev.

These are the steps I took on Arch Linux to get it going:

1. I installed the following from the AUR:

- xfce4-dev-tools-devel
- garcon-devel
- libxfce4util-devel
- libxfce4ui-devel

2. I then ran the following:

```
pacman -S exo gtk-layer-shell
```

```
git clone https://gitlab.xfce.org/kelnos/libxfce4windowing
cd libxfce4windowing
./autogen --prefix=/usr
make
sudo make install
cd ..
```

```
git clone --branch=wayland-xfw https://gitlab.xfce.org/Tamaranch/xfce4-panel
cd xfce-panel
./autogen --prefix=/usr
make
sudo make install
```

johanmalm:
https://user-images.githubusercontent.com/6955353/205457348-e9cd56a0-a0f6-4f19-83ff-d4980e8ed597.mp4

johanmalm:
[recording.webm](https://user-images.githubusercontent.com/6955353/205457696-fb4e6a5d-621d-400b-a111-5f5a73f88ee9.webm)

johanmalm:
@Tamaranch  It works very well. Most impressive.

Found one issue - right clicking on a plugin and select 'move' results in a crash. Not looked at the code (yet).
[recording.webm](https://user-images.githubusercontent.com/6955353/205461836-2323a4a5-239a-4aa1-97ea-24162c5208ae.webm)

johanmalm:
@Tamarance also, there seems to be an occasional glitch with foreign-toplevel in that client `title` / `app_id` is missing for a short while. It might be an out-by-one protocol read/write type issue.

In this example, I've just opened to terminals and you can same the title/app_ids missing.

![20221203_21h00m08s_grim](https://user-images.githubusercontent.com/6955353/205461976-46343195-e11c-41f4-b571-d988c0a8d662.png)

Tamaranch:
> Can raise issues on xfce.gitlab if you prefer

Yes please add new comments on the merge request, it is better to have it centralized there if possible (and we can make threads there too, it's convenient in this case).

[...]

PR #1254, opened 2023-11-25
xwayland: assign view->surface earlier in map handler

johanmalm:
...as it needs to be set before honouring xwayland_surface->fullscreen because that calls desktop_update_top_layer_visiblity() which relies on view->surface being set in view_is_focusable() since 13d0b14.

Fix bug introduced by PR #1237 which fails to hide the xfce4-panel (or any other layer-shell client in the top layer) whilst gaming in fullscreen. The bug can be observed with the following games:

- Alan Wake 2 (wine)
- Starfield (steam+proton)
- Cyberpunk (steam+proton)
- Quake 1 Remaster (steam-native)

Fixes: #661 (the last bit of it)

Reported-by: @ScarecrowDM
Helped-by: @Consolatis

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -518,6 +518,27 @@ xwayland_view_map(struct view *view)
 	ensure_initial_geometry_and_output(view);
 	wlr_scene_node_set_enabled(&view->scene_tree->node, true);
 
+	if (view->surface != xwayland_surface->surface) {
+		if (view->surface) {
+			wl_list_remove(&view->surface_destroy.link);
+		}
+		view->surface = xwayland_surface->surface;
+
+		/* Required to set the surface to NULL when destroyed by the client */
+		view->surface_destroy.notify = handle_surface_destroy;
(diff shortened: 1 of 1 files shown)


johanmalm:
@ScarecrowDM any chance you could test with this patch?

Consolatis:
Could reproduce the issue by using a panel in top layer mode + `GDK_BACKEND=x11 mate-terminal --full-screen`.
This patch hides the panel correctly for me when mate terminal starts up.

There is another issue in that using `A-Space` to minimize a fullscreen window does not trigger the top layer update routine but I think that is a minor issue and can be until after the release.

ScarecrowDM:
Just gave it a try and indeed the panel is hiding correctly now.
Thank you! :tada:

johanmalm:
> Just gave it a try and indeed the panel is hiding correctly now. Thank you! 🎉

Beautiful. Thank you both

PR #1256, opened 2023-11-26
Add omnipresent flag for views

bnason:


docs/labwc-actions.5.scd
```diff
@@ -133,6 +133,9 @@ Actions are used in menus and keyboard/mouse bindings.
 	when defining window-rules for desktop-management tools that do not
 	support the wlr-layer-shell protocol.
 
+*<action name="ToggleOmnipresent" />*
+	Toggle omnipresent for the focused window.
```
Consolatis, line 137: Suggest something like
> Toggle visible-on-all-workspaces for the focused window.

I think we should not assume that the word "Omnipresent" is known widely to mean on-all-workspaces. My documentation skills aren't that great though so a better text is also welcome :)

include/view.h
```diff
@@ -142,6 +142,7 @@ struct view {
 	bool minimized;
 	enum view_axis maximized;
 	bool fullscreen;
+	bool omnipresent;
```
Consolatis, line 145: Suggest renaming this to `visible_on_all_workspaces`

include/view.h
```diff
@@ -389,8 +390,10 @@ void view_toggle_decorations(struct view *view);
 
 bool view_is_always_on_top(struct view *view);
 bool view_is_always_on_bottom(struct view *view);
+bool view_is_omnipresent(struct view *view);
```
Consolatis, line 393: It doesn't seem like this is used anywhere so I'd suggest just removing `view_is_omnipresent`.

src/view.c
```diff
@@ -1049,6 +1049,20 @@ view_toggle_always_on_bottom(struct view *view)
...
+{
+	assert(view);
+	return view->omnipresent;
+}
+
+void
+view_toggle_omnipresent(struct view *view)
+{
+	assert(view);
+	view->omnipresent = !view_is_omnipresent(view);
```
Consolatis, line 1063: So this could then just be a `view->visible_on_all_workspaces = !view->visible_on_all_workspaces`.

src/workspaces.c
```diff
@@ -276,6 +276,16 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
...
 
+	/* Move Omnipresent views to new workspace */
+	struct view *view;
+	enum lab_view_criteria criteria =
+		LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
+	for_each_view(view, &server->views, criteria) {
+		if (view->omnipresent) {
+			view_move_to_workspace(view, target);
+		}
+	}
```
Consolatis, line 287: I think this will currently cause omnipresent views to be always on top of the stacking order when switching workspaces.
But I am not sure how we could prevent that. I am fine with merging this as is for now and think about that part later on.

Consolatis:
Very nice PR, thanks!

We still need to decouple the always-on-top one from being visible-on-all-workspaces but that one is a bit more tricky to do.
IMHO that should not prevent us from adding the real visible-on-all-workspaces implementation though.

tsujan:
As a user, I'm glad to see this activity.

My 2 cents:

I think most users are familiar with the term "sticky" (instead of "visible_on_all_workspaces") and expect that it isn't related to always-on-top either.

bnason:
@Consolatis update code for all reviewed issues and rebased on latest master

Consolatis:
> I think most users are familiar with the term "sticky" (instead of "visible_on_all_workspaces")

As we try to use the same action names as Openbox we kind of have to use the action name `ToggleOmnipresent` here. The docs now include the `sticky` part.

> and expect that it isn't related to always-on-top either.

Yes, that is a quirk from labwc and must be fixed at some point.

> @Consolatis updated code for all reviewed issues and rebased on latest master

Thanks :)

@johanmalm `ToggleOmnipresent` is listed as `5.12` in labwc scope and [marked](https://github.com/labwc/labwc-scope#5-window-actions) as category `C`. However I think we need to support this feature as it is indeed completely different to `ToggleAlwaysOnTop` (and the later behaving partly the same is basically a bug).

This PR looks ready to go from my POV if @johanmalm agrees on changing the scope of the action.
We should sync up the hardcoded default window menu and `docs/menu.xml` at some point though (and maybe use another user-facing name for `ToggleOmnipresent` in there) but we can do that later.

johanmalm:
Really nice. Thanks.

> @johanmalm `ToggleOmnipresent` is listed as `5.12` in labwc scope and [marked](https://github.com/labwc/labwc-scope#5-window-actions) as category `C`. However I think we need to support this feature as it is indeed completely different to `ToggleAlwaysOnTop` (and the later behaving partly the same is basically a bug).

Have updated to a B.

> This PR looks ready to go from my POV if @johanmalm agrees on changing the scope of the action. We should sync up the hardcoded default window menu and `docs/menu.xml` at some point though.

Yep.

PR #1258, opened 2023-11-27
Use the GTK+ 3 notebook header color as the default active title color

dimkr:
The current light gray color is very very very close to the pavucontrol notebook header color, but not enough for "seamless" appearance.

Before:

before

After:

after

### DIFF
--- a/docs/themerc
+++ b/docs/themerc
@@ -14,14 +14,14 @@ padding.height: 3
 # titlebar.height:
 
 # window border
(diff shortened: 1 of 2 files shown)


johanmalm:
Fine with me. Would you mind changing it in `docs/themerc` too though?

dimkr:
@johanmalm Done

johanmalm:
Thanks.

Issue #1268, opened 2023-12-01
DMABUF feedback support


Consolatis:
[wlroots MR](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3726)
[sway scene commit](https://github.com/swaywm/sway/pull/6844/commits/9c04a1894d5989a52d6e0541a3fd04e44b479709)

- #1278

Consolatis:
As I don't really know of a way to test this I'll remove it from the 0.7.0 milestone.

johanmalm:
Closed by #2234

PR #1278, opened 2023-12-04
[wip] Add dmabuf feedback

Consolatis:
Fixes
- #1268

This turned out to be a bit more complex because the `wlr_renderer_init_wl_display()` call we used to do does not store or return the `wlr_linux_dmabuf_v1` pointer anywhere (which we need to supply to `wlr_scene_set_linux_dmabuf_v1()`).

So instead of using `wlr_renderer_init_wl_display()` we just do it manually. This is also what the the sway scene graph [patches](https://github.com/swaywm/sway/pull/6844/commits/9c04a1894d5989a52d6e0541a3fd04e44b479709) do. We could also just use both but that causes duplicated globals warnings in wlroots which might cause further hard to detect / debug issues.

As for how to actually test this.. I have no idea.

Todo:
- [ ] remove logging
- [ ] find a way to actually test this

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -209,6 +209,7 @@ struct server {
 		char pending_output_name[4096];
 	} headless;
 	struct wlr_session *session;
+	struct wlr_linux_dmabuf_v1 *linux_dmabuf;
 
 	struct wlr_xdg_shell *xdg_shell;
 	struct wlr_layer_shell_v1 *layer_shell;
--- a/src/server.c
+++ b/src/server.c
@@ -6,6 +6,7 @@
 #include <wlr/backend/headless.h>
 #include <wlr/backend/multi.h>
 #include <wlr/types/wlr_data_control_v1.h>
+#include <wlr/types/wlr_drm.h>
 #include <wlr/types/wlr_export_dmabuf_v1.h>
 #include <wlr/types/wlr_fractional_scale_v1.h>
 #include <wlr/types/wlr_gamma_control_v1.h>
@@ -35,6 +36,7 @@
 
 #define LAB_WLR_COMPOSITOR_VERSION 5
 #define LAB_WLR_FRACTIONAL_SCALE_V1_VERSION 1
+#define LAB_WLR_LINUX_DMABUF_VERSION 4
 
 static struct wlr_compositor *compositor;
 static struct wl_event_source *sighup_source;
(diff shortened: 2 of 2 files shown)


yuiiio:
can test this with weston-simple-dmabuf-feedback.
and output is same with sway, so works well.
(weston-simple-dmabuf-feedback needs revert this commit https://gitlab.freedesktop.org/wayland/weston/-/commit/34400d7d1686d53ceaf79ff1142349a1a171db68 cause assertion error, even with sway)

jp7677:
Accidentally I stumbled upon the `dmabuf-wayland` video output in `mpv`, usable by e.g. `mpv -v --no-config --hwdec=vaapi --vo=dmabuf-wayland <video-file>`.  
This output mode bails out without this PR. but works nicely with this PR applied. The `mpv` output looks normal as far as I can judge. Also the `labwc` verbose output shows no failing statements from this PR.
So I would also lean towards "This PR works as intended".

ahesford:
The `dmabuf-wayland` output has always worked for me. (It mangles HDR colors, but that's expected because it does no color management.)

jp7677:
ah yeah sorry. I though to be  clever to just set the additions from this PR into comments, but missed the fact that I had to also revert `wlr_renderer_init_wl_shm` back into `wlr_renderer_init_wl_display`.  This resulted in mpv saying `[vo/dmabuf-wayland/wayland] Compositor doesn't support the zwp_linux_dmabuf_v1 (ver. 4) protocol!` which tricked me, but clearly my mistake.

Sorry for the noise.

Consolatis:
I remember reading an anecdote about dmabuf feedback causing an overflow of the wayland connection buffer and thus wlroots killing the client. AFAIR there were talk about wlroots potentially sending too many tranches but I didn't follow up on that. That is another reason why I am a bit skeptical about this PR (I don't see it improving anything other than being able to say "look, we support a new feature" but it might actually cause issues that weren't there before).

yuiiio:
wlroots 0.18 changed ```wlr_renderer_get_dmabuf_texture_formats(renderer)```
to ```wlr_renderer_get_texture_formats(renderer, WLR_BUFFER_CAP_DMABUF)``` 
https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/0686666cf5c0e2788e27c1f9895ce801b2cc4a72

jp7677:
This might become relevant for https://github.com/ammen99/wf-recorder/releases/tag/v0.5.0 i.c.w. VAAPI hardware encoding. But I can't test this since I'm on NVIDIA/nvenc.

jp7677:
I spend this weekend setting up `labwc` master with `wlroots` 0.18 on a Raspberry Pi with a touch screen and an hdmi display. This is essentially a multi  GPU setup. 

It turns out that `dmabuf feedback` is needed to let `labwc`/`wlroots` find the correct EGL-Wayland driver. Without this PR `eglinfo` gave me the mesa `swrast` driver for EGL wayland platform, with this patch it shows the Raspberry Pi driver `drm-rp1-dsi`. 

Using `mpv --vo=gpu --gpu-context=wayland` confirmed that with this patch I was no longer using a software renderer. 

Note that this patch is only needed with `labwc` 0.8. Labwc 0.7.4 with wlroots 0.17 used the non-hardware driver.

johanmalm:
Happy to include in `0.8.1`. If we review/merge now we've got two weeks to get some feedback before the release.

jp7677:
Should I take over with a new PR with @Consolatis as co-author? Only adjustments needed are the changed function that @yuiiio already described and reviewing the log statements.

johanmalm:
> Should I take over with a new PR with @Consolatis as co-author? Only adjustments needed are the changed function that @yuiiio already described and reviewing the log statements.

That'd be great.

johanmalm:
Implemented via #2234

PR #1280, opened 2023-12-04
Add features to make window header icons more like those in Openbox

spl237:
I don't know if this is of interest, but I have been working with labwc on a system which used to run Openbox, and I have noticed a couple of features which Openbox has which are not implemented in labwc.

1) The window header bar button icon for "maximize" is only used on non-maximized windows; on windows which are already maximized, a different "restore" icon is shown.

2) Mouse pointer hover over a window header bar button is shown by a subtle change in the icon which is displayed, usually a slight increase in size.

Neither of these appear to be implemented in labwc at present, and their absence is noticeable when moving from Openbox. This PR implements both, allowing a "hover" icon to be supplied in the theme for each button, and allowing a "restore" icon to be defined for the maximize button in addition to the standard "maximize" icon.

The fallback cases for each are the same as the originals, so unless the relevant icons are defined in the theme, then no change will be seen.

I have only limited experience of the labwc code - this implementation appears to work reliably to me, and should be reasonably efficient; the additional icons are defined as parts of the scene tree for the buttons, and are enabled or disabled as appropriate - but this may not be the best way to do this.

From my reading of the scope document, this appears to be category B functionality, regarded as nice to have, so hopefully this is a useful contribution.

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -107,6 +107,54 @@ load_buttons(struct theme *theme)
 				theme->window_inactive_button_close_unpressed_image_color,
 			},
 		},
+		{
+			"menu_hover",
+			{ 0x00, 0x18, 0x3c, 0x3c, 0x18, 0x00 },
+			{
+				&theme->button_menu_active_hover,
+				theme->window_active_button_menu_unpressed_image_color,
+			},
+			{
+				&theme->button_menu_inactive_hover,
+				theme->window_inactive_button_menu_unpressed_image_color,
+			},
+		},
+		{
+			"iconify_hover",
+			{ 0x00, 0x00, 0x00, 0x00, 0x3f, 0x3f },
+			{
+				&theme->button_iconify_active_hover,
+				theme->window_active_button_iconify_unpressed_image_color,
+			},
+			{
+				&theme->button_iconify_inactive_hover,
+				theme->window_inactive_button_iconify_unpressed_image_color,
+			},
+		},
+		{
+			"max_hover",
+			{ 0x3f, 0x3f, 0x21, 0x21, 0x21, 0x3f },
+			{
+				&theme->button_maximize_active_hover,
+				theme->window_active_button_max_unpressed_image_color,
+			},
+			{
+				&theme->button_maximize_inactive_hover,
+				theme->window_inactive_button_max_unpressed_image_color,
+			},
+		},
+		{
+			"close_hover",
+			{ 0x33, 0x3f, 0x1e, 0x1e, 0x3f, 0x33 },
+			{
+				&theme->button_close_active_hover,
+				theme->window_active_button_close_unpressed_image_color,
+			},
+			{
+				&theme->button_close_inactive_hover,
+				theme->window_inactive_button_close_unpressed_image_color,
+			},
+		},
 	};
 
 	char filename[4096] = {0};
--- a/src/theme.c
+++ b/src/theme.c
@@ -96,6 +96,18 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
+			"max_toggled",
+			{ 0x3f, 0x3f, 0x21, 0x21, 0x21, 0x3f },
+			{
+				&theme->button_restore_active_unpressed,
+				theme->window_active_button_max_unpressed_image_color,
+			},
+			{
+				&theme->button_restore_inactive_unpressed,
+				theme->window_inactive_button_max_unpressed_image_color,
+			},
+		},
+		{
 			"close",
 			{ 0x33, 0x3f, 0x1e, 0x1e, 0x3f, 0x33 },
 			{
@@ -140,6 +152,18 @@ load_buttons(struct theme *theme)
 			},
 			{
 				&theme->button_maximize_inactive_hover,
+				theme->window_inactive_button_max_unpressed_image_color,
+			},
+		},
+		{
+			"max_hover_toggled",
+			{ 0x3f, 0x3f, 0x21, 0x21, 0x21, 0x3f },
+			{
+				&theme->button_restore_active_hover,
+				theme->window_active_button_max_unpressed_image_color,
+			},
+			{
+				&theme->button_restore_inactive_hover,
 				theme->window_inactive_button_max_unpressed_image_color,
 			},
 		},
--- a/src/theme.c
+++ b/src/theme.c
@@ -39,6 +39,7 @@
 
 struct button {
 	const char *name;
+	const char *alt_name;
 	char fallback_button[6];	/* built-in 6x6 button */
 	struct {
 		struct lab_data_buffer **buffer;
@@ -60,7 +61,7 @@ load_buttons(struct theme *theme)
 {
 	struct button buttons[] = {
 		{
-			"menu",
+			"menu", NULL,
 			{ 0x00, 0x18, 0x3c, 0x3c, 0x18, 0x00 },
 			{
 				&theme->button_menu_active_unpressed,
@@ -72,7 +73,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"iconify",
+			"iconify", NULL,
 			{ 0x00, 0x00, 0x00, 0x00, 0x3f, 0x3f },
 			{
 				&theme->button_iconify_active_unpressed,
@@ -84,7 +85,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"max",
+			"max", NULL,
 			{ 0x3f, 0x3f, 0x21, 0x21, 0x21, 0x3f },
 			{
 				&theme->button_maximize_active_unpressed,
@@ -96,7 +97,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"max_toggled",
+			"max_toggled", NULL,
 			{ 0x3e, 0x22, 0x2f, 0x29, 0x39, 0x0f },
 			{
 				&theme->button_restore_active_unpressed,
@@ -108,7 +109,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"close",
+			"close", NULL,
 			{ 0x33, 0x3f, 0x1e, 0x1e, 0x3f, 0x33 },
 			{
 				&theme->button_close_active_unpressed,
@@ -120,7 +121,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"menu_hover",
+			"menu_hover", NULL,
 			{ 0x00, 0x18, 0x3c, 0x3c, 0x18, 0x00 },
 			{
 				&theme->button_menu_active_hover,
@@ -132,7 +133,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"iconify_hover",
+			"iconify_hover", NULL,
 			{ 0x00, 0x00, 0x00, 0x00, 0x3f, 0x3f },
 			{
 				&theme->button_iconify_active_hover,
@@ -144,7 +145,7 @@ load_buttons(struct theme *theme)
 			},
 		},
 		{
-			"max_hover",
+			"max_hover", NULL,
 			{ 0x3f, 0x3f, 0x21, 0x21, 0x21, 0x3f },
 			{
 				&theme->button_maximize_active_hover,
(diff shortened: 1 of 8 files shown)


Consolatis:
For your first PR you chose a pretty hard to understand / deal with area of the codebase (IMO), so kudos to you for figuring out everything correctly!

From a quick look through the diff this looks good.
There are some minor coding style issues but the generic concept seems fine.
On the first glance (again just had a quick look through the diff) it seems the current hover rect (the gray half transparent overlay) is gone, I am not sure if we want to keep that one as a fallback or not.

I will do a proper review in the next days.

Also don't worry about new commits fixing random stuff, we can always clean up the history before merging.

spl237:
Thank you!

Yes, I did get rid of the existing hover overlay - I wanted to try and keep it, but the interaction between it and the restore/maximize icon just ended up with too many special cases, so I thought the code would be cleaner without - if it is wanted, I can have another go.

I can't seem to get a clean build from your build system - it is objecting to an undeclared function, but I can't work out why, as the function in question (lab_scene_tree_from_node) is declared in common/scene-helpers.h, and that is included, so I'm not sure what is going wrong there?

spl237:
> It was removed yesterday: [4189120](https://github.com/labwc/labwc/commit/41891209fc7e3a37350e5b53a31730d755c81a03)
> 
> Edit: We are currently in the process to switch to a new wlroots version (a once in a year situation) so some things are a bit less stable in the master branch as its usually the case.

Ah - that makes sense! I'll submit a fix for that in the morning in that case.

johanmalm:
This is awesome. Thank you!
This is definitely wanted - and other than pipemenus, remains on of last things I personally feel is really needed (ref: https://github.com/johanmalm/roadmap/issues/1)

Have tested with a few themes, and it runs well.

I think we should use the Openbox defaults for max-toggled... don't you think?
https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/obrender/theme.c#L489

src/theme.c
```diff
@@ -107,6 +119,66 @@ load_buttons(struct theme *theme)
...
+				&theme->button_maximize_active_hover,
+				theme->window_active_button_max_unpressed_image_color,
+			},
+			{
+				&theme->button_maximize_inactive_hover,
+				theme->window_inactive_button_max_unpressed_image_color,
+			},
+		},
+		{
+			"max_hover_toggled",
```
jlindgren90, line 159: Something to be aware of is that Openbox used to use "max_toggled_hover" (and this is still what is in the Openbox wiki). A refactoring commit[1] changed it to "max_hover_toggled", which broke existing themes. I haven't done a survey of which themes have or haven't been updated since then, but we may want to consider trying both "max_hover_toggled" and "max_toggled_hover".

[1] https://github.com/danakj/openbox/commit/35e92e4c2a45b28d5c2c9b44b64aeb4222098c94

spl237:
> I think we should use the Openbox defaults for max-toggled... don't you think? https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/obrender/theme.c#L489

Certainly - I will pull that in tomorrow. I thought OB already had a default, but I hadn't had a chance to look for it.

One thing to consider is that I believe the fallback icons in labwc are all 6x6 - is that correct? I don't know how much work it is to expand fallbacks to accept a larger size of icon so that the hover icon can be larger, as it is in OB?

src/theme.c
johanmalm: Ah, so we might soon be more compliant with Openbox spec that openbox itself?
I did a search of my themes and could only find `max_toggle_hover.xbm`

jlindgren90: Indeed :)

Openbox still has an open bug report about this (from 2015) but development there seems to have been dead for a few years now, and even the bug tracker isn't working properly now.

spl237: On Raspberry Pi Desktop, I have only ever seen max_hover_toggled, and I've been working on it since 2014... That said, we have only ever really used our own custom theme.

johanmalm:
Ref: http://openbox.org/wiki/Help:Themes#Button_images

Comments from quick review:

- [ ] Add additionally supported buttons to `labwc-theme(1)` https://github.com/labwc/labwc/blob/3ca0d8a171bdcaa94dc871d3cb7b04f2248053bb/docs/labwc-theme.5.scd?plain=1#L196-L199
- [ ] Suggest `s/add_alt_icon/add_toggled_icon/`
- In `struct ssd_button` I suggest the following name changes:
    - [ ] `s/icon/default/` (or possibly `natural` or `normal`)
    - [ ] `s/alticon/toggled/`
    - [ ] `s/althover/toggled_hover/`
- [ ] I think we should try to get rid of `old_node` and `maximized` in `struct ssd_hover_state`. If feels like we could do this more generically. This is mostly likely outside the scope of the PR - so we either do it first - or sweep it up later. Could be done with a pointer-to-a-pointer `**icon` which we just point to &normal, &toggled, &hover, &toggled_hover in response to cursor movement. Openbox also supports foo_pressed.xbm - so in the long run I think this type of approach is better.
- [ ] We ought to setup a fallback hierarchy (like use `close` if there is no `close_hover`), instead of falling back to built-in xbms for the _hover varieties. Why? Because currently, if a theme is incomplete and for example misses a `close_hover.xbm` we end up with a mix of styles (the close.xbm of the theme and the built-in Clearlooks version of close_hover.xbm). Happy to leave this to a separate commit if needed. Off the top of my head, it's probably best done after load_buttons(). Maybe just check all the `lab_data_buffer`s in `struct theme` populate the empty ones by just copy the filled ones.

Have just tested with Openbox... by default (with Clearlooks) the hover icons are not bigger than the normal ones.

Consolatis:
Didn't think though through completely but I think we could do something like this here:

We do have checks to enable `set_squared_corners()` which basically makes the titlebar react to being maximized / unmaximized. I think we could either (ab)use the `ssd->state.squared_corners != maximized` condition to also switch between the two icons or do something similar for the icon (which seems cleaner). In regards to the hover action, we can put both icons into a tree node and then toggle the visibility of the tree node on hover in/out. That way we have the one part of the code which reacts to the maximized state toggle the visibility of the two icons withing the tree and another one (hover) toggle the generic visibility of the tree itself.

src/theme.c
Consolatis: > but we may want to consider trying both "max_hover_toggled" and "max_toggled_hover".

Agree, lets just support both. Its just the question which one we do mention in `docs/labwc-theme.5.scd`. In either case we should then also mention that the other one is supported as well for backwards compatibility with existing themes.

spl237:
OK, I've made most of the changes suggested above. I've changed the function name so it builds against recent wlroots; I've updated the variable and function names as per Johan's suggestions ("default" is a keyword, so I went with "normal" for the standard icon), I've used the default Openbox toggled icon as the fallback and have added the new icons to the documentation.

The other changes suggested above are probably more appropriate for people who know the code better than I do, so if you don't mind, I'll leave those to you. If there is anything else you need from me to get the PR accepted, please let me know - or indeed, if you would prefer to simply implement the functionality in some other way, I shan't be offended!

johanmalm:
Again, just to say thank you for this work. I'm happy to merge with one exception... the naming of the `xbm` files.

I prefer to use `max_toggled_hover.xbm` instead of `max_hover_toggled.xbm` (but with a fallback to `max_hover_toggled.xbm`.

Why?

- Because we are more like to avoid complaints about regression that way
  - Because the openbox.org wiki says so.
  - Because the only two themes I could find those buttons (see ref below) use `max_toggled_hover.xbm`
  - Because Openbox is patched on Arch Linux (and I think Debian too) only support `max_toggled_hover.xbm` (I've just logged into Openbox on Arch to check)

I'm guessing RPI build their own version of openbox without said patches?

This took me a while to figure out because the source code (`openbox/obrender/theme.c`) parses with `hover` before `toggled` - whereas distros appear to apply patches.

References:
- https://github.com/b1ack0p/pixflat-theme/tree/main/themes/PiX/openbox-3
- https://github.com/the-zero885/Lubuntu-Arc-Round-Openbox-Theme/tree/master/Lubuntu%20Arc-Dots%20Ligth/openbox-3
- https://gitlab.archlinux.org/archlinux/packaging/packages/openbox/-/blob/main/debian-887908.patch?ref_type=heads

johanmalm:
Not sure :smile:

Maybe add `const char *name_compatibility` to `struct button` ...and then do something at the bottom of the for-loop in `load_buttons()`

But it would probably require some refactoring... possibly a version of `button_xbm_load()` without the fallback (so that we know if the load worked or not).

I can have a play tomorrow night if you prefer... although suspect you'd do a better job of it :smile:

spl237:
OK, I have no idea at all how the theme icon loading code works, but I can have a look in the morning. If I can't figure it out, is it ok if I just change the default name from max_hover_toggled to max_toggled_hover and resubmit? I.e. make the change but without the fallback? That's then easy enough for me to cope with in RPIOS.

But I'll see what I can do in terms of getting the fallback working first.

johanmalm:
Yes, happy with name change only. I could add fallback once merged because I think it is neat support both.

spl237:
OK, one way or the other, I'll update the PR tomorrow... :)

spl237:
This should work to support either name now, I think.

johanmalm:
Thanks! 
And great to have you onboard.
This is a feature I've been meaning to add for a long time :smile:

spl237:
Excellent - very glad to be able to contribute!

stefonarch:
It's a nice feature but: it breaks any theme without hover icons. IMO there has to be more attention in releases: never break user experience, it happened also with numlock=active by default in 0.6.6


https://github.com/labwc/labwc/assets/10681413/2824d575-9240-43af-b932-adcc92b9d639

Consolatis:
> It's a nice feature but: it breaks any theme without hover icons.

Agree, we need to adjust this a bit more, personally I'd also like the old hover overlay back. We'll figure something out before 0.7.0. Ideas welcome.

> IMO there has to be more attention in releases: never break user experience, it happened also with numlock=active by default in 0.6.6

To a point we depend on user feedback to ensure that we don't unnecessarily break user experience. We usually try pretty hard not to break old configs (for example when renaming config options we usually keep supporting the old ones but log a deprecation error).

stefonarch:
Yeah, this needs refactoring IMO; fiddling with my theme I noticed it breaks also svg icon hover always, png probably too.

Consolatis:
Maybe we could try to load the `_hover` icons if available and if not simply fall back to the old overlay?

spl237:
> Maybe we could try to load the `_hover` icons if available and if not simply fall back to the old overlay?

I'm fiddling with this now - I have the overlay fallback working; just working on what to do in the theme.

Expect a PR in the next day or so...

spl237:
https://github.com/labwc/labwc/pull/1296 now submitted - should do what is needed.

Consolatis:
> #1296 now submitted - should do what is needed.

Thanks, I'll have a look later and then possibly update #1289

src/theme.c
johanmalm: I've just realised that the same applies to `themerc` variables, for example `window.active.button.toggled.hover.bg`.

http://openbox.org/wiki/Help:Themes#window.active.button.toggled.hover.bg

The Arch/debian patches covers that too:

```
--- a/obrender/theme.c
+++ b/obrender/theme.c
@@ -1532,8 +1534,8 @@ static void read_button_styles(XrmDataba
     READ_BUTTON_APPEARANCE("disabled", disabled, 0);
     READ_BUTTON_APPEARANCE("hover", hover, 0);
     if (toggled_mask) {
-        READ_BUTTON_APPEARANCE("unpressed.toggled", unpressed_toggled, 1);
-        READ_BUTTON_APPEARANCE("pressed.toggled", pressed_toggled, 0);
-        READ_BUTTON_APPEARANCE("hover.toggled", hover_toggled, 0);
+        READ_BUTTON_APPEARANCE("toggled.unpressed", unpressed_toggled, 1);
+        READ_BUTTON_APPEARANCE("toggled.pressed", pressed_toggled, 0);
+        READ_BUTTON_APPEARANCE("toggled.hover", hover_toggled, 0);
     }
 }
```

Let's go for the following syntax:

`BUTTON [TOGGLED] [STATE]`

where
- `BUTTON` can be one of 'max', 'iconify', 'close', 'menu'
- `TOGGLED` is either 'toggled' or nothing
- `STATE` is 'hover' or nothing. In future, 'pressed' may be supported too.

I've done a `ack 'toggled\.hover'` on `/usr/share/themes/` and `~/.local/share/themes` with lots of its. The equivalent with `hover\.toggled` finds nothing. I obviously can't tell to what extent it's representative of the wider community, but my theme folders contains a real mix of themes I've gathered over the years.

PR #1284, opened 2023-12-06
view: continue move-to-edge behavior to adjacent outputs

ahesford:
The `MoveToEdge` action can be extended naturally to adjacent outputs in a matter generally analogous to the behavior of `SnapToEdge`. Such a feature already exists in OpenBox. The proposed behavior is:

1. If the existing move logic determines a nonzero move in either *x* or *y*, keep the existing behavior; the view being moved will advance on the same output to the next edge.
2. When the determined move is zero in both *x* and *y*, find an output adjacent to the current output for the view along the direction in which the move is desired.
3. If an appropriate adjacent output exists, move the view onto that output, at the edge *opposite* the target direction (*i.e.*, along the edge *nearest* the original output).

When the outputs and views are well behaved, that's all there is: the window is moved. However, there are several corner cases that should be considered:
1. The view should never start left or above the usable region of the target output. (Violating this rule would imply that, neglecting any gaps, the view would actually remain on the original display, moving counter to the intended direction.)
2. If the view is more than half off the right (or bottom) edge of the usable view, it is probably best to shift it. Thus, we try to put the right (bottom) edge of the view at the right (bottom) edge of the usable region of the output, unless that would violate the first rule (in which case we apply that rule instead). The combination of these two rules means that corrections to the intended motion to ensure maximal visibility will be the less extreme of aligning the near or far edges of the view and output.

Testing this in a dual-output (side-by-side) configuration against version 0.6.6 seems to produce reasonable behavior even for sub-optimal initial window placement and sizing.

I believe these changes properly consider the effects of server-side decoration and any configured gap distance, but those additional offsets could use further review.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -1229,6 +1229,56 @@ view_on_output_destroy(struct view *view)
 	view->output = NULL;
 }
 
+static struct output *
+view_get_adjacent_output(struct view *view, enum view_edge edge)
+{
+	assert(view);
+	struct output *output = view->output;
+	if (!output_is_usable(output)) {
+		wlr_log(WLR_ERROR,
+			"view has no output, cannot find adjacent output");
+		return NULL;
+	}
+
+	/* Determine any adjacent output in the appropriate direction */
+	struct wlr_output *new_output = NULL;
+	struct wlr_output *current_output = output->wlr_output;
+	struct wlr_output_layout *layout = view->server->output_layout;
+	switch (edge) {
+	case VIEW_EDGE_LEFT:
+		new_output = wlr_output_layout_adjacent_output(
+			layout, WLR_DIRECTION_LEFT, current_output, 1, 0);
+		break;
+	case VIEW_EDGE_RIGHT:
+		new_output = wlr_output_layout_adjacent_output(
+			layout, WLR_DIRECTION_RIGHT, current_output, 1, 0);
+		break;
+	case VIEW_EDGE_UP:
+		new_output = wlr_output_layout_adjacent_output(
+			layout, WLR_DIRECTION_UP, current_output, 0, 1);
+		break;
+	case VIEW_EDGE_DOWN:
+		new_output = wlr_output_layout_adjacent_output(
+			layout, WLR_DIRECTION_DOWN, current_output, 0, 1);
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks for working in this.

> I believe these changes properly consider the effects of server-side decoration and any configured gap distance, but those additional offsets could use further review.

Yes, from a first glance these look good.

I will have a proper look through the diff later on.

ahesford:
Made a couple of tweaks:
1. I think `output_usable_area_in_layout_coords`, not `output_usable_area_scaled`, is the right function to get the target region when jumping outputs.
2. I cleaned up the "shift into region" logic when the view either starts above or left of the usable origin, or the view is more than half outside the usable region. The math is the same, but it should be a bit easier to read and interpret now.

src/view.c
```diff
@@ -1261,7 +1311,75 @@ view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_wind
...
+	struct border margin = ssd_get_margin(view->ssd);
+
+	/* Bounds of the possible placement zone in this output */
+	int left = usable.x + rc.gap + margin.left;
+	int right = usable.x + usable.width - rc.gap - margin.right;
+	int top = usable.y + rc.gap + margin.top;
+	int bottom = usable.y + usable.height - rc.gap - margin.bottom;
+
+	/* Default target position on new output is current target position */
+	dx = view->pending.x;
```
johanmalm, line 1342: At this point we change the meaning of dx,dy from 'delta' (or 'distance') to 'destination'.
Might it be tidier to call them destination_x and destination_y from this point onwards?

johanmalm:
LGTM. Have tested.
Test included large window on large screen moving to a small screen (smaller than window). 

This is a great contribution. Great functionality. Combines common code. Nice and tidy.

Just one minor review comment. Can live without tweaking that, but have to admit it did confuse me for a minute when I first read through.

ahesford:
I changed the names of `dx` and `dy` for clarity, and added an intermediate `midpoint` variable that should better illustrate the logic behind the math in the edge cases.

johanmalm:
Thanks. Nice!

Issue #1285, opened 2023-12-06
gtk-layer-shell exclusive zones do not appear to work


spl237:
With the Raspberry Pi Desktop, the panel window is set to be an exclusive zone using gtk-layer-shell, and when run under wayfire, the desktop background and associated icons are pushed down so that the panel does not cover them.

Under labwc, this doesn't happen - the panel overlays the desktop and covers icons at the top of it.

Are exclusive zones not supported, or is this likely to be a bug in the implementation of the panel as one?

Consolatis:
Labwc does support layershell exclusive zones. But labwc does not prevent windows to be on top of it / under it when manually positioned. We do prevent automatically calculated geometry like maximized / tiled windows to overlap the exclusive zone however. How exactly is the "desktop background with icons" positioned and sized?

spl237:
The desktop is a layer-shell window which is anchored to all four edges of the screen and set to the background layer; the panel is a layer-shell window which is anchored to either the top or bottom of the screen and with exclusive zone set.

Consolatis:
hm.. you are right. We do not update the layershell background layer for new exclusive zones.
This has to be fixed. Thanks for the report!

Consolatis:
Just found this in the wlroots sources:
```
// If the exclusive zone is set to -1, the layer surface will use the
// full area of the output, otherwise it is constrained to the
// remaining usable area.
```
swaybg for example sets it to `-1` which results in not reacting to exclusive zone clients.
There is also an issue in labwc in that it calculates the usable area starting from bottom layer to overlay layer so that even with manually hacking wlroots to always use the usable area it would still not work for non-exclusive background layers with exclusive bottom or top layer clients.

The following PR should fix the issue:
- #1286

spl237:
Tested here - fixes my problem! Many thanks for the very prompt fix.

Consolatis:
Thanks for testing. We have a no-self-merge policy (unless its a crash fix or some minor stuff) so it might take a bit to have the fix in the master branch.

spl237:
No problem at all - we're not in any hurry!

Thanks again.

spl237:
Closing as patch fixes the issue.

PR #1286, opened 2023-12-06
src/layers.c: calculate usable_area before positioning clients

Consolatis:
This ensures that the usable area is completely calculated before non exclusive-zone clients are positioned / resized.

Fixes
- #1285

Reported-by: @spl237


spl237:
Great! Fixes my desktop issue - many thanks!

johanmalm:
Thanks!

PR #1287, opened 2023-12-06
Implement key binds to control virtual outputs

### LINKED ISSUE #1279: Adding and removing virtual outputs (headless)
That would be very handy if labwc supported commands like `VirtualOutputAdd` and `VirtualOutputRemove` that could be bound to keys.

Having such functionality (maybe even barebones) would allow feedback and scenarios from users.

For example, my use case would allow sharing only specific region of a screen during sharing/casting:

1) Create a virtual output via keybind.
2) Set it up using `wlr-randr` or `wdisplays` to overlap real output, but have a smaller vertical resolution. Effectively cutting the bottom part of the screen where panel resides.
3) During screen sharing/casting, choose virtual output.
4) When done, remove virtual output via keybind.

@Consolatis had a PoC patch that creates hardcoded virtual output. I've tried it and it worked well for my scenario.

Extended (not barebones) version could allow specifying virtual output name (for `kanshi` config, for example) and (maybe) resolution/offset/dpi (although this is probably a job for kanshi).

kyak:
It is now possible to use keybinds to add and remove virtual outputs (also called headless backend in wlroots terminology).

Closes #1279 

Thanks to @Consolatis for original PoC patch and guidance!

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -216,8 +216,8 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="VirtualOutputRemove" output_name="value" />*
 	Remove virtual output (headless backend).
 
-	*output_name* The name of virtual output. If not supplied, will remove first
-	found virtual output.
+	*output_name* The name of virtual output. If not supplied, will remove the
+	last virtual output added.
 
 *<action name="None" />*
 	If used as the only action for a binding: clear an earlier defined binding.
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -183,11 +183,10 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="VirtualOutputAdd" output_name="value" />*
 	Add virtual output (headless backend).
 
-	Virtual outputs is a useful mechanism. For example, it can be used to overlay
-	virtual output on real output, but with a different resolution (this can be
-	done with `wlr-randr` or `wdisplays`). After that, virtual output can be
-	selected for screen sharing (casting), effectively sharing only the region of
-	the screen.
+	For example, it can be used to overlay virtual output on real output, but with
+	a different resolution (this can be done with `wlr-randr` or `wdisplays`).
+	After that, virtual output can be selected for screen sharing (casting),
+	effectively sharing only the region of the screen.
 
 	It must be noted that overlaying virtual output and real output is not
 	endorsed or explicitely supported by wlroots. For example, after configuring
@@ -207,9 +206,9 @@ Actions are used in menus and keyboard/mouse bindings.
 	Note that the vertical resolution of "ScreenCasting" output is just 50px
 	smaller than "eDP-1" output to cut off bottom panel from screen sharing.
 
-	This setup is also useful for extending the desktop to (maybe mobile) remote
-	systems like tablets. E.g. simply adding a virtual output, attaching wayvnc to
-	it and running a VNC client on the remote system.
+	Virtual output is also useful for extending the desktop to (maybe mobile)
+	remote systems like tablets. E.g. simply adding a virtual output, attaching
+	wayvnc to it and running a VNC client on the remote system.
 
 	*output_name* The name of virtual output. Providing virtual output name is
 	beneficial for further automation. Default is "HEADLESS-X".
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -180,6 +180,46 @@ Actions are used in menus and keyboard/mouse bindings.
 	*wrap* [yes|no] Wrap around from last desktop to first, and vice
 	versa. Default yes.
 
+*<action name="VirtualOutputAdd" output_name="value" />*
+	Add virtual output (headless backend).
+
+	Virtual outputs is a useful mechanism. For example, it can be used to overlay
+	virtual output on real output, but with a different resolution (this can be
+	done with `wlr-randr` or `wdisplays`). After that, virtual output can be
+	selected for screen sharing (casting), effectively sharing only the region of
+	the screen.
+
+	It must be noted that overlaying virtual output and real output is not
+	endorsed or explicitely supported by wlroots. For example, after configuring
+	virtual output, real output must be reconfigured as well (for the overlay
+	configuration to work correctly). This is the example configuration:
+
+```
+<keybind key="W-v">
+  <action name="VirtualOutputAdd" output_name="ScreenCasting"/>
+  <action name="Execute" command='sh -c "wlr-randr --output ScreenCasting --pos 0,0 --scale 2 --custom-mode 3840x2110; wlr-randr --output eDP-1 --pos 0,0 --scale 2 --mode 3840x2160"'/>
+</keybind>
+<keybind key="W-c">
+  <action name="VirtualOutputRemove"/>
+</keybind>
+```
+
+	Note that the vertical resolution of "ScreenCasting" output is just 50px
+	smaller than "eDP-1" output to cut off bottom panel from screen sharing.
+
+	This setup is also useful for extending the desktop to (maybe mobile) remote
+	systems like tablets. E.g. simply adding a virtual output, attaching wayvnc to
+	it and running a VNC client on the remote system.
+
+	*output_name* The name of virtual output. Providing virtual output name is
+	beneficial for further automation. Default is "HEADLESS-X".
+
+*<action name="VirtualOutputRemove" output_name="value" />*
+	Remove virtual output (headless backend).
+
+	*output_name* The name of virtual output. If not supplied, will remove first
+	found virtual output.
+
 *<action name="None" />*
 	If used as the only action for a binding: clear an earlier defined binding.
 
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -471,6 +471,8 @@ struct wlr_box output_usable_area_in_layout_coords(struct output *output);
 struct wlr_box output_usable_area_scaled(struct output *output);
 void handle_output_power_manager_set_mode(struct wl_listener *listener,
 	void *data);
+void virtual_output_add(struct server *server, const char *output_name);
+void virtual_output_remove(struct server *server, const char *output_name);
 
 void server_init(struct server *server);
 void server_start(struct server *server);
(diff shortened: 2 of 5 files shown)


docs/labwc-actions.5.scd
```diff
@@ -180,6 +180,46 @@ Actions are used in menus and keyboard/mouse bindings.
 	*wrap* [yes|no] Wrap around from last desktop to first, and vice
 	versa. Default yes.
 
+*<action name="VirtualOutputAdd" output_name="value" />*
+	Add virtual output (headless backend).
+
+	Virtual outputs is a useful mechanism. For example, it can be used to overlay
+	virtual output on real output, but with a different resolution (this can be
```
Consolatis, line 187: Suggest removing the first sentence about the usefulness.

Could be rewritten like
> Virtual outputs can be used for example to overlay a virtual output on a real output, but with a different resolution [...]

src/action.c
```diff
@@ -617,6 +629,58 @@ run_if_action(struct view *view, struct server *server, struct action *action)
 	}
 }
 
+static void
+virtual_output_add(struct server *server, const char *output_name)
```
Consolatis, line 633: Suggest moving this function and `virtual_output_remove()` to `src/output.c`.

src/output.c
```diff
@@ -194,14 +196,21 @@ new_output_notify(struct wl_listener *listener, void *data)
 	struct server *server = wl_container_of(listener, server, new_output);
 	struct wlr_output *wlr_output = data;
 
+	/*
+	 * Name virtual outputs.
+	 */
+	if (wlr_output_is_headless(wlr_output) && server->headless.pending_output_name[0] != '\0') {
+		wlr_output_set_name(wlr_output, server->headless.pending_output_name);
+	}
```
Consolatis, line 204: Although we do not currently support any other way to add virtual outputs I'd still suggest `server->headless.pending_output_name[0] = '\0'` here so that we don't run into issues when we add other ways to add virtual outputs in the future.

Consolatis:
> Do I need to squash everything into a single commit?
@kyak up to you. In general yes but we can also do that when merging.

As this is a feature that I helped to design and we usually don't self-merge I'd prefer if @johanmalm could have a final look over this PR. From my POV this is ready to go after squashing.

include/labwc.h
```diff
@@ -467,6 +471,8 @@ struct wlr_box output_usable_area_in_layout_coords(struct output *output);
 struct wlr_box output_usable_area_scaled(struct output *output);
 void handle_output_power_manager_set_mode(struct wl_listener *listener,
 	void *data);
+void virtual_output_add(struct server *server, const char *output_name);
```
johanmalm, line 474: As these functions are in `output.c` their function names ought to begin with `output_`

src/output.c
```diff
@@ -194,14 +196,22 @@ new_output_notify(struct wl_listener *listener, void *data)
 	struct server *server = wl_container_of(listener, server, new_output);
 	struct wlr_output *wlr_output = data;
 
+	/*
```
johanmalm, line 199: Suggest a one-line comment (`/* Name virtual output */`)

Also `s/outputs/output/` ?

src/output.c
```diff
@@ -768,3 +778,55 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
 		break;
 	}
 }
+
+void
+virtual_output_add(struct server *server, const char *output_name)
+{
+	if (output_name) {
+		/*
```
johanmalm, line 786: Single line comment?

src/output.c
```diff
@@ -768,3 +778,55 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
...
+		 * Prevent creating outputs with the same name
+		 */
+		struct output *output;
+		wl_list_for_each(output, &server->outputs, link) {
+			if (wlr_output_is_headless(output->wlr_output)) {
+				if (!strcmp(output->wlr_output->name, output_name)) {
+					wlr_log(WLR_DEBUG,
+					"refusing to create virtual output with duplicate name");
```
johanmalm, line 794: Missing tab.

src/output.c
```diff
@@ -768,3 +778,55 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
...
+virtual_output_add(struct server *server, const char *output_name)
+{
+	if (output_name) {
+		/*
+		 * Prevent creating outputs with the same name
+		 */
+		struct output *output;
+		wl_list_for_each(output, &server->outputs, link) {
+			if (wlr_output_is_headless(output->wlr_output)) {
+				if (!strcmp(output->wlr_output->name, output_name)) {
```
johanmalm, line 792: Consider `if (foo && bar) {}` rather than `if (foo) { if (bar) {} }`. Saves a tab on the indentation on the next line.

src/output.c
```diff
@@ -768,3 +778,55 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
...
+					wlr_log(WLR_DEBUG,
+					"refusing to create virtual output with duplicate name");
+					return;
+				}
+			}
+		}
+		strncpy(server->headless.pending_output_name, output_name,
```
johanmalm, line 799: Can't tell if `strncpy()` goes over the 4096 limit.
Suggest using `snprintf()` for terminating null byte. Could then check return value.

src/output.c
```diff
@@ -768,3 +778,55 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
...
+				sizeof(server->headless.pending_output_name));
+	} else {
+		server->headless.pending_output_name[0] = '\0';
+	}
+	wlr_headless_add_output(server->headless.backend, 1920, 1080);
```
johanmalm, line 804: Should we hard-code 1920x1080 here?

src/output.c
```diff
@@ -768,3 +778,55 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
...
+				if (!strcmp(output->wlr_output->name, output_name)) {
+					wlr_output_destroy(output->wlr_output);
+					return;
+				}
+			} else {
+				/*
+				 * When virtual output name was no supplied by user, simply
+				 * destroy the first virtual output found.
+				 */
+				wlr_output_destroy(output->wlr_output);
```
johanmalm, line 827: Outputs are added to list with `wl_list_insert` (rather than `wl_list_append`) so you'd remove the _last_ virtual output.

src/server.c
```diff
@@ -256,6 +258,24 @@ server_init(struct server *server)
 		exit(EXIT_FAILURE);
 	}
 
+	/*
+	 * Create headless backend to enable adding virtual outputs later on.
```
johanmalm, line 262: Single line comment?

src/server.c
```diff
@@ -256,6 +258,24 @@ server_init(struct server *server)
...
+	 */
+	server->headless.backend = wlr_headless_backend_create(server->wl_display);
+	if (!server->headless.backend) {
+		wlr_log(WLR_ERROR, "failed to create virtual output");
```
johanmalm, line 266: `s/virtual output/secondary headless backend for virtual outputs/` ??

src/server.c
```diff
@@ -256,6 +258,24 @@ server_init(struct server *server)
...
+		exit(EXIT_FAILURE);
+	}
+	wlr_multi_backend_add(server->backend, server->headless.backend);
+
+	/*
```
johanmalm, line 271: Sway creates a "FALLBACK" headless output at this point but does not immediate destroy it.
Do we understand why? Are your requirements different?

src/server.c
```diff
@@ -256,6 +258,24 @@ server_init(struct server *server)
...
+	 * If we don't populate headless backend with a virtual output (that we
+	 * create and immediately destroy), then virtual outputs being added
+	 * later do not work properly when overlayed on real output. Content is
```
johanmalm, line 274: `s/overlayed/overlaid/`

johanmalm:
Happy in principle. Just a bunch of fairly trivial comments.

src/server.c
Consolatis: AFAIR the fallback virtual output in sway is a hack so they don't have to deal with a situation where they don't have any more outputs around. Last I heard about that was that they want to remove it.

src/output.c
kyak: I've skipped tab here because otherwise this don't fit into limitation of 100 columns.

src/output.c
kyak: It won't go over the 4096 limit because of `sizeof(server->headless.pending_output_name)`. I double checked and `sizeof` returns 4096.

Suggest to keep it like this?

src/output.c
kyak: I thought about it. In fact, i tried to hardcode `0, 0` so the user would then decide. But wlroots is quirky - setting it to `0, 0` here disallows changing resolution from tools like wlr-randr (returns error).

Any ideas what would be a better default value?

src/output.c
kyak: Indeed. But note that it says "the first virtual output **found**"

kyak:
@johanmalm thanks! I've addressed most of your comments. Kept it in separate to-be-squashed commits for better traceability.

src/output.c
```diff
@@ -768,3 +776,52 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
...
+
+void
+output_add_virtual(struct server *server, const char *output_name)
+{
+	if (output_name) {
+		/* Prevent creating outputs with the same name */
+		struct output *output;
+		wl_list_for_each(output, &server->outputs, link) {
+			if (wlr_output_is_headless(output->wlr_output) &&
+				!strcmp(output->wlr_output->name, output_name)) {
```
johanmalm, line 788: One more tab indent on `!strcmp`

src/output.c
johanmalm: Never mind - let's leave it.
You won't know if it's been truncated, but seems pretty unlikely with that size buffer anyway.

src/output.c
johanmalm: Ah makes sense then. 1920x1080 is as good as any.

Be good to add a comment to that effect for future devs.
`/* Setting it to 0, 0 here disallows changing resolution from tools like wlr-randr (returns error) */`

src/output.c
johanmalm: I'm easy - not really a merge-stopper.

I think that changing the man-page to say "the last virtual output added" is a very good idea though. Explicit is better than implicit and all that :smile:

kyak: Good point!

johanmalm:
Great effort. Thanks!

PR #1289, opened 2023-12-06
src/ssd: refactor button hover and maximize toggle

Consolatis:
Followup from
- #1280

Only slightly tested, no functional change intended.


johanmalm:
LGTM. Didn't test.

PR #1292, opened 2023-12-07
feat: implement <font place="InactiveWindow"> (ref scope: 2.4.7)

ludg1e:
Thanks you @Consolatis for the help of how to implement it.

Is the scope repo automatically updated or should I do it?

It works:
image

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -28,7 +28,7 @@
     <font place="InactiveWindow">
       <name>sans</name>
       <size>10</size>
-      <slant>italic</slant>
+      <slant>normal</slant>
       <weight>normal</weight>
     </font>
     <font place="MenuItem">
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -25,6 +25,12 @@
       <slant>normal</slant>
       <weight>normal</weight>
     </font>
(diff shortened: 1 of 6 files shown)


docs/rc.xml.all
```diff
@@ -25,6 +25,12 @@
       <slant>normal</slant>
       <weight>normal</weight>
     </font>
+    <font place="InactiveWindow">
+      <name>sans</name>
+      <size>10</size>
+      <slant>italic</slant>
+      <weight>normal</weight>
+    </font>
```
Consolatis, line 33: Suggest reverting `slant` to `normal` here to a) match the hardcoded default in `src/config/rcxml.c` and b) to not surprise users by the sudden change in defaults.

Consolatis:
Apart from the `slant` default in the example config this looks great, thanks :)

> Is the scope repo automatically updated or should I do it?

No, that one is updated manually. Feel free to send a PR there or just wait for @johanmalm to update it later.

docs/rc.xml.all
ludg1e: Done

ludg1e: Fixed

Consolatis:
:)

PR #1296, opened 2023-12-07
Use overlay colour rather than fallback icon if hover icons not defined

spl237:
I think this should use the original hover overlay instead of the fallback XBM hover icons if hover icon files are not present in the theme. Seems to work ok for me.

If a fallback XBM has 0x40 as the first byte - which will never be the case for a real icon, as they only use the bottom 6 bits, that triggers the XBM icon loader to just set the icon buffer to NULL. If the SSD code finds a null icon buffer, it loads the overlay instead of the icon as the hover part of the scene.

The only untidy bit is because if the hover icon is used, we need to turn off the normal icon while the hover icon is displayed to prevent one showing through the other, and if the hover overlay is used, we need to leave the normal icon on. I've added a flag to the button structure to indicate which needs to happen - this is a bit clunky, so feel free to improve on it!

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -23,6 +23,7 @@ struct ssd_button {
 	struct wlr_scene_node *toggled;
 	struct wlr_scene_node *toggled_hover;
 	struct wlr_scene_node *background;
+	bool shade_hover;
 
 	struct wl_listener destroy;
 };
--- a/src/button/button-xbm.c
+++ b/src/button/button-xbm.c
@@ -294,7 +294,12 @@ button_xbm_load(const char *button_name, const char *alt_name,
 			}
 		}
 	}
+
 	if (!pixmap.data) {
+		if (fallback_button[0] == 0xFF) {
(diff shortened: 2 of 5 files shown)


spl237:
Yes, there's various other ways that could be managed in the theme - I just did something quick and dirty that works! I did try just setting the fallback icon in the button structures in the button loader to NULL, but the compiler wasn't happy with that, so I just used bit 6 in the first byte as the flag. It's not elegant, but it works - feel free to hack it about!

johanmalm:
Hi there. Have gone with different approach (PR #1339) which I think is nearly there for merge.
Appreciate this work anyway.
Are you okay if we close this one?

spl237:
> Hi there. Have gone with different approach (PR #1339) which I think is nearly there for merge.
> Appreciate this work anyway.
> Are you okay if we close this one?

If there's a better way of doing it which works, fine with me!

Consolatis:
Replaced by
- #1339

PR #1299, opened 2023-12-08
Theme option to configure boxes from workspace OSD

### LINKED ISSUE #1199: Can I hide these boxes in the Desktop Switcher OSD?
image

kyak:
This allows to configure width and height of "boxes" displayed in the workspace OSD.

Closes #1199

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -905,6 +905,12 @@ post_processing(struct theme *theme)
 		memcpy(theme->osd_border_color, theme->osd_label_text_color,
 			sizeof(theme->osd_border_color));
 	}
+	if (theme->osd_workspace_switcher_boxes_width == 0) {
+		theme->osd_workspace_switcher_boxes_height = 0;
+	}
+	if (theme->osd_workspace_switcher_boxes_height == 0) {
+		theme->osd_workspace_switcher_boxes_width = 0;
+	}
 }
 
 void
--- a/src/theme.c
+++ b/src/theme.c
@@ -362,6 +362,9 @@ theme_builtin(struct theme *theme)
 	theme->osd_window_switcher_item_padding_y = 1;
 	theme->osd_window_switcher_item_active_border_width = 2;
 
+	theme->osd_workspace_switcher_boxes_width = 20;
+	theme->osd_workspace_switcher_boxes_height = 20;
+
 	/* inherit settings in post_processing() if not set elsewhere */
 	theme->osd_bg_color[0] = FLT_MIN;
 	theme->osd_border_width = INT_MIN;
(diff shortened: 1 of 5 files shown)


1qh:
@kyak Awesome, thanks a lot

docs/rc.xml.all
```diff
@@ -104,6 +105,7 @@
     <names>
       <name>Default</name>
     </names>
+    <hideBoxes>false</hideBoxes>
```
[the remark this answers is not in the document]
kyak: I was thinking about doing it via themes. But then I thought that it is an on/off toggle and better belongs to rc.xml. Anyway, not a big deal, I'll rework this to use themes.

Consolatis:
Looking great in principle.
Something seems off with the size of the active box though:

old
20231209_22h36m20s_grim

new
20231209_22h38m34s_grim

kyak:
> Looking great in principle. Something seems off with the size of the active box though:
> 
Ah, great catch, I know why. Will fix it

johanmalm:
LGTM. Didn't test.

Yes, should always `cairo_stroke()`

kyak:
I've updated the branch. Here's the setup I've tried.

`~/.config/labwc/themerc-override`:

```
osd.workspace-switcher.boxes.width: 0
osd.workspace-switcher.boxes.height: 0
```

`~/.config/labwc/rc.xml`:

```
  <desktops number="10">
    <names>
      <name>➀·········</name>
      <name>·➁········</name>
      <name>··➂·······</name>
      <name>···➃······</name>
      <name>····➄·····</name>
      <name>·····➅····</name>
      <name>······➆···</name>
      <name>·······➇··</name>
      <name>········➈·</name>
      <name>·········➉</name>
    </names>
  </desktops>
```

```
  <theme>
    ...
    <font place="OSD">
      <name>monospace</name>
      <size>16</size>
    </font>
  </theme>
```

I don't entirely like that OSD font is the same for window switcher and workspace switcher. I'd prefer a smaller font for window switcher and bigger font for workspace switcher. But this is probably fine for the time being.

Consolatis:
> I don't entirely like that OSD font is the same for window switcher and workspace switcher. I'd prefer a smaller font for window switcher and bigger font for workspace switcher. But this is probably fine for the time being.

Yeah, having to define the fonts in the config and not the theme is somewhat strange in the first place in my opinion.
Nothing stopping us from defining another font place which inherits its default value from the OSD one though.

By the way, you don't need to define the numbers of desktops, the number is completely optional, same for the names.
It's enough to define one of the two.

src/theme.c
```diff
@@ -548,6 +551,12 @@ entry(struct theme *theme, const char *key, const char *value)
 	if (match_glob(key, "osd.window-switcher.item.active.border.width")) {
 		theme->osd_window_switcher_item_active_border_width = atoi(value);
 	}
+	if (match_glob(key, "osd.workspace-switcher.boxes.width")) {
+		theme->osd_workspace_switcher_boxes_width = atoi(value);
+	}
+	if (match_glob(key, "osd.workspace-switcher.boxes.height")) {
+		theme->osd_workspace_switcher_boxes_height = atoi(value);
+	}
```
Consolatis, line 559: Maybe we should additionally check If one of these is `0` and then set the other one to `0` as well.
This could happen in `post_processing()`

Otherwise I think only setting `osd.workspace-switcher.boxes.width` to `0` but keeping `.height` as something different will mess with the height calculations (we disable the boxes but still use their height for calculating the workspace name offset).

Other than that this LGTM.

kyak: Thanks, addressed it in this to-be-squashed commit: https://github.com/labwc/labwc/pull/1299/commits/c6f6ebe96694d9765762062158971b78f11686cf

Consolatis:
:)

PR #1305, opened 2023-12-09
Icon refactoring

johanmalm:
- make button array definition easier to read
- use `max_toggled_hover` with `max_hover_toggled` the alternative
- do not support svg+png icons in max_hover_toggled format

Will wait until this is reviewed/merged and then move onto refactoring `button_xbm_load()` and sort out the fallback regression.
Will try to weave #1296 into it at that point.

Cc @spl237 - for awareness


Consolatis:
:)

PR #1309, opened 2023-12-11
Option to show full app_id in window switcher

### LINKED ISSUE #1073: Different app_id in OSD and windowRule
When using Alt-Tab to display OSD (windowSwitcher), I can see that that Telegram app_id is `desktop`.

However, when trying to match this app_id using `identifier` in `windowRule`, the match doesn't happen.

I used waybar to figure out the Telegram app_id, which is apparently `org.telegram.desktop`.

So this now works (matches):

```
    <windowRule identifier="org.telegram.desktop" serverDecoration="no">
    </windowRule>
```

It seems like OSD and windowRule use different approaches to getting `app_id/identifier`.

Can they be consistent with each other?

What is the proper way to get `app_id` in labwc anyway in order to create window rules?

kyak:
It is now possible to specify whether to show full application identifier or the trimmed variant in window switcher OSD.

Closes #1073

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -134,7 +134,7 @@ this is for compatibility with Openbox.
 
 ## WINDOW SWITCHER
 
-*<windowSwitcher show="" preview="" outlines="">*
+*<windowSwitcher show="" preview="" outlines="" full_app_id="">*
 	*show* [yes|no] Draw the OnScreenDisplay when switching between
 	windows. Default is yes.
 
@@ -143,6 +143,10 @@ this is for compatibility with Openbox.
 
 	*outlines* [yes|no] Draw an outline around the selected window when
 	switching between windows. Default is yes.
+
+	*full_app_id* [yes|no] Display full application identifier instead of trimmed
+	variant. Trimming removes the first two nodes of 'org.' strings. Default is
+	no.
 
 *<windowSwitcher><fields><field content="" width="%">*
 	Define window switcher fields.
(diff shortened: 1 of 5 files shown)


Consolatis:
LGTM, up to @johanmalm if we want to keep the new `full_app_id` option or if we should instead turn this into a new `<field content="full_app_id">` that could be used instead of `identifier`.

Technically, we should actually use this as `identifier` and add a new `content` type like `trimmed_app_id` or something. We could then change the default to use the new `trimmed_app_id` field which would then be backwards compatible to all existing configs that do \*not\* define the fields. It would however break for configs that currently define the `identifier` field.

johanmalm:
> Technically, we should actually use this as `identifier` and add a new `content` type like `trimmed_app_id` or something. We could then change the default to use the new `trimmed_app_id` field which would then be backwards compatible to all existing configs that do *not* define the fields. It would however break for configs that currently define the `identifier` field.

I think we should go for correctness and consistency with the wider code-base, and take the regression hit.

So, “identifier” should mean app-id/wm-class (foo.bar.baz). Suggest we add “trimmed_identifier” and make that default. I’m keen not to use app_id as it’s Wayland specific.

kyak:
@Consolatis @johanmalm I've updated the PR with your suggested approach. I've kept the history, but feel free to squash when merging.

johanmalm:
LGTM

johanmalm:
Thanks!

charbelnicolas:
I changed identifier to trimmed_identifier and saw no difference, whats the catch?

Consolatis:
> I changed identifier to trimmed_identifier and saw no difference, whats the catch?

It only matters for things like `org.mozilla.firefox` which should be `firefox` in the trimmed version and the full string in the non-trimmed version.

PR #1310, opened 2023-12-11
Remove headers not being used directly

kyak:
Fix clangd warnings regarding headers not being used directly.


johanmalm:
Thanks. Nice!

Issue #1282, opened 2023-12-04
Release `0.7.0`


johanmalm:
Porting to `wlroots-0.17` has gone well and I think it's important for distributions that we tag a release as soon as we are reasonably comfortable that there are no regressions.

To that intent, I suggest the following steps for `0.7.0`:

1. Keep a merge window open 4-15 Dec.
2. After that, respect a cool-down period for one week (fixes only)
3. Release on 22 Dec.

Cc
@Consolatis @jlindgren90 @heroin-moose @01micko
@stefonarch @Flrian @kyak @redtide @bi4k8 @bdantas @jp7677
@Tamaranch @kode54 @spl237

stefonarch:
Still open?

johanmalm:
Good point. We're a bit behind on cleansing issues and PRs.

PR #1312, opened 2023-12-11
Add "under-cursor" and "automatic" placement strategies

ahesford:
With under-cursor placement, new top-level windows will be centered under the cursor rather than centered on the active view. This works for xdg-shell and xwayland windows.

Currently, this tyep of placement can be faked with window rules:
```xml
<windowRule identifier="*">
  <action name="MoveToCursor" />
</windowRule>
```
However, this has two drawbacks:
1. Child views of other windows (*e.g.*, dialogs) will open under the cursor instead of centered on their parent (which seems like a better placement for these views); and
2. The action defined by the rule will override preferred placement, which means that things like `xclock -geometry -0-0` will ignore the desired geometry.

Because the new `view_place_initial` just uses the `view_move_to_cursor` action, it takes a couple of actions (un-maximizing and un-tiling the window) unnecessary for initial placement, but I believe those extra actions are harmless. More importantly, reusing `view_move_to_cursor` keeps the logic extremely simple. If this is a problem, `view_move_to_cursor` will have to be split so the out-of-bounds calculations can be reused there and in the new initial placement routine.

There is a reason the placement strategy flag is an enum and not an Boolean: I hope, at some point, to think about adding a variation of the Openbox "smart" placement. Using an enum out of the gate means this won't need to be refactored when (or if) that should happen.

EDIT:
This changeset now includes an "automatic" placement mode that was inspired by the "smart" placement of Openbox. The method works as follows:
1. The usable area of the output on which a window will be placed is subdivided into an irregular grid, with rows and columns defined by every on-screen edge of every existing view. The boundary points that define the rows and columns of the grid are sorted and made unique to facilitate binary searches that identify the position on the grid containing arbitrary points.
2. The list of all windows on the target output are walked to generate an "overlap bitmap". Each window is mapped to a list of spaces on the grid that it overlaps. (The grid is defined such that any given window either complete obscures a particular interval, or does not touch the interval at all.) Counters for each overlapping space are incremented, producing a map from grid spaces to an integer number of windows that completely overlap that grid.
3. To determine the best placement for a new window, the entire grid is walked to compute overlap factors. For each grid interval, the window may potentially extend rightward from the left edge of the interval or leftward from the right edge; likewise, it may extend downward from the top edge or upward from the bottom edge. The overlap of the new window in all four orientations for all grid intervals. The grid interval and growth orientation that produce minimal overlap are used to place the window.
4. As an optimization, the first grid interval and orientation identified that produce *zero* overlap with the new window will be selected without testing later intervals.

The worst-case complexity of this method is O(N^2) in the number of windows, N, on the output being examined. This comes from the use of `qsort` to order the grid and the need to walk the placement grid in both dimensions to compute overlaps. The worst case will occur when existing windows are almost completely overlapped and a new window is large enough to cover most of them. (When they are *completely* overlapped, a lot of edges in the grid will coalesce and reduce the complexity of the walk; when they are close, the grid will be very fine as edges will be close, but not unique.) However, I think that, in practice, arrangements with minimal overlap (which we achieve by design with this placement scheme) will place windows that will not cover substantial numbers of other windows in significantly reduced time---more like O(N log N).

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -134,10 +134,12 @@ this is for compatibility with Openbox.
 
 ## PLACEMENT
 
-*<placement><policy>* [center|cursor]
+*<placement><policy>* [center|automatic|cursor]
 	Specify a placement policy for new windows. The "center" policy will
-	always place windows at the center of the active output. The "cursor"
-	policy will center new windows under the cursor. Default is "center".
+	always place windows at the center of the active output. The "automatic"
+	policy will try to place new windows in such a way that they will
+	have minimal overlap with existing windows. The "cursor" policy will
+	center new windows under the cursor. Default is "center".
 
 ## WINDOW SWITCHER
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -132,6 +132,13 @@ this is for compatibility with Openbox.
 	be used with labwc the preferred mode of the monitor is used instead.
 	Default is no.
 
+## PLACEMENT
+
+*<placement><policy>* [center|cursor]
+	Specify a placement policy for new windows. The "center" policy will
+	always place windows at the center of the active output. The "cursor"
+	policy will center new windows under the cursor. Default is "center".
+
 ## WINDOW SWITCHER
 
 *<windowSwitcher show="" preview="" outlines="">*
--- a/src/placement.c
+++ b/src/placement.c
@@ -1,0 +1,520 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include "common/macros.h"
+#include "common/mem.h"
+#include "labwc.h"
+#include "placement.h"
+#include "ssd.h"
+#include "view.h"
+
+#define overlap_bitmap_index(bmp, i, j) \
+	(bmp)->grid[i * ((bmp)->nr_cols - 1) + j]
+
+struct overlap_bitmap {
+	int nr_rows;
+	int nr_cols;
+	int *rows;
+	int *cols;
+	int *grid;
+};
+
+static int
+compare_ints(const void *a, const void *b)
+{
+	return *(const int *)a - *(const int *)b;
+}
+
+static void
+destroy_bitmap(struct overlap_bitmap *bmp)
+{
+	assert(bmp);
+
+	zfree(bmp->rows);
+	zfree(bmp->cols);
+	zfree(bmp->grid);
+
+	bmp->nr_rows = 0;
+	bmp->nr_cols = 0;
+}
+
+/* Count the number of views on view->output, excluding *view itself */
+static int
+count_views(struct view *view)
+{
+	assert(view);
+
+	struct server *server = view->server;
+	assert(server);
+
+	struct output *output = view->output;
+	if (!output_is_usable(output)) {
+		return 0;
+	}
+
+	int nviews = 0;
+
+	struct view *v;
+	for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+		/* Ignore the target view or anything on a different output */
+		if (v == view || v->output != output) {
+			continue;
+		}
+
+		nviews++;
+	}
+
+	return nviews;
+}
+
+/* Sort and de-deplicate a list of points that define a 1-D grid */
+static int
+order_grid(int *edges, int nedges)
+{
+	/* Sort grid edges */
+	qsort(edges, nedges, sizeof(int), compare_ints);
+
+	/* Skip over non-unique edges, counting the unique ones */
+	/* This is taken almost verbatim from Openbox. */
+	int i = 0;
+	int j = 0;
+
+	while (j < nedges) {
+		int last = edges[j++];
+		edges[i++] = last;
+		while (j < nedges && edges[j] == last) {
+			++j;
+		}
+	}
+
+	return i;
+}
+
+/*
+ * Construct an irregular grid that divides the usable area of view->output
+ * by extending the edges of every view on the output (except for *view itself)
+ * to infinity. The resulting grid will consist of rectangular intervals that
+ * are either completely uncovered by any view, or entirely covered.
+ * Furthermore, when any view intersects any interval on the grid, that view
+ * overlaps the whole interval: no view ever partially intersects any interval.
+ */
+static void
+build_grid(struct overlap_bitmap *bmp, struct view *view)
+{
+	assert(bmp);
+	assert(view);
+
+	struct server *server = view->server;
+	assert(server);
+
+	/* Always start with a fresh bitmap */
+	destroy_bitmap(bmp);
+
+	struct output *output = view->output;
+	if (!output_is_usable(output)) {
+		return;
+	}
+
+	int nviews = count_views(view);
+	if (nviews < 1) {
+		return;
+	}
+
+	/* Number of rows/columns is bounded by two per view plus screen edges */
+	int max_rc = 2 * nviews + 2;
+
+	bmp->rows = xzalloc(max_rc * sizeof(int));
+	bmp->cols = xzalloc(max_rc * sizeof(int));
+	if (!bmp->rows || !bmp->cols) {
+		destroy_bitmap(bmp);
+		return;
+	}
+
+	/* First edges of grid are start of usable area of output */
+	struct wlr_box usable = output_usable_area_in_layout_coords(output);
+	int usable_right = usable.x + usable.width;
+	int usable_bottom = usable.y + usable.height;
+
+	bmp->cols[0] = usable.x;
+	bmp->rows[0] = usable.y;
+
+	bmp->cols[1] = usable_right;
+	bmp->rows[1] = usable_bottom;
+
+	int nr_rows = 2;
+	int nr_cols = 2;
+
+	struct view *v;
+	for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+		if (v == view || v->output != output) {
+			continue;
+		}
+
+		struct border margin = ssd_get_margin(v->ssd);
+		int x = v->pending.x - margin.left;
+		int y = v->pending.y - margin.top;
+
+		/* Add a column if the left view edge is in the usable region */
+		if (x > usable.x && x < usable_right) {
+			assert(nr_cols < max_rc);
+			bmp->cols[nr_cols++] = x;
+		}
+
+		/* Add a row if the top view edge is in the usable region */
+		if (y > usable.y && y < usable_bottom) {
+			assert(nr_rows < max_rc);
+			bmp->rows[nr_rows++] = y;
+		}
+
+		x = v->pending.x + v->pending.width + margin.right;
+		y = v->pending.y + v->pending.height + margin.bottom;
+
+		/* Add a column if the right view edge is in the usable region */
+		if (x > usable.x && x < usable_right) {
+			assert(nr_cols < max_rc);
+			bmp->cols[nr_cols++] = x;
+		}
+
+		/* Add a row if the bottom view edge is in the usable region */
+		if (y > usable.y && y < usable_bottom) {
+			assert(nr_rows < max_rc);
+			bmp->rows[nr_rows++] = y;
+		}
+	}
+
+	bmp->nr_rows = order_grid(bmp->rows, nr_rows);
+	bmp->nr_cols = order_grid(bmp->cols, nr_cols);
+
+	int grid_size = (bmp->nr_rows - 1) * (bmp->nr_cols - 1);
+
+	bmp->grid = xzalloc(grid_size * sizeof(int));
+	if (!bmp->grid) {
+		destroy_bitmap(bmp);
+		return;
+	}
+}
+
+/*
+ * Perform a rightmost binary search along a list of edges in a 1-D grid for
+ * the maximum index j such that edges[j] <= val. The list of edges must be
+ * sorted in increasing order.
+ *
+ * For a returned index j:
+ *
+ * - The index j == -1 implies that val < edges[0].
+ * - An index 0 <= j < (nedges - 1) implies that edges[j] <= val < edges[j + 1].
+ * - The index j == (nedges - 1) implies that edges[nedges - 1] <= val.
+ */
+static int
+find_interval(int *edges, int nedges, double val)
+{
+	int l = 0;
+	int r = nedges;
+
+	while (l < r) {
+		int m = (l + r) / 2;
+		if (edges[m] > val) {
+			r = m;
+		} else {
+			l = m + 1;
+		}
+	}
+
+	return r - 1;
+}
+
+/*
+ * Construct an overlap bitmap for the irregular grid, computed by
+ * build_grid(), that spans view->output. The overlap bitmap maps
+ * each interval to the number of views on the output (excluding *view)
+ * that overlap that interval.
+ */
+static void
+build_overlap(struct overlap_bitmap *bmp, struct view *view)
+{
+	assert(bmp);
+	assert(view);
+
+	struct server *server = view->server;
+	assert(server);
+
+	if (bmp->nr_rows < 1 || bmp->nr_cols < 1) {
+		return;
+	}
+
+	struct output *output = view->output;
+	if (!output_is_usable(output)) {
+		return;
+	}
+
+	struct view *v;
+	for_each_view(v, &server->views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+		if (v == view || v->output != output) {
+			continue;
+		}
+
+		/* Find boundaries of the window */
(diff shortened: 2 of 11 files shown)


Consolatis:
First of all: thanks for working on this.

> 1. Child views of other windows (_e.g._, dialogs) will open under the cursor instead of centered on their parent (which seems like a better placement for these views); and

That can be worked around with the `matchOnce` attribute: https://labwc.github.io/labwc-config.5.html#entry_matchonce

According to the Openbox wiki the config format is slightly different: http://openbox.org/wiki/Help:Configuration#Placement

We [do list](https://github.com/labwc/labwc-scope?tab=readme-ov-file#23-configuration---placement) the `<placement>` config as category `C`, thus not intended to be implemented. Due to the minor code changes required to implement this feature (and reusing existing features within the codebase) I would be open to implement it though. But this requires a decision from the project lead @johanmalm. Also keep in mind that its not just the code changes itself but we will also have to maintain this going forward.

CC @jlindgren90

jlindgren90:
I like the idea of implementing smarter window placement (especially since xdg-shell views cannot do it themselves). So I also appreciate the effort on this.

We'd want to check that XWayland views can still request their own window placement. Maybe that's already been considered (I haven't looked at the code changes).

I do wonder if it makes sense at some point or for some period of time to focus on stabilization, so that we don't get ahead of ourselves with adding new features. Maybe this could wait until after 0.7.0?

ahesford:
Thanks for pointing out the scope document.

The configuration differences from Openbox were intentional. Openbox treats the `<center>` option as a *modifier* on the placement policy; because I didn't intend to make centering over the cursor optional, having a dedicated section for a single enum seemed superfluous. Making it an attribute in `core` allows for more terse configuration. Also, because I deviated here, I figured there wasn't a lot of value in preserving the (subjectively ugly) `UnderMouse` label. However, if the feature is desirable and the consensus is that its configuration should parallel that of Openbox, we can move the option.

I have tested some XWayland views (specifically `xclock` and `urxvt`) and can confirm that, when they are invoked with a `-geometry` argument, the new placement strategy respects any position explicitly specified. (In fact, that was the primary motivation for implementing this strategy instead of just relying on window rules---I like to keep an `xclock` in the bottom-right corner with `-geometry -0-0`, but adding a global `MoveToCursor` action would cause the geometry to be overridden.)

If this is merged, I'd also appreciate some advance comment on the possibility of accepting a `smart` policy that would work similar to (but probably a bit different than) the Openbox `Smart` policy. I've been thinking through the problem, but have not yet committed to implement an approach.

docs/labwc-config.5.scd
```diff
@@ -132,6 +133,11 @@ this is for compatibility with Openbox.
 	be used with labwc the preferred mode of the monitor is used instead.
 	Default is no.
 
+*<core><placement>* [center|cursor]
+	Specify placement strategy for new windows. With "cursor" placement,
+	windows will be centered under the cursor. WIth "center" placement,
```
Consolatis, line 138: Minor spelling: `WIth` (uppercase `i`)

johanmalm:
Thanks for working on this.
Happy to upgrade to a cat B.
I'd like to think through config naming. IIRC openbox added `center` quite late, and looking at it now it feels somewhat tied to `smart` (might be wrong on that though).
We're just about entering cool down for 0.7.0, so could we postpone this for a couple of weeks?
https://github.com/labwc/labwc/issues/1282

ahesford:
Waiting until after the release of 0.7.0 is fine with me. Void Linux is ready with a new `wlroots0.17` package.

I'm not sure if `center` in Openbox has any effect with `UnderMouse` placement, although it could conceivably toggle between centering the window on the cursor or placing the top-left corner at the cursor. I don't think that alternative is particularly interesting and isn't worth added complexity. How centering interplays with `smart` placement and whether it is exposed as a configuration would depend on the nature of the implementation for labwc.

ahesford:
For future reference, I've started working through "smart" placement (which I call "automatic" because I hesitate to call any software "smart") that was inspired smart placement in Openbox. Changes are available for preview on [a branch that extends from this PR](https://github.com/ahesford/labwc/commit/hitchcock).

The basic functionality is in place, but I would probably prefer to fine-tune the placement so it will consider placing windows along the bottom and right edges. (Right now, it will never automatically place windows on these edges.)

If I finish my effort before this PR is merged, I may extend this request to include the automatic placement. Otherwise, I will submit a follow-on PR later.

johanmalm:
Oh, I like the look of that! If you're okay to hold on for a few more days, we'll get `0.7.0` out and then have a look at it more properly.

ahesford:
Alright, I think I worked out the kinks, so I might as well submit "automatic" placement for consideration as part of this PR. Waiting until after 0.7.0 for thorough review is appropriate.

I've edited the primary description above to reflect the new additions.

jlindgren90 (commented):
Thanks very much for working on this, I haven't tested it yet but look forward to doing so.

placement.c looks like some genius-level code :) Is it mostly from Openbox, or original (just curious)? It would take me longer than I have tonight to read and understand all of it, so I just skimmed it, but overall it looks well done and nicely organized.

ahesford:
Thanks. I added the assertions and comments, along with several other explanatory comments.

The basic grid construction, which I'm doing in `placement_build_grid` and `placement_order_grid`, is pretty much taken from the [Openbox implementation](https://github.com/danakj/openbox/blob/master/openbox/place_overlap.c) of `make_grid` and `uniquify`, except I take more care to filter out "uninteresting" (out-of-bounds) edges. However, when Openbox scans the grid for optimal placement, it walks the entire list of views ("clients") to check for intersections for every test placement. The overlap bitmap implemented here is original and designed to avoid that repeated walk.

src/placement.c
```diff
@@ -0,0 +1,513 @@
...
+#define overlap_bitmap_index(bmp, i, j) \
+	(bmp)->grid[i * ((bmp)->ncols - 1) + j]
+
+static int
+compare_ints(const void *a, const void *b)
+{
+	return *(const int *)a - *(const int *)b;
+}
+
+static void placement_destroy_bitmap(struct overlap_bitmap *bmp)
```
johanmalm, line 21: Start function name on new line

src/placement.c
```diff
@@ -0,0 +1,513 @@
...
+	bmp->nrows = 0;
+	bmp->ncols = 0;
+	bmp->rows = NULL;
+	bmp->cols = NULL;
+	bmp->grid = NULL;
+}
+
+/* Count the number of views on view->output, excluding *view itself */
+static int
+placement_count_views(struct view *view)
```
johanmalm, line 38: We could drop the `placement_` prefix for static functions (within the translation unit) if you want. I'm easy.

src/placement.c
```diff
@@ -0,0 +1,513 @@
...
+/*
+ * Build an overlap bitmap for view->output, which subdivides view->output into
+ * a grid and then counts the number of windows that overlap each grid region.
+ */
+static struct overlap_bitmap
+placement_build_bitmap(struct view *view)
+{
+	assert(view);
+
+	struct overlap_bitmap bitmap = {
```
johanmalm, line 317: Could just do `struct overlap_bitmap bitmap = {0}`. Again, just a suggestion.

src/placement.c
```diff
@@ -0,0 +1,513 @@
...
+
+	int nri = bmp.nrows - 1;
+	int nci = bmp.ncols - 1;
+
+	/* Walk the grid to test all possible placements of the view */
+	for (int i = 0; i < nri; ++i) {
+		for (int j = 0; j < nci; ++j) {
+			/* Search southeast, northeast, northwest, southwest */
+			for (int ii = 0; ii < 4; ++ii) {
+				bool rt = (ii & 0x1) == 0;
```
johanmalm, line 465: It's not obvious to me how this works. A comment or long variable names or something like that would be helpful.

johanmalm (commented):
Great piece of work!

Have done a few quick review comments, but have not wrapped my head around the logic in `placement.c`. Have tested (one output only) and it seems to work very well :smile:

The code looks to be of high quality - so I'm sure it's fine, but think would be good if @jlindgren90 @Consolatis or myself did a high-level follow through of the logic.

ahesford:
I've adopted all suggestions so far, taking care to more thoroughly describe the overlap computations above the convolution loop. I've also added the new `<placement>` section to `rc.xml.all`, setting the policy to `center` to reflect the default.

I tend to do all of my testing on a dual-output system, and will also stand up nested sessions for testing with, *e.g.*, `WLR_WL_OUTPUTS=2 ./labwc`. (The `ToggleKeybinds` action is *really* useful when nesting one labwc session inside another!)

ahesford:
I realized that some of the changes, like the structure names or moving the configuration to `<placement><policy>`, would also apply to the original under-mouse commit. The latest rebase makes the first commit stand alone with under-mouse placement and the preferred configuration, while the second adds the automatic placement consistent with the preceding change.

Alternatively, the changeset can be squashed into a single commit that adds both placement policies.

src/placement.c
```diff
@@ -0,0 +1,548 @@
...
+
+	int nviews = count_views(view);
+	if (nviews < 1) {
+		return;
+	}
+
+	/* Number of rows/columns is bounded by two per view plus screen edges */
+	int max_rc = 2 * nviews + 2;
+
+	int rows[max_rc];
```
johanmalm, line 130: If you agree with the comment below about doing separate allocations for row/col/grid, I think it would seem cleaner to use `wl_array` here (which is just a simple array implementation). It would save us having to first create `rows[]` and `cols[]` here only to then heap allocate the same.

I haven't tried it, but think `qsort()` works on a `wl_array` array.

src/placement.c
```diff
@@ -0,0 +1,548 @@
...
+ * sorted in increasing order.
+ *
+ * For a returned index j:
+ *
+ * - The index j == -1 implies that val < edges[0].
+ * - An index 0 <= j < (nedges - 1) implies that edges[j] <= val < edges[j + 1].
+ * - The index j == (nedges - 1) implies that edges[nedges - 1] <= val.
+ */
+static int
+find_interval(int *edges, int nedges, double val)
```
johanmalm, line 218: Could we use `bsearch()` for this?
I don't mind keeping it as it is though.

johanmalm (commented):
Have got as far as the start of `build_overlap()` with the review.
Will carry on tomorrow.

docs/labwc-config.5.scd
```diff
@@ -110,6 +110,7 @@ this is for compatibility with Openbox.
   <gap>0</gap>
   <adaptiveSync>no</adaptiveSync>
   <reuseOutputMode>no</reuseOutputMode>
+  <placement>center</placement>
```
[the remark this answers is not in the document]
ahesford: Fixed

src/placement.c
```diff
@@ -0,0 +1,548 @@
...
+		if (y > usable.y && y < usable_bottom) {
+			assert(nr_rows < max_rc);
+			rows[nr_rows++] = y;
+		}
+	}
+
+	nr_rows = order_grid(rows, nr_rows);
+	nr_cols = order_grid(cols, nr_cols);
+
+	int nr_grid = (nr_rows - 1) * (nr_cols - 1);
```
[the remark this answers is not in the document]
ahesford: Fixed

src/placement.c
ahesford: I moved from `malloc` to `xzalloc` as recommended and avoided the stack allocation by just directly allocating (once for each of the arrays, rather than folding them all into a single allocation) enough storage on the heap to store the maximum grid size right in the bitmap struct. I don't think `wl_array` buys anything here except extra complexity.

For one, it will tend to over-allocate on the heap even worse than just asking for the maximum possible grid size up front; `wl_array_add` looks for power-of-two sizes to satisfy the requested addition. For another, we will still be directly manipuatling the underlying data array everywhere, so this would just replace a single call to `xzalloc` with a structure delcaration, `wl_array_init` call and `wl_array_add` call; and all subsequent data access currently of the form `bmp->member` with `bmp->member->data`. This also ties the implementation of the placement algorithm, where the array manipulations are entirely contained, to the that of `wl_array` and might require a refactor of the placement algorithms should the implementation of `wl_array` ever change.

If you really want to see these moved to `wl_array`, I will do so; but I'd rather just leave them as low-level C arrays.

src/placement.c
ahesford: I looked into `bsearch`, but needed a custom implementation because `bsearch` returns `NULL` when there is no exactly matching element. The implementation here relies on inexact interval matches to simplify identification of the start and end of the range of intervals that bound a view.

src/placement.c
```diff
@@ -0,0 +1,552 @@
...
+{
+	return *(const int *)a - *(const int *)b;
+}
+
+static void
+destroy_bitmap(struct overlap_bitmap *bmp)
+{
+	assert(bmp);
+
+	if (bmp->rows) {
```
johanmalm, line 35: We could save a few lines here by just using `zfree()` (labwc helper). For example (`zfree(bmp->rows)`). It does a `free()` followed by setting the variable to zero.

src/placement.c
johanmalm: Good point!

src/placement.c
```diff
@@ -0,0 +1,552 @@
...
+
+		/*
+		 * Increment the last indices to convert them to strict upper
+		 * bounds, then clip them to the limits of the usable grid.
+		 */
+		lc = MIN(bmp->nr_cols - 1, lc + 1);
+		lr = MIN(bmp->nr_rows - 1, lr + 1);
+
+		/*
+		 * Every interval in the region [fr, lr) x [fc, lc) is
```
johanmalm, line 309: Mix of `[` and `)`

src/placement.c
```diff
@@ -0,0 +1,552 @@
...
+/*
+ * Build an overlap bitmap for view->output, which subdivides view->output into
+ * a grid and then counts the number of windows that overlap each grid region.
+ */
+static struct overlap_bitmap
+build_bitmap(struct view *view)
+{
+	assert(view);
+
+	struct overlap_bitmap bitmap = { 0 };
```
johanmalm, line 330: For this type of stuff, it's generally better to declare at the highest stack frame that the variable is used in (and easier for me to read).

Suggest we could get rid of this whole function and just add the following to where `build_bitmap()` is called from:

```
struct overlap_bitmap bitmap = { 0 };
build_grid(&bitmap, view);
build_overlap(&bitmap, view);
```

The `if (bitmap.nr_rows < 1...` check is done in `build_overlap()` anyway.

src/placement.c
```diff
@@ -0,0 +1,552 @@
...
+			int cw = bmp->cols[jj + 1] - bmp->cols[jj];
+
+			/* Width of overlap between this column and test region */
+			int mw = MAX(0, MIN(ww, cw));
+
+			/* Add overlap contribution for this interval */
+			overlap += overlap_bitmap_index(bmp, ii, jj) * mh * mw;
+
+			/* Count the number of overlapping intervals */
+			count++;
```
johanmalm, line 400: It doesn't look like we're using `count`.... unless I'm getting my wires crossed.

johanmalm (commented):
Have got my head around it now.

Have left a few (minor) comments if you'd be happy to consider those.

Then happy to merge

johanmalm:
> I realized that some of the changes, like the structure names or moving the configuration to `<placement><policy>`, would also apply to the original under-mouse commit. The latest rebase makes the first commit stand alone with under-mouse placement and the preferred configuration, while the second adds the automatic placement consistent with the preceding change.
> 
> Alternatively, the changeset can be squashed into a single commit that adds both placement policies.

Good to keep as two separate commits I think.

src/placement.c
johanmalm: Good points. Normal C arrays are fine.

src/placement.c
ahesford: This is intentional; these intervals are half-open, including `fr` and `fc` but excluding `lr` and `lc` (after they've been incremented int he line above).

src/placement.c
ahesford: The `count` variable is used to set the value in `single` before this function returns, which is used to short-circuit the multi-direction check if the view to be placed is not larger than the single grid interval where it starts.

ahesford:
I've incorporated the latest round of suggestions.

Thanks for digging through this. I've kept an eye on Wayland for a few years now, and labwc is the first project that allows me to replicate my longstanding Openbox workflow while bringing some modern conveniences.

src/placement.c
johanmalm: Sorry. Have caught up now. My brain wasn't tuned to interval notation.

src/placement.c
johanmalm: Got it. Not sure why I didn't see that.

johanmalm:
Thanks!

> Thanks for digging through this. I've kept an eye on Wayland for a few years now, and labwc is the first project that allows me to replicate my longstanding Openbox workflow while bringing some modern conveniences.

Great to have you onboard.

We've got (in my opinion) quite a nice setup with a few maintainers/devs who care about getting the details right and working on something that will hopefully stand the test of time.

I enjoyed reviewing your patch - certainly one of the more interesting ones of 2023 :smile:

PR #1313, opened 2023-12-11
docs/labwc-theme.5.scd: made text fit in max 80-column wide lines

domo141:
For general consistency and better readability as source.


johanmalm:
Thanks!

Issue #1303, opened 2023-12-09
Assertion failure when undocking with screens powered off (wlopm)


jlindgren90:
labwc aborted with this backtrace when I undocked my laptop while the screens were off (using wlopm):

```
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at pthread_kill.c:44
Downloading source file /usr/src/debug/glibc/glibc/nptl/pthread_kill.c
44            return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;                                 
[Current thread is 1 (Thread 0x7f498275ed80 (LWP 14216))]
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at pthread_kill.c:44
#1  0x00007f49858ac8a3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007f498585c668 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f49858444b8 in __GI_abort () at abort.c:79
#4  0x00007f49858443dc in __assert_fail_base
    (fmt=0x7f49859bdae8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x561c1b2493a3 "solo->scene_output != so", file=file@entry=0x561c1b24936e "types/scene/output_layout.c", line=line@entry=78, function=function@entry=0x561c1b26e740 <__PRETTY_FUNCTION__.0.lto_priv.19> "wlr_scene_output_layout_add_output") at assert.c:92
#5  0x00007f4985854d26 in __assert_fail
    (assertion=assertion@entry=0x561c1b2493a3 "solo->scene_output != so", file=file@entry=0x561c1b24936e "types/scene/output_layout.c", line=line@entry=78, function=function@entry=0x561c1b26e740 <__PRETTY_FUNCTION__.0.lto_priv.19> "wlr_scene_output_layout_add_output") at assert.c:101
#6  0x0000561c1b1ad4e1 in wlr_scene_output_layout_add_output
    (so=<optimized out>, lo=0x561c1c1d9810, sol=0x561c1bee1050)
[...]

PR #1325, opened 2023-12-13
output: do not call wlr_scene_output_layout_add_output() twice

jlindgren90:
Fixes an assertion failure that occurred if reconfiguring outputs after an output had been disabled using `wlopm --off`.

Fixes: #1303

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -158,10 +158,14 @@ add_output_to_layout(struct server *server, struct output *output)
 			wlr_log(WLR_ERROR, "unable to create scene output");
 			return;
 		}
+		/*
+		 * Note: wlr_scene_output_layout_add_output() is not
+		 * safe to call twice, so we call it only when initially
+		 * creating the scene_output.
+		 */
+		wlr_scene_output_layout_add_output(server->scene_layout,
+			layout_output, output->scene_output);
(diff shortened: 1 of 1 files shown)


Consolatis:
Hm.. shouldn't `need_to_remove` be true which removes the output from the layout? So when its enabled it shouldn't be in the layout anymore.

jlindgren90:
I think the output power management stuff takes a different code path that doesn't call output_config_apply.

Consolatis:
- I think we should refactor the scene output stuff a bit: create it on `new_output_notify()` and destroy it on `output_destroy_notify()` (The comment referring to wlroots destroying the scene output on `wlr_output_layout_remove()` is outdated, it has been [fixed](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e1e911d425fde6ce85475de95b44aa1340725224))
- We could use `if (!wlr_output_layout_get(output->wlr_output))` to test if the output is already in the layout or not.

jlindgren90:
I agree there is some cleanup that could be done. I don't have much more time to work on it right now though.

Consolatis:
As this fixes a crash I am going for the merge. Thanks.

PR #1326, opened 2023-12-13
xwayland: fix issue with panel getting stuck offscreen

jlindgren90:
For unknown reasons, XWayland surfaces that are completely offscreen seem not to generate commit events. In rare cases, this can prevent an offscreen window from moving onscreen (since we wait for a commit event that never occurs). As a workaround, move offscreen surfaces immediately.

This fixes an issue that I can reproduce by having qmpanel displayed on a (larger) external monitor, then undocking the laptop so that qmpanel requests a simultaneous move+resize to the (smaller) laptop display.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -311,9 +311,20 @@ xwayland_view_configure(struct view *view, struct wlr_box geo)
 	wlr_xwayland_surface_configure(xwayland_surface_from_view(view),
 		geo.x, geo.y, geo.width, geo.height);
 
+	/*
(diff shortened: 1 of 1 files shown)


Consolatis:
Apologies for not looking at this earlier.

Do you want this to be included in 0.7.0?

jlindgren90:
Yes, I think it would be good to include. Thanks!

Consolatis:
As its self contained to xwayland and only triggers when the view is not in the layout \*and\* got a move / resize request this seems fine to me. LGTM.

PR #1329, opened 2023-12-15
Add `touchpad` device type

jmbaur:
It is nice to have finer granularity for device types to allow for configurations such as using `naturalScroll` on touchpads, but not on regular pointer devices such as mice.

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -33,17 +33,28 @@ input_device_destroy(struct wl_listener *listener, void *data)
 	free(input);
 }
 
-static bool
-is_touch_device(struct wlr_input_device *wlr_input_device)
+static enum device_type
+device_type_from_wlr_device(struct wlr_input_device *wlr_input_device)
 {
 	switch (wlr_input_device->type) {
 	case WLR_INPUT_DEVICE_TOUCH:
 	case WLR_INPUT_DEVICE_TABLET_TOOL:
-		return true;
+		return TOUCH_DEVICE;
 	default:
 		break;
 	}
-	return false;
+
+	if (wlr_input_device->type == WLR_INPUT_DEVICE_POINTER &&
+			wlr_input_device_is_libinput(wlr_input_device)) {
+		struct libinput_device *libinput_device =
+			wlr_libinput_get_device_handle(wlr_input_device);
+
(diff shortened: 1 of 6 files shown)


src/seat.c
```diff
@@ -32,17 +32,28 @@ input_device_destroy(struct wl_listener *listener, void *data)
 	free(input);
 }
 
-static bool
-is_touch_device(struct wlr_input_device *wlr_input_device)
+static enum device_type
+wlr_device_to_device_type(struct wlr_input_device *wlr_input_device)
```
johanmalm, line 36: We shouldn't use the `wlr_` prefix as that's 'reserved' for wlroots.
Suggest `device_type_from_wlr_device()` or something like that instead.

johanmalm:
Thanks. This looks good. 

We're in 'cool down' for `0.7.0` and are focusing on fixing a couple of regressions before the release, so will postpone the merge for a weekish (or however long it takes to get the release out!). Hope that's okay.

src/seat.c
jmbaur: Renamed to `device_type_from_wlr_device`

johanmalm:
To start with I was a bit confused by this as `touchpad` is not one of the devices here: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_input_device.h?ref_type=heads

Looking at sway's code, I now realize that `touchpad` is a `pointer`-variant. 
https://github.com/swaywm/sway/blob/0aceff74696f03d306dc9de5473d9ed7afb73f67/sway/input/input-manager.c#L87-L123

Would you mind updating `docs/labwc-config.5.scd` too?
https://github.com/labwc/labwc/blob/7c593517749cf58b91f74d453cd13f06c433a4e7/docs/labwc-config.5.scd?plain=1#L435-L442

Other than that, LGTM and happy to merge.

If you feel like it, we should probably improve the libinput category documentation to make it a bit clearer what touch/non-touch/etc are in general, but that's probably a separate exercise - and one that I/other can do.

jmbaur:
Thanks @johanmalm, added `touchpad` to the manpage. I agree some clearing up would be helpful in the documentation, as I was initially confused as to the difference in the input categories as compared to sway.

johanmalm:
Thanks!

PR #1331, opened 2023-12-16
fix: fall-through to default cursor in cursor_get_from_edge()

ahesford:
When assertions are disabled, providing an unexpected input to cursor_get_from_edge will cause the non-void function to terminate without a return value. This may be effectively unreachable in practice. However, returning a default cursor as a fall-through case will both silence a compiler warning and prevent catastrophy should the function ever be called with a permitted value.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -91,6 +91,8 @@ cursor_get_from_edge(uint32_t resize_edges)
 			"Failed to resolve wlroots edge %u to cursor name", resize_edges);
 		assert(false);
 	}
+
+	return LAB_CURSOR_DEFAULT;
 }
 
 static enum lab_cursors


Consolatis:
Thanks! I guess an alternative would be to use `abort()` here but I am also fine with this solution. In dev builds the assert would still trigger which is the important part here. I won't merge this directly to give @johanmalm a chance to voice his opinion as well.

johanmalm:
Nice catch! Thanks.

Re optimised release build in CI, I agree. Good idea. @ahesford What did you compile with to get the warning?

ahesford:
I usually build this using `xbps-src` on Void, which [sets the following meson options](https://github.com/void-linux/void-packages/blob/master/common/build-style/meson.sh#L23-L42) by default. In this case, `${configure_args}` is empty.

PR #1334, opened 2023-12-16
theme: use non-hover button variants as fallback

jlindgren90:
Some themes don't have hover variants for button pixmaps. It looks better visually to use the non-hover variants as fallbacks rather than the built-in 6x6 pixmaps.

### DIFF
--- a/include/button/button-xbm.h
+++ b/include/button/button-xbm.h
@@ -6,6 +6,7 @@ struct lab_data_buffer;
 
 /* button_xbm_load - Convert xbm file to buffer with cairo surface */
 void button_xbm_load(const char *button_name, const char *alt_name,
-	struct lab_data_buffer **buffer, char *fallback_button, float *rgba);
+	struct lab_data_buffer **buffer, const char *fallback_button,
+	float *rgba);
 
 #endif /* LABWC_BUTTON_XBM_H */
--- a/include/theme.h
+++ b/include/theme.h
@@ -92,6 +92,7 @@ struct theme {
 	struct lab_data_buffer *button_iconify_inactive_unpressed;
 	struct lab_data_buffer *button_menu_inactive_unpressed;
 
+	/* hover variants are optional and may be NULL */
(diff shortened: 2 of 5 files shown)


johanmalm:
In terms of button-regression there are two thing to fix before `0.7.0`: 

a. Fallback for missing _hover icons (which you've just done with this patch)
b. Hover overlay for missing _hover icons

I do like your patch in isolation, particularly avoiding the `0x40` hack in PR #1296. BUT, if we go with it, I think (b) might be harder to achieve (although I might of course be wrong on that one!)

My thoughts are that it would be good to create dedicated pixbufs for missing _hover icons which are copies of the non-hover variants and (optionally?) contains the overlay (rather than just pointing to the non-variant pixbufs). I think the advantages of this approach are:

1. It enables simplification of `src/ssd/` because we don't have to add the overlay there.
2. We could have more sophisticated overlays with rounded corners or whatever to avoid them going outside the titlebar when radius is big.
3. I think it would simplify supporting _pressed variants in the future (just adding some pixbufs rather than more logic in `src/ssd/`

If you agree on this approach, I suspect that it would be much easier to just use CARIO_FORMAT_ARGB32 buffers rather than the DRM_FORMAT_ARGB8888 ones. I'm not sure how to do that, but am sure we could work it out.

As an aside, I think we ought to break out `button_xbm_from_bitmap()` from `buttom_xbm_load()` to keep the API easier and move some fallback logic to `theme.c`. That can obviously be done as a separate exercise and is outside of
the regression fixes.

johanmalm:
@Consolatis and I discussed last night.

I think we have a plan for the hover overlay that means we’ll merge this PR and then sort out the hover stuff in a follow-up commit.

Just wanted to note it here to avoid duplicating effort.

Basic idea: https://github.com/Consolatis/labwc/commit/91c7d5d7afb5a288bee373535de716ab68502652

Consolatis:
Parts of this PR have been merged in
- #1339

Thanks

PR #1335, opened 2023-12-16
CONTRIBUTING.md: update deprecated meson command

Flrian:
Fixes "WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.", plus use `meson compile -C build/` like in README.md instead of `ninja -C build/`


johanmalm:
Thanks.

PR #1339, opened 2023-12-18
Icon refactoring4

johanmalm:
- Remove builtin bitmap fallbacks for `_hover` icons
- For `_hover` buttons with an xbm/svg/png, just copy the non-hover variant and add an overlay
- Simplify `button_xbm_load()` by moving some of the fallback logic to `theme.c`

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -219,21 +219,20 @@ load_buttons(struct theme *theme)
 	}, };
 
 	char filename[4096] = {0};
-	char alt_filename[4096] = {0};
 	for (size_t i = 0; i < ARRAY_SIZE(buttons); ++i) {
 		struct button *b = &buttons[i];
 
 		drop(b->active.buffer);
 		drop(b->inactive.buffer);
 
-		/* Try png icon first */
+		/* PNG */
 		snprintf(filename, sizeof(filename), "%s-active.png", b->name);
 		button_png_load(filename, b->active.buffer);
 		snprintf(filename, sizeof(filename), "%s-inactive.png", b->name);
 		button_png_load(filename, b->inactive.buffer);
 
 #if HAVE_RSVG
-		/* Then try svg icon */
+		/* SVG */
 		int size = theme->title_height - 2 * theme->padding_height;
 		if (!*b->active.buffer) {
 			snprintf(filename, sizeof(filename), "%s-active.svg", b->name);
@@ -245,19 +244,47 @@ load_buttons(struct theme *theme)
 		}
 #endif
 
-		/* If there were no png/svg buttons, use xbm */
+		/* XBM */
 		snprintf(filename, sizeof(filename), "%s.xbm", b->name);
-		alt_filename[0] = '\0';
-		if (b->alt_name) {
-			snprintf(alt_filename, sizeof(alt_filename), "%s.xbm", b->alt_name);
-		}
 		if (!*b->active.buffer) {
-			button_xbm_load(filename, alt_filename, b->active.buffer,
-				b->fallback_button, b->active.rgba);
+			button_xbm_load(filename, b->active.buffer, b->active.rgba);
 		}
 		if (!*b->inactive.buffer) {
-			button_xbm_load(filename, alt_filename, b->inactive.buffer,
-				b->fallback_button, b->inactive.rgba);
+			button_xbm_load(filename, b->inactive.buffer, b->inactive.rgba);
+		}
+
+		/*
+		 * XBM (alternative name)
+		 * For example max_hover_toggled instead of max_toggled_hover
+		 */
+		if (b->alt_name) {
+			snprintf(filename, sizeof(filename), "%s.xbm", b->name);
+		}  else {
+			filename[0] = '\0';
+		}
+		if (!*b->active.buffer) {
+			button_xbm_load(filename, b->active.buffer, b->active.rgba);
+		}
+		if (!*b->inactive.buffer) {
+			button_xbm_load(filename, b->inactive.buffer, b->inactive.rgba);
+		}
+
+		/*
+		 * Builtin bitmap
+		 *
+		 * Applicable to basic buttons such as max, max_toggled and
+		 * iconify. There are no bitmap fallbacks for *_hover icons.
+		 */
+		if (!b->fallback_button) {
+			continue;
+		}
+		if (!*b->active.buffer) {
+			button_xbm_from_bitmap(b->fallback_button,
+				b->active.buffer, b->active.rgba);
+		}
+		if (!*b->inactive.buffer) {
+			button_xbm_from_bitmap(b->fallback_button,
+				b->inactive.buffer, b->inactive.rgba);
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


johanmalm:
Ready for review. I think this should be enough button hacking for `0.7.0`

src/theme.c
```diff
@@ -56,6 +57,62 @@ drop(struct lab_data_buffer **buffer)
...
+}
+
+static void
+create_hover_fallback(struct theme *theme, struct lab_data_buffer **hover_buffer,
+		struct lab_data_buffer *icon_buffer)
+{
+	assert(icon_buffer);
+	assert(!*hover_buffer);
+
+	cairo_surface_t *icon = get_surface(icon_buffer);
```
Consolatis, line 90: This may leak the cairo surface on `Reconfigure`. We need some way to `cairo_surface_destroy(icon)` but only when it was actually created in `get_surface()` (the `DRM_FORMAT_ARGB8888`) case.
Maybe something like
```c
struct surface_wrap {
	bool free_surface;
	cairo_surface_t *surface;
};

static struct surface_wrap
get_surface()
{
	..
}

static void
create_hover_fallback(..)
{
	..
	if (icon->free_surface) {
		cairo_surface_destroy(icon->surface);
	}
}
```
?

src/theme.c
```diff
@@ -189,20 +245,74 @@ load_buttons(struct theme *theme)
...
+	for (size_t i = 0; i < ARRAY_SIZE(buttons); i++) {
+		struct button *button = &buttons[i];
+		if (*button->active.buffer) {
+			continue;
+		}
+
+		/* Given foo_hover, copy foo */
+		char *buf = strdup(button->name);
+		size_t len = strlen(buf);
+		buf[len-6] = '\0';
```
Consolatis, line 302: Although this should usually be fine as currently only the hover overlays may be empty this seems like a bug waiting to happen in the future (e.g. imagine it being used on `max`). We should improve this so it only reacts on things ending in `_hover`.

src/theme.c
```diff
@@ -189,20 +245,74 @@ load_buttons(struct theme *theme)
...
+				b->inactive.buffer, b->inactive.rgba);
+		}
+	}
+
+	/* Add fallbacks for missing hover overlays */
+	for (size_t i = 0; i < ARRAY_SIZE(buttons); i++) {
+		struct button *button = &buttons[i];
+		if (*button->active.buffer) {
+			continue;
+		}
```
Consolatis, line 297: This relies on the fact that for xbm buttons its either both set (`->active` and `->inactive`) or none. However, that doesn't seem to hold true for the png and svg variants. So we either need to duplicate some code here to check each one individually or use some `FOR_EACH()` macro or temporary array for the two states.

src/theme.c
```diff
@@ -189,20 +245,74 @@ load_buttons(struct theme *theme)
...
+		buf[len-6] = '\0';
+		for (size_t j = 0; j < ARRAY_SIZE(buttons); j++) {
+			struct button *tmp = &buttons[j];
+			if (!strcmp(buf, tmp->name)) {
+				create_hover_fallback(theme,
+					button->active.buffer,
+					*tmp->active.buffer);
+				create_hover_fallback(theme,
+					button->inactive.buffer,
+					*tmp->inactive.buffer);
```
Consolatis, line 311: Same as above, just because the `->active` buffer isn't set it doesn't necessarily mean that the `->inactive` buffer isn't set either for loaders other than the xbm one.

src/theme.c
```diff
@@ -56,6 +57,62 @@ drop(struct lab_data_buffer **buffer)
...
+	*hover_buffer = buffer_create_cairo(width, height, 1, true);
+
+	cairo_t *cairo = (*hover_buffer)->cairo;
+	cairo_surface_t *surf = cairo_get_target(cairo);
+
+	/* Background */
+	/* FIXME: support downscaling */
+	cairo_set_source_surface(cairo, icon,
+		(width - icon_width) / 2, (height - icon_height) / 2);
+	cairo_paint(cairo);
```
Consolatis, line 107: I think we should implement downscaling before merging this. Otherwise the hover overlay will become very ugly (and I don't actually know what happens if the `x` and `y` coordinates for cairo become negative).

johanmalm:
All review comments have been addressed.
Need to do an interactive-rebase before final merge.

johanmalm:
Couldn't contain myself... have done the fixup onto the middle commit.

src/theme.c
```diff
@@ -56,6 +57,43 @@ drop(struct lab_data_buffer **buffer)
...
+	assert(!*hover_buffer);
+
+	struct surface_context icon =
+		get_cairo_surface_from_lab_data_buffer(icon_buffer);
+	int icon_width = cairo_image_surface_get_width(icon.surface);
+	int icon_height = cairo_image_surface_get_height(icon.surface);
+
+	/* FIXME: need to somehow respect rounded corners */
+	int width = MAX(SSD_BUTTON_WIDTH, icon_width);
+	int height = MAX(theme->title_height, icon_height);
```
Consolatis, line 74: This may break the aspect ratio of the icon. Played around a bit with this and came up with this variant:
```diff
diff --git a/src/theme.c b/src/theme.c
index 6608b780..5a7a5913 100644
--- a/src/theme.c
+++ b/src/theme.c
@@ -69,9 +69,27 @@ create_hover_fallback(struct theme *theme, struct lab_data_buffer **hover_buffer
 	int icon_width = cairo_image_surface_get_width(icon.surface);
 	int icon_height = cairo_image_surface_get_height(icon.surface);
 
-	/* FIXME: need to somehow respect rounded corners */
-	int width = MAX(SSD_BUTTON_WIDTH, icon_width);
-	int height = MAX(theme->title_height, icon_height);
+	/* TODO: need to somehow respect rounded corners */
+	int width = SSD_BUTTON_WIDTH;
+	int height = theme->title_height;
+
+	if (width && height) {
+		/*
+		 * Maybe scale up the buffer proportionally.
+		 * It will get scaled down again by wlroots
+		 * when rendered and being required by the
+		 * current output scale.
+		 *
+		 * This ensures that icons > width or > height
+		 * keep their aspect ratio and are rendered the
+		 * same as without the hover overlay.
+		 */
+		double scale = MAX((double)icon_width / width, (double)icon_height / height);
+		if (scale > 1.0f) {
+			width = (double)width * scale;
+			height = (double)height * scale;
+		}
+	}
 
 	*hover_buffer = buffer_create_cairo(width, height, 1.0, true);
 ```

It is still not perfect:
- tried using `/usr/share/icons/hicolor/256x256/apps/chromium.png` as `max-active.png` and it still changing its aspect ratio slightly on hover
- the 6x6 default icons look even more blurry with output scale 2 when the hover is active, I played around with `CAIRO_ANTIALIAS_NONE` and `CAIRO_FILTER_FAST` but had no success.

Overall I'd consider this good enough to re-enable the fallback hover overlays.

johanmalm: Thanks. Agree. Done (force-pushed against middle commit)

Consolatis:
:)

PR #1342, opened 2023-12-19
Do not deactive windows when layer-shell client takes keyboard focus (to fix sfwbar minimize action)

johanmalm:


src/seat.c
```diff
@@ -352,26 +352,40 @@ focus_change_notify(struct wl_listener *listener, void *data)
 	struct view *view = event->new_surface ?
 		view_from_wlr_surface(event->new_surface) : NULL;
 
+	/* Prevent focus switch to layershell client from updating view state */
+	if (!view && server->active_view && event->new_surface
```
jlindgren90, line 356: It looks like the `server->active_view` condition was copied from the `view_is_related` logic later on? I think it's unnecessary here.

src/seat.c
```diff
@@ -352,26 +352,40 @@ focus_change_notify(struct wl_listener *listener, void *data)
...
+			&& wlr_layer_surface_v1_try_from_wlr_surface(event->new_surface)) {
+		return;
+	}
+
+	/*
+	 * Due to the previous conditions we may end up in a situation where
+	 * we give focus to a previously focused view that was never deactivated
+	 */
+	if (view == server->active_view) {
```
jlindgren90, line 365: This early-out (`view == server->active_view`) is redundant with the reverse condition (`view != server->active_view`) later on. Let's remove one or the other.

src/input/keyboard.c
```diff
@@ -104,8 +104,8 @@ match_keybinding_for_sym(struct server *server, uint32_t modifiers,
 			continue;
 		}
 		if (server->seat.nr_inhibited_keybind_views
-				&& server->focused_view
-				&& server->focused_view->inhibits_keybinds
+				&& server->active_view
+				&& server->active_view->inhibits_keybinds
```
jlindgren90, line 108: There may be a corner-case here: if a view (game, VM, etc.) is inhibiting keybindings and a layer-shell surface receives focus, the keybindings will still be inhibited. Probably not a big deal.

src/seat.c
johanmalm: Do we actually even need the `!view` in the same place?

jlindgren90: Probably not

johanmalm: Fixed.
Tested with sfwbar... works.

src/seat.c
johanmalm: Good catch. Removed the first one.

src/input/keyboard.c
johanmalm: We could add `&& view->surface == server.seat->keyboard_state.focused_surface`

Or wrap it in a `view_has_keyboard_focus()`

jlindgren90: I'm not sure if `view->surface == server.seat->keyboard_state.focused_surface` is quite correct because the view may have multiple "related" surfaces (for example xwayland-unmanaged popups).

Maybe to avoid complicating things too much, we just let this corner case be (i.e. just do the simple `focused_view` -> `active_view` replacement like you have currently).

johanmalm: Yes, good point. If it becomes a problem, we'll have to think about it a bit deeper.

jlindgren90:
LGTM

Issue #1351, opened 2023-12-22
Segmentation fault when opening CLions settings window in certain situations


jp7677:
Hi there, 

running labwc with essentially https://github.com/labwc/labwc/commit/a3f880716fa183d9e76d01f36c864d6fe51feece (and initial graphics tablet support on top of it) I'm seeing a segmentation fault in labwc when  opening CLion's settings windows in some situations. I can reproduce this fairly good when the main window is docked to the side (not sure if this is relevant) and clicking from a kind-of hover menu on the settings button. I hope this screenshot clarifies what I mean:

![out](https://github.com/labwc/labwc/assets/17219293/9fc7f463-2e7b-41e5-a74d-23395f4dd773)

Running with `gdb` reveals this line https://github.com/labwc/labwc/blob/master/src/xwayland.c#L546:

```gdb
Thread 1 "labwc" received signal SIGSEGV, Segmentation fault.
0x00000000004238a5 in init_foreign_toplevel (view=0x12d94c0) at ../src/xwayland.c:546
546		if (!parent->toplevel.handle) {

(gdb) bt
#0  0x00000000004238a5 in init_foreign_toplevel (view=0x12d94c0) at ../src/xwayland.c:546
#1  0x0000000000423acb in xwayland_view_map (view=0x12d94c0) at ../src/xwayland.c:619
#2  0x000000000041f646 in handle_map (listener=0x12d9650, data=0x0) at ../src/view.c:1747
#3  0x00007ffff7f94dcc in wl_signal_emit_mutable () at /lib64/libwayland-server.so.0
#4  0x00007ffff7f94dcc in wl_signal_emit_mutable () at /lib64/libwayland-server.so.0
#5  0x00007ffff7ef3847 in surface_commit_state () at /lib64/libwlroots.so.12
#6  0x00007ffff7724056 in ffi_call_unix64 () at /lib64/libffi.so.8
#7  0x00007ffff77205ef in ffi_call_int.lto_priv () at /lib64/libffi.so.8
#8  0x00007ffff77233fe in ffi_call () at /lib64/libffi.so.8
#9  0x00007ffff7f93843 in wl_closure_invoke.constprop () at /lib64/libwayland-server.so.0
#10 0x00007ffff7f980b4 in wl_client_connection_data () at /lib64/libwayland-server.so.0
#11 0x00007ffff7f968e2 in wl_event_loop_dispatch () at /lib64/libwayland-server.so.0
[...]

PR #1352, opened 2023-12-22
xwayland: Prevents usage of invalid data pointer

jp7677:
There are situations where the data pointer of a parent is NULL. Checking for NULL first fixes e.g. a segmentation fault in CLion when opening the settings window from a popup menu.

See also https://github.com/labwc/labwc/issues/1351#issuecomment-1867475324

Closes https://github.com/labwc/labwc/issues/1351


Consolatis:
Sweet, thanks :)

PR #1370, opened 2023-12-27
[RFC] Sync pending width, height when applying geometry to view

ahesford:
Applications that do not properly respect pending resizes, such as mpv when it enforces a fixed aspect ratio for video, will honor a resize but may substitute different values than that stored in the pending box. This doesn't trigger any of the usual "failed to respond" error paths, but will leave the pending box different from the actual on-screen size, potentially breaking later operations that consider the pending size for geometry changes. To counteract this, the pending width and height can be forced to the value assigned to the current box whenever a specific geometry is applied.

This can be seen, for example, by opening a video with mpv and attempting to interactively resize the view. When clicking on a corner resize handle and dragging, the view will grow or shrink according to a fixed aspect ratio. However, if the mouse is dragged more along one axis (allowing the resize cursor to drift far from actual corner of the window during the resize operation), the `view->pending` box will capture the size implied by the cursor, while `view->current` will capture whatever size mpv allows. Later actions that use the `view->pending` box, *e.g.* the `MoveToEdge` action, will incorrectly calculate the distance to edges, and the window may be prevented from reaching an edge or may extend beyond an edge.

Forcing the width and height of the pending box during calls to `view_impl_apply_geometry` solves this problem by ensuring that the pending box reflects what has actually been assigned to the view. It *seems* to work as expected when using `GrowToEdge` actions as well as resizing windows (both XDG and Xwayland) with the mouse. However, I am not sure if there are other ramifications for this change. It certainly fixes the problem of resizing mpv and trying to move it to edges.

Also, I snuck in a commit to make git ignore vim swap files; I can drop this if desired.


Consolatis:
I am not sure we can do this here, there might be multiple resizes stacked on top of each other and we are using `pending.height/width` in `view_impl_apply_geometry()` for adjusting the top and left edges during interactive resize.
Maybe we could instead move this to `src/xdg.c` and verify that the commit serial is equal to the pending one and then set the pending size? (e.g. in the ` if (serial > 0 && serial == xdg_surface->current.configure_serial) {` branch).

CC @jlindgren90 

> Also, I snuck in a commit to make git ignore vim swap files; I can drop this if desired.

Fine by me.

jlindgren90:
This is wrong for XWayland since the window manager is supposed to get the final say regarding window geometry. We just recently included (in 0.7.0) a change that does the opposite of what this PR is trying to do.

Not sure about xdg-shell.

johanmalm:
> This is wrong for XWayland since the window manager is supposed to get the final say regarding window geometry. We just recently included (in 0.7.0) a change that does the opposite of what this PR is trying to do.
> 
> Not sure about xdg-shell.

@jlindgren90 - do you remember trying to get this right for xfce4-terminal (which only resizes to discrete size increments)? That's a good example of where we let the client decide the size.... is it not?

johanmalm:
I've been able to reproduce with `mpv` by dragging (trying to) the bottom edge. I expected the window to resize as it does when dragging for corners or left/right edges. I observed that it did not visible change size, but (as @ahesford) described, when later using `MoveToEdge`, the calculations are off.

johanmalm:
Might the answer here be to simply `s/pending/current/` in `view_move_to_edge()`?
I think we have gradually drifted towards using `pending` in those type of operations, but maybe that's not right.

jlindgren90:
> @jlindgren90 - do you remember trying to get this right for xfce4-terminal (which only resizes to discrete size increments)? That's a good example of where we let the client decide the size.... is it not?

It's an example of the opposite -- we ignore the client requested geometry. See #790.

jlindgren90:
I would like to look at the `mpv` issue and figure out what precisely is happening. My understanding is that for "normal" (non-override-redirect) XWayland toplevels, it's not possible for the client to reject the size set by the window manager.

Let's start with a simple question, is `mpv` using XWayland or xdg-shell?

johanmalm:
> Let's start with a simple question, is `mpv` using XWayland or xdg-shell?

I'm running it as `xdg-shell`.

johanmalm:
Ah, I forget you force some x11 backends. When I mentioned `xfce4-panel` earlier, I was referring to the xdg-shell variant.

`xfce4-terminal` (xdg-shell) does increment in steps - so must therefore (unless I'm confused) ignore pending at times. I can see this when doing `MoveToEdge` (which is based on pending). I had never really thought about it much, but with `mpv` it's a lot more obvious.

`mpv` (xdg-shell) does maintain its aspect ratio - even when I try to resize it off that ratio. So again, I think that means that pending is ignored.

jlindgren90:
I'm guessing this is a case where we need different logic for xdg-shell than for XWayland. In the X11 model, the client requests a size change and the window manager gets the final say on whether to honor it. In xdg-shell it's kind of the inverse: the window manager has to make the size request and the client gets to decide whether to honor it.

Does `handle_configure_timeout` (in xdg.c) trigger for the `mpv` issue?

jlindgren90:
> When I mentioned xfce4-panel earlier, I was referring to the xdg-shell variant.

Makes sense. I tested xfce4-terminal with xdg-shell just now and it does ignore/modify the size that labwc requests.

For xdg-shell views, would something like this work (not tested)?
```
index 68f2f4ee..98493b5d 100644
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -128,6 +128,9 @@ handle_commit(struct wl_listener *listener, void *data)
 
        if (update_required) {
                view_impl_apply_geometry(view, size.width, size.height);
+               if (!view->pending_configure_serial) {
+                       view->pending = view->current;
+               }
        }
 }
 
```

johanmalm:
> Does `handle_configure_timeout` (in xdg.c) trigger for the `mpv` issue?

Yes.

I tried with grabbing the top/bottom edges (which is when mpv does not resize when I ask it to) and the 100ms WLR_INFO message every time.

johanmalm:
> For xdg-shell views, would something like this work (not tested)?
> 
> ```
> diff --git a/src/xdg.c b/src/xdg.c
> index 68f2f4ee..98493b5d 100644
> --- a/src/xdg.c
> +++ b/src/xdg.c
> @@ -128,6 +128,9 @@ handle_commit(struct wl_listener *listener, void *data)
>  
>         if (update_required) {
>                 view_impl_apply_geometry(view, size.width, size.height);
> +               if (!view->pending_configure_serial) {
> +                       view->pending = view->current;
> +               }
>         }
>  }
>  
> ```

No, doesn't solve the `MoveToEdge` problem with `mpv` after attemped resize with top/bottom edge.

Chatted with ahesford on IRC earlier.

@Consolatis suggested something similar. https://gist.github.com/Consolatis/a5bcae17af3d4042821650432332755d

Need to get off to bed.

johanmalm:
Just one final thought before running...


```
index 638e111..16202fd 100644
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -147,6 +147,7 @@ handle_configure_timeout(void *data)
 
 	view_impl_apply_geometry(view, view->current.width,
 		view->current.height);
+	view->pending = view->current;
 
 	return 0; /* ignored per wl_event_loop docs */
 }
```

This does indeed fix it.

jlindgren90:
> Just one final thought before running...
> 
> ```
> diff --git a/src/xdg.c b/src/xdg.c
> index 638e111..16202fd 100644
> --- a/src/xdg.c
> +++ b/src/xdg.c
> @@ -147,6 +147,7 @@ handle_configure_timeout(void *data)
>  
>  	view_impl_apply_geometry(view, view->current.width,
>  		view->current.height);
> +	view->pending = view->current;
>  
>  	return 0; /* ignored per wl_event_loop docs */
>  }
> ```
> 
> This does indeed fix it.

Makes sense, this would be the other possibility. I think we should add the "pending = current" in both places then.

ahesford:
Thanks all for the insight. I had tested a variant of the `handle_commit` change suggested by @Consolatis on IRC but was still encountering the issue; it turns out this was because I had overlooked the call to `view_impl_apply_geometry` in `handle_configure_timeout`. By resetting `view->pending` after applying the geometry in *both* `handle_commit` and `handle_configure_timeout`, the problem with mpv seems to go away. The latest commit reflects both of these suggestions.

This leaves two questions:
1. Is the call to `view_impl_apply_geometry` in `xdg_toplevel_view_configure`, added in commit 31ec8f05, still necessary after this change?
2. If yes, should this call also be followed by `view->pending = view->current`?

jlindgren90:
Current version looks good to me, thanks very much @ahesford.

> 1. Is the call to `view_impl_apply_geometry` in `xdg_toplevel_view_configure`, added in commit [31ec8f0](https://github.com/labwc/labwc/commit/31ec8f050cf9f2aafed1dd30490dc0aaf8ef32e0), still necessary after this change?

In theory it should be safe to revert 31ec8f050cf9f2aafed1dd30490dc0aaf8ef32e0 after this change. The `current` and `pending` width/height should be the same in that branch now (unless I overlooked something).

johanmalm:
Really nice. Big thanks for discovering + fixing.

johanmalm:
I've tested with 31ec8f0 reverted and xfce4-panel resizing works fine. I'll do a revert PR.

PR #1373, opened 2023-12-27
osd: refactor to eliminate usage of g_strcmp0()

johanmalm:
...to allow inclusion of ./scripts/find-banned in CI


Consolatis:
:)

Also +1 for the `get_title_if_different()` name change.

Issue #1367, opened 2023-12-25
virtual keyboard (wlrctl keyboard) weirdness: letters 'y' and 'z' getting dropped


jonleivent:
This is bizarre.  In labwc 0.7.0, doing:
```
wlrctl keyboard type 'why are y and z missing?'
```
produces:
```
wh are  and  missing?
```
It was not doing that prior to 0.7.0.

johanmalm:
I can re-produce. How weird. 
Any idea what commit brought this on?
If not, I'll try to do some bisecting.

jonleivent:
> Any idea what commit brought this on?

No idea.  I would expect that it's lower level - maybe down in wlroots.  I wouldn't think labwc would have any code that could do something like this.  So to speed up bisecting, maybe try right before and after you up the wlroots version?

Remember that I was previously stuck on the multisock branch, which is 0.6.4 based, prior to trying 0.7.0.  I haven't checked in between.

Narrat:
Took the liberty. Needs confirmation from someone, as I could have made a mixup.
But this did appear in the labwc 0.6.x releases, that I'm sure
```
5a77a3ad5176b09ff8576b8a6d05e53bcb4302d1 is the first bad commit
commit 5a77a3ad5176b09ff8576b8a6d05e53bcb4302d1
Author: John Lindgren <john@jlindgren.net>
Date:   Fri Nov 10 22:00:04 2023 -0500

    keyboard: remove nr_pressed_keys == 1 restriction for keybindings
    
    This restriction should be unnecessary now (see the previous commit for
    details) and caused issues with keybindings not working on some systems
    where irregular keypress events are received (e.g. XF86XK_WakeUp)
    without an accompanying release event.
    
    Kept separate from the previous commit for the sake of potential future
    bisects.
    
    Partially reverts e77330bc3fe7.

 src/input/keyboard.c | 13 -------------
 1 file changed, 13 deletions(-)
```

johanmalm:
Thanks for reporting. And @Narrat thanks for the analysis.

I think I solved it (#1376).

Testing welcome.

jonleivent:
Just tried the fix, and it works.  Thanks!
Now, if only I had a workaround for #1366...

PR #1376, opened 2023-12-28
keyboard: fix virtual keyboard bug

johanmalm:
Do no process virtual keyboard keycodes (just the keysyms).

Reproduce bug by issuing `wlrctl keyboard type xyz` and observe only 'x' when 'xyz' was expected.

The 'y' and 'z' were matched in match_keybinding() in the keycode section and returned keybinds for `XF86_AudioLowerVolume` and `XF86_AudioRaiseVolume` respectively.

Fixes: #1367


Consolatis:
Nice catch, totally makes sense!

PR #1378, opened 2023-12-28
feat: add "AutoPlace" action

ahesford:
The AutoPlace action will apply `placement_find_best()` to an active view, moving it to a position on its output that will minimize overlap with other views. This may be useful interactively to rearrange windows, but is particularly useful when chained with a `ResizeTo` action in a window rule: the initial placement, when "automatic" is selected, will optimize for a size that is then immediately changed; a follow-up `AutoPlace` action can ensure that the new window size will not push the window off the screen.


johanmalm:
Nice!

Issue #1366, opened 2023-12-24
menu Execute item can no longer use wlrctl keyboard to put text into focused window in 0.7.0


jonleivent:
I have a labwc menu Execute action item that uses `wlrctl keyboard` to put some text into the focused window.  It was working prior to 0.7.0 (I had been using the multisock branch - so whatever that was based on).  When a menu is up, where is keyboard data sent to by labwc?  Has it changed to send it to somewhere else other than the currently focused window?  The same command, when run from a labwc keybinding, still works.  So the problem is what labwc does with virtual keyboard text when a menu is up.  I can understand there being a reason to direct that text at the menu if it has a way to use text.  Is there a workaround?

jonleivent:
It looks like this is due to the fix for #1058 and #1059, which were fixed earlier than 0.7.0.  I was stuck on the multisock branch, and didn't pick up those fixes earlier.

Would it be possible to turn off keyboard input to the menu after a menu item is selected but before that item is executed, so that any text the item generates goes to the focused window instead of back to the menu?

Consolatis:
Hm.. good point and nice research. Does it work when you use `Return` / `Enter` to trigger the menu entry?

jonleivent:
> Does it work when you use Return / Enter to trigger the menu entry?

Yes it does work!

PR #1379, opened 2023-12-28
menu: run actions on button release

johanmalm:
...and call actions after closing menus so that virtual keyboard input caused by actions are sent to the surface with keyboard-focus rather than being consumed by the open menu.

Fixes: #1366

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -940,13 +940,11 @@ cursor_button_press(struct seat *seat, uint32_t button,
 	}
 
 	if (server->input_mode == LAB_INPUT_STATE_MENU) {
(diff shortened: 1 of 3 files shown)


johanmalm:
This could do with some testing + review as it was written by a tired author.

johanmalm:
It's the other way around.

With a menu entry as below, this patch makes "abc" appear in the window that had focus before opening the menu - and after.

```
  <item label="abc">
    <action name="Execute">
      <command>wlrctl keyboard type abc</command>
    </action>
  </item>
```

ahesford (approved):
In that case, these changes (rebased) work as expected.

PR #1382, opened 2023-12-29
rc.xml.all: replace "" with "default" in libinput device category

domo141:
As empty string there is not matched as how it was expected.

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -407,7 +407,7 @@
       - tapButtonMap [lrm|lmr]
   -->
   <libinput>
-    <device category="">
+    <device category="default">
       <naturalScroll></naturalScroll>
       <leftHanded></leftHanded>
       <pointerSpeed></pointerSpeed>


johanmalm:
Or is it better to handle `""` so that it creates the `default` profile. Seems more intuitive.

johanmalm:
Just a thought... I wonder if it would be easier to understand (+ make the code a bit easier) if we change the configuration to:

1. Use `<device category="touch|non-touch|touchpad|default">` for defining settings for device 'categories' (but not named devices). And make it clearer in the documentation that the `default` profile is only used if there is no other category match of name match.
2. Use `<device name="foo">` when we want to create a profile of one particular, named device.

That would obviously introduce a regression (but we could be backward compatible).

Thoughts welcome.

domo141:
For handling "" it seems not trivial to do -- or the one solution I saw I did not understand...

(I tried
``` 
+++ b/src/config/rcxml.c
@@ -462 +462,2 @@ fill_libinput_category(char *nodename, char *content)
-               if (!strcmp(content, "touch")
+               if (content[0] == '\0'
+                               || !strcmp(content, "touch")
```
which did not work.

Then, this related gist is uncomprensible to me ;/

https://gist.github.com/Consolatis/8fffc3b100a4a698fada23177cf1de70

( msg'd in IRC 2023-06-16  10:53 UTC )

----

For the other comment, I got so many mixed thoughts in my head have to rethink...

domo141:
hmm, johanmalm has added related PR If it fixes the thing this PR can be abandoned :dagger:

johanmalm:
Need to just mull it over a bit. Could we delay decision for a bit.

domo141:
Sure, I'd like to see your solution work.

Consolatis:
> Then, this related gist is uncomprensible to me ;/
> 
> https://gist.github.com/Consolatis/8fffc3b100a4a698fada23177cf1de70
> 
> ( msg'd in IRC 2023-06-16 10:53 UTC )

Completely forgot about that one. That seems to treat `<device>` (e.g. without any `category` attribute) as default category while

> +++ b/src/config/rcxml.c
> @@ -462 +462,2 @@ fill_libinput_category(char *nodename, char *content)
> -               if (!strcmp(content, "touch")
> +               if (content[0] == '\0'
> +                               || !strcmp(content, "touch")

should treat `<device category="">` as default.

As for why it doesn't appear to work: not sure.

In any case I think this PR makes sense regardless of the other changes.

domo141:
When I added:
```
        if (!strcmp(nodename, "category")) {
+               wlr_log(WLR_ERROR, "category content: '%s'", content);
                if (!strcmp(content, "touch")
```

I recall that log line did not appear when category="" -- and therefore that `content[0] == '\0'`
makes no difference.

Consolatis:
> I recall that log line did not appear when category="" -- and therefore that `content[0] == '\0'` makes no difference.

I wonder if libxml treats an empty argument (`<foo category="" />`) the same as a boolean argument then (`<foo category />`). That would mean we return earlier (in the `!content` branch on top). However that should in turn mean that `category=""` already works by accident as we create it on top and initialize it with `DEFAULT_DEVICE`. I must be missing something here.

domo141:
my suggestion is that this is merged as is (no need to change anything), and then
the content of

 https://github.com/labwc/labwc/pull/1384

is changed so that it works --I don't understand why the change helps, but perhaps
johanmalm does :D

PR #1384, opened 2023-12-29
config: handle libinput <device category=""> as default

johanmalm:
Fixes: #1382


src/config/libinput.c
```diff
@@ -26,7 +26,7 @@ libinput_category_init(struct libinput_category *l)
 enum device_type
 get_device_type(const char *s)
 {
-	if (!s) {
+	if (!s || !*s) {
 		return DEFAULT_DEVICE;
 	}
```
Consolatis, line 31: This is technically not needed, none of the following would match and we'd return `DEFAULT_DEVICE` anyway.

Consolatis:
LGTM

src/seat.c
```diff
@@ -57,6 +57,37 @@ device_type_from_wlr_device(struct wlr_input_device *wlr_input_device)
...
 
+/*
+ * Get applicable profile (category) by matching first by name and secondly be
+ * type (e.g. 'touch' and 'non-touch'). If not suitable match is found based on
+ * those two criteria we fallback on 'default'.
+ */
+static struct libinput_category *
+get_category(struct wlr_input_device *device)
+{
+	enum device_type type = device_type_from_wlr_device(device);
```
domo141: it looks to me that if this could be done in line 79 (split in 2 if c89 required)

src/seat.c
```diff
@@ -57,6 +57,37 @@ device_type_from_wlr_device(struct wlr_input_device *wlr_input_device)
...
+
+	/* By name */
+	struct libinput_category *category = NULL;
```
domo141, line 71: does compiler optimize the writing of NULL to category away.

I did 

`git grep -B1 wl_list_for_each | less`

and did not see settings such vars set to NULL (in the lines it printed out)

include/config/libinput.h
```diff
@@ -7,6 +7,9 @@
 #include <wayland-server-core.h>
 
 enum device_type {
+	LAB_LIBINPUT_DEVICE_NONE = 0,
+
+	// FIXME: Rename the entries below with a LAB_LIBINPUT_ prefix
```
domo141: comment could be ... LAB_LIBINPUT_DEVICE_ prefix ...

domo141 (approved):
LGTM AFAIU, with just some thoughts in comments...

src/seat.c
johanmalm: Good point. Have moved it.

src/seat.c
johanmalm: Fair point. Have removed =NULL

include/config/libinput.h
johanmalm: Overtaken by events. Have already sorted out the enums.

PR #1385, opened 2023-12-29
feat: Add tablet rotate transformation

jp7677:
Same as the other tablet PRs, based on @Consolatis work here https://github.com/labwc/labwc/issues/1060#issuecomment-1786384495

### DIFF
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -83,6 +83,7 @@ struct rcxml {
 
 	/* graphics tablet */
 	struct tablet_config {
+		enum rotation rotation;
 		uint16_t button_map_count;
 		struct button_map_entry button_map[BUTTON_MAP_MAX];
 	} tablet;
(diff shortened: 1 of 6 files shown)


include/config/rcxml.h
```diff
@@ -83,6 +83,7 @@ struct rcxml {
 
 	/* graphics tablet */
 	struct tablet_config {
+		enum rotation rotate;
```
johanmalm, line 86: Suggest `rotation` instead of `rotate`.... so `enum rotation rotation`

johanmalm (commented):
All looks really tidy.
Just one minor suggestion.
I can't test it, but trust it works :smile:

include/config/rcxml.h
jp7677: Adjusted, thanks!

jp7677:
> All looks really tidy. Just one minor suggestion. I can't test it, but trust it works 😄

Very cool!. And yes, the transformation works, using `rotate="180"` I can turn around my tablet and the cursor still follows my stylus moves ;)

johanmalm:
With thanks.

PR #1387, opened 2023-12-30
feat: Add active tablet area specification

jp7677:
This PR allow to specify the responsive/active area of the tablet. The Wacom driver on Windows allows something similar, see e.g. https://101.wacom.com/UserHelp/en/PenMode.htm
I'm using this to truncate the active area so that the resulting aspect ratio matches the one from my monitor. Initially i had something like this https://github.com/labwc/labwc/issues/1060#issuecomment-1826875020 in mind, but accessing the output turned out rather tricky on startup and actually this approach here is much more flexible.

This is the last PR from me for anything with tablet cursor movement (for now) ;)

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -394,14 +394,22 @@
   </mouse>
 
   <!--
+    The active tablet area can be specified by setting the top/left
+    coordinate (in mm) and/or width/height (in mm). If width or height
+    are omitted or default (0.0), width/height will be set to the
+    remaining width/height seen from top/left.
+
     The tablet orientation can be changed in 90 degree steps, thus
-    rotation can be set to [0|90|180|270].
+    rotation can be set to [0|90|180|270]. Rotation will be applied
+    after applying tablet area transformation.
 
     Tablet buttons emulate regular mouse buttons. The tablet *button* can
     be set to any of [tip|stylus|stylus2|stylus3]. Valid *to* mouse buttons
     are [left|right|middle].
   -->
   <tablet rotate="0">
+    <!-- Active area dimensions are in mm -->
+    <area top="0.0" left="0.0" width="0.0" height="0.0">
     <map button="tip" to="left" />
     <map button="stylus" to="right" />
     <map button="stylus2" to="middle" />
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -83,6 +83,7 @@ struct rcxml {
 
 	/* graphics tablet */
 	struct tablet_config {
+		struct wlr_fbox box;
 		enum rotation rotation;
 		uint16_t button_map_count;
 		struct button_map_entry button_map[BUTTON_MAP_MAX];
--- a/include/config/tablet.h
+++ b/include/config/tablet.h
@@ -17,6 +17,7 @@ struct button_map_entry {
 	uint32_t to;
 };
 
+double tablet_get_dbl_if_positive(const char *content, const char *name);
 enum rotation tablet_parse_rotation(int value);
 uint32_t tablet_button_from_str(const char *button);
 uint32_t mouse_button_from_str(const char *button);
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -826,6 +826,14 @@ entry(xmlNode *node, char *nodename, char *content)
 		} else {
 			wlr_log(WLR_ERROR, "Invalid value for <resize popupShow />");
 		}
+	} else if (!strcasecmp(nodename, "left.area.tablet")) {
+		rc.tablet.box.x = tablet_get_dbl_if_positive(content, "left");
+	} else if (!strcasecmp(nodename, "top.area.tablet")) {
+		rc.tablet.box.y = tablet_get_dbl_if_positive(content, "top");
+	} else if (!strcasecmp(nodename, "width.area.tablet")) {
+		rc.tablet.box.width = tablet_get_dbl_if_positive(content, "width");
+	} else if (!strcasecmp(nodename, "height.area.tablet")) {
+		rc.tablet.box.height = tablet_get_dbl_if_positive(content, "height");
 	} else if (!strcasecmp(nodename, "rotate.tablet")) {
 		rc.tablet.rotation = tablet_parse_rotation(atoi(content));
 	} else if (!strcasecmp(nodename, "button.map.tablet")) {
@@ -999,7 +1007,7 @@ rcxml_init(void)
 	rc.doubleclick_time = 500;
 	rc.scroll_factor = 1.0;
 
-	rc.tablet.button_map_count = 0;
+	rc.tablet.box = (struct wlr_fbox){0};
 	tablet_load_default_button_mappings();
 
 	rc.repeat_rate = 25;
--- a/src/config/tablet.c
+++ b/src/config/tablet.c
@@ -7,6 +7,17 @@
 #include "config/tablet.h"
 #include "config/rcxml.h"
 
+double
+tablet_get_dbl_if_positive(const char *content, const char *name)
+{
+	double value = atof(content);
+	if (value < 0) {
+		wlr_log(WLR_ERROR, "Invalid value for tablet area %s", name);
+		return 0;
+	}
+	return value;
+}
+
 enum rotation
 tablet_parse_rotation(int value)
 {
@@ -84,6 +95,7 @@ tablet_button_mapping_add(uint32_t from, uint32_t to)
 void
 tablet_load_default_button_mappings(void)
 {
+	rc.tablet.button_map_count = 0;
 	tablet_button_mapping_add(BTN_TOOL_PEN, BTN_LEFT); /* Used for the pen tip */
 	tablet_button_mapping_add(BTN_STYLUS, BTN_RIGHT);
 	tablet_button_mapping_add(BTN_STYLUS2, BTN_MIDDLE);
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -127,6 +127,8 @@ setup_pen(struct seat *seat, struct wlr_input_device *wlr_device)
 	tablet->tablet->data = tablet;
 	tablet->x = 0.0;
 	tablet->y = 0.0;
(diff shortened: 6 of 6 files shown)


johanmalm:
Nice. No problems in principle, but have a few review comments. Will go through it shortly.

docs/rc.xml.all
```diff
@@ -394,14 +394,28 @@
...
-    rotation can be set to [0|90|180|270].
+    rotation can be set to [0|90|180|270]. Note that rotation will be
+    applied after applying tablet area transformation.
 
     Tablet buttons emulate regular mouse buttons. The tablet *button* can
     be set to any of [tip|stylus|stylus2|stylus3]. Valid *to* mouse buttons
     are [left|right|middle].
   -->
   <tablet rotate="0">
+    <area top="0,0" left="0,0" bottom="0,0" right="0,0">
```
johanmalm, line 418: Should be `.` (point) rather than `,` (comma)

include/config/rcxml.h
```diff
@@ -83,6 +83,7 @@ struct rcxml {
 
 	/* graphics tablet */
 	struct tablet_config {
+		struct tablet_area tablet_area;
```
johanmalm, line 86: I suggest using `struct wlr_fbox box` here instead. It's more consistent with code-base and avoids having to define another area/box struct.

include/config/tablet.h
```diff
@@ -4,6 +4,13 @@
 
 #include <stdint.h>
 
+struct tablet_area {
```
johanmalm, line 7: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/util/box.h?ref_type=heads#L38

src/config/rcxml.c
```diff
@@ -826,6 +826,34 @@ entry(xmlNode *node, char *nodename, char *content)
 		} else {
 			wlr_log(WLR_ERROR, "Invalid value for <resize popupShow />");
 		}
+	} else if (!strcasecmp(nodename, "top.area.tablet")) {
+		double top = atof(content);
```
johanmalm, line 830: There is a lot of duplication in the four branches here. Could we write a helper function like `rc.tablet.box.x = get_non_neg_double(content)`. Could add error message as second argument. Return 0 if negative.

src/input/tablet.c
```diff
@@ -11,6 +11,29 @@
...
+{
+	if ((area_top == 0 && area_right == 0 && area_bottom == 0 && area_left == 0)
+			|| tablet_width == 0 || tablet_height == 0) {
+		return;
+	}
+
+	if (area_right > 0 && area_right <= tablet_width) {
+		double area_width = area_right - area_left;
+		double area_width_offset = area_left / tablet_width; // x is in range 0..1
+		*x = (*x - area_width_offset) * tablet_width / area_width;
```
johanmalm, line 27: I'm a bit confused by the geometry concepts here. Is `x` representing the `1` in `0..1`?
If so, 
1. Is there a better name as `x` makes me think of x0 rather than x1 if you know what I mean.
2. What happens if a user wants to change `left` (in terms of tablet area)?

src/input/tablet.c
```diff
@@ -50,6 +73,10 @@ handle_axis(struct wl_listener *listener, void *data)
 
 		double x = tablet->x;
 		double y = tablet->y;
+		adjust_for_tablet_area(tablet->tablet->width_mm, tablet->tablet->height_mm,
+			rc.tablet.tablet_area.top_mm, rc.tablet.tablet_area.right_mm,
```
johanmalm, line 77: If we use `wlr_fbox` we can pass one argument here rather the four (i.e. `top_mm`, `right_mm`, `bottom_mm`, `left_mm`). Makes it easier to read.

jp7677:
@johanmalm Thanks a lot for the good review and especially for the hint about `wlr_fbox`. Using that as a representation (thus width/height instead of bottom/right) makes things also easier. Let me know what you think about the result.

PS: I didn't really know where to put `get_non_neg_double`, please let me know if there is a better place.

src/input/tablet.c
jp7677: yeah, sorry, that comment is confusing, it was more a note for me that max-x and max-y is always 1. I've converted that into a `const`,  I hope that makes things a bit more readable.

I'm now calculating width/height if it hasn't been specified, with that it also works when just `top` or `left` is specified, thus when only the top of left side should be truncated.

jp7677:
> PS: I didn't really know where to put `get_non_neg_double`, please let me know if there is a better place.

On second thought, I've put this helper now into `config/tablet`. Fits better imho.

src/config/rcxml.c
```diff
@@ -826,6 +826,14 @@ entry(xmlNode *node, char *nodename, char *content)
 		} else {
 			wlr_log(WLR_ERROR, "Invalid value for <resize popupShow />");
 		}
+	} else if (!strcasecmp(nodename, "left.area.tablet")) {
+		rc.tablet.box.x = tablet_get_area_value(content, "left");
+	} else if (!strcasecmp(nodename, "top.area.tablet")) {
+		rc.tablet.box.y = tablet_get_area_value(content, "top");
+	} else if (!strcasecmp(nodename, "width.area.tablet")) {
+		rc.tablet.box.width = tablet_get_area_value(content, "width");
```
johanmalm, line 834: Just a thought, should we bound check the width/height against tablet->mm_width

src/config/tablet.c
```diff
@@ -7,6 +7,17 @@
 #include "config/tablet.h"
 #include "config/rcxml.h"
 
+double
+tablet_get_area_value(const char *content, const char *name)
```
johanmalm, line 11: Consider parsing percentages, for example 50%

src/input/tablet.c
```diff
@@ -11,6 +11,34 @@
 #include "input/cursor.h"
 #include "input/tablet.h"
 
+static void
+adjust_for_tablet_area(double tablet_width, double tablet_height,
+		struct wlr_fbox box, double *x, double *y)
+{
+	if ((box.x == 0 && box.y == 0 && box.width == 0 && box.height == 0)
```
johanmalm, line 18: We prefer !foo instead of foo==0

docs/rc.xml.all
```diff
@@ -394,14 +394,28 @@
   </mouse>
 
   <!--
+    The responsive tablet area can be specified by setting the lop/left
+    and/or width/height coordinates (in mm). This allows, among others,
+    to set the tablet area aspect ratio to the aspect ratio of the
+    output.
+
+    Example:
```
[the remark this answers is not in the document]
jp7677: I've reworked the description. it should clarify now how this works instead of how i use it. Does it makes sense now?
The aspect ratio of the tablet is not maintained, that was just my own use case (to change the active area that the aspect ratio from tablet changes from 16:9 to 21:9 so that it matches the one from my monitor)

include/config/tablet.h
```diff
@@ -17,10 +17,12 @@ struct button_map_entry {
 	uint32_t to;
 };
 
+double tablet_get_area_value(const char *content, const char *name);
```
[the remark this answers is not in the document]
jp7677: yes, done, I went for `tablet_get_dbl_if_positive`.

src/config/rcxml.c
```diff
@@ -999,7 +1007,7 @@ rcxml_init(void)
 	rc.doubleclick_time = 500;
 	rc.scroll_factor = 1.0;
 
-	rc.tablet.button_map_count = 0;
+	tablet_load_default_tablet_area();
```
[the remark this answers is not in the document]
jp7677: ah, there shines my c/cpp knowledge ;), much better indeed, thanks!

src/config/rcxml.c
jp7677: I'm doing the bounds check when doing the actual transformation in  https://github.com/labwc/labwc/blob/27bba861e85283cc7619e512fd0b300ead0d5257/src/input/tablet.c#L30 and https://github.com/labwc/labwc/blob/27bba861e85283cc7619e512fd0b300ead0d5257/src/input/tablet.c#L35

src/config/tablet.c
jp7677: That would indeed be cool, but i don't have the tablet dimensions available here. If we allow this, we would need to do the conversion from relative to absolute when doing the actual conversion in https://github.com/labwc/labwc/blob/27bba861e85283cc7619e512fd0b300ead0d5257/src/input/tablet.c#L15 
We would also need to store if width/height are absolute of relative. This won't fit into `wlr_fbox` .  I would prefer to not do this as part of this PR and eventually come back to this later.

jp7677:
@johanmalm thanks a lot again for your valuable feedback, I've addressed all points in code or as comment.

src/config/tablet.c
johanmalm: Maybe one for the future :smile:

src/config/rcxml.c
johanmalm: Good point!

johanmalm:
Thanks! Good effort.

jp7677:
Very cool, thanks for your patience and guidance!

docs/rc.xml.all
```diff
@@ -394,14 +394,22 @@
...
+    rotation can be set to [0|90|180|270]. Rotation will be applied
+    after applying tablet area transformation.
 
     Tablet buttons emulate regular mouse buttons. The tablet *button* can
     be set to any of [tip|stylus|stylus2|stylus3]. Valid *to* mouse buttons
     are [left|right|middle].
   -->
   <tablet rotate="0">
+    <!-- Active area dimensions are in mm -->
+    <area top="0.0" left="0.0" width="0.0" height="0.0">
```
Narrat, line 412: Noticed this typo after adding updating my config with the new content.
Should be `<area top="0.0" left="0.0" width="0.0" height="0.0" />`

jp7677: Thanks for point this out!
-> https://github.com/labwc/labwc/pull/1425

PR #1389, opened 2023-12-30
view: honor automatic placement when adjusting floating geometry

ahesford:
The `view_adjust_floating_geometry` function is called when un-maximizing a window or changing the output layout to ensure that views are well placed. Rather than always centering these views should they fall offscren, use the automatic placement strategy if so configured.

This alters the interface of `placement_find_best` to support passing an arbitrary geometry, because we wish to pass either `view->natural_geometry` or `view->pending`, depending on the reason for the placement.

### DIFF
--- a/include/placement.h
+++ b/include/placement.h
@@ -3,8 +3,9 @@
 #define LABWC_PLACEMENT_H
 
 #include <stdbool.h>
+#include <wlr/util/box.h>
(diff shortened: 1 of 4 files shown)


johanmalm:
Neat. Tested with `WLR_WL_OUTPUTS=2 ./build/labwc`
LGTM

PR #1390, opened 2023-12-30
Tearing support

Ph42oN:
This adds allowTearing option. `<allowTearing>yes</allowTearing>` will enable tearing support. Also adds action ToggleTearing that can be used to enable it for windows that don't request tearing.

I decided to make it check if WLR_DRM_NO_ATOMIC is set to 1, and disable tearing otherwise, as currently that is required for it to work. That check can be removed later when we get support for atomic tearing page flips.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -133,10 +133,10 @@ this is for compatibility with Openbox.
 	Allow tearing to reduce input lag. Default is no.
 	This open requires setting the environment variable WLR_DRM_NO_ATOMIC=1.
 
-	*yes* allow tearing if an open window requests it.
+	*yes* allow tearing if the active window requests it.
 
-	*fullscreen* allow tearing if an open window requests it, or if any
-	window is in fullscreen mode.
+	*fullscreen* allow tearing if the active window requests it or is in
+	fullscreen mode.
 
 	*always* allow tearing regardless of window status.
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -109,6 +109,7 @@ this is for compatibility with Openbox.
   <decoration>server</decoration>
   <gap>0</gap>
   <adaptiveSync>no</adaptiveSync>
+  <allowTearing>no</allowTearing>
   <reuseOutputMode>no</reuseOutputMode>
 </core>
 ```
@@ -127,6 +128,17 @@ this is for compatibility with Openbox.
 
 	*fullscreen* enables adaptive sync whenever a window is in fullscreen
 	mode.
+
+*<core><allowTearing>* [yes|no|fullscreen|always]
+	Allow tearing to reduce input lag. Default is no.
+	This open requires setting the environment variable WLR_DRM_NO_ATOMIC=1.
+
+	*yes* allow tearing if an open window requests it.
+
+	*fullscreen* allow tearing if an open window requests it, or if any
+	window is in fullscreen mode.
+
+	*always* allow tearing regardless of window status.
 
 *<core><reuseOutputMode>* [yes|no]
 	Try to re-use the existing output mode (resolution / refresh rate).
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -133,8 +133,7 @@ this is for compatibility with Openbox.
 	Allow tearing to reduce input lag. Default is no.
 	This option requires setting the environment variable WLR_DRM_NO_ATOMIC=1.
 
-	*yes* allow tearing if the active window requests it,
-	or its enabled with action ToggleTearing.
+	*yes* allow tearing if requested by the active window.
 
 *<core><reuseOutputMode>* [yes|no]
 	Try to re-use the existing output mode (resolution / refresh rate).
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -129,16 +129,12 @@ this is for compatibility with Openbox.
 	*fullscreen* enables adaptive sync whenever a window is in fullscreen
 	mode.
 
-*<core><allowTearing>* [yes|no|fullscreen|always]
+*<core><allowTearing>* [yes|no]
 	Allow tearing to reduce input lag. Default is no.
-	This open requires setting the environment variable WLR_DRM_NO_ATOMIC=1.
+	This option requires setting the environment variable WLR_DRM_NO_ATOMIC=1.
 
-	*yes* allow tearing if the active window requests it.
-
-	*fullscreen* allow tearing if the active window requests it or is in
-	fullscreen mode.
-
-	*always* allow tearing regardless of window status.
+	*yes* allow tearing if the active window requests it,
+	or its enabled with action ToggleTearing.
 
 *<core><reuseOutputMode>* [yes|no]
 	Try to re-use the existing output mode (resolution / refresh rate).
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -39,6 +39,7 @@
 #include <wlr/types/wlr_drm_lease_v1.h>
 #include <wlr/types/wlr_virtual_pointer_v1.h>
 #include <wlr/types/wlr_virtual_keyboard_v1.h>
+#include <wlr/types/wlr_tearing_control_v1.h>
 #include <wlr/util/log.h>
 #include "config/keybind.h"
 #include "config/rcxml.h"
@@ -318,6 +319,9 @@ struct server {
 	struct wlr_pointer_constraints_v1 *constraints;
 	struct wl_listener new_constraint;
 
+	struct wlr_tearing_control_manager_v1 *tearing_control;
+	struct wl_listener tearing_new_object;
+
 	/* Set when in cycle (alt-tab) mode */
 	struct osd_state {
 		struct view *cycle_view;
@@ -479,6 +483,8 @@ void handle_output_power_manager_set_mode(struct wl_listener *listener,
 void output_add_virtual(struct server *server, const char *output_name);
 void output_remove_virtual(struct server *server, const char *output_name);
 void output_enable_adaptive_sync(struct wlr_output *output, bool enabled);
+void new_tearing_hint(struct wl_listener *listener, void *data);
+void set_tearing(struct output *output);
 
 void server_init(struct server *server);
 void server_start(struct server *server);
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -357,6 +357,7 @@ struct output {
 
 	bool leased;
 	bool gamma_lut_changed;
+	bool tearing;
 };
 
 #undef LAB_NR_LAYERS
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -361,7 +361,6 @@ struct output {
 
 	bool leased;
 	bool gamma_lut_changed;
-	bool tearing;
 };
 
 #undef LAB_NR_LAYERS
@@ -484,7 +483,6 @@ void output_add_virtual(struct server *server, const char *output_name);
 void output_remove_virtual(struct server *server, const char *output_name);
 void output_enable_adaptive_sync(struct wlr_output *output, bool enabled);
 void new_tearing_hint(struct wl_listener *listener, void *data);
-void set_tearing(struct output *output);
 
 void server_init(struct server *server);
 void server_start(struct server *server);
--- a/src/action.c
+++ b/src/action.c
@@ -956,8 +956,9 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_TOGGLE_TEARING:
 			if (view) {
 				view->tearing_hint = !view->tearing_hint;
+				wlr_log(WLR_DEBUG, "tearing %sabled",
+					view->tearing_hint ? "en" : "dis");
 			}
-			wlr_log(WLR_DEBUG, "tearing: %d", view->tearing_hint);
 			break;
 		case ACTION_TYPE_INVALID:
 			wlr_log(WLR_ERROR, "Not executing unknown action");
--- a/src/action.c
+++ b/src/action.c
@@ -102,6 +102,7 @@ enum action_type {
 	ACTION_TYPE_VIRTUAL_OUTPUT_ADD,
 	ACTION_TYPE_VIRTUAL_OUTPUT_REMOVE,
 	ACTION_TYPE_AUTO_PLACE,
+	ACTION_TYPE_TOGGLE_TEARING,
 };
 
 const char *action_names[] = {
@@ -149,6 +150,7 @@ const char *action_names[] = {
 	"VirtualOutputAdd",
 	"VirtualOutputRemove",
 	"AutoPlace",
+	"ToggleTearing",
 	NULL
 };
 
@@ -950,6 +952,12 @@ actions_run(struct view *activator, struct server *server,
 					view_move(view, x, y);
 				}
 			}
+			break;
+		case ACTION_TYPE_TOGGLE_TEARING:
+			if (view) {
+				view->tearing_hint = !view->tearing_hint;
+			}
+			wlr_log(WLR_DEBUG, "tearing: %d", view->tearing_hint);
 			break;
 		case ACTION_TYPE_INVALID:
 			wlr_log(WLR_ERROR, "Not executing unknown action");
--- a/src/output.c
+++ b/src/output.c
@@ -68,9 +68,7 @@ output_frame_notify(struct wl_listener *listener, void *data)
 		return;
 	}
 
-	if (output->tearing) {
-		output->wlr_output->pending.tearing_page_flip = true;
-	}
+	output->wlr_output->pending.tearing_page_flip = output->tearing;
 
 	lab_wlr_scene_output_commit(output->scene_output);
 
@@ -277,7 +275,7 @@ new_output_notify(struct wl_listener *listener, void *data)
 
 	wl_list_init(&output->regions);
 
-	if (rc.allow_tearing == LAB_TEARING_ENABLED) {
+	if (rc.allow_tearing == LAB_TEARING_ALWAYS) {
 		output->tearing = true;
 	} else {
 		output->tearing = false;
--- a/src/output.c
+++ b/src/output.c
@@ -33,13 +33,8 @@ get_tearing_preference(struct output *output)
 	struct server *server = output->server;
 
 	/* Never allow tearing when disabled */
-	if (rc.allow_tearing == LAB_TEARING_DISABLED) {
+	if (!rc.allow_tearing) {
 		return false;
-	}
-
-	/* Allows allow tearing when forced */
-	if (rc.allow_tearing == LAB_TEARING_ALWAYS) {
-		return true;
 	}
 
 	/* Tearing is only allowed for the output with the active view */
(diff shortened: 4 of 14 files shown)


johanmalm:
> What would be good name for adaptive_sync_mode if its used in this too? Another option is to just use bool, rename option to something like tearingFullscreen for example, and drop support for enabling tearing always.

Suggest a dedicated enum as below (or similar). Keeps it flexible/readable.

```
enum tearing_mode {
  LAB_TEARING_DISABLED = 0,
  LAB_TEARING_ENABLED,
  LAB_TEARING_ON_FULLSCREEN_ONLY,
};
```

ahesford:
"forceTearing" seems counterintuitively named and emphasizes an undesirable but permitted side effect rather than the desired outcome, which (I assume) is lower input latency by eliminating extra layers of frame buffers. Nobody *wants* tearing, but people may be willing to tolerate it to reduce latency. Something like "allowTearing" or "reduceLatency" is probably more descriptive of the intent.

johanmalm:
I've never spent much time working on output frames, so it's not an area I've got my head around really.

Technically I'm not sure what we've got to do to achieve 'tearning'.

Does `lab_wlr_scene_output_commit()` enable it? If so, I'm keen to understand that's the right way of going about it. 

Note that sway does this quite differently with `max_render_time`, etc. See: https://github.com/Nefsen402/sway/blob/2f82cdae5a8d2c0204b06855cccf7f470a587cea/sway/desktop/output.c#L267

Ph42oN:
That thing in sway is not the same thing, tearing support for sway is not merged yet https://github.com/swaywm/sway/pull/7598/files.

Comment about function `lab_wlr_scene_output_commit()` says "This is a copy of wlr_scene_output_commit() as it doesn't use the pending state at all.", and pending is used to set tearing here, so it doesnt get applied with `wlr_scene_output_commit()`.

Ph42oN:
> "forceTearing" seems counterintuitively named and emphasizes an undesirable but permitted side effect rather than the desired outcome, which (I assume) is lower input latency by eliminating extra layers of frame buffers. Nobody _wants_ tearing, but people may be willing to tolerate it to reduce latency. Something like "allowTearing" or "reduceLatency" is probably more descriptive of the intent.

Well, i hate added latency of vsync so much that for months before, i was running with wlroots patch that enables tearing always including on desktop. Anyways i can change name to allowTearing.

Sinaaaa:
Now that this is merged, how do I enable tearing?

LAB_TEARING_FULLSCREEN=1    as an ENV?

Ph42oN:
> Now that this is merged, how do I enable tearing?
> 
> LAB_TEARING_FULLSCREEN=1 as an ENV?

Not merged, but if you build from my branch you can use <allowTearing>fullscreen</allowTearing> in rc.xml.

daniel005007:
so setting `<adaptiveSync>yes</adaptiveSync>` and `<allowTearing>fullscreen</allowTearing>` will make all windows except fullscreen windows use adaptive sync?

Sinaaaa:
> Not merged

Is this just a matter of time, or the problem is the base principle that only certain protocols can be supported in labwc?

Ph42oN:
I asked previously if they want to merge it when its only option to enable on fullscreen or always. But now i implemented tearing protocol support, so i think this patch is starting to be ready.

Now `<allowTearing>yes</allowTearing>` will enable tearing protocol to get window hints to enable tearing, and `<allowTearing>fullscreen</allowTearing>` does same and also allow tearing in fullscreen.  `<allowTearing>always</allowTearing>` will enable it always.

Do you think its good like this?

src/view.c
```diff
@@ -273,6 +273,19 @@ set_adaptive_sync_fullscreen(struct view *view)
...
+set_tearing(struct view *view)
+{
+	if (rc.allow_tearing < LAB_TEARING_ENABLED) {
+		return;
+	}
+	if (rc.allow_tearing == LAB_TEARING_FULLSCREEN && view->fullscreen) {
+		view->output->tearing = true;
+	} else {
+		view->output->tearing = view->tearing_hint;
+	}
```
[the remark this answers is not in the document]
ahesford: Thinking about this more, it might be more obvious if this function explicitly sets the tearing state in all cases, rather than relying on just preserving state for the disabled and always cases. For example:

```c
switch (rc.allow_tearing) {
case LAB_TEARING_DISABLED:
    view->output->tearing = false;
    break;
case LAB_TEARING_ALWAYS:
    view->output->tearing = true;
    break;
case LAB_TEARING_FULLSCREEN:
    view->output->tearing = view->fullscreen;
    break;
case LAB_TEARING_ENABLED:
    view->output->tearing - view->tearing_hint;
    break;
default:
    assert(false);
}
```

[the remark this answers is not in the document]
Ph42oN: It was done like that to use this function to set tearing only when its switched on or off depending on view, while disabled and always have tearing constantly off or on. I changed always to -1 to keep code shorter here. But yea switch case sounds like would make this more clear.

src/view.c
```diff
@@ -1869,6 +1884,9 @@ view_destroy(struct view *view)
 		if (rc.adaptive_sync == LAB_ADAPTIVE_SYNC_FULLSCREEN) {
 			wlr_output_enable_adaptive_sync(view->output->wlr_output, false);
 		}
+		if (rc.allow_tearing == LAB_TEARING_FULLSCREEN) {
+			view->output->tearing = false;
+		}
```
ahesford, line 1889: I'm not sure this is required if you use `set_tearing` to force the right value every time (*i.e.*, after every window activation or full-screen state switch). I'm also not sure this properly handles state transitions in some corner cases where, e.g., you have other full-screen views open that should still force tearing but may not be active.

docs/labwc-config.5.scd
```diff
@@ -128,6 +129,17 @@ this is for compatibility with Openbox.
...
+*<core><allowTearing>* [yes|no|fullscreen|always]
+	Allow tearing. Default is no.
+	Enabling requires that WLR_DRM_NO_ATOMIC is set to 1.
+
+	*yes* enables tearing protocol.
+
+	*fullscreen* enables tearing protocol and allows tearing whenever
+	a window is in fullscreen mode.
+
+	*always* enables tearing always.
```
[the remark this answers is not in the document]
Ph42oN: Do you think `fullscreen` option should do it for only fullscreen? Current version enables it for fullscreen and still allows windows that request tearing to tear without fullscreen.

include/labwc.h
```diff
@@ -318,6 +319,10 @@ struct server {
 	struct wlr_pointer_constraints_v1 *constraints;
 	struct wl_listener new_constraint;
 
+	struct wlr_tearing_control_manager_v1 *tearing_control;
+	struct wl_listener tearing_new_object;
+	struct wl_list tearing_controllers;
```
ahesford, line 324: This seems unused...

Sinaaaa:
> Do you think its good like this?

I think it's great. If by Window Hints you mean that we would be able to put a button (menu item) into window decorations that toggles tearing that would be absolutely amazing.

docs/labwc-config.5.scd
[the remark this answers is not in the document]
ahesford: That seems OK, but that's one of the things I find confusing in this description. After reviewing the rest of the implementation, maybe this is a reasonable description:
```suggestion
	*yes* allows tearing when the active window requests it.

	*fullscreen* allows tearing when the active window requests it or is fullscreen.

	*always* allows tearing in at all times.
```

include/labwc.h
Ph42oN: Oh yea wl_list tearing_controllers is unused, i tried if it works without that part and it did so i removed stuff and forgot that. Other 2 are used in server.c.

src/view.c
Ph42oN: Well, i wouldn't think anyone would care as tearing should be reactivated when the view becomes active.

Setting it that way sounds like good idea, at least in terms of keeping code simple.

Ph42oN:
> > Do you think its good like this?
> 
> I think it's great. If by Window Hints you mean that we would be able to put a button (or menu item) into window decorations that toggles tearing that would be absolutely amazing.

It means windows that request tearing will have it enabled. With testing i did many games enabled tearing with that.

ahesford:
I wonder... is it more appropriate to allow tearing if the *active* window requests it, or if *any* window requests it?

If the latter, then it seems like the right thing is to drop the calls to `set_tearing` and just set `output->wlr_output->pending.tearing_page_flip` directly before each commit. If `LAB_TEARING_FULLSCREEN` or `LAB_TEARING_ENABLED`, you can just walk all windows on the output to see if any are fullscreen (for `_FULLSCREEN`) or set a tearing hint to true. If so, set `tearing_page_flip` and you're done.

johanmalm:
> Is this just a matter of time, or the problem is the base principle that only certain protocols can be supported in labwc?

We want to support the tearing protocol. This is a stopgap to support tearing sooner than most people would have access by waiting for implementations+release of wlroots, labwc and whatever client they wish to tear with (because the client would have to implement the protocol too).

Sinaaaa:
> Maybe it could be possible to add an action that forces tearing for a specific window, so it could also be used via window rules?

The same thing occurred to me just now as well. That would be not only amazing, but only thing I would ever need in this regard.

Ph42oN:
> I wonder... is it more appropriate to allow tearing if the _active_ window requests it, or if _any_ window requests it?
> 
> If the latter, then it seems like the right thing is to drop the calls to `set_tearing` and just set `output->wlr_output->pending.tearing_page_flip` directly before each commit. If `LAB_TEARING_FULLSCREEN` or `LAB_TEARING_ENABLED`, you can just walk all windows on the output to see if any are fullscreen (for `_FULLSCREEN`) or set a tearing hint to true. If so, set `tearing_page_flip` and you're done.

Im not sure but only times i care about latency is when the view is active, so i would see it as working how i want is as long as its enabled when active view requests it. I did it this way to because i felt like i wouldn't want to add more complex code inside `output_frame_notify()`.
> Maybe it could be possible to add an action that forces tearing for a specific window, so it could also be used via window rules?

Adding that would be possible.

Ph42oN:
> > Is this just a matter of time, or the problem is the base principle that only certain protocols can be supported in labwc?
> 
> We want to support the tearing protocol. This is a stopgap to support tearing sooner than most people would have access by waiting for implementations+release of wlroots, labwc and whatever client they wish to tear with (because the client would have to implement the protocol too).

Tearing protocol is now supported, but i would want other ways to enable tearing too for cases where its wanted for windows that do not request tearing.

johanmalm:
> Tearing protocol is now supported, but i would want other ways to enable tearing too for cases where its wanted for windows that do not request tearing.

Sorry, I'll try to keep up :smile:

ahesford:
@Ph42oN see [my changes](https://github.com/Ph42oN/labwc/compare/tearing...ahesford:labwc:zonai-master) which will enable tearing on an output if any view requests it or is fullscreen. The principal advantages are that the changes are more localized (they leave out any alterations to `view.c`) and avoid relying on any long-term state consistency through `output->tearing`. Instead, the actual tearing request is made based on the current state of the output *every time*.

Ph42oN:
> @Ph42oN see [my changes](https://github.com/Ph42oN/labwc/compare/tearing...ahesford:labwc:zonai-master) which will enable tearing on an output if any view requests it or is fullscreen. The principal advantages are that the changes are more localized (they leave out any alterations to `view.c`) and avoid relying on any long-term state consistency through `output->tearing`. Instead, the actual tearing request is made based on the current state of the output _every time_.

I think this looks pretty good, But i was thinking that i would want to avoid adding lot of code in that part to avoid any possible extra latency from it.

ahesford:
It's not "extra code" that adds latency, but walking the window tree. However, this is substantially less error prone and should add a negligible amount of extra computation, especially because we bail out of the loop after the first window asking to permit tearing is discovered. If applications require *extremely* low latency:
1. The whole compositor probably needs to be redesigned around that goal.
2. Just set `allowTearing` to `always` and short-circuit the window checks altogether.

Ph42oN:
> It's not "extra code" that adds latency, but walking the window tree. However, this is substantially less error prone and should add a negligible amount of extra computation, especially because we bail out of the loop after the first window asking to permit tearing is discovered. If applications require _extremely_ low latency:
> 
>     1. The whole compositor probably needs to be redesigned around that goal.
> 
>     2. Just set `allowTearing` to `always` and short-circuit the window checks altogether.

Ok i don't know if it really is anywhere near enough to really matter. Anyways just now i made new version using your code to choose if tearing is enabled, but not set it every frame.

ahesford:
This still sacrifices clarity and simplicity for the possibility of a negligible performance benefit. Looping over window objects and testing a couple of variables is asking the PC to count to ~10 every time it updates the scene. (I doubt a workflow where input latency matters this much will have more than ~20 open windows.) This is a negligible effort compared to everything else that is being done in a frame interval.

The tradeoff for this is another state variable that must be maintained in sync with intent, which adds maintenance burden and barriers to understanding as people seek to trace the logic that governs tearing.

ahesford:
Alright, I have an idea that seems both clean and will appeal to your desire for minimal cost. The [changes in my branch](https://github.com/Ph42oN/labwc/compare/tearing...ahesford:labwc:zonai-master) go back to the notion that only the currently active window should govern tearing. We can query `server->active_view` directly:
1. If tearing is always false, so be it.
2. If tearing is always on, enable it for every output.
3. If the active window is not on the output for which we are trying to commit the scene, or there is no active window, we do not enable tearing for that commit.
4. Otherwise, enable tearing if the active window requests it or (when enabled by the config) the active window is fullscreen.

This avoids the walk altogether, and because `allowTearing` is all about input latency and that doesn't really matter if the view isn't active, we don't need to jump through extra hoops to decide whether to enable tearing for the commit. Even better, we still aren't capturing any output state that has to be kept in sync every time we do something with the window. This should be the best of both worlds!

Ph42oN:
Ok i think this is good as i don't see reason why anyone would need tearing for other than active view.

Ph42oN:
Actually, im not sure if we should keep these fullscreen and always options. Other way enable tearing for windows that don't request it would be to add action that can be used with windowrules or keybind.

ahesford:
A `ToggleTearing` action that just flips the `tearing_hint` attribute would seem to satisfy any need that `always` and `fullscreen` might address.

Ph42oN:
> A `ToggleTearing` action that just flips the `tearing_hint` attribute would seem to satisfy any need that `always` and `fullscreen` might address.

I tried doing this, but it appears to run into problem that some windows can request to not tear, and it will be override what ToggleTearing sets. But simple edit to set tearing_hint only when it requests tearing fixes this.

docs/labwc-config.5.scd
```diff
@@ -128,6 +129,13 @@ this is for compatibility with Openbox.
 	*fullscreen* enables adaptive sync whenever a window is in fullscreen
 	mode.
 
+*<core><allowTearing>* [yes|no]
+	Allow tearing to reduce input lag. Default is no.
+	This option requires setting the environment variable WLR_DRM_NO_ATOMIC=1.
+
+	*yes* allow tearing if the active window requests it,
+	or its enabled with action ToggleTearing.
```
ahesford, line 137: "its" should be "it's", but I'd suggest just "allow tearing if requested by the active window". This action just makes the active window request tearing allowance if its application has not done so, and is covered by this single clause.

src/action.c
```diff
@@ -951,6 +953,12 @@ actions_run(struct view *activator, struct server *server,
 				}
 			}
 			break;
+		case ACTION_TYPE_TOGGLE_TEARING:
+			if (view) {
+				view->tearing_hint = !view->tearing_hint;
+			}
+			wlr_log(WLR_DEBUG, "tearing: %d", view->tearing_hint);
```
ahesford, line 960: This should be inside the `if (view)`.

src/output.c
```diff
@@ -27,6 +27,29 @@
...
+	if (!server->active_view || server->active_view->output != output) {
+		return false;
+	}
+
+	/* If the active view requests tearing, or it is toggled on with action, allow it */
+	if (server->active_view->tearing_hint) {
+		return true;
+	}
+
+	return false;
```
ahesford, line 50:
```suggestion
	return server->active_view->tearing_hint;
```

johanmalm:
@ahesford Thanks for doing that review.
When addressed, I'll go for final review + merge.

src/action.c
Ph42oN: Right, also maybe i should make it say enabled/disabled instead of 0/1.

johanmalm:
Thanks. Would you mind adding the option to `docs/rc.xml.all`.
Otherwise LGTM

Sinaaaa:
Okay so I have     `<allowTearing>yes</allowTearing>` in my rc.xml & `WLR_DRM_NO_ATOMIC=1` in my environment file. 

So how do I convince a game to be a client that requests tearing? (I'm mostly using Bottles to run my games)

johanmalm:
> Okay so I have `<allowTearing>yes</allowTearing>` in my rc.xml & `WLR_DRM_NO_ATOMIC=1` in my environment file.
> 
> So how do I convince a game to be a client that requests tearing? (I'm mostly using Bottles to run my games)

It might be that the game doesn't (yet) support the tearing protocols, so yeah, as @Flrian said, just use the `ToggleTearing` action. You could even add it to the window client-menu :smile:

Sinaaaa:
(My setup is as follows: 3 outputs connected to a Haswell igpu, Windows games rendered by an amd dedicated GPU (rx580). My main screen is mirrored all the time, there is a third screen on the side. All this is set up with wlr-randr)

If I have mirroring on, then the tearing enabled window does not tear at all on the mirrored display. But as I drag it across to my side monitor tearing starts working. If I disable mirroring, tearing works, but not always. The toggle sometimes does not seem to work either. Seems like I lost the ability tear any game with the last mesa patch even on xorg. I should probably give this another year & figure out how to bring tearing back on Xorg. 
At least compositing lag in games has improved a lot over the past couple of weeks. 

On a sidenote I tried tearing on Hyprland and there and -there alone- it works, however the cursor will start to draw at the game's framerate. I can play at 30fps, but I cannot if the the cursor is also at 30fps. :-)

Ph42oN:
> If I have mirroring on, then the tearing enabled window does not tear at all on the mirrored display. But as I drag it across to my side monitor tearing starts working. If I disable mirroring, tearing works, but not always.

Does it not tear at all while mirrored? I think maybe it would need some extra checks for mirrored to allow tearing. It is currently enabled only on output of active window, and i don't think it can point to more than 1 output at same time.

> On a sidenote I tried tearing on Hyprland and there and -there alone- it works, however the cursor will start to draw at the game's framerate. I can play at 30fps, but I cannot if the the cursor is also at 30fps. :-)

Are you using adaptive sync, on labwc moving mouse while cursor is visible causes refresh rate to jump to monitors max refresh rate, keeping cursor smooth always, but this has the issue of adaptive sync not working for game content while moving mouse.

Not doing that would be better to me. Or if we could make it do something like depending on how much fps, cursor would get drawn at 2x or 4x the fps game runs, i think that would be best.

Sinaaaa:
> Does it not tear at all while mirrored? I think maybe it would need some extra checks for mirrored to allow tearing, but i would have expected it to work at least on 1 of the mirrored monitors. It is currently enabled only on output of active window, and i don't think it can point to more than 1 output at same time.
Does not seem to tear at all. (I have mirroring on, so I can use my computer from the couch)
> Are you using adaptive sync
Nope, I have regular 60hz displays, all 3 of them.

ahesford:
This conversation is drifting well beyond the original scope of this work. Rather than continuing in a closed PR, please open a new issue (if this behavior is properly considered a bug) or a discussion (if the desire is for an open-ended consideration of how potentially conflicting options should be reconciled).

Issue #1281, opened 2023-12-04
Changes in display power state destroy ouptut configuration and window arrangement


ahesford:
Related issue: https://github.com/labwc/labwc/issues/864

Changes in display power state, for example issuing a `wlopm --off` or just pushing the power button, will cause labwc to completely discard configuration and window arrangement for that output. I don't know if single-output setups are special-cased to some extent, but this is particularly problematic for multiple outputs:

- Any position configured or scale configured for that output will be lost
- All windows on that output will be gathered on another, active output (until the last remaining output is switched off)

Even with a single output, I suspect that a non-unit scale factor will be reset once the display is reactivated. This seems to be a consequence of disabling each powered-off output in `handle_output_power_manager_set_mode()` as [discussed here](https://github.com/labwc/labwc/issues/864#issuecomment-1493121122).

System details:
- Void Linux (rolling, up to date)
- labwc 0.6.6
- wlroots 0.16.2
- wlopm 0.1.0
- Video hardware: AMD Radeon RX 580

Steps to reproduce:
1. Launch `labwc` with a default configuration on multiple displays.
2. Launch a terminal on each output.
3. In one of the terminals, issue the command `wlopm --off '*' && sleep 20 && wlopm --on '*'`; adjust the sleep as necessary to allow each display to shut itself down.
4. Reconfigure the displays with some non-unit scaling. For example, with two 4K displays:
    ```sh
    wlr-randr --output DP-1 --pos 0,0 --scale 1.5 --output DP-2 --pos 2560,0 --scale 1.5
    ```
5. Again launch a terminal on each output.
6. Re-run step 3.

Expected behavior:
After Steps 3 and 6, the output configuration and window arrangement on each display should be the same as before the step was run.

Actual behavior:
All outputs will be reset to a default arrangement with unit scaling, and all windows will be coalesced onto one of the displays.

Comments:
It may be hard to distinguish a physical power-off of a display from pulling the cable, and one might argue that the appropriate response to physically disconnecting a display **should** cause labwc to reconfigure outputs. However, operations like `wolpm --off` seem primarily useful for power management during idle periods, and it seems appropriate in that case to save and restore all possible display state in that case.

johanmalm:
Thanks for reporting. Agree, that doesn’t seem right. Will try with scale 1.5 to see if I can reproduce.

ahesford:
[...]

PR #1392, opened 2024-01-01
view: try to honor original geometry with layout changes

ahesford:
If a view's output disappears during a layout change, its original geometry will be captured as last_layout_geometry, which will remain valid unless the view's natural geometry is modified. If a subsequent layout change adds an output closer to a valid last_layout_geometry, the view will be relocated to that output.

The intent of this change is to more gracefully handle cases like `wlopm --off` and a subsequent `wlopm --on` which, at least on some multihead `amdgpu` setups, will send disconnect events that will trigger labwc to coalesce all views onto the last display to be disconnected. However, it also allows one to connect an external display to a laptop, put some windows there, remove the display and have them coalesce on the internal display, and then reattach the display and have the windows pushed back to where they began.

There are a few points that should be addressed before this is ready to merge:
- [ ] I am currently using `view_store_natural_geometry` to invalidate the `last_layout_geometry`. This means that, after labwc relocates a view because an output was lost, an interactive move, tiling or maximization of the view on its current output will prevent restoration to an original output. This seems clearly desirable, but does not cover cases when the user resizes a floating window (an interactive resize does not update the natural geometry). Should interactive resizes also call `view_store_natural_geometry` to trigger the invalidation, should the interactive resize manually invalidate the `last_layout_geometry`, or is there some better method to invalidate the `last_layout_geometry` that I am not considering?
- [ ] `view_discover_output` has been overloaded to identify an output closest to the midpoint of the original `last_layout_geometry`, which should mean that, with each new layout change, each view not altered by the user will get as close as possible to its original output. This seems like reasonable behavior, but am open to suggestions for refining this behavior.
- [ ] The `view_adjust_floating_geometry` function has been modified to check not whether a view is currently visible *within the layout*, but instead whether it is currently visible *on its assigned output*. This seems like a more proper check anyway, but I am not sure if there are corner cases that I am overlooking.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -180,6 +180,13 @@ struct view {
 	 * maximized/fullscreen/tiled.
 	 */
 	struct wlr_box natural_geometry;
+	/*
+	 * Whenever an output layout change triggers a view relocation, the
+	 * last pending position (or natural geometry) will be saved so the
+	 * view may be restored to its original location on a subsequent layout
+	 * change.
+	 */
+	struct wlr_box last_layout_geometry;
 
 	/* used by xdg-shell views */
 	uint32_t pending_configure_serial;
@@ -420,6 +427,7 @@ bool view_is_floating(struct view *view);
 void view_move_to_workspace(struct view *view, struct workspace *workspace);
 void view_set_decorations(struct view *view, bool decorations);
 void view_toggle_fullscreen(struct view *view);
+void view_invalidate_last_layout_geometry(struct view *view);
 void view_adjust_for_layout_change(struct view *view);
 void view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_windows);
 void view_grow_to_edge(struct view *view, enum view_edge direction);
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -70,6 +70,7 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 		} else {
 			/* Store natural geometry at start of move */
 			view_store_natural_geometry(view);
+			view_invalidate_last_layout_geometry(view);
 		}
 
 		/* Prevent region snapping when just moving via A-Left mousebind */
@@ -86,6 +87,13 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 			 */
 			return;
 		}
+
+		/*
+		 * Resizing overrides any attempt to restore window
+		 * geometries altered by layout changes.
+		 */
+		view_invalidate_last_layout_geometry(view);
+
 		/*
 		 * If tiled or maximized in only one direction, reset
 		 * tiled/maximized state but keep the same geometry as
--- a/src/view.c
+++ b/src/view.c
@@ -252,14 +252,27 @@ view_get_edge_snap_box(struct view *view, struct output *output,
 	return dst;
 }
 
-static void
-view_discover_output(struct view *view)
+static bool
+view_discover_output(struct view *view, struct wlr_box *geometry)
 {
 	assert(view);
 	assert(!view->fullscreen);
-	view->output = output_nearest_to(view->server,
-		view->current.x + view->current.width / 2,
-		view->current.y + view->current.height / 2);
+
+	if (!geometry) {
+		geometry = &view->current;
+	}
+
+	struct output *output =
+		output_nearest_to(view->server,
+			geometry->x + geometry->width / 2,
+			geometry->y + geometry->height / 2);
+
+	if (output && output != view->output) {
+		view->output = output;
+		return true;
+	}
+
+	return false;
 }
 
 static void
@@ -343,7 +356,7 @@ view_moved(struct view *view)
 	 * output when they enter that state.
 	 */
 	if (view_is_floating(view)) {
-		view_discover_output(view);
+		view_discover_output(view, NULL);
 	}
 	ssd_update_geometry(view->ssd);
 	cursor_update_focus(view->server);
@@ -614,11 +627,11 @@ view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
 
 	bool adjusted = false;
 	/*
-	 * First check whether the view is onscreen. For now, "onscreen"
-	 * is defined as even one pixel of the client area being visible.
+	 * First check whether the view is the target screen, meaning that at
+	 * least one client pixel is on the screen.
 	 */
 	if (wlr_output_layout_intersects(view->server->output_layout,
-			NULL, geometry)) {
+			view->output->wlr_output, geometry)) {
 		/*
 		 * If onscreen, then make sure the titlebar is also
 		 * visible (and not overlapping any panels/docks)
@@ -988,6 +1001,7 @@ view_maximize(struct view *view, enum view_axis axis,
 		interactive_cancel(view);
 		if (store_natural_geometry && view_is_floating(view)) {
 			view_store_natural_geometry(view);
+			view_invalidate_last_layout_geometry(view);
 		}
 	}
 	set_maximized(view, axis);
@@ -1200,6 +1214,7 @@ view_set_fullscreen(struct view *view, bool fullscreen)
 		 */
 		interactive_cancel(view);
 		view_store_natural_geometry(view);
+		view_invalidate_last_layout_geometry(view);
 	}
 
 	set_fullscreen(view, fullscreen);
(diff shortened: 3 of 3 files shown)


Consolatis:
Thanks for working on this!

I currently can't wrap my head around all possible outcomes with this applied but the generic idea sounds fine to me.
CC @jlindgren90 

I will try to review this PR in more detail in the coming days.

johanmalm:
LTGM. But have not tested.

Testing ought to include doing `wlr-randr --off` against all outputs to make sure we don't crash the compositor.

Will leave merge as @Consolatis expressed intent to review. Also, this is an area of the code-base that both @Consolatis and @jlindgren90 have spent significant time on, so more eye-balls may help catch edge cases.

In response to your three points:

1. Invalidation

> Should interactive resizes also call `view_store_natural_geometry()`? 

I'm slightly confused by this. When I do an interactive-resize, followed by SnapToEdge, followed by dragging titlebar, I get the natural geometry set by the interactive resize. So it feels like we already do it - but I can't find where.

If it's missing, I think we should call `view_store_natural_geometry()`

2. Find output

Proposal seems very good. Concur.

3. Check visibility wrt `view->output` in `view_adjust_floating_geometry()`

Happy to go with this. Can't think of any undesired side-effects and it's early in the release cycle, so will get some testing.

ahesford:
In the resize -> snap -> drag sequence, the natural geometry is actually saved by the *snap*, not the resize. However, when an interactive move begins, the natural geometry is saved; I imagine this is done in case the move continues to a screen edge and triggers a maximize/snap operation, so that the governing natural geometry is the original position of the view rather than wherever the window ended up when the non-floating state was triggered.

As this PR currently implements the last-layout geometry, the following sequence can occur:
1. Put window somewhere on display 2.
2. Maximize window. The original window position is saved as natural geometry.
3. Disconnect display 2. The natural window position (because the window is not floating) is stored as the last-layout geometry. The window will be assigned to display 1, and the natural geometry will be adjusted to fit within display 1 (Lines 1291--1293). The window will then be maximized on display 1.
4. Un-maximize the window on display 1. The window will move to its now-adjusted natural geometry on display 1.
5. Interactively resize (but do not move) the window on display 1. This will *not* invalidate the last-layout geometry.
6. Reconnect display 2. The window will be reassigned to its original place on display 2 because the last-layout geometry is valid and indicates a preferred location there.

If Step 5 is replaced with *move* the window, or maximize again, or snap to edge, `view_store_natural_geometry` will be called, invalidating the last-layout geometry; in that case, reconnecting the display in Step 6 will *not* cause the window to be relocated to display 2.

On the one hand, it can be argued that this is ok, because maybe the window is being resized to fit better on the screen, but it isn't necessarily "moving" in a sense and users might prefer that the window bounce back to its original placement when display 2 comes back.

On the other hand, it can be argued that *any* time the user changes the geometry of a window, that should stop the compositor from trying to relocate the window should a new display be added to the layout later. If this is preferable, we should either explicitly invalidate the last-layout geometry at the start of any interactive resize, or just call `view_store_natural_geometry` to do a (meaningless) update of the natural geometry in addition to the last-layout invalidation.

Consolatis:
Fortunately wlroots 0.17 brought the availability of enabling / disabling of virtual outputs when running nested.
Thus `WLR_WL_OUTPUTS=2 labwc -d` actually works now with `wlr-randr --output WL-1 --off` and thus can be used to test things like this.

So far I noticed that floating views will correctly transition to the other output. However, a floating view that was positioned on the remaining output (`WL-2`) was also suddenly centered within this output which seems wrong. When re-enabling `WL-1` \*all\* views were transferred to `WL-1` and I could not make them stick to `WL-2` by moving or resizing.

This might have something to do with disabling the first output and thus the layout coordinates of the views now becoming `0,0` for `WL-2` as its now the only one left in the layout.

Another issue is when I tried `wlopm --off WL-1` to disable the output rather than `wlr-randr --output WL-1 --off`, the views from `WL-1` did not actually move over to `WL-1`.

ahesford:
> So far I noticed that floating views will correctly transition to the other output. However, a floating view that was already positioned on the remaining output (`WL-2`) was also suddenly centered within this output which seems wrong. When re-enabling `WL-1` *all* views were transferred to `WL-1` (including the one that was originally spawned there) and I could not make them stick to `WL-2` by moving or resizing.
>
> This might have something to do with disabling the first output and thus the layout coordinates of the views now becoming `0,0` for `WL-2` as its now the only one left in the layout.

You are right. When WL-1 disappears, all windows on WL-1 will trigger a capture of their current position as `last_layout_geometry`, and then `view_discover_output` will put them on WL-2. Because the output changes and they (now) have a last-layout geometry, these windows will be adjusted to fit on WL-2. (If WL-2 has the same shape as WL-1 and now has position `(0,0)`, the "adjustment" should be a no-op.) The natural geometry is then applied, which will put all views originally on WL-1 in the same relative locations on WL-2.

The windows originally on WL-2 will have x-positions that offset them from the original width of WL-1. When WL-1 disappears and WL-2 takes its place, they are suddenly offscreen. However, because their outputs have not disappeared, this *does not* trigger a capture of `last_layout_geometry` in lines 1241--1264. Instead, the later call to `view_discover_output` will find that these windows are not overlapped with their output, and the default placement (in your case, centering) will be applied to force them onscreen. (This kind of behavior is a lot nicer when new `automatic` placement is enabled.)

Later, when you restore WL-1, all of the original WL-1 windows will be yanked back to their original position based on their `last_layout_geometry` (or, if you've moved them around since, their current position, which will now overlap with WL-1). Likewise, all of the windows on the original WL-2, whether you've moved them or not, will not have a valid `last_layout_position` and will be relocated to WL-1 based on their positions when WL-2 started at (0, 0).

> Another issue is when I tried `wlopm --off WL-1` to disable the output rather than `wlr-randr --output WL-1 --off`, the views from `WL-1` did not actually move over to `WL-2`.

This is behavior I see on Intel video adapters even now. While it might require some rethinking of how `wlr-output-power-management-unstable-v1` and the `handle_output_power_manager_set_mode` handler are implemented, it is unrelated to placement. Disabling the output with `wlr_output_enable` does not seem sufficient to trigger a kernel uevent that will cause wlroots to destroy (and, when it is later reattached, re-create) the head. With `amdgpu`, the hardware itself seems to trigger disconnect and reconnect uevents when the display powers back on (hence #1281) and, apparently, so does `wlr-randr --off` using `wlr-output-management-unstable-v1`. My understanding of the issue starts to get shaky at this point...

> 
> Edit: Yet something else, I noticed
> 
> > 00:08:10.390 [../src/output.c:527] failed to get output layout box
> 
> showing up in the logs when disabling. I am not sure if that is related to this PR though.

I don't believe this is related, and should appear in the debug output even without the changes in this PR.

> Edit 2: Regarding the question on where to update `->last_geometry`: the natural geometry was only used to recover from (potentially layered) automatically calculated layouts (e.g. `SnapToEdge` -> `Maximize` -> `Fullscreen`). As such it was only necessary to update it when entering a new automatically calculated layout.
> 
> Although I guess we could connect that one to `->last_layout` I think its easier to maintain and understand in the long term to be more explicit and have a dedicated function like `view_store_last_geometry()` or something along that line rather than hook into natural geometry.

Based on this, I think it would be better to add a `view_invalidate_last_layout_geometry` function, and separately call the invalidator everywhere `view_store_natural_geometry` is called, *plus* at the beginning of an interactive resize.

Consolatis:
`wlr-randr --output WL-1 --off` triggers the output-management protocol handler:

- `handle_output_manager_apply()`
-  -> `output_config_apply()`
-  -> `do_output_layout_change()`
-  -> `output_update_for_layout_change()`
-  -> `output_update_all_usable_areas()`
-  -> `desktop_arrange_all_views()`
-  -> `view_adjust_for_layout_change()` 

`wlopm --off WL-1` in turn only disables the output. For that one we should be able to listen to the wlr_output->commit signal to figure out that the output state has changed and then call `output_update_for_layout_change()`. Or as alternative simply call it in the output-power-management handler (`handle_output_power_manager_set_mode()`) itself.

ahesford:
I added some behavioral improvements in the latest fixup that should reconcile some of the behavioral oddities:
- Expose a `view_invalidate_last_layout_geometry` that separates this behavior from `view_store_natural_geometry` and is called everywhere we set the natural geometry in addition to whenever an interactive resize starts (which means *some* user-initiated geometry change has occurred, so layout changes should not attempt to restore original geometries).
- Helper functions increase conceptual readability.
- The last-layout geometry is applied to the view not only when an output change has occurred, but also if the output stays the same but the pending (or natural, for non-floating windows) geometry is different from a valid last-layout geometry. This handles cases like WL-2 moving from (1280,0) to (0,0), triggering all of its windows to be moved to different coordinates without changing the output; then, when WL-2 is moved back to (1280,0), the new "opportunistic" application will detect that the window which was in the (0,0) origin should be moved back to the original position in the (1280,0) origin.
- To make sure everything works as expected in a number of strange output reconfigurations, a last-layout geometry will be captured for every view after *any* layout change (if a valid last-layout geometry hasn't already been assigned), so every active view will have an idea of where it "prefers" to be placed any time the layout changes.

Regarding `handle_output_power_manager_set_mode`: I added a `do_output_layout_change` call to the end of the function, and this does indeed cause `wdisplays` and `wlr-randr` to show an output as disabled after `wlopm --off`. However, this is not sufficient to cause labwc to relocate views on the disabled output. I'm not sure what the right course is to cause the expected behavior---maybe grabbing a `wlr_output_configuration_v1` struct from... somewhere.. and calling `output_config_apply` instead of `do_output_layout_change`---but I'm punting on this for now.

If the latest round of changes look reasonable, I will squash the fixup.

jlindgren90:
I have not had a chance to look at the code, but just a comment in regard to `wlopm --off`: I think we intentionally do not want to consider this a layout change. `wlopm --off` is intended to be used to turn off the displays when idle for a while so the window positions should not change when the system is waked again. That's why I believe we are not currently calling `do_output_layout_change()` from the output-power-management path.

I am not sure if all the different cases are handled correctly though. For example, what happens if a laptop is undocked from from external displays after `wlopm --off`? Ideally the laptop display should remain off but any views on the (now disconnected) external displays should be relocated to it. I doubt that we handle that case correctly today.

ahesford:
@jlindgren90 This is why I punted on the display-off issue here. I think there are a lot of nuances around what it means to be "off" or "disconnected" that need to be considered, and this PR addresses none of those concerns. Instead, the only purpose of this PR is to take the best possible action *whenever the layout **does** change*.

The layout improvements here provide some of quality-of-life improvements:
1. You can unplug your laptop from an external display, have your windows collected on the internal display, and use it for awhile; if you later plug in the display, your windows will go back to where they started.
2. On my amdgpu hardware, where `wlopm --off` actually causes the card to issue a hotplug disconnect/connect cycle, this will restore all windows to their pre-off state. (Without this PR, every time my displays go to sleep on the amdgpu, all the windows are just gathered, centered, on the first display to wake up.)
3. You can do things like adjust the scale of one display (*e.g.*, from 1 to 2) and have windows that were pushed off the viewable area gathered somewhere that they will be visible. Later, if you restore the scale, the relocated windows will go back to their original placement.

Of course, all of these automatic restorations happen only on views that have not since been relocated by the user.

In addition to the desktop that triggers disconnects every time the display wakes, I've been testing this behavior on a laptop with a couple of external displays, and the relocation and restoration logic seems intuitive over all of the layout changes that I can image.

johanmalm:
LGTM. Looks like all the queries have been dealt with. Will go for the merge.

johanmalm:
Thanks for all the work on this.

PR #1398, opened 2024-01-02
feat: Add initial tablet pad button mappings

jp7677:
This PR wires the tablet pad and lets you map tablet pad buttons to mouse buttons. This is probably not the most important feature but a good stepping stone to bind actions to pad and stylus buttons, like the  mouse-bindings.
 @Consolatis did also some work in his branch here https://github.com/labwc/labwc/issues/1060#issuecomment-1786384495 for that. That said, i guess adding mouse-like bindings needs some thinking up front, so I prefer to just open an issue for that after this PR to have a place to collect all thoughts.

I haven't found any `const`'s for the  pad buttons in wlroots. Please let know if you prefer to define our own `const`'s instead of using directly `0x0` etc.

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -404,8 +404,8 @@
     after applying tablet area transformation.
 
     Tablet buttons emulate regular mouse buttons. The tablet *button* can
-    be set to any of [tip|stylus|stylus2|stylus3]. Valid *to* mouse buttons
-    are [left|right|middle].
+    be set to any of [tip|stylus|stylus2|stylus3|pad|pad2|pad3|..|pad9].
+    Valid *to* mouse buttons are [left|right|middle].
   -->
   <tablet rotate="0">
     <!-- Active area dimensions are in mm -->
--- a/include/input/tablet.h
+++ b/include/input/tablet.h
@@ -20,6 +20,7 @@ struct drawing_tablet {
 	} handlers;
 };
 
-void tablet_setup_handlers(struct seat *seat, struct wlr_input_device *wlr_input_device);
+uint32_t tablet_get_mapped_button(uint32_t src_button);
+void tablet_init(struct seat *seat, struct wlr_input_device *wlr_input_device);
 
 #endif /* LABWC_TABLET_H */
--- a/include/input/tablet_pad.h
+++ b/include/input/tablet_pad.h
@@ -1,0 +1,21 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_TABLET_PAD_H
+#define LABWC_TABLET_PAD_H
+
+#include <wayland-server-core.h>
+struct seat;
+struct wlr_device;
+struct wlr_input_device;
+
+struct drawing_tablet_pad {
+	struct seat *seat;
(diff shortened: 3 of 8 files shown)


Consolatis:
> I haven't found any `const`'s for the pad buttons in wlroots. Please let know if you prefer to define our own `const`'s instead of using directly `0x0` etc.

`/usr/include/linux/input-event-codes.h`

> The CI failure looks like a network issue, unrelated to the PR itself. I cannot trigger a retry unfortunately.

restarted

jp7677:
> `/usr/include/linux/input-event-codes.h`

ah, sry, that was exactly the header I was looking into (close to the `BTN_STYLUS` definitions), I'm realizing now this isn't a `wlroots` header ;). 
 
> restarted

Thanks!

Consolatis:
I see that you are defining your own numbers here, I don't think that will work.
Could you add some `wlr_log(WLR_ERROR, "pad button pressed: 0x%x", ev->button);` in `handle_button()` and post what buttons you get?

jp7677:
> I see that you are defining your own numbers here, I don't think that will work. Could you add some `wlr_log(WLR_ERROR, "pad button pressed: 0x%x", ev->button);` in `handle_button()` and post what buttons you get?

Those do not come from my twisted mind ;) Those are exactly the codes I get from my pad. I have four buttons on the pad (Wacom Intuos M, https://www.wacom.com/en-us/products/pen-tablets/wacom-intuos ), they return 0x0 till 0x3 when pressed. I can bind right mouse button just fine to all four buttons.

Consolatis:
> they return 0x0 till 0x3 when pressed. I can bind right mouse button just fine to all four buttons.

Uh, that is unfortunate. I have a slight feeling that this depends on the specific pad then. Sadly mine doesn't have any buttons at all so I can't really verify that.

jp7677:
> > they return 0x0 till 0x3 when pressed. I can bind right mouse button just fine to all four buttons.
> 
> Uh, that is unfortunate. I have a slight feeling that this depends on the specific pad then. Sadly mine doesn't have any buttons at all so I can't really verify that.

Yeah, I had the same thought. Best case is that all buttons start at 0x0 and count upwards. But also not sure, I have just this one tablet. An alternative approach would be to not use fixed "pad[|2|3|4]" names but just strip "pad" from the string and somehow convert the remaining number to the corresponding hex value.

Consolatis:
Yeah, I think we can just define them in `include/input/tablet.h` like `#define LAB_BTN_PAD 0x0` and so on for now and deal with user feedback for not working pad button mappings when the issue arises.

johanmalm:
Ref: https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#tablet-pad-button-numbers

jp7677:
> Ref: https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#tablet-pad-button-numbers

ah thanks. How far should I go?  From `LAB_BTN_PAD` to `LAB_BTN_PAD9`?

Consolatis:
LGTM, obviously can't test.

johanmalm:
Thanks

Issue #1406, opened 2024-01-04
System vs user configuration


spl237:
Has any consideration been given to system-wide vs user-specific configuration of labwc?

On Raspberry Pi Desktop, I set up most components of the desktop with default configuration options which are installed as part of the packages, or as part of a global configuration package. The user can then apply further customisation, but it means that every user account which is created starts out with the same initial setup.

(It is not possible to install files into the user's directory as part of the installation of a Debian package, so anything which can only be configured with files in the user's directory cannot easily be preconfigured on install.)

For applications which conform to the XDG spec, this is usually achieved by a default system-wide configuration file or files stored in a subdirectory of /etc/xdg, which loads by default, and can then be overridden by files in a subdirectory of ~/.config.

The rather more hack-y alternative is when an application runs, it checks to see if it can find any configuration files in ~/.config, and if not, it copies a default set from somewhere in /etc into ~/.config and then uses them.

I can't find any reference to system vs user configuration of this kind in the documentation, so I assume it isn't currently implemented. (Or is it?) I'm going to need something like this, so am happy to have a look at writing it - would that be something which would be accepted upstream, or is it something I would need to keep for a local version?

ahesford:
This sounds like an x-y problem. Why would you expect all users in a multi-user Raspberry Pi to want the same configuration?

spl237:
As one example, I would want them all to have the same set of keyboard bindings.

ahesford:
You can always add the config to `/etc/skel` and let `useradd` populate the home directory when the user account is created.

Consolatis:
We are doing something similar with the theme + `~/.config/labwc/themerc-override` already.

Personally, I have no preference either way regarding accepting something similar for the config itself. CC @johanmalm

ahesford:
Merging configurations like this is a bad idea. It makes configurations non-obvious and difficult to model mentally. What's worse, it dramatically increases the likelihood that administrative actions (modifying a system configuration) can break established user workflows. This would be extremely frustrating to encounter and even more frustrating to debug, because users will never suspect changes in a config file that is magically included and not under their control.

There may be good arguments for a locally provided default configuration. There may also be good arguments for supporting an `<include>` directive to allow users to opt into relying on the local default while overriding or adding to it. Nevertheless, unless explicitly requested, user configurations should always supersede, not augment, local defaults.

spl237:
[...]

PR #1411, opened 2024-01-05
Load hierarchy of rc.xml files from directories

spl237:
https://github.com/labwc/labwc/issues/1406

Following this discussion yesterday, here is a quick-and-dirty approach to loading multiple rc.xml files which seems to work.

I scan through the list of config directories in reverse order, global system directories first followed by user directories. (If a configuration file is specifically provided on the command line, that file is the only one loaded.) The XML in each is parsed in turn, and values defined in successive files override those in earlier files, while values which are not defined in a newer file take the value from the older file.

This does not do anything about making this behaviour opt-in or opt-out, but it does exactly what I need - allow me to set a global set of defaults in /etc/xdg/labwc.xml and override specific values with ~/.config/labwc/rc.xml.

I'm not expecting this to be taken as-is, but it is, I hope, a useful first attempt at adding this functionality, which I need for my implementation. Some mechanism whereby this functionality is made optional is probably worthwhile, based on the discussion above.

### DIFF
--- a/include/common/dir.h
+++ b/include/common/dir.h
@@ -3,6 +3,7 @@
 #define LABWC_DIR_H
 
 char *config_dir(void);
+char *config_dir_n(int n);
 
 /**
  * theme_dir - find theme directory containing theme @theme_name
--- a/include/common/dir.h
+++ b/include/common/dir.h
@@ -3,7 +3,8 @@
 #define LABWC_DIR_H
 
 char *config_dir(void);
-char *config_dir_n(int n);
+char *user_config_dir(void);
+char *sys_config_dir(void);
 
 /**
  * theme_dir - find theme directory containing theme @theme_name
--- a/include/config/session.h
+++ b/include/config/session.h
@@ -4,17 +4,15 @@
 
 /**
  * session_environment_init - set enrivonment variables based on <key>=<value>
- * @dir: path to config directory
  * pairs in `${XDG_CONFIG_DIRS:-/etc/xdg}/lawbc/environment` with user override
  * in `${XDG_CONFIG_HOME:-$HOME/.config}`
  */
-void session_environment_init(const char *dir);
+void session_environment_init(void);
 
 /**
  * session_autostart_init - run autostart file as shell script
- * @dir: path to config directory
  * Note: Same as `sh ~/.config/labwc/autostart` (or equivalent XDG config dir)
  */
-void session_autostart_init(const char *dir);
+void session_autostart_init(void);
 
 #endif /* LABWC_SESSION_H */
--- a/src/common/dir.c
+++ b/src/common/dir.c
@@ -17,9 +17,13 @@ struct dir {
 	const char *path;
 };
 
-static struct dir config_dirs[] = {
+static struct dir user_config_dirs[] = {
 	{ "XDG_CONFIG_HOME", "labwc" },
 	{ "HOME", ".config/labwc" },
+	{ NULL, NULL }
+};
+
+static struct dir sys_config_dirs[] = {
 	{ "XDG_CONFIG_DIRS", "labwc" },
 	{ NULL, "/etc/xdg/labwc" },
 	{ NULL, NULL }
@@ -74,12 +78,12 @@ build_theme_path(struct ctx *ctx, char *prefix, const char *path)
 }
 
 static char *
-find_dir(struct ctx *ctx, int n)
+find_dir(struct ctx *ctx)
 {
 	char *debug = getenv("LABWC_DEBUG_DIR_CONFIG_AND_THEME");
 
 	for (int i = 0; ctx->dirs[i].path; i++) {
-		struct dir d = ctx->dirs[n != -1 ? n : i];
+		struct dir d = ctx->dirs[i];
 		if (!d.prefix) {
 			/* handle /etc/xdg... */
 			ctx->build_path_fn(ctx, NULL, d.path);
@@ -109,9 +113,6 @@ find_dir(struct ctx *ctx, int n)
 			}
 			g_strfreev(prefixes);
 		}
-		if (n != -1) {
-			break;
-		}
 	}
 	/* no directory was found */
 	ctx->buf[0] = '\0';
@@ -119,7 +120,7 @@ find_dir(struct ctx *ctx, int n)
 }
 
 char *
-config_dir(void)
+user_config_dir(void)
 {
 	static char buf[4096] = { 0 };
 	if (buf[0] != '\0') {
@@ -129,26 +130,35 @@ config_dir(void)
 		.build_path_fn = build_config_path,
 		.buf = buf,
 		.len = sizeof(buf),
-		.dirs = config_dirs
+		.dirs = user_config_dirs
 	};
-	return find_dir(&ctx, -1);
+	return find_dir(&ctx);
 }
 
 char *
-config_dir_n(int n)
+sys_config_dir(void)
 {
-	char buf[4096] = { 0 };
+	static char buf[4096] = { 0 };
+	if (buf[0] != '\0') {
+		return buf;
+	}
 	struct ctx ctx = {
 		.build_path_fn = build_config_path,
 		.buf = buf,
 		.len = sizeof(buf),
-		.dirs = config_dirs
+		.dirs = sys_config_dirs
 	};
-	if (n == 0) {
-		return find_dir(&ctx, getenv("XDG_CONFIG_HOME") ? 0 : 1);
-	} else {
-		return find_dir(&ctx, getenv("XDG_CONFIG_DIRS") ? 2 : 3);
+	return find_dir(&ctx);
+}
+
+char *
+config_dir(void)
+{
+	char *res = user_config_dir ();
+	if (res[0] != '\0') {
+		return res;
 	}
+	return sys_config_dir ();
 }
 
 char *
@@ -162,5 +172,5 @@ theme_dir(const char *theme_name)
 		.dirs = theme_dirs,
 		.theme_name = theme_name
 	};
-	return find_dir(&ctx, -1);
+	return find_dir(&ctx);
 }
--- a/src/common/dir.c
+++ b/src/common/dir.c
@@ -74,12 +74,12 @@ build_theme_path(struct ctx *ctx, char *prefix, const char *path)
 }
 
 static char *
-find_dir(struct ctx *ctx)
+find_dir(struct ctx *ctx, int n)
 {
 	char *debug = getenv("LABWC_DEBUG_DIR_CONFIG_AND_THEME");
 
 	for (int i = 0; ctx->dirs[i].path; i++) {
-		struct dir d = ctx->dirs[i];
+		struct dir d = ctx->dirs[n != -1 ? n : i];
 		if (!d.prefix) {
 			/* handle /etc/xdg... */
 			ctx->build_path_fn(ctx, NULL, d.path);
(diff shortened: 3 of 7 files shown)


spl237:
Thinking about it, this is actually already opt-out behaviour - if a config directory or config file is explicitly specified on the labwc command line, then the hierarchical load is skipped and only the specified rc.xml will be loaded.

ahesford:
This does not respect the intention that `$HOME/.config/labwc` is a fallback for cases where `$XDG_CONFIG_HOME` is undefined, or that `/etc/xdg` is a fallback for cases where `$XDG_CONFIG_DIRS` is undefined. These fallbacks should not be used at all when their respective variables are defined.

In addition, `$XDG_CONFIG_DIRS` can contain an arbitrary number of colon-separated directories, but overloading `find_dir` in this fashion will only select the first such directory that exists and contains a file. If configs are going to be merged, it makes sense to merge all files in multiple `$XDG_CONFIG_DIRS`.

I think a  cleaner option would be to keep the single-file autodetect behavior, but allow multiple `-c` arguments on the command line to merge any specified configuration by simply loading each file in the order provided. Raspberry Pi Desktop could ship a `labwc-session` shell script that discovers which of `${XDG_CONFIG_DIRS:-/etc/xdg}/labwc/rc.xml` and `${XDG_CONFIG_HOME:-${HOME}/.config}/labwc.xml` exist and just launch `labwc` with the appropriate `-c` flags.

spl237:
I've modified the PR so that the merging of rc.xml files is now opt-in only; if you want to merge configs, you need to supply the -m command line option to labwc; if you don't supply that option, you will get the original behaviour. Also, using either the -c or the -C option will disable merging anyway, as they explicitly set a config path. So the behaviour should be unchanged for anyone who doesn't want to merge configs - configs will only be merged if expressly requested.

spl237:
> This does not respect the intention that `$HOME/.config/labwc` is a fallback for cases where `$XDG_CONFIG_HOME` is undefined, or that `/etc/xdg` is a fallback for cases where `$XDG_CONFIG_DIRS` is undefined. These fallbacks should not be used at all when their respective variables are defined.

Now fixed.

> In addition, `$XDG_CONFIG_DIRS` can contain an arbitrary number of colon-separated directories, but overloading `find_dir` in this fashion will only select the first such directory that exists and contains a file. If configs are going to be merged, it makes sense to merge all files in multiple `$XDG_CONFIG_DIRS`.

But labwc wasn't handling that case anyway; it was only loading rc.xml from the first entry in XDG_CONFIG_DIRS, wasn't it? Given the additional complexity involved in doing that, and the fact that RPiOS never has more than one entry in that variable, I'm happy to leave that for future implementation.

> I think a cleaner option would be to keep the single-file autodetect behavior, but allow multiple `-c` arguments on the command line to merge any specified configuration by simply loading each file in the order provided. Raspberry Pi Desktop could ship a `labwc-session` shell script that discovers which of `${XDG_CONFIG_DIRS:-/etc/xdg}/labwc/rc.xml` and `${XDG_CONFIG_HOME:-${HOME}/.config}/labwc.xml` exist and just launch `labwc` with the appropriate `-c` flags.

I disagree. It is far better to have clearly-defined merge behaviour which is correctly implemented - I cannot think of any circumstance in which you would want system config to override user config - than to have merge behaviour which is dependent on correctly entering the order of identically-flagged command-line parameters, not least because it is not intuitive as to what the order of precedence would be.

spl237:
> The current implementation correctly splits `XDG_CONFIG_DIRS` and will select the first component that contains a matching file.

In which case, this will do exactly the same - it will take the first component of XDG_CONFIG_DIRS which contains a matching file as the file to merge, which is reasonable behaviour to expect. Merging multiple system defaults is not, in my opinion, either reasonable or desirable.

ahesford:
All of the index tricks to achieve the desired outcome are unwieldy.

I suggest you split the current `config_dirs` list into two: `usr_config_dirs` and `sys_config_dirs`. Then, the `find_dirs` function can be restored to its original implementation, and be used with each list separately to achieve your desired outcome: returning the first valid path from the given list. Then, the `config_dir` function can be split into two new functions, `config_usr_dir` and `config_sys_dir`, that use `find_dir` to pick from the appropriate category. The existing `config_dir` can now simply call `config_usr_dir` first and, if that function fails to find a match, call `config_sys_dir` for a fallback.

In `rcxml_read`, the loop conceals the intent of the merge. Instead, it would be more clear  to move the core XML parsing to a dedicated parsing function (`rcxml_read_file`) which embodies the interior of your loop without the file-finding logic. The `rcxml_read` function can then handle the file name caching and do the basic merge flow:
1. Initialize the configuration structure.
2. Cache any provided filename.
3. If the cache contains no filename (*i.e.*, you are merging configs), use `config_sys_dir` to find the system path, then call `rcxml_read_file` on a configuration file in the path (if identified).
4. Call `rcxml_read_file` on the user configuration. If no merging was specified, this will just be the value of the cached filename. Otherwise, the cache will be empty, so you should use `config_usr_dir` to find the user path and build the path there.
5. Finalize as the function currently does.

spl237:
> I suggest you split the current `config_dirs` list into two: `usr_config_dirs` and `sys_config_dirs`. Then, the `find_dirs` function can be restored to its original implementation, and be used with each list separately to achieve your desired outcome: returning the first valid path from the given list. Then, the `config_dir` function can be split into two new functions, `config_usr_dir` and `config_sys_dir`, that use `find_dir` to pick from the appropriate category. The existing `config_dir` can now simply call `config_usr_dir` first and, if that function fails to find a match, call `config_sys_dir` for a fallback.

I'll consider that. The current implementation works perfectly well, and I'm not sure what changing it achieves other than perhaps slightly better readability.

> In `rcxml_read`, the loop conceals the intent of the merge. Instead, it would be more clear to move the core XML parsing to a dedicated parsing function (`rcxml_read_file`) which embodies the interior of your loop without the file-finding logic. 

I agree that would be a tidier solution - had already considered doing it like that, but had wanted to minimise code changes. I will modify accordingly.

johanmalm:
Thanks for working on this. I'm happy to roll my sleeves up too if needed (do just say), but might not work to "required" time-scales :smile:

Have had a quick look at code. My 2ps worth:

Consider the following for cleanliness, readability and extensibility (for `environment`, `menu.xml`, etc):

Create a function (in `common/dir.c` I guess) which takes filename (e.g. "rc.xml") and `bool should_merge`. It would return a `wl_list` of the files (e.g. `/etc/xdg/labwc/rc.xml`, `/home/me/.config/labwc/rc.xml`). For should_merge=false it iterates forwards (most important base dir first) and returns only one file. For should_merge=true it iterates backwards and returns all positive `stat()`s. I think that would simplify everything else, for example the `rcxml.c` parsing section then becomes cleaner:

1. Start with `config_file_list_create()` (or whatever)
2. Iterates over files and parse
3. Free the list
4. Call `post_processing()`

I know this is just proof-of-concept... but a couple of nits :smile:

- I prefer `merge-config` to `mergeconfig` as the command line option (as `sway` does it).
- We prefer `strdup_printf()` to `g_strdup_printf()`... and need associated `free()`. `strdup_printf` is just based on the example in `printf(3)`, but avoids mixing glib allocation which has some problems. Ref below.

https://github.com/labwc/labwc/blob/785a34e8ad6e693591cbf657b4b8ff50192d0eec/src/common/string-helpers.c#L53

ahesford:
> I'll consider that. The current implementation works perfectly well, and I'm not sure what changing it achieves other than perhaps slightly better readability.

Your current implementation requires knowledge of a specific ordering of the variables in the configuration directory. Should that order change (or elements added), your logic will break in subtle ways that can be frustrating to debug. In addition to making the intent clearer, my suggested change will be more robust to future modifications in this file.

johanmalm:
Are we doing this for `rc.xml`, `menu.xml`, `autostart`, `environment` and `themerc-override`? If it's for `rc.xml` only for now, I'm keen that we still write it in such a way that it does not require a re-write for the other later if there is a use-case.

Also, just to share my thoughts whilst we're in the process of refactoring/adding - the use of `static char buf[4096]` in `config_dir()` isn't great. It's my doing (sorry), and I suppose I felt that I knew what I was doing, but it can make for hard to find bugs if people rely on it not changing. It'd be better to return a heap allocated buffer (or list as discussed above).

spl237:
> Are we doing this for `rc.xml`, `menu.xml`, `autostart`, `environment` and `themerc-override`? If it's for `rc.xml` only for now, I'm keen that we still write it in such a way that it does not require a re-write for the other later if there is a use-case.

The one it is definitely needed for is rc.xml, as that contains settings that are likely to require configuration by both distributor and user.

I'd have said it wasn't necessary for themerc-override, as that seems to be a file that only needs to be changed by the user; global changes by the distributor can just be made directly to the theme itself.

Similarly, I can't see it being that likely that a user will need to override a distributor setting in menu.xml (but I could be wrong!)

autostart and environment are the other two which are most likely to require both user and distributor changes, although I have worked around that for the time being - distributor environment can be set in a wrapper script around the labwc executable, and distributor autostarts can be added as a script used as the -s argument to the executable. But it would be tidier if they could both be done with overrides.

For now, I need it to work on rc.xml, not the others, but at some point in the future it could be extended.

One other thing that might be worth considering that doesn't seem to happen at the moment is falling back on a per-file basis in the directory hierarchy, even if merging isn't used. So at present if a user config directory is defined and there is no autostart file in that directory, it might be useful for an autostart file from /etc/xdg to be loaded instead. If merging of certain files turns out to be less easy than it is with rc.xml, that would be a useful alternative to implement.

spl237:
> > I'll consider that. The current implementation works perfectly well, and I'm not sure what changing it achieves other than perhaps slightly better readability.
> 
> Your current implementation requires knowledge of a specific ordering of the variables in the configuration directory. Should that order change (or elements added), your logic will break in subtle ways that can be frustrating to debug. In addition to making the intent clearer, my suggested change will be more robust to future modifications in this file.

OK, that makes sense.

johanmalm:
> One other thing that might be worth considering that doesn't seem to happen at the moment is falling back on a per-file basis in the directory hierarchy, even if merging isn't used. So at present if a user config directory is defined and there is no autostart file in that directory, it might be useful for an autostart file from /etc/xdg to be loaded instead. If merging of certain files turns out to be less easy than it is with rc.xml, that would be a useful alternative to implement.

Yes. Good point. I'm not sure what's best here, but have thought about it before. We currently just lock onto one directory (as you know) but I suspect it would make more sense to fallback on a per-file basis.

spl237:
OK, I've done some tidying and refactoring, mostly along the lines of Andrew's suggestions; it's a lot more obvious what is going on now, I think!

I've also addressed Johan's specific points, with the exception of the static buffer in the config_dir function; due to the len parameter in the context passed to the find_dir function, it's not obvious how to get rid of the requirement for a fixed-size buffer without some significant rewriting, and if the buffer is fixed size anyway, there's not much point mallocing it as opposed to just having it as static locally.

spl237:
> > One other thing that might be worth considering that doesn't seem to happen at the moment is falling back on a per-file basis in the directory hierarchy, even if merging isn't used. So at present if a user config directory is defined and there is no autostart file in that directory, it might be useful for an autostart file from /etc/xdg to be loaded instead. If merging of certain files turns out to be less easy than it is with rc.xml, that would be a useful alternative to implement.
> 
> Yes. Good point. I'm not sure what's best here, but have thought about it before. We currently just lock onto one directory (as you know) but I suspect it would make more sense to fallback on a per-file basis.

OK, I've added a change to enable this behaviour. For both environment and autostart, if there is a file in ~/.config/labwc/, it will be used; if not, a file in /etc/xdg/labwc will be used. So this is more fallback behaviour than merge behaviour - if merge behaviour is desired instead in these cases, then that should be reasonably easy to implement - just process the environment file in both cases with the user file overriding any common settings in the system file, and execute both autostarts in an order TBD. I can see arguments for both merge and fallback behaviour here - not sure which people think is more useful.

Issue #1404, opened 2024-01-03
tablet: Add pen/pad button bindings


jp7677:
Follow up from https://github.com/labwc/labwc/issues/1060 and https://github.com/labwc/labwc/pull/1398

We can map pen and pad buttons to mouse buttons. It would be cool to also bind actions to them. Similar like mouse bindings, but for tablet buttons. The branch mentioned here https://github.com/labwc/labwc/issues/1060#issuecomment-1786384495 already contains idea's for that.

(Tablet support for me is very much complete, so this is rather a "nice to have" feature)

jp7677:
Based on above mentioned branch, this is probably the easiest way to allow pen/pad buttons in regular mouse bindings:

<details>

  <summary>Patch</summary>

````diff
diff --git a/src/config/mousebind.c b/src/config/mousebind.c
index c578c46..8e43a2d 100644
--- a/src/config/mousebind.c
+++ b/src/config/mousebind.c
@@ -8,6 +8,7 @@
 #include "common/list.h"
 #include "common/mem.h"
 #include "config/mousebind.h"
+#include "config/tablet.h"
 #include "config/rcxml.h"
 
 uint32_t
@@ -35,6 +36,11 @@ mousebind_button_from_str(const char *str, uint32_t *modifiers)
        } else if (!strcasecmp(str, "Middle")) {
                return BTN_MIDDLE;
        }
+
+       uint32_t tablet_btn = tablet_button_from_str(str);
+       if (tablet_btn != UINT32_MAX) {
+               return tablet_btn;
+       }
 invalid:
        wlr_log(WLR_ERROR, "unknown button (%s)", str);
        return UINT32_MAX;
diff --git a/src/input/tablet.c b/src/input/tablet.c
index 0a24d61..b4d7bea 100644
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -97,7 +97,7 @@ tablet_get_mapped_button(uint32_t src_button)
                }
        }
        wlr_log(WLR_DEBUG, "no button map target for 0x%x", src_button);
-       return 0;
+       return src_button;
 }
 
 static void
diff --git a/src/input/tablet_pad.c b/src/input/tablet_pad.c
index 3c1b2a9..895574b 100644
--- a/src/input/tablet_pad.c
+++ b/src/input/tablet_pad.c
@@ -18,10 +18,6 @@ handle_button(struct wl_listener *listener, void *data)
        struct wlr_tablet_pad_button_event *ev = data;
[...]

PR #1414, opened 2024-01-06
feat: add more mouse button definitions

jp7677:
This allows those other mouse buttons (from https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L355 ) to be bound and also makes more buttons available to map to from tablet pen/pad buttons.
(Since I have only 4 buttons on my pad, this absolutely sufficient for me)

Also drop one method in favor of reuse.

Closes https://github.com/labwc/labwc/issues/1404 if we want to restrict tablets to pure mouse emulations (which i think is just fine).

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -407,6 +407,11 @@ windows using the mouse.
 	- Left
 	- Middle
 	- Right
+	- Side
+	- Extra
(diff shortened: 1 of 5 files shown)


johanmalm:
Nice. Thanks.

Issue #1399, opened 2024-01-02
Applications overlap bottom-placed-status-bars (waybar, sfwbar) at launch.  However, they don't overlap top-placed status bars.


orchardstreet:
This can be annoying for those who prefer bottom-placed status bars.  Status bars are most traditionally at the bottom.  Sfwbar is at bottom by default for example.  So I think it'd be nice if labwc prevented overlapping bottom status bars on application launch, like it currently does if a status bar is at the top.

I was told via IRC that labwc may only be checking not to overlap status bars on the left and top because of the "'titlebar issue".  Indeed, programs don't overlap status bars that are at the top of the screen, unlike the bottom.

Thanks :)

johanmalm:
IIRC we just move to make sure top/left in not outside.
I guess it's only a problem if you have the panel in the "bottom" layer... and at the "bottom" of the screen.

Try:

```
function("SfwbarInit") {
  SetBarId "bar-0"
  SetLayer "top"
}

window {
  -GtkWindow-direction: bottom;
}
```

I suppose we could resize windows on launch to always fit within the output... I think I would quite like that actually regardless of the panel issue.

PR #1415, opened 2024-01-06
Fit within usable area

johanmalm:
In response to https://github.com/labwc/labwc/issues/1399

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -399,6 +399,7 @@ void view_center(struct view *view, const struct wlr_box *ref);
  * @view: view to be placed
  */
 void view_place_initial(struct view *view);
+void view_constrain_size_to_that_of_usable_area(struct view *view);
 
 void view_restore_to(struct view *view, struct wlr_box geometry);
 void view_set_untiled(struct view *view);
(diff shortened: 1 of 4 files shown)


ahesford (approved):
This looks like a reasonable approach to solving the linked issue.

I wonder if triggering a resize before placement will cause a "bounce" where the window is drawn in some initial position before the placement algorithm is applied, and then moved to its final location. However, if there is an apparent jump, I don't think it can be avoided without completely refactoring `view_place_initial` and all of its supporting functions to use a pre-determined width and height apart from whatever is stored in `view->pending`.

johanmalm:
> I wonder if triggering a resize before placement will cause a "bounce" where the window is drawn in some initial position before the placement algorithm is applied, and then moved to its final location. However, if there is an apparent jump, I don't think it can be avoided without completely refactoring `view_place_initial` and all of its supporting functions to use a pre-determined width and height apart from whatever is stored in `view->pending`.

Yes, possibly, although it's before we connect to the commit-signal, so guess it should be fine.

Have tested running nested and seems fine.

Shall we treat this as a sensible first step and consider further refactoring it leads to bounces.

ahesford:
Just did a quick test and confirm your finding---the window starts where I would expect it to start and I don't see it moved. I agree that this should be merged as-is and doubt any issues will come up later.

Issue #1337, opened 2023-12-18
My (multi layer) xkb layout broken on current master


datMaffin:
While investigating another task on a recent master, I noticed that my keyboard layout does not behave as intended on the current labwc master.

The behavior is as if the level5 "lsgt_switch" (German Keyboard "<" key) is pressed down by default. If I press it, it is as if level5 is unpressed. So the level5(lsgt_switch) seems currently be "inverted".

Setup:

```nix
  # Configure xkb keymap
  services.xserver.extraLayouts = {
    de-lsgtmod = {
      description = "German (less-than as modifier)";
      languages = [ "de" ];
      # Best documentation seems to be: https://www.charvolant.org/doug/xkb/html/node5.html#SECTION00053000000000000000
      # Also, look at similar keyboard configurations.
      #
      # GNOME wayland is currently bugged. For layers >= 5 to work, it is
      # necessary to have the layout in the **first position**.
      # See also: https://gitlab.gnome.org/GNOME/mutter/-/issues/790
      symbolsFile = pkgs.writeText "de-lsgtmod" ''
        partial alphanumeric_keys modifier_keys keypad_keys
        xkb_symbols "de-lsgtmod" {
            include "latin(type4)"
            name[Group1]="German (less-than as modifier)";
            key.type[Group1] = "EIGHT_LEVEL_WITH_LEVEL5_LOCK";
            key <TLDE>  { [dead_circumflex, degree,	U2032,    U2033	] };
            key <AE02>  { [         2,   quotedbl,  twosuperior,    oneeighth ]	};
            key <AE03>  { [         3,    section, threesuperior,    sterling ]	};
            key <AE04>  { [         4,     dollar,   onequarter,     currency ]	};
            key <AE07>  { [         7,      slash,    braceleft, seveneighths, backslash ]	};
            key <AE08>  { [         8,  parenleft,  bracketleft,    trademark, bracketleft ]	};
            key <AE09>  { [         9, parenright, bracketright,    plusminus, bracketright ]	};
            key <AE11>  {[ssharp, question, backslash, questiondown, U1E9E ], type[group1]="FOUR_LEVEL_PLUS_LOCK" };
            key <AE12>  { [dead_acute, dead_grave, dead_cedilla,  dead_ogonek ]	};
            key.type[Group1] = "EIGHT_LEVEL_ALPHABETIC_WITH_LEVEL5_LOCK";
            key <AD03>  { [         e,          E,     EuroSign,     EuroSign ]	};
[...]

PR #1416, opened 2024-01-06
Add click method libinput option

### LINKED ISSUE #957: Touchpad two finger right click not working
I've set my rc.xml to use the following libinput config but the touchpad doesn't use two finger rightclick. Based on your current config, you are missing setting libinput click methods to clickfinger for the tapButtonMap -> lrm to work 
 

rc.xml
```xml
<libinput>
    <device category="SynPS/2 Synaptics TouchPad">
      <naturalScroll>no</naturalScroll>
      <leftHanded>no</leftHanded>
      <accelProfile>adaptive</accelProfile>
      <tap>yes</tap>
      <tapButtonMap>lrm</tapButtonMap>
      <middleEmulation>no</middleEmulation>
      <disableWhileTyping>yes</disableWhileTyping>
    </device>
</libinput>
```

Here is my libinput list-device output for my touchpad.
```
Device:           SynPS/2 Synaptics TouchPad
Kernel:           /dev/input/event14
Group:            7
Seat:             seat0, default
Size:             70x50mm
Capabilities:     pointer gesture
Tap-to-click:     disabled
Tap-and-drag:     enabled
Tap drag lock:    disabled
Left-handed:      disabled
Nat.scrolling:    disabled
Middle emulation: disabled
Calibration:      n/a
Scroll methods:   *two-finger edge
Click methods:    *button-areas clickfinger
Disable-w-typing: enabled
Disable-w-trackpointing: enabled
Accel profiles:   flat *adaptive custom
Rotation:         n/a
```

PS - Love all the work everyone has put in this. this is my daily driver and would love to see 2 issues -> this one and other 'Show desktop' fixed.

datMaffin:
Resolves #957 

For this PR to be ready to merge, I still need to 

- [x] change the documentation and 
- [x] the example rc.xml.

I did try it out, and it does work. But sadly I can not daily drive labwc master because of #1337

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -581,10 +581,18 @@ windows using the mouse.
 	after as well.
 
 *<libinput><device><clickMethod>* [none|buttonAreas|clickfinger]
-	Set the kind of click method that shall be used. Either the location of
-	where the click happens (buttonAreas), or the number of fingers
-	(clickfinger) being used determines the kind of mouse button being
-	triggered. Choose *none* if a click shall not trigger a mouse button click.
+	Configure the method by which physical clicks on a touchpad are mapped to
+	mouse-button events. The click methods available are:
+	- *buttonAreas* - The bottom of the touchpad is divided into distinct
+	  regions corresponding to left, middle and right buttons; clicking within
+	  the region will trigger the corresponding event. Clicking the main area
+	  further up produces a left button event.
+	- *clickfinger* - Clicking with one, two or three finger(s) will produce
+	  left, right or middle button event without regard to the location of a
+	  click.
+	- *none* - Physical clicks will not produce button events.
+
+	The default method depends on the touchpad hardware.
 
 ## WINDOW RULES
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -582,7 +582,9 @@ windows using the mouse.
 
 *<libinput><device><clickMethod>* [none|buttonAreas|clickfinger]
 	Configure the method by which physical clicks on a touchpad are mapped to
-	mouse-button events. The click methods available are:
+	mouse-button events.
+
+	The click methods available are:
 	- *buttonAreas* - The bottom of the touchpad is divided into distinct
 	  regions corresponding to left, middle and right buttons; clicking within
 	  the region will trigger the corresponding event. Clicking the main area
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -504,6 +504,7 @@ windows using the mouse.
     <dragLock></dragLock>
     <middleEmulation></middleEmulation>
     <disableWhileTyping></disableWhileTyping>
+    <clickMethod></clickMethod>
   </device>
 </libinput>
 ```
@@ -578,6 +579,12 @@ windows using the mouse.
 	Enable or disable disable while typing for this category. DWT ignores
 	any motion events while a keyboard is typing, and for a short while
 	after as well.
+
+*<libinput><device><clickMethod>* [none|buttonAreas|clickfinger]
+	Set the kind of click method that shall be used. Either the location of
+	where the click happens (buttonAreas), or the number of fingers
+	(clickfinger) being used determines the kind of mouse button being
+	triggered. Choose *none* if a click shall not trigger a mouse button click.
 
 ## WINDOW RULES
 
--- a/src/config/libinput.c
+++ b/src/config/libinput.c
@@ -22,7 +22,7 @@ libinput_category_init(struct libinput_category *l)
 	l->accel_profile = -1;
 	l->middle_emu = -1;
 	l->dwt = -1;
-	l->click_method = LIBINPUT_CONFIG_CLICK_METHOD_BUTTON_AREAS;
+	l->click_method = -1;
 }
 
 enum lab_libinput_device_type
--- a/src/config/libinput.c
+++ b/src/config/libinput.c
@@ -22,6 +22,7 @@ libinput_category_init(struct libinput_category *l)
 	l->accel_profile = -1;
 	l->middle_emu = -1;
 	l->dwt = -1;
+	l->click_method = LIBINPUT_CONFIG_CLICK_METHOD_BUTTON_AREAS;
 }
 
 enum lab_libinput_device_type
--- a/src/seat.c
+++ b/src/seat.c
@@ -15,6 +15,7 @@
 #include "input/keyboard.h"
 #include "input/key-state.h"
 #include "labwc.h"
+#include "libinput.h"
 #include "view.h"
 
 static void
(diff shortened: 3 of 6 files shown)


src/seat.c
```diff
@@ -15,6 +15,7 @@
 #include "input/keyboard.h"
 #include "input/key-state.h"
 #include "labwc.h"
+#include "libinput.h"
```
datMaffin, line 18: my clangd must have added this...

-[ ] Remove.

johanmalm:
Looks fine.

Consider strcasecmp for all.

Assume you’ve got an eye on updating docs/labwc-config.5.scd and rc.xml.all too.

datMaffin:
> Consider strcasecmp for all.

I used strcmp to be consitent with the "tapButtonMap" option:

```c
	} else if (!strcasecmp(nodename, "tapButtonMap")) {
		if (!strcmp(content, "lrm")) {
			current_libinput_category->tap_button_map =
				LIBINPUT_CONFIG_TAP_MAP_LRM;
		} else if (!strcmp(content, "lmr")) {
			current_libinput_category->tap_button_map =
				LIBINPUT_CONFIG_TAP_MAP_LMR;
		} else {
			wlr_log(WLR_ERROR, "invalid tapButtonMap");
		}
	} else if (!strcasecmp(nodename, "tapAndDrag")) {

```

But I will quickly change it. (Easy to revert again in any case)

src/config/libinput.c
```diff
@@ -22,6 +22,7 @@ libinput_category_init(struct libinput_category *l)
 	l->accel_profile = -1;
 	l->middle_emu = -1;
 	l->dwt = -1;
+	l->click_method = LIBINPUT_CONFIG_CLICK_METHOD_BUTTON_AREAS;
```
johanmalm, line 25: We should set this to `-1` by default so that we don't configure devices unless the requested.

src/seat.c
```diff
@@ -189,6 +189,12 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		wlr_log(WLR_INFO, "dwt configured");
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
+	if (libinput_device_config_click_set_method(libinput_dev,
```
johanmalm, line 192: Do a check here to avoid configuring when `dc->click_method < 0`

johanmalm (commented):
Looks good other than the two comments.

johanmalm:
Ref: https://github.com/jiixyj/libinput/blob/5bb9e3b3905a4d03166dd187aa853c87e38288e4/src/libinput.h#L4557-L4572

docs/labwc-config.5.scd
```diff
@@ -579,6 +580,12 @@ windows using the mouse.
 	any motion events while a keyboard is typing, and for a short while
 	after as well.
 
+*<libinput><device><clickMethod>* [none|buttonAreas|clickfinger]
+	Set the kind of click method that shall be used. Either the location of
+	where the click happens (buttonAreas), or the number of fingers
+	(clickfinger) being used determines the kind of mouse button being
+	triggered. Choose *none* if a click shall not trigger a mouse button click.
```
ahesford, line 587:
```suggestion
	Configure the method by which physical clicks on a touchpad are mapped
	to mouse-button events.

		- *buttonAreas* - The touchpad is divided into distinct regions
		corresponding to left, middle or right buttons; clicking within
		each region will trigger the corresponding event.

		- *clickfinger* - Clicking with one, two or three fingers will
		produce left, middle or right button events without regard to
		the location of a click. The specific map from finger count to
		button is infuenced by the *tapButtonMap* option.

		- *none* - Physical clicks will not produce button events.

	The default method depends on the touchpad hardware.
```

ahesford (changes requested):
In addition to expanding and clarifying documentation, I'm not sure about the use of the terms `clickfinger` and `buttonAreas`:
1. It bothers me a bit to see one camel-cased option and another all lowercase, so at a minimum I think either `clickfinger -> clickFinger` everywhere (which runs afoul of [libinput documentation](https://wayland.freedesktop.org/libinput/doc/latest/clickpad-softbuttons.html) that always uses the capitalized term "Clickfinger") or, perhaps better, `buttonAreas` -> `buttonareas` everywhere.
2. More descriptive nomenclature might be clearer even if it diverges from libinput documentation. Perhaps `byArea` and `byFingerCount` are better options for users.

src/config/libinput.c
Consolatis: This also implies not using `enum libinput_config_click_method` as data type.

docs/labwc-config.5.scd
datMaffin: I think the rational for the different names are that:
* "button areas" is a valid english term.
* "clickfinger" however, is a newly invented "name" and therefore a single word.

src/seat.c
datMaffin: basically done.

src/config/libinput.c
datMaffin: done.

src/seat.c
```diff
@@ -189,11 +189,21 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		wlr_log(WLR_INFO, "dwt configured");
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
-	if (libinput_device_config_click_set_method(libinput_dev,
-			dc->click_method) > 0) {
+	if (dc->click_method < 0
+			||
+			(libinput_device_config_click_get_methods(libinput_dev)
+				& dc->click_method) == 0) {
```
datMaffin, line 195: - [ ] should I turn this around? put the `< 0` at the end?

It makes the logic of available click methods a bit weird. I think -1 would be `0b11111...1111`. So the check with the flag would probably always succeed in that case.

docs/labwc-config.5.scd
datMaffin: I improved the documentation similar to what you proposed. However, I added more details about the behavior of the "buttonAreas" option.

And I removed the 
> The specific map from finger count to button is influenced by the *tapButtonMap* option.

This behavior is not documented in the libinput docs. Additionally, I tried it out, and it does not seem to be the case that the * tapButtonMap*  option influences Clickfinger.

I also changed the exact formatting to be more consisting with the already existing bullet point list.

datMaffin: - [ ] Based on the GitHub formatting of *.scd* I think something is still messed up with the formatting.

datMaffin: My change made it a bit better. But the GitHub rendering of this file seems to be very busted in any case.

src/seat.c
Consolatis: I think the whole if could be turned around, something like this (not tested at all)
```c
if (dc->click_method > 0
		&& libinput_device_config_click_get_methods(libinput_dev) & dc->click_method) {
	libinput_device_config_click_set_method(..);
} else {
	wlr_log(WLR_INFO, "click method not configured");
}
```

datMaffin: I just turned it around now. In my head this is most similar to the existing code: In the existing code, the checking whether an option is supported is already done in the same way;  -1 is also passed in that case.

johanmalm:
Okay, looks good, let's go for it.
I just pushed a minor change to change C99 // comment style for /* */

PR #1418, opened 2024-01-08
rcxml: avoid crash when checking environment for WLR_DRM_NO_ATOMIC

ahesford:
Also elevate warning about this variable to an error when tearing is configured, because presumably the user expects the tearing setting to work.

h/t: IRC user Domo


johanmalm:
Thanks for the quick response on this one. Can't believe I didn't spot that one :smile:

PR #1419, opened 2024-01-09
view, xwayland: fixes for new size constraints

ahesford:
Calling the new `view_constrain_size_to_that_of_usable_area` has some unexpected side-effects for multi-output configurations. In particular, setting `box.x = view->pending.x` and `box.y = view->pending.y` before any placement is selected most likely sets them each to zero. Then, calling `view_move_resize` seems to (eventually) trigger `view_discover_output`, which will override any `view->output` set in, *e.g.*, `xdg_toplevel_view_map` to be the output nearest the cursor. The end result is that, in `center` or `automatic` placement modes, the call to `view_constrain_size_to_that_of_usable_area` will force new windows to always appear on the display at (0, 0), regardless of where the cursor is.

To fix this, I've made a couple of changes to `view_constrain_size_to_that_of_usable_area`:
1. If the window is already small enough, just make this function a no-op. We don't need to move a window that will fit where it needs to be, and the placement algorithm will take care of the rest.
2. If the window *is* resized by this function, ensure that the `x` and `y` coordinates are chosen to place the window inside its target display, so that `view_discover_output` doesn't relocate it before the placement algorithm takes over.

While I was at it, I decided to call this function for Xwayland windows only when they don't have a client-requested geometry, because I figure something like `urxvt -geometry [...]` probably ought to override all of the intelligence and do exactly what the user asked.

cc: @johanmalm

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -737,15 +737,42 @@ view_constrain_size_to_that_of_usable_area(struct view *view)
 	if (!view || !view->output) {
 		return;
 	}
-	struct wlr_box *usable_area = &view->output->usable_area;
+
+	struct wlr_box usable_area =
+			output_usable_area_in_layout_coords(view->output);
 	struct border margin = ssd_get_margin(view->ssd);
+
+	int available_width = usable_area.width - margin.left - margin.right;
+	int available_height = usable_area.height - margin.top - margin.bottom;
+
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM

johanmalm:
Thanks!

Issue #1422, opened 2024-01-10
Touchscreen association with output device


spl237:
Under wayfire, it is possible to make configuration settings to bind touchscreens to specific output devices:

https://github.com/WayfireWM/wayfire/wiki/Configuration#input-device-specific-options

I can see the function new_touch in src/seat.c, which looks as if it does the requisite mapping, but I can't see how to set the name of the output device in the input device structure. Is it done as part of a libinput block? Or is this not yet supported?

johanmalm:
See #1417 - thanks to @jp7677

johanmalm:
@jp7677 Is this a discussion you want to be part of to help advise + shape.

Might be as simple as adding an optional name attribute.

If we need to create a list of config objects to match different can explain the preferred way of doing that. 

I can test this myself, so prefer if someone else takes the authoring+testing.

jp7677:
Thanks for asking, sure, I'm happy to help.

Yes, I've indeed never considered the situation that multiple drawing tablets might need different configurations. I guess this is not that common, but this situation might occur.

I'm not that familiar with touch devices. Are we really talking about touch devices without a monitor, like an Apple Magic Touchpad? For a single device, something like `<touch mapToOutput="DP-1" />` should be fine. When multiple such devices are connected (imho not a common situation), I also think that an identifier attribute in the `tablet` (for tablets) and `touch` element would work for the configuration. That said, I don't think `name` is sufficient since users might use exactly the same device twice, so `name` wouldn't be unique. Unfortunately I don't see something like a `serial` or `id` in `wlr_input_device`  :(

If we are talking about (multiple) touchscreens (thus touch is physically linked to an output), may be the current output mapping just fails because the output comes up later than the touch device (thus `touch->output_name` contains a not yet existing output name on startup?). The "wrong" startup order happened for my tablet, that's why I added `seat_output_layout_changed()`.  If that's the case, it is worth a try to extend the switch statement in that function https://github.com/labwc/labwc/blob/master/src/seat.c#L652 with something like this for touch devices: 

```c
case WLR_INPUT_DEVICE_TOUCH:
    struct wlr_touch *touch = wlr_touch_from_input_device(input->wlr_input_device);
    map_input_to_output(seat, input->wlr_input_device, touch->output_name);
```

Eventually that is sufficient to automatically re-map the touch device to its own output once the output comes up.

spl237:
> Thanks for asking, sure, I'm happy to help.
> 
[...]

PR #1426, opened 2024-01-12
feat: touch output mapping

jp7677:
From https://github.com/labwc/labwc/issues/1422

This adds a `<touch mapToOutput="" />` switch to the configuration, similar like the switch for `tablet`. This switch allows to override the linked output name from a touch device.

I don't have suitable hardware, so this needs validation from somebody else...
~Also no docs yet.~

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -447,6 +447,16 @@ windows using the mouse.
 	<mouse><mousebind> entries exist, the same default mousebinds will be
 	loaded even if the <default /> element is not provided.
 
+## TOUCH
+
+```
+<touch mapToOutput="" />
+```
+
+*<touch mapToOutput="" />*
+	Direct cursor movement to a specified output. If the compositor is
+	running in nested mode, this does not take effect.
+
 ## TABLET
 
 ```
--- a/src/seat.c
+++ b/src/seat.c
@@ -521,6 +521,7 @@ seat_reconfigure(struct server *server)
 			configure_libinput(input->wlr_input_device);
 			break;
 		case WLR_INPUT_DEVICE_TOUCH:
+			configure_libinput(input->wlr_input_device);
 			map_touch_to_output(seat, input->wlr_input_device);
 			break;
 		case WLR_INPUT_DEVICE_TABLET_TOOL:
--- a/src/seat.c
+++ b/src/seat.c
@@ -215,6 +215,14 @@ map_input_to_output(struct seat *seat, struct wlr_input_device *dev, char *outpu
 	wlr_cursor_map_input_to_region(seat->cursor, dev, NULL);
 }
 
+static void
+map_pointer_to_output(struct seat *seat, struct wlr_input_device *dev)
+{
+	struct wlr_pointer *pointer = wlr_pointer_from_input_device(dev);
+	wlr_log(WLR_INFO, "map pointer to output %s\n", pointer->output_name);
+	map_input_to_output(seat, dev, pointer->output_name);
+}
+
 static struct input *
 new_pointer(struct seat *seat, struct wlr_input_device *dev)
 {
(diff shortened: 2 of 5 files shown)


src/seat.c
```diff
@@ -515,6 +522,9 @@ seat_reconfigure(struct server *server)
 		case WLR_INPUT_DEVICE_POINTER:
 			configure_libinput(input->wlr_input_device);
 			break;
+		case WLR_INPUT_DEVICE_TOUCH:
+			map_touch_to_output(seat, input->wlr_input_device);
```
jp7677, line 526: I guess we should also call `configure_libinput(input->wlr_input_device);` here.

src/seat.c
```diff
@@ -273,9 +282,7 @@ new_touch(struct seat *seat, struct wlr_input_device *dev)
 
 	/* In support of running with WLR_WL_OUTPUTS set to >=2 */
 	if (dev->type == WLR_INPUT_DEVICE_TOUCH) {
```
jp7677, line 284: Not needed if statement.

tokyo4j:
I confirmed this works in my setup!

jp7677:
> I confirmed this works in my setup!

Very cool, I've added the final bits to this PR. Would you please retry to ensure that I didn't screwed up anything last minute?

jp7677:
Looking at `wlr_pointer`, this one also has a linked `output_name`, I wonder if we should also do this to ensure remapping when outputs wired?

```diff
diff --git a/src/seat.c b/src/seat.c
index a8d4162..a74b8ee 100644
--- a/src/seat.c
+++ b/src/seat.c
@@ -215,6 +215,14 @@ map_input_to_output(struct seat *seat, struct wlr_input_device *dev, char *outpu
        wlr_cursor_map_input_to_region(seat->cursor, dev, NULL);
 }
 
+static void
+map_pointer_to_output(struct seat *seat, struct wlr_input_device *dev)
+{
+       struct wlr_pointer *pointer = wlr_pointer_from_input_device(dev);
+       wlr_log(WLR_INFO, "map pointer to output %s\n", pointer->output_name);
+       map_input_to_output(seat, dev, pointer->output_name);
+}
+
 static struct input *
 new_pointer(struct seat *seat, struct wlr_input_device *dev)
 {
@@ -225,9 +233,7 @@ new_pointer(struct seat *seat, struct wlr_input_device *dev)
 
        /* In support of running with WLR_WL_OUTPUTS set to >=2 */
        if (dev->type == WLR_INPUT_DEVICE_POINTER) {
-               struct wlr_pointer *pointer = wlr_pointer_from_input_device(dev);
-               wlr_log(WLR_INFO, "map pointer to output %s\n", pointer->output_name);
-               map_input_to_output(seat, dev, pointer->output_name);
+               map_pointer_to_output(seat, dev);
        }
        return input;
 }
@@ -659,6 +665,9 @@ seat_output_layout_changed(struct seat *seat)
        struct input *input = NULL;
        wl_list_for_each(input, &seat->inputs, link) {
                switch (input->wlr_input_device->type) {
+               case WLR_INPUT_DEVICE_POINTER:
+                       map_pointer_to_output(seat, input->wlr_input_device);
+                       break;
                case WLR_INPUT_DEVICE_TOUCH:
                        map_touch_to_output(seat, input->wlr_input_device);
                        break;
```

docs/labwc-config.5.scd
```diff
@@ -445,6 +445,18 @@ windows using the mouse.
 	<mouse><mousebind> entries exist, the same default mousebinds will be
 	loaded even if the <default /> element is not provided.
 
+## TOUCH
+
+```
+<touch mapToOutput="" />
+```
+
+*<touch mapToOutput="" />*
```
johanmalm, line 454: Suggest simplifying this to something like:

Direct cursor movement to a specified output. If the compositor is running in nested mode, this does not take effect.

johanmalm (commented):
All looks good except one comment on description in documentation.

johanmalm:
> Looking at `wlr_pointer`, this one also has a linked `output_name`, I wonder if we should also do this to ensure remapping when outputs are wired?

I like this patch. Let's go for it. Looks much nicer.

src/seat.c
```diff
@@ -263,20 +263,25 @@ new_keyboard(struct seat *seat, struct wlr_input_device *device, bool virtual)
 	return (struct input *)keyboard;
 }
 
+static void
+map_touch_to_output(struct seat *seat, struct wlr_input_device *dev)
+{
+	struct wlr_touch *touch = wlr_touch_from_input_device(dev);
+	char *output_name = rc.touch.output_name ? rc.touch.output_name : touch->output_name;
```
johanmalm, line 270: Reverse fallback here:

`char *output_name = touch.output_name ? : rc.touch->output_name;`

`touch.output_name` is set when running nested (thanks @Consolatis ) so we don't want to affect input-output mapping for that.

PR #1430, opened 2024-01-15
ToggleShade, continued

ahesford:
This is a continuation of #1018. I've started chipping away at the points raised by @Consolatis:
- Interaction with various actions that resize the window or return it into its natural geometry (e.g. unmaximize / untile)
  - [x] Maximize - Disable shading when changing maximize state
  - [x] SnapToEdge - Disable shading when snapping
  - [x] SnapToRegion - Disable shading when snapping
  - [x] ResizeRelative - Disable shading when resizing (still needs testing)
  - [x] GrowToEdge - Disable shading when growing to edge
- [x] Interaction with ToggleDecorations (and Reconfigure) is missing, we should check `view->shaded` on ssd creation
    - Shading is not allowed on windows without SSD and a titlebar (We could enable SSD and a titlebar and then shade, but it would be ugly to shade a CSD window, then unshade and have the SSD titlebar and frame persist; it would be uglier still to try to track whether shading added SSD and then try to remove it when unshading.)
    - Toggling decorations is not allowed on shaded windows (Alternatively, we could just forcefully unshade windows when decorations are to be removed.)
    - When SSD are created, the frame uses the "effective" size to draw a border around only the titlebar.
- [x] Not verified with xwayland windows at all - A urxvt window seems to behave as expected, but this should be tested more thoroughly
- [x] The titlebar misses its top border as we disable the whole border tree
    This is addressed by resizing the border to account for an "effective" height of 0 when the window is shaded and leaving it displayed, rather than hiding the entire border tree. I suppose we could hide the sides and bottom of the border, but it seems easier to just shrink the border to fit the titlebar.

These are still missing
- [x] docs
- [x] default mouse bindings?
    Openbox uses scroll-wheel events on the titlebar to toggle shade. (Scrolling up always shades and scrolling down unshades---this might require Shade and Unshade actions in addition to ToggleShade.) This might be a reasonable default for labwc.

Additional points addressed in this PR:
- ShrinkToEdge is not allowed when shading (the notion of shrinking a nominally zero-area window is not well defined)
- Fullscreen windows will be restored to natural geometry before shading is enabled
- Automatic placement and snapping for move-to-edge actions use an effective height of 0 for shaded windows, so that the behavior feels correct for a window consisting only of a titlebar.
- Resize indicators also use the effective height, so the popup appears halfway off the titlebar instead of at the midpoint of an invisible view. This seems like a sensible default; trying to center the popup vertically on the titlebar would be a bit clumsy and would also make it harder to visually distinguish the popup from the bar.

I'll continue working through the remaining points, but any comments on the behavior or suggestions about outstanding issues are welcome.


johanmalm:
Have tested. Works well.

> default mouse bindings? Openbox uses scroll-wheel events on the titlebar to toggle shade. (Scrolling up always shades and scrolling down unshades---this might require Shade and Unshade actions in addition to ToggleShade.) This might be a reasonable default for labwc.

Agree, seems reasonable.

> ShrinkToEdge is not allowed when shading (the notion of shrinking a nominally zero-area window is not well defined)

We allow it in `view_grow_to_edge()`, so ought to be consistent.
I vote for copying that behaviour in `view_shrink_to_edge()`

ahesford:
The latest batch of changes should address the remaining points. In addition:
- There are now three actions, `Shade`, `Unshade`, and `ToggleShade`.
- Default mouse bindings will shade (and unfocus) a window when scrolling up on the title bar and unshade (and focus) a window when scrolling down on the titlebar.
- When an interactive move begins on a non-floating (tiled or maximized) view, a shaded window will be unshaded. This eliminates a *LOT* of edge casing that would be required to behave correctly when preserving the shaded state, and it seems relatively intuitive anyway.
- When a window enters full-screen mode, it is unshaded. (Otherwise, the window would just disappear from view.)
- Edge resistance when moving should respect the effective size of shaded windows, stopping with the titlebar at the bottom edge rather than where the unshaded window would end. This is consisten with move-to-edge behavior.

johanmalm:
Super work. Thanks!

PR #1431, opened 2024-01-15
Configure border and titlebar color per app by using window rules

dmitry-erin:
The PR makes it possible to change title-bar and border's color per app using window rules. The property borderColor was added to the window rules. There are different approaches of changing the color in different cases: just color substituting and the whole theme changing.


src/theme.c
```diff
@@ -1023,3 +1023,25 @@ theme_finish(struct theme *theme)
 	theme->corner_top_right_active_normal = NULL;
 	theme->corner_top_right_inactive_normal = NULL;
 }
+
+void theme_customize_with_border_color(struct theme *theme, float *color)
```
ahesford, line 1027: This should be a static function in `ssd/ssd_titlebar.c`, where it's used.

src/ssd/ssd_titlebar.c
```diff
@@ -25,6 +25,15 @@ ssd_titlebar_create(struct ssd *ssd)
...
 	struct theme *theme = view->server->theme;
+
+	/* Here the whole theme changing is more preferable */
+	struct theme custom_theme = { 0 };
+	float custom_color[4];
+	if (window_rules_get_custom_border_color(view, custom_color)) {
+		theme_customize_with_border_color(&custom_theme, custom_color);
+		theme = &custom_theme;
+	}
+
```
ahesford, line 36: In addition to moving `theme_customize_with_border_color` to this file as a static function, you should probably just make it an omnibus function that returns a `struct theme` that you can use in one shot, *e.g.*:
```c
struct theme theme = get_theme_for_view(view);
```
where the function looks something like
```c
static struct theme
get_theme_for_view(struct view *view)
{
    float custom_color[4];
    if (!window_rules_get_custom_border_color(view, custom_color)) {
        return *view->server->theme;
    }

    struct theme theme = { 0 };
    /* The rest of your original theme_customize_with_border_color() here */
    return theme;
}

File: src/config/rcxml.c [hunk 3]
```diff
@@ -119,6 +120,7 @@ fill_window_rule(char *nodename, char *content)
 {
 	if (!strcasecmp(nodename, "windowRule.windowRules")) {
 		current_window_rule = znew(*current_window_rule);
+		init_window_rule(current_window_rule);
```
ahesford, line 123: Drop this.

src/config/rcxml.c
```diff
@@ -162,6 +165,17 @@ fill_window_rule(char *nodename, char *content)
...
+	/* Custom border properties: color */
+	} else if (!strcasecmp(nodename, "borderColor")) {
+		parse_hexstr(content, current_window_rule->custom_border_color);
+		current_window_rule->has_custom_border = true;
+		wlr_log(WLR_DEBUG,
+			"Custom borderColor was found in config: %s, parsed into: %f, %f, %f, %f\n",
+			content, current_window_rule->custom_border_color[0],
+			current_window_rule->custom_border_color[1],
+			current_window_rule->custom_border_color[2],
+			current_window_rule->custom_border_color[3]);
```
ahesford, line 177: Debug messages here seem inappropriate, as most config parsing doesn't generate these messages.

src/ssd/ssd.c
```diff
@@ -340,9 +340,13 @@ ssd_enable_keybind_inhibit_indicator(struct ssd *ssd, bool enable)
...
-	float *color = enable
-		? rc.theme->window_toggled_keybinds_color
-		: rc.theme->window_active_border_color;
+	float custom_color[4];
+	bool is_custom_color_available =
+		window_rules_get_custom_border_color(ssd->view, custom_color);
+
+	float *color = is_custom_color_available ? custom_color :
+		(enable ? rc.theme->window_toggled_keybinds_color
+				: rc.theme->window_active_border_color);
```
ahesford, line 349: I don't think the custom theme color should override the toggled-keybind color like this. Toggling keybinds is an ephemeral state, the indication of which is probably more important than honoring a color configured by a rule.

src/ssd/ssd_titlebar.c
```diff
@@ -137,11 +147,16 @@ set_squared_corners(struct ssd *ssd, bool enable)
...
+			struct wlr_scene_rect *rect = lab_wlr_scene_get_rect(button->background);
+			/*Check for custom color as well*/
+			float custom_color[4];
+			bool is_custom_color_available =
+				window_rules_get_custom_border_color(ssd->view, custom_color);
+			wlr_scene_rect_set_color(rect, is_custom_color_available ? custom_color
+				: (!enable ? (float[4]) {0, 0, 0, 0}
+					: (subtree == &ssd->titlebar.active
+						? rc.theme->window_active_title_bg_color
+						: rc.theme->window_inactive_title_bg_color)));
```
ahesford, line 159: This is getting really hard to parse; a separate function would probably be a better option here.

src/window-rules.c
```diff
@@ -73,6 +73,16 @@ view_matches_criteria(struct window_rule *rule, struct view *view)
...
+void
+init_window_rule(struct window_rule *rule)
+{
+	if (!rule) {
+		return;
+	}
+	rule->has_custom_border = false;
+	memset(rule->custom_border_color, 0, sizeof(float)*4);
+}
+
```
ahesford, line 85: This function is not necessary because structs will be zero-initialized. If it were necessary, it shouldn't should exist only to set the color.

src/window-rules.c
```diff
@@ -135,3 +145,21 @@ window_rules_get_property(struct view *view, const char *property)
...
+				memcpy(rgba, rule->custom_border_color, sizeof(float)*4);
+				wlr_log(WLR_DEBUG,
+					"Custom color was found in window rules list: %f, %f, %f, %f\n",
+					rgba[0], rgba[1], rgba[2], rgba[3]);
+				return true;
+			}
+		}
+	}
+	return false;
+}
```
ahesford, line 165: 1. I don't think the debug info is particularly useful here.
2. This function should only take the `struct view` argument and return a float pointer to the internal `rule->custom_border_color` (if the custom color has been defined) or `NULL`. There's no value in copying the values around, and using the pointer simplifies logic elsewhere.

ahesford (changes requested):
I'm not convinced that this is the best approach to this feature. The capability is very limited (it doesn't even distinguish between active and inactive border colors) and complicates the logic in several places. It might be cleaner (and simpler) to support entirely custom theme overrides based on window rules and generalize all theming logic accordingly.

In case others disagree and think this feature is desirable as implemented, I've made some suggestions. However, I'd wait for feedback from others on the broader issues before adopting these revisions.

src/theme.c
dmitry-erin: replaced with get_theme_for_view() as suggested

src/ssd/ssd.c
dmitry-erin: To be honest, I haven't gone deep into the meaning of this part, I just noticed that it produces the bar on the top of the window, which is not colored. I will think how to resolve it (suggestions are also welcome!).

src/ssd/ssd_titlebar.c
dmitry-erin: if statement added, the rest of code turned back.

src/theme.c
```diff
@@ -1042,3 +1044,40 @@ theme_finish(struct theme *theme)
...
+
+struct theme
+get_theme_for_view(struct view *view)
+{
+	float *custom_color = window_rules_get_custom_border_color(view);
+	if (!custom_color) {
+		return *view->server->theme;
+	}
+
+	struct theme theme = { 0 };
```
johanmalm, line 1056: Declaring `theme` here and then returning it, doesn't look right. Its life ends with this frame.

src/theme.c
```diff
@@ -1042,3 +1044,40 @@ theme_finish(struct theme *theme)
...
+
+	/* The same code as in theme_init*(), but with introducing custom color*/
+	theme_builtin(&theme);
+
+	struct wl_list paths;
+	paths_theme_create(&paths, rc.theme_name, "themerc");
+	theme_read(&theme, &paths);
```
johanmalm, line 1063: Oh, reading the theme multiple times doesn't feel right. Could we make copies of it (on the heap)? Is it the creation of the rounded corners that becomes a problem?

johanmalm (commented):
Thanks for working on this.

At a glance, the code looks neat, but there are a couple of things that stand out (have only looked very quickly). Have left comments.

If you want I can take a longer look at try help devise a strategy.

johanmalm:
I like the idea of this patch - but think it needs to implemented differently.  It'll probably be easier after SSD refactoring.
Have set to draft, but keen to leave it open for reference.

Issue #1407, opened 2024-01-04
On-the-fly keyboard layout changes?


spl237:
I can change the keyboard layout by setting XKB_DEFAULT_LAYOUT in ~/.config/labwc/environment and restarting labwc.

Is it possible to change the keyboard layout without having to restart labwc? Changing the value in environment and calling labwc --reconfigure doesn't work, and the environment file doesn't seem to be automatically monitored for changes. Updating the environment variable directly in a terminal doesn't work, either with or without a labwc --reconfigure.

I'm assuming this is possible - for some locales, people need to switch between two keymaps on the fly - but I can't see anything in the documentation other than just setting the environment variables.

stefonarch:
Take a look t the example file:
https://github.com/labwc/labwc/blob/6cf19d1cdbffc699ca8f7899d8e2bb680474789a/docs/environment#L14

Any option can be used among the ones listed here:
`localectl list-x11-keymap-options | grep grp:`

spl237:
Ah - while that does indeed handle the on-the-fly switching of keymaps, it doesn't solve the actual problem I am having! If the user wants to change keyboard to one not previously defined - say switching from a US to UK keyboard layout because they have plugged in a new keyboard, or as part of an initial setup wizard where the user sets their keyboard layout before setting their password for the first time - it needs to be possible for the keyboard layout to be changed to any other layout on the fly.

Consolatis:
We do read the environment file on reconfigure and also apply it to the environment so spawing new applications after that point will include the changes.

However, we do currently not re-create the xkb layout (nor the xcursor theme). I think there is an issue for that somewhere. It should be possible to implement, although it might become a bit finicky if combined with [`layoutScope=window`](https://labwc.github.io/labwc-config.5.html#entry_keyboard_layoutscope=) as we only store the "group" modifier index per window. Thus if the new keyboard layout now suddenly changes their meaning (or even the count) things might get a bit quirky.

spl237:
> We do read the environment file on reconfigure and also apply it to the environment so spawing new applications after that point will include the changes.

Just tried that - it doesn't seem to work.
So if I have XKB_DEFAULT_LAYOUT=gb in ~/.config/labwc/environment and hence a GB keyboard layout, if I change that line to XKB_DEFAULT_LAYOUT=us, save the file and then run labwc --reconfigure in a terminal, I still get the GB keyboard layout on newly-opened applications thereafter.

Consolatis:
[...]

PR #1443, opened 2024-01-16
Reload keymap on reconfigure

spl237:
This reloads the keymap according to the current XKB_DEFAULT_ environment variables when labwc --reconfigure is called, as discussed here - https://github.com/labwc/labwc/issues/1407

I'm not sure if this is exactly the right approach, but it is what wayfire does, and it seems to work.

### DIFF
--- a/src/input/input.c
+++ b/src/input/input.c
@@ -7,7 +7,7 @@ void
 input_handlers_init(struct seat *seat)
 {
 	cursor_init(seat);
-	keyboard_init(seat);
+	keyboard_init(seat, false);
 }
 
 void
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -614,6 +614,24 @@ keyboard_init(struct seat *seat)
 }
 
 void
+keyboard_reload(struct seat *seat)
+{
+	struct wlr_keyboard *kb = wlr_seat_get_keyboard(seat->seat);
+	struct xkb_rule_names rules = { 0 };
+	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
+	struct xkb_keymap *keymap = xkb_map_new_from_names(context, &rules,
+		XKB_KEYMAP_COMPILE_NO_FLAGS);
+	if (keymap) {
+		wlr_keyboard_set_keymap(kb, keymap);
+		xkb_keymap_unref(keymap);
+	} else {
+		wlr_log(WLR_ERROR, "Failed to create xkb keymap");
+	}
+	xkb_context_unref(context);
+	wlr_keyboard_set_repeat_info(kb, rc.repeat_rate, rc.repeat_delay);
+}
+
+void
 keyboard_setup_handlers(struct keyboard *keyboard)
 {
 	struct wlr_keyboard *wlr_kb = keyboard->wlr_keyboard;
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -593,10 +593,15 @@ keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout)
 }
 
 void
-keyboard_init(struct seat *seat)
+keyboard_init(struct seat *seat, bool reconfig)
 {
-	seat->keyboard_group = wlr_keyboard_group_create();
-	struct wlr_keyboard *kb = &seat->keyboard_group->keyboard;
+	struct wlr_keyboard *kb;
+	if (reconfig) {
+		kb = wlr_seat_get_keyboard(seat->seat);
+	} else {
+		seat->keyboard_group = wlr_keyboard_group_create();
+		kb = &seat->keyboard_group->keyboard;
+	}
 	struct xkb_rule_names rules = { 0 };
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
 	struct xkb_keymap *keymap = xkb_map_new_from_names(context, &rules,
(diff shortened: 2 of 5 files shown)


ahesford:
I can't say whether `keyboard_reload` is called at the right time in the initialization, but
1. I suspect that `keybind_update_keycodes` should be called to do the right thing should the keymap change. (I'm not sure if this requires special handling of the keybind structures.)
2. This function is a duplicate of `keyboard_init` except for the first two lines (and, currently, without the final `keybind_update_keycodes`). This should be collapsed into a common function.

I think the best way to handle this is to rename

    keyboard_init(struct seat *seat) -> keyboard_configure(struct seat *seat, bool reconfigure)

and then do something like

```c
void
keyboard_configure(struct seat *seat, bool reconfigure)
{
	struct wlr_keyboard *kb = NULL;
	if (reconfigure) {
		kb = wlr_seat_get_keyboard(seat->seat);
	} else {
		seat->keyboard_group = wlr_keyboard_group_create();
		kb = &seat->keyboard_group->keyboard;
	}
	/* Asserts can be disabled at compile-time; should this die another way when !kb? */
	assert(kb);
	
	/* Rest of the original keyboard_init goes here... */
}
```

If there is an argument against calling `keybind_update_keycodes` on reconfigure, then the final line of the function could be cated with `!reconfigure`.

ahesford:
Thanks for pointing that out. In this case, I'd drop the `keybind_update_keycodes` in `reload_config_and_theme` and let a combined `keyboard_configure` function do it.

spl237:
Done...

src/input/keyboard.c
```diff
@@ -593,10 +593,15 @@ keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout)
 }
 
 void
-keyboard_init(struct seat *seat)
+keyboard_init(struct seat *seat, bool reconfig)
 {
-	seat->keyboard_group = wlr_keyboard_group_create();
-	struct wlr_keyboard *kb = &seat->keyboard_group->keyboard;
+	struct wlr_keyboard *kb;
```
ahesford, line 598:
```suggestion
	struct wlr_keyboard *kb = NULL;
```

src/input/keyboard.c
```diff
@@ -593,10 +593,15 @@ keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout)
...
+	if (reconfig) {
+		kb = wlr_seat_get_keyboard(seat->seat);
+	} else {
+		seat->keyboard_group = wlr_keyboard_group_create();
+		kb = &seat->keyboard_group->keyboard;
+	}
```
ahesford, line 604:
```suggestion
	}
	assert(kb);
```

src/input/input.c
```diff
@@ -7,7 +7,7 @@ void
 input_handlers_init(struct seat *seat)
 {
 	cursor_init(seat);
-	keyboard_init(seat);
+	keyboard_init(seat, false);
```
ahesford, line 10:
```suggestion
	keyboard_init(seat, /* reconfig */ false);
```

src/seat.c
```diff
@@ -537,6 +537,7 @@ seat_reconfigure(struct server *server)
 			break;
 		}
 	}
+	keyboard_init(seat, true);
```
ahesford, line 540:
```suggestion
	keyboard_init(seat, /* reconfig */ true);
```

Consolatis:
This PR completely ignores `view->keyboard_layout` (e.g. per-window keyboard layout). We should at least check of the new amount of keyboard layout groups is less than the previous one and then clamp the stored `view->keyboard_layout` ones to the new max index.

We could also compare the new keymap vs the old one (via a simple `!strcmp()`) and if they match simply do nothing at all (e.g. not even set it as a new keymap).

spl237:
> This PR completely ignores `view->keyboard_layout` (e.g. per-window keyboard layout). We should at least check if the new amount of keyboard layout groups is less than the previous one and then clamp the stored `view->keyboard_layout` ones to the new max index.

I don't begin to understand how per-window keyboard layouts work, so I'll leave that to someone else who does! 

This PR is proof-of-concept of a mechanism by which it is possible for keymap reload to work; I'm sure upstream devs have better ideas on how it should be implemented, so I'll leave it to you at this point.

src/input/input.c
ahesford: Duplicating an entire function except for the first two lines is much worse than adding an argument, and the comment is in keeping with the style of other calls that annotate the meaning of non-obvious arguments like this. Also note that my original suggestion was to rename this to `keyboard_configure` so there would be less confusion about repeated calls.

However, as Consolatis pointed out in IRC, you can probably drop the extra argument altogether and check in `keyboard_init` whether `seat->keyboard_group == NULL` to determine whether you need to perform an initial creation or just reconfigure.

spl237: > Duplicating an entire function except for the first two lines is much worse than adding an argument...

Why? In what way is it "much worse"? It's a tiny function - barely ten lines. As a fraction of the size of the thousands of lines of labwc code, the effect of duplicating it is negligible. (Some compilers would even optimise that out anyway.) And calling it "keyboard_configure" doesn't make it any more obvious as to what the argument means when it is called.

My preference would be to retain separate keyboard_init and keyboard_reconfig function calls with, if repeating code is felt to be unacceptable, the common code removed to a third function which they both call. That makes it very obvious what is going on to anyone reading the code with no need for comments at all.

ahesford: Copy-pasting functions leads to divergence and latent bugs when somebody decides to modify the behavior of one and forgets to synchronize the other. The line count is irrelevant; any time you are trying to do the same non-trivial task more than once, that task should be wrapped in a function.

However, as I noted before, there is a better way to do this without the need for any change in the call signature.

johanmalm:
I started committing on this branch, but didn't want to push / force-push without asking. Need to finish for tonight, to have just pushed the whole thing to another branch.

https://github.com/labwc/labwc/pull/1450

Consolatis:
Replaced by
- #1450

PR #1448, opened 2024-01-17
view: assign output on surface creation instead of mapping...

ahesford:
...and notify the client of the preferred output scale when doing so. This should allow clients to better determine an optimal size if they are initially configured (unmapped) with zero size.

In particular, this fixes an issue with foot: https://codeberg.org/dnkl/foot/issues/1579.


Consolatis:
Nice one, LGTM. Didn't test.

johanmalm:
Thanks!

Impressive analysis on the foot issue

PR #1449, opened 2024-01-17
feat: multiple touch configurations with device name identifier

jp7677:
Not a very big fan of this, but it's reasonable and should address the initial issue in https://github.com/labwc/labwc/issues/1422

Still ~WIP...~ needs testing (~also by me, I have just checked that it compiles but wanted CI feedback) and docs~.

PS: the first commit is just a small refactor in the tablet space .

### DIFF
--- a/include/config/touch.h
+++ b/include/config/touch.h
@@ -1,0 +1,14 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_TOUCH_CONFIG_H
+#define LABWC_TOUCH_CONFIG_H
+
+#include <stdint.h>
+
+struct touch_config_entry {
+	char *device_name;
+	char *output_name;
+
+	struct wl_list link;     /* struct rcxml.touch_configs */
+};
+
+#endif /* LABWC_TOUCH_CONFIG_H */
--- a/include/config/touch.h
+++ b/include/config/touch.h
@@ -11,4 +11,6 @@ struct touch_config_entry {
 	struct wl_list link;     /* struct rcxml.touch_configs */
 };
 
+struct touch_config_entry *touch_find_config_for_device(char *device_name);
+
 #endif /* LABWC_TOUCH_CONFIG_H */
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -36,6 +36,7 @@ static bool in_regions;
 static bool in_usable_area_override;
 static bool in_keybind;
 static bool in_mousebind;
+static bool in_touch;
 static bool in_libinput_category;
 static bool in_window_switcher_field;
 static bool in_window_rules;
@@ -46,6 +47,7 @@ static bool in_action_else_branch;
 static struct usable_area_override *current_usable_area_override;
 static struct keybind *current_keybind;
 static struct mousebind *current_mousebind;
+static struct touch_config_entry *current_touch;
 static struct libinput_category *current_libinput_category;
 static const char *current_mouse_context;
 static struct action *current_keybind_action;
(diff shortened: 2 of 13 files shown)


ahesford:
Can the `touch_config_entry` array become a `wl_list`? The number of people wanting more than eight devices is almost certainly small, but may not be trivial.

jp7677:
@spl237 Feel free to give this PR a shot for your Pi's and please report back. It should be good, but my testing was rather limited/theoretical with debug statements since I don't have such a setup at hand.

With this PR you can setup the configuration like this (assuming that I got the Pi touch device names correctly):

```xml
  <touch device="4-0038 generic ft5x06 (79)" mapToOutput="DSI-1" />
  <touch device="6-0038 generic ft5x06 (79)" mapToOutput="DSI-2" />
  <touch mapToOutput="DP-1" /> <!-- fallback for other touch devices, optionally -- >
```

The labwc logs (`labwc -d`) will give you the detected device names.

jp7677:
PS: This is still a draft, so everything is still open for discussion ofc.

src/config/rcxml.c
```diff
@@ -863,8 +869,11 @@ entry(xmlNode *node, char *nodename, char *content)
 		} else {
 			wlr_log(WLR_ERROR, "Invalid value for <resize popupShow />");
 		}
+	} else if (!strcasecmp(nodename, "device.touch")) {
```
johanmalm, line 872: I think we should avoid the term `device` here because it's used in `xml_tree_walk()` which makes it special.
https://github.com/labwc/labwc/blob/4238d7fc33dc2c7c4ee4ecbac81decbc91833c1a/src/config/rcxml.c#L948

Maybe `device_name`?

src/config/rcxml.c
```diff
@@ -863,8 +869,11 @@ entry(xmlNode *node, char *nodename, char *content)
...
+		touch_config_device_name = xstrdup(content);
 	} else if (!strcasecmp(nodename, "mapToOutput.touch")) {
-		rc.touch.output_name = xstrdup(content);
+		char *output_name = xstrdup(content);
+		touch_config_add(touch_config_device_name, output_name);
```
johanmalm, line 876: It doesn't feel right to create the object on `mapToOutput.touch`. Technically you could, but if the device-name is given first you have to keep track of the independently which is messy.

Could we handle `touch` elements in `xml_tree_walk()` (i.e. with an `in_touch` state variable and `fill_touch()` function. Suggest searching for + following `in_region` and `fill_region` for an idea of what I'm thinking.

I suspect you'll find it cleaner :smile:

johanmalm (commented):
I like the look of this. Just a couple of design suggestions.

spl237:
> @spl237 Feel free to give this PR a shot for your Pi's and please report back. It should be good, but my testing was rather limited/theoretical with debug statements since I don't have such a setup at hand.

I'm out of the office and hence away from any touchscreen devices until Monday; will give it a try then! Many thanks in advance!

src/config/rcxml.c
jp7677: yes indeed, much better now, thanks ;)

spl237:
Tested this here with the touch device I have, and it works fine. I don't have multiple touch devices on hand, but given that the mechanism works with a single named device, I can't see why it wouldn't work in that case. I'll try to find a second touch device for some more testing in future, but this looks good to merge to me.

include/config/touch.h
```diff
@@ -0,0 +1,16 @@
...
+#ifndef LABWC_TOUCH_CONFIG_H
+#define LABWC_TOUCH_CONFIG_H
+
+#include <stdint.h>
+
+struct touch_config_entry {
+	char *device_name;
+	char *output_name;
+
+	struct wl_list link;     /* struct rcxml.tablet_configs */
```
johanmalm, line 11: `s/tablet/touch/` ??

johanmalm (commented):
Was going for the merge, but noticed one tiny nit in a comment...

include/config/touch.h
jp7677: I've worked to long on the tablet stuff :) fixed.

johanmalm:
Thanks!

spl237:
Many thanks to everyone for this!

PR #1450, opened 2024-01-17
keyboard: set keyboard layout on reconfigure

### LINKED ISSUE #1407: On-the-fly keyboard layout changes?

johanmalm:
If keyboard-layout-per-toplevel-window is used, reset the group (index) for each window on --reconfigure whenever XKB_DEFAULT_LAYOUT has changed.

Refactor to use a common configure function for reconfigure and keyboard-group creation.

Co-authored-by: @johanmalm

Fixes #1407

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -592,25 +592,72 @@ keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout)
 		kb->modifiers.latched, kb->modifiers.locked, layout);
 }
 
-void
-keyboard_init(struct seat *seat)
+static void
+reset_window_keyboard_layout_groups(struct server *server)
 {
-	seat->keyboard_group = wlr_keyboard_group_create();
-	struct wlr_keyboard *kb = &seat->keyboard_group->keyboard;
+	if (!rc.kb_layout_per_window) {
+		return;
+	}
+
+	/*
+	 * Technically it would be possible to reconcile previous group indices
+	 * to new group ones if particular layouts exist in both old and new,
+	 * but let's keep it simple for now and just reset them all.
+	 */
+	struct view *view;
+	for_each_view(view, &server->views, LAB_VIEW_CRITERIA_NONE) {
+		view->keyboard_layout = 0;
+	}
+
+	struct view *active_view = server->active_view;
+	if (!active_view) {
+		return;
+	}
+	keyboard_update_layout(&server->seat, active_view->keyboard_layout);
+}
+
+/*
+ * Set layout based on environment variables XKB_DEFAULT_LAYOUT,
+ * XKB_DEFAULT_OPTIONS, and friends.
+ */
+static void
+set_layout(struct server *server, struct wlr_keyboard *kb)
+{
 	struct xkb_rule_names rules = { 0 };
 	struct xkb_context *context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
 	struct xkb_keymap *keymap = xkb_map_new_from_names(context, &rules,
 		XKB_KEYMAP_COMPILE_NO_FLAGS);
 	if (keymap) {
-		wlr_keyboard_set_keymap(kb, keymap);
+		if (!wlr_keyboard_keymaps_match(kb->keymap, keymap)) {
+			wlr_keyboard_set_keymap(kb, keymap);
+			reset_window_keyboard_layout_groups(server);
+		}
 		xkb_keymap_unref(keymap);
 	} else {
 		wlr_log(WLR_ERROR, "Failed to create xkb keymap");
 	}
 	xkb_context_unref(context);
-	wlr_keyboard_set_repeat_info(kb, rc.repeat_rate, rc.repeat_delay);
+}
 
+void
+keyboard_configure(struct seat *seat, struct wlr_keyboard *kb, bool is_virtual)
+{
+	if (!is_virtual) {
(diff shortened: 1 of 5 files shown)


johanmalm:
I've given this some basic testing with single+multiple keyboard layouts. Seems to work, but could do with both code review + testing by others.

src/seat.c
```diff
@@ -506,13 +506,51 @@ seat_finish(struct server *server)
...
+	if (!rc.kb_layout_per_window) {
+		return;
+	}
+
+	/*
+	 * We don't want to reset all window keyboard-layout groups on every
+	 * reconfigure unless XKB_DEFAULT_LAYOUT has actully changed.
+	 */
+	static char last_layout[64];
+	char *current_layout = getenv("XKB_DEFAULT_LAYOUT") ? : "";
```
ahesford, line 520: Recommend avoiding the GNU extension to `?:`.

ahesford (commented):
I'm not terribly familiar with input handling, but these changes seem reasonable.

src/seat.c
```diff
@@ -506,13 +506,51 @@ seat_finish(struct server *server)
...
 {
+	if (!rc.kb_layout_per_window) {
+		return;
+	}
+
+	/*
+	 * We don't want to reset all window keyboard-layout groups on every
+	 * reconfigure unless XKB_DEFAULT_LAYOUT has actully changed.
+	 */
+	static char last_layout[64];
```
Consolatis, line 519: I am not sure if 64 is enough. You can put of \*lot\* of options there.

src/input/keyboard.c
```diff
@@ -592,11 +592,13 @@ keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout)
...
 }
 
-void
-keyboard_init(struct seat *seat)
+/*
+ * Set layout based on environment variables XKB_DEFAULT_LAYOUT,
+ * XKB_DEFAULT_OPTIONS, and friends.
+ */
+static void
+set_layout(struct wlr_keyboard *kb)
```
Consolatis, line 600: I think we should prevent setting the same keymap again. We can test for that with `wlr_keyboard_keymaps_match()`. 

That would also allow to call `reset_window_keyboard_layout_groups()` only when changing the keymap without us requiring to compare the single env var. The user might also have changed some system layout files or used other environment vars that influence the layout.

src/input/keyboard.c
```diff
@@ -608,11 +610,30 @@ keyboard_init(struct seat *seat)
...
 	keybind_update_keycodes(seat->server);
 }
 
+void
+keyboard_group_init(struct seat *seat)
+{
+	if (seat->keyboard_group) {
+		wlr_log(WLR_ERROR, "keyboard group already exists - noop");
+		return;
+	}
```
Consolatis, line 631: IMHO this should either be an `assert()` or we don't need to log here.

src/seat.c
johanmalm: Done

src/seat.c
johanmalm: Have increased to 1024.

src/input/keyboard.c
johanmalm: Nice catch. I didn't know about `wlr_keyboard_keymaps_match()`

src/input/keyboard.c
johanmalm: Done. `assert()` used.

src/input/keyboard.c
johanmalm: > That would also allow to call reset_window_keyboard_layout_groups() only when changing the keymap without us being required to compare the single env var. The user might also have changed some system layout files or used other environment vars that influence the keymap.

I'm not sure that works. For example, what if a user has changed group (e.g. `de` -> `se`) before a `--reconfigure`. That would then trigger a `reset_window_keyboard_layout_groups()` even though the `XKB_DEFAULT_LAYOUT` didn't change. Or is my brain just not thinking straight?

Consolatis: > For example, what if a user has changed group (e.g. `de` -> `se`) before a `--reconfigure`.

Not sure I follow. You mean via `XKB_DEFAULT_LAYOUT`?

As long as the user doesn't change any system xkb files or one of the `XKB_` vars before a `--reconfigure` the keymap should be the same and thus we would not call `reset_window_keyboard_layout_groups()`.

johanmalm: Yes, I'm stupid. I was thinking `group` rather than `keymap`. Have pushed fixup. Works fine now.

johanmalm:
Okay. Have squashed. Ready for final review.

Consolatis:
Code LGTM but I did not test at all.

johanmalm:
Thanks for review... much better after your suggestions :smile:
Have tested again tonight. Seems to work fine.

spl237:
Tested here - works fine for me. Many thanks!

PR #1452, opened 2024-01-19
view: fix regression whereby cursor is not updated

### LINKED ISSUE #1393: Cursor disappears when closing wayland clients (reappears when cursor is moved again)
When closing kitty or chromium with their wayland clients, the cursor disappears. It only shows back again if the cursor is moved. This does not happen when using chromium with Xwayland.

johanmalm:
...on view destruction because focus_change_notify() in seat.c changes server->active_view so the logic introduced by 6c6e406 (which checks if view =! active_view) is no longer right.

The glitches described in the commit below have not come back. https://github.com/labwc/labwc/commit/6c6e406507bfb0e9a809b74029b5b05b5a095768

Fixes #1393

@jlindgren90 - This relates to a couple of separate commits from you, so if you've got time, your eyes on this would helpful. Don't worry if you can't, I think the change is fine.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2046,7 +2046,6 @@ view_destroy(struct view *view)
 {
 	assert(view);
 	struct server *server = view->server;
-	bool need_cursor_update = false;
 
 	if (view->mappable.connected) {
 		mappable_disconnect(&view->mappable);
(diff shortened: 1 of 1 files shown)


jlindgren90:
LGTM, thanks

PR #1455, opened 2024-01-21
resistance: add window-edge resistance for interactive moves/resizes

ahesford:
This includes a general refactor of the resistance logic as well as re-use of the major components to allow resistance to apply to window edges in addition to screen edges. Moving behavior seems to be correct, but there are still a couple of bugs with resizing:
- [x] Resizing to the right snaps to another window as expected
- [x] Resizing to the left or to the top does not seem to encounter any resistance
- [x] Resizing to the bottom causes strange jumps in window size

### DIFF
--- a/src/resistance.c
+++ b/src/resistance.c
@@ -12,20 +12,28 @@ static void
 is_within_resistance_range(struct border view, struct border target,
 		struct border other, struct border *flags, int strength)
 {
-	if (view.left >= other.left && target.left < other.left
-			&& target.left >= other.left - strength) {
-		flags->left = 1;
-	} else if (view.right <= other.right && target.right > other.right
-			&& target.right <= other.right + strength) {
-		flags->right = 1;
+	if (view.left >= other.left) {
+		const int lo = other.left - abs(strength);
+		const int hi = other.left - MIN(strength, 0);
+		flags->left = target.left >= lo && target.left < hi;
 	}
 
-	if (view.top >= other.top && target.top < other.top
-			&& target.top >= other.top - strength) {
-		flags->top = 1;
-	} else if (view.bottom <= other.bottom && target.bottom > other.bottom
-			&& target.bottom <= other.bottom + strength) {
-		flags->bottom = 1;
+	if (!flags->left && view.right <= other.right) {
+		const int lo = other.right + MIN(strength, 0);
+		const int hi = other.right + abs(strength);
+		flags->right = target.right > lo && target.right <= hi;
+	}
+
+	if (view.top >= other.top) {
+		const int lo = other.top - abs(strength);
+		const int hi = other.top - MIN(strength, 0);
+		flags->top = target.top >= lo && target.top < hi;
+	}
+
+	if (!flags->top && view.bottom <= other.bottom) {
+		const int lo = other.bottom + MIN(strength, 0);
+		const int hi = other.bottom + abs(strength);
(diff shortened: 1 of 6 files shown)


ahesford:
The resize-toward-bottom issue was a simple typo. As for left/top, it turns out that this is already broken for screen-edge resistance on master. While these changes seem to preserve existing behavior, I'd like to spend a bit of time thinking about how to resolve this issue.

ahesford:
Ok, I think resizing resistance is properly implemented even for left- and top-edge resizes.

src/resistance.c
```diff
@@ -32,160 +35,214 @@ is_within_resistance_range(struct edges view, struct edges target,
...
-	target_edges.bottom = new_view_geo->y + new_view_geo->height
-		+ border.bottom;
+void
+resistance_resize_apply(struct view *view, struct wlr_box *new_geom)
+{
+	assert(view);
 
-	if (!rc.screen_edge_strength) {
+	/* Shaded windows shoudln't be resized, so ignore resistance */
+	if (view->shaded) {
```
johanmalm, line 207: Will `view->shaded` ever happen here? If so, should we carry out that check in the calling function?

johanmalm (commented):
LGTM

In fact, very good.

`resistance.c:10`: Could consider using `struct border` instead of `struct edges` if you want. Will hold off merging to get your thoughts on that. Ref: https://github.com/labwc/labwc/blob/f73e9ded1c8df004acead9a3cfc5ec0fbf3a68d6/include/common/border.h#L5-L10

src/resistance.c
ahesford: This could probably be dropped altogether or, if retained, converted to an assert. I don't think this function will ever be entered with a shaded view.

ahesford:
Reusing `struct border` would be fine. I'm also thinking about an alternative implementation that uses the new automatic-placement grid. The idea would be cache the grid for whatever output maps to `server-> grab_{x,y}` and find the grid interval containing the cursor to determine if we're near the relevant grid edge. This avoids the need to walk the whole view list with every move/resize event. If the output changes because the cursor crosses a boundary, throw out the cache and recompute the grid for the new output. At the end of an interactive move/resize, discard the cache outright.

Give me a bit of time to play with the alternative to see if it is worth the bookkeeping.

johanmalm:
> Give me a bit of time to play with the alternative to see if it is worth the bookkeeping.

:+1: 

There is some overlap with `snap.c` here. Happy to consider sharing code for another time, but thought I'd mention it. We did consider doing a grid at the time, but the dev preferred how it's written and I went with it.

I guess not too dissimilar from the placement policy either :smile:

ahesford:
The latest version of this branch drops the `struct edges` in favor of the pre-existing `struct border`, and also interprets negative resistance values as "attraction", which will cause moves to snap to a relevant edge when the cursor gets within the specified distance.

ahesford:
I put together a test implementation using the placement grid, but it was getting cumbersome and buggy. I'd like to defer use of the placement grid and the possibility of re-using logic with `snap.c` for now.

The biggest question at the moment is whether there are any objections to the negative-strength additions.

johanmalm:
I love the negative. Works a treat. And nice that it doesn't break current behaviour (as set to positive on default).

This looks good. 
Shall I go for the merge?

Consolatis:
We should definitely mention the new default of 20 for window edge in the changelog for the next release (and maybe how to disable it).

PR #1457, opened 2024-01-21
feat: move / fit to output action

jp7677:
As a follow up to https://github.com/labwc/labwc/discussions/1374#discussioncomment-8070782 I wanted to give "~Send~MoveToOutput" another try.

~I guess I'm quite close but I need some hints how to  finish. Playing with it, my code seems to work for snapped and tiled windows,  but I fail at floating windows. What I had in mind was just moving a floating window to the other output and then position it like it was newly started, thus placed and resized according to placement policy (centered or natural). But somehow I haven't found the correct functions for that. Any hints?~

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -163,6 +163,19 @@ Actions are used in menus and keyboard/mouse bindings.
 	to the center of the window. If the given output does not contain
 	any windows, the cursor is centered on the given output.
 
+*<action name="MoveToOutput" name="HDMI-A-1" direction="value" />*
+	Moves active window to other output, unless the window state is
+	fullscreen.
+
+	If *name* is specified, the window will be sent directly to the specified
+	output and *direction* is ignored. If *name* is omitted, *direction* may
+	be one of "left", "right", "up" or "down" to indicate that the window
+	should be moved to the next output in that direction (if one exists). 
+
+*<action name="FitToOutput" />*
+	Resizes active window size to width and height of the output when the
+	window size exceeds the output size.
+
 *<action name="GoToDesktop" to="value" wrap="yes" />*
 	Switch to workspace.
 
--- a/src/action.c
+++ b/src/action.c
@@ -97,6 +97,7 @@ enum action_type {
 	ACTION_TYPE_SNAP_TO_REGION,
 	ACTION_TYPE_TOGGLE_KEYBINDS,
 	ACTION_TYPE_FOCUS_OUTPUT,
+	ACTION_TYPE_MOVE_TO_OUTPUT,
 	ACTION_TYPE_IF,
 	ACTION_TYPE_FOR_EACH,
 	ACTION_TYPE_VIRTUAL_OUTPUT_ADD,
@@ -148,6 +149,7 @@ const char *action_names[] = {
 	"SnapToRegion",
 	"ToggleKeybinds",
 	"FocusOutput",
+	"MoveToOutput",
 	"If",
 	"ForEach",
 	"VirtualOutputAdd",
@@ -375,6 +377,22 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 	case ACTION_TYPE_FOCUS_OUTPUT:
 		if (!strcmp(argument, "output")) {
 			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
+		break;
+	case ACTION_TYPE_MOVE_TO_OUTPUT:
+		if (!strcmp(argument, "name")) {
+			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
+		if (!strcmp(argument, "direction")) {
+			enum view_edge edge = view_edge_parse(content);
+			if (edge == VIEW_EDGE_CENTER) {
+				wlr_log(WLR_ERROR, "Invalid argument for action %s: '%s' (%s)",
+					action_names[action->type], argument, content);
+			} else {
+				action_arg_add_int(action, argument, edge);
+			}
 			goto cleanup;
 		}
 		break;
@@ -892,6 +910,25 @@ actions_run(struct view *activator, struct server *server,
 						/*update_focus*/ true);
 				}
 			}
+			break;
+		case ACTION_TYPE_MOVE_TO_OUTPUT:
+			if (!view) {
+				break;
+			}
+			const char *name = action_get_str(action, "name", NULL);
+			struct output *target = NULL;
+			if (name) {
+				target = output_from_name(view->server, name);
+			} else {
+				/* Config parsing makes sure that direction is a valid direction */
+				enum view_edge edge = action_get_int(action, "direction", 0);
+				target = view_get_adjacent_output(view, edge);
+			}
+			if (!target) {
+				wlr_log(WLR_ERROR, "Invalid output.");
+				break;
+			}
+			view_move_to_output(view, target);
 			break;
 		case ACTION_TYPE_SNAP_TO_REGION:
 			if (!view) {
--- a/src/action.c
+++ b/src/action.c
@@ -98,6 +98,7 @@ enum action_type {
 	ACTION_TYPE_TOGGLE_KEYBINDS,
 	ACTION_TYPE_FOCUS_OUTPUT,
 	ACTION_TYPE_MOVE_TO_OUTPUT,
+	ACTION_TYPE_FIT_TO_OUTPUT,
 	ACTION_TYPE_IF,
 	ACTION_TYPE_FOR_EACH,
 	ACTION_TYPE_VIRTUAL_OUTPUT_ADD,
@@ -150,6 +151,7 @@ const char *action_names[] = {
 	"ToggleKeybinds",
 	"FocusOutput",
 	"MoveToOutput",
+	"FitToOutput",
 	"If",
 	"ForEach",
 	"VirtualOutputAdd",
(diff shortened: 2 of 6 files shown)


src/view.c
```diff
@@ -1624,6 +1624,56 @@ view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_wind
...
+view_move_to_output(struct view *view, struct output *output)
+{
+	assert(view);
+	if (view->fullscreen) {
+		return;
+	}
+
+	view_invalidate_last_layout_geometry(view);
+	view_set_output(view, output);
+	if (view_is_floating(view)) {
```
jp7677, line 1656: Something is missing here.

src/view.c
```diff
@@ -1624,6 +1624,56 @@ view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_wind
...
+			view_center(view, &box);
+		}
+	} else if (view->maximized != VIEW_AXIS_NONE) {
+		view_apply_maximized_geometry(view);
+	} else if (view->tiled) {
+		view_apply_tiled_geometry(view);
+	} else if (view->tiled_region) {
+		struct region *region = regions_from_name(view->tiled_region->name, output);
+		view_snap_to_region(view, region, /*store_natural_geometry*/ false);
+	}
```
jp7677, line 1674: Not sure if I'm missing other cases here...

src/view.c
```diff
@@ -1624,6 +1624,56 @@ view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_wind
...
+		struct wlr_box box = output_usable_area_in_layout_coords(output);
+		if (rc.placement_policy == LAB_PLACE_AUTOMATIC) {
+			if (placement_find_best(view, &box)) {
+				view_move(view, box.x, box.y);
+			} else {
+				view_center(view, &box);
+			}
+		} else {
+			view_center(view, &box);
+		}
```
ahesford, line 1666: `placement_find_best` needs a box that describes the (expected) `width` and `height` of the window, and it overwrites `x` and `y` with the optimum location, so you need to give it a copy of the `pending` size rather than the full extent of the output.

As for `view_center`, it will center on the entire output when the second argument is `NULL`, so there is no need to figure anything more to provide it.

It might be nice to add an argument to `view_place_initiail` that tells it to skip the under-cursor placement step, so that both the existing `AutoPlace` action and this function can just call, *e.g.*, `view_place_initial(view, /* allow_cursor */ false);`. Otherwise:

```suggestion
		if (rc.placement_policy == LAB_PLACE_AUTOMATIC) {
			struct wlr_box &box = view->pending;
			if (placement_find_best(view, &box)) {
				view_move(view, box.x, box.y);
				return;
			} 
		} else {
			view_center(view, NULL);
		}
```

jp7677: Thanks for the clarification, that helped a lot. I think I got it now, at least from testing the behavior looked good. 
I tried to re-use  `view_place_initial` but unfortunately  that didn't worked together with resize.

jp7677:
PS: I'm not sure if it should be "SendToOutput" or "MoveToOutput". I tend to favor the latter, since this action is feels function-wise closer to e.g. MoveToEdge than to SendToDesktop.

johanmalm:
> PS: I'm not sure if it should be "SendToOutput" or "MoveToOutput". I tend to favor the latter, since this action is feels function-wise closer to e.g. MoveToEdge than to SendToDesktop.

Author's choice I guess :smile:
Agree with your point though... seems closer to `MoveToEdge`

src/action.c
```diff
@@ -893,6 +897,20 @@ actions_run(struct view *activator, struct server *server,
 				}
 			}
 			break;
+		case ACTION_TYPE_MOVE_TO_OUTPUT:
+			{
```
johanmalm, line 901: Brace not needed

docs/labwc-actions.5.scd
```diff
@@ -184,6 +184,13 @@ Actions are used in menus and keyboard/mouse bindings.
 	*wrap* [yes|no] Wrap around from last desktop to first, and vice
 	versa. Default yes.
 
+*<action name="MoveToOutput" to="value" />*
+	Moves active window to other output, unless the window state is fullscreen.
+
+	*to* The direction or the output to send the window to. Supported values
+	are "left", "right", "up", "down" or the name of an output like
+	"HDMI-A-1".
```
ahesford, line 192:
```suggestion
	If *name* is specified, the window will be sent directly to the specified output and *direction* is ignored. If *name* is omitted, *direction* may be one of "left", "right", "up" or "down" to indicate that the window should be moved to the next output in that direction (if one exists). 
```
(This needs rewrapping and uses the "direction" nomenclature I advocate below.)

src/action.c
```diff
@@ -357,14 +359,16 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
...
-			action_arg_add_str(action, argument, content);
-			goto cleanup;
-		}
 		if (!strcmp(argument, "wrap")) {
 			action_arg_add_bool(action, argument, parse_bool(content, true));
 			goto cleanup;
 		}
+		/* Falls through to MoveToOutput */
+	case ACTION_TYPE_MOVE_TO_OUTPUT:
+		if (!strcmp(argument, "to")) {
```
ahesford, line 368: I think "direction" is more in line with other actions that manipulate window relocation. The "to" value is only used for desktop operations and also includes "last" and "current" which run afoul of the directional concept applied here.

If this change is adopted, I recommend changing relevant variables and function arguments to remain consistent.

src/view.c
```diff
@@ -1792,6 +1792,58 @@ view_snap_to_region(struct view *view, struct region *region,
...
+	if (view->fullscreen) {
+		return;
+	}
+
+	view_invalidate_last_layout_geometry(view);
+	view_set_output(view, output);
+	if (view_is_floating(view)) {
+		struct wlr_box geo = output_usable_area_in_layout_coords(output);
+		geo.width = MIN(view->pending.width, geo.width);
+		geo.height = MIN(view->pending.height, geo.height);
```
ahesford, line 1827: I'm not convinced the size should be clipped when relocating outputs. Unlike simple moves, resizing can be potentially "destructive" in the sense that it could cause the window to alter its layout and possibly take other actions.

This might be better served by a new action like "FitToOutput" that will clip the width and height, allowing you to chain "MoveToOutput" and "FitToOutput" if you really want.

src/view.c
```diff
@@ -1792,6 +1792,58 @@ view_snap_to_region(struct view *view, struct region *region,
...
+			if (placement_find_best(view, &geo)) {
+				view_move_resize(view, geo);
+			} else {
+				view_center(view, NULL);
+			}
+		} else {
+			view_center(view, NULL);
+		}
+	} else if (view->maximized != VIEW_AXIS_NONE) {
+		view_apply_maximized_geometry(view);
```
johanmalm, line 1838: It feels like we might be duplicating code here with `view_apply_special_geometry()` and `view_adjust_for_layout_change()`
Could we refactor any of it.
Don't worry if there's nothing obvious, or if it just over-complicates things.

johanmalm (commented):
Looks good with the exception of a couple of comments.

src/view.c
jp7677: Thx, will take a look tomorrow if I can see some options to de-duplicate.

src/view.c
jp7677: thanks, I went to splitting into move and fit and also applied all other suggestions. This turned out pretty good since now I was able to re-use `view_place_initial`.

src/view.c
jp7677: Looking at `view_apply_special_geometry()`, I don't think extracting common things would improve things, there are still subtle differences. `view_adjust_for_layout_change` is also quite different.

src/view.c
```diff
@@ -1792,6 +1792,46 @@ view_snap_to_region(struct view *view, struct region *region,
...
+	} else if (view->tiled) {
+		view_apply_tiled_geometry(view);
+	} else if (view->tiled_region) {
+		struct region *region = regions_from_name(view->tiled_region->name, output);
+		view_snap_to_region(view, region, /*store_natural_geometry*/ false);
+	}
+}
+
+void
+view_fit_to_output(struct view *view)
```
johanmalm, line 1821: This looks similar to `view_constrain_size_to_that_of_usable_area()`. Are we duplicating?

src/view.c
johanmalm: Okay.

johanmalm (commented):
Looks good. Just one question there about view_fit_output()... it feels very similar to another function.

src/view.c
jp7677: shame on me, I was ;) Much better now with re-using that method. I also found a small bug while testing, see the last commit.

jp7677: PS: Just to be sure, `view_constrain_size_to_that_of_usable_area()` won't explode for full screen windows?

johanmalm:
Thanks. And great catching the bug in the constrain function.

src/view.c
jp7677: @johanmalm  I'm thinking that this wouldn't hurt, but dunno if it is really needed:

```diff
diff --git a/src/view.c b/src/view.c
index 4849318..b3f5c90 100644
--- a/src/view.c
+++ b/src/view.c
@@ -747,7 +747,7 @@ view_place_initial(struct view *view, bool allow_cursor)
 void
 view_constrain_size_to_that_of_usable_area(struct view *view)
 {
-       if (!view || !view->output) {
+       if (!view || !view->output || view->fullscreen) {
                return;
        }
 

```

Consolatis: ^ merged via
- #1464

Issue #1458, opened 2024-01-21
Wlroots crash due to division by zero when setting custom mode for DRM with refreshrate of 0


Consolatis:
> \<consolers\> `wlr-randr --output eDP-1 --custom-mode 1620x900` used to work with labwc and an earlier version of wlr-randr but now it crashes it
\<consolers\> [..] `0x0000000000464d60 in mhz_to_nsec (mhz=0) at ../subprojects/wlroots/backend/drm/drm.c:1696 return 1000000000000LL / mhz;`
\<consolers\> calling `wlr-randr --output eDP-1 --custom-mode 1600x900@60` with the 60mhz causes it not to crash

The refresh rate seems indeed not be validated anywhere and just carried along until it crashes in `mhz_to_nsec()` while preparing the present event. Did not search the wlroots repo for an existing issue. Also didn't try to reproduce.

We could just tape over it by checking if the output is a DRM output + refresh rate is 0 when applying a custom mode but this is something that we should really get fixed upstream in wlroots. Maybe we should do both though as not everybody runs the latest wlroots 0.17.x.

Consolatis:
Thanks for the pointers!
That PR is also marked for backport so with a future wlroots minor release this should be fixed from their side.

We should still catch this in labwc though, guess we could use `verify_output_config_v1()` for that.

PR #1461, opened 2024-01-22
Implement proper output configuration test and expose it to clients

Consolatis:
Only tested nested.
Also exposed the test routine to clients while at it, `wlr-randr --dryrun` can be used to verify if it works.

We were previously completely ignoring the `test` signal which requires compositors to destroy the supplied config object. As we handle it now this also fixes a memory leak.

Should fix
- #1458

even without the wlroots patch applied.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -292,6 +292,7 @@ struct server {
 
 	struct wl_listener output_layout_change;
 	struct wlr_output_manager_v1 *output_manager;
+	struct wl_listener output_manager_test;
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM. Not tested.

Nice bug fix too :smile:

Can test on metal tomorrow if you want. Otherwise feel free to merge.

Consolatis:
Some test on DRM would be great.

johanmalm:
Have tested on DRM. Works well. Can reproduce bug (before patch). Will go for merge.

Issue #36, opened 2021-06-08
[Question] Display Manager


speltriao:
I installed labwc  using the AUR. By default, my display manager (lightdm) doesn't show an option to labwc.

How can I login into labwc with lightdm? 

Thanks for the project.

johanmalm:
Hi
I’ve not spent much time with Display Manager.
Do these instructions help?
https://github.com/johanmalm/labwc/wiki/Display-Manager

speltriao:
Yep! That Works!

If I might make a suggestion, It would be nice to have a link to this wiki article on the ReadMe.

johanmalm:
I've added a labwc.desktop file to the repository and included it in meson.build to be installed automatically.

PR #1462, opened 2024-01-22
meson.build: install badges for labwc and migrate to data/ 

rcalixte:
A .desktop file was previously added for the labwc session, this adds a
badge for it where it is supported. The .desktop files and the badges
are also moved to `data/`.

Ref: #36, 9fa783e, labwc/labwc-artwork#7

An example from lightdm:

Before:

After:

### DIFF
--- a/data/labwc-symbolic.svg
+++ b/data/labwc-symbolic.svg
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--labwc logo #2 (C) Johan Malm LICENSE: CC BY-SA 4.0-->
+<svg xmlns="http://www.w3.org/2000/svg" width="256px" height="256px"
+     viewbox="0 0 256 256"
+     stroke-linecap="round" stroke-linejoin="round" stroke-width="12">
+  <path fill="#ffffff" stroke="#ffffff" d="m 26 68 91 55 v 108 l -76 -61 z" />
+  <path fill="#000000" stroke="#000000" d="m 229 68 -91 55 v 108 l 76 -61 z" />
+</svg>
--- a/data/labwc.desktop
+++ b/data/labwc.desktop
@@ -2,5 +2,6 @@
 Name=labwc
 Comment=A wayland stacking compositor
 Exec=labwc
+Icon=labwc
 Type=Application
 DesktopNames=wlroots
--- a/data/labwc.svg
+++ b/data/labwc.svg
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--labwc logo #2 (C) Johan Malm LICENSE: CC BY-SA 4.0-->
+<svg xmlns="http://www.w3.org/2000/svg" width="256px" height="256px"
+     viewbox="0 0 256 256"
+     stroke-linecap="round" stroke-linejoin="round" stroke-width="12">
+  <path fill="#f0d70f" stroke="#f0d70f" d="m 26 68 91 55 v 108 l -76 -61 z" />
+  <path fill="#d02f90" stroke="#d02f90" d="m 229 68 -91 55 v 108 l 76 -61 z" />
+</svg>
--- a/meson.build
+++ b/meson.build
@@ -134,4 +134,10 @@ executable(
   install: true,
 )
 
-install_data('docs/labwc.desktop', install_dir: get_option('datadir') / 'wayland-sessions')
+install_data('data/labwc.desktop', install_dir: get_option('datadir') / 'wayland-sessions')
+
+icons = ['labwc-symbolic.svg', 'labwc.svg']
+foreach icon : icons
+  icon_path = join_paths('data', icon)
+  install_data(icon_path, install_dir: get_option('datadir') / 'icons/hicolor/scalable/apps')
+endforeach


redtide:
Hello, out of interest, where these badge icons are used? Is there some kind of specification or link to some use case? It's new to me, TIA.

Consolatis:
Personally I'd prefer the colored one
wlroots

It [seems](https://wiki.archlinux.org/title/display_manager#Session_configuration) we should also add it to our `labwc.desktop` file as `Icon` attribute.

rcalixte:
> Hello, out of interest, where these badge icons are used? Is there some kind of specification or link to some use case? It's new to me, TIA.

That's my bad. I meant to include screenshots. Let me edit in a few minutes.




> Personally I'd prefer the colored one wlroots
> 
> It [seems](https://wiki.archlinux.org/title/display_manager#Session_configuration) we should also add it to our `labwc.desktop` file as `Icon` attribute.

It only supports grayscale unfortunately.

Consolatis:
> It only supports grayscale unfortunately.

Sorry, but what exactly is only supporting grayscale?

rcalixte:
> > It only supports grayscale unfortunately.
> 
> Sorry, but what exactly is only supporting grayscale?

I've attached the screenshots.

rcalixte:
> > It only supports grayscale unfortunately.
> 
> Sorry, but what exactly is only supporting grayscale?

Now that I look at it closer, it actually only seems monochromatic.

rcalixte:
I'm now noticing that I'm not seeing consistent results when I'm compiling locally unless I don't pass a prefix to `meson`. With `--prefix=/usr`, it didn't seem to work correctly no matter how I modified the .desktop file. Somewhat upsetting.

johanmalm:
If we upload the original (https://github.com/labwc/labwc-artwork/blob/main/logos/labwc2.svg), does lightdm turn that monochrome?

rcalixte:
> If we upload the original (https://github.com/labwc/labwc-artwork/blob/main/logos/labwc2.svg), does lightdm turn that monochrome?

Yes. That's what I used for the source.

rcalixte:
Here's [the source](https://git.launchpad.net/ubuntu/+source/lightdm-gtk-greeter/tree/data/badges) for the `lightdm-gtk-greeter` package managed by Canonical.

rcalixte:
> Here's [the source](https://git.launchpad.net/ubuntu/+source/lightdm-gtk-greeter/tree/data/badges) for the `lightdm-gtk-greeter` package managed by Canonical. That provides the examples that I used for the basis here. It would probably be beneficial to have the badge for labwc added there as well.

johanmalm:
> Here's [the source](https://git.launchpad.net/ubuntu/+source/lightdm-gtk-greeter/tree/data/badges) for the `lightdm-gtk-greeter` package managed by Canonical. That provides the examples that I used for the basis here. It would probably be beneficial for the badge for labwc to be added there as well.

Fine with me. Fancy submitting a patch / pull-request to that project?

Consolatis:
I don't get it though. We do they ship a random collection of icons rather than just using the `Icon` attribute?

rcalixte:
> Fine with me. Fancy submitting a patch / pull-request to that project?

I'll look into it. It's unusual that there are so many missing, including from official Flavours. I wonder if they're expecting upstreams to package it while they just provide what they consider the core set.




> I don't get it though. We do they ship a random collection of icons rather than just using the `Icon` attribute?

It's a hard-coded lookup definition: [source code](https://git.launchpad.net/ubuntu/+source/lightdm-gtk-greeter/tree/src/lightdm-gtk-greeter.c#n1071). (I couldn't find appropriate documentation, but code is the best documentation sometimes. 😅)

rcalixte:
Is someone able to test building this locally to see if my issues are my own or if something needs to be adjusted? I'll keep testing on my end as well and rebase if I find any changes to make. I can also change the icon to a monochromatic one if that would help too.

Consolatis:
> > I don't get it though. We do they ship a random collection of icons rather than just using the `Icon` attribute?
>
> It's a hard-coded lookup definition: [source code](https://git.launchpad.net/ubuntu/+source/lightdm-gtk-greeter/tree/src/lightdm-gtk-greeter.c#n1071). (I couldn't find appropriate documentation, but code is the best documentation sometimes. 😅)

Well that sounds like something to fix for lightdm then, I don't think that we should ship a `labwc_badge-symbolic` icon just to work around an issue in lightdm. Completely fine with installing the usual `labwc.svg` icon and adding it to `labwc.deskop` though.

rcalixte:
> Well that sounds like something to fix for lightdm then, I don't think that we should ship a `labwc_badge-symbolic` icon just to work around an issue in lightdm.

The issue is that lightdm doesn't actually provide a greeter by default. Here are some of the supported greeters: [LightDM - ArchWiki](https://wiki.archlinux.org/title/LightDM)

> Completely fine with installing the usual `labwc.svg` icon and adding it to `labwc.deskop` though.

Is this confirmed to work appropriately?

Consolatis:
> > Completely fine with installing the usual `labwc.svg` icon and adding it to `labwc.deskop` though.
> 
> Is this confirmed to work appropriately?

With lightdm? No. They'd need to ship their own `labwc_badge-symbolic` icon or update their code.
I don't know if other login / display managers support the `Icon` value from the session `.desktop` file.
Based on the arch wiki they seem to but I didn't verify that.

Consolatis:
Openbox also seems to add it via `Icon` attribute: https://github.com/Mikachu/openbox/blob/master/data/openbox.desktop.

Opened an issue in the artwork repo to discuss the icon itself:
- https://github.com/labwc/labwc-artwork/issues/7

rcalixte:
> Openbox also seems to add it via `Icon` attribute: https://github.com/Mikachu/openbox/blob/master/data/openbox.desktop.
> 
> Opened an issue in the artwork repo to discuss the icon itself:
> 
> * [Labwc icon to ship within the main repo for its labwc.desktop session file labwc-artwork#7](https://github.com/labwc/labwc-artwork/issues/7)

I wanted to formulate this better for later but I'll share what I'm thinking for now. First, I understand the spirit of what is trying to be accomplished here. I would caution against referring to the Openbox .desktop file since it is over 12 years old. I would instead point you to either [Sway](https://github.com/swaywm/sway/blob/master/sway.desktop) or [Hyprland](https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.desktop)'s implementations that work. I could pick from a few others as well but the point is hopefully clear.

What I am proposing and can realistically get done today (since I fortunately still have the source files for this pull request) is that we create an assets directory, again similar to [Sway](https://github.com/swaywm/sway/tree/master/assets) and [Hyprland](https://github.com/hyprwm/Hyprland/tree/main/assets) as the docs folder is optimally not the best place for these files. The assets directory for now would contain a 64x64 color labwc.svg and a 64x64 monochrome labwc_badge-symbolic.svg. It would also be best practice to include labwc-symbolic.svg that is a copy of labwc.svg as well. If there is a desire to also ship .png files, the convention is typically to scale up from 16x16 to 256x256. I have a preference for the svg files since those are scalable and supported just the same but this is not my project and not my call. I also think this is the best user experience as it would be great to have the icons available for use in things like Waybar or Yambar or any other bars/greeters that comprise the desktop.

The tldr proposal that I can rebase and test today:

```bash
assets/
├── labwc_badge-symbolic.svg
├── labwc.svg
└── labwc-symbolic.svg
```

ahesford:
The case for a "_badge-symbolic" graphic would be a lot stronger if you can demonstrated that it is used somewhere other than lightdm. I don't think adding special icons every time a display manager wants to do its own thing is a worthwhile effort. If there's an independent case to be made for, *e.g.*, a monochrome version of the icon, it should probably be named more descriptively than "badge-symbolic" and lightdm users can put the appropriate icon in the appropriate place.

redtide:
> I would caution against referring to the Openbox .desktop file since it is over 12 years old. I would instead point you to either [Sway](https://github.com/swaywm/sway/blob/master/sway.desktop) or [Hyprland](https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.desktop)'s implementations that work.

Well, the fact that it's old doesn't mean that it's wrong since it respects the [FreeDesktop desktop entry specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/), and I wonder which specification is behind this proposition other than why Icon should not be used in the desktop file; I tested some other screen managers and it seems it has to do only with lightdm, not labwc.

I agree about SVG, wasn't the discussion about that format?

ahesford:
Just as it is unreasonable to expect that every greeter manage icons for every session, it is unreasonable to expect projects like labwc to manage icons for every possible display manager configuration. I advocate against falling down the rabbit hole of including icons for the GTK greeter for lightdm, and later its WebKit greeter, and later GDM, and later SDDM, and later... This is where standards come in. Projects design icons and put them in a common place for all consumers to find.

An appellation like "symbolic" to describe a grayscale image is a strange choice. Grayscale icons should probably just be called "grayscale" unless there is some common standard that uses this nomenclature.

rcalixte:
Just to summarize what has become a long thread, I can rebase and test with the color labwc.svg and the monochrome labwc-symbolic.svg. Whether it works with LightDM or not is out of scope. These files will be placed into an `assets` (would something like `data` be preferred?) top-level directory and installed to the system in `/usr/share/icons/hicolor/scalable/apps`.

The proposed structure:
```bash
assets/
├── labwc.svg
└── labwc-symbolic.svg
```

If this is agreeable, I can probably get this done in short order.

redtide:
> Regarding the Icon key directly, leaving the entry undefined means the display managers will use whatever convention works best. A hard-coded configuration here is suboptimal.

this doesn't make any sense to me, the icon have the application id, which is also what appstream (and so dbus) uses, so wth screen managers should do something outside the standard it's their problem.

So after:
"A hard-coded configuration here is suboptimal" (Icon property it's in the spec)
and "There is a common standard that uses this nomenclature" can we know which standard is it and where is specified?

rcalixte:
> So after: "A hard-coded configuration here is suboptimal" (Icon property it's in the spec)

It's listed as not required in the spec. When undefined, the previously mentioned Icon Theme spec will be used to determine whatever optimal icon is preferred. This is an optimal configuration depending on whether the client prefers a color or monochromatic icon, rather than forcing one that may not be compatible with all clients.

> and "There is a common standard that uses this nomenclature" can we know which standard is it and where is specified?

I searched the FreeDesktop specs and the best I could find was "read GNOME's source" which didn't seem like an appropriate link. That was why I suggested checking your current machine for instances of those files and their contents. Until proper documentation can be found, that should suffice.

rcalixte:
This feels like another tangent. I can just append `Icon=labwc` to the .desktop file is that is what is preferred.

ahesford:
> There is a common standard that uses this nomenclature. 😅 (If you run something like `locate -- -symbolic.svg` on your local machine (assuming some variant of Linux here), you will find them.) From my current machine:
> 
> ```shell
> $ locate -- -symbolic.svg | wc -l
> 9496
> ```

I too have many of these files. They all seem to be "action" icons, not application icons. That is, they seem to be for things that end up in toolbars.

1. You haven't pointed to a standard, and "just read GNOME source" isn't any better than just reading lightdm source. Both are the wrong way to decide what icons to provide and how to name them.
2. The fact that lightdm is abusing "symbolic" icon nomenclature because it wants to show application icons in its application chooser menu doesn't mean that it is making a sensible choice, or even one that we should accommodate.
3. Prevalence of these files notwithstanding, I still have no idea what notion "symbolic" is trying to convey. This is why I want a standard to read---well written standards tend to explain *why* things are done, not just *how*.

redtide:
> I too have many of these files. They all seem to be "action" icons, not application icons. That is, they seem to be for things that end up in toolbars.

I have them all in [icons/hicolor/scalable/places](https://sfz.altervista.org/images/badge-symbolic-icons.png)

> Apparently there is no spec for naming them from what I gathered in this thread

labwc(.svg) looks good.
In the desktop files the [icon name](https://sfz.altervista.org/images/icon-names.png) is usually the same as the application executable or [app id](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic), this is also in the [desktop entry spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html).

rcalixte:
I appreciate all of the constructive feedback. All of the requested changes are in the latest commit. I sourced the .svg file from the attached Issue and created the symbolic version using the instructions [here](https://wiki.gnome.org/HowDoI/CreateSymbolicIconsThatChangeColorAccordingToTheme).

Consolatis:
@rcalixte Thanks. I'll mark this PR as draft until we decided on the actual icon to use. I still think that the GH orga icon looks better because it offsets the cube a bit to the button to also include the implied invisible top side.

There is also still the question raised by @redtide of where to place the icons, in `scalable/apps/` or `scalable/places/`

rcalixte:
> There is also still the question raised by @redtide of where to place the icons, in `scalable/apps/` or `scalable/places/`

The type specified in the .desktop file is Application which is why I opted for apps. I believe 'places' is for things like folder icons, paths, and locations. The badges themselves are a bit out of place there but they've been that way for a while so it's likely going to stick.

redtide:
> I sourced the .svg file from the attached Issue and created the symbolic version using the instructions [here](https://wiki.gnome.org/HowDoI/CreateSymbolicIconsThatChangeColorAccordingToTheme).

I see there the examples seems to point all in the `actions` directory, what I pointed out was from the **entire** `icons` directory in my system and there are no badge icons anywhere outside that directory

> The type specified in the .desktop file is Application which is why I opted for apps.

This sound strange, according to the [specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html) there are only 3 types of files, Application (type 1), Link (type 2) and Directory and almost all desktop files I saw until now was about Application, so I wonder where `places` should belong using that logic; opted between which options?

rcalixte:
> I see there the examples seems to point all in the `actions` directory, what I pointed out was from the **entire** `icons` directory in my system and there are no badge icons anywhere outside that directory

The files in your screenshot appear to be from the lightdm-gtk-greeter package. The linked documentation is just an example application. I stopped at the section "Create an icon" as that is what covers the task at hand.

> This sound strange, according to the [specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html) there are only 3 types of files, Application (type 1), Link (type 2) and Directory and almost all desktop files I saw until now was about Application, so I wonder where `places` should belong using that logic; opted between which options?

Again, these are options made by other projects long ago. Between `places` and `apps`, `apps` feels more in line with the modern scoping. If you look at the other contents of `places`, you may see other files that fall within the description I provided previously. The outlier among these is the one package of lightdm-gtk-greeter that deposits badges in that location. You can verify this against other packages. (Checking all of Debian Trixie's packages, `dwm` deposits a badge into `places` and `enlightenment-data` deposits a badge into both `apps` and `places`. I can attach the full list of packages for Debian Trixie that use `places` if desired but this feels like another tangent. For reference, there are 9 total packages in `places` while `apps` is 924 for Debian Trixie's entire repository.)

Consolatis:
Added this to the milestone for 0.7.1. Logo discussion in
- https://github.com/labwc/labwc-artwork/issues/7

johanmalm:
Oh, wow you're quick!

LGTM

This is very good. Sorry for not looking at this earlier.
Thanks to @Consolatis for helping me navigate this. And @Domo131 of course for working on the icon too.

Consolatis:
Thanks @rcalixte for the PR and keeping up with our slightly chaotic review in this case. Distro packagers might still need to add a symlink for the `badge` variant if this is applied to make it work with lightdm but at least they now have an icon to work with :)

rcalixte:
> Thanks @rcalixte for the PR and keeping up with our slightly chaotic review in this case. Distro packagers might still need to add a symlink for the `badge` variant if this is applied to make it work with lightdm but at least they now have an icon to work with :)

I appreciate the work being done here! And yeah, I hope distros can do the light work needed on that end. I've also taken the opportunity to add the labwc.svg to the Waybar config I'm using. (I opted for the color SVG for now but this is in flux. 😅)

johanmalm:
> I appreciate the work being done here! And yeah, I hope distros can do the light work needed on that end. I've also taken the opportunity to add the labwc.svg to the Waybar config I'm using. (I opted for the color SVG for now but this is in flux. 😅)

Nice. We enjoy a screenshot if you're happy to share.

rcalixte:
> Nice. We enjoy a screenshot if you're happy to share.

Happy to share. My first rice I guess. 😅 (This is a work in progress that I'm enjoying as an exercise.)

PR #1463, opened 2024-01-22
view, xdg: notify clients when tiling windows

ahesford:
Clients may wish to know if they are "tiled" and change their sizing behavior accordingly. For example, if https://codeberg.org/dnkl/foot/pulls/1580 is merged, `foot` will insist by default that its size be a multiple of the character cell size unless a window is maximized, fullscreened or tiled. Without notifying the client that the window is tiled, this will cause strange artifacts like a snapped-to-edge window (especially on the right) pulling away from the edge a bit. Clients like `mpv` also seem to care when they are tiled; currently, snapping `mpv` to the edge causes it to render in one of the corners, but after this change, `mpv` will take up the requested region when snapping and render letterbox/pillarbox regions as needed.

I'm not sure if I've caught all cases properly, so this needs to be better tested. Suggestions for alternative implementations are welcome.

Under the hood, wlroots just sends a configure event to the client with a `tiled` flag set. wlroots does not expose a similar function for Xwayland, so I'm unsure if there is any action that should be taken for Xwayland views.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -236,16 +236,44 @@ this is for compatibility with Openbox.
 
 ## WINDOW SNAPPING
 
-The following two options relate to triggering window actions when moving
-windows using the mouse.
+Windows may be "snapped" to an edge or user-defined region of an output when
+activated with SnapToEdge actions or, optionally, by dragging windows to the
+edges of an output. Edge snapping causes a window to occupy half of its output,
+extending outward from the snapped edge.
 
 *<snapping><range>*
-	The distance in pixels from the edge of an output for window Move
-	operations to trigger SnapToEdge. A range of 0 disables window snapping.
-	Default is 1.
+	If an interactive move ends with the cursor a maximum distance *range*,
+	(in pixels) from the edge of an output, the move will trigger a
+	SnapToEdge action for that edge. A *range* of 0 disables snapping via
+	interactive moves. Default is 1.
 
 *<snapping><topMaximize>* [yes|no]
-	Maximize window if Move operation ends on the top edge. Default is yes.
+	If *yes*, an interactive move that snaps a window to the top edge will
+	maximize the window. If *no*, snapping will behave as it does with other
+	edges, causing the window to occupy the top half of an output. Default
+	is yes.
+
+*<snapping><notifyClient>* [always|region|edge|never]
+	Snapping windows can trigger corresponding tiling events for native
+	Wayland clients. Clients may use these events to alter their rendering
+	based on knowledge that some edges of the view are confined to edges of
+	a snapping region or output. For example, rounded corners may become
+	square when tiled, or media players may letter-box or pillar-box video
+	rather than imposing rigid aspect ratios on windows that will violate
+	the constraints of window snapping.
+
+	- When *always* is specified, any window that is snapped to either an
+	  output edge or a user-defined region will receive a tiling event.
+
+	- When *region* is specified, only windows snapped to a user-defined
+	  region will receive an event.
+
+	- When *edge* is specified, only windows snapped to an output edge will
+	  receive an event.
+
+	- When *never* is specified, tiling events will never be triggered.
+
+	The default is "always".
 
 ## REGIONS
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -78,10 +78,11 @@
     <raiseOnFocus>no</raiseOnFocus>
   </focus>
 
-  <!-- Set range to 0 to disable window snapping completely -->
   <snapping>
+    <!-- Set range to 0 to disable window snapping completely -->
     <range>1</range>
     <topMaximize>yes</topMaximize>
+    <notifyClient>always</notifyClient>
   </snapping>
 
   <!--
--- a/src/view.c
+++ b/src/view.c
@@ -1051,6 +1051,10 @@ view_set_untiled(struct view *view)
 	view->tiled = VIEW_EDGE_INVALID;
 	view->tiled_region = NULL;
 	zfree(view->tiled_region_evacuate);
+
+	if (view->impl->set_tiled) {
+		view->impl->set_tiled(view);
+	}
 }
 
 void
@@ -1771,6 +1775,11 @@ view_snap_to_edge(struct view *view, enum view_edge edge,
 	view_set_untiled(view);
 	view_set_output(view, output);
 	view->tiled = edge;
+
+	if (view->impl->set_tiled) {
+		view->impl->set_tiled(view);
+	}
+
 	view_apply_tiled_geometry(view);
 }
 
@@ -1804,6 +1813,11 @@ view_snap_to_region(struct view *view, struct region *region,
 	}
 	view_set_untiled(view);
 	view->tiled_region = region;
+
+	if (view->impl->set_tiled) {
+		view->impl->set_tiled(view);
+	}
+
 	view_apply_region_geometry(view);
 }
 
--- a/src/view.c
+++ b/src/view.c
@@ -1043,6 +1043,15 @@ view_is_floating(struct view *view)
 		|| view_is_tiled(view));
 }
 
+static void
+view_notify_tiled(struct view *view)
+{
+	assert(view);
+	if (view->impl->notify_tiled) {
+		view->impl->notify_tiled(view);
+	}
+}
+
 /* Reset tiled state of view without changing geometry */
 void
 view_set_untiled(struct view *view)
@@ -1051,10 +1060,7 @@ view_set_untiled(struct view *view)
 	view->tiled = VIEW_EDGE_INVALID;
 	view->tiled_region = NULL;
 	zfree(view->tiled_region_evacuate);
-
-	if (view->impl->set_tiled) {
-		view->impl->set_tiled(view);
-	}
+	view_notify_tiled(view);
 }
 
 void
@@ -1775,11 +1781,7 @@ view_snap_to_edge(struct view *view, enum view_edge edge,
 	view_set_untiled(view);
 	view_set_output(view, output);
 	view->tiled = edge;
-
-	if (view->impl->set_tiled) {
-		view->impl->set_tiled(view);
-	}
-
+	view_notify_tiled(view);
 	view_apply_tiled_geometry(view);
 }
 
@@ -1813,11 +1815,7 @@ view_snap_to_region(struct view *view, struct region *region,
 	}
 	view_set_untiled(view);
 	view->tiled_region = region;
-
-	if (view->impl->set_tiled) {
-		view->impl->set_tiled(view);
-	}
-
+	view_notify_tiled(view);
 	view_apply_region_geometry(view);
 }
 
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -423,6 +423,41 @@ xdg_toplevel_view_set_fullscreen(struct view *view, bool fullscreen)
 		fullscreen);
 }
 
+static void
+xdg_toplevel_view_set_tiled(struct view *view)
+{
+	enum wlr_edges edge;
+
+	/*
+	 * Edge-snapped view are considered tiled on the snapped edge and those
+	 * perpendicular to it.
+	 */
+	switch (view->tiled) {
+	case VIEW_EDGE_LEFT:
+		edge = WLR_EDGE_LEFT | WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
+		break;
+	case VIEW_EDGE_RIGHT:
+		edge = WLR_EDGE_RIGHT | WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
+		break;
+	case VIEW_EDGE_UP:
+		edge = WLR_EDGE_TOP | WLR_EDGE_LEFT | WLR_EDGE_RIGHT;
+		break;
+	case VIEW_EDGE_DOWN:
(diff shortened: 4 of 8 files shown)


src/xdg.c
```diff
@@ -400,6 +400,12 @@ xdg_toplevel_view_set_fullscreen(struct view *view, bool fullscreen)
 		fullscreen);
 }
 
+static void
+xdg_toplevel_view_set_tiled(struct view *view, bool tiled)
+{
+	wlr_xdg_toplevel_set_tiled(xdg_toplevel_from_view(view), tiled);
```
Consolatis, line 406: `subprojects/wlroots/include/wlr/types/wlr_xdg_shell.h`:
```c
/**
 * Request that this toplevel consider itself in a tiled layout and some
 * edges are adjacent to another part of the tiling grid. `tiled_edges` is a
 * bitfield of enum wlr_edges. Returns the associated configure serial.
 */
uint32_t wlr_xdg_toplevel_set_tiled(struct wlr_xdg_toplevel *toplevel,
                uint32_t tiled_edges);
```

So we'd need to convert `view->tiled` to that bitset.
Another question is how / if we should handle `view->tiled_region` here.

src/xwayland.c
```diff
@@ -797,6 +797,7 @@ static const struct view_impl xwayland_view_impl = {
 	.map = xwayland_view_map,
 	.set_activated = xwayland_view_set_activated,
 	.set_fullscreen = xwayland_view_set_fullscreen,
+	.set_tiled = NULL,
```
Consolatis, line 800: We don't set unsupported methods to `NULL` for the other cases.

src/xdg.c
johanmalm: https://github.com/swaywm/sway/blob/e8c421e917ac59cdf50e5fd366155ea875e4a29e/sway/desktop/xdg_shell.c#L181-L185

Consolatis: Especially with `SnapToEdge` I think we could be a bit smarter here (e.g. only set the 3 edges we are actually snapped to).

johanmalm:
I like this. I had previously thought that this just toggled CSD corners between rounded and square, and as I don't used CSD I had forgotten about it.
The use-case described in OP is a good reason for doing this. It might solve my xfce4-panel issue to (which doesn't quite size right with SnapToEdge). Will test that another day.
Regardless, we ought to honour this anyway for those that like to use CSD.

Consolatis:
Agree on setting the edges as well, but maybe we should provide a config option if it should require `rc.gap` being set to `0`.
I assume some people using `rc.gap` may actually want their rounded corners when snapped.

johanmalm:
Does the sizing issue go away even if we set `WLR_EDGE_NONE`? If so, @Consolatis makes a v good point about consistency with SSD in that we ought to treate the rounded/square corner thing separately (if at all)

src/xdg.c
ahesford: Welp, I completely missed that this was not a Boolean.

I restructured this to set all edges tiled for region-snapped views, but only set the three snapped edges for edge-snapped views. This seems to do the right thing for foot (which doesn't actually care about the tiling direction.)

ahesford:
This sizing issue will not be fixed by `WLR_EDGE_NONE`; wlroots uses that flag to signal to clients that the window is not tiled.

johanmalm:
> This sizing issue will not be fixed by `WLR_EDGE_NONE`; wlroots uses that flag to signal to clients that the window is not tiled.

I’m fine with this as is then.

LGTM. Not tested.

ahesford:
I propose a change in behavior that simplifies some state tracking and solves the issue noted by @Flrian: as soon as an interactive move begins om `interactive_begin`, set the window untiled. This will notify the client that it is untiled (which agrees with its visual state) and also mean that we don't have to clean things up (or deal with inconsistencies) in `interactive_end` or `interactive_cancel`. However, it means that dragging to snap the window, then dragging to maximize, and finally un-maximizing (by action) will leave the window un-tiled at its natural geometry (set before the original snap) rather than back to the snapped position. (Maximizing directly from a snapped state by action will still return the window back to the snapped state after un-maximiziing.)

I can see the purpose of preserving the tiling state in this case, but I'd also argue that restoring the window back to its natural position after dragging to maximize is somewhat intuitive: the user has already visually seen the window shrink back to its natural size during the drag, so the mental model should convey that it is back to its natural geometry before the drag is complete.

If we don't adopt this change in behavior, we either live with the visual artifacts caused by letting the client continue to think it is tiled, or have to jump through several more hoops to notify the client during `interactive_begin` and again in `interactive_end` (also `interactive_cancel`, or else live with the de-sync) about the transient state.

As for preserving rounded corners, I am not opposed to adding a config flag that declines to notify the client when tiled. That lets users choose between letting the client draw how it wants when tiled, or forcing the rounded corners and potentially changing what the client draws in some cases.

ahesford:
I've added `<snapping><xdgTiling>no</xdgTiling></snapping>` to disable the tiling notifications, which should preserve untiled CSD even as we resize and relocate windows with snapping.

ahesford:
I renamed the option `notifyClient` and turned it into an enum for finer control.

ahesford:
If somebody wants a "reset-interactive-move" action, I think that is better implemented by capturing all relevant window state (full screen, maximized, tiling) in `interactive_begin` and a) invalidating it within `interactive_finish` or b) re-applying it properly in `interactive_cancel`. That way, we can fully restore the window to its pre-interactive condition, rather than inconsistently retaining only part of the information in a way that de-syncs the model of the window from its actual behavior during dragging.

Consolatis:
> If somebody wants a "reset-interactive-move" action, I think that is better implemented by capturing all relevant window state (full screen, maximized, tiling) in `interactive_begin` and a) invalidating it within `interactive_finish` or b) re-applying it properly in `interactive_cancel`. That way, we can fully restore the window to its pre-interactive condition, rather than inconsistently retaining only part of the information in a way that de-syncs the model of the window from its actual behavior during dragging.

Good point. Agree.

docs/labwc-config.5.scd
```diff
@@ -236,16 +236,41 @@ this is for compatibility with Openbox.
...
 
 *<snapping><topMaximize>* [yes|no]
-	Maximize window if Move operation ends on the top edge. Default is yes.
+	If *yes*, an interactive move that snaps a window to the top edge will
+	maximize the window. If *no*, snapping will behave as it does with other
+	edges, causing the window to occupy the top half of an output. Default
+	is yes.
+
+*<snapping><notifyClient>* [both|region|edge|none]
+	Snapping windows can trigger corresponding tiling events for xdg-shell
```
Consolatis, line 257: Suggest `s/xdg-shell/wayland native/`

docs/labwc-config.5.scd
```diff
@@ -236,16 +236,41 @@ this is for compatibility with Openbox.
...
+	clients. Clients may use these events to alter their rendering based on
+	knowledge that some edges of the view are confined to edges of a
+	snapping region or output.
```
Consolatis, line 260: Maybe we could add something like `for example by making the corners square` here.

docs/labwc-config.5.scd
```diff
@@ -236,16 +236,41 @@ this is for compatibility with Openbox.
...
+
+	- When *region* is specified, only windows snapped to a user-defined
+	  region will receive an event.
+
+	- When *edge* is specified, only windows snapped to an output edge will
+	  receive an event.
+
+	- When *none* is specified, tiling events will never be triggered.
+
+	The default is "both".
```
Consolatis, line 273: I am not sure about the default value. Would be great if we could get some feedback here @Flrian @heroin-moose.

Especially with `rc.gap` set suddenly having square corners for CSD applications because of an labwc update might feel strange.

Consolatis: Currently this PR is enabling the tiling notifications even if `rc.gap != 0`.

ahesford:
Updated the docs for clarity.

We can change the default `notifyClient` if people want. I picked `both` as the default because it unbreaks functional behavior. For example snapping mpv to the bottom of a window will cause it to resize the video to fill the entire width and, without tiling notifications, push half of the video below the screen; with tiling notifications, mpv will resize properly and pillar-box its video. Terminals that resize in fixed increments but allow dynamic padding when tiled also produce unexpected gaps when `notifyClient` is disabled. My preference is to ensure proper behavior over corner rounding.

As for what happens if we fix SSD corners for tiling, I guess I'd suggest we just silently piggyback on `notifyClient` to emulate what CSD might be doing. The option could be called something else, but I think the most relevant feature of the option is that it tells clients they are tiled to let them draw differently. I'm not sure there are other decent names that would convey this key point as well indicating that SSD may be altered.

Consolatis:
Its unfortunate that both things are using the same event. The mpv / foot issues are a nobrainer and we should fix them. However, that also makes GTK corners square which may destroy the expected look and feel of the desktop for some users. 

Being able to configure them goes a long way already to satisfy both cases. So maybe @Flrian is right and instead of looking at the existing users here we should be looking at new users and make it very clear in the changelog that this may cause a regression for some users, why it is being done and how they can fix it if they want to (thus accepting the mpv / foot issues).

The inconsistency of SSD and CSD is a minor thing for me. Its not great but we don't have to fix that before we can merge this PR in my opinion. We should have the basic building blocks available as we already unround SSD corners on maximize.

docs/rc.xml.all
```diff
@@ -78,10 +78,11 @@
     <raiseOnFocus>no</raiseOnFocus>
   </focus>
 
-  <!-- Set range to 0 to disable window snapping completely -->
   <snapping>
+    <!-- Set range to 0 to disable window snapping completely -->
     <range>1</range>
     <topMaximize>yes</topMaximize>
+    <notifyClient>both</notifyClient
```
Consolatis, line 85: missing `>` at the end.

Random thought:
```xml
<notifyClient>always</notifyClient>
```
rather than `both` ?

ahesford: Good suggestions. Adopted (and syntax fixed) in the latest version.

ahesford:
> Another observation: When snapping Nautilus to an edge or a region I can no longer resize the window via the border, only via Alt + Right dragging, which un-tiles the window.
> 
> (In GNOME when trying to resize via the border when snapped to an edge I can actually only do it horizontally and it stays tiled while doing that, which is really nice)

The loss of a resize cursor is unfortunate. I confirmed the issue with a local Nautilus installation, but don't have any other CSD applications that seen to care about tiling. (Both gedit and simple-scan are tiled without even dropping rounded corners, and neither one suppresses the resize cursor.) I'm not sure there's a good way to fix it, and I have a suspicion that's up to the client rather than labwc. (I can trigger an interactive resize on Nautilus using a keybind, which then untiles the window to allow mouse-based resizing.)

While the GNOME behavior sounds like a nice feature, I think it's incompatible with labwc's concept of snapping: a window is either snapped to a zone and fills it, or else it doesn't. Even if we could fix the Nautilus cursor problem and decide that resizing in this fashion should (somehow) prevent a notification to the client that it was no longer tiled, we'd still have to develop a concept for dynamic snap regions that doesn't now exist.

src/view.c
```diff
@@ -1038,6 +1038,10 @@ view_set_untiled(struct view *view)
 	view->tiled = VIEW_EDGE_INVALID;
 	view->tiled_region = NULL;
 	zfree(view->tiled_region_evacuate);
+
+	if (view->impl->set_tiled) {
+		view->impl->set_tiled(view);
+	}
```
Consolatis, line 1044: I think we usually wrap these in a `view_set_tiled(view)` so we don't have to repeat the impl check everywhere.

Consolatis:
One minor nitpicky comment, other than that this LGTM.

I am still unsure about this not respecting `rc.gap` but let's see how users react to the new default and deal with that when (if) we get issues about it. For some this may be seen as a regression when releasing 0.7.1.

johanmalm:
Happy to go with this one, but just to throw in another angle if we want to ponder further.

We could consider being stricter on two states. I'm just going to call them `soft` and `hard` for the moment.

| soft    | hard   | property    |
|---------|--------|-------------|
| rounded | square | corners     |
| N       | Y      | client notified |
| Y       | N      | natural geometry is changed to reflect new geometry |
| Y       | N      | interactive resize is allowed |

I'm thinking that `GrowToEdge` and `ShrinkToEdge` should result in soft state, whereas SnapToEdge and SnapToRegion should result in a hard state.

We ought to be consistent with CSD + SSD.

I suspect we could have a long discussion around this, so may be best to carry this as a separate issue/PR.

ahesford:
I believe the current changes pretty much cover those cases except for consistency between SSD and CSD. Grow and shrink to edge are really just resize operations, so they already update the natural geometry and allow subsequent resizing. This behaves like a "soft" operation. The snap operations (edge/region) will, after this PR, behave like almost like a "hard" operation. Resizing is still allowed on most clients I've tried, but the resize snaps them out of the tiled state to do so.

ahesford:
Implementing `view_set_tiled` felt strange, because it was used in constructs like
```c
view_set_untiled(view);
...
view_set_tiled(view);
```
I changed the `set_tiled` portion of all function names to `notify_tiled`, which is more descriptive and aligned with the user-facing option anyway.

johanmalm:
LGTM

johanmalm:
Thanks!

DrVektor:
> I've just given it a try and I noticed that when dragging Nautilus away from a tiled state the window borders stay like that until released, which doesn't look optimal imho. Not sure if thats easy to fix/important though.
> 
>  output.mp4

what is the name of bottom taskbar? Is there application menu?

PR #1464, opened 2024-01-22
view: do not resize fullscreen windows

jp7677:
Follow up from https://github.com/labwc/labwc/pull/1457#discussion_r1462493089 and based on discussion with @Consolatis


Consolatis:
Thanks :)

Rationale:
- Action is called `FitToOutput` and a fullscreen fits the output perfectly
- Top layer panels will still be hidden but the fullscreen view will get smaller which is strange

PR #1467, opened 2024-01-23
interactive: unify interactive_{cancel,finish} where possible

ahesford:
This also fixes a bug wherein dragging a window and pressing a hot-key to maximize or fullscreen a window could leave a snap-region highlight visible after the interactive move was canceled.

To reproduce the issue, define an arbitrary snap region and assign a keybind like `W-f`, that will fullscreen a window. The `Super` key also causes regions to be highlighted during interactive moves, so start dragging a window, press and hold `Super` to highlight the snap region, and press `f` to fullscreen the window. Release the mouse button and press the `W-f` keybind to bring the window out of fullscreen. The snap region will still be highlighted until another interactive move triggers it to disappear.

### DIFF
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -196,24 +196,18 @@ snap_to_region(struct view *view)
 void
 interactive_finish(struct view *view)
 {
-	if (view->server->grabbed_view == view) {
-		regions_hide_overlay(&view->server->seat);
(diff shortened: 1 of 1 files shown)


Consolatis:
Nice, LGTM. Not tested.

PR #1470, opened 2024-01-23
Fix pygame-zero crash

### LINKED ISSUE #1466: Intermittent crash on launching pygame zero
For reference - https://pygame-zero.readthedocs.io/en/stable/

Pygame Zero programs seem to intermittently crash at startup when running under labwc on Raspberry Pi - the same programs run fine under X11 or Wayfire.

The problem can be seen with even a blank pygame zero file, so it seems to be something in pygame zero itself which is not compatible with labwc.

Just creating an empty file called intro.py and doing pgzrun intro.py is enough to kill the compositor, but irritatingly, it doesn't happen all the time.

The crash kills the desktop and returns to the greeter, so it looks as if labwc itself is being crashed by something in pgzero.

If there is any logging I can do to try to track this down, please let me know and I will see if I can grab some debug.

johanmalm:
...caused by calling mappable_disconnect() before mappable_connect() and thus hitting an assert()

Handle gracefully intead by simply returning.

Fixes #1466

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2015,7 +2015,13 @@ void
 mappable_disconnect(struct mappable *mappable)
 {
 	assert(mappable);
-	assert(mappable->connected);
+	/*
+	 * pygame-zero gets us here without a mappable_connect() first so we
+	 * have to handle this gracefully. See issue #1466
+	 */
+	if (!mappable->connected) {
+		return;
+	}
 	wl_list_remove(&mappable->map.link);
 	wl_list_remove(&mappable->unmap.link);
 	mappable->connected = false;


Consolatis:
Would be interesting to know how we end up in this situation so we don't accidentally cover up the underlying bug here that will bite us again in the future.
CC @jlindgren90

johanmalm:
I'm happy to conclude it's a client doing odd things.

https://github.com/labwc/labwc/issues/1466#issuecomment-1906931839

src/view.c
```diff
@@ -2015,7 +2015,13 @@ void
 mappable_disconnect(struct mappable *mappable)
 {
 	assert(mappable);
-	assert(mappable->connected);
+	/*
+	 * pygame-zero gets us here without a mappable_connect() first so we
+	 * have to handle this gracefully. See issue #1466
+	 */
+	if (!mappable->connected) {
+		return;
```
jlindgren90, line 2023: I'd suggest we still print a warning here ("received dissociate event but no surface was associated" or similar). If it shows up in the log when debugging some other issue, it could provide clues that a client and/or wlroots is doing something weird.

jlindgren90:
My suggestion (includes the fix for the override-redirect case, which may be mostly theoretical): https://github.com/jlindgren90/labwc/commit/39503f6f2869becd500db845a91799aefb6f9ea5

Consolatis:
> My suggestion (includes the fix for the override-redirect case, which may be mostly theoretical): [jlindgren90@39503f6](https://github.com/jlindgren90/labwc/commit/39503f6f2869becd500db845a91799aefb6f9ea5)

That looks good to me. Want to send a PR?
Maybe you could add something to the comment like `TODO: Potentially remove when starting to track wlroots 0.18.x`?

jlindgren90:
I can send one later - a bit drained at the moment. Feel free to cherry-pick in the meantime if you are so inclined.

Consolatis:
> I can send one later - a bit drained at the moment. Feel free to cherry-pick in the meantime if you are so inclined.

Sure, can do.

Consolatis:
Replaced by
- #1471

PR #1471, opened 2024-01-23
xwayland: prevent crash due to unexpected surface dissociate events

Consolatis:
Fixes
- #1360
- #1466

Follow-up from
- #1470

Not personally tested but code LGTM.
Originally written by @jlindgren90, I only added the `TODO` comment and modified the commit body to close the 2 issues.


johanmalm:
Have tested. Works fine.

PR #1473, opened 2024-01-24
docs/environment: remove MOZ_ENABLE_WAYLAND variable

oifj34f34f:
Starting with version 121, Firefox enables Wayland by default: https://bugzilla.mozilla.org/show_bug.cgi?id=1752398.

### DIFF
--- a/docs/environment
+++ b/docs/environment
@@ -23,12 +23,6 @@
 # XKB_DEFAULT_OPTIONS=grp:shift_caps_toggle
 
 ##
-## Force firefox to use wayland backend.
-##
-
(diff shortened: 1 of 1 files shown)


Consolatis:
I don't think that we should remove the env var though.
For example Debian is still on 115 ESR and in general the env var doesn't hurt but may cause issues when removed.

Narrat:
Thunderbird could also still profit from the env var, although it isn't mentioned in the doc

stefonarch:
Without that env var thunderbird starts in xwayland, yes.

Issue #1476, opened 2024-01-25
Resizing screen doesn't move windows, which can leave them largely non-visible


lurch:
I'm using Labwc (on Raspberry Pi 4 - I'm one of @spl237 's colleagues) and my HDMI monitor has a native resolution of 2560x1440.
I can launch `arandr` and it then appears in the centre of the screen (as all newly-launched applications do). If I change the resolution to 1280x1024 and click "Apply", then the screen-resolution-change happens, but the XY offset of the arandr window doesn't change, which means that most of the window is now hidden off the right of the screen, and I'm unable to see the OK / Cancel buttons on the dialog prompting me to confirm the new screen resolution. As shown in this screenshot:
![change_resolution_labwc](https://github.com/labwc/labwc/assets/476186/6823823e-effe-421e-b519-58622dbf7ace)

If I try the same thing in Wayfire, the arandr window gets moved after (during?)  the screen-resolution-change so that it's still fully-visible:
![change_resolution_wayfire](https://github.com/labwc/labwc/assets/476186/5fcccea7-3a20-497f-8e35-6f78550b1c22)

ahesford:
Can you try this on the master branch? Last-layout restoration may help, but I have a suspicion that it won't. We should probably unconditionally call `update_last_layout_geometry(view)` in `view_adjust_for_layout_change` rather than waiting until a view loses its output.

lurch:
The `master` branch exhibits identical behaviour.

spl237:
For future reference, we are using the master branch anyway - at present, I am updating our package from it every couple of days!

ahesford:
The issue here is that we consider a window that is onscreen by even one pixel as good enough to not trigger repositioning. That makes sense when you're restoring from maximized/fullscreen/tiled views (presumably, only one pixel is visible because you put it there), but doesn't really make sense when the layout has changed. Let's make the "visibility" requirement stricter in those cases to get better behavior.

#1478

Consolatis:
Should be fixed on latest master. Thanks for bringing this up.

spl237:
> Regarding `arandr` - I had forgotten about that tool. How do you make it work on Wayland? Have you done a `s/xrandr/wlr-randr/` patch or is it actually supported upstream?

[...]

PR #1478, opened 2024-01-25
view: ensure midpoint is visible on layout change

ahesford:
When restoring a window's natural geometry during a layout change, require a stricter definition of "visible": either a window's midpoint must be visible, or it will be repositioned. We still respect a mostly-offscreen natural geometry when restoring in an unchanged layout.

Fixes: #1476.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -611,10 +611,12 @@ view_compute_centered_position(struct view *view, const struct wlr_box *ref,
 	return true;
 }
 
-bool
-view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
+static bool
+adjust_floating_geometry(struct view *view, struct wlr_box *geometry,
+		bool midpoint_visibility)
 {
 	assert(view);
+
 	if (!output_is_usable(view->output)) {
 		wlr_log(WLR_ERROR, "view has no output, not positioning");
 		return false;
@@ -627,16 +629,10 @@ view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
 	}
 
 	bool adjusted = false;
-	/*
-	 * First check whether the view is the target screen, meaning that at
-	 * least one client pixel is on the screen.
-	 */
+	bool onscreen = false;
 	if (wlr_output_layout_intersects(view->server->output_layout,
 			view->output->wlr_output, geometry)) {
-		/*
-		 * If onscreen, then make sure the titlebar is also
-		 * visible (and not overlapping any panels/docks)
-		 */
+		/* Always make sure the titlebar starts within the usable area */
 		struct border margin = ssd_get_margin(view->ssd);
 		struct wlr_box usable =
 			output_usable_area_in_layout_coords(view->output);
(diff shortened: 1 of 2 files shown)


lurch:
I've just tested this branch, and it seems to fix #1476 nicely!

ahesford:
I decided even the midpoint is kind of weird. Most users will probably not have windows jumping off the screen, and it is probably undesirable to leave a portion of a window clipped when the layout changes. In the latest attempt, I've made the visibility requirement even stricter, so that a window must be completely visible when the layout changes or else it will be automatically placed.

src/view.c
```diff
@@ -645,15 +641,30 @@ view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
...
+			int right = geometry->x + geometry->width + margin.right;
+			int bottom = geometry->y + geometry->height + margin.bottom;
+
+			onscreen = right <= usable.x + usable.width &&
+				bottom <= usable.y + usable.height;
+		}
+	}
+
+	if (!onscreen) {
+		/* Reposition offscreen views as per policy (except for curo*/
```
Consolatis, line 667: ´(except for curo` ?

src/view.c
```diff
@@ -797,7 +809,7 @@ view_apply_natural_geometry(struct view *view)
 	assert(view_is_floating(view));
 
 	struct wlr_box geometry = view->natural_geometry;
-	view_adjust_floating_geometry(view, &geometry);
+	adjust_floating_geometry(view, &geometry, false);
```
Consolatis, line 812: Inline comments for `false`, same for the next 2 `adjust_floating_geometry()` calls.

src/view.c
```diff
@@ -611,10 +611,12 @@ view_compute_centered_position(struct view *view, const struct wlr_box *ref,
 	return true;
 }
 
-bool
-view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
+static bool
+adjust_floating_geometry(struct view *view, struct wlr_box *geometry,
+		bool fully_visible)
```
Consolatis, line 616: Suggestion: `s/fully_visible/ensure_full_visibility/` ?

ahesford:
Argument changed to `ensure_full_visibility`, inline comments added, stray comment fixed. I refactored the return from `adjust_floating_geometry`.

Consolatis:
Code itself LGTM (other than the minor stray `}` causing the build error)

Just to ensure that this is intended behavior:
Tested this a bit nested (via `wlr-randr --output WL1 --custom-mode 1500x600`).
Spawned foot, changed width of the output to 1500, moved the foot window to the right so there is only a small gap to the right edge, changed output width to 1300 via wlr-randr, new logic kicked in correctly and centered foot, changed output width back to 1500 and the view then moved to its old position (assumingly because we didn't update the last layout position).

ahesford:
The observed behavior is expected; a layout change never replaces the last-layout geometry, and every subsequent layout change will try to restore the view to its last-layout geometry as closely as possible until the user explicitly moves the view.

ahesford:
After some IRC discussion, I went back to using the midpoint, rather than the entire view, as the trigger for repositioning. This mitigates some surprising repositioning that might occur when adding or removing panels, but doesn't add a lot of extra complexity to try to figure out if the user actually *wants* views to hang off the edge of an output.

src/view.c
```diff
@@ -645,27 +641,43 @@ view_adjust_floating_geometry(struct view *view, struct wlr_box *geometry)
...
-		if (rc.placement_policy == LAB_PLACE_AUTOMATIC) {
-			if (placement_find_best(view, geometry)) {
-				return true;
-			}
+
+		if (!midpoint_visibility) {
+			/*
+			 * If fully visibility is not required, the view is on
+			 * screen if at least one pixel is visible.
+			 */
```
Consolatis, line 654: Outdated comment

Consolatis:
Thanks :)

Issue #1454, opened 2024-01-20
Questions (programming)


droc12345:
I am trying to figure out where a few things are, in an earlier version (pre scene) I had patches for changing alpha on windows, but I'm not sure where the data for that is now. Any hints would be appreciated.
I'm looking to do something like have non-focused windows be partly transparent.

I would like to add/modify the change workspace to skip empty (no windows) workspaces
OR
add/modify the next window to look at all windows no matter the workspace.

I'm not asking for anyone else to do it, I'm just needing to know where to look.

Thanks

johanmalm:
Regarding desktop_cycle_view() and osd.c I can do some refactoring first (on my TODO list).

We need to common-ize those two before changing view-criteria.

Think I already started that on a branch somewhere. Will have a look Mon/Tue

droc12345:
That would be nice, I assume keep the old behavior and have a new option to see all workspace/windows.

Looking at sway has given me some ideas as to how to implement opacity.
But I have to think about how to set options to use it properly.
Points to consider....
Can all windows but focused be opaqued to some value?
Can opacity be set (config option) for some windows and can it be easily changed or toggled?
I'm looking at having config option(s) for setting.

Thanks for the replies

ahesford:
The function `view_set_activated` is called when windows change focus, so you can set opacity there.

You can define arbitrary opacity values to, *e.g.*, set, increment, decrement or toggle opacity, and then you can bind these as desired, including to window actions that can force them on certain windows when created.

johanmalm:
`add/modify the next window to look at all windows no matter the workspace` should be easy following #1479

droc12345:
applying patch gives me a segment violation.

as soon as I trigger the action, it immediately seg faults and goes back to cmdline.
It doesn't even show a highlight of the window, so something in the way the list is gathered is my best guess.
But I haven't checked the code that much, just a gut feeling.

I'll look at it later, it's given me a direction to go, so I'm happy.

PR #1479, opened 2024-01-25
Refactor cycle view

johanmalm:
Related-to #1454

Cc: @droc12345

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -118,17 +118,7 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 	bool forwards = dir == LAB_CYCLE_DIR_FORWARD;
 	iter = forwards ? view_next_no_head_stop : view_prev_no_head_stop;
 
-	/*
-	 * TODO: These criteria are the same as in display_osd() in osd.c
-	 * for the time being.
-	 *
-	 * A future improvement could be to make this configurable for example
-	 * in rc.xml and then use rc.cycle_view_criteria (or whatever) both
-	 * here and in the osd.c window-switcher code
-	 */
-	enum lab_view_criteria criteria = LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
-		| LAB_VIEW_CRITERIA_ROOT_TOPLEVEL
-		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
+	enum lab_view_criteria criteria = rc.window_switcher.criteria;
 
 	/*
 	 * Views are listed in stacking order, topmost first.  Usually the
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -127,6 +127,7 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 	 * here and in the osd.c window-switcher code
 	 */
 	enum lab_view_criteria criteria = LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
+		| LAB_VIEW_CRITERIA_ROOT_TOPLEVEL
 		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
 
 	/*
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -106,109 +106,45 @@ desktop_focus_view_or_surface(struct seat *seat, struct view *view,
 	}
 }
 
-static struct wl_list *
-get_prev_item(struct wl_list *item)
-{
-	return item->prev;
-}
-
-static struct wl_list *
-get_next_item(struct wl_list *item)
-{
-	return item->next;
-}
-
-static struct view *
-first_view(struct server *server)
-{
-	struct wlr_scene_node *node;
-	struct wl_list *list_head =
-		&server->workspace_current->tree->children;
-	wl_list_for_each_reverse(node, list_head, link) {
-		if (!node->data) {
-			/* We found some non-view, most likely the region overlay */
-			continue;
-		}
-		struct view *view = node_view_from_node(node);
-		if (view_is_focusable(view)) {
-			return view;
-		}
-	}
-	return NULL;
-}
-
 struct view *
 desktop_cycle_view(struct server *server, struct view *start_view,
 		enum lab_cycle_dir dir)
 {
+	/* Make sure to have all nodes in their actual ordering */
+	osd_preview_restore(server);
+
+	struct view *(*iter)(struct wl_list *head, struct view *view,
+		enum lab_view_criteria criteria);
+	bool forwards = dir == LAB_CYCLE_DIR_FORWARD;
+	iter = forwards ? view_next_no_head_stop : view_prev_no_head_stop;
+
 	/*
-	 * Views are listed in stacking order, topmost first.  Usually
-	 * the topmost view is already focused, so we pre-select the
-	 * view second from the top:
+	 * TODO: These criteria are the same as in display_osd() in osd.c
+	 * for the time being.
+	 *
+	 * A future improvement could be to make this configurable for example
+	 * in rc.xml and then use rc.cycle_view_criteria (or whatever) both
+	 * here and in the osd.c window-switcher code
+	 */
+	enum lab_view_criteria criteria = LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
+		| LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP
+		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
+
+	/*
+	 * Views are listed in stacking order, topmost first.  Usually the
+	 * topmost view is already focused, so when iterating in the forward
+	 * direction we pre-select the view second from the top:
 	 *
 	 *   View #1 (on top, currently focused)
 	 *   View #2 (pre-selected)
 	 *   View #3
 	 *   ...
-	 *
-	 * This assumption doesn't always hold with XWayland views,
-	 * where a main application window may be focused but an
-	 * focusable sub-view (e.g. an about dialog) may still be on
-	 * top of it.  In that case, we pre-select the sub-view:
-	 *
-	 *   Sub-view of #1 (on top, pre-selected)
-	 *   Main view #1 (currently focused)
-	 *   Main view #2
-	 *   ...
-	 *
-	 * The general rule is:
-	 *
-	 *   - Pre-select the top view if NOT already focused
-	 *   - Otherwise select the view second from the top
 	 */
-
-	/* Make sure to have all nodes in their actual ordering */
-	osd_preview_restore(server);
-
-	if (!start_view) {
-		start_view = first_view(server);
-		if (!start_view || start_view != server->active_view) {
-			return start_view;  /* may be NULL */
-		}
+	if (!start_view && forwards) {
+		start_view = iter(&server->views, NULL, criteria);
 	}
-	struct view *view = start_view;
-	struct wlr_scene_node *node = &view->scene_tree->node;
 
-	assert(node->parent);
-	struct wl_list *list_head = &node->parent->children;
-	struct wl_list *list_item = &node->link;
-	struct wl_list *(*iter)(struct wl_list *list);
-
-	/* Scene nodes are ordered like last node == displayed topmost */
-	iter = dir == LAB_CYCLE_DIR_FORWARD ? get_prev_item : get_next_item;
-
-	do {
-		list_item = iter(list_item);
-		if (list_item == list_head) {
-			/* Start / End of list reached. Roll over */
-			list_item = iter(list_item);
-		}
-		node = wl_container_of(list_item, node, link);
-		if (!node->data) {
-			/* We found some non-view, most likely the region overlay */
-			view = NULL;
-			continue;
-		}
-		view = node_view_from_node(node);
-
-		enum property skip = window_rules_get_property(view, "skipWindowSwitcher");
-		if (view_is_focusable(view) && skip != LAB_PROP_TRUE) {
-			return view;
-		}
-	} while (view != start_view);
-
-	/* No focusable views found, including the one we started with */
-	return NULL;
+	return iter(&server->views, start_view, criteria);
 }
 
 struct view *
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -127,7 +127,6 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 	 * here and in the osd.c window-switcher code
 	 */
 	enum lab_view_criteria criteria = LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
-		| LAB_VIEW_CRITERIA_NO_ALWAYS_ON_TOP
 		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
 
 	/*
--- a/src/osd.c
+++ b/src/osd.c
@@ -163,6 +163,9 @@ osd_preview_restore(struct server *server)
 {
 	struct osd_state *osd_state = &server->osd_state;
 	if (osd_state->preview_node) {
+		wlr_scene_node_reparent(osd_state->preview_node,
+			osd_state->preview_parent);
+
 		if (osd_state->preview_anchor) {
 			wlr_scene_node_place_above(osd_state->preview_node,
 				osd_state->preview_anchor);
@@ -176,6 +179,7 @@ osd_preview_restore(struct server *server)
 			wlr_scene_node_set_enabled(osd_state->preview_node, false);
 		}
 		osd_state->preview_node = NULL;
+		osd_state->preview_parent = NULL;
 		osd_state->preview_anchor = NULL;
 	}
 }
@@ -190,8 +194,11 @@ preview_cycled_view(struct view *view)
 	/* Move previous selected node back to its original place */
 	osd_preview_restore(view->server);
 
-	/* Remember the sibling right before the selected node */
+	/* Store some pointers so we can reset the preview later on */
 	osd_state->preview_node = &view->scene_tree->node;
+	osd_state->preview_parent = view->scene_tree->node.parent;
+
+	/* Remember the sibling right before the selected node */
 	osd_state->preview_anchor = lab_wlr_scene_get_prev_node(
 		osd_state->preview_node);
 	while (osd_state->preview_anchor && !osd_state->preview_anchor->data) {
@@ -205,6 +212,14 @@ preview_cycled_view(struct view *view)
 	if (!osd_state->preview_was_enabled) {
 		wlr_scene_node_set_enabled(osd_state->preview_node, true);
 	}
+
+	/*
+	 * FIXME: This abuses an implementation detail of the always-on-top tree.
+	 *        Create a permanent server->osd_preview_tree instead that can
+	 *        also be used as parent for the preview outlines.
+	 */
+	wlr_scene_node_reparent(osd_state->preview_node,
+		view->server->view_tree_always_on_top);
 
 	/* Finally raise selected node to the top */
 	wlr_scene_node_raise_to_top(osd_state->preview_node);
--- a/src/osd.c
+++ b/src/osd.c
@@ -397,20 +397,16 @@ render_osd(struct server *server, cairo_t *cairo, int w, int h,
 }
 
 static void
-display_osd(struct output *output)
+display_osd(struct output *output, struct wl_array *views)
 {
 	struct server *server = output->server;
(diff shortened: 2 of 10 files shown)


johanmalm:
Marked as draft in response to https://github.com/labwc/labwc/issues/1454#issuecomment-1911046196

johanmalm:
@droc12345

Any chance of an ASAN log?

```
meson setup build/
meson configure -Db_sanitize=address,undefined build/
meson compile -C build/
./build/labwc 2>log.txt
```

Then force crash and paste `log.txt` here.

droc12345:
Interesting, if I sanitize then no crash, no error.

Single window, no popup osd
Multiple windows (same workspace) popup osd
Multiple windows (multi workspces) popup osd, but only that workspaces windows appear.

I should mention I'm using gcc 13 and only 2 errors/warnings when compiling
one in wlroots, that I had to overrider with -Wno-error=stringop-overflow to get it to compile
one in src/common/string-helpers.c:66:17 null format string (strictly warning)

droc12345:
A little bit of progress, I can now switch amongst multiple windows on different workspaces.

I had to remove LAB_VIEW_CRITERIA_CURRENT_WORKSPACE from enum criteria.
Still running with sanitizer and no crashes.
Still doesn't show osd with just a single window, 
but I don't think it should have ever shown for just one window.

droc12345:
I've gotten it to mostly work, with adding/changing a few more spots.
no windows - no osd
one window - osd
>1 window - osd
and it switches amongst all windows, including proper window preview dimensions (on other workspace)

It doesn't pop up the osd if I have one or more windows but none on the current workspace.

I'm still having to run with sanitize or it will return to cmdline,
with just a single window using next window key, Multiple windows no problem. 

There's not a crash log when this happens (as far as the sanitizer) so I think there's some type logic bug
in properly stopping if just one window. I think it cycles around (next_view) and hits the assert or 
later in the for loop it's trying to check a view that isn't accurate. Just not sure.

droc12345:
I think I have it working, well except for osd display on empty workspace. 
But NO sanitize needed to keep from crashing.

0 windows - no osd
1 window (current workspace) - osd
1 window (empty workspace) - no osd, but it warps to the workspace the window is in 
multi windows  - osd 

I had to replace the first_view stuff with something that works, so that function isn't used
I look for server->views, 
check if returned view is focusable
and if so, server workspace_current isn't equal to the current view workspace (empty workspace)
then I workspace_switch_to with focus

and then I have to check to see how many views the server knows about, 
if less than 2 then I just return the view 
otherwise I fall through to for_each_view_iter for any others.

It seems to work ok, if no problems show up then I'll either whip up a patch for you or 
set it up to send a merge request.

johanmalm:
Thanks for thinking about this.
Just having a quick look, `view_next` and `view_prev` actually stop at head which doesn't seem right for `desktop_cycle_view()`. We'll probably have to come up with a different macro and/or iterators.

droc12345:
Yeah, view next/prev has the assert head at the beginning, but that isn't really a problem
for_each_view_iter (view_next) isn't called unless I have more than 1 view,
so that assert should never trigger.

droc12345:
My first time pushing a commit, hope it's correct.
#1484

src/desktop.c
```diff
@@ -170,42 +139,30 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
+	 * for the time being.
+	 *
+	 * A future improvement could be to make this configurable for example
+	 * in rc.xml and then use rc.cycle_view_criteria (or whatever) both
+	 * here and in the osd.c window-switcher code
+	 */
+	enum lab_view_criteria criteria = LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
+		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
+
+	/* On the first alt-tab, there is no start_view to we have to set it */
```
Consolatis, line 158: `so we have to set it`

src/desktop.c
```diff
@@ -170,42 +139,30 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
+	struct view *view = start_view ? : iter(&server->views, NULL, criteria);
+
+	for_each_view_from_bidirectional(view, &server->views, iter, criteria) {
+		if (view_is_focusable(view)) {
```
Consolatis, line 162: I think we do check `view_is_focusable()` in `matches_criteria()`.

src/desktop.c
```diff
@@ -170,42 +139,30 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 	/* Make sure to have all nodes in their actual ordering */
 	osd_preview_restore(server);
 
-	if (!start_view) {
-		start_view = first_view(server);
-		if (!start_view || start_view != server->active_view) {
-			return start_view;  /* may be NULL */
-		}
-	}
```
Consolatis, line 178: With the removal of this I think we break the the 2nd point about the XWayland views (subview with focus but not on top of the list) in the comment above.

johanmalm: Good point.

Not sure how to best deal with that. 

We could consider a `criteria` for toplevel-with-no-parent.

droc12345:
I didn't check any Xwayland stuff as I run strictly wayland, X not even installed.

johanmalm:
> I didn't check any Xwayland stuff as I run strictly wayland, X not even installed.

Same applies to mousepad with about dialog open on wayland

droc12345:
Multi-workspace works switching windows 
...if... LAB_VIEW_CRITERIA_CURRENT_WORKSPACE is removed from 
desktop_cycle_view and display_osd

A config option would allow either behavior

src/desktop.c
johanmalm: Overtaken by events.

src/desktop.c
johanmalm: Thanks for spotting that. It actually simplifies things ever more because all we need to do it call `iter()`. No more `for_each_view_from_bidirectional` :smile:

src/desktop.c
johanmalm: I've added a root-toplevel criteria, so the problem has gone away.

johanmalm:
Well... ready for review I think.
There isn't much left of `desktop_cycle_view()` which is normally a good sign.
I suggest we leave the addition of `rc.cycle_view_criteria` (or whatever) for another PR.

src/desktop.c
Consolatis: > I've added a root-toplevel criteria, so the problem has gone away.

Didn't test yet but doesn't this break the ability of selecting sub windows (like `Open file` / `Search` / `About` dialogues)?

johanmalm: Yes, intentionally so.

`desktop_focus_view()` calls `view_move_to_front()` which moves all sub-views (e.g. modal dialogs) with it, so selecting the root-toplevel moves the `Open File` / `Search` / `About` with it. 

Having been wanting to make this change for a long time, but it wasn't a sensible change prior to this `osd.c` / `desktop_cycle_view()` refactoring.

I haven't looked at what other OS/DE/WM/Compositors do, but suspect they just show root toplevels in the window-switchers. I might test a couple of other tomorrow.

Consolatis: Well, not all subwindows are modal dialogues. Usually `Search` in a GUI editor is one of those.
We would then not have the ability to switch between the `Search` window and the main window via `A-Tab` anymore.

droc12345: In firefox, manage bookmarks brings up a window, when a selection is made 
the main window gets focus, and the bookmark window is now hidden.
To get to it (without alt-tab) would require me to either move the main window enough to see the 
one behind or possibly minimize the main window. Neither option is great.

Several apps have the behavior of subwindows being there until manually closed.

droc12345:
A comment on switching to view_next_no_head_stop, 
it got rid of those crashes that were there with view_next.

src/desktop.c
johanmalm: Well, we could just apply the root-toplevel when pre-selecting the second window, but please consider the below first:

1. If a "subwindow" (dialog or whatever) is a separate root-toplevel, then it still shows up in the window-switcher (this happens with gimp, inkscape's About window, etc.)
2. If a subwindow is a indeed a child-view then it moves to back/front with its root-toplevel so I struggle to see the benefit of listing it separately in the window-switcher.

Note: "subwindow" used loosely here as you might argue the first case doesn't apply

droc12345: To me there are 2 types of "sub" windows, 
dialogs like open, library, etc
drop down like menu selection (ff, etc) but they usually go away with change of focus.

Waybar's taskbar shows 2 windows for firefox when I have library open for example.
but if I click on a menu then I don't see a second window in taskbar.

I assume there's some way to tell the difference between the 2 types, and maybe what you're
doing might work.

Consolatis: > 2\. If a subwindow is a indeed a child-view then it moves to back/front with its root-toplevel so I struggle to see the benefit of listing it separately in the window-switcher.

For me it's about fast keyboard focus switching without having to use the mouse.

droc12345: Did a quick test, using root toplevel, seems to work ok.

I started ff, and selected manage bookmarks, was able to alt-tab between both windows
and they show up in the osd, firefox has whatever title from main window, bookmarks shows
library as title. 

Not sure how Xwayland windows will behave though, as I don't have that installed.

droc12345:
It seems to work ok, haven't had any crashes, at least due to code and not my fumble fingers.

To make windows show for all workspaces, 
remove criteria LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
from desktop_cycle_view and display_osd

droc12345:
One other little change, allows for doing alt-tab on an empty workspace.
Useful when you've quit an application and want to swap to another workspace/app.

at top of osd_update
```
-   struct wl_list *node_list =
-       &server->workspace_current->tree->children;
-
-   if (wl_list_empty(node_list) || !server->osd_state.cycle_view) {
+   if (!server->osd_state.cycle_view) {
```

It's understandable for it to be there when switch window only handled the current workspace,
but with seeing all workspaces, it's not needed.

I've been running with this change and no LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
and I've not had any problems. 

Anyway thanks for putting out the original patch, and changes.

droc12345:
A little peculiarity in the alt-tab preview and outline, not a show stopper, just something to keep in mind.

When all windows are on the same workspace, then cycling with either preview or outline will actually swap among the windows (as far as viewing) but if the window is on another workspace, preview doesn't work, but you still get the outline. Something about the way wlr_set_node* behaves.

It's just odd to see some windows swapping but others only get an outline.
But it is an indication that the outline only is on another workspace. 
I'm looking at what is needed to fix the preview, but it would involve something like go to workspace 
along with the preview of the window. Don't know if it's worth it.

I also added the workspace name to each node line (left of type indicator) so that one could turn off 
both preview and outline and still tell which window is on what workspace.

johanmalm:
In order to get this merged, could we leave the workspace considerations until later (unless there is something we ought change in this PR to lay the foundations).

For me the way forward looks like:

1. Review + merge this PR
2. Setup infrastructure for cycle view criteria config option
3. Add config option for showing workspaces

droc12345:
You can merge the refactor-cycle-view patch as is.
It works just the same as it's been, you basically cleaned up/replaced code with better implementation,
at least with my testing.

Consolatis:
Another thing I noticed with this PR is that there are still some occurrences of `server->workspace_current->tree->children` in `src/osd.c`. Those should likely be removed / replaced in lockstep with the other changes.

droc12345:
server->workspace_current->tree->children is in display_osd()
and it's only purpose is to pass a node list to render_osd which never uses it.

droc12345:
Just for giggles, this is how I have my version of osd looking

https://ibb.co/mNQS6p2

droc12345:
Still working fine, absolutely no problems from the changes. I'm just waiting for this to be merged so that I can do a pull request.

I've modified the changes I have for all workspace/window switching,
default behavior is to be exactly like it is now, window switching/current ws only,
but with a flag added to next/prev window bind then all workspaces are looked at.
```
<action name="NextWindow" ws_all="yes" />
```
default is ws_all=no and can be left off for default.

Consolatis:
We will delay this PR until 0.7.1 is released (we target for March 1st).

droc12345:
Waiting is fine, I'll just continue to use my patched version. 

All I can do is test, and tell you it works for me, both johan's patches and my patches on top of his.

The all windows/workspaces would obviate the need to worry about how to send workspace info for a window to panels like waybar though, since foreign toplevel doesn't address it.

droc12345:
This PR should be able to go in, as is. Shouldn't have any negative effects.

I've been running this, with my patches on top of it, along with #1533 changes
and have seen no problems. Even did a quick test by installing xwayland and 
a couple of small X apps and saw no problems.

johanmalm:
> Another thing I noticed with this PR is that there are still some occurrences of `server->workspace_current->tree->children` in `src/osd.c`. Those should likely be removed / replaced in lockstep with the other changes.

Done.

johanmalm:
I think I've responded to all review comments thus far, except possibly the inclusion of `<windowSwitcher allToplevels>` (or whatever) to show dialogs - i.e. to `=& ~LAB_VIEW_CRITERIA_ROOT_TOPLEVEL`) if we still want that. I'm happy to go without unless there is a specific request.

I think the six commits make sense they way they are - and it's probably easiest to read the changes by following the commit journey.

Reviews welcome.

droc12345:
This PR commits (6) make a complete and whole change. 
There should be no difference than the way things work today re. window switching.
I've been running with these changes for quite a while now.

Once you've done this, I'll refactor what I have and produce a clean pull for my proposed changes.

src/osd.c
```diff
@@ -458,13 +455,24 @@ display_osd(struct output *output)
...
+}
+
 void
 osd_update(struct server *server)
 {
-	struct wl_list *node_list =
-		&server->workspace_current->tree->children;
-
-	if (wl_list_empty(node_list) || !server->osd_state.cycle_view) {
+	if (!nr_entries(server) || !server->osd_state.cycle_view) {
```
Consolatis, line 469: We are now creating the same array at least two and at most 1 + output count times so I'd suggest creating the array once in `osd_update` and then pass it around to `display_osd()` instead.

droc12345:
Changes don't look too bad (eyeballing it), I'll try the patch tomorrow and see what it looks like in real life.

When looking at all workspaces, the preview option at least should be disabled, and the outline option
doesn't really make a lot of sense unless it's all in the same workspace.

What I had done was just disable preview/outlines when multi-workspace osd was being used.

droc12345:
all workspaces don't work, I'll put in some debug stuff later to see what's going on.
Possibly criteria related.

droc12345:
Still trying to chase down what going on. it has to do with the last 2 additions to the PR.
Something doesn't work right. it's where all the workspace stuff is added.

I recommend dropping the last 2 pieces from the PR, and looking at all workspace stuff separately.
Something in the logic doesn't seem to be right.

Here is a working version, that has all workspaces, but is using a flag on next window to decide 
whether current workspace or all workspaces. 
https://github.com/droc12345/labwc/commit/d233153de394b34af25c75c7aefeb5d0b317bdcb

droc12345:
I made a comment, but it was in the review area, has to do with lines 846 and 847.

I also think it needs these changes to able turn off outlines and previews when running all workspace.

```
--- a/include/config/rcxml.h    2024-03-04 08:53:18.196440378 -0600
+++ b/include/config/rcxml.h    2024-03-04 08:54:41.410149942 -0600
@@ -140,6 +140,7 @@
        bool show;
        bool preview;
        bool outlines;
+       bool allworkspaces;
        uint32_t criteria;
        struct wl_list fields;  /* struct window_switcher_field.link */
    } window_switcher;
diff -u -r -x '.git*' a/src/config/rcxml.c b/src/config/rcxml.c
--- a/src/config/rcxml.c    2024-03-04 08:53:18.196440378 -0600
+++ b/src/config/rcxml.c    2024-03-04 08:56:05.768855531 -0600
@@ -896,6 +896,7 @@
    } else if (!strcasecmp(nodename, "allworkspaces.windowSwitcher")) {
        if (!strcasecmp(content, "yes")) {
+           set_bool(content, &rc.window_switcher.allworkspaces);
            rc.window_switcher.criteria &=
                ~LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
        }
@@ -1154,6 +1154,7 @@
    rc.window_switcher.show = true;
    rc.window_switcher.preview = true;
    rc.window_switcher.outlines = true;
+   rc.window_switcher.allworkspaces = false;
```

johanmalm:
@droc12345 Thanks for spotting that. I had a different syntax in mind when I first started (which is what I had in my local rc.xml which is why I didn't spot the problem when testing myself!). Have force-pushed the fix.

johanmalm:
So, ready for review again.

droc12345:
Looks good to me, compiles clean, works swapping between yes and now with reconfigure.

droc12345:
Preview only works on current workspace (without a lot of work) and I think it will cause problems and questions, like why sometimes do I see a preview and other times not. 

The outline, I have mixed feelings about, yes it gives some type feedback, but without seeing the contents of a window, it could be confusing. I have 3 (same size) terminal windows, the outline doesn't help me much.

Which is why in my personal changes, in the osd, I put the workspace it's on, whether in a special state,
max/min/full and if multiple outputs (monitors) it will tell which one the app is on. 
That's enough info to figure out which selection to make.

What would be useful is thumbnails of the other workspaces, but not sure how involved that would be.

Consolatis:
> Preview only works on current workspace (without a lot of work) and I think it will cause problems and questions, like why sometimes do I see a preview and other times not.

Shouldn't be that much work actually. Something like this should work already (not tested at all though):
```diff
diff --git a/include/labwc.h b/include/labwc.h
index 78b8b3b6..a686d10a 100644
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -328,6 +328,7 @@ struct server {
 		struct view *cycle_view;
 		bool preview_was_enabled;
 		struct wlr_scene_node *preview_node;
+		struct wlr_scene_tree *preview_parent;
 		struct wlr_scene_node *preview_anchor;
 		struct multi_rect *preview_outline;
 	} osd_state;
diff --git a/src/osd.c b/src/osd.c
index d283459f..b0b59a78 100644
--- a/src/osd.c
+++ b/src/osd.c
@@ -163,6 +163,9 @@ osd_preview_restore(struct server *server)
 {
 	struct osd_state *osd_state = &server->osd_state;
 	if (osd_state->preview_node) {
+		wlr_scene_node_reparent(osd_state->preview_node,
+			osd_state->preview_parent);
+
 		if (osd_state->preview_anchor) {
 			wlr_scene_node_place_above(osd_state->preview_node,
 				osd_state->preview_anchor);
@@ -176,6 +179,7 @@ osd_preview_restore(struct server *server)
 			wlr_scene_node_set_enabled(osd_state->preview_node, false);
 		}
 		osd_state->preview_node = NULL;
+		osd_state->preview_parent = NULL;
 		osd_state->preview_anchor = NULL;
 	}
 }
@@ -190,8 +194,11 @@ preview_cycled_view(struct view *view)
 	/* Move previous selected node back to its original place */
 	osd_preview_restore(view->server);
 
-	/* Remember the sibling right before the selected node */
+	/* Store some pointers so we can reset the preview later on */
 	osd_state->preview_node = &view->scene_tree->node;
+	osd_state->preview_parent = view->scene_tree->node.parent;
+
+	/* Remember the sibling right before the selected node */
 	osd_state->preview_anchor = lab_wlr_scene_get_prev_node(
 		osd_state->preview_node);
 	while (osd_state->preview_anchor && !osd_state->preview_anchor->data) {
@@ -206,6 +213,14 @@ preview_cycled_view(struct view *view)
 		wlr_scene_node_set_enabled(osd_state->preview_node, true);
 	}
 
+	/*
+	 * FIXME: This abuses an implementation detail of the always-on-top tree.
+	 *        Create a permanent server->osd_preview_tree instead that can
+	 *        also be used as parent for the preview outlines.
+	 */
+	wlr_scene_node_reparent(osd_state->preview_node,
+		view->server->view_tree_always_on_top);
+
 	/* Finally raise selected node to the top */
 	wlr_scene_node_raise_to_top(osd_state->preview_node);
 }
```

johanmalm:
> > Preview only works on current workspace (without a lot of work) and I think it will cause problems and questions, like why sometimes do I see a preview and other times not.
> 
> Shouldn't be that much work actually. Something like this should do the job (not tested at all though):
> 

That patch did indeed work very well. Thanks! Have added as a commit.

Ready again for final review.

Consolatis:
Nice work!

PR #1482, opened 2024-01-26
Comprehensive snapping/resistance rewrite

ahesford:
This is a comprehensive rewrite of the resistance and snapping algorithm based on a common edge-finding engine to provide consistent behavior for mouse-based resistance and `{Grow,Shrink,Move}ToEdge` actions. The edge engine is generalized from my earlier work in 57ea197e and the first two commits of this PR are taken from #1468 to make sure that all resistance and snapping actions consider `rc.gap` properly and uniformly.

We can consider this for inclusion into 0.7.1 if the rest of the team agrees. However, if this PR is too extensive to be considered for the upcoming 0.7.1, we should still merge #1468 for that release, because it includes some intermediate quality-of-life improvements for users who configure gaps.

The edge engine provides a means to walk the output and view lists, building a `struct border` for the view of interest and its target position, both of which include `rc.gap` padding if so configured. The engine considers that the view will move to its target region through one or more convex regions (*i.e.*, other views) that are represented as a `struct border` defining the limits of the regions *without* any `rc.gap` padding (because we don't want to double-count the gaps). For edge resistance/attraction of outputs, the problem is a bit messier, because outputs are concave regions. Rather than complicate the logic and require an entirely separate set of conditions for testing output edges and view edges, it seemed much cleaner (and simpler) to handle output edges by breaking each output into four half-infinite regions (one per output edge) and apply the same convex logic to each of those.

The edge engine expects to be provided an `edge_validator_t` function pointer that will be provided the current (or pending) edges of the view of interest, the target edges for that view, the edges of a region that may be encountered as the view transitions from its current (pending) boundaries to its target boundaries, and a pointer to a `struct border` that maintains the current list of "nearest" edges for the motion. The edge engine considers that each one of the four edges of a window may move independently and arbitrarily. Thus, the implementation should be able to handle a single complex transition that represents a simultanous move and resize.

In all contexts, the edge algorithms (ab)use `INT_MAX` or `INT_MIN` to indicate an invalid edge. A couple of helpers have been added to detect (`BOUNDED_INT`) or preserve these "invalid" markers (`clipped_add`, `clipped_sub` to allow arithmetic without overflow or underflow) during geometric manipulations.

The resistance mechanism uses two wrappers of the same underlying validator---one to apply screen-edge resistance or attraction, and another to apply window-edge resistance or attraction. The common validator walks all for edges of the view to check for movement, and updates the "nearest" edges in each corresponding direction. For motion of a given edge, the validator will check the opposing edge and the aligned edge of any encountered region (*e.g.*, if a left edge is moving, the validator will check to see if it should be snapped to the right edge, with gaps, of the region; or to the left edge, with no gaps, of the region). A "valid" edge in this search is any edge that has an attract/resist zone which encloses the target position of the moving edge. The "nearest" of the valid edges, if any exists, is then selected for snapping.

The snapping mechanisms for `{Grow,Shrink,Move}ToEdge` are almost identical to the resistance mechanism. These actions simply set target regions that, respectively, would expand a view to the relevant output edge, shrink a view along the relevant edge to half its original size, or move the view to the relevant output edge. Then, the edge engine is used to identify the nearest valid edge between the original and target positions of the view. In these cases, the validatoris somewhat simpler than that for resistance. Because these actions do not need to consider attract/resist zones, any edge between the original and target positions of the corresponding view edge are valid.

The original implementation of the snap actions had some strange behavior when acting on views that were larger than their outputs. For example, `MoveToEdge` rightward for a view that extends off the right edge of the screen would actually move the view *leftward*. To avoid additional complexity and counter-intuitive behavior, I impose some sanity checks that abandon snapping attempts in these cases. I don't believe that these cases are particularly interesting but, if there is a strong desire to handle them, future refinements are possible.

After spending several days (and following several iterations on the method), I believe these changes perform as expected. Furthermore, if we decide in the future to employ the `struct overlap_bitmap` used for automatic window placement instead of walking the window list, the changes should mostly be limited to the internals of `edges_find_outputs` and `edges_find_neighbors`, although it may be necessary to adjust the interface slightly. I have thought briefly about doing so, but there are a some challenges surrounding `rc.gap`: the overlap bitmap identifies *edges*, but does not make it easy to identify whether any given edge should be considered "aligned" (where the moving edge should be placed *without* a gap) or "opposing" (where the moving edge should be placed *with* a gap). I prefer to think about that challenge later (if ever).


johanmalm:
Wow... think this will have to be a weekend review, or early next week (for me anyway)

johanmalm:
I think this is technically fine.

Have tested and note a slightly difference with `MoveToEdge` in that we now swap to every window edge, whereas before the algorithm took into account direction. I think I actually prefer the new behaviour as it gives more control.

I'll have another read through another evening to see if I can think of anything to help make it simpler to understand.

ahesford:
Strange, I remember encountering the double-edge behavior during testing of the original implementation, and replicating that in this rewrite was a nontrivial effort. The double-edge snapping is definitely new for window resistance.

We can eliminate the double-edge behavior by dropping the "align" edge but, like you, I've come to think that this behavior is desirable. Users who want to snap to windows may want to butt up against other windows, but they may also want to align edges against a common line for consistent grid alignment. The more I've used the new snapping and resistance, the more I've come to like it.

johanmalm:
Okay. Have reviewed + tested. 
LGTM.

Two minor points, but consider them optional:

- `s/mgeom/usable/`. I think that variable name was a leftover from the previous resistance implementation. In other other areas of labwc we use `usable` for this.
- When doing `MoveToEdge` between outputs, usable area is not taken into account when receiving output is smaller (we currently do this). See video below.



https://github.com/labwc/labwc/assets/6955353/7b61ee5b-7c95-49f8-8f15-70075d211d2d

johanmalm:
Really nice work btw!! :cake:

ahesford:
I made the `mgeom -> usable` change. The observed panel-covering behavior is actually expected and consistent with the prior implementation, although it is probably surprising and undesirable. Because this would represent a change in behavior and invites some discussion about specific details, I propose an improvement in #1488.

PR #1484, opened 2024-01-28
changes for multi-workspace window switching

droc12345:
These are the changes for multi-window/multi-workspace osd

This my first time doing this so I hope I did it right.

Edit to add; I haven't checked using it in reverse, but it (theoretically) should work with no problems.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -300,6 +300,22 @@ bool view_matches_query(struct view *view, struct view_query *query);
 	     view = view_next(head, view, criteria))
 
 /**
+ * for_each_view_iter() - iterate in given direction against criteria
+ * @view: Iterator.
+ * @head: Head of list to iterate over.
+ * @iter: Iterator function to get next or previous view
+ * @criteria: Criteria to match against.
+ * Example:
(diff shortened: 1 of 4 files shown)


domo141:
I'd nitpick the use of 01/28/2024 in commit message (it looks like a date in u.s. format)

In this particular case there is not much change for confusion as there is not 28 months
in a year. but something like 2/1/2025 would be confusing (saw such a date few days ago,
both 2025-01-02 and 2024-02-01 are thursdays which added to the confusion ;)

maybe showing date there is not needed
if so I'd vote using the ISO format 2024-01-28

ahesford:
The commit message shouldn't contain a date at all. Git metadata already records the date of a commit. The message should be a descriptive summary of the change, preferably with some location information, like

> desktop: support switching to windows in other workspaces

droc12345:
I've messed it up, trying to change things. 
I'll delete this one and redo it properly.

Thanks for all the comments

ahesford:
There's no need to delete this, you just need to force-push over the branch.

droc12345:
I wasn't sure how to fix, so I just deleted and readded the patch and created a new pull #1485

PR #1485, opened 2024-01-29
desktop: support switching to windows in other workspaces

droc12345:
Ok, 2nd try on pull *lol* 
this time with a more proper title, etc

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -106,18 +106,6 @@ desktop_focus_view_or_surface(struct seat *seat, struct view *view,
 	}
 }
 
-static struct wl_list *
-get_prev_item(struct wl_list *item)
-{
-	return item->prev;
-}
-
-static struct wl_list *
-get_next_item(struct wl_list *item)
-{
-	return item->next;
-}
-
 static struct view *
 first_view(struct server *server)
 {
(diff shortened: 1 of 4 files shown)


src/desktop.c
```diff
@@ -167,48 +155,52 @@ desktop_cycle_view(struct server *server, struct view *start_view,
 	 *   - Otherwise select the view second from the top
 	 */
 
+	int views = wl_list_length(&server->views);
+
 	/* Make sure to have all nodes in their actual ordering */
 	osd_preview_restore(server);
 
 	if (!start_view) {
-		start_view = first_view(server);
```
johanmalm, line 174: Should delete `first_view()`

src/osd.c
```diff
@@ -409,9 +409,7 @@ display_osd(struct output *output)
 	struct wl_array views;
 	wl_array_init(&views);
 	view_array_append(server, &views,
-		LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
```
johanmalm, line 412: It'd be better to take a step at a time - and only optionally remove this criteria.

johanmalm (commented):
Good effort on submitting the patch.

It works, but I have a feeling we could simplify it.

johanmalm:
image

This is how I visualize the linked list.
How do you get past the case when elm==head?

droc12345:
Head is actually start_view pointing to link.

If it's equal to itself, then there is only one view, and it's not allowed to get to the iter function if there is less than 2 views. At least that's my understanding.

first_view isn't needed (nothing else calls it) since I replaced the logic inside the 
if null start_head with different logic.

LAB_VIEW_CRITERIA_CURRENT_WORKSPACE I'm not sure if it's necessary to leave it in place or not.
I had thought that it was keeping me from seeing all views, but that was early in my work and I never went
back and tried it once I got logic working right. 

Take what you need from this and play with it, I haven't tested prev window, I didn't think about it when 
testing as I don't use that action. But it's the same thing as next window except for the direction.

Anyway, it does need more testing and some things could be streamlined.

droc12345:
So the logic is 
desktop_cycle_view is called with start_view (which may be null)
get number of server views

if start_view is null
.for_each_view server views
..if current workspace not eq workspace we found (imply one view not on current workspace)
...workspaces_switch_to with follow
..start_view eq view found and break for loop
.if start_view still null or number of server views is less than 2
..return start_view

set criteria, direction, etc
.check again for server views less than 2 and return start_view
 (redundant but the above check could be moved below the if and then this wouldn't be needed)

for_each_view_iter (we know we have more than one view)

There is no sense in trying to look for next view when there is only one.

That particular problem stumped me for a while, as that's where the crashes
came from

droc12345:
This isn't needed now, so closing

PR #1487, opened 2024-01-29
xdg: make sure wlroots knows client size before activation

ahesford:
Calling the `wlr_xdg_toplevel_set_size()` function will cause wlroots to record the size in its `wlr_xdg_toplevel.scheduled` structure, which is never cleared after an XDG configure event is sent. All subsequent XDG configure events will include this size. This means that `wlr_xdg_toplevel_set_activated` will send an activation event that includes the last-configured size.

If a client resizes itself after a call to `wlr_xdg_toplevel_set_size()`, the client will trigger a surface commit that calls labwc's `handle_commit` callback, and we record the proper `view->current` (and `view->pending`) to reflect the client's actual size. However, wlroots still holds the last value set by `wlr_xdg_toplevel_set_size()`, which means that any subsequent XDG configure will sent a different size than the client set, and likely cause the client to resize itself in response to the request. To prevent this, we can call `wlr_xdg_toplevel_set_size()` with the client's actual size right before any actual configure request we wish to make. This is the approach taken by, *e.g.*, [croissant](https://codeberg.org/vyivel/croissant/src/commit/ed62193ccc0d7b3b2ff49f5ec6b560901438cc4a/compositor/layout/xdg-toplevel.c#L168) and [river](https://github.com/riverwm/river/blob/69a51cadb41443103370247ac515f1067da4b932/river/XdgToplevel.zig#L153).

I believe `wlr_xdg_toplevel_set_activated` is the only case where we send an "unguarded" configure request; other configuration requests (to fullscreen or maximize a view, or change its size explicitly) will all schedule a configure request that will be immediately followed up with a call to `wlr_xdg_toplevel_set_size()`. In all cases, I expect these rapid-fire calls will be coalesced into a single event, so the order in which they are called is probably not important.

Another option would be to manage this in `handle_commit`, but the drawback to doing so is that `wlr_xdg_toplevel_set_size()` will actually trigger a size event, which will then be sent back to the client. At best, this seems wasteful; at worst, I imagine some poorly behaved client geting stuck in a loop where it keeps changing its size based on the event, which triggers another callback, which triggers another event, *ad infinitum*.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -142,6 +142,29 @@ handle_commit(struct wl_listener *listener, void *data)
 		 */
 		if (!view->pending_configure_serial) {
 			view->pending = view->current;
+
+			/*
+			 * wlroots retains the size set by any call to
+			 * wlr_xdg_toplevel_set_size and will send the retained
+			 * values with every subsequent configure request. If a
+			 * client has resized itself in the meantime, a
+			 * configure request that sends the now-outated size
+			 * may prompt the client to resize itself unexpectedly.
+			 *
+			 * Calling wlr_xdg_toplevel_set_size to update the
+			 * value held by wlroots is undesirable here, because
+			 * that will trigger another configure event and we
+			 * don't want to get stuck in a request-response loop.
+			 * Instead, just manipulate the dimensions that *would*
+			 * be adjusted by the call, so the right values will
+			 * apply next time.
+			 *
+			 * This is not ideal, but it is the cleanest option.
+			 */
+			struct wlr_xdg_toplevel *toplevel =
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM

ahesford:
On IRC, @Consolatis made a convincing argument that the most robust approach would be to handle this in `handle_commit`, but that we should avoid triggering another event by just setting the dimensions in  `wlr_xdg_toplevel.scheduled` directly. It would be nice if this werent' so low-level, but I agree it's the best option for now.

Consolatis:
LGTM, not tested

johanmalm:
Thanks.

PR #1488, opened 2024-01-30
view: better honor usable areas when changing outputs in view_move_to_edge

ahesford:
If a view is fully within the usable area of its original output, it should be fully within the usable area (if possible) on its new output.

This addresses [an issue discovered by @johanmalm](https://github.com/labwc/labwc/pull/1482#issuecomment-1917587158) where the `MoveToEdge` action kicks a view to another output and covers a panel.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -1540,6 +1540,42 @@ view_get_adjacent_output(struct view *view, enum view_edge edge)
 	return output;
 }
 
+static int
+shift_view_to_usable_1d(int size,
+		int cur_pos, int cur_lo, int cur_extent,
+		int next_pos, int next_lo, int next_extent,
+		int margin_lo, int margin_hi)
+{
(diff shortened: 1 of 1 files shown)


src/view.c
```diff
@@ -1603,26 +1639,19 @@ view_move_to_edge(struct view *view, enum view_edge direction, bool snap_to_wind
 		return;
 	}
 
-	/* If more than half the view is right of usable region, align to right */
-	int midpoint = destination_x + view->pending.width / 2;
+	struct wlr_box original_usable =
```
johanmalm, line 1642: Why the `original_usable` here? Can't spot the difference with `usable`

ahesford: `usable` is the area of the output that will receive the view, `original_usable` is the area of the output that currently holds the view.

johanmalm:
LGTM.
Nice approach using a 1d helper!

PR #1489, opened 2024-01-30
NEWS.md: note breaking changes in snapping behavior

ahesford:
This could probably use some refinement for clarity.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -121,8 +121,34 @@ The format is based on [Keep a Changelog]
   caused by applications choosing not respond to pending resize requests either
   by ignoring them or substituting alternative sizes (for example, when mpv
   constrains resizes to keep its aspect ratio fixed). Written-by: @ahesford
-    
+
 ### Changed
+
+- When a Wayland-native window is snapped to a screen edges or user-defined
+  region, labwc will notify the application that it is "tiled", allowing the
+  application to better adapt its rendering to constrained layouts. Windows
+  with client-side decorations may respond to these notices by squaring off
+  corners and, in some cases, disabling resize abilities. This can be disabled
+  with:
+
+```xml
+<snapping>
+  <notifyClient>never</notifyClient>
+</snapping>
+```
+
+  or limited to only edge-snapped or only region-snapped windows. See the
+  labwc-config(5) manual page for more information.
+
+- When a window is dragged from a snapped position (either a screen edge or a
+  user-defined region), the snapped state is now discarded as soon as the
+  dragging begins. This means that dragging from a snapped position to a
+  maximized state (with the `topMaximize` option enabled) and then
+  un-maxmimizing the window will restore the window to its size and position
+  *before* it was snapped. In previous releases, un-maximizing would restore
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -124,6 +124,32 @@ The format is based on [Keep a Changelog]
     
 ### Changed
 
+- **Breaking change**: When a Wayland-native window is snapped to a screen
```
johanmalm, line 127: No need for `**Breaking change**`. By definition, anything in the change section refers to `changes in existing functionality`.

NEWS.md
```diff
@@ -124,6 +124,32 @@ The format is based on [Keep a Changelog]
...
+  edges or user-defined region, labwc will notify the application that is is
```
johanmalm, line 128: is is

johanmalm (commented):
Two minors.
Other than that, LGTM

NEWS.md
Consolatis: Maybe we could move the `Changed` to the top? At least in my case I usually don't read through all the detailed changes on updates but that section would be important to see.

Also renaming the whole section to `Breaking changes` (or something similar) might help as well to differentiate the importance of the sections.

ahesford: I moved the "Changes" section up top, called it "Breaking Changes", and also reworded the discussion about `app_id` for good measure.

johanmalm: > Maybe we could move the `Changed` section to the top? At least in my case I usually don't read through all the detailed changes on updates but that section would be important to see.

I prefer to keep the order it as it was. We've been sticking with https://keepachangelog.com/en/1.0.0/ for a couple of years, and I think the consistent pattern is important.

We have a `Changed` section for pretty much every release and the entries within these all describe changes to functionality which are "breaking" to to some degree.

> Also renaming the whole section to `Breaking changes` (or something similar) might help as well to differentiate the importance of the sections.

How about adding a note at the top about the `### Changed` label explaining that it refers to breaking changes and that we go to great lengths to avoid these, but if there is a strong argument in aid of consistency, code correctness or maintainability then we accept them - particularly if the breaking change has a low impact or small footprint.
Maybe we should also describe that whilst `labwc` is not a library, it is a platform upon which people build other things,  and that we value stability and a consistent interfaces.

Problems with changing to `Breaking Changes` include (a) that `Added`, `Changed`, etc are verbs in past tense so it breaks the pattern and I can't think of a sensible way of including `Breaking` with a verb. Also it breaks the keepachangelog standard.

ahesford: Ok, back to the original order and heading, but I kept the rewrites from the last revision.

I don't have a strong opinion about some disclaimer text at the top. While it's important to point out behavioral changes that might intrude on somebody's workflow, a statement about preferring to keep stable interfaces probably just invites *post hoc* criticism from people who disagree with arguments for improvement of some feature or another and would prefer old behavior.

johanmalm: > a statement about preferring to keep stable interfaces probably just invites post hoc criticism from people who disagree with arguments for improvement of some feature or another and would prefer old behavior

Yes, good point. Let's leave as you've done it.

Issue #1474, opened 2024-01-24
Strategy for supporting `_NET_WM_WINDOW_TYPE_DESKTOP`


johanmalm:
...and other net-wm-window-type atoms not already covered by wlroots.

Ref:
  - #1185
  - #1472
  - https://github.com/labwc/labwc/blob/59543e476c9b404a8a8dbfd753a0d438a42308f5/docs/rc.xml.all#L519-L525
  - https://github.com/swaywm/sway/blob/master/sway/criteria.c

Assuming we merge #1185, how would we actually implement support for something like `_NET_WM_WINDOW_TYPE_DESKTOP`.

Looking at the window-rules for `pcmanfm-qt --desktop` running in Wayland mode, we would need to consider
  - `<skipTaskbar>`
  - `<skipWindowSwitcher>`
  - `<fixedPosition>`
  - `<action name="ToggleAlwaysOnBottom"/>

Action like `ToggleAlwaysOnBottom` could be processed in `xwayland.c`'s handle_set_window_type() (introduced by #1185), but properties like `fixedPosition` would have to be checked for in the same positions where we currently do `window_rules_get_property()`, with the exception of some things which could be taken care of in `xwayland_view_wants_focus()`.

Considering the amount of window-rule checks (see below), I think it would be ugly to add `xwayland_surface_contains_window_type()` + the `#if HAVE_XWAYLAND` guards in all those places.

`git grep window_rules_get_property`:

```
src/desktop.c:          enum property skip = window_rules_get_property(view, "skipWindowSwitcher");
src/interactive.c:      if (window_rules_get_property(view, "fixedPosition") == LAB_PROP_TRUE
src/view-impl-common.c: enum property ret = window_rules_get_property(view, "skipTaskbar");
src/view.c:             if (window_rules_get_property(view, "skipWindowSwitcher") == LAB_PROP_TRUE) {
src/view.c:     if (window_rules_get_property(view, "fixedPosition") == LAB_PROP_TRUE
src/xdg.c:      switch (window_rules_get_property(view, "serverDecoration")) {
src/xdg.c:      if (window_rules_get_property(view, "ignoreFocusRequest") == LAB_PROP_TRUE) {
src/xwayland.c: switch (window_rules_get_property(view, "serverDecoration")) {
src/xwayland.c: if (window_rules_get_property(view, "ignoreFocusRequest") == LAB_PROP_TRUE) {
```

So, how do we do it?

[...]

PR #1472, opened 2024-01-24
Respect _NET_WM_WINDOW_TYPE_DESKTOP atom

grisha128:
Hi, I'd like to improve condition of desktop file managers (like `pcmanfm --desktop`) in Labwc and Wayland in general.

Back in X11, they indicated that they are desktops with _NET_WM_WINDOW_TYPE_DESKTOP atom.

In Wayland, there's a layer-shell obviously, but it doesn't come with Qt suit. Folks over KDE developed some 3rd-party component, but I doubt it will ever be stable enough for use by regular people/in enterprise: I encountered pretty flimsy bug [480240](https://bugs.kde.org/show_bug.cgi?id=480240) immediately after integrating this component into my Qt project. This is by no means a stone thrown in the garden of layer-shell-qt developers - Qt is too big and fast moving to be able to keep up with it for basically anyone.

There's a not very clean solution of writing window rules for these desktops, like:

```
<windowRule identifier="...">
	<skipTaskbar>yes</skipTaskbar>
	<skipWindowSwitcher>yes</skipWindowSwitcher>
	<fixedPosition>yes</fixedPosition>
	<action name="Maximize"/>
	<action name="ToggleDecorations"/>
	<action name="ToggleAlwaysOnBottom"/>
</windowRule>
```

But these rules are not in every project's goals. For example, [hyprland](https://github.com/hyprwm/Hyprland) seems to be reluctant of it ([#2093](https://github.com/hyprwm/Hyprland/issues/2093)).

So my proposal for solution is adding support for _NET_WM_WINDOW_TYPE_DESKTOP in wlroots and asking compositor developers to check for this atom in their xwayland code. Since I'm a user of labwc, I already made the patch for this. Now we are waiting for wlroots folks. As far as I could test, the only change required for wlroots is:

```diff
diff --git a/include/xwayland/xwm.h b/include/xwayland/xwm.h
index da3515d..66108e2 100644
--- a/include/xwayland/xwm.h
+++ b/include/xwayland/xwm.h
@@ -72,6 +72,7 @@ enum atom_name {
 	NET_WM_WINDOW_TYPE_MENU,
 	NET_WM_WINDOW_TYPE_NOTIFICATION,
 	NET_WM_WINDOW_TYPE_SPLASH,
+	NET_WM_WINDOW_TYPE_DESKTOP,
 	DND_SELECTION,
 	DND_AWARE,
 	DND_STATUS,
diff --git a/xwayland/xwm.c b/xwayland/xwm.c
[...]

PR #1490, opened 2024-01-30
xwayland: add sensible rules for NET_WM_WINDOW_TYPE_DESKTOP windows

grisha128:
Reference:
* https://github.com/labwc/labwc/pull/1185
* https://github.com/labwc/labwc/pull/1472
* https://github.com/labwc/labwc/issues/1474

I tried to avoid using `HAVE_XWAYLAND` macro according to @johanmalm idea:

> @Consolatis and I brainstormed a bit on IRC and thought we might be able to extend the window-rule criteria to accept a wlr_surface rather than identifier, etc. and just add window rules for properties in the set_type handler? If that works, it would potentially be quite neat and self-contained within xwayland.c We'd probably have to deal with set_type()s happening after map().

But couldn't find a better way than this... If it's not good enough I'm ready to cooperate ;)

CC: @johanmalm @Consolatis @jlindgren90


grisha128:
By the way, `set_window_type` handler works, but unfortunately it does much earlier than there's a surface available so I couldn't add window rule with wlr_surface criteria in this handler.

Consolatis:
I actually think that this is even better than to create internal wlr_surface rules as we only need to setup the rule(s) once and it just magically works for all xwayland surfaces. I didn't check in details but I assume that we have to react to the `Reconfigure` event to keep the automatically added rule(s) around.

We might also want to add another function pointer to `struct view_impl` so we don't need the `HAVE_XWAYLAND` check in `view_matches_criteria()`. Something like
```c
struct view_impl {
  ..
  bool (*contains_window_type)(int32_t window_type);
}
// which can then be called like
if (rule->window_type >= 0 && view->impl->contains_window_type) {
  return view->impl->contains_window_type(rule->window_type);
}
```

In general I really like this approach! Thanks for working on it.

grisha128:
Thank you for a review :)

I'm not sure what's wrong with CI - builds fine on my side.

src/config/rcxml.c
```diff
@@ -1101,6 +1103,10 @@ rcxml_init(void)
 
 	rc.workspace_config.popuptime = INT_MIN;
 	rc.workspace_config.min_nr_workspaces = 1;
+
+#ifdef HAVE_XWAYLAND
+	xwayland_create_window_type_rules();
+#endif
```
Consolatis, line 1109: > I'm not sure what's wrong with CI - builds fine on my side.

Should be `if` rather than `ifdef`

src/xwayland.c
```diff
@@ -32,6 +34,17 @@ xwayland_surface_contains_window_type(
 	return false;
 }
 
+static bool
+view_contains_xwayland_window_type(struct view *view, int32_t window_type)
```
Consolatis, line 38: Sugget changing to `xwayland_view_contains_window_type` or something like that, functions starting with `view_` should usually only be defined in `src/view.c`.

src/xwayland.c
```diff
@@ -32,6 +34,17 @@ xwayland_surface_contains_window_type(
...
+{
+	assert(view);
+	if (view->type == LAB_XWAYLAND_VIEW) {
+		struct wlr_xwayland_surface *surface = xwayland_surface_from_view(view);
+		return xwayland_surface_contains_window_type(surface, window_type);
+	}
```
Consolatis, line 44: No need for the `view->type` check here, if we end up in this function it must be a xwayland view as the wayland implementation (`xdg.c`) doesn't set the window pointer. If we ever end up adding something similar to the wayland variant we can always add the check back in here.

However to detect anything crazy going on I suggest adding and additional `assert(view->type == LAB_XWAYLAND_VIEW);` after the `view` assert.

src/xwayland.c
```diff
@@ -32,6 +34,16 @@ xwayland_surface_contains_window_type(
...
 }
 
+static bool
+xwayland_view_contains_window_type(struct view *view, int32_t window_type)
+{
+	assert(view);
+	assert(view->type == LAB_XWAYLAND_VIEW);
+	struct wlr_xwayland_surface *surface = xwayland_surface_from_view(view);
+	return xwayland_surface_contains_window_type(surface, window_type);
+	return false;
```
Consolatis, line 44: Superfluous `return false`

Consolatis:
Other than the last minor nitpick above this looks very good to me.

I do not merge the PR right away to give @johanmalm and possibly @jlindgren90 a chance to state their opinions.

grisha128:
Just figured that `ToggleFullscreen` action is quite handy for this.

Consolatis:
> Just figured that `ToggleFullscreen` action is quite handy for this.

With that one I am not quite sure:
- it disables top layer layershell (wayland native) panels
- It might actually unfullscreen a desktop client that starts up fullscreen by itself
- It doesn't adjust its geometry when a new X11 or wayland native panel requests exclusive screen space

So maybe a `Maximize` (rather than `ToggleMaximize`) might be better suited here?

grisha128:
> > Just figured that `ToggleFullscreen` action is quite handy for this.
> 
> With that one I am not quite sure:
> 
>     * it disables top layer layershell (wayland native) panels
> 
>     * It might actually unfullscreen a desktop client that starts up fullscreen by itself
> 
>     * It doesn't adjust its geometry when a new X11 or wayland native panel requests exclusive screen space
> 
> 
> So maybe a `Maximize` (rather than `ToggleMaximize`) might be better suited here?

The thing is that when I use `Maximize` it doesn't turn off decorations despite having `server_decoration` set to `LAB_PROP_FALSE`. I tested it really quick and found out that it happens for me because I have

```
<windowRule identifier="*"><action name="ToggleDecorations"/></windowRule>
```

in rc.xml. It toggles decorations for `NET_WM_WINDOW_TYPE_DESKTOP` window as well, which is not nice.

This line fixed it though %)

```
<windowRule identifier="*" serverDecoration="no"></windowRule>
```

jlindgren90:
Sorry I haven't had time to really follow this thread or look at the changes. Don't let me hold up progress :)

johanmalm:
Nice work.

Looks good in principle, except that I'm nervous about hard-coding properties and actions with no ability to configure at run-time. I think it's good to pre-configure `labwc` to work out of the box if a generic rule is possible, but my two comments above (against `ignore_focus_request` and `Maximize`) make me thing that a user/maintainer should be able to add/delete these rules via `rc.xml` too.

I could do some testing Thu/Fri. 
Could you help by recommending some clients that use `_NET_WM_WINDOW_TYPE_DESKTOP`. I guess `conky` and `pcmanfm --desktop` use it, but am not sure.

jlindgren90:
xfdesktop4 would be a well-known client to test

Consolatis:
> but my two comments above (against `ignore_focus_request` and `Maximize`)

I don't see them in this PR, stuck review?

> make me think that a user/maintainer should be able to add/delete these rules via rc.xml too.

The internal rules have the lowest priority so a user could modify them by targeting `->identifier` or `->title`. However I sympathize with your general observation. We should support them for `rc.xml` and then throw the hardcoded rules into `rc.xml.all` as well. Same thing as before, they have the lowest priority so if a user would then configure their own rule targeting `_DESKTOP` (or `_DOCK` or whatever) they would be matched first and thus be applied.

grisha128:
> Nice work.

Thank you :^)

> Looks good in principle, except that I'm nervous about hard-coding properties and actions with no ability to configure at run-time. I think it's good to pre-configure labwc to work out of the box if a generic rule is possible, but my two comments above (against ignore_focus_request and Maximize) make me thing that a user/maintainer should be able to add/delete these rules via rc.xml too.

I'm not sure if revealing this functionality to rc.xml makes a lot of use. For example, IceWM handles many different window types internally and has only title and class for window rule criteria. I reckon it's been like that for ages and no one seem to complain. Moreover, as @Consolatis said, user always has the ability to override these rules by title or identifier.

> I could do some testing Thu/Fri. Could you help by recommending some clients that use _NET_WM_WINDOW_TYPE_DESKTOP. I guess conky and pcmanfm --desktop use it, but am not sure.

```
pcmanfm --display=:0 --desktop
```

```
QT_QPA_PLATFORM=xcb pcmanfm-qt --desktop
```

```
conky
```

(`--display=:0` and `QT_QPA_PLATFORM=xcb` force apps to run in X11 mode)

I couldn't get `xfdesktop` to start.

johanmalm:
> I don't see them in this PR, stuck review?

It was :smile:

src/xwayland.c
```diff
@@ -979,6 +991,25 @@ handle_xwm_ready(struct wl_listener *listener, void *data)
...
+void
+xwayland_create_window_type_rules(void)
+{
+	// Desktop window type
+	struct window_rule *rule = znew(*rule);
+	rule->window_type = NET_WM_WINDOW_TYPE_DESKTOP;
+	rule->server_decoration = LAB_PROP_FALSE;
+	rule->skip_taskbar = LAB_PROP_TRUE;
+	rule->skip_window_switcher = LAB_PROP_TRUE;
+	rule->ignore_focus_request = LAB_PROP_TRUE;
```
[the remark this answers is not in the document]
grisha128: I'm not so sure either, but as far as I can understand ignore_focus_request setting makes it so that window unable to take over user focus? If that's the case, can you give an example in which desktop window might want to take over user focus? In my opinion it's not very convenient for users to stumble upon desktop which randomly focuses itself

src/xwayland.c
```diff
@@ -979,6 +991,25 @@ handle_xwm_ready(struct wl_listener *listener, void *data)
...
+	rule->fixed_position = LAB_PROP_TRUE;
+	struct action *action_maximize = action_create("Maximize");
```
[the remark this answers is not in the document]
grisha128: I tested it with transparent `waybar` and regardless of what action was set, `Maximize` or `MoveTo`, desktop surface was always shifted away from the panel, i.e. desktop didn't stay behind transparent panel. I guess that's how layer shell works? So if there's no difference, I guess it's better to go with a shorter code (`MoveTo` version is 2 lines more than `Maximize`):

```diff
diff --git a/src/xwayland.c b/src/xwayland.c
index e468c01..113ca1c 100644
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -1002,10 +1002,12 @@ xwayland_create_window_type_rules(void)
 	rule->skip_window_switcher = LAB_PROP_TRUE;
 	rule->ignore_focus_request = LAB_PROP_TRUE;
 	rule->fixed_position = LAB_PROP_TRUE;
-	struct action *action_maximize = action_create("Maximize");
+	struct action *action_move_to = action_create("MoveTo");
 	struct action *action_on_bottom = action_create("ToggleAlwaysOnBottom");
+	action_arg_add_str(action_move_to, "x", "0");
+	action_arg_add_str(action_move_to, "y", "0");
 	wl_list_init(&rule->actions);
-	wl_list_append(&rule->actions, &action_maximize->link);
+	wl_list_append(&rule->actions, &action_move_to->link);
 	wl_list_append(&rule->actions, &action_on_bottom->link);
 	wl_list_append(&rc.window_rules, &rule->link);
 }
```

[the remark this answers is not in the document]
grisha128: > Also I think X desktop clients can work out their own size relative to the screen and may choose not to take up the whole screen.

Well, if they don't take up the whole screen, they will want to choose their position on the screen. As far as I understand in Wayland it's not possible and `conky -x 100 -y 100` doesn't work.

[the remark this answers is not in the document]
grisha128: Okay, `conky` case demonstrated that `MoveTo` approach is really better. It's able to position itself after `MoveTo` action and can choose not to take up the whole screen.

src/xwayland.c
```diff
@@ -979,6 +991,27 @@ handle_xwm_ready(struct wl_listener *listener, void *data)
...
+	rule->window_type = NET_WM_WINDOW_TYPE_DESKTOP;
+	rule->server_decoration = LAB_PROP_FALSE;
+	rule->skip_taskbar = LAB_PROP_TRUE;
+	rule->skip_window_switcher = LAB_PROP_TRUE;
+	rule->ignore_focus_request = LAB_PROP_TRUE;
+	rule->fixed_position = LAB_PROP_TRUE;
+	struct action *action_on_bottom = action_create("ToggleAlwaysOnBottom");
+	struct action *action_move_to = action_create("MoveTo");
+	action_arg_add_str(action_move_to, "x", "0");
+	action_arg_add_str(action_move_to, "y", "0");
```
Consolatis, line 1008: These should be `action_arg_add_int()` but it doesn't really matter that much as they default to `0` anyway. So these could also be removed.

src/xwayland.c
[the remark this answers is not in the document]
johanmalm: I haven't tested, but imagine an desktop client showing icons (like `pcmanfm --desktop` and `xfdesktop`) would want to take focus to operate right-click menus, rename icons and so on.

johanmalm:
I'll do some testing with clients tomorrow.

src/xwayland.c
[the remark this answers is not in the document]
grisha128: Yeah, it needs focus to do that of course, but it's user-initiated focus, rather than focus requested by window, isn't it? Sorry if I'm talking nonsense, that's just what I figured from the setting name without looking deep into implementation. This is also confirmed by the fact that I can navigate right-click menus and rename files with this setting set to true.

src/window-rules.c
```diff
@@ -67,6 +67,8 @@ view_matches_criteria(struct window_rule *rule, struct view *view)
 			return false;
 		}
 		return match_glob(rule->title, title);
+	} else if (rule->window_type >= 0 && view->impl->contains_window_type) {
+		return view->impl->contains_window_type(view, rule->window_type);
```
xi, line 71: Is it intentional that `window_type` conditions are ignored if any other conditions are provided?

Consolatis: Likely not.

There should be some `bool matches = false` on top and all conditions should result in `matches |= condition` and at the end we return `matches`. However, a condition being absent from the rule should not changes the `matches` bool.

xi: see 327dd48

txgk: Thanks, should be fixed in https://github.com/labwc/labwc/pull/1490/commits/33cb3d47f8cab58166ce216dc4470afa85e7ec78

grisha128: Thanks, should be fixed in https://github.com/labwc/labwc/pull/1490/commits/33cb3d47f8cab58166ce216dc4470afa85e7ec78

Consolatis: First of all, sorry for the delay in this PR.

It seems like #1731 will get merged first (as it doesn't add the internal rules), @grisha128 are you fine with my Co-Authored-By suggestion from https://github.com/labwc/labwc/pull/1731#issuecomment-2066839523 ?

When #1731 is merged, would you be fine with rebasing your work on top? That should also solve the "no way for users to override the internal rules" issue.

grisha128: > First of all, sorry for the delay in this PR.

No problem, all is fine ;)

> It seems like https://github.com/labwc/labwc/pull/1731 will get merged first (as it doesn't add the internal rules), @grisha128 are you fine with my Co-Authored-By suggestion from https://github.com/labwc/labwc/pull/1731#issuecomment-2066839523 ?

I guess that's okay!

> When https://github.com/labwc/labwc/pull/1731 is merged, would you be fine with rebasing your work on top? That should also solve the "no way for users to override the internal rules" issue.

Will do, thank you <3

grisha128:
Rebased on top of master with the work from https://github.com/labwc/labwc/pull/1731 and tested these:

> ```
> pcmanfm --display=:0 --desktop
> ```
> 
> 
>     
>   
> 
> ```
> QT_QPA_PLATFORM=xcb pcmanfm-qt --desktop
> ```
> 
> 
>     
>   
> 
> ```
> conky
> ```

johanmalm:
Are we sure ignore_focus_request should be true. 

We should be able to operate pcmanfm--desktop context menus with the keyboard, which necessitates keyboard-focus.

Consolatis:
> Are we sure ignore_focus_request should be true.
> 
> We should be able to operate pcmanfm--desktop context menus with the keyboard, which necessitates keyboard-focus.

That is only for the window requesting focus itself. AFAIR it should receive keyboard focus just fine when being clicked on.

txgk:
> Are we sure ignore_focus_request should be true.
>
> We should be able to operate pcmanfm--desktop context menus with the keyboard, which necessitates keyboard-focus.

Since context menus usually appear only after user's mouse clicks which set focus, it should work just fine.

If you believe some windows of desktop type could pick up user focus without any input at first, the rule strictness can be lessen then.

txgk:
> Are we sure ignore_focus_request should be true.

We should be able to operate pcmanfm--desktop context menus with the keyboard, which necessitates keyboard-focus.

Since context menus usually appear only after user's mouse clicks which set focus, it should work just fine.

If you believe some windows of desktop type could pick up user focus without any input at first, the rule strictness can be lessen then.

grisha128:
> Are we sure ignore_focus_request should be true.
>
> We should be able to operate pcmanfm--desktop context menus with the keyboard, which necessitates keyboard-focus.

Since context menus usually appear only after user's mouse clicks which set focus, it should work just fine.

If you believe some windows of desktop type could pick up user focus without any input at first, the rule strictness can be lessen then.

johanmalm:
> > Are we sure ignore_focus_request should be true. We should be able to operate pcmanfm--desktop context menus with the keyboard, which necessitates keyboard-focus.
> 
> Since context menus usually appear only after user's mouse clicks which set focus, it should work just fine. If you believe some windows of desktop type could pick up user focus without any input at first, the rule strictness can be lessen then.

Thanks for being patient :smile: ...and sorry for not keeping up with the conversation higher up. The `ignoreFocusRequest` thing makes sense now. I've probably spent too long on layer-shell clients and was stuck thinking that the clients would be able to `grab` focus at all (whereas this window-rules relates to the client requesting focus).

It feels like we ought to:

- [ ] Allow users to opt-out of this hard coded window-rule. With `<keybinds>` and `<mousebinds>` we've taken the approach that we will use builtin values if none were defined by the user. But the user can use the `<default>` child element to use the builtin values as a starting point to build on. We also have a `None` action to "clear" binds. I feel that window-rules are a bit more specialist, so maybe a simple `<windowRules><clear>` would be enough to destroy anything in `rc.window_rules`
- [ ] Document that a default window-rule exists for `NET_WM_WINDOW_TYPE_DESKTOP` and in big handfuls what it looks like.

grisha128:
@johanmalm Hi, sorry for the wait. I adjusted the patch to meet these points ;) Tested the `<clear />` element, seems to be working on my side!

Issue #1486, opened 2024-01-29
Resize-handles only appear on one screen when an application spans two windows


lurch:
I'm using Labwc (built from `master`) running on a Raspberry Pi 4. I have two monitors plugged in (one is 1440p and the other is 1080p - I don't think this has any impact on this bug, but mentioning it just in case!).
If I open a window e.g. file-manager, then if the window is entirely within the left-monitor or entirely within the right-monitor, then the window-resize handles work correctly on every part of the window border. But if I have the application window straddling both monitors, then the resize-handles only appear on one monitor, and on the other monitor the window doesn't have resize-handles on the window border. It appears to be the monitor that contains the "largest" part of the window, that's the one that gets the resize-handles.

I.e.
* if the window is 100% within monitor 1, then the resize-handles work everywhere on monitor 1
* if the window is 70% within monitor 1 and 30% within monitor 2, then the resize-handles work on monitor 1 but don't appear on monitor 2
* if the window is 30% within monitor 1 and 70% within monitor 2, then the resize-handles don't appear on monitor 1, but do work on monitor 2
* if the window is 100% within monitor 2, then the resize-handles work everywhere on monitor 2

Consolatis:
You mentioned a file-manager, does it use server or client side decorations and does the behavior differ between the two?

When using server side decorations there is an invisible resize area around the window. The resize cursor is set on both, the invisible resize area and the borders themselves. Is this happening for both or just the invisible area around the window?

lurch:
@spl237 tells me that we're using server-side decorations, and we're seeing the same behaviour for both the borders themselves and also the invisible resize area.

Consolatis:
Hm.. right. We restrict the invisible resize area to the usable area of `view->output` so we
- don't overlay bottom layer panels
- don't leak the invisible resize area into a neighboring output

I had a quick test nested and the window can actually be resized on the 2nd output but you'd have to hit the (potentially very small e.g. 1px wide) border itself to do so. 

I am not sure how to tackle this properly, we might need to have a list of outputs the window is currently visible on (currently we only have the one where the majority of the window is in) and then include that 2nd output in the condition.

I think we do have logic like that for the wlr-foreign-toplevel protocol already (so panels can be configured to only show the windows on the output where the panel is). Maybe we can generalize that one a bit and then use it for both. wlroots 0.18.x also brought changes to scene node internal output lists so maybe we could also use those directly.

With the following patch the invisible resize area can be debugged a bit easier:
```diff
[...]

PR #1491, opened 2024-01-31
src/ssd: allow invisible resize area across outputs

Consolatis:
This PR
- adds a new `view->outputs` bitset to store the calculation result of the intersection with \*all\* outputs the view is currently visible on. It is replacing the intersection in `src/foreign.c` to notify panels about the outputs so the result can be used for other cases like the invisible resize area around windows.
- uses the new `view->outputs` bitset in `src/ssd/ssd_extents.c`

This ensures that the invisible resize handle works across outputs
while still making sure that it won't leak into neighboring ones just
because it is positioned closely to an output edge (either manually,
maximized or snapped via SnapToEdge or SnapToRegion).

Fixes: #1486
Reported-By: @lurch

TODO:
- [x] remove the last `[wip]` commit that visualizes the invisible resize area to ease testing

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -124,7 +124,7 @@ struct view {
 	struct wl_list link;
 
 	/*
-	 * The output that the view is displayed on. Specifically:
+	 * The primary output that the view is displayed on. Specifically:
 	 *
 	 *  - For floating views, this is the output nearest to the
 	 *    center of the view. It is computed automatically when the
@@ -139,6 +139,16 @@ struct view {
 	 * by calling view_set_output() beforehand.
 	 */
 	struct output *output;
+
+	/*
+	 * The outputs that the view is displayed on.
+	 * This is used to notify the foreign toplevel
+	 * implementation and to update the SSD invisible
+	 * resize area.
+	 * It is a bitset of output->scene_output->index.
+	 */
+	uint64_t outputs;
+
 	struct workspace *workspace;
 	struct wlr_surface *surface;
 	struct wlr_scene_tree *scene_tree;
@@ -460,6 +470,7 @@ void view_move_to_front(struct view *view);
 void view_move_to_back(struct view *view);
 struct view *view_get_root(struct view *view);
 void view_append_children(struct view *view, struct wl_array *children);
+bool view_on_output(struct view *view, struct output *output);
 
 /**
  * view_is_related() - determine if view and surface are owned by the
--- a/src/foreign.c
+++ b/src/foreign.c
@@ -107,11 +107,10 @@ void
 foreign_toplevel_update_outputs(struct view *view)
 {
 	assert(view->toplevel.handle);
-	struct wlr_output_layout *layout = view->server->output_layout;
+
 	struct output *output;
 	wl_list_for_each(output, &view->server->outputs, link) {
-		if (output_is_usable(output) && wlr_output_layout_intersects(
-				layout, output->wlr_output, &view->current)) {
+		if (view_on_output(view, output)) {
 			wlr_foreign_toplevel_handle_v1_output_enter(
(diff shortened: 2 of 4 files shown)


johanmalm:
Nice work!

I like the bitset approach.

I don't follow the logic as to why only using the first intersecting rect is sufficient. Does it work if the output that a view is 'spilling over into' is to the right or below? Maybe I misunderstood the code.

Consolatis:
> I don't follow the logic as to why only using the first intersecting rect is sufficient. Does it work if the output that a view is 'spilling over into' is to the right or below? Maybe I misunderstood the code.

The first intersection rect is a combined rect across all usable areas the window is in. Usually there is only one intersection rect anway. However there are cases like when one of the outputs has a lower resolution or a panel at the bottom, different to the other one and the same invisible area is only partly visible on the first but completely visible on the second output. In that case the resize area on the 2nd output will be smaller than it could be. Otherwise we'd actually have to split for example the bottom scene rect into multiple ones with different heights which really complicates the code for a very small edge case.
Similar thing when having a panel on the right edge of an output but a view spanning both outputs, the result would be a "hole" in the usable area intersection which could only be solved by splitting the scene rect into multiple ones.

Example screenshot of the first case with a panel on the bottom of the left output:
20240131_23h09m33s_grim

johanmalm:
> The first intersection rect is a combined rect across all usable areas the window is in

Thanks. I forgot.

Just a thought (and possibly not a very well thought through one), in terms of general implementation approach, might `scene_surface_set_clip()` be worth considering. IIRC it's a fairly recent improvement in the scene-graph API. I think the logic would still be non-trivial, but possibly a bit simpler than your current proposal.

Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/surface.c?ref_type=heads#L279

Consolatis:
> Just a thought (and possibly not a very well thought through one), in terms of general implementation approach, might `scene_surface_set_clip()` be worth considering. IIRC it's a fairly recent improvement in the scene-graph API. I think the logic would still be non-trivial, but possibly a bit simpler than your current proposal.

That could indeed simplify some stuff, e.g. clipping the whole resize area \*tree\* to the usable areas the current view is visible on. However, it
- doesn't seem to support multiple clipping regions which then would require further calculations from our side to create a unified wlr_box that includes all usable areas. But that won't work if they differ in height (when they are positioned next to each other) or have holes between them (some exclusive layershell client in the edge between the outputs)
- it seems to only support scene surface trees which so it won't work for rects (only public function is `wlr_scene_subsurface_tree_set_clip()`

ahesford (commented):
If a view is dragged off all displays except for server-side decorations, its box will intersect no outputs and `view->outputs` will be zero. It might be better to force the bit corresponding to `view->output` in all cases. (This becomes moot if the view box is extended to account for SSD, which might offer some benefits but might also have some drawbacks.)

Consolatis:
> If a view is dragged off all displays except for server-side decorations, its box will intersect no outputs and `view->outputs` will be zero. It might be better to force the bit corresponding to `view->output` in all cases.

Is this a likely use-case for a user though? I imagine people rather minimizing a window rather than trying to move / resize it completely off the screen so only the decorations are still visible.

> (This becomes moot if the view box is extended to account for SSD, which might offer some benefits but might also have some drawbacks.)

The main issue with this in regards to the invisible resize area is that it will leak into neighboring outputs when being close to the edge. This doesn't happen if `view->outputs` only contains the outputs the actual window surface is on.

ahesford:
I agree that forcing the bit for `view->output` in `view->outputs` when the view is invisible and may not respond to configure requests is undesirable.

johanmalm:
LGTM

PR #1492, opened 2024-02-01
Edge snapping/resistance improvements

ahesford:
These changes combine two improvements to the edge engine for snapping and resistance behavior:
1. The first commit greatly simplifies edge detection for outputs. The existing implementation was a leftover from a draft where `validate_edges` was a function unique to both `resistance.c` and `snap.c` and, therefore, required that `edges_find_outputs` cast the generic region problem into a form that could be applied to both cases. Now that `validate_edges` is a common routine in `edges.c`, it is much cleaner to add a simpler `validate_output_edges` routine that reduces checks against imaginary, infinite edges.
2. The second commit is a quality-of-life improvement for `edges_find_neighbors` that will ignore snapping to any views on outputs that do not intersect the view being moved. First, this (might) have a slight performance advantage in multi-head configurations, where we trade off an extra few comparisons up front to avoid several other comparisons in `validate_edges` and the provided validator. Second (and most important), this "feels" better, because it seems weird that windows will snap to view edges on another output. (It also feels a bit weird that windows snap to edges of other views on the same output when they aren't actually touching, but I can appreciate that some people might like to maintain consistent grid lines across the output even if views aren't touching, and detecting these cases would be a lot more complicated anyway.)

### DIFF
--- a/src/edges.c
+++ b/src/edges.c
@@ -137,8 +137,10 @@ edges_find_neighbors(struct border *nearest_edges, struct view *view,
 	struct border view_edges = { 0 };
 	struct border target_edges = { 0 };
 
-	edges_for_target_geometry(&view_edges, view,
-		use_pending ? view->pending : view->current);
+	struct wlr_box *view_geom =
+		use_pending ? &view->pending : &view->current;
+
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM

PR #1493, opened 2024-02-01
edges: handle cases where neighbor spills into output of moving view

ahesford:
The fix in #1492 ignored cases when the neighbor being snapped is on another display but spills over *slightly* (so as not to affect its designated output) into the output of the view being moved. Rather than just checking that the view being moved or resized intersects with the output containing the neighbor, it is also necessary to check whether the neighbor intersects with the output containing the changing view.

Edit: this has been altered to use the new `view->outputs` bitmap; see comments below.

### DIFF
--- a/src/edges.c
+++ b/src/edges.c
@@ -383,6 +383,11 @@ edges_find_neighbors(struct border *nearest_edges, struct view *view,
 	assert(validator);
 	assert(nearest_edges);
 
+	if (!output_is_usable(view->output)) {
+		wlr_log(WLR_DEBUG, "ignoring edge search for view on unsable output");
+		return;
+	}
+
 	struct border view_edges = { 0 };
 	struct border target_edges = { 0 };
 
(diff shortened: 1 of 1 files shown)


ahesford:
I'm going to let this sit for now. When #1491 is merged, the whole intersection check can just be replaced with `view->outputs & v->outputs`.

ahesford:
I've altered this approach to use the new `view->outputs` bitmap rather than doing more geometry manipulations. This is now ready for review. The change both simplifies and improves the quality detection of neighbor edges that should be considered for resistance, attraction or `*ToEdge` actions. Note that testing things like equality of `view->output` as well as common bits in `view->outputs` is necessary because, in extreme cases when a view is dragged offscreen except for server-side decorations, `view->outputs` may be zero while `view->output` refers to a non-null controlling output.

There is still a corner case where a view is dragged offscreen except for its server-side decorations and straddles two outputs: in this case, moving another view on the "secondary" output (*i.e.*, not the one pointed to by `view->output`) will not encounter resistance at the portion of the SSD that encroaches on its output. However:
- This is almost certainly a very infrequent occurrence, and trying harder to detect it is probably not worthwhile.
- This case is currently not handled by the resistance implementation anyway.
- The change here still fixes detection of edges of views that straddle two outputs and leave at least one client pixel visible, so the overall UX is improved.

johanmalm:
LGTM. Conflict to be resolved.

PR #1499, opened 2024-02-02
edges: limit edge attraction and resistance...

ahesford:
...to edges actually encountered by motion during interactive moves and resizes.

A new `edges_traverse_edge` function determines if the motion of an edge (a move as well as a resize) actually encounters the finite edge of another region. This is used in the `check_edges` routine of `resistance.c` to limit considered edges to those that are actually traversed during an interactive move. The `check_edges` of `snap.c` still considers edges to be infinitely long, so keyboard actions that move windows will snap to an imaginary grid defined by the extension of all window edges. (This could be extended with, *e.g.*, a configuration option that optionall invokes `edges_traverse_edge` in that case as well.)

In addition, edges for minimized windows are ignored in all cases.

### DIFF
--- a/include/edges.h
+++ b/include/edges.h
@@ -50,6 +50,15 @@ edge_get_best(int next, int edge, bool decreasing)
 	return decreasing ? MAX(next, edge) : MIN(next, edge);
 }
 
+struct edge {
+	/* Position of an edge along the axis perpendicular to it */
+	int offset;
+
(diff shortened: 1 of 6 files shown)


johanmalm:
LGTM. Tested. Works well.

johanmalm:
Thanks :)

Issue #1504, opened 2024-02-04
Hover effect on SDD borders does not respect rounded corners


jp7677:
SSD borders, like being used in `gnome-terminal`, can be configured to have rounded corners in their decoration. When hovering with the mouse button over the `close` button, the hover effect however does not respect the rounded corners and stays square. This also means that the hover effect effects a few pixels outside of the border.

Consolatis:
Had some spare time to play around:
<details><summary>git diff</summary>

```diff
commit 89b384bd09ea938b554564349dbff65f04822089
Author: Consolatis <35009135+Consolatis@users.noreply.github.com>
Date:   Sun Feb 4 23:08:05 2024 +0100

    [wip] src/theme.c: use rounded hover overlays

diff --git a/src/theme.c b/src/theme.c
index 856889a8..e69051ad 100644
--- a/src/theme.c
+++ b/src/theme.c
@@ -47,6 +47,22 @@ struct button {
 	} active, inactive;
 };
 
+enum corner {
+	LAB_CORNER_UNKNOWN = 0,
+	LAB_CORNER_TOP_LEFT,
+	LAB_CORNER_TOP_RIGHT,
+};
+
+struct rounded_corner_ctx {
+	struct wlr_box *box;
+	double radius;
+	double line_width;
+	float *fill_color;
+	float *border_color;
+	enum corner corner;
+};
+static struct lab_data_buffer *rounded_rect(struct rounded_corner_ctx *ctx);
+
 static void
 drop(struct lab_data_buffer **buffer)
 {
@@ -57,9 +73,10 @@ drop(struct lab_data_buffer **buffer)
 }
 
 static void
-create_hover_fallback(struct theme *theme, struct lab_data_buffer **hover_buffer,
+create_hover_fallback(struct theme *theme, const char *name, struct lab_data_buffer **hover_buffer,
 		struct lab_data_buffer *icon_buffer)
 {
+	assert(name);
 	assert(icon_buffer);
 	assert(!*hover_buffer);
 
@@ -68,7 +85,6 @@ create_hover_fallback(struct theme *theme, struct lab_data_buffer **hover_buffer
 	int icon_width = cairo_image_surface_get_width(icon.surface);
 	int icon_height = cairo_image_surface_get_height(icon.surface);
 
-	/* TODO: need to somehow respect rounded corners */
 	int width = SSD_BUTTON_WIDTH;
 	int height = theme->title_height;
 
[...]

PR #1505, opened 2024-02-05
src/theme.c: use rounded hover overlays

Consolatis:
Fixes
- #1504

\+ some minor code improvements as 2nd commit.

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -65,7 +65,7 @@ struct rounded_corner_ctx {
 static struct lab_data_buffer *rounded_rect(struct rounded_corner_ctx *ctx);
 
 static void
-drop(struct lab_data_buffer **buffer)
+zdrop(struct lab_data_buffer **buffer)
 {
 	if (*buffer) {
 		wlr_buffer_drop(&(*buffer)->base);
@@ -275,8 +275,8 @@ load_buttons(struct theme *theme)
 	for (size_t i = 0; i < ARRAY_SIZE(buttons); ++i) {
 		struct button *b = &buttons[i];
 
-		drop(b->active.buffer);
-		drop(b->inactive.buffer);
+		zdrop(b->active.buffer);
+		zdrop(b->inactive.buffer);
 
 		/* PNG */
 		snprintf(filename, sizeof(filename), "%s-active.png", b->name);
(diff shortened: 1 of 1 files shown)


src/theme.c
```diff
@@ -101,9 +136,28 @@ create_hover_fallback(struct theme *theme, struct lab_data_buffer **hover_buffer
...
+	} else {
+		struct rounded_corner_ctx rounded_ctx = {
+			.box = &(struct wlr_box) { .width = width, .height = height },
+			.radius = rc.corner_radius,
+			.line_width = theme->border_width,
+			.fill_color = overlay_color,
+			.border_color = overlay_color,
+			.corner = corner
+		};
+		struct lab_data_buffer *overlay_buffer = rounded_rect(&rounded_ctx);
```
johanmalm, line 155: In `create_hover_fallback()`, could we just use `rounded_rect()` for all cases? Just change .radius and .corner depending on `LAB_CORNER_UNKNOWN/TOP_LEFT/TOP_RIGHT`.

I didn't study it for long, but my eyes were drawn to the creation of a `struct lab_data_buffer` for only purpose of calling `cairo_set_source_surface()` and `cairo_paint()`.

johanmalm (commented):
I'm happy with this approach and agree that a deep refactor is required, but in the meantime it's better to show rounded corner overlays when maximized then to go outside the window.

src/theme.c
Consolatis: > In `create_hover_fallback()`, could we just use `rounded_rect()` for all cases? Just change .radius and .corner depending on `LAB_CORNER_UNKNOWN/TOP_LEFT/TOP_RIGHT`.

Theoretically yes. However I didn't see the need for it (which includes creating a new cairo / lab_data_buffer buffer) if we have a cairo primitive which does the same.

> I didn't study it for long, but my eyes were drawn to the creation of a `struct lab_data_buffer` for only purpose of calling `cairo_set_source_surface()` and `cairo_paint()`.

Well, `rounded_rect()` returns a `lab_data_buffer` which we then paint over the centered icon on transparent background. We could change `rounded_rect()` to return a cairo buffer directly but I don't think that would really ease code maintenance nor would it improve performance. We could however supply an existing cairo buffer to `rounded_rect()` and if `!= NULL` the `lab_data_buffer` creation is skipped and the function draws directly onto the given cairo surface and returns `NULL`.

johanmalm: Good points. Let’s leave as is.

johanmalm:
LGTM

PR #1508, opened 2024-02-06
Refinements to edge behavior

ahesford:
This PR includes three commits that address issues in response to comments on IRC:
1. The first commit is a simple fix that avoids having the window attract to an edge, even when attraction is disabled, when a view approaches another view and then the direction of motion is reversed.
2. The second commit weakens the edge-traversal test, which avoids snapping a window "kitty-corner" to another window. This should reduce user surprise.
3. The third commit more closely mimics Openbox behavior, wherein resistance is only applied against "entry" into the space of another window, or departure of a window from an output. Note that attraction may still grab the trailing edge of a window as it slides past the edge of another view as it moves past the view, but a) this feels kind of natural, and b) preventing this kind of attraction would be extremely difficult without rethinking a lot of other edge-handling aspects.

Note that these changes should only apply to mouse-based interactive moves and resizes; action-based snapping should still favor snapping to *every* edge on the output, allowing for finer control of window placement and sizing on a virtual, irregular grid defined by other view boundaries.

cc: @Consolatis @johanmalm @domo141

### DIFF
--- a/include/edges.h
+++ b/include/edges.h
@@ -66,6 +66,7 @@ struct edge {
  * @target: position to which the moving edge will be moved
  * @oppose: opposing edge of encountered region
  * @align: aligned edge of encountered region
+ * @lesser: true if the moving edge is top or left, false otherwise
  *
  * This function will be used by edge_find_neighbors and edge_find_outputs to
  * validate and select the "best" output or neighbor edge against which a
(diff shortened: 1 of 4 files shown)


domo141:
Seems to work as I expect. Did not notice anything "strange"
have to read the description more carefully (and perhaps diffs)...

Consolatis:
Worked fine in my tests. Only issue left for me is that windows that are hidden behind  other windows also create resistance but we can deal with that in a separate PR. I didn't really follow the code changes for all of the new window resistance/snapping refactor so can't really say anything about the code changes itself but the result LGTM.

ahesford:
OK, let's defer solving the "snap to invisible windows" problem until we come up with either a good way to detect visibility, or a way to make it obvious to users why resistance is applied.

Issue #1412, opened 2024-01-05
Couple of issues with new VRR behavior


ScarecrowDM:
Hello! :)

**First Issue:**

While the new `<adaptiveSync>fullscreen</adaptiveSync>` works great (it triggers VRR with no problems), it always fails to disengage VRR when exiting fullscreen mode in any app/games that I commonly use (xnview, mpv, games, etc).

I don't believe this is the intended behavior.

**Second Issue:**

With labwc 0.7, I can just enable\disable VRR passing `wlr-randr --output DP-1 --adaptive-sync enabled/disabled` and it will work like a charm. 

This regressed and doesn't work anymore with current master.

johanmalm:
We've just committed this fix: https://github.com/labwc/labwc/commit/1b5af864023acedc7a84e4958436725d7017adef
Does that help?

If not, @Ph42oN any chance you could help with this?

ScarecrowDM:
Yes, the first issue is fixed by 1b5af864023acedc7a84e4958436725d7017adef ! :tada: 

Not sure if you guys want to address the second one, so feel free to close this issue.
It was more pressing for me because with VRR not turning off when exiting fullscreen, I had no way to disable VRR without changing labwc settings and restarting the session. But now with VRR turning off correctly, it might not be an issue.

Thanks!

Ph42oN:
> Not sure if you guys want to address the second one, so feel free to close this issue. It was more pressing for me because with VRR not turning off when exiting fullscreen, I had no way to disable VRR without changing labwc settings and restarting the session. But now with VRR turning off correctly, it might not be an issue.
> 
> Thanks!

Actually i see why this happens. Setting it with wlr-randr should still work if you use `<adaptiveSync>yes</adaptiveSync>`, but doesn't otherwise. I added `if` in `output_config_apply()` that prevents wlr-randr from changing adaptive sync.

Consolatis:
Should be fixed in latest master, thanks @Ph42oN.

PR #1512, opened 2024-02-07
Allow changing adaptive sync state always with wlr-randr

Ph42oN:
If that i added in output_config_apply in adaptive sync fullscreen patch prevents wlr-randr from changing adaptive sync state while its not set to yes in labwc config. So i think maybe better remove that if. Fixes 2nd issue in #1412.


Consolatis:
Thanks :)

PR #1516, opened 2024-02-10
edges: do not apply resistance to invisible edges

ahesford:
This is mostly the work of @Consolatis, in the [`feature/invisible_edge_snapping`](https://github.com/Consolatis/labwc/tree/feature/invisibile_edge_snapping) branch, to ignore invisible edges when applying resistance:
1. At the start of interactive operations on a particular view, the scene graph is walked to build, for each view, a bitfield of currently visible edges *without regard for the view receiving interaction*.
2. In `edges_find_neighbors`, all views which have no visible edges are skipped completely, unless the `ignore_hidden` flag is set to false (this is done for action-based snapping).
3. The standard validation done by the edge engine is modified so that any invisible edge is effectively moved out to infinity, which should remove it from consideration for snapping.

While I was cleaning up the changes and fixing my broken first attempt at making the edge engine ignore the invisibles, I also dropped the `enum view_edge` in favor of `enum wlr_edges`, which is both more relevant and used for new visibility calculations.

I've run through some tests and this seems to work as intended, although more thorough testing is appropriate.

### DIFF
--- a/src/edges.c
+++ b/src/edges.c
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include <assert.h>
 #include <limits.h>
+#include <pixman.h>
+#include <wlr/util/edges.h>
 #include <wlr/util/box.h>
 #include "common/border.h"
 #include "common/macros.h"
@@ -8,6 +10,7 @@
 #include "edges.h"
 #include "labwc.h"
 #include "view.h"
+#include "node.h"
 
 static void
 edges_for_target_geometry(struct border *edges, struct view *view,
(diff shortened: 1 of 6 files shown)


Calandracas606:
FYI void clang builds are currently broken https://github.com/void-linux/void-packages/pull/48625

src/edges.c
```diff
@@ -90,42 +112,62 @@ validate_single_region_edge(int *valid_edge,
...
+		break;
+	case WLR_EDGE_LEFT:
+		opposing = WLR_EDGE_RIGHT;
+		break;
+	case WLR_EDGE_RIGHT:
+		opposing = WLR_EDGE_LEFT;
+		break;
+	default:
+		/* Should never be reached */
+		assert(false);
```
Consolatis, line 132: I think we had issues with this before when built without asserts (and should likely add an optimized build to the CI).

AFAIR the conclusion was to either
- handle all possible cases and thus get a compile error when we missed a case (`../src/edges.c:45:2: error: enumeration value ‘WLR_EDGE_NONE’ not handled in switch [-Werror=switch]`)
- or keeping the `default` and `assert()` but also `break` or `return` (or even `abort()`) with or without additional logging depending on the case

I am not sure if we could catch this via checkpatch.pl for the future.

There seem to be another `default: assert(false)` in edges.c.

src/edges.c
```diff
@@ -182,27 +222,161 @@ validate_output_edges(struct border *valid_edges,
...
+	switch (overlap) {
+	case PIXMAN_REGION_IN:
+		view->edges_visible = WLR_EDGE_TOP | WLR_EDGE_RIGHT
+			| WLR_EDGE_BOTTOM | WLR_EDGE_LEFT;
+		break;
+	case PIXMAN_REGION_OUT:
+		view->edges_visible = 0;
+		return;
+	case PIXMAN_REGION_PART:
+		{
```
Consolatis, line 266: In other parts of the code we do also have
`; /* works around "a label can only be part of a statement" */`

as first line after the `case`, personally I'd prefer that variant in this case to get less indentation.

src/edges.c
ahesford: Good point. In this case, a return is OK because it will effectively skip validation of an ill-defined edge. In `build_edge`, getting an edge wrong is a completely unacceptable case, so I've added an `abort()` with a descriptive message. We only ever call `build_edge` with hard-coded arguments (and only ever should), so hard failing in all cases seems acceptable.

src/edges.c
ahesford: Done

src/edges.c
Consolatis: > and should likely add an optimized build to the CI to detect this

done.

johanmalm:
LGTM. Have tested.

PR #1520, opened 2024-02-11
CI: update Void-musl base image

ahesford:
Void's container naming [changed in Fall 2023](https://voidlinux.org/news/2023/08/container-updates.html).


Consolatis:
:)

PR #1523, opened 2024-02-12
Add fallback output

spl237:
This PR adds the ability to automatically enable a fallback output when there are no physical displays connected, to address the issue at https://github.com/any1/wayvnc/issues/286

The environment variable LABWC_FALLBACK_OUTPUT needs to be set to enable the new behaviour - if that is not set, no fallback output is created.

If that variable is set, then a default output called HEADLESS is created whenever there are no physical displays connected, and is destroyed as soon as a physical display is connected. This means that it is always possible to create a VNC connection to the compositor.

I am not sure this is the correct way to enable this behaviour, but it seems to work reliably in my testing, and is similar to the approach taken on wayfire to achieve the same functionality.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -141,6 +141,10 @@ output_destroy_notify(struct wl_listener *listener, void *data)
 	 * destroyed automatically when the wlr_output is destroyed
 	 */
 	free(output);
+
+	if (server->fallback_output && wl_list_length(&server->outputs) == 0) {
+		output_add_virtual(server, "HEADLESS");
+	}
 }
 
 static void
@@ -189,6 +193,16 @@ add_output_to_layout(struct server *server, struct output *output)
 		wlr_scene_output_layout_add_output(server->scene_layout,
 			layout_output, output->scene_output);
 	}
+}
+
+static int
+handle_virtual_output_destroy_timeout(void *data)
+{
+	struct server *server = data;
+	output_remove_virtual(server, "HEADLESS");
+	wl_event_source_remove(server->destroy_timeout);
+	server->destroy_timeout = NULL;
+	return 0;
 }
 
 static void
@@ -354,6 +368,19 @@ new_output_notify(struct wl_listener *listener, void *data)
 	server->pending_output_layout_change--;
 	do_output_layout_change(server);
 	seat_output_layout_changed(&output->server->seat);
+
+	if (server->fallback_output && wl_list_length(&server->outputs) > 1) {
+		/* libwayland has a bug when a global is created and
+		 * immediately destroyed, as clients don't have enough time
+		 * to bind it, so the virtual output is destroyed on a timer.
+		 */
+		if (!server->destroy_timeout) {
+			server->destroy_timeout =
+				wl_event_loop_add_timer(server->wl_event_loop,
+					handle_virtual_output_destroy_timeout, server);
+		}
+		wl_event_source_timer_update(server->destroy_timeout, 1000);
+	}
 }
 
 void
@@ -388,6 +415,14 @@ output_init(struct server *server)
 	wl_list_init(&server->outputs);
 
 	output_manager_init(server);
+
+	if (getenv("LABWC_FALLBACK_OUTPUT")) {
+		output_add_virtual(server, "HEADLESS");
+		server->destroy_timeout = NULL;
+		server->fallback_output = true;
+	} else {
+		server->fallback_output = false;
+	}
 }
 
 static void
--- a/src/output.c
+++ b/src/output.c
@@ -143,7 +143,7 @@ output_destroy_notify(struct wl_listener *listener, void *data)
 	free(output);
 
 	if (server->fallback_output && wl_list_length(&server->outputs) == 0) {
-		output_add_virtual(server, "HEADLESS");
+		output_add_virtual(server, "NOOP-1");
 	}
 }
 
@@ -199,7 +199,7 @@ static int
 handle_virtual_output_destroy_timeout(void *data)
 {
 	struct server *server = data;
-	output_remove_virtual(server, "HEADLESS");
+	output_remove_virtual(server, "NOOP-1");
 	wl_event_source_remove(server->destroy_timeout);
 	server->destroy_timeout = NULL;
 	return 0;
@@ -417,7 +417,7 @@ output_init(struct server *server)
 	output_manager_init(server);
 
 	if (getenv("LABWC_FALLBACK_OUTPUT")) {
-		output_add_virtual(server, "HEADLESS");
+		output_add_virtual(server, "NOOP-1");
 		server->destroy_timeout = NULL;
 		server->fallback_output = true;
 	} else {
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -337,6 +337,9 @@ struct server {
 
 	struct menu *menu_current;
 	struct wl_list menus;
+
+	struct wl_event_source *destroy_timeout;
+	bool fallback_output;
 };
 
 #define LAB_NR_LAYERS (4)


src/output.c
```diff
@@ -354,6 +368,19 @@ new_output_notify(struct wl_listener *listener, void *data)
 	server->pending_output_layout_change--;
 	do_output_layout_change(server);
 	seat_output_layout_changed(&output->server->seat);
+
+	if (server->fallback_output && wl_list_length(&server->outputs) > 1) {
+		/* libwayland has a bug when a global is created and
+		 * immediately destroyed, as clients don't have enough time
+		 * to bind it, so the virtual output is destroyed on a timer.
+		 */
```
Consolatis, line 376: This seems like a pretty hacky workaround and may or may not work depending on the clients responding within 1 sec..

Is there a issue for that at wlroots?

src/output.c
```diff
@@ -388,6 +415,14 @@ output_init(struct server *server)
 	wl_list_init(&server->outputs);
 
 	output_manager_init(server);
+
+	if (getenv("LABWC_FALLBACK_OUTPUT")) {
+		output_add_virtual(server, "HEADLESS");
```
Consolatis, line 420: The name prevents wayvnc from automatically resizing the output:
https://github.com/any1/wayvnc/blob/master/src/output.c#L222

src/output.c
```diff
@@ -141,6 +141,10 @@ output_destroy_notify(struct wl_listener *listener, void *data)
 	 * destroyed automatically when the wlr_output is destroyed
 	 */
 	free(output);
+
+	if (server->fallback_output && wl_list_length(&server->outputs) == 0) {
+		output_add_virtual(server, "HEADLESS");
+	}
```
Consolatis, line 147: `s/wl_list_length/wl_list_empty/`
However, this might actually fail because there may be outputs that are disabled or used only for DRM leasing.

We likely need something like this instead:
```c
bool output_available(struct server *server)
{
		if (output_is_usable(output)) {
			return true;
		}
	}
	return false;
}
```

Consolatis:
In general I like the idea with the environment variable.
Is there more detail on the wlroots bug described in the comment somewhere?

src/output.c
spl237: I don't know - I am basing it on the code at https://github.com/WayfireWM/wayfire/blob/master/src/core/output-layout.cpp#L1520, and certainly destroying the virtual output immediately caused segfaults under certain circumstances; specifically when starting up with a display already connected.

spl237:
As comment above - I am basing it on the code at https://github.com/WayfireWM/wayfire/blob/master/src/core/output-layout.cpp#L1520

Without that code, you get a segfault if you start the compositor with a display already connected, as the virtual output seems to be destroyed twice, and on the second occasion, the data in it is invalid.

src/output.c
spl237: Happy to change the name - any alternative suggestions?

spl237: On Wayfire, the fallback output is called NOOP-1, (for "no output") - would that name be ok?

Consolatis: Anything starting with `HEADLESS-` or `NOOP-` would be fine for wayvnc.

src/output.c
Consolatis: I played around a bit with an implementation without timers so we only store the `wlr_output` of the fallback output and decide based on that pointer. That mostly works but is a bit annoying as well because wlroots will emit the new output signal before returning from `wlr_headless_add_output()`.

However, I am not sure if adding and removing the fallback output is the best option here. Maybe we could instead always add the fallback output (before the headless backend has been started to not run into the issue described above) and then keep the fallback output around but disable as soon as some other output is detected + enable when the last output is gone.

spl237: I did try disabling and re-enabling instead of destroying and creating, but it means you always see the headless output in tools like wlr-randr, which is potentially confusing to the user. I have modified arandr to provide a graphical front end to wlr-randr as well as xrandr, and just disabling the headless output would mean it would appear in arandr and the user could then enable and resize it when it wasn't actually valid.

spl237: If you can, that would be great - I spent a couple of days trying everything I could think of to get rid of it, but I suspect you'll have more sensible ideas than I did!

Consolatis: Feel free to give https://github.com/Consolatis/labwc/tree/feature/virtual_fallback_output a try.

I could only test it nested which sadly doesn't allow outputs to re-appear after being destroyed. The name of the fallback output is read from the environment, e.g. `LABWC_FALLBACK_OUTPUT=HEADLESS-fallback labwc`.

spl237: Sorry, that doesn't start the desktop for me if I boot with an external display connected; it just falls back to the greeter every time. I can see the headless output if I boot with no external display connected, and if I connect an external display once booted to headless, that works, but initial boot with external display connected fails. 

BTW, that was exactly the problem I had before adding the timeout on the destroy - there seems to be some sort of startup race condition between the initial creation of the fallback output and its being destroyed when a real output is detected, and I suspect you are seeing the same problem. I was getting two separate destroy notifications, and the second one had bad data resulting in a segfault.

Also - it doesn't seem to make any difference whether I set the environment variable, or what I set it to - I always get a fallback output called "1" whether the variable is set or not.

Consolatis: > Also - it doesn't seem to make any difference whether I set the environment variable, or what I set it to - I always get a fallback output called "1" whether the variable is set or not.

As the name comes from the `LABWC_FALLBACK_OUTPUT` env var it seems be set somewhere (maybe in `~/.config/labwc/environment`?). You could try to `env | grep -i labwc` to verify if its set or not.

> Sorry, that doesn't start the desktop for me if I boot with an external display connected; it just falls back to the greeter every time. I can see the headless output if I boot with no external display connected, and if I connect an external display once booted to headless, that works, but initial boot with external display connected fails.

I'll try to test on another system with DRM output later.

> BTW, that was exactly the problem I had before adding the timeout on the destroy - there seems to be some sort of startup race condition between the initial creation of the fallback output and its being destroyed when a real output is detected, and I suspect you are seeing the same problem. I was getting two separate destroy notifications, and the second one had bad data resulting in a segfault.

At least when running nested this seems to work fine for me, however it might be that the wayland backend is initialized first and then the headless backend while what you are experiencing is headless being initialized first and then the DRM one.

I decoupled the creation / destruction of the fallback output from the `new_output_notify/destroy` handler and moved it into the output layout change event, that way we should not execute anything twice (and also catch the case with disabled or DRM leased outputs).

Consolatis: Had a chance to test on a DRM setup and could a) reproduce the crash and b) hopefully fix it.

The reason for the crash was that the output was destroyed while in the new_output handler of itself, e.g. the original issue.

That in turn happened because we unconditionally added the fallback output when starting up, in the DRM case this means that the actual DRM outputs may be discovered earlier than the fallback output when starting up the backends. Thus we ended up in a situation where we already added the fallback output but get the notification only when the backends start up at which point there are already the DRM outputs present. Which in turn means that we try to destroy it while its being added and wlroots really doesn't like that.

The way I tackled it is that we now a) move the creation of the fallback output \*after\* the backends have already been started and b) guard it with a check that there are indeed no other outputs available at that point.

The branch test branch is updated and could need some more testing.

spl237: Yes, that was what I saw - the problem arose when I created the headless output and then it was immediately destroyed when the real output was detected; I couldn't work out how to conditionally create it only after waiting to see if a real output appeared. So I think that will indeed fix the crash - I'll test some more in the morning.

spl237: OK, just tried this again and I can't repeat that problem - even without LABWC_FALLBACK_OUTPUT being set, I am still able to connect to the physical display. Not sure what was going wrong with this last week, but it seems to be ok now - must have been finger trouble my end. Apologies for the spurious problem report!

spl237: Is it possible to get your branch accepted as a PR into labwc? It seems to work pretty well to me.

Consolatis: Sure, its too risky to merge before 0.7.1 though, once the release is done I'll open a PR.

Consolatis: Created the PR:

- #1586

Refactored it slightly, testing would be appreciated.

Consolatis:
Replaced by
- #1586

Issue #1320, opened 2023-12-13
Chromium always briefly opens unmaximized first even with `<windowRule identifier="*"><action name="Maximize"/></windowRule>`


sgtpep:
Discovered some inconsistencies in how some complex apps (like browsers) react to `<windowRule identifier="*"><action name="Maximize"/></windowRule>`:

- Scenario 1. I don't have `<windowRule identifier="*"><action name="Maximize"/></windowRule>` in `rc.xml`, I open previously unmaximized Chromium, manually maximize it, then reopen it. Chromium always first briefly displays its window unmaximized then maximizes it after some short delay.
- Scenario 2. I have `<windowRule identifier="*"><action name="Maximize"/></windowRule>` in `rc.xml`, I open previously unmaximized Chromium. Chromium always first briefly displays its window unmaximized then `<action name="Maximize"/>` takes effect and Chromium gets maximized after some short delay.
- Scenario 3. I have `<windowRule identifier="*"><action name="Maximize"/></windowRule>` in `rc.xml`, I open previously unmaximized Firefox. Firefox opens maximized and then then breifly switches to the previous unmaximized state.

Expectation: Ideally, all apps should open maximized straight away without flickering with `<action name="Maximize"/>`.

Fwiw, I was a heavy Openbox user in the past with Xorg and actively used window rules, and they worked flawlessly with all the apps and windows.

Consolatis:
It might make a difference if Chromium / Firefox use the X11 or the wayland native backend. Does it happen in all cases?
(`MOZ_ENABLE_WAYLAND=1` for Firefox / `--enable-features=UseOzonePlatform --ozone-platform=wayland` for Chromium)

sgtpep:
- Ironically, Chromium opens correctly maximized with no flickering with `--ozone-platform=x11` under `labwc` but unmaximized-then-maximized with `--enable-features=UseOzonePlatform --ozone-platform=wayland`.
- I couldn't find a way to test Firefox with wayland support disabled: there's a way to enable it with `MOZ_ENABLE_WAYLAND=1` but apparently no clear way to disable it back, and for some reason I have wayland enabled in Firefox by default judging by `about:support`.

sgtpep:
Another scenario: no window rules in `rc.xml`, Chromium was previously unmaximized. Running `chromium --start-maximized` also produces flickering before actually maximizing.

jlindgren90:
[...]

PR #1529, opened 2024-02-13
xwayland: set initial geometry in map_request handler

jlindgren90:
Set the initial geometry of maximized/fullscreen views before actually mapping them, so that they can do their initial layout and drawing with the correct geometry. This avoids visual glitches and also avoids undesired layout changes with some apps (e.g. HomeBank).
    
Depends on: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4546
    
Fixes:
- #1320

### DIFF
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -69,6 +69,7 @@ struct xwayland_view {
 	struct wl_listener set_override_redirect;
 	struct wl_listener set_strut_partial;
 	struct wl_listener set_window_type;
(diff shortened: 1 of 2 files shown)


jlindgren90:
wlroots MR was merged. `about_to_map` was renamed to `map_request`.

jlindgren90:
This should be good to go now that we are on wlroots 0.18.

johanmalm:
Thanks

Issue #1341, opened 2023-12-19
Regression: Xwayland application not showing up on task switcher


BGazotti:
Upgraded from 0.6.3 to 0.6.6 just now, and Wine on XWayland's window does not show up on the task switcher anymore; I can focus on it by clicking or using the taskbar (sfwbar). Additionally, it also now renders _under_ sfwbar while in full screen, essentially having the top cropped off. Does that warrant a separate issue?

labwc 0.6.6
wlroots 0.16.2
wayland 1.22.0

Thank you!

Consolatis:
> Wine on XWayland's window does not show up on the task switcher anymore;

Do you have a window rule with `skipWindowSwitcher` set by chance?
Otherwise wine windows \*should\* show up on the `A-Tab` dialog window switcher the same as all other X11 windows.

Or maybe the wine window somehow isn't focusable, e.g. our `view_wants_focus()` handler for xwayland doesn't like wine for some reason. CC @jlindgren90 

>  Additionally, it also now renders under sfwbar while in full screen, essentially having the top cropped off.

If sfwbar is configured to use anything other than the "overlay" layer the fullscreen window should indeed be on top of the panel. Does it work correctly with another fullscreen application (like some video / youtube in fullscreen)?

BGazotti:
> Do you have any window rule with `skipWindowSwitcher` set by chance?

I do not, just a rule for steam to have no window decorations. Worth noting is that ToggleAlwaysOnTop makes it render on top of all other windows but not sfwbar.

> Does it work correctly with another fullscreen application (like some video / youtube in fullscreen)?

Firefox in fullscreen works as expected and renders on top of the panel.  sfwbar is configured to use the "top" layer. Is that correct?

Also, just tested in 0.6.5, wine window displays in fullscreen AND shows up under alt+tab. There's some weirdness going on when trying to send it to another workspace, though - that causes sfwbar to vanish.

jlindgren90:
The window switcher issue would probably be solved by addressing #1139.

Consolatis:
Just saw that our focus detection logic for xwayland applications may also depend on the window decorations. I am not sure what `WM_CLASS` is used for your wine case but maybe it is matching your window rule; could you test if removing the window rule makes the application show up in `A-Tab`?

> Firefox in fullscreen works as expected and renders on top of the panel.

Hm.. so we somehow don't get the memo that the wine window is in fullscreen.

> sfwbar is configured to use the "top" layer. Is that correct?

[...]

Issue #1139, opened 2023-10-04
xwayland: support _NET_WM_SKIP_PAGER and ignore WM_HINTS.input for Alt-Tab


jlindgren90:
Related to issue #1130 and discussion on #1137.

Since 7e72bf975fb65c8290b398d21b2ad9d87a22880f, XWayland views that set `WM_HINTS.input=false` are skipped in the Alt-Tab pager. This behavior matches Openbox; however, it is inconvenient when dealing with some XWayland applications that set `WM_HINTS.input=false` incorrectly (example: MATLAB).

(`WM_HINTS.input=false` means that a client either "never expects keyboard input" (No Input model) or "wants to prevent the window manager from setting the input focus to any of its windows" (Globally Active input model). See ICCCM section 4.1.7.)

In the (relatively) newer Extended Window Manager Hints spec, there is a `_NET_WM_STATE_SKIP_PAGER` hint which is more narrowly tailored to exclude a view from the Alt-Tab list. If we support this hint, we could probably ignore `WM_HINTS.input` for the purpose of Alt-Tab.

In my opinion, we should still not focus views that set `WM_HINTS.input=false` automatically (e.g. when another view that was on top is closed/minimized). In my workflow, it trips me up when I close a window and the taskbar gains focus instead of the window underneath. It takes a second to figure out "where did the keyboard focus go?"

Consolatis:
> I close a window and the taskbar gains focus instead of the window underneath. It takes a second to figure out "where did the keyboard focus go?"

The taskbar you are using doesn't support the layershell protocol? That one has a `keyboard_interactivity = None` setting which prevents that.

jlindgren90:
> The taskbar you are using doesn't support the layershell protocol? That one has a `keyboard_interactivity = None` setting which prevents that.

Right, I am still using qmpanel under XWayland since I have not yet found a feature-compatible replacement that user layershell.

More generally, several of the applications I use regularly (particular Qt-based ones) still have issues with native Wayland and so I am focusing on getting things usable in XWayland mode for now, as a transitional step. At this point, it's mostly minor issues that still prevent labwc from being a drop-in replacement for Openbox+Picom, while continuing to run an X11-based desktop otherwise. I think it's nice to have this as an option, rather than doing an all-at-once X11 -> native Wayland switchover.

johanmalm:
We could consider doing some window-rule for xwayland-window-types in line with `sway`s approach

[...]

PR #1533, opened 2024-02-15
xwayland: use _NET_WM_WINDOW_TYPE to fix focus issues

jlindgren90:
Now that we have _NET_WM_WINDOW_TYPE support, we can use it to fix focus issues with Globally Active windows more properly, as we've discussed before (https://github.com/labwc/labwc/issues/1139#issuecomment-1763191003), and remove various hacks/workarounds that are no longer needed.

The first commit is the important one. The others aren't urgent but are nice cleanups, and will be more relevant when the wlroots support is in place to merge #1142. Details in the commit messages.

Fixes:
- #1139
- #1341

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -108,8 +108,6 @@ struct view_impl {
 	void (*move_to_back)(struct view *view);
 	struct view *(*get_root)(struct view *self);
 	void (*append_children)(struct view *self, struct wl_array *children);
-	/* determines if view and surface are owned by the same process */
-	bool (*is_related)(struct view *self, struct wlr_surface *surface);
 	struct view_size_hints (*get_size_hints)(struct view *self);
 	/* if not implemented, VIEW_WANTS_FOCUS_ALWAYS is assumed */
 	enum view_wants_focus (*wants_focus)(struct view *self);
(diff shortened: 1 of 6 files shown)


jlindgren90:
Any and all testing appreciated, of course :)

jlindgren90:
For 0.7.1, maybe it makes sense to take only the first commit ("xwayland: treat Globally Active windows according to type") and leave the other for later...

johanmalm:
Thanks for doing this.
I’m a bit nervous about changing any focus logic before 0.7.1
If you agree, I’d prefer to leave the whole patch set until the next merge window to make sure it gets wider test coverage.

jlindgren90:
Sure, that's fine with me.

Consolatis:
Yes, agree. Lets add this after 0.7.1.

The code itself looks good to me though.
We should keep an eye on this PR in combination with
- #1479

Using `view_is_focusable` as a filter for the `A-Tab` window list might turn out to be not that maintainable in the long term.

droc12345:
> Using `view_is_focusable` as a filter for the `A-Tab` window list might turn out to be not that maintainable in the long term.

This shouldn't be a problem with the changes to desktop_cycle_view, as far as window switching.
It's used in one place in osd.c (for window switcher outline) but it shouldn't be a problem.

johanmalm:
Code looks good.
Thanks a bunch for working on this. It feels like we're getting ever closer to getting surface focus right.
I'm always a tad nervous about changing surface focus behavior, but it's at the right point in the release cycle.
Stand-by for Java and Steam niggles.

PR #1543, opened 2024-02-18
session: run shutdown script, clean up activation env before exit

ahesford:
This came up on IRC and seems like it would be a useful addition.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -8,7 +8,7 @@ labwc - configuration files
 
 Labwc uses openbox-3.6 specification for configuration and theming, but does not
 support all options. The following files form the basis of the labwc
-configuration: rc.xml, menu.xml, autostart and environment.
+configuration: rc.xml, menu.xml, autostart, shutdown and environment.
 
 No configuration files are needed to start and run labwc.
 
@@ -34,11 +34,8 @@ alternative.
 The configuration directory location can be override with the -C command line
 option.
 
-All configuration and theme files except autostart are re-loaded on receiving
-signal SIGHUP.
-
-The *autostart* file is executed as a shell script. This is the place for
-executing clients for handling background images, panels and similar.
+All configuration and theme files except autostart and shutdown are re-loaded on
+receiving signal SIGHUP.
 
 The *environment* file is parsed as *variable=value* and sets environment
 variables accordingly. It is recommended to specify keyboard layout settings and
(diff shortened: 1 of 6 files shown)


Consolatis:
Code LGTM, we should likely document it though. Also up to @johanmalm if we want to merge this before or after release.

ahesford:
No rush to merge before release---I can go either way. I'll write up some documentation to go along with the change.

johanmalm:
I like this in principle, but prefer to leave until after release.

`update_activation_env()` contains calls to both `dbus-update-activation-environment` and `systemctl` so it feels like we ought to call both on "unactivation" also.

https://github.com/labwc/labwc/blob/17bad48d87f5550dcdd16a03fd95b9bd9ebc73b3/src/config/session.c#L82

docs/labwc-config.5.scd
```diff
@@ -48,6 +45,20 @@ sourced prior to running openbox.
...
+support) DISPLAY will be defined. This is the place for executing clients for
+handling background images, panels and other tasks that should run automatically
+when labwc launches.
+
+The *shutdown* file is executed as a shell script when labwc is preparing to
+terminate itself. All environment variables, including WAYLAND_DISPLAY and
+DISPLAY, will be available to the script. However, because the script runs
+asynchronously with other termination tasks, the shutdown file should not assume
+that the display will be usable. This is a perform any custom operations
+necessary to finalize a labwc session.
```
Consolatis, line 60: There seems to be missing something

ahesford:
I split the dbus and systemd activation stuff into separate functions given the different set/unset behaviors for each. Suggestions welcome for improving this.

Consolatis:
> I split the dbus and systemd activation stuff into separate functions given the different set/unset behaviors for each. Suggestions welcome for improving this.

Looks good to me.

I am not sure if `systemctl` might end up spawning a new dbus session as well if there was none. Before this PR it wasn't called if `DBUS_SESSION_BUS_ADDRESS` wasn't set while now it is. But again, not sure if that is an actual problem.

ahesford:
Taking a different tack here:
- Restore the dependence of systemctl activation on a valid dbus session-bus address
- With https://github.com/labwc/labwc/pull/1546 pending, add some dynamic behavior that makes it easier to do the right thing if we add or remove variables in the future.

I probably don't need so many length checks, but we're concatenating strings so I took extra precautions... Comments welcome.

johanmalm:
Could we simplify str_concat by using the common/buf.h helpers?

https://github.com/labwc/labwc/blob/1aa906e0de740a6af96c7365e03c137f84bbc414/include/common/buf.h#L46

Or even add a create a buf_str_join(buf, string-array, delim) where the delim could be “ “ and “= “ for the two cases.

ahesford:
I generalized the join and moved it alongside the other string helpers. Trying to use the buffer struct for this might be a little awkward, and seems like it would involve several reallocs.

ahesford:
Now that 0.7.1 is out, I think this is ready for review.

johanmalm (commented):
Two minor comments. Other than those LGTM.
Nice improvement.
Great effort with `str_join()`

ahesford:
Changes adopted

Issue #1363, opened 2023-12-23
Add a short description / more details for the "Obligatory screenshot" in README.md (on the "GitHub front page")?


rxhfcy:
The "obligatory screenshot" on the [project's "GitHub front page"](https://github.com/labwc/labwc#16-screenshot) kinda looks nice, or at least it looks like a much better starting point for labwc newbies than the current default config (empty black screen + nothing works out of the box, #865).

### SUGGESTION: ###

The screen shot is probably intended for newbies who don't know how to configure labwc yet.

Please consider adding a short helpful description above the obligatory screenshot with the following information:
- which panel was used
- which theme
- any other modifications, etc
- (maybe add a direct link to example "dotfiles" to make things look like what's shown in the screenshot)
- (and/or at least a link to "Integration" to make it more likely that the newbie user eventually learns how to make labwc useful and actually becomes a labwc user instead of leaving the page)

Screenshot from [README.md](https://github.com/labwc/labwc#16-screenshot):
(e.g. "The obligatory screenshot: uses X panel with Y theme (link to example dotfiles), see "Integration" below for more costomization hints" etc):

![Obligatory_screenshot](https://github.com/labwc/labwc/assets/42669366/329b3f89-3331-4ba9-b7ab-be9b7efcd96a)

ludg1e:
Maybe something like this?
![image](https://github.com/labwc/labwc/assets/77110723/31697ee2-880a-4cfa-b2f5-0f1c3eda889d)

Consolatis:
Already mentioned in
- #1363

but to reduce confusion:
That panel in the screenshot is an unreleased one named "Bornel". However we should likely not refer to it as its not publicly available and I don't feel like maintaining it currently. Maybe we should look into creating a new screenshot with a panel like sfwbar or xfce4-panel instead (and maybe the labwc logo, centered on a black background set via `swaybg`?).

johanmalm:
Appreciate comments on https://github.com/labwc/labwc.github.io/pull/43

PR #1547, opened 2024-02-19
README.md: Add screenshot description

johanmalm:
...and use the png files on the labwc website rather than imgur.

Fixes: #1363


Consolatis:
:)

Issue #783, opened 2023-02-16
IME support (text_input_v3 protocol  & zwp_input_method_keyboard_grab_v2 ?)


rico256-cn:
https://bugs.chromium.org/p/chromium/issues/detail?id=1039161

google chrome version 112 supports the text_input_v3  protocol.   

Does labwc support text_input_v3 protocol?

Consolatis:
No, neither v2 nor v3 are implemented by labwc.
Marking this as a feature request.

[This](https://github.com/swaywm/sway/pull/4740/files) is the old implementation for sway for future reference.

rico256-cn:
LabWC runs very smoothly. I want to use LabWC to build a minimal wayland desktop environment. chrome needs this feature to use the input method. Hope to see this feature soon.

Consolatis:
I completely agree that we should implement it. Not everybody uses the latin alphabet.

We do support the [virtual-keyboard-unstable-v1](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/protocol/virtual-keyboard-unstable-v1.xml) protocol though,
maybe there is some on-screen-keyboard which could make use of it?

Small overview for future reference : https://dcz_self.gitlab.io/posts/input_method/

rico256-cn:
On-screen-keyboard is not suitable for production use and is inefficient. input-method-v 3 is recommended as a priority。

rico256-cn:
Thank you, I will wait patiently for the implementation of this feature, hope labwc better and better

rico256-cn:
Please refer:
https://github.com/fcitx/fcitx5/issues/206

johanmalm:
Is it enough to just call wlr_text_input_manager_v3_create() from server_init()?

rico256-cn:
FYI:


https://github.com/WayfireWM/wayfire/issues/1160

kmephistoh:
Application/Compositor uses text-input-v3  --> [link](https://fcitx-im.org/wiki/Using_Fcitx_5_on_Wayland#Sway)
Comopositor/Application uses zwp_input_method_v2 (I don't know which version is the right choice,  How sway do it ?) --> May useful [link](https://wayland.app/protocols/text-input-unstable-v3)

tokyo4j:
This is my work-in-progress implementation: [link](https://github.com/tokyo4j/labwc/tree/input-method)

johanmalm:
[...]

PR #1548, opened 2024-02-21
IME: Support input method (without IME popups)

tokyo4j:
This PR adds support for text-input-unstable-v3 protocol and input-method-unstable-v2 protocol.
Addresses #783

This is based on [Sway's implementation](https://github.com/swaywm/sway/blob/master/sway/input/text_input.c), but I made many changes because it had some designs that don't make sense to me, like follows:
-  It uses `sway_text_input.pending_focused_surface` which is used solely for keeping track of focused surface when input-method isn't present and complicates the code.
- When the focused surface is updated, the relay always sends `deactivate` event to input-method for every text-input to send `leave` event.
- When received `commit`/`enable` requests from application, the relay always sends some events to input-method although the focus might be already moved to another application.
- When text-input is destroyed, input-method is always deactivated although another text-input might be already enabled (this causes a bug that opening two Alacritty windows and closing one of them leaves the input-method deactivated although the focus is moved back to the first window).

To address these problems while keeping the code simple, this PR adds following fields to `input_method_relay`:
- `focused_surface`: The surface the seat is currently focusing
- `active_text_input`: The text-input currently communicating with input-method

and adds following functions:
- `update_active_text_input()`
  - Updates `active_text_input` depending on the focused surface and enabled/disabled state of text-inputs.
  - Sends `activate`/`deactivate` events to input-method if the value is changed
- `update_text_inputs_focused_surface()`
  - Updates focused surface of each text-input depending on `focused_surface` and whether input-method exists
  - Sends `enter`/`leave` events to text-inputs whose focused surface is changed

With these changes, I achieved:
- Unify the logic to send `activate`/`deactivate` events to input-method into `update_active_text_input()`
- Unify the logic to send `enter`/`leave` events to text-inputs into `update_text_inputs_focused_surface()`
- Send `deactivate` event to input-method only when there's actually no text-inputs to communicate
- Properly select text-input to communicate with input-method and handle requests and destruction from unfocused text-inputs.

Note that this PR doesn't include IME popups, so I'm gonna make another PR for it once this PR got merged.

You can try AUR ([labwc-im-git](https://aur.archlinux.org/packages/labwc-im-git)/[labwc-im](https://aur.archlinux.org/packages/labwc-im)) or [this branch](https://github.com/tokyo4j/labwc/tree/text-input-v1) to get IME popups. They also includes minimal text-input-unstable-v1 support for Chromium and Electron-based apps (running with `--enable-wayland-ime`).

And of course, not merging this before 0.7.1 is totally fine with me.

## For testing
As far as I know, `fcitx5` is the only viable input method client that implements input-method-unstable-v2 protocol.
To use it, install `fcitx5` and language-specific modules like `fcitx5-mozc` (Japanese) or `fcitx5-rime` (Chinese), run `fcitx5`, and activate it with hotkeys (default is `Ctrl+Space`).

For application side, GTK3/4 (without settings `GTK_IM_MODULE` environment variable), Alacritty, wezterm and kitty implement text-input-unstable-v3 protocol.

### DIFF
--- a/include/input/ime.h
+++ b/include/input/ime.h
@@ -1,0 +1,77 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+
+#ifndef LABWC_IME_H
+#define LABWC_IME_H
+
+#include <wlr/types/wlr_text_input_v3.h>
+#include <wlr/types/wlr_input_method_v2.h>
+#include "labwc.h"
+
+struct keyboard;
+
+/*
+ * The relay structure manages the relationship between text-inputs and
+ * input-method on a given seat. Multiple text-inputs may be bound to a relay,
+ * but at most one will be "active" (commucating with input-method) at a time.
+ * At most one input-method may be bound to the seat. When an input-method and
+ * an active text-input is present, the relay passes messages between them.
+ */
+struct input_method_relay {
+	struct seat *seat;
+	struct wl_list text_inputs; /* struct text_input.link */
+	struct wlr_input_method_v2 *input_method;
+	struct wlr_surface *focused_surface;
+	/*
+	 * Text-input which is enabled by the client and communicating with
+	 * input-method.
+	 * This must be NULL if input-method is not present.
+	 * Its client must be the same as that of focused_surface.
+	 */
+	struct text_input *active_text_input;
+
+	struct wl_listener new_text_input;
+	struct wl_listener new_input_method;
+
+	struct wl_listener input_method_commit;
+	struct wl_listener input_method_grab_keyboard;
+	struct wl_listener input_method_destroy;
+
+	struct wl_listener keyboard_grab_destroy;
+	struct wl_listener focused_surface_destroy;
+};
+
+struct text_input {
+	struct input_method_relay *relay;
+	struct wlr_text_input_v3 *input;
+	struct wl_list link;
+
+	struct wl_listener enable;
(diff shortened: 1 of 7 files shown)


tokyo4j:
Fixed code styling and variable/function names.

johanmalm:
Thanks for working on this! I’m just focusing on the upcoming release for a few days, but will review + test soon.

Consolatis:
I had a quick read through the code and although I miss some context as for how this works together (need to read the protocols and wlroots headers) I have to say this is some really nice and clean code! Thanks for working on this indeed.

tokyo4j:
[Dorota's article](https://dorotac.eu/posts/input_method/) briefly explains how text-input, input-method and virtual keyboard are working together.

tokyo4j:
Oh, I found river is taking similar approach: [TextInput.zig](https://github.com/riverwm/river/blob/master/river/TextInput.zig) [InputRelay.zig](https://github.com/riverwm/river/blob/master/river/InputRelay.zig)

johanmalm:
This looks good. Let's go for it.
I don't seem to be able to get the environment setup to test it, but I struggle on sway too. Happy to go without that though.

johanmalm:
If you ever feel the inclination, a description of how to set this up would be a nice addition to `labwc.github.io`.

Issue #1545, opened 2024-02-19
Headerbars do not respond to touchscreen events


spl237:
Applications themselves respond to touchscreen events, but it looks as if the labwc headerbars do not have touch handling enabled.

If you run labwc on a touchscreen, then you can interact with the contents of windows, scrollbars etc. But it is not possible to activate the minimise / maximise or close buttons, or to drag the headerbar to move the window.

Has this not been implemented, or is there something I need to enable?

Consolatis:
We might not do mouse emulation for touch events, I don't have hardware to test this but `src/input/touch.c` is pretty small.
It seems to indeed completely skip non client surfaces, maybe we could emulate usual cursor events in those cases instead. Similar to `src/input/tablet.c` (e.g. via `cursor_emulate_move_absolute()` and `cursor_emulate_button()` after a timeout on `touch_down()`).

spl237:
OK, I have written a very crude bodge using cursor_emulate functions which means I can operate the close, minimize etc buttons on a headerbar with the touchscreen. I haven't yet got window dragging working, but this seems to at least be a viable solution in the short term - I'll hack about and see what I can get working.

(But if someone else knows more about this than me - which is almost certainly the case - do please feel free to come up with something better!)

spl237:
Right, I have window dragging and button clicking working, but the problem I now have is that this breaks behaviour in a few other areas of the system. I think what I need to do is to ensure that the cursor_emulate functions are only applied when a touch is on an SSD titlebar.

Is there any easy way to determine if the item under the touch point is an SSD titlebar? If I could do that, I think this would work pretty much perfectly...

spl237:
It looks as if just checking if the surface returned from touch_get_coords is valid can be used to check whether you are on the SSD or not, as because the SSD seems to be the only thing that is returned as not a valid touch surface. That assumption may of course be incorrect!

PR #1550, opened 2024-02-22
Use cursor emulate events to handle touch on headerbar

spl237:
https://github.com/labwc/labwc/issues/1545

This PR uses cursor emulate events to respond to touchscreen input on window headerbars and frames.

I have no idea if this is the correct way to do this, but it seems to work pretty well in my testing. Someone who understands more about how this area of the code works should probably review it, as there is probably a better way to do it!

### DIFF
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include <wayland-util.h>
 #include <wlr/types/wlr_touch.h>
+#include <linux/input-event-codes.h>
 #include "common/mem.h"
 #include "common/scene-helpers.h"
 #include "idle.h"
@@ -61,6 +62,8 @@ touch_motion(struct wl_listener *listener, void *data)
 
 			wlr_seat_touch_notify_motion(seat->seat, event->time_msec,
 				event->touch_id, sx, sy);
+			cursor_emulate_move_absolute(seat, &event->touch->base,
+				event->x, event->y, event->time_msec);
 			return;
 		}
 	}
@@ -104,6 +107,12 @@ touch_down(struct wl_listener *listener, void *data)
 		wlr_seat_touch_notify_down(seat->seat, surface,
 			event->time_msec, event->touch_id, sx, sy);
 	}
+	else {
+		cursor_emulate_move_absolute(seat, &event->touch->base,
+			event->x, event->y, event->time_msec);
+		cursor_emulate_button(seat, BTN_LEFT, WLR_BUTTON_PRESSED,
+			event->time_msec);
+	}
 }
 
 static void
@@ -123,6 +132,8 @@ touch_up(struct wl_listener *listener, void *data)
 	}
 
 	wlr_seat_touch_notify_up(seat->seat, event->time_msec, event->touch_id);
+	cursor_emulate_button(seat, BTN_LEFT, WLR_BUTTON_RELEASED,
+		event->time_msec);
 }
 
 void
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -106,8 +106,7 @@ touch_down(struct wl_listener *listener, void *data)
 	if (surface) {
 		wlr_seat_touch_notify_down(seat->seat, surface,
 			event->time_msec, event->touch_id, sx, sy);
-	}
-	else {
+	} else {
 		cursor_emulate_move_absolute(seat, &event->touch->base,
 			event->x, event->y, event->time_msec);
 		cursor_emulate_button(seat, BTN_LEFT, WLR_BUTTON_PRESSED,
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -13,6 +13,7 @@ struct touch_point {
 	int32_t touch_id;
 	uint32_t x_offset;
 	uint32_t y_offset;
+	struct wlr_surface *surface;
 	struct wl_list link; /* seat.touch_points */
 };
 
@@ -62,8 +63,10 @@ touch_motion(struct wl_listener *listener, void *data)
 
 			wlr_seat_touch_notify_motion(seat->seat, event->time_msec,
 				event->touch_id, sx, sy);
-			cursor_emulate_move_absolute(seat, &event->touch->base,
-				event->x, event->y, event->time_msec);
+			if (!touch_point->surface) {
+				cursor_emulate_move_absolute(seat, &event->touch->base,
+					event->x, event->y, event->time_msec);
+			}
 			return;
 		}
 	}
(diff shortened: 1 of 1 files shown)


johanmalm:
@jp7677 With you knowledge of the tablet cursor-emulation, would you have the capacity/time to review this patch?

Consolatis:
I am slightly concerned that this may break some touchscreen enabled applications as they now get the touch events but also the synthesized mouse events. Maybe we could store the surface in `struct touch_point` on `touch_down()` and then based on the pointer being `NULL` emulate either the cursor actions or send the touch events otherwise.

spl237:
> I am slightly concerned that this may break some touchscreen enabled applications as they now get the touch events but also the synthesized mouse events. Maybe we could store the surface in `struct touch_point` on `touch_down()` and then based on the pointer being `NULL` emulate either the cursor actions or send the touch events otherwise.
> 
> Edit: With this PR applied an application could also get a touch event for `down` and then a stray cursor `up` without a matching cursor `down`. That would also be solved if we only send to appropriate events based on `touch_point.surface` being set or not.

I think I am already doing that by checking to see whether surface is NULL in the touch_down handler - if I understand the code correctly, that should be checking to see if a surface is already able to accept touch events, and only applies the touch->mouse conversion if it is not. In my (limited) testing, the only surfaces which were returned as not already accepting touch events were server-side decorations.

Consolatis:
> I think I am already doing that by checking to see whether surface is NULL in the touch_down handler - if I understand the code correctly, that should be checking to see if a surface is already able to accept touch events, and only applies the touch->mouse conversion if it is not. In my (limited) testing, the only surfaces which were returned as not already accepting touch events were server-side decorations.

Yes, `touch_down` handles this but `touch_motion` and `touch_up` do not.

spl237:
Done, I think. Still seems to work fine on my touchscreen here!

Consolatis:
The cursor emulation part is looking good to me. We still do have some asymmetry here (we only relay the touch_down on `->surface` but always relay touch_up and touch_move touch events).

I am also not sure about the frame events, in the tablet case we had to synthesize them as far as I remember. Maybe @jp7677 could confirm that and what happened without them. As long as we only react to the mouse events ourselves (e.g. SSD / root window / menus) it should be fine to not have the frame events but when we deal with another client window that doesn't support the touch protocol (e.g. most likely something not GTK / Qt) it might start to break.

spl237:
> The cursor emulation part is looking good to me. We still do have some asymmetry here (we only relay the touch_down on `->surface` but always relay touch_up and touch_move touch events).

You mean the pre-existing calls to wlr_seat_touch_notify_up and wlr_seat_touch_notify_motion? It would be trivial to make those conditional on the existence of surface - happy to do that if you think it is a good idea.

spl237:
> Yes, this seems like a pre-existing issue with `touch.c`. I am actually not sure if wlroots may just turn the calls into a no-op if the currently focused surface doesn't accept touch inputs. In either case IMHO we should send all of them or none of them and it seems to be the most robust variant if we do not send any touch events unless the surface that was detected during touch_down actually accepts touch input.

Seems sensible - done.

src/input/touch.c
```diff
@@ -100,9 +106,14 @@ touch_down(struct wl_listener *listener, void *data)
 	double sx = lx - x_offset;
 	double sy = ly - y_offset;
 
-	if (surface) {
-		wlr_seat_touch_notify_down(seat->seat, surface,
+	if (touch_point->surface) {
```
jp7677, line 109: The calculation for `sx`, `sy` is only needed within this if statement, so we might move the calculation for `lx`, `ly`, `sx` and `sy` into this condition.

jp7677 (commented):
As far as i can judge, and I don't know much about the touch events, this looks solid and is certainly an improvement.  That said, cursor emulation for tablets was much easier since there it is indeed emulation only and no mixing with native tablet protocol support. About synchronization, if that question was about `wlr_seat_pointer_notify_frame`, we do that in `cursor_emulate_move_absolute`. That might actually only needed when having a surface, but not sure. It doesn't hurt tablet moves without being on a surface, so no sure if not doing this has any practical benefits without negative side effects. I would leave it as is.

Might be interesting to see how e.g. `mpv` with a wayland surface or any SDL wayland surfaces (ScumVM?) behaves with this PR. I know that on Gnome/Mutter I don't have tablet interaction on those windows. But that could be different with touch. 

Just checking, dragging windows on their SSD title bar to another place works?

I've suggested a small optimization, otherwise LGTM, though I don't have the hardware to test.

src/input/touch.c
spl237: Will do.

spl237: Done...

jp7677: Those lines can also go into the branch
```c
double lx, ly;
wlr_cursor_absolute_to_layout_coords(seat->cursor, &event->touch->base,
		event->x, event->y, &lx, &ly);
```

src/input/touch.c
```diff
@@ -47,20 +49,26 @@ touch_motion(struct wl_listener *listener, void *data)
...
 		if (touch_point->touch_id == event->touch_id) {
-			double sx = lx - touch_point->x_offset;
-			double sy = ly - touch_point->y_offset;
-
-			wlr_seat_touch_notify_motion(seat->seat, event->time_msec,
-				event->touch_id, sx, sy);
+			if (touch_point->surface) {
+				/* Convert coordinates: first [0, 1] => layout */
+				double lx, ly;
+				wlr_cursor_absolute_to_layout_coords(seat->cursor,
```
jp7677, line 59: We shouldn't move this block in that function because we are now in a loop. So keeping

```c
/* Convert coordinates: first [0, 1] => layout, then apply offsets */
double lx, ly;
wlr_cursor_absolute_to_layout_coords(seat->cursor, &event->touch->base,
        event->x, event->y, &lx, &ly);
``` 

out of it is better avoid recalculating the same value again.

(Sorry for the confusion, I really just meant the other function where I was commenting on earlier, since that is an `if else` instead of a loop like here.)

spl237: Won't that calculation only ever happen once anyway due to the way the loop is structured? The calculation is only done if the touch_id matches, and at that point the function returns.

Or am I reading it incorrectly?

jp7677: ah yeah, right, sorry, I had missed the return statement.

spl237: Thanks for checking!

spl237:
> Just checking, dragging windows on their SSD title bar to another place works?

On this point, yes, it does.

spl237:
Suggestion from https://github.com/labwc/labwc/issues/1566 added to switch focus on touch.

johanmalm:
LGTM.
Can't test, but code looks good.
Nice addition.

Issue #1525, opened 2024-02-12
Setting the scale via wlr-output-management sometimes fails


Tamaranch:
It's not always reproducible, and it only seems to concern scaling. It's easier to reproduce with current git-master than with 0.7.0, but the offending commit dates from before : 627179b277cba376cc0538128626d64ad6bccbfc

It's rather strange, since as far as I can see this commit has nothing to do with it, but it does seem to be from this commit that it's broken. I can't be completely sure since the problem isn't systematically reproducible, but it's still easy enough to reproduce that I'm pretty sure.

On the client side, I do receive the `zwlr_output_configuration_v1::succeeded` signal, but the scaling doesn't actually take place.

Consolatis:
The commit pointed to above also changes the wlroots subproject commit, if you build the subproject rather than a system wide installation that might be the culprit.

Do you see any difference in the labwc logs (e.g. `labwc -d` and redirection to a file) for when it works and when it doesn't?
We should either commit all wlr-randr changes or none of them. Another question: when the scaling fails, does wlr-randr still display the new scale when executed without arguments?

Tamaranch:
> if you build the subproject rather than a system wide installation that might be the culprit.

That's what I'm doing, otherwise I wouldn't be able to build the project in this case, I think.

> Do you see any difference in the labwc logs (e.g. labwc -d and redirection to a file) for when it works and when it doesn't?

I think the error lies there in the logs: [labwc.log](https://github.com/labwc/labwc/files/14241294/labwc.log)
```
00:00:41.335 [INFO] [../src/output.c:919] adaptive sync disabled for output LVDS-1
00:00:41.335 [INFO] [backend/drm/drm.c:782] connector LVDS-1: Modesetting with 1366x768 @ 60,001 Hz
00:00:41.338 [ERROR] [backend/drm/legacy.c:179] connector LVDS-1: drmModePageFlip failed: Périphérique ou ressource occupé
00:00:41.338 [ERROR] [../src/output.c:439] Output config commit failed
00:00:41.338 [INFO] [../src/output.c:919] adaptive sync disabled for output HDMI-A-1
00:00:41.338 [INFO] [backend/drm/drm.c:782] connector HDMI-A-1: Modesetting with 1920x1080 @ 60,000 Hz
00:00:41.338 [ERROR] [backend/drm/legacy.c:179] connector HDMI-A-1: drmModePageFlip failed: Périphérique ou ressource occupé
[...]

PR #1551, opened 2024-02-22
src/output.c: notify clients about config errors

Consolatis:
Preliminary fix for
- #1525

Based on the protocol we should also revert all previously correctly committed outputs.
#1528 is doing just that but may cause regressions so we need a short term solution and then deal with potential issues in #1528 after the release.

@Tamaranch
Testing would be greatly appreciated so we can include this preliminary fix for the next release.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -405,10 +405,11 @@ output_update_for_layout_change(struct server *server)
 	cursor_update_image(&server->seat);
 }
 
-static void
+static bool
 output_config_apply(struct server *server,
 		struct wlr_output_configuration_v1 *config)
 {
+	bool success = true;
 	server->pending_output_layout_change++;
 
 	struct wlr_output_configuration_head_v1 *head;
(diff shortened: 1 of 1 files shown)


johanmalm:
I know it's just at draft, but LGTM.

Related? https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/1873#note_2292346

Consolatis:
> Related? https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/1873#note_2292346

Yep, I asked in the wlroots channel a while ago and the response was that wlroots needs a atomic multi output commit to handle this properly. So we may be able to use that around the end of the year. Until then I guess we'll have to tackle it on our own.

Tamaranch:
This one works for me too, I'm notified if the configuration application fails :+1:

Consolatis:
> This one works for me too, I'm notified if the configuration application fails 👍

Sweet, thanks for testing. The complete solution will then come after the release.

PR #1555, opened 2024-02-24
config: Remove invalid click method

jp7677:
As far as I understood, there is no "none" click method, it is either software button area or clickfinger. See https://wayland.freedesktop.org/libinput/doc/latest/clickpad-softbuttons.html#clickpad-software-button-behavior

Also our bit logic at https://github.com/labwc/labwc/blob/master/src/seat.c#L193 doesn't support setting anything from e.g. 3 (`LIBINPUT_CONFIG_CLICK_METHOD_BUTTON_AREAS`) to 0 (`LIBINPUT_CONFIG_CLICK_METHOD_NONE`), because `x & 0 == 0`.

I figured this while trying to disable my dual sense touch pad. 
(The actual solution was https://github.com/labwc/labwc/pull/961)

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -680,7 +680,6 @@ extending outward from the snapped edge.
 	- *clickfinger* - Clicking with one, two or three finger(s) will produce
 	  left, right or middle button event without regard to the location of a
 	  click.
-	- *none* - Physical clicks will not produce button events.
 
 	The default method depends on the touchpad hardware.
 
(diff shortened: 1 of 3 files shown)


jp7677:
Cc @datMaffin because may be I misunderstood the documentation. As a consequence that would mean that the bit logic when determining if a click method needs to be set, is incomplete.

jp7677:
ah, looking at https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#ga3e5dc19a83752430d5e8be7fb2020663 I understood this wrongly, will rework ...

jp7677:
Sorry for the noice, I mixed up `libinput_device_config_click_get_methods` with `libinput_device_config_click_get_method`. All is good!

PR #1556, opened 2024-02-24
seat: fix configure condition for click method

jp7677:
Exclude none (zero) from the bitmask test , otherwise the bitmask test is always true when click method is configured to 'none' and as a result the configuration will be skipped.

Follow up from https://github.com/labwc/labwc/pull/1555 (I needed some time to wrap my head around this ;))

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -189,8 +189,10 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		wlr_log(WLR_INFO, "dwt configured");
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
-	if ((libinput_device_config_click_get_methods(libinput_dev)
-				& dc->click_method) == 0
+
+	if ((dc->click_method != LIBINPUT_CONFIG_CLICK_METHOD_NONE
+			&& (libinput_device_config_click_get_methods(libinput_dev)
+				& dc->click_method) == 0)
 			|| dc->click_method < 0) {
 		wlr_log(WLR_INFO, "click method not configured");
 	} else {


Consolatis:
This is some really strange API:
```c

/**
 * @ingroup config
 *
 * The click method defines when to generate software-emulated
 * buttons, usually on a device that does not have a specific physical
 * button available.
 */
enum libinput_config_click_method {
	/**
	 * Do not send software-emulated button events. This has no effect
	 * on events generated by physical buttons.
	 */
	LIBINPUT_CONFIG_CLICK_METHOD_NONE = 0,
	/**
	 * Use software-button areas to generate button events.
	 */
	LIBINPUT_CONFIG_CLICK_METHOD_BUTTON_AREAS = (1 << 0),
	/**
	 * The number of fingers decides which button press to generate.
	 */
	LIBINPUT_CONFIG_CLICK_METHOD_CLICKFINGER = (1 << 1),
};

/**
 * @ingroup config
 *
 * Check which button click methods a device supports. The button click
 * method defines when to generate software-emulated buttons, usually on a
 * device that does not have a specific physical button available.
 *
 * @param device The device to configure
 *
 * @return A bitmask of possible methods.
 *
 * @see libinput_device_config_click_get_methods
 * @see libinput_device_config_click_set_method
 * @see libinput_device_config_click_get_method
 */
uint32_t
libinput_device_config_click_get_methods(struct libinput_device *device);
```

So it returns a "bitmask" of all supported methods with one of the members being `0` ? Consider me confused.

jp7677:
Yeah, it took me some time to get it ;)

src/seat.c
```diff
@@ -189,8 +189,10 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 		wlr_log(WLR_INFO, "dwt configured");
 		libinput_device_config_dwt_set_enabled(libinput_dev, dc->dwt);
 	}
-	if ((libinput_device_config_click_get_methods(libinput_dev)
-				& dc->click_method) == 0
+
+	if ((dc->click_method != LIBINPUT_CONFIG_CLICK_METHOD_NONE
+			&& (libinput_device_config_click_get_methods(libinput_dev)
+				& dc->click_method) == 0)
```
ahesford, line 195:
```suggestion
	if ((libinput_device_config_click_get_methods(libinput_dev)
				& dc->click_method) != dc->click_method)
```

jp7677: I like this, but this results unfortunately in comparison with different signs :(

Consolatis: As discussed via IRC: refactor intended for the whole thing to use `UINT32_MAX` as "not-configured" placeholder instead.

Consolatis:
Thanks :)

PR #1563, opened 2024-02-25
src/server.c: filter out xwayland shell for usual clients

Consolatis:
I'd be fine with delaying this until after the release.


johanmalm:
Great catch. 
Yes, shall we leave it until after release?

Consolatis:
Although I did some quick tests with lazy restarts of xwayland and didn't see anything wrong this PR has the potential (though unlikely) to prevent xwayland from working. Thus lets delay this until after the release.

PR #1564, opened 2024-02-26
cursor: clear pointer focus on window moving/resizing interaction

tokyo4j:
With this change, the pointer focus is cleared from the surface at the start of window moving/resizing interaction.
This prevents sending prevents sending unnecessary `wl_pointer.frame` events while interactions.
I guess this behavior follows the protocol, as it says about `xdg_toplevel.move`:
> If triggered, the surface will lose the focus of the device (wl_pointer, wl_touch, etc) used for the move. It is up to the compositor to visually indicate that the move is taking place, such as updating a pointer cursor, during the move. There is no guarantee that the device focus will return when the move is completed.

Likewise, I removed the lines sending button release event at the end of interaction.

I believe this change doesn't make any functional changes in most applications.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1109,10 +1109,6 @@ cursor_process_button_release(struct seat *seat, uint32_t button,
 	}
 
 	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
-		if (pressed_surface) {
-			/* Ensure CSD clients see the release event */
-			return true;
-		}
 		return false;
 	}
 
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -153,6 +153,8 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 		view_restore_to(view, natural_geo);
 	}
 
+	wlr_seat_pointer_notify_clear_focus(seat->seat);
+
 	if (rc.resize_indicator) {
 		resize_indicator_show(view);
 	}


johanmalm:
All that cursor stuff took a long to get right... almost as hard as focusing the right surface.

Having a quick look at git blame, the following two pop up:

- https://github.com/labwc/labwc/commit/97ac960feebc552437fc2b536874671163b747ce
- https://github.com/labwc/labwc/issues/1053

Are you comfortable that we wouldn't regress on those?

tokyo4j:
I tested and confirmed #1053 doesn't reoccur with this change.

@Consolatis says in #1056:
> This patch fixes the issue by reordering the conditions where we return early. It also ensures that when we finish the move, we still send the release event to CSD clients.

I think the latter fix should be removed if we remove the cursor focus on interactive action. Otherwise, the client receives button release event after the pointer focus is returned back to the surface at the call to `cursor_update_focus()` in `interactive_cancel()`. Most apps will ignore it, but it doesn't make sense.

However, I'm not 100% sure this change doesn't cause any other regressions.

Anyway, this doesn't fix any practical bugs and is just my ego wanting labwc to behave like KWin/Mutter, so I turn this into a draft for now.
Maybe this is something to delegate to someone more familiar with cursor management.

tokyo4j:
@johanmalm

I'd like proceed this with PR again. After #2167, I noticed window menu is opened when I resized Firefox's picture-in-picture window with `W-Right` mousebind and released the button, because it listens for button release event (idk why) and sends `show_window_menu` request. This problem is fixed if we clear the pointer focus in `interactive_begin()`.

I agree that focus management is complex and dangerous to touch, but I think its risk is limited as we early-return in `cursor_process_motion()` and `cursor_process_button_release()` when the input mode is move/resize.

https://github.com/user-attachments/assets/f758a14c-a910-4b32-82ab-b7cfeb00e357

johanmalm:
Happy to proceed.

tokyo4j:
Superseded by #2455.

Issue #1273, opened 2023-12-02
Regression - video blocks in fullscreen with firefox


stefonarch:
As title, it recovers only when moving the cursor or a notification popup appears. In falkon and chromium (which runs under xwayland for some reason) it's not present, audio runs normal on.

In 0.6.6 it's not present too.

https://github.com/labwc/labwc/assets/10681413/d7f0b372-4940-49a2-8b5b-441ca06a6ed8

johanmalm:
Any chance of testing with https://github.com/Nefsen402/sway/tree/scene-graph (i.e. `--branch=scene-graph`)

Consolatis:
> In falkon and chromium (which runs under xwayland for some reason) it's not present, audio runs normal on.

Hm.. That somewhat sounds like a presentation timer issue.
I don' think the new set_gamma handler has anything to do with this.

@stefonarch If you have the chance, could you retest via `MOZ_ENABLE_WAYLAND=0 firefox` (so running firefox via xwayland) ?

stefonarch:
> @stefonarch If you have the chance, could you retest via `MOZ_ENABLE_WAYLAND=0 firefox` (so running firefox via xwayland) ?

No issue if running in xwayland

>Any chance of testing with https://github.com/Nefsen402/sway/tree/scene-graph (i.e. --branch=scene-graph)

You mean compiling this branch and testing with sway? Probably I didn't get it right.

johanmalm:
> > Any chance of testing with https://github.com/Nefsen402/sway/tree/scene-graph (i.e. --branch=scene-graph)
> 
> You mean compiling this branch and testing with sway? Probably I didn't get it right.

Yes, because if we see the problem there, that points to a problem in `wlroots` rather than what we've done in `labwc`. Just helps to fault-find.

```
git clone --branch=scene-graph https://github.com/Nefsen402/labwc.git
cd sway
meson setup build
ninja -C build
./build/sway/sway
```

IIRC Super-d is linked to dmenu by default to launch an app

stefonarch:
Don't ask my why but this issue is gone here, updated both arch and labwc in the meantime.

jp7677:
[...]

PR #1567, opened 2024-02-27
src/output.c: restore flicker free resize when running nested

Consolatis:
PR #1301 did partly revert the first variant because `lab_wlr_scene_output_commit()` (which uses the `output->pending` state in contrast to `wlr_scene_output_commit()`) was deemd to cause 'frozen' frames during video playback (#1273).

We are now back at using `lab_wlr_scene_output_commit()` for unrelated reasons and the cause of the original issue #1273 was likely that `wlr_scene_output_send_frame_done()` was only sent on a successful commit. This was changed to always be sent in the meantime.

So lets restore the flicker free nested resize as well.

---

This should be fine to include in 0.7.1. I don't expect it to make any difference other than stopping the nested resize flicker.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -150,6 +150,32 @@ output_request_state_notify(struct wl_listener *listener, void *data)
 	struct output *output = wl_container_of(listener, output, request_state);
 	const struct wlr_output_event_request_state *event = data;
 
+	/*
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks.

Issue #1557, opened 2024-02-24
Games feel jittery even running at around 100 FPS


Faugus:
I tried with KDE and this problem doesn't happen.

With VRR ON, games are buttery smooth.
I record gameplay videos and of course the VRR doesn't apply to the video and the recording is jittery/choppy.

```
[faugus@archlinux ~]$ labwc -v
labwc 0.7.0-164-g17bad48
```

This is my Environment file:
```
XKB_DEFAULT_LAYOUT=us
XKB_DEFAULT_VARIANT=intl
XCURSOR_THEME=Breeze_Snow
XCURSOR_SIZE=24
WLR_NO_HARDWARE_CURSORS=1
XDG_CURRENT_DESKTOP=gtk
```

Consolatis:
Depending on the game `WLR_NO_HARDWARE_CURSORS` may prevent direct scanout from happen. Unless you have issue with the cursor without it I'd recommend removing the env var.

Faugus:
I set `WLR_NO_HARDWARE_CURSORS` because in some games (like Uncharted) the mouse cursor was stuttering with VRR ON.
I just tried to remove it and nothing changed.

I tried to set -20 priority to labwc and nothing changed aswell.

Consolatis:
Just so that we are on the same page:
- VRR on, no recording, smooth game
- VRR off, no recording, smooth ?
- VRR on, recording, gameplay smooth, recording jittery ?
- VRR off, recording, gameplay smooth, recording jittery ?

Faugus:
- VRR on, no recording, smooth game
- VRR off, no recording, JITTERY game
- VRR on, recording, gameplay smooth, RECORDING JITTERY
- VRR off, recording, GAMEPLAY JITTERY, RECORDING JITTERY

Faugus:
I forgot to mention that on Last Epoch, the MangoHud's frametime graph is a straight line and running at around 100 FPS and feels jittery.

Consolatis:
Alright, so:
- VRR on smooth gameplay and VRR off jittery gameplay.
- the recording is always jittery

Do you have a chance to test the same with sway 1.9?
They are now using the same rendering backend as we do so it might shed a light on the area where we can proceed to debug.

Completely random thought:
Does the same happen when using the vulkan renderer? (`WLR_RENDERER=vulkan` in `~/.config/labwc/environment`)

Faugus:
> Do you have a chance to test the same with sway 1.9?

[...]

PR #1568, opened 2024-02-28
tearing: add fullscreen option

Consolatis:
If set, labwc will try to automatically enable tearing for fullscreen applications.

Fixes:
- #1557

---
@ahesford @Ph42oN review appreciated.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -157,6 +157,7 @@ Actions are used in menus and keyboard/mouse bindings.
 
 *<action name="ToggleTearing" />*
 	Toggles tearing for the focused window.
+	Requires the config option 'allowTearing'.
 
 *<action name="FocusOutput" output="HDMI-A-1" />*
 	Give focus to topmost window on given output and warp the cursor
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -142,11 +142,14 @@ this is for compatibility with Openbox.
 	*fullscreen* enables adaptive sync whenever a window is in fullscreen
 	mode.
 
-*<core><allowTearing>* [yes|no]
+*<core><allowTearing>* [yes|no|fullscreen]
 	Allow tearing to reduce input lag. Default is no.
 	This option requires setting the environment variable
 	WLR_DRM_NO_ATOMIC=1.
-	*yes* allow tearing if requested by the active window.
+
+	*yes* allows tearing if requested by the active window.
+
+	*fullscreen* enables tearing automatically for fullscreen windows.
 
 *<core><reuseOutputMode>* [yes|no]
 	Try to re-use the existing output mode (resolution / refresh rate).
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -35,6 +35,12 @@ enum adaptive_sync_mode {
 	LAB_ADAPTIVE_SYNC_FULLSCREEN,
 };
 
+enum tearing_mode {
+	LAB_TEARING_DISABLED = 0,
+	LAB_TEARING_ENABLED,
+	LAB_TEARING_FULLSCREEN,
+};
+
 enum tiling_events_mode {
 	LAB_TILING_EVENTS_NEVER = 0,
 	LAB_TILING_EVENTS_REGION = 1 << 0,
@@ -65,7 +71,7 @@ struct rcxml {
 	bool xdg_shell_server_side_deco;
 	int gap;
 	enum adaptive_sync_mode adaptive_sync;
-	bool allow_tearing;
+	enum tearing_mode allow_tearing;
 	bool reuse_output_mode;
 	enum view_placement_policy placement_policy;
 
--- a/include/view.h
+++ b/include/view.h
@@ -30,6 +30,12 @@ enum ssd_preference {
 	LAB_SSD_PREF_SERVER,
 };
 
+enum three_state {
+	LAB_STATE_UNSPECIFIED = 0,
+	LAB_STATE_ENABLED,
+	LAB_STATE_DISABLED
+};
+
 /**
  * Directions in which a view can be maximized. "None" is used
  * internally to mean "not maximized" but is not valid in rc.xml.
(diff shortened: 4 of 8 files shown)


Consolatis:
Technically I guess we should actually check for a tearing hint by the application and only if there is none allow tearing for fullscreen. Otherwise applications don't have a chance to disable tearing when fullscreen (end enabled by the user).

That would require `view->tearing_hint` to become some `yes|no|unspec` enum.

ahesford:
If we're going to respect a no-trading hint from the client, I don't think we should further complicate the interface by splitting the fullscreen option in two. If users wish to override the client hint, they can just use the `ToggleTearing` action.

Faugus:
I found a problem. If I play a video with `mpv` on fullscreen, it speeds up the playback.

Consolatis:
Automatically detecting a fullscreen game vs a fullscreen video player doesn't sound that easy. Not sure if we can make the new `fullscreen` work for both. If we end up respecting application / user configuration via `ToggleTearing` even if configured for `fullscreen`, the user could at least manually disable tearing again for cases likes mpv.

jp7677:
Imho video players should set the no-tearing hint (may be mpv already does that, I don’t know), together with https://github.com/labwc/labwc/pull/1568#issuecomment-1968095678 should disable tearing for those cases.

Faugus:
This option `video-sync=display-resample` in MPV's config file with `allowTearing` was causing the issue.
I set this a long time ago and I didn't remember.

Sorry.

Consolatis:
> This option `video-sync=display-resample` in MPV's config file with `allowTearing` was causing the issue. I set this a long time ago and I didn't remember.
> 
> Sorry.

No worries, thanks for mentioning the underlying reason.

Changed the PR to prefer client state (either set by the client or via `ToggleTearing`) over fullscreen.

Faugus:
If you want me to do any tests, just let me know.

Consolatis:
Feel free to test the current state of the PR. What should happen now is that with `<allowTearing>fullscreen</allowTearing>`
- tearing should be automatically enabled when in fullscreen
- when then using `ToggleTearing` you should still be able to disable / re-enable tearing while in fullscreen

Faugus:
On `mpv` with `video-sync=display-resample` in the config file, it works.
But it's not working with games.
If I toggle it, it breaks the gamepad compatibility like before, but it doesn't seems to turn on the `allowTearing`.
The game looks like the `allowTearing` is always off, cause I don't see the jittering.

Ph42oN:
Using this branch, tearing does not seem to work at all in [VRRTest](https://github.com/Nixola/VRRTest). This breaks enabling it even with ToggleTearing that used to work. But it continues to work in games that request tearing. I think it requests to not tear, and i think we should have way to override that, in previous patch i made this work by setting tearing_hint only if it is to enable it.

Ph42oN (commented):
I think this would be simple way to make ToggleTearing override it.
```diff
diff --git a/include/view.h b/include/view.h
index c3363623..c4bfca2f 100644
--- a/include/view.h
+++ b/include/view.h
@@ -170,6 +170,7 @@ struct view {
 	enum view_axis maximized;
 	bool fullscreen;
 	enum three_state tearing_hint;
+	bool tearing_override;
 	bool visible_on_all_workspaces;
 	enum view_edge tiled;
 	uint32_t edges_visible;  /* enum wlr_edges bitset */
diff --git a/src/action.c b/src/action.c
index b564f4fd..bdbee257 100644
--- a/src/action.c
+++ b/src/action.c
@@ -1007,6 +1007,7 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_TOGGLE_TEARING:
 			if (view) {
+				view->tearing_override = !view->tearing_override;
 				switch (view->tearing_hint) {
 				case LAB_STATE_UNSPECIFIED:
 				case LAB_STATE_DISABLED:
diff --git a/src/tearing.c b/src/tearing.c
index 3cf1aa92..4b2470ed 100644
--- a/src/tearing.c
+++ b/src/tearing.c
@@ -14,7 +14,7 @@ set_tearing_hint(struct wl_listener *listener, void *data)
 {
 	struct tearing_controller *controller = wl_container_of(listener, controller, set_hint);
 	struct view *view = view_from_wlr_surface(controller->tearing_control->surface);
-	if (view) {
+	if (view && !view->tearing_override) {
 		/*
 		 * tearing_control->hint is actually an enum:
 		 * WP_TEARING_CONTROL_V1_PRESENTATION_HINT_VSYNC = 0
-- 
2.43.0
```

Faugus:
@Ph42oN I just tested your changes and now I can't get `allowTearing` to work.
```
<core>
  <allowTearing>fullscreen</allowTearing>
</core>
...
<keybind key="C-F11">
  <action name="ToggleTearing" />
</keybind>
```
The game is jittery all the time. If I toggle it, it breaks the gamepad compatibility like before and the game is still jittery even running at high fps.

Faugus:
[G-SYNC Pendulum Demo](https://us.download.nvidia.com/downloads/cool_stuff/demos/Setup_G-SYNC_1.13.exe) is another great tool to test VRR, Tearing, Vsync...
Just run it with Wine.

Ph42oN:
> @Ph42oN I just tested your changes and it's behaving the same way it was.

Well my changes are to fix the problem that ToggleTearing does not override what application requests. That issue must be caused by something else.

Im not sure why you are having problem, it works as expected in games for me, at least on wine with xwayland. There was some talk about wine wayland previously... if its specific to that maybe i can test, but anyway wine wayland is unusable because camera does not turn at all with mouse on it (and this is not labwc specific, i tested on sway too).

Consolatis:
I can't really see an issue in the logic of this PR and can't test it so if somebody else wants to take this over feel free.

Regarding the question of overriding what a client sets with the protocol: unless the client repeatedly sets it to disabled in an endless loop, the `ToggleTearing` action should overwrite whatever was set by a client (with only this PR applied) so I am not sure what is going on here.

Ph42oN:
> Regarding the question of overriding what a client sets with the protocol: unless the client repeatedly sets it to disabled in an endless loop, the `ToggleTearing` action should overwrite whatever was set by a client (with only this PR applied) so I am not sure what is going on here.

That is what i would have expected too. But when making code to work with that expectation it doesn't work.

I believe tearing protocol was made to take info that clients were already giving on presentation, to not need to add tearing protocol support to every client. Maybe it just sets tearing hint based on that every frame.

Consolatis:
Replaced by
- #1941

PR #1571, opened 2024-02-28
docs/rc.xml: remove gap from the example config

Consolatis:
Some distro packages install the example config at `/etc/xdg/labwc/rc.xml` and thus users of those packages were having a gap 10 setting by default.

Lets remove the gap from the example config to match our intended default.


johanmalm:
Thanks

PR #1575, opened 2024-02-29
[wip] layers: play with interactivity

johanmalm:


src/layers.c
```diff
@@ -118,45 +118,55 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+		break;
+	case ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_ON_DEMAND:
+		if (!exclusive_focus(seat)) {
+			seat_set_focus_layer(seat, layer_surface);
+		}
+		break;
+	case ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_NONE:
+		if (seat->focused_layer) {
+			seat_set_focus_layer(seat, NULL);
+		}
```
Consolatis, line 169: I think this should be something like
```
if (seat->focused_layer == layer_surface) {
	focus_next_exclusive_focus_layer() || seat_set_focus_layer(seat, NULL);
}
```

Consolatis:
I like the general idea, we do have to update `src/input/cursor.c` as well though, currently we just give any layer surface focus on click. We should likely have a single entry point to decide if a layer surface should get focus, regardless of being called because of layer map() / keyboard_interactivity change or click.

src/layers.c
johanmalm: Done... in fixup commit.

johanmalm:
> I like the general idea, we do have to update `src/input/cursor.c` as well though, currently we just give any layer surface focus on click. We should likely have a single entry point to decide if a layer surface should get focus, regardless of being called because of layer map() / keyboard_interactivity change or click.

Very good point. Shall we sort `src/layers.c` first and then refactor?

src/layers.c
Consolatis: Still missing the `seat->focused_layer == layer_surface` check, otherwise just setting some random layershell client to no keyboard interactivity would change focus of another client.

Consolatis:
> > I like the general idea, we do have to update `src/input/cursor.c` as well though, currently we just give any layer surface focus on click. We should likely have a single entry point to decide if a layer surface should get focus, regardless of being called because of layer map() / keyboard_interactivity change or click.
> 
> Very good point. Shall we sort `src/layers.c` first and then refactor?

Not sure yet, do we need to handle map / keyboard interactivity change and cursor click on a layer surface differently?

johanmalm:
Superseded by #1599

PR #1578, opened 2024-03-02
Add cs language

zen0bit:
Since I cannot create new language on weblate...

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -1,0 +1,69 @@
+# Labwc pot file
+# Copyright (C) 2024
+# This file is distributed under the same license as the labwc package.
+# zenobit <zenobit@disroot.org>, 2024.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: labwc\n"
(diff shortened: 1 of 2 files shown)


zen0bit:
I did it by hand, not using weblate or poedit etc..

po/cs.po
```diff
@@ -0,0 +1,69 @@
...
+msgid "Always on Top"
+msgstr "Vždy nahoře"
+
+#: src/menu/menu.c:730
+msgid ""
+msgstr "Posunout doleva"
+
+#: src/menu/menu.c:737
+msgid ""
+msgstr "Posunout doprava"
```
Consolatis, line 57: `msgid` seems to be missing for these two.

zen0bit: Overlooked

01micko:
LGTM

Consolatis:
Thanks :)

stefonarch:
> Since I cannot create new language on weblate...

Users have to ask for for upgrading permissions, as there were once issues, there is a info but most users don't see it. Upgraded now for you.

PR #1581, opened 2024-03-02
Add weblate information

stefonarch:


01micko:
Just a thought..

Since you are adding sub headings then maybe one should be added for **Code Contributors** just to make it seem more consistent. @Consolatis @johanmalm ?

johanmalm:
I think the two new sub-heading would be `##` rather than `###`

@01micko Were you thinking of the `# Code Contributor` heading as a parent of `Coding Style`? If so, I think I prefer as is because `Coding Style` has a rather deep heading structure already.

stefonarch:
> I think the two new sub-heading would be `##` rather than `###`

Had that first, was to big IMHO.

01micko:
> I think the two new sub-heading would be `##` rather than `###`
> 
> @01micko Were you thinking of the `# Code Contributor` heading as a parent of `Coding Style`? If so, I think I prefer as is because `Coding Style` has a rather deep heading structure already.

No as a sub-heading of the NLS section, merely to distinguish between translating and coding responsibilities.

Maybe the wording could be different? Maybe just '__Coders__' or '__A Note for Coders__''? 'Code contributors' is mentioned in the first line of the paragraph any way so seems redundant.

From the file:

> Code contributors may need to update relevant files if their additions affect UI elements (at the moment only `src/menu/menu.c`). In this case the `po/labwc.pot` file needs to be updated so that translators can update their translations. Remember, many translators are _not_ coders!

johanmalm:
> No as a sub-heading of the NLS section, merely to distinguish between translating and coding responsibilities.

I see what you mean now. Yes, it feels like we need to split it into `## Translators` and a `## Coders`. Otherwise `### Github Pull Request` covers the rest of the `# NLS` section.

01micko:
@stefonarch I think that @johanmalm and my self are on the same page now :)

---

# Native Language Support

## Translators

### Weblate Instance

Translators can make an account at [LXQt Weblate](https://translate.lxqt-project.org/projects/labwc/labwc/)
and use the web interface. Adding new languages should work, otherwise the
administrators can be contacted.

### Github Pull Request

Translators can add their `MY_LOCALE.po` files to the `po` directory
based on `po/labwc.pot` and issue a pull request. To do this they can
generate their `MY_LOCALE.po` file in a few steps:

1. Edit the `po/LINGUAS` file to add their locale code in English
   alphabetical order to the field of locale codes.
2. Copy the `po/labwc.pot` to `po/MY_LOCALE.po`
3. Edit the newly generated `MY_LOCALE.po` file with some of their
contact and locale details in the header of the file then add the
translation strings under each English string.

[See this tutorial for further guidance](https://www.labri.fr/perso/fleury/posts/programming/a-quick-gettext-tutorial.html)

## Coders

Code contributors may need to update relevant files if their additions
affect UI elements (at the moment only `src/menu/menu.c`). In this case
the `po/labwc.pot` file needs to be updated so that translators can
update their translations. Remember, many translators are _not_ coders!

The process is fairly trivial however does involve some manual steps.

1. After adding and testing your code additions to satisfaction, backup
`po/labwc.pot`. You need the custom header from that file for the newly
generated .pot file in the next step.

2. From the root of the repository run this:

```
xgettext --keyword=_ --language=C --add-comments -o po/labwc.pot src/menu/menu.c
```

This generates a new pot file at `po/labwc.pot`

3. Copy the header from the original `labwc.pot` to the new one, check
for sanity and commit.

---

What do you think?

stefonarch:
Looks good. I forget to mention that people can just make suggestion for existing translations without login on weblate. Rewording:

"Translators can create an account"... and "... can be contacted. Suggestion for improving existing translation can be added without account."

01micko:
Good. Add that and I reckon this PR is ready to go.

01micko:
Just misses the `## Translators` and `## Coders` sub-headings`

Quoting Johan..

> I see what you mean now. Yes, it feels like we need to split it into `## Translators` and a `## Coders`. Otherwise it feels like `### Github Pull Request` covers the rest of the `# NLS` section.

Like I added https://github.com/labwc/labwc/pull/1581#issuecomment-1975735560

stefonarch:
I'm not full awake yet...

01micko:
LGTM 
@johanmalm

Consolatis:
Thanks.

When merging I suggest `Squash and merge`.

johanmalm:
Thanks

Issue #1582, opened 2024-03-02
view_get_adjacent_output() picks wrong output when there are more than two outputs


johanmalm:
[recording.webm](https://github.com/labwc/labwc/assets/6955353/def89613-70f1-4771-ad40-aba51f6a19cb)

johanmalm:
Using `MoveToEdge` action.

PR #1583, opened 2024-03-02
view: fix view_get_adjacent_output() bug and implement wrap argument for MoveToOutput

johanmalm:
...when using more than two outputs.

Use (centre-x,centre-y) rather than (1,0) and (0,1).

Fixes: #1582

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -1530,6 +1530,10 @@ view_get_adjacent_output(struct view *view, enum view_edge edge)
 		return NULL;
 	}
 
+	struct wlr_box box = output_usable_area_in_layout_coords(view->output);
+	int lx = box.x + box.width / 2;
+	int ly = box.y + box.height / 2;
+
 	/* Determine any adjacent output in the appropriate direction */
 	struct wlr_output *new_output = NULL;
 	struct wlr_output *current_output = output->wlr_output;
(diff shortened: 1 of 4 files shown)


johanmalm:
[recording.webm](https://github.com/labwc/labwc/assets/6955353/4d3f11d8-271f-4bd7-ac0b-57c2edfd3e53)

Using `MoveToOutput` with `direction="right"`... and then some `MoveToEdge`

ahesford:
This seems like a good fix. Note that there is some ambiguity if there are multiple displays that are tiled next to the current output:

```
+-----+-----+
|  x  |     |
+-----+  z  |
|  y  |     |
+-----+-----+
```

However, there is never a perfect way to reconcile that ambiguity and using the center of the reference output is as good a choice as any other point on the output.

The `opposite_direction` function duplicates logic in the `validate_single_region_edge‎` function of `edges.c`, so it might be useful to split that off into a helper function.

johanmalm:
> However, there is never a perfect way to reconcile that ambiguity and using the center of the reference output is as good a choice as any other point on the output.

We could use the center of the view rather than the output I suppose.

> The `opposite_direction` function duplicates logic in the `validate_single_region_edge‎` function of `edges.c`, so it might be useful to split that off into a helper function.

That's a different enum. `enum wlr_direction` vs `enum wlr_edges`

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_output_layout.h?ref_type=heads#L148

ahesford:
Woops, you're right. Although I suppose we could move from `wlr_edges` to `wlr_direction` to unify the two...

src/view.c
```diff
@@ -1530,29 +1567,21 @@ view_get_adjacent_output(struct view *view, enum view_edge edge)
 		return NULL;
 	}
 
+	struct wlr_box box = output_usable_area_in_layout_coords(view->output);
```
Consolatis, line 1570: Could just be `output`, it is defined above.

Consolatis:
Other than the nitpick above this LGTM.

`subprojects/wlroots/types/wlr_output_layout.c` `wlr_output_layout_output_in_direction()` indeed uses `ref_lx` and `ref_ly` as absolute layout coordinates rather than relative to the given `reference` output so `0` and `1` is definitely wrong.

src/view.c
johanmalm: Good spot. Done.

johanmalm:
> Other than the nitpick above this LGTM.

Have tidied up commit messages. Will merge shortly.

johanmalm:
> Woops, you're right. Although I suppose we could move from `wlr_edges` to `wlr_direction` to unify the two...

Yes. With `wlr_output_layout_adjacent_output()` we're stuck with `wlr_direction`, but yes +1 for common helpers :smile:

src/view.c
```diff
@@ -1530,32 +1567,28 @@ view_get_adjacent_output(struct view *view, enum view_edge edge)
...
-		break;
+	enum wlr_direction direction = get_wlr_direction(edge);
+	new_output = wlr_output_layout_adjacent_output(layout, direction,
+		current_output, lx, ly);
+
+	/* Try wrapping */
+	if (!new_output) {
+		new_output = wlr_output_layout_farthest_output(layout,
+			opposite_direction(direction), current_output, lx, ly);
 	}
```
Consolatis, line 1586: Should we unconditionally do this? I think it should be a disabled-by-default behavior.

johanmalm:
Have made the whole wrap thing a disabled-by-default option for `MoveToOutput` only.
Can create separate PR, but if we want to go with the wrap option, I guess it saves me the effort of just keeping as is. No rush to look at this of course :smile:

Consolatis:
LGTM, not tested.

Issue #1530, opened 2024-02-13
Cleanup envrionment when exiting


tfemby:
Currently the dbus/systemd environment isn't being cleaned upon exit. This may effect applications such as kitty with a wayland and x11 backend when swapping from Labwc over to Openbox.

I spoke with @Consolatis on irc about this issue and he suggested that I run `systemctl --user unset-envrionment DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP`. This solved the issue for me but as Consolatis suggested, this is likely something that should be done by Labwc upon exit.

Consolatis:
Thinking about this some more.. maybe we should prevent any of the dbus / systemd calls when we are running nested?
Its way to easy to accidentally use the same dbus session in this case and we really don't want a nested compositor to overwrite / delete the main DRM session vars.

tfemby:
I've been looking into this a little myself and have thought about `loginctl terminate-user` although that's maybe a little too aggressive.

ahesford:
The `terminate-user` action is definitely out of scope and probably not what you really desire anyway. You most likely don't want to kick yourself out of an SSH session just because you also quit labwc.

The only session cleanup labwc should attempt is to unset the variables it sets on launch. If you want more complex session management, the combination of `autostart` and the new `shutdown` script will give you all the control you need. (I use the pair to set up the `dinit` environment and start some Wayland services, then tear it all down on exit.)

PR #1585, opened 2024-03-03
session: only update activation environment when running DRM backend or by explicit request

ahesford:
This is an attempt to address https://github.com/labwc/labwc/issues/1530#issuecomment-1975010257 by only updating the dbus/systemd activation environment when the DRM backend is in use or the `LABWC_UPDATE_ACTIVATION_ENV` environment variable explicitly requests an update.

### DIFF
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -6,11 +6,14 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <wlr/backend/drm.h>
+#include <wlr/backend/multi.h>
 #include <wlr/util/log.h>
 #include "common/buf.h"
 #include "common/dir.h"
 #include "common/file-helpers.h"
 #include "common/mem.h"
+#include "common/parse-bool.h"
 #include "common/spawn.h"
 #include "common/string-helpers.h"
 #include "config/session.h"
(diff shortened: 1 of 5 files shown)


src/config/session.c
```diff
@@ -74,14 +77,53 @@ read_environment_file(const char *filename)
 }
 
 static void
-update_activation_env(bool initialize)
+backend_check_drm(struct wlr_backend *backend, void *is_drm)
 {
+	if (is_drm && wlr_backend_is_drm(backend)) {
```
johanmalm, line 82: Do we need the `is_drm &&` bit here?

johanmalm:
One minor comment. Otherwise LGTM.
Not tested.
We ought to document this, but looking at docs I think the whole systemctl/dbus thing has slipped through the net, so could include here or tackle it as a separate PR.

src/config/session.c
```diff
@@ -74,14 +77,53 @@ read_environment_file(const char *filename)
...
+static bool
+should_update_activation(struct server *server)
+{
+	assert(server);
+
+	static const char *act_env = "LABWC_UPDATE_ACTIVATION_ENV";
+	char *env = getenv(act_env);
+	if (env) {
+		/* Respect any valid preference from the environment */
+		int enabled = parse_bool(env, -1);
```
Consolatis, line 96: Should we add "0" and "1" to `src/common/parse-bool.c` ?

Consolatis:
Another minor comment. Otherwise LGTM as well.

src/config/session.c
johanmalm: Good idea. Seems right for environment variables in particular.

droc12345:
I don't set my env vars in the env file, I have a shell script that does the initialization before I fire off a compositor.
It doesn't look like this would conflict with the way I do things, but I'll keep an eye on it.

ahesford:
Minor comments addressed in the latest version; I'll put together some documentation on this behavior.

ahesford:
I took a crack at updating the documentation; comments welcome.

Consolatis:
Perfect. Thanks :)

Issue #1589, opened 2024-03-06
Action MoveToOutput uses name attribute twice


bnason:
The action `MoveToOutput` uses name attribute twice. I expected that the second one would be called `output` instead.

`<action name="MoveToOutput" name="HDMI-A-1" direction="value" wrap="no" />`

[Documentation](https://labwc.github.io/labwc-actions.5.html#entry_action_name=movetooutput_name=hdmi-a-1_direction=value_wrap=no)

Consolatis:
I am not even sure this works at all, the node name should show up as `name.action` and that usually creates a new action rather than modifying the current one.

bnason:
Attempting to use the action as described produces the following error parsing the config file
`Entity: line 1: parser error : Attribute name redefined`

Consolatis:
- #1590

Testing welcome.

bnason:
@Consolatis I have tested this and the action is now working.

PR #1590, opened 2024-03-06
src/action.c: fix MoveToOutput action by renaming 'name' argument

Consolatis:
We were using the argument name 'name' before
which is already used by the action itself:
`<action name="MoveToOutput" name="HDMI-A-1" />`

Change the argument name to "output" which also
matches the `FocusOutput` action.

Fixes:
- #1589
Reported-by: @bnason (thanks!)
---
Only lightly tested.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -163,12 +163,12 @@ Actions are used in menus and keyboard/mouse bindings.
 	to the center of the window. If the given output does not contain
 	any windows, the cursor is centered on the given output.
 
-*<action name="MoveToOutput" name="HDMI-A-1" direction="value" wrap="no" />*
+*<action name="MoveToOutput" output="HDMI-A-1" direction="value" wrap="no" />*
 	Moves active window to other output, unless the window state is
 	fullscreen.
 
-	If *name* is specified, the window will be sent directly to the specified
-	output and *direction* is ignored. If *name* is omitted, *direction* may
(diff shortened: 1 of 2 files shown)


bnason:
I tested this in my environment and confirmed it works now.

johanmalm:
Thanks. Wow, can't believe we let that one in!

jp7677:
Yeah, sorry, that one slipped in while adjusting the original PR to feedback… made the final tests only with `direction`. Learned a valuable lesson …

Consolatis:
We didn't spot it either during review :)
With our review rules we try to prevent things like this from happening but nobody is perfect and things may slip through from time to time.

johanmalm:
> Yeah, sorry, that one slipped in while adjusting the original PR to feedback… made the final tests only with `direction`. Learned a valuable lesson …

Oh yeah... 100% what @Consolatis said. I was just smiling to myself because it just goes to show how easy it is to miss things, both for coders + reviewers. @jp7677 you always seem to take great care with patches - this is nothing. 
Yesterday, we spotted a `LABWC_PWD` instead of a `LABWC_PID` in the documentation. We think we're careful, but just goes to show how the brain works (or sometimes doesn't).

jp7677:
Thanks for the kind words, no worries, I didn't take it too personal ;) and indeed, cool to have this fixed and I'm happy to read that labwc is getting more attention and lots of features are actually getting used in the wild.

PR #1592, opened 2024-03-07
src/output.c: refactor virtual output related functions

Consolatis:
This commit moves the virtual output related functions
into their own file at `src/output-virtual.c` with its
own include file to reduce `include/labwc.h` bit by bit.

Additionally, it removes the need to keep the
`server->headless.pending_output_name` char array around
by temporarily disconnecting the handler when creating a
new virtual output. This allows to set the output name
right in the `output_virtual_add()` call rather than to
store the pending name until the new output event handler
has been called.

It also makes adding a virtual fallback output easier in
a follow-up PR.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -211,7 +211,6 @@ struct server {
 	struct wlr_backend *backend;
 	struct headless {
 		struct wlr_backend *backend;
-		char pending_output_name[4096];
 	} headless;
 	struct wlr_session *session;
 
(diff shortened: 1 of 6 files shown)


Consolatis:
> I have a slight thing about `store_wlr_output` but let's go with it. If we can see a way to simplify when the remainder of the code joins, we can pick it up then.

If we want to use the new `wlr_output` within the new output handlers (e.g. to compare the new output pointer against it) I don't see a way around it. Obviously it could be renamed to something like `struct wlr_output **wlr_output` or something.

johanmalm:
> If we want to use the new `wlr_output` within the new output handlers (e.g. to compare the new output pointer against it) I don't see a way around it. Obviously it could be renamed to something like `struct wlr_output **wlr_output` though.

I'm good with it. Just a hang-up... but probably an irrational one.

PR #1593, opened 2024-03-07
Ignore move/resize request from clients when the view is not pressed

tokyo4j:
This is relevant for touchpad taps with `tapAndDrag` disabled.

When the user taps the client surface (e.g. mpv and un-decorated titlebar of chromium) with the setting above, libinput sends button press & release signals so quickly that the compositor receives move/resize request from the client after the button release signal is processed, so `interactive_finish()` is never called.

Since the checks I added are duplicated in `xdg.c` and `xwayland.c`, maybe it's better to move them into `view-impl-common.c` and share it.

This includes minor style fixes and additional comments on `seat.pressed`.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -135,6 +135,10 @@ struct seat {
 	 * This allows to keep dragging a scrollbar or selecting text even
 	 * when moving outside of the window.
 	 *
+	 * It is also used to:
+	 * - determine the target view for action in "Drag" mousebind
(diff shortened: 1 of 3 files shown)


johanmalm:
LGTM. The world of touch, eh?

I just had to think for a minute about XWayland requests for resize - but I'm pretty sure that's done via a `request_configure` which is separate.

Issue #1572, opened 2024-02-29
Keyboard focus on panel menus


stefonarch:
In v. 0.6.5 at least up/down arrows were working in lxqt-panel `(wayland_taskbar` PR) menus, as seen in KDE Neon. Only compositor where this is working is `kwin_wayland` atm.

* impossible to type in search fields
* no keyboard navigation

Consolatis:
Might update this post later with more thoughts.

For on-demand I think we should only ever give focus automatically during map to keep it the same as for usual xdg and xwayland surfaces. In my understanding on-demand just means "apply usual focus semantics", e.g. focus on map and on cursor button press. So for `4.` `excl-overlay-to-background` sounds fine but I am not sure about `on-demand-overlay-to-background`. If there is no exclusive candidate, the next focus should either be on some remaining layer surface with on-demand set or on some xdg / xwayland surface. But we don't have a combined history of the last keyboard focused layershell / xdg / xwayland surface so not sure how to handle this.

stefonarch:
> 2) Should keyboard-focus be given to layer-shell popups on creation. I'd never thought about that, and the protocol doesn't specify. Based on what you've said about Qt6 on other compositors, I don't suspect anyone else does it either. 

Only with Hyprland and kwin keyboard navigation in lxqt-panel's menus is working atm.

Consolatis:
That might also be relevant here:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L1244

johanmalm:
> As well as this: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/blob/master/unstable/wlr-layer-shell-unstable-v1.xml?ref_type=heads#L275

Good spot. That links says "This setting is inherited by child surfaces set by the get_popup request." I read that as the client should ask for interactivity when opening the popup.

@stefonarch I'll take a look at what hyprland have done later.

johanmalm:
I've got it working with [tint](https://github.com/johanmalm/tint) test panel by setting keyboard-interactivity to on-demand and by applying the patch below.

@Consolatis There is something weird going on here. `ctx->type` is never `LAB_SSD_LAYER_SURFACE` but always `LAB_SSD_LAYER_SUBSURFACE`. Have even tried with a home made thing that doesn't do sub-surfaces just to make sure it's not a Gtk/Qt thing. 

[...]

Issue #1554, opened 2024-02-24
Modifiers not reported in click events on a layer-shell surface in on-demand keyboard interactivity mode


Tamaranch:
They are in fact sometimes reported, but rarely. I have this problem initially with xfce4-panel, but it's reproducible with a simple GTK window to which you apply `gtk_layer_init_for_window()`.

For comparison, it works pretty well with wayfire 0.8.0, although you'd better left-click on the panel first to make sure it's gained focus (which is not at all what a user expects, but hey...).

Originally reported against gtk-layer-shell, but it turns out the problem is with labwc (0.7.0, git master): https://github.com/wmww/gtk-layer-shell/issues/175

johanmalm:
`wlroots` only broadcasts modifiers to the client with the surface that currently has keyboard-focus.

IIRC the protocol is not clear on approach here. Many compositors have opted to take the `wlroots` approach (e.g. Cosmic) - I think to avoid 'waking up' clients every time a modifier is pressed.

I'm of a mind to patch `labwc` to optionally broadcast modifiers to all clients to solve this type of problem. It would also have the benefits of all clients knowing the `keyboard-group` (e.g. what layout is in use).

Consolatis:
> Modifiers not reported in click events on a layer-shell surface in on-demand keyboard interactivity mode

Hm.. that indeed sounds a bug from our side. Maybe it is a race condition between receiving the mouse button event vs the keyboard modifiers. E.g. if you run via `WAYLAND_DEBUG=1`, do the modifiers maybe show up after the button event?

> `wlroots` only broadcasts modifiers to the client with the surface that currently has keyboard-focus.

In this case (with on-demand keyboard interactivity) the keyboard focus should be on the layershell client, if it isn't there might be something wrong.

> I'm of a mind to patch `labwc` to optionally broadcast modifiers to all clients to solve this type of problem. It would also have the benefits of all clients knowing the `keyboard-group` (e.g. what layout is in use).

That would likely interfere with the layout-per-window setting.

Tamaranch:
Ok so I've been fooled once again, sorry. _When_ the layer-shell surface gets the focus, the modifiers are well reported, whether in exclusive or on-demand mode. The problem is that it doesn't always get the focus as it should in on-demand mode.

This can be difficult to check with xfce4-panel, as several things interact (including overlay layer-shell surfaces for external plugins, themselves in on-demand mode), but with a simple GTK window transformed into layer-shell, it's easily reproducible:
[...]

PR #1594, opened 2024-03-07
cursor: process layer subsurfaces in `cursor_button_press()` and fix a layer-surface bug in `get_cursor_context()`

johanmalm:
...to give keyboard focus to layer-shell clients if exclusive or on-demand interactivity is set, so that menu popups can be navigated with the keyboard. This still only works if the client is in top (or overlay) layers. Support for bottom and background to be done as a separate patch set.

Revert 06b19f0 to process layer-shell subsurfaces in `cursor_button_press()`, but only when their parent layer-shell surface has keyboard interactivity.

In `get_cursor_context()` process layer-surfaces and layer-subsurfaces from surfaces rather than nodes.

Background:

Commit 06b19f0 (issue #1131) disabled processing of layer-shell subsurfaces in cursor_button_press() because when pressing a task in Waybar (Gtk panel using layer-shell subsurfaces) the foreign-toplevel minimize-raise action did not work correctly as the action logic relied on the recipient window being activated and by clicking on the panel, the panel itself was both surface-focusd and activated (and thus the window de-activated).

The un-intended consequence was that by not responding to layer-subsurface cursor buttons presses, layer-shell clients (such as panels) were not given keyboard focus if they indeed wanted it by setting exclusive or on-demand keyboard interactivity.

The good news is that that following @jlindgren90's refactoring (various) the only place where we call `view_set_actived()` is in `focus_change_notify()` in `seat.c` and we now only do it for views (bb8f0bc).

Another side-effect (positive) of 06b19f0 was that a Waybar dnd bug was fixed (pointer-serial-number validation failure).

Have tested with sfwbar, waybar and tint (test-panel) the following results:
- Minimize-raise works even when on-demand keyboard interactivity is set
- Keyboard interactivity is given popup-menus (sfwbar and tint) when the panels are in the top layer (support for bottom will be as a separate patch set)
- Waybar dnd still works (even when hard-coding keyboard-interactivity)

References:
- https://github.com/labwc/labwc/commit/bb8f0bc960dca192b8579d67297c0586ec20bfe0
- https://github.com/labwc/labwc/blob/40ce95a68cf19796dd67b0527fddfdbe46181805/src/seat.c#L481-L483
- https://github.com/labwc/labwc/blob/40ce95a68cf19796dd67b0527fddfdbe46181805/src/dnd.c#L24
- https://github.com/johanmalm/tint

Fixes: #1572

### DIFF
--- a/include/common/surface-helpers.h
+++ b/include/common/surface-helpers.h
@@ -1,0 +1,16 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_SURFACE_HELPERS_H
+#define LABWC_SURFACE_HELPERS_H
+
+struct wlr_surface;
+struct wlr_layer_surface_v1;
+
+/**
+ * subsurface_parent_layer() - Get wlr_layer_surface from layer-subsurface
+ * @wlr_surface: The wlr_surface of the wlr_subsurface for which to get the
+ *               layer-surface.
+ */
+struct wlr_layer_surface_v1 *subsurface_parent_layer(
+	struct wlr_surface *wlr_surface);
+
+#endif /* LABWC_SURFACE_HELPERS_H */
--- a/src/common/meson.build
+++ b/src/common/meson.build
@@ -13,6 +13,7 @@ labwc_sources += files(
   'scaled_font_buffer.c',
   'scaled_scene_buffer.c',
   'scene-helpers.c',
+  'surface-helpers.c',
   'spawn.c',
   'string-helpers.c',
 )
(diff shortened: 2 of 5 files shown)


stefonarch:
Tested with `lxqt-panel` (after changing `KeyboardInteractivityNone` to `KeyboardInteractivityOnDemand`)  and `pcmanfm-qt --desktop`, looks working fine.

Tamaranch:
Looks like it would fix #1554 :)

Tamaranch:
I can also confirm that it does allow you to navigate xfce4-panel menus (internal plugins in the panel surface, or external plugins on their own layer-shell surface in overlay), and that it also solves that old problem of not being able to close the menu via the Escape key (we must have talked about this somewhere, but I can't remember where :)).

johanmalm:
Thanks both for testing.
Have refactored a bit following conversation with @Consolatis 
Ready for final review + merge.

Consolatis:
Had a quick test nested it improves things quite a bit, thanks for your work on this :)

PR #1599, opened 2024-03-08
layer: change logic for giving keyboard-interactivity

johanmalm:
Use the following logic:
 - Exclusive: Grant regardless of layer. Previously it was only given if in top or overlay layers. Grant if in the same or higher layer than (nearer overlay) any other client with exclusive interactivity.
 - On-demand: Grant only if no other layer-shell client has exclusive keyboard interactivity. Previously it was treated the same as exclusive.
 - None: Unset focus if the commit associated with the 'none' came from the currently focused layer. Previously it was just unset regardless.

Remove unecessary top/overlay check in seat_set_focus_layer(). Such checks should be made prior to calling the function.

- [x] NOTE 1: This is to follow https://github.com/labwc/labwc/pull/1594
- [x] NOTE 2: We should also call `layer_try_set_focus()` from `src/input/cursor.c`. Should be simple 2-line-change after #1594

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -118,45 +118,64 @@ handle_output_destroy(struct wl_listener *listener, void *data)
 	wlr_layer_surface_v1_destroy(layer->scene_layer_surface->layer_surface);
 }
 
-static void
-process_keyboard_interactivity(struct lab_layer_surface *layer)
+static bool
+focused_layer_has_exclusive_interactivity(struct seat *seat)
 {
-	struct wlr_layer_surface_v1 *layer_surface = layer->scene_layer_surface->layer_surface;
-	struct seat *seat = &layer->server->seat;
+	if (!seat->focused_layer) {
+		return false;
+	}
+	return seat->focused_layer->current.keyboard_interactive ==
+		ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE;
+}
 
-	if (layer_surface->current.keyboard_interactive
-			&& layer_surface->current.layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP) {
-		/*
-		 * Give keyboard focus to surface if
-		 * - keyboard-interactivity is 'exclusive' or 'on-demand'; and
-		 * - surface is in top/overlay layers; and
-		 * - currently focused layer has a lower precedence
-		 *
-		 * In other words, when dealing with two surfaces with
-		 * exclusive/on-demand keyboard-interactivity (firstly the
-		 * currently focused 'focused_layer' and secondly the
-		 * 'layer_surface' for which we're just responding to a
-		 * map/commit event), the following logic applies:
-		 *
-		 * | focused_layer | layer_surface | who gets keyboard focus |
-		 * |---------------|---------------|-------------------------|
-		 * | overlay       | top           | focused_layer           |
-		 * | overlay       | overlay       | layer_surface           |
-		 * | top           | top           | layer_surface           |
-		 * | top           | overlay       | layer_surface           |
-		 */
+/*
+ * Precedence is defined as being in the same or higher (overlay is highest)
+ * than the layer with current keyboard focus.
+ */
+static bool
+has_precedence(struct seat *seat, enum zwlr_layer_shell_v1_layer layer)
+{
+	if (!seat->focused_layer) {
+		return true;
+	}
+	if (!focused_layer_has_exclusive_interactivity(seat)) {
+		return true;
+	}
+	if (layer >= seat->focused_layer->current.layer) {
+		return true;
+	}
+	return false;
+}
 
-		if (!seat->focused_layer || seat->focused_layer->current.layer
-				<= layer_surface->current.layer) {
+void
+layer_try_set_focus(struct seat *seat, struct wlr_layer_surface_v1 *layer_surface)
+{
+	switch (layer_surface->current.keyboard_interactive) {
+	case ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE:
+		wlr_log(WLR_DEBUG, "interactive-exclusive '%p'", layer_surface);
+		if (has_precedence(seat, layer_surface->current.layer)) {
 			seat_set_focus_layer(seat, layer_surface);
 		}
-	} else if (seat->focused_layer
-			&& !seat->focused_layer->current.keyboard_interactive) {
-		/*
-		 * Clear focus if keyboard-interactivity has been set to
-		 * ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_NONE
-		 */
-		seat_set_focus_layer(seat, NULL);
+		break;
+	case ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_ON_DEMAND:
+		wlr_log(WLR_DEBUG, "interactive-on-demand '%p'", layer_surface);
+		if (!focused_layer_has_exclusive_interactivity(seat)) {
+			seat_set_focus_layer(seat, layer_surface);
+		}
+		break;
+	case ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_NONE:
+		wlr_log(WLR_DEBUG, "interactive-none '%p'", layer_surface);
+		if (seat->focused_layer == layer_surface) {
+			wlr_log(WLR_DEBUG, "unset focus '%p'", layer_surface);
+			/*
+			 * TODO: consider transferring focus other layer-shell
+			 * clients with exclusive focus (that in the case of
+			 * multiple clients we could have stolen it from and
+			 * arguably should give it back to).
+			 */
+			seat_set_focus_layer(seat, NULL);
+		}
+		break;
 	}
 }
 
@@ -184,7 +203,8 @@ handle_surface_commit(struct wl_listener *listener, void *data)
 	}
 	/* Process keyboard-interactivity change */
 	if (committed & WLR_LAYER_SURFACE_V1_STATE_KEYBOARD_INTERACTIVITY) {
-		process_keyboard_interactivity(layer);
+		struct seat *seat = &layer->server->seat;
+		layer_try_set_focus(seat, layer_surface);
 	}
 
 	if (committed || layer->mapped != layer_surface->surface->mapped) {
@@ -250,7 +270,8 @@ handle_map(struct wl_listener *listener, void *data)
 	 * the scene. See wlr_scene_surface_create() documentation.
 	 */
 
-	process_keyboard_interactivity(layer);
+	struct seat *seat = &layer->server->seat;
+	layer_try_set_focus(seat, layer->scene_layer_surface->layer_surface);
 }
 
 static void
--- a/src/layers.c
+++ b/src/layers.c
@@ -318,8 +318,10 @@ handle_map(struct wl_listener *listener, void *data)
 	 * the scene. See wlr_scene_surface_create() documentation.
 	 */
 
-	struct seat *seat = &layer->server->seat;
-	layer_try_set_focus(seat, layer->scene_layer_surface->layer_surface);
+	/*
+	 * Processing of keyboard-interactivity changes is done in the
+	 * commit-handler.
+	 */
 }
 
 static void
--- a/src/layers.c
+++ b/src/layers.c
@@ -118,6 +118,62 @@ handle_output_destroy(struct wl_listener *listener, void *data)
 	wlr_layer_surface_v1_destroy(layer->scene_layer_surface->layer_surface);
 }
 
+static inline bool
+has_exclusive_interactivity(struct wlr_scene_layer_surface_v1 *scene)
+{
+	return scene->layer_surface->current.keyboard_interactive
+		== ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE;
+}
+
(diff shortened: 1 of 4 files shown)


src/layers.c
```diff
@@ -118,45 +118,70 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+	if (seat->focused_layer->current.keyboard_interactive !=
+			ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE) {
+		return false;
+	}
+	return true;
+}
+
+/*
+ * Precedence is defined as being in the same or a higher (overlay is highest)
+ * than the layer with current keyboard focus.
```
Consolatis, line 136: `or a higher [..] than the layer`

src/layers.c
```diff
@@ -118,45 +118,70 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+		wlr_log(WLR_DEBUG, "interactive-none '%p'", layer_surface);
+		if (seat->focused_layer == layer_surface) {
+			wlr_log(WLR_DEBUG, "unset focus '%p'", layer_surface);
+			/*
+			 * TODO: consider transferring focus other layer-shell
+			 * clients with exclusive focus (that in the case of
+			 * multiple clients we could have stole it from and
+			 * arguably should give it back to).
+			 */
+			seat_set_focus_layer(seat, NULL);
```
Consolatis, line 182: Should we do this in this PR? `seat_set_focus_layer(seat, NULL)` gives keyboard focus to the topmost usual window currently. I assume we should do something like this instead:
- exclusive-overlay-and-top
- on_demand-overlay-and-top ?
- topmost_view ?
- exclusive-bottom-and-background
- on_demand-bottom-and-background

and break on the first match. Not quite sure about `on_demand-overlay-and-top` before `topmost_view` or the other way around.

src/layers.c
johanmalm: Done.

src/layers.c
johanmalm: I've added `try_to_focus_next_layer()`

It merely does the following (in order of precedence):

- Give focus to last added overlay/top layer-shell client with exclusive interactivity. I think this is quite an edge case in practice, but seems tidy. For example if you run two bemenu-run instances (I'm not sure why you would, but still!) when the first one exits, the second one gets keyboard focus.

- Give focus to topmost toplevel (=view) if one exists. We actually already do this, so am just noting it here for completeness. https://github.com/labwc/labwc/blob/a5fcbfaf720fc6c2486d850012215102e90e04bf/src/seat.c#L664

> maybe some previously-focused?

Well, if we're talking `exclusive` clients, presumable they had focus before it was stolen from them, so logic above just give it back to them.

They've been inserted at the back of the list in wlroots/types/scene/wlr_scene.c (`wl_list_insert(parent->children.prev, &node->link);` so iterating in reverse gives focus to last added client first.

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/scene/wlr_scene.c?ref_type=heads#L72

> on_demand-overlay-and-top?

I haven't implemented this. I think it would only make sense if such a client had focus and then it was stolen from them by another layer-shell client. Otherwise I think it's more intuitive to give it to the topmost toplevel. I suspect it would be hard to implement too and would probably require a list to keep track of clients from whom keyboard interactivity was stolen. Suggest we leave that for a separate PR - if ever.

> exclusive-bottom-and-background? 

Decided not to implement. It would be dead easy to do, but not sure what the use-case is. I don't mind either way though.

> on_demand-bottom-and-background

Same argument as on-demand-overlay-and-top.

> Not quite sure about on_demand-overlay-and-top before topmost_view or the other way around.

If you take the case of a panel (which could be in `top` or `bottom`) the question is really whether it would be better to give focus to the topmost toplevel or the panel.

What we could consider (but I prefer do consider this in a separate PR) is to close popups when focus is stolen. That would feel quite 'normal' and reduce the sense that focus should be given back at some point.

> The logic could theoretically be moved to seat_set_focus_layer(seat, NULL)

Looks like `seat_input_inhibit()` wants to clear focus without fallbacks (i.e. need to keep `seat_set_focus_layer()` as is in support of that)

> and should likely also be called on unmap if the layer was focused at that point.

Done.

johanmalm:
The only other thing I can think of now it so de-conflict layer client keyboard-interactivity with a seats popup-grab. When xdg-shell popup is open (and grab is on) we can't give focus to a layer-shell client.

Issue can be tested with `sleep 5 ; bemenu-run` with menu (popup) open in an xdg-shell application.

Suggest I draft something as a separate PR for that though.

ahesford:
The logic seems reasonable here, although I don't really have a setup that can test these changes and I'm really only juding the flow described in the comments.

src/layers.c
```diff
@@ -118,45 +118,116 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+		== ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE;
+}
 
-	if (layer_surface->current.keyboard_interactive
-			&& layer_surface->current.layer >= ZWLR_LAYER_SHELL_V1_LAYER_TOP) {
+/*
+ * Try to transfer focus other layer-shell clients with exclusive focus on the
+ * output nearest to the cursor. In none exist (which is likely to generally be
+ * the case) just unset layer focus and try to give it to the topmost toplevel
+ * if one exists.
```
Consolatis, line 132: Try to transfer focus `to` other layer-shell clients ?

src/layers.c
```diff
@@ -118,45 +118,116 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+ */
+static void
+try_to_focus_next_layer_or_toplevel(struct server *server)
+{
+	struct seat *seat = &server->seat;
+	struct output *output = output_nearest_to_cursor(server);
+
+	int overlay = ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY;
+	int top = ZWLR_LAYER_SHELL_V1_LAYER_TOP;
```
Consolatis, line 141: Should we `s/int/enum zwlr_layer_shell_v1_layer/` here?

src/layers.c
```diff
@@ -118,45 +118,116 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
-		 * | overlay       | overlay       | layer_surface           |
-		 * | top           | top           | layer_surface           |
-		 * | top           | overlay       | layer_surface           |
+		 * In wlr_scene.c they were added at end of list so we
+		 * iterate in reverse to process last client first.
 		 */
+		wl_list_for_each_reverse(node, &tree->children, link) {
+			struct lab_layer_surface *layer = node_layer_surface_from_node(node);
+			struct wlr_scene_layer_surface_v1 *scene = layer->scene_layer_surface;
+			struct wlr_layer_surface_v1 *layer_surface = scene->layer_surface;
```
Consolatis, line 152: Is it ensured that there can be no other nodes in that tree and that all `lab_layer_surfaces` do in fact have a `scene_layer_surface` at all times and that in turn also has a `layer_surface`? Otherwise we may want to check for that first and `continue`.

src/layers.c
```diff
@@ -118,45 +118,116 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
+focused_layer_has_exclusive_interactivity(struct seat *seat)
+{
+	if (!seat->focused_layer) {
+		return false;
+	}
+	if (seat->focused_layer->current.keyboard_interactive !=
+			ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE) {
+		return false;
+	}
+	return true;
```
Consolatis, line 187: Suggest
```c
return seat->focused_layer->current.keyboard_interactive ==
	ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_EXCLUSIVE
```

Consolatis:
The logic feels right, didn't test yet. Only thing missing (or I didn't catch it) is that we don't focus layershell surfaces on the background layer with keyboard focus (e.g. pcmanfm --desktop and friends)  if `desktop_focus_topmost_view()` doesn't find any views. I think we could either add it there or in the `seat_focus()` `if (!surface)` branch.

Overall this PR clean things up very nicely!

src/layers.c
johanmalm: I'm fairly comfortable (if there is ever such a thing!)

> Is it ensured that there can be no other nodes in that tree

I think so. And in `node_layer_surface_from_node()` we assert() node-descriptor type.

> and that all `lab_layer_surfaces` do in fact have a `scene_layer_surface` at all times and that in turn also has a `layer_surface`? Otherwise we may want to check for that first and `continue`.

This is consistent with the rest of `src/layers.c`. The `wlr_layer_surface_v1` is the primary object. The other two are just wrapper structs.

src/layers.c
johanmalm: Yes, that's better. Done.

src/layers.c
johanmalm: Done.

johanmalm:
> The logic feels right, didn't test yet. Only thing missing (or I didn't catch it) is that we don't focus layershell surfaces on the bottom / background layer with keyboard interactivity (e.g. `pcmanfm --desktop` and friends) if `desktop_focus_topmost_view()` doesn't find any views. I think we could either add it there or in the `seat_focus() if (!surface)` branch.

I started a separate PR for that. https://github.com/labwc/labwc/pull/1598/files

johanmalm:
Thanks for both for reviewing. 
Ready for final review from my perspective (following small number of fixups)

src/layers.c
Consolatis: more eyes -> more spotted :)

src/layers.c
```diff
@@ -184,7 +252,8 @@ handle_surface_commit(struct wl_listener *listener, void *data)
 	}
 	/* Process keyboard-interactivity change */
 	if (committed & WLR_LAYER_SURFACE_V1_STATE_KEYBOARD_INTERACTIVITY) {
-		process_keyboard_interactivity(layer);
+		struct seat *seat = &layer->server->seat;
+		layer_try_set_focus(seat, layer_surface);
 	}
```
Consolatis, line 257: Just trying to think this through, if a layershell client was `INTERACTIVITY_NONE` before and then sets it to `ON_DEMAND`, we would steal focus from an active window or other layershell surface. That feels wrong.
It might be a different story for the initial mapping but this could be triggered anytime during the lifetime of the surface.

johanmalm: Yes, I see what you mean. Hadn't thought of that.

So, `ON_DEMAND` should not do anything at the time of commit, only when receiving a cursor-button-press.

Let me patch and test with my local clients.

johanmalm: Have patched && tested with `sfwbar`, `bemenu`, `tint` and `idesk-morfa`. I think this is better.

Consolatis: Looks good. Sorry but I want to get all the details correct :p

What about the surface was `EXCLUSIVE`, had focus and then changed to `ON_DEMAND`?
I think that we have to let that one through as well, I am not sure if we can do it that easily though.. e.g. where to get the information that it was previously a exclusive surface.. Maybe we could just check for `ON_DEMAND` and compare it to the focused surface and if they match then call `layer_try_set_focus()`? Didn't think this through completely though.

johanmalm: Good point - again - and as usual :smile:
Have tweaked patch.

Consolatis: Can't spot anything else. Nice :)

Consolatis:
LGTM, not tested.

I'll leave the honors to you.

src/layers.c
```diff
@@ -184,8 +259,28 @@ handle_surface_commit(struct wl_listener *listener, void *data)
...
+			if (seat->focused_layer == layer_surface) {
+				/*
+				 * Must be change from EXCLUSIVE to ON_DEMAND,
+				 * so we should give us focus.
+				 */
+				struct server *server = layer->server;
+				try_to_focus_next_layer_or_toplevel(server);
+			}
+			goto out;
+		}
```
Consolatis, line 278: I don't remember why we try to move the focus somewhere else on change from `EXCLUSIVE` to `ON_DEMAND` for the currently focused layer surface. This seems to prevent something like the following client flow: `EXCLUSIVE` -> `map` -> `ON_DEMAND` which ensures that the surface gets focus after map but doesn't prevent the user from giving focus to some other surface after that. Didn't test though.

Context: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/issues/141

CC @johanmalm

johanmalm: I think the context is this:
1. `on-demand` means give keyboard focus when clicked on
2. Some panels set `exclusive` when opening a menu as a result on IPC

We want to cope with this workflow
a. Open panel menu via IPC (set `exclusive`)
b. Press escape to close menu (set `on-demand`)
c. Return focus on topmost window

Consolatis: Hm.. that workflow is incompatible to `exclusive` -> `map` -> `on-demand` though.

If a panel does the above workflow to get keyboard focus for their menus (via inheritance from the panel surface) they should likely set to `none` at the end, that would then return the focus as intended. And if the panel has additional text inputs or similar it can then set it to `on-demand` again.

johanmalm: The layer-shell protocol refers to "an implementation-defined manner".

I don't mind changing it. Suggest:

1. Documenting this. Possibly a `labwc-layer-shell(5)`. I did start one, but haven't finished it :smile:
2. Try to get panel maintainer buy-in and confirmation that we're not breaking user-space (xfce4-panel and wf-panel-pi IIRC)

PR #1600, opened 2024-03-09
session: process environment.d and allow empty variables

ahesford:
1. All '*.env' files in an 'environment.d' directory alongside each potential 'environment' file will be parsed and added to the environment.
2. For the purposes of configuration merging, an environment definition exists at one level if either the 'environment' file is defined or its corresponding 'environment.d' contains any valid '*.env' file.
3. Variable declarations of the form "VARIABLE=", with no following value, will be written to the environment as empty strings.

Quick tests show this working, but it hasn't been rigorously tested, and the merge logic needs testing as well.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -37,13 +37,23 @@ option.
 All configuration and theme files except autostart and shutdown are re-loaded on
 receiving signal SIGHUP.
 
-The *environment* file is parsed as *variable=value* and sets environment
-variables accordingly. It is recommended to specify keyboard layout settings and
-cursor size/theme here; see environment variable section below for details. Note
-that the environment file is treated differently by openbox where it is simply
-sourced prior to running openbox.
-Note: Tilde (~) and environment variables in the value are expanded, but
-subshell syntax and apostrophes are ignored.
+Environment variables may be set within *environment* files, wherein each line
+defines shell variables in the format *variable=value*. It is recommended to
+specify keyboard layout settings and cursor size/theme here; see environment
+variable section below for details. Within an XDG Base Directory, a file named
+"environment" will be parsed first, followed by any file matching the glob
+"environment.d/\*.env". Files within the environment.d directory are parsed in
+an arbitrary order; any variables that must be set in a particular sequence
+should be set within the same file. Unless the --merge-config option is
+specified, labwc will consider a particular XDG Base Directory to have provided
+an environment file if that directory contains either the "environment"
+directory or at least one "environment.d/\*.env" file.
+
+Note: environment files are treated differently by Openbox, which will simply
+source the file as a valid shell script before running the window manager. Files
+are instead parsed directly by labwc, although any environment variables
+referenced as $VARIABLE or ${VARIABLE} will be substituted and the tilde (~)
+will be expanded as the user's home directory.
 
 The *autostart* file is executed as a shell script after labwc has read its
 configuration and set variables defined in the environment file. Additionally,
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -1,11 +1,13 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #define _POSIX_C_SOURCE 200809L
 #include <assert.h>
+#include <dirent.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #include <wlr/backend/drm.h>
 #include <wlr/backend/multi.h>
 #include <wlr/util/log.h>
@@ -49,9 +51,10 @@ process_line(char *line)
 	buf_add(&value, string_strip(++p));
 	buf_expand_shell_variables(&value);
 	buf_expand_tilde(&value);
-	if (string_null_or_empty(key) || !value.len) {
+	if (string_null_or_empty(key)) {
 		goto error;
 	}
+
 	setenv(key, value.buf, 1);
 error:
 	free(value.buf);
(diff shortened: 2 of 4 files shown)


src/config/session.c
```diff
@@ -176,7 +252,15 @@ session_environment_init(void)
...
 	for (struct wl_list *elm = iter(&paths); elm != &paths; elm = iter(elm)) {
 		struct path *path = wl_container_of(elm, path, link);
+
+		/* Process an environment file itself */
 		bool success = read_environment_file(path->string);
+
+		/* Process a correponding environment.d directory */
+		if (read_environment_dir(path->string)) {
+			success = true;
+		}
```
Consolatis, line 262: Nitpick only
```c
success |= read_environment_dir(path->string);
```

src/config/session.c
```diff
@@ -80,6 +83,79 @@ read_environment_file(const char *filename)
...
+static bool
+read_environment_dir(const char *path_prefix)
+{
+	bool success = false;
+
+	char *path = strdup_printf("%s.d", path_prefix);
+	if (!path) {
+		wlr_log(WLR_ERROR, "failed to allocate memory for environment path");
+		return false;
+	}
```
Consolatis, line 123: Not needed, we `xzalloc` the string in `strdup_printf` and thus just die if the allocation fails.

Consolatis (commented):
2 minor nitpicks, not tested.

src/config/session.c
```diff
@@ -80,6 +83,79 @@ read_environment_file(const char *filename)
...
+	char *full_path = strdup_printf("%s/%s", prefix, name);
+	if (!full_path) {
+		return NULL;
+	}
+
+	/* Valid environment file must be a regular file */
+	struct stat statbuf;
+	if (stat(full_path, &statbuf) == 0 && S_ISREG(statbuf.st_mode)) {
+		return full_path;
+	}
```
Consolatis, line 108: Is a symlink counting as regular file as well? If not we may also want to allow them here.

src/config/session.c
ahesford: Good point; removed.

src/config/session.c
ahesford: In my testing, a broken symlink caused no problems, and a valid symlink was followed as expected.

src/config/session.c
ahesford: I don't think this is done anywhere else for Boolean overrides, and I usually like to avoid using bitwise operators when treating Booleans as opaque types.

Consolatis: We do that for various `consumed_by_frame_context |=` operations in `src/input/cursor.c` but again, just a nitpick. Totally fine with keeping the current variant.

ahesford: Somehow I missed those cases. Consistency seems preferable here, so I included the suggestion.

Consolatis:
Code LGTM, not tested.

src/config/session.c
```diff
@@ -80,6 +83,74 @@ read_environment_file(const char *filename)
...
+				"failed to read environment directory: %s",
+				err_msg ? err_msg : "reason unknown");
+		}
+
+		goto env_dir_cleanup;
+	}
+
+	struct dirent *dirent;
+	while ((dirent = readdir(envdir)) != NULL) {
+		char *env_file_path = valid_env_path(path, dirent->d_name);
```
johanmalm, line 136: 1. Consider `s/valid_env_path/path_validate_and_dup/` (just to help the reader understand the allocation). Just a thought though.
2. I'm happy to do `if (dirent->d_type != DT_REG) { continue; }`. It avoids the `stat()` later. Think we have to define `_DEFAULT_SOURCE`, but it's supported "everywhere". It is my preferred approach generally, but happy to go with your preference.
3. Consider supporting symlinks too. Either via `lstat()` or by adding `&& dirent->d_name != DT_LNK`

src/config/session.c
```diff
@@ -80,6 +83,74 @@ read_environment_file(const char *filename)
 	return true;
 }
 
+static char *
+valid_env_path(const char *prefix, const char *name)
```
johanmalm, line 87: Consider splitting this into `has_valid_file_extension()` and `path_join()`

johanmalm (commented):
LGTM

Have made a few suggestions on style + fuller use of `dirent` rather than `stat()`

johanmalm:
Have tested. Seems to work well.
Nice!

src/config/session.c
ahesford: I renamed the function `strdup_env_path_validate` to make the allocation more obvious and to be more descriptive about the particular paths being validated. I've also added some logic to check `dirent->d_type` to confirm that the path actually refers to a regular file (or a symlink to a regular file).

The `dirent->d_type` field will contain `DT_LNK` for symlinks whether or not they are valid, but it may also contain `DT_UNKNOWN` on some filesystems (XFS seems like one of them) that may opt not to tell you the type of the file when reading a directory. We can short-circuit all of the path validity checks if the directory entry is any other type. After the path validation, we can use `dirent->d_type == DT_REG` to short-circuit the `stat` call, but we sill have to fall back to using `stat` for links or unknown types. The `stat` call follows symbolic links, so valid symlinks will just work with this implementaiton.

src/config/session.c
ahesford: I moved the extension check to a more general string helper called `str_endswith`. This also behaves with NULL or empty arguments, so we can eliminate some up-front checks in this function.

ahesford:
Hrm, not sure how to expose `DT_*` macros in FreeBSD. Advice appreciated, or I can just drop the shortcuts.

Consolatis:
> Hrm, not sure how to expose `DT_*` macros in FreeBSD. Advice appreciated, or I can just drop the shortcuts.

@jbeich suggestions?

jbeich:
Remove `_POSIX_C_SOURCE` from the file if you need non-POSIX bits. `_DEFAULT_SOURCE` is a Linux-specific cheat knob. On BSDs non-POSIX extensions are enabled by default and the cheat knob is different (`__BSD_VISIBLE` on DragonFly/FreeBSD/OpenBSD and `_NETBSD_SOURCE` on NetBSD).

Consolatis:
> Remove `_POSIX_C_SOURCE` from the file if you need non-POSIX bits. `_DEFAULT_SOURCE` is a Linux-specific cheat knob. On BSDs non-POSIX extensions are enabled by default and the cheat knob is different (`__BSD_VISIBLE` on DragonFly/FreeBSD/OpenBSD and `_NETBSD_SOURCE` on NetBSD).

Thanks, if that works in this case: great. Otherwise the easiest and most portable solution is likely to just skip the optimization and go with the separate stat call instead. This is not a performance critical path anyway.

ahesford:
With `-Werror` in CI, the FreeBSD build fails because of the unused `_DEFAULT_SOURCE` definition. This could be gated behind a check for `__linux__`, but operating system checks in this file seem kind of ugly to save the `stat`. I dropped the `dt_type` optimization instead.

johanmalm:
Sorry about all this. I think my comment regarding `d_type` was poor. Need to re-evaluate my ways. Looks like I got confused about `stat()` and `lstat()` too. Should look at man pages before I comment!!

Agree with your sentiment. Let's go with `stat` and POSIX.

LGTM.

Ready for merge?

ahesford:
Let's wait. I'm happy with the implementation, but want to move the `ENVIRONMENT VARIABLES` section from `labwc-config.5` and add some proper documentation for environment parsing that includes the new directory scan.

ahesford:
I added some documentation about `environment.d`; if there are no edits, this is ready to merge.

johanmalm:
LGTM. Thanks :smile:

docs/labwc-config.5.scd
```diff
@@ -37,13 +37,23 @@ option.
...
+defines shell variables in the format *variable=value*. It is recommended to
+specify keyboard layout settings and cursor size/theme here; see environment
+variable section below for details. Within an XDG Base Directory, a file named
+"environment" will be parsed first, followed by any file matching the glob
+"environment.d/\*.env". Files within the environment.d directory are parsed in
+an arbitrary order; any variables that must be set in a particular sequence
+should be set within the same file. Unless the --merge-config option is
+specified, labwc will consider a particular XDG Base Directory to have provided
+an environment file if that directory contains either the "environment"
+directory or at least one "environment.d/\*.env" file.
```
Consolatis, line 50: `[..] "environment" directory or at least one "environment.d/*.env" file"`

First one should likely be `"environment" file`?

johanmalm: Good spot. Sorry missed that one!

ahesford: Fixed; thanks!

Issue #1595, opened 2024-03-07
[Feature request] Add " --exit-with-session" option to labwc.


stefonarch:
`kwin_wayland` has this nice option which allows running a full LXQt session and kwin will exit too after the session is closed. Also when only running  a terminal  `kwin_wayland qterminal` it will exit when the terminal is closed. Labwc provides already the "-s" option to start any command.

Background: The problem is that `lxqt-session` can't close or start labwc, an external bash script or application is needed to close both.

ahesford:
The notion of a "session" not defined by the lifetime of the compositor itself seems like a nebulous concept. This works for X11 window managers because the session is defined by the lifetime of the X server, and the window manager will die automatically when its server terminates. 

An argument could be made that, when launching labwc with a single client (as in your `qterminal` example), the compositor should die when the client dies. That's certainly how cage behaves. However, cage is generally built on a "single instance" interaction model, while labwc is designed around simultaneous interactions with many clients and even provides a means to start more that aren't children of the "key" client. Nominating one as special because it was started at launch would be a bit kludgey and would probably feel unintuitive.

Defining an external "session" concept and making labwc tie its fate to that session seems like it would make labwc particularly opinionated in undesirable ways. Unless there's a Wayland protocol for sending session events that we can implement, this starts to wade into the same kind of territory occupied by things like nonstandard IPC; this project expressly avoids that realm.

My view is that this is beyond the scope of labwc. LXQt should implement a helper, launchable via the autostart script, that can monitor whatever defines a session and send SIGTERM to `$LABWC_PID` at the appropriate time.

johanmalm:
@stefonarch 

I had a quick look at `kwin`, but don't follow the logic. Could you help us understand the mechanism you're trying to achieve.

Let me describe how I understand it, so that you can educate me and help the jigsaw pieces fall into place.

If we take the simplest case (according to my understanding) a session is a script that sets up an environment and then launches `labwc` (sometimes using exec so that it is not a separate PID). So if the session closes, in my mind `labwc` goes with it.

Could you describe what starts what - and then how you wish it to close.

stefonarch:
Under X11 `startlxqt` is a script that sets mainly env vars  and launches `lxqt-session` which isn't a script.

Under wayland I'm using a similar script `_startlxqtlabwc_` which at the end starts `labwc -s lxqt-session`.

[...]

PR #1601, opened 2024-03-09
Add `-S|--session` command line option

johanmalm:
See commit message.

Ref #1595

- [x] Documentation
- [ ] Testing

Cc: @stefonarch

### DIFF
--- a/docs/labwc.1.scd
+++ b/docs/labwc.1.scd
@@ -56,6 +56,19 @@ the `--exit` and `--reconfigure` options use.
 *-s, --startup* <command>
 	Run command on startup
 
+*-S, --session* <command>
+	Run command on startup and terminate compositor on exit. This is useful
+	for session management as it allows the session client to terminate
+	labwc by exiting itself. This is a Wayland specific use-case because
+	under X, xinit starts the server and keeps it alive for as long as the
+	session client. Thus either the session client starts the Window
+	Manager, or the Window Manager can be launched independently first. On
+	Wayland, the Compositor is both Display Server and Window Manager, so
+	the described session management mechanisms do not work because the
+	Compositor needs to be running before the session can function.  As some
+	session clients support both X11 and Wayland, this command line option
+	avoids re-writes and fragmentation.
+
 *-v, --version*
 	Show the version number and quit
 
--- a/include/common/spawn.h
+++ b/include/common/spawn.h
@@ -5,6 +5,12 @@
 #include <sys/types.h>
 
 /**
+ * spawn_primary_client - execute asynchronously
+ * @command: command to be executed
+ */
+pid_t spawn_primary_client(const char *command);
+
+/**
  * spawn_async_no_shell - execute asynchronously
  * @command: command to be executed
  */
--- a/src/common/spawn.c
+++ b/src/common/spawn.c
@@ -25,11 +25,22 @@ reset_signals_and_limits(void)
 	signal(SIGPIPE, SIG_DFL);
 }
 
-static void
+static bool
 set_cloexec(int fd)
 {
 	int flags = fcntl(fd, F_GETFD);
-	fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
+	if (flags == -1) {
+		wlr_log_errno(WLR_ERROR,
+			"Unable to set the CLOEXEC flag: fnctl failed");
+		return false;
+	}
+	flags = flags | FD_CLOEXEC;
+	if (fcntl(fd, F_SETFD, flags) == -1) {
+		wlr_log_errno(WLR_ERROR,
+			"Unable to set the CLOEXEC flag: fnctl failed");
+		return false;
+	}
+	return true;
 }
 
 void
@@ -77,6 +88,41 @@ spawn_async_no_shell(char const *command)
 	waitpid(child, NULL, 0);
 out:
 	g_strfreev(argv);
+}
+
+pid_t
+spawn_primary_client(const char *command)
+{
+	assert(command);
+
+	GError *err = NULL;
+	gchar **argv = NULL;
+
+	/* Use glib's shell-parse to mimic Openbox's behaviour */
+	g_shell_parse_argv((gchar *)command, NULL, &argv, &err);
+	if (err) {
+		g_message("%s", err->message);
+		g_error_free(err);
+		return -1;
+	}
+
+	pid_t child = fork();
+	switch (child) {
+	case -1:
+		wlr_log_errno(WLR_ERROR, "Failed to fork");
+		g_strfreev(argv);
+		return -1;
+	case 0:
+		/* child */
+		close(STDIN_FILENO);
+		reset_signals_and_limits();
+		execvp(argv[0], argv);
+		wlr_log_errno(WLR_ERROR, "Failed to execute primary client %s", command);
+		_exit(1);
+	default:
+		g_strfreev(argv);
+		return child;
+	}
 }
 
 pid_t
--- a/src/server.c
+++ b/src/server.c
@@ -90,6 +90,7 @@ handle_sigchld(int signal, void *data)
 {
 	siginfo_t info;
 	info.si_pid = 0;
+	struct server *server = data;
 
 	/* First call waitid() with NOWAIT which doesn't consume the zombie */
 	if (waitid(P_ALL, /*id*/ 0, &info, WEXITED | WNOHANG | WNOWAIT) == -1) {
@@ -102,8 +103,7 @@ handle_sigchld(int signal, void *data)
 	}
 
 #if HAVE_XWAYLAND
-	/* Verify that we do not break xwayland lazy initialization */
-	struct server *server = data;
+	/* Ensure that we do not break xwayland lazy initialization */
 	if (server->xwayland && server->xwayland->server
 			&& info.si_pid == server->xwayland->server->pid) {
 		return 0;
@@ -137,6 +137,11 @@ handle_sigchld(int signal, void *data)
 		wlr_log(WLR_ERROR,
 			"spawned child %ld terminated unexpectedly: %d"
 			" please report", (long)info.si_pid, info.si_code);
+	}
+
+	if (info.si_pid == server->primary_client_pid) {
+		wlr_log(WLR_INFO, "primary client %ld exited", (long)info.si_pid);
+		wl_display_terminate(server->wl_display);
 	}
 
 	return 0;
@@ -281,6 +286,7 @@ static const char helpful_seat_error_message[] =
 void
 server_init(struct server *server)
 {
+	server->primary_client_pid = -1;
 	server->wl_display = wl_display_create();
 	if (!server->wl_display) {
 		wlr_log(WLR_ERROR, "cannot allocate a wayland display");
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -349,6 +349,8 @@ struct server {
 
 	struct menu *menu_current;
 	struct wl_list menus;
+
+	pid_t primary_client_pid;
 };
 
 #define LAB_NR_LAYERS (4)
--- a/src/main.c
+++ b/src/main.c
@@ -24,6 +24,7 @@ static const struct option long_options[] = {
 	{"merge-config", no_argument, NULL, 'm'},
 	{"reconfigure", no_argument, NULL, 'r'},
 	{"startup", required_argument, NULL, 's'},
+	{"session", required_argument, NULL, 'S'},
 	{"version", no_argument, NULL, 'v'},
 	{"verbose", no_argument, NULL, 'V'},
 	{0, 0, 0, 0}
@@ -39,6 +40,7 @@ static const char labwc_usage[] =
 "  -m, --merge-config       Merge user config files/theme in all XDG Base Dirs\n"
 "  -r, --reconfigure        Reload the compositor configuration\n"
 "  -s, --startup <command>  Run command on startup\n"
+"  -S, --session <command>  Run command on startup and terminate on exit\n"
 "  -v, --version            Show version number and quit\n"
 "  -V, --verbose            Enable more verbose logging\n";
 
@@ -87,12 +89,13 @@ main(int argc, char *argv[])
 	textdomain(GETTEXT_PACKAGE);
 #endif
 	char *startup_cmd = NULL;
+	char *primary_client = NULL;
 	enum wlr_log_importance verbosity = WLR_ERROR;
 
 	int c;
 	while (1) {
 		int index = 0;
-		c = getopt_long(argc, argv, "c:C:dehmrs:vV", long_options, &index);
+		c = getopt_long(argc, argv, "c:C:dehmrs:S:vV", long_options, &index);
 		if (c == -1) {
 			break;
 		}
@@ -117,6 +120,9 @@ main(int argc, char *argv[])
 			exit(0);
 		case 's':
 			startup_cmd = optarg;
+			break;
+		case 'S':
+			primary_client = optarg;
 			break;
 		case 'v':
 			printf("labwc " LABWC_VERSION "\n");
@@ -171,6 +177,16 @@ main(int argc, char *argv[])
 
 	menu_init(&server);
 
+	/* Start session-manager if one is specified by -S|--session */
+	if (primary_client) {
+		server.primary_client_pid = spawn_primary_client(primary_client);
+		if (server.primary_client_pid < 0) {
+			wlr_log(WLR_ERROR, "fatal error starting primary client: %s",
+				primary_client);
+			goto out;
+		}
+	}
+
 	session_autostart_init(&server);
 	if (startup_cmd) {
 		spawn_async_no_shell(startup_cmd);
@@ -178,6 +194,7 @@ main(int argc, char *argv[])
 
 	wl_display_run(server.wl_display);
 
+out:
 	session_shutdown(&server);
 
 	server_finish(&server);


ahesford:
I think the concept is good, but "primary-client" might be a bit opaque. What about `-S` and `--session-manager` or `--session-startup`, which might be a bit more intuitive and has the added benefit that it mirrors the `-s` short option. Then people can remember "s" for startup, and capitalizing the argument makes it "important".

stefonarch:
Not sure if I miss something, but
```
labwc --primary-client qterminal
Usage: labwc [options...]
  -c, --config <file>      Specify config file (with path)
  -C, --config-dir <dir>   Specify config directory
```
but thanks, this would solve a quite big issue :)

johanmalm:
> Not sure if I miss something

Sorry, I messed up. Just try with `labwc -p qterminal` for now.

johanmalm:
> I think the concept is good, but "primary-client" might be a bit opaque. What about `-S` and `--session-manager` or `--session-startup`, which might be a bit more intuitive and has the added benefit that it mirrors the `-s` short option. Then people can remember "s" for startup, and capitalizing the argument makes it "important".

Oh, yes, I like that. `S` is great. How about simply `--session` for the long option?

ahesford:
`--session` seems fine to me.

stefonarch:
Preliminary test results:
* just one process is fine `labwc -p qterminal` → `exit` → labwc exits.
* the same, opening `lximage-qt` and lowering it, `exit` in qterminal and labwc doesn't exit.
* with `-p lxqt-session` in the start script it doesn't exit labwc when lxqt-session exits
* with a script `start-lxqt` which launches just panel and desktop opening a terminal and killing both labwc doesn't exit

johanmalm:
I've changed the command line option to `-S|--session` and added documentation.

Also, description added in `labwc.1.scd`

It works here... if I do the following:
1. `labwc -S qterminal` (running nested)
2. Opening `lximage-qt` in the `qterminal`
3. Minimize `lximage-qt`
4. Exit `qterminal` by pressing close-button in top-right corner
5. Observe that `labwc` terminates

johanmalm:
I've just successfully tested the following steps (nested and DRM):

- `labwc -S lxqt-session`
- Open a terminal
- Run `lxqt-leave`
- Click logout
- Observe that compositor terminates.

docs/labwc.1.scd
```diff
@@ -56,6 +56,19 @@ the `--exit` and `--reconfigure` options use.
 *-s, --startup* <command>
 	Run command on startup
 
+*-S, --session* <command>
+	Run command on startup and terminate compositor on exit. This is useful
+	for session management as it allows the session client to terminate
+	labwc be exiting itself. This is a Wayland specific use-case because
```
Consolatis, line 62: `by existing`

include/common/spawn-primary-client.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_SPAWN_PRIMARY_CLIENT_H
+#define LABWC_SPAWN_PRIMARY_CLIENT_H
```
Consolatis, line 3: Why the own file and include rather than `common/spawn.{h,c}`? Because of the license?

stefonarch:
I still can't make it work:

* with only `labw -S qterminal` → `lximage-qt & ` → exit terminal labwc doesn't exit
* same with full session 

```
12222:mar 10 08:42:30 archlinux labwc[71328]: lxqt-session: Seding TERM to child  71845
12223:mar 10 08:42:30 archlinux labwc[71328]: lxqt-session: Seding TERM to child  72846
12224-mar 10 08:42:30 archlinux labwc[71328]: lxqt-session: Child process  71406  terminated on signal  15
12225-mar 10 08:42:30 archlinux labwc[71328]: lxqt-session: Child process  71809  terminated on signal  15
12226-mar 10 08:42:30 archlinux labwc[71328]: lxqt-session: Child process  72846  exited with status  0
12227-mar 10 08:42:30 archlinux labwc[71328]: lxqt-session: Child process  71404  exited with status  0
12228-mar 10 08:42:30 archlinux labwc[71262]: 00:00:53.688 [DEBUG] [xwayland/xwm.c:1626] unhandled X11 event: FocusOut (10)
12229-mar 10 08:42:30 archlinux labwc[71262]: 00:00:53.689 [ERROR] [xwayland/xwm.c:1599] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 323, value 6291459
12230-mar 10 08:42:30 archlinux labwc[71262]: 00:00:53.689 [ERROR] [xwayland/xwm.c:1599] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 325, value 6291459
12231-mar 10 08:42:30 archlinux labwc[71262]: 00:00:53.689 [ERROR] [xwayland/xwm.c:1599] xcb error: op DeleteProperty (no minor), code Window (no extension), sequence 326, value 6291459
12232-mar 10 08:42:30 archlinux labwc[71941]: Exiting due to channel error.
12233-mar 10 08:42:31 archlinux labwc[71802]: Exiting due to channel error.
12234-mar 10 08:42:31 archlinux labwc[72533]: Exiting due to channel error.
12235-mar 10 08:42:31 archlinux labwc[71961]: Exiting due to channel error.
12236-mar 10 08:42:31 archlinux labwc[72112]: Exiting due to channel error.
12237-mar 10 08:42:31 archlinux labwc[71946]: Exiting due to channel error.
12238-mar 10 08:42:31 archlinux labwc[72473]: Exiting due to channel error.
12239-mar 10 08:42:31 archlinux labwc[71262]: 00:00:53.836 [DEBUG] [types/wlr_compositor.c:692] New wlr_surface 0x578941866640 (res 0x578941898940)
12240-mar 10 08:42:31 archlinux labwc[71262]: 00:00:53.837 [DEBUG] [types/xdg_shell/wlr_xdg_surface.c:388] new xdg_surface 0x57894183ae30 (res 0x578941860210)
12241-mar 10 08:42:31 archlinux labwc[71262]: 00:00:53.838 [DEBUG] [types/wlr_xdg_decoration_v1.c:219] new xdg_toplevel_decoration 0x57894160edc0 (res 0x57894170c060)
12242-mar 10 08:42:31 archlinux labwc[71262]: 00:00:53.873 [DEBUG] [../src/view-impl-common.c:49] [map] identifier=com.github.hluk.copyq, title=CopyQ
12243-mar 10 08:42:31 archlinux labwc[71262]: 00:00:53.956 [DEBUG] [types/wlr_compositor.c:692] New wlr_surface 0x578941865120 (res 0x57894185e750)
12244-mar 10 08:42:41 archlinux labwc[71262]: 00:01:04.023 [INFO] [xwayland/server.c:215] Restarting Xwayland (lazy)
12245-mar 10 08:42:43 archlinux labwc[71262]: 00:01:06.302 [DEBUG] [../src/action.c:679] Handling action 14: ShowMenu
12246-mar 10 08:42:46 archlinux labwc[71262]: 00:01:08.763 [DEBUG] [../src/action.c:679] Handling action 6: Exit
12247-mar 10 08:42:46 archlinux labwc[71262]: 00:01:08.764 [INFO] [../src/config/session.c:131] Updating dbus execution environment
12248-mar 10 08:42:46 archlinux labwc[71815]: wl_display_dispatch: Broken pipe

```

stefonarch:
Ok, if I remove/rename `~/.config/autostart`  and my session starts only with swaybg and lxqt-modules it works.

docs/labwc.1.scd
johanmalm: Thanks. Done.

include/common/spawn-primary-client.h
johanmalm: Because:

1. The license & copyright notice.
2. It's quite a specialist thing with a couple helper functions, so nice to keep them isolated/separated.
3. Makes it clear where those bits have come from for reference for us later (or for future devs) in case the cage approach develops of we want to make changes to it.

johanmalm:
Just demonstrating it on my setup. `lxqt-session` does not issue `labwc -e`, it just closes itself.

[recording.webm](https://github.com/labwc/labwc/assets/6955353/bad72183-cdb6-447a-8d00-c4761e696520)

johanmalm:
@Tamaranch  @spl237 Any thoughts from either of you. Would this be a useful feature in conjunction with your respective projects?

spl237:
I don't think it is useful for us, at least not in the way we have implemented wayland sessions at present. We use lightdm to start the wayland sessions directly; we use lxsession for X sessions, but we don't use a session manager for wayland, as lightdm does everything we need.

That's not to say it might not be useful in the future!

stefonarch:
@johanmalm It works here 
* with `lxqt-session` as long there are no applications configured  for autostart (in 
`lxqt-config-session`  → autostart) = `~/.config/autostart/` empty
* as long no command with & is issued in the terminal using   `labwc -S qterminal`

Tamaranch:
Eh, why not. As it happens, I've just merged a [small port to Wayland](https://gitlab.xfce.org/xfce/xfce4-session/-/merge_requests/49) of xfce4-session, where I use labwc as the default compositor, and more specifically the command `labwc --startup xfce4-session`. So this option could be used instead.

That said, until we develop our own compositor (which probably won't be until Xfce 4.22, i.e. not until next year), and even when we do I think, we don't want to impose any particular compositor. So we need something a little generic to leave the session.

For the moment [I've put this in](https://gitlab.xfce.org/xfce/xfce4-session/-/commit/8e623bc442d44f573c7cb4ea69304dfb6f59b5d7), but it can be considered temporary. As discussed in https://github.com/labwc/labwc/issues/1595, the notion of session on Wayland is a bit nebulous at the moment. Perhaps merging the [session-management protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/18) would help clarify things.

stefonarch:
Playing whith my test user I see that it depends on _what_ is set in autostart: I added qterminal, firefox and feathernotes and it exits still fine, so there is some blocking app under my autostart. Nice!

Consolatis:
> Hidden eggs?
> 
> ```
> $ labwc -p qterminal
> 00:00:00.001 [../src/action.c:430] Invalid action: showdesktop       
> 00:00:00.002 [../src/action.c:430] Invalid action: ToggleAlwaysTop
> ```

No, those must be defined in your `rc.xml`.

johanmalm:
> > `00:00:00.001 [../src/action.c:430] Invalid action: showdesktop`

There is an old PR for this, but not implemented yet.

> > `00:00:00.002 [../src/action.c:430] Invalid action: ToggleAlwaysTop`

Missing `On`?

johanmalm:
> For the moment [I've put this in](https://gitlab.xfce.org/xfce/xfce4-session/-/commit/8e623bc442d44f573c7cb4ea69304dfb6f59b5d7), but it can be considered temporary. As discussed in #1595, the notion of session on Wayland is a bit nebulous at the moment. Perhaps merging the [session-management protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/18) would help clarify things.

Thanks for your thoughts and for the the link to that protocol. I wasn't aware of that one. Interesting indeed.
The session living through compositor re-starts would be interesting :smile:
Maybe that points to a two-part session manager with some IPC.
I've not really thought about Compositor re-starts - but it feels pretty deep. Had some discussions on IRC the other day, but I couldn't quite tune in to it.

stefonarch:
>No, those must be defined in your rc.xml.

In fact, I use a different location usually, was on old file, `showdesktop` was a script which I removed.

> The session living through compositor re-starts would be interesting 😄

kwin_wayland with Qt  have this feature, but it works only in plasma, have tested it killing kwin, it restores  quite everything.

stefonarch:
I cleaned up my autostart folder and now it works like a charm.

src/common/spawn-primary-client.c
```diff
@@ -0,0 +1,119 @@
...
+		return false;
+	}
+
+	pid_t pid = fork();
+	if (!pid) {
+		sigset_t set;
+		sigemptyset(&set);
+		sigprocmask(SIG_SETMASK, &set, NULL);
+		close(fd[0]);
+		execvp(argv[0], argv);
```
Consolatis, line 97: I think this misses some other stuff, namely:
- `restore_nofile_limit()`
- `signal(SIGPIPE, SIG_DFL);`

See `src/common/spawn.c`. IMHO we should try to make this a reusable function so we don't accidentally end up adding something that is only reset at one place but not the other. Also: do we need / want `setsid()` ?

Consolatis:
> I cleaned up my autostart folder and now it works like a charm.

Should we try to reproduce the issue and figure out why it wasn't working before?
E.g. do we actually trigger `wl_display_terminate()` or not in this case? It seems we do not which is strange.

stefonarch:
> > I cleaned up my autostart folder and now it works like a charm.
> 
> Should we try to reproduce the issue and figure out why it wasn't working before? E.g. do we actually trigger `wl_display_terminate()` or not in this case? It seems we do not which is strange.

I _guess_ it was because I had modified lxqt-modules (needed time ago for hyprland) in autostart. I put everything back except them as they are not needed anymore. I could load the backup  and retest, removing one by one.

stefonarch:
> The session living through compositor re-starts would be interesting 😄

Found it again, nested session:

`labwc -C /tmp -s lxqt-panel`
closing the window:
`qt.qpa.wayland: Creating a fake screen in order for Qt not to crash`

stefonarch:
@johanmalm I've added in the coming version of `lxqt-labwc-session` a version check:
https://github.com/stefonarch/lxqt-labwc-session/blob/c4b1d71c83eba459462f0c1d0f824c5f71968e25/startlxqtlabwc.in#L99
 Any hope to have this feature in v7.2 ?

johanmalm:
> @johanmalm I've added in the coming version of `lxqt-labwc-session` a version check: https://github.com/stefonarch/lxqt-labwc-session/blob/c4b1d71c83eba459462f0c1d0f824c5f71968e25/startlxqtlabwc.in#L99 Any hope to have this feature in v7.2 ?

I'll take a look tomorrow to see if I can respond to outstanding review comments. Should be fine for `0.7.2`

johanmalm:
> I think this misses some other stuff, namely:
> 
>     * `restore_nofile_limit()`
>     * `signal(SIGPIPE, SIG_DFL);`
> 
> See `src/common/spawn.c`. IMHO we should try to make this a reusable function so we don't accidentally end up adding something that is only reset at one place but not the other.

I have:
- Rebased against master to enable sharing with the spawn-related code that we added with pipemenus.
- Explored `set_cloexec()` and `reset_signals_and_limits()`
- Used those two in `spawn-primary-clients.c` to deduplicate code and address your (@Consolatis) point above.

> Also: do we need / want `setsid()` ?

Not sure. Might be better to leave `setsid()` out (but I'm guess here) to make it easier to track things. Thoughts from from @stefonarch  welcome on that one.

src/common/spawn.c
```diff
@@ -25,11 +25,20 @@ reset_signals_and_limits(void)
...
-	fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
+	if (flags == -1) {
+		wlr_log(WLR_ERROR, "Unable to set the CLOEXEC flag: fnctl failed");
+		return false;
+	}
+	flags = flags | FD_CLOEXEC;
+	if (fcntl(fd, F_SETFD, flags) == -1) {
+		wlr_log(WLR_ERROR, "Unable to set the CLOEXEC flag: fnctl failed");
+		return false;
+	}
```
Consolatis, line 40: We could use `wlr_log_errno()` instead, it just applies a `: strerr(errno)` to the end of the log message.

include/common/spawn.h
```diff
@@ -4,6 +4,8 @@
 
 #include <sys/types.h>
 
+bool set_cloexec(int fd);
```
Consolatis: This should have a `#include <stdbool.h>`

src/common/spawn-primary-client.c
```diff
@@ -0,0 +1,103 @@
...
+#include <sys/socket.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#include <wlr/util/log.h>
+#include "common/spawn.h"
+#include "common/spawn-primary-client.h"
+#include "labwc.h"
+
+static int
+sigchld_handler(int fd, uint32_t mask, void *data)
```
Consolatis, line 42: Could we call that differntly? This is not a `SIGCHLD` handler but rather a usual fd watcher that watches for `WL_EVENT_HANGUP | WL_EVENT_ERROR`.

So maybe something like `handle_pipe_error()` instead.

I am also not convinced that we should use the exact cage implementation and not just write our own in `common/spawn.c`. IMHO it basically only clutters the src/common/ directory and is almost the same code that we already have in `spawn.c`. Maybe we could simply change the `spawn_async_no_shell()` function to accept an optional `int *err_pipe` argument and then create a small wrapper function that calls `spawn_sync_no_shell()` and sets up the error handler?

src/server.c
```diff
@@ -574,6 +574,11 @@ server_finish(struct server *server)
 	}
 	wl_display_destroy_clients(server->wl_display);
 
+	int status;
+	if (primary_client_pid) {
+		waitpid(primary_client_pid, &status, 0);
```
Consolatis, line 579: This shouldn't be required anymore because we have a global SIGCHLD handler since the pipe menu implementation.

Consolatis:
Hm.. played around with this a bit and the cage implementation doesn't really make sense for us: it is based on a pipe which triggers an error once \*all\* of the write ends of the pipe are closed. The issue here is that the session manager may spawn further clients which then also inherit the write end of the pipe. That is likely also the reason for the behavior seen with this PR that sometimes more than the session manager itself needs to close.

To make the cage solution work for us we'd need cooperation of the session manager itself so that it sets cloexec on some fd we supply via environment variable or command line argument and that drastically limits this feature. Or it needs to ensure that \*all\* spawned sub processes that could have inherit the write end of the pipe are terminated before the session manager itself terminates.

I'll experiment a bit with another solution where we only hold the pipe fd open for our direct child which then does a blocking wait for the grandchild and then closes the write end of the pipe. That should then correctly terminate labwc as soon as the session manager itself closes, regardless of how many other childs it spawned. And we'd also not leak the write end of the pipe to the session manager in the first place.

johanmalm:
@Consolatis I like that. Makes sense. So, in summary:
- Spawn in a similar way to `spawn_async_no_shell()` but without double fork.
- In sigchld handler, just terminate if pid==primary-client-pid

Consolatis:
> So, in summary:
> * Spawn in a similar way to `spawn_async_no_shell()` but without double fork.
> * In sigchld handler, just terminate if pid==primary-client-pid

Exactly. If this is the behavior we want (e.g. terminate labwc as soon as application given by `--session` terminates) I think this is way less convoluted than to deal with a random pipe that gets inherited by every process in the chain.

src/common/spawn.c
johanmalm: Use of `wlr_log_errno()` implemented.

johanmalm:
@stefonarch Fancy taking this for another test-drive?

stefonarch:
> @stefonarch Fancy taking this for another test-drive?

Working flawlessly :)

johanmalm:
Interactively rebased.

Ready for final review.

Consolatis:
Thanks for all the testing :)

PR #1604, opened 2024-03-10
client menu: add option to allow opening of menu at cursor

klausman:
Until now, the client menu can be two ways by default:

- left click on client menu button at the top left of the window bar
- right click anywhere on the window bar that isn't otherwise bound

The first option opens the menu button, while the second opens it on the window bar, whereever the mouse cursor is.

If one adds a way to open the menu elsewhere, e.g. with this:

```xml
<mouse>
  <context name="Frame">
    <mousebind button="S-Right" action="Press">
      <action name="ShowMenu">
        <menu>client-menu</menu>
      </action>
    </mousebind>
  <context>
</mouse>
```

the menu will open on the client menu spot.

With OpenBox, a similar configuration will result in the menu opening where the mouse cursor is, which is usually more convenient.

This patch adds an option (`menu_clientMenuAtPointer`) that will enable the same behavior as with OpenBox, and default to off (existing behavior).

I am not tied to the naming of the option (or that it lives in the RC root atm), but I thought I'd get the general ide reviewed and then letting the bikeshedding begin. Naturally, docs and rc.xml.all will also be added once the basics have been decided.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -887,6 +887,11 @@ entry(xmlNode *node, char *nodename, char *content)
 			wlr_log(WLR_ERROR, "ignoring invalid value for notifyClient");
 		}
 
+	} else if (!strcasecmp(nodename, "menu_clientMenuAtPointer")) {
+		if (parse_bool(content, -1) == true) {
+			rc.menu_client_menu_at_pointer = true;
+		};
+
 	/* <windowSwitcher show="" preview="" outlines="" /> */
 	} else if (!strcasecmp(nodename, "show.windowSwitcher")) {
 		set_bool(content, &rc.window_switcher.show);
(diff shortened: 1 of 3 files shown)


ahesford:
I'm not sure there's value in providing an option. It's probably better to only force the top-left when the button is clicked, and let the menu open next to the cursor everywhere else.

src/config/rcxml.c
```diff
@@ -887,6 +887,11 @@ entry(xmlNode *node, char *nodename, char *content)
 			wlr_log(WLR_ERROR, "ignoring invalid value for notifyClient");
 		}
 
+	} else if (!strcasecmp(nodename, "menu_clientMenuAtPointer")) {
```
johanmalm, line 890: We'd need `menu.clientMenuAtPointer` in order to parse

```
<menu>
  <clientMenuAtPointer>yes|no</clientMenuAtPointer>
</menu>
```

Default no.

ahesford:
Yeah, I didn't think about keyboard bindings. Opening the menu in the upper-right seems more natural for a keyboard action, while opening at the cursor seems more natural for a mouse action.

I think this is better as a per-action option rather than a per-menu or global option.

ahesford:
#1605 implements my proposed alternative.

johanmalm:
Good shout. A per action option is much better.
Sorry @klausman bad steer earlier :smile:

ahesford:
Superseded by #1605

PR #1605, opened 2024-03-10
action: allow client-menu to open at mouse cursor

ahesford:
This is an alternative to #1604 that implements my desire to make this a per-action argument. Given that `at_cursor` is ignored for `root-menu`, this could also be a separate option, *e.g.* `ShowClientMenuHere`, which takes no arguments.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -105,8 +105,17 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
 
-*<action name="ShowMenu" menu="value" />*
-	Show menu. Valid menu names are "root-menu" and "client-menu".
+*<action name="ShowMenu" menu="value" atCursor="yes" />*
+	Show a menu.
+
+	*menu* The name of the menu to show. The menus "root-menu" and
+	"client-menu" are guaranteed to exist, but others may be defined
+	explicitly. See labwc-menu(5) for more information.
+
+	*atCursor* [yes|no] When opening a menu, open the menu at the location
+	of the mouse cursor. When set to no, the menu will appear at the
+	upper-left corner of the window associated with the action. Default is
+	yes.
 
 *<action name="ToggleDecorations" />*
 	Toggle decorations of focused window.
--- a/src/action.c
+++ b/src/action.c
@@ -321,6 +321,10 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			action_arg_add_str(action, argument, content);
 			goto cleanup;
 		}
+		if (!strcasecmp(argument, "atCursor")) {
+			action_arg_add_bool(action, argument, parse_bool(content, true));
+			goto cleanup;
+		}
 		break;
 	case ACTION_TYPE_TOGGLE_MAXIMIZE:
 	case ACTION_TYPE_MAXIMIZE:
@@ -571,7 +575,8 @@ action_list_free(struct wl_list *action_list)
 }
 
 static void
-show_menu(struct server *server, struct view *view, const char *menu_name)
+show_menu(struct server *server, struct view *view,
+		const char *menu_name, bool at_cursor)
 {
 	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH
 			&& server->input_mode != LAB_INPUT_STATE_MENU) {
@@ -579,34 +584,25 @@ show_menu(struct server *server, struct view *view, const char *menu_name)
 		return;
 	}
 
-	bool force_menu_top_left = false;
 	struct menu *menu = menu_get_by_id(server, menu_name);
 	if (!menu) {
 		return;
 	}
-	if (!strcasecmp(menu_name, "client-menu")) {
-		if (!view) {
-			return;
-		}
-		enum ssd_part_type type = ssd_at(view->ssd, server->scene,
-			server->seat.cursor->x, server->seat.cursor->y);
-		if (type == LAB_SSD_BUTTON_WINDOW_MENU) {
-			force_menu_top_left = true;
-		} else if (ssd_part_contains(LAB_SSD_PART_TITLEBAR, type)) {
-			force_menu_top_left = false;
-		} else {
-			force_menu_top_left = true;
-		}
+
+	int x = server->seat.cursor->x;
+	int y = server->seat.cursor->y;
+
+	/* The client menu needs an active client */
+	if (!view && strcasecmp(menu_name, "client-menu") == 0) {
+		return;
 	}
 
-	int x, y;
-	if (force_menu_top_left) {
+	/* Place menu in the view corner if desired (and menu is not root-menu) */
+	if (!at_cursor && view) {
 		x = view->current.x;
 		y = view->current.y;
-	} else {
-		x = server->seat.cursor->x;
-		y = server->seat.cursor->y;
 	}
+
 	/* Replaced by next show_menu() or cleaned on view_destroy() */
 	menu->triggered_by_view = view;
 	menu_open(menu, x, y);
@@ -764,7 +760,9 @@ actions_run(struct view *activator, struct server *server,
 			kill(getpid(), SIGHUP);
 			break;
 		case ACTION_TYPE_SHOW_MENU:
-			show_menu(server, view, action_get_str(action, "menu", NULL));
+			show_menu(server, view,
+				action_get_str(action, "menu", NULL),
+				action_get_bool(action, "atCursor", true));
 			break;
 		case ACTION_TYPE_TOGGLE_MAXIMIZE:
 			if (view) {
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -21,6 +21,7 @@
 #include "common/nodename.h"
 #include "common/parse-bool.h"
 #include "common/string-helpers.h"
+#include "config/default-bindings.h"
 #include "config/keybind.h"
 #include "config/libinput.h"
 #include "config/mousebind.h"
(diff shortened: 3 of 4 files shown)


klausman:
FWIW, this approach is fine by me, I have no particular attachment to how #1604 works.

docs/labwc-actions.5.scd
```diff
@@ -105,8 +105,15 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
 
-*<action name="ShowMenu" menu="value" />*
-	Show menu. Valid menu names are "root-menu" and "client-menu".
+*<action name="ShowMenu" menu="value" at_cursor="yes" />*
+	Show a menu.
+
+	*menu* The name of the menu to show, which must be one of "root-menu" or
+	"client-menu".
```
Consolatis, line 112: This is wrong, it can be named anything as long its define in `menu.xml`

src/action.c
```diff
@@ -321,6 +321,10 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			action_arg_add_str(action, argument, content);
 			goto cleanup;
 		}
+		if (!strcmp(argument, "at_cursor")) {
```
Consolatis, line 324: I think we usually use `CamelCase` for config args so `atCursor` might be more consistent.

johanmalm:
Nice work @ahesford 

Just thinking... would it make sense to let `atPointer` default to `yes` to clean up `show_menu()` a lot... making the `alt-space` keybind case the special one.

Looks like both `(!strcasecmp(menu_name, "client-menu")` and `ssd_part_contains(LAB_SSD_PART_TITLEBAR, type)` would go away with introducing this option.

We just have to define
- the cursor-button-press on Titlebar to have atPointer==true
- the alt-space key-bind to asPointer==true

ahesford:
In the first of two fixups, I:
1. Renamed `at_cursor -> atCursor` in the configuration
2. Adjusted the documentation to reflect arbitrary menu naming
3. Refactored `show_menu` a bit to take advantage of the new option

In the second fixup, I overhauled the definition of default key and mouse bindings to accommodate the desire to specify the `atCursor` attribute in a couple of `ShowMenu` actions. I considered doing something even more general, like parsing inline XML rather than modifying the combo structs, but that seemed like a much more substantive effort.

This change to the storage format makes declaration of the defaults a bit more cumbersome, but it should make it easy to add extra attributes should the defaults demand them in the future. I've looked at the C99 spec and believe that brace-initialization rules ensure that the attributes array in each struct is NULL terminated, so we should not explicitly require an all-NULL entry in cases where we define at least one attribute. The NULLs that do exist were required to get clang to be quiet about unspecified values. If this refactor is acceptable, we should do a bit more investigation to ensure that the fields really are NULL-terminated.

src/action.c
```diff
@@ -321,6 +321,10 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			action_arg_add_str(action, argument, content);
 			goto cleanup;
 		}
+		if (!strcmp(argument, "atCursor")) {
+			action_arg_add_bool(action, argument, parse_bool(content, true));
```
Consolatis, line 325: I think this has to be `strcasecmp()`, otherwise this will never match (as argument is lower-cased automatically afair)

src/action.c
```diff
@@ -571,42 +575,42 @@ action_list_free(struct wl_list *action_list)
...
+			/* Override placement at cursor when over the window menu */
+			enum ssd_part_type type =
+				ssd_at(view->ssd, server->scene, x, y);
+			at_cursor = type != LAB_SSD_BUTTON_WINDOW_MENU;
 		}
 	}
 
-	int x, y;
-	if (force_menu_top_left) {
+	/* Place cursor in the view corner if desired (and menu is not root-menu) */
```
Consolatis, line 608: `s/Place cursor/Place  menu/` ?

src/config/rcxml.c
```diff
@@ -1163,63 +1163,93 @@ rcxml_init(void)
...
-	{ "XF86_AudioLowerVolume", "Execute", "command", "amixer sset Master 5%-" },
-	{ "XF86_AudioRaiseVolume", "Execute", "command", "amixer sset Master 5%+" },
-	{ "XF86_AudioMute", "Execute", "command", "amixer sset Master toggle" },
-	{ "XF86_MonBrightnessUp", "Execute", "command", "brightnessctl set +10%" },
-	{ "XF86_MonBrightnessDown", "Execute", "command", "brightnessctl set 10%-" },
-	{ NULL, NULL, NULL, NULL },
+	{ "A-Tab", "NextWindow",
+		{ { NULL, NULL } } },
+	{ "W-Return", "Execute",
+		{ { "command", "alacritty" } } },
```
Consolatis, line 1175: Personally I think this is slightly easier to read:
```c
	{ "A-Tab", "NextWindow", {
		{ NULL, NULL } }
	}, {  "W-Return", "Execute", {
		{ "command", "alacritty" } }
	}, {
		..
```

Also, my C foo is a bit lacking here, do we actually need the `{ NULL, NULL }` ones within the initializer? Usually everything not given should be set to a `0` or `NULL` equivalent but I am not sure if that is also happening for arrays within a struct.

Consolatis (commented):
Review below

src/action.c
ahesford: You're right; this baffled me in testing for a few minutes. Fixed.

src/action.c
ahesford: Good catch.

src/config/rcxml.c
ahesford: I adopted your formatting. I tried various combinations of omitting empty values. GCC seems to require at least one field in the first struct of the array, so I was able to use `{ NULL }` instead of `{ NULL, NULL}`; clang complained about that and required at least one full struct initialization in the array before it was happy. The initialization as implemented seems to be the minimum necessary to keep the compilers happy.

src/action.c
```diff
@@ -571,42 +575,34 @@ action_list_free(struct wl_list *action_list)
...
+	if (!view && strcasecmp(menu_name, "client-menu") == 0) {
+		return;
 	}
 
-	int x, y;
-	if (force_menu_top_left) {
+	/* Place menu in the view corner if desired (and menu is not root-menu) */
+	if (!at_cursor && strcasecmp(menu_name, "root-menu") != 0) {
 		x = view->current.x;
 		y = view->current.y;
```
Consolatis, line 603: This will break for custom menu names where there is no active view.
Do we actually need the exception for the "root-menu" here? If somebody sets it up to open as some kind of custom window menu, why not.

So I'd suggest something like this instead
```c
if (!at_cursor && view) {
	x = view->current.x;
	y = view->current.x;
}
```

ahesford: Done

src/config/rcxml.c
```diff
@@ -1163,63 +1163,94 @@ rcxml_init(void)
...
+		{ "command", "brightnessctl set 10%-" } }
+	}, { NULL, NULL, { { NULL, NULL } } },
 };
 
 static void
 load_default_key_bindings(void)
 {
 	struct keybind *k;
 	struct action *action;
+	struct key_combos *current;
```
johanmalm, line 1219: Could reduce scope of `current` (put inside `(for (int i...) { ... }`

src/config/rcxml.c
```diff
@@ -1163,63 +1163,94 @@ rcxml_init(void)
...
-	{ "W-Up", "SnapToEdge", "direction", "up" },
-	{ "W-Down", "SnapToEdge", "direction", "down" },
-	{ "A-Space", "ShowMenu", "menu", "client-menu"},
-	{ "XF86_AudioLowerVolume", "Execute", "command", "amixer sset Master 5%-" },
-	{ "XF86_AudioRaiseVolume", "Execute", "command", "amixer sset Master 5%+" },
-	{ "XF86_AudioMute", "Execute", "command", "amixer sset Master toggle" },
-	{ "XF86_MonBrightnessUp", "Execute", "command", "brightnessctl set +10%" },
-	{ "XF86_MonBrightnessDown", "Execute", "command", "brightnessctl set 10%-" },
-	{ NULL, NULL, NULL, NULL },
+	{ "A-Tab", "NextWindow", {
```
johanmalm, line 1172: Consider simplifying as follows:

```C
struct key_combos {
	const char *binding, *action;
	struct {
		const char *name, *value;
	} attributes[3];
} key_combos[] = { {
		.binding = "A-Tab",
		.action = "NextWindow",
	}, {
		.binding = "A-Space",
		.action = "ShowMenu",
		.attributes[0] = {
			.name = "menu",
			.value = "client-menu",
	       	},
		.attributes[1] = {
			.name = "atCursor",
			.value = "no",
	       	},
	}
};
```

src/config/rcxml.c
```diff
@@ -1163,63 +1163,94 @@ rcxml_init(void)
...
+		{ "command", "amixer sset Master 5%-" } }
+	}, { "XF86_AudioRaiseVolume", "Execute", {
+		{ "command", "amixer sset Master 5%+" } }
+	}, { "XF86_AudioMute", "Execute", {
+		{ "command", "amixer sset Master toggle" } }
+	}, { "XF86_MonBrightnessUp", "Execute", {
+		{ "command", "brightnessctl set +10%" } }
+	}, { "XF86_MonBrightnessDown", "Execute", {
+		{ "command", "brightnessctl set 10%-" } }
+	}, { NULL, NULL, { { NULL, NULL } } },
```
johanmalm, line 1211: The for-loop checks `.binding != NULL` so will be enough to set `.binding = NULL` for last entry.

johanmalm (commented):
Looks good other than the comments.

I love the way `show_menu()` is so much simpler.

src/config/rcxml.c
ahesford: This is, I think, much more readable than other options. I've adopted it int he latest commit.

src/config/rcxml.c
ahesford: Done

src/config/rcxml.c
ahesford: Moved in this function and in `load_default_mouse_bindings`.

ahesford:
Although the latest form of defining `key_combos` and `mouse_combos` looks much cleaner and more obvious, it is also **long**. Thoughts on moving the static variables into `include/config/default-bindings.h` or something like that?

Consolatis:
> Although the latest form of defining `key_combos` and `mouse_combos` looks much cleaner and more obvious, it is also **long**. Thoughts on moving the static variables into `include/config/default-bindings.h` or something like that?

No objections from my side.

ahesford:
I moved the combo definitions and moved to using `ARRAY_SIZE` when iterating over attributes. This will help avoid bugs in the future if we decide to add a third attribtue to a combo and forget to bump the attribute array to 4 elements.

johanmalm:
Just looked through again. Really nice.
Thanks for working on this.
LGTM.

PR #1609, opened 2024-03-11
Add ability to set workspace prefix when using number

droc12345:
Default is still "Workspace" for numbers but this allows user to set the prefix

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -107,6 +107,9 @@
       <prefix>ws</prefix>
     </desktops>
 
+    Or:
+    <desktops number="4" popupTime="500" prefix="ws" />
+
     popupTime defaults to 1000 so could be left out.
     Set to 0 to completely disable the workspace OSD.
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -100,8 +100,17 @@
     Or it can also be configured like this:
     <desktops number="4" />
 
+    Or like this:
+    <desktops>
+      <popupTime>500</popupTime>
+      <number>5</number>
+      <prefix>ws</prefix>
+    </desktops>
+
     popupTime defaults to 1000 so could be left out.
     Set to 0 to completely disable the workspace OSD.
+
+    prefix defaults to "Workspace" when using number instead of names.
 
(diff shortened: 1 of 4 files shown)


docs/rc.xml.all
```diff
@@ -100,9 +100,18 @@
     Or it can also be configured like this:
     <desktops number="4" />
 
+    Or like this:
+    <desktops>
+      <popupTime>500</popupTime>
+      <number>5</number>
+      <prefix>ws</prefix>
+    </desktops>
```
Consolatis, line 108: Could also be written like
```xml
<desktops number="4" popupTime="500"  prefix="ws" />
```

src/config/rcxml.c
```diff
@@ -1465,10 +1467,14 @@ post_processing(void)
 
 	int nr_workspaces = wl_list_length(&rc.workspace_config.workspaces);
 	if (nr_workspaces < rc.workspace_config.min_nr_workspaces) {
+		char *ws_name = xstrdup("Workspace");
+		if (rc.workspace_config.prefix) {
+			ws_name = rc.workspace_config.prefix;
+		}
```
Consolatis, line 1473: Suggest something like this instead
```c
if (!rc.workspace_config.prefix) {
	rc.workspace_config.prefix = xstrdup("Workspace");
}
[..]
workspace->name = strdup_printf("%s %d", rc.workspace_config.prefix, i + 1);
```

This also needs a `zfree(rc.workspace_config.prefix)` in `rcxml_finish()`, otherwise we will leak memory.

Consolatis:
Nice feature, thanks for working on this. Two small comments above.

droc12345:
Ok, made changes. Re-review

Consolatis:
See CodeStyleCheck:
> src/config/rcxml.c:1471: ERROR: spaces required around that '=' (ctx:VxW)
> src/config/rcxml.c:1476: ERROR: trailing whitespace

Other  than that this looks good :)

droc12345:
Ok, recheck ... 

I really hate the online editor *sigh*

Consolatis:
Thanks :)

PR #1610, opened 2024-03-11
keyboard: use 'us' as fallback for XKB_DEFAULT_LAYOUT

johanmalm:
...if keymap cannot be created for the provided XKB_DEFAULT_LAYOUT.

If keymap still cannot be created, exit with a helpful message to avoid crash that is hard to understand.

Fixes: https://github.com/stefonarch/lxqt-labwc-session/issues/7

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
 #include <assert.h>
+#include <stdlib.h>
 #include <wlr/backend/multi.h>
 #include <wlr/backend/session.h>
 #include <wlr/interfaces/wlr_keyboard.h>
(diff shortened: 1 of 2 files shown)


src/input/keyboard.c
```diff
@@ -639,7 +643,14 @@ set_layout(struct server *server, struct wlr_keyboard *kb)
 		}
 		xkb_keymap_unref(keymap);
 	} else {
-		wlr_log(WLR_ERROR, "Failed to create xkb keymap");
+		wlr_log(WLR_ERROR, "failed to create xkb keymap for layout '%s'",
+			getenv("XKB_DEFAULT_LAYOUT"));
+		if (!fallback_mode) {
+			wlr_log(WLR_ERROR, "try to start with layout 'us'");
```
Consolatis, line 649: `Entering fallback mode with layout 'us'` ?

Consolatis:
LGTM other than the minor comment above.

src/input/keyboard.c
johanmalm: Good idea. Have updated with `Entering fallback mode with layout 'us'`

PR #1611, opened 2024-03-11
Window switch, add workspace info and window state (M/m/F)

droc12345:
This shortens xdg-shell/xwayland to single Upper case letter
adds workspace name and state (min, max, full)

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -206,7 +206,7 @@ this is for compatibility with Openbox.
 
 	*content* defines what the field shows and can be any of:
 
-		- *type* Show view type ("xdg-shell" or "xwayland")
+		- *type* Show view type ("W" for xdg-shell or "X" for xwayland)
 
 		- *identifier* Show identifier (app_id for native Wayland
 		  windows and WM_CLASS for XWayland clients)
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -207,7 +207,9 @@ this is for compatibility with Openbox.
 	*content* defines what the field shows and can be any of:
 
 		- *type* Show view type ("xdg-shell" or "xwayland")
-		- *type* Show view type all workspaces ("W" or "X")
+
+		- *winfo* Show window info *type* "X" or "W"
+		  along with *state* M/m/F (max, min, full)
 
 		- *identifier* Show identifier (app_id for native Wayland
 		  windows and WM_CLASS for XWayland clients)
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -208,9 +208,6 @@ this is for compatibility with Openbox.
 
 		- *type* Show view type ("xdg-shell" or "xwayland")
 
-		- *winfo* Show workspace name, *type* "X" or "W"
-		  along with *state* M/m/F (max, min, full)
-
 		- *identifier* Show identifier (app_id for native Wayland
 		  windows and WM_CLASS for XWayland clients)
 
@@ -218,6 +215,14 @@ this is for compatibility with Openbox.
 		  the first two nodes of 'org.' strings.
 
 		- *title* Show window title if different to app_id
+
+		- *workspace* Show workspace name
+
+		- *state* Show window state, M/m/F (max/min/full)
+
+		- *type_short* Show view type ("W" or "X")
+
+		- *output* Show output id, if more than one output detected
 
 	*width* defines the width of the field expressed as a percentage of
 	the overall window switcher width. The "%" character is required.
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -208,7 +208,7 @@ this is for compatibility with Openbox.
 
 		- *type* Show view type ("xdg-shell" or "xwayland")
 
-		- *winfo* Show window info *type* "X" or "W"
+		- *winfo* Show workspace name, *type* "X" or "W"
 		  along with *state* M/m/F (max, min, full)
 
 		- *identifier* Show identifier (app_id for native Wayland
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -206,7 +206,8 @@ this is for compatibility with Openbox.
 
 	*content* defines what the field shows and can be any of:
 
-		- *type* Show view type ("W" for xdg-shell or "X" for xwayland)
+		- *type* Show view type ("xdg-shell" or "xwayland")
+		- *type* Show view type all workspaces ("W" or "X")
 
 		- *identifier* Show identifier (app_id for native Wayland
 		  windows and WM_CLASS for XWayland clients)
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -64,12 +64,21 @@
   </windowSwitcher>
 
   <!--
-    When using all workspaces option of window switcher, there is a replacement
-    for "type" above called "winfo" that will give more info about the window,
-    workspace name, state of window (min, max, full) and output that it's on 
-    if more than one monitor is being used. 
+    When using all workspaces option of window switcher, there are extra fields
+    that can be used, workspace (variable length), state (single space), 
+    type_short (3 spaces), output (variable length), and can be set up 
+    like this. Note: output only shows if more than one output available.
 
-      <field content="winfo" width="25%" />
+    <windowSwitcher show="yes" preview="no" outlines="no" allWorkspaces="yes">
+      <fields>
+        <field content="workspace" width="5%" />
+        <field content="state" width="3%" />
+        <field content="type_short" width="3%" />
+        <field content="output" width="9%" />
+        <field content="identifier" width="30%" />
+        <field content="title" width="50%" />
+       </fields>  
+    </windowSwitcher>
   -->
 
   <!-- edge strength is in pixels -->
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -63,6 +63,15 @@
     </fields>
   </windowSwitcher>
 
+  <!--
+    When using all workspaces optiion of window switcher, there is a replacement
+    for "type" above called "winfo" that will give more info about the window,
+    workspace name, state of window (min, max, full) and output that it's on 
+    if more than one monitor is being used. 
+
+      <field content="winfo" width="25%" />
+  -->
+
   <!-- edge strength is in pixels -->
   <resistance>
     <screenEdgeStrength>20</screenEdgeStrength>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -64,7 +64,7 @@
   </windowSwitcher>
 
   <!--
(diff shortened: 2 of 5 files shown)


docs/labwc-config.5.scd
```diff
@@ -206,7 +206,7 @@ this is for compatibility with Openbox.
 
 	*content* defines what the field shows and can be any of:
 
-		- *type* Show view type ("xdg-shell" or "xwayland")
+		- *type* Show view type ("W" for xdg-shell or "X" for xwayland)
```
Consolatis, line 209: Personally I'd prefer a new content type for this, something like `type_short` (or a better naming).

We try to not surprise users on new releases so changing the look of their setups from one release to the other should not be taken lightly. By defining new types the look doesn't change for existing configs while users can still use the new variations.

src/config/rcxml.c
```diff
@@ -895,9 +895,13 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "outlines.windowSwitcher")) {
 		set_bool(content, &rc.window_switcher.outlines);
 	} else if (!strcasecmp(nodename, "allWorkspaces.windowSwitcher")) {
+		set_bool(content, &rc.window_switcher.allworkspaces);
 		if (parse_bool(content, -1) == true) {
 			rc.window_switcher.criteria &=
 				~LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
```
Consolatis, line 901: No need to parse the bool twice, suggest instead:
```c
if (rc.window_switcher.allworkspaces) {
	[..]
}
```

`allworkspaces` is weird to type and read IMHO, maybe `all_workspaces` for the new config var would be better?

src/config/rcxml.c
```diff
@@ -895,9 +895,13 @@ entry(xmlNode *node, char *nodename, char *content)
...
+		} else {
+			rc.window_switcher.criteria |=
+				LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
```
Consolatis, line 904: This shouldn't be needed as it is initialized that way already.

src/osd.c
```diff
@@ -230,10 +230,10 @@ get_type(struct view *view)
 {
 	switch (view->type) {
 	case LAB_XDG_SHELL_VIEW:
-		return "[xdg-shell]";
+		return "[W] ";
 #if HAVE_XWAYLAND
 	case LAB_XWAYLAND_VIEW:
-		return "[xwayland]";
+		return "[X] ";
```
Consolatis, line 236: See above, suggesting to add a new type instead.

src/config/rcxml.c
droc12345: I wasn't sure about the initialization, thus the else and whether it got reset on a reconfig.

src/config/rcxml.c
droc12345: all_workspaces is fine, so is workspaces_all

src/osd.c
```diff
@@ -362,7 +362,23 @@ render_osd(struct server *server, cairo_t *cairo, int w, int h,
...
+						buf_add(&buf, " m ");
+					} else if ((*view)->fullscreen) {
+						buf_add(&buf, " F ");
+					} else {
+						buf_add(&buf, "   ");
+					}
+				}
 				buf_add(&buf, get_type(*view));
+				if (rc.window_switcher.allworkspaces &&
+					wl_list_length(&server->outputs) > 1) {
```
Consolatis, line 379: Double indent for continued if condition lines.

docs/labwc-config.5.scd
droc12345: I had thought to if/else the change to single letter so that the changes would apply to only the all workspaces option. Is that what you're talking about.

Though I find it easier to glance at a single letter instead of trying to parses x..... when they look similar at first glance.

Consolatis: > I had thought to if/else the change to single letter so that the changes would apply to only the all workspaces option. Is that what you're talking about.

No, I am suggesting to add a completely new [content type](https://labwc.github.io/labwc-config.5.html#entry_windowswitcher_fields_field_content=_width=%), So it can be configured by the user like
```xml
  <windowSwitcher allWorkspaces="yes">
    <fields>
      <field content="type_compact" width="25%" />
      <field content="trimmed_identifier" width="25%" />
    </fields>
  </windowSwitcher>
```
With `type_compact` being the new type.

src/config/rcxml.c
Consolatis: `rcxml_init()` is called on every reconfigure.

docs/labwc-config.5.scd
droc12345: Ok, let me look at what all needs changing for that.

I've made the other changes.

droc12345:
Ok, should encompass the changes mentioned, so recheck.
I'll look at any further comments/requested changes in the morning.

src/osd.c
```diff
@@ -364,6 +378,25 @@ render_osd(struct server *server, cairo_t *cairo, int w, int h,
...
+					if ((*view)->maximized) {
+						buf_add(&buf, " M ");
+					} else if ((*view)->minimized) {
+						buf_add(&buf, " m ");
+					} else if ((*view)->fullscreen) {
+						buf_add(&buf, " F ");
+					} else {
+						buf_add(&buf, "   ");
+					}
+				}
```
Consolatis, line 393: Should we really guard everything with `rc.window_switcher.all_workspaces` here?
The states (max, min, fullscreen) also make sense without the all_workspaces flag.

Similar thing for the output name below, that could have another `windowSwitcher` option like `showOutput`, thoughts @johanmalm ?

src/osd.c
```diff
@@ -364,6 +378,25 @@ render_osd(struct server *server, cairo_t *cairo, int w, int h,
...
+				buf_add(&buf, get_winfo(*view));
+				if (rc.window_switcher.all_workspaces &&
+						wl_list_length(&server->outputs) > 1) {
+					buf_add(&buf, (*view)->output->wlr_output->name);
+				}
```
Consolatis, line 398: I am not sure right now if `view->output` is always set, likely it is not. So we need an additional `&& output_is_usable(*view->output)` in the condition.

src/osd.c
droc12345: I was originally leaving out the if all workspaces, and I have mixed feeling about producing the info
when only current workspace. But either way is really fine. 

The reason for short id vs long name, is just space saving. I didn't want to have so much info that
it would interfere with the next field (no trimming on type/winfo) and along with W/X, M/m/F is 
pretty much understood by most people having used a computer for any time.

src/osd.c
droc12345: All I can say is that the data was never null or wrong when I tested, but I don't use both monitors
that much. And I would argue that if the data is incorrect, then it needs to be fixed to be accurate.

I would say leave it as is, and IF there are reports from those with multi-monitor setups, then 
address this. I'm always open to fixing what I write. ~lol~

droc12345:
This is what it looks like right now. I have increased the window width, but I'm on a 4k monitor.
https://github.com/droc12345/labwc/blob/master/docs/ws-all-osd.jpg

No windows on the 2nd monitor but the system is aware of it.

src/osd.c
droc12345: Alright, removed the if *all_workspaces. If they select "winfo" instead of "type" they will see
changes.

src/osd.c
droc12345: I thought about it, and made the change, they seem to work alright.
Whether more correct, hopefully yes.

Still won't trigger off, unless there's a second monitor.

droc12345:
I think I've fixed all the changes, so ready for re-review.

I see it failed the arch build, but I don't think it has anything to do with these changes, all the others passed

Consolatis:
> I see it failed the arch build, but I don't think it has anything to do with these changes, all the others passed

Yeah, seems like a CI issue for arch with conflicting packages and waiting for user input or something like that.

src/osd.c
```diff
@@ -364,6 +378,23 @@ render_osd(struct server *server, cairo_t *cairo, int w, int h,
 			case LAB_FIELD_TYPE:
 				buf_add(&buf, get_type(*view));
 				break;
+			case LAB_FIELD_WINFO:
+				buf_add(&buf, (*view)->workspace->name);
```
Consolatis, line 382: For the workspace name I think it makes sense to guard it by the `all_workspaces` setting, otherwise its just useless space wasted by duplicating the same workspace name. We could also check if there is more than one workspace defined (the `show_workspace` bool could be reused for that I guess).

droc12345: I put an if all workspaces around workspace name, but replaced with 4 spaces as an else.
But now that I've done it, I'm not sure about that. 
What do you think?

Consolatis: I don't think the 4 spaces are necessary. Suggest something like this:
```c
if (rc.window_switcher.all_workspaces && show_workspace) {
	buf_add(&buf, (*view)->workspace->name);
}
```
This should ensure that the workspace name only occupies space when its actually required (e.g. no space wasted when the switcher is only showing the windows from the current workspace or when there is just one workspace defined anyway).

droc12345: Alright, should be good to go

docs/labwc-config.5.scd
```diff
@@ -208,6 +208,9 @@ this is for compatibility with Openbox.
 
 		- *type* Show view type ("xdg-shell" or "xwayland")
 
+		- *winfo* Show workspace name, *type* "X" or "W"
```
johanmalm, line 211: I prefer not to do a compound type like this. It feels to 'hardcoded' to suit one particular preference, but with not flexibility.

Consider implementing `type_short`, `state` and `workspace` as separate items.

Consider a `custom` type where we could support fields in a string, for example "%w %s %t" or something like that. That way we keep the flexibility in terms of separators, spaces, etc.

docs/rc.xml.all
```diff
@@ -63,6 +63,15 @@
     </fields>
   </windowSwitcher>
 
+  <!--
+    When using all workspaces optiion of window switcher, there is a replacement
```
johanmalm, line 67: `optiion` typo

include/config/rcxml.h
```diff
@@ -141,6 +142,7 @@ struct rcxml {
 		bool show;
 		bool preview;
 		bool outlines;
+		bool all_workspaces;
```
johanmalm, line 145: We should not have a dedicated `all_workspace` bool here. It's already covered by one of the bits in `criteria`. Just do a bitwise check.

johanmalm (changes requested):

droc12345:
I made all the changes yesterday, and it's ready for review.

johanmalm:
@droc12345 Are you okay? I'm conscious we said slightly different things in review comments. Hope this wasn't unhelpful.

droc12345:
I'm fine, I didn't like all the changes that were piling up, so I deleted what I had.

I'll post a clean set of changes.

Issue #1587, opened 2024-03-05
Mouse cursor size not updated on --reconfigure


spl237:
Relatively minor issue, but if you update XCURSOR_SIZE or XCURSOR_THEME in the environment file and then call labwc --reconfigure, the mouse cursor does not change.

The environment variables are reloaded as expected, but it looks to me as if the only place those two variables are checked is in cursor_init, which is only called from server_init at startup.

Is it possible to re-call cursor_init as part of the reload on a SIGHUP? Or does there need to be a different cursor_update function to handle this?

johanmalm:
Ah yes, we ought to get this done. On the TODO list.

Consolatis:
> Is it possible to re-call cursor_init as part of the reload on a SIGHUP? Or does there need to be a different cursor_update function to handle this?

Sadly there doesn't seem to be a way to just update the theme / size of an existing xcursor manager so we'd need a small function that basically contains everything up to `/* Set the initial cursor image so the cursor is visible right away */`. That function can then be called on reconfigure and on `cursor_init()`. The new function would also need to clean up the old `seat->xcursor_manager` if already set (e.g. `wlr_xcursor_manager_destroy()`).

PR #1612, opened 2024-03-12
Reload cursor theme and size on reconfigure

spl237:
As per https://github.com/labwc/labwc/issues/1587 - this seems to work for reloading cursor size and theme on a labwc --reconfigure.

### DIFF
--- a/include/input/cursor.h
+++ b/include/input/cursor.h
@@ -119,6 +119,7 @@ void cursor_update_focus(struct server *server);
 void cursor_update_image(struct seat *seat);
 
 void cursor_init(struct seat *seat);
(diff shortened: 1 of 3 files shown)


spl237:
Looks like something unrelated to this PR is broken in your Arch build system...

spl237:
Now working!

Consolatis:
Code looks good and seems to work fine. Thanks!

I wonder if we should also try to set the cursor image again so that users get instant feedback on reconfigure (if its currently a server side one or set via the cursor shape protocol, in comparison to the application setting it to a surface). But that can be its own PR, I'll try to look into this later.

johanmalm:
Thanks!

PR #1615, opened 2024-03-13
session: Don't close a null dir in read_environment_dir

ScarcelyThere:
There was a cleanup path in read_environment_dir that called closedir() on a null DIR* from opendir(). This removes closedir() from that path.

We also log when environment.d doesn't exist, as opposed to only logging when another error occurred.

### DIFF
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -118,18 +118,14 @@ read_environment_dir(const char *path_prefix)
 	DIR *envdir = opendir(path);
 
 	if (!envdir) {
-		if (errno == ENOENT) {
-			wlr_log(WLR_INFO,
-				"no %s.d directory found",
-				path_prefix);
-		} else {
+		if (errno != ENOENT) {
 			const char *err_msg = strerror(errno);
 			wlr_log(WLR_INFO,
 				"failed to read environment directory: %s",
 				err_msg ? err_msg : "reason unknown");
 		}
 
-		goto no_dir_cleanup;
+		goto env_dir_cleanup;
 	}
 
 	struct dirent *dirent;
@@ -148,7 +144,7 @@ read_environment_dir(const char *path_prefix)
 
 	closedir(envdir);
 
-no_dir_cleanup:
+env_dir_cleanup:
 	free(path);
 	return success;
 }
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -118,14 +118,18 @@ read_environment_dir(const char *path_prefix)
 	DIR *envdir = opendir(path);
 
 	if (!envdir) {
-		if (errno != ENOENT) {
+		if (errno == ENOENT) {
+			wlr_log(WLR_INFO,
+				"no %s.d directory found",
+				path_prefix);
+		} else {
 			const char *err_msg = strerror(errno);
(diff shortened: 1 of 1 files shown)


ScarcelyThere:
This fixed a segmentation fault when launching labwc without an environment.d/ in my .config/labwc/.

src/config/session.c
```diff
@@ -118,14 +118,18 @@ read_environment_dir(const char *path_prefix)
 	DIR *envdir = opendir(path);
 
 	if (!envdir) {
-		if (errno != ENOENT) {
+		if (errno == ENOENT) {
+			wlr_log(WLR_INFO,
+				"no %s.d directory found",
+				path_prefix);
+		} else {
```
ahesford, line 125: The messaging explicity checks for `ENOENT` because a non-existent directory is generally the "normal" case, so I don't think it merits printing a log message. Please drop this block.

src/config/session.c
```diff
@@ -118,14 +118,18 @@ read_environment_dir(const char *path_prefix)
...
 			const char *err_msg = strerror(errno);
 			wlr_log(WLR_INFO,
 				"failed to read environment directory: %s",
 				err_msg ? err_msg : "reason unknown");
 		}
 
-		goto env_dir_cleanup;
+		goto no_dir_cleanup;
```
ahesford, line 132: Please keep the label the same; because we have more than one label in this file, I want to be as specific as possible about its purpose and think "env" is more helpful than "no" for this task.

src/config/session.c
```diff
@@ -142,8 +146,9 @@ read_environment_dir(const char *path_prefix)
 		free(env_file_path);
 	}
 
-env_dir_cleanup:
 	closedir(envdir);
+
+no_dir_cleanup:
```
ahesford, line 151:
```suggestion
env_dir_cleanup:
```

ahesford (changes requested):
Thanks for catching this. What C library are you using? `closedir(NULL)` works fine for me with Linux and glibc 2.38.

Consolatis:
> This fixed a segmentation fault when launching labwc without an environment.d/ in my .config/labwc/.

> What C library are you using? closedir(NULL) works fine for me with Linux and glibc 2.38.

I would also be interested in what system this happens on, maybe we can add it to our CI to prevent things like this from happening again in the future. We could start labwc in headless mode and then terminate it again, it should catch issues like that.

ahesford (approved):

ahesford:
Thanks!

ScarcelyThere:
> > What C library are you using? closedir(NULL) works fine for me with Linux and glibc 2.38.
> 
> I would also be interested in what system this happens on, maybe we can add it to our CI to prevent things like this from happening again in the future. We could start labwc in headless mode and then terminate it again, it should catch issues like that.

I'm sorry I didn't notice this question earlier! I'm running musl 1.2.5 on Alpine Linux, with Linux kernel 6.8.0-rc4.

Consolatis:
> I'm running musl 1.2.5 on Alpine Linux, with Linux kernel 6.8.0-rc4.

Thanks, we added a runtime test to our Void-musl runner that should spot things like this in the future.

PR #1616, opened 2024-03-13
session: abort process_line early if line contains no key

ahesford:
I was going to make the `error` label more descriptive, but realized we don't even need the label. What's more, we can short-circuit some string parsing when would would just wind up ignoring the results.


Consolatis:
:)

PR #1617, opened 2024-03-13
CI: add simple runtime test

Consolatis:
To (hopefully) prevent things like
- #1615

from going undetected in the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,8 @@ jobs:
           pacman-key --init
           pacman -Syu --noconfirm
           pacman -S --noconfirm git meson clang wlroots libdrm libinput \
-            wayland-protocols cairo pango libxml2 xorg-xwayland librsvg libdisplay-info
+            wayland-protocols cairo pango libxml2 xorg-xwayland librsvg \
+            libdisplay-info
 
       - name: Install Debian Testing dependencies
         if: matrix.name == 'Debian'
@@ -78,7 +79,7 @@ jobs:
           sed -i '/^Types/ s/deb/& deb-src/' /etc/apt/sources.list.d/debian.sources
           apt-get update
           apt-get upgrade -y
-          apt-get install -y git gcc clang
+          apt-get install -y git gcc clang gdb xwayland
           apt-get build-dep -y labwc
 
       - name: Install FreeBSD dependencies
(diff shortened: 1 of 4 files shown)


Consolatis:
> /usr/sbin/ld: cannot find -lubsan: No such file or directory

@ahesford any clue what needs to be installed for Void to support `meson configure -DBsanitize=undefined` ?

ahesford:
`libsanitizer-devel` should get you what you need on Void.

scripts/ci/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+	"$1/labwc"
+	-C scripts/ci
+	-d
+)
+
+export XDG_RUNTIME_DIR=$(mktemp -d)
+export WLR_BACKENDS=headless
+
+echo "Starting ${args[@]}"
+"${args[@]}"
```
ahesford, line 18: Capturing output will let us exit without any noise if everything is happy:
```suggestion
output="$("${args[@]}" 2>&1)"
ret=$?

if test $ret -eq 0; then
    echo "labwc terminated cleanly"
    exit 0
fi

echo "${output}"
```

Consolatis: True but I think that having the debug log in there doesn't really harm, as long as the job didn't fail I doubt anyone is actually looking at the logs in the first place.

Consolatis:
Restricted the runtime test to Debian (the Arch repo has issues lately)
Two tests are done:
- compiling with gcc and `-Db_sanitize=undefined` and when labwc exists with `>= 128` ( e.g. 128 + signal) it is ran again in gdb to get a stack trace and the surrounding code. It could potentially be changed to simply load the core dump instead but I am not sure if that is always enabled in our Debian environment.
- doing the same with clang, that returns `1` in case of crashes though and prints its own stacktrace so we don't run it again in gdb

The test could potentially be enhanced in the future to also run logic tests like window operations or triggering keybinds via `wlrctl`. But for now I just want to ensure that we do not crash during startup.

johanmalm:
LGTM

Just a couple of queries:
1. What's the reason for now just building with both `ASAN` and `UBSAN`? I see comment in the notes about it running more slowly, which seems odd to me. If we ran with `address`, would that alleviate the required for `gdb` and keep things simpler?
2. Consider calling it `smoke-test.sh` rather than `ci_test.sh`

Consolatis:
> 1. What's the reason for now just building with both `ASAN` and `UBSAN`? I see comment in the notes about it running more slowly, which seems odd to me. If we ran with `address`, would that alleviate the required for `gdb` and keep things simpler?

Added a comment about that in `smoke-test.sh`, basically as discussed via IRC, in my opinion it is way too spammy (memory-still-in-use-by-external-library) and adds quite some seconds to build up that spam list.

> 2. Consider calling it `smoke-test.sh` rather than `ci_test.sh`

Done.

Thanks for feedback @ahesford and @johanmalm.
This now also catches the earlier Void-musl segfault on `closedir(NULL)` and it doesn't significantly increase the CI runtime.

PR #1623, opened 2024-03-14
Window switch, Osd info enhancement

droc12345:
This should, hopefully, be a clean set of changes to show more info in the osd.

I've modularized the different fields.


src/osd.c
```diff
@@ -364,6 +378,31 @@ render_osd(struct server *server, cairo_t *cairo, int w, int h,
...
+				break;
+			case LAB_FIELD_WIN_STATE:
+				if ((*view)->maximized) {
+					buf_add(&buf, "M");
+				} else if ((*view)->minimized) {
+					buf_add(&buf, "m");
+				} else if ((*view)->fullscreen) {
+					buf_add(&buf, "F");
+				} else {
+					buf_add(&buf, " ");
```
Consolatis, line 395: Do we need this alternate case here? Same question for the `LAB_FIELD_OUTPUT` case below.
Other than that, this LGTM.

Consolatis:
I [played around](https://github.com/Consolatis/labwc/commit/ddd6e4575ff237e815be95a6c1d8486a7551f619) a bit based on this PR and implemented a custom type that can be used like
```xml
<windowSwitcher allWorkspaces="yes">
	<fields>
		<field content="custom" format="foobar %b %3s %-10o %-20W %-10i %t" width="100%" />
	</fields>
</windowSwitcher>
```
- `%b` -> `type_short`
- `%B` -> `type`
- `%s` -> `state`
- `%o` -> `output`

and so on.
If combined with setting a monospace font for the osd this actually works pretty nice.

src/osd.c
Consolatis: I see, without it we may reuse the earlier buffer results because we only reset `buf.len`. We need a `buf_clear(buf)` that also resets the first byte of the internal allocation to `\0`.
Feel free to leave this workaround in for now, we can add the `buf_clear()` in a follow up PR.

droc12345: Yeah, I noticed when I had output set, but nothing was produced due to the if, it repeated what was in the buffer, but I hadn't investigated. The one for  state, I kept so that the output would
stay aligned (I use a mono font)

droc12345:
Custom field looks good.

droc12345:
I did a quick merge (on my system) 1st my patch, then your's over it.

Worked well with either the old way, my newer fields or using the custom field. Good flexibility.

droc12345:
You have my permission, to gut what you need from this PR and merge it into the custom fields you're
working on.

There is no sense in adding some of the code then removing it the next PR.
Better to just take and merge and make one clean addition.
What you need is the *get_type_short function* from osd.c
LAB_FIELD_* stuff from rcxml.c & rcxml.h

You're already producing functions from converting the case statements.

Consolatis:
> You have my permission, to gut what you need from this PR and merge it into the custom fields you're working on.

I am all for merging this PR, it implements the necessary features for the custom field thingie but also is very worth on its own as it defines new fields that users can use in their config.

> There is no sense in adding some of the code then removing it the next PR. Better to just take and merge and make one clean addition. What you need is the _get_type_short function_ from osd.c LAB_FIELD_* stuff from rcxml.c & rcxml.h

Well, we do that all the time. That is how codebases evolve over time :)
You implement a nice feature and that then gets refactored / moved around / re-used for other cases.

droc12345:
That's fine too. Then it should be good to go on my end.

With your changes, on top, I think that osd is in good shape, both PRs should address the issue that users
have with which window is on which workspace. 
I do like the format ability, nice to be able to easily customize.

johanmalm:
Thanks. I squashed and modified the commit-message a bit.

stefonarch:
There was a discussion somewhere to export workspace and window state information so that it can be hooked  inby a workspace switcher in panels, I can't find it atm.

Is this a possibility? Open a feature request?

Consolatis:
> export workspace

ext-workspaces protocol, not accepted so far.

> window state information

wlr-foreign-toplevel protocol and possibly a ext follow up with the missing states.

> Open a feature request?

Unlikely that labwc will implement any workarounds without some "official" wayland (or wlroots) protocol. For me personally, I really don't want to partake into actively further fragmenting the wayland eco system. I know this sucks but what is the point of having various implementations in every panel for every compositor out there.

stefonarch:
Fully understandable. A simple desktop switcher is already possible, using the shortcut, `wtype`  and `customcommand` plugin in `lxqt-panel`.

PR #1627, opened 2024-03-16
CI smoke test

Consolatis:
I've seen a segfaut on void with the new runtime test once (on a PR without any actual changes that could cause it) but retrying the job didn't show the segfault again. Unfortunately, only the first test failed and the 2nd run under gdb that should provide the stacktrace didn't fail. So lets do repeated runs for a bit.

Not yet sure if we should actually add this to the CI, likely not. Or maybe as a custom job that can only be triggered manually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
   #        ./scripts/find-banned.sh
   build:
     name: Build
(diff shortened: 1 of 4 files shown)


Consolatis:
A CI smoke-test has long been merged into master.

PR #1629, opened 2024-03-16
Add and use buf_finish() + fix buf reuse issue in osd.c + some minor buf related improvements

Consolatis:
Only slightly tested.

### DIFF
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -69,7 +69,7 @@ buf_expand_shell_variables(struct buf *s)
 			buf_add_char(&new, s->buf[i]);
 		}
 	}
-	free(environment_variable.buf);
+	buf_finish(&environment_variable);
 	free(s->buf);
 	s->buf = new.buf;
 	s->len = new.len;
@@ -79,7 +79,10 @@ buf_expand_shell_variables(struct buf *s)
 void
 buf_init(struct buf *s)
 {
-	/* we can't assert(!s->buf) here because struct may be uninitialized */
+	/*
+	 * We can't assert(!s->buf) here because
+	 * the supplied struct may be uninitialized.
+	 */
 
 	s->alloc = 256;
 	s->buf = xmalloc(s->alloc);
@@ -130,6 +133,21 @@ buf_clear(struct buf *s)
 void
 buf_reset(struct buf *s)
 {
-	zfree(s->buf);
+	assert(s->buf);
+
+	buf_finish(s);
 	buf_init(s);
+}
+
+void
+buf_finish(struct buf *s)
+{
+	assert(s->buf);
+
+	/*
+	 * Using zfree rather than free ensures that the
+	 * asserts will be triggered whenever something
+	 * other than buf_init() is called on the buffer.
+	 */
+	zfree(s->buf);
 }
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -5,13 +5,40 @@
 #include "common/buf.h"
 #include "common/mem.h"
 
+#define BUF_INITIAL_SIZE 256
+
+#define BUF_CLEARED(s) ((s) && !(s)->buf && !(s)->alloc && !(s)->len)
+#define BUF_INITIALIZED(s) ((s) && (s)->buf && (s)->alloc >= BUF_INITIAL_SIZE)
+
+/*
+ * Replace dst by src, src is invalid after this call
+ * but it may be reused when calling buf_init(src).
+ *
+ * dst must either have been initialized with buf_init()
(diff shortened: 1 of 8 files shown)


src/common/buf.c
```diff
@@ -5,6 +5,15 @@
 #include "common/buf.h"
 #include "common/mem.h"
 
+static void
+buf_replace_by(struct buf *dst, struct buf *src)
```
johanmalm, line 9: Consider calling it `buf_take()` or `buf_steal()`. Common when transferring ownership of heap-allocated memory.

src/common/buf.c
```diff
@@ -5,6 +5,15 @@
...
+{
+	free(dst->buf);
+	dst->buf = src->buf;
```
jlindgren90, line 12: Maybe reset `src->buf` to `NULL` also?

Consolatis: Replaced everything in the src buffer by `*src = (struct buf){0}`.

Also added some asserts which try to ensure that the dst buffer is either cleared out or initialized, otherwise supplying just a stack allocated dst buffer would cause us to try to free a random memory address.

src/common/buf.c
Consolatis: Personally I like `buf_replace_by()` better, especially as it somewhat implies what is replaced by what (e.g. first arg is the buffer this API targets / `self` and second then must be the src). But its just some internal static function so also fine changing it to something else if you insist.

src/common/buf.c
jlindgren90: Trying to detect an uninitialized dst buffer seems like it may not be reliable (the garbage values might look valid). So I would lean toward omitting those checks. OK with me either way though.

jlindgren90:
Thinking out loud...

It seems confusing to have `buf_clear`, `buf_reset`, and `buf_finish`, which are all slightly different but all leave the buf in some sort of "empty" state.

Could we consider a zeroed buf (e.g. `struct buf s = {0};`) a valid state, make `buf_init` optional, and make `buf_add[_char]` do the initial allocation automatically if needed?

Then it would no longer be necessary to have a separate `buf_reset` and `buf_finish`. We could just have a single function that does `free(s->buf); *s = (struct buf){0};` -- that would work whether or not the buf is going to be re-used, because it's a valid state but no extra memory is allocated.

`buf_init` could still exist in case we specifically want a buf containing an empty string and not NULL.

Consolatis:
> It seems confusing to have `buf_clear`, `buf_reset`, and `buf_finish`, which are all slightly different but all leave the buf in some sort of "empty" state.

`buf_finish` leaves the buffer in an invalid state while `buf_clear` and `buf_reset` do leave the buffer in a usable but empty state.
`buf_reset` isn't actually used yet AFAIR so we could simply remove `buf_reset`.

> Could we consider a zeroed buf (e.g. `struct buf s = {0};`) a valid state, make `buf_init` optional, and make `buf_add[_char]` do the initial allocation automatically if needed?

I kind of like this idea. If we leave `buf_init()` in but optional (if struct is zeroed) that would also not require any code changes.

src/common/buf.c
Consolatis: > Trying to detect an uninitialized dst buffer seems like it may not be reliable (the garbage values might look valid).

Agree, we'd need to have some defined magic number in one of the fields and at that point we are way overkill for a static function that is only called within the same small `.c` file. I'll remove them.

johanmalm:
Git inits strings by pointing them to `STRBUF_INIT`.

https://github.com/git/git/blob/11c821f2f2a31e70fb5cc449f9a29401c333aad2/strbuf.h#L80-L81

Not sure what I think of it, but avoids the NULL pointer. Obviously will require some additional lines in several functions, but potentially makes it easier for the end-user.

src/common/buf.c
jlindgren90: I would vote for `buf_move` since this is analogous to move-assignment from C++.

Consolatis:
Do we want this in 0.7.2?

Otherwise we should replace the osd.c usage of `buf.len = 0;` with `buf.len = 0; buf.buf[0] = '\0';` separably.

jlindgren90:
> Git inits strings by pointing them to `STRBUF_INIT`.

I kind of like this. We can define:
```
#define BUF_INIT ((struct buf){.buf = ""})
```

... and simply not do `free(s->buf)` when `s->alloc` is 0.

jlindgren90:
> Do we want this in 0.7.2?

I'm working on a rebase that includes the `BUF_INIT` idea, will submit a separate PR shortly.

jlindgren90:
Obsolete now that #1713 has been merged?

PR #1631, opened 2024-03-17
common: render text buffers with opaque background

jlindgren90:
After a roundabout discussion[1] with wlroots devs, it's become apparent that subpixel text rendering (a.k.a. "ClearType") does not work properly when rendering over a transparent background, as labwc currently does.

Basically it comes down to the fact that the color of semi-transparent pixels (which is adjusted redder or bluer to compensate for RGB subpixel alignment) depends somewhat on background color. When rendering over transparency, the text engine doesn't know the intended background color and can't adjust the pixel colors correctly.

With Pango/Cairo, the end result can range from grayscale rendering (no subpixel rendering at all) to wrong/oversaturated colors (for example, bright pink pixels when rendering white text on blue background).

This change solves the issue by first filling the text buffer with an opaque background color before rendering the text over it. Currently, this is easy since the background is always a solid color. It may be a little more complex (but doable) if we implement gradients in future.

Note that GTK 4 (and to some degree, recent versions of Microsoft Windows) avoid this issue by disabling subpixel rendering altogether. I would much prefer that labwc NOT do this -- it results in noticeably blurrier text on non-retina LCD screens, which are still common.

[1] https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3822

Screenshots (zoom in to see detail):

White on blue, `WLR_RENDERER=vulkan`, before/after:
vulkan-before
vulkan-after

Black on white, `WLR_RENDERER=pixman`, before/after:
pixman-bw-before
pixman-bw-after

### DIFF
--- a/include/common/font.h
+++ b/include/common/font.h
@@ -42,11 +42,12 @@ int font_width(struct font *font, const char *string);
  * @text: text to be generated as texture
  * @font: font description
  * @color: foreground color in rgba format
+ * @bg_color: background color in rgba format
  * @arrow: arrow (utf8) character to show or NULL for none
  */
 void font_buffer_create(struct lab_data_buffer **buffer, int max_width,
-	const char *text, struct font *font, float *color, const char *arrow,
-	double scale);
+	const char *text, struct font *font, const float *color,
+	const float *bg_color, const char *arrow, double scale);
 
 /**
  * font_finish - free some font related resources
--- a/include/common/graphic-helpers.h
+++ b/include/common/graphic-helpers.h
@@ -42,7 +42,7 @@ void multi_rect_set_size(struct multi_rect *rect, int width, int height);
  * Sets the cairo color.
  * Splits a float[4] single color array into its own arguments
  */
-void set_cairo_color(cairo_t *cairo, float *color);
+void set_cairo_color(cairo_t *cairo, const float *color);
 
 /* Draws a border with a specified line width */
 void draw_cairo_border(cairo_t *cairo, struct wlr_fbox fbox, double line_width);
--- a/include/common/scaled_font_buffer.h
+++ b/include/common/scaled_font_buffer.h
@@ -17,6 +17,7 @@ struct scaled_font_buffer {
 	char *text;
 	int max_width;
 	float color[4];
+	float bg_color[4];
 	char *arrow;
 	struct font font;
 	struct scaled_scene_buffer *scaled_buffer;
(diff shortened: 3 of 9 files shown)


ahesford:
Nice find! I noticed the strange coloration with sub-pixel rendering (which seems worse with Vulkan), but never would have figured this was related to transparent backgrounds.

Consolatis:
Very nice find indeed!

LGTM, didn't test.

jlindgren90:
> Edit 2: Actually, after reading though the linked wlroots issue: We do create the cairo surface with `CAIRO_FORMAT_ARGB32` and then set the wlroots buffer format to `DRM_FORMAT_ARGB8888`. I don't know enough about color formats to verify if that is actually correct.

For what it's worth, most of the discussion on the wlroots issue (before Félix weighed in) ended up being irrelevant. My original theories were wrong, and the pixel/color formats themselves are fine.

The issue is really just that Pango/Cairo does not behave properly (and in some cases actually produces out-of-range RGB values, which is a bug) when doing subpixel rendering over transparency. The behavior could be improved (e.g. limit the RGB values to be no greater than the alpha value -- I tried that as a post-processing step, and it helped) but the result would still be less than ideal. The subpixel rendering really needs to be done over an opaque background to work correctly. That's my understanding, anyway.

Consolatis:
> For what it's worth, most of the discussion on the wlroots issue (before Félix weighed in) ended up being irrelevant. My original theories were wrong, and the pixel/color formats themselves are fine.

Well it was interesting nevertheless. Even though I will likely have forgotten everything within two weeks again.
The intimate details of color formats are just not my cup of tea, I prefer things to be slightly higher level.

> The subpixel rendering really needs to be done over an opaque background to work correctly. That's my understanding, anyway.

Should we add the [cairo issue](https://gitlab.freedesktop.org/cairo/cairo/-/issues/829) within the comment?

Anyway, feel free to go for the rebase and merge if you are happy with the PR. You have invested way more time in debugging and testing the solution / workaround than all of us and are thus the most qualified person to judge it.

ahesford:
Just tested this on my laptop, and the difference is immediately visible. I always assumed the color artifacts were caused by resampling issues with fractional scaling.

johanmalm:
Thanks. Fantastic work.

This adds a limitation to how we composite SSD, but that's a discussion for a different day.

For what it's worth, I'm increasingly thinking that that we ought to consider assembling multiple "parts" in one cairo-surface anyway. This needs to experimentation, speed-testing and a strategy. For example, I think we ought to consider just creating one surface for the whole titlebar (incl. corners).

PR #1632, opened 2024-03-17
NLS: translate "Workspaces" in rcxml.c. closes #1630

### LINKED ISSUE #1630: NLS: "Workspace"
At the time of writing, `rcxml.c` at line ~1484~ contains the string "Workspace". 

Being an UI element it probably should be internationalised.

Whoops line 1384

01micko:
Translators - no need to update as "Workspaces" is already covered in `menu.c` and gettext is smart enough to re-use it.

I took the liberty to update all the comments in the `po` files too. tested about 6 locales and all works, including RTL in Arabic (https://github.com/labwc/labwc/issues/1630#issuecomment-2002306206). Would appreciate some CJK testing as I don't have those fonts, but there should be no problem.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -377,9 +377,10 @@ translation strings under each English string.
 ## Coders
 
 Code contributors may need to update relevant files if their additions
-affect UI elements (at the moment only `src/menu/menu.c`). In this case
-the `po/labwc.pot` file needs to be updated so that translators can
-update their translations. Remember, many translators are _not_ coders!
+affect UI elements (at the moment only `src/menu/menu.c` and 
+`src/config/rcxml.c`). In this case the `po/labwc.pot` file needs to be
+updated so that translators can update their translations. Remember,
+many translators are _not_ coders!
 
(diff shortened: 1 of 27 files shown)


Consolatis:
If you remove the 2 additional spaces I am ready to merge.
I can squash the commit directly within the PR so just pushing a fixup commit is totally fine.

01micko:
Oh, both eh?

Lol, just removed end one, will do other now.

Consolatis:
Thanks as always :)

01micko:
> Would appreciate some CJK testing as I don't have those fonts, but there should be no problem.

As a follow up to this @Consolatis , @johanmalm I installed `wqy-microhei.ttc` CJK font and added this to my `rc.xml`

```xml
    <font>
      <name>WenQuanYi Micro Hei</name>
      <size>10</size>
      <slant>normal</slant>
      <weight>normal</weight>
    </font>
```

_osd_ and _workplaces_ work correctly as far as I can tell just by comparing the strings to the respective `zh_CN.po` and `ja.po` for Chinese and Japanese.

PR #1635, opened 2024-03-17
xwayland: exclude unfocusable views from wlr-foreign-toplevel

jlindgren90:
These views (notifications, floating toolbars, etc.) should not be shown in taskbars/docks/etc. (which are the stated use-case of the wlr-foreign-toplevel protocol).

See also: https://github.com/jlindgren90/qmpanel/issues/7


johanmalm:
LGTM

PR #1636, opened 2024-03-17
xwayland: fix segv bug when starting game

johanmalm:
@kode54 testing welcome

We could just return on `!root` but I suspect moving the supplied view to front is better.


johanmalm:
Think #1637 is preferable in term of implementation (thanks @Consolatis), but both should fix the issue in the same way.

PR #1638, opened 2024-03-18
workspaces.c: fix bug with RTL locales, closes #1633

### LINKED ISSUE #1633: workspaces: when cycling through desktops RTL languages shift right too much.
So, when a feature is added then a bug is found! Re [5a20014](https://github.com/labwc/labwc/commit/5a20014f5027c72519579010469b05b75dd50526)

I found that when cycling trough desktops with an RTL locale (which I didn't test :( ) that the text is apparently on the wrong side of the desktop number. EDIT: no, the whole thing is pushed right, text is correct in relation to the number.

ar-bug

Works fine in the main window switcher osd so is probably a fairly simple fix. Maybe a pango option missing or something.

01micko:
Thanks to @Consolatis  for the patch.

### DIFF
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -124,22 +124,23 @@ _osd_update(struct server *server)
 		/* Text */
 		set_cairo_color(cairo, server->theme->osd_label_text_color);
 		PangoLayout *layout = pango_cairo_create_layout(cairo);
-		pango_layout_set_width(layout, (width - 2 * margin) * PANGO_SCALE);
 		pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
 
-		PangoFontDescription *desc = font_to_pango_desc(&rc.font_osd);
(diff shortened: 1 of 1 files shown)


src/workspaces.c
```diff
@@ -124,22 +124,23 @@ _osd_update(struct server *server)
...
-		PangoFontDescription *desc = font_to_pango_desc(&rc.font_osd);
-		pango_layout_set_font_description(layout, desc);
-
 		/* Center workspace indicator on the x axis */
-		x = font_width(&rc.font_osd, server->workspace_current->name);
-		x = (width - x) / 2;
+		int req_width = font_width(&rc.font_osd, server->workspace_current->name);
+		if (req_width > width - 2 * margin) {
+			req_width = width - 2 * margin;
+		}
```
Consolatis, line 133: Suggest this one instead of the if:
```c
req_width = MIN(req_width, width - 2 * margin);
```

PR #1639, opened 2024-03-18
workspaces.c: fix bug with RTL locales, closes #1633

### LINKED ISSUE #1633: workspaces: when cycling through desktops RTL languages shift right too much.

01micko:


01micko:
I don't know, haven't force-pushed in awhile so something got messed up!

01micko:
Ah, git fu returns!

Consolatis:
Whoops, I seem to accidentally have created the diff I posted in the issue with an older version of labwc that I had checked out for something else, sorry about that.

Changes LGTM but I don't want to self-merge without a second LGTM as I am also the author of the changes and this doesn't deserve an override of our review rules (e.g. not a crash fix or minor documentation adjustment).

01micko:
> Whoops, I seem to accidentally have created the diff I posted in the issue with an older version of labwc that I had checked out for something else, sorry about that.


Yes I know that and I patched in manually. I can wait for @johanmalm or anyone else.

johanmalm:
Thanks.

PR #1646, opened 2024-03-19
layers: unconstrain popup within usable area

johanmalm:
Helped-by: @Consolatis

@jlindgren90 Might this one help with the position of the `qmpanel` menu?

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -469,6 +469,7 @@ handle_new_popup(struct wl_listener *listener, void *data)
 	wlr_output_layout_get_box(server->output_layout,
 		output->wlr_output, &output_box);
 
+	struct wlr_box usable = output_usable_area_in_layout_coords(output);
 	/*
 	 * Output geometry expressed in the coordinate system of the toplevel
 	 * parent of popup. We store this struct the lab_layer_popup struct
@@ -476,10 +477,10 @@ handle_new_popup(struct wl_listener *listener, void *data)
 	 * the bottom to the top layer.
 	 */
 	struct wlr_box output_toplevel_sx_box = {
-		.x = output_box.x - lx,
-		.y = output_box.y - ly,
-		.width = output_box.width,
-		.height = output_box.height,
+		.x = usable.x - lx,
+		.y = usable.y - ly,
+		.width = usable.width,
+		.height = usable.height,
 	};
 	struct lab_layer_popup *popup = create_popup(wlr_popup, surface->tree);
 	popup->output_toplevel_sx_box = output_toplevel_sx_box;


src/layers.c
```diff
@@ -469,17 +469,18 @@ handle_new_popup(struct wl_listener *listener, void *data)
 	wlr_output_layout_get_box(server->output_layout,
 		output->wlr_output, &output_box);
 
+	struct wlr_box usable = output_usable_area_in_layout_coords(output);
```
Consolatis, line 472: `struct wlr_box output_box` can be removed.

jlindgren90: And `struct server *server` too

jlindgren90:
> @jlindgren90 Might this one help with the position of the `qmpanel` menu?

This one does help somewhat, thanks. Top-level menus don't overlap the panel any more. Bottom-anchored menus still don't adjust position when changing size, and sub-menus are not aligned with the menu item they opened from.

I don't really understand layer-shell positioning, so I don't know if the remaining issues are the fault of labwc or Qt/LayerShellQt (or a limitation in the protocol itself).

tokyo4j:
I wonder if this is the right way to go.
I'm afraid any layer shell clients want to place popups at cursor position, rather than next to the area it excludes.

For qmpanel, seems like the problem lies in client side, as its anchor rect is too small.

This is the log when I clicked the clock item in qmpanel:
```
[ 665627.721]  -> xdg_positioner@54.set_anchor_rect(1615, 24, 1, 1)
[ 665627.726]  -> xdg_positioner@54.set_anchor(7)
[ 665627.736]  -> xdg_positioner@54.set_gravity(8)
[ 665627.740]  -> xdg_positioner@54.set_size(304, 171)
[ 665627.744]  -> xdg_positioner@54.set_constraint_adjustment(3)
[ 665627.749]  -> xdg_surface@53.get_popup(new id xdg_popup@55, nil, xdg_positioner@54)
```

And this is the log when I hovered cursor over taskbar item in Waybar:
```
[ 559811.695]  -> xdg_positioner@54.set_size(308, 44)
[ 559811.699]  -> xdg_positioner@54.set_anchor_rect(-4, -4, 350, 48)
[ 559811.703]  -> xdg_positioner@54.set_offset(21, 0)
[ 559811.706]  -> xdg_positioner@54.set_anchor(2)
[ 559811.709]  -> xdg_positioner@54.set_gravity(2)
[ 559811.712]  -> xdg_positioner@54.set_constraint_adjustment(9)
[ 559811.720]  -> xdg_surface@55.get_popup(new id xdg_popup@63, nil, xdg_positioner@54)
```

Consolatis:
> I wonder if this is the right way to go.
> I'm afraid any layer shell clients want to place popups at cursor position, rather than next to the area it excludes.

Hmm, that is a good point. In some layershell GTK panel I use openAtWidget with `north` (if panel on the bottom) so the popup renders outside of the exclusive zone but other layershell clients may not want that behavior.

jlindgren90:
> I'm afraid any layer shell clients want to place popups at cursor position, rather than next to the area it excludes.

For what it's worth, in the "wild west" of X11, popups usually remain within the usable area (per _NET_WORKAREA property) but this is enforced on the client side, typically at the toolkit level. For example, Qt does not place menus outside _NET_WORKAREA -- but other X11 clients not using Qt are able to so if they want.

I don't know how this is supposed to work in Wayland/layer-shell, with the different (compared to X11) division of compositor and client roles.

> For qmpanel, seems like the problem is in client side, as its anchor rect is too small.

Quite possible. This is all managed by LayerShellQt though (or possibly Qt itself); I have not done anything with anchor rects in the application-level code.

tokyo4j:
I found following piece of code from git://code.qt.io/qt/qtwayland.git:
```c
QRect placementAnchor = m_window->window()->property("_q_waylandPopupAnchorRect").toRect();
    if (!placementAnchor.isValid()) {
        placementAnchor = QRect(m_window->geometry().topLeft() - parent->geometry().topLeft(), QSize(1,1));
    }
...
    positioner->set_anchor_rect(placementAnchor.x(),
                                placementAnchor.y(),
                                placementAnchor.width(),
                                placementAnchor.height());
```

So it looks like manually setting `_q_waylandPopupAnchorRect` would mitigate this problem.
I need more investigation into Qt as I'm totally new to it, though.

tokyo4j:
Anyway, I think it's fine to merge this PR for now as a workaround.
I'd love to try qmpanel as an alternative to waybar/sfwbar 😄

stefonarch:
>I found following piece of code from git://code.qt.io/qt/qtwayland.git:

That could be interesting for us too, thanks!

> I'd love to try qmpanel as an alternative to waybar/sfwbar 😄

FYI in a few days lxqt-panel PR for wayland should be merged in git.

Consolatis:
> Anyway, I think it's fine to merge this PR for now as a workaround.

You raised a good point in that this will prevent layershell clients to position popups on top of reserved area, potentially including their own. This might also apply to things like tooltips, depending on how they are implemented (as popups or subsurfaces). If we merge this we should make it very clear that this is a temporary workaround for Qt-layershell specifically.

stefonarch:
> You raised a good point in that this will prevent layershell clients to position popups on top of reserved area, potentially including their own. 

I didn't test it but with` lxqt-panel` on bottom or right position some menus and all tooltips are opened at screen border atm.

tokyo4j:
I got it working somehow!

stefonarch:
Tested it now,  tooltips and "add-favorites" popup are  aligned to panel border instead of screen border with right and bottom panel. It's something but it should be aligned to item as it is in top or left panel but this is probably a layer-shell-qt issue, as top and left had always been working without it too.

johanmalm:
> I managed to integrate xdg-positioner into qt somehow!

Nice demo @tokyo4j!

I will set this PR to draft, as I'm not convinced it right to merge it. I've had a check with [tint](https://github.com/johanmalm/tint) and not realise that it moves both popups + tooltips away from the pointer to the usable area. Whilst this might be great for some use-cases, it's probably not something the compositor should enforce.

jlindgren90:
Following up on this -- for the purposes of qmpanel, I don't think this is needed after all. I was able to adapt @tokyo4j's demo and get top-level menus positioned correctly (without overlapping the panel).

> tooltips and "add-favorites" popup are aligned to panel border instead of screen border with right and bottom panel (only in kwin it is also like that).

I think this is a Qt bug. There is a similar issue with nested menus. Qt tries to limit popup position within the screen area but it doesn't know the position of the panel on-screen (due to Wayland policy/limitation) and assumes the panel is at (0,0). Therefore Qt thinks any region above or to the left of the panel is off-screen, and doesn't want to put popups there.

PR #1647, opened 2024-03-21
rcxml: add missing check for calibrationMatrix

SnowNF:
It's my fault, forgetting to test for calibration matrix elements less than 6. ;)


johanmalm:
Thanks. Good fix. I should have spotted that!

PR #1648, opened 2024-03-21
ssd: add window drop-shadows

cillian64:
Add optional drop-shadows to windows using server-side decoration. The presence and appearance of the shadows can be configured in rc.xml separately for active and inactive windows.  The default is no shadows to preserve current behaviour.

The shadows are drawn in two fixed buffers containing a corner and edge, then the edges are scaled according to window size.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -130,6 +130,20 @@ This syntax is not documented on the openbox.org wiki, but is supported by
 openbox and is used by many popular themes. For the sake of brevity, these
 elements are not listed here, but are supported.
 
+*window.active.shadow.size*
+	Size of the drop-shadow for the focused window, in pixels. Default is 60.
+
+*window.inactive.shadow.size*
+	Size of drop-shadows for non-focused windows, in pixels. Default is 40.
+
+*window.active.shadow.color*
+	Color of the drop-shadow for the focused window, including opacity. Default
+	is #00000060 (black with 38% opacity).
+
+*window.inactive.shadow.color*
+	Color of drop-shadows for non-focused windows, including opacity. Default is
+	#00000040 (black with 25% opacity).
+
 *menu.items.bg.color*
 	Background color of inactive menu items.
 
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -884,6 +884,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.corner_radius = atoi(content);
 	} else if (!strcasecmp(nodename, "keepBorder.theme")) {
 		set_bool(content, &rc.ssd_keep_border);
+	} else if (!strcasecmp(nodename, "dropShadows.theme")) {
+		set_bool(content, &rc.shadows_enabled);
 	} else if (!strcmp(nodename, "name.font.theme")) {
 		fill_font(nodename, content, font_place);
 	} else if (!strcmp(nodename, "size.font.theme")) {
@@ -1192,6 +1194,7 @@ rcxml_init(void)
 	rc.xdg_shell_server_side_deco = true;
 	rc.ssd_keep_border = true;
 	rc.corner_radius = 8;
+	rc.shadows_enabled = false;
 
 	init_font_defaults(&rc.font_activewindow);
 	init_font_defaults(&rc.font_inactivewindow);
--- a/src/ssd/ssd_shadow.c
+++ b/src/ssd/ssd_shadow.c
@@ -1,0 +1,316 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+#include <assert.h>
+#include "common/scene-helpers.h"
+#include "labwc.h"
+#include "buffer.h"
+#include "ssd-internal.h"
+#include "theme.h"
+#include "view.h"
+#include <cairo.h>
+
+#define FOR_EACH_STATE(ssd, tmp) FOR_EACH(tmp, \
+	&(ssd)->shadow.active, \
+	&(ssd)->shadow.inactive)
+
+/*
+ * Implements point_accepts_input for a buffer which never accepts input
+ * because drop-shadows should never catch clicks!
+ */
+static bool
+never_accepts_input(struct wlr_scene_buffer *buffer, double *sx, double *sy)
+{
+	return false;
+}
+
+/*
+ * Setup transform and scale for shadow corner buffers. Cropping is applied if
+ * the window is short or narrow enough that corners would overlap, the amount
+ * to crop is controlled by vertical_overlap and horizontal_overlap.  Cropping
+ * is applied before rotation so switch_axes should be true for the bottom-left
+ * and top-right corners to crop horizontally instead of vertically.
+ */
+static void
+corner_scale_crop(struct wlr_scene_buffer *buffer, int horizontal_overlap,
+		int vertical_overlap, int corner_size, bool switch_axes)
+{
+	int width = corner_size - horizontal_overlap;
+	int height = corner_size - vertical_overlap;
+	/* Crop is applied before rotation so gets the axis flip */
+	struct wlr_fbox src_box = {
+		.x = switch_axes ? vertical_overlap : horizontal_overlap,
+		.y = switch_axes ? horizontal_overlap : vertical_overlap,
+		.width = switch_axes ? height : width,
+		.height = switch_axes ? width : height,
+	};
+	wlr_scene_buffer_set_source_box(buffer, &src_box);
+	/* But scaling is applied after rotation so no axis flip */
+	wlr_scene_buffer_set_dest_size(buffer, width, height);
+}
+
+/*
+ * Set the position, scaling, and visibility for a single part of a window
+ * drop-shadow.
+ */
+static void
+set_shadow_part_geometry(struct ssd_part *part, int width, int height,
+		int titlebar_height, int corner_size, int inset,
+		int visible_shadow_width)
+{
+	struct wlr_scene_buffer *scene_buf =
+		wlr_scene_buffer_from_node(part->node);
+	/*
+	 * If the shadow inset is greater than half the overall window height
+	 * or width (eg. becaused the window is shaded or because we have a
+	 * small window with massive shadows) then the corners would overlap
+	 * which looks horrible.  To avoid this, when the window is too narrow
+	 * or short we hide the edges on that axis and clip off the portion of
+	 * the corners which would overlap.  This does produce slight
+	 * aberrations in the shadow shape where corners meet but it's not too
+	 * noticeable.
+	 */
+	bool show_topbottom = width > inset * 2;
+	bool show_sides = height > inset * 2;
+	/* These values are the overlap on each corner (half total overlap) */
+	int horizontal_overlap = MAX(0, inset - width / 2);
+	int vertical_overlap = MAX(0, inset - height / 2);
+
+	/*
+	 * If window width or height is odd then making the corners equally
+	 * sized when the edge is hidden would leave a single pixel gap
+	 * between the corners. Showing a single pixel edge between clipped
+	 * corners looks bad because the edge-piece doesn't match up with the
+	 * corners after the corners are clipped. So fill the gap by making
+	 * the top-left and bottom-right corners one pixel wider (if the width
+	 * is odd) or taller (if the height is odd).
+	 */
+	if (part->type == LAB_SSD_PART_CORNER_TOP_LEFT
+			|| part->type == LAB_SSD_PART_CORNER_BOTTOM_RIGHT) {
+		if (horizontal_overlap > 0) {
+			horizontal_overlap -= width % 2;
+		}
+		if (vertical_overlap > 0) {
+			vertical_overlap -= height % 2;
+		}
+	}
+
+	int x;
+	int y;
+
+	switch (part->type) {
+	case LAB_SSD_PART_CORNER_BOTTOM_RIGHT:
+		x = width - inset + horizontal_overlap;
+		y = -titlebar_height + height - inset + vertical_overlap;
+		wlr_scene_node_set_position(part->node, x, y);
+		corner_scale_crop(scene_buf, horizontal_overlap,
+			vertical_overlap, corner_size, false);
+		break;
+	case LAB_SSD_PART_CORNER_BOTTOM_LEFT:
+		x = -visible_shadow_width;
+		y = -titlebar_height + height - inset + vertical_overlap;
+		wlr_scene_node_set_position(part->node, x, y);
+		corner_scale_crop(scene_buf, horizontal_overlap,
+			vertical_overlap, corner_size, true);
+		break;
+	case LAB_SSD_PART_CORNER_TOP_LEFT:
+		x = -visible_shadow_width;
+		y = -titlebar_height - visible_shadow_width;
+		wlr_scene_node_set_position(part->node, x, y);
+		corner_scale_crop(scene_buf, horizontal_overlap,
+			vertical_overlap, corner_size, false);
+		break;
+	case LAB_SSD_PART_CORNER_TOP_RIGHT:
+		x = width - inset + horizontal_overlap;
+		y = -titlebar_height - visible_shadow_width;
+		wlr_scene_node_set_position(part->node, x, y);
+		corner_scale_crop(scene_buf, horizontal_overlap,
+			vertical_overlap, corner_size, true);
+		break;
+	case LAB_SSD_PART_RIGHT:
+		x = width;
+		y = -titlebar_height + inset;
+		wlr_scene_node_set_position(part->node, x, y);
+		wlr_scene_buffer_set_dest_size(
+			scene_buf, visible_shadow_width, height - 2 * inset);
+		wlr_scene_node_set_enabled(part->node, show_sides);
+		break;
+	case LAB_SSD_PART_BOTTOM:
+		x = inset;
+		y = -titlebar_height + height;
+		wlr_scene_node_set_position(part->node, x, y);
+		wlr_scene_buffer_set_dest_size(
+			scene_buf, width - 2 * inset, visible_shadow_width);
+		wlr_scene_node_set_enabled(part->node, show_topbottom);
+		break;
+	case LAB_SSD_PART_LEFT:
+		x = -visible_shadow_width;
+		y = -titlebar_height + inset;
+		wlr_scene_node_set_position(part->node, x, y);
+		wlr_scene_buffer_set_dest_size(
+			scene_buf, visible_shadow_width, height - 2 * inset);
+		wlr_scene_node_set_enabled(part->node, show_sides);
+		break;
+	case LAB_SSD_PART_TOP:
+		x = inset;
+		y = -titlebar_height - visible_shadow_width;
+		wlr_scene_node_set_position(part->node, x, y);
+		wlr_scene_buffer_set_dest_size(
+			scene_buf, width - 2 * inset, visible_shadow_width);
+		wlr_scene_node_set_enabled(part->node, show_topbottom);
+		break;
+	default:
+		break;
+	}
+}
+
+static void
+set_shadow_geometry(struct ssd *ssd)
+{
+	struct view *view = ssd->view;
+	struct theme *theme = view->server->theme;
+	int titlebar_height = ssd->titlebar.height;
+	int width = view->current.width;
+	int height = view_effective_height(view, false) + titlebar_height;
+
+	struct ssd_part *part;
+	struct ssd_sub_tree *subtree;
+
+	FOR_EACH_STATE(ssd, subtree) {
+		if (!subtree->tree) {
+			/* Looks like this type of shadow is disabled */
+			continue;
+		}
+
+		bool active = subtree == &ssd->shadow.active;
+		int visible_shadow_width = active
+			? theme->window_active_shadow_size
+			: theme->window_inactive_shadow_size;
+		/* inset as a proportion of shadow width */
+		double inset_proportion = SSD_SHADOW_INSET;
+		/* inset in actual pixels */
+		int inset = inset_proportion * (double)visible_shadow_width;
+
+		/*
+		 * Total size of corner buffers including inset and visible
+		 * portion.  Top and bottom are the same size (only the cutout
+		 * is different).  The buffers are square so width == height.
+		 */
+		int corner_size = active
+			? theme->shadow_corner_top_active->unscaled_height
+			: theme->shadow_corner_top_inactive->unscaled_height;
+
+		wl_list_for_each(part, &subtree->parts, link) {
+			set_shadow_part_geometry(part, width, height,
+				titlebar_height, corner_size, inset,
+				visible_shadow_width);
+		}
+	} FOR_EACH_END
+}
+
+static void
+make_shadow(struct wl_list *parts, enum ssd_part_type type,
+	struct wlr_scene_tree *parent, struct wlr_buffer *buf,
+	enum wl_output_transform tx)
+{
+	struct ssd_part *part = add_scene_buffer(
+		parts, type, parent, buf, 0, 0);
+	struct wlr_scene_buffer *scene_buf =
+		wlr_scene_buffer_from_node(part->node);
+	wlr_scene_buffer_set_transform(scene_buf, tx);
+	scene_buf->point_accepts_input = never_accepts_input;
+	/*
+	 * Pixman has odd behaviour with bilinear filtering on buffers only one
+	 * pixel wide/tall. Use nearest-neighbour scaling to workaround.
+	 */
+	scene_buf->filter_mode = WLR_SCALE_FILTER_NEAREST;
+}
+
+void
+ssd_shadow_create(struct ssd *ssd)
+{
+	assert(ssd);
+	assert(!ssd->shadow.tree);
+
+	ssd->shadow.tree = wlr_scene_tree_create(ssd->tree);
+
+	struct theme *theme = ssd->view->server->theme;
+	struct wlr_buffer *corner_top_buffer;
+	struct wlr_buffer *corner_bottom_buffer;
+	struct wlr_buffer *edge_buffer;
+	struct ssd_sub_tree *subtree;
+	struct wlr_scene_tree *parent;
+
+	FOR_EACH_STATE(ssd, subtree) {
+		if (subtree == &ssd->shadow.active) {
+			if (theme->window_active_shadow_size == 0) {
+				/* Active window shadows are disabled */
+				continue;
+			}
+		} else {
+			if (theme->window_inactive_shadow_size == 0) {
+				/* Inactive window shadows are disabled */
+				continue;
+			}
+		}
+
+		subtree->tree = wlr_scene_tree_create(ssd->shadow.tree);
+		parent = subtree->tree;
+		if (subtree == &ssd->shadow.active) {
+			corner_top_buffer = &theme->shadow_corner_top_active->base;
+			corner_bottom_buffer = &theme->shadow_corner_bottom_active->base;
+			edge_buffer = &theme->shadow_edge_active->base;
+		} else {
+			corner_top_buffer = &theme->shadow_corner_top_inactive->base;
+			corner_bottom_buffer = &theme->shadow_corner_bottom_inactive->base;
+			edge_buffer = &theme->shadow_edge_inactive->base;
+		}
+		wl_list_init(&subtree->parts);
+
+		make_shadow(&subtree->parts,
+			LAB_SSD_PART_CORNER_BOTTOM_RIGHT, parent,
+			corner_bottom_buffer, WL_OUTPUT_TRANSFORM_NORMAL);
+		make_shadow(&subtree->parts, LAB_SSD_PART_CORNER_BOTTOM_LEFT,
+			parent, corner_bottom_buffer, WL_OUTPUT_TRANSFORM_90);
+		make_shadow(&subtree->parts, LAB_SSD_PART_CORNER_TOP_LEFT,
+			parent, corner_top_buffer, WL_OUTPUT_TRANSFORM_180);
+		make_shadow(&subtree->parts, LAB_SSD_PART_CORNER_TOP_RIGHT,
+			parent, corner_top_buffer, WL_OUTPUT_TRANSFORM_270);
+		make_shadow(&subtree->parts, LAB_SSD_PART_RIGHT, parent,
+			edge_buffer, WL_OUTPUT_TRANSFORM_NORMAL);
+		make_shadow(&subtree->parts, LAB_SSD_PART_BOTTOM, parent,
+			edge_buffer, WL_OUTPUT_TRANSFORM_90);
+		make_shadow(&subtree->parts, LAB_SSD_PART_LEFT, parent,
+			edge_buffer, WL_OUTPUT_TRANSFORM_180);
+		make_shadow(&subtree->parts, LAB_SSD_PART_TOP, parent,
+			edge_buffer, WL_OUTPUT_TRANSFORM_270);
+
+	} FOR_EACH_END
+
+	ssd_shadow_update(ssd);
+}
+
+void
+ssd_shadow_update(struct ssd *ssd)
+{
+	assert(ssd);
+	assert(ssd->shadow.tree);
+
+	struct view *view = ssd->view;
+	bool maximized = view->maximized == VIEW_AXIS_BOTH;
+	bool show_shadows =
+		rc.shadows_enabled && !maximized && !view_is_tiled(ssd->view);
+	wlr_scene_node_set_enabled(&ssd->shadow.tree->node, show_shadows);
+	if (show_shadows) {
+		set_shadow_geometry(ssd);
+	}
+}
+
+void
+ssd_shadow_destroy(struct ssd *ssd)
+{
+	assert(ssd);
+	assert(ssd->shadow.tree);
+
+	wlr_scene_node_destroy(&ssd->shadow.tree->node);
+	ssd->shadow.tree = NULL;
+}
--- a/src/theme.c
+++ b/src/theme.c
@@ -25,6 +25,7 @@
 #include "common/match.h"
 #include "common/mem.h"
 #include "common/parse-bool.h"
+#include "common/parse-double.h"
 #include "common/string-helpers.h"
 #include "config/rcxml.h"
 #include "button/button-png.h"
@@ -500,6 +501,11 @@ theme_builtin(struct theme *theme)
 	parse_hexstr("#000000",
 		theme->window_inactive_button_close_unpressed_image_color);
 
+	theme->window_active_shadow_size = 60;
+	theme->window_inactive_shadow_size = 40;
+	parse_hexstr("#00000060", theme->window_active_shadow_color);
+	parse_hexstr("#00000040", theme->window_inactive_shadow_color);
+
 	parse_hexstr("#fcfbfa", theme->menu_items_bg_color);
 	parse_hexstr("#000000", theme->menu_items_text_color);
 	parse_hexstr("#e1dedb", theme->menu_items_active_bg_color);
@@ -675,6 +681,30 @@ entry(struct theme *theme, const char *key, const char *value)
 	if (match_glob(key, "window.inactive.button.close.unpressed.image.color")) {
 		parse_hexstr(value,
 			theme->window_inactive_button_close_unpressed_image_color);
+	}
+
+	/* window drop-shadows */
+	if (match_glob(key, "window.active.shadow.size")) {
+		theme->window_active_shadow_size = atoi(value);
+		if (theme->window_active_shadow_size < 0) {
+			wlr_log(WLR_ERROR, "window.active.shadow.size cannot "
+				"be negative, clamping it to 0.");
+			theme->window_active_shadow_size = 0;
+		}
+	}
+	if (match_glob(key, "window.inactive.shadow.size")) {
+		theme->window_inactive_shadow_size = atoi(value);
+		if (theme->window_inactive_shadow_size < 0) {
+			wlr_log(WLR_ERROR, "window.inactive.shadow.size cannot "
+				"be negative, clamping it to 0.");
+			theme->window_inactive_shadow_size = 0;
+		}
+	}
+	if (match_glob(key, "window.active.shadow.color")) {
+		parse_hexstr(value, theme->window_active_shadow_color);
+	}
+	if (match_glob(key, "window.inactive.shadow.color")) {
+		parse_hexstr(value, theme->window_inactive_shadow_color);
 	}
 
 	if (match_glob(key, "menu.width.min")) {
@@ -1035,6 +1065,185 @@ create_corners(struct theme *theme)
 	theme->corner_top_right_inactive_normal = rounded_rect(&ctx);
 }
 
+/*
+ * Draw the buffer used to render the edges of window drop-shadows. The buffer
+ * is 1 pixel tall and `visible_size` pixels wide and can be rotated and scaled for the
+ * different edges.  The buffer is drawn as would be found at the right-hand
+ * edge of a window. The gradient has a color of `start_color` at its left edge
+ * fading to clear at its right edge.
+ */
+static void
+shadow_edge_gradient(struct lab_data_buffer *buffer,
+		int visible_size, int total_size, float start_color[4])
+{
+	if (!buffer) {
+		/* This type of shadow is disabled, do nothing */
+		return;
+	}
+
+	assert(buffer->format == DRM_FORMAT_ARGB8888);
+	uint8_t *pixels = buffer->data;
+
+	/* Inset portion which is obscured */
+	int inset = total_size - visible_size;
+
+	/* Standard deviation normalised against the shadow width, squared */
+	double variance = 0.3 * 0.3;
+
+	for (int x = 0; x < visible_size; x++) {
+		/*
+		 * x normalised against total shadow width. We add on inset here
+		 * because we don't bother drawing inset for the edge shadow
+		 * buffers but still need the pattern to line up with the corner
+		 * shadow buffers which do have inset drawn.
+		 */
+		double xn = (double)(x + inset) / (double)total_size;
+
+		/* Gausian dropoff */
+		double alpha = exp(-(xn * xn) / variance);
+
+		/* RGBA values are all pre-multiplied */
+		pixels[4 * x] = start_color[2] * alpha * 255;
+		pixels[4 * x + 1] = start_color[1] * alpha * 255;
+		pixels[4 * x + 2] = start_color[0] * alpha * 255;
+		pixels[4 * x + 3] = start_color[3] * alpha * 255;
+	}
+}
+
+/*
+ * Draw the buffer used to render the corners of window drop-shadows.  The
+ * shadow looks better if the buffer is inset behind the window, so the buffer
+ * is square with a size of radius+inset.  The buffer is drawn for the
+ * bottom-right corner but can be rotated for other corners.  The gradient fades
+ * from `start_color` at the top-left to clear at the opposite edge.
+ *
+ * If the window is translucent we don't want the shadow to be visible through
+ * it.  For the bottom corners of the window this is easy, we just erase the
+ * square of the buffer which will be behind the window.  For the top it's a
+ * little more complicated because the titlebar can have rounded corners.
+ * However, the titlebar itself is always opaque so we only have to erase the
+ * L-shaped area of the buffer which can appear behind the non-titlebar part of
+ * the window.
+ */
+static void
+shadow_corner_gradient(struct lab_data_buffer *buffer, int visible_size,
+	int total_size, int titlebar_height, float start_color[4])
+{
+	if (!buffer) {
+		/* This type of shadow is disabled, do nothing */
+		return;
+	}
+
+	assert(buffer->format == DRM_FORMAT_ARGB8888);
+	uint8_t *pixels = buffer->data;
+
+	/* Standard deviation normalised against the shadow width, squared */
+	double variance = 0.3 * 0.3;
+
+	int inset = total_size - visible_size;
+
+	for (int y = 0; y < total_size; y++) {
+		uint8_t *pixel_row = &pixels[y * buffer->stride];
+		for (int x = 0; x < total_size; x++) {
+			/* x and y normalised against total shadow width */
+			double x_norm = (double)(x) / (double)total_size;
+			double y_norm = (double)(y) / (double)total_size;
+			/*
+			 * For Gaussian drop-off in 2d you can just calculate
+			 * the outer product of the horizontal and vertical
+			 * profiles.
+			 */
+			double gauss_x = exp(-(x_norm * x_norm) / variance);
+			double gauss_y = exp(-(y_norm * y_norm) / variance);
+			double alpha = gauss_x * gauss_y;
+
+			/*
+			 * Erase the L-shaped region which could be visible
+			 * through a transparent window but not obscured by the
+			 * titlebar. If inset is smaller than the titlebar
+			 * height then there's nothing to do, this is handled by
+			 * (inset - titlebar_height) being negative.
+			 */
+			bool in1 = x < inset && y < inset - titlebar_height;
+			bool in2 = x < inset - titlebar_height && y < inset;
+			if (in1 || in2) {
+				alpha = 0.0;
+			}
+
+			/* RGBA values are all pre-multiplied */
+			pixel_row[4 * x] = start_color[2] * alpha * 255;
+			pixel_row[4 * x + 1] = start_color[1] * alpha * 255;
+			pixel_row[4 * x + 2] = start_color[0] * alpha * 255;
+			pixel_row[4 * x + 3] = start_color[3] * alpha * 255;
+		}
+	}
+}
+
+static void
+create_shadows(struct theme *theme)
+{
+	/* Size of shadow visible extending beyond the window */
+	int visible_active_size = theme->window_active_shadow_size;
+	int visible_inactive_size = theme->window_inactive_shadow_size;
+	/* How far inside the window the shadow inset begins */
+	int inset_active = (double)visible_active_size * SSD_SHADOW_INSET;
+	int inset_inactive = (double)visible_inactive_size * SSD_SHADOW_INSET;
+	/* Total width including visible and obscured portion */
+	int total_active_size = visible_active_size + inset_active;
+	int total_inactive_size = visible_inactive_size + inset_inactive;
+
+	/*
+	 * Edge shadows don't need to be inset so the buffers are sized just for
+	 * the visible width.  Corners are inset so the buffers are larger for
+	 * this.
+	 */
+	if (visible_active_size > 0) {
+		theme->shadow_edge_active = buffer_create_cairo(
+			visible_active_size, 1, 1.0, true);
+		theme->shadow_corner_top_active = buffer_create_cairo(
+			total_active_size, total_active_size, 1.0, true);
+		theme->shadow_corner_bottom_active = buffer_create_cairo(
+			total_active_size, total_active_size, 1.0, true);
+		if (!theme->shadow_corner_top_active
+				|| !theme->shadow_corner_bottom_active
+				|| !theme->shadow_edge_active) {
+			wlr_log(WLR_ERROR, "Failed to allocate shadow buffer");
+			return;
+		}
+	}
+	if (visible_inactive_size > 0) {
+		theme->shadow_edge_inactive = buffer_create_cairo(
+			visible_inactive_size, 1, 1.0, true);
+		theme->shadow_corner_top_inactive = buffer_create_cairo(
+			total_inactive_size, total_inactive_size, 1.0, true);
+		theme->shadow_corner_bottom_inactive = buffer_create_cairo(
+			total_inactive_size, total_inactive_size, 1.0, true);
+		if (!theme->shadow_corner_top_inactive
+				|| !theme->shadow_corner_bottom_inactive
+				|| !theme->shadow_edge_inactive) {
+			wlr_log(WLR_ERROR, "Failed to allocate shadow buffer");
+			return;
+		}
+	}
+
+	shadow_edge_gradient(theme->shadow_edge_active, visible_active_size,
+		total_active_size, theme->window_active_shadow_color);
+	shadow_edge_gradient(theme->shadow_edge_inactive, visible_inactive_size,
+		total_inactive_size, theme->window_inactive_shadow_color);
+	shadow_corner_gradient(theme->shadow_corner_top_active,
+		visible_active_size, total_active_size,
+		theme->title_height, theme->window_active_shadow_color);
+	shadow_corner_gradient(theme->shadow_corner_bottom_active,
+		visible_active_size, total_active_size, 0,
+		theme->window_active_shadow_color);
+	shadow_corner_gradient(theme->shadow_corner_top_inactive,
+		visible_inactive_size, total_inactive_size,
+		theme->title_height, theme->window_inactive_shadow_color);
+	shadow_corner_gradient(theme->shadow_corner_bottom_inactive,
+		visible_inactive_size, total_inactive_size, 0,
+		theme->window_inactive_shadow_color);
+}
+
 static void
 fill_colors_with_osd_theme(struct theme *theme, float colors[3][4])
 {
@@ -1154,6 +1363,7 @@ theme_init(struct theme *theme, const char *theme_name)
 	post_processing(theme);
 	create_corners(theme);
 	load_buttons(theme);
+	create_shadows(theme);
 }
 
 void
@@ -1163,4 +1373,10 @@ theme_finish(struct theme *theme)
 	zdrop(&theme->corner_top_left_inactive_normal);
 	zdrop(&theme->corner_top_right_active_normal);
 	zdrop(&theme->corner_top_right_inactive_normal);
+	zdrop(&theme->shadow_corner_top_active);
+	zdrop(&theme->shadow_corner_bottom_active);
+	zdrop(&theme->shadow_edge_active);
+	zdrop(&theme->shadow_corner_top_inactive);
+	zdrop(&theme->shadow_corner_bottom_inactive);
+	zdrop(&theme->shadow_edge_inactive);
 }
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -401,6 +401,9 @@ extending outward from the snapped edge.
 	Even when disabling server side decorations via ToggleDecorations,
 	keep a small border (and resize area) around the window. Default is yes.
 
+*<theme><dropShadows>* [yes|no]
+	Should drop-shadows be rendered behind windows. Default is no.
+
 *<theme><font place="">*
 	The font to use for a specific element of a window, menu or OSD.
 	Places can be any of:
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -24,6 +24,7 @@
     <name></name>
     <cornerRadius>8</cornerRadius>
     <keepBorder>yes</keepBorder>
+    <dropShadows>no</dropShadows>
     <font place="ActiveWindow">
       <name>sans</name>
       <size>10</size>
--- a/docs/themerc
+++ b/docs/themerc
@@ -33,6 +33,12 @@ window.label.text.justify: center
 window.active.button.unpressed.image.color: #000000
 window.inactive.button.unpressed.image.color: #000000
 
+# window drop-shadows
+window.active.shadow.size: 60
+window.inactive.shadow.size: 40
+window.active.shadow.color: #00000060
+window.inactive.shadow.color: #00000040
+
 # Note that "menu", "iconify", "max", "close" buttons colors can be defined
 # individually by inserting the type after the button node, for example:
 #
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -64,10 +64,12 @@ struct rcxml {
 	char *theme_name;
 	int corner_radius;
 	bool ssd_keep_border;
+	bool shadows_enabled;
 	struct font font_activewindow;
 	struct font font_inactivewindow;
 	struct font font_menuitem;
 	struct font font_osd;
+
 	/* Pointer to current theme */
 	struct theme *theme;
 
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -76,6 +76,12 @@ struct ssd {
 		struct ssd_sub_tree inactive;
 	} border;
 
+	struct {
+		struct wlr_scene_tree *tree;
+		struct ssd_sub_tree active;
+		struct ssd_sub_tree inactive;
+	} shadow;
+
 	/*
 	 * Space between the extremities of the view's wlr_surface
 	 * and the max extents of the server-side decorations.
@@ -149,5 +155,9 @@ void ssd_border_destroy(struct ssd *ssd);
 void ssd_extents_create(struct ssd *ssd);
 void ssd_extents_update(struct ssd *ssd);
 void ssd_extents_destroy(struct ssd *ssd);
+
+void ssd_shadow_create(struct ssd *ssd);
+void ssd_shadow_update(struct ssd *ssd);
+void ssd_shadow_destroy(struct ssd *ssd);
 
 #endif /* LABWC_SSD_INTERNAL_H */
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -10,6 +10,13 @@
 #define SSD_EXTENDED_AREA 8
 
 /*
+ * Shadows should start at a point inset from the actual window border, see
+ * discussion on https://github.com/labwc/labwc/pull/1648.  This constant
+ * specifies inset as a multiple of visible shadow size.
+ */
+#define SSD_SHADOW_INSET 0.3
+
+/*
  * Sequence these according to the order they should be processed for
  * press and hover events. Bear in mind that some of their respective
  * interactive areas overlap, so for example buttons need to come before title.
--- a/include/theme.h
+++ b/include/theme.h
@@ -93,6 +93,12 @@ struct theme {
 	float snapping_preview_region_border_color[3][4];
 	float snapping_preview_edge_border_color[3][4];
 
+	/* window drop-shadows */
+	int window_active_shadow_size;
+	int window_inactive_shadow_size;
+	float window_active_shadow_color[4];
+	float window_inactive_shadow_color[4];
+
 	/* textures */
 	struct lab_data_buffer *button_close_active_unpressed;
 	struct lab_data_buffer *button_maximize_active_unpressed;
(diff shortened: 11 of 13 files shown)


src/config/rcxml.c
```diff
@@ -708,6 +714,20 @@ enum_font_place(const char *place)
 	return FONT_PLACE_UNKNOWN;
 }
 
+static enum dropshadow_type
+enum_dropshadow_type(const char *type)
+{
+	if (!type || type[0] == '\0') {
```
Consolatis, line 720: Suggest `string_null_or_empty()` here (`common/string-helpers.h`)

Consolatis:
First, thanks for your PR!
The implementation looks great on the first glance.

Regarding the feature itself, we need to discuss this among the core devs.
My personal opinion is that this will create two small cairo buffers on reconfigure but that's about it, the added code is also nicely contained within `src/ssd/ssd_shadow.c` so maintenance overhead seems very minimal.

We may want to split the config part a bit by moving the shadow settings to `src/theme.c` and just keep a single generic shadows yes / no in rc.xml.

johanmalm:
I like this. Nice implementation.

My top two considerations:

1. Where we put the config options? My preference would be `themerc` because (a) it just feels like it is part of the theme; (b) it makes it easier in theme engines like `labwc-gtktheme` to just write to one simple file rather than processing XML; and (c) it allows globbing so we can do `window.*.drop-shadow.foo` to cover both active and inactive in the case that they are the same. I welcome thoughts/considerations from others. For reference, we do support `themerc-override` (in $XDG_CONFIG_HOME/DIRS) so if not specified in exising themes, there is still a way to configure it on par with `rc.xml`

2. Build confidence that we're choosing the right approach for the settings - both what we call them and what they do. `radius` seems to make sense, but I'm not sure about `inset`. 

2a. I had a quick look at [picom.conf](https://github.com/yshui/picom/blob/next/picom.sample.conf). They use radius and offset, but by default the magnitude of their offset is the greater than the radius, so I'm guessing that they mean something different to our `inset` and `radius`.

2b. GTK themes seems to use a start- and finish-color, so probably work around the inset/offset that way IIUC. For example, from deepin:

`box-shadow: 0 0 0 1px rgb(1,147,240), 0 1px 2px rgba(0,0,0,0.3);`

2c. The corner radius becomes the same as the gradient radius. I'm guessing that's the most sane way of doing this... but for me that means I can' adjust the corner radius independingly of the color gradient (because they're both a function of the inset). I'm not sure if I'm explaining that very well, but hopefully you get what I mean. For example if I want a fairly light color, I have to set inset to a big number (e.g. 0.7 of radius) which means that the corner radius if very big. I suppose having a start and finish color is a way of solving that.

I'm all for starting small (small number of settings and so on), but the reason for trying to think this through in some detail is to avoid having to change them later and break backward-compatibility.

Nits:

- [ ] Add section in `labwc-theme.5.scd`
- [ ] We handle `maximized` in `ssd_shadow_create/update()`. Do we need to do anything for fullscreen?

Other considerations (most likely for the future):

- Consider constraining drop-shadows to usable-area to that they don't spill over onto panels (in background layer). I'm not sure what's visually right here, but for consideration nonetheless.
- Window rule for exluding clients from shadows
- Shadows for popups.

johanmalm:
```
radius=50
inset=25
opacity=1.0
```


20240321_20h02m56s_grim

tsujan:
Nice!

As a tester, I found two problems that I think may have the same cause.

First, it's drawn behind the window too. Since on Wayland windows can be translucent, it can look unnatural:

shadow1

Second, when I shade a window, this happens:

shadow2

jlindgren90:
I haven't looked at the code, but I have no issues with this in principle, if the issues that have been mentioned already are fixed.

Consolatis:
> First, it's drawn behind the window too. Since on Wayland windows can be translucent, it can look unnatural:

I think that could be fixed with `wlr_scene_buffer_set_source_box()`.

stefonarch:
Could it be that any value for `opacity` below 1.0 disables shadows or it's that just me?
I tried with the default values from `rc.xml.all` first.

Really nice feature btw.

stefonarch:
> I've given it a test run and it looks quite nice, but I've found some issues when hiding the decorations or shading a window.

If a window is shaded a ghost is remaining, but when the bar is moved it is corrected:

screen_area_ven_08:42:28_

But when unshaded no way, window needs restart:

screen_area_ven_08:43:06_

For toggling decoration the ghost of the titlebar remains always, toggling all decorations works fine.

cillian64:
Thanks all for the testing/feedback/comments, lots of things I didn't think of!

It sounds like the main issues are the following, I'll work through these over the next few days.

- [ ] Move the shadow config to theme
- [ ] Issues with shading
- [ ] Issues when toggling window decoration
- [ ] Issues with transparent/translucent windows (I can't off the top of my head think of a simple strategy which works for translucent windows with rounded corners, if translucent+rounded+shadows is a requirement it might bump up the complexity.  To start with I'll try just handling translucent rectangular windows)
- [ ] Consider how shadows are configured (agreed that the current radius/inset/opacity is awkward, finding values which look reasonable was a real pain.  Start-opacity and corner radius do sound more sensible.  I'll have a play around with this.)
- [ ] `string_null_or_empty()` in config
- [ ] Update `labwc-theme.5.scd`
- [ ] Check fullscreen works like maximized

Responding to some other points:

> Shadows for popups.

I wasn't sure what to do about this.  Wayfire and GNOME don't seem to do shadows on popups/menus so I decided to copy that behaviour (also because it let me easily piggy-back on the existing SSD code!).  Windows 10 seems to have an extremely subtle shadow on dropdown menus and the start menu but it's hard to tell.

> Could it be that any value for opacity below 1.0 disables shadows or it's that just me?

I've done most of my testing with opacity less than 1.0 so I'm not sure what's going wrong there.  If you could share the rest of the config I'll try to reproduce it.

tsujan:
Just to be thorough, I would like to add that there's also another look-and-feel issue with a zero or low value of "inset", apart from that of rounded corners. The shadow doesn't look natural at the corners; it seems intensified:

shadow4

A relatively large inset would be a workaround (as in @johanmalm's [screenshot](https://github.com/labwc/labwc/pull/1648#issuecomment-2013602598)), but it would cause a problem with translucent windows, as mentioned above.

tsujan:
> Wayfire and GNOME don't seem to do shadows on popups/menus

IMHO, that's a correct approach. Even if there's a shadow for popups/menus, it should be optional. There are at least two reasons for that:

 1. A Menu may have a rounded corner, whose roundness is determined by the Qt/GTK style, not by labwc; and
 2. Some widget styles (like Kvantum) can have their own menu shadows, in which case, we'll have double shadows.

stefonarch:
>I've done most of my testing with opacity less than 1.0 so I'm not sure what's going wrong there. If you could share the rest of the config I'll try to reproduce it.

I removed most but still it doesn't work. Here active windows have no shadow with attached config while inactive do. Theme doesn't matter too.

[themerc.tar.gz](https://github.com/labwc/labwc/files/14723194/themerc.tar.gz)

tsujan:
Strange! I can confirm that @stefonarch's settings result in active windows without shadows, while my settings work without problem:

```xml
    <dropShadow type="active">
      <enabled>yes</enabled>
      <radius>30</radius>
      <opacity>0.5</opacity>
      <inset>2</inset>
    </dropShadow>
    <dropShadow type="inactive">
      <enabled>yes</enabled>
      <radius>20</radius>
      <opacity>0.5</opacity>
      <inset>2</inset>
    </dropShadow>
```

Consolatis:
> It sounds like the main issues are the following, I'll work through these over the next few days.
> [..]
> Responding to some other points:
>
> > Shadows for popups.
> 
> I wasn't sure what to do about this. Wayfire and GNOME don't seem to do shadows on popups/menus so I decided to copy that behaviour (also because it let me easily piggy-back on the existing SSD code!). Windows 10 seems to have an extremely subtle shadow on dropdown menus and the start menu but it's hard to tell.

Yeah, lets not do shadows for popups for now, it complicates everything for a minor benefit but has lots of external undefined variables (e.g. rounded corners set by clients, own client side shadows, some toolkits seem to use popups for tooltips as well as for actual menu popups and so on, other toolkits seem to use subsurfaces for actual menu popups that would then not get a shadow, etc. etc).

tsujan:
BTW, I applied the patch to 0.7.1. It's both usable and enjoyable (the issues are quite tolerable). @cillian64, thanks a lot for it!

cillian64:
> Just to be thorough, I would like to add that there's also another look-and-feel issue with a zero or low value of "inset", apart from that of rounded corners.
> A relatively large inset would be a workaround (as in @johanmalm's [screenshot](https://github.com/labwc/labwc/pull/1648#issuecomment-2013602598)), but it would cause a problem with translucent windows, as mentioned above.

Right, I'm planning to solve this by having a better way to get good-looking shadow corners without needing to inset them behind the window (and with a less awkward config parameter than "inset").

I just tested with @stefonarch 's config on a white background and the active window does have a shadow, just a very subtle one 
shadows

The issue is that opacity and inset interact: opacity sets how dark the shadow is at its inner edge, but when inset is a large value you don't actually see that inner edge because it's behind the window, so increasing inset implicitly decreases opacity.  I agree this is unhelpful and this is why I'm going to change the config so you separately set the shadow inner corner radius and how dark it is at the window edge.

stefonarch:
It's  not subtle here, it's absent, when I shade the window with 0.9 I should see something:

screen_area_lun_14:25:49_

cillian64:
@stefonarch could you try with https://github.com/cillian64/labwc/tree/drop-shadows-debug and see what the logs print and attach `/tmp/{corner,edge}_{active,inactive}.png` ?

stefonarch:
```
stef@archlinux:build$ ./labwc -s lxqt-panel
amdgpu: amdgpu_cs_ctx_create2 failed. (-13)
00:00:00.245 [../src/theme.c:1031] Drawing active shadows 40 15 0,000000 1
00:00:00.245 [../src/theme.c:1036] Drawing inactive shadows 40 14 1,000000 1
StatusNotifierProxy, services: QList()
new toplevel created
new toplevel created
new toplevel created
new toplevel created
new toplevel created

```

So nothing really new..., before changing opacity I got 
```
00:00:00.252 [../src/theme.c:1031] Drawing active shadows 40 15 1,000000 1                                                            
00:00:00.252 [../src/theme.c:1036] Drawing inactive shadows 40 14 1,000000 1                ```

![screen_area_mer_07:57:40_](https://github.com/labwc/labwc/assets/10681413/c2f4619d-5f73-4313-b5ff-49e4cbd3a162)

![screenshot_2024-03-27_07 45 56](https://github.com/labwc/labwc/assets/10681413/8d59c718-92b5-44ab-81e7-d00d06885672)

**Comment by cillian64** [comment]
Aha, that explains it! Looks like your locale uses `,` instead of `.` for decimal separators, and (very unexpectedly to me) `atof` respects locale here.  It should work if you use `,` for decimal separators in opacity.

I think this is very wrong and the config/theme file syntax should *not* be locale dependant.  But a bunch of existing config options already use `atof` so I think this would be best addressed as a separate issue.

**Comment by 01micko** [comment]
`,` (comma) is invalid in XML so file won't be parsed so this is a bug in labwc that has to be overcome by setting locale to `C` in all source files that use `atof()` or `strtod()` or similar apparently.

**Comment by stefonarch** [comment]
Can confirm that `<opacity>0,8</opacity>` works...

**Comment by droc12345** [comment]
You could use setlocale in conjunction with atof to standardize.

**Comment by ahesford** [comment]
I think it's better to write the parsing to avoid floats, so using an integer percent seems like a reasonable option.

If we don't use percent, it's better to respect locale than to force people to write numbers that don't conform to their intuition. Furthermore, if we are going to adjust the locale, we had better be sure to set it back afterwards.

**Comment by cillian64** [comment]
I've made a new issue #1665 for the locale-dependent-floats issue, please could all discussion on that topic happen there.  This issue is not new in my patch and I'm not going to drive-by fix the existing cases in this patch.

**Comment by Consolatis** [comment]
> I've made a new issue #1665 for the locale-dependent-floats issue, please could all discussion on that topic happen there. This issue is not new in my patch and I'm not going to drive-by fix the existing cases in this patch.

Cleaned up the comments a bit, thanks for debuggingt and creating the issue.

**Comment by ahesford** [comment]
Alternatively, a number 0-255, as I suspect people dealing with colors and transparency are somewhat used to dealing with 8-bit ints.

**Comment by tsujan** [comment]
> Alternatively, a number 0-255

I also wanted to suggest that the issue would be avoided in this way.

**Comment by Consolatis** [comment]
Lets not make that a factor here. Labwc needs to fix its float parsing, the decision to use percent or 0-255 or floats should not be influenced by it.

**Comment by cillian64** [comment]
I've added some more commits with my work so far (I'll squash and tidy when finished).  There are some places where the code could be tidied but I'd like to get the rough shape of things figured out before fine-tuning.

I think there's still room to make the configuration/theming more intuitive.  The reason this looks different to other shadow configurations is I'm not doing a real Gaussian blur, I'm faking it with an `x^2` drop-off.  The outer edge of the shadow is pretty similar but the inner part looks quite different which is why it's important to hide it with the inset.

I've changed `radius` to apply to the visible portion of the shadow (excluding `inset`) and `color` now applies from the window-edge so isn't affected by `inset`.  I think this solves the biggest issues with configuration.  I'm considering if it's better to remove the `inset` parameter entirely and hard-code it at 0.5: I feel that `radius` and `color` provide enough flexibility and there's a pretty narrow range of `inset` values which look at all decent.

I'm not quite convinced if it's better to have shadow colour vs just opacity.  I couldn't find any non-black options which I liked, it would be useful to hear if anyone else can find a coloured shadow they like more than black.

**Comment by Consolatis** [comment]
Can't really say anything about the feature itself, I am really not the target audience for it.

However, I noticed that `src/theme.c` doesn't set the default values in `theme_builtin()` which is required for proper `Reconfigure` support (if the whole line is removed / commented out or the theme is changed via `rc.xml`).

**Comment by tsujan** [comment]
I confirm that the issues I reported have been fixed. The new code results in a very nice shadow.

The only issue I can see is what @Flrian reported in the above comment. It doesn't depend on activeness/inactiveness though:

shaded

File: docs/labwc-theme.5.scd [hunk 2]
```diff
@@ -126,6 +126,32 @@ This syntax is not documented on the openbox.org wiki, but is supported by
 openbox and is used by many popular themes. For the sake of brevity, these
 elements are not listed here, but are supported.
 
+*window.active.shadow.enabled*
```
stefonarch, line 129: This sounds strange, conditions in a theme. Couldn't those be replaced simply by a check for  presence or not of `window.*.shadow.*` ?

cillian64: Sure, can avoid having the boolean in the theme if that would be weird.

@Consolatis suggested putting the on/off in rc.xml and having the other settings in themerc.  To me it feels a bit odd to have them in separate files, but I'm not very familiar with labwc config so I will follow other people's lead here.

cillian64:
> The only issue I can see is what @Flrian reported in the above comment. It doesn't depend on activeness/inactiveness though:

The problem with shaded windows happens when `inset` is greater than half the height of the titlebar, so will depend on whether you have a different `inset` and `radius` for active/inactive windows.

I've thought about this a lot and I just can't think of a good way to fix it with how the shadows are currently drawn.  For large insets and small titlebars, the top and bottom corners overlap which looks bad.  Only options I can think of are:

1. Disable shadows on shaded windows (not keen on this)
2. Add a limit on inset to be no greater than half of titlebar height so the overlap doesn't happen (easy to implement and probably not a major problem but seems not ideal)
3. Add yet another set of shadow buffers just used for shaded windows (This will be a bunch more complexity but seems like the only proper way to handle it).

docs/labwc-theme.5.scd
johanmalm: I'm fine with the `.enabled` in themerc. I think it's more explicit and obvious to users compared with parsing `.color` and others for magic-numbers, `-1`, `INT_MIN`, etc.

I suspect @Consolatis's thought on `rc.xml` was to have a bigger override for modest hardward, but we could always add that later.

johanmalm:
> > The only issue I can see is what @Flrian reported in the above comment. It doesn't depend on activeness/inactiveness though:
> 
> The problem with shaded windows happens when `inset` is greater than half the height of the titlebar, so will depend on whether you have a different `inset` and `radius` for active/inactive windows.
> 
> I've thought about this a lot and I just can't think of a good way to fix it with how the shadows are currently drawn. For large insets and small titlebars, the top and bottom corners overlap which looks bad. Only options I can think of are:
> 
>     1. Disable shadows on shaded windows (not keen on this)
> 
>     2. Add a limit on inset to be no greater than half of titlebar height so the overlap doesn't happen (easy to implement and probably not a major problem but seems not ideal)
> 
>     3. Add yet another set of shadow buffers just used for shaded windows (This will be a bunch more complexity but seems like the only proper way to handle it).

Suggest 2 - and just add a check in `post_processing()`

tsujan:
> Add a limit on inset to be no greater than half of titlebar height

That could negatively affect how shadows look without shading. Relatively large insets can make shadows look more natural.

ahesford:
I haven't followed the details of this work, but managing a separate set of shadow buffers to handle shaded views seems highly undesirable. Shading is implemented such that a view simply has an effective height of zero (hence the `view_effective_height` function). Any shadow rendering should be implemented uniformly and correctly to views of *any* height, so there should be no need for special cases when wkndows are shaded.

I'll read more closely through the implementation and see if I can offer some better advice.

johanmalm:
> That could negatively affect how shadows look without shading. Relatively large insets can make shadows look more natural.

@tsujan You're right. I missed that (the risk of joining half way through the party!) Thanks for keeping an eye on the thinking here. Agree, limiting `inset` to `titlebar_height / 2` isn't great for the non-shaded case.

Is there a forth option where we crop the surfaces in the shaded-case to prevent overlap. @Consolatis mentioned `wlr_scene_buffer_set_source_box` which should be the one to use. Seems cleaner than having additional surfaces. I have never used that function myself though, so might be missing something.

https://github.com/swaywm/sway/blob/dcb142bf5e390250939544075b5852ca21eaf721/sway/sway_text_node.c#L67

cillian64:
@ahesford essentially my implementation assumes that we can separate the shadow into corners and edges which are the same for all windows and we just need to scale the edges for window size.  But this breaks down for windows which are too small in either axis.  A more general technique is to draw a black rectangle and apply a Gaussian blur, but this inherently needs re-drawing any time a window is resized, which I don't think will be acceptable in performance terms.

Between these is a trick where we assume unshaded windows are large enough for separable corners/edges and have a special-case for shaded windows.  This means an extra pair of buffers, but avoids having to do real-time redrawing while windows are resized.  But this won't work for arbitrarily sized small windows.

@johanmalm I could do a crop to stop the corners overlapping, but the shadows would end up a weird shape.  I've drawn a quick sketch to show this:

image

1. a normal window with lots of height - corners and edges are separable
2. A small window with inset (which is analogous to curve radius here) equal to half of height
3. A small window with inset greater than half height. This is the nasty overlap we're seeing now
4. A small window with inset greater than half height and corners cropped.  The shadows will have an odd-looking discontinuity at the half-way point.

cillian64:
> Correct, at least from my side that would be something that I would be *very* uncomfortable with, even for a opt-in feature.

Agreed, rendering efficiency is pretty much my top priority.

I did a few experiments.  Leaving out the lower corners or the entire lower edge looks pretty weird.  If I extended the right and left edges down to the bottom corners it would look better but still not great.

without_lower_corners
without_lower

Using `wlr_scene_buffer_set_source_box` works much better than I expected.  There is a slightly odd shape on the sides you can see when exaggerated using a really large radius, large inset, and high opacity:

source_box

But it's barely noticeable with more normal configuration, and would be even less obvious with something other than a plain white background, so maybe this is the best solution.

source_box_normal

I'll get this cleaned up and push some commits so others can play with it.

ahesford:
I think the `wlr_scene_buffer_set_source_box` solution looks like a winner.

docs/labwc-theme.5.scd
Consolatis: > I suspect @Consolatis's thought on `rc.xml` was to have a bigger override for modest hardward

Right, with the enabled option being in the theme it is the theme used that decides about a potential performance impact. So the user would need to use `themerc-override` to "fix" that which might not be obvious. (Thinking longer term here, no theme currently sets those properties because, well, they do only exist within this PR for now.)

> but we could always add that later

Agree.

tsujan:
It's perfect now:

shaded_good

Consolatis:
Is there any (software) design work left to do?

I think overall this PR is looking great (same for the shadows themselves).
There are some minor issues within the code / style but I held of on them for now to not disrupt the feature development.

cillian64:
> Is there any (software) design work left to do?

Broadly I think this covers the functionality and I can't find any more obvious bugs.  The remaining things on my TODO list are:

- [ ] Considering what the config/theme looks like
  - [ ] Should we keep `inset` as an option or hard-code it as 0.5?
  - [ ] Should we specify a shadow color (with opacity) or only specify the opacity?  Color makes it more similar to other options but I'm not sure anyone will ever use a color other than black. Plain opacity feels simpler to me.
- [ ] Add some more comments, in particular explaining the trick used to draw an `x^2`-dropoff gradient in Cairo

>  There are some minor issues within the code / style but I held of on them for now to not disrupt the feature development.

Thanks for this - I'm happy to start covering these issues now.

tsujan:
>  Should we keep inset as an option or hard-code it as 0.5?

Please don't hard-core it! How the shadow looks depends on it, and different users may have different tastes.

> I'm not sure anyone will ever use a color other than black.

Since a well known compositor like KWin has a color option, I think there are users who may like colors other than black — although I always use black.

docs/labwc-theme.5.scd
```diff
@@ -126,6 +126,28 @@ This syntax is not documented on the openbox.org wiki, but is supported by
 openbox and is used by many popular themes. For the sake of brevity, these
 elements are not listed here, but are supported.
 
+*window.active.shadow.inset*
+	How far behind the focused window should its drop-shadow be inset, as a
+	proportion of shadow radius. This determines the corner radius of the
+	shadow.
```
Consolatis, line 132: All descriptions are missing the default values which also clears up the unit expected here, e.g. something like `Default is 0.5.`

src/config/rcxml.c
```diff
@@ -855,6 +855,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.corner_radius = atoi(content);
 	} else if (!strcasecmp(nodename, "keepBorder.theme")) {
 		set_bool(content, &rc.ssd_keep_border);
+	} else if (!strcasecmp(nodename, "dropShadows.theme")) {
+		set_bool(content, &rc.shadows_enabled);
```
Consolatis, line 859: Misses setting the default value in `rcxml_init()`, important when removing / commenting out / having a spelling mistake on an option and then doing a `Reconfigure`.

src/ssd/ssd_shadow.c
```diff
@@ -0,0 +1,269 @@
...
+			case LAB_SSD_PART_CORNER_BOTTOM_RIGHT:
+				x = width - inset;
+				y = -titlebar_height + height - inset
+					+ overlap_px - avoid_gap;
+				wlr_scene_node_set_position(part->node, x, y);
+
+				src_box.x = 0;
+				src_box.y = overlap_px;
+				src_box.width = corner_size;
+				src_box.height = corner_size - overlap_px;
```
Consolatis, line 88: Can we somehow reduce code duplication here? There seems to be only one box with its axis changed on two occasions. Maybe we could define it once outside of the loop and have a static inline function to create a copy and switch `.x` and `.y` and `.width` and `.height` before returning it?

Or as alternative, what about something like this:
```c
void
buffer_transform(struct wlr_scene_buffer *buffer, int overlap_px,
		int corner_size, bool switch_axis)
{
	struct wlr_fbox box = {
		.x = switch_axis ? overlap_px : 0,
		.y = switch_axis ? 0 : overlap_px,
		.width = switch_axis ? (corner_size - overlap_pix) : corner_size,
		.height = switch_axis ? corner_size : (corner_size - overlap_pix)
	};
	wlr_scene_buffer_set_source_box(buffer, &box);
	wlr_scene_buffer_set_dest_size(buffer, box.width, box.height);
}
```

src/theme.c
```diff
@@ -627,6 +627,26 @@ entry(struct theme *theme, const char *key, const char *value)
 			theme->window_inactive_button_close_unpressed_image_color);
 	}
 
+	/* window drop-shadows */
+	if (match_glob(key, "window.active.shadow.radius")) {
+		theme->window_active_shadow_radius = atoi(value);
+	}
```
Consolatis, line 633: Misses setting the default values in `theme_builtin()`, same reason as for `rcxml.c`.

Consolatis (commented):
Some minor things below:

src/theme.c
```diff
@@ -627,6 +627,26 @@ entry(struct theme *theme, const char *key, const char *value)
...
+	if (match_glob(key, "window.inactive.shadow.color")) {
+		parse_hexstr(value, theme->window_inactive_shadow_color);
+	}
+	if (match_glob(key, "window.active.shadow.inset")) {
+		theme->window_active_shadow_inset = atof(value);
+	}
+	if (match_glob(key, "window.inactive.shadow.inset")) {
+		theme->window_inactive_shadow_inset = atof(value);
+	}
```
Consolatis, line 648: `atof()` can now be replaced with `set_float()` or `set_double()` via `common/parse-double.h`.

cillian64:
Thanks for the comments!  I've resolved these and had another look through all of it.  I'm trying to work out if there's a good way to simplify or split up `set_shadow_geometry` even more as there's still a lot going on in there.  Also not overjoyed by the `avoid_gap` bodge so trying to work out if there's a better way to do that.

johanmalm:
It feels like this one is getting close to merging. It's had plenty of interest, reviews and testing which is great. Based on your last comment, it feels like you're still tying up a few loose ends, so I'll hold off. Please could you post a "Ready for final review" or similar so I know when to go.

I've just had an 'interim' look and note the following:

- [ ] 1. Did you want to do some interactive-rebasing, or shall I just do a squash-merge when it comes to it?

- [ ] 2. I know it's petty, but would you be okay to keep a line clear at the start of multi-line comments (example below)

```
/*
 * This is a multi-line comment which is much much much much much much
 * longer.
 */
```

https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#linux-kernel-style-basics

- [x] 3. Let's stick with the `rc.xml` / `themerc` split the way you've done it. Seems fine.

- [ ] 4. Are we sure about the names `inset` and `radius`? I'm fine with how we use them to generate shadows, but think that the terms might be slightly misleading. I'm happy to draw to diagrams later, but it'd be good to agree a quick sketch and what we actually mean be things. I've had a quick go (below). Am I right in thinking that A=radius and inset=(B-A)/A (the way it's defined by the user in themerc).

Playing around, it seems like the actual corner radius ends up being B (which I guess is what you're saying in the man page: "This determines the corner radius of the shadow). If I'm right in what I've said above, I think this is a tad confusing that the radius is not actually the radius the user defined.

It might be simpler to just call A `thickness` - and define it as how much the shadow visibily extends outside the window (or something like that). Could consider `size` too, but I think `width` and `height` are best avoided as it becomes orientation dependent.

The word `inset` doesn't quite seem to fit the bill either (unless I'm not familiar with the meaning / usage). It sort of feels like it ought to fit - like "inset in plot/diagram" or opposite to offset. If it's inspired by other shadow implementations, CSS or similar, than fine (for consistency). Otherwise, I'd suggest considering `overlap` or `undercut` or similar.

Another approach here is to actually let B=`window.active.shadow.corner-radius`; and A=`window.active.shadow.thickness` - and then you can calculate `inset` (or whatever we call it) based on those. To me this seems more intuitive to the end-user. If we want to use this approach, I suggest `corner-radius` because picom uses `radius` to mean `gaussian blur radius` - and you never know what might get considered in future.

terminology

Example ascii diagrams (I can create later):

- https://github.com/labwc/labwc/blob/67669dcf1e7dcc6536d02cd007c8f10aa93a7279/src/theme.c#L779-L793
- https://github.com/labwc/labwc/blob/67669dcf1e7dcc6536d02cd007c8f10aa93a7279/src/theme.c#L864-L882

- [ ] 5. Regarding `inset = inset / (1.0 + inset);` in `shadow_corner_gradient()`, could we rename it to inset_as_ratio, inset_in_pixels, whatever?

- [ ] 6. With `WLR_RENDERER=pixman ./build/labwc` the corners don't look right. Not sure what's doing on there. Maybe the x^2 trick? See scrot below (using default settings). Thanks to @Consolatis for thinking of that test-case.

scrot

Consolatis:
> * [ ]  6. With `WLR_RENDERER=pixman ./build/labwc` the corners don't look right. Not sure what's doing on there. Maybe the x^2 trick? See scrot below (using default settings). Thanks to @Consolatis for thinking of that test-case.

Can confirm: all default settings + `window.*.shadow.color: #000000 100` in `themerc-override`:
gles2
20240404_10h19m18s_grim
pixman:
20240404_10h19m42s_grim

Which is pretty strange, the cairo buffers should be exactly the same for both so this might actually indicate a wlroots bug within the pixman renderer.

cillian64:
> It feels like this one is getting close to merging. It's had plenty of interest, reviews and testing which is great. Based on your last comment, it feels like you're still tying up a few loose ends, so I'll hold off. Please could you post a "Ready for final review" or similar so I know when to go.

Yep, will do!  Remaining things on my TODO list are the config options and pixman rendering issues.

> 1. Did you want to do some interactive-rebasing, or shall I just do a squash-merge when it comes to it?

I'll do a tidy-up rebase at the end.

> 2. I know it's petty, but would you be okay to keep a line clear at the start of multi-line comments

Sure, done!

> 4. Are we sure about the names `inset` and `radius`?

I think you're right that we can do better than these and it's still not very clear.  You're correct that currently `radius=A` and `inset=(B-A)/A`, or `B=A(inset + 1)`.  I do think "shadow corner radius" would be clearer than the current proportional inset.  However, I've found things usually work best if `B` changes proportionally to `A`, and I've found `A` amenable to tweaking while `inset` (as a proportion of `A`) seems best left alone.  I wonder if there's a good way to specify "shadow corner radius as a multiple of shadow size" (which would default to `1.5` currently) or whether it's better to just have "shadow size" and "shadow corner radius" and hope users will figure out to change them in proportion.

> It might be simpler to just call A thickness - and define it as how much the shadow visibily extends outside the window (or something like that). Could consider size too, but I think width and height are best avoided as it becomes orientation dependent.

Agreed that "radius" and "width" are already too overloaded.  I like "shadow size" for this, "thickness" feels a little odd for something potentially large and soft-edged.

> 6. With `WLR_RENDERER=pixman ./build/labwc` the corners don't look right.

This is actually a pixman issue (I've made a reproducer using pixman directly).  If you make a 1xN image then scale it to be wider (or vice-versa transposed) with bilinear filtering then you get an odd triangle shape.  I can't tell if it's actually a pixman bug or just unexpected behaviour.  Regardless, specifying nearest-neighbour scaling makes the problem go away (and might be faster too).

While playing around I spotted an entirely separate issue with pixman and shaded windows:

pixman_issue_2

I haven't dug into it yet but my guess is that this one is a wlroots bug where source-crop and rotation are applied in a different order on GLES vs pixman.

heroin-moose:
I suggest (optional?) support for hiding shadows during effective tiling (following GTK example that unrounds corners and hides drop shadows).

Consolatis:
> I suggest (optional?) support for hiding shadows during effective tiling (following GTK example that unrounds corners and hides drop shadows).

What what its worth, we only unround corners on maximize.

heroin-moose:
Also it would be super cool to have drop shadows for menus.

tsujan:
> Also it would be super cool to have drop shadows for menus.

→ https://github.com/labwc/labwc/pull/1648#issuecomment-2015848098

heroin-moose:
> > Also it would be super cool to have drop shadows for menus.
> 
> → [#1648 (comment)](https://github.com/labwc/labwc/pull/1648#issuecomment-2015848098)

:(((((((((((((((

heroin-moose:
> > I suggest (optional?) support for hiding shadows during effective tiling (following GTK example that unrounds corners and hides drop shadows).
> 
> What what its worth, we only unround corners on maximize.

Nope, at least the git versions unrounds corners upon moving to regions. GTK3 also has a bug that rounds them back, but loses shadows :D

Consolatis:
> > > I suggest (optional?) support for hiding shadows during effective tiling (following GTK example that unrounds corners and hides drop shadows).
> > 
> > What what its worth, we only unround corners on maximize.
> 
> Nope, at least the git version unrounds corners upon moving to regions. GTK3 also has a bug that rounds them back after moving out of the region, but loses shadows :D

Should have been more specific, labwc only unrounds SSD corners on maximize.
The shadow implementation is only for server side decoration clients as CSD clients may render their own ones.

Consolatis:
About the pixman to gles2 rendering differences regarding the order in which transforms are applied, should we simply disable the shadows when running on the pixman renderer for now (until we start to track wlroots 0.18)? Even if it is fixed at wlroots and backported to 0.17.x we would still render it wrong for pixman users of earlier wlroots 0.17.x versions.

It can be checked for by `if (wlr_renderer_is_pixman(view->server->renderer)) { .. }`.

Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3830

johanmalm:
Agree

cillian64:
> About the pixman to gles2 rendering differences regarding the order in which transforms are applied, should we simply disable the shadows when running on the pixman renderer for now (until we start to track wlroots 0.18)?

I think something like this is a good idea.  However, selfishly, I'd prefer if we could do it slightly differently.  The issue only appears with shaded windows and large-ish shadows.  On Raspberry Pi OS we do use the pixman backend and our default config has smallish shadows *and* no option for window shading.  So, could we disable shadows if the renderer is pixman and the shadow size is larger than the window height?  This means pixman users can still have shadows most of the time and only lose them in a pretty narrow situation.

cillian64:
> Actually another option would be to simply clamp the inset in case of shaded + pixman renderer.

That could work, but I'm not super keen on having the shadow appear subtly wrong or not as configured.  Hopefully having the shadow disappear entirely makes it look a bit more deliberate.

ahesford:
Having shadows mysteriously disappear because of an implementation detail in shadow rendering will be more surprising to the user than having the inset clamped to whatever upper limit produces a reasonable approximation.

All of the visual artifacts with shadows are caused by hacks that approximate the actual shadow calculation to improve performance; insisting on "exact" rendering of an approximation to the real thing and bailing out entirely when you can't meet the constraints seems worse than just tweaking the approximation a bit more in these limiting cases.

src/ssd/ssd_shadow.c
```diff
@@ -0,0 +1,279 @@
...
+		int inset = inset_proportion * (double)visible_shadow_width;
+
+		/*
+		 * Total size of corner buffers including inset and visible
+		 * portion.  Top and bottom are the same size (only the cutout
+		 * is different).  The buffers are square so width == height.
+		 */
+		int corner_size = active
+			? theme->shadow_corner_top_active->unscaled_height
+			: theme->shadow_corner_top_inactive->unscaled_height;
```
ahesford, line 170: This crashes labwc if buffer allocation for any shadow part fails. For example,
```
window.active.shadow.inset: 0
window.active.shadow.radius: 0
```
will cause labwc to segfault the first time a window is drawn. The shadow logic needs to handle these cases more gracefully, see comments in `create_shadows` below.

src/theme.c
```diff
@@ -950,6 +978,186 @@ create_corners(struct theme *theme)
...
+{
+	int inset_active = (double)theme->window_active_shadow_radius
+		* theme->window_active_shadow_inset;
+	int inset_inactive = (double)theme->window_inactive_shadow_radius
+		* theme->window_inactive_shadow_inset;
+	/* Total width including visible and obscured portion */
+	int total_active_width =
+		theme->window_active_shadow_radius + inset_active;
+	int total_inactive_width =
+		theme->window_inactive_shadow_radius + inset_inactive;
```
ahesford, line 1107: It seems perfectly reasonable to allow a user to set a zero radius and inset for one or both of these parameters and react accordingly. For example, one might enable shadows globally and completely disable them in the theme, or disable only inactive window shadows by setting the radius and inset to 0. When one of `total_active_width` or `total_inactive_width` is zero, I suggest you avoid trying to create all buffers associated with the corresponding shadow, and then adapt other logic to avoid trying to draw shadows for that case.

The parsing of the themerc should properly validate all numeric arguments. For example, a single negative value will crash labwc, while two negative values will efefctively disable shadows but allow the program to continue. When parsing these parameters, they should be clipped to zero and a warning issued if negative values are specified.

Consolatis: We do have `post_processing()` for that and I agree that the values should be validated / clamped.

It is also a broader issue in labwc in general, we likely have quite some more ints that aren't being checked but that is unrelated this PR.

cillian64:
> Having shadows mysteriously disappear because of an implementation detail in shadow rendering will be more surprising to the user than having the inset clamped to whatever upper limit produces a reasonable approximation.
> 
> All of the visual artifacts with shadows are caused by hacks that approximate the actual shadow calculation to improve performance; insisting on "exact" rendering of an approximation to the real thing and bailing out entirely when you can't meet the constraints seems worse than just tweaking the approximation a bit more in these limiting cases.

You've convinced me and I just tried to do it this way by clamping inset.  But I hit an unexpected problem: The cutouts from the shadow buffers (to make them not appear through translucent windows) are positioned based on the inset.  If we reduce inset then the cutouts become visible as they no longer line up with window edges.  The only solutions to this I can think of would be either to redraw the shadow corner buffers under `ssd_shadow_update` each time geometry changes, or to have another set of corner buffers without cutouts which we use when this happens (either forcing inset=0 or just accepting it'll look odd with translucent windows).  To me these both feel excessively complicated for a pretty niche workaround.  Is either acceptable to you or can you think of another way around it?

ahesford:
> You've convinced me and I just tried to do it this way by clamping inset. But I hit an unexpected problem: The cutouts from the shadow buffers (to make them not appear through translucent windows) are positioned based on the inset. If we reduce inset then the cutouts become visible as they no longer line up with window edges. The only solutions to this I can think of would be either to redraw the shadow corner buffers under `ssd_shadow_update` each time geometry changes, or to have another set of corner buffers without cutouts which we use when this happens (either forcing inset=0 or just accepting it'll look odd with translucent windows). To me these both feel excessively complicated for a pretty niche workaround. Is either acceptable to you or can you think of another way around it?

Pre-rendered shadow buffers are essentially a performance-enhancing cache. I think you should remember the parameters used to render the shadows alongside the shadow buffers (in this case, at least the radius and inset) and use them to validate the cached buffers. When you update the geometry, you check to make sure that the current effective parameters (which should usually match the configured values, except when the window gets too small and they are clipped) match whatever values were used to draw the cached buffer in the first place. If they don't match, you throw away the buffers and redraw them with the new parameters, then remember those. This minimizes redrawing but also makes sure that we always render buffers with parameters that are expected. It might even simplify logic when the user changes the theme values and reconfigures, because it unifies all of these cases.

cillian64:
Actually it's even worse than that because the inset clamping is done *per window*.  So we'd need to potentially render and store a different set of shadow buffers for each window (even if we never actually hit this case, the code would still need the added complexity of handling it).  I think it would be too weird to have resizing/shading one window change the shadow appearance on all other windows.

cillian64:
At this point I'm leaning towards going back to the original suggestion of disabling shadows with the pixman renderer for now, while adding a way to force-enable pixman shadows.  That's a simple workaround until this is hopefully fixed in wlroots 0.18.

For rpi-os we can just add a downstream patch to fix the issue in wlroots 0.17, make our labwc package depend on new enough wlroots, and force-enable pixman shadows in labwc.  If any other pixman users are desperate for shadows right now then they can also force-enable (and install an updated wlroots if they would be affected by the issue).

ahesford:
I've had a chance to play with this a bit more, and have some more opinions:
1. The `inset` parameter should never be exposed to the user. It is a detail of your approximation that, if you use, should be selected correctly by the code and not configurable. The only tunable parameters should be color and radius.
2. The `inset` parameter as implemented doesn't really behave properly. Attached images show the result of setting a dark shadow color `#000000 100` and a radius of `64` with inset parameters of `0.0`, `0.5` and `1.0`, respectively. Setting to `0.0` produces some strange discontinuities in the corners, while `0.5` and `1.0` show strange behavior where the darker regions of the shadow "clip" at a 45-deg angle behind the corners. None of these are physical.

Properly stitching corners might get complicated, so I can appreciate the value of picking some reasonable default that probably isn't too obvious for "typical" shadow coloration, but the need to compromise based on the approximation is all the more reason not to expose this parameter to the user.
 
inset-0 0
inset-0 5
inset-1 0

cillian64:
>  The `inset` parameter should never be exposed to the user. It is a detail of your approximation that, if you use, should be selected correctly by the code and not configurable. The only tunable parameters should be color and radius.

I would also prefer to remove this option and fix it somewhere around `0.5` (with the current `inset` parameter) looked best.  Not because I'm trying to do an accurate approximation of something physical, but because I don't think you get good-looking (by anybody's opinion) results by varying it.  At least @tsujan still wanted the option to tune it (https://github.com/labwc/labwc/pull/1648#issuecomment-2034152913), but I'll let them make the argument for that.

>  The `inset` parameter as implemented doesn't really behave properly. Attached images show the result of setting a dark shadow color `#000000 100` and a radius of `64` with inset parameters of `0.0`, `0.5` and `1.0`, respectively. Setting to `0.0` produces some strange discontinuities in the corners, while `0.5` and `1.0` show strange behavior where the darker regions of the shadow "clip" at a 45-deg angle behind the corners. None of these are physical.

I generated the following image in GIMP by making a white background, a layer with a black square, and a layer with a red square the same size and location.  I applied a Gaussian blur to the layer with the black square.  I then used the levels-curve tool to increase the shadow opacity a bit to make it clearer what's happening.  I think this is basically what we're trying to approximate, and in my opinion this looks pretty close to the `0.5` option (or would if you had set a lower opacity, it's always going to look pretty weird with 100% opacity.).

square_shadow

It's correct that if you follow a contour of constant colour around the shadow you get something a bit like a rounded rectangle which will "clip" behind the corners of the original rectangle (this is true for both a Gaussian blurred rectangle and for my approximation).  If you could see behind the window corner you would see a curve, not a 45-degree chamfer (you can check this by commenting out the cutouts and rendering a transparent window).  Regardless, I think subjective appearance matters more than whether it's an accurate simulation of something physical.  I also don't think it's reasonable to try and make it so all possible configurations will look good - setting basically any theme option to extreme values will look bad.

droc12345:
Rounded shadows on a square corner window looks a little ... weird.

cillian64:
> Rounded shadows on a square corner window looks a little ... weird.

Are you saying the shadow shouldn't be rounded or just that the amount of/style of rounding looks wrong?

droc12345:
Well, in real life, something that would produced shadows would show squarish shadow corners.
But you're not alone in doing that, I've seen other examples.

But for the red with shadows, the rounding is a little too much, IMO.
It's my 1st impression when I looked at it.

ahesford:
I don't want this discussion to be sidelined by a bunch of "this would look better" back-and-forth. We need a reasonable, low-latency and maintainable implementation that looks halfway decent. People can propose additional improvements in subsequent PRs if they can do a better job, although such requests should be about improving the implementation, not endlessly tweaking default parameters. Let users pick their own opacity and radius rather than worrying about what's on by default.

I raised my points above to highlight that we shouldn't expose the inset to users because it doesn't really offer any intuitive meaning and is intimately tied to implementation details of an approximation to an actual shadow render. The fact that we've seen weird special cases and bugs appear also alarms me, because it highlights limits of the approximation. While this needs to be fast, we also don't want the implementation devolving into a ton of "if-this-then-that" hacks. That isn't maintainable and also doesn't make a very good approximation.

The discontinuities at zero inset suggest to me a math or logic error somewhere, and a nonzero inset might just paper over the error by hiding the unexpected discontinuities behind the window. It also complicates the implementation. I need to look more closely to determine whether I'm right, but I suspect that shadow stitching can be done "correctly" without any inset.

ahesford:
OK, I've had a chance to do some testing, and can confirm that the weird artifacts when stitching corners with zero inset are a consequence of the discontinuous derivative of the quadratic function used to approximate a Gaussian fade. The first attachment (`quadratic-profiles.pdf`) shows a shadow constructed by applying a quadratic rolloff in both directions, with obvious banding as the corners set in. Using a Gaussian rolloff ensures a continuous derivative that eliminates this band and remains separable. The second attachment (`gaussian-profiles.pdf`) shows the resulting shadow and its profiles, with the quadratic profile in blue for comparison.

For these tests, I used a 256-by-256 grid and applied the Gaussian and quadratic profiles over 64 pixels along each edge. The quadratic profile is simply the square of a linear variation from 0 to 1 over the 64-pixel interval, while the Gaussian is half of a 128-pixel-wide Gaussian window with a standard deviation of 21 pixels.

The Gaussian rolloff can be handled in the same fashion as the current implementation with no `inset` parameter:
1. For a given "radius", which is probably better called "width", pick a standard deviation that is 1/3 of the parameter, and compute the half Gaussian for that standard deviation and a width equal to the parameter value (thus, truncate the Gaussian from [0, 3] standard deviations).
2. Use this simple profile on each of the edge buffers; the buffer can be stretched simply along the direction perpendicular to the rolloff, using any arbitrary and simple interpolation scheme (nearest neighbor would be fine, or just picking a pixel and replicating that value along the entire edge).
3. A single horizontal edge can be computed once using the simple scheme and rotated 180 degrees for the opposite edge; likewise for the vertical edges.
4. The corners are just a product of two perpendicular half-Gaussians, and can be rotated to each of the four orientations.

In other words, you compute one half-Gaussian profile for the desired radius, compute an outer product of the profile with itself to produce a single corner that you clone and rotate to each of the four window corners, and then also stretch the profile along each edge up to the corners. No inset, no mess, no artifacts, should be just as fast as what you're doing now.

[quadratic-profiles.pdf](https://github.com/labwc/labwc/files/14934561/quadratic-profiles.pdf)
[gaussian-profiles.pdf](https://github.com/labwc/labwc/files/14934562/gaussian-profiles.pdf)

ahesford:
I've marked this draft for now, because I want to find time (maybe this weekend) to try out the Gaussian idea atop this branch. The result should keep the same general concept and most of the implementation as is.

cillian64:
@ahesford I appreciate you digging into this in such detail.  I agree we could use Gaussian drop-offs instead of quadratic and this improves the discontinuity where the corners meet the edges.  I'm pretty surprised that multiplication of the horizontal and vertical Gaussian kernels gives the exact same result as separately convolving by each, but I've convinced myself this works so I could relatively easily change my implementation to use Gaussian drop-offs instead of quadratic.

However, even using the Gaussian drop-off, I feel that we still need the inset.  Looking at aesthetics, this is what a Gaussian dropoff shadow with full opacity and no inset looks like (if you think the shadow size is too large then just zoom out and imagine it's a smaller object):

No inset, full shadow opacity

And this is what it looks like when you reduce the opacity a bit:

No inset, partial shadow opacity

I think both of these look really bad.  This is what it looks like with some inset added:

Inset, partial shadow opacity

To me, that looks much better.

In physical terms, whether you want inset depends whether you imagine the window as a cuboid sitting on a surface, or as a 2d lamina floating above the surface.  If the former, then I think having no inset is correct.  Personally I imagine it as the latter, which does require an inset:

Imagine a distant point-source of light shining vertically down onto an object floating a small distance above a white surface.  The distant point-source produces effectively parallel rays so on the surface we see a black shadow with a perfectly hard edge the same size/shape as the object.  As you diffuse the light source, the shadow edge softens (becomes blurred).  Some rays of light are at an angle and are now blocked by the object so the outermost edge of the blurred shadow grows.   However, some sideways rays of light now get under the object causing the inner edge of the blurred shadow to shrink.  The width (standard deviation) of the half-Gaussian increases but its mid-point (where its value is 50% of peak) is what stays the same size/shape as the object, not the top of the curve.  So if this is what you want to emulate then the inset should be fixed at half the total shadow width (where "total shadow width" includes both the part under the object and outside the object) or equal to the exposed shadow width.  I think you could justify a smaller inset by having the light source be closer by.

And as a final appeal to authority, this (inset equals half total shadow width) is exactly what CSS shadows do if you don't specify an additional spread and so is presumably also what GNOME/GTK does.

```html
<style>
div {
  margin: 2rem 2rem;
  height: 100px;
  width: 100px;
  background-color: #ff0000;
  filter: drop-shadow(black 0 0 10px);
}
</style>

<div></div>
```

image

In summary, I am fine with choosing a fixed inset and removing the config option, but I'm pretty sure that having no inset at all is going to look unacceptably bad.  I can change the drop-off function from quadratic to Gaussian if you want, but I don't think this is going to remove the need for inset and so we'd still need to work out how to deal with the wlroots bug.

ahesford:
You've convinced me that the inset has value, but I advocate for keeping that a fixed parameter that is not exposed to the user for now. I think using a Gaussian profile is important to preserve continuity of the derivative when transitioning from edge to corner, avoiding any unnatural transition bands. And yes, the Gaussian is separable by construction, so you can apply the rolloff independently along the two directions. (As an aside, the term *blur* is misused here, because we are not doing any convolutionggs; this is just a multiplicative roll-off.)

Furthermore, the inset should apply, effectively, to the edges as well as the corner. (You can just start the Gaussian for the edge at the point corresponding to the inset---there is no need to start at the peak and then try to clip the invisible portion.) Unless I've missed something, you aren't currently applying the inset to the edges, which is why you get strange darker bands near the edges that fade out as you approach the corner, contributing to the "chamfered" look of the shadow.

I don't have a good answer or a strong preference regarding pixman renderer issues with wlroots. If wlroots is going to fix this issue in an 0.17 backport, I'd rather just let users of older wlroots live with the bugs or disable shadows on their own if they are bothered by the bug; we shouldn't complicate our lives to work around a fixed bug when we can just tell people to upgrade. If it's not going to be backported, I guess it's fine to disable drop shadows on too-small windows for now and just keep things simple.

ahesford:
Did you compare performance of this implementation to just rendering a partially transparent rectangle with Gaussian rolloff on the edges? Remember, a full convolution is unnecessary; just make the rectangle one full Gaussian wider and taller then the actual window, then multiply the alpha along each edge of the rectangle by half the Gaussian. This will get your desired inset behavior and will work without issue as long as the window is twice the size of the inset. We can truncate the multiplications for smaller windows to get a close and simple approximation (or bite the bullet and do a full convolution along each narrow axis as necessary).

This will render the shadow behind transparent windows, but that isn't necessarily wrong because partially transparent films still cast shadows in the real world too. Even if cutting the shadow out from the window is preferable, rendering one simple shape and cutting out the master window would be a lot easier and may not be meaningfully slower that trying to stitch together several pieces.

src/theme.c
```diff
@@ -950,6 +978,186 @@ create_corners(struct theme *theme)
...
+	cairo_save(cr);
+	cairo_set_operator(cr, CAIRO_OPERATOR_CLEAR);
+	/* top left */
+	cairo_move_to(cr, 0.0, 0.0);
+	/* top right */
+	cairo_line_to(cr, inset * radius, 0.0);
+	/* bottom right, with radius */
+	cairo_arc(cr, inset * radius - cutout_corner_radius,
+		inset * radius - cutout_corner_radius,
+		cutout_corner_radius, 0.0, pi / 2.0);
```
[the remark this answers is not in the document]
cillian64: Code removed, no longer need pi here

src/ssd/ssd_shadow.c
```diff
@@ -0,0 +1,279 @@
...
+	struct wlr_fbox src_box = {
+		.x = switch_axes ? overlap_px : 0,
+		.y = switch_axes ? 0 : overlap_px,
+		.width = switch_axes ? height : width,
+		.height = switch_axes ? width : height,
+	};
+	wlr_scene_buffer_set_source_box(buffer, &src_box);
+	/* But scaling is applied after rotation so no axis flip */
+	wlr_scene_buffer_set_dest_size(buffer, width, height);
+}
```
ahesford, line 49: It seems that the overlap concerns also apply to width, if somebody makes the radius large and a window is narrow. Shouldn't the crop consider widthwise overlap as well?

cillian64: Theoretically this could happen but I haven't seen any cases in testing nor has anyone testing reported any, so I didn't think it was worth the extra complication to handle it.  Seems like the vertical case is only actually seen in practice because of window shading.

cillian64: The halves aren't equal, one is an issue which was quickly noticed by actual users, the other is still theoretical and to me seems very unlikely.  The time I can spend on this is not unlimited and there's still plenty more to do so I'd appreciate if we could prioritise, and we can make further improvements once we find out the wild ways users break this in practice.

src/ssd/ssd_shadow.c
cillian64: Fixed

src/ssd/ssd_shadow.c
ahesford: The logic is the same, you should just add another overlap parameter that applies to the width and replace the zeros in this function with the other parameter. I understand that your time is not infinite, but I'm also opposed to releasing a feature that is known to be defective for limiting cases *in the same manner as you're already fixing*.

cillian64: I'll have a look at fixing this.  Please can you let me know when you've finished reviewing so I can address any remaining issues in one go in order to avoid any more re-rewriting and multiple rounds of testing/checking.

johanmalm:
Looks like this is shaping up nicely :smile:

I've just had a quick glance through code/comments with two quick points:

1. Color format has recently changed in master. Would you mind changing `#000000 80` to the `#rrggbbaa` format? Sorry about the API change there - the former is deprecated.

Ref: [2c23415](https://github.com/labwc/labwc/commit/2c2341530bce800c90e6cb015dd274dedfea6d6c)

2. Regarding the `pixman` issue, I suspect the `wlroots` team would accept a backport if we ask nicely - particularly if we prepare the MR so that they just review/merge.

Sorry for being a bit behind here. Is it already fixed in `0.18` or is this the issue: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3830

> I don't know yet whether wlroots will backport the change. I don't mind either way whether we disable shadows on pixman or leave them as-is. Consolatis and johanmalm initially said to disable shadows with pixman for now, so I'll go with that.

Yes, although that was before the penny dropped and I realized RPi uses `pixman` (and before I had given it much thought - sorry). I'm easy on this, but it would feel right that you @cillian64 (and possibly @spl237) take a view on what you think will work best for you. I mean, we could add a 'disable-if-pixman' and then remove it later when the fix is in the latest `wlroots` release, but that seems messy in terms of knowingly setting ourselves up for changing behavior later. I'm guessing the easiest route would be to NOT hard-code a `disable-if-pixman` but let distros change the `dropShadow` setting `/etc/xdg/labwc/rc.xml` when the `wlroots` package is patched.

ahesford:
If you address the concerns I've already raised, I'm not going to stand in the way of this PR any longer.

I've raised all of these issues because they are direct consequences of trying to approximate the intended shadow form rather than just drawing it. The stitching is clever, but it's also complicated and, as we've seen, gets more complicated as we hit the limits of the approximation. Even if we don't want to draw a full rectangle behind the window, we can directly calculate Gaussian edge buffers in about 700k floating-point multiplications. Even on a low-spec RPi (*i.e.*, older than a 3A/B), we get better than 1 GFLOPS according to my quick search, and we can live-update a 4k-by-2k shadow during resize at better than 1000 FPS. I'm not convinced that all of the stitching and buffer manipulations are worth any savings---or that there is any real savings---over just implementing a straightforward (maintainable and readily understandable) direct render. Yes, we'd have to populate buffers rather than telling wlroots or cairo to rotate and scale a few that pre-computed images; but under the hood, wlroots and cairo are doing this work anyway. Stretching out a 1-pixel buffer via nearest-neighbor interpolation isn't going to be much better than manually copying a column of data over a bunch of rows and, as we approach the corners, rolling off the edges. If the wlroots/cairo nearest-neighbor interpolator doesn't special-case a 1-D image, then the manual manipulation might actually be better. The same goes for corner rotation, which might be smart about just flipping pixels, or might be naïve and actually involve a bunch of matrix multiplies.

I'm not trying to be a jerk about this or run you through the wringer for fun. However, after this is merged, the only people we can be sure will be around to fix visual glitches and refine the implementation are project members. While we all hope that contributors will remain around indefinitely as stewards of their work and appreciate their ongoing support, we also need to be prepared to assume responsibility when they decide to move on to other things. To that end, I want to make ensure that the implementation isn't so obscure that we'll have to re-teach ourselves its implementation in three months, and also that we aren't starting with known glitches that will lead to new issues in short order.

cillian64:
> Color format has recently changed in master. Would you mind changing `#000000 80` to the `#rrggbbaa` format?

Sure, I'll update this and rebase.  I did find the `#rrggbb aaa` format a bit surprising.

> Is it already fixed in `0.18` or is this the issue: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3830

That is the issue, so no fix yet but it's high on my TODO list.  I'll be backporting to wlroots 0.17 anyway so I'll submit the backport upstream too.

> I'm guessing the easiest route would be to NOT hard-code a `disable-if-pixman` but let distros change the `dropShadow` setting `/etc/xdg/labwc/rc.xml` when the `wlroots` package is patched.

That works for me.  On rpi-os we won't enable shadows until we have the wlroots fix rolled out and we'll make our labwc package depend on the fixed wlroots package.  I imagine other distros won't be enabling shadows by default for a while so any users who turn it on and have problems will know how to turn it off again.

@ahesford:

In terms of code complexity, in some ways having single shadow buffers per-window would be simpler than scaling up several small buffers, but in other ways it would add some complications (adding shadow buffers per window, reallocating the buffers when needed for resizes).  All the annoying transform and rotation logic currently manipulating buffers wouldn't go away, it would just move to the code which draws the shadow buffers.  And we'd still have the same issues around small windows (assuming we don't start convolving the whole thing. Special-casing the convolution for small windows would add a separate set of complexity).

In terms of performance, I agree that just drawing the full shadow buffers wouldn't be totally unachievable, but it is a reasonable amount of compute, memory, and memory bandwidth which we don't need to spend.  I'm also thinking in terms of performance when compositing the desktop, in particular memory bandwidth.  On Raspberry Pis we've seen a lot of memory bandwidth get eaten up during composition, and in this regard scaling up small buffers is substantially better than blitting a single large translucent buffer as you only need to read the row/column of pixels once.  This is especially true if you had a large translucent window where you would have to read every pixel in the shadow buffer for blending (Right now the same also applies to opaque windows when using pixman because it isn't very smart about occlusion, but that's on my TODO list...).  When doing desktop composition using dedicated composition hardware (i.e. libliftoff/DRM) instead of GPU/CPU scaling up small buffers is trivial compared to blitting large mostly-transparent buffers.

If I was proposing really substantially increased complexity then I'd properly go through and benchmark to prove it's actually necessary.  But to me it looks like the complexity is at least roughly comparable either way, so being reasonably sure performance will be better (and certainly not worse) seems like sufficient reason to go with the separate shadow buffers.

>  However, after this is merged, the only people we can be sure will be around to fix visual glitches and refine the implementation are project members. While we all hope that contributors will remain around indefinitely as stewards of their work and appreciate their ongoing support, we also need to be prepared to assume responsibility when they decide to move on to other things. To that end, I want to make ensure that the implementation isn't so obscure that we'll have to re-teach ourselves its implementation in three months, and also that we aren't starting with known glitches that will lead to new issues in short order.

For what it's worth,  @spl237 and I are definitely planning and expecting to be around for support, bugfixes, and improvements in labwc for a long time.  But I totally take your point here and in general try to minimise complexity unless there is a good reason not to.  Would it help if I added more localised comments or a single large comment explaining the strategy?

> If you address the concerns I've already raised, I'm not going to stand in the way of this PR any longer.

Just to check, are there any outstanding other than shadow inset with not-wide-enough windows?

ahesford:
The only issues I'm aware of are handling too-narrow and too-short windows uniformly, and gracefully handling cases where shadows are enabled in rc.xml but at least one radius is not positive.

cillian64:
> gracefully handling cases where shadows are enabled in rc.xml but at least one radius is not positive.

I believe I fixed this a couple of days ago, but let me know if not.

My latest push squashes everything so far, rebases on latest master, updates the shadow colour format, and fixes the shadows for narrow windows (I concede that it was much simpler than I expected).  I've taken out the fix for the single pixel gap because I noticed an issue with how I was solving this for the horizontal case, I'm going to re-do this tomorrow for both narrow and short windows (as well as having a careful check over everything, today's push is a bit hurried).

cillian64:
I've finished all the things on my TODO list and had another round of sanity-checking and testing.  So it's probably a good time for any final review comments, style issues, things which need explaining better, etc.  Also probably worth checking that the default theming seems reasonable to others.

ahesford (changes requested):
This is getting close, but the overlap calculations can be simplified a bit.

Also, during testing in a nested labwc session, I discovered running without `swaybg` (resulting in a default, undecorated black background) or with `swaybg -c '#000000'` and a shadow of any color will completely skip the Gaussian rolloff and draw completely solid shadows with the configured color (including transparency). See attached for an exaggerated shadow with color `#ff0000ff` and size `256`.

badshadow

Consolatis:
Something doesn't seem right here:
20240418_17h39m41s_grim

If using '#00000080` that doesn't happen.

Consolatis:
20240418_18h14m39s_grim

When a window is snapped to a side with an adjacent output (e.g. on the left of the right monitor) the shadow bleeds into the adjacent output, likely the same thing happens with panels or other exclusive zone clients. I would be fine accepting this for now because the shadows don't have an input region (the input region was the reason that `ssd_extents.c` has to do some annoying clipping to the usable area of the output). So this is only a visual surprise rather than a functional one.

src/ssd/ssd_shadow.c
```diff
@@ -62,21 +62,38 @@ set_shadow_part_geometry(struct ssd_part *part, int width, int height,
...
+	 * aberrations in the shadow shape where corners meet but it's not too
+	 * noticeable.
 	 */
-	int horizontal_overlap = inset > (width / 2 - 1) ? inset - width / 2 : 0;
-	int vertical_overlap = inset > (height / 2 - 1) ? inset - height / 2 : 0;
+	bool hide_topbottom = width <= inset * 2;
+	bool hide_sides = height <= inset * 2;
+	/* These values are the overlap on each corner (half total overlap) */
+	int horizontal_overlap = hide_topbottom ? inset - width / 2 : 0;
+	int vertical_overlap = hide_sides ? inset - height / 2 : 0;
```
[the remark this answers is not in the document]
cillian64: Agreed this is clearer.

cillian64:
I've fixed the premultiplied alpha colors in a slightly simpler way to avoid the division by zero, and applied the refactors in `set_shadow_part_geometry()`.

> When a window is snapped to a side with an adjacent output (e.g. on the left of the right monitor) the shadow bleeds into the adjacent output, likely the same thing happens with panels or other exclusive zone clients.

Tiled windows do indeed cast shadows over the panel.  Also, if you have two windows tiled side-by-side then the active one casts a shadow over the inactive one, this could be considered a feature but I think it looks weird.  Perhaps the best solution would be to disable shadows on windows which are snapped (as suggested by heroin-moose some time back) -- this is what both GNOME and Wayfire do.

Consolatis:
> Tiled windows do indeed cast shadows over the panel. Also, if you have two windows tiled side-by-side then the active one casts a shadow over the inactive one, this could be considered a feature but I think it looks weird. Perhaps the best solution would be to disable shadows on windows which are snapped (as suggested by heroin-moose some time back) -- this is what both GNOME and Wayfire do.

That in turn might look weird for users who use [gaps](https://labwc.github.io/labwc-config.5.html#entry_core_gap). I'd be fine with both approaches for now though (just letting it bleed into adjacent outputs / panels or disabling shadows on `view_is_tiled(view)`, that one also catches `SnapToRegion`).

cillian64:
I'd wager more people would be annoyed by shadows on snapped windows than by their absence (I fall in the former camp now I've noticed it!) so I'll go with disabling shadows when `view_is_tiled(view)`.  If it turns out to be particularly controversial then it could be a config option easily enough.

johanmalm:
Re: pre-multiplied colours

- @Consolatis: thanks for catching!
- @cillian64: this was changed recently which is probably why it didn't happen until after rebase against master. Ref: https://github.com/labwc/labwc/commit/c35ba02ffad0e68f5a9d0d19361e965fd15c4e3e


Re: bleeding into adjacent output

Agree, let's leave that for now.

> I'll go with disabling shadows when `view_is_tiled(view)`

Fine. I think it's a sane default, and we can add options later to fine-tune if necessary.

@cillian64 - is it ready for merge from your perspective?

johanmalm:
Have tested latest - with two observations:

(1) The shadows are pretty tight around the corners. Are we happy with that? Have just set color to `#ff000080` to make it more visible.

shadow1

(2) The shadows are still visible in tiled state (used `SnapToEdge` action)

shadow2

ahesford:
> (1) The shadows are pretty tight around the corners. Are we happy with that? Have just set color to `#ff0000ff` to make it more visible.

I personally find the "tight" appearance a little weird, but this is a consequence of tying the Gaussian roll-off to the `inset` parameter. I think the way around this would be defining a second parameter that would clip the Gaussian at some peak value less than unity.

The main reason that I wanted to keep the "inset" parameter out of user configurations is because I don't want to lock us into supporting that parameter indefinitely if we decide there are better ways to tune the behavior. I think we should live with the current corner behavior for now, and let those interested propose refinements in the future.

cillian64:
> @cillian64 - is it ready for merge from your perspective?

There are still a couple more things:

> The shadows are pretty tight around the corners. Are we happy with that? Have just set color to #ff0000ff to make it more visible.

I'm also not entirely sure that the inset/corner radius at the moment is correct.  I left the value of `SSD_SHADOW_INSET` at `0.5` which I liked with the quadratic dropoff but feels like a bit too much with the gaussian dropoff.  I was going to wait for some more people to weigh in on this.  I think 0.5 might be too high, maybe we want more like 0.25 or 0.3.  (With this reduced inset you'll want to reduce shadow opacity to something like `#00000080` for active and `#00000040` for inactive.)

> The shadows are still visible in tiled state (used SnapToEdge action)

I haven't implemented hiding-when-snapped yet, but I'll do that now.

Consolatis:
> May be a bit late here, but I'd like to mention [SceneFX](https://github.com/wlrfx/scenefx), an effort of the creators of SwayFX, to make it's FX renderer available to other Wayland compositors. It is also now used in SwayFX itself in the main branch. So far it supports box shadows, rounded corners and kawase blur. It would be nice to collaborate with other Wayland compositors on this to have a nice unified solution for effects.
> 
> Perhaps LabWC could benefit from it as well?

Only speaking for myself here but I'd prefer to keep the dependencies limited to wlroots in this regard. Labwc also needs to run at a decent performance when using the pixman renderer. What is the likelihood of scenefx trying to upstream their changes into wlroots (e.g. `render/*/pass.c` and scene API wrappers)? That would be the easiest way so other wlroots based compositors could benefit from the work and it would reduce the maintenance overhead required from the scenefx team.

ahesford:
I have no interest in adopting SceneFX. There are plenty of eye-candy compositors to satisfy the market; we aren't one of them and I don't think we should try to be. Using a library that emphasizes "eye-candy effects" seems to be a mismatch that will either distract from the focus of labwc or result in most of the functionality of the library going unused. Furthermore, the project is still in its infancy and doesn't appear to have tagged any releases.

Shinyzenith:
@ahesford I totally see where you're coming from -- However, isn't this a eye-candy PR that is being well received from what I can tell? Not trying to be rude, I am just a little confused.

Consolatis:
Please keep the discussion here on-topic of this PR. If you want to talk about labwc potentially using sceneefx feel free to open an issue or start a discussion.

Consolatis:
There are some issues around `Reconfigure` support:
- tiled windows will suddenly have shadows after `Reconfigure`
- shadows are not disable after a `Reconfigure` before resizing the window

Consolatis:
Yep, thanks for your work on this feature and being that responsive to feedback, really appreciated.

ahesford:
I've enabled drop shadows for my regular sessions and find the effect very pleasant. Thanks for putting up with my relentless criticism!

cillian64:
Thanks for merging and thanks to everyone for all the testing and feedback!  This did turn out a lot more complicated than I had expected but I'm very happy with the end result (both code and appearance).

Issue #1577, opened 2024-03-01
New labwc release: 0.7.1


Consolatis:
We just released labwc version 0.7.1 which depends on wlroots 0.17.x.
The changelog can be seen [here](https://github.com/labwc/labwc/blob/master/NEWS.md#071).

We are very happy to see labwc showing up in some distro user repositories and are grateful to the respective maintainers:

- @ptrcnull (Alpine)
- @narrat (Arch)
- @b1rger (Debian)
- @jbeich (FreeBSD)
- @AndersonTorres (NixOS)
- @adcdam (Slackware)
- @bdantas (Tiny Core Linux)
- @Visone-Selektah (Venom Linux)
- @tranzystorekk (Void Linux)

---

This release adds two new labwc icons:
- `labwc.svg`
- `labwc-symbolic.svg`

and adds `Icon=labwc` to the labwc session file.
By default the icons are installed into `get_option('datadir') / 'icons/hicolor/scalable/apps'`.

Some login / display managers may expect a specific naming of the icon in another directory, lightdm-gtk-greeter for example seems to expect the icon as `$data_dir/icons/hicolor/scalable/places/labwc_badge-symbolic.svg`.

---

If you don't want to be mentioned on future releases or have any suggestions or questions regarding packaging feel free to post here. If you are maintaining labwc for another distro and want to be notified on packaging related changes for future releases feel free to post as well.

Visone-Selektah:
Hi!
Congrats to all team!!!

AndersonTorres:
Done :)
https://github.com/NixOS/nixpkgs/pull/292679

dirk-hub:
Great news and thank you all for your hard work! 🥰

PR #1650, opened 2024-03-22
CONTRIBUTING.md: add Gentoo overlay maintainer

Consolatis:
Ref
- https://github.com/bsd-ac/wayland-desktop/issues/78

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,7 @@ Some distributions carry labwc in their repositories or user repositories.
 - @narrat (Arch)
 - @b1rger (Debian)
 - @jbeich (FreeBSD)
+- @epsilon-0 (Gentoo)
 - @AndersonTorres (NixOS)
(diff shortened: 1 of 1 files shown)


tfemby:
First off, thanks for finding my issue there. Life's been too busy for me lately to properly dig into the ebuild there and provide a version bump myself.

Regarding  the change to the contribution page, I think it would be a good idea to include "via `wayland-desktop` overlay". Gentoo has a system to enable/disable these types of overlays that aren't in the main `gentoo` overlay tree.

The Gentoo wiki also lists LabWC on the following wiki page as well. https://wiki.gentoo.org/wiki/List_of_software_for_Wayland

Consolatis:
We don't specific that `Arch` in fact means `AUR` either and the list is basically only used for writing the packaging notifications like #1577. So I don't think it is that important. If labwc ends up in some non-overlay Gentoo repo we could just add another maintainer that also simply states `(Gentoo)`.

PR #1652, opened 2024-03-23
overlay: add snap-to-edge overlay

tokyo4j:
...and unifies region overlay and snap-to-edge overlay into `overlay.c`.

Snap-to-edge overlay is delayed for 150ms to prevent flickering when the view is dragged from an output to another (demo in discussion #1613).

This also fixes a bug that region overlay is not shown when a modifier key is re-pressed.

I removed `seat->region_active`, so my changes will incur some extra calculations. It's a TODO for now.

One thing I'm not sure is where to put `edge_from_cursor()` which is called from `snap_to_edge()` and `overlay_show()`.

### DIFF
--- a/src/overlay.c
+++ b/src/overlay.c
@@ -1,0 +1,172 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <wlr/render/pixman.h>
+#include "labwc.h"
+#include "overlay.h"
+#include "view.h"
+
+static void
+create_overlay(struct seat *seat)
+{
+	assert(!seat->overlay.tree);
+
+	struct server *server = seat->server;
+	struct wlr_scene_tree *parent = wlr_scene_tree_create(&server->scene->tree);
+
+	seat->overlay.tree = parent;
+	wlr_scene_node_set_enabled(&parent->node, false);
+	if (!wlr_renderer_is_pixman(server->renderer)) {
+		/* Hardware assisted rendering: Half transparent overlay */
+		float color[4] = { 0.25, 0.25, 0.35, 0.5 };
+		seat->overlay.rect = wlr_scene_rect_create(parent, 0, 0, color);
+	} else {
+		/* Software rendering: Outlines */
+		int line_width = server->theme->osd_border_width;
+		float *colors[3] = {
+			server->theme->osd_bg_color,
+			server->theme->osd_label_text_color,
+			server->theme->osd_bg_color
+		};
+		seat->overlay.pixman_rect = multi_rect_create(parent, colors, line_width);
+	}
+}
+
+static void
+cancel_pending_overlay(struct overlay *overlay)
+{
+	if (!overlay->timer) {
+		return;
+	}
+	wl_event_source_remove(overlay->timer);
+	overlay->timer = NULL;
+	overlay->view = NULL;
+	overlay->box = (struct wlr_box){0};
+}
+
+static void
+show_overlay(struct seat *seat, struct view *view, struct wlr_box *box)
+{
+	if (!seat->overlay.tree) {
+		create_overlay(seat);
+	}
+
+	/* Update overlay */
+	struct server *server = seat->server;
+	struct wlr_scene_node *node = &seat->overlay.tree->node;
+	if (!wlr_renderer_is_pixman(server->renderer)) {
+		/* Hardware assisted rendering: Half transparent overlay */
+		wlr_scene_rect_set_size(seat->overlay.rect,
+			box->width, box->height);
+	} else {
+		/* Software rendering: Outlines */
+		multi_rect_set_size(seat->overlay.pixman_rect,
+			box->width, box->height);
+	}
+	if (node->parent != view->scene_tree->node.parent) {
+		wlr_scene_node_reparent(node, view->scene_tree->node.parent);
+		wlr_scene_node_place_below(node, &view->scene_tree->node);
+	}
+	wlr_scene_node_set_position(node, box->x, box->y);
+	wlr_scene_node_set_enabled(node, true);
+
+	cancel_pending_overlay(&seat->overlay);
+}
+
+static int
+handle_overlay_timeout(void *data)
+{
+	struct seat *seat = data;
+	show_overlay(seat, seat->overlay.view, &seat->overlay.box);
+	return 0;
+}
+
+static void
+show_overlay_delayed(struct seat *seat, struct view *view, struct wlr_box *box)
+{
+	if (seat->overlay.timer
+			&& seat->overlay.view == view
+			&& wlr_box_equal(box, &seat->overlay.box)) {
+		return;
+	}
+	cancel_pending_overlay(&seat->overlay);
+	seat->overlay.view = view;
+	seat->overlay.box = *box;
+	seat->overlay.timer = wl_event_loop_add_timer(
+		seat->server->wl_event_loop,
+		handle_overlay_timeout, seat);
+	wl_event_source_timer_update(seat->overlay.timer, 150);
+}
+
+void
+overlay_show(struct seat *seat, struct view *view)
+{
+	struct server *server = seat->server;
+
+	/*
+	 * TODO: cache return value of regions_from_cursor() or
+	 * edge_from_cursor() to eliminate duplicated calls to show_overlay() /
+	 * show_overlay_delayed().
+	 */
+	if (regions_should_snap(server)) {
+		/* Region overlay */
+		struct region *region = regions_from_cursor(server);
+		if (region) {
+			show_overlay(seat, view, &region->geo);
+			return;
+		}
+	} else {
+		/* Snap-to-edge overlay */
+		struct output *output;
+		enum view_edge edge = edge_from_cursor(seat, &output);
+		if (edge != VIEW_EDGE_INVALID) {
+			/* TODO: share logic with view_get_edge_snap_box() */
+			struct wlr_box box =
+				output_usable_area_in_layout_coords(output);
+			switch (edge) {
+			case VIEW_EDGE_RIGHT:
+				box.x += box.width / 2;
+				/* fallthrough */
+			case VIEW_EDGE_LEFT:
+				box.width /= 2;
+				break;
+			case VIEW_EDGE_DOWN:
+				box.y += box.height / 2;
+				/* fallthrough */
+			case VIEW_EDGE_UP:
+				box.height /= 2;
+				break;
+			case VIEW_EDGE_CENTER:
+				/* <topMaximize> */
+				break;
+			case VIEW_EDGE_INVALID:
+				/* not reached */
+				assert(false);
+			}
+			/*
+			 * Delay showing overlay to prevent flickering when
+			 * dragging view across output edges in multi-monitor
+			 * setup
+			 */
+			show_overlay_delayed(seat, view, &box);
+			return;
+		}
+	}
+	overlay_hide(seat);
+}
+
+void
+overlay_hide(struct seat *seat)
+{
+	cancel_pending_overlay(&seat->overlay);
+
+	struct server *server = seat->server;
+	struct wlr_scene_node *node = &seat->overlay.tree->node;
+	if (!node) {
+		return;
+	}
+
+	wlr_scene_node_set_enabled(node, false);
+	if (node->parent != &server->scene->tree) {
+		wlr_scene_node_reparent(node, &server->scene->tree);
+	}
+}
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -47,6 +47,7 @@
 #include "config/rcxml.h"
 #include "input/cursor.h"
 #include "input/ime.h"
+#include "overlay.h"
 #include "regions.h"
 #include "session-lock.h"
 #if HAVE_NLS
@@ -159,9 +160,8 @@ struct seat {
 		struct wlr_scene_tree *icons;
 	} drag;
 
-	/* Private use by regions.c */
-	struct region *region_active;
-	struct region_overlay region_overlay;
+	/* Private use by overlay.c */
+	struct overlay overlay;
 	/* Used to prevent region snapping when starting a move with A-Left */
 	bool region_prevent_snap;
 
@@ -479,6 +479,8 @@ void seat_output_layout_changed(struct seat *seat);
 void interactive_begin(struct view *view, enum input_mode mode, uint32_t edges);
 void interactive_finish(struct view *view);
 void interactive_cancel(struct view *view);
+/* Possibly returns VIEW_EDGE_CENTER if <topMaximize> is yes */
+enum view_edge edge_from_cursor(struct seat *seat, struct output **dest_output);
 
 void output_init(struct server *server);
 void output_manager_init(struct server *server);
--- a/include/overlay.h
+++ b/include/overlay.h
@@ -1,0 +1,25 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_OVERLAY_H
+#define LABWC_OVERLAY_H
+
+#include <wlr/util/box.h>
+#include "common/graphic-helpers.h"
+
+struct overlay {
+	struct wlr_scene_tree *tree;
+	union {
+		struct wlr_scene_rect *rect;
+		struct multi_rect *pixman_rect;
+	};
+
+	/* For delayed overlay */
+	struct view *view;
+	struct wlr_box box;
+	struct wl_event_source *timer;
+};
+
+/* Calls overlay_hide() internally if the view is not to be snapped */
+void overlay_show(struct seat *seat, struct view *view);
+void overlay_hide(struct seat *seat);
+
+#endif
--- a/include/regions.h
+++ b/include/regions.h
@@ -25,14 +25,6 @@ struct region {
 	} center;
 };
 
-struct region_overlay {
-	struct wlr_scene_tree *tree;
-	union {
-		struct wlr_scene_rect *overlay;
-		struct multi_rect *pixman_overlay;
-	};
-};
-
 /* Returns true if we should show the region overlay or snap to region */
 bool regions_should_snap(struct server *server);
 
@@ -72,8 +64,5 @@ void regions_destroy(struct seat *seat, struct wl_list *regions);
 /* Get output local region from cursor or name, may be NULL */
 struct region *regions_from_cursor(struct server *server);
 struct region *regions_from_name(const char *region_name, struct output *output);
-
-void regions_show_overlay(struct view *view, struct seat *seat, struct region *region);
-void regions_hide_overlay(struct seat *seat);
 
 #endif /* LABWC_REGIONS_H */
--- a/src/debug.c
+++ b/src/debug.c
@@ -136,10 +136,10 @@ get_special(struct server *server, struct wlr_scene_node *node)
 	if (node == &server->seat.drag.icons->node) {
 		return "seat->drag.icons";
 	}
-	if (server->seat.region_overlay.tree
-			&& node == &server->seat.region_overlay.tree->node) {
+	if (server->seat.overlay.tree
+			&& node == &server->seat.overlay.tree->node) {
 		/* Created on-demand */
-		return "seat->region_overlay";
+		return "seat->overlay";
 	}
 	if (server->seat.input_method_relay->popup_tree
 			&& node == &server->seat.input_method_relay->popup_tree->node) {
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -225,16 +225,7 @@ process_cursor_move(struct server *server, uint32_t time)
 	resistance_move_apply(view, &dx, &dy);
 	view_move(view, dx, dy);
 
-	/* Region overlay */
-	if (!regions_should_snap(server)) {
-		return;
-	}
-	struct region *region = regions_from_cursor(server);
-	if (region) {
-		regions_show_overlay(view, &server->seat, region);
-	} else {
-		regions_hide_overlay(&server->seat);
-	}
+	overlay_show(&server->seat, view);
 }
 
 static void
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -82,9 +82,9 @@ keyboard_modifiers_notify(struct wl_listener *listener, void *data)
 		seat->region_prevent_snap = false;
 	}
 
-	if (server->osd_state.cycle_view || server->grabbed_view
+	if (server->osd_state.cycle_view
 			|| seat->workspace_osd_shown_by_modifier) {
-		if (!keyboard_any_modifiers_pressed(wlr_keyboard))  {
+		if (!keyboard_any_modifiers_pressed(wlr_keyboard)) {
 			if (server->osd_state.cycle_view) {
 				if (key_state_nr_bound_keys()) {
 					should_cancel_cycling_on_next_key_release = true;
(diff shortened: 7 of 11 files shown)


src/overlay.c
```diff
@@ -0,0 +1,172 @@
...
+		return;
+	}
+	cancel_pending_overlay(&seat->overlay);
+	seat->overlay.view = view;
+	seat->overlay.box = *box;
+	seat->overlay.timer = wl_event_loop_add_timer(
+		seat->server->wl_event_loop,
+		handle_overlay_timeout, seat);
+	wl_event_source_timer_update(seat->overlay.timer, 150);
+}
```
Consolatis, line 98: I don't think we need to recreate the timer here. Suggest something like this instead:
```c
	if (seat->overlay.timer
			&& seat->overlay.view == view
			&& wlr_box_equal(box, &seat->overlay.box)) {
		return;
	}
	seat->overlay.view = view;
	seat->overlay.box = *box;
	if (!seat->overlay.timer) {
		seat->overlay.timer = wl_event_loop_add_timer(
			seat->server->wl_event_loop,
			handle_overlay_timeout, seat);
	}
	wl_event_source_timer_update(seat->overlay.timer, 150);
```

src/overlay.c
```diff
@@ -0,0 +1,172 @@
...
+				/* fallthrough */
+			case VIEW_EDGE_UP:
+				box.height /= 2;
+				break;
+			case VIEW_EDGE_CENTER:
+				/* <topMaximize> */
+				break;
+			case VIEW_EDGE_INVALID:
+				/* not reached */
+				assert(false);
```
Consolatis, line 143: The assert can be replaced with a simple `break;`, we do check for `!= VIEW_EDGE_INVALID` directly above anyway.

src/overlay.c
```diff
@@ -0,0 +1,172 @@
...
+	overlay_hide(seat);
+}
+
+void
+overlay_hide(struct seat *seat)
+{
+	cancel_pending_overlay(&seat->overlay);
+
+	struct server *server = seat->server;
+	struct wlr_scene_node *node = &seat->overlay.tree->node;
```
Consolatis, line 163: This might segfault if we call `overlay_hide()` before `show_overlay()`, suggest simply returning if `!seat->overlay.tree`

Consolatis (commented):
Thanks for the PR!
Just a quick review, will look into this in more depth in the next days.

src/overlay.c
```diff
@@ -0,0 +1,172 @@
...
+void
+overlay_show(struct seat *seat, struct view *view)
+{
+	struct server *server = seat->server;
+
+	/*
+	 * TODO: cache return value of regions_from_cursor() or
+	 * edge_from_cursor() to eliminate duplicated calls to show_overlay() /
+	 * show_overlay_delayed().
+	 */
```
Consolatis, line 109: I assume we could just add some embedded state struct to struct overlay, something like
```c
struct overlay {
	[..]
	struct {
		struct region *active_region;
		enum view_edge active_edge;
	} state;
}
```

We just have to ensure that those are reset to `VIEW_EDGE_INVALID` / `NULL` when
- we hide the overlay
- the view is destroyed (so likely already done via `overlay_hide()`)
- the region is destroyed (`regions_reconfigure_output()`).

src/overlay.c
```diff
@@ -0,0 +1,172 @@
...
+	if (regions_should_snap(server)) {
+		/* Region overlay */
+		struct region *region = regions_from_cursor(server);
+		if (region) {
+			show_overlay(seat, view, &region->geo);
+			return;
+		}
+	} else {
```
tokyo4j, line 117: I found this "else" should be removed, otherwise snap-to-edge overlay is not shown if any modifier key is pressed.

src/overlay.c
tokyo4j: Done.

src/overlay.c
tokyo4j: Done.
Actually, this doesn't segfault because node is never dereferenced, but it depended on the assumption of `&tree.node == &tree`.

src/overlay.c
tokyo4j: I decided to separate the logic of getting snapped box into `get_edge_snap_box()`, so I will keep the assert there for readability.

src/overlay.c
tokyo4j: Done.

src/overlay.c
Consolatis: but `seat->overlay.tree` could be null theoretically, so any access to `->node` would already be trying to access a NULL pointer.

Consolatis:
Really nice stuff. The overlay implementation looks very clean. I'll give it a test run and also look into the keyboard.c changes in more detail tomorrow.

tokyo4j:
Maybe I can remove `view` from the arguments of `overlay_show()` then just use `seat->server->grabbed_view` and remove `pending.view` from `struct overlay`.

Consolatis:
> Maybe I can remove `view` from the arguments of `overlay_show()` then just use `seat->server->grabbed_view` and remove `pending.view` from `struct overlay`.

Yes, that sounds sensible.

src/overlay.c
tokyo4j: I assumed `&(pointer)->(field)` is translated to `(pointer) + (offset of field)` so dereference doesn't happen at this point. I wonder if it is defined by C specification, though.

ahesford: That would actually make things worse, because if `pointer` were null and the compiler just blindly added an offset of a struct field, attempting to do would leave the value of `addr = &pointer->field` at a non-null that might circumvent later sanity checks on `addr` even though it is invalid. That would also be hard to debug.

I would hope that a compiler would actually try to deference a pointer when you use `->`, even if you are just computing an address of a field, to trigger a segfault where it would be expected rather than let a bogus address float around.

tokyo4j:
I removed `overlay.pending` and add `overlay.active.output` then used `overlay.active.edge` and `overlay.active.output` to show delayed snap-to-edge overlay instead. I also added handling for output destruction.
This fixes a bug that snap-to-edge overlay is not updated if the cursor moves across outputs but on the same edge.
Other changes are:
- call `overlay_hide()` on region destruction
- change location of `overlay_show()` in `keyboard.c`.

Consolatis:
Tested a bit nested, 150ms seems a bit much too me, 100ms might seem better. But again, I only tested it nested and the experience might be different on a true DRM session.

Maybe we could alternatively make the timeout a bit smarter and only use the timer if there is another output in the edge direction?

E.g. something like if coordinates of the edge is equal to one of the edges in `wlr_output_layout_get_box(layout, NULL, &layout_box)` we don't use _delayed but show the overlay directly? Another option might be `view_get_adjacent_output()` or parts of it.

The keyboard.c changes look good to me. I noticed that `overlay_show()` is used for both, showing and hiding the overlay. Maybe renaming that function to `overlay_update()` would be more intuitive?

tokyo4j:
> I've actually had the opposite experience, if I'm not fast enough when snapping a window the flicker of the overlay popping up and then the window resizing immediately after can be just a bit annoying. Not a huge deal, but I thought I'd throw in my opinion.

Thank you for testing!
Do you mean flicker like below?

https://github.com/labwc/labwc/assets/22029524/f9593f31-d468-4af1-a8de-8ef7edfe912c

Consolatis:
I have to say I kind of like the feel with `WLR_RENDERER=pixman`.

@Flrian @tokyo4j @johanmalm @ahesford (and others) what are your thoughts on configuration options and defaults, should we
- show the snap overlay by default
- keep the look of the overlay depend on the renderer, e.g. some outlines with pixman and the half transparent overlay with gles2 / vulkan or allow users to configure that as well? This was originally implemented to improve performance when running with the pixman renderer but it may be worth to allow users to use the outlines on gles2 / vulkan as well
- what should we set the timeout default to (if there is an adjacent output)
- should we keep the show-instant-unless-adjacent output option as currently implemented or instead go back to always using the delay with something like 500ms to reduce the flicker during usual snap operations?

This really is a feature that should "feel right" for most users out of the box and if it doesn't, at least allow the user to make it feel right (including the ability to disable it completely) so the more opinions we have here the better.

My current thought would be something like
```xml
<snapping>
	<range>1</range>
	<preview>outlines | area | none </preview> (not sure what to set the default here to)
	<previewTimeout>150</previewTimeout> (same here, 150? 500? 0?)
</snapping>
```

tokyo4j:
> - show the snap overlay by default

I prefer showing it by default.

> - keep the look of the overlay depend on the renderer, e.g. some outlines with pixman and the half transparent overlay with gles2 / vulkan or allow users to configure that as well? This was originally implemented to improve performance when running with the pixman renderer but it may be worth to allow users to use the outlines on gles2 / vulkan as well

I guess we should make it configurable. Actually, I prefer outlines for snap-to-edge overlay, too.
I'm not sure there should be separate settings for region overlay and for snap-to-edge overlay, though.

Maybe we should also allow users to configure the color and opacity of area overlay so users can choose less flashy color depending on their background image or `wlr-gamma-control` settings.

> - what should we set the timeout default to (if there is an adjacent output)
> - should we keep the show-instant-unless-adjacent-output option as currently implemented or instead go back to always using the delay with something like 500ms to reduce the flicker during usual snap operations?

Related to this, I found flicker on unmaximizing window by dragging it on its titlebar is a bit annoying: 

https://github.com/labwc/labwc/assets/22029524/2a8de654-0d96-450f-99a7-7ae9deef9411

As this problem happens also on edges which don't have adjacent outputs, maybe we should (optionally) apply some delay for all edges regardless of whether they have adjacent outputs. I think the timeout should be configurable, too.

I find no reasons to apply delay for region snapping, but making it possible is a piece of cake as it's already done for snap-to-edge overlay. So making it possible to configure them separately would be better.

I have no idea what value should be the default though. My preference is `outline` preview with `150`ms delay for snap-to-edge and `area` or `outline` preview with `0`ms delay for region snapping.

So my current thought is like:
```xml
<snapping>
	<range>1</range>

	<preview type="region">
		<show>yes</show>
		<style>area</style>
		<color>#80808080</color> (should be in themerc?)
		<timeout>0</timeout>
	</preview>
	<preview type="edge">
		<show>yes</show>
		<style>outlines</style>
		<color>#ffffff</color>
		<timeout>150</timeout>
	</preview>
</snapping>
```

ahesford:
I finally had time to test this and am really happy with the implementation and its feel. Regarding issues raised above:
- Snapping previews should definitely be enabled by default.
- The timer feels OK to me at 100 ms, but making it configurable would be nice.
- I don't really care if we implement a timer for region snapping. Region snapping is opt-in by holding a modifier, which is more "determined" than edge snapping, so having it flash between regions quickly as you cross between boundaries should be acceptable.
- Allowing configurable shading/border choices, with a default to shading for hardware-accelerated rendering and borders for software rendering, seems like the best choice.
- I don't mind if we only apply the timer to shared, "interior" edges only and show previews instantly for exterior edge, but I understand how the un-snap behavior can be annoying. Maybe the right behavior is to apply the timer to edge previews on exterior edges *only if the window is already snapped to that edge*. This may or may not be straightforward to figure and I haven't thought about the details---if tracking this behavior is too much hassle, maybe applying the edge-snap timer uniformly to all edges is the best compromise.

These changes have me thinking about some other behavior changes that are out of scope of this PR, but I might think about after this is merged:
- If previews are enabled, it might be nice to only allow edge snapping if the preview is currently showing. In other words, honor the preview delay as a "wait for snapping" delay. 
- An option to disable all edge snapping for shared, interior edges might be nice and fit better with my own expectations. That might require some accommodation here as well, but is not something we should think about at the moment.

Consolatis:
> Maybe we should also allow users to configure the color and opacity of area overlay so users can choose less flashy color depending on their background image or `wlr-gamma-control` settings.

That is a good point, I am not sure we handle `#RRGGBBAA` correctly currently, AFAIR we need pre-multiplied  colors internally so we'd need to convert them. In any case this indeed feels like something for `themerc` (or `themerc-override`). Maybe there is even something already defined by Openbox for that that we could reuse.

> * I don't really care if we implement a timer for region snapping. Region snapping is opt-in by holding a modifier, which is more "determined" than edge snapping, so having it flash between regions quickly as you cross between boundaries should be acceptable.

Personally, I don't think we need the timer for region snapping due to it being opt-in. I feel like that would actually be counter productive as it would make labwc feel "laggy" and in most cases I'd assume the user actually needs the overlay to see which region the view will be snapped to (other than in the edge-to-snap case).

> I guess we should make it configurable. Actually, I prefer outlines for snap-to-edge overlay, too.
> I'm not sure there should be separate settings for region overlay and for snap-to-edge overlay, though.

> * Allowing configurable shading/border choices, with a default to shading for hardware-accelerated rendering and borders for software rendering, seems like the best choice.

Using `auto` would be backwards compatible behavior with what we are currently doing. We could move this to theme settings as well, then we could use something like `snap.overlay.*.type: auto | outlines | area | none` to match both or `snap.overlay.edge.type` to match just the snap-to-edge case. Same story for `.color`.

> Related to this, I found flicker on unmaximizing window by dragging it on its titlebar is a bit annoying:

> * I don't mind if we only apply the timer to shared, "interior" edges only and show previews instantly for exterior edge, but I understand how the un-snap behavior can be annoying. Maybe the right behavior is to apply the timer to edge previews on exterior edges _only if the window is already snapped to that edge_. This may or may not be straightforward to figure and I haven't thought about the details---if tracking this behavior is too much hassle, maybe applying the edge-snap timer uniformly to all edges is the best compromise.

The more I think about it the more I feel like just going back to applying it for all snap-to-edge scenarios is the better option. Less complexity in the code and we don't have to deal with the corner cases. If we go that route: sorry that my initial suggestion caused unnecessary code to be written.

Another thought I had regarding that topic: the edge snap overlay is especially useful for new users as they can discover the snap-to-edge feature that way. However, after the initial discovery the usefulness for the overlay is drastically reduced from an UX standpoint. Thus my thought with increasing it to 500ms to reduce the chances of it causing unwanted behavior like accidental flashing.

This is just my personal opinion though so other standpoints are very welcome.

> These changes have me thinking about some other behavior changes that are out of scope of this PR, but I might think about after this is merged:
> * If previews are enabled, it might be nice to only allow edge snapping if the preview is currently showing. In other words, honor the preview delay as a "wait for snapping" delay.

This is an interesting idea, it would slow down the snapping ability by users but they could disable the overlays and get back to snap in an instant. Lets keep that in the back of our minds for potential follow-up PRs.

tokyo4j:
Thank you for suggestions!
I totally agree with @Consolatis.

With my latest fixups,
- Add `<snapping><previewTimeout edgeType="interior|exterior">` and allow configuring timeouts for interior/exterior edge snapping separately
  - I added `edgeType` (see 84e9eb73d331c7728df694cd7630864f015815ca for diff.) to see if it's useful, but it's totally okay to revert this, as setting timeout to 100ms for both edge types is just fine for me.
- Set default timeout to 500ms to never hurt UX
- No timeout for region-snapping

And in the following PRs:
- Make preview style configurable
- Make preview color configurable
  - We need to handle alpha properly before this. If we allow alpha for themes in general, maybe #1631 is problematic?

And about @ahesford's suggestions:
- Allow snapping only when preview is shown
- Add option to disable edge-snapping on interior edges

I feel both suggestions are interesting and worth implementing.
The former might be natural for some users and can be done with small additional checks like `if (seat->overlay.active.edge != VIEW_EDGE_INVALID)`
Also, the latter can be done with additional few lines of code to my latest fixup.

Consolatis:
Thanks for keeping to work on this, I'll try to do a another test run during the weekend but I think this is coming together really nicely!

src/config/rcxml.c
```diff
@@ -735,6 +735,12 @@ entry(xmlNode *node, char *nodename, char *content)
 	/* current <theme><font place=""></font></theme> */
 	static enum font_place font_place = FONT_PLACE_NONE;
 
+	static enum {
+		EDGE_TYPE_INTERIOR = 1,
+		EDGE_TYPE_EXTERIOR = 2,
+		EDGE_TYPE_BOTH = 1 | 2,
```
ahesford, line 741:
```suggestion
		EDGE_TYPE_INTERIOR = (1 << 0),
		EDGE_TYPE_EXTERIOR = (1 << 1),
		EDGE_TYPE_BOTH = (EDGE_TYPE_INTERIOR | EDGE_TYPE_EXTERIOR),
```

src/config/rcxml.c
```diff
@@ -906,6 +912,19 @@ entry(xmlNode *node, char *nodename, char *content)
...
+		} else if (!strcasecmp(content, "exterior")) {
+			edge_type = EDGE_TYPE_EXTERIOR;
+		}
+	} else if (!strcasecmp(nodename, "previewTimeout.snapping")) {
+		if (edge_type & EDGE_TYPE_INTERIOR) {
+			rc.snap_preview_interior_timeout = atoi(content);
+		}
+		if (edge_type & EDGE_TYPE_EXTERIOR) {
+			rc.snap_preview_exterior_timeout = atoi(content);
+		}
```
ahesford, line 927: The word "timeout" here is a misnomer; I suggest the "delay" instead. Also, we'll need documentation on these user-facing options.

ahesford (changes requested):

src/config/rcxml.c
```diff
@@ -906,6 +912,19 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.window_edge_strength = atoi(content);
 	} else if (!strcasecmp(nodename, "range.snapping")) {
 		rc.snap_edge_range = atoi(content);
+	} else if (!strcasecmp(nodename, "edgeType.previewTimeout.snapping")) {
+		if (!strcasecmp(content, "interior")) {
+			edge_type = EDGE_TYPE_INTERIOR;
+		} else if (!strcasecmp(content, "exterior")) {
+			edge_type = EDGE_TYPE_EXTERIOR;
+		}
```
Consolatis, line 920: Can we come up with a more intuitive config name for non-devs (regarding `interior` and `exterior`) ?
If we ship this in the next release we really don't want to ever change it to something else so now would be the time.

Just to reiterate: this allows changing the timeout based on the fact that there is another adjacent output in the direction of the edge.

src/config/rcxml.c
Consolatis: Good call, another dev vs user terminology difference here.

Consolatis:
Just had another test nested and it feels very nice and intuitive to me with the default settings.

Example of how a custom config could look like:
```xml
  <snapping>
    <range>1</range>
    <topMaximize>yes</topMaximize>
    <previewTimeout edgeType="interior">1000</previewTimeout>
    <previewTimeout edgeType="exterior">0</previewTimeout>
    <notifyClient>always</notifyClient>
  </snapping>
```
Do we want to provide a way to disable the whole overlay for interior / exterior / globally?

It was stated earlier in this PR that defining the overlay color (and opacity) should be configurable as well and while I agree that we should support that I think we should move that to a follow-up PR.

src/config/rcxml.c
tokyo4j: Done.

src/config/rcxml.c
tokyo4j: Done.

src/config/rcxml.c
tokyo4j: My random ideas are:
- `edgePosition="inner|outer"`
- `edgeType="withAdjacentOutput|withoutAdjacentOutput"`

docs/rc.xml.all
```diff
@@ -99,6 +99,9 @@
   <snapping>
     <!-- Set range to 0 to disable window snapping completely -->
     <range>1</range>
+		<!-- edgeType is optional -->
+		<previewDelay edgeType="interior">500</previewDelay>
+		<previewDelay edgeType="exterior">500</previewDelay>
```
Consolatis, line 104: `rc.xml.all` uses spaces for some history reason

src/config/rcxml.c
Consolatis: `inner | outer` seems easier to understand to me. Thoughts @johanmalm ?

docs/rc.xml.all
tokyo4j: Done.
The problem is VSCode by default uses `.vscode/settings.json` and ignores `.editorconfig`.
Maybe I'll make a PR to add `.vscode/settings.json` or mention it in `CONTRIBUTING.md`.

tokyo4j:
> Do we want to provide a way to disable the whole overlay for interior / exterior / globally?

I think it would be nice to add an option to disable it globally. Workarounds are (1)setting the delay so long like 100000ms or (2)setting the opacity of the overlay (will be added in following PR) to 0, but something like `<showPreview>no</showPreview>` would be cleaner.

However, I don't think we need to make it configurable differently for internal / external edges. IMO offering an option to snapping at interior edges (like @ahesford said earlier) would be nicer.

src/config/rcxml.c
johanmalm: - `shared` and `layoutPerimeter`
- `betweenOutputs` and `perimeterOfOutputLayout'

docs/rc.xml.all
Consolatis: Still a tab in front on the latest fixup

Consolatis:
> > Do we want to provide a way to disable the whole overlay for interior / exterior / globally?
> 
> I think it would be nice to add an option to disable it globally. Workarounds are (1)setting the delay so long like 100000ms or (2)setting the opacity of the overlay (will be added in following PR) to 0, but something like `<showPreview>no</showPreview>` would be cleaner.

What about something like this?
```xml
  <snapping>
    <range>1</range>
    <topMaximize>yes</topMaximize>
    <preview enabled="yes>
      <delay inner="1000" outer="0" />
    </preview>
    <notifyClient>always</notifyClient>
  </snapping>
```

src/config/rcxml.c
ahesford: I like `edgeType="inner|outer"`, although "shared" and one of "exterior" or "boundary" would be acceptable alternatives. No matter what, we'll have to explain the meaning in the documentation, and I think there's enough mnemonic value in any of these labels to be understood.

ahesford:
> What about something like this?
> 
> ```
>   <snapping>
>     <range>1</range>
>     <topMaximize>yes</topMaximize>
>     <preview enabled="yes">
>       <delay inner="1000" outer="0" />
>     </preview>
>     <notifyClient>always</notifyClient>
>   </snapping>
> ```

I like this syntax. In addition, we could support specification of a negative value to indicate that the associated preview timer is disabled.

docs/rc.xml.all
tokyo4j: Done. Sorry for my oversight.

Consolatis:
> > I like this syntax. In addition, we could support specification of a negative value to indicate that the associated preview timer is disabled.
> 
> Do you mean negative values disable the preview? I already allow the value `0` to disable timer and show the preview immediately. If so, maybe we can omit `<preview enabled="yes|no">`.

I think having a single enabled flag makes it a bit easier to use. If we ever end up with user requests to disable only one of the two overlays we could still implement something like enabled="yes|no|edge|region|both" with "both" being a synonym for "yes".

Consolatis:
Anything left to do? Otherwise this PR is ready to go after squashing from my perspective.

ahesford (approved):
Looks pretty good to me

Consolatis:
Had another quick and everything seems to work as expected thus going for the merge.
Some thing we could potentially tackle in a follow up PR is defining the style (e.g. outlines-only vs half transparent overlay and their colors / opacity) via theme.

Thanks again @tokyo4j for the initial PR and all your work on this!

tokyo4j:
Thank you!

tokyo4j:
I feel `edge_from_cursor()` in `interactive.c` is inappropriate for its naming or the file it's placed, so please modify it if you have better ideas for code organization.

Consolatis:
Yeah, that whole edge geometry calculation and the decision to actually snap or not needs some refactor, even before this PR.
It is spread around `src/view.c`, `src/interactive.c` and now `src/overlay.c`. Some of it also implements things slightly different, e.g. the overlay is not respecting gaps while the final snapping operation is. Then there is also `src/edges.c` which has a slightly different goal but is still related.

johanmalm:
Thanks. Great effort.

PR #1655, opened 2024-03-24
labwc-config(5): clarify applicability of element/attribute

johanmalm:
...agnostic parsing.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -100,20 +100,22 @@ Configuration must be wrapped in a <labwc_config> root-element like this:
 decision to increase config file flexibility and keep code simple. In practical
 terms, this means that `<a><b>c</b></a>` is usually equivalent to `<a b="c" />`.
 However, there are some caveats:
-- In menu.xml, the attributes `id` and `label` *must* be expressed as attributes
-  rather than as separate child tags.
-- Ordering of attributes can sometimes be singificant. For example, when setting
-  fonts via attributes rather than children,
-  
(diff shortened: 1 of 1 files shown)


docs/labwc-config.5.scd
```diff
@@ -100,6 +100,9 @@ Configuration must be wrapped in a <labwc_config> root-element like this:
 decision to increase config file flexibility and keep code simple. In practical
 terms, this means that `<a><b>c</b></a>` is equivalent to `<a b="c" />`.
 
+Note: This does not apply to all aspects of menu.xml, specifically the
+attributes id, label and execute.
```
ahesford, line 104:
```suggestion
decision to increase config file flexibility and keep code simple. In practical
terms, this means that `<a><b>c</b></a>` is usually equivalent to `<a b="c" />`.
However, there are some caveats:
- In menu.xml, the attributes `id` and `label` *must* be expressed as attributes
  rather than as separate child tags.
- Ordering of attributes can sometimes be singificant. For example, when setting
  fonts via attributes rather than children,
  
      <font place="ActiveWindow" name="..." size="..." />

  will set a font for active window titles, while
  
       <font name="..." size="..." place="ActiveWindow" />
       
  will set a font as if `place` were unspecified, because `name` and `size` will be
  processed before `place`.
  
In general, it is recommended that attributes not be used to collapse more than
one level of the configuration hierarchy.
```

johanmalm:
@ahesford Thanks. Pushed a fixup to make it `scdoc` compliant.

Issue #1653, opened 2024-03-23
Regression for lxqt-runner: no focus if not clicked by mouse


stefonarch:
In current master `lxqt-runner` which is launched only by shortcut doesn't receive focus anymore automatically.

Keyboardinteractivity was set originally to `OnDemand` but even setting it to `Exclusive` doesn't change behavior - launching it and typing the typed keys go to the open document or terminal.

Reverting https://github.com/labwc/labwc/commit/6990ff713bc817fba28146617db4c78d843bb019 solves.

Runner is still a PR: https://github.com/lxqt/lxqt-runner/pull/277

tsujan:
Just wanted to add that https://github.com/lxqt/lxqt-runner/pull/277 doesn't have any problem in labwc 0.7.1 — I'm using them — and just sets the layer at the top.

johanmalm:
We've made quite a few changes to layer-shell keyboard-interactivity recently (#1599).

I do really appreciate you reporting this, because there isn't much 'standard implementation' to go on and the protocol is quite vague on [on-demand](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity:entry:on_demand) handling.

I have taken "Request regular keyboard focus semantics" to mean that the surface gets focus when receiving mouse-button-press, but not whenever the surfaces changes its keyboard-interactivity to 'on-demand'. I'm quite happy to revert https://github.com/labwc/labwc/commit/6990ff713bc817fba28146617db4c78d843bb019 because I think it's highly likely that the only time a layer-shell surface changes its interactivity to on-demand is on map (or shortly after).

The only consideration I can think of before reverting is that this means that any layer-shell client with on-demand interactivity will 'grab' the keyboard when started (assuming that's when it sets the interactivity to on-demand, which is most likely the case). Appreciate help in thinking through if there are cases when this is not desirable... to avoid chasing our tail on this one.

As a side-note, I'm in the process of creating a `labwc-layer-shell(5)` man page to describe the implementation.

In terms of the described issue, please could confirm:

1. Is the observation made on first launch, or is the client already running in the background and is being invoked by a keyboard-shortcut (via a unix-socket presumably?)
2. What is the context of choosing on-demand rather than exclusive for it. Apologies - I've never used `lxqt-runner` and can't get it to work on my system :blush: (didn't try for that long though).
[...]

PR #1656, opened 2024-03-24
layer: try to set keyboard focus on map

johanmalm:
...for the following reasons:

1. We interpret 'normal input-focus semantics' for clients with on-demand keyboard interactivity to means that a surface receives input focus on cursor-button-press AND on map (the latter previously missing), just like a normal window would. In this regard, we do not differentiate between layers.

2. Most layer-surfaces set the keyboard interactivity at a similar time to their first (and normally only) map, so the absence of an explicit attempt to focus on map does not make a difference. However, for a long-running layer-shell client (such as lxqt-runner) which sets the interactivity on launch and then maps/unmaps many times throughout its lifetime, a specific focus-attempt is required on map to avoid the client itself having to keep resetting its interactivity to grab the keyboard on map.

3. Compositors like sway and river process focus (for clients with keyboard-interactivity)  in their map-handlers, so this makes for a common approach.

Fixes: #1653

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -337,6 +337,7 @@ handle_map(struct wl_listener *listener, void *data)
 	if (wlr_output) {
 		output_update_usable_area(wlr_output->data);
 	}
+
 	/*
 	 * Since moving to the wlroots scene-graph API, there is no need to
 	 * call wlr_surface_send_enter() from here since that will be done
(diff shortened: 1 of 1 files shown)


ahesford:
Seems OK to me

Consolatis:
:)

Issue #1657, opened 2024-03-24
Strange doubleclick on title behavior


winteriscariot:
So this is a bit of a weird one, but I noticed this today.

I currently have the following mousebind set:

    <context name="Title">
      <mousebind button="Left" action="DoubleClick">
        <action name="Iconify"/>
      </mousebind>
      [...]
ie, doubleclick titlebar to minimize

however, I noticed that labwc seems to register any click within the application itself as a titlebar click, so long as the second click ends on the titlebar. so for instance, if I have featherpad open, I can click in the empty space next to the toolbar icons, then very quickly move my mouse to the titlebar and click a second time. if done fast enough (within the double click time window) this causes the window to iconify.

it appears that labwc is reading that first click, the one within the application content, as a click on the titlebar. double clicking within the application context (so the second click is *not* on the titlebar) behaves as expected.

latest labwc git (11b6836) on arch

Consolatis:
Yes, I think we only use the cursor position of the second click to determine the target.
Is this something that results in unwanted behavior? E.g. is this an issue outside of manually reproducing it?

If yes we could potentially store the first target and on the 2nd click compare the two targets and if they differ do not call the bound action.

winteriscariot:
The only reason I noticed this behavior is because it was unexpected, and a window disappeared when I didn't expect it to. This is also not the way it behaves in openbox, fwiw, so maybe worth restricting it to the titlebar imo.

winteriscariot:
iirc, the first time i noticed, I had clicked something in the window, then went to immediately drag it, but it disappeared instead.

so in my *particular* case, could maybe not register that second click if it's a click-and-hold?

Consolatis:
> iirc, the first time i noticed, I had clicked something in the window, then immediately went to drag it, but it disappeared instead.

That sounds like something that totally can happen. Lets fix this.

tokyo4j:
I experienced this problem too when I first started using labwc. Setting `<mouse><doubleClickTime>` (default is `500`) to `200` mitigated this, but more sophisticated solution would be ideal.

johanmalm:
Does #1659 solve this?

Consolatis:
Should be fixed in latest master. Thanks for the report.

PR #1659, opened 2024-03-25
cursor: validate double-click against SSD part type

johanmalm:
...because click on different parts of a client should not be interpreted as a double click.

Previously only cursor-button and view were validated to be the same between clicks. This resulted in, for example a click on the client surface itself quickly followed by a click on the SSD titlebar being interpreted as a double-click on the titlebar.

Fixes: #1657

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -855,8 +855,10 @@ handle_release_mousebinding(struct server *server,
 }
 
 static bool
-is_double_click(long double_click_speed, uint32_t button, struct view *view)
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, but lets wait for test results.

johanmalm:
@winteriscariot @tokyo4j  Gentle nudge re testing if you've got time.

tokyo4j:
It works fine and prevents unintentional double click on titlebar! Thank you!

PR #1660, opened 2024-03-25
Window switcher, allow to use percent of monitor width.

droc12345:
OSD will size per monitor.  With this the width stays relative sized on my 4k as well as my 1080p monitor

example from my theme
osd.window-switcher.width.percent: 60%

Review and let me know what you think.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -165,6 +165,9 @@ elements are not listed here, but are supported.
 *osd.window-switcher.width*
 	Width of window switcher in pixels. Default is 600.
 
+*osd.window-switcher.width.percent*
+	Width of window switcher in percent. Default is 0. Max 100%
+
 *osd.window-switcher.padding*
 	Padding of window switcher in pixels. This is the space between the
 	window-switcher border and its items. Default is 4.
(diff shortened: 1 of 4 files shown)


johanmalm:
I think it would be preferable to just use `osd.window-switcher.width` and interpret it as a percentage if the value contains `%`.

droc12345:
Easily done. But will the end users be confused with an either/or?

johanmalm:
> Easily done. But will the end users be confused with an either/or?

Hopefully not. Explicit is better than implicit and all that. I think it's potentially more confusing to have two config options to set the same variable.

johanmalm:
Suggestion....

```
bool should_parse_as_percentage = false;
char *p = strrchr(content, '%');
if (p) {
  *p = '\0';
  should_parse_as_percentage = true;
}
....
```

droc12345:
Ok, I'll rework it, and expand on documentation.

Actually atoi doesn't care if the % is at the end, it ignores it. shouldn't matter whether 600 or 50%
then I should only have to check if there is a % in osd.window-switcher.width from the theme
to decide whether to use percent width in display_osd.

Hopefully I made sense with the above.

PR #1661, opened 2024-03-25
Window switcher, allow to use percent of monitor width.

droc12345:
OSD will size per monitor. With this the width stays relative sized on my 4k as well as my 1080p monitor

example from my theme
osd.window-switcher.width: 60%

Accidentally deleted last PR. ~rolls eyes~ 

Ready for review, I think

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -164,6 +164,8 @@ elements are not listed here, but are supported.
 
 *osd.window-switcher.width*
 	Width of window switcher in pixels. Default is 600.
+	Width can also be percent of the width of the monitor.
+	% is mandatory as last character, max 100%
 
 *osd.window-switcher.padding*
 	Padding of window switcher in pixels. This is the space between the
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -165,7 +165,7 @@ elements are not listed here, but are supported.
 *osd.window-switcher.width*
 	Width of window switcher in pixels. Default is 600.
 	Width can also be percent of the width of the monitor.
-	% is mandatory as last character, max 100%
+	% is mandatory as last character in this case, max 100%
 
 *osd.window-switcher.padding*
 	Padding of window switcher in pixels. This is the space between the
--- a/docs/themerc
+++ b/docs/themerc
@@ -60,7 +60,10 @@ osd.border.color: #000000
 osd.border.width: 1
 osd.label.text.color: #000000
 
+# width can be set as percent (of screen width)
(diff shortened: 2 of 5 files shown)


docs/labwc-theme.5.scd
```diff
@@ -164,6 +164,8 @@ elements are not listed here, but are supported.
 
 *osd.window-switcher.width*
 	Width of window switcher in pixels. Default is 600.
+	Width can also be percent of the width of the monitor.
+	% is mandatory as last character, max 100%
```
Consolatis, line 168: `% is mandatory as last character in this case, max 100%.`

Consolatis:
Small nitpick above, other than that LGTM. Didn't test.

droc12345:
I actually did some testing, modify theme osd width, reconfigure and then do alt-tab
went back and forth from percent to normal width, no problems.

Consolatis:
> I actually did some testing, modify theme osd width, reconfigure and then do alt-tab went back and forth from percent to normal width, no problems.

Try setting it to some %, then remove / comment out the whole line from your theme and reconfigure. I think that would then use 600% because `osd_width_should_parse_as_percentage` is never reset. So for theme related options, we usually initialize them in `theme_builtin()` and then validate them in `post_processing()`. E.g. something like
```c
if (osd_width_should_parse_as_percentage) {
	theme->osd_window_switcher_width = MIN(theme->osd_window_switcher_width, 100)
}
```

droc12345:
Any time theme->osd_window_switcher_width is parsed the bool is set to false first.

Shouldn't be a problem but I added it to builtin anyway (I thought I had done it earlier, but evidently didn't)
anyway it's there now.

Consolatis:
> Any time theme->osd_window_switcher_width is parsed the bool is set to false first.

Right, but it will not be parsed if the line was removed from the theme / themerc-override file.

droc12345:
I do check for greater than 100% and just clip to 100 if over.

Consolatis:
> I do check for greater than 100% and just clip to 100 if over.

Personally I'd prefer if we use the `post_processing()` function for that but the way you implemented it is also fine.
LGTM when merged with squash-and-merge. Thanks.

Consolatis:
Hopefully last comment of this slightly disorganized review: we should also check that `osd_window_switcher_width > 0`, regardless of using pecent or not. This is a pre-existing bug, unrelated to the PR.

droc12345:
I have somehow gotten my github repo and local repo really out of whack.
So I'm going to delete this PR, make a clean pull github and local, apply changes and make a new PR.
With all we've talked about above.

Thanks
Don

PR #1664, opened 2024-03-26
docs: Refer to calibration matrix

jp7677:
Let's link tablet rotation and calibration matrix since the calibration matrix can also achieve the rotation, but also more.


johanmalm:
Good thinking. Thanks.

Issue #1665, opened 2024-03-27
Float parsing in rc.xml is locale-dependent


cillian64:
In `rcxml.c` we use `atof` to parse floats from the config file in various places.  It turns out `atof` looks at user locale and so may expect floats to be written like `3,14` rather than `3.14`.  This means that config files can be incompatible across locales, which is very unexpected.

Solutions could be:
- Parse floats in a locale-independent way so they always look the same
- Avoid floats in the config file, e.g. using integer percent
- Keep the current locale-specific behaviour but document it and try to make it obvious

Consolatis:
Reference: https://github.com/labwc/labwc/pull/1648#issuecomment-2022370726

Consolatis:
> > I think it's better to write the parsing to avoid floats, so using an integer percent seems like a reasonable option.
>
> If we don't use percent, it's better to respect locale than to force people to write numbers that don't conform to their intuition. 
_Originally posted by @ahesford in https://github.com/labwc/labwc/issues/1648#issuecomment-2022968434_

As somebody whose native locale is also using `3,14` rather than `3.14` for floats I am very aware of the issue in general. My general assumptions is always to use `3.14` in configs and similar. So rather than trying to support the native locale variants of this we IMHO
- should try to ensure that we always parse `3.14` rather than whatever locale is configured (so config definitions are always the same, regardless of the locale)
- potentially remove the need for it in the first place unless we break existing configs

jlindgren90:
Is the only difference the decimal separator (comma vs dot) or are there others?

Can we make the parsing permissive and allow either comma or dot? Then we would avoid breaking existing configs, but they would also be portable across locales.

jlindgren90:
Possible quick-and-dirty approach:
1) determine locale separator character (e.g. sprintf a float and check the resulting string)
2) replace both '.' and ',' with that character
3) call atof() as usual

Not sure if there are cases that wouldn't work for.

Consolatis:
> Is the only difference the decimal separator (comma vs dot) or are there others?

I think that is the only difference, AFAIR sorting order may also be different for other locales but we don't really rely on that in labwc as far as I know.

jlindgren90:
> AFAIR sorting order may also be different for other locales

Ugh yes, sorting order is even different between "C" and "en_US" locales in some cases. As you say, hopefully that doesn't matter for labwc.

droc12345:
If you want consistency, the wrap the calls to atof in setlocale.
[...]

PR #1666, opened 2024-03-28
common: validate and properly parse floats

ahesford:
This is my (probably over-engineered) attempt to properly parse floats with regard to locale. We ensure that there is at most one `,` or `.` and, if so, convert it to the locale-dependent separator before parsing with `strtod`. If there are too many separators, or the numeric conversion fails, or there is trailing garbage after the parsed number, an error is emitted and the function does not write any converted value.

I tested some basic inputs with `<mouse><scrollFactor>` and this seems to work as anticipated: `1.05` and `1,05` are both recognized as valid, `1,000.5` and `0.1.2` are both rejected.

Fixes: #1665.

### DIFF
--- a/src/common/parse-double.c
+++ b/src/common/parse-double.c
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <locale.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wlr/util/log.h>
+#include "common/mem.h"
+#include "common/parse-double.h"
+
+struct dec_separator {
+	int index;
+	bool multiple;
+};
+
+struct converted_double {
+	double value;
+	bool valid;
+};
+
+static struct dec_separator
+find_dec_separator(const char *str)
+{
+	struct dec_separator loc = {
+		.index = -1,
+		.multiple = false
+	};
+
+	for (int i = 0; *str; i++, str++) {
+		switch (*str) {
+		case ',':
+		case '.':
+			if (loc.index >= 0) {
+				loc.multiple = true;
+				return loc;
+			} else {
+				loc.index = i;
+			}
+			break;
+		}
+	}
+
+	return loc;
+}
+
+static struct converted_double
+convert_double(const char *str)
+{
+	struct converted_double result = {
+		.value = 0,
+		.valid = true,
+	};
+
+	char *eptr = NULL;
+
+	errno = 0;
+	result.value = strtod(str, &eptr);
+
+	if (errno) {
+		wlr_log(WLR_ERROR, "value '%s' is out of range", str);
+		result.valid = false;
+	}
+
+	if (*eptr) {
+		wlr_log(WLR_ERROR, "value '%s' contains trailing garbage", str);
+		result.valid = false;
+	}
+
+	return result;
+}
+
+bool
+set_double(const char *str, double *val)
+{
+	assert(str);
+	assert(val);
+
+	struct dec_separator dloc = find_dec_separator(str);
+	if (dloc.multiple) {
+		wlr_log(WLR_ERROR,
+			"value '%s' contains multiple decimal markers", str);
+		return false;
+	}
+
+	char *lstr = NULL;
+
+	if (dloc.index >= 0) {
+		lstr = xstrdup(str);
+		struct lconv *lc = localeconv();
+		lstr[dloc.index] = *lc->decimal_point;
+		str = lstr;
+	}
+
+	struct converted_double conv = convert_double(str);
+	if (conv.valid) {
+		*val = conv.value;
+	}
+
+	free(lstr);
+	return conv.valid;
+}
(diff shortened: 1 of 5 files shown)


src/common/parse-double.c
```diff
@@ -0,0 +1,94 @@
...
+		lstr = xstrdup(str);
+		struct lconv *lc = localeconv();
+		lstr[sep - str] = *lc->decimal_point;
+		str = lstr;
+	}
+
+	struct converted_double conv = convert_double(str);
+	if (conv.valid && val) {
+		*val = conv.value;
+	}
```
Consolatis, line 90: Do we want to actually support detecting invalid values here (e.g. calling with `val == NULL`)?
Otherwise I don't think we need the `&& val` here, IMHO having an `assert(val)` on top of the function would be better.

src/config/rcxml.c
```diff
@@ -537,7 +538,10 @@ fill_libinput_category(char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "leftHanded")) {
 		set_bool_as_int(content, &current_libinput_category->left_handed);
 	} else if (!strcasecmp(nodename, "pointerSpeed")) {
-		current_libinput_category->pointer_speed = atof(content);
+		double v;
+		if (parse_double(content, &v)) {
+			current_libinput_category->pointer_speed = v;
+		}
```
Consolatis, line 544: What is the reason for checking the result here? I assume it is because of double -> float, should we create a small wrapper like `set_float()` instead?

src/common/parse-double.c
```diff
@@ -0,0 +1,94 @@
...
+	if (*eptr) {
+		wlr_log(WLR_ERROR, "double value contains trailing garbage");
+		result.valid = false;
+	}
+
+	return result;
+}
+
+bool
+parse_double(const char *str, double *val)
```
Consolatis, line 63: To keep it consistent with `common/parse-bool.h` I'd suggest `set_double()` here.

01micko:
FWIW I tested this with `it` locale (since this is where @stefonarch exposed the bug! So I _know_ it uses `comma` for decimal separator) again wirh `<scrollFactor>`. Works. :)

src/common/parse-double.c
```diff
@@ -0,0 +1,94 @@
...
+
+	if (errno) {
+		wlr_log(WLR_ERROR, "double value is out of range");
+		result.valid = false;
+	}
+
+	if (*eptr) {
+		wlr_log(WLR_ERROR, "double value contains trailing garbage");
+		result.valid = false;
+	}
```
Consolatis, line 57: Suggest also having `str` in the log message, otherwise a user is unable to figure out where the error comes from.

droc12345:
>  1,000.5 and 0.1.2 are both rejected.

Not surprised, 1,000.5 would be valid in US and some other places, 
0.1.2 is invalid everywhere as far as locale type data.

if 1,000.5 gets parsed properly, then the last comma or decimal gets changed to decimal_point
but to the left of the decimal point is the thousands_sep 

so strrchr "." for decimal_point, then keep going left but change to thousand separator for all other matches

then 1,000.5 should change properly no matter the locale

https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html for all variations.

droc12345:
while (end of string != beginning of string)
 while (char == number)
  continue
 -> we have found radix, do something with it and bump to next char
 while (char == number)
  continue
 -> we have found thousand sep, do something, and repeat this while() until all string is parsed.

Something like this should handle **all monetary locale data**

Consolatis:
We usually don't use monetary locale data to set scroll speeds though. The restriction to not support thousand seps is fine IMHO.

ahesford:
> Not surprised, 1,000.5 would be valid in US and some other places, 0.1.2 is invalid everywhere as far as _money value using radix and thousand sep. (eta)_
> 
> if 1,000.5 gets parsed properly, then the last comma or decimal gets changed to decimal_point but to the left of the decimal point is the thousands_sep
> 
> so strrchr "." for decimal_point, then keep going left but change to thousand separator for all other matches
> 
> then 1,000.5 should change properly no matter the locale

First, I don't think `strtod` handles the thousands separator at all. Even in a US locale, it truncates at the comma and (in this implementation) triggers the "trailing garbage" error.

Second, even if we could recognize a thousands separator, we shouldn't attempt to support parsing them. The whole notion of this change is that we are indiscriminate about `,` or `.` as a decimal separator. The configuration file does  not indicate the *intended* locale, so we can't distinguish between "1,000.500" as a number slightly larger than one thousand, an intent to write a number larger than one million with one mistyped thousands separator, or a pasting error of a number that was supposed to be equal to or near unity.

This parsing change attempts to be much stricter, but trying to discern thousands separators would undercut most of that effort.

droc12345:
If you only care about the radix, then search for the last non-digit and convert it to locale,
and if more than one non-digit char, then flag string as invalid.

ahesford:
The first commit now includes all suggestions from @Consolatis. The fixup commit refactors the separator search to perform a single pass, and stores an integer index rather than a character pointer to the separator because I absolutely hate doing pointer arithmetic.

Consolatis:
LGTM (including fixup if squashed), not tested.

Issue #1640, opened 2024-03-19
Window switcher switching back to the previous window immediately


charbelnicolas:
One of the last few commits in the last day or two has introduced a bug in the window switcher that makes it switch back to the previous window immediately after selecting a window. It happens most of the time when cycling through the windows switcher. Please see the attached video. This is with the latest git version.

https://github.com/labwc/labwc/assets/14960414/79b9ab90-61c3-41a4-8a51-31065c79e4bb

charbelnicolas:
It's actually quite a showstopper bug not being able to switch windows....

Consolatis:
Thanks for the report. Will try to investigate later.
For now I suggest using an older commit or the last release.

charbelnicolas:
I don't know what triggered it but it's working fine now... It had never happened to me before and I've been using labwc for more than a month now...

Consolatis:
Depending on how you compile labwc, sometimes meson seems to mix things up and doesn't re-compile some parts that it should re-compile. That can then cause various random symptoms like this, recreating the build directory (e.g. via `meson setup new_build; meson compile -C new_build`) should help.

I am closing this issue for now but please comment / reopen if it happens again. We had some major changes to how the OSD shows and switches between windows somewhat recently so it is quite possible that we missed something there.

droc12345:
Looks like it's the original way (not all workspaces).

Which version of labwc does this happen on, or which git commit?

droc12345:
I usually use foot with firefox also, FF is the wayland only version.
And I can't seem to reproduce that problem.

Are you running the latest git version?

and by disappear, you mean it doesn't show the window at all, but it's still active?
ie, if you do a ps does foot still show?

droc12345:
are you using outlines and/or preview (window switcher rc setting)?

tokyo4j:
I don't know if this is the same problem, but by following the procedure below, a window can be hidden, although it seems to be activated from waybar: 

1. Minimize a window
2. Select the window by Alt-Tab
3. Move cursor on the preview
4. Release Alt-Tab

https://github.com/labwc/labwc/assets/22029524/8d3e4eaa-f7f5-41f9-819d-839a17b7518f

tokyo4j:
@charbelnicolas Does applying this patch solve the problem?
```diff
diff --git a/src/input/cursor.c b/src/input/cursor.c
index ad5782e..3115522 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
[...]

PR #1668, opened 2024-03-29
osd: prevent handling cursor actions in preview mode

Consolatis:
Fixes:
- #1640

Other than fixing that bug It also doesn't make sense to allow control of windows from within the window switcher in the first place. Testing appreciated.

### DIFF
--- a/src/foreign.c
+++ b/src/foreign.c
@@ -34,6 +34,12 @@ handle_request_activate(struct wl_listener *listener, void *data)
 {
 	struct view *view = wl_container_of(listener, view, toplevel.activate);
 	// struct wlr_foreign_toplevel_handle_v1_activated_event *event = data;
+
+	if (view->server->osd_state.cycle_view) {
+		wlr_log(WLR_INFO, "Preventing focus request while in window switcher");
+		return;
+	}
+
 	/* In a multi-seat world we would select seat based on event->seat here. */
 	desktop_focus_view(view, /*raise*/ true);
 }
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -529,7 +529,8 @@ process_cursor_motion(struct server *server, uint32_t time)
 		dnd_icons_move(seat, seat->cursor->x, seat->cursor->y);
 	}
 
-	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse) {
+	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse
+			&& !server->osd_state.cycle_view) {
 		desktop_focus_view_or_surface(seat, ctx.view, ctx.surface,
 			rc.raise_on_focus);
 	}
@@ -806,6 +807,10 @@ static bool
 handle_release_mousebinding(struct server *server,
 		struct cursor_context *ctx, uint32_t button)
 {
+	if (server->osd_state.cycle_view) {
+		return false;
+	}
+
 	struct mousebind *mousebind;
 	bool consumed_by_frame_context = false;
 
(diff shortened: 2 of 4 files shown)


tokyo4j:
Tested. I confirmed it fixes the issue well.

However, for future refactoring, I suggest moving checking for `server->osd_state.cycle_view` into `desktop_focus_view()`, as it will catch cases like https://github.com/labwc/labwc/issues/1640#issuecomment-2026680577. It will require some refactoring across the codebase though (for example we need to modify `end_cycling()` in `keyboard.c`).

Consolatis:
> However, as a future refactoring, I suggest moving checking for `server->osd_state.cycle_view` into `desktop_focus_view()`, as it will simplify the code and catch cases like [#1640 (comment)](https://github.com/labwc/labwc/issues/1640#issuecomment-2026680577). It will require some refactoring across the codebase though (for example we need to modify `end_cycling()` in `keyboard.c`).
> 
> EDIT: After some thoughts, I realized my suggestion might not be a better option, as we need to add checking in functions like `view_move_to_front()` and `view_move_to_back()` if we take that path. For [#1640 (comment)](https://github.com/labwc/labwc/issues/1640#issuecomment-2026680577), just checking `server->osd_state.cycle_view` in `handle_request_activate()` will be enough.

Yeah, had the same [idea](https://github.com/labwc/labwc/issues/1640#issuecomment-2025566972) but also then saw the need to modify `end_cycling()` and potentially others. It may be worth it but I don't really feel like working on that currently.

Good find with the foreign-toplevel one, I assume something similar may also happen for xdg-activation. I'll add checks in both places for now.

tokyo4j:
I was missing that `handle_request_minimize()` eventually calls `desktop_focus_view()` when a minimized window is un-minimized from waybar, so I guess the same check as done in `handle_request_activate()` should be applied there.

Consolatis:
> I was missing that `handle_request_minimize()` eventually calls `desktop_focus_view()` when a minimized window is un-minimized from waybar, so I guess the same check as done in `handle_request_activate()` should be applied there.

Good point. I wonder if we should try to figure out a more generic solution here though like the proposed check in `desktop_focus_view()` and make that work properly even when in OSD preview mode.

I have a slight feeling that what we are doing in this PR are accumulating band aids (although the cursor specific ones should likely still be implemented because it just makes sense that windows should not be controlled while in OSD preview mode).

tokyo4j:
> I have a slight feeling that what we are doing in this PR is accumulating band aids (although the cursor specific ones should likely still be implemented because it just makes sense that windows should not be controlled while in OSD preview mode).kk

Yeah, I'm feeling the same. So another approach will be:

- Reverse the order of `desktop_focus_view()` and `osd_finish()` in `end_cycling()`
- Add `server->osd_state.cycle_view` check in `view_minimize()`, `view_move_to_back()`, `view_move_to_front()`, and `desktop_focus_view()` (not sure these are enough)

In my testing, the check in `desktop_focus_view()` was actually not necessary to prevent the inconsistent state because it calls `view_minimize()` and returns for minimized windows, but it was necessary for me because the Alt key-press event (preceding Tab key-press) is forwarded to Fcitx5 then it sends Alt key-release event via virtual keyboard on focus changes between un-minimized windows, which unexpectedly terminates OSD.

johanmalm:
LGTM. Have tested.

Thanks for all the work here.

I think this (protecting with some if-statements) is a good solution - particularly at this stage of the cycle.

I've also had a think/play with refactoring functions associated with raise/focus/activate/minimize/etc. but nothing obvious springs to mind and it's a finicky area to get right.

PR #1670, opened 2024-03-30
OSD window switcher, based on work done by @Consolatis, it uses a custom

droc12345:
field (printf style formatting) to replace the original field formats. I added documentation, and fixed the few style check complaints.

Example:
 ```
 <windowSwitcher preview="no" outlines="no" allWorkspaces="yes">
    <fields>
      <field content="custom" format="foobar %b %3s %-10o %-20W %-10i %t" width="100%" />
    </fields>
  </windowSwitcher>

```

Mono space font recommended, along with possibly adjusting width of OSD window, 
can be used in conjunction wiith #1662

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -243,6 +243,30 @@ this is for compatibility with Openbox.
 
 		- *output* Show output id, if more than one output detected
 
+		- *custom* A printf style config that can replace all the above
+		  fields are:
+			- 'B' - shell type, values [xwayland|xdg-shell]
+			- 'b' - shell type (short form), values [X|W]
+			- 'S' - state of window, values [M|m|F] (3 spaces allocated)
+			        (maximized, minimized, fullscreen)
+			- 's' - state of window (short form), values [M|m|F] (1 space)
+			- 'I' - wm-class/app-id
+			- 'i' - wm-class/app-id trimmed, remove "org." if available
+			- 'W' - workspace name
+			- 'w' - workspace name (if more than 1 ws configured)
+			- 'O' - output name
+			- 'o' - output name (show if more than 1 monitor active)
+			- 'T' - title of window
+			- 't' - title of window (if different than wm-class/app-id)
+		  Recommend using with a mono space font, to keep alignment.
+		- *custom - subset of printf options allowed -- man 3 printf*
+			- random text may be inserted
+			- field length, example "%10" use 10 spaces, even if text uses less
+			- left jusify text, example "%-"
+			- right justify text, example "%" instead of "%-"
+			- example, %-10 would left justify and make room for 10 charaters
+		- Only one custom format allowed now. Future enhancements may allow more than one.
+
 	*width* defines the width of the field expressed as a percentage of
 	the overall window switcher width. The "%" character is required.
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -77,7 +77,22 @@
         <field content="output" width="9%" />
         <field content="identifier" width="30%" />
         <field content="title" width="50%" />
-       </fields>
+      </fields>
+    </windowSwitcher>
+
+    custom format - (introduced in 0.7.2)
+    It allows one to replace all the above "fields" with one line, using a
+    printf style format. For field explanations, "man 5 labwc-config".
+
+    The example below would print "foobar",then type of window (wayland, X),
+    then state of window (M/m/F), then output (shows if more than 1 active),
+    then workspace name, then identifier/app-id, then the window title.
+    Uses 100% of OSD window width.
+
+    <windowSwitcher show="yes" preview="no" outlines="no" allWorkspaces="yes">
+      <fields>
+        <field content="custom" format="foobar %b %3s %-10o %-20W %-10i %t" width="100%" />
+      </fields>
     </windowSwitcher>
   -->
 
--- a/src/osd_field.c
+++ b/src/osd_field.c
@@ -1,0 +1,365 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <ctype.h>
+#include <wlr/util/log.h>
+#include "common/mem.h"
+#include "config/rcxml.h"
+#include "view.h"
+#include "workspaces.h"
+#include "labwc.h"
+#include "osd.h"
+
+/* includes '%', terminating 's' and NULL byte, 8 is enough for %-9999s */
+#define LAB_FIELD_SINGLE_FMT_MAX_LEN 8
+static_assert(LAB_FIELD_SINGLE_FMT_MAX_LEN <= 255, "fmt_position is a unsigned char");
+
+/* forward declares */
+typedef void field_conversion_type(struct buf *buf, struct view *view, const char *format);
+struct field_converter {
+	const char fmt_char;
+	field_conversion_type *fn;
+};
+
+static const struct field_converter field_converter[];
+
+/* Internal helpers */
+
+static const char *
+get_app_id_or_class(struct view *view, bool trim)
+{
+	/*
+	 * XWayland clients return WM_CLASS for 'app_id' so we don't need a
+	 * special case for that here.
+	 */
+	const char *identifier = view_get_string_prop(view, "app_id");
+
+	/* remove the first two nodes of 'org.' strings */
+	if (trim && identifier && !strncmp(identifier, "org.", 4)) {
+		char *p = (char *)identifier + 4;
+		p = strchr(p, '.');
+		if (p) {
+			return ++p;
+		}
+	}
+	return identifier;
+}
+
+static const char *
+get_type(struct view *view, bool short_form)
+{
+	switch (view->type) {
+	case LAB_XDG_SHELL_VIEW:
+		return short_form ? "[W]" : "[xdg-shell]";
+#if HAVE_XWAYLAND
+	case LAB_XWAYLAND_VIEW:
+		return short_form ? "[X]" : "[xwayland]";
+#endif
+	}
+	return "???";
+}
+
+static const char *
+get_title(struct view *view)
+{
+	return view_get_string_prop(view, "title");
+}
+
+static const char *
+get_title_if_different(struct view *view)
+{
+	const char *identifier = get_app_id_or_class(view, /*trim*/ false);
+	const char *title = get_title(view);
+	if (!identifier) {
+		return title;
+	}
+	return (!title || !strcmp(identifier, title)) ? NULL : title;
+}
+
+/* Field handlers */
+static void
+field_set_type(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type conversion-specifier: B (backend) */
+	buf_add(buf, get_type(view, /*short_form*/ false));
+}
+
+static void
+field_set_type_short(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type conversion-specifier: b (backend) */
+	buf_add(buf, get_type(view, /*short_form*/ true));
+}
+
+static void
+field_set_workspace(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type conversion-specifier: W */
+	buf_add(buf, view->workspace->name);
+}
+
+static void
+field_set_workspace_short(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type conversion-specifier: w */
+	if (wl_list_length(&rc.workspace_config.workspaces) > 1) {
+		buf_add(buf, view->workspace->name);
+	}
+}
+
+static void
+field_set_win_state(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type conversion-specifier: s */
+	if (view->maximized) {
+		buf_add(buf, "M");
+	} else if (view->minimized) {
+		buf_add(buf, "m");
+	} else if (view->fullscreen) {
+		buf_add(buf, "F");
+	} else {
+		buf_add(buf, " ");
+	}
+}
+
+static void
+field_set_win_state_all(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type conversion-specifier: S */
+	buf_add(buf, view->minimized ? "m" : " ");
+	buf_add(buf, view->maximized ? "M" : " ");
+	buf_add(buf, view->fullscreen ? "F" : " ");
+	/* TODO: add always-on-top and omnipresent ? */
+}
+
+static void
+field_set_output(struct buf *buf, struct view *view, const char *format)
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

**Comment by droc12345** [comment]
This is ready for review. I've been using this for a while now.

**Comment by johanmalm** [comment]
> This is ready for review. I've been using this for a while now.

Thanks. Will prioritize review with view to merge. Got side-tracked on drop-shadows this evening.

File: docs/rc.xml.all [hunk 1]
```diff
@@ -77,7 +77,20 @@
         <field content="output" width="9%" />
         <field content="identifier" width="30%" />
         <field content="title" width="50%" />
-       </fields>  
+      </fields>  
+    </windowSwitcher>
+
+    A new OSD format - custom
+    It effectively replaces all the above "fields" with one line, using a
+    printf style format. The example below would print "foobar",then type of
```
johanmalm, line 85: Presumably this (`all the above "fields" with one line`) is one way of using it.
Other uses-cases could be to have multiple `custom` fields for more flexibility.
That description ought to reflect this.

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+#include <assert.h>
+#include <wlr/util/log.h>
+#include "common/mem.h"
+#include "config/rcxml.h"
+#include "view.h"
+#include "workspaces.h"
+#include "labwc.h"
+#include "osd.h"
+
+/* includes '%', terminating 's' and NULL byte, 8 is enough for %-9999W */
```
johanmalm, line 11: Because we replace the `W` with `s`?

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+#define LAB_FIELD_SINGLE_FMT_MAX_LEN 8
+static_assert(LAB_FIELD_SINGLE_FMT_MAX_LEN <= 255, "fmt_len is a unsigned char");
+
+/* forward declares */
+typedef void field_set_fn_t(struct buf *buf, struct view *view, const char *format);
```
johanmalm, line 16: We haven't used `typedef` for functions before.

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#typedefs
https://git.sr.ht/~sircmpwn/cstyle#features-to-avoid

Might a simple `void (fn *)(struct buf *buf, struct view *view, const char *format)` suffice in the struct below.

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+struct field_handlers {
+	const char fmt_char;
```
johanmalm, line 18: Suggest sticking with `printf` language, so better to call this `conversion_specifier`.

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+	field_set_fn_t *fn;
```
johanmalm, line 19: Could this method dispatch be called `convert`?

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+{
+	switch (view->type) {
+	case LAB_XDG_SHELL_VIEW:
+		return short_form ? "[W]" : "[xdg-shell]";
+#if HAVE_XWAYLAND
+	case LAB_XWAYLAND_VIEW:
+		return short_form ? "[X]" : "[xwayland]";
+#endif
+	}
+	return "???";
```
johanmalm, line 57: Why the three question marks?
Should we assert here as we should never reach this point.

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+field_set_type_short(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type identifier: b (backend) */
+	buf_add(buf, get_type(view, /*short_form*/ true));
+}
+
+static void
+field_set_workspace(struct buf *buf, struct view *view, const char *format)
+{
+	/* custom type identifier: W */
```
johanmalm, line 95: `s/identifier/conversion-specifier/`???

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+static void
+field_set_custom(struct buf *buf, struct view *view, const char *format)
+{
+	if (!format) {
+		wlr_log(WLR_ERROR, "Missing format for custom window switcher field");
+		return;
+	}
+
+	char fmt[LAB_FIELD_SINGLE_FMT_MAX_LEN];
+	unsigned char fmt_len = 0;
```
johanmalm, line 189: Possibly a moot point, but `pos` is more obvious to me here compared with `fmt_len`

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+
+	struct buf field_result;
+	buf_init(&field_result);
+
+	char field_formatted[4096];
```
johanmalm, line 194: There are four buffers here (`fmt`, `field_formatted`, `field_result` and `format`) which is a bit hard for the brain to keep up with. Consider refactoring.

Possibly just reducing the scope of `field_formatted` might help (it's only used within a loop at the end. Could call it `converted` or `expanded` to avoid using the word `format` to mean so many things.

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+			}
+			continue;
+		}
+
+		/* Allow string formatting */
+		/* TODO: add . for manual truncating? */
+		if (*p == '-' || *p == '#' || (*p >= '0' && *p <= '9')) {
+			if (fmt_len >= LAB_FIELD_SINGLE_FMT_MAX_LEN - 2) {
+				/* Leave space for terminating 's' and NULL byte */
+				wlr_log(WLR_ERROR,
```
johanmalm, line 215: Should we continue if we exceed max-len? Might be cleaner to take a more defensive approach and just goto error rather than trying to interpret something that's "undefined".

src/osd_field.c
```diff
@@ -0,0 +1,359 @@
...
+
+reset_format:
+		/* Reset format string and tmp field result buffer */
+		buf_clear(&field_result);
+		fmt_len = 0;
+	}
+	free(field_result.buf);
+}
+
+static const struct field_handlers field_handlers[LAB_FIELD_COUNT] = {
```
johanmalm, line 255: I'm okay with this, but my preferred style is 
```
static struct foo foo[] { {
  .content = WHATEVER,
  .conversion_specifier = 'B',
  .convert = some_function_pointer
}, {
...etc..
};
```

...and then use `ARRAY_SIZE()`.

johanmalm (commented):
It's a nice PR.

Although it turned out to be a large number of review comments, most of them are trivial.

Moving the xml parsing to osd_field.c is nice. Same for `osd_field_validate()`

I like the approach of throwing the `%*s` at `snprintf()` to make use of width specifiers as such things.

It would be easier to review if this was split into two commits - the first one just refactoring in preparation (like the creation of `osd.h` and moving some stuff from rcxml.h to osd_field.h). Anyway, we can live with it :smile: It's probably more hassle than it's worth to split it out now.

The biggest consideration for me is if we should try to find a way to avoid hard coding strings like `min`, `max`, `XWayland`, `W`, etc. It might not be that straight forward and we run the risk of inventing our own markup language in the process.
I know I'm pointing out the obvious (and we of course already have hard-coded strings in the OSD generator - for which I am to blame, but two wrongs don't make a right):

1. They are hard to change later without causing regression
2. If we hardcode strings that are part of the `gui` we should really translate them
3. It makes for an inflexible design that's hard to configure or extend. The hard-coded values might suit a small number of people today, but will they stand the test of time?

docs/labwc-config.5.scd
```diff
@@ -243,6 +243,22 @@ this is for compatibility with Openbox.
 
 		- *output* Show output id, if more than one output detected
 
+		- *custom* A printf style config that can replace all the above
```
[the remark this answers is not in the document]
Consolatis: Right, currently we only support the length modifier and alignment (e.g. `-` or `#`). I thought about potentially adding a max-field-width modifier (like `.`) which we would then apply manually to the result of `snprintf()`. AFAIR truncating isn't supported via printf.

Now while looking through the man page it seems `#` isn't actually part of the C spec. Not sure why I had in mind that that one align to the right. Based on the man page it seems the string is always aligned to the right unless `-` is used so we should likely remove that one (after testing that `%20W` is indeed aligned to the right and `%-20W` aligned to the left).

docs/rc.xml.all
```diff
@@ -77,7 +77,20 @@
         <field content="output" width="9%" />
         <field content="identifier" width="30%" />
         <field content="title" width="50%" />
-       </fields>  
+      </fields>  
+    </windowSwitcher>
+
+    A new OSD format - custom
```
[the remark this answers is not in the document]
Consolatis: Good point.

src/osd_field.c
Consolatis: right, the comment should likely state `%-9999s` instead as the `W` never ends up in the fmt buffer.

src/osd_field.c
Consolatis: https://git.sr.ht/~sircmpwn/cstyle#variable-declarations states that typedefs `MUST` be used for function pointers that are arguments for other functions and also states later
> Typedefs are usually a bad idea, but function pointers are so verbose and weird and difficult to read that typedefs help out here.

src/osd_field.c
Consolatis: What about `fmt_pos`? I don't really like the `pos` naming usually if there may be multiple "positions", you never really know what position this refers to otherwise.

src/osd_field.c
Consolatis: This array is accessed directly via the enum as index as well (via `osd_field_get_content()` from `osd.c`).

src/osd_field.c
Consolatis: 3 because this function always returns three chars.
Agree on some `assert(false)`, we still need some return after the assert though in case labwc is compiled without asserts.

docs/labwc-config.5.scd
```diff
@@ -243,6 +243,22 @@ this is for compatibility with Openbox.
...
+		  fields are:
+		  'B' - xwayland/xdg-shell window
+		  'b' - X/W (short form)
+		  'S' - M/m/F (max, min, full) (3 chars)
```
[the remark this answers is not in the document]
Consolatis: This long form contains all of `MmF` or a space if the condition is not true. E.g. `M F` would mean maximized and fullscreen and `m  ` would mean minimized while `   ` would mean non of those states.

docs/labwc-config.5.scd
```diff
@@ -243,6 +243,22 @@ this is for compatibility with Openbox.
...
+		  's' - M/m/F (single char used)
```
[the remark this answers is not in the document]
Consolatis: Its an interesting idea. However, I don't really see how a configuration option for those could look like. Including it in the user supplied format string sounds \*very\* messy.

Consolatis:
@droc12345 Feel free to update the PR with all things you are comfortable updating, I can then do the rest but it may take a while.

droc12345:
Made many of the changes, but not the ones in osd_field, dealing with max length, and array size.
I'll let y'all decide what to do there.

docs/labwc-config.5.scd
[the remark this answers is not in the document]
droc12345: What we could do is refer the user to their man page, typically "man 3 printf" and make it clear
that we only support a subset.

docs/rc.xml.all
droc12345: Done

src/osd_field.c
droc12345: field_set_fn_t becomes field_conversion_type

field_handlers becomes field_converter

src/osd_field.c
droc12345: Putting out an error and skipping unknowns might be a better option.

droc12345: I didn't remove the *p == '#', doesn't really do anything as that should never show as a format option.

droc12345: I actually did wind up removing the check for '#' as that denotes an alternate format,
and we don't use that, I also changed the >=0 and <=9 to isdigit()

Todo: check that 1st digit after '%' is not zero, as that says zero fill the output
if '-' is before the '0' then the zero is ignored. But I wasn't sure how much effort 
to put into such a thing.

droc12345:
This should be ready for a review,

I didn't change the ??? from get_type, as I don't think it hurts anything to be there.

I did change 
if (*p == '-' || *p == '#' || (*p >= '0' && *p <= '9')) {
to
if (*p == '-' || isdigit(*p)) {
in field_set_custom().
Reason: the "*" is not needed nor used, it is for alternate forms for number and
I swapped out the check on 0-9, using isdigit instead. Even though some of the is*() functions do
look at locale, and it does affect things like radix, thousand sep, etc, but all locales look at the same
10 digits, ie 0-9.

src/osd_field.c
johanmalm: Yes, good point. Happy with approach.

src/osd_field.c
johanmalm: You’re right. I stand corrected. Happy with approach.

docs/labwc-config.5.scd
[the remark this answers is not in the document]
johanmalm: It’d be good to be explicit in what we support. Agree with sign-posting to man page, but feel we ought to describe at some level what’s in vs out.

droc12345:
I think I'm done, unless a faux pas is found. 
I enhanced the documentation in labwc-config with explanations of what we use from printf style formatting
with examples, and did the same thing to rc.xml.all

docs/rc.xml.all
```diff
@@ -82,19 +82,31 @@
...
+
+    Options:
+     random text may be inserted
+     field length, example "%10" use 10 spaces, even if text uses less
+     left jusify text, example "%-"
+     right justify text, example "%" instead of "%-"
+
+    example, %-10 would left justify and make room for 10 charaters
+
+    Future enhancements may allow more than one custom format.
```
Consolatis, line 109: Suggest replacing this with a reference to the labwc-config manpage.

droc12345: Ok, remove from rc.xml.all and clarified in labwc-config

johanmalm:
LGTM

Branch has conflicts with master. Would you be happy to resolve and push?

droc12345:
This branch has no conflicts with the base branch

this is what I see,

droc12345:
I'm going to try and clean up my local osd-custom and overwrite the one from this PR.
Hopefully that makes things merge right.

droc12345:
This should be good to go, collapsed all commits into one and not showing any conflicts

johanmalm:
Thanks

PR #1674, opened 2024-04-02
snap: cache and ignore last-snapped edge when growing or shrinking

ahesford:
When growing or shrinking a view by snapping to an edge, a client may ignore the requested size and instead keep its original size or substitute a different (possibly constrained) size. In this case, the view may not actually contact the snapped edge, and a subsequent snap attempt will just keep re-trying (and failing) to contact the same ege.

To mitigate this, remember the last-snapped view, snapping direction and offset of the snapping edge in snap.c; when re-attempting a snap for the same view in the same direction, ignore the edge that was last "hit", to allow snapping to progress beyond the problematic edge.

The implementaion considers a single, static cache in an attempt to minimize intrusion into other parts of the program. This means there are some... *edge*... cases where a snap tries to hit an edge, client constraints make it fall short, then the user selects a different window and attempts a snap. Switching windows will invalidate the edge cache, which means that selecting the original window and attempting a snap will re-consider the problematic edge. Thus, you get past a "sticky" edge by trying *twice* to grow or shrink. Avoiding the cache thrashing would require per-view state caching, which would be complex and ugly. 

This needs some thorough testing, and I'm open to suggestions for improvement. Some quick tests indicate that this generally works as expected and is minimally invasive. I don't believe we need to handle `snap_move_to_edge` because the client can't override our move operation; I also think handling cache invalidating when we switch from growing to shrinking is superfluous; the issue should be limited to cases where we try, but fail to reach, a given edge; switching between grow/shrink should prevent the problem edge from being considered in the first place.

This explicitly invalidates the snap cache in `view_destroy` in case the view pointer gets reused for another view in the future. Furthermore, the cache is invalidated at the start of interactive operations, because any manual move or resize means we've potentially changed the window sizing dynamics and will need to re-evaluate every candidate edge in a subsequent snap operation. I'm not thrilled about leaking the cache in these two places, but don't see a cleaner option.

### DIFF
--- a/src/snap.c
+++ b/src/snap.c
@@ -4,10 +4,9 @@
 #include <wlr/util/box.h>
 #include "common/border.h"
 #include "common/macros.h"
-#include "config/rcxml.h"
 #include "edges.h"
 #include "labwc.h"
-#include "resistance.h"
+#include "snap-constraints.h"
 #include "snap.h"
 #include "view.h"
 
@@ -121,9 +120,8 @@ snap_move_to_edge(struct view *view, enum view_edge direction,
 		struct border next_edges;
 		edges_initialize(&next_edges);
 
-		edges_find_neighbors(&next_edges,
-			view, target, output, check_edge,
-			/* use_pending */ true, /* ignore_hidden */ false);
+		edges_find_neighbors(&next_edges, view, view->pending, target,
+			output, check_edge, /* ignore_hidden */ false);
 
 		/* If any "best" edges were encountered, limit motion */
 		edges_adjust_move_coords(view, next_edges,
@@ -135,8 +133,8 @@ snap_move_to_edge(struct view *view, enum view_edge direction,
 }
 
 void
-snap_grow_to_next_edge(struct view *view, enum view_edge direction,
-		struct wlr_box *geo)
+snap_grow_to_next_edge(struct view *view,
+		enum view_edge direction, struct wlr_box *geo)
 {
 	assert(view);
 	assert(!view->shaded);
@@ -196,17 +194,26 @@ snap_grow_to_next_edge(struct view *view, enum view_edge direction,
 	struct border next_edges;
 	edges_initialize(&next_edges);
 
+	/* Use a constrained, effective geometry for snapping if appropriate */
+	struct wlr_box origin = snap_constraints_effective(view, direction);
+
 	/* Limit motion to any intervening edge of other views on this output */
-	edges_find_neighbors(&next_edges,
-		view, *geo, output, check_edge,
-		/* use_pending */ true, /* ignore_hidden */ false);
+	edges_find_neighbors(&next_edges, view, origin, *geo,
+		output, check_edge, /* ignore_hidden */ false);
+
 	edges_adjust_resize_geom(view, next_edges,
 		resize_edges, geo, /* use_pending */ true);
+
+	/*
+	 * Record effective geometry after snapping in case the client opts to
+	 * ignore or modify the configured geometry
+	 */
+	snap_constraints_set(view, direction, *geo);
 }
 
 void
-snap_shrink_to_next_edge(struct view *view, enum view_edge direction,
-		struct wlr_box *geo)
+snap_shrink_to_next_edge(struct view *view,
+		enum view_edge direction, struct wlr_box *geo)
 {
 	assert(view);
 	assert(!view->shaded);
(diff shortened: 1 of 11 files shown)


src/snap.c
```diff
@@ -196,10 +279,21 @@ snap_grow_to_next_edge(struct view *view, enum view_edge direction,
...
 	/* Limit motion to any intervening edge of other views on this output */
 	edges_find_neighbors(&next_edges,
 		view, *geo, output, check_edge,
 		/* use_pending */ true, /* ignore_hidden */ false);
+
+	/*
+	 * Cache the snapped edge so that two successive snapping attempts
+	 * don't get stuck continually trying to hit the same edge if the
+	 * client fails to honor a geometry change.
+	 */
```
Consolatis, line 294: Maybe something like `[..] if the client doesn't want to honor a geometry change due to client side size constrains.` ?

src/snap.c
```diff
@@ -11,6 +11,85 @@
 #include "snap.h"
 #include "view.h"
 
+static struct {
```
johanmalm, line 14: Suggest adding a comment here to describe why we have the cache. The text in the commit message would be good.

src/snap.c
```diff
@@ -196,10 +279,21 @@ snap_grow_to_next_edge(struct view *view, enum view_edge direction,
 	struct border next_edges;
 	edges_initialize(&next_edges);
 
+	/* Invalidate last-snapped-edge cache if view or direction changes */
+	validate_snap_cache(view, direction);
```
johanmalm, line 283: Consider refactoring this to make it easier to read... as follows:

```
if (!cached_is_valid(view, direction)) {
  reset cache
}

File: src/snap.c [hunk 4]
```diff
@@ -11,6 +11,85 @@
...
+		offset = edges.top;
+		break;
+	case VIEW_EDGE_DOWN:
+		offset = edges.bottom;
+		break;
+	default:
+		return;
+	}
+
+	if (view && BOUNDED_INT(offset)) {
```
johanmalm, line 58: Could add a `if (!view) {return}` at top of function and remove it from here.

johanmalm (commented):
Should we consider putting the cache code in a dedicated `snap-cache.c` for increased separation boundaries?

I always think it seems appealing to keep a translation unit really small when it contains a static variable / state like that.

It would probably require an additional `snap_cache_set(view, direction, offset)` or something.

ahesford:
Still playing around with this. In the latest commit, I remember the desired geometry after the snapping operation (which will become `view->pending` after the snapping functions return, and will be sent via an XDG configure. Then, in the XDG commit handler, where we update `view->pending` to match `view->current` in recognition that the client may have altered the desired size, I call a cache update function to replace the desired geometry with the actual geometry set by the client---*if the view in question is still the cached view*. When deciding whether to skip the last snapped edge in a subsequent snap operation, we only do so if the view's pending geometry at the time we're snapping matches whatever we cached as the desired (or client-overridden) geometry during the last snap. Effectively, if the window has not changed from what we would expect from the last snap, we don't bother trying the same edge again.

I think this gets us closer to the "right" behavior, and it might be even more correct if we move from a single, global cache to a per-view cache if we decide to do so.

I have not yet incorporated any suggested changes.

Consolatis:
> Still playing around with this. In the latest commit, I remember the desired geometry after the snapping operation (which will become `view->pending` after the snapping functions return, and will be sent via an XDG configure. Then, in the XDG commit handler, where we update `view->pending` to match `view->current` in recognition that the client may have altered the desired size, I call a cache update function to replace the desired geometry with the actual geometry set by the client---_if the view in question is still the cached view_. 

I think this could also be done in `view_moved()` which AFAIR is called by both, xwayland and xdg commit handlers.

ahesford:
I took another pass, and tried to incorporate previous suggestions into the rewrite:
1. Logic has now moved into a separate `snap-constraints.c` to limit the scope of the cache.
2. Comments have been rewritten for clarity.
3. The entire approach has been rethought.

First, I can't update the "expected" geometry with every commit in `xdg.c`, because then every subsequent resize (whether interactive, snapped, or client-initiated) will replace the expected geometry in the cache and cause the last snapped edge to be ignored even when it should be considered (e.g., if you manually resize the view away from the edge, the client's constrains apply differently and may "catch" a subsequent snapped resize differently). To get around this, I keep a `pending` flag in the cache, which is set when the cache is first set by any snapped resize operation. When the XDG commit handler wants to update the pending geometry to reflect whatever the actual final client geometry is, this flag is cleared and the expected snapping geometry is updated at most once. (The *best* way to do this would be to track the configure serial in the cache and only allow an update if the commit serial matches what we intended to set the first time. However, we don't have this information at the time the snapping logic is applied, and trying to expose that information would greatly increase the complexity. I think the current cache-update logic gets close enough.)

Second, ignoring the "last" snapped edge isn't actually good enough. If there are two nearby edges and client constraints would prevent the view from hitting *either* (e.g., if two edges fall within a single character cell of a terminal that forces cell-based resizing), then the "last" check would just thrash between both unreachable edges, and the view would never move. To get around this, I instead capture the position of the moving edge *as was requested* and, when a subsequent snapping attempt is made, use that position to alter the effective geometry of the view. Thus, we substitute the size we *last wanted* for the size *we actually have*, which means that we can pretend for the purposes of snapping that the client always honors our requests exactly.

ahesford:
After running this for a few days, I think this generally does what I expect.

Consolatis:
To be honest I can't really wrap my head around the changes. But I also didn't spot anything obviously wrong. You have by far the most knowledge of the edge snapping implementation and behavior so feel free to rebase and merge if you are satisfied with the PR.

johanmalm:
I think the code looks fine.

I have found a bug though. Can help debug, but have just recorded it for now.

[recording.webm](https://github.com/labwc/labwc/assets/6955353/53bc50e8-dd00-4aee-a203-c4668ba0babc)

ahesford:
I'm able to reproduce the `xfce4-terminal` issue. It does not happen with `foot`.

The issue seems to be that, once the window is shrunk back to just above the "sticky" edge, each subsequent "GrowToEdge" again attempts to increase the view slightly beyond the edge. Labwc sends the client a configure request, but `xfce4-terminal` determines that its size is already correct and never sends a commit in response to the request. (At this point, labwc reports "client (xfce4-terminal) did not respond to configure request in 100 ms".)

The snap cache stores the "intended" (pending) geometry when a snap action is performed. Normally, the commit handler would see that the size has been finalized, overwrite `view->pending` with `view->current` and also replace the intended geometry in the cache with the actual geometry of the view. Because the commit handler never fires here, the snap cache always retains the old pending geometry, so every subsequent check of the snap cache determines that the view geometry has changed and the cache is considered invalid.

This is fixed in the latest version by the addition of `snap_constraints_update(view)` to `handle_configure_timeout` at line 193 of `xdg.c`. When we reset `view->pending` to reflect the actual `view->current`, we also update the snap cache (if appropriate) to reflect the current geometry of the view.

johanmalm:
Thanks. Great stuff.

PR #1678, opened 2024-04-05
feat: add partial tablet v2 protocol support

jp7677:
As a follow up to https://github.com/labwc/labwc/issues/1672 and the ideas mentioned there, I ~played/hacked~ put this together. Running `gtk4-demo` and there the `Paint` demo, I can draw now in Stylus only mode and pen pressure is applied. So far so good.

~Looking at the general UI, just blindly switching from mouse emulation to tabled mode disables a lot of things that are imho not so easy to get correctly implemented, e.g. focus switch between applications, resize applications, drag & drop and surely a few others. That said, having a configuration switch like 'tablet-mode' (tablet-native|mouse-emulation) and/or an action 'toogle-tablet-mode' might be a way to mitigate this and allow a user to consciously switch between modes with all the consequences and restrictions applied.~ Actually this is now in a very usable state, see below. A switch for mouse emulation is also part of this PR.

~There are more tablet axis that aren't covered yet in this PR.~ I also haven’t done much testing with side businesses like disconnecting/reconnecting the tablet, multiple outputs, output mapping.

So, open for feedback (esp. with a real drawing app like Krita) and thoughts.

Status / TODO:

- [x] Tablet tool wiring/signalling 
- [x] Axis/proximity handling
- [x] Hovering between tablet-capable / non-tablet-capable surfaces / desktop
- [x]  Focus switching
- [x] Idle inhibit
- [x] Client cursors (set_cursor), tricky part...
- [x] Move/resize CSD applications
- [x] Out of surface scrolling
- [x] Button mapping (for non-tip)
- [x] Force-cursor-emulation configuration option
- [x] Add a few safety checks like tool type
- [ ] Drag n drop (if this turns out to be doable)
- [x] Document force-mouse-emulation and button mapping implications
- [x] Testing with device reconnects 
- [x] Testing with output mapping
- [ ] General testing
- [ ] Review, polish and reorganize commits
 
Currently known issues:
- Notify cursor from pointer and tablet can still interfere, noted on out-of-surface scrolling in GTK4 applications. The resize cursor shines through when scrolling out of surface.
- ~The scrollbar in VS-Code, when running as native Wayland application, only when being adjacent or above another window, jumps to the top when scrolling out of surface. This does not happen with VS-Code in XWayland or when the out of scrolling happens on the desktop.~ A similar issue can be seen with the volume slider in Shortwave/GTK4 (Slider on a popup, thus also scrolling out of surface above another surface).

Ideas:
- ~`wlr_tablet_v2_tablet_tool` has a `focused_surface` field which is updated on proximity-in/out.  May be we can use that one instead of our own `drawing_tablet_tool.focus_surface`.~ (Done)
-  Notify cursor interference could be solved by an early return in `request_cursor_notify` based on `drawing_tablet_tool.focus_surface`. Unfortunately that means we need to add and maintain the tool, the `focused_surface` or just a new `boolean tablet_tool_in_proximity` in `seat`.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -42,6 +42,7 @@
 #include <wlr/types/wlr_tearing_control_v1.h>
 #include <wlr/types/wlr_text_input_v3.h>
 #include <wlr/types/wlr_input_method_v2.h>
+#include <wlr/types/wlr_tablet_v2.h>
 #include <wlr/util/log.h>
 #include "config/keybind.h"
 #include "config/rcxml.h"
@@ -332,6 +333,8 @@ struct server {
 
 	struct wlr_input_method_manager_v2 *input_method_manager;
 	struct wlr_text_input_manager_v3 *text_input_manager;
+
+	struct wlr_tablet_manager_v2 *tablet_manager;
 
 	/* Set when in cycle (alt-tab) mode */
 	struct osd_state {
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -354,8 +354,9 @@ update_pressed_surface(struct seat *seat, struct cursor_context *ctx)
 	return false;
 }
 
-static void
-process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
+static bool
+process_cursor_motion_out_of_surface(struct server *server,
+		double *sx, double *sy)
 {
 	struct view *view = server->seat.pressed.view;
 	struct wlr_scene_node *node = server->seat.pressed.node;
@@ -374,28 +375,29 @@ process_cursor_motion_out_of_surface(struct server *server, uint32_t time)
 #endif
 	} else {
 		wlr_log(WLR_ERROR, "Can't detect surface for out-of-surface movement");
-		return;
+		return false;
 	}
 
-	double sx = server->seat.cursor->x - lx;
-	double sy = server->seat.cursor->y - ly;
+	*sx = server->seat.cursor->x - lx;
+	*sy = server->seat.cursor->y - ly;
 	/* Take into account invisible xdg-shell CSD borders */
 	if (view && view->type == LAB_XDG_SHELL_VIEW) {
 		struct wlr_box geo;
 		wlr_xdg_surface_get_geometry(xdg_surface_from_view(view), &geo);
-		sx += geo.x;
-		sy += geo.y;
+		*sx += geo.x;
+		*sy += geo.y;
 	}
 
-	wlr_seat_pointer_notify_motion(server->seat.seat, time, sx, sy);
+	return true;
 }
 
 /*
- * Common logic shared by cursor_update_focus() and process_cursor_motion()
+ * Common logic shared by cursor_update_focus(), process_cursor_motion()
+ * and cursor_axis()
  */
-static void
+static bool
 cursor_update_common(struct server *server, struct cursor_context *ctx,
-		uint32_t time_msec, bool cursor_has_moved)
+		uint32_t time_msec, bool cursor_has_moved, double *sx, double *sy)
 {
 	struct seat *seat = &server->seat;
 	struct wlr_seat *wlr_seat = seat->seat;
@@ -407,7 +409,7 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 		 * Prevent updating focus/cursor image during
 		 * interactive move/resize
 		 */
-		return;
+		return false;
 	}
 
 	/* TODO: verify drag_icon logic */
@@ -422,9 +424,9 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 			 * we can keep scrolling or selecting text even
 			 * if the cursor moves outside of the surface.
 			 */
-			process_cursor_motion_out_of_surface(server, time_msec);
+			return process_cursor_motion_out_of_surface(server, sx, sy);
 		}
-		return;
+		return false;
 	}
 
 	if (ctx->surface && !input_inhibit_blocks_surface(seat,
@@ -457,8 +459,9 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 			seat->server_cursor = LAB_CURSOR_CLIENT;
 		}
 		if (cursor_has_moved) {
-			wlr_seat_pointer_notify_motion(wlr_seat, time_msec,
-				ctx->sx, ctx->sy);
+			*sx = ctx->sx;
+			*sy = ctx->sy;
+			return true;
 		}
 	} else {
 		/*
@@ -477,6 +480,7 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 			cursor_set(seat, cursor);
 		}
 	}
+	return false;
 }
 
 uint32_t
@@ -495,16 +499,16 @@ cursor_get_resize_edges(struct wlr_cursor *cursor, struct cursor_context *ctx)
 	return resize_edges;
 }
 
-static void
-process_cursor_motion(struct server *server, uint32_t time)
+bool
+cursor_process_motion(struct server *server, uint32_t time, double *sx, double *sy)
 {
 	/* If the mode is non-passthrough, delegate to those functions. */
 	if (server->input_mode == LAB_INPUT_STATE_MOVE) {
 		process_cursor_move(server, time);
-		return;
+		return false;
 	} else if (server->input_mode == LAB_INPUT_STATE_RESIZE) {
 		process_cursor_resize(server, time);
-		return;
+		return false;
 	}
 
 	/* Otherwise, find view under the pointer and send the event along */
@@ -514,7 +518,7 @@ process_cursor_motion(struct server *server, uint32_t time)
 	if (ctx.type == LAB_SSD_MENU) {
 		menu_process_cursor_motion(ctx.node);
 		cursor_set(&server->seat, LAB_CURSOR_DEFAULT);
-		return;
+		return false;
 	}
 
 	if (seat->drag.active) {
@@ -543,7 +547,8 @@ process_cursor_motion(struct server *server, uint32_t time)
 		}
 	}
 
-	cursor_update_common(server, &ctx, time, /*cursor_has_moved*/ true);
+	return cursor_update_common(server, &ctx, time,
+		/* cursor_has_moved */ true, sx, sy);
 }
 
 static uint32_t
@@ -569,8 +574,9 @@ _cursor_update_focus(struct server *server)
 			ctx.surface, rc.raise_on_focus);
 	}
 
+	double sx, sy;
 	cursor_update_common(server, &ctx, msec(&now),
-		/*cursor_has_moved*/ false);
+		/*cursor_has_moved*/ false, &sx, &sy);
 }
 
 void
@@ -738,7 +744,11 @@ preprocess_cursor_motion(struct seat *seat, struct wlr_pointer *pointer,
 	 * without any input.
 	 */
 	wlr_cursor_move(seat->cursor, &pointer->base, dx, dy);
-	process_cursor_motion(seat->server, time_msec);
+	double sx, sy;
+	bool notify = cursor_process_motion(seat->server, time_msec, &sx, &sy);
+	if (notify) {
+		wlr_seat_pointer_notify_motion(seat->seat, time_msec, sx, sy);
+	}
 }
 
 static void
@@ -947,9 +957,8 @@ handle_press_mousebinding(struct server *server, struct cursor_context *ctx,
 
 static uint32_t press_msec;
 
-static void
-cursor_button_press(struct seat *seat, uint32_t button,
-		enum wlr_button_state button_state, uint32_t time_msec)
+bool
+cursor_process_button_press(struct seat *seat, uint32_t button, uint32_t time_msec)
 {
 	struct server *server = seat->server;
 	struct cursor_context ctx = get_cursor_context(server);
@@ -972,7 +981,7 @@ cursor_button_press(struct seat *seat, uint32_t button,
 		 * so subsequent release always closes menu or selects menu item.
 		 */
 		press_msec = 0;
-		return;
+		return false;
 	}
 
 	/*
@@ -1012,7 +1021,7 @@ cursor_button_press(struct seat *seat, uint32_t button,
 		 * Note: This does not work for XWayland clients
 		 */
 		wlr_seat_pointer_end_grab(seat->seat);
-		return;
+		return false;
 	}
 
 	/* Bindings to the Frame context swallow mouse events if activated */
@@ -1021,14 +1030,15 @@ cursor_button_press(struct seat *seat, uint32_t button,
 
 	if (ctx.surface && !consumed_by_frame_context) {
 		/* Notify client with pointer focus of button press */
-		wlr_seat_pointer_notify_button(seat->seat, time_msec,
-			button, button_state);
+		return true;
 	}
+
+	return false;
 }
 
-static void
-cursor_button_release(struct seat *seat, uint32_t button,
-		enum wlr_button_state button_state, uint32_t time_msec)
+bool
+cursor_process_button_release(struct seat *seat, uint32_t button,
+		uint32_t time_msec)
 {
 	struct server *server = seat->server;
 	struct cursor_context ctx = get_cursor_context(server);
@@ -1043,24 +1053,20 @@ cursor_button_release(struct seat *seat, uint32_t button,
 				menu_call_selected_actions(server);
 			} else {
 				menu_close_root(server);
+				double sx, sy;
 				cursor_update_common(server, &ctx, time_msec,
-					/*cursor_has_moved*/ false);
+					/*cursor_has_moved*/ false, &sx, &sy);
 			}
 		}
-		return;
+		return false;
 	}
 
 	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
 		if (pressed_surface) {
 			/* Ensure CSD clients see the release event */
-			wlr_seat_pointer_notify_button(seat->seat, time_msec,
-				button, button_state);
+			return true;
 		}
-
-		/* Exit interactive move/resize mode */
-		interactive_finish(server->grabbed_view);
-
-		return;
+		return false;
 	}
 
 	if (pressed_surface && ctx.surface != pressed_surface) {
@@ -1068,9 +1074,7 @@ cursor_button_release(struct seat *seat, uint32_t button,
 		 * Button released but originally pressed over a different surface.
 		 * Just send the release event to the still focused surface.
 		 */
-		wlr_seat_pointer_notify_button(seat->seat, time_msec,
-			button, button_state);
-		return;
+		return true;
 	}
 
 	/* Bindings to the Frame context swallow mouse events if activated */
@@ -1079,9 +1083,25 @@ cursor_button_release(struct seat *seat, uint32_t button,
 
 	if (!consumed_by_frame_context) {
 		/* Notify client with pointer focus of button release */
-		wlr_seat_pointer_notify_button(seat->seat, time_msec,
-			button, button_state);
+		return true;
 	}
+
+	return false;
+}
+
+bool
+cursor_finish_button_release(struct seat *seat)
+{
+	struct server *server = seat->server;
+
+	if (server->input_mode == LAB_INPUT_STATE_MOVE
+			|| server->input_mode == LAB_INPUT_STATE_RESIZE) {
+		/* Exit interactive move/resize mode */
+		interactive_finish(server->grabbed_view);
+		return true;
+	}
+
+	return false;
 }
 
 static void
@@ -1095,12 +1115,24 @@ cursor_button(struct wl_listener *listener, void *data)
 	struct wlr_pointer_button_event *event = data;
 	idle_manager_notify_activity(seat->seat);
 
+	bool notify;
 	switch (event->state) {
 	case WLR_BUTTON_PRESSED:
-		cursor_button_press(seat, event->button, event->state, event->time_msec);
+		notify = cursor_process_button_press(seat, event->button,
+			event->time_msec);
+		if (notify) {
+			wlr_seat_pointer_notify_button(seat->seat, event->time_msec,
+				event->button, event->state);
+		}
 		break;
 	case WLR_BUTTON_RELEASED:
-		cursor_button_release(seat, event->button, event->state, event->time_msec);
+		notify = cursor_process_button_release(seat, event->button,
+			event->time_msec);
+		if (notify) {
+			wlr_seat_pointer_notify_button(seat->seat, event->time_msec,
+				event->button, event->state);
+		}
+		cursor_finish_button_release(seat);
 		break;
 	}
 }
@@ -1129,7 +1161,11 @@ cursor_emulate_move_absolute(struct seat *seat, struct wlr_input_device *device,
 		dx, dy, dx, dy);
 
 	wlr_cursor_move(seat->cursor, device, dx, dy);
-	process_cursor_motion(seat->server, time_msec);
+	double sx, sy;
+	bool notify = cursor_process_motion(seat->server, time_msec, &sx, &sy);
+	if (notify) {
+		wlr_seat_pointer_notify_motion(seat->seat, time_msec, sx, sy);
+	}
 	wlr_seat_pointer_notify_frame(seat->seat);
 }
 
@@ -1139,12 +1175,20 @@ cursor_emulate_button(struct seat *seat, uint32_t button,
 {
 	idle_manager_notify_activity(seat->seat);
 
+	bool notify;
 	switch (state) {
 	case WLR_BUTTON_PRESSED:
-		cursor_button_press(seat, button, state, time_msec);
+		notify = cursor_process_button_press(seat, button, time_msec);
+		if (notify) {
+			wlr_seat_pointer_notify_button(seat->seat, time_msec, button, state);
+		}
 		break;
 	case WLR_BUTTON_RELEASED:
-		cursor_button_release(seat, button, state, time_msec);
+		notify = cursor_process_button_release(seat, button, time_msec);
+		if (notify) {
+			wlr_seat_pointer_notify_button(seat->seat, time_msec, button, state);
+		}
+		cursor_finish_button_release(seat);
 		break;
 	}
 }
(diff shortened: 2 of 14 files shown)


src/input/tablet.c
```diff
@@ -63,27 +66,97 @@ adjust_for_rotation(enum rotation rotation, double *x, double *y)
...
+	struct wlr_surface *surface = tablet_get_coords(tablet, x, y, &nx, &ny);
+
+	if (surface && tablet->tablet_v2_tool) {
+		wlr_tablet_v2_tablet_tool_notify_proximity_in(
+			tablet->tablet_v2_tool, tablet->tablet_v2, surface);
+
+		if (ev->updated_axes & (WLR_TABLET_TOOL_AXIS_X | WLR_TABLET_TOOL_AXIS_Y)) {
+			// Visualize where we are and what we are
+			wlr_cursor_warp_absolute(tablet->seat->cursor, tablet->base, x, y);
+			cursor_set(tablet->seat, LAB_CURSOR_PENCIL);
```
jp7677, line 121: Most likely not the best idea ..

src/input/tablet.c
```diff
@@ -63,27 +66,163 @@ adjust_for_rotation(enum rotation rotation, double *x, double *y)
...
+
+	double lx, ly;
+	struct wlr_surface *surface = tablet_get_coords(tablet, x, y, &lx, &ly);
+
+	if (surface) {
+		wlr_log(WLR_DEBUG, "  got proximity event on surface");
+		// TODO: This seems just plain wrong and causes crashes
+		if (!tablet->tablet_v2_tool || tablet->tablet_v2_tool->wlr_tool != ev->tool) {
+			// Unfortunately I found `wlr_tool` only in the events,
+			// so use this for creating a `wlr_tablet_v2_tablet_tool`
```
Consolatis, line 205: `struct wlr_tablet_pad` appears to have a signal for attaching a tool (`attach_tablet`).

src/input/tablet.c
Consolatis: I don't think we actually need to set the cursor manually here at all.
`struct wlr_tablet_v2_tablet_tool` has a `set_cursor()` signal though that supplies a wl_surface, so I assume we can handle it similar to `request_cursor_notify()` in `src/input/cursor.c`.

src/input/tablet.c
```diff
@@ -63,27 +66,183 @@ adjust_for_rotation(enum rotation rotation, double *x, double *y)
...
+				ev->distance);
+		}
+		if (ev->updated_axes & WLR_TABLET_TOOL_AXIS_PRESSURE) {
+			wlr_tablet_v2_tablet_tool_notify_pressure(tablet->tablet_v2_tool,
+				ev->pressure);
+		}
+		if (ev->updated_axes
+				& (WLR_TABLET_TOOL_AXIS_TILT_X | WLR_TABLET_TOOL_AXIS_TILT_Y)) {
+			wlr_tablet_v2_tablet_tool_notify_tilt(tablet->tablet_v2_tool,
+				tablet->tilt_x, tablet->tilt_y);
```
jp7677, line 178: Do we need to transform (area, rotation) tilt?

src/input/tablet.c
```diff
@@ -63,27 +66,183 @@ adjust_for_rotation(enum rotation rotation, double *x, double *y)
...
+		wlr_log(WLR_DEBUG, "  got proximity event on surface");
+		// TODO: This seems just plain wrong and causes crashes
+		if (!tablet->tablet_v2_tool || tablet->tablet_v2_tool->wlr_tool != ev->tool) {
+			// Unfortunately I found `wlr_tool` only in the events,
+			// so use this for creating a `wlr_tablet_v2_tablet_tool`
+			wlr_log(WLR_DEBUG, "  creating tablet_v2_tool");
+			if (tablet->tablet_v2_tool) {
+				wlr_log(WLR_DEBUG, "  tablet_v2_tool was already created");
+			}
+			tablet->tablet_v2_tool =
```
jp7677, line 226: Free somewhere?

src/input/tablet.c
```diff
@@ -133,13 +311,23 @@ tablet_init(struct seat *seat, struct wlr_input_device *wlr_device)
 	wlr_log(WLR_DEBUG, "setting up tablet");
 	struct drawing_tablet *tablet = znew(*tablet);
 	tablet->seat = seat;
+	tablet->base = wlr_device;
 	tablet->tablet = wlr_tablet_from_input_device(wlr_device);
 	tablet->tablet->data = tablet;
+	if (seat->server->tablet_manager) {
+		tablet->tablet_v2 = wlr_tablet_create(
```
jp7677, line 318: Free somewhere?

src/input/tablet_tool.c
```diff
@@ -0,0 +1,94 @@
...
+		wl_container_of(listener, tool, handlers.set_cursor);
+	struct wlr_tablet_v2_event_cursor *ev = data;
+
+	struct seat *seat = tool->tablet->seat;
+
+	if (seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
+		return;
+	}
+
+	wl_list_remove(&tool->handlers.surface_destroy.link);
```
jp7677, line 52: Do I need to guard this against concurrent calls from multiple client?

src/input/tablet_tool.c
```diff
@@ -0,0 +1,94 @@
...
+		ev->hotspot_x, ev->hotspot_y);
+}
+
+static void
+handle_destroy(struct wl_listener *listener, void *data)
+{
+	struct drawing_tablet_tool *tool =
+		wl_container_of(listener, tool, handlers.destroy);
+
+	wl_list_remove(&tool->link);
```
jp7677, line 70: Also clean up listeners?

src/input/tablet_tool.c
johanmalm: I don't know, but it feels like only the 'focused' client should be allowed to request `set_cursor` (just guessing though).

src/input/tablet_tool.c
johanmalm: Yes, most likely. `set_cursor` feels like a certainty. Not sure about `proximity`, etc - possible they get destroyed by the manager.
Might be worth doing some printf()s to make sure it balances. Always a tricky area to get right.

src/input/tablet_tool.c
```diff
@@ -0,0 +1,97 @@
...
+#include "config/rcxml.h"
+#include "input/cursor.h"
+#include "input/tablet_tool.h"
+#include "labwc.h"
+
+void
+tablet_tool_cursor_reset(struct drawing_tablet_tool *tool)
+{
+	cursor_set(tool->seat, LAB_CURSOR_DEFAULT);
+	cursor_update_image(tool->seat);
```
jp7677, line 18: note to self: probably no longer needed?

src/input/tablet.c
Consolatis: I looked a bit into wlroots and the issue is that wlroots destroys the tool on proximity_out when it can't be uniquely identified. So using the proximity_in signal to create the tool wrapper seems like the correct solution.

Consolatis:
Managed to somewhat fix my test setup (the battery in the pen was leaking and corroded some contacts) and this seems to work great. Only issue I saw while testing is that I was unable to resize the `gtk4-demo --run=paint` window with the pen. I am not quite sure why that is though, it might be a GTK bug in their tablet input handling. Usually CSD clients should request entering interactive resize mode from the compositor when a client clicks on the invisible area around the window but that doesn't appear to happen when the input comes from a tablet. Same thing for moving the (CSD) window when dragging the titlebar.

I am also not sure about the cursor surface destroy listener, we do not use it in `src/cursor.c` either and I'd assume (not actually verified) that wlroots handles that internally. Another thing is that we IMHO should not hardcode the focus behavior here but rather call the same user defined action as when pressing the left mouse button.

jp7677:
> Managed to somewhat fix my test setup (the battery in the pen was leaking and corroded some contacts) and this seems to work great. 

Thanks a lot for starting to look into this and cool to hear that you got your tablet working again.

> Only issue I saw while testing is that I was unable to resize / move the `gtk4-demo --run=paint` window with the pen. I am not quite sure why that is though, it might be a GTK bug in their tablet input handling. Usually CSD clients should request entering interactive resize mode from the compositor when a client clicks on the invisible area around the window but that doesn't appear to happen when the input comes from a tablet. Same thing for moving the (CSD) window when dragging the titlebar.

Yeah, I haven't figured yet what is need for resizing or moving CSD applications, I can't resize/move GTK3 nor GTK4 applications. I'm somehow convinced that _I'm_ missing something ;). I've already tried adding `wlr_tablet_tool_v2_start_implicit_grab()` to tip-down, but that didn't do the trick.

> I am also not sure about the cursor surface destroy listener, we do not use it in `src/input/cursor.c` either and I'd assume (not actually verified) that wlroots handles that internally. 

`wlroots` adds its own listener which effectively hides the cursor on surface destroy (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_cursor.c#L665 / https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_cursor.c#L633). I've added this to prevent the behavior described in https://github.com/labwc/labwc/issues/1672#issuecomment-2045775374 . But yeah, I should try again without because I've changed a few things in the meantime.

> Another thing is that we IMHO should not hardcode the focus behavior here but rather call the same user defined action as when pressing the left mouse button, similar to what is done in [touch.c](https://github.com/labwc/labwc/blob/master/src/input/touch.c#L115).

Ah right, I can do that instead of `desktop_focus_view`.

Consolatis:
> > I am also not sure about the cursor surface destroy listener, we do not use it in `src/input/cursor.c` either and I'd assume (not actually verified) that wlroots handles that internally.
> 
> `wlroots` adds its own listener which effectively hides the cursor on surface destroy (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_cursor.c#L665 / https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_cursor.c#L633). I've added this to prevent the behavior described in [#1672 (comment)](https://github.com/labwc/labwc/issues/1672#issuecomment-2045775374) . But yeah, I should try again without because I've changed a few things in the meantime.

Assuming you were referring to this part of the comment:

> I've noticed that when the last window of Firefox or Thunderbird gets closed and the cursor falls to the desktop then it will disappear until it hovers a new window. I couldn't reproduce it with any other GTK/Qt/Xwayland app.

As long as the cursor is set to `LAB_CURSOR_CLIENT` the `cursor_update_focus()` call on view destruction should handle that.

jp7677:
> As long as `seat->server_cursor` is set to `LAB_CURSOR_CLIENT` the `cursor_update_focus()` call on view destruction should handle that.

yes, it was needed anymore. Nice!

include/labwc.h
```diff
@@ -114,6 +115,13 @@ struct seat {
 
 	struct wlr_pointer_constraint_v1 *current_constraint;
 
+	/*
+	 * Tracks the top-most surface underneath a tablet tool when the
+	 * tool is in proximity.
+	 * Updating happens on tablet proximity and tablet axis events.
+	 */
+	struct wlr_surface *tablet_tool_surface;
```
jp7677, line 123: note to self: might also track the tool here instead, may be that allows a cleaner plumbing of the tablet functions into the general cursor logic.

jp7677:
@Consolatis I've managed to integrate the tablet handling into the existing cursor handling by exposing a few general methods that return if a notify (pointer or tablet) should follow. From my testing and after a few iterations this does the job quite good, even exotic things like continue scrolling with moving the stylus out of surface works. That said, the changes I made to the cursor functions are a bit hard to digest and certainly introduced complexity and ugliness. On the plus  side, it might be feasible now to extract the pointer specific signal handling from `cursor.c` into a another file for better separation. 
Please let me know what you think of that direction and also ofc if you have an inspiration on how to merge the pointer and tablet stuff more elegantly.
PS: The commits are currently somewhat messy, I'll clean this up once finished.

src/input/cursor.c
```diff
@@ -140,6 +140,14 @@ request_cursor_notify(struct wl_listener *listener, void *data)
 		return;
 	}
 
+	if (!wl_list_empty(&seat->server->tablet_manager->clients)) {
```
jp7677, line 143: note to self: TODO: this prevents the drag cursor going back to a normal cursor when finishing moving a CSD application.

jp7677: Remove that part. While it fixed the hovering into a GTK4 app that wants a different cursor when entering its surface (thus when somewhat still being over de resize area), it also prevented setting a cursor in XWayland apps without tablet support (e.g. VS Code) next to the remaining cursor after moving.

src/interactive.c
```diff
@@ -251,4 +251,5 @@ interactive_cancel(struct view *view)
 
 	/* Update focus/cursor image */
 	cursor_update_focus(view->server);
+	cursor_set(&view->server->seat, LAB_CURSOR_DEFAULT);
```
jp7677, line 254: note to self: probably controversial.

src/input/cursor.c
jp7677: Re-added, somehow still makes sense to prevent those conflicts.Mitigated by resetting the cursor in interactive_cancel. Let's play some more.

Consolatis:
> Edit: please don’t do a full code review yet, I need to e.g. polish and add code comments first, for now it’s about the direction of combining the cursor and tablet magic.

I think the general idea of having the cursor functions return a bool which is then used by the caller to either notify the client or not is sensible. It should likely be split into its own PR when everything works as intended for the tablet use-case (e.g. there aren't any missing pieces in the cursor part to make the tablet part make use if it). That makes the review a bit simpler.

jp7677:
> I think the general idea of having the cursor functions return a bool which is then used by the caller to either notify the client or not is sensible. It should likely be split into its own PR when everything works as intended for the tablet use-case (e.g. there aren't any missing pieces in the cursor part to make the tablet part make use if it). That makes this PR (and its review) a bit simpler. 

Yes, good idea. I've already isolated that to a single commit https://github.com/labwc/labwc/pull/1678/commits/ccd715660758d4abfe9d44bfb9f481399c88d462 , should be trivial to make that a separate PR once everything is in place and confirmed to work.

> Overall I think we should not ship this feature in `0.7.2` (targeted feature deadline on April 25th).

Fully agree, I'm sure I'll still need a few weeks to get everything sorted ;)

jp7677:
I'm running this PR now for daily use for a few days with no mouse attached and it works imho pretty good. A few minor quirks I have noticed:
- When moving or resizing the Steam window, the cursor (sometimes) in Steam itself doesn't change shape anymore, e.g. no hand cursor above links. The UI still reacts to the cursor and click works, just the shape is off. The cursor shape restores as soon as the stylus goes proximity out/in or hovering out/in again.
- When two application windows are side by side, the left one is CSD/GTK4 and the right one is focused, I cannot grab the scrollbar of the GTK4 application. Click on scrollbar does work and scrolls, but not grabbing the scrollbar. Once the GTK4 application is focused, grabbing the scrollbar works again. Grabbing works when the left application was e.g. Firefox or Tilix (both GTK3).

@rhyses-pieces experiences cursor flickering above XWayland applications, notably Krita, but this also happens in Sway and might be rooted outside of this PR. Furthermore a labwc restart is needed for restoring pen functionality after a reconnect (the display of the tablet does get connected again though). Both scenarios do work fine on my setup.

jp7677:
> When two application windows are side by side, the left one is CSD/GTK4 and the right one is focused, I can grab the scrollbar of the GTK4 application but cannot move it (or just a tiny bit) . Sometimes it also doesn't work when the GTK4 app is focused. Click on scrollbar does work and scrolls, but not grabbing the scrollbar. Grabbing works when the left application is e.g. Firefox or Tilix (both GTK3).

I'm slightly lost with this issue. Scrolling works fine when no other application is adjacent to e.g. nautilus, but as soon as another application is (and that other application has focus), I cannot get hold of the scrollbar for as long as I hold tap down, only for a tiny moment. The behavior is fine when using a mouse or mouse emulation and also when using the stylus in Gnome. Any idea's?

01micko:
@jp7677 probably pay to ping @Consolatis @jlindgren90 @ahesford @tokyo4j :)

re https://github.com/labwc/labwc/pull/1678#issuecomment-2081349929

Er, done!

johanmalm:
What you describe sounds familiar, but I've not looked at patch, so might be barking up the wrong tree.

When grabbing a scrollbar, you have to keep the associated surface focused even if the cursor moves out over other clients/surfaces.

Ref:
- https://github.com/labwc/labwc/blob/2603dbf396bddeebefeb8e6e1f4ef7cce3b4eb31/src/input/cursor.c#L327
- https://github.com/labwc/labwc/blob/2603dbf396bddeebefeb8e6e1f4ef7cce3b4eb31/src/input/cursor.c#L417-L424

jp7677:
> What you describe sounds familiar, but I've not looked at patch, so might be barking up the wrong tree.
> 
> When grabbing a scrollbar, you have to keep the associated surface focused even if the cursor moves out over other clients/surfaces.

Thanks for the reply and sorry for the late response, I had been away from keyboard for a while. 

Actually I'm reusing all the common cursor logic you are referring to, so the tablet code hits all the code places you mentioned. I still don't know what exactly is causing this weird scrolling behavior I'm seeing when another application is adjacent to a GTK4 application. The scrolling behavior works perfectly when no other application is adjacent. Whatever I do code-wise, I cannot get the scrolling to work correctly when using tablet notifications. I tend to think that this is more a GTK4 quirk (their tablet protocol implementation i.c.w. focus behavior?), but dunno. I don't have this issue with e.g. GTK3 applications and, interestingly, enabling focus-follows-mouse also completely mitigates this issue.

jp7677:
> When moving or resizing the Steam window, the cursor (sometimes) in Steam itself doesn't change shape anymore, e.g. no hand cursor above links. The UI still reacts to the cursor and click works, just the shape is off. The cursor shape restores as soon as the stylus goes proximity out/in or hovering out/in again.

This should be resolved now, I'm doing now an additional proximity-out/proximity-in after an interactive move/resize which seems to do the trick. With this I could also remove a few hacky lines in the last commit.

So only the GTK4 scrollbar quirk is left.

jp7677:
> So only the GTK4 scrollbar quirk is left (based on my own testing).

I found a way to work around this issue, by doing additionally `wlr_seat_pointer_notify_motion(tool->seat->seat, time, 0, 0)` to `wlr_tablet_v2_tablet_tool_notify_motion()`, see last commit.  I haven't seen yet any negative side effects, but feels rather ugly. That said, from a functional perspective, all should be good now.

johanmalm:
I'm only half way through the commit in terms of review.

Comments so far:

1. `69e3c36`: When calling `cursor_update_common()` (nx, ny) has been used rather than (sx, sy).
   See: https://github.com/labwc/labwc/blob/2f1fcb4468d1d6e3442a3cce2d13ea26f3a7b805/CONTRIBUTING.md?plain=1#L311
   
2. `6c22773` Same comment for `tablet_get_coords()`
  
3. `d020bf5` In `struct drawing_tablet` you have added `struct wlr_input_device *base;`. Normally wlroots (and us sometimes) use base to refer to a base struct within a derived struct in the OOP-inheritance sense. It allows getting the inherited struct from the base with constructs such as the one below. It doesn't look like you've used it in that way though, so `struct wlr_input_device *wlr_input_device;` is easier to read where it's dereferenced - and more consistent with code base.


```
struct input *input;
wl_list_for_each(input, &seat->inputs, link) {
  if (input->wlr_input_device->type == WLR_INPUT_DEVICE_KEYBOARD) {
    struct keyboard *keyboard = (struct keyboard *)input;
    ...
  }
}
```

- https://github.com/labwc/labwc/blob/2f1fcb4468d1d6e3442a3cce2d13ea26f3a7b805/src/seat.c#L30
- https://github.com/labwc/labwc/blob/2f1fcb4468d1d6e3442a3cce2d13ea26f3a7b805/include/labwc.h#L84

4. I don't get why we do `wlr_seat_pointer_notify_motion(tool->seat->seat, time, -1, -1);` (probably just me being a bit slow).

jp7677:
> I'm only half way through the commit in terms of review.

Thanks a lot for already looking into the details!
 
> Comments so far:

I've addressed the renames in point 1, 2 and 3.

>  I don't get why we do `wlr_seat_pointer_notify_motion(tool->seat->seat, time, -1, -1);` (probably just me being a bit slow).

By re-using the existing cursor logic, this PR also sets pointer focus (`wlr_seat_pointer_notify_clear_focus` / `wlr_seat_pointer_notify_enter`) when a tablet pen enters a surface. This is a good thing. A side effect is though, that a client will be notified about a pointer event (pointer enter at coordinate) and tablet events (proximity-in, move to coordinate) at the same time. The following tablet actions, as long as the pen stays on the surface, emit only tablet events. That said, the client still thinks that there is a pointer at the coordinate where the tablet entered the surface. This can conflict with e.g. menu or scrollbars because the client might be conflicted about which coordinates (pointer or tablet) have priority when both coordinates are on the same menu or scrollbar.  I figured accidentally that -1/-1 are valid coordinates (well, in the sense that it works). Notifying pointer motion to -1/-1, the pointer is kind of out of the way and is never on the same control like the tablet pen, so no conflicts anymore for the client. 
Does this makes sense?

Note that Gnome/Mutter sends a pointer-leave notification on tablet proximity-in to the client to avoid this conflict. Going that way would probably involve much more refactoring in labwc and I'm not sure what will break since I have the feeling that a lot of internals rely on correct pointer focus.

johanmalm:
Makes sense now. There is already a fairly length comment, but suggest adding some/all of the above for future devs :smile:

jp7677:
> Makes sense now. There is already a fairly length comment, but suggest adding some/all of the above for future devs 😄

Thanks and done!

jp7677:
I think this should be good now. Please let me know if i should open a separate PR for the first commit.

johanmalm:
LGTM - but please don't merge yet.
The @Consolatis is also reviewing and they have a track-record of finding things!

Consolatis:
This looks good to me as well. Didn't detect any issue while testing and overall feels very complete.

With the comment added as raised by malm I'd say lets go for the merge here.

I am not particularly happy about the cursor parking at -1 -1 but we can deal with that in another PR. Same story for drag and drop support. Thanks a lot for all the work you've put into this!

jp7677:
> This looks good to me as well. Didn't detect any issue while testing and overall feels very complete. With the comment added as raised by malm I'd say lets go for the merge here.

Cool, I amended two commits with comments. I hope they make some sense.

> I am not particularly happy about the cursor parking at -1 -1 but we can deal with that in another PR. Same story for drag and drop support. Thanks a lot for all the work you've put into this!

It was quite a ride  ;)

src/input/tablet.c
```diff
@@ -93,60 +177,175 @@ handle_axis(struct wl_listener *listener, void *data)
...
+		tablet->tilt_x = ev->tilt_x;
+	}
+	if (ev->updated_axes & WLR_TABLET_TOOL_AXIS_TILT_Y) {
+		tablet->tilt_y = ev->tilt_y;
+	}
+
+	double x, y;
+	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y);
+
+	if (tool && ((surface
```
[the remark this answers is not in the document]
jp7677: Should be good now.

src/input/tablet.c
```diff
@@ -75,77 +82,410 @@ adjust_for_rotation(enum rotation rotation, double *x, double *y)
...
+{
+	idle_manager_notify_activity(tool->seat->seat);
+
+	bool enter_surface = false;
+	/* Postpone proximity-in on a new surface when the tip is down */
+	if (surface != tool->tool_v2->focused_surface && !tool->tool_v2->is_down) {
+		enter_surface = true;
+		wlr_tablet_v2_tablet_tool_notify_proximity_in(tool->tool_v2,
+			tablet->tablet_v2, surface);
+	}
```
tokyo4j, line 129: @jp7677 What's the reason behind `!tool->tool_v2->is_down`?

(I'm now reading `src/input/tablet.c` to refactor the entire logic of cursor handling and reuse `cursor_process_*` functions in `src/input/touch.c`).

jp7677: From the commit message of https://github.com/labwc/labwc/pull/1678/commits/594717d03b6043b6106854aa45bbd3d16059e819

`Also postpone proximity-in when moving to a new surface with the tip down.`

Its been a while, but i think it had to do with not interrupting drag&drop from not-tablet capable surfaces to tablet capable surfaces. 

Please be careful with the tablet stuff, it was far from trivial to get the tablet interaction right considering that the tablet protocol is completely separated from the normal pointer but somehow had to be bolted into it to allow normal desktop/window interactions and not just actual drawing.

tokyo4j: > Its been a while, but i think it had to do with not interrupting drag&drop from not-tablet capable surfaces to tablet capable surfaces.

But isn't it already handled in the check against `is_down_mouse_emulation`?

https://github.com/labwc/labwc/blob/7bf08af210968fb174149754ce341988365a131b/src/input/tablet.c#L372-L376

I don't know applications that support DnD and don't support tablet protocol, but at least out-of-surface dragging from a non-tablet-capable surface to a tablet-capable surface worked without the `!tool->tool_v2->is_down` check.

Please note that I never intend to blame you. I just want to make sure that my future changes don't break any tablet functionalities.

jp7677: Looking more closely, I think now that it was related to out-of-surface scrolling for a tablet-capable application when the out-surface area is covered by a second tablet-capable surface (e.g. out-of-surface scrolling in firefox above another GTK application). So to prevent a surface change for the tool (proximity out/in) as long as the out-of-surface scrolling is performed.

> Please note that I never intend to blame you. I just want to make sure that my future changes don't break any tablet functionalities.

No worries, all good. Thank you for asking in the first place and once we have the reason, it surely justifies a comment.

tokyo4j: I don't think the check for `tool->tool_v2->is_down` is needed for out-of-surface scrolling from a tablet-capable surface to another tablet-capable surface either, because we set implicit grab when the tip is down, which makes `wlr_tablet_v2_tablet_tool_notify_proximity_in()` no-op: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/tablet_v2/wlr_tablet_v2_tool.c?ref_type=heads#L748

jp7677: I think I wasn't aware of this fact when writing that code. I think I just looked for a counterpart of `is_down_mouse_emulation` but from a tablet-capable surface. Sorry, I don't remember exactly anymore which scenario/edge case I had in mind back then and if I had tested without `is_down`.

tokyo4j: Thanks, that's totally fine. I was just wondering if I'm missing any edge-cases to test with.

I really appreciate your effort on this PR that made labwc more accessible to tablet users!

Issue #1684, opened 2024-04-06
Problem with title-bar translucency


tsujan:
I explain the issue with an example:

When I use `window.active.title.bg.color: #3a3a3b 95`, I apparently have a nice, translucent title-bar:

![95](https://github.com/labwc/labwc/assets/981076/8c820513-8bb3-44a8-ac4d-23de67542cd9)

However, there is a problem, which is made visible when I decrease the opacity, like `window.active.title.bg.color: #3a3a3b 70`:

![70](https://github.com/labwc/labwc/assets/981076/85e4c17b-a97a-4899-9e32-0e993b92656c)

As the screenshot shows, the opacity of the the left and right corners is different from that of the title-bar. It should be related to how the rounded corners are drawn; it doesn't exist in maximized windows.

I use 0.7.1 plus the patch for shadows.

P.S. There's also another, minor issue with title-bar translucency, which I may explain elsewhere.

ahesford:
Does the problem go away for non-maximized windows if you set the corner radius to 0?

tsujan:
No, it doesn't, contrary to what I expected: The issue is still visible with `<cornerRadius>0</cornerRadius>` and `window.active.title.bg.color: #3a3a3b 70`.

Consolatis:
The corners are drawn by cairo, the titlebar and other button backgrounds are drawn by wlroots via the scene graph API. Maybe this is an issue of pre-multipled vs not premultipled rgb values.

tsujan:
> Maybe this is an issue of pre-multipled vs non pre-multipled RGB values

@Consolatis, does the following observation confirm that?

If I change the color to back, everything is OK. I tested with `window.active.title.bg.color: #000000 70` and `window.active.title.bg.color: #000000 50` to make sure.

This may also be related to the second issue that I found (but didn't explain above): With translucency, the title-bar color isn't exactly what it should be. To make the translucent title-bar match my Qt windows, I used `#3a3a3b`, while the color of my Qt windows was `#3d3d3e` — in other words, I had to decrease the color value of the title-bar from 62 to 59 to make it match the window.

tsujan:
BTW, I encountered the second issue in Wayfire too (the necessity of decreasing the color value with translucency).

Consolatis:
>  in other words, I had to decrease the color value of the title-bar from 62 to 59 to make it match the window.

That sounds more like a rounding issue (alpha is usually between 0-255 and we are using percent value e.g. 0-100).
[...]

PR #1685, opened 2024-04-06
Use pre-multiplied colors by default

Consolatis:
wlr_scene_rects expect their color to be pre-multiplied while cairo_set_source_rgba() expects them to not be pre-multiplied.
With this patch we now use premultiplied colors internally by default and then reverse it when setting cairo colors.

This ensures the titlebar uses a consistent color in case it was defined with some transparency by the user.

Fixes:
- #1684

### DIFF
--- a/src/common/graphic-helpers.c
+++ b/src/common/graphic-helpers.c
@@ -85,7 +85,19 @@ draw_cairo_border(cairo_t *cairo, struct wlr_fbox fbox, double line_width)
 void
 set_cairo_color(cairo_t *cairo, const float *c)
 {
-	cairo_set_source_rgba(cairo, c[0], c[1], c[2], c[3]);
+	/*
+	 * We are dealing with pre-multiplied colors
+	 * but cairo expects unmultiplied colors here
+	 */
+	float alpha = c[3];
+
+	if (alpha == 0.0f) {
+		cairo_set_source_rgba(cairo, 0, 0, 0, 0);
+		return;
+	}
+
+	cairo_set_source_rgba(cairo, c[0] / alpha, c[1] / alpha,
+		c[2] / alpha, alpha);
 }
 
 struct surface_context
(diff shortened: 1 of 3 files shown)


Consolatis:
This might need a follow-up PR to adjust hardcoded values for cairo colors. I don't have the time to look through the codebase currently but AFAIR it is mostly the fallback button hover overlays and the region overlays that use transparency.

johanmalm:
LGTM. Not tested.

Consolatis:
Translated the hardcoded fallback button hover overlay color into pre-multiplied, the snap / region overlays already used pre-multiplied before because they are fed directly into a wlr_scene_rect anyway.

I stumbled on the `argb_32()` function `src/button/button-xbm.c`. I am not sure about that one. I assume it should also pre-multiply because (I assume) wlroots expects pre-multiplied pixels in its `DRM_FORMAT_ARGB8888` buffers but couldn't verify that yet.

Consolatis:
Handled the case where we could end up with a division by zero. Thanks for notifying @jlindgren90.

Unless there are any objections I'll merge later.

Joshua-Ashton:
I know this has been merged but it seems wrong to use premul when you have different parts of the pipeline wanting different. There's obvious issues with quantization and 0 there.

jlindgren90:
Quantization should not be an issue since we use floating point, I think? Is there a case you can think of where it will be a problem?

johanmalm:
@Joshua-Ashton Appreciate your thought on thinking this through. My brain has never quite tuned into the pre-multiplied discussions - wish I had paid more attention now :smile:

Joshua-Ashton:
Ah yes, it should be fine in this impl. I thought we were putting stuff back as uint8_t somewhere.

johanmalm:
> Ah yes, it should be fine in this impl. I thought we were putting stuff back as uint8_t somewhere.

Thanks. Appreciate the independent check.

PR #1691, opened 2024-04-07
src/theme.c: support inline alpha encoding like #aabbccff

Consolatis:


Consolatis:
Added a log message and note in the docs for the deprecation of the current `#rrggbb aaa` encoding as discussed via IRC.

johanmalm:
Thanks

johanmalm:
Did you want to add a note in the color section of https://labwc.github.io/labwc-theme.5.html

Consolatis:
I did add
> Note: the #rrggbb aaa notation is deprecated starting from
	labwc 0.7.2 and may be removed in future releases.

But didn't want to blow up the section with rationales of why we don't support the other strange Openbox color formats.

johanmalm:
> I did add
> 
> > Note: the #rrggbb aaa notation is deprecated starting from
> > labwc 0.7.2 and may be removed in future releases.
> 
> But didn't want to blow up the section with rationales of why we don't support the other strange Openbox color formats.

Sorry. Ignore my comment. Got my wires crossed.

Issue #1667, opened 2024-03-28
Assert failure in lab_wlr_scene_output_commit()


Tamaranch:
I don't think I can reproduce this, and it's not quite with the latest commit (I was running 11b6836a3834a77bf9603a7e123fd3ee75a14b72), but I think it's best to report it if it helps.

After doing some display manipulation (scaling etc. via wlr-output-management), I switched to my X11 session in tty 7, then back to my Wayland session in tty 3 labwc crashed:
```
labwc: ../src/common/scene-helpers.c :44 : lab_wlr_scene_output_commit:  l'assertion « scene_output » a échoué.

Core was generated by `labwc --startup xfce4-session'.
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44	      return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
[Current thread is 1 (Thread 0x717a46452f80 (LWP 67473))]
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x0000717a494ab393 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x0000717a4945a6c8 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x0000717a494424b8 in __GI_abort () at abort.c:79
#4  0x0000717a494423dc in __assert_fail_base
    (fmt=0x717a4a5a7355 "%s%s%s :%u : %s%s l'assertion « %s » a échoué.\n%n", assertion=assertion@entry=0x5e5a9c40d513 "scene_output", file=file@entry=0x5e5a9c40d4f0 "../src/common/scene-helpers.c", line=line@entry=44, function=function@entry=0x5e5a9c40d5a0 <__PRETTY_FUNCTION__.0> "lab_wlr_scene_output_commit") at assert.c:94
#5  0x0000717a49452d46 in __assert_fail
[...]

PR #1693, opened 2024-04-08
output: ensure we don't run into an assert() on VT switch

Consolatis:
Also reduce log spam from failed output commits that can happen for various reasons outside of our control.

Fixes:
- #1667

---

There is also a wip branch [here](https://github.com/Consolatis/labwc/tree/issue/1667_scene_output_assert) which tries to fix this more generally but I am not feeling confident enough with that one for the current point in the dev cycle.


johanmalm:
LGTM.
Not tested.
Agree with cautious approach.

PR #1694, opened 2024-04-08
CI: enable gcc static analyzer

Consolatis:
This may help uncover potential bugs in the future.
If we end up getting too many false positives we can still disable it again.


johanmalm:
LGTM. Not tested.

Issue #1098, opened 2023-09-23
Window switcher styling not working under some conditions


stefonarch:
With condition=unknown,

All lines with "window-switcher" in `themerc` have no effect here:
```
osd.bg.color: #ffffff
osd.border.color: #3294a8
osd.border.width: 3
osd.label.text.color: #484848
osd.window‐switcher.width: 700
osd.window‐switcher.padding: 4
osd.window‐switcher.item.padding.x: 3
osd.window‐switcher.item.padding.y: 3
osd.window‐switcher.item.active.border.width: 2
```
I noticed it first on my laptop, while on the PC it was working until some day ago. 
It works with default or or no config files. Using `themerrc-override` doesn't make any difference.

stefonarch:
After lots of test and exclusion it _seems_ that the `-d `option is part of the issue.

With identical config folders using
`exec labwc -d -C $XDG_CONFIG_HOME/lxqt-wayland/labwc -s lxqt-session` the issue is present, while with `exec labwc  -s lxqt-session`  `the file `themerc-override` works as expected, but in `themerc`  settings have no effect still.

Consolatis:
> All lines with "window-switcher" in `.themerc` have no effect here:
> It works with default or or no config files. Using `themerrc.override` doesn't make any difference.

Just to ensure it was a typo while posting the report: it is `themerc-override`, e.g. `-` rather than `.` and `$theme_dir/openbox-3/themerc` rather than `.themerc`.

> exec labwc -d -C $XDG_CONFIG_HOME/lxqt-wayland/labwc -s lxqt-session the issue is present, while with exec labwc  -s lxqt-session the file themerc-override works as expected, but in themerc settings have no effect still.

Didn't look into it yet but the `-C` is also a difference here.

johanmalm:
I'll have a look post `0.6.5` bump. Suspect we don't handle `$XDG_CONFIG_HOME` properly.

stefonarch:
Fixed the typos, and it's non -d but -C (took -d for directory...).
It doesn't look that using `$XDG_CONFIG_HOME` or using full path makes a difference, it looks like just another path doesn't work here. But I like to test also it on the other installation. too many factors around.

stefonarch:
Tested on the labwc, basically the same findings:
*  with custom folder (labwc -d -C $XDG_CONFIG_HOME/lxqt-wayland/labwc) no way to style the window switcher
[...]

PR #1702, opened 2024-04-11
 overlay: make overlay's style and colors configurable

tokyo4j:
This PR adds following theme settings:
- `snapping-preview.region.area.color`
- `snapping-preview.edge.area.color`
- `snapping-preview.region.rect-type` (`Auto|Area|Outlines`)
- `snapping-preview.edge.rect-type` (`Auto|Area|Outlines`)

I named the theme entry `rect-type` to make `labwc-theme.5.scd` easy to follow, but maybe it should be just `type`.

I restructured `struct overlay` to allow independent styles for region/edge snapping overlay.
I also added `inactivate_overlay()` as a refactoring.

In the future, maybe we can also make the colors of outlines configurable like:
`snapping-preview.region.outlines.colors: #ffffff,#000000,#ffffff`
In that case, I think the colors of OSD preview should also be configurable.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -205,6 +205,34 @@ elements are not listed here, but are supported.
 	Otherwise, an outlined rectangle is shown instead.
 	If software rendering is used, an outlined rectangle is always shown.
 
+*snapping.preview.region.bg.color*
+	Color of a filled rectangle shown as the preview when a window is snapped to
+	a region. Default is #8080b380.
+
+*snapping.preview.edge.bg.color*
+	Color of a filled rectangle shown as the preview when a window is snapped to
+	an edge. Default is #8080b380.
+
+*snapping.preview.region.border.width*
+	Border width of an outlined rectangle shown as the preview when a window is
+	snapped to a region. Inherits `osd.border.width` if not set.
+
+*snapping.preview.edge.border.width*
+	Border width of an outlined rectangle shown as the preview when a window is
+	snapped to an edge. Inherits `osd.border.width` if not set.
+
+*snapping.preview.region.border.color*
+	Color(s) of an outlined rectangle shown as the preview when a window is
+	snapped to a region. Possible value is a color or up to 3 colors separated
+	by commas (e.g. "#ffffff,#000000,#ffffff"). When multiple colors are
+	specified, a multi-line rectangle with each line having the specified color
+	is drawn. If not set, this inherits the on-screen-display theme
+	("[*osd.bg.color*],[*osd.label.text.color*],[*osd.bg.color*]").
+
+*snapping.preview.edge.border.color*
+	Color(s) of an outlined rectangle shown as the preview when a window is
+	snapped to an edge. See *snapping.preview.region.border.color* for details.
+
 *border.color*
 	Set all border colors. This is obsolete, but supported for backward
 	compatibility as some themes still contain it.
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -195,6 +195,16 @@ elements are not listed here, but are supported.
 	Height of boxes in workspace switcher in pixels. Setting to 0 disables
 	boxes. Default is 20.
 
+*snapping.preview.region.fill* [yes|no]
+	Show a filled rectangle as the preview when a window is snapped to a region.
+	Otherwise, an outlined rectangle is shown instead.
+	If software rendering is used, an outlined rectangle is always shown.
+
+*snapping.preview.edge.fill* [yes|no]
+	Show a filled rectangle as the preview when a window is snapped to an edge.
+	Otherwise, an outlined rectangle is shown instead.
+	If software rendering is used, an outlined rectangle is always shown.
+
 *border.color*
 	Set all border colors. This is obsolete, but supported for backward
 	compatibility as some themes still contain it.
--- a/docs/themerc
+++ b/docs/themerc
@@ -71,3 +71,6 @@ osd.window-switcher.item.active.border.width: 2
 
 osd.workspace-switcher.boxes.width: 20
 osd.workspace-switcher.boxes.height: 20
+
+snapping.preview.region.fill: yes
+snapping.preview.edge.fill: yes
--- a/docs/themerc
+++ b/docs/themerc
@@ -74,3 +74,9 @@ osd.workspace-switcher.boxes.height: 20
 
 snapping.preview.region.fill: yes
 snapping.preview.edge.fill: yes
+snapping.preview.region.bg.color: #8080b380
+snapping.preview.edge.bg.color: #8080b380
+snapping.preview.region.border.width: 1
+snapping.preview.edge.border.width: 1
+snapping.preview.region.border.color: #dddda6,#000000,#dddda6
+snapping.preview.edge.border.color: #dddda6,#000000,#dddda6
--- a/include/overlay.h
+++ b/include/overlay.h
@@ -7,12 +7,19 @@
 #include "regions.h"
 #include "view.h"
 
-struct overlay {
-	struct wlr_scene_tree *tree;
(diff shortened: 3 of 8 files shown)


johanmalm:
Thanks. Looks good in general.

>     * `snapping-preview.region.area.color`
>     * `snapping-preview.edge.area.color`

For consistency with the rest of the `themerc` file I'd suggest `bg` instead of `area`.

`snapping-preview.bg.color`
`snapping-preview.border.color`

>     * `snapping-preview.region.rect-type` (`Auto|Area|Outlines`)
>     * `snapping-preview.edge.rect-type` (`Auto|Area|Outlines`)

> I named the theme entry `rect-type` to make `labwc-theme.5.scd` easy to follow, but maybe it should be just `type`.

I'm fine with `rect-type`. If you want to consider something else, we could go with `fill` (yes|no|auto). Fill is consistent with cairo-speak (fill/stroke).

For reference, `rc.xml` uses `<resize><drawContents>` - although that feels slightly different.

> In the future, maybe we can also make the colors of outlines configurable like: `snapping-preview.region.outlines.colors: #ffffff,#000000,#ffffff` In that case, I think the colors of OSD preview should also be configurable.

Good idea. Agree. Suggest calling it `border` rather than `outlines` for consistency.

tokyo4j:
I implemented @johanmalm's suggestions, as they make sense to me. `fill` sounds straightforward rather than defining `rect-type` type, as long as we only support `wlr_scene_rect` and `multi_rect`. If we may support another rectangle style in the future, that should be reverted.

I also implemented `snapping-preview.[region|edge].border.color`, whose possible values are two colors separated by a comma like `#ffffff,#000000` or `auto` keyword which falls back to OSD's background/text colors.

I haven't made OSD preview color configurable yet as it's not in the scope of this PR, but it should be an easy fix.

tokyo4j:
I reorganized my commits and some made changes in `snapping-preview.[region|edge].border.color`.

`snapping-preview.[region|edge].border.color` now allows specifying 1-3 colors, and only single line rectangle is drawn when only one color is specified.
Its default value is now set to `#ffffff,#000000,#ffffff` rather than falling back to OSD's colors.

I also added `snapping-preview.[region|edge].border.width`.

docs/labwc-theme.5.scd
```diff
@@ -195,6 +195,52 @@ elements are not listed here, but are supported.
 	Height of boxes in workspace switcher in pixels. Setting to 0 disables
 	boxes. Default is 20.
 
+*snapping-preview.region.fill* [yes|no|auto]
+	Show filled rectangle as preview when window is snapped to a region.
+	Otherwise, outlined rectangle is shown instead.
```
Consolatis, line 200: `Otherwise, an (a?) outlined rectangle is shown instead`.

docs/labwc-theme.5.scd
```diff
@@ -195,6 +195,52 @@ elements are not listed here, but are supported.
...
+
+*snapping-preview.edge.fill* [yes|no|auto]
+	Show filled rectangle as preview when window is snapped to an edge.
+	See *snapping-preview.region.fill* above for details.
+	Default is *auto*.
+
+*snapping-preview.region.bg.color*
+	Color of filled rectangle shown as preview when window is snapped to a
+	region.
+	Default is #8080b380.
```
Consolatis, line 214: Default can be in the previous line, same for the following 3 entries. (I think it looks better when the the "Default is" line is shorter than the line above, in all other cases it should be in the previous line instead).

docs/labwc-theme.5.scd
```diff
@@ -195,6 +195,52 @@ elements are not listed here, but are supported.
...
+
+*snapping-preview.edge.border.width*
+	Border width of outlined rectangle shown as preview when window is snapped to
+	an edge.
+	Default is 1.
+
+*snapping-preview.region.border.color*
+	Color(s) of outlined rectangle shown as preview when window is snapped to a
+	region.
+	Possible value is a color or up to 3 colors separated by a comma
```
Consolatis, line 233: Similar thing here, I think the line break is not necessary to split the description and the value definition. Same for `snapping-preview.edge.border.color`.

docs/themerc
```diff
@@ -71,3 +71,12 @@ osd.window-switcher.item.active.border.width: 2
...
 osd.workspace-switcher.boxes.height: 20
+
+snapping-preview.region.fill: auto
+snapping-preview.edge.fill: auto
+snapping-preview.region.bg.color: #8080b380
+snapping-preview.edge.bg.color: #8080b380
+snapping-preview.region.border.width: 1
+snapping-preview.edge.border.width: 1
+snapping-preview.region.border.color: #ffffff,#000000,#ffffff
+snapping-preview.edge.border.color: #ffffff,#000000,#ffffff
```
Consolatis, line 82: Do we want to change the default to white-black-white here?
Using the OSD colors instead as default will somewhat match the theme in use and should also provide a high enough contrast ratio (as otherwise the text on the OSD wouldn't be readable).

tokyo4j: That's right.
I didn't know that some OSD styles already falls back to values like `window.active.border.color` when I made this change, and thought that I need to allow any dedicated value like `auto` to make it fall back, which sounds tedious.

I will fix it soon.

tokyo4j:
Changes are:
- Adding a/an/the and minor fixes in the documentation
- Letting `snapping-preview.*.border.width` and `snapping-preview.*.border.color` fall back to OSD theme and fixing its documentation

tokyo4j:
I'm thinking of forcing outlines to be rendered when `WLR_RENDERER=pixman`.
I tried `snapping-preview.*.fill: yes` with `WLR_RENDERER=pixman` on my Ryzen 3 4300U laptop, and found it lags so badly (300ms~600ms in my sense).

So possible changes would be like:
- Possible values in `snapping-preview.*.fill` are just `yes|no`
- The default value for `snapping-preview.*.fill` is `yes`
- If the current rendering backend is pixman, outlines are always chosen

Consolatis:
Had a look through the diff again and didn't see anything obvious. 

- tested `Reconfigure` while a preview was shown: it just hides and everything works as expected thereafter
- tested `xcalc &; sleep 5; kill $(pidof xcalc)` and making xcalc show a preview: works as expected
- tested setting border width
- tested setting `.fill` to `no` on the gles2 renderer
- tested colors for border and area

Only slight nitpick I can think of:
`snapping-preview.` could be replaced by `snapping.preview.` (we had issues in the past where the `-` character used was actually some weird locale depended unicode character and not the ASCII version thus we failed to parse it correctly).

Other than that this looks great!
After addressing the `snapping-preview` vs `snapping.preview` point, do you want to squash yourself or should I do so when merging?

tokyo4j:
Thank you!

I'm fine with `snapping.preview`.

Honestly, I don't think the unicode issue (#1098?) can be a strong reason to choose it over `snapping-preview`, but `snapping.preview` looks more natural as there's `<snapping><preview>` in `rc.xml` so I like it better.

>  After addressing the snapping-preview vs snapping.preview point, do you want to squash yourself or should I do so when merging?

I'll do it myself.

Consolatis:
The last Last commit contains `snapping-overlay` in the commit body and the subject is 75 characters wide, suggest something like `overlay: add theme settings for colors and border width` instead. The commit body of the 2nd to last commit is slightly outdated.

tokyo4j:
Thank you! I've learned so much from you about how to contribute to OSS!

Consolatis:
Well, thank you for your great code and responsiveness to feedback. Really appreciated :)

I think the whole snapping overlay implementation is really nice out of the box but still flexible in case a user wants to change the behavior.

tokyo4j:
I'll also make the outline colors of window-switcher preview configurable in follow-up PRs.
It'll more likely to be worth changing from the default as its default colors are similar to that of window borders, which makes a bit harder to distinguish currently previewed window from others.

PR #1706, opened 2024-04-12
input: notify idle manager when emulating cursor move

jp7677:
Looks like we forgot that one earlier, contrary to emulating cursor buttons (https://github.com/labwc/labwc/blob/master/src/input/cursor.c#L1120) 

From https://github.com/labwc/labwc/issues/1672#issuecomment-2052534441 , found by @DynamoFox


Consolatis:
Thanks :)

Issue #1618, opened 2024-03-13
Only create the headless backend if there isn't one already


Consolatis:
This may happen when starting labwc with `WLR_BACKENDS=headless`.

PR #1708, opened 2024-04-13
server: don't create headless backend twice

Consolatis:
This may happen when running with WLR_BACKENDS=headless.

Fixes:
- #1618


johanmalm:
Thanks.

Issue #1298, opened 2023-12-08
Should "Execute an action when ForEach failed to find a match #12" be implemented in labwc? (Mikachu/openbox)


ludg1e:
Hello,

While I assume that labwc doesn't aim by any mean to become an Openbox clone, I think this may be a good addition to the ForEach action.
Issue in Openbox mirror repo: https://github.com/Mikachu/openbox/issues/12

It basically allows to run an action (else) in case the ForEach action has failed (tried to explain).
The user even implemented it by itself so the code for labwc could even be inspired (in some way) by his code (https://github.com/tnymlr/openbox/commit/8f45daa0bc1ba2e6ae6112a589a5ad22efed06e1).

Thank you and have a nice day

johanmalm:
Happy to treat this as in-scope.

Did you want to try the patch?

I think action.c:924 is good starting point.
Maybe make run_if_action() return the number of "then" actions run... if zero, do the "none" actions.

PR #1710, opened 2024-04-13
add none option to ForEach action

nicolas3121:
adds a none branch to ForEach whose actions will be ran when no matches were found
from issue  #1298

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -515,7 +515,7 @@ action_is_valid(struct action *action)
 	case ACTION_TYPE_IF:
 	case ACTION_TYPE_FOR_EACH:
 		; /* works around "a label can only be part of a statement" */
-		static const char * const branches[] = { "then", "else" };
+		static const char * const branches[] = { "then", "else", "none" };
 		for (size_t i = 0; i < ARRAY_SIZE(branches); i++) {
 			struct wl_list *children = action_get_actionlist(action, branches[i]);
 			if (children && !action_list_is_valid(children)) {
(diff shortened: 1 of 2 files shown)


Consolatis (commented):
Thanks for the PR!

Some minor comments below. I didn't yet had a chance to test and look in more detail into the logic, might take a few days to do so.

Consolatis:
Tested with
```xml
<keybind key="F12">
	<!-- run or raise specific application -->
	<action name="ForEach">
		<query identifier="run_or_raise" />
		<then>
			<action name="Raise" />
			<action name="Focus" />
		</then>
		<none>
			<action name="Execute" command="foot -a run_or_raise -o colors.alpha=0.8" />
		</none>
	</action>
</keybind>
[..]
<windowRule identifier="run_or_raise" serverDecoration="no" skipWindowSwitcher="yes" skipTaskbar="yes">
	<action name="SnapToEdge" direction="up" />
</windowRule>
```

Works great!

Consolatis:
> Hi do you know if one can make a scratchpad sort of thing with this? (like just a single keybinding designated to launching/raising/minimizing a particular terminal window (like the one named run_or_raise in your case). I tried but couldn't figure it out how to include the "minimize if it's raised and focused" and "raise and focus if it exists and is minimized" parts. 😅

I tried that one as well but so far wasn't able to make something work either. I did add a `state` query option for testing (supporting `focused` and `minimized`) but as `ForEach` and `If` can not be nested and there is no `return` or `ElseIf` or state variables I didn't manage to make it work using only a single keybind. It would either minimize the terminal and then raise it again or the other way around. I guess there could be some really ugly hack by using wtype to insert key presses to execute another keybind but I would not call that a satisfying solution (and I am not even sure that it wouldn't cause issues).

PR #1712, opened 2024-04-14
Small fixes in `doc/themerc` and `docs/labwc-theme.5.scd`

tokyo4j:
- The default value of `osd.bg.color` is `#e1dedb`, as it falls back to `window.active.title.bg.color` (I may be missing some context here)
- "all border colors" in the description of `border.color` sounds like it includes `osd.border.color`

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -155,16 +155,19 @@ elements are not listed here, but are supported.
 	Menu separator color. Default #888888.
 
 *osd.bg.color*
-	Background color of on-screen-display.
+	Background color of on-screen-display. Inherits
+	*window.active.title.bg.color* if not set.
 
 *osd.border.color*
-	Border color of on-screen-display.
+	Border color of on-screen-display. Inherits *osd.label.text.color* if not
+	set.
 
 *osd.border.width*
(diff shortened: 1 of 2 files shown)


docs/labwc-theme.5.scd
```diff
@@ -196,8 +196,9 @@ elements are not listed here, but are supported.
 	boxes. Default is 20.
 
 *border.color*
-	Set all border colors. This is obsolete, but supported for backward
-	compatibility as some themes still contain it.
+	Set both *window.active.border.color* and *window.active.border.color*.
```
Consolatis, line 199: I assume one of them should read `window.inactive.border.color`.

tokyo4j: Fixed.

tokyo4j:
I added missing documentations on fallback values of `osd.bg.color`, `osd.border.color` and `osd.label.text.color`.

johanmalm:
Thanks. Great tidy-ups. I think some of those go right back to the very, very early days of `labwc`.

It's possible that at the "early" time `border.color` did set "all" borders (that existed at the time) :smile:

And I possibly wasn't so careful with documentation either!

PR #1713, opened 2024-04-14
common: remove buf_init(), add BUF_INIT and buf_move()

jlindgren90:
Rework of #1629 incorporating some of the ideas mentioned in discussion there.

Also includes the first commit from #1629 as-is (thanks @Consolatis!) which looks like a simple fix and could be merged separately.

### DIFF
--- a/include/common/buf.h
+++ b/include/common/buf.h
@@ -8,15 +8,27 @@
 #ifndef LABWC_BUF_H
 #define LABWC_BUF_H
 
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-
 struct buf {
+	/**
+	 * Pointer to underlying string buffer. If alloc != 0, then
+	 * this was allocated via malloc().
+	 */
 	char *buf;
+	/**
+	 * Allocated length of buf. If zero, buf was not allocated
+	 * (either NULL or literal empty string).
+	 */
 	int alloc;
+	/**
+	 * Length of string contents (not including terminating NUL).
+	 * Currently this must be zero if alloc is zero (i.e. non-empty
+	 * literal strings are not allowed).
+	 */
 	int len;
 };
+
+/** Value used to initialize a struct buf to an empty string */
+#define BUF_INIT ((struct buf){.buf = ""})
 
 /**
  * buf_expand_tilde - expand ~ in buffer
@@ -30,13 +42,6 @@ void buf_expand_tilde(struct buf *s);
  * Note: $$ is not handled
  */
 void buf_expand_shell_variables(struct buf *s);
-
-/**
- * buf_init - allocate NULL-terminated C string buffer
- * @s: buffer
- * Note: use free(s->buf) to free it
- */
-void buf_init(struct buf *s);
 
 /**
  * buf_add - add data to C string buffer
@@ -55,13 +60,35 @@ void buf_add_char(struct buf *s, char data);
 /**
  * buf_clear - clear the buffer, internal allocations are preserved
  * @s: buffer
+ *
+ * The buffer will be set to a NUL-terminated empty string.
+ *
+ * This is the appropriate function to call to re-use the buffer
+ * in a loop or similar situations as it reuses the existing heap
+ * allocation.
  */
 void buf_clear(struct buf *s);
 
 /**
  * buf_reset - reset the buffer, internal allocations are free'd
  * @s: buffer
+ *
+ * The buffer will be re-initialized to BUF_INIT (empty string).
+ *
+ * Inside a loop, consider using buf_clear() instead, as it allows
+ * reusing the existing heap allocation. buf_reset() should still be
+ * called after exiting the loop.
  */
 void buf_reset(struct buf *s);
+
+/**
+ * buf_move - move the contents of src to dst, freeing any previous
+ * allocation of dst and resetting src to BUF_INIT.
+ *
+ * dst must either have been initialized with BUF_INIT
+ * or zeroed out (e.g. created by znew() or on the stack
+ * with something like struct buf foo = {0}).
+ */
+void buf_move(struct buf *dst, struct buf *src);
 
 #endif /* LABWC_BUF_H */
--- a/include/common/buf.h
+++ b/include/common/buf.h
@@ -13,9 +13,9 @@ struct buf {
 	 * Pointer to underlying string buffer. If alloc != 0, then
 	 * this was allocated via malloc().
 	 */
-	char *buf;
+	char *data;
 	/**
-	 * Allocated length of buf. If zero, buf was not allocated
+	 * Allocated length of buf. If zero, data was not allocated
 	 * (either NULL or literal empty string).
 	 */
 	int alloc;
@@ -28,7 +28,7 @@ struct buf {
 };
 
 /** Value used to initialize a struct buf to an empty string */
-#define BUF_INIT ((struct buf){.buf = ""})
+#define BUF_INIT ((struct buf){.data = ""})
 
 /**
  * buf_expand_tilde - expand ~ in buffer
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -2,14 +2,15 @@
 #include <assert.h>
 #include <ctype.h>
 #include <stdbool.h>
+#include <string.h>
 #include "common/buf.h"
+#include "common/macros.h"
 #include "common/mem.h"
 
 void
 buf_expand_tilde(struct buf *s)
 {
-	struct buf new;
-	buf_init(&new);
+	struct buf new = BUF_INIT;
 	for (int i = 0 ; i < s->len ; i++) {
 		if (s->buf[i] == '~') {
 			buf_add(&new, getenv("HOME"));
@@ -17,10 +18,7 @@ buf_expand_tilde(struct buf *s)
 			buf_add_char(&new, s->buf[i]);
 		}
 	}
-	free(s->buf);
-	s->buf = new.buf;
-	s->len = new.len;
-	s->alloc = new.alloc;
+	buf_move(s, &new);
 }
 
 static void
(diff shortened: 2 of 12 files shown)


src/common/buf.c
```diff
@@ -69,37 +65,38 @@ buf_expand_shell_variables(struct buf *s)
...
 }
 
 void
 buf_add(struct buf *s, const char *data)
 {
-	assert(s->buf);
-
 	if (!data || data[0] == '\0') {
 		return;
 	}
```
Consolatis, line 96: Could use `string_null_or_empty()` via `include/common/string-helpers.c`

src/common/buf.c
```diff
@@ -69,37 +65,38 @@ buf_expand_shell_variables(struct buf *s)
...
 	int len = strlen(data);
-	if (s->alloc <= s->len + len + 1) {
-		s->alloc = s->alloc + len;
-		s->buf = xrealloc(s->buf, s->alloc);
-	}
+	buf_expand(s, s->len + len + 1);
+	assert(s->alloc);  // dumb but makes gcc static analyzer happy
```
Consolatis, line 99: I wonder why this is necessary here but not in the `buf_add_char()` case. Maybe the analyzer treats `memcpy()` different to direct assignment. Also all of this are ints so technically `s->alloc` (and `s->len`) could be negative due to overflow but it doesn't seem to bother the analyzer in `buf_add_char()`.

include/common/buf.h
```diff
@@ -55,13 +60,35 @@ void buf_add_char(struct buf *s, char data);
...
 
+/**
+ * buf_move - move the contents of src to dst, freeing any previous
+ * allocation of dst and resetting src to BUF_INIT.
+ *
+ * dst must either have been initialized with BUF_INIT
+ * or zeroed out (e.g. created by znew() or on the stack
+ * with something like struct buf foo = {0}).
+ */
+void buf_move(struct buf *dst, struct buf *src);
```
Consolatis, line 92: Should this be public?

include/common/buf.h
```diff
@@ -55,13 +60,35 @@ void buf_add_char(struct buf *s, char data);
...
  * buf_reset - reset the buffer, internal allocations are free'd
  * @s: buffer
+ *
+ * The buffer will be re-initialized to BUF_INIT (empty string).
+ *
+ * Inside a loop, consider using buf_clear() instead, as it allows
+ * reusing the existing heap allocation. buf_reset() should still be
+ * called after exiting the loop.
  */
 void buf_reset(struct buf *s);
```
Consolatis, line 82: While I think what `buf_reset()` is now doing is good I am not sure about using `xxx_reset()` to free resources really matches the rest of the codebase. We usually use `_destroy()` instead. Maybe we could just add another wrapper?

src/common/buf.c
```diff
@@ -2,25 +2,23 @@
...
 #include "common/buf.h"
+#include "common/macros.h"
 #include "common/mem.h"
 
 void
 buf_expand_tilde(struct buf *s)
 {
-	struct buf new;
-	buf_init(&new);
+	struct buf new = BUF_INIT;
```
Consolatis, line 13: This feels a bit weird, what about `struct buf new = BUF_INIT()` instead?

include/common/buf.h
jlindgren90: In my mind `reset` suggests "free resources and set object back to initial state", i.e. it's a combined `destroy` + `init`. This has precedent outside labwc, for example (C++ again) `std::unique_ptr::reset`.

Consolatis:
In general I like this. Some comments above. Also suggest amending the subject of the first commit to something like `src/osd.c: properly clear the buffer`, it is not really treewide anymore.

src/common/buf.c
jlindgren90: I don't have a strong preference, but the parentheses seem unnecessary to me since it's just a predefined value, not a functional macro.

src/common/buf.c
jlindgren90: I think it's because of the `buf_clear` + `buf_add` sequence used in `buf_expand_shell_variables`.

Modifying the first `if` in `buf_expand` might also prove to the analyzer that `s->buf` is allocated after return:
```
if (s->alloc && new_alloc <= s->alloc) {
	return;
}
```

include/common/buf.h
jlindgren90: No strong preference either way. It's not used outside `buf.c` right now, but I can imagine it being useful in future.

jlindgren90:
> Also suggest amending the subject of the first commit to something like `src/osd.c: properly clear the buffer`, it is not really treewide anymore.

Hmm, the first commit does still have a fix in `buf.c` also.

Consolatis:
> > Also suggest amending the subject of the first commit to something like `src/osd.c: properly clear the buffer`, it is not really treewide anymore.
> 
> Hmm, the first commit does still have a fix in `buf.c` also.

Whoops, disregard my comment.

src/common/buf.c
Consolatis: A define "returning" a struct feels different to a macro returning a struct.

include/common/buf.h
Consolatis: Right, and IMHO we should keep the `buf_reset()` call. It is being forced to use it to free something that feels a bit strange to me.

include/common/buf.h
Consolatis: No strong preference here either.

include/common/buf.h
jlindgren90: I think if the `struct buf` itself was allocated then `destroy` would make more sense. Since the `struct buf` itself is not freed, `reset` emphasizes that it's left in a valid state (doesn't contain a stale `buf` pointer etc.)

src/common/buf.c
jlindgren90: I'm not following. It's not "returning" anything, I don't think? It's just a literal struct value.

Consolatis: Yes,but usually a define defines a simple value (some number or string) and not a struct initializer.

The way I see it: if you imagine there is no `#define` involved at all you wouldn't see code like this, e.g. initializing a struct (not a struct pointer like `struct foo *bar`) by assigning it a single variable. Instead you would have a function that returns a struct which is then copied over to the new instance.

To mimic that usual behavior I think having `BUF_INIT()` here would be way easier to read, even without knowing what the `BUF_INIT` define / macro is actually doing.

jlindgren90: Code assigning/initializing a whole struct at once is legal in C even if uncommon -- it's more common in other languages maybe.

Anyway this seems like a minor detail. It's not a big deal to me either way.

johanmalm:
I like this.

My thoughts are that we do this for two reasons:

1. It's simpler to write `struct buf foo = BUF_INIT` than the current two lines `struct buf foo; buf_init(&foo);
2. It enables usage of foo.buf without NULL checking

If I'm thinking about this right, the advantage of the git approach (with `extern char strbuf_slopbuf[];` is that it could potentially be freed in a lower frame and re-gain the `.strbuf_slopbuf` state, whereas we couldn't assign `""` in a lower frame because it would be lost.

src/common/buf.c
```diff
@@ -108,28 +105,38 @@ buf_add(struct buf *s, const char *data)
...
 
 void
 buf_reset(struct buf *s)
 {
-	zfree(s->buf);
-	buf_init(s);
+	if (s->alloc) {
+		free(s->buf);
+	}
+	*s = BUF_INIT;
```
johanmalm, line 131: Can we do this here? Would the `""` not be lost?

include/common/buf.h
johanmalm: I'm good with `reset`. For comparison `git` uses `release`. 

https://github.com/git/git/blob/8f7582d995682f785e80e344197cc715e6bc7d8e/strbuf.c#L75

include/common/buf.h
johanmalm: Public seems fine.

include/common/buf.h
```diff
@@ -8,16 +8,28 @@
...
 
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-
 struct buf {
+	/**
+	 * Pointer to underlying string buffer. If alloc != 0, then
+	 * this was allocated via malloc().
+	 */
```
johanmalm, line 15: I know this is "original" - but we could consider `s/buf/data/` here. Seems consistent with other string implementations, including that of glib. Also avoid `buf->buf` which looks a big ugly.

src/common/buf.c
jlindgren90: String literals have static storage duration according to C99. Does that address the concern?

johanmalm: I see what you mean. It just feels wrong.

include/common/buf.h
jlindgren90: Renamed to `data`.

src/common/buf.c
jlindgren90: Done.

src/common/buf.c
jlindgren90: I went with this approach since it seems more logical to me, and added a longer comment.

johanmalm: Happy with this.

include/common/buf.h
johanmalm: Fine.

src/common/buf.c
johanmalm: I prefer `BUF_INIT` and it similar to `git`'s approach.

src/common/buf.c
johanmalm: You know, I don't think I would have passed a technical interview question on string literals having static storage, but now that the penny has dropped I like it. Makes sense.
I'm not sure I've ever used this approach as a design pattern though - so that's quite exciting in itself.
Yes, agree it is particularly clean just doing `""`.
Let's go for it.

johanmalm:
LGTM.

I think it's better to keep changes as separate commits for this, but did you want to interactive-rebase the "dumb assert" and the final commit?

jlindgren90:
> LGTM.
> 
> I think it's better to keep changes as separate commits for this, but did you want to interactive-rebase the "dumb assert" and the final commit?

Thanks. I squashed the analyzer fix commits into the 2nd commit (which triggered the analyzer warning) but left the others separate. This should be good to go now.

johanmalm:
Thanks.

PR #1714, opened 2024-04-15
graphic-helpers: prevent multi-rect edges from overlapping at corner

tokyo4j:
This is a follow-up to #1702

Before this change:
screenshot


Consolatis:
Nice find! Thanks :)

tokyo4j:
I just wanted to clarify this should be tested after #1702

Consolatis:
> I just wanted to clarify this should be tested after #1702

I did rebase this on top of master before testing, everything was working as expected :)

PR #1715, opened 2024-04-15
osd: add theme settings for window-switcher preview

tokyo4j:
Adds theme settings like:

```
osd.window-switcher.preview.border.width: 2
osd.window-switcher.preview.border.color: #ffffff,#00a2ff,#ffffff
```

20240416_02h56m34s_grim

20240416_02h56m53s_grim


Consolatis:
Thanks :)

PR #1716, opened 2024-04-15
Add labwc to XDG_CURRENT_DESKTOP to support a portals.conf

rcalixte:
The previous `UseIn` key was deprecated in xdg-desktop-portal 1.17/1.18. It has been superceded by the portals.conf structure so that each desktop can configure the precise desired structure for portals. In addition, support was added to the Desktop Entry Specifications to support a `DesktopNames` key that login managers will use to set XDG_CURRENT_DESKTOP.

* [portals.conf Documentation](https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals.conf.rst.in)
* [Example sway-portals.conf](https://salsa.debian.org/swaywm-team/sway/-/blob/debian/sid/debian/sway-portals.conf)
* [Desktop Entry Specifications](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)

Ref: flatpak/xdg-desktop-portal#955

### DIFF
--- a/data/labwc-portals.conf
+++ b/data/labwc-portals.conf
@@ -1,0 +1,2 @@
+[preferred]
+default=wlr;*
--- a/data/labwc.desktop
+++ b/data/labwc.desktop
@@ -4,4 +4,4 @@ Comment=A wayland stacking compositor
 Exec=labwc
 Icon=labwc
 Type=Application
-DesktopNames=wlroots
+DesktopNames=labwc;wlroots
--- a/docs/environment
+++ b/docs/environment
@@ -60,13 +60,13 @@
 
 ##
 ## This allows xdg-desktop-portal-wlr to function (e.g. for screen-recording).
-## It is automatically set to "wlroots" by labwc though, so it is only
+## It is automatically set to "labwc:wlroots" by labwc though, so it is only
 ## included here for completeness. Again, labwc will not over-write an
 ## already set environment variable, so if you need it set to something else,
 ## then uncomment and adjust.
 ##
 
-# XDG_CURRENT_DESKTOP=wlroots
+# XDG_CURRENT_DESKTOP=labwc:wlroots
 
 ##
 ## This causes a virtual output to be created automatically whenever there
--- a/meson.build
+++ b/meson.build
@@ -140,6 +140,8 @@ executable(
 
 install_data('data/labwc.desktop', install_dir: get_option('datadir') / 'wayland-sessions')
 
+install_data('data/labwc-portals.conf', install_dir: get_option('datadir') / 'xdg-desktop-portal')
+
 icons = ['labwc-symbolic.svg', 'labwc.svg']
 foreach icon : icons
   icon_path = join_paths('data', icon)
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -223,7 +223,7 @@ session_environment_init(void)
 	 * May be overridden either by already having a value set or by the user
 	 * supplied environment file.
 	 */
-	setenv("XDG_CURRENT_DESKTOP", "wlroots", 0);
+	setenv("XDG_CURRENT_DESKTOP", "labwc:wlroots", 0);
 
 	/*
 	 * Set default for _JAVA_AWT_WM_NONREPARENTING so that Java applications


Consolatis:
https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals.conf.rst.in#L136C1-L140C32

describes the content of the `XDG_CURRENT_DESKTOP` env var as:
> A colon-separated list of desktop environments, most specific first

However, in this PR the env var is using a semi-colon `;` rather than a colon `:` to separate `labwc` and `wlroots`.

rcalixte:
> https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals.conf.rst.in#L136C1-L140C32
> 
> describes the content of the `XDG_CURRENT_DESKTOP` env var as:
> 
> > A colon-separated list of desktop environments, most specific first
> 
> However, in this PR the env var is using a semi-colon `;` rather than a colon `:` to separate `labwc` and `wlroots`.

I'd seen that and I'm not sure what it pertains to, perhaps some environment variable in the portals.conf? (I've not seen this convention in use anywhere yet.)

From the [Desktop Entry Specifications](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html):
>  `$XDG_CURRENT_DESKTOP` should have been set by the login manager, according to the value of the `DesktopNames` found in the session file. The entry in the session file has multiple values separated in the usual way: with a semicolon.

rcalixte:
> This is kind of messy.

We're used to this by now. 😄

Consolatis:
> From the [Desktop Entry Specifications](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html):
> 
> > `$XDG_CURRENT_DESKTOP` should have been set by the login manager, according to the value of the `DesktopNames` found in the session file. The entry in the session file has multiple values separated in the usual way: with a semicolon.

It also states directly before that quote:
> If $XDG_CURRENT_DESKTOP is set then it contains a colon-separated list of strings.

So using colon `:` separated content of the env var seems correct (and thus should be fixed in this PR).

rcalixte:
> With the latest changes the `.desktop` file now seems wrong, at least compared to how I understand the linked documentation.

So it's the same value. Depending on your login manager, you can test this by adding the `DesktopNames` key to your .desktop file before logging in. So these should match, whether a colon or semi-colon. Let me see if I can find other examples in the wild.

Consolatis:
The sentence you quoted clearly stated that the `.desktop` file uses semi-colons to separate the items. Random anecdote [here](https://github.com/flatpak/xdg-desktop-portal/pull/955#issuecomment-1705301396) which also uses a trailing `;`.

rcalixte:
It's the same value. I don't believe there should a semi-colon in one place and a colon in the other.

Here's an example from Budgie's [budgies-desktop.desktop](https://github.com/BuddiesOfBudgie/budgie-desktop/blob/bfab2ff/src/session/budgie-desktop.desktop.in.in#L8).

rcalixte:
This value directly feeds into the environment variable.

Consolatis:
I usually prefer to implement something based on specs rather than look at what other people do.
The `budgie-desktop.desktop` actually uses a semi-colon as well.

data/labwc-portals.conf
```diff
@@ -0,0 +1,2 @@
+[preferred]
+default=wlr;gtk;
```
Consolatis, line 2: What if a user only has the kde portals installed? Wouldn't this completely break all portals for that user?
It seems there is a `*` default as well: https://github.com/flatpak/xdg-desktop-portal/blob/main/doc/portals.conf.rst.in#L117

So I assume a better default would be `default=wlr;*` here? (recognizing that labwc is not a desktop environment and does not prefer one portal implementation over another, at least outside of the -wlr portal interfaces).

rcalixte: I thought about this and chose poorly. Apologies.

Consolatis:
As I personally don't use desktop portals it would be great if somebody could test this PR and verify that screensharing still works and it doesn't change anything regarding the used portals (e.g. kde ones rather than gtk or the other way around).

stefonarch:
For LXQt I'm using `XDG_CURRENT_DESKTOP=LXQt:labwc:wlroots`  and according to the [testpage](https://mozilla.github.io/webrtc-landing/gum_test.html) screensharing should work.

jp7677:
Not sure if this is relevant here, I've also added `org.freedesktop.impl.portal.Inhibit=none`, see`https://github.com/labwc/labwc/discussions/1503 for some background.

Consolatis:
> For LXQt I'm using `XDG_CURRENT_DESKTOP=LXQt:labwc:wlroots` and according to the [testpage](https://mozilla.github.io/webrtc-landing/gum_test.html) screensharing should work.

Is that with
```ini
[preferred]
default=wlr;*
```
stored at `/usr/share/xdg-desktop-portal/labwc-portals.conf` ?

Although I guess in your case `/usr/share/xdg-desktop-portal/lxqt-portals.conf` would get priority.

johanmalm:
It's not clear to me what a sensible way ahead is here - and it's not something I use myself or have researched, so struggle to take a view without spending time on it. On the one hand I'm keen that labwc just works out of the box, but if distros/DEs have different wishes regarding xdg-desktop-portal configuration then we might be causing more hassle to shipping this.

What do other compositors do? I looked at sway, but it seems that they have chosen to "stay out of it".

Unless we're confident that this is a change-for-good, I suggest we document a suggested setup on the "getting-started" page on https://labwc.github.io

Looks like fedora do this:

```
[preferred]
--
  | # Use xdg-desktop-portal-gtk for every portal interface...
  | default=gtk
  | # ... except for the ScreenCast, Screenshot and Secret
  | org.freedesktop.impl.portal.ScreenCast=wlr
  | org.freedesktop.impl.portal.Screenshot=wlr
  | org.freedesktop.impl.portal.Secret=gnome-keyring
```

But LXQt want `XDG_CURRENT_DESKTOP=LXQt:labwc:wlroots`

Ref:
- https://src.fedoraproject.org/rpms/sway/blob/rawhide/f/sway-portals.conf

stefonarch:
> But LXQt want XDG_CURRENT_DESKTOP=LXQt:labwc:wlroots

We intend anyway  to ship custom configs for all supported compositors except for labwc and kwin (as both support launching `lxqt-session` directly) - but actually this could be changed and would be more consequent.

johanmalm:
@Vladimir-csp  I find this one really hard to evaluate. Would you be happy to offer some independent review/advice.
I'm keen that we offer a setup that works well out of the box, but don't want to step too far into Desktop Environment / Distribution space.

Vladimir-csp:
As I see the current set of changes (6998d8856500d81849d881834ba4f84e32a8834a), all seems good.

It provides sensible default to be set either by DM/SM when acting on desktop entry, or by labwc itself if variable is left empty.

User units or autostart entries can hook up to whatever item is more fit for them (more common `wlroots` or `labwc` specifically).

Alternative DEs using labwc could provide their own desktop entries or configurations, where `DesktopNames` or `XDG_CURRENT_DESKTOP` would be different.

Formats are correct. Desktop entries use ';' to delimit items, `XDG_*` vars use ':'.

`labwc` and `wlroots` are the same names I use for labwc in uwsm via plugin logic.

Consolatis:
> Not sure if this is relevant here, I've also added `org.freedesktop.impl.portal.Inhibit=none`, see`#1503 for some background.

If we really want to ship the `labwc-portals.conf` file we should indeed add that one because otherwise firefox (and potentially others) will fail to prevent screenlockers kicking in.

It would also be great of somebody could test screensharing via desktop portal using `labwc-portals.conf` from this PR when setting `XDG_CURRENT_DESKTOP=labwc:wlroots` (or even `XDG_CURRENT_DESKTOP=labwc`).

stefonarch:
Don't know how reliable this test is, it looks working with `XDG_CURRENT_DESKTOP=LXQt:labwc:wlroots`

https://mozilla.github.io/webrtc-landing/gum_test.html

johanmalm:
Tried screen recording with OBS. Works well.
Will go for merge.

PR #1717, opened 2024-04-16
session: sort directory entries in environment.d

ahesford:
Some people might benefit from reading `environment.d` in sorted order, and that's easy to achieve with `scandir`, a POSIX 2008 routine.


Consolatis:
Thanks!

Issue #1433, opened 2024-01-15
[Feature request] Directional focus


oredaze:
Openbox has the ability to focus windows in a given direction (left right up down).
The action name is `DirectionalTargetWindow`.
It is an extremely intuitive way to navigate. The lack of this feature is a complete deal breaker.

ahesford:
I wasn't aware there was a deal on the table.

Seriously, though, this might be achievable using some of the same machinery as MoveToEdge finding the next window.

johanmalm:
Interesting. I had never noticed.
Do you use it with a keybind?
Ref: http://openbox.org/wiki/Help:Actions#DirectionalTargetWindow

johanmalm:
> > Do you use it with a keybind?
> 
> Yes, what other ways are there?

Mousebind, client-menu or script with wtype.

PR #1720, opened 2024-04-16
DirectionalTargetWindow action

nicolas3121:
adds DirectionalTargetWindow action 
from issue #1433

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -70,6 +70,11 @@ Actions are used in menus and keyboard/mouse bindings.
 
 	*direction* [left|up|right|down] Direction in which to shrink.
 
+*<action name="DirectionalTargetWindow" direction="value" />*
+	 Focus jumps to closest window in given direction.
+
+	*direction* [left|up|right|down] Direction in which to shrink.	
+
 *<action name="MoveTo" x="" y="" />*
 	Move to position (x, y).
 
--- a/src/action.c
+++ b/src/action.c
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: GPL-2.0-only
-#include <limits.h>
-#include <wayland-util.h>
 #define _POSIX_C_SOURCE 200809L
 #include <assert.h>
 #include <signal.h>
 #include <string.h>
 #include <strings.h>
 #include <unistd.h>
+#include <limits.h>
+#include <wayland-util.h>
 #include <wlr/util/log.h>
 #include "action.h"
 #include "common/macros.h"
--- a/src/action.c
+++ b/src/action.c
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0-only
+#include <limits.h>
+#include <wayland-util.h>
 #define _POSIX_C_SOURCE 200809L
 #include <assert.h>
 #include <signal.h>
@@ -71,6 +73,7 @@ enum action_type {
 	ACTION_TYPE_SNAP_TO_EDGE,
 	ACTION_TYPE_GROW_TO_EDGE,
 	ACTION_TYPE_SHRINK_TO_EDGE,
+	ACTION_TYPE_DIRECTIONAL_TARGET_WINDOW,
 	ACTION_TYPE_NEXT_WINDOW,
 	ACTION_TYPE_PREVIOUS_WINDOW,
 	ACTION_TYPE_RECONFIGURE,
@@ -124,6 +127,7 @@ const char *action_names[] = {
 	"SnapToEdge",
 	"GrowToEdge",
 	"ShrinkToEdge",
+	"DirectionalTargetWindow",
 	"NextWindow",
 	"PreviousWindow",
 	"Reconfigure",
@@ -304,6 +308,7 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 		/* Falls through */
 	case ACTION_TYPE_SNAP_TO_EDGE:
 	case ACTION_TYPE_GROW_TO_EDGE:
+	case ACTION_TYPE_DIRECTIONAL_TARGET_WINDOW:
 	case ACTION_TYPE_SHRINK_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
 			enum view_edge edge = view_edge_parse(content);
@@ -492,6 +497,7 @@ action_is_valid(struct action *action)
 	case ACTION_TYPE_EXECUTE:
 		arg_name = "command";
 		break;
+	case ACTION_TYPE_DIRECTIONAL_TARGET_WINDOW:
 	case ACTION_TYPE_MOVE_TO_EDGE:
 	case ACTION_TYPE_SNAP_TO_EDGE:
 	case ACTION_TYPE_GROW_TO_EDGE:
@@ -746,6 +752,53 @@ actions_run(struct view *activator, struct server *server,
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
 				view_shrink_to_edge(view, edge);
+			}
+			break;
+		case ACTION_TYPE_DIRECTIONAL_TARGET_WINDOW:
+			if (view) {
+				/* Config parsing makes sure that direction is a valid direction */
+				struct wl_array views;
+				struct view **item;
+				enum view_edge direction = action_get_int(action, "direction", 0);
+				int distance, dx, dy;
+				int min_distance = INT_MAX;
+				struct view *nearest_view = NULL;
+				wl_array_init(&views);
+				view_array_append(server, &views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE);
+				wl_array_for_each(item, &views) {
+					if (*item == view) {
+						continue;
+					}
+					if ((*item)->minimized) {
+						continue;
+					}
+					dx = (*item)->current.x - view->current.x;
+					dy = (*item)->current.y - view->current.y;
+					switch (direction) {
+					case VIEW_EDGE_LEFT:
+						if (dx >= 0) continue;
+						break;
+					case VIEW_EDGE_RIGHT:
+						if (dx <= 0) continue;
+						break;
+					case VIEW_EDGE_DOWN:
+						if (dy <= 0) continue;
+						break;
+					case VIEW_EDGE_UP:
+						if (dy >= 0) continue;
+						break;
+					default:
+						continue;
+					}
+					distance = dx*dx + dy*dy;
+					if (distance < min_distance) {
+						min_distance = distance;
+						nearest_view = *item;
+					}
+				}
+				if (nearest_view) {
+					desktop_focus_view(nearest_view, /*raise*/ true);
+				}
 			}
 			break;
 		case ACTION_TYPE_NEXT_WINDOW:
--- a/src/action.c
+++ b/src/action.c
@@ -756,48 +756,32 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_DIRECTIONAL_TARGET_WINDOW:
 			if (view) {
-				/* Config parsing makes sure that direction is a valid direction */
-				struct wl_array views;
-				struct view **item;
+				int dx, dy, distance;
 				enum view_edge direction = action_get_int(action, "direction", 0);
-				int distance, dx, dy;
+				struct view *v;
+				struct view *closest_view = NULL;
 				int min_distance = INT_MAX;
-				struct view *nearest_view = NULL;
-				wl_array_init(&views);
-				view_array_append(server, &views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE);
-				wl_array_for_each(item, &views) {
-					if (*item == view) {
+				for_each_view(v, &server->views,
+					      LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+					if (v->minimized) {
 						continue;
 					}
-					if ((*item)->minimized) {
+					dx = v->current.x - view->current.x;
+					dy = v->current.y - view->current.y;
+					distance = dx * dx + dy * dy;
+					if ((direction == VIEW_EDGE_UP && dy >= 0)
+					    || (direction == VIEW_EDGE_LEFT && dx >= 0)
+					    || (direction == VIEW_EDGE_DOWN && dy <= 0)
+					    || (direction == VIEW_EDGE_RIGHT && dx <= 0)) {
 						continue;
 					}
-					dx = (*item)->current.x - view->current.x;
-					dy = (*item)->current.y - view->current.y;
-					switch (direction) {
-					case VIEW_EDGE_LEFT:
-						if (dx >= 0) continue;
-						break;
-					case VIEW_EDGE_RIGHT:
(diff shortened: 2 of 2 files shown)


docs/labwc-actions.5.scd
```diff
@@ -70,6 +70,11 @@ Actions are used in menus and keyboard/mouse bindings.
 
 	*direction* [left|up|right|down] Direction in which to shrink.
 
+*<action name="DirectionalTargetWindow" direction="value" />*
+	Focus jumps to closest window in given direction.
```
tokyo4j, line 74: I prefer "Move focus to ~"

docs/labwc-actions.5.scd
```diff
@@ -70,6 +70,11 @@ Actions are used in menus and keyboard/mouse bindings.
...
+
+	*direction* [left|up|right|down] Direction in which to jump.
```
tokyo4j, line 76: There's a tab character at the end of this line.

src/action.c
```diff
@@ -748,6 +754,37 @@ actions_run(struct view *activator, struct server *server,
...
 			break;
+		case ACTION_TYPE_DIRECTIONAL_TARGET_WINDOW:
+			if (view) {
+				int dx, dy, distance;
+				enum view_edge direction = action_get_int(action, "direction", 0);
+				struct view *v;
+				struct view *closest_view = NULL;
+				int min_distance = INT_MAX;
+				for_each_view(v, &server->views,
+					LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
```
tokyo4j, line 765: This line should be indented.

src/action.c
```diff
@@ -748,6 +754,37 @@ actions_run(struct view *activator, struct server *server,
...
+					if (v->minimized) {
+						continue;
+					}
+					dx = v->current.x - view->current.x;
+					dy = v->current.y - view->current.y;
+					distance = dx * dx + dy * dy;
+					if ((direction == VIEW_EDGE_UP && dy >= 0)
+						|| (direction == VIEW_EDGE_LEFT && dx >= 0)
+						|| (direction == VIEW_EDGE_DOWN && dy <= 0)
+						|| (direction == VIEW_EDGE_RIGHT && dx <= 0)) {
```
tokyo4j, line 775: Multi-line if statements should be indented too. I suggest following change.
```suggestion
					if ((direction == VIEW_EDGE_UP && dy >= 0)
							|| (direction == VIEW_EDGE_LEFT
								&& dx >= 0)
							|| (direction == VIEW_EDGE_DOWN
								&& dy <= 0)
							|| (direction == VIEW_EDGE_RIGHT
								&& dx <= 0)) {
```

tokyo4j:
I suggest moving the logic into another function as it seems too long and nested, but it's just my preference.

I guess the closest window should be searched based on its center position, rather than left-corner position.
Here's OpenBox's code:
https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/openbox/focus_cycle.c#L191-L192

nicolas3121:
wrapping seems to have some issues when using multiple monitors so might just have to leave that out for now

ahesford:
I don't think the distance should go midpoint to midpoint. Instead, the reference for the current view should be the midpoint, but the "nearest" window in a given direction should be the window whose opposite edge (*e.g.*, the nearest left edge when moving right, the nearest top edge when moving down, etc.) is nearest the current view's reference point. That might produce more intuitive results.

Maybe even better would be defining the current reference as the same edge that you're comparing for targets; *e.g.*, left to left when moving right, top to top when moving down.

nicolas3121:
will give that a go when i find some time to mess around with it again but with the way it is right now i haven't found any cases where it behaves in a strange way other than the ones where you're moving at 90° to where the window actually is like in the example.

When only taking into account bottom to bottom edge distance while moving up for example you could get cases like this though where the current one would make the more logical pick.
image

nicolas3121:
ended up sticking with center to center distance because when using edge to edge distance you could get stuck on a window if they shared two edges and it didn't make that much of a difference.

ahesford:
I'm not sure the "wrap" feature is particularly intuitive or all that useful. We have a "wrap" feature in desktop manipulations because desktops are discrete, disjoint, and you only ever see the active one, so it makes sense to treat them as slides in a carousel. Windows, on the other hand, are clearly laid out in a specific arrangement and it's obvious at a glance what's to the left or the right. Attempting to shift focus to the right and having it jump to the other side is going to be very surprising to users. It also complicates the focus logic.

tokyo4j:
### About reference point

> I don't think the distance should go midpoint to midpoint. Instead, the reference for the current view should be the midpoint, but the "nearest" window in a given direction should be the window whose opposite edge (e.g., the nearest left edge when moving right, the nearest top edge when moving down, etc.) is nearest the current view's reference point. That might produce more intuitive results.

A problem with this approach is that it makes it hard to guarantee the focus moves back to the original window if you invoke the action twice in the opposite directions.

Let's assume we have a large window (A) and a small window (B) whose center point is in A:



With the suggested approach with dynamic reference points, the focus can move from A to B with `down` direction, but the focus cannot move from B to A with `up` direction because the reference point of A is its bottom border which has higher y position than center of B.

Therefore I personally prefer the simple approach with just central reference points.

### About implementation

I don't understand what the algorithm is doing in the PR, but I feel it's doing it in a too complex way.
How about a simple algorithm like below?
`wlr_output_layout_get_box()` returns the entire extents of all the output if parameter `reference` is NULL.
```
function directional_target_window(view, direction, wrap):
  layout_extents = wlr_output_layout_get_box()
  min_distance = INT_MAX
  view_ret = NULL;
  ref_x = view.x + view.width / 2
  ref_y = view.y + view.height / 2
  foreach views as v:
    if v == view:
      continue
    dx = v.x + v.width / 2 - ref_x
    dy = v.y + v.height / 2 - ref_y
    switch(direction):
      case RIGHT:
        if dx < 0:
          if wrap:
            dx += layout_extents.width
          else:
            continue
        distance = dx^2+dy^2
      case LEFT:
      ...
    if distance < min_distance:
      min_distance = distance
      view_ret = v

  return view_ret
```

What this algorithm do is like shown in the picture below.
If the searched view is in the opposite direction and `wrap` is `true`, the distance is calculated like the searched view is shifted by the width/height of the extents. 

IMG_3012

If you think your approach is better, please explain what yours is doing and why it's better.

> I'm not sure the "wrap" feature is particularly intuitive or all that useful. We have a "wrap" feature in desktop manipulations because desktops are discrete, disjoint, and you only ever see the active one, so it makes sense to treat them as slides in a carousel. Windows, on the other hand, are clearly laid out in a specific arrangement and it's obvious at a glance what's to the left or the right. Attempting to shift focus to the right and having it jump to the other side is going to be very surprising to users. It also complicates the focus logic.

If "wrap" feature is simple enough like I explained above, I think it's fine to implement it. I'm not sure how much useful it is though.

nicolas3121:
> I don't understand what the algorithm is doing in the PR, but I feel it's doing it in a too complex way.
How about a simple algorithm like below?

made a little diagram, should be easier to understand this way but could just write it out in full in the actual code instead of just substituting that part with dx so it's more obvious
image
image

ahesford:
The examples here are illustrating a lot of strange corner cases that are going to be produce strange behavior. Also, I don't think you can ever fully expect a directional focus shift to be invertible, because the "nearest" relationship is not transitive in general.

I have an alternative proposal based on the concept of the original view casting a "shadow" and picking the nearest window. This can be done with some polygon intersections, which reduces to simple rectangle intersections if the shadow is "strict".

In the first figure, a strict shadow is cast in the intended direction of focus change, bounded in the perpendicular direction by the extent of the window along that axis. The next window chosen is the window whose intersection with the shadow extends furthest in the direction opposite the intended focus (*e.g.*, the leftmost intersection point for a right shift, or the topmost intersection point for a bottom shift; mirror these images appropriately for left or up focus). This is very simple and should provide reasonably intuitive results, but has the drawback that some regions may be entirely unreachable. This isn't necessarily bad, but might not be what users expect.

In the second figure, the shadow has been generalized to extend from the corners of the window at the edge in the direction of the shift, outward at 45 degrees. Effectively, the window subdivides unbounded space into four infinitely tall trapezoids that extend outward from its edges. The decision logic is the same: the next window is the one whose intersection extends furthest in the direction opposite the intended shift; however, this provides more complete coverage and might be more intuitive for users. The drawback is that polygon intersections become somewhat more complex, because the intersection of a rectangular window and a trapezoid can be any region with up to five sides. However, we aren't actually interested in representing the geometry of the intersection, only finding its maximum extent in one direction; I think we can find this with a few `if` clauses.

Neither of these approaches require any special consideration for output layout, because we can treat the shadows as unbounded and just walk the view list.

shadow1
shadow2

tokyo4j:
### About the behavior of "wrap"

I understand why just using `wlr_output_layout_get_box()` may result in counter-intuitive window selections.
Your approach would indeed make sense for some users.

### About implementation in PR

It seems the first figure doing the same as my pseudo-code.
The last expression (missing small $`1`$ in $`W`$ ?) is equal to:
$` dx' = W_0 + W_1 + dx `$

This corresponds to my pseudo-code:
```
        if dx < 0:
          if wrap:
            dx += layout_extents.width
```
So, the code in the PR should be corrected like this?
(let's put aside the problem that `rightmost_visible_point()` returns the position, not the width)
```diff
- dx = rightmost_visible_point(server, v) - dx;
+ dx = rightmost_visible_point(server, v) + dx;
```

The second figure made me understand what `* (1 + dy*dy/(dx*dx))` in the code is doing. Thanks.

### About the algorithm

I personally prefer the way OpenBox handles this.

OpenBox took a quite simple approach: the score is the Manhattan distance of the two center points and `1000000` is added to it if the candidate window is more than 45 degrees off the direction.
In this way, windows within 45 degrees off the direction are always prioritized, and by using Manhattan distance the score is more sensible to the deviation from the direction.

Another approach would be multiplying `dy` (for left&right direction) / `dx` (for up&down direction) with some constant then using Manhattan distance as a score.

<details>
<summary>Implementation in OpenBox</summary>

```c
static ObClient *focus_find_directional(ObClient *c, ObDirection dir,
                                        gboolean dock_windows,
                                        gboolean desktop_windows)
{
    gint my_cx, my_cy, his_cx, his_cy;
    gint offset = 0;
    gint distance = 0;
    gint score, best_score;
    ObClient *best_client, *cur;
    GList *it;


    if (!client_list)
        return NULL;


    /* first, find the centre coords of the currently focused window */
    my_cx = c->frame->area.x + c->frame->area.width / 2;
    my_cy = c->frame->area.y + c->frame->area.height / 2;


    best_score = -1;
    best_client = c;


    for (it = g_list_first(client_list); it; it = g_list_next(it)) {
        cur = it->data;


        /* the currently selected window isn't interesting */
        if (cur == c)
            continue;
        if (!focus_cycle_valid(it->data))
            continue;


        /* find the centre coords of this window, from the
         * currently focused window's point of view */
        his_cx = (cur->frame->area.x - my_cx)
            + cur->frame->area.width / 2;
        his_cy = (cur->frame->area.y - my_cy)
            + cur->frame->area.height / 2;


        if (dir == OB_DIRECTION_NORTHEAST || dir == OB_DIRECTION_SOUTHEAST ||
            dir == OB_DIRECTION_SOUTHWEST || dir == OB_DIRECTION_NORTHWEST)
        {
            gint tx;
            /* Rotate the diagonals 45 degrees counterclockwise.
             * To do this, multiply the matrix /+h +h\ with the
             * vector (x y).                   \-h +h/
             * h = sqrt(0.5). We can set h := 1 since absolute
             * distance doesn't matter here. */
            tx = his_cx + his_cy;
            his_cy = -his_cx + his_cy;
            his_cx = tx;
        }


        switch (dir) {
        case OB_DIRECTION_NORTH:
        case OB_DIRECTION_SOUTH:
        case OB_DIRECTION_NORTHEAST:
        case OB_DIRECTION_SOUTHWEST:
            offset = (his_cx < 0) ? -his_cx : his_cx;
            distance = ((dir == OB_DIRECTION_NORTH ||
                         dir == OB_DIRECTION_NORTHEAST) ?
                        -his_cy : his_cy);
            break;
        case OB_DIRECTION_EAST:
        case OB_DIRECTION_WEST:
        case OB_DIRECTION_SOUTHEAST:
        case OB_DIRECTION_NORTHWEST:
            offset = (his_cy < 0) ? -his_cy : his_cy;
            distance = ((dir == OB_DIRECTION_WEST ||
                         dir == OB_DIRECTION_NORTHWEST) ?
                        -his_cx : his_cx);
            break;
        }


        /* the target must be in the requested direction */
        if (distance <= 0)
            continue;


        /* Calculate score for this window.  The smaller the better. */
        score = distance + offset;


        /* windows more than 45 degrees off the direction are
         * heavily penalized and will only be chosen if nothing
         * else within a million pixels */
        if (offset > distance)
            score += 1000000;


        if (best_score == -1 || score < best_score) {
            best_client = cur;
            best_score = score;
        }
    }


    return best_client;
}
```
</details>

nicolas3121:
> So, the code in the PR should be corrected like this?
(let's put aside the problem that rightmost_visible_point() returns the position, not the width)

for the case where you're going to the left it's - dx 
image

> Manhattan distance of the two center points and 1000000 is added to it if the candidate window is more than 45 degrees off the direction.

could give this a go and see what difference it makes feeling wise but if you strip out the wrapping stuff i don't know if this  actually ends up being that much simpler 

> In the second figure, the shadow has been generalized to extend from the corners of the window at the edge in the direction of the shift, outward at 45 degrees. Effectively, the window subdivides unbounded space into four infinitely tall trapezoids that extend outward from its edges. The decision logic is the same: the next window is the one whose intersection extends furthest in the direction opposite the intended shift; however, this provides more complete coverage and might be more intuitive for users. The drawback is that polygon intersections become somewhat more complex, because the intersection of a rectangular window and a trapezoid can be any region with up to five sides. However, we aren't actually interested in representing the geometry of the intersection, only finding its maximum extent in one direction; I think we can find this with a few if clauses.

this seems like a pretty fun approach to try but might take a few days until i can get to it

ahesford:
If you're interested in looking at the "trapezoid" approach, I can look at reusing some of the edge resistance/logic for doing quick intersection tests.  I'll try to find some time to put together some prototypes.

Jackojc:
I love how much attention has been given to how windows should be focused. Would be great to get this in upstream.

nicolas3121:
will probably get back to finishing this during summer break if i don't forget about it again

lobre:
I came across this PR while checking whether labwc had directional focus. I’m impressed — I didn’t expect it to involve so many interesting algorithmic considerations. I had thought the problem was simpler! Looking forward to seeing how this exploration and its implementation evolve.

philocalyst:
@nicolas3121 Awesome PR! Let me know if you need any help, want to see this upstreamed!

nicolas3121:
I'm currently busy with uni stuff but I'm still on labwc and have been kinda missing it so i will unironically finish it this summer this time but feel free to give it a shot in the meantime.
I did probably over complicate it a bit on my first attempt so would probably just go to L1 norm with maybe a configurable multiplier for the cost of going in the perpendicular direction or something.
Integrating it more with the alt tab thing so directional cycling can be an optional strategy there is probably also not a bad idea though this would require to keep the wrapping system around I think.

PR #1721, opened 2024-04-17
IME: prevent virtual keyboard from unintentionally releasing modifiers

tokyo4j:
This MR fixes that workspace-switcher is immediately terminated when focus is moved between windows while Fcitx5 is running. https://github.com/fcitx/fcitx5/commit/b2924bd361680c493463d240a375b3f0948ae48d is also required to fix the issue.

Demos below is performed with following settings. I kept pressing Alt+Shift and clicked 1 and 2 alternately.
```xml
  <desktops>
    <names>
      <name>Workspace 1</name>
      <name>Workspace 2</name>
    </names>
  </desktops>
  <keyboard>
    <keybind key="A-S-1"><action name="GotoDesktop" to="Workspace 1" /></keybind>
    <keybind key="A-S-2"><action name="GotoDesktop" to="Workspace 2" /></keybind>
...
```

Before:

https://github.com/labwc/labwc/assets/22029524/e7f0ba77-0639-45e8-9bf5-7f7f1139e77e

After:

https://github.com/labwc/labwc/assets/22029524/920d4caf-e257-4f6d-9c52-c2d05fbe461f

### Commit Message

When Fcitx5 is activated, it creates a virtual keyboard to send keycodes to applications, then creates a keyboard grab to capture keycodes the user typed.

Before this commit, we set keyboard grab's modifiers to that of currently active keyboard, which is the virtual keyboard created in the case described above. However, since the modifiers of the virtual keyboard is empty at first, we actually set empty modifiers, even when the user is pressing modifiers. Then, Fcitx5 assumes no modifiers is pressed and redirect the modifier state back to the compositor via the virtual keyboard. As a result, when the focus is switched between windows by workspace-switcher, the workspace-switcher is immediately terminated.

To fix this issue, with this commit, the modifier state of the currently active keyboard is not set to the keyboard grab if the keyboard is a virtual keyboard created by the same input-method client.

Fcitx5's commit below is also required to fix the issue. https://github.com/fcitx/fcitx5/commit/b2924bd361680c493463d240a375b3f0948ae48d

### DIFF
--- a/src/input/ime.c
+++ b/src/input/ime.c
@@ -11,6 +11,20 @@
 	(wl_resource_get_client((wlr_obj1)->resource) \
 	== wl_resource_get_client((wlr_obj2)->resource))
 
+static bool
+is_keyboard_emulated_by_input_method(struct wlr_keyboard *keyboard,
+		struct wlr_input_method_v2 *input_method)
+{
+	if (!keyboard || !input_method) {
+		return false;
+	}
+
+	struct wlr_virtual_keyboard_v1 *virtual_keyboard =
+		wlr_input_device_get_virtual_keyboard(&keyboard->base);
+
+	return virtual_keyboard && SAME_CLIENT(virtual_keyboard, input_method);
+}
+
 /*
  * Get keyboard grab of the seat from keyboard if we should forward events
  * to it.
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, not tested.

johanmalm:
Thanks.

PR #1722, opened 2024-04-18
add .clang-format (with a disclaimer)

jlindgren90:
.clang-format gets picked up nowadays by some IDEs (tested Qt Creator) for auto-formatting aids. With the large caveat that it's not 100% accurate to labwc's preferred coding style, I've found that this config gets it right 90% of the time, which is enough to be helpful.

I added a prominent disclaimer comment to the top of the file to warn people away from relying completely on clang-format. This isn't meant to replace manual attention to formatting details (or to replace ./scripts/check) but I think it doesn't hurt.


.clang-format
```diff
@@ -0,0 +1,24 @@
...
+AlwaysBreakAfterDefinitionReturnType: true
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeBraces: Linux
+IndentCaseLabels: false
+SpaceBeforeParens: ControlStatementsExceptControlMacros
+ForEachMacros: ['for_each_view',
+                'wl_array_for_each',
+                'wl_list_for_each',
+                'wl_list_for_each_reverse',
+                'wl_list_for_each_safe']
```
Consolatis, line 24: There is also `wl_list_for_each_reverse_safe`.

Fine with adding `.clang-format`.

johanmalm: I'm fine with adding it too.

We did actually used to have one :smile:

jlindgren90: Added `wl_list_for_each_reverse_safe`

PR #1724, opened 2024-04-18
Distinguish between "should it be decorated" and "should it use SSD"

xi:
The `ToggleDecorations` action in Openbox was used to decide whether decorations should be drawn or not. The `keepBorder` setting is used to decide whether "undecorated" windows should still have a border or not.

In Wayland this is more complicated because we have to consider client side decorations (CSD) in addition to the traditional server side decorations (SSD).

Initially, `ToggleDecorations` in labwc toggled SSD on and off. Since support for `keepBorder` was added in #876, `ToggleDecorations` cycles through full SSD, only borders, and no SSD. This approach has some issues:

- Whether SSD is enabled initially is negotiated for each window individually, there is no way to know what `<windowRule><action name="ToggleDecorations"/>` would do.
- It is not clear which of the 3 states the `Decorate` and `Undecorate` actions are supposed to transition to

My proposal is to split this into two different topics: `ToggleDecorations`, `Decorate`, and `Undecorate` toggle SSD on and off, like before. Adding `Decorate` and `Undecorate` allows to target a specific state without needing to know the initial state.

In addition to that, the new set of actions `ToggleTitlebar`, `ShowTitlebar`, and `HideTitlebar` can be used to modify the SSD. If `keepBorder` is off, the border will be hidden along with the titlebar.

This gives us all the combinations there are. However, there is some redundancy:

- `HideTitlebar` with `keepBorder=no` is effectively the same as `Undecorate`
- `<windowRules><action name="Undecorate">` is effectively the same as `<windowRule serverDecoration="no"/>`

i therefore went ahead and removed `keepBorder` again. I did not remove `serverDecoration` because I figured that it is more performant/causes less flicker than the corresponding actions.

Some examples:

```xml
<!-- hide titlebar for all windows -->
<windowRule identifier="*">
  <action name="HideTitlebar"/>
</windowRule>

<!-- hide titlebar on maximize -->
<keybind key="W-Up">
  <action name="HideTitlebar"/>
  <action name="Maximize"/>
</keybind>
```

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -118,13 +118,13 @@ Actions are used in menus and keyboard/mouse bindings.
 	yes.
 
 *<action name="ToggleDecorations" />*
-	Toggle decorations of focused window.
+	Toggle server side decorations of focused window.
 
 *<action name="Decorate" />*
-	Enable decorations of focused window.
+	Enable server side decorations of focused window.
 
 *<action name="Undecorate" />*
-	Disable decorations of focused window.
+	Disable server side decorations of focused window.
 
 *<action name="ToggleFullscreen" />*
 	Toggle fullscreen state of focused window.
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -129,9 +129,9 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="ToggleTitlebar" />*
 	Toggle titlebar of focused window.
 
-	If the 'keepBorder' configuration option is enabled (as in the default
-	configuraton), then a border will be left as the only decorations around the
-	window.
+	A border will be left as the only decorations around the window.
+	If you completely want to remove all server side decorations, use
+	'ToggleDecorations' instead.
 
 	This has no effect on windows that do not have server side decorations.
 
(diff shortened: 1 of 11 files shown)


ahesford:
I like the idea of `Decorate` and `Undecorate`, but think that the `*Titlebar` actions are redundant and unnecessary. Also, we need to keep around the current `ToggleDecorations` and `theme.keepBorder` behavior for a long time, because we don't want to break existing configurations.

My proposal would be:
1. Add the new `Decorate` and `Undecorate` actions.
2. Support a new `keepBorder` argument to both the `ToggleDecorations` and `Undecorate` actions. If `keepBorder` is `yes` on a `ToggleDecorations` action, then `ToggleDecorations` cycles through the three states (full decoration -> border only -> no decoration); when its `keepBorder` action is `no`, it cycles through only two states (full decoration -> nodecoration). [This is the current behavior, just with `keepBorder becoming an argument to the action itself.] In the `Undecorate action, `keepBorder` will behave intuitively: when `yes`, `Undecorate` will remove only the titlebar; otherwise, it will remove the border *and* titlebar. The `Decorate` action should always show **both** the titlebar and border.
3. Make the existing `theme.keepBorder` field the default value of the `keepBorder` arguments for any `ToggleDecorations` and `Undecorate` actions that do not provide a specific value.

ahesford:
I find #1725 preferable, but let's see if others agree.

xi:
Closed in favor of #1733

PR #1725, opened 2024-04-18
Implement Decorate/Undecorate

xi:
This is an alternative approach for #1724 as proposed by @ahesford. This approach is so different that I thought a new pull request makes more sense.

This keeps backwards compatibility and thereby also the confusion about SSD and titlebar. I still think the other approach is cleaner. But as far as I was able to figure out, both can accomplish the same.

I was not sure how to handle the situation in which a window has no SSD and a user uses `Undecorate` with `keepBorder="yes"`: My first instinct was to make this idempotent, i.e. you end up in the same state no matter where you started from, in this case: borders. On the other hand, it is a bit weird that a call to `Undecorate` actually adds decorations.

My original motivation for this is that I want to remove title bars by default (without adding borders to CSD windows). This is only possible with the latter approach.  It is still possible to get something idempotent by calling `Decorate` first. So that's what I ended up with.

I hope you get why I think the other approach is cleaner.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -117,7 +117,7 @@ Actions are used in menus and keyboard/mouse bindings.
 	upper-left corner of the window associated with the action. Default is
 	yes.
 
-*<action name="ToggleDecorations" />*
+*<action name="ToggleDecorations" keepBorder="value" />*
 	Toggle decorations of focused window.
 
 	This is a 3-state action which can be executed multiple times:
@@ -125,14 +125,18 @@ Actions are used in menus and keyboard/mouse bindings.
 	- Remaining decorations will be disabled
 	- Decorations will be shown normally
 
-	By disabling the theme configuration 'keepBorder' the first step
-	will be removed and the action only toggles between on and off.
+	*keepBorder* [yes|no] If set to no, the first step will be removed and
+	the action only toggles between on and off. Default is the theme
+	configuration of the same name.
 
 *<action name="Decorate" />*
 	Enable decorations of focused window.
 
-*<action name="Undecorate" />*
+*<action name="Undecorate" keepBorder="value" />*
 	Disable decorations of focused window.
+
+	*keepBorder* [yes|no] Keep a small border (and resize area) around the
+	window. Default is the theme configuration of the same name.
 
 *<action name="ToggleFullscreen" />*
 	Toggle fullscreen state of focused window.
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -128,6 +128,12 @@ Actions are used in menus and keyboard/mouse bindings.
 	By disabling the theme configuration 'keepBorder' the first step
 	will be removed and the action only toggles between on and off.
 
+*<action name="Decorate" />*
+	Enable decorations of focused window.
+
+*<action name="Undecorate" />*
+	Disable decorations of focused window.
+
 *<action name="ToggleFullscreen" />*
 	Toggle fullscreen state of focused window.
 
--- a/include/view.h
+++ b/include/view.h
@@ -30,6 +30,12 @@ enum ssd_preference {
 	LAB_SSD_PREF_SERVER,
 };
 
+enum ssd_mode {
+	LAB_SSD_MODE_NONE,
+	LAB_SSD_MODE_BORDER,
+	LAB_SSD_MODE_FULL
+};
+
 /**
  * Directions in which a view can be maximized. "None" is used
  * internally to mean "not maximized" but is not valid in rc.xml.
@@ -452,7 +458,7 @@ void view_toggle_visible_on_all_workspaces(struct view *view);
 bool view_is_tiled(struct view *view);
 bool view_is_floating(struct view *view);
 void view_move_to_workspace(struct view *view, struct workspace *workspace);
-void view_set_decorations(struct view *view, bool decorations);
+void view_set_decorations(struct view *view, enum ssd_mode mode);
 void view_toggle_fullscreen(struct view *view);
 void view_invalidate_last_layout_geometry(struct view *view);
 void view_adjust_for_layout_change(struct view *view);
--- a/include/view.h
+++ b/include/view.h
@@ -446,7 +446,7 @@ void view_maximize(struct view *view, enum view_axis axis,
 	bool store_natural_geometry);
 void view_set_fullscreen(struct view *view, bool fullscreen);
 void view_toggle_maximize(struct view *view, enum view_axis axis);
-void view_toggle_decorations(struct view *view);
+void view_toggle_decorations(struct view *view, bool keep_border);
 
 bool view_is_always_on_top(struct view *view);
 bool view_is_always_on_bottom(struct view *view);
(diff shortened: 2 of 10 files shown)


docs/labwc-actions.5.scd
```diff
@@ -117,16 +117,26 @@ Actions are used in menus and keyboard/mouse bindings.
 	upper-left corner of the window associated with the action. Default is
 	yes.
 
-*<action name="ToggleDecorations" />*
+*<action name="ToggleDecorations" keepBorder="yes" />*
```
ahesford, line 120: Placeholder text uses `value` or an empty string rather than one of the specific options.

```suggestion
*<action name="ToggleDecorations" keepBorder="value" />*
```

docs/labwc-actions.5.scd
```diff
@@ -117,16 +117,26 @@ Actions are used in menus and keyboard/mouse bindings.
...
-	By disabling the theme configuration 'keepBorder' the first step
-	will be removed and the action only toggles between on and off.
+	*keepBorder* [yes|no] If set to no, the first step will be removed and
+	the action only toggles between on and off. Default is the theme
+	configuration of the same name.
+
+*<action name="Decorate" />*
+	Enable decorations of focused window.
+
+*<action name="Undecorate" keepBorder="yes" />*
```
ahesford, line 135: As above, prefer `value` or an empty string to one of the specific choices.
```suggestion
*<action name="Undecorate" keepBorder="value" />*
```

include/view.h
```diff
@@ -30,6 +30,12 @@ enum ssd_preference {
 	LAB_SSD_PREF_SERVER,
 };
 
+enum ssd_mode {
+	LAB_SSD_MODE_NONE = false,
+	LAB_SSD_MODE_FULL = true,
+	LAB_SSD_MODE_BORDER,
```
ahesford, line 36: This is more readable if it progresses from least to most decorated, and doesn't abuse Booleans for their integer values. Setting `_NONE = 0` is sufficient to allow things like `if (!mode)` without over-specifying the other don't-care values.

```suggestion
	LAB_SSD_MODE_NONE = 0,
	LAB_SSD_MODE_BORDER,
	LAB_SSD_MODE_FULL,
```

src/action.c
```diff
@@ -340,6 +344,14 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		break;
+	case ACTION_TYPE_TOGGLE_DECORATIONS:
+	case ACTION_TYPE_UNDECORATE:
+		if (!strcasecmp(argument, "keepBorder")) {
+			action_arg_add_bool(action, argument,
+				parse_bool(content, rc.ssd_keep_border));
```
ahesford, line 351: I'm not sure that this will work as intended; `rc.ssd_keep_border` might not be properly set, depending on the order of lines in `rc.xml` and the state of the parser. Some testing will be required.

ahesford (commented):
I haven't tested this, but it seems straightforward, respects my preference, and maintains backward compatibility.

I disagree that your other approach is cleaner, because it adds a lot of duplicate options to achieve the same effect. If you really think that there is merit in providing a way to implement your intended decorative scheme for SSD only, I'd suggest that the cleanest approach is this implementation, with an additional flag for `Decorate`, `Undecorate` and `ToggleDecorations` that would optionally limit the action to SSD only.

src/action.c
Consolatis: Agree, this should likely just add the argument. `action_get_bool()` in `actions_run()` already falls back to `rc.ssd_keep_border`.

Consolatis:
First of all, thanks for your 2 PRs.
I only had a chance to look into both of them briefly but so far I prefer this one as well.

What I am not sure about is the `Decorate` action always adding both, border and titlebar. Maybe it should also get a new argument (but default to show titlebar and borders). Otherwise things are pretty confusing if a users wants to use borders only.

> On the other hand, it is a bit weird that a call to Undecorate actually adds decorations.

This indeed feels weird. Does Undecorate actually need the argument if we have one for `Decorate`?

I'll look into both PRs in more details in the coming days.

include/view.h
xi: Unfortunately it's not that simple. I used this mapping so that existing calls to `view_set_decorations()` continue to work. I will have to change them, too. Maybe I will have time for that tomorrow.

src/action.c
xi: I wasn't sure what `action_arg_from_xml_node()` is used for. If you say it is fine to use `true` here I can change that.

xi:
> I disagree that your other approach is cleaner, because it adds a lot of duplicate options to achieve the same effect.

Note that it adds more options, but needs less code to do that. For me it is easier to grasp and has fewer unexpected edge cases. But it also requires a deeper understanding of the low-level details (e.g. the difference between SSD and CSD). I also don't think the previous approach was perfect.

Maybe we can find an even better approach that is both clear and high level and also doesn't break backwards compatibility.

Then again, I believe that breaking backwards compatibility might be justified in some cases. The 3-state `ToggleDecorations` and the redundancy between `<windowRule><action name="Undecorate">` and `<windowRule serverDecoration="no"/>` are not ideal, and if we want to fix that, maybe it's better now then later.

What is your general position on breaking changes? Should I try to come up with a proposal or can I save myself the trouble?

include/view.h
Consolatis: `false` should be the same as `0` and `true` should be the same as `1` so keeping the order but just removing `= true` and replacing `= false` with `= 0` should keep working. It is a bit strange though to upcast `bool` in such a way and IMHO we should update the callers as well.

Consolatis:
> What is your general position on breaking changes? Should I try to come up with a proposal or can I save myself the trouble?

We usually try pretty hard to prevent breaking changes for user facing configuration, same for UX patterns.
If there are situations where this is justified we usually try to accept existing configs for at least one release cycle and print out a deprecation warning but still accept the old variant. In some cases we accept that something is a breaking change and then note it as such in the [changelog](https://github.com/labwc/labwc/blob/master/NEWS.md#changed) but we try to keep the impact minimal.

To save some work it is always suggested to discuss some approach beforehand.

ahesford:
Remember that there is more to "clean" than lines of code. If we have a lot of overlap with the effects of actions, it muddies the user interface for configuring these things, and also allows for weird and buggy interactions when chaining together multiple actions that partially overlap. This produces *more* edge cases, not fewer. Simplifying the action list and adding a few options to make user intent more clear is worth a few extra lines to parse and interpret the actions.

I'm not really advocating for exposing SSD/CSD limiters in the config; I only suggested that this is a possibility if the distinction is appropriate. After thinking some more, I think the best option is to leave a `keepBorders` argument to `Undecorate`, but make sure the logic behind the whole action is correct. The `Undecorate` option should *always* disable a titlebar, and additionally disable the border (if it exists) iff `keepBorder` is not true. Under no circumstances should `Undecorate` **add** a border.

Consolatis:
Just for completeness, there were ideas floating around at some point to add something like `<action name="SetDecoration" type="full|border|none" />`.

xi:
Closed in favor of #1733

Issue #1728, opened 2024-04-19
Cleaner approach to decorations


xi:
I believe the current approach to decorations is a bit messy and hard to extend. I tried to propose extensions in #1724 and #1725 with mixed success. If we want to avoid breaking changes in the future, I think it is better to think hard about the design first before merging any changes.

I am probably overthinking this. But now that the overthinking has already happened, I might as well share it.

## Why is this complicated?

We cannot directly port concepts from openbox because Wayland distinguishes between client side decorations (CSD) and server side decorations (SSD). 

## The current situation

As of 

- There are 3 modes of decoration: 'none', 'borders', and 'full'.
- The `ToggleDecoration` action is implemented, `Decorate` and `Undecorate` are not.
- CSD negotiation is implemented for both xdg-shell and xwayland.
  - The results can be overwritten using `<windowRule serverDecoration="">`
  - A default can be provided using `<core><decoration>`
  - CSD windows get initialized with 'none' decoration, SSD windows get initialized with 'full' decoration
- `<theme><keepBorder>` is implement, but behaves differently from openbox (and also from its documentation) (see #876)
  - If it is disabled, `ToggleDecorations` toggles between 'none' and 'full'
  - If it is enabled, `Toggledecorations` cycles between 'border', 'none', and 'full'
  - It has no effect on the initial decoration of CSD windows (as claimed by the documentation)
- The implementation doesn't use the 3 modes. Instead it uses the following:
  - `view->ssd_enabled` and `view->ssd_titlebar_hidden` to track whether ssd and titlebar are enabled
  - `view->ssd` and `view->ssd->titlebar.tree->node->enabled` to track whether ssd and titlebar actually exist. For example, full screen windows do not have decoration, even if it is enabled.
  - `has_ssd(view)` stores the result of the CSD/SSD negotiation

## What do we want?

As a start for discussion:

- We want to keep backwards compatibility as much as possible. Among other things, this means that we have to stick with the the 3-state `ToggleDecorations` and `keepBorder`.
- For CSD windows, undecorated means 'none'
- For SSD windows, the meaning of undecorated depends on `keepBorder`
  - if `keepBorder=yes`, undecorated means 'borders'
  - if `keepBorder=yes`, undecorated means 'none'
- Decorated always means 'full'
[...]

PR #1729, opened 2024-04-19
decorations: allow to globally disable titlebars

xi:
Mostly for reference: This is my implementation for one of the ideas from #1728. While it does cover my main use case, it is very inflexible. Feel free to close this.


xi:
Closed in favor of #1733

PR #1730, opened 2024-04-19
Implement Decorate/Undecorate with smart mode selection

xi:
Yet another approach for #1728.

This one is basically the `<action name="SetServerSideDecorationMode" mode="none|border|all"/>` approach. But on top of the three explicit values there are also "yes", "no", and "initial" which are meant to do the right thing™ automatically:

- "yes" is just an alias for "full"
- the behavior of "no" depends on `keepBorder` and whether the window uses client side decorations
- the behavior of "initial" depends on whether the window uses client side decorations

I did not yet change anything about `ToggleDecorations`. But with this approach there is a straight forward option: `<action name="ToggleDecorations" mode1="value" mode2="value"/>` toggles between mode1 and mode2 (i.e. `new_mode = current_mode == mode1 ? mode2 : mode1`). If they are not both provided, the old 3-state behavior is used instead.

While very similar, I prefer this approach over #1725 because it provides both a high-level and a low-level API for users. The defaults make sense, but users still have full control over the exact behavior.


ahesford:
The six-way selection here is far too complicated and overloaded. This is a non-starter for me. We aren't trying to be Turing complete with window decorations!

 In #1725, adding a `keepBorders` override to `Undecorate` and implementing it correctly provides conceptual simplicity: remove a title bar if one is shown and, if `keepBorders` is not "yes", also remove the borders if they are shown. Adding an optional `mode` argument to `Decorate`, which may be something like `borders|all`, complements that by allowing you to explicitly set a desired state right out of the gate.

At this point, you now have four competing PRs open to adjudicate your decoration issue. Please let discussion percolate on that issue and await a consensus before opening any more alternatives, and maybe use a few of your alternatives for awhile to get a sense of what you think feels right before proposing all of them for inclusion.

xi:
Closed in favor of #1733

PR #1731, opened 2024-04-19
Implement type filter for window rules

xi:
This builds on #1490. I reverted the hard-coded rules for desktop windows for now. Instead I exposed the feature to users and added heuristics for xdg-shell, as [proposed by @Consolatis](https://github.com/labwc/labwc/compare/master...Consolatis:labwc:feature/window_rule_root_only). See also the [relevant openbox documentation](http://openbox.org/wiki/Help:Applications).

I personally want to use it like this:

```xml
<windowRule type="normal">
  <action name="Maximize"/>
</windowRule>
```

It works quite well, but not perfect. For example, the dialog to enter the primary password for thunderbird behaves correctly under xwayland, but not under native wayland. I intentionally kept the description of the internal heurustics in the docs vague so we can refine them later (e.g. based on size hints).

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -820,7 +820,7 @@ defined as shown below.
 <windowRules>
 
   <!-- Action -->
-  <windowRule identifier="" title="">
+  <windowRule identifier="" title="" type="">
     <action name=""/>
   </windowRule>
 
@@ -832,10 +832,10 @@ defined as shown below.
 
 *Criteria*
 
-*<windowRules><windowRule identifier="" title="" matchOnce="">*
+*<windowRules><windowRule identifier="" title="" type="" matchOnce="">*
 	Define a window rule for any window which matches the criteria defined
-	by the attributes *identifier* or *title*. If both are defined, AND
-	logic is used, so both have to match.
+	by the attributes *identifier*, *title*, or *type*. If more than one
+	is defined, AND logic is used, so all have to match.
 	Matching against patterns with '\*' (wildcard) and '?' (joker) is
 	supported. Pattern matching is case-insensitive.
 
@@ -843,6 +843,12 @@ defined as shown below.
 	for XWayland clients.
 
 	*title* is the title of the window.
+
+	*type* [desktop|dock|toolbar|menu|utility|splash|dialog|dropdown_menu|
+	popup_menu|tooltip|notification|combo|dnd|normal] relates to
+	NET_WM_WINDOW_TYPE for XWayland clients. Native wayland clients have
+	type "dialog" when they have a parent or a fixed size, or "normal"
+	otherwise.
 
 	*matchOnce* can be true|false. If true, the rule will only apply to the
 	first instance of the window with the specified identifier or title.
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -77,6 +77,45 @@ enum font_place {
 static void load_default_key_bindings(void);
 static void load_default_mouse_bindings(void);
 
+static int
+parse_window_type(const char *type)
+{
+	if (!type) {
+		return -1;
+	}
+	if (!strcasecmp(type, "desktop")) {
+		return NET_WM_WINDOW_TYPE_DESKTOP;
+	} else if (!strcasecmp(type, "dock")) {
+		return NET_WM_WINDOW_TYPE_DOCK;
+	} else if (!strcasecmp(type, "toolbar")) {
+		return NET_WM_WINDOW_TYPE_TOOLBAR;
+	} else if (!strcasecmp(type, "menu")) {
+		return NET_WM_WINDOW_TYPE_MENU;
+	} else if (!strcasecmp(type, "utility")) {
+		return NET_WM_WINDOW_TYPE_UTILITY;
+	} else if (!strcasecmp(type, "splash")) {
+		return NET_WM_WINDOW_TYPE_SPLASH;
+	} else if (!strcasecmp(type, "dialog")) {
+		return NET_WM_WINDOW_TYPE_DIALOG;
+	} else if (!strcasecmp(type, "dropdown_menu")) {
+		return NET_WM_WINDOW_TYPE_DROPDOWN_MENU;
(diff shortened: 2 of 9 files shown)


src/xdg.c
```diff
@@ -17,6 +17,23 @@
 #define LAB_XDG_SHELL_VERSION (2)
 #define CONFIGURE_TIMEOUT_MS 100
 
+static bool
+xdg_toplevel_view_contains_window_type(struct view *view, int32_t window_type)
+{
+	assert(view);
+	assert(view->type == LAB_XDG_SHELL_VIEW);
```
Consolatis, line 24: The 2nd assert is not necessary, the abstraction and the function being static already makes sure that its always a xdg view.

src/xwayland.c
```diff
@@ -32,6 +34,15 @@ xwayland_surface_contains_window_type(
 	return false;
 }
 
+static bool
+xwayland_view_contains_window_type(struct view *view, int32_t window_type)
+{
+	assert(view);
+	assert(view->type == LAB_XWAYLAND_VIEW);
```
Consolatis, line 41: Same story here, the 2nd assert is not necessary.

Consolatis (commented):
Looks good on the first glance, two minor nitpicks below. Will test in the coming days.

src/config/rcxml.c
```diff
@@ -35,6 +35,7 @@
 #include "view.h"
 #include "window-rules.h"
 #include "workspaces.h"
+#include "xwayland.h"
```
Consolatis, line 38: This include doesn't seem necessary

src/xwayland.c
```diff
@@ -6,6 +6,8 @@
 #include "common/array.h"
 #include "common/macros.h"
 #include "common/mem.h"
+#include "common/list.h"
+#include "action.h"
```
Consolatis, line 10: Same for these two

Consolatis:
Tested and seems to work fine for the most part.
Some windows are a bit weird (noticed with the VLC preference window which marks itself with `Dialog` and `Normal`) but that doesn't seem like something we can properly handle here.

Could you squash everything into a single commit (and add a `Co-Authored-By: Grigory Kirillov <txgk@bk.ru>`)?
Otherwise I can do so later before merging.

With that and the last two nitpicks about the includes: LGTM.

docs/labwc-config.5.scd
```diff
@@ -844,6 +844,9 @@ defined as shown below.
 
 	*title* is the title of the window.
 
+	*type* relates to NET_WM_WINDOW_TYPE for XWayland cleints and internal
+	heuristics for native Wayland windows.
```
Consolatis, line 848: `s/cleints/clients/`

xi: That's a long list.

Consolatis: I'd just add the whole list for now. We can always update it later.
Without that list the feature is basically useless without reading the code.

xi:
I noticed that `<windowSwitcher><fields><field content="" width="%">` also has a content value `type` which is either "xdg-shell" or "xwayland". Is it a problem that these are called the same?

Consolatis:
> I noticed that `<windowSwitcher><fields><field content="" width="%">` also has a content value `type` which is either "xdg-shell" or "xwayland". Is it a problem that these are called the same?

No, they have different parents which are also considered during parsing so that should be fine.

johanmalm:
I'm fine with this, but have three comments for consideration:

1. We should describe what we mean by "internal heuristics" for native Wayland windows (i.e. that we determie NORMAL vs DIALOG based the toplevel being the root-toplevel or not).
2. I think some xdg-shell 'dialogs' are root-toplevels themselves (IIRC gimp's about dialog might be one). I don't think there is a whole lot we can do about this so I'm still okay with accepting the PR as written as it will provide value to users. I suggest we document that quering 'dialog' on an xdg-shell dialog may not always return true.
3. Consider implementing a `view_contains_window_type()` instead of calling the impl directly.

xi:
@johanmalm as described in above, I intentionally kept the docs vague so we can refine the heuristics later (e.g. based on size hints).

xi:
I noticed that I had missed `other_instances_exist()`, so I fixed that with the latest push.

johanmalm:
> @johanmalm as described above, I intentionally kept the docs vague so we can refine the heuristics later (e.g. based on size hints).

Yes, understand, but I feel uncomfortable with leaving docs unclear because (a) people don’t like surprises if the behaviour changes later; and (b) we tend to get questions on this sort of stuff later (when I will have forgotten this conversation) and if the docs aren’t clear it takes me/others longer to provide help than we wish it to.

As I believe that relatively few people will use this, I’m okay with merging it in an “experimental” state accepting that it may introduce a regression later. But it’s important to me that we point out in the docs that it is “experimental” and may change later.

The alternative here (and possibly a better one) is to not merge the xdg-shell bit of the patch until we’ve worked out how we want it behave so that we reduce the risk of regression.

I don’t understand the size-hint proposal, but suspect that using PIDs might help.

https://github.com/swaywm/sway/blob/087226d997c15f4df30542778854999c632642a3/sway/tree/view.c#L499

xi:
What I mean by size hints: https://github.com/swaywm/sway/blob/087226d997c15f4df30542778854999c632642a3/sway/desktop/xdg_shell.c#L218 or https://codeberg.org/dwl/dwl/src/commit/577d8da6d17dad5f847795582ebf813f3c5c04a7/client.h#L237

xi:
> I feel uncomfortable with leaving docs unclear because (a) people don’t like surprises if the behaviour changes later; and (b) we tend to get questions on this sort of stuff later (when I will have forgotten this conversation) and if the docs aren’t clear it takes me/others longer to provide help than we wish it to.

I implemented what sway does and described it in the documentation. Sway has been using this heuristic for a while now, so I believe it is good enough.

Consolatis:
> 3\. Consider implementing a `view_contains_window_type()` instead of calling the impl directly.

This is still missing, otherwise this LGTM.

Consolatis:
Thanks :)

PR #1733, opened 2024-04-20
add SetDecorations action

### LINKED ISSUE #1728: Cleaner approach to decorations

xi:
as discussed in IRC. resolves #1728

We also discussed adding a "decorations" query for the if-action. I will work on that separately, but this already lays some ground work (e.g. `view_get_ssd_mode`).


johanmalm:
I appreciate the work on this and noted the lengthy analysis. I haven’t been able to keep up with the story/thinking so if you want me to review anything, please could you point me to it.

If @ahesford or @Consolatis are already on the case I’m happy not to duplicate effort, but as I said if independent view is required (by me) please make it clear.

docs/labwc-actions.5.scd
```diff
@@ -117,6 +117,16 @@ Actions are used in menus and keyboard/mouse bindings.
 	upper-left corner of the window associated with the action. Default is
 	yes.
 
+*<action name="SetDecorations" decorations="value" force_ssd="no" />*
```
Consolatis, line 120: Suggest using `forceSSD="no"` to match the usual user-facing argument style.

docs/labwc-actions.5.scd
```diff
@@ -117,6 +117,16 @@ Actions are used in menus and keyboard/mouse bindings.
...
+	Set decorations of focused window.
+
+	*decorations* [full|border|none] On border, only the titlebar will be
+	hidden, borders and resize area are kept. Default is full.
```
Consolatis, line 124: This implies that the SSD is currently shown, suggest rewording it to something like

```
- *full* enables the whole server side decorations
- *border* only the border and invisible resize area are enabled
- *none* disables everything
```
It doesn't have to be a list.

include/view.h
```diff
@@ -497,6 +505,7 @@ struct output *view_get_adjacent_output(struct view *view, enum view_edge edge,
 	bool wrap);
 enum view_axis view_axis_parse(const char *direction);
 enum view_edge view_edge_parse(const char *direction);
+enum ssd_mode ssd_mode_parse(const char *mode);
```
Consolatis, line 508: This function (and the enum) should likely live in `ssd.{h,c}`

include/ssd.h
```diff
@@ -87,6 +87,7 @@ bool ssd_is_button(enum ssd_part_type type);
 bool ssd_part_contains(enum ssd_part_type whole, enum ssd_part_type candidate);
 
 /* TODO: clean up / update */
+bool has_ssd(struct view *view);
```
Consolatis, line 90: If this has to be public, it should be prefixed with `ssd_`.

ahesford: It's also misnamed, because the function indicates whether the view *wants* SSD, not whether it has them.

docs/labwc-actions.5.scd
xi: I had checked and found `output_name`. But you are right, camelcase that seems to be the exception.

Consolatis: Whoops, that should likely be fixed then (e.g. updated in the docs, parsing `outputName` and logging an deprecation error when parsing output_name).

johanmalm:
Thanks for working on this. Sorry it took a while to get into the merge window.

PR #1734, opened 2024-04-20
reuse query logic for window rules

xi:
I noticed that `view_matches_criteria()` and `view_matches_query()` are basically the same, so I reused some code. This builds on #1731  to avoid merge conflcts.

### DIFF
--- a/src/window-rules.c
+++ b/src/window-rules.c
@@ -14,37 +14,13 @@
 #include "window-rules.h"
 
 static bool
-matches_criteria(struct window_rule *rule, struct view *view)
-{
-	const char *id = view_get_string_prop(view, "app_id");
-	const char *title = view_get_string_prop(view, "title");
-
-	if (rule->identifier) {
-		if (!id || !match_glob(rule->identifier, id)) {
-			return false;
-		}
-	}
-	if (rule->title) {
-		if (!title || !match_glob(rule->title, title)) {
-			return false;
-		}
-	}
-	if (rule->window_type >= 0) {
-		if (!view_contains_window_type(view, rule->window_type)) {
-			return false;
-		}
-	}
-	return true;
-}
-
-static bool
-other_instances_exist(struct window_rule *rule, struct view *self)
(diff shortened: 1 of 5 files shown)


Consolatis:
Its an interesting approach (e.g. creating the `query` on-demand).

src/window-rules.c
```diff
@@ -55,10 +30,15 @@ other_instances_exist(struct window_rule *rule, struct view *self)
 static bool
 view_matches_criteria(struct window_rule *rule, struct view *view)
 {
-	if (rule->match_once && other_instances_exist(rule, view)) {
+	struct view_query query;
```
Consolatis, line 33: Suggest `struct view_query query = {0};`, otherwise stuff might be uninitialized and when we add further conditions in the future this might break horribly.

xi: Just so don't do something dumb: This is the definition of `struct view_query`:

```c
struct view_query {
	struct wl_list link;
	char *identifier;
	char *title;
	int window_type;
};
```

`link` is not used in `view_matches_query()`. Do we have to initialize it anyway? Would that work with both options you proposed? (I am not that experienced with C)

Consolatis: Yes, AFAIR that should always work.
Everything not defined within the initializer is set to `0` or `0 equivalent` (e.g. `NULL` pointer).

CC @ahesford to ensure I don't talk nonsense.

ahesford: Yes, if you initialize with ` = { something, ... }` and the list of specified values is not complete, every other field in the struct will be default initialized (which would be `0` in these cases).

If you don't initialize at all, then static variables are default initialized but local variables will be uninitialized, inheriting whatever garbage was on the stack (or heap) at the time they are created.

johanmalm:
Oh, I like the look of this. I think I had in mind doing something similar to this, but never got back to it.
Haven't looked through in detail yet.

Consolatis:
LGTM, not tested.

PR #1736, opened 2024-04-20
add .scd to editorconfig

xi:


droc12345:
I'm not a fan of trying to **enforce** 80 col line lengths on .scd file, by way of editorconfig.

It's a good guideline to stay at or below 80 columns, but there are a few cases 
where trying to force lines to 80 cols doesn't look right.

xi:
As far as I know, editorconfig doesn't enforce anything. It just sets editor defaults. I have tabs set to 2 characters by default, so editing the documentation files was a little awkward. So I thought adding this could make life easier for contributors like me.

droc12345:
Adding it to xml/build files would be better, or make one just for scd files with the few settings you need.
c/h files enforce things like tab size, indent size, etc that are different than text files.

Consolatis:
I really do not have any opinion on the various editor configs. I am just using nano like a caveman.

johanmalm:
I like this.
My vim defaults to tabs anyway, so I hadn't thought about it.

PR #1737, opened 2024-04-20
overlay: s/preview/overlay/ and draw filled/outlined rectangle independently

### LINKED ISSUE #1735: PR #1702 after thoughts
@tokyo4j

Ref:
- https://github.com/labwc/labwc/pull/1702/commits/5b1af2f0d9af3a21f045ef0a60fb2b076eb0a482
- https://github.com/labwc/labwc/pull/1702
- https://github.com/labwc/labwc/pull/1702#issuecomment-2056446222
- https://labwc.github.io/labwc-theme.5.html#entry_snapping.preview.edge.bg.color

Whilst reviewing recent commits I got to `overlay: add theme settings for colors and border width` and tried to configure the border in accordance with the commit message.

```
snapping.overlay.[region|edge].bg.color: #8080b380
snapping.overlay.[region|edge].border.color: #ffffff,#000000,#ffffff
snapping.overlay.[region|edge].border.width: 1
```

I note two things:

1. The man page and parser use `preview` instead of `overlay`. We could simply just accept that we made a mistake in the commit message (no big deal) and when I update `NEWS.md` I use `preview` instead. BUT, in hindsight I wonder if `overlay` is actually a better word because we use `preview` in other places (e.g. `rc.xml`) to mean that the window content is shown. Welcome thoughts on that.
2. The `border.color` and `border.width` only take effect when `fill` is set to `no` (it took me a while to work out why I couldn't get the border to show :smile:). I'm okay with this as an approach but we ought to describe it in the man page. As an after-thought consideration, would we want to ever support showing both fill AND border at the same time? If so that could easily be achieved by making show-border independent of the fill-boolean, and just set the default border.width to 0 or something like that.

It'd be good to conclude this before the next release to avoid regression.

tokyo4j:
Resolves #1735

As now whether to draw a filled rectangle or an outlined rectangle is independently configured, I added `snapping.overlay.[region|edge].stroke`.

However, I'm thinking of replacing `*.fill`/`*.stroke` with `*.bg.enabled`/`*.border.enabled`. I think this will make it easier to understand.

Another approach is removing `*.fill`/`*.stroke` and letting users disable them by setting `*.bg.color` to `#0000000` and `*.border.width` to `0`.

Give me your opinions.

### DIFF
--- a/docs/themerc
+++ b/docs/themerc
@@ -74,11 +74,11 @@ osd.window-switcher.preview.border.color: #dddda6,#000000,#dddda6
 osd.workspace-switcher.boxes.width: 20
 osd.workspace-switcher.boxes.height: 20
 
-snapping.preview.region.fill: yes
-snapping.preview.edge.fill: yes
-snapping.preview.region.bg.color: #8080b380
-snapping.preview.edge.bg.color: #8080b380
-snapping.preview.region.border.width: 1
-snapping.preview.edge.border.width: 1
-snapping.preview.region.border.color: #dddda6,#000000,#dddda6
-snapping.preview.edge.border.color: #dddda6,#000000,#dddda6
+snapping.overlay.region.fill: yes
+snapping.overlay.edge.fill: yes
+snapping.overlay.region.bg.color: #8080b380
+snapping.overlay.edge.bg.color: #8080b380
+snapping.overlay.region.border.width: 1
+snapping.overlay.edge.border.width: 1
+snapping.overlay.region.border.color: #dddda6,#000000,#dddda6
+snapping.overlay.edge.border.color: #dddda6,#000000,#dddda6
--- a/docs/themerc
+++ b/docs/themerc
@@ -74,8 +74,18 @@ osd.window-switcher.preview.border.color: #dddda6,#000000,#dddda6
 osd.workspace-switcher.boxes.width: 20
 osd.workspace-switcher.boxes.height: 20
 
-snapping.overlay.region.fill: yes
-snapping.overlay.edge.fill: yes
+# Default values for following options change depending on the rendering
+# backend. For software-based renderers, *.bg.enabled is "no" and
+# *.border.enabled is "yes" if not set. For hardware-based renderers,
+# *.bg.enabled is "yes" and *.border.enabled is "no" if not set.
+# Setting *.bg.enabled to "yes" for software-based renderer with translucent
+# background color may severely impact performance.
+#
+# snapping.overlay.region.bg.enabled:
+# snapping.overlay.edge.bg.enabled:
+# snapping.overlay.region.border.enabled:
+# snapping.overlay.edge.border.enabled:
+
 snapping.overlay.region.bg.color: #8080b380
 snapping.overlay.edge.bg.color: #8080b380
 snapping.overlay.region.border.width: 1
--- a/src/overlay.c
+++ b/src/overlay.c
@@ -1,62 +1,54 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include <assert.h>
-#include <wlr/render/pixman.h>
 #include "labwc.h"
 #include "overlay.h"
 #include "view.h"
+#include "theme.h"
 
 static void
-create_overlay_rect(struct seat *seat, struct overlay_rect *rect, int fill,
-		float bg_color[4], int border_width, float border_colors[3][4])
+create_overlay_rect(struct seat *seat, struct overlay_rect *rect,
+		struct theme_snapping_overlay *theme)
 {
 	struct server *server = seat->server;
 
-	rect->fill = fill;
-	/* For pixman renderer, always render outlines to save CPU resource */
-	if (wlr_renderer_is_pixman(server->renderer)) {
-		rect->fill = false;
+	rect->bg_enabled = theme->bg_enabled;
+	rect->border_enabled = theme->border_enabled;
+	rect->tree = wlr_scene_tree_create(&server->scene->tree);
+
+	if (rect->bg_enabled) {
+		/* Create a filled rectangle */
+		rect->bg_rect = wlr_scene_rect_create(
+			rect->tree, 0, 0, theme->bg_color);
 	}
 
-	if (rect->fill) {
-		/* Create a filled rectangle */
-		rect->scene_rect = wlr_scene_rect_create(&server->scene->tree,
-			0, 0, bg_color);
-		rect->node = &rect->scene_rect->node;
-	} else {
+	if (rect->border_enabled) {
 		/* Create outlines */
 		float *colors[3] = {
-			border_colors[0],
-			border_colors[1],
-			border_colors[2],
+			theme->border_color[0],
+			theme->border_color[1],
+			theme->border_color[2],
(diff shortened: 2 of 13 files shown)


docs/themerc
```diff
@@ -74,11 +74,20 @@ osd.window-switcher.preview.border.color: #dddda6,#000000,#dddda6
...
-snapping.preview.edge.bg.color: #8080b380
-snapping.preview.region.border.width: 1
-snapping.preview.edge.border.width: 1
-snapping.preview.region.border.color: #dddda6,#000000,#dddda6
-snapping.preview.edge.border.color: #dddda6,#000000,#dddda6
+# Default values for following options change depending on the rendering
+# backend. For software-based renderers, *.fill is "no" and *.stroke is "yes"
+# if not set. For hardware-based renderers, *.fill is "yes" and *.stroke is
+# "no" if not set.
+# Setting *.fill to "yes" for software-based renderer is highly discouraged.
```
Consolatis, line 81: Suggest something like "may severely impact performance" instead of "is highly discouraged".

Consolatis:
> However, I'm thinking of replacing `*.fill`/`*.stroke` with `*.bg.enabled`/`*.border.enabled`. I think this will make it easier to understand.

Potential options seem to look like something like this:
- single config option with something like `border|solid|combined|none`
- two config options with `yes|no`
- implied option (e.g. no specific option) by parsing `border.width` and `bg.color`

Of those 3 I prefer the 2nd one, e.g. two boolean config options.
Then the question is how they should be named and I slightly prefer your suggestion with `bg.enabled` and `border.enabled` as well. However, other opinions welcome.

Side note: the 1st commit body states `snapping.overlay.overlay` rather than `snapping.overlay.*`: 
> `snapping.preview.*` in themerc is now replaced with `snapping.overlay.overlay`

docs/themerc
tokyo4j: > I think this is also only an issue when using transparency.

Confirmed in my test. I didn't know that!
Seems like blending in CPU is much heavier operation than just drawing.

src/overlay.c
```diff
@@ -65,20 +80,20 @@ show_overlay(struct seat *seat, struct overlay_rect *rect, struct wlr_box *box)
...
-	if (!node) {
+	if (!rect->tree) {
 		overlay_reconfigure(seat);
-		node = rect->node;
-		assert(node);
+		assert(rect->tree);
 	}
 
-	if (rect->fill) {
+	if (rect->scene_rect) {
```
tokyo4j, line 89: I don't know why I made this change. This segfaults.

src/overlay.c
```diff
@@ -1,62 +1,78 @@
...
 
-	rect->fill = fill;
-	/* For pixman renderer, always render outlines to save CPU resource */
-	if (wlr_renderer_is_pixman(server->renderer)) {
-		rect->fill = false;
+	/*
+	 * Default settings is:
+	 * pixman renderer: bg=enabled, border=disabled
+	 * Other renderers: bg=disabled, border=enabled
+	 */
```
Consolatis, line 20: This comment seems the wrong way around.

src/overlay.c
```diff
@@ -1,62 +1,78 @@
...
+	if (bg_enabled == INT_MIN) {
+		rect->bg_enabled = !wlr_renderer_is_pixman(server->renderer);
+	} else {
+		rect->bg_enabled = bg_enabled;
 	}
+	if (border_enabled == INT_MIN) {
+		rect->border_enabled = wlr_renderer_is_pixman(server->renderer);
+	} else {
+		rect->border_enabled = border_enabled;
+	}
```
Consolatis, line 30: I am not sure about this, seems to fit better into `theme.c` within `post_processing()` and supplying `server` as argument to `theme_init()`.

tokyo4j: How about having `struct server *server` in `struct theme`?

Consolatis: I think having `server` be an argument to `theme_init()` is cleaner (and also matches better with the other stuff we call on `Reconfigure`, see `reload_config_and_theme`).

I am not sure if additionally having `server` in `struct theme` really improves things much. Note that there is also `rc.theme` globally available already (with the theme pointer being there maybe not being the best idea) so `server` would then (indirectly) be globally available as well via `rc.theme->server` which might lead to some sloppy code in the future. Thoughts @johanmalm ?

tokyo4j: I added `server` to `theme_init()` and `theme_builtin()` and am pretty happy with it.
I initially thought we need to add `server` to every function in `theme.c` and was hesitant about it, but it turned out only `theme_builtin()` needs `server`.

src/theme.c
```diff
@@ -754,29 +766,35 @@ entry(struct theme *theme, const char *key, const char *value)
 	if (match_glob(key, "osd.label.text.color")) {
 		parse_hexstr(value, theme->osd_label_text_color);
 	}
-	if (match_glob(key, "snapping.preview.region.fill")) {
-		theme->snapping_preview_region_fill = parse_bool(value, true);
+	if (match_glob(key, "snapping.overlay.region.bg.enabled")) {
+		theme->snapping_overlay_region_bg_enabled = parse_bool(value, -1);
```
tokyo4j, line 770: This is wrong. `-1` is not boolean.

src/overlay.c
```diff
@@ -1,62 +1,62 @@
...
 #include "labwc.h"
 #include "overlay.h"
 #include "view.h"
 
 static void
-create_overlay_rect(struct seat *seat, struct overlay_rect *rect, int fill,
-		float bg_color[4], int border_width, float border_colors[3][4])
+create_overlay_rect(struct seat *seat, struct overlay_rect *rect,
+		bool bg_enabled, float bg_color[4], bool border_enabled,
+		int border_width, float border_colors[3][4])
```
Consolatis, line 10: If we use a substruct in `theme.h` which defines an overlay and then just have a `region` or `edge` overlay pointer would simplify the args and the caller quite a bit. Not a requirement for me though.

src/overlay.c
johanmalm: > I am not sure if additionally having `server` in `struct theme` really improves things much. Note that there is also `rc.theme` globally available already (with the theme pointer being there maybe not being the best idea) so `server` would then (indirectly) be globally available as well via `rc.theme->server` which might lead to some sloppy code in the future. Thoughts @johanmalm ?

Agree because it invites bad design patterns. If it's proving hard (such as in callbacks, I think it's preferable to make it static variable within a translation unit (as we've done with `g_server` in files like `session-lock.c`. It avoids having a global `server`. 

For simplicity / readability, we made an exception with `rc` (and therefore also `rc.theme`) because they are set once at the start and are used so widely.

See #1744

src/overlay.c
tokyo4j: Good idea. Done.

Consolatis:
- [x] tested with pixman and gles2 and defaults work fine
- [x] Reconfigure support also seems to to work as expected
- [x] Having both enabled on pixman also worked

LGTM

johanmalm:
Really nice. Thanks.

tokyo4j:
Thank you!

Issue #1446, opened 2024-01-16
Add window rule to prevent X11 clients from sizing and positioning themselves


bdantas:
I'm on labwc 0.7.0 and have various window rules in my `~/.config/labwc/rc.xml`. The ResizeTo and MoveTo rules work for wayland applications (`[xdg-shell]` in `Alt+Tab` menu) but not for xwayland applications (`[xwayland]` in `Alt+Tab` menu).

I'm not sure if this is bug or a known limitation due to how labwc handles xwayland windows.

johanmalm:
I've just tried `MoveTo` with xterm and it works fine here.

bdantas:
Maybe my syntax is wrong? I'm trying to move Thunderbird (AppImage of a gtk2 version) using this rule in `~/.config/labwc/rc.xml`, just as an example:
```
<windowRules>
	<windowRule identifier="" title="Thunderbird"><action name="MoveTo" x="10" y="10" /></windowRule>
</windowRules>
```
Definitely not working for me. The window opens in the center of my laptop's screen. What gives?

bdantas:
The syntax is correct and it works with xterm. 

Hmm, with Thunderbird (xwayland app, gtk2 widget library) and with Xfe (xwayland, fox widget library) `MoveTo` does not work at all. Any ideas?

johanmalm:
Are you sure the identifier / title is correct. Might not work with `identifier=""`
Run `labwc -d` to see the identifier + title when you launch (map) an application.

Consolatis:
Just a guess, maybe the title is not correct during startup and thus the `Thunderbird` rule doesn't match.
You could try to use the identifier rather than the title, see `wlrctl window list`.

bdantas:
Thank you both.
```
$ wlrctl window list
foot: foot
Thunderbird: Inbox - Mozilla Thunderbird
```
```
	<windowRule identifier="Thunderbird"><action name="ResizeTo" width="900" height="550" /><action name="MoveTo" x="10" y="10" /></windowRule>
```
Now it works!

I assumed the output of `wlrctl window list` were titles. What's the difference between "identifier" and "title" in this context?

johanmalm:
> I assumed the output of `wlrctl window list` were titles. What's the difference between "identifier" and "title" in this context?

Depends on the client. If you run `labwc` nested (from the terminal) you can see what they are.

bdantas:
[...]

PR #1742, opened 2024-04-21
window-rules: add ignoreConfigureRequest

Consolatis:
This allows to ignore X11 client side configure requests like positioning and resizing and can be used to fight some X11 applications that persist to have their windows spawn at specific places and sizes.

Fixes:
- #1446

---

I don't really like the runtime `strcmp()` calls with the rules, at some point we could consider moving those to enums instead.


johanmalm:
Thanks.

PR #1744, opened 2024-04-21
server: refactor to avoid *g_server

johanmalm:


Consolatis:
:)

johanmalm:
We could probably do something similar in `session-lock.c`, but might take a bit more refactoring :smile:

PR #1745, opened 2024-04-21
Resolve errors from clangd

tokyo4j:
I switched to clangd from VSCode's intellisense today and it showed me some errors regarding headers.

- Added headers like <stddef.h> for `NULL`, <wayland-util.h> for `wl_link` and <stdbool.h> for `bool`.
- Fixed circular dependency of `labwc.h` and `ime.h`

### DIFF
--- a/include/action.h
+++ b/include/action.h
@@ -2,6 +2,7 @@
 #ifndef LABWC_ACTION_H
 #define LABWC_ACTION_H
 
+#include <stdbool.h>
 #include <wayland-util.h>
 
 struct view;
--- a/include/button/common.h
+++ b/include/button/common.h
@@ -2,6 +2,8 @@
 #ifndef LABWC_BUTTON_COMMON_H
 #define LABWC_BUTTON_COMMON_H
 
+#include <stddef.h>
+
 /**
  * button_filename() - Get full filename for button.
  * @name: The name of the button (for example 'iconify.xbm').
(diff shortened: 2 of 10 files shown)


Consolatis:
Thanks. Can / should we add this to the CI?

tokyo4j:
> Thanks. Can / should we add this to the CI?

Maybe we should, but I currently don't know how to do it from command line.

johanmalm:
LGTM
Fancy self-merge?

johanmalm:
Oh - too late :smile:

Consolatis:
> Oh - too late 😄

Sorry :p
Took the liberty to slightly change the commit subject.

tokyo4j:
> > Thanks. Can / should we add this to the CI?
> 
> Maybe we should, but I currently don't know how to do it from command line.

Actually, how I applied clangd for header files is a bit tricky.

Meson creates `compile_commands.json` and clangd use it to analyze source files.
However, `compile_commands.json` has only entries of "how to compile a `*.c` file". So when I open a header file in VSCode, clangd cannot find an matched entry in `compile_commands.json` and chooses a wrong entry (`cursor-shape-v1-protocol.c` in my case).
So as a workaround, I put only the entry for `desktop.c` in `compile_commands.json` to prevent clangd from picking up wrong entries. It works for my use case as labwc uses the same compiler flags against all source files.  

I'll make another PR if I find a maintainable way to integrate this into CI, but I don't know how to do it for now.

Issue #1739, opened 2024-04-20
crash on Kill action with xwayland views


xi:
labwc crashes when I use the Kill action on xwayland views (tested with firefox, thunderbird, and xeyes). It does not crash on native wayland clients or when I manually call `pkill -SIGTERM firefox`.

Consolatis:
Hrm, right. We just kill the pid of the wayland client which in this case is xwayland itself.

So there are two issues here:
- get the right pid (pid of the unix connection to the wayland socket for wayland native clients or `->pid` for xwayland ones)
- figure out why labwc is crashing when xwayland is killed

xi:
Thanks for the quick fix!

PR #1746, opened 2024-04-22
Fix pid lookup for the Kill action

Consolatis:
Before this patch, labwc would happily kill itself when the user called the `Kill` action when any xwayland view had focus.

The reason this happened was that wlroots creates the xwayland wayland client via socketpair() and thus a lookup of the pid of the socket connection would return the pid of labwc itself.

This patch fixes that by implementing different pid lookup mechanisms based on the view implementation backend.

Fixes:
- #1739


johanmalm:
LGTM. Nice work.

PR #1747, opened 2024-04-22
Tweaks to .editorconfig

johanmalm:


tokyo4j:
Oh, I didn't know that commit messages should be within 74 characters, even though It's noted in `CONTRIBUTING.md`.
I've wrapped them at 80 characters. Sorry.

Issue #1750, opened 2024-04-24
menu press-drag-release doesn't select the menu item


ec1oud:
I'm very used to it working in openbox, so it feels really awkward trying to switch to labwc that I have to actually right-click, leave the menu open, and then click again on a menu item.

This applies to the desktop application menu, and menus in Qt applications (not sure if that could be a Qt bug), but not GTK applications.

tokyo4j:
Oops, I didn't know mentioning an GitHub issue in commit message and pushing it always adds an item.

Consolatis:
> Oops, I didn't know mentioning an GitHub issue in commit message and pushing it always adds an item in the issue.

Yeah that is a bit annoying but IMHO a GH UI issue and should not prevent us from working the way we want.

> and menus in Qt applications (not sure if that could be a Qt bug), but not GTK applications.

I vaguely remember some work around this topic about a year or so ago. I think it had something to do with popup grabs and Qt and GTK using different approaches for this is exact scenario. So it \*might\* be something that we could look into from the labwc side. This may also be different for wayland native and for xwayland windows.

tokyo4j:
Reopened for Qt menu issue

ec1oud:
Thanks for the quick response!  Yes that took care of the desktop menu.  In Qt widget applications it seems that the press-drag-release works the first time, but not subsequent times.

ec1oud:
But with the same Qt application on Sway, the menu works fine every time.

ec1oud:
I think the problem is Qt needs to see correct mouse button state for every mouse event, including mouse moves.  Don't assume that if you told us at the time that the left mouse button was pressed, that we will remember that it's still pressed while the user is dragging.  Reporting complete state in every event makes software more robust, in case some event went missing or there is some sort of bug on some system.

tokyo4j:
The difference between labwc and sway is that labwc sends `wl_pointer.enter` to notify the client that the cursor focus is moved to the popup surface, while sway doesn't. So at least by applying the following patch labwc never sends `wl_pointer.enter` in this case and press-move-release selection always works:
```diff
diff --git a/src/input/cursor.c b/src/input/cursor.c
index df03dcf2..e89229c5 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -326,6 +326,7 @@ input_inhibit_blocks_surface(struct seat *seat, struct wl_resource *resource)
 static bool
[...]

PR #1751, opened 2024-04-24
cursor: do actino/close menu in successive press & release

tokyo4j:
This PR just removes `close_menu` in `cursor.c`.
I'm missing the context of it and will close this PR if the current behavior is intentional and better.

Fixes: #1750


Consolatis:
We used to handle menu actions on press and thus needed a way to eat the corresponding release event. Now we do handle the menu actions on release (to allow for example `wtype` to work) and seem to have missed removing this part.

LGTM, not tested.

Consolatis:
Nitpick: commit subject has a spelling error: `actino`

jp7677:
to be honest. this change makes it pretty difficult/awkward to use e.g. the root menu with a stylus. The former behavior with one tap to open and a second tap to hit a menu item fits much better imho.

Consolatis:
Just tested again with the mouse and initially clicking, releasing worked and shows the menu just as before and allows using another click on some item.

However, if clicking, moving the mouse outside of the menu and then releasing the menu closes.

Is this the issue you are referring to?

tokyo4j:
Sorry, I didn't realize clicking without any movement is difficult for some devices. I think @Consolatis's fix would be the best solution. Closing menu in single press&release would have no advantages for most users.

jp7677:
> EDIT: But I'm afraid that, if the small movement while clicking with stylus is in right-bottom direction, an action is unexpectedly triggered.

Yes, this. Above patch does improves the behavior, but due the sensitivity of the stylus, half of the time I'm still triggering the first menu entry by accident.

tokyo4j:
I tested how OpenBox handles this.
It seems to use a timer and run action or close menu if enough time has been passed after the cursor is pressed. Maybe we should take this behavior.

Consolatis:
> > EDIT: But I'm afraid that, if the small movement while clicking with stylus is in right-bottom direction, an action is unexpectedly triggered.
> 
> Yes, this. Above patch does improves the behavior, but due the sensitivity of the stylus, half of the time I'm still triggering the first menu entry by accident.

Hm.. this might also happen from some mice with high resolution. But I also see the benefit of the single press-select-release behavior.

> It seems to use a timer and run action or close menu if enough time has been passed after the cursor was pressed. Maybe we should copy this behavior.

That might be an option as well. Maybe we could even use a timestamp rather than a bool for `close_menu`? That way we wouldn't need an actual timer here.

jp7677:
Or make this configurable, though this comes ofc with the burden of another configuration option.

Consolatis:
> Or make this configurable, though this comes ofc with the burden of another configuration option.

If we can find a way to solve this internally I'd very much prefer that. (also: updated the diff a 2nd time, sorry).

tokyo4j:
> Maybe we could even use a timestamp rather than a bool for close_menu?

Yeah, I think this is the way to go.

jp7677:
> It is already kind of user configurable right now btw: using `<mousebind button="Left" action="Click">` rather then `Press`. This is done by default for the window menu so that one should work as before using the stylus. We should still implement the timestamp idea though and if that works well we have the best of both worlds.

Thanks, that does the trick as well. Cool. 
But that said, not sure if I hit an unrelated issue here, I have to remove the default mouse bindings, otherwise the new binding has no effect (with just changing `Press` to `Click`).

tokyo4j:
Maybe we should support `<menu><hideDelay>`.
> hideDelay How long (in milliseconds) you have to hold the mouse button down for it to be hidden automatically when you release it. If you hold shorter, it will stay up when you release.

http://openbox.org/wiki/Help:Configuration#Menu

Consolatis:
> But that said, not sure if I hit an unrelated issue here, I have to remove the default mouse bindings, otherwise the new binding has no effect (with just changing `Press` to `Click`).

Yes, you can theoretically have a `Press` and `Click` event with all other arguments being equal for the same mouse context.
So just defining a `Click` event won't remove an existing `Press` one (which would be created due to `<default />`).



> Maybe we should support `<menu><hideDelay>`.

And then also use it for the execute?
Nice find! Sounds exactly like what we need anyway.

tokyo4j:
> And then also use it for the execute?

Strangely though, Openbox already uses `<menu><hideDelay>` for running action.
The logic for handling mouse button with menu opened is [here](https://github.com/danakj/openbox/blob/master/openbox/event.c#L1829-L1855).
Action is executed or menu is closed if the timer timeouted (`menu_hide_delay_reached()` returns true).
I think Openbox's naming and documentation is bad.

johanmalm:
IIRC the hideDelay setting in Openbox is really a show-delay or a hover-timer. The purpose of it is to delay the opening of a submenu when the user moves the selection (by cursor or keyboard) to a submenu entry.

It's how menus commonly work, for example with Gtk and Qt.

It affects the UX in two notable ways:

1. When moving up/down fast across many submenu entries, the submenus do not open until you actually stops on one
2. You can move diagonally from the submenu entry to its associated submenu without other submenus opening should you "accidentally" glide across some submenu entries.

Consolatis:
> IIRC the hideDelay setting in Openbox is really a show-delay or a hover-timer. The purpose of it is to delay the opening of a submenu when the user moves the selection (by cursor or keyboard) to a submenu entry.

Isn't that [submenuShowDelay](https://github.com/labwc/labwc-scope?tab=readme-ov-file#211-configuration---menu)?

johanmalm:
> Isn't that [submenuShowDelay](https://github.com/labwc/labwc-scope?tab=readme-ov-file#211-configuration---menu)?

Yes, sorry. Got confused.

PR #1759, opened 2024-04-26
docs: Fixed some typos, 80 column lines (where possible)

domo141:


domo141:
the 'mono space' -> 'monospace' is something I am not sure about (or should it be mono-space)...

also, some tabs after non-tab changed to spaces -- noticed when manpage spaced 'weirdly'.
those tabs were probably introduced by me in previous 80-col-filling commits i've been sending.
`grep -nP '[^\t]\t' {files}` helped there.

Consolatis:
> the 'mono space' -> 'monospace' is something I am not sure about (or should it be mono-space)...

I think `monospace` is fine.

> also, some tabs after non-tab changed to spaces -- noticed when manpage spaced 'weirdly'. those tabs were probably introduced by me in previous 80-col-filling commits i've been sending. `grep -nP '[^\t]\t' {files}` helped there.

Nice catch! The Github UI seems somewhat lacking in this regard.
Maybe we should just throw that into the CI and additionally use `-R` to scan `include/` `src/` and `docs/`.

LGTM.

johanmalm:
Oh I like that ^
Yes, we ought to include it.

PR #1760, opened 2024-04-26
cursor: implement <menu><hideDelay>

tokyo4j:
This fixes the UX degradation introduced by #1751 that the menu is closed or an action is executed unexpectedly when the cursor is moved a bit while clicking, by requiring a delay to close the menu or execute an action in a single sequence of button press & release.

I'm not sure just copying the name `hideDelay` from Openbox is the best option, as it also delays running action.

As we're in cooldown period, maybe we should postpone this fix and revert #1751 for now.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -891,6 +891,24 @@ situation.
 	can be caused by *<margin>* settings or exclusive layer-shell clients
 	such as panels.
 
+## MENU
+
+```
+<menu>
+  <ignoreButtonReleasePeriod>250</ignoreButtonReleasePeriod>
+</menu>
+```
+
+*<menu><ignoreButtonReleasePeriod>*
+	How long (in milliseconds) the initial button release event is ignored
+	for. The reason for this logic and behaviour is to avoid a fast
+	press-move-release sequence indended to just open the menu resulting in
+	the closure of the menu or the selection of (typically the first) menu
+	item. This behaviour only affects the first button-release. It is not
+	anticipated that most users will want to change this, but the config
+	option has been exposed for unusual use-cases. It is equivalent to
+	Openbox's `<hideDelay>`. Default is 250 ms.
+
 ## ENVIRONMENT VARIABLES
 
 *XCURSOR_THEME* and *XCURSOR_SIZE* are supported to set cursor theme
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -586,4 +586,7 @@
     </windowRules>
   -->
 
+  <menu>
+    <ignoreButtonReleasePeriod>250</ignoreButtonReleasePeriod>
+  </menu>
 </labwc_config>
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -929,12 +929,17 @@ handle_press_mousebinding(struct server *server, struct cursor_context *ctx,
 	return consumed_by_frame_context;
 }
 
+static uint32_t press_msec;
+
 static void
 cursor_button_press(struct seat *seat, uint32_t button,
 		enum wlr_button_state button_state, uint32_t time_msec)
 {
 	struct server *server = seat->server;
 	struct cursor_context ctx = get_cursor_context(server);
+
+	/* Used on next button release to check if it can close menu or select menu item */
+	press_msec = time_msec;
 
 	/* Determine closest resize edges in case action is Resize */
 	uint32_t resize_edges = cursor_get_resize_edges(seat->cursor, &ctx);
@@ -946,6 +951,11 @@ cursor_button_press(struct seat *seat, uint32_t button,
 	}
 
 	if (server->input_mode == LAB_INPUT_STATE_MENU) {
+		/*
+		 * If menu was already opened on press, set a very small value
+		 * so subsequent release always closes menu or selects menu item.
+		 */
+		press_msec = 0;
 		return;
 	}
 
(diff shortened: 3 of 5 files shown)


Consolatis:
> As we're in cooldown period, maybe we should postpone this fix and revert #1751 for now.

I am easy on this as this is a fix. Ultimately up to @johanmalm to decide.

> I'm not sure just copying the name `hideDelay` from Openbox is the best option, as it also delays running action.

Agree, this is kind of weird. Not quite sure how to name it instead.

src/input/cursor.c
```diff
@@ -1011,12 +1026,14 @@ cursor_button_release(struct seat *seat, uint32_t button,
 	seat_reset_pressed(seat);
 
 	if (server->input_mode == LAB_INPUT_STATE_MENU) {
-		if (ctx.type == LAB_SSD_MENU) {
-			menu_call_selected_actions(server);
-		} else {
-			menu_close_root(server);
-			cursor_update_common(server, &ctx, time_msec,
-				/*cursor_has_moved*/ false);
+		if (close_menu == 0 || time_msec - close_menu > rc.menu_hide_delay) {
```
Consolatis, line 1029: Could be `!close_menu` or we could just remove the condition here, `time_msec - 0` should always be `> rc.menu_hide_delay`.

tokyo4j:
A completely different approach would be requiring a certain amount of cursor movement (instead of delay) to close menu or execute action.

To test it, I made a [quick prototype](https://github.com/tokyo4j/labwc/tree/drag-menu-distance).

I don't have any preference at the moment though.

johanmalm:
Thanks for researching and coming up with solutions.

> As we're in cooldown period, maybe we should postpone this fix and revert #1751 for now.

Let's forward fix. I'll test both approaches - but probably tomorrow.

docs/labwc-config.5.scd
```diff
@@ -890,6 +890,18 @@ situation.
...
+
+```
+<menu>
+  <hideDelay>200</hideDelay>
+</menu>
+```
+
+*<menu><hideDelay>*
+	How long you have to hold the mouse button down to close the menu or to
+	execute an action when you release it. Default is 200ms.
```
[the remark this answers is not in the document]
tokyo4j: Fixed.

src/input/cursor.c
tokyo4j: > I never verified if time_msec is actually the current time in ms, it *might* just be a serial.

I haven't either, but this worked fine in tests with x11, wayland, libinput+drm backends.

johanmalm:
Thanks for working on this.

I suggest we go for the time-based approach because (a) it has worked on Openbox; and (b) I wouldn't be surprised if touch events (emulating cursor motion) have a large sideway movement associated with one click.

- [ ] We don't often deviate from Openbox spec, but `<hideDelay>` is pretty weird and I don't think many users actually every change that value to I suggest using something like `<ignoreButtonReleasePeriod>` instead.
- [ ] Openbox uses 250 ms, so vote for copying that.
- [ ] Did we consider overflow (i.e. `time_msec < press_msec` after `time_msec` roll-over)? I'm not sure if it's worth dealing with, but maybe we ought to just make a comment about it if we don't.
- [ ] `menu_opened_at` is okay, but could consider `press_msec` (just enjoy brainstorming variable names).


Suggested words for the man page:


```																			x
*<menu><ignoreButtonReleasePeriod>*
	How long (in milliseconds) the initial button release event is ignored
	for. The reason for this logic and behaviour is to avoid a fast
	press-move-release sequence resulting the closure of the menu or the
	selection of (typically the first) menu item. This behaviour only affects
	the first button-release. It is not anticipated that most users will want
	to change this, but the config option has been exposed for unusual
	use-cases. It is equivalent to Openbox's `<hideDelay>`.
	Default is 250 ms.
```

Ref:
- http://openbox.org/wiki/Help:Configuration#Menu

Consolatis:
This is actually happening even without movement when the menu opens above the cursor (e.g. when clicking on the bottom of the screen) so we should really fix that. Should have tested that scenario.

johanmalm:
> The immediate-close is actually happening even without movement when the menu opens above or on the left of the cursor (e.g. when clicking on the bottom or right of the screen) so we should really fix that. Should have tested that scenario.

...by positioning the menu 1 pixel further to the left?

johanmalm:
> Yeah, repositioning the menu would work around the issue and it is at least deterministic (e.g. always under the cursor or never under the cursor). But the actual issue here is the immediate closing of the menu

Ah yes, of course :smile: Got it now.

Consolatis:
Just so we are on the same page, I think we should still fix the positioning of the menu in the long term but I don't think its that important.

tokyo4j:
- Added `indended to just open the menu` to the suggested man page.
- Added `/* TODO: take into account overflow of time_msec */`.
  - The timestamp is the uptime in my system, so it will take 50 days to overflow.

docs/labwc-config.5.scd
```diff
@@ -890,6 +890,24 @@ situation.
 	can be caused by *<margin>* settings or exclusive layer-shell clients
 	such as panels.
 
+## MENU
+
+```
+<menu>
+  <hideDelay>200</hideDelay>
+</menu>
```
Consolatis, line 898: This is referring to the openbox config name.

docs/labwc-config.5.scd
```diff
@@ -890,6 +890,24 @@ situation.
...
+```
+
+*<menu><ignoreButtonReleasePeriod>*
+	How long (in milliseconds) the initial button release event is ignored
+	for. The reason for this logic and behaviour is to avoid a fast
+	press-move-release sequence indended to just open the menu resulting the
+	closure of the menu or the selection of (typically the first) menu item.
```
Consolatis, line 905: resulting \*in\* the closure

docs/rc.xml.all
```diff
@@ -586,4 +586,7 @@
     </windowRules>
   -->
 
+  <menu>
+    <hideDelay>200</hideDelay>
+  </menu>
```
Consolatis, line 591: openbox config name

docs/labwc-config.5.scd
tokyo4j: Should have run `hideDelay`. Thanks.

docs/rc.xml.all
tokyo4j: Fixed.

docs/labwc-config.5.scd
```diff
@@ -890,6 +890,24 @@ situation.
 	can be caused by *<margin>* settings or exclusive layer-shell clients
 	such as panels.
 
+## MENU
+
+```
+<menu>
+  <ignoreButtonReleasePeriod>200</ignoreButtonReleasePeriod>
```
Consolatis, line 897: This is still stating `200` ms, same for the `rc.xml.all` snippet. Sorry, missed that in the earlier review.
With that fixed this LGTM so feel free to Rebase and Merge. Thanks!

tokyo4j: Done. Thank you!

Issue #1756, opened 2024-04-25
Awkward keybinding problem


spl237:
In RPiOS, ever since we started customising the desktop, we have had the Windows key on the keyboard open and close the main application launch menu.

I have added a keybinding to our default labwc.config on Super_L, which generates the relevant call to the panel application to show or hide the menu, which works fine.

However, there are default keybindings which use the Windows key as a modifier - W+cursor snaps windows to the relevant half of the screen, for example. We'd like to keep these, as people use them, but the problem is that the initial keybinding on Super_L is triggered first, so the main menu opens (and stays open) when the Windows key is pressed and held, and then the window snaps when a cursor key is subsequently pressed.

Is there any way of making keybindings which are set up on modifier keys only not take effect until the modifier key is released if no other keys have been pressed in the interim?

This would allow me to use the Windows key in this way. So the actions would be:

Windows key pressed - Windows key released - binding on Super_L is triggered.
Windows key pressed - cursor key pressed - binding on Win + cursor is triggered, subsequent release of Windows key is ignored.

It's not a showstopper, but if keybindings on modifiers alone are allowed - and it seems they are - then at present they can trigger this multiple binding behaviour.

I'm happy to have a look at fixing this myself if it is thought that it would be useful, if someone can just give me a pointer in the right sort of direction! Many thanks in advance...

Consolatis:
This is something that has come up before. I am not sure how we could handle that due to us processing the keybinds on press.
Maybe we could add an optional `on="release"` keybind argument or something like that.. And then check if no other key is currently pressed. However, that would then still trigger when pressing e.g. `W-Left`, then releasing `Left` and finally releasing `W`.

spl237:
OK, if others have found this a problem as well, I'll have a poke about in the code and see if I can think of a solution...

stefonarch:
> Maybe we could add an optional `on="release"` keybind argument or something like that..

We did something similar in `lxqt-panel` to allow super/winkey to open the menu.

spl237:
Possible fix suggested in PR - works for me. No need to explicitly add an on-release flag to keybinds - I just assume that any modifiers-only keybind is to trigger on release, all others trigger on press. Feel free to try - comments welcome!

Issue #351, opened 2022-05-29
Modifier-only keybind


skerit:
I'm missing the ability to configure a modifier-only keybind.
I'd like to perform an action just after I've pressed and released the Super key on its own, but the config currently seems to ignore that.

01micko:
Did you try `Super_L` ?

This works in my `rc.xml` to open the menu:

```xml
    <keybind key="Super_L">
      <action name="ShowMenu"><menu>root-menu</menu></action>
    </keybind>
```

What version of `labwc` are you on?

01micko:
@Flrian interesting, sounds like a bug!

johanmalm:
In accordance with Openbox spec, `<keyboard><keybind>` does not support an action attribute to specify "press" or "release".
http://openbox.org/wiki/Help:Bindings#Key_bindings

As a side-note, `<mousebind>` does.

I guess, for openbox/blackbox/fluxbox/etc that why applications such as [ksuperkey](https://github.com/hanschen/ksuperkey) and 
[xcape](https://github.com/alols/xcape) exist.

So, it's not a bug but rather a requirement we haven't yet considered.
We could consider adding it as a section 8 requirement
https://github.com/labwc/labwc-acceptance-criteria/blob/main/Acceptance-criteria.md

Consolatis:
Using `release` for keybinds is tricky because we don't want to send the corresponding `press` event to the currently focused application. If we instead trigger on `press` we can store that key and then prevent the corresponding `release` to be sent.

I guess it could be done but would require to slow down the keyboard input by waiting for the `release` event (in case it would actually trigger some compositor keybinding) and only then either send both at once or none of them.
That would then cause some troubles with applications that actually depend on receiving the `press` and `release` events at proper times (like crouching via `CTRL`  in games). That in turn could possibly be worked around by using some timer that prevents holding onto the `press` event for longer than the timer runs, but all of this would still increase input latency (by the configured time span of the timer) and complicate the whole input handling.

So all in all I don't see much benefit of implementing a `release` handler for keybinds as in my opinion the costs don't outweigh the benefits.

skerit:
I don't expect a lot of people would want this to apply to keys like Control or Alt. Only Super and maybe the Fn key, right?
So could it be limited to those keys?

Also: there's no need to delay anything if no keybind for it exists, right? Then any negative side effect like that is up to the user's configuration.

johanmalm:
[...]

PR #1761, opened 2024-04-26
Only respond to modifier-only keybindings on release

spl237:
This is a possible fix for https://github.com/labwc/labwc/issues/1756

I have modified the keybinding code so that a flag is set in the keybinding data structure if the keybinding only involves modifier keys.

If this flag is set for a particular keybinding, then that keybinding is detected on key press and stored, and executed on key release if no other (non-modifier) keybinding has executed in the meantime.

This fixes my immediate problem, that of wanting Win+cursor to activate snap behaviour, and Win on its own to launch the application menu, but I have no idea if this is the best way to do this. It seems to work for me, though!

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -38,6 +38,8 @@ struct keyinfo {
 
 static bool should_cancel_cycling_on_next_key_release;
 
+struct keybind *cur_keybind = NULL;
+
 static void
 change_vt(struct server *server, unsigned int vt)
 {
@@ -207,8 +209,8 @@ process_syms:
 	return NULL;
 }
 
-static bool
-is_modifier_key(xkb_keysym_t sym)
+bool
+keyboard_is_modifier_key(xkb_keysym_t sym)
 {
 	switch (sym) {
 	case XKB_KEY_Shift_L:   case XKB_KEY_Shift_R:
@@ -273,7 +275,7 @@ get_keyinfo(struct wlr_keyboard *wlr_keyboard, uint32_t evdev_keycode)
 	keyinfo.is_modifier = false;
 	for (int i = 0; i < keyinfo.translated.nr_syms; i++) {
 		keyinfo.is_modifier |=
-			is_modifier_key(keyinfo.translated.syms[i]);
+			keyboard_is_modifier_key(keyinfo.translated.syms[i]);
 	}
 
 	return keyinfo;
@@ -407,7 +409,14 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 		keyinfo.is_modifier);
 
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-		return handle_key_release(server, event->keycode);
+		if (cur_keybind && cur_keybind->mod_only) {
+			key_state_store_pressed_key_as_bound(event->keycode);
+			actions_run(NULL, server, &cur_keybind->actions, 0);
+			cur_keybind = NULL;
+			return true;
+		} else {
+			return handle_key_release(server, event->keycode);
+		}
 	}
 
 	/* Catch C-A-F1 to C-A-F12 to change tty */
@@ -443,16 +452,17 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 	/*
 	 * Handle compositor keybinds
 	 */
-	struct keybind *keybind =
+	cur_keybind =
 		match_keybinding(server, &keyinfo, keyboard->is_virtual);
-	if (keybind) {
+	if (cur_keybind && !cur_keybind->mod_only) {
 		/*
 		 * Update key-state before action_run() because the action
 		 * might lead to seat_focus() in which case we pass the
 		 * 'pressed-sent' keys to the new surface.
 		 */
 		key_state_store_pressed_key_as_bound(event->keycode);
-		actions_run(NULL, server, &keybind->actions, 0);
+		actions_run(NULL, server, &cur_keybind->actions, 0);
+		cur_keybind = NULL;
 		return true;
 	}
 
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -38,7 +38,7 @@ struct keyinfo {
 
 static bool should_cancel_cycling_on_next_key_release;
 
-struct keybind *cur_keybind = NULL;
+struct keybind *cur_keybind;
 
 static void
 change_vt(struct server *server, unsigned int vt)
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -412,7 +412,7 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 		if (cur_keybind && cur_keybind->on_release) {
 			key_state_bound_key_remove(event->keycode);
 			if (seat->active_client_while_inhibited
-					|| seat->server->session_lock) {
+					|| seat->server->session_lock_manager->locked) {
 				cur_keybind = NULL;
 				return true;
 			}
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -415,7 +415,6 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 				cur_keybind = NULL;
 				return false;
 			}
-			key_state_store_pressed_key_as_bound(event->keycode);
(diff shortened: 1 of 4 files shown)


src/input/keyboard.c
```diff
@@ -407,7 +409,14 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 		keyinfo.is_modifier);
 
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-		return handle_key_release(server, event->keycode);
+		if (cur_keybind && cur_keybind->mod_only) {
+			key_state_store_pressed_key_as_bound(event->keycode);
+			actions_run(NULL, server, &cur_keybind->actions, 0);
```
Consolatis, line 414: This would allow modifier only keybinds to be execute when the session is locked.

Consolatis:
I am also not sure the key tracking is correct here, if we eat the release event but let the press event through to the focused client that will likely result in the client repeating the key on a timer.

src/input/keyboard.c
spl237: OK, that's not an area I understand - can you please explain a bit more, and I will see if I can fix it?

spl237:
Agree this shouldn't be in 0.7.2 - it's a quick hack I put together this morning, and needs a lot of testing for corner cases etc, but if the general approach is ok, I'm happy to look at improvements.

src/input/keyboard.c
Consolatis: The check for being locked (or keyboard inhibited) comes a bit further below as that was only be relevant for the press events before. It might be enough to just guard the `actions_run()` and `key_state_store_pressed_key_as_bound()` with a `if (!>server->session_lock && !seat->active_client_while_inhibited)`.

spl237: Ah yes - I see what you mean, thanks. I'll change accordingly.

spl237:
> I am also not sure the key state tracking is correct here, if we eat the release event but let the press event through to the focused client that will likely result in the client repeating the key on a timer.

I'm not sure I completely understand the key state tracking, but in an attempt to deal with this, I have modified the code so that the stored key state is only updated on press and not on release - does this look like a better approach?

Interestingly, it has also meant that a modifier-only keybinding is only executed if the key is released relatively quickly, not if it is held down for more than a second or so, which I think is actually better.

Consolatis:
> There is also another, older issue for modifier-only keybinds #351 ^ Just remembered that one and thought I'd mention it, so it gets closed if this PR fixes the issue.

For some reason I can't link that issue to the PR (it doesn't show up in the link issue popup). Well, guess we'll have to manually close it then once this is merged.

ahesford:
Linked. The issue just had to be pinged to convince GitHub that it was still active.

johanmalm:
Sorry it's taken a while to get to this one. Just working down the list.

What's the best way of testing it? I tried binding to `W` but can't get it to work.

```
    <keybind key="W">
      <action name="Execute" command="rofi -show drun"/>
    </keybind>
```

spl237:
Bind to Super_L:

`    <keybind key="Super_L">
      <action name="Execute">
        <command>wfpanelctl smenu menu</command>
      </action>
    </keybind>`

johanmalm:
> Bind to Super_L:
> 
> ` <keybind key="Super_L"> <action name="Execute"> <command>wfpanelctl smenu menu</command> </action> </keybind>`
> 
> I think W is the key when used as a modifier, and Super_L is the same key as a plain keystroke.

Genius. Will have a play :smile:

src/input/keyboard.c
```diff
@@ -407,7 +409,18 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 		keyinfo.is_modifier);
 
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-		return handle_key_release(server, event->keycode);
+		if (cur_keybind && cur_keybind->mod_only) {
+			if (seat->active_client_while_inhibited
+					|| seat->server->session_lock) {
+				cur_keybind = NULL;
+				return false;
```
johanmalm, line 416: I think should be `true` to that the event is not passed onto the client.

src/input/keyboard.c
```diff
@@ -407,7 +409,18 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 		keyinfo.is_modifier);
 
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
-		return handle_key_release(server, event->keycode);
+		if (cur_keybind && cur_keybind->mod_only) {
+			if (seat->active_client_while_inhibited
+					|| seat->server->session_lock) {
+				cur_keybind = NULL;
```
johanmalm, line 415: We store key as bound (https://github.com/labwc/labwc/blob/1cd9940307c4b54503cefcd643e92314ade88c45/src/input/keyboard.c#L454) so need to call `key_state_bound_key_remove()` here.

johanmalm:
I've tested and it works well.

I'm pretty close to merging, but feel we ought to consider a few points.

1. `s/TRUE/true/` and `s/FALSE/false/`. I think the capitalized versions are some glib thing from the pre `stdbool.h` era.

2. I suggest changing `mod_only` to `on_release`. I think it's easier for the reader to understand what's going on.

3. Is is better to be explicit about when we handle the `release` event. For example: `<keybind key="Super_L" event="onRelease">` instead of calling `keyboard_is_modifier_key()`. I think it's pretty unlikely that anyone will want to bind to the press-event of a modifier, but you can't beat spelling it out. I can have a go at doing the parsing if you want, but let's have the discussion first.

4. Consider if we should allow multiple modifiers. It looks like it currently does (although not tested). I think the main use-case is `Super_L` so we could just set a rule to only allow on-release binds with `keysyms_len==1`. It's likely to make the key-state logic simpler too.

5. I've followed the key-state logic through and made a couple of suggestions. When merge-ready, I can follow it through with `LABWC_DEBUG_KEY_STATE=1` to double-check. It's an area of the code where it's particularly easy to make thought-mistakes.

spl237:
OK, thanks for that - will have a look at those changes tomorrow.

src/input/keyboard.c
```diff
@@ -38,6 +38,8 @@ struct keyinfo {
 
 static bool should_cancel_cycling_on_next_key_release;
 
+struct keybind *cur_keybind;
```
Consolatis, line 41: This should be static.

I am also not sure about `Reconfigure` happening between press and release, this seems to miss some reconfigure hook to reset `cur_keybind` in that case.

src/input/keyboard.c
spl237: Done

src/input/keyboard.c
spl237: Done

src/input/keyboard.c
spl237: Done

spl237:
> I've tested and it works well.
> 
> I'm pretty close to merging, but feel we ought to consider a few points.
> 
> 1. `s/TRUE/true/` and `s/FALSE/false/`. I think the capitalized versions are some glib thing from the pre `stdbool.h` era.

Fixed.

> 2. I suggest changing `mod_only` to `on_release`. I think it's easier for the reader to understand what's going on.

Fixed.

> 3. Is is better to be explicit about when we handle the `release` event. For example: `<keybind key="Super_L" event="onRelease">` instead of calling `keyboard_is_modifier_key()`. I think it's pretty unlikely that anyone will want to bind to the press-event of a modifier, but you can't beat spelling it out. I can have a go at doing the parsing if you want, but let's have the discussion first.

That sounds sensible to me.

> 4. Consider if we should allow multiple modifiers. It looks like it currently does (although not tested). I think the main use-case is `Super_L` so we could just set a rule to only allow on-release binds with `keysyms_len==1`. It's likely to make the key-state logic simpler too.

I can't think of any situations where you would want to bind to multiple modifiers for a single action, but I've got nothing against the idea if others think it would be useful.

> 5. I've followed the key-state logic through and made a couple of suggestions. When merge-ready, I can follow it through with `LABWC_DEBUG_KEY_STATE=1` to double-check. It's an area of the code where it's particularly easy to make thought-mistakes.

I think those should be fixed, but do please check.

Many thanks!

Consolatis:
Merged via
- #1888

PR #1762, opened 2024-04-26
ssd_titlebar.c: don't pass negative value to wlr_scene_rect_set_size()

tokyo4j:
I noticed following error was printed when a view is resized to minimum width.
I re-installed pixman Arch Linux package with symbols to debug it and found this is due to a negative value passed to `wlr_scene_rect_set_size()`. 

```
*** BUG ***
In pixman_region32_union_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug
```

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -3,14 +3,15 @@
 #define LABWC_VIEW_H
 
 #include "config.h"
+#include "ssd.h"
 #include <stdbool.h>
 #include <stdint.h>
 #include <wayland-util.h>
 #include <wlr/util/box.h>
 #include <xkbcommon/xkbcommon.h>
 
-#define LAB_MIN_VIEW_WIDTH  100
-#define LAB_MIN_VIEW_HEIGHT  60
+#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT)
+#define LAB_MIN_VIEW_HEIGHT 60
 
 /*
  * In labwc, a view is a container for surfaces which can be moved around by
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -229,6 +229,16 @@ ssd_update_geometry(struct ssd *ssd)
 	int eff_width = current.width;
 	int eff_height = view_effective_height(ssd->view, /* use_pending */ false);
 
+	if (eff_width > 0 && eff_width < LAB_MIN_VIEW_WIDTH) {
+		/*
+		 * Prevent negative values in calculations like
+		 * `width - SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT`
+		 */
+		wlr_log(WLR_ERROR,
+			"view width is smaller than its minimal value");
+		return;
+	}
+
 	if (eff_width == cached.width && eff_height == cached.height) {
 		if (current.x != cached.x || current.y != cached.y) {
 			/* Dynamically resize extents based on position and usable_area */


src/ssd/ssd_titlebar.c
```diff
@@ -204,9 +204,10 @@ ssd_titlebar_update(struct ssd *ssd)
 		wl_list_for_each(part, &subtree->parts, link) {
 			switch (part->type) {
 			case LAB_SSD_PART_TITLEBAR:
+				/* LAB_MIN_VIEW_WIDTH < SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT */
```
Consolatis, line 207: Hm.. I wonder how that can happen.

tokyo4j: They're defined like this:
```c
#define LAB_MIN_VIEW_WIDTH  100
#define SSD_BUTTON_COUNT 4
#define SSD_BUTTON_WIDTH 26
```
Feel free to close this PR if another solution like changing these values is better.
I'm not familiar with `ssd*.c` and just blindly fixed the error from pixman.

Consolatis: Whoops, that seems pretty wrong then.

Personal view: what about us doing `#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT)` but also detect the ssd_titlebar case and log an error before fixing it?

I am not quite sure currently if a view really can never be smaller than `LAB_MIN_VIEW_HEIGHT`.

tokyo4j: I changed it to `#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT)`.
Also, I updated `ssd_update_geometry()` to log error and early-return if the view width is smaller than that, since the same pixman error also happens in the following part in `ssd_border_update()` if the view width is very small.
```c
			case LAB_SSD_PART_TOP:
				if (ssd->titlebar.height > 0) {
					wlr_scene_rect_set_size(rect,
						width - 2 * SSD_BUTTON_WIDTH,
						theme->border_width);
``` 
> I am not quite sure currently if a view really can never be smaller than LAB_MIN_VIEW_WIDTH.

Yeah, I confirmed clients can ignore the requested geometry and `view->current.width` can be smaller than `LAB_MIN_VIEW_WIDTH` ([source](https://github.com/tokyo4j/wayland-demo/tree/deco-width-10)).
20240427_16h37m33s_grim

Consolatis: > I changed it to `#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT)`.
> Also, I updated `ssd_update_geometry()` to log error and early-return if the view width is smaller than that, since the same pixman error also happens in the following part in `ssd_border_update()` if the view width is very small.

Thanks, I'll look into the diff in a few hours.

> So I suggest enforcing the minimal geometry in `view_impl_apply_geometry()`.

There must be some discussion around this somewhere, I thought we actually did that already but apparently we did not. It should also only be applied when there is actually a SSD with titlebar shown, otherwise there is not really a point in preventing smaller windows. This is something that should likely wait until after the release `0.7.2` though.

Consolatis:
Tested and works fine when the client cooperates.

When using your test client the SSD is kind of broken due to the early return in `ssd_update()` (e.g. menu button on the right and there is no way to resize a view from that state other than using `A-RMB` or some custom `Resize` keybind) but I am fine with this for 0.7.2. This should likely be revisited though.

Thanks!

PR #1763, opened 2024-04-29
view: update top layer visibility on map

### LINKED ISSUE #1752: Steam running in big picture mode under xwayland makes waybar disappear
When I open steam in big picture mode, it runs under xwayland, and if I open an app like MPV in fullscreen, when I exit out of fullscreen the waybar has gone, even though its still running, when I close steam then the panel reappears again, I've made a short video of the issue below, and if I don't run steam in big picture mode then I have even bigger issues, this also happens on other panels too.

https://www.youtube.com/watch?v=WirdzEA4eZY

tokyo4j:
Fixes #1752.

Steam's Big Picture mode runs in fullscreen mode, but requests it before being mapped.
When a client requests fullscreen mode, `desktop_update_top_layer_visiblity()` checks if there's a fullscreen window in the output to determine the top layer should be hidden or not, but unmapped views are always ignored.
So `desktop_update_top_layer_visiblity()` should be called also on map. 

Not sure this is following the spec.

### DIFF
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -46,6 +46,12 @@ view_impl_map(struct view *view)
 		}
 	}
 
+	/*
+	 * Some clients (e.g. Steam's Big Picture Mode window) request
+	 * fullscreen before mapping.
+	 */
+	desktop_update_top_layer_visiblity(view->server);
+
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks for investigating and fixing this :)

Issue #1767, opened 2024-04-30
"Desktop" context for mouse bindings doesn't work


spl237:
When creating a mouse binding, the "Desktop" context appears to be the same as the "Root" context - in context_from_str in src/config/mousebind.c, both options are translated to LAB_SSD_ROOT.

If I have no application running to draw a desktop, then this works ok - the Root context is then valid. But if I am running an application to draw a desktop (pcmanfm in our case) then the Root context is no longer in effect, so binding to the Desktop context means the binding is no longer in effect. The Frame context also doesn't work for the desktop background.

Is there some valid context for the desktop background?

(Or even a context for "anywhere at all"? I am trying to define a global mouse binding, and I had hoped that setting it for Frame and Desktop would work.)

Consolatis:
> Is there some valid context for the desktop background?

No, there isn't. I don't think we even check any mousebinds for interactions with a layershell surface.
You could theoretically make your desktop implementation use an empty wayland input region, that should then let clicks "go through" the desktop and thus trigger the `Desktop` / `Root` context. That is what for example `swaybg` is doing.

> I am trying to define a global mouse binding, and I had hoped that setting it for Frame and Desktop would work.

Hm.. Maybe we could add some `Layershell` context (or individual ones like `LayerBackground` / `LayerBottom` / `LayerTop`) or something along that line. Thoughts @johanmalm ?

spl237:
> You could theoretically make your desktop implementation use an empty wayland input region, that should then let clicks "go through" the desktop and thus trigger the `Desktop` / `Root` context. 

I don't think that will work for us - the user needs to interact with icons on the desktop, so presumably making it an empty input region will prevent that?

Consolatis:
> I don't think that will work for us - the user needs to interact with icons on the desktop, so presumably making it an empty input region will prevent that?

Yes, that would prevent any mouse interaction with the surface that owns the input region.

droc12345:
Your desktop app is simply another application, not what's referred to as "the desktop" ie, root.
That's why all the settings have the word "context" in front of the name.

When you completely cover the root window, with your desktop app, 
then indeed you quit interacting with the root window. Not unlike having a fullscreen window and not being able to interact with the root window.

AFAIK, there is no such thing as a global binding. 

What are you trying to accomplish with a global binding?

spl237:
> Your desktop app is simply another application, not what's referred to as "the desktop" ie, root. That's why all the settings have the word "context" in front of the name.
> 
[...]

PR #1768, opened 2024-05-01
Add All context for mouse events

spl237:
Fairly trivial change to add an "All" context for mouse bindings which need to be handled irrespective of the mouse pointer location, as discussed at https://github.com/labwc/labwc/issues/1767

Tested locally - seems to work fine for me!


tokyo4j:
Documentation is missing. I suggest something like:
```diff
diff --git a/docs/labwc-config.5.scd b/docs/labwc-config.5.scd
index acc59ada..2c55763a 100644
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -559,6 +559,7 @@ extending outward from the snapped edge.
 	  through to the application.
 	- Desktop: The desktop background, where no windows are present.
 	- Root: A synonym for Desktop (for compatibility).
+	- All: Any of the contexts described above.
 
 	Supported mouse *buttons* are:
 	- Left
```

spl237:
> Documentation is missing. I suggest something like:
> 
> ```diff
> diff --git a/docs/labwc-config.5.scd b/docs/labwc-config.5.scd
> index acc59ada..2c55763a 100644
> --- a/docs/labwc-config.5.scd
> +++ b/docs/labwc-config.5.scd
> @@ -559,6 +559,7 @@ extending outward from the snapped edge.
>  	  through to the application.
>  	- Desktop: The desktop background, where no windows are present.
>  	- Root: A synonym for Desktop (for compatibility).
> +	- All: Any of the contexts described above.
>  
>  	Supported mouse *buttons* are:
>  	- Left
> ```
> 
> Feel free to enhance my words. I'm poor at writing English.

Good idea - thanks for the suggestion!

It's not quite accurate to say that 'All' is any of the contexts, because it also includes things like the desktop background which isn't covered by any of the existing contexts, so I have described it as "Anywhere on the screen".

droc12345:
All: Supersedes of all of the other contexts, plus places where context might not apply.

Consolatis:
> On testing, scrolling over a window without the modifier key held causes the window to scroll; scrolling with the modifier held only triggers the desired mouse binding - so this appears to do what would be expected.

This is likely only because of the `Scroll` bindings, if you try the same with a `Press` binding I assume the click will go through to the application. I think this needs to update `consumed_by_frame_context`.

spl237:
> > On testing, scrolling over a window without the modifier key held causes the window to scroll; scrolling with the modifier held only triggers the desired mouse binding - so this appears to do what would be expected.
> 
> This is likely only because of the `Scroll` bindings, if you try the same with a `Press` binding I assume the click will go through to the application. I think this needs to update `consumed_by_frame_context`.
> 
Does that change do the right thing?

Consolatis:
> > > On testing, scrolling over a window without the modifier key held causes the window to scroll; scrolling with the modifier held only triggers the desired mouse binding - so this appears to do what would be expected.
> > 
> > 
> > This is likely only because of the `Scroll` bindings, if you try the same with a `Press` binding I assume the click will go through to the application. I think this needs to update `consumed_by_frame_context`.
> 
> Does that change do the right thing?

Theoretically yes. But it looks like labwc doesn't really do the right thing in regards to eating the `Release` event for `Press` mousebinds in the `Frame` (and now `All` context). `Click` on the other side works (but only executes the first matching binding found). That is somewhat outside the scope of this PR though and needs some work.

droc12345:
Why would you ever not want the mouse press, scroll, modifier, etc in mousebind to be consumed, ie never pass to whatever is underneath?

johanmalm:
LGTM. Have tested with config below (after rebasing onto `master`). Works well.

```
  <mouse>
    <context name="All">
      <mousebind direction="W-Up" action="Scroll">
        <action name="ZoomIn"/>
      </mousebind>
      <mousebind direction="W-Down" action="Scroll">
        <action name="ZoomOut"/>
      </mousebind>
    </context>
  </mouse>
```

spl237:
Excellent - many thanks!

PR #1771, opened 2024-05-03
Magnifier

spl237:
This PR integrates the screen magnifier written by @Consolatis (https://github.com/Consolatis/labwc/tree/feature/magnifier) with the user interface and the configuration system.

It requires two currently-pending PRs for correct functionality - https://github.com/labwc/labwc/pull/1761 to prevent modifier-only keybindings being triggered until release, and https://github.com/labwc/labwc/pull/1768 to add an "All" context for mouse bindings.

There are two bindings added to control the magnifier. Ctrl-Alt-m toggles the magnifier window on and off, and Win+scroll adjusts the magnification. (The second of these is the only part which requires the two PRs mentioned above.)

I have added a magnifier section to rc.xml, which contains parameters for width and height of the magnifier window, initial magnification scale, border width and colour, and whether or not to bilinear filter the output. (When I use a magnifier, I tend to prefer it unfiltered, as I am usually using it to inspect details of pixels in an icon or similar, but I suspect most people would prefer it filtered). All options are documented in rc.xml.all.

The only change I have made to the original magnifier code from @Consolatis is to make the size of the magnifier window fixed rather than dependent on the magnification, but this could also be a config parameter if that is felt to be useful.

Many thanks to @Consolatis for the original magnification code - I'd never have worked that out myself! But hopefully this is felt to be a reasonable integration of the functionality from a user's perspective.


spl237:
Pulled in some unintended commits while merging - will try to re-create!

Consolatis:
Thanks for the PR.
Suggest using `git rebase -i` rather than `git merge`, could likely also be used to clean up the branch.

spl237:
I'm trying to create a branch which doesn't have any of the commits from the two pending PRs on it - is that the right thing to do? Or would it be better to include them?

Consolatis:
Both would be fine, AFAIR there are no direct code dependencies between the two, only logical ones.

ahesford:
Nothing about this pull request should depend on either of your other two requests. Also, I don't think this should have a default keybind assigned. labwc already has more default keybinds than it probably should, and we shouldn't clutter up the space for users who have configured others. Let people opt into the magnifier.

spl237:
> Nothing about this pull request should depend on either of your other two requests. Also, I don't think this should have a default keybind assigned. labwc already has more default keybinds than it probably should, and we shouldn't clutter up the space for users who have configured others. Let people opt into the magnifier.

OK, that's sensible - it is only the keybinds which depend on the other PRs. I'll remove them for now.

PR #1772, opened 2024-05-03
[RFC] config: use automatic placement by default

ahesford:
Automatic placement has been in place for a couple of releases now, and I think it makes the compositor much more usable to have windows try to avoid collisions rather than centering everything. Maybe it's time to switch the default policy.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -196,12 +196,12 @@ this is for compatibility with Openbox.
 
 ## PLACEMENT
 
-*<placement><policy>* [center|automatic|cursor]
-	Specify a placement policy for new windows. The "center" policy will
-	always place windows at the center of the active output. The "automatic"
-	policy will try to place new windows in such a way that they will
-	have minimal overlap with existing windows. The "cursor" policy will
-	center new windows under the cursor. Default is "center".
+*<placement><policy>* [automatic|center|cursor]
+	Specify a placement policy for new windows. The "automatic" policy will
+	try to place new windows in such a way that they will have minimal
+	overlap with existing windows. The "center" policy will always place
+	windows at the center of the active output. The "cursor" policy will
+	center new windows under the cursor. Default is "automatic".
 
 ## WINDOW SWITCHER
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -16,7 +16,7 @@
   </core>
 
   <placement>
-    <policy>center</policy>
+    <policy>automatic</policy>
   </placement>
 
   <!-- <font><theme> can be defined without an attribute to set all places -->
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -871,12 +871,12 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "reuseOutputMode.core")) {
 		set_bool(content, &rc.reuse_output_mode);
 	} else if (!strcmp(nodename, "policy.placement")) {
-		if (!strcmp(content, "automatic")) {
-			rc.placement_policy = LAB_PLACE_AUTOMATIC;
+		if (!strcmp(content, "center")) {
+			rc.placement_policy = LAB_PLACE_CENTER;
 		} else if (!strcmp(content, "cursor")) {
 			rc.placement_policy = LAB_PLACE_CURSOR;
 		} else {
-			rc.placement_policy = LAB_PLACE_CENTER;
+			rc.placement_policy = LAB_PLACE_AUTOMATIC;
 		}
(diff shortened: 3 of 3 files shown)


Consolatis:
I don't know about this. For me personally as a user it would actually be a surprise to have that as a default setting.
What does Openbox do?

ahesford:
Openbox [enables "smart" placement by default](https://github.com/Mikachu/openbox/blob/53737a844d2bba4bb5892737b05a01d6d06fe84b/openbox/config.c#L1100), but separates the concept of centering from the actual policy (which is only "smart" or "under mouse") and will center each new window in its optimal area.

It would not be difficult to adapt an option centering for smart placement, but would require a change in configuration that makes things a bit more complicated.

Automatic placement was born because I found the labwc default surprising. Most other compositors and X11 window managers I've used make some attempt to vary new window positions as clients are launched.

Consolatis:
Other thoughts @johanmalm @tokyo4j @jlindgren90 ?

tokyo4j:
I'm fine with enabling smart placement by default, but I think it should be after `<placement><center>` is implemented.
All the stacking WMs I've used enable smart placement (with different algorithm) by default, but all of them open the first window at center of the screen.

I personally like the placement policy of Windows or KWin which place the first window at center and the second window at center too but shifted to bottom-right a bit.

jlindgren90:
I haven't tested the smart placement yet myself, but I like the idea of enabling it by default if it feels intuitive enough. Maybe after 0.7.2?

domo141:
I have smart placement, and sometimes i find it hard to find these small windows at some small free space in a desktop.
also some startup windows that used, on X11, open in center, now opens laid out next to some other window (near an edge/corner of an output, usually...)

As in wayland one cannot position these (special-case) windows it is somehat weird to have these open at random places ;/

OTOH, opening every window in the center is IMO more disturbing than some windows that would be nice to get center are opened at random places.

Consolatis:
Just speaking as a user here, what I like about default open-at-center is that I can predict what happens. I open a window and already know where it will be. With smart placement that is not the case and I have to go on a hunt for the window. The always-center default makes it much easier to understand the rules of the compositor in my opinion.

jlindgren90:
> also some startup windows that used, on X11, open in center, now opens laid out next to some other window

Startup windows a.k.a. "splash screens"? I would think these ought to be excluded from the smart placement.

jlindgren90:
> With smart placement that is not the case and I have to go on a hunt for the window.

Are there any adjustments that could make the smart placement more predictable? Such as, only use smart placement if there is an obvious/large blank area to place the window in?

I recall Metacity had a design goal of having just one placement scheme that was "good enough" for most use cases -- it would be nice if we could achieve that.

johanmalm:
I really like automatic placement, but don't think we should change the default behaviour.

stefonarch:
> Just speaking as a user here, what I like about default open-at-center is that I can predict what happens. I open a window and already know where it will be. With smart placement that is not the case and I have to go on a hunt for the window. The always-center default makes it much easier to understand the rules of the compositor in my opinion.

That's  exactly the reason why I use the  "cursor" placement,otherwise I had often to move it all over the monitor to get it to the opened window.
 Maybe a mix between "automatic" and "center" would be the best, opening the windows more or less at center, especially small windows.

tokyo4j:
I just implemented [KWin's algorithm](https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/placement.cpp#L589) which tries to shift a window not to cover existing ones when `center` policy is used.

https://github.com/tokyo4j/labwc/tree/kwin-placement

I think something like this, or a mix of `automatic` and `center` as @stefonarch suggests should be the default.

stefonarch:
> I just implemented [KWin's algorithm](https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/placement.cpp#L589) which tries to shift a window not to cover existing ones when `center` policy is used.

Works nicely:

immagine

johanmalm:
@ahesford Are you okay if we close this one leaving 'center' as default?

ahesford:
Sure thing.

tokyo4j:
I said:
> All the stacking WMs I've used enable smart placement (with different algorithm) by default, but all of them open the first window at center of the screen.

But this was wrong. Xfce4 enables smart placement and opens the first window at top-left corner by default, but it has configurable "minimal window size to trigger smart placement" so tiny windows like dialogs are opened at center.

As a former xfce4 user, now I'm not strongly against enabling the current `smart` placement by default, but I understand why other devs are against it and I still prefer `center` placement.

johanmalm:
I’ve added placement-policy to labwc-tweaks as a convenient way of changing it.

Issue #1619, opened 2024-03-14
Try to re-set cursor image on reconfigure if either cursor theme or cursor size changed


Consolatis:
This gives instant feedback when changing cursor theme or size. It only works for server side cursors or clients using the cursor-shape protocol.

---

- #1612

PR #1773, opened 2024-05-03
cursor: reload cursor on Reconfigure

Consolatis:
This gives instant feedback when changing cursor theme or size. It only works for server side cursors or clients using the cursor-shape protocol.

Fixes:
- #1619


johanmalm:
LGTM. Have tested - works well.

PR #1774, opened 2024-05-03
Add screen magnifier

spl237:
This PR integrates the screen magnifier written by @Consolatis with the user interface and the configuration system.

I have added a magnifier section to rc.xml, which contains parameters for width and height of the magnifier window, initial magnification scale, border width and colour, and whether or not to bilinear filter the output. (When I use a magnifier, I tend to prefer it unfiltered, as I am usually using it to inspect details of pixels in an icon or similar, but I suspect most people would prefer it filtered). All options are documented in rc.xml.all and labwc-config.5.scd.

I've added two functions which can be used in mouse and keyboard bindings to control the magnifier - 'ToggleMagnify' to toggle the magnifier window on and off, and 'SetMagnification scale up/down' to change the magnification factor.

The only change I have made to the original magnifier code from @Consolatis is to make the size of the magnifier window fixed rather than dependent on the magnification, but this could also be a config parameter if that is felt to be useful.

Many thanks to @Consolatis for the original magnification code - I'd never have worked that out myself! But hopefully this is felt to be a reasonable integration of the functionality from a user's perspective.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -264,6 +264,17 @@ Actions are used in menus and keyboard/mouse bindings.
 	decorations (including those for which the server-side titlebar has been
 	hidden) are not eligible for shading.
 
+*<action name="ToggleMagnify">*
+	Toggle the screen magnifier on or off at the last magnification level
+	used.
+
+*<action name="ZoomIn">*++
+*<action name="ZoomOut">*
+	Increase or decrease the magnification level for the screen magnifier.
+	If the magnifier is currently off, ZoomIn will enable it at the lowest
+	magnification (x2). If the magnifier is on and at a magnification of
+	x2, ZoomOut will turn it off.
+
 *<action name="None" />*
 	If used as the only action for a binding: clear an earlier defined
 	binding.
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -272,8 +272,9 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="ZoomOut">*
 	Increase or decrease the magnification level for the screen magnifier.
 	If the magnifier is currently off, ZoomIn will enable it at the lowest
-	magnification (x2). If the magnifier is on and at a magnification of
-	x2, ZoomOut will turn it off.
+	magnification, equal to (1 + the magnifier increment set in the theme).
+	If the magnifier is on and at the lowest magnification, ZoomOut will
+	turn it off.
 
 *<action name="None" />*
 	If used as the only action for a binding: clear an earlier defined
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -277,6 +277,27 @@ elements are not listed here, but are supported.
 	*window.inactive.border.color*. This is obsolete, but supported for
 	backward compatibility as some themes still contain it.
 
+*magnifier.width*
+	Width of magnifier window in pixels.
+
+*magnifier.height*
+	Height of magnifier window in pixels.
+
+*magnifier.border.width*
+	Width of magnifier window border in pixels.
+
+*magnifier.border.color*
+	Color of the magnfier window border.
+
+*magnifier.init-scale*
+	Initial number of times by which magnified image is scaled. Value
+	is the default at boot; can be modified at run-time	in a keyboard
+	or mouse binding by calling 'ZoomIn' or 'ZoomOut'.
+
+*magnifier.filter*
+	Set to 1 to apply a bilinear filter to the magnified image.
+	Set to 0 to use nearest-neighbour.
+
 # BUTTONS
 
 The images used for the titlebar icons are referred to as buttons.
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -296,6 +296,10 @@ elements are not listed here, but are supported.
 	is the default at boot; can be modified at run-time	in a keyboard
 	or mouse binding by calling 'ZoomIn' or 'ZoomOut'. Default is x2.
 
+*magnifier.increment*
+	Step by which magnification changes on each call to 'ZoomIn' or
+	'ZoomOut'. Default is 0.2.
+
 *magnifier.filter*
 	Set to 1 to apply a bilinear filter to the magnified image.
 	Set to 0 to use nearest-neighbour. Default is 1 - bilinear filtered.
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -278,25 +278,25 @@ elements are not listed here, but are supported.
 	backward compatibility as some themes still contain it.
 
 *magnifier.width*
-	Width of magnifier window in pixels.
+	Width of magnifier window in pixels. Default is 400.
 
 *magnifier.height*
-	Height of magnifier window in pixels.
+	Height of magnifier window in pixels. Default is 400.
 
 *magnifier.border.width*
-	Width of magnifier window border in pixels.
+	Width of magnifier window border in pixels. Default is 1.
 
 *magnifier.border.color*
-	Color of the magnfier window border.
+	Color of the magnfier window border. Default is #ff0000 (red).
 
 *magnifier.init-scale*
 	Initial number of times by which magnified image is scaled. Value
 	is the default at boot; can be modified at run-time	in a keyboard
-	or mouse binding by calling 'ZoomIn' or 'ZoomOut'.
+	or mouse binding by calling 'ZoomIn' or 'ZoomOut'. Default is x2.
 
 *magnifier.filter*
 	Set to 1 to apply a bilinear filter to the magnified image.
-	Set to 0 to use nearest-neighbour.
+	Set to 0 to use nearest-neighbour. Default is 1 - bilinear filtered.
 
 # BUTTONS
 
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -279,9 +279,11 @@ elements are not listed here, but are supported.
 
 *magnifier.width*
 	Width of magnifier window in pixels. Default is 400.
+	Set to -1 to use fullscreen magnifier.
 
 *magnifier.height*
 	Height of magnifier window in pixels. Default is 400.
+	Set to -1 to use fullscreen magnifier.
 
 *magnifier.border.width*
 	Width of magnifier window border in pixels. Default is 1.
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -277,32 +277,11 @@ elements are not listed here, but are supported.
 	*window.inactive.border.color*. This is obsolete, but supported for
 	backward compatibility as some themes still contain it.
 
-*magnifier.width*
-	Width of magnifier window in pixels. Default is 400.
-	Set to -1 to use fullscreen magnifier.
-
-*magnifier.height*
-	Height of magnifier window in pixels. Default is 400.
-	Set to -1 to use fullscreen magnifier.
-
 *magnifier.border.width*
 	Width of magnifier window border in pixels. Default is 1.
 
 *magnifier.border.color*
 	Color of the magnfier window border. Default is #ff0000 (red).
-
-*magnifier.init-scale*
-	Initial number of times by which magnified image is scaled. Value
-	is the default at boot; can be modified at run-time	in a keyboard
-	or mouse binding by calling 'ZoomIn' or 'ZoomOut'. Default is x2.
-
-*magnifier.increment*
-	Step by which magnification changes on each call to 'ZoomIn' or
-	'ZoomOut'. Default is 0.2.
-
-*magnifier.filter*
-	Set to 1 to apply a bilinear filter to the magnified image.
-	Set to 0 to use nearest-neighbour. Default is 1 - bilinear filtered.
 
 # BUTTONS
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -603,7 +603,7 @@
   <magnifier>
     <width>400</width>
     <height>400</height>
-    <initScale>2</initScale>
+    <initScale>2.0</initScale>
     <increment>0.2</increment>
     <useFilter>true</useFilter>
   </magnifier>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -593,12 +593,18 @@
   <!--
     Magnifier settings
 
-    'scale' sets the initial magnification factor when first booted.
     'size' sets the width and height in pixels of the magnifier window.
+    'initScale' sets the initial magnification factor when first booted.
+    'borderColour' is an XML RGB string with a leading # which sets the
+      colour of the border for the magnifier window.
+    'borderWidth' sets the width in pixels of the border for the
+      magnifier window.
   -->
     <magnifier>
-        <scale>2</scale>
         <size>400</size>
+        <initScale>2</initScale>
+        <borderColour>#ff0000</borderColour>
+        <borderWidth>1</borderWidth>
     </magnifier>
 
 </labwc_config>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -589,4 +589,23 @@
   <menu>
     <ignoreButtonReleasePeriod>250</ignoreButtonReleasePeriod>
   </menu>
+
+  <!--
+    Magnifier settings
+    'width' sets the width in pixels of the magnifier window.
+    'height' sets the height in pixels of the magnifier window.
+    'initScale' sets the initial magnification factor at boot.
+    'increment' sets the amount by which the magnification factor
+      changes when 'ZoomIn' or 'ZoomOut' are called.
+    'useFilter' sets whether to use a bilinear filter on the magnified
+      output or simply to take nearest pixel.
+  -->
+  <magnifier>
+    <width>400</width>
+    <height>400</height>
+    <initScale>2</initScale>
+    <increment>0.2</increment>
+    <useFilter>true</useFilter>
+  </magnifier>
+
 </labwc_config>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -589,5 +589,4 @@
   <menu>
     <ignoreButtonReleasePeriod>250</ignoreButtonReleasePeriod>
   </menu>
-
 </labwc_config>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -599,12 +599,15 @@
       colour of the border for the magnifier window.
     'borderWidth' sets the width in pixels of the border for the
       magnifier window.
+    'useFilter' sets whether to use a bilinear filter on the magnified
+      output or simply to take nearest pixel.
   -->
     <magnifier>
         <size>400</size>
         <initScale>2</initScale>
         <borderColour>#ff0000</borderColour>
         <borderWidth>1</borderWidth>
+        <useFilter>true</useFilter>
     </magnifier>
 
 </labwc_config>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -590,26 +590,4 @@
     <ignoreButtonReleasePeriod>250</ignoreButtonReleasePeriod>
   </menu>
 
-  <!--
-    Magnifier settings
-
-    'width' sets the width in pixels of the magnifier window.
-    'height' sets the height in pixels of the magnifier window.
-    'initScale' sets the initial magnification factor at boot.
-    'borderColour' is an XML RGB string with a leading # which sets the
-      colour of the border of the magnifier window.
-    'borderWidth' sets the width in pixels of the border of the
-      magnifier window.
-    'useFilter' sets whether to use a bilinear filter on the magnified
-      output or simply to take nearest pixel.
-  -->
-    <magnifier>
-        <width>400</width>
-        <height>400</height>
-        <initScale>2</initScale>
-        <borderColour>#ff0000</borderColour>
-        <borderWidth>1</borderWidth>
-        <useFilter>true</useFilter>
-    </magnifier>
-
 </labwc_config>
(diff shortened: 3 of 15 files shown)


src/action.c
```diff
@@ -1046,6 +1057,19 @@ actions_run(struct view *activator, struct server *server,
 				view_set_shade(view, false);
 			}
 			break;
+		case ACTION_TYPE_TOGGLE_MAGNIFY:
+			magnify_toggle();
+			break;
+		case ACTION_TYPE_SET_MAGNIFICATION:
+			{
+				const char *dir = action_get_str(action, "scale", NULL);
+				if (!strcmp(dir, "up")) {
```
Consolatis, line 1066: This will break when the argument is not set, please add a branch in `action_is_valid()`

Another option could be to use two actions: `ZoomIn` and `ZoomOut`.

docs/rc.xml.all
```diff
@@ -589,4 +589,27 @@
...
+      magnifier window.
+    'useFilter' sets whether to use a bilinear filter on the magnified
+      output or simply to take nearest pixel.
+  -->
+    <magnifier>
+        <width>400</width>
+        <height>400</height>
+        <initScale>2</initScale>
+        <borderColour>#ff0000</borderColour>
+        <borderWidth>1</borderWidth>
```
Consolatis, line 611: I think these two belong to the theme, then we also don't need to add the rgb parser in `rcxml.c`.

src/action.c
spl237: > Another option could be to use two actions: `ZoomIn` and `ZoomOut`.

Much better idea! I'll do that.

Consolatis:
Just a quick review, will do another one after the release so might take a week or two.
This is also missing documention for the new actions.

spl237:
> Just a quick review, will do another one after the release so might take a week or two. This is also missing documention for the new actions.

Now added.

docs/rc.xml.all
Consolatis: Also, please use the american spelling of `Color` for the setting, it matches the rest of the codebase.

spl237: Will do - didn't know what standard was being used here...

spl237: Should "colour" be used in documentation as well, or just in code?

Consolatis:
Did some quick test nested by just adding the `ZoomIn` and `ZoomOut` actions to the `Root` context with `W-Up/Down` and it feels very intuitive. Obviously works even better with support for`All` / `Global` context. Functionality wise and in regards of how it can be controlled by the user I think this is great!

spl237:
> Did some quick test nested by just adding the `ZoomIn` and `ZoomOut` actions to the `Root` context with `W-Up/Down` and it feels very intuitive. Obviously works even better with support for`All` / `Global` context. Functionality wise and in regards of how it can be controlled by the user I think this is great!

Thanks very much! 

I find it works really well with the scroll wheel - it's almost not necessary to have the global toggle as well, but there may be people who find using the scroll awkward, so I think it is worth having the option.

Would be great if this can indeed get into the 0.7.3 milestone as you have suggested, as I think this is the last remaining missing feature compared to our current compositor, so we could then start to think about switching over to labwc permanently.

docs/labwc-theme.5.scd
```diff
@@ -277,6 +277,27 @@ elements are not listed here, but are supported.
...
+	Color of the magnfier window border.
+
+*magnifier.init-scale*
+	Initial number of times by which magnified image is scaled. Value
+	is the default at boot; can be modified at run-time	in a keyboard
+	or mouse binding by calling 'ZoomIn' or 'ZoomOut'.
+
+*magnifier.filter*
+	Set to 1 to apply a bilinear filter to the magnified image.
+	Set to 0 to use nearest-neighbour.
```
tokyo4j, line 299: Default values should be documented.

spl237: No problem - will do that!

spl237:
I've been doing some more testing on this, and there is a problem with the code at present if you choose a very large magnifier window; I was trying to get a window which filled the screen, and while that worked fine on a small-ish monitor, on a 4K monitor, it produced display corruption, I think due to the temporary render buffer not being large enough.

I'm currently working on a fix - I have large windows working fine with a different set of calculations for x, y, w and h - just need to try to find a way which works for small windows as well!

spl237:
OK, I've added an explicit "fullscreen" mode for now - this emulates the behaviour of the zoom plugin in Wayfire; that of magnifying the entire screen rather than a smaller window over the top. I'm not 100% happy with this yet; I've had to do the display area calculations in a completely different way from those used for the windowed version, and I don't think this ought to be necessary, so I am going to play with this a bit further and try to align them a bit better. But I wanted to get something at least checked in so I didn't lose it...

spl237:
I think fullscreen is now good to go - to enable it, set the width or the height in the theme to -1.

tsujan:
Just a user's comment: This is very handy. Thanks a lot!

Consolatis:
> On playing with this a bit, particularly when using the scroll wheel to adjust magnification, it feels better to me to have magnification change in smaller steps than 1 each time; 0.2 feels nicer to me - open to opinions from anyone else who tries it!
> 
> It can be made a config parameter if necessary - it is on the wayfire zoom plugin - but that seems to me to be something where it is better to make a sensible choice upfront than to allow the user to fiddle with it.

Good points. I think having something like 0.2 as a default but still provide a config option is fine for this. As an alternative, we could add an optional argument to the `ZoomIn` and `ZoomOut` actions (which then defaults to 0.2) instead of having a global setting.

spl237:
> > On playing with this a bit, particularly when using the scroll wheel to adjust magnification, it feels better to me to have magnification change in smaller steps than 1 each time; 0.2 feels nicer to me - open to opinions from anyone else who tries it!
> > It can be made a config parameter if necessary - it is on the wayfire zoom plugin - but that seems to me to be something where it is better to make a sensible choice upfront than to allow the user to fiddle with it.
> 
> Good points. I think having something like 0.2 as a default but still provide a config option is fine for this. As an alternative, we could add an optional argument to the `ZoomIn` and `ZoomOut` actions (which then defaults to 0.2) instead of having a global setting.

OK, will make it a config parameter for now - will do that in the morning! I think that's tidier than having to add an argument to ZoomIn/Out, as that then gives scope for error if the user sets different values for in and out... ;)

src/theme.c
```diff
@@ -853,6 +854,9 @@ entry(struct theme *theme, const char *key, const char *value)
 	if (match_glob(key, "magnifier.filter")) {
 		theme->mag_filter = atoi(value);
 	}
+	if (match_glob(key, "magnifier.increment")) {
+		theme->mag_increment = atof(value);
```
Consolatis, line 858: Please use `set_float(value, theme->mag_increment)` here, that one is locale aware.

spl237: Thanks - didn't know about that function!

stefonarch:
Testing this I stumbled upon 2 things more related to how to trigger it:

```
    <keybind key="W-0">
      <action name="ToggleMagnify"/>
    </keybind>
    <keybind key="W-+">
      <action name="ZoomIn"/>
    </keybind>
    <keybind key="W--">
      <action name="ZoomOut"/>
    </keybind>
```
  
 W-+ works, but how about W-- ? Some special trick needed or impossible?

  I liked this tool under wayfire, where you can use Meta+Scroll for zooming in/out. I tried the following but it activates the magnifier at every scroll event without left mouse button pressed, but it doesn't if it's pressed:
 
```
<context name="Client">
      <mousebind button="Left" action="Scroll" direction="Up">
        <action name="ZoomIn"/>
      </mousebind>
      <mousebind button="Left" action="Scroll" direction="Down">
        <action name="ZoomOut"/>
      </mousebind>
....
```

Consolatis:
You could try `W-0x2d` (locale aware)

spl237:
> I liked this tool under wayfire, where you can use Meta+Scroll for zooming in/out. I tried the following but it activates the magnifier at every scroll event without left mouse button pressed, but it doesn't if it's pressed:
> 
> ```
> <context name="Client">
>       <mousebind button="Left" action="Scroll" direction="Up">
>         <action name="ZoomIn"/>
>       </mousebind>
>       <mousebind button="Left" action="Scroll" direction="Down">
>         <action name="ZoomOut"/>
>       </mousebind>
> ....
> ```

I don't think that is how mousebinds are supposed to work - in all the examples I have seen, you either bind to a button action (click / release / press / drag) for a particular mouse button (left / right), or to a scroll action and a direction for the scroll (up/down); you have bound to a scroll action and a direction, but also to a button without a corresponding button action, because "scroll" isn't a button action. As far as I can tell, that binding is not valid, but I may be wrong!

stefonarch:
> You could try `W-0x2d` (locale aware)
> 
> Edit: `W-minus` should work as well.

True for both, thanks! But I found `W-Up|down` more convenient, it's only that +/- are used in other apps too.

>As far as I can tell, that binding is not valid, but I may be wrong!

Probably, someone that it could work with a mouse instead of touchpad, but I think there's no way to have this although the manpage says "scroll" being an button action. 

```
Mouse buttons and directions can be combined with modifier-keys (shift (S), super/logo (W), control (C), alt (A), meta (M) and
           hyper (H)), for example: <mousebind button="A-Right" action="Press">

           Supported mouse actions include:
           •   Press: Pressing the specified button down in the context.
           •   Release: Releasing the specified button in the context.
           •   Click: Pressing and then releasing inside of the the context.
           •   DoubleClick: Two presses within the doubleClickTime.
           •   Drag: Pressing the button within the context, then moving the cursor.
           •   Scroll: Scrolling in specified direction in the context.
```

droc12345:
From context titlebar

        <action name="Unshade" />
        <action name="Focus" />
      </mousebind>
        <action name="Unfocus" />
        <action name="Shade" />
      </mousebind>

Note the use of direction

stefonarch:
Yes, the latter works, but the downside is that the view belows scrolls too I see.

droc12345:
Zoomin/out should consume the scroll, not propagate it.

stefonarch:
>Zoomin/out should consume the scroll, not propagate it.

https://github.com/labwc/labwc/assets/10681413/5f7c221e-2624-4157-bcd6-184e2272649f

spl237:
> > Zoomin/out should consume the scroll, not propagate it.
> 
That's a known problem - discussion on here - https://github.com/labwc/labwc/pull/1768

droc12345:
I knew that was being discussed.

To properly do zoom, bind wise is to have a global context (don't care what it's called)
that takes precedence over other contexts, and it needs to consume the scroll event (in this case)

Consolatis:
That is already handled via
- #1768

docs/labwc-actions.5.scd
```diff
@@ -264,6 +264,17 @@ Actions are used in menus and keyboard/mouse bindings.
...
 
+*<action name="ToggleMagnify">*
+	Toggle the screen magnifier on or off at the last magnification level
+	used.
+
+*<action name="ZoomIn">*++
+*<action name="ZoomOut">*
+	Increase or decrease the magnification level for the screen magnifier.
+	If the magnifier is currently off, ZoomIn will enable it at the lowest
+	magnification (x2). If the magnifier is on and at a magnification of
```
Consolatis, line 275: This looks like it needs an update

spl237: Well spotted! Hang on...

src/common/scene-helpers.c
```diff
@@ -34,6 +44,285 @@ lab_wlr_scene_get_prev_node(struct wlr_scene_node *node)
 	return prev;
 }
 
+static double constrain(double lower, double in, double upper)
+{
```
Consolatis, line 48: Tthis looks like it could be replaced by
```c
MAX(MIN(in, upper), lower)
```

Consolatis: We could also add a new `CLAMP(in, lower, upper)` macro which just does the `MAX(MIN())` dance.

Consolatis:
Had another quick test, including the new fullscreen option. Seems to work fine.
One thing I noticed is that when initially enabling the zoom the action is delayed until we have a first damage somewhere (e.g. hover buttons). That can likely be fixed with something like
```c
struct output *output = output_nearest_to_cursor(server);
if (output) {
	wlr_output_schedule_frame(output->wlr_output);
}
```
whenever enabling the magnifier.

Code wise I think we should move everything to `src/magnifier.c` or something along that line, that could also be done as follow-up of this PR, up to you.

src/common/scene-helpers.c
spl237: Yes, that sounds better - will do that.

spl237:
> One thing I noticed is that when initially enabling the zoom the action is delayed until we have a first damage somewhere (e.g. hover buttons).

Hmm. That's weird - I don't see that; I get the magnified view immediately on calling either of ZoomIn or ToggleMagnifier. It did initially not update until I added some code in lab_wlr_scene_output_commit to detect whether the magnifier state or zoom had changed since the last call, but once I added that, I got immediate update when the actions were called.

I wonder why I am seeing different behaviour? It'll be hard for me to know if a change has fixed this, as I don't see the problem you are seeing. Shall I just add your suggested additional lines to the end of magnify_toggle and magnify_set_scale, and you can let me know if it has worked?

Having everything in a separate magnifier file makes sense - will do that.

Consolatis:
The difference is likely that I test nested and that by default uses something similar in behavior to hardware cursors (e.g. moving the cursor does not produce damage nor requests a new frame as its handled by the parent compositor). The commit callchain is only triggered when there is a frame pending which in the case of nested on top of wayland only occurs when either the compositor changes the state of some scene node or some client informs the compositor about some damage on its own surface.

Thus the commit handler is never called in the first place unless we `wlr_output_schedule_frame()` manually.

Consolatis:
Yes, absolutely. I can verify if it works then.

spl237:
OK, now reorganised to have all the magnifier code in a separate module; have also added the CLAMP macro and the calls to force a refresh when enabling and disabling the magnifier. Hopefully this should now be reasonably close to being good to go...

src/magnifier.c
```diff
@@ -0,0 +1,296 @@
...
+			.clip = NULL,
+		};
+		wlr_render_pass_add_rect(tmp_render_pass, &bg_opts);
+	}
+
+	/* Paste the magnified result back into the output buffer */
+	if (!tmp_texture) {
+		tmp_texture = wlr_texture_from_buffer(server->renderer, tmp_buffer);
+		assert(tmp_texture);
+	}
```
Consolatis, line 174: AFAIR `wlr_texture_from_buffer()` may fail, thus this should `return;` instead of the assert() and the whole block should be moved to the top after the `wlr_log(WLR_ERROR, "Failed to allocate temporary magnifier buffer")` branch.

src/magnifier.c
```diff
@@ -0,0 +1,296 @@
...
+		cropped_height += y;
+		dst_y = y * -1;
+		y = 0;
+	}
+	cropped_width = MIN(cropped_width, (double)output_buffer->width - x);
+	cropped_height = MIN(cropped_height, (double)output_buffer->height - y);
+
+	/* (Re)create the temporary buffer if required */
+	if (tmp_buffer && (tmp_buffer->width != width || tmp_buffer->height != height)) {
+		wlr_log(WLR_ERROR, "tmp buffer size changed, dropping");
```
Consolatis, line 99: This could be a `WLR_DEBUG`

src/magnifier.c
```diff
@@ -0,0 +1,296 @@
...
+		wlr_log(WLR_ERROR, "Failed to allocate temporary magnifier buffer");
+		return;
+	}
+
+	/* Extract source region into temporary buffer */
+
+	struct wlr_render_pass *tmp_render_pass = wlr_renderer_begin_buffer_pass(
+		server->renderer, tmp_buffer, NULL);
+
+	/* FIXME, try to re-use the existing output texture instead */
```
Consolatis, line 120: I don't think we can do this so suggest removing the comment.

src/magnifier.c
spl237: Done

src/magnifier.c
spl237: Done

src/magnifier.c
spl237: Done

src/magnifier.c
```diff
@@ -0,0 +1,300 @@
...
+		&& fullscreen) {
+		return;
+	}
+
+	if (mag_scale == 0.0) {
+		mag_scale = theme->mag_scale;
+	}
+	if (mag_increment == 0.0) {
+		mag_increment = theme->mag_increment;
+	}
```
Consolatis, line 65: This causes an issue on the first `ZoomIn` action, it will set the mag_scale to `1.0` (`1.0 + 0.0`).
Can we use `theme->mag_increment` everywhere and remove the local var for `mag_increment`?

spl237: Ah yes - we can do now, because I am passing the server to the action functions, so they have access to the theme directly. Hang on...

src/magnifier.c
```diff
@@ -112,16 +112,26 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box
...
 	}
 
+	/* Paste the magnified result back into the output buffer */
+	if (!tmp_texture) {
+		tmp_texture = wlr_texture_from_buffer(server->renderer, tmp_buffer);
+	}
+	if (!tmp_texture) {
+		wlr_log(WLR_ERROR, "Failed to allocate temporary texture");
+		return;
+	}
```
Consolatis, line 122: This may cause us to run into the earlier `assert(tmp_texture)` when the dimension of the buffer change (tmp_buffer set but no tmp_texture). Suggest adding a `wlr_buffer_drop(tmp_buffer); tmp_buffer = NULL;` into the branch.

src/magnifier.c
spl237: Done

Consolatis:
> Just a comment on the settings: Should they really be in the themerc? It feels very weird for me to have them there, except for the border nothing really is related to the theme. I could imagine people being confused if the theme controls the options of an accessibility tool.

I agree, this feels a bit weird. I think we should keep at least the border settings in the theme, not sure about the size. The other others should indeed likely be handled within rc.xml. Thoughts @johanmalm ?

> OK, now reorganised to have all the magnifier code in a separate module; have also added the CLAMP macro and the calls to force a refresh when enabling and disabling the magnifier. Hopefully this should now be reasonably close to being good to go...

Can confirm that manually scheduling the frame fixed the issue I've seen while testing.
With the comment from @Flrian addressed this LGTM. Thanks for all your work on this PR!

spl237:
> > Just a comment on the settings: Should they really be in the themerc? It feels very weird for me to have them there, except for the border nothing really is related to the theme. I could imagine people being confused if the theme controls the options of an accessibility tool.
> 
> I agree, this feels a bit weird. I think we should keep at least the border settings in the theme, not sure about the size. The other settings should indeed likely be handled within rc.xml. Thoughts @johanmalm ?

I think the border settings should be in the same place, and having them in the theme makes sense, not least due to the theme already being able to parse colours. But the rest could go in rc.xml, and I think that is more suitable. Will move them back! ;)

> > OK, now reorganised to have all the magnifier code in a separate module; have also added the CLAMP macro and the calls to force a refresh when enabling and disabling the magnifier. Hopefully this should now be reasonably close to being good to go...
> 
> Can confirm that manually scheduling the frame fixed the issue I've seen while testing. With the comment from @Flrian addressed this LGTM. Thanks for all your work on this PR!

Excellent - many thanks for the review! (And your original work, without which I'd never have got anywhere with this...)

spl237:
Magnifier config moved into rc.xml as suggested; border appearance still in theme. I think all the relevant documentation has also been adjusted accordingly.

johanmalm:
> > I agree, this feels a bit weird. I think we should keep at least the border settings in the theme, not sure about the size. The other settings should indeed likely be handled within rc.xml. Thoughts @johanmalm ?
> 
> I think the border settings should be in the same place, and having them in the theme makes sense, not least due to the theme already being able to parse colours. But the rest could go in rc.xml, and I think that is more suitable. Will move them back! ;)

As done now feels good (mostly in `rc.xml` but border-width/color in `themerc`)

src/config/rcxml.c
```diff
@@ -1036,6 +1036,16 @@ entry(xmlNode *node, char *nodename, char *content)
 		}
 	} else if (!strcasecmp(nodename, "ignoreButtonReleasePeriod.menu")) {
 		rc.menu_ignore_button_release_period = atoi(content);
+	} else if (!strcasecmp(nodename, "width.magnifier")) {
+		rc.mag_width = atoi(content);
+	} else if (!strcasecmp(nodename, "height.magnifier")) {
+		rc.mag_height = atoi(content);
+	} else if (!strcasecmp(nodename, "initScale.magnifier")) {
+		rc.mag_scale = atoi(content);
```
Consolatis, line 1044: Just noticed that this might be set to `0` by the user.
This could then cause divisions by zero in magnify(), so may we should make sure that this is always > 0 in `post_processing()`.

spl237: To take the final question, no, I think 1.0 + increment is correct rather than starting at the default zoom. My intention was that default zoom is more useful when toggling magnifier on and off. Zooming feels more like something where you just turn the scroll wheel to wherever looks right, so starting at minimum feels right to me.

But yes, good point that the init scale can indeed be a fractional value, so float is more consistent with the increment being a float - will change that. And will also check for zeros - good spot!

Consolatis: > To take the final question, no, I think 1.0 + increment is correct rather than starting at the default zoom.

Yes I think you are right, feels more intuitive.

The only thing missing apart from the 0-check and s/int/float/ would then be squashing. Do you want to do that or should I do so when merging?

spl237: OK, that should all be done, I think.

Do please do the squash when you merge - probably easier that way. 

Many thanks again for all the help and feedback!

Consolatis: Merged :)

spl237: Cool! Many thanks again!

Issue #1588, opened 2024-03-05
Add support for listening and reconfiguring the keyboard settings from locale1


Conan-Kudo:
### Proposal

Please add support for listening and dynamically configuring keyboard settings based on [`org.freedesktop.locale1`](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.locale1.html).

### Use-cases

This is needed in order to handle cases like keyboard layout switching from the login manager (e.g. SDDM) or the installer (e.g. Calamares or Anaconda).

### References

Here are a couple of example implementations for other compositors:

* KWin: https://invent.kde.org/plasma/kwin/-/merge_requests/2921
* Sway: https://github.com/alebastr/sway-systemd/blob/main/src/locale1-xkb-config

ahesford:
Neither of your example cases would be addressed by monitoring DBus for these messages. An installer will set some appropriate configuration to indicate the default locale, while the display manager should be setting environment variables to indicate the locale even before labwc launches.

The DBus message is relevant when a user updates the locale settings whole labwc is running, but monitoring DBus seems outside the scope of labwc. Unless there is a Wayland protocol that can be used for this purpose, I think your best bet would be to adapt the KWin or Sway helpers to monitor the bus, then write locale variables to `~/.config/labwc/environment` and trigger a labwc reconfigure when locale messages are received.

Conan-Kudo:
Installers change the locale live in order to allow input to work as you progress through the install process.

johanmalm:
Hi @Conan-Kudo 
Thanks for bringing this up. I'm interested in exploring this as getting/setting keyboard layout is something we struggle with in a standardized way.
I don't know much about the topic.
Any idea what other clients use `org.freedesktop.locale1` ?
We've probably got some soul-searching between the core devs because we've been clinging on to Wayland/wlr protocols only for IPC - but that's primarily to avoid custom IPC (which this isn't in my mind).

Conan-Kudo:
SDDM is another one, as the keyboard layout switching logic for Wayland is intended to use the locale1 interface too. And SDDM can be used with a variety of compositors, such as the default Weston, KWin (from KDE), Sway, and so on.

Fedora's Initial Setup tool is built on top of Anaconda and will also use this interface. I expect we'll see more of this sort of thing over time.

Consolatis:
SDDM could also just set the env vars directly before launching the compositor. I get the point about installers and live environments though.

Conan-Kudo:
[...]

PR #1777, opened 2024-05-04
src/main: delay startup of applications until event loop is ready

Consolatis:
This mostly allows for using scripts that trigger a labwc SIGHUP to work as expected. Before this patch, there was a race between the event loop starting up and the autostart scripts executing.

One example of such a use-case is dynamically setting the keyboard layout or cursor theme via environment variable based on feedback from a dbus service like discussed in #1588.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -80,6 +80,35 @@ send_signal_to_labwc_pid(int signal)
 	kill(pid, signal);
 }
 
+struct idle_ctx {
+	struct server *server;
+	const char *primary_client;
+	const char *startup_cmd;
+};
+
+static void
+idle_callback(void *data)
+{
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -177,24 +206,16 @@ main(int argc, char *argv[])
 
 	menu_init(&server);
```
domo141, line 208: instead of the change below, would 

(void)wl_event_loop_dispatch(server.wl_event_loop, 0);

work ?

Consolatis: Potentially that could work (although I didn't look into it in detail). However, I think we should not interfere with eventloop internals here.

domo141:
right -- I see your point there.

johanmalm:
Thanks :smile: Nice improvement.

Issue #1753, opened 2024-04-24
Shaded XWayland window on top of another XWayland gets (mouse) events


domo141:
In case where XWayland window (e.g. urxvt) is stacked on top of another XWayland window (e.g. urxvt)
and the window on top is shaded, the shaded area gets the events even when focus is moved to
the window lower in stacking order.

tokyo4j:
> the shaded area gets the events even when focus is moved to the window lower in stacking order.

Sorry, what does "events" mean here?

Consolatis:
> Sorry, what does "events" mean here?

from IRC history:
> \<Domo> hmm, scroll-wheel events seem to go to shaded view (just now)
> \<Domo> and not just those. cannot drag scrollbar in urxvt window when shaded view (which would unshade over the scrollbar) is on top of it

johanmalm:
Consider `xcb_change_property()` as Openbox uses `XChangeProperty()` on shade/unshade.

`client_shade()`
https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/openbox/client.c#L3535

`client_change_wm_state()`
https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/openbox/client.c#L2502

`OBT_PROP_SETA32`
https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/obt/prop.h#L301-L303

`obt_prop_set_array32()`
https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/obt/prop.c#L529

Consolatis:
Agree that this is likely the better solution but we can't really use it without creating a temporary xcb connection.
We basically need support for that in wlroots / xwm.c.

For now:
- #1778

Would appreciate testing.

Consolatis:
Should be fixed on latest master.

PR #1778, opened 2024-05-04
xwayland: prevent shaded views from getting mouse events

Consolatis:
Fixes:
- #1753

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -127,6 +127,7 @@ struct view_impl {
 	void (*minimize)(struct view *view, bool minimize);
 	void (*move_to_front)(struct view *view);
 	void (*move_to_back)(struct view *view);
+	void (*shade)(struct view *view, bool shaded);
 	struct view *(*get_root)(struct view *self);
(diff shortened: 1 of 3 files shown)


johanmalm:
I think approach is fine until we can access the connection for `xcb_change_property()`.

Would we need to check if the view is shaded in `xwayland_adjust_stacking_order()` to prevent reverting the `XCB_STACK_MODE_BELOW` on changing workspaces?

Consolatis:
> Would we need to check if the view is shaded in `xwayland_adjust_stacking_order()` to prevent reverting the `XCB_STACK_MODE_BELOW` on changing workspaces?

`xwayland_adjust_stacking_order()` uses `view_move_to_front()` which in turn uses `xwayland_view_move_to_front()` which this patch adds a check for.

johanmalm:
> `xwayland_adjust_stacking_order()` uses `view_move_to_front()` which in turn uses `xwayland_view_move_to_front()` which this patch adds a check for.

Sorry - can't read properly.
LGTM
Have tested. Works well.

PR #1780, opened 2024-05-05
menu: reset parser state in `menu_finish()`

tokyo4j:
This fixes use-after-free in `fill_item()` on Reconfigure with invalid `menu.xml` like below:

```xml
<openbox_menu>
  <menu id="root-menu">
    <item id="rofi-run" label="Run command">
      <action name="Execute" command="rofi-run" />
    </item>
  </menu>
</openbox_menu>
```

Fixes #1776

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -960,6 +960,11 @@ void
 menu_finish(struct server *server)
 {
 	menu_free_from(server, NULL);
+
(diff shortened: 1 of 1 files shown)


Consolatis:
Nice catch!
Can't believe that we missed resetting the state vars for the menu.

PR #1781, opened 2024-05-05
NEWS.md: update for 0.7.2

johanmalm:
For Friday

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog]
 
 | Date       | All Changes   | wlroots version | lines-of-code |
 |------------|---------------|-----------------|---------------|
-| 2024-04-22 | [unreleased]  | 0.17.2          | 21300         |
+| 2024-05-10 | [0.7.2]       | 0.17.3          | 21368         |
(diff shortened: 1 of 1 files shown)


tokyo4j:
I suggest `- Prevent Chromium from crashing when started after a virtual keyboard is destroyed. #1789`.

johanmalm:
> I suggest `- Prevent Chromium from crashing when started after a virtual keyboard is destroyed. #1789`.

Thanks. Done.

NEWS.md
```diff
@@ -1326,7 +1345,8 @@ Compile with wlroots 0.12.0 and wayland-server >=1.16
   ShowMenu
 
 [Keep a Changelog]: https://keepachangelog.com/en/1.0.0/
-[unreleased]: https://github.com/labwc/labwc/compare/0.7.1...HEAD
+[unreleased]: https://github.com/labwc/labwc/compare/0.7.2...HEAD
+[0.7.2]: https://github.com/labwc/labwc/compare/0.7.0...0.7.2
```
Consolatis, line 1349: this should likely compare to 0.7.1 rather than 0.7.0

johanmalm: Thanks for catching. Will change tomorrow.

PR #1786, opened 2024-05-07
view: fix top-layer not showing with minimized full-screen view

fberg:
If a full-screen window is minimized, top-layer views like waybar are kept hidden. This PR tries to fix this behavior.

I noticed this with some Wine games when they minimized themselves after losing focus (for instance by using the window switcher), but of course it's easy to reproduce once you know about it.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -648,6 +648,11 @@ view_minimize(struct view *view, bool minimized)
 	struct view *root = view_get_root(view);
 	_minimize(root, minimized);
 	minimize_sub_views(root, minimized);
+
+	/* Enable top-layer when full-screen views are minimized */
+	if (view->fullscreen && view->output) {
+		desktop_update_top_layer_visiblity(view->server);
+	}
 }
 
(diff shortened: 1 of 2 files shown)


src/view.c
```diff
@@ -648,6 +648,11 @@ view_minimize(struct view *view, bool minimized)
 	struct view *root = view_get_root(view);
 	_minimize(root, minimized);
 	minimize_sub_views(root, minimized);
+
+	/* Enable top-layer when full-screen views are minimized */
+	if (view->output) {
+		desktop_update_top_layer_visiblity(view->server);
+	}
```
Consolatis, line 655: Nice catch, thanks for the PR!

Can you add a check for fullscreen here (e.g. `view->fullscreen && view->output`)?
Otherwise we recalculate the visibility on every unrelated minimize event.

fberg: Done.

Consolatis:
Thanks :)

Issue #1770, opened 2024-05-03
Chromuim 124 cashed in labwc 7.1(Vmware 17.5.1 arch linux)


knm100:
labwc 7.1+arch+vmware 17.5.1
Since Chromium was upgraded to version 124, it crashes frequently and irregularly when launching from the menu. However, it has been launching fine from the terminal. Every time it crashes, it has to be launched from the terminal once before it can be launched from the menu. Otherwise, it never starts from the menu.

knm100:
It seems to have been fixed in the latest GIT rev. 0.7.1.r148.g4b6efb73-1-x86_64.   More testing is needed to confirm.

knm100:
It feels like it's happening randomly, with no pattern at all

tokyo4j:
Setting up menu like below may help debugging.
```xml
  <menu id="root-menu">
    <item label="Chromium">
      <action name="Execute" command="sh -c 'exec chromium --enable-logging=stderr 2> chromium-log.txt'" />
    </item>
  </menu>
```

knm100:
Here is the log
[chromium-log.txt](https://github.com/labwc/labwc/files/15200214/chromium-log.txt)

tokyo4j:
Maybe something is going wrong with GPU, from the output:
```
[5190:8:0503/210426.968552:ERROR:command_buffer_proxy_impl.cc(131)] ContextResult::kTransientFailure: Failed to send GpuControl.CreateCommandBuffer.
```
But I can't tell anything from this other that.

BTW, I sometimes have a similar experience with VSCode. When I start VSCode from the menu, sometimes a window appears but crashes immediately.

The difference between running commands from terminal and from menu is:
- From terminal, the command process is spawned as a child process of the terminal.
- From menu, the command process is spawned as an orphan process, thus a child process of `systemd`.

So, I patched labwc to spawn an action command as a child process of `labwc` to see if it solves the problem.

https://github.com/tokyo4j/labwc/tree/no-orphan

Then, I ran patched/unpatched labwc alternately and opened VSCode from menu 3 times in each session.
The result is shown below. VSCode never crashed in the patched labwc, while it crashed with 2/3 chance in the unpatched labwc.

@knm100 Could you try my patch and see it solves your problem with Chromium?

|patched|crashed
|-|-|
|yes|no|
||no|
||no|
|no|no|
||yes|
||yes|
|yes|no|
||no|
||no|
|no|no|
||yes|
||yes|
|yes|no|
||no|
[...]

PR #1789, opened 2024-05-08
seat: prevent Chromium from crashing when active keyboard is gone

tokyo4j:
Slurp and Chromium expect `wl_keyboard.keymap` event to be sent before `wl_keyboard.modifiers` event.
Normally, `wl_keyboard.keymap` event is sent on the client first obtains `wl_keyboard` with `wl_seat.get_keyboard` request.
However, after the active (especially virtual) keyboard is destroyed, wlroots doesn't respond to `wl_seat.get_keyboard` request with `wl_keyboard.keymap` event since there's no active keyboard on the seat.

Therefore, if we run commands like `sleep 2; wtype hello; chromium` (without fcitx5), the active keyboard is destroyed when `wtype` finishes, and `wl_keyboard.modifiers` event is sent to Chromium when Chromium first maps the surface and the keyboard focus is moved to it, then Chromium crashes. I guess something similar is happening in #1770.

https://github.com/labwc/labwc/assets/22029524/9d09f54f-61f2-43ca-ae4e-23d4dbe343c8

I found the same problem happens even when a physical keyboard is unplugged.

So I fixed this by making the active keyboard fall back to `seat->keyboard_group->keyboard` when it's destroyed. The active keyboard is then updated to the individual keyboard as soon as the user types something.

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -31,6 +31,19 @@ input_device_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&keyboard->key.link);
 		wl_list_remove(&keyboard->modifier.link);
 		keyboard_cancel_keybind_repeat(keyboard);
+		/*
+		 * If the active keyboard on the seat is destroyed, fall back
+		 * to the keyboard from keyboard group so we can respond to
+		 * wl_seat.get_keyboard requests with wl_keyboard.keymap event.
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, issue and logic to solve it is straight forward but I didn't personally test. Vote for having this PR be part of 0.7.2.

PR #1805, opened 2024-05-13
desktop: show/hide `top` layer more smartly

### LINKED ISSUE #1792: Show layershell panel when another window is rendered on top of a fullscreen one
Someone correct me if I've got this wrong, but supposedly 0.7.2 fixed this bug https://github.com/labwc/labwc/issues/1752 where alt tabbing out of a full screen window like steam big picture or a full screen mpv video would result in waybar being below the full screen window that was just alt tabbed out of, I just upgraded to the latest labwc-git, running labwc -v in the terminal gives me 0.7.2, but I'm still experiencing this bug, thanks.

tokyo4j:
Fixes #1792

A caveat is that panels in `bottom` layer is still always hidden by a fullscreen window.

Testing is welcome.

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -229,17 +229,23 @@ desktop_update_top_layer_visiblity(struct server *server)
 		wlr_scene_node_set_enabled(&output->layer_tree[top]->node, true);
 	}
 
-	/* And disable them again when there is a view in fullscreen */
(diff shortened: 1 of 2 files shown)


Consolatis:
Tested and works great, thanks!

PR #1813, opened 2024-05-13
keybind: allow keybinding "-"

CosmicToast:
This is entirely untested, as of yet.
I'll test it tomorrow if no one gets around to it :)

### DIFF
--- a/src/config/keybind.c
+++ b/src/config/keybind.c
@@ -124,6 +124,25 @@ keybind_create(const char *keybind)
 	gchar **symnames = g_strsplit(keybind, "-", -1);
 	for (size_t i = 0; symnames[i]; i++) {
 		char *symname = symnames[i];
+		/*
+		 * Since "-" is used as a separator, a keybind string like "W--"
+		 * becomes "W", "", "". This means that it is impossible to bind
+		 * an action to the "-" key in this way.
+		 * We detect empty ""s outputted by g_strsplit and treat them as
+		 * literal "-"s.
+		 */
+		if (!symname[0]) {
+			/*
+			 * You might have noticed that in the "W--" example, the
+			 * output is "W", "", ""; which turns into "W", "-",
+			 * "-". In order to avoid such duplications, we perform
+			 * a lookahead on the tokens to treat that edge-case.
+			 */
+			if (symnames[i+1] && !symnames[i+1][0]) {
+				continue;
+			}
+			symname = "-";
+		}
 		uint32_t modifier = parse_modifier(symname);
 		if (modifier != 0) {
 			k->modifiers |= modifier;


CosmicToast:
Latest force-push addresses a concern about W-- translating to "W", "", ""; and thus "W", "-", "-".
The approach used is to perform a lookahead on the tokens and capture a following "-", only when we're currently processing a "-".

tokyo4j:
- Suggest adding a comment like:
```c
/*
 * Consecutive whitespaces are converted to a minus
 * (e.g. W-- => "W","","" => "W","-")
 */
```
- Why did you define a variable `minus`? Did you get any compile error just with `symname = "-"`?

CosmicToast:
> I suggest adding a comment.

That's a good idea.
I did think of it after I went to bed (I wrote this at 11pm local time), and will do so for sure now.

> Why did you define a variable minus?

I write code like this by default.
If symname can be made a `const char *` and minus a `const char const *`, it unlocks quite a few desirable compiler optimizations.
Additionally, it prevents potential drift, since it makes it clear as to what it's for.

However, thinking back on it, it should probably be inside the function, and be called "sep".
I'll make these changes on my lunch break after I test the functionality.

tokyo4j:
> it unlocks quite a few desirable compiler optimizations.

Really? I did `objdump -d build/labwc` after `meson setup build -Dreleasetype=release && ninja -C build` and saw almost no difference even when I updated `symname` to `const char *`.

<details><summary>source diff</summary>

```diff
diff --git a/src/config/keybind.c b/src/config/keybind.c
index c1bd05fe..274362b3 100644
--- a/src/config/keybind.c
+++ b/src/config/keybind.c
@@ -115,20 +115,22 @@ keybind_update_keycodes(struct server *server)
 	}
 }
 
+static const char *const minus = "-";
+
 struct keybind *
 keybind_create(const char *keybind)
 {
 	xkb_keysym_t sym;
 	struct keybind *k = znew(*k);
 	xkb_keysym_t keysyms[MAX_KEYSYMS];
-	gchar **symnames = g_strsplit(keybind, "-", -1);
+	gchar **symnames = g_strsplit(keybind, minus, -1);
 	for (size_t i = 0; symnames[i]; i++) {
 		const char *symname = symnames[i];
 		if (!symname[0]) {
 			if (symnames[i+1] && !symnames[i+1][0]) {
 				continue;
 			}
-			symname = "-";
+			symname = minus;
 		}
 		uint32_t modifier = parse_modifier(symname);
 		if (modifier != 0) {

```
</details>

<details><summary>objdump diff</summary>

```diff
--- objdump-no-opt.txt	2024-05-14 19:13:57.387840816 +0900
+++ objdump-opt.txt	2024-05-14 19:14:08.794593195 +0900
@@ -47380,7 +47380,7 @@
    3ad1c:	0f 85 00 ff ff ff    	jne    3ac22 <keybind_create+0xe2>
    3ad22:	66 0f 1f 44 00 00    	nopw   0x0(%rax,%rax,1)
    3ad28:	49 89 d8             	mov    %rbx,%r8
-   3ad2b:	b9 99 00 00 00       	mov    $0x99,%ecx
+   3ad2b:	b9 9b 00 00 00       	mov    $0x9b,%ecx
    3ad30:	48 8d 15 c0 43 01 00 	lea    0x143c0(%rip),%rdx        # 4f0f7 <_IO_stdin_used+0x20f7>
    3ad37:	bf 01 00 00 00       	mov    $0x1,%edi
    3ad3c:	48 8d 35 ff 43 01 00 	lea    0x143ff(%rip),%rsi        # 4f142 <_IO_stdin_used+0x2142>
@@ -47394,7 +47394,7 @@
    3ad60:	48 8d 1d 90 43 01 00 	lea    0x14390(%rip),%rbx        # 4f0f7 <_IO_stdin_used+0x20f7>
    3ad67:	31 c0                	xor    %eax,%eax
    3ad69:	41 b8 20 00 00 00    	mov    $0x20,%r8d
-   3ad6f:	b9 a1 00 00 00       	mov    $0xa1,%ecx
+   3ad6f:	b9 a3 00 00 00       	mov    $0xa3,%ecx
    3ad74:	48 89 da             	mov    %rbx,%rdx
    3ad77:	48 8d 35 a2 85 01 00 	lea    0x185a2(%rip),%rsi        # 53320 <_IO_stdin_used+0x6320>
    3ad7e:	bf 01 00 00 00       	mov    $0x1,%edi
@@ -47402,7 +47402,7 @@
    3ad88:	4c 8b 44 24 08       	mov    0x8(%rsp),%r8
    3ad8d:	48 89 da             	mov    %rbx,%rdx
    3ad90:	31 c0                	xor    %eax,%eax
-   3ad92:	b9 a3 00 00 00       	mov    $0xa3,%ecx
+   3ad92:	b9 a5 00 00 00       	mov    $0xa5,%ecx
    3ad97:	48 8d 35 ca 85 01 00 	lea    0x185ca(%rip),%rsi        # 53368 <_IO_stdin_used+0x6368>
    3ad9e:	bf 01 00 00 00       	mov    $0x1,%edi
    3ada3:	e8 28 47 fd ff       	call   f4d0 <_wlr_log@plt>

```
</details>

>  Additionally, it prevents potential drift, since it makes it clear as to what it's for.

I don't think substituting `"-"` with a variable makes it clear. Hardcoding is not always bad.

Sorry for nitpicking, but I'd like to keep the codebase as simple as possible.

src/config/keybind.c
```diff
@@ -124,6 +124,25 @@ keybind_create(const char *keybind)
 	gchar **symnames = g_strsplit(keybind, "-", -1);
 	for (size_t i = 0; symnames[i]; i++) {
 		char *symname = symnames[i];
+		/*
+		 * Since "-" is used as a separator, a keybind string like "W--"
+		 * becomes "W", "", "". This means that it is impossible to bind
+		 * an action to the "-" key in any way.
```
Consolatis, line 130: This is not technically correct, it can be defined like `W-minus` or `W-0x21` (in for example the `us` layout) so I suggest `s/any way/this way/`.

Consolatis:
Tested and works fine, with the comment update above this LGTM. Thanks!

src/config/keybind.c
CosmicToast: Done.

PR #1816, opened 2024-05-14
seat: prevent xwayland from using incorrect keymap

### LINKED ISSUE #1796: wtype not working with labwc 0.7.2
I was on labwc 0.7.0, which was working fine with wtype.

Since upgrading to 0.7.2, wtype stopped working.

For example, if I click on a file in my GUI file manager running in xwayland, running this command...

```
sleep 5; wtype -M ctrl c -m ctrl
```
 
...is supposed to send Ctrl+c to the window and put the filename in the clipboard. Since 0.7.2 upgrade, nothing happens when running the above command (i.e., clipboard contains whatever I last copied into it manually).

Any idea what broke labwc's ability to play nice with wtype?

tokyo4j:
Fixes #1796

#### commit message:
With commit https://github.com/labwc/labwc/commit/cafdcd8e19c99d141afb5a631d1eaadad0a9f5f4, the keyboard from the keyboard group is set to the
seat when the active keyboard is destroyed, but this broke `wtype`
against XWayland surfaces. This is likely because XWayland caches
keycodes from the compositor and convert them with the last keymap
received from the compositor. So when we run `wtype a` over a XWayland
surface, the keycode for `a` is cached by XWayland, the keymap from the
keyboard group is sent to XWayland, then Xwayland converts the keycode
using the keymap from the keyboard group, not the one `wtype` set.

This commit fixes this by setting the keyboard from the keyboard group
to the seat when the keyboard focus is moved to a surface, not when the
active keyboard is destroyed.

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -31,19 +31,6 @@ input_device_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&keyboard->key.link);
 		wl_list_remove(&keyboard->modifier.link);
 		keyboard_cancel_keybind_repeat(keyboard);
-		/*
-		 * If the active keyboard on the seat is destroyed, fall back
-		 * to the keyboard from keyboard group so we can respond to
-		 * wl_seat.get_keyboard requests with wl_keyboard.keymap event.
-		 * This prevents Chromium from crashing when started just after
-		 * the active keyboard is destroyed.
-		 */
-		struct wlr_keyboard *active_keyboard =
-			wlr_seat_get_keyboard(input->seat->seat);
-		if (!active_keyboard || active_keyboard == keyboard->wlr_keyboard) {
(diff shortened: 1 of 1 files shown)


Consolatis:
Fine with the workaround here. Although this seems to cover up two separate bugs:
- wlroots bug in that it does send the modifiers without having sent the keymap first
- xwayland bug in that it doesn't seem to deal with receiving keycodes and a new layout immediately after

tokyo4j:
Thank you for clarification.

> wlroots bug in that it does send the modifiers without having sent the keymap first

Though this is expected to be one of the causes of `chromium`/`slurp`'s crash, I'm not sure we can say it's a "bug", because wayland protocol says nothing about this. Perhaps the protocol should be updated first or we should say it's a client-side bug.

johanmalm:
Thanks.

PR #1817, opened 2024-05-14
Add support for security-context-v1

nesteroff:
This adds support for the [security-context-v1](https://wayland.app/protocols/security-context-v1) protocol. It is used by sandbox engines to restrict the features that sandboxed connections can use. For example, it is currently supported by Flatpak.

The implementation is based on the Sway code but adapted to Labwc. The `sandbox_engine` and `sandbox_app_id` parameters are added to window rules in order to be able to apply different rules for sandboxed clients.

The easiest way to test it is probably to take a recent version of Flatpak. Use it to run a Wayland application (for example, Firefox) with `WAYLAND_DEBUG` enabled and check for `wp_security_context` in the logs.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -189,6 +189,12 @@ fill_window_rule(char *nodename, char *content)
 		current_window_rule->window_type = parse_window_type(content);
 	} else if (!strcasecmp(nodename, "matchOnce")) {
 		set_bool(content, &current_window_rule->match_once);
+	} else if (!strcasecmp(nodename, "sandboxEngine")) {
+		free(current_window_rule->sandbox_engine);
+		current_window_rule->sandbox_engine = xstrdup(content);
+	} else if (!strcasecmp(nodename, "sandboxAppId")) {
+		free(current_window_rule->sandbox_app_id);
+		current_window_rule->sandbox_app_id = xstrdup(content);
 
 	/* Event */
 	} else if (!strcmp(nodename, "event")) {
@@ -323,6 +329,10 @@ fill_action_query(char *nodename, char *content, struct action *action)
 		current_view_query->title = xstrdup(content);
 	} else if (!strcmp(nodename, "type")) {
 		current_view_query->window_type = parse_window_type(content);
+	} else if (!strcasecmp(nodename, "sandboxEngine")) {
+		current_view_query->sandbox_engine = xstrdup(content);
+	} else if (!strcasecmp(nodename, "sandboxAppId")) {
+		current_view_query->sandbox_app_id = xstrdup(content);
 	}
 }
 
@@ -1508,6 +1518,8 @@ rule_destroy(struct window_rule *rule)
 	wl_list_remove(&rule->link);
 	zfree(rule->identifier);
 	zfree(rule->title);
+	zfree(rule->sandbox_engine);
+	zfree(rule->sandbox_app_id);
 	action_list_free(&rule->actions);
 	zfree(rule);
 }
@@ -1576,7 +1588,8 @@ validate(void)
 	/* Window-rule criteria */
 	struct window_rule *rule, *rule_tmp;
 	wl_list_for_each_safe(rule, rule_tmp, &rc.window_rules, link) {
-		if (!rule->identifier && !rule->title && rule->window_type < 0) {
+		if (!rule->identifier && !rule->title && rule->window_type < 0
+				&& !rule->sandbox_engine && !rule->sandbox_app_id) {
 			wlr_log(WLR_ERROR, "Deleting rule %p as it has no criteria", rule);
 			rule_destroy(rule);
 		}
--- a/src/view.c
+++ b/src/view.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <strings.h>
 #include <wlr/types/wlr_output_layout.h>
+#include <wlr/types/wlr_security_context_v1.h>
 #include "common/macros.h"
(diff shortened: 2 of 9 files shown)


Consolatis:
First of all, thanks for the PR.

What is the purpose of adding support for the sandbox manager engine / app_id to the window rules?

I am not that much in favor of hardcoding a list of "privileged" protocols for any sandboxed client, it seems cleaner to allow the user to define the protocols instead based on the sandbox engine name and potentially the sandbox engine app_id.

nesteroff:
> What is the purpose of adding support for the sandbox manager engine / app_id to the window rules?

It can be used to help users visually identify applications running in sandboxes. One use case is when multiple instances of a single application are running under different security policies. For example, a user might have two browsers running: one for untrusted web apps, which is isolated in a sandbox, and another running normally. In this case, the user can configure different window styles with different colors to quickly identify which is which.
 
> I am not that much in favor of hardcoding a list of "privileged" protocols for any sandboxed client, it seems cleaner to allow the user to define the protocols instead based on the sandbox engine name and potentially the sandbox engine app_id.

Sure, it definitely needs to be more flexible. I just thought we could start with a simple implementation, where there is a reasonable set of protocols blocked for sandboxed clients by default. Later on, we could add the ability to whitelist or blacklist other protocols.

I guess it could be something like this:
```xml
<sandboxRules>
    <sandboxRule app_id="org.mozilla.firefox" engine="org.flatpak">
        <protocol name="output_power_manager_v1" restrict="no" />
    </sandboxRule>
</sandboxRules>
```

Consolatis:
> > I am not that much in favor of hardcoding a list of "privileged" protocols for any sandboxed client, it seems cleaner to allow the user to define the protocols instead based on the sandbox engine name and potentially the sandbox engine app_id.
> 
> Sure, it definitely needs to be more flexible. I just thought we could start with a simple implementation, where there is a reasonable set of protocols blocked for sandboxed clients by default. Later on, we could add the ability to whitelist or blacklist other protocols.
> 
> I guess it could be something like this:

If we make the hardcoded list depend on the sandbox engine being `org.flatpak` for now I would feel more comfortable with that approach.

Just for reference, some earlier work (without the security context protocol but single wayland sockets instead):
- #1004

src/window-rules.c
```diff
@@ -13,11 +14,25 @@
...
 
+static const struct wlr_security_context_v1_state *
+view_security_context(struct view *view)
+{
+	if (view && view->surface && view->surface->resource) {
+		struct wl_client *client = wl_resource_get_client(view->surface->resource);
+		const struct wlr_security_context_v1_state *security_context =
+		wlr_security_context_manager_v1_lookup_client(
+		view->server->security_context_manager_v1, client);
+		return security_context;
```
Consolatis, line 25: This could just be
```c
return wlr_security_context_manager_v1_lookup_client(
	view->server->security_context_manager_v1, client);
```

src/window-rules.c
```diff
@@ -29,6 +44,18 @@ matches_criteria(struct window_rule *rule, struct view *view)
 			return false;
 		}
 	}
+	if (rule->sandbox_engine) {
+		if (!security_context || !security_context->sandbox_engine
+			|| !match_glob(rule->sandbox_engine, security_context->sandbox_engine)) {
```
Consolatis, line 49: code style nit: double indent for multi line ifs. Same in the next condition.

src/config/rcxml.c
```diff
@@ -185,6 +185,12 @@ fill_window_rule(char *nodename, char *content)
 	} else if (!strcmp(nodename, "title")) {
 		free(current_window_rule->title);
 		current_window_rule->title = xstrdup(content);
+	} else if (!strcmp(nodename, "sandbox_engine")) {
+		free(current_window_rule->sandbox_engine);
+		current_window_rule->sandbox_engine = xstrdup(content);
+	} else if (!strcmp(nodename, "sandbox_app_id")) {
+		free(current_window_rule->sandbox_app_id);
+		current_window_rule->sandbox_app_id = xstrdup(content);
```
Consolatis, line 193: Nit: we usually use the `sandboxEngine` format for user facing settings.

nesteroff:
Sorry for the delay. I was traveling last week. The code has been updated and rebased. Please have a look.

src/view.c
```diff
@@ -67,6 +81,8 @@ view_matches_query(struct view *view, struct view_query *query)
 {
 	bool match = true;
 	bool empty = true;
+	const struct wlr_security_context_v1_state *security_context =
```
johanmalm, line 84: I haven't analysed this, but suspect `view_matches_query()` is called a lot, thus making anything within it is therefore an expensive operation.

Unless I'm wrong in that, would it make sense to declare + initialize it to NULL at the top; and then set it to `security_context_from_view()` iff `match && (query->sandbox_engine || query->sandbox_app_id)`?

johanmalm:
LGTM
Keen to understand what the next PR might look like - but not a prerequisite to merging.

src/view.c
nesteroff: Sure, I reduced the scope of the security_context variable to be used only when needed.

nesteroff:
> LGTM Keen to understand what the next PR might look like - but not a prerequisite to merging.

Thanks. I think the next step is to add the ability to configure which protocols are available for sandboxed clients, similar to what @Consolatis shared in this thread. It probably shouldn't be too hard to adapt it to the secure context.

johanmalm:
Thanks. Excellent work.

PR #1819, opened 2024-05-15
impr: small tablet improvements

jp7677:
These are the first three commits from https://github.com/labwc/labwc/pull/1678 which are imho a good idea regardless of support for more axis. See the separate commits for what gets improved.


johanmalm:
Thanks.

PR #1820, opened 2024-05-16
Disable magnifier for pixman renderer

tokyo4j:
I tried magnifier and it works very nicely, but I found that magnifier becomes quite laggy and damage tracking doesn't work correctly when I enabled magnifier with `WLR_RENDERER=pixman`. So this PR disables magnifier for pixman renderer for now.

The first commit is just a minor refactor.

### DIFF
--- a/include/magnifier.h
+++ b/include/magnifier.h
@@ -3,10 +3,11 @@
 #define LABWC_MAGNIFIER_H
 
 #include <stdbool.h>
-#include <wayland-server-core.h>
 
 struct server;
 struct output;
+struct wlr_buffer;
+struct wlr_box;
 
 enum magnify_dir {
 	MAGNIFY_INCREASE,
--- a/src/magnifier.c
+++ b/src/magnifier.c
@@ -2,15 +2,15 @@
 
 #include <assert.h>
 #include <wlr/types/wlr_output.h>
-#include "magnifier.h"
-#include "labwc.h"
-#include "theme.h"
 #include "common/macros.h"
+#include "labwc.h"
+#include "magnifier.h"
+#include "theme.h"
 
-bool magnify_on;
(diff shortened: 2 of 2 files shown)


Consolatis:
Nice catch, there are indeed issues with pixman. The border isn't shown and the damage tracking doesn't work correctly.
As the border is just a rect that is also used for damage tracking those two issues do likely have the same cause.

I suspect that there is some 0.5 difference in the pixman renderer (e.g. pixel coordinates at center or on the top-left) so I'd prefer if we can work around it instead of disabling the magnifier outright. In the wlroots master branch [there is a fix for that](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4659/diffs?commit_id=77006e55652c9e2777ad51755cea30e3ed6142a8).

tokyo4j:
> I suspect that there is some 0.5 difference in the pixman renderer (e.g. pixel coordinates at center or on the top-left) so I'd prefer if we can work around it instead of disabling the magnifier outright. In the wlroots master branch [there is a fix for that](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4659/diffs?commit_id=77006e55652c9e2777ad51755cea30e3ed6142a8).

Thank you for the information.

As I don't know how wlroots renderer works, I won't make a workaround for it with this PR.

Should I drop the second commit and leave the fix to you (or other devs more familiar with wlroots renderer)? Or should we merge this PR so we can fix it later or wait for wlroots version-up?

Consolatis:
Thoughts @spl237 ?

tokyo4j:
OK, let's merge only the first commit. Then I'll create an issue and add it to 0.7.3 milestone.

Consolatis:
:)

PR #1823, opened 2024-05-16
IME: support multiple IME popups

### LINKED ISSUE #1794: request to allow multiple IME popups
at src/input/ime.c, there says
                 * With current input-method-v2 protocol, creating multiple IME
		 * popups is not useful because they just stack up at the same
		 * position.

It's not true, I'm a china input method author, in my case there need two popup, one for normal candidate panel, one for tip the current status which shown at focus or some key pressed.

tokyo4j:
Closes #1794.

We didn't support multiple IME popups since input-method-v2 protocol has no way to position them individually, but we should support it to provide IME developers with more programming flexibility.

### DIFF
--- a/include/input/ime.h
+++ b/include/input/ime.h
@@ -29,7 +29,7 @@ struct input_method_relay {
 	 */
 	struct text_input *active_text_input;
 
-	struct wlr_input_popup_surface_v2 *popup_surface;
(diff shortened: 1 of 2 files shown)


Consolatis:
LGTM, didn't test. Feel free to rebase and merge if you are satisfied with the PR.

PR #1832, opened 2024-05-20
Mag fix scene helper

droc12345:
This fixes the logic (I think) inside lab_wlr_scene_output_commit in scene-helpers.c

### DIFF
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -47,14 +47,11 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output)
 	struct wlr_output_state *state = &wlr_output->pending;
 	struct output *output = wlr_output->data;
 	bool wants_magnification = output_wants_magnification(output);
-	static bool last_mag = false;
 
 	if (!wlr_output->needs_frame && !pixman_region32_not_empty(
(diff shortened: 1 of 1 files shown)


tokyo4j:
I tried this PR and the magnifier wasn't drawn when I'm not moving the cursor.

I think the fix can be simpler like this:

```diff
diff --git a/src/common/scene-helpers.c b/src/common/scene-helpers.c
index f09cd9c0..536fa53d 100644
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -51,7 +51,7 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output)
 
 	if (!wlr_output->needs_frame && !pixman_region32_not_empty(
 			&scene_output->damage_ring.current) && !wants_magnification
-			&& last_mag != is_magnify_on()) {
+			&& last_mag == is_magnify_on()) {
 		return false;
 	}
```

I confirmed the CPU time consumption reduced from around 5% to less than 1% with the fix above.

droc12345:
Ok.
I'll close this PR, and you can add the fix to whatever else you're doing in that area.

tokyo4j:
Anyway, thank you for your report!
We should never miss performance degradation like this.

Issue #1822, opened 2024-05-16
Magnifier is broken for pixman renderer


tokyo4j:
See #1820.

On pixman renderer, the border of magnifier is not drawn and damage tracking for magnifier doesn't work correctly.

According to @Consolatis, this is already fixed in wlroots (https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4659).

I think either of following options should be taken before the next release:
- Confirm the wlroots fix is backported and the issue is solved
- Highly discourage users from using magnifier on pixman renderer
- Disable magnifier for pixman renderer

Consolatis:
My 2 cents:
Variant 1 would be the best, variant 3 is a good fallback approach, 2 doesn't sound that great.

For anybody wanting to test the fixes backported to wlroots 0.17:
https://gitlab.freedesktop.org/Consolatis/wlroots/-/tree/backport_017/pixman_render_perf

At least nested it did fix the pixman + magnifier issue for me but also causes a pixman + dropshadow issues instead so there is further work to be done on the wlroots side.

CC @spl237 @cillian64

droc12345:
I'm not sure the logic is 100% correct in scene-helpers.c
I upgraded to the latest git yesterday, and I noticed my idle watts went up from 3.1 to 5.1 (on average)
I chased it down to lab_wlr_scene_output_commit in scene-helpers..

I reworked it slightly and my watts went back to normal.
Basically I moved the checks for last_mag and wants_magnification to where they made sense.
I haven't tested the magnifier part, as I've not used it yet.

If interested I'll pull up a PR.

spl237:
Just to confirm, I've tried the wlroots patch referred to above on Pi, and it does indeed fix the magnifier under pixman for me as well, both windowed and full-screen. (And does also break drop shadows!)

droc12345:
Try this patch #1832  and see if it still works for magnification. 
I think it should I just rearranged and added extra checks.

Consolatis:
It seems like the magnifier and drop shadows are both [fixed](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4690) in wlroots master / 0.18 (not yet tested though). However there is some hesitation to backport the two changesets to 0.17 (as its more than just a bugfix). So the question is if we can fix the wlroots 0.17 branch separately for this magnifier bug. Thoughts @cillian64 ?

cillian64:
[...]

PR #1833, opened 2024-05-20
magnifier: fix high power consumption even with magnifier disabled

tokyo4j:
Fixes the high power consumption issue reported by @droc12345.

My fix is based on my understanding that the output state should be committed when the cursor has been moved or magnifier on/off state is changed.

I confirmed the CPU consumption reduced from around 5% to lower than 1% with this fix.

### DIFF
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -47,15 +47,17 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output)
 	struct wlr_output_state *state = &wlr_output->pending;
 	struct output *output = wlr_output->data;
 	bool wants_magnification = output_wants_magnification(output);
-	static bool last_mag = false;
 
+	/*
+	 * FIXME: Regardless of wants_magnification, we are currently adding
+	 * damages to next frame when magnifier is shown, which forces
+	 * rendering on every output commit and overloads CPU.
+	 * We also need to verify the necessity of wants_magnification.
+	 */
 	if (!wlr_output->needs_frame && !pixman_region32_not_empty(
-			&scene_output->damage_ring.current) && !wants_magnification
-			&& last_mag != is_magnify_on()) {
+			&scene_output->damage_ring.current) && !wants_magnification) {
 		return false;
 	}
-
-	last_mag = is_magnify_on();
 
 	if (!wlr_scene_output_build_state(scene_output, state, NULL)) {
 		wlr_log(WLR_ERROR, "Failed to build output state for %s",


droc12345:
As I mentioned in #1822, it does fix the increased power draw I was seeing.
Looks good from my end, others can check that the magnification still works, 
between this and the wlroots pixman fix, maybe this will fix all the problems.

Consolatis:
I think this is correct.
However, I am not sure if we need the `last_mag` state in the first place:

When enabling the magnifier (either via `magnify_toggle()` or `magnify_set_scale()`) we do call `wlr_output_schedule_frame()` which should set `wlr_output->needs_frame` and thus already skip the early out.

droc12345:
I was also pondering whether we need the last_mag, all we really care about is
do they want magnification and is the mag flag set (these conditions should always match, IMO, but who knows), I don't see the need to care about last_mag.

droc12345:
As a side note, while looking at scene-helper I saw the fixme comment, and that MR was done 10 months ago, Do we need to look at this?

     * FIXME: Remove the following line as soon as
     * https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4253
     * is merged. At that point wlr_scene handles damage tracking internally
     * again.
     wlr_damage_ring_rotate(&scene_output->damage_ring);

Consolatis:
> As a side note, while looking at scene-helper I saw the fixme comment, and that MR was done 10 months ago, Do we need to look at this?

Only relevant once we start to track wlroots 0.18.x. That wlroots MR didn't make it into 0.17.

tokyo4j:
I removed `last_mag`.

I also found removing `wants_magnification` doesn't make any difference in my testing.
@Consolatis Could you explain why it's necessary?

Consolatis:
> I also found removing `wants_magnification` doesn't make any difference in my testing.

For software cursors that should indeed not make a difference as we don't early out as the damage region is not empty.
For hardware cursors I am not sure if wlroots actually schedules a new frame (and thus sets `wlr_output->needs_frame`).
There is also the wayland backend which might yet again behave slightly different in regards to the cursor details.

droc12345:
Not sure if _wants_magnification_ is necessary, but it does call _is_magnify_on_, and returns true/false on that.
But that did indeed make _last_mag/is_magnify_on_ unnecessary as it's already being called.

tokyo4j:
> For software cursors that should indeed not make a difference as we don't early out as the damage region is not empty.
For hardware cursors I am not sure if wlroots actually schedules a new frame (and thus sets wlr_output->needs_frame).
There is also the wayland backend which might yet again behave slightly different in regards to the cursor details.

OK. I'm not sure removing `wants_magnification` is fine for all backends, either. In addition to wayland backend, I have no idea how cursor is handled in headless backend.

> ETA2: in output_wants_magnification, it checks for cursor not having moved, do we really care about that.
If not, then the whole function could go away, and be replaced with the simple call is_magnify_on
which simply checks the flag.

I think this is correct, given that the early return never happens when the magnifier is shown. I replaced `wants_magnification` with `is_magnify_on()` in the fixup commit.

However, I'm feeling the current situation where the screen is always redrawn when the magnifier is shown is not optimal.
Is there any way to improve this? Is this the limitation of scene-graph API?

droc12345:
cripes, what a convoluted mess. 

This is about output_wants_magnification ie. wants_magnification

It's set to return false for !is_magnify_on 
it then checks for cursor position not having moved (using seat's view) if not moved then return false
if it passes the above, then it sets a new position, 
though the last line is ... weird ... I thiink that should just be replaced with "true"

This would allow the early bail if the cursor hasn't moved
the flip side is if the cursor is not moved, then mag won't kick in, ie early return.

I'm not sure how to solve this though.

maybe the original patch was for the best and leave everything else alone (as long as everything works)
```    if (!wlr_output->needs_frame && !pixman_region32_not_empty(
            &scene_output->damage_ring.current) && !wants_magnification
-           && last_mag != is_magnify_on()) {
+           && last_mag == is_magnify_on()) {
        return false;
    }```

Consolatis:
> However, I'm feeling the current situation where the screen is always redrawn when the magnifier is shown is not optimal.

Agree.

> Is there any way to improve this? Is this the limitation of scene-graph API?

We somehow need to mark the magnifier region damaged but maybe we could do so before the early out check based on moved cursor || non-empty damage ring rather than always after modifying the buffer. That should then theoretically allow us to use the early out branch. What complicates this is that the whole function is called with different outputs so we can't just make the `additional_damage` box static as its output specific.

droc12345:
Consolatis, then we do want _wants_magnification_ for its checks on flags, movement and output.

The extra rendering with mag disabled is easy, (the original patch worked)
either change
last_mag != is_magnify_on()
to
last_mag == is_magnify_on()

OR just check !is_magnify_on()  and ignore last_mag.
either one will return early (if allowed by other conditions)

Not sure how the performance of mag on would be, but as you mention it's a lower priority
and many might never use it so it makes little difference to spend lots of time on it.

Consolatis:
Thanks for taking care of this regression.

droc12345:
Ok, added mag stuff to my config, and with the latest changes, it seems to work alright.
Power consumption is like it was before the magnifier merge, and even with mag on
the power consumption seems low. I think that the original problem with power usage
was due to the fact of using != instead of == for the check.

tokyo4j:
> Power consumption is like it was before the magnifier merge, and even with mag on the power consumption seems low.

Really? In my test, `htop` shows CPU usage goes up from 0% to 5% when I enable magnifier.

droc12345:
You're right, I did re-check, and the power usage does go up, but it's only for the time the magnification is in effect.

With the fix, without mag, the power usage is a before.
with mag on, the power does increase, though it's still low, IMO, in my case going from ~3.2 watts to ~5.5
watts, but as soon as mag is toggled off, the power usage drops to normal (all this on an idle system)
This doesn't mean that the power usage will always ~double, but it will increase, mine just happens
to be a new low power system I bought, so I was monitoring the power anyway.

So bottom line, IMO, it's good enough for general usage, if you're going to use magnification, expect it to
use a little more power. At this point, I don't think it's worth putting in a lot of work to make it better.
If there's an easy fix, to the power problem fine, otherwise I'd just live with it, at least put it on the bottom
of the stack of things to do, ie not as important as other things.

PR #1842, opened 2024-05-23
decorations: fix handling of SSD/non-SSD windows

### LINKED ISSUE #1841: SSD: `SetDecorations 'full'` does not work without `forceSSD.action="yes"`
@xi - I have found a problem with `SetDecorations`. Have not had a chance to re-visit the code. Any ideas?

Expected:
- Using the configuration below I expect to be able to switch between `full` and `border` on windows using server side decoration (SSD) with those respective keybinds.

Observed:
- I can set `border` - but cannot change back to `full`.
- Have tried with `xterm`, `foot` and `firefox`. Get the same result.
- I can change back to `full` with `SetDecorations` if I use `forceSSD.action="yes"` (but believe that should only be required for CSD windows).
- I can still cycle between `full`, `none` and `border` with `ToggleDecorations`

```xml
<keyboard>
  <keybind key="W-q" name.action="SetDecorations" decorations.action="full"/>
  <keybind key="W-v" name.action="SetDecorations" decorations.action="border"/>
</keyboard>
```

xi:
Fixup 41a3b6884630eb3299aa66c2152f1b639242999e

Fixes #1841

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -1254,7 +1254,7 @@ view_set_decorations(struct view *view, enum ssd_mode mode, bool force_ssd)
 {
 	assert(view);
 
-	if (force_ssd || !view_wants_decorations(view)
+	if (force_ssd || view_wants_decorations(view)
 			|| mode < view_get_ssd_mode(view)) {
 		view_set_ssd_mode(view, mode);
 	}


johanmalm:
Thanks. Will test tormorrow.

johanmalm:
Thanks.

Found another issues with this. Have logged at #1851

Issue #1700, opened 2024-04-10
handling SIGUSR1 and SIGUSR2


domo141:
the default action for process when receiving SIGUSR1 or SIGUSR2 is to terminate the process

`man 7 signal | grep Term`

in the future we might want to execute some action when such a signal is received.

if that would be the case, good immediate (and simple (*)) action would be to ignore those
two signals -- so that when time comes and one added actions to be executed on receiving
these signals (and announced such feature is available), recent previous versions 
-- instead of terminating -- just stayed running when these signals were received.

(*) ignore those in the same place where SIGPIPE is ignored (src/server.c) and restore where SIGPIPE
is restored (src/common/spawn.c)

But for now, considering it might be never that anyone creates anything more when SIGUSR[12] are
received, does the action of ignoring those have any downsides (e.g. xkcd 1172) ?

PR #1845, opened 2024-05-24
Ignore SIGUSR1 and SIGUSR2

domo141:
Some day there may be features bound to receiving these signals.

For now, ignore those -- this gives a period of time when sending SIGUSR1 and SIGUSR2 to a labwc process is safe, before someone may implement way to activate features bound to these signals.

Related-to: #1700

### DIFF
--- a/src/common/spawn.c
+++ b/src/common/spawn.c
@@ -23,6 +23,8 @@ reset_signals_and_limits(void)
 
 	/* Restore ignored signals */
 	signal(SIGPIPE, SIG_DFL);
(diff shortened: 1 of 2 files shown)


domo141:
I did notice the 'the' failure, but not the 'the the' in the text (until now) -- I thought it was smarter
in grammar check. Anyway, I did not like the comments -- maeby those are now better...

johanmalm:
I prefer to not touch USR1/2 unless we actually decide to use them one day.

Issue #1852, opened 2024-05-27
View matching broken for ForEach action


heroin-moose:
The following action does not work anymore:

```
   <keybind key="W-bracketleft">
      <action name="ForEach">
        <query identifier="nheko"/>
        <then>
          <action name="Focus"/>
          <action name="Raise"/>
        </then>
        <none><action name="Execute" command="nheko"/></none>
      </action>
    </keybind>
```

For some reason even if the XDG toplevel with proper app_id exists, the "none" branch is always executed.

Labwc version: `labwc 0.7.2-29-g8a5602e`.

heroin-moose:
This patch fixes things for me:

```
diff --git a/src/view.c b/src/view.c
index 8d3d091..dfe0a36 100644
--- a/src/view.c
+++ b/src/view.c
@@ -79,11 +79,12 @@ view_matches_query(struct view *view, struct view_query *query)
 		empty = false;
 		match &= title && match_glob(query->title, title);
 	}
-
+#if 0
 	if (match && query->window_type >= 0) {
 		empty = false;
 		match &= view_contains_window_type(view, query->window_type);
 	}
+#endif
 
 	return !empty && match;
 }
```

Something is off with 7bc3cae9.

Consolatis:
Yeah, it should be initialized with `-1`.

Either quick-and-dirty here:
https://github.com/labwc/labwc/blob/95e4c0d8d70838895d1285e904cc29b7a0c03d53/src/config/rcxml.c#L316
Or we create some `window_query_create()` function that does the `znew()` and initializes the window type to `-1`.

PR #1854, opened 2024-05-27
ForEach: initialize view query window_type correctly

Consolatis:
Before this patch, the window type would be checked even if not actually requested to do so.

Fixes:
- #1852


johanmalm:
LGTM

PR #1856, opened 2024-05-28
input: send tablet proximity-in only above a valid surface

jp7677:
A resize with a tablet tool might end above a non-tablet capable surface, discard the proximity-in in that case.

Fixes a segfault for this situation.


Consolatis:
Thanks!

PR #1857, opened 2024-05-29
input: prevent cursor notifications from pointer and tablet tool

jp7677:
... at the same time. Omit cursor notifications from a pointer when a tablet tool (stylus/pen) is in proximity. We expect to get cursor notifications from the tablet tool instead. A side effect from giving pointer focus on tablet proximity.

My suggestion to handle this glitch, which sound actually scarier than it is. I'm only seeing the "wrong" cursor when doing out-of-surface scrolling in GTK4 applications.

Follow up from https://github.com/labwc/labwc/pull/1678 and "draft" for now to let this sink in.

### DIFF
--- a/include/input/tablet-tool.h
+++ b/include/input/tablet-tool.h
@@ -14,6 +14,7 @@ struct drawing_tablet_tool {
 		struct wl_listener set_cursor;
 		struct wl_listener destroy;
 	} handlers;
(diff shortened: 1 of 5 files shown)


johanmalm:
Seems okay to me.

Consolatis:
Thanks :)

PR #1858, opened 2024-05-29
session-lock: fix flashing & update cursor shape & refactoring

### LINKED ISSUE #1453: Locked screen issues (black flash, cursor showing not hidden/disabled) both with swaylock and waylock
I've been comparing hyprland and river against labwc for the last couple of months and I've noticed some differences when locking a session. I think hyprland and river are both doing it a bit better than labwc in this regard and I hope that labwc can improve upon this.

**Issue number one:**

When locking a session, both hyprland and river (with either waylock or swaylock) will switch to the lock screen immediately without flashing a black frame before setting whatever color you've chosen on said lock screen. For example, if your background is set to pure white and your lock screen is set to a blue color it will go directly from white to blue. In labwc, it will go from white to a black screen and then to blue (a quick black frame flash). I know It's a very small nuissance but it should not be there IMO.

**Issue number two:**

Also, both hyprland and river will disable and hide the cursor as soon as the session is locked but labwc keeps the cursor enabled and visible. I read a little bit about the wayland specification for the session locking and I think it was mostly left up to the compositor to decide to hide or not hide the cursor. I think the cursor is unnecessary and can sometimes be misleading since you won't notice that you've locked your session if you happen to have the same background color as the locked screen color (just as I have it set up -_-).

Sorry if this is too picky but I want labwc to be the best it can be and it is now my daily driver and it actually allowed me to get rid of awesome wm at last :)

tokyo4j:
Fixes #1453 by adding a timer to delay filling the screen with opaque color and by adding calls to `cursor_update_focus()`, but also includes many refactoring like replacing `screen_lock` with `screen_lock_manager`. Note that I don't have strong opinions about function/variable names.

I did following tests with swaylock:
- Plug / unplug an output while the session is locked
- Replace session lock with: `swaylock & sleep 1;pkill swaylock;sleep 1;swaylock`
- Patch swaylock to remove some calls to `wl_surface_commit()` and confirm the screen is filled with opaque color


src/session-lock.c
```diff
@@ -152,12 +172,21 @@ static void
...
+	wlr_scene_node_set_enabled(&lock_output->background->node, true);
+	return 0;
+}
+
 void
-session_lock_output_create(struct session_lock *lock, struct output *output)
+session_lock_output_create(struct session_lock_manager *manager, struct output *output)
 {
 	struct session_lock_output *lock_output = znew(*lock_output);
 	if (!lock_output) {
```
tokyo4j, line 192: Is this check necessary? I don't think `lock_output` can be NULL here.

johanmalm: Agree. `znew()` "handles" it.

src/session-lock.c
```diff
@@ -211,130 +210,125 @@ session_lock_output_create(struct session_lock *lock, struct output *output)
...
+		wl_list_remove(&manager->lock_unlock.link);
+		wl_list_remove(&manager->lock_new_surface.link);
 	}
-	if (!lock->abandoned) {
-		wl_list_remove(&lock->destroy.link);
-		wl_list_remove(&lock->unlock.link);
-		wl_list_remove(&lock->new_surface.link);
-	}
-	free(lock);
+	manager->lock = NULL;
```
johanmalm, line 261: Should we not `free(manager->lock)`?

johanmalm:
LGTM other than `free()` query above (which I might be wrong on).

The only unexpected thing I observed when testing was that with `swaylock` I don't get the cursor back when moving it. IIRC pre-session-lock, the cursor disappeared on swaylock startup, but if the cursor was later moved it re-appeared.

johanmalm:
...and thanks a lot for working on this one. The refactoring is really nice. A spare pair of eyes helps a lot.

IIRC @Consolatis wasn't convinced by hiding the cursor, but I personally really like it. Feels cleaner.

src/session-lock.c
tokyo4j: `manager->lock` is just `wlr_session_lock_v1` so wlroots handles it.

Sway wraps `wlr_session_lock_v1` with `sway_session_lock` and separates it from session lock manager like:
```c
struct sway_session_lock {
	struct wlr_session_lock_v1 *lock;
	struct wlr_surface *focused;
	bool abandoned;

	struct wl_list outputs; // struct sway_session_lock_output

	// invalid if the session is abandoned
	struct wl_listener new_surface;
	struct wl_listener unlock;
	struct wl_listener destroy;
};
```
```c
struct sway_server {
	...
	struct {
		struct sway_session_lock *lock;
		struct wlr_session_lock_manager_v1 *manager;

		struct wl_listener new_lock;
		struct wl_listener manager_destroy;
	} session_lock;
	...
}
```

So sway frees `sway_session_lock` every time a session-lock is destroyed, but I prefer just creating `session_lock_manager` which is persistent throughout the session because there cannot be more than 1 session-lock:
```c
struct session_lock_manager {
	struct server *server;
	struct wlr_session_lock_manager_v1 *wlr_manager;
	struct wlr_session_lock_v1 *lock;
	struct wlr_surface *focused;
	bool locked;

	struct wl_list session_lock_outputs;

	struct wl_listener new_lock;
	struct wl_listener destroy;

	struct wl_listener lock_new_surface;
	struct wl_listener lock_unlock;
	struct wl_listener lock_destroy;
};
```

tokyo4j:
> IIRC @Consolatis wasn't convinced by hiding the cursor, but I personally really like it. Feels cleaner.

Oh I didn't know that. Personally I prefer hiding it too and think it's fine as long as it's requested by the session-lock client.

src/session-lock.c
johanmalm: Great table. I think we should add the second table as a comment in the source file. It's likely to be a useful reminder in years to come :smile:

johanmalm:
LGTM
Feel free to go for merge.

src/session-lock.c
tokyo4j: Added:
```c
	/*
	 * When not locked: lock=NULL, locked=false
	 * When locked: lock=non-NULL, locked=true
	 * When lock is destroyed without being unlocked: lock=NULL, locked=true
	 */
```

johanmalm:
Thanks

Consolatis:
> > IIRC @Consolatis wasn't convinced by hiding the cursor, but I personally really like it. Feels cleaner.
> 
> Oh I didn't know that. Personally I prefer hiding it too and think it's fine as long as it's requested by the session-lock client.

To clarify, my issue is unconditionally hiding the cursor as there might be screenlockers (already or in the future) that want to allow interactions with the cursor. As long as screenlockers can still set their own cursor shape if they want to everything in fine from my POV.

tokyo4j:
While investigating #1943, I found I forgot to set `manager->lock` so it's always `NULL`, which conflicts with my comments in `session-lock.h`. I'll fix it soon. I don't know why I made such a stupid mistake.

Consolatis:
No worries, we've all been there :)

PR #1863, opened 2024-05-31
Implement <resize><drawContents>

### LINKED ISSUE #1328: Feature request: <resize><drawContents>no</drawContents> ... </resize>
With compositing on resizing a Chromium window on an aging intel IGPU is often laggy experience, with a high polling rate gaming mouse on Wayland it's horrible. Only on Hyprland it's not, maybe because of some wlroots ENVs I have on there, I don't yet know, but in KDE, Gnome, labwc it is. (Maybe the trick is that the border I'm dragging gets more CPU time on Hyprland and also maybe they are doing a trick with window content warping for resize, which is a fancy way of doing the same thing I'm asking for here in this post) 
Openbox has a feature that helps sidestepping this problem, if 
```
 <resize>
    <drawContents>no</drawContents>
...
  </resize>
```
is added to the rc.xml the system doesn't have to struggle redrawing everything, battery is saved on a laptop etc etc..
Please consider adding this, if you feel like it someday ^

Thank you (and contributors as well) for developing labwc, it's already so much better than openbox in some fundamental ways. (window aero snap without janky half-assed workarounds and the horrible titlebar seperator that ruins consistent theming is not in labwc at all!)

tokyo4j:
Closes #1328.

https://github.com/labwc/labwc/assets/22029524/e59a08f6-e3d1-4806-afb1-d95e3d4e8c74

The theme for the frame shown to indicate resized window when `<resize><drawcontents>` is set to no is configured with:
- `resize-frame.width`
- `resize-frame.color`

But I'm not sure about these names. Perhaps `resize.frame.{width,color}` or other names are better.

I initially implemented the frame as a combination of multi_rect and scene_rect like I did for snapping overlay, but I removed it for simplicity (we need to add `resize-frame.{bg,border}.enabled` and disable translucent scene_rect for pixman renderer).
If anyone wants it, I can move it back from [here](https://github.com/tokyo4j/labwc/tree/drawcontents-first).

The first commit adds `zero_array()` macro in `theme.c` to shorten the lengthy `memcpy()` calls like `memset(theme->snapping_overlay_region.border_color, 0, sizeof(theme->snapping_overlay_region.border_color))`.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -456,6 +456,11 @@ extending outward from the snapped edge.
 
 	Default is Never.
 
+*<resize><drawContents>* [yes|no]
+	Let the application redraw its contents while resizing. If disabled, an
+	outlined rectangle is shown to indicate the geometry of resized window.
+	Default is yes.
+
 ## KEYBOARD
 
 *<keyboard><numlock>* [on|off]
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -103,8 +103,12 @@
     <windowEdgeStrength>20</windowEdgeStrength>
   </resistance>
 
-  <!-- Show a simple resize and move indicator -->
-  <resize popupShow="Never" />
+  <resize>
+    <!-- Show a simple resize and move indicator -->
+    <popupShow>Never</popupShow>
+    <!-- Let client redraw its contents while resizing -->
+    <drawContents>yes</drawContents>
+  </resize>
 
   <focus>
     <followMouse>no</followMouse>
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -118,6 +118,7 @@ struct rcxml {
 	enum tiling_events_mode snap_tiling_events_mode;
 
 	enum resize_indicator_mode resize_indicator;
+	bool resize_draw_contents;
 
 	struct {
 		int popuptime;
--- a/include/resize-outlines.h
+++ b/include/resize-outlines.h
@@ -1,0 +1,13 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_RESIZE_OUTLINES_H
+#define LABWC_RESIZE_OUTLINES_H
+
+#include <wlr/util/box.h>
+
+struct view;
+
+void resize_outlines_update(struct view *view, struct wlr_box new_geo);
+void resize_outlines_finish(struct view *view);
+bool resize_outlines_enabled(struct view *view);
+
+#endif /* LABWC_RESIZE_OUTLINES_H */
--- a/include/view.h
+++ b/include/view.h
@@ -238,6 +238,10 @@ struct view {
 		struct wlr_scene_rect *background;
 		struct scaled_font_buffer *text;
 	} resize_indicator;
+	struct resize_outlines {
+		struct wlr_box view_geo;
+		struct multi_rect *rect;
+	} resize_outlines;
 
 	struct foreign_toplevel {
 		struct wlr_foreign_toplevel_handle_v1 *handle;
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1030,6 +1030,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		} else {
 			wlr_log(WLR_ERROR, "Invalid value for <resize popupShow />");
 		}
+	} else if (!strcasecmp(nodename, "drawContents.resize")) {
+		set_bool(content, &rc.resize_draw_contents);
 	} else if (!strcasecmp(nodename, "mouseEmulation.tablet")) {
 		set_bool(content, &rc.tablet.force_mouse_emulation);
 	} else if (!strcasecmp(nodename, "mapToOutput.tablet")) {
@@ -1269,6 +1271,7 @@ rcxml_init(void)
 		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
 
 	rc.resize_indicator = LAB_RESIZE_INDICATOR_NEVER;
+	rc.resize_draw_contents = true;
 
 	rc.workspace_config.popuptime = INT_MIN;
 	rc.workspace_config.min_nr_workspaces = 1;
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -23,6 +23,7 @@
 #include "menu/menu.h"
 #include "regions.h"
 #include "resistance.h"
+#include "resize-outlines.h"
 #include "ssd.h"
 #include "view.h"
 
@@ -307,7 +308,11 @@ process_cursor_resize(struct server *server, uint32_t time)
 			server->grab_box.width - new_view_geo.width;
 	}
 
-	view_move_resize(view, new_view_geo);
+	if (rc.resize_draw_contents) {
+		view_move_resize(view, new_view_geo);
+	} else {
+		resize_outlines_update(view, new_view_geo);
+	}
 }
 
 void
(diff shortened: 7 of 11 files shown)


include/resize-frame.h
```diff
@@ -0,0 +1,11 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_RESIZE_FRAME_H
+#define LABWC_RESIZE_FRAME_H
+
+#include "view.h"
+#include <wlr/util/box.h>
+
+void resize_frame_update(struct view *view, struct wlr_box new_geo);
+void resize_frame_finish(struct view *view);
```
Consolatis, line 9: `struct view` could be forward declared rather than including `view.h`

Consolatis:
Diff LGTM.

> But I'm not sure about these names. Perhaps resize.frame.{width,color} or other names are better.

Agree, personally I also prefer `resize.frame.{width,color}`.

If openbox provides a setting for the feature (and the theme vars) we should use whatever openbox uses though, I didn't look into it yet.

johanmalm:
I prefer thickness to width because width has multiple meanings (i.e. also width of window)

tokyo4j:
> Agree, personally I also prefer resize.frame.{width,color}.

I don't have a strong preference here. I used `resize-frame` because I thought `frame` might be a bit vague so I defined `resize-frame` as a proper noun. But it won't make any big difference.

I'll change it to `resize.frame`. 

> If openbox provides a setting for the feature (and the theme vars) we should use whatever openbox uses though, I didn't look into it yet.

Openbox doesn't help here because it implements `<resize><drawContents>` differently and there's no theme entries for it. If we can implement it in the same way that doesn't hurt our codebase, switching to it would be another option.

https://github.com/labwc/labwc/assets/22029524/43f81fad-e88d-4373-9474-d4363f66879d

> Somewhat related: the preview outlines (osd_update_preview_outlines) and this feels like kind of the same thing. Maybe the preview outlines should be configurable and let them both be set from the same options?

I added `osd.window-switcher.preview.border.{width,color}` in #1715. But if we follow its names here, we should use `resize.frame.border.{width,color}`.

> I prefer thickness to width because width has multiple meanings (i.e. also width of window)

Agree. I think `resize.frame.thickness` or `resize.frame.border.width` would be better for clarity. 

So, my current candidates are:
1. `resize.frame.border.{width,color}`
2. `resize.frame.{thickness,color}`

Any opinions?

include/resize-frame.h
tokyo4j: Thanks. Fixed.

Consolatis:
> So, my current candidates are:
> 
> 1. `resize.frame.border.{width,color}`
> 2. `resize.frame.{thickness,color}`
> 
> Any opinions?

With 1 we could add the openbox behavior in the future if we want to (e.g. drawing a solid rect behind the window contents on resize) and then use something like `resize.frame.color` to define its color. On the other side we could also use `resize.frame.bg.color` there. That would then be a bit weird to have `resize.frame.color` be the border color though.

So I'd prefer `1` of the two prefixes. Regarding `width` vs `thickness` I don't really have an opinion.

tokyo4j:
> With 1 we could add the openbox behavior in the future if we want to (e.g. drawing a solid rect behind the window contents on resize) and then use something like `resize.frame.color` to define its color. On the other side we could also use `resize.frame.bg.color` there. That would then be a bit weird to have `resize.frame.color` be the border color though.

> So I'd prefer 1 of the two prefixes. Regarding width vs thickness I don't really have an opinion.

If we consider the case we implement Openbox's behavior in the future, I don't think `resize.frame.bg` is a right name there. I named `resize.frame` with the look of multi-rect in mind, and "background of frame" doesn't make sense to me (it may be just my English issue though).

When we implement the Openbox's behavior, I think a possible configuration is like this:

```xml
<!-- no is equal to frame(?) -->
<resize><drawContents>yes|no|frame|fill</drawContents></resize>
```

```
resize.frame.thickness: 1
resize.frame.color: #dddda6,#000000,#dddda6
resize.fill.color: #dddda6
```

So I now prefer `resize.frame.{thickness,color}`.

tokyo4j:
The last thing I'm wondering is `frame` vs `outlines`. Perhaps `outlines` is better if we follow `<windowSwitcher outlines="yes">`.

BTW I'm now regretting I named `window-switcher.preview.border.{width,color}` as I didn't know the outlines can be drawn even with `<windowSwitcher preview="no" />`. I think we can change it to `window-switcher.outlines.{thickness,color}`.

tokyo4j:
I squashed the fixup commit and updated the commit message.
This PR is ready now, but I'll wait for LTGM from @Consolatis just in case. There's no need to hurry.

tokyo4j:
Regarding to https://github.com/labwc/labwc/issues/1328#issuecomment-2142565270, I think applying the same behavior while moving window with a configuration like `<move><drawContents>` is also worth implementing to reduce the size of the updated area.

In fvwm, they're configured with:
```
# Draw a frame on resize and move
Style * ResizeOutline
OpaqueMoveSize 0

# Let clients draw their content on resize and move
Style * ResizeOpaque
OpaqueMoveSize unlimited 
```

johanmalm:
LGTM - except we ought to add the config option in `docs/rc.xml.all` too.

Have tested and it works well.

Nice contribution.

Consolatis:
Had a quick test again and noticed two things:
- If the resize popup is enabled it feels a bit weird that it doesn't update on movement (neither in value nor in position).
- Sometimes the border is a bit hard to see, especially on white application backgrounds.

Do we want to fix both of these within this PR?

tokyo4j:
>  - If the resize popup is enabled it feels a bit weird that it doesn't update on movement (neither in value nor in position).

I fixed this in the fixup commit.

> - Sometimes the border is a bit hard to see, especially on white application backgrounds.
 
> Do we want to fix both of these within this PR? The 2nd issue could be fixed with using an (additional?) wlr_scene_rect for the background.

Hmm... For me the outline with white/black/white colors is recognizable enough even on white backgrounds. You can increase `resize.outlines.thickness` to make it more recognizable.

Of course we can add `wlr_scene_rect`, but then we need to repeat what we did when implementing snapping overlay: adding `resize.overlay.bg.{enabled,color}` and `resize.overlay.border.{enabled,color,width}`, and disabling `wlr_scene_rect` for pixman renderer. Also, it's not what remote desktop users would want. So I'm a bit less motivated for it.

johanmalm:
LGTM.
Won't merge as there is a '[fixup]' commit still to be rebased at the end.

tokyo4j:
I removed theme entries as I'm not 100% sure that the current approach with outlines is right.
Let's discuss about it later when users request settings for it or when we find a better approach.

This PR is good to go now. Sorry for my indecisiveness.

johanmalm:
Sounds sensible. I've often gone for as few config/theme options as possible to start with. Can always add later as you say.

Thanks for PR. It's been on the cards for a while. Nice to get this in.

Consolatis:
> Of course we can add `wlr_scene_rect`, but then we need to repeat what we did when implementing snapping overlay: adding `resize.overlay.bg.{enabled,color}` and `resize.overlay.border.{enabled,color,width}`, and disabling `wlr_scene_rect` for pixman renderer. Also, it's not what remote desktop users would want. So I'm a bit less motivated for it.

wlr_scene_rects should be no issue for pixman, its only if they are combined with transparency which causes visible performance degradation, at least from my experience. Fine with delaying this though. Thanks for the PR :)

tokyo4j:
> wlr_scene_rects should be no issue for pixman, its only if they are combined with transparency which causes visible performance degradation, at least from my experience. Fine with delaying this though. Thanks for the PR :)

I thought showing resizing overlay with opaque color which hides the window contents is unacceptable, but I tried it and found it's not that bad (it flickers when finishing resizing though):

https://github.com/labwc/labwc/assets/22029524/69176d35-448d-424d-b783-237fcd508650

Perhaps this also works nicely with compression algorithm for remote desktop user.

Let's discuss it later anyway.

Consolatis:
Technically, we could likely disable `view->scene_node` and delay disabling the overlay until the client commits a new size where we also enable `view->scene_node` again. That should get rid of the flicker and make the overlay a replacement for the surface itself.

PR #1866, opened 2024-06-01
Add CycleNearRegion action

### LINKED ISSUE #1001: Feature Request: NextWindowInRegion/PreviousWindowInRegion
Regions &the  SnapToRegion action are nice features, and would be well complimented by NextWindowInRegion & PreviousWindowInRegion actions, which would cycle through all the windows that are (mostly?) within that region.

I currently have this workflow with OpenBox (& a custom CLI), using W-KP_* to focus and W-S-KP_* to move+resize windows, and it's quite a nice way to do manual tiling.

rmt:
CycleInRegion (edit: now CycleNearRegion) will cycle the focus between all views that are "mostly" within the given region.

The implementation defines mostly as at least 80% of the view being within the region's geometry, or 80% of the region's geometry being in the view's geometry.

See https://github.com/labwc/labwc/issues/1001#issuecomment-2125118483 for more information and an example config that uses the numeric keypad to define, move and focus windows to corner/side/center regions.

Closes #1001

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -394,6 +394,10 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			}
 			goto cleanup;
 		}
+		if (!strcasecmp(argument, "topmostFirst")) {
+			action_arg_add_bool(action, argument, parse_bool(content, false));
+			goto cleanup;
+		}
 		if (!strcasecmp(argument, "region")) {
 			action_arg_add_str(action, argument, content);
 			goto cleanup;
@@ -1195,7 +1199,8 @@ run_action(struct view *view, struct action *action,
 		if (server.input_mode == LAB_INPUT_STATE_CYCLE) {
 			cycle_step(dir);
 		} else {
-			cycle_begin(dir, filter);
+			bool topmost_first = action_get_bool(action, "topmostFirst", false);
+			cycle_begin(dir, filter, topmost_first);
 		}
 		break;
 	}
@@ -1204,7 +1209,8 @@ run_action(struct view *view, struct action *action,
 		enum lab_cycle_dir dir = (action->type == ACTION_TYPE_NEXT_WINDOW_IMMEDIATE) ?
 			LAB_CYCLE_DIR_FORWARD : LAB_CYCLE_DIR_BACKWARD;
 		struct cycle_filter filter = cycle_filter_from_action(action);
-		cycle_immediate(dir, filter);
+		bool topmost_first = action_get_bool(action, "topmostFirst", false);
+		cycle_immediate(dir, filter, topmost_first);
 		break;
 	}
 	case ACTION_TYPE_RECONFIGURE:
--- a/src/action.c
+++ b/src/action.c
@@ -2,6 +2,7 @@
 #define _POSIX_C_SOURCE 200809L
 #include "action.h"
 #include <assert.h>
+#include <errno.h>
 #include <signal.h>
 #include <string.h>
 #include <strings.h>
@@ -288,6 +289,23 @@ action_get_actionlist(struct action *action, const char *key)
 	return arg ? &arg->value : NULL;
 }
 
+static bool
+parse_percentage(const char *content, int *percentage)
+{
+	assert(content);
+	assert(percentage);
+
+	errno = 0;
+	char *end;
+	long value = strtol(content, &end, 10);
+	if (errno == ERANGE || end == content || *end != '\0'
+			|| value < 0 || value > 100) {
+		return false;
+	}
+	*percentage = value;
+	return true;
+}
+
 void
 action_arg_from_xml_node(struct action *action, const char *nodename, const char *content)
 {
@@ -372,6 +390,22 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 				action_arg_add_int(action, argument, CYCLE_APP_ID_CURRENT);
 			} else {
 				wlr_log(WLR_ERROR, "Invalid argument for action %s: '%s' (%s)",
+					action_names[action->type], argument, content);
+			}
+			goto cleanup;
+		}
+		if (!strcasecmp(argument, "region")) {
+			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
+		if (!strcasecmp(argument, "windowOverlapPercent")
+				|| !strcasecmp(argument, "regionOverlapPercent")) {
+			int percentage;
+			if (parse_percentage(content, &percentage)) {
+				action_arg_add_int(action, argument, percentage);
+			} else {
+				wlr_log(WLR_ERROR, "Invalid argument for action %s: "
+					"'%s' (%s); expected an integer from 0 to 100",
 					action_names[action->type], argument, content);
 			}
 			goto cleanup;
@@ -622,7 +656,7 @@ action_branches_are_valid(struct action *action)
 	return true;
 }
 
-/* Checks for *required* arguments */
+/* Check required arguments and values that need post-parse validation */
 bool
 action_is_valid(struct action *action)
 {
@@ -641,6 +675,12 @@ action_is_valid(struct action *action)
 		arg_name = "direction";
 		arg_type = LAB_ACTION_ARG_INT;
 		break;
+	case ACTION_TYPE_NEXT_WINDOW:
+	case ACTION_TYPE_PREVIOUS_WINDOW:
+	case ACTION_TYPE_NEXT_WINDOW_IMMEDIATE:
+	case ACTION_TYPE_PREVIOUS_WINDOW_IMMEDIATE:
+		return cycle_regions_are_valid(
+			action_get_str(action, "region", NULL));
 	case ACTION_TYPE_SHOW_MENU:
 		arg_name = "menu";
 		break;
@@ -1054,6 +1094,22 @@ warp_cursor(struct view *view, const char *to, const char *x, const char *y)
 	cursor_update_focus();
 }
 
+static struct cycle_filter
+cycle_filter_from_action(struct action *action)
+{
+	return (struct cycle_filter) {
+		.workspace = action_get_int(action, "workspace",
+			rc.window_switcher.workspace_filter),
+		.output = action_get_int(action, "output", CYCLE_OUTPUT_ALL),
+		.app_id = action_get_int(action, "identifier", CYCLE_APP_ID_ALL),
+		.region_names = action_get_str(action, "region", NULL),
+		.window_overlap_percent = CLAMP(
+			action_get_int(action, "windowOverlapPercent", 0), 0, 100),
+		.region_overlap_percent = CLAMP(
+			action_get_int(action, "regionOverlapPercent", 0), 0, 100),
+	};
+}
+
 static void
 run_action(struct view *view, struct action *action,
 	struct cursor_context *ctx)
@@ -1135,14 +1191,7 @@ run_action(struct view *view, struct action *action,
 	case ACTION_TYPE_PREVIOUS_WINDOW: {
 		enum lab_cycle_dir dir = (action->type == ACTION_TYPE_NEXT_WINDOW) ?
 			LAB_CYCLE_DIR_FORWARD : LAB_CYCLE_DIR_BACKWARD;
-		struct cycle_filter filter = {
-			.workspace = action_get_int(action, "workspace",
-				rc.window_switcher.workspace_filter),
-			.output = action_get_int(action, "output",
-				CYCLE_OUTPUT_ALL),
-			.app_id = action_get_int(action, "identifier",
-				CYCLE_APP_ID_ALL),
-		};
+		struct cycle_filter filter = cycle_filter_from_action(action);
 		if (server.input_mode == LAB_INPUT_STATE_CYCLE) {
 			cycle_step(dir);
 		} else {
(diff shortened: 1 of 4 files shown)


src/action.c
```diff
@@ -1026,6 +1037,29 @@ actions_run(struct view *activator, struct server *server,
...
 			break;
+		case ACTION_TYPE_CYCLE_IN_REGION:
+			{
+				if (!view) {
+					break;
+				}
+				struct output *output = view->output;
+				if (!output) {
+					break;
+				}
```
Consolatis, line 1048: Should likely be `output_is_usable(view->output)` rather than `!output` to catch leased / disabled outputs.

include/labwc.h
```diff
@@ -454,6 +454,20 @@ enum lab_cycle_dir {
 struct view *desktop_cycle_view(struct server *server, struct view *start_view,
 	enum lab_cycle_dir dir);
 
+/**
+ * is_mostly_overlapping - return true if box A or B overlap each other
+ * by more than 80% of their own area
+ */
+int is_mostly_overlapping(struct wlr_box *a, struct wlr_box *b);
```
Consolatis, line 461: This could be a static function in `desktop.c` as it appears to only be called within that file.

src/desktop.c
```diff
@@ -139,6 +139,57 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
 }
 
+/*
+ *  is either box overlapping the other by 80% of its own size?
+ */
+int
+is_mostly_overlapping(struct wlr_box *a, struct wlr_box *b)
+{
+	inline int min(int a, int b) { return (a < b) ? a : b; }
+	inline int max(int a, int b) { return (a > b) ? a : b; }
```
Consolatis, line 149: We do have `MAX()` and `MIN()` macros in `common/macros.h`

Consolatis (commented):
Thanks for your work on this.

I've added some review comments but I am not sure if this feature is in scope for labwc in the first place. It seems like pretty low maintenance overhead though. Thoughts @johanmalm (and others)?

src/desktop.c
```diff
@@ -139,6 +139,57 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
+
+	struct view *(*iter)(struct wl_list *head, struct view *view,
+		enum lab_view_criteria criteria);
+	bool forwards = true;
+
+	// if the currently active/focused window is within the region,
+	// then we want to cycle through in reverse order so that we
+	// cycle through all windows, and not just the two most recent.
+	if (active_view && is_mostly_overlapping(&active_view->current, &region->geo)) {
+		forwards = false;
```
johanmalm, line 176: Rather than changing direction couldn't we just `cur = iter(&server->views, NULL, criteria);`?

(probably after initializing `iter` and `criteria`.

johanmalm:
This looks nicely implemented.

High level feedback:

1. In case we ever want to cycle windows that have actually been snapped to a region (rather than just being near it) we ought to call things `near` rather than `in`, for example `CycleNearRegion`.
2. It feels like a slight "misuse" of `regions`, but I can live with that :smile: I.e. it's being used to define an area for reasons other snapping windows to it.
3. I think the use-case is limited, but agree with @Consolatis that it's low overhead so I'm happy on that basis.
4. We use `/*` style comments rather than `//`
5. Prefer no capitalization in variable names (e.g. `area_a` rather than `areaA`)

src/desktop.c
rmt: I tried a few different ways here, and this one felt the best to me (and was simpler than my first attempt), with the window ordering for same-region focus cycling now feeling correct.  Give it another shot in a second.

src/action.c
rmt: Thanks, fixed.  Your comment is probably relevant to the function I copied it from too, but that'd be for another patch.

include/labwc.h
rmt: Thanks!

johanmalm:
Please can we dwell a pause on this one. I think we should explore of this "requirement" can be solved in a more generic way.

It feels like `desktop_cycle_view()` and `osd_update()` are often (not checked thoroughly) called together. If we were to pass the `criteria` to those functions, maybe we would just add `minimized` and `near_regions` criteria and re-use current infrastructure.

Welcome thoughts / investigation.

droc12345:
Possible to enhance the window switcher code to handle regions, add a region criteria to enum,
but that's just off the top of my head.

rmt:
I could imagine view_next / view_next_no_head_stop / view_prev_no_head_stop having an optional region and mostly/fuzzy/near bool argument, along with desktop_cycle_view.  Going this route would give us the osd display (which I personally find a little distracting, but I digress).  I'm not quite sure how this would interact with the other window switching actions (eg. cycle view with a region, then cycle view with tab/shift-tab).

Anyway - I didn't want to start touching code that I didn't fully understand (especially in my first patch!)  Let me know if the team would prefer a different approach, and I'll see what I can do.

droc12345:
Duplicating code and making minor mods rather than enhance the original code usually winds up
biting one in the long run. My preference is to enhance desktop_cycle_view and osd_update even at the
cost of some extra code. 

The osd window doesn't have to display, that's a setting, you can turn off display and window switching 
still works (though a bit like a blind man looking for something), plus the elements of osd can be turned off,
ie preview and preview-outline, while leaving the main window.

rmt:
Just the current workspace's regions.  If you're navigating focus by keyboard, then you'd want to explicitly jump between workspaces, monitors, etc, and then hone in on the specific window that you want.

I'm now looking at adding a lab_view_extra_criteria struct to view_next/view_next_no_head_stop/previous_no_head_stop with a region variable (maybe fuzzy_region too, allowing for both exact & fuzzy/near matching), and then passing that through to matches_criteria() in addition to the lab_view_criteria enum.  Those two could then be passed explicitly to desktop_cycle_view from the actions.  I'd also add another enum value to ignore minimized windows.

I'm not sure if we would want to change the NextWindow/PreviousWindow actions or go with a separate action yet.  I haven't dived into the OSD code yet.

droc12345:
I'm changing my mind on what needs doing.
I've looked at what it takes to modify the different pieces of the window switcher and it would be involved,
more than what's needed for your simple case, so ....

Your original code, is probably fine, but rather than put the modified desktop_cycle* in desktop.c
I think it fits better being added to the regions.c code as it really only pertains to that, and it keeps
it out of the way of the mainstream code. 

Johan, Consolatis, ahesford comments?

src/desktop.c
```diff
@@ -139,6 +140,56 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
+	}
+	iter = forwards ? view_next_no_head_stop : view_prev_no_head_stop;
+
+	enum lab_view_criteria criteria = rc.window_switcher.criteria;
+
+	struct view *cur;
+	cur = iter(&server->views, active_view, criteria);
+	while (cur && cur != active_view) {
+		if (!cur->minimized
+			&& is_mostly_overlapping(&cur->current, &region->geo)) {
```
Consolatis, line 185: CodeStyle nitpick: double intend for condition continuation lines.
E.g. something like
```c
		if (!cur->minimized && is_mostly_overlapping(
				&cur->current, &region->geo)) {
			return cur;
		}
```
Or just keep everything in a single line as long as its < 100 cols.

Consolatis:
> Your original code, is probably fine, but rather than put the modified desktop_cycle* in desktop.c I think it fits better being added to the regions.c code as it really only pertains to that, and it keeps it out of the way of the mainstream code.

`desktop.c` is fine, its intended for functions that depend on multiple views. For `is_mostly_overlapping()` in specific that could indeed be moved to `regions.c` but it feels kind of pointless to make that globally available. I don't have a strong preference there.

> Edit to add: your desktop_cycle* shouldn't be struct view *, as you return null at the end, change to void and just do the return. Remove the assert on region, and use, if null then return, instead.

I don't understand what you mean by that. The function returns the next view or `NULL`.

droc12345:
Looks good to me, I briefly tested, didn't try every combination though. 
Didn't crash the system, and the windows cycled when in or near the region.

droc12345:
> > Edit to add: your desktop_cycle* shouldn't be struct view *, as you return null at the end, change to void and just do the return. Remove the assert on region, and use, if null then return, instead.
> 
> I don't understand what you mean by that. The function returns the next view or `NULL`.

I don't know what I was thinking when I said that.

johanmalm:
Through IRC discussion, @Consolatis made a good point which is that it's not easy to create a `near_region()` criteria without changing the current infrastructure a lot, so let's not go there.

The only thing that doesn't sit right with me is the reversal of direction. Had a quick tinker, but can't get it working just now.

droc12345:
In my original thoughts, when looking at desktop_cycle_view_near_region()
was that the whole top part could be replaced by the standard desktop_cycle_view which just feeds a list
to something like osd_update, and if so then the standard forward/reverse should work.
The bottom part of desktop_cycle_view_near_region() would need to be modified along with fixing
the action trigger to mimic NEXT_WINDOW/PREV_WINDOW actions.

Either way works, I don't think that that many people will use this, so not worth putting in an lot of work for little to no gain.

johanmalm:
Sorry this has taken a while to think through. 

I feel there are three options:

1. Basic approach (as this PR is implemented). Just jump to previous window without waiting for alt-release. This has to go backwards because otherwise we just keep switching between the two top-most ones (like `tinywl`). I think the UX is very different to that of the window-switcher and it's quite a big step in implementation to do Alt + multiple tabs to keep going through list, so if we ever wanted to improve it we would most likely have to swap to option 2 anyway and then cause a regression.
2. Re-use the window-switcher infra (passing a context struct to the iterating function). This is yet to be proven by writing the code, but seems the cleanest / most maintainable / extensible approach. Feels like this type of attribute would work: `<action name="NextWindow" region="foobar" />`
3. Duplicate a lot of the window-switcher related infra (like `server->cycle_view` + lots of stuff in `src/input/keyboard.c` to handle alt-press/release + handle views that get destroyed whilst cycling, etc.). This feels like a bad option.

So my preference is to can this PR or re-write as described in point 2.

src/desktop.c
```diff
@@ -139,6 +140,58 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
+/*
+ *  is either box overlapping the other by 80% of its own size?
+ */
+static int
+is_mostly_overlapping(struct wlr_box *a, struct wlr_box *b)
+{
+	int x_overlap = MAX(0, MIN(a->x + a->width, b->x + b->width)
+				- MAX(a->x, b->x));
+	int y_overlap = MAX(0, MIN(a->y + a->height, b->y + b->height)
+				- MAX(a->y, b->y));
```
tokyo4j, line 152: I suggest using `wlr_box_intersection()` here.

droc12345:
Johan, to implement option 2 cleanly, I think I would have to modify the desktop_cycle_view() by adding a region as a 4th parameter. This involves modifying all current calls to desktop_cycle_view() by adding NULL as a 4th parm, then inside desktop_cycle_view make a decision based on whether region is null or not.
This doesn't take into account using the osd part of window_switcher, I would probably follow what was
originally written and just change to the next window that's in the region. 
To use the osd for display, would require more changes, but I haven't thought all of them through.

I've tried a couple of different approaches, and none did what I wanted, so I'll try the above approach later
to see if it will work.

rmt:
If we want extra criteria for NextWindow/PreviousWindow (osd), then I suspect that the most consistent approach would be to store the extra criteria in server->osd_state and have osd_update() and desktop_cycle_view pass them into the view_* functions from there.  The extra criteria should be cleared when the OSD disappears.

Alternatively, we don't touch that at all, and have desktop_cycle_view() take extra context and do the extra filtering itself - this will result in the full list of windows appearing in the OSD, but only a subset of those being selectable.  There is still the issue of the second window being focused in a region instead of the top-most unfocused, though (see https://github.com/labwc/labwc/pull/1479#pullrequestreview-1849829864 .. although I don't quite understand the need for the change there).

Regardless, after having played around a bit and having made it work in a couple of ways (and still not being happy with my code), I realise that the OSD approach just doesn't scratch my itch for super-fast-focus-switching. :cry: 

I still think there is room for CycleNearRegion or perhaps a more generic non-OSD CycleWindowFocus variant.  Perhaps along the lines of `<CycleWindowFocus region="x" desktop="x" output="x">`.  CycleWindowFocus would immediately switch output & workspace if needed and focus the top window in the region or just to the top-most window if region isn't provided. If no output/workspace change is needed, then it would maintain the current behaviour of CycleNearRegion (top-most if unselected, otherwise reverse order).  `<CycleWindowFocus>` by itself would immediately focus the LRU window on the current output+workspace, which I suspect could be an accessibility feature in and of itself.

rmt:
After sitting on this for far too long (and happily using myself for over a year now), I recently spent some time refactoring this into a more coherent CycleInRegion action, supporting explicit region matching as well as support for optionaloverlap thresholds for windows-in-region and region-in-window, which should give users ample power to customize this to their individual needs.

Let me know if you'd prefer a different name for the action (eg. CycleFocusInRegion), or think the functions should be located elsewhere.  If the region is not supplied, it will cycle across all windows on the given output.  Minimized windows are ignored.

Consolatis:
I think it would be great if this could be implemented on top of #3271, e.g. via `<action name="NextWindow" region="current" />` (or a named region). And potentially another argument like `nearRegion="current"`.

tokyo4j:
I think it's acceptable to add `region=all|current` on top of #3271 argument which only cycles through windows that intersect with the same region as the active window, but I don't think we should add anything more like `windowOverlapPercent`. As far as I can tell, @rmt is the only user who relies on this feature and maintaining it forever is not cheap.

rmt:
I hope you mean `region=all|current|ExplicitRegionName` there.  I would not recommend looking for _any_ intersection, or it will become annoying when windows overlap by 16 pixels.

It should either be an _explicit_ region match (I will then maintain a forever fork, extending it with the overlap logic), or it should support fuzzy is-mostly-in-region logic with opinionated presets or explicitly configurable presets.

Another Q here is around immediate focus.. right now, NextWindow/PreviousWindow pop up the OSD, which is something that I am explicitly avoiding in my implementation.  Would it be OK to add support for immediate focus (or alternatively an arg to hide the OSD during cycling) to a NextWindow/PreviousWindow implementation?  Immediate focus requires the MRU-then-LRU switch to enable user-friendly cycling.

rmt:
Also see https://github.com/rmt/labwc/tree/FixedSizeRegions for a variation where I introduced a fixedSize window flag (similar to fixedPosition) and SnapToRegion would align the window within the region without resizing it, effectively replacing the (optional) region/window overlap detection from the current MR and allowing fixedSize windows (which are common for audio plugins/VSTs) to be placed in regions (and still be manually moved around within that region, without losing the association).

PR #1878, opened 2024-06-05
view: fix child views can open outside of usable area to right/bottom

### LINKED ISSUE #1876: Child dialogs may be off-screen when their parents are
This happens with Qt and GTK alike:

 1. Place a window — like FeatherPad's — in such a way that it's partially outside the visible screen *at right and/or bottom*.
 2. Open a child dialog — like Preferences dialog or open-file dialog.

Then, depending on the position of the main window, the dialog may be off-screen partially. The reason is that it's opened centered relative to the main window. Since the important buttons of a dialog (OK, Cancel,...) my not be visible in this case, I think this is a UX issue, although it isn't a critical problem.

It doesn't happen when the main window is off-screen *at top and/or left*. It doesn't happen under Wayfire either. As the GTK window I used for testing was `gtk3-widget-factory`, I don't think it's related to server-side decoration (of Qt windows).

IMHO, a new window should always be opened inside the visible screen as far as possible (i.e., when its size is smaller than that of the screen), regardless of being a main window or a child dialog.

EDIT: I use LabWC 0.7.2, but it happened with previous versions too.

tokyo4j:
Fixes #1876.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -728,12 +728,16 @@ view_compute_centered_position(struct view *view, const struct wlr_box *ref,
 	*x = ref->x + (ref->width - width) / 2;
 	*y = ref->y + (ref->height - height) / 2;
 
-	/* If view is bigger than usable area, just top/left align it */
+	/* Fit the view within the usable area */
 	if (*x < usable.x) {
 		*x = usable.x;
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks!

PR #1880, opened 2024-06-05
Draft: Cycle Near Region, rework to fit better

droc12345:
This is a rough draft, I haven't taken care of everything, this is more to test proof of concept
and wait for comments on improvement.

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -150,13 +150,13 @@ desktop_cycle_view_near_region(struct server *server, struct view *active_view,
 		LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
 
 	struct view *cur;
-        struct wlr_box intersect;
+	struct wlr_box intersect;
 
 	iter = view_prev_no_head_stop;
 
 	cur = iter(&server->views, active_view, criteria);
 	while (cur && cur != active_view) {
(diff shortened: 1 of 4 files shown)


src/desktop.c
```diff
@@ -139,6 +140,32 @@ desktop_cycle_view(struct server *server, struct view *start_view,
...
+
+	struct view *cur;
+	struct wlr_box intersect;
+
+	iter = view_prev_no_head_stop;
+
+	cur = iter(&server->views, active_view, criteria);
+	while (cur && cur != active_view) {
+		wlr_box_intersection(&intersect, &cur->current, &region->geo);
+		if (!cur->minimized && !wlr_box_empty(&intersect)) {
```
rmt, line 160: This would also match a 1% overlap by the looks of it, which isn't ideal.  80% seems to be a kind of magic after much experimentation.  Do you have a numeric keypad?  Try the config I posted (actually, CycleNearRegion region="NE" and not BNE is better), and place windows in each corner, big and small, and in the center.  With the 80% logic, you'd only get the big & small corner windows and not the center window.

I'd consider exact by default with a flag for fuzzy overlap matching.

droc12345:
Good luck.

PR #1882, opened 2024-06-05
feat: add tablet v2 pad support

jp7677:
Looking at the pad support in the tablet v2 protocol, this seems pretty straight forward and works with the "paint" demo in `gtk4-demo` by letting me use the pad buttons to change colors.

```
[ 142826.733] zwp_tablet_seat_v2@29.pad_added(new id zwp_tablet_pad_v2@4278190081)
[ 142826.734] zwp_tablet_pad_v2@4278190081.buttons(4)
[ 142826.735] zwp_tablet_pad_v2@4278190081.path("/sys/devices/pci0000:00/0000:00:14.0/usb1/1-4/1-4.2/1-4.2:1.0/0003:056A:0378.0005/input/input22/event5")
[ 142826.737] zwp_tablet_pad_v2@4278190081.group(new id zwp_tablet_pad_group_v2@4278190082)
[ 142826.738] zwp_tablet_pad_group_v2@4278190082.modes(1)
[ 142826.739] zwp_tablet_pad_group_v2@4278190082.buttons(array[16])
[ 142826.740] zwp_tablet_pad_group_v2@4278190082.done()
[ 142826.741] zwp_tablet_pad_v2@4278190081.done()
[ 143091.732] zwp_tablet_pad_v2@4278190081.enter(49, zwp_tablet_v2@4278190080, wl_surface@34)
[ 143091.760] zwp_tablet_pad_group_v2@4278190082.mode_switch(491748, 49, 0)
[ 151208.708] zwp_tablet_pad_v2@4278190081.leave(73, wl_surface@34)
[ 151208.711] zwp_tablet_pad_v2@4278190081.enter(74, zwp_tablet_v2@4278190080, wl_surface@56)
[ 151208.728] zwp_tablet_pad_group_v2@4278190082.mode_switch(609613, 74, 0)
[ 151209.181]  -> zwp_tablet_pad_v2@4278190081.set_feedback(1, "Black", 74)
[ 151209.186]  -> zwp_tablet_pad_v2@4278190081.set_feedback(2, "Pink", 74)
[ 151209.189]  -> zwp_tablet_pad_v2@4278190081.set_feedback(3, "Green", 74)
[ 154437.283] zwp_tablet_pad_v2@4278190081.button(5553838, 3, 1)
[ 154567.236] zwp_tablet_pad_v2@4278190081.button(5553968, 3, 0)
[ 159764.940] zwp_tablet_pad_v2@4278190081.leave(92, nil)
[ 159764.971] zwp_tablet_pad_v2@4278190081.enter(93, zwp_tablet_v2@4278190080, wl_surface@34)
[ 159765.023] zwp_tablet_pad_group_v2@4278190082.mode_switch(164874, 93, 0)
```

The way of linking by device group is directly from Sway. There is though an `attach_tablet` signal, but I don't know how this is supposed to be wired. I don't see this signal being emitted on my setup.

I don't have rings or strips on my tablet, so just theoretical, but inspired by Sway, implementation of those events.

Draft for now since there might be some rough edges and still needs some research regarding `attach_tablet`. Also needs some additions in the docs.

TODO:

- [x] Look at XWayland, does this actually work?
Seems to be ok and behaves the same now as stylus buttons
- [x] Should we fallback at all to emulate mouse button when mouse emulation is off? Currently we do that when no app has focus, but this only happens when the desktop is empty. Might just be confusing. Alternatively, may be we should always stick to mouse emulation for buttons when there are configured pad mappings?
-> I went for the same approach as for stylus buttons (but without the ability to change buttons)
- [ ] Find out what `attach_tablet` signal is about
- [x] Clarify button mapping in the docs

PS: PR https://github.com/labwc/labwc/pull/1883 is included, so https://github.com/labwc/labwc/pull/1883 has to go first.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -680,28 +680,39 @@ extending outward from the snapped edge.
 	also omits tablet specific features like reporting pen pressure.
 
 *<tablet><map button="" to="" />*
-	Pen buttons emulate regular mouse buttons. If not specified
-	otherwise, the first pen button (Stylus) is mapped to right mouse
-	button click and the second pen button (Stylus2) emulates a middle
-	mouse button click. For mouse emulation, the tip (Tip) is mapped to
-	left mouse click.
+	Pen and pad buttons behave like regular mouse buttons.With mouse
+	emulation set to "no", which is the default, and if not	specified
+	otherwise, the first pen button is mapped to the right mouse button,
+	the second pen button to the middle mouse button and a third pen
+	button is mapped to the side mouse button.
 
 	Supported map *buttons* are:
-	- Tip (mouse emulation only)
+	- Stylus
+	- Stylus2
+	- Stylus3
+
+	The stylus buttons can be mapped *to*:
+	- Right
+	- Middle
+	- Side
+
+	The tip cannot be remapped.
+
+	When using mouse emulation, all pen buttons emulate regular mouse
+	buttons. The tip, stylus and pad buttons can be mapped to all
+	available mouse	buttons. If not specified otherwise, the tip is
+	mapped to left mouse click, the first pen button (Stylus) is mapped
+	to right mouse button click and the second pen button (Stylus2)
+	emulates a middle mouse	button click.
+
+	Supported map *buttons* for mouse emulation are:
+	- Tip
 	- Stylus
 	- Stylus2
 	- Stylus3
 	- Pad
 	- Pad2..Pad9
 
-	The stylus buttons (Stylus, Stylus2 and Stylus3) can be mapped *to*:
-	- Right
-	- Middle
-	- Side
-
-	The tablet pad buttons can be mapped to all available mouse buttons.
-	When using mouse emulation only, the tip and stylus buttons can also be
-	mapped to all available mouse buttons.
 	See mouse section above for all supported mouse buttons.
 
 ## LIBINPUT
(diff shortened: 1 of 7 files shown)


jp7677:
I'm happy with this PR and works on my setup as intended. So ready for review to make us tablet-v2 complete.

I'm still confused though about `attach_tablet` (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_tablet_pad.h#L33). From what I'm seeing, it should be raised after calling `wlr_tablet_v2_tablet_pad_notify_enter` (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/wayland/tablet_v2.c?ref_type=heads#L360) but I don't get that signal if wired and tbh, I don't really see the point of it.

johanmalm:
Thanks. Will review tomorrow unless someone else gets there first.

johanmalm:
It looks fine to me, but I have one question before merge:

Why do we have to reset _all_ tablet-pad links in `tablet_pad_attach_tablet()` - i.e. why iterate over all tablets rather than just process the subject-tablet.

jp7677:
> Why do we have to reset _all_ tablet-pad links in `tablet_pad_attach_tablet()` - i.e. why iterate over all tablets rather than just process the subject-tablet.

I'm also calling this function in tablet_destroy:

https://github.com/labwc/labwc/blob/e74e9640bbda534b452b8db3a66f4d4ec88f0454/src/input/tablet.c#L514

thus it resets all tablet - pad links first to ensure that we have dealt with removed tablets. That said, I got the order wrong, I need to clear an to be removed tablet from the list first. Will fix, give me a minute ;)

johanmalm:
Looks good. Let's go for it.

PR #1890, opened 2024-06-09
view: implement `cascade` placement policy

tokyo4j:
This is what I suggested in https://github.com/labwc/labwc/pull/1772#issuecomment-2093799066.

This adds following settings:
```xml
<placement>
  <policy>cascade</policy>
  <cascadeOffset x="40" y="30" />
</placement>
```

`cascade` policy places a new window at the center of the screen like `center` policy, but possibly shifts its position to bottom-right so the new window doesn't cover existing windows.
Its algorithm is copied from Kwin ([link](https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/placement.cpp#L589)), combined with some helper functions I added (`wlr_box_contains()`, `wlr_box_intersects()` and `wlr_box_union()`).

https://github.com/labwc/labwc/assets/22029524/df90c7d2-7fa3-4550-8fa9-eb2d83a709dc

I think this policy achieves a good balance of predictability and visibility when opening multiple windows.

But please treat this PR as just a suggestion. In KDE, this was proposed in 2003 ([link](https://bugs.kde.org/show_bug.cgi?id=58063)) and implemented as an enhancement in `center` policy in 2022. So adding this to labwc may also require some discussion as well.

And I also want to emphasize I agree with @jlindgren90's comment in #1772:

> I recall Metacity had a design goal of having just one placement scheme that was "good enough" for most use cases -- it would be nice if we could achieve that.

So, closing this PR and seeking another way that fits our goal better is welcome. Then maybe I can personally maintain a patch and use `cascade` policy until we manage to make an alternative.

### DIFF
--- a/include/common/box.h
+++ b/include/common/box.h
@@ -1,0 +1,17 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_BOX_H
+#define LABWC_BOX_H
+
+#include <wlr/util/box.h>
+
+bool
+box_contains(struct wlr_box *box_super, struct wlr_box *box_sub);
+
+bool
+box_intersects(struct wlr_box *box_a, struct wlr_box *box_b);
+
+/* Returns the bounding box of 2 boxes */
+void
+box_union(struct wlr_box *box_dest, struct wlr_box *box_a, struct wlr_box *box_b);
+
+#endif /* LABWC_BOX_H */
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -901,6 +901,10 @@ entry(xmlNode *node, char *nodename, char *content)
 		}
 	} else if (!strcasecmp(nodename, "xwaylandPersistence.core")) {
 		set_bool(content, &rc.xwayland_persistence);
+	} else if (!strcasecmp(nodename, "x.cascadeOffset.placement")) {
+		rc.placement_cascade_offset_x = atoi(content);
+	} else if (!strcasecmp(nodename, "y.cascadeOffset.placement")) {
+		rc.placement_cascade_offset_y = atoi(content);
 	} else if (!strcmp(nodename, "name.theme")) {
 		rc.theme_name = xstrdup(content);
 	} else if (!strcmp(nodename, "cornerradius.theme")) {
@@ -1229,6 +1233,8 @@ rcxml_init(void)
 	has_run = true;
 
 	rc.placement_policy = LAB_PLACE_CENTER;
+	rc.placement_cascade_offset_x = 0;
+	rc.placement_cascade_offset_y = 0;
 
 	rc.xdg_shell_server_side_deco = true;
 	rc.ssd_keep_border = true;
--- a/src/view.c
+++ b/src/view.c
@@ -4,6 +4,7 @@
 #include <strings.h>
 #include <wlr/types/wlr_output_layout.h>
 #include <wlr/types/wlr_security_context_v1.h>
+#include "common/box.h"
 #include "common/macros.h"
 #include "common/match.h"
 #include "common/mem.h"
@@ -877,6 +878,94 @@ view_center(struct view *view, const struct wlr_box *ref)
 	}
 }
 
+/*
+ * Algorithm based on KWin's implementation:
+ * https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/placement.cpp#L589
+ */
+static void
+view_cascade(struct view *view)
+{
+	/* "cascade" policy places a new view at center by default */
+	struct wlr_box center = view->pending;
+	view_compute_centered_position(view, NULL,
+		center.width, center.height, &center.x, &center.y);
+	struct border margin = ssd_get_margin(view->ssd);
+	center.x -= margin.left;
+	center.y -= margin.top;
+	center.width += margin.left + margin.right;
+	center.height += margin.top + margin.bottom;
+
+	/* Candidate geometry to which the view is moved */
+	struct wlr_box candidate = center;
+
+	struct wlr_box usable = output_usable_area_in_layout_coords(view->output);
+
+	/* TODO: move this logic to rcxml.c */
+	int offset_x = rc.placement_cascade_offset_x;
+	int offset_y = rc.placement_cascade_offset_y;
+	struct theme *theme = view->server->theme;
+	int default_offset = theme->title_height + theme->border_width + 5;
+	if (offset_x <= 0) {
+		offset_x = default_offset;
+	}
+	if (offset_y <= 0) {
+		offset_y = default_offset;
+	}
+
+	/*
+	 * Keep updating the candidate until it doesn't cover any existing views
+	 * or doesn't fit within the usable area.
+	 */
+	bool candidate_updated = true;
+	while (candidate_updated) {
+		candidate_updated = false;
+		struct wlr_box covered = {0};
+
+		/* Iterate over views from top to bottom */
+		struct view *other_view;
+		for_each_view(other_view, &view->server->views,
+				LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+			struct wlr_box other = ssd_max_extents(other_view);
+			if (other_view == view
+					|| view->minimized
+					|| !box_intersects(&candidate, &other)) {
+				continue;
+			}
+			/*
+			 * If the candidate covers an existing view whose
+			 * top-left corner is not covered by other views,
+			 * shift the candidate to bottom-right.
+			 */
+			if (box_contains(&candidate, &other)
+					&& !wlr_box_contains_point(
+						&covered, other.x, other.y)) {
+				candidate.x = other.x + offset_x;
+				candidate.y = other.y + offset_y;
+				if (!box_contains(&usable, &candidate)) {
+					/*
+					 * If the candidate doesn't fit within
+					 * the usable area, fall back to center
+					 * and finish updating the candidate.
+					 */
+					candidate = center;
+					break;
+				} else {
+					/* Repeat with the new candidate */
+					candidate_updated = true;
+					break;
+				}
+			}
+			/*
+			 * We use just a bounding box to represent the covered
+			 * area, which would be fine for our use-case.
+			 */
+			box_union(&covered, &covered, &other);
+		}
+	}
+
+	view_move(view, candidate.x + margin.left, candidate.y + margin.top);
+}
+
 void
 view_place_by_policy(struct view *view, bool allow_cursor,
 		enum view_placement_policy policy)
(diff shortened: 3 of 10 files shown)


droc12345:
cascaded should be cascade in labwc-actions, seems to work alright.

I've long missed this particular "action", going all the way to the corner is too far, for my uses.

tokyo4j:
>cascaded should be cascade in labwc-actions, seems to work alright.

> I've long missed this particular "action", going all the way to the corner is too far, for my uses.

> Edit to add: In some places it's cascadeOffset and others cascadeOffsets

Thank you. Fixed.

src/view.c
```diff
@@ -874,6 +875,69 @@ view_center(struct view *view, const struct wlr_box *ref)
...
+			/*
+			 * Fall back to center if the canditate geometry
+			 * doesn't fit within the usable area.
+			 */
+			if (!wlr_box_contains(&usable, &self)) {
+				view_move(view, center.x, center.y);
+				return;
+			}
+			/* Repeat with the new candidate geometry */
+			goto cascade_if_covers;
```
tokyo4j, line 928: I used `goto` for simplicity, but KWin uses a flag variable `noOverlap` + `break;` instead ([link](https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/placement.cpp#L613-L620)).

Perhaps we should only use `goto` for cleanup logic and avoid using it for loop.

johanmalm: I'm normally quite relaxed about `goto`s, but they are often derided by more learned people (than me).

I think `kwin`'s `while (!noOverlap)` certainly makes it easier for the reader to understand what's going on. So on the basis that readability is important, I think we ought to wrap it in a while-loop.

include/common/box.h
```diff
@@ -0,0 +1,17 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_BOX_H
+#define LABWC_BOX_H
+
+#include <wlr/util/box.h>
+
+bool
+wlr_box_contains(struct wlr_box *box_super, struct wlr_box *box_sub);
```
johanmalm, line 8: I prefer leaving `wlr_` out here on the basis that `wlr_` should be reserved for the `wlroots` API.

Aside: see 3rd paragraph of https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#naming-conventions

droc12345:
The main reason for people's disdain for goto's, is that they have a tendency to be overused and abused.
Though I try and stay away from them as a general rule.

include/common/box.h
droc12345: Change wlr_ to lab_, for the new functions?'s, is t

johanmalm:
This looks great!

I admire the pursuit of the _the_ policy :smile:

Out of curiosity, what does the kwin `center` policy do? Does it just center the first one and then proceed to cascade?

droc12345:
Looks like kwin, does a cascade of overlap at the bottom of several placement functions.
I do remember openbox having a cascade effect, but I don't remember whether it happened
with all the placement options.

droc12345:
If I open multiple foot terminals, they cascade, the same if I open more than one calculator.
The calculator is a small window, and if I open it first then foot, foot doesn't cascade, it centers.
Is this expected behavior?

src/view.c
```diff
@@ -874,6 +875,69 @@ view_center(struct view *view, const struct wlr_box *ref)
...
+	struct wlr_box covered = {0};
+
+	/* Iterate over views from top to bottom */
+	struct view *other_view;
+	for_each_view(other_view, &view->server->views,
+			LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) {
+		struct wlr_box other = ssd_max_extents(other_view);
+		if (other_view == view || !wlr_box_intersects(&self, &other)) {
+			continue;
+		}
```
Consolatis, line 910: Does this miss a `other_view->minimized` check?

src/view.c
```diff
@@ -874,6 +875,69 @@ view_center(struct view *view, const struct wlr_box *ref)
...
+		 * If the candidate geometry covers an existing view which is not
+		 * already covered, shift the candidate geometry to bottom-right.
+		 */
+		if (wlr_box_contains(&self, &other)
+				&& !wlr_box_contains(&covered, &other)) {
+			self.x = other.x + rc.placement_cascade_offset_x;
+			self.y = other.y + rc.placement_cascade_offset_y;
+			/*
+			 * Fall back to center if the canditate geometry
+			 * doesn't fit within the usable area.
```
Consolatis, line 921: Nitpick: `s/canditate/candidate/`

src/config/rcxml.c
```diff
@@ -1223,6 +1227,8 @@ rcxml_init(void)
 	has_run = true;
 
 	rc.placement_policy = LAB_PLACE_CENTER;
+	rc.placement_cascade_offset_x = 40;
+	rc.placement_cascade_offset_y = 30;
```
Consolatis, line 1231: Should `x` really be different to `y` by default? I don't know anything about how this feature works for other implementations but it feels kind of off to me when the values aren't the same.

Consolatis:
In general I like this policy, and I am completely fine with adding it to labwc.
Personally I'd also be open to use this as a default when everything is working correctly.

src/config/rcxml.c
01micko: Personally, I would have `y > x` if I had a vertical monitor so as is is ok IMHO for horizontal orientation which would be the majority of use cases.

Consolatis: The settings for the titlebar height (for example a font size 20 for (in)active window) may also play a role here.

tokyo4j: KWin sets the offset to 1/48 of the usable area ([link](https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/placement.cpp#L364-L368)). So maybe we can allow `%` in the settings to set the offsets relative to the size of usable area.

Another alternative is set the offset to title bar height (perhaps x1.2 or something of it to make it easier to distinguish windows without custom border color or drop shadows) by default and use it for both x and y.

tokyo4j:
> Something feels off when testing, depending on what window has focus when spawning a new window the placements can differ drastically. Sometimes the original center is used even though there are other windows below (and usable area isn't yet reached), sometimes it skips a step when there are multiple other windows behind it.

I can't tell the exact behavior you describe, but I found a weird behavior like below. Is this the same as yours?

https://github.com/labwc/labwc/assets/22029524/d682cc0b-84f7-4ce2-ade7-94fde0cd6039

This happens in the following algorithm:
- The default candidate geometry at the center covers the first existing window, so the candidate geometry cascades against it.
- The candidate geometry covers the second existing window, so the candidate geometry cascades against it.
- The candidate geometry is now not within the usable area, so fall back to center.

I'll try to come up with a better algorithm to avoid this.

src/view.c
tokyo4j: You're right. Will fix.

tokyo4j:
> Out of curiosity, what does the kwin center policy do? Does it just center the first one and then proceed to cascade?

Yes, they implemented this by overwriting the behavior of center policy. Someone complained about it and created a PR to allow to disable cascading ([link](https://invent.kde.org/plasma/kwin/-/merge_requests/5522)), but it was rejected. But it should be noted that for KWin it's possible to disable cascading by using custom script.

We have `<action name="AutoPlace" policy="center">` which should just center the window without any magic, so I think implementing it as a separate policy would suit labwc better.

droc12345:
I'm happy with it as is, as the most common use (for me) would be multiple foot terms, which works ok.
Though I only opened a couple of windows to check workability, not 10's of windows to stress test.

If the 1st window is too small, in relation to the overlapping window, then maybe something like automatic
placement, where it goes all the way to the corner might be an option.

src/view.c
tokyo4j: Done.

src/view.c
tokyo4j: Fixed.

src/view.c
tokyo4j: Done.

droc12345:
Still using this with no problems, might want to change from draft to PR and set merge for after 7.3 release.

tokyo4j:
> Still using this with no problems, might want to change from draft to PR and set merge for after 7.3 release.

Agreed. I'll daily-drive this PR myself from today, too.

I've just pushed a small fixup that ignores windows whose top-left corner is covered, rather than windows completely covered. This fixes the problem I mentioned in https://github.com/labwc/labwc/pull/1890#issuecomment-2157351051, but I'm not 100% sure this is right as it causes a new window to cover cascaded-and-lowered windows like the picture below:

droc12345:
That's acceptable (to me) because as long as some part of the window shows, it's easy to click on it and bring it
forward. Besides there's only so much cascading one can do. I typically deal with 3-4 windows and it works well
for that.

I run with this patch and the wlr 0.18 patch from Consolatis, and I'm pretty happy with things.
Note: I don't do decorations, shading, etc so I''m pretty easy to please.

tokyo4j:
> Edit to add: when running out of space heading towards the bottom right, just start from center and cascade to the upper left.
With enough windows you will eventually overlap no matter which direction but anyone that opens that many windows in a single workspace should understand that that is going to happen.

I don't think cascading to upper left is a good idea because a new window should not hide the titlebar of previous window. Positioning a new window to top-left corner or somewhere if it goes out of workarea is an option though. However, I currently don't know what algorithm can be used for this.

droc12345:
No problem it was just a random thought. 
As I say, I'm happy with it right now, I get center with 1st window and cascade after that. 

Using a 100 offset with my 4k monitor gives about 1/2 inch (real value) cascade, which works for me.

tokyo4j:
I think this PR is ready.

I personally want to set `cascade` by default in the future as it feels the most "natural" policy for me, but let's do it after we get more reviews.

johanmalm:
LGTM. Have tested.

> I personally want to set cascade by default in the future as it feels the most "natural" policy for me, but let's do it after we get more feedbacks.

I'm fine with that - but as you say - let's wait and see how it survives the encounter with "users" first :smile:

knm100:
Is this behavior normal?

Uploading 154530 .mp4…

tokyo4j:
> Is this expected?

The algorithm seems to be working as I expected. What did you find unexpected? (any feedback is welcome)

droc12345:
Looks like it's working as intended.
What do you think should happen when you run out of cascade space?

knm100:
I know the reason, it should be that the pcmanfm-qt window is too large, which causes it to go beyond the bottom screen, which is why the behavior in my video appears.    indeed, it's working as intended.     Thank you for your reply
swappy-20240730-164117

PR #1895, opened 2024-06-09
xwayland: reset cursor image on cursor theme reload

Consolatis:
As wlr_xwayland caches the pixel data when not yet started up due to the delayed lazy startup approach, we do have to re-set the xwayland cursor image when reloading the cursor theme. Otherwise the first X11 client connected will cause the xwayland server to use the cached (and destroyed) pixel data.

To reproduce:
- Compile with b_sanitize=address,undefined
- Start labwc (nothing in autostart that could create a X11 connection, e.g. no GTK or X11 application)
- Reconfigure
- Start some X11 client

### DIFF
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -95,5 +95,7 @@ void xwayland_adjust_usable_area(struct view *view,
 
 void xwayland_update_workarea(struct server *server);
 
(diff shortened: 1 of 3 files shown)


johanmalm:
Thanks.

PR #1899, opened 2024-06-10
keyboard: handle onRelease keybind after long press

johanmalm:
...that is longer than the specified repeatDelay interval.

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -235,6 +235,20 @@ is_modifier_key(xkb_keysym_t sym)
 	}
 }
 
+static bool
+is_modifier(struct wlr_keyboard *wlr_keyboard, uint32_t evdev_keycode)
+{
+	const xkb_keysym_t *syms = NULL;
+	int nr_syms = xkb_state_key_get_syms(wlr_keyboard->xkb_state,
+		evdev_keycode + 8, &syms);
+	for (int i = 0; i < nr_syms; i++) {
+		if (is_modifier_key(syms[i])) {
+			return true;
+		}
+	}
+	return false;
+}
+
 static struct keyinfo
 get_keyinfo(struct wlr_keyboard *wlr_keyboard, uint32_t evdev_keycode)
 {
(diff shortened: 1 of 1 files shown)


johanmalm:
@spl237 Test welcome.

I found that when holding `Super_L` for a while, it doesn't trigger the action on release... which is what this patch fixes.

spl237:
Tested here - looks good to me. 

I'd spotted that holding the key for a long time meant the binding didn't trigger, but had no idea why! Many thanks for tracking it down.

Consolatis:
LGTM, not tested.

stefonarch:
Previously I used patches for this and magnifier and else and all was fine, now I've compiled last master and can't use `Super_L` for lxqt-panel's menu anymore as I use shortcuts like  W-p for applications and every time the menu pops up now. Not sure if it this commit as I used a patch which wasn't necessary the last commit for this feature. Unrelated but `-1` doesn't work too anymore for magnifier. Can try to bisect later today.

spl237:
I am seeing one problem with the magnifier - I have magnifier zoom on W+mouse scroll. If I zoom, the menu pops up when I release the W key after I have finished scrolling. I suspect there needs to be a line somewhere which cancels the on-release behaviour if a mouse action has happened, not just a keyboard action - will have a look.

spl237:
Further - the problem with mouse scroll isn't related to this patch; it just makes it more obvious because previously holding down the W key for more than a second or so (i.e. while scrolling) didn't trigger an on-release event. But if you hold down W, scroll very quickly and release (without this patch), the menu pops up.

spl237:
Actually, I am wondering if it would be cleaner to just set cur_keybind back to NULL in actions_run rather than in individual keyboard and mouse event handlers? That will then reset it if any binding has actually triggered, and leave it pending otherwise so it is used by the release event. You can also then remove a couple of explicit calls setting it to NULL in keyboard.c.

I've just tried this here, and it seems to work well - all my bindings still work, but I no longer get the menu popping up when I release the windows key at the end of a mouse scroll.

spl237:
Submitted a PR - https://github.com/labwc/labwc/pull/1906

I'm not 100% convinced the fix is really as simple as this (!), but it seems to work for me.

PR #1906, opened 2024-06-13
Fix release keybindings being triggered after mouse bindings

spl237:
This fixes the problem where if a mousebinding uses a modifier key which has an onRelease binding, the mousebinding does not cancel the pending onRelease binding if it is triggered.

The approach is to clear the pending release binding whenever any bound action occurs, which means it only needs to take place at one point in the code.

There may be other issues here which I have not considered, but this appears to fix the problem for me without breaking any other bindings - more testing and/or advice much appreciated.

### DIFF
--- a/include/input/keyboard.h
+++ b/include/input/keyboard.h
@@ -9,6 +9,8 @@ struct seat;
 struct keyboard;
 struct wlr_keyboard;
 
+extern struct keybind *cur_keybind;
+
 void keyboard_reset_current_keybind(void);
 void keyboard_configure(struct seat *seat, struct wlr_keyboard *kb,
 	bool is_virtual);
--- a/include/input/keyboard.h
+++ b/include/input/keyboard.h
@@ -9,8 +9,6 @@ struct seat;
 struct keyboard;
 struct wlr_keyboard;
 
-extern struct keybind *cur_keybind;
-
 void keyboard_reset_current_keybind(void);
 void keyboard_configure(struct seat *seat, struct wlr_keyboard *kb,
 	bool is_virtual);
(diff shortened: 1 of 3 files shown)


johanmalm:
I'm good with adding this to `actions_run()` but we ought to do it with a function like `keyboard_reset_current_keybind()` with a comment to explain that `/* This cancels any pending on-release keybinds */`

When I first read the patch my initial feeling was that it goes against the grain to add the coupling to/from `actions_run()`, but looking at all the places it gets called from I think it's safer to do it to save future bugs caused by people forgetting to call the reset separately.

If it becomes a problem in future - for example if we ever need `actions_run()` to be called without resetting the current-keybind - we can just add a wrapper function.

spl237:
OK, makes sense - will change accordingly tomorrow and submit.

spl237:
Done - much tidier!

johanmalm:
Looks like `keyboard_reset_current_keybind()` is missing. Guess it compiles locally - so maybe you haven't pushed it?

I was expecting something like...

```
void
keyboard_reset_current_keybind()
{
  cur_keybind = NULL;
}
```

spl237:
Weird. That function is in my version, and I didn’t write it - it was already in keyboard.c. And it built fine in the commit tests, which all passed.

Are you trying it against an old version, perhaps?

spl237:
Yes, just checked - that function is already on master - see https://github.com/labwc/labwc/blob/4ee538c6cd60b3fd7daeaa4e394d2b6a519e7642/src/input/keyboard.c#L44

It was pulled in as part of the previous PR - are you trying this new PR against code which doesn't include that?

johanmalm:
Apologies. Ignore me. I'll to keep up in future :smile:

johanmalm:
Thanks for reporting + fixing.

spl237:
Excellent - many thanks!

PR #1907, opened 2024-06-13
menu: ignore <item> without parent <menu>

johanmalm:
...to avoid assert() in item_create() because current_menu is NULL.

Reproduce crash with...

    <?xml version="1.0" encoding="utf-8"?>
    <openbox_menu>
      <item label="foo"/>
    </openbox_menu>


Consolatis:
:)

PR #1910, opened 2024-06-14
menu: capitalize Roll Up/Down

johanmalm:
...because it seems like better English and is more consistent with other menu entries.

Same for 'Move Left' and 'Move Right'

### DIFF
--- a/docs/menu.xml
+++ b/docs/menu.xml
@@ -12,7 +12,7 @@
   <item label="Fullscreen">
     <action name="ToggleFullscreen" />
   </item>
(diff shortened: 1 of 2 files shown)


01micko:
Nothing against this PR but we'll have to fix up NLS also. Can of course be done later.

johanmalm:
> Nothing against this PR but we'll have to fix up NLS also. Can of course be done later.

Thanks. I'll the labwc.pot change so I don't mess up the LXQt translation service.

PR #1915, opened 2024-06-16
feat: add ToggleTabletMouseEmulation action

jp7677:
Useful for switching between application-defined and mapped tablet  pad buttons. Also for quickly troubleshooting tablet behavior.

Probably not the most useful action, but simple implementation so why not.


johanmalm:
Looks good.

Issue #1916, opened 2024-06-16
Assertion failed at `wlr_libinput_get_device_handle()` in nested session


tokyo4j:
CC @jp7677

After 3c0cea92f26557103285a716ef7fd6ee5807b925, assertion fails in nested session with the following message:
```
labwc: backend/libinput/tablet_tool.c:56: device_from_tablet: Assertion `wlr_tablet->impl == &libinput_tablet_impl' failed.
```

Perhaps this doesn't occur in PCs without touchscreen (I have envy 13 x360).

I don't fully understand the code, but I suggest a change like this:
```diff
diff --git a/src/input/tablet-pad.c b/src/input/tablet-pad.c
index 1729e356..df913fc8 100644
--- a/src/input/tablet-pad.c
+++ b/src/input/tablet-pad.c
@@ -25,6 +25,9 @@ tablet_pad_attach_tablet(struct seat *seat)
 	/* loop over all tablets and all pads and link by device group */
 	struct drawing_tablet *tablet;
 	wl_list_for_each(tablet, &seat->tablets, link) {
+		if (!wlr_input_device_is_libinput(tablet->wlr_input_device)) {
+			continue;
+		}
 		struct libinput_device *tablet_device =
 			wlr_libinput_get_device_handle(tablet->wlr_input_device);
 		struct libinput_device_group *tablet_group =

```

jp7677:
Thanks for the heads up, does https://github.com/labwc/labwc/pull/1917 works for you?

tokyo4j:
Yes. Thank you!

PR #1917, opened 2024-06-16
input: test if tablet device is a libinput device

jp7677:
Otherwise we might end with a failed assertion if the tablet or pad isn't a libinput device.

Fixes https://github.com/labwc/labwc/issues/1916

### DIFF
--- a/src/input/tablet-pad.c
+++ b/src/input/tablet-pad.c
@@ -25,12 +25,25 @@ tablet_pad_attach_tablet(struct seat *seat)
 	/* loop over all tablets and all pads and link by device group */
 	struct drawing_tablet *tablet;
 	wl_list_for_each(tablet, &seat->tablets, link) {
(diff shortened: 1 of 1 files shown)


Consolatis:
I assume this is happening nested when the the outer compositor has a tablet attached and its exposed via wayland protocol to the nested compositor? So what ends up in the list is now wayland protocol backed tablet / pad? If yes maybe we could add a comment to explain the situation.

jp7677:
> I assume this is happening nested when the outer compositor has a tablet attached and its exposed via wayland protocol to the nested compositor? So what ends up in the list is now wayland protocol backed tablet / pad? If yes maybe we could add a comment to explain the situation.

Thanks, added a small comment. Let me know if you prefer this to be more detailed.

johanmalm:
Thanks

PR #1919, opened 2024-06-17
PrevioiusWindow - fix problems with cycling not working correctly

### LINKED ISSUE #449: PreviousWindow is the same as NextWindow
PreviousWindow walks through the window list in the same direction as NextWindow.
Reproduce:
1. Open 5 windows.
2. Start walking through the windows using NextWindow hotkey.
3. Now use PreviousWindow hotkey.

Expected: PreviousWindow walks in the opposite direction.
Result: PreviousWindow and NextWindow do the same thing.

droc12345:
Should fix #449 PreviousWindow is the same as NextWindow

Edit to add: With the addition of osd_state.direction, there really is no need for LAB_CYCLE_*
enums or the check for them. It's mostly a change to desktop_cycle_view but it doesn't hurt to leave
the enum stuff till after the freeze. Actually this whole PR could wait, but it's not particularly invasive. 

ETA2: forget what I said above about LAB_CYCLE_* as I refactored it to keep it and fix the directional cycling.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -99,11 +99,14 @@ Actions are used in menus and keyboard/mouse bindings.
 	Resize and move active window according to the given region.
 	See labwc-config(5) for further information on how to define regions.
 
-*<action name="NextWindow" />*
-	Cycle focus to next window.
-
+*<action name="NextWindow" />*++
 *<action name="PreviousWindow" />*
-	Cycle focus to previous window.
+	Cycle focus to next/previous window respectively.++
+	Default keybind for NextWindow is Alt-Tab.
+
+	The shift key is used to reverse direction while cycling.
+
+	The arrow keys are used to move forwards/backwards while cycling.
 
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -210,6 +210,12 @@ struct seat {
 struct lab_data_buffer;
 struct workspace;
 
+enum lab_cycle_dir {
+	LAB_CYCLE_DIR_NONE,
+	LAB_CYCLE_DIR_FORWARD,
+	LAB_CYCLE_DIR_BACKWARD,
+};
+
 struct server {
 	struct wl_display *wl_display;
 	struct wl_event_loop *wl_event_loop;  /* Can be used for timer events */
@@ -349,6 +355,8 @@ struct server {
 		struct wlr_scene_tree *preview_parent;
 		struct wlr_scene_node *preview_anchor;
 		struct multi_rect *preview_outline;
+		enum lab_cycle_dir initial_direction;
+		bool initial_keybind_contained_shift;
 	} osd_state;
 
 	struct theme *theme;
@@ -441,12 +449,6 @@ struct view *desktop_topmost_focusable_view(struct server *server);
  * based on the existence of a fullscreen window on the current workspace.
  */
 void desktop_update_top_layer_visiblity(struct server *server);
-
-enum lab_cycle_dir {
-	LAB_CYCLE_DIR_NONE,
-	LAB_CYCLE_DIR_FORWARD,
-	LAB_CYCLE_DIR_BACKWARD,
-};
 
 /**
  * desktop_cycle_view - return view to 'cycle' to
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -384,6 +384,16 @@ handle_menu_keys(struct server *server, struct keysyms *syms)
 }
 
 static void
+toggle_direction(enum lab_cycle_dir *direction)
+{
+	if (*direction == LAB_CYCLE_DIR_FORWARD) {
+		*direction = LAB_CYCLE_DIR_BACKWARD;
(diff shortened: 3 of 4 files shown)


droc12345:
jp7677, I modified what you had there, as it wasn't working right.
It should be correct now, whether using alt-tab with/without arrow keys

droc12345:
Made some more changes, and checked to make sure the logic was correct for all cases.

next window, goes to next window, repeatedly hitting tab goes forward
arrow keys (up/left) (right/down) work scrolling in direction expected.
the same applies to previous window.

This adds a bool direction flag in struct osd_state.
The direction flag is necessary to make handle_cycle_view_key work correctly, 
which is why repeatedly scrolling backwards was not correct.

After doing the above, I looked at removing enum LAB_CYCLE_* and replaced with using the direction bool.

It works for me, even applies cleanly on top of Consolatis work for wlr 0.18

docs/labwc-actions.5.scd
```diff
@@ -99,11 +99,16 @@ Actions are used in menus and keyboard/mouse bindings.
...
 *<action name="PreviousWindow" />*
-	Cycle focus to previous window.
+	Cycle focus to next/previous window respectively.++
+	Default keybind for NextWindow is Alt-Tab.
+
+	The shift key is used to reverse direction while cycling.
+
+	The arrow keys are used to move forwards/backwards while cycling.
+
+	*Note:* Shift key should not be used in keybind for these 2 actions.
```
johanmalm, line 111: This is a no-go for me, because I use `S-A-Tab` for `PreviousWindow` and I think that's common (therefore arguably expected) behaviour.

include/labwc.h
```diff
@@ -345,6 +345,7 @@ struct server {
 	struct osd_state {
 		struct view *cycle_view;
 		bool preview_was_enabled;
+		bool direction;
```
johanmalm, line 348: Suggest a better word than `direction` - for example `forwards`.
Although given how we use it, I think it's easier to read the enum anyway, so prefer to keep that.

src/input/keyboard.c
```diff
@@ -397,19 +397,28 @@ handle_cycle_view_key(struct server *server, struct keyinfo *keyinfo)
...
+			}
+		}
+		bool dir =  server->osd_state.direction;
+		if (back_key >= 0) {
+			/* an arrow key has been hit */
+			dir =  back_key;
+		}
+		if (keyinfo->modifiers & WLR_MODIFIER_SHIFT) {
+			/* Shift held down - change direction */
+			dir = 1 - dir;
```
johanmalm, line 420: I prefer `dir = !dir`

johanmalm (commented):
The main comment here is that we must support `S-A-Tab`. I don't want to break that.
Let me submit a quick patch to compare thinking.

include/labwc.h
droc12345: I'm not hung up on the name, I picked direction because that's what it controls, forwards/backwards.

src/input/keyboard.c
droc12345: No personal preference, I just want to swap the bool, ie direction

droc12345: if "& !dir" is added after WLR_MODIFIER_SHIFT, then S-A-Tab should work properly, but
then you lose the shift key to reverse direction, when doing PreviousWindow.

docs/labwc-actions.5.scd
droc12345: Fixed, borrowed a little logic from your WIP.

droc12345:
Ok, do a review now.

I think I have taken care of all concerns, and the directional changes seem to work.
I made an else if with arrow keys & shift key, as only one should be active at any one time
otherwise you might get unpredictable results.

johanmalm:
Looks good at a glance.
I'll try to test + review properly tomorrow.

johanmalm:
`A-S-Tab` (for `PrevWindow`) starts backwards, but continues backwards regardless of shift being pressed or not.

I have pushed a [fixup] commit. I have tested and am happy to go as is except it would be good to tidy up your commit message (space after first line, etc).

Suggest some along the below (wrapping at 74 - which I can't see in the browser):

```
action: remember initial direction of {Next,PreviouWindow}

...when cycling windows. Also make the direction-toggle when shift is pressed
relative to the initial direction. For example if W-j is bound to PreviousWindow, subsequent key presses will continue to cycle backwards unless shift if pressed.
```

droc12345:
Any time I submit something, and you think the commit message needs changing, 
I have no problem with you changing it. 

It's running the checks now, but should pass.

droc12345:
> `A-S-Tab` (for `PrevWindow`) starts backwards, but continues backwards regardless of shift being pressed or not.

That's because it starts off from next/prev  action, but repeat cycling is done with handle_cycle_view_key.

My original thought was it needed A-S to continue woriking, but now that I think about it, that was a wrong
assumption. Once it's set motion, only the A key needs to be kept down. 

Good catch, good fix, though I don't normally use goto's there is a place for them, sometimes they make
the code path a little cleaner. Anyway, it compiles and works. I do like the ability to shift directions while
in PreviousWindow and still be able to use S-A-Tab to change dir.

johanmalm:
Thanks

Issue #1905, opened 2024-06-13
Flickering with magnifier and gammastep


stefonarch:
It doesn't matter the dimension, the magnified area flickers always at some seconds interval:

https://github.com/labwc/labwc/assets/10681413/a066af9f-5b5b-46b7-aed2-7a7ae5a9b332

Consolatis:
Is this on the DRM backend or on the wayland backend?
Also what renderer is in use, the usual gles2 one or something other like vulkan?

stefonarch:
Some responses, not sure about the right commands:

```
glxinfo | grep "OpenGL renderer"
OpenGL renderer string: AMD Radeon Vega 3 Graphics (radeonsi, raven2, LLVM 17.0.6, DRM 3.54, 6.6.32-1-lts)


$ eglinfo | grep -i "EGL_VENDOR"
$ vulkaninfo | grep "deviceName"
        deviceName        = AMD Radeon Vega 3 Graphics (RADV RAVEN2)

```

droc12345:
Turn on debugging, and startw labwc with stdout/stderr redirections, then exit labwc.
View the log file, it will tell you what it's using, the backend as well as the renderer.

stefonarch:
```
00:00:00.050 [render/egl.c:536] Using EGL_PLATFORM_DEVICE_EXT
00:00:00.051 [render/gles2/renderer.c:842] Creating GLES2 renderer
...
[backend/wayland/backend.c:436] Starting Wayland backend

```

tokyo4j:
Could you post the full debug log when you experience the flicker?
Perhaps it includes messages like "Failed to build output state for HDMI-1" or "Failed to commit output HDMI-1".

stefonarch:
Not very helpful I guess:
```

tokyo4j:
I think I figured out the cause.

When I run a command like `gammastep -t 2000:6000 -l 0:0`, gammastep regularly updates the gamma setting in some interval.
And labwc currently doesn't draw the magnifier at the frame when the gamma setting is updated:
https://github.com/labwc/labwc/blob/1dadb128075864f0c0449700f2b6f9723ef842fc/src/output.c#L77-L101

So, drawing the magnifier also inside the if-block would solve the issue.

Consolatis:
[...]

PR #1920, opened 2024-06-17
magnifier: fix flickering on simultaneous gamma changes

Consolatis:
Gamma changes take another code path and thus did not render the magnifier.
This patch adds the magnfifier to the gamma code path and also slightly refactors the gamma handling.

Fixes:
- #1905

---

Completely untested.

### DIFF
--- a/include/common/scene-helpers.h
+++ b/include/common/scene-helpers.h
@@ -7,6 +7,7 @@
 struct wlr_scene_node;
 struct wlr_surface;
 struct wlr_scene_output;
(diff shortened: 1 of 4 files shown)


johanmalm:
LGTM. Not merging as still set to 'draft' + has one `[fixup]` commit.

johanmalm:
Have tested. 
Without patch magnifier disappears on gammastep change (but comes back on next frame).
With patch the magnifiers stays as one would expect.

Issue #1862, opened 2024-05-31
Set squared SSD corners when SSD client is tiled and notified about being tiled


jp7677:
Follow up from https://github.com/labwc/labwc/pull/1463

It would be nice if the SSD corners would follow GTK3/4 corners of getting squared when the client is tiled and tiling notifications for clients are enabled. 

Since we have already squared corners for maximized windows, I guess the infrastructure is somewhat there.

Consolatis:
Yes, somewhat. We'd have to selectively round / square the left / right ones individually though.

I am also not convinced that this is a good idea for people using gaps and thus have some margin between their snapped windows to the screen edge. However, as we already support the tiling notification setting for this purpose I guess that should be fine. We might want to reconsider the default though if it would now also matter for SSD windows (it is currently `always`).

salahoued:
If I understand correctly removing rounded corners when a window is tiled/snapped to a region is an intended behaviour in `0.7.3` and when not the client/window should keep its rounded corners like so:

### Not tiled:
![20240719_19h02m20s_grim](https://github.com/user-attachments/assets/738cddac-711f-4078-a524-3d7b39970d29)

### Tiled:
![20240719_19h02m43s_grim](https://github.com/user-attachments/assets/616c1269-433a-4d69-8040-95edbd306270)

### But I get this when a window is tiled and I run a `Reconfigure`:
![20240719_19h03m27s_grim](https://github.com/user-attachments/assets/0e5a9c14-747d-4a66-ab7f-007511fe17f2)

Consolatis:
Can confirm, SSD tiled + Reconfigure is broken.

PR #1926, opened 2024-06-22
feat: set squared SSD corners when SSD client is tiled and notified about being tiled

jp7677:
My attempt to solve https://github.com/labwc/labwc/issues/1862 ~Still draft to let this land slowly and give it some time for testing.~

With this PR, my SSD corners look and behave the same as CSD GTK corners which gives a beautiful consistent look.   
A side effect is unfortunately that the "rounded hover effect on square  buttons" can be seen (much) more often.

Closes https://github.com/labwc/labwc/issues/1862

### DIFF
--- a/src/ssd/ssd-border.c
+++ b/src/ssd/ssd-border.c
@@ -86,6 +86,38 @@ ssd_border_update(struct ssd *ssd)
 	int height = view_effective_height(view, /* use_pending */ false);
 	int full_width = width + 2 * theme->border_width;
 
+	/*
+	 * From here on we have to cover the following border scenarios:
+	 * Non-tiled (partial border, rounded corners):
+	 *    _____________
+	 *   o           oox
+	 *  |---------------|
+	 *  |_______________|
+	 *
+	 * Tiled (full border, squared corners):
+	 *   _______________
+	 *  |o           oox|
+	 *  |---------------|
+	 *  |_______________|
+	 *
+	 * Tiled or non-tiled with zero title height (full boarder, no title):
+	 *   _______________
+	 *  |_______________|
+	 */
+
+	int side_height = ssd->state.was_tiled_not_maximized
+		? height + ssd->titlebar.height
+		: height;
+	int side_y = ssd->state.was_tiled_not_maximized
+		? -ssd->titlebar.height
+		: 0;
+	int top_width = ssd->titlebar.height <= 0 || ssd->state.was_tiled_not_maximized
+		? full_width
+		: width - 2 * SSD_BUTTON_WIDTH;
+	int top_x = ssd->titlebar.height <= 0 || ssd->state.was_tiled_not_maximized
+		? 0
+		: theme->border_width + SSD_BUTTON_WIDTH;
+
 	struct ssd_part *part;
 	struct wlr_scene_rect *rect;
 	struct ssd_sub_tree *subtree;
@@ -95,34 +127,35 @@ ssd_border_update(struct ssd *ssd)
 			switch (part->type) {
 			case LAB_SSD_PART_LEFT:
 				wlr_scene_rect_set_size(rect,
-					theme->border_width, height);
+					theme->border_width,
+					side_height);
+				wlr_scene_node_set_position(part->node,
+					0,
+					side_y);
 				continue;
 			case LAB_SSD_PART_RIGHT:
 				wlr_scene_rect_set_size(rect,
-					theme->border_width, height);
+					theme->border_width,
+					side_height);
 				wlr_scene_node_set_position(part->node,
-					theme->border_width + width, 0);
+					theme->border_width + width,
+					side_y);
 				continue;
 			case LAB_SSD_PART_BOTTOM:
 				wlr_scene_rect_set_size(rect,
-					full_width, theme->border_width);
+					full_width,
+					theme->border_width);
 				wlr_scene_node_set_position(part->node,
-					0, height);
+					0,
+					height);
 				continue;
 			case LAB_SSD_PART_TOP:
-				if (ssd->titlebar.height > 0) {
-					wlr_scene_rect_set_size(rect,
-						width - 2 * SSD_BUTTON_WIDTH,
-						theme->border_width);
-					wlr_scene_node_set_position(part->node,
-						theme->border_width + SSD_BUTTON_WIDTH,
-						-(ssd->titlebar.height + theme->border_width));
-				} else {
-					wlr_scene_rect_set_size(rect,
-						full_width, theme->border_width);
-					wlr_scene_node_set_position(part->node,
-						0, -theme->border_width);
-				}
+				wlr_scene_rect_set_size(rect,
+					top_width,
+					theme->border_width);
+				wlr_scene_node_set_position(part->node,
+					top_x,
+					-(ssd->titlebar.height + theme->border_width));
 				continue;
 			default:
 				continue;
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -117,10 +117,15 @@ ssd_titlebar_create(struct ssd *ssd)
 
 	ssd_update_title(ssd);
 
-	if (view->maximized == VIEW_AXIS_BOTH) {
+	bool maximized = view->maximized == VIEW_AXIS_BOTH;
+	if (maximized) {
 		set_squared_corners(ssd, true);
 		set_maximize_alt_icon(ssd, true);
 		ssd->state.was_maximized = true;
+	}
+	if (view_is_tiled(view) && !maximized) {
+		set_squared_corners(ssd, true);
+		ssd->state.was_tiled_not_maximized = true;
 	}
 }
 
@@ -187,11 +192,16 @@ ssd_titlebar_update(struct ssd *ssd)
 	int width = view->current.width;
 	struct theme *theme = view->server->theme;
 
-	bool maximized = (view->maximized == VIEW_AXIS_BOTH);
-	if (ssd->state.was_maximized != maximized) {
-		set_squared_corners(ssd, maximized);
-		set_maximize_alt_icon(ssd, maximized);
+	bool maximized = view->maximized == VIEW_AXIS_BOTH;
+	bool tiled_not_maximized = view_is_tiled(ssd->view) && !maximized;
+	if (ssd->state.was_maximized != maximized
+			|| ssd->state.was_tiled_not_maximized != tiled_not_maximized) {
+		set_squared_corners(ssd, maximized || tiled_not_maximized);
+		if (ssd->state.was_maximized != maximized) {
+			set_maximize_alt_icon(ssd, maximized);
+		}
 		ssd->state.was_maximized = maximized;
+		ssd->state.was_tiled_not_maximized = tiled_not_maximized;
 	}
 
 	if (width == ssd->state.geometry.width) {
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -123,7 +123,7 @@ ssd_titlebar_create(struct ssd *ssd)
 		set_maximize_alt_icon(ssd, true);
 		ssd->state.was_maximized = true;
 	}
-	if (view_is_tiled(view) && !maximized) {
+	if (view_is_tiled_and_notify_tiled(view) && !maximized) {
 		set_squared_corners(ssd, true);
 		ssd->state.was_tiled_not_maximized = true;
 	}
@@ -193,7 +193,9 @@ ssd_titlebar_update(struct ssd *ssd)
 	struct theme *theme = view->server->theme;
 
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
-	bool tiled_not_maximized = view_is_tiled(ssd->view) && !maximized;
+	bool tiled_not_maximized = view_is_tiled_and_notify_tiled(ssd->view)
+		&& !maximized;
+
 	if (ssd->state.was_maximized != maximized
 			|| ssd->state.was_tiled_not_maximized != tiled_not_maximized) {
 		set_squared_corners(ssd, maximized || tiled_not_maximized);
--- a/src/view.c
+++ b/src/view.c
@@ -1172,6 +1172,23 @@ view_is_tiled(struct view *view)
 }
 
 bool
+view_is_tiled_and_notify_tiled(struct view *view)
+{
+	switch (rc.snap_tiling_events_mode) {
+	case LAB_TILING_EVENTS_NEVER:
+		return false;
+	case LAB_TILING_EVENTS_REGION:
+		return view->tiled_region || view->tiled_region_evacuate;
+	case LAB_TILING_EVENTS_EDGE:
+		return view->tiled;
+	case LAB_TILING_EVENTS_ALWAYS:
+		return view_is_tiled(view);
+	}
+
+	return false;
+}
+
+bool
 view_is_floating(struct view *view)
 {
 	assert(view);
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -50,6 +50,7 @@ struct ssd {
 	 */
 	struct {
 		bool was_maximized;   /* To un-round corner buttons and toggle icon on maximize */
+		bool was_tiled_not_maximized;   /* To un-round corner buttons */
 		struct wlr_box geometry;
 		struct ssd_state_title {
 			char *text;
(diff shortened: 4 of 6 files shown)


src/ssd/ssd-titlebar.c
```diff
@@ -444,4 +455,18 @@ ssd_update_button_hover(struct wlr_scene_node *node,
 	}
 }
 
+static bool
+view_is_tiled_and_notify_tiled(struct view *view)
```
jp7677, line 459: Should this go into `view.h`?

src/ssd/ssd-part.c
```diff
@@ -142,18 +143,31 @@ get_scale_box(struct wlr_buffer *buffer, double container_width,
...
+		offset_x = 0;
+	} else {
+		assert(false && "invalid corner button type");
+		wlr_log(WLR_ERROR, "invalid corner button type");
+		abort();
+	}
+
 	/* Background */
 	struct ssd_part *bg_rect = add_scene_rect(part_list, type, parent,
-		SSD_BUTTON_WIDTH, rc.theme->title_height, 0, 0, bg_color);
```
jp7677, line 156: mmh, just resizing this part to include the borders isn't enough. Looks fine when the view is activated, but the boarder color is not correct when the view is not activated. That is with my theme.

src/ssd/ssd-titlebar.c
johanmalm: Yes, I think that would feel better.

src/ssd/ssd-part.c
johanmalm: Is it because the button are positioned at intervals of `SSD_BUTTON_WIDTH`? At a glance, I'm guessing that if you make the buffer wider it will probably overlap adjacent buttons.

https://github.com/labwc/labwc/blob/1495708095b998f03aa8f07a6796764981e2dcd6/src/ssd/ssd-titlebar.c#L203-L228

jp7677: Thanks for the hint. I don't (yet) understand exactly what you mean, but mostly because all SSD painting is still magic to me ;). 

Here is a screenshot that shows what I mean:

screenshot

These are both SSD borders, both tiled and the left view (gnome-terminal) has the focus. The corner from gnome-terminal (left) looks fine, but the button on the CLion SSD doesn't shows the thin border line. I don't know yet where the thin boarder comes from. I do think though, that by making the button background bigger, the buttons are now on top of it and my theme colors are masking this when the view is activated.

jp7677: After discussion on chat, this should be good now and it's also much cleaner code-wise.

jp7677:
> A side effect is unfortunately that the "rounded hover effect on square buttons" can be seen (much) more often.

A feasible mitigation would be imho to change the hover effect into a circle which just covers the icon, like the hover effect in firefox CSD. That way the same hover effect would look nice on both squared and rounded corner buttons.

src/ssd/ssd-border.c
```diff
@@ -86,6 +86,19 @@ ssd_border_update(struct ssd *ssd)
 	int height = view_effective_height(view, /* use_pending */ false);
 	int full_width = width + 2 * theme->border_width;
 
+	int side_height = ssd->state.was_tiled_not_maximized
```
jp7677, line 89: TODO: Needs comment.

src/view.c
```diff
@@ -1171,6 +1171,18 @@ view_is_tiled(struct view *view)
...
+view_is_tiled_and_notify_tiled(struct view *view)
+{
+	return
+		((view_is_tiled(view)
+			&& rc.snap_tiling_events_mode == LAB_TILING_EVENTS_ALWAYS)
+		|| (view->tiled
+			&& rc.snap_tiling_events_mode == LAB_TILING_EVENTS_EDGE)
+		|| ((view->tiled_region || view->tiled_region_evacuate)
+			&& rc.snap_tiling_events_mode == LAB_TILING_EVENTS_REGION));
+}
```
Consolatis, line 1184: Just a personal nitpick, I think this is a bit easier to read:
```c
bool
view_is_tiled_and_notify_tiled(struct view *view)
{
	switch (rc.snap_tiling_events_mode) {
	case LAB_TILING_EVENTS_NEVER:
		return false;
	case LAB_TILING_EVENTS_REGION:
		return view->tiled_region || view->tiled_region_evacuate;
	case LAB_TILING_EVENTS_EDGE:
		return view->tiled;
	case LAB_TILING_EVENTS_ALWAYS:
		return view_is_tiled(view);
	}
}
```

It will also complain loudly if we ever end up adding new tiling event states to the enum.

tokyo4j: How about creating a function like `view_notified_tiled_edges()` which returns the parameter for `wlr_xdg_toplevel_set_tiled()` and share it with `xdg_toplevel_view_notify_tiled()`.

src/ssd/ssd-border.c
jp7677: My best ascii artwork ever ;)

src/view.c
jp7677: > It will also complain loudly if we ever end up adding new tiling event states to the enum.

Thanks, looks indeed easier. That said, the compiler unfortunately still wanted a default case ...

```
../src/view.c:1187:1: warning: control reaches end of non-void function [-Wreturn-type]
 1187 | }
```

jp7677: > How about creating a function like `view_notified_tiled_edges()` which returns the parameter for `wlr_xdg_toplevel_set_tiled()` and share it with `xdg_toplevel_view_notify_tiled()`?

Sorry, I don't understand your suggestion. `xdg_toplevel_view_notify_tiled` is about notifying the client, on the other hand  `view_is_tiled_and_notify_tiled` is about the client being tiled and the client was notified that it is tiled. May be `view_is_tiled_and_notified_tiled` is a better name to avoid this confusion?

jp7677:
> A side effect is unfortunately that the "rounded hover effect on square buttons" can be seen (much) more often.

Running this now for a few days, the wrong hover effect is actually much more in your face. I think we should find a solution for that as well before 0.7.3 when merging this soon, or postpone to have more time. I still think just using circles like firefox does as hover effect is the easiest and imho also a pretty looking solution. That said I'm quite certain that I wont have a PR ready for this until the 27th.

johanmalm:
> Running this now for a few days, the wrong hover effect is actually much more in your face. I think we should find a solution for that as well before 0.7.3 when merging this soon, or postpone to have more time. I still think just using circles like firefox does as hover effect is the easiest and imho also a pretty looking solution. That said I'm quite certain that I wont have a PR ready for this until the 27th.

I was quite tempted to go for circles last time. My reservation was that with large icons (small padding) they may go outside the circle unless we "post-process" and force a certain size padding which gets a bit messy.

jp7677:
> > Running this now for a few days, the wrong hover effect is actually much more in your face. I think we should find a solution for that as well before 0.7.3 when merging this soon, or postpone to have more time. I still think just using circles like firefox does as hover effect is the easiest and imho also a pretty looking solution. That said I'm quite certain that I wont have a PR ready for this until the 27th.
> 
> I was quite tempted to go for circles last time. My reservation was that with large icons (small padding) they may go outside the circle unless we "post-process" and force a certain size padding which gets a bit messy.

Having icons larger than a hover circle would only be the case when custom button icons/images are used, correct? That scenario makes it indeed more difficult. Only simple solution I might have would be to document the hover restrictions and leave it to the user to respect those limits. But not sure if that is feasible/desired.

src/view.c
Consolatis: > > It will also complain loudly if we ever end up adding new tiling event states to the enum.
> 
> Thanks, looks indeed easier. That said, the compiler unfortunately still wanted a default case ...
> 
> ```
> ../src/view.c:1187:1: warning: control reaches end of non-void function [-Wreturn-type]
>  1187 | }
> ```

Compilers can be a bit stupid at times. Just adding a `return false;` after the switch fixes that for me. So there is no `default` case required. I also commented out some enum handler to verify that it still works as intended and got
`../src/view.c:1177:2: error: enumeration value ‘LAB_TILING_EVENTS_ALWAYS’ not handled in switch [-Werror=switch]`.

Consolatis:
An alternate approach would be to add yet another hover icon to the theme and button hover tree: the squared non-toggled hover icon and then add something like `set_maximize_alt_icon()` when switching between square and rounded corners.

src/view.c
tokyo4j: > > How about creating a function like `view_notified_tiled_edges()` which returns the parameter for `wlr_xdg_toplevel_set_tiled()` and share it with `xdg_toplevel_view_notify_tiled()`?
> 
> Sorry, I don't understand your suggestion. `xdg_toplevel_view_notify_tiled` is about notifying the client, on the other hand `view_is_tiled_and_notify_tiled` is about the client being tiled and the client was notified that it is tiled. May be `view_is_tiled_and_notified_tiled` is a better name to avoid this confusion?

I meant something like this:

```diff
diff --git a/include/view.h b/include/view.h
index b4f90e6a..2bb5fe18 100644
--- a/include/view.h
+++ b/include/view.h
@@ -495,7 +495,7 @@ void view_toggle_always_on_bottom(struct view *view);
 void view_toggle_visible_on_all_workspaces(struct view *view);
 
 bool view_is_tiled(struct view *view);
-bool view_is_tiled_and_notify_tiled(struct view *view);
+uint32_t view_notified_edges(struct view *view);
 bool view_is_floating(struct view *view);
 void view_move_to_workspace(struct view *view, struct workspace *workspace);
 enum ssd_mode view_get_ssd_mode(struct view *view);
diff --git a/src/ssd/ssd-titlebar.c b/src/ssd/ssd-titlebar.c
index 32d6131a..ac771af0 100644
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -123,7 +123,7 @@ ssd_titlebar_create(struct ssd *ssd)
 		set_maximize_alt_icon(ssd, true);
 		ssd->state.was_maximized = true;
 	}
-	if (view_is_tiled_and_notify_tiled(view) && !maximized) {
+	if (view_notified_edges(view) && !maximized) {
 		set_squared_corners(ssd, true);
 		ssd->state.was_tiled_not_maximized = true;
 	}
@@ -193,8 +193,7 @@ ssd_titlebar_update(struct ssd *ssd)
 	struct theme *theme = view->server->theme;
 
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
-	bool tiled_not_maximized = view_is_tiled_and_notify_tiled(ssd->view)
-		&& !maximized;
+	bool tiled_not_maximized = view_notified_edges(ssd->view) && !maximized;
 
 	if (ssd->state.was_maximized != maximized
 			|| ssd->state.was_tiled_not_maximized != tiled_not_maximized) {
diff --git a/src/view.c b/src/view.c
index eff9b1c0..42f5c632 100644
--- a/src/view.c
+++ b/src/view.c
@@ -1171,16 +1171,49 @@ view_is_tiled(struct view *view)
 		|| view->tiled_region_evacuate);
 }
 
-bool
-view_is_tiled_and_notify_tiled(struct view *view)
-{
-	return
-		((view_is_tiled(view)
-			&& rc.snap_tiling_events_mode == LAB_TILING_EVENTS_ALWAYS)
-		|| (view->tiled
-			&& rc.snap_tiling_events_mode == LAB_TILING_EVENTS_EDGE)
-		|| ((view->tiled_region || view->tiled_region_evacuate)
-			&& rc.snap_tiling_events_mode == LAB_TILING_EVENTS_REGION));
+uint32_t
+view_notified_edges(struct view *view)
+{
+	/* Take no action if xdg-shell tiling is disabled */
+	if (rc.snap_tiling_events_mode == LAB_TILING_EVENTS_NEVER) {
+		return 0;
+	}
+
+	enum wlr_edges edge = WLR_EDGE_NONE;
+
+	bool want_edge = rc.snap_tiling_events_mode & LAB_TILING_EVENTS_EDGE;
+	bool want_region = rc.snap_tiling_events_mode & LAB_TILING_EVENTS_REGION;
+
+	/*
+	 * Edge-snapped view are considered tiled on the snapped edge and those
+	 * perpendicular to it.
+	 */
+	if (want_edge) {
+		switch (view->tiled) {
+		case VIEW_EDGE_LEFT:
+			edge = WLR_EDGE_LEFT | WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
+			break;
+		case VIEW_EDGE_RIGHT:
+			edge = WLR_EDGE_RIGHT | WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
+			break;
+		case VIEW_EDGE_UP:
+			edge = WLR_EDGE_TOP | WLR_EDGE_LEFT | WLR_EDGE_RIGHT;
+			break;
+		case VIEW_EDGE_DOWN:
+			edge = WLR_EDGE_BOTTOM | WLR_EDGE_LEFT | WLR_EDGE_RIGHT;
+			break;
+		default:
+			edge = WLR_EDGE_NONE;
+		}
+	}
+
+	if (want_region && view->tiled_region) {
+		/* Region-snapped views are considered tiled on all edges */
+		edge = WLR_EDGE_LEFT | WLR_EDGE_RIGHT |
+			WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
+	}
+
+	return edge;
 }
 
 bool
diff --git a/src/xdg.c b/src/xdg.c
index 2b8bf96e..203a4f42 100644
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -422,46 +422,10 @@ xdg_toplevel_view_set_fullscreen(struct view *view, bool fullscreen)
 static void
 xdg_toplevel_view_notify_tiled(struct view *view)
 {
-	/* Take no action if xdg-shell tiling is disabled */
-	if (rc.snap_tiling_events_mode == LAB_TILING_EVENTS_NEVER) {
-		return;
+	uint32_t edge = view_notified_edges(view);
+	if (edge) {
+		wlr_xdg_toplevel_set_tiled(xdg_toplevel_from_view(view), edge);
 	}
-
-	enum wlr_edges edge = WLR_EDGE_NONE;
-
-	bool want_edge = rc.snap_tiling_events_mode & LAB_TILING_EVENTS_EDGE;
-	bool want_region = rc.snap_tiling_events_mode & LAB_TILING_EVENTS_REGION;
-
-	/*
-	 * Edge-snapped view are considered tiled on the snapped edge and those
-	 * perpendicular to it.
-	 */
-	if (want_edge) {
-		switch (view->tiled) {
-		case VIEW_EDGE_LEFT:
-			edge = WLR_EDGE_LEFT | WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
-			break;
-		case VIEW_EDGE_RIGHT:
-			edge = WLR_EDGE_RIGHT | WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
-			break;
-		case VIEW_EDGE_UP:
-			edge = WLR_EDGE_TOP | WLR_EDGE_LEFT | WLR_EDGE_RIGHT;
-			break;
-		case VIEW_EDGE_DOWN:
-			edge = WLR_EDGE_BOTTOM | WLR_EDGE_LEFT | WLR_EDGE_RIGHT;
-			break;
-		default:
-			edge = WLR_EDGE_NONE;
-		}
-	}
-
-	if (want_region && view->tiled_region) {
-		/* Region-snapped views are considered tiled on all edges */
-		edge = WLR_EDGE_LEFT | WLR_EDGE_RIGHT |
-			WLR_EDGE_TOP | WLR_EDGE_BOTTOM;
-	}
-
-	wlr_xdg_toplevel_set_tiled(xdg_toplevel_from_view(view), edge);
 }
 
 static struct view *

```

</details>

jp7677: > Compilers can be a bit stupid at times. Just adding a `return false;` after the switch fixes that for me. So there is no `default` case required. I also commented out some enum handler to verify that it still works as intended and got `../src/view.c:1177:2: error: enumeration value ‘LAB_TILING_EVENTS_ALWAYS’ not handled in switch [-Werror=switch]`.

ah of course. Adjusted and should be good now.

jp7677: > I meant something like this:

...snip...

> Not a strong preference though.

Thanks, I see. Looks also good and as as far as I see (based on the diff, not tested), the result should be indeed the same. Which version should I go for?

jp7677:
Any thoughts on how to proceed code wise and timing wise with this PR? I'm still in doubt if we should depend this PR to also getting the hover effects correct. I mean it is already incorrect on master when a view is maximized.  

I can try to look at the squared hovers during the weekend based on @ Consolatis  hints, but cannot promise that I get this wired correctly (succeeding with this is probably more wishful thinking than anything else on my side).

johanmalm:
I'm happy that we proceed without the hover-icon fix. I've only glanced at this PR though. Do you want me to review with a view to merge?

jp7677:
> I'm happy that we proceed without the hover-icon fix. I've only glanced at this PR though. Do you want me to review with a view to merge?

Yes, that would be nice. Ready to review from my side.

johanmalm:
LGTM
In fact, it looks very tidy indeed. Nice work.

PR #1931, opened 2024-06-23
[wip]

johanmalm:
RFC and to promote discussion on #1919 

Cc: @droc12345


droc12345:
Almost works, arrow key (down) is wrong when doing PreviousWindow,
NextWindow works both directions.

Which is why I had an int set to -1, and will not apply arrow keys logic if the int is < 0

PR #1935, opened 2024-06-25
Remove fuzzy for capitalized translations

stefonarch:


tokyo4j:
Thank you.

I wonder if it is Weblate's bug that it didn't `#, fuzzy`.

In 9fc41a9730fcdeeefd69de6a9f35774e30f0ce26, `#, fuzzy` is removed when `PO-Revision-Date` is updated, but otherwise it isn't. That's strange.

stefonarch:
>In https://github.com/labwc/labwc/commit/9fc41a9730fcdeeefd69de6a9f35774e30f0ce26, #, fuzzy is removed when PO-Revision-Date is updated, but otherwise it isn't. That's strange.

that's because me and the estonian translator updated it, de, and et translations

Consolatis:
Thanks!

PR #1936, opened 2024-06-25
dnd: show dnd icon above layer-shell surfaces

### LINKED ISSUE #1933: Dnd icon is rendered below layer-shell surface
Labwc 0.7.2/git master, it's a regression introduced in 0.7.0 (it worked with 0.6.6)

Use case: xfce4-panel with tasklist plugin, try to drag and drop a window button from the tasklist (so you must have a terminal window open, or something). The dnd icon doesn't appear, because it's actually rendered under the panel, as can be seen by moving the pointer out of the panel.

tokyo4j:
Fixes #1933.


Consolatis:
Thanks!

PR #1941, opened 2024-06-29
feat: add tearing fullscreen options

jp7677:
This is a slightly different approach to https://github.com/labwc/labwc/pull/1568

Whereas https://github.com/labwc/labwc/pull/1568 (forcefully) enables tearing when the active window is full screen and still allows tearing for non-fullscreen windows when the window hints tearing, this PR interprets  the 'fullscreen' option  as a restriction on top of the existing allow tearing ('yes') option. So 'fullscreen' here means "allow tearing *only* in fullscreen state". Additionally  the toggle action overrides the application preference. So once you started toggling, the application preference is ignored. I hope I described this clearly in the docs as part of this PR. 

I'm using the gsync demo from https://github.com/labwc/labwc/pull/1568#issuecomment-1986931068 for testing. Starting this via wine, the tearing hint is set repeatedly by the application. I found it slightly confusing that the toggle action didn't allowed me to disable tearing (on master). Technically the toggle did disable tearing, but it got immediately overridden again by the window, so it looked liked the toggle had no effect. Hence the changed behavior to ignore the hint when toggle to tearing disabled.

What the 'fullscreen' option should do is a matter of preference.  The idea from https://github.com/labwc/labwc/pull/1568 where 'fullscreen' ignores the tearing hint, is still valid and may be this option should be available as well. Also not sure which interpretation makes the most sense. May be it depends on how wisely applications set the tearing hint. To my surprise, the tearing hint was set with wine and demo above (unless I screwed up and something is wrong with the tearing hint now).  

Open for thoughts and feedback.

cc: @Ph42oN @Faugus

PS: Most of this PR is still based on @Consolatis work, hence the Co-Authored in the commit message.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -185,9 +185,18 @@ this is for compatibility with Openbox.
 	*fullscreen* enables adaptive sync whenever a window is in fullscreen
 	mode.
 
-*<core><allowTearing>* [yes|no]
-	Allow tearing, if requested by the active window, to reduce input lag.
-	Default is no.
+*<core><allowTearing>* [yes|no|fullscreen|fullscreenForced]
+	Allow tearing to reduce input lag. Default is no.
+
+	*yes* allows tearing if requested by the active window.
+
+	*fullscreen* allows tearing if requested by the active window, but
+	only when the window is in fullscreen mode.
+
+	*fullscreenForced* enables tearing whenever the active window is in
+	fullscreen mode, whether or not the application has requested tearing.
+
+	Use the *ToggleTearing* action for forcefully enable tearing.
 
 	Note: Enabling this option with atomic mode setting is experimental. If
 	you experience undesirable side effects when tearing is allowed,
--- a/src/action.c
+++ b/src/action.c
@@ -1109,9 +1109,22 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_TOGGLE_TEARING:
 			if (view) {
-				view->tearing_hint = !view->tearing_hint;
-				wlr_log(WLR_DEBUG, "tearing %sabled",
-					view->tearing_hint ? "en" : "dis");
+				switch (view->force_tearing) {
+				case LAB_STATE_UNSPECIFIED:
+					view->force_tearing =
+						output_get_tearing_allowance(view->output)
+							? LAB_STATE_DISABLED : LAB_STATE_ENABLED;
+					break;
+				case LAB_STATE_DISABLED:
+					view->force_tearing = LAB_STATE_ENABLED;
+					break;
+				case LAB_STATE_ENABLED:
+					view->force_tearing = LAB_STATE_DISABLED;
+					break;
+				}
+				wlr_log(WLR_ERROR, "force tearing %sabled",
+					view->force_tearing == LAB_STATE_ENABLED
+						? "en" : "dis");
 			}
 			break;
 		case ACTION_TYPE_TOGGLE_SHADE:
--- a/src/output.c
+++ b/src/output.c
@@ -38,28 +38,50 @@ get_tearing_retry_count(struct output *output)
 	return refresh > 0 ? refresh / 500 : 120;
 }
 
-static bool
-get_tearing_preference(struct output *output)
+bool
+output_get_tearing_allowance(struct output *output)
 {
 	struct server *server = output->server;
 
-	/* Never allow tearing when disabled */
+	/* never allow tearing when disabled */
 	if (!rc.allow_tearing) {
 		return false;
 	}
 
-	/* Tearing is only allowed for the output with the active view */
-	if (!server->active_view || server->active_view->output != output) {
+	struct view *view = server->active_view;
+
+	/* tearing is only allowed for the output with the active view */
+	if (!view || view->output != output) {
 		return false;
 	}
 
-	/* Tearing should not have failed too many times */
+	/* tearing should not have failed too many times */
 	if (output->nr_tearing_failures >= get_tearing_retry_count(output)) {
 		return false;
 	}
 
-	/* If the active view requests tearing, or it is toggled on with action, allow it */
-	return server->active_view->tearing_hint;
+	/* allow tearing for any window when requested or forced */
+	if (rc.allow_tearing == LAB_TEARING_ENABLED) {
+		if (view->force_tearing == LAB_STATE_UNSPECIFIED) {
+			return view->tearing_hint;
+		} else {
+			return view->force_tearing == LAB_STATE_ENABLED;
+		}
+	}
+
+	/* remaining tearing options apply only to full-screen windows */
+	if (!view->fullscreen) {
+		return false;
+	}
+
+	if (view->force_tearing == LAB_STATE_UNSPECIFIED) {
+		/* honor the tearing hint or the fullscreen-force preference */
+		return view->tearing_hint ||
+			rc.allow_tearing == LAB_TEARING_FULLSCREEN_FORCED;
+	}
+
+	/* honor tearing as requested by action */
+	return view->force_tearing == LAB_STATE_ENABLED;
 }
 
 static void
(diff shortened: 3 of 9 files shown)


Faugus:
@jp7677 I just tested your patch and looks like everything is working as intended.

Great work and thank you!

Consolatis:
> What the 'fullscreen' option should do is a matter of preference. The idea from #1568 where 'fullscreen' ignores the tearing hint, is still valid and may be this option should be available as well.

I think we should not have a `fullscreen` option at all then. But instead something like `fullscreenOnly` for this PR and potentially something like `enforceOnFullscreen` for the alternate case.

jp7677:
> > What the 'fullscreen' option should do is a matter of preference. The idea from #1568 where 'fullscreen' ignores the tearing hint, is still valid and may be this option should be available as well.
> 
> I think we should not have a `fullscreen` option at all then. But instead something like `fullscreenOnly` for this PR and potentially something like `enforceOnFullscreen` for the alternate case.

Yes agree, I had also something like this mind. Will sleep a night over it and use your suggestions if I don’t find better fitting terms.

jp7677:
> I think we should not have a `fullscreen` option at all then. But instead something like `fullscreenOnly` for this PR and potentially something like `enforceOnFullscreen` for the alternate case.

I kept `fullscreen` since this matches the meaning of `fullscreen` for the `adaptiveSync` option. I've also added the always-enable-tearing-in-fullscreen-mode and named it `fullscreenUnconditionally`.  I like the ring of it, but happy to change that ofc.

jp7677:
> > I think we should not have a `fullscreen` option at all then. But instead something like `fullscreenOnly` for this PR and potentially something like `enforceOnFullscreen` for the alternate case.
> 
> I kept `fullscreen` since this matches the meaning of `fullscreen` for the `adaptiveSync` option from my understanding. I've also added the always-enable-tearing-in-fullscreen-mode and named it `fullscreenUnconditionally`. I like the ring of it, but happy to change that ofc (`fullscreenForced` also fits imho and has the advantage of being shorter).

I reconsidered and went for `fullscreen` and `fullscreenForced`. Shorter and I guess more "common".

jp7677:
PS: This is ready from my side. But ofc perfectly cool to wait until the release is tagged.

johanmalm:
> PS: This is ready from my side. But ofc perfectly cool to wait until the release is tagged.

Thanks, that's great. Yes, let's wait until after release.

johanmalm:
This has now got a merge conflict.

In principle this LGTM, but I can't test and haven't followed the tearing discussions sufficiently to make a sensible judgement on whether or not the logic + exposed config options align with where we are going.

@ahesford  - would you be happy to review and merge if content (after conflict gets resolved)
@kode54 - appreciate your thoughts on this too.

docs/labwc-config.5.scd
```diff
@@ -181,11 +181,20 @@ this is for compatibility with Openbox.
...
 	WLR_DRM_NO_ATOMIC=1.
-	*yes* allow tearing if requested by the active window.
+
+	*yes* allows tearing if requested by the active window.
+
+	*fullscreen* allows tearing if requested by the active window and
+	restricts this to fullscreen mode only.
+
+	*fullscreenForced* enables tearing if the active window is in
+	fullscreen mode, regardless if the application hinted tearing or not.
```
ahesford, line 195:
```suggestion
	*fullscreen* allows tearing if requested by the active window, but
	only when the window is in fullscreen mode.

	*fullscreenForced* enables tearing whenever the active window is in
	fullscreen mode, whether or not the application has requested tearing.
```

src/action.c
```diff
@@ -1109,9 +1109,18 @@ actions_run(struct view *activator, struct server *server,
 			break;
 		case ACTION_TYPE_TOGGLE_TEARING:
 			if (view) {
-				view->tearing_hint = !view->tearing_hint;
-				wlr_log(WLR_DEBUG, "tearing %sabled",
-					view->tearing_hint ? "en" : "dis");
+				switch (view->force_tearing) {
+				case LAB_STATE_UNSPECIFIED:
```
ahesford, line 1113: I think this behavior would be more pleasant if the UNSPECIFIED behavior did

    view->force_tearing = !get_tearing_preference(...)

where `get_tearing_public` from `output.c` will have to be made public and renamed accordingly (*e.g.*, `output_get_tearing_allowance`). This truly toggles even from an indeterminate state.

src/output.c
```diff
@@ -34,18 +34,41 @@ get_tearing_preference(struct output *output)
...
+	/* allow tearing in fullscreen or windows mode */
+	if (rc.allow_tearing == LAB_TEARING_ENABLED) {
+		if (view->force_tearing == LAB_STATE_UNSPECIFIED) {
+			return view->tearing_hint;
+		} else {
+			return view->force_tearing == LAB_STATE_ENABLED;
+		}
+	}
+
+	return false;
```
ahesford, line 71: Rearranging the conditionals and short-circuiting improves clarity, I think:
```suggestion
	/* allow tearing for any window when requested or forced */
	if (rc.allow_tearing == LAB_TEARING_ENABLED) {
		if (view->force_tearing == LAB_STATE_UNSPECIFIED) {
			return view->tearing_hint;
		} else {
			return view->force_tearing == LAB_STATE_ENABLED;
		}
	}
	
	/* remaining tearing options apply only to full-screen windows */
	if (!view->fullscreen) {
		return false;
	}

	if (view->force_tearing == LAB_STATE_UNSPECIFIED) {
		/* honor the tearing hint or the fullscreen-force preference */
		return view->tearing_hint ||
			rc.allow_tearing == LAB_TEARING_FULLSCREEN_FORCED;
	} else {
		/* honor tearing as requested by action */
		return view->force_tearing == LAB_STATE_ENABLED;
	}

	return false;
```

ahesford (changes requested):
The concept and implementation are generally good, but I suggest a few refinements.

I'm ambivalent about the `fullscreenForced` value. Another possibility would be splitting another core Boolean option, *e.g.*, `forceTearing`, that could apply when `allowTearing` is either `yes` or `fullscreen`. (When both options are `yes`, tearing would always be active on the output that contains the active window.) I'm not convinced this is better, but might be.

Ph42oN:
I think this looks good and will propably be able to give wanted behavior with tearing for everyone. I will test it soon.

Another thing regarding tearing is that we should do something different than checking WLR_DRM_NO_ATOMIC to find if tearing is supported. Thats just how i did it because i thought it was the simplest way, but now that kernel got support for atomic tearing page flips it will not be right way to do it. I heard that it requires wlroots 0.18, so i think this may be thing for new PR and i will try to work on it some time.

Consolatis:
> Another thing regarding tearing is that we should do something different than checking WLR_DRM_NO_ATOMIC to find if tearing is supported.

Its already merged in the master branch:
- #1996

src/action.c
jp7677: yes, implemented like you suggested and this indeed works nicely.

jp7677:
@ahesford Thanks for the review. I've rebased this PR and addressed your points and  the toggle works now even better, thanks!

jp7677:
> Also, I'm ambivalent about the `fullscreenForced` value. Another possibility would be splitting another core Boolean option, _e.g._, `forceTearing`, that could apply when `allowTearing` is either `yes` or `fullscreen`. (When both options are `yes`, tearing would always be active on the output that contains the active window.) I'm not convinced this is better, but might be.

No strong preference from my side. For now I kept `fullscreenForced` for keeping the configuration simple, since I guess tearing is all about gaming which I guess mostly happens fullscreen only.

johanmalm:
I'm happy with suggested config option.
@ahesford - were you okay to lift the merge-block?
If so, I'll do a final review with a view to merge.

ahesford (approved):

ahesford:
Sure, this looks good to me.

jp7677:
(last forced-push was just a styling nit, removed not needed `else` at the end of the function)

johanmalm:
LGTM. Thanks for the work on this.

Issue #1150, opened 2023-10-09
Pipemenu and submenuShowDelay


johanmalm:
Implement cat B+ items `7.1.4` and `2.11.3`

https://github.com/labwc/labwc-scope/blob/main/incomplete-cat-B-items.md

- [ ] Split `parse_xml()` to take fp from `fopen()` or `popen()`
- [ ] Replace bespoke array with `wl_list` to make iteration easier when splitting `menu.c`
- [ ] Split `menu.c`
- [ ] Move `handle_menu_keys()` to `menu/input.c`
- [ ] Refactor `menu_configure()` to be called in real time rather just once on initialization (which is fine for a static menu, but not so good for a dynamic one).
- [ ] Re-consider what `server->menu` variables we want (currently just `menu_current`, but no selection)
- [ ] Should `menu struct` contain size/offset?
- [ ] Implement pipe menus
- [ ] Implement submenu delay

Consolatis:
> Move `handle_menu_keys()` to `menu/input.c`

Do we really want to do that? It uses various `XKB_KEY_` defines and basically just maps keys to actions.
Moving the handlers that are called by that function somewhere else within `src/menu/` sounds totally fine, just not sure about `handle_menu_keys()` itself.

johanmalm:
Okay. I'll split without moving the key-handling thing into it.

Yes, good idea about a private header.

Consolatis:
Marked the pipe menu entry as done. Is there anything you want to see in 0.7.2? Otherwise I'd suggest that we move this to the 0.7.3 milestone.

johanmalm:
Thanks. Nothing else is imminent.  I've moved it to `0.7.3`

PR #1944, opened 2024-06-29
menu: move menu_tree to front

johanmalm:
Also, always top-align submenus intead of inheriting the vertical alignment of their parent. This is consitent with Openbox behaviour.

Related-to: #1150

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -369,6 +369,7 @@ new_output_notify(struct wl_listener *listener, void *data)
 	 * Set the z-positions to achieve the following order (from top to
 	 * bottom):
 	 *	- session lock layer
+	 *	- compositor menu
 	 *	- layer-shell popups
 	 *	- overlay layer
 	 *	- top layer
@@ -381,6 +382,7 @@ new_output_notify(struct wl_listener *listener, void *data)
 	wlr_scene_node_raise_to_top(&output->layer_tree[2]->node);
 	wlr_scene_node_raise_to_top(&output->layer_tree[3]->node);
 	wlr_scene_node_raise_to_top(&output->layer_popup_tree->node);
+	wlr_scene_node_raise_to_top(&server->menu_tree->node);
 	wlr_scene_node_raise_to_top(&output->session_lock_tree->node);
 
 	/*
--- a/src/server.c
+++ b/src/server.c
@@ -423,6 +423,7 @@ server_init(struct server *server)
 	 * | Type              | Scene Tree       | Per Output | Example
 	 * | ----------------- | ---------------- | ---------- | -------
 	 * | ext-session       | lock-screen      | Yes        | swaylock
+	 * | compositor-menu   | menu_tree        | No         | root-menu
 	 * | layer-shell       | layer-popups     | Yes        |
 	 * | layer-shell       | overlay-layer    | Yes        |
 	 * | layer-shell       | top-layer        | Yes        | waybar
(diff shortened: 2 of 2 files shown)


Consolatis:
Tested a bit nested and IMHO it feels a bit weird:

When you open the main menu close to the bottom and then select a submenu and that submenu starting from the selected main menu item doesn't fit without "overshooting" the bottom it will get moved up. That causes it to not be aligned (with either the top or bottom element in the submenu) to the parent item anymore. Could we instead try to open it
- wherever the parent menu was aligned to, when that doesn't work due to the size constrains
- at the inverse position, when that doesn't work due to size constrains either (e.g. now overshoots the other end of the usable area)
- align it to wherever the parent was aligned to + reposition so that it fits in that direction (e.g. what this PR is doing but with following the parent alignment rather than hardcoding it to bottom)

Should we also take the chance to do the same for the x axis? Currently a submenu just overshoots the right / left.

Something else I noticed is that the left / right keys are the wrong way around when the menu opens to the left due to root menu space constrains (e.g. opening it close to the right of the output).

johanmalm:
> Tested a bit nested and IMHO it feels a bit weird: When you open the main menu close to the bottom and then select a submenu and that submenu starting from the selected main menu item doesn't fit without "overshooting" the bottom it will get moved up. That causes it to not be aligned (with either the top or bottom element in the submenu) to the parent item anymore. 

Yes.

Happy to brainstorm other solutions (probably best on IRC), but I think we ought to ponder the adoption of this (PR) behavior. It seems to me consistent with GTK + Openbox (see videos attached) which probably makes it the most intuitive to most users.

[rec-panel.webm](https://github.com/labwc/labwc/assets/6955353/63448e88-0127-4aa6-b4ef-64a1e0f649f1)
[rec-openbox.webm](https://github.com/labwc/labwc/assets/6955353/6724878a-3685-4611-a13c-51d21d88642d)

> Should we also take the chance to do the same for the x axis? Currently a submenu just overshoots the right / left.

Yes. I did notice that it uses the widths of both the current+submenu when calculating when to flip to right-aligned (OPEN_LEFT). Thought I'd just take a small step at the time though considering where we are in terms of cool-down.

> Something completely different I noticed is that the left / right keys are the wrong way around when the menu opens to the left due to root menu space constrains (e.g. opening it close to the right of the output).

Having (just) played with GTK+Openbox, this also seems consistent with those two. Simply speak: right arrow open; left arrow closes - regardless of alignment.

droc12345:
> Simply speak: right arrow open; left arrow closes - regardless of alignment.

It's always been that way for me, especially in firefox where I might have many submenus.
after you hit the right edge of the screen it opens to the left, but still acts like a right open.
That bit me a few times till I learned better.

Consolatis:
Just because others have the same issue it still feels wrong to me. I instinctively press the arrow in the direction where the menu I want to select is and would assume that for most people it would be the same (?).

johanmalm:
Have removed everything related to keeping the menus inside the output as now taken care of by #2408.

The only remaining commit now relates to the z-position of the menu_tree.

Ready for review + merge.

src/output.c
```diff
@@ -382,6 +382,7 @@ new_output_notify(struct wl_listener *listener, void *data)
 	wlr_scene_node_raise_to_top(&output->layer_tree[3]->node);
 	wlr_scene_node_raise_to_top(&output->layer_popup_tree->node);
 	wlr_scene_node_raise_to_top(&output->session_lock_tree->node);
+	wlr_scene_node_raise_to_top(&server->menu_tree->node);
```
Consolatis, line 385: Should the menu really be on top of the lockscreen?
I assume when you got a menu open and then the lockscreen is started due to swayidle it would allow to execute actions from that menu.

johanmalm: Good point. Didn't think of that. Fixed.

tokyo4j:
LGTM. Thanks.

PR #1949, opened 2024-07-01
ssd: allow ssd to be smaller than minimal size

### LINKED ISSUE #1947: Labwc misplaces titlebar on small window
Here's what it looks like:
swappy-20240629_221901

Unfortunately, the titlebar doesn't have a place where a pointer can hold and drag to move the window (but Alt-click-drag still works).

This window is from Workrave.

tokyo4j:
Another approach for #1762.

Note that I don't understand the logic of `ssd_*_update()`. I just fixed pixman errors while testing with [my test client with 1x1 geometry](https://github.com/tokyo4j/wayland-demo/tree/deco-1x1).

Fixes #1947.

### DIFF
--- a/src/ssd/ssd-border.c
+++ b/src/ssd/ssd-border.c
@@ -114,6 +114,7 @@ ssd_border_update(struct ssd *ssd)
 	int top_width = ssd->titlebar.height <= 0 || ssd->state.was_tiled_not_maximized
 		? full_width
 		: width - 2 * SSD_BUTTON_WIDTH;
+	top_width = MAX(0, top_width);
 	int top_x = ssd->titlebar.height <= 0 || ssd->state.was_tiled_not_maximized
 		? 0
 		: theme->border_width + SSD_BUTTON_WIDTH;
--- a/src/ssd/ssd-extents.c
+++ b/src/ssd/ssd-extents.c
@@ -113,6 +113,9 @@ ssd_extents_update(struct ssd *ssd)
 	int side_width = full_width + extended_area * 2 - corner_size * 2;
 	int side_height = full_height + extended_area * 2 - corner_size * 2;
 
+	side_width = MAX(0, side_width);
+	side_height = MAX(0, side_height);
+
 	struct wlr_box part_box;
 	struct wlr_box result_box;
 	struct ssd_part *part;
--- a/src/ssd/ssd-shadow.c
+++ b/src/ssd/ssd-shadow.c
@@ -131,7 +131,7 @@ set_shadow_part_geometry(struct ssd_part *part, int width, int height,
 		y = -titlebar_height + inset;
 		wlr_scene_node_set_position(part->node, x, y);
 		wlr_scene_buffer_set_dest_size(
-			scene_buf, visible_shadow_width, height - 2 * inset);
+			scene_buf, visible_shadow_width, MAX(0, height - 2 * inset));
 		wlr_scene_node_set_enabled(part->node, show_sides);
 		break;
 	case LAB_SSD_PART_BOTTOM:
@@ -139,7 +139,7 @@ set_shadow_part_geometry(struct ssd_part *part, int width, int height,
 		y = -titlebar_height + height;
 		wlr_scene_node_set_position(part->node, x, y);
 		wlr_scene_buffer_set_dest_size(
-			scene_buf, width - 2 * inset, visible_shadow_width);
+			scene_buf, MAX(0, width - 2 * inset), visible_shadow_width);
 		wlr_scene_node_set_enabled(part->node, show_topbottom);
 		break;
 	case LAB_SSD_PART_LEFT:
@@ -147,7 +147,7 @@ set_shadow_part_geometry(struct ssd_part *part, int width, int height,
 		y = -titlebar_height + inset;
 		wlr_scene_node_set_position(part->node, x, y);
 		wlr_scene_buffer_set_dest_size(
-			scene_buf, visible_shadow_width, height - 2 * inset);
+			scene_buf, visible_shadow_width, MAX(0, height - 2 * inset));
 		wlr_scene_node_set_enabled(part->node, show_sides);
 		break;
 	case LAB_SSD_PART_TOP:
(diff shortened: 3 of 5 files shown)


droc12345:
> if (eff_width > 0 && eff_width < LAB_MIN_VIEW_WIDTH) {

I would have just removed the second part and left the check for greater than 0.
Along with changing the message below that.
You don't really want a negative value.

Consolatis:
> if window width - all button width is less than or equal to 0, then don't produce buttons,
> just do the rest of the ssd.

That really doesn't sound right either.
From what I remember, currently the title is shrunk up to width 0. I assume that when the width is smaller than 4 * button_width we could simply let the SSD stick out to the right of the view, otherwise without buttons you can't control the window anymore.

Consolatis:
Thoughts about this alternate version? https://github.com/Consolatis/labwc/tree/fix/min_view_size

droc12345:
> otherwise without buttons you can't control the window anymore.

Indeed, but when the window is very small, you have limited choices.
1. shrink the buttons (maybe a lot)
2. put a large titlebar w/buttons that is much larger than the width of the window (looks very weird)
3. drop the buttons for a very small window and depend on client-menu or keybinds (not optimal either)

Consolatis:
With the branch above this now looks like
20240701_19h18m25s_grim

Note that the black background is in fact not black but invisible, so whatever is behind that window will show through. Maybe we could put some solid rect behind or (although that would kind of complicate everything because it has to be enabled only in this specific situation, otherwise it would break half transparent window surfaces).

tokyo4j:
I like @Consolatis's approach from the developer's point of view because it's clear and solid, but I'm not sure workrave's users expect this behavior that ssd outline is larger than the surface.

Workrave's window size is a few pixels smaller than the minimal width and buttons overlap a bit in this PR, but I think it's acceptable and look a little bit nicer.

However, for even smaller windows like my test client with geometry of 1x1, this PR makes it look ugly @Consolatis's approach looks nicer.

||this PR|Consolatis's branch|
|-|-|-|
|workrave|20240702_11h29m13s_grim|20240702_11h29m56s_grim|
|1x1 window|20240702_11h43m07s_grim|20240702_11h44m12s_grim|

@jjramsey What do you think of this? Do you find top-right picture in the table above acceptable?

tokyo4j:
I'll close this PR and delegate it to @Consolatis, but keep the branch just in case.

ahesford:
I'd rather see the titlebar drawn properly and the window drawn too small if the client isn't well-behaved enough to honor the configure request. Even better would be detecting this case and drawing fatter borders to fill in the gaps. Another option would be to disable full SSD (leaving borders only) when the window is too small, but that might be bad from a usability standpoint.

Consolatis:
Switching to border-only on a 1px x 1px surface indeed sounds bad for UX.
Interesting idea to use the border to fill up the remaining part. I'll play around with both variants (additional rect using the border color which is usually disabled vs making the right border width wider + reposition).

tokyo4j:
In addition, I think we can increase `LAB_MIN_VIEW_WIDTH` a little bit so workrave window can be grabbed by pressing on the title like:

```diff
diff --git a/include/view.h b/include/view.h
index b4f90e6a..4d85743a 100644
--- a/include/view.h
+++ b/include/view.h
@@ -10,7 +10,7 @@
 #include <wlr/util/box.h>
 #include <xkbcommon/xkbcommon.h>
 
-#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT)
+#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT + 10)
 #define LAB_MIN_VIEW_HEIGHT 60
 
 /*
diff --git a/src/ssd/ssd-titlebar.c b/src/ssd/ssd-titlebar.c
index 828958ae..9b0502ee 100644
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -210,7 +210,7 @@ ssd_titlebar_update(struct ssd *ssd)
 		return;
 	}
 
-	int title_bar_size = MAX(0, width - SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT);
+	int title_bar_size = MAX(LAB_MIN_VIEW_WIDTH, width) - SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT;
 
 
 	struct ssd_part *part;

```

jp7677:
just to add my 2cents: From a usability point of view I guess it would be fine to hide/discard the minimize/maximize/close buttons once the width is smaller than 4 x button-width. All three actions can be performed with the menu from the button on the left side, so all functionality is still there. That would leave  you with a smaller minimum width (just 1 * button-width instead of 4 * button-width) which hopefully covers even exotic cases like the one here. 
Having seen a few bits from the ssd code, this might be annoying to implement though. 

My point is mood if the client menu can be customized and those entries can be removed. Not sure from the top of my head. 

First choice would be ofc that client behaves.

Consolatis:
Some thoughts regarding some border colored replacement background:
- could also be used during interactive resize with `drawContents=false`.
- using the right border doesn't work because it will then trigger a interactive resize (along with changing the cursor) which feels strange
- needs separate states for activate and inactive views
- parenting it to the titlebar tree doesn't work either because it may be hidden in border-only mode

So I chose a single `(struct wlr_scene_rect *) ssd->view_background` which is created on-demand if needed + handle the different active / inactive colors in `ssd_set_active()`. Branch updated: https://github.com/labwc/labwc/compare/master...Consolatis:labwc:fix/min_view_size

If @tokyo4j is fine with it I'd squash everything and create a PR.

johanmalm:
> > we don't need to pay much attention to peculiar clients that refuses to resize its window and keep it smaller than the minimal size
> 
> Agree, this is a niche case.
> 
> Whatever we go with, having the buttons rendered the wrong way around feels like a bug. On the other side, that look-through empty space when we enforce the minimal decoration size also feels strange.
> 
> Thoughts @johanmalm @ahesford ?

I think the Consolatis approach is better - i.e. render buttons propertly and accept a 'look-through empty space'

tokyo4j:
> If @tokyo4j is fine with it I'd squash everything and create a PR.

I did a quick test with your branch and I like it so far.

> - could also be used during interactive resize with `drawContents=false`.

Perhaps we can implement Openbox-like behavior of `drawContents=false` if we combine `ssd->view_background`  with `wlr_scene_buffer_set_dest_size()` for surfaces. I'll investigate that potential and see if its complexity is acceptable.

Consolatis:
- #1959

I didn't add the minimal title_bar_width restriction because its not strictly related to the issue at hand. I like the idea though, currently we a user can only move a window like that with `A-LeftMouseButton`.

PR #1950, opened 2024-07-01
xdg: support xdg-shell v3 with popup repositioning

jlindgren90:
See https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3514 which added support on the wlroots side.

We now re-run popup positioning (for both xdg-shell and layer-shell popups) when the "reposition" event is received. This allows popups that change size (such as qmpanel's applications menu) to be positioned correctly.

xdg-shell v3 also gives the compositor some additional "hints" for popup positioning (reactive, parent_size, and parent_configure_serial) which are available but we don't make use of currently.

### DIFF
--- a/include/layers.h
+++ b/include/layers.h
@@ -32,6 +32,7 @@ struct lab_layer_popup {
 	struct wl_listener commit;
 	struct wl_listener destroy;
 	struct wl_listener new_popup;
+	struct wl_listener reposition;
 };
 
 void layers_init(struct server *server);
--- a/src/layers.c
+++ b/src/layers.c
@@ -356,6 +356,7 @@ popup_handle_destroy(struct wl_listener *listener, void *data)
 		wl_container_of(listener, popup, destroy);
 	wl_list_remove(&popup->destroy.link);
 	wl_list_remove(&popup->new_popup.link);
+	wl_list_remove(&popup->reposition.link);
 
 	/* Usually already removed unless there was no commit at all */
 	if (popup->commit.notify) {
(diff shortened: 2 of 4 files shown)


Consolatis:
Looks good, we should likely delay this until after 0.7.3 though unless @johanmalm is fine with taking this in before the release.

johanmalm:
I'm good with merging this now in principle as it's early in the cool-down and fixes bugs.

However, I can't get it to work. I tried with `qmpanel` (master branch) but popups are still position wrongly. Am I missing something. On Arch... all up to date. See video.

[recording.webm](https://github.com/labwc/labwc/assets/6955353/2b583d19-83fa-4a77-ac57-9ad220c20718)

jlindgren90:
It needs a change on the Qt side too, which was merged for 6.8 but isn't in 6.7.x.

johanmalm:
> It needs a change on the Qt side too, which was merged for 6.8 but isn't in 6.7.x.
> 
> Edit: [qt/qtwayland@ec932c9](https://github.com/qt/qtwayland/commit/ec932c9d24ee0559d15251c05960e4ec031318a6)

Ah okay. I shouldn't have doubted.
Have you built qt6-wayland locally then?

jlindgren90:
I have rebuilt both qt6-wayland and qt6-base (takes a while) locally with some patches including that one. Popup positioning in general is still on Qt's "known Wayland issues" list (don't have the link atm). Qt very much wants to position popups itself since every other display system works that way. Wayland breaks that model.

johanmalm:
> I have rebuilt both qt6-wayland and qt6-base (takes a while) locally with some patches including that one. 

Wow!

Well, LGTM, let's go for the merge.

droc12345:
Going with at least version 3 lets you do a few things to the xdg-shell/surface/popup, things like reposition.
At least that's what I think when reading the xml. Which should make it easier for things like qt to do their own
thing,

PR #1952, opened 2024-07-02
session-lock: fix some bugs related to keyboard focus etc.

### LINKED ISSUE #1943: FollowMouse and swaylock: wrong window active
A very minor issue, but one that I find annoying.

I've got the following in my `rc.xml`:
```
  <focus>
    <followMouse>yes</followMouse>
    <followMouseRequiresMovement>no</followMouseRequiresMovement>
  </focus>
```
I also have the following in my `autostart` script:
```
swayidle -w \
	timeout 550 'swaylock -f -c 000000' \
	timeout 600 'wlopm --off \*' \
	resume 'wlopm --on \*' \
	before-sleep 'swaylock -f -c 000000' >/dev/null 2>&1 &
```
This works fairly well, in the sense that the active window is always the one under the mouse, except in one case: if swayidle triggers and swaylock locks the system, it is sometimes the case that the active window is not the one under the mouse.  Moving the mouse clears the discrepancy, but I occasionally end up typing in the wrong window, which is mildly annoying.

tokyo4j:
- Fix #1943
- Fix a bug I introduced in #1858 that allowed multiple clients to lock the screen.
- Clear keyboard focus on lock so keyboard events are never sent to normal surfaces once the screen is locked, even when the session-lock client haven't mapped its surface yet.

### DIFF
--- a/src/session-lock.c
+++ b/src/session-lock.c
@@ -287,6 +287,7 @@ handle_lock_destroy(struct wl_listener *listener, void *data)
 	wl_list_remove(&manager->lock_destroy.link);
 	wl_list_remove(&manager->lock_unlock.link);
 	wl_list_remove(&manager->lock_new_surface.link);
+	manager->lock = NULL;
 }
 
 static void
@@ -323,6 +324,7 @@ handle_new_session_lock(struct wl_listener *listener, void *data)
 	wl_signal_add(&lock->events.destroy, &manager->lock_destroy);
 
 	manager->locked = true;
+	manager->lock = lock;
 	wlr_session_lock_v1_send_locked(lock);
 }
 
--- a/src/session-lock.c
+++ b/src/session-lock.c
@@ -276,12 +276,6 @@ handle_lock_destroy(struct wl_listener *listener, void *data)
 	struct session_lock_manager *manager =
 		wl_container_of(listener, manager, lock_destroy);
 
-	float *black = (float[4]) { 0.f, 0.f, 0.f, 1.f };
-	struct session_lock_output *lock_output;
-	wl_list_for_each(lock_output, &manager->session_lock_outputs, link) {
-		wlr_scene_rect_set_color(lock_output->background, black);
-	}
-
 	wl_list_remove(&manager->lock_destroy.link);
 	wl_list_remove(&manager->lock_unlock.link);
 	wl_list_remove(&manager->lock_new_surface.link);
--- a/src/session-lock.c
+++ b/src/session-lock.c
@@ -269,7 +269,14 @@ handle_lock_unlock(struct wl_listener *listener, void *data)
 		wl_container_of(listener, manager, lock_unlock);
 	session_lock_destroy(manager);
 	manager->locked = false;
-	desktop_focus_topmost_view(manager->server);
+
+	if (manager->last_active_view) {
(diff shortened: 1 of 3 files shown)


src/view.c
```diff
@@ -2396,6 +2396,10 @@ view_destroy(struct view *view)
 		server->active_view = NULL;
 	}
 
+	if (server->session_lock_manager->last_active_view == view) {
+		server->last_raised_view = NULL;
+	}
```
Consolatis, line 2401: Looks like this resets the wrong view

src/session-lock.c
```diff
@@ -266,25 +266,30 @@ handle_lock_unlock(struct wl_listener *listener, void *data)
 		wl_container_of(listener, manager, lock_unlock);
 	session_lock_destroy(manager);
 	manager->locked = false;
-	desktop_focus_topmost_view(manager->server);
+
+	if (manager->last_active_view) {
+		desktop_focus_view(manager->last_active_view, /* raise */ false);
+	}
```
Consolatis, line 272: Should we have an else branch that focuses the topmost view here?

Consolatis:
Other than the UX question above regarding focusing the topmost view if there isn't an last known active view this LGTM, not tested.

src/session-lock.c
```diff
@@ -211,7 +211,10 @@ session_lock_output_create(struct session_lock_manager *manager, struct output *
 		goto exit_session;
 	}
 
-	/* Delay blanking output by 100ms to prevent flashing */
+	/*
+	 * Delay blanking output by 100ms to prevent flciker. If the session is
+	 * already locked, blank immediately.
```
Consolatis, line 216: Nit: `s/flciker/flicker/`

tokyo4j: Thanks.

src/session-lock.c
tokyo4j: Done.

I tried KDE Plasma (which doesn't use ext-session-lock) just now and I found it focuses on the last-focused window on unlock even when no window is focused on lock, so it will make a similar feeling.

tokyo4j:
While testing this, I found killing `gtklock` reveals the normal views, which should be hidden by solid-color rectangles. I'll investigate and fix it.

tokyo4j:
@Consolatis @johanmalm
Just to confirm, can I merge this PR?

All the functionalities are working in my test, including (un)plugging outputs while session is locked, killing `swaylock` and re-launching it, and launching multiple `swaylock` instances.

johanmalm:
LGTM
Yes, please merge.
Grateful for the work on this.

Consolatis:
> @Consolatis @johanmalm Just to confirm, can I merge this PR?

As a general statement, I completely trust your judgment in regards to code and testing PRs even if I didn't test them myself. Once you get a LGTM from at least one other project member feel free to go for the merge.

tokyo4j:
> > @Consolatis @johanmalm Just to confirm, can I merge this PR?
> 
> As a general statement, I completely trust your judgment in regards to code and testing PRs even if I didn't test them myself. Once you get a LGTM from at least one other project member feel free to go for the merge.

Alright. I was a bit uncertain because small changes were made since your last LGTM, but yes, those changes don't require another LGTM as those were just adding `desktop_focus_topmost_view()` which you mentioned and non-functional refactors.

I'll have more confidence next time :)

PR #1954, opened 2024-07-02
ssd/ssd-shadow.c: fix memory leak

Consolatis:
Before this patch, we were leaking memory [0] because the shadow implementation did not free the ssd_parts on destruction.

There was also no check if shadows were actually enabled via rc.xml or not so this also impacted people who were not using shadows but were not setting the shadow size via their theme to 0.

[0] 44 bytes per ssd_part * 8 parts * 2 states == 704 bytes per view closed. Note that Reconfigure also re-creates the SSD, thus we were also leaking 704 bytes * nr_views per Reconfigure.


johanmalm:
Thanks

PR #1959, opened 2024-07-03
ssd: allow ssd to be smaller than minimal size

Consolatis:
With 2603dbf labwc refused to update ssd geometry when its geometry is smaller than minimal size in order to prevent passing negative values in wlr_scene_rect_set_size(), but it made ssd for small windows like workrave ugly. So this commit fixes the negative values for each call to wlr_scene_rect_set_size() individually rather than just early-return in ssd_update_geometry().

---

Follow-up from
- #1949

### DIFF
--- a/src/ssd/ssd-extents.c
+++ b/src/ssd/ssd-extents.c
@@ -104,7 +104,7 @@ ssd_extents_update(struct ssd *ssd)
 
 	struct theme *theme = view->server->theme;
 
-	int width = view->current.width;
+	int width = MAX(view->current.width, LAB_MIN_VIEW_WIDTH);
 	int height = view_effective_height(view, /* use_pending */ false);
 	int full_height = height + theme->border_width * 2 + ssd->titlebar.height;
 	int full_width = width + 2 * theme->border_width;
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -138,6 +138,11 @@ ssd_get_part_type(const struct ssd *ssd, struct wlr_scene_node *node)
 			}
 		}
 	}
+
+	if (ssd->view_background && node == &ssd->view_background->node) {
+		return LAB_SSD_CLIENT;
+	}
+
 	return LAB_SSD_NONE;
 }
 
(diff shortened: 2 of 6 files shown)


src/ssd/ssd-extents.c
```diff
@@ -104,7 +104,7 @@ ssd_extents_update(struct ssd *ssd)
 
 	struct theme *theme = view->server->theme;
 
-	int width = view->current.width;
+	int width = MAX(view->current.width, LAB_MIN_VIEW_WIDTH);
```
tokyo4j, line 107: Should we limit the size of extents?

I found today that `xterm` can be resized to 1x1, and that this change makes cursor position and view geometry out of sync while resizing like the video below:

https://github.com/labwc/labwc/assets/22029524/e26b6c5d-0686-4c08-8eaf-60c3c15d0c19

Consolatis: Nice catch! I only tested with the test client which doesn't allow resizing.
Limiting the extents seems correct to me but something seems off with the offset when then enlarging the surface again.

I'll look into it later.

jlindgren90:
This doesn't seem to work as desired when the view starts at less than the minimum size, for example `xterm -geom 1x1`:

Screenshot 2024-07-05 10:02:36

Maybe the logic in `ssd_titlebar_create` needs similar changes as `ssd_titlebar_update`?

jlindgren90:
By the way, I am looking into an alternative approach that would just scale down the SSD buttons as the view gets smaller. I think I would personally prefer that behavior if it's not too hard to implement.

Consolatis:
> By the way, I am looking into an alternative approach that would just scale down the SSD buttons as the view gets smaller. I think I would personally prefer that behavior if it's not too hard to implement.

That would be an option as well. Might get pretty tricky with the icons though as they are scaled and positioned within the button to match the hardcoded sizes. Additionally those and also the hover effects are created once when loading the theme and then shared across all SSD button instances.

@heroin-moose is currently working on configurable button width via theme settings but that doesn't have the same issue as this one in regards to possibly having separate sizes per SSD instance.

src/ssd/ssd.c
```diff
@@ -267,6 +262,25 @@ ssd_update_geometry(struct ssd *ssd)
...
+		 * view for poorly behaving clients that don't follow
+		 * our requested minimal size.
+		 */
+		if (!ssd->view_background) {
+			ssd->view_background = wlr_scene_rect_create(ssd->tree, 0, 0,
+				ssd->view == ssd->view->server->active_view
+					? rc.theme->window_active_border_color
+					: rc.theme->window_inactive_border_color);
+		}
+		wlr_scene_rect_set_size(ssd->view_background, LAB_MIN_VIEW_WIDTH, current.height);
```
tokyo4j, line 278: Maybe we should resize the background to the size of padding area rather than the whole view size, because the surface may be translucent.

johanmalm:
Set to draft until discussion in #1964 is concluded.

ahesford:
Superseded by #2116

Issue #1958, opened 2024-07-03
X11 resources set by xrdb don't survive xwayland restart


jech:
I've got the following in my `autostart` file:
```
xrdb < ~/.Xresources
```
Occasionally, the X resources are not taken into account after labwc starts, and I need to manually run `xrdb < ~/.Xresources` from a terminal.

Is that an issue with labwc?  Or am I putting the X resources at the wrong place?

ahesford:
`xrdb` commands in the autostart will never have the intended effect, because `Xwayland` will terminate itself after a period of inactivity. Thus, you launch the X server to set the resources, and it dies if you don't start an X11 client right away.

Xwayland seems to read `~/.Xdefaults` by default (not sure if this is a wlroots thing or an Xwayland thing), so I suggest

    ln -s .Xresources ~/.Xdefaults

and drop the autostart.

jech:
> Xwayland seems to read ~/.Xdefaults by default

I could be wrong, but I belive that `.Xdefaults` is an older mechanism, that was made obsolete by xrdb.  As far as I rememebr, Xdefaults are read by the older applications at startup, they are not stored in the server.

What's the right way to run a script whenever Xwayland is started?

ahesford:
You might be right about the application taking responsibility for reading Xdefaults.

If your applications aren't reading Xdefaults (it seems to work for urxvt and gimp in my tests; I don't routinely run other clients), you'll have to either keep the X server alive by running a client constantly, wrap Xwayland to do what you need, or patch wlroots.

jech:
I see, thanks.

> If your applications aren't reading Xdefaults

Xdefaults is not a suitable replacement for Xresources, since it is not forwared over `ssh -X`.  I'll file an issue against wlroots, and see what they have to say.

Consolatis:
xwayland can be run in two separate ways within wlroots
- always running 
- the "lazy" approach: only open the X11 socket and wait for a client to connect and tell xwayland to terminate itself when it hasn't seen a client in like 10 seconds

Labwc uses the "lazy" approach due to obvious benefits when not using X11 clients continuously. Technically, labwc could implement another script that is run automatically whenever xwayland starts up to cater for use-cases like xrdb.

jech:
A simple solution would be to run a user-defined script, say `autostart-x11`, whenever we launch Xwayland.

ahesford:
[...]

PR #1961, opened 2024-07-03
xwayland: allow persistence

ahesford:
This seems to work as expected, keeping the server alive indefinitely at launch rather than allowing it to die.
- Changing the configuration option and reconfiguring does not change the behavior, because the wlroots xwayland object is still the same as was created at launch. Changing the behavior requires a compositor restart. Do we care? Attempting to fix this would seem to require terminating the X server, which could be catastrophic for any connected clients.
- Is there a better name or placement for the configuration option?

Closes: #1958.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -194,6 +194,13 @@ this is for compatibility with Openbox.
 	be used with labwc the preferred mode of the monitor is used instead.
 	Default is no.
 
+*<core><xwaylandPersistence>* [yes|no]
+	Keep XWayland alive even when no clients are connected, rather than
+	using a "lazy" policy that allows the server to launch on demand and die
+	when it is no longer needed. Default is no.
+
+	Note: changing this setting requires a restart of labwc.
+
 ## PLACEMENT
 
 *<placement><policy>* [center|automatic|cursor]
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -13,6 +13,7 @@
     <adaptiveSync>no</adaptiveSync>
     <allowTearing>no</allowTearing>
     <reuseOutputMode>no</reuseOutputMode>
+    <xwaylandPersistence>no</xwaylandPersistence>
   </core>
 
   <placement>
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -55,6 +55,7 @@ struct rcxml {
 	bool allow_tearing;
 	bool reuse_output_mode;
 	enum view_placement_policy placement_policy;
+	bool xwayland_persistence;
 
 	/* focus */
 	bool focus_follow_mouse;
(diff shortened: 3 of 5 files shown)


Consolatis:
> * Changing the configuration option and reconfiguring does not change the behavior, because the wlroots xwayland object is still the same as was created at launch. Changing the behavior requires a compositor restart. Do we care?

No, I think that limitation is fine. I don't expect users wanting to keep changing this. We should add that limitation to the man page though.

Consolatis:
LGTM, not tested.

Personally, I am fine with having this bypass the release cooldown. Thoughts @johanmalm ?

ahesford:
I did some more investigation with the `xinitrc` addition:
1. When Xwayland is not running, running `xrdb -query` locally will print no content (even when I have an .Xresources) the first time. Likewise, running something like `gimp` when Xwayland is not currently running will cause the application to behave as if no resources were set.
2. Once Xwayland has launched, running the client (`xrdb -query` or `gimp`) will behave as expected, because the `xinitrc` will have loaded the resource database.
3. Connecting to a remote host with SSH forwarding when Xwayland is NOT running, and running a client like `xrdb -query` seems to show the right behavior immediately.

The problem is that we aren't notified that Xwayland is ready until it is already responding to clients, so the local application that triggers its launch has already queried the resource database before we've had a chance to (asynchronously) populate it. The X11 proxy built into SSH seems to delay things long enough that the local `xinitrc` can populate the database before the remote client accesses it.

This is a race we will seem destined to lose for local clients, and I don't believe there is anything that we can do about it. I don't even know if wlroots can do anything about it. I also don't know if my observation of "correct" behavior on remote, SSH-tunneled clients is a necessary consequence of the implementation, or just a race we are less likely to lose.

While there might be some value in supporting xinitrc (and, by reusing the session machinery, it doesn't really cost us anything to do so), I think the only reliable way to solve the configuration issue in #1958 is to configure Xwayland for persistence.

johanmalm:
So if making Xwayland persistent is the only way to reliably solve the
issue, are we sure that the benefits of xinitrc are greater than the
confusion it might cause.

The other core configuration options are nouns so suggest making it
xwaylandPersistence  (rather than persists). Happy to merge that commit it
cool down.

I'm not convinced by the net benefit of the xinitrc, but may be missing
some context.

Sorry, have to IRC or browser access for a few days.

jech:
> So if making Xwayland persistent is the only way to reliably solve the issue,

I'm sure we can solve the race (which might require changes to wlroots), but I agree that it'd be better to commit just the persistence change now, and delay committing the xinitrc support until we've solved the race.

ahesford:
I moved "persists" to "persistence" in the first commit, and decided to move the `xinitrc` support to #1963 for separate consideration. As described in that PR, I believe there is still merit in supporting an `xinitrc` script.

Consolatis:
Thanks!

jech:
I've just moved to 0.7.3, and this solves my issue quite nicely.  Thanks.

Issue #1957, opened 2024-07-03
Feature Request: Tablet Relative Tracking Mode


justin330:
I'd noticed the recent addition of mouseEmulation since updating, and got my hopes up that relative tracking support had finally  been added to labwc, but to my disappointment it seems to be for pen pressure only?

I used relative tracking extensively on Xorg desktops because it is much easier to write with when using my 'signature' sized Wacom tablet. I'd really appreciate if such a feature could be added if its not too much trouble.

jp7677:
Well, sorry to having you disappointed ;)

If you are able to compile labwc, you might try this PR https://github.com/labwc/labwc/pull/1962

For relative motion, you'll also need this configuration option:

```xml
    <tablet motion="relative" />
```

Note that this configuration switch will very likely change.

justin330:
Wow, that was fast! Thank you!
I just tried it out and it mostly works; but there is a diagonal drift that occurs whenever there is motion.

justin330:
I just tried a couple things and it seems to occur when the rotation is changed

jp7677:
Thanks for the fast feedback :). Could you explain a bit more detailed what you mean by diagonal drift and what motion is needed (eg. straight or also diagonal). Is this always like this (eg desktop vs surface). Do you have other tablet settings set?

justin330:
The drift also seems to occur whenever there is an motion event, such as when the x, y, or proximity to the pad changes.

justin330:
Another request if I may? Would you be able to add an entry to change the speed of the pen while in relative mode?

jp7677:
> I just tried a couple things and it seems to occur when the rotation is changed from 0

Thanks for the hint, the "drift" should be gone now i.c.w. rotation.

justin330:
Thank you so much for your speedy work! 

The drift is now gone while the pen is within the pad's workable area, though if the pen is moved outside the workable
area there is a drift in that direction. 

To be honest, though unintended its kind of nice in a way, its a bit the auto-scroll when dragging selected files to the top or bottom of the list in a file manager.

Here's a brief illustration:
* Orange is the tablet's workable area, corresponding to the white lines on the tablet
![diagram](https://github.com/labwc/labwc/assets/9954903/171cda39-cf17-494d-a355-b45da5e3b91a)

jp7677:
> Thank you so much for your speedy work!

You are welcome!

[...]

PR #1962, opened 2024-07-03
feat: support relative tablet motion

jp7677:
Still draft since early days and there is already a big flaw:  the motion configuration overwrites all tools (Stylus, Pen, Tablet Mouse etc), but actually this should be on a per-tool base. Since we have `tool->hardware_serial` and/or `tool->hardware_wacom`, this should be doable. 

TODO:
- [x] ~configuration per tool~
- [x] add/improve code comments
- [x] documentation

PS: Multiple tool configurations could also be a separate PR, or may be even wait until someone hits this to have a real life demand.

Closes https://github.com/labwc/labwc/issues/1957

### DIFF
--- a/include/input/tablet.h
+++ b/include/input/tablet.h
@@ -9,12 +9,18 @@ struct seat;
 struct wlr_device;
 struct wlr_input_device;
 
+enum lab_tablet_motion_mode {
+	LAB_TABLET_MOTION_ABSOLUTE = 0,
+	LAB_TABLET_MOTION_RELATIVE,
+};
+
 struct drawing_tablet {
 	struct wlr_input_device *wlr_input_device;
 	struct seat *seat;
 	struct wlr_tablet *tablet;
 	struct wlr_tablet_v2_tablet *tablet_v2;
-	double x, y;
+	enum lab_tablet_motion_mode motion_mode;
+	double x, y, dx, dy;
 	double distance;
 	double pressure;
 	double tilt_x, tilt_y;
--- a/include/input/tablet.h
+++ b/include/input/tablet.h
@@ -9,17 +9,12 @@ struct seat;
 struct wlr_device;
 struct wlr_input_device;
 
-enum lab_tablet_motion_mode {
-	LAB_TABLET_MOTION_ABSOLUTE = 0,
-	LAB_TABLET_MOTION_RELATIVE,
-};
-
 struct drawing_tablet {
 	struct wlr_input_device *wlr_input_device;
 	struct seat *seat;
 	struct wlr_tablet *tablet;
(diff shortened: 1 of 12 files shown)


src/input/tablet.c
```diff
@@ -224,15 +263,15 @@ handle_axis(struct wl_listener *listener, void *data)
 	struct drawing_tablet *tablet = ev->tablet->data;
 	struct drawing_tablet_tool *tool = ev->tool->data;
 
-	if (!tool_supports_absolute_motion(ev->tool)) {
-		return;
-	}
-
+	tablet->dx = 0;
```
jp7677, line 266: Note to self: no need to remember dx, dy since they are relative. May be the same applies for tilt?

include/input/tablet.h
```diff
@@ -9,12 +9,18 @@ struct seat;
...
+};
+
 struct drawing_tablet {
 	struct wlr_input_device *wlr_input_device;
 	struct seat *seat;
 	struct wlr_tablet *tablet;
 	struct wlr_tablet_v2_tablet *tablet_v2;
-	double x, y;
+	enum lab_tablet_motion_mode motion_mode;
+	double x, y, dx, dy;
```
jp7677, line 23: Note to self: no need to remember dx, dy since they are relative. May be the same applies for tilt?

src/input/tablet.c
```diff
@@ -83,30 +92,76 @@ adjust_for_rotation(enum rotation rotation, double *x, double *y)
...
+
+	/* initialize here to avoid a maybe-uninitialized warning */
+	double lx = -1, ly = -1;
+	switch (tablet->motion_mode) {
+	case LAB_TABLET_MOTION_ABSOLUTE:
+		wlr_cursor_absolute_to_layout_coords(tablet->seat->cursor,
+			tablet->wlr_input_device, *x, *y, &lx, &ly);
+		break;
+	case LAB_TABLET_MOTION_RELATIVE:
+		/* dx,dy are already layout coordinate deltas */
```
jp7677, line 154: Note to self, not really true https://wayland.freedesktop.org/libinput/doc/latest/api/group__event__tablet.html#ga868763e264a48cb8df72f5b4724a3c4e

jp7677:
I made up my mind for the configuration section. We are setting tablet mouses and lenses always to relative mode, regardless the configuration. Sway does the same. That way there is a lot less need to configure multiple tools separately. Postpone this until there is a real demand.

include/input/tablet.h
jp7677: Let's keep it like this for readability.

jp7677:
All good from my side, thus ready to review. This is ofc 0.7.4 material.

johanmalm:
Thanks. Nice work!

PR #1963, opened 2024-07-04
xwayland: support xinitrc scripts to configure server on launch

ahesford:
Cf. #1961. This splits the xinitrc support for separate consideration.

Even though this change doesn't address the issue in #1958 because of a [race that we can't control](https://github.com/labwc/labwc/pull/1961#issuecomment-2207000149), I think it still warrants inclusion for four reasons:
1. It seems to partially address the issue, especially for remote clients where symlinking `.Xdefaults -> .Xresources` server-side is insufficient to trigger Xlib behavior that will cause clients to read resource configuration on disk.
2. It allows users of even local clients to force things like an `xrdb` load every time Xwayland is lazily restarted. This means that, even if the intitial client triggering its launch must fall back on `.Xdefaults` behavior for proper configuration, subsequent clients that connect during the lifetime of a particular Xwayland session can avoid disk access by reading the resource database from the properties of the root window.
3. Users may wish adopt other workflows that opportunistically launch X11 clients every time that Xwayland is started, without regard for sequencing with other clients triggering an auto-launch.
4. The `xinitrc` script permits a logical separation between "Xwayland configuration" and "labwc client configuration" as done in the `autostart` script.

### DIFF
--- a/docs/README
+++ b/docs/README
@@ -5,6 +5,7 @@ Config layout for ~/.config/labwc/
 - rc.xml
 - shutdown
 - themerc-override
+- xinitrc
 
 See `man labwc-config and `man labwc-theme` for further details.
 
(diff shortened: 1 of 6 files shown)


johanmalm:
Thanks.

PR #1964, opened 2024-07-05
ssd: scale titlebar buttons to fit for very small views

jlindgren90:
Alternative to:
- https://github.com/labwc/labwc/pull/1959

The rendering is not pixel-perfect since we still use some shared/pre-rendered buffers from the theme, but IMHO it looks fairly decent given the tricky geometry.

xterm at 1x4, 3x4, 5x4, 7x4, and 9x4:

xterm-1x4
xterm-3x4
xterm-5x4
xterm-7x4
xterm-9x4

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -65,6 +65,7 @@ struct ssd {
 	/* The top of the view, containing buttons, title, .. */
 	struct {
 		int height;
+		int button_width;
 		struct wlr_scene_tree *tree;
 		struct ssd_sub_tree active;
 		struct ssd_sub_tree inactive;
@@ -129,15 +130,15 @@ struct ssd_part *add_scene_button(
 	struct wl_list *part_list, enum ssd_part_type type,
 	struct wlr_scene_tree *parent, float *bg_color,
 	struct wlr_buffer *icon_buffer, struct wlr_buffer *hover_buffer,
-	int x, struct view *view);
+	int x, int width, struct view *view);
 void add_toggled_icon(struct ssd_button *button, struct wl_list *part_list,
 	enum ssd_part_type type, struct wlr_buffer *icon_buffer,
-	struct wlr_buffer *hover_buffer);
+	struct wlr_buffer *hover_buffer, int width);
 struct ssd_part *add_scene_button_corner(
 	struct wl_list *part_list, enum ssd_part_type type,
 	enum ssd_part_type corner_type, struct wlr_scene_tree *parent,
 	struct wlr_buffer *corner_buffer, struct wlr_buffer *icon_buffer,
-	struct wlr_buffer *hover_buffer, int x, struct view *view);
+	struct wlr_buffer *hover_buffer, int x, int width, struct view *view);
 
 /* SSD internal helpers */
 struct ssd_part *ssd_get_part(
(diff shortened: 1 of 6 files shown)


droc12345:
That actually looks pretty decent.

Consolatis:
I am currently very low on time and didn't yet have a chance to test this.

From the visual perspective I think I'd prefer this over #1959.
From the implementation perspective I don't particularly like that we recreate the whole titlebar during interactive resize. This may be very expensive (due to the interval in which we do so, especially when using high polling mice), on the other hand this does only happen for very small sizes and clients that don't behave as expected so should actually be fine. (Didn't yet look into why xterm can be interactively resized to < VIEW_MIN_WIDTH.)

Third general observation, I'd vote for delaying the imminent release until we decide on how to fix this and go through the remaining issue on the [milestone](https://github.com/labwc/labwc/milestone/6) and either move them to the 0.7.4 milestone or fix them.

CC
@ahesford @johanmalm @tokyo4j

ahesford:
I like the idea.

It might be nice if we could adopt a bit more sophistication by limiting the minimum size and hiding buttons according to priority if further size reductions are necessary, in order of the first to be excluded:
1. Minimize
2. Maximize
3. Menu
4. Close

That might be better handled by an implementation that would also allow configurable button layouts; just select the layout dynamically based on the window size; because that is more extensive, I don't know that waiting for this behavior is worth delaying a fix.

I don't have a strong preference about delaying the release, or even how to address this sizing problem.

droc12345:
I would place menu last, simply because you can access what the buttons do from the menu.

This small size problem is only going to happen for a very small amount of users with a very small amount of apps.
Most people will not start something like xterm at a very small size, and the apps that do something 
like start with a 1x1 are apps based on tk or other semi-ancient toolkits, very few in use nowadays.

johanmalm:
I'm not particularly keen on the shrinking the buttons. Prefer something simpler like just hiding those that don't fit.

> As a more general observation, I'd vote for delaying the imminent release until we decide on how to fix this and go through the remaining issue on the [milestone](https://github.com/labwc/labwc/milestone/6) and either move them to the 0.7.4 milestone or fix them.

Have discussed on IRC. Will hold release as planned on Friday unless unexpected bugs emerge.
Prefer time-based rather than feature/fix based.
@Consolatis and I have discussed and will try a different approach regarding issues against milestones.

domo141:
I (also) like hiding instead of resizing. i'd also liked titlebar always have region which is not button (TitleBar or Title <context> (I don't know the difference), which can be bound to raise, move (single-click default), maximize(double-click default), shade(my binding)and so on -- IMO titlebar w/ single close button is not too useful (just as an example)

johanmalm:
Have tested with openbox. Looks like it first shrinks title to 'nothing' and then clips the buttons (rather than removing a whole button at a time). I think this is a good approach. Might be able to use `wlr_scene_subsurface_tree_set_clip()` for this.

Ref:
- https://github.com/swaywm/sway/blob/274a5fcb73b1252e05543669c956a370516f052c/sway/tree/view.c#L964C3-L964C37

johanmalm:
Have set this to draft so we don't merge by mistake... until we agree an approach.

Consolatis:
Personally I really don't have an opinion here
- resizing the buttons via scaling
- removing buttons
- clipping buttons
- preventing ssd resize and fill the space with a rect

All sound fine to me.

johanmalm:
I think openbox approach is best (shrink title first - then clip buttons)

tokyo4j:
> I think openbox approach is best (shrink title first - then clip buttons)

I agree. Maybe we can do it after we implement customizable button layout.

PR #1965, opened 2024-07-06
theme: allow to set window button size

heroin-moose:
This commit introduces new option "window.button.width". Despite the name it currently affects both width and height.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -114,6 +114,10 @@ labwc-config(5).
 	Specifies how window titles are aligned in the titlebar for both
 	focused and unfocused windows. Type justification. Default Left.
 
+*window.button.width*
+	Width of a titlebar button, in pixels.
+	Default is 26.
+
 *window.active.button.unpressed.image.color*
 	Color of the images in titlebar buttons in their default, unpressed,
 	state. This element is for the focused window.
--- a/docs/themerc
+++ b/docs/themerc
@@ -29,6 +29,9 @@ window.active.label.text.color: #000000
 window.inactive.label.text.color: #000000
 window.label.text.justify: center
 
+# window button width
+window.button.width: 26
+
 # window buttons
 window.active.button.unpressed.image.color: #000000
 window.inactive.button.unpressed.image.color: #000000
--- a/include/ssd.h
+++ b/include/ssd.h
@@ -6,7 +6,6 @@
 #include "common/border.h"
 
 #define SSD_BUTTON_COUNT 4
-#define SSD_BUTTON_WIDTH 26
 #define SSD_EXTENDED_AREA 8
 
 /*
--- a/include/theme.h
+++ b/include/theme.h
@@ -45,6 +45,9 @@ struct theme {
 	float window_inactive_label_text_color[4];
 	enum lab_justification window_label_text_justify;
 
+	/* button width */
+	int window_button_width;
+
 	/* button colors */
 	float window_active_button_menu_unpressed_image_color[4];
 	float window_active_button_iconify_unpressed_image_color[4];
--- a/include/view.h
+++ b/include/view.h
@@ -10,7 +10,6 @@
 #include <wlr/util/box.h>
 #include <xkbcommon/xkbcommon.h>
 
-#define LAB_MIN_VIEW_WIDTH (SSD_BUTTON_WIDTH * SSD_BUTTON_COUNT)
 #define LAB_MIN_VIEW_HEIGHT 60
 
 /*
--- a/src/snap.c
+++ b/src/snap.c
@@ -220,6 +220,7 @@ snap_shrink_to_next_edge(struct view *view,
 
 	*geo = view->pending;
 	uint32_t resize_edges;
+	int min_view_width = rc.theme->window_button_width * SSD_BUTTON_COUNT;
 
 	/*
 	 * First shrink the view along the relevant edge. The maximum shrink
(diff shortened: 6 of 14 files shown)


johanmalm:
I'm happy to go with this (after the release) based on the fact that it doesn't change anything other than exposing a `#define` as a theme option. However it feels like we're touching an area that needs to TLC and exposing a variable as a theme option we probably set an expectation that it's well thought through and won't change in future, so I'd appreciate thoughts comments aginst the following:

1. All other button theme options use the format `window.(in)active.button.foo`. Should we the same. Can always use `window.*.button.width` to configure it across both active and inactive.
2. Consider describing in the man page that this also sets height for the time being.
3. We ought to describe in the man page what happens if `window.button.width` is greater than `titlebar.height - 2 * padding.height`?
4. Same for when `window.button.width` is greater than the size of an xbm icon? Assume we don't scale up (can't remember).
5. Is defult of 26 a good value? Feels like icons normally come in 16, 22, 24, 36, etc. Just checked a couple of GTK themes with labwc-gtktheme.py (with the icon PR). I suspect I just chose a hard-coded 26 at the time when we only had xbm support and no padding.width (which we still don't).
6. We ought to add padding.width at some point... and at that point we should probably decrease the default button.width, so do we want to do this now as part of this patch set?
7. We ought to define (and document) the logic between titlebar.height, window.button.width (and therefore height) as well as font-height+2*padding.height... i.e. which takes precedence. Currently titlebar.height can be set to reduce padding.height (but IIRC can never be smaller than font_height). What should the logic be wrt the button size?

heroin-moose:
> All other button theme options use the format window.(in)active.button.foo. Should we the same. Can always use window.*.button.width to configure it across both active and inactive.

We've discussed this with @Consolatis and it seems weird to have two seemingly separate options that do the same thing.

>  Consider describing in the man page that this also sets height for the time being.

Makes sense, will do.

> We ought to describe in the man page what happens if window.button.width is greater than titlebar.height - 2 * padding.height?

I'm not sure. Do we do this with other sizes?

> Is defult of 26 a good value? Feels like icons normally come in 16, 22, 24, 36, etc. Just checked a couple of GTK themes with labwc-gtktheme.py (with the icon PR). I suspect I just chose a hard-coded 26 at the time when we only had xbm support and no padding.width (which we still don't).

I've just ported the original 26 and I was wondering why it is 26 myself. Happy to change it to 24 or 36.

> We ought to add padding.width at some point... and at that point we should probably decrease the default button.width, so do we want to do this now as part of this patch set?

You mean making the default button smaller and using padding.width so visually things stay as before?

> We ought to define (and document) the logic between titlebar.height, window.button.width (and therefore height) as well as font-height+2*padding.height... i.e. which takes precedence. Currently titlebar.height can be set to reduce padding.height (but IIRC can never be smaller than font_height). What should the logic be wrt the button size?

This is very dark magic to me. Honestly it would've been better just to set individual things and dropping titlebar.height.

johanmalm:
Sorry - you got my 'unfiltered' thoughts.

I think this part of the code originates from pretty early commits, and tbg it's not gained sophistication with the rest. The `SSD_BUTTON_WIDTH` is a rather crude instrument and does in fact deviate from openbox spec which uses the size of the icon and just inserts `padding.width` between them.

My main reservation with this patch is that by exposing this variable without thinking through what end-state we want, we either accept an interface change later or just stick with this width variable.

I suggest we think through + agree the basic logic before proceeding. Intuitively I prefer just using the icon-size itself + `padding.width`. I feel we ought to limit the max size of the icon though, but not sure what's best here - possibly just doing a `MAX()` with `titlebar->height`.

> > We ought to describe in the man page what happens if window.button.width is greater than titlebar.height - 2 * padding.height?
> 
> I'm not sure. Do we do this with other sizes?

Good point - but we ought to define + agree it (i.e. should `titlebar.height` grow if icon is big?)

> > Is defult of 26 a good value? Feels like icons normally come in 16, 22, 24, 36, etc. Just checked a couple of GTK themes with labwc-gtktheme.py (with the icon PR). I suspect I just chose a hard-coded 26 at the time when we only had xbm support and no padding.width (which we still don't).
> 
> I've just ported the original 26 and I was wondering why it is 26 myself. Happy to change it to 24 or 36.
> 
> > We ought to add padding.width at some point... and at that point we should probably decrease the default button.width, so do we want to do this now as part of this patch set?
> 
> You mean making the default button smaller and using padding.width so visually things stay as before?

Yes.

> 
> > We ought to define (and document) the logic between titlebar.height, window.button.width (and therefore height) as well as font-height+2*padding.height... i.e. which takes precedence. Currently titlebar.height can be set to reduce padding.height (but IIRC can never be smaller than font_height). What should the logic be wrt the button size?
> 
> This is very dark magic to me. Honestly it would've been better just to set individual things and drop titlebar.height.

Yes, I must have had a weak moment merging that `titlebar.height` override variable! Okay, let's not worry about that logic for now.

johanmalm:
I've had a re-think and am happy to merge this... except that it segfaults on `window.button.width < 1`. Would you be happy to fix?

johanmalm:
Have force pushed with the following changes:

1. Clamped at 1 instead of 0 (to avoid segfault)
2. Resolved merge conflict with https://github.com/labwc/labwc/commit/3be8fe25f33bdf968a28aeffc08a370ab2d38b49

stefonarch:
Congrats to the release!

Just read and tried out 

>* Add theme option window.button.width to set window button size.
Written-by: @heroin-moose PR https://github.com/labwc/labwc/pull/1965

Actually here it affects only the distance between the buttons, with any height and any theme.  Maybe rewording "size" as in the man page?

Issue #1826, opened 2024-05-16
Feature request: focus/urgency handling for new/activating windows


Vladimir-csp:
There needs to be a policy for new windows, or windows requesting activation. Switching focus and workspace suddenly is not always a good thing.

IIRC Openbox had focus stealing prevention based on keyboard input. That would be great thing to have.

But this brings the opposite problem: events can be missed. Especially if window rules with `SendToDesktop` action are involved. Window may appear silently (and steals focus due to #1800).

To summarize:
- There needs to be a way to prevent focus stealing (as a global policy, probably based on keyboard input).
- There needs to be a way of dealing with windows activating/appearing on other workspaces other than immediately switching to them.
- There needs to be a way to alert user of new/activating windows that did not receive focus (due to focus stealing prevention, or `ignoreFocusRequest` rule applied) or were not shown entirely (due to being sent to another workspace by a rule, or due to being a modal window of a window on another workspace, or due to just being obscured by another windows).

Classic X11 urgency hints conveyed via taskbars were ideal: window's icon on a taskbar is flashing (possibly indefinitely), conveys an event clearly, without clogging notifications or getting in the way.
Sadly, urgency as a separate concept (as a feedback to user) seems to be missing from wayland protocols.
Sway implemented it for themselves, and waybar's sway-workspaces supports it. Window is flagged urgent either on activation event, or via a window rule. Window border becomes brightly colored, workspace button on waybar is brightly colored too (can be made flashing). Urgency is removed upon receiving focus (with a configurable delay). This is a very good implementation.

GNOME people suggest using notifications for this, but IMHO this is an abuse of unrelated channel of interaction.

What labwc could do without relying on any external integration?
- Convey urgency via window borders (or semi-transparent overlays for windows with hidden borders), add urgency-related parameters to themes.
- In absence of integration solution with taskbars via wayland or custom protocols, there still needs to be a way to convey urgency for a window on another workspace or completely obscured by other windows. A semi-transparent flashing OSD, or window outline?
- `NextUrgentWindow` action.

What could be done with external interaction?
- Go the Sway way, at least by providing ways for taskbar/workspace widgets to know when to show hints. Maybe someday something useful will be added to wayland protocols.

Narrat:
There is one mechanic integrated in labwc to achieve that. Via window rules and the use of  `<ignoreFocusRequest>yes</ignoreFocusRequest>`

johanmalm:
[...]

PR #1974, opened 2024-07-10
src/xdg.c: verify source surface for xdg_activation request

Consolatis:
wlroots < 0.17 didn't allow to reliably check the source surface of an xdg activation request as it reset the surface to NULL when it was destroyed before the token was used. This happens regularly for notifications for example. Thus we treated the token as valid even without checking for the source surface.

wlroots 0.17 added a new_token signal where we can attach information to the existing token which we can then use when evaluating activation requests. This patch implements that check.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -248,6 +248,7 @@ struct server {
 
 	struct wlr_xdg_activation_v1 *xdg_activation;
 	struct wl_listener xdg_activation_request;
+	struct wl_listener xdg_activation_new_token;
 
 	struct wl_list views;
 	struct wl_list unmanaged_surfaces;
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -665,10 +665,39 @@ static const struct view_impl xdg_toplevel_view_impl = {
 	.get_pid = xdg_view_get_pid,
 };
 
+struct token_data {
+	bool had_valid_surface;
+	bool had_valid_seat;
+	struct wl_listener destroy;
+};
+
+static void
+xdg_activation_handle_token_destroy(struct wl_listener *listener, void *data)
+{
+	struct token_data *token_data = wl_container_of(listener, token_data, destroy);
+	wl_list_remove(&token_data->destroy.link);
+	free(token_data);
+}
+
+static void
+xdg_activation_handle_new_token(struct wl_listener *listener, void *data)
+{
+	struct wlr_xdg_activation_token_v1 *token = data;
+	struct token_data *token_data = znew(*token_data);
+	token_data->had_valid_surface = !!token->surface;
+	token_data->had_valid_seat = !!token->seat;
+	token->data = token_data;
+
+	token_data->destroy.notify = xdg_activation_handle_token_destroy;
+	wl_signal_add(&token->events.destroy, &token_data->destroy);
+}
+
 static void
 xdg_activation_handle_request(struct wl_listener *listener, void *data)
 {
 	const struct wlr_xdg_activation_v1_request_activate_event *event = data;
+	struct token_data *token_data = event->token->data;
+	assert(token_data);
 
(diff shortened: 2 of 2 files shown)


johanmalm:
LGTM. Assume post release?

Is this implemented in `sway`? I can't see it.

Consolatis:
> Assume post release?

Yeah, post release.

> Is this implemented in `sway`? I can't see it.

It seems so:
- https://github.com/LBCrion/sfwbar/issues/183#issuecomment-2221086024

tokyo4j:
I'm afraid users see this fix as a regression.

I noticed `xdg-open` command no longer raises an existing Firefox window. Also, VSCode's `code [path to workspace]` command no longer raises the existing window, which seems like VSCode failed to launch at first glance.

I suggest following options:
- Revert this and delay until we implement urgency hint (#1826)
- Add an option to always allow focus request like `<windowRule ignoreFocusRequest="Never">`

stefonarch:
I don't see a regression in daily use with raising the firefox window, but notifications → activate → thunderbird is broken I noticed.

>What about some global setting like <focusRequestPolicy> and more options for the ignoreFocusRequest window rule?

Sounds good.

Consolatis:
> I don't see a regression in daily use with raising the firefox window, but notifications → activate → thunderbird is broken I noticed.

What notification daemon are you using, it sounds like it does not request a proper xdg activation token.

tokyo4j:
> I'd prefer to not revert (its only in the master branch for now)

Agreed. We have plenty of time before the next release and this PR certainly enhances security.

> Foot for example also uses xdg-open when opening a link from within foot but it requests a valid token based on the user input and its own surface and then sets that token as an env var which should then be delivered to firefox and thus firefox should be able to activate itself.

Thank you for providing an example that works. I've been wondering how to use xdg-activation protocol properly.

> What about some global setting like <focusRequestPolicy> and more options for the ignoreFocusRequest window rule?

Sounds good, but I wonder if we need to add more options to `ignoreFocusRequest`.
How about following format (not sure about the name too)?:
```xml
<core>
    <!-- Global policy for focus requests. For users who never get annoyed by focus stealing, set both to "no".  -->
    <focusRequestRequirement input="yes" focus="yes" />
</core>
<windowRules>
    <!-- Allow focus requests regardless of focusRequestRequirement for certain applications.  -->
    <windowRule identifier="firefox" ignoreFocusRequest="no" />
    <!-- Disallow focus requests regardless of focusRequestRequirement for certain applications  -->
    <windowRule identifier="annoying-app" ignoreFocusRequest="yes" />
</windowRules>
```

> It is hard to find self describing naming for those though, we basically have two variables and thus 4 potential states:
> - token requester had a recent input (key or mouse event) -> token->had_valid_seat=true
> - token requester had keyboard or mouse focus -> token->had_valid_surface=true

So my understanding is we need to make following checks for each state. Am I right?
- Measure the time from the last input to the activation request
- Check if the surface is the focused one or not

stefonarch:
> > I don't see a regression in daily use with raising the firefox window, but notifications → activate → thunderbird is broken I noticed.
> 
> What notification daemon are you using? It sounds like it does not request a valid xdg activation token.

In fact, afaik we do not request anything yet in `lxqt-notificationd`.

Consolatis:
> * Measure the time from the last input to the activation request

That is already done by wlroots.
If `token->had_valid_seat` is set, wlroots should have already verified that the serial was fine.
wlroots will also invalidate tokens that are older than 30 seconds automatically and we don't even get the activation request from what I remember. However, those 30 seconds can be configured by the compositor.

> * Check if requester's surface is the focused one or not

Wlroots already did so, if `token->had_valid_surface` is set, wlroots verified that the supplied surfaces had keyboard or mouse focus at the time the token was requested.

PR #1976, opened 2024-07-12
menu: support titles defined by `<separator label="">`

johanmalm:
For visibility only so nobody else duplicates effort.

Cc: @heroin-moose  (as you mentioned this some time ago).

Have implemented `menu.title.bg.color: #589bda`

TODO:

- [x] Documentation
- [x] Handle title before first item

foo

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -200,12 +200,7 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
 
 	const char *arrow = show_arrow ? "›" : NULL;
 
-	/* TODO: Consider setting this somewhere else */
-	if (!menu->item_height) {
-		menu->item_height = font_height(&rc.font_menuitem)
-			+ 2 * theme->menu_item_padding_y;
-	}
-	menuitem->height = menu->item_height;
+	menuitem->height = theme->menu_item_height;
 
 	int x, y;
 	menuitem->native_width = font_width(&rc.font_menuitem, text);
@@ -225,11 +220,11 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
 	/* Item background nodes */
 	menuitem->normal.background = &wlr_scene_rect_create(
 		menuitem->normal.tree,
-		menu->size.width, menu->item_height,
+		menu->size.width, theme->menu_item_height,
 		theme->menu_items_bg_color)->node;
 	menuitem->selected.background = &wlr_scene_rect_create(
 		menuitem->selected.tree,
-		menu->size.width, menu->item_height,
+		menu->size.width, theme->menu_item_height,
 		theme->menu_items_active_bg_color)->node;
 
 	/* Font nodes */
@@ -258,9 +253,9 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
 
 	/* Center font nodes */
 	x = theme->menu_item_padding_x;
-	y = (menu->item_height - menuitem->normal.buffer->height) / 2;
+	y = (theme->menu_item_height - menuitem->normal.buffer->height) / 2;
 	wlr_scene_node_set_position(menuitem->normal.text, x, y);
-	y = (menu->item_height - menuitem->selected.buffer->height) / 2;
+	y = (theme->menu_item_height - menuitem->selected.buffer->height) / 2;
 	wlr_scene_node_set_position(menuitem->selected.text, x, y);
 
 	/* Position the item in relation to its menu */
@@ -289,7 +284,7 @@ separator_create(struct menu *menu, const char *label)
 	struct theme *theme = server->theme;
 
 	if (menuitem->type == LAB_MENU_TITLE) {
-		menuitem->height = menu->item_height;
+		menuitem->height = theme->menu_item_height;
 		menuitem->native_width = font_width(&rc.font_menuitem, label);
 	} else if (menuitem->type == LAB_MENU_SEPARATOR_LINE) {
 		menuitem->height = theme->menu_separator_line_thickness +
@@ -328,7 +323,7 @@ separator_create(struct menu *menu, const char *label)
 		/* Center font nodes */
 		int x, y;
 		x = theme->menu_item_padding_x;
-		y = (menu->item_height - menuitem->normal.buffer->height) / 2;
+		y = (theme->menu_item_height - menuitem->normal.buffer->height) / 2;
 		wlr_scene_node_set_position(menuitem->normal.text, x, y);
 	} else {
 		int nominal_width = theme->menu_min_width;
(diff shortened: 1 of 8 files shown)


src/menu/menu.c
```diff
@@ -114,23 +114,32 @@ menu_update_width(struct menu *menu)
...
-					max_width);
-				scaled_font_buffer_set_max_width(item->selected.buffer,
-					max_width);
-			}
+				width, theme->menu_separator_line_thickness);
+		}
+
+		if (!item->selected.background) {
+			continue;
+		}
```
Consolatis, line 132: What exactly does this mean? Does the earlier comment still apply here that this is another way for defining a separator?

johanmalm: Will make it more readable (maybe by using `menuitem->type` or something).

Basically, `item->selected` doesn't exist for separator because they're not selectable :smile: ... but that should be a lot more obvious to the reader.

We might be able to get rid of `menu_update_width()` all together, but will try to take small steps.

droc12345:
???

test

johanmalm:
> ???
> 
> test

Yes... looks like we don't yet handle a title before the first item. I had something similar happen. Not investigated yet.

droc12345:
I put a separator with label part way down the menu and it does something similar to the above
the separator label is highlighted, and it's the right width but the rest of the menu both above and below
do not pick up the width from the label, so still the same width as the normal items.

droc12345:
Finally figured out what was going on. The separator_create needed to have menu->item_height set because it
was 0 for the 1st entry.

Consolatis:
I feel like we should also add a font face for headers so users can for example make headers bold + use a smaller fonts size compared to the other menu items.

droc12345:
We don't handle a pipemenu without a parent (well technically the parent is "openbox_menu", not "current_menu".)
But a pipemenu without a parent should be similar to setting up root-menu/client-menu.

So, I'm thinking of copying init_rootmenu, adding another parm ie title, where the title matches the id= from
the pipemenu definition, and setting up a new menu.

Does this sound like the direction to go?

johanmalm:
> I feel like we should also add a font face for headers so users can for example make headers bold / italic + use a smaller or bigger fonts size compared to the other menu items.

Yes, good point. Will do. 

I'll probably leave this though until we're on 0.18

droc12345:
Ok, got pipemenus straightened out. Leaving out angle brackets (does funny formatting)
- This works 
openbox_menu
...
menu id="test-menu" label="TestMenu"
__menu id="date-menu" label="cal" execute="/home/don/bin/date-menu.sh"

- This works also
openbox_menu
__menu id="date-menu" label="cal" execute="/home/don/bin/date-menu.sh"
,,,
menu id="test-menu" label="TestMenu"
  menu id="date-menu"
/menu


Either way shows the pipe menu, part of it was I can't use duplicate id's
So when I have just a menu id I have to use an odd name (I'll randomize it)

droc12345:
These are the changes, your code for menu and then my changes.

https://github.com/droc12345/labwc/commit/ec82aa9129ce69910cc592913558ef0e5fd77074

https://github.com/droc12345/labwc/commit/ec82aa9129ce69910cc592913558ef0e5fd77074.patch

src/menu/menu.c
droc12345: needed for both separator and pipemenu

src/menu/menu.c
```diff
@@ -265,34 +276,66 @@ separator_create(struct menu *menu, const char *label)
 	struct menuitem *menuitem = znew(*menuitem);
 	menuitem->parent = menu;
 	menuitem->selectable = false;
+	menuitem->type = label ? LAB_MENU_TITLE : LAB_MENU_SEPARATOR_LINE;
 	struct server *server = menu->server;
 	struct theme *theme = server->theme;
-	menuitem->height = theme->menu_separator_line_thickness +
-			2 * theme->menu_separator_padding_height;
```
droc12345, line 271: menuitem->height needs check for being zero, otherwise you get funny results
copy lines 197-200 and you should be fine.

droc12345:
Ok, so far, I have pipemenus working, both with and without parent
also have x,y menu placement (similar to openbox) as well as original atCursor y/n

I have started working on client-list-menu.

droc12345:
From the todo list only menu.title.bg.border.color is not done, as I didn't venture into the border area.
I've documented the stuff I added but not sure if it's enough, when you review my code you can check that out.

Only thing wrong with this code is it needs to pass code check, scripts/check will point that out.
You don't handle the first title properly, but my first set of menu changes takes care of that.

If you make many more changes than what you have already, 
then let me know, so I can rebase/readjust/add/remove.

droc12345:
src/menu/menu.c:281: ERROR: space required before the open brace '{'

src/menu/menu.c
johanmalm: Have just replaced this with `item->selectable` which makes it more readable.

johanmalm:
> I feel like we should also add a font face for headers so users can for example make headers bold / italic + use a smaller or bigger fonts size compared to the other menu items.

@droc12345 is adding that in a subsequent commit

johanmalm:
Ready for review from my perspective.
All comments so far addressed.
I fixed the item-height problem when a title is the first item by simply removing it from the menu struct and making it part of the "global" theme struct instead - which I think is a more correct solution anyway.

johanmalm:
Something weird going with on with the Void check which feel like it could be related to the patch.

droc12345:
Not sure what's going on with void. I used your current code as a base and void didn't complain, compiled fine.

I looked at the output log, but didn't see anything that stood out. and it took 3 minutes, so maybe it was
something to do with their container stuff. ??? 
I just checked with my additions it only took 2 min for the void compile.

johanmalm:
On Void it compiles fine, but the smoke test fails and the gdb stack-trace points to code in the PR. So doesn't feel right to ignore without investigating.

So here goes the investigation:

1. In `scaled-font-buffer.c` `_create_buffer()` we don't initialize `buffer` to `NULL`.
2. ...which becomes a problem because in font_buffer_create() we bail out on string_null_or_empty()
3. ...whereas in `menu.c` `separator_create()` we only NULL-check label.
4. ...and of course in in `init_windowmenu()` we call `separator_create()` with `label=""`. Boom.

So that `buffer` was pointed to garbage on Void (guess on Arch/FreeBSD CI we were just lucky and it pointed to a block of zeroed memory).

Hooray for smoke test.

Right, will sort some patches later.

droc12345:
On my end, I just made sure never to send a "" label, as it didn't parse well.
Inside separator_create, it make a decision on label being null, but sending "" to some font mechanism is wrong.

How does one do anything to a zero length string?
So I just converted "" to a separator line, and if someone wants a blank title size line use " "

johanmalm:
Have force pushed with fixes for the Void-smoke-test-failure. Also added documentation.
It feels like this needs a review from scratch + some testing.

Consolatis:
LGTM

johanmalm:
Thanks.

johanmalm:
@droc12345 
It's ready for your follow PRs now.
I'll try not to touch `menu.c` for a while :smile:

PR #1988, opened 2024-07-15
menu: fix crash triggered by pipemenu without parent <item>

johanmalm:
...such as the one in the example below:

    <openbox_menu>
      <menu id="root-menu" label="obmenu-generator" execute="obmenu-generator"/>
    </openbox_menu>

We should consider supporting this construct in future.

Reported-by: cry0xen via IRC


johanmalm:
Sorry - force pushed to fix a typo in commit message

Consolatis:
LGTM.

This might actually warrant a fixup release (maybe from a new 0.7 branch based on the 0.7.3 tag + backport of this change?) because crashing on user input is kind of bad.

johanmalm:
> This might actually warrant a fixup release (maybe from a new 0.7 branch based on the 0.7.3 tag + backport of this change?) because crashing on user input is kind of bad.

Thanks. Yes, was thinking the same.

Consolatis:
Should we wait until the end of the week for a fixup release? Just in case we receive more crash reports.

Consolatis:
> it did not crash, at least from that. But it also didn't recognize the resulting menu.

Move the pipemenu to the top of all menus so its
```xml
<openbox_menu>
  <menu id="pipe-menu" label="foobar" execute="barfoo" />
  [..]
</openbox_menu>
```
If that doesn't crash for you (and this MR is not applied) I assume you compiled without asserts.

droc12345:
Probably had something to do with having cascade, wlr 018 and original menu patches applied
but not this patch, no asserts disabled, and no crash, but as I say I can't reference by way of a "menu id=" 
inside a regular menu.

Issue #1980, opened 2024-07-13
Magnifier doesn't work with full-screen surfaces (direct scanout?)


jech:
The magnifier is a great feature for us middle-aged people, thanks a lot.

Unfortunately, it doesn't work with Brave (a Chromium variant) running in native Wayland mode ("Preferred Ozone Platform" set to "auto").  When toggling the magnifier over a Brave window, the maginifier keeps zooming in, as though it were magnifying itself.  When disabling the magnifier, the Brave window is corrupted where the magnifier used to reside, until I force a redraw by switching desktops and switching back again.

I've tried both by using the "ToggleMagnify" and "ZoomIn" actions, and the corruption is the same.

Since the magnifier is most useful with badly formatted web pages, this sadly makes it mostly useless for my usage.

johanmalm:
Could you confirm that you're using `wlroots 0.17.4`
There was a bug that affect zoom before that version.

jech:
I've just upgraded to 0.17.4, and the issue is still there.

johanmalm:
> I've just upgraded to 0.17.4, and the issue is still there.

Might be worth doing a re-build + restarting the compositor if you haven't already.

What's in your `~/.config/chromium-flags.conf`?

I've got the following and it's working fine here

```
--enable-features=UseOzonePlatform
--ozone-platform=wayland
```

jech:
> Might be worth doing a re-build + restarting the compositor if you haven't already.

I'm using Debian's pre-built version.  I have restarted after upgrading the shared library.  I can rebuild from sources if you think it's useful.

What's in your ~/.config/chromium-flags.conf?

I have no such file.  However, `chrome://gpu` indicates the following command line:
```
/opt/brave.com/brave/brave --disable-domain-reliability --enable-dom-distiller --enable-distillability-service --origin-trial-public-key=bYUKPJoPnCxeNvu72j4EmPuK7tr1PAC7SHh8ld9Mw3E=,fMS4mpO6buLQ/QMd+zJmxzty/VQ6B1EUZqoCU04zoRU= --sync-url=https://sync-v2.brave.com/v2 --lso-url=https://no-thanks.invalid --variations-server-url=https://variations.brave.com/seed --variations-insecure-server-url=https://variations.brave.com/seed --flag-switches-begin --ozone-platform-hint=auto --flag-switches-end --ozone-platform=wayland --eye-dropper-not-supported --component-updater=url-source=https://go-updater.brave.com/extensions
```

[...]

PR #1989, opened 2024-07-16
magnifier: disable direct scanout when active

Consolatis:
Fixes:
- #1980

Draft while waiting for confirmation that disabling direct scanout fixes the issue.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -256,6 +256,7 @@ struct server {
 	struct seat seat;
 	struct wlr_scene *scene;
 	struct wlr_scene_output_layout *scene_layout;
+	bool direct_scanout_enabled;
 
 	/* cursor interactive */
 	enum input_mode input_mode;
--- a/src/magnifier.c
+++ b/src/magnifier.c
@@ -112,7 +112,6 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box 
 		return;
 	}
 
-	/* Paste the magnified result back into the output buffer */
 	if (!tmp_texture) {
 		tmp_texture = wlr_texture_from_buffer(server->renderer, tmp_buffer);
 	}
@@ -124,7 +123,6 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box 
 	}
 
 	/* Extract source region into temporary buffer */
-
 	struct wlr_render_pass *tmp_render_pass = wlr_renderer_begin_buffer_pass(
 		server->renderer, tmp_buffer, NULL);
 
@@ -198,6 +196,7 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box 
 		dst_box.y = oy - (height / 2);
 	}
 
+	/* Paste the magnified result back into the output buffer */
 	opts = (struct wlr_render_texture_options) {
 		.texture = tmp_texture,
 		.src_box = src_box,
@@ -241,18 +240,21 @@ output_wants_magnification(struct output *output)
 	return output_nearest_to_cursor(output->server) == output;
 }
 
+static void
+enable_magnifier(struct server *server, bool enable)
+{
+	magnify_on = enable;
+	server->scene->direct_scanout = enable ? false
+		: server->direct_scanout_enabled;
+}
+
 /* Toggles magnification on and off */
 void
 magnify_toggle(struct server *server)
 {
+	enable_magnifier(server, !magnify_on);
+
 	struct output *output = output_nearest_to_cursor(server);
-
-	if (magnify_on) {
-		magnify_on = false;
-	} else {
-		magnify_on = true;
-	}
-
 	if (output) {
 		wlr_output_schedule_frame(output->wlr_output);
 	}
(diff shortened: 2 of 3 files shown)


johanmalm:
I haven't looked through `types/scene/wlr_scene.c` properly to see how `scene->direct_scanout` is used, but other than that it seems fine.

Consolatis:
> I haven't looked through `types/scene/wlr_scene.c` properly to see how `scene->direct_scanout` is used, but other than that it seems fine.

At least in 0.17 and 0.18 it is set on `wlr_scene_create()` and then never modified. It is only checked against when deciding if to allow direct scanout or not. As mentioned by @Nefsen402 in https://github.com/labwc/labwc/issues/1980#issuecomment-2231547950 it is marked as private though.

Nefsen402:
The magnifier in labwc sounds like a massive hack to me. It should be implemented by compositing into a staging buffer, then taking that staging buffer and rendering the magnifier. I believe this is possible to do with wlroots without any hacks with 0.18.

Since 0.18, a custom swapchain can be passed to the scene to render into a set of buffers we control and that we will use for creating the magnifier. Here is some pseudo code to achieve this (I'm not going to bother looking up all the exact function names and signatures - I'm just going off of memory):
```c
struct wlr_scene_output *main_scene = ...;

// this output_scene will be replaced with a new scene when the magnifier is active
struct wlr_scene_output *output_scene = main_scene;

if (is_magnifier_used) {
   struct swapchain *schain =  wlr_swapchain_create(output->allocator, output->width, output->height, output->swapchain->format);

   struct wlr_output_state *state;
   wlr_output_state_init(&state);

   wlr_scene_build_output_state(main_scene, &state, &(struct output_options) {
    .swapchain = schain,
  });

  struct wlr_buffer *composited = state->buffer;

  // the composited buffer should then be placed into a secondary scene and the magnifier on top of that.
  output_scene = wlr_scene_create();
  struct wlr_scene_buffer *scene_buffer = wlr_scene_buffer_create(&output_scene->tree);
  wlr_scene_buffer_set_buffer_with_damage(scene_buffer, state->buffer, state->damage);
  labwc_render_maginifier();

  // make sure to free everything
  wlr_output_state_finish(&state);
}

wlr_scene_output_commit(output->scene_output, output_scene);
```

modulo some code to make sure we're not leaking memory and reusing objects like the swapchain.

Basically, how this works is that we have two scenes: Our primary scene that we would normally use like business as usual for the compositor. Then when we activate the magnifier, we create another scene that will handle compositing of the magnifier element after the main compositing has been finished. We use our own custom swapchain to create a place for where staging buffers can live. We can then access the buffer that scene generated for us, and render that into the secondary scene. We can then just use the secondary scene to render the magnifier without worrying about trampling buffer contents we didn't mean to (like with this bug and direct scanout). This should be higher performance as it will never disable direct scanout.

Consolatis:
We need to fix this for wlroots 0.17.x thus we can't rely on anything that requires wlroots 0.18.

Thanks for the pointers, I am completely fine with revisiting the design of the magnifier for labwc 0.8 / wlroots 0.18.
What we currently do is copying the area around the cursor to a temporary wlr_buffer via render_pass, creating a texture out of it and pasting it back with an enlarged dst_box to the output buffer. Then adding the region to the scene output damage ring.

While we do reuse the existing output buffer I would not strictly call this a hack, we are using the wlroots / wlr_scene rendering primitives but missed the direct scanout case that (rightfully I guess) assumes that there can't be any damage on the frame while its active and thus directly maps the client damage to the output damage.

johanmalm:
> Since 0.18, a custom swapchain can be passed to the scene to render into a set of buffers we control and that we will use for creating the magnifier. Here is some pseudo code to achieve this (I'm not going to bother looking up all the exact function names and signatures - I'm just going off of memory):

@Nefsen402 Really appreciate the steer here. That does indeed look a neat way of doing it.

johanmalm:
Patch LGTM. Suggest we merge this to fix the scanout problem.

Thanks for realising what the issue was. I was thinking scanout but when they said "maximized" I stupidly filtered it out being stuck in SSD thinking.

Consolatis:
> It should be implemented by compositing into a staging buffer, then taking that staging buffer and rendering the magnifier. I believe this is possible to do with wlroots without any hacks with 0.18.

What would be the difference (other than one more frame in memory)? The secondary scene would then also need to modify the build_state() buffer (which sounds the same as what we are doing currently) or place another scene node on top for the magnifier (which would disable direct scanout). Do I miss something?

Consolatis:
We do use `wlr_renderer_begin_buffer_pass()` + `wlr_render_pass_add_texture()` with a temporary (but reused within the magnifier) buffer allocated via `wlr_allocator_create_buffer()` with the same format as `state->buffer`. Shouldn't that make it work without CPU copy?

Nefsen402:
Ah, in that case that would be on the GPU. Not a bad solution, also uses less memory. It's just not supported because it violated the ownership model wlroots has for buffers and scene.

Nefsen402:
I wonder if there is a way we can offer a supported way of doing this

PR #1993, opened 2024-07-17
src/output-state.c: add a FPS debug option

Consolatis:
Logs the FPS per output every $timeout ms via wlr_log() if enabled.

By default this patch does nothing unless the FPS_TIMEOUT define has been set, for example via `meson configure build -DFPS_TIMEOUT=5000`.

---

That was added for the chase 0.18 PR, I don't know if its actually useful or not. Feel free to close if not.


Consolatis:
Due to no feedback I assume this isn't really wanted. Thus closing.

Issue #1994, opened 2024-07-17
chromium won't open fully maximized on launch


charbelnicolas:
I just updated to the latest labwc-git build and chrome is now throwing this error when launching, it wasn't happening with wlroots 0.17.

```
00:01:55.849 [ERROR] [types/xdg_shell/wlr_xdg_surface.c:169] A configure is scheduled for an uninitialized xdg_surface 0x5a15eae0d0f0
00:01:55.849 [ERROR] [types/xdg_shell/wlr_xdg_surface.c:169] A configure is scheduled for an uninitialized xdg_surface 0x5a15eae0d0f0
00:01:57.989 [ERROR] [types/xdg_shell/wlr_xdg_surface.c:169] A configure is scheduled for an uninitialized xdg_surface 0x5a15eac47720
00:01:57.989 [ERROR] [types/xdg_shell/wlr_xdg_surface.c:169] A configure is scheduled for an uninitialized xdg_surface 0x5a15eac47720
00:02:07.377 [ERROR] [types/xdg_shell/wlr_xdg_surface.c:169] A configure is scheduled for an uninitialized xdg_surface 0x5a15eae1e6a0
00:02:07.377 [ERROR] [types/xdg_shell/wlr_xdg_surface.c:169] A configure is scheduled for an uninitialized xdg_surface 0x5a15eae1e6a0
```

I noticed this because chromium won't open fully maximized on launch (a small 1/2 inch gap is left at the bottom between the display's border and chromium's border), it is only after a few milliseconds that it will restore the window to its full maximized state.

I don't know who is at fault here, if it's either chromium, labwc or wlroots... but I figured I could start reporting the issue here. THis does not happen with other programs only with chromium.

Consolatis:
That sounds like a labwc issue, quite possible that we missed something.

charbelnicolas:
These are the only rules I have in my config:

```
<windowRules>

	<windowRule identifier="*" serverDecoration="no" fixedPosition="yes" />

	<windowRule identifier="xdg-desktop-portal-gtk" serverDecoration="no" fixedPosition="yes">
		<action name="Maximize" />
	</windowRule>

</windowRules>
```

charbelnicolas:
Could it have to do with server decorations? Seems the gap's height is about the same space that the server decorations would take...

Consolatis:
Could reproduce the issue, it happens with `--start-maximized` so its unrelated to window rules or decorations.
[...]

PR #1995, opened 2024-07-17
[wip] src/xdg.c: prevent configure requests before map

Consolatis:
Fixes:
- #1994

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -279,9 +279,12 @@ handle_request_maximize(struct wl_listener *listener, void *data)
 	if (!view->mapped && !view->output) {
 		view_set_output(view, output_nearest_to_cursor(view->server));
 	}
-	bool maximized = xdg_toplevel_from_view(view)->requested.maximized;
-	view_maximize(view, maximized ? VIEW_AXIS_BOTH : VIEW_AXIS_NONE,
-		/*store_natural_geometry*/ true);
+
+	if (view->mapped) {
+		bool maximized = xdg_toplevel_from_view(view)->requested.maximized;
+		view_maximize(view, maximized ? VIEW_AXIS_BOTH : VIEW_AXIS_NONE,
+			/*store_natural_geometry*/ true);
+	}
 }
 
 static void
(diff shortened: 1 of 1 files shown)


Consolatis:
@jlindgren90 any objections to this PR?

Obviously the logging will be removed.

jlindgren90:
Haven't had time to test. Does e.g. `foot --maximized` still work with this?

jlindgren90:
I'd also be curious if/how this interacts with #1956. I may find time to check later, don't hold things up waiting for me though.

Consolatis:
> Haven't had time to test. Does e.g. `foot --maximized` still work with this?

Yes, it ignores the first maximize request (before mapping) but it is then maximized within the map handler.

I can't reliably test timings and glitches like repositioning currently as my Debian install is too old for compiling the DRM backend and (gles2 renderer) for wlroots 0.18 thus I have to test with pixman and nested only. I didn't verify the startup flow for `foot -m` via `WAYLAND_DEBUG` either.

tokyo4j:
I believe this PR is doing right.

But as I said in https://github.com/labwc/labwc/issues/1994#issuecomment-2235765192, the bug is due to [wlroots change which changes the timing of `new_surface` event](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e8413660686bf94322799470ef5b109ea4580cc9) and [similar change for layer shell](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e8413660686bf94322799470ef5b109ea4580cc9) follows it. So we should make similar fix also for layer shell (not sure it should be done in this PR).

droc12345:
tokyo4j, patch works for me and gets rid of layer-shell message.

No opinion on whether to add to this PR or a new one.

Consolatis:
I can add the layerchange to this PR but #1956 might fix the issue as well. Didn't yet have time to look into that PR in detail.

jlindgren90:
I think this is correct but only half the fix. More complete fix in #2007.

I didn't look into the similar layer-shell warning.

Consolatis:
> I didn't look into the similar layer-shell warning.

I am actually not sure if that whole block should be removed. If the surface isn't mapped it doesn't make sense to arrange it either.
It seems this was introduced waaay back in 2020 via the initial layershell implementation a4c22f7c4de1319e1e252d4a1d65d033fcb93ef2.

tokyo4j:
> I am actually not sure if that whole block should be removed. If the surface isn't mapped it doesn't make sense to arrange it. And the surface should not be mapped in the new_surface handler either.
> 
> It seems this was introduced waaay back in 2020 via the initial layershell implementation https://github.com/labwc/labwc/commit/a4c22f7c4de1319e1e252d4a1d65d033fcb93ef2.

Agreed.

PR #1996, opened 2024-07-18
backend/drm: Fix up tearing for atomic modesetting

kode54:
Remove the NO_ATOMIC check, and add a conditional fallback for errors, which usually result from properties being applied during a flip, such as hardware cursor image changes. Falling back is limited so that persistent errors do not keep immediate flips running forever, but instead disable them until an output is reset. Verified working at least on AMD Radeon 6700 XT.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -186,10 +186,13 @@ this is for compatibility with Openbox.
 	mode.
 
 *<core><allowTearing>* [yes|no]
-	Allow tearing to reduce input lag. Default is no.
-	This option requires setting the environment variable
-	WLR_DRM_NO_ATOMIC=1.
-	*yes* allow tearing if requested by the active window.
+	Allow tearing, if requested by the active window, to reduce input lag.
+	Default is no.
+
+	Note: Enabling this option with atomic mode setting is experimental. If
+	you experience undesirable side effects when tearing is allowed,
+	consider setting the environment variable WLR_DRM_NO_ATOMIC=1 when
+	launching labwc.
 
 *<core><reuseOutputMode>* [yes|no]
 	Try to re-use the existing output mode (resolution / refresh rate).
--- a/src/output.c
+++ b/src/output.c
@@ -30,6 +30,14 @@
 #include "view.h"
 #include "xwayland.h"
 
+static unsigned int
+get_tearing_retry_count(struct output *output)
+{
+	/* Two seconds worth of frames, guessing 60Hz if refresh is invalid */
+	int refresh = output->wlr_output->refresh;
+	return refresh > 0 ? refresh / 500 : 120;
+}
+
 static bool
 get_tearing_preference(struct output *output)
 {
@@ -42,6 +50,11 @@ get_tearing_preference(struct output *output)
 
 	/* Tearing is only allowed for the output with the active view */
 	if (!server->active_view || server->active_view->output != output) {
+		return false;
+	}
+
+	/* Tearing should not have failed too many times */
+	if (output->nr_tearing_failures >= get_tearing_retry_count(output)) {
 		return false;
 	}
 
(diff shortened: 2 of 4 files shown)


src/output.c
```diff
@@ -113,10 +113,23 @@ output_frame_notify(struct wl_listener *listener, void *data)
...
 		output->pending.tearing_page_flip =
-			get_tearing_preference(output);
-
-		lab_wlr_scene_output_commit(output->scene_output,
-			&output->pending);
+			get_tearing_preference(output) &&
+			output->tearing_fail_count < 120;
+
+		if (!lab_wlr_scene_output_commit(output->scene_output,
+			&output->pending) && output->pending.tearing_page_flip) {
```
Consolatis, line 120: Second line of the if condition should be double indented to stand out from the code below.

src/output.c
```diff
@@ -113,10 +113,23 @@ output_frame_notify(struct wl_listener *listener, void *data)
...
+					" frames, disabling");
+			}
+			output->pending.tearing_page_flip = false;
+			lab_wlr_scene_output_commit(output->scene_output,
+				&output->pending);
+		} else {
+			if (output->pending.tearing_page_flip) {
+				output->tearing_fail_count = 0;
+			}
+		}
```
Consolatis, line 132: Could written as
```c
} else if (output->pending.tearing_page_flip) {
	output->tearing_fail_count = 0;
}

**Comment by Consolatis** [comment]
Thanks for the PR.

LGTM, not tested.
Not merging right away to allow for further feedback.

File: src/output.c [hunk 3]
```diff
@@ -113,10 +116,21 @@ output_frame_notify(struct wl_listener *listener, void *data)
...
+			if (++output->tearing_fail_count >= TEARING_FAIL_LIMIT) {
+				wlr_log(WLR_INFO, "Tearing failed for %u consecutive"
+					" frames, disabling", (unsigned int)TEARING_FAIL_LIMIT);
+			}
+			output->pending.tearing_page_flip = false;
+			lab_wlr_scene_output_commit(output->scene_output,
+				&output->pending);
+		} else if (output->pending.tearing_page_flip) {
+			output->tearing_fail_count = 0;
+		}
```
ahesford, line 133: I think restructing this makes the logic a bit easier to read through (the setting of `scene_output` could be moved further up and used in the sanity check at the top of the function):
```suggestion
		struct wlr_scene_output *scene_output = output->scene_output;
		struct wlr_output_state *pending = &output->pending;
		
		bool committed =
			lab_wlr_scene_output_commit(scene_output, pending);
		
		if (!pending->tearing_page_flip) {
			return;
		}
		
		if (committed) {
			output->tearing_fail_count = 0;
		} else {
			if (++output->tearing_fail_count >= TEARING_FAIL_LIMIT) {
				wlr_log(WLR_INFO, "setting tearing allowance failed "
					"for %u consecutive frames, disabling",
					(unsigned int)TEARING_FAIL_LIMIT);
			}
			pending->tearing_page_flip = false;
			lab_wlr_scene_output_commit(scene_output, pending);
		}
```

ahesford (changes requested):
I don't think we should expose a new configuration option for this. It's awkward, confusing, and will be obsolete whenever we (or wlroots, or whoever is the authority here) considers tearing allowance with atomic modesetting non-experimental. At best, we should note in the manual page (and perhaps in a comment in rc.xml.all) that enabling this option without setting `WLR_DRM_NO_ATOMIC` is experimental and not yet recommended for "production". That has the added benefit of simplifying the code, because we can drop the checks entirely instead of gating them behind another option. (Alternatively, we can log a warning if `WLR_DRM_NO_ATOMIC` is not set when tearing is enabled, and then remove the warning at some point in the future.)

docs/labwc-config.5.scd
```diff
@@ -183,8 +183,10 @@ this is for compatibility with Openbox.
 
 *<core><allowTearing>* [yes|no]
 	Allow tearing to reduce input lag. Default is no.
-	This option requires setting the environment variable
-	WLR_DRM_NO_ATOMIC=1.
+	Use of this variable without setting the environment variable
+	WLR_DRM_NO_ATOMIC=1 is considered experimental at this time, and
+	may have minor issues if used in combination with hardware cursors,
+	though none were observed yet as a noticeable problem in testing.
 	*yes* allow tearing if requested by the active window.
```
ahesford, line 190: It seems the existing description was not properly edited when we added it. Please drop the final sentence, and maybe we can refine the description for clarity:

    Allow tearing, if requested by the active window, to reduce input lag.
    Default is no.
    
    Note: enabling this option with atomic mode setting is experimental. If
    you experience undesirable side effects when tearing is allowed,
    consider setting the environment variable WLR_DRM_NO_ATOMIC=1 when
    launching labwc.

src/output.c
```diff
@@ -30,6 +30,14 @@
 #include "view.h"
 #include "xwayland.h"
 
+static unsigned int
+get_tearing_retry_count(struct output *output)
+{
+	/* Two seconds worth of frames, guessing 60Hz if refresh is zero */
+	unsigned int refresh = output->wlr_output->refresh;
+	return refresh ? refresh / 500 : 120;
```
ahesford, line 38: In the interest of Postel's law:
```suggestion
	return refresh > 0 ? refresh / 500 : 120;
```

ahesford (changes requested):
I think this is looking good. With a nitpick and some trivial documentation cleanup, I think this will be ready to go.

ahesford (approved):

ahesford:
Thanks!

PR #1997, opened 2024-07-18
backend/drm: Implement support for renderer loss recovery

kode54:
This implementation is nearly identical to Sway's, except that it also reloads the configuration, to spur on reloading the server-side decorations.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -318,6 +318,8 @@ struct server {
 	 */
 	int pending_output_layout_change;
 
+	struct wl_listener renderer_lost;
+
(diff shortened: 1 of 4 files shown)


Consolatis:
See CodeStyleCheck output:
> src/server.c:260: CHECK: Comparison to NULL could be written "!renderer"
> src/server.c:267: CHECK: Comparison to NULL could be written "!allocator"

I think we also need some way to reset the magnifier buffer if it exists. Not quite sure yet on how to do that properly, maybe we need some `magnifier_reset_buffer()` call and move the buffer from a function static to a global within `magnifier.c`.

Consolatis:
Thanks for your work on this.

PR #1998, opened 2024-07-18
config: initialize some missing defaults

kode54:
Add some defaults which were missing from initialization.


Consolatis:
Thanks!

PR #2006, opened 2024-07-19
Pipemenu, fix so both inline and stand alone work

droc12345:
This is based on Johanmalm's wip menu patches with changes to make pipemenus work 
in either inline or standalone mode.


Consolatis:
Please don't mix those two commits into one.

If you want to base this on top of @johanmalm branch please do so via cherry-pick of your single commit change on top of the malms branch. That way we can drop it here before merging.

Another option is to simply base your changes on top of master as they are unrelated to the header implementation.

PR #2007, opened 2024-07-19
xdg: update initial maximize logic for wlroots 0.18

### LINKED ISSUE #1994: chromium won't open fully maximized on launch

jlindgren90:
The initial configure event is now sent explicitly by labwc rather than by wlroots. We need to move the maximize/fullscreen logic to the initial commit handling accordingly.

Updates #1956, fixes #1994, replaces #1995.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -651,10 +651,6 @@ xdg_toplevel_view_map(struct view *view)
 	if (!view->been_mapped) {
 		init_foreign_toplevel(view);
 
-		/*
-		 * FIXME: is this needed or is the earlier logic in
-		 * xdg_surface_new() enough?
-		 */
 		if (view_wants_decorations(view)) {
 			view_set_ssd_mode(view, LAB_SSD_MODE_FULL);
 		} else {
@@ -907,13 +903,6 @@ xdg_toplevel_new(struct wl_listener *listener, void *data)
 	CONNECT_SIGNAL(xdg_surface, xdg_toplevel_view, new_popup);
 
 	wl_list_insert(&server->views, &view->link);
-
-	/* FIXME: is view_wants_decorations() reliable this early? */
-	if (view_wants_decorations(view)) {
-		view_set_ssd_mode(view, LAB_SSD_MODE_FULL);
-	} else {
-		view_set_ssd_mode(view, LAB_SSD_MODE_NONE);
-	}
 }
 
 void
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -76,6 +76,18 @@ handle_new_popup(struct wl_listener *listener, void *data)
 }
 
 static void
+set_fullscreen_from_request(struct view *view,
+		struct wlr_xdg_toplevel_requested *requested)
+{
+	if (!view->fullscreen && requested->fullscreen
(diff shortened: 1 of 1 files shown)


jlindgren90:
Maybe the `view_wants_decorations()` check should be moved too.

Consolatis:
This prevents setting the output before map, intended?
From memory that is required for scale notifications to work before map but I am not sure right now. @ahesford might be able to confirm.

jlindgren90:
> This prevents setting the output before map, intended?

There is still this code in `xdg_toplevel_new()`:

	/*
	 * Pick an output for the surface as soon as its created, so that the
	 * client can be notified about any fractional scale before it is given
	 * the chance to configure itself (and possibly pick its dimensions).
	 */
	view_set_output(view, output_nearest_to_cursor(server));
	if (view->output) {
		wlr_fractional_scale_v1_notify_scale(xdg_surface->surface,
			view->output->wlr_output->scale);
	}

Does that address the concern?

ahesford:
I think the assignment of an output in `xdg_toplevel_new`, which was added to make sure foot sets its initial size based on a correct output scale, is probably sufficient even in cases when a view is initially maximized.

Consolatis:
In that case I think we can remove the `view_set_output()` calls in the fullscreen and maximized handlers as output will always be set via `xdg_toplevel_new()`.

jlindgren90:
I agree this later assignments of `view->output` are probably a bit paranoid. I think it's a bit out of scope for this PR to remove them so I think I'll leave them as-is for now.

Consolatis:
Alright, PR LGTM, not tested. Feel free to merge if you are satisfied with it.

Issue #1621, opened 2024-03-14
Dragging a tiled window in nested session causes flickering


tokyo4j:
This issue is not visible in bare metal. 

Demo:

https://github.com/labwc/labwc/assets/22029524/9a4c5ff2-c633-4cb5-a0b7-844787138153

After some debugging, I found the cause is that `view->pending.height` is used to calculate the final position of the view in the process of applying resistance at the first call to `process_cursor_move()` after `interactive_begin()` stores `view->natural_geometry` in `view->pending`.

A small change below fixed the issue. Would it cause any side effects, @ahesford?

```diff
diff --git a/src/edges.c b/src/edges.c
index 7253569..a3f8db5 100644
--- a/src/edges.c
+++ b/src/edges.c
@@ -530,7 +530,7 @@ edges_adjust_move_coords(struct view *view, struct border edges,
 	if (view_geom->y != *y) {
 		int tshift = border.top + rc.gap;
 		int bshift = border.bottom + rc.gap
-			+ view_effective_height(view, /* use_pending */ true);
+			+ view_effective_height(view, use_pending);
 
 		adjust_move_coords_1d(y, edges.top, tshift,
 			edges.bottom, bshift, *y < view_geom->y);
```

ahesford:
Nice find! I think your fix is correct.

PR #2009, opened 2024-07-20
Implement resistance for tiled windows

tokyo4j:
Closes #1968.

This PR adds `<snapping><dragResistance>` which configures the distance of cursor movement for dragging a tiled/maximized window.

I found dragging an initially maximized window places the window at a bit incorrect position. Not sure I can fix it in this PR.

~This PR also temporarily includes the commit from #2010.~

PR #2010, opened 2024-07-20
edges.c: fix flicker of snapped windows in nested session

tokyo4j:
Fixes: #1621

I haven't made a PR for this because I don't understand how resistance works and the issue only happened when I drag a window tiled to the bottom edge in nested session, but now the issue is more apparent while testing #2009.

CC @ahesford


ahesford:
Thanks!

PR #2016, opened 2024-07-20
src/xdg-popup.c: choose output depending on xdg-positioner

### LINKED ISSUE #2012: Application menus open in incorrect location when portion of window overlaps multiple outputs
I noticed that if I have an application window in which part of the window is on a second display, any in-application menus that are opened, open in the wrong location, on the wrong monitor.

Since it's hard to see it with screencap, I've attached a video recorded with my smartphone

I've confirmed this is the case on all applications, GTK and Qt applications, running in native Wayland (not XWayland, though XWayland apps like Steam exhibit similar behaviors)

Arch Linux, 3 monitors (tie-fighter layout, portraits on the side with a landscape in the middle), latest git build of labwc

https://github.com/user-attachments/assets/22610800-ee4c-4e5f-a189-5818df38299b

tokyo4j:
Fixes #2012

`wlr_popup->current.geometry.{x,y}` are usually zero on initial commit, so xdg-popups were always unconstrained against the output which contains the top-left of the parent toplevel.

This PR changes xdg-popups to be unconstrained against the output which contains the top-left of preliminary popup geometry designated by xdg-positioner given as an argument of "get_popup" or "reposition" requests.

EDIT:
~Currently this doesn't work correctly with nested popups.~

EDIT2:
Now correct outputs are chosen also for nested popups.
Also added  a fix for #2017.

### DIFF
--- a/src/xdg-popup.c
+++ b/src/xdg-popup.c
@@ -28,19 +28,33 @@ popup_unconstrain(struct xdg_popup *popup)
 	struct view *view = popup->parent_view;
 	struct server *server = view->server;
 
+	/* Get position of parent toplevel/popup */
 	int parent_lx, parent_ly;
 	struct wlr_scene_tree *parent_tree = popup->wlr_popup->parent->data;
 	wlr_scene_node_coords(&parent_tree->node, &parent_lx, &parent_ly);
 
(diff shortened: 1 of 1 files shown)


Consolatis:
:)

PR #2020, opened 2024-07-23
src/ssd/ssd.c: fix incorrect condition in `ssd_part_contains()`

tokyo4j:


tokyo4j:
Sorry I should've add some comment as this subtly changes the behavior.

The comment in the release note would be like:
```
Fixed mousebinds with "Top" context being incorrectly applied on bottom-left corner of a window. 
```

Consolatis:
Wow, that one must have been there for quite a while. Thanks!

PR #2021, opened 2024-07-23
layer-shell: stop sending configure events on surface creation

tokyo4j:
With wlroots 0.18, layer-shell's `new_surface` event is emitted on `zwlr_layer_shell_v1.get_layer_surface` request rather than the first commit. This change layer-shell clients like `fuzzel` to flicker on launch because labwc was sending a configure event with fullscreen geometry due to the absence of geometry hints on `get_layer_surface` requests.

This PR removes the code that updates the usable area from `new_surface` handler, preventing unintended configure events with fullscreen geometry.

Suggested by https://github.com/labwc/labwc/pull/1995#issuecomment-2240718847


Consolatis:
Thanks :)

PR #2024, opened 2024-07-24
multirect: remove destroy listener before freeing

Consolatis:
Detected by `-Db_sanitize=address,undefined` for libwayland


johanmalm:
LGTM
Good catch

PR #2030, opened 2024-07-26
Initial implementation of cosmic-workspace-unstable-v1

Consolatis:
I finally have decided to not wait another 3 years until the [official workspace protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40) is accepted. This goes very much against my own believes that labwc should only use wlroots or wayland protocols but version 1 of the [cosmic-workspaces protocol](https://github.com/pop-os/cosmic-protocols/blob/main/unstable/cosmic-workspace-unstable-v1.xml) is basically the same as what is proposed in the upstream protocol. Using the cosmic variant also allows using the protocol without hiding it behind feature flags as the upstream protocol may change in incompatible ways before being accepted.

This PR is an initial implementation of the protocol, mostly written in a compositor agnostic way so there is the potential to also propose this implementation for wlroots with minor changes at some point (e.g. mostly `znew()` vs `calloc()` and `wl_list_append()` vs `wl_list_insert(x.prev, y)`).

The integration part of the protocol into labwc is purposefully minimal, it only tells the client about the workspaces available, their state + name and allows the client to activate them. The protocol theoretically also allows for clients to configure the workspaces and the protocol implementation allows for that but the labwc integration does not.

Draft PR for sfwbar and Waybar to have something for testing:
- https://github.com/LBCrion/sfwbar/pull/212
- https://github.com/Alexays/Waybar/pull/3481

<details><summary>Test config for Waybar</summary>

`~/.config/waybar/config`
```json
    "cosmic/workspaces": {
        "format": "{name}",
        "on-click": "activate",
        "all-outputs": true,
    },
```

`~/.config/waybar/style.css` (modified the existing `button:hover`)
```css
button:hover,
#workspaces button.active {
    background: inherit;
    box-shadow: inset 0 -3px #ffffff;
}
```

</details>


tokyo4j:
I don't have any strong preference here since I usually use only one workspace and the integration part is minimal as you said, but what will we do when the `ext-workspace-unstable` is finally merged and wlroots adds implementation for it? Will we temporarily support both of `cosmic-workspace-unstable-v1` and `ext-workspace-unstable` for compatibility? (just for question)

Consolatis:
> what will we do when the `ext-workspace-unstable` is finally merged and wlroots adds implementation for it? Will we temporarily support both of `cosmic-workspace-unstable-v1` and `ext-workspace-unstable` for compatibility? (just for question)

Yes, depending on the availability of clients that would then support ext-workspaces the idea would be to support both for labwc.

Consolatis:
Waybar PR:  https://github.com/Alexays/Waybar/pull/3481

01micko:
Test run, seems to work fine - screencap, is this what I'm supposed to be looking at? (methinks yes)
cosmic

Back on my Slack-current install with updated latest stuff so it builds nice :)

Consolatis:
> is this what I'm supposed to be looking at? (methinks yes)

Yes, the 1-4 on the left look like the pinned workspaces configured in the sfwbar config by default. They can be removed by commenting out the `pins` line in the `pager` section.

01micko:
> > is this what I'm supposed to be looking at? (methinks yes)
> 
> Yes, at least with the stock `sfwbar.config` or a derivation of it. The 1-4 on the left look like the pinned workspaces configured in the sfwbar config by default. They can be removed by commenting out the `pins` line in the `pager` section.

Yeah I figured I had to use stock first up, so just renamed my user config. I'll play with it some more later.

jlindgren90:
There is also https://wayland.app/protocols/kde-plasma-virtual-desktop - not sure of the pros/cons of following Cosmic vs. KDE here.

stefonarch:
> There is also https://wayland.app/protocols/kde-plasma-virtual-desktop - not sure of the pros/cons of following Cosmic vs. KDE here.

The lxqt-panel will use this one if the compositor is kwin_wayland. Desktop pager and "Move to Desktop x" in  the windows button right click menu of the taskbar are working fine with it.

Consolatis:
> There is also https://wayland.app/protocols/kde-plasma-virtual-desktop - not sure of the pros/cons of following Cosmic vs. KDE here.

From a quick look:
- The cosmic variant is an older version of the protocol submitted at wayland-protocols and should be easy to adjust to the official one once merged
- The KDE one looks completely independent and also way less flexible. E.g. it doesn't seem  to notify about any output <--> workspace relation, doesn't allow the compositor to send independent workspaces per output nor allow a client to change the workspaces on a defined output. It also doesn't provide any form of capability systems so panel can disable options not supported by the compositor. The upstream + cosmic variant allow for output independent workspaces via the `group` concept and has a capability system.

jlindgren90:
If the cosmic protocol is closer than the upstream MR, then that's reason enough, you've convinced me :)

My only other thought: there was a wlr protocol proposed 5(!) years ago, https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/35. It was closed in favor of the upstream wayland-protocols MR, and there's been no discussion on it recently. Would it be worth checking with the wlroots folks if it could be revived, given that the upstream MR seems to have stalled?

I do realize wlroots development also happens slowly... if there is no interest there, then I'm fine with adopting the cosmic namespace variant. But it might be worth at least asking.

Consolatis:
> My only other thought: there was a wlr protocol proposed 5(!) years ago, https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/35. It was closed in favor of the upstream wayland-protocols MR, and there's been no discussion on it recently. Would it be worth checking with the wlroots folks if it could be revived, given that the upstream MR seems to have stalled?

That is an ever older variant of the current upstream MR (and thus also an older version of the cosmic vendored one). I don't think we want yet another version of the same protocol floating around. My hope is that either the upstream one is merged within the next year or if that is not the case that wlroots will accept an implementation of the cosmic variant instead (e.g. for example this one that could be slightly adopted for wlroots). In the meantime, compositors and panels can add support for the cosmic vendored one which can be very easily adapted for the upstream MR once merged (as of now I think it only replaced the wl_array capabilities with a proper single uint32_t bitset).
 
> I do realize wlroots development also happens slowly... if there is no interest there, then I'm fine with adopting the cosmic namespace variant. But it might be worth at least asking.

As both cosmic and wlroots are members of the wayland-protocols governance body and protocols in the ext- namespace only need 2 ACKS from members I do not think that wlroots would accept a new wlr vendored protocol. At that point it would make way more sense to get the upstream one merged. See also https://gitlab.freedesktop.org/wlroots/wlr-protocols#submitting-new-protocols.

jlindgren90:
> My hope is that either the upstream one is merged within the next year or if that is not the case that wlroots will accept an implementation of the cosmic variant instead (e.g. for example this one that could be slightly adopted for wlroots).

Okay, this makes sense... a wlroots implementation would continue to use the "cosmic-" protocol namespace then (until the "ext-" version is merged). I am good with this approach.

src/protocols/cosmic_workspaces/cosmic-workspaces.c
```diff
@@ -0,0 +1,675 @@
...
+#include <wlr/util/log.h>
+#include "common/mem.h"
+#include "common/list.h"
+#include "cosmic-workspace-unstable-v1-protocol.h"
+#include "protocols/cosmic-workspaces.h"
+#include "protocols/cosmic-workspaces-internal.h"
+
+/*
+ *	.--------------------.
+ *	|        TODO        |
```
johanmalm, line 13: Should this `TODO` still be here?

src/protocols/cosmic_workspaces/cosmic-workspaces.c
```diff
@@ -0,0 +1,675 @@
...
+	CW_WS_STATE_HIDDEN  = 1 << 2,
+
+	/*
+	 * Set when creating a new workspace so we
+	 * don't end up having to send the state twice.
+	 */
+	CW_WS_STATE_INVALID = 1 << 31,
+};
+
+#define wl_arr_add(_arr, _val) (*(__typeof__(_val) *)wl_array_add( \
```
johanmalm, line 54: It's not obvious to me what `wl_arr_add()` does. Should we add to `include/common/array.h` and add a comment to describe it.

src/workspaces.c
```diff
@@ -260,7 +292,17 @@ _osd_show(struct server *server)
 void
 workspaces_init(struct server *server)
 {
-	wl_list_init(&server->workspaces);
+	server->workspaces.cosmic_manager = wlr_cosmic_workspace_manager_create(
+		server->wl_display, CW_CAP_WS_ACTIVATE | CW_CAP_GRP_WS_CREATE, 1);
```
johanmalm, line 296: In `workspaces_init()` using `1` as last argument in `wlr_cosmic_workspace_manager_create()` is not easy to read. Suggest a `#define` (is it COSMIC_WORKSPACE_V1_VERSION?)

src/workspaces.c
```diff
@@ -260,7 +292,17 @@ _osd_show(struct server *server)
 void
 workspaces_init(struct server *server)
 {
-	wl_list_init(&server->workspaces);
+	server->workspaces.cosmic_manager = wlr_cosmic_workspace_manager_create(
```
johanmalm, line 295: I'm sure it's right, but why do we pass `CW_CAP_WS_ACTIVATE` in `wlr_cosmic_workspace_manager_create()`

src/workspaces.c
```diff
@@ -172,6 +173,28 @@ _osd_update(struct server *server)
 	}
 }
 
+/* cosmic workspace handlers */
+static void
+handle_workspace_create(struct wl_listener *listener, void *data)
```
johanmalm, line 178: Consider `s/handle_workspace_create/handle_cosmic_workspace_create/`

src/workspaces.c
```diff
@@ -172,6 +173,28 @@ _osd_update(struct server *server)
 	}
 }
 
+/* cosmic workspace handlers */
+static void
```
johanmalm, line 177: Consider creating a separate `workspace_create()` which `handle_workspace_create()` can call. Makes it easier to share code

src/protocols/cosmic_workspaces/transactions.c
```diff
@@ -0,0 +1,89 @@
...
+		ctx = znew(*ctx);
+		wl_list_init(&ctx->transactions);
+	}
+	addon->ctx = ctx;
+	addon->ctx->ref_count++;
+	return addon;
+}
+
+void
+add_ws_transaction(struct wlr_cosmic_workspace *ws,
```
johanmalm, line 58: Feels like we should start public function names with `transaction_`

johanmalm (commented):
Have left a bunch of review comments - but other than those it LGTM. Not tested.

A couple of general points:

- Should we prefix with `wlr_cosmic_`? Couldn't we just do `cosmic`. For example `wlr_cosmic_workspace_group_output_enter()`
- We ought to consider ways to not lose created workspace on Reconfigure. Can do that on separate PR though if we can think of something.

src/protocols/cosmic_workspaces/transactions.c
Consolatis: Those are cosmic internal but can change to transaction prefix.

src/workspaces.c
Consolatis: That commit (adding support for clients to create workspaces) is supposed to be deleted before we merge, it was only created to aid in testing of the sfwbar implementation. Most of the reason for me not wanting to implement that part in labwc is that it suddenly shares the responsibility of who configures workspaces with external clients and thus rc.xml doesn't show the full picture. We could consider allowing clients to create / remove workspaces at a later point though, the protocol side of the implementation is already there.

src/workspaces.c
Consolatis: removed completely

src/workspaces.c
Consolatis: added

src/protocols/cosmic_workspaces/cosmic-workspaces.c
Consolatis: Created a new helper in `common/array.h`: `xwl_array_add()` (in the style of `xstrdup()`) which asserts against memory allocation errors.

src/protocols/cosmic_workspaces/cosmic-workspaces.c
Consolatis: Will work on the remaining items tomorrow and then remove the TODO but none of them should cause any issues.

Consolatis:
> Have left a bunch of review comments - but other than those it LGTM. Not tested.

Thanks.

> * Should we prefix with `wlr_cosmic_`? Couldn't we just do `cosmic`. For example `wlr_cosmic_workspace_group_output_enter()`

I've changed everything to a `lab_cosmic_` prefix, the `wlr_` one was intended to upstream the protocol implementation at some point but that is really easy to fix with some `sed 's/lab_cosmic/wlr_cosmic/g'`.

> * We ought to consider ways to not lose created workspace on Reconfigure. Can do that on separate PR though if we can think of something.

I've removed the feature for clients to create workspaces completely as its unclear who is in control of the final workspace configuration. We could consider adding that back at some point (it just requires reacting to two signals).

include/common/array.h
```diff
@@ -41,4 +41,18 @@ wl_array_len(struct wl_array *array)
 		pos && (const char *)pos >= (const char *)(array)->data;               \
 		(pos)--)
 
+/**
+ * xwl_array_add() - add item to array and crash on allocation error
+ * @_arr: wl_array to add the item to
+ * @_val: the item to add to the array
+ * Note: asserts() when the memory allocation for the new item failed
+ */
+#define xwl_array_add(_arr, _val) ({                       \
```
johanmalm, line 50: Now looks like ({}) GNU statement expression.

Are we sure we want that? I don’t personally mind them, but it does move us into a space some people resist.

Consolatis: What do you suggest instead, a `do { .. } while (0)` loop?

include/common/array.h
```diff
@@ -41,4 +41,17 @@ wl_array_len(struct wl_array *array)
 		pos && (const char *)pos >= (const char *)(array)->data;               \
 		(pos)--)
 
+/**
+ * xwl_array_add() - add item to array and crash on allocation error
+ * @_arr: wl_array to add the item to
+ * @_val: the item to add to the array
+ * Note: asserts() when the memory allocation for the new item failed
+ */
+#define xwl_array_add(_arr, _val) do {                     \
```
johanmalm, line 50: This is indeed a nice wrapper. I now get it :smile:

Suggest that we avoid `assert()` and just `exit()` with an error message.

Suggest that we call it `array_add()` becase:
1. `xwl` makes me think of XWayland
2. It actually does more than wl_array_add() + die-on-null

Also suggest adding the following to the comment in the header to make it easier to wrap one's head around it:

```
Let us illustrate the function of this macro by an example:

    uint32_t value = 5;
    array_add(array, value);

...is the equivalent of the code below which is how you would
otherwise use the wl_array API:

uint32_t *elm = wl_array_add(array, sizeof(uint32_t));
    if (!elm) {
    	perror("failed to allocate memory");
    	exit(EXIT_FAILURE);
    }
    *elm = value;
```

johanmalm:
Have tested with waybar. Works well.

LGTM other than the suggestion I just left on the array-macro.

include/common/array.h
Consolatis: Renamed to `array_add()`, replaced the assert with `exit()` and added your comment.

g7fya:
hello, installed labwc and sfwbar from Consolatis repositorie, work  great. However, is there a way where  sfwbar taskbar to show only windows from the current workspace?

Consolatis:
> is there a way where sfwbar taskbar to show only windows from the current workspace?

No, that requires a further protocol to match windows with workspaces. This one only supports controlling the workspaces themselves.

LBCrion:
> No, that requires a further protocol to match windows with workspaces. This one only supports controlling the workspaces themselves.

We could implement cosmic-toplevel-info-unstable and cosmic-toplevel-management-unstable with a fallback to wlr-foreign-top-level in theory.  Although I suspect that's out of scope for now ;)

Consolatis:
> > No, that requires a further protocol to match windows with workspaces. This one only supports controlling the workspaces themselves.
> 
> We could implement cosmic-toplevel-info-unstable and cosmic-toplevel-management-unstable with a fallback to wlr-foreign-top-level in theory. Although I suspect that's out of scope for now ;)

Kind of, it also becomes really messy once (and if) the upstream protocols are merged. Then we'd have
- wlr-foreign-toplevel
- ext-foreign-toplevel-list
- ext-foreign-toplevel-info
- ext-foreign-toplevel-management
- cosmic-foreign-toplevel-info
- cosmic-foreign-toplevel-management

I am already quite happy when we can finally switch workspaces via panel.

LBCrion:
Don't forget the yet-to-be-named ext protocols for toplevel-workspace
interaction (probably two of them as well? Info and management), since
ext-topvel-* and ext-workspace are currently completely unaware of each
other and the plan it to have bridging protocols linking them.

Conan-Kudo:
> > There is also https://wayland.app/protocols/kde-plasma-virtual-desktop - not sure of the pros/cons of following Cosmic vs. KDE here.
> 
> The lxqt-panel will use this one if the compositor is kwin_wayland. Desktop pager and "Move to Desktop x" in the windows button right click menu of the taskbar are working fine with it.

I don't know this for sure, I think @zamundaaa would know for sure, but I think the goal is to support `ext-workspace` in KWin.

Zamundaaa:
> I think the goal is to support ext-workspace in KWin.

At some point, yes.

01micko:
Wow! Excellent!

PR #2032, opened 2024-07-27
Update meson.build

tfemby:
This is a minor thing but when running `labwc --version` on the git version of labwc, I'm currently getting `labwc 0.7.3-52-g750d37b-dirty`.

I think that with the 0.7.4 release, changing meson.build was missed on the main branch. It seems that the 0.7.4 tag has the correct version numbering. Upon applying this commit, the output of `labwc --version` is now `labwc 0.7.4`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
   'labwc',
   'c',
-  version: '0.7.3',
+  version: '0.7.4',
   license: 'GPL-2.0-only',
   meson_version: '>=0.59.0',
   default_options: [


stefonarch:
I wondered where the dirty came from, but here I see it only when I apply a patch manually, otherwise I get `labwc 0.7.3-50-ga7024dd`.

ahesford:
What you're seeing is the output of `git describe`, and it means this was built on a branch that is 52 commits ahead of the 0.7.3 tag. The 0.7.4 tag is not on the master branch; it is on the v0.7 branch maintained for cutting stable releases against wlroots 0.17 until we decide to tag a release against wlroots 0.18. The "dirty" means you've included some uncommitted changes.

Updating the meson project will not fix the reported version you're seeing.

tfemby:
> I wonder where the `-dirty` comes from, but here I see it only when I apply a patch manually, otherwise I get `labwc 0.7.3-50-ga7024dd`.

Pretty much what's happening with my situation. I'm building on gentoo using an overlay which has a patch for doc installation path so it adheres with gentoo guidelines. Said patch isn't my work but if you're curious, [here it is](https://github.com/bsd-ac/wayland-desktop/blob/master/gui-wm/labwc/files/labwc-0.7.4-meson_doc_path.patch).



> What you're seeing is the output of `git describe`, and it means this was built on a branch that is 52 commits ahead of the 0.7.3 tag. The 0.7.4 tag is not on the master branch; it is on the v0.7 branch maintained for cutting stable releases against wlroots 0.17 until we decide to tag a release against wlroots 0.18. The "dirty" means you've included some uncommitted changes.
> 
> Updating the meson project will not fix the reported version you're seeing.

I wasn't completely aware and I appreciate the explainer. Since what I submitted won't achieve the desired outcome, shall I close the PR?

Consolatis:
I think wlroots uses a `-dev` suffix for the version number and then drops it when tagging. So we could use `0.8.0-dev` in meson.build (but that won't help with git describe as mentioned above).

What you could do however is doing something like this:
```diff
diff --git a/meson.build b/meson.build
index 267b91b3..28453da3 100644
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project(
   'labwc',
   'c',
-  version: '0.7.3',
+  version: '0.8.0-dev',
   license: 'GPL-2.0-only',
   meson_version: '>=0.59.0',
   default_options: [
@@ -43,7 +43,7 @@ git = find_program('git', native: true, required: false)
 if git.found()
   git_commit = run_command([git, 'describe', '--dirty'], check: false)
   if git_commit.returncode() == 0
-    version = '"@0@"'.format(git_commit.stdout().strip())
+    version = '"@0@ (@1@)"'.format(meson.project_version(), git_commit.stdout().strip())
   endif
 endif
 add_project_arguments('-DLABWC_VERSION=@0@'.format(version), language: 'c')
```

tfemby:
> I think wlroots uses a `-dev` suffix for the version number and then drops it when tagging. So we could use `0.8.0-dev` in meson.build (but that won't help with git describe as mentioned above).

That's what I see when building lab. Meson finds (for example) `wlroots-0.8.0-dev`.

> What we could do however is something like this:
> 
> ```diff
> diff --git a/meson.build b/meson.build
> index 267b91b3..28453da3 100644
> --- a/meson.build
> +++ b/meson.build
> @@ -1,7 +1,7 @@
>  project(
>    'labwc',
>    'c',
> -  version: '0.7.3',
> +  version: '0.8.0-dev',
>    license: 'GPL-2.0-only',
>    meson_version: '>=0.59.0',
>    default_options: [
> @@ -43,7 +43,7 @@ git = find_program('git', native: true, required: false)
>  if git.found()
>    git_commit = run_command([git, 'describe', '--dirty'], check: false)
>    if git_commit.returncode() == 0
> -    version = '"@0@"'.format(git_commit.stdout().strip())
> +    version = '"@0@ (@1@)"'.format(meson.project_version(), git_commit.stdout().strip())
>    endif
>  endif
>  add_project_arguments('-DLABWC_VERSION=@0@'.format(version), language: 'c')
> ```
> 
> which results in `labwc 0.8.0-dev (0.7.3-56-g7489bff9-dirty)` (on some feature branch of mine)

I actually do like this solution because it pretty clearly shows what build you're using.

It doesn't exactly solve the issue as you said but it does solve the overall issue in showing clarity when running `labwc --version`. Should I add your suggestions to the PR or just amend this locally when testing stuff and close the PR?

Consolatis:
> It doesn't exactly solve the issue as you said but it does solve the overall issue in showing clarity when running `labwc --version`. Should I add your suggestions to the PR or just amend this locally when testing stuff and close the PR?

Feel free to add the suggestion to this PR. No objections from my side to apply those changes but maybe the other devs see it differently.

CC @ahesford @tokyo4j @johanmalm @jlindgren90

johanmalm:
I'm not keen on this because it adds complexity to the release process (which I find takes focus to get right as it is already) and also adds commit noise... for no discernable (to me) benefit.

tfemby:
> I'm not keen on this because it adds complexity to the release process (which I find takes focus to get right as it is already) and also adds commit noise... for no discernable (to me) benefit.

I think that ease of maintenance is a valid enough reason to close this. Thank you everybody for explaining this to me and I'll just apply this locally when testing :)

PR #2035, opened 2024-07-29
Menu changes/additions

droc12345:
This draft contains 4 commits, ETA - 6 commits
1. #1976 draft by johanmalm (only here so that the latter changes can be looked at)
2. Pipemenu - make non-parent work
3. Menu placement, x,y placement (global) except for client-menu ie window menu
4. client-list-combined-menu - internal menu based on openbox
ETA
5. client-send-to-menu - internal menu, shows all workspaces and left/right
6. separator label changes, text color, font for label and justification (center)

I'm having second thoughts about making x,y placement to be global, even though that's what openbox does
It would be more flexible to allow per menu placement. I'm open on going either way

This is a WIP, it works for me, and I drive it daily so, I'm keeping an eye out for unwanted behavior.

Anyway, comments, suggestions, etc.

Reference: #1945 and  #1976

Edit to add: I hadn't realized about using sprintf and strcat/strcpy, I'll change those before this goes to PR.

ETA2: Fixed logic flaw/oversight and added another commit, client-send-to-menu, replace the workspace
section of client-list, allows sending current window to any workspace, along with left/right,

ETA3: 07/31 
added send-to menu (shows all workspaces, along with left/right
added separator label enhancements, text color, font, justification (center)

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -498,6 +498,21 @@ extending outward from the snapped edge.
 	outlined rectangle is shown to indicate the geometry of resized window.
 	Default is yes.
 
+*<resize><popupPosition>* see rc.xml.all for example
+	Lets the menu placement be set by a fixed position or centered on
+	screen. If set then the atCursor placement is overridden.  Not enabled
+	by default. Fixed takes x,y coordinates for placement.
+
+	Note: openbox allows 'Center', 'Top', or 'Fixed' we use Center & Fixed.
+
+	Positioning is per monitor based on which monitor the cursor is in.
+
+*<resize><popupFixedPosition>*
+	set x, y coordinates for menu placement
+	- x,y accept positive offsets only.
+	- 0,0 places menu at top left corner of desktop
+	- monitor width -1, monitor height -1 places menu at bottom right.
+
 ## KEYBOARD
 
 *<keyboard><numlock>* [on|off]
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -448,6 +448,7 @@ extending outward from the snapped edge.
 	- ActiveWindow - titlebar of active window
 	- InactiveWindow - titlebar of all windows that aren't focused by the
 	  cursor
+	- MenuHeader - menu header font ie separator labels
 	- MenuItem - menu item (currently only root menu)
 	- OnScreenDisplay - items in the on screen display
 	If no place attribute is provided, the setting will be applied to all
@@ -460,7 +461,7 @@ extending outward from the snapped edge.
 	Font size in pixels. Default is 10.
 
 *<theme><font place=""><slant>*
-	Font slant (normal or italic). Default is normal.
+	Font slant (normal, oblique or italic). Default is normal.
 
 *<theme><font place=""><weight>*
 	Font weight (normal or bold). Default is normal.
--- a/docs/labwc-menu.5.scd
+++ b/docs/labwc-menu.5.scd
@@ -25,13 +25,16 @@ The menu file must be entirely enclosed within <openbox_menu> and
   <!-- A submenu defined elsewhere -->
   <menu id="" />
 
-  <!-- Horizontal line >
+  <!-- Horizontal line -->
   <separator />
 
   <!-- An inline submenu -->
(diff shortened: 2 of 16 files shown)


droc12345:
getting an odd sigsegv with void, it errors out in  _create_buffer in scaled-font-buffer.c.
It doesn't happen on arch or freebsd, so it's probably peculiar to using musl.

What I think is going on, is inside _create_buffer
a struct buffer is created but not initialized, 
then that buffer is passed to font_buffer_create, where if it return early nothing is done to the buffer 
it's still uninitialized.
then a check is done using an element of the buffer and that's where the sigsegv happens.

I'm thinking that setting the struct to be empty might fix the problem, ie struct... = {0};
There is no downside to adding this.

Again normal glibc and whatever bsd uses don't show a problem.

fungilife:
I esperienced an issue with menu and separator, when trying to move one up before a pipe menu.  When it followed a pipe menu it was OK, couldn't figure out any other change that affected menu blowing off to <configure/exit>
This is on an arch-like system (glibc, no systemd no logind no dbus) .. seatd!
x2 x3 pipe menus work flawlessly (dev branch r52

With void if you speculate it is musl do an identical /minimal glibc installation and see if it acts the same of not.  Very few distros you can do this and reliably say it is musl that is affecting it.

droc12345:
I've made a few changes, fixed a logic error, added a few more checks.

When you say separator before a pipemenu, are you talking about something like this
menu
..separator (not sure if you are using title or regular sep line)
..pipemenu

droc12345:
I think this is in shape to be reviewed, if johan merges his  patches, then I'll rebase these to remove that commit.
And then draft could be changed to PR.

johanmalm:
Amazing. Thanks for working on this.

Will try to finish #1976 over the next fortnight. Suggest we aim to merge shortly after the 0.8.0 release.

droc12345:
Your code is in good shape, the only thing I had to do to add my changes was put a space between ){ 
in your code (line 286 in your draft) so that it passed code check.
~95% of code changes stay in menu.c

As far as when it gets put in, that's up to you.

droc12345:
Current version

johanmalm:
Looks fantastic. 
Will sort out my commit after Friday... then we can get your patches reviewed + merged.

droc12345:
I do have to add, that the menus suffer the same problem as the osd window switcher,
ie they don't scroll with more data than can fit on the screen. 
In my 4k monitor with size 18 font, it took ~45 lines to misbehave (5 workspace lines, 5 go to lines and ~35 applications) shouldn't be a problem 99% of the time, but it still exists.

A potential cure is to create a frame that can be written into, then the data can be scrolled within the frame.

Also after the merges we might consider factoring the menu.c file into 2 or 3 files.
I know there is a TODO comment at the top about this, but I thought I'd mention it.

johanmalm:
Agree. We need to fix that.

droc12345:
A couple more changes while waiting. I added center along with fixed placement.
So now you can put center and the menu will try to center itself on both x and y axis.

Fixed a hidden gotcha, I just found. With a 2nd monitor active, the menu kept going to the 1st monitor
with my enhancements, at cursor worked because it's based on seeing a view.
So, with my latest fixes, fixed/center placement will apply to the monitor where the cursor resides.

johanmalm:
@droc12345

These features are great improvements and will be liked by many. In principle I want all these.

If it's okay with you, what I'd suggest to get through them more quickly, is that after the merge of #1976 that you create a separate PR for each of the commits to keep the review simpler. Would that be okay? It avoids delaying all of them in case we stuck on any single one... and it's likely to be be easier to follow the review-history/changes.

I suggest doing the last commit first (the menu font/color/justify configuration). Other than that - whatever order you prefer.

droc12345:
I'll have to look at that, some of the patches set things up for later use.
But it's also why I have 6 commits in the PR, that way each commit can be looked at separately.

I'll check on getting them split out

droc12345:
I've been able to split into 3 pieces, builtin menus, theme (for most part) and menu fixes (pipemenu, etc)
but there no sense in me putting out a PR until you get your draft for menu stuff put into place.
I'm tired of chasing maybe's.

Issue #864, opened 2023-03-31
wlr-randr: display off/on looses window's fullscreen state


m1027:
Hi,

I have got key bindings to switch the display off and on via wlr-randr. When the display is swiched on again, all maximized windows appear unmaximized.

See my config below.

Thanks

`<keybind key="C-A-1"><action name="Execute"><command>wlr-randr --output eDP-1 --off</command></action></keybind>`

`<keybind key="C-A-2"><action name="Execute"><command>wlr-randr --output eDP-1 --on</command></action></keybind>`

Consolatis:
I assume you have a single display?

m1027:
Yes, a single display only.

The version is this commit from master:

commit e44d56567284c5860c1c3148c09628b50ce477db (HEAD -> master, origin/master, origin/HEAD)
Author: Consolatis <35009135+Consolatis@users.noreply.github.com>
Date:   Tue Mar 28 19:23:49 2023 +0200

And maybe I should be more precise: I am not 100% sure but it is probably the fullscreen state that gets lost, not the maximized state.

Thanks for investigating this issue.

Consolatis:
Yes, the fullscreen state is indeed removed when turning the monitor off.
This was a deliberate choice because usually the windows will be transferred to a still running output (if there are multiple ones).
And as the resolution of the new output may be different to the old one we just disable fullscreen in that case to not confuse applications that are in fullscreen and suddenly get their output / dimensions changed.
If there is only a single output available, disabling and re-enabling it is basically the same as transferring applications to another output (that just happens to have the same dimensions and layout position).

The maximized state (same for the tiling states) should not be lost.

m1027:
Thanks again for your effort in this.

While understanding the reasons for your choice I permit myself to see things differently here: :-)

- I am just saving energy by switching the display off when not needed. So that's quite often and not for attaching a new monitor.
- When looking for code how sway handles it, I often see snippets that save energy by swichting the display off together with a screen locker.
- I've been using wayfire for some time and they just keep the fullscreen / maximized state even after display off/on.
- There are probably a lot of users with a single screen scenario only and I believe most users tend to be a little surprised when their windows get resized by display off/on.

I hope you see the use case here and that there will be a way in labwc too to not loose the fullscreen state in case its the same display being off and on (or one with the same dimensions). 

Thanks again for this awesome WM.

johanmalm:
I haven't looked into this, but quick response nonetheless:

[...]

PR #2040, opened 2024-08-02
view: stay fullscreen when view's output is disconnected

jlindgren90:
I intended to fix this quite some time ago but didn't get around to it.

I don't think there's any good reason why we need to un-fullscreen a view when its output is disconnected. We can handle it the same as a maximized view, and move it to a new output (remaining fullscreen) or, if all outputs are disconnected, just leave it as-is.

This is helpful for a media-center use-case, where you have just one view (e.g. Kodi) fullscreen all the time, but the TV might appear to be disconnected if you switch it to a different source.

Tested with a couple different scenarios:

1. Single output disconnected and re-connected: view stayed fullscreen.
2. Secondary output disconnected: view stayed fullscreen but moved to the primary output, and the layer-shell panel on that output was hidden as expected. When the secondary output was re-connected, the view was moved back (still fullscreen) and the panel on the primary appeared again.

Fixes: #864

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -349,7 +349,6 @@ static bool
 view_discover_output(struct view *view, struct wlr_box *geometry)
 {
 	assert(view);
-	assert(!view->fullscreen);
 
 	if (!geometry) {
 		geometry = &view->current;
@@ -362,6 +361,10 @@ view_discover_output(struct view *view, struct wlr_box *geometry)
 
 	if (output && output != view->output) {
 		view->output = output;
+		/* Show fullscreen views above top-layer */
+		if (view->fullscreen) {
+			desktop_update_top_layer_visiblity(view->server);
+		}
 		return true;
 	}
 
@@ -409,12 +412,15 @@ void
 view_set_output(struct view *view, struct output *output)
 {
 	assert(view);
-	assert(!view->fullscreen);
 	if (!output_is_usable(output)) {
 		wlr_log(WLR_ERROR, "invalid output set for view");
 		return;
 	}
 	view->output = output;
+	/* Show fullscreen views above top-layer */
+	if (view->fullscreen) {
+		desktop_update_top_layer_visiblity(view->server);
+	}
 }
 
 void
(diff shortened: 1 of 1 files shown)


Consolatis:
No objections from my side, code LGTM on the first glance but I didn't look deeper into it nor tested.

CC @ahesford

ahesford:
This seems to work as advertised in my dual-head configuration (which throws a fast disconnect-reconnect sequence when a display resumes from sleep) that motivated the geometry-restoration feature in the first place.

However, it poses a question of intent. While there is no question that it would be desirable to restore full-screen (or even maximized) states when restoring a window *to its original display*, is it generally desirable to keep a window fullscreen when it is moved to another display? There might be multiple full-screen windows across multiple displays, each of which may have a unique geometry, and it may be preferable to demote a full-screen window when its display is lost rather than have it compete with whatever has been assigned to a still-active output.

To get a "proper" feel should the answer be that windows should be restored to natural geometry when assigned on a new output, we could simply save a last-fullscreen state along with last geometry, and re-apply the fullscreen mode when the last geometry would be restored on output layout changes.

jlindgren90:
> is it generally desirable to keep a window fullscreen when it is moved to another display?

I suppose it comes down to personal preference, but for me, yes. To me, full screen is a user-requested state that the WM should not automatically change unless necessary, and also it has more to do with the content/purpose of a window than what display it's on.

Use-case: I'm watching a movie on a laptop docked to a TV, and undock to go to a different room. If I'm still watching the movie, I still want it full screen on the laptop. If I'm done watching, I'll pause and/or close the player manually.

ahesford:
That's a reasonable position, and in that case, this PR seems to work as expected.

Consolatis:
> There might be multiple full-screen windows across multiple displays, each of which may have a unique geometry, and it may be preferable to demote a full-screen window when its display is lost rather than have it compete with whatever has been assigned to a still-active output.

Should we unfullscreen if there is already a fullscreen view on the new output?

ahesford:
I don't have a strong opinion here, because the only time I ever full-screen a window is when I'm watching a video in mpv, and there's only ever a single window in that case.

jlindgren90:
> Should we unfullscreen if there is already a fullscreen view on the new output?

My vote would be no - this seems like trying to be too "smart" and guess what the user wants when we don't really know. Also there is nothing wrong with having two full screen views - you can still Alt-Tab between them just fine. I do this sometimes on the "media center" PC.

Consolatis:
Alright, going for the merge then. Thanks!

PR #2041, opened 2024-08-02
Implement locked keybindings

### LINKED ISSUE #2034: [Feature] Allow certain keybinds to be used during a locked state
As the title suggests, it would be useful to allow certain keybinds (volume and media keys to name a few) to be used while the session is locked.

xi:
fixes #2034

There is a [similar feature in sway](https://github.com/swaywm/sway/blob/bc258a3be2f946c1c93bcbe40735b2db068e0ea8/sway/sway.5.scd?plain=1#L409).

### DIFF
--- a/include/config/keybind.h
+++ b/include/config/keybind.h
@@ -18,6 +18,7 @@ struct keybind {
 	xkb_keycode_t keycodes[MAX_KEYCODES];
 	size_t keycodes_len;
 	int keycodes_layout;
+	bool allow_when_locked;
 	struct wl_list actions;  /* struct action.link */
 	struct wl_list link;     /* struct rcxml.keybinds */
 	bool on_release;
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -495,7 +495,7 @@ extending outward from the snapped edge.
 	Stores the keyboard layout either globally or per window and restores
 	it when switching back to the window. Default is global.
 
-*<keyboard><keybind key="" layoutDependent="">*
+*<keyboard><keybind key="" layoutDependent="" onRelease="">*
 	Define a *key* binding in the format *modifier-key*, where supported
 	modifiers are:
 	- S (shift)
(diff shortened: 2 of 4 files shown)


src/config/keybind.c
```diff
@@ -120,6 +120,7 @@ keybind_create(const char *keybind)
 {
 	xkb_keysym_t sym;
 	struct keybind *k = znew(*k);
+	k->locked = false;
```
Consolatis, line 123: This isn't needed, `znew()` will zero out all memory which means it default to `false` anyway.

Consolatis:
Code looks good to me apart from the comment above (the CI static sanitizer is a bit stupid, `k` may be set to `NULL`).
Didn't test.

I don't really have an opinion if we want the feature itself or not although I think it makes sense to e.g. control volume buttons and similar while locked.

CC @ahesford @jlindgren90 @johanmalm @tokyo4j

jlindgren90:
+1 to the idea, this would be useful to me for volume and media keys.

ahesford:
The idea isn't bad, but I don't like the keyword "locked" because it has a bunch of competing connotations that make the intent non-obvious.

Something like "allowWhenLocked" would be clearer.

tokyo4j:
I like this feature, but you should also modify the logic for key release:
```diff
diff --git a/src/input/keyboard.c b/src/input/keyboard.c
index 0caaf406..3529f410 100644
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -458,7 +458,7 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED) {
 		if (cur_keybind && cur_keybind->on_release) {
 			key_state_bound_key_remove(event->keycode);
-			if (seat->server->session_lock_manager->locked) {
+			if (locked && !cur_keybind->allowWhenLocked) {
 				cur_keybind = NULL;
 				return true;
 			}

```

include/config/keybind.h
```diff
@@ -18,6 +18,7 @@ struct keybind {
 	xkb_keycode_t keycodes[MAX_KEYCODES];
 	size_t keycodes_len;
 	int keycodes_layout;
+	bool allowWhenLocked;
```
tokyo4j, line 21: nit: we use pascal case in `rc.xml`, but use snake case for variable names.
```suggestion
	bool allow_when_locked;
```

Consolatis:
Tested and works as expected. Thanks!

PR #2044, opened 2024-08-03
xdg: account for drag resistance in do_late_positioning()

jlindgren90:
The position where the view should be anchored can now be slightly different from the current cursor position.

Addresses a TODO from #2009.


tokyo4j:
Tested with `chromium --start-maximized`. Thank you for your fixup!

PR #2048, opened 2024-08-05
wayland-wrap: add wayland-1.23

jp7677:
Stable Fedora and Debian distributions are still on 1.22.

### DIFF
--- a/subprojects/wayland.wrap
+++ b/subprojects/wayland.wrap
@@ -1,0 +1,7 @@
+[wrap-git]
+url = https://gitlab.freedesktop.org/wayland/wayland.git
+revision = 1.23.0
+
+[provide]
+dependency_names = wayland-1.23.0
+wayland-1.23=wayland


jp7677:
mmh, this fixes building on my Fedora machine, but it's not enough for Debian CI ;), I'll keep this locally for now.

```
Dependency libdrm found: NO. Found 2.4.121 but need: '>=2.4.122'
```

Consolatis:
> mmh, this fixes building on my Fedora machine, but it's not enough for Debian CI ;), I'll keep this locally for now.
> 
> ```
> Dependency libdrm found: NO. Found 2.4.121 but need: '>=2.4.122'
> ```

It should be available in Debian Testing within the next hours: https://tracker.debian.org/pkg/libdrm.
There is also a wlroots 0.18 package but only in experimental so far: https://tracker.debian.org/pkg/wlroots

jp7677:
> Edit: Also not sure why https://tracker.debian.org/pkg/wayland isn't migrating from Unstable to Testing, I've asked in their IRC channel about it.

Looks like everything has landed, but CI is still failing, see recent run from another PR: https://github.com/labwc/labwc/actions/runs/10260492408/job/28386657415?pr=1941

domo141:
Got the same results w/ that .wrap file in Fedora 40, so for future reference:
```
localhost$ ldd build/labwc | sed -n "s,$HOME.*labwc,/path/to/labwc,p"
  libwayland-server.so.0 => /path/to/labwc/build/subprojects/wayland/src/libwayland-server.so.0 ...
  libwayland-client.so.0 => /path/to/labwc/build/subprojects/wayland/src/libwayland-client.so.0 ...
```
and
```
localhost$ grep -aom1 'ORIGIN[^[:cntrl:]]*' build/labwc
ORIGIN/subprojects/wayland/src
```

lazy me copies the libs to relative path from where I run self-built labwc(1) until November (Fedora 41) fixes this problem for me :D

Consolatis:
> > Edit: Also not sure why https://tracker.debian.org/pkg/wayland isn't migrating from Unstable to Testing, I've asked in their IRC channel about it.
> 
> Looks like everything has landed, but CI is still failing, see recent run from another PR: https://github.com/labwc/labwc/actions/runs/10260492408/job/28386657415?pr=1941

> wlroots| Build-time dependency hwdata found: NO (tried pkgconfig and cmake)
> wlroots| Message: Required for the DRM backend.

The CI in the master branch isn't prepared to compile the wlroots sub project. But all the dependencies should now be available in Debian. However I am not sure if we should add the wayland.wrap file and potentially mess with user systems. Although from what I remember, building via .wrap files makes static builds.

droc12345:
You absolutely do not want to put subproject .so files into the running filesystem (potentially causing problems)

What I do with subprojects, if used is to make libs be static and never install them into the running system.

Consolatis:
I've added `libwlroots-0.18-dev` to the Debian runner via #2050
Once that one is available in the repors the CI should start working again.

PR #2052, opened 2024-08-07
 interactive: allow moving horizontally/vertically maximized window

### LINKED ISSUE #1521: Window unmaximizes upon movement
I don't know if this behavior is intended, I hope not.
Issue:
Maximize a window - try to move it - it unmaximizes.

Expected behavior: if window is maximized in only 1 dimension - it's movement is locked to only the opposite dimension. If it is fully maximized: the window is unmovable. Of course in both cases it keeps the maximized status.

It is useful for example for a workflow with tall and narrow terminal windows.

This is also the openbox behavior.

tokyo4j:
Fixes #1521

Applies drag resistance unidirectionally for horizontally/vertically maximized windows, allowing them to be dragged without being untiled immediately.

While dragging a horizontally/vertically maximized window, edge/region snapping is disabled to prevent unintentional snapping and overlays.

The first commit stops unshading maximized/tiled views at the start of interactive move as it did't seem to be useful and caused flicker.

This PR also includes some refactoring to simplify the logic.

I'm totally fine delaying this PR after the next release. But I want to move forward the discussion below:

## Discussion

As I said in IRC yesterday, I'm thinking of changing `<snapping><dragResistance>` to something like `<resistance><snapOffThreshold>` (or maybe `<resistance><unSnapThreshold>`) as `<resistance>` feels more suitable place for this setting. But while testing this PR, I started to think that it may be also useful to have another setting to configure the resistance for horizontally/vertically maximized windows as it requires more carefulness to drag them in one direction without un-tiling them.

What I came up with are:
- `<resistance><snapOffThreshold>` - for normally maximized or tiled windows
- `<resistance><snapOffThresholdHVMaximized>` - for horizontally/vertically maximized windows

Any ideas?

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -330,6 +330,11 @@ this is for compatibility with Openbox.
 	Sets the movement of cursor in pixel required for a tiled or maximized
 	window to be moved with an interactive move. Default is 20.
 
+*<resistance><unMaximizeThreshold>*
+	Sets the one-dimentional movement of cursor in pixel required for a
+	*vertically or horizontally* maximized window to be moved with an
+	interactive move. Default is 150.
+
 ## FOCUS
 
 *<focus><followMouse>* [yes|no]
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -107,7 +107,10 @@
   <resistance>
     <screenEdgeStrength>20</screenEdgeStrength>
     <windowEdgeStrength>20</windowEdgeStrength>
+    <!-- resistance for maximized/tiled windows -->
     <unSnapThreshold>20</unSnapThreshold>
+    <!-- resistance for vertically/horizontally maximized windows -->
+    <unMaximizeThreshold>150</unMaximizeThreshold>
   </resistance>
 
   <resize>
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -125,6 +125,7 @@ struct rcxml {
 	int screen_edge_strength;
 	int window_edge_strength;
 	int unsnap_threshold;
+	int unmaximize_threshold;
 
 	/* window snapping */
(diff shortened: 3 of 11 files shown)


johanmalm:
I really like `<resistance><unSnapThreshold>` but am fine with either. It would be good to get that change in before the release.

Why do we need a different setting for the uni-directional maximized case? Is it because we're likely to need a greater threshold because people are unlikely to be able to drag vertically/horizontally in a "straight" line (i.e. straight within a tolerance plus/minus `unSnapThreshold`?

tokyo4j:
> I really like <resistance><unSnapThreshold> but am fine with either. It would be good to get that change in before the release.

Opened #2056.

> Is it because we're likely to need a greater threshold because people are unlikely to be able to drag vertically/horizontally in a "straight" line (i.e. straight within a tolerance plus/minus unSnapThreshold?

Yes. Not a strong preference though, as I don't usually unidirectionally maximize windows.

Maybe just using `<resistance><unSnapThreshold>` for normally/vertically/horizontally maximized windows is enough. KWin for example has a common threshold of (screen size) / 16 ([ref](https://github.com/KDE/kwin/blob/df9f8f8346b5b7645578e37365dabb1a7b02ca5a/src/workspace.cpp#L2642)).

johanmalm:
Is this one ready for final review?

tokyo4j:
> Is this one ready for final review?

I'll do some refactors and ensure I don't break anything. I'll ping and mark this PR as ready for review when I finish it.

tokyo4j:
@johanmalm Now ready for review!

johanmalm:
Documentation + `rc.xml.all` missing.

Have tested. Works really well :smile:

Not looked through code yet (other than spotting that docs were missing).

Consolatis:
Tested and works great. Thanks!

PR #2055, opened 2024-08-07
output: set 'labwc' as app_id and title for nested outputs

Consolatis:
Running nested now shows the labwc icon + uses labwc in its title rather than "wlroots".

Example:
20240807_23h27m46s_grim

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -323,6 +323,13 @@ new_output_notify(struct wl_listener *listener, void *data)
 		}
 	}
 
+	if (wlr_output_is_wl(wlr_output)) {
+		char title[64];
+		snprintf(title, sizeof(title), "%s - %s", "labwc", wlr_output->name);
+		wlr_wl_output_set_title(wlr_output, title);
+		wlr_wl_output_set_app_id(wlr_output, "labwc");
+	}
+
 	/*
 	 * We offer any display as available for lease, some apps like
 	 * gamescope want to take ownership of a display when they can


tokyo4j:
Looks nice.
Sadly waybar doesn't show labwc's icon, but I guess that's due to lack of `/usr/share/applications/labwc.desktop`.

Consolatis:
> I needed to run `sudo cp /usr/share/wayland-sessions/labwc.desktop /usr/share/applications/` to make waybar show labwc's icon though.

That is up to how panels look up the icons based on the app_id. There should be a `labwc.svg` in `icons/hicolor/scalable/apps` usually and some panels have a fallback to load a icon with the name of the app_id if there is no `.desktop` file available or it doesn't have an icon defined. Its possible that waybar doesn't implement a fallback like that.

johanmalm:
Consider using `strdup_printf()` to avoid truncation and possibly future error messages from static analyzers on not handling the `snprintf()` return value. Or handle the return value I suppose.

https://github.com/labwc/labwc/blob/670cc0f511cef7e82ba4024d4c50f980de479b4c/include/common/string-helpers.h#L44C7-L44C20

Other than that, LGTM.

Narrat:
>> Looks nice. I needed to run `sudo cp /usr/share/wayland-sessions/labwc.desktop /usr/share/applications/` to make waybar show labwc's icon though.
>
> (...)
> Personally I just have a ~/.icons/labwc.svg and that is enough in my case.

Additional note: If the .desktop file is preferred over the icon, than there is also `$HOME/.local/share/applications/`.

PR #2057, opened 2024-08-08
input: move <scrollFactor> to <libinput> section

### LINKED ISSUE #2033: Move `scrollFactor` to `libinput` section
Hello. I suggest moving the `scrollFactor` option from the `mouse` section to the `libinput.device` section, so that you can set a different scroll speed for each peripheral device. For example, I need to set a lower scroll factor for my laptop's touchpad as it seems too fast, but then my mouse scroll speed becomes very slow.

tokyo4j:
Closes #2033.

But my concern about this PR is whether we should allow users to configure per-device scroll factor also for tablet devices. Looking at `handle_axis()` in `src/input/tablet.c`, it seems that we are not applying any multiplier to `distance` or `wheel_delta`. @jp7677 Do you have any opinions on this?

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -590,9 +590,6 @@ extending outward from the snapped edge.
 *<mouse><doubleClickTime>*
 	Set double click time in milliseconds. Default is 500.
 
-*<mouse><scrollFactor>*
-	Set scroll factor. Default is 1.0.
-
 *<mouse><context name=""><mousebind button="" direction="" action=""><action>*
 	Multiple *<mousebind>* can exist within one *<context>*; and multiple
 	*<action>* can exist within one *<mousebind>*.
@@ -798,6 +795,7 @@ extending outward from the snapped edge.
     <clickMethod></clickMethod>
     <sendEventsMode></sendEventsMode>
     <calibrationMatrix></calibrationMatrix>
+    <scrollFactor>1.0</scrollFactor>
   </device>
 </libinput>
 ```
@@ -922,6 +920,9 @@ The most common matrices are:
 
 	visit https://wayland.freedesktop.org/libinput/doc/latest/absolute-axes.html#calibration-of-absolute-devices
 	for more information.
+
+*<libinput><scrollFactor>*
+	Set scroll factor. Default is 1.0.
 
 ## WINDOW RULES
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -323,7 +323,6 @@
 
     <!-- time is in ms -->
     <doubleClickTime>500</doubleClickTime>
-    <scrollFactor>1.0</scrollFactor>
 
     <context name="Frame">
       <mousebind button="A-Left" action="Press">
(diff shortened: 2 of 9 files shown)


jp7677:
No opinion from my side.

I don't think `scrollFactor` is relevant for `distance`. About `wheel_delta`, I'm not completely sure what exactly this event represents, my tablet does not emit such event. I would tend to ignore tablets for now until there is a real life case.

docs/labwc-config.5.scd
```diff
@@ -798,6 +795,7 @@ extending outward from the snapped edge.
     <clickMethod></clickMethod>
     <sendEventsMode></sendEventsMode>
     <calibrationMatrix></calibrationMatrix>
+		<scrollFactor>1.0</scrollFactor>
```
johanmalm, line 798: Looks like you've used a tab here rather than the two-spaces-per-level indentation.

johanmalm (commented):
LGTM me - just one tiny pick up in the documentation.

tokyo4j:
> No opinion from my side.

> I don't think scrollFactor is relevant for distance. About wheel_delta, I'm not completely sure what exactly this event represents, my tablet does not emit such event. I would tend to ignore tablets for now until there is a real life case.

Alright, then let's proceed with the current approach.

tokyo4j:
I'll delay this PR to 0.8.1 as I'm not 100% sure I'm doing right (e.g. handling virtual pointer).

tokyo4j:
Now this is ready to be merged.

But I want to make sure: is it okay to fall back to scroll factor of `1.0` for wayland/x11 backends or virtual pointers because they are not libinput devices?

I don't come up with any real world cases where configuring scroll factor for those devices is useful. But as we highly values stability, if this breaks any existing workflow, we can retain the support for `<mousebind><scrollFactor>` and let `<libinput><scrollFactor>` overwrite it.

docs/rc.xml.all
```diff
@@ -568,6 +567,7 @@
       <clickMethod></clickMethod>
       <sendEventsMode></sendEventsMode>
       <calibrationMatrix></calibrationMatrix>
+      <scrollFactor>1.0</scrollFactor>
```
Consolatis, line 570: The format could be additionally added to the comment above for completeness.

Consolatis:
> But I want to make sure: is it okay to fall back to scroll factor of `1.0` for wayland/x11 backends or virtual pointers because they are not libinput devices?

I think so, I'd assume any scroll factor modifications for those devices will be set on the host compositor instead thus we already get the adjusted events in.

> I don't come up with any real world cases where configuring scroll factor for those devices is useful. But as we highly values stability, if this breaks any existing workflow, we can retain the support for `<mousebind><scrollFactor>` and let `<libinput><scrollFactor>` overwrite it.

I don't think this is necessary.

I didn't test this yet but one thing that might be an issue here is if we get scroll events from tablets. The `assert()` in `cursor_axis()` only seems to handle pointer and touchpad devices. I don't have the hardware to test this.

docs/rc.xml.all
tokyo4j: Thanks. Fixed.

tokyo4j:
> I didn't test this yet but one thing that might be an issue here is if we get scroll events from tablets. The assert() in cursor_axis() only seems to handle pointer and touchpad devices. I don't have the hardware to test this.

This should be fine because `wlr_cursor.events.tablet_tool_axis` event is emitted instead of `wlr_cursor.events.axis`.

tokyo4j:
As there seems to be no further feedback, I'll merge this PR now.

PR #2060, opened 2024-08-08
input: subscribe to tablet tool events from cursor

### LINKED ISSUE #1990: Stylus input doesn't rotate.
* Version: labwc 0.7.3
* Distro: Arch

I am using a laptop with a touchscreen and stylus and using rot8 to rotate the screen based on orientation.
This works well under Xorg and Sway, but under labwc the rotation is only applying to the finger input, not the stylus input.

These are the lines from my `rc.xml`
```xml
<touch deviceName="Wacom HID 5284 Finger" mapToOutput="eDP-1" />
<touch deviceName="Wacom HID 5284 Pen" mapToOutput="eDP-1" />
```

The same works under Sway when I run these commands:
```bash
swaymsg "input 1386:21124:Wacom_HID_5284_Pen map_to_output eDP-1"
swaymsg "input 1386:21124:Wacom_HID_5284_Finger map_to_output eDP-1"
```

jp7677:
Contrary to the raw tablet events, the cursor events transform the coordinates based on a mapped output orientation. Otherwise those events are the same. See https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_cursor.c#L897

Note that this only applies to absolute coordinates, the relative ones (`dx`, `dy`) are not transformed, in that case a user still has to manually set rotation in our tablet config.

~Additionally our own rotation is applied after area transformation, but the auto rotation is applied before. No real issue, just a slightly inconsistent behavior. We could turn around our behavior though to make this fully consistent, which is a minimal change.~

TODO:
- [x]  More testing
- [x] Clarify auto rotation in the docs (tablet -> mapToOutput)
- [x] Apply rotation before (instead after) area transformation for consistent rotation behavior

Closes #1990 

cc: @dashavoo

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -696,7 +696,7 @@ extending outward from the snapped edge.
 
 *<tablet rotate="" />* [0|90|180|270]
 	The tablet orientation can be changed in 90 degree steps. Default is
-	no rotation (0). Rotation will be applied after applying tablet area
+	no rotation (0). Rotation will be applied before applying tablet area
 	transformation.
 
 	See also *calibrationMatrix* in libinput section below for advanced
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -691,6 +691,9 @@ extending outward from the snapped edge.
 	If the output name is left empty or the output does not exists, the
 	tablet will span all outputs.
 
+	The tablet cursor automatically applies the orientation/rotation of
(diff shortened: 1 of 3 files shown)


dashavoo:
Yes, this does the trick! Sorry I didn't reply last night, I had trouble applying the patch you suggested and then quite literally fell asleep in front of the computer. Thank you so much for fixing this.

johanmalm:
LGTM. Thanks.

PR #2065, opened 2024-08-09
fix: move cursor event listeners from tablet to seat 

jp7677:
Otherwise we would subscribe multiple times to the same event when having multiple tablets.

Follow up from https://github.com/labwc/labwc/pull/2060 after sleeping a night and looking again how the eventing is setup for `touch` (which also subscribes to cursor events).

### DIFF
--- a/include/input/tablet.h
+++ b/include/input/tablet.h
@@ -23,15 +23,13 @@ struct drawing_tablet {
 	double slider;
 	double wheel_delta;
 	struct {
-		struct wl_listener tablet_tool_proximity;
-		struct wl_listener tablet_tool_axis;
-		struct wl_listener tablet_tool_tip;
-		struct wl_listener tablet_tool_button;
 		struct wl_listener destroy;
 	} handlers;
 	struct wl_list link; /* seat.tablets */
 };
 
-void tablet_init(struct seat *seat, struct wlr_input_device *wlr_input_device);
+void tablet_init(struct seat *seat);
+void tablet_finish(struct seat *seat);
+void tablet_create(struct seat *seat, struct wlr_input_device *wlr_input_device);
 
 #endif /* LABWC_TABLET_H */
--- a/include/input/tablet-pad.h
+++ b/include/input/tablet-pad.h
@@ -36,7 +36,7 @@ struct drawing_tablet_pad {
 	struct wl_list link; /* seat.tablet_pads */
 };
 
-void tablet_pad_init(struct seat *seat, struct wlr_input_device *wlr_input_device);
+void tablet_pad_create(struct seat *seat, struct wlr_input_device *wlr_input_device);
 void tablet_pad_attach_tablet(struct seat *seat);
 void tablet_pad_enter_surface(struct seat *seat, struct wlr_surface *wlr_surface);
 
--- a/include/input/tablet-tool.h
+++ b/include/input/tablet-tool.h
@@ -17,7 +17,7 @@ struct drawing_tablet_tool {
 	struct wl_list link; /* seat.tablet_tools */
 };
 
-void tablet_tool_init(struct seat *seat,
+void tablet_tool_create(struct seat *seat,
 	struct wlr_tablet_tool *wlr_tablet_tool);
 bool tablet_tool_has_focused_surface(struct seat *seat);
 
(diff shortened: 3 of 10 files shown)


include/input/tablet.h
```diff
@@ -23,15 +23,13 @@ struct drawing_tablet {
...
-		struct wl_listener tablet_tool_button;
 		struct wl_listener destroy;
 	} handlers;
 	struct wl_list link; /* seat.tablets */
 };
 
-void tablet_init(struct seat *seat, struct wlr_input_device *wlr_input_device);
+void tablet_init(struct seat *seat);
+void tablet_finish(struct seat *seat);
+void tablet_drawing_tablet_init(struct seat *seat, struct wlr_input_device *wlr_input_device);
```
jp7677, line 33: May be there is a better/shorter name...

jp7677:
I would appreciate a thorough look at the wiring now. Is is safe to assume that a tablet event (`tablet_tool_axis` and friends) from `wlr_cursor` will never happen before we have initialized our own tablet structure in `tablet_drawing_tablet_init`?

Consolatis:
I am not sure if listening to the seat pointer signals is the right approach, it shouldn't matter for a single pen setup but if there are multiple ones wouldn't they all see the same mixed up events now and if delivered to an application supporting the tablet protocol they would get the same events multiple times, one per pen?

jp7677:
> Edit: Although it looks like the tablet events detect the tablet / pen being used by their event structs so maybe its actually fine to use the seat signals here. And in that case only having a single listener for the seat tablet events sounds like the correct approach.

Yes, those are my thoughts. Our own created `drawing_tablet` structs are always accessed though the `wlr_tablet_tool_*_event's` (`ev->tablet->data` and `ev->tool->data`). The thing I'm not so sure about is, can I assume that `ev->tablet->data` is always valid. Thus will a tablet tool event from `seat->cursor` always preceded by an initial `seat->new_input` event which leads to the creation of our own `drawing_tablet`.

include/input/tablet.h
jp7677: I want for `tablet_create` and also changed that for pad and tool accordingly.

jp7677:
> That is a good question, in either case we should `assert()` against that if we are sure that to be the case or otherwise early return from the event (e.g. ignore it). I assume there can't be any device events before the devices are added via the libinput / wayland backend but I'd have to look into it.

Thanks, I've added `asserts`. 
I've looked at the sway code. They also subscribe to the tablet events of `wlr_cursor`, so I guess that's good. The initialization of their own structs is done from the proximity event and not on `seat->new_input` like we do, so slightly different approach there.

Consolatis:
I can't currently wrap my head around all variables here so need to tap a bit into your judgement in regards to merging this before or after the release:

- what is the worst that could happen when we don't merge this for 0.8.0, duplicated events when there are multiple pens active?
- how likely is that merging this PR before 0.8.0 creates other issues, I feel like its pretty solid if its verified that the seat cursor can't send events until after we created the tablet and assigned the data pointer.

jp7677:
> * what is the worst that could happen when we don't merge this for 0.8.0, duplicated events when there are multiple pens active?

We would definitively get duplicated events with multiple tablets. I don't think is is an issue for axis or proximity events, but tip/button events could be problematic (multiple down and later multiple up/release). Assuming that we can get events from a tablet before `seat->new_input`, which I actually don't think can happen, we could also on master have the scenario that the first tablet subscribed to the cursor events and a second tablet sends such an event without being initialized first. This would explode because due to the first tablet, we are subscribed to the tablet events from `wlr_cursor`.

> * how likely is that merging this PR before 0.8.0 creates other issues, I feel like its pretty solid if its verified that the seat cursor can't send events until after we created the tablet and assigned the data pointer.

I'm convinced that having this PR merged before the next labwc release is the better option, the wiring looks much more sane. What about replacing the `asserts` from the last commit with early returns (and log statements) when the data pointer is `NULL`? That way I think we are definitively on the safe side.

Consolatis:
> I'm convinced that having this PR merged before the next labwc release is the better option, the wiring looks much more sane. What about replacing the `asserts` from the last commit with early returns (and log statements) when the data pointer is `NULL`? That way I think we are definitively on the safe side.

Sounds good, lets do this.

jp7677:
> Sounds good, lets do this.

Nice, done with the last (force) push.

Consolatis:
Thanks!

PR #2066, opened 2024-08-09
xdg-activation: temporarily disable source surface verification

tokyo4j:
I believe #1974 is the way to go, but it prevents activation of some major applications like thunderbird.

As I suggested #1974, adding option to configure policy for focus requests like `<core><focusRequestRequirement input="yes|no" focus="yes|no">` or implementing urgency hint will be the optimal way to prevent arbitrary focus-stealing without degrading UX. But as we're almost in the cool-down period, let's disable the source surface verification temporarily and move it forward later.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -785,10 +785,17 @@ xdg_activation_handle_request(struct wl_listener *listener, void *data)
 		return;
 	}
 
-	if (!token_data->had_valid_surface) {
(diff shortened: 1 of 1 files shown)


Consolatis:
Fine with both, either this variant or implementing the config for overriding it (I'd count that as a fix).

Consolatis:
As there was no further feedback lets go with the relaxed check for this release.

PR #2069, opened 2024-08-11
xdg: fix error when launching windowed Chromium

tokyo4j:
This PR fixes an error message when launching a windowed Chromium: `00:00:01.778 [../src/view.c:720] view has empty geometry, not centering`.
That occurred because Chromium sends 2 commits before the mapping commit.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -149,7 +149,7 @@ handle_commit(struct wl_listener *listener, void *data)
 	 * the pending x/y is also unset and we still need to position
 	 * the window.
 	 */
-	if (wlr_box_empty(&view->pending)) {
+	if (wlr_box_empty(&view->pending) && !wlr_box_empty(&size)) {
 		view->pending.width = size.width;
 		view->pending.height = size.height;
 		do_late_positioning(view);


Consolatis:
I am a bit out of the loop with the late repositioning.
CC @jlindgren90

jlindgren90:
This is probably fine, I would like to just double-check that this doesn't result in the positioning happening one commit too late.

Is been_mapped already set before the commit that results in the view being visible?

tokyo4j:
> Is been_mapped already set before the commit that results in the view being visible?

I did a quick debug with `wlr_log()`s.
`xdg_toplevel_view_map()` is called before `handle_commit()`, so mapping commits can trigger `do_late_positioning()`.

But it may be too much dependent on the order we register commit/map handlers or wlroots internals. I'm thinking of replacing `view->been_mapped` with `wlr_box_empty(size)`.

jlindgren90:
> I'm thinking of replacing view->been_mapped with wlr_box_empty(size).

Sounds good, I like that approach.

jlindgren90:
I tried with google-chrome-stable here and couldn't reproduce the error, but this LGTM anyway. Thanks!

PR #2072, opened 2024-08-12
Abstract foreign-toplevel interfaces

Consolatis:
And add support for ext-toplevel-list.

This will allow to easily
- add support for toplevel screencapture once implemented in wlroots
- add support for some wlr-foreign-toplevel or ext-foreign-toplevel-list extension for cosmic/ext workspace interop

Note:
There is currently a serious bug in the ext-foreign-toplevel-list implementation in wlroots that causes everything to crash, fix has been sent upstream at https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4782


include/foreign-toplevel.h
```diff
@@ -0,0 +1,13 @@
...
+
+struct view;
+struct foreign_toplevel;
+
+struct foreign_toplevel *foreign_toplevel_create(struct view *view);
+void foreign_toplevel_set_parent(struct foreign_toplevel *toplevel,
+	struct foreign_toplevel *parent);
+void foreign_toplevel_destroy(struct foreign_toplevel *toplevel);
+
+#endif /* LABWC_FOREIGN_TOPLEVEL_INTERNAL_H */
```
Consolatis, line 13: wrong comment

include/foreign-toplevel-internal.h
```diff
@@ -0,0 +1,83 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_FOREIGN_TOPLEVEL_INTERNAL_H
+#define LABWC_FOREIGN_TOPLEVEL_INTERNAL_H
+
+#include <wayland-server-core.h>
+#include "foreign-toplevel.h"
```
Consolatis, line 6: `<stdbool.h>`

Consolatis:
- Rebased on master
- Moved creating the toplevels above the initial positioning so the `new_outputs` event is triggered and updates the toplevel outputs
- Fixed the minor stuff above

I only had a quick test with the new state but I don't expect any serious issues.

There is one more thing I am not sure about, as wlroots was broken for ext-foreign-toplevel-list in 0.18.0 (fixed in 0.18.1) and this PR implements support for that protocol should we increase the wlroots dependency to 0.18.1? Otherwise users might use that protocol and crash their compositor.

johanmalm:
> There is one more thing I am not sure about, as wlroots was broken for ext-foreign-toplevel-list in 0.18.0 (fixed in 0.18.1) and this PR implements support for that protocol should we increase the wlroots dependency to 0.18.1? Otherwise users might use that protocol and crash their compositor.

Yes we ought to bump dependency

johanmalm:
LGTM
Really nice abstraction.
Not tested.

johanmalm:
Do we have an `ext-toplevel-list` test client?

Consolatis:
I've used [wl_framework](https://github.com/Consolatis/wl_framework) but I don't think I've actually pushed the protocol implementation. Could do so later.

https://git.sr.ht/~leon_plickat/lswt also implements it but I never actually tested it.

PR #2073, opened 2024-08-12
xwayland: remove dissociate work-around

johanmalm:
...to cope with pre-wlroots-0.18.0 behaviour resulting in dissociate events being emitted without a prior associate one.

As discussed via IRC with @vyivel and @Consolatis 

Suggest merging after `0.8.0`

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -279,19 +279,8 @@ handle_dissociate(struct wl_listener *listener, void *data)
 	struct xwayland_view *xwayland_view =
 		wl_container_of(listener, xwayland_view, dissociate);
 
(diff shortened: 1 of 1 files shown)


Consolatis:
Do we really need the comment there? Otherwise LGTM.

johanmalm:
Regarding the comment, I just though it’d be easier in future to git-blame to re-trace our steps if there was a problem.

Consolatis:
I think an `assert()` is enough and technically still works for git-blame. Thoughts @jlindgren90 ?

jlindgren90:
I like the assert() idea - it seems clear enough.

Consolatis:
:)

PR #2075, opened 2024-08-13
Fixes for foreign-toplevel when a view is re-mapped

### LINKED ISSUE #2074: `zwlr_foreign_toplevel_handle_v1::output_enter` is not sent when xwayland surface is re-mapped
When I close a window of Slack desktop app, its xwayland surface is unmapped and the corresponding taskbar item disappears (and then slack app runs in background). However, when I open Slack again and the xwayland surface is re-mapped, its taskbar item is still missing.

I confirmed following patch fixes this issue:

```diff
diff --git a/src/xwayland.c b/src/xwayland.c
index 738c2b17..e7b4b148 100644
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -771,6 +771,8 @@ xwayland_view_map(struct view *view)
 	if (xwayland_surface->strut_partial) {
 		output_update_all_usable_areas(view->server, false);
 	}
+
+	foreign_toplevel_update_outputs(view);
 }
 
 static void
```

In my observation, the reason why xdg surfaces don't suffer from this bug is that `handle_commit() -> view_impl_apply_geometry() -> view_moved() -> view_update_outputs()` sends `output_{enter,leave}` events when re-mapped, since `view->{pending,current}` and `view->outputs` are reset to zero on unmap (if I understand correctly).

~I'd like to delegate this to other devs as I don't know how commit/map/unmap work in xwayland.~

tokyo4j:
1. Fixes #2074
2. Just like we're doing for xwayland surfaces, destroy foreign-toplevel handle rather than hiding it when xdg surface is unmapped. This behavior follows sway and xdg-shell protocol:
>  All active operations (e.g., move, resize) are canceled and all attributes (e.g. title, state, stacking, ...) are discarded for an xdg_toplevel surface when it is unmapped.

(2.) is tested with https://github.com/tokyo4j/wayland-demo/tree/remap

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -738,15 +738,6 @@ xwayland_view_map(struct view *view)
 		view->scene_node = &tree->node;
 	}
 
-	/*
-	 * Exclude unfocusable views from wlr-foreign-toplevel. These
-	 * views (notifications, floating toolbars, etc.) should not be
-	 * shown in taskbars/docks/etc.
-	 */
-	if (!view->toplevel.handle && view_is_focusable(view)) {
-		init_foreign_toplevel(view);
-	}
-
 	if (!view->been_mapped) {
 		check_natural_geometry(view);
 		set_initial_position(view, xwayland_surface);
@@ -758,6 +749,16 @@ xwayland_view_map(struct view *view)
 		 */
 		view->current = view->pending;
 		view_moved(view);
+	}
+
+	/*
+	 * Exclude unfocusable views from wlr-foreign-toplevel. These
+	 * views (notifications, floating toolbars, etc.) should not be
+	 * shown in taskbars/docks/etc.
+	 */
+	if (!view->toplevel.handle && view_is_focusable(view)) {
+		init_foreign_toplevel(view);
+		foreign_toplevel_update_outputs(view);
 	}
 
 	/* Add commit here, as xwayland map/unmap can change the wlr_surface */
@@ -797,7 +798,6 @@ xwayland_view_unmap(struct view *view, bool client_request)
 out:
 	if (client_request && view->toplevel.handle) {
 		wlr_foreign_toplevel_handle_v1_destroy(view->toplevel.handle);
-		view->toplevel.handle = NULL;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/xwayland.c
```diff
@@ -797,7 +798,6 @@ xwayland_view_unmap(struct view *view, bool client_request)
 out:
 	if (client_request && view->toplevel.handle) {
 		wlr_foreign_toplevel_handle_v1_destroy(view->toplevel.handle);
-		view->toplevel.handle = NULL;
```
tokyo4j, line 800: Removed as we're already doing it in `handle_destroy()` in `foreign.c`

Consolatis:
Setting view->outputs to `0` in the xwayland unmap handler (on `client_request`) doesn't work?
If it does I'd prefer that one over this PR for the release as it offers less options to cause side effects.

tokyo4j:
> Doing `view->outputs = 0` in the xwayland unmap handler (on `client_request`) doesn't work? If it does I'd prefer that one over this PR for the release as it offers less options to cause side effects.

It alone doesn't work because `view_moved()` is not called on remap as the window geometry is not updated. `view_moved()` isn't called even if I reset `view->{pending,current}` to 0x0 on unmap because `ensure_initial_geometry_and_output()` in `xwayland_view_map()` syncs those geometries before calling `view_moved()`.
I don't know how to fix the issue with this approach in maintainable and least impactful way. CC @jlindgren90

I don't insist on merging this PR for 0.8.0 as no one else has reported the issue and I use git AUR, but I think that this is a "better to have" fix. If we decide to merge this for 0.8.0, I'll remove the second commit.

Consolatis:
In general I think this PR is perfectly fine, including destroying the foreign toplevel handle on xdg umap. Just not sure if we should take the risk for the 0.8.0 release as this also moves creating the foreign toplevel in the xwayland case to a later point (after the positioning logic) which might subtly change the events sent.

So my vote would be to use this PR but delay it until after the release.
CC @johanmalm

johanmalm:
Thanks for the PR. Looks good, but let’s merge after release to play it safe.

Issue #2078, opened 2024-08-14
lab_wlr_scene_output_commit should update after wltoots 0.18 update


yuiiio:
since 
https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/06d301bf1a15826fb6ab89b9c40d40ecb86060cc

https://github.com/labwc/labwc/blob/master/src/common/scene-helpers.c#L45
```diff
        if (!wlr_output->needs_frame && !pixman_region32_not_empty(
-                       &scene_output->damage_ring.current) && !wants_magnification) {
+                       &scene_output->pending_commit_damage) && !wants_magnification) {
                return true;
        }
```
this and wlroots !4753 fixes fullscreen(direct scan-out) VRR.(broken since wlroots 0.18 update)

Consolatis:
This will sadly break the magnifier but after some talks with @Nefsen402 we found a suitable solution that should work with wlroots 0.18.0 and 0.18.1 (when the wlroots fixes are backported).

I'll open a PR later, thanks for bringing this up!

yuiiio:
thank you! Consolatis

PR #2079, opened 2024-08-14
scene-helpers: use pending_commit_damage, chases wlr!4253

Consolatis:
Using the output damage_ring for early out will break VRR in direct scanout mode. The reason being that the damage_ring will be completely ignored in that mode so we need to check `output->pending_commit_damage` instead. This matches with what wlroots has been doing since [0] and it was missed in the initial port to wlroots 0.18.x.

However, that would then break the magnifier which only adds its damage to the damage ring. After some discussion with the wlroots devs we came up with a solution that should work for both, wlroots 0.18.0 and when [1] is backported to 0.18.1.

Note that even with this PR, VRR in direct scanout mode is broken in 0.18.0 from the wlroots side and will be fixed once [1] is backported to the 0.18 branch and 0.18.1 is released.

Fixes: #2078

[0] https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4253
[1] https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4753

---

I'd strongly vote for having this be part of 0.8.0 because otherwise VRR in direct scanout mode will keep being broken even if wlroots releases 0.18.1.

Note, only tested this nested:
- on the current 0.18 branch
- on a local 0.18 branch with cherry-picked backport of [1]

It would be great if somebody could test this on the DRM side but I don't expect any issues there.

### DIFF
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -4,6 +4,8 @@
 #include <wlr/types/wlr_output.h>
 #include <wlr/types/wlr_scene.h>
 #include <wlr/util/log.h>
+#include <wlr/util/region.h>
+#include <wlr/util/transform.h>
 #include "common/scene-helpers.h"
 #include "labwc.h"
 #include "magnifier.h"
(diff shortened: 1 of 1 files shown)


Nefsen402:
> I'd strongly vote for having this be part of 0.8.0

Note that the magnifier will be broken when the backport happens. Needs to be in NEWS and part of this release.

Consolatis:
> > I'd strongly vote for having this be part of 0.8.0
> 
> Note that the magnifier will be broken when the backport happens. Needs to be in NEWS and part of this release.

Why is that? I tested it with https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4753 backported to 0.18.

Nefsen402:
Because we're not updating `pending_commit_damage` from the current labwc release, as soon as the backport comes through, we would be submitting bad damage to the backend as it won't correctly add the damage to `pending_commit_damage`.

Consolatis:
> Because we're not updating `pending_commit_damage` from the current labwc release, as soon as the backport comes through, we would be submitting bad damage to the backend.

Right, but the current labwc release is based on wlroots 0.17. And with this PR we do add the magnifier damage to `pending_commit_damage` so I am slightly confused.

Nefsen402:
Ah, I thought a labwc release already exists of 0.18. Nevermind

johanmalm:
LGTM. I can test tomorrow night.

Consolatis:
Had a chance for a quick test of this PR with the 0.18 branch of wlroots and https://gitlab.freedesktop.org/bl4ckb0ne/wlroots/-/tree/backport-0.18.1 with the mentioned backports included on DRM and everything seems fine. I don't have anything supporting VRR so I am not sure if that is fixed on the backport-0.18.1 branch with this PR or not.

johanmalm:
I have tested on DRM against `0.18.0` as well as `0.18.0` with the `4753` commits cheery-picked.
Have tested Zoom and fullscreen applications. All works fine. Can't test fullscreen.

johanmalm:
LGTM. Thanks for this.

PR #2080, opened 2024-08-15
Add `labwc` to DesktopNames

Vladimir-csp:
A bit of personality to go in `XDG_CURRENT_DESKTOP`

### DIFF
--- a/data/labwc.desktop
+++ b/data/labwc.desktop
@@ -4,4 +4,4 @@ Comment=A wayland stacking compositor
 Exec=labwc
 Icon=labwc
 Type=Application
-DesktopNames=wlroots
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. Will take a look after the release tomorrow.

johanmalm:
I find it hard to navigate what is right here.
Comparing with https://github.com/labwc/labwc/pull/1716 the same change is missing in some other places.
Help with understanding welcome.

Vladimir-csp:
Sorry, I haven't searched existing PRs well enough. #1716 is more comprehensive, you should definitely choose it.
Entry change would only take effect if labwc is started via this entry by a DM/SM.

Issue #857, opened 2023-03-29
Implement `titleLayout` feature


Flrian:
[B+ - 2.4.2](https://github.com/labwc/labwc-scope#24-configuration---theme)
See [here](http://openbox.org/wiki/Help:Configuration#Theme) for more info.

I'd like to be able to remove to remove the client-menu button, because I can always just right click the title bar anyways so it seems unnecessary to me. Also having the option to move the window buttons to the left might be nice for some people.

CosmicFusion:
i want them on the left

johanmalm:
Until titleLayout is implemented you can hide the menu buttons by adding the following to `~/.config/labwc/themerc-override`:

```
window.active.button.menu.unpressed.image.color: #000000 0
window.inactive.button.menu.unpressed.image.color: #000000 0
```

Note the `0` at the end for zero alpha (i.e. transparent)

PR #2088, opened 2024-08-17
theme: implement titleLayout

xi:
An attempt at implementing #857 

I don't really understand `add_scene_button_corner()`, so I would be happy to get some help with that.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -432,6 +432,18 @@ extending outward from the snapped edge.
 *<theme><name>*
 	The name of the Openbox theme to use. It is not set by default.
 
+*<theme><titleLayout>*
+	Selection and order of buttons in a window's titlebar.
+	The following letters can be used, each only once:
+	- L: window label (aka. title)
+	- |: empty space (can be used instead of a title)
+	- W: window menu
+	- I: iconify
+	- M: maximize
+	- C: close
+
+	Example: WLIMC
+
 *<theme><cornerRadius>*
 	The radius of server side decoration top corners. Default is 8.
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -28,6 +28,7 @@
   <!-- <font><theme> can be defined without an attribute to set all places -->
   <theme>
     <name></name>
+    <titleLayout>WLIMC</titleLayout>
     <cornerRadius>8</cornerRadius>
     <keepBorder>yes</keepBorder>
     <dropShadows>no</dropShadows>
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -14,6 +14,7 @@
 #include "config/tablet-tool.h"
 #include "config/libinput.h"
 #include "resize-indicator.h"
+#include "ssd.h"
 #include "theme.h"
 
 enum view_placement_policy {
@@ -43,6 +44,11 @@ enum tiling_events_mode {
 	LAB_TILING_EVENTS_EDGE = 1 << 1,
 	LAB_TILING_EVENTS_ALWAYS =
 		(LAB_TILING_EVENTS_REGION | LAB_TILING_EVENTS_EDGE),
+};
+
+struct title_button {
+	enum ssd_part_type type;
+	struct wl_list link;
 };
 
 struct usable_area_override {
@@ -75,7 +81,11 @@ struct rcxml {
 
 	/* theme */
 	char *theme_name;
+	struct wl_list title_buttons_left;
+	struct wl_list title_buttons_right;
 	int corner_radius;
+	bool show_title;
+	bool title_layout_loaded;
 	bool ssd_keep_border;
 	bool shadows_enabled;
 	struct font font_activewindow;
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -22,7 +22,6 @@ struct ssd_button {
 	struct wlr_scene_node *hover;
 	struct wlr_scene_node *toggled;
 	struct wlr_scene_node *toggled_hover;
-	struct wlr_scene_node *background;
 	struct wlr_scene_tree *icon_tree;
 	struct wlr_scene_tree *hover_tree;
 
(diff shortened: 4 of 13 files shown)


xi:
It's mostly working now. The major thing that is still missing is rounded corners on the title itself (if there are no buttons on one side. I am also not sure how to do a default value and what to do if the layout does not include the title.

I rebased to 0.7.3 because that is much easier to build for me. I can rebase back to current master once the basic design is settled.

johanmalm:
Thanks for working on this. It's definitely high up on the wish-list.

I've not looked at the SSD-rounded-corner code for a long time so can't intuitively help without spending some time looking into it (which I can do if needed).

I did ponder at one point whether we ought to introduce a colon (to indicate right/left alignment) like gnome does:

`gsettings set org.gnome.desktop.wm.preferences button-layout "menu:minimize,maximize,close"`

It would help if users do not want the `L`. 

AFAIR xfce also use colon, but can't find a link to their config file. Only found this: https://docs.xfce.org/xfce/xfwm4/preferences

We don't normally like to deviate from Openbox spec, but Consolatis and I have disucssed this one before and both concluded that an exception may be warranted here. I'm not sure what syntax we really want, but thought I'd through it in early before we go too far down one route.

For example `L` seems rather weird to describe 'title'.

I did try your patch with no `L` and that results in a weird background.

Other that that it seems to work great except for the rounded corners.

johanmalm:
Sorry - xfwm4 is quite similar to openbox actually, for example `button_layout=O|SHMC` where `|=title`

https://gitlab.xfce.org/xfce/xfwm4/-/blob/master/defaults/defaults?ref_type=heads
https://wiki.xfce.org/howto/xfwm4_theme

xi:
I checked with openbox: If `L` is not included, the remaining buttons are right-aligned. We could add `:` or `|` as meaning "leave space, but do not render a title on top of that space". In that case, only one of them should be allowed.

johanmalm:
> I checked with openbox: If `L` is not included, the remaining buttons are right-aligned. We could add `:` or `|` as meaning "leave space, but do not render a title on top of that space". In that case, only one of them should be allowed.

Was thinking exactly the same.

xi:
I think I'm done. Looking forward to the review!

Note that I removed the background from buttons. This made it much easier to deal with rounded corners, especially if there are no buttons on a side so that the title background needs to be rounded.

xi:
One detail I noticed: If I start with something like `LC` and switch to something like `|C`, the title on existing windows is not removed. Not sure how important that is or how it could be fixed.

johanmalm:
There is a lot to take in here, so have just done a 'light' review to start off :smile: Hope that's okay.

In documentation: Describe that L and | also separate left-hand from right-hand aligned buttons.

Feels like `title_layout_loaded` should not be needed. I can only see it used in `post_processing()`. Suggest just doing a check like if `if (!wl_list_length(&rc.title_buttons_left) && !wl_list_length(&rc.title_buttons_right)) {...}`

When we walk the titleLayout string (where the `c2` variable is used) - suggest a separate function to check for duplicates characters (possibly in rcxml.c post processing).

johanmalm:
Have played around with it, and it seems to work really well. The only oddity I spotted was the client-menu opens in the top left corner even if the button is over on the right. Suggest we fix that in a separate PR though. Should be some.

Regarding the title on reconfigure. The title texture is attached to each window (as they're unique). Not looked into it (yet) but suspect we could do something with `ssd_update_title()`

xi:
> Feels like `title_layout_loaded` should not be needed. I can only see it used in `post_processing()`. Suggest just doing a check like `if (!wl_list_length(&rc.title_buttons_left) && !wl_list_length(&rc.title_buttons_right)) {...}`

I was thinking about that, but then it would no longer be possible to select no buttons at all. IMHO that should definitely be possible.

johanmalm:
> > Feels like `title_layout_loaded` should not be needed. I can only see it used in `post_processing()`. Suggest just doing a check like `if (!wl_list_length(&rc.title_buttons_left) && !wl_list_length(&rc.title_buttons_right)) {...}`
> 
> I was thinking about that, but then it would no longer be possible to select no buttons at all. IMHO that should definitely be possible.

Ah sorry - I'm not thinking straight. Ignore that :smile:

johanmalm:
Just having a discussion with @Consolatis on IRC about the `titleLayout` syntax.

The general consensus seems to be that `menu:title:minimize,maximize,close` is more intuitive for the user (inspired by gnome and gsettings).

...so the colon would be the separator for left, center, right alignment - and obviously the commas are just field separators after that.

Thoughts on that?

I think that in practice it would be unusual for someone to want buttons in the centre, so the openbox/xfce4/your impelementations are likely to technically work for all people, but it's nice to ponder how intuitive things are to configure when we have the chance.

If you're okay with that syntax, we've used `g_strsplit()` in a few places to make splitting easier. We avoid `strtok()`.

I'm sure you've already seen it, but in `themerc` there is a `window.label.text.justify` to set the title-alignment. But I think that should still work.

xi:
> The general consensus seems to be that menu:title:minimize,maximize,close is more intuitive for the user (inspired by gnome and gsettings).

Would it be ok leave that for someone else? I am not super experienced with string handling in C. Also, the interaction with `window.label.text.justify` seems complicated to me. This PR is already quite big. So I would personally prefer to merge this as a first step, maybe remove it from the documentation, and then do the settings parsing separately.

johanmalm:
> > The general consensus seems to be that menu:title:minimize,maximize,close is more intuitive for the user (inspired by gnome and gsettings).
> 
> Would it be ok leave that for someone else? I am not super experienced with string handling in C. Also, the interaction with `window.label.text.justify` seems complicated to me. This PR is already quite big. So I would personally prefer to merge this as a first step, maybe remove it from the documentation, and then do the settings parsing separately.

Yep, fine. If nobody else is up for it, I'll do it.

xi:
To summarize next steps:

- change the configuration to something like `menu:title:minimize,maximize,close`, where colons separate left, center, and right
  - rephrase the documentation
  - refactor the code that prevents duplicated buttons
  - figure out what to do with `window.label.text.justify` when alignment can also be configured using `titleLayout`
- fix removing titles on `labwc --reconfigure` when it has been removed from the setting
- fix opening client-menu next to the button

Which of these should be done before we can merge this PR?

johanmalm:
> To summarize next steps:
> 
>     * change the configuration to something like `menu:title:minimize,maximize,close`, where colons separate left, center, and right
>       
>       * rephrase the documentation
>       * refactor the code that prevents duplicated buttons
>       * figure out what to do with `window.label.text.justify` when alignment can also be configured using `titleLayout`
> 
>     * fix removing titles on `labwc --reconfigure` when it has been removed from the setting
> 
>     * fix opening client-menu next to the button
> 
> 
> Which of these should be done before we can merge this PR?

Would you be able to do the reconfigure? And we'll leave the rest on another PR.

xi:
> Would you be able to do the reconfigure?

Done. This was easier than I had expected. I had simply missed resetting the booleans in `rcxml_init()`.

johanmalm:
Okay.. so ready for review+merge?

johanmalm:
Thanks. Great effort. I'll sort the remaining items before the next release.

Issue #1985, opened 2024-07-15
0.7.3: Tablet: Thunar always makes multi selection on left click


Narrat:
Since the labwc 0.7.3 release I notice a different behaviour with the thunar file manager and a tablet.
If I don't directly left click with the pen on a file or directory it will create a multi selection with a random starting point (most of the time top left in the file manager) like when creating a selection with pressed left mouse button. Trying to select multiple files the regular way doesn't work now.
![labwc_thunar](https://github.com/user-attachments/assets/00b4873a-c03d-41fb-ab09-a0425106d2c0)

After bisecting it, this change was introduced with fb57d61f3547d4826d47459b15292389b6a0838c

But I'm unsure if this is an intended change, something broken within thunar or gtk3 or misconfiguration on my side. Any pointers?

johanmalm:
@jp7677 would you be happy to have a look a this one?

jp7677:
I’m currently not in front of my machine, will look more closely at thunar soon.In the meantime, you can switch back to mouse emulation for tablets, see the tablet section in the documentation.On 15 Jul 2024, at 09:07, Johan Malm ***@***.***> wrote:﻿
@jp7677 would you be happy to have a look a this one?

jp7677:
Looking more closely at this, Thunar draws a selection from the last reported pointer coordinate to the current tablet coordinate on stylus down. Since we "park" the pointer cursor at -1/-1 (https://github.com/labwc/labwc/blob/master/src/input/tablet.c#L233), the selection starts at upper-left. Marking the pointer cursor "left" on tablet events might fix this, but (see the comment in the linked source) this leads to other side effects. 

I tend to say that this should be addressed in Thunar, also considering that the GTK3-Demo "Icon View Basics" doesn't show a selection on stylus down. As stated above, switching to mouse emulation mitigates this.

Narrat:
Thank you for looking into this.
Indeed, after setting mouse emulation active helped, when you first mentioned it. I hadn't realized that I changed behaviour, when I merged new config elements.
If needed I can set a Button for toggling. And maybe sometimes in the future this can be achieved via `WindowRules`?

Is here someone around that is involved with XFCE? I thought I had seen a name, but I'm not sure. Would be easier, before I go on the journey on recovering my old account for the bugtracker there... :D

Consolatis:
[...]

PR #2091, opened 2024-08-18
action: add Enable/DisableTabletMouseEmulation

jp7677:
Useful for window rules.

This is a follow up from https://github.com/labwc/labwc/issues/1985#issuecomment-2271575221
Well, actually it is a work around for this issue, esp.  https://github.com/labwc/labwc/issues/1985#issuecomment-2271586923

@Narrat I've tested this briefly with the GTK4-Demo and Nautilus with:

```xml
  <windowRules>
    <windowRule title="*">
      <action name="EnableTabletMouseEmulation"/>
    </windowRule>
    <windowRule title="Paint">
      <action name="DisableTabletMouseEmulation"/>
    </windowRule>
  </windowRules>
```

Would be cool if you could confirm that this works for you.

PS: Not sure about the naming though like I wrote here https://github.com/labwc/labwc/discussions/2070#discussioncomment-10361475

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -294,8 +294,10 @@ Actions are used in menus and keyboard/mouse bindings.
 	decorations (including those for which the server-side titlebar has been
 	hidden) are not eligible for shading.
 
+*<action name="EnableTabletMouseEmulation" />*++
+*<action name="DisableTabletMouseEmulation" />*++
 *<action name="ToggleTabletMouseEmulation">*
-	Toggle mouse emulation for drawing tablets on or off.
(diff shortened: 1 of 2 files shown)


johanmalm:
I'm good with enable/disable. We should use the same language in the documentation though (you've said set/unset in text).
Other than that it LGTM.
Not tested.

johanmalm:
It's not the perfect fit for window rules in that the rules are only applied on first-map.
But as actions I'm fine with them (i.e. the PR is fine).

Narrat:
Did a quick test with various programs (like thunar, gimp) and it did work for me :)

PR #2093, opened 2024-08-19
buf: add buf_add_fmt()

johanmalm:
Cc @tokyo4j Is this useful for the YAML patch instead of the `BUF_FMT` macro?

Haven't tested it yet. Waiting for a test directory (#1099)

### DIFF
--- a/include/common/buf.h
+++ b/include/common/buf.h
@@ -44,6 +44,13 @@ void buf_expand_tilde(struct buf *s);
 void buf_expand_shell_variables(struct buf *s);
 
 /**
+ * buf_add_fmt - add format string to C string buffer
+ * @s: buffer
+ * @fmt: format string to be added
+ */
+void buf_add_fmt(struct buf *s, const char *fmt, ...);
+
+/**
  * buf_add - add data to C string buffer
  * @s: buffer
  * @data: data to be added
--- a/src/common/buf.c
+++ b/src/common/buf.c
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include <assert.h>
 #include <ctype.h>
+#include <stdarg.h>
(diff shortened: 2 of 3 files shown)


src/common/buf.c
```diff
@@ -95,6 +97,39 @@ buf_expand(struct buf *s, int new_alloc)
...
+{
+	if (string_null_or_empty(fmt)) {
+		return;
+	}
+	size_t size = 0;
+	char *p = NULL;
+	va_list ap;
+
+	va_start(ap, fmt);
+	int n = vsnprintf(p, size, fmt, ap);
```
Consolatis, line 111: Could use `NULL` directly rather than `p`.

johanmalm: Good point. Done.

include/common/buf.h
```diff
@@ -43,6 +43,13 @@ void buf_expand_tilde(struct buf *s);
  */
 void buf_expand_shell_variables(struct buf *s);
 
+/**
+ * buf_addf - add format string to C string buffer
+ * @s: buffer
+ * @fmt: format string to be added
+ */
+void buf_addf(struct buf *s, const char *fmt, ...);
```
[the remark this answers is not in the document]
johanmalm: I did initially think of calling it that.
Done.

tokyo4j:
> Cc @tokyo4j Is this useful for the YAML patch instead of the BUF_FMT macro?

Yes, I'll use it for my YAML patch when it's merged.

johanmalm:
@tokyo4j Fancy some reviewing?

tokyo4j:
> @tokyo4j Fancy some reviewing?

LGTM from me. I replaced `BUF_FMT` with `buf_add_fmt` in my local branch for yaml support and it worked without any issues. Tests also worked. Thank you!

PR #2095, opened 2024-08-20
[0.7 backport] src/menu: prevent delayed pipe menu response on item destroy

Consolatis:
Without this patch we would happily access `pipe_ctx->item` members even if the item had been destroyed in the meantime.


johanmalm:
With thanks :smile:

Issue #2099, opened 2024-08-21
Change "FocusOuput" to be more in-line with "MoveToOutput"


orfeasxyz:
These two actions are quite similar, yet the current version of "FocusOutput" does not support directional focus.

ahesford:
Closed in #2100

PR #2100, opened 2024-08-21
Change "FocusOutput" to be more in-line with "MoveToOutput"

orfeasxyz:
This is a small patch to address https://github.com/labwc/labwc/issues/2099.

### DIFF
--- a/include/common/direction.h
+++ b/include/common/direction.h
@@ -1,0 +1,10 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_DIRECTION_H
+#define LABWC_DIRECTION_H
+
+#include "view.h"
+
+enum wlr_direction direction_from_view_edge(enum view_edge edge);
+enum wlr_direction direction_get_opposite(enum wlr_direction direction);
+
+#endif /* LABWC_DIRECTION_H */
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -460,6 +460,7 @@ void desktop_focus_view_or_surface(struct seat *seat, struct view *view,
 
 void desktop_arrange_all_views(struct server *server);
 void desktop_focus_output(struct output *output);
+void warp_cursor(struct view *view);
 struct view *desktop_topmost_focusable_view(struct server *server);
 
 /**
@@ -539,6 +540,21 @@ struct output *output_from_wlr_output(struct server *server,
 struct output *output_from_name(struct server *server, const char *name);
 struct output *output_nearest_to(struct server *server, int lx, int ly);
 struct output *output_nearest_to_cursor(struct server *server);
+
+/**
+ * output_get_adjacent() - get next output, in a given direction,
+ * from a given output
+ *
+ * @output: reference output
+ * @edge: direction in which to look for the nearest output
+ * @wrap: if true, wrap around at layout edge
+ *
+ * Note: if output is NULL, the output nearest the cursor will be used as the
+ * reference instead.
+ */
+struct output *output_get_adjacent(struct output *output,
+	enum view_edge edge, bool wrap);
+
 bool output_is_usable(struct output *output);
 void output_update_usable_area(struct output *output);
 void output_update_all_usable_areas(struct server *server, bool layout_changed);
--- a/src/action.c
+++ b/src/action.c
@@ -19,7 +19,6 @@
 #include "menu/menu.h"
 #include "osd.h"
 #include "output-virtual.h"
-#include "placement.h"
 #include "regions.h"
 #include "ssd.h"
 #include "view.h"
@@ -413,11 +412,6 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 		}
 		break;
 	case ACTION_TYPE_FOCUS_OUTPUT:
-		if (!strcmp(argument, "output")) {
-			action_arg_add_str(action, argument, content);
-			goto cleanup;
-		}
-		break;
 	case ACTION_TYPE_MOVE_TO_OUTPUT:
 		if (!strcmp(argument, "output")) {
 			action_arg_add_str(action, argument, content);
@@ -551,9 +545,6 @@ action_is_valid(struct action *action)
 		break;
 	case ACTION_TYPE_SNAP_TO_REGION:
 		arg_name = "region";
-		break;
-	case ACTION_TYPE_FOCUS_OUTPUT:
-		arg_name = "output";
 		break;
 	case ACTION_TYPE_IF:
 	case ACTION_TYPE_FOR_EACH:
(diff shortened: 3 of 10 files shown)


ahesford (changes requested):
Conceptually, I think it would be good to mimic the `MoveToOutput` behavior here, but there is far too much redundancy. The function `output_get_adjacent_output` (which I might call `output_get_adjacent` for brevity) is essentially identical to `view_get_adjacent_output` except for the latter using `view->output` to grab the output. The function `view_get_adjacent_output` should be eliminated in favor of a common function in `src/action.c` that unifies the output selection before calling `desktop_focus_output` or `view_move_to_output`, depending on the action.

Likewise, the `get_wlr_direction` and `opposite_direction` functions from `view.c` should be made public (and probably relocated into `src/common`) for use without duplication. The names of the functions should be made a bit more descriptive when exposing them publicly. (There is similar duplication of `get_wlr_direction` in `src/overlay.c`, but that is slightly different because it aliases `VIEW_EDGE_CENTER` to `WLR_DIRECTION_UP` rather than throwing an error. This is the version you copied here, but you should instead be using the version from `src/view.c`. I'm not sure if we  still need the unique handling in `src/overlay.c`, but we can keep a separate static function if we do.)

Finally, the configuration parsing for both the `MoveToOutput` and `FocusOutput` actions can be deduplicated.

orfeasxyz:
> Conceptually, I think it would be good to mimic the `MoveToOutput` behavior here, but there is far too much redundancy. The function `output_get_adjacent_output` (which I might call `output_get_adjacent` for brevity) is essentially identical to `view_get_adjacent_output` except for the latter using `view->output` to grab the output. The function `view_get_adjacent_output` should be eliminated in favor of a common function in `src/action.c` that unifies the output selection before calling `desktop_focus_output` or `view_move_to_output`, depending on the action.
> 
> Likewise, the `get_wlr_direction` and `opposite_direction` functions from `view.c` should be made public (and probably relocated into `src/common`) for use without duplication. The names of the functions should be made a bit more descriptive when exposing them publicly. (There is similar duplication of `get_wlr_direction` in `src/overlay.c`, but that is slightly different because it aliases `VIEW_EDGE_CENTER` to `WLR_DIRECTION_UP` rather than throwing an error. This is the version you copied here, but you should instead be using the version from `src/view.c`. I'm not sure if we still need the unique handling in `src/overlay.c`, but we can keep a separate static function if we do.)
> 
> Finally, the configuration parsing for both the `MoveToOutput` and `FocusOutput` actions can be deduplicated.

I rewrote the patch following your suggestions.

src/output.c
```diff
@@ -880,6 +881,49 @@ output_nearest_to_cursor(struct server *server)
 		server->seat.cursor->y);
 }
 
+struct output *
+output_get_adjacent(struct output *output, enum view_edge edge, bool wrap)
+{
+	assert(output);
```
ahesford, line 887: If we let output be NULL, we can use `output_get_adjacent` as a full replacement for `view_get_adjacent_output` and just kill that function entirely:
```suggestion
	if (!output) {
		output = output_nearest_to_cursor(server);
	}
	
	if (!output_is_usable(output)) {
		wlr_log(WLR_ERROR,
			"output is not usable, cannot find adjacent output");
		return NULL;
	}
```

src/action.c
```diff
@@ -1060,7 +1052,24 @@ actions_run(struct view *activator, struct server *server,
...
+				if (!target) {
+					/*
+					 * Most likely because we're already on the
+					 * output furthest in the requested direction
+					 * or the output or direction was invalid.
+					 */
+					wlr_log(WLR_DEBUG, "Invalid output");
+					break;
+				}
+				desktop_focus_output(target);
```
ahesford, line 1072: With the change to `output_get_adjacent` suggested below, the entire block for this action (except for the final call to `desktop_focus_output(target)` can be wrapped in a static function in this file, something like
```c
static struct output *
adjacent_output(struct output *output, struct server *server, struct action *action)
{
	const char *output_name = action_get_str(action, "output", NULL);
	struct output *target = NULL;
	if (output_name) {
		target = output_from_name(server, output_name);
	} else {
		enum view_edge edge =
			action_get_int(action, "direction", VIEW_EDGE_INVALID);
		bool wrap = action_get_bool(action, "wrap", false);
		target = output_get_adjacent(output, edge, wrap);
	}

	if (!target) {
		wlr_log(WLR_DEBUG, "Invalid output");
	}

	return target;
}
```

Then, this action can just do
```c
struct output *target = adjacent_output(NULL, server, action);
if (target) {
	desktop_focus_output(target);
}
break;
```
while the `ACTION_TYPE_MOVE_TO_OUTPUT` block can just do
```c
if (!view) {
	break;
}
struct output *target = adjacent_output(NULL, server, action);
if (target) {
	view_move_to_output(view, target);
}
break;
```

ahesford (changes requested):
This is looking a lot better, but I have a few more refinements to recommend.

orfeasxyz:
I did another set of changes as per suggestion.

include/labwc.h
```diff
@@ -537,6 +538,8 @@ struct output *output_from_wlr_output(struct server *server,
 struct output *output_from_name(struct server *server, const char *name);
 struct output *output_nearest_to(struct server *server, int lx, int ly);
 struct output *output_nearest_to_cursor(struct server *server);
+struct output *output_get_adjacent(struct server *server, struct output *output,
+								   enum view_edge edge, bool wrap);
```
ahesford, line 542:
```suggestion
/**
 * output_get_adjacent() - get next output, in a given direction, from a given output
 * @output: reference output
 * @edge: direction in which to look for the nearest output
 * @wrap: if true, wrap around at layout edge
 *
 * Note: if output is NULL, the output nearest the cursor will be used as the reference.
 */
struct output *output_get_adjacent(struct server *server,
	struct output *output, enum view_edge edge, bool wrap);
```

include/common/direction.h
```diff
@@ -0,0 +1,9 @@
+#ifndef LABWC_DIRECTION_H
```
ahesford, line 1:
```suggestion
// SPDX-License-Identifier: GPL-2.0-only
#ifndef LABWC_DIRECTION_H
```

src/common/direction.c
```diff
@@ -0,0 +1,42 @@
+#include <assert.h>
```
ahesford, line 1:
```suggestion
// SPDX-License-Identifier: GPL-2.0-only
#include <assert.h>
```

ahesford (changes requested):
With a few tweaks to satisfy the linter, this looks pretty good to me.

orfeasxyz:
If everything looks ok then we can merge it. Also, thank you a lot for your suggestions.

orfeasxyz:
I made the final adjustments but the CodeStyleCheck keeps failing, is this a problem with the indentation?

ahesford:
It seems I led you astray... the license style in the headers should use `/* ... */` comments, while the style in the source files should use `//` comments.

I made this change, along with a couple of other cleanups. (I think your changes broke `view_get_adjacent_output`, and it's obsolete anyway, so I killed it.) Everything should work as expected now. Please take a look at my changes, test that `FocusOutput` and `MoveToOutput` both behave as expected, and I'll squash and merge if you are happy.

orfeasxyz:
With some quick testing everything seems to be working fine. I think we can go ahead and merge.

ahesford:
Thanks!

PR #2101, opened 2024-08-21
built in menus

droc12345:
client-list-combined-menu
show all workspaces and all applications in them 

client-send-to-menu
select from all workspaces along with left/right, works on current view. 
Works best as part of client-menu

In both menus active workspace is denoted by ">workspace name<"


src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+	if (menu) {
+		struct menuitem *item, *next;
+		wl_list_for_each_safe(item, next, &menu->menuitems, link) {
+			item_destroy(item);
+		}
+	} else {
+		menu = menu_create(server, "client-send-to-menu", "Send to...");
+	}
+
+	menu->label = xstrdup("Send to...");
```
Consolatis, line 921: This will leak memory of the old `menu->label` and I don't think its necessary as its already supplied as 3rd argument to `menu_create()`.

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+
+	if (wl_list_empty(&menu->menuitems)) {
+		struct workspace *workspace;
+
+		wl_list_init(&menu->menuitems);
+		wl_list_for_each(workspace, &server->workspaces, link) {
+			if (!strcmp(workspace->name,
+						server->workspace_current->name)) {
+				current_item = item_create(menu, strdup_printf(">%s<",
+							workspace->name), false);
```
Consolatis, line 932: This `strdup_printf()` also leaks, suggest either using a stack based approach with `snprintf()` or a temporary variable that is then freed within the loop when going with the heap based `strdup_printf()`.

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+			} else {
+				current_item = item_create(menu, workspace->name, false);
```
Consolatis, line 934: Please add an inline comment like `item_create(menu, workspace->name, /*show_arrow*/ false);`, same for the other `item_create()` calls.

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+	if (menu) {
+		struct menuitem *item, *next;
+		wl_list_for_each_safe(item, next, &menu->menuitems, link) {
+			item_destroy(item);
+		}
+	} else {
+		menu = menu_create(server, "client-list-combined-menu", "Clients...");
+	}
+
+	menu->label = xstrdup("Clients...");
```
Consolatis, line 979: Same here with leaking memory and I don't think its needed as the label is supplied by the 3rd argument in `menu_create()`.

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+	menu->size.height = 0;
+	current_menu = NULL;
+
+	if (wl_list_empty(&menu->menuitems)) {
```
Consolatis, line 983: This should always be empty so the condition can be removed (or replaced with an `assert()` if deemed necessary).

johanmalm:
With the "client-list-combined-menu", I'm not keen on the column 1,2,3 characters to describe window states.
Suggest just making it Openbox like.

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+		wl_list_for_each(workspace, &server->workspaces, link) {
+			if (!strcmp(workspace->name,
+						server->workspace_current->name)) {
+				buf_add(&buffer, ">");
+			}
+			buf_add(&buffer, workspace->name);
+			if (!strcmp(workspace->name,
+						server->workspace_current->name)) {
+				buf_add(&buffer, "<");
+			}
```
Consolatis, line 998: Could be replaced with the newly added `buf_add_fmt()` function like
```c
buf_add_fmt(&buffer, workspace == server->workspace_current ? "<%s>" : "%s", workspace->name);
```

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+			current_item = separator_create(menu, buffer.data);
+			buf_reset(&buffer);
```
Consolatis, line 1000: please use `buf_clear()` here as it keeps the heap allocations, same in the following inner loop.
Then after the outer loop `buf_reset()` needs to be called once to free the internal allocation.

src/menu/menu.c
```diff
@@ -893,6 +895,161 @@ menu_hide_submenu(struct server *server, const char *id)
...
+
+			wl_list_for_each(view, &server->views, link) {
+				if (!strcasecmp(view->workspace->name,
+							workspace->name)) {
```
Consolatis, line 1004: No need to compare the workspace name here, `view->workspace == workspace` should be enough.

droc12345:
Ok, should have taken care of comments.

Ready for review.

src/action.c
```diff
@@ -633,6 +633,10 @@ show_menu(struct server *server, struct view *view,
 		return;
 	}
 
+	/* need to be current to be useful */
+	create_client_list_combined_menu(menu->server);
+	create_client_send_to_menu(menu->server);
```
Consolatis, line 638: This can be removed completely as its already handled in `menu_open_root()`.

src/menu/menu.c
```diff
@@ -908,6 +910,117 @@ menu_hide_submenu(struct server *server, const char *id)
...
+
+		wl_list_for_each(view, &server->views, link) {
+			if (view->workspace->name == workspace->name) {
+				buf_add(&buffer, view_get_string_prop(view, "title"));
+
+				current_item = item_create(menu, buffer.data, false);
+				current_item->id = xstrdup(menu->id);
+				current_item->client_list_view = view;
+				fill_item("name.action", "GoToDesktop");
+				fill_item("to.action", workspace->name);
```
Consolatis, line 1008: I don't think we need this action, at least from memory focusing a window on another workspace should switch there automatically.

src/menu/menu.c
```diff
@@ -983,6 +1096,8 @@ void
 menu_init(struct server *server)
 {
 	wl_list_init(&server->menus);
+	create_client_list_combined_menu(server);
+	create_client_send_to_menu(server);
```
Consolatis, line 1100: I don't think we need to initialize the menus on startup as they are cleared and then reinitialized when used anyway.

src/menu/menu.c
```diff
@@ -1153,6 +1268,13 @@ menu_open_root(struct menu *menu, int x, int y)
 		destroy_pipemenus(menu->server);
 	}
 	close_all_submenus(menu);
+	/*
+	 * dynamic internal client
+	 * similar to pipemenu but no script to execute
+	 */
+	create_client_list_combined_menu(menu->server);
+	create_client_send_to_menu(menu->server);
```
Consolatis, line 1276: Can we make this depend on some `menu->rebuild_on_call` function callback? There is not really a point to reload both when a static menu is opened. E.g. something like this:
```c
struct menu_item {
    ..
    void (*rebuild_on_call)(struct menu *menu);
};

..

if (menu->rebuild_on_call) {
    menu->rebuild_on_call(menu);
}
```

src/menu/menu.c
```diff
@@ -908,6 +910,117 @@ menu_hide_submenu(struct server *server, const char *id)
...
+	}
+
+	menu->size.height = 0;
+
+		struct workspace *workspace;
+		struct view *view;
+		struct buf buffer = BUF_INIT;
+
+		wl_list_init(&menu->menuitems);
+		wl_list_for_each(workspace, &server->workspaces, link) {
```
Consolatis, line 994: Something is off with the indentation here.

src/action.c
droc12345: It didn't update properly (latest running apps) without this. There is hardly any time spent in the creating
the menu items, basically spinning through a couple of in memory structures.

src/menu/menu.c
droc12345: I took out a surrounding if and evidently forgot to adjust.

droc12345:
> How do we deal with users configuring `client-send-to-menu` or `client-list-combined-menu` in their `menu.xml` or in some pipe menu?

Not 100% sure what you mean

I have < menu id="client-list-combined-menu" /> in root-menu
and 
< keybind key="A-Space"><action name="ShowMenu" menu="client-list-combined-menu" /></keybind> for direct use

It is a true menu, but it's kind of in-between a regular menu and a pipemenu.

src/menu/menu.c
droc12345: The reason I have that there is the client menus didn't update properly if a submenu of something like the root menu.

I'm not sure what to do.

include/menu/menu.h
```diff
@@ -47,6 +47,7 @@ struct menuitem {
 	struct menu_scene normal;
 	struct menu_scene selected;
 	struct menu_pipe_context *pipe_ctx;
+	struct view *client_list_view;  /* used by internal client-list */
```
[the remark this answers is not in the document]
droc12345: I'm not sure how likely that is to happen in practice, someone would have to delete a view and almost immediately open the menu and make that selection.

Keeping the view current is one of the reasons I call create_client_list_* just before putting the menu up.

src/menu/menu.c
droc12345: Needs to be there, or it won't show up until some other menu is selected first.

src/menu/menu.c
droc12345: Just tested, and works with just focus/raise, at least in my limited testing.

src/action.c
droc12345: Interesting, I was pretty sure I tested with out these and it didn't work, but I just tried it and
we can remove those and the need for public.

src/menu/menu.c
Consolatis: Right, we need to create the menus themselves on startup because otherwise the action will fail. But we don't need to fill them with content.

src/menu/menu.c
Consolatis: We also need to have that check in `menu_process_item_selection()` when opening a sub menu.

include/menu/menu.h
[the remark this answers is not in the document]
Consolatis: It should be enough to do something like this:
```bash
xcalc &
sleep 5; kill %1
```
and then opening the client menu, waiting for xcalc to close and then select it in the menu.

We do the cleanup for `triggered_by_view` in `src/view.c` -> `view_destroy()`. However it seems that we should have some public `menu_on_view_destroy handler(struct view *view)` instead that can then also deal with the new `client-list-combined-menu`.

droc12345:
Sorry to do this, but I've had some personal issues arise and I won't be able to continue developing for now.
So take the code and do what you need with it. 

Blessings

johanmalm:
> Sorry to do this, but I've had some personal issues arise and I won't be able to continue developing for now. So take the code and do what you need with it.
> 
> Blessings

Look after yourself and those close to you. Hope all goes well.

I won't have access to a computer for a few days, but will sort something out with the PRs after that... unless anyone else gets there first.

Appreciate the efforts you've put into this.

droc12345:
I had a little time, so I worked on this and the positioning #2102 PR. 
Comment on it, review it, I won't be able to get back to it till at least next weekend.

johanmalm:
Okay, so this is the last of the menu related PRs. Nearly there! As always, I prefer to just do one thing per PR where possible because it's easier to review/discuss/agree/merge.

I like `client-list-combined-menu`. It seems fairly consistent with Openbox and is a really nice addition that I think many people will like. It's the default Left-Click on desktop menu in Openbox, so one that many people are likely to be familiar with too, and one that makes up for lack for workspace protocols.

Regarding the `client-send-to-menu` I'd prefer to strip that out of this PR and try to add the functionality to the `client-menu` `Workspace` submenu (just like Openbox does). In that submenu we currenly only have 'Send Left/Right' and I think it would feel a lot better to list the workspaces.

So comments against `client-list-combined-menu`

- [ ] 1. It doesn't feel right have `client-list-view` in the menuitem struct. Would feel better to just create it on the spot when the menu is launched. Am I missing something here?
- [ ] 2. In `menu_execute_item()` I don't like the new if-statements around `actions_run()` because it's hard to follow the flow and will get increasingly ugly if we add more menus that we need to `strcmp()` for. I think we should put that type of logic in `action.c`'s `show_menu()`.

Ref:

johanmalm:
@droc12345 Was it your intent to close this?

droc12345:
> @droc12345 Was it your intent to close this?

I was doing stuff yesterday, and didn't realize I closed it.

droc12345:
Nothing to do with this PR per se, but I think the way to handle the menu being too long (ie too many entries)
is to create a frame (size can be set) using viewporter, then do cropping of data to feed the menu frame.
This requires putting data into something like a linked list, so that scrolling (when too much data) works.

This requires some planning to work right, but would be applicable to any output that labwc generates by way of
menu/window, could be made to work with osd for example. Or an online debug window for something like
config errors.  Lots of potential uses for this, which is why it needs good planning from the start.

droc12345:
The flow for client list is like this:
1., when show menu is selected, a new client list is generated, as it needs to be current and we can't depend on just the menu title, because client list can be a submenu of something like root menu.
2. during the client list create we know which title belongs to which view and this is where I save a  link between the menu item and the view
3. when selected, in menu init, we have modified action run to either send current view (as it is now) or
4. if the action is because of a selection in client list then we send the saved menu view not the current view because when the item is selected we do a focus & raise on the view

johanmalm:
Okay, I get the view-struct in the menuitem now. Sorry - penny didn't drop yesterday.
Regarding the `strcmp()` around `actions_run()`, yeah okay, let's leave as is. If we need to extend some other way in the future we can re-visit it then.
I'm just playing a bit with it. Will push a fixup commit shortly with a couple of minor things.

johanmalm:
@droc12345 Are you okay with these 'fixup' tweaks? If so, I'll squash and merge.

droc12345:
Looks good to me.
I wasn't entirely happy with the name create_client_* 
but I didn't think that calling it init_* would be good outside of an initial setting.
update_* works.

johanmalm:
Got there.
Thanks for your efforts on this.
It's a nice addition.

PR #2102, opened 2024-08-22
menu enhancements placement

droc12345:
menu placement
. x,y coordinates ie. 0,0 or 100,100
. centered on output monitor

< popupPosition>Fixed</popupPosition>

< popupFixedPosition>
  < x>10</x>
  < y>10</y>
< /popupFixedPosition>

OR

< popupPosition>Center< /popupPosition>

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -343,6 +343,14 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			action_arg_add_bool(action, argument, parse_bool(content, true));
 			goto cleanup;
 		}
+		if (!strcasecmp(argument, "x.position")) {
+			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
+		if (!strcasecmp(argument, "y.position")) {
+			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
 		break;
 	case ACTION_TYPE_TOGGLE_MAXIMIZE:
 	case ACTION_TYPE_MAXIMIZE:
@@ -611,7 +619,8 @@ action_list_free(struct wl_list *action_list)
 
 static void
 show_menu(struct server *server, struct view *view,
-		const char *menu_name, bool at_cursor)
+		const char *menu_name, bool at_cursor,
+		const char *pos_x, const char *pos_y)
 {
 	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH
 			&& server->input_mode != LAB_INPUT_STATE_MENU) {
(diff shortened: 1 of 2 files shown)


Consolatis:
This setting does not belong to `<resize>`, if we want this feature it should likely be an argument of the `ShowMenu` action.

droc12345:
The reason I used resize is that's because openbox puts it there.

johanmalm:
In Openbox the `ShowMenu` action has a `position` option. That feels like a sensible way of doing it.

Ref:
- https://github.com/danakj/openbox/commit/238ab3f742644af7f3b4d2a7a7860fedd290ad29

It was one of the last things they added to Openbox.

droc12345:
I had a little time, so I rewrote this, to be like openbox's showmenu approach.

Review it, comment on it, and I might be able to work on it next weekend.

src/action.c
```diff
@@ -679,15 +679,15 @@ show_menu(struct server *server, struct view *view,
 
 		if (!strcasecmp(pos_x, "center")) {
 			x = (output->usable_area.width / 2) - (max_width / 2);
-		} else if (strcasestr(pos_x, "%")) {
+		} else if (strstr(pos_x, "%")) {
```
johanmalm, line 682: I think `strchr()` would be even better.

droc12345: All I'm doing is searching for a "%", I'm not doing anything with the position, as atoi will stop at the first non digit character. Either way should work. Easy fix.

johanmalm:
Hi there. Sorry it's taken a while to get back. Been on holiday, etc. I need to do some follow-up work on the `<titleLayout>` PR before the next release, but will prioritize these menu PRs if you're up for it.

I tend to just do 30 minutes a day at this sort of time, so much prefer longer, more consider messages rather than lots of short back-and-forth (in the interest of getting them through quickly anyway). Also, just doing one at the time is easier for me in terms of rembering what we're doing. So, I've just picked this one to start with; hope that's okay. It seemed straight forward enough.

1. Unless there is a really good reason, I think we ought to follow openbox spec, so I prefer:

```xml
<action name="ShowMenu">
  <menu>root-menu</menu>
  <position>
    <x></x>
    <y></y>
  </position>
</action>
```

rather than how you've currently written it - which is:

```xml
<action name="ShowMenu">
  <menu>root-menu</menu>
  <x></x>
  <y></y>
  <atCursor>false</atCursor>
</action>
```

2. Entry in `labwc-actions(5)` man page is missing. The wording in the openbox wiki tends to be very good, and can be a good starting point.

3. I think it would be better to not explicitly have to set `atCursor` to false. Suggest changing the logic so that `if (pos_x && pos_y)`... then set `atCursor=false`. If we wanted to be really clever we could do a `post_processing()` check, but happy to leave that out because I don't think many people use `atCursor` other than for client-menus.

4. IIRC openbox treats negative pos_x/pos_y as distance from the right/bottom endges respectively. This is particularly useful when wanting to launch a menu from a panel. See https://redtide.github.io/openbox-wiki/Help/Actions/#moveresizeto  

5. With commit messages - try to stick with these guidelines: https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#commit-messages   So probably starting with `action: add <position> argument to ShowMenu` or something like that. Ideally also add an example of what the new config looks like in the commit message - for example like this: https://github.com/labwc/labwc/commit/84c222a84f7e04c1282729d6c1376f334878666d

droc12345:
The atCursor I left alone, as I wasn't sure how many people actually used it, so I worked around it.
It would be easy enough to make atCursor change depending on x/y.

I'm not sure how to do the position braceting, but I'll look at some other setting and try and get an idea.

I really didn't want to mess with negative numbers, but I'm already looking at monitor size anyway,
I'll see what needs doing for that.

The documentation I was just leaving till I saw whether this version of placement was good enough.
But I always try and put documentation if possible.

johanmalm:
That's great.

Just strcasecmp against x.position rather than just x

droc12345:
I think I have addressed all of your comments/suggestions. 

using position
labwc-action
atCursor, I didn't have to set it, x/y overrules atCursor anyway.
negative number ie -0,-0
I fixed up commit message

johanmalm:
Thanks.

Did the following patch:

```
index b36c2ea..a157d6e 100644
--- a/src/action.c
+++ b/src/action.c
@@ -672,7 +672,7 @@ show_menu(struct server *server, struct view *view,
 		} else {
 			if (pos_x[0] == '-') {
 				int neg_x = strtol(pos_x, NULL, 10);
-				x = output->usable_area.width - neg_x;
+				x = output->usable_area.width + neg_x;
 			} else {
 				x = atoi(pos_x);
 			}
@@ -685,7 +685,7 @@ show_menu(struct server *server, struct view *view,
 		} else {
 			if (pos_y[0] == '-') {
 				int neg_y = strtol(pos_y, NULL, 10);
-				y = output->usable_area.height - neg_y;
+				y = output->usable_area.height + neg_y;
 			} else {
 				y = atoi(pos_y);
 			}
```

...to make the negative x/y work.

Have tested and LGTM so going for the merge.

droc12345:
Should be fine, I tested multiple options, positive and negative number, percent works, center works.
multiple line action works, single line action works.

PR #2103, opened 2024-08-22
menu enhancements pipemenu

droc12345:
pipemenus work both parented non parented

< openbox_menu>
< menu id="app-menu" label="Applications" execute="labwc-menu-generator -p" /> ...
< menu id="root-menu" label="">
  ...
  < menu id="app-menu" />

Or

< menu id="root-menu" label="">
  < separator label="Root Menu" />
  < menu id="weather-menu" label="Weather" execute="openweather.sh" />


src/menu/menu.c
```diff
@@ -581,27 +581,32 @@ handle_menu_element(xmlNode *n, struct server *server)
...
-			 * TODO: Consider supporting this
 			 */
-			wlr_log(WLR_ERROR,
-				"pipemenu '%s:%s:%s' has no parent <menu>",
-				id, label, execute);
-			goto error;
+			current_menu = menu_create(server, id, label);
+			current_menu->is_pipemenu = true;
+			current_item = item_create(current_menu, label,
+					/* arrow */ true);
```
Consolatis, line 596: I am not sure if we need to adjust `menu_level` here. CC @johanmalm

droc12345: All I know is it works in practice, haven't seen any glitches.
Typically you would want menu_level for submenus, but we're not doing that.

Whatever way y'all want it, as long as it works.

fungilife:
A few commits ago when first tests of <separator begun if it was placed between <menu tags the entire labwc would segfault and exit and only when placed within menu and between <items it would work, and this behavior seemed to have gone away.

Would it be possible to add color and maybe font features to <separator so under some theme coloring they don't blend in with menu items and pipe-menus?

joborun has been offering both stable versions of labwc and labwc-git all the way to r32 b34f2c9d commit 
 https://git.disroot.org/joborun-pkg/jobcomm/src/branch/main/labwc-git/versions-built   

Although we started as openbox/jwm support distro we have become very labwc oriented recently :)

recent screenshot of setup here:
https://sourceforge.net/projects/joborun/

johanmalm:
Shall we go for this one next? I'll review + test tomorrow.

droc12345:
That works, thanks

johanmalm:
I tried with the following `menu.xml`:

```
<openbox_menu>
  <menu id="root-menu" label="obmenu-generator" execute="obmenu-generator"/>
</openbox_menu>
```

I expected a "typical" applications menu with all my system categories and applications.

I observed:
- One menu item called "obmenu-generator"
- When I hovered over said item I got this message `[../src/menu/menu.c:1337] duplicate id 'root-menu'; abort pipemenu`

I have had a look at the code and think that this needs to be implemented quite differently. I haven't studied it for long but think we ought to call a variant of `parse_pipemenu()` from `handle_menu_element()` rather than creating new menus/itmes there.

I suspect it won't be a trivial change because the associated functions are all written on the assumption that there is a parent item, so in order to avoid creating a complete mess we'd most likely have to refactor all that to make it independent of a parent item. For example, look at `handle_pipemenu_readable()` and `create_pipe_menu()`.

I had a quick go at using a dummy-item, but it got very messy and I couldn't get it to work.

I'm intending to start the freeze for the next release in ca 1 month and have already "committed" to a few other things before that.

Please advise how you want to proceed.

droc12345:
Something slightly wrong with my logic. 
I had it working once upon a time, but when I split out the different menu PRs something is missing.

Leave this for now, and let me try and figure out what's going on.

droc12345:
It works for me and to my satisfication. 
If the criteria mentioned above, ie can't use root-menu as pipemenu id is not acceptable, then just ignore this PR. 
As I said earlier openbox never allowed that as far as I know though.

droc12345:
I'll close this for now, maybe revisit later.

johanmalm:
Okay. 
Yeah someone raised a few weeks ago that Openbox supports pipemenus directly as the rootmenu... which is what I automatically assumed this PR was about.
Having looked at the code, there might be simpler ways of achieving this - like just ask the menu-generators to provide static menu output and add support for parsing that directly from an executable. 
Anyway... probably for another day/week/month.

PR #2104, opened 2024-08-22
ssd: add optional shade button

ahesford:
With the excellent work in #2088, it is now relatively straightforward to add support for a shade button. A lot of this was just cargo-culted from the handling of the maximize button, but seems to work as expected.


ahesford:
While I was at it, I added the "omnipresent" button, which should complete the set of buttons from Openbox.

Consolatis:
Thanks!

stefonarch:
Nice!
Typo in `rc.xml.all ` "Ominpresent".

PR #2105, opened 2024-08-23
ssd: fix incorrect cursor shape on titlebar corner without buttons

tokyo4j:
Fixes a problem like the picture below:
20240823_13h22m19s_grim

I believe this doesn't break `ssd_part_contains()`.


Consolatis:
Thanks!

PR #2077, opened 2024-08-13
[wip] Support rc.yaml

johanmalm:
Tried with:

    ---
    theme:
    - name: Adwaita-light
    - cornerRadius: 2
    ...

Needs lots of testing.

PR #2106, opened 2024-08-23
Support `rc.yaml` (and experimentally `menu.yaml`)

tokyo4j:
This adds support for `rc.yaml` with some tests.

<details><summary>rc.yaml</summary>

```yaml
core:
  xwaylandPersistence: yes
placement:
  policy: cascade
theme:
  dropShadows: yes
  fonts:
    - place: ActiveWindow
      weight: bold
    - place: InactiveWindow
      weight: normal
desktops:
  number: 2
windowSwitcher:
  fields:
    - content: type
      width: 15%
    - content: title
      width: 85%
resize:
  drawContents: no
focus:
  followMouse: yes
  followMouseRequiresMovement: yes
  raiseOnFocus: no
snapping:
  range: 20
  overlay:
    delay:
      inner: 100
      outer: 0
regions:
  - { name: top-left,     x: 0%,  y: 0%,  width: 50%, height: 50% }
  - { name: top-right,    x: 50%, y: 0%,  width: 50%, height: 50% }
  - { name: bottom-left,  x: 0%,  y: 50%, width: 50%, height: 50% }
  - { name: bottom-right, x: 50%, y: 50%, width: 50%, height: 50% }
keyboard:
  repeatRate: 25
  repeatDelay: 600
  keybinds:
    - { key: W-bracketRight,        action: { name: ZoomIn } }
    - { key: W-bracketLeft,         action: { name: ZoomOut } }
    - { key: W-d,                   action: { name: Debug } }
    - { key: W-1,                   action: { name: GoToDesktop, to: Workspace 1 } }
    - { key: W-2,                   action: { name: GoToDesktop, to: Workspace 2 } }
    - { key: W-m,                   action: { name: ToggleKeybinds } }
    - { key: W-q,                   action: { name: Close } }
    - { key: A-F4,                  action: { name: Close } }
    - { key: A-Tab,                 action: { name: NextWindow } }
    - { key: W-e,                   action: { name: Exit } }
    - { key: W-v,                   action: { name: Execute, command: sh -c 'cliphist list | fuzzel --dmenu | cliphist decode | wl-copy' } }
    - { key: W-Tab,                 action: { name: ToggleMaximize } }
    - { key: W-s,                   action: { name: Execute, command: foot } }
    - { key: W-a,                   action: { name: Execute, command: fuzzel } }
    - { key: XF86_AudioLowerVolume, action: { name: Execute, command: pactl set-sink-volume @DEFAULT_SINK@ -2dB } }
    - { key: XF86_AudioRaiseVolume, action: { name: Execute, command: pactl set-sink-volume @DEFAULT_SINK@ +2dB } }
    - { key: XF86_AudioMute,        action: { name: Execute, command: pactl set-sink-mute @DEFAULT_SINK@ toggle } }
    - { key: XF86_AudioMicMute,     action: { name: Execute, command: pactl set-source-mute @DEFAULT_SOURCE@ toggle } }
    - { key: Print,                 action: { name: Execute, command: grim } }
mouse:
  doubleClickTime: 200
  default:
  contexts:
    - name: Frame
      mousebinds:
        - { button: W-Left,  action: Press, actions: [ { name: Raise }, { name: Move } ] }
        - { button: W-Right, action: Drag, action: { name: Raise } }
        - { button: A-Left,  action: Press }
        - { button: A-Left,  action: Drag }
        - { button: A-Right, action: Press }
        - { button: A-Right, action: Drag }
touch:
  deviceName: ELAN2514:00 04F3:2AF1
  mapToOutput: eDP-1
tablet:
  mapToOutput: eDP-1
libinput:
  - category: non-touch
    pointerSpeed: -0.4
  - category: touchpad
    naturalScroll: yes
    tapAndDrag: no
windowRules:
  - identifier: ristretto
    serverDecoration: no
  - identifier: code-url-handler
    action: { name: SetDecorations, decorations: border, forceSSD: yes }
```
</details>

The above `rc.yaml` has duplicated keys in `{ button: W-Right, action: Drag, action: { name: Raise } }`. This should be addressed later in another PR.

The first commit fixes assertion failure when `<touch><deviceName>` is used rather than `<touch deviceName="">`. 

I also experimentally added `menu.yaml` support in the last commit, but I'm thinking of removing it because I found YAML doesn't work well for menu layouts with `<separator>` or nested `<menu>`. For example, a possible `menu.yaml` looks like this:

<details><summary>menu.yaml</summary>

```yaml
menu:
  - id: client-menu
    item:
      label: Minimize
      action: { name: Iconify }
    item:
      label: Maximize
      action: { name: ToggleMaximize }
    menu:
      id: workspaces
      label: Workspace
      item:
        label: Move Left
        action: { name: SendToDesktop, to: left }
      item:
        label: Move Right
        action: { name: SendToDesktop, to: right }
      separator:
      item:
        label: Always on Visible Workspace
        action: { name: ToggleOmnipresent }
    item:
      label: Close
      action: { name: Close }

  - id: root-menu
    items:
      - label: Terminal
        action: { name: Execute, command: alacritty }
      - label: Reconfigure
        action: { name: Reconfigure }
      - label: Exit
        action: { name: Exit }
```
</details>

But I don't think this looks nicer than XML because this has duplicated `item` keys and YAML cannot represent attributes like `label` and `id`.


ahesford:
Although I would prefer YAML to XML for all labwc input files, I think this should only be merged if we intend to deprecate and eventually remove XML support. Retaining eternal support for both formats will increase maintenance burden and user confusion.

tokyo4j:
> Although I would prefer YAML to XML for all labwc input files, I think this should only be merged if we intend to deprecate and eventually remove XML support. Retaining eternal support for both formats will increase maintenance burden and user confusion.

Yeah, I've been having your concern as well. Adding complexity and confusion to make XML haters happy may be against our policy. But I believe providing another configuration format will make labwc more attractive to new users.

I'm looking into [KDL](https://kdl.dev/) (suggested by @vyivel in IRC) as another alternative. It seems to be XML-compatible, so maintaining KDL support would be much smaller than YAML.

However, it's new and breaking changes in [version 2](https://github.com/kdl-org/kdl/pull/286) is pending. Maybe we can see if it survives for some years and finally add support for it someday.

ahesford:
If KDL is up for discussion, then I definitely don't think we should be adding parallel YAML support; thrashing XML -> YAML -> KDL in a few years time will be a bad user experience.

I support moving to YAML (or anything more pleasant than XML to write and parse) with a reasonable deprecation period for XML. We can even offer an out-of-tree tool that converts openbox configurations to our expected format, which will both help users transition cleanly and allow us to reduce the deprecation period.

vyivel:
AFAICT labwc doesn't differentiate between attributes and children nodes when parsing XML, so switching to another node-based language like KDL doesn't make as much sense as I originally thought.

tokyo4j:
> If KDL is up for discussion, then I definitely don't think we should be adding parallel YAML support; thrashing XML -> YAML -> KDL in a few years time will be a bad user experience.

Agreed.

> I support moving to YAML (or anything more pleasant than XML to write and parse) with a reasonable deprecation period for XML. We can even offer an out-of-tree tool that converts openbox configurations to our expected format, which will both help users transition cleanly and allow us to reduce the deprecation period.

Deprecating XML sounds nice from developer side, but @Consolatis is against it (https://github.com/labwc/labwc/pull/2077#issuecomment-2292511206) and I support him. It sounds too breaking and may more or less block transition from Openbox.
Also, I think we should not deprecate `menu.xml` because layout configuration doesn't work well with YAML as I described above. But supporting only `rc.yaml` and `menu.xml` is confusing to users.

So, a possible solution is adding support for any XML-compatible language like KDL while keeping XML support with minimal maintenance cost and confusion to users.

> AFAICT labwc doesn't differentiate between attributes and children nodes when parsing XML, so switching to another node-based language like KDL doesn't make as much sense as I originally thought.

Yeah, though I think configuration in KDL looks nicer, it may be not worth it.

We currently allow `rc.xml` like:

```xml
  <core decoration="server"
        gap="0"
        adaptiveSync="no"
        allowTearing="no"
        reuseOutputMode="no"
        xwaylandPersistence="no" />
```

This is indeed not much different from KDL like:

```kdl
core {
  decoration="server"
  gap=0
  adaptiveSync="no"
  allowTearing="no"
  reuseOutputMode="no"
  xwaylandPersistence="no"
}
```

However, I think KDL works better for `menu.xml`.
For example, follwing `menu.xml`:

```xml
    <item label="Move Right">
      <action name="SendToDesktop" to="right" />
    </item>
    <separator />
    <item label="Always on Visible Workspace">
      <action name="ToggleOmnipresent" />
    </item>
```

can be written in KDL like:

```kdl
item label="Move Left" {
  action name="SendToDesktop" to="left"
}
item label="Move Right" {
  action name="SendToDesktop" to="right"
}
separator
item label="Always on Visible Workspace" {
  action name="ToggleOmnipresent"
}
```

But I'm not sure it's worth it.

johanmalm:
Thanks @tokyo4j ; this is a really nice PR.

I haven't yet looked at code in detail, but let's start with the broader direction.

I feel that we out ought to proceed with YAML in some way.

In my opinion KDL does not feel right for us; partly because it seems to lack a widely packaged C library, but also because it does not appear as elegant as YAML for our use-case (which in the case of rc.xml relates to a very simple grammar consisting of XML nodes and content only - where attributes are just treated as child nodes, and importantly where a single node never contain child-nodes AND content - only one or the other).

I am disregarding `menu.xml` in the discussion below because it contains more complex grammar (e.g. where attributed have a semantic meaning) and it feels more like a standard which external tools like obmenu-generator and lots of pipemenu extensions comply with.

So ignoring a do-nothing option here are four potential directions (1a, 1b, 1c, 2):

1. Leave `rcxml.c` nodename parsing as is. Just convert YAML to XML and pass it to `rcxml_parse_xml()`. This technically leaves the under-the-bonnet approach an XML one, albeit with a restrained grammar (as described above). Technically (read C code rather than documentation) this appears a low maintenance option and certainly does not duplicated effort. We would probably want some more unit tests with complete YAML/XML files and to give it some time to gain confidence that there aren't any odd corner cases. An importance consideration here is that this way maintains backward compatibility. If we decide on this as the direction, there are many approaches for the User Experience (UX). I will outline three here:

  a. Leave XML as the officially supported language in documentation. Make it clear that YAML is a second-class plugin. In `labwc-config(5)` only describe XML variants.

  b. Give YAML and XML equal importance in documentation and examples.

  c. Move to a position (possibly gradually) where YAML is the official config lingo. Obviously XML would still be supported, but would not be something we shout about or encourage. Documentation, examples and the `labwc-tweaks` tool would be YAML only. In way of an improvement we could consider stricter grammar + type schemas without breaking backward compatibility - which we can't at the moment. Another advantage to consider is increased interoperability between different libraries (which is hard with the current element/attribute agnostic approach) which could be useful if writing a tool in python, C++, rust, whatever to modify the config file.

2. Give up on `rcxml.c` and its XML/nodename parsing in favour of a direct YAML parsing implementation. This would simpify the code a bit (amount to be debated), but would most likely lead to a period of subtle breakages / regressions and could potentially take a while to grind out. We need to consider here that `action.c` (possibly others - cannot remember off the top of my head) is closely linked with `rcxml.c` and does some "downstream" parsing based on truncated nodenames.

My head is in the space of 1c as the best option - possibly using 1b as an interim stage... and feel we should take the plunge because:

- We should listen to users - and I've had many over the years who have brought this up
- I have now seen how much cleaner YAML can look compared with lab-flavoured XML
- YAML is widely packaged/supported
- It lays the foundation for using other libs/langs for config tools; as well as schemas. 
- It maintains backward compatibility.

tokyo4j:
Hmm... my mind is going back and forth as to how we should support another file format.

> c. Move to a position (possibly gradually) where YAML is the official config lingo. Obviously XML would still be supported, but would not be something we shout about or encourage. Documentation, examples and the labwc-tweaks tool would be YAML only.

This sounds nice, but I'm afraid this isn't much different from deprecating `rc.xml` because users already using XML may get confused when they see man pages to try out new features, even though it's not hard to find out how the news config options can be added to the existing `rc.xml`. So I think we'd better provide a CLI tool (which may be embedded in labwc binary) to convert `rc.xml` to `rc.yaml` if we take this approach.

> 2. Give up on rcxml.c and its XML/nodename parsing in favour of a direct YAML parsing implementation.

I think this is what we should do in the end if we take the option c. I agree that this is a dangerous approach, but I guess this keeps the codebase clean and straightforward, which labwc aims to be. In addition, if we internally convert XML to YAML in this approach, its logic can be shared with the CLI tool I mentioned above.

But... I'm starting to realize there's no perfect solution that keeps beauty, compatibility, maintainability and stability and feeling that we should stick to XML. I agree that `rc.yaml` looks much cleaner than `rc.xml` that we describe in the man page or `rc.xml.all`, but if we advertise more about how `rc.xml` could be written in cleaner way (e.g. `<focus  followMouseRequiresMovement="yes"  />` vs `<focus><followMouseRequiresMovement>yes</followMouseRequiresMovement></focus>`), I think we can make users realize XML isn't that bad. I want to emphasize that I'll never get less motivated even if this PR is discontinued. Exploring ways to improve software is always fun.

johanmalm:
I have been pondering what you wrote in your last post.

How about converting the code in this PR into a command line tool - like yaml2xml where we started and we then move that repo to the labwc orga. 

That way people can use yaml with a conversion tool with touching current code + documentation.

ahesford:
I like the idea of an external tool, maybe with "lab" in the name because it might do labwc-specific conversions. People can maintain YAML if they want, with a script that converts it before launching labwc. I can write a wrapper that we can maintain if there is interest.

jp7677:
I would  go fully `yaml` and additionally add a path for automatic conversion within `labwc` itself.  So first startup with a `yaml` capable labwc version should create a backup of `rc.xml`, write its own `rc.yaml` based on that and never look back after that. A few labwc major versions later the conversion code could be dropped completely.

System wide configurations would need manual conversion, but that would be unfortunately the burden of package maintainers.

johanmalm:
> I like the idea of an external tool, maybe with "lab" in the name because it might do labwc-specific conversions. People can maintain YAML if they want, with a script that converts it before launching labwc. I can write a wrapper that we can maintain if there is interest.

Thanks @ahesford 

@tokyo4j  - your call regarding @ahesford's offer there.

Ref: 
- https://github.com/johanmalm/yaml2xml/ (nowhere near as sophisticated/elegant as what @tokyo4j has done, but still).

johanmalm:
> I would go fully `yaml` and additionally add a path for automatic conversion within `labwc` itself. So first startup with a `yaml` capable labwc version should create a backup of `rc.xml`, write its own `rc.yaml` based on that and never look back after that. A few labwc major versions later the conversion code could be dropped completely.
> 
> System wide configurations would need manual conversion, but that would be unfortunately the burden of package maintainers.

Well - personally I'm up for this and have not particular attachment to XML. For me there are three prerequisites though:

1. Confidence by using CLI tools first (as discussed above)
2. Unanimous agreement from @ahesford @Consolatis @tokyo4j @jlindgren90 
3. A PR to prove that replacing the current node-name approach in `rcxml.c` and `action.c` feels tidy and elegant. I don't normally shy away from getting stuck in to write code, but I have to admit - on this one I prefer if somebody else did it. Suggest starting with (1) and (2) though to avoid misplaced effort.

jlindgren90:
I'm late to the party, sorry. At first I thought "this looks cool" but I'm less sold on it as I think about it. My concerns with adding a second config format are:

1. I do not think we should deprecate XML or make it a lesser-supported option any time soon. For me, easy migration from Openbox was a strong attraction of labwc initially. I would like to keep this ability.

2. Is it worth the extra maintenance burden for what is (unless I'm missing something) essentially a cosmetic improvement to a config file?

XML is not my favorite either, but my personal feeling is that continuity with existing configs (including Openbox configs) should be the most important consideration.

jlindgren90:
Agree with the idea of using an external tool to prove out any new syntax, rendering back to XML for now.

tokyo4j:
After some thoughts, my conclusion is supporting `rc.yaml` just as an external tool like suggested.

> How about converting the code in this PR into a command line tool - like yaml2xml where we started and we then move that repo to the labwc orga.
>
> That way people can use yaml with a conversion tool with touching current code + documentation.

> I like the idea of an external tool, maybe with "lab" in the name because it might do labwc-specific conversions. People can maintain YAML if they want, with a script that converts it before launching labwc. I can write a wrapper that we can maintain if there is interest.

I really like this idea. This avoids the confusion brought by mixing `rc.yaml` and `rc.xml` in the labwc documentation. For those who want to use `rc.yaml`, we can just say "go to https://github.com/labwc/labwc-yaml2xml" and isolate other YAML-related things like documentation and example `rc.yaml` into that repo.

@ahesford I appreciate if you write a wrapper. Just as a random idea, I thought rewriting the conversion tool in Python is more convenient for end users as it doesn't require compiling, but I'm not sure how much it benefits as it still requires a yaml library like PyYAML.

Conan-Kudo:
If there was an alternative config format to use instead of XML, I would put my hat in for TOML, because it's simpler and easier to understand. But it's not node-based and it is not possible to represent everything that XML can do in TOML (though obviously vice versa is possible).

The larger question to ask is what do we want out of configuration in labwc. If Openbox compatibility is key, then it needs to stay XML and we leave everything as-is. If simplicity and readability is key, then it would be prudent to at least consider TOML.

jjramsey:
I'll be honest; I'm not a huge fan of YAML, mainly because it uses whitespace in a way that has tripped me up (or at least it did when I had configured Yambar way back when). It's not quite as straightforward as Python's use of indentation.

While XML can be annoyingly verbose, I don't find that a huge problem with LabWC's XML files in particular, and it is fairly simple (or at least straightforward) to understand.

ringo32:
why change to yaml in the spirit of openbox ? if you know yaml sure would be easier, for me is same as xml :)  action namens in yaml looks just like leaves also lot space for user errors... dont see the difference...

Coccocoahelper:
As others have stated, I also believe the best way to go is having a conversion tool (YAML → XML, to be clear), instead of adding more code complexity. However, I haven't messed with labwc's code yet, nor actually used it (I will once I have a USB pen to install a distro on), so take this with a mountain of salt.

Note: johanmalm actually made [such a tool](https://github.com/johanmalm/yaml2xml), if anyone is curious.

Issue #2112, opened 2024-08-24
RFC: nomenclature for "Omnipresent" button and its theming


ahesford:
From https://github.com/labwc/labwc/issues/2111#issuecomment-2308527094:

I didn't read the Openbox spec when adding the omnipresent button. It seems what I call the "Omnipresent" mouse context is called "AllDesktops" in Openbox.

If we care about keeping this compatible with Openbox, I'd propose one of two things:

1. Moving the context name to "AllDesktops" and changing the theme variable to "desk" to agree with the image names, or
2. Accepting "AllDesktops" and "Omnipresent" as synonyms for the mouse context, and then accepting both "desk" and "omnipresent" as synonyms in the theme file. If both exist, the "omnipresent" variant would take precedence because it would be specific to labwc.

I think I favor the second option because "Omnipresent" seems a nicer description than "AllDesktops", but I could go either way.

Let's decide whether to act and, if so, which path to take, and I'll PR the favored fix.

Consolatis:
Personally I'd favor variant 1. Supporting both for the icon would mean 4 more icons (`(normal, toggled) * (no_hover, hover)`) and also complicate the logic when either loading the icons or selecting them. Regarding the mouse context I don't really mind either way; IMHO we should support whatever Openbox calls the context but may additionally also support `Omnipresent` but I don't feel strongly about it.

01micko:
With png/svg icons it's 8 more re #2114

Consolatis:
Implemented via
- #2115

PR #2114, opened 2024-08-25
docs: add description for new theme buttons 'shade' and desk'

### LINKED ISSUE #2113: document new ssd buttons
A description in `labwc-theme.5` is needed for the new 'desk' and 'shade' buttons; `xbm`, `png` and `svg`

01micko:
fixes #2113

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -318,6 +318,10 @@ file within a particular theme. The following xbm buttons are supported:
 - close.xbm
 - menu.xbm
 - max_toggled.xbm
+- desk.xbm
+- desk_toggled.xbm
+- shade.xbm
+- shade_toggled.xbm
 
 Additional icons can be defined to be shown when the mouse pointer is hovering
 over the button in question:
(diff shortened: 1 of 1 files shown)


01micko:
Making WIP because 
- labwc ootb doesn't have the additional icons by default
- they are supported so maybe they should go in the first section?

RFC

ahesford:
This should also wait for a resolution of #2112, so that we have consistent terminology.

What do you mean by "doesn't have the additional icons by default"?

01micko:
> What do you mean by "doesn't have the additional icons by default"?

When I switched to latest master the defaults were the same (maybe because of my rc.xml), that's all. Just was'nt sure to put them in the first or second stanza (for xbm). Happy to change if necessary.

I'm happy to wait. Just added the full set for svg/png and tested working well.

01micko:
Added some svg icons for testing if anyone likes. 

https://github.com/labwc/labwc-artwork/pull/19

Consolatis:
Thanks!

01micko:
Always a pleasure :)

PR #2116, opened 2024-08-25
ssd: allow ssd to be smaller than minimal size by removing buttons

### LINKED ISSUE #1947: Labwc misplaces titlebar on small window

tokyo4j:
Another alternative for #1959 and #1964.

This fixes #1947 by removing some buttons for tiny windows. For windows smaller than (button width)*2, corners
are un-rounded so a small titlebar can be rendered with scene-rect.

Demo (extents are colored for visualization):

https://github.com/user-attachments/assets/f155c9dc-6f20-438f-8b16-74e2603d19b0

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -53,12 +53,19 @@ struct ssd {
 		bool was_omnipresent;
 
 		/*
-		 * Corners need to be (un)rounded when toggling tiling or
-		 * maximization, and the button needs to be swapped on
+		 * Corners need to be (un)rounded and borders need be shown/hidden
+		 * when toggling maximization, and the button needs to be swapped on
 		 * maximization toggles.
 		 */
 		bool was_maximized;
-		bool was_tiled_not_maximized;
+
+		/*
+		 * Corners need to be (un)rounded but borders should be kept shown when
+		 * the window is (un)tiled and notified about it or when the window may
+		 * become so small that only a squared scene-rect can be used to render
+		 * such a small titlebar.
+		 */
+		bool was_squared;
 
 		struct wlr_box geometry;
 		struct ssd_state_title {
@@ -109,9 +116,6 @@ struct ssd_part {
 	/* This part represented in scene graph */
 	struct wlr_scene_node *node;
 
-	/* Targeted geometry. May be NULL */
-	struct wlr_box *geometry;
-
 	struct wl_list link;
 };
 
@@ -151,6 +155,7 @@ void ssd_destroy_parts(struct wl_list *list);
 void ssd_titlebar_create(struct ssd *ssd);
 void ssd_titlebar_update(struct ssd *ssd);
 void ssd_titlebar_destroy(struct ssd *ssd);
+bool ssd_should_be_squared(struct ssd *ssd);
 
 void ssd_border_create(struct ssd *ssd);
 void ssd_border_update(struct ssd *ssd);
--- a/src/ssd/ssd-extents.c
+++ b/src/ssd/ssd-extents.c
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0-only
 
+#include <assert.h>
 #include <pixman.h>
 #include "common/mem.h"
 #include "common/scene-helpers.h"
@@ -14,14 +15,7 @@ add_extent(struct wl_list *part_list, enum ssd_part_type type,
 {
 	float invisible[4] = { 0.0f, 0.0f, 0.0f, 0.0f };
 	struct ssd_part *part = add_scene_part(part_list, type);
-	/*
-	 * Extents need additional geometry to enable dynamic
-	 * resize based on position and output->usable_area.
-	 *
-	 * part->geometry will get free'd automatically in ssd_destroy_parts().
-	 */
 	part->node = &wlr_scene_rect_create(parent, 0, 0, invisible)->node;
-	part->geometry = znew(struct wlr_box);
 	return part;
 }
 
@@ -32,8 +26,6 @@ ssd_extents_create(struct ssd *ssd)
 	struct theme *theme = view->server->theme;
 	struct wl_list *part_list = &ssd->extents.parts;
 	int extended_area = SSD_EXTENDED_AREA;
-	int corner_size = extended_area + theme->border_width +
-		theme->window_button_width / 2;
 
 	ssd->extents.tree = wlr_scene_tree_create(ssd->tree);
 	struct wlr_scene_tree *parent = ssd->extents.tree;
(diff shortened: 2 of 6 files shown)


tokyo4j:
I'll also need to modify `tiled_not_maximized` in `ssd_update_geometry()`, but I don't get why `view_is_tiled()` is used here rather than `view_is_tiled_and_notify_tiled()`. @jp7677 any ideas?

https://github.com/labwc/labwc/blob/72f59c32a2d2ee33bf74452cea39c821ce60fbb2/src/ssd/ssd.c#L252

jp7677:
> I'll also need to modify `tiled_not_maximized` in `ssd_update_geometry()`, but I don't get why `view_is_tiled()` is used here rather than `view_is_tiled_and_notify_tiled()`. @jp7677 any ideas?
> 
> https://github.com/labwc/labwc/blob/72f59c32a2d2ee33bf74452cea39c821ce60fbb2/src/ssd/ssd.c#L252

From my memory: `view_is_tiled_and_notify_tiled()` includes if the view should be notified that it is tiled. But a view can also be tiled without being notified about it, since the notification is a user setting. So for purely describing the SSD state (which `tiled_not_maximized` is about), the notify setting should not be taken into account, thus `view_is_tiled()` fits better.
Does this makes sense?

tokyo4j:
> So for purely describing the SSD state (which tiled_not_maximized is about), the notify setting should not be taken into account, thus view_is_tiled() fits better.

Well, it makes sense if `view_is_tiled() && !maximized` is assigned to `ssd->state.was_tiled_not_maximized` and it's purely used for tracking state updates. However, in reality, `view_is_tiled_and_notify_notified() && !maximized` is assigned to `ssd->state.was_tiled_not_maximized` in `ssd_titlebar_update()`:

https://github.com/labwc/labwc/blob/72f59c32a2d2ee33bf74452cea39c821ce60fbb2/src/ssd/ssd-titlebar.c#L252-L263

and the logic of rendering borders in `ssd_border_update()` is directly dependent on it.

So I guess this inconsistency is just a small mistake (and I think this is not causing any bugs). But I wanted to make sure this is not intentional and modifying this doesn't trigger any side effects.

jp7677:
> So I guess this inconsistency is just a small mistake (and I think this is not causing any bugs). But I wanted to make sure this is not intentional and modifying this doesn't trigger any side effects.

Thanks. Yes, I think your assessment is correct and `view_is_tiled_and_notify_notified()` should be used indeed. May be wise to add a comment here https://github.com/labwc/labwc/blob/master/include/ssd-internal.h#L61 that this property makes the corners squared, but this behavior is only wanted when the view is also getting notified about being tiled.

src/ssd/ssd-extents.c
```diff
@@ -232,18 +220,12 @@ ssd_extents_update(struct ssd *ssd)
...
-			wlr_scene_node_set_position(part->node, target->x, target->y);
-		}
-		if (target->width != rect->width || target->height != rect->height) {
-			wlr_scene_rect_set_size(rect, target->width, target->height);
+				target.x + (result_box.x - part_box.x),
+				target.y + (result_box.y - part_box.y));
+		} else {
+			/* Fully visible */
+			wlr_scene_node_set_position(part->node, target.x, target.y);
+			wlr_scene_rect_set_size(rect, target.width, target.height);
```
tokyo4j, line 228: I removed checks like `target->x != part->node->x` as it's done in wlroots side.

src/ssd/ssd-extents.c
```diff
@@ -111,14 +76,15 @@ ssd_extents_update(struct ssd *ssd)
 	int full_width = width + 2 * theme->border_width;
 	int extended_area = SSD_EXTENDED_AREA;
 	int corner_size = extended_area + theme->border_width +
-		theme->window_button_width / 2;
+		MIN(theme->window_button_width, width) / 2;
```
tokyo4j, line 79: This fixes the overlap of corner extents and negative widths of top/bottom extents.

include/ssd-internal.h
```diff
@@ -109,9 +119,6 @@ struct ssd_part {
 	/* This part represented in scene graph */
 	struct wlr_scene_node *node;
 
-	/* Targeted geometry. May be NULL */
-	struct wlr_box *geometry;
-
```
tokyo4j, line 114: I removed `ssd_part.geometry` since now we calculate all the x/y/width/height of extent parts in `ssd_extents_update()`.

ahesford:
I pushed some suggested improvements to your branch. Please adopt as appropriate or drop the changes where you disagree:

1. Do not cache button counts. The arithmetic operations counts for calculating the counts on the fly, and the cost offsets the potential bugs that may result from de-syncing the cache:
   - In most cases, when the window is large enough, we only add a few memory accesses, an integer addition, an integer multiplication, and a comparison.
   - In cases where the sizes are reduced, we only have a couple more comparisons, and a division and subtraction each.

2. Simplify the counting and attempt to take approximately equally from both sides as the window size is reduced.

ahesford:
Also, I don't like overloading the name "tiled_not_maximized" to also mean "tiled and not maximized or too small". I think we need a better name for the variable and for the function that computes it.

tokyo4j:
Thank you for advises.

> Do not cache button counts. The arithmetic operations counts for calculating the counts on the fly, and the cost offsets the potential bugs that may result from de-syncing the cache:

I don't think your change prevents the potential de-synchronization because with your change, forgetting to call `update_visible_buttons()` leads to out-of-sync return values from `count_visible_buttons()`. I think the most reliable approach is to count the number of enabled scene nodes for buttons in `count_visible_buttons()` rather than re-calculating it. But I prefer just adding a comment like "these should always be synchronized with enabled/disabled states for button scene-nodes" besides `visible_button_count_{left,right}` declarations.

> 2. Simplify the counting and attempt to take approximately equally from both sides as the window size is reduced.

Your code looks clean, but I noticed the corner-left button is not lastly hidden if `(left button count) < (right button count)`. I want the corner-left button to lastly hidden because it's the window-menu button by default, which has the functionalities of all the other buttons. Also if we replace it with the app icon in the future, it will be more meaningful than the close button.

It can be easily fixed by changing like:
```diff
- num_visible.left = MIN(num_visible.left, num_visible_total / 2);
+ num_visible.left = MIN(num_visible.left, (num_visible_total+1) / 2);
```
...but I came up with the dumbest, but quite readable approach:

```c
	while (width < button_width * (button_count_left + button_count_right)) {
		if (button_count_left > button_count_right) {
			button_count_left--;
		} else {
			button_count_right--;
		}
	}
```

I believe the overhead of this is negligible as the maximum number of button is small and limited.

src/ssd/ssd-titlebar.c
```diff
@@ -241,6 +244,58 @@ set_alt_button_icon(struct ssd *ssd, enum ssd_part_type type, bool enable)
...
+	 * The corner-left button is lastly removed as it's usually a window
+	 * menu or app icon button.
+	 */
+	while (width < button_width * (button_count_left + button_count_right)) {
+		if (button_count_left > button_count_right) {
+			button_count_left--;
+		} else {
+			button_count_right--;
+		}
+	}
```
ahesford, line 271: Here's a deterministic algorithm that can be executed without loops and should preserve the behavior you desire:
```c
int max_buttons = width / button_width; 
int lb = MIN(button_count_left, (max_buttons + 1) / 2);
int rb = MIN(button_count_right, max_buttons / 2);
button_count_left = MIN(button_count_left, max_buttons - rb);
button_count_right = MIN(button_count_right, max_buttons - button_count_left);
```
I've enumerated all possible combinations of (left, right) configurations in our six-button set and confirmed that the removals are approximately uniform, while ensuring that the first left button does not disappear unless it was never configured in the first place, or the number of buttons allowed is identically zero.

src/ssd/ssd-titlebar.c
```diff
@@ -241,6 +244,58 @@ set_alt_button_icon(struct ssd *ssd, enum ssd_part_type type, bool enable)
...
+		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
+			part = ssd_get_part(&subtree->parts, b->type);
+			wlr_scene_node_set_enabled(part->node,
+				button_count < button_count_right);
+			button_count++;
+		}
+	} FOR_EACH_END
+
+	ssd->state.visible_button_count_left = button_count_left;
+	ssd->state.visible_button_count_right = button_count_right;
```
ahesford, line 296: I understand the remaining de-sync issue, but still don't like caching these. Would looping through the SSD buttons and counting whether they are enabled in the scene be terrible?

ahesford (commented):
I like the change to "squared".

src/ssd/ssd-titlebar.c
tokyo4j: Well, I usually don't like math and I prefer the simplest code as long as it doesn't have a noticeable overhead.

My code makes it clear that `button_count_{left,right}` are just `wl_list_length(&rc.title_buttons_{left,right})` when the view width is large enough and how button counts are decremented as the view shrinks.

src/ssd/ssd-titlebar.c
tokyo4j: Not terrible at all. It's just my weak preference and I understand adding states to `struct ssd` for this minor feature is undesirable.

I made a fixup commit that counts the number of enabled button nodes instead of caching it.
I still slightly prefer caching it as it's less verbose, but I'll squash it if you like it.

ahesford: I think the uncached version is better, but we could see what the others think.

src/ssd/ssd-titlebar.c
ahesford: If you want to keep the loop, you might consider setting

```c
int button_width = MAX(0, view->server->theme->window_button_width);
```

to guard against some invalid value inadvertently triggering an infinite loop.

tokyo4j: Good idea. I added `assert(button_width > 0)` as negative values are already checked when reading `themerc`:

https://github.com/labwc/labwc/blob/b34f2c9d85a52bc19a9458b6d38bae64fae55baa/src/theme.c#L740-L747

src/ssd/ssd-titlebar.c
johanmalm: Yeah, I much prefer the un-cached version (fixup-commit).
Caching is always harder to maintain and given the context, a few extra loops is highly unlikely to be expensive.
Thanks for having the discussion on this.

johanmalm:
This looks a great PR. Really like the solution. Thanks for the work on it.

src/ssd/ssd-titlebar.c
tokyo4j: Alright. I squashed the commit with some minor improvements. Thank your for your advises!

PR #2118, opened 2024-08-26
"WarpCursor" action

orfeasxyz:
This is a simple action that warps the cursor to the middle of the currently focused view. Useful when moving windows or changing focus using keyboard shortcuts.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -340,6 +340,19 @@ Actions are used in menus and keyboard/mouse bindings.
 	decorations (including those for which the server-side titlebar has been
 	hidden) are not eligible for shading.
 
+*<action name="WarpCursor" to="output" x="center" y="center" />*
+	Warp the cursor to a position relative to the active output or window.
+
+	*to* [output|window] Specifies the target area of the warp.
+	Default is "output"
+	*x* [center|value] Specifies the horizontal warp position within the target area.
+	"center": Moves the cursor to the horizontal center of the target area.
+	Positive or negative integers warp the cursor to a position offset by the specified
+	number of pixels from the left or right edge of the target area respectively.
+	Default is "center"
+	*y* [center|value] Equivalent for the vertical warp position within the target area.
+	Default is "center"
+
 *<action name="EnableTabletMouseEmulation" />*++
 *<action name="DisableTabletMouseEmulation" />*++
 *<action name="ToggleTabletMouseEmulation">*
--- a/src/action.c
+++ b/src/action.c
@@ -121,7 +121,8 @@ enum action_type {
 	ACTION_TYPE_TOGGLE_TABLET_MOUSE_EMULATION,
 	ACTION_TYPE_TOGGLE_MAGNIFY,
 	ACTION_TYPE_ZOOM_IN,
-	ACTION_TYPE_ZOOM_OUT
+	ACTION_TYPE_ZOOM_OUT,
+	ACTION_TYPE_WARP_CURSOR,
 };
 
 const char *action_names[] = {
@@ -186,6 +187,7 @@ const char *action_names[] = {
 	"ToggleMagnify",
 	"ZoomIn",
 	"ZoomOut",
+	"WarpCursor",
 	NULL
 };
 
@@ -472,6 +474,12 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			} else {
 				action_arg_add_int(action, argument, policy);
 			}
+			goto cleanup;
+		}
+		break;
+	case ACTION_TYPE_WARP_CURSOR:
+		if (!strcmp(argument, "to") || !strcmp(argument, "x") || !strcmp(argument, "y")) {
+			action_arg_add_str(action, argument, content);
 			goto cleanup;
 		}
 		break;
@@ -822,6 +830,43 @@ get_target_output(struct output *output, struct server *server,
 	}
 
 	return target;
+}
+
+static void
+warp_cursor(struct view *view, struct output *output, const char *to, const char *x, const char *y)
+{
+	struct wlr_box target_area = {0};
+	int goto_x;
+	int goto_y;
+
+	if (!strcasecmp(to, "output") && output) {
+		target_area = output_usable_area_in_layout_coords(output);
+	} else if (!strcasecmp(to, "window") && view) {
+		target_area = view->current;
+	} else {
+		wlr_log(WLR_ERROR, "Invalid argument for action WarpCursor: 'to' (%s)", to);
+	}
+
+	if (!strcasecmp(x, "center")) {
+		goto_x = target_area.x + target_area.width / 2;
+	} else {
+		int offset_x = atoi(x);
+		goto_x = offset_x >= 0 ?
+			target_area.x + offset_x :
+			target_area.x + target_area.width + offset_x;
+	}
+
+	if (!strcasecmp(y, "center")) {
+		goto_y = target_area.y + target_area.height / 2;
+	} else {
+		int offset_y = atoi(y);
+		goto_y = offset_y >= 0 ?
+			target_area.y + offset_y :
+			target_area.y + target_area.height + offset_y;
+	}
+
+	wlr_cursor_warp(output->server->seat.cursor, NULL, goto_x, goto_y);
+	cursor_update_focus(output->server);
 }
 
 void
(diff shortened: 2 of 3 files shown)


ahesford (commented):
I wouldn't use this feature if merged, but don't really have a problem with it. However, I don't think `WarpCursor` is the right name. It's too vague to convey that the cursor will be centered on the active view.

docs/labwc-actions.5.scd
```diff
@@ -80,6 +80,9 @@ Actions are used in menus and keyboard/mouse bindings.
 
 	*height* The height to resize the window to in pixels.
 
+*<action name="WarpCursor" />*
+    Warp the cursor to the middle of the currently focused view.
```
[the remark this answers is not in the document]
01micko: Apart from that `scdoc` only accepts tabs, will error otherwise.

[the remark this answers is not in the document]
01micko: A bit strange that only the `void` runner error-ed on sdcoc.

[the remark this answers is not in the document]
01micko: I see, not a hard dependency, debian, if working would have failed.

orfeasxyz:
I made a few changes, including renaming, proper checking and fixing the doc.

src/view.c
```diff
@@ -2296,6 +2296,16 @@ view_toggle_keybinds(struct view *view)
 	}
 }
 
+void
+view_warp_cursor_to_focused(struct view *view)
```
ahesford, line 2300: `view_center_cursor` is probably a better name for this function, because its action has nothing to do with the view being focused.

docs/labwc-actions.5.scd
```diff
@@ -80,6 +80,9 @@ Actions are used in menus and keyboard/mouse bindings.
 
 	*height* The height to resize the window to in pixels.
 
+*<action name="WarpCursorToFocused" />*
```
ahesford, line 83: `CenterCursor` seems like a brief but more meaningful name, and could also be extended to center on the currently active output if no window is focused.

orfeasxyz:
I reworked the action to optionally center the cursor to the output if no view is currently being focused and did a proper renaming as it was suggested.

tokyo4j:
I see moving the cursor to the center of the focused view is useful, but I'm not sure how we should design an action for it in a future-proof way.

Personally, I prefer the original `WarpCursor` so we can add optional parameters like `<action name="WarpCursor" to="window|output" x="50%" y="50%" />` in the future. I found an action (dispatcher) in Hyprland called `movecursortocorner` which moves the cursor to the corner of the active window. So there may be some demand for setting destination of the cursor, though I'm not sure if it's actually useful.

ahesford:
If we're going to admit the possibility of warping to arbitrary (proportional) coordinates, I'd rather see that done right away. Calling it "WarpCursor" with no arguments isn't very obvious, and calling it "CenterCursor" now and then adding a new action in the future is undesirable.

Extending this now to support configurable fractions other than 50% should be straightforward.

tokyo4j:
> If we're going to admit the possibility of warping to arbitrary (proportional) coordinates, I'd rather see that done right away. Calling it "WarpCursor" with no arguments isn't very obvious, and calling it "CenterCursor" now and then adding a new action in the future is undesirable.

Agreed. Anyway I want to hear opinions from @johanmalm @Consolatis when they're back.

johanmalm:
I like `WarpCursor`. Suggest making the `to`, `x` and `y` options mandatory.

Maybe something like:

```
<action name="WarpCursor">
    <to>window|output</to>
    <x>center</x>
    <y>center</y>
</action>
```

Ref:
- https://redtide.github.io/openbox-wiki/Help/Actions/#moveresizeto

johanmalm:
@orfeasxyz  - gentle nudge on this one. Did you want to finish off implementation?

orfeasxyz:
I got hooked with the query related stuff and forgot about this. I will work on this over the next few days and send the patch.

orfeasxyz:
I pushed my changes. I ended up following @johanmalm's suggestion of using "to" "x" and "y" as action arguments to specify different offsets and targets (either the current view or the current desktop). If everything is ok I can go ahead and write the docs.

src/action.c
```diff
@@ -824,6 +832,53 @@ get_target_output(struct output *output, struct server *server,
...
 
+static void
+warp_cursor(struct view *view, struct output *output, const char *to, const char *x, const char *y)
+{
+	struct wlr_box target_area = {};
+	int goto_x;
+	int goto_y;
+
+	if (!strcasecmp(to, "output") && output) {
+		target_area = output->usable_area;
```
tokyo4j, line 843: You can use `output_usable_area_in_layout_coords()` to the geometry of the usable area in layout coodinates. That way you can simplify the logic below in 868-877.

src/action.c
```diff
@@ -824,6 +832,53 @@ get_target_output(struct output *output, struct server *server,
...
+	} else if (!strcasecmp(to, "view") && view) {
```
tokyo4j, line 844: "View" is just an internal term and we should not expose it to users. "Window" should be suitable here.

src/action.c
```diff
@@ -824,6 +832,53 @@ get_target_output(struct output *output, struct server *server,
 	return target;
 }
 
+static void
+warp_cursor(struct view *view, struct output *output, const char *to, const char *x, const char *y)
+{
+	struct wlr_box target_area = {};
```
tokyo4j, line 838: nit: we usually use `= {0}` rather than `= {}`.

src/action.c
```diff
@@ -121,7 +121,8 @@ enum action_type {
 	ACTION_TYPE_TOGGLE_TABLET_MOUSE_EMULATION,
 	ACTION_TYPE_TOGGLE_MAGNIFY,
 	ACTION_TYPE_ZOOM_IN,
-	ACTION_TYPE_ZOOM_OUT
+	ACTION_TYPE_ZOOM_OUT,
+	ACTION_TYPE_WARP_CURSOR
```
tokyo4j, line 125: Not specific to this PR, but I suggest adding a comma also to the last element in the enum so that we can minimize the diff when we add more actions. AFAIK we don't have any coding rule to prohibit that.

src/action.c
```diff
@@ -1287,6 +1342,16 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_ZOOM_OUT:
 			magnify_set_scale(server, MAGNIFY_DECREASE);
 			break;
+		case ACTION_TYPE_WARP_CURSOR:
+		{
```
tokyo4j, line 1346: I guess the open bracket should be at the end of the previous line. We should document that in `CONTRIBUTING.md`.

```suggestion
		case ACTION_TYPE_WARP_CURSOR: {
```

tokyo4j: Oh, sorry I didn't realize there's already `case ACTION_TYPE_VIRTUAL_OUTPUT_REMOVE` that follows your indentation.

Fair enough. Let's keep it as it is.

tokyo4j:
Thanks. Code LGTM and it works fine in my test.

I think we can merge this PR after you write the documentation for WarpCursor.

orfeasxyz:
FYI I've written the documentation, you may go ahead and merge.

johanmalm:
Thanks for the work on this. LGTM.

ManuLinares:
I don't get the name. I don't know what "warp" is. This action should be called "MoveCursorTo"

johanmalm:
> I don't get the name. I don't know what "warp" is. This action should be called "MoveCursorTo"

Guess it comes from XWarpPointer

https://tronche.com/gui/x/xlib/input/XWarpPointer.html

Choosing names is hard. In hindsight I think your suggestion is better, but a breaking change is undesireable so prefer to leave as is.

ahesford:
"Warp" conveys more than just motion, but an instantaneous teleportation to the destination.

ManuLinares:
"Warp" is not a common term outside X11 systems, while "MoveCursorTo" is immediately understandable.

> Choosing names is hard. In hindsight I think your suggestion is better, but a breaking change is undesireable so prefer to leave as is.

The goal of pre-1.0.0 versions is to experiment, iterate, and refine the API before committing to long-term stability. The naming should align with standard user expectations and other APIs.

Maybe a deprecation system is needed, even though this is a pre-1.0.0 version, users may already rely on the existing API. A deprecation warning ensures that users are informed of upcoming changes without breaking their workflow immediately.


I suggest discussing/implementing a deprecation system.

1. Keep WarpCursor but mark it as deprecated.
2. Introduce MoveCursorTo as the preferred function.
3. Log a warning when WarpCursor is used, encouraging users to switch.
4. Fully remove WarpPointer in a later version (e.g., 0.9.0 or 1.0.0).

ahesford:
We've already noted
1. The suitability of the term "warp" because of the connotation of the word,
2. A desire not to introduce breaking changes, with or without a deprecation.

The matter is settled. Please refrain from further debate.

Salmanfarisy:
> This is a simple action that warps the cursor to the middle of the currently focused view. Useful when moving windows or changing focus using keyboard shortcuts.

How do you warp the cursor when changing focus using Alt-tab?

PR #2124, opened 2024-09-02
Add openbox style decoration actions

droc12345:
- Decorate, set full decorations, ie titlebar and border
- Undecorate, turn off all decorations
- Border, border only decoration, unaffected by keepBorder setting

Brought about by discussions in  #2070

Edit to add: this does not undo the 3-way toggle decorations, though IMO, we should go back to OB behavior
ie, two states and use the border action if a border is wanted.
keepBorder yes - toggle between full and border
keepBorder no  - toggle between full and none (current behavior)

ETA2: On my home system, I changed the 3 way action back to 2 way (matching openbox)
and it seems to work well, toggles as mentioned in previous ETA, and the direct actions of
Decorate, Undecorate and Border handle just those states. Works well on my end.
Let me know if you want to change the 3 state toggle back to 2 states.
Also not sure about the action name "Border" maybe change to "BorderOnly"?

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -134,6 +134,18 @@ Actions are used in menus and keyboard/mouse bindings.
 	windows that have client side decorations if it would increase the
 	amount of decoration. Default is no.
 
+*<action name="Decorate" />*
+	Similar to ToggleDecorations below but a single state
+	- Titlebar and borders
+
+*<action name="Undecorate" />*
+	Similar to ToggleDecorations below but a single state
+	- No titlebar or borders
+
+*<action name="Border" />*
+	Similar to ToggleDecorations below but a single state
+	- Borders only
+
 *<action name="ToggleDecorations" />*
 	Toggle decorations of focused window.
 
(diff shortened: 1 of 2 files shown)


johanmalm:
@xi I remember you did some analysis on this. Any thoughts on actions/options regarding this?

xi:
Yes, there was quite some discussion around this. We went through 5 different designs and eventually settled on `SetDecorations` in #1733. @droc12345 I think your actions all have equivalents with the solution:

- Undecorate: `<action name="SetDecorations" decorations="none" forceSSD="yes" />`
- Border: `<action name="SetDecorations" decorations="border" forceSSD="yes" />`
- Decorate: `<action name="SetDecorations" decorations="full" forceSSD="yes" />`

Or am I missing something?

For what it's worth, I also don't like the 3-way toggle (see https://github.com/labwc/labwc/pull/1725#issuecomment-2065260652), but I understand if backwards compatibility is more important.

droc12345:
All I've done is access part of ToggleDecorations, by supplying an action name.

Since it's not needed, I'll close it shortly.

PR #2127, opened 2024-09-07
theme: add circle hover effect for window buttons

jp7677:
I've played a bit with the hover effects and came up with this simple version that is nicely consistent with GTK CSD applications I’m using and also plays nicely with squared corners ( win-win ;) ).

ssd

I would love to get this ready for review/merge. Obviously I should put this behind an option to not break existing setups. Furthermore I guess, but not sure if it is needed, the radius of the hover circle should also be a theme option. Any thoughts on where to put and how to name those options (`rc.xml` or `theme`)?

My suggestions:
Since we already have `<theme><cornerRadius>` may be `<theme><hoverRadius>` could work, whereas  `hoverRadius == 0` would be no rounded hover effects (thus keeping the current hover effect) and any value `> 1` would be applied as is (if it fits the max possible radius).

Alternatively `<theme><roundedHover>` could be a `true/false` switch and we hard code the radius to 70% like I did now.

PS: What about disabling the CI run on Debian until Debian is ready to produce builds again? Getting the failed CI notifications is, well, misleading.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -47,12 +47,18 @@ labwc-config(5).
 # THEME ELEMENTS
 
 *border.width*
-	Line width (integer) of border border drawn around window frames.
+	Line width (integer) of border drawn around window frames.
 	Default is 1.
 
+*padding.width*
+	Horizontal padding size, in pixels, between border and first
+	button on the left/right.
+	Default is 0.
+
 *padding.height*
-	Vertical padding size, used for spacing out elements in the window
-	decorations. Default is 3.
+	Vertical padding size, in pixels, used for spacing out elements
+	in the window decorations.
+	Default is 3.
 
 *titlebar.height*
 	Window title bar height.
@@ -121,6 +127,10 @@ labwc-config(5).
 *window.button.width*
 	Width of a titlebar button, in pixels.
 	Default is 26.
+
+*window.button.spacing*
+	Space between titlebar buttons, in pixels.
+	Default is 0.
 
 *window.active.button.unpressed.image.color*
 	Color of the images in titlebar buttons in their default, unpressed,
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -132,6 +132,10 @@ labwc-config(5).
 	Space between titlebar buttons, in pixels.
 	Default is 0.
 
+*window.button.hover.bg.shape*
+	The shape of the hover effect of a titlebar button: "rectangle" or "circle".
+	Default is "rectangle".
+
 *window.active.button.unpressed.image.color*
 	Color of the images in titlebar buttons in their default, unpressed,
 	state. This element is for the focused window.
--- a/include/view.h
+++ b/include/view.h
@@ -538,6 +538,7 @@ const char *view_get_string_prop(struct view *view, const char *prop);
 void view_update_title(struct view *view);
 void view_update_app_id(struct view *view);
 void view_reload_ssd(struct view *view);
+int view_get_min_width(void);
 
 void view_set_shade(struct view *view, bool shaded);
 
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -116,6 +116,7 @@ ssd_titlebar_create(struct ssd *ssd)
 	struct view *view = ssd->view;
 	struct theme *theme = view->server->theme;
 	int width = view->current.width;
+	int corner_width = ssd_get_corner_width();
 
 	float *color;
 	struct wlr_scene_tree *parent;
@@ -144,25 +145,25 @@ ssd_titlebar_create(struct ssd *ssd)
 
 		/* Background */
 		add_scene_rect(&subtree->parts, LAB_SSD_PART_TITLEBAR, parent,
-			width - theme->window_button_width * 2, theme->title_height,
-			theme->window_button_width, 0, color);
+			width - corner_width * 2, theme->title_height,
+			corner_width, 0, color);
 		add_scene_buffer(&subtree->parts, LAB_SSD_PART_TITLEBAR_CORNER_LEFT, parent,
 			corner_top_left, -rc.theme->border_width, -rc.theme->border_width);
 		add_scene_buffer(&subtree->parts, LAB_SSD_PART_TITLEBAR_CORNER_RIGHT, parent,
-			corner_top_right, width - theme->window_button_width,
+			corner_top_right, width - corner_width,
 			-rc.theme->border_width);
 
 		/* Buttons */
 		struct title_button *b;
-		int x = 0;
+		int x = theme->padding_width;
 		wl_list_for_each(b, &rc.title_buttons_left, link) {
 			add_button(ssd, subtree, b->type, x);
-			x += theme->window_button_width;
+			x += theme->window_button_width + theme->window_button_spacing;
 		}
 
-		x = width;
+		x = width - theme->padding_width + theme->window_button_spacing;
 		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
-			x -= theme->window_button_width;
+			x -= theme->window_button_width + theme->window_button_spacing;
 			add_button(ssd, subtree, b->type, x);
 		}
 	} FOR_EACH_END
@@ -197,11 +198,12 @@ set_squared_corners(struct ssd *ssd, bool enable)
 {
 	struct view *view = ssd->view;
 	int width = view->current.width;
+	int corner_width = ssd_get_corner_width();
 	struct theme *theme = view->server->theme;
 
 	struct ssd_part *part;
 	struct ssd_sub_tree *subtree;
-	int x = enable ? 0 : theme->window_button_width;
+	int x = enable ? 0 : corner_width;
 
 	FOR_EACH_STATE(ssd, subtree) {
 		part = ssd_get_part(&subtree->parts, LAB_SSD_PART_TITLEBAR);
@@ -252,18 +254,23 @@ static void
 update_visible_buttons(struct ssd *ssd)
 {
 	struct view *view = ssd->view;
-	int width = view->current.width;
+	int width = view->current.width - (2 * view->server->theme->padding_width);
 	int button_width = view->server->theme->window_button_width;
+	int button_spacing = view->server->theme->window_button_spacing;
 	int button_count_left = wl_list_length(&rc.title_buttons_left);
 	int button_count_right = wl_list_length(&rc.title_buttons_right);
 
 	/* Make sure infinite loop never occurs */
 	assert(button_width > 0);
+
 	/*
 	 * The corner-left button is lastly removed as it's usually a window
 	 * menu button (or an app icon button in the future).
 	 */
-	while (width < button_width * (button_count_left + button_count_right)) {
+	while (width <
+		((button_width * (button_count_left + button_count_right)) +
+			(MAX((button_count_right - 1), 0) * button_spacing) +
+			(MAX((button_count_left - 1), 0) * button_spacing))) {
 		if (button_count_left > button_count_right) {
 			button_count_left--;
 		} else {
@@ -299,6 +306,7 @@ ssd_titlebar_update(struct ssd *ssd)
 {
 	struct view *view = ssd->view;
 	int width = view->current.width;
+	int corner_width = ssd_get_corner_width();
 	struct theme *theme = view->server->theme;
 
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
@@ -335,27 +343,29 @@ ssd_titlebar_update(struct ssd *ssd)
 	struct ssd_part *part;
 	struct ssd_sub_tree *subtree;
 	struct title_button *b;
-	int bg_offset = maximized || squared ? 0 : theme->window_button_width;
+	int bg_offset = maximized || squared ? 0 : corner_width;
 	FOR_EACH_STATE(ssd, subtree) {
 		part = ssd_get_part(&subtree->parts, LAB_SSD_PART_TITLEBAR);
 		wlr_scene_rect_set_size(
 			wlr_scene_rect_from_node(part->node),
 			width - bg_offset * 2, theme->title_height);
 
-		x = 0;
+		x = theme->padding_width;
 		wl_list_for_each(b, &rc.title_buttons_left, link) {
 			part = ssd_get_part(&subtree->parts, b->type);
 			wlr_scene_node_set_position(part->node, x, 0);
-			x += theme->window_button_width;
+			x += theme->window_button_width + theme->window_button_spacing;
 		}
 
-		x = width - theme->window_button_width;
+		x = width - corner_width;
 		part = ssd_get_part(&subtree->parts, LAB_SSD_PART_TITLEBAR_CORNER_RIGHT);
 		wlr_scene_node_set_position(part->node, x, -rc.theme->border_width);
+
+		x = width - theme->padding_width + theme->window_button_spacing;
 		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
 			part = ssd_get_part(&subtree->parts, b->type);
+			x -= theme->window_button_width + theme->window_button_spacing;
 			wlr_scene_node_set_position(part->node, x, 0);
-			x -= theme->window_button_width;
 		}
 	} FOR_EACH_END
 	ssd_update_title(ssd);
@@ -457,19 +467,23 @@ get_title_offsets(struct ssd *ssd, int *offset_left, int *offset_right)
 {
 	struct ssd_sub_tree *subtree = &ssd->titlebar.active;
 	int button_width = ssd->view->server->theme->window_button_width;
-	*offset_left = 0;
-	*offset_right = 0;
+	int button_spacing = ssd->view->server->theme->window_button_spacing;
+	int padding_width = ssd->view->server->theme->padding_width;
+	*offset_left = padding_width;
+	*offset_right = padding_width;
 
 	struct title_button *b;
 	wl_list_for_each(b, &rc.title_buttons_left, link) {
 		struct ssd_part *part = ssd_get_part(&subtree->parts, b->type);
 		if (part->node->enabled) {
+			*offset_left += *offset_left > padding_width ? button_spacing : 0;
 			*offset_left += button_width;
 		}
 	}
 	wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
 		struct ssd_part *part = ssd_get_part(&subtree->parts, b->type);
 		if (part->node->enabled) {
+			*offset_right += *offset_right > padding_width ? button_spacing : 0;
 			*offset_right += button_width;
 		}
 	}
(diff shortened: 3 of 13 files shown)


johanmalm:
I like this - in the sense that I much prefer circles over rectangles. However, in terms of how we configure this, we need to tread carefully and with thought. The overlay is fundamentally an old hack - not in a technical sense but in terms of not being given much thought regarding extensibility and maintainability. Also, it does completely ignore openbox spec.

Initial considerations:

1. Openbox uses themerc settings for the hover textures for example `window.active.button.hover.bg` (see references). If you look at the 'Texture' section you'll see definitions like 'Solid Flat'. As a starting point I think it is preferable to align with the openbox way and define the hover overlays with `.bg` syntax.

2. Openbox doesn't support shapes other than squares so we could consider just adding a `.bg.shape` variable rather than making the `.bg` syntax more complex. Although it may clash with 'Raised' and 'Sunken' if we do it that way.

3. The default Openbox theme `Clearlooks-3.4` defaults to `Flat Border Gradient Vertical` for the button `.bg`s which I'm pretty sure we don't want to copy. So if we just agree that any future support of `.bg` would default to a solid colour (i.e. not a gradient) and 'no border' then we don't have to do anything with that for now.

4. Let's consider how to configure hover-overlay sizes next. An openbox `.bg` texture just fills the 'allocated' space - i.e. there is no way of setting `.bg` texture sizes directly. I think that the best way to move forward on this is to complete the theme variables relating to buttons and their position. For example, the horizontal spacing variable (`padding.width`) and the width/height variable (for which we already have `window.button.width`).

As an aside, I suggest using `2 * M_PI` instead of `360 * deg` (I'm probably guilty of introducing the original `deg` thing, but hey :smile:).

Ref:
- https://redtide.github.io/openbox-wiki/Help/Themes/#textures
- https://redtide.github.io/openbox-wiki/Help/Themes/#window.active.button.hover.bg
- https://redtide.github.io/openbox-wiki/Help/Themes/#paddingwidth
- https://github.com/labwc/labwc/commit/e4afa10fe42ad0716f75ec47f4b3051f65cf59f0

droc12345:
Maybe a silly question, since the original wasn't extensible, wouldn't this be the perfect time to re-engineer things?
I have not looked at the code, so no idea how much work would be needed.

jp7677:
> I like this - in the sense that I much prefer circles over rectangles. However, in terms of how we configure this, we need to tread carefully and with thought. The overlay is fundamentally an old hack - not in a technical sense but in terms of not being given much thought regarding extensibility and maintainability. Also, it does completely ignore openbox spec.

Thanks  for the detailed reply. I don't have a history with openbox, but sounds good to keep as close as possible!

>  2. Openbox doesn't support shapes other than squares so we could consider just adding a `.bg.shape` variable rather than making the `.bg` syntax more complex. Although it may clash with 'Raised' and 'Sunken' if we do it that way.

I've added `window.button.hover.bg.shape`. Since this applies to active and inactive windows, I think this fits better than `window.active.button.hover.bg.shape` . It also aligns with `window.button.width`. Ok like this?

> 4. Let's consider how to configure hover-overlay sizes next. An openbox `.bg` texture just fills the 'allocated' space - i.e. there is no way of setting `.bg` texture sizes directly. I think that the best way to move forward on this is to complete the theme variables relating to buttons and their position. For example, the horizontal spacing variable (`padding.width`) and the width/height variable (for which we already have `window.button.width`).

So you are suggesting we introduce `window.button.hover.bg.padding.width` and `window.button.hover.bg.padding.width`. Then calculate the circle radius with `(MIN(width - (2 * padding_width), height - (2 * padding_height)) / 2)`?
 
> As an aside, I suggest using `2 * M_PI` instead of `360 * deg` (I'm probably guilty of introducing the original `deg` thing, but hey 😄).

Even when including `math.h`, I couldn't find `M_PI`. Looks like there it is not that easy and more needed to have this macro available.

johanmalm:
> I've added `window.button.hover.bg.shape`. Since this applies to active and inactive windows, I think this fits better than `window.active.button.hover.bg.shape` . It also aligns with `window.button.width`. Ok like this?

Yes. Happy with `window.button.hover.bg.shape`

> So you are suggesting we introduce `window.button.hover.bg.padding.width` and `window.button.hover.bg.padding.height`. Then calculate the circle radius with `(MIN(width - (2 * padding_width), height - (2 * padding_height)) / 2)`?

Just `padding.width`. It's one of those lonely variables at the op of themerc that feels somewhat global.

I think the hover-circle should have the diameter of `window.button.width`.

But if we implemented `.bg.shape` for the `unpressed` state, we'd have three (for iconify, maximize and close) circles touching each other because there is no spacing. So we need `padding.width` between them.

I'm making an assumption here that `window.button.width` is generally greater than the icon itself - but can't remember if we scale up. If we do, we should probably stop doing that!

johanmalm:
> Maybe a silly question, since the original wasn't extensible, wouldn't this be the perfect time to re-engineer things? I have not looked at the code, so no idea how much work would be needed.

I didn't say that it wasn't extensible... just that no thought was given to how it would be extended. So now that we're at the point of wanting to extend it, I want to carefully consider how we do it.

jp7677:
> Just `padding.width`. It's one of those lonely variables at the op of themerc that feels somewhat global.

Thanks! I've pushed `shape` and `padding.width`. 
 
> I think the hover-circle should have the diameter of `window.button.width`.

When `padding == 0` (it cannot, see next sentence) it would be indeed the same as `window.button.width`.
 
> But if we implemented `.bg.shape` for the `unpressed` state, we'd have three (for iconify, maximize and close) circles touching each other because there is no spacing. So we need `padding.width` between them.

Yes, I figured. I've added a minimum padding of 1 for the hover effect to avoid touching. Assuming that I correctly understood what you wrote.

> I'm making an assumption here that `window.button.width` is generally greater than the icon itself - but can't remember if we scale up. If we do, we should probably stop doing that!

I'm not familiar with the icon magic. Did you thought about this:  https://github.com/labwc/labwc/blob/master/src/theme.c#L135-L151 ?

src/theme.c
```diff
@@ -164,7 +167,13 @@ create_hover_fallback(struct theme *theme, const char *icon_name,
 	float overlay_color[4] = { 0.15f, 0.15f, 0.15f, 0.3f};
 	enum corner corner = corner_from_icon_name(icon_name);
 
-	if (corner == LAB_CORNER_UNKNOWN) {
+	if (theme->window_button_hover_bg_shape == LAB_CIRCLE) {
```
jp7677, line 170: I could turn this into a switch statement to make it more obvious that we are handling two different shapes here.

johanmalm:
I feel a picture might be helpful to make sure we're talking the same. I don't think we mean the same by `padding.width`. I'm pretty sure that's how it is in Openbox, but could do some testing.

variables

jp7677:
Ah, there might indeed be a misunderstanding ;) 

The definition I have in my mind for padding is the space to the inside of a certain element, like it is used in html/css. There `margin` describes the space to the outside and `padding` the space to the inside of a container. See e.g. https://stackoverflow.com/questions/5958699/difference-between-margin-and-padding

So for the hover effect I saw we have width/height for drawing the hover rectangle which covers the whole button (https://github.com/labwc/labwc/blob/master/src/theme.c#L169). For drawing the hover effect as a circle, I need to derive the diameter from somewhere. Introducing `padding` as you suggested and applying the padding from the definition I used, the diameter is the button width (or height, depending which one is smaller) minus twice the padding  (to apply the padding on both sides). So by increasing the padding, the diameter gets smaller. 

```
----------------------------------
| o |  Title   | o |  | o |  | o |
----------------------------------
```
With by best ascii art, padding like I implemented it, is the space between `|` and `o` .

(Having a `window.button.hover.bg.shape.circle.width` instead would an alternative which describes the diameter directly.)

Looking at your picture, I think `padding` there is the padding from the title border to the inside (which also describes the space between the buttons). I don't see (yet) how to get from there to the hover circle diameter. 

Note that this is really just about the hover effect. I'm not changing any button or icon positioning/size or any space anywhere else on the title bar. I have the feeling you have the latter in mind.

I hope above doesn't make things even more confusing ;)

jp7677:
Ok, re-reading everything. I think I understand now what you meant to say. When you wrote about `padding.width` I thought you meant it in the context of `window.button.hover.bg`. I didn’t realized that openbox has a global `padding.width`. It doesn’t help that I’m really not familiar with openbox ;)

> I feel a picture might be helpful to make sure we're talking the same. I don't think we mean the same by `padding.width`. I'm pretty sure that's how it is in Openbox, but could do some testing.

I’ve indeed implemented a different `padding.width`. Fortunately I also gave it a different name (`window.button.hover.bg.padding.width`). That said, I’m not sure if the global padding setting would solve my case. I do want the hover circle to be smaller than the button itself. I think adding space between the buttons would not change the button size itself, so without “my” padding, the hover circle would be to large for my taste. May be there is a use case for both padding’s.

johanmalm:
> May be there is a use case for both padding’s.

Yes. Quite possibly.

I think we should draw it all and conclude a way ahead. Of course it doesn't all need to be part of this PR, but I fear we might make a mistake if we decide on a particular case in isolation.

Just thinking aloud - a few things to establish:

- [ ] 1. As we have a `window.button.width`, should we also have a `window.button.height`?
- [ ] 2. The current hover-overlay height is the full height of the titlebar, so presumably `padding.height` has no effect there. Feel like hover-overlay height should only equal `titlebar.height` when `padding.height` is zero.
- [ ] 3. Need to establish which variables are + should be fixed vs variable (i.e. those that are derived by adding/subtracting the fixed ones).
- [ ] 4. Should we define `padding.height` and (a possibly future) `padding.width` as only being the padding of the titlebar (rather than spacing between icons and so on).
- [ ] 5. If the box defined by `window.button.width` and (a possibly future) `window.button.height` primarily exists to define the size of the bg-texture and a MAX size of any icon within it... then what's the use-case for padding of such as box (see diagram at https://github.com/labwc/labwc/pull/2127#issuecomment-2335842959)
- [ ] 6. The circular backgrounds are nice - and could be used for icon states other than `hover`. The current infrastructure isn't quite ready for that, but we ought to design any theme-options with that in mind. How does that affect things? Is it conceivable that different size will be needed for different states? Unusual use-cases could be met by creating specific svg/png icons, so we don't have to cater for all use-cases.

> That said, I’m not sure if the global padding setting would solve my case. I do want the hover circle to be smaller than the button itself.

Not smaller than the 'icon' surely? Just smaller than the full height of the titlebar - and equivalent width? So to achieve that couldn't we just make `window.button.width` smaller - assuming the circle diameter is based on that. And then introduce some new variable to enable spacing between the boxes defined by `window.button.width`? Example below.

image

johanmalm:
Just some thoughts. Enough for one day for me :smile:

variables2

jp7677:
Thanks. that is a really good picture and makes it easy to work with.

I've added some proof-of-concept code in the last commit that adds `padding.width` and `window,button.spacing`. It is completely broken code and the effects are only applied after a reconfigure. That said, it allows to play with the concept and I figured that I would achieve what I want with those options (with setting the hover circle diameter to the button width, so discarding "my" padding). 

Some rabbit holes though: 
- `theme->window_button_width` is used in a lot of places in the code (close to 30) base which needs careful adjustments since sometimes you need to add the padding into considerations, sometimes not. I don't think that I understand the logic everywhere to judge this correctly.
- The rectangle hover effect with rounded corners looks very out of place when padding is applied, since the hover is shifted together with the button
- Personally I would not make the hover circle smaller as the actual icon. I could imagine though that there are people that might want to do this with certain icons. But that is probably a very special case.

johanmalm:
Wow - you don't hang about!

Looks good. Not tested though.

Comments from quick review:

- Put `padding_width` next to `padding_height` in `theme.h`
- In `ssd-titlebar.c` where we have things like `int button_width = ssd->view->server->theme->window_button_width + ssd->view->server->theme->window_button_spacing;`, we should probably rename `button_width` to something like `step` or similar.

Looking at the code now, I think it's really tidy to handle the titlebar padding separately from icon-spacing (as you've done).

jp7677:
> Wow - you don't hang about!

The brain kept working during sleep ;)

> Looks good. Not tested though.

Well, the last commit was really just about getting a feeling how this would look UI wise and configuration wise. The code changes are far from complete.

tokyo4j:
Hmm.... I don't understand why we need `window.button.hover.bg.shape`/`padding.width`/`window.button.spacing` in the first place.

How about just supporting `hover-overlay.{png,svg,xbm}` (with alpha channel) along with other button files? For example, if `close-active.{png,svg,xbm}` and `hover-overlay.{png,svg,xbm}` is found but `close_hover-active.{png,svg,xbm}` is not found, we draw `hover-overlay.{png,svg,xbm}` over `close-active.{png,svg,xbm}` when cursor is on the close button.

In this way, users can easily extend the hover effect to various shapes (e.g. rounded rectangle) without needing to create `*_hover-*.{png,svg,xbm}` files for all buttons and we can use the existing infrastructure to configure the hover effect size (image size for png/xbm files and `title_height - 2 * padding_height` for svg files).

Sorry if I'm missing the point. I haven't read all of the comments in this PR.

jp7677:
@johanmalm I've still one todo left and the notes about the rectangle hover effect i.c.w. padding-width still applies, but otherwise I'm pretty happy how this is shaping up. I would be interested in your intermediate review. 
What sticks out now is the title bar corner width being independent from the button width. Those where, I guess, tied together for convenience and because the outer buttons had the same position, but that is no longer the case when padding-width is greater than zero. I think the separation also increases readability, I was pretty confused initially since I thought the corner related code was also about buttons ;) 

PS: Also needs commit clean up.

@tokyo4j May be you are right and such a scenario, the rounded hover effects from my screenshot, should only be supported with your suggestion. I've never set up custom icons because I'm pretty happy with the default styling, so I just tried what is possible with the theming without firing up a drawing application. Even if this PR makes it in, I think your suggestion should definitively (also) be implemented for allowing the scenarios you mentioned.

include/view.h
```diff
@@ -538,6 +538,7 @@ const char *view_get_string_prop(struct view *view, const char *prop);
 void view_update_title(struct view *view);
 void view_update_app_id(struct view *view);
 void view_reload_ssd(struct view *view);
+int view_min_width(void);
```
jp7677, line 541: May be `view_get_min_width()`?

tokyo4j:
> May be you are right and such a scenario, the rounded hover effects from my screenshot, should only be supported with your suggestion. I've never set up custom icons because I'm pretty happy with the default styling, so I just tried what is possible with the theming without firing up a drawing application. Even if this PR makes it in, I think your suggestion should definitively (also) be implemented for allowing the scenarios you mentioned.

Okay. Although I prefer the current hover effect, I understand the rounded hover effect may be worth supporting without requiring custom image files.

I tested this PR and it looks great. I really like the separation of the corner buffer width and button width.

As for my window icon support PR (#2128), I can update it to place window icon following `padding.{height,width}`. I initially thought the default `padding.height` of `3` is too large for window icon and I set the hardcoded padding of 1px, but I checked how Openbox works and I confirmed the geometry of window icon follows `padding.{height,width}`.

src/theme.c
```diff
@@ -682,6 +705,9 @@ entry(struct theme *theme, const char *key, const char *value)
 	if (match_glob(key, "border.width")) {
 		theme->border_width = atoi(value);
 	}
+	if (match_glob(key, "padding.width")) {
```
jp7677, line 708: Do we need bounds checking here and for `window.button.spacing`?

src/ssd/ssd-titlebar.c
```diff
@@ -252,18 +254,23 @@ static void
 update_visible_buttons(struct ssd *ssd)
```
jp7677, line 254: I'm not sure I got this correctly. What is a good app to test the resizing to a very small size with?

johanmalm: xterm

jp7677: thanks, should be the same behavior now.

docs/labwc-theme.5.scd
```diff
@@ -122,6 +128,14 @@ labwc-config(5).
 	Width of a titlebar button, in pixels.
 	Default is 26.
 
+*window.button.spacing*
+	Space between titlebar buttons, in pixels.
+	Default is 0.
+
+*window.button.hover.bg.shape*
+	The shape of the hover effect of a titlebar button: "rectangle" or "circle".
```
jp7677, line 136: May be “squared” and “rounded” instead?

src/theme.c
johanmalm: Yes, would make sense. We've not been super strict with that in other places, but it's good practice.

docs/labwc-theme.5.scd
johanmalm: I think circle and rectangle are good words

johanmalm:
> Hmm.... I don't understand why we need `window.button.hover.bg.shape`/`padding.width`/`window.button.spacing` in the first place.
> 
> How about just supporting `hover-overlay.{png,svg,xbm}` (with alpha channel) along with other button files? For example, if `close-active.{png,svg,xbm}` and `hover-overlay.{png,svg,xbm}` is found but `close_hover-active.{png,svg,xbm}` is not found, we draw `hover-overlay.{png,svg,xbm}` over `close-active.{png,svg,xbm}` when cursor is on the close button.
> 
> In this way, users can easily extend the hover effect to various shapes (e.g. rounded rectangle) without needing to create `*_hover-*.{png,svg,xbm}` files for all buttons and we can use the existing infrastructure to configure the hover effect size (image size for png/xbm files and `title_height - 2 * padding_height` for svg files).
> 
> Sorry if I'm missing the point. I haven't read all of the comments in this PR.

It's a good point and we should have had that discussion here. I personally run https://github.com/labwc/labwc-artwork/tree/main/button-themes/svg/Adwaita which works very well. Provided `_hover` variants of the icons exist, it works really well.

I have pondered this before, and think there is likely to be ongoing demand for both the `*_hover.{svg,png,xbm}` approach as well as a openbox style `.bg` texture backgrounds.

We already support both except that the `.bg` texture (i.e. the built-in hover-overlay) is not configurable.

Personally I think that supporting both is fine. The `.bg` texture approach is more convenient and provides compatibility with openbox themes, whereas the `_hover` icon approach provides great flexibility for unusual designs.

src/theme.c
jp7677: Lower bound is easy, `0`, but upper bound is tricky? I could just use some arbitrary number like 50, but not sure if this makes sense and things might still fall apart. So just checking for non-negative numbers here and actually also at other places in this function?

jp7677:
If it makes the discussion easier, I could offer to open the first commit (padding, spacing) as a separate PR.

src/theme.c
jp7677: I've added a commit in between which clamps negative values to zero for some of the configuration options.

src/theme.c
johanmalm: Yes. I think that is a good idea.

johanmalm:
LGTM.

I think your suggestion regarding a switch-statement for the `theme->window_button_hover_bg_shape == LAB_CIRCLE` stuff is good thought. Makes it more obvious what's going on.

johanmalm:
Tested. Works fine.

johanmalm:
Thanks.

PR #2128, opened 2024-09-07
ssd: support window icons

tokyo4j:
This PR adds support for window icons with the help of [libsfdo](https://gitlab.freedesktop.org/vyivel/libsfdo) for parsing desktop files and looking up icon files.

- The default `titleLayout` is updated to `NLIMC` which follows Openbox's default value.
- The window menu icon is initially shown and then its normal&hover buffer is replaced with the window icon when app_id is provided and the corresponding icon file is loaded.
- Applications need to provide `/usr/share/applications/[app_id].desktop` for the icons to be loaded. For example, icons are not loaded by default for thunderbird, which has app_id `thunderbird` but provides a dbus-style desktop file `/usr/share/applications/org.mozilla.Thunderbird.desktop`.
- Icon files in XPM format (e.g. xterm) are not supported.
- The window icon have a hardcoded margin of 1px. We can make this configurable later.
- Since the output scale is not considered for the icon lookup, rendered window icons can be blurry with scale > 1. But we can improve this later as the normal button icons can also be blurry in the current implementation.
- The logic in `button/button-{png,svg,xbm}.c` is generalized for window icons and the files are renamed to `img/img-{png,svg,xbm}.c`.

I'm not sure if we should make libsfdo an optional dependency. But I think it's fine to always require and statically link it because it's small and doesn't have any further dependencies (no `dependency()` in `meson.build`).

Lastly, I want to show my gratitude to @vyivel for writing such an amazing library that greatly simplifies the task of looking up icons. Without it, supporting window icons would have been much more challenging.

20240907_17h31m38s_grim


src/icon-loader.c
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	int lookup_options = SFDO_ICON_THEME_LOOKUP_OPTIONS_DEFAULT;
+#if !HAVE_RSVG
+	lookup_options |= SFDO_ICON_THEME_LOOKUP_OPTION_NO_SVG;
+#endif
+	struct sfdo_icon_file *icon_file = sfdo_icon_theme_lookup(
+		loader->icon_theme, icon_name, SFDO_NT, size, scale,
+		lookup_options);
+	if (!icon_file) {
```
vyivel, line 108: This should also check for `SFDO_ICON_FILE_INVALID`.

https://gitlab.freedesktop.org/vyivel/libsfdo/-/blob/a3e4935526f145326a6fce46006989090fc7d2df/include/sfdo-icon.h#L117

src/icon-loader.c
```diff
@@ -0,0 +1,134 @@
...
+	struct sfdo_icon_ctx *icon_ctx;
+	struct sfdo_desktop_db *desktop_db;
+	struct sfdo_icon_theme *icon_theme;
+};
+
+void
+icon_loader_init(struct server *server)
+{
+	struct icon_loader *loader = znew(*loader);
+	loader->basedir_ctx = sfdo_basedir_ctx_create();
```
vyivel, line 31: If you don't plan to use `basedir_ctx` for anything else (e.g. configuration loading), it can be destroyed immediately after creating desktop and icon contexts.

(I should document that...)

subprojects/libsfdo.wrap
```diff
@@ -0,0 +1,3 @@
+[wrap-git]
+url = https://gitlab.freedesktop.org/vyivel/libsfdo.git
+revision = v0.1.2
```
vyivel, line 3: You might want to fetch the latest (at the moment) commit instead or wait for 0.1.3, the latest release is kinda buggy (due to me not writing tests sooner).

01micko:
I think this is a great leap forward.

> I'm not sure if we should make libsfdo an optional dependency. But I think it's fine to always require and statically link it because it's small and doesn't have any further dependencies

Regarding this, your statement makes sense however debian as of now has no package for `libsfdo`  and it is the type of distro that likes to have everything dynamically linked so just for now it should be optional.

Will be testing tomorrow. Thank you for your great efforts @tokyo4j !

01micko:
Oh and thanks @vyivel for the library!

droc12345:
I have been looking into what is needed to add icons to menus.
So it would be nice if all logic about icons other than the actual loading was put into separate functions,
ie searching for icons, returning full icon path, etc.

You may have done that already, I haven't looked at the code yet, but thought I'd mention it is more useful than just window decoration.
Thanks.

johanmalm:
Fantastic.

Happy that we set NLMIC as default.

Nice refactoring moving `get_button_filename()` to `theme.c`.

I think it's a good choice to use `theme->window_button_width` for the icon size. Not sure if you've seen #2127, but @jp7677 is working on some stuff that may result in `padding.width` theme setting too, which will help create some space around the icon.

LGTM in principle, but think we should address @vyivel's comments.

Also, for consideration:

1. I suggest we enable icons by default, but add an option to disable to avoid ding-dong with distros that don't want to bundle.

2. In ssd-titlebar.c you can use `zfree()` instead of 

```
if (ssd->state.app_id) {
  free(ssd->state.app_id);
  ssd->state.app_id = NULL;
}
```

In fact, no need to NULL check before free, regardless. But zfree() sets to NULL after which is nice.

3. I can't work out how the theme name is set (for example 'Papirus'). Do we need to add a config-file option for that?

johanmalm:
Just catching up... looks like the second argument in `sfdo_icon_theme_load()` is the theme name.
If set to NULL we revert to `hicolor` which is a good fallback, but not what most people will probably want.

https://gitlab.freedesktop.org/vyivel/libsfdo/-/blob/main/sfdo-icon/load.c?ref_type=heads#L14

We can do it as a later PR if you want, but I think it'd be good to give the user the option to set that in `rc.xml`. 
Maybe just something like this:

```xml
<theme>
  <icon>Papirus</icon>
</theme>
```

src/theme.c
```diff
@@ -362,47 +389,47 @@ load_buttons(struct theme *theme)
 		zdrop(b->inactive.buffer);
 
 		/* PNG */
-		snprintf(filename, sizeof(filename), "%s-active.png", b->name);
-		button_png_load(filename, b->active.buffer);
-		snprintf(filename, sizeof(filename), "%s-inactive.png", b->name);
-		button_png_load(filename, b->inactive.buffer);
+		get_button_filename(b->name, "-active.png", filename, sizeof(filename));
+		img_png_load(filename, b->active.buffer);
+		get_button_filename(b->name, "-inactive.png", filename, sizeof(filename));
```
tokyo4j, line 394: Bug: `get_button_filename()` doesn't touch `filename` if the button file is not found, so old `filename` is passed to `img_{svg,png,xbm}_load()` again and again. I'll fix it.

src/icon-loader.c
tokyo4j: Thank you. Fixed.

src/icon-loader.c
tokyo4j: Fixed.

subprojects/libsfdo.wrap
tokyo4j: Okay. Updated to `head`.

tokyo4j:
> So it would be nice if all logic about icons other than the actual loading was put into separate functions,
ie searching for icons, returning full icon path, etc.

I separated the logic of looking up desktop files & icon files and loading it into a buffer into `icon-loader.c`.

> 1. I suggest we enable icons by default, but add an option to disable to avoid ding-dong with distros that don't want to bundle.

I made libsfdo an optional dependency. libsfdo is fetched by default and can be disabled with `-Dicon=disabled`. It should be noted that it breaks existing build scripts with `--wrap-mode nodownload` like Arch's `arch-meson`.

> 2. In ssd-titlebar.c you can use zfree() instead

Fixed.

> 3. We can do it as a later PR if you want, but I think it'd be good to give the user the option to set that in rc.xml.
> Maybe just something like this:
> ```xml
> <theme>
>   <icon>Papirus</icon>
> </theme>
> ```

It makes sense. I added `<theme><icon>`.

I'm marking this PR as a Draft for now to wait for another libsfdo release. Also, `padding.width` sounds nice and maybe we can replace `APP_ICON_MARGIN` with it in this PR.

johanmalm:
Thanks!
LGTM.
Happy to go for merge before next `libsfdo` tag.
Do we depend on '>0.1.2' for the next release of `labwc`?

johanmalm:
Looks like both Void and FreeBSD CI are having problems with `libsfdo`

```
In function 'exec_add_quoted',
    inlined from 'load_exec' at ../subprojects/libsfdo/sfdo-desktop/db.c:783:8:
../subprojects/libsfdo/sfdo-desktop/db.c:537:41: error: 'escape_i' may be used uninitialized [-Werror=maybe-uninitialized]
  537 |                                         logger_write(logger, SFDO_LOG_LEVEL_ERROR,
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  538 |                                                         "%d:%d: unexpected field code in a quoted argument at position %zu",
      |                                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  539 |                                                         scanner->line, scanner->column, escape_i);
      |                                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../subprojects/libsfdo/sfdo-desktop/db.c: In function 'load_exec':
../subprojects/libsfdo/sfdo-desktop/db.c:518:16: note: 'escape_i' was declared here
  518 |         size_t escape_i;
      |                ^~~~~~~~
cc1: all warnings being treated as errors
ninja: build stopped: subcommand failed.
```

tokyo4j:
> Looks like both Void and FreeBSD CI are having problems with libsfdo

I tried the release build with GCC 13.3.0 (installed via `gcc13` package) on my Arch system and I got the error on `escape_i` which also appeared in Void's CI. But I couldn't reproduce the error on `best_dist`. So it looks quite dependent on GCC versions and distros.

> Happy to go for merge before next libsfdo tag.
Do we depend on '>0.1.2' for the next release of labwc?

@vyivel said in the review:

> You might want to fetch the latest (at the moment) commit instead or wait for 0.1.3, the latest release is kinda buggy (due to me not writing tests sooner).

So merging when libsfdo 0.1.3 released feels desirable from the stability perspective. I don't think we should rush here.

knm100:
The three programs "pcmanfm-qt, qterminal, qpdfview" did not display the icons with  `<icon></icon>` in rc.xml.  they all are have the app-id.desktop at /usr/share/applications/.   set up the `<icon>Breeze</icon>` in rc.xml,   All programs did not display the icons.

vyivel:
@knm100 The icon theme name is "breeze", all lowercase.

stefonarch:
Nice!
I notice that `window.button.width: ` does affect its size while for the other buttons (.xbm) it affects only the distance between them:

immagine
immagine

knm100:
> @knm100 The icon theme name is "breeze", all lowercase.

Thanks, It worked,  and pcmanfm-qt display the icons now, but the qterminal and qpdfviewer still don't display the icon.

tokyo4j:
>  I notice that window.button.width: does affect its size while for the other buttons (.xbm) it affects only the distance between them:

That's because we shrink the button image to fit within (`window.button.width`x`titlebar.height`), but never increase the size of it. So as long as your XBM is smaller than `window.button.width`x`titlebar.height`, the button image size is constant even if you change `window.button.width`.

> Is there some way for a padding(-left)?

I rebased the branch to the latest master which has `padding.width`, so now you can control the left-padding with it.

stefonarch:
> I rebased the branch to the latest master which has `padding.width`, so now you can control the left-padding with it.

Perfect! It's not documented in manpage yet afaik.

jp7677:
> > I rebased the branch to the latest master which has `padding.width`, so now you can control the left-padding with it.
> 
> Perfect! It's not documented in manpage yet afaik.

Did you look for https://labwc.github.io/labwc-theme.5.html#entry_padding.width ?

stefonarch:
I looked at `man labwc-theme` locally:

immagine

jp7677:
> I looked at `man labwc-theme` locally:

The web site and the man page are from the same source, it looks like you man pages are not built from latest git version.

stefonarch:
Think so too, it's a checkout of https://github.com/tokyo4j/labwc/tree/app-icon

knm100:
qpwgraph and qalculate provide the dbus-style desktop file as well, but they can display the icons correctly.  
thunderbird can't do it.

tokyo4j:
> qpwgraph and qalculate provide the dbus-style desktop file as well, but they can display the icons correctly.
thunderbird can't do it.

Ah I forgot to mention that if the desktop file is not found, we use the bare app_id as the key for icon lookup (I just copied the behavior from croissant compositor). So the app_id `org.rncbc.qpwgraph` is used to directly find the icon file `/usr/share/icons/hicolor/scalable/apps/org.rncbc.qpwgraph.svg`.

knm100:
> > qpwgraph and qalculate provide the dbus-style desktop file as well, but they can display the icons correctly.
> > thunderbird can't do it.
> 
> Ah I forgot to mention that if the desktop file is not found, we use the bare app_id as the key for icon lookup (I just copied the behavior from croissant compositor). So the app_id `org.rncbc.qpwgraph` is used to directly find the icon file `/usr/share/icons/hicolor/scalable/apps/org.rncbc.qpwgraph.svg`.

So if  I make a copy of  org.mozilla.Thunderbird.svg to /usr/share/icons/hicolor/scalable/apps/ , thunderbird shoulde can display icon.  but I just tried.  the icon of thunderbird didn't display.     and the qpwgraph's app_id is qpwgraph, isn't `org.rncbc.qpwgraph`

stefonarch:
FYI I've no issues anywhere with Papirus icons.
immagine

tokyo4j:
> So if I make a copy of org.mozilla.Thunderbird.svg to /usr/share/icons/hicolor/scalable/apps/ , after gtk-update-icon-cache  /usr/share/icons/hicolor, thunderbird shoulde can display icon. but I just tried. the icon of thunderbird didn't display.
> 
> Edited:
Indeed, if chang the Thunderbird to thunderbird, It will work.

Yeah the icon lookup is case-sensitive, so you need to create the image with the exact filename `[app_id].svg` (for thunderbird `{/usr,~/.local}/share/icons/hicolor/scalable/apps/thunderbird.svg`).

But I'd rather recommend running `cp /usr/share/applications/org.mozilla.Thunderbird.desktop ~/.local/share/applications/thunderbird.desktop`. In this way, the compositor can access all the icon files for the app and pick the best one for the titlebar size even when the app only provides PNG icons. This will also work for other compositors like KWin. As a side note, adding `NoDisplay=true` to the copied desktop file prevents duplicated entries in application launchers.

> and the qpwgraph's app_id is qpwgraph, isn't org.rncbc.qpwgraph

How did you see the app_id? In my setup, `wlrctl toplevel list` reports that the app_id is `org.rncbc.qpwgraph`.

knm100:
> How did you see the app_id? In my setup, `wlrctl toplevel list` reports that the app_id is `org.rncbc.qpwgraph`.

I use command ps -A | grep qp.   I should make an mistaken about app_id.

knm100:
> But I'd rather recommend running `cp /usr/share/applications/org.mozilla.Thunderbird.desktop ~/.local/share/applications/thunderbird.desktop`. In this way, the compositor can access all the icon files for the app and pick the best one for the titlebar size even when the app only provides PNG icons. This will also work for other compositors like KWin. As a side note, adding `NoDisplay=true` to the copied desktop file prevents duplicated entries in application launchers.

Thank you very much! It is perfect for me

knm100:
> But I'd rather recommend running `cp /usr/share/applications/org.mozilla.Thunderbird.desktop ~/.local/share/applications/thunderbird.desktop`. In this way, the compositor can access all the icon files for the app and pick the best one for the titlebar size even when the app only provides PNG icons. This will also work for other compositors like KWin. As a side note, adding `NoDisplay=true` to the copied desktop file prevents duplicated entries in application launchers.

Thank you very much! It is perfect for me .

jp7677:
@tokyo4j i think I have placed (yet again) another stone in your road with this PR by adding https://github.com/labwc/labwc/pull/2146 . With your PR it is possible to have multiple buttons that open the client menu, but I didn’t think of this when writing #2146. Not sure yet how to deal with this elegantly, expect from being quite drastic and making the icon and window menu button mutually exclusive. Something to think about..

tokyo4j:
I'll merge this tomorrow.

johanmalm:
Good plan. LGTM.

tokyo4j:
It's time to merge. We'll need to request package maintainers to add libsfdo as a dependency or to add `-Dicon=disable` in their build script for the next release.

tokyo4j:
I found starting an application in fullscreen causes SEGFAULT because `view->ssd_enabled` doesn't mean `view->ssd != NULL`. I'll push a fix soon.

johanmalm:
Thanks @tokyo4j Nicely implemented :smile:

PR #2136, opened 2024-09-11
theme: compensate the hover corner radius with padding

jp7677:
When padding.width is set, the corner radius of the rectangle hover effect can be smaller since there is more space available.

This avoid weird looking rounded corners of the hover effect when there is enough space due to padding, that said, this is still a band aid.


johanmalm:
LGTM. Nice "fixup"

PR #2144, opened 2024-09-14
docs: add a note on padding.height

tokyo4j:
Currently, unlike `padding.width`, `padding.height` is not applied for the button images in PNG/XBM format, or builtin hover effects. I personally don't like this inconsistency, but I have no idea how to fix this without breaking compatibility. At least we should document this behavior to avoid confusion, which I'm doing in this PR.

CC @johanmalm @jp7677

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -57,8 +57,9 @@ labwc-config(5).
 
 *padding.height*
 	Vertical padding size, in pixels, used for spacing out elements
-	in the window decorations.
-	Default is 3.
+	in the window decorations. Unlike *padding.width*, this padding is
+	not applied for window buttons in PNG/XBM format, or the builtin
+	hover effects. Default is 3.
 
 *titlebar.height*
 	Window title bar height.


jp7677:
My 2cents: if we do this correctly for svg buttons, I would consider it a "fix" to also honor `padding.height` for png/xbm buttons.  I realize that this might break capability, but may be that is acceptable in this case? May be postpone such a change until labwc 0.9?. The circle hover effect is easily adjustable (`radius = MIN(height, width)` when height and width are set) and it would makes sense to add this restriction. I don't know what to do with the rectangle hover effect though, I could imagine that it would look weird when the height is truncated, also considering the rounded corner effect.

johanmalm:
I think we should:
- Implement `window.button.height`
- Make the buttons (i.e. the rectangle represented by `window.button.width` x `window.button.height` respect `padding.height`.
- Change the default shape to `circle` (which already nicely stays within the `window.button.{width,height}` rectangle. We ought to check with point with @Consolatis to make sure there are no strong feelings about this.
- Change the shape=rectangle version (current default) to respect `window.button.{width,height}` and possibly issue warning in the documentation (and possibly in `rcxml.c` too - if `padding.height != 0`.

I'm not sure if you described a difference between xbm/png/svg in OP or if you just inadvertently left svg out in the sentence. If you described a differences in behaviour, we ought to fix it. It was probably an oversight when I first implemented.

I normally try to guard carefully against regressions/changes, but we have to accept that life isn't perfect and with things like this, I think it's much better to get the code correct and consistent rather than building on an originally ill though through implementation.

jp7677:
> Change the default shape to circle (which already nicely stays within the window.button.{width,height} rectangle.

Here is an even more radical thought: drop the rectangle hover altogether. We could clean up some code, remove the newly introduced `shape` theme option again and this would also definitively solve the awkward rounded corner hover effect.

johanmalm:
> Here is an even more radical thought: drop the rectangle hover altogether. We could clean up some code, remove the newly introduced `shape` theme option again and this would also definitively solve the awkward rounded corner hover effect.

Oh that is radical! 😄 

Well, I think there is a community of people for whom consistency with openbox themes is important and keeping the rectangular shape is essential for that.

Personally I don't mind getting rid of the hover with rounded corner (which brings some code complexity)... but I think some people use it and it took quite a while to get it to work consistently. Getting Consolatis's view would be a good first step with that.

tokyo4j:
> - Implement window.button.height
> - Make the buttons (i.e. the rectangle represented by window.button.width x window.button.height respect padding.height.

I suggest we just set `window.button.height` to `titlebar.height - 2*padding.height` (which is the same as the title font extent) and allow arbitrary button heights regardless of `padding.height`, rather than always respecting `padding.height`. This allows users to set the padding of the title text and that of the buttons separately (ref: [topic from arch forum](https://bbs.archlinux.org/viewtopic.php?id=215551)).

Another random idea for my window icon PR is supporting `window.button.{max,iconify,...}.height` to set padding around window icon differently from other buttons.

> - Change the default shape to circle (which already nicely stays within the window.button.{width,height} rectangle. We ought to check with point with @Consolatis to make sure there are no strong feelings about this.
> - Change the shape=rectangle version (current default) to respect window.button.{width,height} and possibly issue warning in the documentation (and possibly in rcxml.c too - if padding.height != 0.

Although I prefer the current hover effect, I agree that changing the default behavior will make the logic simpler and intuitive. We'll need to explain to users the way to bring back the old behavior though.

> I'm not sure if you described a difference between xbm/png/svg in OP or if you just inadvertently left svg out in the sentence. If you described a differences in behavior, we ought to fix it. It was probably an oversight when I first implemented.

Currently, we always draw the svg image on the buffer with fixed size of `titlebar.height - 2*padding.height` while we use the size of the image itself for other formats (PNG and XBM).

https://github.com/labwc/labwc/blob/21e92632a656a2b3e9584b4d2e64bdf912fed522/src/theme.c#L386-L390

So I think we can fix this by using `<svg width="" height="">` attributes in the SVG file as the buffer size if they exist and they're not so big. I agree that this behavior is unwanted and we should fix it.

I'll close this PR in favor of the suggested ideas.

johanmalm:
> I suggest we just set `window.button.height` to `titlebar.height - 2*padding.height` (which is the same as the title font extent)

That feel too implicit to me... like set your icons size by picking a font and font-size. 

> and allow arbitrary button heights regardless of `padding.height`, rather than always respecting `padding.height`. This allows users to set the padding of the title text and that of the buttons separately.

My thoughts:

1. It feels too inconsistent to have padding.height/width apply to all objects within the titlebar except button-height.
2. I'm nervous about deriving `window.button.height` from other variables, partly because the icon aspect ratio is likely to important but also because icon "crispness" would be affected if scaled down. For example, imagine if a user creates a 24x24 icon, and we then calculate window.button.height to be 21.

I prefer to make the objects within the titlebar the primary objects. What I mean by that is that buttons and the title self have explicit dimensions (i.e. not derived from others) whereas the titlebar itself is a secondary object that adjusts its size depending on those primary ones.

So my proposed way forward (using 'height' to describe things)
- The button-allocated-space would be determined by `window.button.height` so that the user can be sure that icons are not scaled down and that aspect ratios are maintained.
- The `titlebar.height` would be 'derived` by taking the max of button-height & font-height, and then adding the padding (i.e. just add button-height to current calculation).
- If a user wants icons or icon-hover effects that go all the way to the titlebar edge, then set padding.height to 0.
- If the same user wants the titlebar to be taller than the font-height, just increase `window.button.height`.

> Another random idea for my window icon PR is supporting `window.button.{max,iconify,...}.height` to set padding around window icon differently from other buttons.

Yes, I think this is a good idea. The window-icon is likely to have different requirements in this regard (height/padding/hover-effect/etc). We already support setting colors for individual buttons, so feels consistent.

> Currently, we always draw the svg image on the buffer with fixed size of `titlebar.height - 2*padding.height` while we use the size of the image itself for other formats (PNG and XBM).

Thanks for spotting that. We need to get that sorted. At the back of my mind there is the case where svgs don't specify size (but can't remember if that's legal). To be investigated!

johanmalm:
@tokyo4j - for info... I've started working towards a PR for this.
https://github.com/johanmalm/labwc/tree/window-button-height

PR #2145, opened 2024-09-15
theme: round the hover effect along the window border

tokyo4j:
This enhances the mitigation in #2136 by using the buffer returned from `rounded_corner()` as a mask for the hover overlay buffer.

Please note that this is my first time writing code with cairo.

20240916_00h14m09s_grim


johanmalm:
Great idea to mask. Have tested. Works well.

PR #2146, opened 2024-09-15
action: open the client menu underneath the window menu button

jp7677:
It looks slightly awkward when the client menu shows up in the left corner of the view and the window menu button is configured to be on the right side.

Before:
Screenshot

After:
Screenshot

The appearance is unchanged when the menu button is the first one and when no padding is applied. 

My initial version placed the menu in the center underneath the button (like GTK4 does), but I think this version fits better considering how the menu looks.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -617,6 +617,38 @@ action_list_free(struct wl_list *action_list)
 	}
 }
 
+static int
+get_window_menu_button_offset(struct server *server, struct view *view)
+{
+	int padding_width = server->theme->padding_width;
+	int button_width = server->theme->window_button_width;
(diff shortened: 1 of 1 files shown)


src/action.c
```diff
@@ -617,6 +617,38 @@ action_list_free(struct wl_list *action_list)
 	}
 }
 
+static int
+get_window_menu_button_offset(struct server *server, struct view *view)
```
jp7677, line 621: Not sure if this is the right place. Alternatively it could also be part of `view` with the button type as argument. Something like `view_get_ssd_button_offset()`.

johanmalm:
Beautiful. Thanks.

PR #2149, opened 2024-09-16
action: try to find the button that triggered the window menu

jp7677:
This should make https://github.com/labwc/labwc/pull/2146 more robust against not so common configurations. It should also work with a future window icon (#2128 ). If both icon and menu button are present, it should fall back to the left corner.

cc @tokyo4j

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -6,6 +6,7 @@
 #include <strings.h>
 #include <unistd.h>
 #include <wlr/util/log.h>
+#include <linux/input-event-codes.h>
 #include "action.h"
 #include "common/macros.h"
 #include "common/list.h"
@@ -24,6 +25,7 @@
 #include "view.h"
 #include "workspaces.h"
 #include "input/keyboard.h"
+#include "config/mousebind.h"
 
 enum action_arg_type {
 	LAB_ACTION_ARG_STR = 0,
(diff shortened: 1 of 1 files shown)


tokyo4j:
I was working on this in another approach ([link](https://github.com/labwc/labwc/compare/master...tokyo4j:labwc:refactor-action-ctx)) in which `cursor_context` is passed to `actions_run()`. In this way, we can reuse `cursor_context.node` to reference the button node and call `wlr_scene_node_coord()` to get the position of it.

Sorry for my miscommunication. I saw [your comment](https://github.com/labwc/labwc/pull/2128#issuecomment-2352077857) but didn't fully understand what you meant. I should've asked you how you were going to solve this.

jp7677:
> I was working on this in another approach ([link](https://github.com/labwc/labwc/compare/master...tokyo4j:labwc:refactor-action-ctx)) in which `cursor_context` is passed to `actions_run()`. In this way, we can reuse `cursor_context.node` to reference the button node and call `wlr_scene_node_coord()` to get the position of it. I prefer my approach as it's more straightforward, though I need to refine it more as it touches more parts of the codebase.
> 
> Sorry for my miscommunication. I saw [your comment](https://github.com/labwc/labwc/pull/2128#issuecomment-2352077857) but didn't fully understand what you meant. I should've asked you how you were going to solve this.

No worries, I'm completely happy with closing this PR and going with your solution, which indeed does look cleaner and much easier to read.

PR #2150, opened 2024-09-16
theme: finish titleLayout implementation

johanmalm:
Change Openbox style `WLIMC` syntax to `menu:title:minimize,maximize,close` as was agreed when PR #2088 was merged.

Ref:
- https://github.com/labwc/labwc/pull/2088#issuecomment-2295730704

Cc: @xi

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -437,19 +437,24 @@ extending outward from the snapped edge.
 *<theme><name>*
 	The name of the Openbox theme to use. It is not set by default.
 
-*<theme><titleLayout>*
+*<theme><titlebar><layout>*
 	Selection and order of buttons in a window's titlebar.
-	The following letters can be used, each only once:
-	- L: window label (aka. title)
-	- |: empty space (can be used instead of a title)
-	- W: window menu
-	- I: iconify
-	- M: maximize toggle
-	- C: close
-	- S: shade toggle
-	- D: all-desktops toggle
+	The following identifiers can be used, each only once:
+	- 'menu': window menu
+	- 'iconify': iconify
+	- 'max': maximize toggle
+	- 'close': close
+	- 'shade': shade toggle
+	- 'desk': all-desktops toggle
 
-	Example: WLIMC
+	A colon deliminator is used to separate buttons on the left and right,
+	whereas commas are used to separate items within a section. It is
+	mandatory to use one colon.
+
+	Default: menu:iconify,max,close
+
+*<theme><titlebar><showTitle>* [yes|no]
+	Show the window title in the titlebar. Default is yes.
 
 *<theme><cornerRadius>*
 	The radius of server side decoration top corners. Default is 8.
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -119,61 +119,139 @@ parse_window_type(const char *type)
 	}
 }
 
+/*
+ * Openbox/labwc comparison
+ *
+ * Instead of openbox's <titleLayout>WLIMC</title> we use
+ *
+ *     <titlebar>
+ *       <layout>menu:iconfiy,max,close</layout>
+ *       <showTitle>yes|no</showTitle>
+ *     </titlebar>
+ *
+ * ...using the icon names (like iconify.xbm) without the file extension for the
+ * identifier.
+ *
+ * labwc        openbox     description
+ * -----        -------     -----------
+ * menu         W           Open window menu (client-menu)
+ * iconfiy      I           Iconify (aka minimize)
+ * max          M           Maximize toggle
+ * close        C           Close
+ * shade        S           Shade toggle
+ * desk         D           All-desktops toggle (aka omnipresent)
+ */
 static void
-fill_title_layout(char *nodename, char *content)
+fill_section(const char *content, struct wl_list *list)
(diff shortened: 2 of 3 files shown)


johanmalm:
@tokyo4j @ahesford @jlindgren90 - would one of you fancy a code review of this PR?

docs/labwc-config.5.scd
```diff
@@ -439,17 +439,30 @@ extending outward from the snapped edge.
...
+	section, whereas buttons are only allowed in the left/right sections.
+	However, the title can be right/center/left aligned by setting
+	'window.title.text.justify' in themerc.
+	The underlying behavior is consistent with Openbox and the `titleLayout`
+	syntax deviation is simply a superficial one in the interest of being
+	more verbose and understandable, for example:
+	'menu:title:maximize,minimize,close' is the equivalent of 'WLIMC'.
+	The two colons are manadatory.
+
+	Example: menu:title:shade,alldesktops,minimize,maximize,close
```
tokyo4j, line 465: I suggest showing the default value rather than an example.

```suggestion
	Default is 'menu:title:minimize,maximize,close'.
```

ahesford:
I didn't join the original conversation, but I don't really like this syntax. Only honoring "title" in the center position is both confusing and redundant; if the title is only valid in the center, then the title *is* the separator, and the "title" keyword should just be eliminated. In other words,

    left1,left2,left3:right1,right2

is sufficient to describe the entire state.

If we want to allow a blank space (as we do now, but why?), we could, for example, use ":" for the title separator and something else (like "_") to mean "separate with blank space instead". Either of these is a bit easier and cleaner to validate, because we can just check that there is only a single separator in the string.

 Even better might be a more descriptive configuration, for example

    <leftButtons>...</leftButtons>
    <rightButtons>...</rightButtons>
    <blankTitle>yes|no</blankTitle>

johanmalm:
Thanks both for taking a look. 

Regarding the syntax, @Consolatis and I have had a couple of lengthy discussions on IRC... and got to a compromise as PR is written. I take your points though. 

Let's tease the syntax a bit further.

Could do something like:

```xml
<theme>
  <titlebar>
    <layout>left1,left2:right1,right2</layout>
    <showTitle>yes|no</showTitle>
  </titlebar>
</theme>
```

We're deviating from openbox anyway, so don't have to stick with `titleLayout`.

ahesford:
I think your alternative is clean and readable.

jlindgren90:
I can't do a detailed review right now, but I trust the other reviewers who have already looked at it :)

No opinion on syntax since I always use the default layout.

tokyo4j:
(Sorry for late reply. Now I returned from 2-week intern and hopefully I'll have more free time from today)

Although I don't have a strong preference in here, I slightly prefer the alternative configuration (with `<titlebar><layout>` and `<titlebar><showTitle>`). Also from the developer's perspective, I guess the alternative configuration is easier to work with.

Another thing I'm wondering is the button names. For example, we have `Iconify` action and `iconify.{png,svg,xbm}`, but use the term `minimize` only in the documentation. I personally prefer to align the names with the icon file names (`max`,`iconify`,`close`,`menu`,`desk`,`shade`)

johanmalm:
> with the icon file names (max,iconify,close,menu,desk,shade)

Good point. That would indeed be more consistent, and just as readable.

I'll do a fixup to that effect.

```xml
<theme>
  <titlebar>
    <layout>menu:iconify,max,close</layout>
    <showTitle>yes|no</showTitle>
  </titlebar>
</theme>
```

johanmalm:
Okay, ready for final review.
If anyone could give me an `LGTM` I'll merge :smile:

src/config/rcxml.c
```diff
@@ -119,61 +119,145 @@ parse_window_type(const char *type)
...
+
+	/*
+	 * The string typically looks like: 'menu:iconfiy,max,close' so we
+	 * replace colons with commas and then split on commas to get a simple
+	 * string list of all the identifiers.
+	 */
+	while ((p = strchr(s, ':'))) {
+		*p = ',';
+	}
+	gchar **idents = g_strsplit(s, ",", -1);
```
tokyo4j, line 203: I think `g_strsplit_set()` can be used here to split a string with multiple delimiters.

tokyo4j:
LGTM other than my small suggestion. Thank you for working on this!

src/config/rcxml.c
johanmalm: Nice. Didn't know about that one. Done.

jp7677:
After adjusting my settings to the new syntax, I saw that `<titlebar layout=":menu,iconify,max,close" />` hides the title, I need to explicitly add it with `<titlebar layout=":menu,iconify,max,close" showTitle="yes" />`. Omitting the `titlebar` element completely also hides the title, so looks like the implemented default acts as `no` instead of `yes`.

johanmalm:
> After adjusting my settings to the new syntax, I saw that `<titlebar layout=":menu,iconify,max,close" />` hides the title, I need to explicitly add it with `<titlebar layout=":menu,iconify,max,close" showTitle="yes" />`. Omitting the `titlebar` element completely also hides the title, so looks like the implemented default acts as `no` instead of `yes`.

Thanks for catching. Fixed with #2179

PR #2153, opened 2024-09-18
client list combined menu fix for submenu use

droc12345:
Without this fix it doesn't update properly as a submenu.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -665,10 +665,15 @@ show_menu(struct server *server, struct view *view,
 		return;
 	}
 
-	/* Need to refresh to show current windows and recalculate width */
-	if (!strcasecmp(menu_name, "client-list-combined-menu")) {
-		update_client_list_combined_menu(menu->server);
-	}
+	/*
+	 *  We always refresh the client-list-combined-menu so that it is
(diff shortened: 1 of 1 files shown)


src/action.c
```diff
@@ -666,9 +666,7 @@ show_menu(struct server *server, struct view *view,
 	}
 
 	/* Need to refresh to show current windows and recalculate width */
-	if (!strcasecmp(menu_name, "client-list-combined-menu")) {
-		update_client_list_combined_menu(menu->server);
-	}
+	update_client_list_combined_menu(menu->server);
```
johanmalm, line 669: Suggest adding a comment here to make the reason for always refreshing this clear to other users. Otherwise they're likely to do what I did in the fixup.

Maybe something like:

```C
/*
 * We always refresh the client-list-combined-menu so that it is up-to-date
 * whether it is used as a menu with `menu_name` set to `client-list-combined-menu`
 * or used as a submenu which we don't know at this point.
 */
```

droc12345: Done, just letting ci checks run.

johanmalm:
Thanks.

PR #1671, opened 2024-03-31
[wip] Add ToggleSnapToEdge

Consolatis:
This allows to replace the usual maximize behavior that does not respect gaps with SnapToEdge direction="center" which does respect gaps. It can be bound to the SSD Maximize button and Title double click handlers in rc.xml.

Missing:
- [ ] docs
- [ ] some enum for `<snapping topMaximize="yes|no">` like `yes|no|respectGaps`
- [ ] an entry in the website FAQ like "How to make maximize respect gaps"

Ref: https://www.reddit.com/r/labwc/comments/1brvsj8/

---

Not sure when I'll be working on this again, if somebody else wants to take this over feel free.

To test:
```xml
<labwc_config>

    <keyboard>
        <default />
        <keybind key="W-a">
            <action name="ToggleSnapToEdge" direction="center" />
        </keybind>
    </keyboard>

    <mouse>
        <default />
        <context name="Maximize">
            <mousebind button="Left" action="Click">
                <action name="ToggleSnapToEdge" direction="center" />
            </mousebind>
        </context>

        <context name="Title">
            <mousebind button="Left" action="DoubleClick">
                <action name="ToggleSnapToEdge" direction="center" />
            </mousebind>
        </context>
    </mouse>

</labwc_config>
```

PR #2154, opened 2024-09-18
action: add ToggleSnapToEdge/Region and UnSnap

jp7677:
Behaves identical like SnapToEdge and SnapToRegion, but untiles the window when already being tiled to the given region or direction.

I somehow missed this since there is already `ToggleMaximize` ;). `ToggleSnapToEdge` is also identical with the behavior I'm seeing on Gnome Mutter w.r.t. tiling.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -1172,8 +1172,9 @@ actions_run(struct view *activator, struct server *server,
 			}
 			break;
 		case ACTION_TYPE_UNSNAP:
-			if (view && view->maximized == VIEW_AXIS_NONE && !view->fullscreen
-					&& view_is_tiled(view)) {
+			if (view && !view->fullscreen && !view_is_floating(view)) {
+				view_maximize(view, VIEW_AXIS_NONE,
+					/* store_natural_geometry */ false);
 				view_set_untiled(view);
 				view_apply_natural_geometry(view);
 			}
--- a/src/action.c
+++ b/src/action.c
@@ -69,6 +69,7 @@ enum action_type {
 	ACTION_TYPE_EXECUTE,
 	ACTION_TYPE_EXIT,
 	ACTION_TYPE_MOVE_TO_EDGE,
+	ACTION_TYPE_TOGGLE_SNAP_TO_EDGE,
 	ACTION_TYPE_SNAP_TO_EDGE,
 	ACTION_TYPE_GROW_TO_EDGE,
 	ACTION_TYPE_SHRINK_TO_EDGE,
@@ -99,6 +100,7 @@ enum action_type {
 	ACTION_TYPE_MOVE_RELATIVE,
 	ACTION_TYPE_SEND_TO_DESKTOP,
 	ACTION_TYPE_GO_TO_DESKTOP,
+	ACTION_TYPE_TOGGLE_SNAP_TO_REGION,
 	ACTION_TYPE_SNAP_TO_REGION,
 	ACTION_TYPE_TOGGLE_KEYBINDS,
 	ACTION_TYPE_FOCUS_OUTPUT,
@@ -130,6 +132,7 @@ const char *action_names[] = {
 	"Execute",
 	"Exit",
 	"MoveToEdge",
+	"ToggleSnapToEdge",
 	"SnapToEdge",
 	"GrowToEdge",
 	"ShrinkToEdge",
@@ -160,6 +163,7 @@ const char *action_names[] = {
 	"MoveRelative",
 	"SendToDesktop",
 	"GoToDesktop",
+	"ToggleSnapToRegion",
 	"SnapToRegion",
 	"ToggleKeybinds",
 	"FocusOutput",
@@ -319,12 +323,15 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		/* Falls through */
+	case ACTION_TYPE_TOGGLE_SNAP_TO_EDGE:
 	case ACTION_TYPE_SNAP_TO_EDGE:
 	case ACTION_TYPE_GROW_TO_EDGE:
 	case ACTION_TYPE_SHRINK_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
 			enum view_edge edge = view_edge_parse(content);
-			if ((edge == VIEW_EDGE_CENTER && action->type != ACTION_TYPE_SNAP_TO_EDGE)
+			bool allow_center = action->type == ACTION_TYPE_TOGGLE_SNAP_TO_EDGE
+				|| action->type == ACTION_TYPE_SNAP_TO_EDGE;
+			if ((edge == VIEW_EDGE_CENTER && !allow_center)
 					|| edge == VIEW_EDGE_INVALID) {
 				wlr_log(WLR_ERROR, "Invalid argument for action %s: '%s' (%s)",
 					action_names[action->type], argument, content);
@@ -413,6 +420,7 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		break;
+	case ACTION_TYPE_TOGGLE_SNAP_TO_REGION:
 	case ACTION_TYPE_SNAP_TO_REGION:
 		if (!strcmp(argument, "region")) {
 			action_arg_add_str(action, argument, content);
@@ -538,6 +546,7 @@ action_is_valid(struct action *action)
 		arg_name = "command";
 		break;
 	case ACTION_TYPE_MOVE_TO_EDGE:
+	case ACTION_TYPE_TOGGLE_SNAP_TO_EDGE:
 	case ACTION_TYPE_SNAP_TO_EDGE:
 	case ACTION_TYPE_GROW_TO_EDGE:
 	case ACTION_TYPE_SHRINK_TO_EDGE:
@@ -551,6 +560,7 @@ action_is_valid(struct action *action)
 	case ACTION_TYPE_SEND_TO_DESKTOP:
 		arg_name = "to";
 		break;
+	case ACTION_TYPE_TOGGLE_SNAP_TO_REGION:
 	case ACTION_TYPE_SNAP_TO_REGION:
 		arg_name = "region";
 		break;
@@ -889,10 +899,20 @@ actions_run(struct view *activator, struct server *server,
 				view_move_to_edge(view, edge, snap_to_windows);
 			}
 			break;
+		case ACTION_TYPE_TOGGLE_SNAP_TO_EDGE:
 		case ACTION_TYPE_SNAP_TO_EDGE:
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
+				if (action->type == ACTION_TYPE_TOGGLE_SNAP_TO_EDGE
+						&& view->maximized == VIEW_AXIS_NONE
+						&& !view->fullscreen
+						&& view_is_tiled(view)
+						&& view->tiled == edge) {
+					view_set_untiled(view);
+					view_apply_natural_geometry(view);
+					break;
+				}
 				view_snap_to_edge(view, edge,
 					/*across_outputs*/ true,
 					/*store_natural_geometry*/ true);
@@ -1122,6 +1142,7 @@ actions_run(struct view *activator, struct server *server,
 			}
 			view_constrain_size_to_that_of_usable_area(view);
 			break;
+		case ACTION_TYPE_TOGGLE_SNAP_TO_REGION:
 		case ACTION_TYPE_SNAP_TO_REGION:
 			if (!view) {
 				break;
@@ -1133,6 +1154,15 @@ actions_run(struct view *activator, struct server *server,
 			const char *region_name = action_get_str(action, "region", NULL);
 			struct region *region = regions_from_name(region_name, output);
 			if (region) {
+				if (action->type == ACTION_TYPE_TOGGLE_SNAP_TO_REGION
+						&& view->maximized == VIEW_AXIS_NONE
+						&& !view->fullscreen
+						&& view_is_tiled(view)
+						&& view->tiled_region == region) {
+					view_set_untiled(view);
+					view_apply_natural_geometry(view);
+					break;
+				}
 				view_snap_to_region(view, region,
 					/*store_natural_geometry*/ true);
 			} else {
--- a/src/action.c
+++ b/src/action.c
@@ -102,6 +102,7 @@ enum action_type {
 	ACTION_TYPE_GO_TO_DESKTOP,
 	ACTION_TYPE_TOGGLE_SNAP_TO_REGION,
 	ACTION_TYPE_SNAP_TO_REGION,
+	ACTION_TYPE_UNSNAP,
 	ACTION_TYPE_TOGGLE_KEYBINDS,
 	ACTION_TYPE_FOCUS_OUTPUT,
 	ACTION_TYPE_MOVE_TO_OUTPUT,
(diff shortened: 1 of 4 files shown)


src/action.c
```diff
@@ -912,10 +918,17 @@ actions_run(struct view *activator, struct server *server,
...
 			break;
+		case ACTION_TYPE_TOGGLE_SNAP_TO_EDGE:
 		case ACTION_TYPE_SNAP_TO_EDGE:
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
+				if (action->type == ACTION_TYPE_TOGGLE_SNAP_TO_EDGE
+						&& view->tiled == edge) {
+					view_set_untiled(view);
+					view_restore_to(view, view->natural_geometry);
```
jp7677, line 929: Can we end up without a valid `natural_geometry` here?

src/action.c
```diff
@@ -1154,6 +1168,12 @@ actions_run(struct view *activator, struct server *server,
 			const char *region_name = action_get_str(action, "region", NULL);
 			struct region *region = regions_from_name(region_name, output);
 			if (region) {
+				if (action->type == ACTION_TYPE_TOGGLE_SNAP_TO_REGION
+					&& view->tiled_region == region) {
+					view_set_untiled(view);
+					view_restore_to(view, view->natural_geometry);
```
jp7677, line 1174: Same question as above.

jp7677:
Alternatively, or additionally, an "UnSnap" action which just untiles and restore the last natural position, might also be a good idea.

src/action.c
johanmalm: Sorry - can't remember. @jlindgren90 @ahesford - do you know off the top of your heads?

johanmalm:
> Alternatively, or additionally, an "UnSnap" action which just untiles and restore the last natural position, might also be a good idea.

Yes, I have thought about that before. Definitely up for that one too.

jp7677:
> > Alternatively, or additionally, an "UnSnap" action which just untiles and restore the last natural position, might also be a good idea.
> 
> Yes, I have thought about that before. Definitely up for that one too.

Done ;), but not yet tested...

src/action.c
jlindgren90: natural_geometry can be empty for xdg views, but I believe this is correct anyway - the 0x0 gets passed to the client to choose its own size and then we center the view in the commit handler.

jp7677:
I've tested this now on a multi-monitor setup. While it works nice in principle, untiling moved the window back to the old monitor when the tiled windows was moved to the new output before. The last commit addresses this by tweaking the natural geometry before moving a tiled window to a new output. With that, untiling keeps the window on the new output. 

Note that the size of window might be bigger than the output after untiling. Since move-to-desktop also doesn't touches width/height of a windows, untiling doesn't do either. To mitigate this, a fit-to-output action should follow unsnap or toggle-snap-edge/region , like 

```xml
    <keybind key="W-Down">
      <action name="UnMaximize" />
      <action name="UnSnap" />
      <action name="FitToOutput" />
    </keybind>
``` 

when monitor sizes differ.

jp7677:
PS: The CI failure on Arch seems unrelated.

src/view.c
```diff
@@ -2116,8 +2117,22 @@ view_move_to_output(struct view *view, struct output *output)
...
 		view_apply_maximized_geometry(view);
 	} else if (view->tiled) {
+		/*
+		 * Tweak the natural geometry to also be on the new output.
+		 * This ensures that untiling keeps the view on the new
+		 * output and does not jump back to the old output.
+		 */
+		view->pending.x = output_area.x;
+		view->pending.y = output_area.y;
+		view->natural_geometry = view->pending;
```
jp7677, line 2127: I can't  use `view_store_natural_geometry` (https://github.com/labwc/labwc/blob/master/src/view.c#L843) here since that function exists early for tlied windows.

ahesford: This looks strange. Won't the pending geometry have the tiled width and height at this point? We should just copy the new `x` and `y` fields into `view->natural_geometry` instead of also copying the width and height.

jp7677: Thanks, adjusted and untiling still works nicely after moving a tiled view to another output.

jp7677:
I’m just seeing now that there is already a draft PR https://github.com/labwc/labwc/pull/1671 with a similar approach.

tokyo4j:
I suggest the following change:

```diff
diff --git a/include/view.h b/include/view.h
index 3464b4e9..d69a80c3 100644
--- a/include/view.h
+++ b/include/view.h
@@ -460,6 +460,11 @@ bool view_compute_centered_position(struct view *view,
 	const struct wlr_box *ref, int w, int h, int *x, int *y);
 void view_set_fallback_natural_geometry(struct view *view);
 void view_store_natural_geometry(struct view *view);
+/**
+ * view_apply_natural_geometry - applies view->natural geometry, but
+ * repositions it if it doesn't intersect with the geometry of view->output.
+ */
+void view_apply_natural_geometry(struct view *view);
 
 /**
  * view_effective_height - effective height of view, with respect to shaded state
diff --git a/src/action.c b/src/action.c
index 56e5c41a..fef7ded1 100644
--- a/src/action.c
+++ b/src/action.c
@@ -935,7 +935,7 @@ actions_run(struct view *activator, struct server *server,
 				if (action->type == ACTION_TYPE_TOGGLE_SNAP_TO_EDGE
 						&& view_is_tiled(view) && view->tiled == edge) {
 					view_set_untiled(view);
-					view_restore_to(view, view->natural_geometry);
+					view_apply_natural_geometry(view);
 					break;
 				}
 				view_snap_to_edge(view, edge,
@@ -1181,7 +1181,7 @@ actions_run(struct view *activator, struct server *server,
 						&& view_is_tiled(view)
 						&& view->tiled_region == region) {
 					view_set_untiled(view);
-					view_restore_to(view, view->natural_geometry);
+					view_apply_natural_geometry(view);
 					break;
 				}
 				view_snap_to_region(view, region,
@@ -1193,7 +1193,7 @@ actions_run(struct view *activator, struct server *server,
 		case ACTION_TYPE_UNSNAP:
 			if (view && view_is_tiled(view)) {
 				view_set_untiled(view);
-				view_restore_to(view, view->natural_geometry);
+				view_apply_natural_geometry(view);
 			}
 			break;
 		case ACTION_TYPE_TOGGLE_KEYBINDS:
diff --git a/src/view.c b/src/view.c
index ce459c89..33c24e2c 100644
--- a/src/view.c
+++ b/src/view.c
@@ -1036,7 +1036,7 @@ view_constrain_size_to_that_of_usable_area(struct view *view)
 	view_move_resize(view, box);
 }
 
-static void
+void
 view_apply_natural_geometry(struct view *view)
 {
 	assert(view);
@@ -2117,20 +2117,8 @@ view_move_to_output(struct view *view, struct output *output)
 	} else if (view->maximized != VIEW_AXIS_NONE) {
 		view_apply_maximized_geometry(view);
 	} else if (view->tiled) {
-		/*
-		 * Tweak the natural geometry to also be on the new output.
-		 * This ensures that untiling keeps the view on the new
-		 * output and does not jump back to the old output.
-		 */
-		view->natural_geometry.x = output_area.x;
-		view->natural_geometry.y = output_area.y;
-
 		view_apply_tiled_geometry(view);
 	} else if (view->tiled_region) {
-		/* see comment above */
-		view->natural_geometry.x = output_area.x;
-		view->natural_geometry.y = output_area.y;
-
 		struct region *region = regions_from_name(view->tiled_region->name, output);
 		view_snap_to_region(view, region, /*store_natural_geometry*/ false);
 	}
```

</details>

This reuses `view_apply_natural_geometry()` which adjusts `view->natural_geometry` to make it intersect with `view->output`.

ahesford:
I think exposing `view_apply_natural_geometry` is better than creating a new `view_unsnap`.

jp7677:
> I suggest the following change:

Done, thanks!
(and it still works ;))

tokyo4j:
I tested with `<action name="ToggleSnapToEdge" direction="center" />` which is the original purpose of this feature in #1671, but I encountered SIGABRT due to `VIEW_EDGE_INVALID` passed to `view_snap_to_edge()`. Also, invalid `direction` (e.g. `direction="foo"`) throws SIGABRT as well. So the following change is needed:

```diff
diff --git a/src/action.c b/src/action.c
index a7b62c39..f1e03365 100644
--- a/src/action.c
+++ b/src/action.c
@@ -331,7 +331,9 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 	case ACTION_TYPE_SHRINK_TO_EDGE:
 		if (!strcmp(argument, "direction")) {
 			enum view_edge edge = view_edge_parse(content);
-			if ((edge == VIEW_EDGE_CENTER && action->type != ACTION_TYPE_SNAP_TO_EDGE)
+			bool allow_center = action->type == ACTION_TYPE_TOGGLE_SNAP_TO_EDGE
+				|| action->type == ACTION_TYPE_SNAP_TO_EDGE;
+			if ((edge == VIEW_EDGE_CENTER && !allow_center)
 					|| edge == VIEW_EDGE_INVALID) {
 				wlr_log(WLR_ERROR, "Invalid argument for action %s: '%s' (%s)",
 					action_names[action->type], argument, content);
@@ -546,6 +548,7 @@ action_is_valid(struct action *action)
 		arg_name = "command";
 		break;
 	case ACTION_TYPE_MOVE_TO_EDGE:
+	case ACTION_TYPE_TOGGLE_SNAP_TO_EDGE:
 	case ACTION_TYPE_SNAP_TO_EDGE:
 	case ACTION_TYPE_GROW_TO_EDGE:
 	case ACTION_TYPE_SHRINK_TO_EDGE:

```

</details>

LGTM other than that. Thank you for working on this!

jp7677:
> We will also need similar checks for `ToggleSnapToRegion`.

Sorry for missing those checks, I've added them now. If my logical mind hasn't left me (which it does once in a while), the check should be the other way around though, thus bail out when direction is center and we deal with (Toggle)SnapToEdge. I would appreciate a good look at my version (which you actually always do).

I've also added explicit checks for `!view->maximized` and `!view->fullscreen`. The former `view->view_restore_to` handled those cases, but `view_apply_maximized_geometry` just assert when those conditions are hit.

tokyo4j:
> Sorry for missing those checks, I've added them now. If my logical mind hasn't left me (which it does once in a while), the check should be the other way around though, thus bail out when direction is center and we deal with (Toggle)SnapToEdge. I would appreciate a good look at my version (which you actually always do).

I'm not sure if I understand what you mean, but I think you're misunderstanding what (Toggle)SnapToEdge does. They should accept `center` and it behaves like `Maximize` but respects `<core><gap>` (which the reddit user mentioned in #1671 wanted). On the other hand, `direction` argument in `GrowToEdge` and `ShrinkToEdge` just means the direction the window edge should move rather than the tiling state, so `center` is invalid for them. 

> I've also added explicit checks for !view->maximized and !view->fullscreen. The former view->view_restore_to handled those cases, but view_apply_maximized_geometry just assert when those conditions were hit. A maximized/fullscreen view returns true for view_is_tiled when it has been titled before getting maximized/set to fullscreen.

Nice catch! Then I think `!view_is_floating(view)` can be used instead.

src/action.c
```diff
@@ -889,10 +901,20 @@ actions_run(struct view *activator, struct server *server,
 				view_move_to_edge(view, edge, snap_to_windows);
 			}
 			break;
+		case ACTION_TYPE_TOGGLE_SNAP_TO_EDGE:
 		case ACTION_TYPE_SNAP_TO_EDGE:
 			if (view) {
 				/* Config parsing makes sure that direction is a valid direction */
 				enum view_edge edge = action_get_int(action, "direction", 0);
+				if (action->type == ACTION_TYPE_TOGGLE_SNAP_TO_EDGE
+						&& !view->maximized
```
tokyo4j, line 910: nit: let's use `view->maximized != VIEW_AXIS_NONE` to keep consistency with other parts of the codebase.

jp7677:
> ... but I think you're misunderstanding what `(Toggle)SnapToEdge` does.

Yes, this is exactly what happened, somehow I indeed had in mind that center is an invalid edge. Adjusted and thanks for your patience.

src/action.c
jp7677: Actually I would prefer to keep `!view->maximized` to have a consistent toggle-tiling-back-from-maximed-state for all maximized variations.

tokyo4j: Sorry I should've said `view->maximized == VIEW_AXIS_NONE`. I'm so bad at understanding logic.

Well, I prefer `view->maximized == VIEW_AXIS_NONE` (which is exactly the same as `!view->maximized`) since it makes it clear that the view is not maximized in any direction.

> have a consistent toggle-tiling-back-from-maximed-state for all maximized variations.

Sorry I couldn't understand what you mean by this (I guess this is my English skill issue).

jp7677: Nah, I'm even worse at simply reading, somehow I read `VIEW_AXIS_BOTH` while you wrote `VIEW_AXIS_NONE` ;). I've pushed that adjustments and agree that consistency matters.

tokyo4j:
I tested this and it worked fine also with invalid `direction`/`region`.

The last thing I noticed is that it might be more natural to unmaximize the window with `UnSnap`. Of course the combination of `UnMaximize` and `UnSnap` (ref: https://github.com/labwc/labwc/pull/2154#issuecomment-2362931748) will also work, but I don't come up with any demand for `UnSnap` to do nothing when the window is maximized. CC @johanmalm @ahesford

ahesford:
I think making UnSnap imply UnMaximize is appropriate.

jp7677:
> Oops, I tried `gh pr update-branch` command for the first time and did something wrong... I thought it would pull the @jp7677's branch into my local branch, but looks like my understanding was incorrect.
> 
> EDIT: Removed the merge commit I accidentally made.

What I do to pull in remote changes into a local branch is `git pull https://github.com/<user>/<repo> <branch>`, so in this case `git pull https://github.com/jp7677/labwc toggle-tiling`, with a space between repo and branch.

jp7677:
> I tested this and it worked fine also with invalid `direction`/`region`.

Thanks!
 
> The last thing I noticed is that it might be more natural to unmaximize the window with `UnSnap`. Of course the combination of `UnMaximize` and `UnSnap` (ref: [#2154 (comment)](https://github.com/labwc/labwc/pull/2154#issuecomment-2362931748)) will also work, but I don't come up with any demand for `UnSnap` to do nothing when the window is maximized. CC @johanmalm @ahesford

Any thoughts how this should exactly behave? Currently UnSnap has no arguments, but UnMaximize has a direction argument. To be honest, I tend to keep actions separate, also because chaining works nice in this case (which could ofc be mentioned in the docs).

ahesford:
I think UnSnap should behave as if UnMaximize were called with a "both" argument before the action is applied. That is, set the maximization to VIEW_AXIS_NONE and then set untiled and apply the natural geometry.

tokyo4j:
> Any thoughts how this should exactly behave? Currently UnSnap has no arguments, but UnMaximize has a direction argument. To be honest, I think keeping both actions separate is less confusing, also because chaining works nice in this case (which could ofc be mentioned in the docs).

I think it's intuitive to consider `UnSnap` to "make the window floating" (but ignore fullscreened), so unmaximizing into `VIEW_AXIS_NONE` sounds fine to me.

> Edit: like this?

Yes, it is. But we should also handle the cases when `view_is_tiled(view) && view->maximized != VIEW_AXIS_NONE`, like:

```c
		case ACTION_TYPE_UNSNAP:
			if (view && !view->fullscreen && !view_is_floating(view)) {
				view_maximize(view, VIEW_AXIS_NONE,
					/* store_natural_geometry */ false);
				view_set_untiled(view);
				view_apply_natural_geometry(view);
			}
			break;
```

jp7677:
Alright, pushed and also added @tokyo4j as a co-author.

tokyo4j:
LGTM. Thank you for your patience!

PR #2155, opened 2024-09-18
client list combined menu examples

droc12345:
2 examples in rc.xml.all 
one for mouse bind, one for keybind

1 example in menu.xml
show client-list-combined-menu as submenu of root-menu

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -316,6 +316,16 @@
       <action name="SnapToRegion" region="bottom-right" />
     </keybind>
   -->
+  <!-- keybind for client-list-combined-menu - will center in middle of screen -->
+  <!--
+    <keybind key="W-Space">
+      <action name="ShowMenu" menu="client-list-combined-menu" />
+      <position>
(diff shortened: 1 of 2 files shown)


docs/rc.xml.all
```diff
@@ -491,7 +501,7 @@
         <action name="ShowMenu" menu="root-menu" />
       </mousebind>
       <mousebind button="Middle" action="Press">
-        <action name="ShowMenu" menu="root-menu" />
+        <action name="ShowMenu" menu="client-list-combined-menu" />
```
johanmalm, line 504: `rc.xml.all` reflect the built-in default keybinds/mousebinds, so suggest removing this line or adding is (the `client-list-combined-menu`) as an example.

johanmalm:
Just one review-comment. Otherwise happy.

docs/rc.xml.all
droc12345: Done, I put back root-menu and made client-list-combined-menu a commented out example.

Issue #2159, opened 2024-09-19
"go there..." and I noticed an empty space


droc12345:
Hm, didn't quite understand what happens with "go there..." and I noticed an empty space, maybe not the right place to discuss this.
![immagine](https://github.com/user-attachments/assets/a1d0a0ad-db80-4069-82c6-eb6694732746)

_Originally posted by @stefonarch in https://github.com/labwc/labwc/issues/2158#issuecomment-2360938506_

droc12345:
I don't run lxqt, so I've never seen that.

It looks like there are 2 views, that don't have titles.
if you move down from LXQT Appearance Config* does it highlight?

stefonarch:
Just happened again, so my edit wasn't the reason. I can highlight the empty spaces  and it will finish at the top of the list. With some fiddling it looks like VLC is the reason, also if minimized to tray. Opened a bunch of other apps, all fine. No idea why, it doesn't show unusual things with `lswt`.

droc12345:
I don't have vlc installed, but I may install just to see what's going on.

droc12345:
Is vlc running under Xwayland?

droc12345:
I guess the easy way is to filter out titles if null. I'm assuming it's null and not just spaces.
I can't install vlc as it wants me to install the whole X world.

@johanmalm would this work?

```
 		wl_list_for_each(view, &server->views, link) {
 			if (view->workspace == workspace) {
+				char *Title = view_get_string_prop(view, "title");
+				if (string_null_or_empty(line)) {
+					continue;
+				}
+
 				if (view == server->active_view) {
 					buf_add(&buffer, "*");
 				}
-				buf_add(&buffer, view_get_string_prop(view, "title"));
+				buf_add(&buffer, Title);
 
 				current_item = item_create(menu, buffer.data, /*show arrow*/ false);
 				current_item->id = xstrdup(menu->id);
```

stefonarch:
Just checked, in fact it's an xwayland issue, I checked with element and it's identical. It was just by chance that there was VLC in the quicklaunch of my testuser. `QT_QPA_PLATFORM=xcb lxqt-config-input ` produces the same.

droc12345:
When labwc goes to update foreign title, we check -- if (!view->toplevel.handle || !title) and then bail
which is why lswt and others don't show an empty title.

I think we want to do the same thing, but I'll wait for johan to say something.

stefonarch:
[...]

PR #2161, opened 2024-09-19
client list combined empty title fix

droc12345:
this filters out views that have no title or xwayland toplevel

@johanmalm  this is referenced in #2159 discussion

I'm not sure if I have to zfree the assisgnment though after I'm done checking.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1001,10 +1001,15 @@ update_client_list_combined_menu(struct server *server)
 
 		wl_list_for_each(view, &server->views, link) {
 			if (view->workspace == workspace) {
+				const char *title = view_get_string_prop(view, "title");
+				if (!view->toplevel.handle || string_null_or_empty(title)) {
+					continue;
+				}
+
 				if (view == server->active_view) {
 					buf_add(&buffer, "*");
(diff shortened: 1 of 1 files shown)


src/menu/menu.c
```diff
@@ -1001,10 +1001,15 @@ update_client_list_combined_menu(struct server *server)
 
 		wl_list_for_each(view, &server->views, link) {
 			if (view->workspace == workspace) {
+				const char *tmp_title = view_get_string_prop(view, "title");
```
johanmalm, line 1004: Please could we just call this `title` rather than `tmp_title`.

johanmalm:
Looks good. Thanks for catching this.
Just one common on variable name.
No need to zfree() - in fact, we mustn't.

src/menu/menu.c
droc12345: Ok, will change.

droc12345:
> Looks good. Thanks for catching this. Just one common on variable name. No need to zfree() - in fact, we mustn't.

I wasn't completely sure, thus the reason I asked.
I was thinking  it should fall out of scope once it leaves the if ()

PR #2165, opened 2024-09-20
Refactor the logic of placing client-menu with ShowMenu action

tokyo4j:
Supersedes #2149.

In #2146, we assumed `ShowMenu` action is not bound to any buttons other than window menu button and always place the client-menu under the window-menu button when `atCursor="no"`. Also, it was going to be difficult to distinguish whether the action is executed from the window menu button or the window icon, which will be added soon.

This PR fixes it to open the menu under the actually-clicked button by passing `cursor_context` to `actions_run()`, with some refactoring:
- `seat->pressed.resize_edges` is removed and it's calculated from the cursor position and `seat->pressed.type` just before running Resize action. This slightly changes the existing logic to determine the resizing edges with Alt-Right + Drag mousebinding, but `seat->pressed.type` is still stored on button press so it doesn't bring back https://github.com/labwc/labwc/issues/543.
- `seat->pressed.toplevel` is removed and `get_toplevel()` in `update_pressed_surface()` may be called more often, but its overhead will be negligible.


johanmalm:
Nice. Thanks.

johanmalm:
Might it also make sense to do the patch below to avoid the gap (=padding.width) between the client-menu and window edge when the button that triggers the menu is the left-most one.

```
index 07b2fb3..132127a 100644
--- a/src/action.c
+++ b/src/action.c
@@ -655,6 +655,7 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 		update_client_send_to_menu(menu->server);
 	}
 	/* Place menu in the view corner if desired (and menu is not root-menu) */
+	struct theme *theme = server->theme;
 	if (!at_cursor && view) {
 		x = view->current.x;
 		y = view->current.y;
@@ -664,6 +665,14 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 			assert(ctx->node);
 			int ly;
 			wlr_scene_node_coords(ctx->node, &x, &ly);
+
+			/*
+			 * ...but if it is the left-most button, just align it
+			 * to the window edge and disregarding titlebar padding.
+			 */
+			if (x - view->current.x <= theme->padding_width) {
+				x = view->current.x;
+			}
 		}
 	}
 
@@ -675,7 +684,6 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 		struct output *output = output_nearest_to(server,
 				server->seat.cursor->x, server->seat.cursor->y);
 		struct menuitem *item;
-		struct theme *theme = server->theme;
 		int max_width = theme->menu_min_width;
 
 		wl_list_for_each(item, &menu->menuitems, link) {
```

PR #2166, opened 2024-09-21
xdg: don't always adjust the window size before configure

### LINKED ISSUE #2142: When the width of the window is less than the length of the title, there will be a noticeable shrinkage process
Please See the video 

https://github.com/user-attachments/assets/51a08691-2133-46f0-8eef-69d601a38207

tokyo4j:
Fixes #2142 by just removing `view_adjust_size()` from `xdg_toplevel_view_configure()`.

Maybe we should still adjust the window size before when resizing windows via actions (https://github.com/labwc/labwc/pull/83#issuecomment-945166313), but I wonder if we really need to do it.

At least we currently have a strange behavior with `<keybind key="" name.action="ResizeRelative" left="-10" />` like the video below and we should fix it when we decide to adjust the window size before resizing windows via actions.

https://github.com/user-attachments/assets/bb1e9ab6-757c-4705-b19f-41ad74001959

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -397,15 +397,6 @@ xdg_toplevel_view_configure(struct view *view, struct wlr_box geo)
 	uint32_t serial = 0;
 
 	/*
-	 * Leave a size of 0x0 unchanged; this has special meaning in
-	 * an xdg-toplevel configure event and requests the application
-	 * to choose its own preferred size.
-	 */
-	if (!wlr_box_empty(&geo)) {
-		view_adjust_size(view, &geo.width, &geo.height);
-	}
-
-	/*
 	 * We do not need to send a configure request unless the size
 	 * changed (wayland has no notion of a global position). If the
 	 * size is the same (and there is no pending configure request)


tokyo4j:
I forgot to explain the cause of #2142; `xdg_toplevel_view_map()` calls `set_initial_position()` which eventually calls `xdg_toplevel_view_configure()`. So even when the client initially submits the window geometry smaller than the minimum size, we send a configure event with the minimum size and the client re-submits the window geometry, which causes a strange shrinkage process shown in #2142.

tokyo4j:
> Looks like the lines this PR is removing were added by commit https://github.com/labwc/labwc/commit/3be8fe25f33bdf968a28aeffc08a370ab2d38b49 to avoid a flicker when some clients start maximized.

I don't think that PR is related to this. That PR just wrapped the `view_adjust_size()` in `if (!wlr_box_empty(&geo))` because it changed the semantics of `view->natural_geometry` to accept an empty box in order to let clients submit any window geometry they want when an empty is passed.

From my git-blaming, #83 introduced the enforcement of minimal values before configure and 8e1f11548622297f7810ee153ffd52ff795c3e9d introduced `view_adjust_size()` which also takes into account the size hints from X11 clients.

jlindgren90:
I think I agree with removing view_adjust_size() from xdg_toplevel_view_configure(). We can't force xdg-shell clients to accept a size they don't want, so we should probably just accept the initial size they request as this PR proposes.

I wonder though if we want to do the view_adjust_size() at the next higher level - perhaps in view_move_resize()? Not tested, just thinking out loud...

johanmalm:
Got it. Okay LGTM. Let's go for it.

PR #2168, opened 2024-09-21
docs: some fixes about `<action name="ShowMenu" atCursor="no">`

tokyo4j:
- Add some missing `atCursor="no"` in `rc.xml.all`
- Describe the behavior updated in #2146
  - The added sentence is currently a bit inaccurate as we just search for the window menu button rather than the actual button that opened the menu. #2165 will fix this.

I won't self-merge just in case because I'm bad at writing English.


johanmalm:
Thanks.

PR #2171, opened 2024-09-22
action: place menu at topleft of window when opened by leftmost button

tokyo4j:
It will be more natural to place the menu at the topleft of the window when opened by the leftmost button just like before #2146.

Suggested by @johanmalm (https://github.com/labwc/labwc/pull/2165#issuecomment-2365269283).

Demo after I locally rebased on the window icon PR:
 
https://github.com/user-attachments/assets/cfaf5a79-6eb6-4129-926e-7c1c74a4652a

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -617,6 +617,16 @@ action_list_free(struct wl_list *action_list)
 	}
 }
 
+static enum ssd_part_type
+get_leftmost_button_type(void)
+{
+	struct title_button *b;
+	wl_list_for_each(b, &rc.title_buttons_left, link) {
+		return b->type;
+	}
+	return LAB_SSD_NONE;
+}
+
 static void
 show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 		const char *menu_name, bool at_cursor,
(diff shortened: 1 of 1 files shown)


ahesford:
I disagree with this behavior on the basis of both form and function:
1. Special-casing the behavior for the left-most button is really ugly.
2. With an extreme padding, *e.g.*, `padding.width: 128`, the space between the edge of the view and the leftmost button can be *wider than the menu*, which means you wouldn't even start with the mouse above the menu with this patch.

The beauty of the new behavior is that it makes the menu button behave like it actually spawns the menu, rather than a menu spawning abstractly in an arbitrary location.

tokyo4j:
> 1. Special-casing the behavior for the left-most button is really ugly.
> 2. With an extreme padding, e.g., padding.width: 128, the space between the edge of the view and the leftmost button can be wider than the menu, which means you wouldn't even start with the mouse above the menu with this patch.

I agree. I think we were biased that top-left of the window is the standard position of client-menus. (2.) really makes sense and my suggested change is rather unnatural.

I'll close this PR if @johanmalm agrees with this.

johanmalm:
Okay, @ahesford makes good points.

What I don't like is when we get a small gap between the menu and window edge, for example with `padding.width=2`. Maybe a more acceptable approach would be to introduce a configurable menu-snap setting to snap to the window edge if gap is less than a small number.

Anyway, it's not really a big deal for me. Suggest we close this PR for now.

PR #2174, opened 2024-09-23
Don't detect edge-resizes in configure timeout

### LINKED ISSUE #1922: XDG configure state gets out of sync when clients timeout
I've been seeing an intermittent issue when repeatedly going in and out of fullscreen in Firefox when it's playing back video on a heavily loaded Raspberry Pi 4, using labwc with the gles2 renderer.  I can't get a screenshot, but what I see is the fullscreen video but with a transparent band at the bottom through which I can see my desktop (and can actually click and interact with the desktop too).

I can try to write up a more detailed reproducer if needed.  I'm happy to try fixing this myself if I can get some guidance on context/strategy.

What I believe is happening in my case is:

1. Toggle window maximization, Firefox window size changes from A to B
2. labwc sends XDG configure to Firefox to go from size A to B
3. Firefox doesn't respond in `CONFIGURE_TIMEOUT_MS` because it's being busy/slow
4. labwc discards the pending configure and assumes Firefox is still in state A.
5. Firefox eventually applies the configure and is now in state B
6. Toggle window maximization, Firefox window size changes from B to A
7. labwc thinks Firefox is already in state A so doesn't bother sending a configure
8. Now Firefox and labwc are out of sync, labwc expects size A but Firefox is sending surfaces of size B.  We get the "Window geometry for client (%s) appears to be incorrect - ignoring" error and things look a bit strange.

I'm not sure what the original logic was behind having configures time out. I'm also not sure whether other clients apply configures with a delay or just ignore them entirely. So, it's not obvious to me the best way to fix this, but a few ideas I have are:

- If a configure times out then always send the next configure, even if it appears pointless.  I'm not 100% sure this will entirely solve the problem as stopping after step 5 might still leave things inconsistent.
- If a configure appears to timeout then mark the pending config as "probably forgotten" but still keep track of it.  Then if the client later starts sending surfaces which match this config then apply it.
- Increase `CONFIGURE_TIMEOUT_MS` to make the problem less common.  I expect making the timeout 1000ms would make the problem go away for me at least.  Not sure what the downsides of a longer timeout are.

cillian64:
If a client times out responding to a configure request then the handle_configure_timeout() callback is run.  This cleans up the pending state and moves the view to the pending location, but keeps the current size. The idea is to stop slow applications causing too much lag when the user manipulates the window.  This callback used view_impl_apply_geometry() to actually apply the changes.

view_impl_apply_geometry() contains some heuristics to detect if we're resizing a window from the top, left, or top-left, and if so to do the expected behaviour of keeping the window's bottom/right corner in the same place.  However, that code was erroneously triggering in the case when the user requests to change a window from maximized to fullscreen but the client times out on the configure request. handle_configure_timeout() decides to enact the movement of the window but keep its size at the old size and tells view_impl_apply_geometry() to do that.

The current view position and size is 0,64 1920x1016, the pending position/size is 0,0 1920x1080, and the last committed size is 1920x1016.  Looking at the current and pending position and size, the height changes while the bottom edge stays in the same place so this looks like a top-edge-resize and view_impl_apply_geometry() decides to keep the window's bottom edge in the same place while setting the position according to the last-committed size.  This results in the window staying at position 0,64 size 1920x1016 despite being marked as fullscreen.

My solution to this is just to change handle_configure_timeout() to directly change the view's position and call view_moved() if necessary. The idea of handle_configure_timeout() is to action the window movement now while discarding the size change, and let the size change take place later on when the client catches up.  The logic of view_impl_apply_geometry() doesn't make sense in this case so just avoid it entirely.

Fixes #1922

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -244,8 +244,18 @@ handle_configure_timeout(void *data)
 	view->pending_configure_serial = 0;
 	view->pending_configure_timeout = NULL;
 
-	view_impl_apply_geometry(view,
-		view->current.width, view->current.height);
+	/*
+	 * We do not use view_impl_apply_geometry() here since in the timeout
+	 * case we prefer to always put the top-left corner of the view at the
+	 * desired position rather than anchoring some other edge or corner
+	 */
+	bool moved = view->current.x != view->pending.x
+		|| view->current.y != view->pending.y;
+	view->current.x = view->pending.x;
+	view->current.y = view->pending.y;
+	if (moved) {
+		view_moved(view);
+	}
 
 	/* Re-sync pending view with current state */
 	snap_constraints_update(view);


johanmalm:
Thanks for working this out.
Happy to go with this, but will hang on for half a day or so to see if @jlindgren90 et al any thoughts on the contrary.

jlindgren90:
From a quick look, I think I agree with the code change but not the comment ("the change we're making here is not resizing"). This code doesn't know that the view isn't being resized - but I think it is what we want in the case of a resize too.

Suggest the comment should say something like this instead: "We do not use view_impl_apply_geometry() here since in the timeout case, we prefer to always put the top-left corner of the view at the desired location rather than anchoring some other edge or corner."

jlindgren90:
LGTM, thanks

johanmalm:
Thanks.

PR #2187, opened 2024-09-28
Fix absolute icon name

johanmalm:
@vyivel @tokyo4j would one of you mind doing a review of this one.
I've tested and it works with `gtk-lshw` which has an absolute icon-name.

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -118,7 +118,20 @@ out:
 static int
 process_abs_name(struct icon_ctx *ctx, const char *icon_name)
 {
-	wlr_log(WLR_ERROR, "absolute paths not yet supported");
+	ctx->path = xstrdup(icon_name);
+	if (str_endswith(icon_name, ".png")) {
+		ctx->format = SFDO_ICON_FILE_FORMAT_PNG;
+	} else if (str_endswith(icon_name, ".svg")) {
+		ctx->format = SFDO_ICON_FILE_FORMAT_SVG;
+	} else if (str_endswith(icon_name, ".xpm")) {
+		ctx->format = SFDO_ICON_FILE_FORMAT_XPM;
+	} else {
+		goto err;
+	}
+	return 0;
+err:
+	wlr_log(WLR_ERROR, "'%s' has invalid file extension", icon_name);
+	free(ctx->path);
 	return -1;
 }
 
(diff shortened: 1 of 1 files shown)


src/icon-loader.c
```diff
@@ -101,37 +143,35 @@ icon_loader_lookup(struct server *server, const char *app_id, int size, int scal
 		icon_name = app_id;
```
vyivel, line 156: This creates an interesting case where an application can set its app_id to an absolute path and get it loaded as its icon. An absolute path must only come from `sfdo_desktop_entry_get_icon()`.

src/icon-loader.c
```diff
@@ -120,7 +120,20 @@ process_rel_name(struct icon_ctx *ctx, const char *icon_name,
 static int
 process_abs_name(struct icon_ctx *ctx, const char *icon_name)
 {
-	wlr_log(WLR_ERROR, "absolute paths not yet supported");
+	ctx->path = xstrdup(icon_name);
+	if (str_endswith(icon_name, "png")) {
```
vyivel, line 124: I feel like checking for ".png" (with a dot) would be slightly more robust.

src/icon-loader.c
```diff
@@ -82,6 +83,60 @@ icon_loader_finish(struct server *server)
...
+#endif
+	struct sfdo_icon_file *icon_file = sfdo_icon_theme_lookup(
+		loader->icon_theme, icon_name, SFDO_NT, size, scale,
+		lookup_options);
+	if (!icon_file || icon_file == SFDO_ICON_FILE_INVALID) {
+		ret = -1;
+		goto out;
+	}
+	const char *path = sfdo_icon_file_get_path(icon_file, NULL);
+	if (string_null_or_empty(path)) {
```
vyivel, line 109: This should always be `false`.

src/icon-loader.c
johanmalm: Yes, that's a weird one. Have tweaked logic to avoid that.

src/icon-loader.c
johanmalm: Good point. Done.

src/icon-loader.c
johanmalm: Okay, have just gone for a straight `xstrdup(sfdo_icon_file_get_path())`

I did actually look at the `sfdo_icon_file_get_path()` source - but I'm just paranoid :smile: I'll try to relax.

vyivel:
LGTM

johanmalm:
Thanks. Will rebase and go.

01micko:
Works well.

PR #2189, opened 2024-09-29
Titlebar related breaking changes

johanmalm:
As we are approach the cool-down for the next release I am pondering the
titlebar settings. We have added some slightly conflicting theme options and it
is important that we strike a balance between sane default value and avoiding
some visible breaking changes.

Firstly we have added support for icons which are enabled by default.

Secondly - and over slightly longer than this release cycle - we have moved
from a very simple BUTTON_WIDTH constant which originated in early labwc
commits, to a few more knobs namely:

```
window.button.hover.bg.shape: circle|rectangle (default rectangle)
padding.width: 0
window.button.spacing: 0
window.button.width: 26
```

Whilst these have been given _default_ values to maintain the current look,
themes containing for example `padding.width` will result in a different look.

So, where am I going with this?

1. Our theme-spec is crying out for a `window.button.height`, unless of course
   we just accept that button height will always be the same as button width.
   If you have followed the commits in this release cycle, these dimentions
   relate to the allocated space for each button. Buttons can be smaller (and
   will then center-align) but cannot be any bigger. Also, this allocated space
   is what determines the size of the button hover effect.
2. It feels like the titlebar height should be derived from now only the
   font-height but now also this new `window.button.height` plus of course
   `2 * padding.height` which has always been the case.
3. Following (2) we need `window.button.height` to be 23 for the same 'look'
   based on the fact that on my system font_height() is 17 (using default font)
   so 17 + 2 * 3 (`padding.height`) = 23
4. So now have an odd situation whereby window.button.{width,height} is 26x23,
   which is unlikely to be particularly helpful in keeping icons look crisp.
   For example if you try to use a 24x24 icon, you'll end up distorted. 
   So, I suggest just setting `window.button.height` to 26.
5. Well, now for the next inconsistency. Until the introduction of
   `window.button.height` in this PR, buttons have actually used the whole
   titlebar height. That works fine for rectangular hover-effects spanning
   the full height, but if you want a circular/rectangular hover-effect that's
   a bit smaller you'll be very grateful for `window.button.height`. BUT as a
   result, buttons now respect `padding.height` which is really nice, except
   that the default `padding.height` of 3 changes the default look.
   I suggest therefore that we set the `padding.height` to 0 (as in this PR).
6. Follwoing (3) I'm suggesting that we actually get rid of the theme setting
   `titlebar.height`. It doesn't exist in Openbox spec and was one of those
   "additional" levers that someone wanted (and I had a weak moment and said
   yes :smile:) but it just offers an alternative way (more direct) of setting
   the titlebar height but creates logic that's harder for users to follow.
   I think it's important with UI design that primary objects (buttons in our
   case) decide the geometry and that secondary objects (the titlebar) follow.
   So giving users the `titlebar.height` setting just creates more code for us
   and a false sense of how it works for the users.

What I haven't added in the PR - but we really ought to consider at this time
is that if `padding.{height,width}` is set to zero (and assuming that
`window.button.height` is greater than `font_height()`) then there is no
padding around the window-icon which looks a bit "tight".  So we ought to
consider either/both a setting for the icon-size and icon-margin. For example,
we could change `<theme><icon>` before the next release (beyond which point it
becomes less appealing) to enable something like:

```xml
<theme>
  <iconTheme></iconTheme>
  <iconSize><iconSize>
</theme>
```

--EDIT-- We may also want to consider icon-margin because the icon is a bit special.

Also on the TODO list is 

- [ ] Draw svg icons to the size specified by the icon rather than the
      space allocated by window.button.{width,height}
      Ref: https://github.com/labwc/labwc/pull/2144#issuecomment-2359397672

- [ ] Change `theme->title_height` to `theme->titlebar_height`. Just seems a
      better variable name.

20240929_20h18m48s_grim

### DIFF
--- a/docs/themerc
+++ b/docs/themerc
@@ -8,7 +8,7 @@
 # general
 border.width: 1
 padding.width: 0
-padding.height: 3
+padding.height: 0
 
 # window border
 window.active.border.color: #e1dedb
--- a/docs/themerc
+++ b/docs/themerc
@@ -7,8 +7,13 @@
 
 # general
 border.width: 1
-padding.width: 0
-padding.height: 0
+
+#
+# We do not support the global padding.{width,height} of openbox because
+# the default labwc button geometry has deviates from that of openbox
+#
+window.titlebar.padding.width: 0
+window.titlebar.padding.height: 0
 
 # window border
 window.active.border.color: #e1dedb
--- a/docs/themerc
+++ b/docs/themerc
@@ -10,10 +10,6 @@ border.width: 1
 padding.width: 0
 padding.height: 3
 
-# The following options has no default, but fallbacks back to
-# font-height + 2x padding.height if not set.
-# titlebar.height:
-
 # window border
 window.active.border.color: #e1dedb
 window.inactive.border.color: #f6f5f4
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -75,11 +75,15 @@ ssd_titlebar_create(struct ssd *ssd)
 		/* Buttons */
 		struct title_button *b;
 		int x = theme->padding_width;
+
+		/* Center vertically within titlebar */
+		int y = (theme->title_height - theme->window_button_height) / 2;
+
 		wl_list_for_each(b, &rc.title_buttons_left, link) {
 			struct lab_data_buffer **buffers =
 				theme->window[active].buttons[b->type];
 			add_scene_button(&subtree->parts, b->type, parent,
-				buffers, x, view);
+				buffers, x, y, view);
 			x += theme->window_button_width + theme->window_button_spacing;
 		}
 
@@ -89,7 +93,7 @@ ssd_titlebar_create(struct ssd *ssd)
 			struct lab_data_buffer **buffers =
 				theme->window[active].buttons[b->type];
 			add_scene_button(&subtree->parts, b->type, parent,
-				buffers, x, view);
+				buffers, x, y, view);
 		}
 	} FOR_EACH_END
 
@@ -293,6 +297,8 @@ ssd_titlebar_update(struct ssd *ssd)
 
 	update_visible_buttons(ssd);
 
+	/* Center buttons vertically within titlebar */
+	int y = (theme->title_height - theme->window_button_height) / 2;
 	int x;
 	struct ssd_part *part;
 	struct ssd_sub_tree *subtree;
@@ -307,7 +313,7 @@ ssd_titlebar_update(struct ssd *ssd)
 		x = theme->padding_width;
 		wl_list_for_each(b, &rc.title_buttons_left, link) {
 			part = ssd_get_part(&subtree->parts, b->type);
-			wlr_scene_node_set_position(part->node, x, 0);
+			wlr_scene_node_set_position(part->node, x, y);
 			x += theme->window_button_width + theme->window_button_spacing;
 		}
 
@@ -319,7 +325,7 @@ ssd_titlebar_update(struct ssd *ssd)
 		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
 			part = ssd_get_part(&subtree->parts, b->type);
 			x -= theme->window_button_width + theme->window_button_spacing;
-			wlr_scene_node_set_position(part->node, x, 0);
+			wlr_scene_node_set_position(part->node, x, y);
 		}
 	} FOR_EACH_END
 
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -609,7 +609,7 @@ ssd_update_window_icon(struct ssd *ssd)
 	 */
 	int hpad = theme->window_button_width / 10;
 	int icon_size = MIN(theme->window_button_width - 2 * hpad,
-		theme->title_height - 2 * theme->window_titlebar_padding_height);
+		theme->window_button_height);
 
 	/*
 	 * Load/render icons at the max scale of any usable output (at
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -74,7 +74,7 @@ ssd_titlebar_create(struct ssd *ssd)
 
 		/* Buttons */
 		struct title_button *b;
-		int x = theme->padding_width;
+		int x = theme->window_titlebar_padding_width;
 
 		/* Center vertically within titlebar */
 		int y = (theme->title_height - theme->window_button_height) / 2;
@@ -87,7 +87,7 @@ ssd_titlebar_create(struct ssd *ssd)
 			x += theme->window_button_width + theme->window_button_spacing;
 		}
 
-		x = width - theme->padding_width + theme->window_button_spacing;
+		x = width - theme->window_titlebar_padding_width + theme->window_button_spacing;
 		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
 			x -= theme->window_button_width + theme->window_button_spacing;
 			struct lab_data_buffer **buffers =
@@ -211,7 +211,7 @@ static void
 update_visible_buttons(struct ssd *ssd)
 {
 	struct view *view = ssd->view;
-	int width = view->current.width - (2 * view->server->theme->padding_width);
+	int width = view->current.width - (2 * view->server->theme->window_titlebar_padding_width);
 	int button_width = view->server->theme->window_button_width;
 	int button_spacing = view->server->theme->window_button_spacing;
 	int button_count_left = wl_list_length(&rc.title_buttons_left);
@@ -310,7 +310,7 @@ ssd_titlebar_update(struct ssd *ssd)
 			wlr_scene_rect_from_node(part->node),
 			width - bg_offset * 2, theme->title_height);
 
-		x = theme->padding_width;
+		x = theme->window_titlebar_padding_width;
 		wl_list_for_each(b, &rc.title_buttons_left, link) {
 			part = ssd_get_part(&subtree->parts, b->type);
 			wlr_scene_node_set_position(part->node, x, y);
@@ -321,7 +321,7 @@ ssd_titlebar_update(struct ssd *ssd)
 		part = ssd_get_part(&subtree->parts, LAB_SSD_PART_TITLEBAR_CORNER_RIGHT);
 		wlr_scene_node_set_position(part->node, x, -rc.theme->border_width);
 
-		x = width - theme->padding_width + theme->window_button_spacing;
+		x = width - theme->window_titlebar_padding_width + theme->window_button_spacing;
 		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
 			part = ssd_get_part(&subtree->parts, b->type);
 			x -= theme->window_button_width + theme->window_button_spacing;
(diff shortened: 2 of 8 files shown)


jlindgren90:
I'll try to take a closer read through this tonight. Initial thought - maybe we set the default horizontal padding/spacing values to something small but non-zero, like 2px? So that the window icon doesn't butt right up to the text, or to the screen edge when maximized.

jlindgren90:
What if we change the default button width to 22px but add default padding of 2px on all sides? I'm hoping this would match the current default titlebar height (26px) while giving a little bit of space around the icon. An added benefit is that 22px is one of the recommended raster icon sizes for freedesktop icon themes - 26px is not.

johanmalm:
> What if we change the default button width to 22px but add default padding of 2px on all sides? I'm hoping this would match the current default titlebar height (26px) while giving a little bit of space around the icon. An added benefit is that 22px is one of the recommended raster icon sizes for freedesktop icon themes - 26px is not.

The dilemma with that approach is that the button-hover effect (for icons that don't have a dedicated `foo_hover.{xbm,png,svg}`) would change from going all the way to the end (as it currently is) to a smaller rectangle.

Now we could live with that as a breaking change, but the real problem is that there is no way to configure button-hover effects that go to the edge (i.e. no padding) AND the window icon having some space around it.

Note: The 26px is just the allocated space that the icons are 'capped' at. If a user supplies a 22px button then it'll stay that size (except for SVGs - but that's on the list to fix). For the window icon I think it's important that we specify the size we want (hence the suggestion for an additional setting).

tokyo4j:
I like @johanmalm's proposal.

I gave some thought not to break existing themes, but I gave up. So instead, we should carefully inform users of how they can recover existing themes. Also we should update labwc-artwork with the next labwc release.

> So we ought to consider either/both a setting for the icon-size and icon-margin

Agreed. I suggest adding `window.icon.padding` with a default value of `2` or `3`.

Another thought for circle hover effect - how about replacing `window.button.hover.bg.shape` `window.button.hover.bg.corner-radius`? This will allow both circle and rounded-rect hover effect. My gedit window has rounded-rect hover effects, this may be attractive for some users. For circle hover effect, users can set `window.button.height` to the same value as `window.button.width` and `window.button.hover.bg.corner-radius` to half of them.

jlindgren90:
> The dilemma with that approach is that the button-hover effect ... would change from going all the way to the end (as it currently is) to a smaller rectangle.

Okay, good point. Current default behavior is that the hover/clickable area (what I would think of as the "button size") extends to the edge of the titlebar and has no gaps between buttons. Let's keep it this way (by default).

On the other hand, the window icon shouldn't take up the whole hover area ("button size") so we need a way to set the *icon* size separately. To me it's most clear to add a new `icon.size` setting, less confusing than another `padding` setting. We could set this to 22px by default, which fits nicely in the current 26px button size.

jlindgren90:
Re: the various height settings:

- I like the current behavior that the titlebar height is computed by default from font height + padding.
- I also like the fact that the button (hover area) height is the full height of the titlebar by default. 
- I think we could add an optional `button.height` setting. If it's set and less than the full titlebar height, then the buttons should be centered vertically within the titlebar. If it's unset, then keep the current behavior.

I am not sure what should happen if button height is set and > font height. Do we expand the titlebar to at least button.height + 2x padding? Or just to button.height? Or not at all (and let the buttons be cropped vertically)? I haven't thought this through.

I do not have a much of an opinion either way on `titlebar.height`. I can see some use for it if themes want an easy way to make the titlebar consistent even if the font is changed - but agree there are downsides too.

jlindgren90:
Another idea (for later): make some of the default values proportional, rather than a fixed pixel size. For example the default button width could be the same as the titlebar height, so the buttons are square regardless of the font used. Likewise the icon size could be maybe 85% of the button size (22/26 ≈ 0.846) - could be rounded to common raster sizes if we wanted to get really fancy.

Bonus points: allow percentages in themerc files too. It would take some care to document clearly though.

johanmalm:
Thanks for taking time to think about this.

The more I think about this, the more I feel that there is something missing whatever way we do this because buttons (like max/close) are fundamentally different to the window-icon. A user may wish to have small buttons (6px) close together with a small over effect (12px) - whilst at the same time have a much large icon (22px); or vice versa. What ever way we look at this, with the introduction of window-icons, we need more fine tuning of the padding/margin options.

We could consider overloading `struct ssd_part` for buttons (including the window icons) to add per-object properties like margin and padding.

I think that we fundamentally need to treat buttons diffrently. 

I think we need to conclude something before the next release though.

> Another idea (for later): make some of the default values proportional, rather than a fixed pixel size. For example the default button width could be the same as the titlebar height, so the buttons are square regardless of the font used. Likewise the icon size could be maybe 85% of the button size (22/26 ≈ 0.846) - could be rounded to common raster sizes if we wanted to get really fancy.

> Bonus points: allow percentages in themerc files too. It would take some care to document clearly though.

Yeah!

jlindgren90:
An option for the upcoming release could be to hard-core the icon size as some percentage (not 100%) of the button size, and leave the remainder as padding around it (within the hover area). IMHO that would make it look good enough in most cases, and buy some time to think through additional theme options.

johanmalm:
I dropped the last commit which related to the window-icon size because this was fixed by: https://github.com/labwc/labwc/commit/db3aab77fef5ba4ad98e53cc9e1c2908751dedc3

So in summary this now does the following:

1. Add `window.button.height`
2. Remove `titlebar.height` option
3. Include `window.button.height` in `title_height` calculation
4. Change `padding.height` default to 0

Ready for final review and test.

tokyo4j:
The following patch should round the corner buttons correctly when `padding.height` is not zero:

```diff
diff --git a/src/theme.c b/src/theme.c
index 65d7d1f..d89b152 100644
--- a/src/theme.c
+++ b/src/theme.c
@@ -181,7 +181,7 @@ create_rounded_buffer(struct theme *theme, enum corner corner,
 	struct rounded_corner_ctx rounded_ctx = {
 		.box = &(struct wlr_box){
 			.width = theme->padding_width + width,
-			.height = height,
+			.height = theme->padding_height + height,
 		},
 		.radius = rc.corner_radius,
 		.line_width = theme->border_width,
@@ -189,16 +189,12 @@ create_rounded_buffer(struct theme *theme, enum corner corner,
 		.border_color = white,
 		.corner = corner,
 	};
-	int mask_offset;
-	if (corner == LAB_CORNER_TOP_LEFT) {
-		mask_offset = -theme->padding_width;
-	} else {
-		mask_offset = 0;
-	}
 	struct lab_data_buffer *mask_buffer = rounded_rect(&rounded_ctx);
 	cairo_set_operator(cairo, CAIRO_OPERATOR_DEST_IN);
 	cairo_set_source_surface(cairo,
-		cairo_get_target(mask_buffer->cairo), mask_offset, 0);
+		cairo_get_target(mask_buffer->cairo),
+		corner == LAB_CORNER_TOP_LEFT ? -theme->padding_width : 0,
+		-theme->padding_height);
 	cairo_paint(cairo);
 
 	cairo_surface_flush(cairo_get_target(cairo));

```

</details>

Also, adding a link to following image might help future development:

corner-rounding

johanmalm:
> The following patch should round the corner buttons correctly when `padding.height` is not zero:

Thanks. Will incorporate.

Also, I think I know what we should also do!

1. Stop supporting `padding.width` and `padding.height`. These are old openbox options which many themes contain. As we've deviated from openbox spec by "allocating" much more space for icons (26x26) - adding padding around the icons will lead to much bigger titlebar heights, etc. For example Numix theme uses `padding.height: 6`
2. Instead let's introduce `window.titlebar.padding.{width,height}`. This will give users the same control, but not screw up the default look - nor the look of many common themes (except by maybe 1-3 pixels, but that's a lot better than 6px in the case of Numix).

jp7677:
> 1. Stop supporting `padding.width` and `padding.height`.
> 2. Instead let's introduce `window.titlebar.padding.{width,height}`.

just to be sure, this is really just a rename of the existing options? I prefer a Gnome look and feel without a window icon in the title bar https://github.com/labwc/labwc-artwork/tree/main/themes/Adwaita-Papirus-dark . The sizing and button positioning relies on setting a huge `padding.width` and `titlebar.height` (be it either the old option or the calculated one based on `button.height` and `padding.height` from this current PR). I’m all in of course for a simple rename.

johanmalm:
Yes, just a rename - with the subtle difference that one is part of openbox spec and one isn’t.

jp7677:
> Yes, just a rename - with the subtle difference that one is part of openbox spec and one isn’t.

All good, thanks for the clarification!

tokyo4j:
> 1. Stop supporting padding.width and padding.height. These are old openbox options which many themes contain. As we've deviated from openbox spec by "allocating" much more space for icons (26x26) - adding padding around the icons will lead to much bigger titlebar heights, etc. For example Numix theme uses padding.height: 6
> 2. Instead let's introduce window.titlebar.padding.{width,height}. This will give users the same control, but not screw up the default look - nor the look of many common themes (except by maybe 1-3 pixels, but that's a lot better than 6px in the case of Numix).

I support this change.

Also, I suggest making `window.height` inherit `window.width` by default. This is not a strong preference, but these values are the same in many use-cases and I think this change makes the configuration a little bit easier.

Likewise, I suggest making `window.button.spacing` inherit `padding.width` (or `window.titlebar.padding.width`). This can be done in another PR though.

johanmalm:
Okay. Ready to go. Looking for LGTM.
Suggest we do a merge-commit for this on keep the link to this PR with the pictures.

johanmalm:
> The following patch should round the corner buttons correctly when `padding.height` is not zero:
> patch
> 
> Also, adding a link to following image might help future development:
> 
> corner-rounding

I got a bit confused trying to add this patch. Would you mind doing it on top of this PR?

johanmalm:
> Also, I suggest making `window.button.height` inherit `window.button.width` by default. This is not a strong preference, but these values are the same in many use-cases and I think this change makes the configuration a little bit easier.
> 
> Likewise, I suggest making `window.button.spacing` inherit `padding.width` (or `window.titlebar.padding.width`). This can be done in another PR though.

Suggest we let it settle down a bit first with all the changes we've made.

I think my preference would be not to inherit unless there is a strong case (like with colors).

src/ssd/ssd-titlebar.c
```diff
@@ -603,7 +609,7 @@ ssd_update_window_icon(struct ssd *ssd)
 	 */
 	int hpad = theme->window_button_width / 10;
 	int icon_size = MIN(theme->window_button_width - 2 * hpad,
-		theme->title_height - 2 * theme->padding_height);
+		theme->title_height - 2 * theme->window_titlebar_padding_height);
```
jp7677, line 612: This should be the same as `theme->window_button_height`.

johanmalm: Thanks. Done.

tokyo4j:
> I got a bit confused trying to add this patch. Would you mind doing it on top of this PR?

Alright. Pushed a commit.

tokyo4j:
LGTM. Tested and it worked well.

But I'd like to wait for response from @jlindgren90 since he has better understanding of this issue than me.

docs/themerc
```diff
@@ -7,12 +7,13 @@
...
 border.width: 1
-padding.width: 0
-padding.height: 3
 
-# The following options has no default, but fallbacks back to
-# font-height + 2x padding.height if not set.
-# titlebar.height:
+#
+# We do not support the global padding.{width,height} of openbox because
+# the default labwc button geometry has deviates from that of openbox
```
jlindgren90, line 13: Nit: "has deviates" reads awkwardly. Maybe "the default button geometry of labwc deviates"?

jlindgren90 (approved):
LGTM. For what it's worth, I tested this on my Openbox theme.

What Openbox looks like:
openbox-untweaked

What labwc master looks like (without any labwc-specific adjustments):
labwc-master-untweaked

What this branch looks like:
branch-untweaked

The reduction in vertical padding looks more consistent, and the window buttons seem to be more square now. It feels a bit cramped compared to Openbox, but overall I would still call it an improvement over labwc master. Maybe the padding defaults could be tweaked later.

docs/themerc
johanmalm: Thanks. Will fix this latet.

PR #2190, opened 2024-09-29
xdg: fix de-synced SSD when shrinking Thunderbird window

### LINKED ISSUE #2183: Thunderbird cannot shrink back from maximized to normal window properly
Please see the video.

https://github.com/user-attachments/assets/2035f25c-7d31-49e4-9a94-2e5266826463

Is it an issue of thunderbird or labwc? 

Edited:
the output messeage of command `WAYLAND_DEBUG=client  /usr/bin/thunderbird`
[1.txt](https://github.com/user-attachments/files/17142004/1.txt)

tokyo4j:
Fixes #2183.

This PR modifies the workaround for Qt apps introduced in #1229 by using the toplevel surface size (`surface->current.{width,height}`) instead of the whole surface extent (returned from `wlr_surface_get_extends()`), to update `view->current`.

Using toplevel surface size as the window geometry doesn't sound not right in regard to the xdg-protocol, but I think this change is OK as we're already not strictly following the protocol since #1229 and this doesn't complicate the codebase so much.

CC @jlindgren90 as I'm not 100% sure that this PR doesn't reintroduce any problems.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -161,8 +161,15 @@ handle_commit(struct wl_listener *listener, void *data)
 	 */
 	if (size.width != view->pending.width
 			|| size.height != view->pending.height) {
-		struct wlr_box extent;
-		wlr_surface_get_extends(xdg_surface->surface, &extent);
+		/*
+		 * Not using wlr_surface_get_extend() since Thunderbird
+		 * sometimes resizes the window geometry and the toplevel
+		 * surface size, but not the subsurface size (see #2183).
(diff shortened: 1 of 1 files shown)


jlindgren90:
LGTM. I wasn't ever able to reproduce the original issue with Qt apps, but I think this is the safest approach. Since it fixes Thunderbird, let's go for it.

tokyo4j:
Alright. I'll see the original issue later.

PR #2192, opened 2024-09-30
icon-loader: match more application icons via fallbacks

jlindgren90:
Add fallbacks to:

1) match reverse-URL .desktop file names when the app_id is just the
   base name (e.g. "thunderbird" -> org.mozilla.Thunderbird.desktop)

~2) look in /usr/share/pixmaps when the icon isn't present in the
   hicolor theme (still the case for many applications)~

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -2,6 +2,8 @@
 #include <sfdo-desktop.h>
 #include <sfdo-icon.h>
 #include <sfdo-basedir.h>
+#include <string.h>
+#include <strings.h>
 #include <wlr/util/log.h>
 #include "common/mem.h"
 #include "common/string-helpers.h"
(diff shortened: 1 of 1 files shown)


vyivel:
> look in /usr/share/pixmaps when the icon isn't present in the hicolor theme

libsfdo is supposed to do this already.

jlindgren90:
> libsfdo is supposed to do this already.

Hmm, it is not working for me (tested with asunder, xterm, xmaxima...)

src/icon-loader.c
```diff
@@ -135,6 +137,33 @@ process_abs_name(struct icon_ctx *ctx, const char *icon_name)
...
+	struct sfdo_desktop_entry **entries = sfdo_desktop_db_get_entries(db, &n_entries);
+
+	for (size_t i = 0; i < n_entries; i++) {
+		struct sfdo_desktop_entry *entry = entries[i];
+		const char *desktop_id = sfdo_desktop_entry_get_id(entry, NULL);
+		/* Get portion of desktop ID after last '.' */
+		const char *dot = strrchr(desktop_id, '.');
+		const char *desktop_id_base = dot ? (dot + 1) : desktop_id;
+
+		if (g_ascii_strcasecmp(app_id, desktop_id_base) == 0) {
```
johanmalm, line 159: Prefer `!strcasecmp()`

johanmalm (commented):
LGTM. 
Just one preference to use `strcasecmp()` rather than the glib variant.

src/icon-loader.c
jlindgren90: My concerns with strcasecmp is the locale-specific behavior - it is "implementation defined" in all but the C/POSIX locale. But maybe it's not an issue in practice.

johanmalm: Okay. Let's leave as is.

johanmalm: Ah... sorry you've changed it :smile:

jlindgren90: Oh, I already switched it to strcasecmp. Either way is fine with me.

johanmalm:
Thanks

PR #2188, opened 2024-09-29
icon-loader: support xpm icons

johanmalm:
Cc @01micko @kode54 @vyivel @tokyo4j

PR #2193, opened 2024-09-30
icon-loader: add a new loader for XPM icons, used by many legacy X11 apps

jlindgren90:
Adapted from gdk-pixbuf, see the original at:

  https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/blob/master/gdk-pixbuf/io-xpm.c

rgb.txt is from X.org, see:

  https://gitlab.freedesktop.org/xorg/app/rgb/-/blob/master/rgb.txt

Differences from the gdk-pixbuf version:

- GdkPixbuf replaced with struct lab_data_buffer
- Progressive and in-memory loading code removed
- Two functions that had separate BSD copyright rewritten
- Stores colors as ARGB32 earlier in the decoding process for efficiency
- Limited to 1024x1024 px and 1024 colors to prevent abuse
- Uses struct buf to build strings, in place of manual g_new/g_realloc
- Uses xzalloc/xznew_n for other memory allocations
- Uses g_strlcpy in place of banned strcpy/strncpy/strncat
- Uses standard C types (int, bool, etc.) in place of the GLib ones
- Follows labwc coding style (whitespace, braces, letter case, etc.)
- Et cetera ...


jlindgren90:
No hurry on this :)

If there's a small library we could use instead, that would be fine too. libxpm has an X11 depedency, and I didn't think we wanted to depend on gdk-pixbuf necessarily.

01micko:
For completeness,  I tested out this iteration and I can't tell the difference. All works for me :)

src/img/gen-color-table.pl
```diff
@@ -0,0 +1,78 @@
...
+# Adapted from gdk-pixbuf (LGPL-2.0-or-later)
+
+if (@ARGV != 1) {
+    die "Usage: gen-color-table.pl rgb.txt > xpm-color-table.h\n";
+}
+
+open IN, $ARGV[0] || die "Cannot open $ARGV[0]: $!\n";
+
+@colors = ();
+while (defined($_ = <IN>)) {
```
domo141, line 13: there is always newline (or last line w/o ! ... )so while (<IN>) would suffice, but if this is style somewhere so be it :D

src/img/gen-color-table.pl
```diff
@@ -0,0 +1,78 @@
...
+    next if /^!/;
+    if (!/^\s*([0-9]+)\s+([0-9]+)\s+([0-9]+)\s+(.*\S)\s+$/) {
```
domo141, line 15: could use \d in place of [0-9]

but more importantly, if last line were without newline, the \s+ before $/ might not match,
so more robust alternative would be to have \s*$/ at the end.

src/img/gen-color-table.pl
```diff
@@ -0,0 +1,78 @@
...
+	die "Cannot parse line $_";
+    }
+
+    push @colors, [$1, $2, $3, $4];
+}
+
```
domo141, line 21: close IN or die "close failed: $!\n";

src/img/gen-color-table.pl
```diff
@@ -0,0 +1,78 @@
...
+@colors = sort { lc($a->[3]) cmp lc($b->[3]) } @colors;
```
domo141, line 22: hmm, lowercasing the lines for comparing (but keeping the original casing in @colors.

is the hash code (which the change in check hints) going to do case-insensitive searching....

... if so, one could also remove whitespace (and then remove duplicates) to shorten the list
(there could be duplicates even without doing this suggested whitespace removal..)

anyway that is just a thought that came in my mind, i don't know how good it is

src/img/gen-color-table.pl
```diff
@@ -0,0 +1,78 @@
...
+static const char color_names[] =
+EOT
+
+for $color (@colors) {
+    $name = $color->[3];
+
+    if ($offset != 0) {
+	print qq(\n);
+    }
+    print qq(	"$name\\0");
```
domo141, line 45: it took me while to understand this concatenates all strings to one long one
-- good we're not in c89 where the max string length was like <1000 octets :D

src/img/gen-color-table.pl
```diff
@@ -0,0 +1,78 @@
...
+
+print ";\n\n";
+
+print <<EOT;
+struct xcolor_entry {
+	uint16_t name_offset;
+	uint8_t red;
+	uint8_t green;
+	uint8_t blue;
+};
```
domo141, line 59: 6 octets between 2 items in xcolors[] array. I tried. looks smart.

johanmalm:
Oh wow. We've done a similar thing (see #2188). 

Think yours looks better.

Do you want to cast you eyes over mine to see if anything stands out (like a 180 degree review).

Then I'll review this with a view to merge.

domo141 (commented):
close IN or die ... significant (IMO). that \*$/ may be reasonable. others mainly notes

src/img/gen-color-table.pl
jlindgren90: This whole script was pretty much copied from gdk-pixbuf. I know very little perl - would you care to post a diff of how to improve it?

Alternatively, we could just drop the script entirely - the X11 color definitions have not changed in over a decade and are unlikely to ever again at this point, I expect.

src/img/gen-color-table.pl
jlindgren90: We could rewrite this (simplified?) in Python or something if we want. No need to do the same as gdk-pixbuf exactly, I just took what they had out of laziness.

jlindgren90:
> Oh wow. We've done a similar thing (see #2188).
> 
> Think yours looks better.
> 
> Do you want to cast you eyes over mine to see if anything stands out (like a 180 degree review).
> 
> Then I'll review this with a view to merge.

I left a few comments/questions on yours last night, did not take an exhaustive look through. It looks like it may have been from an earlier port, with all the glib stuff removed? (I left a few glib bits, including the hash table optimization.)

Yours also omits the color table generator. Maybe we just go that route, it's not really needed and I don't have the Perl kung-fu to maintain that script anyway.

src/img/gen-color-table.pl
domo141: So you mean save the generated code ?
(I was wondering how is this used, running at build time? -- or just run once and use the generated code.)
-- for the case this script is going to be preserved in repo, I'll paste diff here:

```
--- gen-color-table.pl-orig	2024-09-30 21:45:22.134618483 +0300
+++ gen-color-table.pl	2024-09-30 21:47:51.115934367 +0300
@@ -7,18 +7,20 @@
     die "Usage: gen-color-table.pl rgb.txt > xpm-color-table.h\n";
 }
 
-open IN, $ARGV[0] || die "Cannot open $ARGV[0]: $!\n";
+open IN, '<', $ARGV[0] or die "Cannot open $ARGV[0]: $!\n";
 
 @colors = ();
-while (defined($_ = <IN>)) {
+while (<IN>) {
     next if /^!/;
-    if (!/^\s*([0-9]+)\s+([0-9]+)\s+([0-9]+)\s+(.*\S)\s+$/) {
+    if (!/^\s*([0-9]+)\s+([0-9]+)\s+([0-9]+)\s+(.*\S)\s*$/) {
 	die "Cannot parse line $_";
     }
 
     push @colors, [$1, $2, $3, $4];
 }
 
+close IN or die "close IN failed: $!\n";
+
 @colors = sort { lc($a->[3]) cmp lc($b->[3]) } @colors;
 
 $offset = 0;
```

src/img/gen-color-table.pl
jlindgren90: The lookup is a binary-search, ASCII case-insensitive (using `g_ascii_strcasecmp`). I suppose removing whitespace would also possible, but a little more complicated and I'd rather not mess with this too much.

src/img/gen-color-table.pl
jlindgren90: I added these fixes to the PR, thanks :)

johanmalm:
LGTM
Happy to go with this.
Go for merge yourself if you want, or let me know if there is nothing else we ought to do and I'll merge.

jlindgren90:
I think I'm happy with this. Let's go for it.

PR #2194, opened 2024-09-30
ssd: also add spacing between innermost button and title

jlindgren90:
Currently, left-aligned title text gets pushed right up to the window icon with no spacing at all, which looks bad IMO.

Before: 
Screenshot 2024-09-30 12:30:03

After:
Screenshot 2024-09-30 12:30:47

themerc has:
```
padding.width: 4
padding.height: 2
window.button.width: 22
window.button.spacing: 4
```

### DIFF
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -272,11 +272,12 @@ update_visible_buttons(struct ssd *ssd)
 	/*
 	 * The corner-left button is lastly removed as it's usually a window
 	 * menu button (or an app icon button in the future).
+	 *
+	 * There is spacing to the inside of each button, including between the
+	 * innermost buttons and the window title. See also get_title_offsets().
 	 */
-	while (width <
-		((button_width * (button_count_left + button_count_right)) +
-			(MAX((button_count_right - 1), 0) * button_spacing) +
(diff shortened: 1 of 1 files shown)


jp7677:
Nice improvement! I'm using different sizes for button spacing and padding:

```
padding.width: 10
window.button.width: 24
window.button.spacing: 13
```

I think it would look slightly better for my case when `padding.width` is applied between buttons and title instead of `window.button.spacing`. That said, the code would certainly look worse. I'm torn between those.

jlindgren90:
That should be doable too. I haven't looked closely at the docs or what Openbox does to see if there's precedent one way or the other.

jp7677:
Actually, I think there is no perfect solution. Also both button spacing and padding width are zero by default.  Considering the simplifications in the code, I tend to prefer your version.

johanmalm:
Thanks.
It feels right to use `spacing` between the objects in the titlebar. Just hadn't thought about this case.
And code looks nicer too.
LGTM. Will go for the merge.

Regarding the comment by @jp7677 - we may need to `icon.margin` type variable because the icon is always going to be a bit special. I think @jp7677 already suggested this in some early discussions on IRC.

@jlindgren90 - If you've got capacity, your views on #2189 would be gratefully received.

PR #2196, opened 2024-10-01
cursor: prevent Drag mousebinds from running without button press

### LINKED ISSUE #2175: Move/resize is unfinished when `Move`/`Resize` action is bound to both `Press` and `Drag` events
With following "odd" configuration, interactive move/resize is not finished after the button is released.
I encountered this bug about a year ago, and am posting it here just because I recalled it now.

```xml
  <mouse>
    <default />
    <context name="Frame">
      <mousebind button="W-Left" action="Press">
        <action name="Move" />
      </mousebind>
        <action name="Move" />
      </mousebind>
    </context>
  </mouse>
```

EDIT: I found a similar bug can happen when I run `Move` from a keybind while the cursor is on the titlebar.

tokyo4j:
Fixes #2175.

For `Drag` mousebinds, `mousebind->pressed_in_context` is set on button press and cleared on cursor motion, which executes actions bounded to the mousebind. However, when interactive move/resize has been already started on cursor motion by other keybinds/mousebinds, `mousebind->pressed_in_context` is not cleared, so the mousebind could be unexpectedly executed after the button is released and the interactive move/resize is finished.

So this PR fixes it by always clearing `pressed_in_context` on button releases.

### DIFF
--- a/include/input/cursor.h
+++ b/include/input/cursor.h
@@ -142,7 +142,7 @@ bool cursor_process_button_release(struct seat *seat, uint32_t button, uint32_t 
  * Finishes cursor button release. The return value indicates if an interactive
  * move/resize had been finished. Should be called after notifying a client.
  */
-bool cursor_finish_button_release(struct seat *seat);
+bool cursor_finish_button_release(struct seat *seat, uint32_t button);
 
 void cursor_init(struct seat *seat);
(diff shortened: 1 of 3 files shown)


johanmalm:
Thanks. LGTM.

Issue #2191, opened 2024-09-30
closed thunderbird from maximization, restarted,then the window appeared in the upper left corner at the first non-maximization 


knm100:
This issue is not caused by the fixing of #2183, Regardless of whether #2183 is fixed or not, it is always there.

https://github.com/user-attachments/assets/1a59f110-4b19-4a84-a755-fd8f9467a775

the output message of command `WAYLAND_DEBUG=client  /usr/bin/thunderbird 2> log1.txt`
[log1.txt](https://github.com/user-attachments/files/17186348/log1.txt)

the output message of command `WAYLAND_DEBUG=server labwc -C /tmp -s thunderbird -d 2> log2.txt`
[log2.txt](https://github.com/user-attachments/files/17186351/log2.txt)

tokyo4j:
Hmm... I don't know how to fix this / whether we should fix this. When thunderbird is launched in maximized state, labwc doesn't know the natural windowed geometry is, so when the window is unmaximized, labwc lets thunderbird submit a window geometry it wants. However, when thunderbird delays to submit the windowed geometry, labwc forgets about the unmaximize event it sent. Then when thunderbird finally submits the windowed geometry, labwc sees as if thunderbird submitted a new window geometry without labwc's intention and just updates the window size without updating its position.

So while #2183 was clearly a bug caused by the workaround for Qt apps, this one is a problem of how labwc perceives clients' states and I'm not sure we can/should fix this. It will be worth looking at implementations of other stacking compositors.

jlindgren90:
Hmm. This is an interesting corner case. Maybe we should use some kind of falllback position if we get to the configure timeout with the unset (0x0) pending geometry - I think that's what happens here.

jlindgren90:
By the way, the `cpulimit` utility is helpful to reproduce this.

knm100:
@tokyo4j @jlindgren90   thank you!

PR #2198, opened 2024-10-01
xdg: try to handle slow un-maximize with empty natural geometry better

jlindgren90:
In the case of an initially-maximized view which is taking a long time to un-maximize (seen for example with Thunderbird on slow machines), we may end up in handle_configure_timeout() with an empty pending geometry. In that case we have no great options (we can't center the view since we don't know the un-maximized size yet), so set a fallback position.

Fixes: #2191

Video running nested (slowed down to 50%):
https://github.com/user-attachments/assets/dfb35d39-e023-4733-a8c3-26f5b8184e34

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -251,16 +251,35 @@ handle_configure_timeout(void *data)
 	view->pending_configure_serial = 0;
 	view->pending_configure_timeout = NULL;
 
-	/*
-	 * We do not use view_impl_apply_geometry() here since in the timeout
-	 * case we prefer to always put the top-left corner of the view at the
-	 * desired position rather than anchoring some other edge or corner
-	 */
-	bool moved = view->current.x != view->pending.x
-		|| view->current.y != view->pending.y;
-	view->current.x = view->pending.x;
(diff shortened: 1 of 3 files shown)


src/xdg.c
```diff
@@ -251,16 +251,36 @@ handle_configure_timeout(void *data)
...
-	 * We do not use view_impl_apply_geometry() here since in the timeout
-	 * case we prefer to always put the top-left corner of the view at the
-	 * desired position rather than anchoring some other edge or corner
-	 */
-	bool moved = view->current.x != view->pending.x
-		|| view->current.y != view->pending.y;
-	view->current.x = view->pending.x;
-	view->current.y = view->pending.y;
-	if (moved) {
+	if (view->pending.x != view->current.x
```
jlindgren90, line 254: Hmm. Maybe we should check `wlr_box_empty(&view->pending)` here first. It occurs to me that view->current and view->pending might both be at 0,0 with an initially-fullscreen view.

johanmalm:
LGTM.
Feels like a good fix for this.

PR #2202, opened 2024-10-02
ssd: unround hover effects when the window is tiled

### LINKED ISSUE #1682: Shadow of close-button on fullscreen window with rounded corners
> Slightly hijacking this thread as reading rounded corners reminded me of something. Had a default labwc session and in the default config rounded corners are active. If a window is set to ~fullscreen~ maximized (`ToggleMaximize`) those corners won't be round anymore (dunno if that is to be expected or intended), but the shadow on the close-button (if the mouse hovers above) will still be with rounded corners.

As requested.
Just noticed this, personally I don't care about rounded corners and therefore no preference in what to expect (should the window still have rounded corners or should the shadow be drawn without).
IIRC was with a build of labwc up to d68376f2

tokyo4j:
Fixes #1682.

This also  fixes another problem that the window menu button as the fallback of the window icon is not rounded correctly:

20241002_16h43m45s_grim

Now the translucent hover effect is stored separately from the non-hover icons and rendered on top of them. This might incur some overhead on pixman renderer, but it's not obvious in my testing.

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -18,12 +18,18 @@
 struct ssd_button {
 	struct view *view;
 	enum ssd_part_type type;
+
 	struct wlr_scene_node *normal;
-	struct wlr_scene_node *hover;
 	struct wlr_scene_node *toggled;
+	/*
+	 * Hover icons provided by user or builtin translucent hover overlay.
+	 * Hover overlays are rendered on top of normal/toggled nodes.
+	 */
+	struct wlr_scene_node *hover;
 	struct wlr_scene_node *toggled_hover;
-	struct wlr_scene_tree *icon_tree;
-	struct wlr_scene_tree *hover_tree;
+
+	struct wlr_scene_tree *untoggled_tree;
+	struct wlr_scene_tree *toggled_tree;
 
 	struct wl_listener destroy;
 };
--- a/include/theme.h
+++ b/include/theme.h
@@ -22,6 +22,12 @@ enum lab_shape {
 	LAB_CIRCLE,
 };
 
+enum ssd_corner {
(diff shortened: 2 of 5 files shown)


tokyo4j:
Now I wonder if this is the best option...

Suppose a user want to set the red hover effect on close button that covers the full height of the titlebar (like native windows in Windows 11). To achieve that, he can provide a rectangular icon file filled with red, but then it's not rounded when the window is untiled in both the current master and this PR. Another possible solution is to add theme properties like `window.button.hover.color: #ff0000` in the future and use the builtin overlay, but this PR makes it harder to achieve that because the hover overlay is stored as `theme->button_hover_overlay_{left,right,middle}`, rather than per-button buffers.

Now I think the better idea is to have something like `button->hover_rounded` and `button->hover_toggled_rounded` that holds rounded variants of `button->hover` and `button->toggled_hover` regardless of it's a fallback hover icon or provided by the user.

For the problem of unrounded window menu button as a fallback of the window icon, we can simply make `icon` and `menu` exclusive in `<theme><titlebar><layout>` as the temporary solution for the next release.

tokyo4j:
I worked on the approach I described above: https://github.com/tokyo4j/labwc/tree/refactor-buttons. But my patch became so large (though it compacts the codebase by 200 LoC) and I guess this can't be merged before the next release.

Here's the demo with rectangular close icon files:

https://github.com/user-attachments/assets/71e17343-fcd0-4ab8-826c-18442fee5e8b

I'll open a PR and close this PR later.

johanmalm:
I like the look of refactor-buttons branch. I’ve had that sort of thing in mind before, but as you note, it’s a big re-write 😊

I’ll review on a big screen at the weekend to give some feedback.

PR #2207, opened 2024-10-03
theme: scale corner radius in hover effect rendering

tokyo4j:
The current `create_hover_fallback()` doesn't scale up the corner radius (and `padding.width`) correctly when the fallback hover effect is drawn on top of a non-hover icon buffer that's larger than the window button size.

So for example, when `close-active.png` with the size of 800x200 is provided, the hover effect looks as if it's not rounded at all:

20241003_09h39m37s_grim 

This PR fixes it by switching the scale of the cairo context with `cairo_scale()`.

This also includes a minor refactor of #2145.


johanmalm:
Thanks

PR #2209, opened 2024-10-03
ssd: limit icon size to ~85% of window_button_width

jlindgren90:
This ensures that the icon doesn't push right up to the window edge (or left-aligned title) when using a large font or narrow button width. In the default config (with 10pt font), it makes no difference since the limiting factor on the icon size is the titlebar height anyway.

I spent way too much time over-thinking how to compute the padding. I think 2px on each side is reasonable (and it should be equal on each side), so window_button_width/10 (rounded down) should be fine. Eventually the padding should probably be configurable anyway.

Previously:
- #2189 (broader discussion covering various options, including this)
- #2194 (works only if themes to set a non-zero `window.button.spacing`)

Before/after, default theme except for left-aligned title, with 10pt font:
Screenshot 2024-10-03 07:40:10
Screenshot 2024-10-03 08:23:51

Before/after, default theme except for left-aligned title, with 16pt font:
Screenshot 2024-10-03 07:41:03
Screenshot 2024-10-03 08:25:38

### DIFF
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -654,7 +654,16 @@ ssd_update_window_icon(struct ssd *ssd)
 
 	struct theme *theme = ssd->view->server->theme;
 
-	int icon_size = MIN(theme->window_button_width,
+	/*
+	 * Ensure a small amount of horizontal padding within the button
+	 * area (2px on each side with the default 26px button width).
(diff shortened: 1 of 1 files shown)


knm100:
Indeed, looks more harmonious.

johanmalm:
Thanks

PR #2210, opened 2024-10-04
icon-loader: add a log handler for sfdo messages

jlindgren90:
Useful for debugging why an icon isn't being loaded.

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -25,6 +25,19 @@ struct icon_loader {
 	struct sfdo_icon_theme *icon_theme;
 };
 
+static void
+log_handler(enum sfdo_log_level level, const char *fmt, va_list args, void *tag)
+{
+	/* add a prefix if the format length is reasonable */
+	char buf[256];
+	if (snprintf(buf, sizeof(buf), "[%s] %s", (const char *)tag, fmt)
+			< (int)sizeof(buf)) {
+		fmt = buf;
+	}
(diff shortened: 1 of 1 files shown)


src/icon-loader.c
```diff
@@ -25,6 +25,18 @@ struct icon_loader {
 	struct sfdo_icon_theme *icon_theme;
 };
 
+static void
+log_handler(enum sfdo_log_level level, const char *fmt, va_list args, void *data)
+{
+	// add a prefix if the format length is reasonable
+	char buf[256];
+	if (snprintf(buf, sizeof(buf), "[sfdo] %s", fmt) < (int)sizeof(buf)) {
```
tokyo4j, line 33: I suggest setting prefixes like `[sfdo-desktop]` and `[sfdo-icon]`. I think the argument `data` can be used for this.

src/icon-loader.c
```diff
@@ -25,6 +25,18 @@ struct icon_loader {
 	struct sfdo_icon_theme *icon_theme;
 };
 
+static void
+log_handler(enum sfdo_log_level level, const char *fmt, va_list args, void *data)
+{
+	// add a prefix if the format length is reasonable
```
tokyo4j, line 31: nit: `/* */` should be used for comments.

tokyo4j:
LGTM other than my minor comments.

johanmalm:
Nice. Thanks.

PR #2211, opened 2024-10-04
build: allow disabling icon when libsfdo is installed system-wide

tokyo4j:
`-Dicon=disabled` should disable window icons even when libsfdo is installed system-wide.


johanmalm:
Thanks. Whoops - should have caught that in original review :smile:

PR #2167, opened 2024-09-21
xdg: handle `show_window_menu` requests

tokyo4j:
Screenshot with gedit:

20240921_19h53m54s_grim

PR #2214, opened 2024-10-04
xdg: fix crash on actions from menu opened by show_window_menu requests

tokyo4j:
Fix crash when executing view-related actions like `Maximize` from a menu opened by `show_window_menu` requests (https://github.com/labwc/labwc/pull/2167#issuecomment-2393015264).

Note that I don't know internals of `menu.c`. I just copied the lines from `show_menu()` in `action.c`.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -418,8 +418,11 @@ handle_request_show_window_menu(struct wl_listener *listener, void *data)
 	struct xdg_toplevel_view *xdg_toplevel_view = wl_container_of(
 		listener, xdg_toplevel_view, request_show_window_menu);
 	struct server *server = xdg_toplevel_view->base.server;
+
 	struct menu *menu = menu_get_by_id(server, "client-menu");
 	assert(menu);
+	menu->triggered_by_view = &xdg_toplevel_view->base;
+
 	struct wlr_cursor *cursor = server->seat.cursor;
 	menu_open_root(menu, cursor->x, cursor->y);
 }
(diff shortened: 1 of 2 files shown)


jp7677:
Yes, this fixes what I described in https://github.com/labwc/labwc/pull/2167#issuecomment-2393015264
Thanks for the fast response!

src/xdg.c
```diff
@@ -418,8 +418,14 @@ handle_request_show_window_menu(struct wl_listener *listener, void *data)
 	struct xdg_toplevel_view *xdg_toplevel_view = wl_container_of(
 		listener, xdg_toplevel_view, request_show_window_menu);
 	struct server *server = xdg_toplevel_view->base.server;
+
 	struct menu *menu = menu_get_by_id(server, "client-menu");
 	assert(menu);
+
+	/* do the same as show_menu() in action.c */
+	menu->triggered_by_view = &xdg_toplevel_view->base;
+	server->menu_current = menu;
```
tokyo4j, line 427: @johanmalm @droc12345
Is `server->menu_current = menu;` needed?

It's added in #2102 and I blindly copied it from `show_menu()`, but looks like it's handled in `menu_open_root()`.

johanmalm: Have had a look at the code. Agree, I don't think it's needed as set in `menu_open_root()`. 

I think we can remove that line from `show_menu()` too.

johanmalm (commented):
LGTM

PR #2215, opened 2024-10-04
icon-loader: also match desktop entries by executable path

jlindgren90:
It's relatively common for X11 applications to set WM_CLASS to the name of the executable, which may not be the same as the name of the .desktop file (for example, "gimp-2.10" vs. gimp.desktop). We can still match these applications by looking at the Exec key.

This currently makes use of an internal libsfdo header since the needed API doesn't exist yet. I have submitted a merge request upstream: https://gitlab.freedesktop.org/vyivel/libsfdo/-/merge_requests/5

I am not sure whether this makes sense to merge now (using the internal header) or wait for the upstream MR and a new libsfdo release.

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -5,6 +5,12 @@
 #include <string.h>
 #include <strings.h>
 #include <wlr/util/log.h>
+/*
+ * FIXME: remove when this is merged:
+ * https://gitlab.freedesktop.org/vyivel/libsfdo/-/merge_requests/5
+ */
+#include <sfdo-desktop/internal.h>
+
 #include "common/mem.h"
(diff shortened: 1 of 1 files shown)


vyivel:
Note that internal libsfdo headers are not installed and therefore this will only work if libsfdo is built as a subproject.

For X11 programs, it would make sense to match by `StartupWMClass`, which is set to `gimp-2.10` in `gimp.desktop`.

jlindgren90:
> For X11 programs, it would make sense to match by StartupWMClass, which is set to gimp-2.10 in gimp.desktop.

Oh, nice, I'll try that.

jlindgren90:
Closing in favor of #2216.

PR #2216, opened 2024-10-04
icon-loader: also match desktop entries StartupWMClass

jlindgren90:
GIMP for example sets its WM_CLASS to "gimp-2.10", which does not match the "gimp.desktop" filename, but matches the StartupWMClass key listed in it.

Thanks to @vyivel for the suggestion.

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -160,6 +160,13 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		if (!strcasecmp(app_id, desktop_id_base)) {
 			return entry;
 		}
+
+		/* Try desktop entry's StartupWMClass also */
+		const char *wm_class =
+			sfdo_desktop_entry_get_startup_wm_class(entry, NULL);
+		if (wm_class && !strcasecmp(app_id, wm_class)) {
+			return entry;
+		}
 	}
 
 	return NULL;


johanmalm:
It was noted by @vyivel on IRC that this would resolve the "XTerm" matches "xterm.desktop" logic in the fuzzy function. Consider removing that.

vyivel:
> Consider removing that.

I don't think it should be removed? The function still does what it says, more or less.

jlindgren90:
There were several other apps besides xterm that needed the existing "fuzzy" logic. I would like to be sure those all still work before removing it.

johanmalm:
> There were several other apps besides xterm that needed the existing "fuzzy" logic. I would like to be sure those all still work before removing it.

Sorry John. I got my wires cross reading vyivel's comment.
LGTM.

jlindgren90:
Quick check of applications on my system that don't get icons with just the StartupWMClass logic, but do with the existing fuzzy logic:

- Beyond Compare
- Leafpad
- xmaxima
- xsane

Also, neither the existing fuzzy logic nor StartupWMClass work for:

- Graphical Disk Map
- Scid vs. PC
- Spotify

... yet qmpanel can find icons for all of these. (I think libsfdo has trouble parsing Spotify's .desktop file, not sure about the others.)

vyivel:
> I think libsfdo has trouble parsing Spotify's .desktop file

Can you send me the file (or a link to it)? Spotify desktop entry is correctly loaded and used to find the icon on my system (installed from Flathub).

jlindgren90:
> Can you send me the file (or a link to it)? Spotify desktop entry is correctly loaded and used to find the icon on my system (installed from Flathub).

The problem is originating from the AUR package. It is adding a `--uri=%U` argument which should use a lowercase `%u` (single URI) instead: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=spotify#n70

PR #2218, opened 2024-10-05
icon-loader: fix a few more missing icons

jlindgren90:
By:
- stripping extensions from relative icon filenames (for example, gdmap.desktop has "Icon=gdmap_icon.png")
- ~lowercasing the app_id before the fallback lookup since WM_CLASS is often capitalized ("Scid" now matches /usr/share/pixmaps/scid.png)~
- using the WM_CLASS "instance" rather than "class" string, which is usually the same but lowercase

~Change strcasecmp() to locale-independent g_ascii_strcasecmp() throughout the file, for consistency with g_ascii_strdown().~

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -114,6 +114,23 @@ struct icon_ctx {
 };
 
 /*
+ * Return the length of a filename minus any known extension
+ */
+static size_t
+length_without_extension(const char *name)
+{
+	size_t len = strlen(name);
+	if (len >= 4 && name[len - 4] == '.') {
+		const char *ext = &name[len - 3];
+		if (!strcmp(ext, "png") || !strcmp(ext, "svg")
+				|| !strcmp(ext, "xpm")) {
+			len -= 4;
+		}
+	}
+	return len;
+}
+
+/*
  * Return 0 on success and -1 on error
  * The calling function is responsible for free()ing ctx->path
  */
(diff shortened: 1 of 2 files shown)


vyivel:
> stripping extensions from relative icon filenames (for example, gdmap.desktop has "Icon=gdmap_icon.png")

This violates the spec. The updated version (from a fork AFAICT) has this problem fixed:

https://salsa.debian.org/debian/gdmap/-/blob/master/data/gdmap.desktop.in

> lowercasing the app_id before the fallback lookup since WM_CLASS is often capitalized ("Scid" now matches /usr/share/pixmaps/scid.png)

Consider matching by WM_INSTANCE as well? This is what GNOME shell does. (Looking at its app matching logic in general might be useful: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/368a685717bb5ae0f7b07e5ff9a80edf7b3a9fa7/src/shell-window-tracker.c#L392)

jlindgren90:
> This violates the spec. The updated version (from a fork AFAICT) has this problem fixed:

I realize it violates the spec, but GLib has basically the exact same workaround:
https://github.com/GNOME/glib/blob/main/gio/gdesktopappinfo.c#L1998

I don't think labwc should stand on principle and refuse to show an icon for a .desktop file that GNOME and other GLib-based environments accept.

> Consider matching by WM_INSTANCE as well? This is what GNOME shell does. (Looking at its app matching logic in general might be useful

Looking into this, thanks again for the suggestion.

jlindgren90:
Updated to use the WM_CLASS 'instance' string. The gnome-shell logic that uses both 'instance' and 'class' seems to be mainly for Chrome webapps, so I am not going to bother with it right now -- it could be something to explore more in future.

johanmalm:
Thanks. LGTM.

PR #2221, opened 2024-10-06
xdg: take into account minimum window size for resizing

tokyo4j:
This is especially relavant for `<resize drawContents="no">`.

Demo (left: before, right: after):

https://github.com/user-attachments/assets/c0363f17-17b9-404d-a425-1b8c8ef8413a

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -44,6 +44,20 @@ xdg_toplevel_from_view(struct view *view)
 	return xdg_surface->toplevel;
 }
 
+static struct view_size_hints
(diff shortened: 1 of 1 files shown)


johanmalm:
Nice improvement.

tokyo4j:
Note: With this PR, some apps that send `xdg_toplevel.xdg_set_min_size` requests with very small size (e.g. foot sets 16x19) can be resized into window sizes smaller than `view_get_min_width()xLAB_MIN_VIEW_HEIGHT`.

PR #2222, opened 2024-10-06
buffer: reduce unnecessary painting to new cairo surfaces

jlindgren90:
Add buffer_adopt_cairo_surface(), which allows wrapping an existing cairo image surface in a struct lab_data_buffer. This is useful when loading PNGs since most will be loaded as ARGB32 already.

Fix a memory leak in the non-ARGB32 PNG case, where we do still need to paint to a new image surface -- we were leaking the original surface.

Eliminate an unnecessary temporary image surface in SVG loading and just render the SVG to the image surface held by the lab_data_buffer.

I also cleaned up and clarified the buffer API a bit:

- Add a pointer to the held cairo_surface_t (so we can still access it if there is no cairo_t).
- Remove the free_on_destroy bool (it was always true).
- Rename unscaled_width/height to logical_width/height and add an explanatory comment. It was unclear what "unscaled" meant.
- Rename buffer_create_wrap() to buffer_create_from_data().

This is laying groundwork for some more icon fixes I am working on (making sure icons are loaded and rendered at the correct scale).

### DIFF
--- a/include/buffer.h
+++ b/include/buffer.h
@@ -32,21 +32,43 @@
 struct lab_data_buffer {
 	struct wlr_buffer base;
 
-	cairo_t *cairo;
-	void *data;
+	cairo_surface_t *surface; /* optional */
+	cairo_t *cairo;           /* optional */
+	void *data; /* owned by surface if surface != NULL */
 	uint32_t format;
 	size_t stride;
-	bool free_on_destroy;
-	uint32_t unscaled_width;
-	uint32_t unscaled_height;
+	/*
+	 * The logical size of the surface in layout pixels.
+	 * The raw pixel data may be larger or smaller.
+	 */
+	uint32_t logical_width;
+	uint32_t logical_height;
 };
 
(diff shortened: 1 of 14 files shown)


johanmalm:
LGTM

There is a lot there in one commit - but have looked through and it seems fine.

As a point of reflection, there is an inconsistency in that some lab-buffers use `cairo_surface_set_device_scale()` (with the expected `wlr_scene_buffer_set_dest_size()`) and others don't.

Of course the inconsistency in using the 'abstract' co-ordinate space already existed before, but it becomes more obvious with this refactoring... which is a good thing I think and which should make it easier to implement for example PNGs at different scales in future.

jlindgren90:
Yep, I have noticed some inconsistency with the logical/device pixels too. The logical size stored with the buffer (which is related to the cairo device scale) isn't being considered in some places. I have some draft fixes for a future PR.

jlindgren90:
I'll try to keep commits smaller in the next PR too. Agree this one probably could have been multiple commits.

PR #2224, opened 2024-10-06
xdg-shell: fix popup on wrong output

johanmalm:
With Qt apps I get a scheduled x coordinate of `-6` for the `File` menu (the left-most one) pushing it over to the next output.

[recording.webm](https://github.com/user-attachments/assets/c7174596-1624-45e0-9759-b7f39dc1fbea)

### DIFF
--- a/src/xdg-popup.c
+++ b/src/xdg-popup.c
@@ -7,6 +7,7 @@
  *	- keeping non-layer-shell xdg-popups outside the layers.c code
  */
 
+#include "common/macros.h"
 #include "common/mem.h"
 #include "labwc.h"
 #include "node.h"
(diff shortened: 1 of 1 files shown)


tokyo4j:
What Qt theme are you using? I'd like to reproduce the problem if I can.

johanmalm:
> What Qt theme are you using? I'd like to reproduce the problem if I can.

I set `QT_QPA_PLATFORMTHEME=qt6ct`

In `qt6ct` I set style to `kvantum`

In `kvantummanager` I currently use `KvGnome`

tokyo4j:
LGTM.

Tested and it worked well.

Maybe we should apply more sophisticated policy; for example, if the gravity of xdg-positioner is bottom-left, pass the top-right corner of the scheduled popup geometry to `output_nearest_to()` and clamp it with the parent window width. But this PR should cover most use-cases of xdg popup.

PR #2225, opened 2024-10-07
Improve icon display especially at larger output scales

jlindgren90:
Very short summary: make icon-loader and ssd agree on what size icons
will be displayed at, and then load icons at the best available quality
for that size.

Before/after comparison, logical icon size 16px, various scales:
icons2

Before/after comparison, logical icon size 22px, various scales:
icons1

(Left-to-right: Xfce4 Terminal, XTerm, Spotify. Xfce4 Terminal has PNGs
at various sizes (but notably not 22px) plus an SVG. XTerm has a single
48px XPM. Spotify has a single huge 512px PNG, which as you can see is
downscaled badly in master.)

Changes in detail (just copied from commit messages):

### icon-loader: load SVG icons at max scale of any usable output

- Add a new function to get the maximum scale of all usable outputs
- Pass the maximum output scale through to img_svg_load(), which
  ultimately calls cairo_surface_set_device_scale() before rendering

### ssd: respect logical size when displaying window icon

get_scale_box() was using the raw pixel size of the icon buffer for
layout, which caused the icon to be incorrectly scaled up if the buffer
scale was >1.

### common/box: add box_fit_within()

Factor out common math from ssd's get_scale_box() for use elsewhere.

### buffer: add buffer_convert_cairo_surface_for_icon()

Which handles:
 - conversion from other pixel formats to ARGB32
 - setting the logical size to the desired display size
 - downscaling very large images using CAIRO_FILTER_GOOD

### DIFF
--- a/include/buffer.h
+++ b/include/buffer.h
@@ -62,6 +62,18 @@ struct lab_data_buffer *buffer_create_cairo(uint32_t logical_width,
 	uint32_t logical_height, float scale);
 
 /*
+ * Create a buffer from an image surface, for display as an icon.
+ *
+ * The surface is either adopted by the buffer (which takes ownership),
+ * or copied and then destroyed.
+ *
+ * This function allows non-ARGB32 source images and converts to
+ * CAIRO_FORMAT_ARGB32 if needed.
+ */
+struct lab_data_buffer *buffer_convert_cairo_surface_for_icon(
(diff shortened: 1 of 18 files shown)


johanmalm:
Thanks. Looks amazing. Have tested. Seems to work fine too.

PR #2226, opened 2024-10-09
cursor: fix button release events sometimes not being sent 

### LINKED ISSUE #2220: After opening a menu from `show_window_menu` request, DnD stops working
After I opened a menu from `xdg_toplevel::show_window_menu` request, drag-and-drop stops working, emitting an error message:

```
00:00:06.990 [../src/dnd.c:24] wrong source for drag request
```

https://github.com/user-attachments/assets/02eaf1a7-6d15-42b0-bbac-d41a8e8dcb30

tokyo4j:
Fixes #2220.

When `wlr_seat_pointer_notify_button()` is called on a button press event, that function must also be called on the subsequent button release event, because otherwise wlroots thinks the button is kept pressed and it causes issues with validating DnD requests from clients, where only one button must be pressed.

This was the case when a CSD client opens a client-menu via `show_window_menu` request after pressing its window with
the right button because we were always not notifying button release events while a menu is open.

So let's keep track of bound (pressed but not notified) buttons and notify button release events only when the button is not bound, like we are doing for key-state:

https://github.com/labwc/labwc/blob/a5d89a2e4c610b5ed51011a49815d8466ac46643/src/input/keyboard.c#L312

### DIFF
--- a/include/common/set.h
+++ b/include/common/set.h
@@ -1,0 +1,19 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_SET_H
+#define LABWC_SET_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#define LAB_SET_MAX_SIZE 16
+
+struct lab_set {
+	uint32_t values[LAB_SET_MAX_SIZE];
+	int size;
+};
+
+bool lab_set_contains(struct lab_set *set, uint32_t value);
+void lab_set_add(struct lab_set *set, uint32_t value);
+void lab_set_remove(struct lab_set *set, uint32_t value);
+
+#endif /* LABWC_SET_H */
--- a/src/common/set.c
+++ b/src/common/set.c
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <wlr/util/log.h>
+#include "common/set.h"
+
+bool
+lab_set_contains(struct lab_set *set, uint32_t value)
+{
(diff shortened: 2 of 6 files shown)


src/common/set.c
```diff
@@ -0,0 +1,42 @@
...
+}
+
+void
+lab_set_remove(struct lab_set *set, uint32_t value) {
+	bool shifting = false;
+
+	for (int i = 0; i < LAB_SET_MAX_SIZE; ++i) {
+		if (set->values[i] == value) {
+			--set->size;
+			shifting = true;
```
vyivel, line 35: If the order doesn't matter:

```c
set->values[i] = set->values[--set->size];
break;
```

tokyo4j: That sounds efficient, but I'd like to stick the existing method for clarity.

include/common/set.h
```diff
@@ -0,0 +1,19 @@
...
+#include <stdint.h>
+
+#define LAB_SET_MAX_SIZE 16
+
+struct lab_set {
+	uint32_t values[LAB_SET_MAX_SIZE];
+	int size;
+};
+
+bool lab_set_exists(struct lab_set *set, uint32_t value);
```
jlindgren90, line 15: I like the new `set` utilities, nice work. I'd suggest naming this `lab_set_contains` as "contains" is the more typical name for this method.

tokyo4j: Good idea. Fixed.

johanmalm:
LGTM. Have tested and seems fine.
Thanks for catching + implementing this.

PR #2231, opened 2024-10-12
theme: replace `window.button.hover.bg.shape` with `window.button.hover.bg.corner-radius`

tokyo4j:
My GTK theme (Arc-Dark) has a hover effect with rounded corners, so I think it's worth supporting it rather than only supporting `rectangle` and `circle`. The existing `circle` hover effect can be reproduced if you set `window.button.hover.bg.corner-radius` more than half the button size and `window.button.{width,height}` equal.

20241012_15h18m11s_grim

I will not stick with this change, however, as I'll keep using the default theme anyway. I just wanted to confirm the right approach for the customization of the hover effect before the next release.

CC @johanmalm @jp7677

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -126,9 +126,13 @@ labwc-config(5).
 	Space between titlebar buttons, in pixels.
 	Default is 0.
 
-*window.button.hover.bg.shape*
-	The shape of the hover effect of a titlebar button: "rectangle" or "circle".
-	Default is "rectangle".
+*window.button.hover.bg.corner-radius*
+	Radius of the hover effect of a titlebar button, in pixels.
+	Default is 0.
+
+	Note: for a circular hover effect, set *window.button.width* and
+	*window.button.height* equal and *window.button.hover.bg.corner-radius* half
+	of them.
 
 *window.active.button.unpressed.image.color*
 	Color of the images in titlebar buttons in their default, unpressed,
--- a/docs/themerc
+++ b/docs/themerc
@@ -36,7 +36,7 @@ window.button.width: 26
 window.button.spacing: 0
 
 # window button hover effect
-window.button.hover.bg.shape: rectangle
+window.button.hover.bg.corner-radius: 0
 
 # window buttons
 window.active.button.unpressed.image.color: #000000
--- a/include/theme.h
+++ b/include/theme.h
@@ -18,11 +18,6 @@ enum lab_justification {
 	LAB_JUSTIFY_RIGHT,
 };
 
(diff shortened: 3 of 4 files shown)


jp7677:
I like this refinement. 

What comes to my mind is that the title bar corner radius is configured in `rc.xml` (https://labwc.github.io/labwc-config.5.html#entry_theme_cornerradius), I guess it would be more consistent to have both options close to each other, but not sure where this should be and things are breaking we shuffle the existing option around.

> The existing circle hover effect can be reproduced if you set window.button.hover.bg.corner-radius more than half the button size and window.button.{width,height} equal.

May be good to document this. This is also the upper limit, so higher values have no effect anymore?

tokyo4j:
> What comes to my mind is that the title bar corner radius is configured in rc.xml (https://labwc.github.io/labwc-config.5.html#entry_theme_cornerradius), I guess it would be more consistent to have both options close to each other, but not sure where this should be and things are breaking we shuffle the existing option around.

I agree that having those options separately in `rc.xml` and `themerc` is not optimal. But I think the corner radius of hover button should be accessible to theme providers, so it should at least be in `themerc`.

I think we can deprecate `<theme><cornerRadius>` and move it to something like `window.titlebar.corner-radius` (or maybe `window.corner-radius` for when rounded bottom corners is added) in the future.

> May be good to document this.

That's fair. I'll add a note for it soon.

> This is also the upper limit, so higher values have no effect anymore?

Yes, the value is capped at half the button size:

```c
	int min_button_hover_radius =
		MIN(theme->window_button_width, theme->window_button_height) / 2;
	if (theme->window_button_hover_bg_corner_radius > min_button_hover_radius) {
		theme->window_button_hover_bg_corner_radius = min_button_hover_radius;
	}
```

jp7677:
> I think we can deprecate `<theme><cornerRadius>` and move it to something like `window.titlebar.corner-radius` (or maybe `window.corner-radius` for when rounded bottom corners is added) in the future.

Yes, this is probably the best way forward. Considering that the next release already has breaking changes, may be now is also a good moment. But that is a tough decision and fortunately not mine to take ;)

johanmalm:
I like this. It's nice to cover more use-cases with just the one setting.

LGTM and have tested, so will go for merge.

> I think we can deprecate `<theme><cornerRadius>` and move it to something like `window.titlebar.corner-radius` (or maybe `window.corner-radius` for when rounded bottom corners is added) in the future.

Have thought about this before and agree that it would sit better in themerc. Suggest we leave it until after release to honour the 'cool down' period. I think this ought to have a soft deprecation where we support both for a couple of releases.

PR #2236, opened 2024-10-13
src/theme.c: fix misuse of wl_list_for_each

tokyo4j:
I misunderstood when I was writing #2213; the iterator value of `wl_list_for_each()` is initialized whether or not the list is empty.

This reminds me of [a topic from linux kernel](https://lwn.net/ml/linux-kernel/20220217184829.1991035-1-jakobkoschel@gmail.com/).


johanmalm:
Good catch. Better (!) to only use iterator pointer inside loop.

Issue #2238, opened 2024-10-14
Pipemenus Work in Openbox, Not Labwc


ElliotLurie:
I have the following pipemenu setup from `obmenu-generator` in both `~/.config/labwc/menu.xml` and `~/.config/openbox/menu.xml`:
```
<?xml version="1.0" encoding="utf-8"?>
<openbox_menu xmlns="https://openbox.org/"
 xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="https://openbox.org/">
    <menu id="root-menu" label="obmenu-generator" execute="/usr/bin/obmenu-generator" />
</openbox_menu>
```

In Openbox, the menu is displayed properly. However, Labwc just displays the fallback one. I've tried other menu generators with the same results. I am on Arch Linux and have experienced this with both the latest versions of `labwc` and `labwc-git` from the AUR.

01micko:
You should use [labwc-menu-generator](https://github.com/labwc/labwc-menu-generator).

I've generated the following from a script that starts at login:

```xml
<?xml version="1.0" encoding="UTF-8"?>

<openbox_menu>
  <menu id="root-menu" label="root-menu">
    <item label="Web browser"><action name="Execute" command="x-www-browser" /></item>
    <item label="Terminal"><action name="Execute" command="x-terminal-emulator" /></item>
    <item label="Text Editor"><action name="Execute" command="geany" /></item>
    <item label="File Manager"><action name="Execute" command="thunar" /></item>
    <item label="Screen Magnifier"><action name="ToggleMagnify" /></item>
  <menu id="Windows" label="Windows" execute="app-list-cut.sh" /> <!-- pipemenu -->
  <menu id="Applications" label="Applications" execute="labwc-menu-generator -p -t 'lxterminal -e'" /> <!-- pipemenu -->
  <separator />
  <menu id="Exit" label="Exit">
    <item label="Reconfigure"><action name="Reconfigure" /></item>
    <item label="Logout"><action name="Execute" command="wlogout" /></item>
  </menu> <!-- exit-menu -->
  </menu> <!-- root-menu -->
</openbox_menu>
```

Native openbox menus are not a drop in replacement.

BlindRepublic:
I've tried `labwc-menu-generator` for static menus before and, yes, they work. I'm talking about pipe menus here and they do not, even with your suggested binary.

The manpage for `labwc-menu` confirms that the following _should_ work:
```
<openbox_menu>
[...]

PR #2239, opened 2024-10-14
menu: support pipemenu with the toplevel `<menu>` element

johanmalm:
For example:

```xml
<?xml version="1.0"?>
<openbox_menu>
  <menu id="root-menu" label="" execute="obmenu-generator"/>
</openbox_menu>
```

Fixes: #2238

Intended for after `0.8.1`

### DIFF
--- a/docs/labwc-menu.5.scd
+++ b/docs/labwc-menu.5.scd
@@ -112,6 +112,16 @@ ID attributes are unique. Duplicates are ignored.
 When writing pipe menu scripts, make sure to escape XML special characters such
 as "&" ("&amp;"), "<" ("&lt;"), and ">" ("&gt;").
 
+A pipemenu can also be used to define the toplevel *<menu>* element. In this
+case the entire menu.xml file would be reduced to something like this (replacing
+obmenu-generator with the menu generator of your choice):
+
+```
+<?xml version="1.0"?>
+<openbox_menu>
+  <menu id="root-menu" label="" execute="obmenu-generator"/>
+</openbox_menu>
+```
 
 # LOCALISATION
 
--- a/include/menu/menu.h
+++ b/include/menu/menu.h
@@ -42,6 +42,7 @@ struct menu {
 	char *id;
 	char *label;
 	struct menu *parent;
+	struct menu_pipe_context *pipe_ctx;
 
 	struct {
 		int width;
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -48,6 +48,7 @@ static struct menuitem *selected_item;
 struct menu_pipe_context {
 	struct server *server;
 	struct menuitem *item;
+	struct menu *top_level_menu;
 	struct buf buf;
 	struct wl_event_source *event_read;
 	struct wl_event_source *event_timeout;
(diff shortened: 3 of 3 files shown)


tokyo4j:
I guess this runs the pipemenu script only once at the start up, which differs from Openbox's behavior.

As a more general approach, I think we should refactor `menu.c` to recreate scene-nodes of menus every time they're opened. This may incur some overhead, but I don't think it becomes a problem since we're already recreating the scene-nodes for `client-list-combined-menu` and `client-send-to-menu` whenever any menu is opened. Also, this approach allows us to resize a menu (with scrolling?) when it doesn't fit in the output.

Consolatis:
This looks like it would block the whole compositor when waiting for obmenu-generator due to use of `g_spawn_command_line_sync()`.

johanmalm:
> This looks like it would block the whole compositor when waiting for obmenu-generator due to use of `g_spawn_command_line_sync()`.

Yes. But it'll only be once at startup. I think it would be hard to do this async, and not convinced there is a need.

Consolatis:
Hm.. so this makes the generated menu static (or require a Reconfigure to change) then. As its executed when parsing `menu.xml` it can also be used for built-in menus like `root-menu` and `client-menu`. I still didn't read through the dynamic menu changes (the ones implementing listing current windows and workspaces) to know if this might interfere here. I'll try to look into this in detail this week. If I forget feel free to ping me here or at IRC.

johanmalm:
> Hm.. so this makes the generated menu static (or require a Reconfigure to change) then.

Yes. I never actually used this approach with Openbox, but I do really like it.

> I'll try to look into this in detail this week. 

Ta

johanmalm:
@tokyo4j @ahesford @jlindgren90 
Consolatis and I have worked on this together so a light-touch independent review would be welcome.

Usage:

<details>
<summary>menu.xml</summary>

```
<?xml version="1.0"?>
<openbox_menu>
  <menu id="root-menu" label="" execute="obmenu-generator"/>
</openbox_menu>
```

</details>

Or if you don't use `obmenu-generator`, you can replace it with a simple script like this:

<details>
<summary>labwc-menu.sh</summary>

```
#!/bin/sh

terminal="foot"

printf '%b\n' '<openbox_pipe_menu>
  <item label="Web Browser">
    <action name="Execute"><command><![CDATA[firefox labwc.github.io]]></command></action>
  </item>
  <item label="Terminal">
  </item>
  <item label="Tweaks">
    <action name="Execute" command="labwc-tweaks"/>
  </item>
  <separator />'

# -b|--bare    Disables header and footer
# -i|--ignore  Excludes all .desktop files listed at the top of this file
labwc-menu-generator -b -t "${terminal}"

printf '%b\n' '<separator />
  <item label="Reconfigure">
    <action name="Reconfigure" />
  </item>
  <item label="Exit">
    <action name="Exit" />
  </item>
</openbox_pipe_menu>'
```

</details>

johanmalm:
Sorry - just squashed two commits. It is ready for review + merge.

jlindgren90:
I am not familiar with this area of the code at all. No objections to merging from my side.

Consolatis:
It just occurred to me that when there is a Reconfigure with a pending toplevel pipemenu this would cause a use-after-free (because the new reconfigure would free the menu that is still references in the pipe context). So two reconfigures in a row with a pipe menu that takes longer to execute  than the delta between those reconfigures (and < 4 seconds because that is our max time before killing the pipemenu process) it would crash.

Thus we need some kind of running pipe invalidation for menus, we do have them for menu items in`item_destroy()` but are missing something in `menu_free()`.

johanmalm:
Hi @tokyo4j I note your "I'll review" on IRC, so will hold off merging until I get a nod (or otherwise).

johanmalm:
> It just occurred to me that when there is a Reconfigure with a pending toplevel pipemenu this would cause a use-after-free (because the new reconfigure would free the menu that is still referenced in the pipe……

Great catch. Thanks.

tokyo4j:
I suggest the following addition for readability:

```diff

diff --git a/src/menu/menu.c b/src/menu/menu.c
index 1f395331..e0182b33 100644
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1372,6 +1372,12 @@ static void
 create_pipe_menu(struct menu_pipe_context *ctx)
 {
 	if (ctx->top_level_menu) {
+		/*
+		 * We execute the scripts for the toplevel pipemenus at startup
+		 * or Reconfigure, but they can be opened before they finish
+		 * execution, usually with their content empty. Make sure they
+		 * are closed and emptied.
+		 */
 		if (ctx->server->menu_current == ctx->top_level_menu) {
 			menu_close_root(ctx->server);
 		}
@@ -1379,6 +1385,7 @@ create_pipe_menu(struct menu_pipe_context *ctx)
 		wl_list_for_each_safe(item, tmp, &ctx->top_level_menu->menuitems, link) {
 			item_destroy(item);
 		}
+
 		menu_level++;
 		current_menu = ctx->top_level_menu;
 		if (!parse_buf(ctx->server, &ctx->buf)) {
```

LGTM other than that. Thanks for working on this.

I'm not still very happy with the fact that the script is executed only at startup or Reconfigure, but it'll require further rewrite of `menu.c`, so this PR will be sufficient as an intermediate solution. I may address in the future.

Consolatis:
Added the comment and empty line. Thanks for the review.

Issue #2148, opened 2024-09-16
menu contrast 


davidphilipbarr:
The menus can visually blend in with certain applications as there is no contrast to separate them, such as a border (or a shadow!). 
Obviously this can be over come by setting a different colour for the menu bg, this might create some visual inconsistencies with other desktop apps. 

i guess this lays somewhere between a feature request and a usability issue, probably mostly the former :)

screen shot for example.
![lab_menu](https://github.com/user-attachments/assets/0ef0ac05-0563-4b82-ba6d-2e55be9e86bd)

tokyo4j:
Yeah, I agree we should support a border for menus. I haven't realized the UX degradation by the lack of border because I always use dark theme, but in your screenshot, it's indeed a problem.

jp7677:
GTK also applies shadows to menus, may be that would also (additionally/alternatively) fit here.

johanmalm:
Openbox theme specification include `menu.border.width` and `menu.border.color`. We just haven't implemented them yet.

droc12345:
I like to set my menu background with slightly off white color, with a hint of transparency

menu.items.bg.color: #f5f2efdd

tokyo4j:
Implemented by #2376.

PR #2240, opened 2024-10-15
add menu border option

davidphilipbarr:
Adds an option for a menu border with menu.border.width and menu.border.color 

not entirely sure this is the best method? I'm also not sure if default should be off, will need some tidying if it looks functionally ok?

#2148 

menb


tokyo4j:
Please fix the style errors from `scripts/checkpatch.pl` and remove the redundant `theme.c`.

tokyo4j:
I opened #2241 that adds `outline_rect`, a rectangular outline that consist of 4 scene_rects. Your solution (drawing a border as difference of two scene_rects) will work, but I think using the new `outline_rect` will make the codebase more consistent as we already have `multi_rect` (a rectangle that consists of nested 3 rectangular outlines with different colors).  

As for this PR, please fix errors from CI and remove the redundant `theme.c`. The check by CI can be locally executed with `scripts/check`.

davidphilipbarr:
Thanks for looking at this, i figured it might not be the *best* way to do it, i'll tidy up as per your advice but feel free to ignore if you have alternative plans, i'm just getting familiar with the code base so my changes are quite basic.  

...how embarrassing leaving that themerc in there :)

johanmalm:
Just thinking ahead - we ought to consider the possibility of padding + rounded corners in future.
Therefore, would it make sense to just generate a background surface and include the border in that?

Ref:
- https://github.com/labwc/labwc/blob/da68df5f5b6b1c7e8729bdfcd1e527e83f547067/src/osd.c#L210-L223
- https://github.com/labwc/labwc/blob/da68df5f5b6b1c7e8729bdfcd1e527e83f547067/src/common/graphic-helpers.c#L81

davidphilipbarr:
I've fixed the CI errors and the separator placement now.

tokyo4j:
> Just thinking ahead - we ought to consider the possibility of padding + rounded corners in future.
Therefore, would it make sense to just generate a background surface and include the border in that?

I think we should stick to scene-rect-based approach in this PR because the widths of menu items are dynamically updated after parsing `menu.xml` or when opening `client-list-combined-menu`/`client-send-to-menu`  in the current implementation.

tokyo4j:
> I think we should stick to scene-rect-based approach in this PR because the widths of menu items are dynamically updated after parsing menu.xml or when opening client-list-combined-menu/client-send-to-menu in the current implementation.

I worked around this problem and implemented a prototype of background + rounded corner + padding support for menus.

https://github.com/labwc/labwc/compare/master...tokyo4j:labwc:rounded-menu

20241018_21h36m20s_grim

If we decide to proceed on this approach, I'll open a PR later.

@davidphilipbarr Sorry if we decide to take this PR over. I really appreciate your contribution that showed a solution for menu borders and moved our discussion forward.

davidphilipbarr:
I much prefer your solution! Happy if i just helped in any way to move things along!

johanmalm:
> I worked around this problem and implemented a prototype of background + rounded corner + padding support for menus.
> 
> [master...tokyo4j:labwc:rounded-menu](https://github.com/labwc/labwc/compare/master...tokyo4j:labwc:rounded-menu)
> 
> If we decide to proceed on this approach, I'll open a PR later.

This patch looks very promising and I think it's the right direction. Suggest we proceed with this approach.

TheESN:
How is this PR going?

tokyo4j:
> How is this PR going?

We've finish some refactoring as preparations for borders + rounded corners + padding around items, so I'll open another PR soon.

PR #2241, opened 2024-10-16
graphic-helper: add outline_rect

tokyo4j:
This PR adds `outline_rect`, a rectangular outline with a single color as a preparation for menu borders (#2148 #2240).

`multi_rect` now uses three `outline_rect`s.

Intended for after 0.8.1


src/common/graphic-helpers.c
```diff
@@ -57,22 +53,63 @@ multi_rect_set_size(struct multi_rect *rect, int width, int height)
...
+		line_width, height - line_width * 2);
+}
+
+static void
+multi_rect_destroy_notify(struct wl_listener *listener, void *data)
+{
+	struct multi_rect *rect = wl_container_of(listener, rect, destroy);
+	for (size_t i = 0; i < 3; i++) {
+		wlr_scene_node_destroy(&rect->outlines[i]->tree->node);
+	}
```
Consolatis, line 84: Is this necessary? wlr_scene should do this anyway if those are children of the main tree node.

Consolatis:
In general this looks good to me. Is there a use-case where we require a single rect rather than a multi rect?

tokyo4j:
> In general this looks good to me. Is there a use-case where we require a single rect rather than a multi rect?

I made this PR to be used for menu borders, but this is superseded by `scaled_scene_rect`. So this PR can be closed if `scaled_scene_rect` works fine.

However, if we conclude that #2394 cannot be fixed as long as we stick to `scaled_scene_rect`, perhaps we should give up on implementing rounded menus (#2352) and replace `scaled_scene_rect` with `outline_rect` in this PR. The cause of #2394 is, I guess, the semantic gap between wlroots and cairo; in 150% scale, wlroots's render pass api may round 1.5px into 2px but cairo may round it into 1px (or keep it 1.5px with anti-aliasing?). So drawing everything through `wlr_scene_rect` should fix the issue.

Consolatis:
I didn't look into that issue yet. However one (to me surprising) point of cairo is that it always treats the coordinates as the center of a line. That is why there are `+ line_width / 2` offsets in `draw_cairo_border()`. I don't know if that is related to the issue though.

src/common/graphic-helpers.c
tokyo4j: Nice catch. Fixed.

tokyo4j:
- I concluded #2394 cannot be fixed as long as we use cairo buffers to draw menu borders. Thus my 3rd commit replaces `scaled_scene_buffer` with `border_rect` (formerly `outline_rect`). This means #2352 is given up for now.
- I renamed `outline_rect` to `border_rect` and added `bg_color` for `border_rect_create()`. This is required to for it to be used as the OSD background.
- As a preparation for wlroots 0.19, my 2nd commit makes sure the size of the `border_rect`s inside `multi_rect` is larger than `2 * border_width`.

> However one (to me surprising) point of cairo is that it always treats the coordinates as the center of a line. That is why there are + line_width / 2 offsets in draw_cairo_border(). I don't know if that is related to the issue though.

This is not directly related the issue. The problem in #2394 is that wlroots renders scene-nodes in position-independent way which cairo is never aware of. [wlroots#3793](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3793) is related.

Consolatis:
> The problem in #2394 is that wlroots renders scene-nodes in position-independent way which cairo is never aware of. 
[wlroots#3793](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3793) is related.

I assume this would help: [wlroots#5027](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5027)?

tokyo4j:
> I assume [wlroots!5027](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5027) would help?

It will theoretically help, if it gives full control of physical coordinates of rendered objects. But I'm so skeptical that we can tolerate its complexity. If we try to fix #2394 with the idea of that wlroots PR, I guess we'll need to compute the physical sizes of the menu scene-nodes and pass fractional coordinates to `wlr_scene_rect_set_size()` or `wlr_scene_buffer_set_dest_size()` (or maybe `cairo_set_line_width()` in `scaled_scene_rect.c`), which seems to be eliminating much of the benefit of abstraction by scene-graph API.

So I'd rather like to render menu borders with scene-rects and let wlroots render all the menu components.

tokyo4j:
> @tokyo4j - are we waiting for this one to be reviewed?

Yes, I appreciate if anyone reviews this one. I've made many changes since I got the last LGTM.

But I gave up rounded corner support for menus with this PR, so please challenge me if you have other opinions.

include/common/graphic-helpers.h
```diff
@@ -9,29 +9,49 @@ struct wlr_scene_tree;
 struct wlr_scene_rect;
 struct wlr_fbox;
 
+struct border_rect {
```
johanmalm, line 12: Toolkits and libraries normally just call "rectangles with borders" rects. In other words there is an assumptions that when you draw a rectangle, you can specify borders. I suppose the `wlr_scene_rect` is quite a basic object in that sense. For a more intuitive API, I'd suggest dropping the `border_` prefix.

So, suggest considering calling it `lab_scene_rect` or simply `rect`

include/common/graphic-helpers.h
```diff
@@ -9,29 +9,49 @@ struct wlr_scene_tree;
...
+	int border_width; /* read-only */
+
+	/* Private */
+	struct wlr_scene_rect *top;
+	struct wlr_scene_rect *bottom;
+	struct wlr_scene_rect *left;
+	struct wlr_scene_rect *right;
+	struct wlr_scene_rect *bg;
+
+	struct wl_listener destroy;
```
johanmalm, line 23: I can live with it, but there is something that seems a bit unbalanced about having nested trees and destroy-listeners in the case of the multi_rect case.

Suggest considering breaking out the code below into a separate struct:

```
int border_width;
struct wlr_scene_rect *top;
struct wlr_scene_rect *bottom;
struct wlr_scene_rect *left;
struct wlr_scene_rect *right;
struct wlr_scene_rect *bg;
```

...leaving the (currently named) `border_rect` with the `tree`, a `*border_rect` and a `destroy`.

johanmalm (commented):
I agree with moving to `wlr_scene_rect` instead of the cairo buffer for the described reasons relating to fractional scaling.

I'm good with the code in principle, but have left some nits/suggestions.

include/common/graphic-helpers.h
johanmalm: Just thinking aloud...

```C
struct rect {
	int border_thickness;
	struct wlr_scene_rect *top;
	struct wlr_scene_rect *bottom;
	struct wlr_scene_rect *left;
	struct wlr_scene_rect *right;
	struct wlr_scene_rect *fill;
}

struct scene_object {
	struct wlr_scene_tree *tree;
	struct wl_listener destroy;
};

struct lab_scene_rect {
	struct scene_object base;
	struct rect rect;
};

struct lab_scene_multi_rect {
	struct scene_object base;
	struct rect rects[3];
};
```

tokyo4j: Yeah I had a similar feeling.

I opened #2809 as an alternative solution. What do you think of it?

johanmalm: > Yeah I had a similar feeling.
> 
> I opened #2809 as an alternative solution. What do you think of it?

That looks a much nicer interface. Great work!

tokyo4j:
Superseded by #2809

Issue #2242, opened 2024-10-16
How is touchscreen cursor emulation supposed to work?


spl237:
In touch.c, the functions touch_down, touch_up and touch_motion can all call cursor emulation functions in cursor.c (cursor_emulate_move_absolute and cursor_emulate_button). I've been playing with this - it is currently impossible to generate a double-click on a touchscreen by tapping, as there seems to be no such thing as a double-tap equivalent to a double-click. Hence I was wondering about forcing touch events to be interpreted as mouse emulations - I found that by hacking seat_update_capabilities in seat.c so that a touch input device returned its capability as WL_SEAT_CAPABILITY_POINTER, this forced the use of the cursor emulation functions.

They don't quite work, unfortunately. If I drag a selection rectangle in the file manager, the rectangle is created on tap and resized on drag, but releasing the screen, while it does generate a cursor_emulate_button RELEASE, does not remove the selection rectangle. 

It also seems to take two taps for a button press to register on our taskbar - not sure if this is something to do with the GTK gesture recognisers I have implemented to detect tap-and-hold behaviour, but I don't think so - it seems that the "clicked" signal on a button is only generated when the button is clicked for the second time.

Is this how touch emulation is supposed to work? I note that for touchpad devices, you can set mouseEmulation to be yes in the rc.xml file, but that this has no effect for touchscreens. As a first step, if it is relatively easy to do, allowing mouseEmulation on touchscreens might be useful?

johanmalm:
I'm not too clued up on the cursor emulation tbh. Have discussed with @jp7677 + others on IRC.

The emulation was introduced for tablets tools, but there is nothing 'tablet' specific about it.

The double-click functionality goes via `is_double_click()` in `cursor.c`

- https://github.com/labwc/labwc/blob/da68df5f5b6b1c7e8729bdfcd1e527e83f547067/src/input/cursor.c#L894

A couple have tried on touchscreens and confirms that "everything works" - i.e. single tap to open menu, double-tap on SSD titlebar to maximize window, tap-and-drag on SSD for moving a window.... except with Gtk applications! Not sure why.

jp7677:
Trying this on a Pi touchscreen with labwc master on Raspberry OS I'm seeing the following:

gnome-terminal (GTK, SSD) 
- interaction with application, e.g. opening a menu, on tap works 
- double tap on the SSD titlebar maximizes and umaximizes
- tap-drag on the SSD titlebar moves the application window

nautilus (GTK, CSD):
[...]

PR #2244, opened 2024-10-17
cursor: send a frame event after emulated button events

tokyo4j:
This fixes the issues reported in #2242 that emulated tablet/touchscreen button events sometimes don't take effect on applications immediately.


johanmalm:
LGTM

Issue #2126, opened 2024-09-04
Enriching the <query> API to match that of openbox'


orfeasxyz:
Openbox offers a lot of different query options especially ones related to the current window state.  

Reference: https://redtide.github.io/openbox-wiki/Help/Actions/#if

johanmalm:
Sounds interesting.

We generally take the view that we wait for someone to write a patch or describe a particular use-case rather than just copying all of Openbox's options. But having said that - am happy to merge to merge those types of query variants.

Out of interest - is there a particular use-case you foresee?

orfeasxyz:
> Sounds interesting.
> 
> We generally take the view that we wait for someone to write a patch or describe a particular use-case rather than just copying all of Openbox's options. But having said that - am happy to merge to merge those types of query variants.
> 
> Out of interest - is there a particular use-case you foresee?

I see your concern regarding the use-case of these features. One quick use case I can think of is the W-Right followed by W-Up to snap a window to the upper right region without having a separate keybind for it like this:

```xml
...
<keybind key="W-U">
  <action name="If">
    <query>
      <snapped>right</snapped>
    </query>
    <then>
      <action name="SnapToRegion" region="up-right"/> <!-- up-right region defined elsewhere -->
    </then>
  </action>
</keybind>
...
```

Issue #2029, opened 2024-07-25
Action Snap to Region and Edge Snap to Region


googl267:
Custom regions in labwc feel under utilized as you cannot use them without a direct keybinding. I propose an action that resizes the window to the nearest user defined region. Additionally replacing the current edge snapping to snap to users region would be nice aswell

Consolatis:
Just for completeness, you can also use the mouse to snap a window to a region by pressing any keyboard modifier after the move has started.

droc12345:
Should be able to add custom regions to client-menu (assuming you have a menu.xml and are defining client-menu)
but you would have to add entries for all regions as an item.

googl267:
My first thought as well, however a nice keybinding for the nearest region and as Consolatis suggested a drag and drop would make things a lot more clean/intuitive for the end user

calston:
Please don't forget about touchscreen users... modifier keys aren't helpful there

Issue #2023, opened 2024-07-24
UnMaximized reverts to previous state, even if that was maximized in one direction


xi:
Related to https://github.com/labwc/labwc/issues/1825

When I execute the sequence `floating` -> `SnapToEdge right` -> `Maximize` -> `UnMaximize`, the window resizes to the SnapToEdge size and position instead of the initial floating one.

To restore the last state that was not maximized in any direction, you can use the `MoveRelative` without any attributes. However, this is not very intuitive.

@Consolatis wrote in IRC that this behavior is intentional. So this issue mostly exists for people like me who are confused about this. `MoveRelative` works as a workaround for now. If you think anything should be changed about this, leave a comment or :+1:.

orfeasxyz:
Not sure of you are still interested in this but you can achieve this behavior by using conditional actions as such:

```xml
<keybind key="W-a">
  <action name="If">
    <query maximized="both"/>
    <then name.action="UnSnap"/>
    <else name.action="Mazimize"/>
  </action>
</keybind>
```

Issue #1837, opened 2024-05-22
Feature Request: Swap windows in a workspace


Fuyukai:
I often have an editor or terminal open on my second screen that I'd like to move back to my primary screen temporarily; and likewise, to get all the windows on my primary screen out of the way but still visible. It would be nice if there was a built-in way to swap them all with a single action without needing to focus the windows and moving them manually in some way.

Consolatis:
Potential approach:

Using the ForEach action with some additional `output="some_output"` query argument. Although that would then likely only work with two monitors because we don't have a `else if` condition.

PR #2245, opened 2024-10-18
query: additional condition options

orfeasxyz:
A patch following up the issue I opened a while back https://github.com/labwc/labwc/issues/2126.

Other (more or less) related issues that I've come across:
https://github.com/labwc/labwc/issues/1837
https://github.com/labwc/labwc/issues/2023
https://github.com/labwc/labwc/issues/2029

List of options from Openbox and their implementation status (reference: http://openbox.org/help/Actions#If) :
- [x] shaded
- [x] maximized
- [x] maximizedhorizontal
- [x] maximizedvertical
- [x] iconified
- [x] focused
- [ ] urgent: There is no concept of urgent views yet.
- [x] undecorated: implemented differently (see below).
- [x] omnipresent
- [ ] activedesktop: Not implemented yet since it requires some extra changes to the way conditional actions are checked and executed. Currently all conditional actions are associated with one or more views, whereas "activedesktop" is associated with the active workspace. 
- [x] desktop
- [x] monitor

⭐ New in Labwc:
- [x] tiled ["up" | "right" | "down" | "left" | "center"]: succeeds if the current view is tiled in the respective direction.
- [x] tiled_region <region_name>: succeeds if the current view is tiled in the respective region.
- [x] decoration ["none" | "border" | "full"]: succeeds if the decoration mode of the current view matches with the given argument. Replaces openbox' \<undecorated\> query.
---

Some example uses:

A custom "ToggleMaximize" that UnSnaps the window back to its floating state
```xml
<keybind key="W-a">
  <action name="If">
    <query maximized="no"/>
    <then name.action="Maximize"/>
    <else name.action="UnSnap"/>
  </action>
</keybind>
```

Tile to top left region by tiling up and then left (for this to work for "down-left" as well, we need a way to express an `else if` statement).
```xml
<keybind key="W-Left">
  <action name="If">
    <query tiled="up"/>
    <then>
      <action name="SnapToRegion" region="up-left"/>
    </then>
    <else>
      <action name="SnapToEdge" direction="left"/>
    </else>
  </action>
</keybind>
```

Close all windows in other workspaces
```xml
<keybind key="W-S-q">
  <action name="ForEach">
    <query desktop="other"/>
    <then name.action="Close"/>
  </action>
</keybind>
```

Toggle between full and border decorations
```xml
<keybind key="W-d">
  <action name="If">
    <query decoration="full"/>
    <then>
      <action name="SetDecorations" decorations="border"/>
    </then>
    <else>
      <action name="SetDecorations" decorations="full"/>
    </else>
  </action>
</keybind>
```


johanmalm:
Thanks for working on this. It's a nice addition.

Suggest we work towards merging after release of `0.8.1` (in a week's time).

There are some `./script/check` warnings to fix. All minor stuff.

```
src/config/rcxml.c:498: CHECK: Blank lines aren't necessary before a close brace '}'
src/view.c:154: CHECK: line length of 105 exceeds 100 columns
src/view.c:160: CHECK: line length of 101 exceeds 100 columns
src/view.c:178: CHECK: line length of 105 exceeds 100 columns
src/view.c:197: CHECK: line length of 108 exceeds 100 columns
src/view.c:199: CHECK: line length of 107 exceeds 100 columns
```

Would you mind updating `docs/labwc-actions.5.scd` with these additions?

The code looks good - except on construct that I'm not sure about. Will add review comment.

src/view.c
```diff
@@ -68,19 +70,30 @@ struct view_query *
 view_query_create(void)
 {
 	struct view_query *query = znew(*query);
-	query->window_type = -1;
+	*query = (struct view_query) {
```
johanmalm, line 73: Can we do this against a heap allocated struct? Feels like `*query` will point to this static memory, which doesn't seem right if we have multiple queries and actually later change one of the members.

@ahesford @jlindgren90 - Does this construct make sense (and I'm just having a wobbly) or should it be changed?

johanmalm (commented):
See note on initializing `query`

src/view.c
orfeasxyz: The object on the right is stack allocated and copied into the memory pointed to by `query` (much like calling `memcpy`). It's also very likely the compiler will optimize it completely and write to the heap object directly.

ahesford (changes requested):
I really don't like the reliance on magic `int` values to distinguish between unset and set true/false, especially because it leverages implementation details of the actual `bool` types we care about, when we should treat them as opaque, abstract types. I also don't like abusing bitwise operators to effect logical constructs (you're just inheriting this, but we might as well fix it right). I suggest using the already defined `enum three_state` in place of all the integers. I suggest the more explicit

```c
static bool
match_tristate(enum three_state desired, bool actual)
{
	switch (desired) {
	case LAB_STATE_ENABLED:
		retrun actual;
	case LAB_STATE_DISABLED:
		return !actual;
	default:
		return false;
	}
}
```

```c
if (match && query->some_tristate) {
	empty = false;
	match = match_tristate(query->some_tristate, view->actual_bool);
}
```

We don't need to combine the results of matches because we short-circuit tests after the first failure.

It might even be cleaner if we drop the `empty` and `match` pair altogether, redefine the match helper as

```c
static bool
match_tristate(enum three_state desired, bool actual)
{
	switch (desired) {
	case LAB_STATE_ENABLED:
		retrun actual ? LAB_STATE_ENABLED : LAB_STATE_DISABLED;
	case LAB_STATE_DISABLED:
		return actual ? LAB_STATE_DISABLED : LAB_STATE_ENABLED;
	default:
		return LAB_STATE_UNSPECIFIED;
	}
}
```

and then make `match` in the `view_matches_query` an `enum three_state`, allowing constructs like

```c
enum three_state match = LAB_STATE_UNSPECIFIED;
[...]
match = match_three_state(query->some_three_state, view->actual_bool);
if (match != LAB_STATE_UNSPECIFIED) {
	goto end_match;
}
[...]

end_match:
return match == LAB_STATE_ENABLED;
```
but this will require proper handling of the `match` variable in the string comparison blocks.

src/view.c
johanmalm: Got it. Need to get more used to writing things that way :smile:

orfeasxyz:
Thanks for pointing out the `struct three_state`, I will stick with the first implementation and refactor the function.

As a side note: I noticed that the <desktop> option is not working as intended. More specifically, one must specify the entire desktop name including the prefix (e.g. "Workspace 1"). I am thinking of using the `workspaces_find` function which means that the <desktop> query works as expected + we get a few extra options such as querying whether a view is on the last used workspace. A small problem however is that we cannot configure the wrap behavior for the "left" and "right" options so it either has to be hard-coded or implemented some other time.

orfeasxyz:
@ahesford @johanmalm

I made some refactoring and added a few more options, if you have any suggestions LMK so I can finish everything up and write the docs.

johanmalm:
> @ahesford @johanmalm
> 
> I made some refactoring and added a few more options, if you have any suggestions LMK so I can finish everything up and write the docs.

Look like you've inadvertently pushed some rogue commits. Please could you get rid of those.

@ahesford  - Would you be happy to continue the review / merge of this one.

orfeasxyz:
> > @ahesford @johanmalm
> > I made some refactoring and added a few more options, if you have any suggestions LMK so I can finish everything up and write the docs.
> 
> Look like you've inadvertently pushed some rogue commits. Please could you get rid of those.
> 
> @ahesford - Would you be happy to continue the review / merge of this one.

Yeah it was an accident while I was trying to rebase, I will revert.

ahesford (commented):
I pushed a commit that includes some doc cleanup and, more importantly, changes the interface for `maximized` queries from three separate booleans (which is cumbersom) to a multi-state `yes|no|horizontal|vertical`. Internally, we keep the three tri-state variables (with some renames to make them easier to distinguish), although we can refactor that later if desire strikes. I've also dropped all of your `wlr_log` calls in `view_matches_query`, because they were not universally applied to all tests anyway.

Please review these changes and, if you find them agreeable (and functional---I haven't actually run-tested them), we can squash and merge everything.

ahesford:
Incidentally, if we were to completely eliminate the three tri-states for maximization, I would add a `VIEW_AXIS_INVALID = -1` to the definition of `enum view_axis` (making sure it doesn't break existing logic) and use that in the query struct.

PR #2246, opened 2024-10-18
src/action.c: fix client-send-to-menu not properly updated

tokyo4j:
Before this PR, `client-menu` needed to be opened for `client-send-to-menu` to be updated.


johanmalm:
Thanks.

PR #2249, opened 2024-10-19
src/output.c: (output_frame_notify): skip painting output when sessio…

enometh:
…n is not active.

On switching to a console vt when an application is painting, labwc produces a stream of log messages of the form
```
 `00:00:52.345 [ERROR] [../src/output-state.c:45] Failed to commit frame` lines.
``

This patch tries to follow the lead from the solution to this problem
in https://github.com/WayfireWM/wayfire/pull/2484.patch

As noted in that commit message, this patch depends on unreleased
patches to seatd (master) to work reliably, and there may be a a risk
of locking up the labwc if it is used with the seatd-0.8.0.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -128,6 +128,13 @@ output_frame_notify(struct wl_listener *listener, void *data)
 		return;
 	}
 
+	/*
+	 * skip painting the session when it exists but is not active.
+	 */
+	if (output->server->session && !output->server->session->active) {
+		return;
+	}
+
 	if (!output->scene_output) {
 		/*
 		 * TODO: This is a short term fix for issue #1667,

### CODE REVIEW

**Comment by enometh** [comment]
Please review.his change should be postponed till the next release of seatd., also I'm` not sure if this is the correct place for this change  (I'm currently prevented from running this with the devs first
on chat)

**Comment by johanmalm** [comment]
Thanks. I like the look of this. Will set to 'draft' until `seatd` has been tagged at `>0.8.0`

**Comment by enometh** [comment]
Maybe this can be fixed in wlroots for all wlr clients before that

**Comment by Consolatis** [comment]
Just to mention, the Void-musl runtime tests fails because `output_frame_notify()` is triggered before the session is initialized (`server->session` still at `NULL`). This might happen when labwc runs with headless outputs only which are immediately started by wlroots in contrast to DRM outputs. I assume a simple change to `if (output->server->session && !output->server->session->active)` is enough here.

We will also need to raise the seatd requirement to `>= 0.8.0` if we merge this to prevent labwc suddenly locking up.

**Comment by Consolatis** [comment]
Just asked in the wlroots channel regarding the unreliable `session->active` state and it turns out on affected versions of libseat this may be `true` instead of `false`. So on these affected versions the worst thing that will happen is that we spend some CPU as we do without this PR.

```
<Consolatis> kennylevinsen: soreau: quick question regarding libseat: https://github.com/WayfireWM/wayfire/pull/2484 mentions issues that the VT state (session->active) was not always reliable. Is this only an issue between "logind: Skip session activation" and "libseat/logind: Check initial active state" and thus all released versions are reliable or are there indeed releases where that state is unreliable?
<soreau> Consolatis: from my testing, AFAIK, latest release should have the fix but I have experienced the bug in past versions
<Consolatis> labwc is considering adding the same but suddenly stopping to render feels worse than what this fixes. So if there are affected versions out there (>= 0.2.0 as that is what wlroots itself requires) we'd have to ensure that we don't run that check if on such a version
<soreau> Consolatis: I don't think that consideration is necessary, because in the broken case, the active state variable is true during tty switch
<Consolatis> oh, I see. so it keeps wasting CPU but not stop rendering. Thanks!
<soreau> so it would be the same is without having the check, unless it works (which it does intermittently)
```

So with this: LGTM and thanks for the PR!

PR #2251, opened 2024-10-20
menu: fix the calculation for centered titles

tokyo4j:
Before:

20241020_15h22m41s_grim

After:

20241020_15h22m32s_grim


johanmalm:
Thanks.

PR #2252, opened 2024-10-20
menu: fix memory leak in update_client_send_to_menu()

tokyo4j:


johanmalm:
Thanks. Can't believe that one slipped in! Must have had an unfocused moment.

Issue #2163, opened 2024-09-20
Should the mouse pointer relocate to the location of a touch event on a touchscreen?


spl237:
A while back I raised this issue - https://github.com/labwc/labwc/issues/1566 - whereby touching a window headerbar didn't bring the window to the front, which was fixed by forcing a simulated mouse click on the touch location.

This issue is related. On X, if you use a touchscreen and you touch somewhere on it, the mouse pointer moves to that location. This doesn't happen on labwc (or indeed on wayfire or sway); the touch events are processed completely independently of mouse pointer location.

I am assuming this is deliberate? But is there a reason for it? 

The reason I ask is that I am trying to find a way to be able to simulate right-clicks on a touchscreen, and have come across this - https://github.com/bareboat-necessities/evdev-right-click-emulation/tree/hold-right-click, which generates right-click events at the mouse pointer location when the touchscreen is held for a second. But this only works correctly if touching the screen also moves the mouse pointer to the location that was touched - so while it could in theory work on wayland and provide an easy fix for the right-click problem, at present it does not.

Would it be possible to add behaviour whereby the mouse pointer does move to the touched location? Possible even as a configuration option if it is felt to be undesirable all the time?

johanmalm:
> This issue is related. On X, if you use a touchscreen and you touch somewhere on it, the mouse pointer moves to that location. This doesn't happen on labwc (or indeed on wayfire or sway); the touch events are processed completely independently of mouse pointer location.
> 
> I am assuming this is deliberate? But is there a reason for it?

I don't know and I'm not aware that it's deliberate.

I suspect that not many `labwc` users actually use "touch" - and looking at `src/input/touch.c` it has a rather short history which sort of backs up that supposition.

> Would it be possible to add behaviour whereby the mouse pointer does move to the touched location?

Yes, feels like we ought to do that.

> Possible even as a configuration option if it is felt to be undesirable all the time?

If that's what X does, then I think it's a reasonable starting point.

We generally try pretty hard to not surprise users by introducing changes to the UX, but if we go on the assumption that not many people use it, we could try some changes and see what responses we get... and then decide on the extent to which we make this configurable.

I've not got access to a touch screen with `labwc`, but could experiment on a Chromebook to see what 'normal' behaviours are.

PR #2253, opened 2024-10-20
Add mouseEmulation parameter to touch devices

spl237:
As per discussion on https://github.com/labwc/labwc/issues/2242, this PR adds the same mouseEmulation parameter to touch devices as already exists for tablet devices, and sets the device capabilties for the seat accordingly.

I have no idea if this is the best solution, or even the right thing to do, but it seems to work for me!


tokyo4j:
I like this addition.

For consistency, I suggest adding an early return in `touch_get_coords()` rather than removing the capability, just like what we're doing for tablets in `tablet_get_coords()`:

https://github.com/labwc/labwc/blob/c6a6be8219fbeaf99e035ef30a874bbf93d8cf1b/src/input/tablet.c#L134-L137

The difference from your solution would be whether the application know the seat has touchsceen/tablets, and I'm not sure which one is better. But I think we should keep consistency as long as there's no specific reason.

spl237:
Yes, just tried that, and it works just as well - will modify the PR accordingly. Many thanks for the suggestion - you know far more about this stuff than I do!

johanmalm:
Seems fine I think. Will hang fire until after the release of `0.8.1` this week.

@jp7677  - as main contributor of emulation stuff, would you have time to cast your eyes over this one?

spl237:
I should add - this also addresses the issue I raised in https://github.com/labwc/labwc/issues/2163

jp7677:
> I should add - this also addresses the issue I raised in #2163

I assume only with mouse emulation? With actual touch support it becomes more difficult, I started to add some thoughts here https://github.com/labwc/labwc/issues/2242#issuecomment-2417932320

src/input/touch.c
```diff
@@ -24,6 +24,13 @@ static struct wlr_surface*
 touch_get_coords(struct seat *seat, struct wlr_touch *touch, double x, double y,
 		double *x_offset, double *y_offset)
 {
+	/* Fall out if mouse emulation is set */
```
jp7677, line 27: Suggestion: be a bit more verbose with this comment since why this enforces mouse emulation is not directly obvious. I did the same just now for tablets (https://github.com/labwc/labwc/pull/2255) ;)

jp7677:
LGTM (but not tested)! Nice to keep this consistent with the tablet implementation. 
Added a small suggestion.

spl237:
> > I should add - this also addresses the issue I raised in #2163
> 
> I assume only with mouse emulation? With actual touch support it becomes more difficult, I started to add some thoughts here [#2242 (comment)](https://github.com/labwc/labwc/issues/2242#issuecomment-2417932320)

If mouse emulation is enabled for touch screens, then the mouse pointer moves to the touched point, which is what I was trying to get from that issue.

src/input/touch.c
spl237: Fine - will look at your explanation and produce something similar!

jp7677:
> > > I should add - this also addresses the issue I raised in #2163
> > 
> > 
> > I assume only with mouse emulation? With actual touch support it becomes more difficult, I started to add some thoughts here [#2242 (comment)](https://github.com/labwc/labwc/issues/2242#issuecomment-2417932320)
> 
> If mouse emulation is enabled for touch screens, then the mouse pointer moves to the touched point, which is what I was trying to get from that issue.

Thanks, yes, then we are on the same page ;). We should address this also for native touch support, not just for mouse emulation. But as stated, there is more to it and requires some thoughts.

spl237:
I think that on Pi, for the time being, we will be enabling mouse emulation by default for any touchscreens we detect, as it makes a bunch of stuff just work. If there are future improvements to native touch support that make this unnecessary, we can always turn it off again...

One thing which, as far as I can tell, is not supported in general on touch displays is the concept of double-clicking; there is no double-tap event that I know of, so anything which requires double-click (as our file manager does) will either need to use mouse emulation, or have some form of custom gesture added. For now, mouse emulation is just easier!

jp7677:
> One thing which, as far as I can tell, is not supported in general on touch displays is the concept of double-clicking; there is no double-tap event that I know of, so anything which requires double-click (as our file manager does) will either need to use mouse emulation, or have some form of custom gesture added. For now, mouse emulation is just easier!

Yes. From understanding, it depends heavily on how clients have implemented the touch protocol. Similar with the tablet protocol, results can be very different depending on toolkit and application itself.

src/input/touch.c
spl237: Done...

src/input/touch.c
```diff
@@ -24,6 +24,16 @@ static struct wlr_surface*
 touch_get_coords(struct seat *seat, struct wlr_touch *touch, double x, double y,
 		double *x_offset, double *y_offset)
 {
+	/* Do not return a surface when mouse emulation is enforced. Not
```
jp7677, line 27: nit: First comment line in a c style multi line comment should be empty, see e.g. https://github.com/labwc/labwc/blob/master/src/server.c#L233-L239

jp7677 (approved):
Not tested, but all good code-wise!

johanmalm:
LGTM.
Can't merge now has it has conflicts.
@spl237 Would you be happy to resolve the conflicts?

spl237:
Yes, I’ll sort that. Give me a day or so - up to my ears in release hell at the moment, but will do as soon as that’s a bit calmer.

jp7677:
> LGTM. Can't merge now has it has conflicts.

Actually, i don't see conflicts here (at least not in the GitHub UI) ...

<details>
  <summary>GitHub UI</summary>

image

</details>

spl237:
Yes - I see the same, so it ought to just merge?

spl237:
Closed and replaced by clean version at https://github.com/labwc/labwc/pull/2277

PR #2255, opened 2024-10-20
input: add comments about tablet mouse emulation

jp7677:
This mechanism is only indirect, so it deserves a comment.


johanmalm:
I'm not sure what `a.o.` is.

johanmalm:
Thanks.

PR #2257, opened 2024-10-21
src/main.c: add environment variable LABWC_VER

01micko:
src/config/session.c add env var LABWC_VER
This could be useful going forward for helper programs such as `labwc-menu-generator` or `labwc-tweaks`

### DIFF
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -29,6 +29,7 @@ static const char *const env_vars[] = {
 	"XCURSOR_THEME",
 	"XDG_SESSION_TYPE",
 	"LABWC_PID",
+	"LABWC_VER",
 	NULL
 };
 
--- a/src/main.c
+++ b/src/main.c
@@ -191,6 +191,13 @@ main(int argc, char *argv[])
 		wlr_log(WLR_DEBUG, "LABWC_PID=%s", pid);
 	}
 
+	/* useful for helper programs */
+	if (setenv("LABWC_VER", LABWC_VERSION, true) < 0) {
+		wlr_log_errno(WLR_ERROR, "unable to set LABWC_VER");
+	} else {
+		wlr_log(WLR_DEBUG, "LABWC_VER=%s", LABWC_VERSION);
+	}
+
 	if (!getenv("XDG_RUNTIME_DIR")) {
 		wlr_log(WLR_ERROR, "XDG_RUNTIME_DIR is unset");
 		exit(EXIT_FAILURE);


domo141:
good idea, I'd name it `LABWC_VERSION`, but there may be reasons not to.

currently one can run `grep -aom1 'labwc [0-9][--y]*' /proc/$LABWC_PID/exe` (on linux) to see version but having it in
environment is more portable for sure...

one thing to notice: when not doing full rebuild after change in git repo, that version string will not
be up-to-date -- that is just a minor nuisance for those who build without cleaning after pulling new changes
(or committing oneself)

01micko:
> good idea, I'd name it `LABWC_VERSION`, but there may be reasons not to.
> 
Main reason I didn't was to save confusion, the other is less typing! Still intuitive enough I think.

> currently one can run `grep -aom1 'labwc [0-9][--y]*' /proc/$LABWC_PID/exe` (on linux) to see version but having it in environment is more portable for sure...
> 

Nice, that works well, almost makes this PR redundant except that that would be more cumbersome in a `C` program.

> one thing to notice: when not doing full rebuild after change in git repo, that version string will not be up-to-date -- that is just a minor nuisance for those who build without cleaning after pulling new changes (or committing oneself)

I did notice that but tested by patching `master`.

domo141:
now that I looked `git grep | grep LABWC` -- `LABWC_VER` would list nicely with `LABWC_PID` ...

01micko:
The env var does make things so easy!

<details><summary>testver.c</summary>

```c
#include <stdlib.h>
#include <stdio.h>

struct lab_ver {
	unsigned maj;
	unsigned min;
	unsigned pat;
};

struct lab_ver get_lab_ver(char *ver) {
	struct lab_ver s;
	unsigned major = 0, minor = 0, bugfix = 0;
	sscanf(ver, "%u.%u.%u", &major, &minor, &patch);
	s.maj = major;
	s.min = minor;
	s.pat = bugfix;
	return s;
}

int main() {
	char *labwc_is_version = getenv("LABWC_VER");
	struct lab_ver p = get_lab_ver(labwc_is_version);
	char *labwc_is_running = getenv("LABWC_PID");
	if (labwc_is_running) {
		printf("labwc %u.%u.%u, pid:%s\n", p.maj, p.min, p.pat, labwc_is_running);
		return 0;
	} else {
		return 1;
	}
	// test with  LABWC_VER=0.8.0-132-5f8e8a7-dirty ./labwc-test (or whatever)
}
```
</details>

johanmalm:
I'm good with this, but let's merge after the release.

01micko:
Absolutely :smile:

01micko:
Thanks!

PR #2259, opened 2024-10-21
session-lock: make session_lock_output_create() safe to call twice

jlindgren90:
session_lock_output_create() can safely no-op if the lock output has already been created for the specified output. This scenario doesn't happen currently, and the change is in preparation for some other output-related changes I am working on. But I think it's a nice code improvement worth merging separately.

For after 0.8.1.


johanmalm:
Thanks.

PR #2260, opened 2024-10-21
output: remove ported wlr_output functions

jlindgren90:
We have several wlr_output_* functions which are just wrappers around corresponding wlr_output_state_* functions and don't actually touch the wlr_output itself. These probably made some sense historically, but IMHO they are just confusing now. So remove them and call wlr_output_state_* directly.

Rename wlr_output_commit() (still useful) to output_state_commit().

For after 0.8.1.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -549,7 +549,7 @@ struct wlr_box output_usable_area_in_layout_coords(struct output *output);
 struct wlr_box output_usable_area_scaled(struct output *output);
 void handle_output_power_manager_set_mode(struct wl_listener *listener,
 	void *data);
-void output_enable_adaptive_sync(struct wlr_output *output, bool enabled);
+void output_enable_adaptive_sync(struct output *output, bool enabled);
 
 /**
  * output_max_scale() - get maximum scale factor of all usable outputs.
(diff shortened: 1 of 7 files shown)


Consolatis:
> We have several wlr_output_* functions which are just wrappers around corresponding wlr_output_state_* functions and don't actually touch the wlr_output itself. These probably made some sense historically,

They were added to keep the rebase conflicts low during the lifetime of the 0.18 chase PR.

> but IMHO they are just confusing now. So remove them and call wlr_output_state_* directly.

Sounds good.

jlindgren90:
> We could potentially remove the whole file and move the remaining functions to `src/output.c`.

Yes, I like this idea. It also reminds me that I would like to create an `output.h` rather than having all the `output` functions in `labwc.h`.

How about leaving the functions in `output-state.c/h` for now and leaving the move for a follow-up PR?

Consolatis:
> It also reminds me that I would like to create an `output.h` rather than having all the `output` functions in `labwc.h`.

Anything that shrinks `labwc.h` is a good thing in my book. +1 for that idea.

> How about leaving the functions in `output-state.c/h` for now and leaving the move for a follow-up PR?

Sure, just wanted to mention the possibility.

johanmalm:
LGTM

PR #2263, opened 2024-10-23
menu: fix UAF in client-list-combined-menu after window destruction

tokyo4j:
client-list-combined-menu should be updated when a window is destroyed, otherwise we get SEGFAULT when selecting an old window entry in it (https://github.com/labwc/labwc/pull/2101#discussion_r1727053580).

In my opinion, we should merge fix this before the next release as we should always try not to introduce bugs that trigger SEGFAULT.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2473,5 +2473,6 @@ view_destroy(struct view *view)
 	wl_list_remove(&view->link);
 	free(view);
 
+	update_client_list_combined_menu(server);
 	cursor_update_focus(server);
 }


tokyo4j:
I initially added `menu_on_view_destroy()` that included the check for `server->menu_current->triggered_by_view`, but removed it to make this PR least impactful.

droc12345:
The problem is not that the client list menu needs updating.

an app is started using the example from #2101, while it's sleeping show client list menu is triggered, sleep finishes, then action is selected from menu. 
Once the menu is displayed, updating the menu should have no effect until the next time it's triggered.
I'm not sure your code will keep all segfaults from happening, for the reasons below.

The real problem is this, a pointer to a "view" is kept, pointing to the application.
When selected FOCUS is called on that "view" the action is triggered.
The 1st thing focus does is check if the pointer to the view is null ie if (view) do some action
It is not checking if the view is valid, only that it doesn't point to a NULL 
This type of "checking" is through pretty much all the code.

The next action after focus is raise, same problem, same underlying cause. 

All views should be checked for validity, not just nullness. 

I've been working on other things so haven't been keeping up with labwc, 
and I just happened to come by and see this thread, so I offer my views.
Have fun, enjoy life.

tokyo4j:
Sorry, I don't get what you mean...

> The real problem is this, a pointer to a "view" is kept, pointing to the application.

Which pointer do you mean? `item->client_list_view`?

> All views should be checked for validity, not just nullness.

What's "validity"? If it means whether the variable is `free()`ed or not, I don't think there's a way to check the validity of variables without modifying the compiler or libc's allocator, because the allocator may reuse the memory for `free()`ed `view` for another newly-allocated `view`, which is valid.

BTW, I realized my fix is incomplete, because `update_client_list_combined_menu()` doesn't touch `menu->selection.item`
and UAF can still happen when the selected item is updated in `menu_item_select()`. I'll try to come up with a better solution that can fit in the scope of the next release.

johanmalm:
Could do something like:

```
index bcb86ea..9a03119 100644
--- a/src/view.c
+++ b/src/view.c
@@ -2453,10 +2453,18 @@ view_destroy(struct view *view)
 		}
 	}
 
-	/* If we spawned a window menu, close it */
-	if (server->menu_current
-			&& server->menu_current->triggered_by_view == view) {
-		menu_close_root(server);
+	if (server->menu_current) {
+		/* If a window menu is open, close it */
+		if (server->menu_current->triggered_by_view == view) {
+			menu_close_root(server);
+		}
+
+		/* Also close client-list-combined-menu if open */
+		struct menu *client_list_combined_menu =
+			menu_get_by_id(server, "client-list-combined-menu");
+		if (server->menu_current == client_list_combined_menu) {
+			menu_close_root(server);
+		}
 	}
 
 	/*
```

johanmalm:
Like this?? #2264

droc12345:
Let me try and explain it better.

foot starts, with 5 second delay until it dies
2 seconds later, client list menu is activated,

scenario 1
2 seconds later, selection is made (should be no problem, focus/raise should happen then window deletion)

scenario 2
5 seconds later, selection is made (should be no problem, as focus/raise should see pointer to view being null, since window destruction happened a few seconds ago.

scenario 3
~3 seconds later selection is made (possible segfault at this point as window is being destroyed at same time it's trying to be accessed, and with timing focus/raise might not see the pointer as null and thus try and execute code it shouldn't) 

A potential fix is to have a semaphore that mediates between selection of windows and destruction of windows.

2nd thing, update_client_list is separate from the menu that is displayed on the screen.
In other words, if client list menu is activated and displaying, running update_client_list will have no effect on what is displayed, ie the menu is not redrawn.

Consolatis:
> scenario 2
> 5 seconds later, selection is made (should be no problem, as focus/raise should see pointer to view being null, since window destruction happened a few seconds ago.

I don't see where `client_list_view` is ever reset to `NULL`. Thus the UAF.

What we need at a minimum is something like this (pseudo code, not tested)
```c
void
menu_on_view_destroy(struct view *view) {
	struct menu *menu = menu_get_by_id(server, "client-list-combined-menu");
	if (!menu) {
		return;
	}
	struct menuitem *item;
	wl_list_for_each(item, &menu->menuitems, link) {
		if (item->client_list_view == view) {
			item->client_list_view = NULL;
		}
	}
}
```
And potentially a slight modification for `menu_execute_item`.

A proper solution would involve destroying that item and then recalculating the positions and selection status of the menu items so the experience ends up being the same as for the OSD where destroying a view while its open should work in all cases and just remove it from the list (even if its currently selected).

The workaround suggested by malm (just closing the menu if any view is destroyed) should also work although the experience for the user might be slightly confusing.

tokyo4j:
> scenario 3
~3 seconds later selection is made (possible segfault at this point as window is being destroyed at same time it's trying to be accessed, and with timing focus/raise might not see the pointer as null and thus try and execute code it shouldn't)

> A potential fix is to have a semaphore that mediates between selection of windows and destruction of windows.

Perhaps you are missing that labwc is single-threaded? Your statement would make sense if labwc were to process input events and client state updates with multiple threads in parallel.

johanmalm:
Covered by #2264

PR #2264, opened 2024-10-23
menu: fix UAF in client-list-combined-menu after window destruction

johanmalm:
Update client-list-combined-menu when a window is destroyed to avoid SEGFAULT when selecting an that window entry in it.

### DIFF
--- a/include/menu/menu.h
+++ b/include/menu/menu.h
@@ -85,6 +85,7 @@ bool menu_call_selected_actions(struct server *server);
 
 void menu_init(struct server *server);
 void menu_finish(struct server *server);
+void menu_on_view_destroy(struct view *view);
 
 /**
  * menu_get_by_id - get menu by id
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1190,6 +1190,36 @@ menu_finish(struct server *server)
 	current_menu = NULL;
 }
 
+void
+menu_on_view_destroy(struct view *view)
+{
+	struct server *server = view->server;
+
+	/* If the view being destroy has an open window menu, then close it */
+	if (server->menu_current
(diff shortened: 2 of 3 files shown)


johanmalm:
See #2263

tokyo4j:
I like this approach, but I suggest adding `menu_on_destroy()` like:

```c
void
menu_on_view_destroy(struct view *view)
{
	struct server *server = view->server;

	/* If we spawned a window menu, close it */
	if (server->menu_current
			&& server->menu_current->triggered_by_view == view) {
		menu_close_root(server);
	}

	struct menu *combined_menu =
		menu_get_by_id(server, "client-list-combined_menu");
	if (combined_menu) {
		struct menuitem *item;
		wl_list_for_each(item, &combined_menu->menuitems, link) {
			if (item->client_list_view == view) {
				item->client_list_view = NULL;
			}
		}
	}
}
```

The reason I removed `menu_on_view_destroy()` in #2263 is that I didn't want to make side effects on the existing lines:

https://github.com/labwc/labwc/blob/2b1a6941545b2d55ddf901c1358f9580a1018d79/src/view.c#L2456-L2460

In my initial approach in #2263, `menu_on_view_destroy()` called `update_client_list_combined_menu()` which iterates over `server->views`, so I had to move `menu_on_view_destroy()` to after `wl_list_remove(&view->link)`, which made me feel a bit anxious.

Now that your approach has minimal impact on the existing control flow, I think adding `menu_on_destroy()` is much cleaner.

johanmalm:
Updated as per discussion.
Have tested and seems to work fine.
Ready to merge from my perspective.

Consolatis:
With comment added as discussed via IRC: LGTM

PR #2266, opened 2024-10-24
Add final check for matching partial strings when looking for icons

spl237:
This adds a final check to the fuzzy matching for application icons, whereby the number of characters in whichever is shorter of the desktop file name and the app id is compared.

I have already implemented heuristics similar to this in my desktop panel application to get icons for the task switcher - https://github.com/raspberrypi-ui/wf-panel-pi/blob/efafe5800adbf76869c8b7f1e03b6cf6dd4d4622/src/widgets/window-list/toplevel.cpp#L701 - and this final test matches a few common applications which are otherwise missed, such as GIMP.

### DIFF
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -245,7 +245,7 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		int alen = strlen(app_id);
 		int dlen = strlen(desktop_id_base);
 
-		if (!strncmp(app_id, desktop_id, alen > dlen ? dlen : alen)) {
+		if (!strncasecmp(app_id, desktop_id, alen > dlen ? dlen : alen)) {
 			return entry;
 		}
 	}
--- a/src/icon-loader.c
+++ b/src/icon-loader.c
@@ -236,6 +236,20 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		}
 	}
 
+	/* Try matching partial strings - catches GIMP, among others */
+	for (size_t i = 0; i < n_entries; i++) {
+		struct sfdo_desktop_entry *entry = entries[i];
+		const char *desktop_id = sfdo_desktop_entry_get_id(entry, NULL);
+		const char *dot = strrchr(desktop_id, '.');
+		const char *desktop_id_base = dot ? (dot + 1) : desktop_id;
+		int alen = strlen(app_id);
+		int dlen = strlen(desktop_id_base);
+
+		if (!strncmp(app_id, desktop_id, alen > dlen ? dlen : alen)) {
+			return entry;
+		}
+	}
+
 	return NULL;
 }
 


johanmalm:
Thanks. 
Out of curiosity, what are the partial strings we're trying to compare?
Gimp works on my current setup (Arch).
Should we use `strncasecmp()`. Not that wf-panel-pi uses `toLower`

spl237:
strncasecmp is a good idea - I'll change that.

On my current Debian bookworm install, GIMP's app-id is "gimp-2.10", while the desktop file is "gimp.desktop", so I don't get an icon for GIMP without this change.

jlindgren90:
GIMP should be matched by the StartupWMClass as was discussed in a previous PR (don't have the link right now). I wonder why that's not working for you, it works here.

I think I would prefer matching on the executable name as was previously proposed, rather than a substring of the desktop file name. It seems like the latter might result in some unintended/incorrect matches. For a (theoretical) example, some short app I'd like "bc" could match a completely different application like "bcompare".

johanmalm:
Okay, I'll set this to draft for the minute then so I don't merge it by mistake before the `labwc` version bump tomorrow.

spl237:
> GIMP should be matched by the StartupWMClass as was discussed in a previous PR (don't have the link right now). I wonder why that's not working for you, it works here.

OK, that's odd - I'll do some testing and see if I can figure that out.

> I think I would prefer matching on the executable name as was previously proposed, rather than a substring of the desktop file name. It seems like the latter might result in some unintended/incorrect matches. For a (theoretical) example, some short app id like "bc" could match a completely different application like "bcompare".

Bear in mind this is a fallback - it only happens if a "proper" match hasn't already been found. So it is unlikely to find incorrect matches of that sort, as most of those would already have been found as exact string matches. All this does is to slightly reduce the possibility of some applications having no icon at all by finding an icon which is at least likely to be correct.

spl237:
> GIMP should be matched by the StartupWMClass as was discussed in a previous PR (don't have the link right now). I wonder why that's not working for you, it works here.

I've just looked into this. On Debian bookworm, the .desktop file for the version of GIMP which is in the upstream repo does not define StartupWMClass, so this test simply returns a null string. (In fact, very few of the packages we install by default do seem to define this - Firefox and LibreOffice are the only ones which do.)

I'm not sure where the version of GIMP you are installing comes from? But for the one from Debian, that check doesn't help.

jlindgren90:
It's the Arch Linux package, same as johanmalm. I checked the packaging and Arch doesn't appear to be modifying the .desktop file, so I'm not sure why it's different in Debian (unless an older version?)

I think I would still prefer comparing Exec[0] rather than substring matching (and I still have that change in a branch somewhere if wanted), but it's not a big deal one way or the other.

spl237:
> I think I would still prefer comparing Exec[0] rather than substring matching (and I still have that change in a branch somewhere if wanted), but it's not a big deal one way or the other.

I don’t think that will help in the case of GIMP - the name of the executable is the same as the app_id, gimp-2.10, so it still requires a substring match.

jlindgren90:
> the name of the executable is the same as the app_id

That's the point. I'm suggesting matching the name of the executable to the app_id.

jlindgren90:
Note that for X11 apps, the "app_id" is actually the "instance name" from the WM_CLASS property. In most cases (as with GIMP) this is exactly "the trailing part of the name used to invoke the program (argv[0] stripped of any directory names)". Which is what I propose matching on.

https://tronche.com/gui/x/icccm/sec-4.html#WM_CLASS

01micko:
@jlindgren90 and @spl237 this is only for context, I'm currently on debian/testing. I don't know if it helps or not but I am following this discussion.

```
mick@dellhome:~$ type gimp
gimp is /usr/bin/gimp
mick@dellhome:~$ readlink /usr/bin/gimp
gimp-2.10
mick@dellhome:~$ type gimp-2.10
gimp-2.10 is /usr/bin/gimp-2.10
mick@dellhome:~$ grep 'gimp' /usr/share/applications/*.desktop
/usr/share/applications/gimp.desktop:Exec=gimp-2.10 %U
/usr/share/applications/gimp.desktop:TryExec=gimp-2.10
/usr/share/applications/gimp.desktop:Icon=gimp
/usr/share/applications/gimp.desktop:MimeType=image/bmp;image/g3fax;image/gif;image/x-fits;image/x-pcx;image/x-portable-anymap;image/x-portable-bitmap;image/x-portable-graymap;image/x-portable-pixmap;image/x-psd;image/x-sgi;image/x-tga;image/x-xbitmap;image/x-xwindowdump;image/x-xcf;image/x-compressed-xcf;image/x-gimp-gbr;image/x-gimp-pat;image/x-gimp-gih;image/x-sun-raster;image/tiff;image/jpeg;image/x-psp;application/postscript;image/png;image/x-icon;image/x-xpixmap;image/x-exr;image/webp;image/x-webp;image/heif;image/heic;image/avif;image/jxl;image/svg+xml;application/pdf;image/x-wmf;image/jp2;image/x-xcursor;
/usr/share/applications/gimp.desktop:StartupWMClass=gimp-2.10
```

spl237:
> > the name of the executable is the same as the app_id
> 
> That's the point. I'm suggesting matching the name of the executable to the app_id.

Ah - my apologies; I misunderstood. You are proposing matching the executable name with the Exec lines in the .desktop files? That would indeed be a better solution if there is a suitable way of accessing the Exec lines.

Hopefully, this is all really only temporary stuff until something like https://wayland.app/protocols/xdg-toplevel-icon-v1 can be implemented. Assuming that does what it looks like it should!

jlindgren90:
See also https://gitlab.freedesktop.org/vyivel/libsfdo/-/merge_requests/5. It was closed because we thought StartupWMClass was a better solution, but maybe we should resurrect it.

jlindgren90:
Here is the previous PR that would match based on app_id == Exec[0]:
- https://github.com/labwc/labwc/pull/2215

Downside is, it depends on a private sfdo header and so does not work with a separately installed copy of the lib. Hence the libsfdo MR linked above. So, I am also okay with the substring-matching if it is the most pragmatic solution for now.

PR #2273, opened 2024-10-27
feat: hide cursor for touch and prevent pointer cursor interference

jp7677:
The last commit is work in progress, I'm Just opening this for visibility and early feedback.

This needs to be based on https://github.com/labwc/labwc/pull/2253 . Once that one is in, I'll probably open a separate PR with only the first two commits, which are a follow up from https://github.com/labwc/labwc/issues/2242#issuecomment-2417932320. Well, after some testing…

About the WIP commit, hiding the cursor for touch feels very natural for me (Sway does that as well, see https://github.com/swaywm/sway/blob/master/sway/input/cursor.c#L408), but unfortunately it is slightly more work to keep the cursor hidden and affects other input handlers.
Not sure though if the cursor should still be hidden when cursor emulation for touch is enabled, I tend to keep the cursor visible in that case. Any thoughts on this direction?

cc @spl237

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -146,6 +146,14 @@ request_cursor_notify(struct wl_listener *listener, void *data)
 	}
 
 	/*
+	 * Omit cursor notifications when the current cursor is
+	 * invisible, e.g. on touch input.
+	 */
+	if (!seat->cursor_visible) {
+		return;
+	}
+
+	/*
 	 * Omit cursor notifications from a pointer when a tablet
 	 * tool (stylus/pen) is in proximity. We expect to get cursor
 	 * notifications from the tablet tool instead.
@@ -193,6 +201,14 @@ request_set_shape_notify(struct wl_listener *listener, void *data)
 
 	/* Prevent setting a cursor image when moving or resizing */
 	if (seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
+		return;
+	}
+
+	/*
+	 * Omit set shape when the current cursor is
+	 * invisible, e.g. on touch input.
+	 */
+	if (!seat->cursor_visible) {
 		return;
 	}
 
@@ -347,15 +363,34 @@ cursor_set(struct seat *seat, enum lab_cursors cursor)
 		return;
 	}
 
-	wlr_cursor_set_xcursor(seat->cursor, seat->xcursor_manager,
-		cursor_names[cursor]);
+	if (seat->cursor_visible) {
+		wlr_cursor_set_xcursor(seat->cursor, seat->xcursor_manager,
+			cursor_names[cursor]);
+	}
 	seat->server_cursor = cursor;
+}
+
+void
+cursor_set_visible(struct seat *seat, bool visible)
+{
+	if (seat->cursor_visible == visible) {
+		return;
+	}
+
+	seat->cursor_visible = visible;
+	cursor_update_image(seat);
 }
 
 void
 cursor_update_image(struct seat *seat)
 {
 	enum lab_cursors cursor = seat->server_cursor;
+
+	if (!seat->cursor_visible) {
+		wlr_cursor_unset_image(seat->cursor);
+		return;
+	}
+
 	if (cursor == LAB_CURSOR_CLIENT) {
 		/*
 		 * When we loose the output cursor while over a client
@@ -817,6 +852,7 @@ cursor_motion(struct wl_listener *listener, void *data)
 	struct server *server = seat->server;
 	struct wlr_pointer_motion_event *event = data;
 	idle_manager_notify_activity(seat->seat);
+	cursor_set_visible(seat, /* visible */ true);
 
 	wlr_relative_pointer_manager_v1_send_relative_motion(
 		server->relative_pointer_manager,
@@ -843,6 +879,7 @@ cursor_motion_absolute(struct wl_listener *listener, void *data)
 		listener, seat, cursor_motion_absolute);
 	struct wlr_pointer_motion_absolute_event *event = data;
 	idle_manager_notify_activity(seat->seat);
+	cursor_set_visible(seat, /* visible */ true);
 
 	double lx, ly;
 	wlr_cursor_absolute_to_layout_coords(seat->cursor,
@@ -1144,6 +1181,7 @@ cursor_button(struct wl_listener *listener, void *data)
 	struct seat *seat = wl_container_of(listener, seat, cursor_button);
 	struct wlr_pointer_button_event *event = data;
 	idle_manager_notify_activity(seat->seat);
+	cursor_set_visible(seat, /* visible */ true);
 
 	bool notify;
 	switch (event->state) {
@@ -1324,6 +1362,7 @@ cursor_axis(struct wl_listener *listener, void *data)
 
 	struct cursor_context ctx = get_cursor_context(server);
 	idle_manager_notify_activity(seat->seat);
+	cursor_set_visible(seat, /* visible */ true);
 
 	/* Bindings swallow mouse events if activated */
 	bool handled = handle_cursor_axis(server, &ctx, event);
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1194,8 +1194,6 @@ void
 cursor_emulate_move_absolute(struct seat *seat, struct wlr_input_device *device,
 		double x, double y, uint32_t time_msec)
 {
-	idle_manager_notify_activity(seat->seat);
-
 	double lx, ly;
 	wlr_cursor_absolute_to_layout_coords(seat->cursor,
 		device, x, y, &lx, &ly);
@@ -1210,8 +1208,6 @@ void
 cursor_emulate_button(struct seat *seat, uint32_t button,
 		enum wl_pointer_button_state state, uint32_t time_msec)
 {
-	idle_manager_notify_activity(seat->seat);
-
 	bool notify;
 	switch (state) {
 	case WL_POINTER_BUTTON_STATE_PRESSED:
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -61,6 +61,7 @@ handle_touch_motion(struct wl_listener *listener, void *data)
 {
 	struct seat *seat = wl_container_of(listener, seat, touch_motion);
 	struct wlr_touch_motion_event *event = data;
+
 	idle_manager_notify_activity(seat->seat);
 
 	int touch_point_count = wl_list_length(&seat->touch_points);
@@ -109,6 +110,8 @@ handle_touch_down(struct wl_listener *listener, void *data)
 {
 	struct seat *seat = wl_container_of(listener, seat, touch_down);
 	struct wlr_touch_down_event *event = data;
+
+	idle_manager_notify_activity(seat->seat);
 
 	/* Compute layout => surface offset and save for this touch point */
 	struct touch_point *touch_point = znew(*touch_point);
@@ -166,6 +169,8 @@ handle_touch_up(struct wl_listener *listener, void *data)
 {
 	struct seat *seat = wl_container_of(listener, seat, touch_up);
 	struct wlr_touch_up_event *event = data;
+
+	idle_manager_notify_activity(seat->seat);
 
 	/* Remove the touch point from the seat */
 	struct touch_point *touch_point, *tmp;
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -125,6 +125,9 @@ handle_touch_down(struct wl_listener *listener, void *data)
 	wl_list_insert(&seat->touch_points, &touch_point->link);
 	int touch_point_count = wl_list_length(&seat->touch_points);
 
+	/* hide the cursor when starting touch input */
+	cursor_set_visible(seat, /* visible */ false);
+
 	if (touch_point->surface) {
 		seat_pointer_end_grab(seat, touch_point->surface);
 		/* Clear focus to not interfere with touch input */
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -126,6 +126,7 @@ handle_touch_down(struct wl_listener *listener, void *data)
 	int touch_point_count = wl_list_length(&seat->touch_points);
 
 	if (touch_point->surface) {
+		seat_pointer_end_grab(seat, touch_point->surface);
 		/* Clear focus to not interfere with touch input */
 		wlr_seat_pointer_notify_clear_focus(seat->seat);
 
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -123,6 +123,9 @@ handle_touch_down(struct wl_listener *listener, void *data)
 	int touch_point_count = wl_list_length(&seat->touch_points);
 
 	if (touch_point->surface) {
+		/* Clear focus to not interfere with touch input */
+		wlr_seat_pointer_notify_clear_focus(seat->seat);
+
 		/* Convert coordinates: first [0, 1] => layout */
 		double lx, ly;
 		wlr_cursor_absolute_to_layout_coords(seat->cursor,
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -77,6 +77,8 @@ handle_touch_motion(struct wl_listener *listener, void *data)
 				double sx = lx - touch_point->x_offset;
 				double sy = ly - touch_point->y_offset;
 
+				wlr_cursor_warp_absolute(seat->cursor,
+					&event->touch->base, event->x, event->y);
 				wlr_seat_touch_notify_motion(seat->seat, event->time_msec,
 					event->touch_id, sx, sy);
 			} else {
@@ -133,6 +135,8 @@ handle_touch_down(struct wl_listener *listener, void *data)
 			}
 		}
 
+		wlr_cursor_warp_absolute(seat->cursor,
+			&event->touch->base, event->x, event->y);
 		wlr_seat_touch_notify_down(seat->seat, touch_point->surface,
 			event->time_msec, event->touch_id, sx, sy);
 	} else {
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -63,6 +63,8 @@ handle_touch_motion(struct wl_listener *listener, void *data)
 	struct wlr_touch_motion_event *event = data;
 	idle_manager_notify_activity(seat->seat);
 
+	int touch_point_count = wl_list_length(&seat->touch_points);
+
 	/* Find existing touch point to determine initial offsets to subtract */
 	struct touch_point *touch_point;
 	wl_list_for_each(touch_point, &seat->touch_points, link) {
@@ -77,13 +79,17 @@ handle_touch_motion(struct wl_listener *listener, void *data)
 				double sx = lx - touch_point->x_offset;
 				double sy = ly - touch_point->y_offset;
 
-				wlr_cursor_warp_absolute(seat->cursor,
-					&event->touch->base, event->x, event->y);
+				if (touch_point_count == 1) {
+					wlr_cursor_warp_absolute(seat->cursor, &event->touch->base,
+						event->x, event->y);
+				}
 				wlr_seat_touch_notify_motion(seat->seat, event->time_msec,
 					event->touch_id, sx, sy);
 			} else {
-				cursor_emulate_move_absolute(seat, &event->touch->base,
-					event->x, event->y, event->time_msec);
+				if (touch_point_count == 1) {
+					cursor_emulate_move_absolute(seat, &event->touch->base,
+						event->x, event->y, event->time_msec);
+				}
 			}
 			return;
 		}
@@ -114,6 +120,7 @@ handle_touch_down(struct wl_listener *listener, void *data)
 	touch_point->y_offset = y_offset;
 
 	wl_list_insert(&seat->touch_points, &touch_point->link);
+	int touch_point_count = wl_list_length(&seat->touch_points);
 
 	if (touch_point->surface) {
 		/* Convert coordinates: first [0, 1] => layout */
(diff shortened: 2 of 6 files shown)


johanmalm:
> hiding the cursor for touch feels very natural for me (Sway does that as well, see https://github.com/swaywm/sway/blob/master/sway/input/cursor.c#L408), but unfortunately it is slightly more work to keep the cursor hidden and affects other input handlers. 

Yeah, feels right to hide cursor on touch.

> Not sure though if the cursor should still be hidden when cursor emulation for touch is enabled, I tend to keep the cursor visible in that case. Any thoughts on this direction?

I think it'd be better to hide the cursor by default - even with cursor emulation. I think it would be less 'surprising' for users from other operating systems. Just my 2 pennies worth though... I'm not really a big users of touch screens and we can always make it configurable if there are many wishes here.

tokyo4j:
>> Not sure though if the cursor should still be hidden when cursor emulation for touch is enabled, I tend to keep the cursor visible in that case. Any thoughts on this direction?

> I think it'd be better to hide the cursor by default - even with cursor emulation. I think it would be less 'surprising' for users from other operating systems. Just my 2 pennies worth though... I'm not really a big users of touch screens and we can always make it configurable if there are many wishes here.

I agree with hiding cursor whenever the finger is down. I tested KWin, Mutter and sway and they all hide the cursor whether or not the finger is on client surfaces or compositor-side components.

jp7677:
This is ready to review and up for more testing/feedback. Note that I keep the cursor hidden not only during touch motion, but also after touch input until a mouse/pointer or tablet event is signaled.

Please take a good look at the last commit. I went for extending `lab_cursors` with `LAB_CURSOR_INVISIBLE`, this fits quite good, but not 100% because  being invisible is more of a sub state of the other cursors. Initially I went for a separate `bool cursor_hidden` which does the job as well. Not sure which of those two approaches is better. 

This PR should have no impact on non-touch usage.

jp7677:
Back to draft, found a case where this PR ended up loading an xdg/x11 NULL cursor which ended in a segfault.

jp7677:
> Back to draft, found a case (wpolm toggle with hidden cursor) where this PR ended up loading an xdg/x11 NULL cursor which ended in a segfault.

I've found the issue (`systemd-coredump` to the rescue), passing `NULL` into `wlr_cursor_set_xcursor` isn't a good idea, even unintentionally. I happy now with how this works, so up again for feedback, both UI- and code-wise, and testing. 

I haven't yet found a way to close a labwc menu (root or client) on touch down, at least not without pulling in all of the general cursor button functions (which comes with other side effects). I'm happy to address this as well, but since this is also an annoyance on current master, I guess doing this later is fine.

tokyo4j:
I tested this PR and it worked perfectly so far!

However, I'm not a fan of how `cursor_set_client_cursor()` doesn't set `LAB_CURSOR_CLIENT` when the cursor is hidden, and how `cursor_set(seat, LAB_CURSOR_NONE);` hides the cursor but `cursor_set(seat, LAB_CURSOR_DEFAULT);` doesn't show it again.

After some thoughts, I think adding `LAB_CURSOR_NONE` in `enum lab_cursors` is not a good idea because whether the cursor is hidden or not is a concept orthogonal to the cursor shape or whether the cursor server-side or client-side. I'm even not sure `LAB_CURSOR_CLIENT` should be in `enum lab_cursors`, but at the moment it'll be cleaner because switching to server-side cursor usually involves setting server-side cursor shape (e.g. when starting interactive move/resize).

As a reference, I made a patch based on this PR. I made some other changes like adding `cursor_set_visible(struct seat *seat, bool visible)` as that's how wlroots APIs are designed, but you don't need to follow all of it.

```diff
diff --git a/include/input/cursor.h b/include/input/cursor.h
index 12d5870a..b52d9218 100644
--- a/include/input/cursor.h
+++ b/include/input/cursor.h
@@ -16,7 +16,6 @@ enum wl_pointer_button_state;
 /* Cursors used internally by labwc */
 enum lab_cursors {
 	LAB_CURSOR_CLIENT = 0,
-	LAB_CURSOR_NONE,
 	LAB_CURSOR_DEFAULT,
 	LAB_CURSOR_GRAB,
 	LAB_CURSOR_RESIZE_NW,
@@ -73,12 +72,7 @@ struct cursor_context get_cursor_context(struct server *server);
  */
 void cursor_set(struct seat *seat, enum lab_cursors cursor);
 
-/**
- * cursor_ensure_visible - sets the cursor to the default
- * cursor if it was invisible, does nothing otherwise.
- * @seat - current seat
- */
-void cursor_ensure_visible(struct seat *seat);
+void cursor_set_visible(struct seat *seat, bool visible);
 
 /**
  * cursor_get_resize_edges - calculate resize edge based on cursor position
diff --git a/include/labwc.h b/include/labwc.h
index bcbfd195..e50bcc45 100644
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -108,6 +108,7 @@ struct seat {
 	 * cursor_set().  Cleared when a client surface is entered
 	 * (in that case the client is expected to set its own cursor image).
 	 */
+	bool cursor_visible;
 	enum lab_cursors server_cursor;
 	struct wlr_cursor *cursor;
 	struct wlr_xcursor_manager *xcursor_manager;
diff --git a/src/input/cursor.c b/src/input/cursor.c
index d4a2a783..5fdbcbf5 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -37,7 +37,6 @@ static const char * const *cursor_names = NULL;
 /* Usual cursor names */
 static const char * const cursors_xdg[] = {
 	NULL,
-	"",
 	"default",
 	"grab",
 	"nw-resize",
@@ -53,7 +52,6 @@ static const char * const cursors_xdg[] = {
 /* XCursor fallbacks */
 static const char * const cursors_x11[] = {
 	NULL,
-	"",
 	"left_ptr",
 	"grabbing",
 	"top_left_corner",
@@ -151,7 +149,7 @@ request_cursor_notify(struct wl_listener *listener, void *data)
 	 * Omit cursor notifications when the current cursor is
 	 * invisible, e.g. on touch input.
 	 */
-	if (seat->server_cursor == LAB_CURSOR_NONE) {
+	if (!seat->cursor_visible) {
 		return;
 	}
 
@@ -210,7 +208,7 @@ request_set_shape_notify(struct wl_listener *listener, void *data)
 	 * Omit set shape when the current cursor is
 	 * invisible, e.g. on touch input.
 	 */
-	if (seat->server_cursor == LAB_CURSOR_NONE) {
+	if (!seat->cursor_visible) {
 		return;
 	}
 
@@ -355,20 +353,6 @@ process_cursor_resize(struct server *server, uint32_t time)
 	}
 }
 
-static void
-cursor_set_client_cursor(struct seat *seat)
-{
-	/*
-	 * Keep the cursor invisible if it has been hidden due
-	 * to e.g. touch input.
-	 */
-	if (seat->server_cursor == LAB_CURSOR_NONE) {
-		return;
-	}
-
-	seat->server_cursor = LAB_CURSOR_CLIENT;
-}
-
 void
 cursor_set(struct seat *seat, enum lab_cursors cursor)
 {
@@ -379,28 +363,20 @@ cursor_set(struct seat *seat, enum lab_cursors cursor)
 	 * setting the cursor when the cursor is invisible, e.g.
 	 * on touch input.
 	 */
-	if (seat->server_cursor == cursor
-			|| seat->server_cursor == LAB_CURSOR_NONE) {
+	if (seat->server_cursor == cursor) {
 		return;
 	}
-
-	if (cursor == LAB_CURSOR_NONE) {
-		wlr_cursor_unset_image(seat->cursor);
-	} else {
-		wlr_cursor_set_xcursor(seat->cursor, seat->xcursor_manager,
-			cursor_names[cursor]);
-	}
 	seat->server_cursor = cursor;
+	cursor_update_image(seat);
 }
 
 void
-cursor_ensure_visible(struct seat *seat)
+cursor_set_visible(struct seat *seat, bool visible)
 {
-	if (seat->server_cursor != LAB_CURSOR_NONE) {
+	if (seat->cursor_visible == visible) {
 		return;
 	}
-
-	seat->server_cursor = LAB_CURSOR_DEFAULT;
+	seat->cursor_visible = visible;
 	cursor_update_image(seat);
 }
 
@@ -408,6 +384,12 @@ void
 cursor_update_image(struct seat *seat)
 {
 	enum lab_cursors cursor = seat->server_cursor;
+
+	if (!seat->cursor_visible) {
+		wlr_cursor_unset_image(seat->cursor);
+		return;
+	}
+
 	if (cursor == LAB_CURSOR_CLIENT) {
 		/*
 		 * When we loose the output cursor while over a client
@@ -561,7 +543,7 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 			}
 			wlr_seat_pointer_notify_enter(wlr_seat, ctx->surface,
 				ctx->sx, ctx->sy);
-			cursor_set_client_cursor(seat);
+			seat->server_cursor = LAB_CURSOR_CLIENT;
 		}
 		if (cursor_has_moved) {
 			*sx = ctx->sx;
@@ -869,7 +851,7 @@ cursor_motion(struct wl_listener *listener, void *data)
 	struct server *server = seat->server;
 	struct wlr_pointer_motion_event *event = data;
 	idle_manager_notify_activity(seat->seat);
-	cursor_ensure_visible(seat);
+	cursor_set_visible(seat, true);
 
 	wlr_relative_pointer_manager_v1_send_relative_motion(
 		server->relative_pointer_manager,
@@ -1197,7 +1179,7 @@ cursor_button(struct wl_listener *listener, void *data)
 	struct seat *seat = wl_container_of(listener, seat, cursor_button);
 	struct wlr_pointer_button_event *event = data;
 	idle_manager_notify_activity(seat->seat);
-	cursor_ensure_visible(seat);
+	cursor_set_visible(seat, true);
 
 	bool notify;
 	switch (event->state) {
@@ -1378,7 +1360,7 @@ cursor_axis(struct wl_listener *listener, void *data)
 
 	struct cursor_context ctx = get_cursor_context(server);
 	idle_manager_notify_activity(seat->seat);
-	cursor_ensure_visible(seat);
+	cursor_set_visible(seat, true);
 
 	/* Bindings swallow mouse events if activated */
 	bool handled = handle_cursor_axis(server, &ctx, event);
@@ -1475,6 +1457,7 @@ cursor_init(struct seat *seat)
 
 	/* Set the initial cursor image so the cursor is visible right away */
 	cursor_set(seat, LAB_CURSOR_DEFAULT);
+	cursor_set_visible(seat, true);
 
 	dnd_init(seat);
 
diff --git a/src/input/tablet.c b/src/input/tablet.c
index fba1bde5..889b603f 100644
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -251,7 +251,7 @@ handle_tablet_tool_proximity(struct wl_listener *listener, void *data)
 	}
 
 	idle_manager_notify_activity(tablet->seat->seat);
-	cursor_ensure_visible(tablet->seat);
+	cursor_set_visible(tablet->seat, true);
 
 	if (ev->state == WLR_TABLET_TOOL_PROXIMITY_IN) {
 		tablet->motion_mode =
@@ -314,7 +314,7 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 	}
 
 	idle_manager_notify_activity(tablet->seat->seat);
-	cursor_ensure_visible(tablet->seat);
+	cursor_set_visible(tablet->seat, true);
 
 	/*
 	 * Reset relative coordinates. If those axes aren't updated,
@@ -472,7 +472,7 @@ handle_tablet_tool_tip(struct wl_listener *listener, void *data)
 	}
 
 	idle_manager_notify_activity(tablet->seat->seat);
-	cursor_ensure_visible(tablet->seat);
+	cursor_set_visible(tablet->seat, true);
 
 	double x, y, dx, dy;
 	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
@@ -552,7 +552,7 @@ handle_tablet_tool_button(struct wl_listener *listener, void *data)
 	}
 
 	idle_manager_notify_activity(tablet->seat->seat);
-	cursor_ensure_visible(tablet->seat);
+	cursor_set_visible(tablet->seat, true);
 
 	double x, y, dx, dy;
 	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
diff --git a/src/input/touch.c b/src/input/touch.c
index 132f72b4..2d69814a 100644
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -119,7 +119,7 @@ handle_touch_down(struct wl_listener *listener, void *data)
 	wl_list_insert(&seat->touch_points, &touch_point->link);
 
 	/* hide the cursor when starting touch input */
-	cursor_set(seat, LAB_CURSOR_NONE);
+	cursor_set_visible(seat, false);
 
 	if (touch_point->surface) {
 		seat_pointer_end_grab(seat, touch_point->surface);

```

</details>

jp7677:
> I tested this PR and it worked perfectly so far!

Thanks, cool to read that!

> However, I'm not a fan of how `cursor_set_client_cursor()` doesn't set `LAB_CURSOR_CLIENT` when the cursor is hidden, and how `cursor_set(seat, LAB_CURSOR_NONE)` hides the cursor but `cursor_set(seat, LAB_CURSOR_DEFAULT)` doesn't show it again.

Yeah, I wasn't sure either of that approach. Thanks for your opinion and the provided alternative. I've changed the last commit to use your patch (which is actually pretty close to what I had initially).

src/input/cursor.c
```diff
@@ -347,15 +363,31 @@ cursor_set(struct seat *seat, enum lab_cursors cursor)
 		return;
 	}
 
-	wlr_cursor_set_xcursor(seat->cursor, seat->xcursor_manager,
-		cursor_names[cursor]);
 	seat->server_cursor = cursor;
+	cursor_update_image(seat);
```
jp7677, line 367: @tokyo4j I have taken this over from your suggestion. `cursor_update_image()` does more than just `wlr_cursor_set_xcursor()`. This is OK and not to expensive?

tokyo4j: Hmm... now I think we should keep it as it is.
I thought it would be cleaner if we could centralize the logic updating the cursor image into `cursor_update_image()` and made the change without much thought, but `cursor_update_image()` seems to be intended for refreshing the cursor on output layout/scale changes and indeed it does more than just setting the cursor image. Maybe we can refactor this later.

jp7677: Should be good now.

jp7677: PS: A side effect is now that the cursor is initially hidden, even without touch, and becomes visible with the first mouse action. I like it, but this affects all setups. To revert to the current behavior we just have to initialize `cursor_visible` with `true`.

src/input/touch.c
```diff
@@ -133,6 +145,8 @@ handle_touch_down(struct wl_listener *listener, void *data)
 			}
 		}
 
+		wlr_cursor_warp_absolute(seat->cursor,
```
jp7677, line 148: Note to self, look briefly if this behaves with multiple touch points, thus when more than one finger is down.

jp7677: Answer: the cursor does jump back and forth between two finger when moving both. But this is only visible when not hiding the cursor, so probably not an issue.

(I have a branch that adds a "hideCursor" option for touch/tablet for testing, not sure yet how useful this is).

AlessandroTischer:
Hi! I would really like to test this implementation, as I'm using a rpi4 with a touchscreen display and in the latest rpi OS they switched to labwc. 
Is it something difficult to test? If not, would you please tell me how? If it helps, I could give you some feedbacks

jp7677:
> Hi! I would really like to test this implementation, as I'm using a rpi4 with a touchscreen display and in the latest rpi OS they switched to labwc. Is it something difficult to test? If not, would you please tell me how? If it helps, I could give you some feedbacks

Thanks for offering to test this PR. This is certainly appreciated.
For testing you would have to checkout the code from this PR (https://github.com/jp7677/labwc/tree/touch) on your Pi, compile Labwc and then start it from a TTY. Are you familiar with compiling on your PI? 
Once you have everything running, it would be nice to play with and without mouse emulation for touch (https://labwc.github.io/labwc-config.5.html#touch).

AlessandroTischer:
> > Hi! I would really like to test this implementation, as I'm using a rpi4 with a touchscreen display and in the latest rpi OS they switched to labwc. Is it something difficult to test? If not, would you please tell me how? If it helps, I could give you some feedbacks
> 
> Thanks for offering to test this PR. This is certainly appreciated.
> For testing you would have to checkout the code from this PR (https://github.com/jp7677/labwc/tree/touch) on your Pi, compile Labwc and then start it from a TTY. Are you familiar with compiling on your PI? 

I am definitely not familiar unfortunately. I think I would need a step by step guide... 

> Once you have everything running, it would be nice to play with and without mouse emulation for touch (https://labwc.github.io/labwc-config.5.html#touch).

I can happily test it if useful, but I will definitely prefer the simple touch-only experience for my application.

jp7677:
> I am definitely not familiar unfortunately. I think I would need a step by step guide...

Well, sorry, I don't a watertight step-by-step guide at hand. I'm also not that familiar with Raspberry Pi OS to provide one in a few minutes. Essentially it comes to installing build tools and compiler (gcc, meson and friends), installing all dependencies from the `labwc` packages from the distribution (there is an `apt` commend that does that conveniently), checking out the code using `git` from above mentioned link and then follow https://github.com/labwc/labwc?tab=readme-ov-file#2-build-and-installation . Sorry for being a better assistant.

johanmalm:
Based on https://github.com/labwc/labwc/issues/2290#issuecomment-2453118145

(not tried)

Assume you already have git, gcc, meson and ninja, otherwise install those too.

```
sudo apt install hwdata libgbm-dev libdisplay-info-dev libseat-dev libinput-dev libpango1.0-dev libpangocairo-1.0-0 libcairo2-dev libglib2.0-dev libpixman-1-dev libxkbcommon-dev liblcms2-dev libxcb-xinput-dev libxcb-errors-dev libxcb-render-util0-dev libxcb-present-dev libxcb-res0-dev libxcb-dri3-dev libxcb-ewmh-dev libxcb-icccm4-dev libxcb-composite0-dev cmake libxml2-dev libliftoff-dev

git clone https://github.com/labwc/labwc
cd labwc
git clone https://gitlab.freedesktop.org/mesa/drm.git subprojects/libdrm
git clone https://gitlab.freedesktop.org/vyivel/libsfdo.git subprojects/libsfdo
git clone https://gitlab.freedesktop.org/wayland/wayland.git subprojects/wayland
git clone https://gitlab.freedesktop.org/wayland/wayland-protocols.git subprojects/wayland-protocols
meson setup build
meson compile -C build
./build/labwc
```

If you notice anything additional that's needed let me know. I ought to add this to the wiki really but it'd be good to validate first  :smile:

src/input/touch.c
jp7677: The last commit addresses this and warps the cursor only when exactly one finger is down.

carpajr:
Could someone guide me to where I should set the cursor to hide?

@johanmalm with your mini guide I was able to recompile quickly the code. Thanks!

@jp7677, I'm able to do some tests. Is there a critical aspect that I must consider when taking the test?

jp7677:
> @jp7677, I'm able to do some tests. Is there a critical aspect that I must consider when taking the test?

Very nice, thanks a lot. Nothing in particular, there shouldn't be a regression and no weird hover/side effects. It would be nice to play with and without mouse emulation. Application react differently, e.g. firefox allows zooming-in when pinching and when mouse emulation is off. Having mouse emulation on should play better with e.g. select or drag and drop in file managers.

jp7677:
PS: With Johans mini guide you'll build labwc from master. For building this PR you'll also need to add my fork as a remote and checkout the `touch` branch. From the top of my head:

```shell
git remote add jp7677 https://github.com/jp7677/labwc.git
git checkout jp7677/touch
```

carpajr:
@jp7677,
I'm using labwc in a raspberryPi 5 for a few days in a raspberryPi 5. 
I am sorry for the stupid question, but how do I hide the cursor? I've tried using rc.xml setting there 
```
<mouse>
  <cursor>none</cursor>
</mouse>
``` 
But it seems not work.

jp7677:
@carpajr nice to read that. The configuration option you mentioned doesn’t exist, you can remove that. This PR wil hide the cursor automatically for touch input. 
(Or to be exact: the cursor is initially hidden and only becomes visible on mouse or tablet input.

ameinhardt:
I'm showing a kiosk chrome dashboard on non-touch screen (TV). How could I hide it there?

carpajr:
@jp7677 thanks. I did many tests, and today I'll continue a little bit.

@ameinhardt just updates labwc with the PR and the cursor will be hidden.

AlessandroTischer:
> PS: With Johans mini guide you'll build labwc from master. For building this PR you'll also need to add my fork as a remote and checkout the `touch` branch. From the top of my head:
> 
> ```shell
> git remote add jp7677 https://github.com/jp7677/labwc.git
> ```

here you need to add `git fetch jp7677`.

> ```shell
> git checkout jp7677/touch
> ```

Other than that, it seems to work fine so far.

src/input/cursor.c
jp7677: `cursor_visible` is now initialized with `true` to keep the original behavior.

tokyo4j:
Thanks!

MathisP75:
I am on 0.8.3 and I still see the cursor on my touch screen. Has this been removed?

tokyo4j:
> I am on 0.8.3 and I still see the cursor on my touch screen. Has this been removed?

This hasn't been removed. Could you check if the mouse emulation for touch devices is enabled (`<touch mouseEmulation="yes" />`)?

MathisP75:
I tried with it on and off, it didn't seem to change anything.

tokyo4j:
That's strange. With this PR, the cursor should be hidden as soon as the finger touches the touchscreen.

Perhaps your touchscreen is detected as a pointer device instead of a touchscreen device. Could you run `libinput debug-events` from your terminal, touch the touchscreen with your finger and see if prints `TOUCH_DOWN` or `TOUCH_FRAME` events?

MathisP75:
It prints both...
```
mathis@raspberrypi5:~ $ libinput debug-events
-event4   DEVICE_ADDED            WaveShare WS170120                seat0 default group1  cap:t  size 205x150mm ntouches 5 calib
-event0   DEVICE_ADDED            vc4-hdmi-0                        seat0 default group2  cap:kp left scroll-nat scroll-button
-event2   DEVICE_ADDED            vc4-hdmi-1                        seat0 default group3  cap:kp left scroll-nat scroll-button
-event4   TOUCH_DOWN              +0.000s       0 (0) 44.39/77.04 (91.00/115.75mm)
 event4   TOUCH_FRAME             +0.000s
 event4   TOUCH_UP                +1.434s       0 (0)
 event4   TOUCH_FRAME             +1.434s
 event4   TOUCH_DOWN              +3.203s       0 (0)  4.10/ 7.82 ( 8.40/11.75mm)
 event4   TOUCH_FRAME             +3.203s
 event4   TOUCH_UP                +3.295s       0 (0)
 event4   TOUCH_FRAME             +3.295s
```

tokyo4j:
Then I have no idea why the cursor is not hidden on your system.

I built the latest labwc locally on raspberry pi 5 and confirmed the cursor is hidden as soon as I touch the touchscreen.

Issue #2271, opened 2024-10-26
Ctrl+click does not work in unfocused xdg-shell windows


jlindgren90:
When doing a Ctrl+click on an xdg-shell window that is not the active/focused window, it seems that the client doesn't see the Ctrl key state, and just handles it like a normal click. Example: Ctrl+click on a file in Thunar should add to the current selection, but instead selects only the single file clicked on. Both GTK and Qt-based apps are affected.

This is my current labwc config (I don't think there is anything in there that would affect this behavior):
https://github.com/jlindgren90/Config/blob/master/templates/labwc-rc.xml

I haven't yet started looking in the input and/or action code to see how to solve this.

jlindgren90:
It appears that Wayland mouse events don't include modifier state, and in fact clients aren't informed of modifier state at all until a (keyboard) enter event. So hooray for another basic functionality gap in the protocol, I guess?

One option (as a workaround) could be for Ctrl or Shift+click to match modifier-less button actions (like the default Focus+Raise) unless a modifier-specific action is found.

johanmalm:
IIRC the protocol is not clear on whether to broadcast modifiers or not.

I would quite like because all clients would then be aware of current keyboard layout.

I believe sway used to broadcast modifiers but chose not in order to not "disturb" unfocused clients.

Broadcasting to all may also resolve weird alt states on some gtk versions.

jlindgren90:
> IIRC the protocol is not clear on whether to broadcast modifiers or not.

Interesting, did not know that.

Proof-of-concept to apply Focus/Raise on Ctrl+click: https://github.com/jlindgren90/labwc/commit/cf80689f6f0265b77dc6477b1d4d5666947be293

I think I prefer that behavior anyway, but others may not of course. Maybe it could be an option in that case.

vyivel:
Some data points: KWin and Weston do pass modifiers to unfocused clients, Mutter does not. (Tested with foot and Ctrl+scroll to change font size.)

jlindgren90:
What event is used to "broadcast" the modifiers? Can we just send that to the client receiving the click, right before the button press?

vyivel:
> What event is used to "broadcast" the modifiers?

`wl_keyboard.modifiers`.

> Can we just send that to the client receiving the click, right before the button press?

[...]

Issue #1507, opened 2024-02-06
Errors when switching windows while using blender (blender loses input on viewport)


charbelnicolas:
When using blender and switching focus from the blender window to another window I sometimes get xcb errors... I don't know if this is related to a bug which makes blender unable to receive any input from the mouse on the current viewport after switching a window with the window switcher. Apparently this happens on SwayWM too per the blender issue tracker.

[https://projects.blender.org/blender/blender/issues/117225#issuecomment-1114837](https://projects.blender.org/blender/blender/issues/117225#issuecomment-1114837)

Here's the screenshot of the errors I get while using blender:

![ss_2024_02_03_17_32_48](https://github.com/labwc/labwc/assets/14960414/37308058-044b-490f-b809-32dad229034c)

I tried Hyprland today using similar shortcuts and didn't have any problems at all with blender so there must be something going on with labwc...

_Originally posted by @charbelnicolas in https://github.com/labwc/labwc/discussions/1502_

charbelnicolas:
This happens to me mostly when switching windows from chromium to blender... Chromium uses xwayland still and blender is wayland native.

Consolatis:
Based on the linked bug report from blender it seems to bail out in some capacity if the keystate after a receiving focus is not consistent with blender's internal xkb state:

`WARN (ghost.wl.keyboard.depressed): intern/ghost/intern/GHOST_SystemWayland.cc:7079 getModifierKeys: modifier (OS) state is inconsistent (GHOST held keys do not match XKB)`

The xcb errors can be related but don't have to. Do you get the same issue when running blender wayland native (it showing up with `[xdg-shell]` in `A-Tab`)? That would eliminate xwayland as potential contributor the the issue.

charbelnicolas:
Yes, this happens when running blender in wayland ([xdg-shell])...

I also got these errors when trying to reproduce the bug:

```
02:55:18.911 [ERROR] [../labwc/src/view.c:980] view has no output, not updating geometry
02:55:18.911 [ERROR] [../labwc/src/view.c:626] view has no output, not positioning
```

```
03:08:54.437 [ERROR] [xwayland/xwm.c:1599] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 8017, value 14680065
03:08:54.437 [ERROR] [xwayland/xwm.c:1599] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 8019, value 14680065
[...]

PR #2274, opened 2024-10-27
keyboard: broadcast modifiers

johanmalm:
...to all clients rather than just the one with keyboard focus on keyboard enter/create, modifer press/release and wlr_seat_set_keyboard().

This enables:

- Clients such as panels to display the current keyboard layout without introducing new wayland protocols or other IPC.

- Unfocused xdg-shell clients to understand button press with keyboard modifiers for example Ctrl+click.

The keymap is forwarded to all clients in wlr_seat_set_keyboard(). When a keymap contains multiple layouts, the selection is made via modifiers, which previously were only sent to the client with keyboard focus.

Tested with: https://github.com/johanmalm/keyboard-layout

Ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4496

Fixes: #2271

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -80,6 +80,43 @@ end_cycling(struct server *server)
 	should_cancel_cycling_on_next_key_release = false;
 }
 
+static struct wlr_seat_client *
+seat_client_from_keyboard_resource(struct wl_resource *resource)
+{
+	return wl_resource_get_user_data(resource);
+}
+
+static void
+broadcast_modifiers_to_unfocused_clients(struct wlr_seat *seat,
+		const struct wlr_keyboard_modifiers *modifiers)
+{
+	struct wlr_seat_client *client;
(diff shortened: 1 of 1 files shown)


jlindgren90:
Yep, this works for me. Thanks!

johanmalm:
v2: Added a longer comment and changed the function name.

Ready for review from my perspective.

If we find that this has side-effects, I suggest we make it optional before the next release.

If we do make it optional, we ought to consider pass the modifiers to the client with pointer-focus even if `broadcastModifiersToUnfocusedClients == false`

jlindgren90:
LGTM

Consolatis:
LGTM as well, timing of the PR is also nice as we are at the start of a release cycle and thus have plenty time for user feedback in case this breaks some clients.

johanmalm:
> Ohh I thought this was going to fix #1507. Any news on that? :)

If you've got the setup, any chance of trying?

PR #2276, opened 2024-10-27
menu: use "MenuHeader" font height in separators with label

domo141:


johanmalm:
Thanks.

domo141:
thanks for quick merge - the "fix" was probably as obvious as i thought

PR #2277, opened 2024-10-29
Add mouse emulation for touch devices

spl237:
Cleaned-up version of mouse emulation code for touch devices.


johanmalm:
Thanks.

PR #2283, opened 2024-10-30
output: split up new_output_notify()

jlindgren90:
... to make it possible in future to *not* immediately auto-configure
new outputs when connected (which will be a future PR).

In particular:
    
- Move regions_reconfigure_output() and session_lock_output_create()
  into add_output_to_layout(), which would be called also if the output
  is enabled later.
- Move seat_output_layout_changed() to do_output_layout_change() so it
  is called for any layout change, not just adding new outputs.
- Then factor out configure_new_output() as a new function.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -316,6 +316,69 @@ add_output_to_layout(struct server *server, struct output *output)
 }
 
 static void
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM

PR #2284, opened 2024-10-31
config/rcxml.c: fix parsing of three-state query parameters

ahesford:
Parsing directly as a bool was setting incorrect values in the query struct.

Reported on IRC by @Consolatis


Consolatis:
LGTM

Issue #2288, opened 2024-10-31
Weirdness with window rules and calibre after 96da82c


Flrian:
I have a window rule that is supposed to make sure the calibre window is maximized and on desktop 3 when I open it. After 96da82c something weird is happening. 
If I close the calibre window while it is maximized (so it also should start maximized), it doesn't get moved to desktop 3. If I close the window while it is unmaximized, it gets maximized and moved to desktop 3. But the same thing happens with any sub windows I open in calibre too, for example with the preferences, add books window or an error message, even though `matchOnce="true"` should make sure that only applies to the main calibre window.

https://github.com/user-attachments/assets/4e51f2b5-8082-4fcb-ab48-585b491247a3

```
    <windowRule identifier="calibre-gui" matchOnce="true">
      <action name="Maximize"/>
      <action name="SendToDesktop" to="3" follow="no"/>
    </windowRule>
```

ahesford:
Sorry, I overlooked some logic errors and some defaults when merging the identified commit. Would you mind testing #2289? It seems to work for my simple test case:

```xml
<windowRule identifier="foot" matchOnce="true">
  <action name="Maximize" />
  <action name="SetDecorations" decorations="border" />
</windowRule>
```

PR #2289, opened 2024-10-31
query: fix default values in query, fix match logic

ahesford:
The `view_query.maximized` field should be initialized to the non-zero `VIEW_AXIS_INVALID` where used, rather than assuming the default zero value of `VIEW_AXIS_NONE`. Also, all of the early (pre-existing) tests in `view_matches_query` were not effectively distinguishing between "not tested" and "tested but passed" with the refactored match structure.

Fixes: #2288.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -71,6 +71,7 @@ view_query_create(void)
 {
 	struct view_query *query = znew(*query);
 	query->window_type = -1;
+	query->maximized = VIEW_AXIS_INVALID;
 	return query;
 }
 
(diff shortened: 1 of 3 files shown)


src/view.c
```diff
@@ -114,38 +115,50 @@ view_matches_query(struct view *view, struct view_query *query)
...
 	}
 
 	if (query->title) {
 		const char *title = view_get_string_prop(view, "title");
-		if (!(title && match_glob(query->title, title))) {
+		match = bool_to_tristate(title && match_glob(query->title, title));
+		if (match == LAB_STATE_DISABLED) {
 			return false;
 		}
 	}
```
Consolatis, line 131: I don't quite see why we have to use `match` here (or in fact anywhere at all). As these are all optional `and` conditions, can't we just `return false` if its requested in the query but doesn't match? Something along this line
```c
static bool
query_str_fail(const char *condition, const char *value)
{
	if (!condition) {
		return false;
	}
	return !(value && match_glob(condition, value));
}

bool
view_matches_query(struct view *view, struct view_query *query)
{
	if (query_str_fail(query->identifier, view_get_string_prop(view, "app_id")) {
		return false;
	}
	if (query_str_fail(query->title, view_get_string_prop(view, "title")) {
		return false;
	}
	if (query->window_type != - 1 && !view_contains_window_type(view, query->window_type)) {
		return false;
	}
	[..]
	if (query->shaded && query->shaded != bool_to_tristate(view->shaded)) {
		return false;
	}
	[..]
}
```

ahesford: We need to distinguish between everything matching and nothing tested; a view never matches an empty query. In the earlier implementation, this was done with a separate `empty` Boolean that was tracked alongside a match variable. The recent change dropped `empty` by making `match` a tristate, but the tristate was never triggered in these early checks. Hence, a positive match in these early tests was counted as an empty query, and the match was failing.

Consolatis: This raises the questions of why do we need to return `false` for an empty query in the first place. The query argument itself is optional but that is handled in `run_if_action()`.

And if we really need it, a dedicated `if (view_query_is_empty(query)) { return false; }` seems easier to read.

ahesford: I don't know why an empty query should fail to match, so I applied the principle of Chesterton's fence: without understanding the behavior, I was unwilling to remove it.

ahesford:
After some consideration, I refactored based on advice from Consolatis to greatly simplify the match logic.

Changes in the latest commit:
- Any empty query will not always match, instead of never match. I did some tests and can't identify a functional difference resulting from this change.
- The function `query_str_fail` was implemented in the affirmative form as `query_str_match`, because I thought that was a bit easier to read.
- Matching tile-to-region now supports globs, which is both cleaner to implement and more functional.

I tested this with the following keybind actions:

```xml
    <keybind key="W-A-S-x">
      <action name="If">
        <query />
        <then>
          <action name="Execute">
            <command>mimic -voice rms -t "It Matches"</command>
          </action>
        </then>
        <else>
          <action name="Execute">
            <command>mimic -voice rms -t "It does not match"</command>
        </action>
        </else>
      </action>
    <action name="If">
      <query identifier="foot" maximized="none" />
      <then>
          <action name="Execute">
            <command>mimic -voice rms -t "Will iconify"</command>
          </action>
        <action name="Iconify" />
      </then>
      <else>
          <action name="Execute">
            <command>mimic -voice rms -t "Not about to iconify that one"</command>
          </action>
      </else>
    </action>
    </keybind>
```

and the following window rules:
```xml
    <windowRule identifier="foot">
      <action name="SetDecorations" decorations="border" />
    </windowRule>

    <windowRule identifier="foot" matchOnce="yes" >
      <action name="Maximize"/>
    </windowRule>
```

The first `foot` instance opens maximized, all others open regularly but with border-only decorations; when I press W-S-A-x, my computer says "It matches" for every window, and (simultaneously, for the non-maximized windows only) says "Will iconify" while iconifying the windows or (simultaneously, for the maximized window) says "Not about to iconify that one" while taking no other action.

@Flrian, please confirm that the latest version works as you expect.

src/view.c
```diff
@@ -88,144 +89,122 @@ view_query_free(struct view_query *query)
...
+			// TODO: perhaps wrap "left" and "right" workspaces
 			struct workspace *target =
-				workspaces_find(view->server->workspaces.current,
-						query->desktop, false);
-			match = bool_to_tristate(target &&
-					!strcasecmp(view->workspace->name, target->name));
-		}
-		if (match == LAB_STATE_DISABLED) {
-			return false;
+				workspaces_find(current, query->desktop, false);
```
Consolatis, line 188: inline comment for `/*wrap*/ false`

Consolatis (commented):
Minor nitpick below, other than that LGTM.

PR #2291, opened 2024-10-31
menu: use only menu_separator_padding_width to pad separators w/o label

domo141:
I.e. drop the extra padding from menu_item_padding_x that got there when separators with labels were implemented.


domo141:
it looks like 0552c6b7f0008 changed the behavior this change "restores"

johanmalm:
Thanks.

Issue #2229, opened 2024-10-09
Tearing preference is fully denied as result of tearing page-flip commit failures


RicArch97:
Since kernel 6.11.1, tearing (async pageflips) now work with the DRM atomic interface, without having to launch with the `WLR_DRM_NO_ATOMIC=1` variable. However it does not currently work as expected in LabWC.

Here's the scenario: i was playing a game, initially tearing was working fine. After some time it no longer did. Checking the logs there seemed to be commit failures where tearing was requested. In MR https://github.com/labwc/labwc/pull/1996 a change was made to fully disable the tearing functionality after a certain amount of failures has happened, though I'm not convinced that this is the right approach.

With the Sway tearing implementation, we also take possible failures into account: https://github.com/swaywm/sway/blob/master/sway/desktop/output.c#L291 If the output state cannot be committed when requesting tearing, then we simply commit the frame without tearing. Commits where tearing is requested could basically fail indefinitely without having impact on usability. I tested the same game in Sway master with tearing enabled and there were no issues (tearing not getting disabled, nor any frame related issues).

I don't think we should just full disable the user defined preference for tearing if commits with tearing page flip enable fail an X amount of times, but rather just handle the failures in a better way. What does everyone think?

johanmalm:
Sounds sensible, but have to admit that this is not an area where I've been following the "thinking" too much.
@kode54 @ahesford 
Any views on this?

ahesford:
I don't have any particular insight or preference here. If the output test as done by sway really caries no appreciable cost, checking indefinitely seems fine.

johanmalm:
> I don't have any particular insight or preference here. If the output test as done by sway really caries no appreciable cost, checking indefinitely seems fine.

Yeah.

@RicArch97 Would you have the means, ability (and interest) to draft a patch for discussion with @kode54 ?

RicArch97:
Actually i was wrong about us fully negating the tearing preference if it fails for X times, but then it is finally committed without tearing page-flip.  One difference i see is that in Sway, the output is state is **tested** first (without committing it yet) with tearing enabled, and then actually **committed** with tearing if the test is positive, or else committed without tearing. 

[...]

PR #2295, opened 2024-11-02
tearing: avoid permanent disable due to rejected commits

RicArch97:
Currently, the cursor plane does not allow async page flips which causes tearing page flips to be rejected if the cursor was moved.

However, in games where no cursor image is present, the async page flips can still work as expected.

Instead of permanently disabling tearing after too many failures, test the output state first before each frame to see if we can commit with tearing_page_flip set to true.

### What is the extra processing time caused by the test commit?

Snippet used:
```c
if (state->tearing_page_flip) {
  struct timespec start, end;
  clock_gettime(CLOCK_MONOTONIC, &start);

  if (!wlr_output_test_state(wlr_output, state)) {
    wlr_log(WLR_DEBUG, "Output test for tearing failed on %s, "
      "trying page-flip without tearing", wlr_output->name);
    state->tearing_page_flip = false;
  }

  clock_gettime(CLOCK_MONOTONIC, &end);

  long seconds = end.tv_sec - start.tv_sec;
  long nanoseconds = end.tv_nsec - start.tv_nsec;
  double elapsed = seconds + nanoseconds*1e-9;

  wlr_log(WLR_DEBUG, "test commit time: %.9f sec", elapsed);
}
```

Commit with `tearing_page_flip` set to true, that doesn't fail
```
00:00:16.648 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000014819 sec
00:00:16.649 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000014560 sec
00:00:16.649 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000014680 sec
00:00:16.649 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000017460 sec
00:00:16.649 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000014560 sec
00:00:16.649 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000014729 sec
00:00:16.650 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000014490 sec
```

Commit with `tearing_page_flip` set to true, that fails
```
00:00:16.658 [DEBUG] [backend/drm/atomic.c:79] connector DP-2: Atomic commit failed: Invalid argument
00:00:16.658 [DEBUG] [backend/drm/atomic.c:84] (Atomic commit flags: PAGE_FLIP_ASYNC | ATOMIC_TEST_ONLY)
00:00:16.658 [DEBUG] [../src/common/scene-helpers.c:118] Output test for tearing failed on DP-2, trying page-flip without tearing
00:00:16.658 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000028439 sec
00:00:16.664 [DEBUG] [backend/drm/atomic.c:79] connector DP-2: Atomic commit failed: Invalid argument
00:00:16.664 [DEBUG] [backend/drm/atomic.c:84] (Atomic commit flags: PAGE_FLIP_ASYNC | ATOMIC_TEST_ONLY)
00:00:16.664 [DEBUG] [../src/common/scene-helpers.c:118] Output test for tearing failed on DP-2, trying page-flip without tearing
00:00:16.664 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000035509 sec
00:00:16.670 [DEBUG] [backend/drm/atomic.c:79] connector DP-2: Atomic commit failed: Invalid argument
00:00:16.670 [DEBUG] [backend/drm/atomic.c:84] (Atomic commit flags: PAGE_FLIP_ASYNC | ATOMIC_TEST_ONLY)
00:00:16.670 [DEBUG] [../src/common/scene-helpers.c:118] Output test for tearing failed on DP-2, trying page-flip without tearing
00:00:16.670 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000028369 sec
00:00:16.676 [DEBUG] [backend/drm/atomic.c:79] connector DP-2: Atomic commit failed: Invalid argument
00:00:16.676 [DEBUG] [backend/drm/atomic.c:84] (Atomic commit flags: PAGE_FLIP_ASYNC | ATOMIC_TEST_ONLY)
00:00:16.676 [DEBUG] [../src/common/scene-helpers.c:118] Output test for tearing failed on DP-2, trying page-flip without tearing
00:00:16.676 [DEBUG] [../src/common/scene-helpers.c:130] test commit time: 0.000027770 sec
```

The change isn't really noticeable when playing a game, and this allows the tearing to keep working regardless of the failures, without having to use `WLR_DRM_NO_ATOMIC=1`. Note that Sway also currently does the test commit.

Fixes: https://github.com/labwc/labwc/issues/2229

### DIFF
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -118,14 +118,22 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
 		magnify(output, state->buffer, &additional_damage);
 	}
 
-	if (state == &output->pending) {
-		if (!output_state_commit(output)) {
-			wlr_log(WLR_INFO, "Failed to commit output %s",
-				wlr_output->name);
-			return false;
+	bool committed = wlr_output_commit_state(wlr_output, state);
+	/*
+	 * Handle case where the ouput state test for tearing succeeded,
+	 * but actual commit failed. Retry wihout tearing.
+	 */
+	if (!committed && state->tearing_page_flip) {
+		state->tearing_page_flip = false;
+		committed = wlr_output_commit_state(wlr_output, state);
+	}
+	if (committed) {
+		if (state == &output->pending) {
+			wlr_output_state_finish(&output->pending);
+			wlr_output_state_init(&output->pending);
 		}
-	} else if (!wlr_output_commit_state(wlr_output, state)) {
-		wlr_log(WLR_INFO, "Failed to commit state for output %s",
+	} else {
+		wlr_log(WLR_INFO, "Failed to commit output %s",
 			wlr_output->name);
 		return false;
 	}
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -107,6 +107,14 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
 		return false;
 	}
 
+	if (state->tearing_page_flip) {
+		if (!wlr_output_test_state(wlr_output, state)) {
+			wlr_log(WLR_DEBUG, "Output test for tearing failed on %s, "
+				"trying page-flip without tearing", wlr_output->name);
+			state->tearing_page_flip = false;
+		}
+	}
+
 	struct wlr_box additional_damage = {0};
 	if (state->buffer && is_magnify_on()) {
 		magnify(output, state->buffer, &additional_damage);
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -109,8 +109,6 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
 
 	if (state->tearing_page_flip) {
 		if (!wlr_output_test_state(wlr_output, state)) {
-			wlr_log(WLR_DEBUG, "Output test for tearing failed on %s, "
-				"trying page-flip without tearing", wlr_output->name);
 			state->tearing_page_flip = false;
 		}
 	}
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -406,8 +406,6 @@ struct output {
 
 	bool leased;
 	bool gamma_lut_changed;
-
-	uint32_t nr_tearing_failures;
 };
 
 #undef LAB_NR_LAYERS
--- a/src/output.c
+++ b/src/output.c
@@ -32,14 +32,6 @@
 #include "view.h"
 #include "xwayland.h"
 
-static unsigned int
-get_tearing_retry_count(struct output *output)
-{
-	/* Two seconds worth of frames, guessing 60Hz if refresh is invalid */
-	int refresh = output->wlr_output->refresh;
-	return refresh > 0 ? refresh / 500 : 120;
-}
-
 bool
 output_get_tearing_allowance(struct output *output)
 {
@@ -54,11 +46,6 @@ output_get_tearing_allowance(struct output *output)
 
 	/* tearing is only allowed for the output with the active view */
 	if (!view || view->output != output) {
-		return false;
-	}
-
-	/* tearing should not have failed too many times */
-	if (output->nr_tearing_failures >= get_tearing_retry_count(output)) {
 		return false;
 	}
 
@@ -154,22 +141,7 @@ output_frame_notify(struct wl_listener *listener, void *data)
 
 		pending->tearing_page_flip = output_get_tearing_allowance(output);
 
-		bool committed =
-			lab_wlr_scene_output_commit(scene_output, pending);
-
-		if (pending->tearing_page_flip) {
-			if (committed) {
-				output->nr_tearing_failures = 0;
-			} else {
-				if (++output->nr_tearing_failures >=
-						get_tearing_retry_count(output)) {
-					wlr_log(WLR_INFO, "setting tearing allowance failed "
-						"for two consecutive seconds, disabling");
-				}
-				pending->tearing_page_flip = false;
-				lab_wlr_scene_output_commit(scene_output, pending);
-			}
-		}
+		lab_wlr_scene_output_commit(scene_output, pending);
 	}
 
 	struct timespec now = { 0 };


Consolatis:
Fine with me in general. Thoughts @kode54 (and others)?

src/common/scene-helpers.c
```diff
@@ -107,6 +107,14 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
 		return false;
 	}
 
+	if (state->tearing_page_flip) {
+		if (!wlr_output_test_state(wlr_output, state)) {
+			wlr_log(WLR_DEBUG, "Output test for tearing failed on %s, "
+				"trying page-flip without tearing", wlr_output->name);
```
ahesford, line 113: I doubt we should fire a log message every time this fails, even in debug mode, especially when it could potentially fail with every single frame.

 If there is value in logging these failures at all, I would wrap the logging in a static function with some static counters every M seconds or N frames, whichever is most sensible and easiest to support.

ahesford (changes requested):
This is fine in general, although it would be nice to have some hard evidence of the performance cost from check-every-commit compared to that of try-then-fail, in "typical" workflows and the worst case where every frame might be a failure.

src/common/scene-helpers.c
RicArch97: >  especially when it could potentially fail with every single frame.

Only when tearing is requested by a surface, which shouldn't happen all that often, mostly games (where the failures mostly happen in menus).

But I agree, I think we can just get rid of this log message.  The atomic backend will log as well when this check fails.

src/common/scene-helpers.c
```diff
@@ -107,6 +107,14 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
...
+			state->tearing_page_flip = false;
```
Consolatis, line 114: In case the test commit returns `true` but the actual commit fails we should likely have another safeguard in place to retry the commit with tearing disabled. I don't know if that is a possible scenario for any GPU driver but if it is we would skip the frame completely with this PR.

RicArch97: Is that also the case when `state == &output->pending`?

kode54:
I'm fine with making it less annoying. I'd also like to see that kernel patch I mentioned in the tearing issue merged into the kernel, so at least this becomes less of a problem. I'm applying it to a kernel and building now, and will test further.

Though I don't expect a whole lot of tearing from my current setup, since I can reach 165Hz with sync enabled.

src/common/scene-helpers.c
Consolatis: Yes, that issue would exist with both states. We could refactor this slightly to only have a single `wlr_output_commit_state()` and merge the state reset in `output_state_commit()` from `src/output-state.c` into `lab_wlr_scene_output_commit()` and remove `output_state_commit()` to have less duplication. E.g. something like (completely untested pseudo code):
```c
lab_wlr_scene_output_commit([..]) {
	[..]

	if (state->buffer && is_magnify_on()) {
		magnify(output, state->buffer, &additional_damage);
	}

	if (!wlr_output_commit_state(wlr_output, state)) {
		if (state->tearing_page_flip) {
			/*
			 * Test commit with tearing went through but actual
			 * commit failed, lets try again without tearing
			 */
			state->tearing_page_flip = false;
			if (wlr_output_commit_state(wlr_output, state)) {
				goto commit_ok;
			}
		}
		wlr_log(WLR_INFO, "Failed to commit output %s", wlr_output->name);
		return false;
	}

commit_ok:
	if (state == &output->pending) {
		wlr_output_state_finish(&output->pending);
		wlr_output_state_init(&output->pending);
	}

	if (!wlr_box_empty(&additional_damage)) {
		[..]
	}
	return true;
}
```

CC @jlindgren90

RicArch97: To see it was a bit confusing why there were 2 different functions for commit used, with almost identical names (for someone pretty new to LabWC and wlroots at least). 

So guessing from `state == &output->pending` and `wlr_output_state_finish` being called on the state it is the commit for the final output state? The function name might need some more context

RicArch97: I think something like this would be the cleanest;

```c
bool committed = false;
if (state == &output->pending) {
  committed = output_state_commit(output);
} else {
  committed = wlr_output_commit_state(wlr_output, state);
}
/* 
 * Handle case where the ouput state test for tearing succeeded, 
 * but actual commit failed. Retry wihout tearing.
 */
if (!committed && state->tearing_page_flip) {
  state->tearing_page_flip = false;
  committed = wlr_output_commit_state(wlr_output, state);
}
if (!committed) {
  wlr_log(WLR_INFO, "Failed to commit output %s", wlr_output->name);
  return false;
}
```

Or we could use the conditional operator if you prefer that;

```c
bool committed = (state == &output->pending)
       ? output_state_commit(output)
       : wlr_output_commit_state(wlr_output, state);
/* 
 * Handle case where the ouput state test for tearing succeeded, 
 * but actual commit failed. Retry wihout tearing.
 */
if (!committed && state->tearing_page_flip) {
  state->tearing_page_flip = false;
  committed = wlr_output_commit_state(wlr_output, state);
}
if (!committed) {
  wlr_log(WLR_INFO, "Failed to commit output %s", wlr_output->name);
  return false;
}
```

Consolatis: This would not reset the shared `output->pending` state on non-tearing retry.
We should simply get rid of `output_state_commit()` in this function and reset `output->pending` directly in case that is what we are committing.

 > To me it was a bit confusing why there are 2 different functions for commit used, with almost identical names (for someone pretty new to LabWC and wlroots at least).

Historical reason, needs to be cleaned up.

> So guessing from `state == &output->pending` and `wlr_output_state_finish` being called on the state it is the commit for the final output state? The function name might need some more context

No, we do have a shared pending state for the next commit in `output->pending`. If a commit went through we need to reset it (finish -> init) so we don't end up committing the same things over and over again. Some parts of the code base do use their own state instead of the shared `output->pending` one in which case we must not reset the shared state because we would loose some uncommitted changes.

src/common/scene-helpers.c
RicArch97: Removed the log.

src/common/scene-helpers.c
RicArch97: Changed, should cover all cases now

Consolatis:
LGTM, not actually tested.

johanmalm:
Thanks.

PR #2300, opened 2024-11-02
data: disable file-chooser in xdg-portals configuration

jp7677:
Similar like the Inhibit interface, the FileChooser does not work on labwc and thus preventing a file dialog from appearing in applications/toolkits using this interface (e.g. GTK 4.16).

Disabling this interfaces makes those GTK4 application fallback to an alternative file open/safe dialog which works again normally on labwc.

Observed and tested with Gnomes Simple-Scan and Clapper i.c.w. GTK 4.16.

### DIFF
--- a/data/labwc-portals.conf
+++ b/data/labwc-portals.conf
@@ -1,3 +1,4 @@
 [preferred]
 default=wlr;*
 org.freedesktop.impl.portal.Inhibit=none
+org.freedesktop.impl.portal.FileChooser=none


Consolatis:
Why doesn't the file-chooser portal work? Pretty sure I've seen the gtk 3 version working before.

jp7677:
> Why doesn't the file-chooser portal work? Pretty sure I've seen the gtk 3 version working before. Note, this also disables the file-chooser for other portal backends like KDE.

Without this setting I don't have the file chooser in above mentioned applications (like: nothing happens after clicking on open/save button) after having updated to Fedora 41 which ships GTK4 4.16. I haven't had this issue with GTK4 4.14 on Fedora 40, nor with GTK3 applications. I don't know if KDE uses a portal for their file chooser, so yeah, may be wise to keep this draft until others hit this.

Consolatis:
IMHO the main thing here is apparently some bug in connection with the GTK 4.16 implementation of the file chooser portal. If its on the labwc side we should fix it and if its on the GTK side we should report it. I don't think a bug in a portal implementation of a specific portal backend warrants the complete deactivation of the whole portal.

jp7677:
I still have to test this, but according to https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7932 there is `GDK_DEBUG=no-portals` which might also do the trick. Will report back.

Consolatis:
@jp7677 could you give
- #2400

a try to see if it fixes the GTK4 file chooser portal backend?

jp7677:
@Consolatis Sorry for the late response. No, unfortunately not. With labwc master from today I still need `GDK_DEBUG=no-portals` for the save dialog in Gnomes Simple-Scan and Clapper.

Consolatis:
Well, it was worth a try. Thanks for testing :)

PR #2305, opened 2024-11-05
add Iconify shortcut to rc.xml.all

muezabdalla:
add Iconify shortcut to rc.xml.all so that new users will find it on the example config file


Consolatis:
`rc.xml.all` should replicate the default keybinds and settings as hardcoded in labwc. As there is no `Iconify` keybind available by default in labwc we also can't add one to `rc.xml.all`.

I think the minimize window button is a well established pattern so I also don't see a reason to add a keybind for it by default.

PR #2314, opened 2024-11-07
Fix buf off by one bug

johanmalm:


t/buf-simple.c
```diff
@@ -69,11 +69,32 @@ test_buf_add_fmt(void **state)
...
+	struct buf s = BUF_INIT;
+
+	/*
+	 * Start off with a long string so that the allocated buffer is only
+	 * just large enough to contain the string including the NULL
+	 * termination. This way we prove that the allocation triggered by the
+	 * subsequent buf_add_char() is appropriate.
+	 */
+	buf_add(&s, "123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890");
+	buf_add_char(&s, '+');
```
Consolatis, line 86: I think this would need to be on the end of the initial allocation. e.g. initial string 255 chars + the single `buf_add_char()` call. The result should be `s->alloc == 384` in the fixed implementation with `s->alloc * 3 / 2`. In the error case it would still be `256` (because `255 + 1` != < `s->alloc`).

Consolatis: Nevermind about the `255` part. Having an initial allocation > 256 should be fine.
I'd still add the `s->len + 1 <= s->alloc` condition to verify we didn't write outside the allocation.

johanmalm: Done.

jlindgren90:
Thanks!

PR #2319, opened 2024-11-08
menu: correctly center a menu opened with <position x="center" y="center">

tokyo4j:
Prior to this PR, a menu opened with:
```xml
  <action name="ShowMenu">
    <position x="center" y="center" />
  </action>
```
was not correctly centered when `menu.items.padding.x` is non-zero.


johanmalm:
Thanks

Issue #2320, opened 2024-11-08
Mouse pointer speed setting bug


spl237:
### Problem description

The range of pointerSpeed in the libinput section of rc.xml is quoted as -1.0 to +1.0, but setting it to -1.0 has no effect

### Steps to reproduce

https://github.com/labwc/labwc/blob/master/src/seat.c#L189 has a problem, because a > is used which should be a >=.

https://github.com/labwc/labwc/blob/master/docs/rc.xml.all#L598 states the valid range is -1.0 to 1.0.

But if you set the pointerSpeed to a value other than -1.0 and --reconfigure, the new speed loads. If you then set the speed to -1.0 and --reconfigure, the speed does not change from the previous value, because this test fails. If -1.0 is a valid value - and it is according to the documentation - then I think this test is wrong.

### labwc build source

Release

### labwc version

0.7.4

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

RPiOS

spl237:
PR created - https://github.com/labwc/labwc/pull/2321

PR #2321, opened 2024-11-08
Fix bug whereby pointer speed of -1 is not valid

spl237:
As per https://github.com/labwc/labwc/issues/2320


Consolatis:
Its initialized as `-2` so this seems fine. Thanks!

PR #2322, opened 2024-11-08
Fix broken links to Openbox

tinyboxvk:


Consolatis:
Looks like the openbox site has been moved to https://github.com/openboxwm/openboxwm.github.io by the creator of Openbox which is also delivered via https://openbox.org. So these changes look fine. Thanks!

PR #2323, opened 2024-11-08
Bump actions/checkout to v4

tinyboxvk:


Consolatis:
What would be the benefit of switching to v4?

tinyboxvk:
> What would be the benefit of switching to v4?

v2 and above should have improved performance but node.js is added as a dependency. As a result, it breaks the Void-musl build.

PR #2325, opened 2024-11-08
environment: disallow recursive env var assignments

johanmalm:
...like `FOO=$FOO:bar` to avoid environment variables growing on reconfigure.

Cc @spl237

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -50,9 +50,14 @@ that directory contains either the "environment" file or at least one
 
 Note: environment files are treated differently by Openbox, which will simply
 source the file as a valid shell script before running the window manager. Files
-are instead parsed directly by labwc, although any environment variables
-referenced as $VARIABLE or ${VARIABLE} will be substituted and the tilde (~)
-will be expanded as the user's home directory.
+are instead parsed directly by labwc so that environment variables can be
+re-loaded on --reconfigure.
+
+Any environment variables referenced as $VARIABLE or ${VARIABLE} will be
+substituted and the tilde (~) will be expanded as the user's home directory.
+
+Please note that as labwc reloads the environment file(s) on reconfigure,
+recursive/circular assignments (for example FOO=$FOO:bar) should not be made.
 
 The *autostart* file is executed as a shell script after labwc has read its
 configuration and set variables defined in the environment file. Additionally,
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -33,6 +33,7 @@ static const char *const env_vars[] = {
 	NULL
 };
 
+#define LAB_ENV_VAR_MAX_SIZE 1024
 static void
 process_line(char *line)
 {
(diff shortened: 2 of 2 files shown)


Consolatis:
This doesn't catch stuff like
```sh
foo=bar${baz}
baz=$foo
```
Honestly I am not sure if its worth the additional code to prevent this (or recursion in general), at some point users should be able to shoot themselves in the foot if they choose to.

My take on this in general is that we should look at what might be the worst thing to happen due to the recursion. And for me I think that is running OOM due to some automated `labwc --reconfigure` calls like from labwc-tweaks and similar implementations. So just checking the size before setting an env var seems enough for me.

jlindgren90:
Would it be difficult to save off the original value of environment variables that are set, and then use the original value in later substitutions? It seems like that would better match the intent of self-referencing variables.

johanmalm:
> Would it be difficult to save off the original value of environment variables that are set, and then use the original value in later substitutions? It seems like that would better match the intent of self-referencing variables.

Yeah, that would probably be the way to implement this... if we want to support it. Have to admit, the more I think about it I can't see a use case that can't be sorted by a shell script. Suggest we just do something around size (as @Consolatis suggested) and offer a warning.

spl237:
I think something to restrict maximum size of an envvar is probably the most robust (and simplest) solution.

johanmalm:
Okay. Attempt number 2. Now just ignoring assignments if size is greater than 1k.

docs/labwc-config.5.scd
```diff
@@ -50,9 +50,14 @@ that directory contains either the "environment" file or at least one
...
-referenced as $VARIABLE or ${VARIABLE} will be substituted and the tilde (~)
-will be expanded as the user's home directory.
+are instead parsed directly by labwc so that environment variables can be
+re-loaded on --reconfigure.
+
+Any environment variables referenced as $VARIABLE or ${VARIABLE} will be
+substituted and the tilde (~) will be expanded as the user's home directory.
+
+Note: Recursive variable assignments (for example FOO=$FOO:bar) is disallowed to
+prevent environment variables from growing on each --reconfigure.
```
Consolatis, line 60: Outdated note

src/config/session.c
```diff
@@ -33,6 +34,36 @@ static const char *const env_vars[] = {
...
+	}
+
+	char c, *end = NULL;
+	long val = strtol(str, &end, 0);
+	if (!val || val == LONG_MAX) {
+		return -1;
+	}
+	while ((c = *end++) != 0) {
+		switch (c) {
+		case 'k':
```
Consolatis, line 51: lowercase `k` but uppercase `M` and `G` seems a bit weird to me. Also: do we actually need the string parsing in the first place? I think the limit could also be a simple define.

johanmalm:
@Consolatis Thanks for review. Have simplified and addressed all. Ready for final review from my perspective.

Consolatis:
LGTM, if 1024 byte ends up being too small for users we can always increase it later.

PR #2326, opened 2024-11-10
input: add hold gestures, notify idle-manager and ensure visible cursor on gestures

jp7677:
Adds hold gestures (added in `wlroots` with https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/20d9448257f4871d3a57deab2f105d7335a64fb2) and also add the usual ceremony for the gesture handlers.

Not tested because no touch pad on my development machine. That said, i don't expect any side effects (famous last words ..). Keeping this draft until confirmation that there are indeed no side effects.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -179,6 +179,8 @@ struct seat {
 	struct wl_listener swipe_begin;
 	struct wl_listener swipe_update;
 	struct wl_listener swipe_end;
+	struct wl_listener hold_begin;
+	struct wl_listener hold_end;
 
 	struct wl_listener request_cursor;
 	struct wl_listener request_set_shape;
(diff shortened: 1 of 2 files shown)


Consolatis:
LGTM code wise but didn't test. Is there some application which makes use of the hold gesture?

jp7677:
Actually I don't know. I only wanted to add the idle notifications and the cursor visibility and stumbled upon that holds weren't implemented.

johanmalm:
Would `firefox` be a candidate for testing this? Seems to handle gestures.

tokyo4j:
I confirmed kinetic scrolling can be paused when I fold fingers on touchpad. Feels pretty nice.

jp7677:
Just to be sure, I can also confirm that this PR doesn't interfere with touch screens. Gestures in `firefox` were still working and the cursor stayed hidden. So ready for review.

Consolatis:
Thanks!

PR #2331, opened 2024-11-11
Fix memory leaks detected by ASAN

tokyo4j:
This PR fixes some memory leaks (including ones for singletons) detected by ASAN.
The first 5 commits fixes memory leaks for `output->osd_buffer`, `icon_loader->icon_theme`, `seat->workspace_osd_timer`, `cosmic_manager` and `server->ssd_hover_state` respectively, and the last commit finally releases all the resources like `wlr_allocator`, `wlr_renderer`, `wlr_backend` and `wlr_scene`.

To suppress all the errors from ASAN with EGL renderer, https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/ce615a44c06ed10ab51b9c5a630acb43ef014efb needs to be reverted.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -399,8 +399,6 @@ struct output {
 
 	struct wl_list regions;  /* struct region.link */
 
-	struct lab_data_buffer *osd_buffer;
-
 	struct wl_listener destroy;
 	struct wl_listener frame;
 	struct wl_listener request_state;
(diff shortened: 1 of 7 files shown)


Consolatis:
LGTM, not tested.

johanmalm:
@tokyo4j I am delighted with this patch set. This is super work!
I have tried before to get a clean ASAN run, but not achieved it. It's so much easier to spot memory leaks once a clean run has been achieved.

We could consider adding an `lsan.supp` with `leak:libEGL_mesa.so` for the EGL stuff.

Issue #2329, opened 2024-11-11
Weird cycling focus logic when switching between workspaces


pirat89:
### Problem description

When switching between workspaces/desktops, each time a different window (available/visible on the target workspace) is activated/focused, even when the window originally focused/active is available on both (original and target) workspaces. This is weird and kind of annoying as many times I start to type on keyboard when a different window is focused after the workspace switch, so I guess this is not intended behavior. Witch 3+ windows set as omnipresent, the focus cycle through all of them after each change of the workspace.

Additional Notes:
* I realized that when using multiple outputs and I haven't tried it yet with a single output device to see whether it depends to that fact anyhow. Both omnipresent windows are displayed in the same output.
* I guess there is a logic to unfocus current active window when switching a workspace which leads to the situation the next window is picked.
* If I have a time, I will take a look at it if I can find more or come up with a fix.
* The behaviour seems same in labwc `0.7.4` and newer - including `0.8.1`. (I haven't tried versions before 0.7.4)

### Steps to reproduce

1. open 2+ windows and set both of them omnipresent
2. go to a different workspace

the focus is switched each time.

### labwc build source

Release

### labwc version

0.8.1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Fedora 40+

Consolatis:
Can confirm, something is off with omnipresent + desktop switching.

Consolatis:
Can you give #2332 a try?

pirat89:
@Consolatis works like a charm. thanks!

Consolatis:
Thanks for reporting and testing. Should be fixed in the master branch.

Issue #1151, opened 2023-10-10
Decouple stickyness from alwaysOnBottom/Top


stefonarch:
I didn't find any open issue about that,but it was mentioned somewhere already.
`action name="sticky"` and remove it from` alwaysOn*`
Maybe this could be also in client menu by default.

Related https://github.com/labwc/labwc/issues/856

tsujan:
This could be used for putting Conly or Qt/GTK widgets (like analog clocks) on all desktops. `ToggleAlwaysOnBottom` couldn't be used if PCManFM-Qt (or any FM) draws the desktop, because in that case, the window will go behind the drawn desktop once the latter is clicked.

Consolatis:
We do have a `ToggleOmnipresent` now (personally I would be fine with adding a `ToggleSticky` alias which simply triggers the same action). Agree that we should get rid of the `on-all-workspaces` behavior of `ToggleAlwaysOn{Bottom,Top}`.

Maybe we could simply have the always-on-top/always-on-bottom scene trees be parented to the workspace trees rather than the layout tree.

tsujan:
> We do have a ToggleOmnipresent now

It's nice but, sadly, it didn't work for what I had in mind. The reason is that it makes the window come to the front on switching the desktop. So, for example, when I put an analog clock (or Conky) on desktop and want it to appear on all desktops, I can't use it (I don't want the clock in front of Firefox when the desktop is changed to where Firefox is). Moreover, I can't add `ToggleAlwaysOnBottom` to the clock because it would easily make it go behind pcmanfm-qt's desktop.

Vladimir-csp:
Window seems to loose Always On Top state after workspace switch.

PR #2332, opened 2024-11-12
workspaces: keep focus on omnipresent windows when switching workspaces

Consolatis:
Fixes:
- #2329

### DIFF
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -339,12 +339,19 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	server->workspaces.current = target;
 
 	/*
-	 * Make sure we are focusing what the user sees.
-	 * Only refocus if the focus is not already on an always-on-top view.
+	 * Make sure we are focusing what the user sees. Only refocus if
+	 * the focus is not already on an omnipresent or always-on-top view.
+	 *
+	 * TODO: Decouple always-on-top views from the omnipresent state.
+	 *       One option for that would be to create a new scene tree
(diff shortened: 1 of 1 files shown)


tokyo4j:
I think `!view->visible_on_all_workspaces` clarifies the intention better than `view->workspace != target`. Also I suggest adding a TODO comment to decouple the sticky behavior from AlwaysOnTop windows (#1151).

johanmalm:
LGTM. Not tested.

Consolatis:
Addressed comment.

One more related issue I noticed is the following:
- two omnipresent views and a usual one
- focus the usual one
- switch workspace
- one of the two omnipresent views gets focus
- switch back to the previous workspace
- focus the usual one
- switch workspace
- now the other omnipresent view gets focus

I think this is because we are using `view_for_each()` to move the omnipresent views to the new workspace. However, `server->views` is ordered topmost first and we only reparent the view scene tree without changing the ordering of `server->views`. `desktop_focus_topmost_view()` in contrast uses the scene node ordering to find the topmost view and then raises that one.

An easy 'fix' would be to implement `view_for_each_reverse()` and use that one to move the omnipresent views to the new workspace. I've tested that and it looks like it at least makes things deterministic. I feel like focus handling of omnipresent views on workspace switching needs some more thought instead of adding band aids like `view_for_each_reverse()` though. Currently omnipresent views `*always*` get focus on workspace switching because they are the topmost views in terms of the scene graph. That feels slightly wrong (and also desyncs `server->views` and `server->workspaces.current->tree->children`).

PR #2334, opened 2024-11-12
Add lsan.supp

johanmalm:
...so suppress errors from ASAN with EGL renderer.

Also add Makefile to simplify running with sanitizer and this suppression file.

Related-to: #2331

### DIFF
--- a/scripts/asan_leak_suppressions
+++ b/scripts/asan_leak_suppressions
@@ -1,1 +1,2 @@
 leak:libfontconfig.so
+leak:libEGL_mesa.so
--- a/scripts/run-with-asan.sh
+++ b/scripts/run-with-asan.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+: ${BUILD_DIR="build"}
+
+[ -d "${BUILD_DIR}" ] || meson setup "${BUILD_DIR}"
+meson configure -Db_sanitize=address,undefined "${BUILD_DIR}"
+ninja -C "${BUILD_DIR}"
+LSAN_OPTIONS=suppressions=scripts/asan_leak_suppressions "${BUILD_DIR}/labwc" -d 2>log.txt
+


Consolatis:
The `Makefile` in the root directory feels weird IMHO. It could make users expect to run `make` rather than `meson setup build; [..]`.

This feels more like it would belong in some test, scripts or ci directory.

tokyo4j:
Strangely, I can no longer see memory leak errors from ASAN (which were locally present in 2 weeks ago) in the current master. Were you able to see the errors and confirm this PR suppresses them?

johanmalm:
> Strangely, I can no longer see memory leak errors from ASAN (which were locally present in 2 weeks ago) in the current master. Were you able to see the errors and confirm this PR suppresses them?

Yes, I saw the EGL memory leaks, and the lsan.supp file suppresses them.

johanmalm:
> The `Makefile` in the root directory feels weird IMHO. It could make users expect to run `make` rather than `meson setup build; [..]`.
> 
> This feels more like it would belong in some test, scripts or ci directory.

Yes, I did ponder that. It's nice to keep the toplevel directory clean actually isn't it. I'll do a v2.

johanmalm:
Okay, ready for review.

tokyo4j:
> > Strangely, I can no longer see memory leak errors from ASAN (which were locally present in 2 weeks ago) in the current master. Were you able to see the errors and confirm this PR suppresses them?
> 
> Yes, I saw the EGL memory leaks, and the lsan.supp file suppresses them.

I can see the error messages again after downgrading mesa from 1:24.2.7-1 to 1:24.2.5-1... that's strange.

Consolatis:
I've added a runtime leak test to the CI in #2383.

That one uses `LSAN_OPTIONS=suppressions=scripts/asan_leak_supressions` (which only contains `libfontconfig.so` because the version shipped by Debian Testing seems to leak).

Consolatis:
> Well, shall we delete this PR. Probably not needed.

Up to you, I don't see anything wrong with adding `libEGL_mesa.so` to `scripts/asan_leak_suppression` and have a small wrapper to compile and start labwc with the leak detection enabled.

johanmalm:
Like so…?

Consolatis:
Yep, LGTM

PR #2335, opened 2024-11-12
workspaces: use view_for_each_reverse() to move omnipresent views

Consolatis:
Follow-up from #2332 

One more related issue I noticed is the following:
- two omnipresent views and a usual one
- focus the usual one
- switch workspace
- one of the two omnipresent views gets focus
- switch back to the previous workspace
- focus the usual one
- switch workspace
- now the other omnipresent view gets focus

I think this is because we are using `view_for_each()` to move the omnipresent views to the new workspace. However, `server->views` is ordered topmost first and we only reparent the view scene tree without changing the ordering of `server->views`. `desktop_focus_topmost_view()` in contrast uses the scene node ordering (which is topmost last) to find the topmost view and then raises that one (which then in turn moves it to the front of `server->views`).

An easy 'fix' would be to implement `view_for_each_reverse()` and use that one to move the omnipresent views to the new workspace. I've tested that and it looks like it at least makes things deterministic. I feel like focus handling of omnipresent views on workspace switching needs some more thought instead of adding band aids like `view_for_each_reverse()` though. Currently omnipresent views `*always*` get focus on workspace switching because they are the topmost views in terms of the scene graph. That feels slightly wrong (and also desyncs `server->views | LAB_VIEW_CRITERIA_CURRENT_WORKSPACE` and `server->workspaces.current->tree->children`).

This PR implements the `view_for_each_reverse()` band aid.

Ideas how to proceed welcome.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -381,10 +381,26 @@ bool view_matches_query(struct view *view, struct view_query *query);
  *		printf("%s\n", view_get_string_prop(view, "app_id"));
  *	}
  */
-#define for_each_view(view, head, criteria)		\
-	for (view = view_next(head, NULL, criteria);	\
-	     view;					\
+#define for_each_view(view, head, criteria)           \
+	for (view = view_next(head, NULL, criteria);  \
+	     view;                                    \
 	     view = view_next(head, view, criteria))
+
+/**
+ * for_each_view_reverse() - iterate over all views which match criteria
+ * @view: Iterator.
+ * @head: Head of list to iterate over.
+ * @criteria: Criteria to match against.
+ * Example:
+ *	struct view *view;
+ *	for_each_view_reverse(view, &server->views, LAB_VIEW_CRITERIA_NONE) {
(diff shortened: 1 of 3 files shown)


johanmalm:
LGTM. I think we should merge this until such time we can refactor the omnipresent-view code wrt workspces.

> I feel like focus handling of omnipresent views on workspace switching needs some more thought instead of adding band aids like `view_for_each_reverse()` though. Currently omnipresent views `*always*` get focus on workspace switching because they are the topmost views in terms of the scene graph. That feels slightly wrong (and also desyncs `server->views | LAB_VIEW_CRITERIA_CURRENT_WORKSPACE` and `server->workspaces.current->tree->children`).

The way forward is not obvious to me. I did have a go at de-coupling omnipresent from always-on-top some time ago, but got stuck on some odd behaviors which were probably related to what you describe above.

Could a solution be for each workspace to remember its `active_view` - and should that `active_view` no longer exist, we just choose `desktop_topmost_focusable_view()` which iterates over `server->workspaces.current` (and thus ignores always-on-top)

PR #2336, opened 2024-11-12
menu: refactor in preparation for richer menu design (+ minor fixes)

tokyo4j:
This PR is a preparation for richer menu design including borders, rounded corners and padding around items (ref: https://github.com/tokyo4j/labwc/tree/rounded-menu, https://github.com/labwc/labwc/pull/2240#issuecomment-2422437018).

The first & second commits split the initialization of menu items into `item_create()`/`separator_create()` and `item_create_scene()`/`separator_create_scene()`/`title_create_scene()` so that we can create all the scene nodes in items after knowing the final menu width. This will allow us to smoothly add support for rounded corners without bloating `menu_update_width()` (renamed to `menu_update_scene()` in this PR) as we cannot resize cairo surfaces unlike scene-rects.

The third commit fixes that `menu.title.text.justify: right` worked like `menu.title.text.justify: left`.

The forth commit fixes that `menu.width.{min,max}` was applied to the width of menu item content rather than the menu width including `menu.items.padding.x`. (cc @Consolatis as I don't fully understand the current calculation)

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -339,15 +339,14 @@ menu_update_scene(struct menu *menu)
 	menu->scene_tree = wlr_scene_tree_create(menu->server->menu_tree);
 	wlr_scene_node_set_enabled(&menu->scene_tree->node, false);
 
-	/* Get widest menu item, clamped by menu_max_width */
-	int max_width = theme->menu_min_width;
+	/* Menu width is the maximum item width, capped by menu.width.{min,max} */
+	menu->size.width = 0;
 	wl_list_for_each(item, &menu->menuitems, link) {
-		if (item->native_width > max_width) {
-			max_width = item->native_width < theme->menu_max_width
-				? item->native_width : theme->menu_max_width;
-		}
+		menu->size.width = MAX(menu->size.width,
+			item->native_width + 2 * theme->menu_items_padding_x);
 	}
-	menu->size.width = max_width + 2 * theme->menu_items_padding_x;
+	menu->size.width = MAX(menu->size.width, theme->menu_min_width);
+	menu->size.width = MIN(menu->size.width, theme->menu_max_width);
 
 	/* Update all items for the new size */
 	int item_y = 0;
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -15,6 +15,7 @@
 #include "common/dir.h"
 #include "common/font.h"
 #include "common/list.h"
+#include "common/macros.h"
 #include "common/mem.h"
 #include "common/nodename.h"
 #include "common/scaled-font-buffer.h"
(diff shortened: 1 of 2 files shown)


johanmalm:
This is really nice. Thank you.
The new `menu_update_scene()` is much nicer :smile:

src/menu/menu.c
```diff
@@ -253,46 +182,34 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
...
+		theme->menu_items_text_color,
+		theme->menu_items_bg_color, menuitem->arrow);
+	scaled_font_buffer_update(menuitem->selected.buffer, menuitem->text,
+		menuitem->native_width, &rc.font_menuitem,
+		theme->menu_items_active_text_color,
+		theme->menu_items_active_bg_color, menuitem->arrow);
+	scaled_font_buffer_set_max_width(menuitem->normal.buffer,
+		menu->size.width - 2 * theme->menu_items_padding_x);
+	scaled_font_buffer_set_max_width(menuitem->selected.buffer,
+		menu->size.width - 2 * theme->menu_items_padding_x);
```
Consolatis, line 201: These calls are redundant and force a new render. Also I am slightly confused why the `max_width` argument in `scaled_font_buffer_update()` is different to the one in `scaled_front_buffer_set_max_width()`.

src/menu/menu.c
```diff
@@ -253,46 +182,34 @@ item_create(struct menu *menu, const char *text, bool show_arrow)
...
-		wlr_log(WLR_ERROR, "Failed to create menu item '%s'", text);
-		/*
-		 * Destroying the root node will destroy everything,
-		 * including the node descriptor and scaled_font_buffers.
-		 */
-		wlr_scene_node_destroy(&menuitem->tree->node);
-		free(menuitem);
-		return NULL;
-	}
+	assert(menuitem->normal.buffer && menuitem->selected.buffer);
```
Consolatis, line 185: From memory this has been added to catch using a font size of `0` but I didn't verify.
Also, some recent refactoring seems to have broken `lab_data_buffer` with either height or width of `0` as it now asserts in `data_buffer_begin_data_ptr_access()` whereas before it would return `NULL` and thus trigger this check.

src/menu/menu.c
tokyo4j: Yeah, it's my mistake. I carelessly copied and modified the code, so I didn't realize `scaled_font_buffer_update()` contains `max_width`. I'll open a PR to fix it.

src/menu/menu.c
tokyo4j: I see. I thought it was just for OOM and didn't think about that. But when would anyone want to set the font sizes to zero? Can't we just force them to be positive?

Consolatis: I think a font size of 0 was used at some to hide the title (or even the whole titlebar, can't remember right now). Personally I don't feel strongly about supporting a font size of a `0` vs enforcing a font size `> 0`. If we go the 2nd route we should still add an assert when creating a lab_data_buffer though.

Issue #2337, opened 2024-11-13
Locking session blindly with no outputs causes labwc to crash


kode54:
### Problem description

I disconnected my monitors, and then proceeded to lock the session blindly with a keyboard shortcut. This caused labwc to crash back to GDM.

### Steps to reproduce

1. Launch labwc with one monitor
2. Disconnect the monitor
3. Lock the session with swaylock using a keyboard shortcut

### labwc build source

Local build

### labwc version

0.8.1-44-g2d3be5ab

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

### Crash dump

```
#0  wlr_output_state_set_adaptive_sync_enabled (state=state@entry=0x20, 
    enabled=enabled@entry=false)
    at ../wlroots-hidpi-xprop/types/output/state.c:59
#1  0x000056d6c53ce938 in output_enable_adaptive_sync (output=0x0, 
    enabled=false) at ../labwc/src/output.c:1051
#2  0x000056d6c53e1945 in set_adaptive_sync_fullscreen (view=0x56d6d29347e0)
    at ../labwc/src/view.c:461
#3  set_adaptive_sync_fullscreen (view=0x56d6d29347e0)
    at ../labwc/src/view.c:455
#4  view_set_activated (view=0x56d6d29347e0, activated=<optimized out>, 
    activated@entry=false) at ../labwc/src/view.c:486
#5  0x000056d6c53e1a4e in focus_change_notify (listener=<optimized out>, 
    data=<optimized out>) at ../labwc/src/seat.c:512
#6  0x00007e11bb1fa47e in wl_signal_emit_mutable (
    signal=signal@entry=0x56d6d1ddb038, data=data@entry=0x7ffc0a776400)
    at ../wayland-1.23.1/src/wayland-server.c:2314
#7  0x00007e11bb143c2c in wlr_seat_keyboard_enter (seat=0x56d6d1ddae50, 
    surface=0x0, keycodes=<optimized out>, num_keycodes=<optimized out>, 
    modifiers=0x0) at ../wlroots-hidpi-xprop/types/seat/wlr_seat_keyboard.c:294
#8  0x000056d6c53d9bb5 in seat_focus (seat=0x7ffc0a777658, surface=0x0, 
    is_lock_surface=false) at ../labwc/src/seat.c:667
#9  seat_focus_surface (seat=0x7ffc0a777658, surface=0x0)
    at ../labwc/src/seat.c:711
[...]

PR #2338, opened 2024-11-13
view: ensure output is usable before setting adaptive sync

Consolatis:
Fixes
- #2337

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -1048,6 +1048,8 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
 void
 output_enable_adaptive_sync(struct output *output, bool enabled)
 {
+	assert(output_is_usable(output));
+
 	wlr_output_state_set_adaptive_sync_enabled(&output->pending, enabled);
 	if (!wlr_output_test_state(output->wlr_output, &output->pending)) {
 		wlr_output_state_set_adaptive_sync_enabled(&output->pending, false);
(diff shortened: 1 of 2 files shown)


Consolatis:
Also needs to be backported to v0.7 which doesn't check `view->output` either.

johanmalm:
LGTM. @kode54 I'd be good to get your verification if you've got time.

kode54:
@johanmalm Verified working. Thanks again!

jlindgren90:
This causes an `assert` fail and crash for me, when enabling an output that was previously disabled.
Partial backtrace (line numbers are off, it's my `dev` branch):

```
#4  0x00007087fd644177 in __assert_fail () at /usr/lib/libc.so.6
#5  0x00006525829f83a8 in output_enable_adaptive_sync (output=<optimized out>, enabled=<optimized out>)
    at ../src/output.c:1092
#6  0x00006525829f9582 in output_config_apply (server=0x7ffcce1d9740, config=0x652588361ca0) at ../src/output.c:586
#7  handle_output_manager_apply (listener=0x7ffcce1d9ee0, data=0x652588361ca0) at ../src/output.c:728
#8  0x00007087fe67747e in wl_signal_emit_mutable () at /usr/lib/libwayland-server.so.0
```

It looks like when the output is in the process of being enabled, but the new state has not been committed yet, `wlr_output->enabled` is still `false`. So it's not safe to `assert` that it's `true` in `output_enable_adaptive_sync`.

Consolatis:
Somehow missed  the pre-initial commit code path. The backport in #2347 is not using the assert() so that one should be safe.
Thanks for fixing

PR #2339, opened 2024-11-13
menu: invalidate nested duplicated menus

tokyo4j:
Prior to this PR, nesting the same menus (e.g. `<menu id="foo"><menu id="foo" /></menu>`) caused stack overflow at `close_all_submenus()` when trying to open it.


Consolatis:
LGTM, not tested.

PR #2340, opened 2024-11-13
menu: fix UAF of server->menu_current

tokyo4j:
This fixes segfault when exiting with a menu opened, which is a regression from eaf11fac.


Consolatis:
:)

PR #2344, opened 2024-11-13
menu: fix pipemenu bug causing parent menu to disappear

johanmalm:
...introduced by 7651531632a345d65737e631c484b2e20d05a988

Only post-process the pipemenu instead of processing all menus because post_processing() calls menu_update_scene() which is a bit heavier than its predecessor and calls wlr_scene_node_set_enabled() which disables existing menus.

@tokyo4j - would you mind reviewing in case I'm missing something here.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -368,6 +368,21 @@ menu_update_scene(struct menu *menu)
 }
 
 static void
+post_processing_from(struct server *server, struct menu *from)
+{
+	bool should_process = false;
+	struct menu *menu;
+	wl_list_for_each(menu, &server->menus, link) {
+		if (menu == from) {
+			should_process = true;
+		}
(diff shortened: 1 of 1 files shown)


tokyo4j:
> which is a bit heavier than its predecessor and calls wlr_scene_node_set_enabled() which disables existing menus

From my testing, the cause of the bug is that `menu_update_scene()` **destroys** the menu scene-tree and recreates it, so the position of the parent of the pipemenu always goes to (0,0) and the pipemenu is incorrectly positioned.

Your solution will work, but I opened #2346 as my preference. It guards `menu_update_scene()` with `if (!menu->scene_tree)` instead of starting iteration the pipemenu onward.

tokyo4j:
Superseded by #2346 (sorry if its PR title was confusing).

PR #2345, opened 2024-11-13
placement: set default placement policy to `cascade`

tokyo4j:
I've been using `cascade` placement policy for several months and it's been working very well. It avoids the perfect overlap with `center` policy when user opens multiple windows consecutively, but its behavior is more predictable than `automatic` policy in my opinion.

I think we can provide better out-of-the-box experience to users by setting this policy to the default value.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -232,7 +232,7 @@ this is for compatibility with Openbox.
 	center new windows under the cursor. The "cascade" policy will try to
 	place new windows at the center of the active output, but possibly
 	shifts its position to bottom-right not to cover existing windows.
-	Default is "center".
+	Default is "cascade".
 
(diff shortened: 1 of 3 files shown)


Consolatis:
I have no opinion in this regard.

CC @ahesford @jlindgren90 @johanmalm

johanmalm:
I agree with this. I actually had it on my "TODO" list to propose changing the default to this :smile:

jlindgren90:
Sounds good to me.

Consolatis:
Looks like there is an agreement regarding changing the default, merging.

> The little confirmation prompt gets pushed outside or to the edge of taskmanager's window with automatic

That sounds like we either don't check if the confirmation prompt is a child of the task manager or xfce4-taskmanager isn't setting the parent of the dialog correctly.

PR #2346, opened 2024-11-13
menu: fix incorrect positioning of pipemenus

tokyo4j:
7651531 introduced a regression: `menu_update_scene()` which re-creates a menu scene was called for all the menus when a pipemenu is created, so the menus (parent of the pipemenu) were always moved to (0,0) and hidden, and the pipemenu was incorrectly positioned.

This PR fixes it by calling `menu_update_scene()` only for the pipemenu when it's created.

The second commit just eliminates dead code in `menu_hide_submenu()` I added in the prior PR by mistake.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -322,6 +322,7 @@ title_create_scene(struct menuitem *menuitem, int *item_y)
 	*item_y += theme->menu_header_height;
 }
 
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. LGTM. Not tested.

PR #2350, opened 2024-11-14
menu: fix UAFs in menu_destroy() and item_destroy()

tokyo4j:
This fixes use-after-free when there's only 1 desktop and `menu_hide_submenu()` is called to delete "Workspaces" submenu in client-menu before menu scenes are initialized.

As `menu_create()` and `item_create()` no longer initialize scenes after 76515316, `menu->scene_tree` and `item->tree` should be null-checked.

Sorry for another regression!

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -437,7 +437,9 @@ item_destroy(struct menuitem *item)
 	}
 	wl_list_remove(&item->link);
 	action_list_free(&item->actions);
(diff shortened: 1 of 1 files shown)


Consolatis:
Didn't test but code LGTM and makes sense when the scene nodes are not created automatically anymore.

> Sorry for another regression!

No worries, it's the master branch after all. Thanks for detecting and fixing.

PR #2352, opened 2024-11-14
menu: [wip] support borders, rounded corners and padding around items

tokyo4j:
Following theme configurations are added:

```
menu.padding.width: 0
menu.padding.height: 0
menu.corner-radius: 0
menu.border.width: 1
menu.border.color: #a8a5a2
menu.bg.color: #fcfbfa
menu.items.corner-radius: 0
menu.items.border.width: 0
menu.items.border.color: #fcfbfa
menu.items.active.border.color: #a8a5a2
```

- Changes to the default look are:
  - gray boders around menus
  - gray borders at top & bottom of titles. This follows what Openbox does.
- The rounded corners and borders are implemented with the new `scaled_rect_buffer`.
- The scaled-rect-buffers for unselected items are created as well as selected ones, but they are invisible by default, by setting `menu.items.border.color` and `menu.items.bg.color` the same as `menu.bg.color`. I wonder if we need this, but I added it because Openbox allows setting borders around unselected items with `border` texture.
- As Openbox allows setting border colors around items by applying `border` texture to their background, I allowed `menu.items(.active).bg.border.color` as aliases to `menu.items(.active).border.color`
- I'm not sure about `menu.padding.{width,height}` vs `menu.padding.{x,y}`. We're in a weird situation where we have both `menu.separator.padding.width` and `menu.items.padding.x`, so I think we should choose either of width/height or x/y and rename those existing entries.
- The separator title as the topmost item in a menu is a bit special, as it's top edge is rounded and not stroked.

- [ ] Add documentation

Example:

20241114_19h22m19s_grim

<details><summary>themerc</summary>

```
menu.corner-radius: 5
menu.padding.width: 3
menu.padding.height: 3
menu.items.active.border.color: #10acfa
menu.items.active.bg.color: #c3ebff
menu.items.corner-radius: 5
menu.title.bg.color: #d2d2d2
menu.title.text.color: #000000
```

</details>


Consolatis:
The usage of the scaled-scene-buffer API looks good to me. Didn't check cairo and menu.

Commit message states that docs aren't added yet but just to have it mentioned, IMHO `docs/themerc` should contain all default values (similar to `docs/rc.xml.all`).

One thing I noticed on a quick test: `menu.border.color` isn't inherited from `border.color` which I personally would expect to happen. I didn't check what Openbox does.

Overall: great work!

tokyo4j:
> One thing I noticed on a quick test: menu.border.color isn't inherited from border.color which I personally would expect to happen. I didn't check what Openbox does.

Your suggested behavior can be achieved by letting the border color of menus inherit `window.active.border.color` and this is what Openbox does. However, we set the default value of `window.active.border.color` to white (`#e1dedb`), so border will be invisible by default if we do that. Perhaps we can inherit `window.active.label.text.color` just like what we're doing for `osd.border.color`, but it's dark black and I wonder if it's suitable for default look.

> Just thinking longer term: with this PR it seems as if the background is re-created for every menu item. As all menu items (within a menu) share the same size and attributes (like color, radius and so on) and there are only 3 different variants (correct me if wrong, not looked in detail in menu.c) it should be enough to create at most that many scaled-scene-rects per menu and then re-use the wlr_buffers for all menu items.

You're right. The current implementation in this PR duplicates many buffers.

Your sketch branch looks nice. Maybe I'll open another PR that only adds scaled-rect-buffer incorporating your modification later.

tokyo4j:
@Consolatis Could you have a look at my branch? https://github.com/labwc/labwc/compare/master...tokyo4j:labwc:scaled-scene-buffer-sharing

I added `bool equal(struct scaled_scene_buffer *, struct scaled_scene_buffer *)` to `scaled_scene_buffer_impl` and moved the buffer sharing mechanism to `scaled-scene-buffer.c`. I applied it for scaled-font-buffer, and I had to remove `width` and `height` fields from `scaled_font_buffer` since `create_buffer()` is no longer always called and `width` and `height` may not be set.

Consolatis:
I really like that approach! Especially with having the layer 2 cache in the impl and under full control of `wlr-scene-buffer.c`. Will play around with it a bit tonight. If we end up implementing this for ssd buttons we could even use some button enum + hover bool + state and wouldn't have to compare everything. Similar story for the window icons, just using the app_id should already be unique.

Consolatis:
Had a chance to play around with your branch a bit: https://github.com/labwc/labwc/compare/master...Consolatis:labwc:sketch/scaled_rect_buffer_reuse_tokyo4j
Overall I think its a great design, do you want to open a PR for it?

I've removed:
3e7eb358f72e4df9603e5484a330cbeab341ea8f, 64a3e0e07fd6de452bb5731a9f1bf44bd775d452 and 5e1b6beefa9ed6b0293e05ae88250cd64ac63b21 (alternative approach in dc3bc53ec21ede300ff05239bd4b53b50261775d).

I've also added a few minor fixups in 96dc7c4532a7d0aa6b75ee6d0f9ae1f65eec50ff.
When doing a PR we should maybe split the changes to font-buffer.c from the main commit into an own one.

tokyo4j:
>  Overall I think its a great design, do you want to open a PR for it?

OK, I'll open a PR.

Consolatis:
> One thing I'm worried with this approach is that, if we apply scaled-scene-buffer to window decorations, the buffers are released when all the window are closed and recreated every time the first window is opened. Maybe this is not much of a problem though, considering that we are always recreating buffer for window icons. Or maybe we can set drop_buffer to false for the scaled-scene-buffers for window decoration parts and drop them all on reconfigure or exit.

Yes, I've noticed that too. And it goes further than that: If the first window closes it will remove the cache entry along with it. Thus when then spawing a new window it wouldn't find the old cache entry anymore (even though its wlr_buffer is used by other windows). One option could be to create a new local cache item per scene buffer when reusing the wlr_buffer from another scene buffer. That should take care of the issue when the original owner of the cache_entry gets destroyed. But it would slightly increase the lookup time because there is more cache items to go through.

Recreating the window decorations itself when there are no windows open seems fine to me, it is a trade-off of complexity in the code base vs a (usual) 1 time operation when spawing the first window.

tokyo4j:
Maybe we can cache handlers for images in a lower layer than scaled-scene-buffers. In my quick benchmark, `load_buttons()` took 18ms when I placed 8 svg icons in the theme directory, and its flamegraph from `perf` showed that `rsvg_handle_new_from_file()` accounted for more than half of the loading time besides `rsvg_handle_render_document()`. So I think when we implement `scaled-icon-buffer`, we can replace the icon buffers in `struct theme` with handlers for icon images so that the loading time of svg images is halved.

Consolatis:
Sorry for derailing this PR so much with the cache discussion.

Appreciate the new PR and I think it makes it easier to review.

Consolatis:
What is left in this PR, mostly rounded corners?

tokyo4j:
> What is left in this PR, mostly rounded corners?

#2394 is the blocking issue for this PR. As I said in https://github.com/labwc/labwc/pull/2241#issuecomment-2550328298, the issue will be fixed if we draw menu borders only with wlr_scene_buffers, but it means this PR will be abandoned. It may look trivial, but I think we should address it first before adding rounded corners which depend on the cairo-based buffers.

TheESN:
What's the status of this PR right now?

tokyo4j:
> What's the status of this PR right now?

It's abandoned now. #2394 is very hard to avoid if we draw curved strokes for menus. Also, rounded corners of the top/bottom items, labels and separators is too annoying to implement and maintain.

PR #2353, opened 2024-11-16
Fix some small config related memory leaks

Consolatis:
We didn't free `keybind->keysyms` on
- keybind de-duplication
- keybind removal due to empty action list

This PR adds a `keybind_destroy()` function which takes care of properly freeing the keybind.

---
For some `char *` vars in `struct rc` we were using `= xstrdup()` without freeing the previous value. This caused a mem leak with a config like
```xml
<theme>
  <name>Numix</name>
</theme>
<theme name="Numix" />
```

This PR adds a new `xstrdup_replace(char *ptr, const char *val)` helper which does a `free(ptr); ptr = xstrdup(val)`.

---
Also thrown in a  small fix to remove line terminators from some log messages.

### DIFF
--- a/include/common/mem.h
+++ b/include/common/mem.h
@@ -45,6 +45,14 @@ void *xrealloc(void *ptr, size_t size);
 char *xstrdup(const char *str);
 
 /*
+ * Same as ptr = xstrdup(str) but free
+ * <ptr> before assigning the result.
+ */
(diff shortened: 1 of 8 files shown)


johanmalm:
Nice!
I had got as far as detecting the keysym leaks, but not managed to translate it into a fix yet :smile:

PR #2356, opened 2024-11-16
theme: allow negative values for menu overlap

### LINKED ISSUE #2294: menu.overlap.x does not work with negative values
Hello!
Thank you very much for your hard work and great project.

I have this problem in **labwc 0.8.1**

When using in **themerc-override**
`menu.overlap.x: 5`
the submenu moves to the left and overlaps the main menu,

and with negative values
`menu.overlap.x: -5`
nothing happens at all, although it should be shifted to the right.

P.S.
I would like you to add menu border parameters (like in Openbox).
`menu.border.width`
`menu.border.color`
because without border menu often merges with the background.

jp7677:
This fixes a regression introduced with:
https://github.com/labwc/labwc/commit/dcd9b47e5b584a2559d41cd255bb043f11970d12

A weak moment, missed that those can be negative while adding the checks ;)

Fixes #2294

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -851,12 +851,10 @@ entry(struct theme *theme, const char *key, const char *value)
 	}
 
 	if (match_glob(key, "menu.overlap.x")) {
-		theme->menu_overlap_x = get_int_if_positive(
-			value, "menu.overlap.x");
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks.

PR #2360, opened 2024-11-17
feat: add libsfdo desktop entry name as OSD element

jp7677:


src/sfdo.c
```diff
@@ -210,11 +211,19 @@ process_abs_name(struct icon_ctx *ctx, const char *icon_name)
 static struct sfdo_desktop_entry *
 get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 {
+	struct sfdo_desktop_entry *entry;
+
+	/* try direct match first */
+	entry = sfdo_desktop_db_get_entry_by_id(db, app_id, SFDO_NT);
```
jp7677, line 217: This is still OK for icon look up?

include/sfdo.h
```diff
@@ -0,0 +1,14 @@
...
+#ifndef LABWC_SFDO_H
+#define LABWC_SFDO_H
+
+struct server;
+
+void sfdo_init(struct server *server);
+void sfdo_finish(struct server *server);
+struct lab_data_buffer *sfdo_icon_lookup(struct server *server,
+	const char *app_id, int size, float scale);
+const char *sfdo_desktop_entry_name_lockup(struct server *server,
```
Consolatis, line 11: We can't (or rather: should not) use the `sfdo_` prefix because it belongs to libsfdo and might clash in the future.

Consolatis (commented):
We should likely have a generic `get_desktop_entry()` function within the loader which does all the weird matching (app_id, lowercase(app_id), wm_class match, partial match, ..). That function can then be used by both, the name and the icon lookup.

Regarding the feature itself I don't really have an opinion.

src/sfdo.c
Consolatis: Its doing duplicated work in case of icon lookup, suggest something like a generic `get_desktop_entry()` (or whatever its named) function within the loader which does all the weird matching (app_id, lowercase(app_id), wm_class match, partial match, ..). That function can then be used by both, the name and the icon lookup.

include/sfdo.h
jp7677: Ah yeah, good point. Do you have a suggestion? What comes to my mind is `desktop_entry` or `desktop_entries_db`, but I’m not convinced that this is the best fit.

src/sfdo.c
jp7677: Thx, I had indeed missed that this function is already used for icon look up.

include/sfdo.h
Consolatis: Can't come up with anything better than the `desktop_entry_` prefix either right now.

tokyo4j: In addition, I think we should add a comment about the lifetime of returned value. In this case, its lifetime will be the same as that of `sfdo_desktop_db`.

Also, there's a typo: `s/lockup/lookup/`.

jp7677: Thanks, addressed. I wasn't sure about the new name of the `struct icon_loader`. Using `desktop_entry` looks a bit weird, so I kept that as `sfdo`. Please let me know if you prefer `desktop_entry` of a different name instead.

Consolatis: `struct sfdo` sounds fine to me.

Consolatis:
Thanks, code LGTM and didn't see issues while testing (only tested custom `%n` format with and without libsfdo enabled though).
I don't want to overrule @johanmalm with a decision of the feature itself so not merging right away.

johanmalm:
Thanks. I’m good with feature. Had been thinkink that this was the way to go for a long time… probably since implementing that “trimmed” hack.

johanmalm:
In fact, we ought to use the by default, particularly as it falls back on the trimmed version.

Consolatis:
Do we want to do this in this PR or as a follow up?

include/desktop-entry.h
```diff
@@ -0,0 +1,23 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_ICON_LOADER_H
+#define LABWC_ICON_LOADER_H
```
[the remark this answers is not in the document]
jp7677: Thanks, fixed.

johanmalm:
> Do we want to do this in this PR or as a follow up?

I'm easy. Happy to change default later.

johanmalm:
Thanks.

PR #2363, opened 2024-11-18
scaled-scene-buffer: implement buffer sharing mechanism

tokyo4j:
Suggested in #2352, combined with @Consolatis's work.

This PR adds `equal()` API in scaled-scene-buffer which can be used to share buffers for visually duplicated elements (e.g. scaled-font-buffer with the same text, font color and size). This PR itself isn't much beneficial as most `scaled-font-buffer`s are not duplicated, but this will greatly reduce the increased memory usage in #2352 as `scaled-rect-buffer`s used for backgrounds of menu and menu items will be visually the same.

### DIFF
--- a/include/common/scaled-scene-buffer.h
+++ b/include/common/scaled-scene-buffer.h
@@ -17,6 +17,9 @@ struct scaled_scene_buffer_impl {
 		(struct scaled_scene_buffer *scaled_buffer, double scale);
 	/* Might be NULL or used for cleaning up */
 	void (*destroy)(struct scaled_scene_buffer *scaled_buffer);
+	/* Returns true if the two buffers are visually the same */
+	bool (*equal)(struct scaled_scene_buffer *scaled_buffer_a,
+		struct scaled_scene_buffer *scaled_buffer_b);
 };
 
 struct scaled_scene_buffer {
@@ -28,11 +31,18 @@ struct scaled_scene_buffer {
 	/* Private */
 	bool drop_buffer;
 	double active_scale;
+	/* cached wlr_buffers for each scale */
 	struct wl_list cache;  /* struct scaled_buffer_cache_entry.link */
 	struct wl_listener destroy;
 	struct wl_listener output_enter;
 	struct wl_listener output_leave;
 	const struct scaled_scene_buffer_impl *impl;
+	/*
+	 * Pointer to the per-implementation list of scaled-scene-buffers.
+	 * This is used to share the backing wlr_buffers.
+	 */
+	struct wl_list *cached_buffers;
+	struct wl_list link; /* struct scaled_scene_buffer.cached_buffers */
 };
 
 /**
@@ -52,6 +62,11 @@ struct scaled_scene_buffer {
  * wlr_scene_buffer is being destroyed. If implementation->destroy is set
  * it will also get called so a consumer of this API may clean up its own
  * allocations.
+ *
+ * Besides caching buffers for each scale per scaled_scene_buffer, we also
+ * store all the scaled_scene_buffers in a per-implementer list passed as
+ * @cached_buffers in order to reuse backing buffers for visually duplicated
+ * scaled_scene_buffers found via impl->equal().
  *
  * All requested lab_data_buffers via impl->create_buffer() will be locked
  * during the lifetime of the buffer in the internal cache and unlocked
@@ -79,7 +94,7 @@ struct scaled_scene_buffer {
 struct scaled_scene_buffer *scaled_scene_buffer_create(
 	struct wlr_scene_tree *parent,
 	const struct scaled_scene_buffer_impl *implementation,
-	bool drop_buffer);
+	struct wl_list *cached_buffers, bool drop_buffer);
 
 /* Clear the cache of existing buffers, useful in case the content changes */
 void scaled_scene_buffer_invalidate_cache(struct scaled_scene_buffer *self);
--- a/src/common/scaled-font-buffer.c
+++ b/src/common/scaled-font-buffer.c
@@ -7,9 +7,13 @@
 #include <wlr/util/log.h>
 #include "buffer.h"
 #include "common/font.h"
+#include "common/list.h"
 #include "common/mem.h"
 #include "common/scaled-scene-buffer.h"
 #include "common/scaled-font-buffer.h"
+
+/* This holds all the scaled-font-buffers and used for sharing backing buffers */
+static struct wl_list cached_buffers = WL_LIST_INIT(&cached_buffers);
 
 static struct lab_data_buffer *
 _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
@@ -25,8 +29,6 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
 		wlr_log(WLR_ERROR, "font_buffer_create() failed");
 	}
 
-	self->width = buffer ? buffer->logical_width : 0;
-	self->height = buffer ? buffer->logical_height : 0;
 	return buffer;
 }
 
@@ -42,9 +44,33 @@ _destroy(struct scaled_scene_buffer *scaled_buffer)
 	free(self);
 }
 
+static bool str_equal(const char *a, const char *b)
+{
+	return a == b || (a && b && !strcmp(a, b));
+}
+
+static bool
+_equal(struct scaled_scene_buffer *scaled_buffer_a,
+	struct scaled_scene_buffer *scaled_buffer_b)
+{
+	struct scaled_font_buffer *a = scaled_buffer_a->data;
+	struct scaled_font_buffer *b = scaled_buffer_b->data;
+
+	return str_equal(a->text, b->text)
+		&& a->max_width == b->max_width
+		&& str_equal(a->font.name, b->font.name)
+		&& a->font.size == b->font.size
+		&& a->font.slant == b->font.slant
+		&& a->font.weight == b->font.weight
+		&& !memcmp(a->color, b->color, sizeof(a->color))
+		&& !memcmp(a->bg_color, b->bg_color, sizeof(a->bg_color))
+		&& str_equal(a->arrow, b->arrow);
+}
+
 static const struct scaled_scene_buffer_impl impl = {
 	.create_buffer = _create_buffer,
-	.destroy = _destroy
+	.destroy = _destroy,
+	.equal = _equal,
 };
 
 /* Public API */
@@ -54,7 +80,7 @@ scaled_font_buffer_create(struct wlr_scene_tree *parent)
 	assert(parent);
 	struct scaled_font_buffer *self = znew(*self);
 	struct scaled_scene_buffer *scaled_buffer = scaled_scene_buffer_create(
-		parent, &impl, NULL, /* drop_buffer */ true);
+		parent, &impl, &cached_buffers, /* drop_buffer */ true);
 	if (!scaled_buffer) {
 		free(self);
 		return NULL;
@@ -96,6 +122,10 @@ scaled_font_buffer_update(struct scaled_font_buffer *self, const char *text,
 
 	/* Invalidate cache and force a new render */
 	scaled_scene_buffer_invalidate_cache(self->scaled_buffer);
+
+	/* Ensure the height / width is up-to-date */
+	self->width = self->scaled_buffer->width;
+	self->height = self->scaled_buffer->height;
 }
 
 void
@@ -103,4 +133,8 @@ scaled_font_buffer_set_max_width(struct scaled_font_buffer *self, int max_width)
 {
 	self->max_width = max_width;
 	scaled_scene_buffer_invalidate_cache(self->scaled_buffer);
+
+	/* Ensure the height / width is up-to-date */
+	self->width = self->scaled_buffer->width;
+	self->height = self->scaled_buffer->height;
 }
--- a/src/common/scaled-font-buffer.c
+++ b/src/common/scaled-font-buffer.c
@@ -53,8 +53,8 @@ scaled_font_buffer_create(struct wlr_scene_tree *parent)
 {
 	assert(parent);
 	struct scaled_font_buffer *self = znew(*self);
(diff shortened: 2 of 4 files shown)


src/common/scaled-font-buffer.c
```diff
@@ -42,9 +40,33 @@ _destroy(struct scaled_scene_buffer *scaled_buffer)
...
+	struct scaled_font_buffer *b = scaled_buffer_b->data;
+
+	return str_equal(a->text, b->text)
+		&& a->max_width == b->max_width
+		&& str_equal(a->font.name, b->font.name)
+		&& a->font.size == b->font.size
+		&& a->font.slant == b->font.slant
+		&& a->font.weight == b->font.weight
+		&& !memcmp(a->color, b->color, sizeof(float[4]))
+		&& !memcmp(a->bg_color, b->bg_color, sizeof(float[4]))
```
Consolatis, line 62: `sizeof(a->color)` (or `b->color`) feels safer to me here in case we ever end up changing `float[4]` to something else.

src/common/scaled-scene-buffer.c
```diff
@@ -41,29 +41,71 @@ _cache_entry_destroy(struct scaled_scene_buffer_cache_entry *cache_entry, bool d
...
+			continue;
 		}
+
+		/* Ensure self->width and self->height are set correctly */
+		self->width = scene_buffer->width;
+		self->height = scene_buffer->height;
+
+		wlr_scene_buffer_set_buffer(self->scene_buffer, cache_entry->buffer);
+		wlr_scene_buffer_set_dest_size(self->scene_buffer, self->width, self->height);
+		return;
```
Consolatis, line 108: I still think we should create a local cache entry when reusing a buffer. Otherwise it will not be reused once the original owner is closed. Maybe we can just do a `struct wlr_buffer *wlr_buffer = NULL` on top, set it here, remove the `return` and then do
```c
static void
_update_buffer(struct scaled_scene_buffer *self, double scale) {
	[..]
	if (!wlr_buffer) {
		/* Create new buffer, will get destroyed along the backing wlr_buffer */
		struct lab_data_buffer *buffer = self->impl->create_buffer(self, scale);
		self->width = buffer ? buffer->logical_width : 0;
		self->height = buffer ? buffer->logical_height : 0;
		wlr_buffer = &buffer->base;
	}
	if (wlr_buffer) {
		/* Ensure the buffer doesn't get deleted behind our back */
		wlr_buffer_lock(wlr_buffer);
	}
	/* Create or reuse cache entry */
	[..]
	cache_entry->buffer = wlr_buffer;
	[..]
}
```

Consolatis (commented):
> This PR itself isn't much beneficial as most `scaled-font-buffer`s are not duplicated

If you spawn multiple windows with the same title (e.g. foot) then it will actually reuse the title. But agree that that is not the main benefit for this PR, that would be scalable window icon / button / menu background lookups.

Another pair of eyes on this PR would be good as tokyo4j and I were both involved in writing it.

include/common/scaled-scene-buffer.h
```diff
@@ -17,6 +17,10 @@ struct scaled_scene_buffer_impl {
 		(struct scaled_scene_buffer *scaled_buffer, double scale);
 	/* Might be NULL or used for cleaning up */
 	void (*destroy)(struct scaled_scene_buffer *scaled_buffer);
+	/* Returns true if the two buffers are visually the same */
+	bool (*equal)(struct scaled_scene_buffer *scaled_buffer_a,
+		struct scaled_scene_buffer *scaled_buffer_b);
+	struct wl_list cached_buffers;
```
johanmalm, line 23: We shouldn't put this list in the `impl` struct (which is in effect the method dispatch vtable).

include/common/scaled-scene-buffer.h
```diff
@@ -32,7 +36,8 @@ struct scaled_scene_buffer {
 	struct wl_listener destroy;
 	struct wl_listener output_enter;
 	struct wl_listener output_leave;
-	const struct scaled_scene_buffer_impl *impl;
+	struct scaled_scene_buffer_impl *impl;
+	struct wl_list link;
```
johanmalm, line 40: We should say what linked list this is for. Looks like `other cached scaled-scene-buffers`.

src/common/scaled-scene-buffer.c
```diff
@@ -49,21 +49,63 @@ _cache_entry_destroy(struct scaled_scene_buffer_cache_entry *cache_entry, bool d
...
+		 * If found in our local cache,
+		 * - self->width and self->height are already set
+		 * - wlr_scene_buffer_set_dest_size() has already been called
+		 */
+		return;
+	}
+
+	/* Search from other cached scaled-scene-buffers */
+	struct scaled_scene_buffer *scene_buffer;
+	wl_list_for_each(scene_buffer, &self->impl->cached_buffers, link) {
```
johanmalm, line 87: I struggle to get my head around this. Does each scaled-scene-buffer contain a list of 'other scaled scene buffers' in the impl-struct?

src/common/scaled-scene-buffer.c
```diff
@@ -177,6 +220,10 @@ scaled_scene_buffer_create(struct wlr_scene_tree *parent,
 	self->active_scale = 1;
 	self->drop_buffer = drop_buffer;
 	wl_list_init(&self->cache);
+	if (!impl->cached_buffers.next) {
```
johanmalm, line 223: It is not clear to me what this does.

What does `!impl->cached_buffers.next` mean? Couldn't it mean that the list has never been initialized OR contains items. Feels like red flag to me.

johanmalm:
I've had a read through, and have to admit that I find it hard to follow. You're two very clever people, so I feel sure there is soundness in the approach, but I would feel happier if I could snap into the thinking.

I've left a few comments to show where I get confused as a starting point for discussion.

src/common/scaled-scene-buffer.c
Consolatis: It basically just means that its calloc'd and was not initialized. An empty list points to itself for both `.prev` and `.next` and a list with items points to these items. Maybe we could add some `wl_list_maybe_init()` helper or something? Other ideas welcome, the issue here is that this is a shared list between all scaled-buffers of a given implementation (just the the scaled-font-buffer at the moment). See also my reply to the comment regarding the placement of the list itself above.

include/common/scaled-scene-buffer.h
Consolatis: An alternative approach I could think of is using a static var per impl and then defining something like `scaled_font_buffer_init()` method that has to be called from somewhere like `server.c` and then does something like
```c
static wl_list cache;
[..]
void
scaled_font_buffer_init() {
	wl_list_init(&cache);
}
```

We could then supply that `&cache` to `scaled_scene_buffer_create()` whenever the impl is creating a new scaled buffer. It would cost one more pointer per `struct scaled_scene_buffer` but would be an alternative. Although I personally think that this is slightly more code compared to storing the list in the `impl` struct and doing a lazy `wl_list_init()`.

johanmalm: I think a static variable at the start of the `scaled-scene-buffer.c` would be by far the clearest approach for me.

We could just init it in `scaled_scene_buffer_create()` with a quick check to make sure we only do it the first time.

Consolatis: The cache belongs to the implementation of a scaled-scene-buffer, e.g. only scaled-font-buffer.c for now but each new implementation might want its own cache.

 > We could just init it in scaled_scene_buffer_create() with a quick check to make sure we only do it the first time.

That was the idea with the `if (!list.next) { wl_list_init(list); }`

johanmalm: Yeah, it's just a fine line between "clear code" that someone like me (as reviewer) quickly gets - and using "tricks" to enable optimization.

johanmalm: Yep. Maybe just `init_wl_list()` :smile:

johanmalm: Or... how about just add a comment to describe this.

```
/*
 * As this is a shared list, it should only be initialized the first time.
 * When just `calloc()d` and not yet initialized, `prev`/`next` that is used to avoid
 * initizing a list in use
 */
```

Consolatis: Fine with me as well. If we ever see the need to do this in other places we can always add a helper later.

src/common/scaled-scene-buffer.c
tokyo4j: Yeah I agree. I didn't do it just because it will incur some overhead for iterating over caches of each buffers when creating the first buffer for a certain scale (`impl->equal()` always returns `true` but `find_cache_for_scale()` returns `NULL`). But the overhead will be negligible and I agree that we should not differentiate owner / borrower buffers.

src/common/scaled-scene-buffer.c
Consolatis: Each scaled-scene-buffer has an impl pointer which in turn has a single list for that implementation (e.g. scaled-font-buffer) where all scene_buffers it has ever created (and their scene nodes are still alive) are stored. This would not change if we use a static wl_list per implementation and provide a `impl->get_cache()` getter for it. Do you have a suggestion to state that more clearly in the comment?

include/common/scaled-scene-buffer.h
tokyo4j: So, is adding a comment describing the intent of `impl.cached_buffer` enough for now? I'm not sure I understood the discussion correctly...

Another approach for one-shot initialization of `wl_list` would be defining a macro like this:

```c
#define WL_LIST_INIT(head) ((struct wl_list){.prev = (head), .next = (head)})

static struct scaled_scene_buffer_impl impl = {
	.cached_buffers = WL_LIST_INIT(&impl.cached_buffers),
};
```

I personally prefer this approach as it moves the task of initialization from runtime to compile-time and  eliminates additional checks.

src/common/scaled-scene-buffer.c
tokyo4j: Exactly. I should have added more detailed comments about it.

src/common/scaled-font-buffer.c
tokyo4j: Done

src/common/scaled-scene-buffer.c
tokyo4j: Done

include/common/scaled-scene-buffer.h
tokyo4j: Added a comment:
```
	/*
	 * This holds all the created scaled-*-buffers and is referenced for
	 * sharing backing buffers
	 */
```

src/common/scaled-font-buffer.c
Consolatis: 2nd one should be `sizeof(a->bg_color)` (its `sizeof(a->color)` currently).

tokyo4j: Oops. Fixed.

include/common/scaled-scene-buffer.h
tokyo4j: OK. I moved to the static var approach and added the following line in `scaled-font-buffer.c`:
```c
static struct wl_list cached_buffers = WL_LIST_INIT(&cached_buffers);
```

include/common/scaled-scene-buffer.h
```diff
@@ -32,7 +35,13 @@ struct scaled_scene_buffer {
 	struct wl_listener destroy;
 	struct wl_listener output_enter;
 	struct wl_listener output_leave;
-	const struct scaled_scene_buffer_impl *impl;
+	struct scaled_scene_buffer_impl *impl;
```
Consolatis, line 38: Can be turned back into a `const`.

src/common/scaled-scene-buffer.c
```diff
@@ -176,7 +232,9 @@ scaled_scene_buffer_create(struct wlr_scene_tree *parent,
 	self->impl = impl;
 	self->active_scale = 1;
 	self->drop_buffer = drop_buffer;
+	self->cached_buffers = cached_buffers;
 	wl_list_init(&self->cache);
+	wl_list_insert(self->cached_buffers, &self->link);
```
Consolatis, line 237: I think we should either `assert(cached_buffers)` or check for `self->cached_buffers` in `_update_buffer()`. Personally I prefer the 2nd variant as that allows future implementations that don't want to get their things cached (e.g. they pass `NULL` as arg and don't provide `impl->equal()`).

Consolatis: Regarding the `wl_list_remove(&self->link)` in `_handle_node_destroy()`: that can be solved by either checking for `self->cached_buffers` before calling or by `wl_list_init(&self->link)` in `_create()` on `!cached_buffers`.

Or with an alternative: a pattern often used in wayland protocol implementations within wlroots. E.g. if you have a `struct wl_list` which is either part of some list or not it can always be removed via `wl_list_remove()` if it has been initialized after removal.

tokyo4j: I agree. Added checks for `self->cached_buffers` in `scaled_scene_buffer_create()` and `_update_buffer()`.

Consolatis: `wl_list_remove(&self->link);` in `_handle_node_destroy` will crash. See above

tokyo4j: Sorry, I missed your second comment. I copied your suggested code.

Consolatis: I thought so thus the 2nd reply :)

Consolatis:
Added some buffer tracking and tested: no issues detected. The title buffers are correctly reused when spawning a 2nd foot window. When then closing the first one and opening a third foot window it now reuses the buffer from the 2nd instance. Everything seems to work as expected. Code LGTM as well. Great work!

@johanmalm this is ready from my point of view but please say if there is anything missing or should be documented better.

johanmalm:
LGTM. This is a great piece of work laying some very useful building blocks for the future.

PR #2364, opened 2024-11-18
src/menu.c: remove unnecessary re-renderings of font buffers

tokyo4j:
Reported in https://github.com/labwc/labwc/pull/2336#discussion_r1845286237

I missed that `scaled_font_buffer_update()` has `max_width` argument and the subsequent calls to `scaled_font_buffer_set_max_width()` caused unnecessary re-renderings.


Consolatis:
Thanks :)

PR #2365, opened 2024-11-19
Protocol implementation: ext-workspaces

Consolatis:
Minor final touches missing from the implementation but should be fine to test clients against.
The first two commits are prep work for supporting both, cosmic-workspaces and ext-workspaces and will be split into an own PR so this one is easier to review.

~~Note: the [protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40) is still not merged so it may be changed further. Do not use this protocol in non-dev builds or without hiding it behind feature flags.~~

CC @LBCrion in case you want to start playing with an implementation in sfwbar and need a compositor to test against.

### DIFF
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -16,11 +16,13 @@
 #include "input/keyboard.h"
 #include "labwc.h"
 #include "protocols/cosmic-workspaces.h"
+#include "protocols/ext-workspace.h"
 #include "view.h"
 #include "workspaces.h"
 #include "xwayland.h"
 
 #define COSMIC_WORKSPACES_VERSION 1
+#define EXT_WORKSPACES_VERSION 1
 
 /* Internal helpers */
 static size_t
@@ -185,6 +187,15 @@ handle_cosmic_workspace_activate(struct wl_listener *listener, void *data)
 	wlr_log(WLR_INFO, "cosmic activating workspace %s", workspace->name);
 }
 
+/* ext workspace handlers */
+static void
+handle_ext_workspace_activate(struct wl_listener *listener, void *data)
+{
+	struct workspace *workspace = wl_container_of(listener, workspace, on_ext.activate);
+	workspaces_switch_to(workspace, /* update_focus */ true);
+	wlr_log(WLR_INFO, "ext activating workspace %s", workspace->name);
+}
+
 /* Internal API */
 static void
 add_workspace(struct server *server, const char *name)
@@ -210,6 +221,17 @@ add_workspace(struct server *server, const char *name)
 	workspace->on_cosmic.activate.notify = handle_cosmic_workspace_activate;
 	wl_signal_add(&workspace->cosmic_workspace->events.activate,
 		&workspace->on_cosmic.activate);
+
+	/* ext */
+	workspace->ext_workspace = lab_ext_workspace_create(
+		server->workspaces.ext_manager, /*id*/ NULL);
+	lab_ext_workspace_assign_to_group(workspace->ext_workspace, server->workspaces.ext_group);
+	lab_ext_workspace_set_name(workspace->ext_workspace, name);
+	lab_ext_workspace_set_active(workspace->ext_workspace, active);
+
+	workspace->on_ext.activate.notify = handle_ext_workspace_activate;
+	wl_signal_add(&workspace->ext_workspace->events.activate,
+		&workspace->on_ext.activate);
 }
 
 static struct workspace *
@@ -288,8 +310,15 @@ workspaces_init(struct server *server)
 		server->wl_display, /* capabilities */ CW_CAP_WS_ACTIVATE,
 		COSMIC_WORKSPACES_VERSION);
 
+	server->workspaces.ext_manager = lab_ext_workspace_manager_create(
+		server->wl_display, /* capabilities */ WS_CAP_WS_ACTIVATE,
+		EXT_WORKSPACES_VERSION);
+
 	server->workspaces.cosmic_group = lab_cosmic_workspace_group_create(
 		server->workspaces.cosmic_manager);
+
+	server->workspaces.ext_group = lab_ext_workspace_group_create(
+		server->workspaces.ext_manager);
 
 	wl_list_init(&server->workspaces.all);
 
@@ -319,6 +348,8 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 
 	lab_cosmic_workspace_set_active(
 		server->workspaces.current->cosmic_workspace, false);
+	lab_ext_workspace_set_active(
+		server->workspaces.current->ext_workspace, false);
 
 	/* Move Omnipresent views to new workspace */
 	struct view *view;
@@ -374,6 +405,7 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	desktop_update_top_layer_visiblity(server);
 
 	lab_cosmic_workspace_set_active(target->cosmic_workspace, true);
+	lab_ext_workspace_set_active(target->ext_workspace, true);
 }
 
 void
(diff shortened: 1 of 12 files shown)


src/protocols/cosmic_workspaces/cosmic-workspaces.c
```diff
@@ -212,17 +213,27 @@ workspace_set_state(struct lab_cosmic_workspace *workspace,
...
+	struct lab_wl_resource_addon *addon = wl_resource_get_user_data(resource);
 	if (!addon) {
 		return;
 	}
 
 	struct lab_cosmic_workspace_group *group = addon->data;
-	transaction_add_workspace_group_ev(group, resource, CW_PENDING_WS_CREATE, name);
+	struct lab_transaction *transaction = lab_transaction_add(
+		addon->ctx, CW_PENDING_WS_CREATE, group, xstrdup(name));
+	wl_signal_add(&transaction->events.destroy, &group->on.transaction_destroy);
```
Consolatis, line 236: Hrm. This will leak when there are multiple create-workspace transactions within a single `commit()`. Maybe revert the whole destroy handler for the transaction idea and instead go back to manually `free(->data)` in `manager_handle_commit()`. Same issue for cosmic-workspaces.

src/workspaces.c
```diff
@@ -319,6 +350,11 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	lab_cosmic_workspace_set_hidden(
 		server->workspaces.current->cosmic_workspace, true);
 
+	lab_ext_workspace_set_active(
+		server->workspaces.current->ext_workspace, false);
+	lab_ext_workspace_set_hidden(
+		server->workspaces.current->ext_workspace, true);
```
Consolatis, line 356: The integration of both, cosmic and ext are likely using the the `hidden` flag wrong. It appears to mean something like `disabled` instead which is a concept labwc doesn't support. We should likely simply remove both calls.

include/protocols/ext-workspaces.h
```diff
@@ -0,0 +1,99 @@
...
+void lab_ext_workspace_group_destroy(struct lab_ext_workspace_group *group);
+
+struct lab_ext_workspace *lab_ext_workspace_create(struct lab_ext_workspace_group *group);
+void lab_ext_workspace_set_name(struct lab_ext_workspace *workspace, const char *name);
+void lab_ext_workspace_set_active(struct lab_ext_workspace *workspace, bool enabled);
+void lab_ext_workspace_set_urgent(struct lab_ext_workspace *workspace, bool enabled);
+void lab_ext_workspace_set_hidden(struct lab_ext_workspace *workspace, bool enabled);
+void lab_ext_workspace_set_coordinates(struct lab_ext_workspace *workspace,
+	struct wl_array *coordinates);
+void lab_ext_workspace_destroy(struct lab_ext_workspace *workspace);
```
Consolatis, line 97: Misses an `lab_ext_workspace_assign_to_group()` (even though the labwc integration will never use it).

Consolatis:
> Wow! There is a lot of work in here. Do you want me to review now - or wait until the protocol is upstreamed?

I'll open a new PR with the transaction abstraction, might be easiest to start there. The rest of this PR is more or less a copy of the cosmic-workspace implementation.

LBCrion:
I merged initial support for ext-workspace into the latest git version of
sfwbar, but couldn't test it c with this PR so far. I'm getting a segfault
initializing a layer shell window via gtk layer shell.  Sounds like an
issue unrelated to this PR.

Consolatis:
> I merged initial support for ext-workspace into the latest git version of
sfwbar

Nice!

> I'm getting a segfault initializing a layer shell window via gtk layer shell.

In sfwbar or in labwc?

LBCrion:
The segfault is in sfwbar, in a call to gtk_layer_init_for_window. I only
get this crash in the git version of labwc wip/ext-workspace branch. Stable
version of labwc works fine. The crash occurs before sfwbar even gets to
initializing ext-workspace related code.

Consolatis:
> The segfault is in sfwbar, in a call to gtk_layer_init_for_window. I only get this crash in the git version of labwc wip/ext-workspace branch. Stable version of labwc works fine. The crash occurs before sfwbar even gets to initializing ext-workspace related code.

When running with `WAYLAND_DEBUG=1` the last message I see is an information about a new workspace. This seems to make things work:
```diff
diff --git a/src/ipc/ext-workspace.c b/src/ipc/ext-workspace.c
index 5dfc5b4..d8f1b18 100644
--- a/src/ipc/ext-workspace.c
+++ b/src/ipc/ext-workspace.c
@@ -201,6 +201,13 @@ static void ew_workspace_manager_handle_workspace_group(void *data,
     workspace_group, &ew_workspace_group_impl, NULL);
 }
 
+static void ew_workspace_manager_handle_new_workspace(void *data,
+    struct ext_workspace_manager_v1 *workspace_manager,
+    struct ext_workspace_handle_v1 *workspace)
+{
+  g_warning("New workspace");
+}
+
 static void ew_workspace_manager_handle_done(void *data,
     struct ext_workspace_manager_v1 *workspace_manager)
 {
@@ -218,6 +225,7 @@ static const struct ext_workspace_manager_v1_listener
   workspace_manager_impl =
 {
   .workspace_group = ew_workspace_manager_handle_workspace_group,
+  .workspace = ew_workspace_manager_handle_new_workspace,
   .done = ew_workspace_manager_handle_done,
   .finished = ew_workspace_manager_handle_finished,
 };
```

I do get misaligned address errors though when compiling sfwbar with `meson configure -Db_sanitize=address,undefined`.

LBCrion:
Indeed, this is the cause. ext-workspace handles the new workspaces via
.workspace rather than .workspace_enter in the group. Should be fixed in
the latest git and it seems to work correctly with the labwc PR. Odd that
the backtrace showed the crash long before we get to ext-workspace.

Consolatis:
Sorry, I know this is a giant commit but I can't figure out how to logically split it. Everything that is unrelated was already split off and was merged into master separately.

This still needs minor stuff:
- [ ] `assert(argument)` for the public API
- [ ] a final sanity check if all events and requests from the XML are implemented

I'll work on the two remaining things the next time. Nevertheless, this is ready for review / test from my point of view.

johanmalm:
LGTM except that `[wip]` in the commit message.

Just a couple of thoughts:
- Looks like we don't use the `assign` listener. Is that intentional?
- It is curious that the ext-workspaces protocol went with a `uint` for capabilities rather than an `array` like cosmic. Don't suspect it'll matter for a long time - if ever - but still!

johanmalm:
I should have said: I tested with `sfwbar` at `master` and that all seems to work fine (confirmed it's using ext-workspaces rather than cosmic)

Consolatis:
Thanks for the review and test.

> * Looks like we don't use the `assign` listener. Is that intentional?

Yes, that is intentional to keep the configuration in one place, e.g. in `rc.xml`. This PR is also the base for an implementation in wlroots which is why it implements everything but only hooks up the things that are relevant for labwc.

> * It is curious that the ext-workspaces protocol went with a `uint` for capabilities rather than an `array` like cosmic. Don't suspect it'll matter for a long time - if ever - but still!

That was one of the changes. Having an array for states or capabilities is a mess because those are dynamic in nature and may depend on the version supported by the client. This means that they have to be allocated for every change for every client which is extremely messy and - at least from what I've seen so far - are in most cases bitfields on the compositor side, translated into heap allocated arrays, send though the wire and then translated back into a bitfield by the client. Thus using bitfields is much easier to work with.

Consolatis:
A wlroots implementation MR is still pending.

The last pieces of the puzzle are:
- ext-foreign-toplevel-state, ext-foreign-toplevel-management ([MR](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196))
- ext-foreign-toplevel-workspaces (will likely be added to the MR above)

I did play around a bit with a ext-foreign-toplevel-list-handle <--> ext-workspace implementation and it seems to work fine. That is is not yet part of the upstream MR but we should get there eventually.

PR #2376, opened 2024-11-23
menu: support borders

tokyo4j:
Since #2352 is so large and contains many things to discuss (e.g. default values of new theme configs), I decided to create another PR that only adds support for menu borders.

This PR adds following theme configurations:
```
menu.border.width: 1
menu.border.color: #a8a5a2
```

Currently, `menu.border.width` inherits `border.width` and `menu.border.color` inherits `window.active.border.color` just like Openbox does. However, as I explained in https://github.com/labwc/labwc/pull/2352#issuecomment-2478479590, the default value of `window.active.border.color` is the same as that of `window.active.title.bg.color` and `menu.items.active.bg.color`, so menu borders can be invisible with the default configuration like this:

So, I'm thinking of setting the default value of `menu.border.color` to another constant value (for example, I set it to `#a8a5a2` in #2352). Another option is to make it inherit `menu.items.text.color` (black by default), which is similar to what we're doing for `osd.border.color`:

https://github.com/labwc/labwc/blob/c2928027becba26f058b544d4bd8c72895c8d7cf/src/theme.c#L1488-L1501

@johanmalm @Consolatis Do you have any ideas on this?

### DIFF
--- a/docs/themerc
+++ b/docs/themerc
@@ -16,8 +16,8 @@ window.titlebar.padding.width: 0
 window.titlebar.padding.height: 0
 
 # window border
-window.active.border.color: #e1dedb
-window.inactive.border.color: #f6f5f4
+window.active.border.color: #aaaaaa
+window.inactive.border.color: #aaaaaa
 
 # ToggleKeybinds status indicator
 window.active.indicator.toggled-keybind.color: #ff0000
--- a/docs/themerc
+++ b/docs/themerc
@@ -59,6 +59,8 @@ menu.overlap.x: 0
 menu.overlap.y: 0
 menu.width.min: 20
 menu.width.max: 200
+menu.border.width: 1
+menu.border.color: #aaaaaa
 menu.items.bg.color: #fcfbfa
 menu.items.text.color: #000000
 menu.items.active.bg.color: #e1dedb
--- a/include/common/scaled-rect-buffer.h
+++ b/include/common/scaled-rect-buffer.h
@@ -1,0 +1,31 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_SCALED_RECT_BUFFER_H
+#define LABWC_SCALED_RECT_BUFFER_H
+
+#include <stdint.h>
+
+struct wlr_scene_tree;
+struct wlr_scene_buffer;
+struct scaled_scene_buffer;
+
+struct scaled_rect_buffer {
+	struct wlr_scene_buffer *scene_buffer;
+	struct scaled_scene_buffer *scaled_buffer;
+	int width;
+	int height;
+	int border_width;
+	float fill_color[4];
+	float border_color[4];
+};
+
+/*
+ * Create an auto scaling borderd-rectangle buffer, providing a wlr_scene_buffer
+ * node for display. It gets destroyed automatically when the backing
+ * scaled_scene_buffer is being destroyed which in turn happens automatically
+ * when the backing wlr_scene_buffer (or one of its parents) is being destroyed.
+ */
+struct scaled_rect_buffer *scaled_rect_buffer_create(
+	struct wlr_scene_tree *parent, int width, int height, int border_width,
+	float fill_color[4], float border_color[4]);
+
+#endif /* LABWC_SCALED_RECT_BUFFER_H */
--- a/src/common/scaled-rect-buffer.c
+++ b/src/common/scaled-rect-buffer.c
@@ -1,0 +1,113 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wayland-server-core.h>
+#include <wlr/util/log.h>
+#include "buffer.h"
+#include "common/graphic-helpers.h"
+#include "common/list.h"
+#include "common/macros.h"
+#include "common/mem.h"
+#include "common/scaled-scene-buffer.h"
+#include "common/scaled-rect-buffer.h"
+
+static struct wl_list cached_buffers = WL_LIST_INIT(&cached_buffers);
+
+static void
+draw_rectangle_path(cairo_t *cairo, int width, int height, int border_width)
+{
+	double offset = border_width / 2.0;
+	double right_x = width - offset;
+	double bottom_y = height - offset;
+
+	cairo_move_to(cairo, offset, offset);
+	cairo_line_to(cairo, right_x, offset);
+	cairo_line_to(cairo, right_x, bottom_y);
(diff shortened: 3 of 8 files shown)


tokyo4j:
Another thing to mention: Openbox also draws borders around titles, but this PR currently doesn't. I think we can do it in a following PR.

johanmalm:
Deciding on default values is really tricky because:
1. Openbox defaults to black/white values which most users wouldn't want
2. Openbox does not actually let you start without a theme - and defaults to Clearlooks-3.4
3. Common Openbox themes set theme variables so that default values are not used (broad statement, I know, but useful context here I think)

I don't think there is a particularly clean way through this but suggest the following as a rule of thumb:

a. When we add options that exist in openbox theme spec, we ought to inherit as per openbox spec so that we don't break existing themes.
b. When we add options that DO NOT exist in openbox theme spec we choose some sensible color/width/etc that avoid breaking current look as far as possible. For example, if we add rounded corners, then lets set them to zero so that people have to opt-in rather than breaking current look.

This means that we don't get the best look out-of-the-box, but hey I prefer that to upsetting existing users.

There is one thing to consider here though. Our current default values contain a mix of Clearlooks and Adwaita. For example the titlebar color is grey like adwaita - but the menu labels are blue like Clearlooks. I'm happy to (and actually think we should) change the menu theme settings to align more with Adwaita but cause that blue clashes with most things.

So, in answer to your questions:

> Currently, menu.border.width inherits border.width and menu.border.color inherits window.active.border.color just like Openbox does.

Yes, I think we should stick with that.

> However, as I explained in #2352 (comment), the default value of window.active.border.color is the same as that of window.active.title.bg.color and menu.items.active.bg.color, so menu borders can be invisible with the default configuration...

I'd rather change the default value of `window.active.border.color: #e1dedb` to be different from `window.active.title.bg.color: #e1dedb` - but keen the as-per-openbox inheritence.

Suggest changing to `window.active.title.bg.color: #a8a5a2` (as per this PR :smile:) and then just inherit from it.

> Another thing to mention: Openbox also draws borders around titles, but this PR currently doesn't. I think we can do it in a following PR.

Agree.

src/common/scaled-rect-buffer.c
```diff
@@ -0,0 +1,111 @@
...
+#include "common/list.h"
+#include "common/macros.h"
+#include "common/mem.h"
+#include "common/scaled-scene-buffer.h"
+#include "common/scaled-rect-buffer.h"
+
+static struct wl_list cached_buffers = WL_LIST_INIT(&cached_buffers);
+
+static void
+draw_rectangle(cairo_t *cairo, int w, int h)
```
johanmalm, line 19: I think it would be better to pass the border thickness to this function in case we want to draw thicker borders in future - and so that the offset can be handled here. 

IIRC at a line-thickness of one we have to offset by 0.5. Otherwise cairo draws the color fainter across two pixels (wide).

Example: https://github.com/labwc/labwc/blob/c2928027becba26f058b544d4bd8c72895c8d7cf/src/theme.c#L1152

include/common/scaled-rect-buffer.h
```diff
@@ -0,0 +1,33 @@
...
+	struct wlr_scene_buffer *scene_buffer;
+	struct scaled_scene_buffer *scaled_buffer;
+	int width;
+	int height;
+	int border_width;
+	float fill_color[4];
+	float border_color[4];
+};
+
+/**
```
johanmalm, line 21: We've not been super strict on this, but when using `/**` I prefer that we stick to kernel doc format.

- https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#linux-kernel-style-basics
- https://docs.kernel.org/doc-guide/kernel-doc.html

docs/themerc
```diff
@@ -59,6 +59,8 @@ menu.overlap.x: 0
 menu.overlap.y: 0
 menu.width.min: 20
 menu.width.max: 200
+menu.border.width: 1
+menu.border.color: #a8a5a2
```
johanmalm, line 63: This should always correspond to the default value (via inheritance or explicitly set)

johanmalm (commented):
Sorry - I means to inherit `window.active.border.color: #e1dedb` above.

https://github.com/labwc/labwc/blob/c2928027becba26f058b544d4bd8c72895c8d7cf/docs/themerc#L19

src/common/scaled-rect-buffer.c
tokyo4j: Done.

include/common/scaled-rect-buffer.h
tokyo4j: I moved `TODO: ...` to `scaled_rect_buffer_create()` so that it doesn't look like the description of a parameter.
I didn't add descriptions of every parameter since they're self-explanatory.

docs/themerc
tokyo4j: Thanks. Fixed.

tokyo4j:
I changed the default value `window.active.border.color` to `#aaaaaa` which is the value of `menu.border.color` in Clearlook.
I also changed the default value of  `window.inactive.border.color` to `#aaaaaa` too as I think it's a bit strange to have the border color and titlebar color different in active windows but not in inactive windows.

johanmalm:
LGTM except that the commit message of `menu: support borders` should now `s/a8a5a2/aaaaaa/`

johanmalm:
Sorry, just one thing I discovered on further testing.

I feel that we shouldn't include the border-thickness in the overlap calculation. So with an `overlap.x` of `-1` there should be a visible gap between the menu-windows (including the border).

tokyo4j:
> I feel that we shouldn't include the border-thickness in the overlap calculation. So with an overlap.x of -1 there should be a visible gap between the menu-windows (including the border).

Sorry I should've mentioned that earlier. Though I don't have a strong preference on that, I think we should follow Openbox's behavior that submenus overlap with its parent by `menu.border.width`. Another option is making `menu.overlap.{x,y}` inherit `menu.border.width` by default and I think it's slightly more straightforward, but I don't think it's worth deviating from Openbox spec.

johanmalm:
> > I feel that we shouldn't include the border-thickness in the overlap calculation. So with an overlap.x of -1 there should be a visible gap between the menu-windows (including the border).
> 
> Sorry I should've mentioned that earlier. That behavior is copied from Openbox and was added with [1b64d86](https://github.com/labwc/labwc/commit/1b64d863c90896005ebb1ca3c63363cc6babbe98) in this PR. Without it, the border becomes twice as thick on the side where a submenu is opened and y-coordinates of submenus don't align nicely.

Okay. Have just been playing on openbox and agree with you. Ignore my previous comment.

johanmalm:
Big thank you. Very nice.

PR #2377, opened 2024-11-23
keyboard: optionally send release-events

johanmalm:
...for modifiers which are part of keybinds. This supports clients (for example blender) that want to see the modifier-release-event even if it was part of a keybinds. This is treated as a special case and can only be achieved by configuration.

Most clients (including those using Qt and GTK) are setup to not see these modifier release events - and actually misbehave if they do.  For example Firefox shows the menu bar if alt is pressed and then released, whereas if only pressed (because the release is absorbed) nothing happens. So, if Firefox saw bound modifier-release-events it would show the menu bar every time the window-switcher is used with alt-tab.

Issue #1507

Cc @charbelnicolas 

Testing appreciated.

TODO

- [x] Testing
- [x] Config options
- [x] Documentation
- [x] Consider making this app_id specific in the configuration (e.g. window-rule)

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -15,6 +15,7 @@
 #include "osd.h"
 #include "regions.h"
 #include "view.h"
+#include "window-rules.h"
 #include "workspaces.h"
 
 enum lab_key_handled {
@@ -363,7 +364,8 @@ get_keyinfo(struct wlr_keyboard *wlr_keyboard, uint32_t evdev_keycode)
 }
 
 static bool
-handle_key_release(struct server *server, uint32_t evdev_keycode)
+handle_key_release(struct server *server, uint32_t evdev_keycode,
+		bool is_modifier_key)
 {
 	/*
 	 * Release events for keys that were not bound should always be
@@ -386,11 +388,33 @@ handle_key_release(struct server *server, uint32_t evdev_keycode)
 		end_cycling(server);
 	}
 
+	key_state_bound_key_remove(evdev_keycode);
+
+	/*
+	 * There are some clients (for example blender) that want to see the
+	 * modifier-release-event even if it was part of a keybinds. This is
+	 * treated as a special case and can only be achieved by configuration.
+	 *
+	 * Most clients (including those using Qt and GTK) are setup to not see
+	 * these modifier release events - and actually misbehave if they do.
+	 * For example Firefox shows the menu bar if alt is pressed and then
+	 * released, whereas if only pressed (because the release is absorbed)
+	 * nothing happens. So, if Firefox saw bound modifier-release-events it
+	 * would show the menu bar every time the window-switcher is used with
+	 * alt-tab.
+	 */
+	struct view *view = server->active_view;
+	if (is_modifier_key && view) {
+		if (window_rules_get_property(view, "wantAbsorbedModifierReleaseEvents")
+				== LAB_PROP_TRUE) {
+			return false;
+		}
+	}
+
 	/*
 	 * If a press event was handled by a compositor binding, then do
 	 * not forward the corresponding release event to clients.
 	 */
-	key_state_bound_key_remove(evdev_keycode);
 	return true;
 }
 
@@ -525,7 +549,8 @@ handle_compositor_keybindings(struct keyboard *keyboard,
 			actions_run(NULL, server, &cur_keybind->actions, NULL);
 			return true;
 		} else {
-			return handle_key_release(server, event->keycode);
+			return handle_key_release(server, event->keycode,
+				keyinfo.is_modifier);
 		}
 	}
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -1055,6 +1055,18 @@ situation.
 	can be caused by *<margin>* settings or exclusive layer-shell clients
 	such as panels.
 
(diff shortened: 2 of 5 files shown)


charbelnicolas:
I'll try to apply the patch and report back, thank you!

By the way, do you think I should bring this issue with the blender developers too and see what they have to say?

charbelnicolas:
Hey @johanmalm, I've tested this patch and it works pretty well! Thank you :)

[https://drive.google.com/file/d/1nZefB9n1_25MdnlgQa_P69HgbCjliR-t/view?usp=sharing](https://drive.google.com/file/d/1nZefB9n1_25MdnlgQa_P69HgbCjliR-t/view?usp=sharing)

src/input/keyboard.c
```diff
@@ -386,11 +386,33 @@ handle_key_release(struct server *server, uint32_t evdev_keycode)
...
+	 *
+	 * Most clients (including those using Qt and GTK) are setup to not see
+	 * these modifier release events - and actually misbehave if they do.
+	 * For example Firefox shows the menu bar if alt is pressed and then
+	 * released, whereas if only pressed (because the release is absorbed)
+	 * nothing happens. So, if Firefox saw bound modifier-release-events it
+	 * would show the menu bar every time the window-switcher is used with
+	 * alt-tab.
+	 */
+	bool rc_send_absorbed_modifier_release_events_to_client = true;
```
Consolatis, line 404: I guess this could be a window rule property then?

The only other solution to this that I can see is delaying modifier press signals until either a release or a press of the next non-modifier key (and then send press events for both if its not a keybind or don't send anything in case it is). That has its own issues though.. e.g. consider a game where you press `CTRL` to crouch not  actually crouching until you also press some other key.

charbelnicolas: I think this is one of the reasons the SUPER/windows key exists, to separate it from other software using key modifiers like ALT, SHIFT, CTRL, etc... is that not the case?

charbelnicolas: I say this because I use the SUPER key for keybinds, in that way it doesn't interfere in applications.

Consolatis: I see your point but key-combos like `A-Tab`, `A-F4` and similar are kind of a standard for decades. In a perfect world I'd agree that compositors / window managers should only use super + key combinations and also be the only ones allowed to use them. That is not the world we live in though.

tokyo4j:
I reproduced the issue with Alt-Tabbing Blender window in labwc, but not in KWin. Debugging with `WAYLAND_DEBUG=client` showed that KWin sends `wl_keyboard.leave` at the start of window-switching. Maybe we can copy this behavior?

charbelnicolas:
> I reproduced the issue with Alt-Tabbing Blender window in labwc, but not in KWin. Debugging with `WAYLAND_DEBUG=client` showed that KWin sends `wl_keyboard.leave` at the start of window-switching. Maybe we can copy this behavior?

The problem is that it also happens when using other keybinds (like doing a screen video capture with `<keybind key="W-Print">`) and not only when window switching.

johanmalm:
> By the way, do you think I should bring this issue with the blender developers too and see what they have to say?

Yes. It might be worth trying with `mutter, `kwin` and another wlroots compositor like `wayfire` or `sway` first. The observation that @tokyo4j made on `kwin` is interesting, and it'd probably be helpful for the blender devs to know how blender works with keybinds on other compositors.

And if other compositors handle this differently, we should probably reevaluate our approach!

johanmalm:
Anyway. This is ready for final review now.

charbelnicolas:
> Anyway. This is ready for final review now.

Thank you Johan! I appreciate it!

src/input/keyboard.c
```diff
@@ -386,11 +387,34 @@ handle_key_release(struct server *server, uint32_t evdev_keycode)
...
+	 * alt-tab.
+	 */
+	struct view *view = server->active_view;
+	if (view && window_rules_get_property(view,
+			"wantAbsorbedModifierReleaseEvents") == LAB_PROP_TRUE) {
+		struct wlr_keyboard *kb = &server->seat.keyboard_group->keyboard;
+		if (is_modifier(kb, evdev_keycode)) {
+			return false;
+		}
+	}
```
Consolatis, line 412: Could we turn this around, e.g. check `is_modifier()` first? As this is a somewhat hot path and called on every key release I'd prefer if we could reduce the overhead as much as possible. At some point we should also convert the string properties to enums similar to what we do with actions.

Consolatis: We can pass `struct keyinfo` to this function rather than just the keycode. It contains a `.is_modifier` bool.

johanmalm: Okay - will do. 

Just to note though - it's only for release events of keys that were bound.

Consolatis: > Just to note though - it's only for release events of keys that were bound.

Right, missed the early out in `handle_key_release()`.
She the other branch in `handle_compositor_keybindings()` with `cur_keybind && cur_keybind->on_release` also return false in this case?

johanmalm: ^ No.

`handled==false` means that the key-event should be passed onto the focused client.

`cur_keybind->on_release` means that we have bound an action to the release event (like super to launch menu), and in this case we do NOT want to pass on the event to the client.

Consolatis:
LGTM

PR #2378, opened 2024-11-23
output: fix assert() fail when enabling an output that was disabled

jlindgren90:
When the output is in the process of being enabled, but the new state has not been committed yet, wlr_output->enabled is still false. So it's not safe to assert that it's true in output_enable_adaptive_sync().

Fixes: 36e099fc93b3caf87e7f390a732fcec69373de8c
"view: ensure output is usable before setting adaptive sync"


Consolatis:
Whoops. Thanks

PR #2380, opened 2024-11-25
menu: invert the y-offset of submenus applied by menu.overlap.y

tokyo4j:
In Openbox, `menu.overlap.y: 3` is applied like:

20241126_06h02m59s_grim

However, in the current labwc, the y-offset is inverted like:

20241126_06h03m49s_grim

This PR fixes it to follow Openbox's behavior.


johanmalm:
Thanks.

Interestingly, our documentation already describes the openbox behaviour!

"menu.overlap.y Vertical offset in pixels between submenus and their parents. Positive values for downwards and negative for upwards. Default is 0."

PR #2382, opened 2024-11-26
menu: don't include menu.border.width for menu title height

tokyo4j:
It was a cruft from my experiment of adding borders around titles.

It even caused an integer overflow because `theme->menu_border_width` can be `INT_MIN` when `menu_header_height` is calculated. This didn't cause sigfault, however, because `INT_MIN * 2` becomes 0.

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -1457,8 +1457,7 @@ post_processing(struct theme *theme)
 		+ 2 * theme->menu_items_padding_y;
 
 	theme->menu_header_height = font_height(&rc.font_menuheader)
-		+ 2 * theme->menu_items_padding_y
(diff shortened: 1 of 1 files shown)


tokyo4j:
I stopped the FreeBSD CI as it kept saying "VM is booting..." and didn't make any progress.

Consolatis:
> I stopped the FreeBSD CI as it kept saying "VM is booting..." and didn't make any progress.

Yeah, happens from time to time. It will timeout after 20 minutes from what I remember. I guess we could reduce the timeout to 10 minutes as the execution got much faster due to GH setting up new runners some time ago which support hardware virtualization.

PR #2388, opened 2024-11-27
Update for wlroots 0.19

Consolatis:
You know the drill.

Not intended to be merged anytime soon but it can be used to develop against the wlroots master branch for changes proposed there. Feel free to push chase commits to this branch (include the wlroots MR in he commit subject) but please don't do force pushes.

I do not recommend running this branch outside of dev work.

Upstream list of breaking changes:
https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3869

Issues:
- [x] when the SSD titlebar gets created the view width is now `0` for some reason and thus it triggers an assert() in wlroots because we now set the width to a negative value. Bypassed with a hacky `MAX(0, width - 2 * x)`. Needs further investigation
- [x] `wlr_damage_ring_add()` removed the return value, evaluate if we need to check against wlr_scene_output size when adding damage due to the magnifier
- [x] commits miss references to the actual wlroots commits, are in the wrong order and do not have proper subjects
- [x] wlroots.wrap file is not synced with each chase commit

JFYI @jlindgren90

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,7 +92,7 @@ jobs:
           apt-get upgrade -y
           apt-get install -y git gcc clang gdb xwayland
           apt-get build-dep -y labwc
-          #apt-get build-dep -y libwlroots-0.18-dev
+          apt-get build-dep -y libwlroots-0.18-dev
 
       - name: Install FreeBSD dependencies
         if: matrix.name == 'FreeBSD'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -117,7 +117,7 @@ jobs:
           xbps-install -Syu
           xbps-install -y git meson gcc clang pkg-config scdoc \
             cairo-devel glib-devel libpng-devel librsvg-devel libxml2-devel \
-            pango-devel wlroots0.18-devel gdb bash xorg-server-xwayland \
+            pango-devel wlroots0.19-devel gdb bash xorg-server-xwayland \
             dejavu-fonts-ttf libsfdo-devel foot
 
       # These builds are executed on all runners
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -21,7 +21,6 @@
 #include <wlr/types/wlr_keyboard.h>
 #include <wlr/types/wlr_keyboard_group.h>
 #include <wlr/types/wlr_layer_shell_v1.h>
-#include <wlr/types/wlr_matrix.h>
 #include <wlr/types/wlr_output.h>
 #include <wlr/types/wlr_output_management_v1.h>
 #include <wlr/types/wlr_output_power_management_v1.h>
--- a/include/view.h
+++ b/include/view.h
@@ -140,7 +140,7 @@ struct view_impl {
 	 * minimizing we don't destroy the foreign toplevel handle).
 	 */
 	void (*unmap)(struct view *view, bool client_request);
-	void (*maximize)(struct view *view, bool maximize);
+	void (*maximize)(struct view *view, enum view_axis maximized);
 	void (*minimize)(struct view *view, bool minimize);
 	struct view *(*get_root)(struct view *self);
 	void (*append_children)(struct view *self, struct wl_array *children);
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -59,9 +59,6 @@ struct xwayland_view {
 	struct view base;
 	struct wlr_xwayland_surface *xwayland_surface;
 
-	/* Used to detect stacking order updates */
-	int stacking_order;
-
 	/* Events unique to XWayland views */
 	struct wl_listener associate;
 	struct wl_listener dissociate;
@@ -84,8 +81,6 @@ void xwayland_unmanaged_create(struct server *server,
 
 void xwayland_view_create(struct server *server,
 	struct wlr_xwayland_surface *xsurface, bool mapped);
-
-void xwayland_adjust_stacking_order(struct server *server);
 
 struct wlr_xwayland_surface *xwayland_surface_from_view(struct view *view);
 
--- a/meson.build
+++ b/meson.build
@@ -50,9 +50,9 @@ endif
 add_project_arguments('-DLABWC_VERSION=@0@'.format(version), language: 'c')
 
 wlroots = dependency(
-  'wlroots-0.18',
+  'wlroots-0.19',
   default_options: ['default_library=static', 'examples=false'],
-  version: ['>=0.18.1', '<0.19.0'],
+  version: ['>=0.19.0', '<0.20.0'],
 )
 
 wlroots_has_xwayland = wlroots.get_variable('have_xwayland') == 'true'
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -49,6 +49,7 @@ data_buffer_destroy(struct wlr_buffer *wlr_buffer)
 	if (!buffer->surface_owns_data) {
 		free(buffer->data);
 	}
+	wlr_buffer_finish(wlr_buffer);
 	free(buffer);
 }
 
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -58,8 +58,8 @@ scene_output_damage(struct wlr_scene_output *scene_output,
 
 	if (pixman_region32_not_empty(&clipped)) {
 		wlr_damage_ring_add(&scene_output->damage_ring, &clipped);
-		pixman_region32_union(&scene_output->pending_commit_damage,
-			&scene_output->pending_commit_damage, &clipped);
+		pixman_region32_union(&scene_output->WLR_PRIVATE.pending_commit_damage,
+			&scene_output->WLR_PRIVATE.pending_commit_damage, &clipped);
 	}
 
 	pixman_region32_fini(&clipped);
@@ -85,8 +85,10 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
 	 * rendering on every output commit and overloads CPU.
 	 * We also need to verify the necessity of wants_magnification.
 	 */
-	if (!wlr_output->needs_frame && !pixman_region32_not_empty(
-			&scene_output->pending_commit_damage) && !wants_magnification) {
+	if (!wlr_output->needs_frame
+			&& !pixman_region32_not_empty(
+				&scene_output->WLR_PRIVATE.pending_commit_damage)
+			&& !wants_magnification) {
 		return true;
 	}
 
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -50,9 +50,7 @@ static void
 scene_output_damage(struct wlr_scene_output *scene_output,
 		const pixman_region32_t *region)
 {
-	if (!wlr_damage_ring_add(&scene_output->damage_ring, region)) {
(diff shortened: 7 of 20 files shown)


jlindgren90:
Separate horizontal & vertical maximize support is interesting. There are still some minor issues I run into with that (on my to-do list...) so I wonder if the wlroots support will help.

tokyo4j:
>  when the SSD titlebar gets created the view width is now 0 for some reason and thus it triggers an assert() in wlroots because we now set the width to a negative value. Bypassed with a hacky MAX(0, width - 2 * x). Needs further investigation

I think this is due to my poor implementation in #2213:

https://github.com/labwc/labwc/blob/d2a864dfe799d542441701af8d86fdc3b3ba1789/src/ssd/ssd-titlebar.c#L98

I'll open a PR later.

Consolatis:
> Separate horizontal & vertical maximize support is interesting. There are still some minor issues I run into with that

Maybe the last commit helps? We are using `view_toggle_maximize()` in the xwayland request handler which seems wrong. Should only matter when the X11 client requests the maximize state itself though, e.g. via client side deco or something like `wmctrl`.

jlindgren90:
@Consolatis I created a branch based on yours, with pinning to subproject commits:
https://github.com/jlindgren90/labwc/commits/wlroots-0.19/

I added one more fix but did not include 50af9e2cff67931916d28816cf45216de6bd4182 nor 4bf5d8a85d12218ab3b0a1387fccbdf81aa768f3 since I'm not sure which wlroots commits those should go with.

Consolatis:
> @Consolatis I created a branch based on yours, with pinning to subproject commits: https://github.com/jlindgren90/labwc/commits/wlroots-0.19/

Nice! You could take over this PR as well if you want to (e.g. just force-push whatever you want and I'd only push normally from now on).

> did not include [50af9e2](https://github.com/labwc/labwc/commit/50af9e2cff67931916d28816cf45216de6bd4182)

I \*think\* this one can be dropped completely.

> nor [4bf5d8a](https://github.com/labwc/labwc/commit/4bf5d8a85d12218ab3b0a1387fccbdf81aa768f3) since I'm not sure which wlroots commits those should go with.

If it makes sense in the first place it should likely be a fixup for "chase wlroots: xwayland separate maximized axis MR 4670".

jlindgren90:
@Consolatis thanks. I'm not looking to take this over entirely, let's work together on it. I went ahead and rebased it on top of my wlroots-0.19 branch, leaving to two remaining commits on top for now.

Consolatis:
There are a lot of shutdown asserts due to https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3926 when compiling this branch against current wlroots master. I'll work on that via the labwc master branch.

There is also a shadow assert with a negative height due to initial view geometry being `0` now (still not sure what this is about), I'll push a fixup commit here later.

Consolatis:
> There are a lot of shutdown asserts due to https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4918 when compiling this branch against current wlroots master. I'll work on that via the labwc master branch.

Decided to add the commit as first commit in this branch for now as to not interfere with the pending release. We can move it over when 0.8.3 is released.

> There is also a shadow assert with a negative height due to initial view geometry being `0` now (still not sure what this is about), I'll push a fixup commit here later.

Added a fix to the existing "[FIXME] ssd titlebar width is 0 on initialization" commit.

alexsch01:
`xdg-toplevel-icon-v1` is a nice Wayland protocol for having icons work for Wayland applications (that support the protocol) running inside of a distrobox/toolbox without adding a desktop file

tokyo4j:
Damage tracking of magnifier doesn't work correctly with transformed outputs in this branch.

The quirk in `scene_output_damage()` in wlroots 0.18 was that it accepted a damage region in un-transformed coordinates (not in output buffer coordinates), as I mentioned:

https://github.com/labwc/labwc/blob/58d22d05e22745a626ab1ec66aa1cb8802d12812/src/common/scene-helpers.c#L146-L149

And `wlr_damage_ring` accepted the un-transformed coordinates, while `wlr_scene_output->pending_commit_damage` accepted output buffer coordinates.

But [wlroots!4845](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4845) cleaned up this weird situation, so our scene helpers can be rewritten in a much cleaner way.

I applied the following patch on top of this branch and the magnifier worked correctly in transformed outputs (I almost copied `scene_output_damage()` from the latest wlroots master branch):

```diff
diff --git a/src/common/scene-helpers.c b/src/common/scene-helpers.c
index 20507d59..ea450e3a 100644
--- a/src/common/scene-helpers.c
+++ b/src/common/scene-helpers.c
@@ -48,31 +48,22 @@ lab_wlr_scene_get_prev_node(struct wlr_scene_node *node)
  */
 static void
 scene_output_damage(struct wlr_scene_output *scene_output,
-		const pixman_region32_t *region)
+		const pixman_region32_t *damage)
 {
-	wlr_damage_ring_add(&scene_output->damage_ring, region);
-
 	struct wlr_output *output = scene_output->output;
-	enum wl_output_transform transform =
-		wlr_output_transform_invert(scene_output->output->transform);
-
-	int width = output->width;
-	int height = output->height;
-	if (transform & WL_OUTPUT_TRANSFORM_90) {
-		width = output->height;
-		height = output->width;
-	}
 
-	pixman_region32_t frame_damage;
-	pixman_region32_init(&frame_damage);
-	wlr_region_transform(&frame_damage, region, transform, width, height);
+	pixman_region32_t clipped;
+	pixman_region32_init(&clipped);
+	pixman_region32_intersect_rect(&clipped, damage, 0, 0, output->width, output->height);
+
+	if (!pixman_region32_empty(&clipped)) {
+		wlr_damage_ring_add(&scene_output->damage_ring, &clipped);
+
+		pixman_region32_union(&scene_output->WLR_PRIVATE.pending_commit_damage,
+			&scene_output->WLR_PRIVATE.pending_commit_damage, &clipped);
+	}
 
-	pixman_region32_union(&scene_output->WLR_PRIVATE.pending_commit_damage,
-		&scene_output->WLR_PRIVATE.pending_commit_damage, &frame_damage);
-	pixman_region32_intersect_rect(&scene_output->WLR_PRIVATE.pending_commit_damage,
-		&scene_output->WLR_PRIVATE.pending_commit_damage, 0, 0,
-		output->width, output->height);
-	pixman_region32_fini(&frame_damage);
+	pixman_region32_fini(&clipped);
 }
 
 /*
@@ -144,12 +135,6 @@ lab_wlr_scene_output_commit(struct wlr_scene_output *scene_output,
 		pixman_region32_init_rect(&region,
 			additional_damage.x, additional_damage.y,
 			additional_damage.width, additional_damage.height);
-		/*
-		 * Region passed to scene_output_damage() should have the same
-		 * scale as the output buffer but have a different transform.
-		 */
-		wlr_region_transform(&region, &region, wlr_output->transform,
-			wlr_output->width, wlr_output->height);
 
 		scene_output_damage(scene_output, &region);
 		pixman_region32_fini(&region);
```

</details>

Consolatis:
> I applied the following patch on top of this branch and the magnifier worked correctly in transformed outputs

Nice, feel free to push a commit to this branch, otherwise I can add it in a few days. I can then also take care of moving it to the appropriate place and syncing the wlroots.wrap file.

Consolatis:
Perfect, thanks. I'll take care of that

Consolatis:
Rebased on master and moved the new commit into the right spot. Also changed `!pixman_region32_empty()` to `pixman_region32_not_empty()` as the _empty() variant was added recently and is not supported with older pixman versions. We could add another commit to change it back as wlroots increased the pixman dep version at some point.

All commits test compiled with the wlroots.wrap version checked out.

johanmalm:
LGTM.

Is #2727  a prerequisite to this one?

Consolatis:
> Is #2727 a prerequisite to this one?

yep

johanmalm:
Should we still use `wlr_xwayland_surface_restack()`? I remember a wlroots IRC conversation shortly after 0.18 saying that compositors shouldn't do this anymore. Looks like sway has got rid of theirs.

Ref:
- https://github.com/labwc/labwc/blob/568bc71ca2cccdbf323a0b31484aebb22ada8561/src/xwayland.c#L1187
- https://github.com/swaywm/sway/commit/951a22c2445f5c32b831bac0db86869627940402

tokyo4j:
> Should we still use wlr_xwayland_surface_restack()? I remember a wlroots IRC conversation shortly after 0.18 saying that compositors shouldn't do this anymore. Looks like sway has got rid of theirs.

Hm, I really like the sway's approach to delegate works to scene-graph API, but a tricky part for us is that we are moving shaded windows and windows in other workspace to the bottom of the scene:

https://github.com/labwc/labwc/blob/849fd7a96f54521b8e6e4e8f14d4774f3b9ddffb/src/xwayland.c#L1117-L1145

So in order to get rid of `wlr_xwayland_surface_restack()` without making regressions, we will need another way to tell xwayland that some windows are not actually shown.

tokyo4j:
@Consolatis I pushed a commit to remove `xwayland_adjust_stacking_order()`. The commit should be relocated to between [3548b0a](https://github.com/labwc/labwc/pull/2388/commits/3548b0abe8be1bf00d8d4cd63493cc70ad6f6cab) and [9aacada](https://github.com/labwc/labwc/pull/2388/commits/9aacada091854cf42248c3abc2143a0147d2bc9f).

Consolatis:
> Should we use `wlr_scene_output_needs_frame()`?

Potentially, but I feel like we should keep this PR as minimal as possible and do things like that as follow-up PRs.

I've rebased on master, removed the last work-around commit, moved the new commit into the right spot and compile tested every commit with the appropriate wlroots commit checked out.

I'll un-draft this PR but due to the changes of the xwayland stacking order I'd suggest someones testing the current state for a day or so on DRM, focusing on "unusual" X11 behavior like scrolling on X11 windows which don't have keyboard focus after workspace switching. Personally I only very lightly tested this branch nested.

johanmalm:
> > Should we use `wlr_scene_output_needs_frame()`?
> 
> Potentially, but I feel like we should keep this PR as minimal as possible and do things like that as follow-up PRs.

Okay, sounds sensible. Have made a note of in #2729 so we don't forget.

jlindgren90:
I'll try to do some testing over the next 24 hrs or so too

johanmalm:
Thanks @tokyo4j. I can reproduce that cursor-update issue. Have tried inserting a few `cursor_update_focus()` but nothing seems to help. Thanks for bisecting. I'm short of ideas for moving this one forward atm.

I've been running the 0.19 branch for a while now and other than the cursor-update everything seems to run really well - including xwayland stacking.

johanmalm:
Fabulous effort and big thanks!! LGTM. Let's go for it.

Consolatis:
@Narrat the master branch / labwc-git is now based on wlroots 0.19.

Narrat:
Alright :)
Unfortunately I will temporarily need to switch to static compiled wlroots then. Repo `wlroots` will likely stay on 0.18 until one package will require 0.19 (which was in the past the following sway release). And `wlroots-git` is providing 0.20.
But building the `wlroots` subproject currently fails for me with the restrictions I set :D

PR #2389, opened 2024-11-27
buffer: always set buffer->surface and remove buffer->cairo

tokyo4j:
I think `memcpy()` in `get_cairo_surface_from_lab_data_buffer()` is redundant as we can use `cairo_image_surface_create_for_data()` instead. Also we can always set `buffer->{cairo,surface}` so that we can just access them when we start writing on/from the buffer.

### DIFF
--- a/src/common/scaled-rect-buffer.c
+++ b/src/common/scaled-rect-buffer.c
@@ -39,7 +39,7 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
 		return NULL;
 	}
 
-	cairo_t *cairo = buffer->cairo;
+	cairo_t *cairo = cairo_create(buffer->surface);
 
 	/* Clear background */
 	cairo_set_operator(cairo, CAIRO_OPERATOR_CLEAR);
@@ -57,6 +57,8 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
 	cairo_set_line_width(cairo, self->border_width);
 	set_cairo_color(cairo, self->border_color);
 	cairo_stroke(cairo);
+
+	cairo_destroy(cairo);
 
 	return buffer;
 }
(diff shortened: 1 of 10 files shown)


jlindgren90:
Nice cleanup. I have no objection to creating the cairo_surface always. We do a lot of drawing with cairo so it's handy.

I wonder if creating the cairo_t all the time makes sense though. It's only used when drawing *to* the buffer, which for many buffers (e.g. application icons) never happens. So it's unnecessary memory overhead for those.

It's more typical usage (in code I've seen) for cairo_t to have a temporary lifetime - created right before drawing and destroyed once finished drawing. I think we should consider moving in that direction, and possibly not storing the cairo_t with the buffer at all.

tokyo4j:
> I wonder if creating the cairo_t all the time makes sense though. It's only used when drawing to the buffer, which for many buffers (e.g. application icons) never happens. So it's unnecessary memory overhead for those.

Agreed. It will also prevent accidentally reusing its previous state, without requiring to always call `cairo_save()` and `cairo_restore()`.

I pushed the second commit that removes `buffer->cairo`.

I needed to modify `copy_icon_buffer()` (which copies the icon buffer for hover/rounded variant) to also return `cairo_t` since its callers use the scale bound to it. But this workaround will be removed with a following PR for major refactor that dynamically updates the scale of icon buffers by introducing `scaled_img_buffer`: https://github.com/tokyo4j/labwc/tree/scaled-img-buffer

src/common/scaled-rect-buffer.c
```diff
@@ -58,6 +58,8 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
 	set_cairo_color(cairo, self->border_color);
 	cairo_stroke(cairo);
 
+	cairo_destroy(cairo);
```
jlindgren90, line 61: Do we need a cairo_surface_flush() here? (Not sure, I just noticed we have one in most of the other cases.)

jlindgren90 (commented):
LGTM, with one question.

src/common/scaled-rect-buffer.c
tokyo4j: Nice catch. I'll add it.

PR #2393, opened 2024-11-30
NEWS.md: interim update

johanmalm:


NEWS.md
```diff
@@ -71,6 +116,16 @@ Notes to package maintainers:
 
 ### Fixed
 
+- Fix incorrect focus behaviour when switching between workspaces with
+  omnipresent windows open [#2335]
+- Fall back to loading icon based on app-id when `Icon` defined in .desktop file
+  can not be loaded [#2361]
+- Fix regression introduced with dcd9b47 to allow negative values for theme
+  option `menu.overlap` [#2356]
```
Consolatis, line 124: Suggest referencing `0.8.1` rather than the commit id.

NEWS.md
```diff
@@ -34,15 +34,60 @@ The format is based on [Keep a Changelog]
...
+packaging.
+
+We do not normally describe behind-the-scenes work in this log, but will mention
+two here as an exception:
+
+1. A clean run with gcc/clang address-sanitizer has been achieved which is a
+   great position to protect from memory leaks. Thanks to @tokyo4j for fantastic
+   work with this [#2331]
+2. A buffer-sharing mechanism has been merged to improve both processor and
+   memory usage. Credits to @tokyo4j and @Consolatis for this one. [#2363]
```
Consolatis, line 48: Maybe add a "in the long term" here. Currently the effects are minimal and only the SSD title may be reused.

NEWS.md
```diff
@@ -34,15 +34,60 @@ The format is based on [Keep a Changelog]
...
+This is a shorter release cycle compared with the usual 10-week one because it
+contains a significant number of stability and cleanliness fixes which warrant
+packaging.
+
+We do not normally describe behind-the-scenes work in this log, but will mention
+two here as an exception:
+
+1. A clean run with gcc/clang address-sanitizer has been achieved which is a
+   great position to protect from memory leaks. Thanks to @tokyo4j for fantastic
+   work with this [#2331]
```
Consolatis, line 46: The way this is described my make it seems as if address-sanitizer was never used before. Suggest removing it completely and instead use something like "A clean run with gcc/clang memory leak check has been archived [..]".

Consolatis:
Another one in the already merged changelog regarding `Improve support for non-compliant icons by matching partial strings to handle for example app-id="gimp-2.10" with file "gimp.desktop`: `s/icons/icons and .desktop files/` or just `.desktop` files as the example uses that one.

johanmalm:
@Consolatis @tokyo4j Thanks for reviewing. All comments actioned.
Need to `fixup` befor emerge.

PR #2395, opened 2024-12-01
ssd: fix pixman error when SSD is created for tiny windows

tokyo4j:
I think it's fine to merge this PR for the next release.

This fixes the pixman error (which causes `assert()` failure in wlroots 0.19 https://github.com/labwc/labwc/pull/2388#issue-2696997166) when a tiny window is decorated.

In #2213, `set_squared_corners(false)` was always called when titlebar is created. However, `set_squared_corners(false)` sets the width of the titlebar background buffer to `(view width) - (corner radius)`, which causes pixman errors due to the negative width set for titlebar background buffer when the view is so small.

Note: `ssd_should_be_squared()` returns `true` when the titlebar corners should be un-rounded as the view is so small or tiled, but maximization state is not considered. `set_squared_corners()` rounds/un-rounds the titlebar corners, so `true` should be passed when `ssd_should_be_squared()` returns true or the view is maximized. This might be unclear and need refactoring or renaming.

### DIFF
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -95,14 +95,16 @@ ssd_titlebar_create(struct ssd *ssd)
 	ssd_update_title(ssd);
 	ssd_update_window_icon(ssd);
 
-	set_squared_corners(ssd, false);
-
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks.

PR #2399, opened 2024-12-03
 scaled-scene-buffer: reduce unnecessary renderings

tokyo4j:
For post-0.8.2.

Prior to this PR, a backing buffer with scale 1 was always created for a `scaled_scene_buffer` before showing it, and backing buffers for specific scales were created on `output_enter` events.

This PR removes this redundant re-renderings by calling `wlr_scene_buffer_set_dest_size()` upon `scaled_scene_buffer` creation so that we can receive `output_enter` events without creating a backing buffer, and delaying the first rendering to the first `output_enter` event.

This PR includes some refactors in `font.c` and `menu.c`, so I'm thinking of splitting this to smaller ones after we release 0.8.2.

### DIFF
--- a/src/common/font.c
+++ b/src/common/font.c
@@ -80,6 +80,18 @@ font_width(struct font *font, const char *string)
 }
 
 void
+font_get_buffer_size(int max_width, const char *text, struct font *font,
+	int *width, int *height)
+{
+	PangoRectangle text_extents = font_extents(font, text);
+	if (max_width > 0 && text_extents.width > max_width) {
+		text_extents.width = max_width;
+	}
+	*width = text_extents.width;
+	*height = text_extents.height;
+}
+
+void
 font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	const char *text, struct font *font, const float *color,
 	const float *bg_color, double scale)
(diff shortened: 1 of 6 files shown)


Consolatis:
> Prior to this PR, a backing buffer with scale 1 was always created for a `scaled_scene_buffer` before showing it, and backing buffers for specific scales were created on `output_enter` events.

Just for my understanding, you are only talking about the case where all outputs use a different scale than 1 here, right?

tokyo4j:
> Just for my understanding, you are only talking about the case where all outputs use a different scale than 1 here, right?

That's the main target case in this PR, but another motivation for this optimization is that I want to minimize the latency of opening the first window when we apply scaled-scene-buffer for titlebar buttons (ref: https://github.com/labwc/labwc/pull/2352#issuecomment-2481362924).

tokyo4j:

src/common/font.c
```diff
@@ -79,6 +79,18 @@ font_width(struct font *font, const char *string)
...
 
+void
+font_get_buffer_size(int max_width, const char *text, struct font *font,
+	int *width, int *height)
+{
+	PangoRectangle text_extents = font_extents(font, text);
+	if (max_width > 0 && text_extents.width > max_width) {
+		text_extents.width = max_width;
+	}
+	*width = text_extents.width;
```
Consolatis, line 90: This looks like it could be removed completely and the assignment changed to `*width = MAX(text_extents.width, max_width)` instead.

Consolatis:
LGTM, not tested.

PR #2401, opened 2024-12-04
NEWS.md: update for 0.8.2

johanmalm:
In readiness for 13-Dec

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog]
 
 | Date       | All Changes   | wlroots version | lines-of-code |
 |------------|---------------|-----------------|---------------|
+| 2024-12-13 | [0.8.2]       | 0.18.2          | 26277         |
 | 2024-10-25 | [0.8.1]       | 0.18.1          | 25473         |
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -144,6 +151,11 @@ menu.border.color: #aaaaaa
 
 ### Changed
 
+- Temporarily set xwaylandPersistence to 'yes' by default to avoid a bug that is
+  present in wlroots <0.18.2 and which has the potential to crash the compositor
+  when performing a drag-and-drop action at the same time as the XWayland server
+  is shutting down. This will be reverted when wlroots-0.18.2 can be linked
+  with. [#2371] [#2414]
```
Consolatis, line 158: Could be rewritten to something like

> Set xwaylandPersistence default value to `yes` when compiled with wlroots < 0.18.2. This prevents a bug which has the  potential to crash the compositor when performing a drag-and-drop action at the same time as the XWayland server is shutting down.

The revert part isn't required anymore as with 0.18.2 and later we will be back to `no` by default automatically.

Consolatis:
Other than the nitpick above: LGTM

NEWS.md
johanmalm: Thanks. Done.

PR #2402, opened 2024-12-05
query: extend "monitor" query & add missing docs

orfeasxyz:


johanmalm:
Thanks. Will review after release in a week's time if that's okay.

orfeasxyz:
No problem. I reference the issue https://github.com/labwc/labwc/issues/1837 since it is partially solved. We are still missing the ability to express "else if" in conditional actions and it's one thing I might work on in the future.

johanmalm:
Thanks. Seems fine.

PR #2403, opened 2024-12-06
menu: slightly slide menus opened from buttons

tokyo4j:
Suggested by stormshadow in IRC.

This PR slides menus opened from buttons (with `atCursor="no"`) to the left by `menu.border.width`.

|before|after|
|-|-|
|20241206_13h17m28s_grim|20241206_13h17m11s_grim|

This change is primarily for the default window menu opened from the left-corner window icon, but I think it doesn't sacrifice the look when the button from which the menu is opened is not placed in the left corner:

20241206_13h18m26s_grim

I think we can merge this PR before the next release, as the change is quite small.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -678,6 +678,7 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 			assert(ctx->node);
 			int ly;
 			wlr_scene_node_coords(ctx->node, &x, &ly);
+			x -= server->theme->menu_border_width;
 		}
 	}
 


Consolatis:
LGTM, up to @johanmalm to decide regarding merge before release.

johanmalm:
Thanks

st0rm-shad0w:
> This change is primarily for the default window menu opened from the left-corner window icon, but I think it doesn't sacrifice the look when the button from which the menu is opened is not placed in the left corner:

It might seem pedantic, but I feel it'll be visually more consistent if we have the window menu to show in top left corner for either cases. It does look odd when it opens with a margin. 🙈😅

PR #2405, opened 2024-12-07
cosmic-workspaces: abstract transaction-addon

Consolatis:
This allows to use it for a future ext-workspace implementation. It is also more generalized so can be used for other protocol implementation in the future in case the protocols require some kind of transaction management.

---

Split of from
- #2365

Todo:
- [x] document `transaction-addon.h`
- [x] remove refcount debug logging from `transaction-addon.c`


Consolatis:
> Assume we're looking at merging after release?

Yes, this shouldn't change anything (aside from additionally fixing some hard-to-trigger mem leaks) but you never know.

Consolatis:
Added documentation to the header + removed debug logging. Ready to go after 0.8.2 if there are no further review comments.

tokyo4j:
Doesn't it make more sense to rename `lab_transaction` to `lab_transaction_op` and `lab_transaction_session_context` to `lab_transaction`? AFAIK, a "transaction" means a set of operations that should be atomically committed.

Consolatis:
> Doesn't it make more sense to rename `lab_transaction` to `lab_transaction_op`
> AFAIK, a "transaction" means a set of operations that should be committed at once.

I see your point. `lab_transaction_operation` maybe? Personally I think the `op` word is slightly overloaded.

> and `lab_transaction_session_context` to `lab_transaction`

The session context is a more general concept because its shared between all wl_resources up to the initial bind of a registry global. It is currently only used for sharing transactions among these wl_resources and verifying that 2 given resources belong to the same session but in general the transactions are only a subset of the session context.

include/protocols/transaction-addon.h
```diff
@@ -0,0 +1,83 @@
...
+ * If it is NULL it will be created automatically which can be used for top
+ * level wl_resources (when a client binds a wl_global from the registry).
+ *
+ * The context refcount is increased by one after this call.
+ */
+struct lab_wl_resource_addon *lab_resource_addon_create(
+	struct lab_transaction_session_context *ctx);
+
+/*
+ * A generic transaction attached to a session context list.
```
tokyo4j, line 44: I think this doesn't explain what `lab_transaction_add()` does.

Consolatis: Suggestions appreciated, I am pretty bad at describing things.

tokyo4j:
> I see your point. `lab_transaction_operation` maybe? Personally I think the op word is slightly overloaded.

Agreed.

> The session context is a more general concept because its shared between all wl_resources up to the initial bind of a registry global. It is currently only used for sharing transactions among these wl_resources and verifying that 2 given resources belong to the same session but in general the transactions are only a subset of the session context.

OK, I see.

But after reading `cosmic-workspaces.c`, I'm wondering why we can't just directly bind `lab_cosmic_workspace{,_manager,_group}` to each `wl_resource` and store pending operations in something like `lab_cosmic_workspace_manager.pending_ops` instead of storing the operations in ref-counted `session_context`. I think this approach makes the codebase easier to read and more closely follows what wlroots does.

Consolatis:
> But after reading `cosmic-workspaces.c`, I'm wondering why we can't just directly bind `lab_cosmic_workspace{,_manager,_group}` to each `wl_resource` and store pending operations in something like `lab_cosmic_workspace_manager.pending_ops` instead of storing the operations in ref-counted `session_context`. I think this approach makes the codebase easier to read and more closely follows what wlroots does.

Because the transactions are per wayland client (or more precise per client manager bind). If we were to store them in `lab_cosmic_workspace_manager` they would be shared between different clients.

tokyo4j:
> Because the transactions are per wayland client (or more precise per client manager bind). If we were to store them in lab_cosmic_workspace_manager they would be shared between different clients.

Ah I see! Now I understand why you added `wl_resource_addon` in the conversion from per-client `wl_resource` into global `lab_cosmic_workspace{,manager,group}`!

Sorry for noises. LGTM other than the "transaction" wording.

Consolatis:
> Sorry for noises.

I actually very much appreciate your question. The more people know how and especially why things are written the way they are written the better :) I also tend to unnecessarily over complicate code at times so a critical view on the things I write is always appreciated.

tokyo4j:
As I said in the IRC, LGTM. Maybe I can open another PR for `transaction` renaming later.

Consolatis:
Went for `lab_transaction_op`, the `_operation` suffix was getting too long. This is ready from my point of view.

johanmalm:
Thanks.

PR #2406, opened 2024-12-07
menu: fix misplaced top-left window menu when window is maximized

### LINKED ISSUE #2404: Window menu fails to appear at top-left corner after right-clicking titlebar in maximized state
### Problem description

**Description:** There is a bug with the window menu in the maximized state. After right-clicking on the titlebar, the menu appears at the location of the right-click instead of the top-left corner. If you then click the dedicated menu button, the menu continues to appear at the right-click position. This does not happen in the normal (non-maximized) state.

https://github.com/user-attachments/assets/57e8c15d-934f-4081-8c2c-a6ac3f96bf56

### Steps to reproduce

**Steps to Reproduce:**

1. Maximize a window.
2. Right-click on the titlebar.
3. Observe that the window menu appears at the right-click position.
4. Click the dedicated menu button.
5. Observe that the window menu appears at the right-click position instead of the top-left corner.

### labwc build source

Local build

### labwc version

labwc 0.8.1-115-g2ade6a1e-dirty

### labwc environment

### Distribution

Void Linux

tokyo4j:
Fixes #2404.

With 2ade6a1, negative x-values can be passed to `menu_configure()` when the window menu is opened from top-left button in a maximized window, but `wlr_output_layout_output_at()` couldn't find the output for the coordinate since it's out of the output layout, thus it just opened the menu without updating its position.

So let's replace `wlr_output_layout_output_at()` in `menu_configure()` with `output_nearest_to()`, which searches for the output more roughly.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -867,10 +867,7 @@ menu_configure(struct menu *menu, int lx, int ly, enum menu_align align)
 	/* Get output local coordinates + output usable area */
 	double ox = lx;
 	double oy = ly;
-	struct wlr_output *wlr_output = wlr_output_layout_output_at(
-		menu->server->output_layout, lx, ly);
-	struct output *output = wlr_output ? output_from_wlr_output(
-		menu->server, wlr_output) : NULL;
+	struct output *output = output_nearest_to(menu->server, lx, ly);
 	if (!output) {
 		wlr_log(WLR_ERROR,
 			"Failed to position menu %s (%s) and its submenus: "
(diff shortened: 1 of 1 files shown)


tokyo4j:
Hmm... I'm not sure this is suitable for merging before the next release though, because this PR still allows negative values in `lx` and `ox` in `menu_configure()`, which doesn't expect out-of-output coordinates AFAIK. I admit I don't fully understand the logic for positioning menus.

I suggest reverting #2403 for now and fix the issue later, along with another issue reported by @st0rm-shad0w in IRC that the menu opened with Alt-Space still isn't placed naturally.

Consolatis:
So our options appear to be:
- reverting 2ade6a1 for now
- clamping x to 0 (not sure about edge cases with multiple outputs)
- using `output_nearest()`

Personally I think reverting is the best option for the release because the benefits are slightly better alignment but the downsides is more or less new untested code paths with edge cases.

tokyo4j:
Superseded by #2408.

PR #2407, opened 2024-12-08
Revert "menu: slightly slide menus opened from buttons"

tokyo4j:
This PR reverts #2403 due to the regression (#2404). #2406 might be fine for fixing the regression, but introducing an unknown code path would increase the risk of another regression.


johanmalm:
Agree with reverting.

LGTM

tokyo4j:
@st0rm-shad0w Sorry for reverting the fix. We'll reland it in the next release cycle.

st0rm-shad0w:
@tokyo4j No worries at all! More important to have a stable release. 

We can fix all related issues post 0.8.2 launch.

PR #2408, opened 2024-12-08
menu: improve algorithm for menu placement with xdg-positioner

tokyo4j:
For post-0.8.2. I wrote this patch some weeks ago, but now I think it's more useful than just a refactoring since this PR fixes the the root cause of #2404.

---

Since wlroots already provides a mechanism to place popups via `wlr_xdg_positioner_rules_get_geometry()` and `wlr_xdg_positioner_rules_unconstrain_box()`, I think we can simplify the codebase by applying them also for compositor-side menus. 

Notable functional changes are:
- Slide the menu to fit in the output when it's opened out of the output (e.g. when top-left window menu is opened when the window is overflowing to the left), rather than not updating the menu at all. This should prevent reintroducing #2404 when we reland #2403.

|before|after|
|-|-|
|20241208_17h39m10s_grim|20241208_17h39m29s_grim|

- The horizontal alignment of menus is now determined based on the size of each (sub)menu alone rather than the total width of entire menu tree. This means submenus can now overlap with its parents, but this is no longer a problem since we recently added support for menu borders.

|before|after|
|-|-|
|20241208_17h30m13s_grim|20241208_17h29m39s_grim|

- Fixed that pipemenus always follow the alignment of its parent even when it overflows from the output.

|before|after|
|-|-|
|20241208_17h32m45s_grim|20241208_17h33m01s_grim|

### DIFF
--- a/include/menu/menu.h
+++ b/include/menu/menu.h
@@ -12,14 +12,6 @@ struct wlr_scene_tree;
 struct wlr_scene_node;
 struct scaled_font_buffer;
 
-enum menu_align {
-	LAB_MENU_OPEN_AUTO   = 0,
-	LAB_MENU_OPEN_LEFT   = 1 << 0,
-	LAB_MENU_OPEN_RIGHT  = 1 << 1,
-	LAB_MENU_OPEN_TOP    = 1 << 2,
-	LAB_MENU_OPEN_BOTTOM = 1 << 3,
-};
-
 enum menuitem_type {
 	LAB_MENU_ITEM = 0,
 	LAB_MENU_SEPARATOR_LINE,
@@ -63,7 +55,7 @@ struct menu {
 	} selection;
 	struct wlr_scene_tree *scene_tree;
 	bool is_pipemenu;
-	enum menu_align align;
+	bool align_left;
 
 	/* Used to match a window-menu to the view that triggered it. */
 	struct view *triggered_by_view;  /* may be NULL */
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -9,6 +9,7 @@
 #include <string.h>
 #include <strings.h>
 #include <wayland-server-core.h>
+#include <wlr/types/wlr_xdg_shell.h>
 #include <wlr/util/log.h>
 #include "action.h"
 #include "common/buf.h"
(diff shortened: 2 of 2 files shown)


ahesford:
Your second "after" case, where menus open up willy-nilly over their ancestors, should never be allowed to happen. That completely destroys the visual representation of the hierarchy and makes it extremely difficult to make sense of what's going on.

In that case, the old behavior should be preferred, although that still provides surprising behavior where menus open to the left despite the arrows pointing right. Two even better options for menu placement would be:
1. Figure out when submenus would open leftward, and draw a left arrow before the submenu entry rather than a right arrow after it; or
2. Continue to allow menus to open rightward, but slide the entire menu hierarchy leftward (probably with some cursor warping) to make space for the submenu when it would otherwise collide with the screen edge.

tokyo4j:
> Your second "after" case, where menus open up willy-nilly over their ancestors, should never be allowed to happen. That completely destroys the visual representation of the hierarchy and makes it extremely difficult to make sense of what's going on.

I don't think overlapping submenus is visually that bad. It might be a bit confusing from the picture, but the user knows which menu is the root menu and which menu is the lastly opened one. And what's more, this behavior is the standard among most applications and WMs including Openbox.

> 1. Figure out when submenus would open leftward, and draw a left arrow before the submenu entry rather than a right arrow after it

This is an interesting approach and I think it's worth discussing. But the problem is that we currently don't allocate space for left-arrow before the item label, except for the user-defined `menu.items.padding.x`. We can slide the label to the right and insert a left-arrow, but I wonder how natural it looks.

> 2. Continue to allow menus to open rightward, but slide the entire menu hierarchy leftward (probably with some cursor warping) to make space for the submenu when it would otherwise collide with the screen edge.

I don't think this is what most users expect. I would be surprised if the entire menu hierarchy suddenly slides when the submenu can just be opened to the left. I may be biased, but again, I don't think we should deviate from the standard here.

ahesford:
I guess I misremembered how macOS behaves. Both macOS and Windows 11 leave the arrow pointing right, but the menus open to the left. Keeping our menus opening leftwards would follow that same convention. I could have sworn I've seen windowing systems that slide menus over as their children need room to display...

I still think allowing further children to open over their ancestors is really bad UX. Users are drilling down a tree, and we should establish a convention *at the moment the root is opened* that downward is either rightward (if there's space) or leftward and stick to that. Yes, the user is aware at the moment of how the menu configuration came to be, but it's really easy to glance away for a second, forget where you were, and get lost in the visual clutter of menus stacking atop each other. It's also just hard to make sense of a bunch of overlapping lines of text, whether or not the menus have borders (which the user might make the same color anyway).


There is certainly an extra constraint here---if a user has incredibly deep hierarchies, submenus may span the entire width of the screen no matter which way we open, so we'd have to reverse direction when we collide with the other end. Still, it's better to do that as few times as possible (think one long zig-zag back and forth across the screen as the menu gets ever deeper) rather than flipping direction every time space would allow.

Imagine looking at a tree diagram where the descendants of the root were placed essentially randomly on the page, and tracing arrows in the graph was the only visual queue you had of the relationships.

Another interesting design option would be to allow children to *replace* the parents, so there's only ever one menu on screen at a time... but that requires a "Back" button and a lot of thought (and probably animation) that we don't need to consider at the moment.

johanmalm:
Nice idea to use xdg-positioner.

You pick some great topics to work on! Keeping menus + OSDs within the output is possibly the only thing left for me to consider a `1.0.0` tag (although let's not start that debate in this thread).

I did start some work on this in #1944 and there is some commentary there expressing views on menu behaviors.

I think the openbox behavior is a good starting point.

Have recorded a quick gif of it.

Peek 2024-12-08 14-45

<details>

<summary>menu.xml code for nested submenus</summary>

```
  <menu id="a" label="a">
    <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
    <menu id="b" label="b">
      <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
      <menu id="c" label="c">
        <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
        <menu id="d" label="d">
          <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
          <menu id="e" label="e">
            <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
            <menu id="f" label="f">
              <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
              <menu id="g" label="g">
                <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
                <menu id="h" label="h">
                  <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
                  <menu id="i" label="i">
                    <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
                    <menu id="j" label="j">
                      <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
                      <menu id="k" label="k">
                        <item label="1 - long text to make it wide"/><item label="2"/><item label="3"/><item label="4"/><item label="5"/><item label="6"/><item label="7"/>
                      </menu>
                    </menu>
                  </menu>
                </menu>
              </menu>
            </menu>
          </menu>
        </menu>
      </menu>
    </menu>
  </menu>
```

</details>

johanmalm:
> Another interesting design option would be to allow children to _replace_ the parents, so there's only ever one menu on screen at a time... but that requires a "Back" button and a lot of thought (and probably animation) that we don't need to consider at the moment.

Wow this takes me back to writing `jgmenu` where we indeed allow opening nested in one window with a back button - as well as optionally flipping the submenu-arrows to the other side. 

I don't think there is a particularly clean way of doing deep hierarchies with exploring quite different UI approaches.

ahesford:
This video generally shows the behavior I'm advocating for. Note especially that, after the "c" menu is opened in the opposite direction, Openbox *keeps* opening leftward until it hits the left edge. In particular, the deeper children do *not* attempt to open rightward as space permits, so the new child is never drawn over any ancestor until the next screen edge is reached and it becomes strictly necessary.

Of course, when it is necessary,  we see the problem I'm complaining about. The "g" menu opens the "h" menu over its parent which, in your particular example, perfectly covers the parent and the entire hierarchy is visually broken at that point. The "h" menu (which actually shows the "i" item to open the "i" submenu) now looks like a child of the "e" submenu (which shows the "f" entry. Furthermore, even when you are looking at the screen and *expecting* the behavior, it is extremely hard to tell in this case that the child has even opened, because the only queue is the deselection of the bottom child and the change from "g" to "i" in the text of the entry. These queues are incredibly subtle.

Because this is the most extreme case of an incredibly deep hiearchy on a very small screen, we can't really do anything in that case, so we have to live with it. But the bulk behavior should be

1. Set the child-opening direction to rightward.
2. Open all descendant submenus rightward until a submenu will collide with the screen edge.
3. Switch the child-opening direction to leftward.
4. Open all descendant submenus leftward until a submenu will collide with the screen edge.
5. Return to 1.

johanmalm:
Have just tested branch. This is already a lot better than current behavior.

Should we consider continue to vertically 'bottom' align until we hit the bottom. Openbox does that and IIRC @Consolatis preferred that too. Note the difference in vertically alignment in the video below compared with the openbox one above.

[recording.webm](https://github.com/user-attachments/assets/8a53da12-357d-480c-a86c-25ee211bf018)

Consolatis:
Quick list of thoughts:
- spanning the menu across difference outputs seems fine to me (for both x and y axis)
- agree on the general idea to open in the same x direction until reaching the end, reversing the direction until reaching the end and so on
- regarding the y axis, if the last menu entries would be invisible and the whole menu fits within the output layout I think we should align the bottom of the submenu to the bottom of the layout. That ensures that all entries can be selected. E.g. neither of the two variants (first one hides the last entries, 2nd one wastes spaces). Maybe we could do the same for the top.
- in general I think we need some scrolling of menu entries in the longer term for the y axis if the height of a menu is > the output layout (e.g. not the height of a single output but the whole layout).

tokyo4j:
> But the bulk behavior should be
> 1. Set the child-opening direction to rightward.
> 2. Open all descendant submenus rightward until a submenu will collide with the screen edge.
> 3. Switch the child-opening direction to leftward.
> 4. Open all descendant submenus leftward until a submenu will collide with the screen edge.
> 5. Return to 1.

Sorry, I think my English in the description was bad. This is what I implemented, as the video by Johan's video shows. Whether the direction should be inverted is determined by the menu size alone, but the menus usually keep opening its children in the same direction as long as they can.

> Should we consider continue to vertically 'bottom' align until we hit the bottom. Openbox does that and IIRC @Consolatis preferred that too. Note the difference in vertically alignment in the video below compared with the openbox one above.

> - regarding the y axis, if the last menu entries would be invisible and the whole menu fits within the output layout I think we should align the bottom of the submenu to the bottom of the layout. That ensures that all entries can be selected. E.g. neither of the two variants (first one hides the last entries, 2nd one wastes space). Maybe we could do the same for the top.

Yeah I slightly prefer Opebox's behavior. That will add some lines of code since we'll have to allow vertically flipping the root menu around the cursor while we don't for the child menus, but that's certainly acceptable.

> spanning the menu across different outputs seems fine to me (for both x and y axis)

I agree, but it's annoying to implement since it's hard to calculate the box to constraint menus by with different sizes of monitors.

> in general I think we need some scrolling of menu entries in the longer term for the y axis if the height of a menu is > the output layout (e.g. not the height of a single output but the whole layout).

I agree.

tokyo4j:
Now this PR is ready for review.

johanmalm:
Thanks.
LGTM. Have tested it works well.
There is a merge conflict. Are you happy to sort?

PR #2409, opened 2024-12-08
action: fix menu position with "x"/"y" arguments in multi-monitor setup

tokyo4j:
Prior to this PR, output-relative coordinate was passed to `menu_open_root()` as the menu position when it's designated via `position.x`/`position.y` arguments in ShowMenu action, so menu can be misplaced to the output other than the one in which the cursor is in.

Intended for merging before 0.8.2, but it's up to other devs.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -688,40 +688,41 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 	if (pos_x && pos_y) {
 		struct output *output = output_nearest_to(server,
 				server->seat.cursor->x, server->seat.cursor->y);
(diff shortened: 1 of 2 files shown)


src/action.c
```diff
@@ -688,40 +688,41 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 	if (pos_x && pos_y) {
 		struct output *output = output_nearest_to(server,
 				server->seat.cursor->x, server->seat.cursor->y);
+		struct wlr_box usable = output_usable_area_in_layout_coords(output);
```
Consolatis, line 691: Pre-existing issue but couldn't `output_nearest_to()` return NULL (e.g. if there is no output connected at all)?

PR #2411, opened 2024-12-08
labwc-actions(5): add missing position option to ShowMenu

johanmalm:
Noticed when merging #2409

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -124,8 +124,19 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
 
-*<action name="ShowMenu" menu="value" atCursor="yes" />*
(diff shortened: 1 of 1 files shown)


docs/labwc-actions.5.scd
```diff
@@ -124,9 +124,21 @@ Actions are used in menus and keyboard/mouse bindings.
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
 
-*<action name="ShowMenu" menu="value" atCursor="yes" />*
+*<action name="ShowMenu"/>*
```
Consolatis, line 127: This makes it seem as if `ShowMenu` as only argument would work. Suggest adding `menu="root-menu"` or something.

johanmalm: Fixed.

johanmalm:
Ready to go.

Consolatis:
:)

PR #2414, opened 2024-12-08
Temporarily set xwaylandPersistence to `yes` by default

johanmalm:


Consolatis:
> which has the
potential to crash the compositor when performing a drag-and-drop action
with an XWayland client at the same time as the XWayland server is
shutting down.

The DND doesn't have to involve a XWayland client at all. The only condition is that xwayland is running at the start of the drag and terminating while the drag (or the drag source itself) is still active.

johanmalm:
Thanks for review. I changed the commit message to not mention XWayland client.

PR #2417, opened 2024-12-09
scripts: add wl_compcheck.py

Consolatis:
Simple utility to allow protocol and version comparison between different running compositors. This can be used to detect missing protocols or versions in labwc. See for example this result when running against the current labwc master and sway 1.10 (nested, results on DRM may differ slightly):

```
  Protocols missing from labwc @ wayland-0
  ext_transient_seat_manager_v1                   1
  wl_compositor                                   6  (has version 5)
  wp_alpha_modifier_v1                            1
  wp_content_type_manager_v1                      1
  xdg_wm_base                                     5  (has version 3)
  zwp_keyboard_shortcuts_inhibit_manager_v1       1

  Protocols missing from sway @ wayland-1
  wl_drm                                          2
  zcosmic_workspace_manager_v1                    1
```

### DIFF
--- a/scripts/wl_compcheck.py
+++ b/scripts/wl_compcheck.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python3
+
+import os
+import socket
+import struct
+
+# object ids
+WL_DISPLAY = 1
+WL_REGISTRY = 2
+WL_SYNC_DONE = 3
+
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. Really nice. And will be helpful to keep up with sway.

PR #2420, opened 2024-12-11
config: only default to xwayland persistence on old wlroots versions

ahesford:
This is a compile-time check, so it will be too conservative if somebody updates wlroots after building labwc, but that is still better than the alternatives.

I don't like the idea of hard-depending on `wlroots >= 0.18.2` after that is released. It isn't necessary for core (or even expected) functionality in labwc, and changing the default persistence flag effectively renders inoperable the bug we are trying to avoid. Strict version requirements are an unnecessary burden on distribution packagers as well as end users that may be balancing competing interests (imaging an end user trying to build the latest release, but who prefers to keep a system version of wlroots that has not yet been updated).

At the same time, forcing xwayland to persist by default when somebody *has* updated wlroots is also undesirable, so let's split the baby and choose the bug-avoiding default only when it seems necessary. If there is a runtime check for wlroots versions, we can consider doing that instead, but a compile-time check is straightforward and we can change the wording of the warning to suggest to users that they can avoid this problem either by changing settings or recompiling.

I think this behavior should persist until we have an actual hard requirement on newer wlroots, which probably means 0.19.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -14,6 +14,7 @@
 #include <wayland-server-core.h>
 #include <wlr/util/box.h>
 #include <wlr/util/log.h>
+#include <wlr/version.h>
 #include "action.h"
 #include "common/dir.h"
 #include "common/list.h"
@@ -35,6 +36,9 @@
 #include "view.h"
 #include "window-rules.h"
 #include "workspaces.h"
+
+#define LAB_WLR_VERSION_OLDER_THAN(major, minor, micro) \
+	(WLR_VERSION_NUM < (((major) << 16) | ((minor) << 8) | (micro)))
 
 static bool in_regions;
 static bool in_usable_area_override;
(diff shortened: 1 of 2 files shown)


src/config/rcxml.c
```diff
@@ -1089,16 +1090,14 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "xwaylandPersistence.core")) {
 		set_bool(content, &rc.xwayland_persistence);
 
-		/*
-		 * TODO: Temporary warning message. Revert when wlroots-0.18.2
-		 * can be linked with.
-		 */
+#if WLR_VERSION_NUM < 0x001202
```
Consolatis, line 1093: I think we should use some kind of
```c
#define LAB_WLR_VERSION_OLDER_THAN(major, minor, micro) (WLR_VERSION_NUM < (((major) << 16) | ((minor) << 8 )) | (micro)))

#if LAB_WLR_VERSION_OLDER_THAN(0, 18, 2)
    [..]
```

ahesford: Nice suggestion; adopted.

ahesford:
Doc fixes have been added.

Consolatis:
Thanks :)

PR #2422, opened 2024-12-12
docs/environment: remove example MOZ_ENABLE_WAYLAND=1

johanmalm:
...because it has not been relevant since before firefox v121 and even firefox-esr on Debian bullseye is now at v128.

Suggested-by: cry0xen


Consolatis:
No clue about Thunderbird but change LGTM.

johanmalm:
Just checked Thunderbird. Looks like it's v128, which against it what's being used on from bullseye onwards.

PR #2427, opened 2024-12-13
query: allow nested if actions

orfeasxyz:
This patch allows for nested "If" as well as "ForEach" actions. I did initially face some regressions however after reverting some changes and after some testing everything seems to be working.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -49,7 +49,8 @@ struct parser_state {
 	bool in_touch;
 	bool in_libinput_category;
 	bool in_window_switcher_field;
-	bool in_window_rules;
+	bool in_window_rule;
+	bool in_action;
 	bool in_action_query;
 	bool in_action_then_branch;
 	bool in_action_else_branch;
@@ -60,14 +61,13 @@ struct parser_state {
 	struct touch_config_entry *current_touch;
 	struct libinput_category *current_libinput_category;
 	const char *current_mouse_context;
-	struct action *current_keybind_action;
-	struct action *current_mousebind_action;
 	struct region *current_region;
 	struct window_switcher_field *current_field;
 	struct window_rule *current_window_rule;
-	struct action *current_window_rule_action;
 	struct view_query *current_view_query;
-	struct action *current_child_action;
+	struct action *current_inline_action;
+	struct wl_list *current_action_siblings;
+	struct action *current_action;
 };
 
 /* for backward compatibility of <mouse><scrollFactor> */
@@ -279,10 +279,14 @@ fill_window_rule(char *nodename, char *content, struct parser_state *state)
 		state->current_window_rule->window_type = -1; // Window types are >= 0
 		wl_list_append(&rc.window_rules, &state->current_window_rule->link);
 		wl_list_init(&state->current_window_rule->actions);
+		state->current_action = NULL;
+		state->current_action_siblings = &state->current_window_rule->actions;
+		state->current_inline_action = NULL;
 		return;
 	}
 
 	string_truncate_at_pattern(nodename, ".windowrule.windowrules");
+
 	if (!content) {
 		/* nop */
 	} else if (!state->current_window_rule) {
@@ -328,16 +332,16 @@ fill_window_rule(char *nodename, char *content, struct parser_state *state)
 
 	/* Actions */
 	} else if (!strcmp(nodename, "name.action")) {
-		state->current_window_rule_action = action_create(content);
-		if (state->current_window_rule_action) {
+		state->current_inline_action = action_create(content);
+		if (state->current_inline_action) {
 			wl_list_append(&state->current_window_rule->actions,
-				&state->current_window_rule_action->link);
+				&state->current_inline_action->link);
 		}
-	} else if (!state->current_window_rule_action) {
+	} else if (!state->current_inline_action) {
 		wlr_log(WLR_ERROR, "expect <action name=\"\"> element first. "
 			"nodename: '%s' content: '%s'", nodename, content);
 	} else {
-		action_arg_from_xml_node(state->current_window_rule_action, nodename, content);
+		action_arg_from_xml_node(state->current_inline_action, nodename, content);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks for your continuous work in this area!

Just to set expectations, this might take a while to review properly.

tokyo4j:
I suggest packing __all__  of the parser state into `parser_state`.  I think it feels strange have `parset_state` that only contains a part of the parser state.

Also, for easiness of review, I suggest splitting this into 2 commits or PRs that (1)adds `parser_state` and (2)adds support for nested if actions.

Keep in mind that I'm 100% not sure introducing `parser_state` is the way to go, though I personally support this approach.

orfeasxyz:
> I suggest packing all of the parser state into parser_state. I think it feels strange have parset_state that only contains a part of the parser state.

I agree, looking back to this it does feel inconsistent to have separate global and non-global state.

> Also, for easiness of review, I suggest splitting this into 2 commits or PRs that (1)adds parser_state and (2)adds support for nested if actions.

That's fine by me.

> Keep in mind that I'm 100% not sure introducing parser_state is the way to go, though I personally support this approach.

What would be an alternative way?

tokyo4j:
> What would be an alternative way?

I just wanted to tell you my guide may be inaccurate since I'm not very familiar with codes in `rcxml.c`. A drastic approach I came up with is rewriting `rcxml.c` to have functions like `struct keybind *consume_keybind(xmlNode *node)` that converts `<keybind>` xml node into `struct keybind` instead of chaining if-else statements in `entry()`, but that's too drastic and will incur many bugs. So as I said, I like your approach at the moment.

Consolatis:
The config parsing now does way more than it was originally the case. New things have been bolted on top and so far it kind of works for the most part, although not without issues (e.g. user supplied attribute order matters in some cases). I think in the really long term we need to re-design the whole thing from the ground up (but keep it attribute / child element agnostic). For this specific case of nested Ifs I can't really tell if we can bolt it on as well or if that causes a maintenance nightmare in the future, didn't yet read the diff in detail.

orfeasxyz:
> > What would be an alternative way?
> 
> I just wanted to tell you my guide may be inaccurate since I'm not very familiar with codes in `rcxml.c` and other devs may come up with a better way. A drastic approach I came up with is rewriting `rcxml.c` to have functions like `struct keybind *consume_keybind(xmlNode *node)` that converts `<keybind>` xml node into `struct keybind` instead of chaining if-else statements in `entry()`, but that's too drastic and will incur many bugs. So as I said, I like your approach at the moment.

Honestly while looking through the parser's code I did find the logic somewhat messy. I would not mind doing a larger rewrite, however as you also mentioned this may introduce a lot of bugs in a very sensitive part of the code base.

One change I want to suggest however, is to *not* fallback to the default config in the case of a configuration error. I tent to experiment a lot when I first started using labwc and if I missed a closing tag, suddenly all of my keybinds would stop working until I fixed the config and reloaded.

Consolatis:
> One change I want to suggest however, is to **not** fallback to the default config in the case of a configuration error. I tended to experiment a lot when I first started using labwc and if I missed a closing tag, suddenly all of my keybinds would stop working until I fixed the config and reloaded.

That should only happen if the xml file itself is broken, e.g. can't be parsed by libxml2. Maybe we could try to throw it at libxml2 before doing the `rcxml_finish` -> `rcxml_read` -> `rcxml_init` -> `rcxml_parse_xml` routines. Or do some kind of dry run before.

johanmalm:
Set to `draft` as it does not look ready for merge.

Consolatis:
> Honestly while looking through the parser's code I did find the logic somewhat messy. I would not mind doing a larger rewrite, however as you also mentioned this may introduce a lot of bugs in a very sensitive part of the code base.

A potential approach could be adding some `config/rcxml_new.c`, moving `config/rcxml.c` to `config/rcxml_old.c`, creating a new `config/rcxml.c` which only exposes the public function and acts as wrapper based on some temporary env var which if set uses the new implementation to fill the global `struct rcxml rc` and the old one if not.

So rather than potentially breaking the existing implementation we provide some kind of opt-in for users willing to test the new implementation. That is also something that could be done for `menu/menu.c` to split out the menu.xml parsing out of it.

Once everyone is satisfied we can then remove the old implementation and the wrapper.

orfeasxyz:
I've split the patch into 2 commits as it was suggested by @tokyo4j. From all the personal testing I have done, everything seems to be working smoothly.

johanmalm:
Has branch conflict

orfeasxyz:
Should be fine now

johanmalm:
I am away from the computer for a week and find this one too much to review on a phone.

I will take a look when I get back unless anyone else gets there first.

tokyo4j:
Sorry, I couldn't understand the code changes. I guess that's not because your code is bad, but it lacks explanation of why these changes are needed.

Could you give us the overview of your changes? Something like this:
- I removed `xxx` and added `yyy` in `struct parser_state` because ...
- I changed the signature of `fill_action_query()` because ...
- When the parser encounters nested `If` / `Else` actions, it currently ... but in this PR, ...
 
You can either post a comment here or add comments in GitHub's diff viewer.

src/config/rcxml.c
```diff
@@ -1352,10 +1380,22 @@ xml_tree_walk(xmlNode *node, struct parser_state *state)
...
+		if (!strcasecmp((char *)n->name, "action")) {
+			struct parser_state new_state = *state;
+			new_state.current_action = NULL;
+			new_state.current_view_query = NULL;
+			new_state.in_action = true;
+			new_state.in_action_query = false;
+			new_state.in_action_then_branch = false;
+			new_state.in_action_else_branch = false;
+			new_state.in_action_none_branch = false;
+			traverse(n, &new_state);
```
orfeasxyz, line 1398: In order to allow actions to be nested without losing track, we create a copy of the current parser state, initialize all fields that are relevant to the current action and descend.

orfeasxyz:
I added a few comments trying to explain most of my changes. I will leave 2 changes as a TODO in the future.

1) Find some way for all action logic to be proxied through the `fill_action()`. The current solution of keeping a `current_inline_action` is not something I am proud of.
2) Rename `fill_child_action()` to something like `initialize_parent_action()`.

tokyo4j:
Sweet. Your comments are very helpful. I'll review this PR again tonight.

> Find some way for all action logic to be proxied through the fill_action(). The current solution of keeping a current_inline_action is not something I am proud of.

Agreed. Having both `current_action` and `current_inline_action` was the most confusing part for me when I read your code yesterday.

I'm even wondering if we should keep supporting the shortened form like `<keybind key="A-F4" name.action="Close" />`. I actually tried to refactor the entire `rcxml.c` some months ago and tried to use recursion instead of encoding the node names, but keeping the compatibility of this shortened form was one of the biggest pain. @johanmalm @Consolatis @jlindgren90 @ahesford any ideas?

orfeasxyz:
IMO completely dropping the `name.action` syntax will be quite breaking. One of the ideas I got while working on this is to convert XML attributes to elements since we use them interchangeably. This way we could keep the syntax, while lowering the overall complexity of the parser.

tokyo4j:
> One of the ideas I got while working on this is to convert XML attributes to elements since we use them interchangeably.

I've got the same idea (as I said in the IRC), and am implementing it. I can open a PR for it later.

orfeasxyz:
Closing this as https://github.com/labwc/labwc/pull/2667 got merged.

PR #2429, opened 2024-12-14
font: remove arrow character from font buffer

tokyo4j:
Split from #2399. This PR removes `arrow` argument from `font_buffer_create()` and `scaled_font_buffer_update()` so that we can simplify the calculation of the size of a font buffer. Now, 2 separate scaled-font-buffers for a label and a arrow sign are created for a submenu item. 

Other changes are:
- Now `font_buffer_create()` accepts non-positive numbers for `max_width`, in which case the natural font width is always used for buffer creation.
- `menuitem->{normal,selected}` are replaced by simple `menuitem->{normal_tree,selected_tree}`, since we no longer dynamically update buffers of existing menuitems.

### DIFF
--- a/include/common/font.h
+++ b/include/common/font.h
@@ -44,11 +44,10 @@ int font_width(struct font *font, const char *string);
  * @font: font description
  * @color: foreground color in rgba format
  * @bg_color: background color in rgba format
- * @arrow: arrow (utf8) character to show or NULL for none
  */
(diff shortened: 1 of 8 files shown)


johanmalm:
Thanks. That looks much better. That `arrow` argument never looked right (my fault).

This also open the door for supporting `bullet.xbm` which is the way the arrow should really be done :smile:

https://openbox.org/help/Themes#Submenu_bullet

PR #2432, opened 2024-12-14
build: rename meson_options.txt to meson.options

johanmalm:
...because:

- this is the preferred name since Meson 1.1
- wlroots did it


meson.build
```diff
@@ -3,7 +3,7 @@ project(
   'c',
   version: '0.8.2',
   license: 'GPL-2.0-only',
-  meson_version: '>=0.59.0',
+  meson_version: '>=1.1',
```
Consolatis, line 6: Debian Stable (and this likely a bunch of other distros using it as a base) is still on meson 1.0.1: https://tracker.debian.org/pkg/meson

As a result I am a bit skeptical about this change at the current point in time as it isn't really needed from the technical side. wlroots 0.18.x also still uses `meson_options.txt`.

johanmalm:
Okay. Good point. Will close for now. Can reconsider in future. Just liked the look of it so much.

PR #2436, opened 2024-12-15
Some more minor prep work for ext-workspaces

Consolatis:
Split of from
- #2365

Some more minor prep work:
- Removes `lab_cosmic_workspace_set_hidden()` calls as its a concept labwc does not support.
- Prefixes some cosmic internal API naming to not clash with a future ext-workspace implementation


johanmalm:
LGTM

PR #2437, opened 2024-12-16
keyboard: fix missing release event when IME is enabled while key press 

tokyo4j:
After commit [e2189903](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e21899037a97dcade40149f05ab54229c0018644) in wlroots, when ctrl-f is pressed in firefox with a IME client running, the following key-release event for "f" is not sent, thus "f" is repeated like "ffffffffff..." in the input box of firefox. This is because the key-release event for "f" is firstly relayed to the IME client and then sent via the virtual keyboard created by the IME client while the key-press event is sent via physical keyboard, and with [e2189903](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e21899037a97dcade40149f05ab54229c0018644), key-release events without a corresponding key-press event on the same keyboard is not emitted to the compositor.

https://github.com/user-attachments/assets/b87e7e7f-c288-4f50-8902-361a6f0be678

So this PR fixes this problem by not relaying the key-release event to the IME client when the corresponding key-press event was previously sent to the application.

This follows sway's behavior:
https://github.com/swaywm/sway/blob/f293418d9d8a9e71fc52bc75a83d24ec2cc934c9/sway/input/keyboard.c#L548-L561

### DIFF
--- a/include/input/ime.h
+++ b/include/input/ime.h
@@ -21,6 +21,7 @@ struct input_method_relay {
 	struct wl_list text_inputs; /* struct text_input.link */
 	struct wlr_input_method_v2 *input_method;
 	struct wlr_surface *focused_surface;
+	struct lab_set pressed_keys;
 	/*
 	 * Text-input which is enabled by the client and communicating with
 	 * input-method.
--- a/src/input/ime.c
+++ b/src/input/ime.c
@@ -73,9 +73,25 @@ bool
 input_method_keyboard_grab_forward_key(struct keyboard *keyboard,
 		struct wlr_keyboard_key_event *event)
 {
+	/*
+	 * We should not forward key-release events without corresponding
+	 * key-press events forwarded
+	 */
+	struct lab_set *pressed_keys =
+		&keyboard->base.seat->input_method_relay->pressed_keys;
+	if (event->state == WL_KEYBOARD_KEY_STATE_RELEASED
+			&& !lab_set_contains(pressed_keys, event->keycode)) {
+		return false;
+	}
+
 	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
 		get_keyboard_grab(keyboard);
 	if (keyboard_grab) {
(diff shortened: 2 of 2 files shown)


src/input/keyboard.c
```diff
@@ -677,7 +686,8 @@ keyboard_key_notify(struct wl_listener *listener, void *data)
 				&& event->state == WL_KEYBOARD_KEY_STATE_PRESSED) {
 			start_keybind_repeat(seat->server, keyboard, event);
 		}
-	} else if (!input_method_keyboard_grab_forward_key(keyboard, event)) {
+	} else if (pressed_sent_key_is_released
```
johanmalm, line 689: I find this logic hard to follow. Could we considering refactoring it with the aim of keeping `keyboard_key_notify()` easy on the brain. For example:

1. Could the `pressed_sent_key_is_released` logic go at the top of `input_method_keyboard_grab_forward_key()`.
2. Or maybe a `ime_grab()` wrapper containing both.

johanmalm (commented):
Thanks. And great work on tracking this down!

Have added a suggestion to refactor.

Is it safe to use this when linking to a wlroots version which does not contain the fix?

tokyo4j:
I've updated the patch with another approach that only touches `ime.c`, since my previous patch had an issue with normal applications like foot with IME enabled; when you press and release a character, the release event was not forwarded to the IME while the press event was forwarded, so the IME began sending repeated press events as it could observe the press event.

Instead, I added a `lab_set` that tracks the key-press events forwarded to IME. If a key-release event doesn't have corresponding key-press event in the `lab_set`, the key-release event is not forwarded. I also like how my new approach doesn't touch the normal code paths without IME.

> Is it safe to use this when linking to a wlroots version which does not contain the fix?

Yeah it should be safe. I tested with wlroots 0.18.1 and it works without any issue.

johanmalm:
LGTM. Not tested.

PR #2439, opened 2024-12-16
labwc-config(5): describe how keys are processed

johanmalm:
...when matching against keybinds.

Copied from:

https://github.com/labwc/labwc/blob/63dc609085d42d16401e23079bb50e2f07e31de5/src/input/keyboard.c#L215-L238


Consolatis:
LGTM

PR #2442, opened 2024-12-18
menu: slightly slide menus opened with atCursor="no"

tokyo4j:
This is reland of #2403 which was reverted with #2407, but this prevents the x-position of the menu from being negative and slides menus opened with Alt-Space too. CC @st0rm-shad0w

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -670,14 +670,17 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 	}
 	/* Place menu in the view corner if desired (and menu is not root-menu) */
 	if (!at_cursor && view) {
-		x = view->current.x;
+		struct wlr_box extent = ssd_max_extents(view);
+		x = extent.x;
 		y = view->current.y;
 		/* Push the client menu underneath the button */
 		if (is_client_menu && ssd_part_contains(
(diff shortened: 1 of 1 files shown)


src/action.c
```diff
@@ -670,14 +670,17 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
...
 		/* Push the client menu underneath the button */
 		if (is_client_menu && ssd_part_contains(
 				LAB_SSD_BUTTON, ctx->type)) {
 			assert(ctx->node);
-			int ly;
-			wlr_scene_node_coords(ctx->node, &x, &ly);
+			int lx, ly;
+			wlr_scene_node_coords(ctx->node, &lx, &ly);
+			/* MAX() prevents negative x when the window is maximized */
+			x = MAX(x, lx - server->theme->menu_border_width);
```
tokyo4j, line 683: I initially thought negative x values can be handed with #2408, but `MAX()` is needed anyway because otherwise the menu can be constrained in the left-adjacent output when the window is maximized.

st0rm-shad0w:
Thank you @tokyo4j ! 🙌
I wrecked my machine, will have it fixed by the weekend and try.

PR #2443, opened 2024-12-18
xdg-popup: update pointer focus on popup destruction

### LINKED ISSUE #2438: Strange problem with (right) clicking after a context menu is closed
### Problem description

This happens in both Qt and GTK apps. After a context menu is opened by right clicking and closed by clicking a menu item *that doesn't open a new window*, right clicking or clicking won't work until the mouse cursor is moved.

I'm not sure that the problem is new or existed before — I just encountered it today. It doesn't happen when the menu is closed by the Escape key, or another window is opened as the result of clicking a menu item.

It doesn't happen under kwin_wayland either. I haven't tried other Wayland compositors.

### Steps to reproduce

 1. Open any Qt or GTK app, in which you can right click.
 2. Click any context menu item that doesn't open a new window.
 3. Make sure you do not move the mouse cursor, and try to right click again.

Neither right clicking nor clicking will work until you move the cursor a little.

### labwc build source

Local build

### labwc version

0.8.2

### labwc environment

### Distribution

Manjaro

EDIT: Tested under Wayfire; the issue didn't happen in it.

tokyo4j:
Fixes #2438.

@jlindgren90 Should we do this also for unmanaged xwayland surfaces?

### DIFF
--- a/include/layers.h
+++ b/include/layers.h
@@ -25,6 +25,7 @@ struct lab_layer_surface {
 struct lab_layer_popup {
 	struct wlr_xdg_popup *wlr_popup;
 	struct wlr_scene_tree *scene_tree;
+	struct server *server;
 
 	/* To simplify moving popup nodes from the bottom to the top layer */
 	struct wlr_box output_toplevel_sx_box;
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1118,9 +1118,7 @@ cursor_process_button_release(struct seat *seat, uint32_t button,
 				menu_call_selected_actions(server);
 			} else {
 				menu_close_root(server);
-				double sx, sy;
-				cursor_update_common(server, &ctx, time_msec,
-					/*cursor_has_moved*/ false, &sx, &sy);
+				cursor_update_focus(server);
 			}
 		}
 		return notify;
--- a/src/layers.c
+++ b/src/layers.c
@@ -363,6 +363,8 @@ popup_handle_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&popup->commit.link);
 	}
 
+	cursor_update_focus(popup->server);
+
 	free(popup);
 }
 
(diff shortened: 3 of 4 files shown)


jlindgren90:
> Should we do this also for unmanaged xwayland surfaces?

I think we already do, see `handle_unmap()` in `xwayland-unmanaged.c`.

johanmalm:
LGTM

However, I don't think it fixes #2438 
I was able to re-produce that one and guess it may relate to:
1. XDG window-menu requests (https://github.com/labwc/labwc/blob/ac7c94db12d9b599050122a00b922b82e21cb16b/src/xdg.c#L431) 
2. Absorbing some mouse buttons in `input/cursor.c`

I reproduced it with `gnome-weather` using CSD (which that one insists on) and just right clicking on the titlebar.

tokyo4j:
> However, I don't think it fixes https://github.com/labwc/labwc/issues/2438

I think there's a confusion of compositor-side menu vs client-side menu. I guess "menu" in #2438 is just about the client-side menu as it says wayfire (which doesn't have compositor-side menu) works fine. So, I think this PR fixes #2438.

> I reproduced it with gnome-weather using CSD (which that one insists on) and just right clicking on the titlebar.

Did you right-click the titlebar to open menu and left-click to close it, _without moving the cursor_? That's what I reproduced; when I click the _border_ of already opened menu, the menu is closed, but the pointer focus doesn't move back to the window. This is because `get_cursor_context()` returns `type=LAB_SSD_NONE` when the cursor is on the menu border and `cursor_update_common()` clears the pointer focus in the button-release handler:

https://github.com/labwc/labwc/blob/ac7c94db12d9b599050122a00b922b82e21cb16b/src/input/cursor.c#L1106-L1123

tsujan:
The strange thing is that this patch fixes the issue for me everywhere except for the place where I encountered it the first time, namely, pcmanfm-qt's Desktop, which is on the background layer (context menus are OK inside pcmanfm-qt's windows).

It gets even stranger because, on Desktop, I can consistently reproduce it by clicking "Hide Desktop Items" in the context menu, while it only randomly happens by clicking "Select All".

I recompiled labwc 0.8.2 against wlroots 0.18.2 today, and applied this patch afterward.

tokyo4j:
Oh, that's a good catch. I forgot about popups created as a child of layer-shell surface. I have no idea why it randomly happens with "Select All" though.

tsujan:
I confirm that now the patch fixes all cases that I'd encountered. Thanks!

PR #2444, opened 2024-12-20
ssd: rework titlebar button rendering 

### LINKED ISSUE #2286: Title-bar buttons not aligned correctly in 0.8.1
This screenshot shows the left title-bar buttons of a normal window with Labwc 0.8.0:

titlebar-old

The buttons are perfectly aligned.

After I upgraded to 0.8.1, I added these lines to my theme to have the same sizes

```
window.button.width: 20
window.button.spacing: 6
window.titlebar.padding.width: 3
window.titlebar.padding.height 3
```

The result is this:

titlebar-new

The close button at the left is one pixel off vertically (you could check it in GIMP). The same happens with my right-most menu button. The other buttons are OK.

However, this offset doesn't exist in maximized or tiled windows. I have `<cornerRadius>8</cornerRadius>` in `rc.xml`.

tokyo4j:
## What this PR does

- Fix #2286
- Fix blurry window button icons in outputs with a scale larger than 1

### `lab_img`

- `lab_img` holds a cache for a loaded image (`lab_data_buffer` for PNG/XBM/XPM images and `RsvgHandle` for SVG images).
- `lab_img` is created with `lab_img_load()` and rendered to a buffer with `lab_img_render()`. This allows rendering the same images for different sizes/scales.
- __"Modifier"__ functions can be applied to `lab_img` with `lab_img_add_modifier()`. This applies additional drawing operations in `lab_img_render()`. In this PR, this is used for overlaying hover effects on window buttons and rounding the corner buttons.
- The cache in `lab_img` is refcounted and can be shared by multiple `lab_img`s when they are duplicated with `lab_img_copy()`. In this PR, `lab_img`s for builtin normal/hovered buttons share the same caches while having different "modifiers".

### `scaled_img_buffer`

- This inherits `scaled_scene_buffer` and wraps `lab_img` to render it as a scene-node. When it enters an output with a different scale, it calls `lab_img_render()` to render the image in that scale.
- `scaled_scene_buffer_update()` allows updating its `lab_img` and geometry. This PR uses this to update the image when the new icon is loaded from app-id.

### Additional notes

- Now the process of rendering button icons are split into 2 phases: loading with lab_img_load() and creating scene-nodes for them with scaled_img_buffer_create(). This might incur some additional overhead since we no longer preload icon textures, but I believe its overhead won't be noticeable. In addition, the rendering of icon only happens for the first window as their backing buffers are shared (#2363).
- We had some complexity around placing icon buffer at the center of the space for titlebar button, and around calculating the geometry for hovered/rounded icon buffers. This PR simplifies it by always rendering icon on a buffer in a fixed geometry (`window.button.width`x`window.button.height`). This might incur some additional overhead for copying buffers or shrinking icons with `CAIRO_FILTER_GOOD` (which was previously handled by `buffer_convert_cairo_surface_for_icon()`), but I believe its overhead is also not noticeable and its simplification is worth it.

### DIFF
--- a/include/img/img-xbm.h
+++ b/include/img/img-xbm.h
@@ -5,18 +5,15 @@
 struct lab_data_buffer;
 
 /**
- * img_xbm_from_bitmap() - create button from monochrome bitmap
+ * img_xbm_load_from_bitmap() - create button from monochrome bitmap
  * @bitmap: bitmap data array in hexadecimal xbm format
- * @buffer: cairo-surface-buffer to create
  * @rgba: color
  *
  * Example bitmap: char button[6] = { 0x3f, 0x3f, 0x21, 0x21, 0x21, 0x3f };
  */
-void img_xbm_from_bitmap(const char *bitmap, struct lab_data_buffer **buffer,
-	float *rgba);
+struct lab_data_buffer *img_xbm_load_from_bitmap(const char *bitmap, float *rgba);
 
 /* img_xbm_load - Convert xbm file to buffer with cairo surface */
-void img_xbm_load(const char *filename, struct lab_data_buffer **buffer,
-	float *rgba);
+struct lab_data_buffer *img_xbm_load(const char *filename, float *rgba);
 
 #endif /* LABWC_IMG_XBM_H */
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -7,11 +7,13 @@
 #include "config.h"
 #include "common/mem.h"
 #include "common/scaled-font-buffer.h"
+#include "common/scaled-img-buffer.h"
 #include "common/scene-helpers.h"
 #include "common/string-helpers.h"
 #if HAVE_LIBSFDO
 #include "desktop-entry.h"
 #endif
+#include "img/img.h"
 #include "labwc.h"
 #include "node.h"
 #include "ssd-internal.h"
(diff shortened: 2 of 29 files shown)


tokyo4j:
Another discussion: How much does it cost to reload icon when the window enters a new output with a different scale?

@jlindgren90 stated:
https://github.com/labwc/labwc/blob/b2a45ac8afed0186a32e0b319c0481b7f16d4d17/src/ssd/ssd-titlebar.c#L600-L607

But if we accept this overhead, we can encapsulate the icon lookup in `scaled_img_buffer`, which will make it easier to support icons in menus and window-switcher.

jlindgren90:
If libsfdo has implemented icon theme caching well, then maybe reloading icons while dragging a window is fine. I grew up with "spinning rust" disks (not to mention floppies) and maybe my intuition for what is "slow" is out of date now. :)

johanmalm:
This looks fantastic. I really like the modifier-functions as well as separating out the texture creation. Tidy.

Have scanned through it a few times, but not yet tested or followed the logic closely. If nobody gets there before me, I'll review with a view to merge tomorrow.

> But if we accept this overhead, we can encapsulate the icon lookup in scaled_img_buffer, which will make it easier to support icons in menus and window-switcher.

I like the sounds of this. Assume that's for a different PR though?

include/img/img-xbm.h
```diff
@@ -4,19 +4,8 @@
 
 struct lab_data_buffer;
 
-/**
```
johanmalm, line 7: It would be good to keep comment documenting the function in `img-xbm.h`

Consolatis:
> If libsfdo has implemented icon theme caching well, then maybe reloading icons while dragging a window is fine. I grew up with "spinning rust" disks (not to mention floppies) and maybe my intuition for what is "slow" is out of date now. :)

As long as its only done once when changing scales and not repeatedly when moving a window from one scale to the other I think its fine.

include/img/img-xbm.h
tokyo4j: Alright. I once thought it could be removed as it's duplicated in `img.h`, but I agree that having a comment is better than none.

tokyo4j:
>> But if we accept this overhead, we can encapsulate the icon lookup in scaled_img_buffer, which will make it easier to support icons in menus and window-switcher.
>
> I like the sounds of this. Assume that's for a different PR though?

Yeah it should be done in another PR. My suggestion was just to see how `scaled_img_buffer` should be extended in the future. As it seems to be acceptable to reload icon on entering a new output, I'll open another PR with that approach.

johanmalm:
Have tested - works well.
LGTM.

src/ssd/ssd-titlebar.c
```diff
@@ -593,9 +598,9 @@ ssd_update_window_icon(struct ssd *ssd)
 	 * was considered, but these settings have distinct purposes
 	 * already and are zero by default.
 	 */
-	int hpad = theme->window_button_width / 10;
-	int icon_size = MIN(theme->window_button_width - 2 * hpad,
-		theme->window_button_height);
+	int icon_padding = theme->window_button_width / 10;
+	int icon_size = MIN(theme->window_button_width - 2 * icon_padding,
+		theme->window_button_height - 2 * icon_padding);
```
tokyo4j, line 603: Let me confirm before merging. @jlindgren90 Is there any reason for applying the padding around the icon horizontally, but not vertically? I think it's more straightforward to apply padding in both directions.

tokyo4j: Ah I see. I didn't think about the space between icons since I usually don't place buttons next to the window icon. 

For simplification, I'll keep the patch as it is.

PR #2446, opened 2024-12-23
src/output.c: remove output_usable_area_scaled()

tokyo4j:
I'm opening a PR as I just recalled an old discussion: #1620

`output_usable_area_scaled()` doesn't make sense; the geometry returned from `output_usable_area_in_layout_coords()` is already divided by the output scale, so we don't need to divide it again.


johanmalm:
LGTM. Have tested.

PR #2449, opened 2024-12-23
theme: demote padding.height warning

johanmalm:
...from WLR_ERROR to WLR_INFO to avoid 'false' logging when Openbox themes contain the padding.height settings. Whilst this option is not supported, the visual appearance is very close to that of Openbox with common themes.

The background to this is that Labwc theme specification has diverged slightly from that of Openbox with respect to titlebar padding to support more contemporary looks whilst avoiding breaking changes. For full details see commit: e16e78e7a4a6e40c8184edd7bb483b0f395482d0

Suggested-by: @Consolatis


tokyo4j:
Thanks!

Issue #2450, opened 2024-12-24
issues with application icon scaling after #2444


Flrian:
### Problem description

I've noticed that the way most of my application icons in the titlebar are smaller and some even get cut off after #2444.

As you can see in the screenshot, in the nested labwc instance running with the commit prior to the pr foot has a bigger application icon, while in the main instance it is smaller and the discord/vscode icons just get cut off. 

![2024-12-24T02:16:11](https://github.com/user-attachments/assets/f108a03a-282b-4414-8dbe-c6f0c789563c)

### Steps to reproduce

Might have something to do with the combination of window button height and titlebar padding in my theme?
```
window.button.height: 16
window.titlebar.padding.height: 8
window.titlebar.padding.width: 4
```

### labwc build source

Local build

### labwc version

16dbdc64e58d66011bbf319b92de844dab0ca8d9

### labwc environment

From a TTY or some display manager like lightdm

tokyo4j:
The cut-off icons should be fixed by #2451, but let me think about the problem of smaller icons for a while. Perhaps we should go back to the previous behavior that the padding of `window.button.width / 10` was applied only horizontally for window icons.

johanmalm:
Should be fixed with #2451.

tokyo4j:
The problem of smaller icon is not resolved yet.

Before #2444, the padding around icon with the size of (button size) / 10 is applied only horizontally, but #2444 changed it to apply in both directions as I felt it's more natural.

But I think we should at least provide a way for @Flrian to restore the original icon size. We can:
1. Make the padding configurable via `window.icon.padding` with default size of 2px
2. Or add a more sophisticated heuristic to determine the padding size based on `window.titlebar.padding.width`, `window.titlebar.padding.height` and `window.button.spacing`.
3. Or rollback to the original behavior that applies the padding only horizontally.

I've made a patch for (2.) like below:

<details><summary>patch</summary>

```diff
diff --git a/src/ssd/ssd-titlebar.c b/src/ssd/ssd-titlebar.c
index 9c24b431..ed58918f 100644
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -569,6 +569,41 @@ ssd_should_be_squared(struct ssd *ssd)
 		&& view->maximized != VIEW_AXIS_BOTH;
 }
 
+static int
+get_icon_padding(struct theme *theme)
+{
+	/*
+	 * First, get the space around the icon inside the button if no icon
[...]

PR #2451, opened 2024-12-24
img: fix scaling for rendering large image in non-square rectangle

tokyo4j:
Fix the cut-off icons reported in #2450. I think I was careless when writing that part.


johanmalm:
LGTM

Issue #1537, opened 2024-02-16
ToggleKeybinds: keyboardmodifier stick in wlfreerdp


dmitry-azaraev:
I'm use wlfreerdp and toggle keybinds with shortcut like W+F6. For testing it is easier to use S+F6 instead.

Once I'm press S+F6 - shift modifier stick in pressed state, until I'm did not press shift again (same for Super modifier).

However, I'm can't reproduce this in kitty, foot, chromium, so I'm very not sure what it is really labwc keyboard handling issue. Application which most requires this feature - possibly having buggy implementation, lol. :)

dmitry-azaraev:
I'm trace problem a bit, and the problem I'm think at labwc side, I guess.

For first normal Shift keydown/keyup sequence:
1. Shift pressed: `wl_keyboard_listener::key` - OK (indicates key pressed)
2. Shift pressed: bunch of `wl_keyboard_listener::modifiers` - OK
4. Shift released: `wl_keyboard_listener::key` - OK (indicates key released)
5. Shift released: bunch of `wl_keyboard_listener::modifiers` - OK 

When we use any key-bound labwc shortcut next situation happens (assuming we pressing something like Shift+F12), it can be bound to `Debug`, or `ToggleKeybinds` is not matters:
1. Shift pressed: `wl_keyboard_listener::key` - OK
2. Shift pressed: bunch of `wl_keyboard_listener::modifiers` - OK
3. F12 pressed: (event for shortcut key is suppressed is not generated) - OK
4. F12 released: (event for shortcut key is suppressed is not generated) - OK 
5. Shift released: `wl_keyboard_listener::key` - FAIL (key event is not generated)
6. Shift released: bunch of `wl_keyboard_listener::modifiers` - OK 

`wlfreerdp` rely on key event, is then sent to remote machine, while other software i guess did not require this event, they smart enough to "unpress" key from modifiers event(s) [but this kinda dirty workaround for something].

PS: I'm not sure - is `wl_keyboard_listener::modifiers` should really be sent for every modifier? There is 7 events every time generated, looks like event designed to deliver all state at once. However I'm not familiar with all this xkb stuff.

Consolatis:
Where do these `wl_keyboard_listener::{key,modifiers}` come from? `WAYLAND_DEBUG` from wlfreerdp or labwc internal callbacks?

dmitry-azaraev:
I'm just printf'ed this events in wlfreerdp directly.

[...]

PR #2452, opened 2024-12-24
Always send release events for modifier keys

tokyo4j:
Background: #2448

While not sending key-release events for modifier keys when they are bound to a keybind prevents Firefox to show its menu bar, it causes problems with some applications like Blender(#1507), wlfreerdp (#1537) or gzdoom (reported by 60IQ). #2377 solved this by adding `wantAbsorbedModifierReleaseEvents`, but forcing users to find it from our documentation and to apply it is not user-friendly.

So this PR reverts #1226 and #2377, and instead, sends `wl_keyboard.{leave,enter}` events on action invocations when key-press events for any modifiers have been sent. I confirmed this prevents firefox from showing its menu bar on Alt-* keybind and the issue with blender (#1507) doesn't happen. In addition, this PR prevents firefox from showing its menu bar also with the default Alt-Drag ~keybind~ mousebind.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -1301,4 +1301,18 @@ actions_run(struct view *activator, struct server *server,
 				" This is a BUG. Please report.", action->type);
 		}
 	}
+
+	/*
+	 * We send wl_keyboard.{leave,enter} to the clint when we have sent
+	 * key-press events to it since the last wl_keyboard.leave event.
+	 * This prevents firefox to show its menu bar when a keybind with
+	 * Alt key is executed.
+	 */
+	struct seat *seat = &server->seat;
+	if (seat->modifier_press_sent) {
+		struct wlr_surface *focused_surface =
+			seat->seat->keyboard_state.focused_surface;
+		seat_focus_surface(seat, NULL);
+		seat_focus_surface(seat, focused_surface);
+	}
 }
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -1099,18 +1099,6 @@ situation.
 	can be caused by *<margin>* settings or exclusive layer-shell clients
 	such as panels.
 
-*<windowRules><windowRule wantAbsorbedModifierReleaseEvents="">* [yes|no|default]
-	*wantAbsorbedModifierReleaseEvents* allows clients to receive modifier
-	release events even if they are part of keybinds. Most clients should
-	not receive these, but some (for example blender) need it in some
-	situations.
-
-```
-<windowRules>
-  <windowRule identifier="blender" wantAbsorbedModifierReleaseEvents="yes"/>
-</windowRules>
-```
-
 ## MENU
 
 ```
--- a/include/input/key-state.h
+++ b/include/input/key-state.h
@@ -19,7 +19,7 @@
 uint32_t *key_state_pressed_sent_keycodes(void);
 int key_state_nr_pressed_sent_keycodes(void);
 
-void key_state_set_pressed(uint32_t keycode, bool is_pressed, bool is_modifier);
+void key_state_set_pressed(uint32_t keycode, bool is_pressed);
 void key_state_store_pressed_key_as_bound(uint32_t keycode);
 bool key_state_corresponding_press_event_was_bound(uint32_t keycode);
 void key_state_bound_key_remove(uint32_t keycode);
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -148,6 +148,14 @@ struct seat {
 
 	struct lab_set bound_buttons;
 
+	/*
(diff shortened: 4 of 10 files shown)


src/action.c
```diff
@@ -1301,4 +1301,18 @@ actions_run(struct view *activator, struct server *server,
...
 	}
+
+	/*
+	 * We send wl_keyboard.{leave,enter} to the clint when we have sent
+	 * key-press events to it since the last wl_keyboard.leave event.
+	 * This prevents firefox to show its menu bar when a keybind with
+	 * Alt key is executed.
+	 */
+	struct seat *seat = &server->seat;
+	if (seat->modifier_press_sent) {
```
jlindgren90, line 1312: This is only necessary for keyboard driven actions, right? I think actions_run() is also called on every mouse click in the default config.

tokyo4j: Well, sending `wl_keyboard.{leave,enter}` also prevents firefox to show menu bar with the default Alt-Drag mousebind. You can see the current problem by running labwc with the default config and dragging firefox window with Alt key pressed.

But I agree that sending those events on every mouse click (by `Focus` & `Raise` mousebinds) doesn't feel right. Perhaps we should send those events only with keybinds and we can clear keyboard focus separately in `interactive_begin()`.

jlindgren90: That sounds like a good approach 👍

tokyo4j: Hmm... sorry, I was totally wrong in this part:

> But I agree that sending those events on every mouse click (by Focus & Raise mousebinds) doesn't feel right. 

In this PR, leave & enter events are only sent when any modifier is pressed, so the default Focus & Raise mousebind doesn't trigger leave & enter events.

Furthermore, I came up with another case where sending leave & enter events also after mousebinds is beneficial; consider following mousebinds:

```xml
    <context name="All">
      <mousebind action="Scroll" direction="A-Up">
        <action name="ZoomIn" />
      </mousebind>
      <mousebind action="Scroll" direction="A-Down">
        <action name="ZoomOut" />
      </mousebind>
    </context>
```

In this case, I guess the user doesn't want firefox to show menu bar after updating magnification rate.

Therefore, now I think putting the logic for sending leave & enter events in `actions_run()` is a better idea.

jlindgren90: That's a good thought. It still seems like we should send the enter/leave when the modifier is part of a compositor action (like the magnifier thing) but not when it's a client action (like Ctrl+Scroll to zoom). I'm not sure of the best mechanism to do that.

tokyo4j: Sorry, I don't get what you mean. In this PR, leave & enter events are sent when a keybind or mousebind is executed while a modifier has been pressed. So by default, we never send leave & enter when user presses Ctrl+Scroll. Of course, when the user sets mousebinds like `<mousebind action="Scroll" direction="C-Up">` it may interfere with the client-side action, but I think that can be helped and the user should not set such a mousebind anyway.

My current concerns with this PR are:
- It deviates from what other compositors like KWin do. So _perhaps_ it causes issues with applications that don't expect the behavior in this PR.
- It sometimes sends unnecessary leave & enter events. For example, we practically don't need to send those events when launching a new terminal window because a leave event is sent to the original window anyway when moving the keyboard focus to the new window. But in this PR, we send leave -> enter -> leave events to the original window, which feels a bit strange.

But overall, I think this PR does its job well.

jlindgren90: Ctrl+Scroll may have been a bad example. In my setup I like to have Ctrl+Click of any mouse button bound to "Focus" and "Raise". But I don't want every Ctrl+Click to send keyboard enter & leave events.

tokyo4j: OK. I see sending leave & enter events only for keybinds is preferable on your setup.

However, now I wonder if we have a good reason to differentiate keybinds vs mousebinds in regard to whether to send the leave & enter events, as both keybinds and mousebinds can unexpectedly open firefox's menu bar.

The more I consider this problem, the more I think we should take the option (2) in #2448: just send a leave event on window-switcher activation. This only works for the default Alt-Tab and will reintroduce the firefox's menu bar issue with some existing keybinds like:

```xml
    <keybind key="A-Left">
      <action name="MoveToEdge" direction="left" />
    </keybind>
```

But Kwin also has this problem and I think it's a much more minor issue than that with Alt-Tab keybind. As we advertise labwc as a lightweight and simple compositor, I prefer this way.

I'll open another PR with this approach.

jlindgren90: I'm fine with that. Thanks for all the work on this - it's tricky to get right.

I think the leave/enter approach could also still work, if we sorted out what actions it should or shouldn't trigger for. But maybe it needs more thought.

tokyo4j:
Superseded by #2455

PR #2455, opened 2024-12-27
Always send modifier release events, and clear focus on Move/Resize, window switcher and menu interactions

tokyo4j:
First, this PR reverts #1226 and #2377, because:
- Not sending modifier release events (#1226) for keybinds (for not showing firefox's menu bar) is very problematic with some applications (#1537, #1507).
- Forcing users to define `wantAbsorbedModifierReleaseEvents` window rule (#2377) to fix it is also undesirable, since it doesn't provide a good out-of-the-box experience and requires users to read through our documentation.

After some discussion in #2448 and #2452, we decided to follow KWin's approach that always sends modifier release events, but clears keyboard focus when activating window switcher (typically triggered by Alt-Tab) to prevent firefox from showing menu bar. Of course, this doesn't work for other actions like `Execute` and makes existing Alt-* keybinds show firefox's menu bar, but we concluded this approach is the most robust.

So in this PR, keyboard and pointer focus are temporarily cleared while Move/Resize, window switcher and menu interactions and restored after them, which basically follows KWin's behavior.

We slightly deviate from KWin as KWin doesn't clear the keyboard focus while Move/Resize, but it solves our existing problem that Firefox shows its menu bar after dragging it with default Alt-Drag mousebind, and this is what Mutter does.

Testing is appreciated.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -116,6 +116,16 @@ struct seat {
 		double x, y;
 	} smooth_scroll_offset;
 
+	/*
+	 * The surface whose keyboard focus is temporarily cleared with
+	 * seat_focus_override_begin() and restored with
+	 * seat_focus_override_end().
+	 */
+	struct {
+		struct wlr_surface *surface;
+		struct wl_listener surface_destroy;
+	} focus_override;
+
 	struct wlr_pointer_constraint_v1 *current_constraint;
 
 	/* In support for ToggleKeybinds */
@@ -273,9 +283,13 @@ struct server {
 	 * 'active_view' is generally the view with keyboard-focus, updated with
 	 * each "focus change". This view is drawn with "active" SSD coloring.
 	 *
-	 * The exception is when a layer-shell client takes keyboard-focus in
-	 * which case the currently active view stays active. This is important
-	 * for foreign-toplevel protocol.
+	 * The exceptions are:
+	 * - when a layer-shell client takes keyboard-focus in which case the
+	 *   currently active view stays active
+	 * - when keyboard focus is temporarily cleared for server-side
+	 *   interactions like Move/Resize, window switcher and menus.
+	 *
+	 * Note that active_view is synced with foreign-toplevel clients.
 	 */
 	struct view *active_view;
 	/*
@@ -502,6 +516,20 @@ void seat_set_focus_layer(struct seat *seat, struct wlr_layer_surface_v1 *layer)
 void seat_set_pressed(struct seat *seat, struct cursor_context *ctx);
 void seat_reset_pressed(struct seat *seat);
 void seat_output_layout_changed(struct seat *seat);
+
+/*
+ * Temporarily clear the pointer/keyboard focus from the client at the
+ * beginning of interactive move/resize, window switcher or menu interactions.
+ * The focus is kept cleared until seat_focus_override_end() is called or
+ * layer-shell/session-lock surfaces are mapped.
+ */
+void seat_focus_override_begin(struct seat *seat, enum input_mode input_mode,
+	enum lab_cursors cursor_shape);
+/*
+ * Restore the pointer/keyboard focus which was cleared in
+ * seat_focus_override_begin().
+ */
+void seat_focus_override_end(struct seat *seat);
 
 /**
  * interactive_anchor_to_cursor() - repositions the geometry to remain
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1336,8 +1336,9 @@ menu_open_root(struct menu *menu, int x, int y)
 	menu_configure(menu, (struct wlr_box){.x = x, .y = y});
 	wlr_scene_node_set_enabled(&menu->scene_tree->node, true);
 	menu->server->menu_current = menu;
-	menu->server->input_mode = LAB_INPUT_STATE_MENU;
 	selected_item = NULL;
+	seat_focus_override_begin(&menu->server->seat,
+		LAB_INPUT_STATE_MENU, LAB_CURSOR_DEFAULT);
 }
 
 static void
@@ -1628,8 +1629,7 @@ menu_execute_item(struct menuitem *item)
 	 */
 	struct server *server = item->parent->server;
 	menu_close(server->menu_current);
-	server->input_mode = LAB_INPUT_STATE_PASSTHROUGH;
-	cursor_update_focus(server);
+	seat_focus_override_end(&server->seat);
 
 	/*
 	 * We call the actions after closing the menu so that virtual keyboard
@@ -1739,7 +1739,7 @@ menu_close_root(struct server *server)
 		server->menu_current = NULL;
 		destroy_pipemenus(server);
 	}
-	server->input_mode = LAB_INPUT_STATE_PASSTHROUGH;
+	seat_focus_override_end(&server->seat);
 }
 
 void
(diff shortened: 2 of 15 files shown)


tokyo4j:
"Temporarily clearing keyboard focus" in a reliable way is harder than I thought... I'll un-draft this PR when it's ready.

tokyo4j:
I think now this is ready. I'll daily-drive this PR this from today.

jlindgren90:
The code looks sane to me. I have not tested it yet.

include/labwc.h
```diff
@@ -115,6 +115,14 @@ struct seat {
...
 	} smooth_scroll_offset;
 
+	/*
+	 * The surface whose keyboard focus is temporarily cleared with
+	 * seat_begin_server_action() and restored with
+	 * seat_finish_server_action().
+	 */
+	struct wlr_surface *last_focused_surface;
+	struct wl_listener last_focused_surface_destroy;
+
```
tokyo4j, line 125: Note: Perhaps this can be substituted by `server->active_view` and `seat->focused_layer`, but an advantage of this approach is that it handles all types of surfaces including unmanaged-xwayland surfaces.

johanmalm: IIRC `view->active_view` isn't always in sync either (e.g. when override redirect surfaces are focused). I might be getting my wires crossed on that one though.

src/menu/menu.c
```diff
@@ -1625,7 +1626,8 @@ menu_execute_item(struct menuitem *item)
 	struct server *server = item->parent->server;
 	menu_close(server->menu_current);
 	server->input_mode = LAB_INPUT_STATE_PASSTHROUGH;
-	cursor_update_focus(server);
+	seat_finish_server_action(&server->seat,
```
johanmalm, line 1629: Should we consider setting `server->input_mode` in `seat_{begin,finish}_server_action()` functions? They might be set in other places - not looked - but it would feel like a neat thing to do.

include/labwc.h
```diff
@@ -115,6 +115,14 @@ struct seat {
 		double x, y;
 	} smooth_scroll_offset;
 
+	/*
+	 * The surface whose keyboard focus is temporarily cleared with
+	 * seat_begin_server_action() and restored with
+	 * seat_finish_server_action().
+	 */
+	struct wlr_surface *last_focused_surface;
```
johanmalm, line 123: Consider wrapping these two variables in a `struct { } seat_server_mode` or something like that for ease of reading.

johanmalm (commented):
Have just left some quick comments for now.
Will use this PR for work tomorrow and then do a deeper review to follow the logic through.

include/labwc.h
jlindgren90: Just agreeing with both - I like the last_focused_surface approach as @tokyo4j has it. The active view vs. focused surface duality is annoying but I think we have to deal with it, because of the non-view surface types.

src/seat.c
```diff
@@ -794,3 +814,54 @@ seat_output_layout_changed(struct seat *seat)
...
+handle_last_focused_surface_destroy(struct wl_listener *listener, void *data)
+{
+	struct seat *seat = wl_container_of(listener, seat,
+		last_focused_surface_destroy);
+	wl_list_remove(&seat->last_focused_surface_destroy.link);
+	seat->last_focused_surface = NULL;
+}
+
+void
+seat_begin_server_action(struct seat *seat, enum lab_cursors cursor_shape)
```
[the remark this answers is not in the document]
jlindgren90: seat_focus_override_begin/end?

"action" and "mode" are both so heavily overloaded already...

src/menu/menu.c
tokyo4j: Yeah you're absolutely right, and that's what I described as TODO in the header file. As I described there, I think the cleanest option is to add `LAB_INPUT_STATE_WINDOW_SWITCHER` in the `input_mode` enum and set it in those functions, but I didn't in this PR since it's kind of unpredictable how it affects other places like `input_mode != LAB_INPUT_STATE_PASSHROUGH` checks.

I'm thinking of opening another PR for just adding `LAB_INPUT_STATE_WINDOW_SWITCHER`.

src/seat.c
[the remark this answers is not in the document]
tokyo4j: Another option is `seat_set_input_mode()` and `seat_clear_input_mode()`, which depends on the change suggested in https://github.com/labwc/labwc/pull/2455#discussion_r1898728253 and #2459.

But I really like seat_focus_override_begin/end as it tells exactly what they do. I'll rename to it.

tokyo4j:
I added a condition in `cursor_update_common()` to ensure pointer focus is kept cleared while server-side interactions.

Also, I guess the following code is no longer needed, because it was a workaround to make clients update the cursor on finishing Move/Resize while keeping pointer focus on the client:

https://github.com/labwc/labwc/blob/1550eb9c26fcf756c3915e8ab9a777245c8f558c/src/input/cursor.c#L530-L548

@jlindgren90 @johanmalm Is my assumption right?

jlindgren90:
Hmm, I am not sure about the pointer focus bits. I am less familiar with how that works.

include/labwc.h
tokyo4j: Good point. Wrapped with `focus_override`.

src/menu/menu.c
tokyo4j: Done, with help of #2459.

tokyo4j:
I rebased this PR on #2459, and addressed the comments. I removed the part I quoted above.

johanmalm:
LGTM - other than rebasing against master now that #2459 has been merged.

Have run with it all day and it seems fine.

I suggest that we go for a merge soon. That way we're relatively early in the release cycle which gives us time to catch any unintended glitches.

Many thanks for working on this. It's really nice to get the fundamentals right.

johanmalm:
> Also, I guess `wlr_seat_pointer_notify_clear_focus()` in the following code is no longer needed, because it was a workaround to make clients update the cursor on finishing Move/Resize while keeping pointer focus on the client:

> Is my assumption right?

I think it's fine. Have had some git-blaming to see if those commits help job my memory, but can't think of anything else.

Getting all the keyboard-focus + cursor-updates right is quite finicky, but if we've missed some odd edge-case, we'll re-discover it and can document it better next time.

tokyo4j:
> I suggest that we go for a merge soon. That way we're relatively early in the release cycle which gives us time to catch any unintended glitches.

> Getting all the keyboard-focus + cursor-updates right is quite finicky, but if we've missed some odd edge-case, we'll re-discover it and can document it better next time.

Yeah I agree. I realized it's almost impossible to 100% make sure this doesn't cause any regression.

As I described as a `FIXME`, a known issue is that the focus is transiently given to the old surface at the end of window-switching. But I guess it won't cause any problems with most apps.

tokyo4j:
The changes in this PR will be described in the next release note like:

```
CHANGED:
- When a keybind is triggered while focusing on an application, the key-release event of the modifier key
  is now sent to the application. This fixes the problem of stuck modifier keys in some applications
  including Blender and wlfreerdp.
- Removed "wantAbsorbedModifierReleaseEvents" window rule as it's no longer needed
- Now the keyboard/pointer focus is cleared while a window is dragged, the window switcher is activated
  or a menu is opened.
```

Issue #2453, opened 2024-12-26
Crash when using window switcher


Flrian:
### Problem description

labwc crashes for me when switching from a nested mutter window (that is being run in an arch nspawn container) via the window switcher to a terminal window.

<details>
<summary>log</summary>

```
00:00:01.033 [DEBUG] [../src/input/keyboard.c:252] keycode matched
00:00:01.033 [DEBUG] [../src/action.c:894] Handling action 5: Execute
warn: fcft.c:583: failed to get full font name
warn: wayland.c:1619: compositor does not implement the XDG toplevel icon protocol
warn: fcft.c:583: failed to get full font name
warn: fcft.c:583: failed to get full font name
warn: fcft.c:583: failed to get full font name
warn: fcft.c:583: failed to get full font name
00:00:01.057 [DEBUG] [types/wlr_compositor.c:771] New wlr_surface 0x5190015e0080 (res 0x50c00026c080)
00:00:01.057 [DEBUG] [types/wlr_compositor.c:771] New wlr_surface 0x5190015e0580 (res 0x50c00026c380)
00:00:01.057 [DEBUG] [types/xdg_shell/wlr_xdg_surface.c:401] new xdg_surface 0x512000082e40 (res 0x50c00026c680)
00:00:01.057 [DEBUG] [types/wlr_xdg_decoration_v1.c:216] new xdg_toplevel_decoration 0x5110001c6dc0 (res 0x50c00026ca40)
00:00:01.057 [DEBUG] [../src/desktop-entry.c:320] loading icon file /usr/share/icons/hicolor/scalable/apps/foot.svg
00:00:01.058 [DEBUG] [../src/xdg.c:131] scheduling configure
00:00:01.058 [DEBUG] [types/wlr_compositor.c:771] New wlr_surface 0x5190015e1480 (res 0x50c00026df40)
00:00:01.062 [DEBUG] [render/vulkan/renderer.c:326] Created new vk staging buffer of size 2800000
00:00:01.065 [DEBUG] [../src/view-impl-common.c:57] [map] identifier=foot, title=foot
00:00:10.671 [INFO] [xwayland/server.c:107] Starting Xwayland on :0
00:00:10.723 [DEBUG] [types/wlr_compositor.c:771] New wlr_surface 0x5190015ebe80 (res 0x50c00027afc0)
00:00:10.723 [DEBUG] [xwayland/server.c:272] Xserver is ready
[...]

PR #2457, opened 2024-12-27
view: fix NULL string_prop crash

johanmalm:
...when app_id is NULL.

Make sure view_get_string_prop() never returns NULL because it is so easy to misuse. Same for the respective xwayland/xdg impl methods in case anyone decides to (incorrectly) call them directly in future.

Fixes: #2453

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -3,6 +3,7 @@
 #include <wlr/types/wlr_scene.h>
 #include "common/graphic-helpers.h"
 #include "common/scene-helpers.h"
+#include "common/string-helpers.h"
 #include "debug.h"
 #include "input/ime.h"
 #include "labwc.h"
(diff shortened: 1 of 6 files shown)


Consolatis:
If we go this route we can remove quite some checks for it being NULL. If you are fine with it I could push the cleanup to this PR as separate commit.

johanmalm:
> If we go this route we can remove quite some checks for it being NULL. If you are fine with it I could push the cleanup to this PR as separate commit.

Yes - good plan.

tokyo4j:
I don't have strong preference on `NULL` vs `""`, but if we choose `""`, I think we should add `string_null_or_empty()` check and early-return in `desktop_entry_icon_lookup()` as an easy optimization.

johanmalm:
I think we're aligned on this one, but looking for a LGTM before I merge.

Consolatis:
Original  commit LGTM. Added another one to deal with the ensured non-null return value, including in `desktop_entry_icon_lookup()` and `desktop_entry_name_lookup()`.

johanmalm:
Thanks. Second commit LGTM.

PR #2459, opened 2024-12-28
Add LAB_INPUT_STATE_WINDOW_SWITCHER

tokyo4j:
...as a preparation for #2455.

In #2455, all of following operations clear keyboard/pointer focus:
- Activating window switcher
- Starting Move/Resize
- Opening menus

So adding `LAB_INPUT_STATE_WINDOW_SWITCHER` into `input_mode` enum will make sense.

In addition, the third commit prevents starting window switcher or opening menus while `input_mode` is set. This will make the transition of the server state more predictable and prevent potential bugs.


tokyo4j:
Note that this change affects all the early-returns like `if(server->input_mode != LAB_INPUT_STATE_PASSTHROUGH)`. This includes:
- `request_cursor_notify()`
- `request_cursor_notify()`
- `request_cursor_notify()`
- `cursor_process_button_release()`
- `show_menu()`
- `desktop_focus_output()`

But I think it's safe to early-return in these sites.

johanmalm:
LGTM. Thanks. This looks altogether more consistent.

PR #2463, opened 2024-12-29
keyboard: set numlock to off by default

johanmalm:
...to avoid surprises on Acer Aspire One laptops where the numeric keyboard does not physically exist but "overlaps" the qwerty keyboard.

Reported-by: staryvyr on IRC

Note: We do not like changing defaults, but feel that there is a good reason for this one.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -565,7 +565,7 @@ extending outward from the snapped edge.
 
 *<keyboard><numlock>* [on|off]
 	When recognizing a new keyboard enable or disable Num Lock.
-	Default is on.
+	Default is off.
 
 *<keyboard layoutScope="">* [global|window]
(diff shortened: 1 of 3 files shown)


Staryvyr:
I observed this problem on the Acer Aspire One D 270 and on the Acer Aspire One D 255 E.

tokyo4j:
LGTM. I remember I was a bit surprised when numlock was enabled on startup. Sway also disables numlock by default.

Consolatis:
I am not convinced that either on or off should be set as default. This new default now makes numpads not work while the old default breaks non-numpad keyboards. Maybe this should be a three-state instead, with the default being unspecified which would then simply not touch numlock? The alternative would be to somehow figure out if the keyboard does have a numpad and use a default based on that.

I wonder how other compositors / X11 environments deal with this.

CC @jlindgren90

PR #2464, opened 2024-12-29
osd: some refactor

tokyo4j:
1. Remove the call to `osd_review_restore()` in `desktop_cycle_view()`. I guess this is just a cruft from days when the next view to cycle is determined by traversing scene-nodes (ref. #1479).
2. Make `osd_preview_restore()` private in `osd.c` and call it inside `osd_finish()`. I renamed it to `restore_preview_node()`.
3. Replace `seat->server->osd_state.cycle_view || seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH` with just `seat->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH`. I missed to do this when rebasing #2455.
4. Refactor to not update values in `server->osd_state` outside `osd.c` by adding `osd_begin()` and `osd_cycle()`.
5. Make `desktop_cycle_view()` private in `osd.c`, renaming it to `get_next_cycle_view()`.

The primary goal in this PR is (4). In the future, `osd_begin()` will create the scene graph for the window switcher (rather than drawing it on a single texture) and `osd_cycle()` will update the scene-node of the highlight for the selected window. This will make it easier to add mouse interactions or other forms of window switcher like #2172.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -482,14 +482,6 @@ struct view *desktop_topmost_focusable_view(struct server *server);
 void desktop_update_top_layer_visiblity(struct server *server);
 
 /**
- * desktop_cycle_view - return view to 'cycle' to
- * @start_view: reference point for finding next view to cycle to
- * Note: If !start_view, the second focusable view is returned
- */
-struct view *desktop_cycle_view(struct server *server, struct view *start_view,
-	enum lab_cycle_dir dir);
-
-/**
  * desktop_focus_topmost_view() - focus the topmost view on the current
  * workspace, skipping views that claim not to want focus (those can
  * still be focused by explicit request, e.g. by clicking in them).
--- a/include/osd.h
+++ b/include/osd.h
@@ -43,9 +43,6 @@ void osd_update(struct server *server);
 /* Closes the OSD */
 void osd_finish(struct server *server);
 
-/* Moves preview views back into their original stacking order and state */
-void osd_preview_restore(struct server *server);
-
 /* Notify OSD about a destroying view */
 void osd_on_view_destroy(struct view *view);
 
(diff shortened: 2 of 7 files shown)


johanmalm:
LGTM. Not tested.

Consolatis:
LGTM to me as well.

What are your plans for adding other osd render backends, some `impl` struct like for `view->impl`?

E.g. API interface for rest of the code base and some generic functions in `osd.c` + the current OSD in something like `osd_simple.c` and the preview one in `osd_fancy.c` + the implementations would then only fill the impl struct with function pointers which are then called by `osd.c`?

tokyo4j:
> What are your plans for adding other osd render backends, some impl struct like for view->impl?

> E.g. API interface for rest of the code base and some generic functions in osd.c + the current OSD in something like osd_simple.c and the preview one in osd_fancy.c + the implementations would then only fill the impl struct with function pointers which are then called by osd.c?

Yeah that's what I'm thinking of. The functions in the impl struct will be like:

1. Function that initializes the scene-graph for window switcher, called from `osd_begin()`.
2. Function that destroys the scene-graph, called from `osd_finish()`.
3. Function that updates the selection of the cycled window, called from `osd_finish()`.
4. Function that processes mouse hover events (which takes the `cursor_context.node` as an argument?)

Consolatis:
Sounds good. Using the whole `struct cursor_context` as argument rather than just the scene node may be more future proof. We may also need to handle clicks and potentially wheel events if we want to allow mouse control of the OSD. The state tracking might get slightly annoying though, not sure.

In the menu implementation we added some generic selection functions and then some key event handlers and mouse event handlers which call those. That made the selection logic itself share most of the code, regardless how the selection was triggered. Maybe we can do the same or something similar with the OSD.

PR #2473, opened 2024-12-31
keyboard: cancel keybind_repeat on reconfigure

Consolatis:
Currently we may end up in an endless loop of Reconfigure requests if the Reconfigure action was called by a keybind. If the reconfigure takes too long (which may happen on slow systems with libsfdo full debug logging for example) the reconfigure might be triggered again and again.

To prevent that, simply cancel all keybind_repeat timers on reconfigure.

---

Only slightly tested.


tokyo4j:
LGTM.

Issue #2475, opened 2024-12-31
Keyboard focus is moved to the surface below the cursor on finishing window switcher with followMouse="yes" & followMouseRequiresMovement="no"


tokyo4j:
### Problem description

#2455 reintroduced #849.

In `osd_finish()`, we call `seat_focus_override_end()` which clears `server->input_mode` and then call `cursor_update_focus()` which moves the keyboard focus to the surface below the cursor with `followMouse="yes"`, `followMouseRequiresMovement="no"` and `server->input_mode == PASSTHROUGH`.

### Steps to reproduce

1. Set `followMouse="yes"` and `followMouseRequiresMovement="no"` in `rc.xml`.
2. Open 2 windows and put cursor on one of them.
3. Start window switcher and try to move the cursor on the other window.
4. Keyboard focus stays on the window.

### labwc build source

Local build

### labwc version

0.8.2-37-gbad788cc-dirty

### labwc environment

From a TTY or some display manager like lightdm

tokyo4j:
Perhaps we don't need to consider this issue as a bug since #849 is an issue from days when `followMouseRequiresMovement` was always `no`, if I get it correctly.

I even wonder if there are users who actually prefer `followMouseRequiresMovement=no`... Maybe we can remove that option?

Consolatis:
I think it is actually more correct now. The default for `followMouseRequiresMovement` is 'yes' anyway and I assume that is what most people who want to use follow mouse in the first place will stick with. The other mode is a more strict variant with the idea of 'focus-what-is-under-the-cursor-at-all-times` which is what happens now. So I think we are fine as is.

tokyo4j:
OK, I agree with you now.

Then we should describe this change in the 0.8.3 release note. @johanmalm

tokyo4j:
For what it's worth, I realized that window switcher becomes almost unusable with  `<focus followMouse="yes" followMouseRequiresMovement="no" raiseOnFocus="yes" />`, because the window below the cursor is always raised, discarding the selection with window switcher. But the experience with that configuration before #2459 was already bad  since the the window below the cursor was raised as soon as the cursor is moved.

Anyway, there would be little amount of users who prefers `<focus followMouse="yes" followMouseRequiresMovement="no" raiseOnFocus="yes" />`, so this can be addressed later.

johanmalm:
> Then we should describe this change in the 0.8.3 release note.

[...]

PR #2476, opened 2024-12-31
Remove cruft related to window switcher and keyboard focus

tokyo4j:
This PR cleans up the cruft and comments from e45fe08 (ref. #2475).

Background:

- With e45fe08, the keyboard focus is always moved to the switched window on finishing window switcher, even with `<focus followMouse="yes">`. Since `followMouseRequiresMovement` was not implemented at that time (behaved like it was always "no"), e45fe08 was necessary to allow users users who use `followMouse="yes"` to move the keyboard focus with window switcher.
- 9a9e20d added `followMouseRequiresMovement`, but it kept the behavior described above even with `followMouse="yes"` and `followMouseRequiresMovement="no"`.
- 398b80b accidentally invalidated e45fe08, which means the keyboard focus is now always moved to the window below the cursor on finishing window switcher with `followMouse="yes"` and `followMouseRequiresMovement="no"`.

Although the invalidation was a accident, I think always setting the keyboard focus on the window below the cursor is what users expect from `followMouse="yes"` and `followMouseRequiresMovement="no"`.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -637,10 +637,11 @@ _cursor_update_focus(struct server *server)
 	struct cursor_context ctx = get_cursor_context(server);
 
 	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse
-			&& !rc.focus_follow_mouse_requires_movement
-			&& server->input_mode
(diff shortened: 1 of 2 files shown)


Consolatis:
Thanks, tested and didn't spot any issue.

PR #2477, opened 2024-12-31
osd: make window switcher more Openbox-like in terms of key precessing

### LINKED ISSUE #2462: window switching order stays the same as when first started
### Problem description

When you press a key that has NextWindow or PreviousWindow as an action the windows are always rotated in the initial direction even when the order is reversed.

### Steps to reproduce

1. Assign keys for NextWindow and PreviousWindow in rc.xml for example W-n and W-p
2. Press one of them, while keeping the modifier key pressed, e.g. W-n
3. While still keeping the modifier key pressed press the opposite direction key, e.g. W-n
4. The rotation direction stays the same

### labwc build source

Release

### labwc version

0.8.2

### labwc environment

### Distribution

Arch Linux

tokyo4j:
Closes #2462.

Before this PR, keystrokes were interpreted based on following hard-coded rules while the window switcher is active:

  1. Up/Left arrow keys cycle the window forward.
  2. Down/Right arrow keys cycle the window backward.
  3. Other keystrokes cycle the window in the initial direction specified by `NextWindow`/`PreviousWindow` actions. But while Shift key is pressed, the direction is inverted.

...and keybind actions were never executed.

However, this lead to a counter-intuitive behavior for new, especially pre-Openbox users. For example, in the following keybinds, after the user activates the window switcher with `Super+n`, `Super+p` cycles the window _forward_:

```xml
  <keybind key="W-n">
    <action name="NextWindow" />
  </keybind>
  <keybind key="W-p">
    <action name="PreviousWindow" />
  </keybind>
```

This is because the key `p` is recognized just as a normal key in the third hard-coded rule.

So this PR changes the rules to be more Openbox-like:

  1. Up/Left arrow keys cycles the window forward.
  2. Down/Right arrow keys cycles the window backward.
  3. Other keystrokes are matched against keybinds and execute their actions. If they include NextWindow/PreviousWindow action, it cycles the selected window forward/backward even while the window switcher is active.

To keep using Alt-Shift-Tab to cycle the window backward, this PR adds following keybind to the default setting:
```xml
    <keybind key="A-S-Tab">
      <action name="PreviousWindow" />
    </keybind>
```

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -447,58 +447,35 @@ handle_menu_keys(struct server *server, struct keysyms *syms)
 	}
 }
 
-static void
-toggle_direction(enum lab_cycle_dir *direction)
-{
-	if (*direction == LAB_CYCLE_DIR_FORWARD) {
-		*direction = LAB_CYCLE_DIR_BACKWARD;
-	} else if (*direction == LAB_CYCLE_DIR_BACKWARD) {
-		*direction = LAB_CYCLE_DIR_FORWARD;
-	}
-}
-
-static void
+/* Returns true if the keystroke is consumed */
+static bool
 handle_cycle_view_key(struct server *server, struct keyinfo *keyinfo)
 {
-	for (int i = 0; i < keyinfo->translated.nr_syms; i++) {
-		if (keyinfo->translated.syms[i] == XKB_KEY_Escape) {
-			/* cancel view-cycle */
-			osd_finish(server);
-			return;
-		}
+	if (keyinfo->is_modifier) {
+		return false;
 	}
 
 	/* cycle to next */
-	if (!keyinfo->is_modifier) {
-		enum lab_cycle_dir direction = server->osd_state.initial_direction;
-		for (int i = 0; i < keyinfo->translated.nr_syms; i++) {
-			if (keyinfo->translated.syms[i] == XKB_KEY_Up
-					|| keyinfo->translated.syms[i] == XKB_KEY_Left) {
-				direction = LAB_CYCLE_DIR_BACKWARD;
-				goto miss_shift_toggle;
-			}
-			if (keyinfo->translated.syms[i] == XKB_KEY_Down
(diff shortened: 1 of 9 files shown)


Consolatis:
> To keep using Alt-Shift-Tab to cycle the window backward, this PR adds following keybind to the default setting:

This is a breaking change for users who do no use `<default />` but I think its still the right change to make.

src/input/keyboard.c
```diff
@@ -447,58 +447,36 @@ handle_menu_keys(struct server *server, struct keysyms *syms)
...
-				direction = LAB_CYCLE_DIR_BACKWARD;
-				goto miss_shift_toggle;
-			}
-			if (keyinfo->translated.syms[i] == XKB_KEY_Down
-					|| keyinfo->translated.syms[i] == XKB_KEY_Right) {
-				direction = LAB_CYCLE_DIR_FORWARD;
-				goto miss_shift_toggle;
-			}
+	for (int i = 0; i < keyinfo->translated.nr_syms; i++) {
+		if (keyinfo->translated.syms[i] == XKB_KEY_Up
```
Consolatis, line 468: Looks like we could combine this loop with the `Escape` one on top and move the `is->modifier` check before that one.

tokyo4j: Nice suggestion. Fixed.

johanmalm:
LGTM. Have tested and it works well.

Suggest amending documentation with the following:

1. In `labwc-actions(5)` we ought to remove `The shift key is used to reverse direction while cycling.` in the `PreviousWindow` section. See https://labwc.github.io/labwc-actions.5.html#entry_action_name=nextwindow
2. In `labwc-config(1)` under `<keyboard><default/>` we should add `A-S-Tab - previous window`. See https://labwc.github.io/labwc-config.5.html#entry_keyboard_default

As an observation, this now means that we honour keybinds during window-switching. Whilst window-switching I can press `C-A-t` to issue `foot`. I don't think this is a problem - and if it turns out to be one, we could patch that later.

I really like this change. It is more intuitive :smile:

PR #2478, opened 2024-12-31
config/rcxml.c: encapsulate parser state into a struct

orfeasxyz:
Staging for https://github.com/labwc/labwc/pull/2427.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -41,33 +41,35 @@
 #define LAB_WLR_VERSION_OLDER_THAN(major, minor, micro) \
 	(WLR_VERSION_NUM < (((major) << 16) | ((minor) << 8) | (micro)))
 
-static bool in_regions;
-static bool in_usable_area_override;
-static bool in_keybind;
-static bool in_mousebind;
-static bool in_touch;
-static bool in_libinput_category;
-static bool in_window_switcher_field;
-static bool in_window_rules;
-static bool in_action_query;
-static bool in_action_then_branch;
-static bool in_action_else_branch;
-static bool in_action_none_branch;
+struct parser_state {
+	bool in_regions;
+	bool in_usable_area_override;
+	bool in_keybind;
+	bool in_mousebind;
+	bool in_touch;
+	bool in_libinput_category;
+	bool in_window_switcher_field;
+	bool in_window_rules;
+	bool in_action_query;
+	bool in_action_then_branch;
+	bool in_action_else_branch;
+	bool in_action_none_branch;
+	struct usable_area_override *current_usable_area_override;
+	struct keybind *current_keybind;
+	struct mousebind *current_mousebind;
+	struct touch_config_entry *current_touch;
+	struct libinput_category *current_libinput_category;
+	const char *current_mouse_context;
+	struct action *current_keybind_action;
+	struct action *current_mousebind_action;
+	struct region *current_region;
+	struct window_switcher_field *current_field;
+	struct window_rule *current_window_rule;
+	struct action *current_window_rule_action;
+	struct view_query *current_view_query;
+	struct action *current_child_action;
+};
 
-static struct usable_area_override *current_usable_area_override;
-static struct keybind *current_keybind;
-static struct mousebind *current_mousebind;
-static struct touch_config_entry *current_touch;
-static struct libinput_category *current_libinput_category;
-static const char *current_mouse_context;
-static struct action *current_keybind_action;
-static struct action *current_mousebind_action;
-static struct region *current_region;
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, not merging right away to give @johanmalm a chance comment.

johanmalm:
I'm not happy with this one.

I see the merit of creating the state-struct, but this patch also includes changes to `entry()`, `traverse()` and `process_node()` which are not trivial to evaluate in one's head given the recursive nature of the parsing.

There is no indication of data validation. How much testing has been carried out?

It feels like we should write a unit-test to prove that we get the same nodenames before and after given some XML files with mixed elements, attributes and content.

We did have one a long time ago which may provide a useful starting point. https://github.com/labwc/labwc/blob/adf0f70a04fa8e9d24da4f5931618eddef95ec1b/tests/t1001-rcxml-nodenames-simple.c

`LABWC_DEBUG_CONFIG_NODENAMES` can of course also be used.

src/config/rcxml.c
```diff
@@ -1274,135 +1281,123 @@ entry(xmlNode *node, char *nodename, char *content)
...
-	content = (char *)node->content;
 	if (xmlIsBlankNode(node)) {
 		return;
 	}
-	name = nodename(node, buffer, sizeof(buffer));
-	entry(node, name, content);
-}
 
-static void xml_tree_walk(xmlNode *node);
-
```
Consolatis, line 1313: The only logical difference I see is that `xmlIsBlankNode()` used to guard only `entry()` and now guards `traverse()`. If a 'blank node' is ensured to not have properties nor children I think there is no logical difference.

Other than that, the `nodename` used to be built in the `process_node()` wrapper for `entry()` and is now built in `entry()` itself and the `process_node()` wrapper is removed.

I might have missed something though.

orfeasxyz: From a logical side there is no difference. I simply merged `process_node()` into `traverse()` because it didn't make much sense to exist separately IMO.

Also passing the `xmlNode` struct along with `nodename` and `content` to `entry()` seemed counterintuitive since the `xmlNode` already contains that information.

domo141:
I looked this (again) today, and to me it looks like Consolatis mentioned -- `xml_tree_walk()` 's to child nodes
are not called if `xmlIsBlankNode()` returns true is (only) logical change (and I don't see anyhing missed there). 
If one wanted to have logically equivalent, just `entry()` could be skipped when `xmlIsBlankNode()` returns true
(but I may have missed something there)

As there is just one file changed in this commit (`src/config/rcxml.c`), I did the following to minimise
changes in last view of my review:
```
git fetch upstream 02763d3945b31a3cf023ce7c9ac651b591ec5897
git -P show --raw 02763d3945b31a3cf023ce7c9ac651b591ec5897
git cat-file blob f7d92e4c > a
git cat-file blob 6649a724 > b
xxdiff a b
sed -i 's/state->//g' b
xxdiff a b
sed -i s/nodename/name/g a b
xxdiff a b
```

src/config/rcxml.c
Consolatis: Just wanted to sum it up for @johanmalm based on the comment above.

I'd prefer if we could get this merged or have a clear route of what to change to get it merged (e.g. keep the change to `entry()` `traverse()` and `process_node()  out of this PR?).

johanmalm: I’m looking for confidence that we’ve not functionally changed anything by refactoring those functions. 

If it were me authoring / reviewing I’d check that nodenames+values are the same before and after with a couple a config files (by setting the env var). I don’t suspect that would take more than 10 minutes - but would provide some verification that we’ve not overlooked something.

johanmalm:
It feels like it should be two  separate commits, but happy to ignore that.

src/config/rcxml.c
orfeasxyz: I went ahead and did the test you mentioned.

```bash
git checkout master # 
meson compile -C build/
LABWC_DEBUG_CONFIG_NODENAMES=1 ./build/labwc > pre_change.log
git checkout query_encapsulate_parser_state
meson compile -C build/
LABWC_DEBUG_CONFIG_NODENAMES=1 ./build/labwc > post_change.log
diff pre_change.log post_change.log # returned nothing
```

Got equivalent log output for both runs.

orfeasxyz: @johanmalm @Consolatis any update on this?

orfeasxyz:
I went ahead and reverted the `s/nodename/name/` in `entry()` as you suggested.

johanmalm:
LGTM

PR #2481, opened 2024-12-31
img: fix UAF on Reconfigure by refcounting

tokyo4j:
Before this commit, there was a use-after-free bug on Reconfigure:
- `theme_finish()` destroys lab_imgs for titlebar icons
- For some reason, `undecorate()` eventually calls `_create_buffer()` in `scaled-img-buffer.c`, which calls `img_render()` on a destroyed `lab_img`.

So this commit adds `wlr_buffer`-like APIs: `lab_img_lock()`, `lab_img_unlock()` and `lab_img_drop()`. This ensures a `lab_img` is never destroyed until no one references it.

### DIFF
--- a/include/img/img.h
+++ b/include/img/img.h
@@ -3,6 +3,7 @@
 #define LABWC_IMG_H
 
 #include <cairo.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <wayland-util.h>
 
@@ -19,6 +20,9 @@ struct lab_img {
 	struct theme *theme; /* Used by modifier functions */
 	struct wl_array modifiers; /* lab_img_modifier_func_t */
 	struct lab_img_cache *cache;
+
+	bool dropped;
+	int nr_locks;
 };
 
 struct lab_img *lab_img_load(enum lab_img_type type, const char *path,
@@ -69,10 +73,9 @@ void lab_img_add_modifier(struct lab_img *img, lab_img_modifier_func_t modifier,
 struct lab_data_buffer *lab_img_render(struct lab_img *img,
 	int width, int height, int padding, double scale);
 
-/**
- * lab_img_destroy() - destroy lab_img
- * @img: lab_img to destroy
- */
-void lab_img_destroy(struct lab_img *img);
+/* These functions closely follow the APIs of wlr_buffer */
+void lab_img_lock(struct lab_img *img);
+void lab_img_unlock(struct lab_img *img);
+void lab_img_drop(struct lab_img *img);
 
 #endif /* LABWC_IMG_H */
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -79,7 +79,6 @@ struct ssd {
 		} title;
 
 		char *app_id;
-		struct lab_img *icon_img;
 	} state;
 
 	/* An invisible area around the view which allows resizing */
--- a/src/common/scaled-img-buffer.c
+++ b/src/common/scaled-img-buffer.c
@@ -26,6 +26,7 @@ static void
 _destroy(struct scaled_scene_buffer *scaled_buffer)
 {
 	struct scaled_img_buffer *self = scaled_buffer->data;
+	lab_img_unlock(self->img);
 	free(self);
 }
 
@@ -57,6 +58,7 @@ scaled_img_buffer_create(struct wlr_scene_tree *parent, struct lab_img *img,
 	struct scaled_img_buffer *self = znew(*self);
 	self->scaled_buffer = scaled_buffer;
 	self->scene_buffer = scaled_buffer->scene_buffer;
+	lab_img_lock(img);
 	self->img = img;
 	self->width = width;
 	self->height = height;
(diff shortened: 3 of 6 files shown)


Consolatis:
> * For some reason, `undecorate()` eventually calls `_create_buffer()` in `scaled-img-buffer.c`, which calls `img_render()` on a destroyed `lab_img`.

Did you figure out why this happens? I have a feeling that we don't need the refcounting if we can figure out that one.

tokyo4j:
Looks like here's what's happening:

1. When 2 foot window is opened in the nested labwc session, a window completely overlaps another window with `cascade` placement policy.
2. Window decorations are refreshed in stacking order from front to back, so the foot window in front is temporarily undecorated, which exposes the decoration in the foot window below it.
3. As the scaled-img-buffer is now exposed, it receives `output_enter` event. The buffer is not reused here since the decoration in the front foot window has been destroyed.

Consolatis:
I tried to debug this a bit myself and I don't get it:
[Added a destroy listener](https://github.com/Consolatis/labwc/compare/labwc%3Alabwc%3Amaster...debug/buffer_crash) for the buffer which resets `self->img` to `NULL` on lab_img destroy. Then added a WLR_ERROR log in `_create()` to confirm that it correctly detects the issue. But that log message never triggers and I fail to reproduce the issue now.

I am very confused what is going on here.

tokyo4j:
> [Added a destroy listener](https://github.com/Consolatis/labwc/compare/labwc%3Alabwc%3Amaster...debug/buffer_crash) for the buffer which resets self->img to NULL on lab_img destroy. Then added a WLR_ERROR log in _create() to confirm that it correctly detects the issue. But that log message never triggers and I fail to reproduce the issue now.

I guess that's because you set `self->img` to `NULL` without flushing caches.

In your patch, when scaled_img_buffers for "inactive" window buttons receive `output_enter` event in the situation I described above, the cached buffers for "active" window buttons are incorrectly reused as the two scaled_img_buffers for different window states are considered _equal_ after you set `self->img` to `NULL` for both.

So calling `scaled_scene_buffer_request_update()` after setting `self->img` to `NULL` may fix it.

However, I prefer `wlr_buffer`-like approach in this PR, because it's simpler and removes `ssd->state.icon_img`.

Consolatis:
Hmm, I see.

So during a reconfigure we currently destroy all theme lab_img and then do `undecorate()` and `decorate()`. So that leaves 3 potential options (+ any that I missed):
- change to undecorate() -> theme_finish() -> theme_init() -> decorate()
- make lab_img outlive theme_finish() (e.g. what this PR is doing)
- destroy or otherwise disable scaled-img-buffer when the backing lab_img is destroyed (my branch + `instance->img == NULL` always returning `false` in `_equal()`

The first option feels a bit finicky because changing the order of reconfigure events in the future seems possible so if we go that route we need to at least add comments about the dependency in there. Other than that it seems like a route we could go.

The second option slightly increases the complexity of the lab_img API, manually doing refcounting is easy to get wrong at random points in time (for example during unrelated refactorings) and could then easily lead to mem leaks.

The third option somewhat leaks the lifetime of a lab_img into scaled-img-buffer but only rendering inoperable rather than destroying it which also feels a bit weird.

Did I miss any option or argument?

tokyo4j:
Your assumption is correct. As an implementer I liked the `lab_img_{lock,unlock,drop}` approach because it's _syntactically_ simpler, but I understand your concern.

The other 2 approaches you suggested seems to be valid, but I'd like to propose another approach:

- Cloning the image `lab_img_copy()` when a new scaled-img-buffer is created (or updated)
- Destroying the image with `lab_img_destroy()` when the scaled-img-buffer is destroyed (or updated)

Note that `lab_img_copy()` doesn't copy the actual content of the image; it just increments the refcount in `lab_img_cache` which holds the image content. The primary purpose of `lab_img_copy()` was to share the loaded image buffer between different variants of titlebar buttons (e.g. normal/hovered/rounded buttons), but I think we can reuse this to safely manage the lifetime of `lab_img`, in a way similar to what this PR does, without complicating the `lab_img` API.

I'll open another PR for it later.

tokyo4j:
Your assumption is correct. As an implementer I liked the lab_img_{lock,unlock,drop} approach because it's syntactically simpler, but I understand your concern.

The other 2 approaches you suggested seems to be valid, but I'd like to propose another approach:

    Cloning the image lab_img_copy() when a new scaled-img-buffer is created (or updated)
    Destroying the image with lab_img_destroy() when the scaled-img-buffer is destroyed (or updated)

Note that lab_img_copy() doesn't copy the actual content of the image; it just increments the refcount in lab_img_cache which holds the image content. The primary purpose of lab_img_copy() was to share the loaded image buffer between different variants of titlebar buttons (e.g. normal/hovered/rounded buttons), but I think we can reuse this to safely manage the lifetime of lab_img, in a way similar to what this PR does, without complicating the lab_img API.

I'll open another PR for it later.

Consolatis:
My personal preference before looking into your follow up PR was option 1. After reading through your follow-up PR changes I agree and prefer options 2 as implemented with #2489. It simply makes use of the existing `labwc_img` implementation which already implements automatic refcounting anyway. I really like that approach.

PR #2458, opened 2024-12-28
rcxml: add core.autoEnableOutputs option

jlindgren90:
rcxml: add core.autoEnableOutputs option

Currently, labwc automatically enables outputs at startup and when new
outputs are connected. Make this behavior optional (but still enabled by
default).

With autoEnableOutputs disabled, tools such as kanshi can be used to
give finer-grained control of which outputs are enabled and when.
Personally, I use this to avoid re-enabling outputs that disconnect and
reconnect during powersave, but I can imagine other use-cases for it
also.

The first two commits address modesetting failures that are revealed by
this option but could also be merged separately.

PR #2486, opened 2025-01-01
output: ignore wlr_output_state.mode/custom_mode except for client request

jlindgren90:
The wlr_output_state passed to the new_output signal sometimes has a non-preferred mode set, which we were ignoring before 03be489eb973. Ignore it again to prevent the wrong mode from being set.

The first commit adds some debugging output which might be useful again in future, so I left it.

See discussion on #2458.


johanmalm:
Thanks :smile:

PR #2488, opened 2025-01-02
Use 'labwc' directory for themes as well as 'openbox-3'

johanmalm:
Rationale in commit message.

Have set to draft because it could do with some more testing + discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ For a step-by-step initial configuration guide, see [getting-started].
 
 ## 4. Theming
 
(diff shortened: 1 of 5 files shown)


Consolatis:
Didn't yet look at the changes but the idea is fine with me. Might need updates in labwc-tweaks and friends.

JFYI @stefonarch

tokyo4j:
LGTM. Have tested and it worked well.

johanmalm:
Thanks. Okay, let's go for it then.

PR #2489, opened 2025-01-03
img: fix UAF on Reconfigure by refcounting (attempt 2)

tokyo4j:
This PR implements the approach I proposed in https://github.com/labwc/labwc/pull/2481#issuecomment-2569384105.

Like #2481, this PR makes `lab_img` outlive `theme_finish()` by refcounting, but without adding new APIs:
- When creating a `scaled_img_buffer`, `lab_img` is replicated via `lab_img_copy()`
- When destroying a `scaled_img_buffer`, `lab_img` is destroyed via `lab_img_destroy()`

Let me explain how this works.

## The lifetime of `lab_img` and `lab_img_cache`

After startup, `theme` struct references `lab_img`, which references `lab_img_cache` with `refcount=1`. Here, `lab_img_cache` is the wrapper for `lab_data_buffer` or `RsvgHandle` and represents the actual content of the loaded image file.

first

When a window is opened and its decoration is created, the `lab_img` is copied via `lab_img_copy()` and a new `scaled_img_buffer` references it. Note that the image content is not copied here; instead, the refcount of `lab_img_cache` is incremented.

second

When the theme is de-initialized in `theme_finish()` and the `lab_img` referenced by `theme` is destroyed, the `lab_img` referenced by the `scaled_img_buffer` and `lab_img_cache` outlive. Therefore, it's safe for `_update_buffer()` in `scaled-img-buffer.c` to be called.

third

And when the decoration is destroyed via `undecorate()`, the `scaled_img_buffer`, `lab_img` and `lab_img_cache` are finally destroyed.

## Motivation of `lab_img_copy()` and `lab_img_cache`

For better understanding of `lab_img` API in general, let me explain the initial motivation of `lab_img_copy()` and `lab_img_cache` in #2444. `lab_img_copy()` was introduced to share the image content with different variants of buttons including normal, hovered, rounded, rounded-hovered buttons:

foo

For example, when `close_hover-active.png` is not found, the `lab_img` for `close-active.png` is copied via `lab_img_copy()` and a "modifier" function that draws a hover overlay on it is added to the copied `lab_img` via `lab_img_add_modifier()`. And if the close button is placed at the corner of the titlebar, the `lab_img` is further copied and the modifier function that cuts its corner is applied.

### DIFF
--- a/src/common/scaled-img-buffer.c
+++ b/src/common/scaled-img-buffer.c
@@ -26,6 +26,7 @@ static void
 _destroy(struct scaled_scene_buffer *scaled_buffer)
 {
 	struct scaled_img_buffer *self = scaled_buffer->data;
+	lab_img_destroy(self->img);
 	free(self);
 }
 
@@ -36,7 +37,7 @@ _equal(struct scaled_scene_buffer *scaled_buffer_a,
 	struct scaled_img_buffer *a = scaled_buffer_a->data;
 	struct scaled_img_buffer *b = scaled_buffer_b->data;
 
-	return a->img == b->img
+	return lab_img_equal(a->img, b->img)
 		&& a->width == b->width
 		&& a->height == b->height
 		&& a->padding == b->padding;
@@ -52,12 +53,13 @@ struct scaled_img_buffer *
 scaled_img_buffer_create(struct wlr_scene_tree *parent, struct lab_img *img,
 	int width, int height, int padding)
 {
+	assert(img);
 	struct scaled_scene_buffer *scaled_buffer = scaled_scene_buffer_create(
 		parent, &impl, &cached_buffers, /* drop_buffer */ true);
 	struct scaled_img_buffer *self = znew(*self);
 	self->scaled_buffer = scaled_buffer;
 	self->scene_buffer = scaled_buffer->scene_buffer;
-	self->img = img;
+	self->img = lab_img_copy(img);
 	self->width = width;
 	self->height = height;
 	self->padding = padding;
@@ -73,7 +75,9 @@ void
 scaled_img_buffer_update(struct scaled_img_buffer *self, struct lab_img *img,
 	int width, int height, int padding)
 {
-	self->img = img;
+	assert(img);
+	lab_img_destroy(self->img);
+	self->img = lab_img_copy(img);
 	self->width = width;
 	self->height = height;
 	self->padding = padding;
--- a/src/img/img.c
+++ b/src/img/img.c
@@ -106,10 +106,8 @@ lab_img_copy(struct lab_img *img)
 }
 
 void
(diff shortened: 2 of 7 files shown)


tokyo4j:
Note: The first commit is not directly related to the topic. It just removes `lab_img.theme`.

src/common/scaled-img-buffer.c
```diff
@@ -57,7 +58,7 @@ scaled_img_buffer_create(struct wlr_scene_tree *parent, struct lab_img *img,
 	struct scaled_img_buffer *self = znew(*self);
 	self->scaled_buffer = scaled_buffer;
 	self->scene_buffer = scaled_buffer->scene_buffer;
-	self->img = img;
+	self->img = lab_img_copy(img);
```
Consolatis, line 61: Can we add an `assert(img)` here and in `scaled_img_buffer_update()`? That should ensure future accidental misuse of the API.

src/img/img.c
```diff
@@ -218,3 +218,18 @@ lab_img_destroy(struct lab_img *img)
...
+	if (img_a->cache != img_b->cache
+			|| img_a->modifiers.size != img_b->modifiers.size) {
+		return false;
+	}
+	if (img_a->modifiers.size == 0) {
+		return true;
+	} else {
+		return !memcmp(img_a->modifiers.data, img_b->modifiers.data,
+			img_a->modifiers.size);
+	}
```
Consolatis, line 234: The else branch isn't necessary here

src/img/img.c
```diff
@@ -220,3 +218,18 @@ lab_img_destroy(struct lab_img *img)
 	wl_array_release(&img->modifiers);
 	free(img);
 }
+
+bool
+lab_img_equal(struct lab_img *img_a, struct lab_img *img_b)
+{
+	if (img_a->cache != img_b->cache
```
Consolatis, line 225: I think this should either `assert()` both lab_img args or handle one or both of them being `NULL`.

Consolatis (commented):
So this basically depends on the underlying img cache buffers already being refcounted.

I like this much better, nice work!
Three small nitpicks below. With those addressed this LGTM.

src/common/scaled-img-buffer.c
tokyo4j: Agreed. Added `assert()`.

src/img/img.c
tokyo4j: I removed the if-else statement and replaced it with `img_a->modifiers.size == 0 || !memcmp(img_a->modifiers.data, img_b->modifiers.data, img_a->modifiers.size);`

src/img/img.c
tokyo4j: Handled `NULL`.

tokyo4j:
@Consolatis I addressed your comment. Please merge this PR if you're fine with it.

Consolatis:
Thanks!

PR #2492, opened 2025-01-04
img: apply padding around window icon only horizontally

### LINKED ISSUE #2450: issues with application icon scaling after #2444

tokyo4j:
Closes #2450.

16dbdc64 changed the padding around the app icon in the titlebar to be applied both vertically and horizontally rather than only horizontally because it was more natural from a developer's perspective, but some users complained about the smaller icons in certain configurations.

So let's undo the change in 16dbdc64 and apply the icon padding only horizontally for now.

We can add configurations for the icon padding (or icon size independent from `window.button.{width,height}`?) later.

### DIFF
--- a/include/img/img.h
+++ b/include/img/img.h
@@ -63,11 +63,11 @@ void lab_img_add_modifier(struct lab_img *img, lab_img_modifier_func_t modifier,
  * @img: source image
  * @width: width of the created buffer
  * @height: height of the created buffer
- * @padding: padding around the rendered image in the buffer
+ * @padding_x: horizontal padding around the rendered image in the buffer
  * @scale: scale of the created buffer
  */
 struct lab_data_buffer *lab_img_render(struct lab_img *img,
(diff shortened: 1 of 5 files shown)


johanmalm:
LGTM

PR #2495, opened 2025-01-05
img: don't abort with non-ARGB32 PNG files

tokyo4j:
Cairo surfaces loaded with `cairo_image_surface_create_from_png()` are not always ARGB32 format. So this PR converts non-ARGB32 surfaces to ARGB32 to not abort in `buffer_adopt_cairo_surface()`.

### DIFF
--- a/src/img/img-png.c
+++ b/src/img/img-png.c
@@ -59,5 +59,22 @@ img_png_load(const char *filename)
 		return NULL;
 	}
 
-	return buffer_adopt_cairo_surface(image);
+	if (cairo_image_surface_get_format(image) == CAIRO_FORMAT_ARGB32) {
+		return buffer_adopt_cairo_surface(image);
+	} else {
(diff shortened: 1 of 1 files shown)


Consolatis:
Technically other pixel formats should be fine as well, we'd just need to translate the cairo format into the drm format and ensure that that specific format is supported by the renderer rather than hardcoding `DRM_FORMAT_ARGB8888`. Also fine with this variant though.

tokyo4j:
> Technically other pixel formats should be fine as well, we'd just need to translate the cairo format into the drm format and ensure that that specific format is supported by the renderer rather than hardcoding DRM_FORMAT_ARGB8888.

Yeah it makes sense. I added a `TODO` comment for it.

For now, I think copying the surface to a buffer is fine since this PR handles such an edge-case of non-ARGB32 PNG files.

PR #2498, opened 2025-01-06
IME: fix stuck Ctrl when pressed Ctrl+F in Firefox with Fcitx5

tokyo4j:
This PR fixes a problem after 683f67b that Ctrl modifier state is stuck when pressed Ctrl+F (shortcut for word-search) in Firefox while Fcitx5 is running, which makes mouse scrolls zoom in/out the UI even after Ctrl key is released.

This doesn't affect the code path without input method running. Below is the detailed explanation:

<details><summary>details</summary>

When Ctrl+F is pressed, an input box is opened in the application and Fcitx5 creates a new virtual keyboard
(VK), whose initial modifiers is empty. Then prior to 683f67b, the key/modifiers events flowed like this:

- The compositor detects F key-release
  - Modifiers (Ctrl pressed) are notified via `wlr_input_method_keyboard_grab_v2_set_keyboard()`
  - F key-release is forwarded to IM
- IM sends modifiers (Ctrl) back to the compositor via VK
  - **The modifiers on VK is updated (empty->Ctrl)**
  - **Modifers (Ctrl) are notified to the app**
- IM sends F key-release back to the compositor via VK
  - F key-release is notified to the app
- The compositor detects Ctrl key-release
  - Ctrl key-release is forwarded to IM
  - Modifiers (empty) are forwarded to IM
- IM sends Ctrl key-release back to the compsitor via VK
  - Ctrl key-release is notified to IM
- IM sends modifiers (empty) back to the compositor via VK
  - **The modifiers on VK is updated again (Ctrl->empty)**
  - **Modifiers (empty) are notified to the app**

Thus, the final modifiers (empty) is notified to the application a expected. However, after 683f67b, the key/modifiers events flowed like this:

- The compositor detects F key-release
  - F key-release is directly notified to the app
  - The modifiers (Ctrl) is also notified to the app
- The compositor detects Ctrl key-release
  - Ctrl key-release is directly notified to the app
  - Modifiers (empty) are forwarded to IM
- IM sends modifiers (empty) back to the compositor via VK
  - **Modifier on VK is not updated (empty->empty)**
  - **The compositor ignores it**

So the final modifier (empty) is never notified to the application, which causes stuck Ctrl modifier.

This PR fixes this by not forwarding the modifiers when it hasn't been updated since it was forwarded previously. So after this commit, the key/modifiers events flow like this:

- The compositor detects F key-release
  - F key-release is directly notified to the app
  - The modifiers (Ctrl) is also notified to the app
- The compositor detects Ctrl key-release
  - Ctrl key-release is directly notified to the app
  - The modifiers are directly notified to the app because the modifiers (empty) are the same as the last forwarded modifier (empty).

</details>

I've been testing this PR for a day and it's been working well so far.

### DIFF
--- a/include/input/ime.h
+++ b/include/input/ime.h
@@ -21,7 +21,7 @@ struct input_method_relay {
 	struct wl_list text_inputs; /* struct text_input.link */
 	struct wlr_input_method_v2 *input_method;
 	struct wlr_surface *focused_surface;
-	struct lab_set pressed_keys;
+	struct lab_set forwarded_pressed_keys;
 	/*
 	 * Text-input which is enabled by the client and communicating with
 	 * input-method.
--- a/include/input/ime.h
+++ b/include/input/ime.h
@@ -21,7 +21,10 @@ struct input_method_relay {
 	struct wl_list text_inputs; /* struct text_input.link */
 	struct wlr_input_method_v2 *input_method;
 	struct wlr_surface *focused_surface;
+
 	struct lab_set forwarded_pressed_keys;
+	struct wlr_keyboard_modifiers forwarded_modifiers;
+
 	/*
 	 * Text-input which is enabled by the client and communicating with
 	 * input-method.
--- a/src/input/ime.c
+++ b/src/input/ime.c
@@ -58,11 +58,25 @@ input_method_keyboard_grab_forward_modifiers(struct keyboard *keyboard)
 {
 	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
 		get_keyboard_grab(keyboard);
+
+	struct wlr_keyboard_modifiers *forwarded_modifiers =
+		&keyboard->base.seat->input_method_relay->forwarded_modifiers;
+	struct wlr_keyboard_modifiers *modifiers =
+		&keyboard->wlr_keyboard->modifiers;
+
+	if (forwarded_modifiers->depressed == modifiers->depressed
+			&& forwarded_modifiers->latched == modifiers->latched
+			&& forwarded_modifiers->locked == modifiers->locked
+			&& forwarded_modifiers->group == modifiers->group) {
+		return false;
+	}
+
 	if (keyboard_grab) {
+		*forwarded_modifiers = keyboard->wlr_keyboard->modifiers;
 		wlr_input_method_keyboard_grab_v2_set_keyboard(keyboard_grab,
 			keyboard->wlr_keyboard);
 		wlr_input_method_keyboard_grab_v2_send_modifiers(keyboard_grab,
-			&keyboard->wlr_keyboard->modifiers);
+			modifiers);
 		return true;
 	} else {
 		return false;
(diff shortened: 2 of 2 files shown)


johanmalm:
I'm happy with this except that I don't think we should use `memcmp()` to check if structs are equal because of potential random (compiler generated) padding between fields in structs. I believe that a member-by-member comparison is the correct, safe way.

tokyo4j:
I updated to compare the fields one-by-one. I'll merge this PR once you gave LGTM.

Just my excuse: honestly, I really don't like making this kind of PR because this is very hard to explain, specific to Fcitx5 and more like a workaround. But I decided to make this PR because this is not a problem of Fcitx5, but a problem of the immature IME support in wayland-protocol that lacks proper synchronization of keyboard state between input method and compositor ([reference](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/39#note_2371724)). I really hope the situation gets better.

johanmalm:
LGTM.

tokyo4j:
My note after some analysis: the root cause of this kind of de-synchronized modifier state between the compositor and Fcitx5 is that we are not correctly advertising the modifier state to Fcitx5 when it grabs the keyboard. This is because, when Fcitx5 is activated, Fcix5 firstly creates a virtual keyboard which updates `wlr_seat.keyboard_state.keyboard` in `new_keyboard()` and then grabs the keyboard, so the branch below is never taken:

https://github.com/labwc/labwc/blob/f4d581ef91366dcd83eeebec2beb13ee90d2de32/src/input/ime.c#L355-L360

As a result, Fcitx5 cannot receive the modifier state when it's activated and it assumes no modifiers are depressed. To fix this, I think we can call `wlr_seat_set_keyboard()` in `new_keyboard()` only when no keyboard is set on `wlr_seat`, after which we can just revert this PR.

But one thing to consider with this approach is that, when the user types something with Fcitx5 enabled and switch the keyboard focus to another window, Fcitx5 destroys the virtual keyboard even though `wlr_seat.keyboard_state.keyboard` is already set to the virtual keyboard. This means `wlr_seat.keyboard_state.keyboard` becomes `NULL` and the modifier state is never sent to Fcitx5 when Fcitx5 is activated again anyway. So to mitigate this issue, we will need to revert #1816 to ensure `wlr_seat.keyboard_state.keyboard` is never `NULL` as long as there's any physical keyboards connected. This re-introduces #1796, but after some analysis, I concluded this is a fundamental issue with X11 that clients don't have a way to obtain the keymap exactly at the time when the key press/release events are received. At least for `wtype`, the issue can be mitigated by delaying destroying the virtual keyboard like `wtype -k xyz -s 200`.

fenevadkan:
I have an issue on zen-browser (which is basically firefox), where I have enabled ctrl-tab (recent order tab switching) and sometimes when i press it, the wide tab selector bar on screen will be stuck, which is quite frustrating. So it seems to be related.

PR #2499, opened 2025-01-06
cursor: fix invisible cursor on application after reconfigure

tokyo4j:
On reconfigure, we should send `wl_pointer.{leave,enter}` events if the cursor is on an application surface to let the application update the cursor, but #2455 prevented these events from being sent.


Consolatis:
LGTM, not tested.

johanmalm:
Thanks.

PR #2501, opened 2025-01-06
osd: prevent preview outlines to be above OSD

### LINKED ISSUE #2465: Window switcher preview boarder/outline is in front of the OSD
### Problem description

See this video, the border/outline of the preview window when cycling in the window switcher is in front of the OSD. It should be, like the previewed window content, behind of the OSD.

https://github.com/user-attachments/assets/3fb8a158-b79d-4ba3-9aed-a2eef7a65edb

Note that the last merged PR (https://github.com/labwc/labwc/pull/2464) is not the cause of this issue, this got sneaked in earlier. If the source of this issue isn't that obvious, I can bisect of course.

PS: Nice issue template ;)

### Steps to reproduce

1. Run labwc with window switcher preview and outlines enabled

```xml
  <windowSwitcher show="yes" preview="yes" outlines="yes">
    <fields>
      <field content="desktop_entry_name" width="25%" />
      <field content="title" width="75%" />
    </fields>
  </windowSwitcher>
``` 

2. Cycle though the OSD with an application where the border crossed the OSD

### labwc build source

Local build

### labwc version

https://github.com/labwc/labwc/commit/a98b535

### labwc environment

### Distribution

Fedora 41

tokyo4j:
Fixes #2465 by placing `osd_tree` above `menu_tree`.

This also prevents layer-shell surfaces from covering the OSD. I'm not sure whether to be higher, I prefer server-side component to be higher.

Before (higher-first):
- preview outlines (just above menu)
- menu
- layer-shell (top/overlay/popup)
- osd

After (higher-first):
- osd
- preview outlines (just above menu)
- menu
- layer-shell (top/overlay/popup)

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -511,6 +511,7 @@ new_output_notify(struct wl_listener *listener, void *data)
 	 * Set the z-positions to achieve the following order (from top to
 	 * bottom):
 	 *	- session lock layer
+	 *	- window switcher osd
 	 *	- compositor menu
 	 *	- layer-shell popups
 	 *	- overlay layer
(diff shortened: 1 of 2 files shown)


Consolatis:
LGTM as a stopgap measure until we restructure the scene tree for the OSD (+ preview + outlines) in general.

johanmalm:
Thanks :smile:

PR #2502, opened 2025-01-07
theme: don't try to load "themerc"s if theme_name is NULL

domo141:
theme_name is NULL when <theme><name></name>...</theme> is in rc.xml

--8<--

I'd  guess trying to load e.g. `~/.local/share/themes/(null)/labwc/themerc`
is not a feature.

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -1496,15 +1496,18 @@ theme_init(struct theme *theme, struct server *server, const char *theme_name)
 	 */
 	theme_builtin(theme, server);
 
-	/*
-	 * Read
-	 *   - <data-dir>/share/themes/$theme_name/labwc/themerc
-	 *   - <data-dir>/share/themes/$theme_name/openbox-3/themerc
-	 */
 	struct wl_list paths;
-	paths_theme_create(&paths, theme_name, "themerc");
-	theme_read(theme, &paths);
-	paths_destroy(&paths);
+
+	if (theme_name) {
+		/*
+		 * Read
+		 *   - <data-dir>/share/themes/$theme_name/labwc/themerc
+		 *   - <data-dir>/share/themes/$theme_name/openbox-3/themerc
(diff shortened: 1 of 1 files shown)


src/theme.c
```diff
@@ -1496,15 +1496,18 @@ theme_init(struct theme *theme, struct server *server, const char *theme_name)
...
-	 * Read
-	 *   - <data-dir>/share/themes/$theme_name/labwc/themerc
-	 *   - <data-dir>/share/themes/$theme_name/openbox-3/themerc
-	 */
 	struct wl_list paths;
-	paths_theme_create(&paths, theme_name, "themerc");
-	theme_read(theme, &paths);
-	paths_destroy(&paths);
+
+	if (theme_name) {
```
tokyo4j, line 1501: I suggest `string_null_or_empty()` for robustness (though I guess an empty strong is never set in `rcxml.c`).

tokyo4j:
Nice find! I tried creating `~/.themes/(null)` directory and indeed it was scanned.

LGTM other than my minor suggestion.

src/theme.c
domo141: That is a good suggestion; if someone(tm) were able to sneak in empty string then the scan would go
one subdirectory shallower which is also ugly... I'll update.

tokyo4j: OK. Then we'll need to add some utility function that checks if a string can be treated as a single directory and use it for both labwc theme and icon theme. I'll merge this PR for now.

domo141:
tnx. wrapper function which checks and trims content before calling xstrdup_replace() could do the trick,
in these lines:
```
src/config/rcxml.c:1110:                xstrdup_replace(rc.theme_name, content);
src/config/rcxml.c:1112:                xstrdup_replace(rc.icon_theme_name, content);
```

PR #2503, opened 2025-01-08
Minor prep work for menu icons

Consolatis:


Consolatis:
Meh, static analyzer hits again:
> ../src/desktop-entry.c:323:22: error: dereference of NULL ‘icon_name’ [CWE-476] [-Werror=analyzer-null-dereference]
>  323 |         if (icon_name[0] == '/') {

I am pretty sure this can't be reached because we return early on `string_null_or_empty()`. ideas? Should I just throw a `|| !icon_name` into the condition even though it already checks for it internally?

tokyo4j:
I personally prefer defining `string_null_or_empty()` as a macro rather than a function since I don't like adding a dead code path to make the static analyzer happy. But I don't insist on it as it loses the type information for arguments.

Also, I suggest following function names as I think they clarify what they do better:
- `desktop_entry_load_icon()`
- `desktop_entry_load_icon_from_app_id()`

src/desktop-entry.c
```diff
@@ -326,28 +321,48 @@ desktop_entry_icon_lookup(struct server *server, const char *app_id, int size,
...
+	}
 
+	const char *icon_name = NULL;
+	struct sfdo_desktop_entry *entry = get_desktop_entry(sfdo, app_id);
+	if (entry) {
+		icon_name = sfdo_desktop_entry_get_icon(entry, NULL);
+	}
+
+	struct lab_img *img = NULL;
+	img = desktop_entry_icon_lookup(server, icon_name, size, scale);
```
tokyo4j, line 361: nit: I think the first assignment to `NULL` is doing nothing.

Consolatis: changed to `struct lab_img *img = desktop_entry_load_icon()`

Consolatis:
Changed function names as suggested, not sure about the macro variant of `string_null_or_empty()`. I think loosing the type information is worse than adding a useless `NULL` check.

PR #2508, opened 2025-01-10
scaled-scene-buffer: use outputs_update signal

Consolatis:
This fixes an issue with buffers not updating when an output is configured for a new scale. It also supports windows being on more than 2 outputs at once and in general simplifies the code.


johanmalm:
LGTM. Not tested

PR #2509, opened 2025-01-10
menu icon support

Consolatis:
20250129_12h00m18s_grim

Can be tested with a toplevel "applications" menu (or some usual pipe menu)
```xml
<menu id="applications" label="foo" execute="labwc-menu-generator --icons --pipemenu" />
```
or a usual static menu item
```xml
<menu id="root-menu">
	<item label="Terminal" icon="foot">
	</item>
	[..]
</menu>
```

Open questions:
- [x] do we want this feature in the first place
- [x] how can we provide a toggle to deactivate menu icons
- [x] how do we allow a theme to define the width of menu icons, should a width of 0 be used to disable menu icons?

### DIFF
--- a/docs/labwc-menu.5.scd
+++ b/docs/labwc-menu.5.scd
@@ -2,11 +2,11 @@ labwc-menu(5)
 
 # NAME
 
-labwc - menu files
+labwc - menu file
 
 # DESCRIPTION
 
-Static menus are built based on content of XML files located at
+Static menus are built based on the menu.xml file located at
 "~/.config/labwc" and equivalent XDG Base Directories.
 
 # SYNTAX
--- a/docs/labwc-menu.5.scd
+++ b/docs/labwc-menu.5.scd
@@ -15,29 +15,37 @@ The menu file must be entirely enclosed within <openbox_menu> and
 </openbox_menu> tags.  Inside these tags, menus are specified as follows:
 
 ```
+<!-- A toplevel menu -->
+<menu id="external_menu_settings" label="External Settings" icon="preferences-other">
+  <item label="Display setting">
+    <action name="Execute" command="wdisplays" />
+  </item>
+</menu>
+
+<!-- Another toplevel menu -->
 <menu id="">
 
   <!-- A menu entry with an action, for example to execute an application -->
-  <item label="">
+  <item label="" icon="">
     <action></action>
   </item>
 
-  <!-- A submenu defined elsewhere -->
-  <menu id="" />
+  <!-- A submenu defined elsewhere, uses external label and icon attributes -->
+  <menu id="external_menu_settings" />
 
   <!-- Horizontal line -->
   <separator />
 
   <!-- An inline submenu -->
-  <menu id="" label="">
+  <menu id="" label="" icon="">
     ...some content...
   </menu>
 
   <!-- Title -->
-  <separator label=""/>
+  <separator label="" />
 
   <!-- Pipemenu -->
-  <menu id="" label="" execute="COMMAND"/>
+  <menu id="" label="" icon="" execute="COMMAND" />
 
 </menu>
 ```
@@ -56,14 +64,27 @@ The menu file must be entirely enclosed within <openbox_menu> and
 		  view to that workspace when selected.
 
 *menu.id* (when nested under other *<menu>* element)
-	Link to a submenu defined elsewhere (by a *<menu id="">* at toplevel)
+	Link to a submenu defined elsewhere (by a *<menu id="">* at toplevel).
 
 *menu.label*
 	The title of the menu, shown in its parent. A label must be given when
 	defining a menu.
 
+*menu.icon*
+	An icon to be rendered, shown in its parent. This argument is optional.
+	See *menu.item.icon* for further details.
+
 *menu.item.label*
 	The visible name of the menu item.
+
+*menu.item.icon*
+	The name of an icon to be rendered in front of the menu entry. This
+	attribute is optional. The name follows naming conventions of "Icon="
+	entries in .desktop files. If used, it is recommended to use the name of
+	the icon only rather than a full path. This ensures that the icon will
+	be looked up in the scale of the output where the menu is rendered. E.g.
+	use of icon="vlc" is suggested over using
+	icon="/usr/share/icons/hicolor/16x16/apps/vlc.png".
 
 *menu.item.action*
 	See labwc-actions(5). Note: XML CDATA is supported for this node in
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -459,8 +459,7 @@ fill_item(char *nodename, char *content)
 			"nodename: '%s' content: '%s'", nodename, content);
 	} else if (!strcmp(nodename, "icon")) {
 #if HAVE_LIBSFDO
-		// TODO: add some rc.menu_icons bool
-		if (true && !string_null_or_empty(content)) {
+		if (rc.menu_show_icons && !string_null_or_empty(content)) {
 			xstrdup_replace(current_item->icon_name, content);
 			current_menu->has_icons = true;
 		}
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -20,6 +20,7 @@
 #include "common/mem.h"
 #include "common/nodename.h"
 #include "common/scaled-font-buffer.h"
+#include "common/scaled-icon-buffer.h"
 #include "common/scaled-rect-buffer.h"
 #include "common/scene-helpers.h"
 #include "common/spawn.h"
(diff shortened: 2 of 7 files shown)


johanmalm:
I like this. Feels right to include.

Openbox supports this with menu.shoeIcon in rc.xml

https://openbox.org/help/Menus#Value_of_attribute_.22icon.22

Not sure whats best regarding size. Needs some thinking.

If all items in a menu-window don't have in icon, should we adjust the menu widtg? This might be relevant for pipemenus

Noy yet tested.

Consolatis:
> If all items in a menu-window don't have in icon, should we adjust the menu width? This might be relevant for pipemenus

Yes, this PR handles that. If a menu contains at least one icon if has an additional padding of padding_x + icon_size. If the menu has no icons at all it works as before, e.g. no additional padding.

tokyo4j:
Can we just set the icon size to  (menu item height) - 2 * `menu.items.padding.y`? I guess that's how Openbox basically works: https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/openbox/menuframe.c#L541-L546

Consolatis:
Changed to `scaled_icon_buffer` and derived icon size from `item_height - 2 * padding_y` (which itself is derived from the font height).

Consolatis:
> Openbox supports this with menu.showIcon in rc.xml
> 
> https://openbox.org/help/Menus#Value_of_attribute_.22icon.22

Added:
```xml
<menu>
    <showIcons>yes</showIcons>
</menu>
```
or in the more compact form of
```xml
<menu showIcons="yes" />
```

> Not sure whats best regarding size. Needs some thinking.

I've gone with tokyo4j's suggestion of calculating the icon size based on the menu item height.

Things missing:
- [ ] support for the `icon` attribute for submenus (inline and external), assuming that we want to support it
- [ ] squashing of the commits

docs/labwc-menu.5.scd
```diff
@@ -65,6 +65,14 @@ The menu file must be entirely enclosed within <openbox_menu> and
 *menu.item.label*
 	The visible name of the menu item.
 
+*menu.item.icon*
+	An icon to be rendered in front of the menu entry. This attribute is
+	optional. If used, it is recommended to use the name of the icon only
+	rather than a full path. This ensures that the icon will be looked up
+	in the scale of the output where the menu is rendered. E.g. use of
+	icon="vlc" is suggested over using
+	icon="/usr/share/icons/hicolor/16x16/apps/vlc.png".
```
tokyo4j, line 74: I'm not sure if this recommendation is necessary here, because many apps are shipped with SVG icons and scales don't matter for them. We can just say that the value can be either of full path or icon name, like `Icon=` entries in desktop files.

src/menu/menu.c
```diff
@@ -363,6 +379,12 @@ menu_update_scene(struct menu *menu)
 			+ 2 * theme->menu_border_width;
 		menu->size.width = MAX(menu->size.width, width);
 	}
+
+	if (menu->has_icons) {
+		menu->size.width += theme->menu_items_padding_x
+			+ theme->menu_item_height
+			- 2 * theme->menu_items_padding_y;
```
tokyo4j, line 386: I think we should make it clearer that the icon size is `theme->menu_item_height - 2 * theme->menu_items_padding_y;`. How about defining a macro like `#define ICON_SIZE(theme) ((theme)->menu_item_height - 2 * (theme)->menu_items_padding_y)`?

src/menu/menu.c
```diff
@@ -436,10 +458,13 @@ fill_item(char *nodename, char *content)
 		wlr_log(WLR_ERROR, "expect <item label=\"\"> element first. "
 			"nodename: '%s' content: '%s'", nodename, content);
 	} else if (!strcmp(nodename, "icon")) {
-		/*
-		 * Do nothing as we don't support menu icons - just avoid
-		 * logging errors if a menu.xml file contains icon="" entries.
-		 */
+#if HAVE_LIBSFDO
+		if (rc.menu_show_icons && !string_null_or_empty(content)
+				&& strcmp(content, "(null)") != 0) {
```
tokyo4j, line 463: Can `content` take `(null)`?

tokyo4j:
Overall it worked great! I left some bikeshedding comments, but I think we can merge this PR once submenu icons are supported.

I'm a bit afraid that we are not handling the case where menus are so small that icons don't fit in it, but looks like the issue has been already present even without this PR. Here's an example configuration:
```
menu.items.padding.x: 50
menu.width.max: 50
menu.width.min: 50
```

src/menu/menu.c
Consolatis: Yes, its just a string. I saw this behavior in `openbox-menu`. I assume they are using snprintf without checking for `NULL`. And the chances of a legitimate icon using the name `(null)` seems pretty minimal.

src/menu/menu.c
Consolatis: Good idea, will add.

Consolatis:
> Overall it worked great! I left some bikeshedding comments, but I think we can merge this PR once submenu icons are supported.

Added support for submenu icons. Also added a commit to make the icon theme fall  back to the openbox theme name when not set.

> I'm a bit afraid that we are not handling the case where menus are so small that icons don't fit in it, but looks like the issue has been already present when large paddings are set, even without this PR. Here's an example configuration:
> 
> ```
> menu.items.padding.x: 50
> menu.width.max: 50
> menu.width.min: 50
> ```

Hm.. yes. That looks pretty broken indeed. Might be best to tackle this in another PR.

docs/labwc-menu.5.scd
tokyo4j: Sorry, I realized my comment didn't make sense (I think my brain was bugged at the time)... Yes, using icon names rather than full path should be encouraged and it's worth mentioning.

I think my intention was that `An icon to be rendered in front of the menu entry` is not clear enough and we should clarify what value can be set.

So something like:

```
The name of an icon to be rendered in front of the menu entry. This attribute is
optional. The name follows naming conventions of "Icon=" entries in desktop files. 

If used, it is recommended to use the name of...
```

src/menu/menu.c
tokyo4j: Not a strong preference, but I prefer to treat it as just a random invalid value which makes icon lookups always fail, because it looks as if `(null)` comes from a bug in our parser at a first glance.

If we want to keep it, we should add a comment saying that it's handling bugs in menu generators.

johanmalm:
LGTM and thank you!

Have tested with `labwc-menu-generator -I -p` and `obmenu-generator -i`. Works great.

Have a couple of minor nits:

1. Suggest `s/Defaults to the Openbox theme name/Inherits <theme><name> if not set/`. Prefer to avoid using `Openbox` unless there is a good reason.
2. In `labwc-menu(5)` consider adding an icon attribute to the "externally" defined `<menu>` - and a comment by the `<menu id="">` pointing to a menu defined externally.
3. Regarding the `(null)`, I think it's only `labwc-menu-generator` that produces `(null)`. `obmenu-generator` does not seem to. I think it'd be better to fix that upstream - which of course is within our own gift :smile:

01micko:
Just found a chance to test and works nicely. So a LGTM from me :smile:

docs/labwc-menu.5.scd
Consolatis: changed

src/menu/menu.c
Consolatis: Removed the `(null)` check for now. I think we should still have it somewhere though, maybe in form of a block list at `desktop_entry_load_icon()`.

Consolatis:
> 1. Suggest `s/Defaults to the Openbox theme name/Inherits <theme><name> if not set/`. Prefer to avoid using `Openbox` unless there is a good reason.

changed

> 2. In `labwc-menu(5)` consider adding an icon attribute to the "externally" defined `<menu>` - and a comment by the `<menu id="">` pointing to a menu defined externally.

Added a new "external" menu and referenced it via comment

> 3. Regarding the `(null)`, I think it's only `labwc-menu-generator` that produces `(null)`. `obmenu-generator` does not seem to. I think it'd be better to fix that upstream - which of course is within our own gift 😄

I've seen this in the `openbox-menu` generator as well. Removed the `(null)` check.

This PR is ready from my point of view.

johanmalm:
Great effort. Let's go for it :smile:

Consolatis:
> Great effort.

It's mostly using the scaled-icon-buffer as implemented by tokyo4j which now made this very easy to support. Credit where credit is due :)

01micko:
Big thanks for this @Consolatis and @johanmalm and a special thanks to @b1rger for getting it up to debian/unstable rather promptly :smiley: 

Here is bunsenlabs (on testing/unstable) running it labwc

@hhhorb is great with theming while I helped out converting the colors in the boron jgmenu icon theme (material-solarized-suruplusplus), which was as easy as a sed one liner. Thankfully they are SVG :)

PR #2510, opened 2025-01-10
src/img/img.c: cache and reuse lab_img_data

Consolatis:
We do already cache wlr_buffers for scaled-scene-buffer implementations but we still allocate new svg handles / lab_data_buffers for each window icon. And if we end up merging the menu icon support that gets amplified by a lot.

What this patch does is to add another caching layer in `src/img/img.c` by simply storing the path requested within the lab_img_data struct (which will be different for differently sized png and xpm images but the same for svg ones). When creating a new lab_img_data buffer we first check to see if a lab_img_data source is already using that same path and if yes directly use that one.

This patch does not implement caching for xbm icons because they are copies of the ones in theme anyway (so already share the same lab_data_buffer) and also require another variable to compare (their color).

Currently this should slightly increase memory consumption because we store a path per lab_img_data instance and only very slightly reduce the memory usage as only window icons really benefit from this changes. I've added a 2nd commit which tracks the number of `lab_img` and `lab_img_data` being active. If you spawn multiple instance of the same application those numbers should differ because the window icon lab_img_data will be reused rather than recreated.

Draft for now because it has only been slightly tested.


tokyo4j:
I agree with the file-based caches in `lab_img_data` and buffer-based caches in `scaled_scene_buffer`.

However, as I said in IRC, this conflicts with my work-in-progress [scaled_icon_buffer](https://github.com/tokyo4j/labwc/tree/scaled-icon-buffer) implementation that loads `lab_img` from app-id and destroys it in `_create_buffer()`... I've spend hours thinking about how to resolve this problem, but I haven't found a good way for it. Do you have any ideas?

Consolatis:
I didn't think this through completely but I think when we switch the window (and menu) icons over to scaled_icon_buffer this PR isn't needed anymore. If we for example have 2 windows with the same window icon, the 2nd window would use the wlr_buffer of the first one. If the scale changes for both, the first one handled would load a temporary lab_img with the new path, create a wlr_buffer out of it, use it and store it in its internal LRU cache. The second window would then find that wlr_buffer and use it as well. Same story for menus.

I think the only "issue" (e.g. not optimal solution) with that approach is svg handles, as the lab_imgs created are temporary they would not be kept alive and would need to be reopened rather than just using the existing handle to render in a new scale. I am not sure if that is an issue in practice though. When moving windows between outputs with different scales, the shared LRU caches of the scaled_scene_buffer should ensure that there are at least 2 different scales available per icon in the form of wlr_buffers (which are most likely already textures on the wlroots side). So having 5 `icon` parts in the titlelayout should only make one of them load a file and render a wlr_buffer and the other 4 reusing that buffer.

tokyo4j:
OK. I've been seeking a solution that allows us to share icon buffers with different sizes (e.g. window icon, menu icon and window switcher icons) and to reuse the same svg handle on scale updates, which my `scaled_icon_buffer` implementation can't. But at least, my implementation allows us to share buffers for titlebar icons between multiple windows of the same app and share buffers for menu icons with the same icon name, and this will be enough for most cases.

I'll open for a PR for `scaled_icon_buffer` later.

> And if we end up merging the menu icon support that gets amplified by a lot.

Just for curiosity, do you mean users often set multiple menu icons with the same name? I haven't extensively used Openbox's menus and I'm not sure how it is used in practice...

Consolatis:
> Just for curiosity, do you mean users often set multiple menu icons with the same name? I haven't extensively used Openbox's menus and I'm not sure how it is used in practice...

I think that is something that might happen with automatic application menu generators (be it via pipemenu or generating `menu.xml`). Foot for example has 3 entries all sharing the same icon and its just one example. Additionally, the icons might be shared with window icons. Some menu generators might also use generic icon names like "terminal" for every terminal found on the system.

You may also have multiple application menus, for example an own toplevel menu with its own key/mousebind and a separate one embedded in other menus like `root-menu`. In that case you basically have every application icon on the system at least twice.

Consolatis:
I'll close this PR for now as we now have the scaled icon buffer API. If we end up needing a lab_img cache as well in the future we can always revive it.

Issue #2513, opened 2025-01-10
labwc in wayvnc does not apply keyboard repeatRate/repeatDelay until Reconfigure


quite:
### Problem description

I'm running labwc 0.8.2 in wayvnc 0.9.1.

It seems the repeat rate/delay and possibly also other details about the keyboard are not applied until doing action Reconfigure.

With the following patch, the error output does not appear until action Reconfigure (following the steps below):

```
diff --git a/src/seat.c b/src/seat.c
index a1924f42..2cdaf933 100644
--- a/src/seat.c
+++ b/src/seat.c
@@ -641,7 +641,9 @@ seat_reconfigure(struct server *server)
        cursor_reload(seat);
        overlay_reconfigure(seat);
        keyboard_reset_current_keybind();
+       wlr_log(WLR_ERROR, "gonna iterate");
        wl_list_for_each(input, &seat->inputs, link) {
+               wlr_log(WLR_ERROR, "inp: %s", input->wlr_input_device->name);
                switch (input->wlr_input_device->type) {
                case WLR_INPUT_DEVICE_KEYBOARD:
                        configure_keyboard(seat, input);
```

Upon action Reconfigure, I get:

```
00:00:20.326 [../src/seat.c:644] gonna iterate
00:00:20.326 [../src/seat.c:646] inp: wlr_virtual_pointer_v1
00:00:20.326 [../src/seat.c:646] inp: wlr_virtual_keyboard_v1
```

### Steps to reproduce

- Launch a headless labwc with config for a high keyboard repeatRate that you can distinquish.
- Attach wayvnc to it
- Run your vnc client
- Notice that the repeat rate is not that high.
- Run Reconfigure from the menu
- Enjoy your turbo keyboard input

### labwc build source

Release

### labwc version

0.8.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch

Consolatis:
That looks like the same underlying issue as #2511. We use the keyboard group to sync up the repeat information to all keyboards and virtual keyboards are not part of that keyboard group. Added a fix in #2514. Testing appreciated.

Issue #2511, opened 2025-01-10
labwc in wayvnc does not handle modifiers for mouse binding


quite:
### Problem description

I'm running labwc 0.8.2 in wayvnc 0.9.1 and can't hold Alt and left-drag to move a window by clicking inside the frame. Are modifiers from virtual keyboards not handled for mouse bindings?

### Steps to reproduce

Launch a headless labwc with bundled config. Attach wayvnc and try to move a window by holding Alt and left-click-dragging inside the window.

### labwc build source

Release

### labwc version

0.8.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch

tokyo4j:
I think you are right.

The cause is that we read the modifier state from `wlr_keyboard_group` when matching mousebinds, but we don't add virtual keyboards to `wlr_keyboard_group`. On the other hand, keybinds with modifiers work fine because the modifer state is read from the very keyboard which has sent the key-press .

tokyo4j:
I don't know how to fix this problem. If you've got an idea, I appreciate your PR.

I even don't know the motivation of separating virtual keyboards from the keyboard group... what happens if we don't?

Consolatis:
Attaching virtual keyboards to the keyboard group makes all members of the group share a keymap. As virtual keyboards usually have a completely custom keymap things start to get funky. Virtual keyboards may also be created due to use of `wtype` or `wltrctl` so a hacky use of either to for example trigger some labwc keybind would change the keymap of all physical keyboards.

My rough idea to fix this is to keep getting the modifiers from the keyboard group but then `|=` all modifiers of virtual keyboards into it and then make the mousebinds call that new function instead.

Consolatis:
The keyboard group automatically syncs the state of multiple physical keyboards. I can't remember on the top of my head why we needed that though. Maybe @johanmalm or @jlindgren90 remember the reason why we implemented it in the first place?

> Then I think we should apply it for all other parts that read modifier state from the keyboard group, like region snapping or deactivation of window/workspace switcher.

Good point, will include a fix for those ones as well.

Consolatis:
@quite testing of #2514 welcome, it should fix this issue and the ones brought up by tokyo4j.

tokyo4j:
[...]

PR #2514, opened 2025-01-10
keyboard: add keyboard_get_all_modifiers()

Consolatis:
And make mousebind handlers use that one.
Also remove `keyboard_any_modifiers_pressed()` and replace its usage with the new function.

Without this patch we would only request the modifier state of the keyboard group which makes mousebinds involving keyboard modifiers break for virtual keyboards like when using wayvnc. Same story for hiding the workspace overlay or snapping to regions.

Also includes a commit to set the repeat info for new virtual keyboards as we usually use the keyboard group for synchronizing that property.

Fixes:
- #2511
- #2513

### DIFF
--- a/include/input/keyboard.h
+++ b/include/input/keyboard.h
@@ -22,6 +22,6 @@ void keyboard_update_layout(struct seat *seat, xkb_layout_index_t layout);
 void keyboard_cancel_keybind_repeat(struct keyboard *keyboard);
 void keyboard_cancel_all_keybind_repeats(struct seat *seat);
 
-bool keyboard_any_modifiers_pressed(struct wlr_keyboard *keyboard);
+uint32_t keyboard_get_all_modifiers(struct seat *seat);
 
 #endif /* LABWC_KEYBOARD_H */
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -17,9 +17,10 @@
 #include "dnd.h"
 #include "idle.h"
 #include "input/gestures.h"
-#include "input/touch.h"
+#include "input/keyboard.h"
 #include "input/tablet.h"
 #include "input/tablet-tool.h"
+#include "input/touch.h"
 #include "labwc.h"
 #include "layers.h"
 #include "menu/menu.h"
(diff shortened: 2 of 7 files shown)


quite (commented):
I built and run this. It makes it possible to hold Alt and move/resize window by left/right click-dragging!

quite (commented):
There is also no need to run action Reconfigure to get correct keyboard repeat after attaching my vnc client

tokyo4j:
LGTM, but I suggest adding some comments in `keyboard_get_all_modifiers()` about we need to sum up the modifiers from the keyboard group and virtual keyboards.

johanmalm:
> LGTM, but I suggest adding a comment in `keyboard_get_all_modifiers()` about we need to sum up the modifiers from the keyboard group and virtual keyboards.

Good point :smile:

LGTM from me too.

Brought back memories of this one: https://github.com/labwc/labwc/commit/e77dddbc59836b2a4250529433b5f11399722d16
...but I don't think (fingers crossed) we have to worry about that for this PR.

Consolatis:
Added a comment that hopefully explains the reasoning.

> Brought back memories of this one: [e77dddb](https://github.com/labwc/labwc/commit/e77dddbc59836b2a4250529433b5f11399722d16)
> ...but I don't think (fingers crossed) we have to worry about that for this PR.

I don't think that is an issue here. The modifiers are a bitset of `enum wlr_keyboard_modifier` so should be stable across different layouts and keymaps.

tokyo4j:
Thanks!

Just my random thought: maybe we should see the global modifier state rather than the per-keyboard state also for keybinds. Maybe it helps minor users who simultaneously use two keyboards.

Consolatis:
From what I remember, the keyboard group synchronizes the modifier state between all members so I assume that would already work but I never tested it. And mixing keyboard modifiers from a virtual and physical keyboard sounds like a extremely unlikely use-case.

tokyo4j:
Oh, I forgot that modifiers are also synchronized. I tested it by plugging a USB keyboard to my laptop and it indeed worked!

PR #2518, opened 2025-01-12
Introduce scaled_icon_buffer and <theme><fallbackAppIcon>

tokyo4j:
- The 3rd commit is just a refactor of `lab_img`. The horizontal padding in titlebar app icon is now applied by horizontally shrinking and shifting the icon buffer rather than adding empty spaces in the buffer in `lab_img_render()`.

- The 4th commit adds `str_equal()`. This is used by `_equal()` in `scaled_img_buffer` and `scaled_font_buffer`.

- The 5th commit introduces `scaled_icon_buffer`. Its benefits are:
  - The best icon file is now dynamically chosen for each output scales.
  - The titlebar app icons for the same application now shares its buffers and file handlers. This will also work for menu icons (#2509).

- The 6th commit adds `<theme><fallbackAppIcon>` used by `scaled_icon_buffer` when the icon for an app-id cannot be loaded. Its default value is 'labwc', which means we now fall back to the labwc icon (rather than 'menu' icon) when titlebar icon cannot be not loaded.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -483,6 +483,12 @@ extending outward from the snapped edge.
 *<theme><icon>*
 	The name of the icon theme to use. It is not set by default.
 
+*<theme><fallbackAppIcon>*
+	The name of the icon to use as a fallback when the application icon
+	(e.g. window icon in the titlebar) is not available. The name follows
+	the ones specified in "Icon=" entries in desktop files.
+	Default is 'labwc'.
+
 *<theme><titlebar><layout>*
 	Selection and order of buttons in a window's titlebar.
 	The following identifiers can be used, each only once:
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -30,6 +30,7 @@
   <theme>
     <name></name>
     <icon></icon>
+    <fallbackAppIcon>labwc</fallbackAppIcon>
     <titlebar>
       <layout>icon:iconify,max,close</layout>
       <showTitle>yes</showTitle>
--- a/src/common/scaled-icon-buffer.c
+++ b/src/common/scaled-icon-buffer.c
@@ -1,0 +1,123 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include <string.h>
+#include "common/macros.h"
+#include "common/mem.h"
+#include "common/scaled-icon-buffer.h"
+#include "common/scaled-scene-buffer.h"
+#include "common/string-helpers.h"
+#include "config.h"
+#include "config/rcxml.h"
+#include "desktop-entry.h"
+#include "img/img.h"
+#include "node.h"
+
+static struct lab_data_buffer *
+_create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
+{
+#if HAVE_LIBSFDO
+	struct scaled_icon_buffer *self = scaled_buffer->data;
+	int icon_size = MIN(self->width, self->height);
+	struct lab_img *img = NULL;
+
+	if (self->icon_name) {
+		img = desktop_entry_load_icon(self->server,
+			self->icon_name, icon_size, scale);
+	} else if (self->app_id) {
+		img = desktop_entry_load_icon_from_app_id(self->server,
+			self->app_id, icon_size, scale);
+	}
+
+	if (!img) {
+		return NULL;
+	}
+
+	struct lab_data_buffer *buffer =
+		lab_img_render(img, self->width, self->height, scale);
+	lab_img_destroy(img);
+
+	return buffer;
+#else
+	return NULL;
+#endif
+}
+
+static void
+_destroy(struct scaled_scene_buffer *scaled_buffer)
+{
+	struct scaled_icon_buffer *self = scaled_buffer->data;
+	free(self->app_id);
+	free(self->icon_name);
+	free(self);
+}
+
+static bool
+_equal(struct scaled_scene_buffer *scaled_buffer_a,
+	struct scaled_scene_buffer *scaled_buffer_b)
+{
+	struct scaled_icon_buffer *a = scaled_buffer_a->data;
+	struct scaled_icon_buffer *b = scaled_buffer_b->data;
+
+	return str_equal(a->app_id, b->app_id)
+		&& str_equal(a->icon_name, b->icon_name)
+		&& a->width == b->width
+		&& a->height == b->height;
+}
+
+static struct scaled_scene_buffer_impl impl = {
+	.create_buffer = _create_buffer,
+	.destroy = _destroy,
+	.equal = _equal,
+};
+
+struct scaled_icon_buffer *
+scaled_icon_buffer_create(struct wlr_scene_tree *parent, struct server *server,
+	int width, int height)
+{
+	struct scaled_scene_buffer *scaled_buffer = scaled_scene_buffer_create(
+		parent, &impl, /* drop_buffer */ true);
+	struct scaled_icon_buffer *self = znew(*self);
+	self->scaled_buffer = scaled_buffer;
+	self->scene_buffer = scaled_buffer->scene_buffer;
+	self->server = server;
+	self->width = width;
+	self->height = height;
+
+	scaled_buffer->data = self;
+
+	return self;
+}
+
+void
+scaled_icon_buffer_set_app_id(struct scaled_icon_buffer *self,
+	const char *app_id)
+{
+	assert(app_id);
+	if (str_equal(self->app_id, app_id)) {
+		return;
+	}
+	xstrdup_replace(self->app_id, app_id);
+	scaled_scene_buffer_request_update(self->scaled_buffer, self->width, self->height);
+}
+
+void
+scaled_icon_buffer_set_icon_name(struct scaled_icon_buffer *self,
+	const char *icon_name)
+{
+	assert(icon_name);
+	if (str_equal(self->icon_name, icon_name)) {
+		return;
+	}
+	xstrdup_replace(self->icon_name, icon_name);
+	scaled_scene_buffer_request_update(self->scaled_buffer, self->width, self->height);
+}
+
+struct scaled_icon_buffer *
+scaled_icon_buffer_from_node(struct wlr_scene_node *node)
+{
+	struct scaled_scene_buffer *scaled_buffer =
+		node_scaled_scene_buffer_from_node(node);
+	assert(scaled_buffer->impl == &impl);
+	return scaled_buffer->data;
+}
--- a/src/common/scaled-icon-buffer.c
+++ b/src/common/scaled-icon-buffer.c
@@ -27,6 +27,10 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
 	} else if (self->app_id) {
 		img = desktop_entry_load_icon_from_app_id(self->server,
 			self->app_id, icon_size, scale);
+		if (!img) {
+			img = desktop_entry_load_icon(self->server,
+				rc.fallback_app_icon_name, icon_size, scale);
+		}
 	}
 
 	if (!img) {
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1073,6 +1073,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		xstrdup_replace(rc.theme_name, content);
 	} else if (!strcmp(nodename, "icon.theme")) {
 		xstrdup_replace(rc.icon_theme_name, content);
+	} else if (!strcasecmp(nodename, "fallbackAppIcon.theme")) {
+		xstrdup_replace(rc.fallback_app_icon_name, content);
 	} else if (!strcasecmp(nodename, "layout.titlebar.theme")) {
 		fill_title_layout(content);
 	} else if (!strcasecmp(nodename, "showTitle.titlebar.theme")) {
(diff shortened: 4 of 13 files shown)


tokyo4j:
Looks like the icons are not updated on Reconfigure when multiple `scaled_icon_buffer`s are spawned... I guess an invalidation mechanism like #2510 is required.

src/common/scaled-icon-buffer.c
```diff
@@ -0,0 +1,129 @@
...
+#include "common/mem.h"
+#include "common/scaled-icon-buffer.h"
+#include "common/scaled-scene-buffer.h"
+#include "common/string-helpers.h"
+#include "config.h"
+#include "config/rcxml.h"
+
+#if HAVE_LIBSFDO
+#include "desktop-entry.h"
+#endif
```
Consolatis, line 16: Can we do the `#if HAVE_LIBSFDO` check in `desktop-entry.h` instead? Similar to `xwayland.h`. That prevents having to guard it for each include.

src/common/scaled-icon-buffer.c
```diff
@@ -0,0 +1,129 @@
...
+static bool
+_equal(struct scaled_scene_buffer *scaled_buffer_a,
+	struct scaled_scene_buffer *scaled_buffer_b)
+{
+	struct scaled_icon_buffer *a = scaled_buffer_a->data;
+	struct scaled_icon_buffer *b = scaled_buffer_b->data;
+
+	return str_equal(a->app_id, b->app_id)
+		&& a->width == b->width
+		&& a->height == b->height;
```
Consolatis, line 75: misses a `&& str_equal(a->icon_name, b->icon_name)`

src/common/scaled-icon-buffer.c
```diff
@@ -0,0 +1,129 @@
...
+	return self;
+}
+
+void
+scaled_icon_buffer_set_app_id(struct scaled_icon_buffer *self,
+	const char *app_id)
+{
+	assert(app_id);
+	free(self->app_id);
+	self->app_id = xstrdup(app_id);
```
Consolatis, line 108: could use `xstrdup_replace()` from `common/mem.h`. Same for `set_icon_name()`.

src/common/string-helpers.c
```diff
@@ -200,3 +200,8 @@ str_starts_with(const char *s, char needle, const char *ignore_chars)
 	return (s + strspn(s, ignore_chars))[0] == needle;
 }
 
+bool
+str_equal(const char *a, const char *b)
+{
+	return a == b || (a && b && !strcmp(a, b));
+}
```
Consolatis, line 207: Is it intentional that this returns `true` on both args being `NULL`? If yes, suggest adding it to the docs of `str_equal()`.

src/config/rcxml.c
```diff
@@ -157,7 +157,13 @@ fill_section(const char *content, struct wl_list *list)
...
 		enum ssd_part_type type = LAB_SSD_NONE;
 		if (!strcmp(identifier, "icon")) {
-			type = LAB_SSD_BUTTON_WINDOW_ICON;
+			if (HAVE_LIBSFDO) {
+				type = LAB_SSD_BUTTON_WINDOW_ICON;
+			} else {
+				wlr_log(WLR_ERROR, "libsfdo is not linked. "
+					"Replacing 'icon' in titlebar layout with 'menu'.");
+				type = LAB_SSD_BUTTON_WINDOW_MENU;
+			}
```
Consolatis, line 166: Why the runtime check rather than `#if #else #endif`?

src/config/rcxml.c
```diff
@@ -1714,8 +1679,21 @@ post_processing(void)
 		load_default_mouse_bindings();
 	}
 
+	if (!rc.fallback_app_icon_name) {
+		rc.fallback_app_icon_name = xstrdup("labwc");
+	}
```
Consolatis, line 1684: Should this use `string_null_or_empty()`? An empty string could never be an icon to load.

src/config/rcxml.c
```diff
@@ -1714,8 +1679,21 @@ post_processing(void)
...
+		if (HAVE_LIBSFDO) {
+			fill_title_layout("icon:iconify,max,close");
+		} else {
+			/*
+			 * 'icon' is replaced with 'menu' in fill_title_layout()
+			 * when libsfdo is not linked, but we also replace it
+			 * here not to show error message with default settings.
+			 */
+			fill_title_layout("menu:iconify,max,close");
+		}
```
Consolatis, line 1696: Same question as above, why no compile time check?

src/ssd/ssd-part.c
```diff
@@ -100,18 +101,37 @@ add_scene_button(struct wl_list *part_list, enum ssd_part_type type,
...
+	 * already and are zero by default.
+	 */
+	int icon_padding = button_width / 10;
+
+	if (type == LAB_SSD_BUTTON_WINDOW_ICON) {
+		struct scaled_icon_buffer *icon_buffer =
+			scaled_icon_buffer_create(parent, view->server,
+				button_width - 2 * icon_padding, button_height);
+		assert(icon_buffer);
+		nodes[0] = &icon_buffer->scene_buffer->node;
```
Consolatis, line 121: I can't quite follow this code, why is a window icon node always `node[0]`? The branch below seems to potentially also write to `node[0]` as `state_set` starts at `0` as well.

src/ssd/ssd-titlebar.c
```diff
@@ -624,22 +605,11 @@ ssd_update_window_icon(struct ssd *ssd)
...
-					scaled_img_buffer_from_node(node);
-				scaled_img_buffer_update(img_buffer, icon_img,
-					theme->window_button_width,
-					theme->window_button_height,
-					icon_padding);
-			}
-		}
+		struct scaled_icon_buffer *icon_buffer =
+			scaled_icon_buffer_from_node(button->nodes[0]);
+		scaled_icon_buffer_set_app_id(icon_buffer, app_id);
```
Consolatis, line 611: Is it ensure that `nodes[0]` is always available? If yes I think an `assert(icon_buffer)` and potentially a short comment explaining why its always available would be good here.

src/common/scaled-icon-buffer.c
Consolatis, line 108: Should this (and `set_icon_name()`) do an additional `str_equal()` here to prevent unnecessarily reloading icons?

Consolatis (commented):
Just a quick review, didn't test nor looked in depth into the logic so far.

Consolatis:
> Looks like the icons are not updated on Reconfigure when multiple `scaled_icon_buffer`s are spawned... I guess an invalidation mechanism like #2510 is required.

I think we should also change the reconfigure logic to theme_finish() -> undecorate() (or the other way around) -> theme_load() -> decorate(). I have the slight feeling that this will bite us over and over again.

src/config/rcxml.c
Consolatis: On further consideration, the current behavior of accepting an empty string might actually be beneficial if a user prefers to not have any icon rather than a fallback icon.

tokyo4j:
Sorry, my assumption was wrong. I replaced `fallbackAppIcon` with an empty string and the icon was not updated, but it was due to missing `zfree()` in `rcxml_finish()`. But still, `scaled_icon_buffer` has a problem of missed updates on Reconfigure when the icon theme is updated.

> I think we should also change the reconfigure logic to theme_finish() -> undecorate() (or the other way around) -> theme_load() -> decorate(). I have the slight feeling that this will bite us over and over again.

This will work for most cases, but if `scaled_icon_buffer` is used in other places like menu and window switcher and their sizes are coincidentally identical, your solution might not work.

I think the solutions for this are:
- Pass `server->sfdo` rather than the entire `server` to `scaled_icon_buffer` and compare it too in `_equal()`
- Clear the `cached_buffers` at the beginning of reconfigure

src/common/scaled-icon-buffer.c
tokyo4j: Yeah I agree with your idea. Will fix.

src/config/rcxml.c
tokyo4j: I assume an empty string is never accepted since `xmlIsBlankNode()` returns true:

https://github.com/tokyo4j/labwc/blob/3a3c60fe33f68e774ff54b5d2a2bbe84e0f609e5/src/config/rcxml.c#L1276

src/common/scaled-icon-buffer.c
tokyo4j: Thanks.

src/common/scaled-icon-buffer.c
tokyo4j: Good point. Added early-returns.

tokyo4j: Done.

src/common/string-helpers.c
tokyo4j: Updated the docs.

src/config/rcxml.c
tokyo4j: I thought it was more easier to read, but now I think it's better to keep consistency.

I updated to use macro instead.

src/config/rcxml.c
tokyo4j: Fixed.

src/ssd/ssd-part.c
tokyo4j: `nodes[0]` represents the normal (not hovered, toggled or rounded) button state. Previously `button->nodes[]` were set to the same values as the `menu` icon which has hovered/toggled/rounded states and they were later replaced with the same buffer for an icon looked up from app-id. But in this PR (before updating the patch), `scaled_icon_buffer` is set to `button->nodes[0]` on creation as the window icon doesn't have hovered/toggled/rounded states.

I updated the patch to store the window icon separately in `button->window_icon` and `button->nodes` are replaced with `button->img_buffers` which provides more type information. Hopefully it's easier to follow now.

Consolatis: Separating these two make sense to me. Should we also add a `LAB_BS_DEFAULT = 0`  (or `LAB_BS_NORMAL` or similar) to reduce confusion?

tokyo4j: Added `LAB_BS_DEFAULT`

docs/rc.xml.all
```diff
@@ -30,6 +30,7 @@
   <theme>
     <name></name>
     <icon></icon>
+    <fallbackAppIcon>labwc</fallbackAppIcon>
```
Consolatis, line 33: We could consider using `image-missing` as default but it fails to load for me even though its available in multiple themes. Not quite sure why, maybe we need to tell libsfdo about the used theme?

Consolatis:
I've split the first commits into an own PR. When testing everything seems to work as expected and code wise nothing stood out to me in particular either. I'll do another review once this PR is rebased on master with the prep PR merged but I think this looks very close to the finish line.

docs/labwc-config.5.scd
```diff
@@ -483,6 +483,11 @@ extending outward from the snapped edge.
 *<theme><icon>*
 	The name of the icon theme to use. It is not set by default.
 
+*<theme><fallbackAppIcon>*
+	The name of the icon to use as a fallback when the application icon
+	(e.g. window icon in the titlebar) is not available. The name follows
+	the ones specified in "Icon=" entries in desktop files.
```
Consolatis, line 489: Default value missing.

src/common/scaled-icon-buffer.c
```diff
@@ -0,0 +1,127 @@
...
+#include "desktop-entry.h"
+#include "img/img.h"
+#include "node.h"
+
+static struct lab_data_buffer *
+_create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
+{
+#if HAVE_LIBSFDO
+	struct scaled_icon_buffer *self = scaled_buffer->data;
+	int icon_size = MIN(self->width, self->height);
```
Consolatis, line 21: Why `MIN()` rather than `MAX()`?

src/ssd/ssd-part.c
```diff
@@ -92,36 +93,57 @@ add_scene_button(struct wl_list *part_list, enum ssd_part_type type,
...
+			struct scaled_img_buffer *img_buffer = scaled_img_buffer_create(
+				parent, imgs[state_set], rc.theme->window_button_width,
+				rc.theme->window_button_height);
+			assert(img_buffer);
+			icon_part->node = &img_buffer->scene_buffer->node;
+			wlr_scene_node_set_enabled(icon_part->node, false);
+			button->img_buffers[state_set] = img_buffer;
+		}
+		/* Initially show non-hover, non-toggled, unrounded variant */
+		wlr_scene_node_set_enabled(&button->img_buffers[0]->scene_buffer->node, true);
```
Consolatis, line 144: `img_buffers[LAB_BS_DEFAULT]->scene_buffer->node` ?

Consolatis (commented):
Some minor nitpicks, with those addressed this LGTM. Thanks for working on the final piece of the scaled window icons!

src/ssd/ssd-part.c
tokyo4j: Nice catch. Replaced `0` with `LAB_BS_DEFAULT`.

docs/labwc-config.5.scd
tokyo4j: Thanks. Added `Default is 'labwc'.`

src/common/scaled-icon-buffer.c
tokyo4j: The buffer is a `(self->width)x(self->heigh)` rectangle and the squared icon will be rendered to fit within the rectangle.

foo

Consolatis: Right, we would search for a 16x16 icon for a button with for example size 16x48. And on further consideration that appears as what should happen as we scale proportionally anyway. So having a bigger icon would not enhance anything because it would still be scaled down to the axis with less space. So `MIN()` seems correct here.

Consolatis:
Thanks!

PR #2520, opened 2025-01-12
scaled-scene-buffer: block sharing of buffers created before reconfigure

tokyo4j:
A possible solution for the problem in #2518 that icon theme updates are not applied to the titlebar icons on reconfigure if multiple windows with the same app-id are present.

This PR moves `cached_buffers` in `scaled-{font,rect,img}-buffer.c` into `scaled-scene-buffer.c` and clears it at the beginning of Reconfigure to prevent them from being reused by newly created scaled_scene_buffers.

### DIFF
--- a/include/common/scaled-scene-buffer.h
+++ b/include/common/scaled-scene-buffer.h
@@ -128,6 +128,14 @@ struct scaled_scene_buffer *scaled_scene_buffer_create(
 void scaled_scene_buffer_request_update(struct scaled_scene_buffer *self,
 	int width, int height);
 
+/**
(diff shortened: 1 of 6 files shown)


Consolatis:
The general idea sounds fine to me. The change to a single list will slightly increase the lookup time when there are lots of icon buffers (from menus) in the list. I don't think it is that important though. I also see the issue in the implementations not being tracked in `scaled_scene_buffer.c` so it would be difficult to iterate over them, even if we add some `impl->invalidate_buffer_cache()` API.

So: LGTM, not tested.

PR #2524, opened 2025-01-13
rcxml: ignore duplicated buttons in `<titlebar><layout>` individually

tokyo4j:
and internally replace `LAB_SSD_BUTTON_WINDOW_ICON` with `LAB_SSD_BUTTON_WINDOW_MENU` when libsfdo is not linked.

Extracted from #2518.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -147,7 +147,7 @@ parse_window_type(const char *type)
  * desk         D           All-desktops toggle (aka omnipresent)
  */
 static void
-fill_section(const char *content, struct wl_list *list)
+fill_section(const char *content, struct wl_list *list, uint32_t *found_buttons)
 {
 	gchar **identifiers = g_strsplit(content, ",", -1);
 	for (size_t i = 0; identifiers[i]; ++i) {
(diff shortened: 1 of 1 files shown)


Consolatis:
I don't quite understand why we need to prevent multiple instances of the same button in the first place but as it is already the behavior in master this LGTM. Not tested.

tokyo4j:
> I don't quite understand why we need to prevent multiple instances of the same button in the first place but as it is already the behavior in master

Well, that's an implementation issue.

The prior `contains_duplicates()` detects duplicates in `<titlebar><layout>` based on the raw string values and permits values like `menu,icon:`. But if we translate `icon` into `LAB_SSD_BUTTON_WINDOW_MENU` (when libsfdo is not linked), `menu,icon:` is now translated into two `LAB_SSD_BUTTON_WINDOW_MENU` in the internal list, which can cause unexpected behaviors. So the motivation for the first commit is to detect duplicates _after_ the strings are converted to the `LAB_SSD_BUTTON_*` list.

Of course we could keep the existing behavior that falls back to the default value when we encounter duplicated buttons in `<titlebar><layout>`, but I find my approach easier and it seems to follow Openbox's behavior in my quick test.

Consolatis:
> The reason why we need to prevent multiple instances of the same button is that we access to buttons in `ssd-titlebar.c` like `ssd_get_part(&subtree->parts, LAB_SSD_BUTTON_WINDOW_ICON);`. We can only get a single `ssd_part` with the same button type with the current API.

That makes sense. Technically we could also get multiple ones by not using `ssd_get_part()` but instead just looking through the list ourselves. But as said, that limitation is already there in master so it has nothing to do with this PR. And we should likely not change any of that while simultaneously also replacing the whole buffer architecture :)

PR #2529, opened 2025-01-17
ssd: expose resize.minimumArea parameter

ahesford:
When themes call for thick borders, having an even thicker extended area can be annoying, especially when trying to activate windows behind another when only the target window's border is visible. Let's provide a mechanism for adjusting or disabling the extension.

I'm open to suggestions for better configuration placement or option naming.

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -132,6 +132,8 @@
     <!-- Let client redraw its contents while resizing -->
     <drawContents>yes</drawContents>
     <cornerRange>8</cornerRange>
+    <!-- Borders are effectively 8 pixels wide regardless of visual appearance -->
+    <minimumArea>8</minimumArea>
   </resize>
 
   <focus>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -131,9 +131,16 @@
     <popupShow>Never</popupShow>
     <!-- Let client redraw its contents while resizing -->
     <drawContents>yes</drawContents>
-    <cornerRange>8</cornerRange>
     <!-- Borders are effectively 8 pixels wide regardless of visual appearance -->
     <minimumArea>8</minimumArea>
+
+    <!--
+      Set cornerRange to a positive value to increase the size of corner
+      regions for mouse actions and diagonal window resizing. When omitted,
+      the default size of the corner region is half of the titlebar height.
+
+      <cornerRange>8</cornerRange>
+    -->
   </resize>
 
   <focus>
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -152,6 +152,7 @@ struct rcxml {
 	enum resize_indicator_mode resize_indicator;
 	bool resize_draw_contents;
 	int resize_corner_range;
+	int resize_minimum_area;
 
 	struct {
 		int popuptime;
--- a/src/ssd/ssd-extents.c
+++ b/src/ssd/ssd-extents.c
@@ -4,6 +4,7 @@
 #include <pixman.h>
 #include "common/mem.h"
 #include "common/scene-helpers.h"
+#include "config/rcxml.h"
 #include "labwc.h"
 #include "ssd-internal.h"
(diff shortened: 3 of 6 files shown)


ahesford:
These are good ideas.

Driving the extended area to zero breaks bidirectional resizing the window corners. I need to understand a bit better how this part of the program works. After this is worked out, I'll refactor the extent calculation along the lines of these suggestions.

ahesford:
Alright, this seems to be in a decent state:
- Instead of exposing an "extension" parameter, we instead expose `theme.minimumEffectiveBorder`, which is (as the name suggests) the smallest allowed "effective" border zone that will receive mouse actions and induce cursor changes. The actual border zone is the larger of this parameter or the visible border.
- To fix diagonal corner resizing, we now expose a `theme.cornerWidth` parameter, which defines the realm in which mousing over a single-edge border will be interpreted as a corner hover.

I'm not sure how to cleanly handle not adding extents when their size goes to zero, especially when reconfiguring may cause them to suddenly appear or disappear. For now, just drawing zero-thickness zones seems to work as expected even when reconfiguring causes them to appear or disappear.

There might still be some quirks with overlapping corner extents and edges (either visible borders or hidden extents), especially in the titlebar when rounded corners are enabled. The overlap can be seen by forcing borders to each have unique, partially transparent colors while forcing the invisible extents to have clashing, partially transparent colors. However, the overlap does not seem to cause obvious behavioral bugs, so I don't know that we need to address the overlap.

ahesford:
Taking this back to draft until #2552 is settled, which expands on the corner work I started above.

ahesford:
This is now called `resize.minimumArea` and rebased on top of the merged border-handling changes, and is ready for review.

include/config/rcxml.h
```diff
@@ -88,6 +88,7 @@ struct rcxml {
 	struct wl_list title_buttons_left;
 	struct wl_list title_buttons_right;
 	int corner_radius;
+	int corner_size;
```
Consolatis, line 91: This looks like a left-over from an earlier iteration

docs/rc.xml.all
```diff
@@ -132,6 +132,8 @@
     <!-- Let client redraw its contents while resizing -->
     <drawContents>yes</drawContents>
     <cornerRange>8</cornerRange>
+    <!-- Borders are effectively 8 pixels wide regardless of visual appearance -->
+    <minimumArea>8</minimumArea>
```
Consolatis, line 136: Missed that in the earlier PR, `<cornerRange>` should be removed and potentially replaced with a comment about it defaulting to half of titlebar size. Not sure about the comment though.

src/ssd/ssd-extents.c
```diff
@@ -68,7 +69,8 @@ ssd_extents_update(struct ssd *ssd)
 	int height = view_effective_height(view, /* use_pending */ false);
 	int full_height = height + theme->border_width * 2 + ssd->titlebar.height;
 	int full_width = width + 2 * theme->border_width;
-	int extended_area = SSD_EXTENDED_AREA;
+	int border_width = MAX(0, MAX(rc.resize_minimum_area, theme->border_width));
```
Consolatis, line 72: Why do we need the outer `MAX(0, ..)` here? if either of the vars are negative I'd assume other things will break as well.

Consolatis (commented):
Some comments below. With those addressed this LGTM. Not tested.

include/config/rcxml.h
ahesford: Removed

src/ssd/ssd-extents.c
ahesford: We don't, really, because `rc.resize_minimum_area` is already clipped to a non-negative value when parsing `rc.xml`. Removed.

docs/rc.xml.all
ahesford: Good catch. I replaced the stale entry with a comment explaining its use.

PR #2531, opened 2025-01-18
Pipemenu rework

tokyo4j:
In this PR, top-level pipemenus are dynamically generated rather than generating only on initialization.

But this PR also includes many refactors about how menus are generated and opened. Notably:
- All the menu scene-graphs are now lazily generated and repositioned when being opened. This fixes the problem of incorrect submenu position in configurations like:
```xml
  <menu id="foo" label="foo">
    <item label="aaaaaa"/>
    <item label="bbbbbb"/>
  </menu>
  <menu id="root-menu">
    <menu id="foo" />
    <menu id="foo" />
  </menu>
```
- `struct menu` for pipemenus are created on initialization rather than after the execution of pipemenu script
- `update_client_list_combined_menu()` and `update_client_send_to_menu()` are called only when the corresponding menus are being opened

I admit this is a dangerous move as this touches the core of menu management, but I believe the UX enhancement and simplification in this PR is worth the risk.

I tested with following conditions:
- Closing menu while pipemenu script is running
- Running `labwc -r` while pipemenu script is running
- Let the pipemenu script print an invalid output

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1062,7 +1062,6 @@ update_client_list_combined_menu(struct server *server)
 				buf_add(&buffer, title);
 
 				current_item = item_create(menu, buffer.data, /*show arrow*/ false);
-				current_item->id = xstrdup(menu->id);
 				current_item->client_list_view = view;
 				fill_item("name.action", "Focus");
 				fill_item("name.action", "Raise");
@@ -1070,7 +1069,6 @@ update_client_list_combined_menu(struct server *server)
 			}
 		}
 		current_item = item_create(menu, _("Go there..."), /*show arrow*/ false);
-		current_item->id = xstrdup(menu->id);
 		fill_item("name.action", "GoToDesktop");
 		fill_item("to.action", workspace->name);
 	}
@@ -1682,7 +1680,7 @@ menu_execute_item(struct menuitem *item)
 	 * menu_close() and destroy_pipemenus() which we have to handle
 	 * before/after action_run() respectively.
 	 */
-	if (item->id && !strcmp(item->id, "client-list-combined-menu")
+	if (!strcmp(item->parent->id, "client-list-combined-menu")
 			&& item->client_list_view) {
 		actions_run(item->client_list_view, server, &item->actions, NULL);
 	} else {
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -49,9 +49,8 @@ static bool waiting_for_pipe_menu;
 static struct menuitem *selected_item;
 
 struct menu_pipe_context {
-	struct server *server;
-	struct menuitem *item;
-	struct menu *top_level_menu;
+	struct wlr_box anchor_rect;
+	struct menu *pipemenu;
 	struct buf buf;
 	struct wl_event_source *event_read;
 	struct wl_event_source *event_timeout;
@@ -88,8 +87,7 @@ menu_create(struct server *server, const char *id, const char *label)
 	menu->label = xstrdup(label ? label : id);
 	menu->parent = current_menu;
 	menu->server = server;
-	menu->is_pipemenu = waiting_for_pipe_menu;
-	menu->size.width = server->theme->menu_min_width;
+	menu->is_pipemenu_child = waiting_for_pipe_menu;
 	return menu;
 }
 
@@ -481,16 +479,11 @@ fill_item(char *nodename, char *content)
 static void
 item_destroy(struct menuitem *item)
 {
-	if (item->pipe_ctx) {
-		item->pipe_ctx->item = NULL;
-	}
 	wl_list_remove(&item->link);
 	action_list_free(&item->actions);
 	if (item->tree) {
 		wlr_scene_node_destroy(&item->tree->node);
 	}
-	free(item->execute);
-	free(item->id);
 	free(item->text);
 	free(item->icon_name);
 	free(item);
@@ -636,32 +629,6 @@ static bool parse_buf(struct server *server, struct buf *buf);
 static int handle_pipemenu_readable(int fd, uint32_t mask, void *_ctx);
 static int handle_pipemenu_timeout(void *_ctx);
 
-static void
-parse_root_pipemenu(struct menu *top_level_menu, const char *execute)
-{
-	int pipe_fd = 0;
-	pid_t pid = spawn_piped(execute, &pipe_fd);
-	if (pid <= 0) {
-		wlr_log(WLR_ERROR, "Failed to spawn pipe menu process %s", execute);
-		return;
-	}
-
-	struct menu_pipe_context *ctx = znew(*ctx);
-	ctx->server = top_level_menu->server;
-	ctx->top_level_menu = top_level_menu;
-	ctx->pid = pid;
-	ctx->pipe_fd = pipe_fd;
-	ctx->buf = BUF_INIT;
-	top_level_menu->pipe_ctx = ctx;
-
-	ctx->event_read = wl_event_loop_add_fd(ctx->server->wl_event_loop,
-		pipe_fd, WL_EVENT_READABLE, handle_pipemenu_readable, ctx);
-
-	ctx->event_timeout = wl_event_loop_add_timer(ctx->server->wl_event_loop,
-		handle_pipemenu_timeout, ctx);
-	wl_event_source_timer_update(ctx->event_timeout, PIPEMENU_TIMEOUT_IN_MS);
-}
-
 /*
  * <menu> elements have three different roles:
  *  * Definition of (sub)menu - has ID, LABEL and CONTENT
@@ -678,24 +645,24 @@ handle_menu_element(xmlNode *n, struct server *server)
 
 	if (execute && label && id) {
 		wlr_log(WLR_DEBUG, "pipemenu '%s:%s:%s'", id, label, execute);
+
+		struct menu *pipemenu = menu_create(server, id, label);
+		pipemenu->execute = xstrdup(execute);
 		if (!current_menu) {
 			/*
-			 * Handle pipemenu as the root-menu such this:
+			 * A pipemenu may not have its parent like:
 			 *
 			 * <?xml version="1.0" encoding="UTF-8"?>
 			 * <openbox_menu>
 			 *   <menu id="root-menu" label="foo" execute="bar"/>
 			 * </openbox_menu>
 			 */
-			struct menu *menu = menu_create(server, id, label);
-			parse_root_pipemenu(menu, execute);
 		} else {
 			current_item = item_create(current_menu, label,
 				/* arrow */ true);
 			fill_item("icon", icon_name);
 			current_item_action = NULL;
-			current_item->execute = xstrdup(execute);
-			current_item->id = xstrdup(id);
+			current_item->submenu = pipemenu;
 		}
 	} else if ((label && id) || is_toplevel_static_menu_definition(n, id)) {
 		/*
@@ -749,7 +716,7 @@ handle_menu_element(xmlNode *n, struct server *server)
 		 * pipemenu opening the "root-menu" or similar.
 		 */
 
-		if (current_menu && current_menu->is_pipemenu) {
+		if (waiting_for_pipe_menu) {
 			wlr_log(WLR_ERROR,
 				"cannot link to static menu from pipemenu");
 			goto error;
@@ -1178,6 +1145,8 @@ nullify_item_pointing_to_this_menu(struct menu *menu)
 	}
 }
 
+static void pipemenu_ctx_destroy(struct menu_pipe_context *ctx);
+
 static void
 menu_free(struct menu *menu)
 {
@@ -1194,7 +1163,8 @@ menu_free(struct menu *menu)
 	}
 
 	if (menu->pipe_ctx) {
-		menu->pipe_ctx->top_level_menu = NULL;
+		pipemenu_ctx_destroy(menu->pipe_ctx);
+		assert(!menu->pipe_ctx);
 	}
 
 	/*
@@ -1208,6 +1178,7 @@ menu_free(struct menu *menu)
 	zfree(menu->id);
 	zfree(menu->label);
 	zfree(menu->icon_name);
+	zfree(menu->execute);
 	zfree(menu);
 }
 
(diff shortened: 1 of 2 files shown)


src/menu/menu.c
```diff
@@ -1359,14 +1285,15 @@ menu_open_root(struct menu *menu, int x, int y)
...
+	struct wlr_box anchor_rect = {.x = x, .y = y};
+	if (menu->execute && !menu->scene_tree) {
+		open_pipemenu_async(menu, anchor_rect);
+	} else {
+		open_menu(menu, anchor_rect);
 	}
-	close_all_submenus(menu);
-	menu_set_selection(menu, NULL);
-	menu_configure(menu, (struct wlr_box){.x = x, .y = y});
-	wlr_scene_node_set_enabled(&menu->scene_tree->node, true);
```
tokyo4j, line 1369: I believe this is not needed anymore as `if (menu->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) { return; }` in the first line guarantees that all the menus are closed.

tokyo4j, line 1369: I believe we don't need to close menus here as `if (menu->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) { return; }` in the first line guarantees that all the menus are closed.

src/menu/menu.c
```diff
@@ -1082,8 +1014,6 @@ init_rootmenu(struct server *server)
 	if (!menu) {
 		current_menu = NULL;
 		menu = menu_create(server, "root-menu", "");
-	}
-	if (wl_list_empty(&menu->menuitems)) {
```
tokyo4j, line 1086: With this change, defining an empty `root-menu` in `menu.xml` makes an empty menu rather than filling it with default items. This is needed as otherwise `root-menu` defined as a pipemenu is also filled with the default items.

src/menu/menu.c
```diff
@@ -1359,14 +1285,15 @@ menu_open_root(struct menu *menu, int x, int y)
...
+	struct wlr_box anchor_rect = {.x = x, .y = y};
+	if (menu->execute) {
+		open_pipemenu_async(menu, anchor_rect);
+	} else {
+		open_menu(menu, anchor_rect);
 	}
-	close_all_submenus(menu);
-	menu_set_selection(menu, NULL);
-	menu_configure(menu, (struct wlr_box){.x = x, .y = y});
-	wlr_scene_node_set_enabled(&menu->scene_tree->node, true);
```
tokyo4j, line 1369: I believe we don't need to close menus here as `if (menu->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) { return; }` in the first line guarantees that all the menus are closed.

Consolatis:
I didn't look through the diff in detail yet but have a more general question:

Do we want toplevel pipemenus to load when being opened rather than on startup / reconfigure only? This seems to drastically increase the latency every time the menu is loaded, especially when also using menu icons for cases like application menus.

CC @johanmalm

tokyo4j:
> Do we want toplevel pipemenus to load (e.g. spawn the generator process and wait for it being done) when being opened rather than on startup / reconfigure only? This seems to drastically increase the latency every time the menu is loaded, especially when also using menu icons for cases like big application menus.

To me the overhead is acceptable, given that Openbox does it.

I actually don't have a strong preference on whether it should be generated on startup or when opened, but if we decide to generate them on startup, it feels more natural to generate all the submenus including non-top-level ones on startup.

tokyo4j:
Rebased on #2537

Conan-Kudo:
From a user experience perspective, it would make sense to generate it on request every time, because then you don't need to reload the compositor when you install or remove applications.

johanmalm:
> I didn't look through the diff in detail yet but have a more general question:
> 
> Do we want toplevel pipemenus to load (e.g. spawn the generator process and wait for it being done) when being opened rather than on startup / reconfigure only? This seems to drastically increase the latency every time the menu is loaded, especially when also using menu icons for cases like big application menus.
> 
> CC @johanmalm

I think that re-reading each time seems better.

This way we are consistent with the submenu-pipemenu behaviour and probably provide the least amount of user-surprise.

IIRC obmenu-generator caches the output and we ought to do the same with labwc-menu-generator to avoid reading all desktop files each time (different project/issue but thought I would mention it for context).

For the labwc side, I think a better answer to any overhead concern is to just use a static menu.

When first introducing pipemenus we considered having options for different cache-lifetimes, but unless there is a strong argument  for this (which I currently do not see) I prefer keeping it simple with just one mode.

src/menu/menu.c
```diff
@@ -1120,8 +1050,6 @@ init_rootmenu(struct server *server)
 	if (!menu) {
 		current_menu = NULL;
 		menu = menu_create(server, "root-menu", "");
-	}
-	if (wl_list_empty(&menu->menuitems)) {
```
tokyo4j, line 1124: With this change, defining an empty `root-menu` in `menu.xml` makes an empty menu rather than filling it with default items. This is needed because otherwise an empty root-menu defined as a pipemenu is also filled with the default items.

src/action.c
```diff
@@ -664,16 +664,6 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
...
-	/*
-	 * We always refresh client-list-combined-menu and client-send-to-menu
-	 * so that they are up-to-date whether they are directly opened as a
-	 * top-level menu or opened as a submenu which we don't know at this
-	 * point. It is also needed to calculate the proper width for placement
-	 * as it fluctuates depending on application/workspace titles.
-	 */
-	update_client_list_combined_menu(menu->server);
-	update_client_send_to_menu(menu->server);
-
```
tokyo4j, line 676: These menus are now updated in `open_menu()` which is called every time a (sub)menu is opened.

src/menu/menu.c
```diff
@@ -362,19 +360,30 @@ title_create_scene(struct menuitem *menuitem, int *item_y)
...
+}
+
+static void
+menu_create_scene(struct menu *menu)
+{
+	struct menuitem *item;
+	struct theme *theme = menu->server->theme;
+
+	assert(!menu->scene_tree);
+
```
tokyo4j, line 386: `menu_update_scene()` is now renamed to `menu_create_scene()`. It creates scene-nodes for the menu and asserts that the scene-nodes hasn't been created.

src/menu/menu.c
```diff
@@ -683,24 +645,33 @@ handle_menu_element(xmlNode *n, struct server *server)
...
-			struct menu *menu = menu_create(server, id, label);
-			parse_root_pipemenu(menu, execute);
 		} else {
 			current_item = item_create(current_menu, label,
 				/* arrow */ true);
 			fill_item("icon", icon_name);
 			current_item_action = NULL;
-			current_item->execute = xstrdup(execute);
-			current_item->id = xstrdup(id);
+			current_item->submenu = pipemenu;
```
tokyo4j, line 674: For root pipemenus, we just set `pipemenu->execute` and without creating its items or scene-nodes. They are created when being opened.

src/menu/menu.c
```diff
@@ -748,18 +719,7 @@ handle_menu_element(xmlNode *n, struct server *server)
...
-		 * nested (inline) menu definitions, otherwise we could have a
-		 * pipemenu opening the "root-menu" or similar.
 		 */
-
-		if (current_menu && current_menu->is_pipemenu) {
-			wlr_log(WLR_ERROR,
-				"cannot link to static menu from pipemenu");
-			goto error;
-		}
-
```
tokyo4j, line 762: I think this is no longer needed as we check for nested menus with the same id in the code below.

src/menu/menu.c
```diff
@@ -914,10 +874,8 @@ get_item_anchor_rect(struct theme *theme, struct menuitem *item)
 }
 
 static void
-menu_configure(struct menu *menu, struct wlr_box anchor_rect)
+menu_reposition(struct menu *menu, struct wlr_box anchor_rect)
```
tokyo4j, line 877: Renamed to `menu_reposition()` as it explains better what this function does.

`menu_reposition()` is called every time (sub)menu is opened with `open_menu()` and it no longer repositions submenus of the given menu.

src/menu/menu.c
```diff
@@ -1006,20 +955,13 @@ init_client_send_to_menu(struct server *server)
...
-		struct menuitem *item, *next;
-		wl_list_for_each_safe(item, next, &menu->menuitems, link) {
-			item_destroy(item);
-		}
-	}
+	struct menu *menu = menu_get_by_id(server, "client-send-to-menu");
+	assert(menu);
 
-	menu->size.height = 0;
+	reset_menu(menu);
```
tokyo4j, line 964: `reset_menu()` destroys all the items and scene-nodes of the menu. After that, we fill the items again and create the scene-nodes with `menu_create_scene()`.

src/menu/menu.c
```diff
@@ -1234,7 +1161,8 @@ menu_free(struct menu *menu)
 	}
 
 	if (menu->pipe_ctx) {
-		menu->pipe_ctx->top_level_menu = NULL;
+		pipemenu_ctx_destroy(menu->pipe_ctx);
```
tokyo4j, line 1164: I think destroying `menu->pipe_ctx` here makes more sense. It removes the need to check for `pipe_ctx->menu` in `handle_pipemenu_readable()`.

src/menu/menu.c
```diff
@@ -1248,34 +1176,17 @@ menu_free(struct menu *menu)
...
+	zfree(menu->execute);
 	zfree(menu);
 }
 
-/**
- * menu_free_from - free menu list starting from current point
- * @from: point to free from (if NULL, all menus are freed)
- */
-static void
-menu_free_from(struct server *server, struct menu *from)
```
tokyo4j, line 1259: `menu_free_from()` is removed in favor of simplicity.

src/menu/menu.c
```diff
@@ -1376,6 +1300,25 @@ menu_close(struct menu *menu)
 	_close(menu);
 }
 
+static void
+open_menu(struct menu *menu, struct wlr_box anchor_rect)
```
tokyo4j, line 1304: This function is called every time a (sub)menu is opened.

src/menu/menu.c
```diff
@@ -683,24 +645,33 @@ handle_menu_element(xmlNode *n, struct server *server)
...
-			parse_root_pipemenu(menu, execute);
 		} else {
 			current_item = item_create(current_menu, label,
 				/* arrow */ true);
 			fill_item("icon", icon_name);
 			current_item_action = NULL;
-			current_item->execute = xstrdup(execute);
-			current_item->id = xstrdup(id);
+			current_item->submenu = pipemenu;
 		}
```
tokyo4j, line 675: For pipemenus, a `menu` with `menu->execute` set is created rather than setting `item->execute` without creating a menu. This allows us to equally treat pipemenus with/without parent menu.

src/menu/menu.c
```diff
@@ -1815,11 +1682,12 @@ void
...
-		menu_close(server->menu_current);
-		server->menu_current = NULL;
-		destroy_pipemenus(server);
-	}
+	assert(server->menu_current);
+
+	menu_close(server->menu_current);
+	server->menu_current = NULL;
+	reset_pipemenus(server);
+
```
tokyo4j, line 1690: I replaced the check against `server->menu_current` as I don't think `server->menu_current` can be set when `server->input_mode == LAB_INPUT_STATE_MENU`.

tokyo4j:
Rebased and added some comments. This PR is now ready for review.

johanmalm:
Sorry - Have only just seen that you've set to draft. Have done some testing. See observations below.


1. Nested pipemenus are no longer supported. They should be. The browse.sh script at https://github.com/labwc/labwc/pull/1216#issuecomment-2008591173 can be used to test.

2. I get `00:00:00.174 [../src/menu/menu.c:79] menu id root-menu already exists` whereas I never used to with:


```
<?xml version="1.0" encoding="utf-8"?>
<openbox_menu>

<menu id="root-menu" label="root-menu">
  <item label="File Manager"><action name="Execute"><command><![CDATA[xdg-open .]]></command></action></item>
  <item label="Terminal"><action name="Execute"><command><![CDATA[xterm]]></command></action></item>
  <item label="Web Browser"><action name="Execute"><command><![CDATA[xdg-open http://]]></command></action></item>
  <item label="Run command"><action name="Execute"><command><![CDATA[gmrun]]></command></action></item>
  <separator/>
  <menu label="BIG  big-pipe.sh" id="big-pipe.sh" execute="big-pipe.sh ."/>
  <menu label="SLOW browse.sh" id="browse.sh" execute="browse.sh ."/>
  <menu label="SLOW pipemenu2.sh" id="pipemenu2.sh" execute="pipemenu2.sh"/>
  <menu label="pipemenu.sh" id="pipemenu.sh" execute="pipemenu.sh"/>
  <menu label="labwc-menu-generator -p" id="labwc-menu-generator-p" execute="labwc-menu-generator -p"/>
  <menu label="labwc-menu-generator -p -n" id="labwc-menu-generator-p-n" execute="labwc-menu-generator -p -n"/>
  <menu label="obmenu-generator -p" id="obmenu-generator -p" execute="obmenu-generator -p"/>
  <menu label="bl-places-pipemenu" id="bl-places-pipemenu" execute="bl-places-pipemenu"/>
  <menu label="bl-places-pipemenu" id="root-menu" execute="bl-places-pipemenu"/>
  <separator/>
  <item label="abc"><action name="Execute"><command>wlrctl keyboard type abc</command></action></item>
  <item label="Lock"><action name="Execute"><command><![CDATA[xscreensaver-command -lock]]></command></action></item>
  <item label="Exit"><action name="Exit"/></item>
</menu>

</openbox_menu>
```

tokyo4j:
> Nested pipemenus are no longer supported. They should be. The browse.sh script at https://github.com/labwc/labwc/pull/1216#issuecomment-2008591173 can be used to test.

Okay, I updated the PR to allow nested pipemenus. I initially prohibited nested pipemenus just to make sure things never go wrong, but looks like that check was not necessary.

> I get 00:00:00.174 [../src/menu/menu.c:79] menu id root-menu already exists whereas I never used to with:

Hmm, I assume the issue you describe is:
  Before: we get `duplicate id 'root-menu'; abort pipemenu` when opening the pipemenu
  After: we get `menu id root-menu already exists` error when on initialization

This is because we now create `struct menu` also for pipemenus on initialization, rather than just setting `item->execute`, in order to equally treat pipemenus with/without parent. I think we can tolerate this behavioral change as duplicated ids to static/pipe menus doesn't sound like something we should officially support.

johanmalm:
LGTM. Have tested. Works well.

PR #2537, opened 2025-01-24
menu: fix unexpected behavior when a menu is opened from another menu

### LINKED ISSUE #2526: Menu broken if invoked from another menu
### Problem description

Menu is not functional if invoked from another menu.
Also broken menu do not disappear from the screen.

### Steps to reproduce

Example menu.xml:
```
<openbox_menu>
<menu id="root-menu">
  <separator label="Main Menu" />
  <item label="Terminal"><action name="Execute" command="foot" /></item>
<separator />
  <item label="Settings menu"><action name="ShowMenu" menu="settings-menu" /></item>
</menu>
<menu id="settings-menu">
  <item label="back to main menu"><action name="ShowMenu" menu="root-menu" /></item>
  <separator label="Edit config file..."/>
  <item label="rc.xml"><action name="Execute" command="geany ~/.config/labwc/rc.xml" /></item>
  <item label="autostart"><action name="Execute" command="geany ~/.config/labwc/autostart" /></item>
  <item label="environment"><action name="Execute" command="geany ~/.config/labwc/environment" /></item>
</menu>
</openbox_menu>
```

### labwc build source

Release

### labwc version

0.8.2,  0.8.2-88-g08d31145

### labwc environment

### Distribution

Mabox Linux

tokyo4j:
Fixes #2526.

`server->menu_current` should be cleared before calling `actions_run()` as it may internally call `menu_open_root()`. Clearing it after `actions_run()` leads to an inconsistent state where a menu is opened but `server->menu_current` is
`NULL`. It even lead to a segfault when the item opening another menu is contained in a pipemenu, because `menu_open_root()` calls `destroy_pipemenu()` when `server->menu_current` is set, which makes accessing `item->actions` a UAF.

This PR also replaces some code that ensures the menu is closed in `menu_open_root()` with `assert(!menu->server->menu_current);`.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1327,20 +1327,6 @@ menu_set_selection(struct menu *menu, struct menuitem *item)
 	menu->selection.item = item;
 }
 
-static void
-close_all_submenus(struct menu *menu)
-{
-	struct menuitem *item;
-	wl_list_for_each(item, &menu->menuitems, link) {
-		if (item->submenu) {
-			wlr_scene_node_set_enabled(
-				&item->submenu->scene_tree->node, false);
-			close_all_submenus(item->submenu);
-		}
-	}
-	menu->selection.menu = NULL;
-}
-
 /*
  * We only destroy pipemenus when closing the entire menu-tree so that pipemenu
  * are cached (for as long as the menu is open). This drastically improves the
@@ -1388,17 +1374,14 @@ menu_close(struct menu *menu)
 void
 menu_open_root(struct menu *menu, int x, int y)
 {
+	assert(menu);
+
 	if (menu->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
 		return;
 	}
 
-	assert(menu);
-	if (menu->server->menu_current) {
-		menu_close(menu->server->menu_current);
-		destroy_pipemenus(menu->server);
-	}
-	close_all_submenus(menu);
-	menu_set_selection(menu, NULL);
+	assert(!menu->server->menu_current);
+
 	menu_configure(menu, (struct wlr_box){.x = x, .y = y});
 	wlr_scene_node_set_enabled(&menu->scene_tree->node, true);
 	menu->server->menu_current = menu;
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1720,13 +1720,9 @@ menu_execute_item(struct menuitem *item)
 		return false;
 	}
 
-	/*
-	 * We close the menu here to provide a faster feedback to the user.
-	 * We do that without resetting the input state so src/cursor.c
-	 * can do its own clean up on the following RELEASE event.
-	 */
 	struct server *server = item->parent->server;
 	menu_close(server->menu_current);
+	server->menu_current = NULL;
 	seat_focus_override_end(&server->seat);
 
 	/*
@@ -1746,7 +1742,6 @@ menu_execute_item(struct menuitem *item)
 				&item->actions, NULL);
 	}
 
-	server->menu_current = NULL;
 	destroy_pipemenus(server);
 	return true;
 }
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -115,7 +115,12 @@ validate_menu(struct menu *menu)
 	struct action *action, *action_tmp;
 	wl_list_for_each(item, &menu->menuitems, link) {
 		wl_list_for_each_safe(action, action_tmp, &item->actions, link) {
-			if (!action_is_valid(action)) {
+			bool is_show_menu = action_is_show_menu(action);
+			if (!action_is_valid(action) || is_show_menu) {
+				if (is_show_menu) {
+					wlr_log(WLR_ERROR, "'ShowMenu' action is"
+						" not allowed in menu items");
+				}
 				wl_list_remove(&action->link);
 				action_free(action);
 				wlr_log(WLR_ERROR, "Removed invalid menu action");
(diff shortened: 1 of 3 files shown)


src/menu/menu.c
```diff
@@ -1686,13 +1668,9 @@ menu_execute_item(struct menuitem *item)
 		return false;
 	}
 
-	/*
-	 * We close the menu here to provide a faster feedback to the user.
-	 * We do that without resetting the input state so src/cursor.c
-	 * can do its own clean up on the following RELEASE event.
-	 */
```
tokyo4j, line 1693: The second sentence no longer holds true after f3b68b8f which added `server->input_mode = LAB_INPUT_STATE_PASSTHROUGH;`.

I'm not sure about the first sentence though. CC @Consolatis

Consolatis:
Another approach could be to simply prevent this situation by early returning on the `ShowMenu` action when there is already a menu open.

src/menu/menu.c
Consolatis: Hm.. does that mean we now leak the mouse button release event to the client under the cursor? Or is that now caught with the `seat->bound_buttons` set?

The first sentence should still hold true, although it isn't very descriptive. The idea was to close the menu on press rather than release. But that then requires the release handler to not notify the client under the cursor with the now closed menu because it would then get a mouse button release without a mouse button press.

tokyo4j: > Hm.. does that mean we now leak the mouse button release event to the client under the cursor? Or is that now caught with the seat->bound_buttons set?

Yes, `seat->bound_buttons` will prevent leaking release events in usual cases like:
- `press button to open a menu with a mousebind` -> `release button on an item to execute it`
- `press button to open a menu with a mousebind` -> `release button` -> `press & release button on an item to execute it`
- `open a menu with a keybind` -> `press & release button on an item to execute it`

An exceptional case is `press button` -> `open a menu with a keybind` -> `release button on an item to execute it`. In this case, I do see an warning from foot terminal: `warn: input.c:2551: stray button release event (compositor bug?)`, because after #2455, we send `wl_pointer.leave` when opening menus, which should reset the button state. But I think this is such an edge case and not an urgent bug.

> The first sentence should still hold true, although it isn't very descriptive. The idea was to execute the actions / close the menu on press rather than release.

I see. I'll replace the comment with a more descriptive one.

tokyo4j:
> Another approach could be to simply prevent this situation by early returning on the ShowMenu action when there is already a menu open.

Yes adding an early-return like this will also fix the issue:

```diff
diff --git a/src/menu/menu.c b/src/menu/menu.c
index ad049e34..c8b93ad2 100644
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1357,6 +1357,9 @@ menu_open_root(struct menu *menu, int x, int y)
 	if (menu->server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
 		return;
 	}
+	if (menu->server->menu_current) {
+		return;
+	}
 
 	assert(menu);
 	if (menu->server->menu_current) {
```

But I think it's unclear why we need to have checks for both `server->input_mode != LAB_INPUT_STATE_PASSTHROUGH` and `server->menu_current` and I'd rather like to enforce a constraint that `server->menu_current` should be cleared while `server->input_mode != LAB_INPUT_STATE_MENU`. Keeping `server->input_mode` as `LAB_INPUT_STATE_MENU` while executing actions will also work, but I'm afraid it will affect the executions of other actions in unexpected ways.

So if we decide to reject running ShowMenu actions from a menu, I think the most robust approach is to remove ShowMenu actions in `validate()` on menu initialization. I don't have a strong preference on whether we should reject ShowMenu actions or continue with the approach in this PR.

src/menu/menu.c
Consolatis: Thanks for looking into this! In that case it seems as if we can indeed remove the whole comment.

Consolatis:
I am slightly worried that this might break in the future. As an example: a menu action spawns a new toplevel pipemenu (which currently is static but that might not always be the case). After execution of the actions we destroy pipemenus which would then include the newly opened one. In general I have a feeling that allowing `ShowMenu` while processing actions from a menu item might turn in a unmanageable mess full of edge cases.

tokyo4j:
> I am slightly worried that this might break in the future. As an example: a menu action spawns a new toplevel pipemenu (which currently is static but that might not always be the case). After execution of the actions we destroy pipemenus which would then include the newly opened one. In general I have a feeling that allowing ShowMenu while processing actions from a menu item might turn into a unmanageable mess full of edge cases.

I agree. Currently we need to deal with a strange situation where menus are closed but old pipemenus need to stay, but this might cause another unexpected bug and simply skipping menu interactions in that situation makes sense.

I'd like to still keep the changes in this PR as closing menus in `menu_open_root()` or delaying `server->menu_current = NULL;` after action executions don't make sense, but I will add another commit that removes `ShowMenu` actions from menu items.

Consolatis:
I'd prefer if we could keep the `action_type` enum internal to `action.c`. We could instead add a small function to `action.c` like `action_is_show_menu()`, similar to `actions_contain_toggle_keybinds()`. I am not particularly happy that we started to leak other internal details of `action.c` to the config parser either, that was required for the `ForEach` implementation from what I remember. But lets not continue this trend if not strictly necessary.

tokyo4j:
> I'd prefer if we could keep the action_type enum internal to action.c. We could instead add a small function to action.c like action_is_show_menu(), similar to actions_contain_toggle_keybinds(). I am not particularly happy that we started to leak other internal details of action.c to the config parser either, that was required for the ForEach implementation from what I remember. But lets not continue this trend if not strictly necessary.

Alright. I updated the patch to use `action_is_show_menu()`.

Consolatis:
Thanks :)

The Debian build error is unrelated to this PR.

Issue #2542, opened 2025-01-24
Incorrect coordinates given to pointer_motion listener during a drag from a sub-surface


CyprinusCarpio:
### Problem description

I'm developing a widget docking implementation for the FLTK toolkit. It allows detaching widget groups from the main window, which on Wayland creates a child window, or a sub-surface in Wayland parlance, the parent being the top-level. The sub surface can be moved outside the top level, and/or rescaled. However, if the user starts a drag in the detached window, and the pointer goes outside the detached window, FLTK and in turn my code get wrong coordinates. I have narrowed down the problem to the `pointer_motion` listener, which gets incorrect coordinates during a drag initiated in a sub-surface if the pointer is dragged outside said sub-surface. The coordinates appear to be in relation to the top-level instead. This is not the case in KWin and Weston, the other two compositors I tried. I've also modified the [hello-wayland](https://github.com/emersion/hello-wayland) to exhibit the same behavior without the overhead of FLTK: [modified main.c](https://pastebin.com/HR9TAejn)

<details><summary>modified main.c</summary>

```c
/*#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mman.h>
#include <unistd.h>
#include <wayland-client.h>
#include <wayland-client-protocol.h>
#include <linux/input-event-codes.h>

#include "cat.h"
#include "shm.h"
#include "xdg-shell-client-protocol.h"

static const int width = 128;
static const int height = 128;

static bool configured = false;
static bool running = true;

static struct wl_shm *shm = NULL;
static struct wl_compositor *compositor = NULL;
static struct xdg_wm_base *xdg_wm_base = NULL;

static void *shm_data = NULL;
static struct wl_surface *surface = NULL;
static struct xdg_toplevel *xdg_toplevel = NULL;

static void noop() {
	// This space intentionally left blank
}

static void xdg_wm_base_handle_ping(void *data,
		struct xdg_wm_base *xdg_wm_base, uint32_t serial) {
	// The compositor will send us a ping event to check that we're responsive.
	// We need to send back a pong request immediately.
	xdg_wm_base_pong(xdg_wm_base, serial);
}

static const struct xdg_wm_base_listener xdg_wm_base_listener = {
	.ping = xdg_wm_base_handle_ping,
};

static void xdg_surface_handle_configure(void *data,
		struct xdg_surface *xdg_surface, uint32_t serial) {
	// The compositor configures our surface, acknowledge the configure event
[...]

PR #2547, opened 2025-01-25
cursor.c: use subsurface as reference for out-of-surface movement

Consolatis:
The protocol states that the wl_pointer motion coordinates must be relative to the focused surface (e.g. the surface that last received a wl_pointer enter event).

Before this patch, the coordinates were relative to the toplevel surface instead, resulting in subsurface events having the wrong coordinates when pressing a button over a subsurface and moving the cursor outside of that subsurface.

Fixes:
- #2542

### DIFF
--- a/src/common/surface-helpers.c
+++ b/src/common/surface-helpers.c
@@ -10,7 +10,6 @@ subsurface_parent_layer(struct wlr_surface *wlr_surface)
 	struct wlr_subsurface *subsurface =
 		wlr_subsurface_try_from_wlr_surface(wlr_surface);
 	if (!subsurface) {
-		wlr_log(WLR_DEBUG, "surface %p is not subsurface", subsurface);
 		return NULL;
 	}
 	struct wlr_surface *parent = subsurface->parent;
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -455,9 +455,18 @@ process_cursor_motion_out_of_surface(struct server *server,
 	assert(surface);
 	int lx, ly;
 
-	if (view) {
+	if (node && wlr_subsurface_try_from_wlr_surface(surface)) {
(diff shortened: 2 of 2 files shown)


tokyo4j:
Logic LGTM, but I suggest moving the part that applies the shadow offset for xdg_toplevels like:

```diff
diff --git a/src/input/cursor.c b/src/input/cursor.c
index 5af3bd0e..dc3851e0 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -458,6 +458,14 @@ process_cursor_motion_out_of_surface(struct server *server,
 	if (view) {
 		lx = view->current.x;
 		ly = view->current.y;
+		/* Take into account invisible xdg-shell CSD borders */
+		if (view->type == LAB_XDG_SHELL_VIEW) {
+			struct wlr_box geo;
+			wlr_xdg_surface_get_geometry(
+				xdg_surface_from_view(view), &geo);
+			lx -= geo.x;
+			ly -= geo.y;
+		}
 	} else if (node && wlr_layer_surface_v1_try_from_wlr_surface(surface)) {
 		wlr_scene_node_coords(node, &lx, &ly);
 #if HAVE_XWAYLAND
@@ -471,13 +479,6 @@ process_cursor_motion_out_of_surface(struct server *server,
 
 	*sx = server->seat.cursor->x - lx;
 	*sy = server->seat.cursor->y - ly;
-	/* Take into account invisible xdg-shell CSD borders */
-	if (view && view->type == LAB_XDG_SHELL_VIEW) {
-		struct wlr_box geo;
-		wlr_xdg_surface_get_geometry(xdg_surface_from_view(view), &geo);
-		*sx += geo.x;
-		*sy += geo.y;
-	}
 
 	return true;
 }
```

and removing `is_subsurface`.

Consolatis:
Good point, changed. Thanks for the review.

PR #2550, opened 2025-01-27
wlr-foreign: send `output_enter` on initialization

tokyo4j:
This fixes a bug that `output_enter` events are not sent when a Slack window running in background is re-mapped, which caused missing taskbar items in Waybar when it's configured to show windows per output.


Consolatis:
Thanks. Looks like the underlying issue is that we don't clear view->outputs on unmap. This fix should be fine as well though.

PR #2552, opened 2025-01-29
ssd: eliminate corner extents and correct corner behavior

ahesford:
Handling of SSD corners is currently broken and overly complicated.

See the attached image for a pictorial view of the current handling of window edges:

1. SSD borders are shown in light green (top and bottom) and red (left and right).
2. Invisible extents, which extend the effective border region, are shown in dark blue (top and bottom), dark green (left and right) and cyan (corners).
3. All components are partially transparent, so the overlap of the bottom corners with the bottom and side edges is shown as a more vibrant coloration on the borders. The brighter regions are NOT independent rects. There is no overlap between the top corners and any border segments.

Each segment (four window borders, four single-edge extents, and four corner extents) has a unique enumeration that we use to determine whether mouse resizes act on a single edge or two perpendicular edges simultaneously. The unique corner segments are also mapped to mousebind contexts for each respective corner.

The corners complicate `ssd-extents.c` because we have to compute extra offsets and sizes for all eight components. It isn't obvious how these things will be positioned without forcing them to be rendered visibly. Even worse, this leads to buggy and unexpected behavior:
1. The overlap of the bottom and side borders with the bottom corners *should* trigger corner-based resizing and mousebinds, but the mouse context in any border segment is always just the single-edge border. We don't see a corner context until the mouse is moved outside of the window and its border.
2. With rounded corners in the titlebar, the portion of the visible corner extents that is within the rectangular bounds of window borders but outside of the radius produces a context of `LAB_SSD_PART_TITLEBAR_CORNER_{RIGHT,LEFT}`, which does not match our expected corner contexts. Thus, even when you are outside the visible region of the window near the titlebar, you may not be within what we detect as corners---until you move even further away!

This could be fixed by breaking border rects into segments that map to each of the `LAB_SSD_PART_CORNER_*` enumerations and other segments that map to the single-edge enumerations, but that gets even more complicated. We'd have a bunch of extra sizes and offsets to compute, and we'd have extra logic to handle lower bounds when the window gets smaller than any of the components. We'd also have to break the corner extents into extra pieces, and everything would turn into a mess.

Instead, this PR corrects and improves the behavior while simplifying the implementation:
1. Corner extents disappear entirely.
2. Top and bottom extents run the entire length of the window (and its invisible extended area); left and right extents run the entire visible height of the window with decorations, butting up to the top and bottom extents without overlapping (as we currently do with the bottom and side borders).
3. Corner detection is done in `ssd_resize_edges` by determining if the mouse context is one of the four single-edge zones and, if so, checking whether the cursor is near the ends of that zone. The definition of "near" is now made configurable with the `theme.cornerSize` parameter, meaning people can elminate corners entirely or make them as large as they wish.
4. Mapping of mousebind contexts reuses the corner detection in `ssd_resize_edges`, so we simplify and unify corner handling for both resizes and configurable bindings.
5. `ssd_resize_edges` now treats `LAB_SSD_PART_TITLEBAR_CORNER_*` as a corner resize edge, so we don't get weird deadzones in corner or edge detection where, in the image, we see no top border.

corners

The second image shows the borders and extents after the changes in this PR (ignore the size difference in the extents---that was a consequence of other changes). Note that, even though the titlebar corners do not correspond to any border segment, they now behave as corner resize and action zones.

nocorners


tokyo4j:
I like your approach.

I think `<theme><cornerSize>` is confusing as the new configuration doesn't visually change anything. I suggest changing it to `<theme><cornerSize>` or `<theme><cornerRange>`. My preference is `<theme><cornerRange>`, but I welcome any ideas.

Also, I made a patch to improve this PR:

```diff
diff --git a/src/ssd/ssd.c b/src/ssd/ssd.c
index b952244c..d3657a50 100644
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -146,58 +146,37 @@ uint32_t
 ssd_resize_edges(struct cursor_context *ctx, struct wlr_cursor *cursor)
 {
 	assert(ctx);
+	assert(cursor);
 	struct view *view = ctx->view;
-	uint32_t edge = WLR_EDGE_NONE;
 
-	switch (ctx->type) {
-	case LAB_SSD_PART_TOP:
-		edge = WLR_EDGE_TOP;
-		break;
-	case LAB_SSD_PART_RIGHT:
-		edge = WLR_EDGE_RIGHT;
-		break;
-	case LAB_SSD_PART_BOTTOM:
-		edge = WLR_EDGE_BOTTOM;
-		break;
-	case LAB_SSD_PART_LEFT:
-		edge = WLR_EDGE_LEFT;
-		break;
-	case LAB_SSD_PART_CORNER_TOP_LEFT:
-	case LAB_SSD_PART_TITLEBAR_CORNER_LEFT:
-		return WLR_EDGE_TOP | WLR_EDGE_LEFT;
-	case LAB_SSD_PART_CORNER_TOP_RIGHT:
-	case LAB_SSD_PART_TITLEBAR_CORNER_RIGHT:
-		return WLR_EDGE_RIGHT | WLR_EDGE_TOP;
-	case LAB_SSD_PART_CORNER_BOTTOM_RIGHT:
-		return WLR_EDGE_BOTTOM | WLR_EDGE_RIGHT;
-	case LAB_SSD_PART_CORNER_BOTTOM_LEFT:
-		return WLR_EDGE_BOTTOM | WLR_EDGE_LEFT;
-	default:
+	if (!view || !view->ssd_enabled || view->fullscreen) {
 		return WLR_EDGE_NONE;
 	}
 
-	if (!view || !cursor || !view->ssd) {
-		return edge;
+	struct wlr_box view_box = view->current;
+	view_box.height = view_effective_height(view, /* use_pending */ false);
+	if (view->ssd_enabled && !view->fullscreen && !view->ssd_titlebar_hidden) {
+		int titlebar_height = view->server->theme->titlebar_height;
+		view_box.y -= titlebar_height;
+		view_box.height += titlebar_height;
+	}
+	if (wlr_box_contains_point(&view_box, cursor->x, cursor->y)) {
+		return WLR_EDGE_NONE;
 	}
 
-	if (edge == WLR_EDGE_BOTTOM || edge == WLR_EDGE_TOP) {
-		int width = view->current.width;
-		int corner_size = MAX(0, MIN(rc.corner_size, width / 2));
+	uint32_t edge = WLR_EDGE_NONE;
+	int corner_height = MAX(0, MIN(rc.corner_size, view_box.height / 2));
+	int corner_width = MAX(0, MIN(rc.corner_size, view_box.width / 2));
 
-		if (cursor->x < view->current.x + corner_size) {
-			edge |= WLR_EDGE_LEFT;
-		} else if (cursor->x > view->current.x + width - corner_size) {
-			edge |= WLR_EDGE_RIGHT;
-		}
-	} else if (edge == WLR_EDGE_LEFT || edge == WLR_EDGE_RIGHT) {
-		int height = view_effective_height(view, /* use_pending */ false);
-		int corner_size = MAX(0, MIN(rc.corner_size, height / 2));
-
-		if (cursor->y < view->current.y + corner_size) {
-			edge |= WLR_EDGE_TOP;
-		} else if (cursor->y > view->current.y + height - corner_size) {
-			edge |= WLR_EDGE_BOTTOM;
-		}
+	if (cursor->x < view_box.x + corner_width) {
+		edge |= WLR_EDGE_LEFT;
+	} else if (cursor->x > view_box.x + view_box.width - corner_width) {
+		edge |= WLR_EDGE_RIGHT;
+	}
+	if (cursor->y < view_box.y + corner_height) {
+		edge |= WLR_EDGE_TOP;
+	} else if (cursor->y > view_box.y + view_box.height - corner_height) {
+		edge |= WLR_EDGE_BOTTOM;
 	}
 
 	return edge;
@@ -346,29 +325,31 @@ ssd_part_contains(enum ssd_part_type whole,
 {
 	assert(ctx);
 	enum ssd_part_type candidate = ctx->type;
-	if (whole == candidate || whole == LAB_SSD_ALL) {
+
+	switch (whole) {
+	case LAB_SSD_ALL:
 		return true;
-	}
-	if (whole == LAB_SSD_BUTTON) {
+	case LAB_SSD_BUTTON:
 		return candidate >= LAB_SSD_BUTTON_CLOSE
 			&& candidate <= LAB_SSD_BUTTON_OMNIPRESENT;
-	}
-	if (whole == LAB_SSD_PART_TITLEBAR) {
+	case LAB_SSD_PART_TITLEBAR:
 		return candidate >= LAB_SSD_BUTTON_CLOSE
 			&& candidate <= LAB_SSD_PART_TITLE;
-	}
-	if (whole == LAB_SSD_PART_TITLE) {
+	case LAB_SSD_PART_TITLE:
 		/* "Title" includes blank areas of "Titlebar" as well */
 		return candidate >= LAB_SSD_PART_TITLEBAR
 			&& candidate <= LAB_SSD_PART_TITLE;
-	}
-	if (whole == LAB_SSD_FRAME) {
+	case LAB_SSD_FRAME:
 		return candidate >= LAB_SSD_BUTTON_CLOSE
 			&& candidate <= LAB_SSD_CLIENT;
-	}
-
-	/* Perform corner tests if necessary */
-	switch (whole) {
+	case LAB_SSD_PART_TOP:
+		return ssd_resize_edges(ctx, cursor) == WLR_EDGE_TOP;
+	case LAB_SSD_PART_BOTTOM:
+		return ssd_resize_edges(ctx, cursor) == WLR_EDGE_BOTTOM;
+	case LAB_SSD_PART_LEFT:
+		return ssd_resize_edges(ctx, cursor) == WLR_EDGE_LEFT;
+	case LAB_SSD_PART_RIGHT:
+		return ssd_resize_edges(ctx, cursor) == WLR_EDGE_RIGHT;
 	case LAB_SSD_PART_CORNER_TOP_LEFT:
 		return ssd_resize_edges(ctx, cursor) == (WLR_EDGE_TOP | WLR_EDGE_LEFT);
 	case LAB_SSD_PART_CORNER_TOP_RIGHT:
@@ -378,10 +359,8 @@ ssd_part_contains(enum ssd_part_type whole,
 	case LAB_SSD_PART_CORNER_BOTTOM_RIGHT:
 		return ssd_resize_edges(ctx, cursor) == (WLR_EDGE_BOTTOM | WLR_EDGE_RIGHT);
 	default:
-		break;
+		return whole == candidate;
 	}
-
-	return false;
 }
 
 enum ssd_mode
```

</details>

This patch does:
- Call `ssd_resize_edges()` in `ssd_part_contains()` also when `LAB_SSD_PART_{TOP,BOTTOM,LEFT,RIGHT}` is given as `whole`. This makes it clear that `ssd_resize_edges()` always determines the direction of resize and I think this is easier to comprehend. This is a bit breaking though, because `Left` context no longer include `TLCorner`. I slightly prefer the change in my patch as it follows Openbox's behavior and the code looks slightly nicer, but we can keep the existing behavior by replacing `==` with `&`.
- Determine the direction of resize only based on the window geometry and cursor position in `ssd_resize_edges()`. I think this is easier to follow as it we no longer have to care `ctx->type`. This also prevents the bug fixed by #2105 to be reintroduced.

tokyo4j:
I opened #2553 that supersedes this PR as I realized that we can modify `ssd_get_part_type()` rather than `ssd_part_contains()` and `ssd_resize_edges()` to achieve the same goal. I think performing the calculation when setting `cursor_context->type` in `get_cursor_context()` and keeping `ssd_part_contains()` and `ssd_resize_edges()` "pure" makes the code easier to follow.

Consolatis:
> I think `<theme><cornerSize>` is confusing as the new configuration doesn't visually change anything. I suggest changing it to `<theme><cornerSize>` or `<theme><cornerRange>`. My preference is `<theme><cornerRange>`, but I welcome any ideas.

I assume the suggestions should be `<resize><cornerSize>` / `<resize><cornerRange>` based on the new PR?
Personally I like putting it at `<resize>` rather than `<theme>`. What about about simply `<resize><range>` as the setting is not actually related to corners only?

Consolatis:
> What about about simply `<resize><range>` as the setting is not actually for corners only?

Sorry, confused this with #2529. Why do we need a configuration option at all here? Can't we just use something like button_width / 2 for x and button height / 2 for y?

ahesford:
I want it configurable because I want large corner targets, bigger than the buttons. Usually I'm not interested in diagonal resizing than single-edge sizes.

tokyo4j:
> > I think `<theme><cornerSize>` is confusing as the new configuration doesn't visually change anything. I suggest changing it to `<theme><cornerSize>` or `<theme><cornerRange>`. My preference is `<theme><cornerRange>`, but I welcome any ideas.
> 
> I assume the suggestions should be `<resize><cornerSize>` / `<resize><cornerRange>` based on the new PR? Personally I like putting it at `<resize>` rather than `<theme>`. What about about simply `<resize><range>` as the setting is not actually for corners only?

Sorry they are typos :sweat_smile: I intended to suggest `<resize><cornerRange>`/ `<resize><cornerSize>`. Sorry for confusion.

swmaazx:
``

PR #2558, opened 2025-02-03
Add helper to extract refs from `NEWS.md`

johanmalm:
@Consolatis We talked some time ago about doing this. Have eventually got around to it :smile: Thought I'd better do it before starting the `NEWS.md` update for `0.8.3` to safe myself some time.

### DIFF
--- a/scripts/rip.c
+++ b/scripts/rip.c
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Helper to extract references from NEWS.md and print them as github URLs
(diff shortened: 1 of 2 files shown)


scripts/rip.c
```diff
@@ -0,0 +1,81 @@
...
+			return false;
+		}
+	}
+	return true;
+}
+
+void process_line(const char *line)
+{
+	const char *p = line;
+	while (p = strchr(p, '[')) {
```
Consolatis, line 43: Suggest adding an extra pair of parentheses in the while assignment.

Consolatis:
Apart from the nitpick above this LGTM.

Issue #2539, opened 2025-01-24
use-after-free on shutdown when running nested with WLR_WL_OUTPUTS=2


Consolatis:
### Problem description

When `Ctrl-C`ing a nested instance with more than a single virtual output, labwc does run into a UAF while trying to set a cursor image in the output_layout_changed handler. Didn't yet investigate what causes that to happen, I think this was not the case in the labwc 0.8 branch with earlier versions.

Note: when closing the virtual output windows manually before doing a `ctrl-c` everything is fine.

<details><summary>stacktrace</summary>

```
=================================================================
==3924482==ERROR: AddressSanitizer: heap-use-after-free on address 0xffff9b607cb0 at pc 0xaaaac3148e00 bp 0xfffffa3b57c0 sp 0xfffffa3b57d8
READ of size 8 at 0xffff9b607cb0 thread T0
    #0 0xaaaac3148dfc in wlr_texture_from_buffer ../subprojects/wlroots/render/wlr_texture.c:116
    #1 0xaaaac340e928 in wlr_output_cursor_set_buffer ../subprojects/wlroots/types/output/cursor.c:340
    #2 0xaaaac32df15c in output_cursor_set_xcursor_image ../subprojects/wlroots/types/wlr_cursor.c:478
    #3 0xaaaac32e1384 in cursor_output_cursor_update ../subprojects/wlroots/types/wlr_cursor.c:583
    #4 0xaaaac32e1ab0 in cursor_update_outputs ../subprojects/wlroots/types/wlr_cursor.c:610
    #5 0xaaaac32e2088 in wlr_cursor_set_xcursor ../subprojects/wlroots/types/wlr_cursor.c:627
    #6 0xaaaac30be74c in cursor_update_image ../src/input/cursor.c:417
    #7 0xaaaac2fd6d08 in output_update_for_layout_change ../src/output.c:585
    #8 0xaaaac2fd9b80 in do_output_layout_change ../src/output.c:837
    #9 0xaaaac2fd9e74 in handle_output_layout_change ../src/output.c:853
    #10 0xffffa83d97a0 in wl_signal_emit_mutable (/lib/aarch64-linux-gnu/libwayland-server.so.0+0x97a0)
    #11 0xaaaac3329d64 in output_layout_reconfigure ../subprojects/wlroots/types/wlr_output_layout.c:115
    #12 0xaaaac332a390 in addon_destroy ../subprojects/wlroots/types/wlr_output_layout.c:147
[...]

PR #2560, opened 2025-02-04
server: on exit, destroy backend first

Consolatis:
This fixes a UAF caused by trying to update the xcursor in a output destroy handler, in turn caused by destroying the backend.

Fixes:
- #2539


johanmalm:
LGTM. Thanks.

PR #2572, opened 2025-02-09
Protocols: add generic output tracker

Consolatis:
Tracking outputs for protocol implementations is a reoccurring thing to do. The main complexity is that we can only send events for outputs actually bound by the client, so rather than just sending the event for all clients using the given client protocol + have the respective output bound we also need to listen for future output bind signals by clients and then send the `output_enter` event at that point again. This is a generic restriction of various wayland protocols and so far no conclusion could be reached upstream to enhance the situation on the protocol level.

We are currently implementing the logic for this to work as good as possible (given the generic protocol restriction) in the ext-workspace and cosmic-workspace implementations. I am also working on protocol implementations for the not-yet-merged ext-foreign-toplevel-info protocol which also requires output tracking. As I am lazy I don't want to implement the same thing a third time so this PR abstracts that logic into an own protocol helper and uses that one for the existing ext-workspace and cosmic-workspace implementations (and in future for the ext-foreign-toplevel-info one). And while at it also gets rid of the internal headers for both implementations as its not required anymore after abstracting the output tracking logic away.

For after 0.8.3.


johanmalm:
I like the look of this. Not been through it in any detail though.

tokyo4j:
I haven't looked into this PR in detail, but I'd like to express my opinion.

> Additionally, I am working on protocol implementations for the not-yet-merged ext-foreign-toplevel-info protocol which also requires output tracking.

I think we should be really careful when implementing a protocol on our own.

In my opinion, `cosmic-workspace` was an exceptional one which was worth implementing, because (1) workspace management on client side was a pretty basic functionality that most X11 desktop environments have provided via their panel and (2) the protocol had been actually proved to be working by cosmic-comp and cosmic-panel. But now that ext-workspace protocol has been merged in wayland-protocols, I think we can drop `cosmic-workspace` support in some release cycles (probably when the wlroots implements ext-workspace support?) so that we go back to follow our original policy:

> We think that standardisation and de-fragmentation is a route to greater Wayland adoption, and wanting to play our part in this, Labwc only understands [wayland-protocols](https://gitlab.freedesktop.org/wayland/wayland-protocols) & [wlr-protocols](https://gitlab.freedesktop.org/wlroots/wlr-protocols), and it cannot be controlled with dbus, sway/i3/custom-IPC or other technology.

I will never exclude the option to implement `ext-foreign-toplevel-state` protocol as it will be a really important protocol for us, too (e.g. `urgent` state I suggested there is essential for us to enforce more strict xdg-activation token validation). However, if we do that, we will need to (1) add `lab-foreign-toplevel-state` protocol and let clients use that protocol, or (2) implement the work-in-progress `ext-foreign-toplevel-state` and tolerate breakages in clients who use another snapshot of the protocol. I honestly don't want to see them to happen...

So, for now, I would rather like to see draft PRs in both wlroots & labwc that implement the ext-foreign-toplevel and hopefully help the upstream discussion.

Consolatis:
> I think we should be really careful when implementing a protocol on our own.

Absolutely agree. This is for implementing the upstream protocol and will only be merged once https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196 is accepted.

tokyo4j:
> Absolutely agree. This is for implementing the upstream protocol and will only be merged once https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196 is accepted.

So do you want to support the protocol on labwc side before wlroots adds an implementation for it? My slight preference is waiting for a major wlroots update which implements the protocol rather than duplicating the effort, but yeah I think we can tolerate the maintenance cost as we can switch to the wlroots implementation once it's available.

Consolatis:
> So do you want to support the protocol on labwc side before wlroots adds an implementation for it? My slight preference is waiting for a major wlroots update which implements the protocol

Yes, that was the idea.

> rather than duplicating the effort, 

It is written in such a way that it only requires minimal changes to make it work in wlroots itself (I think its mostly `znew()` -> `calloc()`). Same is true for the ext-workspace protocol but I am slightly procrastinating with proposing that one for wlroots.

src/protocols/output-tracker.c
```diff
@@ -0,0 +1,175 @@
...
+	const struct output_tracker_impl *impl;
+
+	struct {
+		struct wl_listener output_bind;
+		struct wl_listener output_destroy;
+	} on;
+	struct wl_list link;
+};
+
+static struct wl_list objects = WL_LIST_INIT(&objects);
```
tokyo4j, line 21: I suggest renaming this to `object_outputs`.

tokyo4j:
The most confusing part for me in this PR is that there's no `struct output_tracker`. I think using only either of `output_tracker` or `output_object` will make the code easier to follow.

Consolatis:
> The most confusing part for me in this PR is that there's no `struct output_tracker`. I think using only either of `output_tracker` or `output_object` (I prefer `output_object`) will make the code easier to follow.

This helper started its life as `object-output.c` but I ended up not liking the name so I renamed the public interfaces to `output-tracker`. Would renaming the internal `struct object_output` to `struct output_tracker_output` help to reduce the confusion?

Looking at the diff again it also needs
- [ ] comments about what it is and how to use the functions in the header
- [ ] renaming `output_tracker_destroy()` to something like `output_tracker_finish_object()` or `output_tracker_stop_tracking(void *object)`

The destroy/finish function is required because the earlier protocol specific implementations provided a destroy signal but due to the new implementation being protocol agnostic (and thus operating on a generic \*void pointer) we can't really use any kind of protocol internal destroy signals.

tokyo4j:
How about this? https://github.com/Consolatis/labwc/compare/feature/protocols_generic_output_tracking...tokyo4j:labwc:output-tracker

Rather than storing the states in the global list, I think explicitly adding `struct output_tracker` in `lab_{ext,cosmic}_workspace_group` and storing the entered outputs there makes the data flow clearer.

PR #2574, opened 2025-02-10
src/overlay.c: clean up timer on exit

Consolatis:
Prevents a leak notification on exit. I think we can merge this within the cooldown period as side effects are unlikely.
CC @tokyo4j


tokyo4j:
LGTM. Thanks.

Issue #2576, opened 2025-02-12
wlopm --on fails to wake up display from screen blank after changing video mode with wlr-randr --custom-mode


qrp73:
### Problem description

I'm using RPI4 with raspi-os and labwc, with this labwc/autostart config for display blank:
```
/usr/bin/swayidle -w timeout 300 'wlopm --off \*' resume 'wlopm --on \*' &
```

Most of time it works ok, but once I change video mode using ```wlr-randr```  with --custom-mode, like this:
```
wlr-randr --output HDMI-A-1 --custom-mode 1280x1024@50Hz
```

it works until first display blank and unable to wake up anymore.

When this happens, it is possible to wake up the display by changing the display mode again using wlr-randr from remote ssh with no --custom-mode.

Further investigation shows that this issue happens when you set any --custom-mode. 

I found 2 stable reproduced issues:
**1) Custom Display Mode Activation**: When switching to a display mode not included in the standard list, the wlopm --on command is failed. Additionally, subsequent attempts to reapply the same mode is failed for an unknown reason. Furthermore, wlopm --on only starts functioning correctly after switching to a mode listed in the standard modes. With any custom mode, even very slightly different from standard one, wlopm --on fails.

**2) Swayidle Timer Reset** (moved to separate issue #2579): Upon resuming the display from sleep mode with a display mode change from ssh, the swayidle timer does not restart. As a result, the system does not enter sleep mode again anymore. Manual interaction, such as moving the mouse or pressing a key, is required to reactivate the timer.

Also sometimes, very-very rare I catch issue when ```wlopm --on``` fails even for standard display mode, but it happens very-very rare, its almost impossible to catch and very hard to reproduce, but sometimes it happens.

See original [issue #321 on raspberrypi/bookworm-feedback](https://github.com/raspberrypi/bookworm-feedback/issues/321) issue tracker.

### Steps to reproduce

1) open terminal and run ```swayidle -w timeout 10 'wlopm --off \*' resume 'wlopm --on \*'```, it should put display into sleep mode after 10 seconds idle.

2) change display mode to any mode not included in standard list with using command:
```
wlr-randr --output HDMI-A-1 --custom-mode 1280x1024@50Hz
```

where ```HDMI-A-1``` is your display id, ```1280x1024``` is your display resolution and ```50Hz``` is your display refresh rate. 

Just slightly change the display refresh rate so that it is still accepted by your display, but the mode will not be listed in the standard mode list.

[...]

PR #2578, opened 2025-02-12
output: set custom output mode on wlopm --on

Consolatis:
This works around a wlroots issue: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3946

Fixes:
- #2576

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -1073,6 +1073,28 @@ handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
 		break;
 	case ZWLR_OUTPUT_POWER_V1_MODE_ON:
 		wlr_output_state_set_enabled(&output->pending, true);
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. Wow. That's some fix! LGTM. Are we considering this for `0.8.3`?

Consolatis:
> Are we considering this for `0.8.3`?

Yes, once confirmed that it fixes the issue I'd say lets go with it. The amount of users potentially hitting this is pretty low (as it requires the user to a) run on the DRM backend and b) use a custom resolution rather than what their monitor offers). But it feels better to me to fix that so those people don't end up with an basically unusable labwc session after swayidle kicks in.

cillian64:
LGTM, and I've confirmed the latest commit fixes the issue for me.

Issue #2579, opened 2025-02-13
Swayidle timer reset issue after display mode switch


qrp73:
### Problem description

Upon resuming the display from sleep mode after a display mode change from ssh session, the swayidle timer does not restart. As a result, the system does not enter sleep mode again anymore. Manual interaction, such as moving the mouse or pressing a key, is required to reactivate the timer.

### Steps to reproduce

1. open terminal and start swayidle to put display into sleep mode after 10 seconds idle: ```swayidle -w timeout 10 'wlopm --off \*' resume 'wlopm --on \*'```
2. wait until display going to sleep mode
3. connect through ssh and try to change display mode with wlr-randr. For example ```wlr-randr --output HDMI-A-1 --custom-mode 1280x1024@50Hz```, you can use any other display mode which is supported by your display. It leads to display wake up.

Expected result: after 10 seconds the display going to sleep mode again

Actual result: if you don't touch keyboard or mouse, the display don't going to sleep mode again.
 

### labwc build source

Release

### labwc version

0.8.1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Raspberry Pi OS Bookworm: Linux raspi 6.6.74+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.74-1+rpt1 (2025-01-27) aarch64 GNU/Linux

cillian64:
Thanks for making the separate issue for this.

I've just done some testing:

- If I start swayidle, then do the custom mode change (from SSH) while the display is still awake, don't touch the mouse/keyboard, then it still successfully goes to sleep once the original timeout is hit.
- If I do the custom mode change while the display is asleep, this wakes up the display. With no mouse or keyboard input, it never goes back to sleep.  After touching the mouse or pressing a key there is a brief black flash but the monitor stays on.  Then after 10s it goes to sleep.

So I suspect the cause is quite simple, mode changes re-enable the output but don't boop the idle activity thing. So swayidle doesn't know the display was powered back on and doesn't know to restart its idle timer.

The black flash is caused by doing `wlopm --on` on a display which is already on.  This probably counts as a separate bug (every frame is perfect?) but I don't know if it's worth pursuing.  You can reproduce it without mode changes or swayidle by just running `wlopm --on HDMI-A-1` repeatedly.

Consolatis:
In general I really suggest not mixing wlopm and wlr-randr usage. They are using different protocols for different reasons. With that said, labwc should deal better with the situation.

[...]

PR #2580, opened 2025-02-13
output: early return on wlopm if requested state is already set

Consolatis:
This prevents a unnecessary output commit without buffer, showing up as black flash when repeatedly calling wlopm --on.

---
Reported by @cillian64 in #2579.

Should be fine to merge within the cooldown period as chances of this having side effects is minimal.


cillian64:
Confirmed that this fixes the black flash.

johanmalm:
Thanks both. LGTM.

PR #2583, opened 2025-02-16
osd: prevent preview outlines from overlapping OSD in first output

tokyo4j:
See https://github.com/labwc/labwc/issues/2465#issuecomment-2661511564.

Honestly I'm not 100% sure this PR doesn't introduce another regression... but I believe the issue is something we should fix before 0.8.3.

Testing is appreciated.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -512,20 +512,22 @@ new_output_notify(struct wl_listener *listener, void *data)
 	 * bottom):
 	 *	- session lock layer
 	 *	- window switcher osd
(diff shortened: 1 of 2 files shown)


johanmalm:
Thanks for tracking this down. The logic seems fine to me. Will do some testing later and let you know.

johanmalm:
Have tested and can confirm that the bug is fixed. Cannot detect any unintended consequences. 
LGTM

Issue #2582, opened 2025-02-16
Switch to window after iconifying it does non display it


gtellalov:
### Problem description

When iconifying a maximised window it can no longer be displayed even though when switching to it, it gets focused.

### Steps to reproduce

1. Maximise a window
2. Iconify it
3. Switch to it with NextWindow or PreviousWindow

### labwc build source

Local build

### labwc version

labwc 0.8.2-125-ga4d1fcfe

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

johanmalm:
Works fine here. What application does it happen with?

domo141:
hmm, happens to me too

2 urxvt windows, one maximized, then that iconofied - alt-tab shows it, but after releasing, that maximized window disappears..

domo141:
eventually I got that one back to view with sfwbar...

tokyo4j:
I can't reproduce it... Does it happen when you start labwc with the default config like `labwc -C /tmp -s urxvt`? Also, did you maximize/iconify the window with the titlebar buttons?

gtellalov:
No, it doesn't happen with the default config. My config has titlebar disabled, but I could reproduce it both form a keyboard shortcut and the window menu (ShowMenu). 
Here's my config in case it helps reproduce it: http://0x0.st/8NIw.xml

domo141:
actually, in my case, the window does not need to be maximized... i'll explore more

domo141:
now I have 2 "normal" urxvt windows, both minimized, neither gets back. I had to skip the view first
time in the "switcher" alt-tab popped up (i.e. choose other view). If I chose the latest minimized view
immediately in the window switcher then the view came back to be useable. Keyboard focus were given
to the view that is invisible, though (I have focus-follow mouse, but mouse not over any of the othet views...)

domo141:
```
   <keybind key="A-Tab">
      <action name="NextWindow" />
    </keybind>
```

domo141:
I can reproduce in nested labwc. there I also tried w/ fedora-41 -shipped labwc 0.8.2 (where deiconifying works)
In 0.8.2 windowSwitcher shows small 'm' for minimized views -- latest github master the space is empty (i.e. no 'm')

tokyo4j:
[...]

PR #2584, opened 2025-02-17
cursor: don't un-minimize previewed window while window switching

tokyo4j:
This PR fixes the issue reported by @domo141 in #2582, by restoring the check I removed in #2476.

Without the check, if `followMouse` is yes and `followMouseRequiresMovement` is no, `osd_update()` => `cursor_update_focus()` => `desktop_focus_view()` unexpectedly un-minimizes the window on cursor even when the window is just a preview of window switcher. This caused some strange behavior that a minimized window selected with window switcher is immediately hidden after finishing window switching (please ignore the preview outline overlapping OSD):

https://github.com/user-attachments/assets/6ae8a5b5-23c1-4eb1-88c8-48f801dc78bb

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -641,10 +641,15 @@ _cursor_update_focus(struct server *server)
 	struct cursor_context ctx = get_cursor_context(server);
 
 	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse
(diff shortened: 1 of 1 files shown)


tokyo4j:
Side note: I think we can remove `cursor_update_focus()` in `osd_update()` as now we always set the default cursor while window switching since #2455. But it can be done after Friday anyway.

PR #2588, opened 2025-02-19
action: ignore Focus action while window switching

### LINKED ISSUE #2582: Switch to window after iconifying it does non display it

tokyo4j:
Fixes #2582.

We could check against `server->input_mode` inside `desktop_focus_view()`, but I think this is the most predictable and least impactful approach that we can merge before 0.8.3.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -1042,7 +1042,12 @@ actions_run(struct view *activator, struct server *server,
 			}
 			break;
 		case ACTION_TYPE_FOCUS:
-			if (view) {
+			if (view && server->input_mode
+					!= LAB_INPUT_STATE_WINDOW_SWITCHER) {
+				/*
+				 * TODO: check against server->input_mode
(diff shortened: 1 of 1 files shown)


tokyo4j:
I turned this into a draft because I'd like to sort out the current situation first (https://github.com/labwc/labwc/issues/2582#issuecomment-2669358695).

tokyo4j:
We concluded in IRC that we should deal with the NextWindow+Focus issue after 0.8.3 release.

PR #2596, opened 2025-02-22
project wide: clean up event listeners on shutdown

Consolatis:
This ensures all event listeners are removed before the emitting wlroots object is being destroyed. This will be enforced with asserts in wlroots 0.19 but there is no reason to not do it right now either.

This change in wlroots 0.19 is implemented via commit 8f56f7ca43257cc05c7c4eb57a0f541e05cf9a79
"Assert (almost all) signals have no attached listeners on destroy"

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -752,12 +752,26 @@ server_finish(struct server *server)
 		wl_event_source_remove(sighup_source);
 	}
 	wl_display_destroy_clients(server->wl_display);
+
+	seat_finish(server);
+	output_finish(server);
+	xdg_shell_finish(server);
+	layers_finish(server);
+	kde_server_decoration_finish(server);
+	xdg_server_decoration_finish(server);
+	wl_list_remove(&server->new_constraint.link);
+	wl_list_remove(&server->output_power_manager_set_mode.link);
+	wl_list_remove(&server->tearing_new_object.link);
+
 	wlr_backend_destroy(server->backend);
 	wlr_allocator_destroy(server->allocator);
+
+	wl_list_remove(&server->renderer_lost.link);
 	wlr_renderer_destroy(server->renderer);
-	seat_finish(server);
+
 	workspaces_destroy(server);
 	wlr_scene_node_destroy(&server->scene->tree.node);
+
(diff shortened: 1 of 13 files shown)


tokyo4j:
I slightly prefer synchronizing the lifetime of the listeners and the globals by listening for globals' destroy signals (e.g. `wlr_server_decoration_manager.events.destroy`). But other devs may have different views.

tokyo4j:
I found following signals that are not cleaned up:
- `server->drm_lease_request`
- `workspace->on_cosmic.activate`
- workspace->on_ext.activate

Also, can we remove this comment?

https://github.com/labwc/labwc/blob/392db7c04dd30d7137b3c22f32021bcd1baeb984/src/input/cursor.c#L1514

LGTM other than that.

johanmalm:
@tokyo4j LGTM too. Do you fancy doing a separate PR for those three listeners you found and I merge this one as is?

By the way, how did you find those listeners? Do you use some static/dynamic analyzer or was it just by review?

tokyo4j:
> Do you fancy doing a separate PR for those three listeners you found and I merge this one as is?

Sure. I'll open a follow-up PR later.

> By the way, how did you find those listeners? Do you use some static/dynamic analyzer or was it just by review?

By hand. I searched for `wl_signal_add` with VSCode's project-wide search and used Ctrt-F to find corresponding `wl_list_remove`.

src/server.c
```diff
@@ -752,12 +752,26 @@ server_finish(struct server *server)
...
+	wl_list_remove(&server->new_constraint.link);
+	wl_list_remove(&server->output_power_manager_set_mode.link);
+	wl_list_remove(&server->tearing_new_object.link);
+
 	wlr_backend_destroy(server->backend);
 	wlr_allocator_destroy(server->allocator);
+
+	wl_list_remove(&server->renderer_lost.link);
 	wlr_renderer_destroy(server->renderer);
-	seat_finish(server);
```
tokyo4j, line 758: @Consolatis Why did you move the position of `seat_finish()`? This question is not blocking this PR from merging, but I just want to know for my curiosity.

Issue #2591, opened 2025-02-21
Magnifier is misaligned and presents graphical artifacts when used in a mixed-orientation multi-monitor setup


helpvisa:
### Problem description

Attempting to use the magnifier in an environment that combines multiple monitors with different orientations (i.e. horizontal + vertical monitor setup) results in a misaligned magnifier on 'secondary' monitor that leaves behind  a trail of graphical artifacts.

https://github.com/user-attachments/assets/7242a1ea-fe4f-4f6f-a196-df64789d266b

### Steps to reproduce

1. Launch labwc and use wlr-randr to configure a mixed-orientation multi-monitor setup.
2. Enable the magnifier on either display.
3. Observe the magnifier appear in an incorrect position when the mouse moves between the 'primary' horizontal and 'secondary' vertical monitor, leaving a trail behind it.

### labwc build source

Release

### labwc version

0.8.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Fedora 41

Consolatis:
Thanks for bringing this up, the magnifier indeed has an issue with rotated displays.

helpvisa:
Have been playing with this again and using a patch of the current diffs at #2598 (which I know is not intended to solve this particular bug yet) and one curiosity I've noticed that may be worth mentioning is that the area of the screen marked damaged seems to be correct, it's just the magnifier box that is misaligned.

I experimented with hard-coding the transformation of my vertical monitor on the mag_box coordinates, and it indeed made the magnifier show up centered around my cursor, however now the damage_box was in the incorrect spot just as the mag_box itself was previously (this makes sense given the damage_box is derived from the mag_box coordinates). I'm not sure why the cursor coordinates / damage_box coordinates would be correct where the mag_box coordinates would be incorrect if they are derived from the same coordinates in code.

I can't say that I have a full understanding of this code or the wlroots code, but I figured it was worth mentioning in this issue since it's tagged as 'investigation required'. Could it be possible that the cursor coordinates are correct, but somehow the texture / buffer from which the mag_box is being sampled from / rendered to is not transformed correctly and this is what causes the misalignment?

tokyo4j:
Looks like `damage_box`, which is in turn passed as `region` in `scene_output_damage()`, should be in the "transformed" coordinates (not the coordinates in the raw buffer), whereas `mag_box` is in un-transformed coordinates.

So the following patch fixes the artifact:

```diff
[...]

Issue #1836, opened 2024-05-21
magnifier: improve performance by preventing useless rendering when enabled


Consolatis:
We currently add damage after we modified the frame buffer. This causes a new render on the next frame event even if nothing has been changed because the damage region is not empty. We could instead add the last magnifier frame damage to the output damage ring before the early out check and only do so based on
- damage ring currently non-empty
- or cursor moved

What complicates this approach is that the damage is output coordinate specific so we need to either store it somewhere per output temporarily or we create some `magnifier_get_damage_box(struct output *output)` as the region of the magnifier is always the same per output. When we go the 2nd route (which would be my preference) we also need to make sure that we properly handle `Reconfigure` which might cause a different damage region based on new user settings for the magnifier size or border.

https://github.com/labwc/labwc/blob/48742163fd0c1626897b607b3c8adfa7f42093da/src/common/scene-helpers.c#L51-L56

See also
- #1833

droc12345:
2nd approach would be my choice too.

PR #2598, opened 2025-02-23
magnifier: refactor

tokyo4j:
...in preparation for fixing #2591.

- Functions defined in `magnifier.c` are renamed to have the prefix `magnifier_`.
- Use `wlr_output.swapchain.format` to obtain buffer format.
- Make sure magnifier scale never goes below 1.0. This should remove the need for `CLAMP()` macro defined in `magnifier.c`.
- Refactor `magnifier_draw()` (previously `magnify()`) to make it easier to follow.

### DIFF
--- a/src/magnifier.c
+++ b/src/magnifier.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
 
 #include <assert.h>
+#include <wlr/render/swapchain.h>
 #include <wlr/types/wlr_output.h>
 #include "common/macros.h"
 #include "labwc.h"
@@ -24,21 +25,6 @@ magnifier_draw(struct output *output, struct wlr_buffer *output_buffer, struct w
 	struct wlr_box border_box, dst_box;
 	struct wlr_fbox src_box;
 	bool fullscreen = false;
-
-	/* TODO: This looks way too complicated to just get the used format */
-	struct wlr_drm_format wlr_drm_format = {0};
-	struct wlr_shm_attributes shm_attribs = {0};
-	struct wlr_dmabuf_attributes dma_attribs = {0};
-	if (wlr_buffer_get_dmabuf(output_buffer, &dma_attribs)) {
-		wlr_drm_format.format = dma_attribs.format;
-		wlr_drm_format.len = 1;
-		wlr_drm_format.modifiers = &dma_attribs.modifier;
-	} else if (wlr_buffer_get_shm(output_buffer, &shm_attribs)) {
-		wlr_drm_format.format = shm_attribs.format;
-	} else {
-		wlr_log(WLR_ERROR, "Failed to read buffer format");
-		return;
-	}
 
 	/* Fetch scale-adjusted cursor coordinates */
 	struct server *server = output->server;
@@ -105,7 +91,8 @@ magnifier_draw(struct output *output, struct wlr_buffer *output_buffer, struct w
 	}
 	if (!tmp_buffer) {
 		tmp_buffer = wlr_allocator_create_buffer(
-			server->allocator, width, height, &wlr_drm_format);
+			server->allocator, width, height,
+			&output->wlr_output->swapchain->format);
 	}
 	if (!tmp_buffer) {
 		wlr_log(WLR_ERROR, "Failed to allocate temporary magnifier buffer");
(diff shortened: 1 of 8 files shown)


src/magnifier.c
```diff
@@ -180,37 +154,30 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box
...
-	dst_box.height = height;
+	struct wlr_fbox src_box_for_paste;
+	src_box_for_paste.width = mag_box.width / mag_scale;
+	src_box_for_paste.height = mag_box.height / mag_scale;
 
 	if (fullscreen) {
-		src_box.x = CLAMP(ox - (ox / mag_scale), 0.0,
-			width * (mag_scale - 1.0) / mag_scale);
-		src_box.y = CLAMP(oy - (oy / mag_scale), 0.0,
-			height * (mag_scale - 1.0) / mag_scale);
```
tokyo4j, line 197: I assumed `CLAMP()` is not needed if `mag_scale` never becomes smaller than `1.0`. CC @Consolatis

src/magnifier.c
```diff
@@ -64,38 +49,20 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box
...
+	struct wlr_box mag_box;
 	if (fullscreen) {
-		width = output_buffer->width;
-		height = output_buffer->height;
-		x = 0;
-		y = 0;
+		mag_box = output_box;
 	} else {
-		width = rc.mag_width + 1;
-		height = rc.mag_height + 1;
```
tokyo4j, line 74: I don't get what these `+ 1` were for.

src/magnifier.c
```diff
@@ -221,10 +188,7 @@ magnify(struct output *output, struct wlr_buffer *output_buffer, struct wlr_box
 	}
 
 	/* And finally mark the extra damage */
-	*damage = border_box;
-	damage->width += 1;
-	damage->height += 1;
```
tokyo4j, line 226: I don't know what these incrementations were for.

src/magnifier.c
Consolatis: I don't quite remember, the magnifier implementation started its life as a proof-of-concept implementation and was then modified before being merged. I think the reason for the `+1` here was that the cursor coordinates themselves where not part of the box, e.g. a width of `100` would result in `50 + 1 for the cursor + 50 == 101 pixels`.

src/magnifier.c
Consolatis: Don't remember either, removing them seems fine to me.

src/magnifier.c
Consolatis: As this PR now uses intersection with the output geometry it seems fine to remove.

Consolatis:
Tested nested with `WLR_WL_OUTPUTS=2` and didn't spot any difference in behavior so far. Didn't yet have time to look into the diff in detail. Moving the cursor between two nested outputs always had issues with the initial rendering of the rect on the 2nd output from what I remember and is likely related to wayland backend optimizations (e.g. the movement of the cursor needs to produce some kind of damage on the new output before it refreshes once and thus kickstarts the magnifier on that output).

Overall the code looks a lot cleaner to me now.

tokyo4j:
> Moving the cursor between two nested outputs always had issues with the initial rendering of the rect on the 2nd output from what I remember and is likely related to wayland backend optimizations (e.g. the movement of the cursor needs to produce some kind of damage on the new output before it refreshes once and thus kickstarts the magnifier on that output).

Yeah it should be addressed when fixing #1836 later.

johanmalm:
LGTM. Is this waiting for anything before merge?

I don't get the `wlr_output.swapchain.format` - but happy to go with it.

Suggest doing a merge-commit on this one to get the reference for the PR (and the diagram in it)

johanmalm:
Have tested. Seems to work the same as before.

tokyo4j:
> I don't get the wlr_output.swapchain.format - but happy to go with it.

I'm not sure this is the best approach, either. For example, if the renderer is pixman and the allocator is gbm, we allocate gbm buffer as a temporary buffer for copy-and-paste, but I'm a bit afraid it ends up reading a big chunk of data from gbm buffer which I guess is not optimized for read operations from CPU (I realized this when I was implementing #2172).

But I think the change in this PR is fine because at least this PR will not change the buffer format of the temporary buffer.

PR #2599, opened 2025-02-23
ssd-titlebar: show fallback icon when no app_id is set

tokyo4j:
Before this PR, nothing was rendered in window icon button if the application doesn't set its app_id (e.g. nested kwin_wayland). This was my oversight in #2518.

This PR makes sure the fallback icon (set via `<theme><fallbackAppIcon>`) is shown instead, by removing `ssd->state.app_id` and the early-return in `ssd_update_window_icon()`. Note that `view_get_string_prop(view, "app_id")` returns an empty string if app id is not set, and passing an empty string to `scaled_icon_buffer_set_app_id()` always renders the fallback icon.

Removing `ssd->state.app_id` doesn't affect performance because `scaled_icon_buffer` caches app_id and `scaled_icon_buffer_set_app_id()` does nothing when a duplicated app_id is passed.

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -83,8 +83,6 @@ struct ssd {
 			struct ssd_state_title_width active;
 			struct ssd_state_title_width inactive;
 		} title;
-
-		char *app_id;
 	} state;
 
 	/* An invisible area around the view which allows resizing */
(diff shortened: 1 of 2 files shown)


Consolatis:
Thanks!

PR #2603, opened 2025-02-28
config: make prefix.desktops default to nothing

johanmalm:
...because the parser ignores empty nodes, so users cannot get rid of the previous default of "Workspace" without using <names> which is not that obvious.

This is a breaking change, but feels like a soft one for a good reason.

Fixes: #2601

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1025,7 +1025,24 @@ entry(xmlNode *node, char *nodename, char *content, struct parser_state *state)
 		return;
 	}
 
-	/* handle the rest */
+	if (!strcasecmp(nodename, "prefix.desktops")) {
+		xstrdup_replace(rc.workspace_config.prefix, content ? content : "");
+		return;
+	}
+
+	/*
+	 * Nodenames where we want to honour !content have to be parsed above
+	 * this point. An example of this is:
+	 *
+	 *     <desktops>
+	 *       <prefix></prefix>
+	 *     </desktops>
+	 *
+	 * In the case of the <prefix> element having content, the node will be
+	 * processed twice; first for the element itself (with no content) and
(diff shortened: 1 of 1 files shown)


tokyo4j:
I think this PR is too breaking. I actually use `<desktops number="2">` and `<action name="GoToDesktop" to="Workspace 1" />` keybinds, and this PR suddenly invalidates those keybinds. I don't think that breakage is worth the benefit of allowing an empty prefix, as number-only workspace names has been already achievable by manually setting `<desktop><names><name>`.

An alternative approach is to disable the `xmlIsBlankNode()` check for `<desktops><prefix>` nodes, but I'm not sure how complicated the implementation will be.

johanmalm:
You're right. Thanks for challenging.
I fell into the trap of not taking breaking changes seriously enough - particularly when the net-benefit is not clear.
I have refactored to honor `<desktops><prefix>` nodes with no content.

johanmalm:
@tokyo4j - would you be happy to review this one with a view to merge?

tokyo4j:
This PR still breaks the default value of `<desktops><prefix>`. I opened #2611 which just allows an empty string.

johanmalm:
Have fixed. Ready for review again.

tokyo4j:
LGTM

Honestly I'm not very happy about the fact that `<desktop prefix="" />` doesn't work, but I think this is acceptable as we already have some oddity about the compatibility between `<foo><bar>` and `<foo bar="">` like the constraint of attribute ordering. I think we'd better fix the issue about XML parsing in general before 1.0.

johanmalm:
Oh, I did not even think to check the attribute case! Sorry. That explains why your branch did not work for me.

Hmm, yeah agree - there is some parsing refinement required.

Issue #2602, opened 2025-02-27
Difficulty snapping windows in multi-monitor setup


EbonJaeger:
### Problem description

I have two monitors, left and right. Snapping windows on the edge between the monitors is difficult; it feels like I have to have near pixel perfect placement of my mouse to get it to snap, on either monitor.

Additionally, the detection region for snapping to the left half of the right-side monitor actually appears to be at the right edge of the left-side monitor.

### Steps to reproduce

1. Launch labwc with a multi-monitor setup where the displays are next to each other horizontally
2. Open a window of any application
3. Try to snap the window to the right-side monitor by dragging the window the the left edge of the monitor
AND/OR
3. Try to snap the window to the left-side monitor by dragging the window the the right edge of the monitor

### labwc build source

Release

### labwc version

0.8.3

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Solus 4.7

tokyo4j:
You can set `<snapping><range>` (default: 1) to a larger value in order to allow window snapping over larger areas.

EbonJaeger:
Ah, that does make the experience *much* better. I'm not sure if you want to keep this open or not. I would consider the default to be not great, so maybe it could serve as a conversation about having a better default.

johanmalm:
I'm happy to increase this a bit. @tokyo4j - thoughts?

tokyo4j:
> I'm happy to increase this a bit. [@tokyo4j](https://github.com/tokyo4j) - thoughts?

Agreed. Something like 10px would be nice.

PR #2608, opened 2025-03-04
config: increase default <snapping><range> to 10

johanmalm:
...to make it easier to snap windows on the edge between two monitors.

Fixes: #2602

@Consolatis @jlindgren90 @ahesford @tokyo4j - at the back of my mind there is something telling me that we picked `1` to avoid some other behaviour, but I have tested it with `10` on a multi-output setup and it seems to work really well. Just wanted to see if I could jog your memories on anything. If not, I suggest we try this value.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -409,7 +409,7 @@ extending outward from the snapped edge.
 	If an interactive move ends with the cursor a maximum distance *range*,
 	(in pixels) from the edge of an output, the move will trigger a
 	SnapToEdge action for that edge. A *range* of 0 disables snapping via
-	interactive moves. Default is 1.
+	interactive moves. Default is 10.
 
 *<snapping><overlay><enabled>* [yes|no]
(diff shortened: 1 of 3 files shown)


ahesford:
I don't have a strong opinion, because I don't really use snapping.

jlindgren90:
Fine with me

tokyo4j:
Perhaps we should increase 
<snapping><overlay><delay><inner> to prevent flashy overlays, because the bigger <snapping><range> is, the longer the cursor stays in the snapping area when dragging a window between outputs.

johanmalm:
Thanks :smile:

Consolatis:
> at the back of my mind there is something telling me that we picked 1 to avoid some other behaviour,

Just for reference, the original default of `1` was from the initial implementation and was based on a single output and "throwing" a window as much into a direction as possible (and thus the cursor always landing on the outer most edge so 1px was sufficient). The increase here absolutely makes sense to handle multiple outputs.

PR #2610, opened 2025-03-06
call selected menu action with keypad enter

zeussuse:
I find myself trying to select menu items with the keypad enter key all the time. I'm not a coder but this seems easy enough. Please forgive me if egregious!


johanmalm:
Thanks. Looks good.

tokyo4j:
Just a trivial suggestion for C programming: you can write multiple switch cases like:

```c
		case XKB_KEY_Return:
		case XKB_KEY_KP_Enter:
			menu_call_selected_actions(server);
			break;
```

PR #2611, opened 2025-03-08
rcxml: allow empty <desktops><prefix>

tokyo4j:
Supersedes #2603.

In this PR, the early-return with `xmlIsBlankNode()` is ignored if the node name is `prefix.desktops`, which allows an empty string in `<desktops><prefix>`.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -941,12 +941,6 @@ entry(xmlNode *node, char *nodename, char *content, struct parser_state *state)
 
 	static uint32_t button_map_from;
 
-	if (!nodename) {
-		return;
-	}
-	string_truncate_at_pattern(nodename, ".openbox_config");
-	string_truncate_at_pattern(nodename, ".labwc_config");
-
 	if (getenv("LABWC_DEBUG_CONFIG_NODENAMES")) {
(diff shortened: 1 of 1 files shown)


src/config/rcxml.c
```diff
@@ -1765,6 +1773,7 @@ post_processing(void)
 			workspace = znew(*workspace);
 			workspace->name = strdup_printf("%s %d",
 				rc.workspace_config.prefix, i + 1);
+			string_strip(workspace->name);
```
tokyo4j, line 1776: This is slightly breaking, but I don't think there's much demand for whitespaces in workspace names.

johanmalm:
This one doesn't work for me. Have tried with `<prefix></prefix>` and `<prefix/>`. Get the full "Workspace 1", etc either way which is obviously not the intent.

I think there is a problem using `string_strip()` in this way in that it only does the `rstrip()` bit and if the format was `"%s %d"` then presumably with an empty `"%s"` the resulting string will start with a space...?

I have updated my PR as an alternative.

tokyo4j:
You're right. I only tested with the shortened form (`<dekstops prefix="">`) and didn't notice the problem.

PR #2612, opened 2025-03-09
menu: lazily generate menu scenes

tokyo4j:
Split from #2531.

In this PR, scene-nodes of (sub)menus are lazily generated when they are being opened. This removes the need to call `update_client_list_combined_menu()` and `update_client_send_to_menu()` every time a root menu is opened. This also fixes the incorrect menu position with following configuration:
```xml
  <menu id="foo" label="foo">
    <item label="aaaaaa"/>
    <item label="bbbbbb"/>
  </menu>
  <menu id="root-menu">
    <menu id="foo" />
    <menu id="foo" />
  </menu>
```

This PR also:
- Removes `menu_free_from()` in favor of simplicity.
- Allows linking to static menu from pipemenu. I think #2339 was enough to prevent ~static menus from opening in unexpected position~ circular menus.
- Renames `menu_configure()` to `menu_reposition()`.
- Asserts `server->menu_current != NULL` in `menu_close_root()`.

### DIFF
--- a/include/menu/menu.h
+++ b/include/menu/menu.h
@@ -127,7 +127,4 @@ void menu_close_root(struct server *server);
 /* menu_reconfigure - reload theme and content */
 void menu_reconfigure(struct server *server);
 
-void update_client_list_combined_menu(struct server *server);
-void update_client_send_to_menu(struct server *server);
-
 #endif /* LABWC_MENU_H */
--- a/src/action.c
+++ b/src/action.c
@@ -664,16 +664,6 @@ show_menu(struct server *server, struct view *view, struct cursor_context *ctx,
 		return;
 	}
 
-	/*
-	 * We always refresh client-list-combined-menu and client-send-to-menu
-	 * so that they are up-to-date whether they are directly opened as a
-	 * top-level menu or opened as a submenu which we don't know at this
-	 * point. It is also needed to calculate the proper width for placement
-	 * as it fluctuates depending on application/workspace titles.
-	 */
-	update_client_list_combined_menu(menu->server);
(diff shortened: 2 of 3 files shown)


johanmalm:
This one looks much easier to review with separate commits. Shall I start with this PR?

tokyo4j:
> This one looks much easier to review with separate commits. Shall I start with this PR?

Yes, please. I'll rebase the original PR when this PR is merged.

tokyo4j:
> Why are we removing the check in menu: allow linking to static menu from pipemenu? In my mind that prevents circular behaviour. Surely pipemenus should not open menus higher up in the hierarchy?

I thought circular menus are prohibited with the following code introduced in #2339:

https://github.com/labwc/labwc/blob/7cc0c479230ad4262d481d2c8c5432bf5320e20d/src/menu/menu.c#L766-L773

At least this prohibits linking to `root-menu` from pipemenus. But I'm open to revert the change if this code doesn't cover all the circumstances.

johanmalm:
I think that is different. We do not want pipemenus "pointing" to any static menu-ids anywhere (which might be outside the iter->parent lineage. 

Unless there is a use-case I would feel more comfortable dropping that commit.

tokyo4j:
Okay, I'll drop that commit.

But I just want to make sure: don't we want pipemenus linking to static menus in order to prevent circular menus? If so, I think we should expand that prevention also to static-to-static menu relationship, since we should regard both `menu.xml` and pipemenu outputs as "untrusted" and prevent any unexpected behaviors when their contents are invalid. Anyway, it can be done in follow-up PRs.

PR #2613, opened 2025-03-09
Some fixes about focus or stacking order management while window switching

tokyo4j:
- Don't call `cursor_update_focus()` every time the OSD is updated as it is no-op while window switching. Instead, they should be called at the beginning/end of window switching. This change functionally does nothing. Ref: https://github.com/labwc/labwc/pull/2584#issuecomment-2663425104
- Make it harder to mess up the window stacking order while window switching by adding a check for `server->input_mode` in `desktop_focus_view()` and `view_minimize()`. Especially the check in `view_minimize()` should prevent panels from minimizing a window while window switching.
- Allow only `PreviousWindow`/`NextWindow` actions while window switching. This should catch some actions that could not be filtered by the 2nd commit e.g. `SendToDesktop`. This was suggested by @Consolatis in IRC.

Both/either of the 2nd and 3rd commit prevent `<action name="NextWindow" /><action name="Focus" />` keybinds from messing up the window stacking order while window switching, which was found in #2582.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -877,6 +877,14 @@ actions_run(struct view *activator, struct server *server,
 	}
 
 	wl_list_for_each(action, actions, link) {
+		if (server->input_mode == LAB_INPUT_STATE_WINDOW_SWITCHER
+				&& action->type != ACTION_TYPE_NEXT_WINDOW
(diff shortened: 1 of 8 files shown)


johanmalm:
This looks great. Thanks very much.
LGTM :smile:

PR #2618, opened 2025-03-13
project wide: clean up event listeners on shutdown (part 2)

tokyo4j:
Follow-up for #2596

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -762,6 +762,10 @@ server_finish(struct server *server)
 	wl_list_remove(&server->new_constraint.link);
 	wl_list_remove(&server->output_power_manager_set_mode.link);
 	wl_list_remove(&server->tearing_new_object.link);
+	if (server->drm_lease_request.notify) {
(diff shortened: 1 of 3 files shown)


src/server.c
```diff
@@ -762,6 +762,10 @@ server_finish(struct server *server)
 	wl_list_remove(&server->new_constraint.link);
 	wl_list_remove(&server->output_power_manager_set_mode.link);
 	wl_list_remove(&server->tearing_new_object.link);
+	if (server->drm_lease_request.notify) {
+		wl_list_remove(&server->drm_lease_request.link);
+		server->drm_lease_request.notify = NULL;
```
tokyo4j, line 767: This line is actually not needed, but I added it to harden the logic and align with:

https://github.com/labwc/labwc/blob/9e6aaa689aafecff7cd38e1c501ffd04219d077f/src/decorations/xdg-deco.c#L23-L26

johanmalm:
Thanks.

PR #2619, opened 2025-03-13
data: center logo vertically

jlindgren90:
The current logo has asymmetrical 62/19 px top/bottom padding. Let's center it (40/41 px top/bottom) so it looks better in the titlebar.


johanmalm:
Good shout.

PR #2620, opened 2025-03-14
osd: draw window switcher with scene-nodes

tokyo4j:
This PR is a preparation for icons in the window switcher (https://github.com/labwc/labwc/discussions/2172#discussioncomment-12297481) which will be achieved by using `scaled_icon_buffer`. Also, this PR may reduce the overhead by not redrawing the whole osd when the selected item is updated.

This PR adds `output->osd_scene` which has 2 fields:
- `tree`: a scene-tree of per-output osd inserted as a child of `output->osd_tree`.
- `items`: a list of `osd_scene_item`s which are used to toggle the highlights for selected window item. In the future, `osd_scene_item` may be extended to support filled highlights for selected window item, or mouse interaction.

A concern about this PR is the ugly look of the window switcher when `osd.window-switcher.width` is too small:
|Before|After|
|-|-|
|20250314_18h57m38s_grim|20250314_18h56m10s_grim|

But I think this is tolerable and I'd like to deliver icon support first. Any ideas to fix this without adding too much complexity are welcome.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -412,6 +412,12 @@ struct output {
 	struct wlr_scene_tree *osd_tree;
 	struct wlr_scene_tree *session_lock_tree;
 	struct wlr_scene_buffer *workspace_osd;
+
+	struct osd_scene {
+		struct wl_array items; /* struct osd_scene_item */
+		struct wlr_scene_tree *tree;
+	} osd_scene;
+
 	/* In output-relative scene coordinates */
 	struct wlr_box usable_area;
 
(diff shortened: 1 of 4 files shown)


johanmalm:
Thanks LGTM.

Regarding `theme->osd_window_switcher_width`, I think we ought to do some sanity checking in `post_processing()`. It's a bit like `theme->titlebar_height` in that there is a relationship with the object within it. In the case of the window-switcher I suspect we ought to set an arbitrary minimum and but on the whole I think it would be better to shrink the contents to fit. Anyway, I feel that is a separate PR because although it looks a bit different now, the fundamental issue is pre-existing.

PR #2621, opened 2025-03-14
osd: support app icons in window switcher

tokyo4j:
Based on #2620.

Example configuration:
```xml
  <windowSwitcher>
    <fields>
      <field content="icon" width="5%" />
      <field content="title" width="95%" />
    </fields>
  </windowSwitcher>
```

20250314_19h41m02s_grim


tokyo4j:

I'd like to modify the defaults to show app icons in the window switcher by default, but it can be done in another PR as we should consider performance degradation due to icon lookups and support for systems without libsfdo.

johanmalm:
LGTM. Have tested and confirm it works well. Agree that we ought to enable icons be default - in a follow up PR.

Please consider icon-size options. I think that for this PR it's fine to set to `font_height` for simplicity, but we may want to add a dedicated theme option for this (possibly with a fallback to use the titlebar icon size).

I've just had a look at Openbox and the icons there are much bigger than the text (too big in my view!).

tokyo4j:
> Please consider icon-size options. I think that for this PR it's fine to set to font_height for simplicity, but we may want to add a dedicated theme option for this (possibly with a fallback to use the titlebar icon size).

Agreed. I think we can add something like `osd.window-switcher.icon-size` and calculate the item height like:
```diff
diff --git a/src/theme.c b/src/theme.c
index 51468e02..38b23aea 100644
--- a/src/theme.c
+++ b/src/theme.c
@@ -1380,9 +1380,10 @@ post_processing(struct theme *theme)
 	theme->menu_header_height = font_height(&rc.font_menuheader)
 		+ 2 * theme->menu_items_padding_y;
 
-	theme->osd_window_switcher_item_height = font_height(&rc.font_osd)
-		+ 2 * theme->osd_window_switcher_item_padding_y
-		+ 2 * theme->osd_window_switcher_item_active_border_width;
+	theme->osd_window_switcher_item_height =
+		MAX(font_height(&rc.font_osd), theme->osd_window_switcher_icon_size)
+			+ 2 * theme->osd_window_switcher_item_padding_y
+			+ 2 * theme->osd_window_switcher_item_active_border_width;
 
 	if (rc.corner_radius >= theme->titlebar_height) {
 		rc.corner_radius = theme->titlebar_height - 1;
```

jlindgren90:
I love this, it was the biggest outstanding feature I was missing from Openbox. Thanks @tokyo4j !

protostork:
Agree @jlindgren90, absolutely amazing and long-missed feature, thank you so much @tokyo4j!

PR #2622, opened 2025-03-14
string-helpers: rtrim() with just char *s, call it later in string_strip

domo141:
char **s not needed to get trailing whitespace trimmed, and rtrim() does not return anything

if there is leading whitespace in *s in call to string_strip(), there is less chars left to scan in rtrim().

---
noticed this while doing something; if it supposed to be as it is, i can amend this commit w/ comment

### DIFF
--- a/src/common/string-helpers.c
+++ b/src/common/string-helpers.c
@@ -30,14 +30,14 @@ trim_last_field(char *buf, char delim)
 }
 
 static void
-rtrim(char **s)
+rtrim(char *s)
 {
-	size_t len = strlen(*s);
+	size_t len = strlen(s);
 	if (!len) {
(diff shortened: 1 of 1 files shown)


tokyo4j:
LGTM. I like the change from `char **` to `char *`. But I wonder:

> if there is leading whitespace in *s in call to string_strip(), there is less chars left to scan in rtrim().

Does calling `rtrim()` later reduces the number of characters to scan? Do you mean `strlen()` scanning the string for null byte?

domo141:
I meant if there is leading whitespace that is now removed (and s advanced), there is left
for rtrim() to scan.

johanmalm:
Thanks

PR #2633, opened 2025-03-23
action: Add HideCursor action

### LINKED ISSUE #2614: [ Feature Request ] Option to hide mouse cursor
I think it would be nice to have an option that would allow you to have the cursor hide after X seconds of inactivity and reappear on movement. An example of usage would allow taking screenshots with grimp and not having the cursor in the shot. Is this something that others would be interested in too ?

jp7677:
Very simple action which, well, hides the cursor. 

Tested with:

```xml
<keybind key="A-W-h">
  <action name="HideCursor" />
  <action name="WarpCursor" x="-1" y="-1" />
</keybind>
``` 

Because naming is hard, I'm not sure if this is the best name for this action. The original issue says `ToggleCursor` or `SetCursorVisibility`, but I think having a toggle or visible/not-visible like action is overkill because the cursor, when hidden, becomes visible again as soon as pointer/gestures or stylus input is detected.

This also works directly on startup by, for above key binding, adding `wtype -M alt -M logo h -m alt -m logo` to `.config/labwc/autostart`. 

CC: @justinesmithies
Closes #2614

PS: Would be nice actually to directly execute an action incl. argument from the labwc executable, similar to `labwc --reconfigure`, to not need a key binding for the "hide only at startup" use case. Hint.. hint.. ;)

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -351,6 +351,12 @@ Actions are used in menus and keyboard/mouse bindings.
 	*y* [center|value] Equivalent for the vertical warp position within the target area.
 	Default is "center"
 
+*<action name="HideCursor" />*
+	Hide the pointer or stylus cursor. The cursor becomes visible again on following
+	pointer actions, stylus actions or touchpad gestures.
+	Use together with the WarpCursor action to not just hide the cursor but to
+	additionally move it away to prevent e.g. hover effects.
+
 *<action name="EnableTabletMouseEmulation" />*++
 *<action name="DisableTabletMouseEmulation" />*++
 *<action name="ToggleTabletMouseEmulation">*
--- a/src/action.c
+++ b/src/action.c
@@ -123,6 +123,7 @@ enum action_type {
 	ACTION_TYPE_ZOOM_IN,
 	ACTION_TYPE_ZOOM_OUT,
 	ACTION_TYPE_WARP_CURSOR,
+	ACTION_TYPE_HIDE_CURSOR,
 };
 
 const char *action_names[] = {
@@ -188,6 +189,7 @@ const char *action_names[] = {
 	"ZoomIn",
 	"ZoomOut",
 	"WarpCursor",
+	"HideCursor",
 	NULL
 };
 
@@ -1334,6 +1336,9 @@ actions_run(struct view *activator, struct server *server,
 
 				warp_cursor(view, output, to, x, y);
 			}
+			break;
+		case ACTION_TYPE_HIDE_CURSOR:
+			cursor_set_visible(&server->seat, false);
 			break;
 		case ACTION_TYPE_INVALID:
 			wlr_log(WLR_ERROR, "Not executing unknown action");


johanmalm:
Thanks.

> PS: Would be nice actually to directly execute an action incl. argument from the labwc executable, similar to `labwc --reconfigure`, to not need a key binding for the "hide only at startup" use case. Hint.. hint.. ;)

Well, it's hard to know where the best separation line is for `labwc` to flourish in the long run. I think custom IPC would be preferable to a "startup-only action" from the command line... but if we open the plugin / custom-IPC door it is then hard to ever draw a sensible line on what is in or out and the balance then becomes scope vs maintainer focus/interest.

jp7677:
> Thanks.
> > PS: Would be nice actually to directly execute an action incl. argument from the labwc executable, similar to `labwc --reconfigure`, to not need a key binding for the "hide only at startup" use case. Hint.. hint.. ;)
> 
> Well, it's hard to know where the best separation line is for `labwc` to flourish in the long run. I think custom IPC would be preferable to a "startup-only action" from the command line... but if we open the plugin / custom-IPC door it is then hard to ever draw a sensible line on what is in or out and the balance then becomes scope vs maintainer focus/interest.

Thanks for the answer. Yeah, I agree that this is indeed out of scope for such a limiting use case. I never looked at the code behind “reconfigure”, but having done now, I realized this is much more simple and straightforward (in a positive sense) than i thought and not extendable for executing arbitrary actions.

tokyo4j:
I tested `HideCursor` and I like it. At first I thought it's more intuitive to add an action (e.g. `ToggleCursorVisibility`) which permanently shows/hides the cursor by adding an variable like `seat->force_visible`, but `HideCursor` in this PR is simpler and sufficient for such a limited use-case.

yulipumal:
I'm trying to get it working on my raspi 4 (Wayland labwc).
I added the following to the rc.xml:
```
<?xml version="1.0"?>
<openbox_config xmlns="http://openbox.org/3.4/rc">
	<touch deviceName="WingCoolTouch WingCoolTouch" mapToOutput="HDMI-A-2" mouseEmulation="yes"/>
</openbox_config>
<labwc_config>
	<keyboard>
		<keybind key="A-W-h">
		<action name="HideCursor" />
		<action name="WarpCursor" x="-1" y="-1" />
		</keybind>
	</keyboard>
</labwc_config>
``` 
But I'm not able to do the keyboard shortcut. Is that correctly applied?

jp7677:
Should be good. Please ensure that you are running labwc 0.8.4 which got released three days ago. Older versions do not know this action. 
Side note: it's better to open a discussion for questions like this. Commenting on a closed PR is a hit and miss.

beep41:
I've been having trouble with this. I'm on version 0.8.4. I'm also very new to this, so maybe it's a mistake I made.

bdantas:
I love this HideCursor action, which allows behavior similar to `unclutter` on X. I did a labwc keybind for `Alt+h` to trigger the HideCursor action.

There is one fly in the ointment, though. This is my config file for `swayidle`:
```
timeout 5 "ydotool key 56:1 35:1 35:0 56:0" resume "echo I'm awake"
timeout 10 "some_other_command"
timeout 15 "another_command"
```
 
After five seconds, swayidle triggers the `ydotool` command (which sends `Alt+h` and successfully triggers the HideCursor action), but the keyboard input generated by ydotool causes swayidle to run the resume command and restart the idle clock! The end result is that, even if I don't touch the computer, every 5 seconds the ydotool command is run and "I'm awake" appears in terminal, while `some_other_command` and `another_command` never run :(

Is there some way to trigger the HideCursor action in a way that doesn't involve keyboard activity (e.g., with a signal to labwc) so that the idle clock can keep ticking?

Frederick888:
@yulipumal You need to remove the `<openbox_config>` section. (Sorry I don't know where you should put `<touch>` after removing that. I guess it can also be in `<labwc_config>` if https://github.com/odoo-dev/odoo/commit/3263f9029205bbce16815366b0bb9574d9e08005#diff-34fb11edf051e9dcab3cf30ef6836ce8ce92fd94f15a2ab7924c9280c1f9cbdbR4 is right.)

Issue #2627, opened 2025-03-18
A new window is not focused if it has been minimized before being mapped


tokyo4j:
### Problem description

When an application (e.g. VSCode, ref: https://github.com/microsoft/vscode/issues/242726) creates a `xdg_toplevel`, **minimizes** it and maps it, I observe 2 weirdnesses:
- The new window is not automatically focused
- ~The previous window cannot be selected with window switcher~ (fixed by #2638)

I made a simple reproducer: https://github.com/tokyo4j/wayland-demo/tree/minimize-before-map

https://github.com/user-attachments/assets/d4adcdcc-3d7d-44b2-b4ff-065df3abfb4a

### Steps to reproduce

1. Clone and build the reproducer above
2. Run `./client` from a terminal emulator window
3. Try to select the terminal emulator window with window switcher

### labwc build source

Local build

### labwc version

0.8.3

### labwc environment

From a TTY or some display manager like lightdm

linux36524:
"What’s your taskbar? Are you using tint2 or waybar?"

tokyo4j:
> "What’s your taskbar? Are you using tint2 or waybar?"

It's waybar, but don't ask me such a off-topic thing here. Instead you can ask me via IRC or email.

tokyo4j:
Updated the title and description to follow the current state.

For `xdg-shell` clients that minimizes itself before mapping it, KWin seems to open a window minimized (thus its presence is only indicated in the panel), while Mutter seems to just ignore the minimization request and open the window as usual. I prefer to follow Mutter's behavior as it will only require an early-return in `view_minimize()`.

But I wonder how we should handle this for X11 clients. @jlindgren90 Is it allowed to request minimization before mapping a window in X11?

jlindgren90:
> But I wonder how we should handle this for X11 clients. [@jlindgren90](https://github.com/jlindgren90) Is it allowed to request minimization before mapping a window in X11?

I don't know off-hand. Do you know of an X11 client that does this, and can we test it under Openbox to see what happens?

tokyo4j:
I asked ChatGPT to write a C program that does it:

<details><summary>code</summary>

```c
#include <X11/Xatom.h>
#include <X11/Xlib.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

int main() {
    Display *display = XOpenDisplay(NULL);
    if (!display) {
        fprintf(stderr, "Unable to open X display\n");
        return 1;
    }

[...]

PR #2638, opened 2025-03-27
Remove `server->last_raised_view` and fix incorrect stacking order of always-on-top windows in xwayland

### LINKED ISSUE #2629: [Regression] Window switcher misbehaves when new window is mapped while window switching
#2613 caused a regression: when a new window is mapped while window switcher is activated, the previously focused window cannot be selected (raised) with the window switcher.

https://github.com/user-attachments/assets/b6f8d792-4356-4b31-a58d-a984f745083a

The cause is that, since #2613, `view_move_to_front()` is not called for new windows in `view_impl_map()` while window switching, so `server->last_raised_view` can be still the previously focused window and thus it cannot be raised at the end of window switching despite that the new window is actually on top of it in the scene graph.

I guess such an out-of-sync state is the root cause of #1640, #2582 and #2627, too.

tokyo4j:
Summary: This PR removes `server->last_raised_view` to fix #2629 and mitigate #2627, but also fixes incorrect stacking order of always-on-top windows in xwayland.

## 1st commit: Remove `server->last_raised_view`

We were skipping in `view_move_to_front()` if the raised view is already cached in `server->last_raised_view`. But this was prone to bugs that windows cannot be raised even though they are actually not placed at the top in the scene.

This happened when a window is mapped but `view_move_to_front()` is not called in `view_impl_map()` for some reason. Example cases were:
- a window is minimized before being mapped (#2627)
- a window is mapped while window switcher is active (#2629)

Also, this problem was the root cause of #1640 and #2582, though they are already fixed.

So this PR removes `server->last_raised_view` in the 1st commit and adds `server->stacked_xsurfaces` as an alternative caching mechanism in the 2nd commit explained below.

## 2nd commit: Fix incorrect stacking order of always-on-top (AOT) windows in xwayland

Currently, when a normal window is raised, xwayland thinks it's above AOT windows even though it's actually below AOT windows in the scene. This means mouse scroll events may be unexpectedly sent to normal windows below AOT windows even when the cursor is hovering over a AOT window. I guess this is what @jlindgren90 meant in the last part of the comment below:

https://github.com/labwc/labwc/blob/d6edb8b2738c894307acccde57a45eca9e3fa113/src/xwayland.c#L833-L836

So this PR fixes it by notifying the correct stacking order (where AOT windows are placed above normal windows) to xwayland every time the stacking order is updated.

I also did following things in this commit:
- Aggregated other logic about stacking order management in xwayland (e.g. shaded windows or windows in other workspaces should be notified to xwayland as being placed at the bottom) into `xwayland_adjust_stacking_order()` in order to improve readability and predictability.
- Added `server->stacked_xsurfaces` as a caching mechanism alternative to `server->last_raised_view` removed in the 1st commit.


jlindgren90:
Nice analysis and write-up! The concept seems sounds to me.

I'm not thrilled about the potential for dangling pointers in `server->stacked_xsurfaces`. Maybe it ends up being okay in practice, but it still makes me uncomfortable.

`append_xsurfaces()` also seems like it's doing a lot of memory allocation, for a function that may (if I'm understanding correctly) run on every single mouse movement. Again - maybe my concern is overblown. (In my day job, I work on embedded systems, so memory allocations are evil ;)

Brainstorming for an alternative implementation ...

What if we had a numeric scheme for the stacking position of each xsurface, say a uint32, where:

- The high 2 bits are the "group" (0 = hidden, 1 = bottom, 2 = normal, 3 = top)
- The remaining 30 bits are the stacking position within that group (equivalent to that xsurface's index in the 4 wl_arrays you have currently)

Then, each `struct xwayland_view` contains a `uint32_t stacking_order`. The separate `server->stacked_xsurfaces` list goes away, because it's now embedded into the existing structs directly.

`append_xsurfaces()` and `xwayland_adjust_stacking_order()` get rolled into one function, and we change up the approach. Rather than iterating over `server->views` once and then sorting arrays, we iterate 4 times to enumerate the xsurfaces directly in the order we want them. First the hidden views, then bottom, then normal, then top.

In each iteration we can compute what the new/desired `stacking_order` for each xsurface should be right away. As soon as this doesn't match the current `stacking_order` of any view, we start restacking them all, bottom-to-top, just like you're doing now at the very end.

It may be a little more involved to code up. But I think it does the logical equivalent of your current approach, while eliminating all the wl_arrays and with them the potential for dangling pointers.

Let me know what you think. I could potentially prototype something if there's interest (unless you'd rather do so yourself). I could also be okay with merging what you have, and trying to improve/optimize it later.

tokyo4j:
> for a function that may (if I'm understanding correctly) run on every single mouse movement

I thought this never happens, but it turned out that `view_move_to_front()` is called on every single mouse movement if `<focus><followMouse>` and `<focus><raiseOnFocus>` are both set to yes.

> What if we had a numeric scheme for the stacking position of each xsurface, say a uint32, where:
> 
> - The high 2 bits are the "group" (0 = hidden, 1 = bottom, 2 = normal, 3 = top)
> - The remaining 30 bits are the stacking position within that group (equivalent to that xsurface's index in the 4 wl_arrays you have currently)
>
> Then, each struct xwayland_view contains a uint32_t stacking_order. The separate server->stacked_xsurfaces list goes > away, because it's now embedded into the existing structs directly.
>
> append_xsurfaces() and xwayland_adjust_stacking_order() get rolled into one function, and we change up the approach. Rather than iterating over server->views once and then sorting arrays, we iterate 4 times to enumerate the xsurfaces directly in the order we want them. First the hidden views, then bottom, then normal, then top.
> 
> In each iteration we can compute what the new/desired stacking_order for each xsurface should be right away. As soon as this doesn't match the current stacking_order of any view, we start restacking them all, bottom-to-top, just like you're doing now at the very end.

I really like the basic idea of this.

I pushed a simpler version of this approach, where `xwayland_view` contains `stacking_order` and we iterate over the windows 4 times to check `xwayland_view->stacking_order` in the correct stacking order. I'll squash the commit once we confirm this is the way to go.

jlindgren90:
Nice, that is indeed simpler than I was imagining. I have but tested it yet but the code looks good to me. 

You can also remove server->stacked_xsurfaces now, I think.

tokyo4j:
Thanks. Removed `server->stacked_xsurfaces`. This PR is now ready for final review.

johanmalm:
@jlindgren90  - Did you want to carry on the review + merge of this one? Let me know if not and I'll take it on tomorrow night.

jlindgren90:
The code looks good to me. I will not have time to test it for a little bit, so please go ahead.

PR #2639, opened 2025-03-27
add realtime round-robin scheduling capability

zeusgoose:
This patch is based on https://github.com/swaywm/sway/commit/a3a82efbf6b5b3af840c70038b1b599ba29003ac
Eliminates the libinput error "your system is too slow".
Enable by giving the labwc binary CAP_SYS_NICE capability, ie `setcap cap_sys_nice+ep usr/bin/labwc`

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -584,6 +584,8 @@ void output_enable_adaptive_sync(struct output *output, bool enabled);
  */
 float output_max_scale(struct server *server);
 
+void set_rr_scheduling(void);
+
 void new_tearing_hint(struct wl_listener *listener, void *data);
 
 void server_init(struct server *server);
--- a/src/main.c
+++ b/src/main.c
@@ -205,6 +205,8 @@ main(int argc, char *argv[])
 
 	increase_nofile_limit();
 
+	set_rr_scheduling();
+	
 	struct server server = { 0 };
 	server_init(&server);
 	server_start(&server);
--- a/src/meson.build
+++ b/src/meson.build
@@ -18,6 +18,7 @@ labwc_sources = files(
   'output-virtual.c',
   'overlay.c',
   'placement.c',
+  'realtime.c',
   'regions.c',
   'resistance.c',
   'resize-outlines.c',
--- a/src/realtime.c
+++ b/src/realtime.c
@@ -1,0 +1,37 @@
+#include <sys/resource.h>
+#include <sched.h>
+#include <unistd.h>
+#include <pthread.h>
+#include "labwc.h"
+ 
+static void child_fork_callback(void) {
+ 
+	struct sched_param param;
+	param.sched_priority = 0;
+	int ret = pthread_setschedparam(pthread_self(), SCHED_OTHER, &param);
+	if (ret != 0) {
+		wlr_log(WLR_ERROR, "Failed to reset scheduler policy on fork");
+	}
+}
+
+void set_rr_scheduling(void) {
+	int prio = sched_get_priority_min(SCHED_RR);
+	int old_policy;
+	int ret;
+	struct sched_param param;
+	ret = pthread_getschedparam(pthread_self(), &old_policy, &param);
+	if (ret != 0) {
+		wlr_log(WLR_DEBUG, "Failed to get old scheduling priority");
+		return;
+	}
+  
+	param.sched_priority = prio;
+	ret = pthread_setschedparam(pthread_self(), SCHED_RR, &param);
+  
+	if (ret != 0) {
+		wlr_log(WLR_INFO, "Failed to set scheduling priority to %d", prio);
+		return;
+	}
+  
+	pthread_atfork(NULL, NULL, child_fork_callback);
+}


johanmalm:
@zeussuse  Any reason `chrt --reset-on-fork` could not be used instead? We prefer to keep things outside compositor space where practicable.

johanmalm:
Hey - appreciate you joining us and feel a bit guilty shutting this one down but feel it's better done outside the compositor. Some of hang out on IRC if you want to join and discuss areas to get involved :smile:

tokyo4j:
Just to share my experiment, running following commands successfully launched labwc with RR scheduling policy:
```
sudo setcap cap_sys_nice+ep /usr/bin/chrt
chrt --reset-on-fork -r 1 labwc
```

danielfikko:
@johanmalm @Consolatis @tokyo4j @jlindgren90 @ahesford Could you reconsider this PR?

I discovered it recently and am currently using it this way on Arch Linux:

<details>
<summary>$ cat /etc/pacman.d/hooks/util-linux.hook</summary>

```
[Trigger]
Operation = Install
Operation = Upgrade
Type = Path
Target = usr/bin/chrt

[Action]
Exec = /usr/bin/setcap cap_sys_nice+ep /usr/bin/chrt
When = PostTransaction
```
</details>

<details>
<summary>$ cat ~/.zprofile</summary>

```
[[ "$(tty)" = "/dev/tty1" ]] && exec chrt -r -R 1 labwc &> /dev/null
```
</details>

...it works to set `SCHED_RR` for labwc. However, wlroots has a feature to give high priority to the Vulkan queue, but it currently doesn't work with labwc:

```
[render/vulkan/vulkan.c:578] Requesting a high-priority device queue
[render/vulkan/vulkan.c:614] Failed to obtain a high-priority device queue, falling back to regular queue priority
```

You need to set `cap_sys_nice+ep` on the labwc binary for it to work. Of course, I can just create this file myself:

<details>
<summary>$ cat /etc/pacman.d/hooks/labwc.hook</summary>

```
[Trigger]
Operation = Install
Operation = Upgrade
Type = Path
Target = usr/bin/labwc

[Action]
Exec = /usr/bin/setcap cap_sys_nice+ep /usr/bin/labwc
When = PostTransaction
```
</details>

...but it would be much simpler (and less redundant than having two hooks) if labwc implemented `SCHED_RR`, so that distributions could set `cap_sys_nice+ep` on the labwc binary when they package it. Using Arch Linux as an example:

<details>
<summary>$ git diff</summary>

```diff
diff --git a/PKGBUILD b/PKGBUILD
index ab70844..aa479c3 100644
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -8,6 +8,7 @@ pkgdesc='stacking wayland compositor with look and feel from openbox'
 url="https://github.com/labwc/labwc"
 arch=('x86_64')
 license=('GPL-2.0-only')
+install=labwc.install
 depends=(
   cairo
   glib2
diff --git a/labwc.install b/labwc.install
new file mode 100644
index 0000000..a62d373
--- /dev/null
+++ b/labwc.install
@@ -0,0 +1,7 @@
+post_install() {
+  setcap cap_sys_nice+ep /usr/bin/labwc
+}
+
+post_upgrade() {
+  setcap cap_sys_nice+ep /usr/bin/labwc
+}
```
</details>

In addition, it would be in line with the sway: https://gitlab.archlinux.org/archlinux/packaging/packages/sway/-/blob/main/sway.install. And that's something jay does, too: https://github.com/mahkoh/jay/blob/master/book/src/installation.md#cap_sys_nice-optional. jay is a very competent compositor when it comes to performance and latency!

johanmalm:
I am fine with this. Does not feel like a big deal either way from a maintainability perspective, and it is good to listen to what users want.

If sway had not implemented this, I would probably be more weary in term of side-effects and ongoing support with issues, but as it is, it seems pretty low risk from that perspective.

Could reach out to swaywm team for a second opinion.

Any reservations from anyone else?

danielfikko:
Thank you for considering this.

I've opened an MR for the Arch Linux labwc package at: https://gitlab.archlinux.org/archlinux/packaging/packages/labwc/-/merge_requests/2.

This will at least make the high-priority Vulkan queue work now. And if there are no objections, it will eventually make `SCHED_RR` work as well.

danielfikko:
When you grant `CAP_SYS_NICE` to the labwc binary, labwc no longer respects the `XKB_DEFAULT_LAYOUT` environment variable and uses the `English (US)` layout instead.

I opened: https://github.com/labwc/labwc/issues/3662

ahesford:
I no longer retain editorial authority on this project, but I remain opposed to this change. It is inappropriate for the compositor to impose its own scheduling priority on processes it launches. The settling applied by the operator, either by kernel default or specific invocation, should always be respected.

Even if there were a valid argument for labwc to impose scheduling decisions, it should be gated by some user flag or configuration option to allow the user to govern the behavior. This entails exposing additional complexity to the user and just competes with existing control mechanisms.

This is a bad idea all around.

danielfikko:
It's really best to get a second opinion from the wlroots developers (or @mahkoh), because granting `CAP_SYS_CAP` to a binary causes environment variables, such as `VK_DRIVER_FILES`, to break when used in conjunction with that binary.
https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderDriverInterface.md#exception-for-elevated-privileges

I'm just a user, but since sway and jay already do that, I don't think it's a big deal (?)

ahesford:
Quit tagging people into a discussion over a matter closed more than a year ago.

What other compositors do is immaterial. They have different philosophies and impose different standards, as evidenced even by their central strategies for window management.

danielfikko:
If you realized it was an old discussion, you could just ignore it. It's not like a notification on GitHub is the same as me knocking on your door and bothering you.

mahkoh:
>The settling applied by the operator, either by kernel default or specific invocation, should always be respected.

This does not work in general because a process might consist of multiple threads that might want to run with different scheduling priorities. For example, in a compositor you might want to give RR to a thread handling input but not to an llvmpipe software rendering thread spawned by mesa, as doing so could lead to unintentional DOS.

Sysadmins can in general only set these things up per-process. The exception is if you only ever want the main thread to have RR, as the reset-on-fork flag resets the policy in spawned threads.

---

As for CAP_SYS_NICE: As a sysadmin I would not want to set the effective flag on an inode since that might lead to privilege escalation. Setting only the permitted flag ensures that the application only runs with the effective capability if it explicitly enables it.

---

As for environment variables: Some libraries allow you to read environment variables even when running with the secure-execution flag set in the AUXV, which is set when the inode has filesystem capabilities. libxkbcommon is one such library.

danielfikko:
Thank you for your explanation!

Unfortunately, setting only `cap_sys_nice=p` causes the Vulkan high-priority queue to not work.

mahkoh:
The application needs to add the capability to the effective set at runtime.

johanmalm:
Let's start with my situation and perspective. Will then ask some technical question to help me evaluate.

1. I am having a "slow period" so response times will be accordingly.
2. I will propose some more formal notes to help navigate this type of evaluation, but fundamentally - for me - this is a balancing act between staying viable in the long run, user affinity, things just working out-of-the-box and straying into adjacent domains (e.g. distro and DE). The latter is hard to undo in practice and can be quite unhelpful.
3. I appreciate and respect the opinions of ahesford in thinking this through.
4. I do not have much knowledge in this domain, so spelling things out in simple ways helps me.

Now technical...

a. If we do not merge this, can distributions and/or users make the high-priority Vulkan queue work? I get the points about avoid custom hooks, but am looking for a black-and-white to start with.

b. Same question for SCHED_RR as well. Apologies if that is effectively the same question. It is not a space I have any knowledge in.

c. Any help you can provide in describing the use-cases for the high-priority Vulkan queue and SCHED_RR. It is important to describe value before considering new code. Often, the value is obvious, but it is not - for me - with this one.

d. Regardless of anything else, the problem with XKB_DEFAULT_LAYOUT would be a blocker for this for some time if I understand what you have written correctly.

e. I do not understand the context/implications made about p and pe in the last two points. Appreciate help in unpacking this.

danielfikko:
I'm not a developer, but I think I can answer a few questions.

a. Yes, as for Vulkan’s high-priority queue, labwc doesn’t need to do anything about it. All that’s needed is for the distributions to grant the `CAP_SYS_NICE` capability when packaging labwc. Sorry for getting a little mixed up about where labwc and the distribution come in. Obviously, the distribution doesn’t have to do this either, the user can simply set up a hook (just like I did).

b. I’m not sure if distributions would be able to do this one. Here, I think it’s labwc’s responsibility to implement what this PR does: attempt to gain `SCHED_RR`. And it’s the distribution’s responsibility to grant `CAP_SYS_NICE` to the labwc binary when packaging it so that `SCHED_RR` works. The latter doesn’t require the distribution. The user could set up a hook (just like I did).

I mixed a little of `a` and `b` because, with labwc implementing `SCHED_RR`, it would be another reason (in addition to Vulkan's high-priority queue) for distributions to grant `CAP_SYS_NICE` to the labwc binary. Because it would be much simpler if the distribution granted `CAP_SYS_NICE` rather than having to deal with various hooks. In my opinion, it's an interesting feature and that would make it work out of the box, without needing to configure hooks.

But honestly, I’m giving up on advocating for this feature because of `d`.

c. The use case is simply to keep labwc smooth and responsive during heavy load.

d. Yes, I think this is a blocker. So much so that I gave up on the MR for the Arch Linux package because of it.

e. I don’t know how to answer that, best to leave it to mahkoh.

Consolatis:
The `XKB_DEFAULT_LAYOUT` issue could likely be worked around by making libxkbcommon not use `secure_getenv()`. Labwc already has a special case for Android to set `XKB_CONTEXT_NO_SECURE_GETENV` for the xkb context (`src/input/keyboard.c:771`). From a quick glance this seems also what sway is doing. That won't help for the Vulkan case though.

As for the responsibility concerns: I don't think there is anything inherently wrong with trying to enable `SCHED_RR`. If labwc doesn't have the permission to do so its a no-op anyway. So its up to the distros to decide what permissions to give to the labwc executable by their installation procedure and ultimately up to the users to overrule that packaging decision or not.

The question is if we want to go on full realtime though, it might for example interfere with audio latency. A simple nice value of like `-1` (e.g. `echo -1 > /proc/self/autogroup || nice(-1)`) \*might\* also be enough. Or we could add a `<core>` setting offering the options of `realtime|-20 to 20|none`.

zeusgoose:
Yo, I'm mostly a illiterate tard and shouldn't be monkeying with anything code wise. This PR was purely copy/paste, please forgive me.
That said, just out of curiosity I looked at what gamescope does, they do `nice -20`. 
[        int nNewNice = nice( -20 );](https://github.com/ValveSoftware/gamescope/blob/5fb8dce4a09d0a68d097b9faf9513782106bc843/src/Utils/Process.cpp#L437)
Which in my pea-brain would make sense: if you're running a wiz-bang SCX scheduler you'd probably want that instead of RR/FIFO.
As for packaging: I *think* giving the binary cap_sys_nice can be done with a  `systemd-tmpfiles.d`: from `man tmpfiles.d` 
` k     /path-or-glob/to/set/caps                -    -    -     -           file capabilities`

zeusgoose:
I had an idea. working off of tokyo4j's suggestion: setting RR could be accomplished by the user with lab's `autostart`:
`sh -c 'sleep 1;exec foot --title=labRR -o locked-title=yes -w 600x100 -e sh -c "run0 chrt --pid -Rr 1 $(pidof labwc) && chrt --pid $(pidof labwc) && sleep 3"' >/dev/null 2>&1 &`
change foot to your favorite terminal, and run0 to sudo if need be.
you'll be annoyed with a prompt every log-in but thats okay? idk how wrong/bad/stupid it is.. just vomiting thoughts
_________
Edit: thanks for the reminder @elviosak !
allowing the command with `sudoers` looks easy enough. Doing it with systemd + polkit was a little more hoop-jumpy, and it's still probably bad

1. create a unit file: `/etc/systemd/system/labwcrr.service`
```
[Unit]
Description=Labwc Round Robin
Wants=labwc-session.target
After=labwc-session.target

[Service]
Type=oneshot
ExecStart=
ExecStart=sh -c "/usr/bin/chrt --pid -Rr 1 $(pidof labwc)"

[Install]
WantedBy=default.target
```
2. create a polkit rule that allows the user to run said service: could probably limited to your specific user?
`/etc/polkit-1/rules.d/10-labwcrr.rules`
```
polkit.addRule(function(action, subject) {
    if (action.id == "org.freedesktop.systemd1.manage-units") {
        if (action.lookup("unit") == "labwcrr.service") {
            var verb = action.lookup("verb");
            if (verb == "start" || verb == "restart") {
                return polkit.Result.YES;
            }
        }
    }
});
```

3. add to `labwc/autostart`:
`systemctl start labwcrr.service >/dev/null 2>&1 &`

elviosak:
np, i looked on my phone and thought i replied to an old message so i deleted it, but i'm glad it was useful. I know almost nothing about polkit, just know it's there and can be used to execute stuff that needs root.

zeusgoose:
i've tested getting high-priority queue working with my suggestions; none of them work:
Setting cap_sys_nice with a tmpfile does NOT work.
Setting RR with a autostart method does NOT work.

Conclusion: the patch is required
Apologies for blowing smoke

PR #2644, opened 2025-03-30
src/magnifier: fix compile on ancient GCC versions

ahesford:
[On behalf of @Consolatis]

Fixes
```
../src/magnifier.c:153:3: error: missing initializer for field ‘clip’ of ‘struct wlr_render_rect_options’ [-Werror=missing-field-initializers]
  153 |   };
      |   ^
In file included from ../subprojects/wlroots/include/wlr/render/wlr_renderer.h:14,
                 from ../subprojects/wlroots/include/wlr/types/wlr_output.h:17,
                 from ../src/magnifier.c:5:
../subprojects/wlroots/include/wlr/render/pass.h:115:27: note: ‘clip’ declared here
  115 |  const pixman_region32_t *clip
```

### DIFF
--- a/src/magnifier.c
+++ b/src/magnifier.c
@@ -150,6 +150,7 @@ magnifier_draw(struct output *output, struct wlr_buffer *output_buffer, struct w
 				.b = theme->mag_border_color[2],
 				.a = theme->mag_border_color[3]
 			},
+			.clip = NULL,
 		};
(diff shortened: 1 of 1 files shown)


tokyo4j:
Hm, I'm just wondering why that version of GCC only complains about the missing initializer of `wlr_render_rect_options.clip`, but not about `wlr_render_rect_options.blend_mode`. Maybe GCC treats pointers and integers (or enums) differently?

PR #2648, opened 2025-03-31
osd: show icon in window switcher by default

tokyo4j:
Follow-up for #2621.

The default window switcher layout is updated to:

```xml
  <windowSwitcher>
    <fields>
      <field content="icon" width="5%" />
      <field content="title" width="95%" />
    </fields>
  </windowSwitcher>
```

The layout is unchanged when libsfdo is not linked.

I was a bit afraid that showing icons increases the latency when starting window switcher, but I didn't notice the latency when I opened 15-20 windows and started window swither on my rpi5.

Any ideas for better documentation or default configuration are welcome.


johanmalm:
I like the change in principle.

Suggest the following:

1. Make the removal of `trimmed_identifier` a separate commit.
2. It looks like icon size is equal to font-height (https://github.com/labwc/labwc/blob/2404297f2d6838d49b5b6704ea34149de47b1826/src/osd.c#L356). Consider if we want to hard-code a field width of `5%` for an object that does not change size (for example by ellipsis in the case of text). I tried with `osd.window-switcher.width: 200` and `<font><size>14</size></font>` and found that the icon clashes with the text in the window switcher.
3. Rather than just leaving `title` without an `identifier` variant, consider the config below as a contender (where `%n` is the desktop entry/file application name).

```xml
  <windowSwitcher>
    <allWorkspaces>yes</allWorkspaces>
    <fields>
      <field content="icon" width="5%"/>
      <field content="custom" format="%n - %t" width="95%"/>
    </fields>
  </windowSwitcher>
```

tokyo4j:
> Make the removal of trimmed_identifier a separate commit.

Is your concern that the commit title "osd: show icon in window switcher by default" was missing the context that it removes `type` and `trimmed_identifier`? I updated the title to "osd: update the default window switcher layout" and I hope it helps. I'm kind of reluctant to make a commit just for removal of trimmed_identifier since it will require updating the same part of the documentation twice.

> Consider if we want to hard-code a field width of 5% for an object that does not change size (for example by ellipsis in the case of text). I tried with osd.window-switcher.width: 200 and <font><size>14</size></font> and found that the icon clashes with the text in the window switcher.

#2651 should make the icon shrink (instead of clashing with the text) if the allocated width is smaller than the icon size.

> Rather than just leaving title without an identifier variant, consider the config below as a contender (where %n is the desktop entry/file application name).

I like the addition of desktop entry name, but I'm not a fun of the custom format.

For example my VSCode sets its title like: "rc.xml - labwc - Visual Studio Code". When I use your custom format, it looks like "Visual Studio Code - rc.xml - labwc -Visual Studio Code", which breaks the hierarchy that "Visual Studio Code" is opening the project "labwc" and the file "rc.xml" in it.

How about configuration like this?

```xml
  <windowSwitcher>
    <fields>
      <field content="icon" width="5%"/>
      <field content="desktop_entry_name" width="30%"/>
      <field content="title" width="65%"/>
    </fields>
  </windowSwitcher>
```

tokyo4j:
I've just saw https://github.com/labwc/labwc/discussions/2607#discussioncomment-12681287 - Let's see if localized app name can be used instead of `desktop_entry_name` first.

johanmalm:
> > Make the removal of trimmed_identifier a separate commit.
> 
> Is your concern that the commit title "osd: show icon in window switcher by default" was missing the context that it removes `type` and `trimmed_identifier`? I updated the title to "osd: update the default window switcher layout" and I hope it helps. I'm kind of reluctant to make a commit just for removal of trimmed_identifier since it will require updating the same part of the documentation twice.

I'm fine either way.

> > Consider if we want to hard-code a field width of 5% for an object that does not change size (for example by ellipsis in the case of text). I tried with osd.window-switcher.width: 200 and 14 and found that the icon clashes with the text in the window switcher.
> 
> #2651 should make the icon shrink (instead of clashing with the text) if the allocated width is smaller than the icon size.

Great.

> How about configuration like this?
> 
> ```
>   <windowSwitcher>
>     <fields>
>       <field content="icon" width="5%"/>
>       <field content="desktop_entry_name" width="30%"/>
>       <field content="title" width="65%"/>
>     </fields>
>   </windowSwitcher>
> ```

Yes - that is better! I like that. I had forgotten that we had `desktop_entry_name`.

tokyo4j:
OK, I updated the default layout as I suggested. I also updated the default layout without `libsfdo` to:
```xml
      <field content="desktop_entry_name" width="30%"/>
      <field content="title" width="70%"/>
```
for more consistency.

I don't have a strong preference on the widths of the entries. For applications that I use daily, desktop entry names fit in the width of `30%` without ellipsis. However, some names like `Advanced Network Configuration` or localized names like `Thunar ファイルマネージャー` (with the patch in #2653 applied) are shortened with ellipsis.

johanmalm:
Thanks. Looks great.

PR #2649, opened 2025-03-31
docs: labwc-actions.5.scd style updates

domo141:
- some newlines to dedent and separate content
- comma (,) before respectively
- reflowd some lines to fit in 80 columns

--- end of commit message --

to see how the content looks before this commit, execute

`scdoc < docs/labwc-actions.5.scd| nroff -mandoc -rLL=79n -rLT=79n -Tutf8 | less`

and then search (/) for content nearby ( AutoP, NextW, warp (skip a few first), HideC, EnableT)

(and finally, searched in the Net, respectively, is with leading comma (,) and either trailing comma 
or period -- I just had a feeling as not being native speaker...)


johanmalm:
Thanks

PR #2651, opened 2025-04-01
osd: add osd.window-switcher.item.icon.size

tokyo4j:
Suggested in https://github.com/labwc/labwc/pull/2621#issuecomment-2727469687.

This allows users to make the icon in window switcher bigger (or smaller) than the font size, which enables more Openbox-like appearance.

Example:
```
  osd.window-switcher.item.icon.size: 50
```

```xml
  <windowSwitcher>
    <fields>
      <field content="icon" width="15%" />
      <field content="title" width="85%" />
    </fields>
  </windowSwitcher>
```

20250401_19h26m10s_grim

The default value is set to `font_height(&rc.font_osd)`, as it is right now. @johanmalm suggested deriving from the icon size in the titlebar (https://github.com/labwc/labwc/pull/2621#issuecomment-2727469687), but I personally don't think it's intuitive that changing the button size in titlebar changes the icon size in window switcher. This is not a strong preference, though.

This PR also makes the icon smaller than the font size when the width allocated with `<windowSwitcher><fields><field width="">` is smaller than that.

The first commit fixes the placement of workspace name at the top of window switcher. Previously we didn't take into account `theme->osd_window_switcher_item_padding_y` when centering the workspace name. This problem seems to have been around since the introduction of item paddings.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -256,6 +256,11 @@ all are supported.
 	Border width of the selection box in the window switcher in pixels.
 	Default is 2.
 
+*osd.window-switcher.item.icon.size*
+	Size of the icon in window switcher, in pixels.
+	If not set, the font size derived from <theme><font place="OnScreenDisplay">
+	is used.
+
 *osd.window-switcher.preview.border.width*
 	Border width of the outlines shown as the preview of the window selected
 	by window switcher. Inherits *osd.border.width* if not set.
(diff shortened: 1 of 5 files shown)


johanmalm:
Thanks. This is great.

LGTM. Not tested.


> The default value is set to `font_height(&rc.font_osd)`, as it is right now. @johanmalm suggested deriving from the icon size in the titlebar ([#2621 (comment)](https://github.com/labwc/labwc/pull/2621#issuecomment-2727469687)), but I personally don't think it's intuitive that changing the button size in titlebar changes the icon size in window switcher. This is not a strong preference, though.

I think you're right - following font height is more intuitive

PR #2652, opened 2025-04-01
cursor: let followMouse update focus only when cursor enters a surface

### LINKED ISSUE #2650: followMouse & followMouseRequiresMovement has a horrible experience when using keyboard to focus
### Problem description

When using the keyboard to switch focus (eg. alt-tab), the cursor may remain over an existing window.. a tiny bump of the mouse (or table upon which it rests) will result in focus being immediately changed to the window under the mouse.

A better (and more common) approach is to require the mouse to actually enter a window... which in the example above would require it to exit the window first.

Note: There also seems to be a bug related to followMouse=yes; followMouseRequiresMovement=no (the combination of which I tried while debugging this) but I suppose that this is covered by #905.

### Steps to reproduce

1. followMouse=yes; followMouseRequiresMovement=yes
2. create two windows (eg. one on the left of the screen, one on the right of the screen)
3. move the mouse to the right window
4. alt-tab to focus the left window
5. bump mouse slightly while typing in the left window
6. result: focus has moved to the right window

### labwc build source

Local build

### labwc version

master / 0.8.3-gitref

### labwc environment

### Distribution

Arch Linux

tokyo4j:
Fixes #2650.

This needs more testing and maybe this can be implemented in a cleaner way.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -599,11 +599,6 @@ cursor_process_motion(struct server *server, uint32_t time, double *sx, double *
 		dnd_icons_move(seat, seat->cursor->x, seat->cursor->y);
 	}
 
-	if ((ctx.view || ctx.surface) && rc.focus_follow_mouse) {
-		desktop_focus_view_or_surface(seat, ctx.view, ctx.surface,
-			rc.raise_on_focus);
-	}
-
 	struct mousebind *mousebind;
 	wl_list_for_each(mousebind, &rc.mousebinds, link) {
 		if (mousebind->mouse_event == MOUSE_ACTION_DRAG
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM.

I don't mind the old/new thing. Seems clean enough.

tokyo4j:
I noticed that we should update keyboard focus when entering a window including SSD rather than a surface... I'll make a follow-up PR later.

johanmalm:
Okay. Should be surface || view?
I did ponder that when reviewing. Silly me - should have said something.
Thanks for catching.

tokyo4j:
> Should be surface || view?
> I did ponder that when reviewing. Silly me - should have said something.

Yeah... I was kind of confused when writing this patch and I overlooked such a obvious difference. I admit, I'm not smart enough to understand how focus management works.

I'm now wondering how we should remember the cursor context of a previous motion handler call, which is necessary to detect whether the cursor is "entering" a window. It will take me some time to determine what's the best approach...

tokyo4j:
> Suggest trying the kwin approach if it simplifes things.

Just for clarification, labwc in the master branch currently have the same behavior as kwin that updates keyboard focus only when the cursor enters the content of a window. If we decide to follow kwin's behavior, all we have to do for the next release is just to describe the change in `NEWS.md` like:

```
CHANGED
- If <focus><followMouse> is set to yes, the focus is now updated only when the cursor enters
  a window content, not when the cursor moves within the window content, or when the cursor
  enters a window's titlebar (#2652).
```

ahesford:
I just ran into this behavior when I shaded a window and de-focused it, and found it impossible to refocus with the mouse. Because there is no surface to enter on a shaded window, you can only focus by clicking the titlebar.

In general, losing the ability to focus a window by mousing over the titlebar feels broken, but I don't know how long I'll put up with it before I decide to figure out a proper fix. This is one of those "we should probably fix this some day" things.

PR #2657, opened 2025-04-06
osd: workspace switcher boxes improvements

czkz:
Adds a theme option to configure border width of workspace boxes in osd.

Previews:
- master defaults
- new defaults
- example osd.workspace-switcher.boxes.border.width: 8

Note that master boxes are not square and 1px off-center (blue line).

preview

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -281,6 +281,9 @@ all are supported.
 	Height of boxes in workspace switcher in pixels. Setting to 0 disables
 	boxes. Default is 20.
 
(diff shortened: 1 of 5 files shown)


johanmalm:
Thanks for this contributions, and apologies that it took a few days to get to it.

Looks fine except in the first commit message, should we say `width` instead of `height` and `22x20` instead of `20x22`?

czkz:
> should we say `width` instead of `height` and `22x20` instead of `20x22`?

Not sure I understand why - you can see in the first image attached, the box width (including the black borders) is 20px, and height is 22px.

Issue #2391, opened 2024-11-28
XML attributes are forced to be ordered


RicArch97:
### Problem description

I wrote a stylesheet using XSLT for converting from the Nix language to XML to autogenerate the rc.xml from Nix.

However i ran into an ordering problem when defining keybinds:
```
00:00:00.000 [../src/config/rcxml.c:559] expect <action name=""> element first. nodename: 'command.action' content: '/nix/store/nkjjc70cidfc3wbpas7kqr7agasma59c-rofi-1.7.5+wayland3/bin/rofi -show drun -config $XDG_CONFIG_HOME/rofi/menu.rasi'
00:00:00.000 [../src/action.c:591] Missing required argument for Execute: command
00:00:00.000 [../src/config/rcxml.c:1759] Removed invalid keybind action
```

Note that the XML specification (https://www.w3.org/TR/REC-xml/#sec-starttags) states:
> Note that the order of attribute specifications in a start-tag or empty-element tag is not significant.

Which kind of makes sense, they are named attributes, why would order matter?

Edit: it doesn't seem to be the case for every setting. For example it tried it with margin:
```
<margin bottom="10" left="10" right="10" top="10"/>
```
It's a different order than how it's defined in the man page, but I don't see any errors because of it. So it's somewhat inconsistent across the options.

### Steps to reproduce

1. Define a keybind where the attributes are not in the same order as they appear in the man page.

### labwc build source

Release

### labwc version

0.8.1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

NixOS

Consolatis:
Yes, actions in specific require the `name` attribute to come first. I am also not that happy about it and we should figure out how to make the argument ordering irrelevant.

The main issue with actions in specific is that we do create the action struct with the name argument. Maybe we could have a generic `action_create(void)` and provide some additional `action_set_name(const char *name)` instead. Or make `action_arg_from_xml_node()` also handle the name (+ potentially rename the function). It might require some more refactoring though because that function also does argument verification based on the action type (which in turn is set by the `name` tag / child-node). Or we need a temporary struct which holds all the XML tags / child-nodes before creating the action struct.

PR #2665, opened 2025-04-10
rcxml: convert dotted properties into nested nodes before processing

tokyo4j:
For example, the following node:

```xml
  <keybind name.action="ShowMenu" menu.action="root-menu"
            x.position.action="1" y.position.action="2" />
```
is converted to:
```xml
 <keybind>
   <action>
     <name>ShowMenu</name>
     <menu>root-menu</menu>
     <position>
       <x>1</x>
       <y>2</y>
     </position>
   </action>
 </keybind>
```
...before processing the entire xml tree.

The long-term plan I'm thinking of is:
1. Merge this PR.
2. Remove `current_inline_action` from #2427 and merge it.
3. Refactor the entire `rcxml.c` (perhaps `menu.c` and `action.c` too) to use recursion rather encoding each node. This will need careful testing and reviewing, but I think we should do this before 1.0 as it will make the code a lot easier to follow and solve #2391 ~and https://github.com/labwc/labwc/pull/2603#issuecomment-2708643312.~

### DIFF
--- a/include/common/xml.h
+++ b/include/common/xml.h
@@ -5,7 +5,7 @@
 #include <libxml/tree.h>
 
 /*
- * Converts dotted properties into nested nodes.
+ * Converts dotted attributes into nested nodes.
  * For example, the following node:
  *
  *   <keybind name.action="ShowMenu" menu.action="root-menu"
@@ -24,6 +24,6 @@
  *     </action>
  *   </keybind>
  */
-void lab_xml_expand_dotted_props(xmlNode *root);
+void lab_xml_expand_dotted_attributes(xmlNode *root);
 
 #endif /* LABWC_XML_H */
--- a/include/common/xml.h
+++ b/include/common/xml.h
@@ -1,0 +1,29 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_XML_H
+#define LABWC_XML_H
+
+#include <libxml/tree.h>
+
+/*
+ * Converts dotted properties into nested nodes.
+ * For example, the following node:
+ *
+ *   <keybind name.action="ShowMenu" menu.action="root-menu"
+ *            x.position.action="1" y.position.action="2" />
+ *
+ * is converted to:
+ *
+ *   <keybind>
+ *     <action>
+ *       <name>ShowMenu</name>
+ *       <menu>root-menu</menu>
+ *       <position>
+ *         <x>1</x>
+ *         <y>2</y>
+ *       </position>
+ *     </action>
+ *   </keybind>
+ */
+void lab_xml_expand_dotted_props(xmlNode *root);
+
+#endif /* LABWC_XML_H */
--- a/src/common/meson.build
+++ b/src/common/meson.build
@@ -23,4 +23,5 @@ labwc_sources += files(
   'surface-helpers.c',
   'spawn.c',
   'string-helpers.c',
+  'xml.c',
 )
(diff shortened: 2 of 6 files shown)


johanmalm:
Thanks for working on this. It is no small task.

Have not looked through detail as still just on phone. Just a few questions in the meantime:

1. By "recursive" are you referring to the functions with the prefix fill_ ? The xml_tree_walk/traverse functions are already recursive. Or do you have a different approach in mind for those too?
2. Should we consider merging this after the next version-bump to give it more testing?
3. In order to go the full way with recursive parsing, would we not have to also convert non-dotted attributes to elements too (to stay element/attribute agnostic).
4. I think we ought to consider changing "property"/"prop" to "attribute". I might be mistaken here, but I think libxml2 unhelpfully mix the terms.

tokyo4j:
> By "recursive" are you referring to the functions with the prefix fill_ ? The xml_tree_walk/traverse functions are already recursive. Or do you have a different approach in mind for those too?

Sorry, you are correct. "Recursive" was a wrong word. I was thinking of something like this:
```c

static void
fill_keybind(xmlNode *keybind_node)
{
	char *key;
	struct keybind *keybind;
	bool layout_dependent;
	bool on_release;

	if (lab_xml_get_string(keybind_node, "key", &key)) {
		keybind = keybind_create(key);
	} else {
		return;
	}
	if (lab_xml_get_bool(keybind_node, "layoutDependent", &layout_dependent)) {
		keybind->use_syms_only = layout_dependent;
	}
	if (lab_xml_get_bool(keybind_node, "onRelease", &on_release)) {
		keybind->on_release = on_release;
	}

	for (xmlNode *node = keybind_node->children; node; node = node->next) {
		if (!strcasecmp((char *)node->name, "action")) {
			struct action *action = parse_action(node);
			if (action) {
				wl_list_append(&keybind->actions, &action->link);
			}
		}
	}
}
```

> In order to go the full way with recursive parsing, would we not have to also convert non-dotted attributes to elements too (to stay element/attribute agnostic).

I converted only dotted attributes because I was a bit afraid of changing code paths in `rcxml.c` for users not using dotted attributes at all. But I agree with your opinion. If we fully refactor `rcxml.c` like above, converting all of the attributes will make things a lot easier.

> Should we consider merging this after the next version-bump to give it more testing?

I initially thought we can merge this PR before the next release as it would not affect users not using dotted attributes, but now I think we should also convert non-dotted attributes and do more testing.

One thing I'm concerning is that we're parsing 'action' attribute in `<mousebind />` and `<action />` differently. I'm not sure this PR & #2665 don't break it if we also convert non-dotted attributes.

orfeasxyz:
I don't mind the delay. Also, since I enjoy working on this part of the codebase, I could also take on the refactoring of rcxml. I suggest opening a meta issue to track all the changes we must do. On top of my head:

1) Convert attributes to elements
2) Make element parsing, order independent
3) Make it possible to dry run the config and not fallback to default values when parsing fails.
4) Support nested "If" actions.

tokyo4j:
> I suggest opening a meta issue to track all the changes we must do.

Great idea! Opened #2666.

tokyo4j:
Superseded by #2667

johanmalm:
> I don't mind the delay. Also, since I enjoy working on this part of the codebase, I could also take on the refactoring of rcxml. I suggest opening a meta issue to track all the changes we must do. On top of my head:
> 
> 1. Convert attributes to elements
> 2. Make element parsing, order independent
> 3. Make it possible to dry run the config and not fallback to default values when parsing fails.
> 4. Support nested "If" actions.

Looks like @tokyo4j  is deep into this work already. Could you discuss with them directly on ways to contribute. It might be the peer reviews is the best way atm.

PR #2667, opened 2025-04-12
rcxml: big refactor

tokyo4j:
- Remove state-machine variables in `rcxml.c`
- Convert all the attributes into nested/non-nested nodes so that we can just traverse `node->children`. This should also fix https://github.com/labwc/labwc/pull/2603#issuecomment-2708643312
- Support nested `If` actions
- Support `CDATA`
- Remove ordering constraint of attributes in `<keybind>`, `<mousebind>` and `<windowRule>` (#2391)

TODO:
- [ ] Do testing
- [ ] Decide how we should merge this giant patch. Perhaps we can create another branch in labwc repo, merge my patches above to it one-by-one, and eventually merge it to the master branch.
- [ ] Decide whether we should go with `LAB_XML_FOR_EACH()` macro. `scripts/check` is not happy with that.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -432,112 +432,141 @@ fill_region(char *nodename, char *content, struct parser_state *state)
 }
 
 static void
-fill_action_query(char *nodename, char *content, struct action *action, struct parser_state *state)
+fill_action_query(struct action *action, xmlNode *node, struct view_query *query)
 {
-	if (!action) {
-		wlr_log(WLR_ERROR, "No parent action for query: %s=%s", nodename, content);
-		return;
-	}
-
-	string_truncate_at_pattern(nodename, ".keybind.keyboard");
-	string_truncate_at_pattern(nodename, ".mousebind.context.mouse");
-
-	if (!strcasecmp(nodename, "query.action")) {
-		state->current_view_query = NULL;
-	}
-
-	string_truncate_at_pattern(nodename, ".query.action");
-
-	if (!content) {
-		return;
-	}
-
-	if (!state->current_view_query) {
-		struct wl_list *queries = action_get_querylist(action, "query");
-		if (!queries) {
-			action_arg_add_querylist(action, "query");
-			queries = action_get_querylist(action, "query");
+	xmlNode *child;
+	char *key, *content;
+	LAB_XML_FOR_EACH(node, child, key, content) {
+		if (!strcasecmp(key, "identifier")) {
+			xstrdup_replace(query->identifier, content);
+		} else if (!strcasecmp(key, "title")) {
+			xstrdup_replace(query->title, content);
+		} else if (!strcmp(key, "type")) {
+			query->window_type = parse_window_type(content);
+		} else if (!strcasecmp(key, "sandboxEngine")) {
+			xstrdup_replace(query->sandbox_engine, content);
+		} else if (!strcasecmp(key, "sandboxAppId")) {
+			xstrdup_replace(query->sandbox_app_id, content);
+		} else if (!strcasecmp(key, "shaded")) {
+			query->shaded = parse_three_state(content);
+		} else if (!strcasecmp(key, "maximized")) {
+			query->maximized = view_axis_parse(content);
+		} else if (!strcasecmp(key, "iconified")) {
+			query->iconified = parse_three_state(content);
+		} else if (!strcasecmp(key, "focused")) {
+			query->focused = parse_three_state(content);
+		} else if (!strcasecmp(key, "omnipresent")) {
+			query->omnipresent = parse_three_state(content);
+		} else if (!strcasecmp(key, "tiled")) {
+			query->tiled = view_edge_parse(content);
+		} else if (!strcasecmp(key, "tiled_region")) {
+			xstrdup_replace(query->tiled_region, content);
+		} else if (!strcasecmp(key, "desktop")) {
+			xstrdup_replace(query->desktop, content);
+		} else if (!strcasecmp(key, "decoration")) {
+			query->decoration = ssd_mode_parse(content);
+		} else if (!strcasecmp(key, "monitor")) {
+			xstrdup_replace(query->monitor, content);
 		}
-		state->current_view_query = view_query_create();
-		wl_list_append(queries, &state->current_view_query->link);
-	}
-
-	if (!strcasecmp(nodename, "identifier")) {
-		xstrdup_replace(state->current_view_query->identifier, content);
-	} else if (!strcasecmp(nodename, "title")) {
-		xstrdup_replace(state->current_view_query->title, content);
-	} else if (!strcmp(nodename, "type")) {
-		state->current_view_query->window_type = parse_window_type(content);
-	} else if (!strcasecmp(nodename, "sandboxEngine")) {
-		xstrdup_replace(state->current_view_query->sandbox_engine, content);
-	} else if (!strcasecmp(nodename, "sandboxAppId")) {
-		xstrdup_replace(state->current_view_query->sandbox_app_id, content);
-	} else if (!strcasecmp(nodename, "shaded")) {
-		state->current_view_query->shaded = parse_three_state(content);
-	} else if (!strcasecmp(nodename, "maximized")) {
-		state->current_view_query->maximized = view_axis_parse(content);
-	} else if (!strcasecmp(nodename, "iconified")) {
-		state->current_view_query->iconified = parse_three_state(content);
-	} else if (!strcasecmp(nodename, "focused")) {
-		state->current_view_query->focused = parse_three_state(content);
-	} else if (!strcasecmp(nodename, "omnipresent")) {
-		state->current_view_query->omnipresent = parse_three_state(content);
-	} else if (!strcasecmp(nodename, "tiled")) {
-		state->current_view_query->tiled = view_edge_parse(content);
-	} else if (!strcasecmp(nodename, "tiled_region")) {
-		xstrdup_replace(state->current_view_query->tiled_region, content);
-	} else if (!strcasecmp(nodename, "desktop")) {
-		xstrdup_replace(state->current_view_query->desktop, content);
-	} else if (!strcasecmp(nodename, "decoration")) {
-		state->current_view_query->decoration = ssd_mode_parse(content);
-	} else if (!strcasecmp(nodename, "monitor")) {
-		xstrdup_replace(state->current_view_query->monitor, content);
 	}
 }
 
+static void append_actions(xmlNode *node, struct wl_list *list);
+
 static void
-fill_child_action(char *nodename, char *content, struct action *parent,
-	const char *branch_name, struct parser_state *state)
+parse_action_args(xmlNode *node, struct action *action)
 {
-	if (!parent) {
-		wlr_log(WLR_ERROR, "No parent action for branch: %s=%s", nodename, content);
-		return;
-	}
-
-	string_truncate_at_pattern(nodename, ".keybind.keyboard");
-	string_truncate_at_pattern(nodename, ".mousebind.context.mouse");
-	string_truncate_at_pattern(nodename, ".then.action");
-	string_truncate_at_pattern(nodename, ".else.action");
-	string_truncate_at_pattern(nodename, ".none.action");
-
-	if (!strcasecmp(nodename, "action")) {
-		state->current_child_action = NULL;
-	}
-
-	if (!content) {
-		return;
-	}
-
-	struct wl_list *siblings = action_get_actionlist(parent, branch_name);
-	if (!siblings) {
-		action_arg_add_actionlist(parent, branch_name);
-		siblings = action_get_actionlist(parent, branch_name);
-	}
-
-	if (!strcasecmp(nodename, "name.action")) {
-		if (!strcasecmp(content, "If") || !strcasecmp(content, "ForEach")) {
-			wlr_log(WLR_ERROR, "action '%s' cannot be a child action", content);
-			return;
+	xmlNode *child;
+	char *key, *content;
+	LAB_XML_FOR_EACH(node, child, key, content) {
+		if (!strcasecmp(key, "query")) {
+			struct wl_list *querylist =
+				action_get_querylist(action, "query");
+			if (!querylist) {
+				action_arg_add_querylist(action, "query");
+				querylist = action_get_querylist(action, "query");
+			}
+			struct view_query *query = view_query_create();
+			fill_action_query(action, child, query);
+			wl_list_append(querylist, &query->link);
+		} else if (!strcasecmp(key, "then")) {
+			struct wl_list *actions =
+				action_get_actionlist(action, "then");
+			if (!actions) {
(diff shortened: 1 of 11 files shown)


tokyo4j:
>  Find another approach alternative to LAB_XML_FOR_EACH(). scripts/check is not happy with that.

Perhaps we can do something like:
```c


static void
handle_keybind_node(xmlNode *node, const char *content)
{
	struct keybind *keybind = NULL;
	char keyname[256];
	bool on_release;
	bool layout_dependent;
	bool allow_when_locked;

	if (lab_xml_get_string(node, "key", keyname, sizeof(keyname))) {
		keybind = keybind_create(keyname);

...
	if (lab_xml_get_bool(node, "allowWhenLocked", &allow_when_locked)) {
		keybind->allow_when_locked = allow_when_locked;
	}

	append_actions(node, &keybind->actions);
}

static void handle_decoration_node(xmlNode *node, const char *content)
{
	if (!strcmp(content, "client")) {
		rc.xdg_shell_server_side_deco = false;
	} else {
		rc.xdg_shell_server_side_deco = true;
	}
}

struct config_def {
	const char *nodename;
	void (*handle_node)(xmlNode *node, const char *content);
} config_defs[] = { {
	.nodename = "keybind.keyboard",
	.handle_node = handle_keybind_node,
}, {
	.nodename = "decoration.core",
	.handle_node = handle_decoration_node,
} };
``` 

...but this requires definitions of every config entry, which looks very verbose.

At the moment, I prefer sticking to using macros. For CI errors, we can just suppress them or change the macro definition to something like:

```c
        LAB_XML_FOR_EACH(node, child, key, content, {
		if (!strcasecmp(key, "deviceName")) {
			xstrdup_replace(touch_config->device_name, content);
		} else if (!strcasecmp(key, "mapToOutput")) {
			xstrdup_replace(touch_config->output_name, content);
		} else if (!strcasecmp(key, "mouseEmulation")) {
			set_bool(content, &touch_config->force_mouse_emulation);
		} else {
			wlr_log(WLR_ERROR, "Unexpected data in touch parser: %s=\"%s\"",
				key, content);
		}
        })
```

johanmalm:
In the non-macro version, why do we need the local `allow_when_locked` boolean? Couldn't `lab_xml_get_bool()` just not touch the variable rather when returning false?

johanmalm:
I don't really have anything against C macros from a pious, philosophical perspective - but I think there is a bit too much 'hidden' in that for-each macro which is why it doesn't fit the `do { } while(0)` model nor the more traditional usage which holds the `for (some really; awkward and lengthy code; no doubt involving pointer arithmetic and stuff)` but no subsequent lines ending in semi-colon.

tokyo4j:
> In the non-macro version, why do we need the local allow_when_locked boolean? Couldn't lab_xml_get_bool() just not touch the variable rather when returning false?

I suppose your question applies both the code snippet above and the codes in this PR.
Yeah, you're right. I think `lab_xml_get_bool()` returning a boolean can be useful, but at least for `allow_when_locked`, the local variable is not needed at all.

> I don't really have anything against C macros from a pious, philosophical perspective - but I think there is a bit too much 'hidden' in that for-each macro which is why it doesn't fit the do { } while(0) model nor the more traditional usage which holds the for (some really; awkward and lengthy code; no doubt involving pointer arithmetic and stuff) but no subsequent lines ending in semi-colon.

You are right. I've just realized `LAB_XML_FOR_EACH()` macro can be transformed like below. I didn't notice `xmlFree(content)` can be put in the third statement in for (;;) :sweat_smile:

```c
static inline xmlNode *
lab_xml_get_next_child(xmlNode *child)
{
	if (!child) {
		return NULL;
	}
	do {
		child = child->next;
	} while (child && child->type != XML_ELEMENT_NODE);

	return child;
}

static inline void
lab_xml_get_key_and_content(xmlNode *node, char **name, char **content)
{
	if (node) {
		*name = (char *)node->name;
		*content = (char *)xmlNodeGetContent(node);
	}
}

#define LAB_XML_FOR_EACH(parent, child, key, content) \
	for ((child) = (parent)->children, \
		lab_xml_get_key_and_content((child), &(key), &(content)); \
		(child); \
		xmlFree((xmlChar *)(content)), \
		(child) = lab_xml_get_next_child(child), \
		lab_xml_get_key_and_content((child), &(key), &(content)))
```

tokyo4j:
Let me raise another consideration. This PR includes the following code:

<details><summary>code</summary>

```c
	char *key, *content;
	xmlNode *child;
	LAB_XML_FOR_EACH(node, child, key, content) {
		/* Criteria */
		if (!strcmp(key, "identifier")) {
			xstrdup_replace(window_rule->identifier, content);
		} else if (!strcmp(key, "title")) {
			xstrdup_replace(window_rule->title, content);
		} else if (!strcmp(key, "type")) {
			window_rule->window_type = parse_window_type(content);
		} else if (!strcasecmp(key, "matchOnce")) {
			set_bool(content, &window_rule->match_once);
		} else if (!strcasecmp(key, "sandboxEngine")) {
			xstrdup_replace(window_rule->sandbox_engine, content);
		} else if (!strcasecmp(key, "sandboxAppId")) {
			xstrdup_replace(window_rule->sandbox_app_id, content);

		/* Event */
		} else if (!strcmp(key, "event")) {
			/*
			 * This is just in readiness for adding any other types of
			 * events in the future. We default to onFirstMap anyway.
			 */
			if (!strcasecmp(content, "onFirstMap")) {
				window_rule->event = LAB_WINDOW_RULE_EVENT_ON_FIRST_MAP;
			}

		/* Properties */
		} else if (!strcasecmp(key, "serverDecoration")) {
			set_property(content, &window_rule->server_decoration);
		} else if (!strcasecmp(key, "skipTaskbar")) {
			set_property(content, &window_rule->skip_taskbar);
		} else if (!strcasecmp(key, "skipWindowSwitcher")) {
			set_property(content, &window_rule->skip_window_switcher);
		} else if (!strcasecmp(key, "ignoreFocusRequest")) {
			set_property(content, &window_rule->ignore_focus_request);
		} else if (!strcasecmp(key, "ignoreConfigureRequest")) {
			set_property(content, &window_rule->ignore_configure_request);
		} else if (!strcasecmp(key, "fixedPosition")) {
			set_property(content, &window_rule->fixed_position);
		}
	}
```

</details>

And this could also be written like this:

<details><summary>code</summary>

```c
	char buf[256];

	/* Criteria */
	if (!lab_xml_get_string(node, "identifier", buf, sizeof(buf))) {
		xstrdup_replace(window_rule->identifier, buf);
	}
	if (!lab_xml_get_string(node, "title", buf, sizeof(buf))) {
		xstrdup_replace(window_rule->title, buf);
	}
	if (!lab_xml_get_string(node, "type", buf, sizeof(buf))) {
		window_rule->window_type = parse_window_type(buf);
	}
	if (!lab_xml_get_string(node, "matchOnce", buf, sizeof(buf))) {
		set_bool(buf, &window_rule->match_once);
	}
	if (!lab_xml_get_string(node, "sandboxEngine", buf, sizeof(buf))) {
		xstrdup_replace(window_rule->sandbox_engine, buf);
	}
	if (!lab_xml_get_string(node, "sandboxAppId", buf, sizeof(buf))) {
		xstrdup_replace(window_rule->sandbox_app_id, buf);
	}
	/* Event */
	if (!lab_xml_get_string(node, "event", buf, sizeof(buf))) {
		/*
		 * This is just in readiness for adding any other types of
		 * events in the future. We default to onFirstMap anyway.
		 */
		if (!strcasecmp(buf, "onFirstMap")) {
			window_rule->event = LAB_WINDOW_RULE_EVENT_ON_FIRST_MAP;
		}
	}
	/* Properties */
	if (!lab_xml_get_string(node, "serverDecoration", buf, sizeof(buf))) {
		set_property(buf, &window_rule->server_decoration);
	}
	if (!lab_xml_get_string(node, "skipTaskbar", buf, sizeof(buf))) {
		set_property(buf, &window_rule->skip_taskbar);
	}
	if (!lab_xml_get_string(node, "skipWindowSwitcher", buf, sizeof(buf))) {
		set_property(buf, &window_rule->skip_window_switcher);
	}
	if (!lab_xml_get_string(node, "ignoreFocusRequest", buf, sizeof(buf))) {
		set_property(buf, &window_rule->ignore_focus_request);
	}
	if (!lab_xml_get_string(node, "ignoreConfigureRequest", buf, sizeof(buf))) {
		set_property(buf, &window_rule->ignore_configure_request);
	}
	if (!lab_xml_get_string(node, "fixedPosition", buf, sizeof(buf))) {
		set_property(buf, &window_rule->fixed_position);
	}
```

</details>

This eliminates an indentation, but this is involves more lines of codes and is slower since `lab_xml_get_string()` iterates over all the children of the node. So I chose the former approach.

orfeasxyz:
@tokyo4j
I personally prefer the macro approach for the first snippet and the iterative approach for the second one.

tokyo4j:
>  I personally prefer the macro approach for the first snippet and the iterative approach for the second one.

Sorry, I'm confused. What do you mean by "macro approach", "iterative approach" and "snippet"?

orfeasxyz:
Sorry I wrote this in hurry and wasn't very clear. The [first discussion](https://github.com/labwc/labwc/pull/2667#issuecomment-2798707681) was about using macros vs defining callback functions for every config entry. I would stick with macros because they are less verbose.

For the [second discussion](https://github.com/labwc/labwc/pull/2667#issuecomment-2800638848) I prefer iterating through every child node for:
1) Avoiding re-iterating every time with `lab_xml_get_*`
2) Adding an `else` block at the end and throwing a warning for all incorrectly passed child nodes (e.g. <core> cannot have a <placement> child node).

tokyo4j:
Thanks for clarification.

> Adding an else block at the end and throwing a warning for all incorrectly passed child nodes (e.g. cannot have a child node).

Good catch. In fact, `fill_region()` this PR shows warning for missing `%` in `<region>` arguments, inside `LAB_XML_FOR_EACH()`.

orfeasxyz:
@tokyo4j I've been quite inactive on this one, if you have anything ready for review you may ping me.

tokyo4j:
> @tokyo4j I've been quite inactive on this one, if you have anything ready for review you may ping me.

Thanks. I resolved the conflicts and made some minor fixes. This PR is now ready. I appreciate if you review it.

johanmalm:
@tokyo4j - might it make sense to merge this one after `0.9.0` to maximize user feedback. I just feel this on has significant risk of subtle regressions.

orfeasxyz:
I have some time during this weekend, so I will take a look.

Consolatis:
I just quickly scrolled through the diff so can't really say anything about the PR itself.
One thing I noticed: if we rewrite rcxml.c anyway, this seems like a good opportunity to also get rid of all the `!strcasecmp()` and `!strcmp()` calls. I really don't like
- these two being mixed
- using `strcmp()`with a value using camelCase is a potential reason for bugs (and we've run into that before)
- the need to check for its return value of `0` meaning `true` is also weird to read

 I thought about something like:
```c
static inline bool
str_matches(const char *s1, const char *s2)
{
	return strcasecmp(s1, s2) == 0;
}
```
Or as macro variant
```c
#define str_matches(s1, s2) (strcasecmp((s1), (s2)) == 0)
```

tokyo4j:
> @tokyo4j - might it make sense to merge this one after 0.9.0 to maximize user feedback. I just feel this on has significant risk of subtle regressions.

Totally agreed. I know this PR is really scary and I often make mistakes. Though it's a bit annoying to resolve conflicts every time any changes are made in `rcxml.c`, reducing the risk of regressions should be prioritized.

@Consolatis I agree with your opinion on `!strcasecmp()`. I will add `str_matches()` function (I prefer inline functions over macros for more type information).

johanmalm:
Agree with fixing the inconsistent use of case sensitive and insensitive.

I'm not bothered about `!strcasecmp()`. I know it's a C oddity, but we know what it means and it's plain and simple. The drawback of using a wrapper is of course that it's not obvious to a new reader what it does. For example, is `str_match()` case-sensitive?

Ref:
- https://github.com/rustyrussell/ccan/blob/ca0940392e3cf17a0537d66ee6db3a4931cd3821/ccan/str/str.h#L21

src/config/rcxml.c
```diff
@@ -365,220 +334,237 @@ clear_window_switcher_fields(void)
...
+	xmlNode *child;
+	char *key, *content;
+	LAB_XML_FOR_EACH(node, child, key, content) {
+		if (strcasecmp(key, "action")) {
+			continue;
 		}
-		state->current_child_action = action_create(content);
-		if (state->current_child_action) {
-			wl_list_append(siblings, &state->current_child_action->link);
+		if (lab_xml_node_is_leaf(child)) {
```
orfeasxyz, line 541: Maybe turn the comment into "Ignore empty actions"

orfeasxyz:
The change looks good overall. From my understanding action parsing is the largest change whereas the rest of the changes are small rewrites thus they will (probably - hopefully!) not cause any regressions. I think we should prioritize this once 0.9.0 so we don't have to do duplicate work as @tokyo4j also said.

tokyo4j:
> I'm not bothered about !strcasecmp(). I know it's a C oddity, but we know what it means and it's plain and simple. The drawback of using a wrapper is of course that it's not obvious to a new reader what it does. For example, is str_match() case-sensitive?

That's a very good point. If we add another function to replace `!strcasecmp(a, b)`, it should be named as something like `str_case_match()` to avoid confusion. But I feel its advantage of removing the odd `!` before `strcasecmp()` is not worth the unclarity for new users.

> I think we should prioritize this once 0.9.0 so we don't have to do duplicate work as @tokyo4j also said.

Yeah, I'll be happy if this PR is merged in the early stage of the next release cycle, as changes to `rcxml.c` always requires me to resolve conflicts. I'll resolve the conflicts, replace some `strcmp()`s to `strcasecmp()`, and un-draft this PR when 0.9.0 is released. I'll review the labnag PR in the meanwhile.

src/config/rcxml.c
tokyo4j: That's not meant to ignore empty actions, but to ignore `action` attributes in `<mousebind>` which is totally different normal actions. For example, `action="Click"` below should be ignore here:
```xml
  <action name="Close" />
</mousebind>
```

AFAIR @Consolatis suggested changing the `action` attribute to `event`, but we will need to keep compatibility anyway.

I updated the comment to be more descriptive. I also added a comment for `lab_xml_node_is_leaf()` in the header.

orfeasxyz: Oh I totally forgot about that case

tokyo4j:
Now ready for review again.

EDIT: I forgot to mention that I didn't replace `strcmp()` with `strcasecmp()`, because we will also need to modify `action.c`, `osd-field.c` etc. Let's do it in follow-up PR.

johanmalm:
Have reviewed again. LGTM.

Have tested with a couple of my config files and it seems to work well.

I've got to say, `rcxml.c` looks so much better now :smile:

orfeasxyz:
Thanks for all the work on this, my base config as well as nested If actions seem to be working just fine.

PR #2669, opened 2025-04-12
rcxml: handle duplicated `<theme><titlebar><layout>` and `<windowSwitcher><fields>` entries

### LINKED ISSUE #2646: Multiple `<theme><titlebar><layout>` nodes in `rc.xml` breaks titlebar
### Problem description

When we have multiple `<theme><titlebar><layout>` in `rc.xml` like below, the titlebar looks broken (button icons are placed on the far left of the window).
```xml
  <theme><titlebar><layout>icon,menu:iconify,max,close</layout></titlebar></theme>
  <theme><titlebar><layout>icon,menu:iconify,max,close</layout></titlebar></theme>
```

Image

The problem is that, `rc.title_buttons_{left,right}` can be filled with duplicated button types by having multiple `<theme><titlebar><layout>`, and `ssd_get_part()` in `ssd-titlebar.c` can select only a single button node per type.

This won't be difficult to fix, but I think we should validate if we are properly handling duplicated nodes for other config entries (e.g. `<windowSwitcher><fields>`), too.

This kind of issue can be critical for Raspberry Pi users, because they launch labwc with `-m` option and the user may try to overwrite `/etc/xdg/labwc/rc.xml` with `~/.config/labwc/rc.xml`, which means labwc loads duplicated nodes across those files. JFYI: @spl237

### Steps to reproduce

1. Have multiple `<theme><titlebar><layout>` (filled with valid contents) in `rc.xml`
2. Launch labwc
3. Open a `alacritty` window.

### labwc build source

Local build

### labwc version

0.8.3-30-g4ef881bb-dirty

### labwc environment

### Distribution

_No response_

tokyo4j:
Fixes #2646.

Also handles duplicated `<windowSwitcher><fields>` entries so that users can overwrite the window switcher layout defined in `/etc/xdg/labwc/rc.xml` with `~/.config/labwc/rc.xml`, without overflowing contents.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -202,8 +202,25 @@ fill_section(const char *content, struct wl_list *list, uint32_t *found_buttons)
 }
 
 static void
+clear_title_layout(void)
+{
+	struct title_button *button, *button_tmp;
+	wl_list_for_each_safe(button, button_tmp, &rc.title_buttons_left, link) {
+		wl_list_remove(&button->link);
+		zfree(button);
+	}
+	wl_list_for_each_safe(button, button_tmp, &rc.title_buttons_right, link) {
+		wl_list_remove(&button->link);
+		zfree(button);
+	}
+	rc.title_layout_loaded = false;
+}
+
+static void
 fill_title_layout(char *content)
 {
+	clear_title_layout();
+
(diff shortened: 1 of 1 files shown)


johanmalm:
That's great. Thanks.

PR #2673, opened 2025-04-14
menu: refactor handle_menu_element() (part 1)

tokyo4j:
This should not logically change anything other than an error message when encountering `<menu>` without `id`.


johanmalm:
Oh, this is a tricky one. Need to read through it when I have a bit more focus. Will try tomorrow.

johanmalm:
LGTM and indeed seems to simplify the code.

Have tested and appears to work fine.

PR #2674, opened 2025-04-14
menu: refactor handle_menu_element() (part 2)

tokyo4j:
Based on #2673.

I believe `!current_menu` is equivalent to `is_toplevel_static_menu_definition()`.

This PR and #2673 are not urgent, so delaying them to after the next release is totally fine.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -42,7 +42,6 @@ static bool in_item;
 static struct menuitem *current_item;
 static struct action *current_item_action;
 
-static int menu_level;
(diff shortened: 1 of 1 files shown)


johanmalm:
Have tested with all menu scenarios I can think of (external, inline, pipemenu at topmost level and otherwise) and it seems to work fine.

Could you rebase it and I'll just read through the code again. It looks like a nice simplification.

johanmalm:
LGTM. Have tested. Really nice simplification.

PR #2676, opened 2025-04-16
view: disallow minimizing unmapped view 

### LINKED ISSUE #2627: A new window is not focused if it has been minimized before being mapped

tokyo4j:
Fixes #2627.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -764,6 +764,11 @@ _minimize(struct view *view, bool minimized)
 	if (view->minimized == minimized) {
 		return;
 	}
+	if (!view->mapped && minimized) {
+		/* xdg_toplevel.set_minimized can be sent before mapping */
+		wlr_log(WLR_DEBUG, "not minimizing unmapped view");
+		return;
+	}
 
 	if (view->impl->minimize) {
 		view->impl->minimize(view, minimized);


jlindgren90:
I can kind of see why the first part makes sense (prohibit minimizing unmapped view). But I don't understand the second part (prohibit unminimizing mapped view). What's the reasoning behind that?

jlindgren90:
Actually I guess I'm not 100% convinced of the first part either. Does xdg-shell protocol forbid minimize before map? If it's allowed by the protocol we should probably try to honor it.

tokyo4j:
> I can kind of see why the first part makes sense (prohibit minimizing unmapped view). But I don't understand the second part (prohibit unminimizing mapped view). What's the reasoning behind that?

Yeah, I changed my mind as soon as I opened this PR and pushed an updated version.

> Actually I guess I'm not 100% convinced of the first part either. Does xdg-shell protocol forbid minimize before map? If it's allowed by the protocol we should probably try to honor it.

I agree. I just didn't want to add another field to `view_xdg_toplevel` like I suggested in https://github.com/labwc/labwc/issues/2627#issuecomment-2807969964 for such a minor use-case, but I've just realized there's already `wlr_xdg_toplevel.requested.minimized`. I'll update my patch to see that field in map handler.

johanmalm:
Should this be closed now?

jlindgren90:
I assume so, it would run contrary to #2688 which was just merged.

PR #2678, opened 2025-04-16
feat: add scroll wheel emulation for cursor motion

jp7677:
Solution for https://github.com/labwc/labwc/discussions/2672

Simple keybinding example:

```xml
<keybind key="A-W-s">
  <action name="ToggleScrollWheelEmulation" />
</keybind>
```

But this should also work for mouse bindings, like enabling emulation on button down and disabling on button up.

Includes https://github.com/labwc/labwc/pull/2704

### DIFF
--- a/include/input/cursor.h
+++ b/include/input/cursor.h
@@ -156,6 +156,10 @@ void cursor_emulate_move_absolute(struct seat *seat,
 		double x, double y, uint32_t time_msec);
 void cursor_emulate_button(struct seat *seat,
 		uint32_t button, enum wl_pointer_button_state state, uint32_t time_msec);
+void cursor_emulate_axis(struct seat *seat,
+		struct wlr_input_device *device,
+		enum wl_pointer_axis orientation, double delta, double delta_discrete,
+		enum wl_pointer_axis_source source, uint32_t time_msec);
 void cursor_finish(struct seat *seat);
 
 #endif /* LABWC_CURSOR_H */
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1195,7 +1195,7 @@ struct scroll_info {
 };
 
 static struct scroll_info
-compare_delta(const struct wlr_pointer_axis_event *event, double *accum)
+compare_delta(double delta, double delta_discrete, double *accum)
 {
 	/*
 	 * Smooth scroll deltas are in surface space, so treating each unit as a
@@ -1210,21 +1210,21 @@ compare_delta(const struct wlr_pointer_axis_event *event, double *accum)
 	const double SCROLL_THRESHOLD = 10.0;
 	struct scroll_info info = {0};
 
-	if (event->delta_discrete < 0 || event->delta < 0) {
+	if (delta_discrete < 0 || delta < 0) {
 		info.direction = -1;
-	} else if (event->delta_discrete > 0 || event->delta > 0) {
+	} else if (delta_discrete > 0 || delta > 0) {
 		info.direction = 1;
 	}
 
-	if (event->delta == 0.0) {
+	if (delta == 0.0) {
 		/* Delta 0 marks the end of a scroll */
 		*accum = 0.0;
 	} else {
 		/* Accumulate smooth scrolling until we hit threshold */
-		*accum += event->delta;
+		*accum += delta;
 	}
 
-	if (event->delta_discrete != 0 || fabs(*accum) > SCROLL_THRESHOLD) {
+	if (delta_discrete != 0 || fabs(*accum) > SCROLL_THRESHOLD) {
 		*accum = fmod(*accum, SCROLL_THRESHOLD);
 		info.run_action = true;
 	}
@@ -1233,25 +1233,28 @@ compare_delta(const struct wlr_pointer_axis_event *event, double *accum)
 }
 
 static bool
-process_cursor_axis(struct server *server, struct cursor_context *ctx,
-		struct wlr_pointer_axis_event *event)
+process_cursor_axis(struct server *server, enum wl_pointer_axis orientation,
+		double delta, double delta_discrete)
 {
-	struct mousebind *mousebind;
-	bool handled = false;
+	struct cursor_context ctx = get_cursor_context(server);
 	uint32_t modifiers = keyboard_get_all_modifiers(&server->seat);
 
 	enum direction direction = LAB_DIRECTION_INVALID;
 	struct scroll_info info = {0};
 
-	if (event->orientation == WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
-		info = compare_delta(event, &server->seat.smooth_scroll_offset.x);
+	if (orientation == WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
+		info = compare_delta(delta, delta_discrete,
+			&server->seat.smooth_scroll_offset.x);
+
 		if (info.direction < 0) {
 			direction = LAB_DIRECTION_LEFT;
 		} else if (info.direction > 0) {
 			direction = LAB_DIRECTION_RIGHT;
 		}
-	} else if (event->orientation == WL_POINTER_AXIS_VERTICAL_SCROLL) {
-		info = compare_delta(event, &server->seat.smooth_scroll_offset.y);
+	} else if (orientation == WL_POINTER_AXIS_VERTICAL_SCROLL) {
+		info = compare_delta(delta, delta_discrete,
+			&server->seat.smooth_scroll_offset.y);
+
 		if (info.direction < 0) {
 			direction = LAB_DIRECTION_UP;
 		} else if (info.direction > 0) {
@@ -1261,27 +1264,36 @@ process_cursor_axis(struct server *server, struct cursor_context *ctx,
 		wlr_log(WLR_DEBUG, "Failed to handle cursor axis event");
 	}
 
-	if (direction == LAB_DIRECTION_INVALID) {
-		return false;
-	}
-
-	wl_list_for_each(mousebind, &rc.mousebinds, link) {
-		if (ssd_part_contains(mousebind->context, ctx->type)
-				&& mousebind->direction == direction
-				&& modifiers == mousebind->modifiers
-				&& mousebind->mouse_event == MOUSE_ACTION_SCROLL) {
-			handled = true;
-			/*
-			 * Action may not be executed if the accumulated scroll
-			 * delta on touchpads doesn't exceed the threshold
-			 */
-			if (info.run_action) {
-				actions_run(ctx->view, server, &mousebind->actions, ctx);
+	bool handled = false;
+	if (direction != LAB_DIRECTION_INVALID) {
+		struct mousebind *mousebind;
+		wl_list_for_each(mousebind, &rc.mousebinds, link) {
+			if (ssd_part_contains(mousebind->context, ctx.type)
+					&& mousebind->direction == direction
+					&& modifiers == mousebind->modifiers
+					&& mousebind->mouse_event == MOUSE_ACTION_SCROLL) {
+				handled = true;
+				/*
+				 * Action may not be executed if the accumulated scroll
+				 * delta on touchpads doesn't exceed the threshold
+				 */
+				if (info.run_action) {
+					actions_run(ctx.view, server, &mousebind->actions, &ctx);
+				}
 			}
 		}
 	}
 
-	return handled;
+	/* Bindings swallow mouse events if activated */
+	if (ctx.surface && !handled) {
+		/* Make sure we are sending the events to the surface under the cursor */
+		double sx, sy;
+		cursor_update_common(server, &ctx, /*cursor_has_moved*/ false, &sx, &sy);
+
+		return true;
+	}
+
+	return false;
 }
 
 static void
@@ -1292,8 +1304,10 @@ handle_axis(struct wl_listener *listener, void *data)
 	 * event, for example when you move the scroll wheel.
 	 */
 	struct seat *seat = wl_container_of(listener, seat, on_cursor.axis);
-	struct wlr_pointer_axis_event *event = data;
 	struct server *server = seat->server;
+	struct wlr_pointer_axis_event *event = data;
+	idle_manager_notify_activity(seat->seat);
+	cursor_set_visible(seat, /* visible */ true);
 
 	/* input->scroll_factor is set for pointer/touch devices */
 	assert(event->pointer->base.type == WLR_INPUT_DEVICE_POINTER
@@ -1301,18 +1315,10 @@ handle_axis(struct wl_listener *listener, void *data)
 	struct input *input = event->pointer->base.data;
 	double scroll_factor = input->scroll_factor;
 
-	struct cursor_context ctx = get_cursor_context(server);
-	idle_manager_notify_activity(seat->seat);
-	cursor_set_visible(seat, /* visible */ true);
+	bool notify = process_cursor_axis(server, event->orientation,
+		event->delta, event->delta_discrete);
 
-	/* Bindings swallow mouse events if activated */
-	bool handled = process_cursor_axis(server, &ctx, event);
-
-	if (ctx.surface && !handled) {
-		/* Make sure we are sending the events to the surface under the cursor */
-		double sx, sy;
-		cursor_update_common(server, &ctx, /*cursor_has_moved*/ false, &sx, &sy);
-
+	if (notify) {
 		/* Notify the client with pointer focus of the axis event. */
 		wlr_seat_pointer_notify_axis(seat->seat, event->time_msec,
 			event->orientation, scroll_factor * event->delta,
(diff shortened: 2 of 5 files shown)


jp7677:
In response to https://github.com/labwc/labwc/discussions/2672#discussioncomment-12870890 : 

Thanks a lot for the extensive report, much more detailed than I expected for an early look! Very nice!

I've pushed a commit which now uses fixed values for the scroll values, based on the hardware i have here. Combining this with a low scroll factor work nicely here, both in firefox and Gnome document viewer (which uses page-wise scrolling). 

The  config I'm using:

```xml
<mouse>
  <default />
  <context name="All">
    <mousebind button="Side" action="Press"><action name="EnableScrollWheelEmulation" /></mousebind>
    <mousebind button="Side" action="Release"><action name="DisableScrollWheelEmulation" /></mousebind>
  </context>
</mouse>

<libinput>
  <device category="non-touch">
    <naturalScroll>yes</naturalScroll>
    <scrollFactor>0.2</scrollFactor>
  </device>
</libinput>
``` 

Could you try this version? 

With regard to "Major issue: mousebind makes whole desktop unusable", did you by chance omitted the `<default />`  element in your mouse binding section which (afaik) would require you to setup the "normal" actions like focus manually?

Robert42:
> Could you try this version?

Awesome! Now I have smooth scrolling in every application I've tested.

It is now in a very usable state.

There are two new rough edges (see below).

> With regard to "Major issue: mousebind makes whole desktop unusable", did you by chance omitted the `<default />` element in your mouse binding section which (afaik) would require you to setup the "normal" actions like focus manually?

Yep that was it. adding `<default/>` did the trick!
"Major issue: mousebind makes whole desktop unusable" was indeed a layer 8 issue ;)
No more problems here.

## Rough Edge: clamped scroll speed

I have the impression that the maximum scroll speed is clamped.

When the scoll speed is not clamped, you can get a trackball to spin by giving it one powerful swing. This way you can scroll very quickly to other parts of a long document.

When it's clamped, you have to rotate the trackball a lot more times to navigate in a large document.

## Rough Edge: Scroll Speed Variation

From program to program the scroll speed is noticably different (for examle I've opened the labwc-config manual in firefox and in chromium).

It's noticable but a minor issue.

Maybe even a non-issue once the scroll speed is not clamped anymore (I could then set the scroll factor a little lower and use a larger spin whenever I want to navigate faster).

I don't know whether it's possible/feasible to allow configuring different scroll speeds per application. I would need to invest a little more time to test, whether I can spot a hidden pattern (For example x11 vs wayland applications. Or those application which were previously scrolled pagewise vs those which were smooth).

Anyway. Thank you for the time you've already invested. It's much appreciated!

jp7677:
> Awesome! Now I have smooth scrolling in every application I've tested.

Very nice, cool to read that!

> Yep that was it. … No more problems here.

Also very nice :)
 
> I have the impression that the maximum scroll speed is clamped.

Yes, that is indeed the case since I used “fixed” steps for each motion event and I discard the actual distance/delta of the mouse event. So the motion sample rate is the clamping factor, at least that’s what I think. Not sure yet if this is the way to go, but let’s see where we end up.

> When the scoll speed is not clamped, you can get a trackball to spin by giving it one powerful swing. This way you can scroll very quickly to other parts of a long document.

I see and I have an idea how to handle this. Fortunately we have the motion delta (which includes mouse acceleration) and also the raw/unaccelerated delta when a motion event triggered. Applying the factor between those to the simulated scroll delta should accelerate the scrolling. That’s the theory though, I haven’t tried yet :)

> From program to program the scroll speed is noticably different (for examle I've opened the labwc-config manual in firefox and in chromium).

Is this also the case with a normal mouse wheel?

> I don't know whether it's possible/feasible to allow configuring different scroll speeds per application.

Unfortunately that is not possible currently since the scroll factor is a per-device setting. 

> Anyway. Thank you for the time you've already invested. It's much appreciated!

You are very welcome and it’s actually also quite fun to get this right. Getting such detailed feedback also certainly helps.

jp7677:
@Robert42, https://github.com/labwc/labwc/pull/2678/commits/06f071a152a605a35843def2ed6dc97f2fefa933 includes acceleration. Could you please retry with that version and report how the spinning goes?

Robert42:
> Is this also the case with a normal mouse wheel?

Yes it does - good catch! I can observe the same behavior in the labwc version that's preinstalled on the raspberry pi OS.  
=> Not related to this feature.

> @Robert42, [06f071a](https://github.com/labwc/labwc/commit/06f071a152a605a35843def2ed6dc97f2fefa933) includes acceleration. Could you please retry with that version and report how the spinning goes?

Awesome! I've just tested it.

1. :x: In the applications that were previously smooth (chromium, firefox) the accelerated scrolling does not work. When I get the ball spinning, the scroll bar wobbles a little, but mostly stays at the same spot. It feels as if a car-wheel was spinning on ice surface without the car getting tracing and moving forward. Good news: When I rotate the ball slowly, the scrolling still works just like before. The new commit did not break anything :+1: . If helpful I could add some `printf`s to relevent lines.
2. :white_check_mark: In the applications that were previously scrolling page-wise (lxterminal, mousepad, inkscape) spinning the wheel works perfectly! I can get the ball spinning and the scrolling accelerates as you would expected. I can get to the beginning/end of long documents with ease. Awesome! :1st_place_medal:

jp7677:
I wonder if there is too much acceleration when you go full spinning ;). 
What happens when you set the mouse acceleration profile to `flat` ( https://labwc.github.io/labwc-config.5.html#libinput )? I'm indeed curious what kind of delta's are reported in your setup, can you log `event->delta_x ` and `event->delta_y ` and post what ranges (min/max) you are seeing?

Robert42:
That's it! It works! :tada: 

In every application I've tried. With and without getting the ball to spin.

johanmalm:
> That's it! It works! 🎉

Great effort!

src/input/cursor.c
```diff
@@ -663,8 +658,7 @@ _cursor_update_focus(struct server *server)
 	}
 
 	double sx, sy;
-	cursor_update_common(server, &ctx, msec(&now),
-		/*cursor_has_moved*/ false, &sx, &sy);
+	cursor_update_common(server, &ctx, /*cursor_has_moved*/ false, &sx, &sy);
```
Consolatis, line 661: Looks like the first two lines (now + clock_gettime) of the function can now be removed.

include/input/cursor.h
```diff
@@ -13,6 +13,9 @@ struct wlr_surface;
 struct wlr_scene_node;
 enum wl_pointer_button_state;
 
+/* Defines the delta of a single step for a mouse wheel */
+#define LABWC_POINTER_AXIS_STEP 15
+
```
Consolatis, line 18: Suggest moving this to the top of `src/input/cursor.c` and using `.. of a single step for mouse wheel emulation`.

Consolatis (commented):
Nice stuff!

Two small nitpicks below.
Also thanks for converting the signals to the macro.

jp7677:
> That's it! It works! 🎉
> 
> In every application I've tried. With and without getting the ball to spin.

Ah very nice! That also means that code can be simplified even more :)

> When scrolling slowly, I get absolute values around 5, with moderate speeds 10 or 20. When I go fast, 120. The maximum absolute value was 360.

oh wow, that is very different than the max 5 or so I saw with my normal mouse and explains why I didn’t encountered the spinning effects.

> Are there more things you want me to try?

All good for now. I’ll probably ask for a review of the documentation once ready since I have a tendency to keep docs briefly and technical :) That said, what is the configured scroll factor you ended up with? I had to use 0.2 for reasonable scrolling speed. If yours is similar, I could add another factor in code which makes the scrolling speed hopefully useable with the default scrolling factor of 1.0, at least for similar setups like ours. Regardless, the docs should refer to tweaking the scrolling factor for personal preference.

> You guys rock!

Thanks a lot and as stated, your high quality feedback contributed a lot, so I’m gladly returning the compliment.

include/input/cursor.h
jp7677: I think I’ll remove this define (but take over the comment) altogether since the value is only used once.

Robert42:
> I’ll probably ask for a review of the documentation once ready since I have a tendency to keep docs briefly and technical :)

Sure, why not. But I can only review whether the text is understandable (I have a blindness for typos ;) )

> That said, what is the configured scroll factor you ended up with? I had to use 0.2 for reasonable scrolling speed. If yours is similar, I could add another factor in code which makes the scrolling speed hopefully useable with the default scrolling factor of 1.0, at least for similar setups like ours.

During testing I've used 0.5. That was uncomfortable fast, but I wanted to keep it consistent across tests to better see the effects of each commit.

I've just tried 0.1, 0.2, 0.25 and 0.3 ad 0.5.
0.2 is actually the most comfortable one. I can scroll with precision when rotating the ball gently. It's not hectic when rotating with a normal speed and still am able to reach the start and and of the document with a single swing.

tokyo4j:
I like the conversion to macros, but I suggest wrapping the cursor listeners like:
```c
	struct {
		struct wl_listener motion;
		struct wl_listener motion_absolute;
		struct wl_listener button;
		struct wl_listener axis;
		struct wl_listener frame;
	} on_cursor;
```

...since `seat->frame` looks ambiguous to me. This follows `workspace->on_{cosmic,ext}`.

src/input/cursor.c
```diff
@@ -848,14 +862,39 @@ handle_motion(struct wl_listener *listener, void *data)
...
 
-	wlr_relative_pointer_manager_v1_send_relative_motion(
-		server->relative_pointer_manager,
-		seat->seat, (uint64_t)event->time_msec * 1000,
-		event->delta_x, event->delta_y, event->unaccel_dx,
-		event->unaccel_dy);
+	if (seat->cursor_scroll_wheel_emulation) {
+		double natural_scroll_factor = get_natural_scroll_factor(&event->pointer->base);
+		struct input *input = event->pointer->base.data;
+		double scroll_factor = input->scroll_factor * natural_scroll_factor;
```
tokyo4j, line 868: nit: I suggest defining `get_configured_scroll_factor()` like:
```c
static double
get_configured_scroll_factor(struct wlr_input_device *device)
{
	/* input->scroll_factor is set for pointer/touch devices */
	assert(device->type == WLR_INPUT_DEVICE_POINTER
		|| device->type == WLR_INPUT_DEVICE_TOUCH);
	struct input *input = device->data;
	return input->scroll_factor;
}
```

and use it like:
```c
double scroll_factor = get_configured_scroll_factor(&event->pointer->base) * get_natural_scroll_factor(&event->pointer->base);
```

jp7677: Thanks! This was needed after all. I had looked up the scroll factor twice, here and in  `cursor_emulate_axis()`. Now the scroll factor is only read once (in `cursor_emulate_axis()`) with a gentle check for the device type and a fallback to the default of `1.0`.

jp7677:
Assuming that I didn't messed up with adjusting to comments and slight refactor, this should be good now, thus ready for review from my side. 

@Robert42 would be nice if you could briefly confirm that everything is still good for you and that the default scroll factor of `1.0` gives reasonable speed.

Robert42:
> @Robert42 would be nice if you could briefly confirm that everything is still good for you and that the default scroll factor of `1.0` gives reasonable speed.

Yup! I can verify: All scrolling is still fine in all tested applications.

(tested with `<scrollFactor>1.0</scrollFactor>`)

tokyo4j:
@jp7677 Could you open a separate PR for the first 2 commits and the relocation of `cursor_emulate_{move,move_absolute,button}()` in the 4th commit? I think it will make our review easier.

tokyo4j:
I think `delta_discrete` passed to `compare_delta()` should be zero for scrolls emulated by cursor motion, since cursor motions should be considered "continuous" (not "discrete") events just like two-finger scrolling on touchpad, unlike 15-degree scrolls on physical mice. Otherwise, mousebinds like below are executed too frequently with emulated scrolls:

```xml
    <context name="all">
      <mousebind direction="W-Down" action="Scroll">
        <action name="ZoomIn" />
      </mousebind>
    </context>
```

So you can either always pass zero to `cursor_emulate_axis()` or remove `delta_discrete` from `cursor_emulate_axis()`.

LGTM other than that. Great work!

jp7677:
> I think `delta_discrete` passed to `compare_delta()` should be zero for scrolls emulated by cursor motion, since cursor motions should be considered continuous (not discrete) events just like two-finger scrolling on touchpad, unlike 15-degree scrolls on physical mice. Otherwise, mousebinds like below are executed too frequently with emulated scrolls:
> 
> ```
>     <context name="all">
>       <mousebind direction="W-Down" action="Scroll">
>         <action name="ZoomIn" />
>       </mousebind>
>     </context>
> ```
> 
> So you can either always pass zero to `cursor_emulate_axis()` or remove `delta_discrete` from `cursor_emulate_axis()`.
> 
> LGTM other than that. Great work!

What about this alternative?

```patch
index 6fd7b2ad..acdb7e68 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1400,7 +1400,7 @@ cursor_emulate_axis(struct seat *seat, struct wlr_input_device *device,
                scroll_factor = input->scroll_factor;
        }
 
-       bool notify = process_cursor_axis(server, orientation, delta, delta_discrete);
+       bool notify = process_cursor_axis(server, orientation, delta, 0);
        if (notify) {
                /* Notify the client with pointer focus of the axis event. */
                wlr_seat_pointer_notify_axis(seat->seat, time_msec,
```

That way `delta_discrete` is indeed zero for `compare_delta()`, but still used for `wlr_seat_pointer_notify_axis()` which is, based on my testing with firefox, still needed for smooth scrolling. See also the discussion above about finding a good  value for delta_discrete for a track ball device.

tokyo4j:
I read the wayland protocol spec and found `WL_POINTER_AXIS_SOURCE_CONTINUOUS` (instead of `WL_POINTER_AXIS_SOURCE_WHEEL`) is suited for scrolls emulated by motion:
> The "continuous" axis source is a device generating events in a continuous coordinate space, but using something other than a finger. One example for this source is button-based scrolling where the vertical motion of a device is converted to scroll events while a button is held down.

So I think adding `source` argument to `cursor_emulate_axis()` like below is a better option.

```diff
diff --git a/include/input/cursor.h b/include/input/cursor.h
index 5ceabf0f..c49606e2 100644
--- a/include/input/cursor.h
+++ b/include/input/cursor.h
@@ -159,7 +159,7 @@ void cursor_emulate_button(struct seat *seat,
 void cursor_emulate_axis(struct seat *seat,
 		struct wlr_input_device *device,
 		enum wl_pointer_axis orientation, double delta, double delta_discrete,
-		uint32_t time_msec);
+		enum wl_pointer_axis_source source, uint32_t time_msec);
 void cursor_finish(struct seat *seat);
 
 #endif /* LABWC_CURSOR_H */
diff --git a/src/input/cursor.c b/src/input/cursor.c
index 6fd7b2ad..8f48c80f 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -885,7 +885,7 @@ handle_motion(struct wl_listener *listener, void *data)
 			orientation,
 			pointer_axis_step * scroll_factor * delta,
 			WLR_POINTER_AXIS_DISCRETE_STEP * scroll_factor * delta,
-			event->time_msec);
+			WL_POINTER_AXIS_SOURCE_CONTINUOUS, event->time_msec);
 	} else {
 		wlr_relative_pointer_manager_v1_send_relative_motion(
 			server->relative_pointer_manager,
@@ -1388,7 +1388,7 @@ handle_frame(struct wl_listener *listener, void *data)
 void
 cursor_emulate_axis(struct seat *seat, struct wlr_input_device *device,
 		enum wl_pointer_axis orientation, double delta, double delta_discrete,
-		uint32_t time_msec)
+		enum wl_pointer_axis_source source, uint32_t time_msec)
 {
 	struct server *server = seat->server;
 	struct input *input = device->data;
@@ -1406,7 +1406,7 @@ cursor_emulate_axis(struct seat *seat, struct wlr_input_device *device,
 		wlr_seat_pointer_notify_axis(seat->seat, time_msec,
 			orientation, scroll_factor * delta,
 			round(scroll_factor * delta_discrete),
-			WL_POINTER_AXIS_SOURCE_WHEEL, WL_POINTER_AXIS_RELATIVE_DIRECTION_IDENTICAL);
+			source, WL_POINTER_AXIS_RELATIVE_DIRECTION_IDENTICAL);
 	}
 }
```

A random idea for another use-case of this function is scrolls emulated by tablet rings, in which case `WL_POINTER_AXIS_SOURCE_WHEEL` is passed.

In addition, I suggest adding `wlr_seat_pointer_notify_frame(seat->seat);` at the end of `cursor_emulate_axis()`. This is currently unnecessary since it's called from pointer motion handlers, but let's follow other `cursor_emulate_*()` functions for portability.

jp7677:
Nice find and that looks indeed better. I've added this as a WIP commit for now (to be squashed later).
@Robert42 does this version still works for you?

Robert42:
>  @Robert42 does this version still works for you?

Yup works perfect.

jp7677:
> > @Robert42 does this version still works for you?
> 
> Yup works perfect.

Nice! I've squashed the last commit into the earlier ones and should be good now.

PR #2684, opened 2025-04-18
labwc-config(5): add mousebind direction example

johanmalm:
Helped-by: @Consolatis
Related-to: #2677

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -760,7 +760,11 @@ extending outward from the snapped edge.
 	Mouse buttons and directions can be combined with modifier-keys
 	(shift (S), super/logo (W), control (C), alt (A), meta (M) and
 	hyper (H)), for example:
-	<mousebind button="A-Right" action="Press">
+
+```
+<mousebind button="A-Right" action="Press">...
+<mousebind direction="W-Up" action="Scroll">...
+```
 
 	Supported mouse *actions* include:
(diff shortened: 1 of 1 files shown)


tokyo4j:
I understand the main focus in this PR is to clarify that `direction` should be used instead of `button` for mousebinds with `Scroll` action, but I feel it's a bit strange to show only such mousebinds as examples.

I suggest showing both `Scroll` and `Drag` (or other) keybinds like:
```xml
<?xml version="1.0"?>
<labwc_config>
<mouse>
  <default />
  <context name="Frame">
    <mousebind direction="W-S-Up" action="Scroll">
      <action name="ResizeRelative" left="10%" right="10%" top="10%" bottom="10%" />
    </mousebind>
      <action name="Move" />
    </mousebind>
  </context>
</mouse>
</labwc_config>
```

johanmalm:
Very good point. Changed as you suggested.

Consolatis:
Suggest also adding a scroll example below `<mousebind button="A-Right" action="Press">` (line 763)

johanmalm:
> Suggest also adding a scroll example below `<mousebind button="A-Right" action="Press">` (line 763)

Like so? Ready for review again.

Consolatis:
> Like so? Ready for review again.

Yep, LGTM

tokyo4j:
Thanks!

PR #2686, opened 2025-04-22
Basic (vertical) titlebar gradient support

jlindgren90:
For after 0.8.4. ~It needs better commit messages and documentation updates still.~

This is enough to get basic vertical titlebar gradients working.
Some examples that are working fairly well (Openbox top, labwc bottom):

Clearlooks (mild SplitVertical gradient):
clearlooks
clearlooks-labwc

Clearlooks-Olive (very subtle simple Vertical gradient):
clearlooks-olive
clearlooks-olive-labwc

Onyx (obvious SplitVertical gradient):
onyx
onyx-labwc

They don't look identical due to various border options that labwc doesn't support.


src/theme.c
```diff
@@ -452,6 +465,56 @@ parse_hexstrs(const char *hexes, float colors[3][4])
...
+			rgba[0] = ((argb >> 16) & 0xFF) / 255.0;
+			rgba[1] = ((argb >> 8) & 0xFF) / 255.0;
+			rgba[2] = (argb & 0xFF) / 255.0;
+			rgba[3] = 1.0;
+		}
+	}
+}
+
+static enum lab_gradient
+parse_gradient(const char *str)
```
johanmalm, line 500: What did you base the parsing methodology on?
1. Openbox source code
2. An interpretation of the openbox wiki documentation: https://openbox.org/help/Themes#Textures
3. Reverse engineering
4. Other

johanmalm (commented):
Nice! I think many people will like this :smile:

If we wanted to avoid that `_GNU_SOURCE` we could do a compat thing like: https://github.com/git/git/blob/master/compat/strcasestr.c

src/theme.c
jlindgren90: I had to look at the Openbox source code to see what it was doing, because the themes shipped with Openbox itself don't follow what one would expect from the wiki. Our `parse_gradient` is not a copy of the Openbox equivalent though (it's much simpler).

johanmalm: :+1: 

Yes, I remember getting confused trying to understand the texture logic before. Think I was looking at the border part that time.

jlindgren90:
> If we wanted to avoid that _GNU_SOURCE we could do a compat thing

Up to you. I think I slightly prefer using the libc version of `strcasestr` since it's available.

tokyo4j:
Do we support vertical gradients or more complex patterns in the future? If so, I think we can redraw the whole titlebar every time a window is resized, rather than having 1-pixel-width buffer and widen it.

My idea is having `scaled_titlebar_buffer` which have following properties:
- Renders the whole titlebar (including corners) based on:
  - Output scale (which this PR doesn't consider IIUC)
  - Window width
  - Corner radius
  - Border width
  - Border color
  - Background color
- If the background color and border color are the same and the corner radius is zero, renders a rectangle with `wlr_scene_rect` for optimization. This can be used by raspberry pi where saving memory & CPU usage may be considered critical.

What do you think of this?

johanmalm:
> > If we wanted to avoid that _GNU_SOURCE we could do a compat thing
> 
> Up to you. I think I slightly prefer using the libc version of `strcasestr` since it's available.

I'm not super precious about it, but prefer to keep GNUisms out where we can avoid it... which I think hitherto we have.
https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#the-use-of-gnu-extensions

The thing that could possibly swing me is the `#define X(Y) ({ Z })` macro :smile: but doesn't seem worth it for `strcasestr()`

jlindgren90:
> I'm not super precious about it, but prefer to keep GNUisms out where we can avoid it...

Technically `strcasestr` is a BSD rather than GNU extension - not sure if that makes a difference to you or not. It's a available on e.g. FreeBSD by default. It just didn't make it into POSIX for whatever reason.

I have no problem with adding the fallback though.

jlindgren90:
> Will we support horizontal gradients or more complex patterns in the future? If so, I think we can redraw the whole titlebar every time a window is resized, rather than having 1-pixel-width buffer and widen it.
> ...
> What do you think of this?

I'm not opposed to any of the ideas you mention, but I'm not volunteering to implement it either :)

Stretching the 1px-wide buffer horizontally seemed like a fair compromise between performance/simplicity and features in my mind, and I think it's still a nice incremental improvement over supporting only solid backgrounds. I don't know anything about what it translates to in terms of HW acceleration on RPi, though.

tokyo4j:
> I'm not opposed to any of the ideas you mention, but I'm not volunteering to implement it either :)
> 
> Stretching the 1px-wide buffer horizontally seemed like a fair compromise between performance/simplicity and features in my mind, and I think it's still a nice incremental improvement over supporting only solid backgrounds. I don't know anything about what it translates to in terms of HW acceleration on RPi, though.

OK. I agree that your approach in this PR is the best for the goal of supporting vertical gradients and I'm not opposed to it either :smile:.

Maybe I can work on the idea I suggested above some months later (aiming for full compatibility with Openbox if possible) and discuss if the performance degradation is acceptable there.

johanmalm:
@tokyo4j I like you proposal and have thought for some time that we should go that way to support non-vertical gradients (it's just not that trivial to implement).

Presumably this PR wouldn't make it harder to go down that route at a later stage?

tokyo4j:
> Presumably this PR wouldn't make it harder to go down that route at a later stage?

Yeah, I think we can merge this PR for now and I can work on my idea later (sorry I forgot to say this in my comment above). I can also help review this PR.

johanmalm:
> > I'm not super precious about it, but prefer to keep GNUisms out where we can avoid it...
> 
> Technically `strcasestr` is a BSD rather than GNU extension - not sure if that makes a difference to you or not. It's a available on e.g. FreeBSD by default. It just didn't make it into POSIX for whatever reason.
> 
> I have no problem with adding the fallback though.

Have discussed on IRC and got majority preference for re-implementing to avoid the GNU extension. I know it's more ethereal than technical - but feels better to pursue the line.

@tokyo4j - do you fancy taking the lead on reviewing this one then, with a view to merge after the release.

jlindgren90:
- Replaced `strcasestr` with `g_ascii_strdown` + `strstr`
- Rebased onto master to fix merge conflicts

johanmalm:
> * Replaced `strcasestr` with `g_ascii_strdown` + `strstr`

:+1:  Good choice

include/common/scaled-font-buffer.h
```diff
@@ -17,7 +17,7 @@ struct scaled_font_buffer {
 	char *text;
 	int max_width;
 	float color[4];
-	float bg_color[4];
+	struct _cairo_pattern *bg_pattern;
```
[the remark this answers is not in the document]
jlindgren90: Sure, I can do that, it just needs pulling in `cairo.h`.

jlindgren90:
- Replaced `struct _cairo_pattern` with `cairo_pattern_t`
- Rebased

jlindgren90:
Rebased to fix merge conflicts

jlindgren90:
> So for now, can we disable subpixel antialiasing for texts on gradient background

It's not worth it to me to merge partial gradient support right now, if it means losing subpixel text rendering. I would rather wait until it can be implemented in a way that both will work.

So, maybe we should just leave this PR as draft for now, and eventually cancel it once #2695 is ready?

tokyo4j:
> It's not worth it to me to merge partial gradient support right now, if it means losing subpixel text rendering. I would rather wait until it can be implemented in a way that both will work.

Is it such a big drawback to lose subpixel text rendering, only on gradient background? Of course we should do it when we can, but I think it's still appealing to many users that they can use vertical gradient titlebar in the expense of losing subpixel antialiasing. I don't want you and users to wait for my PR to be finalized for months, as it requires more work like optimization for solid color background and the rethink of keybind toggle indicator which is currently drawn as the top window border.

In the screenshot below, the titlebar placed above is rendered with gray antialiasing (note: in this PR, we will be able to keep subpixel antialiasing for solid background color), and the one placed below is rendered with subpixel antialiasing. I will not notice its difference until someone tells me.

20250526_13h25m34s_grim

CC @johanmalm @Consolatis @ahesford

johanmalm:
> I am ok with disabling subpixel rendering only on gradient backgrounds.

Same. The images indeed look very similar.

It would be good to get consensus with @jlindgren90 on this though. It is not an area I know much about, but I think they do :smile: based on various loosely related commits in both wlroots and labwc.

An important question for me here is whether this PR moves us in the right direction - or if it makes it harder to later get to the right space (sorry not looked at code in detail - but can do if you want me to).

jlindgren90:
> We should vertically slide the pattern drawn on the corner buffers, otherwise they don't align with the pattern on the titlebar rectangle:
> 
> 20250525_23h32m49s_grim

@tokyo4j can you share the theme/config you used for this screenshot? I don't see the issue here, but I use a thin border and square corners.

tokyo4j:
> > An important question for me here is whether this PR moves us in the right direction - or if it makes it harder to later get to the right space (sorry not looked at code in detail - but can do if you want me to).
> 
> If the "right direction" is to eventually render the title text into the same buffer as the background, as discussed in [#2695 (comment)](https://github.com/labwc/labwc/discussions/2695#discussioncomment-12973599), then most of the code touched by this PR is going to get reworked anyway. I don't personally anticipate this PR as making the future work either easier or harder.
> 
> If the future gradient work is indeed months away, then I think this vertical-pattern approach still has some value for the short term, but does not add much value in the long term (except for the config parsing bits).

I'm afraid this PR makes the future work harder if we change the APIs of font buffers like as it is in this PR, as it complicates the APIs and it may give readers a wrong impression that we can extend this approach (draw gradients as the background of font buffers and vertically center the text) into other components like menu items, which we can't.

But without the API changes in font buffers, I'd say this PR doesn't make the future work easier or harder.

The reason why I want to make this PR less complicated by giving up subpixel antialiasing and push this PR forward is that there are still some uncertainty in my approach I suggested in #2695 and I'm not 100% sure I can finalize my approach in a short term. I'm currently facing two problems with it:
- (1) How should we re-implement the keybind toggle indicator which is currently implemented by changing the top border color?
- (2) Can we not introduce performance degradation I showed in #2695 for the default configuration (with rounded corners)? There is a consensus in #2695 that we can ignore the performance degradation in usual desktop uses, but it would be nicer if we can apply some optimization for the default configuration so that we don't need to redraw the entire titlebar every time the window is resized.

I guess I can open a PR with my approach in a short term if we resolve these problems by placing the keybind toggle indicator on top of the titlebar and by making (2) out-of-scope, but my taste is pushing forward this PR for now and having more time for addressing the problems above.

> can you share the theme/config you used for this screenshot? I don't see the issue here, but I use a thin border and square corners.

Sorry, I forgot to say that the pattern on the corner buffers are shifted __by border width__. Here's my configuration:

```
*.title.bg: Raised Gradient vertical
*.title.bg.color: #ff0000
*.title.bg.color.splitTo: #ff0000
*.title.bg.colorTo: #00ff00
*.title.bg.colorTo.splitTo: #0000ff
border.width: 6
```

jlindgren90:
- Rebased
- Split up commits further
- Added `zfree_pattern()` to graphics-helpers.h
- Added separate `scaled_font_buffer_set_fixed_height/bg_pattern()`
- Restored original `scaled_font_buffer_update()` parameters
- Fixed issue with pattern offset in corner buffers

tokyo4j:
OK, I still prefer dropping subpixel antialiasing for gradients in favor of simplicity, but I can live with your approach as this PR no longer touches all the users of `scaled_font_buffers` like OSD.

I pushed some changes to `scaled_font_buffers` interfaces to clearly split the code paths between normal font buffers and title font buffers.

I'll do another review of the entire PR later. Please un-draft this PR if it's ready.

jlindgren90:
Okay, I would normally prefer that you ask first before pushing changes, but that's fine. The `create_for_titlebar` idea does simplify the API, I like that. I'll take a closer look later on today.

tokyo4j:
> I would normally prefer that you ask first before pushing changes, but that's fine.

Sorry, next time I'll push to my own branch and show you diff.

> I did not incorporate all the changes to scaled_font_buffer since some of them seemed a bit arbitrary. For example, I prefer having the separate tmp_bg_pattern pointer to hold the temporarily-allocated solid color pattern, and kept the fixed_height vs computed_height variable names for clarity.

That's fine, but I think we can rename `tmp_bg_pattern` to e.g. `solid_bg_pattern` for clarity.

jlindgren90:
Rebased to fix merge conflicts.

If someone has time and would like to help out with the documentation updates (new theme options), I'd welcome the help. Otherwise, I'll get to it eventually.

johanmalm:
> If someone has time and would like to help out with the documentation updates (new theme options), I'd welcome the help. Otherwise, I'll get to it eventually.

I have typed up documentation.

<details>

```
diff --git a/docs/labwc-theme.5.scd b/docs/labwc-theme.5.scd
index 5b267f35..57b480a0 100644
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -41,6 +41,8 @@ labwc-config(5).
 	- #rrggbb (hexadecimal RGB values)
 	- #rrggbb aaa (same but with decimal alpha value percentage)
 	- #rrggbbaa (same but with inline alpha value in hex encoding)
+	- #rgb (hexadecimal RGB values where #fff is shorthand for #f0f0f0)
+	  (since 0.9.0)
 
 	Note: the #rrggbb aaa notation is deprecated starting from
 	labwc 0.7.2 and may be removed in future releases.
@@ -49,6 +51,52 @@ labwc-config(5).
 	Justification determines the horizontal alignment of text.
 	Valid options are Left, Center and Right.
 
+*textures*
+	Textures determine the visual look of an element. They are specified
+	through a text string with a number of fields. Capitalization is not
+	significant. The format is as follows:
+
+```
+(solid | gradient gradient-type)
+```
+
+	... where *|* stands for "or".
+
+	- *solid* means that the background of the texture is filled with a
+	  single color. The texture must be accompanied by a single color field.
+	  For example:
+
+```
+window.*.title.bg: Solid Flat
+window.*.title.bg.color: #f8f5f2
+```
+
+	- *gradient* means that the background of the texture is filled with a
+	  gradient from one color to another. When gradient is specified, it
+	  must be followed by the gradient's type. It must also be accompanied
+	  by the fields *color* and *colorTo*. Valid gradient types are:
+	  (1) *Vertical* where the gradient goes from the top edge to the bottom
+	  (2) *SplitVertical* where the gradient is split in the middle and goes
+	  out toward the top and bottom edges. SplitVertical gradients have two
+	  extra, optional color fields: *color.splitTo* and *colorTo.splitTo*.
+	  These colors are the light colors used on the far top and bottom of
+	  the SplitVertical gradient. When these are omitted, the default values
+	  for these are *color x 5/4* and *colorTo x 17/16*. Examples:
+
+```
+# Vertical gradient
+window.*.title.bg: Gradient Vertical
+window.*.title.bg.color: #658fb5
+window.*.title.bg.colorTo: #4d6982
+
+# SplitVertical gradient
+window.*.title.bg: Gradient SplitVertical
+window.*.title.bg.color: #658fb5
+window.*.title.bg.color.splitTo: #7595b9
+window.*.title.bg.colorTo: #4d6982
+window.*.title.bg.colorTo.splitTo: #557485
+```
+
 # THEME ELEMENTS
 
 *border.width*
@@ -75,12 +123,22 @@ labwc-config(5).
 	value as set for window.active.border.color to disable the status
 	indicator. Default is #ff0000.
 
+*window.active.title.bg*
+	Texture for the focused window's titlebar. See texture section above.
+	Default is *Solid*
+
 *window.active.title.bg.color*
-	Background color for the focused window's titlebar.
+	Background color for the focused window's titlebar. See texture section
+	for additional color fields associated with gradients.
 	Default is #e1dedb.
 
+*window.inactive.title.bg*
+	Texture for non-focused windows' titlebars. See texture section above.
+	Default is *Solid*
+
 *window.inactive.title.bg.color*
-	Background color for non-focused windows' titlebars.
+	Background color for non-focused windows' titlebars. See texture section
+	for additional color fields associated with gradients.
 	Default is #f6f5f4.
 
 *window.active.label.text.color*
diff --git a/docs/themerc b/docs/themerc
index 7ac293f7..045c1021 100644
--- a/docs/themerc
+++ b/docs/themerc
@@ -25,6 +25,14 @@ window.active.indicator.toggled-keybind.color: #ff0000
 # window titlebar background
 window.active.title.bg.color: #e1dedb
 window.inactive.title.bg.color: #f6f5f4
+window.*.title.bg: Solid
+
+# Vertical and SplitVertical gradients are supported for window.*.title.bg
+# with the following additional color fields. See labwc-theme(5) for examples.
+#
+#     window.*.title.bg.colorTo:
+#     window.*.title.bg.color.splitTo:
+#     window.*.title.bg.colorTo.splitTo:
 
 # window titlebar text
 window.active.label.text.color: #000000
```

</details>

If of satisfactory standard, I can push directly to this PR if you want - or to `johanmalm` and you can pull from there.

jlindgren90:
@johanmalm thanks. I like how you organized it. Direct push to the PR is fine.

edit: committed your diff from above (no changes) with you as author

johanmalm:
Thanks.

jlindgren90:
I think this is ready for review now. Thanks @tokyo4j and @johanmalm for all the feedback and help so far!

johanmalm:
LGTM

> I'll do another review of the entire PR later. Please un-draft this PR if it's ready.

I feel @tokyo4j ought to have a chance to veto though.

docs/labwc-theme.5.scd
```diff
@@ -49,6 +51,52 @@ labwc-config(5).
 	Justification determines the horizontal alignment of text.
 	Valid options are Left, Center and Right.
 
+*textures*
+	Textures determine the visual look of an element. They are specified
+	through a text string with a number of fields. Capitalization is not
+	significant. The format is as follows:
+
+```
+(solid | gradient gradient-type)
```
tokyo4j: nit: it may be a bit confusing that actual strings (`solid` and `gradient`) and a type (`gradient-type`) are placed just like the same. Suggest something like this:
```
(Solid | Gradient [gradient-type])
```

Or
```
(Solid | Gradient (Vertical | SplitVertical))
```

I also used PascalCase for consistency with the examples below.

docs/labwc-theme.5.scd
```diff
@@ -49,6 +51,52 @@ labwc-config(5).
...
+```
+
+	... where *|* stands for "or".
+
+	- *solid* means that the background of the texture is filled with a
+	  single color. The texture must be accompanied by a single color field.
+	  For example:
+
+```
+window.*.title.bg: Solid Flat
```
tokyo4j: I think `Flat` should be removed as it's not described anywhere.

tokyo4j:
LGTM other than minor suggestions in the docs. Thanks for your effort and cooperation!

docs/labwc-theme.5.scd
johanmalm: It's heavily inspired by https://openbox.org/help/Themes#Textures

`[ ]` indicates optional, so don't agree with using that for the gradient-type.

There are many other gradients which we may consider using in future, so I think a simple `type` identifier is simpler when describing the syntax.

How about this: `(Solid | Gradient <type>)`

docs/labwc-theme.5.scd
johanmalm: Agree. Good point.

@jlindgren90 Are you happy to lead with the fixups so we don't tread on each others commit-toes?

jlindgren90: Yep, I pushed a fixup commit just now. I'll squash it before merging.

johanmalm: LGTM. @jlindgren90 do you want to go for the final squash + merge

PR #2688, opened 2025-04-22
view: handle initially minimized views (simple approach)

jlindgren90:
Taking the simplistic approach of waiting until the view is mapped to minimize it. A more optimal but invasive approach could be to modify the `view->impl->map` functions to directly support map-while-minimized, but this would require some careful disentangling of the concepts of "minimized" and "unmapped", which overlap a bit in labwc.

Fixes: #2627

@tokyo4j if you could test that this fixes the issue you saw, I'd appreciate it.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2436,7 +2436,18 @@ static void
 handle_map(struct wl_listener *listener, void *data)
 {
 	struct view *view = wl_container_of(listener, view, mappable.map);
-	view->impl->map(view);
+	if (view->minimized) {
+		/*
+		 * The view->impl functions do not directly support
+		 * mapping a view while minimized. Instead, mark it as
+		 * not minimized, map it, and then minimize it again.
+		 */
+		view->minimized = false;
+		view->impl->map(view);
+		view_minimize(view, true);
+	} else {
+		view->impl->map(view);
+	}
 }
 
 static void


jlindgren90:
Also see https://github.com/labwc/labwc/issues/2627#issuecomment-2810626634 for a test-case using Qt.

jlindgren90:
Eventually I think we need a way to track "mapped but minimized" state.

For a quick fix, I suppose `(view->mapped || (view->minimized && view->been_mapped))` could address the particular concern.

jlindgren90:
I did some more mulling this over and came to the conclusion that the only really necessary change is to `handle_map()`. I couldn't find any case where the other changes actually matter.

> However, I'm a bit afraid anything wrong happens if view_is_focusable() is called when a window is minimized before mapping:

I'm curious if you can actually make this happen. I am not sure that labwc sees the `request_minimize` event until the window is already in the process of being mapped. The whole process is a bit difficult to follow on the wlroots side.

I did test the case of a X11 client-requested unmap (`hide()` in Qt terms) and in that case `view->surface` gets set to NULL, so `view_is_focusable()` correctly returns false.

Anyway, I'm marking this PR as "ready" again with just the `handle_map()` change. That seems to be enough to fix #2627 on its own.

tokyo4j:
> I'm curious if you can actually make this happen. I am not sure that labwc sees the request_minimize event until the window is already in the process of being mapped. The whole process is a bit difficult to follow on the wlroots side.

I've just realized my comment saying "this PR breaks this assumption" was completely wrong, since it's already possible that `view->minimized` is set before mapping a surface. I was simply confused when I said that.

And as for your comment, I _guess_ `view_is_focusable()` being called for minimized-but-unmapped surfaces is possible in a very limited situation. But as I said above, this PR doesn't change assumptions about `view->mapped` or `view->minimized` and I think we can address it later when it becomes a problem.

LGTM. I really like how your updated implementation is so minimal and not touching the complicated parts of map handlers.

@johanmalm @Consolatis @ahesford Any thoughts on whether to merge this PR before next version bump? I personally think we can merge this PR now, but merging after the version bump will be fine too, as the VSCode issue #2627 is confirmed to be a application bug by its developers (https://github.com/microsoft/vscode/issues/242726#issuecomment-2783708602) and it's already fixed.

johanmalm:
Thanks a lot for the work on this - both of you.

Let's go for it. Doesn't seem right to hold back on a bug fix of this nature with, as I see it, low risk of unintended side-effects.

PR #2689, opened 2025-04-22
osd: suppress errors with null window switcher fields

tokyo4j:
Intended to be merged before the next release.

Fixes a regression in #2620 that emits errors like:

```
[../labwc/src/common/scaled-font-buffer.c:26] font_buffer_create() failed
```

...when `osd_field_get_content()` doesn't set non-null text.

Can be tested with the following configuration (`output` shows nothing when only one output is connected):
```xml
  <windowSwitcher>
    <fields>
      <field width="100%" content="output" />
    </fields>
  </windowSwitcher>
```


Consolatis:
Thanks :)

PR #2692, opened 2025-04-23
Add support for all Pango font weight options

spl237:
The current labwc code recognises all three valid Pango font slant options (normal, italic and oblique), but only two of the twelve valid Pango font weight options (normal and bold). I assume this is because these were the only font weight options supported by openbox.

This trivial patch adds support for the remaining ten Pango font weight options so that the title bar font can be specified to use any valid weight.


Consolatis:
Thanks, from my point of view this could be merged before the release as I don't expect any side effects from the changes.
With the second commit squashed into the first this LGTM.

spl237:
Note that this code could be simplified substantially by making the enum for font_weight a direct copy of the PangoWeight enum, with the options in the same order, but I have retained the existing order with NORMAL as zero.

src/common/font.c
```diff
@@ -23,8 +23,44 @@ font_to_pango_desc(struct font *font)
...
+			break;
+		case FONT_WEIGHT_HEAVY:
+			pango_font_description_set_weight(desc, PANGO_WEIGHT_HEAVY);
+			break;
+		case FONT_WEIGHT_ULTRAHEAVY:
+			pango_font_description_set_weight(desc, PANGO_WEIGHT_ULTRAHEAVY);
+			break;
+		default:
+			pango_font_description_set_weight(desc, PANGO_WEIGHT_NORMAL);
+			break;
```
Consolatis, line 63: One small point for improvement here: you could omit the default cause and instead add a `case FONT_WEIGHT_NORMAL` here. That would ensure that the compiler complains if the enum and this switch ever gets out of sync.

ahesford, line 63: For consistency with other switch blocks, please put the opening brace on the same line as the `switch` statement and remove one level of indentation inside the block, aligning the `case` delcarations with the top-level `switch`.

spl237: Fine  - will do. I didn't know if not having a default case was a code style no-no for labwc, but if not, then your suggestion is cleaner.

spl237: Fine - will do.

Consolatis: In general for cases like enums having a default case might cover up a bug so I consider it good practice to not have one in those cases. I prefer a compile error over something working (although incorrectly) by accident.

ahesford:
I wouldn't copy the pango enum, but instead just use it directly. Wrapping it in our own namespace doesn't add a lot and, as noted, requires we maintain a giant switch to map them.

spl237:
In that case, for consistency, it would probably make sense to also use the PangoStyle enum instead of font_slant, as the two situations are analogous.

Consolatis:
> I wouldn't copy the pango enum, but instead just use it directly. Wrapping it in our own namespace doesn't add a lot and, as noted, requires we maintain a giant switch to map them.

I don't really have a preference regarding that. As we currently use our own enum I feel like we should keep doing so until after the release though and can then go ahead and change both to the pango enums outside of the cooldown.

ahesford:
Taking this as-is and refactoring after the release is OK with me.

Consolatis:
Thanks :)

spl237:
You're very welcome!

Once the release is out of the way, I'm very happy to make the changes to use the mango types. Will submit an appropriate PR then.

PR #2699, opened 2025-04-28
Nag

johanmalm:
--EDIT-- For after 0.8.4

For anyone wanting to test this (and assuming you're not installing it system wide), you need to create a symlink from build/clients/labnag to a location in `$PATH`.

The exit code relates to the position of the button as demonstrated in this example: ~~`labnag -m 'Foo' -Z 1 : -Z 2 : -Z 3 : -s 0`~~ (--EDITED-- See https://github.com/labwc/labwc/pull/2699#issuecomment-2993658830 for up-to-date example)

This PR does not deal with displaying config file error messages. Suggest we explore that in a separate PR - possibly by exploring something like `xmlCtxtSetErrorHandler()` or similar.

TODO:

- ~~In `action_prompt_create()` we ought to consider adding a config option
      for the command.~~ (--EDIT-- Do this at some other stage)
- [x] --EDIT-- Change examples to use `<prompt>` in `menu.xml`
- [x] --EDIT-- Update `labwc-config(5)`

### DIFF
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -1,0 +1,1652 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Based on https://github.com/swaywm/sway/tree/master/swaynag
+ *
+ * Copyright (C) 2016-2017 Drew DeVault
+ * Copyright (C) 2025 Johan Malm
+ */
+#define _POSIX_C_SOURCE 200809L
+#include <assert.h>
+#include <cairo.h>
+#include <ctype.h>
+#include <getopt.h>
+#include <glib.h>
+#include <pango/pangocairo.h>
+#include <poll.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <strings.h>
+#include <sys/signalfd.h>
+#include <sys/stat.h>
+#include <sys/timerfd.h>
+#include <sys/wait.h>
+#include <time.h>
+#include <unistd.h>
+#include <wayland-client.h>
+#include <wayland-cursor.h>
+#include <wlr/util/log.h>
+#include "pool-buffer.h"
+#include "cursor-shape-v1-client-protocol.h"
+#include "wlr-layer-shell-unstable-v1-client-protocol.h"
+
+#define LABNAG_MAX_HEIGHT 500
+#define LAB_EXIT_FAILURE 255
+#define LAB_EXIT_SUCCESS 0
+
+struct conf {
+	PangoFontDescription *font_description;
+	char *output;
+	uint32_t anchors;
+	int32_t layer; /* enum zwlr_layer_shell_v1_layer or -1 if unset */
+
+	/* Colors */
+	uint32_t button_text;
+	uint32_t button_background;
+	uint32_t details_background;
+	uint32_t background;
+	uint32_t text;
+	uint32_t border;
+	uint32_t border_bottom;
+
+	/* Sizing */
+	ssize_t bar_border_thickness;
+	ssize_t message_padding;
+	ssize_t details_border_thickness;
+	ssize_t button_border_thickness;
+	ssize_t button_gap;
+	ssize_t button_gap_close;
+	ssize_t button_margin_right;
+	ssize_t button_padding;
+};
+
+struct nag;
+
+struct pointer {
+	struct wl_pointer *pointer;
+	uint32_t serial;
+	struct wl_cursor_theme *cursor_theme;
+	struct wl_cursor_image *cursor_image;
+	struct wl_surface *cursor_surface;
+	int x;
+	int y;
+};
+
+struct seat {
+	struct wl_seat *wl_seat;
+	uint32_t wl_name;
+	struct nag *nag;
+	struct pointer pointer;
+	struct wl_list link; /* nag.seats */
+};
+
+struct output {
+	char *name;
+	struct wl_output *wl_output;
+	uint32_t wl_name;
+	uint32_t scale;
+	struct nag *nag;
+	struct wl_list link; /* nag.outputs */
+};
+
+struct button {
+	char *text;
+	char *action;
+	int x;
+	int y;
+	int width;
+	int height;
+	bool expand;
+	bool dismiss;
+	struct wl_list link;
+};
+
+enum {
+	FD_WAYLAND,
+	FD_TIMER,
+	FD_SIGNAL,
+
+	NR_FDS,
+};
+
+struct nag {
+	bool run_display;
+
+	struct wl_display *display;
+	struct wl_compositor *compositor;
+	struct wl_seat *seat;
+	struct wl_shm *shm;
+	struct wl_list outputs;
+	struct wl_list seats;
+	struct output *output;
+	struct zwlr_layer_shell_v1 *layer_shell;
+	struct zwlr_layer_surface_v1 *layer_surface;
+	struct wp_cursor_shape_manager_v1 *cursor_shape_manager;
+	struct wl_surface *surface;
+
+	uint32_t width;
+	uint32_t height;
+	int32_t scale;
+	struct pool_buffer buffers[2];
+	struct pool_buffer *current_buffer;
+
+	struct conf *conf;
+	char *message;
+	struct wl_list buttons;
+	struct pollfd pollfds[NR_FDS];
+
+	struct {
+		bool visible;
+		char *message;
+		char *details_text;
+		int close_timeout;
+		bool use_exclusive_zone;
+
+		int x;
+		int y;
+		int width;
+		int height;
+
+		int offset;
+		int visible_lines;
+		int total_lines;
+		struct button *button_details;
+		struct button button_up;
+		struct button button_down;
+	} details;
+};
+
+static int exit_status = LAB_EXIT_FAILURE;
+
+static void close_pollfd(struct pollfd *pollfd);
+
+static PangoLayout *
+get_pango_layout(cairo_t *cairo, const PangoFontDescription *desc,
+		const char *text, double scale, bool markup)
+{
+	PangoLayout *layout = pango_cairo_create_layout(cairo);
+	pango_context_set_round_glyph_positions(pango_layout_get_context(layout), false);
+
+	PangoAttrList *attrs;
+	if (markup) {
+		char *buf;
+		GError *error = NULL;
+		if (pango_parse_markup(text, -1, 0, &attrs, &buf, NULL, &error)) {
+			pango_layout_set_text(layout, buf, -1);
+			free(buf);
+		} else {
+			wlr_log(WLR_ERROR, "pango_parse_markup '%s' -> error %s",
+				text, error->message);
+			g_error_free(error);
+			markup = false; /* fallback to plain text */
+		}
+	}
+	if (!markup) {
+		attrs = pango_attr_list_new();
+		pango_layout_set_text(layout, text, -1);
+	}
+
+	pango_attr_list_insert(attrs, pango_attr_scale_new(scale));
+	pango_layout_set_font_description(layout, desc);
+	pango_layout_set_single_paragraph_mode(layout, 1);
+	pango_layout_set_attributes(layout, attrs);
+	pango_attr_list_unref(attrs);
+	return layout;
+}
+
+static void
+get_text_size(cairo_t *cairo, const PangoFontDescription *desc, int *width, int *height,
+		int *baseline, double scale, bool markup, const char *fmt, ...)
+{
+	va_list args;
+	va_start(args, fmt);
+	gchar *buf = g_strdup_vprintf(fmt, args);
+	va_end(args);
+	if (!buf) {
+		return;
+	}
+
+	PangoLayout *layout = get_pango_layout(cairo, desc, buf, scale, markup);
+	pango_cairo_update_layout(cairo, layout);
+	pango_layout_get_pixel_size(layout, width, height);
+	if (baseline) {
+		*baseline = pango_layout_get_baseline(layout) / PANGO_SCALE;
+	}
+	g_object_unref(layout);
+
+	g_free(buf);
+}
+
+static void
+render_text(cairo_t *cairo, const PangoFontDescription *desc, double scale,
+		bool markup, const char *fmt, ...)
+{
+	va_list args;
+	va_start(args, fmt);
+	gchar *buf = g_strdup_vprintf(fmt, args);
+	va_end(args);
+	if (!buf) {
+		return;
+	}
+
+	PangoLayout *layout = get_pango_layout(cairo, desc, buf, scale, markup);
+	cairo_font_options_t *fo = cairo_font_options_create();
+	cairo_get_font_options(cairo, fo);
+	pango_cairo_context_set_font_options(pango_layout_get_context(layout), fo);
+	cairo_font_options_destroy(fo);
+	pango_cairo_update_layout(cairo, layout);
+	pango_cairo_show_layout(cairo, layout);
+	g_object_unref(layout);
+
+	g_free(buf);
+}
+
+static void
+cairo_set_source_u32(cairo_t *cairo, uint32_t color)
+{
+	cairo_set_source_rgba(cairo,
+			(color >> (3*8) & 0xFF) / 255.0,
+			(color >> (2*8) & 0xFF) / 255.0,
+			(color >> (1*8) & 0xFF) / 255.0,
+			(color >> (0*8) & 0xFF) / 255.0);
+}
+
+static uint32_t
+render_message(cairo_t *cairo, struct nag *nag)
+{
+	int text_width, text_height;
+	get_text_size(cairo, nag->conf->font_description, &text_width,
+		&text_height, NULL, 1, true, "%s", nag->message);
+
+	int padding = nag->conf->message_padding;
+
+	uint32_t ideal_height = text_height + padding * 2;
+	uint32_t ideal_surface_height = ideal_height;
+	if (nag->height < ideal_surface_height) {
+		return ideal_surface_height;
+	}
+
+	cairo_set_source_u32(cairo, nag->conf->text);
+	cairo_move_to(cairo, padding, (int)(ideal_height - text_height) / 2);
+	render_text(cairo, nag->conf->font_description, 1, false,
+			"%s", nag->message);
+
+	return ideal_surface_height;
+}
+
+static void
+render_details_scroll_button(cairo_t *cairo, struct nag *nag,
+		struct button *button)
+{
+	int text_width, text_height;
+	get_text_size(cairo, nag->conf->font_description, &text_width,
+		&text_height, NULL, 1, true, "%s", button->text);
+
+	int border = nag->conf->button_border_thickness;
+	int padding = nag->conf->button_padding;
+
+	cairo_set_source_u32(cairo, nag->conf->details_background);
+	cairo_rectangle(cairo, button->x, button->y,
+			button->width, button->height);
+	cairo_fill(cairo);
+
+	cairo_set_source_u32(cairo, nag->conf->button_background);
+	cairo_rectangle(cairo, button->x + border, button->y + border,
+			button->width - (border * 2),
+			button->height - (border * 2));
+	cairo_fill(cairo);
+
+	cairo_set_source_u32(cairo, nag->conf->button_text);
+	cairo_move_to(cairo, button->x + border + padding,
+			button->y + border + (button->height - text_height) / 2);
+	render_text(cairo, nag->conf->font_description, 1, true,
+			"%s", button->text);
+}
+
+static int
+get_detailed_scroll_button_width(cairo_t *cairo, struct nag *nag)
+{
+	int up_width, down_width, temp_height;
+	get_text_size(cairo, nag->conf->font_description, &up_width, &temp_height,
+		NULL, 1, true, "%s", nag->details.button_up.text);
+	get_text_size(cairo, nag->conf->font_description, &down_width, &temp_height,
+		NULL, 1, true, "%s", nag->details.button_down.text);
+
+	int text_width =  up_width > down_width ? up_width : down_width;
+	int border = nag->conf->button_border_thickness;
+	int padding = nag->conf->button_padding;
+
+	return text_width + border * 2 + padding * 2;
+}
+
+static uint32_t
+render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
+{
+	uint32_t width = nag->width;
+
+	int border = nag->conf->details_border_thickness;
+	int padding = nag->conf->message_padding;
+	int decor = padding + border;
+
+	nag->details.x = decor;
+	nag->details.y = y + decor;
+	nag->details.width = width - decor * 2;
+
+	PangoLayout *layout = get_pango_layout(cairo, nag->conf->font_description,
+			nag->details.message, 1, false);
+	pango_layout_set_width(layout,
+			(nag->details.width - padding * 2) * PANGO_SCALE);
+	pango_layout_set_wrap(layout, PANGO_WRAP_WORD_CHAR);
+	pango_layout_set_single_paragraph_mode(layout, false);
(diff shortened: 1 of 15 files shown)


johanmalm:
Example usage:

```xml
<keybind key="W-q">
  <action name="If">
    <prompt message="Quit?"/>
    <then><action name="Exit"/></then>
  </action>
</keybind>
```

--edited-- to align with https://github.com/labwc/labwc/pull/2699#issuecomment-2993658830

tokyo4j:
Do we really need to support both `<prompt text="Quit?" option1="Yes" option2="No"/>` and `<prompt>Quit?</prompt>` syntaxes? I suggest supporting only the latter syntax and make the prompt command configurable like this:
```xml
<core>
   <promptCommand>labnag - %s -Z "No" : -s "Yes"</promptCommand>
</core>
```

Regarding @Consolatis's comment:
```c
		/*
		 * TODO: figure out how to have more than 2 choices,
		 *       maybe by using branch names like cond_0, cond_1
		 *       and so on and falling back to 'then' and 'else'
		 *       if not found?
		 */
```

I think we can set this out-of-scope. For those who want to do complex things like having multiple choices, they can write their own script and run `wtype` to trigger actions.

Also, I wonder if we should support the following options in `labnag`:
- `-B, --button-no-terminal`
- `-Z, --button-dismiss-no-terminal`
- `-s, --dismiss-button`

https://github.com/i3/i3/issues/2199 is the background for the first 2 options. In short, `i3-nagbar` spawns commands in terminal emulator (set with `$TERMINAL`) with `--button`/`--button-dismiss` options for historical reasons, and `*-no-terminal` options are added to allow launching commands without terminal emulator installed. And then swaybar supported them for compatibility.

Can we just not run commands in terminal emulator and drop those 2 options? IMO, we don't need to focus on compatibility with `swaybar`/`i3-nagbar` and we should prioritize simplicity to make options easier for starters.

Also for simplicity, I suggest dropping the last option. I was confused to find there are both `--dismiss-button` and `--button-dismiss`.

Consolatis:
> I think we can set this out-of-scope. For those who want to do complex things like having multiple choices, they can write their own script and run wtype to trigger actions.

Agreed.

Another issue with the prompt implementation I just noticed is that if a prompt is spawned and then a `Reconfigure` occurs we will execute free'd actions. So this need some reconfigure trigger to simply destroy all prompts, a button click on the GUI prompt would then not do anything but I think that's fine when somebody does a reconfigure in between spawning the prompt and clicking a button.

johanmalm:
Thanks for looking at this.

> Do we really need to support both `<prompt text="Quit?" option1="Yes" option2="No"/>` and `<prompt>Quit?</prompt>` syntaxes? I suggest supporting only the latter syntax

Agree with removing the options attributes.

Might it be worth using `<prompt text="foo"/>` in case we want to extend in future. If we use `<prompt>foo</prompt>` we cannot add attributes or child elements in future without breaking our 'clean' syntax.

> ...and make the prompt command configurable like this:
> 
> ```
> <core>
>    <promptCommand>labnag - %s -Z "No" : -s "Yes"</promptCommand>
> </core>
> ```

I like this in principle. Assume you mean `-m` instead of the lonely `-`?

By the way, have you explore the stdin functionality for future config-error-logging. Try `echo "foo\nbar" | labnag -l -m baz`


> Also, I wonder if we should support the following options in `labnag`:
>     * `-B, --button-no-terminal`
>     * `-Z, --button-dismiss-no-terminal`

Do you mean that we should support `Z` and `B` but remove support for `z` and `b`? I'm thinking it's better to just pass things to `execlp("sh", "sh", "-c", button->action, NULL);` rather than invoking terminals.

>     * `-s, --dismiss-button`

Do you mean to remove support for `-s` and simply use `-z foo :` instead?

> IMO, we don't need to focus on compatibility with `swaybar`/`i3-nagbar` and we should prioritize simplicity to make options easier for starters.

Agree.

Consolatis:
> Do we really need to support both `<prompt text="Quit?" option1="Yes" option2="No"/>` and `<prompt>Quit?</prompt>` syntaxes? I suggest supporting only the latter syntax and make the prompt command configurable like this:
> 
> ```
> <core>
>    <promptCommand>labnag - %s -Z "No" : -s "Yes"</promptCommand>
> </core>
> ```

One thing that doesn't work with this approach is having multiple actions with different option values.

tokyo4j:
> Might it be worth using <prompt text="foo"/> in case we want to extend in future. If we use <prompt>foo</prompt> we cannot add attributes or child elements in future without breaking our 'clean' syntax.

Agreed. If we decide to put `<prompt>` in `<action name="If">`, it means we already deviate from Openbox and it sounds sensible to use `<prompt text="foo" />` syntax.

> By the way, have you explore the stdin functionality for future config-error-logging. Try echo "foo\nbar" | labnag -l -m baz

I confirmed it worked, but I wonder why we need it. Can't we just pass the error messages via command line? Or are you planning to launch `labnag` as a daemon and let it show error messages by receiving messages from labwc via pipe?

> Assume you mean -m instead of the lonely -?

Yes.

> Do you mean that we should support Z and B but remove support for z and b? I'm thinking it's better to just pass things to execlp("sh", "sh", "-c", button->action, NULL); rather than invoking terminals.

Yes, I meant that we can support only `--button`/`--button-dismiss` and make them run a command without terminal emulator, just like what `--button-no-terminal`/`--button-dismiss-no-terminal` are currently doing.

> Do you mean to remove support for -s and simply use -z foo : instead?

Yes.

> One thing that doesn't work with this approach is having multiple actions with different option values.

Yeah, it's a trade-off between flexibility and simplicity. I guess you mean something like having `Yes`/`No` options for some actions but also having `Exit`/`Don't Exit` options for `<action name="Exit">`. But I think we can prioritize simplicity here since `Yes`/`No` options are sufficient for most cases.

johanmalm:
I have done the following:

- Rebased
- As described in the 'Add labnag' commit message at v2:
    - Removed option -s and the default "X" button
    - Removed options -b and -z because there is no requirement to run in a terminal.
    - Removed *-no-terminal from options --button and --button-dismiss because commands are now always run directly without a terminal.
- Fixed `-fanalyzer` and `-Wshadow` warnings (see full commit history at https://github.com/johanmalm/nag)

So, example usage now looks like this:

```
<keybind key="W-q">
  <action name="If">
    <prompt message="Quit?"/>
    <then>
      <action name="Exit"/>
    </then>
  </action>
</keybind>
```


Exit codes are demonstrated by this example `labnag -m EXITCODES -Z 2 -Z 1 -Z 0` - where the 0 button is further to the right (--edited after v3)

Also, have added the swaywm commit message that swaynag was copied at so that we can gitk that directory and chase any improvements.

I have not done anything with the suggestion below, but I propose that we go without it initially and then progressively improve.

```
<core>
   <promptCommand>labnag - %s -Z "No" : -s "Yes"</promptCommand>
</core>
```

johanmalm:
v3:
- Added documentation
- Squashed a few commits and tidied up commit messages

This is now ready for review. I'll leave PR at draft until after `0.9.0`

clients/labnag.c
```diff
@@ -0,0 +1,1643 @@
...
+			wl_display_cancel_read(nag->display);
+		}
+		if (timer_fd >= 0 && fds[1].revents & POLLIN) {
+			close(timer_fd);
+			timer_fd = -1;
+			fds[1].fd = -1;
+			nag->run_display = false;
+		}
+	}
+}
```
tokyo4j: This is missing `wl_display_flush()`, so requests are often buffered and not sent immediately. In fact, the cursor shape isn't updated when I hover the cursor over the labnag surface.

src/action.c
```diff
@@ -773,6 +778,112 @@ view_for_action(struct view *activator, struct server *server,
...
+		}
+
+		wlr_log(WLR_INFO, "Found pending prompt for exit code %d", exit_code);
+		struct wl_list *actions = NULL;
+		/*
+		 * TODO: figure out how to have more than 2 choices,
+		 *       maybe by using branch names like cond_0, cond_1
+		 *       and so on and falling back to 'then' and 'else'
+		 *       if not found?
+		 */
```
tokyo4j: This comment should be removed as we agreed that this is out of scope.

tokyo4j:
I wrote a patch to make `<prompt text="">` work for `ForEach` action without adding complexity: https://github.com/johanmalm/labwc/compare/nag...tokyo4j:labwc:nag-foreach.

In my patch, when `ForEach` contains `<prompt text="">`, it spawns a dialog once. And if it's accepted, it behaves just like `ForEach` which may contains `<query>` nodes. If it's rejected, it simply executes `else` and `none` branches. The first commit just splits the switch statement in `actions_run()` into `run_action()` and the second commit is the main part.

If you agree with my approach, I can open a PR for the first commit to minimize the diff in this PR.

src/action.c
```diff
@@ -491,6 +491,11 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		break;
+	case ACTION_TYPE_IF:
+		if (!strcmp(argument, "text.prompt")) {
+			action_arg_add_str(action, "prompt_text", content);
```
tokyo4j: nit: I prefer storing the argument just as `text.prompt` rather than `prompt_text`. It's consistent with how `x.position` for `ShowMenu` action is stored.

tokyo4j:
I wonder which of `<prompt>` or `<query>` should be prioritized. In my patch above, when there are both `<prompt>` and `<query>` in a `If` action, it always spawns a window dialog, and if "yes" is selected, `<query>` is checked. But this PR seems to prioritize `<prompt>`.

Consolatis:
> I wonder which of `<prompt>` or `<query>` should be prioritized. In my patch above, when there are both `<prompt>` and `<query>` in a `If` action, it always spawns a window dialog, and if "yes" is selected, `<query>` is checked. But this PR seems to prioritize `<prompt>`.

Well, technically it should depend on the order of the queries / prompts. But in practice this seems way too complicated to implement. I don't have a strong opinion about all prompts first or all queries first.

Edit:
One option that might make the the ordering work is to turn the prompt into a usual query + store it in the prompt struct. Then we could resume on the next query on success of the prompt. That would also mean that if you add 3 prompts to an `If` action it would spawn one at a time.

tokyo4j:
>  One option that might make the the ordering work is to turn the prompt into a usual query + store it in the prompt struct. Then we could resume on the next query on success of the prompt. That would also mean that if you add 3 prompts to an If action it would spawn one at a time.

That might work, but I don't know how we can manage usual queries synchronously and prompts asynchronously at the same time, without breaking `ForEach` action or adding much complexity.

I also wonder anyone would want to check prompts and queries with OR operation. I think a more reasonable situation to have both prompts and queries is like, "when pressed W-x in firefox window, open a prompt and then make it fullscreen", but this requires AND operation rather than OR.

So now I think my approach that explicitly splits prompts and queries is the way to go. When the user wants to have complex conditions like above, we can enable it by supporting nested If actions.

clients/labnag.c
johanmalm: Thanks. Have added `wl_display_flush()` in a fixup commit.

src/action.c
johanmalm: Done

johanmalm:
Rebased to resolve merge conflicts.

src/action.c
johanmalm: Done. Thanks.

johanmalm:
> I wrote a patch to make `<prompt text="">` work for `ForEach` action without adding complexity: [johanmalm/labwc@nag...tokyo4j:labwc:nag-foreach](https://github.com/johanmalm/labwc/compare/nag...tokyo4j:labwc:nag-foreach).
> 
> In my patch, when `ForEach` contains `<prompt text="">`, it spawns a dialog once. And if it's accepted, it behaves just like `ForEach` which may contains `<query>` nodes. If it's rejected, it simply executes `else` and `none` branches. The first commit just splits the switch statement in `actions_run()` into `run_action()` and the second commit is the main part.
> 
> If you agree with my approach, I can open a PR for the first commit to minimize the diff in this PR.

I like this approach.

How do we best implement? Shall I squash all the `[fixup]` comments - and then you just push to this branch?

tokyo4j:
> How do we best implement? Shall I squash all the [fixup] comments - and then you just push to this branch?

I opened #2960. Could you review & merge it first? Then you can rebase your branch on master.

tokyo4j:
Hm, I think my suggested approach was actually not the best.

I was tempted to allow `<prompt>` inside `ForEach` action after reading the FIXME comment, but now I don't think it's actually demanded. When a user wants to combine `ForEach` action and `<prompt>`, he can just use nested actions. I think it's more cleaner to not share `run_action()` for `action_check_prompt_result()` and `actions_run()`.

I updated [my branch](https://github.com/johanmalm/labwc/compare/nag...tokyo4j:labwc:nag-foreach). I also restored the behavior of not running child actions in `If` action when no window is focused, which I changed in #2960. Sorry for my bad steering!

johanmalm:
> Hm, I think my suggested approach was actually not the best.
> 
> I was tempted to allow `<prompt>` inside `ForEach` action after reading the FIXME comment, but now I don't think it's actually demanded. When a user wants to combine `ForEach` action and `<prompt>`, he can just use nested actions. I think it's more cleaner to not share `run_action()` for `action_check_prompt_result()` and `actions_run()`.
> 
> I updated [my branch](https://github.com/johanmalm/labwc/compare/nag...tokyo4j:labwc:nag-foreach). I also restored the behavior of not running child actions in `If` action when no window is focused, which I changed in #2960. Sorry for my bad steering!

Okay. Shall I rebase my nag branch and then we can apply your new commit on top of it?

tokyo4j:
> Okay. Shall I rebase my nag branch and then we can apply your new commit on top of it?

Yes, or you can write your own commit and mention my name as `Co-authored-by`.

johanmalm:
Groan. Finding it too hard to rebase on top of master now.

tokyo4j:
OK, I can help that. I will create `nag` branch to my repo so that you can use it.

EDIT: Here's my branch: https://github.com/tokyo4j/labwc/tree/nag

As #2667 seems to somehow broke parsing of `<prompt text="" />` due to the part below in `nodename()`, I renamed it to `<prompt message="" />`. I also updated corresponding documentations.

https://github.com/labwc/labwc/blob/b9c84f9c3848eabf55e0acfb670da85075ec2c62/src/common/nodename.c#L13-L16

johanmalm:
> OK, I can help that. I will create `nag` branch to my repo so that you can use it.

Thanks. That helped a lot.

Okay, I think we're finally ready for review with a view to merge.

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+	optind = 1;
+	while (1) {
+		int c = getopt_long(argc, argv, "B:Z:c:de:y:f:hlL:m:o:s:t:vx", opts, NULL);
+		if (c == -1) {
+			break;
+		}
+		switch (c) {
+		case 'B': /* Button */
+		case 'Z': /* Button (Dismiss) */
+			if (nag) {
```
tokyo4j: Can `nag` be NULL?

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+	ssize_t button_gap;
+	ssize_t button_gap_close;
+	ssize_t button_margin_right;
+	ssize_t button_padding;
+};
+
+struct nag;
+
+enum action_type {
+	LABNAG_ACTION_DISMISS,
```
tokyo4j: I guess `LABNAG_ACTION_DISMISS` is not used anymore.

tokyo4j:
When I run labnag with `labnag -m hello -Z close :` and press the close button with my touchpad, labnag is not closed immediately. It's because `poll()` is called even after `nag->run_display` is set to `false` in `wl_display_dispatch_pending()`. The patch below fixes this:

```diff
diff --git a/clients/labnag.c b/clients/labnag.c
index f0b17f77..0d2b32a2 100644
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -1192,7 +1192,7 @@ nag_run(struct nag *nag)
 			.events = POLLIN,
 		}
 	};
-	while (nag->run_display) {
+	while (true) {
 		while (wl_display_prepare_read(nag->display) != 0) {
 			wl_display_dispatch_pending(nag->display);
 		}
@@ -1202,6 +1202,10 @@ nag_run(struct nag *nag)
 			break;
 		}
 
+		if (!nag->run_display) {
+			break;
+		}
+
 		poll(fds, 2, -1);
 		if (fds[0].revents & POLLIN) {
 			if (timer_fd >= 0 && nag->details.close_timeout_cancel) {
```

tokyo4j:
I find it a bit strange that `labnag -m hello -B option1 : -B option2 :` places "option1" button at right and "option2" button at left. Can we invert the button order?

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+			break;
+		}
+		switch (c) {
+		case 'B': /* Button */
+		case 'Z': /* Button (Dismiss) */
+			if (nag) {
+				if (optind >= argc) {
+					fprintf(stderr, "Missing action for button %s\n", optarg);
+					return LAB_EXIT_FAILURE;
+				}
```
tokyo4j: Rather than requiring the action argument, can we just make it optional so that it can also function as swaynag's `-s` which was removed in response to my comment (https://github.com/labwc/labwc/pull/2699#issuecomment-2875286896)?

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+	wlr_log_init(debug ? WLR_DEBUG : WLR_ERROR, NULL);
+
+	if (!nag.message) {
+		wlr_log(WLR_ERROR, "No message passed. Please provide --message/-m");
+		exit_status = LAB_EXIT_FAILURE;
+		goto cleanup;
+	}
+
+	if (nag.details.message) {
+		nag.details.button_details = calloc(1, sizeof(struct button));
```
tokyo4j, line 1627: I think we should use `znew()` instead.

EDIT: Never mind. Currently it cannot be used since functions in `common/` are unavailable from `labnag.c`. Maybe we can link some files in `common/` to `labnag`, but we can do it later.

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+
+	return LAB_EXIT_SUCCESS;
+}
+
+static struct nag *_nag;
+
+static void
+sig_handler(int signal)
+{
+	nag_destroy(_nag);
```
tokyo4j: I know this follows `swaynag`, but I wonder if this is safe. Isn't it possible that `SIGTERM` is called while mutating `nag` and cause unexpected behavior? If so, maybe we should use `signalfd()` instead. It will also allow us to remove `_nag` static variable.

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+	struct conf *conf;
+	char *message;
+	struct wl_list buttons;
+
+	struct {
+		bool visible;
+		char *message;
+		char *details_text;
+		int close_timeout;
+		bool close_timeout_cancel;
```
tokyo4j: When is this set to false?

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+
+	if (nag->display) {
+		wl_display_disconnect(nag->display);
+	}
+	pango_cairo_font_map_set_default(NULL);
+}
+
+static void
+button_execute(struct nag *nag,
+		struct button *button)
```
tokyo4j: Nit: unnecessary indentation. Same goes for other functions.

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+			break;
+		}
+
+		poll(fds, 2, -1);
+		if (fds[0].revents & POLLIN) {
+			if (timer_fd >= 0 && nag->details.close_timeout_cancel) {
+				close(timer_fd);
+				timer_fd = -1;
+				fds[1].fd = -1;
+			}
```
tokyo4j: This feels like a bit hack to me. I think the timer for auto-closing should be cancelled by user inputs, not by wayland events including output plug/unplug. I think we should do this in `wl_pointer.motion` event handler.

EDIT: I guess this can be more problematic when we extend labnag to support button selection with keyboards (probably with layer-shell's exclusive keyboard interactivity), because we broadcast modifiers to all the clients which have created `wl_keyboard`.

clients/labnag.c
```diff
@@ -0,0 +1,1649 @@
...
+
+	if (nag->details.visible) {
+		h = render_detailed(cairo, nag, max_height);
+		max_height = h > max_height ? h : max_height;
+	}
+
+	int border = nag->conf->bar_border_thickness;
+	if (max_height > nag->height) {
+		max_height += border;
+	}
```
tokyo4j, line 488: What is this `if` branch for? This is causing 1 more call to `wl_display_roundtrip()` in `render_frame()` when the detail is being opened.

tokyo4j:
Not sure we should do this in this PR, but I think we need to set a limit for the detail message. When I passed the output of `labwc -d` which is 517 lines long into stdin of labnag, opening the detail message is noticeably slow, and scrolling the message is painfully slow and takes 100% of CPU usage.

clients/pool-buffer.h
```diff
@@ -0,0 +1,30 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+/*
+ * Copied from https://github.com/swaywm/sway
+ *
+ * Copyright (C) 2016-2017 Drew DeVault
+ */
+#ifndef BUFFERS_H
+#define BUFFERS_H
```
tokyo4j: I suggest more unique values like `LAB_POOL_BUFFER_H`

johanmalm:
Thanks for the review. Very thorough. I'll work through the comments, but might take a few days.

clients/pool-buffer.h
johanmalm: Fixed. Used `LAB_POOL_BUFFER_H`

johanmalm:
@tokyo4j Thanks for your `nag` branch. I've cherry-picked those four commits.

Also updated `labnag(1)` to reflect the change relating to the `-B/-Z` options (optional action argument and the order the buttons are shown).

Have squashed, but updated the commit message with a `v3` section to reflect this.

johanmalm:
I had to add a commit to make it build on FreeBSD. Lucky we've got CI :smile:

Ready for final review again.

@tokyo4j I think all your comments above have been answered (almost entirely by the four commit on your nag branch :smile:)

tokyo4j:
I hid my comments that are now resolved. Remaining are:

1. Remove `close_timeout_cancel`?
2. Remove unnecessary indentations
3. Remove `if (max_height > nag->height)` if it's not needed
4. Limit the length of detail messages to prevent overloading CPU

I think (1) and (2) are easily addressed. I'm not sure about (3) and (4). If they are not easy, let's add a `TODO` comment and address them later.

LGTM other than that. Actually I don't understand all the logic of `labnag.c`, but everything seems to be working well in my testing.

clients/labnag.c
johanmalm: Thanks. Have removed redundant variable `close_timeout_cancel`

clients/labnag.c
johanmalm: Have fixed indentation at `button_excecute()`

johanmalm:
> Not sure we should do this in this PR, but I think we need to set a limit for the detail message. When I passed the output of `labwc -d` which is 517 lines long into stdin of labnag, opening the detail message is noticeably slow, and scrolling the message is painfully slow and takes 100% of CPU usage.

I have limited the detailed message (via stdin) to 200 lines.

Tested with `seq 10000 | build/clients/labnag -m foo -l -Z Exit`

clients/labnag.c
johanmalm: Yeah, I think the logic could be improved, simplified, re-factored... but I have not touched this in the basis that it's stood the test of time and I feel it's safer at this stage rather than embarking on an overhaul

tokyo4j: OK, let's keep it as it is

johanmalm:
Have adjusted indentation throughout `labnag.c`.
Also amended `--help` message to reflect that the action argument is now optional.

PR #2700, opened 2025-04-29
Parse touch configs in reverse order to enable overriding

spl237:
I'm not sure if this is the best fix for the problem I have found, but it is the shortest...

If multiple touch configurations for the same device are defined - so more than one is defined in /etc/xdg/labwc/rc.xml, or one is defined in /etc/xdg/labwc/rc.xml and another is defined in ~/.config/labwc/rc.xml, I would expect the one which is read in last to be the one which takes effect. So I would expect the config in the local file to override that in the global file, or a config at the bottom of the global file to override one higher up in the file.

But what happens is that the first config found is the one which takes effect. This means that it is impossible to override a global config for a touch device with one in the local config file; the one in the global file takes precedence.

This seems to be handled in the cases of mouse and keyboard keybinds by the deduplicate functions which are used to validate the bindings after they are loaded, which may be a better solution, but simply reversing the order in which the loaded touch configs are parsed so they are searched from end to start for the relevant device rather than from start to end seems to work for me.

### DIFF
--- a/src/config/touch.c
+++ b/src/config/touch.c
@@ -9,7 +9,7 @@ static struct touch_config_entry *
 find_default_config(void)
 {
 	struct touch_config_entry *entry;
-	wl_list_for_each(entry, &rc.touch_configs, link) {
+	wl_list_for_each_reverse(entry, &rc.touch_configs, link) {
 		if (!entry->device_name) {
 			wlr_log(WLR_INFO, "found default touch configuration");
 			return entry;
@@ -23,7 +23,7 @@ touch_find_config_for_device(char *device_name)
 {
 	wlr_log(WLR_INFO, "find touch configuration for %s", device_name);
 	struct touch_config_entry *entry;
-	wl_list_for_each(entry, &rc.touch_configs, link) {
+	wl_list_for_each_reverse(entry, &rc.touch_configs, link) {
 		if (entry->device_name && !strcasecmp(entry->device_name, device_name)) {
 			wlr_log(WLR_INFO, "found touch configuration for %s", device_name);
 			return entry;


ahesford:
The order in which rules are applied (first wins, last wins) arbitrary---SSH configs, for example, are first-win to allow more specific rules up front and broader rules to follow, setting only what hasn't been set---but should be consistent across our entire configuration parser. If the other config sections are last-win, so should these be.

However, simply reversing the order of parsing may have unintended consequences, either now or in the future, if there is ever an interdependency between rules at different levels. For example, bindings allow a default rule specification followed by specific overrides, which would break if we simply reverse the processing sequence.

Furthermore, the handling of multiple-level merging should always be consistent across all sections. If the other sections process in general-to-specific order and provide a mechanism to override previously set variables, this section should be processed in the same fashion.

tokyo4j:
You should also reverse the searching order in `find_default_config()`. In this way, the touch config is searched with the rules:
- Device-specific touch configs are always prioritized over default configs
- When duplicated device-specific or default configs are found, ones defined later are prioritized

This is consistent with how `<libinput><device>` sections are applied.

spl237:
> The order in which rules are applied (first wins, last wins) arbitrary---SSH configs, for example, are first-win to allow more specific rules up front and broader rules to follow, setting only what hasn't been set---but should be consistent across our entire configuration parser. If the other config sections are last-win, so should these be.

That is my understanding, and indeed we had a discussion on this last year which came to the conclusion that local settings should override global ones : https://github.com/labwc/labwc/issues/1406 , and this was then implemented.
> 
> However, simply reversing the order of parsing may have unintended consequences, either now or in the future, if there is ever an interdependency between rules at different levels. For example, bindings allow a default rule specification followed by specific overrides, which would break if we simply reverse the processing sequence.
> 
> Furthermore, the handling of multiple-level merging should always be consistent across all sections. If the other sections process in general-to-specific order and provide a mechanism to override previously set variables, this section should be processed in the same fashion.

As above - I am not proposing that this is the correct solution - I will be the first to admit that it is quick and dirty, and that a deduplication mechanism like that on keyboard and mouse bindings may be more appropriate, but I wanted to raise the issue with at least a potential solution.

tokyo4j:
> I've been looking through the libinput code, and it looks to me as if this behaviour is purely enabled by the use of wl_list_for_each_reverse in libinput_category_get_default - so doing what I am doing in touch_find_config_for_device, changing wl_list_for_each to wl_list_for_each_reverse, looks to be directly analogous and will result in the same behaviour - is that correct? So it may be all that is required here, with the addition of also reversing the search, as you point out, in find_default_config.

I'm not sure if I understand your question correctly, but the lookup of `<libinput><device>` is done by `get_category()`:

https://github.com/labwc/labwc/blob/58d22d05e22745a626ab1ec66aa1cb8802d12812/src/seat.c#L67-L90

...and `libinput_category_get_default()`:

https://github.com/labwc/labwc/blob/58d22d05e22745a626ab1ec66aa1cb8802d12812/src/config/libinput.c#L63-L77

So just like you've done, reversing the searching order in `touch_find_config_for_device()` and `find_default_config()`, which are very similar to the functions above, will make the parsing of `<libinput><device>` and `<touch>` consistent.

Could you squash the 2 commits into 1 commit? LGTM other than that.

@johanmalm +1 for merging this PR before 0.8.4, but it's up to you. I tested this PR on my laptop (which has builtin touchscreen) with configurations like:
```xml
  <touch deviceName="ELAN2514:00 04F3:2AF1" mapToOutput="eDP-1" />
  <touch deviceName="ELAN2514:00 04F3:2AF1" mapToOutput="DP-1" />
  <touch mapToOutput="eDP-1" />
  <touch mapToOutput="DP-1" />
```

and it worked as expected when I reordered the lines above.

spl237:
OK, will squash the commits in the morning...

johanmalm:
I see the point @ahesford is making here - so we ought to make sure we're not affected (nor setting a trap for the future). The man page states that `If device name is left empty, the touch configuration applies to all touch devices`. Are we sure that reversing the order does not introduce a breaking change?

```
<touch mapToOutput="eDP-1" />
<touch deviceName="foo" mapToOutput="DP-1" />
```

Would `foo` be mapped to the same if we reverse the order?

@tokyo4j I'm not sure if the above is indeed what your answer addressed. Confirmation appreciated.

spl237:
Commits now squashed.

johanmalm:
> Commits now squashed.

Thanks @spl237.

I won't squash until someone can convince me that we're not breaking something. Or I'll test/analyze myself - but might not be in time for Friday :smile:

I think the correct solution is to de-duplicate when parsing, but will accept above if a unanimous, confident position is achieved.

spl237:
I don't pretend to know the code well enough to comment, so will leave the decision to wiser heads than mine - but let me know if there is anything anyone wants me to do / change / test.

tokyo4j:
Hmm, I might be a bit too focused on "not introducing new bugs" but not on "not making breaking changes".

> Are we sure that reversing the order does not introduce a breaking change?

This PR is _breaking_, though its impact is minimal IMO. Consider strange configurations like:

```xml
  <touch deviceName="foo" mapToOutput="eDP-1" />
  <touch deviceName="foo" mapToOutput="DP-1" />
```

or

```xml
  <touch mapToOutput="eDP-1" />
  <touch mapToOutput="DP-1" />
```

This PR makes the touch device mapped to "DP-1" instead of "eDP-1" with the both configurations above.

> ```
> <touch mapToOutput="eDP-1" />
> <touch deviceName="foo" mapToOutput="DP-1" />
> ```
> 
> Would foo be mapped to the same if we reverse the order?

Yes, `foo` is mapped to the same output even after this PR. This PR just reverses the search order of multiple `<touch>` sections with the `deviceName`, or multiple `<touch>` sections without `deviceName`.

I think we can merge this PR before 0.8.4 since I don't see any potential for regressions and it's only breaking for unusual configurations. But as I said, it's up to you. I suppose this PR is important mainly for RPi, who it sells its own touch devices and uses `labwc -m` to let users overwrite their configuration, but RPi team can also apply this patch to their labwc branch even if we give up merging this PR before 0.8.4.

> I think the correct solution is to de-duplicate when parsing, but will accept above if a unanimous, confident position is achieved.

I'm not sure how cleaner it is to de-duplicate the `<touch>` sections rather than reversing the search order. I'm not against it, but to me just reversing order is easy enough to follow and maintain.

Consolatis:
From memory: we were using the same trick for mouse and keybinds long ago (pre scene-graph, e.g. just reversing the lookup to find a matching keybind). What changed there was that at some point we added the ability to remove bindings by having an empty bind or one with the `None` action only. For that to work slightly faster at runtime (e.g. reducing the amount of bindings to loop through) we added the de-duplication logic. I don't remember any other reason for it.

johanmalm:
Right, let's go for it then :smile:

PR #2710, opened 2025-05-03
rcxml: validate total osd field width

tokyo4j:
Before this patch, window switcher fields could overflow if the sum of `<windowSwitcher><fields><field width="">` exceeds 100%.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1941,9 +1941,11 @@ validate(void)
 	validate_actions();
 
 	/* OSD fields */
+	int field_width_sum = 0;
 	struct window_switcher_field *field, *field_tmp;
 	wl_list_for_each_safe(field, field_tmp, &rc.window_switcher.fields, link) {
-		if (!osd_field_validate(field)) {
+		field_width_sum += field->width;
+		if (!osd_field_is_valid(field) || field_width_sum > 100) {
 			wlr_log(WLR_ERROR, "Deleting invalid window switcher field %p", field);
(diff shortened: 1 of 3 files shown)


src/config/rcxml.c
```diff
@@ -1941,9 +1941,20 @@ validate(void)
...
+
+		invalid_osd_field |= !osd_field_is_valid(field);
+
+		if (field_width_sum > 100) {
+			wlr_log(WLR_ERROR, "sum of osd field widths must not exceed 100%%");
+			invalid_osd_field |= true;
+		}
+
+		if (invalid_osd_field) {
 			wlr_log(WLR_ERROR, "Deleting invalid window switcher field %p", field);
```
Consolatis, line 1958: Suggest something like
```c
bool is_valid = osd_field_is_valid(field);
if (field_width_sum > 100) {
	wlr_log(..);
	is_valid = false;
}
if (!is_valid) {
	..
}
```
Or maybe even
```c
bool is_valid = osd_field_is_valid(field) && field_width_sum <= 100;
```
Although that would miss the comment.

Consolatis:
Small nitpick above but other than the bitwise op + double negative this LGTM

src/config/rcxml.c
tokyo4j: I prefer the latter approach. I thought we should print a error message for overflowing fields, but users will easily see what's wrong without it. In fact, before #2620, we were just drawing the overflowing fields outside of the buffer region.

PR #2711, opened 2025-05-03
ssd-shadow: flip top-right/bottom-left corner instead of rotating

tokyo4j:
This simplifies `corner_scale_crop()`.

Just for testing, this PR also includes a commit that paints the corner shadow buffer with green.

CC @cillian64


cillian64:
Thanks for this, I'm away for a long weekend but can have a look on Tuesday.

cillian64:
Changes make sense to me and work fine in my testing - thanks for the improvement!

Consolatis:
Tested nested with pixman and gles2 and didn't spot a difference. With the 2nd commit removed this LGTM.

Issue #1526, opened 2024-02-12
labwc may crash if starts without any installed fonts


dmitry-azaraev:
Hello.

During new minimal Arch Linux install, i'm hit into crash when run labwc. After some trial and error, found what it is because my system have no any fonts installed. Interestingly what it been able to run without fonts at 5-10% rate. I'm okay with such behavior, but would be nice to have bit more meaningful error.

```
Core was generated by `labwc'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005981116d1e99 in rounded_rect (ctx=ctx@entry=0x7fff33bc5ed0) at ../labwc/src/theme.c:772
772		cairo_t *cairo = buffer->cairo;
[Current thread is 1 (Thread 0x773b25eceac0 (LWP 1408))]
(gdb) bt
#0  0x00005981116d1e99 in rounded_rect (ctx=ctx@entry=0x7fff33bc5ed0) at ../labwc/src/theme.c:772
#1  0x00005981116dce8c in create_corners (theme=0x7fff33bc6100) at ../labwc/src/theme.c:937
#2  theme_init (theme=0x7fff33bc6100, theme_name=<optimized out>) at ../labwc/src/theme.c:1033
#3  0x00005981116c388d in main (argc=<optimized out>, argv=<optimized out>) at ../labwc/src/main.c:168
```

Consolatis:
Do you run with a border width of zero by chance?

Consolatis:
Strange, the rounded rects should have nothing to do with font sizes but it looks like allocating the buffer fails for some reason (which usually is a `0` width or height). Not sure how to reproduce this currently without breaking my system, maybe there is some env var one could set to prevent cairo / pango to find the fonts or something..

dmitry-azaraev:
@Consolatis it should be easy:

Create empty config file:
```xml
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "urn:fontconfig:fonts.dtd">
<fontconfig>

</fontconfig>
```

Then run:
`FONTCONFIG_FILE=${HOME}/path_to_empty_config.conf labwc`

Consolatis:
Yep, that seems to work when running nested: getting lots of "Failed to create font buffer" in the logs but no crash so far. I may investigate this later in more detail.

Consolatis:
I am not sure how to reproduce this currently so a potential fix will likely not land in the next release. I'll move the issue to the followup release.

[...]

PR #2713, opened 2025-05-08
Gracefully exit when no fonts are installed

tokyo4j:
...rather than emitting ugly errors like:

```
labwc: ../src/buffer.c:85: buffer_adopt_cairo_surface: Assertion `cairo_image_surface_get_format(surface) == CAIRO_FORMAT_ARGB32' failed.
```

Ref: #1526, [labwc-git AUR](https://aur.archlinux.org/packages/labwc-git#comment-984158)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -262,3 +262,15 @@ jobs:
             meson compile -C build-gcc-gdb
             LABWC_RUNS=2 scripts/ci/smoke-test.sh build-gcc-gdb
           ' | $TARGET
+
+      - name: Build with gcc - catch no font installed case
+        if: matrix.name == 'Void-musl'
+        run: |
+          echo '
+            cd "$GITHUB_WORKSPACE"
+            xbps-remove -y dejavu-fonts-ttf
+            export CC=gcc
+            meson setup build-gcc-nofont -Dxwayland=enabled --werror
+            meson compile -C build-gcc-nofont
+            LABWC_EXPECT_RETURNCODE=1 scripts/ci/smoke-test.sh build-gcc-nofont
+          ' | $TARGET
--- a/scripts/ci/smoke-test.sh
+++ b/scripts/ci/smoke-test.sh
@@ -2,6 +2,8 @@
 
 : ${LABWC_RUNS:=1}
 : ${LABWC_LEAK_TEST:=0}
+: ${LABWC_EXPECT_RETURNCODE:=0}
+: ${LABWC_VERBOSE:=0}
 
 if ! test -x "$1/labwc"; then
 	echo "$1/labwc not found"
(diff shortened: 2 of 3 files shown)


Consolatis:
If this works to detect those random results from `font_height()` I am all for it. But the failed CI runs for Arch and Void make me think this might actually be worse now, something is definitely fishy with pango.

tokyo4j:
> If this works to detect those random results from `font_height()` I am all for it.  But the failed CI runtime tests for Arch and Void make me think this might actually be worse now, something is definitely fishy with pango.

Agreed. I'll investigate the cause of CI failures and hopefully fix it in this release cycle.

Consolatis:
Maybe it is also due to the smoke-test environment. It starts foot for a second and then terminates everything (see https://github.com/labwc/labwc/blob/master/scripts/ci/ci_autostart.sh). Maybe that is not enough time for pango initialization or something.

tokyo4j:
I think I found the cause.

Before this PR:
- `main()` > `server_init()` > `wl_event_loop_add_signal()` masks `SIGTERM` and registers our handler for it with `signalfd()`.
- `main()` > `theme_init()` > ... > `font_extents()` > `pango_cairo_create_layout()` spawns pango's thread

In this case, the state of masked signals are inherited to the pango's thread, so it never receives `SIGTERM`.

After this PR:
- `main()` > `die_on_no_fonts()` > `pango_cairo_font_map_get_default()` spawns pango's thread
- `main()` > `server_init()` > `wl_event_loop_add_signal()` masks `SIGTERM` and registers our handler for it

In this case, `SIGTERM` is not masked in the pango's thread, so GDB sees `SIGTERM` delivered to the pango's thread and exits with status 255.

I'm not sure what's the cleanest way to fix this... maybe we should temporarily mask signals at the start of `main()`?

tokyo4j:
> If this works to detect those random results from font_height() I am all for it.

I tested this PR on my server without any fonts installed and it successfully detected that no fonts are available.

Also, I tried installing `ttf-dejavu` (only contains basic characters) and rendering CJK fonts in the titlebar. This case is not covered by `die_on_no_fonts()`, but pango/fontconfig seems to successfully render a fallback font like below and I don't think we need to exit in this case:

20250509_17h08m10s_grim

Consolatis:
Nice! Let me add a missing font test-case to the CI to ensure it also works there. If you are fine with it I'd push it to this PR in the next 24 hours.

tokyo4j:
> If you are fine with it I'd push it to this PR in the next 24 hours.

Yes, that sounds nice!

Consolatis:
Seems to work as expected: https://github.com/labwc/labwc/actions/runs/14930660893/job/41945842419?pr=2713.
LGTM, Looks like we finally found a solution to that annoying issue :)

PR #2727, opened 2025-05-17
Guard against negative sizes in `wlr_scene_rect/buffer`

tokyo4j:
This PR ensures that the sizes set in `wlr_scene_{rect,buffer}` never goes negative, which crashes in wlroots 0.19.

The 1st commit temporarily updates `wlroots.wrap` to depend on my fork of wlroots which has cherry-picked the commits that added abortions for negative sizes. This should be dropped before merging.

The 2nd commit guards against negative sizes in menu scenes (possible with `menu.width.{min,max}: 1`). It checks the lengths marked with red in the diagram below:

menu-coordinates

Likewise, the 3rd commit handles negative sizes in window switcher OSD (possible with `osd.window-switcher.width: 1`):

osd-coordinates

The 4th commit handles negative sizes in SSD by just adding `MAX()` macros. Another possible solution suggested by @Consolatis in IRC is to enforce minimal window size so that sizes of components like edge drop-shadows never become negative, but I prefer just adding `MAX()` macros because:
- Enforcing minimal window sizes via `xdg_toplevel.set_min_size` doesn't prevent ridiculously small window sizes, since some clients (e.g. workrave #1947) can still submit tiny buffers. Maybe we can add some padding between the client surface and SSD like #1959 in such cases to prevent broken look, but I prefer the current behavior and implementation.
- We are always calling `ssd_create()` for 0x0 window size when a client enables decorations before mapping a surface. If we try to prevent negative sizes by not creating SSD for tiny windows, we will need change this to call `ssd_create()` on map, which sounds a bit annoying to address.

The 5th commit handles negative sizes in `multi_rect`, which can happen when a tiny window is selected with window switcher and `osd.window-switcher.preview.border.width` is large, by just adding `MAX()` macros. This can be improved by #2241 which hides inner outlines when the rectangle size is smaller than (border width) * 6 and bigger than (border width) * 2.

The 6th commit adds asserts for negative sizes passed to `scaled_*_buffer`s to make it easier to find bugs in the future.

### DIFF
--- a/src/common/graphic-helpers.c
+++ b/src/common/graphic-helpers.c
@@ -8,6 +8,7 @@
 #include <wlr/util/box.h>
 #include "buffer.h"
 #include "common/graphic-helpers.h"
+#include "common/macros.h"
 #include "common/mem.h"
 
 static void
@@ -57,7 +58,7 @@ multi_rect_set_size(struct multi_rect *rect, int width, int height)
 	 * +-+-----+-+   |
 	 * +---------+  ---
 	 */
-	for (size_t i = 0; i < 3; i++) {
+	for (int i = 0; i < 3; i++) {
 		/* Reposition, top and left don't ever change */
 		wlr_scene_node_set_position(&rect->right[i]->node,
 			width - (i + 1) * line_width, (i + 1) * line_width);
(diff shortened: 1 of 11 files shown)


johanmalm:
LGTM. Not tested.

Should we remove the first commit and merge this before the 0.19 port?

Those pictures are great. We ought to consider adding them to the documentation.

tokyo4j:
> Should we remove the first commit and merge this before the 0.19 port?

Yes, I prefer handling all the negative sizes in this PR and focusing on 0.19 port in #2388 without causing abortions.

> We ought to consider adding them to the documentation.

Agreed. I've been wondering how we should upload diagram files regarding https://github.com/labwc/labwc.github.io/pull/68, but now I think we can just upload plain `.svg` or `.png` files and help users and contributors.

PR #2728, opened 2025-05-17
xwayland.c: support both axis for client side maximize requests

Consolatis:
We used `view_toggle_maximize()` before which was always wrong. It didn't matter much because it's only relevant for repeated client side maximize requests.

With wlroots 0.18.x we still only get the `request_maximize` event when either both axis are maximized or not but with wlroots 0.19 we will get the event on any change of the two bools.

Split off from
- #2388

CC @jlindgren90


jlindgren90:
LGTM

Issue #2720, opened 2025-05-15
Login sound in labwc


tuxslack:
**Login sound is not working in labwc, it worked in openbox**

$ ls /usr/share/sounds/freedesktop/stereo/service-login.oga
/usr/share/sounds/freedesktop/stereo/service-login.oga

$ cat ~/.config/labwc/autostart

play /usr/share/sounds/freedesktop/stereo/service-login.oga &

Labwc starts without sound.

cillian64:
IIRC pipewire only starts with the graphical session, so it could be that the sound is trying to play before pipewire is ready.

Just as an experiment, does it work if you put a `sleep 5` or `sleep 10` line before `play` in  the autostart file?

stefonarch:
Not a big fan of talking DE's but adding `mpv  /usr/share/sounds/freedesktop/stereo/service-login.oga` in LXQt's autostart is working fine here.

tuxslack:
$ cat ~/.config/labwc/autostart

...

exec pipewire &



**Method 1 (It didn't work):**

exec (sleep 3 && play /usr/share/sounds/freedesktop/stereo/service-login.oga) &


**Method 2 (It didn't work):**

sleep 3 && play /usr/share/sounds/freedesktop/stereo/service-login.oga &


**Method 3 (It worked):**

sleep 3 ; play /usr/share/sounds/freedesktop/stereo/service-login.oga &

mpv /usr/share/sounds/freedesktop/stereo/service-login.oga &

...


Could the problem be the use of && in the file?

**I replaced "&&" with ";"**

johanmalm:
We fundamentally just run `sh autostart` (via `fork()` and `execvp()`)

`labwc` is agnostic to what you put in `autostart` but I suggest the following:

1. Don't use `exec` as that  replaces the shell with `command` without creating a  new  process. See `man 1p exec`.
2. If you want to delay the start of a particular client, use the syntax `( sleep X ; foo >/dev/null 2>&1 ) &`. We should probably add a note in the example autostart to that effect.

Ref:
- https://github.com/labwc/labwc/blob/05ff914cfe0b76973ac23e4b6e44b6822358b040/src/config/session.c#L306-L307
- https://github.com/labwc/labwc/blob/05ff914cfe0b76973ac23e4b6e44b6822358b040/src/common/spawn.c#L47

We're trying to keep issues for bugs only. This seems unlikely to be a labwc bug, so will close on that basis. Please say if you think otherwise.

Ref:
[...]

PR #2731, opened 2025-05-20
docs/autostart: add example for delayed start

johanmalm:


Consolatis:
I am kind of skeptical about actual user needs for a sleep in the autostart file. Labwc should be completely initialized at the point the autostart file is executed.

I also feel like this is more of generic shell documentation rather than labwc specific.

johanmalm:
Reasons I thought it would be useful:
- A user was struggling with it last week (#2720) needing to wait for `pipewire` before issusing `play`. I considered adding it to the faq, but thought the core repo is generally better.
- This was a common usage pattern described in Openbox forums - and it's not necessarily obvious to users.
- I found myself needing it some time ago with `xrdb` and an XWayland client that relied on it.

Consolatis:
In that case, LGTM.

sefernet:
Im a user with a case : launching a notification before the actual command (ie reboot/shutdown)

I tried to test this :
```
 <action name="Execute" command="notify-send  Reboot ; sleep 3; notify-send  HI " />
```

Or :
```
 <action name="Execute" command="notify-send  Reboot ; sh sleep 3; notify-send  HI" />
```

And didnt work either, the log says:
```
 [../src/action.c:1536] Handling action 5: Execute Invalid number of options.
```

And when split this in several command lines, then the commands run without the waiting.

There is a way to achieve this ?

Thanks!

Consolatis:
The `Execute` action is not executed within a shell so things like `;` won't work. You could do something like this instead
```xml
<action name="Execute" command="sh -c 'notify-send one; sleep 3; notify-send two'" />
```
or in the longer form which makes the quotes a bit easier to manage and/or allows to use quotes within your sh command:
```xml
<action name="Execute">
    <command>sh -c 'notify-send "first message"; sleep 3; notify-send "second message"'</command>
</action>
```

And lastly you could throw it into a script instead and then call the script with the action.

PR #2734, opened 2025-05-21
config: drop xwayland persistence workaround for wlroots < 0.18.2

tokyo4j:
We enabled xwayland persistence for wlroots < 0.18.2 to prevent crashes, but it's no longer needed since we now depend on wlroots 0.19.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -173,7 +173,7 @@ this is for compatibility with Openbox.
   <allowTearing>no</allowTearing>
   <autoEnableOutputs>yes</autoEnableOutputs>
   <reuseOutputMode>no</reuseOutputMode>
-  <xwaylandPersistence>yes</xwaylandPersistence>
+  <xwaylandPersistence>no</xwaylandPersistence>
 </core>
 ```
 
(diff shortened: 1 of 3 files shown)


docs/labwc-config.5.scd
```diff
@@ -228,12 +228,7 @@ this is for compatibility with Openbox.
 *<core><xwaylandPersistence>* [yes|no]
 	Keep XWayland alive even when no clients are connected, rather than
 	using a "lazy" policy that allows the server to launch on demand and die
-	when it is no longer needed.
-
-	When labwc is compiled with wlroots 0.18.2 or later, the default is no.
-	When labwc is compiled with wlroots < 0.18.2, the default is yes to
-	avoid a wlroots bug that can result in compositor crashes when
-	performing drag-and-drop actions when the XWayland server is shut down.
+	when it is no longer needed. Default is yes.
```
Consolatis, line 231: Should be: "Default is no."

src/config/rcxml.c
```diff
@@ -1127,14 +1123,6 @@ entry(xmlNode *node, char *nodename, char *content, struct parser_state *state)
...
 		set_bool(content, &rc.xwayland_persistence);
 
-#if LAB_WLR_VERSION_OLDER_THAN(0, 18, 2)
-		if (!rc.xwayland_persistence) {
-			wlr_log(WLR_ERROR, "to avoid the risk of a fatal crash, "
-				"setting xwaylandPersistence to 'no' is only "
-				"recommended when labwc is compiled against "
-				"wlroots >= 0.18.2. See #2371 for details.");
-		}
-#endif
```
Consolatis, line 1137: Stray empty line above this block can be removed as well.

Consolatis:
Two nitpicks above, other than that LGTM.

src/config/rcxml.c
tokyo4j: Removed the empty line. Also reordered the if-statement for authentics.

PR #2735, opened 2025-05-21
xwayland: use wlr_xwayland_surface_has_window_type()

jlindgren90:
This eliminates a bit of logic, including an extra XWayland connection.

See also:
https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4553

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -84,8 +84,15 @@ enum view_wants_focus {
 	VIEW_WANTS_FOCUS_OFFER,
 };
 
+/*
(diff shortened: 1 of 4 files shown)


Consolatis:
LGTM, not tested.

Also nice to see the additional xcb connection go away. We may need access to the xwm xcb connection again at some point in the future but since wlroots 0.18 we can actually use that one instead of having to connect ourselves.

johanmalm:
Excellent.

PR #2743, opened 2025-05-22
treewide: remove empty statements in switch cases

jlindgren90:
To address:
- https://github.com/labwc/labwc/discussions/2739

For longer cases, factor out the logic to new functions. For very short cases, just move the declaration before the switch.


jlindgren90:
v2: in one case, replace the switch with if/else.

johanmalm:
Thanks.

PR #2745, opened 2025-05-22
layer-shell: only configure initialized layer surfaces

### LINKED ISSUE #2736: unmapping layer surface crash
### Problem description

After a layer surface has unmapped itself, labwc will still try to send configure events to the unmapped surface when arranging all the layers. This causes labwc to crash because of a failed assertion in `wlr_scene_layer_surface_v1_configure`.

### Steps to reproduce

To reproduce you need to run a layer shell program that unmaps itself. The kitty terminal has a new feature that does this. With the latest version of kitty installed, bind `kitten quick-access-terminal` to some keybinding and activate this binding twice. labwc should now have crashed.

### labwc build source

Local build

### labwc version

labwc 0.8.4-38-g95552c26

### labwc environment

### Distribution

Arch Linux

alex-huff:
Fixes #2736

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -46,6 +46,9 @@ arrange_one_layer(const struct wlr_box *full_area, struct wlr_box *usable_area,
 	wl_list_for_each(node, &tree->children, link) {
 		struct lab_layer_surface *surface = node_layer_surface_from_node(node);
 		struct wlr_scene_layer_surface_v1 *scene = surface->scene_layer_surface;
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. LGTM.

Matches sway, which is good for confidence.

https://github.com/swaywm/sway/blob/534491d3aa0c60d5c941f2f46b5d7f8978901f84/sway/desktop/layer_shell.c#L67

PR #2747, opened 2025-05-22
layer-shell: null check surface before configure

johanmalm:
...to avoid crash during node/surface destruction.

No crash has been observed, so this is merely a preventative measure.


Consolatis:
I don't think they can be `NULL` here: `node_layer_surface_from_node()` would run into an assert in that case.

johanmalm:
What if we do #1154?

tokyo4j:
Now that #1154 has been merged, can we close this PR now?

PR #2749, opened 2025-05-23
Add `-Wshadow` + reformat switch cases

tokyo4j:
This PR adds `-Wshadow` compiler option to prevent unintentional variable overrides.

This PR also reformats switch cases with variable declarations by wrapping them with `{}`. This will make our logic more robust by limiting lifetimes of variables. I believe this syntax got some consensus in #2739.

I also documented the use of braces in switch statements in `CONTRIBUTING.md`.


Consolatis:
Very nice, also tested to compile on gcc 10.2 and no issues.

LGTM

tokyo4j:
Thanks, but let's wait for reaction from @johanmalm.

johanmalm:
I like this. Thanks for grabbing + running with it.

PR #2751, opened 2025-05-23
server.c: add missing privileged protocols

Consolatis:
Also use `ext-workspace-v1.xml` from upstream wayland-protocols rather than carrying around our own copy.


johanmalm:
Thanks

PR #2753, opened 2025-05-24
doc: s/view/window/

zeusgoose:
for the wlr18 to 19 check list
ref: https://github.com/swaywm/sway/commit/10e50e6bf9b63b205c141f97a5709fd4d405542f

### DIFF
--- a/docs/autostart
+++ b/docs/autostart
@@ -29,7 +29,7 @@ mako >/dev/null 2>&1 &
 #
 # Note that in the context of idle system power management, it is *NOT* a good
 # idea to turn off displays by 'disabling outputs' for example by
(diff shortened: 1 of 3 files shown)


johanmalm:
Thanks for contributing. Looks good, except that we prefer a prefix in the commit message. Ref: https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#commit-messages

Suggest the following to be consistent with `git log --oneline` history.

- `docs/autostart:`
- `labwc-config(5):`
- `labwc-menu(5):`

Just shout if git isn't your first language. Suggest `git rebase -i HEAD~3`

PR #2760, opened 2025-05-25
xwayland: use wlr_xwayland_surface_fetch_icon()

tokyo4j:
Based on #2755.

This PR currently depends on my fork of wlroots: https://gitlab.freedesktop.org/tokyo4j/wlroots/-/tree/net-wm-icon

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -25,18 +25,6 @@
 #include "window-rules.h"
 #include "workspaces.h"
 
-enum atoms {
-
-	ATOM_COUNT,
-};
-
-static const char * const atom_names[] = {
-};
-
-static_assert(ARRAY_SIZE(atom_names) == ATOM_COUNT, "atom names out of sync");
-
-static xcb_atom_t atoms[ATOM_COUNT] = {0};
-
 static void set_surface(struct view *view, struct wlr_surface *surface);
 static void handle_map(struct wl_listener *listener, void *data);
 static void handle_unmap(struct wl_listener *listener, void *data);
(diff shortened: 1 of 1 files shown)


johanmalm:
Nice!

I haven't tested, but do we want to make this optional?

Take xterm as an example - it might be preferable to use the xterm.{png,svg} supplied by the icon-theme rather than the built in one (which isn't that pretty imho).

Consolatis:
wlroots MR: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5079

I'll investigate if we can implement the X11 bits on the labwc side until we start tracking wlroots 0.20.

Consolatis:
I've updated my branch and cherry-picked your commit on top + slightly modified it to fit the new xdg icon implementation (e.g. mostly `struct *wl_array` arg rather than `struct wl_array` + changed the signal source). The original proof of concept commit was updated to emulate the wlroots one. If you are fine with it I could force push it to this PR. This should work with wlroots 0.19.x and be easy to update once your wlroots MR is merged + we started to track wlroots 0.20.x.

Edit:
Moved some changes accidentally added to the cherry picked commit into the first one, no change in the code itself.

tokyo4j:
wlroots PR has just been merged. Let's use `wlr_xwayland_surface_fetch_icon()` when wlroots 0.20 is released.

Consolatis:
Master is on 0.20 now, this PR needs a rebase.

jlindgren90:
Rebased here: https://github.com/jlindgren90/labwc/commit/dee3546516be1e2f0a0ea1ebc68f8fed1eabada6

src/xwayland.c
Consolatis: Unrelated to this PR and no functional change but `stride` is usually defined as bytes-per-row. so it should be
```c
size_t stride = iter.width * 4;
uint32_t *buf = xzalloc(iter.height * stride);
```

Or just
```c
uint32_t *buf = xzalloc(iter.width * iter.height * 4);
```

tokyo4j: That was fixed by #2909

tokyo4j:
This PR is now ready. I split it into two commits so that we can easily restore the atom synchronization code if we want to fetch another atom in the future.

EDIT: I haven't tested this yet after wlroots 0.20 release. Just a moment...

tokyo4j:
I tested with `wine notepad` and it worked fine.

Consolatis:
Thanks!

PR #2768, opened 2025-05-27
src/layers.c: send fractional scale before map

Consolatis:
Written trying to solve an issue reported with layershell clients getting blurry after system resume when using an output scale. Didn't fix that issue but still feels like the correct thing to do.


tokyo4j:
Thanks!

PR #2778, opened 2025-06-01
input: support tablet tool mouse buttons

### LINKED ISSUE #2775: Button events on Wacom mouse tool are ignored.
### Problem description

I'm using Wacom CTE-430 (Graphire3 4x5) pen tablet. With the stylus/eraser tool everything works fine. With the mouse tool, only movements are handled. Any button clicks are completely ignored. If labwc is started with `--debug` option the following messages are logged when the buttons are pressed and released:
```
[DEBUG] [../labwc-0.8.4/src/config/tablet.c:121] no button map target for 0x110
[DEBUG] [../labwc-0.8.4/src/config/tablet.c:121] no button map target for 0x111
[DEBUG] [../labwc-0.8.4/src/config/tablet.c:121] no button map target for 0x112
```
The reported codes just happen to correspond to `BTN_LEFT`, `BTN_RIGHT` and `BTN_MIDDLE`, as defined by libinput.
A quick look at the labwc code suggests that it always expects to find a button mapping for each event, even in the cases like this, when no mapping is necessary. The configuration file syntax, on the other hand, does not allow to specify mappings for these events.

### Steps to reproduce

1. Launch labwc with default configuration.
2. Try to use Wacom mouse with labwc itself or any application.

Observed result: mouse clicks have no effect.

### labwc build source

Release

### labwc version

0.8.4

### labwc environment

### Distribution

Artix Linux

jp7677:
Those are regular mouse buttons, but attached to a tablet tool (which looks like a mouse).
Not tested since I don't have such a device.

Closes #2775

### DIFF
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -119,7 +119,8 @@ adjust_for_motion_sensitivity(double motion_sensitivity, double *dx, double *dy)
 }
 
 static struct wlr_surface*
-tablet_get_coords(struct drawing_tablet *tablet, double *x, double *y, double *dx, double *dy)
+tablet_get_coords(struct drawing_tablet *tablet, enum motion motion_mode,
+		double *x, double *y, double *dx, double *dy)
 {
 	*x = tablet->x;
 	*y = tablet->y;
@@ -146,7 +147,7 @@ tablet_get_coords(struct drawing_tablet *tablet, double *x, double *y, double *d
 
 	/* initialize here to avoid a maybe-uninitialized compiler warning */
 	double lx = -1, ly = -1;
-	switch (tablet->motion_mode) {
+	switch (motion_mode) {
 	case LAB_TABLET_MOTION_ABSOLUTE:
 		wlr_cursor_absolute_to_layout_coords(tablet->seat->cursor,
 			tablet->wlr_input_device, *x, *y, &lx, &ly);
@@ -177,8 +178,8 @@ tablet_get_coords(struct drawing_tablet *tablet, double *x, double *y, double *d
 
 static void
 notify_motion(struct drawing_tablet *tablet, struct drawing_tablet_tool *tool,
-		struct wlr_surface *surface, double x, double y, double dx, double dy,
-		uint32_t time)
+		struct wlr_surface *surface, enum motion motion_mode,
+		double x, double y, double dx, double dy, uint32_t time)
 {
 	bool enter_surface = false;
 	/* Postpone proximity-in on a new surface when the tip is down */
@@ -188,7 +189,7 @@ notify_motion(struct drawing_tablet *tablet, struct drawing_tablet_tool *tool,
 			tablet->tablet_v2, surface);
 	}
 
-	switch (tablet->motion_mode) {
+	switch (motion_mode) {
 	case LAB_TABLET_MOTION_ABSOLUTE:
 		wlr_cursor_warp_absolute(tablet->seat->cursor,
 			tablet->wlr_input_device, x, y);
@@ -253,11 +254,6 @@ handle_tablet_tool_proximity(struct wl_listener *listener, void *data)
 	idle_manager_notify_activity(tablet->seat->seat);
 	cursor_set_visible(tablet->seat, /* visible */ true);
 
-	if (ev->state == WLR_TABLET_TOOL_PROXIMITY_IN) {
-		tablet->motion_mode =
-			tool_motion_mode(rc.tablet_tool.motion, ev->tool);
-	}
-
 	/*
 	 * Reset relative coordinates, we don't want to move the
 	 * cursor on proximity-in for relative positioning.
@@ -268,8 +264,9 @@ handle_tablet_tool_proximity(struct wl_listener *listener, void *data)
 	tablet->x = ev->x;
 	tablet->y = ev->y;
 
+	enum motion motion_mode = tool_motion_mode(rc.tablet_tool.motion, ev->tool);
 	double x, y, dx, dy;
-	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
+	struct wlr_surface *surface = tablet_get_coords(tablet, motion_mode, &x, &y, &dx, &dy);
 
 	/*
 	 * Do not attempt to create a tablet tool when mouse emulation is
@@ -292,7 +289,8 @@ handle_tablet_tool_proximity(struct wl_listener *listener, void *data)
 	 */
 	if (tool && surface) {
 		if (tool->tool_v2 && ev->state == WLR_TABLET_TOOL_PROXIMITY_IN) {
-			notify_motion(tablet, tool, surface, x, y, dx, dy, ev->time_msec);
+			notify_motion(tablet, tool, surface, motion_mode,
+				x, y, dx, dy, ev->time_msec);
 		}
 		if (tool->tool_v2 && ev->state == WLR_TABLET_TOOL_PROXIMITY_OUT) {
 			wlr_tablet_v2_tablet_tool_notify_proximity_out(tool->tool_v2);
@@ -355,8 +353,9 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 		tablet->wheel_delta = ev->wheel_delta;
 	}
 
+	enum motion motion_mode = tool_motion_mode(rc.tablet_tool.motion, ev->tool);
 	double x, y, dx, dy;
-	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
+	struct wlr_surface *surface = tablet_get_coords(tablet, motion_mode, &x, &y, &dx, &dy);
 
 	/*
 	 * We are sending tablet notifications on the following conditions:
@@ -373,7 +372,7 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 			&& tablet->seat->server->input_mode == LAB_INPUT_STATE_PASSTHROUGH)
 			|| wlr_tablet_tool_v2_has_implicit_grab(tool->tool_v2))) {
 		/* motion seems to be supported by all tools */
-		notify_motion(tablet, tool, surface, x, y, dx, dy, ev->time_msec);
+		notify_motion(tablet, tool, surface, motion_mode, x, y, dx, dy, ev->time_msec);
 
 		/* notify about other axis based on tool capabilities */
 		if (ev->tool->distance) {
@@ -422,7 +421,7 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 					tool->tool_v2);
 			}
 
-			switch (tablet->motion_mode) {
+			switch (motion_mode) {
 			case LAB_TABLET_MOTION_ABSOLUTE:
 				cursor_emulate_move_absolute(tablet->seat,
 					&ev->tablet->base,
@@ -474,8 +473,9 @@ handle_tablet_tool_tip(struct wl_listener *listener, void *data)
 	idle_manager_notify_activity(tablet->seat->seat);
 	cursor_set_visible(tablet->seat, /* visible */ true);
 
+	enum motion motion_mode = tool_motion_mode(rc.tablet_tool.motion, ev->tool);
 	double x, y, dx, dy;
-	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
+	struct wlr_surface *surface = tablet_get_coords(tablet, motion_mode, &x, &y, &dx, &dy);
 
 	uint32_t button = tablet_get_mapped_button(BTN_TOOL_PEN);
 
@@ -554,8 +554,9 @@ handle_tablet_tool_button(struct wl_listener *listener, void *data)
 	idle_manager_notify_activity(tablet->seat->seat);
 	cursor_set_visible(tablet->seat, /* visible */ true);
 
+	enum motion motion_mode = tool_motion_mode(rc.tablet_tool.motion, ev->tool);
 	double x, y, dx, dy;
-	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
+	struct wlr_surface *surface = tablet_get_coords(tablet, motion_mode, &x, &y, &dx, &dy);
 
 	uint32_t button = tablet_get_mapped_button(ev->button);
 
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -273,8 +273,13 @@ handle_tablet_tool_proximity(struct wl_listener *listener, void *data)
 	 * enforced. Not having a tool or tablet capable surface will trigger
 	 * the fallback to cursor move/button emulation in the tablet signal
 	 * handlers.
+	 * Also stick to mouse emulation when the current tool is a tablet mouse.
+	 * Client support for tablet mouses in tablet mode is often incomplete
+	 * and no functionality is lost since those device do not support tool
+	 * specific axis like pressure or distance.
 	 */
 	if (!rc.tablet.force_mouse_emulation
+			&& ev->tool->type != WLR_TABLET_TOOL_TYPE_MOUSE
 			&& tablet->seat->server->tablet_manager && !tool) {
 		/*
 		 * Unfortunately `wlr_tool` is only present in the events, so
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -439,6 +439,21 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 				break;
 			}
 		}
+		if (ev->updated_axes & WLR_TABLET_TOOL_AXIS_WHEEL) {
+			/*
+			 * libinput reports delta_discrete for tablet tool mouses,
+			 * but unfortunately wlroots doesn't expose it. That said,
+			 * based on the libinput source (tablet_device_has_axis()),
+			 * we only have to deal with non-high-res mouses here, so
+			 * it is relatively safe to set a fixed value.
+			 */
+			int delta_discrete = tablet->wheel_delta >= 0 ? 1 : -1;
+			cursor_emulate_axis(tablet->seat, &ev->tablet->base,
+				WL_POINTER_AXIS_VERTICAL_SCROLL,
+				tablet->wheel_delta,
+				WLR_POINTER_AXIS_DISCRETE_STEP * delta_discrete,
+				WL_POINTER_AXIS_SOURCE_WHEEL, ev->time_msec);
+		}
 	}
 }
 
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -417,7 +417,7 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 		}
 	} else {
 		if (ev->updated_axes & (WLR_TABLET_TOOL_AXIS_X | WLR_TABLET_TOOL_AXIS_Y)) {
-			if (tool && tool->tool_v2->focused_surface) {
+			if (tool && tool->tool_v2 && tool->tool_v2->focused_surface) {
 				wlr_tablet_v2_tablet_tool_notify_proximity_out(
 					tool->tool_v2);
 			}
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -895,13 +895,15 @@ extending outward from the snapped edge.
 	available mouse	buttons. If not specified otherwise, the tip is
 	mapped to left mouse click, the first pen button (Stylus) is mapped
 	to right mouse button click and the second pen button (Stylus2)
-	emulates a middle mouse	button click.
+	emulates a middle mouse	button click. Buttons of a tablet tool mouse
+	are by default mapped to their (regular) mouse counterparts.
 
 	Supported map *buttons* for mouse emulation are:
 	- Tip
 	- Stylus
 	- Stylus2
 	- Stylus3
+	- Left..Task
 	- Pad
 	- Pad2..Pad9
 
--- a/include/input/tablet.h
+++ b/include/input/tablet.h
@@ -14,7 +14,6 @@ struct drawing_tablet {
 	struct seat *seat;
 	struct wlr_tablet *tablet;
 	struct wlr_tablet_v2_tablet *tablet_v2;
-	enum motion motion_mode;
 	double x, y, dx, dy;
 	double distance;
 	double pressure;
--- a/src/config/tablet.c
+++ b/src/config/tablet.c
@@ -103,7 +103,10 @@ void
 tablet_load_default_button_mappings(void)
 {
 	rc.tablet.button_map_count = 0;
-	tablet_button_mapping_add(BTN_TOOL_PEN, BTN_LEFT); /* Used for the pen tip */
+
+	/* used for the pen tip */
+	tablet_button_mapping_add(BTN_TOOL_PEN, BTN_LEFT);
+
 	tablet_button_mapping_add(BTN_STYLUS, BTN_RIGHT);
 	tablet_button_mapping_add(BTN_STYLUS2, BTN_MIDDLE);
 }
--- a/src/config/tablet.c
+++ b/src/config/tablet.c
@@ -51,6 +51,22 @@ tablet_button_from_str(const char *button)
 		return BTN_STYLUS2;
 	} else if (!strcasecmp(button, "Stylus3")) {
 		return BTN_STYLUS3;
+	} else if (!strcasecmp(button, "Left")) {
+		return BTN_LEFT;
+	} else if (!strcasecmp(button, "Right")) {
+		return BTN_RIGHT;
+	} else if (!strcasecmp(button, "Middle")) {
+		return BTN_MIDDLE;
+	} else if (!strcasecmp(button, "Side")) {
+		return BTN_SIDE;
+	} else if (!strcasecmp(button, "Extra")) {
+		return BTN_EXTRA;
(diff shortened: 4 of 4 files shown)


Consolatis:
Yes, this feels cleaner.

LGTM, the CI error for Arch is unrelated.

tokyo4j:
LTGM, but maybe we should also fall back to the same button code as its source like this? I guess this prevents potential bugs like #2775 when we encounter other button codes.

```diff
diff --git a/src/config/tablet.c b/src/config/tablet.c
index ef8059d4..790e83a1 100644
--- a/src/config/tablet.c
+++ b/src/config/tablet.c
@@ -119,5 +119,5 @@ tablet_get_mapped_button(uint32_t src_button)
 		}
 	}
 	wlr_log(WLR_DEBUG, "no button map target for 0x%x", src_button);
-	return 0;
+	return src_button;
 }
```

jp7677:
@sfzhi thanks for your detailed report in https://github.com/labwc/labwc/issues/2775 . Let's continue here:

> The scroll wheel on the Wacom mouse tool does not work at all (it never did, I just didn't pay attention to that earlier). No messages are logged by labwc when trying to use it.

Yes, we indeed don't simulate cursor axis on a tablet tool wheel axis changes for mouse emulation. We could try to do that. That said, I'm not event sure if the wheel axis is the correct one. Keep in mind though that applications that support the tablet protocol (e.g. GTK3/4), needs to handle this by themself. I don't know if they do.

> The left button works in labwc, alacritty and foot, does not work in the rest. When it doesn't work, labwc logs:
[INFO] [../src/input/tablet.c:595] invalid stylus button mapping for tablet mode. Coincidentally or not, the applications where it doesn't work are all GTK-based.

My assumption is, that from your list only GTK applications implement the tablet protocol. All others I guess do not and in that case labwc switches automatically back to mouse emulation (which is the same as `mouseEmulation="yes"`)

I've added a commit which simulates a tool tip down/up when a left mouse button code came in (or is the result of the  mapping). Does this fixes the left mouse button on your tablet mouse for GTK applications when mouse emulation is off?

> There is some strange behavior with menus.

I have also seen strange menu behavior in tablet mode. My guess is that those are GTK issues but since not that many people use a tablet and report those issues, it isn't high on the list. Most recent GTK3/4 version do contain some fixes in that regard as far as I remember, e.g. https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8240 .  A good way to verify this, is to try to reproduce in recent Gnome. If Gnome is fine, it could be indeed our issue.

src/input/tablet.c
```diff
@@ -581,16 +581,31 @@ handle_tablet_tool_button(struct wl_listener *listener, void *data)
...
+						? ZWP_TABLET_PAD_V2_BUTTON_STATE_PRESSED
+						: ZWP_TABLET_PAD_V2_BUTTON_STATE_RELEASED);
+			} else {
+				/*
+				 * Simulate a tool tip down/up, reason for that
+				 * could be an uncommon mapping or a left button
+				 * click of a tablet tool mouse.
+				 */
+				if (ev->state == WLR_BUTTON_PRESSED) {
+					wlr_tablet_v2_tablet_tool_notify_down(tool->tool_v2);
```
jp7677: GTK4 only handles BTN_STYLUS{2|3} in its button handler (https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/wayland/gdkseat-wayland.c?ref_type=heads#L3088). So unfortunately we cannot just pass `BTN_{LEFT|RIGHT|MIDDLE}` from the tablet tool mouse into `wlr_tablet_v2_tablet_tool_notify_button`. One could argue this to be a GTK issue, because we kind of have to map the tool mouse buttons back to have basic functionality in tablet mode. 

A completely alternative (and I guess even simpler and better working) approach would be to always use mouse emulation when the tablet tool is of type `WLR_TABLET_TOOL_TYPE_MOUSE`. We would loose any specific functionalities by clients, but I wonder if there any for tablet tool mouses.

Like this instead of the last commit:

<p>

```diff
diff --git a/src/input/tablet.c b/src/input/tablet.c
index 7d27dde2..9564961e 100644
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -357,11 +357,15 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 
        double x, y, dx, dy;
        struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
+       bool is_tablet_tool_mouse = tool
+               && tool->tool_v2->wlr_tool->type == WLR_TABLET_TOOL_TYPE_MOUSE;
+
 
        /*
         * We are sending tablet notifications on the following conditions:
         * - a tablet tool (aka pen/stylus) had been created earlier on
         *   proximity-in
+        * - the tablet tool is no tablet tool mouse
         * - there is no current tip or button press (e.g. from out-of-surface
         *   scrolling) that started on a non tablet capable surface
         * - the surface below the tip understands the tablet protocol and is in
@@ -369,7 +373,7 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
         *   don't have a tablet-capable surface but are still having an active
         *   grab (e.g. from out-of-surface scrolling).
         */
-       if (tool && !is_down_mouse_emulation && ((surface
+       if (tool && !is_tablet_tool_mouse && !is_down_mouse_emulation && ((surface
                        && tablet->seat->server->input_mode == LAB_INPUT_STATE_PASSTHROUGH)
                        || wlr_tablet_tool_v2_has_implicit_grab(tool->tool_v2))) {
                /* motion seems to be supported by all tools */
@@ -558,16 +562,19 @@ handle_tablet_tool_button(struct wl_listener *listener, void *data)
        struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
 
        uint32_t button = tablet_get_mapped_button(ev->button);
+       bool is_tablet_tool_mouse = tool
+               && tool->tool_v2->wlr_tool->type == WLR_TABLET_TOOL_TYPE_MOUSE;
 
        /*
         * We are sending tablet notifications on the following conditions:
         * - a tablet tool (aka pen/stylus) had been created earlier on
         *   proximity-in
+        * - the tablet tool is no tablet tool mouse
         * - there is no current tip or button press (e.g. out of surface
         *   scrolling) that started on a non tablet capable surface
         * - the surface below the tip understands the tablet protocol.
         */
-       if (tool && !is_down_mouse_emulation && surface) {
+       if (tool && !is_tablet_tool_mouse && !is_down_mouse_emulation && surface) {
                if (button && ev->state == WLR_BUTTON_PRESSED) {
                        struct view *view = view_from_wlr_surface(surface);
                        struct mousebind *mousebind;
```
</p>
</details>

sfzhi: I've tested the updated branch and the alternative patch. Both work well and solve the problem with the left button. The patch also has a bonus that it avoids the weird issue with menus in GTK (but only for the mouse tool, of course).

I don't know much about Wayland protocols, so I'can't tell which solution is more technically correct. From general perspective I would choose the patch (always use mouse emulation). One reason is that it feels conceptually cleaner. The mouse tool is just a mouse, which happens to use the tablet as a communication channel. It is different from a regular mouse in that it supports absolute positioning, but that fact is already hidden from applications, so there is nothing left to make it special. It's not unlike having multiple regular mice connected at the same time: applications cannot tell them apart, but nobody seems to be complaining about that. The other reason is that, if I understand correctly, with "permanent" mouse emulation there will be no need to be concerned that applications might not understand scroll wheel events.

jp7677: Thanks a lot for the feedback on both versions. Having slept a night, my opinion also shifts towards always using mouse emulation for a tablet tool mouse. The solution fully in line with the protocol would leave you with non-working buttons in (at least in current GTK4 since they don't handle the buttons reported from the tool mouse) and, based on your testing, without scroll wheel support. May be this changes in future. But for now, let's go practical, even this is not what the tablet protocol had in mind, and go for a solution that works and also allows for scroll emulation in the future.

I'll adjust this PR in later today.

jp7677:
> LTGM, but maybe we should also fall back to the same button code as its source like this? I guess this prevents potential bugs like #2775 when we encounter other button codes.

I'm a bit torn about adding the fallback. My argument would be that hitting the fallback would hide the fact that the mapping is incomplete or that such a fallback button cannot be mapped at all since we forgot to add that button code to the mapping entries. No strong preference from my side though, so if you still prefer the fallback, I'm happy to add that.

src/input/tablet.c
jp7677: I've reworked the last commit, so all ready from my side.

jp7677: PS: Does that snippet on top of this PR effects your scroll wheel?

<p>

```diff
diff --git a/src/input/tablet.c b/src/input/tablet.c
index 8c46376d..37845587 100644
--- a/src/input/tablet.c
+++ b/src/input/tablet.c
@@ -438,6 +438,13 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
                                break;
                        }
                }
+               if (ev->updated_axes & WLR_TABLET_TOOL_AXIS_WHEEL) {
+                       cursor_emulate_axis(tablet->seat,
+                               &ev->tablet->base,
+                               WL_POINTER_AXIS_VERTICAL_SCROLL,
+                               tablet->wheel_delta, 0,
+                               WL_POINTER_AXIS_SOURCE_CONTINUOUS, ev->time_msec);
+               }
        }
 }
``` 

</p>
</details>

sfzhi: With this last patch the scroll wheel kind of works, but not quite.

- It works fine in terminal emulators and in Gedit (regular scrolling and ctrl+scroll to change the font size). I don't see any difference in behavior compared to a regular mouse.
- In GIMP some thing work, some don't. Regular scrolling and zoom-in/zoom-out with ctrl+scroll work fine. Changing some numeric input fields by hover+scroll does not work (no reaction at all).
- Strange things happen in awf. Switching tabs by hover+scroll works in GTK4, but not in GTK3. But it does work in GTK3 via XWayland, although in this case it skips tabs sometimes (one scroll step moves tab selection by two instead of one).
- Another strange thing happens with sliders (when using hover+scroll). They kind of work, but the size of the scroll step differs from the one with a regular mouse. And that difference is not always the same. For example, in awf-GTK4 it takes about 6.5 scroll steps to cover the whole slider range, while with a regular mouse it takes 10 steps. In afw-GTK3 there is some difference too, but it's barely noticeable. The things get really crazy in pavucontrol. With a regular mouse, one scroll step changes the volume by 5%. With the Wacom mouse, it takes 44 scroll steps to change the volume by 1%.

jp7677: Thanks, that is already much better than I had expected. I can do a follow up PR after this PR has landed since I guess we need some rounds until we have a good working solution.

jp7677:
@sfzhi ~I've enforced relative motion for a mouse tool, does this makes sense?~ Do you actually have relative or absolute motion with your mouse? This should already be in place, but I could never test this, so curious if this works as intended.

I've also added the scroll wheel emulation with a slight adjustment, does this works somewhat better now? Note the comment about `delta_discrete` unfortunately not being available (https://wayland.freedesktop.org/libinput/doc/latest/api/group__event__tablet.html#gac983a446dab6d1d6835273f2b6110595 / https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/libinput/tablet_tool.c#L179).

Edit: Reverted the commit about relative motion, that should already be in place. Also CI failure looks like a hickup.

src/input/tablet.c
```diff
@@ -357,19 +357,22 @@ handle_tablet_tool_axis(struct wl_listener *listener, void *data)
 
 	double x, y, dx, dy;
 	struct wlr_surface *surface = tablet_get_coords(tablet, &x, &y, &dx, &dy);
+	bool is_tablet_tool_mouse = tool && tool->tool_v2
```
jp7677: Note to self, may be we can do this more elegantly by manually destroying the tool in proximity-in.

sfzhi: Indeed, I have relative mouse motion (which is how I like it). Although, I saw absolute motion once or twice in the past. It was with the last release, before this PR. But now I can't reproduce that (not even when going back to that version). It must have been a glitch of some sort. I saw in the code that the choice between absolute and relative was made upon receiving proximity-in event. My theory at the time was that if labwc was started when the mouse was already on the tablet then it would never receive that event. Now that theory seems incorrect, because I can see "setting up tablet tool" logged at startup if the mouse is on the tablet. It must be triggered by the proximity-in event, if I understand correctly.

Regarding the scroll wheel, I don't see any change in behavior compared to my previous tests.

sfzhi:
> Do you actually have relative or absolute motion with your mouse?

Indeed, I have relative mouse motion (which is how I like it). Although, I saw absolute motion once or twice in the past. It was with the last release, before this PR. But now I can't reproduce that (not even when going back to that version). It must have been a glitch of some sort. I saw in the code that the choice between absolute and relative was made upon receiving proximity-in event. My theory at the time was that if labwc was started when the mouse was already on the tablet then it would never receive that event. Now that theory seems incorrect, because I can see "setting up tablet tool" logged at startup if the mouse is on the tablet. It must be triggered by the proximity-in event, if I understand correctly.

> I've also added the scroll wheel emulation with a slight adjustment, does this works somewhat better now?

I don't see any change in behavior compared to my previous tests.

jp7677:
> Indeed, I have relative mouse motion (which is how I like it). Although, I saw absolute motion once or twice in the past. 

Thanks, I've reworked that part slightly to ensure that the motion mode is always up to date for the currently used tool. 

> I don't see any change in behavior compared to my previous tests.

I've added a cheap (and not tested) way to calculate the wheel discrete number based on a fixed click angle. Do you see a difference?  

(Also converted this PR to draft while playing with the wheel support)

sfzhi:
The behavior of the scroll wheel in pavucontrol has slightly improved, but is still far from normal. It now takes 24 scroll steps to change the volume by 1% (was 44 previously). In awf-GTK4 scrolling has become much slower and feels about the same as in pavucontrol (which also uses GTK4). In other applications, including awf-GTK3, scrolling has become extremely slow, to the point that at first I thought it didn't work at all. It takes quite some amount of scrolling before the application reacts as it would normally react to a single scroll step.

sfzhi:
I have also conducted an experiment with patching wlroots to report the actual delta_discrete. I added a field to `struct wlr_tablet_tool_axis_event` and set it to the value returned by `libinput_event_tablet_tool_get_wheel_delta_discrete()`. That seemed like the obvious solution, but I really don't know if it was the right way to do it. Anyway, it turned out the value from libinput was always 1 or -1, depending on the scroll direction. That matched the calculated value exactly (`tablet->wheel_delta` was 15.0 or -15.0).

jp7677:
> I have also conducted an experiment with patching wlroots to report the actual delta_discrete. I added a field to `struct wlr_tablet_tool_axis_event` and set it to the value returned by `libinput_event_tablet_tool_get_wheel_delta_discrete()`. That seemed like the obvious solution, but I really don't know if it was the right way to do it. Anyway, it turned out the value from libinput was always 1 or -1, depending on the scroll direction. That matched the calculated value exactly (`tablet->wheel_delta` was 15.0 or -15.0).

Cool, thanks for trying and confirming that (at least) my math hasn't left me ;)

Could you run `wev` and look there how an axis event is reported from your regular mouse and from the tablet mouse? Here is to hoping that there is a difference that reveals a hint.

sfzhi:
My next experiment was to multiply the `delta_discrere` value by 120 before passing it to `cursor_emulate_axis()`. The factor was guessed based on the scroll behavior in pavucontrol. With that change everything magically started to work. Now I don't see any difference in scroll behavior between the Wacom mouse and a regular mouse. Pure magic :).

sfzhi:
Events from a regular mouse are reported by wev like this:
```
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 265412879; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
```
From the Wacom mouse I can see only a series of `frame` and nothing in between. Once in a while this shows up:
```
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 265599034; axis: 0 (vertical), value: 1800.000000
[13:      wl_pointer] frame
```
I can't help noticing that the "value" differs by the factor of exactly 120. If add this factor to the code, as I mentioned above, the events from both mice look identical in wev.

Consolatis:
There are two different scroll events from the kernel, the older `REL_WHEEL` and the newer `REL_WHEEL_HI_RES`. The newer one reports values multiplied by 120. So it looks like what is being used by the tool mouse is the ~~HI_RES~~ older variant.

Can you run `evtest` and compare the two mice?

sfzhi:
The regular mouse:
```
Input driver version is 1.0.1
Input device ID: bus 0x3 vendor 0x46d product 0xc077 version 0x111
Input device name: "Logitech USB Optical Mouse"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 272 (BTN_LEFT)
    Event code 273 (BTN_RIGHT)
    Event code 274 (BTN_MIDDLE)
  Event type 2 (EV_REL)
    Event code 0 (REL_X)
    Event code 1 (REL_Y)
    Event code 8 (REL_WHEEL)
    Event code 11 (REL_WHEEL_HI_RES)
  Event type 4 (EV_MSC)
    Event code 4 (MSC_SCAN)
Properties:
Testing ... (interrupt to exit)
Event: time 1749212699.817162, type 2 (EV_REL), code 8 (REL_WHEEL), value -1
Event: time 1749212699.817162, type 2 (EV_REL), code 11 (REL_WHEEL_HI_RES), value -120
Event: time 1749212699.817162, -------------- SYN_REPORT ------------
Event: time 1749212700.897109, type 2 (EV_REL), code 8 (REL_WHEEL), value 1
Event: time 1749212700.897109, type 2 (EV_REL), code 11 (REL_WHEEL_HI_RES), value 120
Event: time 1749212700.897109, -------------- SYN_REPORT ------------
```
The Wacom mouse:
```
Input driver version is 1.0.1
Input device ID: bus 0x3 vendor 0x56a product 0x13 version 0x100
Input device name: "Wacom Graphire3 Pen"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 272 (BTN_LEFT)
    Event code 273 (BTN_RIGHT)
    Event code 274 (BTN_MIDDLE)
    Event code 320 (BTN_TOOL_PEN)
    Event code 321 (BTN_TOOL_RUBBER)
    Event code 326 (BTN_TOOL_MOUSE)
    Event code 330 (BTN_TOUCH)
    Event code 331 (BTN_STYLUS)
    Event code 332 (BTN_STYLUS2)
  Event type 2 (EV_REL)
    Event code 8 (REL_WHEEL)
  Event type 3 (EV_ABS)
    Event code 0 (ABS_X)
      Value   5360
      Min        0
      Max    10208
      Fuzz       4
      Resolution      80
    Event code 1 (ABS_Y)
      Value   3457
      Min        0
      Max     7424
      Fuzz       4
      Resolution      80
    Event code 24 (ABS_PRESSURE)
      Value      0
      Min        0
      Max      511
    Event code 40 (ABS_MISC)
      Value      6
      Min        0
      Max        0
Properties:
  Property type 0 (INPUT_PROP_POINTER)
Testing ... (interrupt to exit)
Event: time 1749212818.274912, type 2 (EV_REL), code 8 (REL_WHEEL), value -1
Event: time 1749212818.274912, -------------- SYN_REPORT ------------
Event: time 1749212818.898728, type 2 (EV_REL), code 8 (REL_WHEEL), value 1
Event: time 1749212818.898728, -------------- SYN_REPORT ------------
```
I guess this explains the mystery.

Consolatis:
Alright, then the question becomes if we can do the scroll wheel emulation in such a way that both cases work correctly. So tool mice which only support `REL_WHEEL` as well as tool mice which only support `REL_WHEEL_HI_RES`. I am not sure how wlroots handles that internally but it looks like we need to convert everything so it always ends up with either the new or the old format before processing it further in `cursor.c`.

sfzhi:
I've spent some time browsing through the forest of previously unfamiliar code and got the impression that there may not be any other use cases that need to be supported. I can see that for a regulat mouse libinput has two different APIs for `REL_WHEEL` and `REL_WHEEL_HI_RES` events (they call the latter "v120"), and wlroots uses both. But there is nothing similar for tablet devices. It looks like only `REL_WHEEL` is supported: https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/src/evdev-tablet.c?ref_type=heads#L958 . Out of curiosity I checked how it's done in xf86-input-wacom driver for Xorg. They don't seem to support hi-res events either: https://github.com/linuxwacom/xf86-input-wacom/blob/master/src/wcmUSB.c#L1946 . Unless I misunderstood something, this suggests that may be tablets with hi-res mouse tools simply do not exist (or if they do, they also emit corresponding lo-res events). In other words, with the current implementation in libinput and wlroots, it doesn't look like labwc will ever be receiving hi-res scroll events for a mouse tool, even if those are emitted by the hardware. If that's the case, then we could just assume that any scroll event for a mouse tool originates from `REL_WHEEL` and stupidly multiply the reported value by 120.

jp7677:
Thanks @Consolatis for joining! Based on https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/src/evdev-tablet.c?ref_type=heads#L148 I would say a tablet tool mouse is always a non-high-res mouse, so I think it is safe to always set `delta_discrete` to `WLR_POINTER_AXIS_DISCRETE_STEP` or `WLR_POINTER_AXIS_DISCRETE_STEP * -1`.

I don't yet fully understand the magic in https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/seat/wlr_seat_pointer.c#L288 and why `tablet->wheel_delta` can stay as is but I think we are quite close...

jp7677:
@sfzhi Seems like we both reached the same conclusion independently.

jp7677:
@sfzhi Just to be sure that I didn't messed up, last version of this PR fully works for you for both tablet tools?

sfzhi:
Everything works fine, except for the menus in GTK when using the stylus. The same weird things still happen. I assume it is an unrelated issue. By the way, it acts exactly the same with sway. So it's probably an issue in wlroots or GTK.

sfzhi:
Besides, I strongly doubt that the menu weirdness is specific to my hardware. If you have any pen tablet you may be able to reproduce at least some of it. For example, you could try the following scenario:
- Start GIMP
- Create a new image
- Use the stylus on the image: click the equivalent of the right mouse button - popup menu opens.
- **Without releasing the button** move the stylus slightly right and down - to select the first menu item ("File").

The "File" submenu is supposed to open next to its parent item (in the popup menu). Instead it opens in the top left corner next to the "File" item in the main menu. I can't imagine how labwc could possibly cause such behavior, even if it wanted to.

jp7677:
> Everything works fine, except for the menus in GTK when using the stylus. The same weird things still happen. I assume it is an unrelated issue. By the way, it acts exactly the same with sway. So it's probably an issue in wlroots or GTK.

Yeah, as stated, that is a GTK3 issue. GTK3 3.24.49 should improve, if not fix, this behavior. At least I saw that the menu in `gnome-terminal` was fixed after that GTK3 version landed.

sfzhi:
> GTK3 3.24.49 should improve, if not fix, this behavior.

I'm already using 3.24.49. I guess they didn't fix it completely.

Consolatis:
I am not too deep in the tablet and input emulation part of labwc but the code LGTM and I didn't spot anything obvious in the diff. I also can't test it because I don't have a mouse tool but based on the tests by @sfzhi it seems to work fine.

Thanks @sfzhi for testing.
And obviously thanks @jp7677 for your relentless work to improve the tablet experience in labwc!

I am not merging this right away to give @tokyo4j a chance to raise any objections to https://github.com/labwc/labwc/pull/2778#issuecomment-2936359790.

As for the assumed GTK bug, maybe this is somewhat related to #2787.

tokyo4j:
LGTM. Thanks for working on this.

I'm still not sure about always falling back to the original button, but it can be discussed separately.

tokyo4j:
@jp7677 Let me ask you a question about the current tablet implementation.

While reading `input/tablet.c`, I was puzzled to see that tablet buttons are mapped to the mouse button and then reverse-mapped again:

https://github.com/labwc/labwc/blob/33fc1c6d0543f577dafbf4587fecdbc9e21bbced/src/input/tablet.c#L581-L614

Why did you do this? Sorry if it was already discussed.

jp7677:
> While reading `input/tablet.c`, I was puzzled to see that tablet buttons are mapped to the mouse button and then reverse-mapped again ...  Why did you do this?

There were/are several reasons for that: 

- History because initially there was only mouse emulation, so the button mapping was already there and the actual tablet protocol implementation was added later. 
- It turned out quite convenient for the button mapping configuration. That way we could keep the mapping configuration clean and always just map to mouse buttons and don't had to introduce separate mappings for mouse emulation and for tablet mode which would be quite confusing. 
- The most important point: it also allowed me to reverse map in a way that it matches GTK so that the pen buttons are the same for mouse emulation and tablet mode. Note that, contrary to the default mappings, the reverse mapping maps BTN_RIGHT to STYLUS2 and BTN_MIDDLE to STYLUS. Without this, the second and third stylus button would be reversed between tablet and mouse emulation which is really annoying. That used to be the way in Gnome/Mutter between native and XWayland until it got fixed, really annoying.     

Fun fact: When in tablet mode, GTK maps the stylus buttons again back to mouse buttons, so it's indeed back and forth several times.

PR #2783, opened 2025-06-02
labwc-config(5): describe environment file comment syntax

johanmalm:


docs/labwc-config.5.scd
```diff
@@ -56,6 +56,9 @@ re-loaded on --reconfigure.
 Any environment variables referenced as $VARIABLE or ${VARIABLE} will be
 substituted and the tilde (~) will be expanded as the user's home directory.
 
+Any line beginning with the character '#' in an environment file will be treated
+as a comment. But trailing comments are not supported.
```
Consolatis: The word `But` feels a bit weird to me here.

Consolatis (commented):
Minor nitpick below, English is not my native langue though so take it with a grain of salt.

Even if kept as is: LGTM

docs/labwc-config.5.scd
johanmalm: Yes, good point. Not great starting a sentence with "But"

Issue #2774, opened 2025-06-01
mpv when opened in full screen start in a small window


muezabdalla:
### Problem description

I have mpv configured to run in full screen but when it starts it only start in a small window without decoration 
but the bar disappear like if mpv was in full screen
here is a screenshot:

![Image](https://github.com/user-attachments/assets/4efacfbe-54b1-45c5-ae4b-312ff7021d33)

 and I could not replicate it in wayfire or icewm

### Steps to reproduce

make mpv open in fullscreen by defualt
open any video with mpv

### labwc build source

Local build

### labwc version

from git and 0.8.4

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

void linux

tokyo4j:
Mpv seems to be sending `xdg_toplevel.set_fullscreen()` and `xdg_surface.set_window_geometry()` with a windowed geometry. As a result, labwc mis-recognizes that mpv refused to set the window geometry offered by labwc and updates the scheduled geometry to the windowed geometry. 

![Image](https://github.com/user-attachments/assets/f5f63582-c685-47f3-bce1-1e0f3bdc6272)

Removing the following code will fix this, but I don't know what's the best solution at the moment.

https://github.com/labwc/labwc/blob/949d769198bd9ee1884fdccc673740ab11e0c678/src/xdg.c#L246-L247

tokyo4j:
Related: #2327

Consolatis:
Just a slightly related note: The reason the smaller surface is shown on the top-left and the other windows being visible is due to labwc being [non-spec compliant](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L1078): the surface should be centered on the fullscreen output in case the dimensions are smaller than the full output geometry. We should also add a solid background behind it, covering the full output. Technically we are also supposed to put a solid background behind any fullscreen surface if the surface is not completely opaque (which I am absolutely not a fan of). Ref: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/116

danielfikko:
[...]

PR #2788, opened 2025-06-04
xdg: add missing tracking of configure serials

jlindgren90:
In some cases, we would schedule a configure event but not record that we had done so, leaving `view->pending_configure_serial` incorrect. This caused incorrect handling of in-flight commits that had not yet taken the new configure into account, such as assuming that the client had ignored the configured size and incorrectly resetting `view->pending`.

Fixes: #2774

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -120,6 +120,9 @@ do_late_positioning(struct view *view)
 	}
 }
 
+/* TODO: reorder so this forward declaration isn't needed */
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM. Not tested.

tokyo4j:
I really like this. Feels much cleaner than sway's approach.

I confirmed this fixes #2774.

Consolatis:
Yep, LGTM as well, not tested either.

Issue #1797, opened 2024-05-12
Shadows for snapped windows


Vladimir-csp:
Snapped windows loose shadows, please make them stay.

stefonarch:
If yes only as an option... kwin_wayland has it and it looks very  bad IMHO.

tsujan:
They had shadow at first, but then it was decided that they shouldn't.

I almost agree with @stefonarch, although IMO it wouldn't look so bad. However, the lack of shadow serves as a sign of the snapped state (in contrast to a window with the same size and at the same position).

Consolatis:
I really can't say anything about this as the whole concept of shadows feels strange to me personally as a user.

tsujan:
> the whole concept of shadows feels strange to me personally as a user.

Can I ask why? To me, shadows are not only nice but useful too: they're good for distinguishing the boundaries of overlapped, borderless windows (`border.Width: 0`).

Consolatis:
> Can I ask why? To me, shadows are not only nice but useful too: they're good for distinguishing the boundaries of overlapped, borderless windows (`border.Width: 0`).

I can see the point of using shadows instead of borders. However personally I prefer borders.

Vladimir-csp:
Indication for a snapped window (to edge) could be a missing border in corresponding direction (or lack of rounded corners on titlebar in case of top edge and shown titlebar).

BTW, maximized horizontal/vertical states could also drop corresponding borders.

stefonarch:
>Indication for a snapped window (to edge) could be a missing border in corresponding direction (or lack of rounded corners on titlebar in case of top edge and shown titlebar).

I'm using gaps which would need an exception.

johanmalm:
Let's close this for now. 

When someone has time/means/ability to contribute to the drop-shadow implementation I suggest we take a broader view of scope and config options.

PR #2789, opened 2025-06-05
ssd: add configurable drop-shadows for tiled windows

diredocks:
This patch introduces optional drop shadows for tiled windows. The feature is controlled via configuration, allowing users to enable or disable shadows on a per-preference basis.

related issue https://github.com/labwc/labwc/issues/1797

### DIFF
--- a/src/ssd/ssd-shadow.c
+++ b/src/ssd/ssd-shadow.c
@@ -293,15 +293,15 @@ ssd_shadow_update(struct ssd *ssd)
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
 	bool tiled_shadows = false;
 	if (rc.shadows_on_tiled) {
-		if (rc.gap >= theme->window[THEME_ACTIVE].shadow_size &&
-			rc.gap >= theme->window[THEME_INACTIVE].shadow_size) {
+		if (rc.gap >= theme->window[THEME_ACTIVE].shadow_size
+				&& rc.gap >= theme->window[THEME_INACTIVE].shadow_size) {
 			tiled_shadows = true;
 		} else {
 			wlr_log(WLR_INFO, "gap size < shadow_size, ignore rc.shadows_ontiled");
 		}
 	};
-	bool show_shadows = rc.shadows_enabled && !maximized &&
-		(!view_is_tiled(ssd->view) || tiled_shadows);
+	bool show_shadows = rc.shadows_enabled && !maximized
+		&& (!view_is_tiled(ssd->view) || tiled_shadows);
 	wlr_scene_node_set_enabled(&ssd->shadow.tree->node, show_shadows);
 	if (show_shadows) {
 		set_shadow_geometry(ssd);
--- a/src/ssd/ssd-shadow.c
+++ b/src/ssd/ssd-shadow.c
@@ -291,7 +291,8 @@ ssd_shadow_update(struct ssd *ssd)
 	struct view *view = ssd->view;
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
 	bool show_shadows =
-		rc.shadows_enabled && !maximized && !view_is_tiled(ssd->view);
+		rc.shadows_enabled && !maximized &&
+		(!view_is_tiled(ssd->view) || rc.shadows_on_tiled);
 	wlr_scene_node_set_enabled(&ssd->shadow.tree->node, show_shadows);
 	if (show_shadows) {
 		set_shadow_geometry(ssd);
--- a/src/ssd/ssd-shadow.c
+++ b/src/ssd/ssd-shadow.c
@@ -289,10 +289,19 @@ ssd_shadow_update(struct ssd *ssd)
 	assert(ssd->shadow.tree);
 
 	struct view *view = ssd->view;
+	struct theme *theme = ssd->view->server->theme;
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
-	bool show_shadows =
-		rc.shadows_enabled && !maximized &&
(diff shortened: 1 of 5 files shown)


cillian64:
I haven't tested but the implementation looks reasonable to me.  rpi-os will keep this option turned off but I don't object to the option - it makes sense if the tiled/snapped windows have decent clearance gaps.

Consolatis:
Thanks for the PR.

IMHO we need to ensure that shadows won't leak into a neighboring monitor when tiled to an output edge (either via regions or SnapToEdge). Same story with some bar in exclusive mode. I am not quite sure how to do that properly without unreasonably complicating everything. See for example `ssd_extents_update()` in `src/ssd/ssd-extents.c` for the calculation of the sizes of the invisible resize area which faces the same challenge.

diredocks:
> Thanks for the PR.
> 
> IMHO we need to ensure that shadows won't leak into a neighboring monitor when tiled to an output edge (either via regions or SnapToEdge). Same story with some bar in exclusive mode. I am not quite sure how to do that properly without unreasonably complicating everything. See for example `ssd_extents_update()` in `src/ssd/ssd-extents.c` for the calculation of the sizes of the invisible resize area which faces the same challenge.

Thanks for the feedback, I understand your concern.

However, I believe this issue is somewhat orthogonal to what this PR addresses. The shadow leakage can happen even without tiling — for example, with regular floating windows placed near the screen edge. So while the concern about shadows leaking into neighboring outputs is valid, I don't think it's directly caused or worsened by the changes in this PR.

That said, I agree it's worth exploring a more general solution in the future.

Consolatis:
> The shadow leakage can happen even without tiling — for example, with regular floating windows placed near the screen edge.

The difference between the two is that in the non-tiling / region case the user is fully aware of the situation as they moved the window into that direction. Same story for borders and other visual parts of the window content, that is usual operation of any stacking compositor supporting multiple monitors.

In the tiling case however the user tells the compositor to place a window in some defined region or at some screen edge. IMHO the compositor should ensure in those cases that the window is only placed there and won't spill into any other regions / screens.

Another option could be to include the size of the shadow in the calculation of the tiled geometry, that still feels a bit weird because then the window would not technically be tiled to that exact region (although it already works the same for the border and is a visual element so I guess its fine). See `view_apply_region_geometry()` and `view_apply_tiled_geometry()` in `src/view.c` and corresponding `ssd_thickness()` in `src/ssd/ssd.c`.

diredocks:
> Another option could be to include the size of the shadow in the calculation of the tiled geometry, that still feels a bit weird.

Including the shadow in the tiled geometry could work, but it might make the gaps between windows look too wide, especially when rc.gap is already in effect. Visually this could result in more spacing than intended.

diredocks:
My use case:



This means users can choose to add their own gaps if they want to avoid shadow leakage across outputs — or simply disable shadows when windows are snapped.

Consolatis:
> This means users can choose to add their own gaps if they want to avoid shadow leakage across outputs — or simply disable shadows when windows are snapped.

If you have gaps that are >= shadow size that would work. However, what in case of gaps < shadow size? In the window-to-window case like on your screenshot it would render the shadow on top of the contents of the neighboring window which also changes with the window focus. And on the screen / panel cases it would render those shadows on top of those / spill into the neighboring screen.

diredocks:
> what in case of gaps < shadow size? ...

That’s a fair point — shadows overlapping neighboring windows or panels when the gap is smaller than the shadow is definitely a valid concern, especially in tightly packed layouts.

Just to clarify: **this PR doesn’t change the default behavior and shadows on tiled windows remain disabled by default.** It only adds an option to explicitly enable shadows for tiled windows if the user prefers that look and is okay with the trade-offs.

For users who do enable it, they still have control — **they can adjust gaps or disable shadows for specific layouts.** This patch is mainly about making that choice available.

Consolatis:
A compromise would be to simply prevent activating the shadow-on-tile setting when there is no gap >= shadow-size.

That way there is no leaking shadows and the feature works fine when using gaps.

diredocks:
Done with https://github.com/labwc/labwc/pull/2789/commits/e08f8af667e5a7f3b728dc660edaf0192b10963d

When shadows are very subtle but large in size, they can appear almost invisible to the eye. Currently, we don’t have a way to render that kind of shadow for tiled windows. But as a trade-off, I’m personally satisfied with it.

src/ssd/ssd-shadow.c
```diff
@@ -289,10 +289,19 @@ ssd_shadow_update(struct ssd *ssd)
...
 	struct view *view = ssd->view;
+	struct theme *theme = ssd->view->server->theme;
 	bool maximized = view->maximized == VIEW_AXIS_BOTH;
-	bool show_shadows =
-		rc.shadows_enabled && !maximized &&
-		(!view_is_tiled(ssd->view) || rc.shadows_on_tiled);
+	bool tiled_shadows = false;
+	if (rc.shadows_on_tiled) {
+		if (rc.gap >= theme->window[THEME_ACTIVE].shadow_size &&
+			rc.gap >= theme->window[THEME_INACTIVE].shadow_size) {
```
Consolatis: Second line of a if condition needs to be double indented. E.g.
```c
if (rc.gap >= theme->window[THEME_ACTIVE].shadow_size
		&& rc.gap >= theme->window[THEME_INACTIVE].shadow_size) {
	tiled_shadows = true;
}
```

Consolatis:
Small nitpick above, other than that this LGTM.
Thanks!

Edit:
Commits should be squashed, we can also do that when merging though.

PR #2791, opened 2025-06-06
fix: missed judgment for some X11 property

DreamMaoMao:
refer to:

https://github.com/swaywm/sway/issues/7823

https://codeberg.org/dwl/dwl/pulls/1110

this issue also exists in labwc

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -117,10 +117,34 @@ xwayland_view_wants_focus(struct view *view)
 	case WLR_ICCCM_INPUT_MODEL_GLOBAL:
 		/*
 		 * Assume that NORMAL and DIALOG windows always want
-		 * focus. These window types should show up in the
+		 * focus, unless they are one of the special window types.
+		 * These window types should show up in the
 		 * Alt-Tab switcher and be automatically focused when
 		 * they become topmost.
 		 */
+		if (wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_COMBO) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DND) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DROPDOWN_MENU) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_MENU) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NOTIFICATION) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_POPUP_MENU) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_SPLASH) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DESKTOP) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_TOOLTIP) ||
+			wlr_xwayland_surface_has_window_type(xsurface,
+				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_UTILITY)) {
+			return VIEW_WANTS_FOCUS_NEVER;
+		}
+
 		return (wlr_xwayland_surface_has_window_type(xsurface,
 				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NORMAL)
 			|| wlr_xwayland_surface_has_window_type(xsurface,


johanmalm:
Ah, interesting.

@jlindgren90 - any views before we go for a merge?

jlindgren90:
This needs way more explanation and looks wrong at first glance. Combos, menus, and desktop types definitely want focus.

DreamMaoMao:
> This needs way more explanation and looks wrong at first glance. Combos, menus, and desktop types definitely want focus.

Could you provide some examples of application scenarios? I can do some tests.

He is doing well in my application scenario and has not broken anything that previously worked properly。

jlindgren90:
I think it's fair to ask first what problem you are trying to solve, and for an explanation of how your change solves it.

DreamMaoMao:
> I think it's fair to ask first what problem you are trying to solve, and for an explanation of how your change solves it.我认为

Just as mentioned in the link, it is to solve the problem that the drop-down menus or pop-up panels of some x11 Windows cannot be clicked. These Windows will directly exit when the composiotr attempts to actively focus on them and thus cannot be clicked. The above attributes are from hyprland(xwayland surface wantsFocus func). It works very well in my actual tests. I have checked the relevant attribute documents but couldn't find the reasons why the several attributes you mentioned need to obtain the focus. If you can provide the relevant documents for which they need the focus, I can check again.


https://specifications.freedesktop.org/wm-spec/latest/ar01s05.html

jlindgren90:
- What application do you see the issue with?
- What version of labwc (if built from git, give the commit hash)?
- What do you mean by the "synthesizer"?

DreamMaoMao:
synthesizer -> comositor.

version : labwc-git (c5b9e2665d4d17ace6a9027a065c746c70d967e1)

I can't demonstrate to you for the time being. You may not be able to use my software. Because it is exclusively used within a certain country.

There are also some other cases in other repositories, such as dwl. This also solved his problem, but I can't ask them what software they are using now either.

It sounds like it has temporarily reached a verification deadlock.

I'm unable to conduct further verification here for the time being. If you encounter this issue in the future, you can conduct your own research. I'll close it for now.

Consolatis:
If you can't disclose the software which shows this behavior for whatever reason then you could potentially find an alternative software which shows the same behavior or write a small reproducer for that.

I do understand the intention to not focus for example tooltips but as @jlindgren90 said above, things like combo-boxes, menus and similar may want keyboard focus to allow users to select things via the keyboard. Note that we are not actually giving keyboard focus in this case but sending a `WM_TAKE_FOCUS` request. The client is free to ignore that and thus not get focus.

I am not very versed in the input model of X11, if we go against the spec here that is something to discuss but I agree on not simply merging this PR just because it apparently fixes some random application.

DreamMaoMao:
> If you can't disclose the software which shows this behavior for whatever reason then you could potentially find an alternative software which shows the same behavior or write a small reproducer for that.
> 
> I do understand the intention to not focus for example tooltips but as @jlindgren90 said above, things like combo-boxes, menus and similar may want keyboard focus to allow users to select things via the keyboard. Note that we are not actually giving keyboard focus in this case but sending a `WM_TAKE_FOCUS` request. The client is free to ignore that and thus not get focus.
> 
> I am not very versed in the input model of X11, if we go against the spec here that is something to discuss but I agree on not simply merging this PR just because it apparently fixes some random application.


This workload exceeded my expectations. I don't have much extra time to spend here. So sorry, I can't work further on this. But, it can be used as a reference.

diredocks:
We're encountering this issue with WeChat, a widely used instant messaging app in China. You can install it via `flatpak install flathub com.tencent.WeChat` However, the app requires logging in via QR code, which means you'll also need to install it on your phone. 😅

On KDE, when you click one of the sidebar buttons after logging in, a popup appears as expected.



However, on labwc, the popup simply vanishes immediately after showing up.

The same behavior can also be observed in Zoom's popup menus, as described in [this Sway issue comment](https://github.com/swaywm/sway/issues/7823#issuecomment-1931912437). This issue also provided another Java application example that might be easier to test with.

@jlindgren90 @johanmalm @Consolatis

jlindgren90:
Please re-test with latest master, that should have fixed the issue with Zoom popups. Right @tokyo4j ?

I'm not personally willing to provide support for WeChat.

Edit: I tested zoom 6.4.12-1 from AUR and I don't see any issue with popups closing.

diredocks:
Changes in wlroots 0.19.0 related to this seem to be mostly function renames — see the [changelog](https://gitlab.freedesktop.org/wlroots/wlroots/-/releases/0.19.0) for details. For example, `wlr_xwayland_or_surface_wants_focus` has been replaced by `wlr_xwayland_surface_override_redirect_wants_focus`.

I'm not sure why the wlroots devs made this function, but judging by the name, it seems to suggest that certain types of windows are not intended to take focus.

That said, I'd like to clarify that this report is not about supporting WeChat specifically. Rather, WeChat is just a convenient example of how certain types of window can misbehave if not handled correctly.

Since behavior for these kinds of windows may not be clearly defined in the spec, it's important for compositors to align with de facto standards — i.e., how other major compositors like KWin, GNOME Mutter, or Sway behave in these cases. Ensuring consistent behavior across compositors improves compatibility and avoids app-specific issues caused by implementation differences.

As for the Zoom issue, it's possible that the recent fix came from Zoom itself, rather than from wlroots. So while it may appear resolved now, that might not be due to compositor-side changes.

jlindgren90:
Please provide a minimal reproducible test case and let's test it to see how it works under Openbox.

johanmalm:
Maybe my antennae are sensitive, but I detect slight tension in this conversation.

I'm adding the context below in the hope that it unites our thinking:

1. The fact that hyprland has added these and that WeChat works better on that platform is an interesting reference point and one we ought to take note of.
2. We (labwc maintainers) have found that keyboard focus is the hardest thing to get right in a compositor - harder even than stacking surfaces well.
3. We have spent literally years trying to get all toolkits, etc working well with focus. In that processes we have taken a couple of wrong turns which have been painful to users/devs and have meant that we have taken a few steps backwards before moving forward again.
4. Consequently - unless we are pretty sure we know what we're doing with focus behaviour changes, we like to do significant testing and push the change early in a release cycle to get user input. For example, with XWayland focus changes we really ought to give the Raspberry Pi platform a good chance to do some testing as they run some X11 clients with toolkits other than Qt and Gtk. Considering that we've just ported to wlroots-0.19 there may even be some sense in leaving any potential focus change to the next release to find it easier to diagnose breaking changes.
5. It really helps having a test case for us to experiment with given the context I have described above. I've never used WeChat, but looking at it quickly it does not appear to be the easiest for many of us to play with.
6. On a more practical note, I can often only log in once every 24 hours to review/merge/comment, so find it hard to keep up with the quick twists/turns of this PR. I also see these types of changes taking longer for the benefit of stability.
8. The Zoom popups were fixed by #1142 (which depended on wlroots-0.19 and was deeper than function name changes).
9. In response to the statement on specifications and what it is important for compositors to do...

> Since behavior for these kinds of windows may not be clearly defined in the spec, it's important for compositors to align with de facto standards — i.e., how other major compositors like KWin, GNOME Mutter, or Sway behave in these cases. Ensuring consistent behavior across compositors improves compatibility and avoids app-specific issues caused by implementation differences.

Yes, agree very much but have also had my fingers burnt many times so like to build confidence that we're not introducing unintended side-effects. Help is comparing behavious of compositors (like in this PR) is gratefully received.

10. If there is any chance of finding a test case (preferably not WeChat by the sounds of it) then I think that's the best way forward to evaluate this further. I suspect a Java one will be the mostly likely candidate. I can try some Netbeans / CLion stuff at some point if nobody else has the means, ability or appetite. I have not used them for a long time, but they definitely were problematic with wlroots in the early days.

I hope you don't mind if I re-open + set this one to draft so we can keep it for reference and evaluation. I think that will be the most helpful position.

diredocks:
I created a small program that partially reproduces the behavior I encountered in WeChat. Under labwc, the program vanishes immediately after startup, whereas it works correctly under KWin.

Log when running this program in labwc on my machine:

```
00:00:05.510 [xwayland/server.c:107] Starting Xwayland on :1
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Could not resolve keysym XF86RefreshRateToggle
> Warning:          Could not resolve keysym XF86Accessibility
> Warning:          Could not resolve keysym XF86DoNotDisturb
Errors from xkbcomp are not fatal to the X server
00:00:05.588 [types/wlr_compositor.c:771] New wlr_surface 0x14985980 (res 0x14985d50)
00:00:05.589 [xwayland/server.c:272] Xserver is ready
00:00:05.589 [xwayland/xwm.c:1983] xfixes version: 6.0
00:00:05.589 [xwayland/xwm.c:2004] xres version: 1.2
00:00:05.591 [../src/xwayland.c:1042] Connected to xwayland
00:00:05.591 [../src/xwayland.c:998] Syncing X11 atoms
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_DESKTOP: 293
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_DOCK: 311
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_TOOLBAR: 312
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_MENU: 290
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_UTILITY: 284
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_SPLASH: 292
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_DIALOG: 313
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_DROPDOWN_MENU: 287
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_POPUP_MENU: 288
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_TOOLTIP: 285
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_NOTIFICATION: 291
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_COMBO: 289
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_DND: 286
00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_NORMAL: 283
00:00:05.591 [../src/xwayland.c:1044] Disconnecting from xwayland
00:00:05.592 [types/wlr_compositor.c:771] New wlr_surface 0x149b4b90 (res 0x149b4f60)
00:00:05.592 [xwayland/xwm.c:1752] New xwayland surface: 0x149b4b90
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Unsupported maximum keycode 708, clipping.
>                   X11 cannot support keycodes above 255.
> Warning:          Could not resolve keysym XF86RefreshRateToggle
> Warning:          Could not resolve keysym XF86Accessibility
> Warning:          Could not resolve keysym XF86DoNotDisturb
Errors from xkbcomp are not fatal to the X server
00:00:05.603 [xwayland/xwm.c:887] unhandled X11 property 271 (WM_STATE) for window 4194305
00:00:05.603 [xwayland/xwm.c:1660] unhandled X11 event: MappingNotify (34)
00:00:05.603 [xwayland/xwm.c:1660] unhandled X11 event: MappingNotify (34)
00:00:05.603 [../src/tearing.c:48] New presentation hint 0 received for surface 0x149b4b90
00:00:05.603 [render/gles2/renderer.c:158] Created GL FBO for buffer 202x102
00:00:05.622 [../src/desktop-entry.c:334] failed to load icon file labwc
00:00:05.622 [../src/desktop-entry.c:334] failed to load icon file labwc
00:00:05.622 [../src/view-impl-common.c:40] [map] identifier=, title=
00:00:05.624 [xwayland/xwm.c:1660] unhandled X11 event: FocusOut (10)
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op ConfigureWindow (no minor), code Window (no extension), sequence 167, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op SendEvent (no minor), code Window (no extension), sequence 168, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op SendEvent (no minor), code Window (no extension), sequence 170, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op SetInputFocus (no minor), code Window (no extension), sequence 171, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 172, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 175, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 178, value 4194305
00:00:05.624 [xwayland/xwm.c:1633] xcb error: op DeleteProperty (no minor), code Window (no extension), sequence 179, value 4194305
```

Log when trying to make the WeChat popup appear:

```
00:01:55.410 [xwayland/xwm.c:1384] Unhandled NET_WM_STATE property change 341 (_NET_WM_STATE_ABOVE)
00:01:55.410 [xwayland/xwm.c:1384] Unhandled NET_WM_STATE property change 343 (_NET_WM_STATE_STAYS_ON_TOP)
00:01:55.410 [xwayland/xwm.c:1384] Unhandled NET_WM_STATE property change 342 (_NET_WM_STATE_BELOW)
00:01:55.410 [xwayland/xwm.c:887] unhandled X11 property 295 (XdndAware) for window 4194366
00:01:55.411 [xwayland/xwm.c:887] unhandled X11 property 338 (_NET_WM_ICON_NAME) for window 4194366
00:01:55.411 [xwayland/xwm.c:887] unhandled X11 property 345 (_NET_WM_USER_TIME) for window 4194366
00:01:55.411 [types/wlr_compositor.c:771] New wlr_surface 0x14a6f900 (res 0x149bc510)
00:01:55.411 [xwayland/xwm.c:1752] New xwayland surface: 0x14a6f900
00:01:55.412 [xwayland/xwm.c:887] unhandled X11 property 271 (WM_STATE) for window 4194366
00:01:55.412 [../src/tearing.c:48] New presentation hint 0 received for surface 0x14a6f900
00:01:55.414 [render/gles2/renderer.c:158] Created GL FBO for buffer 172x208
00:01:55.414 [../src/view-impl-common.c:40] [map] identifier=wechat, title=wechat
00:01:55.414 [xwayland/xwm.c:1660] unhandled X11 event: FocusOut (10)
00:01:55.415 [render/gles2/renderer.c:158] Created GL FBO for buffer 837x720
00:01:55.430 [render/gles2/renderer.c:158] Created GL FBO for buffer 172x208
00:01:55.510 [xwayland/xwm.c:887] unhandled X11 property 271 (WM_STATE) for window 4194366
00:01:55.510 [xwayland/xwm.c:1192] Failed to get window property
00:01:55.510 [xwayland/xwm.c:1192] Failed to get window property
```

The test program:

```c
#include <X11/Xlib.h>
#include <X11/Xutil.h>
#include <X11/Xatom.h>  // for XA_ATOM
#include <stdio.h>
#include <stdlib.h>

int main(void) {
    Display *dpy = XOpenDisplay(NULL);
    if (!dpy) {
        fprintf(stderr, "Unable to open display\n");
        return 1;
    }

    int screen = DefaultScreen(dpy);
    Window root = RootWindow(dpy, screen);

    /* Create a menu window to be managed by the compositor */
    XSetWindowAttributes attrs;
    attrs.override_redirect = False;
    attrs.background_pixel  = WhitePixel(dpy, screen);
    Window menu = XCreateWindow(
        dpy, root,
        200, 200, 200, 100, 1,
        CopyFromParent, CopyFromParent,
        CopyFromParent,
        CWBackPixel,
        &attrs
    );

    /* Set EWMH _NET_WM_WINDOW_TYPE to UTILITY */
    Atom netWmWindowType        = XInternAtom(dpy, "_NET_WM_WINDOW_TYPE", False);
    Atom netWmWindowTypeUtility = XInternAtom(dpy, "_NET_WM_WINDOW_TYPE_UTILITY", False);
    XChangeProperty(dpy, menu, netWmWindowType, XA_ATOM, 32,
                    PropModeReplace,
                    (unsigned char *)&netWmWindowTypeUtility,
                    1);

    /* Select events: button press, key press, expose, focus change, and structure notify */
    XSelectInput(dpy, menu,
                 ButtonPressMask | KeyPressMask | ExposureMask |
                 FocusChangeMask | StructureNotifyMask);

    /* Map (display) the window */
    XMapWindow(dpy, menu);
    XFlush(dpy);

    /* Wait for MapNotify to ensure the window is visible */
    XEvent ev;
    while (1) {
        XNextEvent(dpy, &ev);
        if (ev.type == MapNotify && ev.xmap.window == menu) {
            break;
        }
    }

    /* Request keyboard input focus now that the window is visible (avoids BadMatch) */
    XSetInputFocus(dpy, menu, RevertToNone, CurrentTime);
    XFlush(dpy);

    /* Draw simple text */
    GC gc = XCreateGC(dpy, menu, 0, NULL);
    XSetForeground(dpy, gc, BlackPixel(dpy, screen));

    int running = 1;
    while (running) {
        XNextEvent(dpy, &ev);
        switch (ev.type) {
        case Expose:
            XDrawString(dpy, menu, gc, 20, 50,
                        "Override-Redirect Utility", 25);
            break;
        case ButtonPress:
        case KeyPress:
            running = 0;
            break;
        case FocusOut:
            /* Close the window when focus is lost */
            running = 0;
            break;
        }
    }

    XFreeGC(dpy, gc);
    XDestroyWindow(dpy, menu);
    XCloseDisplay(dpy);
    return 0;
}
```

Notice that `[xwayland/xwm.c:1660] unhandled X11 event: FocusOut (10)`, which is quite suspicious for me.

johanmalm:
Thanks. That's interesting.

It's intermittent for me (sometimes the window stays and sometimes it doesn't). This is the case for both labwc-master and with the code in this PR.

In other words, there is something else going on other than what this PR does.

johanmalm:
If I run with this...

```C
        case FocusOut:
	    fprintf(stderr, "FocusOut\n");
            break;
        case FocusIn:
	    fprintf(stderr, "FocusIn\n");
            break;
        }
```


...I get a mix of either

```
FocusIn
```

or

```
FocusIn
FocusOut
FocusIn
```

So, it looks like the `FocusOut` is a temporary state and that it always goes back to `FocusIn`.

diredocks:
What's really strange is that the patch in this PR actually fixed the popup issue in WeChat — and I’m not entirely sure why.

The test program I used only partially reproduces the popup behavior we see in WeChat, but it still highlights some inconsistencies across different compositors.

For example, when testing with Niri, the popup does appear, but it keeps sending Enter keystrokes back to the terminal that launched it — which seems unintended.

diredocks:
I tried backporting this patch to release 0.8.4, but it didn't fix the popup issue in WeChat.
There must be some other changes between 0.8.4 and master that, together with this patch, made it work.

DreamMaoMao:
I feel the direction of the discussion here has deviated. The fundamental reason these windows exit is not because they themselves receive focus, but because their gaining focus causes the parent window to lose focus. Even if these menu child windows are not focused, but other windows are focused instead, these child windows will still exit—because their parent window has lost focus. Therefore, studying the events of these menu windows isn't very meaningful. We just need to confirm why these windows require focus—what would break if they don't receive it?

By the way, the fact that this window has not gained focus does not mean that the surface on this window cannot gain focus. The surface in the menu options can still actively pointer focus.

Consolatis:
> 00:00:05.591 [../src/xwayland.c:1013] Got X11 atom for _NET_WM_WINDOW_TYPE_DESKTOP: 293

What labwc version are you testing this with? This log should not appear on the master branch.

Edit:
My viewpoint in general is that there are two ways to approach situations like this:

First one is:
We observe examples which trigger the issue and then figure out if the reason for that is labwc breaking the spec or the examples breaking the spec. If it turns out its labwc then we should fix it and follow the spec. If its the examples then we evaluate the potential side effects of a workaround in labwc and if they seems acceptable add a workaround. If the potential side effects are not acceptable we keep it as is.

Second one is:
We study the X11 input model specs and try to confirm we are spec compliant and if not fix that if possible. That might not really help us get to the ground of the underlying issue here though.

DreamMaoMao:
what do you mean by "example"? 

If you are talking about wlroots, as far as I know, its current support for the relevant attributes of xwayland is incomplete and is still being improved.


However, x11 has already made standard descriptions of the Windows represented by these attributes, and hyprland determines whether focus is needed based on the functions of these Windows

https://specifications.freedesktop.org/wm-spec/1.5/ar01s05.html


for example:

_NET_WM_WINDOW_TYPE_DESKTOP

He represents the desktop background window at the back. Why does this kind of window need to be focused on?

but jlindgren90 think they need focus,I don't know what his basis is
image

jlindgren90:
> I tried backporting this patch to release 0.8.4, but it didn't fix the popup issue in WeChat. There must be some other changes between 0.8.4 and master that, together with this patch, made it work.

Now it comes out. 

There is an entire series of focus changes in wlroots and labwc fixing focus issues which you are missing. I know because I authored them.

Please re-test with labwc master and let us know if the problem still exists. I suspect it is the same exact issue as we already fixed.

jlindgren90:
> > I tried backporting this patch to release 0.8.4, but it didn't fix the popup issue in WeChat. There must be some other changes between 0.8.4 and master that, together with this patch, made it work.
> 
> Now it comes out.
> 
> There is an entire series of focus issues in wlroots and labwc fixing focus changes which you are missing. I know because I authored them.
> 
> Please re-test with labwc master and let us know if the problem still exists. I suspect it is the same exact issue as we already fixed.

diredocks:
Following tests are done in labwc `0.8.4-79-gcc84cee7`

In Wechat, when releasing the sidebar button, the popup briefly appeared and then vanished:

```
00:00:30.434 [xwayland/xwm.c:1619] Unhandled NET_WM_STATE property change 350 (_NET_WM_STATE_STAYS_ON_TOP)
00:00:30.434 [xwayland/xwm.c:1105] unhandled X11 property 307 (XdndAware) for window 4194355
00:00:30.434 [xwayland/xwm.c:1105] unhandled X11 property 348 (_NET_WM_ICON_NAME) for window 4194355
00:00:30.434 [xwayland/xwm.c:1105] unhandled X11 property 351 (_NET_WM_USER_TIME) for window 4194355
00:00:30.435 [render/gles2/renderer.c:157] Created GL FBO for buffer 837x640
00:00:30.436 [types/wlr_compositor.c:786] New wlr_surface 0x249ee9e0 (res 0x249f4120)
00:00:30.436 [xwayland/xwm.c:2055] New xwayland surface: 0x249ee9e0
00:00:30.436 [xwayland/xwm.c:695] XCB_ATOM_NET_STARTUP_ID: (null)
00:00:30.436 [xwayland/xwm.c:1105] unhandled X11 property 279 (WM_STATE) for window 4194355
00:00:30.436 [../src/tearing.c:48] New presentation hint 0 received for surface 0x249ee9e0
00:00:30.439 [render/gles2/renderer.c:157] Created GL FBO for buffer 172x208
00:00:30.439 [../src/view-impl-common.c:40] [map] identifier=wechat, title=wechat
00:00:30.439 [render/swapchain.c:101] Allocating new swapchain buffer
00:00:30.440 [render/allocator/gbm.c:146] Allocated 32x32 GBM buffer with format AR24 (0x34325241), modifier GFX9,64KB_D_X,PIPE_XOR_BITS=2,BANK_XOR_BITS=0,RB=1,PIPE=2,DCC,DCC_MAX_COMPRESSED_BLOCK=64B,DCC_INDEPENDENT_64B,DCC_CONSTANT_ENCODE,DCC_PIPE_ALIGN (0x020000044051BA01)
00:00:30.440 [render/gles2/renderer.c:157] Created GL FBO for buffer 32x32
00:00:30.440 [backend/wayland/output.c:217] DMA-BUF imported into parent Wayland compositor
00:00:30.440 [render/swapchain.c:101] Allocating new swapchain buffer
00:00:30.440 [render/allocator/gbm.c:146] Allocated 32x32 GBM buffer with format AR24 (0x34325241), modifier GFX9,64KB_D_X,PIPE_XOR_BITS=2,BANK_XOR_BITS=0,RB=1,PIPE=2,DCC,DCC_MAX_COMPRESSED_BLOCK=64B,DCC_INDEPENDENT_64B,DCC_CONSTANT_ENCODE,DCC_PIPE_ALIGN (0x020000044051BA01)
00:00:30.440 [render/gles2/renderer.c:157] Created GL FBO for buffer 32x32
00:00:30.441 [backend/wayland/output.c:217] DMA-BUF imported into parent Wayland compositor
00:00:30.441 [render/swapchain.c:101] Allocating new swapchain buffer
00:00:30.441 [render/allocator/gbm.c:146] Allocated 16x16 GBM buffer with format AR24 (0x34325241), modifier GFX9,64KB_D_X,PIPE_XOR_BITS=2,BANK_XOR_BITS=0,RB=1,PIPE=2,DCC,DCC_MAX_COMPRESSED_BLOCK=64B,DCC_INDEPENDENT_64B,DCC_CONSTANT_ENCODE,DCC_PIPE_ALIGN (0x020000044051BA01)
00:00:30.441 [render/gles2/renderer.c:157] Created GL FBO for buffer 16x16
00:00:30.441 [backend/wayland/output.c:217] DMA-BUF imported into parent Wayland compositor
00:00:30.442 [render/gles2/renderer.c:157] Created GL FBO for buffer 837x640
00:00:30.442 [xwayland/xwm.c:1947] unhandled X11 event: FocusOut (10)
00:00:30.454 [render/swapchain.c:101] Allocating new swapchain buffer
00:00:30.455 [render/allocator/gbm.c:146] Allocated 16x16 GBM buffer with format AR24 (0x34325241), modifier GFX9,64KB_D_X,PIPE_XOR_BITS=2,BANK_XOR_BITS=0,RB=1,PIPE=2,DCC,DCC_MAX_COMPRESSED_BLOCK=64B,DCC_INDEPENDENT_64B,DCC_CONSTANT_ENCODE,DCC_PIPE_ALIGN (0x020000044051BA01)
00:00:30.455 [render/gles2/renderer.c:157] Created GL FBO for buffer 16x16
00:00:30.455 [backend/wayland/output.c:217] DMA-BUF imported into parent Wayland compositor
00:00:30.455 [render/gles2/renderer.c:157] Created GL FBO for buffer 172x208
00:00:30.537 [xwayland/xwm.c:1410] Failed to get window property
00:00:30.537 [xwayland/xwm.c:1947] unhandled X11 event: FocusOut (10)
00:00:30.537 [xwayland/xwm.c:1410] Failed to get window property
00:00:30.537 [xwayland/xwm.c:1410] Failed to get window property
00:00:30.537 [xwayland/xwm.c:1410] Failed to get window property
```

Running the test program:

```
00:02:11.189 [xwayland/server.c:107] Starting Xwayland on :1
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Could not resolve keysym XF86RefreshRateToggle
> Warning:          Could not resolve keysym XF86Accessibility
> Warning:          Could not resolve keysym XF86DoNotDisturb
Errors from xkbcomp are not fatal to the X server
00:02:11.260 [types/wlr_compositor.c:786] New wlr_surface 0x24936cb0 (res 0x248b3020)
00:02:11.261 [xwayland/server.c:283] Xserver is ready
00:02:11.261 [xwayland/xwm.c:2286] xfixes version: 6.0
00:02:11.261 [xwayland/xwm.c:2307] xres version: 1.2
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Warning:          Unsupported maximum keycode 708, clipping.
>                   X11 cannot support keycodes above 255.
> Warning:          Could not resolve keysym XF86RefreshRateToggle
> Warning:          Could not resolve keysym XF86Accessibility
> Warning:          Could not resolve keysym XF86DoNotDisturb
Errors from xkbcomp are not fatal to the X server
00:02:11.269 [xwayland/xwm.c:1947] unhandled X11 event: MappingNotify (34)
00:02:11.269 [xwayland/xwm.c:1947] unhandled X11 event: MappingNotify (34)
00:02:11.271 [types/wlr_compositor.c:786] New wlr_surface 0x2493d1d0 (res 0x2493d5a0)
00:02:11.271 [xwayland/xwm.c:2055] New xwayland surface: 0x2493d1d0
00:02:11.272 [xwayland/xwm.c:695] XCB_ATOM_NET_STARTUP_ID: (null)
00:02:11.273 [xwayland/xwm.c:1105] unhandled X11 property 279 (WM_STATE) for window 4194305
00:02:11.273 [../src/tearing.c:48] New presentation hint 0 received for surface 0x2493d1d0
00:02:11.273 [render/gles2/renderer.c:157] Created GL FBO for buffer 202x102
00:02:11.274 [../src/common/scaled-icon-buffer.c:75] loading icon by app_id
00:02:11.274 [../src/common/scaled-icon-buffer.c:80] loading fallback icon
00:02:11.286 [../src/desktop-entry.c:334] failed to load icon file labwc
00:02:11.287 [../src/common/scaled-icon-buffer.c:75] loading icon by app_id
00:02:11.287 [../src/common/scaled-icon-buffer.c:80] loading fallback icon
00:02:11.287 [../src/desktop-entry.c:334] failed to load icon file labwc
00:02:11.287 [../src/common/scaled-icon-buffer.c:75] loading icon by app_id
00:02:11.287 [../src/common/scaled-icon-buffer.c:80] loading fallback icon
00:02:11.287 [../src/desktop-entry.c:334] failed to load icon file labwc
00:02:11.287 [../src/common/scaled-icon-buffer.c:75] loading icon by app_id
00:02:11.287 [../src/common/scaled-icon-buffer.c:80] loading fallback icon
00:02:11.287 [../src/desktop-entry.c:334] failed to load icon file labwc
00:02:11.287 [../src/view-impl-common.c:40] [map] identifier=, title=
00:02:11.288 [xwayland/xwm.c:1947] unhandled X11 event: FocusOut (10)
00:02:11.288 [xwayland/xwm.c:1947] unhandled X11 event: FocusOut (10)
00:02:11.288 [xwayland/xwm.c:1410] Failed to get window property
00:02:11.288 [xwayland/xwm.c:1947] unhandled X11 event: FocusOut (10)
00:02:11.288 [xwayland/xwm.c:1920] xcb error: op ConfigureWindow (no minor), code Window (no extension), sequence 187, value 4194305
00:02:11.288 [xwayland/xwm.c:1920] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 190, value 4194305
00:02:11.288 [xwayland/xwm.c:1920] xcb error: op ConfigureWindow (no minor), code Window (no extension), sequence 192, value 4194305
00:02:11.288 [xwayland/xwm.c:1920] xcb error: op ChangeProperty (no minor), code Window (no extension), sequence 198, value 4194305
00:02:11.288 [xwayland/xwm.c:1920] xcb error: op DeleteProperty (no minor), code Window (no extension), sequence 199, value 4194305
```

jlindgren90:
Thanks for testing. It is interesting that in labwc, the test program receives a FocusOut right at startup. I tested under Openbox and no FocusOut is received there (until Alt-Tabbing back to some other window).

However, as @johanmalm pointed out, the patch in this PR doesn't make any difference. The FocusOut is still received at startup (under labwc with the patch) and the test program exits.

diredocks:
Thanks for checking out. I’d like to point out that this patch did fix the issue with WeChat:

```
00:01:35.514 [../src/action.c:889] Handling action 25: Focus
00:01:35.514 [../src/action.c:889] Handling action 29: Raise
00:01:35.516 [xwayland/xwm.c:1105] unhandled X11 property 351 (_NET_WM_USER_TIME) for window 4194324
00:01:37.374 [xwayland/xwm.c:1619] Unhandled NET_WM_STATE property change 350 (_NET_WM_STATE_STAYS_ON_TOP)
00:01:37.375 [xwayland/xwm.c:1105] unhandled X11 property 307 (XdndAware) for window 4194372
00:01:37.375 [xwayland/xwm.c:1105] unhandled X11 property 348 (_NET_WM_ICON_NAME) for window 4194372
00:01:37.375 [xwayland/xwm.c:1105] unhandled X11 property 351 (_NET_WM_USER_TIME) for window 4194372
00:01:37.376 [types/wlr_compositor.c:786] New wlr_surface 0x2bf6cc50 (res 0x2beb62f0)
00:01:37.376 [xwayland/xwm.c:2055] New xwayland surface: 0x2bf6cc50
00:01:37.376 [xwayland/xwm.c:695] XCB_ATOM_NET_STARTUP_ID: (null)
00:01:37.376 [xwayland/xwm.c:1105] unhandled X11 property 279 (WM_STATE) for window 4194372
00:01:37.376 [../src/tearing.c:48] New presentation hint 0 received for surface 0x2bf6cc50
00:01:37.379 [render/gles2/renderer.c:157] Created GL FBO for buffer 172x208
00:01:37.379 [../src/view-impl-common.c:40] [map] identifier=wechat, title=wechat
00:01:37.379 [render/swapchain.c:101] Allocating new swapchain buffer
00:01:37.379 [render/allocator/gbm.c:146] Allocated 32x32 GBM buffer with format AR24 (0x34325241), modifier GFX9,64KB_D_X,PIPE_XOR_BITS=2,BANK_XOR_BITS=0,RB=1,PIPE=2,DCC,DCC_MAX_COMPRESSED_BLOCK=64B,DCC_INDEPENDENT_64B,DCC_CONSTANT_ENCODE,DCC_PIPE_ALIGN (0x020000044051BA01)
00:01:37.379 [render/gles2/renderer.c:157] Created GL FBO for buffer 32x32
00:01:37.380 [backend/wayland/output.c:217] DMA-BUF imported into parent Wayland compositor
```

After releasing the button, the popup appears as expected, and no FocusOut is triggered for it.

While the patch seems to help in at least one real-world scenario, it’s unclear what exactly changed or why it worked there but not in the minimal test case.

Consolatis:
> `00:02:11.288 [xwayland/xwm.c:1947] unhandled X11 event: FocusOut (10)`

This also makes me curious though. Maybe that must be handled to make things work properly? (e.g. via wlroots in the long term + `wlr_xwayland->user_event_handler` in the short term).

Edit:
> However, x11 has already made standard descriptions of the Windows represented by these attributes, and hyprland determines whether focus is needed based on the functions of these Windows

Its not really that easy: https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#input_focus

Edit 2:
Maybe we also need to parse the `WM_PROTOCOLS` property of the given window to decide when to send a `WM_TAKE_FOCUS` message and when not.

jlindgren90:
> This also makes me curious though. Maybe that must be handled to make things work properly?

FocusIn and FocusOut typically come in pairs. wlroots listens only to FocusIn and infers that the previously focused window lost focus. I don't think there's any need to also explicitly listen for FocusOut.

jlindgren90:
> Maybe we also need to parse the `WM_PROTOCOLS` property of the given window to decide when to send a `WM_TAKE_FOCUS` message and when not.

We do already. That's what `view_want_focus` looks at.

Consolatis:
Looking a bit further into the `WLR_ICCCM_INPUT_MODEL_GLOBAL` branch in `xwayland_view_wants_focus()`, the logic seems to contradict the comment above it (and the two comments seems to contradict each other). For `NORMAL` and `DIALOG` we do give focus but the comment states
> It wants to prevent the window manager from setting the input focus to any of its windows [because it may or may not want focus].
> [..]
>  Assume that NORMAL and DIALOG windows always want focus. These window types should show up in the Alt-Tab switcher and be automatically focused when they become topmost.

Why are we returning `VIEW_WANTS_FOCUS_ALWAYS` in that case rather than `VIEW_WANTS_FOCUS_OFFER`? And does it fix the issue here if we always return `VIEW_WANTS_FOCUS_OFFER` in that branch?

diredocks:
> Why are we returning `VIEW_WANTS_FOCUS_ALWAYS` in that case rather than `VIEW_WANTS_FOCUS_OFFER`? And does it fix the issue here if we always return `VIEW_WANTS_FOCUS_OFFER` in that branch?

Always returning `VIEW_WANTS_FOCUS_OFFER` fixes the issue with WeChat, but no luck with out test program—and what are the side effects of it?

Consolatis:
> and what are the side effects of it?

Likely that the windows won't show up in `A-Tab` and friends but that is a labwc internal issue and not related to the spec and should be decoupled. So that fact should not prevent us from following the spec if it is indeed the right thing to never give focus to a globally active client but only send a `WM_TAKE_FOCUS` request.

jlindgren90:
> Likely that the windows won't show up in `A-Tab` and friends

Yes, precisely.

> So that fact should not prevent us from following the spec if it is indeed the right thing to never give focus to a globally active client but only send a `WM_TAKE_FOCUS` request.

We still send these windows WM_TAKE_FOCUS rather than setting the focus forcibly (this is handled by wlroots internally). However I am not sure what happens if a NORMAL or DIALOG window actually refuses focus - that would be rather strange and I doubt it has been tested.

jlindgren90:
> Always returning `VIEW_WANTS_FOCUS_OFFER` fixes the issue with WeChat

That's bizarre. It would be good to understand why that's the case.

diredocks:
I think I’ve figured out the key difference between our test program and the popup behavior in WeChat.

In WeChat, the popup is a separate window that doesn't actively receive focus. In fact, I’m not even sure it can receive focus. When the popup appears, it doesn’t seem to be focused.

Even if I click on the blank areas of the popup in KWin, the taskbar doesn’t indicate that it's selected.
Instead, it just clears the `demandsAttention` flag, and I still can’t actually focus on it. Keyboard events continue to be sent to the main window that spawned the popup.

When the popup appears:
image
image

Clicking on the popup:
image
image

As I mentioned earlier, always returning `VIEW_WANTS_FOCUS_OFFER` fixes the issue with WeChat. This might be because it sends a focus offer to the popup rather than forcing focus. As a result, the parent window retains focus, which prevents the popup from being closed. In other words, the popup doesn’t vanish because it itself loses focus—but because its parent window does.

diredocks:
Btw, the behavior we see in test program is still pretty strange. When running it in KWin Wayland, it doesn't seem to be get any `FOCUS_OUT` event until we actually click on somewhere else:

image
image

johanmalm:
@diredocks As you've got the test bed, any chance of evaluating which of the properties in this PR make(s) it work?

diredocks:
The popup in WeChat is a UTILITY-type window, and this PR prevents it from being focused. Always returning a focus offer also allows it to choose whether to take focus or not. That’s why both approaches end up working.

jlindgren90:
> The popup in WeChat is a UTILITY-type window, and this PR prevents it from being focused. Always returning a focus offer also allows it to choose whether to take focus or not.

The current code should already return OFFER for a UTILITY type window though. So what am I missing here?

If someone could describe clearly what's happening (which branch is taken etc.) in the current code, that would be helpful.

Consolatis:
> The current code should already return OFFER for a UTILITY type window though

I wondered the same. My only guess is that the popup has either `NORMAL` or `DIALOG` set as well, additionally to `UTILITY`.

> We still send these windows WM_TAKE_FOCUS rather than setting the focus forcibly (this is handled by wlroots internally).

I am not sure about that, at least I didn't spot this behavior in `xwm.c` on a quick scan but maybe I missed something.

Regarding the related not-showing-up-in-a-tab-and-taskbar when we switch over to always send an offer in those cases: now with wlroots 0.19.x we do have access to `surface->skip_taskbar` and `surface->skip_pager` (and correlated event emitters), that in combination with `wlr_xwayland_surface_has_window_type()` should allow us to decouple the focus semantics from the visibility in a-tab / taskbars from what I remember.

jlindgren90:
> > We still send these windows WM_TAKE_FOCUS rather than setting the focus forcibly (this is handled by wlroots internally).
> 
> I am not sure about that, at least I didn't spot this behavior in `xwm.c` on a quick scan but maybe I missed something.

See https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c?ref_type=heads#L406 added by https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/d0b3aed584ecb4cbaccb94e76ca5bcbc952075de some years back.

diredocks:
I added some log like this:

```c
	case WLR_ICCCM_INPUT_MODEL_GLOBAL:
		/*
		 * Assume that NORMAL and DIALOG windows always want
		 * focus. These window types should show up in the
		 * Alt-Tab switcher and be automatically focused when
		 * they become topmost.
		 */
		if (wlr_xwayland_surface_has_window_type(xsurface, WLR_XWAYLAND_NET_WM_WINDOW_TYPE_UTILITY)
				&& wlr_xwayland_surface_has_window_type(xsurface, WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NORMAL)) {
			wlr_log(WLR_DEBUG, "LMAO it's utility and normal");
		}
		if (wlr_xwayland_surface_has_window_type(xsurface, WLR_XWAYLAND_NET_WM_WINDOW_TYPE_UTILITY)
				&& wlr_xwayland_surface_has_window_type(xsurface, WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DIALOG)) {
			wlr_log(WLR_DEBUG, "LOL it's utility and dialog");
		}
		return (wlr_xwayland_surface_has_window_type(xsurface,
				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NORMAL)
			|| wlr_xwayland_surface_has_window_type(xsurface,
				WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DIALOG)) ?
			VIEW_WANTS_FOCUS_ALWAYS : VIEW_WANTS_FOCUS_OFFER;
```

And here's what's interesting: the popup in Wechat does set two flags: 

image

Consolatis:
> > > We still send these windows WM_TAKE_FOCUS rather than setting the focus forcibly (this is handled by wlroots internally).
> > 
> > 
> > I am not sure about that, at least I didn't spot this behavior in `xwm.c` on a quick scan but maybe I missed something.
> 
> See https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xwayland/xwm.c?ref_type=heads#L406 added by https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/d0b3aed584ecb4cbaccb94e76ca5bcbc952075de some years back.

Whoops, indeed missed that completely for some reason. The only two relevant differences I see is that this code path doesn't check if the window actually contains `WM_TAKE_FOCUS` in its protocols (assumingly it does not which would explain the issue we see) and `wlr_xwayland_surface_activate()` unconditionally calls `xwm_set_focused_window()` and makes it the focused window, regardless of the internal fallback (might also explain the issue). Both of these don't feel right to me.

johanmalm:
Would you be able to pinpoint which "type" (for example WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NOTIFICATION) it is.

diredocks:
> Would you be able to pinpoint which "type" (for example WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NOTIFICATION) it is.

The popup has two types:
```
00:00:30.730 [../src/xwayland.c:133] Window type: WLR_XWAYLAND_NET_WM_WINDOW_TYPE_UTILITY
00:00:30.730 [../src/xwayland.c:142] Window type: WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NORMAL
```

Code I use for logging this:
```c
#define CHECK_TYPE(x) \
	if (wlr_xwayland_surface_has_window_type(xsurface, x)) { \
		wlr_log(WLR_DEBUG, "Window type: " #x); \
	}

...
	case WLR_ICCCM_INPUT_MODEL_GLOBAL:
		/*
		 * Assume that NORMAL and DIALOG windows always want
		 * focus. These window types should show up in the
		 * Alt-Tab switcher and be automatically focused when
		 * they become topmost.
		 */
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DESKTOP);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DOCK);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_TOOLBAR);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_MENU);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_UTILITY);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_SPLASH);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DIALOG);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DROPDOWN_MENU);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_POPUP_MENU);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_TOOLTIP);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NOTIFICATION);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_COMBO);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_DND);
		CHECK_TYPE(WLR_XWAYLAND_NET_WM_WINDOW_TYPE_NORMAL);
		return (wlr_xwayland_surface_has_window_type(xsurface,
...
```

jlindgren90:
> The only two relevant differences I see is that this code path doesn't check if the window actually contains `WM_TAKE_FOCUS` in its protocols (assumingly it does not which would explain the issue we see) and `wlr_xwayland_surface_activate()` unconditionally calls `xwm_set_focused_window()` and makes it the focused window, regardless of the internal fallback (might also explain the issue). Both of these don't feel right to me.

Yes, there's definitely some cause for concern there - it seems like the actual input focus could get out-of-sync with the various NET_WM properties and the internal wlroots state.

> Regarding the related not-showing-up-in-a-tab-and-taskbar when we switch over to always send an offer in those cases: now with wlroots 0.19.x we do have access to surface->skip_taskbar and surface->skip_pager (and correlated event emitters), that in combination with wlr_xwayland_surface_has_window_type() should allow us to decouple the focus semantics from the visibility in a-tab / taskbars from what I remember.

I think you've convinced me, this is what we should do (decouple the focus semantics from the visibility). It does open up a possibility that certain "weird" windows like the WeChat popup would show up in the Alt-Tab switcher and taskbar, but not actually be focusable. That's certainly less than ideal, but from https://github.com/labwc/labwc/pull/2791#issuecomment-2954247992 it sounds like KWin does something similar.

jlindgren90:
How about something like this: https://github.com/jlindgren90/labwc/commit/b13db707abcc6f554856153943eba13a1b4d25ef

@diredocks Can you please test whether this also fixes the WeChat issue?

@Consolatis I did also take note of your suggestion to pay attention to skip_pager/skip_taskbar hints, but I think that can be done later.

diredocks:
> @diredocks Can you please test whether this also fixes the WeChat issue?

It works like a charm!

Consolatis:
> How about something like this: [jlindgren90@b13db70](https://github.com/jlindgren90/labwc/commit/b13db707abcc6f554856153943eba13a1b4d25ef)

Looks fine to me as an intermediate step, e.g. we still use the `NORMAL || DIALOG` hint to decide on the visibility but always offer the focus rather than forcing it.

>  It does open up a possibility that certain "weird" windows like the WeChat popup would show up in the Alt-Tab switcher and taskbar, but not actually be focusable.

@diredocks could you check to see if the popup has either `xsurface->skip_pager` or `xsurface->skip_taskbar` set?

jlindgren90:
Thanks for testing. I created a new PR:
- #2811

johanmalm:
> for example:
> 
> _NET_WM_WINDOW_TYPE_DESKTOP
> 
> He represents the desktop background window at the back. Why does this kind of window need to be focused on?

Could an example be `pcmanfm --desktop` which sets the background image but also has icons with context menus which a user can navigate with a pointer or keyboard?

diredocks:
> @diredocks could you check to see if the popup has either `xsurface->skip_pager` or `xsurface->skip_taskbar` set?

None of these attributes are set

Consolatis:
> > @diredocks could you check to see if the popup has either `xsurface->skip_pager` or `xsurface->skip_taskbar` set?
> 
> None of these attributes are set

Interesting. We may need to read into what exactly are the X11 policies for a-tab / taskbar of `NORMAL | UTILITY` windows without those flags set. Or wechat is simply broken or its intended that the popup shows up in taskbars / alt-tab.

Anyway, this uncovered an issue in labwc and a connected one in wlroots. We can work around it on the labwc side and may also fix it on the wlroots side long term so thanks for reporting and testing!

diredocks:
Regarding the WeChat popup, I have something to add: Neither labwc nor KDE shows this popup in the window switcher (KDE calls it the tabbox), but both display it in the taskbar(waybar in labwc).

image
(what shown in tabbox here is the main window of wechat)

image

jlindgren90:
Closing this now that #2811 has been merged.

I think the unexpected initial FocusOut+FocusIn event pair is a separate issue. My guess is that it's a result of wlroots trying to prevent focus-stealing initially, and labwc only later processing the Focus action to give the new window focus again. I don't know if it necessarily *needs* to be fixed but it's something to keep an eye on.

PR #2793, opened 2025-06-06
Enable content-type-v1 protocol

Sunderland93:
The [content-type-v1](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/content-type/content-type-v1.xml?ref_type=heads) protocol allows clients to send compositor a hint about the type of content being displayed on its surface so it could potentially make some sort of optimization. The protocol itself exposes 4 different types of content: none, photo, video, and game.  Tested on RetroArch (always set `game` type) and MPV (set `video` type by default)

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3599


ahesford:
This change isn't doing anything.

tokyo4j:
I think @ahesford is right. As long as we don't actually have any different optimizations in rendering or presenting window contents, there's not much point in supporting the protocol.

Issue #2729, opened 2025-05-19
Interesting sway changes between wlroots `0.18` and `0.19`.


johanmalm:
Intend processing after `0.9.0` release.

# General

- [ ] Drop wl_drm https://github.com/swaywm/sway/commit/1d4632f97fb6ee61abe350ae9f76270562396553
- [ ] Use pthread_atform() https://github.com/swaywm/sway/commit/0e19d85d37e556721f982c3f63d4e2927f306b18
- [ ] Replace signal() with sigaction() https://github.com/swaywm/sway/commit/38a42f97d46931e97693610c999ea51834b71352
- [ ] Use wl_event_loop_add_signal for exit signals https://github.com/swaywm/sway/commit/8f089f0229eabbcd3f68d641d5b826220f1edb0c
- [ ] Use SIG_IGN for SIGCHLD instead of our own handler https://github.com/swaywm/sway/commit/8238e5242bdbbc4c3b7cba0651c620a89b872a27
- [ ] docs: use "window" instead of "view" throughout https://github.com/swaywm/sway/commit/10e50e6bf9b63b205c141f97a5709fd4d405542f
- [ ] has_prefix https://github.com/swaywm/sway/commit/0c60d1581f7b12ae472c786b7dfe27a1c6ec9a47
- [ ] drag-lock enabled sticky https://github.com/swaywm/sway/commit/bbadf9b8b10d171a6d5196da7716ea50ee7a6062
- [ ] dump output state https://github.com/swaywm/sway/commit/9765c29be11f04f903a45f34b9142a865784fc46
- [ ] Use `wlr_scene_output_needs_frame()` https://github.com/labwc/labwc/pull/2388#issuecomment-2889167576 https://github.com/swaywm/sway/commit/5a3621460f193416605ad786e33687952527df21
- [ ] wmenu-run https://github.com/swaywm/sway/commit/b44015578a3d53cdd9436850202d4405696c1f52
- [x] restack https://github.com/swaywm/sway/commit/951a22c2445f5c32b831bac0db86869627940402

New Protos

- [ ] ext-data-control manager https://github.com/swaywm/sway/commit/5d7b9a8320f8999059f287734c1df76289b01a27
- [ ] Add ext-image-copy-capture-v1 and ext-image-capture-source-v1 https://github.com/swaywm/sway/commit/8acb0482da68af69d52ab168f9e30e2464b9c7a3
[...]

Issue #1156, opened 2023-10-12
Allow actions to be triggered by touchpad gestures


FriederHannenheim:
Launching programs or maybe moving windows by 3 or 4 finger drag would be awesome

stefonarch:
I'm using 3-fingerswipe for workspace switching:
```

 cat .config/libinput-gestures.conf 
#$ libinput-gestures-setup stop
#$ libinput-gestures-setup start
# https://github.com/bulletmark/libinput-gestures

# sway
gesture: swipe left 3   swaymsg -t command workspace next_on_output
gesture: swipe right 3  swaymsg -t command workspace prev_on_output

# labwc
gesture: swipe left 3   wtype  -M ctrl -M alt -P right
gesture: swipe right 3 wtype  -M ctrl -M alt -P left

# openbox

#gesture: swipe left 3 xdotool key ctrl+alt+Right
#gesture: swipe right 3  xdotool key ctrl+alt+Left


```

Consolatis:
Work-in-progress PR:
- #3227

PR #2795, opened 2025-06-06
libinput: implement threeFingerDrag feature

### LINKED ISSUE #2647: support 3-finger drag
Libinput 1.28 with 3(or 4)-finger drag support has been released. I think we can adopt it like:
```xml
<libinput>
  <device>
    <!-- 'yes' is equal to '3' -->
    <threeFingerDrag>no|yes|3|4</threeFingerDrag>
  </device>
</libinput>
```

I'll mark this issue as a "good first issue" and wait for 1-2 weeks or so to see if there's any interested contributor. I remember my first contribution to labwc was adding `naturalScroll` and `dragLock`, and now I'm a maintainer of labwc!

m4rch3n1ng:
closes #2647.

i took most of the docs directly [from libinput](https://gitlab.freedesktop.org/libinput/libinput/-/blob/49eb6cb8ab2695a4ac8883d2be5148472563b8db/src/libinput.h?page=6#L5071-5074), i hope that's fine.

i tested it locally and it worked fine, but i haven't done extensive testing.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -959,6 +959,7 @@ extending outward from the snapped edge.
     <tapButtonMap></tapButtonMap>
     <tapAndDrag></tapAndDrag>
     <dragLock>sticky</dragLock>
+    <threeFingerDrag></threeFingerDrag>
     <middleEmulation></middleEmulation>
     <disableWhileTyping></disableWhileTyping>
     <clickMethod></clickMethod>
@@ -1032,6 +1033,16 @@ extending outward from the snapped edge.
 	release of a finger during tap-and-dragging. Both *yes* and *sticky* enable
 	drag lock, but if *yes* is set, the drag lock expires after a timeout.
 	Default is *sticky*.
+
+*<libinput><device><threeFingerDrag>* [yes|no|3|4]
+	Enable or disable the three-finger drag feature. When enabled, three
+	fingers down will result in a button down event and subsequent finger
+	motions triggers a drag.
+
+	The available options are:
+	- *no* - Disable three-finger drag.
+	- *yes* | *3* - Enable three-finger drag for 3 fingers.
+	- *4* - Enable three-finger drag for 4 fingers.
 
 *<libinput><device><middleEmulation>* [yes|no]
 	Enable or disable middle button emulation for this category. Middle
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -788,6 +788,27 @@ fill_libinput_category(char *nodename, char *content, struct parser_state *state
 		state->current_libinput_category->drag_lock = ret
 			? LIBINPUT_CONFIG_DRAG_LOCK_ENABLED
 			: LIBINPUT_CONFIG_DRAG_LOCK_DISABLED;
+	} else if (!strcasecmp(nodename, "threeFingerDrag")) {
+#if HAVE_LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG
+		if (!strcmp(content, "3")) {
+			state->current_libinput_category->three_finger_drag =
+				LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG;
+		} else if (!strcmp(content, "4")) {
+			state->current_libinput_category->three_finger_drag =
+				LIBINPUT_CONFIG_3FG_DRAG_ENABLED_4FG;
+		} else {
+			int ret = parse_bool(content, -1);
+			if (ret < 0) {
+				return;
+			}
+			state->current_libinput_category->three_finger_drag = ret
+				? LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG
+				: LIBINPUT_CONFIG_3FG_DRAG_DISABLED;
+		}
+#else
+		wlr_log(WLR_ERROR, "<threeFingerDrag> is only"
+			" supported in libinput >= 1.28");
+#endif
 	} else if (!strcasecmp(nodename, "accelProfile")) {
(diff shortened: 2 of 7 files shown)


Consolatis:
Thanks for the PR.

From the FreeBSD CI run:
```
../src/seat.c: In function 'configure_libinput':
../src/seat.c:172:17: error: implicit declaration of function 'libinput_device_config_3fg_drag_set_enabled'; did you mean 'libinput_device_config_dwtp_set_enabled'? [-Werror=implicit-function-declaration]
  172 |                 libinput_device_config_3fg_drag_set_enabled(
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                 libinput_device_config_dwtp_set_enabled
```

I assume this is a relatively new feature in libinput so if we want to support this we'd need to increase the libinput version requirement.

m4rch3n1ng:
yeah, the version with that feature (1.28) was only released ~2 months ago (2025-03-24).

it seems freebsd does not have that version in their repos yet though (https://cgit.freebsd.org/ports/tree/x11/libinput)

Consolatis:
@johanmalm @tokyo4j any thoughts on how to deal with this?

I have a feeling raising the required version is too soon.
We could do a version check in meson.build and then add a feature flag, similar to what we are doing with librsvg and libsfdo. And then put the new code path into a `#if HAVE_LIBINPUT_THREE_FINGER_DRAG .. #endif`.

johanmalm:
> I have a feeling raising the required version is too soon.

Agree. We don't want to make life difficult for non-rolling-release distros. Looks like Debian bookworm is currently on 1.22. 

> We could do a version check in meson.build and then add a feature flag, similar to what we are doing with librsvg and libsfdo. And then put the new code path into a `#if HAVE_LIBINPUT_THREE_FINGER_DRAG .. #endif`.

Good plan. Feels like it's enough to add that `#if` in `configure_libinput()`

tokyo4j:
> We could do a version check in meson.build and then add a feature flag, similar to what we are doing with librsvg and libsfdo. And then put the new code path into a `#if HAVE_LIBINPUT_THREE_FINGER_DRAG .. #endif`.

I agree.

I'll open a PR to address "drag-lock enabled sticky" in #2729 which will also use similar libinput version check so that @m4rch3n1ng can see it as a reference.

docs/labwc-config.5.scd
```diff
@@ -947,6 +947,7 @@ extending outward from the snapped edge.
     <tapButtonMap></tapButtonMap>
     <tapAndDrag></tapAndDrag>
     <dragLock></dragLock>
+	<threeFingerDrag></threeFingerDrag>
```
tokyo4j: Inside triple backticks, spaces should be used instead of a tab.

tokyo4j:
~Opened #2803. You can copy my modification to `meson.build` there and replace the symbol with `LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG`, or you can rebase on the master branch after my PR is merged.~

You can now:
- Rebase you branch on the master branch
- Add `LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG` to the the array (`['LIBINPUT_CONFIG_DRAG_LOCK_ENABLED_STICKY']`) in `meson.build`
- Wrap some of the new lines with `#if HAVE_LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG ... #endif` so that it can be compiled with older libinput versions
- Show an error message if `<threeFingerDrag>` is provided from the user with older libinput versions.

src/config/rcxml.c
```diff
@@ -778,6 +778,19 @@ fill_libinput_category(char *nodename, char *content, struct parser_state *state
...
+				LIBINPUT_CONFIG_3FG_DRAG_DISABLED;
+		} else if (!strcmp(content, "yes") || !strcmp(content, "3")) {
+			state->current_libinput_category->three_finger_drag =
+				LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG;
+		} else if (!strcmp(content, "4")) {
+			state->current_libinput_category->three_finger_drag =
+				LIBINPUT_CONFIG_3FG_DRAG_ENABLED_4FG;
+		} else {
+			wlr_log(WLR_ERROR, "invalid threeFingerDrag");
+		}
```
tokyo4j: You can compare `content` with "3" and "4" and then use `parse_bool()`.

docs/labwc-config.5.scd
m4rch3n1ng: done

src/config/rcxml.c
m4rch3n1ng: done

m4rch3n1ng:
rebased on master now and added the `HAVE_LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG` directives, i think it should work now.

(note: i accidentally auto-formatted a file i shouldn't have and forgot to restore that, sorry for wasting your ci minutes ><)

tokyo4j:
Tested and it worked well. Thanks for your contribution!

concussious:
Hi, how do I use this? I want to do something like:
```
<mousebind button="threeFingerDrag" action="left"><action name="NextWindow" /></mousebind>
```
But I didn't see anything in the doc that seemed like what I was looking for.

m4rch3n1ng:
what you are looking for is gesture bindings, which, as far as i can tell, haven't yet been implemented: #1156.

the feature that this pr implements has very little to do with swipe gestures, it just makes it so that you can select text / drag windows without clicking first by just dragging with three fingers. neat feature, but not what you are looking for i imagine.

PR #2797, opened 2025-06-06
view.c: free icon name on view destroy

Consolatis:
Discovered via `-Db_sanitize=address` with `Ctrl-C` of a nested instance with foot still running.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2595,6 +2595,7 @@ view_destroy(struct view *view)
 	if (view->icon.buffers.data) {
 		view_set_icon(view, NULL, NULL);
 	}
+	zfree(view->icon.name);
 
 	/*
 	 * The layer-shell top-layer is disabled when an application is running


Consolatis:
The CI leak test should have caught that but didn't: https://github.com/labwc/labwc/actions/runs/15498221139/job/43639945054?pr=2794#step:18:794

The reason is that foot didn't detect the toplevel icon protocol which doesn't make sense to me.
`warn: wayland.c:1766: compositor does not implement the XDG toplevel icon protocol`

johanmalm:
Thanks. 

> The reason is that foot didn't detect the toplevel icon protocol which doesn't make sense to me. `warn: wayland.c:1766: compositor does not implement the XDG toplevel icon protocol`

Interestingly I get the same foot message on sway (with latest sway master)

tokyo4j:
> The reason is that foot didn't detect the toplevel icon protocol which doesn't make sense to me. `warn: wayland.c:1766: compositor does not implement the XDG toplevel icon protocol`

My PR to fix it was merged last week: https://codeberg.org/dnkl/foot/pulls/2113

johanmalm:
> > The reason is that foot didn't detect the toplevel icon protocol which doesn't make sense to me. `warn: wayland.c:1766: compositor does not implement the XDG toplevel icon protocol`
> 
> My PR to fix it was merged last week: https://codeberg.org/dnkl/foot/pulls/2113

Beautiful :smile:

I just ran with foot-master and it works now.

Consolatis:
> > The reason is that foot didn't detect the toplevel icon protocol which doesn't make sense to me. `warn: wayland.c:1766: compositor does not implement the XDG toplevel icon protocol`
> 
> My PR to fix it was merged last week: https://codeberg.org/dnkl/foot/pulls/2113

Well that explains it. I did my test for the icon PR with a recent foot master branch but its likely not released and thus not in Arch yet so the runtime leak test didn't actually test the window icon name set by foot.

Issue #2798, opened 2025-06-07
View creates SSD before view_init() can initialize the signals


kode54:
### Problem description

Server side decoration is crashing on NULL wl_list in the signals, due to SSD being hit before view_init, as mentioned on IRC. It has the following backtrace log:

```
#0  0x00007f64530c7109 in wl_list_insert (list=0x0, elm=elm@entry=0x55ec0b71c488) at ../wayland-1.23.1/src/wayland-util.c:48
48              elm->next = list->next;
[Current thread is 1 (Thread 0x7f644ed89600 (LWP 717394))]
Function(s) ^std::(move|forward|as_const|(__)?addressof) will be skipped when stepping.
Function(s) ^std::(shared|unique)_ptr<.*>::(get|operator) will be skipped when stepping.
Function(s) ^std::(basic_string|vector|array|deque|(forward_)?list|(unordered_|flat_)?(multi)?(map|set)|span)<.*>::(c?r?(begin|end)|front|back|data|size|empty) will be skipped when stepping.
Function(s) ^std::(basic_string|vector|array|deque|span)<.*>::operator.] will be skipped when stepping.
(gdb) bt
#0  0x00007f64530c7109 in wl_list_insert (list=0x0, elm=elm@entry=0x55ec0b71c488) at ../wayland-1.23.1/src/wayland-util.c:48
#1  0x000055ebf2f3293e in wl_signal_add (signal=0x55ec0ae9d698, listener=0x55ec0b71c488) at /usr/include/wayland-server-core.h:476
#2  scaled_icon_buffer_set_view (self=0x55ec0b71c440, view=0x55ec0ae9d3c0) at ../labwc/src/common/scaled-icon-buffer.c:228
#3  0x000055ebf2f48d82 in add_scene_button.isra.0 (part_list=part_list@entry=0x55ec0b833270, type=LAB_SSD_BUTTON_WINDOW_ICON, parent=0x55ec0b71c010,
    parent@entry=0x55ec0b71bab0, imgs=0x7ffeb514e010, x=x@entry=9, y=y@entry=3, view=0x55ec0ae9d3c0) at ../labwc/src/ssd/ssd-part.c:116
#4  0x000055ebf2f27a90 in ssd_titlebar_create (ssd=0x55ec0b833200) at ../labwc/src/ssd/ssd-titlebar.c:79
#5  ssd_create (view=0x55ec0ae9d3c0, active=false) at ../labwc/src/ssd/ssd.c:249
[...]

PR #2799, opened 2025-06-07
Ensure the view signals are initialized early

Consolatis:
This prevents a crash observed by the scaled-icon-buffer listening to the `view->events.set_icons` signal before it is initialized. For this to happen, the application needs to use the kde decoration protocol and set it on the xdg_surface before creating the xdg_toplevel.

Fixes:
- #2798


Consolatis:
Confirmed fixing the issue via IRC.

Merging due to it being a crash fix.

johanmalm:
Thanks.

tokyo4j:
Thanks!

PR #2801, opened 2025-06-07
view: simplify freeing of view->icon

tokyo4j:
fixup for #2797


johanmalm:
Ah yeah, that's nicer.

LGTM

PR #2803, opened 2025-06-07
libinput: support <dragLock>sticky<dragLock> and enable it by default

tokyo4j:
Ref: [swaywm/sway#8674](https://github.com/swaywm/sway/commit/bbadf9b8b10d171a6d5196da7716ea50ee7a6062)

Quote from my added documentation:
> Drag lock ignores a temporary release of a finger during tap-and-dragging. Both *yes* and *sticky* enable drag lock, but if *yes* is set, the drag lock expires after a timeout. Default is *sticky*.


johanmalm:
The meson `has_header_symbol()` method is nice. I did not know about that one.

Suggest adding a warning in `rcxml.c` if the user tries to set sticky but it is not supported.

tokyo4j:
> The meson has_header_symbol() method is nice. I did not know about that one.

Me neither. It was copied from sway's PR.

> Suggest adding a warning in rcxml.c if the user tries to set sticky but it is not supported.

Nice idea. I added an error message: `<dragLock>sticky</dragLock> is only supported in libinput >= 1.27`.

johanmalm:
LGTM

PR #2807, opened 2025-06-08
refactor: move tablet handlers to tablet tool

jp7677:
This is _just_ a refactoring which moves the tablet tool axis handlers from tablet to tablet-tool which feels slightly more correctly and imho easier to follow. 

All preparations are done in the first commit so that the second commit, which diffs horrible, is really just a big move. 
From the first commit: 
Move tablet axis members to tablet tool: It doesn't matter much since those are about the combination of tablet and tablet tool. That said, this feels slightly more natural. As a consequence we always create a tablet tool and decide indirectly via  `tablet_get_coords()` and the returning surface if mouse emulation should be used or not. Now we can also add a `motion_mode` to the tablet tool which is slightly cleaner.

From a functional perspective everything should behave the same, but I guess this needs some cooking and testing.

Also no drama if we don't do this and stick to the current version.


src/input/tablet-tool.c
```diff
@@ -70,6 +76,17 @@ tablet_tool_create(struct seat *seat,
...
+	tool->motion_mode = LAB_TABLET_MOTION_ABSOLUTE;
+	tool->x = 0.0;
+	tool->y = 0.0;
+	tool->distance = 0.0;
+	tool->pressure = 0.0;
+	tool->tilt_x = 0.0;
+	tool->tilt_y = 0.0;
+	tool->rotation = 0.0;
+	tool->slider = 0.0;
+	tool->wheel_delta = 0.0;
```
Consolatis: These aren't necessary as `znew()` was used to initialize the struct with 0. The only exception might be `LAB_TABLET_MOTION_ABSOLUTE` if its defined as something else than `0`.

Consolatis:
> This is _just_ a refactoring which moves the tablet tool axis handlers from tablet to tablet-tool which feels slightly more correctly and imho easier to follow.

Agree. One thing I wonder is if we should keep the `tablet_init()` calls though which then in turn can call `tablet_tool_init()`. That allows better modularization of the whole "tablet" handling in a way that there is a central entry point which then can initialize / finalize its sub modules.

Also CC @tokyo4j as they mentioned interest in some refactoring of the tablet logic before.

jp7677:
> Agree.

Thanks for already looking at this PR!

> One thing I wonder is if we should keep the `tablet_init()` calls though which then in turn can call `tablet_tool_init()`. That allows better modularization of the whole "tablet" handling in a way that there is a central entry point which then can initialize / finalize its sub modules.

My reasoning to also move the `init`/`finish` functions to `tablet_tool` was because they only setup handlers for `tablet_tool_*` (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_cursor.h#L72), but do nothing else. Other initialization functions (`tablet_create`/`tablet_pad_create`) are called later from `seat.new_input_notify()` or internally from `tablet_tool.handle_tablet_tool_proximity()`. So it is in my opinion not really a submodule relation between tablet and tablet-tool/tablet-pad. 
An alternative thought I had was to merge everything from `tablet`/`tablet-tool` and `tablet-pad` into `tablet` to avoid that `tablet` appears to be a module above tool/pad. Not sure yet..

Edit: I've now put everything into tablet. May be that is a bit too much ;). I could also just merge tablet and tablet-tool, but leave tablet-pad separated.

Edit 2: I've moved tablet-pad back to its separated h/c file, thus only tablet and tablet-tool are merged into tablet.h/.c. Now we are getting somewhere..

src/input/tablet-tool.c
jp7677: Done as a separate commit.

jp7677:
This feels good now. I haven’t seen any issues and the relationship tablet vs tablet-tool is imho less confusing. Review per commit works probably better.

tokyo4j:
Sorry for keeping this PR on hold for a while. I'm not following all the changed lines, but I like this cleanup.

LGTM other than:
- The subject of the third commit should be "... into tablet.c" rather than "... into tablet" (sorry for nitpicking :sweat_smile:)
- I was a bit confused to find there are both `rc.tablet.force_mouse_emulation` and `drawing_tablet_tool.force_mouse_emulation`. I suggest adding a comment above `drawing_tablet_tool.force_mouse_emulation` definition in the header file or renaming it to something else.

jp7677:
> The subject of the third commit should be "... into tablet.c" rather than "... into tablet" (sorry for nitpicking 😅)

All good, I'm all in for the nits ;)

>  I was a bit confused to find there are both `rc.tablet.force_mouse_emulation` and ... 

I went for the comment in the header.

johanmalm:
Thanks.

PR #2808, opened 2025-06-08
cursor: omit pointer cursor shape for tablet tool

jp7677:
Omit cursor notifications from a pointer when a tablet tool (stylus/pen) is in proximity. This is equivalent to `handle_request_set_cursor()` and prevents a resize cursor for out-of-surface scrolling with a tablet tool in recent GTK4 (which uses the cursor shape protocol).

This fixes this issue:

https://github.com/user-attachments/assets/de763a4e-a8d2-4022-ab63-b98dc803a622

(the cursor is from a tablet tool)

This is not a GTK4 issue since this works in Gnome, so really up to us to fix this.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -224,6 +224,16 @@ handle_request_set_shape(struct wl_listener *listener, void *data)
 		return;
 	}
 
+	/*
+	 * Omit cursor notifications from a pointer when a tablet
+	 * tool (stylus/pen) is in proximity.
+	 */
+	if (tablet_tool_has_focused_surface(seat)
+			&& event->device_type
+				!= WLR_CURSOR_SHAPE_MANAGER_V1_DEVICE_TYPE_TABLET_TOOL) {
+		return;
+	}
+
 	wlr_log(WLR_DEBUG, "set xcursor to shape %s", shape_name);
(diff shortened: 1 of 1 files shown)


jp7677:
mmh, with this fix for the out-of-surface scrolling cursor shape, I'm seeing https://gitlab.gnome.org/GNOME/mutter/-/issues/4099 on labwc, so could be that this change uncovers an actual GTK4 issue. 

Edit: I've compared mutter and labwc with this PR applied and the GTK4 issue is exactly the same based on my testing. ~I've updated the mutter bug report and asked for it to be moved to the GTK4 repository.~ Moved to https://gitlab.gnome.org/GNOME/gtk/-/issues/7593

jp7677:
As stated, this exposes a GTK4 issue, but I'm still in favor of fixing the out-of-surface bug (which seems to me the correct thing to do) and hope that the cursor shape issue gets resolved in GTK4. Until that there is mouse emulation in labwc for tablet input.

jp7677:
> As stated, this exposes a GTK4 issue

PR which fixes this in GTK4: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8674 (merged)

So this PR here is good to land imho.

johanmalm:
Thanks

PR #2809, opened 2025-06-09
Replace `multi_rect` and `scaled_rect_buffer` with `lab_scene_rect`

tokyo4j:
Supersedes #2241.

Rather than having two different interfaces for rectangles with 3 borders and ones with 1 border, this PR unifies them into `lab_scene_rect` that can accept arbitrary numbers of borders.

I decided to move its implementation to `lab-scene-rect.c`.


src/common/lab-scene-rect.c
```diff
@@ -0,0 +1,112 @@
...
+}
+
+struct lab_scene_rect *
+lab_scene_rect_create(struct wlr_scene_tree *parent,
+		struct lab_scene_rect_options *opts)
+{
+	struct lab_scene_rect *rect = znew(*rect);
+	rect->border_width = opts->border_width;
+	rect->nr_borders = opts->nr_borders;
+	rect->borders = calloc(opts->nr_borders, sizeof(struct border_scene));
```
Consolatis: Suggest `sizeof(rect->borders[0])` here to prevent memory access errors / random memory writes when we ever decide to change the type of `rect->borders`.

src/common/lab-scene-rect.c
```diff
@@ -0,0 +1,112 @@
...
+	 * +-+-----+-+   |
+	 * +---------+   v
+	 */
+
+	if ((width < border_width * 2) || (height < border_width * 2)) {
+		wlr_scene_node_set_enabled(&border->tree->node, false);
+		return;
+	} else {
+		wlr_scene_node_set_enabled(&border->tree->node, true);
+	}
```
Consolatis: The `else` branch isn't necessary here as we early return in the first branch.

src/common/lab-scene-rect.c
```diff
@@ -0,0 +1,112 @@
...
+		width - border_width, border_width);
+
+	wlr_scene_rect_set_size(border->top,
+		width, border_width);
+	wlr_scene_rect_set_size(border->bottom,
+		width, border_width);
+	wlr_scene_rect_set_size(border->left,
+		border_width, height - border_width * 2);
+	wlr_scene_rect_set_size(border->right,
+		border_width, height - border_width * 2);
```
Consolatis: We do have a 80 column width recommendation but can exceed it up to 100 columns. IMHO this looks way too packed and decreases readability. So I suggest to use single lines whenever a single line is <= 100 (or at least 80) columns here.

src/common/lab-scene-rect.c
Consolatis: Also, this should be `znew_n()` rather than `calloc()`. Somthing like 
```c
rect->borders = znew_n(rect->borders[0], opts->nr_borders);
// or
rect->borders = znew_n(*rect->borders, opts->nr_borders);
```

tokyo4j: Updated to `znew_n(rect->borders[0], opts->nr_borders)`. I didn't know that we have `znew_n()`!

src/common/lab-scene-rect.c
tokyo4j: Removed the `else` branch.

src/common/lab-scene-rect.c
tokyo4j: Updated as suggested.

tokyo4j:
Thanks for reviewing!

PR #2810, opened 2025-06-09
CI: use wlroots0.19 package for Arch Linux

tokyo4j:
CC @Consolatis 

EDIT: also updated to use `wlroots019` for FreeBSD


Consolatis:
Thanks :)

PR #2811, opened 2025-06-09
xwayland: always offer focus in Globally Active case

jlindgren90:
In 9e3785f8cd7a, a heuristic was added to assume that NORMAL and DIALOG window types were always focusable. (This was before we had the "offer focus" mechanism in place.)

However, we should still call `wlr_xwayland_surface_offer_focus()` for these views, in case they actually don't want focus. (This is uncommon but has recently been seen with WeChat popups, which have both NORMAL and UTILITY type.)

To make this possible, refine `view_wants_focus()` to return either LIKELY or UNLIKELY for Globally Active input windows. This decouples the question of "should we try to focus this view" from the actual mechanism used to do so.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -75,13 +75,18 @@ enum view_wants_focus {
 	/* View wants focus */
 	VIEW_WANTS_FOCUS_ALWAYS,
 	/*
-	 * View should be offered focus and may accept or decline
-	 * (a.k.a. ICCCM Globally Active input model). Labwc generally
-	 * avoids focusing these views automatically (e.g. when another
-	 * view on top is closed) but they may be focused by user action
-	 * (e.g. mouse click).
+	 * The following values apply only to XWayland views using the
+	 * Globally Active input model per the ICCCM. These views are
+	 * offered focus and will voluntarily accept or decline it.
+	 *
+	 * In some cases, labwc needs to decide in advance whether to
+	 * focus the view. For this purpose, these views are classified
+	 * (by a heuristic) as likely or unlikely to want focus. However,
+	 * it is still ultimately up to the client whether the view gets
+	 * focus or not.
 	 */
-	VIEW_WANTS_FOCUS_OFFER,
+	VIEW_WANTS_FOCUS_LIKELY,
+	VIEW_WANTS_FOCUS_UNLIKELY,
 };
 
 /*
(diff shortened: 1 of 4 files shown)


jlindgren90:
The more people who can test this before merging, the better - to avoid regressions.

Consolatis:
LGTM so feel free to merge at your own judgement.

johanmalm:
@jlindgren90 Thanks for navigating this. Seems good to me too.

@cillian64 @spl237 - Would you have the capacity to do some testing with XWayland clients. We're looking for any changes to input focus or any other odd behaviour. Seem to remember that you've got a few X11 apps based on various toolkits :smile:

jlindgren90:
Let's go ahead with this. We know that it fixes the WeChat popup issue (the "likely" case) and I've tested it with qmpanel in XWayland mode (the "unlikely" case) which was tricky to get right in the past.

cillian64:
> @cillian64 @spl237 - Would you have the capacity to do some testing with XWayland clients. We're looking for any changes to input focus or any other odd behaviour. Seem to remember that you've got a few X11 apps based on various toolkits 😄

I'm afraid I'm pretty booked up for the next week or two.  @spl237 if you wanted to test this I've got a wlroots 0.19 branch with rpi customisations I can point you at

jp7677:
Sorry to bring the bad news , but this PR/Commit effects CLion (and probably other JetBrains IDE's) in that some tool windows (Settings icon -> View mode or Settings icon -> Keymap) won't show up anymore.

jlindgren90:
> Sorry to bring the bad news , but this PR/Commit effects CLion (and probably other JetBrains IDE's) in that some tool windows (Settings icon -> View mode or Settings icon -> Keymap) won't show up anymore.

Hmm. The windows don't show up in labwc at all now? Or they just don't show up in the Alt-Tab switcher?

jp7677:
> > Sorry to bring the bad news , but this PR/Commit effects CLion (and probably other JetBrains IDE's) in that some tool windows (Settings icon -> View mode or Settings icon -> Keymap) won't show up anymore.
> 
> Hmm. The windows don't show up in labwc at all now? Or they just don't show up in the Alt-Tab switcher?

They don't show up at all, so like nothing happens when clicking the menu entry. That said those tool windows look quite special, see this screenshot from Gnome:

image

Those windows don't show up in the window switcher, but they also don't show up in Gnome or Labwc before that commit, so that seems fine.

jlindgren90:
It appears that when clicking on the gear/settings icon in the top toolbar, the main CLion window loses focus now. Before this PR/commit, and under Openbox, it keeps focus. I'm not sure why -- need to debug it further.

PR #2819, opened 2025-06-11
CLion focus fixes

jlindgren90:
#2811 fixed an issue with WeChat but unfortunately exposed multiple focus-related bugs, affecting at least JetBrains CLion (probably other apps too).

These bugs were introduced by the addition of `view_offer_focus()` in #1142 but had remained masked due to treating NORMAL and DIALOG type windows as always wanting focus.

The general approach of #1142 still feels right to me, but it introduced some asynchronous behavior (labwc offers focus, client accepts it some time later) which is more of a headache to deal with, and broke some assumptions in the code.

Please see the individual commit messages for details.

### DIFF
--- a/src/xwayland-unmanaged.c
+++ b/src/xwayland-unmanaged.c
@@ -89,11 +89,29 @@ focus_next_surface(struct server *server, struct wlr_xwayland_surface *xsurface)
 	}
 
 	/*
-	 * If we don't find a surface to focus fall back
-	 * to the topmost mapped view. This fixes dmenu
-	 * not giving focus back when closed with ESC.
+	 * Unmanaged surfaces do not clear the active view when mapped.
+	 * Therefore, we can simply give the focus back to the active
+	 * view when the last unmanaged surface is unmapped.
+	 *
+	 * Also note that resetting the focus here is only on the
+	 * compositor side. On the xwayland server side, focus is never
+	 * given to unmanaged surfaces to begin with - keyboard grabs
+	 * are used instead.
+	 *
+	 * In the case of Globally Active input windows, calling
+	 * view_offer_focus() at this point is both unnecessary and
+	 * insufficient, since it doesn't update the seat focus
+	 * immediately and ultimately results in a loss of focus.
+	 *
+	 * For the above reasons, we avoid calling desktop_focus_view()
+	 * here and instead call seat_focus_surface() directly.
+	 *
+	 * If modifying this logic, please test for regressions with
+	 * menus/tooltips in JetBrains CLion or similar.
 	 */
-	desktop_focus_topmost_view(server);
+	if (server->active_view) {
+		seat_focus_surface(&server->seat, server->active_view->surface);
+	}
 }
 
 static void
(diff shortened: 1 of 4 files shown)


jlindgren90:
@jp7677 please test if you get a chance

johanmalm:
@jlindgren90 You're amazing finding the time to get your head around this. My brain is mash tonight, but will try to understand/follow logic tomorrow.

jlindgren90:
> @jlindgren90 You're amazing finding the time to get your head around this. My brain is mash tonight, but will try to understand/follow logic tomorrow.

Haha, I had the day off and the kids are away. Otherwise it would not have happened :)

Force-pushed with a couple of comment updates only.

src/xwayland-unmanaged.c
```diff
@@ -209,9 +227,12 @@ handle_request_activate(struct wl_listener *listener, void *data)
...
 	 * Validate that the unmanaged surface trying to grab focus is actually
-	 * a child of the topmost mapped view before granting the request.
+	 * a child of the active view before granting the request.
+	 *
+	 * FIXME: this logic appears to always allow activating an unmanaged
+	 * surface if an xdg-shell view is active - which seems inconsistent.
 	 */
-	struct view *view = desktop_topmost_focusable_view(server);
+	struct view *view = server->active_view;
 	if (view && view->type == LAB_XWAYLAND_VIEW) {
```
Consolatis: Looks like that could be fixed with another branch like
```c
if (view && view->type == LAB_XWAYLAND_VIEW) {
	[..]
} else if (view) {
	/* Don't allow unmanaged windows to steal focus from xdg views */
	return;
}
seat_focus_surface(seat, xsurface->surface);
```

Consolatis:
Small nitpick above, other than that code LGTM but didn't test.

> The general approach of #1142 still feels right to me

Yep, same. The various X11 focus models are a complete mess but the more we are able to stick to the specs the better and less chance that we pile workaround on top of workaround and still get reports that X11 application $x broke.

src/xwayland-unmanaged.c
jlindgren90: I think there are still inconsistencies even with that. It's a bit of a can of worms which is why I didn't want to do more in this PR than add a FIXME comment.

Consolatis: In that case I suggest specifying the weird edge cases / inconsistencies in the comment so we don't end up forgetting about them.

jlindgren90: I started a new discussion thread about this to capture some of my concerns/questions:

- #2821

Given that, is it okay to leave the FIXME as-is for this PR?

Consolatis: Could we add the link https://github.com/labwc/labwc/discussions/2821 to the comment?

I just want to make sure that if we don't come to a conclusion and fix soon and then forget about it, the next person seeing this will just put a branch in there, similar to what I proposed above. That would then for example also ignore the `grab_focus` handler but assume that the issue is fixed and thus remove the FIXME comment.

jlindgren90: Good idea

jlindgren90: Added a link to the discussion

jp7677:
> @jp7677 please test if you get a chance

I did a brief test and CLion popups/menus do work again as intended, very nice! 
Thanks a lot for having looked into this!

johanmalm:
LGTM. Thanks.

Issue #2722, opened 2025-05-15
focus fight between compositor and client


domo141:
### Problem description

Compositor gives focus to a view, while (the same?) application wants to give it back to another

Apparently the relevant configuration options are:
`<followMouse>yes</followMouse>`
`<followMouseRequiresMovement>yes</followMouseRequiresMovement>`
(I also have `<raiseOnFocus>no</raiseOnFocus>`)

Tried env:
- Fedora 42
- labwc 0.8.4 (*) 
- scribus 1.6.3 (X application)

Scribus starts ok. It shows 2 windows, the main application window and on top of that a
popup file requestor.

When moving mouse from that requestor over to the main window, compositor freezes.

After quite a few of tries (of all kinds) my understanding of the issue is that compositor
gives focus to that main window of scribus -- and scribus immeidately requests focus
to that file requestor -- and this keeps looping frantically.

When I change that `requresMovement` to `no`, then this focus changing is more apparent,
as (only) the movements of a mouse makes the focus move happen (back and forth)

(*) does not happen with labwc 0.8.3 (current Fedora 42 package), when I made (not so educated)
guesses what made this broke, it could be that the commit de49054d75 did it...

### Steps to reproduce

1. configure followMouse yes, followMouseRequiresMovement no
2. launch labwc 0.8.4 - preferably nested instance
3. run scribus 
4. if focus on top of the file requester, move mouse on top of the "main" scribus window

### labwc build source

Release

### labwc version

0.8.4

### labwc environment

Nested on Wayland

jlindgren90:
There are some focus related fixes in wlroots 0.19 that may improve/fix this.

tokyo4j:
This can be fixed by not calling `cursor_update_focus()` if the stacking order is not changed just like before #2638.

But in the first place, I guess we should not give keyboard focus to a parent window which has child windows. This seems to be a common sense in Openbox/KWin/Mutter.

jlindgren90:
> But in the first place, I guess we should not give keyboard focus to a parent window which has child windows.

I think it's not quite this simple, it depends on whether the child window is a "modal" dialog or not. Possibly the `view_wants_focus` tri-state hint could help to determine the correct behavior. Or, we could just try giving the parent window focus one time (i.e. when the cursor first enters the window) but not continue to "fight" with the application if it gives focus back to a modal dialog immediately.

tokyo4j:
Sorry for late reply.

[...]

PR #2822, opened 2025-06-12
desktop: enforce followMouse focus policy more strictly

jlindgren90:
With `followMouse=yes` in the config, a client activation request was still allowed to focus a view that wasn't under the mouse cursor, but it would then be immediately unfocused again. With some clients using modal dialogs (e.g. Scribus), this could lead to an infinite cycle of the client and labwc continually switching the focus back and forth.

To break the cycle, only allow focus updates to be initiated from the cursor logic when `followMouse=yes`.

Fixes: #2722


jlindgren90:
Lightly tested, but more testing would be appreciated

jlindgren90:
Also a question: if `followMouseRequiresMovement=yes`, should a new window that pops up under the cursor still automatically gain focus? I am not sure of the desired behavior in that case.

jlindgren90:
v2: allow focusing a new/raised view appearing under the cursor

This needs some work/debugging still. With `followMouseRequiresMovement=yes`, if Alt-Tabbing to Scribus (with the New Document dialog open), then neither the main window nor the dialog gets focused. I'm not sure exactly why.

tokyo4j:
> This needs some work/debugging still. With followMouseRequiresMovement=yes, if Alt-Tabbing to Scribus (with the New Document dialog open), then neither the main window nor the dialog gets focused. I'm not sure exactly why.

In addition, with `followMouse=yes` and `followMouseRequiresMovement=yes`, new windows are no longer automatically focused when the cursor is somewhere else. This (and the Alt-Tab issue you described) is because `desktop_focus_view()` does nothing when there's no view on the cursor.

However, changes like below doesn't work well either.

```diff
diff --git a/src/desktop.c b/src/desktop.c
index 70a376fa..f12253f0 100644
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -108,7 +108,9 @@ _desktop_focus_view(struct view *view, bool raise, bool for_cursor_update)
 	 */
 	if (rc.focus_follow_mouse && !for_cursor_update) {
 		struct cursor_context ctx = get_cursor_context(view->server);
-		if (ctx.view && view_get_root(ctx.view) == view_get_root(view)) {
+		if (!ctx.view) {
+			set_or_offer_focus(view);
+		} else if (ctx.view && view_get_root(ctx.view) == view_get_root(view)) {
 			set_or_offer_focus(ctx.view);
 		}
 	} else {
```

...because it overrides the selection from window switcher; when we switch from window A to window B with window switcher while the cursor is on window A, we expect B to be focused (at least when `followMouseRequiresMovement=yes`), but window A is focused instead with the patch above.

I don't know how we can fix #2722 with your approach that `desktop_focus_view()` doesn't focus the view at the cursor with `followMouse=yes` & `followMouseRequiresMovement=no` when it's called from window activation handler. Honestly I'd like to leave the current logic as it is if possible so that we can keep the simple assumption that we can focus on the view at the cursor from anywhere, like we are doing in `_cursor_update_focus()`.

If we can fix #2722 by not focusing on a parent window which has a modal window like I described in https://github.com/labwc/labwc/issues/2722#issuecomment-2885969492, I'm all for it. It also follows the behaviors of other compositors and WMs including KWin, Mutter and Openbox, though I don't know how they handle follow-mouse.

jlindgren90:
> In addition, with `followMouse=yes` and `followMouseRequiresMovement=yes`, new windows are no longer automatically focused when the cursor is somewhere else.

I actually thought that was the desired behavior with followMouse. But I don't use it personally so my opinion is uneducated here.

There are other issues still with this PR even in the latest version. New xdg-shell views aren't getting focused even if under the cursor. So I'm fine with abandoning it in favor of a different approach.

jlindgren90:
> New xdg-shell views aren't getting focused even if under the cursor.

I fixed this part, it wasn't too bad. Of course it doesn't address the larger concerns with the approach.

jlindgren90:
Closing in favor of #2828.

PR #2824, opened 2025-06-12
PoC: view: Add simple window animations

any1:
This is a proof-of-concept for simple window animations. As such, this is not meant to be a complete implementation so a few things are missing, such as proper cleanup when the window is destroyed while it's being animated. I am also not sure if this is the right place to start the animations. Suggestions would be welcome!

The following modifications to wlroots are required to build and run this:  https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5082

cc: @cillian64

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -201,6 +201,9 @@ struct view {
 	struct wlr_scene_tree *scene_tree;
 	struct wlr_scene_tree *content_tree;
 
+	struct wlr_scene_buffer *animation;
+	uint64_t animation_start_time;
+
 	bool mapped;
 	bool been_mapped;
 	bool ssd_enabled;
@@ -278,6 +281,7 @@ struct view {
 	struct wl_listener request_maximize;
 	struct wl_listener request_fullscreen;
 	struct wl_listener set_title;
+	struct wl_listener output_commit;
 
 	struct foreign_toplevel *foreign_toplevel;
 
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -2,11 +2,92 @@
 /* view-impl-common.c: common code for shell view->impl functions */
 #include <stdio.h>
 #include <strings.h>
+#include <math.h>
+#include <assert.h>
+#include <wlr/types/wlr_output.h>
 #include "foreign-toplevel.h"
 #include "labwc.h"
 #include "view.h"
 #include "view-impl-common.h"
 #include "window-rules.h"
+#include "workspaces.h"
+
+static inline uint64_t
+timespec_to_us(const struct timespec *ts)
+{
+	return (uint64_t)ts->tv_sec * UINT64_C(1000000) +
+		(uint64_t)ts->tv_nsec / UINT64_C(1000);
+}
+
+static inline uint64_t
+gettime_us(void)
+{
+	struct timespec ts = { 0 };
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	return timespec_to_us(&ts);
+}
+
+static void
+adjust_scaled_position(struct view *view, double scale)
+{
+	int x = view->current.x;
+	int y = view->current.y;
+	double width = view->current.width;
+	double height = view->current.height;
+
+	int x_pos = x + round(width * (1.0 - scale) / 2.0);
+	int y_pos = y + round(height * (1.0 - scale) / 2.0);
+
+	wlr_scene_node_set_position(&view->scene_tree->node, x_pos, y_pos);
+}
+
+static void
+view_run_animation(struct view *view, uint64_t now)
+{
+	uint64_t elapsed_us = now - view->animation_start_time;
+	double t = (double)elapsed_us * 1e-6;
+	double duration = 0.2; // s
+	double scale = fmax(t / duration, 0.01);
+
+	if (scale < 1.0) {
+		wlr_scene_node_set_scale(&view->scene_tree->node, scale);
+		adjust_scaled_position(view, scale);
+	} else {
+		int x = view->current.x;
+		int y = view->current.y;
+		wlr_scene_node_set_position(&view->scene_tree->node, x, y);
+		wlr_scene_node_set_scale(&view->scene_tree->node, 1.0);
+		wl_list_remove(&view->output_commit.link);
+		wl_list_init(&view->output_commit.link);
+		view->animation_start_time = 0;
+	}
+}
+
+static void
+handle_output_commit(struct wl_listener *listener, void *data)
+{
+	struct wlr_output_event_commit *event = data;
+	struct view *view = wl_container_of(listener, view, output_commit);
+
+	if (!(event->state->committed & WLR_OUTPUT_STATE_BUFFER)) {
+		return;
+	}
+
+	view_run_animation(view, gettime_us());
+
+	wlr_output_update_needs_frame(event->output);
+}
+
+static void
+view_start_animation(struct view *view)
+{
+	view->output_commit.notify = handle_output_commit;
+	wl_signal_add(&view->output->wlr_output->events.commit, &view->output_commit);
+
+	uint64_t now = gettime_us();
+	view->animation_start_time = now;
+	view_run_animation(view, now);
+}
 
 void
 view_impl_map(struct view *view)
@@ -16,6 +97,7 @@ view_impl_map(struct view *view)
 	view_update_app_id(view);
 	if (!view->been_mapped) {
 		window_rules_apply(view, LAB_WINDOW_RULE_EVENT_ON_FIRST_MAP);
+		view_start_animation(view);
 	}
 
 	/*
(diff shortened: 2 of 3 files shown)


Consolatis:
Thanks for the PR.

I have a feeling that I am missing some context here.

tokyo4j:
This is very interesting and I really enjoy seeing this kind of experiment, but I'm strongly opposed to bringing animations to labwc. Maybe it's just that I'm dumb, but I think animations makes state transitions harder and is easy to cause unexpected bugs. If someone can maintain a fork of labwc that adds animations to labwc just like swayfx is doing for sway, it's totally fine.

cillian64:
> I have a feeling that I am missing some context here.

Sorry, maybe I should have explained beforehand.  This was instigated by us - one of the last things we feel we're missing in the labwc-based rpi-os desktop compared to wayfire is some basic animations (window close/open/minimize/restore).  We might end up default-disabling them on older Pis, but at least the 4 and 5 are more than capable of handling this.

I think @spl237 discussed this with some labwc devs about a year ago, IIRC the median response was "possibly, if they are lightweight/simple and have zero performance impact when disabled".  So we commissioned @any1 to investigate a proof of concept to see what they might look like, so we have something more concrete to discuss.

> This is very interesting and I really enjoy seeing this kind of experiment, but I'm strongly opposed to bringing animations to labwc. Maybe it's just that I'm dumb, but I think animations makes state transitions harder and is easy to cause unexpected bugs. If someone can maintain a fork of labwc that adds animations to labwc just like swayfx is doing for sway, it's totally fine.

I totally agree that even if there's no performance impact when disabled, there's still a downside of added code complexity and that is really important to consider.  If labwc consensus is that this wouldn't fit with the project, I think we would consider maintaining animations as downstream patches for rpi-os.

Consolatis:
> one of the last things we feel we're missing in the labwc-based rpi-os desktop compared to wayfire is some basic animations (window close/open/minimize/restore).

> I totally agree that even if there's no performance impact when disabled, there's still a downside of added code complexity and that is really important to consider. If labwc consensus is that this wouldn't fit with the project, I think we would consider maintaining animations as downstream patches for rpi-os.

A middle ground would be to move all the animation logic into an own file and base it on `view->events`. We could expose additional events if necessary and the animations being in an own file would make rebasing trivial.

As for having the feature in labwc itself (even if disabled by default).. I am not sure about that. For reference, this compositor started with something like `no icons, no animations, no gradients, no bling`. At this point we do already support icons and shadows and have gradient PRs in the pipeline. So far we managed to keep the resource fallout of those features relatively restricted to the case where they are enabled but things start to get slightly complex on the code maintenance side.

My personal opinion (e.g. not a statement for labwc): If you manage to put everything into an own file, the costs of having downstream accidentally break when rebasing (and thus users ending up here with hard to diagnose issues) might outweigh us simply shipping it without really introducing maintenance overhead (if raspberrypi is willing to maintain the animation implementation itself that is). We could potentially also throw a meson feature option around it. That approach only works if the animations are completely modularized and contained within a single file (or directory) though. Obviously we need some `server->events.new_view` signal and some call to start setting up the initial listener but that is completely fine.

Another issue I have with this in general is the timeline, having a feature within labwc (even if compile time opt-in) depend on downstream patches for wlroots is not something I am particularly enthusiastic about. So even if the wlroots MR is accepted upstream this would only start to work when we start tracking wlroots 0.20.x which will most likely take at least half a year.

johanmalm:
@any1: Thanks for this contribution. 

Setting a direction on 'animations' is an important part of maintainership so here is my view:

1. Cards on the table, I am up for it provided that coupling can be kept low and that we reach a unanimous position amongst the five maintainers - which I think is important as we have so far considered animations to be out of scope. It is important to emphasize that surviving in the long run depends on having people who are interested in the ongoing maintenance, which in turn is more likely if we stand united on direction/principle and keep the code nice to work on (i.e. simple, modular and maintainable).

2. From a personal perspective, `labwc` does everything I set out to do, and my purpose is increasingly in enabling Desktop Environments like Raspberry Pi OS, LXQt and XFCE to create an experience on par with what of equivalent X11 Window
Managers like openbox or Xfwm.

3. The reason I pushed back quite hard on icons, gradients, drop shadows and animations a few years ago was that I knew how much work would be required to create a stable product and I had a strong sense that those feature would get in the way at that point and kill the project. However, my perspective has changed with time, and where we stand today the key consideration for me really is whether things increase or decrease our probability of surviving in the long term.

4. Sustainability in software tools like `labwc` does in the long-run very much depend on carefully managing scope and working within limits.  Sometimes it is really hard to feel that you are doing the right thing in saying 'no' because it is 'just one flag' or just one more thing, or just one work-around. But most people (in my view) who have been around the buoy a few times often regret
letting in too many new features and options. That is why we (on this project) generally take a conservative and slow approach.  BUT having said all that, if people do not want to use the software tool in its vanilla state (`labwc` in this case) because it does not support stuff that they feel they need, then that is obviously no good either. So for example, if Desktop Environment communities think labwc is a good compositor for them, but that the lack of certain contemporary features (like drop-shadows or whatever) makes it unattractive, then we end up missing out on a big ingredient in keeping this project relevant in the long run. Now, if said communities are willing to contribute the code plus be there to support and if it can be implemented in a maintainable way which does not add complexity (beyond that associated with the increase of lines of code - proportionately), then I think it is quite an
attractive proposition.

5. If we  proceed on the basis that the feature set of Openbox is the spirit by which we go forward, it is quite a small step to say that picom/compton functionality is included. This includes simple (unlike those of compiz) animations. It may feel like a hard line to draw, but has not been a problem with drop shadows.

6. Willingness to merge new features/code is often a function of the maintainers' belief that they will get help in looking after it over time. So with devs like @any1, @cillian64 and @spl237 who have compositor street cred, for me that is much easier.

7. So, I concur with the verdict of @Consolatis which really argues (I think!) that if coupling is kept to a minimum; and cohesion is kept high, then maintainability is fine. And on that basis the intent of this PR is of net benefit to `labwc`.

jlindgren90:
I am quite concerned about animations that affect window size/position, because I think it will open up a whole host of bugs if we can no longer rely on the scene-graph nodes being in sync with the logical window size/position. We have a hard enough time already trying to keep the scene-graph in sync with the various clients (with xwayland and xdg-shell clients operating differently and each creating their own challenges).

I might be more okay with simple fade animations (that just add an alpha modifier and don't affect window size/position).

(edit: spelling)

jlindgren90:
>  If labwc consensus is that this wouldn't fit with the project, I think we would consider maintaining animations as downstream patches for rpi-os.

I feel like maintaining patches downstream (or in a long-running PR) might be appropriate initially, as kind of an experiment to make sure the approach is maintainable without getting in the way of other development. Then re-evaluate upstream inclusion a little later, say after a couple more labwc releases?

any1:
Thanks for taking the time to respond to this, everyone.

It seems that the main concern that most people have is that of complicated state tracking. I also think that this may become problematic if we don't think things through with this in mind.

What if, instead of "scale" there was an arbitrary transform property akin to the one in QML? See: https://doc.qt.io/qt-6/qml-qtquick-item.html#transform-prop

The transform would just change the way in which the node is presented. If the transform is reserved for animations and if it is the only property used by animations, then there should be no interference with other state.

johanmalm:
I think that a transformation object is a good idea. 

The bigger question I sensed in the responses was the amount of encapsulation that can be achieved (e.g. init on create and then encapsulate in separate translation unit(s) using the view event signals as coupling). It might just sound good, but not be feasible in reality so a quick evaluation of that concept would be helpful.

In Qt I would say there is not much of a separation line - with advance() merhod and similar part of base classes.

tokyo4j:
I still don't like adding animations to labwc, but its complexity may be acceptable if the animations affect only "visuals" and `wlr_scene_node_at()` always returns the node under the closing window even if it's in the middle of fading animation. In that way, we can make sure animations never interfere with our main logic.

any1:
> The bigger question I sensed in the responses was the amount of encapsulation that can be achieved (e.g. init on create and then encapsulate in separate translation unit(s) using the view event signals as coupling). It might just sound good, but not be feasible in reality so a quick evaluation of that concept would be helpful.

Yeah, it's probably a good idea to give it a go. It does seem to me to be quite feasible. If the animation module just applies visuals to the view, then we'll need to know when the view is created and when it is destroyed. It might also be a good idea to have some way to fast-forward all active animations to the end.

Kyuyrii:
As a user, I feel like the lack of simple animations makes the experience a bit awkward, but I don't want Labwc to feel like Wayfire.

Usually apps and even websites have animations, so the rest of the system not having any gives the impression that something is wrong.

For example, when I open Raven in Budgie, it suddenly appears strange, as it is supposed to "slide" from right to left.

Would this type of animation be very difficult to add and maintain?

yuiiio:
I think scaling during animations would be useful for features like the overview, but
even if we don’t use scaling—for example, I’ve implemented a workspace switcher animation using a slide of the entire scene tree in the workspace—is there a chance this could be accepted in the upstream labwc? (single file <= 150 LOC)

johanmalm:
> I think scaling during animations would be useful for features like the overview, but even if we don’t use scaling—for example, I’ve implemented a workspace switcher animation using a slide of the entire scene tree in the workspace—is there a chance this could be accepted in the upstream labwc? (single file <= 150 LOC)

Can the workspace switcher be implemented as a standalone client? If so, it's outside the project scope.

PR #2827, opened 2025-06-13
input: use mouse emulation for touch when labwc menu is open

jp7677:
Only use the touch protocol when no labwc menu is open. This ensure that labwc menus will be closed on first touch down/up since cursor emulation takes care of closing a menu.
This prevents that labwc menu stay open during touch interactions in native touch mode.

### DIFF
--- a/src/input/touch.c
+++ b/src/input/touch.c
@@ -35,6 +35,20 @@ touch_get_coords(struct seat *seat, struct wlr_touch *touch, double x, double y,
 		return NULL;
 	}
 
+	/*
+	 * Do not return a surface when a labwc menu is still open.
+	 * The idea is: when a client supports touch, but a labwc root or
+	 * client menu is open, use mouse emulation for the first touch
+	 * interaction. Mouse emulation for this first touch then runs through
+	 * the normal cursor logic which closes the menu. All touch interaction
+	 * after that do use the touch protocol because the menu is now closed.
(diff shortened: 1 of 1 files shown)


johanmalm:
Is this still the right answer if cursor emulation is not enabled?

jp7677:
> Is this still the right answer if cursor emulation is not enabled?

Yes, this is especially for the touch mode (mouse emulation off for touch in rc.xml). The idea is:
When a client supports touch, but a Labwc root or client menu is open, use mouse emulation for the first touch interaction. Mouse emulation for this first touch then runs through the normal cursor logic which closes the menu. All touch interaction after that, do use the touch protocol because the menu is now closed.

Contrary to tablet input, native touch input doesn’t use the cursor-process functions, hence this way to hold off touch mode for the first touch when a menu is open.

This matches normal pointer/mouse behavior where the first click on a surface closes a root/client menu and the second one (and all that follows) is passed to the client.

Does this makes sense?

johanmalm:
Yes.
Would you mind adding that longer explanation as a comment in the code. That context make it much clearer why the code is written that way.

jp7677:
> Yes. Would you mind adding that longer explanation as a comment in the code. That context make it much clearer why the code is written that way.

Thanks, done and hopefully less cryptic now.

johanmalm:
Ideal.

PR #2828, opened 2025-06-13
desktop: give focus to a modal dialog rather than its parent

jlindgren90:
Fixes: #2722

/cc @tokyo4j


johanmalm:
Looks good. Thanks.

For reference: https://wayland.app/protocols/xdg-dialog-v1

domo141:
that went fast. for sure scribus seems to behave well now.

tokyo4j:
LGTM. Thanks.

PR #2814, opened 2025-06-09
Bump xdg-shell version from 3 to 6

tokyo4j:

PR #2829, opened 2025-06-13
Wayland protocols: support alpha modifier and ext_data_control_manager

Consolatis:
With that we mostly caught up with sway 1.11. Still missing are:
- `ext_transient_seat_manager_v1` which I don't think we need
- `wp_content_type_manager_v1` which is pretty useless right now
- `zwp_keyboard_shortcuts_inhibit_manager_v1` which sounds like a really bad idea, we do have `ToggleKeybinds` for that which is under the control of the user and not some random web app
- `xdg_wm_base` (#2814)

The [fixup commit](https://github.com/swaywm/sway/commit/9a9be01ad4130e4e19b437fd064f90982974971f) for alpha modifier isn't relevant to us from what I can see because we don't have a SetOpacity action and thus don't overwrite the opacity of any scene nodes.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -6,6 +6,7 @@
 #include <sys/wait.h>
 #include <wlr/backend/headless.h>
 #include <wlr/backend/multi.h>
+#include <wlr/types/wlr_alpha_modifier_v1.h>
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks.

PR #2831, opened 2025-06-13
config: deprecate some key- and mousebinds

johanmalm:
...because Alt- keybinds should be for clients to use and the A-<arrow> default combination is a frequent user complaint because it prevents some common usage patterns like alt-left/right in web browers.

#2825

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -689,7 +689,6 @@ extending outward from the snapped edge.
   A-Tab - next window
   A-S-Tab - previous window
   W-Return - alacritty
-  A-F3 - run bemenu
   A-F4 - close window
   W-a - toggle maximize
   W-<arrow> - resize window to fill half the output
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -796,14 +796,14 @@ extending outward from the snapped edge.
 </labwc_config>
 ```
 
-	Example to un-bind the default alt + left-button-press to move window:
+	Example to un-bind the default Super + left-button-press to move window:
 
 ```
 <mouse>
   <default/>
   <context name="Frame">
-    <mousebind button="A-Left" action="Press"/>
-    <mousebind button="A-Left" action="Drag"/>
+    <mousebind button="W-Left" action="Press"/>
+    <mousebind button="W-Left" action="Drag"/>
   </context>
 </mouse>
 ```
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -692,7 +692,6 @@ extending outward from the snapped edge.
   A-F3 - run bemenu
   A-F4 - close window
   W-a - toggle maximize
-  A-<arrow> - move window to edge
   W-<arrow> - resize window to fill half the output
   A-Space - show window menu
 ```
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -354,18 +354,18 @@
     <doubleClickTime>500</doubleClickTime>
 
     <context name="Frame">
-      <mousebind button="A-Left" action="Press">
+      <mousebind button="W-Left" action="Press">
         <action name="Focus" />
         <action name="Raise" />
       </mousebind>
-      <mousebind button="A-Left" action="Drag">
+      <mousebind button="W-Left" action="Drag">
         <action name="Move" />
       </mousebind>
-      <mousebind button="A-Right" action="Press">
+      <mousebind button="W-Right" action="Press">
         <action name="Focus" />
         <action name="Raise" />
       </mousebind>
-      <mousebind button="A-Right" action="Drag">
+      <mousebind button="W-Right" action="Drag">
         <action name="Resize" />
       </mousebind>
     </context>
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -268,18 +268,6 @@
     <keybind key="W-a">
       <action name="ToggleMaximize" />
     </keybind>
-    <keybind key="A-Left">
-      <action name="MoveToEdge" direction="left" />
-    </keybind>
-    <keybind key="A-Right">
-      <action name="MoveToEdge" direction="right" />
-    </keybind>
-    <keybind key="A-Up">
-      <action name="MoveToEdge" direction="up" />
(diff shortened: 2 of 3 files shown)


Consolatis:
Should we do the same for `A-Left/Right` for mousebinds? If we deprecate some default key combos we could go all the way so we don't end up in the same situation a few months from now.

johanmalm:
v2: We move have a total of three commits removing what I think is the totally of what we should consider. Following discussion on IRC it feels better to do one big hit of changing this rather than doing a bit now and a bit more some other time.

Thoughts welcome as any deprecation can be annoying for users, but I think the overall benefit outweighs the pain on this one.

Consolatis:
I think we should keep the `Move` and `Resize` mousebinds but use `W-` rather than `A-`.

Edit:
Regarding `A-F3`: no opinion from my side. I never used that and have a feeling that people who use wofi and friends are able to set up their own binding.

Regarding `W-Return`: As this is a keybind in "our" space (logo modifier) we could keep it but I don't have a strong preference here either. Something slightly related is that we might want consider exposing multiple terminals in the default root menu so that a new user has \*some\* way to start a terminal.

johanmalm:
> I think we should keep the `Move` and `Resize` mousebinds but use `W-` rather than `A-`.

Good idea.

v3: Have changed the last commit to reflect this.

johanmalm:
Ready for review.

jlindgren90:
Fine with me (though I don't use the defaults anyway)

Consolatis:
@ahesford @tokyo4j any additional suggestions / complains?
LGTM from my side but lets give everyone a choice to voice concerns or suggest improvements.

tokyo4j:
> Something slightly related is that we might want consider exposing multiple terminals in the default root menu so that a new user has *some* way to start a terminal.

This is important to me since I often launch `labwc -C /tmp` to check whether a bug is due to my personal configuration or not.

Can we keep the `W-Return` keybinding? Adding a menu item for terminal may be sufficient, but I prefer keeping the keybinding that follows sway.

LGTM other than that.

johanmalm:
> Can we keep the `W-Return` keybinding? Adding a menu item for terminal may be sufficient, but I prefer keeping the keybinding that follows sway.

Well - we can do, but I find there is something quite clean and elegant about not favoring any particular terminal, launcher, etc.

I've drafted an alternative (to #2842) menu generating thing. Will push shortly.

I normally do `labwc -C /dev/null -s foot` for that kind of use case.

tokyo4j:
I think we have 2 topics to discuss:

1. Should we have a default keybind to launch a terminal emulator?
2. Should we hardcode a particular terminal emulator?

As for the first topic, I think having a default keybind to launch a terminal emulator is quite common and convenient. Many other WMs (Wayfire, sway, Hyprland, i3, AwesomeWM, fluxbox) have such a keybind, and `W-Return` is the most common key combo among them. Having a default menu item to launch a terminal emulator may be a good idea, but I still prefer a default keybind as the fastest way to launch a terminal without any configuration. Yes, `labwc -C /dev/null -s foot` works for my debugging purpose, but a default keybind is a bit more convenient for me since it allows me to quickly launch a second window if the first `foot` window is running another application.

As for the second topic, I agree that hardcoding a particular terminal emulator is not very clean. However, I think it's just a developer's perspective and I don't think removing the keybind brings any real benefit for end users.

A possible solution to achieve the two goals of (1) having a default keybind for terminal emulator and (2) not hardcoding a specific terminal emulator is, as I suggested in [#2843](https://github.com/labwc/labwc/pull/2843#issuecomment-2982616751), adding something like `lab-sensible-terminal` that automatically selects an available terminal emulator. But I'm not sure it's worth adding another binary and instructing packagers to ship it. Honestly I'm pretty happy with what we currently have.

ahesford:
There's a big difference between providing a functional example and trying to think for users. I've railed pretty hard against trying to think for users on several other occasions. It's a bad idea that usually ends up being wrong, and eventually just hinders the same people you're trying to help in the first place.

`W-Return` is only an obvious keybinding to users of a few other window managers that use this preconfigured key combination. It isn't obvious or desirable to me. Those users are also the bunch that tend to be capable of writing their own configurations. Let them do so.

If any of you, as labwc maintainers, want to define default keybindings that make your life easier, do so. But if any two of you disagree on what that keybinding should be, then it shouldn't exist. If the keybinding becomes convoluted just to try to appease everybody, it shouldn't exist. Far better to tell everybody that they should make themselves happy than to try unsuccessfully to appease everybody.

As for "fresh" configurations that work out of the box for testing, it would be just as easy for any of us to create an `rc.xml.clean` and invoke labwc with that. You can even make a shell alias!

johanmalm:
Just to make sure there is no misunderstaning in the discussion - `W-Return` is already bound to alacritty by default. This PR aims to remove some default keybinds.

I suggest that we forget all about the `W-Return` (i.e. we leave it as is). I don't think it does any harm - and might just serve as an example to people. I think that philosophically I agree with @ahesford - in a sort of less-is-more type way, but I have also been at the receiving end of users who do not like things being deprecated.

Interestingly, https://wiki.archlinux.org/title/Labwc now talks about `W-Return` in sections 1 and 2 of labwc :smile: so I'm not sure it's a battle worth having.

So, that would leave the following to be removed (when I've done a fixup commit):

Keybinds:
- `A-F3`:bemenu
- `A-<arrow>`: move window to edge

Mousebinds:
- Change `A-Left` and `A-Right` to `W-Left` and W-Right` respectively.

The `A-<arrow>` and `A-{Right/Left}` have caused users grief numerous times, so I feel there is net benefit in proceeding with these.

Suggest we leave the `lab-sensible-terminal` discussion to the other thread that deals with the proposal to add a more user-friendly default root menu.

johanmalm:
Note: Needs interactive-rebase prior to merge.

johanmalm:
Have tidied up and it's now ready to go.

@tokyo4j and @Consolatis have already posted LGTMs, but following the discussion above it would be nice to get a nod from @ahesford too - if you're happy with approach that is.

ahesford:
No complaints from me.

johanmalm:
> No complaints from me.

Thanks :smile: I thought that was the case, but just wanted to double check.

PR #2832, opened 2025-06-13
config: add <core><primarySelection>

johanmalm:
...to enable/disable primary selection clipboard support. This only works on launch.

The reason it is useful to be able to disable this is that some clients (like browsers) support middle-button-click to start scrolling up/down.

With some clients can be disabled via

   gsettings set org.gnome.desktop.interface gtk-enable-primary-paste false

...but for others (like chromium and electron based programs) a compositor setting is required.

Fixes: #2815


src/config/rcxml.c
```diff
@@ -1161,6 +1161,9 @@ entry(xmlNode *node, char *nodename, char *content, struct parser_state *state)
 		set_bool(content, &rc.reuse_output_mode);
 	} else if (!strcasecmp(nodename, "xwaylandPersistence.core")) {
 		set_bool(content, &rc.xwayland_persistence);
+	} else if (!strcasecmp(nodename, "primarySelection.core")) {
+		set_bool(content, &rc.primary_selection);
+
```
Consolatis: Stray empty line

docs/labwc-config.5.scd
```diff
@@ -177,6 +177,7 @@ this is for compatibility with Openbox.
   <autoEnableOutputs>yes</autoEnableOutputs>
   <reuseOutputMode>no</reuseOutputMode>
   <xwaylandPersistence>no</xwaylandPersistence>
+  <primarySelection>yes</primarySelection>
 </core>
```
Consolatis, line 181: Misses the same in `rc.xml.all`

Consolatis (commented):
Two small nitpicks above. LGTM but didn't test at all.

src/config/rcxml.c
johanmalm: Done

docs/labwc-config.5.scd
johanmalm: Thanks. Good catch. Can't believe I forgot that!

PR #2834, opened 2025-06-14
desktop: enforce followMouse focus policy more strictly (redo)

jlindgren90:
When `followMouse=yes`, prevent stealing focus from another view under the cursor. The requested view *is* allowed to take focus if it ends up under the cursor after raising, or if there is no view under the cursor at all. (Non-view surfaces aren't currently considered.)

This is a revived version of #2822. It's no longer related to the Scribus modal dialog issue, but may still be useful on its own.

See also:
- #2821

v2: allow focusing a new/raised view appearing under the cursor
v3: allow focusing a sub-view/sibling of the requested view
v4: fix new xdg-shell views under mouse not being focused
v5: allow focus if no other view is focused; rebase

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -476,10 +476,12 @@ void xdg_shell_finish(struct server *server);
 void desktop_focus_view(struct view *view, bool raise);
 
 /**
- * desktop_focus_view_or_surface() - like desktop_focus_view() but can
- * also focus other (e.g. xwayland-unmanaged) surfaces
+ * desktop_focus_for_cursor_update() - like desktop_focus_view() but can
+ * also focus other (e.g. xwayland-unmanaged) surfaces.
+ *
+ * Used only for cursor-driven focus updates.
  */
-void desktop_focus_view_or_surface(struct seat *seat, struct view *view,
+void desktop_focus_for_cursor_update(struct seat *seat, struct view *view,
 	struct wlr_surface *surface, bool raise);
 
 void desktop_arrange_all_views(struct server *server);
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -58,8 +58,8 @@ set_or_offer_focus(struct view *view)
 	}
 }
 
-void
-desktop_focus_view(struct view *view, bool raise)
+static void
+_desktop_focus_view(struct view *view, bool raise, bool for_cursor_update)
 {
 	assert(view);
 	/*
@@ -101,6 +101,28 @@ desktop_focus_view(struct view *view, bool raise)
 	}
 
 	/*
+	 * When followMouse=yes, prevent stealing focus from another
+	 * view under the cursor. The requested view *is* allowed to
+	 * take focus if it ends up under the cursor after raising, or
+	 * if there is no view under the cursor at all.
+	 *
+	 * If view_move_to_front() also raised child/sibling views,
+	 * then focus the sibling that's under the cursor.
+	 */
+	if (rc.focus_follow_mouse && !for_cursor_update) {
+		struct cursor_context ctx = get_cursor_context(view->server);
+		if (ctx.view && view_get_root(ctx.view) == view_get_root(view)) {
+			view = ctx.view;
+		} else if (ctx.view) {
+			/*
+			 * Do not steal focus from an unrelated view.
+			 * TODO: consider non-view surfaces also?
+			 */
+			return;
+		}
+	}
+
+	/*
 	 * If any child/sibling of the view is a modal dialog, focus
 	 * the dialog instead. It does not need to be raised separately
 	 * since view_move_to_front() raises all sibling views together.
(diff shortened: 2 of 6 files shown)


domo141:
this works for me. focusfollowsmouse yes.

- cursor over a terminal window, a-tab-ing to other views - may raise other views that don't appear under cursor - and focus stays on the terminal.
- starting new terminals on command line - new view does not open over cursor - focus kept under cursor
-  moving cursor out of any views - alt-tab changes (keyboard) focus on the view chosen w/ alt-tab

with just those tries I did not notice any difference in behavior when changed *requiresMovement
from >no< to >yes< -- will change my config to `yes` whenever this (or related) commit gets merged
just to see if that matters (for me) anymore...

tokyo4j:
I use `followMouse=yes` & `followMouseRequiresMovement=yes` and I found it annoying that it doesn't update focus if the cursor stays in the currently focused window.

I think we should get a consensus on how `followMouse` and `followMouseRequiresMovement` should work first.

Here's what I expect:

| Event                             | `followMouse=no`          | `followMouse=yes` & `followMouseRequiresMovement=yes` | `followMouse=yes` & `followMouseRequiresMovement=no`                            |
| --------------------------------- | ------------------------- | ----------------------------------------------------- | ------------------------------------------------------------------------------- |
| Alt-Tab                           | Focus the selected window | Focus the selected window                             | Focus the selected window only if cursor is on it (Currently this doesn't work?) |
| A window is mapped                | Focus the mapped window   | Focus the mapped window                               | Focus the mapped window only if cursor is on it (Currently this doesn't work?)   |
| The moved cursor enters a surface | Do nothing                | Focus the window the cursor is on                     | Focus the window the cursor is on                                               |

@domo141 Do you expect these behaviors too?

domo141:
> @domo141 Do you expect these behaviors too?

I'd expect the same - and your (nice) table is right in those things that doesn't work

On this machine I am running current HEAD (+ 5 unrelated hacks), and I find it annoying
that focus doesn't stay in the window under cursor - and I was narrow-window enough
not to understand that is not the desired behavior for everyone using `focusFollowMouse=yes`.

One more thing with `followMouseRequiresMovement=yes`

Currently, one needs to move focus out of view and then back to re-focus view using pointer.
Should the focus be given to view under cursor just like that - or after one does (sub)pixel
movement to the pointer.

I tried with 0.8.3 nested - focus is restored to the view under cursor (even) after a small touch
to the pointer device - which someone(tm) could find "disturbing" as focus updates (sporadically).

That said, I "like" how current HEAD behaves if one desired requiresmovement `yes`, but who am
I to say, as I want the behavior that requiresmovement `no` used to provide.

domo141:
There is yet another problem with current HEAD - and one reason I have `followMouseRequiresMovement=no`:

In my case the focused view is often not the topmost view of a workspace; I have focus on a view,
then move to another workspace and back -- now focus is not on that same view but some other view.

With that (not remembering the focused view on a workspace but choosing topmost when switched to) makes me actually not like if future `followMouseRequiresMovement=yes` required one to move
cursor leave and re-enter a view to refocus it...

domo141:
> There is yet another problem with current HEAD - and one reason I have `followMouseRequiresMovement=no`:
> 
> In my case the focused view is often not the topmost view of a workspace; I have focus on a view, then move to another workspace and back -- now focus is not on that same view but some other view.
> 
> With that (not remembering the focused view on a workspace but choosing topmost when switched to) makes me actually not like if future `followMouseRequiresMovement=yes` required one to move cursor leave and re-enter a view to refocus it...

Hmm I don't seem to see problem here (anymore?). If I move to a workspace, view under mouse
gets focus. I mouse not over a view, then the "topmost" view gets focus - so this probably works
as it has always been...

domo141:
Hoo, the interesting thing is that if X client is started from any client, the started X client "steals" focus, but If I
start `foot` (wayland client) from urxvt (or another foot), focus stays on that view mouse is on (as desired).

Anyway, annoyances every now and then when I start urxvt windows (to log things) and then
focus is lost -- replacing `urxvt` with `foot` there is "hard" as I cannot request where the foot window
is going to open...

jlindgren90:
> Hoo, the interesting thing is that if X client is started from any client, the started X client "steals" focus, but If I
> start `foot` (wayland client) from urxvt (or another foot), focus stays on that view mouse is on (as desired).

Is that happening on master, or on this PR branch?

domo141:
> > Hoo, the interesting thing is that if X client is started from any client, the started X client "steals" focus, but If I
> > start `foot` (wayland client) from urxvt (or another foot), focus stays on that view mouse is on (as desired).
> 
> Is that happening on master, or on this PR branch?

in "master" - currently running e365d5ea (2025-06-29). happening also in nested, 03004cf44b2 (2025-08-03)

(I would cherry-pick this PR, but then I could not follow how "master" behaves...)

domo141:
When this commit went unrebaseable I dropped using any workaround to the problem (which still exists) but when this got closed I started looking something I could use - and thanks to the
changes here (and small gdb script), I currently have the this applied to my labwc build tree:

https://github.com/domo141/ruudut/blob/main/with-labwc/11-patches/0005-x-tmp-hack-to-keep-focus-when-starting-an-x-client-f.patch
 
It is somewhat strange the problem applies only when having focus on one x client and the
start another (x client) from it. any other X11/wayland client combination does not have this problem...

PR #2839, opened 2025-06-16
[proof-of-concept] icon priority

Consolatis:
My try at implementing a icon lookup priority via window rules.

Can be tested with
```xml
</windowRules>
	<windowRule identifier="*" iconPriority="server" />
</windowRules>
```

Notable:

- This introduces a subtle change in behavior for `scaled_icon_buffer_set_icon_name()` (only used for menu icons right now): if the icon isn't found it will now fall back to the configured app-id fallback.
- The user facing window rule attribute is called `iconPriority` but the internal property is called `iconPreferServer` to make it work with `enum property` so we don't need to add an additional enum and API for that config.

Missing:
- [ ] docs
- [ ] remove some log spam (especially "Trying to load ..", but maybe also "loaded icon by .."?)

### DIFF
--- a/src/common/scaled-icon-buffer.c
+++ b/src/common/scaled-icon-buffer.c
@@ -15,6 +15,7 @@
 #include "img/img.h"
 #include "node.h"
 #include "view.h"
+#include "window-rules.h"
 
 #if HAVE_LIBSFDO
 
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
 	}
 	return best_buffer;
 }
+
+typedef struct lab_data_buffer *(*load_icon_func_t)(
+	struct scaled_icon_buffer *self, const char *icon_name,
+	int icon_size, double scale);
+
+static struct lab_data_buffer *
+load_local_icon(struct scaled_icon_buffer *self,
+		const char *icon_name, int icon_size, double scale)
+{
+	if (!icon_name) {
+		return NULL;
+	}
+	struct lab_img *img = desktop_entry_load_icon(
+		self->server, icon_name, icon_size, scale);
+	if (!img) {
+		return NULL;
+	}
+
+	wlr_log(WLR_DEBUG, "loaded icon by name: %s", icon_name);
+	struct lab_data_buffer *buffer = lab_img_render(
+		img, self->width, self->height, scale);
+	lab_img_destroy(img);
+	return buffer;
+}
+
+static struct lab_data_buffer *
+load_client_icon(struct scaled_icon_buffer *self,
+		const char *icon_name, int icon_size, double scale)
+{
+	wlr_log(WLR_INFO, "Trying to load icon from client");
+	struct lab_data_buffer *buffer = load_local_icon(
+		self, self->view_icon_name, icon_size, scale);
+	if (buffer) {
+		return buffer;
+	}
+
+	buffer = choose_best_icon_buffer(self, icon_size, scale);
+	if (!buffer) {
+		return NULL;
+	}
+
+	wlr_log(WLR_DEBUG, "loaded icon from client buffer");
+	return buffer_resize(buffer,
+		self->width, self->height, scale);
+}
+
+static struct lab_data_buffer *
+load_app_id_icon(struct scaled_icon_buffer *self,
+		const char *icon_name, int icon_size, double scale)
+{
+	wlr_log(WLR_INFO, "Trying to load icon via app id");
+	struct lab_img *img = desktop_entry_load_icon_from_app_id(
+		self->server, self->view_app_id, icon_size, scale);
+	if (!img) {
+		return NULL;
+	}
+
+	wlr_log(WLR_DEBUG, "loaded icon by app_id");
+	struct lab_data_buffer *buffer = lab_img_render(
+		img, self->width, self->height, scale);
+	lab_img_destroy(img);
+	return buffer;
+}
+
 #endif /* HAVE_LIBSFDO */
 
 static struct lab_data_buffer *
@@ -50,51 +115,37 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
 #if HAVE_LIBSFDO
 	struct scaled_icon_buffer *self = scaled_buffer->data;
 	int icon_size = MIN(self->width, self->height);
-	struct lab_img *img = NULL;
 
-	if (self->icon_name) {
-		img = desktop_entry_load_icon(self->server,
-			self->icon_name, icon_size, scale);
-	} else if (self->view) {
-		if (self->view_icon_name) {
-			wlr_log(WLR_DEBUG, "loading icon by name: %s",
-				self->view_icon_name);
-			img = desktop_entry_load_icon(self->server,
-				self->view_icon_name, icon_size, scale);
-		}
-		if (!img) {
-			struct lab_data_buffer *buffer =
-				choose_best_icon_buffer(self, icon_size, scale);
-			if (buffer) {
-				wlr_log(WLR_DEBUG, "loading icon by buffer");
-				return buffer_resize(buffer,
-					self->width, self->height, scale);
-			}
-		}
-		if (!img) {
-			wlr_log(WLR_DEBUG, "loading icon by app_id");
-			img = desktop_entry_load_icon_from_app_id(self->server,
-				self->view_app_id, icon_size, scale);
-		}
-		if (!img) {
-			wlr_log(WLR_DEBUG, "loading fallback icon");
-			img = desktop_entry_load_icon(self->server,
-				rc.fallback_app_icon_name, icon_size, scale);
+	struct {
+		load_icon_func_t fn;
+		const char *icon_name;
+	} choices[] = {
+		{ load_local_icon, self->icon_name },
+		{ NULL, NULL },
+		{ NULL, NULL },
+		{ load_local_icon, rc.fallback_app_icon_name },
+	};
+	if (self->view) {
+		if (window_rules_get_property(self->view, "iconPreferServer") == LAB_PROP_TRUE) {
+			choices[1].fn = load_app_id_icon;
+			choices[2].fn = load_client_icon;
+		} else {
+			choices[1].fn = load_client_icon;
+			choices[2].fn = load_app_id_icon;
 		}
 	}
-
-	if (!img) {
-		return NULL;
+	for (size_t i = 0; i < ARRAY_SIZE(choices); i++) {
+		if (!choices[i].fn) {
+			continue;
+		}
+		struct lab_data_buffer *buffer = choices[i].fn(
+			self, choices[i].icon_name, icon_size, scale);
+		if (buffer) {
+			return buffer;
+		}
 	}
-
-	struct lab_data_buffer *buffer =
-		lab_img_render(img, self->width, self->height, scale);
-	lab_img_destroy(img);
-
-	return buffer;
-#else
-	return NULL;
 #endif /* HAVE_LIBSFDO */
+	return NULL;
 }
 
 static void
--- a/include/window-rules.h
+++ b/include/window-rules.h
@@ -38,6 +38,7 @@ struct window_rule {
 	enum property ignore_focus_request;
 	enum property ignore_configure_request;
 	enum property fixed_position;
+	enum property icon_prefer_server;
 
 	struct wl_list link; /* struct rcxml.window_rules */
 };
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -328,6 +328,15 @@ fill_window_rule(char *nodename, char *content, struct parser_state *state)
 	/* Properties */
 	} else if (!strcasecmp(nodename, "serverDecoration")) {
 		set_property(content, &state->current_window_rule->server_decoration);
+	} else if (!strcasecmp(nodename, "iconPriority")) {
+		if (!strcasecmp(content, "client")) {
+			state->current_window_rule->icon_prefer_server = LAB_PROP_FALSE;
+		} else if (!strcasecmp(content, "server")) {
+			state->current_window_rule->icon_prefer_server = LAB_PROP_TRUE;
+		} else {
+			wlr_log(WLR_ERROR,
+				"Invalid value for window rule property 'iconPriority'");
+		}
 	} else if (!strcasecmp(nodename, "skipTaskbar")) {
 		set_property(content, &state->current_window_rule->skip_taskbar);
 	} else if (!strcasecmp(nodename, "skipWindowSwitcher")) {
--- a/src/window-rules.c
+++ b/src/window-rules.c
@@ -108,6 +108,10 @@ window_rules_get_property(struct view *view, const char *property)
 					&& !strcasecmp(property, "fixedPosition")) {
 				return rule->fixed_position;
 			}
+			if (rule->icon_prefer_server
+					&& !strcasecmp(property, "iconPreferServer")) {
+				return rule->icon_prefer_server;
+			}
 		}
 	}
 	return LAB_PROP_UNSPECIFIED;


johanmalm:
LGTM - nothing the point about `_equal()`

Have tested with `<windowRules><windowRule identifier="*" iconPriority="server" /></windowRules>` and can confirm that `pcmanfm-qt` get the better looking icon back (the one specified in the .desktop file).

Suggest plastering `scaled-icon-buffer.c` with comments as it is a hard file to read and review. Particularly:
1. The functions `load_{local,client,app_id}_icon`
2. Some context for four `choices[]`

Related: 
- https://github.com/labwc/labwc.github.io/pull/68/files

Consolatis:
I wonder if we should default to "server" if no window rule matches. I'd assume the majority of users actually want the app-id based icon from the configured icon theme rather than the generic application supplied one. Even more relevant when we also add the xwayland backend for the icons. We'd still use the client ones if the app-id based one fails anyway.

johanmalm:
> I wonder if we should default to "server" if no window rule matches. I'd assume the majority of users actually want the app-id based icon from the configured icon theme rather than the generic application supplied one. Even more relevant when we also add the xwayland backend for the icons. We'd still use the client ones if the app-id based one fails anyway.

I'm really not sure what's best here. I personally prefer server, but can see both sides.

I suppose it's similar to CSD/SSD - in that its a choice between letting the client decide or not.

Going with server is more likely to result in a more consistent look.

Consolatis:
Using "server" as default would also have the added benefit that we wouldn't change existing behavior but only add another source of icons for applications that didn't have an icon before.

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
 	}
 	return best_buffer;
 }
+
+typedef struct lab_data_buffer *(*load_icon_func_t)(
+	struct scaled_icon_buffer *self, const char *icon_name,
```
johanmalm: Consider using a context struct here. 

With `load_client_icon()` and `load_app_id_icon()` not using `icon_name`, a struct could make that feel a bit cleaner.

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
...
+	int icon_size, double scale);
+
+static struct lab_data_buffer *
+load_local_icon(struct scaled_icon_buffer *self,
```
johanmalm: Suggest calling it 'load_icon_by_name()`

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
 	}
 	return best_buffer;
 }
+
+typedef struct lab_data_buffer *(*load_icon_func_t)(
+	struct scaled_icon_buffer *self, const char *icon_name,
+	int icon_size, double scale);
+
+static struct lab_data_buffer *
```
johanmalm, line 47: Suggest a comment like:

```
/*
 * Load an icon by name e.g. 'firefox'. libsfdo will search for the icon
 * (e.g. firefox.svg) using the theme specified in rc.xml.
 */
```

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
...
+	}
+
+	wlr_log(WLR_DEBUG, "loaded icon by name: %s", icon_name);
+	struct lab_data_buffer *buffer = lab_img_render(
+		img, self->width, self->height, scale);
+	lab_img_destroy(img);
+	return buffer;
+}
+
+static struct lab_data_buffer *
```
johanmalm, line 60: `/* Load an icon based on the name set by the xdg-toplevel-icon protocol */`

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
...
+load_client_icon(struct scaled_icon_buffer *self,
```
johanmalm: Suggest `load_icon_by_xdg_toplevel_icon_name()`

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
...
+	if (!buffer) {
+		return NULL;
+	}
+
+	wlr_log(WLR_DEBUG, "loaded icon from client buffer");
+	return buffer_resize(buffer,
+		self->width, self->height, scale);
+}
+
+static struct lab_data_buffer *
```
johanmalm, line 84: Maybe...

```
/*
 * Load an icon by a view's app_id. For example, if the app_id is 'firefox', then
 * libsfdo will parse firefox.desktop to get the Icon name and then find that icon
 * based on the theme specified in rc.xml.
 */
```

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
...
+load_app_id_icon(struct scaled_icon_buffer *self,
```
johanmalm: How about `load_icon_by_app_id()`

src/common/scaled-icon-buffer.c
```diff
@@ -42,6 +43,70 @@ choose_best_icon_buffer(struct scaled_icon_buffer *self, int icon_size, double s
...
+		return buffer;
+	}
+
+	buffer = choose_best_icon_buffer(self, icon_size, scale);
+	if (!buffer) {
+		return NULL;
+	}
+
+	wlr_log(WLR_DEBUG, "loaded icon from client buffer");
+	return buffer_resize(buffer,
```
johanmalm: Describe why `buffer_resize()` is needed here by not in `load_local_icon()`

src/common/scaled-icon-buffer.c
```diff
@@ -50,51 +115,37 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
...
-			if (buffer) {
-				wlr_log(WLR_DEBUG, "loading icon by buffer");
-				return buffer_resize(buffer,
-					self->width, self->height, scale);
-			}
+
+	struct {
+		load_icon_func_t fn;
+		const char *icon_name;
+	} choices[] = {
```
johanmalm: Consider `s/choices/load_methods/`

src/common/scaled-icon-buffer.c
```diff
@@ -50,51 +115,37 @@ _create_buffer(struct scaled_scene_buffer *scaled_buffer, double scale)
...
+		{ load_local_icon, self->icon_name },
```
johanmalm: I get confused by what this first load-method is... what is the context where we already have a `self->icon_name`?

src/common/scaled-icon-buffer.c
Consolatis: @tokyo4j Would you mind doing a short description that I could add here?

src/common/scaled-icon-buffer.c
Consolatis: Its used for icons defined in `menu.xml` / pipemenus.

Edit:
~~Added a comment above: `/* First try to load a specified icon as used by menu.c */`~~
Mentioned it in the priority list.

Consolatis:
Thanks a lot for your feedback, really appreciate it! Will incorporate.

src/common/scaled-icon-buffer.c
Consolatis: The function first tries to load an icon based on the name from the icon theme if set by the application. If that is not successful it will then go through the buffers supplied by the application to find a matching one (in terms of size and scale) and then use that one.

With the pending xwayland icon backend the icon update will end up here as well so its not just about xdg toplevel and also not only about the icon name.

Consolatis:
Addressed most of the feedback, the points mentioned in the top post still remain.
> This introduces a subtle change in behavior for scaled_icon_buffer_set_icon_name() (only used for menu icons right now): if the icon isn't found it will now fall back to the configured app-id fallback.

It would be relatively easy to change it back to simply not showing an icon in that case like it was before (e.g. if `!self->view` we simply do not add the fallback to the `load_methods` array). The question for me is: do we want the fallback icon for menu entries setting an invalid / non-existing icon property or not?

src/common/scaled-icon-buffer.c
Consolatis: I removed the load-icon-based-on-the-name functionality and made it an own entry in `load_methods` and then renamed the function to `load_icon_from_client_buffer()` + wrote a little comment on top.

src/common/scaled-icon-buffer.c
tokyo4j: Just like what `lab_img_render()` does, we always resize the icon buffer with `CAIRO_FILTER_GOOD` so that the icon never look aliased. See https://github.com/labwc/labwc/pull/2225.

tokyo4j:
Let me propose another approach without function pointers: https://github.com/Consolatis/labwc/compare/feature/icon_priority...tokyo4j:labwc:icon-priority-revised

I initially thought using function pointer like this PR is the best way to conditionally swap the 2 code paths for icon lookup, but now I think my approach is cleaner for those who read `scaled-icon-buffer.c` for the first time. A downside of my approach is `if (buffer) { return buffer; }` repeated 4 times, but I think it's acceptable.

@johanmalm Could you see it and judge which approach is better?

> > I wonder if we should default to "server" if no window rule matches. I'd assume the majority of users actually want the app-id based icon from the configured icon theme rather than the generic application supplied one. Even more relevant when we also add the xwayland backend for the icons. We'd still use the client ones if the app-id based one fails anyway.
> 
> I'm really not sure what's best here. I personally prefer server, but can see both sides.
> 
> I suppose it's similar to CSD/SSD - in that its a choice between letting the client decide or not.
> 
> Going with server is more likely to result in a more consistent look.

I'm not sure about this. I prefer client-side icons, but I also dislike the inconsistent look of icons in window switcher and panels. So maybe we can disable client-side icons by default for now, and then enable it when there is a new protocol that allows the compositor to pass client-side icons to panels?

Consolatis:
> I initially thought using function pointer like this PR is the best way to conditionally swap the 2 code paths for icon lookup, but now I think my approach is cleaner for those who read `scaled-icon-buffer.c` for the first time.

Agree that it feels cleaner and easier to read as we are only dealing with 2 alternatives here. I tend to unnecessarily over-complicate things to make it easier to change the logic in the future (like some setting to prefer the client buffers before the client icon name or a list of fallback icons). However that does take a toll on complexity and associated readability and is useless if we never implement a feature like that anyway.

> A downside of my approach is if (buffer) { return buffer; } repeated 4 times, but I think it's acceptable.

Yes, that looks totally fine to me.

johanmalm:
> @johanmalm Could you see it and judge which approach is better?

Yes, it looks cleaner and simpler without the function pointers.

Consolatis:
Cherry-picked the commit, changed the default to be app-id based (`server`) and rebased on master to we can test with X11 icons as well.

Consolatis:
Hm.. we are currently doing \*way\* more work than necessary. E.g. we reload / resize the icon on any changes (like app-id or title) even if the outcome will be the same (we are unconditionally using `scaled_scene_buffer_request_update()`). I think its fine for a bit but we really need to overhaul the icon related logic to prevent that in the long term. At the same time the buffer sharing mechanism turned out to be more complex and cumbersome to work with than initially anticipated. Maybe we should brainstorm a bit (after 0.9.0) to figure out a general design that prevents that and also doesn't introduce the complexity with `_equal`.

Testing of this branch would be appreciated to ensure everything works as expected.

I can then remove the logging and squash everything. (I left the logging in for now because `Trying to load ..` makes it easy to verify the correct priority being used). In general I'd really like to have this feature in 0.9.0 (cooldown starts at June 27th I think) because otherwise we will get issues like "labwc started to show the wrong icon".

johanmalm:
^ Thanks. Agree with all. I'll run this branch for a day or two.

johanmalm:
Is there a good text client for the XWayland case?

Consolatis:
> Is there a good test client for the XWayland case?

I've just used my ancient VLC version which doesn't use wayland by default. But I assume most GTK stuff should work when used with `GDK_BACKEND=x11`.

johanmalm:
Have tried a few with `GDK_BACKEND=x11`. The ones I have tried look the same, so was looking for one that's different. 

...except for netsurf which looks... well different with _NET_WM_ICON support.
20250622_19h32m54s_grim

src/view.c
```diff
@@ -2400,6 +2400,7 @@ view_update_title(struct view *view)
 	assert(view);
 	ssd_update_title(view->ssd);
 	wl_signal_emit_mutable(&view->events.new_title, NULL);
+	wl_signal_emit_mutable(&view->events.set_icon, NULL);
```
tokyo4j: Is this needed? If so, I think it should be described in a comment.

EDIT: Ah, this is indeed needed, because title updates may change the window rule to be applied.

tokyo4j:
> Hm.. we are currently doing *way* more work than necessary. E.g. we reload / resize the icon on any changes (like app-id or title) even if the outcome will be the same (we are unconditionally using scaled_scene_buffer_request_update()). I think its fine for a bit but we really need to overhaul the icon related logic to prevent that in the long term. At the same time the buffer sharing mechanism turned out to be more complex and cumbersome to work with than initially anticipated. Maybe we should brainstorm a bit (after 0.9.0) to figure out a general design that prevents that and also doesn't introduce the complexity with _equal.

Agreed.

I'll test this branch for some days, too.

src/view.c
Consolatis: As window rules may be matched using the title this is needed, yes. In general I don't really like app-id or title changes to emit that signal though, the scaled-icon-buffer should likely have its own listeners for the existing new_title / new_app_id signals. Those could then check if the result of the `window_rules_get_property()` is now different. And if not (and not in the changed app-id handler and detecting that the app-id actually changed), simply return. Same could be done for the set_icon event listener, e.g. there is not really a point in deleting and re-creating everything if the icon name and the buffers are the same anyway.

Edit:
Pushed a RFC commit to implement the suggestion. Can be merged as part of this PR (keeping it as an own commit) or I can open another PR if deemed necessary.

tokyo4j: >  Pushed a RFC commit to implement the suggestion. Can be merged as part of this PR (keeping it as an own commit) or I can open another PR if deemed necessary.

Let's merge your RFC commit in this PR. I was just too lazy to add another listener for the `new_app_id` signal and your change feels cleaner.

This is just for my two cents, but I feel that `scaled_icon_buffer` listening to `update_title` is actually not very consistent because the window rule to be applied is usually determined when the window is mapped and updates of app-id or title don't reapply a new window rule. But I understand that your current approach is the best solution for now as we are creating `struct ssd` before the window is mapped. So perhaps we can remove the listeners on `scaled_icon_buffer` if we defer the ssd creation to the time the window is mapped. It will also get rid of the current odd behavior that `ssd_create()` is always called for 0x0 window.

src/common/scaled-icon-buffer.c
```diff
@@ -183,22 +238,69 @@ scaled_icon_buffer_create(struct wlr_scene_tree *parent, struct server *server,
...
+		if (icon_buffers_equal(
+				&self->view_icon_buffers,
+				&self->view->icon.buffers)) {
+			return;
+		}
+	} else if (!self->view->icon.name) {
+		zfree(self->view_icon_name);
+	} else {
+		xstrdup_replace(self->view_icon_name, self->view->icon.name);
 	}
```
tokyo4j: These branches feels hard to follow. I suggest changing it to something like this:
```suggestion
	if (str_equal(self->view_icon_name, self->view->icon.name)
			&& icon_buffers_equal(&self->view_icon_buffers,
				&self->view->icon.buffers)) {
		return;
	}

	zfree(self->view_icon_name);
	if (self->view->icon.name) {
		xstrdup_replace(self->view_icon_name, self->view->icon.name);
	}
```

tokyo4j:
Overall the change looks good. I've been running this branch (with ASAN enabled) for a day and I haven't found any problems. I'll do a final review after the log spams are removed.

src/common/scaled-icon-buffer.c
Consolatis: That would unnecessarily free and heap-allocate `self->view_icon_name`, even if it hasn't been changed.
Agree that the branches are hard to follow though. I'll think of something.

tokyo4j: > That would unnecessarily free and heap-allocate self->view_icon_name, even if it hasn't been changed.

This is true, but I'd prefer readability over the overhead that no one notices.

Consolatis: The thing with unnecessary overhead in general is that it stacks up and slowly but steadily the whole thing becomes slow and laggy. And the higher the layer for unnecessary overhead the higher the impact. Some small unnecessary `free() + strlen() + malloc() + strncpy()` isn't that bad in itself but again, it stacks up and should be prevented if possible.

johanmalm:
Have been running this branch for a couple of days too and it seems fine.

src/common/scaled-icon-buffer.c
Consolatis: Added a temporary boolean to make it easier to follow.

Consolatis:
Removed the additional log output + rebased + squashed. Ready from my point of view but I think @tokyo4j wanted to have another look at the diff before we merge.

danielfikko:
Hi, it's not clear to me what server and client are. Server is based on the `app-id`, i.e.  desktop entry and icon with the same name as the `app-id`? So what would the client be? It's based on `xdg-toplevel-icon`?

Consolatis:
> Hi, it's not clear to me what server and client are. Server is based on the `app-id`, i.e. desktop entry and icon with the same name as the `app-id`? So what would the client be? It's based on `xdg-toplevel-icon`?

Yes, xdg-toplevel-icon for wayland native applications or the equivalent `_NET_WM_ICON` for X11 applications. This PR includes these docs (and I am pretty bad at writing them in general), any suggestions to improve them?

> *\<windowRules>\<windowRule iconPriority="">* [client|server]
> By default, labwc is trying to find application icons based on their
> app-id, either via .desktop file or by finding an icon with the same
> name. If that fails labwc will then try to use client supplied icons.
> This property allows prioritizing client supplied icons for specific
> applications.

danielfikko:
> > Hi, it's not clear to me what server and client are. Server is based on the `app-id`, i.e. desktop entry and icon with the same name as the `app-id`? So what would the client be? It's based on `xdg-toplevel-icon`?
> 
> Yes, xdg-toplevel-icon for wayland native applications or the equivalent `_NET_WM_ICON` for X11 applications. This PR includes these docs (and I am pretty bad at writing them in general), any suggestions to improve them?

Maybe something like that?

If that fails labwc will then try to use the built-in icons provided by the client using `xdg-toplevel-icon`/`_NET_WM_ICON`.

danielfikko:
At first I understood that the `server` was based on `xdg-toplevel-icon` because of the “based on their app-id”, since technically `xdg-toplevel-icon` also uses the `app-id`, right? Also, “client supplied icons” can be understood as the desktop entry/icon file packaged with the application, like `/usr/share/icons/hicolor/scalable/apps/firefox.svg`. I suggest something like this:

By default, labwc tries to search for desktop entry files with a name that matches the application's `app-id` and then uses the icon defined in this desktop entry file or labwc will try to use icon files with a name that matches the application's `app-id`. If this fails, labwc will try to use built-in icons using `xdg-toplevel-icon`/`_NET_WM_ICON` extensions for wayland/x11 applications. The default, `server`, will maintain this behavior, while `client` will prioritize built-in icons for applications that implement `xdg-toplevel-icon`/`_NET_WM_ICON`.

EDIT: I think my [first suggestion](https://github.com/labwc/labwc/pull/2839#issuecomment-3010171585), just explaining that the `client` will use the `xdg-toplevel-icon`, clarifies things enough. But, is there anywhere else in the documentation explaining how labwc searches for icons? if not, this might be the chance to explain it in detail.

Consolatis:
> At first I understood that the `server` was based on `xdg-toplevel-icon` because of the “based on their app-id”, since technically `xdg-toplevel-icon` also uses the `app-id`, right?

The xdg-toplevel-icon doesn't use the app-id at all. It supplies an icon name and/or a set of actual pixel data in different resolutions. The `_NET_WM_ICON` property only supplies pixel data in different resolutions.

Rewritten as:
```
*<windowRules><windowRule iconPriority="">* [client|server]
	By default, labwc is trying to find application icons based on their
	app-id, either via .desktop file or by finding an icon with the same
	name. If that fails labwc will then try to use client supplied icons,
	accomplished with the xdg-toplevel-icon protocol for wayland native
	applications or the _NET_WM_ICON property for X11 applications.

	This property allows prioritizing client supplied icons for specific
	applications. Default is server.
```

src/common/scaled-icon-buffer.c
```diff
@@ -183,22 +236,66 @@ scaled_icon_buffer_create(struct wlr_scene_tree *parent, struct server *server,
 	return self;
 }
 
+static void
+update_icon(struct scaled_icon_buffer *self)
+{
+	scaled_scene_buffer_request_update(self->scaled_buffer,
+		self->width, self->height);
+}
+
```
tokyo4j: I think we can directly call `scaled_scene_buffer_request_update()` rather than defining `update_icon()`.

docs/labwc-config.5.scd
```diff
@@ -1210,6 +1210,16 @@ situation.
 	can be caused by *<margin>* settings or exclusive layer-shell clients
 	such as panels.
 
+*<windowRules><windowRule iconPriority="">* [client|server]
+	By default, labwc is trying to find application icons based on their
```
tokyo4j: `s/is trying to/tries to/`

tokyo4j:
LGTM other than my nitpicks above. Thanks for working on this!

PR #2840, opened 2025-06-16
[No wlroots change required] xwayland: support _NET_WM_ICON

tokyo4j:
Support `_NET_WM_ICON` property without waiting for [wlroots MR](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5079)

Just copied from https://github.com/consolatis/labwc/tree/feature/x11_icons

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -3,6 +3,8 @@
 #include <assert.h>
 #include <stdlib.h>
 #include <wlr/xwayland.h>
+#include "buffer.h"
+#include "common/array.h"
 #include "common/macros.h"
 #include "common/mem.h"
 #include "config/rcxml.h"
@@ -16,6 +18,20 @@
 #include "window-rules.h"
 #include "workspaces.h"
 #include "xwayland.h"
+
+enum atoms {
+	ATOM_NET_WM_ICON = 0,
+
+	ATOM_COUNT,
+};
+
+static const char * const atom_names[] = {
+	[ATOM_NET_WM_ICON] = "_NET_WM_ICON",
+};
+
+static_assert(ARRAY_SIZE(atom_names) == ATOM_COUNT, "atom names out of sync");
+
+static xcb_atom_t atoms[ATOM_COUNT] = {0};
 
 static void xwayland_view_unmap(struct view *view, bool client_request);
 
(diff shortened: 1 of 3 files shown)


tokyo4j:
I removed `set_icon` signals for simplicity. Let's restore it later.

I also added `xcb-ewmh` as a dependency because linker errors occurred without it.

src/xwayland.c
```diff
@@ -580,6 +596,62 @@ handle_set_strut_partial(struct wl_listener *listener, void *data)
...
+				dst_pixel[2] = src_pixel[2] * src_pixel[3] / 255;
+				dst_pixel[3] = src_pixel[3];
+			}
+		}
+
+		struct lab_data_buffer *buffer = buffer_create_from_data(
+			buf, iter.width, iter.height, stride);
+		array_add(&buffers, buffer);
+	}
+
```
Consolatis, line 647: Suggest adding a comment like `/* view takes ownership of the buffers */ to make clear that the `calloc()` call above doesn't leak although it might appear like it.

src/xwayland.c
```diff
@@ -580,6 +596,62 @@ handle_set_strut_partial(struct wl_listener *listener, void *data)
...
+	view_set_icon(&xwayland_view->base, NULL, &buffers);
+	wl_array_release(&buffers);
```
Consolatis, line 650: Suggest adding a comment like `/* view takes ownership of the buffers */` to make clear that the `calloc()` call above doesn't leak although it might appear like it.

src/xwayland.c
```diff
@@ -17,6 +19,20 @@
...
+	ATOM_COUNT,
+};
+
+static const char * const atom_names[] = {
+	[ATOM_NET_WM_ICON] = "_NET_WM_ICON",
+};
+
+static_assert(ARRAY_SIZE(atom_names) == ATOM_COUNT, "atom names out of sync");
+
+xcb_atom_t atoms[ATOM_COUNT] = {0};
```
Consolatis: Should be `static xcb_atom_t atoms[ATOM_COUNT] = {0};` (as its static you could also leave out the initialization, up to you).

src/xwayland.c
```diff
@@ -580,6 +596,62 @@ handle_set_strut_partial(struct wl_listener *listener, void *data)
...
+		view_set_icon(&xwayland_view->base, NULL, NULL);
+		goto out;
+	}
+
+	xcb_ewmh_wm_icon_iterator_t iter = xcb_ewmh_get_wm_icon_iterator(&icon);
+	struct wl_array buffers;
+	wl_array_init(&buffers);
+	for (; iter.rem; xcb_ewmh_get_wm_icon_next(&iter)) {
+		size_t stride = iter.height * 4;
+		uint32_t *buf = calloc(1, iter.width * stride);
```
Consolatis: Should use our `xzalloc()` wrapper rather than `calloc()`, this should (hopefully) also fix the CI static analyzer error.

Consolatis:
Some small nitpicks above, other than that LGTM. A third pair of eyes might be good though because we both were involved in writing the code.

johanmalm:
LGTM. Should the commits be squashed?

src/xwayland.c
tokyo4j: Added the comment above `view_set_icon()` in both `xdg.c` and `xwayland.c`.

src/xwayland.c
tokyo4j: Done.

src/xwayland.c
tokyo4j: Added `static`.

Issue #1893, opened 2024-06-09
Debian Packaging - Fixing Terminal Recommends


rcalixte:
This was originally meant for @b1rger (thanks for packaging this!) and I wasn't sure if email or an issue was a better form of contact but then I realized that the reason for the change was due to upstream.

The core issue here is that the [labwc Debian package](https://packages.debian.org/sid/labwc) currently has a **Recommends** of `alacritty` and by default, Debian will install packages labelled as such. The typical convention though is to use `x-terminal-emulator` which allows the user to use any terminal that provides this functionality. As LabWC is a Wayland compositor, that might not be desired, in which case it would make sense to append the **Recommends** to something like `alacritty | foot | kitty` (or `alacritty | x-terminal-emulator`?) or remove the terminal dependency completely as is the most common approach. Maybe this could be an opportunity to introduce a `wayland-terminal-emulator` virtual package?  (Moving it to **Suggests** could be a valid course of action too but it really just moves the core issue somewhere else.)

For comparison, the other wlroots-based compositors in Debian's repositories are configured as such with regards to terminals:

* **cage:** _No terminal application dependencies_
* **gamescope:** _No terminal application dependencies_
* **phoc:** _No terminal application dependencies_
* **sway:** `Recommends: foot`
* **wayfire:** _No terminal application dependencies_

I understand this core issue stems from [this line in default-bindings.h](https://github.com/labwc/labwc/blob/master/include/config/default-bindings.h#L20) so maybe the larger issue is defining a better mechanism to determine the default terminal, possibly among the available ones already on the system or possibly using a list of pre-defined terminals, selected based on availability, since it would most likely be modified by the end user at some point later on.

Consolatis:
This sounds like a packaging related issue on the surface so feel free to use this issue to discuss `Suggests` / `Recommends`.

Regarding the labwc repo internal default configurations:
- hardcoded keybind defaults via https://github.com/labwc/labwc/blob/master/include/config/default-bindings.h
- hardcoded menu defaults via https://github.com/labwc/labwc/blob/master/src/menu/menu.c#L814
- example [rc.xml.all](https://github.com/labwc/labwc/blob/master/docs/rc.xml.all) (this is should reflect the hardcoded defaults)
- example [menu.xml](https://github.com/labwc/labwc/blob/master/docs/menu.xml) (which I think the Debian package actually uses by default by moving it to `/etc/xdg/labwc/`)
[...]

PR #2842, opened 2025-06-17
src/menu: add small selection of default terminals to root-menu

Consolatis:
So new users (or `labwc -C /dev/null` testruns) have some way to spawn a terminal.

---

I've gone with `foot`, `alacritty` and `xterm` but feel free to discuss alternatives likely available on user systems.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -488,7 +488,7 @@ fill_item(char *nodename, char *content)
 #if HAVE_LIBSFDO
 		if (rc.menu_show_icons && !string_null_or_empty(content)) {
 			xstrdup_replace(current_item->icon_name, content);
-			current_menu->has_icons = true;
+			current_item->parent->has_icons = true;
 		}
 #endif
 	} else if (!strcmp(nodename, "name.action")) {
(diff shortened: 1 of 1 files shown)


tokyo4j:
I like these additions. I think their names should be the ones from desktop entries (`Foot`, `Alacritty` and `XTerm`) to make them consistent with other default entries (`Reconfigure` and `Exit`).

I'm not sure if this is in this PR's scope, but I think we can also add a menu entry for `xdg-open https://labwc.github.io` so that new users can open our documentation quickly.

johanmalm:
This doesn't feel quite right to me. Need to have a think.

Ref:
- https://github.com/labwc/labwc/issues/1893
- https://github.com/linuxmint/nemo/blob/88ec5d2/eel/eel-gnome-extensions.c#L40-L60

Consolatis:
#2843 Feels a bit too much to me just for the initial root menu (which will most likely be changed anyway). Its not a hard no from my side though. I'll comment there with further suggestions.

Consolatis:
Closing in favor of
- #2877

PR #2843, opened 2025-06-17
[RFC] menu: add terminal to default root-menu

johanmalm:
Welcome feedback on an approach like this instead of hardcoding `foot`, `alacritty`, etc regardless of whether or not they are installed on the system.

Ref:
- https://github.com/labwc/labwc/issues/1893
- https://github.com/linuxmint/nemo/blob/88ec5d2/eel/eel-gnome-extensions.c#L40-L60
- #2842
- #2831

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1021,6 +1021,79 @@ update_client_list_combined_menu(struct server *server)
 	menu_create_scene(menu);
 }
 
+static bool
+is_ixoth(char *filename)
+{
+	struct stat sb;
+	return stat(filename, &sb) == 0 && (sb.st_mode & S_IXOTH);
+}
+
+static bool
+isprog(const char *filename, gchar **paths)
+{
+	/* handle filename with absolute path */
+	if (filename[0] == '/') {
+		return is_ixoth((char *)filename);
+	}
+
+	/* handle filename with relative path by iterating over $PATH */
+	for (gchar **p = paths; *p; p++) {
+		char path[4096];
+		snprintf(path, sizeof(path), "%s/%s", *p, filename);
+		if (is_ixoth(path)) {
+			return true;
+		}
+	}
+	return false;
+}
+
+static void
+try_to_add_terminal(struct menu *menu)
+{
+	const char * const terminals[] = {
+		"alacritty",
+		"dtterm",
+		"foot",
+		"gnome-terminal",
+		"kgx",
+		"kitty",
+		"konsole",
+		"lxterminal",
+		"mate-terminal",
+		"nxterm",
+		"rxvt",
+		"sakura",
+		"st",
+		"terminator",
+		"terminology",
+		"tilix",
+		"wezterm",
+		"xfce4-terminal",
+		"xterm",
+		"color-xterm",
+	};
+
+	char *path = getenv("PATH");
+	if (!path) {
+		return;
+	}
+	gchar **paths = g_strsplit(path, ":", -1);
+
+	for (size_t i = 0; i < ARRAY_SIZE(terminals); ++i) {
+		const char *terminal = terminals[i];
+		if (isprog(terminal, paths)) {
+			wlr_log(WLR_DEBUG, "add terminal %s to menu", terminal);
+			current_item = item_create(menu, _("Terminal"),
+				/* array */ false);
+			fill_item("name.action", "Execute");
+			fill_item("command.action", (char *)terminal);
+			break;
+		}
+	}
+	g_strfreev(paths);
+
+}
+
 static void
 init_rootmenu(struct server *server)
 {
@@ -1030,6 +1103,9 @@ init_rootmenu(struct server *server)
 	if (!menu) {
 		current_menu = NULL;
 		menu = menu_create(server, "root-menu", "");
+
+		try_to_add_terminal(menu);
+
 		current_item = item_create(menu, _("Reconfigure"), false);
 		fill_item("name.action", "Reconfigure");
 		current_item = item_create(menu, _("Exit"), false);


Consolatis:
-  I'd would welcome the logic to move somewhere else. `menu.c` is already way too big because it does many things at once
   - parsing
   - rendering
   - handling input
I think we don't need to add more like finding available terminals on the system.
- This only uses the first terminal found, I think a `FOR_EACH_TERMINAL()` macro (or some other kind of iterator approach) would be better, then the caller can either only use the first one (e.g. the `W-Return` use-case) or all of the ones found (fallback `root-menu` use-case)

src/menu/menu.c
```diff
@@ -1021,6 +1021,79 @@ update_client_list_combined_menu(struct server *server)
...
+		}
+	}
+	return false;
+}
+
+static void
+try_to_add_terminal(struct menu *menu)
+{
+	const char * const terminals[] = {
+		"alacritty",
```
Consolatis, line 1054: Should this list start with `x-terminal-emulator`? That should respect the user configuration on Debian based systems and be simply ignored on other distros.

ahesford:
I don't really like either option but, if we must choose one, I'd rather just have a static sub-menu of a few common terminals.

Walking the path to find the first of a long list of terminals is error prone. Checking `S_IXOTH` on a file is subject to both false positives and false negatives; filesystems can be mounted `noexec`, or the executable could be owned and executable by the user without allowing others to execute it. There is simply no robust way to identify whether a given path can be executed except to try to execute it, which would clearly be inappropriate here.

Even if this check is a reliable proxy, it is likely to produce undesirable results for the very people it attempts to help. Shortcuts in the root menu are inherently personal, and sort of demand that the user write a proper configuration. I might have `alacritty` and `foot` installed, but I might want `foot` to be the default terminal. Trying to divine user intent for what should be an exceptional condition---users don't write a configuration but expect the program to magically behave exactly as they expect---just complicates everything for little real benefit. After all, rather than hard-coding just a few entries, we're now hard-coding a much longer list of potential terminal candidates. Who curates this list? How should it be sorted?

If we really want to try to divine user intent and select a desirable terminal for the menu that a user didn't bother to write for himself, I supposed I'd rather check to see if `$TERMINAL` is defined in the environment. (At least one other compositor or X11 window manager looks for `$TERMINAL` to point to the preferred terminal.) If so, render a menu item labeled "Terminal Emulator" that will invoke the contents of that variable. If not, just stick a few "common" static entries in there. What is "common"? Nobody knows; pick the terminal emulators that the core team uses (I like `foot`), go by GitHub stars, or use some other arbitrary criteria.

tokyo4j:
I'm not particularly against listing many terminal emulator candidates, but I don't really like such a hidden control that cannot be simply described in our default `menu.xml`. I'd rather like to see "Failed to execute alacritty" message box which encourages the user to overwrite it.

I like adding submenu items with some specific terminal emulators, but another possible solution is to add `lab-sensible-terminal` just like [i3-sensible-terminal](https://github.com/i3/i3/blob/next/i3-sensible-terminal) and execute it from the menu item (and from `W-Return` keybind too?) so that users can easily see what's happening.

Consolatis:
I am mostly interested in providing some way for new users to do something with labwc after installing. Note, most users won't start labwc from the TTY but from a display manager of some kind so you really can't do anything with labwc (or in fact the whole system if labwc is the only installed "enviornment") at the moment. See for example https://youtu.be/lQUCNWuwSqo?t=360 (should go without saying but that person is not really interested in anything wayland but was curious about the newly added labwc option to the installer which is totally fine so please don't dislike the video, write rude comments or anything like that).

I think we can improve the initial impression drastically by enhancing the on-boarding experience by adding a selection of terminal emulators and as mentioned by @tokyo4j in the other PR something like `xdg-open https://labwc.github.io`.
It only matters until the user sets up the system to a point where it isn't necessary anymore. E.g. by using some panel, configuring `menu.xml` or defining own keybinds.

ahesford:
Stealing the `i3-sensible-terminal` script, maybe with some cleanup, seems like the best way to accomplish this goal.

Consolatis:
I fail to see how adding another binary to labwc (which then also need to be packaged by distros) is beneficial compared to having the list in labwc itself.

ahesford:
Adding a script defers the terminal search until it's actually needed, honors a $TERMINAL environment variable to allow simple customization in a way that a few other environments already expect, and does the search the only right way---by trying to exec and, if that fails, moving on to the next option.

If the wrapper script isn't palatable, I'd much rather just hard-code a few well-chosen default entries that may be available rather than trying to be too clever about finding a good choice within labwc.

johanmalm:
> I am mostly interested in providing some way for new users to do something with labwc after installing. Note, most users won't start labwc from the TTY but from a display manager of some kind so you really can't do anything with labwc (or in fact the whole system if labwc is the only installed "enviornment") at the moment. See for example https://youtu.be/lQUCNWuwSqo?t=360 (should go without saying but that person is not really interested in anything wayland but was curious about the newly added labwc option to the installer which is totally fine so please don't dislike the video, write rude comments or anything like that).

@Consolatis I think this summarises the use-case beatifully.

@tokyo4j @ahesford You make good points and thanks for pushing back; I think we are potentially straying into a space that is arguably outside of the compositor and there is something elegant about keeping "certain things" outside of the compositor monolith.

I would generally argue against shipping "helpful scripts" with the compositor package, but given the use case above, `lab-sensible-terminal` is quite appealing. -- EDIT -- And most users probably just run `labwc` from a repo so the script then be in $PATH and just work.

For comparison, `sway` provides `swaybar`, `swaymsg`, `swaynag` with the compositor - as well as deskop wallpapers, optional automatic config generation on first start - and not to mention the closely linked swaybg, swayidle, etc... so we are pretty modest in comparison.

johanmalm:
I think the `lab-sensible-terminal` could be a good idea, but just posting a couple of other ideas here for consideration:

1. Import `labwc-menu-generator` to the `labwc` repo so that it gets packaged and we can rely (reasonably) on it being installed with `labwc`.
2. Ship a script that parses desktop files and produces an `Applications` pipemenu.

<details><summary>Sample script:</summary>

```python
#!/usr/bin/env python3

import os
import sys

builtin_desktop_directories = [
    {
	    "Name": "Accessories",
	    "Icon": "applications-accessories",
        "Categories": "Accessibility;Core;Utility;",
    }, {
	    "Name": "Development",
        "Icon": "applications-development",
        "Categories": "Development;",
    }, {
        "Name": "Education",
        "Icon": "applications-science",
        "Categories": "Education;",
    }, {
        "Name": "Games",
        "Icon": "applications-games",
        "Categories": "Game;",
    }, {
        "Name": "Graphics",
        "Icon": "applications-graphics",
        "Categories": "Graphics;",
    }, {
        "Name": "Multimedia",
        "Icon": "applications-multimedia",
        "Categories": "Audio;Video;AudioVideo;",
    }, {
        "Name": "Internet",
        "Icon": "applications-internet",
        "Categories": "Network;",
    }, {
        "Name": "Office",
        "Icon": "applications-office",
        "Categories": "Office;",
    }, {
        "Name": "Settings",
        "Icon": "preferences-desktop",
        "Categories": "Settings;Screensaver;",
    }, {
        "Name": "System",
        "Icon": "applications-system",
        "Categories": "Emulator;System;",
    },
]

def parse_one_app(path):
    app = {}
    try:
        with open(path, "r", encoding="utf-8") as f:
            lines = f.read().split("\n")
            inside = False
            for line in lines:
                if line.startswith("["):
                    inside = line == "[Desktop Entry]"
                    continue
                if inside:
                    if "=" in line:
                        key, value = line.split("=", 1)
                        app[key] = value
    except:
        print(f"warn: error reading '{path}'", file=sys.stderr)
    return app

def parse_apps(paths):
    apps = []
    for path in paths:
        for (_, _, filenames) in os.walk(path):
            for filename in filenames:
                app = parse_one_app(path + '/' + filename)
                if "Type" not in app or app["Type"] != "Application":
                    continue
                if "NoDisplay" in app and app["NoDisplay"] == "true":
                    continue
                if "Terminal" in app and app["Terminal"] == "true":
                    continue
                if "Categories" not in app or "Icon" not in app or "Exec" not in app:
                    continue
                app["cmd"] = app["Exec"].split("%", 1)[0]
                apps.append(app)
    return apps

def traverse_directories(apps):
    for directory in builtin_desktop_directories:
        categories = [s.strip() for s in directory["Categories"].split(";") if s.strip()]
        print(f"<menu id=\"{directory["Name"]}\" label=\"{directory["Name"]}\" icon=\"{directory["Icon"]}\">")
        for category in categories:
            for app in apps:
                if category in app["Categories"]:
                    print(f"  <item label=\"{app["Name"]}\" icon=\"{app["Icon"]}\">")
                    print(f"    <action name=\"Execute\"><command>{app["cmd"]}</command></action>")
                    print(f"  </item> <!-- {app["Name"]} -->")
        print(f"</menu> <!-- {directory["Name"]} -->")

def create_menu():
    print("<?xml version=\"1.0\" encoding=\"utf-8\"?>")
    print("<openbox_menu>")
    print("<menu id=\"root-menu\">")

    paths = ['/usr/share/applications/']
    apps = parse_apps(paths)
    traverse_directories(apps)

    print("<separator/>")
    print("<menu id=\"Preferences\" label=\"Preferences\">")
    print("<item label=\"Reconfigure\"><action name=\"Reconfigure\"/></item>")
    print("</menu> <!-- Preferences -->")
    print("<menu id=\"Exit\" label=\"Exit\">")
    print("<item label=\"Logout\"><action name=\"Exit\"/></item>")
    print("</menu> <!-- Exit -->")
    print("</menu>")
    print("</openbox_menu>")

def main():
    create_menu()

if __name__ == '__main__':
    main()
```

</details>

Consolatis:
I've added both of the PRs to the 0.9.0 milestone. I don't really mind which variant we choose but would really appreciate if we can come to a conclusion to improve the initial experience with labwc to (at least more likely) being able to open some terminal. Adding `xdg-open https://labwc.github.io` or `xdg-open https://labwc.github.io/getting-started.html` would be a bonus.

- Listing some not installed terminal emulators is still better than not listing any terminal emulator.
- Adding a sensible-labwc-terminal executable + fallback root-menu entry for it is also fine.
- Same for adding a pipemenu executable + fallback root-menu entry which shows all installed ones (or even all applications).
- Adding some hardcoded logic within labwc to find some terminal also works for me.

We can also use a less-preferred option for now and then discuss a "proper" solution for 0.9.1.
Note: this really only affects users not using any root menu at all, e.g. freshly installed labwc without the example `menu.xml`.

johanmalm:
> I don't really mind which variant we choose but would really appreciate if we can come to a conclusion to improve the initial experience with labwc to (at least more likely) being able to open some terminal.

Thanks for pushing this.

I also feel that we ought to make the initial experience better.

On the whole I think that our compositor is a system component and anything that serves to setup/configure/create a Desktop Environment is fundamentally outside of its scope. BUT, there are a few small things we can do which would lower the entrance barrier and just make it a little easier to get started.

I think it's important that any such steps we take:
1. Do not become a slippery slope into to the DE domain
2. Do not disproportionately increase maintenance burden
3. Provide solutions which stand the test of the time without having to be constantly improved, tweaked, adapted, translated, "just add another option", etc...
4. Have a clear separation line and finish point, where we say "we will go this far and then draw a line".

>     * Listing some not installed terminal emulators is still better than not listing any terminal emulator.
>     * Adding a sensible-labwc-terminal executable + fallback root-menu entry for it is also fine.
>     * Same for adding a pipemenu executable + fallback root-menu entry which shows all installed ones (or even all applications).
>     * Adding some hardcoded logic within labwc to find some terminal also works for me.

Looks like there is enough discussion above which leans towards discarding #2842 and #2843 (and I agree with this).

I suggest the following steps:

a. Add `lab-sensible-terminal` script before `0.9.0` with an associated `_("Terminal")` entry in the default root-menu. I created one last week which we could use as a starter for ten to critique. (pasted below).
b. After `0.9.0` consider other improvements. I wouldn't be adverse to including a pipemenu scripts for all `.desktop` entries. If we improve the menu first so that it can sensibly handle large numbers of items by allowing up/down scrolling, then we could avoid handling of "directories" which significantly simplifies implementations including i18n/nls.
c. We could also consider a `_("Help")` submenu with our man pages.
d. We could consider making man pages for "getting-started", "autostart", "environment", etc.

> Adding `xdg-open https://labwc.github.io` or `xdg-open https://labwc.github.io/getting-started.html` would be a bonus.

I prefer not depending on `xdg-open`. Suppose we could consider a `lab-sensible-browser` with `xdg-open` as the first entry.

Example `lab-sensible-terminal` script:

```bash
#!/bin/sh

terminals="\
	x-terminal-emulator \
	alacritty \
	foot \
	kitty \
	xfce4-terminal \
	lxterminal \
	qterminal \
	mate-terminal \
	gnome-terminal \
	konsole \
	sakura \
	dtterm \
	kgx \
	nxterm \
	rxvt \
	st \
	terminator \
	terminology \
	tilix \
	color-xterm \
	urxvt \
	termit \
	Eterm \
	aterm \
	uxterm \
	xterm \
	roxterm \
	termite \
	lilyterm \
	terminix \
	guake \
	tilda \
	hyper \
	wezterm \
	rio"

for terminal in $terminals
do
	if type "${terminal}" >/dev/null 2>&1
	then
		exec "$terminal" "$@"
	fi
done

labnag -m 'lab-sensible-terminal could not find a terminal emulator' -Z X :
```

Some inspiration from https://www.bunsenlabs.org/ below. Don't get me wrong here - that level of support would be way too much, but it might just give us ideas of what a friendly welcome could look like.

cool-menu

tokyo4j:
I believe adding `lab-sensible-terminal` is the most sensible solution suggested so far, and I almost agree with @johanmalm's plan, but I still feel not sure that our/packager's cost of shipping another executable is worth the enhancement of out-of-the-box experience.

So let me suggest another idea that achieves a similar goal to `lab-sensible-terminal`, but without shipping another executable - how about adding `OpenTerminal` (and `OpenBrowser`) action? This feels more minimalistic and avoids additional work for packagers, while deferring the detection of available terminal emulators.

Consolatis:
> So let me suggest another idea that achieves a similar goal to `lab-sensible-terminal`, but without shipping another executable - how about adding `OpenTerminal` (and `OpenBrowser`) action? This feels more minimalistic and avoids additional work for packagers, while deferring the detection of available terminal emulators.

How would those actions then find the "right" browser / terminal emulator?
My main goal is to make the initial configuration easier, once a user is able to modify any configs we are already past that goal.

ahesford:
I prefer a new script because it's a simple task in shell, but is much more cumbersome in C. There's no need to complicate the labwc source code to effect a shell loop.

tokyo4j:
> How would those actions then find the "right" browser / terminal emulator?
My main goal is to make the initial configuration easier, once a user is able to modify any configs we are already past that goal.

Just like `lab-sensible-terminal`, it tries running terminal emulators from an array.

Advantages over `lab-sensible-terminal` are:
- No need to ship another executable
- Works without `meson install` or registering the builddir to `$PATH` (but maybe this becomes pointless if we merge labnag in the next release...)

Advantages over detecting terminal emulators at startup are:
- We can defer the detection. If a terminal emulator becomes unavailable for some reason, we can select another available terminal
- We can actually try `execvp()`ing the terminal emulator and see it succeeds

> I prefer a new script because it's a simple task in shell, but is much more cumbersome in C. There's no need to complicate the labwc source code to effect a shell loop.

Hm, that's a good point. If @johanmalm and @Consolatis disagree with my idea, let's go for `lab-sensible-terminal`.

johanmalm:
> packager's cost of shipping another executable

I don't think it adds any overhead/cost for package maintainers.

Consolatis:
Closing in favor of
- #2877

PR #2844, opened 2025-06-17
theme: change builtin menu fallback button to arrow

johanmalm:
What do you think of changing the default menu button from this:

menu xml current

to this:

menu xml new

With `icon` being the default, I don't suspect most users would notice it.

I feel the downward arrow just looks nicer.


Consolatis:
No preference from my side. LGTM

Issue #2845, opened 2025-06-18
SEGFAULT when xdg_popup is destroyed before its parent?


tokyo4j:
### Problem description

When I run my (incomplete) sample program (https://github.com/tokyo4j/wayland-demo/tree/xdg-popup-labwc-crash) and kill it with Ctrl-C, the whole labwc session crashes with SEGFAUT.

Looks like `handle_destroy()` in `xdg-popup.c` is called after the parent `view` is destroyed, so accessing `popup->parent_view` becomes UAF. But I don't know what condition triggers this crash.

<details><summary>output from ASAN</summary>

```
=================================================================
==120052==ERROR: AddressSanitizer: heap-use-after-free on address 0x7cad27c20c80 at pc 0x56437791b99c bp 0x7ffe913504b0 sp 0x7ffe913504a0
READ of size 8 at 0x7cad27c20c80 thread T0
    #0 0x56437791b99b in handle_destroy ../src/xdg-popup.c:85
    #1 0x7f2d2dbd451d in wl_signal_emit_mutable (/usr/lib/libwayland-server.so.0+0x851d) (BuildId: 99ddf009eb5a6bbfb7d34377dd8c6a593ef0cd76)
    #2 0x7f2d2d34b42d  (/usr/lib/libwlroots-0.19.so+0x8442d) (BuildId: fe0111da6b6ed06c643a5b2102d2d6e0ee9d00e3)
    #3 0x7f2d2d34b735  (/usr/lib/libwlroots-0.19.so+0x84735) (BuildId: fe0111da6b6ed06c643a5b2102d2d6e0ee9d00e3)
    #4 0x7f2d2d34b9c6  (/usr/lib/libwlroots-0.19.so+0x849c6) (BuildId: fe0111da6b6ed06c643a5b2102d2d6e0ee9d00e3)
    #5 0x7f2d2d34d605  (/usr/lib/libwlroots-0.19.so+0x86605) (BuildId: fe0111da6b6ed06c643a5b2102d2d6e0ee9d00e3)
    #6 0x7f2d2dbd6bba  (/usr/lib/libwayland-server.so.0+0xabba) (BuildId: 99ddf009eb5a6bbfb7d34377dd8c6a593ef0cd76)
    #7 0x7f2d2dbd6dd1 in wl_client_destroy (/usr/lib/libwayland-server.so.0+0xadd1) (BuildId: 99ddf009eb5a6bbfb7d34377dd8c6a593ef0cd76)
[...]

PR #2846, opened 2025-06-18
Destroy xdg_popups when its parent is destroyed

Consolatis:
Fixes:
- #2845

----

Testing appreciated (especially with layershell popups). Didn't spot any issues but only tested it lightly.


johanmalm:
LGTM

Have verified that it fixes the crash.

Have done some testing with layershell client popups and they seem to work fine (qmpanel and gtk4-layer-demo)

PR #2848, opened 2025-06-19
[proof-of-concept] add --logfile

Consolatis:
Only very lightly tested.


Consolatis:
Well, I have no clue whats up with the static analyzer. It just seems wrong. This works: https://godbolt.org/z/vMMvqodzo and should be the exact same that is used in this PR. Maybe we should just add `-Wno-analyzer-fd-leak` to the CI. Unless I missed something. That `close(STDERR_FILENO)` at the end of the main function is pretty pointless and was only added to make the analyzer happy but it didn't.

PR #2857, opened 2025-06-25
ime: revert workarounds for wlroots 0.18.2

tokyo4j:
The workarounds added in #2498 and #2437 fixed stuck key/modifier bug caused by wlroots commit [e218990](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e21899037a97dcade40149f05ab54229c0018644). But now that the commit was reverted in 0.19, the workarounds are no longer needed.

Removing the workarounds also fixes a minor regression that pressing Ctrl+Enter in an input box in firefox causes stuck modifier.

I've been running this patch for a day and it's been working fine.


Consolatis:
I am not really familiar with this part of the codebase. I do trust your judgement though and didn't spot anything obvious in the diff so this LGTM.

PR #2860, opened 2025-06-26
Strictly use handle_ prefix for signal handlers

tokyo4j:


src/input/keyboard.c
```diff
@@ -358,7 +359,7 @@ get_keyinfo(struct wlr_keyboard *wlr_keyboard, uint32_t evdev_keycode)
 		&keyinfo.raw.syms);
 
 	/*
-	 * keyboard_key_notify() is called before keyboard_key_modifier(),
+	 * handle_key() is called before handle_modifier(),
```
Consolatis: `handle_modifiers()`

Consolatis (commented):
Small nitpick below, other than that LGTM. Not tested.

Issue #2858, opened 2025-06-25
UAF when an output with a panel is destroyed


tokyo4j:
### Problem description

#1154 seems to have caused an UAF when an output with a panel client is destroyed. Here's what's happening:
- `output->layer_tree[1]->node` (bottom layer) is destroyed in `output_destroy_notify()` in `output.c`
- `output->layer_tree[2]->node` (top layer) is destroyed in `output_destroy_notify()` in `output.c`
- `surface->scene_layer_surface->tree` for the top-layer panel is destroyed
- `handle_node_destroy()` in `layer.c` eventually calls `layers_arrange()` which accesses `output->layer_tree[1]`

### Steps to reproduce

1. Build labwc with `-Db_sanitize=address,undefined`
2. Launch nested labwc
3. Close the WL-1 window
4. Labwc is terminated due to UAF

### labwc build source

Local build

### labwc version

0.8.4-114-ga8df2487-dirty

### labwc environment

Nested on Wayland

Consolatis:
So how do we fix this?

If we want to keep arranging layer clients on node destroy the only thing I could think of is `output->layer_tree[x] = NULL` after destroying them and then checking for that in everything that iterates through the layers.

Consolatis:
Something like this:
- #2864

Completely untested so confirmation on fixing the issue welcome.

johanmalm:
Can try to re-produce issue to verify fix at weekend if nobody else gets there first.

PR #2864, opened 2025-06-26
layers: ignore destroyed layers

Consolatis:
Fixes:
- #2858


johanmalm:
LGTM. Not tested. 

Can't think of a better way to do it.

tokyo4j:
Confirmed this fixes the issue. But let's consider my alternative approach: #2867.

johanmalm:
Close in lieu of #2867 ?

PR #2868, opened 2025-06-27
Allow dragged views to be moved to other workspaces

Sumandora:
When configuring labwc for the first time I encountered a - not really a bug but - odd behavior, where if you drag a window and switch workspaces, the window will not carry over. I think that this is an objectively incorrect solution, because you end up dragging a window that you can't even see, and when you switch workspaces while dragging a window, you probably imagined something else to happen.
I initially wanted that behavior because it would allow me to save on the SendToDesktop buttons, which I had to put on an awkward mouse button, because they just didn't fit anywhere else. 
Anyways, I have written a patch, because I'm using Gentoo and I can apply them immediately without needing to do anything besides tell my pkg manager about it. Despite that I think this patch might be useful to others, which is why I'm now trying to upstream it. If there is no desire in merging this then I'll just keep using it locally. I must admit that I'm only semi-positive that there is no way of getting the behavior in mainline labwc, but I searched hard and also didn't find anything in the source code.

### DIFF
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -443,6 +443,11 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	/* Make sure new views will spawn on the new workspace */
 	server->workspaces.current = target;
 
+	struct view *grabbed_view = server->grabbed_view;
+	if(grabbed_view) {
+		view_move_to_workspace(grabbed_view, target);
+	}
+
 	/*
 	 * Make sure we are focusing what the user sees. Only refocus if
 	 * the focus is not already on an omnipresent or always-on-top view.
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -444,7 +444,7 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	server->workspaces.current = target;
 
 	struct view *grabbed_view = server->grabbed_view;
-	if (grabbed_view) {
+	if (grabbed_view && !view_is_always_on_top(grabbed_view)) {
 		view_move_to_workspace(grabbed_view, target);
 	}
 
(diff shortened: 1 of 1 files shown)


domo141:
I wonder how one drags and changes workspace at the same time, but sure if that happened the current
behavior (i.e. without this change) would be undesirable...

Sumandora:
> I wonder how one drags and changes workspace at the same time, but sure if that happened the current behavior (i.e. without this change) would be undesirable...

1. press super key + left mouse button
2. press button that is assigned to GoToDesktop
3. ????
4. profit!!

Didn't feel to adventurous :p

Consolatis:
Thanks for the PR!

> When configuring labwc for the first time I encountered a - not really a bug

I would consider this at least a UX bug.

I think we indeed need to implement one of these:
1. Prevent switching workspaces while moving a window around
2. Move the window to the new workspace when this happens

As the second option provides additional benefits I am fine with going with that route.

src/workspaces.c
```diff
@@ -443,6 +443,11 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	/* Make sure new views will spawn on the new workspace */
 	server->workspaces.current = target;
 
+	struct view *grabbed_view = server->grabbed_view;
+	if (grabbed_view) {
```
Consolatis: This also needs to check for always on top views (e.g. something like `grabbed_view && !view_is_always_on_top(grabbed_view)`. Otherwise always on top windows would loose their state and they should already be visible even when switching workspaces while moving them around.

Sumandora: Applied your suggestion

Consolatis:
Thanks, tested nested and works well.

Issue #2863, opened 2025-06-26
Coredump when switching to another tty


stefonarch:
### Problem description

I use to run an LXQt session with a different user on tty2 for testing purposes. I noticed now that labwc crashes when I switch back to tty7.
```
(gdb) where
#0  0x000074f14e22aded in wlr_scene_output_build_state () at /usr/lib/libwlroots-0.19.so
#1  0x00005aeef2f97985 in lab_wlr_scene_output_commit
    (scene_output=0x5aef0a691a80, state=0x5aef0a904e30)
    at ../src/common/scene-helpers.c:92
#2  0x00005aeef2f720ad in output_frame_notify
    (listener=0x5aef0a904fa0, data=0x5aef0a498c00) at ../src/output.c:152
#3  0x000074f14e2f151e in wl_signal_emit_mutable () at /usr/lib/libwayland-server.so.0
#4  0x000074f14e2f2efc in wl_event_loop_dispatch_idle ()
    at /usr/lib/libwayland-server.so.0
#5  0x000074f14e2f3177 in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#6  0x000074f14e2f51f7 in wl_display_run () at /usr/lib/libwayland-server.so.0
#7  0x00005aeef2f6f77a in main (argc=5, argv=0x7fffad688d48) at ../src/main.c:254
```

### Steps to reproduce

1. Launch `labwc -C /tmp` from tty
2. switch to another tty
3. Switch back.

The issue happens only with an external monitor attached to  the laptop.

### labwc build source

Local build

### labwc version

labwc 0.8.4-130-ge5a236ee

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch

Consolatis:
Thanks for the report! 

On a quick look at the offending lines I don't spot any obvious issue. I assume what is happening is that this is a duplicate of #2858 but as run without sanitizer we randomly overwrite stuff on the heap which then shows up like this.

This is just a theory though.

Consolatis:
Could you give
- #2864

a test?

stefonarch:
Done, still crashing but on console I got an output about `backend/drm/drm.c:1345 connector eDP-1 Failed to disable  CRTC67` and the same for HDMI-A-1 with  CRTC71.

I unplugged the external monitor and no crash, will test later with master.

stefonarch:
Updated the first posting. it happens only with dual monitor setup here.

[...]

PR #2874, opened 2025-06-29
layers.c: fix UAF bug on TTY change

johanmalm:
Call seat_set_focus_layer(seat, NULL) in node-destroy-handler to avoid seat->focused_layer becoming invalid and causing UAF issues in certain situations like when outputs (and therefore layer-trees) are destroyed.

Fixes: #2863


Consolatis:
:)

PR #2875, opened 2025-06-29
NEWS.md for the release

johanmalm:
Just letting this one develop over the next couple of weeks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog]
 
 | Date       | All Changes   | wlroots version | lines-of-code |
 |------------|---------------|-----------------|---------------|
-| 2025-06-24 | [unreleased]  | 0.19.0          | 28504         |
+| 2025-07-11 | [0.9.0]       | 0.19.0          | 28586         |
 | 2025-05-02 | [0.8.4]       | 0.18.2          | 27679         |
 | 2025-02-21 | [0.8.3]       | 0.18.2          | 27671         |
 | 2024-12-13 | [0.8.2]       | 0.18.2          | 26298         |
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -36,18 +36,28 @@ The format is based on [Keep a Changelog]
...
+  press-drag-release mouse action. For context: This is due to ambiguity in the
+  specifications and contrary implementations. For example, Gtk applications are
+  broken under KWin in this regard, while vice versa Qt clients are broken under
+  other compositors like Weston, Mutter and labwc. It has been decided not to
+  block the release due to this regression as it is an eco-system wide issue
+  that has existed for a long time. [#2787]
 
 ### Added
 
+- Enhance `-v|--version` option be adding feature flags like `+xwayland -rsvg`.
```
Consolatis: `s/be/by/`

Consolatis:
The dragLock comment should be updated, see https://github.com/labwc/labwc/pull/2876#issue-3188014183

There is also a comment at the bottom:
> TODO: Default source for icon name has changed with the support of `xdg-toplevel-icon` protocol

This is not true anymore, `xdg-toplevel-icon` / `_NET_WM_ICON` should  only be used when no icon has been found so I think we can simply remove that comment from the changelog.

NEWS.md
johanmalm: Thanks. Done.

johanmalm:
Thanks. All updated. I think we're "current" now.

NEWS.md
```diff
@@ -50,9 +50,13 @@ There is a regression warning worth noting for the switch to wlroots 0.19:
   other compositors like Weston, Mutter and labwc. It has been decided not to
   block the release due to this regression as it is an eco-system wide issue
   that has existed for a long time. [#2787]
+- All output leasing is disabled when building with `wlroots-0.19.0` due to a
+  wlroots use-after-free bug in that version [#2887]
```
Consolatis: Suggest something like this instead:

VR headset support is disabled when compiled with wlroots `0.19.0` to work around a bug on the wlroots side which is expected to be fixed in wlroots `0.19.1`.

johanmalm: Thanks. Done. Will squash all the fixup comments now.

johanmalm:
Ready for tomorrow from my perspective.

Consolatis:
Looking at https://github.com/johanmalm/labwc/blob/NEWS/NEWS.md the `[0.9.0]` link seems to be missing.
Other than that LGTM.

We could also modify the format slightly so the links in the top table go to the changelog items rather than to the compare links. And maybe add a small diff column which then contains the compare like https://github.com/labwc/labwc/compare/0.8.4...0.9.0. We don't have to do that for the release though, just something I noticed.

NEWS.md
```diff
@@ -136,21 +158,18 @@ window.*.title.bg.colorTo.splitTo:
     to use and this one results in frequent user complaints because it prevents
     some common usage patterns like alt-left/right in web browers.
 - Change default titlebar menu button from a dot to an arrow @johanmalm [#2844]
-- "Sticky" mode for drag-locking is enabled by default as recommended by
-  libinput [#2803]. The timeout based behavior can be restored via the snippet
-  below. To disable the feature all together use `no` as the value.
+- When `dragLock` is set to `yes`, `Sticky` mode is used by default as
```
tokyo4j: Sorry for missing this in the interim update: I guess "sticky" is unfamiliar to many users. I think we can instead say: "drag-lock no longer expires".

tokyo4j:

johanmalm:
@Consolatis Have fixed the `0.9.0` link. 

Regarding the refactoring so that the table at the top links to respective sections - I have thought about this too. Suggest we do that after release.

NEWS.md
johanmalm: Done.

PR #2876, opened 2025-06-30
rcxml: revise drag-lock behavior and option names

### LINKED ISSUE #2870: Change libinput dragLock default?
Quote from the wayland channel:
```
<emersion> whot: in the libinput release notes, you recommend switching the default to "sticky" for drag-lock
<emersion> https://lists.freedesktop.org/archives/wayland-devel/2024-November/043860.html
<emersion> is this a recommendation over the global default, or just the default when the user wants to enable drag-lock?
<emersion> more context: https://github.com/swaywm/sway/issues/8758#issuecomment-2989281883
<emersion> IOW: should sticky be the default if the user hasn't configured anything in particular, or only if the user wants to enable drag-lock?
<whot> emersion: recommendation is that if a user enables drag-locking it should be enabled as sticky because that's (probably) the more intuitive behavior
<whot> suggested default for drag lock: off, suggested default behavior if drag lock is on: sticky
```

(with whot being a libinput maintainer)
So should we default to `dragLock="no"` rather than `dragLock="sticky"`?

Ref: https://github.com/swaywm/sway/issues/8758#issuecomment-2989281883

tokyo4j:
Closes #2870

This revises the changes done in #2803:
- Cancel defaulting to `<dragLock>sticky<dragLock>`. So labwc now disables drag-lock by default, as libinput does.
- Update the options from `[yes|no|sticky]` to `[timeout|no|yes]` to emphasize that the `sticky` mode (now `yes`) is recommended when using drag-lock.

So the release note for 0.9.0 should include something like this:
> If you set `<dragLock>yes</dragLock>`, the drag will continue after you release your finger. Use `<dragLock>timeout</dragLock>` to cancel after a certain time as before.


Consolatis:
LGTM, not tested.

PR #2877, opened 2025-06-30
Add lab-sensible-terminal

johanmalm:
Related-to: #2843 #2842

How about this then


meson.build
```diff
@@ -198,6 +198,10 @@ install_data('data/labwc.desktop', install_dir: get_option('datadir') / 'wayland
 
 install_data('data/labwc-portals.conf', install_dir: get_option('datadir') / 'xdg-desktop-portal')
 
+# TODO: move this to clients/meson.build after the labnag PR
+clients = files('clients/lab-sensible-terminal')
+install_data(clients, install_dir: 'bin')
```
Consolatis: I have a feeling that hardcoding `bin` is wrong, there is likely something like `get_option(*bindir')` instead but I didn't check.

clients/lab-sensible-terminal
```diff
@@ -0,0 +1,50 @@
...
+	hyper \
+	wezterm \
+	rio"
+
+for terminal in $terminals
+do
+	if type "${terminal}" >/dev/null 2>&1
+	then
+		exec "$terminal" "$@"
+	fi
```
Consolatis: Even if the `exec` succeeds this might still not open a terminal (e.g. xterm on labwc without Xwayland support).
Suggest something like
```sh
if ..
then
	$terminal "$@" &
	wait %1 && exit 0
fi
```

Didn't check if this is portable sh code.

clients/lab-sensible-terminal
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
```
ahesford, line 3: It would be nice to support `$TERMINAL` and try that first:
```suggestion
if [ -n "${TERMINAL}" ]; then
	exec ${TERMINAL} "$@"
fi
```
Note that this shouldn't be quoted, so things like `TERMINAL="foot -T my-terminal"` work as expected.

meson.build
Narrat: Would be preferable if that is adjustable, yes :) Especially since some distros have completed the `/usr` merge.
And `get_option('bindir')` exists, but iirc needs to be combined with `get_option('prefix')`

clients/lab-sensible-terminal
```diff
@@ -0,0 +1,50 @@
...
+	terminix \
+	guake \
+	tilda \
+	hyper \
+	wezterm \
+	rio"
+
+for terminal in $terminals
+do
+	if type "${terminal}" >/dev/null 2>&1
```
ahesford: `type` is widely supported (including on `dash` and FreeBSD `sh`), but I don't believe it's POSIX compliant. Because we are falling back to `/bin/sh`, I recommand `command -v` in place of `type`.

Consolatis:
Should we also add it to the fallback root menu in this PR?

clients/lab-sensible-terminal
Narrat: Mostly curiosity. Why `type` instead of `command -v`? I see the latter more in checks if a binary exists.

meson.build
ahesford: No, `get_option('bindir')` should do the right thing with regard to prefix. Even `'bin'` will work regardless of usrmerge, but it ignores any `--bindir` argument passed to the meson configure.

clients/lab-sensible-terminal
ahesford: `exec` is preferable here because it avoids keeping around a useless parent shell, and will also confound signal handling in workflows that expect it (*e.g.*, cases were `lab-sensible-terminal` may be launched by a process supervisor).

The missing Xwayland case is probably the most likely cause for these issues (and is probably not a likely cause either way), so I would think moving X terminals to the end of the list is sufficient to remedy this.

meson.build
Narrat: It will work yes, but depending on the tooling to create packages it will complain about the install to a symlink location.
I just had recently a case, where `bindir` didn't work as expected and [this](https://github.com/dunst-project/dunst/issues/1462#issuecomment-2851363909) was the proposed solution by meson upstream. But not sure if that would also apply here. I'm not that knowledgable with the possebilities of `meson`. I suppose it came down to the string representation?

Edit: Having re-read that. Yes, was about wanting to have the string. Internally it knows. So `prefix` isn't needed for installing the binary.

clients/lab-sensible-terminal
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
```
Consolatis, line 1: This might need a License to make distros happy. Suggest something very broad like the Unlicence or MIT / BSD.

clients/lab-sensible-terminal
johanmalm: Because that's what git use.

https://github.com/git/git/blob/cf6f63ea6bf35173e02e18bdc6a4ba41288acff9/Documentation/CodingGuidelines#L87

clients/lab-sensible-terminal
ahesford: I would assume that the project license would apply here. If it's in the labwc repository and doesn't specify a license of its own, `LICENSE` at the project root should be inferred.

clients/lab-sensible-terminal
Narrat: Interesting. Thanks for the link.
In the end it will come down to being POSIX compliant or not then :)

meson.build
johanmalm: Thanks. Have implemented. Seems to work fine.

clients/lab-sensible-terminal
johanmalm: It's a very good thought about making this work when XWayland is not suppoted.

The code above looks awkward to me.

As an alternative, how about we put all the X11 ones last in `terminals` string?

johanmalm:
> Should we also add it to the fallback root menu in this PR?

Ok. Will have to be tomorrow though :smile:

clients/lab-sensible-terminal
johanmalm: Done. Good idea.

clients/lab-sensible-terminal
johanmalm: I didn't know that. Have just got stuck on `type` over the years :smile:

Have changed to `command -v`

johanmalm: Well, let's go with POSIX compliance. POSIX compliance is not everything to me, but we ought to have a reason for not choosing it - and I can't see one with this.

clients/lab-sensible-terminal
johanmalm: I think I prefer sticking with GPL2 to avoid having multiple licenses. What would be the arguments for MIT/BSD?

ahesford (approved):
This seems good to me. I have no objection to updating the root menu to using it.

clients/lab-sensible-terminal
johanmalm: Sorry @ahesford - didn't refresh my browser to see that you'd already said the same thing!

Consolatis: It looses the ability to prioritize some of those X11 terminal emulators before wayland supporting ones but that feels acceptable to me.

clients/lab-sensible-terminal
Consolatis: Might make re-using / forking the script easier. Although now thinking of it, this is not a library or anything like that so using something broad vs `GPL-2.0-only` shouldn't really make a difference in this case.

domo141:
I'd check `$WAYLAND_DISPLAY` & `$DSPLAY` -- quick & lazy skeleton:

```
#!/bin/sh

set -euf

test "${TERMINAL-}" && exec $TERMINAL "$@"

if test "${WAYLAND_DISPLAY-}"
then
         terminals='list-of-wayland-terminals'

         for t in $terminals
         do command -v $t >/dev/null && exec $t "$@"
         done
fi

if test "${DISPLAY-}"
then
     terminals='list-of-x11-terminals'

     for t in $terminals
     do command -v $t >/dev/null && exec $t "$@"
     done
fi

exec labnag no terminal or something...
```

clients/lab-sensible-terminal
domo141: fyi:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html

lists `command` (but not `type`) as (builtin) commands. that also lists `hash` which
some has used to check if command exists -- just that `ksh` returns `0` even if it did
not find the command which path it wanted to remember...

clients/lab-sensible-terminal
```diff
@@ -0,0 +1,56 @@
...
+for terminal in $terminals
+do
+	if command -v "${terminal}" >/dev/null 2>&1
+	then
+		exec "$terminal" "$@"
+	fi
+done
+
+# TODO: uncomment after labnag PR
+#labnag -m 'lab-sensible-terminal could not find a terminal emulator' -Z X :
```
tokyo4j: Instead of spawning an error window by the executable itself, wouldn't it be more general to spawn the error window by labwc when `Execute` action fails? AFAIR it follows Openbox's behavior.

But of course, I know these are just comments and I don't mind keeping them in this PR.

tokyo4j:
LGTM.

clients/lab-sensible-terminal
ahesford: I agree, just `exit 1` at the end seems like a good option to trigger a failure message handled more generically.

clients/lab-sensible-terminal
johanmalm: Okay. Have put X11 terminals at end of list.

clients/lab-sensible-terminal
johanmalm: Have added SPDX GPL-2.0-only identifier in file.

johanmalm:
> I'd check `$WAYLAND_DISPLAY` & `$DSPLAY` -- quick & lazy skeleton:

I don't think there is a consistent way to check - but might be wrong on this.

clients/lab-sensible-terminal
johanmalm: Have added `exit 1` instead of the labnag comment.

src/menu/menu.c
```diff
@@ -1030,6 +1030,13 @@ init_rootmenu(struct server *server)
 	if (!menu) {
 		current_menu = NULL;
 		menu = menu_create(server, "root-menu", "");
+
+		current_item = item_create(menu, _("Terminal"), false);
+		fill_item("name.action", "Execute");
+		fill_item("command.action", "lab-sensible-terminal");
```
Consolatis, line 1036: I think we need to do [some things](https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#coders) here when we add a new translation item.

Consolatis, line 1036: Should we also put that into the example `menu.xml` for the root menu? I am not sure on the right approach, with `lab-sensible-terminal` now expected to be available it might be fine I guess. We could then also use the chance to sync it with our actual hardcoded defaults like we do with `rc.xml.all`.

johanmalm: Have updated po/labwc.pot

johanmalm: Have updated menu.xml

Consolatis:
LGTM, have tested with `-C /dev/null` and works well. Thanks!

johanmalm:
Thanks for "approving review" and LGTM thus far. Considering the amount of change though, I feel this ought to have another final review.

I think I've responded to all suggestions.

This has been rebased and is ready to go from my perspective.

Thanks for all the help in thinking this through :smile:

johanmalm:
> LGTM, have tested with `-C /dev/null` and works well. Thanks!

Like passing ships at night. Thanks.

keithbowes:
This is weird. In my RPM SPEC file, I recommend [xdg-terminal-exec](https://github.com/Vladimir-csp/xdg-terminal-exec), which is a better solution than all these *-sensible-terminal scripts (in particular, it lets users set the order and the command-line arguments, and provides a consistent interface between terminals, which the terminals themselves lack). I can understand older things having these scripts, but it seems weird to create new ones where better solutions already exist. (GTK uses xdg-terminal-exec when launching a terminal program, so it's not like using it is unprecedented.)

johanmalm:
> This is weird. In my RPM SPEC file, I recommend [xdg-terminal-exec](https://github.com/Vladimir-csp/xdg-terminal-exec), which is a better solution than all these *-sensible-terminal scripts (in particular, it lets users set the order and the command-line arguments, and provides a consistent interface between terminals, which the terminals themselves lack). I can understand older things having these scripts, but it seems weird to create new ones where better solutions already exist. (GTK uses xdg-terminal-exec when launching a terminal program, so it's not like using it is unprecedented.)

Yes, I think those types of scripts are better solutions in a generic sense.

We are trying to strike a balance here though whereby the compositor remains an independent system tool without further dependencies (like xdt-terminal-exec or whatever), but where we just become a little more user friendly for people who are trying to get started.

I do not imagine that anyone seriously looking to use `labwc` as their daily drivers will carry out using the default root-menu.

Hope that makes sense.

PR #2882, opened 2025-07-02
Added information about idle screen inhibitor.

Kyuyrii:
I would like you to add this mention about sway-audio-idle-inhibit, it is an idle screen inhibitor that prevents the screen from turning off based on audio, I tested it on Labwc and it worked very well.


Consolatis:
LGTM

johanmalm:
Nice.

PR #2883, opened 2025-07-02
actions: added query tiled=any comparison for rc.xml simplification

1ynxy:
Following up from discussion #2869, this converts the view_edge enum to bitfields (similar to view_axis enum above), and adds a VIEW_EDGE_ANY value which combines all valid options.

This allows the user to query for any snap directions without using multiple query statements, ie:

```xml
<keybind key="W-Down">
  <action name="If">
    <query tiled="any" />
    <then>
      <action name="UnSnap" />
    </then>
    <else>
      <action name="Iconify" />
    </else>
  </action>
</keybind>
```

This has been compiled and tested locally with no issues or aberrant behaviour.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -474,7 +474,7 @@ Actions that execute other actions. Used in keyboard/mouse bindings.
 			The "left" , "right", "left-occupied" and
 			"right-occupied" directions will not wrap.
 
-		*tiled* [up|right|down|left|center]
+		*tiled* [up|right|down|left|center|any]
 			Whether the client is tiled (snapped) along the the
 			indicated screen edge.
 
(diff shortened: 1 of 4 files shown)


Consolatis:
Thanks for the PR!

LGTM.

As we are in cooldown until the 0.9.0 release I think this needs to wait a bit though. (Although personally I wouldn't mind making an exception as this is unlikely to introduce any issues but that is up to @johanmalm to decide).

1ynxy:
Following discussion with Consolatis, I have added checks to prevent users providing the "any" value on "direction" properties for the following actions:

- ToggleSnapToEdge
- SnapToEdge
- GrowToEdge
- ShrinkToEdge
- FocusOutput
- MoveToOutput

Tested behaviour without such restrictions did not result in any errors- in most instances "any" functioned like "center" does- but if we can warn the user that this is not a valid input then we probably should.

johanmalm:
LGTM.

I prefer to honour the cool down period, so let's merge after `0.9.0` (probably Friday)

tokyo4j:
LGTM

PR #2884, opened 2025-07-04
Miscellaneous cleanups of discouraged/obscure C language features

jlindgren90:
Background: I was doing an experiment compiling labwc as C++. (I know we've discussed this before, and I'm not pushing a transition -- it's just a personal experiment.) But in the process, I found some usages of discouraged/obscure C language features, which are forbidden and produce errors in C++.

Some of the fixes seemed like they would be good cleanups even in C, so I cherry-picked those to this branch. Roughly in order of most- to least-important (in my judgment).

Please feel free to take or leave any of them. Or take some now and leave the others till after the release, if that makes sense.

### DIFF
--- a/include/common/string-helpers.h
+++ b/include/common/string-helpers.h
@@ -59,8 +59,7 @@ char *strdup_printf(const char *fmt, ...);
  * The separator is arbitrary. When the separator is NULL, a single space will
  * be used.
  */
-char *str_join(const char *const parts[],
-	const char *restrict fmt, const char *restrict sep);
+char *str_join(const char *const parts[], const char *fmt, const char *sep);
 
 /**
  * str_endswith - indicate whether a string ends with a given suffix
--- a/include/foreign-toplevel-internal.h
+++ b/include/foreign-toplevel-internal.h
@@ -6,62 +6,63 @@
 #include <wayland-server-core.h>
 #include "foreign-toplevel.h"
 
+struct wlr_foreign_toplevel {
+	struct wlr_foreign_toplevel_handle_v1 *handle;
+
+	/* Client side events */
+	struct {
+		struct wl_listener request_maximize;
+		struct wl_listener request_minimize;
+		struct wl_listener request_fullscreen;
+		struct wl_listener request_activate;
+		struct wl_listener request_close;
+		struct wl_listener handle_destroy;
+	} on;
+
+	/* Compositor side state updates */
+	struct {
+		struct wl_listener new_app_id;
+		struct wl_listener new_title;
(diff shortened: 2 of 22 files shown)


tokyo4j:
- b1923c3: Maybe you can just move `string_truncate_at_pattern()` to inside `entry()` rather than allocating another temporary memory for the nodename?
- 134f128: LGTM
- 46284f1: LGTM
- 0982820: LGTM
- 7ce4113: LGTM
- 57d9c30: LGTM
- 89d32aa: My preference is different. I like how those inline struct declaration allows me to see its nested fields without scrolling to outside of the parent struct declaration. But let's see how other maintainers view this.
- db974fe: LGTM. I think those forward variable declarations are sometimes useful to reduce the diff as much as possible when defining a new function, but I think they should be cleaned up eventually, like done in this PR.
- 1ed392e: LGTM
- 959d26a: LGTM
- f71474a: LGTM

jlindgren90:
> * [b1923c3](https://github.com/labwc/labwc/commit/b1923c34f38e25098111e9be3218e8071df00e98): Maybe you can just move `string_truncate_at_pattern()` to inside `entry()` rather than allocating another temporary memory for the nodename?

Very nice idea; I didn't look closely enough to see that. I pushed a [fixup] commit just now.

> * [89d32aa](https://github.com/labwc/labwc/commit/89d32aa6af9a759c7ee27f2e02bc9eb30dec21b8): My preference is different. I like how those inline struct declaration allows me to see its nested fields without scrolling to outside of the parent struct declaration. But let's see how other maintainers view this.

I'm happy to leave this one out if there isn't consensus on it.

Consolatis:
https://github.com/labwc/labwc/pull/2884/commits/098282074c9f5916fcc66d485947ba71493150e6: Con argument: generally trying to prevent `goto`s will result in a more pyramid style of code because it adds more branches and thus nested scopes (or duplicated clean-up code when returning instead). This specific commit looks fine but I don't really see a benefit.

https://github.com/labwc/labwc/commit/89d32aa6af9a759c7ee27f2e02bc9eb30dec21b8: My preference regarding this is different as well, I prefer to see all struct members in the same struct if possible. If we need a pointer to a single substruct we can also name the inline struct.

The others look fine to me. I'd suggest merging after the release though because it adds potential issues without any benefit for the user.

johanmalm:
Thanks. Okay, let's do the final review + merge after `0.9.0`.

jlindgren90:
Rebased, omitting the struct re-org commit that was unpopular. Should be ready for final review now.

Consolatis:
Thanks :)

PR #2885, opened 2025-07-04
Support <snapping><cornerRange>

### LINKED ISSUE #1758: Allow snapping a window to screen corners
My idea is adding a setting like `<snapping cornerRange="10%">` and allow snapping a window to corners if the cursor is in the red region of outputs in the following figure. EDIT: In #2885, I decided to use just pixel-based values for consistency with `<snapping><range>`.

snap-corner

tokyo4j:
Closes #1758.

By default, `<snapping><cornerRange>` is set to 50(px).

`up-left`/`up-right`/`bottom-left`/`bottom-right` are added for `direction` argument of `SnapToEdge`/`ToggleSnapToEdge` actions.

Additionally, this PR changes the edge-snapping overlay to take into account `<core><gap>`, as a result of deduplicating `get_edge_snap_box()` in `interactive.c` and `view_get_edge_snap_box()` in `view.c`. But as this makes the region/edge snapping overlays inconsistent, I'm planning to open another sub-PR to both snapping overlays to take into account `<core><gap>`.

Documentation is also lacking.

### DIFF
--- a/include/common/direction.h
+++ b/include/common/direction.h
@@ -2,9 +2,10 @@
 #ifndef LABWC_DIRECTION_H
 #define LABWC_DIRECTION_H
 
+#include <wlr/types/wlr_output_layout.h>
 #include "view.h"
 
-enum wlr_direction direction_from_view_edge(enum view_edge edge);
+bool direction_from_view_edge(enum view_edge edge, enum wlr_direction *dir);
 enum wlr_direction direction_get_opposite(enum wlr_direction direction);
 
 #endif /* LABWC_DIRECTION_H */
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -145,6 +145,7 @@ struct rcxml {
 
 	/* window snapping */
 	int snap_edge_range;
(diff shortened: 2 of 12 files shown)


johanmalm:
Nice. Assume this is for review + merge after `0.9.0`?

tokyo4j:
> Nice. Assume this is for review + merge after 0.9.0?

Yeah, this is for post-0.9.0 of course. Please review this PR after I mark this PR as ready for review. I have some more work to do as I described above.

tokyo4j:

The tricky part in this PR is that `edge_from_cursor()` now returns `edge1` and `edge2`. `edge1` is the edge determined by the cursor position and `<snapping><range>` and used for `<inner>`/`<outer>` detection of `<snapping><overlay><delay>`. And `edge2` is the edge determined by the cursor position and `<snapping><cornerRange>`. The final edge to snap a window to can be obtained as `edge1 | edge2`.

johanmalm:
LGTM. Have tested.

I think that the `edge{1,2}` is fine.

PR #2886, opened 2025-07-04
resistance: refactor snap constraints, use in interactive resistance

ahesford:
With negative screen or window resistance (attraction), snapping can cause "flickering" in clients (like `foot`) that enforce constraints on sizes. As labwc tries to snap a window to an edge, the client may substitute its own size that does not quite reach the edge; this can leave the cursor in a position that, when it moves slightly, will cause the client to bounce rapidly between a size that puts its resize edge near the cursor and one that puts its resize edge near the snapping edge.

A similar issue appeared earlier with "grow to edge" or "shrink to edge" operations, which was handled by caching the "desired" size set by labwc and using that for subsequent edge checks. When the client overrules a configure request with a different size, we still remember our "intended" size, and use that for subsequent edge checks.

The same cache and logic can be applied to interactive resizes. However, because interactive resizes can act on both a horizontal and a vertical edge, the cache has been extended to apply to either orientation or both simultaneously.

I have tested these changes briefly, but we will need more extensive validation that existing (grow or shrink)-to-edge operations have not been broken, and that both positive and negative resistance values behave intuitively for interactive resizes.

### DIFF
--- a/include/snap-constraints.h
+++ b/include/snap-constraints.h
@@ -2,19 +2,21 @@
 #ifndef LABWC_SNAP_CONSTRAINTS_H
 #define LABWC_SNAP_CONSTRAINTS_H
 
+#include <wlr/util/edges.h>
+
 #include "common/border.h"
 #include "view.h"
 
 struct wlr_box;
 
 void snap_constraints_set(struct view *view,
-	enum view_edge direction, struct wlr_box geom);
+	enum wlr_edges direction, struct wlr_box geom);
 
 void snap_constraints_invalidate(struct view *view);
 
 void snap_constraints_update(struct view *view);
 
 struct wlr_box snap_constraints_effective(struct view *view,
-	enum view_edge direction);
+	enum wlr_edges direction, bool use_pending);
 
 #endif /* LABWC_SNAP_CONSTRAINTS_H */
--- a/src/resistance.c
+++ b/src/resistance.c
@@ -7,6 +7,7 @@
 #include "edges.h"
 #include "labwc.h"
 #include "resistance.h"
+#include "snap-constraints.h"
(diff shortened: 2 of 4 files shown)


ahesford:
It seems using the snap-constraint cache was not sufficient. The cache records the pending client size when a snap is determined, and an XDG client is allowed to update the pending size (at most once) in case the client has substituted its own width or height during a configure. The pending size is used to validate the cache when determining the "effective" geometry for snapping calculations during interactive or action-based resizes.

Originally, `snap_constraints_update` checked that the cached geometry matched the view's pending geometry exactly, or it would disallow the geometry update. This was a sanity check to ensure that the cache wasn't being corrupted should the view size have somehow drifted from what we expected based on the snapping calculations. During interactive resizing, this check may fail even when it should succeed, because the (x, y) coordinates in the view's pending geometry may differ from the cached geometry by one pixel (along either, or both, axes) even if the widths and heights match. Thus, the cache is never updated with client overrides, and validation fails.

In retrospect, checking the cached geometry against the pending geometry during a cache update is too strict. The client will only ever override the width or height of the geometry, which means our sanity check for cache updates can be limited to comparing the widths and heights of the pending view geometry and that cached during snapping calculations. The result is that `snap_constraints_update` will succeed when the client overrides our expected width and height with its own, and we will use the cached "effective" geometry for snapping calculations as we should.

With the second commit applied, there should be no flicker when dragging windows in the vicinity of attractive window or screen edges.

Consolatis:
Have tested nested, works great. This solves constant resizing with a window edge strength of `-50` for me.

johanmalm:
LGTM. Have tested and seems to work well.

PR #2887, opened 2025-07-04
Disable drm leasing for wlroots 0.19 due to UAF

Consolatis:
First commit fixes a labwc UAF caused by wlroots now destroying the `wlr_output` when granting a drm lease request, see commit body for details. This also removes the `output->leased` bool which is not required anymore.

Second commit disables drm leasing completely because of a wlroots UAF in the same code path ([wlroots fix](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5104)).


tokyo4j:
For the second commit, I think we can just not create `server->drm_lease_manager` if wlroots is below 0.19.1 so that clients are aware that drm-leasing is not supported.

Consolatis:
> For the second commit, I think we can just not create `server->drm_lease_manager` if wlroots is below 0.19.1 so that clients are aware that drm-leasing is not supported.

Shouldn't really make a difference in practice and the comment looked like a good place to extend. In `server.c` it would be a bit more complicated because it would then also trigger the log message about not being supported and thus VR not being available.

tokyo4j:
> Shouldn't really make a difference in practice and the comment looked like a good place to extend. In server.c it would be a bit more complicated because it would then also trigger the log message about not being supported and thus VR not being available.

OK, I'm fine with that. Not exposing drm-lease global feels slightly more straightforward, but it will be fine as `wlr_drm_lease_v1_manager_offer_output()` is also not called in nested wayland session.

johanmalm:
LGTM

tokyo4j:
I confirmed that this PR prevents UAF when I run https://github.com/JoseExposito/drm-lease-simple-client.

It's a bit sad that upgrading labwc to 0.9.0 removes a non-negligible feature, but it's a lot better than crashing. I hope distro packagers recompile labwc as soon as new wlroots version is released.

johanmalm:
> I hope distro packagers recompile labwc as soon as new wlroots version is released.

Re-build shouldn't be required though, should it?

Consolatis:
> > I hope distro packagers recompile labwc as soon as new wlroots version is released.
> 
> Re-build shouldn't be required though, should it?

It would be required because the wlroots version we check against comes from `<wlr/version.h>` and thus is set at labwc compile-time.

tokyo4j:
> Re-build shouldn't be required though, should it?

The detection of wlroots version in this PR is based `WLR_VERSION_NUM` macro and done at compile time. So when wlroots 0.19.1 is released, packagers can restore the drm-leasing feature by recompiling labwc with the new wlroots header.

tokyo4j:
This should help: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5106

johanmalm:
> It would be required because the wlroots version we check against comes from `<wlr/version.h>` and thus is set at labwc compile-time.

Got it.

Do we think that's okay then, or is it likely to trip up any distros or package maintainers?

Consolatis:
> Do we think that's okay then, or is it likely to trip up any distros or package maintainers?

I think its fine and I would assume that most distros will recompile if a dependency changes. And even if not, the next labwc release will fix it automatically. Also DRM leasing is an extremely niche feature. For the labwc user base at least, I do not remember a single issue being raised about that here, like ever.

PR #2888, opened 2025-07-04
src/output.c: remove wlroots 0.18 wlopm with custom-mode workaround

Consolatis:
The [fix](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3946) has been included in wlroots 0.19.0.


johanmalm:
Well remembered :smile:
Thanks.

PR #2892, opened 2025-07-05
 Update for wlroots 0.20

Consolatis:
Replaced by
- #2956

---

You know the drill, not intended to be merged anytime soon.
Mainly to ease incremental chase updates so we don't end up with a giant pile of work when we start tracking 0.20.

Not suggested to use as daily driver but pointing out issues in this PR is obviously welcome. The last commit always tracks the wlroots master branch so issues are expected when wlroots breaks API. List of breaking changes from wlroots is [here](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3977).

This PR may also serve as a base for other PRs to test new wlroots things like toplevel capture.

Known defects as of today:
- [ ] remove erroneous IME commit, I mixed up the `input_method` with the `text_input`.

Push policy: feel free to push commits to this PR but please don't force-push, I'll handle the rebasing.

PR #2894, opened 2025-07-06
ime: don't use data argument in handle_input_method_commit()

tokyo4j:
Preparation for #2892.

I'm pretty sure that this doesn't cause any regressions, but feel free to delay merging this as this doesn't change any functionalities.

### DIFF
--- a/src/input/ime.c
+++ b/src/input/ime.c
@@ -265,8 +265,7 @@ handle_input_method_commit(struct wl_listener *listener, void *data)
 {
 	struct input_method_relay *relay =
 		wl_container_of(listener, relay, input_method_commit);
-	struct wlr_input_method_v2 *input_method = data;
-	assert(relay->input_method == input_method);
+	struct wlr_input_method_v2 *input_method = relay->input_method;
 
 	struct text_input *text_input = relay->active_text_input;
 	if (!text_input) {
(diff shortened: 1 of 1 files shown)


Consolatis:
I just checked the [changes](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/2d5492c73770c9de420527df1098fefabe43d689) again and the change is about `struct wlr_text_input_v3` (e.g. the `handle_text_input_xxx` handlers). Those don't seem to use the `data` argument from a quick glance.

I'll remove the temporary commit from the chase PR and delete it from the top post. Sorry for the noise.

tokyo4j:
Hm, it was confusing that only `wlr_text_input_v3` interfaces have been refactored and `wlr_input_method_v2` have not.

I think `NULL` should be emitted in `wlr_input_method_v2` signals as well and this PR is still the way to go. Opened https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5107.

Consolatis:
In that case there is an `assert(relay->input_method == data);` missing in `handle_input_method_destroy()`.

tokyo4j:
> In that case there is an assert(relay->input_method == data); missing in handle_input_method_destroy().

Nice catch. I'll remove that line.

johanmalm:
Thanks. Let's merge after `0.9.0`

Consolatis:
LGTM

PR #2895, opened 2025-07-06
RFC: windowswitcher: show 's' as "state" for shaded views

domo141:
While at it sorted the code to show 'm' before 's' and 's' before 'M'
\-- from the least visible to most visible state.

RFC NOTE 1: docs not (yet) updated

RFC NOTE 2: "hacked" the field_set_win_state_all() to show either 'F' or 's', as F and s cannot be (feasibly) be active the same time. This is inconsistent - just used to make 's' seen if one uses only 'S' in custom="..." format and not content="state" at all...

### DIFF
--- a/src/osd-field.c
+++ b/src/osd-field.c
@@ -125,10 +125,12 @@ static void
 field_set_win_state(struct buf *buf, struct view *view, const char *format)
 {
 	/* custom type conversion-specifier: s */
-	if (view->maximized) {
-		buf_add(buf, "M");
-	} else if (view->minimized) {
+	if (view->minimized) {
 		buf_add(buf, "m");
+	} else if (view->shaded) {
+		buf_add(buf, "s");
+	} else if (view->maximized) {
+		buf_add(buf, "M");
 	} else if (view->fullscreen) {
 		buf_add(buf, "F");
 	} else {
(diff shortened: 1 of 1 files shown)


src/osd-field.c
```diff
@@ -142,7 +144,9 @@ field_set_win_state_all(struct buf *buf, struct view *view, const char *format)
 	/* custom type conversion-specifier: S */
 	buf_add(buf, view->minimized ? "m" : " ");
 	buf_add(buf, view->maximized ? "M" : " ");
-	buf_add(buf, view->fullscreen ? "F" : " ");
+	// XXX WIP -- "s" should be handled separately - consumes space in
+	// XXX custom field -- harder is to choose order...
+	buf_add(buf, view->fullscreen ? "F" : view->shaded ? "s" : " ");
```
Consolatis: I think `s` should just be a usual field here. The difference between the two `set_win_state()` fields is that one only shows a single character indicating state while the other shows all states. Suddenly having fullscreen and shaded share a single character while all other states use a dedicated one sounds incoherent to me.

domo141: Yes - and if I had realized that the windowswitcher can show *many* clients, one with 'F' and another with 's'
(which probably looks pretty awful along with other clients) ;/ I would not have pushed such a version for
(preliminary) (re)view...

domo141:
made the change to separate 's' from 'F' in the state of window (all states) view
- in there just added 's' between 'm' and 'M'

in case of short form state of window field, it is better to see m/s if view is minimized/shaded
instead of 'M'.

Consolatis:
Thanks :)

Should be mentioned in the changelog with something like
> The custom window switcher format modifier `S` now uses 4 characters instead of 3.

@johanmalm I've applied the label `Release Notification` for things to be mentioned in the `Changed` section in NEWS.md. When done you can remove the label. That way we can keep track of those changes. https://github.com/labwc/labwc/issues?q=label%3A%22Release%20Notification%22 should thus always show all pending ones.

PR #2897, opened 2025-07-07
Add a way to replicate KDE Window Control Setups

CosmicFusion:
In KDE you can use KDE Window Control to hide SSDs and merge the window controls in the top panel, i think it works like this in MacOS too, i added a simple change to allow doing this in labwc

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -186,6 +186,11 @@ this is for compatibility with Openbox.
 	that it is not always possible to turn off client side decorations.
 	Default is server.
 
+*<core><disableMaximizedServerDecor>*
+	Hide server side decoration when a window is maximized.
+	Useful for trying to replicate a KDE window controls setup where window controls
+	merge with the top panel
+
 *<core><gap>*
 	The distance in pixels between windows and output edges when using
 	movement actions, for example MoveToEdge. Default is 0.
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1162,6 +1162,8 @@ entry(xmlNode *node, char *nodename, char *content, struct parser_state *state)
 		rc.gap = atoi(content);
 	} else if (!strcasecmp(nodename, "adaptiveSync.core")) {
 		set_adaptive_sync_mode(content, &rc.adaptive_sync);
+	} else if (!strcasecmp(nodename, "disableMaximizedServerDecor.core")) {
+		set_bool(content, &rc.disable_maximized_ssd_decor);
 	} else if (!strcasecmp(nodename, "allowTearing.core")) {
 		set_tearing_mode(content, &rc.allow_tearing);
 	} else if (!strcasecmp(nodename, "autoEnableOutputs.core")) {
--- a/src/ssd/ssd.c
+++ b/src/ssd/ssd.c
@@ -30,7 +30,7 @@ ssd_thickness(struct view *view)
 	 * in border-only deco mode as view->ssd would only be set
 	 * after ssd_create() returns.
 	 */
-	if (!view->ssd_enabled || view->fullscreen) {
+	if (!view->ssd_enabled || view->fullscreen || (view->maximized && rc.disable_maximized_ssd_decor)) {
 		return (struct border){ 0 };
 	}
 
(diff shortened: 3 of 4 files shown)


docs/labwc-config.5.scd
```diff
@@ -186,6 +186,11 @@ this is for compatibility with Openbox.
 	that it is not always possible to turn off client side decorations.
 	Default is server.
 
+*<core><disableMaximizedServerDecor>*
+	Hide server side decoration when a window is maximized.
+	Useful for trying to replicate a KDE window controls setup where window controls
+	merge with the top panel
```
Consolatis: Default value missing.

src/config/rcxml.c
```diff
@@ -1162,6 +1162,8 @@ entry(xmlNode *node, char *nodename, char *content, struct parser_state *state)
 		rc.gap = atoi(content);
 	} else if (!strcasecmp(nodename, "adaptiveSync.core")) {
 		set_adaptive_sync_mode(content, &rc.adaptive_sync);
+	} else if (!strcasecmp(nodename, "disableMaximizedServerDecor.core")) {
+		set_bool(content, &rc.disable_maximized_ssd_decor);
```
Consolatis, line 1166: `rc.disable_maximized_ssd_decor` needs to be set to `false` in `rcxml_init()`, otherwise the following cycle fails: `setting it to true -> Reconfigure -> removing the config -> Reconfigure`.

src/ssd/ssd.c
```diff
@@ -30,7 +30,11 @@ ssd_thickness(struct view *view)
 	 * in border-only deco mode as view->ssd would only be set
 	 * after ssd_create() returns.
 	 */
-	if (!view->ssd_enabled || view->fullscreen) {
+	if (
+		!view->ssd_enabled ||
+		view->fullscreen ||
+		(view->maximized && rc.disable_maximized_ssd_decor)
+	) {
```
Consolatis: Please use something like 
```c
if (!view->ssd_enabled || view->fullscreen
		|| (view->maximized && rc.disable_maximized_ssd_decor)) {
	[..]
}
```

CosmicFusion:
> Thanks for the PR.
> 
❤️
> In general I am not sure if this catches all edge cases and if this is something we want to support in labwc in the first place. I don't see it causing any harm or really add complexity though so I don't really have a preference in one way or the other.
> 
It Deals with SSD to create the functionality that KDE like layout needs, for GTK3/4 CSD a simple CSS hack in your current theme is needed to make it follow the same behavior.

I think it's good enough, and thank you for considering merging it!

> Some small things below.

Hopefully resolved now!

johanmalm:
I like it.

IIRC we have had this as a feature request before, so suspect there is demand for it.

We're in a cool down period this week, so suggest we go for final review with a view to merge after the release next week.

tokyo4j:
You also need to disable the titlebar on maximize, which I think is the most complicated part for this change.

johanmalm:
> You also need to disable the titlebar on maximize, which I think is the most complicated part for this change.

@tokyo4j The PR already does this. Are you happy to proceed?

johanmalm:
The commits need to be squashed, but we could do that on merge.

tokyo4j:
> > You also need to disable the titlebar on maximize, which I think is the most complicated part for this change.
> 
> @tokyo4j The PR already does this. Are you happy to proceed?

Sorry, my sentences didn't make sense. I meant the titlebar is still visible outside of usable area:

20250821_13h03m21s_grim

After some tinkering, I found the following patch works (so it's not complicated as I thought):

```diff
diff --git a/src/ssd/ssd-titlebar.c b/src/ssd/ssd-titlebar.c
index d44adf41..cf4ce42b 100644
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -300,6 +300,9 @@ ssd_titlebar_update(struct ssd *ssd)
 		ssd->state.was_omnipresent = view->visible_on_all_workspaces;
 	}
 
+	wlr_scene_node_set_enabled(&ssd->titlebar.tree->node,
+		!(maximized && rc.disable_maximized_ssd_decor));
+
 	if (width == ssd->state.geometry.width) {
 		return;
 	}
```

EDIT: Opened  #3015 for a better solution

tokyo4j:
Superseded by #3015.

Issue #2880, opened 2025-07-02
how to trigger HideCursor action without restarting swayidle's clock?


bdantas:
Triggering HideCursor action with a keyboard shortcut causes idle clock to reset, as I describe here: https://github.com/labwc/labwc/pull/2633#issuecomment-3027635612

Is there some way to trigger the HideCursor action without a keyboard shortcut (i.e., without restarting system's idle clock)?

Consolatis:
I wonder what your use-case is, hiding the cursor after some timeout when not moved in something like a terminal?

bdantas:
> I wonder what your use-case is
 
Home-brewed family media player. I'd like to hide the cursor after 5 seconds of idle, start screensaver (if there is no sound output) after 60 seconds of idle.

Because of this little problem, swayidle wakes up the system every time it sends the keyboard shortcut to hide the cursor.

Consolatis:
Wayland applications can hide the cursor themselves, so if the media player used is a wayland native application running in fullscreen it could hide the cursor automatically after some timeout.

bdantas:
Hi Consolatis. Yes, the media player uses mpv for video playback, which is wayland-native and hides the cursor when fullscreen. 

Even though I can't use labwc's HideCursor action for the time being, it's not a big deal because cursor is hidden when I most want it hidden (during video playback).

Incidentally, with labwc 0.8.4 the Alt+Tab window list shows icons (which is cute) but no longer makes it obvious which applications are running in xwayland vs. which ones are wayland-native (which was very useful). Is there a way to restore the old behavior?

EDIT: Please ignore my comment about the window switcher--I just found https://labwc.github.io/labwc-config.5.html#window_switcher

Consolatis:
> Please ignore my comment about the window switcher--I just found https://labwc.github.io/labwc-config.5.html#window_switcher

Too late, was just about to link it :p

Regarding the issue itself: No, I don't think there is a way for that currently. Window rules can also execute actions but that requires some application to be spawned. Emulating mouse or key input in turn resets the idle timer, otherwise the screen would be locked when controlled via VNC for example.

bdantas:
> Regarding the issue itself: No, I don't think there is a way for that currently.
 
No worries. Because of mpv's ability to hide the cursor when fullscreen, this is not a pain point for me. It's just a "would be nice to have" feature that would give me more control over the cursor without breaking my screensaver.

[...]

PR #2900, opened 2025-07-10
SIGUSR1 to trigger HideCursor action

bdantas:
Not having an equivalent to X's `unclutter` (which hides the cursor after a period of inactivity) is a big pain point for some users. Sway has `seat * hide_cursor 5000` but in labwc there is currently no way to achieve similar functionality without a kluge (see https://github.com/labwc/labwc/issues/2880). I had raised the possibility of configurable signal handlers to Consolatis but, on second thought, I think that would be overkill. An elegant way to achieve this functionality in labwc would be something like `swayidle timeout 5 "pkill -FOO labwc" &` where `FOO` is a signal that triggers the labwc's HideCursor action. labwc uses SIGHUP but SIGUSR1 and SIGUSR2 are both available. I went with SIGUSR1. This is tested and working as expected: `swayidle timeout 5 "pkill -USR1 labwc" &`

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -247,6 +247,7 @@ struct server {
 	struct wlr_compositor *compositor;
 
 	struct wl_event_source *sighup_source;
+	struct wl_event_source *sigusr1_source;
 	struct wl_event_source *sigint_source;
 	struct wl_event_source *sigterm_source;
 	struct wl_event_source *sigchld_source;
--- a/src/server.c
+++ b/src/server.c
@@ -103,6 +103,15 @@ handle_sighup(int signal, void *data)
 }
 
 static int
+handle_sigusr1(int signal, void *data)
+{
+	struct server *server = data;
+
+	cursor_set_visible(&server->seat, false);
+	return 0;
+}
+
+static int
 handle_sigterm(int signal, void *data)
 {
 	struct wl_display *display = data;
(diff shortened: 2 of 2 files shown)


bdantas:
There was one thing in my pull request I was unsure about, in `server.c`:

```
static int
handle_sigusr1(int signal, void *data)
{
	struct server *server = data;          // is this line necessary?

	cursor_set_visible(&server->seat, false);
	return 0;
}
```

ahesford:
I strongly oppose this.

Using one of a very limited number of signals for some hard-coded, extremely specific magic behavior is unintuitive. The proposal adds a completely unique interface for interacting with labwc that would only exist for a very narrow purpose. This is the opposite of an elegant solution. It also, as noted, violates the spirit of our "no custom IPC" rule.

bdantas:
Hi ahesford. Your objection makes sense. I'm just a bit desperate because I love labwc but am finding that it's currently impossible to have this simple functionality (which X and sway both offer).
 
My pull request may not be elegant but would make possible something that's currently impossible. Something is better than nothing :) If we can find a way to make this functionality possible _and_ elegant _and_ in keeping with labwc design principles...that would be much better, of course. Any ideas?

bdantas:
Maybe implementing something like this in the config file? 

```
<labwc_config>
<mouse>
  <hideCursorTimeout="5" />
</mouse>
</labwc_config>
```
 
I wouldn't be able to implement this myself, though, as much I would like to.

ahesford:
I would not object to an option to just hide the cursor after some idle delay.

Consolatis:
What are your thoughts on a `hideCursorTimeout` setting @jp7677 ?

jp7677:
> What are your thoughts on a `hideCursorTimeout` setting @jp7677 ?

Personally I wouldn't use such a feature, but I wouldn't mind having it either. 
I think `<core hideCursorTime="timeout in ms" />` would be a better UI since the cursor can be controlled by mouse _and_ tablet (touch hides the cursor on touch-down), and also to be consistent with `popupTime`. I guess all mouse/tablet events should reset the timer? What about gesture and keyboard events?

johanmalm:
Closing this on the basis that we don't want to reserve SIGUSR1 for a marginal use-case.

If anyone actually wants a timeout based hide setting, then let's open a new discussion or PR.

"Other" IPC would be far preferable to SIGUSR1 for this type of thing... but would also be quite a big step from where we currently are with scope/thinking.

PR #2907, opened 2025-07-12
Make wlroots DRM backend optional

Consolatis:
Draft because I didn't try to build it without any libdrm headers installed so this might miss something like the legacy `wlr_drm`.

Can be tested with something like
```sh
git fetch origin pull/xx/head && git checkout FETCH_HEAD
meson setup build_tmp -Dwlroots:backends=libinput --force-fallback-for=wlroots
meson compile -C build_tmp
```


johanmalm:
What is the objective/rationale?

Consolatis:
> What is the objective/rationale?

Sorry, should have mentioned that in the first post. There are cases where drm isn't required or not supported. An example is the termux environment on Android which patches an own Android backend into wlroots. Other use-cases include usage of labwc in nested-only settings like only with the X11 or Wayland backends or completely headless, e.g. as a wayvnc-only host using the headless backend.

Currently labwc just fails to link in this cases as can be seen when you use the reproduction steps in the first post on the master branch.

It also serves as a test-case for further work to also make libinput optional which is required for porting labwc to some of the BSDs (or termux) which do not support it.

Updated the first post with the rationale.

Consolatis:
Replaced by
- #3470

PR #2909, opened 2025-07-13
xwayland: fix swapped width/height in _NET_WM_ICON stride calculation

jlindgren90:
Probably slipped through since most window icons are square anyway.

cc @Consolatis


Consolatis:
whoops.

Thanks!

PR #2910, opened 2025-07-13
img: fix apparent double-free in img_svg_render() failure path

jlindgren90:
img_svg_render() calls g_object_unref() on the RsvgHandle in its error path, but the handle is owned by the shared lab_img_data struct and will be double-freed later by lab_img_destroy().

The double-free was introduced when img_svg_load() was split from img_svg_render(). The g_object_unref() should have been removed from img_svg_render() but was missed.

Fixes: 16dbdc64e58d66011bbf319b92de844dab0ca8d9
("ssd: rework titlebar button rendering")

cc @tokyo4j

### DIFF
--- a/src/img/img-svg.c
+++ b/src/img/img-svg.c
@@ -65,6 +65,5 @@ img_svg_render(RsvgHandle *svg, int w, int h, double scale)
 error:
 	wlr_buffer_drop(&buffer->base);
 	cairo_destroy(cr);
-	g_object_unref(svg);
 	return NULL;
 }


tokyo4j:
Thanks!

Consolatis:
Does this end up crashing labwc when loading a broken SVG? If yes it seems like this (and potentially #2909) would deserve a 0.9.1 bugfix release which means that we should likely not merge other non-fixes until then.

CC @ahesford @johanmalm

tokyo4j:
> Does this end up crashing labwc when loading a broken SVG? If yes it seems like this (and potentially https://github.com/labwc/labwc/pull/2909) would deserve a 0.9.1 bugfix release which means that we should likely not merge other non-fixes until then.

I think it's relatively unlikely that the double-free or UAF happens in reality. At least, syntax errors are caught at loading SVG files with `rsvg_handle_new_from_file()`, before rendering them with `rsvg_handle_render_document()`.

As for #2909, I wrote a C program (with ChatGPT) that sets icon 64x32 with `_NET_WM_ICON` to see if it crashes labwc:

<details><summary>program</summary>

```c
#include <X11/Xlib.h>
#include <X11/Xatom.h>
#include <stdlib.h>
#include <unistd.h>
#include <stdint.h>

#define ICON_W 64
#define ICON_H 32

void set_window_icon(Display *dpy, Window win) {
    Atom net_wm_icon = XInternAtom(dpy, "_NET_WM_ICON", False);
    Atom cardinal = XInternAtom(dpy, "CARDINAL", False);

    // Allocate memory for icon data
    size_t array_size = 2 + ICON_W * ICON_H; // [width, height, pixels...]
    unsigned long *icon_data = malloc(array_size * sizeof(unsigned long));
    if (!icon_data) return;

    icon_data[0] = ICON_W; // width
    icon_data[1] = ICON_H; // height

    // Fill the icon with opaque green (ARGB: 0xFF00FF00)
    for (int i = 0; i < ICON_W * ICON_H; ++i) {
        icon_data[2 + i] = 0xFF00FF00;
    }

    // Set the icon property
    XChangeProperty(dpy, win, net_wm_icon, cardinal, 32,
                    PropModeReplace, (unsigned char *)icon_data, array_size);

    free(icon_data);
}

int main() {
    Display *dpy = XOpenDisplay(NULL);
    if (!dpy) return 1;

    int screen = DefaultScreen(dpy);
    Window root = RootWindow(dpy, screen);

    // Create window
    Window win = XCreateSimpleWindow(dpy, root, 100, 100, 200, 200, 1,
                                     BlackPixel(dpy, screen), WhitePixel(dpy, screen));

    // Set window icon
    set_window_icon(dpy, win);

    // Set window title
    XStoreName(dpy, win, "Green Icon Window");

    // Select input and map window
    XSelectInput(dpy, win, ExposureMask | KeyPressMask);
    XMapWindow(dpy, win);

    // Basic event loop
    XEvent ev;
    while (1) {
        XNextEvent(dpy, &ev);
        if (ev.type == KeyPress)
            break;
    }

    XDestroyWindow(dpy, win);
    XCloseDisplay(dpy);
    return 0;
}
```

</details>

As a result, the icon became entirely transparent, but it didn't crash. So I guess (hope) the wrong stride calculation isn't very severe, either.

Anyway, sorry for my stupid mistakes!

johanmalm:
> Does this end up crashing labwc when loading a broken SVG?

I've tried using a borked svg file, but cannot force a crash, so vote for **not** doing a `0.9.1` at a shorter interval than usual.

ahesford:
I haven't seen an issue in my normal workflow, so I don't have an opinion on a quick bug-fix release either way.

Consolatis:
> I think it's relatively unlikely that the double-free or UAF happens in reality. At least, syntax errors are caught at loading SVG files with `rsvg_handle_new_from_file()`, before rendering them with `rsvg_handle_render_document()`.

> As a result, the icon became entirely transparent, but it didn't crash. So I guess (hope) the wrong stride calculation isn't very severe, either.

Thanks for looking into this. As it seems pretty hard to actually produce a crash here I tend to agree that we likely don't need a short-term bug-fix release unless something else pops up in the next days.

> Anyway, sorry for my stupid mistakes!

No worries, it also slipped through review. One of the reasons labwc requires reviews (other than more eyes == more chances to catch issues like this) is also that there is never a single person responsible if a bug ends up in labwc. Bugs happened before and will happen in the future, the important thing is that they are dealt with.

jlindgren90:
I agree with not doing a quick release - since this issue was just noticed in code inspection and hasn't been observed "in the wild."

PR #2929, opened 2025-07-19
menu: add struct menu_parse_context to reduce static vars

jlindgren90:
The lifetime of the "current_" variables (current_menu, current_item, current_item_action) is very difficult to understand from reading the code. It appears that e.g. current_menu could still point to a previous menu when starting to parse a new one, with unpredictable results.
    
Let's use a context struct when parsing, and consistently initialize it when beginning to build a new menu.
    
Lightly tested with:
    
- default menus (no menu.xml)
- example static menu from labwc.github.io/getting-started.html
- an added "client-list-combined-menu" sub-menu
- pipe menu generated by `labwc-menu-generator -p`

Includes one commit from #2884 to avoid conflicts.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -17,7 +17,6 @@
 #include "common/font.h"
 #include "common/lab-scene-rect.h"
 #include "common/list.h"
-#include "common/macros.h"
 #include "common/mem.h"
 #include "common/nodename.h"
 #include "common/scaled-font-buffer.h"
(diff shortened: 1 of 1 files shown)


src/menu/menu.c
```diff
@@ -1333,19 +1339,16 @@ static void
 create_pipe_menu(struct menu_pipe_context *ctx)
 {
 	struct server *server = ctx->pipemenu->server;
-	struct menu *old_current_menu = current_menu;
-	current_menu = ctx->pipemenu;
-	if (!parse_buf(server, &ctx->buf)) {
-		goto restore_menus;
+	struct menu_parse_context parse_ctx = {
+		.server = server, .menu = ctx->pipemenu};
```
tokyo4j:
```suggestion
	struct menu_parse_context parse_ctx = {
		.server = server,
		.menu = ctx->pipemenu,
	};
```

I think each field should be followed by a line break, to be consistent with other parts of our codebase.

tokyo4j:
LGTM other than minor styling suggestion above. Tested and didn't notice any breakages.

I think we can refactor the whole parser logic like #2667 in the future.

src/menu/menu.c
jlindgren90: Sure, I agree. I blame clang-format for this - I got lazy :)

PR #2930, opened 2025-07-19
Drop cosmic-workspace and use wlroots implementation of ext-workspace-v1

tokyo4j:
For testing https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5115

I'm not sure when to merge this PR. IMO, we can merge this as soon as a new wlroots version with the MR above is released.


Consolatis:
I am a bit hesitant to remove cosmic-workspaces. It doesn't really hurt to have it around and some users might still rely on it.
Totally fine with replacing our ext-workspace implementation with the wlroots one though.

tokyo4j:
> I am a bit hesitant to remove cosmic-workspaces. It doesn't really hurt to have it around and some users might still rely on it.
> Totally fine with replacing our ext-workspace implementation with the wlroots one though.

I'm fine with keeping cosmic-workspace too, but then I don't see much benefit of replacing only ext-workspace implementation, as those two implementations are almost identical. I'd rather keep this PR on hold for another 1 or 2 years and then finally merge it when we can assume no one is relying on cosmic-workspace.

sfzaw:
wait till everyone adopts ext-workspace-v1  protocol first. waybar hasnt implemented it yet. until everyone adopts it can be marked as legacy feature.

tokyo4j:
> wait till everyone adopts ext-workspace-v1 protocol first. waybar hasnt implemented it yet. until everyone adopts it can be marked as legacy feature.

I wonder there are any practical applications that still depend on cosmic-workspace protocol. AFAIR, waybar didn't merge cosmic-workspace protocol support and instead supported ext-workspace-v1 protocol recently.

I guess the timing for us to drop cosmic-protocol is when we can assume everyone uses newer versions of [sfwbar](https://github.com/LBCrion/sfwbar/), which firstly supported cosmic-workspace and then ext-workspace-v1 later.

PR #2933, opened 2025-07-21
cursor: prevent hi-res mice triggering scroll actions too often

### LINKED ISSUE #1068: "Duplicate" scroll events on some mice
I have this in my `rc.xml` to change workspaces using the scroll wheel:
```
        <context name="Root">
                <action name="GoToDesktop" to="left" wrap="no" />
            </mousebind>
                <action name="GoToDesktop" to="right" wrap="no" />
            </mousebind>
        </context>
```
Works fine with my old Logitech M500.

With my Logitech M705, though, labwc jumps multiple workspace at once during one "click" of the wheel. (By "click" I mean the tactile feedback while turning the wheel -- I'm not pressing down the wheel, like I would to issue a middle mouse click).

I had a look at the output of `wev`, but I can't make heads or tails of it.

M500 (ok) -- one click of the wheel results in a "frame" line and the three "axis" lines:
```
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6172381; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6172861; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6173341; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6173949; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6174517; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6175077; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6175669; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6176365; axis: 0 (vertical), value: -15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6176397; axis: 0 (vertical), value: -15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6177061; axis: 0 (vertical), value: -15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6177621; axis: 0 (vertical), value: -15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6178189; axis: 0 (vertical), value: -15.000000
```
M705 (not ok):
```
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6218674; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6219670; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: 1
[13:      wl_pointer] axis: time: 6220724; axis: 0 (vertical), value: 15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6222086; axis: 0 (vertical), value: -15.000000
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] frame
[13:      wl_pointer] axis_source: 0 (wheel)
[13:      wl_pointer] axis_stop: axis: 0 (vertical), discrete: -1
[13:      wl_pointer] axis: time: 6223280; axis: 0 (vertical), value: -15.000000
```
During a physical click of the M705's scroll wheel, I see the "axis" lines. I have no idea what those "frame" lines are supposed to mean. They occur in between the clicks. If I turn the scroll wheel *very slowly* (without triggering a click), I can see the "frame" lines appear one by one.

Scrolling in applications like Firefox or l3afpad behaves the same for both mice, though.

For comparison: On plain X11, I only see button events like "button 4". It's the same for both mice.

Any ideas what this could be? (Is this even a labwc bug or should I take this up with the wlroots guys or libinput? Since applications work fine on Wayland, I suspect it has something to do with labwc ...)

tokyo4j:
Fixes #1068

Hi-res mice produces mulitple scroll events with `delta_discrete` != 0 during a single "click". This patch makes them trigger `Scroll` actions only when the accumulated `delta_discrete` exceeds 120 (= 1 click).

See https://lists.freedesktop.org/archives/wayland-devel/2019-April/040377.html for how hi-res scroll events are reported.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -111,9 +111,10 @@ struct seat {
 	bool cursor_visible;
 	struct wlr_cursor *cursor;
 	struct wlr_xcursor_manager *xcursor_manager;
-	struct {
-		double x, y;
-	} smooth_scroll_offset;
+	struct accumulated_scroll {
+		double delta;
+		double delta_discrete;
+	} accumulated_scrolls[2]; /* indexed by wl_pointer_axis */
 	bool cursor_scroll_wheel_emulation;
 
 	/*
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1245,38 +1245,44 @@ struct scroll_info {
 };
 
 static struct scroll_info
-compare_delta(double delta, double delta_discrete, double *accum)
+compare_delta(double delta, double delta_discrete, struct accumulated_scroll *accum)
 {
-	/*
-	 * Smooth scroll deltas are in surface space, so treating each unit as a
-	 * scroll event would result in too-fast scrolling.
-	 *
-	 * This fudge factor (inherited from various historic projects, incl. Weston)
-	 * produces events at a more reasonable rate.
-	 *
-	 * For historic context, see:
-	 * https://lists.freedesktop.org/archives/wayland-devel/2019-April/040377.html
-	 */
-	const double SCROLL_THRESHOLD = 10.0;
 	struct scroll_info info = {0};
 
-	if (delta_discrete < 0 || delta < 0) {
-		info.direction = -1;
-	} else if (delta_discrete > 0 || delta > 0) {
-		info.direction = 1;
-	}
-
-	if (delta == 0.0) {
-		/* Delta 0 marks the end of a scroll */
-		*accum = 0.0;
+	if (delta_discrete) {
+		/* mice */
+		info.direction = delta_discrete > 0 ? 1 : -1;
+		accum->delta_discrete += delta_discrete;
+		/*
+		 * Non-hi-res mice produce delta_discrete of ±120 for every
+		 * "click", so it always triggers actions. But for hi-res mice
+		 * that produce smaller delta_discrete, we accumulate it and
+		 * run actions after it exceeds 120(= 1 click).
+		 */
+		if (fabs(accum->delta_discrete) >= 120.0) {
+			accum->delta_discrete = fmod(accum->delta_discrete, 120.0);
+			info.run_action = true;
+		}
 	} else {
-		/* Accumulate smooth scrolling until we hit threshold */
-		*accum += delta;
-	}
-
-	if (delta_discrete != 0 || fabs(*accum) > SCROLL_THRESHOLD) {
-		*accum = fmod(*accum, SCROLL_THRESHOLD);
-		info.run_action = true;
+		/* 2-finger scrolling on touchpads */
+		if (delta == 0) {
+			/* delta=0 marks the end of a scroll */
+			accum->delta = 0;
+			return info;
+		}
+		info.direction = delta > 0 ? 1 : -1;
+		accum->delta += delta;
+		/*
+		 * The threshold of 10 is inherited from various historic
+		 * projects including weston.
+		 *
+		 * For historic context, see:
+		 * https://lists.freedesktop.org/archives/wayland-devel/2019-April/040377.html
+		 */
+		if (fabs(accum->delta) >= 10.0) {
+			accum->delta = fmod(accum->delta, 10.0);
+			info.run_action = true;
+		}
 	}
 
 	return info;
@@ -1290,21 +1296,16 @@ process_cursor_axis(struct server *server, enum wl_pointer_axis orientation,
 	uint32_t modifiers = keyboard_get_all_modifiers(&server->seat);
 
 	enum direction direction = LAB_DIRECTION_INVALID;
-	struct scroll_info info = {0};
+	struct scroll_info info = compare_delta(delta, delta_discrete,
+		&server->seat.accumulated_scrolls[orientation]);
 
 	if (orientation == WL_POINTER_AXIS_HORIZONTAL_SCROLL) {
-		info = compare_delta(delta, delta_discrete,
-			&server->seat.smooth_scroll_offset.x);
-
 		if (info.direction < 0) {
 			direction = LAB_DIRECTION_LEFT;
 		} else if (info.direction > 0) {
 			direction = LAB_DIRECTION_RIGHT;
 		}
 	} else if (orientation == WL_POINTER_AXIS_VERTICAL_SCROLL) {
-		info = compare_delta(delta, delta_discrete,
-			&server->seat.smooth_scroll_offset.y);
-
 		if (info.direction < 0) {
 			direction = LAB_DIRECTION_UP;
 		} else if (info.direction > 0) {
@@ -1324,8 +1325,8 @@ process_cursor_axis(struct server *server, enum wl_pointer_axis orientation,
 					&& mousebind->mouse_event == MOUSE_ACTION_SCROLL) {
 				handled = true;
 				/*
-				 * Action may not be executed if the accumulated scroll
-				 * delta on touchpads doesn't exceed the threshold
+				 * Action may not be executed if the accumulated scroll delta
+				 * on touchpads or hi-res mice doesn't exceed the threshold
 				 */
 				if (info.run_action) {
 					actions_run(ctx.view, server, &mousebind->actions, &ctx);


Consolatis:
LGTM but lets wait for confirmation that it fixes #1068.

Consolatis:
One slight difference in this PR is that we don't reset the touchpad accumulator anymore, e.g. it keeps the remains of the previous scroll events. If that makes a difference in practice I don't know.

tokyo4j:
> One slight difference in this PR is that we don't reset the touchpad accumulator anymore, e.g. it keeps the remains of the previous scroll events. If that makes a difference in practice I don't know.

That's a good point. I updated the patch to restore the reset.

vain:
Hi,

yes, thank you, this is much better. One “click” on a hi-res wheel now results in one workspace switch. I also tested this with my Synaptics touchpad, appears to work fine.

Almost. 😅

I’m not exactly sure what triggers this, but sometimes no events register anymore when turning the wheel one click up, down, up, down, …

Doing this appears to be able to reproduce it quite often (not 100%, though):

-   Start labwc.
-   Do 4 clicks downwards on a low-res wheel. This switches the workspaces 1->2->3->4->5.
-   Then do 4 clicks upwards on the low-res wheel. This switches back 5->4->3->2->1.
-   Wait for the workspace switcher / OSD to disappear (not 100% sure if this matters).
-   Do some clicks on a hi-res wheel. It’s immediately noticeable that the first click is already missing.
-   No matter which mouse I’m using now, wheel down -> wheel up -> wheel down -> wheel up -> … does nothing anymore (until I restart labwc).

(Since this affects both mice at the same time, I don’t think it’s a physical defect of one of the wheels.)

– edit: Forgot to mention it, I’m on 31608471.

tokyo4j:
> I’m not exactly sure what triggers this, but sometimes no events register anymore when turning the wheel one click up, down, up, down, …

Actually, that's what I expected. When I replay the events from https://github.com/labwc/labwc/issues/1068#issuecomment-3049433203, the first batch of scrolls don't switch the workspace. This is because libinput "accumulates" the scroll events from evdev layer. For example, the scroll events from evdev layer `15,15,15,15,15,30,15` are reported to labwc as `60,15,30`, which don't reach the threshold `120` to trigger actions.

Libinput author says this behavior for working around some "flaky scroll wheels" and I guess there's no way to fix this at the moment:

<details><summary>IRC log</summary>

```
[12:39:33] <tokyo4j> whot: why does libinput "accumulate" hi-res mouse scrolls? M705 mouse reports 15,15,15,15,15,30,15 (120 in total) scroll events at evdev layer, but libinput reports them as 60,15,30 (105 in total). So we have difficulty in reliably executing a mouse binding for exact one scroll click.
[13:22:38] <whot> tokyo4j: user complaints about flaky scroll wheels, so that heuristic was a way to work around that. This was fixed  in https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/1261 and https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/1258, merged a few h ago
[13:24:06] <whot> tokyo4j: current code still has some accumulation but it's now two clicks, i.e. in your case it's 30 (not 60) to start and for mice with low res multipliers there is no accumulation (because it's unlikely to have inadvertant scrolls anyway)
[13:33:20] <whot> tokyo4j: still trying to figure out if *any* devices are actually affected by flaky scroll wheels or just the MX Master 3 series (or at least some of those...) but once the wheel is hi-res enough it's hard to be sure what is flakyness and what is just finger movement
[15:25:03] <tokyo4j> whot: OK, thanks for your explanation. I installed libinput-git from AUR and now the reported scroll events are 15,15,15,15,15,30 (105 in total). I personally expect the scroll events from evdev (including the last 15-degree scroll) to be just passed through to the compositor immediately, at least with M705 mouse, but I'm not following the context of debouncing issue and don't know what's the best solution.
[17:12:43] <whot> tokyo4j: it's one of the many cases where there is no single correct answer because it depends on the mouse, the user and the interactions themselves. but it's basically the same as the de-jittering of the touchpad movement that we already have, the key is finding out what the magic line is where the least number of users is unhappy
```

</details>

@Consolatis do you have any thoughts on this? IMO, we can merge this PR to make the situation better, even though it causes some missed scroll actions.

@vain ~Just to make sure that we're on the same page, could you post the output of `libinput record` again when some scroll clicks are missed as you described? Please note that you should not touch the mouse scroll since you launch labwc and until you start recording, so that we can make sure the internal accumulated scroll values in labwc/libinput are initially untouched and reproduce the issue.~ I rent MX Master 3 from my colleague and confirmed the issue. The problem is, as expected, libinput doesn't always report 120 `delta_discrete` for a single click.

Consolatis:
> @Consolatis do you have any thoughts on this? IMO, we can merge this PR to make the situation better, even though it causes some missed scroll actions.

If libinput eats events that is not good because it means we can not rely on a wheel "click" to be 120 anymore. Only thing I can think of for the switching-direction issue is to add a self-refreshing timer of a couple hundred ms which resets the accumulator. That is pretty hacky though.

I think in general this is a libinput issue. It should not eat any scroll events or at least provide an option to disable that behavior or only apply it to specific models. Or it should ensure that whatever it considers a wheel "click" to then also end up with `120` in total at the compositor.

I agree that this PR is better than the current state in labwc so its definitely an improvement and we can fine-tune it later if necessary.

tokyo4j:
OK, let's go for this for now.

> If libinput eats events that is not good because it means we can not rely on a wheel "click" to be 120 anymore. Only thing I can think of for the switching-direction issue is to add a self-refreshing timer of a couple hundred ms which resets the accumulator. That is pretty hacky though.
> 
> I think in general this is a libinput issue. It should not eat any scroll events or at least provide an option to disable that behavior or only apply it to specific models. Or it should ensure that whatever it considers a wheel "click" to then also end up with 120 in total at the compositor.

I agree. I'll need read though the libinput issues on gitlab about scrolling behaviors.

I also confirmed the same issue in KDE Plasma: when I bind Meta+scrolls to switching workspaces, some scroll clicks from hi-res mice are missed. So this is a common problem among wayland compositors.

I even wonder this is solely a libinput problem. When I dump scroll events from MX Master 3 with `evtest`, sometimes the scroll delta accumulated for a single click is smaller than 120. So I'm afraid there's no way to fully fix this other than patching kernel drivers.

vain:
Thank you for fixing this!

(I’ll try to steer clear of such hi-res wheels in the future. I have zero use for this feature.)

PR #2942, opened 2025-07-24
cursor: also toggle mousebinds with `ToggleKeybinds`

### LINKED ISSUE #1536: ToggleKeybinds and mousebinds with keyboard modifiers (Alt+Click, Alt+Drag)
When input inhibited by application, it is generally expected what all input will be passed to application.
But now, mousebinds like ALT+click or ALT+drag-move ignores this state, that prevents pass mouse click completely to client. I'm having issue with `wlfreerdp`, but this reproducible with any application.

Technically this is not related to "ToggleKeybinds" by itself, just this mouse actions which involve ALT (probably and other modifiers) prevent to event been dispatched. E.g. if i'm remove mouse actions like:

```
    <context name="Frame">
      <mousebind button="A-Left" action="Press">
      </mousebind>
      <mousebind button="A-Left" action="Drag">
      </mousebind>
      <mousebind button="A-Right" action="Press">
      </mousebind>
      <mousebind button="A-Right" action="Drag">
      </mousebind>
    </context>
```

then this starts to work as expected.

However, since application takes all keyboard input, it looks like have sense prevent match mousebinds which includes keyboard modifiers.

tokyo4j:
Closes #1536.

Mousebinds can still be applied when the cursor is over window decoration.

The first commit removes `seat->nr_inhibited_keybind_views` as I don't think it's needed.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -227,10 +227,10 @@ Actions are used in menus and keyboard/mouse bindings.
 	window.
 
 *<action name="ToggleKeybinds" />*
-	Stop handling keybinds other than ToggleKeybinds itself.
-	This can be used to allow A-Tab and similar keybinds to be delivered
-	to Virtual Machines, VNC clients or nested compositors.
-	A second call will restore all original keybinds.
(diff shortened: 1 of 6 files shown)


Consolatis:
LGTM, not tested.

Issue #2944, opened 2025-07-26
Labwc crashes on panel destruction with maximized & unmapped windows


tokyo4j:
### Problem description

Same root cause as #2937.

When a panel is destroyed immediately (<100ms) after a maximized window is unmapped, labwc crashes at assertion: `labwc: types/xdg_shell/wlr_xdg_surface.c:168: wlr_xdg_surface_schedule_configure: Assertion 'surface->initialized' failed`.

Stacktrace:
```
labwc: types/xdg_shell/wlr_xdg_surface.c:168: wlr_xdg_surface_schedule_configure: Assertion `surface->initialized' failed.

Thread 1 "labwc" hit Breakpoint 1, 0x00007ffff6e3555c in abort () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff6e3555c in abort () at /usr/lib/libc.so.6
#1  0x00007ffff6e354e3 in ??? () at /usr/lib/libc.so.6
#2  0x0000555555638813 in wlr_xdg_surface_schedule_configure (surface=0x55555696ce00) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_surface.c:168
#3  0x000055555563b6ee in wlr_xdg_toplevel_set_size (toplevel=0x5555568395b0, width=1280, height=720)
    at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_toplevel.c:579
#4  0x00005555555a9ff6 in xdg_toplevel_view_configure (view=0x5555562b4aa0, geo=...) at ../src/xdg.c:499
#5  0x00005555555a1fce in view_move_resize (view=0x5555562b4aa0, geo=...) at ../src/view.c:636
#6  0x00005555555a3dd2 in view_apply_maximized_geometry (view=0x5555562b4aa0) at ../src/view.c:1334
#7  0x00005555555a3ece in view_apply_special_geometry (view=0x5555562b4aa0) at ../src/view.c:1350
#8  0x00005555555a4fe0 in view_adjust_for_layout_change (view=0x5555562b4aa0) at ../src/view.c:1885
#9  0x0000555555587b6f in desktop_arrange_all_views (server=0x7fffffffd350) at ../src/desktop.c:37
#10 0x0000555555592701 in output_update_usable_area (output=0x55555650aeb0) at ../src/output.c:1049
#11 0x000055555558bcca in handle_unmap (listener=0x55555678dcf8, data=0x0) at ../src/layers.c:373
[...]

Issue #2937, opened 2025-07-23
Assert failure when resized window is unmapped


tokyo4j:
### Problem description

When the window is unmapped while being resized with `Resize` action, labwc crashes with the assertion failure: `labwc: types/xdg_shell/wlr_xdg_surface.c:168: wlr_xdg_surface_schedule_configure: Assertion 'surface->initialized' failed`

Backtrace:
```
#0  0x00007ffff603555c in abort () at /usr/lib/libc.so.6
#1  0x00007ffff60354e3 in ??? () at /usr/lib/libc.so.6
#2  0x000055555611e70e in wlr_xdg_surface_schedule_configure (surface=0x7d2ff22710c0) at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_surface.c:168
#3  0x000055555612e739 in wlr_xdg_toplevel_set_size (toplevel=0x7d3ff2206640, width=608, height=607)
    at ../subprojects/wlroots/types/xdg_shell/wlr_xdg_toplevel.c:579
#4  0x0000555555e68af6 in xdg_toplevel_view_configure (view=0x7d7ff226ac80, geo=...) at ../src/xdg.c:499
#5  0x0000555555e419c2 in view_move_resize (view=0x7d7ff226ac80, geo=...) at ../src/view.c:636
#6  0x0000555555ee2918 in process_cursor_resize (server=0x7bfff14009b0, time=4291529) at ../src/input/cursor.c:361
#7  0x0000555555ee576f in cursor_process_motion (server=0x7bfff14009b0, time=4291529, sx=0x7bfff0fb3c20, sy=0x7bfff0fb3c40) at ../src/input/cursor.c:593
#8  0x0000555555ee8f6b in preprocess_cursor_motion (seat=0x7bfff1400b50, pointer=0x7d3ff21ec040, time_msec=4291529, dx=-2.6875, dy=-1.34375)
    at ../src/input/cursor.c:838
#9  0x0000555555ee9c34 in handle_motion (listener=0x7bfff1400d80, data=0x7bfff0e35130) at ../src/input/cursor.c:903
#10 0x00007ffff7f7551e in wl_signal_emit_mutable () at /usr/lib/libwayland-server.so.0
#11 0x000055555615cb0f in handle_pointer_motion (listener=0x7d5ff223e8b8, data=0x7bfff0e35130) at ../subprojects/wlroots/types/wlr_cursor.c:731
[...]

PR #2948, opened 2025-07-27
src/xdg.c: prevent interacting with un-initialized xdg toplevels after unmap

Consolatis:
Fixes:
- #2937
- #2944

Originally-Reported-By: tranzystorekk via IRC

@tokyo4j could you give this a test?


src/xdg.c
```diff
@@ -513,12 +527,20 @@ xdg_toplevel_view_configure(struct view *view, struct wlr_box geo)
 static void
 xdg_toplevel_view_close(struct view *view)
 {
+	if (!xdg_toplevel_from_view(view)->base->initialized) {
```
vyivel: `xdg_toplevel.close` is not a part of a configure sequence so this check isn't strictly required.

Consolatis: Thanks, removed.

tranzystorekk:
Tested, the issue I reported no longer happens with this patch :)

tokyo4j:
Thanks, I confirmed this PR fixes both of the two issues. LGTM.

PR #2949, opened 2025-07-28
Fix some warnings from clang-tidy

tokyo4j:
Notably this fixes a possible null pointer dereference in `warp_cursor()` when `output_nearest_to_cursor()` returns null.


Consolatis:
Thanks

PR #2950, opened 2025-07-28
Fix header ordering dependencies

jlindgren90:
A couple of the headers under `include` do not compile unless other headers are included first. The first commit fixes those.

The second commit implements the common practice of making each C file include its matching header first -- precisely to ensure that header compiles cleanly without implicit dependencies on other headers.

I used a crude bash script to automate the 2nd commit -- ugly enough that I'm not inclined to share it ;)

### DIFF
--- a/include/common/graphic-helpers.h
+++ b/include/common/graphic-helpers.h
@@ -2,6 +2,8 @@
 #ifndef LABWC_GRAPHIC_HELPERS_H
 #define LABWC_GRAPHIC_HELPERS_H
 
+#include <stdbool.h>
+#include <stdint.h>
 #include <cairo.h>
 
 struct wlr_fbox;
--- a/include/config/touch.h
+++ b/include/config/touch.h
@@ -2,7 +2,7 @@
 #ifndef LABWC_TOUCH_CONFIG_H
 #define LABWC_TOUCH_CONFIG_H
 
-#include <stdint.h>
+#include <stdbool.h>
 #include <wayland-util.h>
 
 struct touch_config_entry {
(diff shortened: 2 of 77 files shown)


Consolatis:
No objections from my side. @ahesford @johanmalm @tokyo4j thoughts?

If we go this route, we should likely mention including the primary header first in the code style guidelines.

tokyo4j:
I prefer keeping the alphabetical order for `#include` directives, as that's what wlroots does and I think it's prettier.

But I can accept it if other devs agree with the proposal.

ahesford:
I don't have a strong preference, but I usually prefer to put system headers before project-local headers.

johanmalm:
I like this.

Suggest updating CONTRIBUTING.md accordingly including the reason ("to ensure that header compiles cleanly without implicit dependencies on other headers"). Other than that, LGTM.

https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#devault-deviations

Consolatis:
So basically the idea is:
- "primary" header first to ensure no implicit dependencies on other headers
- system headers, ordered alphabetically
- internal headers, ordered alphabetically

johanmalm:
> So basically the idea is:
> 
> * "primary" header first to ensure no implicit dependencies on other headers
> * system headers, ordered alphabetically
> * internal headers, ordered alphabetically

Yes, that's how I understand it.

jlindgren90:
Added a section to CONTRIBUTING.md (and a separate commit to fix some other minor issues I noticed while in there).

johanmalm:
Thanks :smile:

PR #2953, opened 2025-07-28
NEWS.md: update with notes on wlroots changes

johanmalm:


johanmalm:
RFC.

Is it useful? Is it pitched right? Is it accurate.

If this is deemed right, I suggest that we combine it with the words under "There is a regression warning worth noting for the switch to wlroots 0.19:" in the `0.9.0` release notes section.

Consolatis:
Sounds about right. The "fake output" could maybe be reworded, its more like wlroots no longer trying to match the existing outputs with the new connector state. I think the word ` "fake"` could just be removed from the sentence.

jlindgren90:
LGTM

johanmalm:
> Sounds about right. The "fake output" could maybe be reworded, its more like wlroots no longer trying to match the existing outputs with the new connector state. I think the word ` "fake"` could just be removed from the sentence.

Thanks. Done.

PR #2957, opened 2025-07-30
include: add output.h and pare down labwc.h

jlindgren90:
`labwc.h` currently has a lot of stuff in it and feels a bit disorganized.
I propose to move a few things to (IMO) more logical places:

- output-related stuff goes to  `output.h` (new file)
- `struct input` goes to `input.h`
- `struct keyboard` goes to `keyboard.h`
- `enum lab_cycle_dir` goes to `osd.h`
- `struct constraint` becomes private to `cursor.c`
- and a long list of `#includes` go to the files that actually need them

Afterward, what remains in `labwc.h` is mainly `struct seat`, `struct server`, and related stuff.
(It could potentially even be renamed to `server.h`, but I haven't done that here.)


include/input/keyboard.h
```diff
@@ -4,10 +4,25 @@
 
 #include <stdbool.h>
 #include <xkbcommon/xkbcommon.h>
+#include "input.h"
```
Consolatis: `input/input.h`

Consolatis:
Didn't check all changes in detail but overall this looks great!

`seat.h` might be another target.

include/input/keyboard.h
jlindgren90: Fixed

johanmalm:
Thanks. LGTM.
Something I've been of a mind to do for a long time actually :smile:

PR #2960, opened 2025-08-01
action: refactor in preparation for labnag

tokyo4j:
As I proposed in https://github.com/labwc/labwc/pull/2699#issuecomment-3082599289.

I think this PR also makes the logic easier to follow.

I advise you to use "hide whitespaces" feature when reviewing this PR with GitHub's diff viewer.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -776,29 +776,24 @@ view_for_action(struct view *activator, struct server *server,
 }
 
 static bool
-run_if_action(struct view *view, struct server *server, struct action *action)
+match_queries(struct view *view, struct action *action)
 {
-	struct view_query *query;
-	struct wl_list *queries, *actions;
-	const char *branch = "then";
+	struct wl_list *queries = action_get_querylist(action, "query");
+	if (!queries) {
+		return true;
+	}
+	if (!view) {
+		return false;
+	}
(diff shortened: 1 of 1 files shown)


tokyo4j:
I'm not sure how `If` action should behave without focused window.

Previously it was just ignored, but the 2nd commit makes it execute `<else>` branch to make things easier when we support `If` action with `<prompt>`, which should be executed whether or not there's a focused window. In #2699 `<then>` branch is executed, but I chose `<else>` instead as it's more natural in my PoV. 

Does anyone rely on `If` action not being executed without focused window? If no, let's go with the approach in this PR.

johanmalm:
> I'm not sure how `If` action should behave without focused window.
> 
> Previously it was just ignored, but the 2nd commit makes it execute `<else>` branch to make things easier when we support `If` action with `<prompt>`, which should be executed whether or not there's a focused window. In #2699 `<then>` branch is executed, but I chose `<else>` instead as it's more natural in my PoV.

I think that what you suggest makes more sense (the `<else>` branch). Suggest we go with that.

johanmalm:
LGTM. 

You've got a good eye for making things easier to read :smile:

PR #2961, opened 2025-08-01
Remaximize after fullscreen move

IMbackK:
I quite like the behavior of openbox that it allows moving of full screen windows by keeping them maximized and snapping them from one output to the next. This is very useful when you have a lot of outputs and most window movements consist of shoving maximized windows around. The current behavior requires you to manually remaximize the windows after every move, which is tedious.

This pr implements similar behavior by remembering if a window started a move maximized, and re maximizing it when the move is finished if it was. Its a draft, as at the very least we probably want this behavior to be optional

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -222,6 +222,7 @@ struct view {
 	enum ssd_preference ssd_preference;
 	bool shaded;
 	bool minimized;
+	bool remaximize;
 	enum view_axis maximized;
 	bool fullscreen;
 	bool tearing_hint;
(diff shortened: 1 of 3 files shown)


Consolatis:
Thanks for the PR.

> The current behavior requires you to manually remaximize the windows after every move, which is tedious.

Couldn't one also move the window to the top of the new output and thus it would get maximized automatically?

On the action side there is also `MoveToOutput`.

IMbackK:
Sure, but that's less worry free than just flinging the mouse anywhere on the other output

tokyo4j:
I personally don't think we should implement this feature, given that we already have `(Toggle)Maximize` action, `MoveToOutput` action and `<snapping><topMaximize>`.

I have a keybind to run `ToggleMaximize` action with `W-Tab`. So when I move a maximized window to another output, I just drag the window with `W-Left` mousebind and then press `Tab` to maximize it again. I'm not bothered by the extra step of pressing `Tab` key.

PR #2965, opened 2025-08-02
overlay: take into account <core><gap> for edge/region overlay

tokyo4j:
Preparation for #2885.

The motivation is to deduplicate `get_edge_snap_box()` in `interactive.c` and `view_get_edge_snap_box()` in `view.c`, but I think it's a bit more intuitive for the overlay to indicate the actual window size after snapping.

|Before|After (the overlay is shrunk by `<core><gap>`)|
|-|-|
|20250802_17h59m16s_grim|20250802_18h01m21s_grim|


johanmalm:
LGTM

PR #2966, opened 2025-08-02
cursor: fix Scroll mousebinds not inhibited with ToggleKeybinds

tokyo4j:
Fixup for #2942. I forgot about `Scroll` mousebindings.


johanmalm:
Thanks.

PR #2967, opened 2025-08-02
common/xml: let LAB_XML_FOR_EACH() skip first child text nodes

tokyo4j:
Before this patch, first text nodes like the spaces between `<a>` and `<b>` below were also travered by LAB_XML_FOR_EACH(), with their keys "text":

```xml
  <a>  <b>foo</b></a>
```

### DIFF
--- a/include/common/xml.h
+++ b/include/common/xml.h
@@ -35,16 +35,13 @@ bool lab_xml_get_string(xmlNode *node, const char *key, char *s, size_t len);
 bool lab_xml_get_int(xmlNode *node, const char *key, int *i);
 bool lab_xml_get_bool(xmlNode *node, const char *key, bool *b);
 
+/* also skips other unusual nodes like comments */
 static inline xmlNode *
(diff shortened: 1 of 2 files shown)


include/common/xml.h
```diff
@@ -58,11 +55,13 @@ lab_xml_get_key_and_content(xmlNode *node, char **name, char **content)
 }
 
 #define LAB_XML_FOR_EACH(parent, child, key, content) \
-	for ((child) = (parent)->children, \
+	for ((child) = lab_xml_skip_text((parent)->children), \
 		lab_xml_get_key_and_content((child), &(key), &(content)); \
+		\
 		(child); \
+		\
```
tokyo4j, line 62: Added empty backslashes for readability.

tokyo4j:
I also pushed another commit to make `CDATA` work. I intended to make CDATA work in #2667, but I forgot about it and it turned out that CDATA was mostly just ignored.

johanmalm:
LGTM

PR #2970, opened 2025-08-03
menu: fix segfault with toplevel <separator>

### LINKED ISSUE #2969: menu: SEGFAULT with toplevel <separator>
### Problem description

Labwc crashes with SEGFAULT when `menu.xml` contains a `<separator>` as a direct child of `<openbox_menu>` rather than `<menu>`.

I don't think this is worth releasing another fix-only 0.9.2 release, as this issue has been present since 0.8.0, at least.

I'm working on refactoring `menu.c` in the same way as #2667, which also fixes this issue as a side effect. But of course we can fix this issue if my refactoring PR doesn't come in time for the next release.

### Steps to reproduce

1. Have `~/.config/labwc/menu.xml` like below:

```xml
<openbox_menu>
  <separator />
</openbox_menu>
```

2. Launch labwc
3. Labwc crashes with SEGFAULT

### labwc build source

Release

### labwc version

0.9.0-56-gb9c84f9c-dirty (+xwayland +nls +rsvg +libsfdo)

### labwc environment

Nested on Wayland

### Distribution

_No response_

tokyo4j:
Fixes #2969

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -269,6 +269,8 @@ item_create_scene(struct menuitem *menuitem, int *item_y)
 static struct menuitem *
 separator_create(struct menu *menu, const char *label)
 {
+	assert(menu);
+
 	struct menuitem *menuitem = znew(*menuitem);
 	menuitem->parent = menu;
 	menuitem->selectable = false;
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks. Good catch.

LGTM

PR #2971, opened 2025-08-04
menu: refactor parser

tokyo4j:
...with the same approach as #2667.

- `If` actions now works for menus
- `name` argument no longer have to be the first argument of `<action>`
- `label` argument no longer have to be the first argument of `<item>`

Plans for further refactoring:
- Refactor `append_parsed_actions()`, `parse_action()`, `parse_action_args()` and `action_arg_from_xml_node()` to treat `<query>`/`<then>`/`<else>`/`<none>` just like other action arguments. And we can move that logic to another file like `src/action/parse.c`.
- Move menu parser logic into `src/menu/parse.c`


johanmalm:
Thanks. I think this looks fine, but feel i ought to do some testing with various menu files I've got kicking around.

johanmalm:
Well, I have tested with externally defined menus as well as internal ones. 

Codes looks really nice.

LGTM

PR #2972, opened 2025-08-05
output: make autoEnableOutputs=no apply only to drm outputs

jlindgren90:
It is not really useful for other output backends and just results in no outputs being enabled at all. (This is mainly an annoyance for developers normally running with drm but occasionally nested.)


Consolatis:
:)

PR #2973, opened 2025-08-05
view: assert internal signals are disconnected before destroy

jlindgren90:
If they are not empty, then we are headed for use-after-free shortly. An assert() failure is easier to debug than UAF, so let's fail early.

Inspired by:
https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/8f56f7ca43257cc05c7c4eb57a0f541e05cf9a79


Consolatis:
Good idea. Thanks!

PR #2974, opened 2025-08-05
foreign-toplevel: simplify and fully separate ext-foreign/wlr-foreign

jlindgren90:
Currently, the dependencies between foreign-toplevel[-internal], ext-foreign, and wlr-foreign are cyclical and a bit complex.

I suggest we reorganize it into a simpler hierarchy:
```
  foreign-toplevel/
    -> foreign.c/h
      -> (depends on) ext-foreign.c/h
      -> (depends on) wlr-foreign.c/h
```
The refactored code is smaller and (IMO) easier to follow.

In detail:

- Add `include/foreign-toplevel` folder mirroring `src/foreign-toplevel`
- Split `foreign-toplevel-internal.h` to `ext-foreign.h` and `wlr-foreign.h`
- Eliminate `ext-/wlr-foreign.c` -> `foreign.c` reverse dependencies (including internal signals and `foreign_request*` functions)
- Make `struct foreign_toplevel` private to `foreign.c`

Lightly tested with qmpanel (which uses wlr-foreign-toplevel).


johanmalm:
@Consolatis  - fancy doing the review on this one?

src/foreign-toplevel/ext-foreign.c
```diff
@@ -22,66 +22,59 @@ handle_handle_destroy(struct wl_listener *listener, void *data)
...
 	if (!ext_toplevel->handle) {
 		return;
 	}
 
 	wlr_ext_foreign_toplevel_handle_v1_destroy(ext_toplevel->handle);
+	assert(!ext_toplevel->handle);
 
 	/* Compositor side state changes */
 	wl_list_remove(&ext_toplevel->on_view.new_app_id.link);
 	wl_list_remove(&ext_toplevel->on_view.new_title.link);
```
Consolatis: If the handle gets destroyed from the wlroots side, the finish call would not remove the app_id / title listeners and when triggered we run into an assert in those handlers.

Suggest to move the two to `handle_handle_destroy()` (preferred) or to return early in the handlers if `ext_toplevel->handle` is `NULL`. This seems like a pre-existing issue.

src/foreign-toplevel/wlr-foreign.c
```diff
@@ -123,87 +124,84 @@ handle_new_outputs(struct wl_listener *listener, void *data)
...
-static void
-handle_toplevel_destroy(struct wl_listener *listener, void *data)
+void
+wlr_foreign_toplevel_finish(struct wlr_foreign_toplevel *wlr_toplevel)
 {
-	struct wlr_foreign_toplevel *wlr_toplevel = wl_container_of(
-		listener, wlr_toplevel, on_foreign_toplevel.toplevel_destroy);
+	if (!wlr_toplevel->handle) {
+		return;
+	}
```
Consolatis: Same story here.

src/foreign-toplevel/wlr-foreign.c
```diff
@@ -213,20 +211,15 @@ handle_toplevel_destroy(struct wl_listener *listener, void *data)
 	wl_list_remove(&wlr_toplevel->on_view.minimized.link);
 	wl_list_remove(&wlr_toplevel->on_view.fullscreened.link);
 	wl_list_remove(&wlr_toplevel->on_view.activated.link);
-
-	/* Internal signals */
-	wl_list_remove(&wlr_toplevel->on_foreign_toplevel.toplevel_parent.link);
-	wl_list_remove(&wlr_toplevel->on_foreign_toplevel.toplevel_destroy.link);
 }
 
 /* Internal API */
```
Consolatis, line 193: Nitpick: The internal API includes the finish function above, suggest moving the finish function to the bottom of the file. same for `ext-foreign.c`.

src/foreign-toplevel/wlr-foreign.c
```diff
@@ -123,87 +124,84 @@ handle_new_outputs(struct wl_listener *listener, void *data)
...
-		toplevel->wlr_toplevel.handle, *activated);
+	wlr_foreign_toplevel_handle_v1_set_activated(wlr_toplevel->handle,
+		*activated);
 }
 
-/* Internal signals */
-static void
-handle_toplevel_parent(struct wl_listener *listener, void *data)
+void
+wlr_foreign_toplevel_set_parent(struct wlr_foreign_toplevel *wlr_toplevel,
```
Consolatis: Nitpick: suggest moving this function to the bottom, between `_init()` and `_finish()`.

jlindgren90:
Good finds @Consolatis. I pushed some fix-up commits that should address all the comments - please take a look. I will re-order/squash commits before merge.

Consolatis:
Looks good. Ready to go from my side.

Issue #2938, opened 2025-07-23
New windows no longer open on current output if on the right or below


Tamaranch:
### Problem description

With labwc 0.9.0, when I run e.g. mousepad in xfce4-terminal on my external monitor, located to the right of or below my laptop's internal monitor, mousepad opens on the internal monitor. On the other hand, if the external monitor is located to the left or above the internal monitor, mousepad opens on it.

With labwc 0.8.4, mousepad always opened on the "active monitor", i.e. the one where xfce4-terminal (and also the pointer) is located.

Is this a bug or the new behavior expected after the switch to wlroots 0.19? Thanks!

### Steps to reproduce

1. In a multi-monitor setup
2. Open a graphics application from a terminal located on a right or bottom monitor

### labwc build source

Release

### labwc version

0.9.0 or 0.9.0+16+gcb79ecce

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

tokyo4j:
I cannot reproduce the issue, but perhaps it's a labwc/wlroots bug due to some changes in how wlroots manages outputs.

Can you reproduce the issue if you launch labwc with the default configurations by `labwc -C /tmp`?

Also, output from `labwc -d 2> log.txt` might be helpful.

Tamaranch:
> Can you reproduce the issue if you launch labwc with the default configurations by `labwc -C /tmp`?

Yes I can, here is the logs when performing steps above: 

[labwc-logs.txt](https://github.com/user-attachments/files/21392135/labwc-logs.txt)

tokyo4j:
I'm still investigating the issue, but these lines stand out:

```
00:00:22.993 [INFO] [../src/xdg.c:281] client (mousepad.orig) did not respond to configure request in 100 ms
00:00:23.000 [INFO] [../src/xdg.c:311] using fallback position
```

This indicates `mousepad` doesn't submit a new buffer fast enough when labwc requests a new or initial window size.

@Tamaranch Does the same issue happen for other applications that run fast enough (e.g. `foot`)?

Tamaranch:
Yes, if the window opens quickly enough it seems to end up on the right monitor (tested with gitk, which is very light). But 100 ms is short when you have a small machine, or when it's doing other things... I've just had the same problem when opening the first terminal with super+enter, because a compilation was running in parallel on X11.

jlindgren90:
[...]

PR #2976, opened 2025-08-06
Fix xdg-shell views moving between outputs due to configure timeout

jlindgren90:
Fixes: #2938

Either commit will actually fix the issue with mousepad, but I think we should do both:

- We shouldn't be setting a fallback position at all for views that aren't mapped yet. Just let the map handler take care of it.
- In the corner case where we really do need a fallback position (the initially-maximized case), we can still make the fallback position relative to the view's current output to reduce the surprise factor.

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -288,6 +288,14 @@ handle_configure_timeout(void *data)
 	view->pending_configure_serial = 0;
 	view->pending_configure_timeout = NULL;
 
+	/*
+	 * No need to do anything else if the view is just being slow to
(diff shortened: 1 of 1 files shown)


tokyo4j:
LGTM. Thanks for working on this.

So this PR fixes:
- New, slow windows are placed at the primary monitor even with muti-monitor setup (#2938)
- Initially-maximized, slow windows (e.g. Thunderbird, according to the comment in `handle_configure_timeout()`) are warped to the primary monitor when unmaximized.

PR #2979, opened 2025-08-07
ime: fix segfault when IME is killed

### LINKED ISSUE #2978: Segfault when killing IME
### Problem description

#2894 caused a regression that killing IME client like `fcitx5` causes segfault in labwc.

The cause is that `relay->input_method` is destroyed before `replay->input_method->keyboard_grab` is destroyed, so `replay->input_method` is `NULL` in keyboard grab's destroy handler.

https://github.com/labwc/labwc/blob/d87ef7568a97f03a84061feb94c7ce1dd37c11da/src/input/ime.c#L307-L308

I'll fix that soon.

### Steps to reproduce

1. Start `fcitx5`
2. Kill it
3. Labwc crashes

### labwc build source

Release

### labwc version

0.9.0-68-ge530f437-dirty (+xwayland +nls +rsvg +libsfdo)

### labwc environment

### Distribution

_No response_

tokyo4j:
Fixes #2978 by partially reverting #2894.

Perhaps we can revert this PR again in wlroots 0.20 by reordering the emitted signals, I'm not sure (ref: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5107#note_3042631)

### DIFF
--- a/src/input/ime.c
+++ b/src/input/ime.c
@@ -304,8 +304,8 @@ handle_keyboard_grab_destroy(struct wl_listener *listener, void *data)
 {
 	struct input_method_relay *relay =
 		wl_container_of(listener, relay, keyboard_grab_destroy);
-	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
-		relay->input_method->keyboard_grab;
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab = data;
+	assert(keyboard_grab);
 
 	wl_list_remove(&relay->keyboard_grab_destroy.link);
 


src/input/ime.c
```diff
@@ -304,9 +304,7 @@ handle_keyboard_grab_destroy(struct wl_listener *listener, void *data)
 {
 	struct input_method_relay *relay =
 		wl_container_of(listener, relay, keyboard_grab_destroy);
-	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab =
-		relay->input_method->keyboard_grab;
-
+	struct wlr_input_method_keyboard_grab_v2 *keyboard_grab = data;
```
Consolatis, line 307: Suggest adding an `assert(data)` or `assert(keyboard_grab)` here to ensure that we remember to handle this when it is changed in future wlroots releases.

Consolatis (commented):
Small future-proofing nitpick below. Other than that: LGTM

src/input/ime.c
tokyo4j: Thanks. Added `assert(keyboard_grab)`.

Issue #2983, opened 2025-08-09
crash on set display scale < 1:  labwc cairo_surface_set_device_scale: Assertion `status == CAIRO_STATUS_SUCCESS' failed


kvaradhan3:
### Problem description

When I set device scale to less than 1, using kanshi or way-displays, labwc crashes.

Details:
arch linux, current.
labwc>0.9  (also seen with AUR package, labwc-noxwayland>0.9),

Cause the crash by opening a terminal window, using W-Enter.

The problem is *not* seen with labwc=0.8.4
The problem is *not* tied to display types, I tested with various laptop and external displays.
The problem is *not* tied to kanshi, the same problem can be created with way-displays. 

### Steps to reproduce

1.` Use the following setup: 
```
% cat ~/.config/labwc/autostart
/usr/bin/kanshi
```
and
```
% cat kanshi/config
output eDP-1 {
  scale 0.99
}

profile crashme {
  output eDP-1 enable
}
```

2. Start `labwc` from the tty.
3. Open a terminal window
*viola*

### labwc build source

Release

### labwc version

0.9.1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

arch linux

kvaradhan3:
`labwc -d >/tmp/labwc-debug.log 2>&1` output is here:  
https://gist.github.com/kvaradhan3/25b1b05767711c0144a4665064936328

kvaradhan3:
Appears loosely related to #2947

Consolatis:
Thanks for the report!

Could reproduce nested with `wlr-randr --output WL-1 --scale 0.99`.
The issue is that `img_to_buffer()` in `src/common/scaled-icon-buffer.c` uses an `int scale` argument rather than `double scale`. This results in us telling cairo that the surface scale is `0` which cairo apparently doesn't like. This also means that this is also a bug outside of this assert as we may basically end up rendering smaller icons than necessary and then scale them up again if using a fractional output scale.

Could you give #2984 a try?

kvaradhan3:
> Thanks for the report!
> 
> Could you give [#2984](https://github.com/labwc/labwc/pull/2984) a try?

Yes, this works.  Thanks for the quick fix.

PR #2984, opened 2025-08-09
scaled-icon-buffer: prevent accidental downcasting of scale

Consolatis:
This causes blurry icons on non-integer scales and triggers an assert within cairo when using a output scale < 1.

Fixes:
- #2983


tokyo4j:
Thanks!

PR #2987, opened 2025-08-09
osd: add null-check for output->osd_scene.tree

tokyo4j:
`wlr_scene_node_destroy(&output->osd_scene.tree->node);` does nothing for null pointer, but ASAN is unhappy about accessing `(NULL)->node`, saying: `runtime error: member access within null pointer of type 'struct wlr_scene_tree'`.


johanmalm:
Thanks

PR #2988, opened 2025-08-09
labnag: return non-zero on timeout

johanmalm:
...to avoid inadvertently interpreting timeout as "Yes"

### DIFF
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -29,13 +29,12 @@
 #include <wayland-client.h>
 #include <wayland-cursor.h>
 #include <wlr/util/log.h>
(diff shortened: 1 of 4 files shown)


tokyo4j:
Should we execute `<else>` branch on timeout? I think we can update `action_check_prompt_result()` to do nothing when the exit code 254 is received.

johanmalm:
> Should we execute `<else>` branch on timeout? I think we can update `action_check_prompt_result()` to do nothing when the exit code 254 is received.

That's a good call. Have implemented as suggested. Have tested and seems to work fine. Much better :smile:

tokyo4j:
Thanks!

PR #2991, opened 2025-08-12
docs/menu.xml: adjust labnag example

Narrat:
It was missing the respective closing tag
____________________
Something I noticed when I tried to use the example for `labnag` in the menu.
Unless this syntax should also work?


johanmalm:
Oh, thank you. Good catch.

PR #2993, opened 2025-08-13
ssd: clean up scene management

tokyo4j:
Our codebase for ssd scenes has grown with a lot of technical debts:
- We needed to call `ssd_get_part()` everywhere to get the scene node of a ssd part. We then needed to cast it to `wlr_scene_rect` and `wlr_scene_buffer`. This bloated our codebase and even blocked duplicated button types in `<titlebar><layout>`.
- `ssd_get_part_type()` was a dirty hack. It compared parent, grandparent and grandgrandparent of a node with each subtree in the ssd to get the part type of the node.

To resolve this issues, this commit changes how ssd scenes are managed:
- Access scene rects and scene buffers just as a member of `struct ssd`.
- `ssd_part` is now a attachment to a scene node that can be accessed via `node_descriptor->data`, with a new node-descriptor type `LAB_NODE_DESC_SSD_PART`. `LAB_NODE_DESC_SSD_BUTTON` is unified into it.

Now the scene graph under ssd->tree looks like below. The parentheses indicate the type of ssd_part attached to the node:

```
ssd->tree (LAB_SSD_NONE)
+--titlebar (LAB_SSD_PART_TITLEBAR)
|  +--inactive
|  |  +--background bar
|  |  +--left corner
|  |  +--right corner
|  |  +--title (LAB_SSD_PART_TITLE)
|  |  +--iconify button (LAB_SSD_BUTTON_ICONIFY)
|  |  |  +--window icon image
|  |  |  +--normal close icon image
|  |  |  +--hovered close icon image
|  |  |  +--...
|  |  +--maximize button (LAB_SSD_BUTTON_MAXIMIZE)
|  |  |  +--...
|  |  +--...
|  +--active
|     +--...
+--border
|  +--inactive
|  |  +--top
|  |  +--...
|  +--active
|     +--top
|     +--...
+--shadow
|  +--inactive
|  |  +--top
|  |  +--...
|  +--active
|     +--top
|     +--...
+--extents
   +--top
   +--...
```

When hovering on SSD, `get_cursor_context()` traverses this scene node from the leaf. If it finds a `ssd_part` attached to the node, it returns `ssd_part_type` that represents the resizing direction, button types or `Title`/`Titlebar`.

This PR is ready for review, but I'm opening this PR as draft since it needs more testing. I'll daily drive this PR for a week or so.


src/config/rcxml.c
```diff
@@ -154,14 +154,6 @@ fill_section(const char *content, struct wl_list *list, uint32_t *found_buttons)
 
 		assert(type != LAB_SSD_NONE);
 
-		if (*found_buttons & (1 << type)) {
```
johanmalm, line 158: Do we actually want to allow duplicated button types in the titlebar?

johanmalm:
I like the node-descriptor approach.

I have had a look through code and it seems fine to me, although I've not followed all the logic changes, nor tested.

If @Consolatis has time to review this one that might be a good idea as they did a lot of the original work on it? Otherwise I'll do a slightly more thorough 'final' review in a few days.

src/config/rcxml.c
tokyo4j: Actually I don't have much preference on it.

If you'd like to keep prohibiting duplicated button types, we can instead just add a comment like `This check is no longer needed, but let's keep it just in case`.

johanmalm: Yeah, I think I prefer that. We can always remove it if there is a use-case.

tokyo4j: Done.

tokyo4j:
Not having observed any regressions so far, I'm un-drafting PR.

I think there's no need to rush - I don't even mind delaying this PR to next release cycle.

include/ssd-internal.h
```diff
@@ -117,16 +156,23 @@ struct ssd {
...
 
 	/* This part represented in scene graph */
 	struct wlr_scene_node *node;
+	struct wl_listener node_destroy;
 
-	struct wl_list link;
+	/* Set when type == LAB_SSD_BUTTON_* */
+	struct ssd_button *button;
+	struct wl_list link; /* ssd_titlebar_subtree.buttons_{left,right} */
 };
```
Consolatis: This is `struct ssd_part`, should we omit these two members and instead create a new struct like
```c
struct ssd_part_button {
	struct ssd_part base;
	struct ssd_button *button;
	struct wl_list link;
}
```
Or as alternative include `struct ssd_part base` as first entry in `struct ssd_button`?
These approaches would prevent having to add optional elements to `struct ssd_part` for different parts.

I don't feel strongly about it though.

src/ssd/ssd-titlebar.c
```diff
@@ -309,34 +306,31 @@ ssd_titlebar_update(struct ssd *ssd)
...
 		x = width - corner_width;
-		part = ssd_get_part(&subtree->parts, LAB_SSD_PART_TITLEBAR_CORNER_RIGHT);
-		wlr_scene_node_set_position(part->node, x, -rc.theme->border_width);
+		wlr_scene_node_set_position(&subtree->corner_right->node,
+			x, -rc.theme->border_width);
 
 		x = width - theme->window_titlebar_padding_width + theme->window_button_spacing;
-		wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
-			part = ssd_get_part(&subtree->parts, b->type);
+		wl_list_for_each(part, &subtree->buttons_right, link) {
```
Consolatis: This changes from `for_each_reverse()` to `for_each()`, intended?

src/ssd/ssd-titlebar.c
```diff
@@ -434,22 +410,20 @@ ssd_update_title_positions(struct ssd *ssd, int offset_left, int offset_right)
...
-		struct ssd_part *part = ssd_get_part(&subtree->parts, b->type);
+	struct ssd_part *part;
+	wl_list_for_each(part, &subtree->buttons_left, link) {
 		if (part->node->enabled) {
 			*offset_left += button_width + button_spacing;
 		}
 	}
-	wl_list_for_each_reverse(b, &rc.title_buttons_right, link) {
-		struct ssd_part *part = ssd_get_part(&subtree->parts, b->type);
+	wl_list_for_each(part, &subtree->buttons_right, link) {
```
Consolatis: Same here

include/ssd-internal.h
```diff
@@ -80,33 +101,51 @@ struct ssd {
...
 		struct ssd_state_title {
 			char *text;
-			struct ssd_state_title_width active;
-			struct ssd_state_title_width inactive;
+			/* indexed by enum ssd_active_state */
+			struct ssd_state_title_width dstates[2];
 		} title;
+
+		char *app_id;
 	} state;
```
Consolatis: What is the app_id for? On a quick scan I didn't find it used anywhere but its possible I missed it.

tokyo4j: Oh, thanks. I guess that's just a cruft from old patch before we added `scaled_icon_buffer`. I'll remove it.

tokyo4j: Removed

src/ssd/ssd-titlebar.c
tokyo4j: Yes, that's intended. Packing from `rc.title_buttons_{left,right}` to `ssd.titlebar.subtrees[].buttons_{left,right}` reverses the order.

include/ssd-internal.h
tokyo4j: I like your idea. What do you think of my latest fixup commit? It adds about 20 lines of code, but if it makes the code easier to follow, I'm all for it.

src/ssd/ssd-titlebar.c
Consolatis: Alright, just wanted to make sure its an intended change.

include/ssd-internal.h
Consolatis: Looks good

kode54:
This needs to be rebased.

Edit: All that needs to be changed is that `ssd_resize_edges` now returns `enum lab_edge`

tokyo4j:
Rebased and squashed the fixup commits.

johanmalm:
Thanks. LGTM. Have done some testing and it seems fine.

PR #2994, opened 2025-08-13
 menu: allow client-{list-combined,send-to}-menu as submenu of static menu

tokyo4j:
This PR allows `client-list-combined-menu` and `client-send-to-menu` as submenus of static menus by creating them first.

Their labels are changed to "Windows" and "Send to desktop" which are the same as Openbox.

This PR also fixes my stupid mistake in #2971 that broke `client-send-to-menu`.

Link: #2992

PR #2995, opened 2025-08-13
menu: use client-send-to-menu as 'Workspace' submenu

johanmalm:
...because that is more flexible and how it is in openbox.

I have had in mind that we should do this since the original implementation, and #2994 just jogged my memory.

20250813_20h41m09s_grim

### DIFF
--- a/docs/menu.xml
+++ b/docs/menu.xml
@@ -25,18 +25,7 @@
     Any menu with the id "workspaces" will be hidden
     if there is only a single workspace available.
   -->
-  <menu id="workspaces" label="Workspace">
-    <item label="Move Left">
-      <action name="SendToDesktop" to="left" />
-    </item>
-    <item label="Move Right">
-      <action name="SendToDesktop" to="right" />
-    </item>
-    <separator />
-    <item label="Always on Visible Workspace">
-      <action name="ToggleOmnipresent" />
-    </item>
-  </menu>
+  <menu id="client-send-to-menu"/>
   <!--
     openbox default workspace selector
     to use replace above workspace menu with the example below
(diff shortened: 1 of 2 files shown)


johanmalm:
Added a couple of commit to update documentation and fix a memory leak (not caused by this patch set).

Ready for final review.

tokyo4j:
Thanks for fixing the leak. I didn't understand the difference between `buf_clear()` and `buf_reset()` :sweat_smile:

In #2994, I used "Send to desktop" as the fixed label for `client-send-to-menu`. To keep consistency, I think we can rename it to "Workspace" or use "Send to desktop" also in the builtin root menu. My preference is the former, since we already have many translations for "Workspace".

johanmalm:
> Thanks for fixing the leak. I didn't understand the difference between `buf_clear()` and `buf_reset()` 😅

I was running with ASAN, so noticed it that way :smile:

> In #2994, I used "Send to desktop" as the fixed label for `client-send-to-menu`. To keep consistency, I think we can rename it to "Workspace" or use "Send to desktop" also in the builtin root menu. My preference is the former, since we already have many translations for "Workspace".

Agree. Have pushed fix.

Ready to go if you agree.

docs/menu.xml
```diff
@@ -25,18 +25,7 @@
...
-    </item>
-    <item label="Move Right">
-      <action name="SendToDesktop" to="right" />
-    </item>
-    <separator />
-    <item label="Always on Visible Workspace">
-      <action name="ToggleOmnipresent" />
-    </item>
-  </menu>
+  <menu id="client-send-to-menu" label="Workspace"/>
```
tokyo4j: I think we should remove `label="Workspace"` since otherwise it's recognized as a new menu definition rather than a link to `client-send-to-menu`.

tokyo4j:
LGTM other than my comment above.

Thanks for working on this. This is indeed more helpful than only moving to left/right workspace.

johanmalm:
Thanks for review.

PR #2996, opened 2025-08-14
chase wlroots: Add wl_fixes interface

kode54:
New fixes interface added upstream, for repairing bugs that can't simply be fixed by changing the affected interfaces.


tokyo4j:
You also need to add `wl_fixes` to `allowed_protocols` array above.

kode54:
Where in the list do you think it should go? It currently is allocated as a base protocol, but it only exists thus far to clean up instances of the registry on display destruction.

jlindgren90:
So, https://en.wikipedia.org/wiki/XFixes all over again? :)

kode54:
> So, https://en.wikipedia.org/wiki/XFixes all over again? :)

It's a little too late to change the registry interface, isn't it?

tokyo4j:
Thanks, LTGM.

> > So, https://en.wikipedia.org/wiki/XFixes all over again? :)
> 
> It's a little too late to change the registry interface, isn't it?

Here's the reason from mahkoh: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/388#note_2504541

jlindgren90:
> > So, https://en.wikipedia.org/wiki/XFixes all over again? :)
> 
> It's a little too late to change the registry interface, isn't it?

Yes, understood, I just found the analogy with XFixes humorous, because XFixes was the same way: improvements/fixes that it was too late to change in the core X11 protocol.

kode54:
> > > So, https://en.wikipedia.org/wiki/XFixes all over again? :)
> > 
> > 
> > It's a little too late to change the registry interface, isn't it?
> 
> Yes, understood, I just found the analogy with XFixes humorous, because XFixes was the same way: improvements/fixes that it was too late to change in the core X11 protocol.

Hehe, indeed. Alas, history repeats itself.

PR #2998, opened 2025-08-14
menu: allow overwriting submenu icon

tokyo4j:
This PR allows overwriting the icon of item linking to another menu like below (the icon for `krita` instead of `mpv` should be shown):

```xml
<openbox_menu>
  <menu id="static-menu" label="Static Menu" icon="mpv" />
  <menu id="root-menu" label="Root">
    <menu id="static-menu" icon="krita" />
  </menu>
</openbox_menu>
```

For comparison, Openbox seems to ignore `icon` property in toplevel menu definition in the first place and always follows `icon` property inside the `<menu>` that links to a static menu. So when we have following `menu.xml`:

```xml
<openbox_menu>
  <menu id="static-menu" label="Static Menu"
    icon="/usr/share/icons/hicolor/32x32/apps/mpv.png" />
  <menu id="root-menu" label="Root">
    <menu id="static-menu"
      icon="/usr/share/icons/hicolor/22x22/apps/krita.png" />
    <menu id="inline-menu" label="Inline Menu"
      icon="/usr/share/icons/hicolor/32x32/apps/chromium.png" />
  </menu>
</openbox_menu>
```

|Openbox|labwc 0.9.0|This PR|
|-|-|-|
| 20250815_02h59m17s_grim | 20250815_03h06m42s_grim | 20250815_03h11m59s_grim |

So I think this PR enhances compatibility to Openbox.

This PR also fixes my mistake in #2971 (`s/parent->icon/menu->icon/`) that showed incorrect icon in an item linking to another menu.

Link: https://github.com/labwc/labwc/discussions/2992#discussioncomment-14101404

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -613,7 +613,7 @@ fill_menu(struct server *server, struct menu *parent, xmlNode *n)
 		}
 
 		struct menuitem *item = item_create(parent, menu->label,
-			parent->icon_name, true);
+			icon_name ? icon_name : menu->icon_name, true);
 		item->submenu = menu;
 	}
 error:


johanmalm:
Thanks for analysis and fix. Makes sense.

LGTM

PR #2999, opened 2025-08-15
include: add common/enum.h and reduce header dependencies

jlindgren90:
(Another cleanup PR, maybe a little OCD. Let me know if it's getting to be too much.)

Currently, many enums are defined in a specific module header (such as `rcxml.h` or `view.h`) even though they would be useful in others also. This leads to an awkward tradeoff where either:

1. additional header dependencies are created just to pull in the enum (example: `view.h` depending on `rcxml.h` just for `enum view_placement_policy`), or
2. to avoid additional dependencies, a generic type is used instead of the enum (for example, `rcxml.h` defining `rc.window_switcher.criteria` as `uint32_t` rather than `enum lab_view_criteria` to avoid a cyclical dependency on `view.h`).

We also have one case where an entire header (`three-state.h`) exists just to define a single enum. I suggest it would be better to have a common location for enums that are used across multiple labwc headers, and propose `common/enum.h`.

It may be easiest to review the changes commit-by-commit. The last commit just shifts #includes around and adds a lot of noise to the overall diff.


tokyo4j:
I prefer placing enums inside the most relevant header files (e.g. `enum ssd_part_type` in `ssd.h`). I agree that we should try to include as fewer headers as possible to reduce compilation time and the chances of circular dependencies, but I'd rather like to see all the enum/struct/functions declarations about ssd inside `ssd.h`, and not deviate from other major C projects like wlroots, sway and linux kernel (or have you seen `enum.h` before?).

> 2. to avoid additional dependencies, a generic type is used instead of the enum (for example, rcxml.h defining rc.window_switcher.criteria as uint32_t rather than enum lab_view_criteria to avoid a cyclical dependency on view.h).

I don't think `uint32_t` is used for `rc.window_switcher.criteria` to avoid additional dependencies, but it's just that we use `uint32_t` for a bitmap of enums. I personally prefer using `uint32_t` for bitmaps of enums so that we can make sure we catch all the cases in `switch` statement. Even if we decide to use enum for bitmaps, I think we can just remove `#include "config/rcxml.h"` from `view.h` to avoid the circular dependency in this case.

> We also have one case where an entire header (three-state.h) exists just to define a single enum. I suggest it would be better to have a common location for enums that are used across multiple labwc headers, and propose common/enum.h.

I agree that it's strange to have a separate header for just defining a single enum. I think we can instead have `util.h` for that kind of miscellaneous enums, functions and macros (e.g. `file_exists()` in `common/file-helpers.h`).

jlindgren90:
> I prefer placing enums inside the most relevant header files (e.g. enum ssd_part_type in ssd.h).

Okay, but we have a number of enums that are set in the config (rcxml) but control the behavior of other modules (ssd, view, etc.) What is the "most relevant" header for those? We haven't been consistent about picking one.

> it's just that we use uint32_t for a bitmap of enums.

I wasn't aware of that convention and haven't followed it with e.g. view_axis. Maybe it should be added to CONTRIBUTING.md.

> I think we can just remove #include "config/rcxml.h" from view.h

In that case ~lab_view_criteria~ view_placement_policy needs to be moved.

jlindgren90:
> I prefer placing enums inside the most relevant header files (e.g. enum ssd_part_type in ssd.h).

On the topic of ssd_part_type in particular, that enum is misnamed. It should be "lab_node_type" or similar since it includes many node IDs (OSD, layer-shell, etc.) that have nothing to do with server-side decorations.

jlindgren90:
> not deviate from other major C projects like wlroots, sway and linux kernel (or have you seen enum.h before)?

Having a separate header for common or shared types/defines is pretty common (no pun intended). Think e.g. stddef.h which many other headers pull in for NULL and size_t. Or glib/gtypes.h.

I called it enum.h because most of our basic shared types are enums, but it could also be types.h or whatever.

tokyo4j:
> I wasn't aware of that convention and haven't followed it with e.g. view_axis. Maybe it should be added to CONTRIBUTING.md.

Storing a bitmap of `view_axis` not in `uint32_t` is fine, since that type has `VIEW_AXIS_BOTH = (VIEW_AXIS_HORIZONTAL | VIEW_AXIS_VERTICAL)` and static analyzer warns us if we forget to catch `case VIEW_AXIS_BOTH` in `switch` statement.

> On the topic of ssd_part_type in particular, that enum is misnamed. It should be "lab_node_type" or similar since it includes many node IDs (OSD, layer-shell, etc.) that have nothing to do with server-side decorations.

Hm, I agree that having `LAB_SSD_OSD` is quite strange and maybe we should rename and move it elsewhere. I'd like to move it to `include/input/cursor.h` instead, but moving it to a common header like this PR might be an option.


> Okay, but we have a number of enums that are set in the config (rcxml) but control the behavior of other modules (ssd, view, etc.) What is the "most relevant" header for those? We haven't been consistent about picking one.
> Having a separate header for common or shared types/defines is pretty common (no pun intended). Think e.g. stddef.h which many other headers pull in for NULL and size_t. Or glib/gtypes.h.

`stddef.h` and `gtypes.h` define pretty basic types like integers without much context, whereas our enums like `enum view_edge` and `enum view_placement_policy` is specific to how we manage windows. I don't think we should put them to such a "common" header.

tokyo4j:
> I wasn't aware of that convention and haven't followed it with e.g. view_axis. Maybe it should be added to CONTRIBUTING.md.

Storing a bitmap of `view_axis` not in `uint32_t` is fine, since that type has `VIEW_AXIS_BOTH = (VIEW_AXIS_HORIZONTAL | VIEW_AXIS_VERTICAL)` and static analyzer warns us if we forget to catch `case VIEW_AXIS_BOTH` in `switch` statement.

> On the topic of ssd_part_type in particular, that enum is misnamed. It should be "lab_node_type" or similar since it includes many node IDs (OSD, layer-shell, etc.) that have nothing to do with server-side decorations.

Hm, I agree that having `LAB_SSD_OSD` in `ssd_part_type` is quite strange and maybe we should rename and move it elsewhere. I'd suggest renaming it to `enum mouse_context` and moving it to `include/input/cursor.h` after #2993 is merged.

> Okay, but we have a number of enums that are set in the config (rcxml) but control the behavior of other modules (ssd, view, etc.) What is the "most relevant" header for those? We haven't been consistent about picking one.

> Having a separate header for common or shared types/defines is pretty common (no pun intended). Think e.g. stddef.h which many other headers pull in for NULL and size_t. Or glib/gtypes.h.

Well, there's not a definitive criteria for "most relevant" header, but enums like `view_placement_policy` is quite specific to how we manege windows or our configuration, whereas `stddef.h` and `gtypes.h` define pretty basic types like integers without much context.

I'm not against creating a common header. Especially, `enum view_edge` is no longer specific to window edges and they are now used to simply represent directions, so it makes sense to rename it (e.g. `lab_direction`) and move it to a common header.

CC @labwc/compositor-team

jlindgren90:
> Well, there's not a definitive criteria for "most relevant" header, but enums like view_placement_policy is quite specific to how we manege windows or our configuration, whereas stddef.h and gtypes.h define pretty basic types like integers without much context.

Interesting that you mention "our configuration" as a specific category. Many of the enums that I moved do represent config settings. If we put those in e.g. `config/types.h` would it be a fair compromise? The header is then a little more specific (configuration types) but still allows `rcxml.h` to pull them from a single location without including the whole world or (vice versa) the whole world including `rcxml.h`.

johanmalm:
> I'm not against creating a common header. Especially, `enum view_edge` is no longer specific to window edges and they are now used to simply represent directions, so it makes sense to rename it (e.g. `lab_direction`) and move it to a common header.

> If we put those in e.g. `config/types.h` would it be a fair compromise?

It feels like a sensible balance is the answer here rather than too far down either route. I suspect many of our enums started of within a fairly small scope and then spread, so refactoring includes files to reflect that makes sense. Suggest doing it bit by bit, but yeah the `config/types.h` ones seems to make a lot of sense.

I think the enums names also need to be re-visited from time to time if they become more general in nature. The edge/direction/border types ones are good examples of this.

jlindgren90:
Updated to:
- add `config/types.h` (instead of `common/enum.h`)
- use `lab_` and `LAB_` prefixes consistently

include/config/types.h
```diff
@@ -0,0 +1,118 @@
...
+	LAB_EDGE_UPLEFT = (LAB_EDGE_UP | LAB_EDGE_LEFT),
+	LAB_EDGE_UPRIGHT = (LAB_EDGE_UP | LAB_EDGE_RIGHT),
+	LAB_EDGE_DOWNLEFT = (LAB_EDGE_DOWN | LAB_EDGE_LEFT),
+	LAB_EDGE_DOWNRIGHT = (LAB_EDGE_DOWN | LAB_EDGE_RIGHT),
+};
+
+enum lab_motion {
+	LAB_MOTION_ABSOLUTE = 0,
+	LAB_MOTION_RELATIVE,
+};
```
tokyo4j, line 43: `lab_motion` sounds like it represents a motion itself. I suggest `enum motion_event_mode` (or `lab_motion_event_mode`) instead and adding a comment `/* represents how a input device reports motion events */`.

tokyo4j:
LGTM other than my comment above.

I like how it is now clearer what types should be put in the common header, thanks to the file name `config/types.h` and the comment at the top there.

include/config/types.h
jlindgren90: Happy to add the comment, but I suggest keeping `lab_motion` for the name, or possibly `lab_tablet_motion`.

Rationale: the field in `rc.xml` is just `motion`, which is parsed by `tablet_parse_motion()` and stored in a variable named `motion`. So I think it makes sense to be consistent and use the just term `motion` here also, or possibly `tablet_motion`, which is still consistent but adds a namespace for context.

jlindgren90: (Note: it was just `enum motion` before, which seems like asking for a name collision.)

tokyo4j: OK, I'm fine with it. I was a bit afraid that the context of `enum motion` become unclearer by migrating it from `input/tablet.h`, but we can rename it later if name collision happens.

tokyo4j:
Thanks.

I think we should also rename `view_edge_invert()` and `view_edge_parse()` and move it to somewhere else (e.g. `src/common/edge.c`). But then I'm a bit afraid that it's confused with `src/edges.c`. So let's do it later.

PR #3000, opened 2025-08-15
input/cursor: fix assignment/equality mix-up

jlindgren90:
Self-explanatory, I think.

edit: also, cheers to PR 3000 :)


johanmalm:
Good catch. How did you find it?

Yeah, can't believe we've got to that number!

jlindgren90:
> How did you find it?

I was fixing a C++ warning nearby (lack of a cast from (void *)) and happened to see it.
Kind of surprised it didn't trigger a compiler warning, actually.

Issue #2997, opened 2025-08-14
Use-after-free when labwc exits while submenu is selected


tokyo4j:
Steps to reproduce:
1. Use following `menu.xml`:
```xml
<openbox_menu>
  <menu id="menu-1">
    <item label="Item in menu-1" />
  </menu>

  <menu id="root-menu">
    <menu id="menu-1" />
  </menu>
</openbox_menu>
```

2. Start labwc
3. Launch terminal emulator
4. Run `sleep 3; labwc -e`
5. Open `root-menu` and select the item in it
6. After 3 seconds, labwc crashes with segfault

<details><summary>Output from ASAN</summary>

```
==104520==ERROR: AddressSanitizer: heap-use-after-free on address 0x7c7f88708f50 at pc 0x55c02f8ad250 bp 0x7fffa9ad6d10 sp 0x7fffa9ad6d00
READ of size 8 at 0x7c7f88708f50 thread T0
    #0 0x55c02f8ad24f in _close ../src/menu/menu.c:1161
    #1 0x55c02f8ad3d8 in _close ../src/menu/menu.c:1166
    #2 0x55c02f8ad628 in menu_close ../src/menu/menu.c:1182
    #3 0x55c02f8b2613 in menu_close_root ../src/menu/menu.c:1562
    #4 0x55c02f8ab6ee in menu_free ../src/menu/menu.c:1042
    #5 0x55c02f8ac1be in menu_finish ../src/menu/menu.c:1075
    #6 0x55c02f761b0c in main ../src/main.c:270
    #7 0x7faf8c227674  (/usr/lib/libc.so.6+0x27674) (BuildId: 1d7c9b9187cec4c03a504fccc917f9af55568a53)
    #8 0x7faf8c227728 in __libc_start_main (/usr/lib/libc.so.6+0x27728) (BuildId: 1d7c9b9187cec4c03a504fccc917f9af55568a53)
    #9 0x55c02f731da4 in _start (/home/hrak/codes/labwc/build/labwc+0x2b0da4) (BuildId: 05a88de34874008062aadcff97bbc97c59ef3418)

0x7c7f88708f50 is located 96 bytes inside of 136-byte region [0x7c7f88708ef0,0x7c7f88708f78)
freed by thread T0 here:
    #0 0x7faf8dd1f79d in free /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:51
[...]

PR #3001, opened 2025-08-15
menu: fix use-after-free at exit with sub-menu selected

jlindgren90:
Sequence of events:

- `menu_finish()` frees the sub-menu first
- the `selection.menu` of the parent menu is now dangling
- `menu_finish()` frees the parent menu
- `menu_free()` calls `menu_close_root()` on the parent menu
- `menu_close_root()` tries to close the (freed) sub-menu
- boom

Extending `nullify_item_pointing_to_this_menu()` avoids the crash.

Fixes:
- #2997


tokyo4j:
Thanks, tested and worked well.

PR #3002, opened 2025-08-15
add app_id and better/different state indicators for client list menu

davidphilipbarr:
adds max, fullscreen and min/iconified states in client-list-combined-menu. Also add app_id to the label to be a bit clearer what is going on. this may not be the best way, it is similar to openbox but extended, not sure if openbox includes the icon in the ( ) for minimised?

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -3,6 +3,8 @@
 <!--
   This file contains all supported config elements & attributes with
   default values.
+
+  Values for [yes|no] can be replaced by [true|false], [on|off] or [1|0].
 -->
 
 <labwc_config>
@@ -609,7 +611,13 @@
       - sendEventsMode [yes|no|disabledOnExternalMouse]
       - calibrationMatrix [six float values split by space]
       - scrollFactor [float]
+   
+    The following <libinput>...</libinput> block may not be complete for
+    your requirements. Default values are device specific. Only set an option
+    if you require to override the default. Valid values must be inserted.
+
   -->
+
   <libinput>
     <device category="default">
       <naturalScroll></naturalScroll>


johanmalm:
I feel this needs some discussion.

I'm not that keen on hard-coding too much of this type of mark-up, but maybe if it's on par with openbox that okay.

Would you mind describing what openbox does in this regard and which bits are "extended".

With the window switch we went with a custom field and printf style markup

davidphilipbarr:
The only behaviour openbox has is round brackets around iconified windows - which I've tried to modify to indicate other states, fullscreen and max are quite niche and only really of use on multi screen setups, but there is no precedent for them. 

I guess the core idea of this PR is to add *some* indicator of iconified and to add the app-id to the label to add some extra clarity. 

I think macos uses a leading outlined diamond, maybe something like this: ♢ for iconified windows or there are utf-8 characters for min/max/close/screen - but I don't know if that is a good idea either?

Perhaps this should just be an 'add iconified state indicator' pr?

johanmalm:
> Perhaps this should just be an 'add iconified state indicator' pr?

I think it's good to have a broad discussion, but in terms of merging it's generally easier with smaller changes :smile:

I've just looked at Openbox and the `(foo)` syntax for minimized looks nice in my opinion.

I generally prefer to not add too many config options unless there is a real need/demand for them. 

If adding the brackets becomes controversial, we can always consider a `<wrapMinimizedWindowsInBraces>` or whatever. Or we could consider re-using or extending the osd-field approach (https://github.com/labwc/labwc/blob/master/src/osd-field.c) with `%x` style formatting.

johanmalm:
I've set this to draft until it's clear where we are going with it.

I'm happy to merge a patch to add brackets around the title for minimized windows (which is then on par with Openbox). For anything more complex like `%s - %s` including the `app_id` I feel this needs a way for users to specify the format because not everyone will like it that way.

Regardless, as a PR is needs:
1. Commits to be squashed
2. One commit per change
3. Sensible commit messages in accordance with CONTRIBUTING.md
4. Fix coding style issues. See CONTRIBUTING.md and the Github CI (or just run `./scripts/check --file=src/menu/menu.c`)

tokyo4j:
I actually don't see the benefit of app_id in `client-list-combined-menu`. Why aren't icons enough?

davidphilipbarr:
This is a fair point, it's more nuanced than I thought. I'll reduce
down and tidy up to brackets around iconified windows using only the title.

src/menu/menu.c
```diff
@@ -907,6 +907,9 @@ update_client_list_combined_menu(struct server *server)
 				if (view == server->active_view) {
 					buf_add(&buffer, "*");
 				}
+				if (view->minimized) {
+				buf_add_fmt(&buffer, "(%s)", title);
```
johanmalm: Could you fix these?

```
src/menu/menu.c:910: WARNING: suspect code indent for conditional statements (32, 32)
src/menu/menu.c:912: WARNING: suspect code indent for conditional statements (32, 32)
src/menu/menu.c:912: CHECK: Unbalanced braces around else statement
```

johanmalm:
@davidphilipbarr I can squash the commits and fix the coding style issues on your branch, but don't want to force-push unless you are happy that I do it.

davidphilipbarr:
I think i've resolved the formatting issue but i'm not sure how to squash, i'm a bit of an idiot :)

johanmalm:
We can squash/merge directly from github, but on the command line you can do something like:

```sh
git rebase -i HEAD~8
```

(8 because you had 8 commits)

Then you edit the file replacing the first word with what you want to do, like `fixup`, `reword`, `squash`, etc.

https://git-scm.com/docs/git-rebase#_interactive_mode

PR #3004, opened 2025-08-16
desktop-entry: fix partial string matching

datMaffin:
Use the base instead of the full string for the comparison.

The desktop_id_base is currently basically unused.

---

I noticed that something was weird when Godot would show the Zeal icon.

Debugging session:

image

---

I think the partial match search could be improved.
The id of the desktop file is: `org.godotengine.Godot4.4`. So there are various difficulties with it finding the correct icon ;).

But I am already pretty happy with it no longer showing a wrong icon, and instead showing the labwc icon (default.)

### DIFF
--- a/src/desktop-entry.c
+++ b/src/desktop-entry.c
@@ -268,7 +268,7 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		int alen = strlen(app_id);
 		int dlen = strlen(desktop_id_base);
 
-		if (!strncasecmp(app_id, desktop_id, alen > dlen ? dlen : alen)) {
+		if (!strncasecmp(app_id, desktop_id_base, alen > dlen ? dlen : alen)) {
 			return entry;
 		}
 	}


jlindgren90:
Previous discussions (edit: added one more):
- #2215
- #2216
- #2266

I am not in favor of the partial string matching, and said already that it would produce incorrect matches. The argument in favor of it seemed to be that that chance of an incorrect icon was better than no icon at all - which I still don't agree with.

datMaffin:
At least with this fix the chance for a wrong icon should now be _very_ unlikely.

johanmalm:
This makes sense to me.

@spl237 We're change the behaviour of one of your patches here (#2266). Would you be happy to review and comment?

spl237:
WIthout trying all my specific test cases - I'm on holiday at the moment so can't do that - I can't say whether this is likely to break them, but it looks sensible enough.

But really, all this ought to be rendered unnecessary if https://wayland.app/protocols/xdg-toplevel-icon-v1 is implemented and used for this purpose instead - has any progress been made on that? I thought I had seen something to suggest it was either underway or completed, but I may have imagined that.

johanmalm:
> WIthout trying all my specific test cases - I'm on holiday at the moment so can't do that - I can't say whether this is likely to break them, but it looks sensible enough.

Shall I go for the merge, and you can do the tests (with gimp, etc) in slow time. Logically, I think it should be fine but there is no substitute for testing. We're unlikely to freeze for the next release until second half of September.

> But really, all this ought to be rendered unnecessary if https://wayland.app/protocols/xdg-toplevel-icon-v1 is implemented and used for this purpose instead - has any progress been made on that? I thought I had seen something to suggest it was either underway or completed, but I may have imagined that.

We have merged `xdg-toplevel-icon` and it's working well, but obviously relies on clients (or associated toolkits) implementing it too. I personally don't use it though because I prefer the uniform look I getting from specifying and using a specific icon theme.

johanmalm:
Let's go for it.

PR #3007, opened 2025-08-17
docs/rc.xml.all: clean up libinput section.

### LINKED ISSUE #3006: warnings/errors due to parsing of the config
### Problem description

My main config is based upon `rc.xml.all`. Since some time I noticed a few red lines (dunno exactly when it started, but still there with 8d60fb27 ).
Did diff the configs for obvious changes, but that didn't solve the "issue" (?).
Then tested the `rc.xml.all` directly and the warnings or errors are also there:

```
$ labwc -c /usr/share/doc/labwc/rc.xml.all
Fontconfig warning: using without calling FcInit() <<--- can be ignored
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/config/rcxml.c:769] invalid tapButtonMap
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/config/rcxml.c:847] invalid clickMethod
00:00:00.015 [../labwc/src/config/rcxml.c:860] invalid scrollMethod
00:00:00.015 [../labwc/src/common/parse-bool.c:29] () is not a boolean value
00:00:00.015 [../labwc/src/config/rcxml.c:885] wrong number of calibration matrix elements, expected 6, got 0
```
It does seem to be warnings in regards to the [libinput](https://github.com/labwc/labwc/blob/master/docs/rc.xml.all#L613-L632) section of the config and it doesn't like the empty settings there.
Does the `rc.xml.all` need an update or is it due to changes to the config parser and shouldn't be thrown?

### Steps to reproduce

1. Launch labwc explictly with the `rc.xml.all` config

### labwc build source

Local build

### labwc version

labwc 0.9.0-93-g8d60fb27 (+xwayland +nls +rsvg +libsfdo)

### labwc environment

Not Applicable

### Distribution

Arch Linux

01micko:
closes #3006

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -3,6 +3,8 @@
 <!--
   This file contains all supported config elements & attributes with
   default values.
+
+  Values for [yes|no] can be replaced by [true|false], [on|off] or [1|0].
 -->
 
 <labwc_config>
@@ -609,7 +611,13 @@
       - sendEventsMode [yes|no|disabledOnExternalMouse]
       - calibrationMatrix [six float values split by space]
       - scrollFactor [float]
+   
+    The following <libinput>...</libinput> block may not be complete for
+    your requirements. Default values are device specific. Only set an option
+    if you require to override the default. Valid values must be inserted.
+
   -->
+
   <libinput>
     <device category="default">
       <naturalScroll></naturalScroll>


docs/rc.xml.all
```diff
@@ -609,7 +609,12 @@
       - sendEventsMode [yes|no|disabledOnExternalMouse]
       - calibrationMatrix [six float values split by space]
       - scrollFactor [float]
-  -->
+
+    Values for [yes|no] can be replaced by [true|false], [on|off] or [1|0]
+   
+    The following <libinput>...</libinput> block is not complete. Valid
```
johanmalm: I agree with commenting out the `<libinput>` section.

Suggest removing the "or keys without values removed." Could say something like "Default values are device specific. Only set an option if you want to override the default.".

tokyo4j:
Let's wait for #3011 which changes how empty values in `<libinput>` are recognized.

01micko:
> Let's wait for #3011 which changes how empty values in `<libinput>` are recognized.

Ok, I can wait, no rush for me.

01micko:
Still, a comment about the alternatives ([1|0] etc) and what @johanmalm suggests as an edit would be worthwhile, but that's a separate PR I can fire in later.

docs/rc.xml.all
tokyo4j: Now that #3011 (which removes the error messages with empty values) is merged, I think we don't need to comment out `<libinput>` section (that's not a strong preference though). Also, the word "only" can be removed from the comment @johanmalm suggested.

I don't think `Values for [yes|no] can be replaced by [true|false], [on|off] or [1|0]` should be here since it's not specific to `<libinput>` section. I suggest moving it to above `<core>`.

tokyo4j:
> Still, a comment about the alternatives ([1|0] etc) and what @johanmalm suggests as an edit would be worthwhile, but that's a separate PR I can fire in later.

You can continue your work in this PR. You can change its title if you'd like.

01micko:
Ok, after a silly typo, ready for review :sweat_smile:

tokyo4j:
Thanks!

PR #3011, opened 2025-08-19
rcxml: allow empty values in <libinput>

tokyo4j:
The first commit restores the old behavior prior to #2667 that empty values in `<libinput>` are ignored.

The second commit implements what I proposed as "future plan" in https://github.com/labwc/labwc/issues/3006#issuecomment-3197033177. Reconfiguring a libinput device with empty values now restores its default values, rather than leaving its old configuration before Reconfigure. This makes `rc.xml` more declarative.

I wonder whether we should allow empty values in other places (e.g. `<margin top="" />`). They were allowed before #2667, but I don't find empty values useful outside of `<libinput>`. If we allow empty values in all the configurations, we will need add `if (string_null_or_empty(identifier)) { continue; }` in `fill_region()`, `fill_window_rule()` etc., or we can skip nodes with an empty value in `LAB_XML_FOR_EACH()` (with additional `skip_empty_node` argument?).

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -723,6 +723,9 @@ fill_libinput_category(xmlNode *node)
 	xmlNode *child;
 	char *key, *content;
 	LAB_XML_FOR_EACH(node, child, key, content) {
+		if (string_null_or_empty(content)) {
+			continue;
+		}
 		if (!strcmp(key, "category")) {
 			/*
 			 * First we try to get a type based on a number of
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM. Not tested.

I like the look of this. Feels like it makes the libinput handling more explicit.

I don't think we need to do the same for other settings. I feel libinput is special in the sense that it handles device defaults rather than compostor defaults.

Looking at the code, I think the "handle empty" could be extended later if a use-case emerges.

I'm not 100% sure that device behaviour will not change on compositor start-up. I "think" it should be fine, but I have been caught out before where libinput devices do not behave as I expect them too.

Suggest we merge now. If there unintended side-effects we've got plenty of time to re-consider approach.

PR #3015, opened 2025-08-21
ssd: allow hiding titlebar on maximization

tokyo4j:
Supersedes #2897.

The 1st commit unifies `view->ssd_enabled` and `view->ssd_titlebar_hidden` into `view->ssd_mode` and adds `view_titlebar_visible()` instead of `view->ssd_titlebar_hidden`. I reordered the entries in `enum ssd_mode` so that we can write `if (view->ssd_mode)` rather than `if (view->ssd_mode != LAB_SSD_MODE_NONE)`.

Then 2nd commit takes @CosmicFusion's patch and implements `<core><disableMaximizedServerDecor>` by checking `rc.disable_maximized_ssd` in `view_titlebar_visible()`.

I believe this is the best way to implement hide-titlebar-on-maximize feature without adding complexity.

@CosmicFusion Sorry for superseding your PR. You will be mentioned in the commit message and NEWS.md.


docs/labwc-config.5.scd
```diff
@@ -186,6 +186,11 @@ this is for compatibility with Openbox.
 	that it is not always possible to turn off client side decorations.
 	Default is server.
 
+*<core><disableMaximizedServerDecor>* [yes|no]
+	Hide server side decoration when a window is maximized.
+	Useful for trying to replicate a KDE window controls setup where window
+	controls merge with the top panel. Default is no.
```
tokyo4j: @CosmicFusion I don't know about the "KDE window controls setup". Could you provide a link to it?

johanmalm:
I think this is good, but let me have a thorough look through. It might take a day or two.

dglava:
I've been one of the people who requested this feature (https://github.com/labwc/labwc/discussions/2685). After a quick test, it seems to work exactly like that. Thanks!

I'm gonna keep testing it to see if there are any issues.

docs/labwc-config.5.scd
jlindgren90: Can we just omit the KDE reference and say "Hide server side decorations (e.g. titlebar) when a window is maximized"? That should be clear enough. Whether there is a panel that shows window controls is kind of irrelevant.

johanmalm:
Agree with jlindgren90 that we ought to drop the KDE reference.

Other than that, LGTM

I like the refactoring in the first commit.

Have tested. Works well.

tokyo4j:
Now I wonder `disableMaximizedServerDecor` is the best name, because I'm afraid `disableMaximizedServerDecor=no` sounds like maximized windows have usual SSDs including titlebar and borders, despite that we already hide borders for maximized windows.

What do you think of `<core maximizedDecoration="titlebar|none" />` (or `maximizedServerDecor`)? And if someone wants to show both titlebar and borders for maximized windows, we can also allow `full` just like the argument in `SetDecorations` action.

docs/labwc-config.5.scd
tokyo4j: Agreed. Removed the KDE reference and took your suggested sentence, though it may be revised later.

johanmalm:
Yeah, I like `<core maximizedDecoration="titlebar|none" />`. It's much cleaner.

tokyo4j:
@johanmalm Please merge this PR if you are happy with my updated documentation.

johanmalm:
Thanks :smile:

tokyo4j:
Sorry, I forgot to update my commit message :cry:

johanmalm:
> Sorry, I forgot to update my commit message 😢

Hey, never mind. I also noticed it afterwards :smile: I've started going an interim update of `NEWS.md` so we can straighten it out there.

PR #3016, opened 2025-08-21
Add common/edge.c

tokyo4j:
Follow-up of #2999.

I'm a bit afraid that `src/common/edge.c` can be confused with `src/edges.c`, but I didn't come up with better names for them.


johanmalm:
LGTM. Thanks.

PR #3019, opened 2025-08-22
common: flesh out enum lab_edge and prefer over wlr_edges/wlr_direction

jlindgren90:
I like the new common/edge.h. I don't like how inconsistently we use it.

Current situation:

 - enum wlr_edges and wlr_direction are designed to be used as bitset, and are defined compatibly

 - enum lab_edge is *also* designed to be used as bitset, but incompatible with the others (LEFT/RIGHT come before UP/DOWN)

 - we use an inconsistent mix of all three *AND* uint32_t (usually with the WLR_EDGE constants rather than the LAB_EDGE constants), and convert between them on an ad-hoc basis, sometimes implicitly

Let's clean this up:

 - flesh out enum lab_edge a little, and reorder it to be compatible with the two wlr enums (check this by static_assert)

 - use lab_edge for all internal edge/direction fields, consistently

 - add lab_edge_is_cardinal() as a sanity check before casting to enum wlr_direction, and then eliminate all of direction.c/h

~~Naming is still awkward because "bottom" is an edge while "down" is a direction -- yet we have (inexplicably) LAB_EDGE_DOWN. I'm leaving this as-is, since there is enough churn already.~~ (edit: decided to rename these)

Instead of "enum wlr_edges direction" (!) we now have "enum lab_edge direction", which is not that much better, but at least we are now clear that we're overloading one enum with two meanings.


jlindgren90:
Thoughts on renaming?

- `LAB_EDGE_UP` -> `LAB_EDGE_TOP`
- `LAB_EDGE_UPLEFT` -> `LAB_CORNER_TOPLEFT`
- etc.

I could do this in a 2nd commit.

jlindgren90:
Renamed to `LAB_EDGE_TOP` (consistent with `WLR_EDGE_TOP`), `LAB_CORNER_TOP_LEFT`, etc.

This caused a conflict in `theme.c` which already had private constants `LAB_CORNER_TOP_LEFT/RIGHT` - I renamed these to `ROUNDED_CORNER_TOP_LEFT/RIGHT` instead.

tokyo4j:
I like this patch.

> Thoughts on renaming?
> 
>  - `LAB_EDGE_UP` -> `LAB_EDGE_TOP`
> 
> - `LAB_EDGE_UPLEFT` -> `LAB_CORNER_TOPLEFT`
> 
> - etc.

`LAB_EDGE_TOP` seems fine to me, but I don't like `LAB_CORNER_TOPLEFT`. It sounds like they are different enum types. I understand calling top-left as an "edge" sounds a bit awkward, but naming consistency within a type is more important to me.

Or we can instead use `enum lab_edges` type containing `LAB_EDGES_TOP` and `LAB_EDGES_TOPLEFT` etc. For reference, Qt has `Qt::Edge` enum for cardinal directions and `Qt::Edges` enum (?) for cardinal+corner directions: https://doc.qt.io/qt-6/qt.html#Edge-enum

src/edges.c
```diff
@@ -258,19 +240,19 @@ compute_edges_visible(const struct wlr_box *view_size,
...
-			edges_visible |= WLR_EDGE_TOP;
+			edges_visible |= LAB_EDGE_TOP;
 		}
 		if (rects[i].x2 == view_rect->x2) {
-			edges_visible |= WLR_EDGE_RIGHT;
+			edges_visible |= LAB_EDGE_RIGHT;
 		}
 		if (rects[i].y2 == view_rect->y2) {
-			edges_visible |= WLR_EDGE_BOTTOM;
+			edges_visible |= LAB_EDGE_BOTTOM;
```
tokyo4j, line 255: I don't think we can use `lab_edge` here, since `edges_visible` may contain both `LAB_EDGE_TOP` and `LAB_EDGE_BOTTOM`. Same goes for other functions/variables. CC @ahesford

jlindgren90: In the most recent commit, I made a note in lab_edge header comment that other combinations (LEFT+RIGHT, TOP+BOTTOM) are allowed.

wlroots uses OR'd combinations (without explicitly listing them) for its edge/direction enums, I think it's okay for us to do the same.

jlindgren90:
> Or we can instead use enum lab_edges type containing LAB_EDGES_TOP and LAB_EDGES_TOPLEFT etc. For reference, Qt has Qt::Edge enum for cardinal directions and Qt::Edges enum (?) for cardinal+corner directions: https://doc.qt.io/qt-6/qt.html#Edge-enum

LAB_EDGES_TOP_LEFT etc. sounds good to me. Prefer to keep the underscore between words unless you feel strongly about it.

Qt has some C++ magic to automatically generate a distinct set-of-enums type from an enum type, with type-safe overloads for the usual bitwise operations. (C++ is also stricter and warns about using integer operations on enums. C is more permissive here.)

(Edit to add: https://doc.qt.io/qt-6/qflags.html)

src/edges.c
jlindgren90: We could also add `LAB_EDGES_LEFT_RIGHT` and `LAB_EDGES_TOP_BOTTOM` but it didn't seem necessary since those combinations are only used in one or two places.

jlindgren90:
Regarding C's permissiveness on storing other values in an enum type, the relevant sections of the standard (C99 draft, I don't have the final version) are:

In 6.3 (Conversions):
> Conversion of an operand value to a compatible type causes no change to the value or the representation.

In 6.7.2.2 (Enumeration specifiers):
> Each enumerated type shall be compatible with char, a signed integer type, or an unsigned integer type. The choice of type is implementation-defined, but shall be capable of representing the values of all the members of the enumeration.

Based on my reading (which seems to agree with the consensus) this means that since `enum lab_edge` includes `LAB_EDGE_NONE` (0) and `LAB_EDGES_ALL` (15), it's compatible with an integer type storing at least values 0 through 15, and thus can safely store those values, even the ones that don't have a matching enum constant.

I would say that storing an OR'd combination of enum values in a enum type is an "old-style" C practice that might be frowned on by some parties (MISRA, C++ committee, etc.) but I don't see an issue with using it in this case. In contexts where not all combinations are allowed/meaningful, we can just include a `default` case or do an explicit check like `lab_edge_is_cardinal()`.

jlindgren90:
I just added the 6 remaining combinations for completeness, should address the concern.

src/edges.c
jlindgren90: Added `LAB_EDGES_LEFT_RIGHT` and `LAB_EDGES_TOP_BOTTOM` to address the concern.

jlindgren90:
Squashed commits and updated commit message.

src/theme.c
```diff
@@ -1202,21 +1193,17 @@ rounded_rect(struct rounded_corner_ctx *ctx)
 	set_cairo_color(cairo, ctx->border_color);
 	cairo_set_line_width(cairo, ctx->line_width);
 	double half_line_width = ctx->line_width / 2.0;
-	switch (ctx->corner) {
-	case LAB_CORNER_TOP_LEFT:
+	if (ctx->corner == ROUNDED_CORNER_TOP_LEFT) {
```
tokyo4j: Why did you replace the `switch` statement with `if` statement? Not a strong preference, but I prefer `switch` statement over `if` statement with `/* ROUNDED_CORNER_TOP_RIGHT */` comment.

tokyo4j:
The whole idea looks good to me now. I'll do a final review in ~2 days.

> I would say that storing an OR'd combination of enum values in a enum type is an "old-style" C practice that might be frowned on by some parties (MISRA, C++ committee, etc.) but I don't see an issue with using it in this case. In contexts where not all combinations are allowed/meaningful, we can just include a default case or do an explicit check like lab_edge_is_cardinal().

The problem for me is that, when a function takes `enum lab_edge` as its argument, I expect only either of the enum entries, not an OR'd combination of them. AFAIR from IRC, that's what @Consolatis expects too.

src/theme.c
jlindgren90: Since the private enum can (provably) have only one of two values, the "if" avoids the need for a dead branch.

I can revert it back to switch though, it's also not a strong preference for me. I'd still remove the dead wlr_log and return though.

jlindgren90:
> The problem for me is that, when a function takes enum lab_edge as its argument, I expect only either of the enum entries, not an OR'd combination of them.

Understood. In the latest version, all the combinations are covered by explicit enum constants, so hopefully the intent is clearer and it avoids surprises.

src/theme.c
tokyo4j: But you removed `LAB_CORNER_UNKNOWN`, so we don't need the dead branch anymore, do we? I prefer switch statement when checking enums like this, because the compiler warns us when we add a new entry to an enum but forget to catch it in switch-cases.

jlindgren90: Sure, let's use a switch with just the two valid cases here

tokyo4j:
LGTM, but please remember to squash the last commit. I didn't find any regressions in my test (e.g. cursor shape around window edges, edge snapping, window resistance).

I really like the unification to `lab_edge`. Dealing with all of `view_edge`, `wlr_direction` and `wlr_edge` was really annoying when I was working on #2885.

jlindgren90:
Thanks, will squash and merge tonight if no one else objects.

Consolatis:
> > I would say that storing an OR'd combination of enum values in a enum type is an "old-style" C practice that might be frowned on by some parties (MISRA, C++ committee, etc.) but I don't see an issue with using it in this case. In contexts where not all combinations are allowed/meaningful, we can just include a default case or do an explicit check like lab_edge_is_cardinal().
> 
> The problem for me is that, when a function takes `enum lab_edge` as its argument, I expect only either of the enum entries, not an OR'd combination of them. AFAIR from IRC, that's what @Consolatis expects too.

Just seen this comment. Yes this sums up my standpoint pretty well.

Issue #3018, opened 2025-08-21
[emacs] Maximize overlooks reserved space when run as a window rule


jberkenbilt:
### Problem description

When I start emacs-pgtk with the following window rule:
```
  <windowRules>
    <windowRule identifier="emacs" matchOnce="true">
      <action name="MoveTo" x="0" y="0" />
      <action name="ToggleMaximize" direction="vertical"/>
    </windowRule>
  </windowRules>
```
it correctly moves to 0, 0 and maximizes vertically, but it covers the entire height of the screen, sitting on top of the panel, which reserves space. Unmaximizing and remaximizing the window correctly honors the reserved space.

As a secondary issue, toggling vertical maximize off returns emacs's height to its original height but also restores its original width if I resized it using the set-frame-width function from emacs. If I manually horizontally resize emacs, toggling vertical maximize only restores vertical width and respects the modified horizontal width. If this is unexpected behavior, I can report a separate issue.

### Steps to reproduce

1. Run a panel that reserves space. For the moment, I am using lxqt-panel.
2. Run a pure gtk build of emacs from a terminal. I use the emacs-pgtk package on Ubuntu 25.04 with above window rule
3. Observe that the emacs window obscures the panel.
4. Toggle vertical maximize; emacs returns to its original height
5. Toggle vertical maximize again; emacs's height goes just to the top of the panel, as expected

I decades past, I did early work in X11 and window managers and am coming up to speed on Wayland. I may take a crack at this and submit a PR, but don't hold your breath.

If the fact that the compositor is not tracking changes made by emacs itself through its own size changing functions for purposes of restoring maximize is also a bug, just say so, and I will create a separate issue. Or I will create a separate issue anyway and  someone can close it.

### labwc build source

Local build

### labwc version

0.9.1 source tarball downloaded from github

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Ubuntu 25.04

jlindgren90:
> it correctly moves to 0, 0 and maximizes vertically, but it covers the entire height of the screen

I'm guessing there's some corner case with the MoveTo and the maximize happening nearly at once, which isn't handled correctly.

> As a secondary issue, toggling vertical maximize off returns emacs's height to its original height but also restores its original width if I resized it using the set-frame-width function from emacs.

To clarify, the `set-frame-width` does change the width (while maximized vertically) but then it reverts back after un-maximizing? That seems like a bug.

[...]

PR #3020, opened 2025-08-23
view: respect client-initiated resize of non-maximized axis

jlindgren90:
When implementing single-axis maximize some time ago, I made the simplifying assumption that a view couldn't be resized while maximized (even in only one axis). And indeed for compositor-initiated resize, we always unmaximize the view first.

However, I didn't account for the client resizing the non-maximized axis, which we can't (and shouldn't) prevent. When this happens, we should also update the natural geometry of that single axis so that we don't undo the resize when un-maximizing.

P.S. xdg-shell clients resizing the *maximized* axis is still an unsolved problem, exacerbated by the fact that xdg-shell protocol doesn't allow clients to even know about single-axis maximize. That's the problem at the root of:
- #3018

P.P.S. the view_invalidate_last_layout_geometry() logic may need similar updates, I'm not sure.

### DIFF
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -90,9 +90,9 @@ interactive_begin(struct view *view, enum input_mode mode, uint32_t edges)
 			return;
 		}
 
+		/* Store natural geometry at start of move */
+		view_store_natural_geometry(view);
 		if (view_is_floating(view)) {
-			/* Store natural geometry at start of move */
-			view_store_natural_geometry(view);
 			view_invalidate_last_layout_geometry(view);
 		}
 
--- a/src/view.c
+++ b/src/view.c
@@ -970,8 +970,11 @@ void
 view_store_natural_geometry(struct view *view)
 {
 	assert(view);
-	if (!view_is_floating(view)) {
-		/* Do not overwrite the stored geometry with special cases */
+	/*
+	 * Do not overwrite the stored geometry if fullscreen or tiled.
+	 * Maximized views are handled on a per-axis basis (see below).
+	 */
+	if (view->fullscreen || view_is_tiled(view)) {
 		return;
 	}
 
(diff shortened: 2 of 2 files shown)


tokyo4j:
It's not very unclear to me what this patch does. I think it's desirable to allow vertically(horizontally) resizing horizontally(vertically)-maximized windows via interactive resize, no matter whether it's initiated by client or server - is it what your patch does? Or is there something special about client-initiated resize?

Edit: another question: is this PR about interactive resize, and not about clients submitting new window geometries? If so, I suggest `s/resize/interactive resize/` in the commit title.

jlindgren90:
> It's not very unclear to me what this patch does. I think it's desirable to allow vertically(horizontally) resizing horizontally(vertically)-maximized windows via interactive resize, no matter whether it's initiated by client or server - is it what your patch does?

This change is a necessary prerequisite for allowing interactive resize in one axis, but does not enable it yet. That requires additional changes in interactive.c.

> Or is there something special about client-initiated resize?

Just the fact that it already happens and isn't handled well. See discussions on #3018 for more background.

jlindgren90:
> is this PR about interactive resize, and not about clients submitting new window geometries?

The reverse of that.

tokyo4j:
LGTM. I skimmed #3018 and understood the purpose of this PR.

I think this PR also fixes vertically maximized windows being moved horizontally. So in summary, this PR does two things below:

1. Remember width of vertically-maximized window resizing itself horizontally:

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/cf3791cd-5558-41cb-9d25-6b13178f967e"/>|<video src="https://github.com/user-attachments/assets/02036320-9609-439a-a39e-050f5559d250" />|

2. Remember x position of vertically maximized window moved horizontally by interactive move etc.:

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/2fc612c5-7a1a-47fd-a6c8-7c50960b2766"/>|<video src="https://github.com/user-attachments/assets/ec81a212-9699-4b98-9d02-134890138214" />|

Both of two follow Openbox. And as you mentioned, I hope we can prevent vertically-maximized windows from being un-maximized when resized horizontally with interactive resize in near future.

> xdg-shell clients resizing the maximized axis is still an unsolved problem, exacerbated by the fact that xdg-shell protocol doesn't allow clients to even know about single-axis maximize.

Perhaps we can use [constrained_left](https://wayland.app/protocols/xdg-shell#xdg_toplevel:enum:state:entry:constrained_left) state etc?

jlindgren90:
> Remember x position of vertically maximized window moved horizontally by interactive move etc.:

I didn't realize we even allowed that - that's a nice bonus fix :)

jlindgren90:
> Perhaps we can use constrained_left state etc?

Worth a try!

PR #3024, opened 2025-08-24
action: add toggle for GoToDesktop

RainerKuemmerle:
Adds an option "toogle" to GoToDesktop.
In case the target equals current, we go back to the last desktop instead if configured.

This is inspired by the tweak available in xfwm4, see https://docs.xfce.org/xfce/xfwm4/wmtweaks:
"Remember and recall previous workspace when switching via keyboard shortcuts".

Example of rc.xml

```xml
<keybind key="C-F1">
  <action name="GoToDesktop">
    <to>1</to>
    <toggle>true</toggle>
  </action>
</keybind>
```

I like this option a lot because I have, e.g., my browser on Desktop 1 and code on another one. Quickly pressing Ctrl-F1 brings me to the browser to read up sth and pressing again Ctrl-F1 allows to be back in code.

If you like the contribution, I'd be happy to contribute also documentation and other parts if needed. First, wanted to understand if you are interested in this small patch.

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -440,6 +440,11 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			action_arg_add_bool(action, argument, parse_bool(content, true));
 			goto cleanup;
 		}
+		if (!strcmp(argument, "toggle")) {
+			action_arg_add_bool(
+				action, argument, parse_bool(content, false));
+			goto cleanup;
+		}
 		break;
 	case ACTION_TYPE_TOGGLE_SNAP_TO_REGION:
 	case ACTION_TYPE_SNAP_TO_REGION:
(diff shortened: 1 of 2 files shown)


src/action.c
```diff
@@ -1228,6 +1233,15 @@ run_action(struct view *view, struct server *server, struct action *action,
 		 */
 		struct workspace *target_workspace = workspaces_find(
 			server->workspaces.current, to, wrap);
+		if (action->type == ACTION_TYPE_GO_TO_DESKTOP) {
+			bool toggle = action_get_bool(
+				action, "toggle", false);
+			// If already on this workspace toggle
```
johanmalm: This is a real nitpik, but we prefer C89 style comments.

```
/* foo */

/*
 * multi
 * line
 */
```

Having said that, given that the comment just describes what the code does - and that it's easy code to read - I'd be happy to just drop the comment.

johanmalm:
Thanks - and welcome to the project :smile:

PR #3026, opened 2025-08-25
window-rules: fix window rules not being applied

tokyo4j:
In 943f5751, I initialized heap-allocated `view_query` used for `If` actions with `decoration=LAB_SSD_MODE_INVALID`, but I forgot to do that for stack-allocated `view_query` used for window rules.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -78,6 +78,7 @@ struct view_query *
 view_query_create(void)
 {
 	struct view_query *query = znew(*query);
+	/* Must be synced with view_matches_criteria() in window-rules.c */
 	query->window_type = -1;
 	query->maximized = VIEW_AXIS_INVALID;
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM

Guess it's because the INVALID enum was previously 0.

Could consider a common init_query() to avoid duplicated code.

01micko:
LGTM - tested, thanks :smile:

tokyo4j:
> Could consider a common init_query() to avoid duplicated code.

Yeah, but I didn't do it because it feels a bit strange to have both `view_query_create()` and `view_query_init()`. Let's revisit when we touch codes around this part in the future.

johanmalm:
> Let's revisit when we touch codes around this part in the future.

Yes, good call. Thanks for fixing this one so quickly.

PR #3028, opened 2025-08-27
seat: improve debug logging when configuring input devices

jlindgren90:
I needed to debug an input configuration issue and found the debug output not-super-helpful, so I made some improvements:

- Print the name and "sysname" (e.g. event11) of the device being configured. Note that the name alone isn't enough since there can be multiple identically-named devices.

- Print the config category matched for each device.

- Print the config values (if any) being applied. For enums, only the numeric value is printed since I'm lazy.

- Don't print "pointer acceleration configured" if neither a pointer speed nor acceleration profile is configured (it's confusing).

### DIFF
--- a/src/seat.c
+++ b/src/seat.c
@@ -149,6 +149,13 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
 	 */
 	assert(dc);
 
+	wlr_log(WLR_INFO, "configuring input device %s (%s)",
+		libinput_device_get_name(libinput_dev),
+		libinput_device_get_sysname(libinput_dev));
+
+	wlr_log(WLR_INFO, "matched category: %s",
+		dc->name ? dc->name : libinput_device_type_name(dc->type));
+
 	libinput_device_config_tap_set_enabled(libinput_dev,
 		libinput_device_config_tap_get_default_enabled(libinput_dev));
 	libinput_device_config_tap_set_button_map(libinput_dev,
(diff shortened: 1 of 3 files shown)


src/seat.c
```diff
@@ -149,14 +149,22 @@ configure_libinput(struct wlr_input_device *wlr_input_device)
...
 	libinput_device_config_tap_set_enabled(libinput_dev,
 		libinput_device_config_tap_get_default_enabled(libinput_dev));
 	libinput_device_config_tap_set_button_map(libinput_dev,
 		libinput_device_config_tap_get_default_button_map(libinput_dev));
 	if (libinput_device_config_tap_get_finger_count(libinput_dev) <= 0) {
 		wlr_log(WLR_INFO, "tap unavailable");
 	} else {
-		wlr_log(WLR_INFO, "tap configured");
+		wlr_log(WLR_INFO, "tap configured (%d, %d)", dc->tap,
+			dc->tap_button_map);
```
tokyo4j:
```suggestion
		wlr_log(WLR_INFO, "tap configured (tap=%d, button_map=%d)",
			dc->tap, dc->tap_button_map);
```

Not a strong preference, but I'd like to make it clear which value corresponds to which device property.

tokyo4j:
Thanks. LGTM other than my minor suggestion above.

src/seat.c
jlindgren90: Good idea, applied.

jlindgren90:
Thanks for the review :) I applied your suggestion and will merge shortly.

PR #3029, opened 2025-08-27
tree-wide: use enum types/constants where appropriate

jlindgren90:
- add `LAB_WINDOW_TYPE_INVALID` in place of literal `-1`
- document more clearly that `enum lab_view_criteria` is a bitset
- other one-off replacements of integer values/types for consistency

Note: variables of type `enum lab_view_criteria` are already used extensively throughout the code to contain combinations of the declared enum values. I am not introducing any new usage here, just changing the single `uint32_t` to be consistent with all the other usages.

### DIFF
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -173,7 +173,7 @@ struct rcxml {
 		bool show;
 		bool preview;
 		bool outlines;
-		uint32_t criteria;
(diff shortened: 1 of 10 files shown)


tokyo4j:
Thanks. I'm not sure if we should keep storing bitset in `lab_view_criteria`, but as you said, we can postpone that discussion as it's already there.

PR #3031, opened 2025-08-29
Drop cosmic-workspace and use wlroots impl of ext-workspace-v1

tokyo4j:
Rebased #2930 on chase/0.20 branch for testing.


Consolatis:
This branch needs rebasing (e.g. all commits but the last one can be omitted).

Consolatis:
Thanks for your work on this PR and the wlroots implementation!

PR #3032, opened 2025-08-30
clients/: add lab-exit to display an exit window with labnag.

### LINKED ISSUE #917: Logout/Restart/Shutdown popup-menu
I have the experience that I misclicked on the Exit button twice.
Now I have installed nwg-bar and replaced the default exit menu item with nwg-bar.
I think new users are turned off by the fact, that misclicking is so easy.
If a user is unlucky important work could be deleted.
Would it be a hassle to implement a Logout/Restart/Shutdown popup-menu?
Or using a existing open-source popup-menu?

01micko:
closes #917
Summary follows:
- src/menu/menu.c: add `leave` action to call lab-exit.
- clients/meson.build: add install routine for lab-exit and lab-sensible-terminal.
- meson.build remove lab-sensible-terminal routine re the comment.
- po/labwc.pot: rebuild with new xgettext commands.
- po/it.po: add translations for lab-exit and edit my email.
- CONTRIBUTING.md#translations: modify the instructions for coders as two xgettext commands are needed since we now have two GUI facing languages that have translatable strings.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -450,27 +450,38 @@ translation strings under each English string.
 ## Coders
 
 Code contributors may need to update relevant files if their additions
-affect UI elements (at the moment only `src/menu/menu.c` and
-`src/config/rcxml.c`). In this case the `po/labwc.pot` file needs to be
+affect UI elements (at the moment only `src/menu/menu.c`, `src/config/rcxml.c`
+and `clients/lab-exit`). In this case the `po/labwc.pot` file needs to be
 updated so that translators can update their translations. Remember,
 many translators are _not_ coders!
 
 The process is fairly trivial however does involve some manual steps.
 
-1. After adding and testing your code additions to satisfaction, backup
-`po/labwc.pot`. You need the custom header from that file for the newly
-generated .pot file in the next step.
+After adding and testing your code additions to satisfaction, there are
+two commands needed to generate a sane `po/labwc.pot` file. Two commands are
+necessary because we now are using `C` and `Shell` in GUI facing files.
 
-2. From the root of the repository run this:
+1. From the root of the repository run this:
 
 ```
-xgettext --keyword=_ --language=C --add-comments -o po/labwc.pot src/menu/menu.c src/config/rcxml.c
+xgettext --keyword=_ --language=C --add-comments -o po/labwc.pot src/menu/menu.c\
+  src/config/rcxml.c --package-name=labwc --msgid-bugs-address=https://github.com/labwc/labwc/issues\
+  --copyright-holder="2025 labwc team"
 ```
 
 This generates a new pot file at `po/labwc.pot`
 
-3. Copy the header from the original `labwc.pot` to the new one, keeping
-the newly generated dates, check for sanity and commit.
+2. Then next run this:
+```
+xgettext -j --keyword=_ --language=Shell --add-comments -o po/labwc.pot  clients/lab-exit\
+  --package-name=labwc --msgid-bugs-address=https://github.com/labwc/labwc/issues\
+  --copyright-holder="2025 labwc team"
+```
+
+This appends the shell script translatable strings to the `C` code strings
+generated with the first command via the `-j` option to `xgettext`
+
+View the `xgettext` manual for explanation of the other options. 
 
 # Upversion
 
--- a/clients/lab-exit
+++ b/clients/lab-exit
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+# logout with labnag
+
+export TEXTDOMAIN=labwc
+export OUTPUT_CHARSET=UTF-8
+alias _='gettext'
+
+QUESTION=$(_ 'Are you sure you want to quit Labwc?')
+YES=$(_ 'Yes')
+NO=$(_ 'No')
+
+labnag \
+ -f "mono 10"\
+ -m "$QUESTION"\
+ -Z " $YES " "labwc -e"\
+ -Z " $NO "\
+ --background 00ffff\
+ --button-background 00ffff\
+ --border 00ccccaa\
+ --text 000000\
+ --button-text 000000\
+ --button-gap 8\
+ --button-margin-right 0\
+ --button-padding 5\
+ --button-border-size 2\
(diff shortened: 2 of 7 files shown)


01micko:
The `it.po` translation is mainly for demonstration/testing purposes so may need attention @stefonarch

tokyo4j:
I don't think we need another `lab-exit` as we can just embed `labnag` command in `<action><command>`.

If you don't like the style, then I suggest opening another PR to change the default style of `labnag`. I think that's better for styling consistency. CC @johanmalm

01micko:
> I don't think we need another `lab-exit` as we can just embed `labnag` command in `<action><command>`.

Yes but if shipped it's it's not translated. 

> If you don't like the style, then I suggest opening another PR to change the default style of `labnag`. I think that's better for styling consistency. CC @johanmalm

I'm not concerned about the defaults styles, they're fine. I'll happily remove mine. I was only trying to match the default `autostart` swaybg colors, even though it's not called in the default `menu.xml`; something else to add? I could even make it user configurable if users want to keep it with a simple `~/.config/labwc/lab-exit.conf`

tokyo4j:
> > I don't think we need another `lab-exit` as we can just embed `labnag` command in `<action><command>`.
> 
> Yes but if shipped it's it's not translated.

Well, I don't get what you mean. Does shipping `lab-exit` as another executable make any difference? For builtin menus created in `menu.c`, we can translate arguments passed to `labnag` like this:

```diff
diff --git a/src/menu/menu.c b/src/menu/menu.c
index 4e483989..7db9d3a4 100644
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -949,7 +949,12 @@ init_rootmenu(struct server *server)
 		item = item_create(menu, _("Reconfigure"), NULL, false);
 		item_add_action(item, "Reconfigure");
 		item = item_create(menu, _("Exit"), NULL, false);
-		item_add_action(item, "Exit");
+		action = item_add_action(item, "Execute");
+		struct buf buf = BUF_INIT;
+		buf_add_fmt(&buf, "labnag -m %s -Z %s 'labwc -e' -Z %s",
+			_("Are you sure to exit Labwc?"), _("Yes"), _("No"));
+		action_arg_add_str(action, "command", buf.data);
+		buf_reset(&buf);
 	}
 }
 
```

As another topic, I'm afraid that replacing `Exit` action in the builtin menu with `labnag` command removes the way to exit labwc when the user just git-clones, builds and launches labwc to try it out. We already replaced `alacritty` with `lab-sensible-terminal`, but I think having a reliable way to exit labwc is more important than that.

So I think we can:
- Create `If` action with `<prompt>` for the builting menu rather than directly invoking `labnag`
- Always execute `<then>` branch if `labnag` fails to launch
 
> > If you don't like the style, then I suggest opening another PR to change the default style of `labnag`. I think that's better for styling consistency. CC @johanmalm
> 
> I'm not concerned about the defaults styles, they're fine. I'll happily remove mine. I was only trying to match the default `autostart` swaybg colors, even though it's not called in the default `menu.xml`; something else to add? I could even make it user configurable if users want to keep it with a simple `~/.config/labwc/lab-exit.conf`

I think `labnag` should be invoked without any styling arguments from labwc since labwc itself is a standalone compositor that doesn't assume any specific utility apps.

01micko:
Granted, I see what you mean. :smile:  I'll mark this as a draft and await @johanmalm 's input.

johanmalm:
@01micko This is a really nice work, but I think it crosses the lines into Desktop Environment space or just plain personal setup.

> So I think we can:
>
>    Create If action with <prompt> for the builting menu rather than directly invoking labnag
>    Always execute <then> branch if labnag fails to launch

I'm happy to debate this, but as a starting point I feel that the safer option is to *not* fallback on the `<then>` branch in the absence of `labnag` for anything that the user/maintainer thought warranted a `<prompt>` in the first place.

> As another topic, I'm afraid that replacing Exit action in the builtin menu with labnag command removes the way to exit labwc when the user just git-clones, builds and launches labwc to try it out. We already replaced alacritty with lab-sensible-terminal, but I think having a reliable way to exit labwc is more important than that.

Yes, that's exactly my thinking.

I did originally ponder changing `item_add_action(item, "Exit")` to a `labnag` command or to something with `<prompt>` but decided not to for the reason you just described. I think that really becomes an exercise of users or Desktop Environment devs.

On a note of interest, I did find in the early days of development that Openbox actually relies on various other things being installed (like a theme!). I have tried to maintain the ability to just use the compositor from a build directory. So far, the only "reduced capabilities" include translations and the use of `clients/*`. I think the more we stick to that line, the better.

@01micko I've read the gettext construct with interest. I wonder if it's something that we ought to consider in BunsenLabs. I have not publicised it, but have started playing with a building a thin layer on top of `labwc`. (https://teallach-desktop.github.io) I'm mostly doing to try to gain a perspective on what it's like to **use** the compositor. Might a lab-exit script with translation belong in such a domain?

@01micko Another thought... if we proceed on the basis that `lab-exit` is out-of-scope, would it be worth adding it (or possibly `nagoff` to the `labnag(1)` man page as an example. We currently do not have any example there, and I'm increasingly thinking that example bring man pages to life. https://labwc.github.io/labnag.1.html

For general context, we have debated anything going into `clients/` quite extensively and on the whole I think it's fair to say that we're all aligned on wanting to keep it very small and modest. `labnag` itself was not an obvious addition as tools like `zenity` can be used for simple yes/no dialogs. The reasons for including a `labnag` was to enabled `<prompt>` without any external dependencies, and to lay the foundation for stderr logging (or possibly just theme/config errors).

01micko:
Thanks @johanmalm .

It makes sense to keep `clients/` small and `lab-exit` was a bit of an exercise which was fun. Of course menus can easily be changed in `menu.xml` and anyone can use `nagoff` or `lab-exit` and change it to suit. An example in `labnag.1` is a good idea. I think `nagoff` is a little more comprehensive, but should be made more `systemd` compliant. I can get the values from `bunsen-exit` :wink: 

I've read with interest just now about `teallach-desktop` and it looks very interesting. I'm sure you'll let us know when it's read.

PR #2981, opened 2025-08-08
osd: support window switcher with thumbnails

tokyo4j:
This PR implements window switcher with thumbnails (window previews):

https://github.com/user-attachments/assets/7a53934a-a1a2-4db9-bf58-3bd8f7421328

New configration `<windowSwitcher style="classic|thumbnail">` is added to switch between the old/new window switcher style.

Following theme entries are added:
```
osd.window-switcher.thumbnail.width.max: 80%
osd.window-switcher.thumbnail.padding: 4
osd.window-switcher.thumbnail.item.width: 300
osd.window-switcher.thumbnail.item.height: 250
osd.window-switcher.thumbnail.item.padding: 10
osd.window-switcher.thumbnail.item.active.border.width: 2
osd.window-switcher.thumbnail.item.active.border.color: #589bda
osd.window-switcher.thumbnail.item.active.bg.color: #c7e2fc
osd.window-switcher.thumbnail.item.icon.size: 60
```

Some theme entries for the old style are renamed with `osd.window-switcher.classic` prefix:
```
osd.window-switcher.classic.width: 600
osd.window-switcher.classic.padding: 4
osd.window-switcher.classic.item.padding.x: 10
osd.window-switcher.classic.item.padding.y: 1
osd.window-switcher.classic.item.active.border.width: 2
osd.window-switcher.classic.item.icon.size: 50
```

Following theme entries are unchanged and used by both old and new styles:
```
osd.bg.color: #e1dedb
osd.border.color: #000000
osd.border.width: 1
osd.label.text.color
```

TODO:
- More testing, especially with outputs and clients with transformation or scaling
- Fill documentations
- ~Find a way to get rid of `scene_buffer->WLR_PRIVATE.texture`, if possible? I'll need to learn APIs of wlroots more...~ I think we can just call `wlr_texture_from_buffer()`. It may incur some overhead compared to the scene renderer of wlroot as it always creates a new texture, but it will be negligible.

I'll appreciate testing or suggestions for the design.

Link: https://github.com/labwc/labwc/discussions/2172

PR #3033, opened 2025-08-30
scaled-icon-buffer: fix large client-side icon not being loaded

tokyo4j:
Fixes https://github.com/labwc/labwc/pull/2981#issuecomment-3238675460.

This commit fixes that client-side icons were not loaded when the rendered icon size is larger than icon sizes from the client. This bug has become more likely to happen due to the new thumnail-style window switcher.

The cause was `abs(INT_MIN)` becomes `INT_MIN` due to integer overflow.


danielfikko:
This fixes the missing icon in the OSD. Thank you very much!

johanmalm:
Good fine! Thanks.

PR #3034, opened 2025-08-30
docs/labnag.1.scd: add a labnag example

01micko:
Just the simple `nagoff` example modified for systemd.


01micko:
I left the word "Example" in like that and not capitalized. I think it's appropriate but if opinions differ I'm happy to make it upper case.

johanmalm:
For consistency I prefer the headings to be capitalized.

Other than that, LGTM

01micko:
Happy to change.

PR #3035, opened 2025-08-30
meson.build: move lab-sensible-terminal install routine.

01micko:
Moved to clients/meson.build as per the comment in meson.build.


01micko:
Just some cleanup before it gets forgotten :)

johanmalm:
Awesome. Thanks.

PR #3038, opened 2025-08-31
overlay: refactor

tokyo4j:
- Use a single `lab_scene_rect` for both overlay background and outlines, like I described in the TODO comment in #2809.
- Simplify the resource management by destroying the overlay tree when it's hidden. I think its overhead is pretty minimal.
- Share a single `lab_scene_rect` for both region/edge overlays.

### DIFF
--- a/include/overlay.h
+++ b/include/overlay.h
@@ -2,10 +2,9 @@
 #ifndef LABWC_OVERLAY_H
 #define LABWC_OVERLAY_H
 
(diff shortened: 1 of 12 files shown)


johanmalm:
This is very nice refactoring. Thanks for the work to keep code clean and simple.

LGTM.

johanmalm:
PS. Have tested.

PR #3041, opened 2025-09-01
view: use fixed default window width 

tokyo4j:
Now it's not very reasonable to determine the default window width based on the titlebar geometry, as the titlebar can be shrunk to 1px.

Let's use the fixed value of 100px for simplification.


johanmalm:
Makes sense. Thanks.

PR #3042, opened 2025-09-01
view: restore initially-maximized window position after unplug/plug

tokyo4j:
`update_last_layout_geometry()` stores `view->natural_geometry` in `view->last_layout_geometry`, but it's empty for initially-maximized windows, so their positions were not restored after outputs are unplugged and plugged (also when VT switching in wlroots 0.19.0).

This PR sets the fallback natural geometry (at the center of the output) so that initially-maximized windows reappears in the same output.


johanmalm:
Thanks.

PR #3043, opened 2025-09-01
interactive: un-maximize only axes that are being resized

jlindgren90:
When resizing in only one axis (horizontal/vertical), there's no reason to un-maximize the other axis.

Most of the supporting logic was landed already in:
- #3020

However, that PR fixed only the client-initiated case. Let's do the same for interactive resize.

The first commit is prep work, eliminating `view_restore_to()` and exposing `view_set_maximized()` instead.
Then all that remains is to make a small change in `interactive_begin()` (second commit).


johanmalm:
Thanks. LGTM.

PR #3045, opened 2025-09-01
default-bindings.h: focus & raise on border press

johanmalm:
...because it is probably what most people expect and it makes the behavior consistent with that of Openbox.

Fixes: #3039

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -376,6 +376,10 @@
     </context>
 
     <context name="Border">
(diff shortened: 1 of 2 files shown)


tokyo4j:
I think we should also update corner contexts (e.g. `TLCorner`).

Can we see #3047 first? It will reduce the diff in this PR.

johanmalm:
Rebased following merge of 3047.

Ready for final review.

tokyo4j:
Thanks!

PR #3046, opened 2025-09-01
NEWS.md: link from TOC to each section

johanmalm:


tokyo4j:
LGTM. I agree that linking to each section makes more sense. Thanks for adding PR links, too.

PR #3047, opened 2025-09-01
mousebind: support `Border` context

tokyo4j:
This new context includes `Top`...`BRCorner` and makes it easier to modify the mousebinds bound to them.


johanmalm:
Nice!

PR #3048, opened 2025-09-02
scaled-scene-buffer: restructure source files

tokyo4j:
- Rename `scaled_scene_buffer` to `scaled_buffer`. This makes it clear that `scaled_{font,img,icon}_buffers` are implementations of it.
- Move the files from `src/common` to `src/scaled-buffer` as `scaled_icon_buffer` heavily depends on `server` and `view` etc.


johanmalm:
LGTM

PR #3050, opened 2025-09-03
Unify node + ssd types/descriptors

jlindgren90:
First of all, as noted in https://github.com/labwc/labwc/pull/2999#issuecomment-3191581771, `enum ssd_part_type` contains several node types that are not actually part of server-side decorations (ROOT, MENU, OSD, etc.) So rename it to `lab_node_type` and move it to a common location `common/node-type.c/h`, along with some related conversion/comparison functions.

Second, eliminate `struct ssd_hover_state` (addressing a FIXME) and just store a pointer to the button instead. The view pointer can be fetched indirectly from that.

Finally, I noticed that `struct ssd_part` and `struct node_descriptor` seem to have essentially the same purpose: tag a `wlr_scene_node` with some extra data indicating what we're using it for. And as with `enum ssd_part_type` (now `lab_node_type`), `ssd_part` is used for several types of nodes that are not part of SSD.

So instead of the current chaining (`node_descriptor` -> `ssd_part`), let's flatten/unify the two structs.

In detail:

- First, merge `enum node_descriptor_type` into `enum lab_node_type`.
- Add a separate view pointer in `node_descriptor`, since in the case of SSD buttons we need separate view and button data pointers.
- Rename `ssd_part_button` to simply `ssd_button`. It no longer contains an `ssd_part` as base.
- Add `node_try_ssd_button_from_node()` which replaces `node_ssd_part_from_node()` + `button_try_from_ssd_part()`.
- Factor out `ssd_button_free()` to be called in node descriptor destroy.
- Finish up by slightly reorganizing `get_cursor_context()` to handle the unified structs.

Overall, this simplifies the code a bit, and in my opinion makes it easier to understand. No functional change intended.

This builds heavily on:
- #2993 

So if @tokyo4j has time to review this, it would be appreciated.

### DIFF
--- a/include/common/node-type.h
+++ b/include/common/node-type.h
@@ -43,12 +43,23 @@ enum lab_node_type {
 	LAB_NODE_CLIENT,
 	LAB_NODE_FRAME,
 	LAB_NODE_ROOT,
-	LAB_NODE_MENU,
+	LAB_NODE_MENUITEM,
 	LAB_NODE_OSD,
 	LAB_NODE_LAYER_SURFACE,
 	LAB_NODE_LAYER_SUBSURFACE,
 	LAB_NODE_UNMANAGED,
 	LAB_NODE_ALL,
+
+	/* translated to LAB_NODE_CLIENT by get_cursor_context() */
+	LAB_NODE_VIEW,
+	LAB_NODE_XDG_POPUP,
+	LAB_NODE_LAYER_POPUP,
+	LAB_NODE_SESSION_LOCK_SURFACE,
+	LAB_NODE_IME_POPUP,
+
+	/* never returned by get_cursor_context() */
+	LAB_NODE_TREE,
+	LAB_NODE_SCALED_BUFFER,
 };
 
 enum lab_node_type node_type_parse(const char *context);
--- a/include/common/node-type.h
+++ b/include/common/node-type.h
@@ -1,0 +1,60 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_NODE_TYPE_H
+#define LABWC_NODE_TYPE_H
+
+#include "common/edge.h"
+
+/*
+ * In labwc, "node type" indicates the role of a wlr_scene_node in the
+ * overall desktop. It also maps more-or-less to the openbox concept of
+ * "context" (as used when defining mouse bindings).
+ *
+ * Node types are defined in the order they should be processed for press
+ * and hover events. Note that some of their respective interactive areas
+ * overlap, so for example buttons need to come before title.
+ */
+enum lab_node_type {
+	LAB_NODE_NONE = 0,
+
+	LAB_NODE_BUTTON_CLOSE = 1,
+	LAB_NODE_BUTTON_MAXIMIZE,
+	LAB_NODE_BUTTON_ICONIFY,
+	LAB_NODE_BUTTON_WINDOW_ICON,
+	LAB_NODE_BUTTON_WINDOW_MENU,
+	LAB_NODE_BUTTON_SHADE,
+	LAB_NODE_BUTTON_OMNIPRESENT,
+	LAB_NODE_BUTTON_FIRST = LAB_NODE_BUTTON_CLOSE,
+	LAB_NODE_BUTTON_LAST = LAB_NODE_BUTTON_OMNIPRESENT,
+	LAB_NODE_BUTTON,
+
+	LAB_NODE_TITLEBAR,
+	LAB_NODE_TITLE,
+
+	LAB_NODE_CORNER_TOP_LEFT,
+	LAB_NODE_CORNER_TOP_RIGHT,
+	LAB_NODE_CORNER_BOTTOM_RIGHT,
+	LAB_NODE_CORNER_BOTTOM_LEFT,
+	LAB_NODE_BORDER_TOP,
+	LAB_NODE_BORDER_RIGHT,
+	LAB_NODE_BORDER_BOTTOM,
+	LAB_NODE_BORDER_LEFT,
+	LAB_NODE_BORDER,
+
+	LAB_NODE_CLIENT,
+	LAB_NODE_FRAME,
+	LAB_NODE_ROOT,
+	LAB_NODE_MENU,
+	LAB_NODE_OSD,
+	LAB_NODE_LAYER_SURFACE,
+	LAB_NODE_LAYER_SUBSURFACE,
+	LAB_NODE_UNMANAGED,
+	LAB_NODE_ALL,
+};
+
+enum lab_node_type node_type_parse(const char *context);
+
+bool node_type_contains(enum lab_node_type whole, enum lab_node_type part);
+
+enum lab_edge node_type_to_edges(enum lab_node_type type);
+
+#endif /* LABWC_NODE_TYPE_H */
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -225,7 +225,7 @@ struct server {
 	 */
 	struct view *active_view;
 
-	struct ssd_hover_state *ssd_hover_state;
+	struct ssd_part_button *hover_button;
 
 	/* Tree for all non-layer xdg/xwayland-shell surfaces */
 	struct wlr_scene_tree *view_tree;
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -225,7 +225,7 @@ struct server {
 	 */
 	struct view *active_view;
 
-	struct ssd_part_button *hover_button;
+	struct ssd_button *hover_button;
 
 	/* Tree for all non-layer xdg/xwayland-shell surfaces */
 	struct wlr_scene_tree *view_tree;
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -263,7 +263,7 @@ get_surface_from_layer_node(struct wlr_scene_node *node)
 struct cursor_context
 get_cursor_context(struct server *server)
 {
-	struct cursor_context ret = {.type = LAB_SSD_NONE};
+	struct cursor_context ret = {.type = LAB_NODE_NONE};
 	struct wlr_cursor *cursor = server->seat.cursor;
 
 	/* Prevent drag icons to be on top of the hitbox detection */
@@ -281,7 +281,7 @@ get_cursor_context(struct server *server)
 
 	ret.node = node;
 	if (!node) {
-		ret.type = LAB_SSD_ROOT;
+		ret.type = LAB_NODE_ROOT;
 		return ret;
 	}
 
@@ -289,7 +289,7 @@ get_cursor_context(struct server *server)
 	if (node->type == WLR_SCENE_NODE_BUFFER) {
 		struct wlr_surface *surface = lab_wlr_surface_from_node(node);
 		if (node->parent == server->unmanaged_tree) {
-			ret.type = LAB_SSD_UNMANAGED;
+			ret.type = LAB_NODE_UNMANAGED;
 			ret.surface = surface;
 			return ret;
 		}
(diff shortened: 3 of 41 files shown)


tokyo4j:
Amazing work! I really like your idea. I'll review this PR in this weekend.

src/desktop.c
```diff
@@ -299,59 +299,67 @@ get_cursor_context(struct server *server)
...
-			case LAB_NODE_DESC_TREE:
-			case LAB_NODE_DESC_SCALED_BUFFER:
+			case LAB_NODE_TREE:
+			case LAB_NODE_SCALED_BUFFER:
+				/* Continue to parent node */
 				break;
+			default:
+				/*
+				 * All other node descriptors (buttons, title,
+				 * frame, etc.) should have an associated view.
```
tokyo4j: I don't think there are "frame" nodes. Let's remove it.

src/node.c
```diff
@@ -3,13 +3,20 @@
...
 {
 	if (!node_descriptor) {
 		return;
 	}
+
+	if (node_descriptor->type >= LAB_NODE_BUTTON_FIRST
+			&& node_descriptor->type <= LAB_NODE_BUTTON_LAST) {
+		ssd_button_free(node_descriptor->data);
+	}
+
```
tokyo4j: - You can use `node_try_ssd_button_from_node()` instead.
- ~I prefer freeing `ssd_button` in a node destroy handler `ssd-button.c` so that we allocate and free `ssd_button` in the same file.~ I prefer freeing `ssd_button` in its own destroy handler like we did in `ssd_part` so that we can make its lifetime more self-contained, not requiring `ssd_button_free()` or downcast.

src/node.c
```diff
@@ -67,24 +73,29 @@ node_menuitem_from_node(struct wlr_scene_node *wlr_scene_node)
...
+struct ssd_button *
+node_try_ssd_button_from_node(struct wlr_scene_node *wlr_scene_node)
 {
 	assert(wlr_scene_node->data);
 	struct node_descriptor *node_descriptor = wlr_scene_node->data;
-	assert(node_descriptor->type == LAB_NODE_DESC_SCALED_BUFFER);
-	return (struct scaled_buffer *)node_descriptor->data;
+
+	if (node_descriptor->type >= LAB_NODE_BUTTON_FIRST
+			&& node_descriptor->type <= LAB_NODE_BUTTON_LAST) {
```
tokyo4j: You can use `node_type_contains(LAB_NODE_BUTTON, type)` instead. If you don't like it, you can remove `LAB_NODE_BUTTON` and replace `node_type_contains(LAB_NODE_BUTTON, type)` (called in `action.c` etc.) with arithmetic comparison, but I suggest keeping it for now.

include/labwc.h
```diff
@@ -225,7 +225,7 @@ struct server {
 	 */
 	struct view *active_view;
 
-	struct ssd_hover_state *ssd_hover_state;
+	struct ssd_button *hover_button;
```
tokyo4j: I suggest `hovered_button` instead. Then you can rename `ssd_update_button_hover()` to `ssd_update_hovered_button()`.

include/common/node-type.h
```diff
@@ -0,0 +1,71 @@
...
+
+/*
+ * In labwc, "node type" indicates the role of a wlr_scene_node in the
+ * overall desktop. It also maps more-or-less to the openbox concept of
+ * "context" (as used when defining mouse bindings).
+ *
+ * Node types are defined in the order they should be processed for press
+ * and hover events. Note that some of their respective interactive areas
+ * overlap, so for example buttons need to come before title.
+ */
```
tokyo4j: I guess the second paragraph is obsolete today. That was written 4 years ago when `ssd_at()` (which is removed long ago) iterated over the `ssd_part_type` entries to get the mouse context:

https://github.com/labwc/labwc/blob/1dc3059707245e4345ae095ff9e215a598b3ba8c/src/ssd.c#L185-L196

Today, as far as I understand, the order of enum entries matters only in `node_type_contains()`. So we can instead say `The order must be in sync with node_type_contains()`.

src/node.c
jlindgren90: I don't really like having two separate `destroy` handlers, because then it's unclear which runs first, and both objects (`node_descriptor` and `ssd_button`) need to be careful because the other may have already been destroyed. Having an explicit, linear ownership chain (`wlr_scene_node` -> `node_descriptor` -> `ssd_button`) is clearer to me.

I do agree with using `node_try_ssd_button_from_node()` here, that's cleaner.

src/desktop.c
jlindgren90: Right, "frame" is a category rather than a specific node. I'll say "borders" here instead.

include/common/node-type.h
jlindgren90: Good catch, I'll update the comment.

src/node.c
jlindgren90: Actually, `node_try_ssd_button_from_node()` doesn't work because we don't have a pointer to the `wlr_scene_node`. But we can used `node_type_contains()` at least.

include/labwc.h
jlindgren90: I'm also going to fix spelling of `LAB_BS_HOVERD`, it's getting on my nerves ;)

jlindgren90:
Fix-ups added to address the comments. I'll rebase and squash these before merge.

src/node.c
tokyo4j: OK, I see your point. Let's keep `ssd_button_free()` and go with your approach then.

tokyo4j:
Please update comments below. `/* List of ssd_parts that represent buttons */` can be replaced with `/* struct ssd_button.link */` following the member declaration, like we usually do for `struct wl_link` members.

```
hrak@ay0x ~/c/labwc (node-type)> grep -Rn "ssd_part" --exclude-dir .git --exclude-dir build
include/ssd-internal.h:17: * ssd_part attached to the node.
include/ssd-internal.h:106:                     /* List of ssd_parts that represent buttons */
include/ssd-internal.h:152:      * When the type of ssd_part pointing to ssd_button is
hrak@ay0x ~/c/labwc (node-type)> grep -Rn "LAB_SSD_" --exclude-dir .git --exclude-dir build | grep -v "LAB_SSD_MODE" | gre
p -v "LAB_SSD_PREF_"
include/theme.h:107:             * buttons[LAB_SSD_BUTTON_ICONIFY][LAB_BS_HOVERD | LAB_BS_TOGGLED]
include/theme.h:109:             * Elements in buttons[0] are all NULL since LAB_SSD_BUTTON_FIRST is 1.
include/ssd-internal.h:19: * ssd->tree (LAB_SSD_NONE)
include/ssd-internal.h:20: * +--titlebar (LAB_SSD_PART_TITLEBAR)
include/ssd-internal.h:25: * |  |  +--title (LAB_SSD_PART_TITLE)
include/ssd-internal.h:26: * |  |  +--iconify button (LAB_SSD_BUTTON_ICONIFY)
include/ssd-internal.h:30: * |  |  +--window icon (LAB_SSD_BUTTON_WINDOW_ICON)
include/ssd-internal.h:153:      * LAB_SSD_BUTTON_WINDOW_ICON, these are all NULL and window_icon is
include/input/cursor.h:44: *    set to LAB_SSD_LAYER_SURFACE and view will be NULL.
include/input/cursor.h:48: *    be set to LAB_SSD_UNMANAGED and view will be NULL.
include/input/cursor.h:55: *    LAB_SSD_CLIENT and return the attached view.
include/input/cursor.h:62: * type will be set to LAB_SSD_ROOT and view will be NULL.
```

src/ssd/ssd.c
```diff
@@ -169,7 +146,8 @@ ssd_create(struct view *view, bool active)
 
 	ssd->view = view;
 	ssd->tree = wlr_scene_tree_create(view->scene_tree);
-	attach_ssd_part(LAB_SSD_NONE, view, &ssd->tree->node);
+	node_descriptor_create(&ssd->tree->node,
+		LAB_NODE_NONE, view, /*data*/ NULL);
```
tokyo4j, line 150: Sorry for missing this in the previous round of review, but binding `LAB_NODE_NONE` to `ssd->tree` in `ssd_create()` looks strange.

This is a bit tricky to explain, but my idea for binding `LAB_SSD_NONE` to `ssd->tree` in #2993 was that, when cursor is on the border or extents, `get_cursor_context()` catches the root node of the ssd and `ssd_get_resizing_type()` returns "resizing" part types (e.g. `LAB_SSD_PART_TOP`) by arithmetically comparing cursor position and ssd geometry. We don't just bind border/extent nodes to the "resizing" types because there are also borders drawn to `theme.windows[].corner_top_{left,right}_normal` buffers. Ref: #2552.

So I suggest adding `LAB_NODE_SSD_ROOT` node type and bind it to `ssd->tree`.

src/desktop.c
tokyo4j: This is a bit tricky, but we don't actually have node descriptors for borders. As explained in https://github.com/labwc/labwc/pull/3050/files#r2328158193, the detection of border/extents is instead done by `ssd_get_resizing_type()`. So I suggest just mentioning button and title (or titlebar too).

tokyo4j:
Could you apply this patch at the first of your branch? I found a mistake in #2993 that confuses my explanation above.

```patch
From 5237b2d0ea358cf23b59cd909bf4f77aac9fe7f4 Mon Sep 17 00:00:00 2001
From: tokyo4j <hrak1529@gmail.com>
Date: Sun, 7 Sep 2025 03:42:50 +0900
Subject: [PATCH] ssd-titlebar: don't bind ssd_part to ssd->tree

Fixes up f347a81. This mistake didn't cause misbehaviors because the
ssd_part bound to ssd->tree is just a placeholder to let
get_cursor_context() call ssd_get_resizing_type() when the cursor is on
border/extents.
---
 src/ssd/ssd-titlebar.c | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

index 13e3ebf6..e2cb2e06 100644
--- a/src/ssd/ssd-titlebar.c
+++ b/src/ssd/ssd-titlebar.c
@@ -33,7 +33,7 @@ ssd_titlebar_create(struct ssd *ssd)
 	int corner_width = ssd_get_corner_width();
 
 	ssd->titlebar.tree = wlr_scene_tree_create(ssd->tree);
-	attach_ssd_part(LAB_SSD_PART_TITLEBAR, view, &ssd->tree->node);
+	attach_ssd_part(LAB_SSD_PART_TITLEBAR, view, &ssd->titlebar.tree->node);
 
 	enum ssd_active_state active;
 	FOR_EACH_ACTIVE_STATE(active) {
-- 
2.51.0

```

src/ssd/ssd.c
jlindgren90: Can that be done in a follow-up PR? I'm reasonably confident that the refactoring done so far doesn't cause any behavioral change, but if we start adding new node types then I'm not so sure. I'd rather keep that separate in case we need to bisect later.

tokyo4j: Sure, I can open a follow-up PR later.

jlindgren90:
> Could you apply this patch at the first of your branch? I found a mistake in #2993 that confuses my explanation above.

I'll add this since you have it all set to go. Can we put any more logical fixes in a follow-up PR though, separate from the refactoring in this one?

tokyo4j:
> I'll add this since you have it all set to go. Can we put any more logical fixes in a follow-up PR though, separate from the refactoring in this one?

Sure, I can do it in the follow-up PR.

LGTM. Feel free to squash the [fixup] comments and merge this PR. Thanks for another cleanup!

src/desktop.c
jlindgren90: Hmm okay, makes sense

tokyo4j: Sorry I removed my comment as I thought you missed it and it's trivial. For our records, I suggested not mentioning "border" as there's no ssd_part (or node_descriptor) bound to a scene-node and cursor hovering on border/extents are arithmetically detected by `ssd_get_resizing_type()`.

jlindgren90:
Rebased, added @tokyo4j 's ssd-titlebar patch, updated one comment in get_cursor_context(). Will merge shortly. Thanks for the thorough review!

src/desktop.c
jlindgren90: I was sneaky and slipped in a comment update in the final rebase :)

PR #3055, opened 2025-09-04
view.c: remove cruft in view_apply_maximized_geometry()

tokyo4j:
I don't know the purpose of the removed lines, but they are from 4 years ago when labwc didn't use scene-graph API and I'm sure they are redundant.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -1302,14 +1302,6 @@ view_apply_maximized_geometry(struct view *view)
 	assert(output_is_usable(output));
 
 	struct wlr_box box = output_usable_area_in_layout_coords(output);
-	if (box.height == output->wlr_output->height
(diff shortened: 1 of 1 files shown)


johanmalm:
@jlindgren90 - Fancy reviewing this one?

tokyo4j:
Additional note: the removed lines are from ed3d0cd523efcc3e91cc6d5889c607fe4421ba16 (HiDPI: fix osd placement and maximized scale). But comparing the usable area size and the physical screen size doesn't make sense to me.

jlindgren90:
I have puzzled over this exact code before and it looked wrong to me too. The box returned by `output_usable_area_in_layout_coords()` should already be in device-independent pixels so dividing by the output scale doesn't make sense.

Let's remove it. LGTM.

PR #3061, opened 2025-09-06
Trivial cleanups

jlindgren90:
A couple of trivial cleanups:

- Remove an unused parameter & struct member
- Make use of a function that exists already, rather than repeating it


tokyo4j:
Thanks.

PR #3062, opened 2025-09-06
Header cleanups

jlindgren90:
- Remove unused #includes
- Make use of forward declarations to reduce indirect #includes (one
  header including another) where it seems appropriate

Really this is just cosmetic as the extra #includes do no harm, but I had these cleanups in a local branch for a while, so submitting them in case they're of interest.


tokyo4j:
Thanks!

PR #3063, opened 2025-09-07
node-type: some cleanups

tokyo4j:
This PR shouldn't change any behaviors.

The 1st commit removes `LAB_NODE_SCALED_BUFFER` node type and `scaled_{img,icon}_buffer_from_node()`. I remember those functions were required for updating window icons some time ago, but they are not used anymore.

The 2nd commit removes `LAB_NODE_TREE` node type. It was attached to `output->layer_popup_tree` etc., but I don't think they are needed. CC @johanmalm

The 3nd commit attaches `LAB_NODE_SSD_ROOT` to `ssd->tree` instead of `LAB_NODE_NONE`, as I described in https://github.com/labwc/labwc/pull/3050#discussion_r2328158193. I also updated `get_cursor_context()` to make my intent clearer.

### DIFF
--- a/include/common/node-type.h
+++ b/include/common/node-type.h
@@ -62,7 +62,6 @@ enum lab_node_type {
 
 	/* never returned by get_cursor_context() */
 	LAB_NODE_TREE,
-	LAB_NODE_SCALED_BUFFER,
 };
 
 enum lab_node_type node_type_parse(const char *context);
--- a/include/common/node-type.h
+++ b/include/common/node-type.h
@@ -59,9 +59,6 @@ enum lab_node_type {
 	LAB_NODE_LAYER_POPUP,
 	LAB_NODE_SESSION_LOCK_SURFACE,
 	LAB_NODE_IME_POPUP,
-
-	/* never returned by get_cursor_context() */
-	LAB_NODE_TREE,
 };
 
 enum lab_node_type node_type_parse(const char *context);
--- a/include/common/node-type.h
+++ b/include/common/node-type.h
@@ -59,6 +59,12 @@ enum lab_node_type {
 	LAB_NODE_LAYER_POPUP,
 	LAB_NODE_SESSION_LOCK_SURFACE,
 	LAB_NODE_IME_POPUP,
+
+	/*
+	 * translated to LAB_CORNER_* or LAB_BORDER* by
+	 * ssd_get_resizing_type()
+	 */
+	LAB_NODE_SSD_ROOT,
 };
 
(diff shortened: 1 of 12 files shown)


johanmalm:
I love this tidying up. It's really great.

I think I'm fine with all the changes, but wonder if @jlindgren90 want to do a quick reviewing following the patches from the last few days. If they don't have time, I'll do another review in a couple of days.

Regarding the LAB_NODE_TREE, I'm fine with removing that. I actually can't remember the reason for first introducing it :smile:

I think the LAB_NODE_SSD_ROOT is a good idea. It's more readable than NONE.

So, am I right in thinking that it's only those layer-shell subsurfaces which do not fit the node-descriptor pattern.

include/common/node-type.h
```diff
@@ -60,9 +60,11 @@ enum lab_node_type {
 	LAB_NODE_SESSION_LOCK_SURFACE,
 	LAB_NODE_IME_POPUP,
 
-	/* never returned by get_cursor_context() */
-	LAB_NODE_TREE,
-	LAB_NODE_SCALED_BUFFER,
+	/*
+	 * translated to LAB_CORNER_* or LAB_CORNER_BUTTON_* by
```
jlindgren90: What is LAB_CORNER_BUTTON_*?

src/desktop.c
```diff
@@ -332,35 +332,46 @@ get_cursor_context(struct server *server)
...
-
+			case LAB_NODE_BUTTON_FIRST...LAB_NODE_BUTTON_LAST:
+			case LAB_NODE_SSD_ROOT:
+			case LAB_NODE_TITLE:
+			case LAB_NODE_TITLEBAR:
 				ret.node = node;
 				ret.view = desc->view;
+				/*
+				 * A node_descriptor attached to a ssd part
+				 * must have and associated view.
```
jlindgren90: Typo: and -> an

jlindgren90 (approved):
Looks fine overall to me, a couple of minor comments. I don't know the background of the node types being removed but if they are unused then let's remove them. 👍

include/common/node-type.h
tokyo4j: Oops, I don't know what I was thinking. Fixed.

tokyo4j:
> So, am I right in thinking that it's only those layer-shell subsurfaces which do not fit the node-descriptor pattern.

I think you're right, but I doubt we really need `LAB_NODE_LAYER_SUBSURFACE` either. I guess we can make it cleaner.

PR #3064, opened 2025-09-07
desktop.c: don't print errors when cursor is on resize-indicator

tokyo4j:
Removes log spams when the cursor is on resize-indicator:

https://github.com/user-attachments/assets/27fdb17a-0fee-4966-9d4b-8604d79b3356

In #2993, I thought there are only window contents and SSD under `view->scene_tree` and forgot about the resize-indicator.

I also refactored the logic around it:
- Remove `ret.node->type == WLR_SCENE_NODE_BUFFER` check since it's already done by `lab_wlr_surface_from_node()`
- Eliminate duplicated call to `lab_wlr_surface_from_node()`

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -303,13 +303,12 @@ get_cursor_context(struct server *server)
 			case LAB_NODE_VIEW:
 			case LAB_NODE_XDG_POPUP:
 				ret.view = desc->view;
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM. Nice spot that the WLR_SCENE_NODE_BUFFER check is in `lab_wlr_surface_from_node()`

PR #3065, opened 2025-09-07
src/output.c: restore drm lease include

Consolatis:
Introduced in #3062.

Its usage is guarded by a wlroots version check
which prevented the CI to detect the issue.

When compiled with a wlroots version > 0.19.0
(like the chase 0.20 PR) the error shows up.


tokyo4j:
Oops, nice catch.

jlindgren90:
Oops, sorry about that

PR #3066, opened 2025-09-08
desktop: remove LAB_NODE_LAYER_SUBSURFACE node type

tokyo4j:
- Remove `LAB_NODE_LAYER_SUBSURFACE` to compact our codebase
- Clarify the semantics of `cursor_context` returned by `get_cursor_context()` as I described in `cursor.h`; when cursor is on a surface (e.g. xdg/x11/layer/session-lock), the returned `ctx.surface` and `ctx.node` points to the (sub)surface rather than its parent.

I believe this improves the quality of our codebase, but actually I'm not 100% sure this doesn't cause regressions. #667, #1030 and #1572 seem to be related, but I couldn't follow them. As for #1030, I confirmed I can still launch applications by clicking menu items of `xfce4-panel`.


src/input/cursor.c
```diff
@@ -1127,7 +1137,7 @@ cursor_process_button_press(struct seat *seat, uint32_t button, uint32_t time_ms
 #endif
 	}
 
-	if (ctx.type != LAB_NODE_CLIENT && ctx.type != LAB_NODE_LAYER_SUBSURFACE
+	if (ctx.type != LAB_NODE_CLIENT && ctx.type != LAB_NODE_LAYER_SURFACE
```
tokyo4j, line 1140: Question: was it intentional that we hadn't had `ctx.type != LAB_NODE_LAYER_SURFACE` along with `ctx.type != LAB_NODE_LAYER_SUBSURFACE`? If the intention was "close popup whenever the cursor presses non-surface areas", I think we should've excluded layer-surfaces, too. Related: #1030, #1031

Consolatis:
AFAIR the main issue was that we need to give keyboard focus to layer subsurfaces, it was something that could be reproduced with the margin settings in `gtk-layer-demo` from what I remember but I am really not sure about the details here.

CC @johanmalm

tokyo4j:
> AFAIR the main issue was that we need to give keyboard focus to layer subsurfaces, it was something that could be reproduced with the margin settings in `gtk-layer-demo` from what I remember but I am really not sure about the details here.

From the commit message in a5fcbfa, I guess the point is that we must give keyboard focus to the layer surface when its subsurface is pressed while the layer surface has on-demand keyboard interactivity.

I reproduced that, like described in #1572, clicking application menu in `lxqt-panel` didn't give it keyboard focus before #1594, and #1594 fixed it. I confirmed this PR doesn't change this behavior.

#697 mentions the margin settings dialog in `gtk-layer-demo`. Before that PR, labwc didn't send enter events or motion events to layer subsurfaces, so clicking on the margin settings dialog just closed it.

But yeah, let's wait for @johanmalm's response anyway.

tokyo4j:
I split this PR into 3 commits:
- 1st commit removes `LAB_NODE_LAYER_SUBSURFACE`. This should not change any behaviors.
- 2nd commit lets `get_cursor_context()` return layer-pop subsurface rather than its parent. ~This will send wl_pointer.enter/motion events to layer-popups subsurfaces rather than their parent.~ With this change, when a layer-popup has a subsurface and cursor is on it, wl_pointer.enter/motion events are sent to the subsurface rather than its parent layer-popup surface. Though I haven't tested this yet, I think this follows wayland spec more closely.
- 3nd commit is just a minor refactoring to make the semantics of `get_cursor_context()` clearer.

EDIT: I made this PR to a draft to further improve the documentation of `get_cursor_context()`.

src/input/cursor.c
johanmalm: I can't remember but I think it's fine.

jlindgren90:
> This will send wl_pointer.enter/motion events to layer-popups subsurfaces rather than their parent. Though I haven't tested this yet, I think this follows wayland spec more closely.

I'm a bit confused by this. Popup != subsurface, which are we talking about here?

Popups (xdg_popup) should definitely get enter/leave events. I'm not sure about sub-surfaces (wl_subsurface). They are definitely *not* supposed to get keyboard input (protocol says this explicitly) and I was under the impression they did not get pointer input either, and were mainly intended as a drawing sub-buffer.

I've looked back at #1594, and from what I can see/recall the main point of it was to send keyboard focus to the layer-shell surface when clicking on a sub-surface -- not to send events to the sub-surface itself.

I probably need to look at the code and/or protocols more closely (with a good night's sleep) to be sure.

Edit: https://ppaalanen.blogspot.com/2013/11/sub-surfaces-now.html seems to say that sub-surfaces should receive pointer enter/motion events, unless they have an empty input region.

jlindgren90:
I read through the code and comments more closely, looks good to me. Nice work.

tokyo4j:
> > This will send wl_pointer.enter/motion events to layer-popups subsurfaces rather than their parent. Though I haven't tested this yet, I think this follows wayland spec more closely.
> 
> I'm a bit confused by this. Popup != subsurface, which are we talking about here?

Sorry, I updated my badly-written sentence. I was talking about the situation where layer-popup has a subsurface and the cursor is on it. I haven't found any apps to test this with, though (EDIT: I wrote a demo: https://github.com/tokyo4j/wayland-demo/tree/layer-popup).

> Popups (xdg_popup) should definitely get enter/leave events. I'm not sure about sub-surfaces (wl_subsurface). They are definitely not supposed to get keyboard input (protocol says this explicitly) and I was under the impression they did not get pointer input either, and were mainly intended as a drawing sub-buffer.
> 
> I've looked back at https://github.com/labwc/labwc/pull/1594, and from what I can see/recall the main point of it was to send keyboard focus to the layer-shell surface when clicking on a sub-surface -- not to send events to the sub-surface itself.
> 
> I probably need to look at the code and/or protocols more closely (with a good night's sleep) to be sure.
> 
> Edit: https://ppaalanen.blogspot.com/2013/11/sub-surfaces-now.html seems to say that sub-surfaces should receive pointer enter/motion events, unless they have an empty input region.

Yeah, this PR implements what you described. When a layer-popup has a subsurface and cursor presses it, the keyboard is given to the parent layer-popup surface but pointer events go to the subsurface.

EDIT: I was mixing up usual layer-surfaces and layer-popups... this PR doesn't change how we give keyboard focus to layer-popups (I'm not sure how we are doing this). This PR just gives pointer focus to a subsurface when cursor is on it, rather than parent layer-popups.

PR #3076, opened 2025-09-12
More small cleanups

jlindgren90:
Some more miscellaneous small cleanups. These are the last of what I've had collecting locally.

- A small refactoring to use fixed arrays for `rc.title_buttons_*`. These are just lists of `enum lab_node_type`, with a bounded size and no middle-insertions/removals, so linked lists are overkill. Also, the use of `wl_list_for_each[_reverse]` just to access the first or last entry in the list (corner button) was weird.

- Some little syntactic cleanups (add `const`, reorder initializers, add braces around a `case` block).

- Some minor tweaks to `.clang-format` to make it match our coding style a little better.

### DIFF
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -14,6 +14,9 @@
 
 #define BUTTON_MAP_MAX 16
 
+/* max of one button of each type (no repeats) */
+#define TITLE_BUTTONS_MAX ((LAB_NODE_BUTTON_LAST + 1) - LAB_NODE_BUTTON_FIRST)
+
 enum adaptive_sync_mode {
 	LAB_ADAPTIVE_SYNC_DISABLED,
 	LAB_ADAPTIVE_SYNC_ENABLED,
@@ -46,11 +49,6 @@ struct buf;
 struct button_map_entry {
 	uint32_t from;
 	uint32_t to;
-};
-
-struct title_button {
-	enum lab_node_type type;
-	struct wl_list link;
 };
 
 struct usable_area_override {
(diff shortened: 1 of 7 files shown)


include/config/rcxml.h
```diff
@@ -14,6 +14,9 @@
 
 #define BUTTON_MAP_MAX 16
 
+/* max of one button of each type (no repeats) */
+#define TITLE_BUTTONS_MAX (LAB_NODE_BUTTON_LAST + 1 - LAB_NODE_BUTTON_FIRST)
```
johanmalm: Maybe just personal preference, but suggest putting the `+1` at the end.

include/config/rcxml.h
```diff
@@ -88,8 +86,12 @@ struct rcxml {
 	char *theme_name;
 	char *icon_theme_name;
 	char *fallback_app_icon_name;
-	struct wl_list title_buttons_left;
-	struct wl_list title_buttons_right;
+
+	enum lab_node_type title_buttons_left[TITLE_BUTTONS_MAX];
+	int num_title_buttons_left;
```
johanmalm: We generally use prefix `nr_` rather than `num_`

src/config/rcxml.c
```diff
@@ -116,7 +116,8 @@ parse_window_type(const char *type)
  * desk         D           All-desktops toggle (aka omnipresent)
  */
 static void
-fill_section(const char *content, struct wl_list *list, uint32_t *found_buttons)
+fill_section(const char *content, enum lab_node_type *list, int *count,
```
johanmalm: Suggest calling `list` something else. Maybe `buttons`?

include/config/rcxml.h
jlindgren90: Hmm, I guess I tend to think of (last + 1) naturally together, as the "end" of the range. How about if I add parentheses, would that make it look better to you?

include/config/rcxml.h
jlindgren90: Ah, wasn't aware. Renamed.

src/config/rcxml.c
jlindgren90: Sure, renamed

johanmalm:
I'm good with it without moving that +1

johanmalm:
Thanks

PR #3080, opened 2025-09-14
view: fix <query monitor="current|left|right" />

### LINKED ISSUE #3077: Query monitor current
### Problem description

I was experimenting with keybinds a bit and found out that querying for "current" monitor inside conditional action is not working as expected. Querying monitor explicitly by name does work. 

### Steps to reproduce

1. Create a keybind

    ```xml
        <keybind key="W-S-Left">
          <action name="If">
            <query monitor="current" />
            <then>
              <action name="MoveToEdge" direction="left" />
            </then>
            <else>
              <action name="MoveToOutput" direction="left" />
            </else>
          </action>
        </keybind>
    ```
2. Test a keybind with window placed on current monitor (or simply with single output)

Expected result: MoveToEdge action is invoked
Actual result: MoveToOutput action is invoked

### labwc build source

Local build

### labwc version

0.9.1+git20250904

### labwc environment

### Distribution

Ubuntu 25.04

tokyo4j:
Fixes #3077.

The problem was the branch at the line 230 was always taken.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -215,20 +215,25 @@ view_matches_query(struct view *view, struct view_query *query)
 
 	if (query->monitor) {
 		struct output *current = output_nearest_to_cursor(view->server);
-
-		if (!strcasecmp(query->monitor, "current") && current != view->output) {
-			return false;
-		}
-		if (!strcasecmp(query->monitor, "left") &&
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, not tested

tokyo4j:
Tested with `ForEach` action + `<query monitor="current" />` and worked fine.

PR #3081, opened 2025-09-14
action: allow SnapToEdge to combine two cardinal directions

tokyo4j:
This PR allows to snap a window to e.g. up-left by running two actions:
- `<action name="SnapToEdge" direction="left" />`
- `<action name="SnapToEdge" direction="up" />`

Then running `<action name="SnapToEdge" direction="down" />` snaps it to left again. This behavior is the same as KWin.

https://github.com/user-attachments/assets/8c1e760d-c0d2-4fda-9656-3c517a7d14bf

This (and snap-to-adjacent-output feature added in #1284) can be disabled by adding `relative="no"` in `SnapToEdge`. This means the exact old behavior (only with snap-to-adjacent-output) cannot be restored, but I don't think there's much demand for it.

Any feedback for the algorithm and configurations are appreciated.


Consolatis:
No issue with the feature itself, just a bit hesitant to break user experience by changing it. We should likely also check what Openbox is doing.

tokyo4j:
> No issue with the feature itself, just a bit hesitant to break user experience by changing it. 

Yeah, I had the same concern. I considered setting `relative="no"` by default, but it also disables our existing behavior that snapping a window to the same direction twice moves it to the adjacent output (snap-to-adjacent-output feature I described above). To put it into a table:

| operation                        | before this PR                | relative="yes" in this PR  (default)   | relative="no" in this PR |
| -------------------------------- | ----------------------------- | ----------------------------- | ------------------------ |
| Snap a left-tiled window to up   | tiled to up                   | tiled to up-left              | tiled to up              |
| Snap a left-tiled window to left | moved to left-adjacent output | moved to left-adjacent output | tiled to left            |

I also considered letting `relative` argument not affect the snap-to-adjacent-output feature, but I prefer the clarity that the window is always snapped to the given `direction` if `relative="no"`.

> We should likely also check what Openbox is doing.

Openbox won't give us any hints as it doesn't support window tiling/snapping in the first place. I invetigated how KWin, Cosmic, Mutter and Xfce4 handles it:

### KWin

KWin has the same behavior as what this PR implements.

### Cosmic

Cosmic has the almost same behavior as KWin and this PR, but the difference is that snapping a left-tiled window to right maximizes it, whereas KWin simply tiles it to the right. I followed KWin's behavior because I don't like the extra step required for a left-tiled window to be right-tiled (https://bugs.kde.org/show_bug.cgi?id=493607).

### Mutter

Mutter doesn't allow to snap a window to up/down or corners by default. But as for left/right-tiling, its behavior is similar to `ToggleSnapToEdge` which this PR doesn't affect. When a window is tiled to left, snapping it to left again makes it floating, and snapping it right makes it right-tiled.

### Xfce4

Xfce4 has the exactly same behavior as `ToggleSnapToEdge`. It allows to snapping a window to up/down/left/right/up-left/up-right/down-left/down-right and then make it float by snapping it to the same direction again. When snapping a left-tiled window to up, it just gets up-tiled.

Consolatis:
> I also considered letting `relative` argument not affect the snap-to-adjacent-output feature, but I prefer the clarity that the window is always snapped to the given `direction` if `relative="no"`.

What about something like `s/relative/stack/` or `s/relative/combine/` and then have it disabled by default? That would then not affect the snap-to-adjescent-output feature and thus keep the existing behavior.

And it would be easy to change the default later if it turns out that its more intuitive for users and thus decouples the decision for UX change from the feature itself.

cunlem:
Hey thanks for your work on this feature. Just throwing my two thoughts

1. Have you considered making ToggleSnapToEdge act in the same way? The only difference would be that if action results in no change then the window becomes unsnapped. The problem I have with SnapToEdge is that I don't see a natural keybind candidate to unsnap action (I know that KWin doesn't have one either) so unsnapping requires mouse usage. 
2. Currently no action exists that allows directional snapping to regions. Is this something on the list of goals and hence needs to be considered here? I feel like snapping to corners is most useful on smaller to medium screens not the large ones which can make better use of custom grid.

tokyo4j:
I decided to take a simpler, non-breaking approach:
- Replace `relative` with `combine`.
- When `combine=yes`, snapping a window twice with two orthogonal directions to a screen corner.
- ~A window snapped to screen corner can no longer "grow" to fill the half of the screen; when snapping a up-right-tiled window to left simply snaps to the left rather than the top.~ EDIT: I restored the original behavior, since it felt weird that  a window shrunk to the corner cannot be grown back. I'm torn on this.
- A window snapped to a screen corner cannot be directly moved to an adjacent corner; when snapping a up-right-tiled window to the right, it's tiled to the right first, then snapping it to the right again moves to the right-adjacent output. The motivation of this change is that it allows us to share the same logic for `combine="yes|no"`, and it feels strange that `combine=yes` suddenly disallows to make a up-right-tiled window right-tiled by snapping it to the right once.
- Make `combine` applicable to `ToggleSnapToEdge`.
- (EDIT2) Enable `combine="yes"` for the default `W-` keybindings.

> 1. Have you considered making ToggleSnapToEdge act in the same way? The only difference would be that if action results in no change then the window becomes unsnapped. The problem I have with SnapToEdge is that I don't see a natural keybind candidate to unsnap action (I know that KWin doesn't have one either) so unsnapping requires mouse usage.

Thanks, I made `combine` applicable to `ToggleSnapToEdge` as described above.

> 2. Currently no action exists that allows directional snapping to regions. Is this something on the list of goals and hence needs to be considered here? I feel like snapping to corners is most useful on smaller to medium screens not the large ones which can make better use of custom grid.

I think we can add another action for that. Feel free to open a [discussion](https://github.com/labwc/labwc/discussions/new?category=feature-requests).

docs/labwc-actions.5.scd
```diff
@@ -92,11 +92,18 @@ Actions are used in menus and keyboard/mouse bindings.
...
+	Resize window to fill half or quarter the output in the given direction.
+
+	*direction* [up|down|left|right|up-left|up-right|down-left|down-right|center]
+		Direction in which to snap the window.
+
+	*combine* [yes|no]
+		Allows to snap a window to an output corner by combining two
+		directions. For example, snapping a window to *right* and then
+		to *up* places it in the *up-right* quarter of the output.
+		Default is no.
```
johanmalm, line 106: The default is `yes` following the last commit.

src/view.c
```diff
@@ -2142,25 +2142,38 @@ view_snap_to_edge(struct view *view, enum lab_edge edge,
 
 	view_set_shade(view, false);
 
-	if (across_outputs && view->tiled == edge && view->maximized == VIEW_AXIS_NONE) {
-		/* We are already tiled for this edge; try to switch outputs */
-		output = output_get_adjacent(view->output, edge, /* wrap */ false);
+	if (lab_edge_is_cardinal(edge) && view->maximized == VIEW_AXIS_NONE) {
+		enum lab_edge invert_edge = lab_edge_invert(edge);
+		enum lab_edge parallel_mask = edge | invert_edge;
```
johanmalm, line 2148: Suggest describing (or giving examples of) `parallel_mask`, `parallel_tiled` and `orthoganol_tiled` to make it easier for the reader.

johanmalm (commented):
Have left a couple of minor comments. Other than those, LGTM

The new code in `view_snap_to_edge()` looks elegant.

Although it's a slightly different behavior, I think the change is so gentle and intuitive that the new default is the way to go. I'll add a comment in the NEWS.md about how the old behavior can be achieved.

tokyo4j:
> Although it's a slightly different behavior, I think the change is so gentle and intuitive that the new default is the way to go. I'll add a comment in the NEWS.md about how the old behavior can be achieved.

Sorry, I think there's a misunderstanding in the default behavior. When I opened this PR, I changed the default behavior of `SnapToEdge` by setting `relative="yes"` by default. After @Consolatis's feedback, I decided to set `combine="no"` by default so that this PR doesn't break any existing user configurations, though I set `combine="yes"` for the default `W-` keybinds that are loaded when `rc.xml` is not found.

I don't have much preference on whether to enable/disable `combine`. I think the behavior with `combine="yes"` is more intuitive, but I also agree with @Consolatis's opinion that we should be really careful when we change the default behavior. So IMO, keeping `combine="no"` by default while setting `combine="yes"` for the builtin keybinds is a good compromise between advertising our new feature and keeping backward compatibility. But it's up to you.

docs/labwc-actions.5.scd
tokyo4j: The default is `no`, and the last commit updated the default keybinds, not the default argument of `(Toggle)SnapToEdge`. See https://github.com/labwc/labwc/pull/3081#issuecomment-3305561403.

src/view.c
tokyo4j: OK, added comments:

```c
		/* Represents axis of snapping direction */
		enum lab_edge parallel_mask = edge | invert_edge;
		/*
		 * The vector view->tiled is split to components
		 * parallel/orthogonal to snapping direction. For example,
		 * view->tiled=TOP_LEFT is split to parallel_tiled=TOP and
		 * orthogonal_tiled=LEFT when edge=TOP or edge=BOTTOM.
		 */
		enum lab_edge parallel_tiled = view->tiled & parallel_mask;
		enum lab_edge orthogonal_tiled = view->tiled & ~parallel_mask;
```

johanmalm:
I'm fine with this. Seems like a good compromise.

> though I set `combine="yes"` for the default `W-` keybinds that are loaded when rc.xml is not found

Just to clarify that we're aligned here though, it's not just when an `rc.xml` is not found, but also in the situation where someone has an `rc.xml` with

```
<keyboard>
<default/>
...
</keyboard>
```

...which I've got - and which I suspect is quite common - then the default (`<default/>`!!) behavior is different.
And just to clarify, I'm good with this and maintain my LGTM :smile:

docs/labwc-actions.5.scd
johanmalm: Thanks. I get it now. Makes sense.  

https://labwc.github.io/labwc-config.5.html#entry_keyboard_default

Would it make sense to tweak this line `W-<arrow> - resize window to fill half the output` to say `resize window to fill half or quarter of the output` or something like that?

tokyo4j:
> I'm fine with this. Seems like a good compromise.
> 
> > though I set `combine="yes"` for the default `W-` keybinds that are loaded when rc.xml is not found
> 
> Just to clarify that we're aligned here though, it's not just when an `rc.xml` is not found, but also in the situation where someone has an `rc.xml` with
> 
> ```
> <keyboard>
> <default/>
> ...
> </keyboard>
> ```
> 
> ...which I've got - and which I suspect is quite common - then the default (`<default/>`!!) behavior is different. And just to clarify, I'm good with this and maintain my LGTM 😄

Yeah, you're right. So this PR is breaking for users using `<default />`. I haven't used `<default />` because I didn't like the prior `A-` bindings, but now looking at `rc.xml.all` again, I agree that `<default />` is quite useful for many users.

docs/labwc-actions.5.scd
tokyo4j: Nice catch. I updated `labwc-config.5.scd` as you described.

PR #3089, opened 2025-09-19
PoC: Selectively enable privileged interfaces

xi:
This is a proof of concept to restrict privileged interfaces. Specifically:

- Disable privileged interfaces by default
- Allow users to selectively enable interfaces…
  - globally via `<core><permission>`
  - for actions via `<action><permission>`
  - for autostart via `<autostart><action><permission>`

I based it on 0.8.3 (because I am currently stuck with wlroots 0.18). I have been running this for some time without major issues. In my personal setup, I have enabled `zwlr_layer_shell_v1`, `zwlr_screencopy_manager_v1`, and `ext_session_lock_manager_v1` by default, and some additional interfaces for specific processes (e.g. `zwlr_data_control_manager_v1` for the clipboard manager).

Let me know what you think! (And of course, feel free to close this pull request if you think this is not a useful direction to investigate.)

## Comparison to security contexts

Support for the [security context protocol extension](https://wayland.app/protocols/security-context-v1) was added in https://github.com/labwc/labwc/pull/1817. There are two main differences from what I am proposing:

- This proposal focuses on authorization by explicitly allowing/preventing access to privileged interfaces. Security contexts on the other hand focus on authentication by providing a `sandbox_engine`, `app_id`, and `instance_id`. There is currently no standardized way how access to privileged interfaces should be influenced by that information. There is a proposal for [adding authorization to the protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/207), but I don't expect that it will be approved.
- This proposal is concerned about *which user interaction triggered the process*. For example, `grim` could get access to the screencopy interface only if it was triggered by a keyboard shortcut. Security contexts are concerned about what a process *is*. This is just a conceptual difference though. On a technical level, security contexts could be used the same way (by using a different app IDs depending on how the process was triggered).
- With this proposal, clients are not privileged by default and *gain* extra privileges by using a separate socket. With security contexts, the assumption seems to be that clients are privileged by default and  *lose* privileged by using a separate socket.
- This proposal does not require any new protocol extensions. It works entirely within the compositor.

This general idea was already discussed [back in 2014](https://mupuf.org/blog/2014/02/19/wayland-compositors-why-and-how-to-handle/). I am not sure what happened in the meantime.

## Limitation: Permissions are not inherited

Only the immediate child gets access to the privileged socket. Consider this example:

```xml
<autostart>
  <action>
    <name>Execute</name>
    <command>swayidle timeout 600 swaylock before-sleep swaylock lock swaylock</command>
    <permission>ext_idle_notifier_v1</permission>
    <permission>ext_session_lock_manager_v1</permission>
  </action>
</autostart>
```
  
`swayidle` will work fine, because it has access to the `ext_idle_notifier_v1` interface. However, swaylock will fail because it does not inherit the permissions.

## Limitation: privileged clients that are not started by the compositor

As far as I understand, xdg-desktop-portal-wlr cannot be started by the compositor. The only way to make it work is to globally allow `zwlr_layer_shell_v1` and `zwlr_screencopy_manager_v1`.

## Limitation: writable config

The one fatal flaw I see is that malicious applications could change the config and reload it without the user even noticing. See https://github.com/labwc/labwc/discussions/3036.

## What's wrong with sandboxes?

An alternative approach would be to wrap every single process that is spawned by the compositor in a sandbox. IMHO that is not realistic. I do most of my development work in a terminal. That includes pulling untrusted code from remote repositories and executing it. But even the terminals available on flathub [run their shells on the host](https://github.com/flatpak/xdg-desktop-portal/discussions/1808#discussioncomment-14387170).

Adding our own layer of sandboxing would also break portals: We would absolutely have to use xdg-dbus-proxy to filter dbus access because both `org.freedesktop.portal.Flatpak` and `org.freedesktop.systemd1.Manager` provide trivial sandbox escapes. But then the portals would see our proxy rather than the one created by flatpak. This would break their approach to authentication.

Sandboxing is great, but it is not (yet) fit for everything.


johanmalm:
I've got no issue with some sort of opt-in approach for protocols, but feels we ought to get a discussion + broad buy-in from @ahesford @Consolatis @tokyo4j and @jlindgren90 as it's quite a fundamental change.

I would prefer not changing defaults, because I suspect most people want to use it as it is.

Rather than the `<autostart>` approach, could we just list client identifiers (app_id or whatever) that should get access to each protocol or something like that? Could just default to `*`

What would stop a bad client from calling itself `grim` to get access?

> I based it on 0.8.3 (because I am currently stuck with wlroots 0.18)

If you want to run `0.9.1` or `master`, just build locally and meson will download `wlroots 0.19` and statically link it. It'll help avoid merge conflicts.

johanmalm:
Have set to draft as I feel it needs discussion before reviewing with a direct view to merge.

jlindgren90:
What is the end goal here from a user perspective?

If sandboxes are not "realistic" and a malicious application on this system can overwrite arbitrary files like labwc config (as mentioned), then I guess it can also `rm -rf $HOME/*`, no?

I fail to see the benefit of restricting layer-shell access in that case. It feels like patching a window crack while leaving the door wide open.

xi:
> Rather than the `<autostart>` approach, could we just list client identifiers (app_id or whatever) that should get access to each protocol or something like that? Could just default to `*`
>
> What would stop a bad client from calling itself grim to get access?

That's exactly the point why `<autostart>` is needed: There is no reliable way to identify running processes, so we need to define permissions based on how the process is spawned.

Consolatis:
Just for reference, some earlier work in this area:
- #1004
- #1002

You can already do something similar today with a small wrapper like [this](https://gitlab.freedesktop.org/wlroots/wlr-clients/-/merge_requests/6) (or one of the other ones). It creates its own restricted wayland listening socket via the security context protocol which can then be used by random applications if their `WAYLAND_DISPLAY` (or `XDG_RUNTIME_DIR`) env var is set accordingly (or bind-mounted into a filesystem namespace).

My main concern with this approach here is that it might imply a false sense of security. I also fail to see the actual use-case, if its untrusted applications then `rm -rf ~` is still possible, same for `echo "keylogger &" >> ~/.profile` or as mentioned changing the labwc config. The only working approach to that is some kind of sandboxing or at least privilege separation like running untrusted applications with a different UID / GID.

xi:
> I fail to see the benefit of restricting layer-shell access in that case. It feels like patching a window crack while leaving the door wide open.

> My main concern with this approach here is that it might imply a false sense of security.

I expected this reaction and I honestly wish I had a good answer.

There are some use cases (e.g. terminals) for which flatpak is not suitable. We have no security story for those cases yet. My hope was that this approach could be a piece of the puzzle. I wanted to share the idea and gather feedback.

I have experimented with different approaches and this one was the best I have found so far. What makes this different from prior discussions like e.g. #1004 is the idea to use `WAYLAND_SOCKET` rather than `WAYLAND_DISPLAY`, which results in a much smaller attack surface because the privileged socket is never exposed to other processes.

But you are absolutely right that this is not a complete solution on its own. Maybe it could be combined with some light-weight sandboxing to make it more complete. If you think there is something here, maybe we can work together to fill in the blanks. Otherwise, feel free to close the PR. It can still serve for future reference.

ahesford:
I don't really have a strong opinion about protocol blacklists in general, but my general sense is that they are more trouble than they are worth. I am, however, strongly opposed to the `autostart` logic. Protocol approval should *always* be gated by application type, not by fragile and racy logic that determines an application was started by the autostart script.

Consolatis:
> Protocol approval should always be gated by application type, not by fragile and racy logic that determines an application was started by the autostart script.

Well, to be fair I don't see the approach as being fragile or racy. `WAYLAND_SOCKET` should be fine for that as long as applications actually prioritize it over `WAYLAND_DISPLAY` or simply scanning `XDG_RUNTIME_DIR` for sockets.

And in this case, failure to use it would leave applications with less privileges so I think that part would work fine. However, I am still not a fan of the approach for the stated reasons above.

> There are some use cases (e.g. terminals) for which flatpak is not suitable.

I completely get the point about flatpak, what about using a different `XDG_RUNTIME_DIR` or `WAYLAND_DISPLAY` for those?

Edit:
I am also missing an actual use-case here, e.g. what application should be prevented to use e.g. the layershell protocol but allowed to modify everything in the home directory and why?

xi:
> what about using a different `XDG_RUNTIME_DIR` or `WAYLAND_DISPLAY` for those?

If you want only specific clients to get access to privileged interfaces, `WAYLAND_SOCKET` is the only way I know to do that. All other options require that you create a named socket in the current mount namespace, at which point any other process in that namespace can also access that socket.

If the only goal is to restrict a terminal, you could just as well leave all interfaces enabled by default and create an unprivileged named socket that is mounted into the terminal's namespace.

>  I am also missing an actual use-case here, e.g. what application should be prevented to use e.g. the layershell protocol but allowed to modify everything in the home directory and why?

I think what you are describing could match the expectations many users have of terminals: They allow you to do basically anything, but still protect you from keyloggers.

But this is also just the  principle of least privileges. I do not expect any random client to be able to make screenshots or grab the keyboard. Admittedly, I also do not expect any random process to be able to delete all of my files. But the latter is out of scope for labwc, while the former is a direct consequence of its choice of protocol extensions.

ahesford:
"Racy" was a misjudgment based on a cursory reading of this PR, but the logic os fragile because it is still often possible to shadow trusted applications with malicious actors just by installing in a PATH location with higher precedence.

Protection from key loggers in terminal emulators is a straw man. If you have filesystem access, you can just wrap your shell profile as Consolatis suggests and grab all you like. It doesn't require Wayland to do so and isn't really any more difficult. If you run untrusted applications, you can't trust what they'll do, period.

Consolatis:
> If you want only specific clients to get access to privileged interfaces, `WAYLAND_SOCKET` is the only way I know to do that. All other options require that you create a named socket in the current mount namespace, at which point any other process in that namespace can also access that socket.
> [..]
> But this is also just the principle of least privileges. I do not expect any random client to be able to make screenshots or grab the keyboard. Admittedly, I also do not expect any random process to be able to delete all of my files. But the latter is out of scope for labwc, while the former is a direct consequence of its choice of protocol extensions.

Hm, this is a valid point. Having a user controllable block list for protocols is something we should definitely implement at some point. But that indeed raises the question of how some specific applications can then still use those protocols. And it gets more tricky if we think about xdg-desktop-portal-wlr which is not spawned by the compositor.

Maybe there could be some simple wrapper which could be used which rather than just spawning a listening socket permanently in `XDG_RUNTIME_DIR` it connects to it itself, sets `WAYLAND_SOCKET` and then closes the listening socket before spawning the client . Or like suggested in this PR via compositor `socketpair()` + `wl_client_create()` which is also what sway is doing AFAIR.

I need to think a bit about this. Further discussions / ideas obviously welcome.

xi:
I think we all agree that for real security, we need to restrict both Wayland interfaces and file system access. The question is where the Wayland restrictions should be implemented: In the compositor or in the sandbox engine (or both).

Arguments for the compositor:

- Lower complexity in the sandbox engine (which might trigger further innovation in the field)
- Better security if a sandbox is not available
- Better security if a sandbox is available, but does not properly restrict Wayland interfaces
- The compositor is uniquely positioned to understand which clients need which interfaces

Arguments against the compositor

- Higher complexity in the compositor
- Backwards-incompatible changes to configuration
- Potential of false sense of security

PR #3096, opened 2025-09-21
font: some minor cleanups

tokyo4j:
The 1st commit removes the 4-pixel padding on the right of font buffers. Actually I have no idea about the purpose of `+= 4;`. I guess the comment (`we put a 2 px edge on each side`) was inaccurate, because it doesn't affect the x-position of the text in the buffer.

The 2nd commit allows null/empty string in `font_buffer_create()` so that so that we don't need to check for NULL/empty strings when calling `scaled_font_buffer_update()`. This also fixes error spams when trying to show the thumbnail-style window switcher for windows with no title.

The 3rd commit is just a minor fixup related to the 2nd commit.


tokyo4j:
I realized this PR makes the menu text and the arrow too close. I'll fix it by sliding the arrow by `menu.items.padding.x`, like we are doing for the menu icons.

EDIT: Done.

johanmalm:
LGTM

Wow, yes those 4 pixels go back to a very early commit! I can't remember, but guess it was easier to hardcode some padding than to create variables/defines at the time :smile:

Possibly inspired by this:
- https://github.com/danakj/openbox/blob/9e8813e111cbe6c1088f6abbc771a29470f05fc2/obrender/font.c#L172

PR #3097, opened 2025-09-22
config: add `<core><promptCommand>`

johanmalm:
-- EDIT --

See commit messages for description.

With zenity:

zenity

### DIFF
--- a/src/action-prompt-command.c
+++ b/src/action-prompt-command.c
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include "action-prompt-command.h"
+#include <stdio.h>
+#include <wlr/util/log.h>
+#include "action.h"
+#include "common/buf.h"
+#include "labwc.h" /* for gettext */
+#include "theme.h"
+
+enum {
+	LAB_PROMPT_NONE = 0,
+	LAB_PROMPT_MESSAGE,
+	LAB_PROMPT_NO,
+	LAB_PROMPT_YES,
+	LAB_PROMPT_BG_COL,
+	LAB_PROMPT_TEXT_COL,
+
+	LAB_PROMPT_COUNT
+};
+
+typedef void field_conversion_type(struct buf *buf, struct action *action,
+		struct theme *theme);
+
+struct field_converter {
+	const char fmt_char;
+	field_conversion_type *fn;
+};
+
+/* %m */
+static void
+set_message(struct buf *buf, struct action *action, struct theme *theme)
+{
+	buf_add(buf, action_get_str(action, "message.prompt", "Choose wisely"));
+}
+
+/* %n */
+static void
+set_no(struct buf *buf, struct action *action, struct theme *theme)
+{
+	buf_add(buf, _("No"));
+}
+
+/* %y */
+static void
+set_yes(struct buf *buf, struct action *action, struct theme *theme)
+{
+	buf_add(buf, _("Yes"));
+}
+
+/* %b */
+static void
+set_bg_col(struct buf *buf, struct action *action, struct theme *theme)
+{
+	buf_add_hex_color(buf, theme->osd_bg_color);
+}
+
+/* %t */
+static void
+set_text_col(struct buf *buf, struct action *action, struct theme *theme)
+{
+	buf_add_hex_color(buf, theme->osd_label_text_color);
+}
+
+static const struct field_converter field_converter[LAB_PROMPT_COUNT] = {
+	[LAB_PROMPT_MESSAGE]           = { 'm', set_message },
+	[LAB_PROMPT_NO]                = { 'n', set_no },
+	[LAB_PROMPT_YES]               = { 'y', set_yes },
+	[LAB_PROMPT_BG_COL]            = { 'b', set_bg_col },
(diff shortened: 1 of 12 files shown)


Consolatis:
Didn't yet look at the diff but some generic remarks:

- Personally I am a fan of using the long-form arguments for defaults because they serve as self-describing docs (e.g. `--button-dismiss` and `--message`).
- The timeout seems weird to me for prompts, should we add a `--timeout 0`?
- Docs are missing

src/common/buf.c
```diff
@@ -128,6 +128,13 @@ buf_add_fmt(struct buf *s, const char *fmt, ...)
 	s->data[s->len] = 0;
 }
 
+void
+buf_add_hex_color(struct buf *s, float color[4])
+{
+	buf_add_fmt(s, "#%02x%02x%02x", (int)(color[0] * 255),
+		(int)(color[1] * 255), (int)(color[2] * 255));
+}
```
Consolatis, line 153: I think we have pre-multiplied alpha in the color argument, but not 100% sure right now.

src/action-prompt-command.c
```diff
@@ -0,0 +1,123 @@
...
+			fmt[fmt_position++] = 's';
+			fmt[fmt_position++] = '\0';
+			snprintf(converted_field, sizeof(converted_field),
+				fmt, field_result.data);
+			buf_add(buf, converted_field);
+			goto next;
+		}
+
+		wlr_log(WLR_ERROR,
+			"invalid format character found for osd %s: '%c'",
```
Consolatis: "for prompt command"

src/action-prompt-command.c
```diff
@@ -0,0 +1,123 @@
...
+				continue;
+			}
+			field_converter[i].fn(&field_result, action, theme);
+			fmt[fmt_position++] = 's';
+			fmt[fmt_position++] = '\0';
+			snprintf(converted_field, sizeof(converted_field),
+				fmt, field_result.data);
+			buf_add(buf, converted_field);
+			goto next;
+		}
```
Consolatis: This seems to be a copy of `osd-field.c` but without support for the length modifier. Do we actually need `char fmt[]` and `char converted_field[]` here? It would always be `%s` anyway so I think we could just remove all of that completely and also remove `snprintf()`.

So the only thing left would be something like `field_converter[i].fn(buf, action, theme)` unless I missed something.

src/common/buf.c
johanmalm: Could catch. Have un-pre-multiplied.

src/action-prompt-command.c
johanmalm: Done. Thanks.

src/action-prompt-command.c
johanmalm: Thanks. Yes, did copy `osd-field.c`. You're absolutely right. Have simplified as suggested. Leaves very little actually :smile:

johanmalm:
>     * Personally I am a fan of using the long-form arguments for defaults because they serve as self-describing docs (e.g. `--button-dismiss` and `--message`).

Good point. Done.

>     * The timeout seems weird to me for prompts, should we add a `--timeout 0`?

Have set `--timeout 0`

That revealed a bug in `labnag` so was worthwhile from that perspective too.

>     * Docs are missing

Added.

I also did the following:
- Reverted the button colors, but it think it looks better.
- Set bottom-border and button-border to 1px rather than the default 2.

With built-in (default) theme:

20250923_19h29m56s_grim

With `teallach` theme:

20250923_19h29m31s_grim


Thank goodness for peer-review :smile: Admittedly, I did rush the original patch a bit because I wanted to get it done before starting cool-down at the end of the week, but it did leave quite a lot to be desired!

johanmalm:
Ready for another review from my perspective.

Needs an interactive rebase + tidying up original commit message before merge.

src/common/buf.c
Consolatis: Needs a `!= 0` check (or something like `if (!alpha) { buf_add(s, "#00000000"); return; }`), otherwise we may divide by zero. `set_cairo_color()` uses `if (alpha == 0.0f) { ..; return; }`.

Consolatis:
Small divide by zero check missing, other than that LGTM. Not actually tested.
Very nice!

As for the design, I can't really say much. Personally I am a bit bothered by the buttons looking like they are not centered correctly but that might very well be a /me issue.

Edit:
One thing I noticed is that `rc.xml.all` is not updated. Might be enough to just put the default string in there and (if at all) refer to the man page for more information.

johanmalm:
> Small divide by zero check missing

Fixed.

> Personally I am a bit bothered by the buttons looking like they are not centered correctly...

Looks like they're centered with the default button-border-size (3px) but not with 2 or 1. Have restored the default to make it look good. Suggest we look at labnag button border in a separate PR.

 
> Edit: One thing I noticed is that `rc.xml.all` is not updated. Might be enough to just put the default string in there and (if at all) refer to the man page for more information.

Done.

Will rebase and tidy up commit messages now.

johanmalm:
Okay. Third time lucky. Ready to go as far as I'm concerned.

I know you said LGTM, but do you want to cast your eyes over it again?

20250924_19h48m26s_grim

src/common/buf.c
johanmalm: Done.

PR #3105, opened 2025-09-26
Fix ci freebsd

johanmalm:


johanmalm:
Whoops... screwed up some commits there. Hang on :smile:

PR #3107, opened 2025-09-26
Add build option to show unused functions; and make some functions private

johanmalm:
See #2975

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -134,6 +134,15 @@ if get_option('static_analyzer').enabled()
   add_project_arguments(['-fanalyzer'], language: 'c')
 endif
 
+link_args = []
+if get_option('sections').enabled()
+  add_project_arguments(['-ffunction-sections'], language: 'c')
+  link_args += [
+    '-Wl,--gc-sections',
+    '-Wl,--print-gc-sections',
+  ]
+endif
+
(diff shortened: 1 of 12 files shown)


meson.build
```diff
@@ -134,6 +134,16 @@ if get_option('static_analyzer').enabled()
...
 
+if get_option('sections').enabled()
+  add_project_arguments(['-ffunction-sections'], language: 'c')
+  link_args = [
+    '-Wl,--gc-sections',
+    '-Wl,--print-gc-sections',
+  ]
+else
+  link_args = []
+endif
```
tokyo4j, line 144: nit: I suggest the style below as we may conditionally add other linker flags in the future.

```suggestion
link_args = []
if get_option('sections').enabled()
  add_project_arguments(['-ffunction-sections'], language: 'c')
  link_args += [
    '-Wl,--gc-sections',
    '-Wl,--print-gc-sections',
  ]
endif
```

tokyo4j:
LGTM other than my suggestion above. Feel free to merge this PR, as this doesn't modify any logic.

meson.build
johanmalm: Good suggestion. Done.

johanmalm:
Thanks for review, and for finding the `-ffunction-sections` stuff in the first place :smile:

PR #3110, opened 2025-09-29
Fix unit tests + add CI

tokyo4j:
- 38a1a9b broke t/xml.c due to macros.h requiring wlr/version.h. This PR fixes it by adding wlroots as a direct dependency of the test executables.
- Use `xmlBufferContent()` instead of `xmlBuffer->content`, which has been deprecated.
- Add unit tests to CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,7 +92,7 @@ jobs:
           apt-get upgrade -y
           apt-get install -y git gcc clang gdb xwayland
           apt-get build-dep -y labwc
-          apt-get build-dep -y libwlroots-0.18-dev
+          apt-get build-dep -y libwlroots-0.19-dev
 
       - name: Install FreeBSD dependencies
         if: matrix.name == 'FreeBSD'


tokyo4j:
I confirmed reverting the second commit (`test/xml: use xmlBufferContent()`) successfully causes CI error.

Consolatis:
While on it, could you `s/libwlroots-0.18-dev/libwlroots-0.19-dev/` in the Debian Build depends? That should fix the Debian CI. wlroots 0.19.x entered Debian Testing 2 days ago.

Consolatis:
The CI part LGTM.
Can't really say anything in regards to the unit test itself and its meson file but as it now apparently successfully runs via CI that seems fine as well.

Thanks for the fix!

johanmalm:
Thanks! I didn't realise we weren't running `meson test` in the CI :smile: Whoops.

We implemented it in `labwc-menu-generator` which is probably why I got my wires crossed on that

https://github.com/labwc/labwc-menu-generator/blob/19ea2d27eaea50a8ef031bc1ea0c2ccfebb92e3c/.github/workflows/build.yml#L100-L105

Issue #3111, opened 2025-09-29
Unroll window on focus / Alt+Tab


Amodio:
### Problem description

Hello,
When I Alt+Tab to a rolled window, it does not unroll it despite my `rc.xml`:
```bash
[...]
  <keyboard>
    <numlock>on</numlock>
    <layoutScope>global</layoutScope>
    <repeatRate>25</repeatRate>
    <repeatDelay>600</repeatDelay>
    <keybind key="A-Tab">
      <action name="NextWindow" />
      <action name="Unshade" />
      <action name="Focus" />
    </keybind>
    <keybind key="A-S-Tab">
      <action name="PreviousWindow" />
      <action name="Unshade" />
      <action name="Focus" />
    </keybind>
[...]
```
Is it currently possible to have this behavior?

Thanks!

### Steps to reproduce

1. Launch labwc (wayland)
2. Roll a window
3. Alt+Tab to it: it is not unrolled automatically.

### labwc build source

Release

### labwc version

labwc 0.9.1 (-xwayland +nls +rsvg +libsfdo)

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Alpine

domo141:
rolled window shouls behave like minimized window when nextwindow/previouswindow reaches it
(with no config change/with "default" config). it is known issue that it (currently) does not.

i.e. unroll when reached, then roll back unless "final selection" (I don't know exact terms :D)

In my case if "final selection" the window in question gets focus (albeit rolled)

Focus is not changing while keeping alt pressed and clicking tab / arrow keys...

johanmalm:
I haven't tested, but suspect you could achieve this by adding `<action name="Unshade" />`

Amodio:
> rolled window shouls behave like minimized window when nextwindow/previouswindow reaches it (with no config change/with "default" config). it is known issue that it (currently) does not.

Yes, it should behave just like a click on the rolled window bar with this configuration (that works):
```
    <context name="TitleBar">
      <mousebind button="Left" action="Press">
        <action name="Focus" />
        <action name="Unshade" />
        <action name="Raise" />
      </mousebind>
[...]
      <mousebind direction="Down" action="Scroll">
        <action name="Unshade" />
        <action name="Focus" />
      </mousebind>
    </context>
```
I did not find any previous issue for that, do you have any discussion about it?

Consolatis:
[...]

PR #3114, opened 2025-09-30
Auto unshade window on Alt+Tab

Amodio:
This PR is for fixing bug issue #3111: a window alt+tabbed should be automatically focused & unrolled.

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -93,8 +93,14 @@ end_cycling(struct server *server)
 	}
 
 	struct view *cycle_view = server->osd_state.cycle_view;
+	/* Store the shaded state before osd_finish() clears the cycle_view */
+	bool was_shaded = cycle_view && cycle_view->shaded;
 	/* FIXME: osd_finish() transiently sets focus to the old surface */
 	osd_finish(server);
+	/* Unshade if we're actually selecting a window (not canceling) */
+	if (cycle_view && was_shaded) {
+		view_set_shade(cycle_view, false);
+	}
 	/* Note that server->osd_state.cycle_view is cleared at this point */
 	desktop_focus_view(cycle_view, /*raise*/ true);
 }


Amodio:
Removed the unnecessary code added to focus the window, only unrolling/unshading is needed.

Consolatis:
Thanks for your PR!

We are currently in the cooldown period before a release so this PR will have to wait a bit.
In general, if Openbox indeed unshades windows when switched to them via A-Tab I would be for doing the same in labwc.

johanmalm:
Yes, for after `0.9.2`

LGTM. Not tested.

I've tried with Openbox, and the can confirm that clients are unshaded on finishing the window-switcher cycling.

We could consider making this configurable with a `<finalaction>` like openbox but suggest just hard-coding for now (i.e. the way this PR is)

Amodio:
Thanks, I am not certain it was the default behavior of openbox because my `rc.xml` was using a `finalactions` block:
```xml
    <!-- Keybindings for window switching -->
    <keybind key="A-Tab">
      <action name="NextWindow">
        <finalactions>
          <action name="Focus"/>
          <action name="Raise"/>
          <action name="Unshade"/>
        </finalactions>
      </action>
    </keybind>
```
I have not found this block implemented in labwc yet, so hardcoding the unshading behavior seems okay to me.

domo141:
IMO unshading/shading should work like unminimizing/minimizing while doing A-tabbing.

i.e. when a-tab reaches shaded window, it would be unshaded (temporarily), and if one continues past that window, it would be re-shaded.
EDIT: and raise temporarily, then "drop" in it's original place in stacking order
if the "final action" is to choose that particular window then unshade it. 

(all of that should not be too difficult, if following the unminimize/minimize path in code is easy)

I recall checking first change of this, and what I understood was it (only) unshaded at "finalaction", not
temporarily in-between -- but I may be wrong :D

Consolatis:
> IMO unshading/shading should work like unminimizing/minimizing while doing A-tabbing.
> 
> i.e. when a-tab reaches shaded window, it would be unshaded (temporarily), and if one continues past that window, it would be re-shaded. EDIT: and raise temporarily, then "drop" in it's original place in stacking order if the "final action" is to choose that particular window then unshade it.

Good point, just tested and we indeed do not temporarily unshade. We do raise temporarily though.

Is the temporary unshade behavior the same in Openbox? Not all users may want to accidentally expose the window contents via A-Tab. Maybe we should add a setting for that.

If we want to add the temporary unshade we'd need to store another flag for the window originally being shaded and restore it in `restore_preview_node()`, similar to minimized / (`preview_was_enabled`). Or we throw the node away and simply store the view in `server->osd_state`.

Rather than using `view_set_shade()` we should likely use `wlr_scene_node_set_enabled()` directly here so we don't interfere with any state in `struct view`. That is debatable though. Anyway, the temporary unshade sounds like it should be an own PR.

As for this PR, the `view_set_shade(view, false)` should likely happen in `end_cycling()` (`src/input/keyboard.c`) rather than in `osd_finish()` because we don't want to unshade if canceling the window switcher via `ESC`.

ahesford:
It isn't a bug that the window doesn't unshade during Alt-Tab switching.

As noted, users may wish for windows to remain shaded for privacy. This applies to minimized windows as well, but there is a key difference: it is possible to interact with shaded windows even though the content is not visible. Thus, it is not unexpected to be able to Alt-Tab over to a shaded window without showing it and take some action (like pausing a video with a keypress, or invoking a close action).

Temporarily raising a minimized window during Alt-Tab makes sense for a couple of reasons that do not really apply to shading. First, this is the only sensible visual representation of the minimized window while cycling through them. What else would we display? An empty frame? This concern doesn't apply to cycling through shaded windows: the rolled-up titlebar *is its own visible representation*. Second, if you are cycling through minimized windows and stop on one, it is clear that you intend to restore the window to interact with it. However, given the concern above (or any number of other reasons why you may not wish to expand the contents permanently), user intent is *not* clear when stopping on a shaded window. How would we distinguish between the desire to simply focus the window for input and the desire to focus and unshade it? What would we do after the Alt-Tab action is completed, restore the shaded state? It would be supremely weird to have a window temporarily unshade while cycling with Alt-Tab, only to have it re-shade itself when the user completes the action with the shaded window closed. But leaving it unshaded is not necessarily the right call, either.

I would accept a configurable option to temporarily unshade windows when they are focused during Alt-Tab cycling, but it definitely should not be a hard-coded change. I also suspect that there are going to be unanticipated weird edge cases in the state transitions for Shaded -> Temporarily Unshaded During Alt-Tab -> Complete With Focus on Shaded Window.

Amodio:
Thanks for your feebacks, yes I do not remember that windows were unrolled on just cycling with Alt+Tab (and I was using the `finalactions` block); I do not have this need.
Feel free to test the modifications and edge cases I may not be aware of, this PR is simpler :)

Consolatis:
After some discussions on IRC, the current plans of how we want to proceed with this are roughly:
- add a config option `<windowSwitcher unshade="yes|no" />`
- if enabled, shaded windows will be temporarily unshaded while switching in A-Tab and permanently unshaded on the final selection
- if disabled, the behavior keeps being the same as right now (no temporary or final unshade)
- at the current state the default for the new config option is "yes" but other opinions definitely welcome

I can do a PR for the config option and a-tab preview in the next weeks.

Amodio:
Alright, sounds good thx.

PR #3117, opened 2025-10-01
action: simplify action_prompt_command()

tokyo4j:
For post-0.9.2.

I think using function pointers was overkill.


johanmalm:
Agree. LGTM

PR #3118, opened 2025-10-01
Add osd.window-switcher.style-classic.item.active.{border,bg}.color

tokyo4j:
For post-0.9.2.

These theme options configures the border/background of selected window item in the `classic` style window siwtcher. Previously, the OSD text color (default: black) was used as the border and background was transparent.

Now the default theme is updated to use #589bda as the border color and #c7e2fc as the background color, which are the same as the `thumbnail` style window switcher.

Screenshot of the new default look:

20251001_14h35m54s_grim

I'll appreciate feedbacks. If the default colors are not preferable, I think we can update the default colors of `thumbnail` style window switcher for consistency before 0.9.2 release.

### DIFF
--- a/src/theme.c
+++ b/src/theme.c
@@ -604,6 +604,8 @@ theme_builtin(struct theme *theme, struct server *server)
 	theme->osd_window_switcher_classic.item_padding_x = 10;
 	theme->osd_window_switcher_classic.item_padding_y = 1;
 	theme->osd_window_switcher_classic.item_active_border_width = 2;
+	theme->osd_window_switcher_classic.item_active_border_color[0] = FLT_MIN;
+	theme->osd_window_switcher_classic.item_active_bg_color[0] = FLT_MIN;
 	theme->osd_window_switcher_classic.item_icon_size = -1;
 
 	theme->osd_window_switcher_thumbnail.max_width = 80;
@@ -988,6 +990,12 @@ entry(struct theme *theme, const char *key, const char *value)
 		switcher_classic_theme->item_active_border_width =
 			get_int_if_positive(value,
 				"osd.window-switcher.style-classic.item.active.border.width");
+	}
+	if (match_glob(key, "osd.window-switcher.style-classic.item.active.border.color")) {
+		parse_color(value, switcher_classic_theme->item_active_border_color);
+	}
+	if (match_glob(key, "osd.window-switcher.style-classic.item.active.bg.color")) {
+		parse_color(value, switcher_classic_theme->item_active_bg_color);
 	}
 	if (match_glob(key, "osd.window-switcher.style-classic.item.icon.size")
 			|| match_glob(key, "osd.window-switcher.item.icon.size")) {
@@ -1745,6 +1753,14 @@ post_processing(struct theme *theme)
 		 */
 		memcpy(theme->osd_border_color, theme->osd_label_text_color,
 			sizeof(theme->osd_border_color));
+	}
+	if (switcher_classic_theme->item_active_border_color[0] == FLT_MIN) {
+		blend_color_with_bg(switcher_classic_theme->item_active_border_color,
+			theme->osd_label_text_color, 0.50, theme->osd_bg_color);
+	}
+	if (switcher_classic_theme->item_active_bg_color[0] == FLT_MIN) {
+		blend_color_with_bg(switcher_classic_theme->item_active_bg_color,
+			theme->osd_label_text_color, 0.15, theme->osd_bg_color);
 	}
 	if (switcher_thumb_theme->item_active_border_color[0] == FLT_MIN) {
 		blend_color_with_bg(switcher_thumb_theme->item_active_border_color,
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -327,6 +327,14 @@ all are supported.
 	Border width of the selection box in the window switcher in pixels.
 	Default is 2.
 
+*osd.window-switcher.style-classic.item.active.border.color*
+	Border color around the selected window switcher item.
+	Default is *osd.label.text.color* with 50% opacity.
+
+*osd.window-switcher.style-classic.item.active.bg.color*
+	Background color of the selected window switcher item.
+	Default is *osd.label.text.color* with 15% opacity.
+
 *osd.window-switcher.style-classic.item.icon.size*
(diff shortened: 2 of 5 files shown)


Consolatis:
This color doesn't really work with dark themes, e.g. `Numix` for example. Reason being that the OSD font color is very bright so you end up with unreadable text under the selection:

20251001_18h17m56s_grim

Consolatis:
This is `Numix` with `#c7e2fc50`:
20251001_18h24m47s_grim

And this is the default theme:
20251001_18h24m12s_grim

tokyo4j:
That's a good point. To fix it, I updated the default colors like this:

- `osd.window-switcher.style-classic.item.active.border` inherits `osd.label.text.color` as before.
- `osd.window-switcher.style-classic.item.active.bg` inherits `osd.window-switcher.style-classic.item.active.border` with 15% opacity.

To keep subpixel rendering for fonts, `switcher_classic_theme->item_active_bg_color` stores the blending result of `osd.bg.color` and 15% opacity `osd.window-switcher.style-classic.item.active.border`.

I personally prefer the blue border and light-blue background like shown in the description, but I think we should make it inherit `osd.label.text.color` as before, since some theme may have light-blue osd text color which makes the text unreadable.

FWIW, Openbox has `osd.active.label.bg` for the background texture of selected window item. So maybe we could use it as the default background/border color, but I'm afraid it complicates our logic for theme inheritances and confuses users.

Also, I updated the default color of `thumbnail` style window switcher to fix the same problem.

Now both styles of the default window switcher look like this:

20251003_03h14m03s_grim

20251003_03h14m25s_grim

src/theme.c
```diff
@@ -1721,6 +1757,26 @@ post_processing(struct theme *theme)
...
+	if (switcher_classic_theme->item_active_border_color[0] == FLT_MIN) {
+		memcpy(switcher_classic_theme->item_active_border_color,
+			theme->osd_label_text_color,
+			sizeof(theme->osd_label_text_color));
+	}
+	if (switcher_classic_theme->item_active_bg_color[0] == FLT_MIN) {
+		blend_color_with_bg(switcher_classic_theme->item_active_bg_color,
+			switcher_classic_theme->item_active_border_color, 0.15,
+			theme->osd_bg_color);
+	}
```
Consolatis, line 1764: Thoughts about this variant? It reduces the bright border in dark themes and seems preferable to me but that is just a personal preference.
```c
	if (switcher_classic_theme->item_active_border_color[0] == FLT_MIN) {
		blend_color_with_bg(switcher_classic_theme->item_active_border_color,
			theme->osd_label_text_color, 0.5,
			theme->osd_bg_color);
	}
	if (switcher_classic_theme->item_active_bg_color[0] == FLT_MIN) {
		blend_color_with_bg(switcher_classic_theme->item_active_bg_color,
			theme->osd_label_text_color, 0.15,
			theme->osd_bg_color);
	}
```

tokyo4j: Looks good to me. I agree that the pure-black border in the default theme and the pure-white border in `Numix` theme are too prominent and I like your 50% gray border.

tokyo4j: Done.

tokyo4j:
Now the window switcher looks like this:

||`style=classic`|`style=thumbnail`|
|-|-|-|
|default|20251003_14h03m03s_grim | 20251003_14h03m18s_grim |
|Numix| 20251003_14h03m31s_grim | 20251003_14h03m40s_grim |

johanmalm:
I think that inheriting `osd.label.text.color` with alpha as you've done it makes a lot of sense to avoid out-of-the-box surprises against existing themes.

I'm confused by the values in `docs/themerc`. How can `r`, `g` and `b` be different if the default is `#000000 50%` and `#000000 15%`? I think I'm right in saying that the defaults comes from falling back on `osd.label.text.color` which in turns falls back on `window.active.label.text.color` which is `#000000` by default.


```
osd.window-switcher.style-thumbnail.item.active.border.color: #706f6d
osd.window-switcher.style-thumbnail.item.active.bg.color: #bfbcba
```

LGTM other that. Have tested - works well.

I wasn't aware of `osd.active.label.bg`. It does not appear to be in the wiki, but interestingly it exists in the Clearlooks-3.4 `themerc` file. I don't think that Openbox's theme spec for `osd` is really sufficient for us, so current trajectory makes sense (i.e. ignore `osd.active.label.bg`).

tokyo4j:
> I'm confused by the values in docs/themerc. How can r, g and b be different if the default is #000000 50% and #000000 15%? I think I'm right in saying that the defaults comes from falling back on osd.label.text.color which in turns falls back on window.active.label.text.color which is #000000 by default.

`#000000 50%` and `#706f6d` are basically the same. The downside of describing `#000000 50%` in `themerc` is that it disables subpixel antialiasing for the window switcher fields/titles, which slightly degrades UX when users copy `docs/themerc` to their theme.

Also, do you have any thoughts on https://github.com/labwc/labwc/pull/3123#issuecomment-3364251460? I prefer merging only the first commit for now, because I think the changes in `osd-classic.c` is bigger and harder to follow compared to the changes in `theme.c`.

johanmalm:
> `#000000 50%` and `#706f6d` are basically the same. The downside of describing `#000000 50%` in `themerc` is that it disables subpixel antialiasing for the window switcher fields/titles, which slightly degrades UX when users copy `docs/themerc` to their theme.

I got it now, it's what we get when we do `#000000 50%` on top of `#e1dedb` (default `osd.bg.color`)

> Also, do you have any thoughts on [#3123 (comment)](https://github.com/labwc/labwc/pull/3123#issuecomment-3364251460)? I prefer merging only the first commit for now, because I think the changes in `osd-classic.c` is bigger and harder to follow compared to the changes in `theme.c`.

Good idea about just doing the first commit. Let's do that before the release and leave the rest for later.

It helps respect the cool-down period and give us longer to think through any potential changes to the classic settings.

Consolatis:
Just wanting to point out that we can also easily delay the release, both options are fine to me.

tokyo4j:
Opened #3127 for 0.9.2.

> Just wanting to point out that we can also easily delay the release, both options are fine to me.

Yes, let's do that if there's still room for improvement in #3127 or we're busy and it takes several days to merge. The release date is much less important than labwc's quality and our health :smile:

tokyo4j:
Rebased.

PR #3119, opened 2025-10-01
labnag: rename options for color

tokyo4j:
This PR renames `labnag` options for colors:

| Before                 | After                        |
| ---------------------- | ---------------------------- |
| `--background`         | `--background-color`         |
| `--border`             | `--button-border-color`      |
| `--border-bottom`      | `--border-bottom-color`      |
| `--button-background`  | `--button-background-color`  |
| `--text`               | `--text-color`            |
| `--button-text`        | `--button-text-color`        |
| `--details-background` | `--details-background-color` |

- Add `-color` suffix to clarify that they configures colors. `--text` looked like it configures the string to show rather than its color.
- Rename `--border` to `--button-border-color` because it only configures the button border. `--border` sounded like `labnag` has borders around the entire surface, but it doesn't.

Background: I was implementing keyboard-based button selection in `labnag` which adds new color options like `--button-selected-border-color`.


Consolatis:
No objections from my side, if we want to change the argument names we should indeed do it before the release. But lets have @johanmalm give a final approval.

johanmalm:
LGTM.

This is much better/clearer than before.

PR #3121, opened 2025-10-01
build: fix build with libinput as a subproject

tokyo4j:
For post-0.9.2.

Unfortunately, `has_header_symbol()` doesn't work with internal dependencies in `subprojects/`.

Ref: https://github.com/mesonbuild/meson/issues/13553


Consolatis:
LGTM, not tested

PR #3123, opened 2025-10-02
doc: note planned updates of default colors in thumbnail-style osd

tokyo4j:
Background: #3118

The current default colors of a selected window item in thumbnail-style window switcher don't work well with some themes like `Numix` (https://github.com/labwc/labwc/pull/3118#issuecomment-3357154192). I wrote a fix for it (https://github.com/labwc/labwc/pull/3118#issuecomment-3362417671), but it's not trivial enough for this cool-down period (or is it?). So this PR notifies users that the default colors are planned to be changed, in order to give them as little surprise as possible.


johanmalm:
I'm up for merging 3118 in cool down and think we're on the right track with that one.

Given that the thumbnail-style window-switcher default colors don't work well with dark themes (that don't have the latest window-switcher settings), I think we ought prioritize fixing that. I feel it's a good example of what the cool-down period is for.

I've run out of time tonight, but can review/test 3118 tomorrow and/or Saturday.

tokyo4j:
Alright, let's go for #3118 first then. That PR contains two commits:
1. Update the default colors of thumbnail-style window switcher by blending `osd.bg.color` and `osd.label.text.color`
2. Introduce new theme options for colors of classic-style window switcher and update its default colors

So I guess we can open a separate PR that only contains the first commit and merge it in this cool-down period?

PR #3124, opened 2025-10-04
osd: allow window switcher to temporary unshade windows

Consolatis:
This can be configured with a new unshade="yes|no" argument for windowSwitcher in rc.xml

Fixes:
- #3111

----

For post-release.
The view preview outlines for temporarily unshaded windows are only visible on the titlebar and don't cover the whole window like for normal windows. It could be changed by changing the order of outlines and preview if preferred but I kept this behavior in for now because it provides additional state to the user.

I decided against bypassing `view_set_shade()` (e.g. enabling / disabling the content scene node manually) because shaded windows have their border disabled and it would require the OSD to manipulate internal SSD state.

Testing appreciated, especially to figure out a sensible default for `<windowSwitcher unshade="yes|no" />`. Personally I don't really use shading so its hard for me to tell what feels better / more consistent.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -352,7 +352,7 @@ this is for compatibility with Openbox.
 </windowSwitcher>
 ```
 
-*<windowSwitcher show="" style="" preview="" outlines="" allWorkspaces="">*
+*<windowSwitcher show="" style="" preview="" outlines="" allWorkspaces="" unshade="">*
 	*show* [yes|no] Draw the OnScreenDisplay when switching between
 	windows. Default is yes.
 
@@ -369,6 +369,9 @@ this is for compatibility with Openbox.
 	*allWorkspaces* [yes|no] Show windows regardless of what workspace
 	they are on. Default no (that is only windows on the current workspace
 	are shown).
+
+	*unshade* [yes|no] Temporarily unshade windows while using A-Tab and
+	permanentele unshade on the final selection. Default is yes.
 
 *<windowSwitcher><fields><field content="" width="%">*
 	Define window switcher fields when using *<windowSwitcher style="classic" />*.
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -79,7 +79,8 @@
     </font>
   </theme>
 
-  <windowSwitcher show="yes" style="classic" preview="yes" outlines="yes" allWorkspaces="no">
+  <windowSwitcher show="yes" style="classic" preview="yes"
+      outlines="yes" allWorkspaces="no" unshade="yes">
     <fields>
       <field content="icon" width="5%" />
       <field content="desktop_entry_name" width="30%" />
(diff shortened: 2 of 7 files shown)


johanmalm:
Thanks.

Have tested. Seems to work well.

I think using `unshade="yes"` as the default feels good.

A couple of comments:

1. Instead of `osd_state->preview_view`, could we get the view by using `node_view_from_node()`?
2. `s/permanentele/permanently/`

Consolatis:
> A couple of comments:
> 
>     1. Instead of `osd_state->preview_view`, could we get the view by using `node_view_from_node()`?
> 
>     2. `s/permanentele/permanently/`

Changed both.

tokyo4j:
>  The view preview outlines for temporarily unshaded windows are only visible on the titlebar and don't cover the whole window like for normal windows. It could be changed by changing the order of calls for outlines and preview if preferred but I kept this behavior in for now because it provides additional state to the user.

I prefer preview outlines covering the entire window rather than only its titlebar. In the current state of this PR, when I open one window, shade it and press Alt-Tab, the preview outline covers only its titlebar. But then if I press Tab again, the preview outline is resized to cover the entire window. I find this behavior odd. I think we can reorder `preview_cycled_view()` and `osd_update_preview_outlines()` in `update_osd()` to fix this (as you described). I don't get what you meant by "it provides additional state to the user".

Consolatis:
> I don't get what you meant by "it provides additional state to the user".

It tells the user that the windows is actually shaded. I am fine with removing that though and instead show the preview outlines for the whole window.

tokyo4j:
> > I don't get what you meant by "it provides additional state to the user".
> 
> It tells the user that the windows is actually shaded. I am fine with removing that though and instead show the preview outlines for the whole window.

I see... then I don't have a strong preference here. I don't like the odd behavior I described above, but it happens only when there's one window in the desktop, where window switcher isn't useful.

I can review this PR in a day or two.

domo141:
unshare behavior work as I'd expected (also tried unshade="no" and it behaved like before)

In case of this temporary unshading is that the outline is around the title bar, not over the
whole window, like with those windows that weren't shaded when started A-tabbing.
Whether this is OK or not is a question there...

Consolatis:
Rebased + preview outlines are now always around the whole window.

docs/labwc-config.5.scd
```diff
@@ -364,6 +364,9 @@ this is for compatibility with Openbox.
 	they are on. Default no (that is only windows on the current workspace
 	are shown).
 
+	*unshade* [yes|no] Temporarily unshade windows while using A-Tab and
+	permanently unshade on the final selection. Default is yes.
```
tokyo4j: nit: I think we should avoid `A-Tab`. I suggest `when switching between windows` (or `when switching between them`) which follows the documentations of `preview`/`outlines` above.

tokyo4j:
LGTM other than the nitpick in the documentation above. Tested and it worked well.

PR #3127, opened 2025-10-07
osd-thumbnail: update default colors of selected window item

tokyo4j:
Extracted the first commit from #3118 for merging before 0.9.2 release.

## Commit message

Previously, the default values of `osd.window-switcher.style-thumbnail.item.active.{bg,border}.color` were blue. But they caused the selected window title in the window switcher to be unreadable due to duplicated colors of the text and background with Openbox themes like Numix.

Instead, this commit updates them to follow other themes configurations. The default border color of the selected window item is now `osd.label.text.color` with 50% opacity and the background is `osd.label.text.color` with 15% opacity.

For subpixel antialiasing to work, the background color is calculated by manually blending `osd.label.text.color` and `osd.bg.color`, rather than just updating the alpha with 50% or 15%.

### DIFF
--- a/docs/labwc-theme.5.scd
+++ b/docs/labwc-theme.5.scd
@@ -358,10 +358,12 @@ all are supported.
 	Border width of selected window switcher items in pixels. Default is 2.
 
 *osd.window-switcher.style-thumbnail.item.active.border.color*
-	Color of border around selected window switcher items. Default is #589bda.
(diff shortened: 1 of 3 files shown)


johanmalm:
Thanks.
LGTM. 
Have tested thumbnail window-switcher with both light and dark themes. Seems fine.

PR #3128, opened 2025-10-08
desktop-entry: fix wrong description of sfdo-icon flags

tokyo4j:


Consolatis:
As this is basically just a docs change I am fine with merging this before release. Thanks!

PR #3129, opened 2025-10-08
desktop-entry.c: on detecting a broken icon theme, fall back to hicolor

Consolatis:
Reported-By: @Kreevoz

Fixes:
- #3126

---

This allows to have an immediate fix in labwc while we can work on the libsfdo side in the background.
Suggest to merge this before release.

CC @tokyo4j @johanmalm

### DIFF
--- a/src/desktop-entry.c
+++ b/src/desktop-entry.c
@@ -37,9 +37,10 @@ log_handler(enum sfdo_log_level level, const char *fmt, va_list args, void *tag)
 
 	/*
 	 * To avoid logging issues with .desktop files as errors, all libsfdo
-	 * error-logging is demoted to info level.
+	 * error-logging is demoted to info level unless running with
+	 * LABWC_DEBUG_LIBSFDO.
 	 */
-	if (level == SFDO_LOG_LEVEL_ERROR) {
+	if (!debug_libsfdo && level == SFDO_LOG_LEVEL_ERROR) {
 		level = SFDO_LOG_LEVEL_INFO;
 	}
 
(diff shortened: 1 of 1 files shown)


tokyo4j:
Thanks. +1 for merging this before 0.9.2 release.

I suggest adding a comment like this to clarify the duplicated logic in labwc and libsfdo:

> `sfdo_icon_theme_load()` falls back to hicolor theme with _ALLOW_MISSING flag when the theme is missing, but just fails when the theme is invalid. So manually call `sfdo_icon_theme_load()` again here.

src/desktop-entry.c
```diff
@@ -108,6 +108,13 @@ desktop_entry_init(struct server *server)
 	sfdo->icon_theme = sfdo_icon_theme_load(
 		sfdo->icon_ctx,
 		rc.icon_theme_name, load_options);
+	if (!sfdo->icon_theme) {
+		wlr_log(WLR_ERROR,
+			"Failed to load icon theme %s, falling back to 'hicolor'",
+			rc.icon_theme_name);
```
vyivel, line 120: Just a suggestion: provide instructions on getting more info on the error (`-d`, `LABWC_DEBUG_LIBSFDO`) here?

tokyo4j: Just a suggestion too: I think we can also add the instruction to `wlr_log(WLR_ERROR, "Failed to initialize icon loader");` at the last line of the function.

Consolatis: Added both and also prevented demoting of error messages to info if `LABWC_DEBUG_LIBSFDO` is enabled. That way the user doesn't need `--debug`.

Consolatis:
> I suggest adding a comment like this to clarify the duplicated logic in labwc and libsfdo:
> 
> > `sfdo_icon_theme_load()` falls back to hicolor theme with _ALLOW_MISSING flag when the theme is missing, but just fails when the theme is invalid. So manually call `sfdo_icon_theme_load()` again here.

Added.

johanmalm:
Thanks.
Have tested. LGTM

PR #3130, opened 2025-10-08
build: use spaces in stead of tab

johanmalm:


Consolatis:
`s/in stead/instead/` ?
Other than that LGTM

PR #3134, opened 2025-10-09
implement is_modal_dialog() for xdg shell

### LINKED ISSUE #1698: Appearance of GTK modal dialogs not being correctly set
Labwc's handling of dialogs which are set as modal in GTK applications is not completely consistent with their behaviour under X.

The Geany editor is a good demonstration, as it has both modal and non-modal dialogs.

Non-modal dialogs, such as the search dialog, behave identically in X and in labwc - they can be opened, they stay in front of the parent but can lose focus if the parent is clicked while remaining open (and the title bar and GTK foreground/background theme is updated accordingly), and keyboard and mouse input can be directed to either the parent or the dialog as expected.

In X, a modal dialog, such as the open file dialog, completely locks out access to the parent. It is not possible to focus the parent by clicking on it, or indeed to interact with it in any way, until the modal dialog is closed. This is, as far as I understand it, the expected behaviour of a modal dialog.

In labwc, it is possible to click away from the open file dialog back to the parent - the title bar colours update to suggest the parent now has focus, and it adopts the GTK theme for a foreground window rather than a background window, while the modal dialog adopts the background GTK theme. However, keyboard and mouse input are (correctly) blocked on the parent while the modal is open, so it appears to be the indication of focus and foreground / background which is incorrect.

I believe the only fix required is to make it such that clicking on the parent when a modal dialog is open does not update the appearance (titlebar and GTK foreground / background theme) of the parent or the dialog.

xi:
Using the new [xdg-dialog](https://wayland.app/protocols/xdg-dialog-v1) protocol extension.

This should fix #1698.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -276,6 +276,7 @@ allow_for_sandbox(const struct wlr_security_context_v1_state *security_state,
 		"zxdg_importer_v1",
 		"zxdg_importer_v2",
 		"xdg_toplevel_icon_manager_v1",
+		"xdg_dialog_v1",
 		/* plus */
 		"wp_alpha_modifier_v1",
 		"wp_linux_drm_syncobj_manager_v1",
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -5,6 +5,7 @@
 #include <wlr/types/wlr_fractional_scale_v1.h>
 #include <wlr/types/wlr_scene.h>
 #include <wlr/types/wlr_xdg_activation_v1.h>
+#include <wlr/types/wlr_xdg_dialog_v1.h>
 #include <wlr/types/wlr_xdg_shell.h>
 #include <wlr/types/wlr_xdg_toplevel_icon_v1.h>
 #include "buffer.h"
@@ -614,6 +615,18 @@ xdg_toplevel_view_append_children(struct view *self, struct wl_array *children)
 	}
 }
 
+static bool
+xdg_toplevel_view_is_modal_dialog(struct view *view)
+{
+	struct wlr_xdg_toplevel *toplevel = xdg_toplevel_from_view(view);
+	struct wlr_xdg_dialog_v1 *dialog =
+		wlr_xdg_dialog_v1_try_from_wlr_xdg_toplevel(toplevel);
+	if (!dialog) {
+		return false;
+	}
+	return dialog->modal;
+}
+
 static void
 xdg_toplevel_view_set_activated(struct view *view, bool activated)
 {
@@ -894,6 +907,7 @@ static const struct view_impl xdg_toplevel_view_impl = {
 	.minimize = xdg_toplevel_view_minimize,
 	.get_root = xdg_toplevel_view_get_root,
 	.append_children = xdg_toplevel_view_append_children,
+	.is_modal_dialog = xdg_toplevel_view_is_modal_dialog,
 	.get_size_hints = xdg_toplevel_view_get_size_hints,
 	.contains_window_type = xdg_toplevel_view_contains_window_type,
 	.get_pid = xdg_view_get_pid,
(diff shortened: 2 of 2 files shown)


Consolatis:
Thanks for the PR!

The changes look good to me but I didn't test it yet.
I don't have any application which uses the xdg-dialog protocol (pretty outdated system by now) and would need to write a dummy client myself.

As this has implications on the focus behavior and likely can't be properly tested before the release I'd prefer if we look at this for the next release cycle (starting in a few days).

tokyo4j:
This PR is currently no-op. You should call `wlr_xdg_wm_dialog_v1_create()` in `xdg_init()` to support the protocol. Then you will also need to add the protocol to the list of unprivileged protocols in `server.c`.

xi:
> You should call `wlr_xdg_wm_dialog_v1_create()` in `xdg_init()` to support the protocol. Then you will also need to add the protocol to the list of unprivileged protocols in `server.c`.

Thanks for pointing that out, I had totally missed it. I am honestly a bit confused by the wlroots interface, so I hope I got it correct this time.

I had thought that I had tested this with geany, but after @Consolatis' comment I am not sure it actually uses xdg-dialog yet.

tokyo4j:
LGTM. I tested with "Dialog" in `gtk4-demo` and `featherpad` and confirmed the parent window no longer gets the keyboard focus.

Unfortunately, `geany` (mentioned in #1698) is based on GTK3 and doesn't seem to support xdg-dialog-v1 protocol. I also tested `citations` which is based on GTK4, but its file chooser is actually spawned by `xdg-desktop-portal-gtk` and it doesn't seem to support xdg-dialog-v1 either.

I wonder if xdg-foreign-v2 (which `xdg-desktop-portal-gtk` uses) and xdg-dialog-v1 can work together... I guess they can.

xi:
> I wonder if xdg-foreign-v2 (which xdg-desktop-portal-gtk uses) and xdg-dialog-v1 can work together... I guess they can.

Should we open an issue at xdg-desktop-portal-gtk about that?

Consolatis:
xdg-dialog operates on a `xdg_toplevel` whereas xdg-foreign exposes its own type on import (mapping the underlying `wl_surface` on the compositor side). So there is no `xdg_toplevel` object which could be used by the importer.

The exporter protocol defines that the `wl_surface` being exported actually has to be a `xdg_toplevel` though (in v2; `xdg_surface` in v1). So maybe the xdg-dialog protocol can be extended by another constructor like `<request name="get_xdg_dialog_from_xdg_imported_v2">`.

Or as alternative, the xdg-importer protocol could define another request like `<request name="set_modal">`.

In any case it seems some protocol work is required to make this work.

tokyo4j:
@Consolatis I think you have the same misconception as I had. xdg-dialog-v1 just allows us to mark a `xdg_toplevel` as a modal and it doesn't require the parent toplevel. So the portal can:

1. Create its own file chooser window
2. Import the handle of the parent window from another GTK app as `xdg_imported_v2`
3. Bind its own file chooser window to the parent window with [`zxdg_imported_v2::set_parent_of`](https://wayland.app/protocols/xdg-foreign-unstable-v2#zxdg_imported_v2:request:set_parent_of) request
4. Create `xdg_dialog_v2` from its own file chooser window
5. Mark its own file chooser window as a modal via [`xdg_dialog_v1::set_modal`](https://wayland.app/protocols/xdg-dialog-v1#xdg_dialog_v1:request:set_modal) request

I installed https://github.com/mahkoh/xdg-desktop-portal-gtk4 and confirmed it uses xdg-dialog-v1 protocol as expected.

> Should we open an issue at xdg-desktop-portal-gtk about that?

~The fundamental problem is xdg-desktop-portal-gtk is based on GTK3 which doesn't support xdg-dialog-v1. So I guess the issue should be titled "Port to GTK4" and xdg-dialog-v1 support should be mentioned as one of its benefits.~

Opened https://github.com/flatpak/xdg-desktop-portal-gtk/issues/532

Consolatis:
Oh I see. I got it backwards then, its actually the client surface which gets imported into the portal and not the other way round. Thanks for clearing that up!

PR #3135, opened 2025-10-09
rcxml: move <maximizedDecoration> from <core> to <theme>

tokyo4j:
For merging before 0.9.2 release.

I put it in `<core>` just because there's `<core><decoration>`, but now I think it should be in `<theme>` instead because `<theme>` contains many other configurations for SSD styles, like `<titlebar>`, `<keepBorder>` and `<dropShadows>`. `<core><decoration>` is a bit different from them as it configures which of SSD and CSD we should advise apps to use, rather than how SSD should look.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -171,7 +171,6 @@ this is for compatibility with Openbox.
 ```
 <core>
   <decoration>server</decoration>
-  <maximizedDecoration>titlebar</maximizedDecoration>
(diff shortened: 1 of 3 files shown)


tokyo4j:
JFYI: @dglava, @CosmicFusion

Consolatis:
I think it makes sense to use `<theme>` for the setting. Up to @johanmalm

johanmalm:
LGTM

PR #3141, opened 2025-10-12
Tiny correctness fixes

jlindgren90:
Some tiny fixes to make the code more technically correct:

- osd,ssd: don't cast away const
- menu: use xmlFree() for return value of xmlGetProp()

There was no functional bug here since:

- the non-const pointers were never actually used to overwrite const data
- in labwc, xmlFree() aliases to free() anyway

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -482,7 +482,7 @@ fill_item(struct menu *menu, xmlNode *node)
 		goto out;
 	}
 
-	struct menuitem *item = item_create(menu, (char *)label, icon_name, false);
+	struct menuitem *item = item_create(menu, label, icon_name, false);
 	lab_xml_expand_dotted_attributes(node);
 	append_parsed_actions(node, &item->actions);
 


src/menu/menu.c
```diff
@@ -487,8 +487,8 @@ fill_item(struct menu *menu, xmlNode *node)
 	append_parsed_actions(node, &item->actions);
 
 out:
-	free(label);
-	free(icon_name);
+	xmlFree(label);
+	xmlFree(icon_name);
```
Consolatis, line 491: They are still casted to `char *`, should we just keep them at `xmlChar *label`? No strong preference though and if we then end up having to cast them to `(char *)` in every function call its not really worth it either.

Edit:
There is a useless cast a bit above in line 485, that could be removed if we keep it at `char *`.

jlindgren90: We pass them to functions expecting `char *` (like `string_null_or_empty` or `xstrdup`) so we still need the casts, unfortunately

(edit to add: xmlChar is `unsigned char`)

Consolatis: > (edit to add: xmlChar is unsigned char)

Hm.. that is actually kind of annoying. AFAIR `char` is actually `unsigned char` on aarch64 and `signed char` on x86. Only matters for characters > 127 obviously but still.

jlindgren90: > Hm.. that is actually kind of annoying.

Yep :)

> There is a useless cast a bit above in line 485

Sure, removed it

Consolatis:
Commits LGTM, thanks!

We should likely check if the `unsigned char*` returned from libxml2 is actually problematic for characters > 127 on x86.

jlindgren90:
> We should likely check if the unsigned char* returned from libxml2 is actually problematic for characters > 127 on x86.

I'm pretty sure this is fine. labwc is certainly not the first piece of software to reinterpret between signed/unsigned 8-bit chars, and I've never heard of it being an issue. It's all twos-complement on x86.

PR #3142, opened 2025-10-12
view: store title/app_id in view

tokyo4j:
This simplifies our codes and eliminates duplicated `view.events.new_{title,app_id}` events. This should not change any behaviors.

I think we can also remove `ssd_state_title.text` in a follow-up PR.

### DIFF
--- a/src/osd/osd-field.c
+++ b/src/osd/osd-field.c
@@ -26,13 +26,9 @@ struct field_converter {
 /* Internal helpers */
 
 static const char *
-get_app_id_or_class(struct view *view, bool trim)
+get_identifier(struct view *view, bool trim)
 {
-	/*
-	 * XWayland clients return WM_CLASS for 'app_id' so we don't need a
-	 * special case for that here.
-	 */
-	const char *identifier = view_get_string_prop(view, "app_id");
+	const char *identifier = view->app_id;
 
 	/* remove the first two nodes of 'org.' strings */
 	if (trim && !strncmp(identifier, "org.", 4)) {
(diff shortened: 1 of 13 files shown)


src/view-impl-common.c
```diff
@@ -10,8 +10,6 @@ void
 view_impl_map(struct view *view)
 {
 	desktop_focus_view(view, /*raise*/ true);
-	view_update_title(view);
-	view_update_app_id(view);
```
tokyo4j, line 14: These lines are added in e140682 and e2ae63a 4 years ago. I guess they were needed to initialize app-id and titles for foreign-toplevel clients, but I think we no longer need them since `foreign_toplevel_create()` now initializes app-id/titles.

jlindgren90:
LGTM, did not test though

src/view-impl-common.c
Consolatis: On a quick look it indeed seems they are not needed anymore.

src/osd/osd-field.c
```diff
@@ -73,17 +68,11 @@ get_type(struct view *view, bool short_form)
...
 static const char *
 get_title_if_different(struct view *view)
 {
-	const char *identifier = get_app_id_or_class(view, /*trim*/ false);
-	const char *title = get_title(view);
+	const char *identifier = get_identifier(view, /*trim*/ false);
+	const char *title = view->title;
 	if (!identifier) {
 		return title;
 	}
```
Consolatis: This doesn't need to check for `NULL` anymore (and also before this PR).

src/ssd/ssd-titlebar.c
```diff
@@ -486,7 +485,7 @@ ssd_update_title(struct ssd *ssd)
 		if (state->text) {
 			free(state->text);
 		}
-		state->text = xstrdup(title);
+		state->text = xstrdup(view->title);
```
Consolatis: This could be `xstrdup_replace()`

Consolatis:
LGTM as well, two semi-related nitpicks above. Not tested either.

tokyo4j:
Thanks. Addressed the comments. Also removed the duplication checks in `wlr-foreign.c` as they are not needed anymore.

I'll merge this PR after CI passes.

PR #3144, opened 2025-10-14
theme: replace THEME_(IN)ACTIVE with SSD_(IN)ACTIVE

tokyo4j:


Consolatis:
Thanks, all active and inactive states match and there is already a `#include "ssd.h"` in `theme.c`.

PR #3148, opened 2025-10-15
Don't remove newlines when parsing config, menu and XBM

### LINKED ISSUE #3145: Some multi-line comments in XML cause parser errors
### Problem description

I use a Raspberry Pi 4 with their spin of debian trixie with labwc 0.8.4

I was trying to use #2633 and stumbled upon this Problem

my user rc.xml is:

```xml
<?xml version="1.0"?>
<!--
 -
 - Ansible managed
 -
-->
<openbox_config xmlns="http://openbox.org/3.4/rc">
  <keyboard>
    <keybind key="A-W-h">
      <action name="HideCursor"/>
    </keybind>
  </keyboard>
</openbox_config>
```

with this the `Alt` + `Super` + `h` does not do anything.

once I remove the comment so the config looks like this:

```xml
<?xml version="1.0"?>
<openbox_config xmlns="http://openbox.org/3.4/rc">
  <keyboard>
    <keybind key="A-W-h">
      <action name="HideCursor"/>
    </keybind>
  </keyboard>
</openbox_config>
```

the combination hides the Cursor successfully.

weirdly enough: commenting out actions works fine

### Steps to reproduce

1. launch labwc with a user config with a XML comment like above
2. try actions you defined in the user config (like pressing `Alt` + `Super` + `h` )

expected behavior: the comment gets detected as a comment and gets ignored by the parsing

actual behavior: the user config gets ignored

### labwc build source

Release

### labwc version

0.8.4

### labwc environment

### Distribution

RaspberryPiOS based on Debian 13

tokyo4j:
Fixes #3145.

Removing newlines in rc.xml and menu.xml caused parser error with following content:

```xml
<!--
 -
 - Some comments
 - 
-->
```

...though it is a valid XML.

Let's not do that. I moved `grab_file()` to `buf.c` and renamed it to `buf_from_file()`, because it now directly touches `struct buf` and I don't like having a source file only for one function.


src/common/buf.c
```diff
@@ -203,3 +203,33 @@ buf_move(struct buf *dst, struct buf *src)
...
+		return buf;
+	}
+	long size = ftell(stream);
+	if (size == -1) {
+		fclose(stream);
+		return buf;
+	}
+	rewind(stream);
+
+	buf_expand(&buf, size + 1);
```
Consolatis, line 230: Why the `+ 1` here?

Other than that looks fine but lets give @johanmalm a chance to review.

src/common/buf.c
```diff
@@ -203,3 +203,33 @@ buf_move(struct buf *dst, struct buf *src)
...
+	if (fread(buf.data, 1, size, stream) == (size_t)size) {
+		buf.len = size;
+	} else {
+		buf_reset(&buf);
```
Consolatis, line 236: Should we log this?

src/common/buf.c
```diff
@@ -203,3 +203,33 @@ buf_move(struct buf *dst, struct buf *src)
...
+buf_from_file(const char *filename)
+{
+	struct buf buf = BUF_INIT;
+	FILE *stream = fopen(filename, "r");
+	if (!stream) {
+		return buf;
+	}
+
+	if (fseek(stream, 0, SEEK_END) == -1) {
+		fclose(stream);
```
johanmalm, line 219: Consider wlr_log() here with `strerror(errno)`.

src/common/buf.c
tokyo4j: For a terminating null character. `buf_add()` and `buf_add_char()` also call `buf_expand()` with the length incremented by one.

src/common/buf.c
```diff
@@ -203,3 +203,33 @@ buf_move(struct buf *dst, struct buf *src)
...
+	}
+	long size = ftell(stream);
+	if (size == -1) {
+		fclose(stream);
+		return buf;
+	}
+	rewind(stream);
+
+	buf_expand(&buf, size + 1);
+	if (fread(buf.data, 1, size, stream) == (size_t)size) {
```
johanmalm, line 231: `buf_expand()` uses `xmalloc()` and `xrealloc()` so will `buf.data` definitely be null terminated here?

src/common/buf.c
johanmalm: Yeah, doing `ftell()` on a FILE generated like this `printf '%b' "123" > test.txt` returns 3, so need to allow space for null byte.

johanmalm:
I like the approach in general. Nice simplification.

src/common/buf.c
Consolatis: Oh, right. I somehow remembered that `buf_expand()` does this internally. Thanks for clarifying.

src/common/buf.c
tokyo4j: True. Added `buf.data[size] = '\0';`.

tokyo4j:
Added error loggings. I excluded `fopen()` for error logging since we should not print anything when the file doesn't exist (e.g. when `~/.config/labwc/rc.xml` doesn't exist but `/etc/xdg/labwc/rc.xml` does).

I didn't add test for now, because `buf_from_file()` is a bit more annoying to test as it deals with a file rather than a string.

johanmalm:
LGTM

PR #3149, opened 2025-10-15
tests: remove placeholders for arguments

tokyo4j:
I don't think they are needed. Let's remove it if the CI passes.


Consolatis:
> Let's remove it if the CI passes.

Heh, looks like we don't have a build trigger for the `t/` directory. See 7166efe7bf3fe239b73b394759f59835653d6d32 for reference.

tokyo4j:
Updated `build.yml`. I thought we could run `scripts/check` for test files, but it caused errors that are annoying to fix:

```
t/buf-simple.c:71: CHECK: line length of 338 exceeds 100 columns
t/buf-simple.c:71: WARNING: const array should probably be static const
t/buf-simple.c:92: CHECK: Avoid CamelCase: <CMUnitTest>
t/str.c:21: CHECK: Avoid CamelCase: <CMUnitTest>
t/xml.c:121: CHECK: Avoid CamelCase: <CMUnitTest>
```

Do you think we should skip `Avoid CamelCase` and fix other errors?

johanmalm:
> Do you think we should skip `Avoid CamelCase` and fix other errors?

Yeah, why not.

The following patch does it:

```
diff --git a/scripts/checkpatch.pl b/scripts/checkpatch.pl
index 933dbc59..217536a6 100755
--- a/scripts/checkpatch.pl
+++ b/scripts/checkpatch.pl
@@ -5661,6 +5661,7 @@ sub process {
 			    $var !~ /^(?:GString|GError|GHashTable)/ &&
 			    $var !~ /^(?:__FreeBSD__)/ &&
 			    $var !~ /^(?:RsvgRectangle|RsvgHandle)/ &&
+			    $var !~ /^(?:CMUnitTest)/ &&
 			    $var !~ /^(?:XKB_KEY_XF86Switch_VT_1)/ &&
 
 #Ignore SI style variants like nS, mV and dB
diff --git a/t/buf-simple.c b/t/buf-simple.c
index ff22aa10..cf615698 100644
--- a/t/buf-simple.c
+++ b/t/buf-simple.c
@@ -74,7 +74,12 @@ test_buf_add_char(void **state)
 {
 	(void)state;
 
-	const char long_string[] = "123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890";
+	const char const long_string[] = "123456789012345678901234567890123456 "
+		"8901234567890123456789012345678901234567890123456789012345678 "
+		"0123456789012345678901234567890123456789012345678901234567890 "
+		"2345678901234567890123456789012345678901234567890123456789012 "
+		"4567890123456789012345678901234567890123456789012345678901234 "
+		"678901234567890";
 	size_t len = strlen(long_string);
 
 	/*
```

johanmalm:
LGTM

tokyo4j:
> Yeah, why not.

A subtle disadvantage of it is that we can no longer detect when a new contributor accidentally define some symbols with camel case, which is unusual in our codebase. But yeah, I think we can catch it with reviews and it's not a realistic problem.

I'll apply your diff and merge this PR.

Consolatis:
> A subtle disadvantage of it is that we can no longer detect when a new contributor accidentally define some symbols with camel case, which is unusual in our codebase. But yeah, I think we can catch it with reviews and it's not a realistic problem.

I think we should not remove the CamelCase check but rather add a exception for this case as done in the diff by @johanmalm.

tokyo4j:
Ah, sorry. I didn't look the diff in detail. Yes, that sounds like a great idea!

Issue #3099, opened 2025-09-24
Invisible input margin to undecorated/layer shell windows


mozlima:
### Problem description

Using wlroots versions 0.19.0-2 through 0.19.0-5 causes labwc to add an invisible input margin to undecorated or layer-shell windows (e.g., Firefox) and panels (tested: sfwbar, xfce4-panel).

The screen renders correctly without visible margins, and the pointer can hover over the region, but input near the screen edges is blocked.

**Environment:**

* labwc 0.9.1-1

**Working version:**

* wlroots 0.19.0-1
* libdisplay-info 0.2.0-2 (also downgraded as a dependency)

**Affected versions:**

* wlroots 0.19.0-2
* wlroots 0.19.0-3
* wlroots 0.19.0-5
* libdisplay-info 0.3.0-1 (as a dependency)

### Steps to reproduce

When the window is decorated, input works correctly. For example, with a maximized decorated window behind and focus given to a smaller centered window, clicking on the very bottom of the maximized window correctly brings focus to it.

### labwc build source

Release

### labwc version

0.9.1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

ahesford:
Client-side decorations are responsible for these borders, not labwc.

mozlima:
> Client-side decorations are responsible for these borders, not labwc.

Are the inputs directly managed by them or proxied by the compositor/wlroots? To be honest, I don’t know if the bug comes from labwc or wlroots. I would say libdisplay-info, but it’s only a required dependency for wlroots 0.19, and wlroots 0.19 is also a dependency for labwc.

ahesford:
It's not a bug. The client adds an invisible region that permits things like window resize. You aren't clicking on the client behind, you're clicking in the topmost client's resize zone.

mozlima:
> It's not a bug. The client adds an invisible region that permits things like window resize. You aren't clicking on the client behind, you're clicking in the topmost client's resize zone.

Of course, it's a bug. I can't say whether it's related to the compositor, wlroots 0.19.0-2 to 0.19.0-5, or libdisplay-info 0.3.0-1

johanmalm:
@mozlima I don't understand the bug report.

Would you mind trying to describe it in with others words, post a screenshot highlighting the area, and/or similar.

Specifically `invisible input margin` is not clear to me.

[...]

PR #3157, opened 2025-10-19
desktop: work around client-side rounding issues at right/bottom pixel

jlindgren90:
Tested with qmpanel (similar to lxqt-panel) at 1x and 2x output scale. Does not help if scaling is done client-side, e.g. QT_SCALE_FACTOR=2.

Fixes: #2379
Fixes: #3099

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -242,6 +242,39 @@ desktop_update_top_layer_visibility(struct server *server)
 	}
 }
 
+/*
+ * Work around rounding issues in some clients (notably Qt apps) where
+ * cursor coordinates in the rightmost or bottom pixel are incorrectly
+ * rounded up, putting them outside the surface bounds. The effect is
+ * especially noticeable in right/bottom desktop panels, since driving
+ * the cursor to the edge of the screen no longer works.
+ *
+ * Under X11, such rounding issues went unnoticed since cursor positions
+ * were always integers (i.e. whole pixel boundaries) anyway. Until more
+ * clients/toolkits are fractional-pixel clean, limit surface cursor
+ * coordinates to (w - 1, h - 1) as a workaround.
+ */
+static void
+avoid_edge_rounding_issues(struct cursor_context *ctx)
+{
+	if (!ctx->surface) {
+		return;
+	}
+
+	int w = ctx->surface->current.width;
+	int h = ctx->surface->current.height;
+	/*
+	 * The cursor isn't expected to be outside the surface bounds
+	 * here, but check (sx < w, sy < h) just in case.
+	 */
+	if (ctx->sx > w - 1 && ctx->sx < w) {
+		ctx->sx = w - 1;
+	}
+	if (ctx->sy > h - 1 && ctx->sy < h) {
+		ctx->sy = h - 1;
+	}
+}
+
 /* TODO: make this less big and scary */
 struct cursor_context
 get_cursor_context(struct server *server)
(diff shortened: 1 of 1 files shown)


src/desktop.c
```diff
@@ -269,6 +269,35 @@ get_cursor_context(struct server *server)
...
+		 * Qt apps) where cursor coordinates in the rightmost or
+		 * bottom pixel are incorrectly rounded up, putting them
+		 * outside the surface bounds.
+		 *
+		 * This also avoids a similar server-side rounding issue
+		 * with some combinations of wlroots and libwayland
+		 * versions (likely to be fixed soon in wlroots). See:
+		 *
+		 * - https://gitlab.freedesktop.org/wayland/wayland/-/issues/555
+		 * - https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5159
```
tokyo4j: nit: I think this paragraph should be moved to the commit message, because "soon" will no longer be "soon" as time passes.

tokyo4j:
Though I haven't tested this, I like this approach.

But I think we should get a consensus on whether to accept this kind of workaround for client (e.g. Qt) issues. I'm for this approach. @labwc/compositor-team What do you think?

jlindgren90:
We have a number of workarounds for client-side issues in place already. Some even for specific clients (check xdg.c for references to Thunderbird).

Obviously there have to be some boundaries, but I think in the past we've leaned toward pragmatism and added workarounds where it fixes a popular use-case and doesn't have bad side effects.

This particular one is important for LXQt, and the issue is not easy to fix in Qt (would require replacing a common but incorrect usage of QRect::contains throughout the codebase).

src/desktop.c
jlindgren90: Sure, I was on the fence whether to mention it here in the comment or not.

johanmalm:
LGTM

I generally try to resist workarounds - as I'm sure we all are - but feel there is sufficient benefit to justify this one; with hopefully no long term side-effects. And we have to recognize that we live in the real world :smile:

Consider putting this in a separate function to avoid the scariness of `get_cursor_content()`. Maybe a `clamp_hack_for_qt_clients()`.

Ref:
- https://github.com/labwc/labwc/blob/261ec3c5093e89a27916210dbddc8e9e31c1d8d7/src/desktop.c#L245

Consolatis:
> But I think we should get a consensus on whether to accept this kind of workaround for client (e.g. Qt) issues.

In general fine by me as well as long as
- it impacts a broad range of applications (e.g. mostly UI frameworks)
- the severity of the issue being worked around is high enough
- it doesn't have negative side effects for other clients
- doesn't introduce a maintenance burden
- is explained clearly within the code (which is the case here)

src/desktop.c
jlindgren90: Removed this bit and mentioned it in the commit message instead.

jlindgren90:
- moved to a separate function (edge_rounding_workaround)
- expanded rationale in comments a bit
- moved mention of the wlroots/wayland issue to the commit message

src/desktop.c
```diff
@@ -242,6 +242,39 @@ desktop_update_top_layer_visibility(struct server *server)
...
+ * especially noticeable in right/bottom desktop panels, since driving
+ * the cursor to the edge of the screen no longer works.
+ *
+ * Under X11, such rounding issues went unnoticed since cursor positions
+ * were always integers (i.e. whole pixel boundaries) anyway. Until more
+ * clients/toolkits are fractional-pixel clean, limit surface cursor
+ * coordinates to (w - 1, h - 1) as a workaround.
+ */
+static void
+edge_rounding_workaround(struct cursor_context *ctx)
```
tokyo4j: nit: I generally prefer a verb for function name e.g. `apply_edge_rounding_workaround()`.

jlindgren90:
- renamed new function to `avoid_edge_rounding_issues`

johanmalm:
This is great. Thank you so much for investigating + fixing.

tokyo4j:
I think we can apply this for tablet tools as well (I guess this is relevant only for tablet mouse/lens which report relative positions?). This will be another reason to unify the duplicated logic in tablet/mouse pointer handling.

PR #3158, opened 2025-10-19
labnag: use overlay layer by default

johanmalm:
...so that the dialog is still visible when some client is using fullscreen mode.


johanmalm:
Have successfully tested with

```xml
<labwc_config>
  <keyboard>
    <default/>
    <keybind key="W-q">
      <action name="If">
        <prompt message="Quit?"/>
        <then>
          <action name="Exit"/>
        </then>
      </action>
    </keybind>
  </keyboard>
</labwc_config>
```

Consolatis:
:)

PR #3159, opened 2025-10-19
docs/: remove XML declarations in .xml files

johanmalm:
...because we don't like them anymore and libxml2 does not consider them mandatory.


tokyo4j:
LGTM.

Consolatis:
I vote for now always adding "because we don't like it anymore" at the end of every commit :)

PR #3164, opened 2025-10-22
ext-workspace: add `workspace_enter/leave` notifications and wire up `enter` on workspace switch

mbroemme:
## Summary
This PR introduces runtime `workspace_enter` / `workspace_leave` event notifications in the `ext-workspace` protocol and wires `workspace_enter` into the compositor's workspace switch path so clients (panels, pagers, taskbars) can react immediately when a workspace becomes visible on the currently bound output group. No protocol XML changes are required; this mirrors the initial-state emission semantics at bind time. Debug messages are emitted on enter/leave to make verification easy with `WLR_DEBUG`.

## Why
Many status bars/pagers rely on knowing when a workspace actually becomes visible on the user's active output group. Previously, clients only received the initial state on bind; dynamic transitions (e.g., switching to workspace 2) weren't explicitly signaled. Emitting `workspace_enter/leave` keeps client state accurate without polling.

## Implementation
The `*_enter/leave` helpers iterate group resources and workspace resources, match by client, and send the appropriate `workspace_enter/leave` to that client only. Include defensive `NULL` checks and debug logging (workspace name). On `workspaces_switch_to`, we emit `workspace_enter` for the target workspace just before setting it active. A `*_leave` hook is provided and can be used where the previous workspace actually leaves the bound group - e.g., when you finalize the old → new transition or in output/group reconfiguration paths.

### New Helpers (internal API)
```
lab_ext_workspace_group_workspace_enter(struct lab_ext_workspace_group *group, struct lab_ext_workspace *workspace)
lab_ext_workspace_group_workspace_leave(struct lab_ext_workspace_group *group, struct lab_ext_workspace *workspace)
```

These notify all clients that have both the group and the workspace bound, mirroring the initial-state emission behavior.

### Compositor Flow Changes
`workspaces_switch_to(...)` now calls `lab_ext_workspace_group_workspace_enter(server->workspaces.ext_group, target->ext_workspace);` right before marking the target ext-workspace active. This ensures clients learn that the workspace has entered the currently bound group as soon as the switch occurs.

### Protocol Impact
* No wire changes: Uses existing ext_workspace_group_handle_v1_send_workspace_enter/leave events.
* Client compatibility: Old clients are unaffected. New clients can subscribe to the events for better UX.

## Testing

1. Build & run
    * Build labwc and run with `labwc -d`.
2. Switch workspaces
    * Use your normal bindings to switch to another workspace in the same output group.
    * Expect logs like:
      * `ext-workspace: enter name='2'`
      * (If/when wired) `ext-workspace: leave name='1'`
3. Regression
    * Open/close views, move omnipresent views, and verify no change in existing behavior except the new notifications.

### DIFF
--- a/include/protocols/ext-workspace.h
+++ b/include/protocols/ext-workspace.h
@@ -106,4 +106,13 @@ void lab_ext_workspace_set_coordinates(struct lab_ext_workspace *workspace,
 
 void lab_ext_workspace_destroy(struct lab_ext_workspace *workspace);
 
+/* Notify clients that a workspace has entered/left the currently bound output group.
+ * This emits workspace_enter/leave for all clients that have both the group and the
+ * workspace bound, mirroring the initial-state emission.
+ */
+void lab_ext_workspace_group_workspace_enter(struct lab_ext_workspace_group *group,
+	struct lab_ext_workspace *workspace);
+void lab_ext_workspace_group_workspace_leave(struct lab_ext_workspace_group *group,
+	struct lab_ext_workspace *workspace);
+
 #endif /* LABWC_PROTOCOLS_EXT_WORKSPACES_H */
--- a/src/protocols/ext-workspace/ext-workspace.c
+++ b/src/protocols/ext-workspace/ext-workspace.c
@@ -760,3 +760,56 @@ lab_ext_workspace_destroy(struct lab_ext_workspace *workspace)
 	zfree(workspace->name);
 	free(workspace);
 }
+
+void
+lab_ext_workspace_group_workspace_enter(struct lab_ext_workspace_group *group,
+	struct lab_ext_workspace *workspace)
+{
+	if (!group || !workspace) {
+		return;
+	}
+
+	/* Verbose logging for debugging: print workspace name on enter */
+	wlr_log(WLR_DEBUG, "ext-workspace: enter name='%s'", workspace->name);
+
+	/* For each manager/group resource pair belonging to the same client,
+	 * emit workspace_enter with the matching workspace resource.
+	 */
+	struct wl_resource *group_resource;
+	wl_resource_for_each(group_resource, &group->resources) {
+		struct wl_client *client = wl_resource_get_client(group_resource);
+		struct wl_resource *ws_resource;
+		wl_resource_for_each(ws_resource, &workspace->resources) {
+			if (wl_resource_get_client(ws_resource) == client) {
+				ext_workspace_group_handle_v1_send_workspace_enter(
+					group_resource, ws_resource);
+			}
+		}
+	}
+	ext_manager_schedule_done_event(group->manager);
+}
+
+void
+lab_ext_workspace_group_workspace_leave(struct lab_ext_workspace_group *group,
+	struct lab_ext_workspace *workspace)
+{
+	if (!group || !workspace) {
+		return;
+	}
+
+	/* Verbose logging for debugging: print workspace name on leave */
+	wlr_log(WLR_DEBUG, "ext-workspace: leave name='%s'", workspace->name);
+
+	struct wl_resource *group_resource;
+	wl_resource_for_each(group_resource, &group->resources) {
+		struct wl_client *client = wl_resource_get_client(group_resource);
+		struct wl_resource *ws_resource;
+		wl_resource_for_each(ws_resource, &workspace->resources) {
+			if (wl_resource_get_client(ws_resource) == client) {
+				ext_workspace_group_handle_v1_send_workspace_leave(
+					group_resource, ws_resource);
+			}
+		}
+	}
+	ext_manager_schedule_done_event(group->manager);
+}
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -237,6 +237,12 @@ add_workspace(struct server *server, const char *name)
 	lab_ext_workspace_set_name(workspace->ext_workspace, name);
 	lab_ext_workspace_set_active(workspace->ext_workspace, active);
 
+	/* Notify ext-workspace listeners for active workspace only */
+	if (active) {
+		lab_ext_workspace_group_workspace_enter(
+			server->workspaces.ext_group, workspace->ext_workspace);
+	}
+
 	workspace->on_ext.activate.notify = handle_ext_workspace_activate;
 	wl_signal_add(&workspace->ext_workspace->events.activate,
 		&workspace->on_ext.activate);
@@ -427,6 +433,10 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	lab_ext_workspace_set_active(
 		server->workspaces.current->ext_workspace, false);
 
+	/* Notify ext-workspace listeners */
+	lab_ext_workspace_group_workspace_leave(
+		server->workspaces.ext_group, server->workspaces.current->ext_workspace);
+
 	/* Move Omnipresent views to new workspace */
 	struct view *view;
 	enum lab_view_criteria criteria =
@@ -482,6 +492,11 @@ workspaces_switch_to(struct workspace *target, bool update_focus)
 	desktop_update_top_layer_visibility(server);
 
 	lab_cosmic_workspace_set_active(target->cosmic_workspace, true);
+
+	/* Notify ext-workspace listeners */
+	lab_ext_workspace_group_workspace_enter(
+		server->workspaces.ext_group, target->ext_workspace);
+
 	lab_ext_workspace_set_active(target->ext_workspace, true);
 }
 


Consolatis:
Thanks for the PR.

Currently on the go, will look into this in more detail later.

Can you elaborate a bit more of why (in the labwc case) we need to send additional `workspace_enter` / `_leave` events (there is only ever a single group because it spans all outputs). And what is the difference to `lab_ext_workspace_assign_to_group()`?

mbroemme:
@Consolatis Thanks for review.

Use case: labwc with 8 workspaces on a single output and a bar (e.g., Waybar) using the `wlr/taskbar` module. Today, Waybar can't show only the tasks from the currently active workspace because there's no event to tell the bar that the user switched workspaces, so the taskbar can't update (show/hide) accordingly. This patch doesn't fix Waybar by itself, but once merged it makes adding that behavior to Waybar straightforward - and I'm already working on it. Since these are standard `ext-workspace` events, any app that implements the protocol can react to them. Other compositors/bars handle this via IPC (e.g., Sway) or custom protocol extensions (AFAIK Cosmic), but event-based signaling is more compositor-agnostic. It also works when a workspace spans two outputs, because the event signals workspace enter/leave rather than output enter/leave.

I'm pretty new to this - after 30+ years on X11 I'm transitioning to Wayland. If there's a better way to handle the use case above using the events labwc already exposes, I'd love to hear it.

johanmalm:
Hi @mbroemme and thanks for this contribution. The code looks very tidy, but please could you help us understand a bit more about the context.

> Many status bars/pagers rely on knowing when a workspace actually becomes visible on the user's active 

I would like to re-produce the issue that this PR fixes? Would you mind:

1. Pointing us to a bar/pager to try this with
2. Describe the steps to take to highlight the current shortcoming

Consolatis:
There is a protocol missing to match workspace handles with wlr-foreign-toplevel handles. As wlr-foreign-toplevel is a wlroots protocol an event to receive a wlr-foreign-toplevel handle will not be accepted in the upstream ext-workspace protocol. And all wlr- protocols are not accepting changes and instead expect the protocols to be upstreamed. There is a [prototype](https://github.com/labwc/labwc/compare/master...Consolatis:labwc:wip/ext_workspsaces_foreign_interop) I made a while ago but it doesn't really help without having ext-foreign-toplevel-state accepted. And we can still not reference a wlr-foreign-toplevel-handle with the ext-foreign-toplevel-handle. This whole thing is kind of a [mess](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196) on the governance side.

It is a very interesting idea to use the group membership of workspaces as a workaround. But it would cause a precedent which might bite us later once we actually have some interaction protocol. For example with this setup, how does a panel even know all the workspaces it \*could\* activate within a group?

mbroemme:
@johanmalm The PR doesn't fix anything by itself; it introduces a baseline capability that bars relying on protocol signals can use. Here's the Waybar issue I'm aiming to address:

1. Install Waybar.
2. Add this to `~/.config/waybar/config.jsonc`:
```jsonc
<...rest of your config...>
"modules-center": [
  "wlr/taskbar"
],

"wlr/taskbar": {
  "format": "{title}",
  "tooltip": false,
  "expand": true,
  "all-outputs": true,
  "on-click": "activate"
},
<...rest of your config...>
```
3. Launch Waybar with `waybar`.
4. Open 3 apps on workspace 1 and 2 apps on workspace 2.

The bar shows all toplevels at once. Switch between workspaces 1, 2, and 3 - you'll still see all 5 apps in the bar. Filtering per workspace (e.g., only 3 apps on workspace 1, only 2 on workspace 2, and an empty bar on workspace 3) is difficult without workspace enter/leave events because there's no reliable signal that a workspace switch occurred - at least with the current labwc events.

As mentioned earlier, the labwc PR won't change Waybar immediately, but it enables Waybar (and other bars) to implement this behavior.

@Consolatis Exactly - that's the problem I hit. I couldn't find a clean, reliable way to combine the existing protocols to implement per-workspace toplevel filtering. :) If a future interaction protocol lands, we can deprecate this approach, but I don't expect that soon. This PR doesn't touch any workspace-grouping code, though a bit of extra testing would still be wise.

Consolatis:
It would break all panels which only show groups (and their workspace) that are visible on the individual screen though. Other compositors do implement multiple workspace groups and thus panels need to support that scenario. So if we were to remove all workspaces from our group which are not active, those workspaces would not show up on those panels at all.

Edit:
I've now looked at this in more detail, how does this allow waybar to hide toplevels from other workspaces? There is already a workspace change event in the form of the `active` state event which panels usually use to highlight the selected workspace.

mbroemme:
@Consolatis Wouldn't that mean on every workspace switch I'd have to iterate over all toplevels and hide them in the bar, since there's no signal telling me which workspace was deactivated?

Edit:
Ok there is `EXT_WORKSPACE_HANDLE_V1_STATE_HIDDEN` but it is for a different case.

Consolatis:
> @Consolatis Wouldn't that mean on every workspace switch I'd have to iterate over all toplevels and hide them in the bar, since there's no signal telling me which workspace was deactivated?

The workspace being deactivated gets a `state` event with the `EXT_WORKSPACE_HANDLE_V1_STATE_ACTIVE` state bit not being set. The workspace being activated gets a `state` event with the `EXT_WORKSPACE_HANDLE_V1_STATE_ACTIVE` state bit set. Note that on other compositors (tag ones a la awesomewm) there can also be multiple active "workspaces" at the same time within a single group.

mbroemme:
@Consolatis You're right. I can do the same with an active state handler in Waybar. I've implemented it quickly and the handler works.

Waybar startup with 8 labwc workspaces:
```
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16cd00, state=1
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=0
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d650, state=0
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16db70, state=0
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16e090, state=0
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16e5b0, state=0
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16ebe0, state=0
[2025-10-23 21:53:12.414] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16f100, state=0
```

Workspace switch:
```
[2025-10-23 21:53:24.397] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16cd00, state=0
[2025-10-23 21:53:24.397] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=1
[2025-10-23 21:53:32.744] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=0
[2025-10-23 21:53:32.744] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d650, state=1
[2025-10-23 21:53:33.233] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=1
[2025-10-23 21:53:33.233] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d650, state=0
[2025-10-23 21:54:18.736] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=0
[2025-10-23 21:54:18.736] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d650, state=1
[2025-10-23 21:54:19.643] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=1
[2025-10-23 21:54:19.643] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d650, state=0
[2025-10-23 21:54:20.041] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16cd00, state=1
[2025-10-23 21:54:20.041] [info] [wlr/taskbar]: workspace state changed: handle=0x5585db16d130, state=0
```

There are still some glitches, sometimes the state handler doesn't send the event notifications and you need to restart labwc (really don't know why), but in general it can be used to catch workspace switch signals.

Consolatis:
I am still not entirely sure what your end goal is. Waybar should already support the ext-workspace protocol (including indicator of the active workspaces) via [ext/workspaces](https://raw.githubusercontent.com/Alexays/Waybar/refs/heads/master/man/waybar-ext-workspaces.5.scd). You seem to modify `wlr/taskbar` which is for the wlr-foreign-toplevel protocol and only handles windows. As said above there is currently no official wayland protocol to allow connecting those two. I am still missing the greater picture here.

mbroemme:
@Consolatis The goal is to show the window's taskbar entry only on the active workspace of the output / output groups. But you're right using `ext/workspaces` is the wrong way. I'm new to this. I will create another labwc PR for review a bit later which add enter/leave event with wlr-foreign-toplevel and this works without any patching Waybar and others. It will extend <windowRule> with a new property and will allow to configure a property taskbar per workspace.

Since the workspace enter/leave can also be used via active state handler, you may close this PR.

Consolatis:
As labwc only has a single workspace group which is always on all outputs this PR is not necessary.

Issue #2467, opened 2024-12-30
Menus opened via xfce4-popup-commands don't get keyborad focus on opening


Tamaranch:
### Problem description

When running e.g. xfce4-popup-applicationsmenu (shipped with xfce4-panel), the xfce app menus opens, but the menu doesn't get focus. It does get focus with wayfire 0.9.0.

Keyboard focus is set to on-demand in xfce4-panel.

### Steps to reproduce

1. Rune xfce4-panel
2. Add xfce4-applicationsmenu-plugin
3. Run xfce4-popup-applicationsmenu
4. Try to navigate in the menu via the arrow keys: it does not work, because the menu doesn't get focus.

### labwc build source

Release

### labwc version

0.8.2 or 0.8.2-29-g1550eb9

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

johanmalm:
Similar to https://github.com/labwc/labwc/issues/1572

I have tried playing around with this before but:
1. I couldn't find a way of grabbing the keyboard when popups open. Welcome help with this.
2. IIRC the protocol states the child surface should inherit the interactivity of their parent. I would personally be happy to consider popups special in this case and only apply that rule to sub-surfaces.

Consolatis:
There is a [xdg_popup.grab()](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L1253) request which might be related or not, not sure about a layershell parent + xdg_popup combination.

johanmalm:
> There is a [xdg_popup.grab()](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L1253) request which might be related or not, not sure about a layershell parent + xdg_popup combination.

Yes. I played with that. IIRC it caused focus problems on some panels’ tooltips.


> The current workaround approach seems to be to request `EXCLUSIVE`, open the popup, return back to `ON_DEMAND`.

Yes

tokyo4j:
@vyivel Does wlroots provide a way to listen for `xdg_popup.grab` requests? I found `wlr_seat.events.keyboard_grab_begin`, but I couldn't manage to tell whether the event comes from `xdg_popup.grab` requests or DnD.

vyivel:
[...]

PR #3165, opened 2025-10-23
Give keyboard focus to xdg-popups of unfocused layer-shell clients

johanmalm:
...in support of enabling panel menus to be opened by keyboard shortcuts and get keyboard focus so that they can be operated with the keyboard.

An example use-case is the xfce4-panel applications-menu being opened by the command xfce4-popup-applicationmenu.

Ref: #1572 

-- EDIT -- Needs extensive testing with different setups.

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -403,11 +403,18 @@ handle_map(struct wl_listener *listener, void *data)
 	layer_try_set_focus(seat, layer->scene_layer_surface->layer_surface);
 }
 
+static bool
+surface_is_focused(struct seat *seat, struct wlr_surface *surface)
+{
+	return seat->seat->keyboard_state.focused_surface == surface;
+}
+
 static void
 handle_popup_destroy(struct wl_listener *listener, void *data)
 {
 	struct lab_layer_popup *popup =
 		wl_container_of(listener, popup, destroy);
+	struct seat *seat = &popup->server->seat;
 
 	struct wlr_xdg_popup *_popup, *tmp;
 	wl_list_for_each_safe(_popup, tmp, &popup->wlr_popup->base->popups, link) {
@@ -423,12 +430,13 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&popup->commit.link);
 	}
 
-	/* Give focus back to whoever had it before the popup */
-	if (popup->parent_was_focused && popup->wlr_popup->parent) {
-		struct seat *seat = &popup->server->seat;
-		seat_force_focus_surface(seat, popup->wlr_popup->parent);
-	} else {
-		desktop_focus_topmost_view(popup->server);
+	if (surface_is_focused(seat, popup->wlr_popup->base->surface)) {
+		/* Give focus back to whoever had it before the popup */
+		if (popup->parent_was_focused && popup->wlr_popup->parent) {
+			seat_force_focus_surface(seat, popup->wlr_popup->parent);
+		} else {
+			desktop_focus_topmost_view(popup->server);
+		}
 	}
 
 	free(popup);
@@ -496,14 +504,12 @@ handle_popup_commit(struct wl_listener *listener, void *data)
 		/* Force focus when popup was triggered by IPC */
 		struct server *server = popup->server;
 		struct seat *seat = &server->seat;
-		if (seat->seat->keyboard_state.focused_surface
-				== popup->wlr_popup->parent) {
-			popup->parent_was_focused = true;
-		}
-		if (popup->wlr_popup->seat) {
-			struct wlr_layer_surface_v1 *layer_surface =
-				popup->lab_layer_surface->layer_surface;
-			seat_force_focus_surface(seat, layer_surface->surface);
+		bool requesting_grab = !!popup->wlr_popup->seat;
+		if (requesting_grab) {
+			if (surface_is_focused(seat, popup->wlr_popup->parent)) {
+				popup->parent_was_focused = true;
+			}
+			seat_force_focus_surface(seat, popup->wlr_popup->base->surface);
 		}
 	}
 }
@@ -573,8 +579,6 @@ handle_popup_new_popup(struct wl_listener *listener, void *data)
 
 	new_popup->output_toplevel_sx_box =
 		lab_layer_popup->output_toplevel_sx_box;
-
-	new_popup->lab_layer_surface = lab_layer_popup->lab_layer_surface;
 }
 
 /*
@@ -640,7 +644,6 @@ handle_new_popup(struct wl_listener *listener, void *data)
 	}
 
 	popup->output_toplevel_sx_box = output_toplevel_sx_box;
-	popup->lab_layer_surface = toplevel;
 
 	if (surface->layer_surface->current.layer
 			<= ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM) {
--- a/include/layers.h
+++ b/include/layers.h
@@ -30,7 +30,6 @@ struct lab_layer_popup {
 	struct wlr_xdg_popup *wlr_popup;
 	struct wlr_scene_tree *scene_tree;
 	struct server *server;
-	struct lab_layer_surface *lab_layer_surface;
 	bool parent_was_focused;
 
 	/* To simplify moving popup nodes from the bottom to the top layer */


johanmalm:
Need to deal with
1. Panel tooltips
2. Giving focus back to client that had it before menu opened

stefonarch:
Gave this a test, it looks fine for the menu. We ship now in git `lxqt-qdbus openmenu` (Alt-F1) which opens/closes the fancy menu on lxqt-panel. With this patch keyboard focus to the menu is fixed.

>Giving focus back to client that had it before menu opened

If you mean keyboard focus no, but the window is always marked active. I don't think this is an issue though, because usually users open the menu to open an app.

Panel tooltip issue:  I see (if focus is elsewhere) first it flickers shortly up and vanishes, wile the second attempt is fine,

elviosak:
> 2\. Giving focus back to client that had it before menu opened

``` c
diff --git a/src/layers.c b/src/layers.c
index 8ee11d64..19b86d82 100644
--- a/src/layers.c
+++ b/src/layers.c
@@ -423,7 +423,7 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&popup->commit.link);
 	}
 
-	cursor_update_focus(popup->server);
+	desktop_focus_topmost_view(popup->server);
 
 	free(popup);
 }
```

I don't know if this could cause other issues, but i tried some things and this fixed the focus issue for me in lxqt-panel with fancymenu and wingmenu (custom lxqt menu), patch is on top of this PR.

Tried `try_to_focus_next_layer_or_toplevel` but didn't work

elviosak:
``` c
diff --git a/src/layers.c b/src/layers.c
index 8ee11d64..eb80eb5d 100644
--- a/src/layers.c
+++ b/src/layers.c
@@ -552,8 +552,10 @@ handle_new_popup(struct wl_listener *listener, void *data)
 	 * of a use-case is the xfce4-panel start menu which can be opened by a
 	 * keyboard shortcut (linked to xfce4-popup-applicationmenu).
 	 */
-	seat_force_focus_surface(&server->seat,
-		toplevel->layer_surface->surface);
+	if (wlr_popup->seat) {
+		seat_force_focus_surface(&server->seat,
+			toplevel->layer_surface->surface);
+	}
 
 	int lx, ly;
 	wlr_scene_node_coords(&surface->tree->node, &lx, &ly);
```

And this fixed the tooltips for me

stefonarch:
Confirming both, nice!

tokyo4j:
I think the more correct approach without a hack like this is to listen to `xdg_popup.grab` requests (for both xdg-shell clients and layer-shell clients), which is currently impossible due to wlroots' limitation (https://github.com/labwc/labwc/issues/2467#issuecomment-2585927886) which originates from the fundamental flaw in wlroots' seat grab design [as explained by @vyivel](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3689). Removing seat grab from wlroots will take a lot of work, but I think wlroots can just add `wlr_xdg_popup.events.grab` for now so that we can at least give the keyboard focus to the popup following `xdg_popup.grab` request.

johanmalm:
> ```c
> diff --git a/src/layers.c b/src/layers.c
> index 8ee11d64..eb80eb5d 100644
> --- a/src/layers.c
> +++ b/src/layers.c
> @@ -552,8 +552,10 @@ handle_new_popup(struct wl_listener *listener, void *data)
>  	 * of a use-case is the xfce4-panel start menu which can be opened by a
>  	 * keyboard shortcut (linked to xfce4-popup-applicationmenu).
>  	 */
> -	seat_force_focus_surface(&server->seat,
> -		toplevel->layer_surface->surface);
> +	if (wlr_popup->seat) {
> +		seat_force_focus_surface(&server->seat,
> +			toplevel->layer_surface->surface);
> +	}
>  
>  	int lx, ly;
>  	wlr_scene_node_coords(&surface->tree->node, &lx, &ly);
> ```
> 
> And this fixed the tooltips for me

When wrapping this on `if (wlr_popup->seat) { }`, `xfce4-popup-applicationsmenu` does not work. Hmm... how frustrating.

johanmalm:
I've added a `grab` event to the `wlroots` `wlr_xdg_popup.events` and have added a handler on the `labwc` side, but.... it just results in protocol error for now. :face_with_diagonal_mouth: Will try to free up some time at the weekend to see if I can get it to work.

Cc: @tsujan

johanmalm:
I've got it working. Have opened a merge request.

https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5265

johanmalm:
Superseded by #3375

johanmalm:
@tokyo4j How about making this optional?

> I think the more correct approach without a hack like this is to listen to `xdg_popup.grab` requests (for both xdg-shell clients and layer-shell clients), which is currently impossible due to wlroots' limitation ([#2467 (comment)](https://github.com/labwc/labwc/issues/2467#issuecomment-2585927886)) which originates from the fundamental flaw in wlroots' seat grab design [as explained by @vyivel](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3689). Removing seat grab from wlroots will take a lot of work, but I think wlroots can just add `wlr_xdg_popup.events.grab` for now so that we can at least give the keyboard focus to the popup following `xdg_popup.grab` request.

Having tried to listen to `xdg_popup.grab` events (#3375) it appears that `labwc` does not receive grab events from Qt Layer Shell clients. I haven't been able to prove why this is, but suspect it is for the following reasons:

1. Per the protocol, the grab must be requested in response to a user input event and requires an event serial (so IPC triggered popups are questionable)
2. Also, arguably the popup parent should be a normal xdg_toplevel or xdg_popup, not a layer surface. I say this because I see the grab then the parent is a popup, but not when it's a layer-shell surface.

I have tried `hyprland` and `wayfire`. The panel IPC does in fact work well on those. I find their code hard to read, but at least with wayfire I think they just focus on new layer-shell-popup.

Ref:
- https://wayland.app/protocols/xdg-shell#xdg_popup:request:grab
- https://wayland-book.com/xdg-shell-in-depth/popups.html#popup-grabs

Consolatis:
> Having tried to listen to `xdg_popup.grab` events (#3375) it appears that `labwc` does not receive grab events from Qt Layer Shell clients.

This feels like the actual issue. Does WAYLAND_DEBUG=1 reveal that the grab request is actually sent? From a broader perspective, I don't really understand what is going with layershell popups + grab request but I also didn't really look that deep into it. From memory wlroots handles grabs internally and re-routes input to a grabbing surface as long as the grab is active, regardless of what other surface the compositor wants to have focus. Even without the new signal.

johanmalm:
> This feels like the actual issue. Does WAYLAND_DEBUG=1 reveal that the grab request is actually sent? From a broader perspective, I don't really understand what is going with layershell popups + grab request but I also didn't really look that deep into it. From memory wlroots handles grabs internally and re-routes input to a grabbing surface as long as the grab is active, regardless of what other surface the compositor wants to have focus. Even without the new signal.

```
~$ WAYLAND_DEBUG=1 lxqt-panel 2>&1 | grep grab
[3545745.174] {Default Queue}  -> xdg_popup#52.grab(wl_seat#11, 1487)
~$ WAYLAND_DEBUG=1 xfce4-panel 2>&1 | grep grab
[3583756.150] {Default Queue}  -> xdg_popup#51.grab(wl_seat#23, 1698)
```

^^^ when clicking on start-menu button rather than using IPC.

tokyo4j:
> Having tried to listen to xdg_popup.grab events (https://github.com/labwc/labwc/pull/3375) it appears that labwc does not receive grab events from Qt Layer Shell clients.

I reproduced this by clicking fancy menu icon in `lxqt-panel` (`lxqt-qdbus openmenu` doesn't work for some reason as said in IRC). The problem was that `lxqt-panel` sends `xdg_popup.grab` before `zwlr_layer_surface_v1.get_popup`. Therefore if we register a popup grab handler in the handler of `zwlr_layer_surface_v1.get_popup`, `xdg_popup.grab` has been already fired at that time and we cannot handle it.

The protocol says:

> ### zwlr_layer_surface_v1::get_popup
> This assigns an xdg_popup's parent to this layer_surface. This popup should have been created via xdg_surface::get_popup with the parent set to NULL, and **this request must be invoked before committing the popup's initial state.**

~If the popup grab is considered one of the "initial state", I think this is something to be fixed in lxqt-panel or layer-shell-qt.~

EDIT: `lxqt-panel` sends `xdg_popup.grab`, `zwlr_layer_surface_v1.get_popup` and `wl_surface.commit`, which doesn't seem to violate the protocol as `xdg_popup.grab` is sent before `wl_surface.commit`.

> How about making this optional?

Well, I'm not keen on it as I don't generally like to implement hacks to certain applications to work. Layer shell clients explicitly requesting keyboard is at least supported at the protocol level via `xdg_popup.grab` and I prefer making use of that.

However, I'm not strongly against *optionally* giving keyboard focus to xdg-popups, since it's unlikely that your `wlr_xdg_popup.events.grab` MR in wlroots will be merged in 0.20.0. If we do that, I think we should clearly say that the option is just a temporary workaround and we will remove it as soon as we start handling `xdg_popup.grab` requests. Something like this:


> ### `<core><alwaysFocusLayerPopup> [yes|no]`
> 
> Whether to give keyboard focus to popups created on top of a layer-shell surfaces, regardless of whether popup requests an explicit grab. Default is no.
> 
> NOTE: This is a temporary workaround for panels with keyboard-controllable menus and will be removed in labwc 0.11. Currently, labwc gives keyboard focus only to layer-shell surfaces but not to their popups. In labwc 0.11, we plan to give keyboard focus to those popups if they sends `xdg_popup.grab` request.

johanmalm:
Ah... interestingly - for both lxqt-panel and xfce4-panel `popup->wlr_popup->seat` is set by the time we get to the popups initial commit... when a grab is wanted... but not for tooltips!

That doesn't explain why the lxqt-panel grab doesn't trickle through to the new popup-grab request handler, but still.

johanmalm:
Right, this seems to actually work :smile:

The menus of both xfce4-panel and lxqt-panel get keyboard focus when their respective IPC are used to open the menus.
Tooltips do not steal keyboard focus.

@spl237 @stefonarch @tsujan @tamaranch - testing appreciated.

No changes to wlroots required for this PR.

I think Wayfire does something similar to this in its map-handler, although find it difficult to follow its logic.
Ref: https://github.com/WayfireWM/wayfire/blob/209ab88c000cc5d1ad076caa0afcc372be51caeb/src/view/xdg-shell.cpp#L230-L233

johanmalm:
It looks like:
1. In `types/xdg_shell/wlr_xdg_popup.c`... `xdg_popup_handle_grab()` sets `popup->seat`.
2. The same function also does a `wl_signal_emit_mutable()` but for layer-shell-qt clients this all takes place before the lab_layer_popup is created.

So, it appears that `popup->seat` can later be used to check if a grab was requested.

tsujan:
I installed my package of lxqt-panel which I'd compiled with `KeyboardInteractivityNone` for labwc. After applying the patch to labwc, Fancy Menu correctly gets the focus when opened by a shortcut or by clicking its button on lxqt-panel, *apparently* without the focus being taken away from the currently active window. Tooltips cause no problem anymore (→ https://github.com/labwc/labwc/issues/1572#issuecomment-3438684241)

I said "apparently" because the title-bar shows that the active window remains active. But there's still a problem: when Fancy Menu is closed (e.g., by pressing Escape), the focus does *not* return to the active window. I tested with some Qt windows, Firefox, Chromium and gtk3/4-widget-factory.

I compared it to lxqt-runner. In the case of lxqt-runner, the focus really returns to the active window.

stefonarch:
Just tested with git panel without `KeyboardInteractivityNone` and it's the same: titlebar remains active, keyboard focus goes to menu (after having made at least one interaction before with the panel because of that Qt bug) but doesn't turn to the "active" window.

And ran into https://github.com/labwc/labwc/issues/3381 too, qterminal dropdown is fine though.

Tamaranch:
This does give focus to the menu when running xfce4-popup-applicationsmenu for me (keyboard interactivity on demand), but the terminal does not regain focus after closing the menu with the Esc key, whereas this is the case with wayfire (as expected, I think).

johanmalm:
@stefonarch @tsujan @Tamaranch - Thanks for testing. I got so excited at finding a way that worked for both Gtk + Qt that I forgot about giving focus back to the last toplevel on popup close. Have now added that.

@tokyo4j - Have added a lengthy comment to the first commit - partly based on some of your feedback/comments.

@tokyo4j @Consolatis - would we be happy to merge this in cool-down?

tsujan:
> I got so excited at finding a way …

Thank you all for making this nice compositor constantly better!

Tamaranch:
> I forgot about giving focus back to the last toplevel on popup close. Have now added that.

I can confirm it works, thanks!

tsujan:
Oh, I forgot to say that I also confirm that it works with lxqt-panel + `KeyboardInteractivityNone` :)

johanmalm:
Thanks for quick responses. I have tested it myself with both `xfce4-panel` and `lxqt-panel`, so - whilst it's good to have independent verification - the other things that's really helpful is trying to make sure there aren't any unintended side-effects on your setups. I don't think there will be, but it's always hard to think of everything.

src/layers.c
```diff
@@ -447,11 +423,75 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&popup->commit.link);
 	}
 
-	cursor_update_focus(popup->server);
+	/* Give focus back to the toplevel that had it before the popup */
+	desktop_focus_topmost_view(popup->server);
```
Consolatis: Does this handle submenus? We might need to check parents here and only if we are the topmost one give focus back to the topmost view.

src/seat.c
```diff
@@ -759,6 +756,17 @@ seat_reconfigure(struct server *server)
 	}
 }
 
+void
+seat_force_focus_surface(struct seat *seat, struct wlr_surface *surface)
+{
```
Consolatis, line 761: I think we should check for an active session lock here.
CC @jlindgren90

Consolatis:
> would we be happy to merge this in cool-down?

From my side: yes, absolutely.

src/layers.c
johanmalm: I have tested with nested menus on both `qmpanel` and `xfce4-panel`. It works fine with both.
However, I know what you mean here. Not sure what's best.

johanmalm: Maybe it's because the popup-grabs are in place that it work :smile:

src/layers.c
```diff
@@ -447,11 +423,75 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
...
+ * - https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5265
+ * - https://github.com/labwc/labwc/issues/2467#issuecomment-2585927886
+ * - https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3689
+ * - https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4950
+ */
+static void
+focus(struct seat *seat, struct wlr_surface *wlr_surface)
+{
+	seat_force_focus_surface(seat, wlr_surface);
+}
```
tokyo4j: I don't think we need another function for just calling `seat_force_focus_surface()`.

src/layers.c
```diff
@@ -447,11 +423,75 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
...
+ *
+ *     This assigns an xdg_popup's parent to this layer_surface. This popup
+ *     should have been created via xdg_surface::get_popup with the parent set
+ *     to NULL, and this request must be invoked before committing the popup's
+ *     initial state.
+ *
+ * lxqt-panel sends xdg_popup.grab, zwlr_layer_surface_v1.get_popup and then
+ * wl_surface.commit, which does not seem to violate the protocol as
+ * xdg_popup.grab is sent before wl_surface.commit. So therefore, we have just
+ * have to deal with it.
```
tokyo4j: This sounds off-topic. Let's remove it.

src/seat.c
jlindgren90: Yes, I think so too. And we should test with "sleep" to make sure popping up a menu (or similar) cannot take focus when locked.

src/layers.c
tokyo4j: I'm not sure always focusing on the topmost view is the right decision... Isn't it possible that the panel itself was focused before the popup grabs focus and the panel expects to regain the focus when the popup is dismissed?

tokyo4j: I found a real example; in the current PR, closing a right-click menu of `qterminal -d` closes the terminal rather than restoring the focus to the terminal.

So it turned out that giving focus to popups is harder than I initially thought... Perhaps we should go for #3375 instead, though it implements something totally undocumented in the protocol? Anyway it needs more thoughts...

tokyo4j: How about something like this?

```diff
diff --git a/include/layers.h b/include/layers.h
index 6cb505c1..5e39868c 100644
--- a/include/layers.h
+++ b/include/layers.h
@@ -31,6 +31,7 @@ struct lab_layer_popup {
 	struct wlr_scene_tree *scene_tree;
 	struct server *server;
 	struct lab_layer_surface *lab_layer_surface;
+	bool parent_was_focused;
 
 	/* To simplify moving popup nodes from the bottom to the top layer */
 	struct wlr_box output_toplevel_sx_box;
diff --git a/src/layers.c b/src/layers.c
index 6c196aac..5222eaa2 100644
--- a/src/layers.c
+++ b/src/layers.c
@@ -424,7 +424,12 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
 	}
 
 	/* Give focus back to the toplevel that had it before the popup */
-	desktop_focus_topmost_view(popup->server);
+	if (popup->parent_was_focused && popup->wlr_popup->parent) {
+		struct seat *seat = &popup->server->seat;
+		seat_force_focus_surface(seat, popup->wlr_popup->parent);
+	} else {
+		desktop_focus_topmost_view(popup->server);
+	}
 
 	free(popup);
 }
@@ -507,6 +512,10 @@ handle_popup_commit(struct wl_listener *listener, void *data)
 		popup->commit.notify = NULL;
 
 		if (popup->wlr_popup->seat) {
+			if (popup->server->seat.seat->keyboard_state.focused_surface
+					== popup->wlr_popup->parent) {
+				popup->parent_was_focused = true;
+			}
 			struct wlr_layer_surface_v1 *layer_surface =
 				popup->lab_layer_surface->layer_surface;
 			focus(&popup->server->seat, layer_surface->surface);

```

tsujan: > I found a real example; in the current PR, closing a right-click menu of qterminal -d closes the terminal rather than restoring the focus to the terminal.

Nice catch!

src/seat.c
johanmalm: Good shout. `sleep 5 ; xfce4-popup-applicationsmenu` followed quickly by running `swaylock` does indeed result in a blackscreen that we cannot get out of.

Have pushed a fix.

src/layers.c
johanmalm: Have removed that wrapper and now just call `seat_force_focus_surface()` directly.

src/layers.c
johanmalm: Have removed excessive, off-topic comment.

src/layers.c
johanmalm: @Consolatis I think this has been overtaken by the patch suggested by @tokyo4j (at bottom of thread) which has now been implemented. Are we happy to go as is?

johanmalm: Thanks for catching the `qterminal -d` case. Have implemented the patch.

johanmalm:
Ready for final review + merge from my perspective. All comments incorporated. Have tested all use-cases that I am aware of and can think of.

tsujan:
I confirm that with the latest version of this patch, when applied to git, the problem of drop-down QTerminal's menus is also fixed :)

jlindgren90:
LGTM. I hope that this can be improved in future with changes on the wlroots side as it seems a bit fragile, but I don't have any better ideas at the moment.

tsujan:
I'm afraid I found another problem:

pcmanfm-qt's desktop is on the background layer (`LayerShellQt::Window::Layer::LayerBackground`). If the desktop has focus and a tooltip is shown for a desktop item, the focus will be taken away from the desktop as soon as the tooltip disappears.

EDIT: The same thing happens when a panel popup/tooltip is shown and closed without further interaction.

The expected behavior in both cases is that the focus should return to the desktop, which had it.

tsujan:
Oh, the same behavior is visible in the dropdown QTerminal: For example, if the "Close Tab" tooltip is shown inside it by putting the mouse cursor over a tab button, the terminal disappears when the tooltip does.

EDIT: And as soon as a tooltip is closed in its Preferences dialog, that dialog will be unfocused. All these symptoms should have the same cause.

EDIT2: Also, when a panel tooltip is shown and closed.

johanmalm:
@tsujan Thanks for catching that regression. I've pushed a fixup which I think does the job. Verification that this does not introduce unwanted side-effects gratefully received.

NOTE: Need to squash last two commits before merge.

src/layers.c
Consolatis: I think its a bit weird that we check the session lock state for popups in layers.c rather than in seat.c as for all other surfaces. Early returning in the seat.c function seems cleaner to me.

Edit:
Whoops, wrong thread.

src/seat.c
Consolatis: I think its a bit weird that we check the session lock state for popups in layers.c rather than in seat.c as for all other surfaces. Early returning in the seat.c function seems cleaner to me.

Edit: fixed

src/layers.c
Consolatis: The whole focus semantics of the layershell protocol is a giant mess. And wlroots not telling us about the grab event + automatically re-routing the focus behind our back isn't making things better. I am not particularly happy about the whole thing but I can't think of something better either.

johanmalm: Okay. Have moved the "locked" check to `seat_force_focus_surface()` in `seat.c`

johanmalm:
Fixups rebased. Ready to go again from my perspective.

tsujan:
It's OK now. As far as I can tell, all problems are fixed. Thanks1

There's a tiny detail that isn't worth mentioning here — maybe later, after labwc 0.9.4 …

src/layers.c
```diff
@@ -465,6 +492,19 @@ handle_popup_commit(struct wl_listener *listener, void *data)
...
+		struct server *server = popup->server;
+		struct seat *seat = &server->seat;
+		if (seat->seat->keyboard_state.focused_surface
+				== popup->wlr_popup->parent) {
+			popup->parent_was_focused = true;
+		}
+		if (popup->wlr_popup->seat) {
+			struct wlr_layer_surface_v1 *layer_surface =
+				popup->lab_layer_surface->layer_surface;
+			seat_force_focus_surface(seat, layer_surface->surface);
```
tokyo4j: @johanmalm I think we have some miscommunication on which surface to gain focus. I supposed that the keyboard focus should be given to the popup surface, not the root layer surface (I thought this PR does this in the earlier review, but it doesn't). The [xdg-shell](https://wayland.app/protocols/xdg-shell#xdg_popup:request:grab) protocol states:

> ### `xdg_popup::grab`
> ... During a popup grab, the client owning the grab will receive pointer and touch events for all their surfaces as normal (similar to an "owner-events" grab in X11 parlance), while **the top most grabbing popup will always have keyboard focus**.

I interpret this sentence that the keyboard focus is given to the popup, not the parent toplevel. And therefore I suppose that the current implementation of xdg-shell and wlr-layer-shell in wlroots-based compositors (e.g. labwc and sway) are violating the protocol as the keyboard focus is never moved during popup grab. It's just that the clients can somehow handle this incorrect compositor implementations.

Therefore, I think we should give keyboard focus to the popup in this PR. And in the future, we should do it also for popups in xdg-shell with some improvements in wlroots.

Here's a patch that works for me:

```diff
diff --git a/include/layers.h b/include/layers.h
index 5e39868c..001347c8 100644
--- a/include/layers.h
+++ b/include/layers.h
@@ -30,7 +30,6 @@ struct lab_layer_popup {
 	struct wlr_xdg_popup *wlr_popup;
 	struct wlr_scene_tree *scene_tree;
 	struct server *server;
-	struct lab_layer_surface *lab_layer_surface;
 	bool parent_was_focused;
 
 	/* To simplify moving popup nodes from the bottom to the top layer */
diff --git a/src/layers.c b/src/layers.c
index cbca97e2..51049b11 100644
--- a/src/layers.c
+++ b/src/layers.c
@@ -403,11 +403,18 @@ handle_map(struct wl_listener *listener, void *data)
 	layer_try_set_focus(seat, layer->scene_layer_surface->layer_surface);
 }
 
+static bool
+surface_is_focused(struct seat *seat, struct wlr_surface *surface)
+{
+	return seat->seat->keyboard_state.focused_surface == surface;
+}
+
 static void
 handle_popup_destroy(struct wl_listener *listener, void *data)
 {
 	struct lab_layer_popup *popup =
 		wl_container_of(listener, popup, destroy);
+	struct seat *seat = &popup->server->seat;
 
 	struct wlr_xdg_popup *_popup, *tmp;
 	wl_list_for_each_safe(_popup, tmp, &popup->wlr_popup->base->popups, link) {
@@ -423,12 +430,13 @@ handle_popup_destroy(struct wl_listener *listener, void *data)
 		wl_list_remove(&popup->commit.link);
 	}
 
-	/* Give focus back to whoever had it before the popup */
-	if (popup->parent_was_focused && popup->wlr_popup->parent) {
-		struct seat *seat = &popup->server->seat;
-		seat_force_focus_surface(seat, popup->wlr_popup->parent);
-	} else {
-		desktop_focus_topmost_view(popup->server);
+	if (surface_is_focused(seat, popup->wlr_popup->base->surface)) {
+		/* Give focus back to whoever had it before the popup */
+		if (popup->parent_was_focused && popup->wlr_popup->parent) {
+			seat_force_focus_surface(seat, popup->wlr_popup->parent);
+		} else {
+			desktop_focus_topmost_view(popup->server);
+		}
 	}
 
 	free(popup);
@@ -496,14 +504,12 @@ handle_popup_commit(struct wl_listener *listener, void *data)
 		/* Force focus when popup was triggered by IPC */
 		struct server *server = popup->server;
 		struct seat *seat = &server->seat;
-		if (seat->seat->keyboard_state.focused_surface
-				== popup->wlr_popup->parent) {
-			popup->parent_was_focused = true;
-		}
-		if (popup->wlr_popup->seat) {
-			struct wlr_layer_surface_v1 *layer_surface =
-				popup->lab_layer_surface->layer_surface;
-			seat_force_focus_surface(seat, layer_surface->surface);
+		bool requesting_grab = !!popup->wlr_popup->seat;
+		if (requesting_grab) {
+			if (surface_is_focused(seat, popup->wlr_popup->parent)) {
+				popup->parent_was_focused = true;
+			}
+			seat_force_focus_surface(seat, popup->wlr_popup->base->surface);
 		}
 	}
 }
@@ -573,8 +579,6 @@ handle_popup_new_popup(struct wl_listener *listener, void *data)
 
 	new_popup->output_toplevel_sx_box =
 		lab_layer_popup->output_toplevel_sx_box;
-
-	new_popup->lab_layer_surface = lab_layer_popup->lab_layer_surface;
 }
 
 /*
@@ -640,7 +644,6 @@ handle_new_popup(struct wl_listener *listener, void *data)
 	}
 
 	popup->output_toplevel_sx_box = output_toplevel_sx_box;
-	popup->lab_layer_surface = toplevel;
 
 	if (surface->layer_surface->current.layer
 			<= ZWLR_LAYER_SHELL_V1_LAYER_BOTTOM) {

```

</details>

This patch includes some other changes:
- Add `requesting_grab` local variable for clarity
- In `handle_popup_destroy()`, skip updating keyboard focus if the popup doesn't have an explicit grab

CC @Consolatis

EDIT: opened https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4048

johanmalm: Thanks this is better.

I tried doing this by giving focus to the popup before, but couldn't get it work, hence the layer-shell surface approach.

I have tested with: `lxqt-panel`, `xfce4-panel`, `qterminal -d`, `pcmanfm-qt --desktop`. Everything seems to work well.

johanmalm:
Right. Will go for the merge now.

tsujan:
Actually, the "tiny detail" that I said wasn't worth mentioning is also fixed in git now :)

Arnaudv6:
I'll document my understanding for those not in the know: to benefit from this: you can use such a wrapper script:
``` bash
#!/bin/sh
# labwc is soon to be white-listed. In the meanwhile...
# https://github.com/lxqt/lxqt-panel/blob/master/panel/lxqtpanel.cpp#L272
XDG_CURRENT_DESKTOP=wayfire "$(command -pv lxqt-panel)" $@
```

stefonarch:
> I'll document my understanding for those not in the know: to benefit from this: you can use such a wrapper script:
> 
> ```shell
> #!/bin/sh
> # labwc is soon to be white-listed. In the meanwhile...
> # https://github.com/lxqt/lxqt-panel/blob/master/panel/lxqtpanel.cpp#L272
> XDG_CURRENT_DESKTOP=wayfire "$(command -pv lxqt-panel)" $@
> ```

That will load the Wayfire IPC backend instead of the generic wlroots one and the desktop/workspace switcher will not work anymore (it looks it won't even be displayed from my test). But
`XDG_CURRENT_DESKTOP=Hyprland "$(command -pv lxqt-panel)" $@` will work as there is no dedicated backend.
PR is here: https://github.com/lxqt/lxqt-panel/pull/2404

Arnaudv6:
thanks @stefonarch for clarifications and MR.

PR #3168, opened 2025-10-24
Avoid consuming scroll on client mousebind

### LINKED ISSUE #3146: mousebinds for scroll in client are not passed to applications
### Problem description

Setting scroll to raise/focus window makes scroll not work inside the applications.
```
  <action name="Focus"/>
  <action name="Raise"/>
</mousebind>
  <action name="Focus"/>
  <action name="Raise"/>
</mousebind>
```

According to this it should work
https://labwc.github.io/labwc-config.5.html
> · Client: The client area of a window, inside its decorations. Events bound to Client are also passed to applications.

Is there another way to do that?

``` bash
$: labwc --version    
labwc 0.9.1 (+xwayland +nls +rsvg +libsfdo)
```
Running it with LXQt (startlxqtwayland)

### Steps to reproduce

1. add this to config
``` xml
<context name="Client">
  <!-- other stuff -->
    <action name="Focus"/>
    <action name="Raise"/>
  </mousebind>
    <action name="Focus"/>
    <action name="Raise"/>
  </mousebind>
</context>
```
2. Try to scroll in some window, actions work (raise/focus), but scrolls are not passed to application.

### labwc build source

Release

### labwc version

0.9.1

### labwc environment

### Distribution

Arch Linux

elviosak:
Closes #3146 

According to suggestions here: https://github.com/labwc/labwc/issues/3146#issuecomment-3408583069

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1346,7 +1346,7 @@ process_cursor_axis(struct server *server, enum wl_pointer_axis orientation,
 		wlr_log(WLR_DEBUG, "Failed to handle cursor axis event");
 	}
 
-	bool handled = false;
+	bool consumed = false;
 	if (direction != LAB_DIRECTION_INVALID) {
 		struct mousebind *mousebind;
 		wl_list_for_each(mousebind, &rc.mousebinds, link) {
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks!
For raising the issue, testing and verifying a fix and then also doing the PR :)

PR #3175, opened 2025-11-01
osd-classic: fix miscalculation for osd width in percentage

tokyo4j:
Fixes a regression in 75eb370.

I guess I thought that we could be completely unaware of output scale when moving from cairo textures to scene nodes, but it was wrong.

### DIFF
--- a/src/osd/osd-classic.c
+++ b/src/osd/osd-classic.c
@@ -86,7 +86,8 @@ osd_classic_create(struct output *output, struct wl_array *views)
 
 	int w = switcher_theme->width;
 	if (switcher_theme->width_is_percent) {
-		w = output->wlr_output->width * switcher_theme->width / 100;
+		w = output->wlr_output->width / output->wlr_output->scale
+			* switcher_theme->width / 100;
 	}
 	int h = wl_array_len(views) * switcher_theme->item_height
 		+ 2 * rc.theme->osd_border_width + 2 * switcher_theme->padding;


johanmalm:
Thanks

tokyo4j:
I realized that output transformation (e.g. 90deg) is not taken into account in this code. I'll open a follow-up PR to use `wlr_output_effective_resolution()`.

EDIT: opened #3177

src/osd/osd-classic.c
```diff
@@ -86,7 +86,8 @@ osd_classic_create(struct output *output, struct wl_array *views)
 
 	int w = switcher_theme->width;
 	if (switcher_theme->width_is_percent) {
-		w = output->wlr_output->width * switcher_theme->width / 100;
+		w = output->wlr_output->width / output->wlr_output->scale
+			* switcher_theme->width / 100;
```
Consolatis, line 90: Can the output scale technically end up being `0`? If yes maybe we should handle that here as it would potentially be a difference of a non working output and a compositor crash.

Consolatis: Oh, never mind. It was replaced in #3177 anyway.

tokyo4j: `wlr_output_effective_resolution()` doesn't seem to care about zero scale. Wlr-output-management protocol prohibits zero scale, but I'm not sure if the scale is always non-zero.

PR #3177, opened 2025-11-01
osd: use wlr_output_effective_resolution() to get output geometry

tokyo4j:
Before this commit, output transformations were not taken into account:

|Before|After|
|-|-|
| 20251102_01h25m10s_grim | 20251102_01h24m59s_grim |

`themerc-override`:

```
osd.window-switcher.style-thumbnail.width: 80%
osd.window-switcher.style-thumbnail.width.max: 80%
```


johanmalm:
Thanks. Great fix.

PR #3179, opened 2025-11-02
osd: place osd at the center of the output rather than the usable area

tokyo4j:


src/osd/osd-classic.c
```diff
@@ -214,10 +215,9 @@ osd_classic_create(struct output *output, struct wl_array *views)
 
 error:;
 	/* Center OSD */
-	struct wlr_box usable = output_usable_area_in_layout_coords(output);
 	wlr_scene_node_set_position(&output->osd_scene.tree->node,
-		usable.x + usable.width / 2 - w / 2,
-		usable.y + usable.height / 2 - h / 2);
+		output_box.x + (output_box.width - w) / 2,
+		output_box.y + (output_box.height - h) / 2);
```
Consolatis, line 220: Heh, I was just typing this up while you already changed it :)

<details><summary>outdated review</summary>
Nitpick:

It feels a bit weird that we have two ways to center things in this PR. Can you change this one to `output_box.x + (output_box.width - w) / 2` (and the same for `y`)? Then all 3 have matching calculations.

With this changed..
</details>

LGTM. Thanks

PR #3180, opened 2025-11-02
cursor: clean up cursor_process_button_{press,release}()

tokyo4j:
This should not change any behaviors.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -953,10 +953,6 @@ static void
 process_release_mousebinding(struct server *server,
 		struct cursor_context *ctx, uint32_t button)
 {
-	if (server->input_mode == LAB_INPUT_STATE_WINDOW_SWITCHER) {
-		return;
-	}
-
 	struct mousebind *mousebind;
 	uint32_t modifiers = keyboard_get_all_modifiers(&server->seat);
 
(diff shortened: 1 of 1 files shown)


src/input/cursor.c
```diff
@@ -1115,8 +1107,11 @@ cursor_process_button_press(struct seat *seat, uint32_t button, uint32_t time_ms
 		 * so subsequent release always closes menu or selects menu item.
 		 */
 		press_msec = 0;
-		lab_set_add(&seat->bound_buttons, button);
-		return false;
+		goto consumed;
+	}
+
+	if (server->input_mode != LAB_INPUT_STATE_PASSTHROUGH) {
+		goto consumed;
```
tokyo4j, line 1114: This makes the code paths below not taken when window switching, resizing etc., but I don't think it makes any changes in behaviors, as `seat_focus_surface()` does nothing when `server->input_mode != LAB_INPUT_STATE_PASSTHROUGH`. Let's align with `cursor_process_button_release()`.

src/input/cursor.c
```diff
@@ -953,10 +953,6 @@ static void
 process_release_mousebinding(struct server *server,
 		struct cursor_context *ctx, uint32_t button)
 {
-	if (server->input_mode == LAB_INPUT_STATE_WINDOW_SWITCHER) {
-		return;
-	}
-
```
tokyo4j, line 959: This deletion is fine, because `cursor_process_button_release()` already checks `server->input_mode != LAB_INPUT_STATE_PASSTHROUGH`

PR #3182, opened 2025-11-02
osd: move end_cycling() into osd_finish()

tokyo4j:
...so that we can use `osd_finish()` to support clicking an osd item to focus its associated window: https://github.com/labwc/labwc/commit/eac5f0f892d7350053b0b070c50ddd3f2b93ccbb


johanmalm:
LGTM. Thanks.

PR #3183, opened 2025-11-02
Add allOutputs option to window switcher

SametAylak:
Hello team,

First, I want to mention that this is a great project. I really appreciate seeing something minimal and modern in the Wayland compositor landscape. This is my first contribution to a Wayland compositor in general, so please forgive me if I don't follow best practices.

This PR introduces a new configuration option for the window switcher that allows users to filter windows by output (monitor).

Changes:
- Added `allOutputs` attribute to <windowSwitcher>
  - Defaults to "no" (shows only windows on the cursor's output)
  - When set to "yes", shows windows from all outputs

- Introduced LAB_VIEW_CRITERIA_CURSOR_OUTPUT flag to filter views by output
  - Windows are matched against the output nearest to the cursor
  - Bit positions of existing criteria flags were shifted to accommodate the new flag

- Modified OSD display behavior to show only on the cursor's output
  - Previously, the OSD appeared on all outputs simultaneously
  - Now displays only on the output nearest to the cursor location

This provides better multi-monitor support by allowing users to focus on windows relevant to their current screen, reducing visual clutter when working across multiple displays.

Should fix https://github.com/labwc/labwc/discussions/3067

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -339,7 +339,7 @@ this is for compatibility with Openbox.
 ## WINDOW SWITCHER
 
 ```
-<windowSwitcher show="yes" style="classic" preview="yes" outlines="yes" allWorkspaces="no" allOutputs="no">
+<windowSwitcher show="yes" style="classic" preview="yes" outlines="yes" allWorkspaces="no" allOutputs="yes">
   <fields>
     <field content="icon" width="5%" />
     <field content="desktop_entry_name" width="30%" />
@@ -367,8 +367,7 @@ this is for compatibility with Openbox.
 	are shown).
 
 	*allOutputs* [yes|no] Show windows regardless of what output
-	they are on. Default no (that is only windows on the cursor output
-	are shown).
+	they are on. Default yes.
 
 	*unshade* [yes|no] Temporarily unshade windows when switching between
 	them and permanently unshade on the final selection. Default is yes.
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -78,7 +78,7 @@
   </theme>
 
   <windowSwitcher show="yes" style="classic" preview="yes"
-      outlines="yes" allWorkspaces="no" allOutputs="no" unshade="yes">
+      outlines="yes" allWorkspaces="no" allOutputs="yes" unshade="yes">
     <fields>
       <field content="icon" width="5%" />
       <field content="desktop_entry_name" width="30%" />
@@ -98,7 +98,7 @@
     Some contents are fixed-length and others are variable-length.
     See "man 5 labwc-config" for details.
 
-    <windowSwitcher show="yes" preview="no" outlines="no" allWorkspaces="yes" allOutputs="no">
+    <windowSwitcher show="yes" preview="no" outlines="no" allWorkspaces="yes" allOutputs="yes">
       <fields>
         <field content="workspace" width="5%" />
         <field content="state" width="3%" />
@@ -118,7 +118,7 @@
     then workspace name, then identifier/app-id, then the window title.
     It uses 100% of OSD window width.
 
-    <windowSwitcher show="yes" preview="no" outlines="no" allWorkspaces="yes" allOutputs="no">
+    <windowSwitcher show="yes" preview="no" outlines="no" allWorkspaces="yes" allOutputs="yes">
       <fields>
         <field content="custom" format="foobar %b %3s %-10o %-20W %-10i %t" width="100%" />
       </fields>
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1218,9 +1218,9 @@ entry(xmlNode *node, char *nodename, char *content)
 				~LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
 		}
 	} else if (!strcasecmp(nodename, "allOutputs.windowSwitcher")) {
-		if (parse_bool(content, -1) == true) {
-			rc.window_switcher.criteria &=
-				~LAB_VIEW_CRITERIA_CURSOR_OUTPUT;
+		if (parse_bool(content, -1) == false) {
+			rc.window_switcher.criteria |=
+				LAB_VIEW_CRITERIA_CURSOR_OUTPUT;
 		}
 	} else if (!strcasecmp(nodename, "unshade.windowSwitcher")) {
 		set_bool(content, &rc.window_switcher.unshade);
@@ -1438,7 +1438,6 @@ rcxml_init(void)
 	rc.window_switcher.outlines = true;
 	rc.window_switcher.unshade = true;
 	rc.window_switcher.criteria = LAB_VIEW_CRITERIA_CURRENT_WORKSPACE
-		| LAB_VIEW_CRITERIA_CURSOR_OUTPUT
 		| LAB_VIEW_CRITERIA_ROOT_TOPLEVEL
 		| LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER;
 


src/osd/osd.c
```diff
@@ -292,11 +292,20 @@ update_osd(struct server *server)
...
+		// wl_list_for_each(output, &server->outputs, link) {
+		// 	if (!output_is_usable(output)) {
+		// 		continue;
+		// 	}
+		// 	if (!output->osd_scene.tree) {
+		// 		osd_impl->create(output, &views);
+		// 		assert(output->osd_scene.tree);
+		// 	}
+		// 	osd_impl->update(output);
+		// }
```
SametAylak, line 305: We might wanna support this by adding a new configuration keyword. I'm open to any suggestions.

tokyo4j:
I think we should keep the existing default behaviors. All the WMs I'm aware of shows windows from all the outputs (see below). And I prefer showing OSD at all outputs, because "focused output" is not always the output that contains the cursor (e.g. after switching focus with Alt-Tab or moving the window with keybinds).

For reference:

|| KDE | Xfce4 | Openbox | Windows 11 |
|-|-|-|-|-|
|shows windows from| all outputs | all outputs | all outputs | all outputs |
|shows OSD at  | output with focused window | all outputs| primary output | primary output |

EDIT: KWin turned out to show the OSD at the output with a focused window, not with the cursor.

src/osd/osd.c
tokyo4j: Hm, maybe we can restructure our configuration like this:

```xml
<windowSwitcher preview="yes|no" outlines="yes|no" unshade="yes|no">
  <osd show="yes|no" style="classic|thumbnail" allOutputs="yes|no" />
  <filter currentOutput="yes|no" currentWorkspace="yes|no" />
</windowSwitcher>
```

I suggest keeping this code as it is for this PR. We can make a follow-up PR to implement my suggestion above, or you can implement it by yourself.

EDIT: added `<filter />` in my suggestion.

EDIT2: Another option is just letting `<windowSwitcher><allOutputs>` control both of whether to show windows from all the outputs and whether to show OSD in the current output. It doesn't require reconstruction of our configuration and can be done within this PR.

SametAylak:
> I think we should keep the existing default behaviors. All the WMs I'm aware of shows windows from all the outputs (see below). And I prefer showing OSD at all outputs, because "focused output" is not always the output that contains the cursor (e.g. after switching focus with Alt-Tab or moving the window with keybinds).
> 
> For reference:
> 
> KDE	Xfce4	Openbox	Windows 11
> shows windows from	all outputs	all outputs	all outputs	all outputs
> shows OSD at	output with cursor	all outputs	primary output	primary output

Yes, that's my bad. It's make sense to keep the existing defaults right now. I'll make the necessary changes.

johanmalm:
Hi @sametaylak and welcome

Looks like you've got your head around coding style and approach quickly.

Suggest we start by talking through behaviour first and then discuss how we structure commits.

I feel this needs some thought because we're talking two things:

1. Which windows are shown in the osd (e.g. all windows or just those windows on the same output as where the osd is shown)
2. Which output(s) the osd(s) is/are shown on (e.g. all outputs, just the output with where the pointer is, or the output when there client with keyboard focus is). I think the option regarding keyboard focus is what @tokyo4j described in the later post about changing focus with alt-tab.

I like @tokyo4j's suggestion regarding a restructured config because it deals with combinations of behavior described in (1) and (2). It also looks like we can deprecate extant attribute `allWorkspaces` for a few releases so that we do not immediately break backward compatibility.

Instead of the suggest `<osd allOutputs="yes|no"/>`, consider something like `<osd output="all|pointer|keyboardFocus">` to cater for options described in point (2) above.

I think it'll be easier to review and discuss all this if we carefully craft small succinct commits to deal with separate concerns one at a time.

It looks like the CI has errored on CodingStyle. Those checks can be run locally with `scripts/check --file=src/osd/osd.c`

tokyo4j:
I agree with @johanmalm. Let's decide the behaviors first.

I'm not sure if we really need both of `pointer` and `keyboardFocus`. I suggest implementing only `pointer` for now and adding `keyboardFocus` later if someone requested it. I also think we should have similar options for `<filter>`.

To sum up, we can have new configurations like this:

```xml
  <windowSwitcher preview="" outlines="" unshade="">
    <osd show="" style="" output="all|pointer" />
    <filter output="all|pointer" workspace="all|current" />
  </windowSwitcher>
```

---

This is a bit off-topic, but I wonder if anyone wants to have `<filter>` option in `NextWindow` actions themselves, not as global configuration. I mean something like this:

```xml
  <action name="NextWindow">
    <filter output="all|pointer" workspace="all|current" />
  </action>
```

This will be more attracting if we support something like `<filter application="all|current" >` that allows switching windows with the same app-id (#3060).

I think we can implement `<windowSwitcher><filter>` (as suggested above) for now and additionally implement `<action name="NextWindow"><filter>` that can overwrite `<windowSwitcher><filter>` later. @johanmalm @Consolatis Do you think we should avoid this duplication and have only `<action name="NextWindow"><filter>`? If so, we should not have `<filter>` option in `<windowSwitcher>`.

johanmalm:
I like the `<filter>` option in `NextWindow`. I think both KDE and Windows11 supporting that type of functionality with Alt-Tab, Super-Tab, or similar.

I think it would be better to only do `<filter>` in one place. Multiple ways of configuring the same thing tends to (in my experience) grow complexity disproportionately (e.g. precedence logic).

Considering `s/filter/windowFilter/` to make it clearer what the attribute `output` refers to.

SametAylak:
What @tokyo4j suggests is actually quite readable and easy to understand.
```xml
  <windowSwitcher preview="" outlines="" unshade="">
    <osd show="" style="" output="all|pointer" />
    <windowFilter output="all|pointer" workspace="all|current" />
  </windowSwitcher>
```

I'm not deeply familiar with the codebase, but this seems like a solid approach for separating responsibilities between the window switcher and OSD.

Consolatis:
So what is the game plan here?

It appears that this could be implemented in two independent parts:
- Only showing the OSD on the output nearest to the cursor / where the majority of the keyboard focused window is / on all outputs (`<osd output="pointer|focused|all|">`). I think its important to have the `focused` value in there because on some workflows the mouse pointer ends up being somewhere random while only working with the keyboard. In general using the mouse pointer as indicator for "this output is active" has drawbacks. We could even consider removing the `pointer` option and always go via `server->active_view->output` and thus only have the options of `<osd output="focused|all">`.
- Filtering the windows which are shown within the OSD, either by a global `<windowSwitcher><windowFilter>` config or a `NextWindow/PreviousWindow` argument. Here the options seem to be `output="current|all"`. If we decide to go the `NextWindow` variant we should then likely also move the `workspace="all|current"` there and provide some fallback for the next versions to parse the old global syntax.

@sametaylak 
The first part seems relatively straight forward while the second part needs some more discussion / design work. So I'd suggest to create a new PR for the first part. We can use this PR for design work on the second part.

SametAylak:
Sounds like a plan. I'll come up with an another PR that implements:
```xml
    <osd show="" style="" output="all|pointer|keyboardFocus" />
```

SametAylak:
We’ve completed the first part, which involved adding the `<osd />` element to improve the separation between windowSwitcher and OSD.

Many thanks to @tokyo4j and @Consolatis for their contributions.

How should we proceed with the second discussion?

tokyo4j:
I think we can proceed in two steps:

- Move `<windowSwitcher allWorkspace="yes|no">` to `<action name="NextWindow"><filter workspace="all|current">` with backward compatibility preserved.
- Add `<action name="NextWindow"><filter output="all|pointer|keyboard">`.

I'm not sure whether to split PRs, but I suggest splitting them. I can take this over if you feel this is too deep into our codebase, but it might take some weeks since I'm currently busy for my university stuffs.

Consolatis:
`<action name="NextWindow"><filter output="all|pointer|keyboard" /></action>` feels a bit weird to me. I don't think we have that nested element syntax for any other usual action (outside of `If` and `ForEach`). What about something like `<action name="NextWindow" output="all|pointer|keyboard" />` and `<action name="NextWindow" workspace="all|current" />` instead? We could potentially also extend this later on to include things like `tile` or `region` (I think there were feature requests for this a while ago).

To make it a bit more self-declarative we could use `filterOutput` and `filterWorkspace` as argument name but I don't feel strongly about that.

SametAylak:
```xml
<windowSwitcher>
  <osd output="pointer" />
  <windowFilter output="pointer" workspace="current" />
</windowSwitcher>
```
This basically makes the window switcher __kinda__ `self-contained`.

Pros:
- Nice and predictable
- If multiple actions call the window switcher (NextWindow, PreviousWindow, mouse bindings…), they all behave the same

Cons:
- Less flexible. All keybinds use the same filter
- Power users may want different behaviors per shortcut

Overall it feels… tidy. Harder to screw up.

---
```xml
<windowSwitcher>
  <osd output="pointer" />
</windowSwitcher>

<keyboard>
  <keybind key="A-Tab">
    <action name="NextWindow" output="pointer" workspace="current" />
  </keybind>
  <keybind key="A-S-Tab">
    <action name="PreviousWindow" output="pointer" workspace="current" />
  </keybind>
</keyboard>
```
This lets each keybinding completely define its switching behavior.

Pros:
- Very flexible
- Configs become programmable in a way

Cons:
- Filters get duplicated across keybinds
- Future changes require updating N locations
- Copy-paste errors are inevitable

Feels powerful, but also a bit dangerous.

---

I kinda prefer having `windowFilter` in the `<windowSwitcher>` section.

This gives:
- Consistent behavior
- OSD + filtering stay in the same scope without needing to duplicate the behavior

I was also planning to bring up having workspace sets per output, which is how wayfire describes it. And I believe having configuration per-keybinding will make things harder to follow and implement, perhaps.

Consolatis:
With the filter per action a user can implement different window switcher keys like `W-Tab` to switch within a workspace and `A-Tab` to switch globally. And potentially within a tile/region in the future. This would not be possible if we make it a global setting.

johanmalm:
I think the point made by Consolatis is important. It doesn't feel good to shut the door on different types of alt-tab.

Are we sure that the second lot of cons are true? I.e. could common code be used?

SametAylak:
Now that I think of it, having filtering on keybindings might be beneficial. Consolatis’s `W-Tab` and `A-Tab` example was enough for me. After that, I also started thinking about having `X-Tab` behavior on the same app_ids that I generally use in KDE.

Once we’re aligned, I’m happy to implement this if that works for you.

PR #3186, opened 2025-11-03
osd: allow clicking an osd item to focus the window

tokyo4j:
https://github.com/user-attachments/assets/a618b88b-0349-4199-a581-d6d6e0bebfca

The 1st & 2nd commits generalize `osd_{classic,thumbnail}_item` so that we can associate it to the node descriptors and get the view from `osd_on_cursor_release()`. This is also useful for scrolling: https://github.com/tokyo4j/labwc/blob/69b539d11928387151880d305ae99b51c2191b76/src/osd/scroll.c#L90-L97


Consolatis:
Code LGTM. The refactoring as well. As for the activate on click feature itself: up to @johanmalm.

I am slightly worried about the growing complexity of the window switcher but as its mostly contained in `src/osd/` I am fine with it. There is also not yet really a way to properly implement the thumbnail variant with an external tool.

johanmalm:
LGTM.

I like this feature and think it helps usability a lot.

PR #3189, opened 2025-11-05
window rules: add `<taskbarScope>` with 4 visibility modes; gate taskbar listing via foreign-toplevel outputs

mbroemme:
## Summary
This PR introduces a per-window rule `<taskbarScope>` that controls where a window's taskbar entry appears. It affects taskbar listing only (via `zwlr_foreign_toplevel_management_v1` output membership); it does not change sticky/omnipresent state, focus, or workspace membership. The initial approach to solving this problem was to implement the `workspace_enter` and `workspace_leave` events in PR #3164 and have certain taskbars react to these notifications. However, the approach presented here is more correct.

## Why
- Users want per-workspace taskbars without losing flexibility for apps that should appear "everywhere".
- Naming emphasizes taskbar behavior to avoid confusion with "sticky/omnipresent" windows.

## Configuration
- `here` - show only when both match: the window's monitor **and** its active workspace
- `monitor` - show only on the window's monitor, across that monitor’s workspaces (**default** if omitted)
- `workspace` - show on all monitors currently on the window’s workspace
- `everywhere` - show on all monitors and all workspaces

This gives users precise control per application (e.g., music/chat apps "everywhere", IDEs "workspace", utilities "monitor", or strict "here"). Example configuration:

```xml
<windowRules>
  <windowRule>
    <identifier>org.gnome.Lollypop</identifier>
    <taskbarScope>everywhere</taskbarScope>
  </windowRule>

  <windowRule>
    <identifier>alacritty</identifier>
    <taskbarScope>monitor</taskbarScope>
  </windowRule>

  <windowRule>
    <identifier>code</identifier>
    <taskbarScope>workspace</taskbarScope>
  </windowRule>

  <windowRule>
    <identifier>org.gnome.Terminal</identifier>
    <!-- key omitted => default here-only -->
  </windowRule>
</windowRules>
```

### Single monitor (MON1), two workspaces (WS1 & WS2)
*Before:* On WS1, taskbar shows apps from WS1 + WS2.
| Workspace | Taskbar |
|-|-|
| 1 | image |
| 2 | image |

*After:* On WS1, taskbar shows only WS1 apps (for windows with `taskbarScope=here`).
| Workspace | Taskbar |
|-|-|
| 1 | image |
| 2 | image |

### Dual monitors (MON1 & MON2), single workspace (WS1)
*Before:* Each monitor's taskbar shows only windows physically on that monitor.
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |

*After:* With `taskbarScope=workspace` for an app, show the same WS1 app list on both monitors.
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |

*After:* With `taskbarScope=monitor` for another app, show it only on its own monitor (still across that monitor's workspaces).
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |

### Dual monitors (MON1 & MON2), two workspaces (WS1 & WS2)
*Before:* Both taskbars list all apps (or are tied only to physical output, depending on baseline).
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |
| 1 | image |
| 2 | image |

*After:* With `taskbarScope=here` windows appear only on the monitor whose workspace matches.
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |
| 1 | image |
| 2 | image |

*After:* With `taskbarScope=workspace` windows appear only on the monitor that is currently showing their workspace.
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |
| 1 | image |
| 2 | image |

*After:* With `taskbarScope=everywhere` windows appear on both taskbars.
| Monitor | Taskbar |
|-|-|
| 1 | image |
| 2 | image |
| 1 | image |
| 2 | image |

## Backward Compatibility
- If `<taskbarScope>` is omitted, behavior remains the classic per-monitor.
- No protocol changes; panels (e.g., Waybar's `wlr/taskbar`) follow compositor-driven `output_enter/leave`.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -1304,6 +1304,16 @@ situation.
 	This property allows prioritizing client supplied icons for specific
 	applications. Default is server.
 
+*<windowRules><windowRule taskbarScope="">* [here|monitor|workspace|everywhere]
+	Controls where this window's taskbar entry is shown (taskbar listing
+	only).
+
+	Values:
+	- *here* — this monitor **and** its current workspace
+	- *monitor* — this monitor across its workspaces (default)
+	- *workspace* — all monitors currently on this workspace
+	- *everywhere* — all monitors & all workspaces
+
 ## MENU
 
 ```


tokyo4j:
Sorry for not participating the discussion in #3164, but I'm not a fan of this approach that works around the protocol's limitation. Waybar, for example, has `all-outputs` option for its `wlr/taskbar` modules that configures whether to show windows on the all outputs or only on the same output as the panel. `taskbarScope="monitor"` in this PR seems to duplicate it, thus it doesn't sound in our scope.

Adding events to describe toplevel-workspace relationship in wayland-protocols will take years to come, which is very frustrating, but I think we should not rush implementing it in compositor side so that we don't have duplicated configurations in both compositor and client side when it's finally supported in wayland-protocol. Once implemented, we'll need maintain it forever and it becomes a technical debt for us. Also, having practical inconvenience due to the immature protocols can be a pressure for wayland-protocol to accelerate the discussion there, hopefully.

mbroemme:
@tokyo4j I can drop `taskbarScope="monitor"`. It was only there to differentiate scopes, and since `monitor` is the default (and backward-compatible) fallback, removing it is fine. Generally I prefer explicit, configurable values so the intent is clear - otherwise, in a few years no one will remember why `monitor` was the default.

Consolatis:
I agree on all points made by @tokyo4j.

Personally I'd rather throw another "unofficial" bridge protocol into the mix (maybe in collaboration with lxqt / xfce / cosmic)  than hacking workarounds into clients and compositors. Last time we did that with the cosmic-workspaces protocol it kind of put more pressure onto wayland-protocols to finally merge ext-workspace. This time its a bit more tricky though because an "official" bridge protocol would connect a ext-workspace handle to a ext-foreign-toplevel handle which is pretty much useless without the -info and -management protocols also merged. So we basically have two options here:
- The potential upstream route:
  - Connect wlr-foreign-toplevel-handle to ext-foreign-toplevel-handle (maybe by using its unique id)
  - Connect ext-foreign-toplevel-handle to ext-workspace-handle
- The downstream route:
  - Connect wlr-foreign-toplevel-handle to ext-workspace-handle directly

Edit:
Actually, I think we could also have a single bridge protocol which has two constructors, one based on a wlr-foreign-toplevel handle and another based on a ext-foreign-toplevel-handle. The resulting object wrapping either of the two would then behave identical so clients would only need to change like 2 lines of code to switch over if it gets upstreamed. That doesn't provide a solution to missing toplevel capture for wlr-foreign-toplevel handles though..

Consolatis:
While this is an interesting approach to bypass the missing protocol I feel like the approach does not match the philosophy of the labwc project. We would rather see this fixed with the introduction of a new (preferably official) wayland protocol so any client using it will function as expected on all compositors supporting the new protocol.

PR #3193, opened 2025-11-05
docs: mention path for custom user keyboard layouts

ManuLinares:
https://github.com/labwc/labwc/discussions/3191#discussioncomment-14885443

### DIFF
--- a/docs/environment
+++ b/docs/environment
@@ -10,6 +10,7 @@
 ## /usr/share/X11/xkb/rules/evdev.lst
 ##
 ## Multiple keyboard layouts can be set by comma-separating the country codes.
+## Custom user layouts lookup path: $HOME/.config/xkb/symbols/
 ## If a variant layout is needed, the syntax is layout(variant)
 ## If multiple layouts are used, specify the toggle-keybind using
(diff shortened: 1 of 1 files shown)


johanmalm:
Would it make sense to add a full example (like you did in #3191)?

ManuLinares:
I'm not much of a documentation guy, more of a technical fix stuff guy.
If somebody wants to document this better I'll write here my thinking just for reference.

--- 

https://wiki.archlinux.org/title/X_keyboard_extension#Editing_the_layout

xkb documentation is horrible. I used the archwiki docs which are far better simplified, and instead of creating a new `/usr/share/X11/xkb/symbols/latam` like `/usr/share/X11/xkb/symbols/latam_myowncustomwhatever` I found this solution (creating a `$HOME/.config/xkb/symbols/mylatam` which includes the original "latam")

This solution was better than to install software input mappers, mess with the udev, etc.

Consolatis:
I wonder if we should put the example from #3191 into https://labwc.github.io/faq.html and then only add a link to the FAQ in the `environment` file with something like:
```
# To create a custom layout or overwrite single keys see https://labwc.github.io/faq.html#custom_layout
```

Issue #3160, opened 2025-10-21
Margins quit being reserved at the edge of the screen after upgrading to 0.9.2


cole-mk:
### Problem description

In my rc.xml I have margins set
``` <margin top="6" bottom="3" left="3" right="3" output="" />```

After upgrading to 0.9.2, my waybar instances are positioned all the way at the top and bottom of the screen, and maximized windows extend all the way to the left and right edges.

I have tried copying the latest rc.xml.all to my config directory and only setting margins. I have also tried installing the labwc-git package from the AUR, which I believe pulls the latest commit from this repo. Neither of those things helped. However, downgrading labwc to version 0.9.1 does restore the margins.

Thank you for taking the time to look at this!

### Steps to reproduce

1. Set margins to some value in rc.xml.
2. reconfigure or restart labwc
3. Open a program and maximize its window.

### labwc build source

Release

### labwc version

0.9.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch linux

tokyo4j:
Sorry, this is due to my PR #2667. As a quick fix, you can remove `output=""` to restore the old behavior.

Like #3006, the cause is that we no longer skip empty values in `<margin />`. I guess we have the same issue in `<touch deviceName="" mapToOutput="" mouseEmulation="no"/>` advertised in `rc.xml.all`.

Now I think we underestimated the number of users who use empty values in `rc.xml` as "ignore" and we should restore the old behavior. @johanmalm @Consolatis What do you think?

cole-mk:
> As a quick fix, you can remove output="" to restore the old behavior.

Thank you! Its working as expected now.

johanmalm:
My intuitive feel is that we should leave it as it is. I'm not basing that sentiment on logical reasoning or analysis, but merely on the fact that we've not had much feedback regarding this, and by now other users might have started relying on it the way it is.

In the manual it says `output is optional; if this attribute is not provided (rather than leaving it an empty string) the margin will be applied to all outputs.` which makes me think that the behavior described in this issue has always been like it.

Ref:
- https://labwc.github.io/labwc-config.5.html#margin

johanmalm:
> However, downgrading labwc to version 0.9.1 does restore the margins.

Sorry - just saw this. Right, so something has changed, but it's curious that the man page seems to describe the new behavior (unless I'm thinking about this the wrong way).

tokyo4j:
[...]

PR #3196, opened 2025-11-07
osd: make empty string "" in thumbnailLabelFormat not show any text

elviosak:
saw this

https://github.com/labwc/labwc/issues/3160#issuecomment-3499753127
> for example the OSD format strings where the user may actually want an empty string and it should not fall back to its default value in that case.

and noticed empty string was ignored in `thumbnailLabelFormat`, instead it used the default format "%T", while `<fields><field content="custom">` does accept the empty string and shows empty space.

with this commit this is possible:

qse_1762484165

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1068,6 +1068,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		load_default_mouse_bindings();
 	} else if (!strcasecmp(nodename, "prefix.desktops")) {
 		xstrdup_replace(rc.workspace_config.prefix, content);
+	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.windowSwitcher")) {
+		xstrdup_replace(rc.window_switcher.thumbnail_label_format, content);
 
 	} else if (!lab_xml_node_is_leaf(node)) {
 		/* parse children of nested nodes other than above */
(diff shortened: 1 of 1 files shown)


src/config/rcxml.c
```diff
@@ -1069,6 +1069,8 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "prefix.desktops")) {
 		xstrdup_replace(rc.workspace_config.prefix, content);
 
+	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.windowSwitcher")) {
+		xstrdup_replace(rc.window_switcher.thumbnail_label_format, content);
```
tokyo4j, line 1072: nit: please move the empty line like below so that the comment `	/* handle nodes without content, e.g. <keyboard><default /> */` is also applied for the added lines.

```suggestion
	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.windowSwitcher")) {
		xstrdup_replace(rc.window_switcher.thumbnail_label_format, content);

```

tokyo4j:

src/config/rcxml.c
elviosak: done and squashed, thanks

tokyo4j:
Thanks!

ManuLinares:
Please revert this merge. It breaks thumbnailLabelFormat.

```xml
  <windowSwitcher style="thumbnail" show="yes" preview="no" outlines="no" 
  allWorkspaces="no" unshade="yes" thumbnailLabelFormat="%b %n"/>
```

Consolatis:
This MR is likely unrelated. The setting was just moved to `<windowSwitcher><osd thumbnailLabelFormat="%T" />`. There is no fallback for the old format because it was only available for a few days and wasn't in any release yet.

PR #3199, opened 2025-11-07
view: update top layer visiblity at unmap instead of destroy

jlindgren90:
It's possible for a fullscreen xwayland view to be unmapped without being destroyed. In this case, we need to update top layer visibility, otherwise panels and the like will remain hidden.

Since unmap is always called before destroy, it's sufficient to do the update only in view_impl_unmap() and not in view_destroy().

Adaptive sync logic needs work still, but I tried to minimize changes to it since I don't have hardware to test it.

Test case (run with no other views open):
```
#include <QApplication>
#include <QWidget>
#include <QTimer>
#include <stdlib.h>

int main(int argc, char *argv[]) {
    setenv("QT_QPA_PLATFORM", "xcb", 1);
    QApplication app(argc, argv);
    QWidget window;
    window.showFullScreen();
    QTimer::singleShot(1000, &window, &QWidget::hide);
    return app.exec();
}
```

### DIFF
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -1,11 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0-only
 /* view-impl-common.c: common code for shell view->impl functions */
 #include "view-impl-common.h"
-#include "config/rcxml.h"
 #include "foreign-toplevel/foreign.h"
 #include "labwc.h"
 #include "output.h"
-#include "output-state.h"
 #include "view.h"
 #include "window-rules.h"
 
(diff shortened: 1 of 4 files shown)


src/view-impl-common.c
```diff
@@ -52,6 +55,21 @@ view_impl_unmap(struct view *view)
 	if (view == server->active_view || !server->active_view) {
 		desktop_focus_topmost_view(server);
 	}
+
+	desktop_update_top_layer_visibility(view->server);
```
Consolatis, line 57: Hm.. stumbled over the removal of the fullscreen check here but I guess it actually makes sense to not check for it: the window might have been the only one in front of a fullscreen window.

Consolatis (commented):
The code looks good to me for now but I wonder if this is the right approach for long term maintainability. We are doing the same thing in multiple places (files in this case) which makes things hard to follow.

jlindgren90:
> We are doing the same thing in multiple places (files in this case) which makes things hard to follow.

I assume this refers to the adaptive sync logic. I considered moving the private function set_adaptive_sync_fullscreen to output.c instead. I can do that still if you like.

I think even with that refactor, the adaptive sync logic will still be wrong, and should probably just re-use the same as we have already for top layer visibility. But that would be a more invasive change, and I can't test it myself.

Consolatis:
Thanks :)

PR #3200, opened 2025-11-08
Map/unmap/minimize logic cleanups

jlindgren90:
A couple of fixes to make the map/unmap/minimize logic a bit easier to follow. There shouldn't be any behavioral change, except *possibly* eliminating some spurious minimize/un-minimize transitions for initially minimized views.

Descriptions below copied-and-pasted from the commit messages.

- xwayland: connect commit and surface_destroy handlers together
    
    Factor out set_surface() which consolidates connecting/disconnecting
    the wlr_surface event listeners in one place.
    
    In theory, this means we can receive commit events for minimized views.
    However, with a test app that resizes itself, I didn't see any change,
    i.e. the commits still don't come through until un-minimize. It's
    possible they are being filtered at wlroots or protocol level.
    
    Also remove an old, semi-related TODO from view.c.

- view: less hacky support for minimize-before-map
    
    The previous "minimal fix" (5148c2aa3140) worked but was a bit of a
    hack, as it basically un-minimized and then immediately minimized the
    view again at map. It's not actually too difficult to make the map
    handlers aware of minimized views, eliminating the need for the hack.
    
    Note: this depends on the previous commit ("xwayland: connect commit
    and surface_destroy handlers together") otherwise the xwayland map
    handler registers the commit handler twice, leading to a crash.

### DIFF
--- a/src/foreign-toplevel/foreign.c
+++ b/src/foreign-toplevel/foreign.c
@@ -18,7 +18,6 @@ struct foreign_toplevel *
 foreign_toplevel_create(struct view *view)
 {
 	assert(view);
-	assert(view->mapped);
 
 	struct foreign_toplevel *toplevel = znew(*toplevel);
 	wlr_foreign_toplevel_init(&toplevel->wlr_toplevel, view);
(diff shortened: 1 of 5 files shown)


johanmalm:
Thanks.

I think it's quite hard to validate the correctness of this, but it LGTM and seems nicer than before.

I appreciate your effort in grinding out the details with this.

PR #3201, opened 2025-11-08
windowSwitcher: add multi-monitor OSD support

SametAylak:
# Add multi-monitor OSD support to window switcher

## Summary
Adds multi-monitor support to the window switcher OSD with configurable output criteria.

## Changes
- New `output` attribute with three modes:
  - `all`: Display OSD on all monitors (default)
  - `pointer`: Display on monitor containing the cursor
  - `keyboard`: Display on monitor with keyboard focus
- Refactored XML structure: OSD settings now nested under `<osd>` element

## Configuration Example
```xml
<windowSwitcher preview="yes" outlines="yes" allWorkspaces="no">
  <osd show="yes" style="classic" output="pointer" />
  <fields>
    ...
  </fields>
</windowSwitcher>
```

## Breaking Changes
The `<windowSwitcher>` element structure has changed. OSD-specific attributes (`show`, `style`, `thumbnailLabelFormat`) must now be placed within an `<osd>` child element. The new `output` attribute is also configured here.

### Migration Example
```diff
- <windowSwitcher show="yes" style="classic" preview="yes">
+ <windowSwitcher preview="yes">
+   <osd show="yes" style="classic" output="all" />
    <fields>...</fields>
  </windowSwitcher>
```

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -339,7 +339,8 @@ this is for compatibility with Openbox.
 ## WINDOW SWITCHER
 
 ```
-<windowSwitcher show="yes" style="classic" preview="yes" outlines="yes" allWorkspaces="no" thumbnailLabelFormat="%T">
+<windowSwitcher preview="yes" outlines="yes" allWorkspaces="no">
+  <osd show="yes" style="classic" output="all" thumbnailLabelFormat="%T" />
   <fields>
     <field content="icon" width="5%" />
     <field content="desktop_entry_name" width="30%" />
@@ -348,14 +349,7 @@ this is for compatibility with Openbox.
 </windowSwitcher>
 ```
 
-*<windowSwitcher show="" style="" preview="" outlines="" allWorkspaces="" unshade="" thumbnailLabelFormat="">*
-	*show* [yes|no] Draw the OnScreenDisplay when switching between
-	windows. Default is yes.
-
-	*style* [classic|thumbnail] Configures the style of the OnScreenDisplay.
-	"classic" displays window information like icons and titles in a vertical list.
-	"thumbnail" shows window thumbnail, icon and title in grids.
-
+*<windowSwitcher preview="" outlines="" allWorkspaces="" unshade="">*
 	*preview* [yes|no] Preview the contents of the selected window when
 	switching between windows. Default is yes.
 
@@ -369,12 +363,26 @@ this is for compatibility with Openbox.
 	*unshade* [yes|no] Temporarily unshade windows when switching between
 	them and permanently unshade on the final selection. Default is yes.
 
+*<osd show="" style="" output="" thumbnailLabelFormat="" />*
+	*show* [yes|no] Draw the OnScreenDisplay when switching between
+	windows. Default is yes.
+
+	*style* [classic|thumbnail] Configures the style of the OSD.
+	"classic" displays window information like icons and titles in a vertical list.
+	"thumbnail" shows window thumbnail, icon and title in grids.
+
+	*output* [all|pointer|keyboard] Configures which monitor(s) show the OSD.
+	"all" displays the OSD on all monitors.
+	"pointer" displays the OSD on the monitor containing the mouse pointer.
+	"keyboard" displays the OSD on the monitor with keyboard focus.
+	Default is "all".
+
 	*thumbnailLabelFormat* Format to be used for the thumbnail label according to *custom*
-	field below, only applied when using *<windowSwitcher style="thumbnail" />*.
+	field below, only applied when using *<osd style="thumbnail" />*.
 	Default is "%T".
 
 *<windowSwitcher><fields><field content="" width="%">*
-	Define window switcher fields when using *<windowSwitcher style="classic" />*.
+	Define window switcher fields when using *<osd style="classic" />*.
 
 	*content* defines what the field shows and can be any of:
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -77,8 +77,8 @@
     </font>
   </theme>
 
-  <windowSwitcher show="yes" style="classic" preview="yes"
-      outlines="yes" allWorkspaces="no" unshade="yes">
+  <windowSwitcher preview="yes" outlines="yes" allWorkspaces="no" unshade="yes">
+    <osd show="yes" style="classic" output="all" thumbnailLabelFormat="%T" />
     <fields>
       <field content="icon" width="5%" />
       <field content="desktop_entry_name" width="30%" />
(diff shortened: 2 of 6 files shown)


src/osd/osd.c
```diff
@@ -318,16 +332,27 @@ update_osd(struct server *server)
...
-		wl_list_for_each(output, &server->outputs, link) {
-			if (!output_is_usable(output)) {
-				continue;
-			}
-			if (!output->osd_scene.tree) {
-				osd_impl->create(output, &views);
-				assert(output->osd_scene.tree);
+		switch (rc.window_switcher.output_criteria) {
+		case OSD_OUTPUT_ALL:
+			{
```
tokyo4j: This opening brace should be put in the same line as `case`

tokyo4j:
```c
		case OSD_OUTPUT_ALL: {
			struct output *output;
				update_osd_on_output(server, output, osd_impl, &views);
			}
			break;
		}
```

src/config/rcxml.c
```diff
@@ -1199,16 +1199,28 @@ entry(xmlNode *node, char *nodename, char *content)
...
-	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.windowSwitcher")) {
+	} else if (!strcasecmp(nodename, "output.osd.windowSwitcher")) {
+		if (!strcasecmp(content, "all")) {
+			rc.window_switcher.output_criteria = OSD_OUTPUT_ALL;
+		} else if (!strcasecmp(content, "pointer")) {
+			rc.window_switcher.output_criteria = OSD_OUTPUT_POINTER;
+		} else if (!strcasecmp(content, "keyboard")) {
+			rc.window_switcher.output_criteria = OSD_OUTPUT_KEYBOARD;
+		}
+	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.osd.windowSwitcher")) {
```
tokyo4j: We must keep backward compatibility for old configurations. We can print `"*** is deprecated. use *** instead"` error message to encourage users to use new configurations, then fully migrate after some version bumps.

src/osd/osd.c
```diff
@@ -318,16 +332,27 @@ update_osd(struct server *server)
...
+			break;
+		case OSD_OUTPUT_POINTER:
+			update_osd_on_output(server,
+				output_nearest_to_cursor(server),
+				osd_impl,
+				&views);
+			break;
+		case OSD_OUTPUT_KEYBOARD:
+			update_osd_on_output(server,
+				server->active_view->output,
```
tokyo4j: This causes segfault `active_view` is NULL. Can we fall back to `pointer` in that case?

SametAylak:
@tokyo4j Thanks for the feedback! I've addressed your comments and pushed the changes. Could you take another look when you have a chance?

src/config/rcxml.c
```diff
@@ -1068,8 +1068,14 @@ entry(xmlNode *node, char *nodename, char *content)
...
 	} else if (!strcasecmp(nodename, "prefix.desktops")) {
 		xstrdup_replace(rc.workspace_config.prefix, content);
+	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.osd.windowSwitcher")) {
+		xstrdup_replace(rc.window_switcher.thumbnail_label_format, content);
+
+	/* The following one is for backward compatibility only. */
 	} else if (!strcasecmp(nodename, "thumbnailLabelFormat.windowSwitcher")) {
 		xstrdup_replace(rc.window_switcher.thumbnail_label_format, content);
+		wlr_log(WLR_ERROR, "<windowSwitcher thumbnailLabelFormat=\"\" /> is deprecated."
+			" Use <osd thumbnailLabelFormat=\"\" />");
```
Consolatis: I don't think we need backwards compatibility for this one as it was never available in a release.

docs/labwc-config.5.scd
```diff
@@ -369,6 +363,20 @@ this is for compatibility with Openbox.
...
+
+	*style* [classic|thumbnail] Configures the style of the OnScreenDisplay.
+	"classic" displays window information like icons and titles in a vertical list.
+	"thumbnail" shows window thumbnail, icon and title in grids.
+
+	*output* [all|pointer|keyboard] Configures which monitor(s) display the on-screen display (OSD).
+	"all" displays the OSD on all monitors.
+	"pointer" displays the OSD on the monitor containing the mouse pointer.
+	"keyboard" displays the OSD on the monitor with keyboard focus.
+	Default is "all"
```
Consolatis: Period at the end of the paragraph is missing.

Consolatis:
Other than the nitpicks above this LGTM. Thanks for your work on the feature!

Edit:
The commits should be squashed (for example via `git rebase -i master` and `f` in front of every commit after the first one) but we can also do that when merging.

Consolatis:
Rebased + added some logging for invalid `output` and `style` values.

docs/rc.xml.all
```diff
@@ -77,8 +77,8 @@
     </font>
   </theme>
 
-  <windowSwitcher show="yes" style="classic" preview="yes"
-      outlines="yes" allWorkspaces="no" unshade="yes">
+  <windowSwitcher preview="yes" outlines="yes" allWorkspaces="no" unshade="yes">
+    <osd show="yes" style="classic" />
```
tokyo4j: `output="all"` and `thumbnailLabelFormat="%T"` are missing. `rc.xml.all` is meant to demonstrate all the default configurations.

src/osd/osd.c
```diff
@@ -318,16 +332,31 @@ update_osd(struct server *server)
...
+				break;
 			}
-			if (!output->osd_scene.tree) {
-				osd_impl->create(output, &views);
-				assert(output->osd_scene.tree);
+		case OSD_OUTPUT_POINTER:
+			update_osd_on_output(server,
+				output_nearest_to_cursor(server),
+				osd_impl,
+				&views);
```
tokyo4j: nit: We don't usually split arguments into different lines. I suggest below
```suggestion
			update_osd_on_output(server,
				output_nearest_to_cursor(server), osd_impl, &views);
```

docs/labwc-config.5.scd
```diff
@@ -369,6 +363,20 @@ this is for compatibility with Openbox.
 	*unshade* [yes|no] Temporarily unshade windows when switching between
 	them and permanently unshade on the final selection. Default is yes.
 
+*<osd show="" style="" output="" thumbnailLabelFormat="" />*
+	*show* [yes|no] Draw the OnScreenDisplay when switching between
+	windows. Default is yes.
+
+	*style* [classic|thumbnail] Configures the style of the OnScreenDisplay.
```
tokyo4j: nit: `s/OnScreenDisplay/on-screen display/`

I know this is copied from the old documentation, but the pascal case in the sentence looks weird.

tokyo4j:
LGTM other than my nitpicks above.

I also suggest using `osd: ` prefix for the commit message. I admit that we are mixing the term "window switcher" and "osd" in our codebase and maybe we should fix that, but let's keep "osd" for now.

Consolatis:
As I've force pushed this branch earlier I'll take care of the review comments and then merge.

docs/labwc-config.5.scd
Consolatis: Changed to `OSD` for now because it would look weird to only change the single one but have other parts still refer to `OnScreenDisplay`.

Consolatis:
Note for changelog: breaking change due to movement of `show` and `style` to a `<osd>` sub-node under `<windowSwitcher>`.

PR #3205, opened 2025-11-11
xwayland: remove obsolete (since wlroots 0.17) NULL check

jlindgren90:
Since d7dc6e01b446 ("Chase wlroots: Unified mapping"), the map event is emitted by the wlr_surface itself, thus it cannot be NULL.

Change the "if" to an "assert" accordingly.


Consolatis:
Thanks!

PR #3207, opened 2025-11-12
view: factor out view_impl_init_foreign_toplevel()

jlindgren90:
Just factoring out some duplicated code between xdg.c and xwayland.c

### DIFF
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -725,35 +725,14 @@ xdg_toplevel_view_notify_tiled(struct view *view)
 	}
 }
 
-static struct view *
-lookup_view_by_xdg_toplevel(struct server *server,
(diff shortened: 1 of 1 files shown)


src/xwayland.c
```diff
@@ -192,6 +192,13 @@ xwayland_view_offer_focus(struct view *view)
 	wlr_xwayland_surface_offer_focus(xwayland_surface_from_view(view));
 }
 
+static struct view *
+xwayland_view_get_parent(struct view *view)
+{
+	struct wlr_xwayland_surface *xsurface = xwayland_surface_from_view(view);
+	return xsurface->parent ? (struct view *)xsurface->data : NULL;
```
jlindgren90: Whoops, should be `xsurface->parent->data`

Consolatis:
Thanks!

PR #3209, opened 2025-11-14
focus: add `<focusOnTitlebarEnter>` option to follow-mouse on SSD titlebars

mbroemme:
## Description
This adds a new `<focusOnTitlebarEnter>` option under the `<focus>` section of `rc.xml`. When enabled, labwc will focus a window when the pointer enters its server-side decorated titlebar - including the titlebar background, window title text, and any titlebar buttons (close, maximize, etc.).

Previously, focus-follows-mouse behavior only applied when entering the client content area, not the titlebar. This change aligns SSD focus behavior with client-side decorated windows for consistent user experience.

New configuration example:

```
    <focus>
        <followMouse>yes</followMouse>
        <followMouseRequiresMovement>no</followMouseRequiresMovement>
        <focusOnTitlebarEnter>yes</focusOnTitlebarEnter>
    </focus>
```

If this setting is left unspecified or set to "no," backward compatibility will be preserved.

## Demo Video
_Before:_
https://github.com/user-attachments/assets/f88e93c1-c333-488f-8175-1bf6ef12473b
_After:_
https://github.com/user-attachments/assets/bb8ef9ff-8dd2-4e76-a6bd-69b8659d5a7a

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -491,12 +491,6 @@ this is for compatibility with Openbox.
 *<focus><raiseOnFocus>* [yes|no]
 	Raise window to top when focused. Default is no.
 
-*<focus><focusOnTitlebarEnter>* [yes|no]
-	When *followMouse* is enabled, also focus the toplevel when the
-	pointer enters the server-side decorated titlebar (including the
-	titlebar background, the window title text, and any titlebar buttons).
-	Default is no.
-
 ## WINDOW SNAPPING
 
 Windows may be "snapped" to an edge or user-defined region of an output when
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -199,9 +199,6 @@ struct rcxml {
 	float mag_scale;
 	float mag_increment;
 	bool mag_filter;
-
-	/* Focus behavior */
-	bool focus_titlebar_on_mouse_enter;
 };
 
 extern struct rcxml rc;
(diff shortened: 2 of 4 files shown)


ahesford:
Why make this optional? It should just be standard behavior, assuming it works as expected for things like shaded windows.

mbroemme:
Simple reason to keep current backward compatible behavior, but of course I can remove the optional part.

domo141:
I've always thought not getting focus when mouse enters titlebar (with focus-follows-mouse) is accidental,
and I don't see reason to make this configurable.
I'd welcome this change - seldom, but just yesterday I had to "hassle" a bit since I did not get keyboard
focus when entering window titlebar (another window covered everything else but part of that).

ahesford:
Too many options are confusing and don't really improve user experience. To anybody but those working on compositors, the difference between a server-side titlebar and the client view is almost always immaterial. The titlebar is just part of the "window". I don't think there is any legitimate use case for not focusing when you mouse over (or click, which happens already anyway) its titlebar.

Consolatis:
I agree that we should simply focus the window if we detect movement on the titlebar, no setting necessary. The current behavior feels like an incomplete implementation. What about the borders? Or to ask the other way around: should we focus the window on any movement on a interactive server side decoration object (other than the invisible resize area)?

tokyo4j:
Note: the current behavior that cursor entering titlebar doesn't update keyboard focus is just a known regression: https://github.com/labwc/labwc/pull/2652#issuecomment-2787672703

mbroemme:
> Note: the current behavior that cursor entering titlebar doesn't update keyboard focus is just a known regression: [#2652 (comment)](https://github.com/labwc/labwc/pull/2652#issuecomment-2787672703)

Cool your #3211 fixes it as well so I believe this PR is not needed anymore.

johanmalm:
@mbroemme Thanks for catching / raising this. Yes, the SSD should be included in the `<focus><followMouse>` behaviour. Looks like 3211 is the way to go.

PR #3210, opened 2025-11-14
cursor: remove process_cursor_motion_out_of_surface()

tokyo4j:
I believe the entire if-else chain in the function can be replaced with one `wlr_scene_node_coords()`.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -464,44 +464,6 @@ update_pressed_surface(struct seat *seat, struct cursor_context *ctx)
 	return false;
 }
 
-static bool
-process_cursor_motion_out_of_surface(struct server *server,
-		double *sx, double *sy)
-{
-	struct view *view = server->seat.pressed.view;
-	struct wlr_scene_node *node = server->seat.pressed.node;
-	struct wlr_surface *surface = server->seat.pressed.surface;
-	assert(surface);
-	int lx, ly;
-
-	if (node && wlr_subsurface_try_from_wlr_surface(surface)) {
-		wlr_scene_node_coords(node, &lx, &ly);
-	} else if (view) {
-		lx = view->current.x;
-		ly = view->current.y;
-		/* Take into account invisible xdg-shell CSD borders */
-		if (view->type == LAB_XDG_SHELL_VIEW) {
-			struct wlr_xdg_surface *xdg_surface = xdg_surface_from_view(view);
-			lx -= xdg_surface->geometry.x;
-			ly -= xdg_surface->geometry.y;
-		}
-	} else if (node && wlr_layer_surface_v1_try_from_wlr_surface(surface)) {
-		wlr_scene_node_coords(node, &lx, &ly);
-#if HAVE_XWAYLAND
-	} else if (node && node->parent == server->unmanaged_tree) {
-		wlr_scene_node_coords(node, &lx, &ly);
-#endif
-	} else {
-		wlr_log(WLR_ERROR, "Can't detect surface for out-of-surface movement");
-		return false;
-	}
-
-	*sx = server->seat.cursor->x - lx;
-	*sy = server->seat.cursor->y - ly;
-
-	return true;
-}
-
 /*
  * Common logic shared by cursor_update_focus(), process_cursor_motion()
  * and cursor_axis()
@@ -536,7 +498,11 @@ cursor_update_common(struct server *server, struct cursor_context *ctx,
 			 * we can keep scrolling or selecting text even
 			 * if the cursor moves outside of the surface.
 			 */
-			return process_cursor_motion_out_of_surface(server, sx, sy);
+			int lx, ly;
+			wlr_scene_node_coords(seat->pressed.node, &lx, &ly);
+			*sx = server->seat.cursor->x - lx;
+			*sy = server->seat.cursor->y - ly;
+			return true;
 		}
 		return false;
 	}


src/input/cursor.c
```diff
@@ -464,44 +464,6 @@ update_pressed_surface(struct seat *seat, struct cursor_context *ctx)
...
-		wlr_scene_node_coords(node, &lx, &ly);
-	} else if (view) {
-		lx = view->current.x;
-		ly = view->current.y;
-		/* Take into account invisible xdg-shell CSD borders */
-		if (view->type == LAB_XDG_SHELL_VIEW) {
-			struct wlr_xdg_surface *xdg_surface = xdg_surface_from_view(view);
-			lx -= xdg_surface->geometry.x;
-			ly -= xdg_surface->geometry.y;
-		}
```
Consolatis, line 487: This seems to be a behavioral change for clients using CSD with shadows.

tokyo4j: I don't think it is. That old logic seems to be functionally identical to `wlr_scene_node_coords()`.

I tested this:
```diff
diff --git a/src/input/cursor.c b/src/input/cursor.c
index 6c12d3da..6cf6da28 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -485,6 +485,9 @@ process_cursor_motion_out_of_surface(struct server *server,
 			lx -= xdg_surface->geometry.x;
 			ly -= xdg_surface->geometry.y;
 		}
+		int x, y;
+		wlr_scene_node_coords(node, &x, &y);
+		wlr_log(WLR_ERROR, "%d,%d %d,%d", x, y, lx, ly);
 	} else if (node && wlr_layer_surface_v1_try_from_wlr_surface(surface)) {
 		wlr_scene_node_coords(node, &lx, &ly);
 #if HAVE_XWAYLAND
```

And the coordinates are identical when I perform out-of-surface motion on `gedit`:
```
00:00:10.946 [../src/input/cursor.c:490] 468,99 468,99
00:00:10.946 [../src/input/cursor.c:490] 468,99 468,99
00:00:10.986 [../src/input/cursor.c:490] 468,99 468,99
00:00:10.986 [../src/input/cursor.c:490] 468,99 468,99
```

Consolatis: I assume it doesn't use CSD shadows, the important one to test would be if the results are identical if `xdg_surface->geometry.{x,y} != 0`. Thinking about this some more I am not actually sure what the coordinates should refer to in that case though, e.g. the top left corner of the surface or it's client side defined x.y starting position.

tokyo4j: `gedit` uses CSD shadows. What the old code does is, getting the window position and subtract the shadow offset to obtain the real surface position. It can be replaced by `wlr_scene_node_coords()` that returns the position of scene-surface.

Here's my update patch and log:
```diff
diff --git a/src/input/cursor.c b/src/input/cursor.c
index 6c12d3da..a534ec20 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -485,6 +485,11 @@ process_cursor_motion_out_of_surface(struct server *server,
 			lx -= xdg_surface->geometry.x;
 			ly -= xdg_surface->geometry.y;
 		}
+		int x, y;
+		wlr_scene_node_coords(node, &x, &y);
+		wlr_log(WLR_ERROR, "wlr_scene_node_coords():%d,%d (lx,ly):%d,%d offsets:%d,%d",
+			x, y, lx, ly,
+			xdg_surface_from_view(view)->geometry.x, xdg_surface_from_view(view)->geometry.y);
 	} else if (node && wlr_layer_surface_v1_try_from_wlr_surface(surface)) {
 		wlr_scene_node_coords(node, &lx, &ly);
 #if HAVE_XWAYLAND
```

```
00:00:04.121 [../src/input/cursor.c:490] wlr_scene_node_coords():186,146 (lx,ly):186,146 offsets:23,15
```

Consolatis: Hmm, might be a subsurface then. Need to read the xdg-shell protocol again later. I am still slightly concerned that this will break some clients using margins but no subsurfaces. I vaguely remember this being hard to get right originally.

Edit:
Although subsurfaces should take the branch above. Really no clue how `wlr_scene_node_coords()` is then able to somehow fetch the xdg shell margin; maybe it actually has its own listener and in that case this PR should be fine.

tokyo4j: I think whether the surface is a xdg-surface or not doesn't really matter here. The tree returned by `wlr_scene_xdg_surface_create()` is placed at the position of the window, but its internal `surface_tree` ([ref](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/106f0f950639924d706df21f48545b09db8197ed/types/scene/xdg_shell.c#L5-13)) is placed at the position of the surface, not the window. Since `ctx.node` is always the scene-surface node for `ctx.surface` (as clarified in 9f7e532852ba199b845f6ac9c98804010fbe378e), `wlr_scene_node_coords(ctx.node)` always returns the exact position of the scene-surface rather than the "window position" in terms of xdg-shell protocol.

Note that the branch above is not taken if `ctx.surface` is a subsurface. There's `if (node && wlr_subsurface_try_from_wlr_surface(surface))` branch before that.

Consolatis: Ah, I see. In `types/scene/xdg_shell.c` `scene_xdg_surface_update_position()` the offset is already applied so `wlr_scene_node_coords()` is correct.

Consolatis:
Thanks!

PR #3211, opened 2025-11-14
cursor: update focus on entering SSD if followMouse=yes

tokyo4j:
Supersedes #3209.
Fixes the regression: https://github.com/labwc/labwc/pull/2652#issuecomment-2787672703

This PR fixes the regression that moving the cursor to window SSD (e.g. titlebar) with `followMouse=yes` doesn't give the keyboard focus to the window. The main challenge of this PR is to track the hovered window/surface in cursor motion handler so that we update the keyboard focus only when the cursor is **entering** the window, not when the cursor is just moving over the window:

https://github.com/user-attachments/assets/03170bd3-a65f-4c1d-8853-a59e26351032

The 1st commit generalizes `seat->pressed` as `struct cursor_context_saved` so that we can also have `seat->hovered` later.

The 2nd commit updates `cursor_update_common()` to take `notified` as an argument. This should not change any behaviors.

The 3rd commit finally fixes the regression. This is achieved by adding `seat->hovered` that tracks the cursor context on every motion.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -78,6 +78,9 @@ struct seat {
 	 */
 	struct cursor_context_saved pressed;
 
+	/* Cursor context of the last cursor motion */
+	struct cursor_context_saved last_cursor_ctx;
+
 	struct lab_set bound_buttons;
 
 	struct {
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -664,9 +664,6 @@ cursor_process_motion(struct server *server, uint32_t time, double *sx, double *
 		}
 	}
 
-	struct wlr_surface *old_focused_surface =
-		seat->seat->pointer_state.focused_surface;
-
 	/*
 	 * Cursor context that is actually interacting with cursor and should
 	 * be notified to the client. E.g. it is cleared when menu is open,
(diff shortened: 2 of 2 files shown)


tokyo4j:
Now ready for review.

tokyo4j:
Decided not to create `src/input/cursor-context.c`.

include/labwc.h
```diff
@@ -78,6 +78,9 @@ struct seat {
 	 */
 	struct cursor_context_saved pressed;
 
+	/* Cursor context of the last cursor motion */
+	struct cursor_context_saved hovered;
```
johanmalm: Suggest `s/hovered/prev_cursor_context/` or something like that. When reading the code further down without the hint that it's a `struct cursor_context_saved` the term hovered isn't that intuitive.

Or possibly just `prev` or `last` because most readers should get what's going on with that.

src/desktop.c
```diff
@@ -118,11 +118,10 @@ void
 desktop_focus_view_or_surface(struct seat *seat, struct view *view,
 		struct wlr_surface *surface, bool raise)
 {
-	assert(view || surface);
 	if (view) {
 		desktop_focus_view(view, raise);
 #if HAVE_XWAYLAND
-	} else {
+	} else if (surface) {
```
johanmalm: What's the logic behind adding this `if (surface)` and removing the assert?

If it's just to make it easier to read, I'm fine with it, but was just wondering if I missing some subtle point otherwise.

src/input/cursor.c
```diff
@@ -437,8 +437,70 @@ cursor_update_image(struct seat *seat)
 		cursor_names[cursor]);
 }
 
+static void
+clear_cursor_context(struct cursor_context_saved *saved_ctx)
+{
+	if (saved_ctx->node_destroy.notify) {
```
johanmalm, line 443: Suggest asserting `saved_ctx` here to avoid null de-reference.

src/input/cursor.c
```diff
@@ -437,8 +437,70 @@ cursor_update_image(struct seat *seat)
...
+cursor_context_save(struct cursor_context_saved *saved_ctx,
+		const struct cursor_context *ctx)
+{
+	clear_cursor_context(saved_ctx);
+	if (!ctx) {
+		return;
+	}
+	saved_ctx->ctx = *ctx;
+	if (ctx->node) {
+		saved_ctx->node_destroy.notify = handle_ctx_node_destroy;
```
johanmalm, line 491: Could we use `CONNECT_SIGNAL()` here?

johanmalm (commented):
Thanks. I like the refactoring with the `cursor_context_saved` struct.
Have left a few comments. Other than those, LGTM. Not tested.

include/labwc.h
tokyo4j: Updated to `last_cursor_ctx`.

src/input/cursor.c
tokyo4j: I don't think we need it in `clear_cursor_context()` because it's only called from the node/surface/view destroy handlers or `cursor_context_save()` which are defined nearby. I added `assert(saved_ctx)` in `cursor_context_save()` instead.

src/input/cursor.c
tokyo4j: We could use it if we rename the handler name to `handle_node_destroy()`, but that sounds too generic to be defined in `cursor.c`. So I'd like to keep `wl_signal_add()` here.

src/desktop.c
tokyo4j: Oh, that's just a leftover from my try and error. I removed the change.

PR #3213, opened 2025-11-16
docs/labwc-config.5.scd: clarify expectations of startup, shutdown files

ahesford:
Referring to these files as shell scripts may yield the expectation that these are piped through a shell, which would imply that these do not need to be executable or need not be fully formed scripts. In reality, labwc couldn't care less whether the file is a shell script, as long as it exists and is executable.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -62,19 +62,22 @@ as a comment. Trailing comments are not supported.
 Please note that as labwc reloads the environment file(s) on reconfigure,
 recursive/circular assignments (for example FOO=$FOO:bar) should not be made.
 
-The *autostart* file is executed as a shell script after labwc has read its
+The *autostart* file, if it exists, is executed after labwc has read its
 configuration and set variables defined in the environment file. Additionally,
 the environment variables WAYLAND_DISPLAY and (when labwc is built with Xwayland
-support) DISPLAY will be defined. This is the place for executing clients for
-handling background images, panels and other tasks that should run automatically
-when labwc launches.
+support) DISPLAY will be defined. This is a convenient place for executing
+clients for handling background images, panels and other tasks that should run
+automatically when labwc launches. While it is generally expected that the
+autostart file be a shell script, labwc does not impose any constraints other
+than that the file exist and be executable.
 
-The *shutdown* file is executed as a shell script when labwc is preparing to
+The *shutdown* file, if it exists, is executed when labwc is preparing to
 terminate itself. All environment variables, including WAYLAND_DISPLAY and
-DISPLAY, will be available to the script. However, because the script runs
-asynchronously with other termination tasks, the shutdown file should not assume
-that the display will be usable. This file is useful to perform any custom
-operations necessary to finalize a labwc session.
+DISPLAY, will be available to the executable. However, because *shutdown* runs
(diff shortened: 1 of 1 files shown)


docs/labwc-config.5.scd
```diff
@@ -62,19 +62,22 @@ as a comment. Trailing comments are not supported.
 Please note that as labwc reloads the environment file(s) on reconfigure,
 recursive/circular assignments (for example FOO=$FOO:bar) should not be made.
 
-The *autostart* file is executed as a shell script after labwc has read its
+The *autostart* file, if it exists, is executed after labwc has read its
```
johanmalm, line 65: `session_run_script()` runs `sh %s`, so I think it's better say "as a shell script"

docs/labwc-config.5.scd
```diff
@@ -62,19 +62,22 @@ as a comment. Trailing comments are not supported.
...
 configuration and set variables defined in the environment file. Additionally,
 the environment variables WAYLAND_DISPLAY and (when labwc is built with Xwayland
-support) DISPLAY will be defined. This is the place for executing clients for
-handling background images, panels and other tasks that should run automatically
-when labwc launches.
+support) DISPLAY will be defined. This is a convenient place for executing
+clients for handling background images, panels and other tasks that should run
+automatically when labwc launches. While it is generally expected that the
+autostart file be a shell script, labwc does not impose any constraints other
+than that the file exist and be executable.
```
johanmalm, line 72: I think it needs to be a shell script (as run with `sh %s`.

I does not need to be executable.

It has been a long time since I thought about the semantics of this, but feel sure it the behavior was based on those of Openbox.

`/lib/usr/openbox/openbox-autostart` my system looks like this:

```sh
#!/bin/sh

# Set a background color
BG=""
if which hsetroot >/dev/null 2>/dev/null; then
  BG=hsetroot
elif which esetroot >/dev/null 2>/dev/null; then
  BG=esetroot
elif which xsetroot >/dev/null 2>/dev/null; then
  BG=xsetroot
fi
test -z $BG || $BG -solid "#303030"

GLOBALAUTOSTART="/etc/xdg/openbox/autostart"
AUTOSTART="${XDG_CONFIG_HOME:-"$HOME/.config"}/openbox/autostart"

# Run the global openbox autostart script
if test -f $GLOBALAUTOSTART; then
    sh $GLOBALAUTOSTART
elif test -f $GLOBALAUTOSTART.sh; then
    sh $GLOBALAUTOSTART.sh
fi

# Run the user openbox autostart script
if test -f $AUTOSTART; then
    sh $AUTOSTART
elif test -f $AUTOSTART.sh; then
    sh $AUTOSTART.sh
fi

# Run the XDG autostart stuff.  These are found in /etc/xdg/autostart and
# in $HOME/.config/autostart.  This requires PyXDG to be installed.
# See openbox-xdg-autostart --help for more details.
/usr/lib/openbox/openbox-xdg-autostart "$@"
```

Note the `sh $AUTOSTART`

ahesford: Whoops, I overlooked the `sh %s` and just saw `spawn_async_no_shell`. Withdrawing.

johanmalm: Hey, no worries :smile:

PR #3214, opened 2025-11-16
[RFC] desktop: keep windows within output on scale change

johanmalm:
Submitting this on an RFC basis at the moment.

Would appreciate:

- Feedback on implementation approach
- Feedback on window position/resize behavior
- Testing

Can be tested with `WLR_WL_OUTPUTS=2 ./build/labwc`; then open a terminal and run `wlr_randr --output WL-1 --scale=2.0` (or similar).

### DIFF
--- a/include/desktop-state.h
+++ b/include/desktop-state.h
@@ -1,0 +1,31 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_DESKTOP_STATE_H
+#define LABWC_DESKTOP_STATE_H
+#include <wayland-util.h>
+
+struct server;
+struct output;
+
+struct desktop_output {
+	struct output *output;
+	double scale;
+	struct wl_array views;
+	struct wl_list link;
+};
+
+/* Helpers to remember windows and scales for each output */
+void desktop_state_create(struct server *server, struct wl_list *desktop_outputs);
+void desktop_state_destroy(struct wl_list *desktop_outputs);
+
+/**
+ * desktop_state_adjust_on_output_scale_change() - adjust views on scale change
+ * @server: server
+ * @desktop_outputs: object list containing scale and array of views
+ *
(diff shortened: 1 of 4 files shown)


jlindgren90:
It seems like this is kind of fighting against/undoing the existing logic in `view_adjust_for_layout_change()`, which calls `view_discover_output()` in order to *move* views to the "nearest" output based on absolute layout coordinates.

Would it make sense to improve the logic in `view_adjust_for_layout_change()` to prioritize keeping views on the same output, instead of adding on a sort of post-processing step after the fact?

spl237:
Hmm - this doesn't seem to work for me. I applied the patch against 0.9.2, but when I increase scaling, windows stay the same size and don't move. Is there something else I need to do to enable it?

Edit - ah, hang on - missed the bit about WL_WLR_OUTPUTS=2; I'll try again...

Hmm - no, that makes no difference - still no resizing or moving when I use wlr-randr to increase scaling from 2.0 to 3.0, or from 1.0 to 2.0.

spl237:
OK, I've put some debug into this to see what is happening. My test case is a Geany window which is taking up around 3/4 of the screen size at scaling x2 - if I change the scaling to x3, this overflows the screen, but the patch does not resize or move it.

I can see that fit_window_within_usable area is called, and it correctly calculates the usable area as 1280 x 658 (the top margin is 62). view_constrain_size_to_that_of_usable_area is called as expected, and it passes a box of the correct size and position to view_move_resize, but it looks as if the view_move_resize silently fails to do anything on the Geany window.

xdg_toplevel_view_configure is being called from the view_move_resize, and I can see that the call to wlr_xdg_toplevel_set_size is being made with appropriate values for width and height - but it doesn't seem to have an effect on the size of the Geany window.

It's not just Geany - Glade windows do the same; same trace through to the toplevel_view_configure, but no size change when requested.

johanmalm:
Fixed by #3310

PR #3215, opened 2025-11-16
view: manage view->surface lifetime via common view_set_surface()

jlindgren90:
- add new `commit()` function to `view->impl` table
- add `view_set_surface()` which encapsulates:
  - setting `view->surface`
  - connecting/disconnecting map, unmap, and commit handlers

For xwayland, `view_set_surface()` is called from `associate/dissociate()` handlers. This means that `view->surface` is now set slightly earlier, in `associate()` rather than `map()`.

It's no longer necessary to have two separate cleanup paths depending on whether the `wlr_surface` or `wlr_xwayland_surface` is destroyed first, because wlroots calls `dissociate()` beforehand in either case.

Also, `view->content_tree` is now destroyed and set to `NULL` at unmap. Previously, we relied on `wlr_scene` to destroy it automatically when the surface was destroyed, but kept a potentially dangling pointer in `view->content_tree` until next `map()`. Similar change for unmanaged.

I've tested lightly with a few xdg-shell and xwayland apps, but more testing would be appreciated, especially on the xwayland side.

P.S. I actually removed an old TODO to do exactly this cleanup in #3200, but changed my mind and decided to implement it after all.

### DIFF
--- a/src/xwayland-unmanaged.c
+++ b/src/xwayland-unmanaged.c
@@ -166,19 +166,6 @@ handle_dissociate(struct wl_listener *listener, void *data)
 	struct xwayland_unmanaged *unmanaged =
 		wl_container_of(listener, unmanaged, dissociate);
 
-	if (!unmanaged->mappable.connected) {
-		/*
-		 * In some cases wlroots fails to emit the associate event
-		 * due to an early return in xwayland_surface_associate().
-		 * This is arguably a wlroots bug, but nevertheless it
-		 * should not bring down labwc.
-		 *
-		 * TODO: Potentially remove when starting to track
-		 *       wlroots 0.18 and it got fixed upstream.
-		 */
-		wlr_log(WLR_ERROR, "dissociate received before associate");
-		return;
-	}
 	mappable_disconnect(&unmanaged->mappable);
 }
 


src/xwayland.c
```diff
@@ -349,8 +351,7 @@ handle_destroy(struct wl_listener *listener, void *data)
 	struct view *view = wl_container_of(listener, view, destroy);
 	struct xwayland_view *xwayland_view = xwayland_view_from_view(view);
 	assert(xwayland_view->xwayland_surface->data == view);
-
-	set_surface(view, NULL);
+	assert(!view->surface); /* dissociate is expected first */
```
jlindgren90: Hmm, this might not hold if called from set_override_redirect. Need to think about it more.

jlindgren90: Solved this by adding a synthetic call to handle_dissociate().

src/xwayland.c
```diff
@@ -319,28 +315,34 @@ handle_associate(struct wl_listener *listener, void *data)
...
-handle_surface_destroy(struct wl_listener *listener, void *data)
-{
-	struct view *view = wl_container_of(listener, view, surface_destroy);
-	assert(data && data == view->surface);
+	view_set_surface(view, NULL);
 
-	set_surface(view, NULL);
+	/* Also destroy content_tree when losing surface */
+	if (view->content_tree) {
+		wlr_scene_node_destroy(&view->content_tree->node);
```
jlindgren90: *sigh* this also needs more thought. The dissociate event can be emitted during wlr_surface destruction, in which case the content_tree may have already been destroyed, and we crash here. Seems to happen particularly at xwayland exit.

jlindgren90: Fixed by destroying content_tree at unmap instead.

jlindgren90:
Rebased on master, should be ready for review.

johanmalm:
Thanks

LGTM

Won't press the merge-button yet as linked to #3218 (via `view_connect_map()`) which has an ongoing discussion.

Seems a really nice cleanup.

jlindgren90:
I'll mark this as draft for now and follow up on the discussion on #3218.

jlindgren90:
Closing in favor of #3218.

PR #3216, opened 2025-11-17
view: separate (un)minimize and (un)map logic

jlindgren90:
Map/unmap logic is currently re-used for minimize/unminimize, but lots of it doesn't actually apply in that case. This is both confusing and creates some extra complexity, such as:

 - extra `client_request` parameter to `unmap()`, in which case it has to still do some cleanup even if `view->mapped` is already false
 - various `view->mapped || view->minimized` checks when we really just mean "is the view mapped"

To clean this all up, let's put the logic that really is common into new `view_on_show/on_hide()` functions, and stop using map/unmap for minimize/unminimize.

Note that this changes the meaning of `view->mapped`, which used to mean "mapped and not minimized" but now really just means "mapped". I left some `view->mapped` conditions as-is (rather than changing to `view->mapped && !view->minimized`) where it seemed to make sense.

Lightly tested, including some initially-minimized and client map/unmap cases. More testing appreciated.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -110,13 +110,7 @@ struct view_impl {
 	void (*set_activated)(struct view *view, bool activated);
 	void (*set_fullscreen)(struct view *view, bool fullscreen);
 	void (*notify_tiled)(struct view *view);
-	/*
-	 * client_request is true if the client unmapped its own
-	 * surface; false if we are just minimizing the view. The two
-	 * cases are similar but have subtle differences (e.g., when
-	 * minimizing we don't destroy the foreign toplevel handle).
-	 */
-	void (*unmap)(struct view *view, bool client_request);
+	void (*unmap)(struct view *view);
 	void (*maximize)(struct view *view, enum view_axis maximized);
 	void (*minimize)(struct view *view, bool minimize);
 	struct view *(*get_parent)(struct view *self);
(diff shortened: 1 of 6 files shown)


jlindgren90:
There will be a superficial merge conflict with #3215.

tokyo4j:
I really like your approach. The logic around `client_request` and `view->mapped` was so confusing and I'm happy to see them sorted.

Instead of exposing `view_on_show()`/`view_on_hide()`, I suggest adding `view_update_visibility()` like this:

```c
/* should be called when the view is (un)mapped/(un)minimized */
void
view_update_visibility(struct view *view)
{
	bool visible = view->mapped && !view->minimized;
	if (visible == view->scene_tree->node.enabled) {
		return;
	}
	wlr_scene_node_set_enabled(&view->scene_tree->node, visible);
	if (visible) {
		// when the view is shown
	} else {
		// when the view is hidden
	}
}
```

EDIT: Ah, https://github.com/labwc/labwc/pull/3218/commits/55709115091ee052149c9d1ee24b4828759c7eea looks similar to my suggestion. But I prefer my approach because we don't need to check `view->minimized` in map/unmap handlers and `view->mapped` in minimize handler.

jlindgren90:
`view_update_visibility()` is a good idea, I'll work on incorporating that.

tokyo4j:
LGTM. I'm not 100% sure that the reordered steps don't cause any regressions, but this seems to work fine in my quick test (with initially-minimized windows, fullscreen windows + top-layer panel, X11 panels).

PR #3217, opened 2025-11-18
wlr-foreign: set all initial states correctly

jlindgren90:
Two were missing: "minimized" and "activated".

At least "minimized" can be set before map, so the initial state needs to be read and set on the foreign-toplevel.

I think in the current code, the foreign-toplevel is always created before a view is activated, but for future-proofing, let's not rely on this, and set "activated" as well. There's no harm since wlroots optimizes away any redundant state changes.


Consolatis:
Thanks!

PR #3218, opened 2025-11-18
view: refactor map/unmap logic to reduce duplication

jlindgren90:
Trying to reduce a bit of the duplication (and accidental differences) between xwayland and xdg-shell map/unmap sequences. This is the basically the end goal I'm going for with #3215, #3216, and #3217 (which are included here).

~Put the common logic in `view_handle_map/unmap()`. Call down to `impl` functions from the common logic, not the other way around.~ Edit: this now takes a different approach as suggested by tokyo4j. See comments.
    
~Draft for now - the final form may end up a bit different, and I'm interested in initial feedback. Some risk of regression here since a few steps have been reordered.~ Edit: less reordering now that a fair amount of it was moved to #3216 instead.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -816,6 +816,8 @@ handle_map(struct wl_listener *listener, void *data)
 		}
 	}
 
+	wlr_scene_node_set_enabled(&view->content_tree->node, !view->shaded);
+
 	if (!view->been_mapped) {
 		check_natural_geometry(view);
 		set_initial_position(view, xwayland_surface);
--- a/src/view.c
+++ b/src/view.c
@@ -2518,7 +2518,11 @@ view_set_shade(struct view *view, bool shaded)
 
 	view->shaded = shaded;
 	ssd_enable_shade(view->ssd, view->shaded);
-	/* An unmapped view may not have a content tree */
+	/*
+	 * An unmapped view may not have a content tree. When the view
+	 * is mapped again, the new content tree will be hidden by the
+	 * map handler, if the view is still shaded at that point.
+	 */
 	if (view->content_tree) {
 		wlr_scene_node_set_enabled(&view->content_tree->node,
 			!view->shaded);


jlindgren90:
With `view_update_visibility()` having been added in #3216, and #3217 also merged, this PR is a bit smaller in scope/impact now, and should be OK to review & merge on its own.

tokyo4j:
I wonder if calling xdg/x11-specific functions from the common map handler is a good idea. As #3216 clarified that "(un)map" means the client (un)mapping a surface but not the compositor (un)minimizing the view, it feels more consistent to handle all the client requests (e.g. map, commit, new title, maximize-request) in xdg/x11-specific handlers first and then call common functions.

How about [this approach](https://github.com/tokyo4j/labwc/tree/map-dedup)? This keeps common functions being called from xdg/x11-specific functions. I also removed `view_impl->{map,unmap}` to clarify that functions in `view_impl` handle compositor-side actions.

johanmalm:
I think this PR looks fine, but @tokyo4j proposal does look particularly clean. There is something convoluted about `{xdg,xwayland}.c` calling `view_connect_map()` only for that function to then call the respective xdg/xwayland impl methods.

Interested to hear thoughts of @jlindgren90

jlindgren90:
https://github.com/tokyo4j/labwc/commit/829b05225b9a8e93be6cfb92c32061b8015a1ec3 (dedup’ing the foreign top-level creation) could be merged on its own, I think.

Let me think about the common->x11/xdg vs. x11/xdg->common question a bit more. One rationale for starting with common functions first is that the events are from wlr_surface, not the xdg/x11 specific surfaces.

jlindgren90:
Okay, I'm on board with @tokyo4j's approach. I've cherry-picked those two commits here and added what still made sense from #3215 on top (basically the same logical changes, except that `set_surface` remains xwayland-specific).

johanmalm:
LGTM. Not tested. I think it would be good to wait for @tokyo4j to give it a glance over though.

src/xwayland-unmanaged.c
```diff
@@ -128,7 +127,13 @@ handle_unmap(struct wl_listener *listener, void *data)
...
 	wl_list_remove(&unmanaged->link);
 	wl_list_remove(&unmanaged->set_geometry.link);
-	wlr_scene_node_set_enabled(unmanaged->node, false);
+
+	/*
+	 * Destroy the scene node. It would get destroyed later when
+	 * the wlr_surface is destroyed, but if the unmanaged surface
+	 * gets converted to a managed surface, that may be a while.
+	 */
+	wlr_scene_node_destroy(unmanaged->node);
```
tokyo4j, line 136: nit: I suggest removing the comment below (`Mark the node as gone...`) since now it's obvious that we should clear the dangling pointer.

src/view-impl-common.c
```diff
@@ -34,15 +16,19 @@ view_impl_map(struct view *view)
 	}
 
 	/*
-	 * It's tempting to just never create the foreign-toplevel handle in the
-	 * map handlers, but the app_id/title might not have been set at that
-	 * point, so it's safer to process the property here
+	 * Create foreign-toplevel handle, respecting skipTaskbar rules.
+	 * Also exclude unfocusable views (popups, floating toolbars,
+	 * etc.) as these should not be shown in taskbars/docks/etc.
```
tokyo4j, line 21: @johanmalm Do you remember why we used to always create foreign-toplevel at xdg/x11-specific map handlers and destroy it in `view_impl_map()` if `skipTaskbar` is set? This question is not blocking this PR from being merged as I confirmed `skipTaskbar` still works fine in my test, but I'm just curious.

This comment was added in #933. There used to be `view_update_title()` and `view_update_app_id()` at the start of `view_impl_map()` but I don't think it changes the return values of (now removed) `view_string_get_prop()`.

tokyo4j:
LGTM other than my trivial suggestion above. Thanks!

src/view-impl-common.c
johanmalm: No can't remember.

I think some non-Qt/Gtk clients (possibly alacrity and foot) set app_id after map, Not sure if that's related here.

IIRC I originally implemented foreign to level in a half-baked way, so might even go back to making things work then.

jlindgren90: Not sure if it was the original reason, but it used to be that some of the foreign top-level initial states were not set correctly if it was created this late. That shouldn't be the case any more since #3217.

src/xwayland.c
```diff
@@ -872,13 +846,24 @@ xwayland_view_map(struct view *view)
...
+
+	/*
+	 * Destroy the content_tree at unmap. It would get destroyed
+	 * with the wlr_surface later, but that might be a while,
+	 * especially if the surface gets converted to unmanaged.
+	 */
+	if (view->content_tree) {
+		wlr_scene_node_destroy(&view->content_tree->node);
+		view->content_tree = NULL;
+	}
```
Consolatis, line 869: Are we sure that there is no code path accidentally calling `view_set_shade()` or anything else operating on the content tree from this view? (example could be unmapping a shaded window while in the OSD)

jlindgren90: I can add a ”mapped” check to view_set_shade() to be sure.

Note that map/unmap already interacts badly with shade - you get a new content tree displayed but SSD is still in "shade" mode, so the border around the window contents is missing.

jlindgren90: I think that nullifying `view->content_tree` is still an improvement since it turns what was previously a use-after-free into a more debuggable segfault.

Consolatis: Did we actually destroy the content tree before though? AFAIR we only destroyed the whole `view->scene_tree` on view destroy but not on unmap.

jlindgren90: For client-side unmap, the `content_tree` was destroyed automatically by wlroots along with the `wlr_surface`, and we kept a dangling pointer in `view->content_tree` until the next map (or destroy).

I guess another option would be to nullify `view->content_tree` in `handle_dissociate`, since that is also called on destruction of the `wlr_surface`. But code using `view->content_tree` would still have to be null-aware, so I'm not sure it's really any better.

jlindgren90:
Updated comments and added some null checks for `view->content_tree`

src/xwayland-unmanaged.c
jlindgren90: Removed the comment

jlindgren90:
Added a fix for the case of unmap+remap while shaded

src/xwayland.c
Consolatis: Hm.. right. Looks like we have to sprinkle NULL checks everywhere we use view->content_tree then. Or we just use an own tree as `view->content_tree` and then add the surface tree under that one. That way we don't really have to care as we have full control of the tree lifetime.

jlindgren90: I've added null checks on `view->content_tree` for now. I'm not against adding a separate tree node, but I'd like to leave that change for a later PR.

jlindgren90:
Changes since the last "LGTM"s are minor, so I'll plan on merging this soon unless there are any objections.

PR #3227, opened 2025-11-24
touchpad: Register an action to a gesture

yuiiio:
```xml
    <mousebind direction="Down" fingers="4" action="swipe">
            <action name="GoToDesktop" to="right" wrap="false" />
    </mousebind>
    <mousebind direction="Up" fingers="4" action="swipe">
            <action name="GoToDesktop" to="left" wrap="false" />
    </mousebind>
```

### DIFF
--- a/include/config/mousebind.h
+++ b/include/config/mousebind.h
@@ -14,6 +14,7 @@ enum mouse_event {
 	MOUSE_ACTION_RELEASE,
 	MOUSE_ACTION_DRAG,
 	MOUSE_ACTION_SCROLL,
+	MOUSE_ACTION_SWIPE,
 };
 
 enum direction {
@@ -29,6 +30,9 @@ struct mousebind {
 
 	/* ex: BTN_LEFT, BTN_RIGHT from linux/input_event_codes.h */
 	uint32_t button;
+
+	/* used for MOUSE_ACTION_SWIPE */
+	uint32_t fingers;
 
 	/* scroll direction; considered instead of button for scroll events */
 	enum direction direction;
--- a/src/config/mousebind.c
+++ b/src/config/mousebind.c
@@ -98,6 +98,8 @@ mousebind_event_from_str(const char *str)
 		return MOUSE_ACTION_DRAG;
 	} else if (!strcasecmp(str, "scroll")) {
 		return MOUSE_ACTION_SCROLL;
+	} else if (!strcasecmp(str, "swipe")) {
+		return MOUSE_ACTION_SWIPE;
 	}
 	wlr_log(WLR_ERROR, "unknown mouse action (%s)", str);
 	return MOUSE_ACTION_NONE;
@@ -111,7 +113,8 @@ mousebind_the_same(struct mousebind *a, struct mousebind *b)
 		&& a->button == b->button
 		&& a->direction == b->direction
 		&& a->mouse_event == b->mouse_event
-		&& a->modifiers == b->modifiers;
+		&& a->modifiers == b->modifiers
+		&& a->fingers == b->fingers;
 }
 
 struct mousebind *
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -591,6 +591,9 @@ fill_mousebind(xmlNode *node, const char *context)
 		mousebind->direction = mousebind_direction_from_str(
 			buf, &mousebind->modifiers);
 	}
+	if (lab_xml_get_string(node, "fingers", buf, sizeof(buf))) {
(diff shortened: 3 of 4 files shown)


yuiiio:
that touchpad gesture feeling wasn't very good(easy to make mistakes)

yuiiio:
now having 
```
struct lab_gesturebind *gesture_binds[LAB_GESTURE_EVENT_COUNT][MAX_FINGERS];
``` 
in struct input (per devices), 
so can avoid device_name search ( strcmp with wl_list_for_each(gb, &rc.gesture_bindings) { } )
in every {gesture_binding_check, gesture_binding_match_and_run_action}  :)

Consolatis:
Thanks for the PR, I try to look into it this weekend.

One thing I noticed is that this is using `<touchpad_gesture deviceName=""><gesturebind>` for the config which seems a bit odd. Maybe something like `<touchpad>` (no device) as parent category and `<gesture gesture="swipe-right">` would be an alternative?

I'll think about the config layout until the weekend but other opinions definitely welcome.

tokyo4j:
> now having
> 
> ```
> struct lab_gesturebind *gesture_binds[LAB_GESTURE_EVENT_COUNT][MAX_FINGERS];
> ```
> 
> in struct input (per devices), so can avoid device_name search ( strcmp with wl_list_for_each(gb, &rc.gesture_bindings) { } ) in every {gesture_binding_check, gesture_binding_match_and_run_action} :)

I don't think we need it for now. We won't have gesture bindings by default, so the performance overhead doesn't matter for the default setup. If it becomes a problem and it's proved by `perf`, we can do it later.

Consolatis:
Short update: there is some discussion in the IRC channel about future proofing the config layout and letting the user decide if a binding is exclusive to labwc or should additionally be relayed to the application.

Another open question is if there should be different contexts for the gesture bindings like `Desktop`/`Root` to react to swipe to change desktops but also being able to relay swipes to the application for example firefox.

One option would thus be to simply add another argument to the existing `<mousebind>` element, similar to this earlier [proof-of-work](https://github.com/labwc/labwc/compare/master...Consolatis:labwc:feature/swipe_mousebinds) but actually respecting the mouse context. When bound to the `All` or `Frame` context the event would be labwc exclusive and shared with the focused application otherwise.

As for the device filter, I don't think we need it at all.

yuiiio:
just to be safe, I made it configurable per device, but it wasn't necessary for my use case.
I agree that simpler is better.

johanmalm:
I was going to say what @Consolatis just said above. As for touchpad binds I think it would make sense to define them as `<mousebind>` as in branch `feature/swipe_mousebind`.

@yuiiio Do you know if we can tell the difference between touchpad (=laptop trackpad) and touchscreen events?

I think that touchscreen events, it would be surprising to the user to find their bind definitions under `<mouse>`.

yuiiio:
wlr_pointer(gestures) and wlr_touch are completely different path.
wlr_touch(libilnput touchscreen api) doesn't provide gesture(swipe, pinch..) event(only tells per-finger touch_{down,up}&motion)

johanmalm:
> wlr_pointer(gestures) and wlr_touch are completely different path. wlr_touch(libilnput touchscreen api) doesn't provide gesture(swipe, pinch..) event(only tells per-finger touch_{down,up}&motion)

Thanks, have spent some time looking at it and have got my head around it. With this applying to touchpads only, I think the `<mousebind>` approach in feature/swipe_mousebind makes a lot of sense.

johanmalm:
@yuiiio I'm conscious we've given you some quite sporadic feedback on this PR. Are you happy to take it to the next level? Is there anything we can do to help?

yuiiio:
Consolatis's POC code works well on myside :)

Consolatis:
It really is just a proof-of-work though, e.g.
- its likely missing a threshold (I can sadly not test it because my only touchpad device doesn't seem to support libinput gesture detection at all)
- doesn't respect the mouse contexts (an example can be seen [here](https://github.com/labwc/labwc/blob/master/src/input/cursor.c#L1398), it also includes the check for pressed keyboard modifiers)
- on the start of a swipe / pinch we likely end up in the weird situation where we can't say if the swipe is part of a binding or not, so we would likely need to send the events to the client and if we then later discover that it was actually a consuming binding we've already sent the start of the sequence. I am not quite sure how to solve this properly. I think we'd need to store and delay sending the events until we are sure its not a consuming binding like `All` and `Frame` and only then send it to the client.
- maybe more I didn't think about

Issue #2779, opened 2025-06-01
Applications using fullscreen with a smaller geometry than the output geometry are misplaced


Consolatis:
We should center the window in those cases and put a solid background behind everything. That is also what the spec requires.

>> Just a slightly related note: The reason the smaller surface is shown on the top-left and the other windows being visible is due to labwc being [non-spec compliant](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml?ref_type=heads#L1078): the surface should be centered on the fullscreen output in case the dimensions are smaller than the full output geometry. We should also add a solid background behind it, covering the full output. Technically we are also supposed to put a solid background behind any fullscreen surface if the surface is not completely opaque (which I am absolutely not a fan of). Ref: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/116
> 
> This explains the strange behavior that happens with labwc and winewayland.drv.
> 
> This is what a game through wine wayland in labwc currently looks like:
> ![Image](https://gitlab.winehq.org/-/project/5/uploads/60010f1bf0218f93e5455081d04c3840/screenshot_05-03-25_02-03-51.png)
> 
> ...and this is what it's supposed to look like (wine xwayland in labwc):
> ![Image](https://gitlab.winehq.org/-/project/5/uploads/48d2fee421f47205d6d2565686f8fc59/screenshot_05-03-25_02-02-46.png)
> 
> Basically, it's impossible to play old games using wine wayland on labwc. 

 _Originally posted by @danielfrrrr in [#2774](https://github.com/labwc/labwc/issues/2774#issuecomment-2927870760)_

qrp73:
I'm not sure that centering will help much. And it happens for native Linux games using SDL2 library either. When you run the game running with 640x480 resolution with default SDL x11 driver you will get this:

<img width="1280" height="1024" alt="Image" src="https://github.com/user-attachments/assets/ab8697a1-97f8-42a1-99af-09c0b1462b64" />



But if you run it with ```SDL_VIDEODRIVER=wayland quake3``` it works as expected:

<img width="1280" height="1024" alt="Image" src="https://github.com/user-attachments/assets/30edac2f-5348-4667-8e0e-a470f347c11c" />

So, it seems like some issue with display mode set or scaling for x11 driver.


[...]

PR #3228, opened 2025-11-24
[RFC] center fullscreen surface if smaller than output

johanmalm:
Quick implementation to try to move the conversation on. There might be more elegant and sensible ways to implement this.

Tested with https://github.com/tokyo4j/wayland-demo/tree/small-fullscreen-window

Only implemented for xdg-shell (not XWayland) as I only had the test client above.

In response to #2779

### DIFF
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -3,6 +3,7 @@
 #include "view-impl-common.h"
 #include "foreign-toplevel/foreign.h"
 #include "labwc.h"
+#include "output.h"
 #include "view.h"
 #include "window-rules.h"
 
@@ -74,6 +75,22 @@ resizing_edge(struct view *view, enum lab_edge edge)
 		&& (server->resize_edges & edge);
 }
 
+static void
+center_fullscreen_view(struct view *view, int w, int h)
+{
+	int x = 0, y = 0;
+	if (view_compute_centered_position(view, NULL, w, h, &x, &y)) {
+		view_move(view, x, y);
+	}
+}
+
+static bool
+surface_is_smaller_than_output(struct view *view, int w, int h)
+{
+	struct wlr_output *output = view->output->wlr_output;
+	return w < output->width || h < output->height;
+}
+
 void
 view_impl_apply_geometry(struct view *view, int w, int h)
 {
@@ -117,5 +134,15 @@ view_impl_apply_geometry(struct view *view, int w, int h)
 
 	if (!wlr_box_equal(current, &old)) {
 		view_moved(view);
+	}
+
+	/*
+	 * Centre fullscreen windows if smaller than output, for example Wine
+	 * game SWAT4. See issue #2779
+	 */
+	if (view->fullscreen && !view->current.x && !view->current.y) {
+		if (surface_is_smaller_than_output(view, w, h)) {
+			center_fullscreen_view(view, w, h);
+		}
 	}
 }
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -133,24 +133,6 @@ do_late_positioning(struct view *view)
 static void set_pending_configure_serial(struct view *view, uint32_t serial);
 
 static void
-center_fullscreen_view(struct view *view, struct wlr_box *size)
-{
-	int x = 0, y = 0;
-	if (view_compute_centered_position(view, NULL, size->width, size->height, &x, &y)) {
-		view_move(view, x, y);
-	}
-}
-
-static bool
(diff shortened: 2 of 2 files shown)


johanmalm:
Should probably only try to center if fullscreen window is smaller and at (0,0) to save cycles.

tokyo4j:
Does it restore the natural geometry when the window is un-fullscreened?

jlindgren90:
Thanks for working on this. I think I have seen the issue on XWayland too, for example Cities Skylines with the game resolution set smaller than the output.

Thinking out loud, would the logic fit well in `view_impl_apply_geometry()`? That has the convenience of already been called from both commit handlers (xdg and xwayland) on surface size change.

johanmalm:
> Does it restore the natural geometry when the window is un-fullscreened?

No. 

> Thinking out loud, would the logic fit well in `view_impl_apply_geometry()`? That has the convenience of already been called from both commit handlers (xdg and xwayland) on surface size change.

Yes, that does indeed sound like a good idea.

johanmalm:
Moved it to `view_impl_appy_geometry()`

@jlindgren90 Any chance of testing with Cities Skylines - or some other XWayland client with a small surface in fullscreen mode?

This is a simple implementation which does **not**:

1. Restore natural geometry on un-fullscreen. In other words, it leaves the window centered.
2. Add a black surface behind the small fullscreen window.

Considering that this is an unusual case, maybe simplicity in functionality is better than complexity in code here?

jlindgren90:
I'll try to test soon (next day or two).

> Restore natural geometry on un-fullscreen. In other words, it leaves the window centered.

I would have expected the existing logic on leaving full screen to handle this. Any idea why it doesn't?

jlindgren90:
> Any chance of testing with Cities Skylines - or some other XWayland client with a small surface in fullscreen mode?

I think the issue with Cities Skylines is on the client side actually, after re-testing. `xwininfo` shows the window as having the same resolution as the output (2560x1440) even though the game only renders in the lower left corner at 1920x1080 (the rest of the window is black).

jlindgren90:
@johanmalm how about this:
https://github.com/labwc/labwc/compare/master...jlindgren90:labwc:center-fullscreen

After trying a few different things, I think that for xdg-shell views, it makes sense to set `view->pending` directly in the commit handler rather than going through `view_move() -> xdg_toplevel_view_configure()`. No new configure event is going to be sent to the client. It also avoids flicker if we center the view immediately when the size changes, without waiting for `!view->pending_configure_serial` as `xdg_toplevel_view_configure()` does.

I'm not sure what would make the most sense for an xwayland client. There I'd be nervous about racing with an xwayland client trying to set its own position on entering fullscreen. It might be safest to make the change apply only to xdg-shell until we have an xwayland test case.

Regarding restoring the natural geometry on leaving fullscreen, I think it still works fine with the existing logic? With @tokyo4j 's test client, the usual dance of sending a configure request 0x0 to get the desired size applies, and the client doesn't actually change size, but I can see it shift from being centered in the full output (when fullscreen) to being centered in the usable area (when not fullscreen).

Consolatis:
Just thinking out loud: we could add a wlr_scene_rect to `view->scene_tree` which is usually disabled. Then when we detect the size being smaller than the output geometry on fullscreen we could enable the new rect and set its size to the output resolution and keep the window (e.g. `view->scene_tree`) at 0,0 of the output but instead move `view->content_tree` so its centered within the parent `view->scene_tree`. That way we don't change any pending or current geometry.

jlindgren90:
> That way we don't change any pending or current geometry.

Just to this specific point, I think there is a fair amount of code that assumes `view->pending` and `view->current` match the client/content geometry. Maybe less so in xdg-shell than xwayland. But I'd be nervous about breaking that assumption.

SSD scene nodes are already positioned at negative coordinates relative to `view->scene_tree`. That could work for the solid black fill rect as well.

mahkoh:
FYI when @danielfrrrr reported this issue to me I solved it by splitting the xdg_surface geometry in two parts: The geometry set by the client (possibly the implicit geometry) and the effective geometry used by the compositor. For non-fullscreen surfaces the effective geometry is identical to the geometry by the client. For fullscreen surfaces the effective geometry has the size of the output and is calculated such that the geometry set by the client is centered within in. The rest of the compositor always uses the effective geometry. This obviously works because for the rest of the compositor it is indistinguishable from the client having set the correct fullscreen geometry itself. The only outside change that was necessary was informing the xdg_surface when it enters or leaves fullscreen.

Consolatis:
Superseded  by #3233

PR #3229, opened 2025-11-25
[Feature] window-switcher: add `order` parameter to `<windowSwitcher>` to switch between MRU and opened

mbroemme:
## Description
Add a new configuration option to control the window switcher traversal order. MRU cycling is convenient for quick toggling, but some users - me as well - prefer a stable taskbar-like order. This change introduces a choice while keeping MRU as the default and aligns the visual highlight with actual focus/stacking to avoid surprises.

New switcher order:
- `order="focus"` (default): cycle by most-recently-used focus order.
- `order="age"`: cycle by window creation/taskbar order.

### DIFF
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1238,6 +1238,9 @@ entry(xmlNode *node, char *nodename, char *content)
 			rc.window_switcher.order = WINDOW_SWITCHER_ORDER_FOCUS;
 		} else if (!strcasecmp(content, "age")) {
 			rc.window_switcher.order = WINDOW_SWITCHER_ORDER_AGE;
+		} else {
+			wlr_log(WLR_ERROR, "Invalid windowSwitcher order %s: "
+				"should be one of focus|age", content);
 		}
 
 	/* The following two are for backward compatibility only. */


johanmalm:
IIRC the taskbar order isn't consistent. I can't remember why, but it's not labwc related. 

I might be wrong on this; memory might be failing me.

If you open two taskbars at different times (with different number of clients open) that should confirm it.

src/osd/osd.c
```diff
@@ -71,23 +71,66 @@ static struct view *
 get_next_cycle_view(struct server *server, struct view *start_view,
 		enum lab_cycle_dir dir)
 {
+	bool forwards = dir == LAB_CYCLE_DIR_FORWARD;
+	if (rc.window_switcher.order == WINDOW_SWITCHER_ORDER_OPENED) {
+		/* Build a stable list of candidates in open (creation) order */
+		struct wl_array views; wl_array_init(&views);
```
Consolatis: Please use two lines here, we don't usually have two statements within the same line.

src/osd/osd.c
```diff
@@ -71,23 +71,66 @@ static struct view *
...
+					break;
+				}
+			}
+			if (idx == -1) {
+				idx = 0; /* fallback */
+			}
+		}
+
+		ssize_t next = forwards ? (idx + 1) % (ssize_t)n :
+			(idx - 1 + (ssize_t)n) % (ssize_t)n;
```
Consolatis: Can you reformat this like so?
```c
ssize_t next = forwards
	? (idx + 1) % (ssize_t)n
	: (idx - 1 + (ssize_t)n) % (ssize_t)n;
```

src/osd/osd.c
mbroemme: Oops, will do. :)

mbroemme:
@Consolatis code has been fixed and commits squashed.

docs/labwc-config.5.scd
```diff
@@ -363,6 +363,33 @@ this is for compatibility with Openbox.
...
+
+	- Pressing Alt+Tab forward moves to the next-most-recent window (the
+	  classic behavior: you usually toggle between the last two, then
+	  further back in time).
+
+	- If you focus a window, it jumps to the front of this list
+	  immediately.
+
+	- Closing or minimizing a window removes it from the MRU list until it's
+	  shown/focused again.
```
Consolatis: Sorry, missed this before: it reads like minimizing a window would remove it from the list completely which is not the case

tokyo4j:
The idea sounds good to me. Openbox has `linear` argument in `NextWindow` which behaves similarly: https://openbox.org/help/Actions#NextWindow. I'm afraid `mru` is too technical, I suggest `focused` instead. Note that we currently sort the windows in window switcher in stacking-order, but I think that's due to our implementation issue and we can instead sort them by focus-order later.

This PR currently has a problem that window order shown in OSD isn't consistent, because the stacking-ordered array of windows is passed to `osd_impl->create()`.

I think this is a good opportunity for us to refactor the window switcher code so that we create a fixed list of windows at the start of window switching and navigate through it, instead using `view_next_no_head_stop()` and `view_prev_no_head_stop()`. Let's merge #3236 first.

docs/labwc-config.5.scd
```diff
@@ -363,6 +363,33 @@ this is for compatibility with Openbox.
 	*unshade* [yes|no] Temporarily unshade windows when switching between
 	them and permanently unshade on the final selection. Default is yes.
 
+	*order* [mru|opened] Cycle windows behavior.
+	"mru" cycles by focus history. The list is sorted by the last time each
+	window had focus—most recently focused first.
+
+	- Pressing Alt+Tab forward moves to the next-most-recent window (the
```
tokyo4j: We usually avoid "Alt+Tab" to describe window switcher because that's just one of the default keybindings. I suggest saying "window switcher" or "window switching" instead. If you think saying "Alt-Tab" makes it easier to understand for new users, you can something like "window switcher (started with alt-tab by default)".

docs/labwc-config.5.scd
mbroemme: Oh yes, you're right. I had added it before for minimized windows, but figured out that it will be hard to show it again without any taskbar to click on. I will remove the sentence completely as it is obvious for closed windows.

Consolatis:
> I'm afraid `mru` is too technical, I suggest `focused` instead.

What about something like `linear` and `focus-history`?

mbroemme:
> This PR currently has a problem that window order shown in OSD isn't consistent, because the stacking-ordered array of windows is passed to `osd_impl->create()`.

Oops, forgot to test it. @tokyo4j Should I fix it or wait for #3235 and #3236 finalization?

mbroemme:
> > I'm afraid `mru` is too technical, I suggest `focused` instead.
> 
> What about something like `linear` and `focus-history`?

Maybe `linear` and `recent`?

docs/labwc-config.5.scd
mbroemme: Thank you very much, rephrased.

johanmalm:
In terms of scope, I'm happy with this. I think it's a nice addition. 

Have not looked at code yet.

In terms of terminology, I'm not fond of `linear` and `recent`.

I'd be okay with `linear` on the basis that it's consistent with Openbox though.

If verbosity was not a consideration, I'd suggest `by-{window,client}-age` or `by-duration-since-keyboard-focus`. For the sake of simplicity I'd vote for `age` and `focus`.

mbroemme:
Actually Openbox does it even a bit different. It has an option `linear` which can be `[yes|no]`. Maybe this is a better way, instead of using `order`. It would be even more aligned with other options like `show`, `preview`, `outlines` and `allWorkspaces` in `<windowSwitcher>`. What do you think?

Consolatis:
That would make the config format more weird when offering other options in the future (outside of linear or history) but that might be acceptable (I don't think there are many more potential sort orders anyway). Thus `linear="yes|no"` would be fine by me but lets give others a chance to voice their opposition.

tokyo4j:
> Should I fix it or wait for https://github.com/labwc/labwc/pull/3235 and https://github.com/labwc/labwc/pull/3236 finalization?

Sorry for ignoring your comment. As #3236 is merged, now you can sort the window list by the creation age. You can refer to my example (sorting windows by their title) in that PR.

I'm not sure about the naming... I'm not a fan of `linear="yes|no"` because "linear" is just a mathematical term for me. If we go with that approach, I would prefer `fixed="yes|no"`.

But for extensibility, I prefer `order="creation|focus"` or `order="age|focus"` (suggested by @johanmalm).

Consolatis:
Lets go with `age|focus` then.

mbroemme:
@tokyo4j I've updated and rebased it on the latest master with #3236 merged. It's much cleaner now since I was able to remove the custom sorting. Please review and test - it's working well on my end.

docs/labwc-config.5.scd
```diff
@@ -363,6 +363,32 @@ this is for compatibility with Openbox.
...
+
+	- Using the window switcher (started with Alt+Tab by default) moves to
+	  the next window in that fixed open order; backward goes the other
+	  way.
+
+	- The order doesn't change when you focus a window, so repeated cycles
+	  are predictable.
+
+	Default is "focus".
+
```
tokyo4j: This looks too verbose. I suggest more simpler documentation like below that fits in one paragraph and aligns with other settings in `<windowSwitcher>` (I'm a non-native English speaker and feel free to modify it). Sorry for not saying this earlier.

```suggestion
	*order* [focus|age] The order in which windows are cycled. *focus* cycles by
	recent focus history, starting with the previously focused window. *age* cycles
	by creation/open order, a stable taskbar-style ordering that doesn’t change on
	focus. Default is *focus*.
```

include/view.h
```diff
@@ -174,6 +174,7 @@ struct view {
 
 	bool mapped;
 	bool been_mapped;
+	uint64_t creation_iid;
```
tokyo4j: nit: I don't know what "iid" means, and I suggest `creation_id` instead. Perhaps it's just I'm too bad at English, though.

src/config/rcxml.c
```diff
@@ -1233,6 +1233,12 @@ entry(xmlNode *node, char *nodename, char *content)
 			wlr_log(WLR_ERROR, "Invalid windowSwitcher output %s: "
 				"should be one of all|focused|cursor", content);
 		}
+	} else if (!strcasecmp(nodename, "order.windowSwitcher")) {
+		if (!strcasecmp(content, "focus")) {
+			rc.window_switcher.order = WINDOW_SWITCHER_ORDER_FOCUS;
+		} else if (!strcasecmp(content, "age")) {
+			rc.window_switcher.order = WINDOW_SWITCHER_ORDER_AGE;
+		}
```
tokyo4j, line 1244: I suggest printing error when any string other than `focus` and `age` is passed.

tokyo4j:
LGTM other than my comments above. Tested and worked well. Thanks!

include/view.h
Consolatis: I've stumbled over this as well. +1 for the change to `creation_id`. Same for `server->next_view_creation_id`.

mbroemme: It was named as `increment` id, but you're right over-complicates naming. Will change it.

docs/labwc-config.5.scd
mbroemme: Thanks for the feedback @tokyo4j much appreciated and changed!

docs/labwc-config.5.scd
```diff
@@ -363,6 +363,10 @@ this is for compatibility with Openbox.
 	*unshade* [yes|no] Temporarily unshade windows when switching between
 	them and permanently unshade on the final selection. Default is yes.
 
+	*order* [focus|age] The order in which windows are cycled. *focus* cycles by
+	recent focus history, starting with the previously focused window. *age* cycles
+	by creation/open order, a stable taskbar-style ordering that doesn’t change on
+	focus. Default is *focus*.
```
tokyo4j, line 369: Please add an empty line below the paragraph for consistency.

mbroemme: Of course, thanks and done!

src/config/rcxml.c
mbroemme: Added!

include/view.h
mbroemme: Renamed:
- `creation_iid` -> `creation_id`
- `next_view_creation_iid` -> `next_view_creation_id`

mbroemme:
All requested changes are done. I can squash the commits if needed, but I recommend keeping them for future context and discussion.

Consolatis:
> All requested changes are done. I can squash the commits if needed, but I recommend keeping them for future context and discussion.

I've squashed the commits into one to keep the history focused on single features. Thanks a lot for your flexibility and great teamwork!

Bypassing cooldown as mentioned in https://github.com/labwc/labwc/discussions/3226#discussioncomment-15179871.

PR #3230, opened 2025-11-25
output: avoid use of wlr_scene_output.WLR_PRIVATE.index

jlindgren90:
We were only using it to allow quick bitset comparisons of sets of outputs (such as `view->outputs`). We can maintain our own bit IDs for this purpose and avoid using the private wlroots field.

Note: from my reading of `wlr_scene_output_create()`, it appears to always take the lowest unused index, resulting in aggressive re-use of index values when outputs are disconnected and reconnected. I've tried to make re-use as infrequent as possible. This could theoretically reduce the chance of a mix-up in `view_update_outputs()`, although I'm not aware of any practical scenario where it matters.

Lightly tested running nested with `WLR_WL_OUTPUTS=2`.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -257,6 +257,7 @@ struct server {
 	struct wl_list outputs;
 	struct wl_listener new_output;
 	struct wlr_output_layout *output_layout;
+	uint64_t next_output_id_bit;
 
 	struct wl_listener output_layout_change;
 	struct wlr_output_manager_v1 *output_manager;
--- a/include/output.h
+++ b/include/output.h
@@ -33,6 +33,18 @@ struct output {
 	struct wl_listener frame;
 	struct wl_listener request_state;
 
+	/*
+	 * Unique power-of-two ID used in bitsets such as view->outputs.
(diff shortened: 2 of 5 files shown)


jlindgren90:
`WLR_WL_OUTPUTS=65 labwc` is fun.

```
labwc: types/scene/wlr_scene.c:1675: wlr_scene_output_create: Assertion `scene_output->index < 64' failed.
```

Consolatis:
> `WLR_WL_OUTPUTS=65 labwc` is fun.
> 
> ```
> labwc: types/scene/wlr_scene.c:1675: wlr_scene_output_create: Assertion `scene_output->index < 64' failed.
> ```

Right, it is a wlroots limitation with the bitset design which should arguably not be an assert but simply prevent adding more wlr_scene_outputs.

jlindgren90:
> Right, it is a wlroots limitation with the bitset design which should arguably not be an assert but simply prevent adding more wlr_scene_outputs.

I suppose we might as well limit it on the compositor side to avoid hitting the assert. I pushed a small update to do that.

Consolatis:
Thanks!

PR #3233, opened 2025-11-26
Center fullscreen views and add black background fill

jlindgren90:
Building on discussions & previous prototypes in:
- #2779
- #3228

Tested with https://github.com/tokyo4j/wayland-demo/tree/small-fullscreen-window and seems to work well for that case.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -1741,6 +1741,12 @@ view_set_fullscreen(struct view *view, bool fullscreen)
 		view_apply_special_geometry(view);
 	}
 	output_set_has_fullscreen_view(view->output, view->fullscreen);
+	/*
+	 * Entering/leaving fullscreen might result in a different
+	 * scene node ending up under the cursor even if view_moved()
+	 * isn't called. Update cursor focus explicitly for that case.
+	 */
+	cursor_update_focus(view->server);
 }
 
 static bool
--- a/src/xdg.c
+++ b/src/xdg.c
@@ -10,6 +10,7 @@
 #include <wlr/types/wlr_xdg_toplevel_icon_v1.h>
 #include "buffer.h"
 #include "common/array.h"
+#include "common/box.h"
 #include "common/macros.h"
 #include "common/mem.h"
 #include "config/rcxml.h"
@@ -129,6 +130,58 @@ do_late_positioning(struct view *view)
 	}
 }
 
+static void
+disable_fullscreen_bg(struct view *view)
+{
+	struct xdg_toplevel_view *xdg_view = xdg_toplevel_view_from_view(view);
+	if (xdg_view->fullscreen_bg) {
+		wlr_scene_node_set_enabled(&xdg_view->fullscreen_bg->node, false);
+	}
+}
+
+/*
+ * Centers any fullscreen view smaller than the full output size.
+ * This should be called immediately before view_moved().
+ */
+static void
+center_fullscreen_if_needed(struct view *view)
+{
+	if (!view->fullscreen || !output_is_usable(view->output)) {
+		disable_fullscreen_bg(view);
+		return;
+	}
+
+	struct wlr_box output_box = {0};
+	wlr_output_layout_get_box(view->server->output_layout,
+		view->output->wlr_output, &output_box);
+	box_center(view->current.width, view->current.height, &output_box,
+		&output_box, &view->current.x, &view->current.y);
+
+	/* Reset pending x/y to computed position also */
+	view->pending.x = view->current.x;
+	view->pending.y = view->current.y;
+
+	if (view->current.width >= output_box.width
+			&& view->current.width >= output_box.height) {
+		disable_fullscreen_bg(view);
+		return;
+	}
+
+	struct xdg_toplevel_view *xdg_view = xdg_toplevel_view_from_view(view);
+	if (!xdg_view->fullscreen_bg) {
+		const float black[4] = {0, 0, 0, 1};
+		xdg_view->fullscreen_bg =
+			wlr_scene_rect_create(view->scene_tree, 0, 0, black);
+		wlr_scene_node_lower_to_bottom(&xdg_view->fullscreen_bg->node);
+	}
+
+	wlr_scene_node_set_position(&xdg_view->fullscreen_bg->node,
+		output_box.x - view->current.x, output_box.y - view->current.y);
+	wlr_scene_rect_set_size(xdg_view->fullscreen_bg,
+		output_box.width, output_box.height);
+	wlr_scene_node_set_enabled(&xdg_view->fullscreen_bg->node, true);
+}
+
 /* TODO: reorder so this forward declaration isn't needed */
 static void set_pending_configure_serial(struct view *view, uint32_t serial);
 
@@ -238,6 +291,7 @@ handle_commit(struct wl_listener *listener, void *data)
 
 	if (update_required) {
 		view_impl_apply_geometry(view, size.width, size.height);
+		center_fullscreen_if_needed(view);
 		view_moved(view);
 
 		/*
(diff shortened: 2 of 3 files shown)


Consolatis:
It looks like this will break partly transparent fullscreen windows, even if they are the same dimensions as the output. While this is spec compliant I'd argue it is just an unnecessary limitation for users.

jlindgren90:
I guess I'm not sure what the desired behavior would be for partly transparent fullscreen views. Is it desirable for views underneath to show through?

I would be inclined to say no, but I guess I've never seen a fullscreen application use transparency in the wild, so I'm not sure what the use case would be.

Consolatis:
I think the use-case for that would mostly be ad-hoc terminal emulators and application launchers. Some discussion around that topic is [here](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/116).

jlindgren90:
I'm also unsure what we would want to do for a view that is just being slow to resize. The approach here blanks out other views immediately, and centers the view on first commit (or configure timeout, whichever comes first).

I could change it to wait and center + fill background only after there is no pending configure, and only if the view is too small, maybe that would be better?

jlindgren90:
> I could change it to wait and center + fill background only after there is no pending configure, and only if the view is too small, maybe that would be better?

The latest commit does more or less this, please take a look.

src/output.c
```diff
@@ -1083,6 +1083,23 @@ output_update_all_usable_areas(struct server *server, bool layout_changed)
...
+	wlr_output_effective_resolution(output->wlr_output,
+		&box.width, &box.height);
+	double ox = 0, oy = 0;
+	wlr_output_layout_output_coords(output->server->output_layout,
+		output->wlr_output, &ox, &oy);
+	box.x -= ox;
+	box.y -= oy;
+	return box;
+}
+
```
tokyo4j: We can just use `wlr_output_layout_get_box()` instead.

jlindgren90: Good catch, thanks.

jlindgren90:
- Fixed a race condition with `handle_configure_timeout()`
- Changed to use `wlr_output_layout_get_box()`
- Rebased

Consolatis:
Thanks to anyone involved. If it is confirmed that the latest state of this PR fixes the remaining issues in #2779 this LGTM. Not tested.

jlindgren90:
It seems there's *still* a race with `handle_configure_timeout()`, specifically if the size of the view doesn't change when going fullscreen. I'll spend some time thinking about how to solve this.

jlindgren90:
After a couple more iterations, this seems to be working well as confirmed in #2779.
I squashed some commits to clean up the history.

src/xdg.c
```diff
@@ -237,8 +264,19 @@ handle_commit(struct wl_listener *listener, void *data)
...
+		 * size. This call updates view->pending directly.
+		 */
+		bool centered_fullscreen = center_fullscreen_if_needed(view,
+			size.width, size.height);
+
+		/* Updates view->current and calls view_moved() if needed */
 		view_impl_apply_geometry(view, size.width, size.height);
 
+		/* Enable background fill if needed (uses view->current) */
+		view_set_fullscreen_bg_enabled(view, centered_fullscreen);
```
tokyo4j: Did you split this into `center_fullscreen_if_needed()` and `view_set_fullscreen_bg_enabled()` just because the background is a child of view's tree and `view_impl_apply_geometry()` updates its position? If so, I suggest making it a sibling of the view's tree instead and merging these function calls.

Also, I think we should not add functions in `view.c` if adding background behind fullscreen windows is exlusive to xdg-shell. We can also move `fullscreen_bg` into `xdg_toplevel_view`.

jlindgren90: Sounds reasonable. Do you mind putting your suggestion into code, to be clear?

We just have to take some care with where pending vs. current are updated, it is a bit intricate. And we need to make sure the background rect stays with the output when the view moves, that's why I had an update in view_moved, but maybe another place would work as well.

jlindgren90: This branch has an alternative approach (and a few more comments): https://github.com/jlindgren90/labwc/tree/center-fullscreen-alt

I'm not sure that I like it any better, but it does keep the logic mostly contained to xdg.c.

tokyo4j: Sorry, I couldn't understand what your alternative version does at a first glance. And sorry for my late reply.

Is there any downside of this? https://github.com/labwc/labwc/compare/master...tokyo4j:labwc:center-fullscreen

As commit handling and configure timeout handling are already a mess (inevitably), I'd like to keep this hack into one place and keep the added complexity as small as possible.

jlindgren90: > Is there any downside of this?

As I mentioned in code comments, the reason for making the background rect a child of the view's scene_tree is so that the stacking order remains correct automatically. With your approach, the background doesn't minimize with the view, close with the view, etc. I don't think wlr_scene has a way to say "position node A relative to node B but stack it relative to node C", so we have to manage either the stacking or positioning explicitly.

Edit: or there could be other approaches such as centering only the `content_tree` while leaving the `scene_tree` aligned with the output. I think @Consolatis suggested something similar to that.

jlindgren90: > Sorry, I couldn't understand what your alternative version does at a first glance.

Logically it does pretty much the same thing as the version here in the PR. It's just refactored so as to avoid modifying common `view` logic. For example, instead of adding a hook inside `view_moved()` to update the background rect position, it adds a xdg-specific wrapper around `view_moved()`.

jlindgren90: I worked on it some more and pushed a substantially simplified version to this PR.

jlindgren90:
Substantially refactored and simplified. Most of the logic has been consolidated to the single, xdg-specific `center_fullscreen_if_needed()` function now.

johanmalm:
LGTM. Not tested.

danielfikko:
I just want to say that I tested your latest changes and everything is working!

jlindgren90:
Thanks. I'll wait for comment from @tokyo4j, but hopefully the latest iteration addresses the remaining concerns, and we can go with it.

src/view.c
```diff
@@ -1762,6 +1741,12 @@ view_set_fullscreen(struct view *view, bool fullscreen)
 		view_apply_special_geometry(view);
 	}
 	output_set_has_fullscreen_view(view->output, view->fullscreen);
+	/*
+	 * Entering/leaving fullscreen might result in a different
+	 * scene node ending up under the cursor even if view_moved()
+	 * isn't called. Update cursor focus explicitly for that case.
+	 */
```
tokyo4j, line 1748: If this is currently only for the fullscreen background, I think we should explicitly mention that.

```suggestion
	/*
	 * Leaving fullscreen might hide the background fill for xdg views and result in a different
	 * scene node ending up under the cursor even if view_moved()
	 * isn't called. Update cursor focus explicitly for that case.
	 */
```

src/xdg.c
```diff
@@ -785,6 +852,9 @@ handle_map(struct wl_listener *listener, void *data)
 			set_initial_position(view);
 		}
 
+		/* Disable background fill at map (paranoid?) */
+		disable_fullscreen_bg(view);
+
```
tokyo4j, line 857: I suggest removing it for now. I'm not 100% sure this isn't necessary either, but I guess `center_fullscreen_if_needed()` in `handle_commit()` is sufficient.

tokyo4j:
Thanks, I like the updated version. LGTM other than minor suggestions above.

src/xdg.c
jlindgren90: I'd prefer to leave it unless we can prove it's not needed. If not needed (bg already disabled) then it does no harm.

Map for xdg-shell is weird since it's actually part of commit. I suspect the view_moved() in map is also redundant, but haven't worked through it to be sure.

src/view.c
jlindgren90: I think we need this also in case the cursor is over the SSD and view doesn't move (however unlikely that is).

src/xdg.c
tokyo4j: Hm, okay. Let's keep it for now.

src/view.c
tokyo4j: Alright. Ignore my suggestion.

PR #3235, opened 2025-11-26
 Use the word "cycle" instead of "osd" across the entire codebase 

tokyo4j:
First round of #3236.

We've been using the word "osd" for window switcher in general. But this can be confusing for readers, especially new contributors. Let's call it "switcher" instead and use "switcher OSD" for window switching OSD to distinguish it from the workspace indicator. 

One concern is that it can be still confusing as we also have `osd.workspace-switcher` theme setting, but I think the word "window switcher" is too long and wastes the line width too much. Xfce4 calls it "tabwin", KWin calls it "tabbox" and Openbox calls it "focus cycle" in their codebase, but they are not common.

I split this into two commits to let `git` track the file renames.


jlindgren90:
I don't mind the change, but I'm not sure I would understand when to use just `switcher` vs. `switcher_osd` (as I see you are using both).

tokyo4j:
> I don't mind the change, but I'm not sure I would understand when to use just switcher vs. switcher_osd (as I see you are using both).

"Switcher OSD" refers to the OSD shown when window switcher is active. "Switcher" refers to window switcher itself, which can happen without OSD if `<windowSwitcher><osd show="no">`.

Consolatis:
Personally I would prefer `cycle` rather than `switcher` (as in `cycle_begin()`, `cycle_next()`, `cycle_finish()`) but I don't have a strong opinion here.

jlindgren90:
> "Switcher" refers to window switcher itself, which can happen without OSD if `<windowSwitcher><osd show="no">`.

Huh, I did not know that was even an option. Then `switcher` is really the behavioral mode/state while `switcher_osd` is the visual UI that goes with it.

I like the "cycle" suggestion. When I hear "switcher", my mind goes to something concrete, like a visual UI element. Whereas `cycle_begin()` is easier to associate with the behavioral mode/state itself.

But I don't have a strong preference either.

ManuLinares:
Then perhaps instead of `switcher_osd` a better descriptive would be `switcher_ui`

01micko:
I'm going to go out on a limb here (as a native English speaker) and say that "switcher" makes no sense. A switch is binary, well from the beginning any way, pre-dating anything electrical.

**Selector**  :question: 

(ignore as required)

01micko:
> Personally I would prefer `cycle` rather than `switcher` (as in `cycle_begin()`, `cycle_next()`, `cycle_finish()`) but I don't have a strong opinion here.

Missed that one. @Consolatis were you on a similar train of thought to me?

tokyo4j:
Hm, I agree that "cycle" is a better term to describe our alt-tab behavior, but we already have `<windowSwitcher>` and `osd.window-switcher` in our configuration for it and I'm afraid calling it as "cycle" introduces another confusion for new contributors.

I'll continue #3236 without this PR for now.

Consolatis:
I think its valid to have different words for concepts within code and user-facing configuration / documentation. Its not optimal but in this case I think `cycle` or `cycler` is pretty clear. IMHO its more important to decouple the meaning of OSD from the window switcher logic itself because that is indeed just the UI part of it (and I am fine with sticking to the term `*_osd` for that within the code).

tokyo4j:
Okay, I'll rename `osd` to `cycle`. I agree that we should clarify the difference of window switching and its UI part anyway. I also like `cycle` is a bit shorter than `switcher`.

include/output.h
```diff
@@ -15,14 +15,14 @@ struct output {
...
 	struct wlr_scene_tree *session_lock_tree;
 	struct wlr_scene_buffer *workspace_osd;
 
-	struct osd_scene {
-		struct wl_list items; /* struct osd_item */
+	struct cycle_osd_scene {
+		struct wl_list items; /* struct cycle_osd_item */
 		struct wlr_scene_tree *tree;
-	} osd_scene;
+	} cycle_osd;
```
tokyo4j, line 25: Note: updated `output->osd_scene` to `output->cycle_osd`.

include/labwc.h
```diff
@@ -302,15 +302,15 @@ struct server {
...
+	struct cycle_state {
+		struct view *selected_view;
 		bool preview_was_shaded;
 		bool preview_was_enabled;
 		struct wlr_scene_node *preview_node;
 		struct wlr_scene_tree *preview_parent;
 		struct wlr_scene_node *preview_anchor;
 		struct lab_scene_rect *preview_outline;
-	} osd_state;
+	} cycle;
```
tokyo4j, line 313: Note:  updated `server->osd_state` to `server->cycle` and updated `osd_state.cycle_view` to `cycle.selected_view`.

elviosak:
I might be late to the discussion, but "switcher" is clearer to me. I'm not a native english speaker, but i've always used software/games in english, and i "think" it's also commonly used in other WM's/Compositors (not sure about that, but at least on windows it is/was called task switcher)

alttab

https://en.wikipedia.org/wiki/Alt-Tab

tokyo4j:
> I might be late to the discussion, but "switcher" is clearer to me. I'm not a native english speaker, but i've always used software/games in english, and i "think" it's also commonly used in other WM's/Compositors (not sure about that, but at least on windows it is/was called task switcher)

Yeah, I agree with your opinion and I still slightly prefer "switcher" for documentation/code consistency. Wayfire also uses "switcher" in its code. But I no longer have strong preference on this and I'd like to go with anything that we maintainers agrees the most.

@malm Would you mind commenting on this? I think neither of @Consolatis, @jlindgren90 and me have strong preference on this and we can follow your decision. I can restore "switcher" version from GitHub if you prefer it.

EDIT: Oh, sorry so much for an annoying notification, Mr. Erik Malm. I meant to mention johanmalm...

johanmalm:
I prefer `cycle` and `cycle_osd` for the code (as you have done it).

LGTM.

Issue #3161, opened 2025-10-21
Problem in switcher's thumbnail style


tsujan:
### Problem description

Let me show it to you with 2 screenshots:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/72f71de6-5657-40ac-8cd7-f9642f7a6307" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/96672799-c1de-4bb4-9454-68c19e2007f6" />

Put into words, the thumbnail of pcmanfm-qt may be wrong: sometimes its icons aren't shown, sometimes it's black. My wild guess is that the thumbnail is created too soon.

### Steps to reproduce

Try to use the thumbnail style with pcmanfm-qt. Sooner or later, you might see wrong thumbnails. It's random though.

So far, I've been able to reproduce it only with pcmanfm-qt.

### labwc build source

Local build

### labwc version

0.9.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Manjaro Testing

tsujan:
I succeeded in reproducing it with lximage-qt and another Qt-based image viewer too. No screenshot needed; they were black. Again, it happened randomly.

With pcmanfm-qt, sometimes the thumbnail belongs to the previous folder, before entering the current folder. With image viewers, it sometimes belongs to the previously shown image.

tokyo4j:
The black box is a know issue. It happens when the client buffer is unavailable when labwc shows the window switcher. We can fix this by using wlroots' private internal struct, or creating another swapchain for each thumbnail.

I have no idea about the missing icon. I cannot reproduce it. Could you provide the output of `labwc -d` when the icon is not shown?

tsujan:
> I have no idea about the missing icon.

It's not related to *icons* themselves. As I tried to describe the problem with different examples, the thumbnail seems to be about a *previous state* of the window — before a folder is populated, or before the directory is changed to the current one, or the previous image shown by an image viewer (not the current image) — in short, the thumbnail may not be about the moment the switcher is used. This happens randomly.

> Could you provide the output of labwc -d 

Sorry, being busy with coding, I can't do tests for a while…

tokyo4j:
[...]

PR #3236, opened 2025-11-26
Refactor window switcher

tokyo4j:
This is motivated by #3229, but this PR contains some cleaups that are not directly related to it.

~The first one is avoiding the word "osd" for describing window switcher in general. The second one is extracting osd initialization code from `update_switcher()` (prior `update_osd()`). The third one is saving a fixed array of windows at the start of window switching, which helps #3229.~

~Let's review & merge one by one.~

## 1. Clarify the lifecycle of window switcher states

The role of `update_cycle()` was a bit complex; it both initializes and updates the window switcher, and calls `cycle_finish()` if it failed to initialize it. Also, it was a bit hard to follow what happens in `cycle_on_view_destroy()` when the window being cycled is destroyed. So this PR splits `update_cycle()` into `init_cycle()` and `update_cycle()`, and clarifies:
- `init_cycle()`: initializes the window switcher (e.g. OSD).
- `update_cycle()`: updates the window switcher states including OSD, preview and outlines.
- `destroy_cycle()`: clears all the window switcher states.

Whenever a window destroyed, `cycle_reinitialize()` (renamed from `cycle_on_view_destroy()` with calling it from map/unmap handler in mind) calls `init_cycle()`, sets the new selected window, and calls `update_cycle()`.

## 2. Remember cycled window list in `server->cycle.views`

We had duplicated logic to determine which windows to be cycled: `updated_cycle()` determined which windows to be shown in the OSD, but `get_next_selected_view()` also determined which window to cycle next via `view_next_no_head_stop()`/`view_prev_no_head_stop()`. Instead, this PR creates a list of windows to be cycled at the start of window switching and then iterate over it when cycling windows. This allows smoothly supports different window orders (#3229). For example, we can sort the cycle windows by its title like this:

```diff
diff --git a/src/cycle/cycle.c b/src/cycle/cycle.c
index 292e6436..87508e4f 100644
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -276,13 +276,28 @@ create_osd_on_output(struct output *output)
 	assert(output->cycle_osd.tree);
 }
 
+static void
+insert_view_ordered_by_title(struct wl_list *views, struct view *new_view)
+{
+	struct wl_list *link = views;
+	struct view *view;
+	wl_list_for_each(view, views, cycle_link) {
+		if (strcmp(new_view->title, view->title) < 0) {
+			break;
+		}
+		link = &view->cycle_link;
+	}
+	wl_list_insert(link, &new_view->cycle_link);
+}
+
 /* Return false on failure */
 static bool
 init_cycle(struct server *server)
 {
 	struct view *view;
 	for_each_view(view, &server->views, rc.window_switcher.criteria) {
-		wl_list_append(&server->cycle.views, &view->cycle_link);
+		// wl_list_append(&server->cycle.views, &view->cycle_link);
+		insert_view_ordered_by_title(&server->cycle.views, view);
 	}
 	if (wl_list_empty(&server->cycle.views)) {
 		wlr_log(WLR_DEBUG, "no views to switch between");

```

Note that this change might degrade user experience by not being able to cycle through windows newly spawned during window switching. Before this PR, they were not shown in the OSD but they can be cycled. I think we can fix this by calling `cycle_reinitialize()` from map/unmap handlers instead of view destroy handler later.

Another change in this PR is that the previous/next window of `server->active_view` (instead of the topmost window) is always selected at the start of window switching. This makes sure that Alt-Tab + Alt-Shift-Tab selects the originally focused window even when it's not the topmost window.

### DIFF
--- a/include/cycle.h
+++ b/include/cycle.h
@@ -87,7 +87,7 @@ struct cycle_osd_impl {
 	 * Create a scene-tree of OSD for an output.
 	 * This sets output->cycle_osd.{items,tree}.
 	 */
-	void (*create)(struct output *output, struct wl_array *views);
+	void (*create)(struct output *output);
 	/*
 	 * Update output->cycle_osd.tree to highlight
 	 * server->cycle_state.selected_view.
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -303,6 +303,7 @@ struct server {
 	/* Set when in cycle (alt-tab) mode */
 	struct cycle_state {
 		struct view *selected_view;
+		struct wl_list views;
 		bool preview_was_shaded;
 		bool preview_was_enabled;
 		struct wlr_scene_node *preview_node;
--- a/include/view.h
+++ b/include/view.h
@@ -134,6 +134,9 @@ struct view {
 	const struct view_impl *impl;
 	struct wl_list link;
 
+	/* This is cleared when the view is not in the cycle list */
+	struct wl_list cycle_link;
+
 	/*
 	 * The primary output that the view is displayed on. Specifically:
 	 *
@@ -382,15 +385,6 @@ struct view *view_next(struct wl_list *head, struct view *view,
  * Returns NULL if there are no views matching the criteria.
  */
 struct view *view_prev(struct wl_list *head, struct view *view,
-	enum lab_view_criteria criteria);
-
-/*
- * Same as `view_next()` except that they iterate one whole cycle rather than
- * stopping at the list-head
- */
-struct view *view_next_no_head_stop(struct wl_list *head, struct view *from,
-	enum lab_view_criteria criteria);
-struct view *view_prev_no_head_stop(struct wl_list *head, struct view *from,
 	enum lab_view_criteria criteria);
 
 /**
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -4,8 +4,8 @@
 #include <wlr/types/wlr_scene.h>
 #include <wlr/util/box.h>
 #include <wlr/util/log.h>
-#include "common/array.h"
 #include "common/lab-scene-rect.h"
+#include "common/list.h"
 #include "common/scene-helpers.h"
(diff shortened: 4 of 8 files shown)


johanmalm:
LGTM. Tested.

johanmalm:
Thanks :smile:

tokyo4j:
Just to note, I assume you tagged this PR as "breaking" following my comment:

> Note that this change might degrade user experience by not being able to cycle through windows newly spawned during window switching. Before this PR, they were not shown in the OSD but they can be cycled. I think we can fix this by calling cycle_reinitialize() from map/unmap handlers instead of view destroy handler later.

I actually tried calling `cycle_initialize()` from `view_impl_map()`, but it caused segfault with `style="thumbnail"` when a new window is spawned during window switching. That's because `view->current` is empty when the map handler (called before commit handler) is called, so the buffer for the thumbnail cannot be created:

https://github.com/labwc/labwc/blob/9f5ff391ccf37a3ad5be33a08f8ff9df161c7ba9/src/cycle/osd-thumbnail.c#L85-L87

At least we can guard against segfault by checking the size of `view->current`, but I'm not sure we should manage to render thumbnails from map handler. Options are:
1. Determine the window size from scene-tree: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/abf80b529e48823e21215a6ccc4653e2c2a4a565/types/ext_image_capture_source_v1/scene.c#L37-72
2. Use dummy `wl_output`s for thumbnails. This will enable live-updated thumbnails and fix #3161.

johanmalm:
> Just to note, I assume you tagged this PR as "breaking" following my comment:
> 
> > Note that this change might degrade user experience by not being able to cycle through windows newly spawned during window switching. Before this PR, they were not shown in the OSD but they can be cycled. I think we can fix this by calling cycle_reinitialize() from map/unmap handlers instead of view destroy handler later.
> 

Yes.

PR #3239, opened 2025-11-28
Add `direction` option to `Resize` action

mbroemme:
This introduces an optional `direction` argument to the `Resize` action, mirroring Fluxbox's `StartResizing [corner]` behavior.

Supported values (case-insensitive) are:
`up-left`, `up`, `up-right`, `left`, `right`, `down-left`, `down`, `down-right`.

If no direction is specified, the existing behavior is preserved and the resize edges are inferred from the current pointer position. The action documentation has been updated to describe the new argument. For example, using `down-right` lets you resize a window with right-click and drag always toward the same corner, no matter how close the pointer is to any other window edge or corner.

I don't know if Openbox has something similar as I'm coming from Fluxbox/X11 to labwc/Wayland. But in general this is a very simple and useful feature.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -51,8 +51,12 @@ Actions are used in menus and keyboard/mouse bindings.
 	another window or screen edge. If set to "no", only move to
 	the next screen edge. Default is yes.
 
-*<action name="Resize" />*
+*<action name="Resize" direction="value" />*
 	Begin interactive resize of window under cursor.
+
+	*direction* [up|down|left|right|up-left|up-right|down-left|down-right]
+	Edge or corner from which to start resizing. If this is not provided,
+	the direction is inferred from the cursor position.
 
 *<action name="ResizeRelative" left="" right="" top="" bottom="" />*
 	Resize window relative to its current size. Values of left, right,
--- a/src/action.c
+++ b/src/action.c
@@ -414,6 +414,20 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		break;
+	case ACTION_TYPE_RESIZE:
+		if (!strcmp(argument, "direction")) {
+			enum lab_edge edge = lab_edge_parse(content,
+				/*tiled*/ true, /*any*/ false);
+			if (edge == LAB_EDGE_NONE || edge == LAB_EDGE_CENTER) {
+				wlr_log(WLR_ERROR,
+					"Invalid argument for action %s: '%s' (%s)",
+					action_names[action->type], argument, content);
+			} else {
+				action_arg_add_int(action, argument, edge);
+			}
+			goto cleanup;
+		}
+		break;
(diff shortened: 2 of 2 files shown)


tokyo4j:
I think this change is acceptable, but do you have any specific use cases for this?

mbroemme:
My primary use case and the main value of `down-right` is predictability: you always know which corner is the anchor and how your drag will affect the window, regardless of where you happened to click.

It is due to my regular working with windowed applications. You will get consistent memory resizing for keyboard and mouse combo workflows, for example:

If doing it manually, I always resize from down-right; it happens to 90% of my application resizes and with `direction` you can:

* Bind something like `Super + Right Mouse Button` to resize with `direction=down-right`.
* No matter where you click on the window, the resize feels the same: drag right/down to make it bigger, left/up to make it smaller.

You don't have to "aim" for the correct corner anymore or worry that starting near the top edge will suddenly flip the resize direction.

src/action.c
```diff
@@ -414,6 +414,21 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		break;
+	case ACTION_TYPE_RESIZE:
+		if (!strcmp(argument, "direction")) {
+			/* Map human-readable corner/edge names to lab_edge bitmask */
```
tokyo4j: Please remove this comment, as it sounds too obvious.

docs/labwc-actions.5.scd
```diff
@@ -51,9 +51,15 @@ Actions are used in menus and keyboard/mouse bindings.
...
 
-*<action name="Resize" />*
+*<action name="Resize" direction="value" />*
 	Begin interactive resize of window under cursor.
 
+	When *direction* is given, start resizing from the specified edge
+	or corner instead of inferring it from the pointer position.
+
+	*direction* [up|down|left|right|up-left|up-right|down-left|down-right]
+	Edge or corner from which to start resizing.
```
tokyo4j: nit: I suggest putting them into one paragraph:

```suggestion
	*direction* [up|down|left|right|up-left|up-right|down-left|down-right]
	Edge or corner from which to start resizing. If this is not provided,
	the direction is inferred from the cursor position.
```

tokyo4j:
Although I will not use this feature myself, you explanation makes sense.

LGTM other than my nitpicks above. Thanks for your contribution!

src/action.c
mbroemme: Fixed with a7a6550596cdd01bea0dbe5724a14f3a004d7f7c

docs/labwc-actions.5.scd
mbroemme: Fixed with a7a6550596cdd01bea0dbe5724a14f3a004d7f7c

mbroemme:
@tokyo4j Thanks for review and feedback, ready to be merged.

tokyo4j:
Thanks!

PR #3240, opened 2025-11-28
rcxml: ignore and warn on empty strings in configuration

### LINKED ISSUE #3160: Margins quit being reserved at the edge of the screen after upgrading to 0.9.2

tokyo4j:
Closes #3160.

I rewrote the config parser in #2667, but it broke certain configurations by changing how empty strings are handled: they were just ignored before my parser rewrite, but after that, they are interpreted as just empty strings. So `<margin output="">` means `apply margin to an output named ""` rather than `apply margin to all the outputs`. In #3011, I wasn't aware this kind of breakages in functionalities.

On the other hand, using empty strings just to let the parser ignore them doesn't make sense and I think this is a bad practice (https://github.com/labwc/labwc/issues/3160#issuecomment-3499753127), though silently changing this behavior was certainly not ideal.

So, this PR restores the previous behavior of ignoring empty strings, and then prints an error message when ignoring them, encouraging users to avoid using empty strings just as a placeholder. Empty strings in `<libinput><device>` are exceptional; they represent device defaults (as described in https://github.com/labwc/labwc/pull/3011#issuecomment-3199450879), so they are still accepted without warnings. I commented out some entries in `rc.xml.all` so that reading it doesn't print any error messages. I think we can remove these checks in some releases later and interpret empty strings as they are.

### DIFF
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -33,8 +33,8 @@
 
   <!-- <font><theme> can be defined without an attribute to set all places -->
   <theme>
-    <name></name>
-    <icon></icon>
+    <!-- <name>Numix</name> -->
+    <!-- <icon>breeze</icon> -->
     <fallbackAppIcon>labwc</fallbackAppIcon>
     <titlebar>
       <layout>icon:iconify,max,close</layout>
@@ -218,9 +218,9 @@
     space automatically, so <margin> is only intended for other, specialist
     cases.
 
-    If output is left empty, the margin will be applied to all outputs.
+    If 'output' is not provided, the margin will be applied to all outputs.
 
-    <margin top="" bottom="" left="" right="" output="" />
+    <margin top="10" bottom="10" left="10" right="10" output="HDMI-A-1" />
   -->
 
   <!-- Percent based regions based on output usable area, % char is required -->
@@ -525,7 +525,11 @@
     If mouseEmulation is enabled, all touch up/down/motion events are
     translated to mouse button and motion events.
   -->
-  <touch deviceName="" mapToOutput="" mouseEmulation="no"/>
+  <touch>
+    <!-- <deviceName>ELAN2514:00 04F3:2AF1<deviceName> -->
+    <!-- <mapToOutput>HDMI-A-1</mapToOutput> -->
+    <mouseEmulation>no</mouseEmulation>
+  </touch>
 
   <!--
     The tablet cursor movement can be restricted to a single output.
@@ -552,7 +556,8 @@
     When using mouse emulation, the pen tip [tip] and the stylus buttons
     can be set to any available mouse button [Left|Right|Middle|..|Task].
   -->
-  <tablet mapToOutput="" rotate="0" mouseEmulation="no">
+  <tablet rotate="0" mouseEmulation="no">
+    <!-- <mapToOutput>HDMI-A-1</mapToOutput> -->
     <!-- Active area dimensions are in mm -->
     <area top="0.0" left="0.0" width="0.0" height="0.0" />
     <map button="Tip" to="Left" />
@@ -586,7 +591,7 @@
       - sendEventsMode [yes|no|disabledOnExternalMouse]
       - calibrationMatrix [six float values split by space]
       - scrollFactor [float]
-   
+
     The following <libinput>...</libinput> block may not be complete for
     your requirements. Default values are device specific. Only set an option
     if you require to override the default. Valid values must be inserted.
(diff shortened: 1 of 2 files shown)


docs/rc.xml.all
```diff
@@ -595,6 +600,7 @@
 
   <libinput>
     <device category="default">
+      <!-- Empty strings mean to use device defaults -->
```
Consolatis: Suggest adding something like "and can be omitted"

docs/rc.xml.all
```diff
@@ -678,7 +684,7 @@
     <height>400</height>
     <initScale>2.0</initScale>
     <increment>0.2</increment>
-    <useFilter>true</useFilter>
+    <useFilter>yes</useFilter>
```
Consolatis, line 686: Nice catch.

src/cycle/osd-field.c
```diff
@@ -289,6 +290,12 @@ void
 osd_field_arg_from_xml_node(struct window_switcher_field *field,
 		const char *nodename, const char *content)
 {
+	if (string_null_or_empty(content)) {
+		wlr_log(WLR_ERROR, "Empty string is not allowed for "
+			"<windowSwitcher><fields><field><%s>", nodename);
+		return;
+	}
```
Consolatis: Doesn't this break an empty custom format?

Consolatis:
With those 2 points addressed this LGTM. Not tested.

src/cycle/osd-field.c
tokyo4j: It's indeed breaking for users upgrading labwc from 0.9.2 to 0.9.3, but empty custom formats were just ignored before 0.9.2. Same goes for other configurations touched in this PR.

I think we should just focus on restoring the behaviors of 0.9.2 in this PR, though it's not a strong preference.

EDIT: Decided to remove that check.

tokyo4j:
I also decided comment out configurations in `<libinput><device>` and print warnings on empty strings for consistency.
If we allow explicitly selecting default values, I think we can allow a string `default`, like we are doing for window rule properties or `<libinput><device category="default">`.

JFYI: @johanmalm @Narrat

docs/rc.xml.all
tokyo4j: Removed the comment as I decided to comment out the empty entries.

docs/rc.xml.all
```diff
@@ -595,21 +600,21 @@
 
   <libinput>
     <device category="default">
-      <naturalScroll></naturalScroll>
-      <leftHanded></leftHanded>
-      <pointerSpeed></pointerSpeed>
-      <accelProfile></accelProfile>
+      <!-- <naturalScroll>no</naturalScroll> -->
```
johanmalm, line 603: I don't think we should be providing values like `no` and `adaptive` here.
The `rc.xml.all` file contains default values, and for libinput devices the default is "nothing" (i.e. we do not change what ever the device property already is).

`tap` is an exception

docs/rc.xml.all
```diff
@@ -33,8 +33,8 @@
 
   <!-- <font><theme> can be defined without an attribute to set all places -->
   <theme>
-    <name></name>
-    <icon></icon>
+    <!-- <name>Numix</name> -->
```
johanmalm, line 36: We state at the top of the file that elements/attributes contain default values. If feel that showing `<name></name>` (or simply `<name/>`) indicates that the name is indeed unset by default.

I don't mind adding the `<!-- <name>Numix</name> -->` as an example of what it could be, but without the original `<name></name>` I think it's potentially confusing and misleading - or at least inconsistent.

src/config/rcxml.c
```diff
@@ -1076,7 +1111,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		return true;
 
 	} else if (str_space_only(content)) {
-		/* ignore empty leaf nodes other than above */
+		wlr_log(WLR_ERROR, "Empty string is currently not "
```
johanmalm: "Empty string is current not allowed" doesn't feel quite right because:

- I would argue that they are not "not allowed", we just interpret them as nothing by ignoring them.
- And "currently" feels like we're in a temporary state, which I prefer not to be on this point. In other words, I feel we ought to make up our mind which way we're going and then stick with it.

johanmalm (commented):
I'm good with going in this direction (ignoring empty strings), bit I feel we ought to think it through to the end to avoid changing this some time in the future.

I therefore suggest that before merging we ought to decide how we would set a string to empty should indeed that requirement emerge in future. For example with `--merge-config` I think it's quite conceivable that a use may wish to override a system wide setting in `/etc/labwc/rc.xml`. It could probably be achieved by specifying one space character or something like that, but suggest we agree approach.

tokyo4j:
@johanmalm Sorry for confusion, but I think you are misunderstanding my intention in this PR. I intend to **read all empty strings as they are** (instead of ignoring them) in the future. And the goals of this PR are:

1. Clarify that we intend to read empty strings as they are in the future
2. Temporarily fix the unintentional change in 0.9.2 that broke some existing configurations (#3160)
3. Encourage users to remove empty strings from their configurations copied from `rc.xml.all`

On a second thought, maybe the second goal (fixing #3160) is not really demanded and rather causing another confusion as I intend to stop ignoring empty strings again in the future. So maybe a simpler approach like https://github.com/tokyo4j/labwc/commit/a218df02e077e1d7567eee8a3f12615dc89ec1d5 is better.

johanmalm:
> On a second thought, maybe the second goal (fixing #3160) is not really demanded and rather causing another confusion as I intend to stop ignoring empty strings again in the future. So maybe a simpler approach like [tokyo4j@a218df0](https://github.com/tokyo4j/labwc/commit/a218df02e077e1d7567eee8a3f12615dc89ec1d5) is better.

Yes I think that would be better.

Considering that we've only had one person raising an issue (the `<margin output="">` I don't think it's worth flip-flopping on this.

So am I right in think that the main point of the PR will be to warn users that `<libinput>` child elements with empty text strings will in future be treated as empty strings (rather than being ignored).

tokyo4j:
Okay, I updated the branch with the simpler version.

> So am I right in think that the main point of the PR will be to warn users that <libinput> child elements with empty text strings will in future be treated as empty strings (rather than being ignored).

Yes, and same goes for other configurations handled in `entry()` (e.g. `<core><decoration>`). I believe this gives more consistency for our parser and removes the need to care about which empty strings are ignored (e.g. #3196).

johanmalm:
LGTM

tokyo4j:
@johanmalm I'm not sure if I can mark all of your comments as resolved. My intention of the changes in `rc.xml.all` is to comment out the empty nodes so that reading it doesn't print warnings introduced in this PR, and to provide working examples. Please tell me if you prefer other approaches.

EDIT: okay, thanks!

PR #3241, opened 2025-11-29
[snapping] replace <snapping><range> with <snapping><range inner="" outer="">

elviosak:
Why? With 2 screens side by side it's a bit hard to snap windows to the center (right of the left monitor or left of the right monitor), because if i increase <range> too much it will trigger the maximize whenever i want to move the windows a little (my windows are usually maximized vertically, some snapped to the sides, some in the middle, and i move dragging the titlebar).

### DIFF
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -174,32 +174,6 @@ interactive_begin(struct view *view, enum input_mode mode, enum lab_edge edges)
 	}
 }
 
-static enum lab_edge
-find_adjacent_edges(struct seat *seat, struct output *output)
-{
-	enum lab_edge adjacent = LAB_EDGE_NONE;
-	if (!output_is_usable(output)) {
-		return adjacent;
-	}
-	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_UP,
-			output->wlr_output, output->usable_area.x, output->usable_area.y)) {
-		adjacent |= LAB_EDGE_TOP;
-	}
-	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_DOWN,
-			output->wlr_output, output->usable_area.x, output->usable_area.y)) {
-		adjacent |= LAB_EDGE_BOTTOM;
-	}
-	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_LEFT,
-			output->wlr_output, output->usable_area.x, output->usable_area.y)) {
-		adjacent |= LAB_EDGE_LEFT;
-	}
-	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_RIGHT,
-			output->wlr_output, output->usable_area.x, output->usable_area.y)) {
-		adjacent |= LAB_EDGE_RIGHT;
-	}
-	return adjacent;
-}
-
 bool
 edge_from_cursor(struct seat *seat, struct output **dest_output,
 		enum lab_edge *edge1, enum lab_edge *edge2)
@@ -223,9 +197,31 @@ edge_from_cursor(struct seat *seat, struct output **dest_output,
 	}
 	*dest_output = output;
 
-	/* Translate into output local coordinates */
 	double cursor_x = seat->cursor->x;
 	double cursor_y = seat->cursor->y;
+
+	int top_range = rc.snap_edge_range_outer;
+	int bottom_range = rc.snap_edge_range_outer;
+	int left_range = rc.snap_edge_range_outer;
+	int right_range = rc.snap_edge_range_outer;
+	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_UP,
+			output->wlr_output, cursor_x, cursor_y)) {
+		top_range = rc.snap_edge_range_inner;
+	}
+	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_DOWN,
+			output->wlr_output, cursor_x, cursor_y)) {
+		bottom_range = rc.snap_edge_range_inner;
+	}
+	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_LEFT,
+			output->wlr_output, cursor_x, cursor_y)) {
+		left_range = rc.snap_edge_range_inner;
+	}
+	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_RIGHT,
+			output->wlr_output, cursor_x, cursor_y)) {
+		right_range = rc.snap_edge_range_inner;
+	}
+
+	/* Translate into output local coordinates */
 	wlr_output_layout_output_coords(seat->server->output_layout,
 		output->wlr_output, &cursor_x, &cursor_y);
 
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks for this contribution. 

I support this in principle, but we should still support `<snapping><range>`. Happy to consider deprecating it over time, but we try pretty hard to give plenty of warning (like 2+ releases).

Would you mind sharing the snap-to-center configuration so that I can try to reproduce the issue that this solves.

elviosak:
> I support this in principle, but we should still support `<snapping><range>`. Happy to consider deprecating it over time, but we try pretty hard to give plenty of warning (like 2+ releases).

``` C
} else if (!strcasecmp(nodename, "range.snapping")) {
	rc.snap_edge_horizontal_range = atoi(content);
	rc.snap_edge_vertical_range = atoi(content);
```
this part keeps `<range>` working the same, i think whatever comes later overrides it right? This works as expected: `<range>50</range><verticalRange>5</verticalRange>`.
Maybe it doesn't even need to be deprecated and i could re-add it to the documentation.


> Would you mind sharing the snap-to-center configuration so that I can try to reproduce the issue that this solves.

no special configuration, i meant that it's hard to snap to the "inner edge" between the screens, without this change, i use `<range>15</range>`, it's not too bad but i often trigger the `<topMaximize>` unintentionally (titlebar height is 20 i think).
With this change i'm using `<verticalRange>5</verticalRange><horizontalRange>100</horizontalRange>` so it's very easy to snap to the sides, and i can move windows without triggering `<topMaximize>` (but can still use it when i want).
Another way to "fix" this would be to have a resistance for the mouse pointer only when dragging a window, but i didn't look into it, this way seemed better/simpler.

tokyo4j:
I think we can add `<snapping><range inner="" outer="">`. That's similar to `<snapping><overlay><delay inner="" outer="">` we already have.

elviosak:
That would work too, i'll try it.

elviosak:
@tokyo4j done, and if we go with this change, should `<cornerRange>` be changed to `<range><corner>` too?

it looks a bit weird on docs:
```
*<snapping><range><inner>*++
*<snapping><range><outer>*++
*<snapping><cornerRange>*```

**Comment by elviosak** [comment]
separated the logic into another function, is it better?

**Comment by tokyo4j** [comment]
> @tokyo4j done, and if we go with this change, should `<cornerRange>` be changed to `<range><corner>` too?
> 
> it looks a bit weird on docs:
> 
> ```
> *<snapping><range><inner>*++
> *<snapping><range><outer>*++
> *<snapping><cornerRange>*```
> ```

I agree that it looks a bit weird, but I think having `<cornerRange>` and `<range>` separately makes some sense because they works in different ways (`<range>` defines the distance from edges while `<cornerRange>` defines the distance from corners).

So +1 for keeping `<cornerRange>`. @johanmalm I'll appreciate your thoughts on this.

> separated the logic into another function, is it better?

I quickly read through your codes and it looks good. I'll review and test later.

**Comment by johanmalm** [comment]
Yes, agree to keep cornerRange as is.

File: src/config/rcxml.c [hunk 1]
```diff
@@ -1180,7 +1180,12 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "unMaximizeThreshold.resistance")) {
 		rc.unmaximize_threshold = atoi(content);
 	} else if (!strcasecmp(nodename, "range.snapping")) {
-		rc.snap_edge_range = atoi(content);
+		rc.snap_edge_range_inner = atoi(content);
+		rc.snap_edge_range_outer = atoi(content);
```
tokyo4j, line 1184: I suggest adding a warning message: `wlr_log(WLR_ERROR, "<snapping><range> is deprecated. Use <snapping><range inner="" outer=""> instead.");`. This might be a bit annoying for users who just want to set universal snapping range, but I prefer letting users not have multiple ways for the same configuration.

src/interactive.c
```diff
@@ -210,13 +236,23 @@ edge_from_cursor(struct seat *seat, struct output **dest_output,
...
-	if (top < rc.snap_edge_range) {
+	enum lab_edge adjacent_edges = find_adjacent_edges(seat, output);
+	int top_range = adjacent_edges & LAB_EDGE_TOP ?
+			rc.snap_edge_range_inner : rc.snap_edge_range_outer;
+	int bottom_range = adjacent_edges & LAB_EDGE_BOTTOM ?
+			rc.snap_edge_range_inner : rc.snap_edge_range_outer;
+	int left_range = adjacent_edges & LAB_EDGE_LEFT ?
+			rc.snap_edge_range_inner : rc.snap_edge_range_outer;
+	int right_range = adjacent_edges & LAB_EDGE_RIGHT ?
+			rc.snap_edge_range_inner : rc.snap_edge_range_outer;
```
tokyo4j: I think we can instead add `static bool edge_has_adjacent_output(struct seat *, struct output *, enum wlr_direction)` and call it for each edge. This is mostly just my preference, but I prefer not having an intermediate variable.

elviosak: then we could use `wlr_output_layout_adjacent_output` directly? the function will basically just call that and return anyway so just revert this? https://github.com/labwc/labwc/pull/3241/commits/27bcffc66f7d2f29f450fd06c96f6a0615a4b526

elviosak: also, do you prefer if/else like before that commit, set default values then re-set like this: 
``` C
	int top_range = rc.snap_edge_range_outer;
	int bottom_range = rc.snap_edge_range_outer;
	int left_range = rc.snap_edge_range_outer;
	int right_range = rc.snap_edge_range_outer;
	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_UP,
			output->wlr_output, cursor_x, cursor_y))
		top_range = rc.snap_edge_range_inner;
	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_DOWN,
			output->wlr_output, cursor_x, cursor_y))
		bottom_range = rc.snap_edge_range_inner;
	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_LEFT,
			output->wlr_output, cursor_x, cursor_y))
		left_range = rc.snap_edge_range_inner;
	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_RIGHT,
			output->wlr_output, cursor_x, cursor_y))
		right_range = rc.snap_edge_range_inner;
```
or ternary?

tokyo4j: > also, do you prefer if/else like before that commit, set default values then re-set like this:
> 
> ```c
> 	int top_range = rc.snap_edge_range_outer;
> 	int bottom_range = rc.snap_edge_range_outer;
> 	int left_range = rc.snap_edge_range_outer;
> 	int right_range = rc.snap_edge_range_outer;
> 	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_UP,
> 			output->wlr_output, cursor_x, cursor_y))
> 		top_range = rc.snap_edge_range_inner;
> 	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_DOWN,
> 			output->wlr_output, cursor_x, cursor_y))
> 		bottom_range = rc.snap_edge_range_inner;
> 	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_LEFT,
> 			output->wlr_output, cursor_x, cursor_y))
> 		left_range = rc.snap_edge_range_inner;
> 	if (wlr_output_layout_adjacent_output(seat->server->output_layout, WLR_DIRECTION_RIGHT,
> 			output->wlr_output, cursor_x, cursor_y))
> 		right_range = rc.snap_edge_range_inner;
> ```

Yeah, I like this. Ternaries are not suitable here, as the condition is quite long.

PR #3243, opened 2025-11-30
 cycle: use dummy node to remember the scene position of previewed window 

tokyo4j:
With this approach, we only need to store `cycle->dummy_node` instead of `cycle->preview_anchor` and `cycle->preview_parent` and we don't need to care about annoying cases like when the previewed window has no siblings or when a window tracked by `cycle->preview_anchor` is destroyed.


Consolatis:
LGTM, not tested.

Nice cleanup!

PR #3245, opened 2025-12-01
 view: decouple always-on-top windows from the omnipresent state

### LINKED ISSUE #1151: Decouple stickyness from alwaysOnBottom/Top

tokyo4j:
Fixes #1151.

Before this PR, always-on-{top,bottom} windows were always visible on all workspaces (omnipresent) because the they were not in per-workspace trees like normal windows. This PR fixes this by removing per-workspace trees and instead showing/hiding views individually based on `view->workspace`.

This PR also updates `desktop_focus_output()` and `desktop_topmost_focusable_view()` to use `for_each_view()` instead of scene graph traversal. The functional changes will be that always-on-{top,bottom} windows can be now focused when layer-shell or session-lock surface loses focus, when switching workspaces, when `FocusOutput` action is executed, etc.

### DIFF
--- a/include/config/types.h
+++ b/include/config/types.h
@@ -62,10 +62,7 @@ enum lab_view_criteria {
 	/* No filter -> all focusable views */
 	LAB_VIEW_CRITERIA_NONE = 0,
 
-	/*
-	 * Includes always-on-top views, e.g.
-	 * what is visible on the current workspace
-	 */
+	/* Includes omnipresent (visible on all desktops) views */
 	LAB_VIEW_CRITERIA_CURRENT_WORKSPACE       = 1 << 0,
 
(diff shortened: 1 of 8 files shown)


Consolatis:
I kind of liked the simplicity of having a single workspace scene tree to enable/disable compared to manually enabling / disabling every window. What are the downsides of having another tree as child of each workspace tree for always on top windows?

tokyo4j:
Superseded by #3356

Issue #3152, opened 2025-10-17
popups disappear


BiliOslavy:
### Problem description

Running Imagemagick's display command brings up an image for a split-second and then it disappears.

labwc version:   0.9.0
Imagemagic version:    7.1.1.47
distro:   gentoo

FWIW, this behaviour is identical to what happens with xshogi as well.    I closed the ticket on xshogi since it was hard for people to work with,  but same failure mode.     Imagemagick is more mainstream, so it should be easier to work with and reproduce. 

### Steps to reproduce

1.   Install Imagemagick
2.   run "display <imagename>" where imagename is a common image file format like JPG.
3.   image comes up briefly and then disappears

### labwc build source

Release

### labwc version

0.9.0

### labwc environment

Nested on Wayland

### Distribution

Gentoo

BiliOslavy:
Just tried on 0.9.2.   Same issue.    An image will come up once after a reboot, 
but successive attempts fail.

johanmalm:
I can reproduce. This is an XWayland client. Help welcome to fault find.

Consolatis:
Maybe @jlindgren90 has an idea?

johanmalm:
I've done some testing:
- It works on sway, which indicates some issue with labwc
- It's intermittent. On drm it seems to only fail every 3-10 times for me. Nested it seems to work almost all the time.
- I've with `xwaylandPersistence` set to both yes/no, but that makes no difference.
- I've added some printfs to the override-redirect functions in `xwayland.c` and `xwayland-unmanaged.c` but that does not appear to show anything.

Smells like a race condition.

tokyo4j:
Strangely, I cannot reproduce the issue.

jlindgren90:
I can't reproduce either. The image window that comes up is very small though (minimum size?)

Maybe it's config-specific?

tsujan:
It also happened to me very rarely in a tk/tcl app, starting with wlroots 0.18. I didn't report it because I'd seen it more often in Wayfire, assuming that it might not be related labwc. Haven't encountered it recently.

UweKrueger:
I cannot reproduce the issue with Imagemagick's display, but for me `xedit` and `xfig` close/crash immediately. I can also reproduce the `xshogi` issue #2928. After some investigations it turned out that there are two different causes:

[...]

PR #3246, opened 2025-12-02
xwayland, xdg: remove useless pings

jlindgren90:
Removing _NET_WM_PING was discussed in #3152. The xdg_wm_base::ping seems to be equally useless, since we don't use the ping_timeout event in either case.


johanmalm:
LGTM.

PR #3248, opened 2025-12-02
 cycle: update config to <osd output="all|cursor|focused">

tokyo4j:
CC @SametAylak

Looking back #3201,  I think `<windowSwitcher><osd output="keyboard">` is a bit unclear and hard to interpret as "show OSD in the output with keyboard focus". Also, we use "cursor" instead of "pointer" in other configurations like `<placement policy="cursor">` and `<action name="ShowMenu" atCursor="">`.

So let's replace `output="all|pointer|keyboard"` with `output="all|cursor|focused"`. In documentation, I reordered them to `output="all|focused|cursor"` as `focused` feels like a bit more sophisticated and general policy.

Sorry for my indecisive steering!

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -371,10 +371,10 @@ this is for compatibility with Openbox.
 	"classic" displays window information like icons and titles in a vertical list.
 	"thumbnail" shows window thumbnail, icon and title in grids.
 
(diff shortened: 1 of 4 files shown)


Consolatis:
Thanks, this indeed feels more in line with the other settings.

PR #3249, opened 2025-12-02
xwayland: flush XCB connection to mitigate race between Raise and input

jlindgren90:
Fixes the mouse input issue with Xaw test client from:
- #3152

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -35,6 +35,7 @@
 
 #if HAVE_XWAYLAND
 #include <wlr/xwayland.h>
+#include "xwayland.h"
 #endif
 
 struct view *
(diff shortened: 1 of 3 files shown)


src/view.c
```diff
@@ -2247,6 +2248,17 @@ view_move_to_front(struct view *view)
 		move_to_front(view);
 	}
 
+#if HAVE_XWAYLAND
+	/*
+	 * view_move_to_front() is typically called on each mouse press
+	 * via Raise action. This means we are restacking windows just
+	 * about at the same time we send the mouse press input to the
+	 * X server, and creates a race where the mouse press could go
+	 * an incorrect X window depending on timing. To mitigate the
```
Consolatis: `[..] could go to an incorrect [..]`

jlindgren90: Oops, fixed

Consolatis:
Nice bisect + fix!

johanmalm:
Fantastic effort :smile:

PR #3251, opened 2025-12-04
cursor: fix a couple issues with motion events running nested

jlindgren90:
- Fix the number of motion events doubling with each nesting level
- Fix pointer going crazy in Wine/Wayland games due to incorrect synthesis of relative motion events

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -981,11 +981,6 @@ handle_motion_absolute(struct wl_listener *listener, void *data)
 	double dx = lx - seat->cursor->x;
 	double dy = ly - seat->cursor->y;
 
-	wlr_relative_pointer_manager_v1_send_relative_motion(
-		seat->server->relative_pointer_manager,
-		seat->seat, (uint64_t)event->time_msec * 1000,
-		dx, dy, dx, dy);
-
 	preprocess_cursor_motion(seat, event->pointer,
 		event->time_msec, dx, dy);
 }


src/input/cursor.c
```diff
@@ -978,6 +992,16 @@ handle_motion_absolute(struct wl_listener *listener, void *data)
...
 
+	/*
+	 * If a "locked" constraint is active, then seat->cursor->x/y
+	 * doesn't update and so the math to synthesize relative motion
+	 * events doesn't work. See also the comment re: running nested
+	 * in handle_motion().
+	 */
+	if (cursor_locked(seat, event->pointer)) {
+		return;
+	}
```
Consolatis: Hm.. Properly handling a absolute-only input + cursor lock scenario seems to require some work to get it right. E.g. looks like we'd need to store `event->{x,y}` between calls, calculate the diff and then emit `wlr_relative_pointer_manager_v1_send_relative_motion()` of the diff.

To unconditionally disable this code path with the assumption that there is another relative input device seems too focused on the wlroots wayland backend.

src/input/cursor.c
```diff
@@ -947,6 +948,19 @@ handle_motion(struct wl_listener *listener, void *data)
...
+		 * events are received from the parent compositor. Ignore
+		 * the relative events since we will synthesize them in
+		 * handle_motion_absolute(). One exception: if a "locked"
+		 * constraint is active, then absolute motion events are
+		 * ignored, so pass the relative motion events through.
+		 */
+		if (wlr_input_device_is_wl(&event->pointer->base)
+				&& !cursor_locked(seat, event->pointer)) {
+			return;
+		}
```
Consolatis: Similar thing here, other host compositors might not send absolute / relative cursor events.

src/input/cursor.c
jlindgren90: Yes, I looked into it a little and there is a lot of tech debt that would have to be addressed first.
With touchscreen input, it *might* work via mouse emulation, however `cursor_emulate_move_absolute()` doesn't respect a "locked" constraint at all.
With tablet input, that code calls `wlr_cursor_warp_absolute()` directly and does not attempt to synthesize relative motion events at all.
Altogether, the pointer input code seems to have a lot of inconsistencies that may need addressing.

src/input/cursor.c
jlindgren90: Absolute motion events come via the base `wl_pointer` interface so those should be available universally.
Relative motion events are via a wlr protocol, so would be missing on non-wlroots compositors. So in that case, the in-game cursor / aim / etc. might just be frozen in-place. (The current behavior is that it tends to rail to some edge of the constraint area and get stuck there.)

Consolatis: Yes, good point. Absolute motion events should be available on every wayland backend. Relative ones might not though.

Slight correction: the relative events come from an official wayland protocol: [relative-pointer-unstable-v1.xml](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/unstable/relative-pointer/relative-pointer-unstable-v1.xml).

jlindgren90: > Slight correction: the relative events come from an official wayland protocol

Ah, you're right. Need sleep...

tokyo4j:
> Fix the number of motion events doubling with each nesting level

I think this should be discussed with sway/wlroots devs. I quickly tested sway and it also seems to send both of relative & absolute motion events to the nested compositor.

A wlroots [commit](https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/1e568d84dfd40e7a106a2afc74cdd7d39ba9a7dd) says:

> Compositors can figure out how absolute and relative events are related
(e.g. whether they have been triggered by the same logical event) with
the frame event.

So I guess compositors are expected to buffer relative & absolute motion events and filter out either of them in frame events, but that sounds too annoying and I think wlroots should just drop relative motion protocol support in wayland backend.

jlindgren90:
> I quickly tested sway and it also seems to send both of relative & absolute motion events to the nested compositor.

I'm not suggesting changing this aspect. To the host compositor, there is nothing special about the nested compositor - it will get the same events as any other xdg-shell surface. 

> wlroots should just drop relative motion protocol support in wayland backend.

It's easy enough for the (nested) compositor to just ignore them if/when they're not wanted. I don't see why a wlroots change would be necessary.

tokyo4j:
> > I quickly tested sway and it also seems to send both of relative & absolute motion events to the nested compositor.
> 
> I'm not suggesting changing this aspect. To the host compositor, there is nothing special about the nested compositor - it will get the same events as any other xdg-shell surface.

Sorry, I was confused and too inaccurate on this. I tried running sway inside sway, and the nested sway seems to also treat relative & absolute motion events equally; it calls both of `wlr_relative_pointer_manager_v1_send_relative_motion()` and `wlr_seat_pointer_notify_motion()` for every relative or absolute motion events from the host compositor. I think this is what labwc is doing, too.

Let me ask you some questions:

1. What's the practical problem of "the number of motion events doubling with each nesting level"?

2. Could you elaborate on the part "pointer going crazy in Wine/Wayland games due to incorrect synthesis of relative motion events"? From my testing, clients running in the nested compositor (sway / labwc) observes following events in a frame:
  - `zwp_relative_pointer_v1.relative_motion` event with some delta
  - `wl_pointer.motion` event with some delta
  - `zwp_relative_pointer_v1.relative_motion` event with zero delta
  
    Is this causing any problems in wine/wayland games?

3. Is the problems addressed in this PR universal in most wlroots-based compositors?.

> > wlroots should just drop relative motion protocol support in wayland backend.
> 
> It's easy enough for the (nested) compositor to just ignore them if/when they're not wanted. I don't see why a wlroots change would be necessary.

Yeah, it's not "necessary", but I generally don't like to add backend-specific workarounds in compositor side. If it's not a problem specific to labwc, I think it should be addressed in wlroots side. The codes added in this PR is small, but understanding their background is not trivial.

Consolatis:
I think what we could do is to
- completely drop all relative motion events if originating from a wayland backend device (e.g. what this PR is doing but even if the cursor is locked)
- fix the absolute motion event to correctly translate into relative ones

That should fix the issue when running nested and also fix the issue with relative motion events on DRM with an input device only offering absolute events.

jlindgren90:
> fix the absolute motion event to correctly translate into relative ones

I responded to this suggestion already in https://github.com/labwc/labwc/pull/3251#discussion_r2587941405.

I tried it, it is more difficult than it sounds. You have to track/update a second cursor x/y position just for the absolute events. And figure out how/whether to update it if relative events are mixed in, or emulated pointer events from touch/tablet sources.

So, not something I can implement for 0.9.3, but either of you are welcome to try and succeed where I gave up.

Consolatis:
> You have to track/update a second cursor x/y position just for the absolute events.

Couldn't we just add a `x` and `y` to `struct input`? and in the absolute movement handler use it for the diff and then set the new absolute value received? I didn't think this trough completely but on the first glance I don't see the issue with that approach.

jlindgren90:
I'm happy to test it. I still think not synthesizing the relative events when locked is a valid short term fix, since it doesn't work currently anyway.

tokyo4j:
I'm still not sure I'm following the discussion correctly, but can't we just send relative cursor motion events only for relative pointer devices like this?:

```diff
diff --git a/src/input/cursor.c b/src/input/cursor.c
index 2a681a6e..1fe73750 100644
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -981,11 +981,6 @@ handle_motion_absolute(struct wl_listener *listener, void *data)
 	double dx = lx - seat->cursor->x;
 	double dy = ly - seat->cursor->y;
 
-	wlr_relative_pointer_manager_v1_send_relative_motion(
-		seat->server->relative_pointer_manager,
-		seat->seat, (uint64_t)event->time_msec * 1000,
-		dx, dy, dx, dy);
-
 	preprocess_cursor_motion(seat, event->pointer,
 		event->time_msec, dx, dy);
 }
```

I executed the following program in KWin and Mutter to see how they handle absolute motion events:

<details>
<summary>C program</summary>

```c
#include <assert.h>
#include <libevdev/libevdev-uinput.h>
#include <libevdev/libevdev.h>
#include <unistd.h>
#include <stdio.h>

#define ARRAY_SIZE(a) (sizeof(a)/sizeof(a[0]))

int
main()
{
	struct libevdev *dev = libevdev_new();
	libevdev_set_name(dev, "abs-only-pointer");

	libevdev_enable_event_type(dev, EV_ABS);
	libevdev_enable_event_code(dev, EV_ABS, ABS_X,
		&(struct input_absinfo){0, 0, 32767, 0, 0, 0});
	libevdev_enable_event_code(dev, EV_ABS, ABS_Y,
		&(struct input_absinfo){0, 0, 32767, 0, 0, 0});

	libevdev_enable_event_type(dev, EV_KEY);
	libevdev_enable_event_code(dev, EV_KEY, BTN_LEFT, NULL);

	struct libevdev_uinput *uidev = {0};
	assert(libevdev_uinput_create_from_device(
		       dev, LIBEVDEV_UINPUT_OPEN_MANAGED, &uidev)
		== 0);

	for (int i = 0; i < 1; i++) {
		int positions[][2] = {
			{15000, 7000},
			{15000, 8000},
			{15000, 9000},
			{15000, 10000},
			{15000, 11000},
			{15000, 12000},
			{15000, 13000},
			{15000, 14000},
		};
		for (int j = 0; j < ARRAY_SIZE(positions); j++) {
			sleep(1);
			fprintf(stderr, "Move to (%d, %d)\n",
				positions[j][0], positions[j][1]);
			libevdev_uinput_write_event(uidev, EV_ABS, ABS_X,
						   positions[j][0]);
			libevdev_uinput_write_event(uidev, EV_ABS, ABS_Y,
						   positions[j][1]);
			libevdev_uinput_write_event(uidev, EV_SYN, SYN_REPORT, 0);
		}
	}

	libevdev_uinput_destroy(uidev);
	libevdev_free(dev);
	return 0;
}
```

</details>

They both just didn't send relative motion events (`zwp_relative_pointer_v1.relative_motion`), but only absolute motion events (`wl_pointer.motion`).

jlindgren90:
> I'm still not sure I'm following the discussion correctly, but can't we just send relative cursor motion events only for relative pointer devices like this?

Really good question, I don't know the answer. I assumed there was a good reason to be synthesizing the relative events, and focused only on disabling the broken case (when cursor is locked), but maybe we don't need to do it at all, if kwin and mutter aren't.

Was that test that you did running kwin and mutter nested? Or do you have real input hardware that emits absolute motion events?

tokyo4j:
> Was that test that you did running kwin and mutter nested? Or do you have real input hardware that emits absolute motion events?

That wasn't nested. I launched `startplasma_wayland` or `dbus-run-session -- gnome-shell --wayland` from VT. Inside it, I executed the C program shown in https://github.com/labwc/labwc/pull/3251#issuecomment-3619470302 to emulate an absolute pointer device (this should be the same as physical device from libinput's perspective). I'll ask Peter Hutterer (libinput maintainer) in IRC which real hardware is recognized as an absolute pointer device rather than touchscreen/tablet/relative pointer device.

EDIT: https://discussion.fedoraproject.org/t/weird-pen-tablet-behaviour-under-gnome-wayland/123821 says OpenTabletDriver supports "absolute mode" that reports `POINTER_MOTION_ABSOLUTE` events.

Consolatis:
> > I'm still not sure I'm following the discussion correctly, but can't we just send relative cursor motion events only for relative pointer devices like this?
> 
> Really good question, I don't know the answer. I assumed there was a good reason to be synthesizing the relative events, and focused only on disabling the broken case (when cursor is locked), but maybe we don't need to do it at all, if kwin and mutter aren't.

Indeed. Good question. I think we've been doing that since forever, potentially by copying what tinywl was doing. Looking into tinywl now it uses `wlr_cursor_warp_absolute()` / `wlr_cursor_move()` in the respective motion handlers and then calls a generic `process_cursor_motion()` handler in both cases. But it obviously doesn't handle pointer locks / confines.

So I guess the question is if we want to support absolute input devices in combination with pointer locks / confines or not. And if not to what extend, e.g. just ignoring the constraints or respecting them but still not synthesizing relative motion events.

CC @jp7677 for the tablet / mouse emulation case.

tokyo4j:
I think absolute pointer device + relative motion events themselves cannot work well in the first place. Let's assume a pointer device as a touchscreen that never reports when the finger is up or down. If the user puts his finger on the screen and swipes on a client surface, we can send relative motion events to the client by sending the delta between the absolute positions reported by the device (this is what we are doing). However, if the user lifts up his finger, moves it to its original place and puts it on the screen to swipe again, we get a quite large delta that will confuse the client. I confirmed we already have this problem for mouse emulation for real touchscreen/tablet devices ([test client](https://github.com/tokyo4j/wayland-demo/tree/relative-pointer)). Of course we can avoid this problem for mouse-emulated touchscreen/tablet devices by storing the last reported position in per-device struct and clearing it on every `TOUCH_UP`/`PROXIMITY` events, but it's annoying to implement and cannot be applied for general absolute pointer devices. This can also be a problem for nested compositors if we consider the case where the host compositor "warps" the cursor e.g. before/after alt-tab window switching. To fix this case, we'll need to `enter`/`leave` event of the host pointer, which is really annoying, too.

So in summary we currently have 2 practical problems:
1. Absolute pointer devices (real absolute pointer device, mouse-emulated touchscreen/tablet, or host pointer in nested compositor) creates insane relative motion events between pointer motions
2. When the pointer is locked, we report insane relative motion events during pointer motions

As these problems are hard to fix, and even impossible to fix for (1.), I think we should just not report relative events for absolute pointer devices.

tokyo4j:
FYI: I got a response from someone in IRC that says QEMU's emulated pointer is recognized as an absolute pointer device.

jlindgren90:
All right. Suppose we simply stop synthesizing relative events from absolute events, then:

- It would solve the 2nd (and worse) issue of this PR (pointer going crazy in Wine/Wayland games)
- Nested clients would see 2 absolute and 1 relative motion event for every motion from the host input device, which is still a 50% increase but not 2x as it is today. Probably not an issue in practice.

Should we go with that for 0.9.3? I can update the PR if others are on board.

jp7677:
> So I guess the question is if we want to support absolute input devices in combination with pointer locks / confines or not. And if not to what extend, e.g. just ignoring the constraints or respecting them but still not synthesizing relative motion events.
> 
> CC @jp7677 for the tablet / mouse emulation case.

Thanks for asking. I've always used a DRM session, never nested, during development. So I've unfortunately no experience with the type of events coming in in that case.

> I'll ask Peter Hutterer (libinput maintainer) in IRC which real hardware is recognized as an absolute pointer device rather than touchscreen/tablet/relative pointer device.

To work around tablet issues in Gnome/mutter, I'm currently using a libinput plugin which maps tablet motion to mouse motion (see the plugin code here https://gitlab.freedesktop.org/libinput/libinput/-/issues/1195). This also ends up emitting absolute motion for a pointer device. This works in mutter. This is a weird edge case but illustrates next to above mentioned devices that absolute pointer motion does happen in real life.

tokyo4j:
> - Nested clients would see 2 absolute and 1 relative motion event for every motion from the host input device, which is still a 50% increase but not 2x as it is today. Probably not an issue in practice.

It will be far less than 50%, as `wlr_seat_pointer_notify_motion()` should mostly return early when it doesn't update the cursor position. I'm not sure if the positions updated with relative / absolute motions can be sometimes slightly different, though.

> Should we go with that for 0.9.3? I can update the PR if others are on board.

I think we should, but let's see @Consolatis's view.

jlindgren90:
> I'm not sure if the positions updated with relative / absolute motions can be sometimes slightly different, though.

In my testing, they were usually slightly different (by a small fraction of a pixel). I think it's because absolute events are unitless (0 to 1 range) while relative events are in 1/256 pixel units, leading to rounding differences.

Pushed commit to disable the relative event synthesis. If @Consolatis agrees, we can go ahead and merge.

Consolatis:
> > Should we go with that for 0.9.3? I can update the PR if others are on board.
> 
> I think we should, but let's see @Consolatis's view.

Yes, lets try. @tokyo4j's point regarding position changes between touch_down / proximity_in events make a lot of sense. I guess the same is true for the wayland backend or qemu viewers if you move the cursor out of the window and then back in from another position.

Suggest updating the commit body though as its a more fundamental issue and not directly related to pointer constraints.

jlindgren90:
I said "There can be other situations as well" in the commit, but happy to update it. Care to suggest some wording to include?

tokyo4j:
> I said "There can be other situations as well" in the commit, but happy to update it. Care to suggest some wording to include?

Something like this?

*This also fixes another problem that happens even without pointer constraint: large deltas are synthesized between two sets of pointer motions coming from absolute pointer devices (e.g. mouse-emulated touchscreen/tablets and emulated pointer in nested compositors or qemu). See https://github.com/labwc/labwc/pull/3251#issuecomment-3620052002 for details.*

jlindgren90:
Thanks, that helps. I pushed a new commit message (no code changes) that reads like this now:

cursor: don't synthesize relative motion events from absolute events

It seems to have been inherited behavior from tinywl, but it's not clear
what purpose it serves, and it causes a couple of issues:

- A new absolute position that's discontinuous with the previous cursor
  position can produce unexpectedly large relative motion deltas. This
  can occur for example when multiple input devices are active, or in
  nested/VM scenarios when the pointer leaves the windowed output and
  re-enters at a different point.

- When the cursor position is locked via constraint, the computed deltas
  continue to get larger as the absolute event position diverges further
  from the locked position. This led to the mouse pointer going crazy in
  applications that use the relative events, such as games under Wine/
  Wayland.

jlindgren90:
Going for the merge - thanks all for the in-depth discussion!

Consolatis:
Sounds good. Thanks!

PR #3252, opened 2025-12-04
cursor: allow movement until entering the constraint surface

jlindgren90:
Fixes the issue pointed out in https://github.com/labwc/labwc/issues/2779#issuecomment-3588121048 where the cursor would get stuck (immovable) outside the window of a Wine/Wayland game, if it was already outside when the game started (common with a 4:3 game on a 16:9 screen). Now one can manually move the cursor into the game window, at which point it becomes locked.

This is a minimal/interim fix. Ideally we should warp the cursor into the constraint area automatically, but that would be a bit more work.

The change to apply_constraint() just turns an assert-failure into a safe no-op return, since the function is now entered for "locked" as well as "confined" constraint types.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -840,10 +840,12 @@ apply_constraint(struct seat *seat, struct wlr_pointer *pointer, double *x, doub
 	if (!seat->server->active_view) {
 		return;
 	}
-	if (!seat->current_constraint || pointer->base.type != WLR_INPUT_DEVICE_POINTER) {
+	if (!seat->current_constraint
+			|| pointer->base.type != WLR_INPUT_DEVICE_POINTER
+			|| seat->current_constraint->type
+				!= WLR_POINTER_CONSTRAINT_V1_CONFINED) {
 		return;
 	}
-	assert(seat->current_constraint->type == WLR_POINTER_CONSTRAINT_V1_CONFINED);
 
 	double sx = seat->cursor->x;
 	double sy = seat->cursor->y;
@@ -866,7 +868,9 @@ cursor_locked(struct seat *seat, struct wlr_pointer *pointer)
 {
 	return seat->current_constraint
 		&& pointer->base.type == WLR_INPUT_DEVICE_POINTER
-		&& seat->current_constraint->type == WLR_POINTER_CONSTRAINT_V1_LOCKED;
+		&& seat->current_constraint->type == WLR_POINTER_CONSTRAINT_V1_LOCKED
+		&& seat->current_constraint->surface
+			== seat->seat->pointer_state.focused_surface;
 }
 
 static void


Consolatis:
I think we should warp the cursor to the locked / confined region when focusing a confined / locked surface or activating a confine / lock on the currently focused surface.

Ph42oN:
I've noticed this issue as well but its minor since i can just focus different window and move cursor.

I think this looks like good way to fix that issue.

jlindgren90:
> I think we should warp the cursor to the locked / confined region when focusing a confined / locked surface or activating a confine / lock on the currently focused surface.

I agree we should do that eventually, but I think my smaller fix here is still an improvement over the status quo until that's implemented. It could also be useful even after, if there are corner cases such as an unmanaged surface (e.g. menu) sitting over the locked/confined region.

I'm happy to test a PR implementing the warp behavior, but I don't feel confident trying to implement it myself before 0.9.3.

src/input/cursor.c
```diff
@@ -840,10 +840,12 @@ apply_constraint(struct seat *seat, struct wlr_pointer *pointer, double *x, doub
...
 		return;
 	}
-	if (!seat->current_constraint || pointer->base.type != WLR_INPUT_DEVICE_POINTER) {
+	if (!seat->current_constraint
+			|| pointer->base.type != WLR_INPUT_DEVICE_POINTER
+			|| seat->current_constraint->type
+				!= WLR_POINTER_CONSTRAINT_V1_CONFINED) {
 		return;
 	}
-	assert(seat->current_constraint->type == WLR_POINTER_CONSTRAINT_V1_CONFINED);
```
Consolatis, line 846: Why the removal of the `assert()`? The added condition should make sure that its `V1_CONFINED`

Consolatis:
Hm.. it seems we already have a `warp_cursor_to_constraint_hint()`. I also wonder how good we deal with two applications simultaneously requesting a lock or confine.

tokyo4j:
> Ideally we should warp the cursor into the constraint area automatically, but that would be a bit more work.

> I think we should warp the cursor to the locked / confined region when focusing a confined / locked surface or activating a confine / lock on the currently focused surface.

I don't think that's ideal. I tested Kwin and Mutter, and they both locks the pointer only after the pointer enters the locking surface. I like this non-intrusive approach.

About this PR, it feels a bit odd that cursor can be moved even after `locked` event is sent. But as the pointer-locking client is unaware of the pointer activity outside of it's surface, I think this is acceptable as a small fix for 0.9.3.

 I implemented similar functionality by synchronizing `seat->current_constraint` with `seat->seat->pointer_state->focused_surface`: https://github.com/tokyo4j/labwc/tree/constraint-rewrite. I think this also improves the consistency of our pointer constraint logic. I can open a PR for it as a post-0.9.3 cleanup.

EDIT: edited comment as it didn't logically make sense.

src/input/cursor.c
jlindgren90: It's dead code now due to the same condition being checked in the `if`.

Consolatis:
ACK to merge for 0.9.3. Not tested.

PR #3258, opened 2025-12-06
NEWS.md: interim update

johanmalm:
...at start of cool-down period for 0.9.3

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog]
 
 | Date       | All Changes   | wlroots version | lines-of-code |
 |------------|---------------|-----------------|---------------|
-| 2025-11-15 | [unreleased]  | 0.19.2          | 28825         |
+| 2025-12-06 | [unreleased]  | 0.19.2          | 28927         |
 | 2025-10-10 | [0.9.2]       | 0.19.1          | 28818         |
 | 2025-08-02 | [0.9.1]       | 0.19.0          | 28605         |
 | 2025-07-11 | [0.9.0]       | 0.19.0          | 28586         |
@@ -116,8 +116,36 @@ differently [#3099].  There is a pending fix [wlroots-5159].
 
 [unreleased-commits]
 
+This release contains a good amount of bug-fixes, code simplification and
+small usability improvements.
+
+With the stability that comes with having tracked `wlroots 0.19` for a decent
+length of time, this feels like the best version of labwc so far.
+
+In terms of new features, it is worth drawing attention to the click support in
+the window-switcher on-screen-display by @tokyo4j [#3186] which has frequently
+been requested by users.
+
+As a general note to users, we discourage the use of empty strings in the
+`rc.xml` configuration file, for example `<theme><name></name></theme>`. There
+are only a few areas left where empty string are ignored (like under
+`<libinput>`) but the intent for future releases is to consistently read empty
+strings as empty strings. As a preparation, this release has added some warnings
+for empty strings that are currently ignored, so that users can take action.
+Also, the example `docs/rc.xml.all` has been updated to remove poor examples in
+this regard.
+
+A big thank you to all involved in this release.
+
 ### Added
 
+- Replace `<snapping><range>` with `<snapping><range inner="" outer="">` to
+  provide more granular control when configuring the size of snapping areas
+  (including `<topMaximize>`) on output edges with and without adjacent outputs.
+  @elviosak [#3241]
+- Add `direction` option to `Resize` action supporting the values `up-left`,
+  `up`, `up-right`, `left`, `right`, `down-left`, `down`, `down-right`. This
+  mirrors Fluxbox's `StartResizing [corner]` behavior. @mbroemme [#3239]
 - Allow the use of the `sendEventsMode` configuration option on keyboards in
   order to disable keyboard input. @cillian64 [#3208]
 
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -116,8 +116,35 @@ differently [#3099].  There is a pending fix [wlroots-5159].
...
+length of time, this feels like best version of labwc so far.
+
+In terms of new features, it is worth drawing attention to the click support in
+the window-switcher on-screen-display by @tokyo4j [#3186] which has frequently
+been requested by users.
+
+As a general note to users, we discourage the use of empty strings in the
+`rc.xml` configuration file, for example `<margin output=""/>`. There are only a
+few areas left where empty string are ignored (like under `<libinput>`) but the
+intent for future releases is to consistently read empty strings as empty
```
jlindgren90: I may have missed some discussion on this point, but I wonder if this "intent for future releases" is a good idea. I personally like the new "ignore and warn" behavior that got merged for 0.9.3. Offhand, I can't think of a good reason to explicitly use an empty string, so to "read empty strings as empty strings" (i.e. to give a different meaning to an empty string vs. an omitted string) seems like it would just cause confusion. The warnings seems like a good hint: "Hey, this probably doesn't do what you intended."

Consolatis: There are some cases like the workspace prefix or the custom window switcher field / thumbnail format strings. If these are set to an empty string they would mean an empty string. In contrast, if absent they would be using the default values.

jlindgren90: Okay, makes sense

NEWS.md
```diff
@@ -116,8 +116,35 @@ differently [#3099].  There is a pending fix [wlroots-5159].
...
+A big thank you to all involved in this release.
+
 ### Added
 
+- Replace `<snapping><range>` with `<snapping><range inner="" outer="">` to
+  provide more granular control when configuring the size of snapping areas
+  (including `<topMaximize>`) on output edges with and without adjacent outputs.
+  @elviosak [#3241]
+- Add directional options to `Resize` action to mirror Fluxbox's
+  `StartResizing [corner]` behavior. @mbroemme [#3239]
```
tokyo4j: We should mention the name of the new option:
```suggestion
- Add `directional` option to `Resize` action to mirror Fluxbox's
  `StartResizing [corner]` behavior. @mbroemme [#3239]
```

NEWS.md
```diff
@@ -181,6 +220,15 @@ differently [#3099].  There is a pending fix [wlroots-5159].
 
 ### Changed
 
+- `<snapping><range>` is deprecated. Use `<snapping><range inner="" outer="">`
+  instead. @elviosak [#3241]
+- When cycling through windows (typically with Alt-Tab) there are two minor
+  user-visible changes. For most users these will not be noticeable, but are
+  mentioned here for completeness.
+  - The initially selected window will now be the one that had keyboard focus
+    when cycling commenced rather than the topmost one. @tokyo4j [#3236]
```
tokyo4j: Our window switcher advances once at the beginning.

```suggestion
  - The initially selected window will now be the one that previously had keyboard focus
    when cycling commenced rather than the second topmost one. @tokyo4j [#3236]
```

NEWS.md
```diff
@@ -181,6 +220,15 @@ differently [#3099].  There is a pending fix [wlroots-5159].
...
+  - Windows that are spawned whilst cycling can no longer be cycled through
+    @tokyo4j [#3236]
```
tokyo4j: I think we should say: `We plan to fix this problem in future releases`

NEWS.md
```diff
@@ -116,8 +116,35 @@ differently [#3099].  There is a pending fix [wlroots-5159].
...
+
+With the stability that comes with having tracked `wlroots 0.19` for a decent
+length of time, this feels like best version of labwc so far.
+
+In terms of new features, it is worth drawing attention to the click support in
+the window-switcher on-screen-display by @tokyo4j [#3186] which has frequently
+been requested by users.
+
+As a general note to users, we discourage the use of empty strings in the
+`rc.xml` configuration file, for example `<margin output=""/>`. There are only a
```
tokyo4j: I suggest `<theme><name>` as an example instead, because we decided to read `<margin output="">` as an empty string without warnings. For background, my config refactoring unintentionally broke `<margin output="">` in the same way as #3160, and in #3240 we decided to keep it broken to avoid "flip-flopping" our parser behavior (https://github.com/labwc/labwc/pull/3240#issuecomment-3608309898).

tokyo4j:
LGTM other than my comments above. Thanks for your work!

NEWS.md
```diff
@@ -116,8 +116,35 @@ differently [#3099].  There is a pending fix [wlroots-5159].
 
 [unreleased-commits]
 
+This release contains a good amount of bug-fixes, code simplification and
+small usability improvements.
+
+With the stability that comes with having tracked `wlroots 0.19` for a decent
+length of time, this feels like best version of labwc so far.
```
Coccocoahelper: *the best version of labwc so far.
Aside from that, you could add a comma before "so far", but that is preferential.

NEWS.md
johanmalm: Done.

NEWS.md
johanmalm: Done.

NEWS.md
johanmalm: Thanks. Good spot. Done.

johanmalm:
Thanks. All suggestions incorporated. Going for the merge :smile:

PR #3259, opened 2025-12-06
docs/rc.xml: comment out example with empty string

johanmalm:


docs/rc.xml
```diff
@@ -6,7 +6,7 @@
 <labwc_config>
 
   <theme>
-    <name></name>
+    <!-- <name>Numix</name> -->
```
tokyo4j: As this is just an example file and there's not the guarantee like `rc.xml.all` that it should only show the default values, I think we don't need to comment it out. Perhaps we should use `Clearlooks` instead as it's the most used theme in Openbox. `Numix` was just my personal preference.

johanmalm: Yeah, good point. I think I was being paranoid that distro maintainers might package it, but it contains other non-default things anyway (like removing the `A-F4` binding).

Have removed the `<!-- -->` and changed to `Clearlooks-3.4` which is the Openbox default.

johanmalm:
Done. Also updated commit message. Ready to go again.

tokyo4j:
Thanks!

Issue #3238, opened 2025-11-28
Alt+F1 goes to virtual terminal


delentef:
### Problem description

Hello,

Sometimes I have this issue where hitting Alt+F1 or Alt+F2 etc; switches to a virtual console (the framebuffer); other times it's the right combination (Alt+Ctrl+F1) to switch to then framebuffer. I'm on labwc 0.9.2 but the issue appeared with  previous versions too.

Thanks for your work anyway, labwc is great!

### Steps to reproduce

1. start labwc
2. hit Alt+F1
3. I get the login prompt on a virtual terminal

### labwc build source

Release

### labwc version

0.9.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Fedora 42

johanmalm:
I'm not sure what's going on there.

Is there anything unusual in your setup. I know you've answered the "issue questions", but it's worth just thinking through in case there is a clue there. For example, do you run any client using a virtual keyboard? Do you have more than one keyboard connected. What type of keyboard(s)?

Do you have anything unusual in the `<keyboard>` section in `rc.xml`?

If you run with `LABWC_DEBUG_KEY_STATE=1 labwc 2> $HOME/log.txt` we can interrogate that log file one when the issue happens.

tokyo4j:
> If you run with `LABWC_DEBUG_KEY_STATE=1 labwc 2> $HOME/log.txt` we can interrogate that log file one when the issue happens.

You can also take my patch in #3262 to visually see the internal key state.

delentef:
Hello,

I'm coming back to this bug since it happened again after a hiatus.

I'm on fedora 43 with lightdm as my session manager. In `/usr/share/wayland-sessions/` I have just one file: `labwc.desktop` which contains

```[Desktop Entry]
Name=/usr/local/bin/labwc_start.sh
Comment=A wayland stacking compositor
Exec=labwc
Icon=labwc
Type=Application
DesktopNames=labwc;wlroots
```

My `/usr/local/bin/labwc_start.sh` script consists of:

```
#!/bin/bash

export LABWC_DEBUG_KEY_STATE=1
labwc  2>&1 | tee /tmp/labwc_start_log.txt &
```

My problem is that no `/tmp/labwc_start_log.txt` file is created. Is there a way to set the debug file name through an environment variable? Thanks!

F. Delente

johanmalm:
[...]

Issue #2771, opened 2025-05-28
labwc window switcher freezes itself from time to time


Bluey26:
### Problem description

Hello, according to what i have been mentioning in the labwc IRC room, i open this issue.

The problem itself is the following:

I use the labwc window switcher a lot, i find it really useful. From time to time, probably after several hours of uptime (like 7 or 8) the window switcher (sometimes) presents the following behavior:

When i try to change from app A to app B, and i have several other opened tabs, the window switcher freezes itself in the app B "vision" but it does not dissapears after the miliseconds of the animation. Instead, you see the task switcher and the window selected, and you cannot click any of the windows in the panel (lxqt-panel) or do any other actions that are not moving the mouse or cycling again in the window switcher.

This is a problem because sometimes i cannot change to app B directly using the window switcher. Instead, i have to unfocus the app A manually and open the app B using the window bar present in lxqt-panel. To do this, i have to go back to app A with the window switcher (i cannot click during this process).

For some reason this happens for some apps, and not for others, which made me think this was related to some kind of the xwayland apps i use from day to day (smplayer,stardict,etc.). But today i also verified that this happened to me even in xdg-shell (wayland native) apps like mozilla firefox.

To give more details about my system, i am using `labwc` from the archlinux official repositories:

`Build Date      : Sat May 3 17:35:35 2025`

I use a LXQt+labwc session, in archlinux. I usually keep the system updated, and do this task like 1 or 2 times per week.

### Steps to reproduce

I do not have steps to reproduce it yet, it happens after several hours in my system, and not always.

### labwc build source

Release

### labwc version

0.8.4

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

Consolatis:
If stuck in the window switcher, pressing `ESC` should at least make everything usable again.

As for why this happens in the first place I am kind of lost. @tokyo4j @jlindgren90 any random ideas what could cause this behavior?

We did have an issue in the past where the order of press and release events of `Alt` and `Tab` mattered and it was extremely difficult to diagnose as well, maybe we somehow re-introduced that issue?

Note: this is reported for 0.8.4 so before the X11 request focus feature was available.

tokyo4j:
Do you have any `virtual-keyboard-v1` client (e.g. fcitx5) or other programs that can modify keystroke?

Bluey26:
> Do you have any virtual-keyboard-v1 client (e.g. fcitx5) or other programs that can modify keystroke?

[...]

PR #3262, opened 2025-12-07
[just for debugging] show key states in the monitor

tokyo4j:
I originally made this to help debugging #2771. I'm opening this PR just to make this more accessible as this seems to be useful for debugging keyboard issues in general like #3238.

Perhaps we can integrate this to `labnag` instead by allowing clearing its message, but I'm not sure.

Shift+Ctrl+Alt+Z: show the key indicator
Shift+Ctrl+Alt+X: hide the key indicator

20251207_17h25m37s_grim

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -436,6 +436,10 @@ Actions are used in menus and keyboard/mouse bindings.
 	If used as the only action for a binding: clear an earlier defined
 	binding.
 
+*<action name="DebugToggleKeyStateIndicator" />*
+	Toggle visibility of key-state on-screen display (OSD). Note: This is for
+	debugging purposes only.
+
 # CONDITIONAL ACTIONS
 
 Actions that execute other actions. Used in keyboard/mouse bindings.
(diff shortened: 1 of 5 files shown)


johanmalm:
Just a thought... how including something like this in the main code base but instead of the `Shift+Ctrl+Alt+Z/X` we create a configuration option for it - some actions like `DebugToggleShowKeyState`.

There are undoubtedly some ongoing keyboard-state issues which are hard to debug, and I suspect that this type of support would make it easier for users to help analyze.

tokyo4j:
> Just a thought... how including something like this in the main code base but instead of the `Shift+Ctrl+Alt+Z/X` we create a configuration option for it - some actions like `DebugToggleShowKeyState`.
> 
> There are undoubtedly some ongoing keyboard-state issues which are hard to debug, and I suspect that this type of support would make it easier for users to help analyze.

I support your idea, but I'm not sure if I can take time to work on it. Feel free to take the code in this PR and make your own.

johanmalm:
@tokyo4j Apologies. I force-pushed to this branch by mistake. Forgive me. Hopefully it doesn't matter.

PR #3265, opened 2025-12-08
Primary workspace

csmccreery:
Adding the "firstDesktop" feature as described in the scope document. Allows for users to set the "primary" workspace in rc.xml which will be switched to immediately after the server is initialized.

### DIFF
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -429,11 +429,6 @@ workspaces_init(struct server *server)
 
 	if (initial_name) {
 		initial = workspace_find_by_name(server, initial_name);
-		if (!initial) {
-			wlr_log(WLR_ERROR,
-				"Initial workspace '%s' not found. Falling back to default",
-				initial_name);
-		}
 	}
 	if (!initial) {
 		initial = first;


src/config/rcxml.c
```diff
@@ -1492,6 +1495,7 @@ rcxml_init(void)
 	rc.resize_minimum_area = 8;
 
 	rc.workspace_config.popuptime = INT_MIN;
+	rc.workspace_config.primary_workspace = NULL;
```
Consolatis: Needs to be free'd in `rcxml_finish()` as well (there is a `zfree(rc.workspace_config.prefix)` already, you can just add a line after that one).

src/main.c
```diff
@@ -255,6 +256,11 @@ main(int argc, char *argv[])
 	rc.theme = &theme;
 	server.theme = &theme;
 
+	/* If the primary workspace was defined, set it */
+	if (rc.workspace_config.primary_workspace) {
+	        workspaces_set_primary(&server);
+	}
+
```
Consolatis: Suggest moving this to the end of `workspaces_init()` in `src/workspace.c`.

src/workspaces.c
```diff
@@ -209,12 +211,13 @@ add_workspace(struct server *server, const char *name)
...
 	if (!server->workspaces.current) {
-		server->workspaces.current = workspace;
+	    server->workspaces.current = workspace;
 	} else {
-		wlr_scene_node_set_enabled(&workspace->tree->node, false);
-	}
-
+	    wlr_scene_node_set_enabled(&workspace->tree->node, false);
+	}  
+
```
Consolatis: Some accidental tab to to space changes, please remove

src/workspaces.c
```diff
@@ -541,6 +552,21 @@ workspaces_find(struct workspace *anchor, const char *name, bool wrap)
 	return NULL;
 }
 
+void
+workspaces_set_primary(struct server *server) {
+        if (rc.workspace_config.primary_workspace) {
+	        char *primary_name = rc.workspace_config.primary_workspace;
+	        struct workspace *primary = workspaces_find(
+		        server->workspaces.current, primary_name, false);
```
Consolatis: Suggest removing this function then adding something like this at the end of `workspaces_init()` directly:
```c
if (!rc.workspace_config.primary) {
	return;
}
struct workspace *primary = workspaces_find(
	server->workspaces.current, rc.workspace_config.primary, /* wrap */ false);

if (primary) {
	workspaces_switch_to(primary, /*update_focus*/ false);
} else {
	wlr_log(WLR_ERROR, "..");
}
```

tokyo4j:
I think we need to add `<!-- <primary>Workspace 1</primary> -->` in `rc.xml.all` as it should contain all the possible configuration entries.

src/workspaces.c
csmccreery: Thanks for the feedback, I ran into some segmentation fault issues when I first tried to implement this at the end of workspaces_init() due to relying on theme_init but I'll go back and see if I can't resolve those. I agree that that's a cleaner implementation. 

Additionally the current prototype I have causes the ODS to show on login, that  was ultimately what caused the seg fault issues (I believe) and I have a working solution to prevent the ODS from showing but it requires updating the workspaces_switch_to() function to take an argument denoting whether this is the "initial" workspace selection. Appreciate the feedback and I'll get to work on cleaning this up!

Consolatis: Hm.. right. I'll look into it. Feel free to keep the callchain via `main.c` as is for now.

csmccreery:
Updated to reflect the logic changes, also changed name to 'firstDesktop' as that's whats in the scope document and closer to the convention elsewhere in rc.xml. This moves the logic for primary selection into workspaces_init and alters workspaces_switch_to to check if this is an "initial" workspace selection. Also updated the docs and rc.xml.all.

Let me know how you feel about it! I think this is a bit cleaner but it does come with the cost of adding an argument to switch_to.

EDIT: Forgot to mention. Tested this under the following conditions 1. firstDesktop matches an existing desktop, 2. firstDesktop does not match (defaults to first), 3. firstDesktop is empty, and 4. firstDesktop is not defined.

Consolatis:
I think something like this is cleaner (only slightly tested though):

```diff
diff --git a/docs/labwc-config.5.scd b/docs/labwc-config.5.scd
index 4e7d7ad8..6771e8f1 100644
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -569,6 +569,12 @@ extending outward from the snapped edge.
 	is 1. The number attribute is optional. If the number attribute is
 	specified, names.name is not required.
 
+*<desktops><firstDesktop>*
+	Define the default starting workspace at login. The default workspace
+	name must match one of the names defined in <dekstops><names>. If no
+	firstDesktop is defined or no matching firstDesktop can be found then
+	the default will be selected as the first workspace in <desktops><names>.
+
 *<desktops><popupTime>*
 	Define the timeout after which to hide the workspace OSD.
 	A setting of 0 disables the OSD. Default is 1000 ms.
diff --git a/docs/rc.xml.all b/docs/rc.xml.all
index f3d046d4..093b42fc 100644
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -175,6 +175,7 @@
     Workspaces can be configured like this:
     <desktops>
       <popupTime>1000</popupTime>
+      <firstDesktop>Workspace 1</firstDesktop>
       <names>
         <name>Workspace 1</name>
         <name>Workspace 2</name>
diff --git a/include/config/rcxml.h b/include/config/rcxml.h
index 3e4f15a2..3b1a6399 100644
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -168,6 +168,7 @@ struct rcxml {
 	struct {
 		int popuptime;
 		int min_nr_workspaces;
+		char *primary_workspace;
 		char *prefix;
 		struct wl_list workspaces;  /* struct workspace.link */
 	} workspace_config;
diff --git a/src/config/rcxml.c b/src/config/rcxml.c
index d3694337..3847e254 100644
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1299,6 +1299,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		wl_list_append(&rc.workspace_config.workspaces, &workspace->link);
 	} else if (!strcasecmp(nodename, "popupTime.desktops")) {
 		rc.workspace_config.popuptime = atoi(content);
+	} else if (!strcasecmp(nodename, "firstDesktop.desktops")) {
+		xstrdup_replace(rc.workspace_config.primary_workspace, content);
 	} else if (!strcasecmp(nodename, "number.desktops")) {
 		rc.workspace_config.min_nr_workspaces = MAX(1, atoi(content));
 	} else if (!strcasecmp(nodename, "popupShow.resize")) {
@@ -1953,6 +1955,7 @@ rcxml_finish(void)
 	zfree(rc.icon_theme_name);
 	zfree(rc.fallback_app_icon_name);
 	zfree(rc.workspace_config.prefix);
+	zfree(rc.workspace_config.primary_workspace);
 	zfree(rc.tablet.output_name);
 	zfree(rc.window_switcher.thumbnail_label_format);
 
diff --git a/src/workspaces.c b/src/workspaces.c
index f56d170a..b14d1d49 100644
--- a/src/workspaces.c
+++ b/src/workspaces.c
@@ -209,18 +209,11 @@ add_workspace(struct server *server, const char *name)
 	workspace->name = xstrdup(name);
 	workspace->tree = wlr_scene_tree_create(server->view_tree);
 	wl_list_append(&server->workspaces.all, &workspace->link);
-	if (!server->workspaces.current) {
-		server->workspaces.current = workspace;
-	} else {
-		wlr_scene_node_set_enabled(&workspace->tree->node, false);
-	}
-
-	bool active = server->workspaces.current == workspace;
+	wlr_scene_node_set_enabled(&workspace->tree->node, false);
 
 	/* cosmic */
 	workspace->cosmic_workspace = lab_cosmic_workspace_create(server->workspaces.cosmic_group);
 	lab_cosmic_workspace_set_name(workspace->cosmic_workspace, name);
-	lab_cosmic_workspace_set_active(workspace->cosmic_workspace, active);
 
 	workspace->on_cosmic.activate.notify = handle_cosmic_workspace_activate;
 	wl_signal_add(&workspace->cosmic_workspace->events.activate,
@@ -231,7 +224,6 @@ add_workspace(struct server *server, const char *name)
 		server->workspaces.ext_manager, /*id*/ NULL);
 	lab_ext_workspace_assign_to_group(workspace->ext_workspace, server->workspaces.ext_group);
 	lab_ext_workspace_set_name(workspace->ext_workspace, name);
-	lab_ext_workspace_set_active(workspace->ext_workspace, active);
 
 	workspace->on_ext.activate.notify = handle_ext_workspace_activate;
 	wl_signal_add(&workspace->ext_workspace->events.activate,
@@ -398,6 +390,26 @@ workspaces_init(struct server *server)
 	wl_list_for_each(conf, &rc.workspace_config.workspaces, link) {
 		add_workspace(server, conf->name);
 	}
+
+	struct workspace *initial = NULL;
+	struct workspace *first = wl_container_of(
+		server->workspaces.all.next, first, link);
+	const char *initial_name = rc.workspace_config.primary_workspace;
+	if (initial_name) {
+		initial = workspaces_find(first, initial_name, /*wrap*/ false);
+		if (!initial) {
+			wlr_log(WLR_ERROR, "No primary workspace matching %s", initial_name);
+		}
+	}
+	if (!initial) {
+		initial = first;
+	}
+
+	server->workspaces.current = initial;
+	wlr_scene_node_set_enabled(&initial->tree->node, true);
+	lab_ext_workspace_set_active(initial->ext_workspace, true);
+	lab_cosmic_workspace_set_active(initial->cosmic_workspace, true);
 }
 
 /*
```

</details>

Basically we manually set the initial workspace and not switch to it after the fact. Another thing I wondered about is if we should change the term `firstDesktop` to something like `initial`, `selected` or `startup`. That would allow a config like `<desktops number="5" startup="3" />`. But it might be confused with some kind of dynamic desktop creation..

csmccreery:
I had thought about putting this in add_workspace but at the time I didn't want to modify the logic in there too much as default workspace selection was already happening there. This is definitely the cleanest way to do this. I'll toy around with this and get a commit out tonight. 

I like initial as I think that's pretty clear as to the intent and I think it would be hard to mistake that to mean anything other than "the workspace I start out on".

csmccreery:
Sorry for the delay, I've pushed the changes discussed and updated the docs as well.

docs/labwc-config.5.scd
```diff
@@ -569,6 +569,12 @@ extending outward from the snapped edge.
 	is 1. The number attribute is optional. If the number attribute is
 	specified, names.name is not required.
 
+*<desktops><initial>*
+	Define the default starting workspace at login. The default workspace
+	name must match one of the names defined in <dekstops><names>. If no
```
Consolatis: Suggest changing into something like
```
[..] must match one of the names defined in <dekstops><names> or be an index lower or equal to *number*. 
```

include/config/rcxml.h
```diff
@@ -168,6 +168,7 @@ struct rcxml {
 	struct {
 		int popuptime;
 		int min_nr_workspaces;
+		char *primary_workspace;
```
Consolatis: Suggest renaming to something like `initial_workspace_name`.

src/config/rcxml.c
```diff
@@ -1299,6 +1299,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		wl_list_append(&rc.workspace_config.workspaces, &workspace->link);
 	} else if (!strcasecmp(nodename, "popupTime.desktops")) {
 		rc.workspace_config.popuptime = atoi(content);
+	} else if (!strcasecmp(nodename, "firstDesktop.desktops")) {
```
Consolatis: This is still using `firstDesktop`.

src/workspaces.c
```diff
@@ -25,6 +25,7 @@
 #include "view.h"
 
 #define COSMIC_WORKSPACES_VERSION 1
+
 #define EXT_WORKSPACES_VERSION 1
```
Consolatis, line 28: Unrelated empty line added, please remove.

src/workspaces.c
```diff
@@ -209,18 +210,10 @@ add_workspace(struct server *server, const char *name)
 	workspace->name = xstrdup(name);
 	workspace->tree = wlr_scene_tree_create(server->view_tree);
 	wl_list_append(&server->workspaces.all, &workspace->link);
-	if (!server->workspaces.current) {
-		server->workspaces.current = workspace;
-	} else {
-		wlr_scene_node_set_enabled(&workspace->tree->node, false);
-	}
-
-	bool active = server->workspaces.current == workspace;
```
Consolatis, line 218: There is a `wlr_scene_node_set_enabled(&workspace->tree->node, false);` missing which hides all workspaces on initialization.

src/workspaces.c
```diff
@@ -398,6 +390,31 @@ workspaces_init(struct server *server)
...
+			wlr_log(WLR_ERROR, "No workspace matching %s found",
+				primary_name);
+		}
+	}
+	if (!initial) {
+		initial = first;
+	}
+
+	server->workspaces.current = initial;
+	wlr_scene_node_set_enabled(&initial->tree->node, false);
```
Consolatis: The argument to `_set_enabled()` should be `true` here.

src/workspaces.c
```diff
@@ -398,6 +390,31 @@ workspaces_init(struct server *server)
...
 		add_workspace(server, conf->name);
 	}
+
+	/*
+	 * After adding workspaces, check if there is a primary selected and set
+	 * that as the primary workspace.
+	 */
+	char *primary_name = rc.workspace_config.primary_workspace;
+	struct workspace* initial = NULL;
+	struct workspace* first = wl_container_of(
```
Consolatis: Please use format `struct workspace *initial` and `*first` (e.g. `*` in front of the variable name).

csmccreery: Sorry missed that, re-commiting.

Consolatis:
Slightly tested and it seems to work fine. LGTM.

Now we just wait for labwc 0.9.3 to release and need to remember to squash the commits when merging.

src/workspaces.c
```diff
@@ -398,6 +390,31 @@ workspaces_init(struct server *server)
...
+	struct workspace *initial = NULL;
+	struct workspace *first = wl_container_of(
+		server->workspaces.all.next, first, link);
+
+	if (primary_name) {
+		initial = workspaces_find(first, primary_name, false);
+		if (!initial) {
+			wlr_log(WLR_ERROR, "No workspace matching %s found",
+				primary_name);
+		}
```
tokyo4j: This is unnecessary as `workspaces_find()` itself prints error when the workspace doesn't exist.

src/workspaces.c
```diff
@@ -398,6 +390,31 @@ workspaces_init(struct server *server)
 	wl_list_for_each(conf, &rc.workspace_config.workspaces, link) {
 		add_workspace(server, conf->name);
 	}
+
+	/*
+	 * After adding workspaces, check if there is a primary selected and set
+	 * that as the primary workspace.
+	 */
+	char *primary_name = rc.workspace_config.initial_workspace_name;
```
tokyo4j: `s/primary/initial/`

src/workspaces.c
```diff
@@ -398,6 +390,31 @@ workspaces_init(struct server *server)
 	wl_list_for_each(conf, &rc.workspace_config.workspaces, link) {
 		add_workspace(server, conf->name);
 	}
+
+	/*
+	 * After adding workspaces, check if there is a primary selected and set
+	 * that as the primary workspace.
+	 */
```
tokyo4j, line 424: `s/primary/initial/`

docs/labwc-config.5.scd
```diff
@@ -574,6 +569,13 @@ extending outward from the snapped edge.
 	is 1. The number attribute is optional. If the number attribute is
 	specified, names.name is not required.
 
+*<desktops><initial>*
+	Define the default starting workspace at login. The default workspace
+	name must match one of the names defined in <dekstops><names> or must
+	be an index equal to or lower than *number*. If no initial workspace is
+	defined or no matching initial workspace can be found then the default
+	will be selected as the first workspace in <names>.
```
tokyo4j: Labwc itself doesn't have the concept of "login". Also, the description of the default value looks a bit redundant and doesn't cover the case where `<number>` is used instead of `<names>`

I suggest a simpler description like below:

> Define the workspace selected at startup. This must match one of the names defined in `<names>` or an index equal to or lower than `<number>`. If not set, the first workspace is used.

src/workspaces.c
```diff
@@ -398,6 +390,31 @@ workspaces_init(struct server *server)
...
+	if (primary_name) {
+		initial = workspaces_find(first, primary_name, false);
+		if (!initial) {
+			wlr_log(WLR_ERROR, "No workspace matching %s found",
+				primary_name);
+		}
+	}
+	if (!initial) {
+		initial = first;
+	}
```
tokyo4j, line 435: We don't need this `if` branch if we declare `initial` like `struct workspace *initial = first;`

tokyo4j:
`workspaces_find()` accepts some fixed strings like `current`. Could you cherry-pick https://github.com/tokyo4j/labwc/commit/0e1626f7187c6d637a6a89ce6dfa38ad9421c260 and use `workspace_find_by_name()` instead?

src/workspaces.c
Consolatis: That wouldn't work in the case where `workspaces_find()` returns `NULL`.

tokyo4j: OK, never mind.

csmccreery:
Appreciate the feedback, I'll work through the docs and fix up the comments.

> `workspaces_find()` accepts some fixed strings like `current`. Could you cherry-pick [tokyo4j@0e1626f](https://github.com/tokyo4j/labwc/commit/0e1626f7187c6d637a6a89ce6dfa38ad9421c260) and use `workspace_find_by_name()` instead?

Just so I understand, would you like me to implement the workspace_find_by_name() function as it shows up in that commit or would you like me implement some of the logic in the workspace searching? I only ask for compatibility with [tokyo4j@0e1626f](https://github.com/tokyo4j/labwc/commit/0e1626f7187c6d637a6a89ce6dfa38ad9421c260)

tokyo4j:
> Just so I understand, would you like me to implement the workspace_find_by_name() function as it shows up in that commit or would you like me implement some of the logic in the workspace searching? I only ask for compatibility with [tokyo4j@0e1626f](https://github.com/tokyo4j/labwc/commit/0e1626f7187c6d637a6a89ce6dfa38ad9421c260)

You can use `git cherry-pick [my commit]` and rebase your branch on it.

Consolatis:
@5trixs0f If you don't want to go into that depth of git we can also modify that part later in another PR and you keep using `workspaces_find()`. I don't think there is anything inherently wrong with supporting `current|left|right` and so on. It doesn't make that much sense for the config option but it shouldn't hurt either.

csmccreery:
I'll see if I can't make them work together. Once everything is said and done I'll squash the commits as well.

csmccreery:
Cherry picked [tokyo4j@0e1626f](https://github.com/tokyo4j/labwc/commit/0e1626f7187c6d637a6a89ce6dfa38ad9421c260) and updated the comments and docs. Squashed commits to a single commit explaining the feature. Also moved the original location of workspace_find_by_name to higher in workspaces.c so it could be used by workspace_init.

tokyo4j:
@5trixs0f Please address my 2 unresolved comments above. LGTM other than them.

Consolatis:
Thanks!

csmccreery:
Thanks! Great working with you guys!

PR #3266, opened 2025-12-08
cycle: add `server->cycle_preview_tree`

tokyo4j:
This PR just addresses the `FIXME` comment. The only change in this PR is that the previewed window is placed above top/overlay layer surfaces.

I also updated the comment in `server_init()` that describes our scene-tree order.


johanmalm:
Forgot this one. Thanks.

LGTM.

PR #3267, opened 2025-12-09
desktop-entry: better handle desktop files with dots in their name

Consolatis:
This fixes - among others - cases like `R.E.P.O..desktop`.

Due to the duplicated dot at the end we were matching against a 0 sized string which was always true and thus would always match the desktop file, regardless of what the app-id was.

A related issue is that an app-id of `osomething` would match a desktop file name called `R.E.P.O.desktop` due to the string being size of 1.

Fix this by requiring the partial desktop filename string to be at least 3 characters wide.

Fixes #3257

### DIFF
--- a/src/desktop-entry.c
+++ b/src/desktop-entry.c
@@ -251,6 +251,9 @@ err:
  * (e.g. "thunderbird" matches "org.mozilla.Thunderbird.desktop"
  * and "XTerm" matches "xterm.desktop"). This is not per any spec
  * but is needed to find icons for existing applications.
+ *
+ * The second loop tries to match more partial strings, for
+ * example "gimp-2.0" would match "org.something.gimp.desktop".
  */
 static struct sfdo_desktop_entry *
 get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
@@ -258,6 +261,7 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 	size_t n_entries;
 	struct sfdo_desktop_entry **entries = sfdo_desktop_db_get_entries(db, &n_entries);
 
+	/* Would match "org.foobar.xterm" when given app-id "XTerm" */
 	for (size_t i = 0; i < n_entries; i++) {
 		struct sfdo_desktop_entry *entry = entries[i];
 		const char *desktop_id = sfdo_desktop_entry_get_id(entry, NULL);
@@ -266,6 +270,8 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		const char *desktop_id_base = dot ? (dot + 1) : desktop_id;
 
 		if (!strcasecmp(app_id, desktop_id_base)) {
+			wlr_log(WLR_DEBUG, "'%s' to '%s.desktop' via case-insensitive match",
+				app_id, desktop_id);
 			return entry;
 		}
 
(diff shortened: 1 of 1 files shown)


src/desktop-entry.c
```diff
@@ -283,15 +283,15 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
...
-		int alen = strlen(app_id);
-		int dlen = strlen(desktop_id_base);
-
-		if (!strncasecmp(app_id, desktop_id_base, alen > dlen ? dlen : alen)) {
+		const int dlen = strlen(desktop_id_base);
+		const int cmp_len = MIN(app_id_len, dlen);
+		if (cmp_len >= 3 && !strncasecmp(app_id, desktop_id_base, cmp_len)) {
 			return entry;
 		}
 	}
```
tokyo4j, line 314: How about this?

```c
	if (strlen(app_id) < 3) {
		return NULL;
	}
	for (size_t i = 0; i < n_entries; i++) {
		struct sfdo_desktop_entry *entry = entries[i];
		const char *desktop_id = sfdo_desktop_entry_get_id(entry, NULL);
		const char *dot = strrchr(desktop_id, '.');
		const char *desktop_id_base = dot ? (dot + 1) : desktop_id;
		if (strlen(desktop_id_base) < 3) {
			continue;
		}
		if (!strcasecmp(app_id, desktop_id_base)) {
			return entry;
		}
	}
```

tokyo4j: No. We need to match `gimp.desktop` and `gimp-2.10`: https://github.com/labwc/labwc/pull/2266#issuecomment-2435914399

tokyo4j: I think we need to add comments on what it is trying to match by showing examples. I'm not sure what `strrchr(desktop_id, '.')` and `strncasecmp()` are for. Though they were introduced in prior PRs.

Consolatis: Added 3 more comments with examples. Please tell if they should be enhanced and if yes in what way.

johanmalm: How about something like this...

Add

```C
/*
 * Here we get the last field of a dot separated string, so 'bar' given an a
 * desktop_id of `org.foo.bar`
 */
```

...before the `const char *desktop_id_base=` line


And something like:

```C
/* 
 * Given a very short desktop_id_base - like 'a' in 'org.foobar.a' - we could end up
 * checking if app_id begins 'a' which is likely to result in many unwanted hits. We
 * therefore check that length is at least an arbitrary 3.
 *
 * A real-life example of this is .desktop files containing more than one dot, for example
 * 'R.E.P.O.desktop'
 */
 ```

...before the `if (cmp_len >= 3)`

**Comment by johanmalm** [comment]
LGTM.

@spl237 IIRC you influenced/shaped/inspired this function. Any thoughts on this minor change.

**Comment by tokyo4j** [comment]
I suggest changes like this:

```diff
diff --git a/src/desktop-entry.c b/src/desktop-entry.c
index 63564ca0..c70256fa 100644
--- a/src/desktop-entry.c
+++ b/src/desktop-entry.c
@@ -270,6 +270,8 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		const char *desktop_id_base = dot ? (dot + 1) : desktop_id;
 
 		if (!strcasecmp(app_id, desktop_id_base)) {
+			wlr_log(WLR_DEBUG, "'%s' to '%s' via case-insensitive match",
+				app_id, desktop_id);
 			return entry;
 		}
 
@@ -282,6 +284,8 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		const char *wm_class =
 			sfdo_desktop_entry_get_startup_wm_class(entry, NULL);
 		if (wm_class && !strcasecmp(app_id, wm_class)) {
+			wlr_log(WLR_DEBUG, "'%s' to '%s' via StartupWMClass",
+				app_id, desktop_id);
 			return entry;
 		}
 	}
@@ -295,7 +299,16 @@ get_db_entry_by_id_fuzzy(struct sfdo_desktop_db *db, const char *app_id)
 		const char *desktop_id_base = dot ? (dot + 1) : desktop_id;
 		const int dlen = strlen(desktop_id_base);
 		const int cmp_len = MIN(app_id_len, dlen);
-		if (cmp_len >= 3 && !strncasecmp(app_id, desktop_id_base, cmp_len)) {
+		if (cmp_len < 3) {
+			/*
+			 * Without this check, app-id "foot" would match
+			 * "foobar.f" and whatever app-id would match "R.E.P.O."
+			 */
+			continue;
+		}
+		if (!strncasecmp(app_id, desktop_id_base, cmp_len)) {
+			wlr_log(WLR_DEBUG, "'%s' to '%s' via partial match",
+				app_id, desktop_id);
 			return entry;
 		}
 	}
@@ -308,11 +321,11 @@ get_desktop_entry(struct sfdo *sfdo, const char *app_id)
 {
 	struct sfdo_desktop_entry *entry = sfdo_desktop_db_get_entry_by_id(
 		sfdo->desktop_db, app_id, SFDO_NT);
-	if (!entry) {
-		entry = get_db_entry_by_id_fuzzy(sfdo->desktop_db, app_id);
+	if (entry) {
+		wlr_log(WLR_DEBUG, "matched '%s' via exact match", app_id);
+		return entry;
 	}
-
-	return entry;
+	return get_db_entry_by_id_fuzzy(sfdo->desktop_db, app_id);
 }
 
 static enum lab_img_type
```

</details>

- A simple description of why `cmp_len > 3` is needed with examples.
- Add debug logs to help debugging issues like #3257 easier.

Consolatis:
Added @tokyo4j's suggestions above, slightly modified to result in the following log lines:
- `matched 'mate-terminal.desktop' via exact match`
- `'mAte-terminal' to 'mate-terminal.desktop' via case-insensitive match`
- `'mAte-terminal.asd' to 'mate-terminal.desktop' via partial match`
- `failed to find .desktop file for 'asd.mAte-terminal'`
- `'footClient' to 'org.codeberg.dnkl.footclienta.desktop' via StartupWMClass`

tokyo4j:
Thanks!

PR #3270, opened 2025-12-11
docs: tab to space: 4, trailing newline: -1, fit some lines to 80-cols

domo141:
--
most of the accidental tabs from my previous 'tidy' commits - and I almost did one more ;/

simple `git grep '[^\t]\t.'` would have shown those, have to keep using in the future

the newline from the end of labnag...scd was done to be consisntent with other .scd files:
was it README which (still) has trailing newline (I personally find it easier to read that is
one of the reasons I kept "status quo" there)

### DIFF
--- a/docs/labnag.1.scd
+++ b/docs/labnag.1.scd
@@ -140,4 +140,3 @@ labnag \\
  --button-border-size 2\\
  -t 60
 ```
-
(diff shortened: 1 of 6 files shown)


johanmalm:
Thanks.

Issue #3071, opened 2025-09-10
NextWindow filters


orfeasxyz:
Currently the `NextWindow` action cycles through all windows and can only be filtered by workspace if `windowSwitcher.allWorkspaces="no"`. It would be useful if we could filter the cycled windows by some predicate (like `identifier`, `iconified` and `monitor`):

1) Filter by a static app-id ([discussion](https://github.com/labwc/labwc/discussions/3060))

```xml
<!-- static app-id -->
<keybind "W-t">
    <action name="NextWindow" identifier="foot"/>
</keybind>
```

2) Filter by the currently focused app-id ([comment](https://github.com/labwc/labwc/discussions/2521#discussioncomment-11814156))

```xml
<!-- dynamic app-id based on what is currently focused -->
<keybind "W-`">
    <action name="NextWindow" current_identifier="true"/>
</keybind>
```

3) Filter by non-minimized ([discussion](https://github.com/labwc/labwc/discussions/2521))

```xml
<!-- only cycle through non-minimized windows -->
<keybind "W-Tab">
    <action name="NextWindow" iconified="no"/>
</keybind>
```

4) Filter by monitor ([discussion](https://github.com/labwc/labwc/discussions/2690))

```xml
<!-- only cycle through non-minimized windows -->
<keybind "W-Tab">
    <action name="NextWindow" monitor="current"/>
</keybind>
```

These are some example window filters that already exist in the codebase and we can reuse them for this purpose.

johanmalm:
Seems reasonable. Any feel for how complex an implementation would be?

orfeasxyz:
> Seems reasonable. Any feel for how complex an implementation would be?

I am not very sure because I haven't seen much of how cycling is handled internally but I would assume not very difficult.

PR #3271, opened 2025-12-12
cycle: add `<action name="NextWindow" workspace="" output="" identifier="" >`

tokyo4j:
Supersedes #3183
Partially implements #3071

This PR adds `workspace="all|current"`, `output="all|focused|cursor"` and `identifier="all|current"` arguments to `NextWindow`/`PreviousWindow` and deprecates `<windowSwitcher><allWorkspaces>`.

Later, we can also implement other things suggested in #3071 like:
- `identifier="foot"`
- `minimized="all|yes|no"`

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -125,8 +125,8 @@ Actions are used in menus and keyboard/mouse bindings.
 	Resize and move the active window back to its untiled or unmaximized
 	position if it had been maximized or tiled to a direction or region.
 
-*<action name="NextWindow" workspace="current" />*++
-*<action name="PreviousWindow" workspace="current" />*
+*<action name="NextWindow" workspace="current" output="all" identifier="all" />*++
+*<action name="PreviousWindow" workspace="current" output="all" identifier="all" />*
 	Cycle focus to next/previous window, respectively.
 
 	Default keybinds for NextWindow and PreviousWindow are Alt-Tab and
@@ -136,6 +136,14 @@ Actions are used in menus and keyboard/mouse bindings.
 	*workspace* [all|current]
 	This determines whether to cycle through windows on all workspaces or the
 	current workspace. Default is "current".
+
+	*output* [all|focused|cursor]
+	This determines whether to cycle through windows on all outputs, the focused
+	output, or the output under the cursor. Default is "all".
+
+	*identifier* [all|current]
+	This determines whether to cycle through all windows or only windows of the
+	same application as the currently focused window. Default is "all".
 
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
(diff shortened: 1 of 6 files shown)


docs/labwc-actions.5.scd
```diff
@@ -125,13 +125,25 @@ Actions are used in menus and keyboard/mouse bindings.
...
+	This determines whether to cycle through windows on all workspaces or the
+	current workspace. Default is "current".
+
+	*output* [all|focused|cursor]
+	This determines whether to cycle through windows on all outputs, the focused
+	output, or the output under the cursor. Default is "all".
+
+	*appId* [all|current]
+	This determines whether to cycle through all windows or only windows of the
+	same application as the currently focused window. Default is "all".
```
Consolatis: Just something I noticed while skimming through the diff: `s/appId/identifier/`

tokyo4j: Done.

tokyo4j:
Rebased.

Consolatis:
LGTM, not tested.

Overall I think this is a nice structure and allows for easy additions of criteria in the future.

Edit:
~~One thing I am not sure about is if this keeps the old defaults. E.g. we might need to change the default keybinds for `A-Tab` / `A-S-Tab` to include `workspaces="current"`.~~ Disregard that, `run_action()` uses the appropriate defaults via `struct rc` which itself is defaulting to `CURRENT`.

src/config/rcxml.c
```diff
@@ -1269,10 +1269,17 @@ entry(xmlNode *node, char *nodename, char *content)
 	} else if (!strcasecmp(nodename, "outlines.windowSwitcher")) {
 		set_bool(content, &rc.window_switcher.outlines);
 	} else if (!strcasecmp(nodename, "allWorkspaces.windowSwitcher")) {
-		if (parse_bool(content, -1) == true) {
-			rc.window_switcher.criteria &=
-				~LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
+		int ret = parse_bool(content, -1);
+		if (ret == 0) {
```
johanmalm: Use `if (!ret)` instead of `if (ret == 0)`

https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#linux-kernel-style-basics

johanmalm:
Agree, LGTM.

Found a minor style nit. Did you want to fix that, or shall we just go for the merge?

src/config/rcxml.c
tokyo4j: Replaced with:
```c
		int ret = parse_bool(content, -1);
		if (ret < 0) {
			wlr_log(WLR_ERROR, "Invalid value for <windowSwitcher"
				" allWorkspaces=\"\">: '%s'", content);
		} else {
			rc.window_switcher.workspace_filter = ret ?
				CYCLE_WORKSPACE_ALL : CYCLE_WORKSPACE_CURRENT;
		}
```

...to clarify that `ret` is just a boolean if it's not negative. I hope this addresses your concern.

johanmalm: Yes, very nice.

PR #3275, opened 2025-12-16
Correct semantically-significant typo in labwc-config docs

Marie-Joseph:
"reserved" here implies that configuration files are parsed in the same order as before; "reversed" correctly implies they are parsed from least important to most important.  This wording is supported by code such as https://github.com/labwc/labwc/blob/7f0abed9c8c9859c4962c8feecfbe0b0a2ebb008/src/config/rcxml.c#L1924-L1931


johanmalm:
Great catch. Thanks.

PR #3282, opened 2025-12-21
Some tweaks for window switcher filters

tokyo4j:
Preparation for #3271.

The 1st commit puts `rc.window_switcher.{criteria,fields,style,output_criteria,thumbnail_label_format}` inside `rc.window_switcher.osd`. I noticed that we forgot to move `<windowSwitcher><fields>` to `<windowSwitcher><osd>`, but we already released 0.9.3 and it's not a big deal anyway.

The 2nd commit renames `cycle_osd_output_criteria` to `cycle_output_filter`. The reason for removal of "osd" is because we will reuse it for filtering of windows to cycle. "Filter" is just my preference. It's shorter than `criteria` and doesn't overlap with `lab_view_criteria`.

The 3rd commit rewrites `init_cycle()` with `get_output_by_filter()`. I'm not sure if this is the best approach, but my intention for it is to allow it to be reused for filtering of windows to cycle.

### DIFF
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -266,14 +266,36 @@ get_osd_impl(void)
 	return NULL;
 }
 
-static void
-create_osd_on_output(struct output *output)
+static uint64_t
+get_outputs_by_filter(struct server *server,
+		enum cycle_output_filter output_filter)
 {
-	if (!output_is_usable(output)) {
-		return;
+	struct output *output = NULL;
+
+	switch (output_filter) {
+	case CYCLE_OUTPUT_ALL:
+		break;
+	case CYCLE_OUTPUT_CURSOR:
+		output = output_nearest_to_cursor(server);
+		break;
+	case CYCLE_OUTPUT_FOCUSED: {
+		struct view *view = server->active_view;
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM, not tested.

src/cycle/cycle.c
```diff
@@ -266,14 +266,25 @@ get_osd_impl(void)
...
-create_osd_on_output(struct output *output)
+static struct output *
+get_output_by_filter(struct server *server,
+		enum cycle_output_filter output_filter)
 {
-	if (!output_is_usable(output)) {
-		return;
+	switch (output_filter) {
+	case CYCLE_OUTPUT_ALL:
+		return NULL;
```
johanmalm: Returning `NULL` for all outputs doesn't feel right to me and I think it makes `init_cycle()` harder to read.

Suggest returning a `wl_array` of outputs and just iterating over those in `init_cycle()`

tokyo4j: That makes sense. I updated that function to return the bitset of output indices.

johanmalm: The use of that `id_bit` is even nicer. I forgot that it existed :smile:

PR #3289, opened 2025-12-23
docs: add example for GTK4 composing

Consolatis:


docs/environment
```diff
@@ -22,6 +22,12 @@
 # XKB_DEFAULT_OPTIONS=grp:alt_shift_toggle
 # XKB_DEFAULT_OPTIONS=grp:shift_caps_toggle
 
+## GTK4 started to require input methods like fcitx5 or ibus to handle
```
johanmalm, line 25: Suggest saying from version 4.20.0.
Consider explaining what compose keys are, for example typing “à” with “`” + “a”.
It may also be worth pointing out what "input methods" are - e.g. `Fcitx5` and `IBus`
See: https://labwc.github.io/integration.html#gtk

Consolatis: Should we just put the link in there; something like `For further information see https://labwc.github.io/integration.html#gtk` ?

johanmalm: Good idea. Makes sense to just document in one place.

Consolatis: added the link

PR #3291, opened 2025-12-26
cycle: implement scrollable OSD 

### LINKED ISSUE #3162: window switcher may not fit on screen
### Problem description

If there are too many windows, the window switcher is not guaranteed to fit on the screen. This is a bigger issue with the thumbnail style (just because it is bigger), but the issue exists with both styles.

I suggest scrolling the switcher so that the currently selected entry is always visible.

### Steps to reproduce

Create a lot of windows, then use either window switcher.

### labwc build source

Release

### labwc version

0.9.2

### labwc environment

### Distribution

_No response_

tokyo4j:
Closes #3162.

This PR introduces scrolling mechanism in the window switcher OSD that scrolls items and hides overflowed ones so that the OSD fits within the screen. This shows a scrollbar indicating the current vertical scrolling position.

| `style="classic"` | `style="thumbnail"` |
|-|-|
| <video src="https://github.com/user-attachments/assets/5f841333-98d9-48bf-9536-8bde4658fcd8"> | <video src="https://github.com/user-attachments/assets/7c89ff7f-0b6e-4d00-b7c2-aafefb8b79ee"> |

FWIW, Openbox has arrow signs at the top and bottom of the OSD like below, but I prefer a scrollbar as it tells the current position of the selected item.

20251226_19h25m25s_grim

### DIFF
--- a/src/cycle/osd-classic.c
+++ b/src/cycle/osd-classic.c
@@ -97,11 +97,16 @@ cycle_osd_classic_init(struct cycle_osd_output *osd_output)
 	if (switcher_theme->width_is_percent) {
 		w = output_box.width * switcher_theme->width / 100;
 	}
-	int h = nr_views * switcher_theme->item_height + 2 * padding;
+	int workspace_name_h = 0;
 	if (show_workspace) {
 		/* workspace indicator */
-		h += switcher_theme->item_height;
+		workspace_name_h = switcher_theme->item_height;
 	}
+	int nr_visible_views = (output_box.height - workspace_name_h - 2 * padding)
+				/ switcher_theme->item_height;
+	nr_visible_views = MIN(nr_visible_views, nr_views);
+	int h = workspace_name_h + nr_visible_views * switcher_theme->item_height
+		+ 2 * padding;
 
 	osd_output->tree = wlr_scene_tree_create(output->cycle_osd_tree);
 
@@ -154,13 +159,17 @@ cycle_osd_classic_init(struct cycle_osd_output *osd_output)
 		goto error;
 	}
 
+	float *active_bg_color = switcher_theme->item_active_bg_color;
+	float *active_border_color = switcher_theme->item_active_border_color;
+	osd_output->items_tree = wlr_scene_tree_create(osd_output->tree);
+
 	/* Draw text for each node */
 	struct view *view;
 	wl_list_for_each(view, &server->cycle.views, cycle_link) {
 		struct cycle_osd_classic_item *item = znew(*item);
 		wl_list_append(&osd_output->items, &item->base.link);
 		item->base.view = view;
-		item->base.tree = wlr_scene_tree_create(osd_output->tree);
+		item->base.tree = wlr_scene_tree_create(osd_output->items_tree);
 		node_descriptor_create(&item->base.tree->node,
 			LAB_NODE_CYCLE_OSD_ITEM, NULL, item);
 		/*
@@ -185,9 +194,6 @@ cycle_osd_classic_init(struct cycle_osd_output *osd_output)
 		item->normal_tree = wlr_scene_tree_create(item->base.tree);
 		item->active_tree = wlr_scene_tree_create(item->base.tree);
 		wlr_scene_node_set_enabled(&item->active_tree->node, false);
-
-		float *active_bg_color = switcher_theme->item_active_bg_color;
-		float *active_border_color = switcher_theme->item_active_border_color;
 
 		/* Highlight around selected window's item */
 		struct lab_scene_rect_options highlight_opts = {
@@ -215,6 +221,17 @@ cycle_osd_classic_init(struct cycle_osd_output *osd_output)
 		y += switcher_theme->item_height;
 	}
 
+	struct wlr_box scrollbar_area = {
+		.x = w - padding - SCROLLBAR_W,
+		.y = padding,
+		.width = SCROLLBAR_W,
+		.height = h - 2 * padding,
+	};
+	cycle_osd_scroll_init(osd_output, scrollbar_area,
+		switcher_theme->item_height,
+		/*nr_cols*/ 1, /*nr_rows*/ nr_views, nr_visible_views,
+		active_border_color, active_bg_color);
+
 error:;
 	/* Center OSD */
 	wlr_scene_node_set_position(&osd_output->tree->node,
@@ -226,6 +243,8 @@ static void
 cycle_osd_classic_update(struct cycle_osd_output *osd_output)
 {
 	struct server *server = osd_output->output->server;
+	cycle_osd_scroll_update(osd_output);
+
 	struct cycle_osd_classic_item *item;
 	wl_list_for_each(item, &osd_output->items, base.link) {
 		bool active = item->base.view == server->cycle.selected_view;
--- a/src/cycle/osd-scroll.c
+++ b/src/cycle/osd-scroll.c
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#include <assert.h>
+#include <wlr/types/wlr_scene.h>
+#include "common/lab-scene-rect.h"
+#include "labwc.h"
+#include "cycle.h"
+#include "output.h"
+
+void
+cycle_osd_scroll_init(struct cycle_osd_output *osd_output, struct wlr_box bar_area,
+		int delta_y, int nr_cols, int nr_rows, int nr_visible_rows,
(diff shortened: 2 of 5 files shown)


Consolatis:
In terms of style I think this nails it. Only showing the bar on the right when necessary and it adjusting its size base on how many items are hidden is great IMHO. Only visual nitpick would be that its covering the right end of the active selection, maybe we should have the scrollbar allocate its own space instead?

Two things I noticed when having a quick test:
- we load the same icon over and over again each time when opening the OSD (e.g. 25 foot terminals -> 25 times .desktop file and icon lookup + loading of icon itself), at least when using the classic osd (obviously unrelated to this specific PR). I wonder if we could re-use the window icon instead if available or if we need a caching layer somewhere.
- i somehow managed to run into a busy loop when first testing with thumbnail view, then changing to classic + `Reconfigure` + `A-Tab`. didn't investigate yet why. Had to kill -9 the nested instance.

Will look more into the PR later. Thanks for your work on this!

johanmalm:
This is amazing. Thanks. Will try to look through code tomorrow if nobody else gets there first.

tokyo4j:
> Only visual nitpick would be that its covering the right end of the active selection, maybe we should have the scrollbar allocate its own space instead?

I actually considered implementing that, but I decided to not do it since I'd like to minimize the complexity of the scrolling mechanism, which is usually not shown to users unless they open so many windows. Also, if we implement it, the OSD could horizontally overflow from the screen due to the additional width of the scrollbar when `osd.window-switcher.style-thumbnail.width.max` is set to `100%` and we would need to recalculate the geometry of thumbnail items in order to prevent that.

> - we load the same icon over and over again each time when opening the OSD (e.g. 25 foot terminals -> 25 times .desktop file and icon lookup + loading of icon itself), at least when using the classic osd (obviously unrelated to this specific PR). I wonder if we could re-use the window icon instead if available or if we need a caching layer somewhere.

I thought this is handled by our buffer sharing mechanism, but adding some `wlr_log()`s shows that we are indeed looking up duplicated icons when opening OSD. I'll look into this further. This might be another reason to drop the buffer sharing mechanism and cache icon files by their file names instead. But yeah, this is unrelated to this PR anyway.

> - i somehow managed to run into a busy loop when first testing with thumbnail view, then changing to classic + Reconfigure + A-Tab. didn't investigate yet why. Had to kill -9 the nested instance.

Perhaps this is due to UAF in `cycle_osd_scroll_update()`. I updated this PR to clear `output->cycle_osd` in `cycle_destroy()`. Could you test it again to see the busy loop still happens?

include/output.h
```diff
@@ -22,6 +22,17 @@ struct output {
 	struct cycle_osd_scene {
 		struct wl_list items; /* struct cycle_osd_item */
 		struct wlr_scene_tree *tree;
+		struct wlr_scene_tree *items_tree;
+
+		struct cycle_osd_scroll_context {
+			/* index of the top visible row */
```
johanmalm: Superfluous comment?

include/output.h
```diff
@@ -22,6 +22,17 @@ struct output {
 	struct cycle_osd_scene {
 		struct wl_list items; /* struct cycle_osd_item */
 		struct wlr_scene_tree *tree;
+		struct wlr_scene_tree *items_tree;
+
+		struct cycle_osd_scroll_context {
```
johanmalm: I have a couple of thoughts here:
1. In a tactical sense, would it make sense to put this `cycle_osd_scroll_context struct` definition in a dedicated header file? They feel like concerns that could/should be separated.
2. In a more strategic sense, I feel we're mixing currencies here: `struct cycle_osd_scene` contains `wlr_scene_tree` (items_tree) but also a context which contains various state variables as well as another `wlr_scene_tree` (bar_tree). I get why we've ended up here - and that it keeps changes to a minimum - but when taking a step back it doesn't seems a clear/separated as it possibly could be.

src/cycle/osd-classic.c
```diff
@@ -216,6 +222,16 @@ cycle_osd_classic_create(struct output *output)
...
 	}
 
+	struct wlr_box scrollbar_area = {
+		.x = w - padding - SCROLLBAR_W,
+		.y = padding,
+		.width = SCROLLBAR_W,
+		.height = h - 2 * padding,
+	};
+	cycle_osd_scroll_init(output, scrollbar_area,
+		switcher_theme->item_height, 1, nr_views, nr_visible_views,
```
johanmalm: Consider inline comment to describe the `1`, possibly `/* nr_cols */`

src/cycle/osd-scroll.c
```diff
@@ -0,0 +1,99 @@
...
+	scroll->nr_visible_rows = nr_visible_rows;
+	scroll->top_row_idx = 0;
+	scroll->bar_area = bar_area;
+	scroll->delta_y = delta_y;
+	scroll->bar_tree = wlr_scene_tree_create(output->cycle_osd.tree);
+	wlr_scene_node_set_position(&scroll->bar_tree->node,
+		bar_area.x, bar_area.y);
+
+	struct lab_scene_rect_options scrollbar_opts = {
+		.border_colors = (float *[1]) { border_color },
```
johanmalm, line 31: Could we pass `float **border_color` to `cycle_osd_scroll_init()` and then simply do `.border_colors = border_color` here.

I might be missing something about how the code works here.

src/cycle/osd-scroll.c
```diff
@@ -0,0 +1,99 @@
...
+
+	int idx = 0;
+	struct cycle_osd_item *item;
+	wl_list_for_each(item, &output->cycle_osd.items, link) {
+		if (item->view == server->cycle.selected_view) {
+			return idx;
+		}
+		idx++;
+	}
+	/* should be unreachable */
```
johanmalm: Is this worthy of an `assert()`?

johanmalm (commented):
Thanks for this.

1. The functionality is great, and I like the generic approach by which both modules (classic and thumbnail) can use `cycle_osd_scroll_update()`.
2. I actually found the patch a bit tricky to read, but think that is simply because it implements something quite complex. I think you have struck a nice balance between keeping it simple versus inventing a whole new GUI toolkit :smile:
3. I have left a few comments, particularly one about separating concerns. If you think I'm over-thinking it, I'm happy to go with a LGTM though.

include/output.h
tokyo4j: Opened #3295 to aggregate types into `cycle.h` and replace `cycle_osd_scene` with `cycle_osd_output`.

In terms of separating concerns, I'm not very happy that `item_tree` is created by each module but also moved by `cycle_osd_scroll_update()`, but I think we have to live with that. I added comments to describe this.

src/cycle/osd-scroll.c
tokyo4j: Added `assert()`.

src/cycle/osd-classic.c
tokyo4j: Done.

include/output.h
tokyo4j: Removed.

src/cycle/osd-scroll.c
tokyo4j: Passing `float **border_color` to `cycle_osd_scroll_init()` doesn't make sense to me since a single color is usually declared as `float *` across our codebase. We could use `.border_colors = &border_color`, but I prefer `(float *[1]) { border_color }` to clarify that it's passing an array of border colors with a single element and to align with other calling sites that creates a rectangle with 3 border colors.

johanmalm: Okay. I'm good with what you've done. I wasn't thinking straight and missed the fact that it relates to colors in plural.

include/output.h
johanmalm: Thanks. Is this PR review for another review?

tokyo4j: > Is this PR review for another review?

Do you mean "Is this PR **ready** for another review"? Then yes, this PR is now ready again and I'd like you to see if the current data structure is okay. Actually I'm not sure if renaming `cycle_osd_scene` to `cycle_osd_output` addressed your concern (2.), but I haven't come up with better ways.

Consolatis:
Tested again, could not reproduce the busy loop. Discovered something new though: changing osd style + `sleep 10; labwc --reconfigure` + `A-Tab` + keeping the osd open until the reconfigure comes in triggers ASAN. Potentially pressing `Tab` once more is required depending on if switching from `classic` to `thumbnail` or the other way around.

johanmalm:
Great. LGTM. Going for merge.

Issue #3293, opened 2025-12-26
sendEventsMode="disabledOnExternalMouse" rejected as non-boolean


ManuLinares:
### Problem description

labwc rejects valid libinput values for <sendEventsMode> when parsing XML device configuration.
Using sendEventsMode="disabledOnExternalMouse" results in the error:
```
[ERROR] [../labwc/src/common/parse-bool.c:29] (disabledOnExternalMouse) is not a boolean value
```

According to this: https://labwc.github.io/labwc-config.5.html#entry_libinput_device_sendeventsmode it should disable my ps2 mouse if my wireless mouse is connected.

### Steps to reproduce

1. My rc.xml:

```xml
...
  <libinput>
    <device category="Logitech USB-PS/2 Optical Mouse">
      <pointerSpeed>0.0</pointerSpeed>
      <accelProfile>adaptive</accelProfile>
      <sendEventsMode>disabledOnExternalMouse</sendEventsMode>
    </device>
  
    <device category="Logitech Wireless Mouse">
      <pointerSpeed>-0.8</pointerSpeed>
      <accelProfile>adaptive</accelProfile>
      <scrollFactor>1.0</scrollFactor>
    </device>
  </libinput>
...
```

2. restart labwc

### labwc build source

Local build

### labwc version

0.9.2.r112.gfd6d62ea-1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Archlinux

Consolatis:
I think this is just a slightly annoying misleading log message. Looking at `get_send_events_mode()`, `disabledOnExternalMouse` should indeed work, regardless of the error log.

Consolatis:
Should be fixed by #3294, testing appreciated.

ManuLinares:
Hi, thanks for the fix but this fixes only the error log.

My issue still persists. According to this: https://labwc.github.io/labwc-config.5.html#entry_libinput_device_sendeventsmode it should disable my ps2 mouse if my wireless mouse is connected with a config like this one:

```xml
	<libinput>
		<device category="Logitech USB-PS/2 Optical Mouse">
			<pointerSpeed>0.0</pointerSpeed>
			<accelProfile>adaptive</accelProfile>
			<sendEventsMode>disabledOnExternalMouse</sendEventsMode>
		</device>
	
		<device category="Logitech Wireless Mouse">
			<pointerSpeed>-0.8</pointerSpeed>
			<accelProfile>adaptive</accelProfile>
[...]

PR #3294, opened 2025-12-26
config/rcxml.c: prevent wrong parse_bool() err message

Consolatis:
by only parsing the boolean when it is not `disabledOnExternalMouse`.

Fixes:
- #3293


johanmalm:
Thanks :smile:

PR #3295, opened 2025-12-28
cycle: refactor to aggregate type definitions into cycle.h

tokyo4j:
Preparation for #3291.

We declared `cycle_state` struct in `labwc.h` and `cycle_osd_scene` struct in `output.h`, which was unclean in terms of separation of concerns.

So this PR firstly moves `cycle_state` to `cycle.h`, then replaces `cycle_osd_scene` in `output.h` with `cycle_osd_output` in `cycle.h` which is dynamically allocated in a similar manner to `session_lock_output`. This ensures that all states about alt-tabbing are stored in server->cycle.

Also, this PR fixes a rare memory leak in `output->cycle_osd.items` when an output is destroyed while alt-tabbing, by freeing it when the osd tree is destroyed.

### DIFF
--- a/include/cycle.h
+++ b/include/cycle.h
@@ -4,6 +4,7 @@
 
 #include <stdbool.h>
 #include <wayland-server-core.h>
+#include <wlr/util/box.h>
 #include "config/types.h"
 
 struct output;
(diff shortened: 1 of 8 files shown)


johanmalm:
LGTM

PR #3300, opened 2026-01-01
view: eliminate store_natural_geometry arguments

jlindgren90:
These were added to fix handling of natural geometry for snap-to-edge behavior back in 9021020f6e37 and seemed like a good idea at the time. Since then, the number of call sites has exploded, so it seems more maintainable to put explicit checks for interactive move within the three functions affected.


johanmalm:
LGTM

PR #3303, opened 2026-01-04
protocols: remove wlr-input-inhibitor-unstable-v1

jlindgren90:
It appears we don't support this protocol any more. It was originally used for swaylock, which now uses ext-session-lock-v1 instead.


Consolatis:
Thanks, looks like somebody forgot to remove the xml in 3e614814feaf25a088a2e27eb5db649af7e3036f when we moved to wlroots 0.18.x.

PR #3306, opened 2026-01-04
view: rework & fix last_layout_geometry logic

jlindgren90:
Two goals here:
- make the code easier to follow
- really fix the handling of initially maximized/fullscreen xdg-shell views, which didn't work correctly in all cases in spite of the attempted fix in #3042

The fix is not practical to separate from the refactor, hence the combined PR.

The main idea of the refactor is to combine natural_geometry and last_layout_geometry, because the existing logic around last_layout_geometry seems unnecessarily complicated. As far as I can see, it's basically just a copy of the natural_geometry with some special semantics, specifically:
    
1. it's copied *from* the natural geometry (or equivalently, the pending geometry if floating) at the first layout change after a user-initiated move or resize, and
    
2. it's copied *back to* the natural geometry at subsequent layout changes and used as the "original" position to restore.

I don't see the point of the copying back and forth. Unless I'm missing something, we can achieve the same semantics in a lot less code by just updating/using the natural_geometry itself at the appropriate times. The only additional data needed is a single bool to track whether a layout change has occurred since the last user-initiated move/resize.
    
The first two commits revert #3042, and the last commit implements a better fix for the issue of invalid fallback geometry with initially maximized xdg-shell views.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -7,6 +7,7 @@
 #include <wayland-util.h>
 #include <wlr/util/box.h>
 #include <xkbcommon/xkbcommon.h>
+#include "common/box.h"
 #include "common/edge.h"
 #include "config.h"
 #include "config/types.h"
@@ -219,6 +220,14 @@ struct view {
 	 * layout change. False if it was due to user-initiated action.
 	 */
 	bool adjusted_for_layout_change;
+	/*
+	 * Midpoint of the view's actual geometry prior to adjusting
+	 * for a layout change. For floating views, this will match the
+	 * natural_geometry, but for maximized/tiled/fullscreen views it
+	 * could be on a different output, in which case it is needed to
+	 * restore those views to the correct output.
+	 */
+	struct point saved_midpoint;
 
 	/* used by xdg-shell views */
 	uint32_t pending_configure_serial;
--- a/src/view.c
+++ b/src/view.c
@@ -446,18 +446,12 @@ view_get_edge_snap_box(struct view *view, struct output *output,
 }
 
 static bool
-view_discover_output(struct view *view, struct wlr_box *geometry)
+view_discover_output(struct view *view, struct point midpoint)
 {
 	assert(view);
 
-	if (!geometry) {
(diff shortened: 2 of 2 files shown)


jlindgren90:
Lightly tested, but more testing appreciated.
/cc @ahesford and @tokyo4j who have worked on this code previously.

ahesford:
It's been awhile since I thought about these problems, but I probably created `last_layout_geometry` as a safety measure because I was introducing new behavior and didn't want to inadvertently break something that relies on `natural_geometry`. Combining them makes sense and probably helps avoid extra cache-invalidation problems.

jlindgren90:
Correction to the description: the case of "maximized view interactively dragged to a different output before the layout changes" still doesn't work correctly, because we're choosing an output based on the natural geometry which remains on the old output. It's a pre-existing issue and would require further changes to fix properly, so I think it's okay to leave as-is for now.

tokyo4j:
I haven't reviewed at the diff, but at least I tested this PR and confirmed that this PR doesn't re-introduce the problem addressed by #3042; the position of an initially-maximized window is properly restored after re-plugging the output the window belongs to. I didn't like having both of `view->last_layout_geometry` and `view->natural_geometry` either and I really like how this PR cleans it up!

jlindgren90:
I have an idea to solve the issues with maximized/full screen views more generally. Will mark this draft for now to think about it more.

jlindgren90:
I replaced the last commit (fix for initially maximized/fullscreen xdg-shell views) with a different approach, which is to save and use the midpoint of the actual (not natural) geometry to pick the output after layout changes.

I think this is more robust, and it also fixes the case of "maximized view interactively dragged to a different output before the layout changes" for real this time.

jlindgren90:
Hmm, I wonder why we don't just save the actual (pending) geometry to begin with, rather than the natural geometry?

jlindgren90:
Abandoning this in favor of a different approach:
- #3309

Sorry for the noise, I think the end result will be better though.

PR #3307, opened 2026-01-05
scripts/check: run max $(nproc) checkpatch.pl processes

domo141:
Each of those checking "number-of-files" / $(nproc) files.

This is significantly faster and uses less resources than running checkpatch.pl separately for every file.

To count the number of files, the file name list was first written to a temporary file.

There is chance, albeit improbable, that shell fails to execute an exit trap when its execution is interrupted. Therefore before the last use of this temporary file shell opens it for fd 8 and then removes its directory entry (along with the exit trap) and finally redirects the file content to xargs with <&8.

--

On one system 4-core system I tested to check all 208 files, best time now 22 secs - before 34 secs.

Cold machine, warm cache, fast SSD 

tried also `echo 3 > /proc/sys/vm/drop_caches` and/or 8 parallel processes but did not see much difference :o

anyway, before 208 forks, execves and perl inintialisations and arg parsing, each checking one file,
now 4 forks, execves... and each checking 52 files

### DIFF
--- a/scripts/check
+++ b/scripts/check
@@ -20,7 +20,7 @@ run_checks () {
 	fi
 
 	find src/ include/ clients/ t/ \( -name "*.c" -o -name "*.h" \) -type f -print0 |
-	nice xargs -0 --max-args 1 --max-procs $(nproc) \
+	nice xargs -0 --max-args 16 --max-procs $(nproc) \
 		scripts/checkpatch.pl --terse --no-tree --strict --file
 	return $?
 }


johanmalm:
I find this one hard to review. Welcome help from others.

In terms of speeding up the script, should we consider just running it on the patch itself - or limit it to the files affected by the PR. 

I very rarely run it on the whole project, but just on the files I'm working on.

For the CI I imagine analysis of the patch makes most sense.

Consolatis:
Isn't there some kind of limit for `argv`; with a single core, this seems to execute a single checkpatch instance with > 208 args?

> This is significantly faster

Debatable, in the context of now running in 22 sec vs 34 secs it is obviously a massive improvement.
However, in the context of CI (and assumingly most users) those 12 seconds don't really matter.

I find it hard to read as well, especially with the exit trap, alternatives like this are not really any better in that regard, sh doesn't seem to have any way to lseek. Edit: actually there is: just use `< /proc/self/fd/x` which reopens the fd.
<details><summary>off-topic shell impl</summary>

```sh
#!/bin/sh
files=$(mktemp /tmp/lawc_check.XXXXXX)
[ -z "$files" ] && exit 1
exec 7>"$files"
exec 8<"$files"
exec 9<"$files"
rm -- "$files"
echo "this is a test line" >&7
echo "characters in line: $(wc -c <&8)"
echo "line was: $(cat <&9)"
```

```sh
#!/bin/sh
files=$(mktemp /tmp/lawc_check.XXXXXX)
[ -z "$files" ] && exit 1
exec 7>"$files"
rm -- "$files"
echo "this is a test line" >&7
echo "characters in line: $(wc -c < /proc/self/fd/7)"
echo "line was: $(cat < /proc/self/fd/7)"
```

</details>

Edit:
To answer myself:
> Isn't there some kind of limit for `argv`;

Yes, there is: `getconf ARG_MAX` is the amount of bytes allowed for combined environment and arguments.
It is apparently set to 2MiB by default.

Edit 2:
Did some unscientific tests on my high end system:
```
      this pr: real    2m15.281s
       master: real    2m19.178s

--max-args  2: real    1m53.492s
--max-args  5: real    1m37.596s
--max-args 10: real    1m31.858s
--max-args 20: real    1m29.274s
--max-args 30: real    1m28.702s
--max-args 40: real    1m49.698s
```
The reason it is only marginally faster with this PR is that for like 60% of the time only two cores were working because the other two were already done with their files (mostly headers). Just adding something like `--max-args 10` seems way more beneficial and much easier to read.

domo141:
> I find this one hard to review. Welcome help from others.
> 
> In terms of speeding up the script, should we consider just running it on the patch itself - or limit it to the files affected by the PR.
> 
> I very rarely run it on the whole project, but just on the files I'm working on.
> 
> For the CI I imagine analysis of the patch makes most sense.

Originally I had thought how to do something to check just the changed files in question,
but that would required to basically rewrite the script (in an idea that it could later do some
other checks too...) so then I just ended up with this (while doing I also hacked `checkpatch.pl`
to do the parallelism -- which was even a bit faster but more intrusive) -- anyway just now
the idea was to lessen the "seconds" spent on github CI...

In the case of checking all changed files in a github PR would require some more investigation
what would be the correct `revision range` in `git log` command line:

`git --no-pager log --name-only --format= HEAD^.. 'src/**.[ch]' 'include/**.[ch]' 'clients/**.[ch]' 't/**.[ch]'`

(note `--format= ` w/ space following) works for local HEAD commit, and

`git --no-pager log --name-only --format= @{u}.. 'src/**.[ch]' 'include/**.[ch]' 'clients/**.[ch]' 't/**.[ch]'`

since "upstream" -- that could find just the changed  c/h files in those particular directories in the whole commit stack of one PR -- "could" since the checkout in gh may work in peculiar way
(in my local labwc copy I git-am patches to a detached head in a git worktree and that @{u} fails
 there for me - happen to be that way by accident) -- and to test how gh behaves one needed to
do a PR with at least 2-commits, log a lot and investigate CI logs...

domo141:
For @Consolatis comment...

In case of Linux, one could do `exec 8<>$file` and then `rm $file` and then have read-write
access to the file via `/dev/fd/8` (Used it successfully in handful of shell scripts so far)
The one I presented could work also in FreeBSD (same grep and xargs opts works according
to man pages) -- on NetBSD (if anyone cared) `grep -z` seems not be (same) option.

The tesst with `--max-args` are interesting - I have to try my self (fun!). Sure, if
`--max-args 10` brings similar results then that would sure be much simpler change.

Your execution times were quite long. Just running one checkpatch.pl with all 
the 208 files given at once from command line executes in 51 seconds on my machine...

domo141:
TL;DR;

I pushed an update -- just `s/max-args 1/max-args 16/` is now the change. nice round 2**4
number. I spent some time checking things (mostly fun, what else can nerd to ;/). wrote some
thoughts to the commit  message. maybe someone(tm) is interested of the message there...:)

-- I wrote too much below - I thought just a simple thing, but IT happens...

In my tests, on this laptop, I got same, 21secs with --max-args 52 and --max-args 16 (and 10 and 20),
but on other, desktop PC, 21secs were (also) the result with 16, 10 and 20 -- but with --max-args 52
the run-time was 27 seconds (and I did all these in all these systems several times). 
My quess is that the processor in that desktop pc is slower, but when enough load is on that laptop
(>300% on 4-core system), it throttles enough to make it as slow as the 52-case (200%). 
And, in that 52-case, first cpu was ready after 8secs, second after 12, third after 19.
-- this (suspected) throttling made me not see it, and understand how xargs distribute the args..

Thank you @Consolatis for this:
```
 files=`mktemp`; exec 7>$files 8<$files 9<$files; rm -f $files
```
Tried that over the 1st version of this pr -- worked fine -- while not needed here
but I can use it elsewhere (where 8</dev/fd/8 is bad example for portability...)

In reply to @johanmalm  I mentioned some command lines to get filelists for commts
range. That was not good enough. One of the working one could be:
```
git diff-tree -r --no-commit-id --name-only --diff-filter=AM @{u}..
```
That doesn't list removed files (any files no longer on clean worktree),
and shows renamed (R) files as added (A) (or would if `--name-status` were there).

I think the current change is small improvement of the execution there. While 
checking things I looked into checkpatch.pl a bit deeper (it just `diff -u /dev/null {file}`
if content not given as patch) -- in case of CI one could provide the diffs from git
-- or it may be checkpatch.pl can do it (I resisted the temptation to check).

Consolatis:
Thanks for your investigation. We can always consider ways to do this even faster but this is a nice incremental and easy update.

PR #3308, opened 2026-01-05
CI: disable no-font check

Consolatis:
Void made the foot package depend on the font.

Running this test now always fails with
> `dejavu-fonts-ttf-2.37_3` in transaction breaks installed pkg `foot-1.25.0_2`

As this test uses both and the feature itself
seems to work well lets just skip it for now.


johanmalm:
Ta

PR #3309, opened 2026-01-05
view: rework saving/restoring geometry across layout changes

jlindgren90:
After several iterations, this is basically a complete re-work. The old implementation was difficult to follow and sometimes failed to restore fullscreen/maximized/tiled geometry correctly, since it was based entirely on natural (floating) geometry.

The new implementation:
 - always saves the actual (pending) geometry at first layout change
 - explicitly tracks whether a view has moved between outputs
 - consolidates invalidating the saved geometry into one place, rather than having lots of invalidate() calls sprinkled everywhere

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -215,11 +215,14 @@ struct view {
 	struct wlr_box natural_geometry;
 	/*
 	 * Whenever an output layout change triggers a view relocation, the
-	 * last pending position (or natural geometry) will be saved so the
-	 * view may be restored to its original location on a subsequent layout
-	 * change.
+	 * last pending position will be saved so the view may be restored
+	 * to its original location on a subsequent layout change.
 	 */
 	struct wlr_box last_layout_geometry;
+	/* Set temporarily when moving view due to layout change */
+	bool adjusting_for_layout_change;
+	/* True if original output was disconnected or disabled */
+	bool lost_output_due_to_layout_change;
 
 	/* used by xdg-shell views */
 	uint32_t pending_configure_serial;
@@ -533,7 +536,6 @@ bool view_titlebar_visible(struct view *view);
 void view_set_ssd_mode(struct view *view, enum lab_ssd_mode mode);
 void view_set_decorations(struct view *view, enum lab_ssd_mode mode, bool force_ssd);
 void view_toggle_fullscreen(struct view *view);
-void view_invalidate_last_layout_geometry(struct view *view);
 void view_adjust_for_layout_change(struct view *view);
 void view_move_to_edge(struct view *view, enum lab_edge direction, bool snap_to_windows);
 void view_grow_to_edge(struct view *view, enum lab_edge direction);
--- a/src/view.c
+++ b/src/view.c
@@ -588,6 +588,21 @@ view_move_resize(struct view *view, struct wlr_box geo)
 	if (view->impl->configure) {
 		view->impl->configure(view, geo);
 	}
+
+	/*
+	 * If the move/resize was user-initiated (rather than due to
+	 * output layout change), then invalidate the saved geometry.
+	 *
+	 * TODO: consider also updating view->output here for floating
+	 * views (based on view->pending) rather than waiting until
+	 * view_moved(). This might eliminate some race conditions with
+	 * view_adjust_for_layout_change(), which uses view->pending.
+	 * Not sure if it might have other side-effects though.
+	 */
+	if (!view->adjusting_for_layout_change) {
+		view->last_layout_geometry = (struct wlr_box){0};
+		view->lost_output_due_to_layout_change = false;
+	}
 }
 
 void
@@ -1405,9 +1420,6 @@ view_maximize(struct view *view, enum view_axis axis)
 		 * a maximized view.
 		 */
 		interactive_cancel(view);
-		if (store_natural_geometry && view_is_floating(view)) {
-			view_invalidate_last_layout_geometry(view);
-		}
 	}
 
 	/*
@@ -1695,7 +1707,6 @@ view_set_fullscreen(struct view *view, bool fullscreen)
 		 */
 		interactive_cancel(view);
 		view_store_natural_geometry(view);
-		view_invalidate_last_layout_geometry(view);
 	}
 
 	set_fullscreen(view, fullscreen);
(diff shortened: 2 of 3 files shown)


johanmalm:
LGTM - not tested.

jlindgren90:
@tokyo4j - care to review/re-test?

src/view.c
```diff
@@ -1713,138 +1725,66 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
 	bool is_floating = view_is_floating(view);
-	bool use_natural = false;
-
-	if (!output_is_usable(view->output)) {
-		/* A view losing an output should have a last-layout geometry */
-		update_last_layout_geometry(view);
-	}
+	bool saved = view->saved_geometry_valid;
+	bool lost_output = view->lost_output_due_to_layout_change
+		|| !output_is_usable(view->output);
```
tokyo4j: What's the purpose of `view->lost_output_due_to_layout_change`? What happens if we just do `bool lost_output = !output_is_usable(view->output);`.

src/view.c
```diff
@@ -1713,138 +1725,66 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
+	 *   is reconnected.
+	 *
+	 * If so, try to find the "nearest" output (in terms of layout
+	 * coordinates) to the saved geometry. Other strategies might be
+	 * possible here -- for example, trying to match the same physical
+	 * output the view was on previously -- but this is simplest.
 	 */
-	if (is_floating || last_geometry || !output_is_usable(view->output)) {
-		/* Move the view to an appropriate output, if needed */
-		bool output_changed = view_discover_output(view, last_geometry);
```
tokyo4j, line 1811: We can let `view_discover_output()` return `void` as we no longer use its return value.

tokyo4j:
I commented on what I noticed at first glance, but it may take me a while to fully understand the changes in this PR.

src/view.c
jlindgren90: Maximized/full screen outputs won't return to their original output once it's reconnected, if they got assigned to a different output in the meantime (e.g. internal laptop display if undocked and then docked again).

The current implementation has similar logic if you look close enough, but it isn't obvious.

src/view.c
jlindgren90: Yep, I could go either way on that - remove the return value or leave it in case we want it again in future.

tokyo4j: I don't have a strong preference on it, either.

tokyo4j:
Could you review #3310? I took some hints from this PR and then replaced `view->saved_geometry` with output name and output-relative view geometry.

src/view.c
```diff
@@ -1713,138 +1725,66 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
-		}
-
-		/*
-		 * Whether or not the view has moved, the layout has changed.
-		 * Ensure that the view now has a valid last-layout geometry.
-		 */
-		update_last_layout_geometry(view);
+	if (is_floating || lost_output) {
+		view_discover_output(view, &view->saved_geometry);
 	}
```
tokyo4j: What happens if we always call `view_discover_output()`? If we can remove `lost_output` here, we can also remove `view->lost_output_due_to_layout_change`.

include/view.h
```diff
@@ -214,12 +214,14 @@ struct view {
...
-	 * view may be restored to its original location on a subsequent layout
-	 * change.
+	 * Geometry of the view prior to adjusting for layout changes.
+	 * This is valid only when the most recent move/resize of the
+	 * view was due to a layout change, and is invalidated by any
+	 * user-initiated move/resize.
 	 */
-	struct wlr_box last_layout_geometry;
+	struct wlr_box saved_geometry;
+	bool saved_geometry_valid;
```
tokyo4j: I prefer using `!wlr_box_empty(saved_geometry)` rather than adding another flag.

I also suggest keeping the name `last_layout_geometry` if this PR doesn't change its semantics, as I will rename them anyway in #3310.

src/view.c
jlindgren90: If we always call `view_discover_output()` then we may unnecessarily move maximized or fullscreens views to a different output. Existing behavior is that we try to keep them on the same output if outputs are rearranged in the layout.

src/view.c
```diff
@@ -588,6 +588,22 @@ view_move_resize(struct view *view, struct wlr_box geo)
...
+	 * (rather than due to output layout change) and invalidate the
+	 * saved geometry. For the few cases where the move/resize is
+	 * due to layout change, these flags are set again afterward by
+	 * view_adjust_for_layout_change().
+	 *
+	 * TODO: consider also updating view->output here for floating
+	 * views (based on view->pending) rather than waiting until
+	 * view_moved(). This might eliminate some race conditions with
+	 * view_adjust_for_layout_change(), which uses view->pending.
+	 * Not sure if it might have other side-effects though.
```
tokyo4j, line 600: I don't think we should updated `view->output` here since `view_move_resize()` doesn't update the actual window geometry.

EDIT: Hm, I'm not sure about this... let me think for a while

jlindgren90: It updates `view->pending` so my thought is that we would use that rather than `view->current`. But I don't think we need to make that change right now.

tokyo4j: OK, ignore me. I'm still not happy about updating `view->output` without updating `view->current` (which means `view->output` and `view->outputs` may be de-synchronized), but `view_adjust_for_layout_change()` actually does it and we will need to sort this in the future. It's worth mentioning.

src/view.c
```diff
@@ -1713,138 +1725,66 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
+		struct wlr_box geometry = view->saved_geometry;
+		adjust_floating_geometry(view, &geometry,
+			/* midpoint_visibility */ true);
+		view_move_resize(view, geometry);
 	}
 
+	/*
+	 * view_move_resize() always clears these two flags.
+	 * Set them back to correct values now that we're done.
+	 */
```
tokyo4j: I think you already know, but I suggest using a flag (like `doing_layout_change` in #3310) rather than setting flags again here.

src/view.c
tokyo4j: OK, then I can remove it later in #3310 which tries to keep windows in the same output no matter it's floating or not.

tokyo4j:
LGTM other than my two open comments above.

include/view.h
jlindgren90: Done

src/view.c
jlindgren90: Done - I decided to add it in `struct view` rather than a file-scope static though

PR #3310, opened 2026-01-08
view: simplify and improve layout geometry restoration

tokyo4j:
Supersedes #3309
Supersedes #3214 (maybe it doesn't if we should resize the window to fit in the output, but we can do it later)

This PR is about simplifying the logic of saving and restoring view geometry on layout changes, and making it more robust.

The first commit handles layout changes in idle context. This is mainly just a preparation for the next commit, but I find this approach more straightforward for deduplicating layout change handlings. Please keep in mind that I'm not much familiar with output managements. Review & testing appreciated.

The second commit in the main part. First it removes many `view_invalidate_last_layout_geometry()` calls scattered in e.g. `view_maximize()` and `view_set_fullscreen()` and instead clears the saved geometry in `view_move_resize()`. The idea is took from #3309. This itself does not affect the behavior.

Second, it replaces `view->last_layout_geometry` with the output name and output-relative view geometry in `view->output_ctx`. This allows more robust restoration of view geometry on layout changes. For example, when there are outputs WL-1, WL-2, and WL-3, and WL-1 has a view, disabling WL-1 and WL-2 and then re-enabling WL-1 should now correctly restore the relative view geometry on WL-1. Also, relative view positions in outputs are now preserved when the output layout changes.

`view->output_ctx` is always saved before output layout changes (in `schedule_output_layout_change()`), and cleared in `view_move_resize()` as mentioned above. As saving `output_ctx` requires `view->output` to be set, I added `schedule_output_layout_change()` at the start of `handle_output_destroy()` (before clearing `view->output`).

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -581,6 +581,8 @@ view_moved(struct view *view)
 	}
 }
 
+static void clear_last_placement(struct view *view);
+
 void
 view_move_resize(struct view *view, struct wlr_box geo)
 {
@@ -600,8 +602,7 @@ view_move_resize(struct view *view, struct wlr_box geo)
 	 * Not sure if it might have other side-effects though.
 	 */
 	if (!view->adjusting_for_layout_change) {
-		view->last_placement.layout_geo = (struct wlr_box){0};
-		view->lost_output_due_to_layout_change = false;
+		clear_last_placement(view);
 	}
 }
 
@@ -1744,7 +1745,11 @@ save_last_placement(struct view *view)
 		wlr_log(WLR_ERROR, "cannot save last placement in unusable output");
 		return;
 	}
+	xstrdup_replace(view->last_placement.output_name, output->wlr_output->name);
 	view->last_placement.layout_geo = view->pending;
+	view->last_placement.relative_geo = view->pending;
+	view->last_placement.relative_geo.x -= output->scene_output->x;
+	view->last_placement.relative_geo.y -= output->scene_output->y;
 }
 
 void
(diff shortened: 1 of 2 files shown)


jlindgren90:
> The first commit handles layout changes in idle context

Is this required for the rest of the changes? I didn't see why.

src/output.c
```diff
@@ -890,20 +897,34 @@ wlr_output_configuration_v1 *create_output_config(struct server *server)
...
+	output_update_for_layout_change(server);
+	seat_output_layout_changed(&server->seat);
+	server->output_layout_change_idle = NULL;
+}
+
+static void
+schedule_output_layout_change(struct server *server)
+{
+	if (server->output_layout_change_idle) {
+		return;
```
jlindgren90: It seems like there is a race if a user-initiated move/resize happens when the idle source is already scheduled. Then `view_save_output_ctx` will not run again before the layout change is processed, and the user change will potentially be reverted. Right?

jlindgren90:
Arch build is failing due to LeakSanitizer finding a memory leak at exit:

```
==10386==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x7fdaf4e3d5dd in calloc /usr/src/debug/gcc/gcc/libsanitizer/asan/asan_malloc_linux.cpp:74
    #1 0x7fdaf4d10369 in wl_event_loop_add_idle (/usr/lib/libwayland-server.so.0+0x8369) (BuildId: b9664217748f523995e3f855fa197cf8e59942d1)
    #2 0x555561ef8336 in schedule_output_layout_change ../src/output.c:922
    #3 0x555561eef592 in handle_output_destroy ../src/output.c:176
    #4 0x7fdaf4d104cf in wl_signal_emit_mutable (/usr/lib/libwayland-server.so.0+0x84cf) (BuildId: b9664217748f523995e3f855fa197cf8e59942d1)
    #5 0x55556216f344 in wlr_output_finish ../subprojects/wlroots/types/output/output.c:380
    #6 0x55556214e61f in output_destroy ../subprojects/wlroots/backend/headless/output.c:94
    #7 0x555562170678 in wlr_output_destroy ../subprojects/wlroots/types/output/output.c:436
    #8 0x55556214d4cb in backend_destroy ../subprojects/wlroots/backend/headless/backend.c:40
    #9 0x55556209e054 in wlr_backend_destroy ../subprojects/wlroots/backend/backend.c:68
    #10 0x55556211664f in multi_backend_destroy ../subprojects/wlroots/backend/multi/backend.c:62
    #11 0x55556209e054 in wlr_backend_destroy ../subprojects/wlroots/backend/backend.c:68
    #12 0x555561f1c615 in server_finish ../src/server.c:792
    #13 0x555561eecc41 in main ../src/main.c:277
```

jlindgren90:
This seems like a bigger/riskier change than #3309, and I wonder if we could merge that one first.

I do like the `static bool doing_layout_change` logic, it's cleaner and I can use that in #3309 also.

src/output.c
tokyo4j: I'm not sure a user-initiated move/resize can be injected before the execution of the idle context. But maybe it can technically happen. I removed the idle context anyway.

tokyo4j:
> > The first commit handles layout changes in idle context
> 
> Is this required for the rest of the changes? I didn't see why.

You are right. We can instead just call `view_save_output_ctx()` (now updated to `views_save_output_ctx()`) for all views before layout changes output destruction, and it turned out to be more robust.

I reverted the most changes in `output.c`.

src/view.c
```diff
@@ -825,6 +839,7 @@ view_compute_centered_position(struct view *view, const struct wlr_box *ref,
 	return true;
 }
 
+/* Make sure the view is visible in view->output */
```
jlindgren90: Thanks for adding the comment. Nit: maybe saying "Make sure the passed-in view geometry is visible..." would make it more clear that this function doesn't actually move the view.

src/view.c
```diff
@@ -1713,137 +1724,95 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
+	struct wlr_box new_geo = view->current;
+	if (!restore_to_output_ctx(view, &new_geo)) {
 		/*
-		 * Whether or not the view has moved, the layout has changed.
-		 * Ensure that the view now has a valid last-layout geometry.
+		 * Or evacuate the view to another output if its output is
+		 * being removed from the layout.
 		 */
-		update_last_layout_geometry(view);
+		view_discover_output(view, NULL);
```
jlindgren90: If the output-name-based restore logic fails, I wonder if it would still make sense to try to discover the new output based on saved layout geometry, rather than `view->current` (which might have been affected by other layout changes). The use-case would be unplugging a monitor and connecting it to e.g. a different USB/DP port, which generally changes the output name. Or a laptop traveling between different (home/work) docking stations.

It may be a bit of a niche case, but since it would technically be a regression vs. current behavior (if I understand the logic correctly), I think it's worth mentioning.

jlindgren90: So what I have in mind as an ideal solution would be a combination of this and #3309, where we would try the output-name-based lookup first (or maybe an EDID-based output match) and fall back to layout-based geometry restore if that fails. In which case I still think we could start with the changes in #3309 (or a refined version of them) and add the output-matching on top afterward.

src/view.c
tokyo4j: Thanks. Updated the comment.

src/view.c
tokyo4j: I think it makes sense. Added `view->last_placement.layout_geo`.

This may be also helpful when the reconnected output appears as a different output for some reason, which reportedly happens in some USB-HDMI connector: https://bugs.kde.org/show_bug.cgi?id=450068

> In which case I still think we could start with the changes in https://github.com/labwc/labwc/pull/3309 (or a refined version of them) and add the output-matching on top afterward.

That sounds sensible. I'll add some comments in #3309.

tokyo4j:
Replaced `view->output_ctx` with:

```c
struct {
		char *output_name;
		struct wlr_box relative_geo;
		struct wlr_box layout_geo;
	} last_placement;
```

I also updated it to save `view->pending` rather than `view->current`, as we currently do it and I don't think it makes noticeable changes.

tokyo4j:
Rebased.

src/view.c
```diff
@@ -1724,61 +1726,99 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
-	 * possible here -- for example, trying to match the same physical
-	 * output the view was on previously -- but this is simplest.
-	 */
-	if (is_floating || view->lost_output_due_to_layout_change) {
-		view_discover_output(view, &view->last_layout_geometry);
+	if (!output_is_usable(output)) {
+		wlr_log(WLR_ERROR, "cannot save last placement in unusable output");
+		return;
+	}
+	view->last_placement.output_name = xstrdup(output->wlr_output->name);
```
jlindgren90: I guess the `wlr_box_empty` check earlier means that `output_name` should be NULL here, but you could also use `xstrdup_replace` to be sure.

src/view.c
```diff
@@ -1724,61 +1726,99 @@ view_set_fullscreen(struct view *view, bool fullscreen)
...
+}
+
+void
+view_adjust_for_layout_change(struct view *view)
+{
+	assert(view);
+
+	view->adjusting_for_layout_change = true;
+
+	if (wlr_box_empty(&view->last_placement.layout_geo)) {
```
jlindgren90, line 1777: If you move this check before `view->adjusting_for_layout_change = true;`, then you could just `return` and eliminate the `goto`.

tokyo4j: Nice catch. Done.

src/view.c
tokyo4j: Done.

PR #2956, opened 2025-07-30
Update for wlroots 0.20

Consolatis:
Replacement PR for #2892.
The only difference is that this branch is now part of the labwc repo so new PRs can target it.

---

You know the drill, not intended to be merged anytime soon.
Mainly to ease incremental chase updates so we don't end up with a giant pile of work when we start tracking 0.20.

Not suggested to use as daily driver but pointing out issues in this PR is obviously welcome. The last commit always tracks the wlroots master branch so issues are expected when wlroots breaks API. List of breaking changes from wlroots is [here](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3977).

This PR may also serve as a base for other PRs to test new wlroots things like toplevel capture.

Known defects as of today:
- none

Push policy: feel free to push commits to this PR but please don't force-push, I'll handle the rebasing.

PR #3312, opened 2026-01-08
Chase/0.20

ManuLinares:
- changes for wl-roots wlr_xwayland_set_cursor
- also has support for color-management-v1

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -1200,10 +1200,9 @@ xwayland_server_init(struct server *server, struct wlr_compositor *compositor)
 		server->seat.xcursor_manager, XCURSOR_DEFAULT, 1);
 	if (xcursor) {
 		struct wlr_xcursor_image *image = xcursor->images[0];
-		wlr_xwayland_set_cursor(server->xwayland, image->buffer,
-			image->width * 4, image->width,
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks, I'll look a bit into the wlroots cursor API change and then potentially remove the whole function if it turns out its not required anymore because wlroots now creates a copy of the pixel data.

ManuLinares:
> Thanks, I'll look a bit into the wlroots cursor API change and then potentially remove the whole function if it turns out its not required anymore because wlroots now creates a copy of the pixel data.

Great! Just let me know and I close this.

Consolatis:
The xcursor change should be handled by #2956.

Regarding the color manager: what benefit does it bring in practice?

Note that we usually do not include support for new protocols in the chase PR itself but you could open a new PR against the `chase/0.20` branch which we can handle once the master branch starts to track wlroots 0.20.

ManuLinares:
> The xcursor change should be handled by #2956.
> 
> Regarding the color manager: what benefit does it bring in practice?
> 
> Note that we do usually not include support for new protocols in the chase PR itself but you could open a new PR against the `chase/0.20` branch which we can handle once the master branch starts to track wlroots 0.20.

thank you.

Issue #3315, opened 2026-01-11
Game Issues: Minecraft problem with lwjgl 2.x and the cursor.


Terromur:
### Problem description

1. Starting with the very first Alpha version (a1.0.1_01) and ending with 1.5.2 (Release), the problem manifests itself in that when left-clicking/right-clicking, the camera in the game centers sharply and for 1 millisecond the cursor appears. When using WASD/space bar and other keys while clicking the mouse, the keys release themselves. The problem only occurs in windowed mode. The problem is fixed if you play at 800x600 and the game is centered.
2. Starting with the Classic version (c0.0.14a_08) and ending with the latest infdev (inf-200100630-1835 (1.369)), the problem manifests itself in that when you click the left mouse button/right mouse button, the main menu simply appears. Problem saved and with fullscreen.

The problem appeared with version 0.9.0 and continues with the latest version. I tried GNOME and found that they have the same problem, but there is no problem with KDE Plasma, niri, or river.

Tested on Freesmlauncher/PrismLauncher/Betacraft Launcher/XMCL Launcher (X Minecraft Launcher).

I decided to check which commit caused this problem. It is related to these changes: [link](https://github.com/labwc/labwc/commit/8fb2ecefcbd66c4836f47d3b9451c8a380b6016f)

Temporary solution: Downgrade to labwc 0.8.4, or try replacing LWGL with version 3.x (I don't know how to do this, but the problem disappeared on Better Than Adventure/Reindev versions Beta 1.7.3 and with Babric beta 1.7.3, after installing the necessary mods). 

Temporarily stopped at 0.9.2 version by making a patch, returning VIEW_WANTS_FOCUS_ALWAYS  VIEW_WANTS_FOCUS_OFFER

### Steps to reproduce

1. Launch labwc 0.9.0-0.9.3/labwc-git or labwc-git (chase/0.20)
2. Launch vanilla Minecraft 1.5.2 or older version

### labwc build source

Release

### labwc version

0.9.3

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

CachyOS

PR #3316, opened 2026-01-11
xwayland: Do not try to focus a window that was already in focus

ventureoo:
Old versions of Minecraft (<= 1.5.2) take input, but don’t set WM input hint, which leads to labwc trying to focus window on every click in the game. We had not this issue before 8fb2ece because `xwayland_view_wants_focus()` returns `VIEW_WANTS_FOCUS_ALWAYS` , and `set_or_offer_focus()` already checks if the focus was received before in this case.

Fixes: #3315

### DIFF
--- a/src/desktop.c
+++ b/src/desktop.c
@@ -55,7 +55,9 @@ set_or_offer_focus(struct view *view)
 		break;
 	case VIEW_WANTS_FOCUS_LIKELY:
 	case VIEW_WANTS_FOCUS_UNLIKELY:
-		view_offer_focus(view);
+		if (view->surface != seat->seat->keyboard_state.focused_surface) {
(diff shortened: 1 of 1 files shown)


Terromur:
My problem has been solved thanks to ventureoo! The problem disappeared in 0.9.3, master branch, and chase/0.20.

Consolatis:
Thanks for the bug report and investigation and thanks for the PR. It looks correct to me but I am not that deep into the X11 focus semantics so CC @jlindgren90 as our X11 focus expert.

jlindgren90:
Thanks, I did not test but the change looks fine to me. I don't remember any specific reason we didn't have that check already and adding it is consistent with the other switch branch, so I'm up for giving it a try.

PR #3323, opened 2026-01-16
view: avoid repeated focus changes unminimizing parent/child views

jlindgren90:
When unminimizing a group of N parent/child views, we currently end up triggering N focus changes (as well as O(N^2) surface restackings) due to calling `desktop_focus_view()` for each view in turn.

I'm not aware of any real issues this is causing, but it's unnecessary and easy to fix. Since `desktop_focus_view()` already raises all sibling views together via `view_move_to_front()`, `view_minimize()` simply needs to call it only once at the very end, once all views are visible.

Test cases:
- Audacious with floating plugin views (XWayland)
- xfce4-terminal with About dialog (xdg-shell)

### DIFF
--- a/src/view-impl-common.c
+++ b/src/view-impl-common.c
@@ -11,6 +11,11 @@ view_impl_map(struct view *view)
 {
 	view_update_visibility(view);
 
+	/* Leave minimized, if minimized before map */
+	if (!view->minimized) {
+		desktop_focus_view(view, /* raise */ true);
+	}
+
 	if (!view->been_mapped) {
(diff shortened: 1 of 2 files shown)


jlindgren90:
Hmm... probably should do the same thing when minimizing (avoid calling `desktop_focus_topmost_view` repeatedly).

jlindgren90:
This should be all set now.

jlindgren90:
@tokyo4j thoughts on this approach? I know you were involved in adding `view_update_visibility()` previously.

johanmalm:
Thanks. LGTM. Going for the merge.

PR #3325, opened 2026-01-19
view: avoid restacking when view (or a sub-view) is already in front

jlindgren90:
Currently, every click within a sub-view results in first restacking the parent view in front, and then the sub-view. This is unnecessary and has caused issues in the past, such as with Xaw popups (which we've worked around in d748dc78bc0e by adding an `xcb_flush()`).

It would be better not to do the unnecessary restacking at all.

Note: this fixes the Xaw popup issue from #3152 on its own, and means we could probably revert d748dc78bc0e if we wanted. Not sure if we want to do that yet, or later, or at all.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2232,6 +2232,18 @@ void
 view_move_to_front(struct view *view)
 {
 	assert(view);
+	struct server *server = view->server;
+	assert(!wl_list_empty(&server->views));
+
+	/*
+	 * Check whether the view is already in front, or is the root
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM

Looks like another good one :smile:

I'm not sure about the `xwayland_flush()`. The bug associated with Imagemagick's `display` command in #3152 was intermittent, so maybe it's safer to leave as is?

jlindgren90:
I'm good with leaving the flush() in place, doesn't do any harm.

PR #3334, opened 2026-01-24
action: support `NextWindow` option `window="root|all"`

johanmalm:
This enables all windows, not just topmost (root) ones, to be included in the window-switcher. This is helpful when cycling through Gtk applications with open dialogs because although these child windows will always be shown on top of their parents, they do not get keyboard focus with the default configuration.

This also applices to `PreviousWindow`.

Fixes: #3332
Suggested-by: @airtower-luna

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -125,8 +125,8 @@ Actions are used in menus and keyboard/mouse bindings.
 	Resize and move the active window back to its untiled or unmaximized
 	position if it had been maximized or tiled to a direction or region.
 
-*<action name="NextWindow" workspace="current" output="all" identifier="all" />*++
-*<action name="PreviousWindow" workspace="current" output="all" identifier="all" />*
+*<action name="NextWindow" workspace="current" output="all" identifier="all" window="root"/>*++
+*<action name="PreviousWindow" workspace="current" output="all" identifier="all" window="root"/>*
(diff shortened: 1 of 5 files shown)


johanmalm:
Tested with mousepad's About dialog with the following in `rc.xml`

```xml
<keyboard>
  <default/>
  <keybind key="A-Tab">
    <action name="NextWindow" window="all"/>
  </keybind>
</keyboard>
```

tokyo4j:
LGTM.

I'm a bit afraid that `window` sounds too generic, but I don't come up with a better name.

johanmalm:
Thanks.

How about `windowType` or `windowFilter` or `windowCriteria`

Consolatis:
Or maybe something like `dialogs="yes|no"` or `includeDialogs="yes|no"`?

tokyo4j:
This may sound aggressive, but how about #3339? That PR shows non-dialog child windows in the window switcher rather than adding an option. Child windows that are explicitly marked as dialogs (e.g. dialog in `featherpad`), which grabs keyboard focus, are kept hidden in the window switcher, while non-dialog child windows (e.g. file-chooser dialog in `firefox`) are shown in the window switcher, addressing the complaint in #3332.

If we go with this PR, +1 for `includeDialogs="yes|no"`.

johanmalm:
#3339 better

PR #3339, opened 2026-01-27
 cycle: show non-dialog child windows in window switcher 

tokyo4j:
Supersedes #3334.

Before this PR, the window switcher skipped all the child windows.

However, as child windows not marked as modal dialogs can lose focus (ref. `desktop_focus_view()`), it will make sense to include them in the window switcher so that users can refocus them with keyboard.

This behavior follows KWin.


johanmalm:
Yes, this is a better solution. It feels like this is how it should be.

I had forgotten that we had the modal_dialog method.

airtower-luna:
Works well in 0.9.4, thank you! :slightly_smiling_face:

PR #3340, opened 2026-01-28
view: don't update top layer visibility in `view_discover_output()`

tokyo4j:
This PR does not change any behaviors.

We don't need to call `desktop_update_top_layer_visibility()` in `view_discover_output()` since it's called in `view_update_outputs()`.


johanmalm:
Thanks.

Issue #3344, opened 2026-01-28
Cannot select a fullscreen Minecraft window with window switcher


tokyo4j:
### Problem description

When I tried to switch focus to a fullscreen Minecraft window from another application, the focus does not change for some reason. Strangely, I can switch focus by clicking the `waybar` taskbar item. The minecraft uses Xwayland (according to `[W]` prefix in the window switcher). 

I'll debug this further later.

### Steps to reproduce

1. Start `prismlauncher`
2. Create an minecraft instance (I selected "Mechanical Mastery" modpack with Minecraft 1.18.2)
3. Launch the minecraft instance
4. Make the minecraft window fullscreen
5. Switch focus to another window (e.g. firefox)
6. Try to switch focus back to minecraft window with window switcher
7. The focus is kept on the previous window

### labwc build source

Local build

### labwc version

labwc 0.9.3-24-g37c7de32-dirty (+xwayland +nls +rsvg +libsfdo)

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

tokyo4j:
Maybe the cause is that we momentarily give keyboard focus to the previous window when finishing window switcher:

https://github.com/labwc/labwc/blob/a62441ff77765cf85fe58d8ff9774cedb1d32a75/src/cycle/cycle.c#L207-L208

EDIT: Turned out that this is not the cause of this issue. But I opened #3346 to make debugging easier.

Consolatis:
Maybe @jlindgren90 has any ideas? Somewhat sounds like the usual X11 input model confusion but then I'd assume that a focus request via panel should behave the same.

jlindgren90:
Not sure. Does the same occur with anything else besides Minecraft?

tokyo4j:
Fixing https://github.com/labwc/labwc/issues/3344#issuecomment-3813239322 didn't fix this issue.
I'll continue debugging, currently with `x11trace`.

I also found that the problem can sometimes happen when focusing on the fullscreen minecraft window by clicking the taskbar item in waybar. So this issue is not specific to the window switcher.

I also reproduced the issue in Wayfire, but not in KWin. So maybe this is related to how wlroots handles X11 messages. 

> Not sure. Does the same occur with anything else besides Minecraft?

Maybe we can reproduce it in other programs using `glfw`. Looks like the issue is related to a fullscreen `glfw` window minimizing itself when it's unfocused.

[...]

PR #3346, opened 2026-01-30
cycle: fix spurious focus changes on finishing window switcher

tokyo4j:
This **DOES NOT FIX** #3344, but I think this is an improvement.

As described in the `FIXME` comment in `cycle.c`, we had spurious focus changes where the keyboard focus is momentarily given to the previously focused window when finishing the window switcher, an then it is given to the selected window.

This commit fixes this by adding a parameter in `seat_focus_override_end()` to avoid restoring the focus to the previously focused window.

I also removed the check for `!seat->seat->keyboard_state.focused_surface` in `seat_focus_override_end()`. I thought it was necessary to avoid updating the keyboard focus if the focus was given to a session-lock surface before e.g. finishing window switching, but `seat_focus()` is no-op in that case anyway.

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -397,10 +397,10 @@ void seat_output_layout_changed(struct seat *seat);
 void seat_focus_override_begin(struct seat *seat, enum input_mode input_mode,
 	enum lab_cursors cursor_shape);
 /*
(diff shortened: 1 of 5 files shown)


johanmalm:
LGTM

PR #3347, opened 2026-01-30
Avoid double use of `struct workspace`

tokyo4j:
`struct workspace` was used both for representing an actual workspace and for an entry of workspace configuration. Avoid it for clarity.

### DIFF
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -57,6 +57,11 @@ struct usable_area_override {
 	struct wl_list link; /* struct rcxml.usable_area_overrides */
 };
 
+struct workspace_config {
+	struct wl_list link; /* struct rcxml.workspace_config.workspaces */
+	char *name;
+};
+
 struct rcxml {
 	/* from command line */
 	char *config_dir;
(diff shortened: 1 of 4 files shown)


include/config/rcxml.h
```diff
@@ -57,6 +57,11 @@ struct usable_area_override {
 	struct wl_list link; /* struct rcxml.usable_area_overrides */
 };
 
+struct workspace_entry {
+	struct wl_list link; /* struct rcxml.workspace_config.workspaces */
+	char *name;
+};
```
Consolatis: Suggest something like `workspace_config` rather than the generic `_entry`.

src/workspaces.c
```diff
@@ -590,36 +590,35 @@ workspaces_reconfigure(struct server *server)
...
-		if (strcmp(actual_workspace->name, configured_workspace->name)) {
+		if (strcmp(workspace->name, entry->name)) {
 			/* Workspace is renamed */
 			wlr_log(WLR_DEBUG, "Renaming workspace \"%s\" to \"%s\"",
-				actual_workspace->name, configured_workspace->name);
-			free(actual_workspace->name);
-			actual_workspace->name = xstrdup(configured_workspace->name);
+				workspace->name, entry->name);
+			free(workspace->name);
+			workspace->name = xstrdup(entry->name);
```
Consolatis: Could use `xstrdup_replace()`

Consolatis:
With those two nitpicks addressed this LGTM. Not tested.

Consolatis:
Thanks

tokyo4j:
Thanks for review! I waited for CI to finish and forgot about it :sweat_smile:

PR #3348, opened 2026-01-30
desktop: use for_each_view() instead of scene graph traversal

tokyo4j:
...for `desktop_focus_output()` and `desktop_topmost_focusable_view()`.


johanmalm:
I like the look of that :)

Thanks.

PR #3350, opened 2026-01-31
view: cancel interactive move/resize when minimized/unmapped

jlindgren90:
Currently, with enough dexterity, you can minimize a window via hotkey with one hand while continuing to move it (now invisible) with the mouse in your other hand. The same can probably happen with XWayland windows that self-unmap at an inconvenient time.

Probably not a common occurrence, but it's trivial to handle.


Consolatis:
Nice catch, thanks!

PR #3353, opened 2026-01-31
desktop: use for_each_view() in desktop_topmost_focusable_view()

tokyo4j:
Fixes a regression in #3348 that the bottom-most view is focused when when an exclusive layer surface (e.g. fuzzel) is unfocused.

Also, added some comments to clarify the order.


Consolatis:
Thanks

PR #3355, opened 2026-02-03
view: fix error messages on region or usable area changes

### LINKED ISSUE #3341: `view has no last placement info` error on region or usable area updates
In #3310, I considered `view_adjust_for_layout_change()` is called only from output destroy handler or layout change handler which explicitly sets `view->last_placement`, but it was actually called from other places, like `regions_reconfigure()` and `output_update_all_usable_areas()`, emitting `view has no last placement info` error.

We will need to fix this before the next release.

Reproducer:
1. Open a window
2. Run `labwc -r` which calls `regions_reconfigure()`
3. `view has no last placement info` error message is printed

tokyo4j:
Fixes #3341.

#3310 implemented output-relative position saving/restoring on output un-plugging/re-plugging. It worked as follows:

1. Store the output-relative view geometry in `view->last_placement` (if not set) before adding/removing an output from the layout.
2. After adding/removing an output, call `view_adjust_for_layout_change()` after the layout change to restore the output-relative view geometry based on `view->last_placement`.

However, it didn't consider `view_adjust_for_layout_change()` being called from other places such as `regions_reconfigure()` and `output_update_all_usable_areas()`. In such cases, `view->last_placement` is empty and it emits an error message "view has no last placement info". This can happen when a panel is mapped or unmapped, or on Reconfigure.

This PR fixes it by changing the life cycle of `view->last_placement`. It used to be set only before output layout changes and cleared on user-initiated moves/resizes, but now it is always set and updated on user-initiated moves/resizes. I think this is more intuitive, too.

CC @jlindgren90

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -214,11 +214,10 @@ struct view {
 	 */
 	struct wlr_box natural_geometry;
 	/*
-	 * last_placement represents the last view position set by the user
-	 * before layout changes. output_name and relative_geo are used to
-	 * keep or restore the view position relative to the output and
-	 * layout_geo is used to keep the global position when the output is
-	 * lost.
+	 * last_placement represents the last view position set by the user.
+	 * output_name and relative_geo are used to keep or restore the view
+	 * position relative to the output and layout_geo is used to keep the
+	 * global position when the output is lost.
 	 */
 	struct {
 		char *output_name;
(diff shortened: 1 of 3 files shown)


tokyo4j:
This PR causes heap allocations at `view->last_placement.output_name` every time the view is moved, but I don't think its overhead is serious. According to `perf`, `save_last_placement()` takes 1.3% execution time of `process_cursor_move()`: 

[flamegraph.html](https://github.com/user-attachments/files/25039167/flamegraph.html)

jlindgren90:
I suppose we could avoid the heap allocation by checking if the output name has actually changed, via strcmp?

jlindgren90:
LGTM with the one suggestion (strcmp check). Not tested.

tokyo4j:
Thanks. Used `str_equal()` instead, since `view->last_placement.output_name` can be NULL and I don't want to add an extra `if (!view->last_placement.output_name || ...` check.

Issue #3357, opened 2026-02-03
Segfault at startup with autoEnableOutputs=no


jlindgren90:
### Problem description

With core.autoEnableOutputs=no, labwc (master) recently started crashing at startup:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  create_output_config (server=0x7fffdef22f70) at ../src/output.c:884
884				head->state.x = output->scene_output->x;
[Current thread is 1 (Thread 0x7f1d1c5f5d00 (LWP 11513))]
(gdb) bt
#0  create_output_config (server=0x7fffdef22f70) at ../src/output.c:884
#1  0x000055e5f7f1ed44 in do_output_layout_change (server=0x7fffdef22f70) at ../src/output.c:896
#2  0x000055e5f7f1e30d in handle_new_output (listener=0x7fffdef237a8, data=0x55e61ac99420) at ../src/output.c:595
#3  0x00007f1d207534d0 in wl_signal_emit_mutable () at /usr/lib/libwayland-server.so.0
#4  0x000055e5f7f9a9e0 in new_output_reemit (listener=0x55e61a220678, data=0x55e61ac99420)
    at ../subprojects/wlroots/backend/multi/backend.c:189
#5  0x00007f1d207534d0 in wl_signal_emit_mutable () at /usr/lib/libwayland-server.so.0
#6  0x000055e5f7f8ab32 in scan_drm_connectors (drm=0x55e61a287f60, event=0x0)
    at ../subprojects/wlroots/backend/drm/drm.c:1856
#7  0x000055e5f7f84191 in backend_start (backend=0x55e61a287f60) at ../subprojects/wlroots/backend/drm/backend.c:25
#8  0x000055e5f7f815fb in wlr_backend_start (backend=0x55e61a287f60) at ../subprojects/wlroots/backend/backend.c:57
#9  0x000055e5f7f9a3fb in multi_backend_start (wlr_backend=0x55e61a28fb10)
    at ../subprojects/wlroots/backend/multi/backend.c:31
#10 0x000055e5f7f815fb in wlr_backend_start (backend=0x55e61a28fb10) at ../subprojects/wlroots/backend/backend.c:57
#11 0x000055e5f7f25e41 in server_start (server=0x7fffdef22f70) at ../src/server.c:747
#12 0x000055e5f7f1cc40 in main (argc=1, argv=0x7fffdef23b08) at ../src/main.c:251
(gdb) p output
[...]

PR #3361, opened 2026-02-04
tree-wide: do not try to use outputs with no scene_output

jlindgren90:
- check for valid `scene_output` in `output_is_usable()`
- change many `output != NULL` checks to use `output_is_usable()`
- remove one now-redundant separate check for valid `scene_output`

Fixes a crash at startup (with `autoEnableOutputs=no`) due to dereferencing null `scene_output` in `create_output_config()` since 7d7ece21d98d9e02b8834feb1138bae30a8ed271 ("output: suppress error when output position is unavailable").

Fixes:
- #3357

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -1409,7 +1409,7 @@ run_action(struct view *view, struct server *server, struct action *action,
 			break;
 		}
 		struct output *output = view->output;
-		if (!output) {
+		if (!output_is_usable(output)) {
 			break;
 		}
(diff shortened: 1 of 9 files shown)


johanmalm:
Sorry about merging 7d7ece2 that caused the crash. Thanks for taking the time to investigating and fixing this. I like the look of it a lot. Seems more wholesome and possibly address a whole class of defects.

PR #3362, opened 2026-02-04
xwayland: flush X11 connection after focus/activate

jlindgren90:
This mitigates a race where the XWayland server may generate an unwanted FocusOut event for the newly activated window, if it receives mouse/ pointer events over the parallel wayland connection first.

In particular, this fixes an issue with certain fullscreen applications (such as Minecraft) that self-minimize when receiving FocusOut.

Also limit a previous similar workaround to apply only to XWayland views.

Fixes:
- #3344

See also: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4044


Consolatis:
:)

PR #3364, opened 2026-02-04
Add an XML formatting tool (merely a simply libxml2 wrapper) and adjust files in `docs/`

johanmalm:
This helps when working with `labwc-tweaks` in that it's easier to spot the actual changes without all the formatting noise.

### DIFF
--- a/docs/menu.xml
+++ b/docs/menu.xml
@@ -1,83 +1,82 @@
 <openbox_menu>
-<!-- Note: for localization support of menu items "client-menu" has to be removed here -->
-<menu id="client-menu">
-  <item label="Minimize">
-    <action name="Iconify" />
-  </item>
-  <item label="Maximize">
-    <action name="ToggleMaximize" />
-  </item>
-  <item label="Fullscreen">
-    <action name="ToggleFullscreen" />
-  </item>
-  <item label="Roll Up/Down">
-    <action name="ToggleShade" />
-  </item>
-  <item label="Decorations">
-    <action name="ToggleDecorations" />
-  </item>
-  <item label="Always on Top">
-    <action name="ToggleAlwaysOnTop" />
-  </item>
-  <!--
-    Any menu with the id "workspaces" will be hidden
-    if there is only a single workspace available.
-  -->
-  <menu id="client-send-to-menu"/>
-  <!--
-    openbox default workspace selector
-    to use replace above workspace menu with the example below
-    the label is required, but you can change the text.
+  <!-- Note: for localization support of menu items "client-menu" has to be removed here -->
+  <menu id="client-menu">
+    <item label="Minimize">
+      <action name="Iconify"/>
+    </item>
+    <item label="Maximize">
+      <action name="ToggleMaximize"/>
+    </item>
+    <item label="Fullscreen">
+      <action name="ToggleFullscreen"/>
+    </item>
+    <item label="Roll Up/Down">
+      <action name="ToggleShade"/>
+    </item>
+    <item label="Decorations">
+      <action name="ToggleDecorations"/>
+    </item>
+    <item label="Always on Top">
+      <action name="ToggleAlwaysOnTop"/>
+    </item>
+    <!--
+      Any menu with the id "workspaces" will be hidden
+      if there is only a single workspace available.
+    -->
+    <menu id="client-send-to-menu"/>
+    <!--
+      openbox default workspace selector
+      to use replace above workspace menu with the example below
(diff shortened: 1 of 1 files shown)


docs/rc.xml
```diff
@@ -2,38 +2,33 @@
...
+    <font name="sans" size="10"/>
   </theme>
-
   <keyboard>
-    <default />
+    <default/>
     <!-- Use a different terminal emulator -->
     <keybind key="W-Return">
-      <action name="Execute" command="foot" />
+      <action name="Execute" command="foot"/>
```
Consolatis, line 15: Personally I think the space before `/` in self-closing arguments is much more readable.

Consolatis:
I don't quite understand what the benefit of this is. You wrote "makes spotting changes easier" but how exactly does this change achieve that?

johanmalm:
> I don't quite understand what the benefit of this is. You wrote "makes spotting changes easier" but how exactly does this change achieve that?

`labwc-tweaks` uses `xmlReadFile()` and `xmlSaveFormatFile()`. So when clicking `Apply`, it is not just the changed node that is updated, the whole document is re-written (like a clang-format for XML :smile:). So, if example files like `rc.xml.all` are already formatted correctly - from the perspective of libxml2 - then it is easier to see the changes.

Personally, when I'm testing `labwc-tweaks` I sometimes use `rc.xml.all` as the master and do a `diff -u` on the changes.

Consolatis:
I have a feeling that this could then be a part of labwc-tweaks instead. Basically everything that would be required is to save without changing anything on startup (or once manually). IMHO the `docs/` directory is primarily targeted at humans and serves as a reference and description of available config options and not as input for external applications.

scripts/helper/format-xml.c
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * Helper to format XML the way libxml2 does it - because it helps stot
```
domo141, line 3: s/stot/spot/ ?

johanmalm: Thanks. Will discuss with @Consolatis to conclude whether or not we want this PR before fixing.

johanmalm:
> I have a feeling that this could then be a part of labwc-tweaks instead. Basically everything that would be required is to save without changing anything on startup (or once manually).

Yes. Can remove `format-xml.c` - it just seemed more convenient to apply the formatting directly with a few lines of a CLI tool as currently `labwc-tweaks` cannot be setup to read `rc.xml.all`.

> IMHO the `docs/` directory is primarily targeted at humans and serves as a reference and description of available config options and not as input for external applications.

With the removal of some empty lines and spaces it can do both. I don't feel that the proposed changes makes any difference to the readability of humans - if anything the increase in consistency probably helps brains which like patterns.

But honestly, this is not a big thing for me, but merely a matter of convenience - and possibly helpful for users.

Having played around a bit with `lxqt-wayland-session`, `xfce4-session-wayland-experimental` (on OpenSUSE) and others, it appears to me that it's not unusual for the xml files in `docs/` to serve as a starting point for projects/devs/users. I was of a mind to submit a couple of patches to the mentioned projects for potential improvements when it occurred to me that they were not consistent with the libxml2 formatting. So my choice became to either hand-amend (which is a pain now that I have a working version of `labwc-tweaks` - or first submit a format-style patch and then use `labwc-tweaks`. Which made me think it's a bit rich to do the latter without "fixing" the formatting of our own upstream `docs/`. Anyways.... this is of course just a personal, subjective opinion, but thought it would make sense to build a bit more context around proposed change.

Consolatis:
I am not against having some helper util around which formats our xml example files and ensures a sane and consistent syntax. But then I think we should define what that syntax should look like with the main goal being easy for users to understand and skim over rather than ensuring `diff -u` with other utilities updating the XML produces no whitespace changes.

It is just my personal opinion but I think especially the removal of  the space before the `/>` in self-closing elements makes reading and skimming through the xml harder. I don't really care about removing some empty lines in terms of readability.

johanmalm:
I think I'm right in saying that `libxml2` does not provide an option to control the whitespace before the / in empty element tags. So I guess the discussion boils down to (a) having aesthetically pleasing spaces before slashes or (b) reduced probability of `labwc-tweaks` modifying xml files for no other reason than that space.

johanmalm:
I implemented this in `labwc-tweaks` instead.

Ref:
- https://github.com/labwc/labwc-tweaks/pull/238

PR #3365, opened 2026-02-05
theme: add option window.button.hover.bg.color

johanmalm:


domo141:
great progress - this obsoletes my hack:
```
-       float overlay_color[4] = { 0.15f, 0.15f, 0.15f, 0.3f};
+       float overlay_color[4] = { 0.0f, 0.0f, 0.0f, 0.6f};
```
I'll test in near future...

domo141:
... and works for me, with

window.button.hover.bg.color: `#00000090`

the effect looks near { 0.0f, 0.0f, 0.0f, 0.6f};
(trial & error is faster than trying to understand premultiplied alpha :D)

One thing that come to my mind while looking the change in labwc-theme.5.scd,
In case with the icon (buttons) that open window menu, there is no "hover effect"
for me (I don't use themes, I just have themerc.override) -- was not before and
now now. Looking the change -- which if effectively the same as my hardcoded
change, that was expected by me ( just that one could have assumed that based
when reading text like:
 "Note: The *window.button.hover* namespace (below) relates to button hover
 +overlay effects which work with all buttons including svg, png and xbm."

johanmalm:
@domo141 Thanks for testing.

Regarding not showing hover effects for the `icon` button, that is how it's designed. IIRC that's partly because some themes specify special hover variants of their buttons, but the `icon` button can't do that so would then look inconsistent. Also, that icon tends to be bigger, so the hover effect might look weird.

johanmalm:
@domo141 Thanks for testing.

Regarding not showing hover effects for the `icon` button, that is how it's designed. IIRC that's partly because some themes specify special hover variants of their buttons, but the `icon` button can't do that so would then look inconsistent. Also, that icon tends to be bigger, so the hover effect might look weird.

docs/labwc-theme.5.scd
```diff
@@ -173,6 +173,18 @@ window.*.title.bg.colorTo.splitTo: #557485
 	Space between titlebar buttons, in pixels.
 	Default is 0.
 
+Note: The *window.button.hover* namespace (below) relates to button hover
+overlay effects which work with all buttons including svg, png and xbm.
```
tokyo4j, line 177: I think we should say "**default** button hover overlay effects" and mention that `window.button.hover.*` is overridden by the icon files like `close_hover-active.svg`.

tokyo4j:
Thanks! LGTM other than my comment above.

docs/labwc-theme.5.scd
johanmalm: Thanks. Have added.

Issue #3371, opened 2026-02-10
Weird mouse-move and keyboard-unmaximize interaction (corner case)


jlindgren90:
### Problem description

While just messing around and trying to break things as usual, I noticed a corner case where the interaction between interactive move (with the mouse) and simultaneous keybindings gets a bit weird. Also it's a bit tricky to fix - I'll create a draft PR shortly, but thought I'd open an issue just to describe the scenario first.

### Steps to reproduce

1. Maximize a smallish window (doesn't matter how)
2. Press down the mouse button on the titlebar, but don't move the mouse (yet)
3. Un-maximize the window via keybinding
4. Move the mouse while still holding the button down

Now moving the mouse will move the window around, but it's at a significant offset - the window starts moving from its un-maximized position, while the mouse cursor is still up near the top of the screen.

### labwc build source

Release

### labwc version

0.9.3 + some patches, nothing that would be related I don't think

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

johanmalm:
Fixed by #3372

PR #3372, opened 2026-02-10
interactive: set grab parameters at cursor press

jlindgren90:
Add `interactive_set_grab_context()` which is called when the mouse button is first pressed, before `interactive_begin()`. This fixes two small issues:

- The cursor origin position for interactive move/resize was slightly off (depending on mouse resolution), because it was set after the mouse had already moved slightly. Now it's exact.

- If app- or keybind-initiated maximize (etc.) happened after the button press but before the mouse was moved, then `interactive_begin()` would still start move/resize even though the view might now be far away from the cursor. Now `interactive_cancel()` works as expected, even if called before interactive_begin().

Also, make sure to call `interactive_cancel()` for un-maximize as well.

Fixes:
- #3371

Creating as draft for now since the code is a bit tricky, there might be some case I missed. Interactive move/resize via touchscreen or tablet seems like it would not work, but likely didn't before either (don't have HW to test it on currently).

### DIFF
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -414,6 +414,7 @@ void seat_focus_override_end(struct seat *seat, bool restore_focus);
  */
 void interactive_anchor_to_cursor(struct server *server, struct wlr_box *geo);
 
+void interactive_set_grab_context(struct cursor_context *ctx);
 void interactive_begin(struct view *view, enum input_mode mode,
 	enum lab_edge edges);
 void interactive_finish(struct view *view);
--- a/src/action.c
+++ b/src/action.c
@@ -1262,6 +1262,14 @@ run_action(struct view *view, struct server *server, struct action *action,
 		break;
 	case ACTION_TYPE_MOVE:
 		if (view) {
+			/*
+			 * If triggered by mousebind, grab context was already
+			 * set by button press handling. For keybind-triggered
+			 * Move, set it now from current cursor position.
+			 */
+			if (view != server->seat.pressed.ctx.view) {
+				interactive_set_grab_context(ctx);
+			}
 			interactive_begin(view, LAB_INPUT_STATE_MOVE,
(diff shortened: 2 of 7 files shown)


johanmalm:
Have tested; seems to work well.

Yes, it's not the easiest of patches to review. Are you of a mind to get this reviewed/merged before or after `0.9.4` for which we anticipate cool-down starting on Friday?

jlindgren90:
I haven't seen/thought of any issues with it after sleeping on it, seems like it should be OK for 0.9.4. I'll mark it ready for review. Thanks!

johanmalm:
LGTM.

It takes quite a few fingers to test this! Tried `SnapToEdge` too.

Just a thought - possibly for a following PR. Should we consider putting `grabbed_view`, `grab_x`, `grab_y` and `grab_box` into something like a `struct grab` or `struct cursor_interactive_grab`. Possibly also `resize_edges`.

jlindgren90:
> Should we consider putting grabbed_view, grab_x, grab_y and grab_box into something like a struct grab or struct cursor_interactive_grab. Possibly also resize_edges.

That would be a nice cleanup. Full disclosure, I experimented with factoring out a `view_grab` "class" in Rust:
https://github.com/jlindgren90/labwc/blob/rust/src/view_grab.rs

(Very much only an experiment, and mainly because I wanted to learn the language.)

johanmalm:
> Full disclosure, I experimented with factoring out a `view_grab` "class" in Rust: https://github.com/jlindgren90/labwc/blob/rust/src/view_grab.rs
> 
> (Very much only an experiment, and mainly because I wanted to learn the language.)

Wow. Every day is full of surprises. I would very much like to discuss your insight and how you are finding it. Probably best done in a dedicated discussion thread though.

Issue #3167, opened 2025-10-24
Setting keyboard interactivity to on-demand steals focus


spl237:
### Problem description

It appears that either the action of setting the keyboard interactivity of a window to on-demand, or opening a window with on-demand set, causes that window to immediately take keyboard focus. As I understand it, an on-demand window should only take focus when keyboard events occur, not by default.

### Steps to reproduce

1. On Raspberry Pi OS, open a terminal window.
2. Type into the window to establish that it has keyboard focus.
3. In the terminal, do "pkill pcmanfm" to kill the pcmanfm process which is drawing the desktop window.
4. When the desktop auto-restarts, type - the text now goes into the desktop search box rather than the terminal.

If I change the setting of keyboard interactivity in the desktop from on-demand to none, this does not happen, and the terminal retains keyboard focus when the desktop restarts.

### labwc build source

Local build

### labwc version

0.9.2

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Raspberry Pi OS

johanmalm:
> It appears that either the action of setting the keyboard interactivity of a window to on-demand, or opening a window with on-demand set, causes that window to immediately take keyboard focus.

See analysis below. In terms of terminology, I (we) use window to mean a "toplevel" - not a layer-shell client.

> As I understand it, an on-demand window should only take focus when keyboard events occur, not by default.

According to the protocol, the on-demand "requests the compositor to allow this surface to be focused and unfocused by the user in an implementation-defined manner."

It also says: "Typically, the compositor will want to use its normal mechanism to manage keyboard focus between layer shell surfaces with this setting and regular toplevels on the desktop layer (e.g. click to focus)."

For windows (toplevels), we do 'focus-on-map', 'click-to-focus' as well as a few more like when using the window-switcher or closing another window.

In `src/layers.c` we call `layers_try_set_focus()` in `handle_map()`. I guess we could consider not doing that for on-demand layer-shell surfaces, so that these only ever get keyboard focus 'on click' (or indeed when a popup like the applications menu is opened).

I completely see your use-case here. In simple words, when on-demand layer-shell desktop components start (map) we don't want them to get keyboard-focus. It probably doesn't matter on initial compositor startup, but when these components are (re-)started later it's potentially quite annoying.

Ref:

- https://github.com/labwc/labwc/blob/df8e3d819cf32ea46a6ff14d8e61135b6900b5e7/src/layers.c#L403

Not tested, but could something like this work?

```
[...]

PR #3373, opened 2026-02-10
layers: restrict layer-shell on-demand focus on map

johanmalm:
...so that clients with on-demand keyboard interactivity in the background/bottom/top layers do not steal focus. The reason for this is to avoid desktop components like panels and desktops from stealing keyboard focus if they are re-started.

We still let such clients in the overlay layer get focus on map so that for example labnag with its default `<prompt>` setting take keyboard focus on map, whilst still letting users click on something else to move focus away.

Fixes: #3167

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -382,12 +382,19 @@ handle_unmap(struct wl_listener *listener, void *data)
 	layer->being_unmapped = false;
 }
 
+static bool
+is_overlay(struct wlr_layer_surface_v1 *layer_surface)
+{
+	return layer_surface->current.layer == ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY;
+}
+
 static void
 handle_map(struct wl_listener *listener, void *data)
 {
 	struct lab_layer_surface *layer = wl_container_of(listener, layer, map);
-	struct wlr_output *wlr_output =
-		layer->scene_layer_surface->layer_surface->output;
+	struct wlr_layer_surface_v1 *layer_surface =
+		layer->scene_layer_surface->layer_surface;
+	struct wlr_output *wlr_output = layer_surface->output;
 	if (wlr_output) {
 		output_update_usable_area(wlr_output->data);
 	}
@@ -398,6 +405,23 @@ handle_map(struct wl_listener *listener, void *data)
 	 * automatically based on the position of the surface and outputs in
 	 * the scene. See wlr_scene_surface_create() documentation.
 	 */
+
+	/*
+	 * Layer-shell clients with exclusive interactivity always get focus,
+	 * whereas on-demand ones only get it when in the overlay layer. We
+	 * could make this configurable, but for the time being this default
+	 * behaviour strikes a balance between:
+	 *
+	 *   1. Giving overlay, on-demand clients (like labnag with the default
+	 *      <prompt> settings) keyboard focus.
+	 *   2. Preventing desktop components like panels and desktops from
+	 *      stealing keyboard focus on re-start. On compositor start, this
+	 *      is not really a problem, but if any such client restarts later,
+	 *      focus-stealing is unlikely to be desirable.
+	 */
+	if (!is_overlay(layer_surface) && is_on_demand(layer_surface)) {
+		return;
+	}
 
 	struct seat *seat = &layer->server->seat;
 	layer_try_set_focus(seat, layer->scene_layer_surface->layer_surface);


jlindgren90:
> ...so that clients with on-demand keyboard interactivity in the background/bottom/top layers do not steal focus.

Just curious, what is the reason for restricting top layer from taking focus? Background and bottom makes sense to me -- we don't want a layer taking focus if it's under the active view.

(I don't have much experience on the layer-client side except for qmpanel, which uses top-layer but isn't on-demand -- it switches back and forth between "none" and "exclusive" depending on whether a menu is popped up or not.)

johanmalm:
> Just curious, what is the reason for restricting top layer from taking focus? Background and bottom makes sense to me -- we don't want a layer taking focus if it's under the active view.

It think it's common for panels to be in either top or bottom, so I guess the same rationale applies regarding them stealing focus on re-start.

I don't think there is a hard and fast rule about where dialogs like `labnag`, `fuzzel` and `rofi` go, but I in my experience `overlay` is common. Welcome perspectives from others here.

So in summary, it's based on an assumption of mine that panels reside in `bottom` or `top`, whereas dialogs - for which we want to make the exception - go in `overlay`.

Interestingly, the protocol treats `overlay` and `top` separately for `exclusive` interactivity.

Ref:
- https://github.com/labwc/labwc/blob/a672e8a9fdd486b1a2c645aa1135a610124438dc/src/output.c#L547-L559
- https://github.com/labwc/labwc/blob/a672e8a9fdd486b1a2c645aa1135a610124438dc/src/server.c#L563-L584
- https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity
- https://codeberg.org/dnkl/fuzzel/src/commit/18bda5015696b5784384d74fac172adf0f0b976c/config.c#L1725

stefonarch:
Is this tested with lxqt-panel's fancy-menu search edit focus when opened by shortcut (`lxqt-qdbus openmenu`)?

johanmalm:
> Is this tested with lxqt-panel's fancy-menu search edit focus when opened by shortcut (`lxqt-qdbus openmenu`)?

No, but I think that's a different scenario though isn't it? Does `lxqt-qdbus openmenu` currently work? Can do some testing tomorrow if you think different.

Ref: 
- https://github.com/labwc/labwc/pull/3165

jlindgren90:
> So in summary, it's based on an assumption of mine that panels reside in bottom or top, whereas dialogs - for which we want to make the exception - go in overlay.

Makes sense. Personally, I came to the conclusion that on-demand is not suitable for panels. At least it didn't work well for qmpanel. Not only because the panel doesn't want focus on startup/map, but also because it wants to give it away again to the topmost view once the last menu is closed. On-demand doesn't work for that, making the none <-> exclusive dance necessary.

The layer-shell protocol is pretty loose about what "on-demand" actually means, but I read it as a gentle suggestion to handle on-demand layer focus more or less like for normal toplevel views. Just my two cents -- not a strong opinion either way.

> Typically, the compositor will want to use its normal mechanism to manage keyboard focus between layer shell surfaces with this setting and regular toplevels on the desktop layer (e.g. click to focus). Nevertheless, it is possible for a compositor to require a special interaction to focus or unfocus layer shell surfaces[...]

tokyo4j:
LGTM. Given that `labnag` on the default overlay layer is automatically focused, I'm fine with this.

tsujan:
> Does lxqt-qdbus openmenu currently work? 

That's an old story, not a new issue: the menu is opened but doesn't get focused on showing if the panel isn't already focused. That points to another issue: if we made the panel non-focusable (`LayerShellQt::Window::KeyboardInteractivityNone`) under labwc, its child popups would never get the focus. Only kwin_wayland, wayfire and hyprland support a non-focusable panel (which is the correct behavior of a panel in general). Under them, `lxqt-qdbus openmenu` works as expected.

johanmalm:
> > Does lxqt-qdbus openmenu currently work?
> 
> That's an old story, not a new issue: the menu is opened but doesn't get focused on showing if the panel isn't already focused. That points to another issue: if we made the panel non-focusable (`LayerShellQt::Window::KeyboardInteractivityNone`) under labwc, its child popups would never get the focus. Only kwin_wayland, wayfire and hyprland support a non-focusable panel (which is the correct behavior of a panel in general). Under them, `lxqt-qdbus openmenu` works as expected.

That's what I thought - hence why I didn't understand the question by @stefonarch

So we've got to fix #3165, right?

tsujan:
> So we've got to fix https://github.com/labwc/labwc/pull/3165, right?

Yes, it seems so.

stefonarch:
Bad memory here, sorry, I remembered wrongly that it did get focus.

PR #3377, opened 2026-02-13
 workspaces: keep unfocusable omnipresent views visible

tokyo4j:
Fixes #3376.

When switching between workspaces, we need to migrate omnipresent views to the new workspace so that they are kept visible. However, by using `for_each_view_reverse()` macro, we were not migrating omnipresent views that `view_is_focusable()` return false (e.g. Conky).

Use `wl_list_for_each_reverse()` macro instead to catch all the omnipresent views.


Consolatis:
Thanks :)

Issue #593, opened 2022-10-16
Render rounded corners via autoscaling buffer


Consolatis:
Otherwise blurry galore happens:

![image](https://user-images.githubusercontent.com/35009135/196048629-196135fa-c40d-4c99-b9dc-fcf10b7ab241.png)

Consolatis:
I looked into this a while ago and for this to be implemented we have to modify the `scaled_scene_buffer` API a bit.
As of right now the buffer implementation will automatically free the backing `wlr_buffer` when the corresponding scene node is destroyed which is something we really don't want for the SSD images as we keep reusing the backing `wlr_buffer` for all future SSDs.

Consolatis:
@tokyo4j just to confirm, we are now doing that with the scaled_img_buffer, correct? If yes this issue can be closed.

tokyo4j:
No. Icons are rendered with `scaled_img_buffer`, but titlebar corners placed below them can still be blurry.

I actually implemented `scaled_corner_buffer` for that, but I found its border width becomes consistent with other straight borders when fractional scales are used, in a the same manner as #2394. I'll open a draft PR for further discussion later.

lszl84:
Any update on this one? The README proudly mentions HiDPI support, but blurry window decorations really spoil the experience...

tokyo4j:
> Any update on this one? The README proudly mentions HiDPI support, but blurry window decorations really spoil the experience...

Sorry for your inconvenience. I think we can fix this problem with the approach proposed here (#2695), which draws the titlebar as a single auto-scaling buffer. We will definitely need to fix this problem.

PR #3382, opened 2026-02-16
[proof of concept] textured borders

jlindgren90:
I played around with implementing a different visual style for borders, with some slight 3D-ish shading and the active view highlighted in color all around (not just the titlebar). In a decade of "flat" UI design, I suppose it probably looks dated, but I still prefer this style or something similar.

The code is only a quick hack, and doesn't play well with other theme options (particularly rounded corners). So this is just to start a discussion and see if anyone else has interest, or has ideas how to code something similar in a better/more flexible way.

Screenshot 2026-02-16 07:18:16

Theme in the screenshot is https://github.com/jlindgren90/old-style-theme.

### DIFF
--- a/include/ssd-internal.h
+++ b/include/ssd-internal.h
@@ -114,7 +114,7 @@ struct ssd {
 		struct wlr_scene_tree *tree;
 		struct ssd_border_subtree {
 			struct wlr_scene_tree *tree;
-			struct wlr_scene_rect *top, *bottom, *left, *right;
+			struct wlr_scene_buffer *top, *bottom, *left, *right;
 		} subtrees[2]; /* indexed by enum ssd_active_state */
 	} border;
 
--- a/include/theme.h
+++ b/include/theme.h
@@ -14,6 +14,11 @@
 
 struct lab_img;
 
+/* XXX: using fixed border widths for now */
+/* (top is narrower to blend with titlebar) */
+#define BORDER_PX_TOP 2
+#define BORDER_PX_SIDE 4
+
 /*
  * Openbox defines 7 types of Gradient background in addition to Solid.
  * Currently, labwc supports only Vertical and SplitVertical.
--- a/src/ssd/ssd-border.c
+++ b/src/ssd/ssd-border.c
@@ -2,7 +2,9 @@
 
 #include <assert.h>
 #include <wlr/types/wlr_scene.h>
+#include "buffer.h"
 #include "common/macros.h"
+#include "common/mem.h"
 #include "labwc.h"
 #include "ssd.h"
 #include "ssd-internal.h"
@@ -17,13 +19,19 @@ ssd_border_create(struct ssd *ssd)
 
 	struct view *view = ssd->view;
 	struct theme *theme = view->server->theme;
+#if 0
+	/* XXX: not using with textured borders */
 	int width = view->current.width;
 	int height = view_effective_height(view, /* use_pending */ false);
 	int full_width = width + 2 * theme->border_width;
 	int corner_width = ssd_get_corner_width();
+#endif
 
 	ssd->border.tree = wlr_scene_tree_create(ssd->tree);
+#if 0
+	/* XXX: eliminated this offset */
 	wlr_scene_node_set_position(&ssd->border.tree->node, -theme->border_width, 0);
+#endif
 
 	enum ssd_active_state active;
 	FOR_EACH_ACTIVE_STATE(active) {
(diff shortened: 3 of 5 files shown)


johanmalm:
Hey, nice idea.

Initial thoughts/questions:

1. Does openbox support this?
2. I have been looking at implementint "texture" borders for buttons and menu headers. Were you thinking of following that sort of syntax?
3. I think I am right in saying that Xfce themes allow pixmaps for borders. Might be worth considering that in early design decisions.

jlindgren90:
Openbox has a different scheme where it puts a wider, colored border only on the bottom edge. I'm not really a fan of it. Some xfwm4 themes had a look more similar to this though.

A themeable solution could be to just bite the bullet and implement xfwm4 theme support, rather than inventing something new. Maybe that would be getting outside labwc's niche a bit though.

johanmalm:
> A themeable solution could be to just bite the bullet and implement xfwm4 theme support, rather than inventing something new. Maybe that would be getting outside labwc's niche a bit though.

I have pondered this before. Have never got deep into xfce4 theme spec though.

Consolatis:
Tried to generalize the idea a bit: https://github.com/labwc/labwc/compare/master...Consolatis:labwc:feature/textured_corners

It accepts a 1 x N pixel pattern and derives the corners automatically + uses wlr_scene transforms to rotate and scale them accordingly. Rather than the square corner derivation we could maybe implement a rounded one as well.

It also decouples the lab_data_buffers from the scene graph itself thus one can create multiple scene trees based on the same textured_rect, each tree having its own geometry.

Idea would be to generate the pixel pattern on theme (re)load and call `textured_rect_create_from_pixels()` once and store it in `rc.theme`. All further uses (for SSD instances / menus / OSD and whatnot) would then only use `textured_rect_create_tree()` based on the initially generated pattern. For the SSD case we may need a separate struct to store the individual node_descriptors though.

tokyo4j:
> Tried to generalize the idea a bit: [master...Consolatis:labwc:feature/textured_corners](https://github.com/labwc/labwc/compare/master...Consolatis:labwc:feature/textured_corners)
> 
> It accepts a 1 x N pixel pattern and derives the corners automatically + uses wlr_scene transforms to rotate and scale them accordingly. Rather than the square corner derivation we could maybe implement a rounded one as well.
> 
> It also decouples the lab_data_buffers from the scene graph itself thus one can create multiple scene trees based on the same textured_rect, each tree having its own geometry.
> 
> Idea would be to generate the pixel pattern on theme (re)load and call `textured_rect_create_from_pixels()` once and store it in `rc.theme`. All further uses (for SSD instances / menus / OSD and whatnot) would then only use `textured_rect_create_tree()` based on the initially generated pattern. For the SSD case we may need a separate struct to store the individual node_descriptors though.

I haven't given it much thoughts into it, but I have one question; how are you going to handle different output scales while keeping rounded corners? As far as I can come up with, adding another `scaled-*-buffer.c` is the only solution to address #593.

jlindgren90:
The style here is intentionally pixelated, and for that I think it's fine to let scene-graph handle the scaling.

Also, in that pixelated vein, the corners here are actually neither rectangular nor round, but instead have a 1px square notch out of them.

All of the above being fine, I think, for this style--but it may not generalize well.

Consolatis:
> I haven't given it much thoughts into it, but I have one question; how are you going to handle different output scales while keeping rounded corners? As far as I can come up with, adding another `scaled-*-buffer.c` is the only solution to address

That is a good point. If we can derive the rounded corners based on the initial pattern I think it should be relatively easy to react to new scale events on the scene buffer corners and then re-render into a new rounded corner buffer based on the initial pixel pattern but with a larger corner buffer (e.g. `(pixel_width * scale) ** 2` rather than `pixel_width**2`). It could then be cached in the pattern itself. That would require an actual struct for the scene nodes though which keeps a reference to the pattern and event listeners for the corner buffers. But that would actually be required for resizing support and SSD attachment of node-descriptors as well.

Note that the scaling is really only an issue for the rounded corners, everything else should scale nicely with `WLR_SCALE_FILTER_NEAREST`.

johanmalm:
> Tried to generalize the idea a bit: [master...Consolatis:labwc:feature/textured_corners](https://github.com/labwc/labwc/compare/master...Consolatis:labwc:feature/textured_corners)

I look the look of that. Am I right in think that it would help separate out some of the mixed concerns in #3463 too?

Issue #3381, opened 2026-02-16
Focus regression with git labwc and lxqt-runner


tsujan:
I upgraded to git labwc https://github.com/labwc/labwc/commit/654da8c8a99ef472cccec23cab7f992559e90dd0 and saw that lxqt-runner didn't get the keyboard focus on showing up (with its shortcut) anymore.

This problem doesn't exist in labwc 0.9.3.

EDIT: lxqt-runner is a `QDialog` on the top layer (`LayerShellQt::Window::Layer::LayerTop`).

tokyo4j:
Since #3373, we don't give keyboard focus to layer-shell clients mapped at `top`/`bottom`/`background` layer with `on_demand` keyboard interactivity. So if `lxqt-runner` is one of those clients, it no longer gets keyboard focus.

I'm not sure if we should revert #3373... Of course we can keep #3373 and advise lxqt devs to make `lxqt-runner` mapped at `overlay` layer or use `exclusive` keyboard interactivity, but on the other hand, I think this issue is more annoying for users than #3167. @johanmalm @Consolatis any thoughts?

tsujan:
> and advise lxqt devs to make lxqt-runner mapped at overlay layer or use exclusive keyboard interactivity,

Sorry, but that's not a good advice ;) Adding codes for circumventing issues in this or that compositor isn't an option, let alone *changing* codes. Should I emphasize that it works in wayfire, niri, kwin_wayland, … as well as labwc ≤ 0.9.3?

tokyo4j:
> Sorry, but that's not a good advice ;) Adding codes for circumventing issues in this or that compositor isn't an option, let alone changing codes. Should I emphasize that it works in wayfire, niri, kwin_wayland, … as well as labwc ≤ 0.9.3?

Yeah, I undestand your pain... I had similar experiences with IME support, whether different versions of `fcitx5` and `labwc` worked differently. I'm currently leaning towards reverting #3167 and then start working to fix #3167 by updating layer-shell protocol, without compositor-side workarounds. But let's wait for reactions from other devs anyway.

Consolatis:
Initial thoughts: keep giving focus on map for ON_DEMAND to top layer and above.

johanmalm:
> Initial thoughts: keep giving focus on map for ON_DEMAND to top layer and above.

[...]

PR #3383, opened 2026-02-16
layer: let top-layer on-demand clients grab focus

johanmalm:
Soften the new rule introduced by 155c153 because it broke lxqt-runner which defaults to the top-layer, and we do not want to knowingly break user-space without a very good reason which we do not have for this.

Fixes: #3381

@tsujan Thanks for catching this before the next release. We really don't like breaking user-space and stuff people rely on. I suspect there will be other layer-shell on-demand clients using the top-layer and which expect to be able to grab the keyboard on map. I think it was a poor judgement on my part to change that.

@tokyo4j - Are you happy that we end up in a good space with this PR?

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -383,9 +383,9 @@ handle_unmap(struct wl_listener *listener, void *data)
 }
 
 static bool
-is_overlay(struct wlr_layer_surface_v1 *layer_surface)
+is_above_toplevels(struct wlr_layer_surface_v1 *layer_surface)
 {
-	return layer_surface->current.layer == ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY;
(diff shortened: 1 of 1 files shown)


Consolatis:
LGTM

tokyo4j:
OK, let's go for this.

However, I believe workarounds like this is just a hack, as panels in popular DEs including KDE Plasma and Windows 11 are placed above windows as if they are mapped at `top` layers. I agree that this PR is the best approach at the moment, but I also want to emphasize that we seriously need to work on fixing this at protocol level by e.g. adding another keyboard interactivity to `wlr-layer-shell-v1` or work-in-progress [`ext-layer-shell-v1`](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/28) protocol.

PR #3387, opened 2026-02-21
Refactor initial view positioning/placement

jlindgren90:
For after 0.9.4.

The main motivation here is to unify the separate "map" and "late" positioning paths for xdg-shell views. Some refactoring had to be done first to allow calling into the policy-based placement logic from the commit handler, where we want to update `view->pending` directly instead of calling `view_move`.

From a user standpoint, the benefit is that this allows policy-based placement to apply when an initially-maximized/fullscreen view is restored to floating geometry -- for both xdg-shell and xwayland views.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -514,10 +514,17 @@ int view_effective_height(struct view *view, bool use_pending);
  */
 void view_center(struct view *view, const struct wlr_box *ref);
 
-/*
- * Like view_place_by_policy() but doesn't actually move the view.
- * Returns false if position could not be computed (for example, if no
- * outputs are connected). In that case, @geom is not modified.
+/**
+ * view_compute_position_by_policy() - compute view placement
+ * @view: view to be placed
+ * @geom: floating view geometry to update (in/out)
+ * @allow_cursor: set to false to ignore center-on-cursor policy
+ * @policy: placement policy to apply
+ *
+ * Computes floating view placement according to configured strategy.
+ * Unlike view_place_by_policy(), this function doesn't actually move
+ * the view. It returns false if position could not be computed (e.g.
+ * if no outputs are connected). In that case, @geom is not modified.
  */
 bool view_compute_position_by_policy(struct view *view, struct wlr_box *geom,
 	bool allow_cursor, enum lab_placement_policy policy);
@@ -527,6 +534,10 @@ bool view_compute_position_by_policy(struct view *view, struct wlr_box *geom,
  * @view: view to be placed
  * @allow_cursor: set to false to ignore center-on-cursor policy
  * @policy: placement policy to apply
+ *
+ * Places a floating view according to configured placement strategy.
+ * Clears any maximized/fullscreen/tiled state and restores natural
+ * geometry of the view before positioning.
  */
 void view_place_by_policy(struct view *view, bool allow_cursor,
 	enum lab_placement_policy policy);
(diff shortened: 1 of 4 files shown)


include/view.h
```diff
@@ -515,6 +514,10 @@ int view_effective_height(struct view *view, bool use_pending);
  */
 void view_center(struct view *view, const struct wlr_box *ref);
 
+/* Like view_place_by_policy() but doesn't actually move the view */
+bool view_compute_position_by_policy(struct view *view, struct wlr_box *geom,
+	bool allow_cursor, enum lab_placement_policy policy);
```
tokyo4j: Just a quick initial review: I think we should clarify when this function returns true or false.

jlindgren90:
Force-pushed with some fixes. In particular, to clean up the logic for setting the initial output, which was a bit messy. Doing it at the initial commit seems to be reliable for all the cases I tested (`foot --maximized` and `--fullscreen`, Thunar initially maximized, xfce4-terminal "About" dialog).

include/view.h
```diff
@@ -515,6 +514,14 @@ int view_effective_height(struct view *view, bool use_pending);
...
+/*
+ * Like view_place_by_policy() but doesn't actually move the view.
+ * Returns false if position could not be computed (for example, if no
+ * outputs are connected). In that case, @geom is not modified.
+ */
+bool view_compute_position_by_policy(struct view *view, struct wlr_box *geom,
+	bool allow_cursor, enum lab_placement_policy policy);
+
 /**
  * view_place_by_policy - apply placement strategy to view
```
tokyo4j, line 533: "apply placement strategy" sounds vague. Let's change this to something like: "Un-tile the view and reposition it based on the given policy"

src/view.c
```diff
@@ -657,17 +639,26 @@ view_move_to_cursor(struct view *view)
...
+	struct border margin = ssd_thickness(view);
+	struct seat *seat = &view->server->seat;
+
+	int total_width = geom->width + margin.left + margin.right;
+	int total_height = geom->height + margin.top + margin.bottom;
+
+	int x = (int)seat->cursor->x - (total_width / 2);
+	int y = (int)seat->cursor->y - (total_height / 2);
+
+	/* Order of MIN/MAX is significant here */
```
tokyo4j: Do you mean we let a large view overflow at its right/bottom edges than at its top/left edges? Then I think we should describe it.

src/xdg.c
```diff
@@ -234,15 +262,15 @@ handle_commit(struct wl_listener *listener, void *data)
...
 
 	/*
-	 * If we didn't know the natural size when leaving fullscreen or
-	 * unmaximizing, then the pending size will be 0x0. In this case,
-	 * the pending x/y is also unset and we still need to position
-	 * the window.
+	 * The pending size will be empty in two cases:
+	 *   (1) when the view is first mapped
+	 *   (2) when leaving fullscreen or un-maximizing,
+	 *       if natural geometry wasn't known
```
tokyo4j: I suggest instead saying something like "An initially-maximized (or initially-fullscreened) view is made floating" so the reader can easily understand and test it.

tokyo4j:
LGTM other than my minor suggestions on documentations. Let's merge after the 0.9.4 release.

Thanks for this cleanup!

tokyo4j:
Thank you!

@johanmalm Please remember that 20254a4b44c00ae830152e68d36605cd735b0255 is slightly changing the behavior; actions like `<action name="AutoPlace" policy="cascade|center|automatic" />` now un-maximizes (un-fullscreens / un-tiles) the window.

jlindgren90:
> Please remember that 20254a4 is slightly changing the behavior; actions like <action name="AutoPlace" policy="cascade|center|automatic" /> now un-maximizes (un-fullscreens / un-tiles) the window.

Note that the behavior before was somewhat broken as it would try to place/center the view *without* un-maximize/full screen/tile. I doubt that anyone was relying on that behavior.

PR #3389, opened 2026-02-22
 layer: update pointer focus on unmap

tokyo4j:
Ensure that pointer focus is restored to the surface below when a layer surface or popup is unmapped. Can be tested like below:

1. Open a terminal and move the cursor over it so it becomes an I-beam
2. Open a launcher (e.g. fuzzel) so the cursor becomes an arrow
3. Close the launcher
4. The cursor should become an I-beam again

https://github.com/user-attachments/assets/f287cfb2-733b-4364-a5dc-1b30f41c6294

Note: this commit restores `cursor_update_focus()` in `handle_popup_destroy()` which was removed in #3165. I don't remember why it was removed.

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -378,6 +378,7 @@ handle_unmap(struct wl_listener *listener, void *data)
 	if (seat->focused_layer == layer_surface) {
 		try_to_focus_next_layer_or_toplevel(layer->server);
 	}
+	cursor_update_focus(layer->server);
 
 	layer->being_unmapped = false;
 }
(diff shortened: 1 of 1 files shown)


tokyo4j:
I think we can merge this PR in this cool-down period. If anyone has any concern, we can restore `cursor_update_focus()` in `handle_popup_destroy()` for this PR and add `cursor_update_focus()` in `handle_unmap()` as another PR.

johanmalm:
Good catch. Let's got for it.

PR #3398, opened 2026-02-24
interactive&action: minor refactor

tokyo4j:
- Add `server` argument to `interactive_set_grab_context()` to clarify it updates something in `server`.
- Add `const` qualifier to `ctx` argument of `interactive_set_grab_context()` and `cursor_get_resize_edges()`
- Reorder `ACTION_TYPE_RESIZE` enum entry for readability

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -97,9 +97,9 @@ enum action_type {
 	ACTION_TYPE_FOCUS,
 	ACTION_TYPE_UNFOCUS,
 	ACTION_TYPE_ICONIFY,
-	ACTION_TYPE_MOVE,
 	ACTION_TYPE_RAISE,
 	ACTION_TYPE_LOWER,
+	ACTION_TYPE_MOVE,
 	ACTION_TYPE_RESIZE,
 	ACTION_TYPE_RESIZE_RELATIVE,
 	ACTION_TYPE_MOVETO,
@@ -1260,6 +1260,16 @@ run_action(struct view *view, struct server *server, struct action *action,
 			view_minimize(view, true);
 		}
 		break;
+	case ACTION_TYPE_RAISE:
+		if (view) {
+			view_move_to_front(view);
+		}
+		break;
+	case ACTION_TYPE_LOWER:
+		if (view) {
+			view_move_to_back(view);
+		}
+		break;
 	case ACTION_TYPE_MOVE:
 		if (view) {
 			/*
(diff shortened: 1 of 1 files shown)


jlindgren90:
Off-topic: I would like to just make `server` global. There is only ever going to be one instance, and storing pointers to it in every random little other struct just seems pointless.

jlindgren90:
Nit: suggest putting RAISE and LOWER above MOVE and RESIZE to keep all the various move/resize variants together.

Other than that, LGTM.

tokyo4j:
> Off-topic: I would like to just make server global. There is only ever going to be one instance, and storing pointers to it in every random little other struct just seems pointless.

Hm, I don't have strong preferences on it. Your point makes sense, but I'm used to passing `server` everywhere and it will feel a bit strange at first to see `get_cursor_context()` taking no arguments if we make `server` global (of course I will get used to that too).

Feel free to open a discussion for it.

johanmalm:
@tokyo4j: is this for after the release?

@jlindgren90: I have thought many times about making server global. IIRC it actually is in sway even though they pass it around. I think one or the other is better - to avoid `-Wshadow`

jlindgren90:
I have a change locally to add `g_server` global and eliminate all the `struct server *server` params. Largely done with perl regexes. I can create a PR later if there's interest.

Other globals I've considered are `g_seat` and `g_theme`, just for brevity as they would already be accessible globally as `g_server.seat` and `*rc.theme` anyway.

johanmalm:
> I have a change locally to add `g_server` global and eliminate all the `struct server *server` params. Largely done with perl regexes. I can create a PR later if there's interest.
> 
> Other globals I've considered are `g_seat` and `g_theme`, just for brevity as they would already be accessible globally as `g_server.seat` and `*rc.theme` anyway.

Suggest separate discussion as there might be strong feelings on this :smile:

Agree with `theme`.

We ought to consider that multiple `seat`s may be implemented in future. I'm not sure if making it global at this stage really makes a difference in that regard, but it's a consideration.

tokyo4j:
> @tokyo4j: is this for after the release?

Yes. I forgot to mention that. I'm setting this PR as a draft just in case.

Consolatis:
Regarding making `struct server` and `struct theme` global: theoretically I don't have objections to them being available globally but I don't really have any preference in one way or the other. `struct rcxml rc` being global is good though I think.

However, if we go that route I'd prefer if we could wrap those pointers (and maybe also `struct rcxml rc`) into a single global struct instance (maybe something like `extern struct global global` or `extern struct global g` with my preference being `g`). That makes it much easier to see what is available globally and what is not rather than scanning through random header files to see what is exposed globally. And when reading code which makes use of those globals it also makes it clear that this var is something available globally rather than per compilation unit (or being declared within a function or supplied as function argument).

Issue #3399, opened 2026-02-24
Blocking apps cause "Data too big for buffer" and get disconnected


jls17:
### Problem description

In case applications do not respond fast enough to events, they get closed instantly. For example if the main thread of an application is blocked for some time and there are lots of mouse events arriving, labwc prints the following

```
00:17:42.169 [INFO] [wayland] Data too big for buffer (4088 + 20 > 4096).
00:17:42.201 [INFO] [wayland] error in client communication (pid 113111)
```

The mouse events are filling the buffer and once the buffer is full, the app will get closed immediately.

The client error may look like this:

```
Gdk-Message: 22:24:24.629: Lost connection to Wayland compositor.
```

... or ...

```
Gdk-Message: 22:41:18.663: Error flushing display: Broken pipe
```

Other compositors do have the same issue (see links below) and the currently recommended fix is to increase the ring-buffer to 1MB.

I've put together a small demo application. Once you click on the button the main thread gets blocked by some calculations. If there are lots of mouse events on that button/app while the app is blocked, labwc will print the error and the client gets disconnected. The application terminates.

The following recording does demonstrate this 4 times, just to show there are different errors on client side. On the left-side you see the output of labwc with verbose (`-V`) logging. On the right side I just use the shell to start the hello_world demo app.

https://github.com/user-attachments/assets/82728678-5b6e-4844-9a8d-11ed5e54433e

- This demo was taken from: https://www.gtk.org/docs/language-bindings/vala/
- just save the [attached file](https://github.com/user-attachments/files/25532345/hello_world.txt) as "hello_world.vala" and compile with `valac --pkg gtk4 hello_world.vala`

#### Possible fix/workarround

By default the ring-buffer is set to 4096 bytes. As suggested in other issues, increasing the buffer size to 1MB should at least help. For a PoC I did give that a try and put the following line into `server_start` function of labwc.

```
    wl_display_set_default_max_buffer_size(server->wl_display, 1024*1024);
```

After that the issue stopped happening and I can use those affected apps/games on labwc without making them disappear. This was just a quick code-adjustment to verify if it will help me. But I don't know if that line is better placed somewhere else. So I hope someone with more knowledge can look at this and decide what's the right way.

#### References
- __Sway__: https://github.com/swaywm/sway/pull/8532
- __Niri__: https://github.com/niri-wm/niri/issues/1989

#### I used the following
LightDM 1.32.0
Budgie-Desktop 10.10.1
labwc 0.9.3

#### Attachments
[...]

PR #3400, opened 2026-02-25
src/server.c: increase max client buffer size to 1 MiB

Consolatis:
Allows slow clients to better deal with large amounts of events coming in from labwc like from high refresh rate mice. Without this patch, the client is likely to get disconnected because the labwc side client write buffer is full.

----

Only compile time tested + single startup nested.

Fixes:
- #3399

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -59,7 +59,7 @@ wlroots = dependency(
 wlroots_has_xwayland = wlroots.get_variable('have_xwayland') == 'true'
 have_libsfdo = not get_option('icon').disabled()
 
-wayland_server = dependency('wayland-server', version: '>=1.19.0')
+wayland_server = dependency('wayland-server', version: '>=1.22.90')
 wayland_protos = dependency('wayland-protocols', version: '>=1.39')
 xkbcommon = dependency('xkbcommon')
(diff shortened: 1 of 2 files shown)


jls17:
I tested this version and the affected apps on my system don't terminate unexpectedly anymore.
Works for me.
Thank you

meson.build
```diff
@@ -59,7 +59,7 @@ wlroots = dependency(
 wlroots_has_xwayland = wlroots.get_variable('have_xwayland') == 'true'
 have_libsfdo = not get_option('icon').disabled()
 
-wayland_server = dependency('wayland-server', version: '>=1.19.0')
+wayland_server = dependency('wayland-server', version: '>=1.22.90')
```
johanmalm, line 62: Are you sure it's `1.22.90`?

sway's meson.build says `wayland_server = dependency('wayland-server', version: '>=1.21.0')` on the master branch.

johanmalm: Looks like it was added here: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/188

johanmalm: ...which was merged on Apr 8, 2024 - so yeah - looks like you're right as always :smile:

PR #3401, opened 2026-02-25
Handle xwayland not started (when HAVE_XWAYLAND, but lacking the binary)

quite:
Closes https://github.com/labwc/labwc/issues/434

### DIFF
--- a/src/config/session.c
+++ b/src/config/session.c
@@ -21,7 +21,6 @@
 #include "labwc.h"
 
 static const char *const env_vars[] = {
-	"DISPLAY",
 	"WAYLAND_DISPLAY",
 	"XDG_CURRENT_DESKTOP",
 	"XCURSOR_SIZE",
@@ -209,6 +208,21 @@ should_update_activation(void)
 }
 
 static void
+execute_update(const char *env_keys, const char *env_unset_keys, bool initialize)
+{
+	char *cmd =
+		strdup_printf("dbus-update-activation-environment %s",
+			initialize ? env_keys : env_unset_keys);
+	spawn_async_no_shell(cmd);
+	free(cmd);
+
+	cmd = strdup_printf("systemctl --user %s %s",
+		initialize ? "import-environment" : "unset-environment", env_keys);
+	spawn_async_no_shell(cmd);
+	free(cmd);
+}
+
+static void
 update_activation_env(bool initialize)
 {
 	if (!should_update_activation()) {
@@ -227,19 +241,18 @@ update_activation_env(bool initialize)
 	char *env_keys = str_join(env_vars, "%s", " ");
 	char *env_unset_keys = initialize ? NULL : str_join(env_vars, "%s=", " ");
 
-	char *cmd =
-		strdup_printf("dbus-update-activation-environment %s",
-			initialize ? env_keys : env_unset_keys);
-	spawn_async_no_shell(cmd);
-	free(cmd);
-
-	cmd = strdup_printf("systemctl --user %s %s",
-		initialize ? "import-environment" : "unset-environment", env_keys);
-	spawn_async_no_shell(cmd);
-	free(cmd);
+	execute_update(env_keys, env_unset_keys, initialize);
 
 	free(env_keys);
 	free(env_unset_keys);
+
+#if HAVE_XWAYLAND
+	if (server.xwayland) {
+		// DISPLAY is only set if xwayland was initialized successfully,
+		// so we only update the env in that case
+		execute_update("DISPLAY", "DISPLAY=", initialize);
+	}
+#endif
 }
 
 void
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -1166,7 +1166,7 @@ cursor_process_button_press(struct seat *seat, uint32_t button, uint32_t time_ms
 		if (layer && layer->current.keyboard_interactive) {
 			layer_try_set_focus(seat, layer);
 		}
-#ifdef HAVE_XWAYLAND
+#if HAVE_XWAYLAND
 	} else if (ctx.type == LAB_NODE_UNMANAGED) {
 		desktop_focus_view_or_surface(seat, NULL, ctx.surface,
 			/*raise*/ false);
--- a/src/server.c
+++ b/src/server.c
@@ -583,6 +583,8 @@ server_init(void)
 	server.workspace_tree = lab_wlr_scene_tree_create(&server.scene->tree);
 	server.xdg_popup_tree = lab_wlr_scene_tree_create(&server.scene->tree);
 #if HAVE_XWAYLAND
+	// Creating/setting this is harmless when xwayland support is built-in
+	// but xwayland could not be successfully started.
 	server.unmanaged_tree = lab_wlr_scene_tree_create(&server.scene->tree);
 #endif
 	server.menu_tree = lab_wlr_scene_tree_create(&server.scene->tree);
--- a/src/view.c
+++ b/src/view.c
@@ -54,6 +54,8 @@ view_from_wlr_surface(struct wlr_surface *surface)
 		return xdg_surface->data;
 	}
 #if HAVE_XWAYLAND
+	// Doing this is harmless even in the case that xwayland could not be
+	// successfully started.
 	struct wlr_xwayland_surface *xsurface =
 		wlr_xwayland_surface_try_from_wlr_surface(surface);
 	if (xsurface) {
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -1218,8 +1218,9 @@ xwayland_server_init(struct wlr_compositor *compositor)
 		wlr_xwayland_create(server.wl_display,
 			compositor, /* lazy */ !rc.xwayland_persistence);
 	if (!server.xwayland) {
-		wlr_log(WLR_ERROR, "cannot create xwayland server");
-		exit(EXIT_FAILURE);
+		wlr_log(WLR_ERROR, "failed to create xwayland server, continuing without");
+		unsetenv("DISPLAY");
+		return;
 	}
 	server.xwayland_new_surface.notify = handle_new_surface;
 	wl_signal_add(&server.xwayland->events.new_surface,
@@ -1308,6 +1309,11 @@ void
 xwayland_server_finish(void)
 {
 	struct wlr_xwayland *xwayland = server.xwayland;
+
+	if (!xwayland) {
+		return;
+	}
+
 	wl_list_remove(&server.xwayland_new_surface.link);
 	wl_list_remove(&server.xwayland_server_ready.link);
 	wl_list_remove(&server.xwayland_xwm_ready.link);


quite:
This is a draft attempt at making labwc okay with not being able to successfully start xwayland (even when built with the support). The issue is: https://github.com/labwc/labwc/issues/434

It contains 1 TODO which I consider blocking -- I could not immediately find a way for how to get hold of `server->xwayland` from `view.c`, for making decisions based on if xwayland has been successfully started or not.

There are also 3 TODOs which I think needs some consideration.

Happy for comments on these, and overall!

quite:
~~And one bigger oversight is that I did not realize that `struct server` does not even have member `xwayland` if HAVE_XWAYLAND is not set. Hm. Wonder if it would make sense to let it always have that, and let that be used as a "bool" for whether xwayland could be started of not. Or is that scope creep...~~

Actually scratch that, cursor.c did wrongly not use `#if` but had `#ifdef HAVE_XWAYLAND`, which did not do what's expected. Have harmonized to it uses `#if` like everywhere else.

johanmalm:
Thanks for working on this. I think it's a good addition that will help end users.

We're outside a merge window at the moment, but should get back into it next week. Can we look more deeply at it then?

src/config/session.c
```diff
@@ -226,6 +226,8 @@ update_activation_env(struct server *server, bool initialize)
 
 	wlr_log(WLR_INFO, "Updating dbus execution environment");
 
+	// TODO? should we avoid passing DISPLAY if we did not set it? when
```
johanmalm: Yes probably.

src/cycle/osd-field.c
```diff
@@ -62,7 +62,11 @@ get_type(struct view *view, bool short_form)
 		return short_form ? "[W]" : "[xdg-shell]";
 #if HAVE_XWAYLAND
 	case LAB_XWAYLAND_VIEW:
-		return short_form ? "[X]" : "[xwayland]";
+		// TODO? maybe type must never be LAB_XWAYLAND_VIEW if the
+		// xwayland server could not be started?
```
johanmalm: True, but probably better to guard against it anyway.

src/server.c
```diff
@@ -588,6 +588,9 @@ server_init(struct server *server)
 	server->workspace_tree = wlr_scene_tree_create(&server->scene->tree);
 	server->xdg_popup_tree = wlr_scene_tree_create(&server->scene->tree);
 #if HAVE_XWAYLAND
+	// TODO? is creating this and server->unmanaged_tree becoming non-NULL
```
johanmalm: I think so. Avoids having to guard against all instances of `unmanaged_tree`

src/view.c
```diff
@@ -54,6 +54,10 @@ view_from_wlr_surface(struct wlr_surface *surface)
 		return xdg_surface->data;
 	}
 #if HAVE_XWAYLAND
+	// TODO We have no access to server->xwayland here, so how can we avoid
+	// running this if we are built with xwayland support, but we want to
```
johanmalm: 1. Pass `*server` to this function - which looks like a pain from `src/tearing.c`
2. Create a translation unit wide server pointer (have done that in a few places search for `g_server`)
3. Make `struct server *server` global. @jlindgren90 mentioned this yesterday, so may be some wider interest in doing this to simplify things.

src/config/session.c
quite: Done

src/cycle/osd-field.c
quite: Done

src/server.c
quite: OK, leaving a comment about that.

src/view.c
quite: 1. Yes that looks esp painful
2. Hm, I can find a couple of static g_server instances in the past, but they were intentionally refactored away from it seems. Perhaps it's not a way forward?
3. Looks like somewhat involved, but if it is of interest for other purposes then... Thoughts @jlindgren90 ?

jlindgren90: I don't see why we need any extra check here to begin with. wlr_xwayland_surface_try_from_wlr_surface() should be benign even if called for an xdg-shell view - but shouldn't even be reached because xdg-shell are handled first.

johanmalm: Fair point.

quite: This is just me being very new to wlr and all. And being extra careful to not cause any mess given that. But that sounds promising..

johanmalm:
@quite Are you able to verify that it works by compiling with XWayland support, then uninstalling XWayland.

src/view.c
quite: So I just left a comment to the effect that this is done with the awareness that xwayland might not have been started.

quite:
> @quite Are you able to verify that it works by compiling with XWayland support, then uninstalling XWayland.

Yes, that works! I think the patch is workable now. I have especially looked at all code which is conditioned on `HAVE_XWAYLAND` at compile-time, and tried to make sure it is run only when `server->xwayland` is true (or if that does not matter).

We can look more in depth later when you have time, or merge-window is opening up. Thank you.

src/cycle/osd-field.c
```diff
@@ -62,7 +62,9 @@ get_type(struct view *view, bool short_form)
 		return short_form ? "[W]" : "[xdg-shell]";
 #if HAVE_XWAYLAND
 	case LAB_XWAYLAND_VIEW:
-		return short_form ? "[X]" : "[xwayland]";
+		if (view->server->xwayland) {
```
jlindgren90: I don't think this check is necessary/useful. If an xwayland view exists, we should label it as such.

src/input/cursor.c
```diff
@@ -1171,8 +1171,8 @@ cursor_process_button_press(struct seat *seat, uint32_t button, uint32_t time_ms
 		if (layer && layer->current.keyboard_interactive) {
 			layer_try_set_focus(seat, layer);
 		}
-#ifdef HAVE_XWAYLAND
-	} else if (ctx.type == LAB_NODE_UNMANAGED) {
+#if HAVE_XWAYLAND
+	} else if (server->xwayland && ctx.type == LAB_NODE_UNMANAGED) {
```
jlindgren90: I don't think this check is necessary/useful. If an unmanaged surface exists, we can treat it as such.

src/input/cursor.c
```diff
@@ -1615,7 +1615,9 @@ cursor_reload(struct seat *seat)
 {
 	cursor_load(seat);
 #if HAVE_XWAYLAND
-	xwayland_reset_cursor(seat->server);
+	if (seat->server->xwayland) {
```
jlindgren90: I would prefer to move this check inside `xwayland_reset_cursor()`, where (presumably) `server->xwayland` is used.

src/debug.c
```diff
@@ -179,7 +179,7 @@ get_special(struct server *server, struct wlr_scene_node *node)
 		return "cycle_state->preview_outline";
 	}
 #if HAVE_XWAYLAND
-	if (node == &server->unmanaged_tree->node) {
+	if (server->xwayland && node == &server->unmanaged_tree->node) {
```
jlindgren90: Please remove this check. This is debug code and just labeling the scene-node, which will still exist whether xwayland is running or not.

src/desktop.c
```diff
@@ -124,7 +124,7 @@ desktop_focus_view_or_surface(struct seat *seat, struct view *view,
 	if (view) {
 		desktop_focus_view(view, raise);
 #if HAVE_XWAYLAND
-	} else {
+	} else if (seat->server->xwayland) {
```
jlindgren90: Don't see the point of this check. `wlr_xwayland_surface_try_from_wlr_surface()` is intended to be safe to call for wayland and non-xwayland surfaces, regardless of whether xwayland is running.

src/desktop.c
```diff
@@ -291,7 +291,7 @@ get_cursor_context(struct server *server)
 
 #if HAVE_XWAYLAND
 	/* TODO: attach LAB_NODE_UNMANAGED node-descriptor to unmanaged surfaces */
-	if (node->type == WLR_SCENE_NODE_BUFFER) {
+	if (server->xwayland && node->type == WLR_SCENE_NODE_BUFFER) {
```
jlindgren90: I don't think this check is necessary/useful. If an unmanaged surface exists, we can treat it as such.

src/output.c
```diff
@@ -1069,12 +1069,14 @@ update_usable_area(struct output *output)
 	layers_arrange(output);
 
 #if HAVE_XWAYLAND
-	struct view *view;
-	wl_list_for_each(view, &output->server->views, link) {
-		if (view->mapped && view->type == LAB_XWAYLAND_VIEW) {
-			xwayland_adjust_usable_area(view,
-				output->server->output_layout,
-				output->wlr_output, &output->usable_area);
+	if (output->server->xwayland) {
```
jlindgren90: I don't think this check is necessary. If xwayland views exist, we should treat them as such.
The only thing to check is whether `xwayland_adjust_usable_area` uses `server->xwayland` internally -- and if it does, the check should take place there, inside `xwayland_adjust_usable_area`.

src/output.c
```diff
@@ -1087,7 +1089,9 @@ output_update_usable_area(struct output *output)
 	if (update_usable_area(output)) {
 		regions_update_geometry(output);
 #if HAVE_XWAYLAND
-		xwayland_update_workarea(output->server);
+		if (output->server->xwayland) {
```
jlindgren90: Would prefer to have this check moved inside `xwayland_update_workarea` (where `server->xwayland` is presumably used).

src/output.c
```diff
@@ -1109,7 +1113,9 @@ output_update_all_usable_areas(struct server *server, bool layout_changed)
 	}
 	if (usable_area_changed || layout_changed) {
 #if HAVE_XWAYLAND
-		xwayland_update_workarea(server);
+		if (server->xwayland) {
```
jlindgren90: Same as above, move the check inside `xwayland_update_workarea`.

src/server.c
```diff
@@ -589,6 +589,8 @@ server_init(struct server *server)
 	server->workspace_tree = wlr_scene_tree_create(&server->scene->tree);
 	server->xdg_popup_tree = wlr_scene_tree_create(&server->scene->tree);
 #if HAVE_XWAYLAND
+	// It is assumed that creating/setting this is harmless when xwayland
```
jlindgren90: You can remove "It is assumed". Creating the extra scene-tree is in fact harmless, even if it doesn't get used for anything.

src/view.c
```diff
@@ -54,6 +54,8 @@ view_from_wlr_surface(struct wlr_surface *surface)
 		return xdg_surface->data;
 	}
 #if HAVE_XWAYLAND
+	// Doing this should be harmless even in the case that xwayland could
```
jlindgren90: `s/should be/is/`

src/view.c
```diff
@@ -2211,7 +2213,7 @@ view_move_to_front(struct view *view)
 	 * to an incorrect X window depending on timing. To mitigate the
 	 * race, perform an explicit flush after restacking.
 	 */
-	if (view->type == LAB_XWAYLAND_VIEW) {
+	if (view->server->xwayland && view->type == LAB_XWAYLAND_VIEW) {
```
jlindgren90: I would prefer this check be moved inside `xwayland_flush()`.

jlindgren90:
@quite thanks for this. I left individual comments, but overall I think we should try to limit/move the checks for NULL `server->xwayland` to where `server->xwayland` is actually used. Most of the other checks should be unnecessary -- especially in code paths where we know an xwayland surface already exists.

src/input/cursor.c
quite: Ah. And such check is already present in there.

src/output.c
quite: Removing it here. And `xwayland_update_workarea` is already checking that of course, nice code.

src/output.c
quite: Removed from here

src/view.c
quite: Done (and the check if already there!)

src/desktop.c
quite: Removed it

src/input/cursor.c
quite: Removed it.

src/cycle/osd-field.c
quite: I removed it.

quite:
> @quite thanks for this. I left individual comments, but overall I think we should try to limit/move the checks for NULL `server->xwayland` to where `server->xwayland` is actually used. Most of the other checks should be unnecessary -- especially in code paths where we know an xwayland surface already exists.

Thank you. That all made sense of course, and I learned a lot about the structure of code and how it actually implements the compositor by going through these. So most of my additions could just go, nice!

jlindgren90 (approved):

quite:
> I did not fully understand all the env-var changes, maybe someone else can review that part.

The idea is that when xwayland could not be started, we don't ever call out to `dbus-update-activation-environment`/`systemctl --user {import,unset}-environment` with the `DISPLAY` variable (but still all other variables that are managed).

jlindgren90:
Don't the `dbus-update-activation-environment` / `import-environment` commands already skip `DISPLAY` if it's unset? I guess I'm not sure why any changes are necessary in that area.

quite:
> Don't the `dbus-update-activation-environment` / `import-environment` commands already skip `DISPLAY` if it's unset? I guess I'm not sure why any changes are necessary in that area.

From what I can see, the possible issue if we do not make this change, is that on session shutdown  `dbus-update-activation-environment DISPLAY=` will be called, and it *will* set an empty DISPLAY in that environment, ignoring that its actually not set in the local environment. But immediately after that, `systemctl --user unset-environment DISPLAY` will be called, and it will unset DISPLAY from some environment. But I'm not entirely sure of if or how the environments that `dbus-update-activation-environment` and `systemctl --user *-environment` affect respectively differ.

In any case, I guess the current code on main will also leave a `DISPLAY=` behind itself in some environment when shutting down -- or it won't then, if the `systemctl --user unset-environment` call overrides that.

johanmalm:
We now have merge conflicts on this one. @quite would you be have to fix?

Is it okay to call `dbus-update-activation-environment` twice when initializing the variables? (e.g. do the variables set with the first call remain set on the second call - which does not include them?).

quite:
> We now have merge conflicts on this one. @quite would you be have to fix?

That was done a few days ago! :)

> Is it okay to call `dbus-update-activation-environment` twice when initializing the variables? (e.g. do the variables set with the first call remain set on the second call - which does not include them?).

Yes. Reading dbus-update-activation-environment.1, it says that it only ever touches variables in that environment which are mentioned on the commandline.

johanmalm:
> > We now have merge conflicts on this one. @quite would you be have to fix?
> 
> That was done a few days ago! :)

Well, I think we have new ones since Sunday.

> > Is it okay to call `dbus-update-activation-environment` twice when initializing the variables? (e.g. do the variables set with the first call remain set on the second call - which does not include them?).
> 
> Yes. Reading dbus-update-activation-environment.1, it says that it only ever touches variables in that environment which are mentioned on the commandline.

Thanks. That was the check I was looking for. Thanks for doing the research.

quite:
Ah, rebased now sry.

quite:
Wonder if it would be useful to add a config option to disable xwayland, even if support is compiled in, and xwayland exists. But perhaps that can be considered out of scope for this patch.

johanmalm:
LGTM. Thanks.

johanmalm:
> Wonder if it would be useful to add a config option to disable xwayland, even if support is compiled in, and xwayland exists. But perhaps that can be considered out of scope for this patch.

Happy to support that. IIRC that's what sway does.

Issue #3410, opened 2026-03-02
mako notification deamon eating up RAM with output off


coruja-eu:
### Problem description

I encountered a quite strange issue with latest labwc (0.9.4), to be precise not with labwc itself but with the mako notification daemon.

When my output is turned off (by swayidle/swaylock and wlopm), and a notification is sent by mako, mako fills up my RAM within short time. With labwc 0.9.3 this issue didn't occur, and doesn't after downgrading to 0.9.3 again, but it is reproducible with 0.9.4.

Together with @Consolatis (thanks for helping me debugging on IRC, btw), I created some mako wayland debug log, which most interesting lines (which are repeated again and again and fill up the log very quickly) are as follow:

> [3383347.133] {Default Queue}  -> wl_compositor#5.create_surface(new id wl_surface#13)
[3383347.141] {Default Queue}  -> zwlr_layer_shell_v1#9.get_layer_surface(new id zwlr_layer_surface_v1#14, wl_surface#13, nil, 3, "notifications")
[3383347.147] {Default Queue}  -> zwlr_layer_surface_v1#14.set_size(420, 94)
[3383347.150] {Default Queue}  -> zwlr_layer_surface_v1#14.set_anchor(9)
[3383347.153] {Default Queue}  -> zwlr_layer_surface_v1#14.set_margin(0, 0, 0, 0)
[3383347.156] {Default Queue}  -> wl_surface#13.commit()
[3383347.232] {Display Queue} wl_display#1.delete_id(12)
[3383347.238] {Display Queue} wl_display#1.delete_id(3)
[3383347.240] {Default Queue} zwlr_layer_surface_v1#14.closed()
[3383347.243] {Default Queue}  -> zwlr_layer_surface_v1#14.destroy()
[3383347.246] {Default Queue}  -> wl_surface#13.destroy()
[3383347.395] {Default Queue}  -> wl_compositor#5.create_surface(new id wl_surface#3)
[3383347.402] {Default Queue}  -> zwlr_layer_shell_v1#9.get_layer_surface(new id zwlr_layer_surface_v1#12, wl_surface#3, nil, 3, "notifications")
[3383347.407] {Default Queue}  -> zwlr_layer_surface_v1#12.set_size(420, 94)
[3383347.410] {Default Queue}  -> zwlr_layer_surface_v1#12.set_anchor(9)
[3383347.412] {Default Queue}  -> zwlr_layer_surface_v1#12.set_margin(0, 0, 0, 0)
[...]

PR #3411, opened 2026-03-02
layers.c: allow layer surfaces on disabled outputs

Consolatis:
#3361 added a check for the output being enabled when creating a layershell surface with automatic output selection, likely on accident. This causes labwc to send a `layer_surface.close()` and `wl_surface.destroy()` to layershell clients when creating a new layershell surface on a disabled output which is still part of the layout (e.g. `wlopm --off`).

This MR replaces the `output_is_usable()` check with a dedicated check, bypassing the check for the output being enabled.

CC @jlindgren90 

Fixes:
- #3410

### DIFF
--- a/src/layers.c
+++ b/src/layers.c
@@ -687,7 +687,14 @@ handle_new_layer_surface(struct wl_listener *listener, void *data)
 
 	if (!layer_surface->output) {
 		struct output *output = output_nearest_to_cursor(server);
-		if (!output_is_usable(output)) {
+		if (!output || !output->scene_output) {
+			/*
+			 * We are not using output_is_usable() here because
+			 * it also checks if the output is enabled which is
+			 * not the case when wlopm is used. Thus we allow
+			 * layer surfaces to spawn on disabled outputs as
+			 * long as it is part of the scene layout.
+			 */
 			wlr_log(WLR_INFO,
 				"No output available to assign layer surface");
 			wlr_layer_surface_v1_destroy(layer_surface);


jlindgren90:
Hmm, I was under the impression that `output_is_usable()` still returned true if an output was turned off via `wlopm`. There might be other issues if it doesn't, like relocating views off of an output in powersave.

jlindgren90:
Anyway, LGTM as fixing the regression.

Consolatis:
> Hmm, I was under the impression that output_is_usable() still returned true if an output was turned off via wlopm. There might be other issues if it doesn't, like relocating views off of an output in powersave.

AFAIR wlopm simply causes labwc to disable the output but doesn't remove it from the layout.

I didn't keep up with the changes to the output evacuation / restore of window position mechanism. I am also not sure of the implications of simply removing the `wlr_output->enabled` check for `output_is_usable()`.

So my question is: is the window position restore on output re-connect (potentially after already turned off via wlopm) affected as well and if yes, should we fix that before the 0.9.5 bugfix release?

tokyo4j:
> So my question is: is the window position restore on output re-connect (potentially after already turned off via wlopm) affected as well and if yes, should we fix that before the 0.9.5 bugfix release?

I don't think window repositioning is affected by #3361. Though I and @jlindgren90 made some simplifications and improvements (trying to match output name & keep relative position), windows in output disabled by `wlopm` should be relocated to another enabled output anyway in both 0.9.3 and 0.9.4.

jlindgren90:
Agree with @tokyo4j after thinking about it. The view positioning logic is already good at handling disconnected/disabled outputs, and `wlopm --off` is handled the same in the regard. So it should be fine.

SjoerdV:
Just wondering is this change has been tested with a full S3 suspend cycle (sleep: swayidle -> `wlopm --off \*` AND resume: swayidle -> `wlopm --on \*`) with `autoEnableOutputs=no`.

I have my own 'connected display re-init' script running in the `after-resume` phase of swayidle to setup my display configuration just right depending on where I am. 

I did notice labwc or wlroots _sometimes_ crashing (probably some kind of race involving external displays and thunderbolt docks...) to the greeter after a resume. So I'm asking for a friend.

johanmalm:
> Just wondering is this change has been tested with a full S3 suspend cycle (sleep: swayidle -> `wlopm --off \*` AND resume: swayidle -> `wlopm --on \*`) with `autoEnableOutputs=no`.
> 
> I have my own 'connected display re-init' script running in the `after-resume` phase of swayidle to setup my display configuration just right depending on where I am.
> 
> I did notice labwc or wlroots _sometimes_ crashing (probably some kind of race involving external displays and thunderbolt docks...) to the greeter after a resume. So I'm asking for a friend.

Just seen this. Don't like the sound of that. If you still get this on the master branch, I'd be grateful for a new issue if that's okay.

PR #3421, opened 2026-03-05
action: sync action arrays

Consolatis:
The previous commit desynchronized the action type arrays causing Move operations to be treated as Raise actions.

Fixes: #3398
Reported-By: Domo via IRC


tokyo4j:
Thanks! Sorry for my mistake.

PR #3422, opened 2026-03-06
[RFC] make struct server global

jlindgren90:
Declare a global instance of `struct server` as `g_server`. (Naming open to discussion of course, but that's my suggestion as it's brief but clearly distinct from local vars.)

Rationale: it is unlikely to ever have more than one instance per process, and needs to last for the lifetime of the process. Accessing it indirectly through pointers embedded in every other struct just adds noise to the code.

The 1st commit contains the only manually authored changes (edit: also the 3rd commit, which is strictly whitespace). The 2nd commit (which is the lion's share of the LOC changed) is entirely automated and reproducible via the bash/perl script in the commit message.

### DIFF
--- a/include/cycle.h
+++ b/include/cycle.h
@@ -133,7 +133,7 @@ struct cycle_osd_impl {
 	 */
 	void (*init)(struct cycle_osd_output *osd_output);
 	/*
-	 * Update the OSD to highlight g_server.cycle.selected_view.
+	 * Update the OSD to highlight server.cycle.selected_view.
 	 */
 	void (*update)(struct cycle_osd_output *osd_output);
 };
@@ -157,7 +157,7 @@ void cycle_osd_scroll_init(struct cycle_osd_output *osd_output,
 	int nr_cols, int nr_rows, int nr_visible_rows,
 	float *border_color, float *bg_color);
 
-/* Scroll the OSD to show g_server.cycle.selected_view if needed */
+/* Scroll the OSD to show server.cycle.selected_view if needed */
 void cycle_osd_scroll_update(struct cycle_osd_output *osd_output);
 
 extern struct cycle_osd_impl cycle_osd_classic_impl;
--- a/include/decorations.h
+++ b/include/decorations.h
@@ -7,12 +7,12 @@ struct view;
 struct wlr_surface;
 
 void kde_server_decoration_init(void);
-void xdg_server_decoration_init(void);
+void xdserver_decoration_init(void);
 
 void kde_server_decoration_update_default(void);
 void kde_server_decoration_set_view(struct view *view, struct wlr_surface *surface);
 
 void kde_server_decoration_finish(void);
-void xdg_server_decoration_finish(void);
+void xdserver_decoration_finish(void);
 
 #endif /* LABWC_DECORATIONS_H */
--- a/include/input/cursor.h
+++ b/include/input/cursor.h
@@ -82,7 +82,7 @@ void cursor_context_save(struct cursor_context_saved *saved_ctx,
 
 /**
  * cursor_get_resize_edges - calculate resize edge based on cursor position
- * @cursor - the current cursor (usually g_server.seat.cursor)
+ * @cursor - the current cursor (usually server.seat.cursor)
  * @cursor_context - result of get_cursor_context()
  *
  * Calculates the resize edge combination that is most appropriate based
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -313,7 +313,7 @@ struct server {
 };
 
 /* defined in main.c */
-extern struct server g_server;
+extern struct server server;
 
 void xdg_popup_create(struct view *view, struct wlr_xdg_popup *wlr_popup);
 void xdg_shell_init(void);
@@ -405,8 +405,8 @@ void seat_focus_override_end(struct seat *seat, bool restore_focus);
 /**
  * interactive_anchor_to_cursor() - repositions the geometry to remain
  * underneath the cursor when its size changes during interactive move.
- * This function also resizes g_server.grab_box and repositions it to remain
- * underneath g_server.grab_{x,y}.
+ * This function also resizes server.grab_box and repositions it to remain
+ * underneath server.grab_{x,y}.
  *
  * geo->{width,height} are provided by the caller.
  * geo->{x,y} are computed by this function.
--- a/include/menu/menu.h
+++ b/include/menu/menu.h
@@ -84,10 +84,10 @@ struct menu *menu_get_by_id(const char *id);
 /**
  * menu_open_root - open menu on position (x, y)
  *
- * This function will close g_server.menu_current, open the
- * new menu and assign @menu to g_server.menu_current.
+ * This function will close server.menu_current, open the
+ * new menu and assign @menu to server.menu_current.
  *
- * Additionally, g_server.input_mode will be set to LAB_INPUT_STATE_MENU.
+ * Additionally, server.input_mode will be set to LAB_INPUT_STATE_MENU.
  */
 void menu_open_root(struct menu *menu, int x, int y);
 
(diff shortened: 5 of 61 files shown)


jlindgren90:
Potential follow-ups:
- `g_seat` (in place of `g_server.seat`)
- `g_theme` (in place of `g_server.theme` or `rc.theme`)

Consolatis:
Copy from https://github.com/labwc/labwc/pull/3398#issuecomment-4004094654
> Regarding making `struct server` and `struct theme` global: theoretically I don't have objections to them being available globally but I don't really have any preference in one way or the other. `struct rcxml rc` being global is good though I think.
> 
> However, if we go that route I'd prefer if we could wrap those pointers (and maybe also `struct rcxml rc`) into a single global struct instance (maybe something like `extern struct global global` or `extern struct global g` with my preference being `g`). That makes it much easier to see what is available globally and what is not rather than scanning through random header files to see what is exposed globally. And when reading code which makes use of those globals it also makes it clear that this var is something available globally rather than per compilation unit (or being declared within a function or supplied as function argument).

So personally I'd prefer something like `g.server` (and `g.rc`, `g.theme`) rather than `g_server` and `g_theme`. `g.server` could also be a pointer to the `main()` stack allocated `struct server`, that removes the need to use `&g.server` (or the need to remember to use `.` rather than `->`) everywhere.

jlindgren90:
I'd like to hear @johanmalm's thoughts, but the `g.` scheme would not be my preference, because:
1. it's "novel" whereas the `g_` prefix convention is fairly widespread AFAIK
2. it would be hidden anywhere we might have used a local variable named `g` (haven't searched)
3. it requires a central header to pull in all the different types (`struct rcxml`, `struct theme`, etc.) which is a step backward in my opinion. I've personally done some work already to break up `labwc.h`, and I was thinking of keeping at least `rc` declared in `rcxml.h` (where it is today), `g_theme` declared in `theme.h`, and the other two in `labwc.h` for now, while maybe splitting out `seat.h` in future. The actual definitions could live in `main.c` if we want a place to see them all together.

Consolatis:
> * it's "novel" whereas the `g_` prefix convention is fairly widespread AFAIK

Is there a widespread convention how to differentiate single source file globals from "project wide" ones?

> * it would be hidden anywhere we might have used a local variable named `g` (haven't searched)

I think we have a check for shadowing enabled by default so we'd know about it automatically.

> * it requires a central header to pull in all the different types (`struct rcxml`, `struct theme`, etc.) which is a step backward in my opinion.

If those are pointers, couldn't they be forward declared? Anything which uses those vars would need to include the relevant headers anyway.

jlindgren90:
> Is there a widespread convention how to differentiate single source file globals from "project wide" ones?

It's certainly not universal, but I guess `g_` is somewhat "commonly known" along with e.g. `s_` for file-scoped static variables and (in C++) `m_` for member variables. Ref: https://en.wikipedia.org/wiki/Hungarian_notation#Examples

> I think we have a check for shadowing enabled by default so we'd know about it automatically.

Sure, it just means we can't use `g` for both purposes in the same function. Probably a minor/theoretical annoyance.

> If those are pointers, couldn't they be forward declared? Anything which uses those vars would need to include the relevant headers anyway.

Sure, that's an option. I still slightly prefer `g_server` to `g->server` but I suppose it's mainly cosmetic/purist to not like the extra indirection.

tokyo4j:
Though I don't really have any preferences on whether to make `server` global, I agree with @jlindgren90's statements. I've seen the `g_` prefix several times and it's easy for me to interpret it as a global.

Consolatis:
> I've seen the `g_` prefix several times and it's easy for me to interpret it as a global.

Agree, interpreting a `g_` var as global is fine. But how does one discover the available globals if they are sprinkled around in header files and there is no single point of reference like a struct?

domo141:
`git grep 'struct .* g_' '**.[ch]'`

or someone(tm) documents all globals (yeah right...)

but does one need to discover all globals  -- one sees globals that are interesting at the moment whenever one encounters those while looking the code.

(using `g_` as prefix "collides" with glib convertion in function names so that makes it a bit harder to "grep" those)

But in any case, whether there is `g_server` or structures inside `g`-structure I like this
approach of removing one pointer indirction.

jlindgren90:
> But how does one discover the available globals if they are sprinkled around in header files and there is no single point of reference like a struct?

We could put all of the actual global definitions in `main.c`, regardless of which header contains the `extern` declaration (and I can move the block comment from labwc.h into main.c as well). Then they are all discoverable in one place. Really not much different from the current scheme where they are all on the stack of `main`.

johanmalm:
I prefer `server`, although am happy to go with `g_server` if that's a consensus.

Suggest that we:
- Move `struct server server = { 0 };` next to `struct rcxml rc = { 0 };` at the start of `main.c`.
- Add a comment there to explain that these are global and that they are `extern`ed in `labwc.h` (or wherever) and `config/rcxml.h`.

Ref:
- https://github.com/labwc/labwc/blob/16c5373be5eafad3d3b854341d39b68c8d21203d/src/main.c#L17

I/we have thus far used `g_server` as a workaround within the confines of a small number of translation units, which probably hasn't been helpful in setting expectations around naming conventions here.

As a slight aside, when I first started writing `labwc`, I did consider following the convention `struct labwc_server *labwc_server` as many other compositors seemed to at the time. But I decided that we're a library when such namespace considerations become more important, and decided that brevity is preferable and stuck with simply `server`.

I also considered making `server` global rather than passing it around. I wish I had, which is why I look at this PR with kind eyes. The whole passing around of the global state possibly makes it feel like the true input/output to each function is clearer, but the reality is that we're still dealing with a global state. A compositor has a high degree of coupling which is hard to avoid and I think it's fairer to be explicit about it.

Interestingly `sway` actually `externs` `sway_server`! I discovered this a long time ago :smile: But they still pass it around most of the time, so the direct usage of the global one is very low.

Ref:
- https://github.com/swaywm/sway/blob/6d25b100a23a17e9663cab5c286934089f2c4460/include/sway/server.h#L153

So in sway, one can do things like this:

```
diff --git a/sway/server.c b/sway/server.c
index 5d446579..61bc7381 100644
--- a/sway/server.c
+++ b/sway/server.c
@@ -240,6 +240,13 @@ static void handle_new_foreign_toplevel_capture_request(struct wl_listener *list
 	wlr_ext_foreign_toplevel_image_capture_source_manager_v1_request_accept(request, view->image_capture_source);
 }
 
+static void chat(void)
+{
+	if (server.wl_display) {
+		sway_log(SWAY_ERROR, "asdf");
+	}
+}
+
 bool server_init(struct sway_server *server) {
 	sway_log(SWAY_DEBUG, "Initializing Wayland server");
 	server->wl_display = wl_display_create();
@@ -251,6 +258,8 @@ bool server_init(struct sway_server *server) {
 	wlr_fixes_create(server->wl_display, 1);
 	root = root_create(server->wl_display);
 
+	chat();
+
 	server->backend = wlr_backend_autocreate(server->wl_event_loop, &server->session);
 	if (!server->backend) {
 		sway_log(SWAY_ERROR, "Unable to create backend");
```

There is something clean about only having one global state `extern` struct, so suggest we make `server` it and only access `rc`, `theme` and `seat` via `server`, so `server.theme`, etc. I don't feel strongly about this last point though, it's just a mental model of mine.

jlindgren90:
Happy to change `g_server` to just `server` if that's the consensus. No strong opinion on my side.

I would suggest keeping `rc` separate, since (a) that's the status quo and (b) `rc.<setting-name>` is repeated quite a lot; changing it to `server.rc` would increase a lot of line lengths.

I'm good with `server.seat` being the canonical location of `struct seat`, rather than a separate global (especially since "seat" can also refer to `struct wlr_seat`).

Currently, `struct theme` is accessible though both `struct server` and `struct rcxml`; maybe we could settle on `rc.theme`, since it's conceptually part of the configuration.

johanmalm:
> I would suggest keeping `rc` separate, since (a) that's the status quo and (b) `rc.<setting-name>` is repeated quite a lot; changing it to `server.rc` would increase a lot of line lengths.

Good point. Happy with `rc` as is.

> Currently, `struct theme` is accessible though both `struct server` and `struct rcxml`; maybe we could settle on `rc.theme`, since it's conceptually part of the configuration.

Yes, `rc.theme` is nice.

jlindgren90:
I added a commit renaming `g_server` back to just `server`. Trying to go back and update the original commits is awkward since it ends up with e.g. `struct server *server = &server`, which doesn't work.

Consolatis:
ACK to global `struct server server`

johanmalm:
Looks good. Thanks @jlindgren90

PR #3250, opened 2025-12-02
add support for color-management-v1

ManuLinares:

PR #3424, opened 2026-03-06
HDR support

kode54:
This patch series also assumes one of my other PRs will be applied first, so I'll rebase it if/when that is applied to chase/0.20.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -175,6 +175,7 @@ this is for compatibility with Openbox.
   <adaptiveSync>no</adaptiveSync>
   <allowTearing>no</allowTearing>
   <autoEnableOutputs>yes</autoEnableOutputs>
+  <hdr>no</hdr>
   <reuseOutputMode>no</reuseOutputMode>
   <xwaylandPersistence>no</xwaylandPersistence>
   <primarySelection>yes</primarySelection>
@@ -239,6 +240,11 @@ this is for compatibility with Openbox.
 	swayidle -w timeout 600 \\
 	    'pkill kanshi ; wlopm --off \*' resume 'kanshi & wlopm --on \*'
 	```
+
+*<core><hdr>* [yes|no]
+	Automatically enable HDR support on outputs when configuring them,
+	where supported by the particular output device and display. Default
+	is no.
 
 *<core><reuseOutputMode>* [yes|no]
 	Try to re-use the existing output mode (resolution / refresh rate).
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -36,6 +36,12 @@ enum tearing_mode {
 	LAB_TEARING_FULLSCREEN_FORCED,
 };
 
+enum render_bit_depth {
+	LAB_RENDER_BIT_DEPTH_DEFAULT = 0,
+	LAB_RENDER_BIT_DEPTH_8,
+	LAB_RENDER_BIT_DEPTH_10,
+};
+
 enum tiling_events_mode {
 	LAB_TILING_EVENTS_NEVER = 0,
 	LAB_TILING_EVENTS_REGION = 1 << 0,
@@ -74,6 +80,7 @@ struct rcxml {
 	int gap;
 	enum adaptive_sync_mode adaptive_sync;
 	enum tearing_mode allow_tearing;
+	enum render_bit_depth target_render_depth;
 	bool auto_enable_outputs;
 	bool reuse_output_mode;
 	uint32_t allowed_interfaces;
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1070,6 +1070,16 @@ set_tearing_mode(const char *str, enum tearing_mode *variable)
 	}
 }
 
+static void
+set_hdr_mode(const char *str, enum render_bit_depth *variable)
+{
+	if (parse_bool(str, -1) == 1) {
+		*variable = LAB_RENDER_BIT_DEPTH_10;
+	} else {
+		*variable = LAB_RENDER_BIT_DEPTH_8;
+	}
+}
+
 /* Returns true if the node's children should also be traversed */
 static bool
 entry(xmlNode *node, char *nodename, char *content)
@@ -1134,6 +1144,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		set_adaptive_sync_mode(content, &rc.adaptive_sync);
 	} else if (!strcasecmp(nodename, "allowTearing.core")) {
 		set_tearing_mode(content, &rc.allow_tearing);
+	} else if (!strcasecmp(nodename, "Hdr.core")) {
+		set_hdr_mode(content, &rc.target_render_depth);
 	} else if (!strcasecmp(nodename, "autoEnableOutputs.core")) {
 		set_bool(content, &rc.auto_enable_outputs);
 	} else if (!strcasecmp(nodename, "reuseOutputMode.core")) {
@@ -1505,6 +1517,7 @@ rcxml_init(void)
 	rc.gap = 0;
 	rc.adaptive_sync = LAB_ADAPTIVE_SYNC_DISABLED;
 	rc.allow_tearing = LAB_TEARING_DISABLED;
+	rc.target_render_depth = LAB_RENDER_BIT_DEPTH_DEFAULT;
 	rc.auto_enable_outputs = true;
 	rc.reuse_output_mode = false;
 	rc.allowed_interfaces = UINT32_MAX;
--- a/src/output-state.c
+++ b/src/output-state.c
@@ -24,6 +24,7 @@ output_state_init(struct output *output)
 			backup_config, output->wlr_output);
 
 	wlr_output_head_v1_state_apply(&backup_head->state, &output->pending);
+
 	wlr_output_configuration_v1_destroy(backup_config);
 }
 
--- a/src/output.c
+++ b/src/output.c
@@ -9,6 +9,7 @@
 #define _POSIX_C_SOURCE 200809L
 #include "output.h"
 #include <assert.h>
+#include <drm_fourcc.h>
 #include <strings.h>
 #include <wlr/backend/wayland.h>
 #include <wlr/config.h>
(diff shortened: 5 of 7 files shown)


kode54:
Damn, I thought I worked out all the kinks. This should be better now.

kode54:
Excellent, tests pass. I already verified it functioned correctly, but hadn't run the battery of automated testing against it, oops. Especially dumb that I missed that legacy function declaration.

Consolatis:
Same here, rebase would be good.

Regarding this PR vs #3250, is there anything missing in either of the PRs? Currently its hard to say because both PRs include the chase commits which make it pretty difficult to see the actual changes.

GitHub sadly doesn't seem to have a proper rebasing feature which allows to drop some of the commits.

src/output.c
```diff
@@ -43,6 +44,45 @@
...
+	enum render_bit_depth render_bit_depth = get_config_render_bit_depth();
+	if (render_bit_depth == LAB_RENDER_BIT_DEPTH_10 &&
+		bit_depth_from_format(render_format) == render_bit_depth) {
+		// 10-bit was set successfully before, try to save some tests by reusing the format
+		wlr_output_state_set_render_format(&output->pending, render_format);
+	} else if (render_bit_depth == LAB_RENDER_BIT_DEPTH_10) {
+		wlr_output_state_set_render_format(&output->pending, DRM_FORMAT_XBGR2101010);
+	} else {
+		wlr_output_state_set_render_format(&output->pending, DRM_FORMAT_XBGR8888);
+	}
```
Consolatis, line 194: It seems we are limiting the render formats to a few hardcoded formats. This feels a bit suspicious.

kode54: What do you propose instead? Probing the DRM for all supported formats and picking an appropriate one? I wouldn't know how to do that, either.

Consolatis:
#3469 has been merged which is the same as parts of 4b545fdcbd388d92895e7dd8d06c42d7ebdd4f1e.

As for the hardcoded output render_formats: I think we should add some more here. This collection seems sensible to me:
```c
uint32_t output_formats_8bit[] = {
	/* 32 bpp RGB with 8 bit component width */
	DRM_FORMAT_XRGB8888,
	DRM_FORMAT_XBGR8888,
	DRM_FORMAT_RGBX8888,
	DRM_FORMAT_BGRX8888,

	DRM_FORMAT_ARGB8888,
	DRM_FORMAT_ABGR8888,
	DRM_FORMAT_RGBA8888,
	DRM_FORMAT_BGRA8888,

	/* 24 bpp RGB */
	DRM_FORMAT_RGB888,
	DRM_FORMAT_BGR888,
};

uint32_t output_formats_10bit[] = {
	/* 32 bpp RGB with 10 bit component width */
	DRM_FORMAT_XRGB2101010,
	DRM_FORMAT_XBGR2101010,
	DRM_FORMAT_RGBX1010102,
	DRM_FORMAT_BGRX1010102,

	DRM_FORMAT_ARGB2101010,
	DRM_FORMAT_ABGR2101010,
	DRM_FORMAT_RGBA1010102,
	DRM_FORMAT_BGRA1010102,
};
```
We should then prioritize from top to bottom for 8 bit and 10 bit respectively. wlroots and thus labwc currently uses `XRGB8888` by default so that should be kept as top choice for the 8 bit render format. So something like this:
```c
bool
output_set_render_format(struct output *output, uint32_t candidates[], size_t count)
{
	for (size_t i = 0; i < count; i++) {
		wlr_output_state_set_render_format(&output->pending, candidates[i]);
		if (wlr_output_test_state(&output->pending)) {
			return true;
		}
	}
	return false;
}
```

src/output.c
Consolatis: I've looked a bit into it and its pretty messy because its basically impossible to figure out the component bit width from `drm_fourcc.h` without adding manual lookup tables. I posted a list of formats which seem reasonable below.

johanmalm:
> We should then prioritize from top to bottom for 8 bit and 10 bit respectively. wlroots and thus labwc currently uses `XRGB8888` by default so that should be kept as top choice for the 8 bit render format.

Thanks @Consolatis 

It feels right, but I'm a bit out of my depth on offering an opinion on the those arrays and the general approach.

How would you feel about us asking kennylevinsen and/or emersion for an ex-officio review as a one off?

kode54:
This should be good to go now.

johanmalm:
@Consolatis Would you be happy to review this one?

Consolatis:
Sorry for the delay, will do a detailed review in the next days. In general the approach seems fine.

docs/labwc-config.5.scd
```diff
@@ -240,6 +241,17 @@ this is for compatibility with Openbox.
 	    'pkill kanshi ; wlopm --off \*' resume 'kanshi & wlopm --on \*'
 	```
 
+*<core><hdr>* [yes|no]
+	Automatically enable HDR support on outputs when configuring them,
+	where supported by the particular output device and display. Default
+	is no.
+
+	Note: Enabling this option will also currently make all non-HDR
+	outputs render to 10 bits per channel as well, but still sRGB.
```
Consolatis: Is this still the case with the new `wlr_output_test_state()` added? I think that one checks if the output allows scanout in the given format but I don't know how display drivers implement the supported formats for DRM scanout, e.g. if they change depending on the capabilities of the connected monitor.

include/config/rcxml.h
```diff
@@ -74,6 +85,7 @@ struct rcxml {
 	int gap;
 	enum adaptive_sync_mode adaptive_sync;
 	enum tearing_mode allow_tearing;
+	enum hdr_mode hdr;
```
Consolatis: I think this can be `enum render_bit_deph target_render_depth` instead, initialized to `LAB_RENDER_BIT_DEPTH_8` in `rcxml_init()`. The `enum hdr_mode` could then be removed completely.

src/config/rcxml.c
```diff
@@ -1070,6 +1070,16 @@ set_tearing_mode(const char *str, enum tearing_mode *variable)
...
 }
 
+static void
+set_hdr_mode(const char *str, enum hdr_mode *variable)
+{
+	if (parse_bool(str, -1) == 1) {
+		*variable = LAB_HDR_ENABLED;
+	} else {
+		*variable = LAB_HDR_DISABLED;
+	}
```
Consolatis, line 1080: This could then directly set `LAB_RENDER_BIT_DEPTH_8` / `LAB_RENDER_BIT_DEPTH_10`.

src/output-state.c
```diff
@@ -24,6 +24,10 @@ output_state_init(struct output *output)
 			backup_config, output->wlr_output);
 
 	wlr_output_head_v1_state_apply(&backup_head->state, &output->pending);
+
+	/* This must be applied outside of config */
+	output_state_setup_hdr(output, true);
+
```
Consolatis: I am slightly confused, why do we need to set this more than once for an output? Same question for the one added in `output_state_commit()`. We also don't set for example the resolution each time we commit a new frame.

src/output.c
```diff
@@ -51,6 +52,126 @@
...
+	DRM_FORMAT_RGBA8888,
+	DRM_FORMAT_BGRA8888,
+
+	/* 24 bpp RGB */
+	DRM_FORMAT_RGB888,
+	DRM_FORMAT_BGR888,
+};
+
+static const size_t output_formats_8bit_count = sizeof(output_formats_8bit) /
+	sizeof(output_formats_8bit[0]);
```
Consolatis: This (and the one for 10 bit) is not needed, we can use `ARRAY_SIZE()` as argument to the iter functions.

src/output.c
```diff
@@ -1136,3 +1260,57 @@ output_set_has_fullscreen_view(struct output *output, bool has_fullscreen_view)
...
+		if (!silent) {
+			wlr_log(WLR_INFO, "Cannot enable HDR on output %s: %s",
+				output->wlr_output->name, unsupported_reason);
+		}
+		enabled = false;
+	}
+
+	if (!enabled) {
+		if (output->wlr_output->supported_primaries != 0 ||
+			output->wlr_output->supported_transfer_functions != 0) {
```
Consolatis: Multiline-ifs should have all lines after the first one double indented so its easier to distinguish from the body.

src/output.c
```diff
@@ -51,6 +52,126 @@
...
+	uint32_t render_format = output->wlr_output->render_format;
+	bool hdr_succeeded = false;
+	enum render_bit_depth render_bit_depth = get_config_render_bit_depth();
+	if (render_bit_depth == LAB_RENDER_BIT_DEPTH_10 &&
+		bit_depth_from_format(render_format) == render_bit_depth) {
+		/* 10-bit was set successfully before, try to save some
+		 * tests by reusing the format
+		 */
+		wlr_output_state_set_render_format(&output->pending, render_format);
+		hdr_succeeded = true;
```
Consolatis, line 172: I am not sure why this needs to be handled in the first place. If the render format is already set, why do we have to set it again?

Consolatis (commented):
There are some things I don't understand in the PR, mostly related to repeatedly setting the image description and render format for the same output. That would be a stark difference to how wlr_output usually behaves: things are set via output state and committed once. And if the commit doesn't error out it is kept being set like that.

Another thing I am curious about is if we could just do something like this once when enabling the output:
```c
if (rc.target_render_depth != LAB_RENDER_BIT_DEPTH_10 || !output_supports_hdr(output)) {
	return;
}
if (!output_set_render_format(output, output_formats_10bit, ARRAY_SIZE(output_formats_10bit))) {
	wlr_log(WLR_ERROR, "Failed to enable 10 bit render format for %s", output->wlr_output->name);
	output_set_render_format(output, output_formats_8bit, ARRAY_SIZE(output_formats_8bit));
	return;
}
const struct wlr_output_image_description image_desc = {
	.primaries = WLR_COLOR_NAMED_PRIMARIES_BT2020,
	.transfer_function = WLR_COLOR_TRANSFER_FUNCTION_ST2084_PQ,
};
wlr_output_state_set_image_description(&output->pending, &image_desc);

if (!wlr_output_test_state(output->wlr_output, &output->pending)) {
	wlr_log(WLR_ERROR, "Failed to enable HDR for %s", output->wlr_output->name);
	wlr_output_state_set_image_description(&output->pending, NULL);
	output_set_render_format(output, output_formats_8bit, ARRAY_SIZE(output_formats_8bit));
}
```

Some detailed comments below.

src/output-state.c
kode54: If we don't apply this every commit, then the image description is reset, and the color transforms are lost. Blame upstream.

src/output.c
kode54: I forgot that there was a convenient macro already declared for this.

src/output.c
kode54: Will fix.

src/output.c
kode54: Maybe it should simply fall back without setting the render format again.

docs/labwc-config.5.scd
kode54: I'm not sure this is the case now. I'll need to test again.

src/output-state.c
Consolatis: Had a quick look into the wlroots sources and it seems to be preserved in `wlr_output` and additionally in `wlr_drm_connector`. Also asked in the `#wlroots` channel and the answer was also that it only needs to be set when it changes.

kode54: This should be resolved now.

kode54: Hold it, I'll check it out. I swear it results in broken transforms if I don't apply those.

kode54: That's a nope, the colors are super poppy if I remove those.

Consolatis: Hm.. could you test with sway v1.12 to see if it also happens there? I don't have any HDR compatible hardware around so can't test myself.

kode54: We determined it to be related to attempting to commit an image_description without a buffer attached, and changed the related code.

src/output.c
```diff
@@ -371,7 +516,7 @@ configure_new_output(struct output *output)
 		output_enable_adaptive_sync(output, true);
 	}
 
-	output_state_commit(output);
+	output_state_setup_hdr(output, false);
 
 	wlr_output_effective_resolution(wlr_output,
 		&output->usable_area.width, &output->usable_area.height);
```
Consolatis: I think this must be moved after the final `lab_wlr_scene_output_commit()` because otherwise the resolution might be wrong.

Consolatis:
Small nitpick above, other than that this LGTM. We can always simplify it a bit further in the future but for now I think its more important that people can try out HDR.

Consolatis:
Thanks!

PR #3426, opened 2026-03-06
fix name of xdg-dialog-v1 global

xi:
I noticed the following line in the log:

```
[ERROR] [../src/server.c:345] Blocking unknown protocol xdg_wm_dialog_v1
```

If I understand correctly,  `allow_for_sandbox()` has a list of globals. The xdg-dialog-v1 defines a global called `xdg_wm_dialog_v1`, so I think this needs to be changed.


Consolatis:
Indeed, thanks!

PR #3427, opened 2026-03-06
labwc-config.5: document mouse cursor variables

concussious:
Everyone knows that mouse cursors can be configured with XCURSOR_SIZE. What was not obvious is that XCURSOR_THEME is also required. Which, is not obvious how to use without XCURSOR_PATH. Rewrite the manual to make these things explicit. I accidentally fixed a missing period too while here, before I noticed that the project wants these in labwc-config(5) instead of labwc(1).

Thanks @jbeich for teaching me how to fix my tiny mouse cursor!

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -1429,9 +1429,25 @@ situation.
 
 ## ENVIRONMENT VARIABLES
 
-*XCURSOR_THEME* and *XCURSOR_SIZE* are supported to set cursor theme
-and size respectively. The default size is 24. System cursor themes can
-typically be found with a command such as:
+*XCURSOR_PATH*
+	Specify a colon-separated list of paths to look for mouse cursors in.
+	Default
+	~/.local/share/icons:
+	~/.icons:
+	/usr/share/icons:
+	/usr/share/pixmaps:
+	~/.cursors:
+	/usr/share/cursors/xorg-x11:
+	/usr/X11R6/lib/X11/icons:
+
+*XCURSOR_SIZE*
+	Specify an alternative mouse cursor size in pixels. Requires
+	XCURSOR_THEME to be set also. Default 24.
+
+*XCURSOR_THEME*
+	Specify a mouse cursor theme within XCURSOR_PATH.
+
+System cursor themes can typically be found with a command such as:
 
 ```
 find /usr/share/icons/ -type d -name "cursors"
--- a/docs/labwc.1.scd
+++ b/docs/labwc.1.scd
@@ -141,7 +141,7 @@ example: *LABWC_DEBUG_FOO=1 labwc*.
 
 *LABWC_DEBUG_KEY_STATE*
 	Enable logging of press and release events for bound keys (generally
-	key-combinations like *Ctrl-Alt-t*)
+	key-combinations like *Ctrl-Alt-t*).
 
 # SEE ALSO
 


ThomasDickey:
XCursor has a manual page, which you might find helpful.  See [this](https://invisible-island.net/xterm/xcursor/xcursor.html).

concussious:
> XCursor has a manual page, which you might find helpful. See [this](https://invisible-island.net/xterm/xcursor/xcursor.html).

Okay, thank you. We probably want to trim this down, and cross-reference that instead to reduce duplication and documentation proliferation. I'll see what I can come up with.

Consolatis:
Thanks for the PR.

Note that those 3 env vars (+ `XDG_DATA_HOME` and `HOME`) seem to be the only ones supported via [labwc](https://github.com/labwc/labwc/blob/master/src/input/cursor.c#L1570) / [wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/tree/master/xcursor). So adding a link to the XCursor documentation might add more confusion.

docs/labwc-config.5.scd
```diff
@@ -1429,9 +1429,17 @@ situation.
 
 ## ENVIRONMENT VARIABLES
 
-*XCURSOR_THEME* and *XCURSOR_SIZE* are supported to set cursor theme
-and size respectively. The default size is 24. System cursor themes can
-typically be found with a command such as:
+*XCURSOR_PATH*
+	Specify a comma-separated list of paths to look for mouse cursors in.
```
Consolatis: The path is colon-separated rather than comma-separated.

We could also add the default here, should be
`~/.local/share/icons:~/.icons:/usr/share/icons:/usr/share/pixmaps:~/.cursors:/usr/share/cursors/xorg-x11:/usr/X11R6/lib/X11/icons`

concussious: > The path is colon-separated rather than comma-separated.

Oops :P fix incoming, thanks.

> We could also add the default here

That default isn't portable; at least BSD has a different directory structure.

Consolatis: That might be true but its the [default](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xcursor/xcursor.c#L488) in wlroots if no `XCURSOR_PATH` is given.

The last entry in the default is a bit more complex because meson automatically [sets](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/xcursor/meson.build) `ICONDIR` to `get_option('prefix') / get_option('datadir') / 'icons'`.

concussious: Okay, I added in the default path. Downstreams can patch this. I had some trouble with this technically though, I couldn't find any scdoc syntax that will display correctly at different manual sizes.

I try to test everything with at least MANWITDH 80, the reference, and MANWITDH 59, to ensure it looks okay on smaller split screens or for people with older eyes.

docs/labwc-config.5.scd
```diff
@@ -1429,9 +1429,25 @@ situation.
...
+*XCURSOR_PATH*
+	Specify a colon-separated list of paths to look for mouse cursors in.
+	Default
+	~/.cursors:
+	~/.icons:
+	~/.local/share/icons:
+	/usr/share/cursors/xorg-x11:
+	/usr/share/icons:
+	/usr/share/pixmaps:
+	/usr/X11R6/lib/X11/icons:
```
Consolatis, line 1441: The ordering of these paths seems wrong (which is relevant because the cursor theme found first is used). With this fixed this looks good to me.

Consolatis:
Thanks!

concussious:
You are welcome, thank you!

Issue #3425, opened 2026-03-06
core dump on task switcher after resuming machine from suspend


kode54:
### Problem description

I just upgraded my machine to the latest Nvidia beta drivers, `595.45.04-1`, using packages supplied by Peter Jung, on Arch Linux, using linux-zen `6.19.6.zen1-1`.

I am using the wlroots 0.20 chase branch, with my gamma commit, and my HDR patches. wlroots is patched with the hidpi-xprop patch set, up to `0.20.0.20260224.132846-1`, which corresponds to a commit from four days ago. I'll try updating that. And I'll also test the 0.19 branch with labwc-git from the master branch.

The new 595 drivers support a new suspend method of restoring video memory allocations, and also eliminates the suspend and hibernate services it previously used. I suspended my desktop while logged in, then woke it shortly thereafter with the power button.

Discord resumed on the main output with the wrong window scale, so I quit and restarted it. Then I attempted to alt-tab to some other window, which caused labwc to crash back to gdm.

It produced this backtrace:
```
#0  0x000055b842198b25 in init_cycle (server=0x7ffc92470950, filter=...)
    at ../labwc/src/cycle/cycle.c:349
#1  0x000055b842170b35 in cycle_begin (server=<optimized out>,
    direction=LAB_CYCLE_DIR_FORWARD, filter=...) at ../labwc/src/cycle/cycle.c:165
#2  run_action (view=<optimized out>, server=<optimized out>,
    action=0x55b87f3cc4d0, ctx=0x7ffc9246f900) at ../labwc/src/action.c:1183
#3  actions_run (activator=0x0, server=<optimized out>, actions=0x55b87fa2b680,
    cursor_ctx=<optimized out>) at ../labwc/src/action.c:1666
#4  0x000055b8421a06a6 in handle_compositor_keybindings (
    keyboard=keyboard@entry=0x55b881072760, event=event@entry=0x7ffc9246fb50)
    at ../labwc/src/input/keyboard.c:545
#5  0x000055b8421a0b2d in handle_key (listener=0x55b8810727c8,
    data=0x7ffc9246fb50) at ../labwc/src/input/keyboard.c:627
#6  0x00007f8c577b04d0 in wl_signal_emit_mutable (
    signal=signal@entry=0x55b87f20e5e0, data=data@entry=0x7ffc9246fb50)
    at ../wayland-1.24.0/src/wayland-server.c:2369
[...]

PR #3430, opened 2026-03-07
src/cycle: handle views without output

Consolatis:
Fixes: #3425
Reported-By: kode54

---

Completely untested. Used a nested if to make it easier to read.

Edit:
Tested by @kode54 at https://github.com/labwc/labwc/issues/3425#issuecomment-4017778820

### DIFF
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -346,8 +346,10 @@ init_cycle(struct server *server, struct cycle_filter filter)
 
 	struct view *view;
 	for_each_view(view, &server->views, criteria) {
-		if (!(cycle_outputs & view->output->id_bit)) {
-			continue;
+		if (filter.output != CYCLE_OUTPUT_ALL) {
+			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+				continue;
+			}
 		}
 		if (cycle_app_id && strcmp(view->app_id, cycle_app_id) != 0) {
 			continue;


src/cycle/cycle.c
```diff
@@ -346,8 +346,10 @@ init_cycle(struct server *server, struct cycle_filter filter)
 
 	struct view *view;
 	for_each_view(view, &server->views, criteria) {
-		if (!(cycle_outputs & view->output->id_bit)) {
-			continue;
+		if (filter.output != CYCLE_OUTPUT_ALL) {
+			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+				continue;
+			}
```
tokyo4j, line 352: If we use this nested *if*, then `get_outputs_by_filter()` returning a bitmask (`UINT64_MAX` when `<action name="NextWindow" output="all">`) doesn't make sense to me.

For background, I first had `get_outputs_by_filter()` return `struct output *` (`NULL` when `<action name="NextWindow" output="all">`), but I updated it to return a bitmask following the feedback: https://github.com/labwc/labwc/pull/3282#discussion_r2637904813.

I suggest removing the nested if, but I'm open to better idea. I know this code is a bit tricky to read 😅

Consolatis: Indeed, I can remove the `UINT64_MAX` return for `get_outputs_by_filter()`.

> I suggest removing the nested if, but I'm open to better idea. I know this code is a bit tricky to read 😅

Do you suggest to combine the 3 conditions into a single `if` or changing the conditions?

tokyo4j: > Indeed, I can remove the UINT64_MAX return for get_outputs_by_filter().

If we go this route, then there's no point in `get_outputs_by_filter()` returning a bitmask and we can just let it return `struct output *`, like I implemented before https://github.com/labwc/labwc/pull/3282#discussion_r2637904813. Here's the patch for it:

```diff
diff --git a/src/cycle/cycle.c b/src/cycle/cycle.c
index 8d1c7612..74334756 100644
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -264,36 +264,25 @@ get_osd_impl(void)
 	return NULL;
 }
 
-static uint64_t
-get_outputs_by_filter(struct server *server,
+static struct output *
+get_output_by_filter(struct server *server,
 		enum cycle_output_filter output_filter)
 {
-	struct output *output = NULL;
-
 	switch (output_filter) {
 	case CYCLE_OUTPUT_ALL:
-		break;
+		return NULL;
 	case CYCLE_OUTPUT_CURSOR:
-		output = output_nearest_to_cursor(server);
-		break;
+		return output_nearest_to_cursor(server);
 	case CYCLE_OUTPUT_FOCUSED: {
 		struct view *view = server->active_view;
 		if (view && output_is_usable(view->output)) {
-			output = view->output;
+			return view->output;
 		} else {
 			/* Fallback to pointer */
-			output = output_nearest_to_cursor(server);
+			return output_nearest_to_cursor(server);
 		}
-		break;
 	}
 	}
-
-	if (output) {
-		return output->id_bit;
-	} else {
-		/* bitmask for all outputs */
-		return UINT64_MAX;
-	}
 }
 
 static void
@@ -336,8 +325,8 @@ init_cycle(struct server *server, struct cycle_filter filter)
 		criteria |= LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
 	}
 
-	uint64_t cycle_outputs =
-		get_outputs_by_filter(server, filter.output);
+	struct output *cycle_output =
+		get_output_by_filter(server, filter.output);
 
 	const char *cycle_app_id = NULL;
 	if (filter.app_id == CYCLE_APP_ID_CURRENT && server->active_view) {
@@ -347,7 +336,7 @@ init_cycle(struct server *server, struct cycle_filter filter)
 	struct view *view;
 	for_each_view(view, &server->views, criteria) {
 		if (filter.output != CYCLE_OUTPUT_ALL) {
-			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+			if (cycle_output != view->output) {
 				continue;
 			}
 		}
@@ -369,12 +358,14 @@ init_cycle(struct server *server, struct cycle_filter filter)
 
 	if (rc.window_switcher.osd.show) {
 		/* Create OSD */
-		uint64_t osd_outputs = get_outputs_by_filter(server,
+		struct output *output_with_osd = get_output_by_filter(server,
 				rc.window_switcher.osd.output_filter);
 		struct output *output;
 		wl_list_for_each(output, &server->outputs, link) {
-			if (!(osd_outputs & output->id_bit)) {
-				continue;
+			if (rc.window_switcher.osd.output_filter != CYCLE_OUTPUT_ALL) {
+				if (output != output_with_osd) {
+					continue;
+				}
 			}
 			if (!output_is_usable(output)) {
 				continue;

```

</details>

> Do you suggest to combine the 3 conditions into a single if or changing the conditions?

I suggested just having `view->output ||` without adding another if. Like this:


```diff
diff --git a/src/cycle/cycle.c b/src/cycle/cycle.c
index 8d1c7612..e9923717 100644
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -346,10 +346,8 @@ init_cycle(struct server *server, struct cycle_filter filter)
 
 	struct view *view;
 	for_each_view(view, &server->views, criteria) {
-		if (filter.output != CYCLE_OUTPUT_ALL) {
-			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
-				continue;
-			}
+		if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+			continue;
 		}
 		if (cycle_app_id && strcmp(view->app_id, cycle_app_id) != 0) {
 			continue;
```

</details>

Which do you prefer? I don't really have any preference on this.

Consolatis: In your second patch we would skip views without output even if there is no output filter.

tokyo4j: Ah, I thought the parent if-statement is just for readability. Then the second patch should look like this:

```diff
diff --git a/src/cycle/cycle.c b/src/cycle/cycle.c
index 8d1c7612..11fd8428 100644
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -346,10 +346,8 @@ init_cycle(struct server *server, struct cycle_filter filter)
 
 	struct view *view;
 	for_each_view(view, &server->views, criteria) {
-		if (filter.output != CYCLE_OUTPUT_ALL) {
-			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
-				continue;
-			}
+		if (view->output && !(cycle_outputs & view->output->id_bit)) {
+			continue;
 		}
 		if (cycle_app_id && strcmp(view->app_id, cycle_app_id) != 0) {
 			continue;
```

This makes views without output showing on all the outputs even with an output filter, which I think is practically harmless.

Consolatis: That feels wrong to me. I don't really use the output filter but I would expect only windows with that output being shown there if I configure the filter.

Edit:
As for the readability statement in the top post I just meant its more readable than cramming everything into a single `if` like
```c
if (filter.output != CYCLE_OUTPUT_ALL
		&& (!view->output || !(cycle_outputs & view->output->id_bit))) {
	..
}
```

tokyo4j: > I would expect only windows with that output being shown there if I configure the filter.

Note that `view->output` becoming NULL even when there are some outputs is another bug (https://github.com/labwc/labwc/issues/3425#issuecomment-4030737799). So the code path newly covered in this PR should be taken only when there are no connected outputs.

Do you dislike the first patch in https://github.com/labwc/labwc/pull/3430/changes#r2904251421? What I don't like about the current implementation in this PR is that it's now unclear why `get_outputs_by_filter()` returns a bitmask.

Honestly I don't intend to stick with my argument any longer, as the patch is small and totally acceptable.

Consolatis: I am fine with your first patch but my main objective here is to stop labwc crashing while not adding more potential bug surface to a 0.9.6 bugfix release.

tokyo4j: > my main objective here is to stop labwc crashing while not adding more potential bug surface to a 0.9.6 bugfix release.

Ah, I didn't considered that. Fair enough, let's go with the current approach. I can apply the first patch when I touch `cycle.c` again.

Just to note, #3425 is less likely to happen in 0.9.5. Consider two situations below:
1. Running `NextWindow` action while all the outputs are disconnected
2. Running `NextWindow` action after all the outputs are disconnected and then reconnected (e.g. after suspend?)

In the master branch between #3387 and #3433 (both after 0.9.5 bugfix release), the crash could occur in both 1st and 2nd situations. But in 0.9.5 bugfix release, the crash could occur in only the 1st situation (which is still undesirable, of course).

tokyo4j:
Thank you!

PR #3433, opened 2026-03-08
 view: save view->last_placement on initial positioning

### LINKED ISSUE #3431: xdg windows can be misplaced on output layout change
When an xdg window is mapped and then the output layout is changed, the window can go somewhere outside of the output layout, making it invisible to the user.

### Steps to reproduce
1. Start nested labwc with `WLR_WL_OUTPUTS=2 labwc`
2. Spawn a `foot` window on the output "WL-1" (**Do not move this window, keep it where it spawned**)
3. Run `wlr-randr --output WL-1 --custom-mode 1300x800`
4. The `foot` window becomes invisible, with the labwc log: `[../labwc/src/view.c:1759] view has no last placement info`

### Analysis

This is a regression caused by the combination of #3310, #3355 and #3387.

Through #3310 and #3355, I added `view->last_placement` so that we can keep or restore the output-relative position of the view:

https://github.com/labwc/labwc/blob/9de35d7b4fbf95c5ed999bfe61d222022b50550e/include/view.h#L223-L235

`view->last_placement` is updated every time the user moves the view through `view_move_resize()`. I assumed that `view_move_resize()` is also called when a view is mapped so that all the mapped views can be safely relocated by `view_adjust_for_layout_change()`.

However, after #3387, `view_move_resize()` is no longer called when an xdg window is mapped, as the PR replaced `view_place_by_policy()` in the map handler with `view_compute_position_by_policy()`. As a result, `view->last_placement` in xdg views is never set until the user manually moves the view, causing the bug.

I'll work on fixing this and open a PR.

tokyo4j:
Fixes #3431.

The 1st commit replaces `view_place_by_policy()` with `view_compute_position_by_policy()` in `set_initial_position()` in `xwayland.c`, just for consistency with `set_initial_position()` in `xdg.c`. CC @jlindgren90

The 2nd commit fixes #3431 by explicitly saving `view->last_placement` on initial positioning. This ensures all the mapped views have `view->last_placement` set when an output layout change occurs. I applied this to both `xdg.c` and `xwayland.c`, as the 1st commit extends #3431 to x11 views too.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -564,7 +564,15 @@ bool view_titlebar_visible(struct view *view);
 void view_set_ssd_mode(struct view *view, enum lab_ssd_mode mode);
 void view_set_decorations(struct view *view, enum lab_ssd_mode mode, bool force_ssd);
 void view_toggle_fullscreen(struct view *view);
+
+/*
+ * Saves the window position in view->last_placement. This should be called
+ * when a view is first mapped or manually moved by the user.
+ */
+void view_save_last_placement(struct view *view);
+/* Restores and adjusts the view's position from the view->last_placement */
 void view_adjust_for_layout_change(struct view *view);
+
 void view_move_to_edge(struct view *view, enum lab_edge direction, bool snap_to_windows);
 void view_grow_to_edge(struct view *view, enum lab_edge direction);
 void view_shrink_to_edge(struct view *view, enum lab_edge direction);
--- a/src/view.c
+++ b/src/view.c
@@ -568,8 +568,6 @@ view_moved(struct view *view)
 	}
 }
 
-static void save_last_placement(struct view *view);
-
 void
 view_move_resize(struct view *view, struct wlr_box geo)
 {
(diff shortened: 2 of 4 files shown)


jlindgren90:
> The 1st commit replaces view_place_by_policy() with view_compute_position_by_policy() in set_initial_position() in xwayland.c, just for consistency with set_initial_position() in xdg.c.

This part is wrong, I think. Xwayland views are aware of their layout position, thus the view_place_by_policy is needed to send a configure to the server & application.

jlindgren90:
The xdg-shell change looks fine - have not tested.

tokyo4j:
> This part is wrong, I think. Xwayland views are aware of their layout position, thus the view_place_by_policy is needed to send a configure to the server & application.

Thanks. I was missing that the client and compositor send the window position to each other in X11. I also thought `view_move_resize()` is not called when an initially-maximized x11 view is mapped, but it's actually called by `handle_map_request()`.

I removed the first commit. We could remove the call to `view_save_last_placement()` in `xwayland.c`, but I kept it for safety.

jlindgren90:
> I removed the first commit. We could remove the call to view_save_last_placement() in xwayland.c, but I kept it for safety.

I think that's a good idea - there are cases where the xwayland client sets its own geometry and we don't need view_move_resize() at map.

PR #3434, opened 2026-03-08
RFC: use X macros to fill up enum action_type and const char *action_…

domo141:
…names[]

To make it harder to have these 2 data structures go out of sync.

To understand X macros better, visit:
- https://en.wikipedia.org/wiki/X_macro
- https://philliptrudeau.com/blog/x-macro

--

whether this is worthwhile to replace the status-quo (current code) is to be discussed
- I thought of this around 2 years ago while I was doing x-macros work elsewhere, but now this one missing change in one commit made the effort of sending this worthwhile :D

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -68,146 +68,109 @@ struct action_arg_list {
 	struct wl_list value;
 };
 
-/* Warning: Any change in this enum needs to be reflected in the action_names[] array below */
+#define ACTION_TYPE_LIST(X) \
+	X(NONE, "None") \
+	X(CLOSE, "Close") \
+	X(KILL, "Kill") \
+	X(DEBUG, "Debug") \
+	X(EXECUTE, "Execute") \
+	X(EXIT, "Exit") \
+	X(MOVE_TO_EDGE, "MoveToEdge") \
+	X(TOGGLE_SNAP_TO_EDGE, "ToggleSnapToEdge") \
+	X(SNAP_TO_EDGE, "SnapToEdge") \
+	X(GROW_TO_EDGE, "GrowToEdge") \
+	X(SHRINK_TO_EDGE, "ShrinkToEdge") \
+	X(NEXT_WINDOW, "NextWindow") \
+	X(PREVIOUS_WINDOW, "PreviousWindow") \
+	X(RECONFIGURE, "Reconfigure") \
+	X(SHOW_MENU, "ShowMenu") \
+	X(TOGGLE_MAXIMIZE, "ToggleMaximize") \
+	X(MAXIMIZE, "Maximize") \
+	X(UNMAXIMIZE, "UnMaximize") \
+	X(TOGGLE_FULLSCREEN, "ToggleFullscreen") \
+	X(SET_DECORATIONS, "SetDecorations") \
+	X(TOGGLE_DECORATIONS, "ToggleDecorations") \
+	X(TOGGLE_ALWAYS_ON_TOP, "ToggleAlwaysOnTop") \
+	X(TOGGLE_ALWAYS_ON_BOTTOM, "ToggleAlwaysOnBottom") \
+	X(TOGGLE_OMNIPRESENT, "ToggleOmnipresent") \
+	X(FOCUS, "Focus") \
+	X(UNFOCUS, "Unfocus") \
+	X(ICONIFY, "Iconify") \
+	X(RAISE, "Raise") \
+	X(LOWER, "Lower") \
+	X(MOVE, "Move") \
+	X(RESIZE, "Resize") \
+	X(RESIZE_RELATIVE, "ResizeRelative") \
+	X(MOVETO, "MoveTo") \
+	X(RESIZETO, "ResizeTo") \
+	X(MOVETO_CURSOR, "MoveToCursor") \
+	X(MOVE_RELATIVE, "MoveRelative") \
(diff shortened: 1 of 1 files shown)


domo141:
if this flies further, have to put the strings in "quotes" to avoid CamelCase warnings...

jlindgren90:
I've definitely seen code where "tricky" macros like this were over-used, but I think I like the approach here.

For comparison, wlroots and some other places in labwc use designated initializers, like:
`[ACTION_TYPE_SHOW_MENU] = "ShowMenu",`

That has the advantage of being a little more obvious, but doesn't prevent a value from being missing, and is still repetitive.

Consolatis:
I am not aware of any nice solution. I think its always one of these three:

1. risk of desync (two manually synced arrays)
2. repetition in case of enum + `[ENUM_VAR] = "value"`
3. macro to create the two arrays based on an intermediate array, potentially confusing IDEs and feeling slightly out of place

My personal preference for this specific case of a large array would be 1 -> 3 -> 2.
To minimally improve things for 1 we could add an `static_assert()` to ensure that at least the sizes match, even though that would not have caught the recent desync.

domo141:
Looks like we 3 are quite like on the same page on this thing... kinda but maybe...

Whatever happens to this change I'll push the "final" version so that it is not incomplete in history.

I've already thought a followup change which includes ACTION_TABLE_COUNT and ARRAY_SIZE(action_names)
(and removal of that NULL in action_names) which can submitted irrespective of this change.

johanmalm:
How about 3 with a slight tweak to increase readability + a `[ENUM_VAR] = "value"` style designated initialisers to ensure the indices always match the enum values.

```
#define ACTION_TYPE_LIST(X) \
	X(INVALID, "INVALID") \
	X(NONE, "None") \
	X(CLOSE, "Close") \
	X(KILL, "Kill") \
	X(DEBUG, "Debug") \
	X(EXECUTE, "Execute") \
	X(EXIT, "Exit") \
	X(MOVE_TO_EDGE, "MoveToEdge") \
	X(TOGGLE_SNAP_TO_EDGE, "ToggleSnapToEdge") \
	X(SNAP_TO_EDGE, "SnapToEdge") \
	X(GROW_TO_EDGE, "GrowToEdge") \
	X(SHRINK_TO_EDGE, "ShrinkToEdge") \
	X(NEXT_WINDOW, "NextWindow") \
	X(PREVIOUS_WINDOW, "PreviousWindow")
	...etc.
```

Generate the enums.

Note that INVALID might not get zero, so would have to make sure we don't rely on that.

```
/*
 * Will expand to:
 *
 * enum action_type {
 *	ACTION_TYPE_INVALID,
 *	ACTION_TYPE_NONE,
 *	ACTION_TYPE_CLOSE,
 *	ACTION_TYPE_KILL,
 *	etc...
 */
enum action_type {
#define X(name, str) ACTION_TYPE_##name,
	ACTION_TYPE_LIST(X)
#undef X
};
```


Generate the string array:

```
/*
 * Will expand to:
 *
 * const char *action_names[] = {
 *	[ACTION_TYPE_INVALID] = "INVALID",
 *	[ACTION_TYPE_NONE] = "None",
 *	[ACTION_TYPE_CLOSE] = "Close",
 *	[ACTION_TYPE_KILL] = "Kill",
 *	etc...
 */
const char *action_names[] = {
#define X(name, str) [ACTION_TYPE_##name] = str,
	ACTION_TYPE_LIST(X)
#undef X
    NULL
};
```

The X-macro generating both the enum and the string table from that single source of truth certainly has the advantage of eliminating the parallel lists getting out of sync.

With some comments (as above) and the slight refactoring, I think this is readable enough and should make for better maintainability.

johanmalm:
@Consolatis @jlindgren90 - would you support this X macro approach if tweaked as I suggested?

domo141:
This is now the version that is like @johanmalm suggested. I took a few iterations with
a bit different naming and other things to play with (and finally check in generated code
nothing changed (except line numbers in __assert_fail() calls). fun and educational).

In the line 166 had to remove space beween ')' and '[' to "satisfy" checkpatch.pl.
maybe no-one notices...

If one wanted to ensure action_names[0] is populated (and not NULL), static_assert
could be adde to ensure ACTION_TYPE_INVALID (or some other val) is 0. Also, literal
strings could be forced in action_names by adding "" before 'str' in 2nd X definition.
I don't see undetectable risks there so I left such a things out of suggestion here.

Consolatis:
To ensure `ACTION_TYPE_INVALID` is `0` we could simply remove it from the macro array and add it manually as first entry in both arrays, with an additional `= 0` in the enum one.

In general I am fine with the change.

johanmalm:
> To ensure `ACTION_TYPE_INVALID` is `0` we could simply remove it from the macro array and add it manually as first entry in both arrays, with an additional `= 0` in the enum one.

The only place I can see where it potentially makes a difference is `action_type_from_str()` where we iterate from `1`.

I like the idea above. @domo141 - do you agree - and do you fancy implementing that?

> In general I am fine with the change.

Looks like we've got thumbs all around then.

Happy to do final review+test+merge if you take a view (and fix) the `ACTION_TYPE_INVALID=0` thing.

domo141:
agreed & undrafted

johanmalm:
LGTM. Have tested and it seems to work well.

Thanks for this contribution.

Issue #3438, opened 2026-03-10
xwayland initial geometry not always constrained to usable area


jlindgren90:
### Problem description

We have some code to constrain initial view geometry to usable area (for both xdg-shell and xwayland views), but it's applied inconsistently for xwayland views.

Specific issues:
1. When an xwayland view starts maximized, the natural geometry isn't constrained, so the view can end up larger than the usable area after un-maximized.
2. For certain applications (tested `gitk`, also `xterm` with `-geometry` option), the initial geometry isn't constrained at all. It may have to do with these apps setting a position hint.

I noticed the issues when working on my experimental rust port and have fixes there, but need to backport them to C still. Just creating this issue to document the current behavior.

The labwc version tested was from my own `stable` branch, but I'm pretty sure 0.9.5 and master branch both behave the same way.

### Working case: Thunar floating

 - export GDK_BACKEND=x11
 - start labwc nested, resize nested output to large size
 - start Thunar, ensure un-maximized, resize to nearly (nested) output size
 - navigate in and out of a folder to trigger geometry save in Thunar
 - close Thunar
 - resize nested labwc smaller and start Thunar again
 
RESULT: Thunar window is constrained to output size (good)

### Broken case 1: Thunar maximized-then-floating

 - export GDK_BACKEND=x11
 - start labwc nested, resize nested output to large size
 - start Thunar, ensure un-maximized, resize to nearly (nested) output size
 - navigate in and out of a folder to trigger geometry save in Thunar
 - close Thunar
 - start Thunar again and maximize
 - navigate in and out of a folder to trigger geometry save in Thunar
 - close Thunar
 - resize nested labwc smaller and start Thunar again (3rd time)
 - unmaximize Thunar

RESULT: Thunar window is NOT constrained to output size (bad)

### Broken case 2: gitk (doesn't restore maximized state)

 - start labwc nested (nested output size doesn't matter)
 - start gitk and maximize it
 - close gitk
 - start gitk again

RESULT: gitk is NOT constrained to output size and its bottom+right borders are slightly offscreen (bad)

### labwc build source

Local build

### labwc version

labwc 0.9.5-16-gbb99a68b5 (+xwayland +nls +rsvg +libsfdo)

### labwc environment

Nested on Wayland

### Distribution

Arch Linux

johanmalm:
Closed by #3439

PR #3439, opened 2026-03-10
xwayland: rework setting initial geometry from surface

jlindgren90:
The basic idea is to set the initial geometry from the surface exactly once, just before we need it, i.e. either (1) when mapping the view or (2) right before processing an initial maximize/fullscreen request.

I've consolidated various parts of the initial geometry setup to take place at this point (in ensure_initial_geometry_and_output()).

The main motivation is to have a valid, adjusted floating geometry for the view *before* saving the natural geometry when processing an initial maximize/fullscreen request. This reduces code duplication and addresses a FIXME in set_initial_position(), as well as fixing an issue where the natural geometry could exceed the usable output area.

Some other minor changes:

- The initial output is now set directly from the surface geometry if the "application/user-set position" hint is given. This is unlikely to matter in practice, but theoretically an initially maximized view could now appear on a different (application-chosen) output.
- Floating view size is now constrained to the usable area even if a position hint is set. It seemed inconsistent that `xterm -g 200x200` was constrained but `xterm -g 200x200+0+0` was not.

Fixes:
- #3438

Not necessarily recommending this for a patch release as it's a substantial change.

### DIFF
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -39,6 +39,7 @@ struct xwayland_view {
 	struct view base;
 	struct wlr_xwayland_surface *xwayland_surface;
 	bool focused_before_map;
+	bool initial_geometry_set;
 
 	/* Events unique to XWayland views */
 	struct wl_listener associate;
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -211,33 +211,6 @@ top_parent_of(struct view *view)
 	return s;
 }
 
-static void
-ensure_initial_geometry_and_output(struct view *view)
-{
-	if (wlr_box_empty(&view->current)) {
-		struct wlr_xwayland_surface *xwayland_surface =
-			xwayland_surface_from_view(view);
-		view->current.x = xwayland_surface->x;
-		view->current.y = xwayland_surface->y;
-		view->current.width = xwayland_surface->width;
-		view->current.height = xwayland_surface->height;
-		/*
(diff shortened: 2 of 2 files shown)


jlindgren90:
Some potential room for improvement: make the `view_constrain_size_to_that_of_usable_area()` and `view_place_by_policy()` calls only compute geometry and not actually call `view_move_resize()`, if we know the view is going to end up maximized/fullscreen. But that would take a little more work. I think the logic here is behaviorally sound, if not optimal due to the couple of extra Configures being generated.

johanmalm:
LGTM

domo141:
after this change in `urxvt -g -0-0` the geometry position has no effect.

`xterm -g -0-0` works ok. Also `xxdiff 3.x` and `WAYLAND_DISPLAY= emacs`  with position args works as expected.

Did some `printf` debugging:

`printf("xxxxxx %p %d\n", xsurface->size_hints, has_position);`

(added as line 280 in src/xwayland.c)

shows that in case of urxvt, `xsurface->size_hints` is `NULL` and therefore 
`has_position` is always `false` -- with `xterm`, `has_position` is `true` when 
`xterm -g -0-0`  (and `false` when only `xterm`) is executed.

I wonder which thing should be fixed (urxvt, wlroots or labwc) here...

jlindgren90:
> I wonder which thing should be fixed (urxvt, wlroots or labwc) here...

- #3466

PR #3440, opened 2026-03-11
output: exit compositor when last nested output is destroyed

marler8997:
When running on the X11 or Wayland backend, each output corresponds to a window in the parent compositor. Unlike DRM outputs, these cannot be reconnected after being destroyed.

Terminate the compositor when the last output is destroyed on these nested backends, matching the expected behavior when the user closes the window.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -12,6 +12,7 @@
 #include <strings.h>
 #include <wlr/backend/drm.h>
 #include <wlr/backend/wayland.h>
+#include <wlr/backend/x11.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
 #include <wlr/types/wlr_gamma_control_v1.h>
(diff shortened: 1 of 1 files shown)


src/output.c
```diff
@@ -193,13 +194,25 @@ handle_output_destroy(struct wl_listener *listener, void *data)
...
-	output->wlr_output->data = NULL;
+	struct wlr_output *wlr_output = output->wlr_output;
+	wlr_output->data = NULL;
 
 	/*
 	 * output->scene_output (if still around at this point) is
 	 * destroyed automatically when the wlr_output is destroyed
 	 */
 	free(output);
+
```
kode54, line 198: Your pointer below becomes invalid after the above free() statement.

kode54: Never mind, I misread. I thought it was an offset within the structure. My bad.

marler8997: Understandable mistake, it was unnecessarily convoluted.  I've simplified it by just moving the check to occur before the free.

Consolatis:
Thanks!

PR #3442, opened 2026-03-11
src/output.c: guard against wlroots not providing the X11 backend

Consolatis:
Reported-By: Flrian via IRC

----

Confirmed fixing compilation with wlroots without X11 backend.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -12,7 +12,7 @@
 #include <strings.h>
 #include <wlr/backend/drm.h>
 #include <wlr/backend/wayland.h>
-#include <wlr/backend/x11.h>
+#include <wlr/config.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_drm_lease_v1.h>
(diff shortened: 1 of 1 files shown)


src/output.c
```diff
@@ -39,6 +39,10 @@
 #include "view.h"
 #include "xwayland.h"
 
+#if WLR_HAS_X11_BACKEND
+	#include <wlr/backend/x11.h>
```
tokyo4j: We usually don't add an indentation for `#include` guarded by a build config. I don't have strong preferences here, but for now let's remove the tab for consistency.

tokyo4j:

src/output.c
Consolatis: Personally I prefer the indentation here but you are already the second person saying they prefer it without indentation. Will remove.

tokyo4j: Yeah I'm totally OK with the indentation and I will welcome a PR to insert an indentation tree-wide.

PR #3443, opened 2026-03-11
ssd-titlebar: update title buffer when title gets empty

tokyo4j:
Reported by @Flrian via IRC

Before this PR, `ssd_update_title()` returned early and the title was not updated when the title gets empty. Furthermore, this caused the title buffer to be misplaced outside of the titlebar when the window is resized.

Reproducer:
1. Build and launch https://github.com/tokyo4j/wayland-demo/tree/repro-labwc-ssd-title-not-updated
2. One second later, the window title gets empty. But the title in SSD is not updated.
3. Un-minimize the window. Now the title buffer is misplaced outside of the titlebar.

I suggest including this PR in the next bug-fix release.

https://github.com/user-attachments/assets/4da72949-971f-49cf-b88a-2636b645ca5b

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -2501,9 +2501,6 @@ view_destroy(struct view *view)
 	wl_list_remove(&view->set_title.link);
 	wl_list_remove(&view->destroy.link);
 
-	zfree(view->title);
-	zfree(view->app_id);
-
 	if (view->foreign_toplevel) {
 		foreign_toplevel_destroy(view->foreign_toplevel);
 		view->foreign_toplevel = NULL;
@@ -2557,6 +2554,9 @@ view_destroy(struct view *view)
 	assert(wl_list_empty(&view->events.activated.listener_list));
 	assert(wl_list_empty(&view->events.set_icon.listener_list));
 	assert(wl_list_empty(&view->events.destroy.listener_list));
+
+	zfree(view->title);
+	zfree(view->app_id);
 
 	/* Remove view from server->views */
 	wl_list_remove(&view->link);


tokyo4j:
Pushed another commit to clear `view->{title,app_id}` later in `view_destroy()` as suggested by @Consolatis via IRC. It's just for safety and no one has reported any bugs caused by `view->{title,app_id}` getting NULL too early.

Consolatis:
Thanks :)

As for the bugfix release, I agree on including a fix for the issue but maybe we should write a new commit for that which just replaces `string_null_or_empty(view->title)` with `!view->title` to be on the safe side.

If there is some extremely weird edge case in master that is fine and we can fix it once discovered / reported. But doing 2 bugfix releases shortly after a regular release and then introducing new bugs within those feels somewhat unstable for the project.

tokyo4j:
Alright, so I think we can include following commits in the next bugfix release:

- 9de35d7b4fbf95c5ed999bfe61d222022b50550e: fix name of xdg-dialog-v1 global for sandbox permission
- 3bb4ccd22d55a787d4e2288185ae547b1da324a3: just for safety, no need to mention in `NEWS.md`
- 89574772d04d05fcb34bc2223db0093611d9154e: handle outputs with all modes failing
- f0589810ad9227fc4faade31ccecbccb51e7a846: fixes a regression in 0.9.4 that `NextWindow` causes segfault while no outputs connected
- ab60379c7f3e46655f1c81dc76f25adae0b6c6f9: this PR (but replace `assert()` with `if (!view->title) return;`)

CC @johanmalm

Consolatis:
Discussed with malm, I'll create a branch + cherry-pick PR for 0.9.6 tomorrow.

PR #3444, opened 2026-03-12
tree-wide: use die_if_null() for wlr_scene alloc failures

jlindgren90:
wlr_scene_*_create() functions all allocate memory via calloc() and return NULL if the allocation fails. Previously, the failures were handled in any of 3 different ways:

 - sending a wayland protocol error
 - exiting labwc with an error
 - segfault (no NULL check at all)

Since labwc does not attempt to survive heap exhaustion in other allocation paths (such as `znew`), it seems more consistent to use the same die_if_null() check used in those paths to exit with an error.

For the three most common create() functions (tree, rect, buffer), add small lab_wlr_ wrappers to common/scene-helpers.

### DIFF
--- a/include/common/scene-helpers.h
+++ b/include/common/scene-helpers.h
@@ -4,12 +4,20 @@
 
 #include <stdbool.h>
 
+struct wlr_buffer;
 struct wlr_scene_node;
 struct wlr_surface;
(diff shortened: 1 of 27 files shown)


johanmalm:
LGTM

Issue #3441, opened 2026-03-11
I can't ```Alt``` + ```Tab``` out of a Wine/xwayland window


danielfrrrr:
### Problem description

When I start a game through Wine (using `winex11.drv`, i.e., xwayland), I can't use `Alt` + `Tab` to switch to other windows.

https://github.com/user-attachments/assets/b6f36707-70db-4c8b-9d7b-a4ca53d0214a

labwc output log:
[labwc-output-log.txt](https://github.com/user-attachments/files/25893400/labwc-output-log.txt)

SWAT output log:
[swat4-output-log-labwc.txt](https://github.com/user-attachments/files/25893404/swat4-output-log-labwc.txt)

### Steps to reproduce

1. Start a game through Wine (xwayland is used by default, so you don't need to do anything).
2. The game will likely start in full screen mode.
3. Try pressing `Alt` + `Tab` to switch to another window, either with the labwc window switcher or through your taskbar.

### labwc build source

Release

### labwc version

0.9.5

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Arch Linux

danielfikko:
This is a regression, `git bisect` says:

```
079b3b8a10392cc196ab37ddc2a97a5843be08e5 is the first bad commit
commit 079b3b8a10392cc196ab37ddc2a97a5843be08e5 (HEAD)
Author: John Lindgren <john@jlindgren.net>
Date:   Sat Feb 28 16:33:08 2026 -0500

    xwayland: honor client request for always-on-top

 include/xwayland.h |  1 +
 src/xwayland.c     | 15 +++++++++++++++
 2 files changed, 16 insertions(+)
```

https://github.com/labwc/labwc/pull/3406

Consolatis:
So the game sets always-on-top on itself?
In that case you should be able to switch to other windows which are also set to always on top + be able to change workspaces. Potentially you could also toggle (and therefore disable) always-on-top via keybind or window menu (`A-Space` by default) when in the game to see if we are on the right track.

@jlindgren90 (or anyone else) Any idea on how to deal with this situation? My thoughts:
- add some specific setting to ignore / allow all x11 always-on-top requests
- add some window rule, similar to `ignoreFocusRequest` and `ignoreConfigureRequest` which ignores the request
- add some window rule which works the other way around, by default we ignore and if set we allow the request
- just accept the fact that some X11 software behaves.. lets say 'interesting'
- any other ideas?

danielfikko:
[...]

PR #3445, opened 2026-03-12
xwayland: make X11 always-on-top request opt-in

Consolatis:
Fixes:
- #3441

---

To be included in 0.9.6.
CC @jlindgren90

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -431,6 +431,11 @@ handle_request_above(struct wl_listener *listener, void *data)
 		wl_container_of(listener, xwayland_view, request_above);
 	struct view *view = &xwayland_view->base;
 
+	if (window_rules_get_property(view, "allowAlwaysOnTop") != LAB_PROP_TRUE) {
+		wlr_log(WLR_INFO, "X11 client side always on top request rejected");
+		return;
+	}
+
 	view_set_layer(view, xwayland_view->xwayland_surface->above
 		? VIEW_LAYER_ALWAYS_ON_TOP : VIEW_LAYER_NORMAL);
 }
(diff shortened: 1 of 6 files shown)


jlindgren90:
Seems reasonable.

A more targeted workaround for #3441 could be to ignore always-on-top for full screen windows, since the combination of both seems a bit aggressive/unnecessary. Non-fullscreen windows would typically have window controls visible and could be at least minimized or moved in a pinch.

jlindgren90:
Is there a view_set_layer() in handle_map_request() that needs the condition added too?

Consolatis:
> Is there a view_set_layer() in handle_map_request() that needs the condition added too?

Indeed, missed that one.

src/xwayland.c
```diff
@@ -431,6 +431,11 @@ handle_request_above(struct wl_listener *listener, void *data)
 		wl_container_of(listener, xwayland_view, request_above);
 	struct view *view = &xwayland_view->base;
 
+	if (window_rules_get_property(view, "allowAlwaysOnTop") != LAB_PROP_TRUE) {
```
johanmalm, line 434: If this window-rule is not set in `rc.xml`, presumably we get `LAB_PROP_UNSPECIFIED` resulting in refusal to handle the `above` request.

Is that the intended behaviour?

johanmalm:
I think this PR changes the behaviour so that unless there specifically is a window-rule which sets `allowAlwaysOnTop` to `TRUE` then we no longer honour those requests.

Is that correct? And is that what we want?

If I have understood correctly, I suggest stating that in the man page and declaring the default.

I'll update `NEWS.md` with a `### Changed` section for the time being.

johanmalm:
Is this the right summary?

Disallow X11 window always-on-top requests by default to fix an issue whereby Alt+Tab cannot be used to switch to other windows when using some XWayland Wine games [#3441]. Add window-rule property `allowAlwaysOnTop` to optionally enable this always-on-top requests. Add the snippet below to the `rc.xml` file to restore the previous behaviour.

```
<windowRules>
  <windowRule identifier="*" allowAlwaysOnTop="yes" />
</windowRules>
```

src/xwayland.c
Consolatis: Yes, apparently some games (or things like wine) use a-o-t together will fullscreen, presumably to still cover up a potential panel when loosing focus or something like that. Thus it is a pretty annoying feature to enable by default.

Consolatis:
I am fine with keeping this PR open for a little longer to discuss alternative approaches including figuring out if there is something we are missing / doing different to X11 (like not reflecting back the new state to the X11 window). We could revert the a-o-t commit for 0.9.6 instead.

johanmalm:
LGTM

PR #3446, opened 2026-03-12
xwayland: sync always on top state back to X11

Consolatis:
Otherwise things like `wmctrl -b toggle,above` doesn't work.

I've chosen to go with a generic always-on-top signal here rather than to implement a new view_impl function because this signal is also useful for future foreign-toplevel protocols and we don't need two code-paths for the same thing.

For after 0.9.6.

CC @jlindgren90

### DIFF
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -59,6 +59,12 @@ struct xwayland_view {
 	/* Not (yet) implemented */
 /*	struct wl_listener set_role; */
 /*	struct wl_listener set_hints; */
+
+	/* Events coming in from the view itself */
+	struct {
+		struct wl_listener always_on_top;
+	} on_view;
+
 };
 
 void xwayland_unmanaged_create(struct server *server,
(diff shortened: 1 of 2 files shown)


johanmalm:
LGTM.

Is there anything to consider with the a-o-t window-rule we merged today, or is this ready to go?

Consolatis:
After seeing that some games are apparently built on workarounds on top of workarounds, I am honestly not quite sure if a `above` state set by the user on some X11 window via keybind / window menu might end up confusing that client. Although I'd assume the chances are rather minimal because users would not usually set always-on-top on fullscreen games and I don't know of any other X11 clients acting like that.

jlindgren90:
LGTM.

For what it's worth: in Audacious, years ago, I gave on paying attention to the always-on-top state published by the WM in _NET_WM_STATE. There were weird race conditions between user/application-side changes and then getting spurious change notifications from the WM (don't remember which one) undoing what the user just requested.

But it could still be useful for other use-cases, such as `wmctrl` as mentioned.

PR #3447, opened 2026-03-12
Cherry-picks for labwc 0.9.6

Consolatis:
This is a collection of cherry-picks for the upcoming labwc 0.9.6 bugfix release targeted at a new temporary `0-9-5-maintenance` branch. 1383e74cb1036428a22d142352c34e09c89e1c4c was simplified to make it more robust by returning on `!view->title` instead of asserting against it.

I did not add 3bb4ccd22d55a787d4e2288185ae547b1da324a3 because I am not sure if there could be side-effects due to the additional calls. Especially because it has never been tested without 3f223fe5.

Intend to also add #3445 after that PR is reviewed.

CC @jlindgren90 @johanmalm @tokyo4j 
If I missed anything or something isn't right feel free to mention.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog]
 
 | Date       | All Changes   | wlroots version | lines-of-code |
 |------------|---------------|-----------------|---------------|
+| 2026-03-15 | [0.9.6]       | 0.19.2          | 29271         |
 | 2026-03-04 | [0.9.5]       | 0.19.2          | 29251         |
 | 2026-02-27 | [0.9.4]       | 0.19.2          | 29225         |
 | 2025-12-19 | [0.9.3]       | 0.19.2          | 28968         |
@@ -41,7 +42,7 @@ The format is based on [Keep a Changelog]
 | 2021-04-15 | [0.2.0]       | 0.13.0          | 5011          |
 | 2021-03-05 | [0.1.0]       | 0.12.0          | 4627          |
 
-[unreleased]: NEWS.md#unreleased
+[0.9.6]: NEWS.md#096---2026-03-15
 [0.9.5]: NEWS.md#095---2026-03-04
 [0.9.4]: NEWS.md#094---2026-02-27
 [0.9.3]: NEWS.md#093---2025-12-19
@@ -108,9 +109,42 @@ There are some regression warnings worth noting for the switch to wlroots 0.19:
 [wlroots-5098]:https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/5098
 [gtk-8792]: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8792
 
-## unreleased
+## 0.9.6 - 2026-03-15
 
-[unreleased-commits]
+[0.9.6-commits]
+
+This is an earlier-than-usual release containing bug fixes only. It has been
+done on a separate branch (0.9.5-maintenance) to avoid the inclusion of
+refactoring and new features.
+
+```
+               0.9.6  <--- bug-fixes only
+                /
+               /
+0.9.4--------0.9.5--------  <-- master
+```
+
+### Fixed
+
+- Disable outputs where all modes fail [#3428] [#3429] @Consolatis @kode54
+- Fix regression in `0.9.4` that causes `NextWindow` action to segfault when
+  no outputs are connected. This fixes a window-switcher crash with some
+  Nvidia GPUs/drivers after suspend [#3425] [#3430] @Consolatis
+- Fix typo to allow `xdg-dialog-v1` global [#3426] @xi
+
+### Changed
+
+- Disallow X11 window always-on-top requests by default to fix an issue whereby
+  Alt+Tab cannot be used to switch to other windows when using some XWayland
+  Wine games [#3441]. Add window-rule property `allowAlwaysOnTop` to optionally
+  allow this always-on-top requests. Add the snippet below to the `rc.xml` file
+  to restore the previous behaviour. [#3445] @Consolatis
+
+```
+<windowRules>
+  <windowRule identifier="*" allowAlwaysOnTop="yes" />
+</windowRules>
+```
 
 ## 0.9.5 - 2026-03-04
 
@@ -2607,7 +2641,7 @@ Compile with wlroots 0.12.0 and wayland-server >=1.16
   ShowMenu
 
 [Keep a Changelog]: https://keepachangelog.com/en/1.0.0/
-[unreleased-commits]: https://github.com/labwc/labwc/compare/0.9.5...HEAD
+[0.9.6-commits]: https://github.com/labwc/labwc/compare/0.9.5...0.9.6
 [0.9.5-commits]: https://github.com/labwc/labwc/compare/0.9.4...0.9.5
 [0.9.4-commits]: https://github.com/labwc/labwc/compare/0.9.3...0.9.4
 [0.9.3-commits]: https://github.com/labwc/labwc/compare/0.9.2...0.9.3
@@ -3135,8 +3169,18 @@ Compile with wlroots 0.12.0 and wayland-server >=1.16
 [#3365]: https://github.com/labwc/labwc/pull/3365
 [#3372]: https://github.com/labwc/labwc/pull/3372
 [#3373]: https://github.com/labwc/labwc/pull/3373
+[#3387]: https://github.com/labwc/labwc/pull/3387
 [#3400]: https://github.com/labwc/labwc/pull/3400
 [#3406]: https://github.com/labwc/labwc/pull/3406
 [#3410]: https://github.com/labwc/labwc/pull/3410
 [#3411]: https://github.com/labwc/labwc/pull/3411
 [#3412]: https://github.com/labwc/labwc/pull/3412
+[#3425]: https://github.com/labwc/labwc/pull/3425
+[#3426]: https://github.com/labwc/labwc/pull/3426
+[#3428]: https://github.com/labwc/labwc/pull/3428
+[#3429]: https://github.com/labwc/labwc/pull/3429
+[#3430]: https://github.com/labwc/labwc/pull/3430
+[#3440]: https://github.com/labwc/labwc/pull/3440
+[#3441]: https://github.com/labwc/labwc/pull/3441
+[#3443]: https://github.com/labwc/labwc/pull/3443
+[#3445]: https://github.com/labwc/labwc/pull/3445


tokyo4j:
Now we will need to decide what to do with a report in IRC:

<details><summary>IRC log</summary>

```
03:56          monash  a backtrace from 2 days ago: https://dpaste.com/2WMSNTCSS.txt
03:57          monash  were the reasons which triggered #3425 understood?
            labwc_bot  Issue 3425 (closed) - core dump on task switcher after resuming machine from suspend - https://github.com/labwc/labwc/issues/3425
04:06      Consolatis  interesting backtrace, that doesn't look like the same issue
04:08          monash  yes the fix fixed the issue, but i coulnt readily see how that situation could have occured
04:09      Consolatis  hm.. it seems like an assert for a title buffer width width 0
04:10      Consolatis  with width 0*
04:11      Consolatis  can you open an issue with that stacktrace?
04:12      Consolatis  including what version of labwc (labwc -v) this was happening with
04:13          monash  it was probably from march 6 sources
               monash  i'll try to open an issue later
               monash  but the other issue - how could a view suddenly lose an output?
04:25      Consolatis  they will loose the output when switching ttys and potentially when the monitor or system suspends
04:28          monash  hmm
04:37      Consolatis  what wlroots version are you using?
04:38          monash  wlroots refs/tags/0.19.2
04:40      Consolatis  hm.. should be fine. really interesting stacktrace
           Consolatis  looking a bit into it and trying to reproduce
04:41      Consolatis  are you still able to reproduce the issue with current master?
04:42          monash  i'll have to try later..
04:43          monash  bbl
                   --  -monash
04:44      Consolatis  ^ it seems to be based on an empty title and us creating a font buffer with width 0 but still getting a wlr_buffer back which doesn't have buffer->data set and triggers an assert
           Consolatis  so its indirectly caused by the change to allow empty titles
04:45      Consolatis  but the buffer creation should fail in that width == 0 case.. so not sure what is going on
05:13      Consolatis  i really don't understand the backtrace. it doesn't look like its possible to happen unless I missed something
05:55            malm  Yes, weird
                 malm  Where does the buffer creation fail on width==0?
                 malm  Maybe we need to protect against width==0 in scaled_font_buffer.c
05:56      Consolatis  common/font.c font_buffer_create() has a string_null_or_empty() check
05:57            malm  What about when we do scaled_font_buffer_update()?
                 malm  It calls font_get_buffer_size()
                 malm  line 136
                 malm  ...which sets self->width (where self is the scaled_font_buffer
05:58            malm  )
           Consolatis  right, but that one is not part of the stacktrace
           Consolatis  at the end it calls scaled_buffer_request_update()
                 malm  At #14
            labwc_bot  Issue 14 (closed) - Grab openbox rc.xml file if it exists. - https://github.com/labwc/labwc/issues/14
05:59            malm  #14 0x0000000000466bbc in scaled_font_buffer_update
                 malm      (self=0x772c20, text=0x1a5f870 "", max_width=696, font=0x5e8280 <rc+192>, color=0x7ffc10db443c, bg_color=0x7ffc10db3e50) at ../src/scaled-buffer/scaled-font-buffer.c:139
06:00      Consolatis  yes, it calls font_get_buffer_size() but I don't see it having an impact
06:02            malm  Ok
           Consolatis  I think the issue is here: #12 0x0000000000467faa in _update_buffer (self=0x18cd4d0, scale=1) at ../src/scaled-buffer/scaled-buffer.c:141
            labwc_bot  Issue 12 (closed) - fluxbox - https://github.com/labwc/labwc/issues/12
           Consolatis  that one calls wlr_scene_buffer_set_buffer() with some buffer which apparently has buffer->data == NULL
06:03            malm  Ok
                 malm  Any idea where data becomes NULL?
                 malm  I haven't followed every frame
06:04            malm  Guess both width and stride will be zero, so feels like that might lead to a NULL somewhere
           Consolatis  that is the thing.. I don't think it becomes NULL anywhere. the only reason I saw which would cause it to be NULL is when creating the buffer
           Consolatis  which should not be possible because font_buffer_create() prevents creating a buffer for an empty string
06:05            malm  But font_buffer_create() is not in stacktrace
06:06      Consolatis  src/scaled-buffer/scaled-buffer.c:105 calls into src/scaled-buffer/scaled-font-buffer.c:15 which calls into font_buffer_create()
                 malm  k
           Consolatis  so unless we somehow end up reusing a cached buffer that should be the source of the invalid buffer being set
                 malm  It's like there's a missing piece
06:07      Consolatis  it might also be random stack trashing.. although that somehow seems unlikely because of the specific empty title + labwc change to allow for that
06:08            malm  But like you say, font_buffer_create() protects against null-or-empty
           Consolatis  and I can't replicate the issue with the master branch at all. buffer creation fails as expected for me when setting an empty title via  printf "\x1b]0;%s\x07" ""
06:09            malm  Which terminal did they use?
06:10      Consolatis  shouldn't matter, don't even know if its a terminal at all. its just some window with empty title
06:12            malm  Is monash the only one who can reproduce?
           Consolatis  i don't know if they can reproduce it either
           Consolatis  but yes, didn't see any other report about it
06:19      Consolatis  my assumption that its actually an empty title is based on this arguments of the backtrace btw: text=0x1a5f870 ""
06:20      Consolatis  + view_set_title (view=0x138bfa0, title=0x12e14d0 "")
                 malm  Yes.
```

</details>

This is the provided stacktrace:

<details><summary>stacktrace</summary>

```
#0  __pthread_kill_implementation
    (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at pthread_kill.c:44
#1  0x00007fb40b0b5c9f in __pthread_kill_internal (signo=6, threadid=<optimized out>)
    at pthread_kill.c:78
#2  0x00007fb40b069a52 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fb40b0534df in __GI_abort () at abort.c:79
#4  0x00007fb40b053415 in __assert_fail_base
    (fmt=0x7fb40b1c0b98 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x540130 "buffer->data", file=file@entry=0x540120 "../src/buffer.c", line=line@entry=58, function=function@entry=0x540280 <__PRETTY_FUNCTION__.3> "data_buffer_begin_data_ptr_access") at assert.c:92
#5  0x00007fb40b062722 in __assert_fail
    (assertion=0x540130 "buffer->data", file=0x540120 "../src/buffer.c", line=58, function=0x540280 <__PRETTY_FUNCTION__.3> "data_buffer_begin_data_ptr_access") at assert.c:101
#6  0x000000000041bcde in data_buffer_begin_data_ptr_access
    (wlr_buffer=0x1a24720, flags=1, data=0x7ffc10db3bb8, format=0x7ffc10db3bb4, stride=0x7ffc10db3ba8) at ../src/buffer.c:58
#7  0x00000000004c8c4a in wlr_buffer_begin_data_ptr_access
    (buffer=0x1a24720, flags=1, data=0x7ffc10db3bb8, format=0x7ffc10db3bb4, stride=0x7ffc10db3ba8) at ../subprojects/wlroots/types/buffer/buffer.c:89
#8  0x00000000004c8d8f in wlr_buffer_is_opaque (buffer=0x1a24720)
    at ../subprojects/wlroots/types/buffer/buffer.c:120
#9  0x00000000004b230c in scene_buffer_set_buffer
    (scene_buffer=0x15963a0, buffer=0x1a24720)
    at ../subprojects/wlroots/types/scene/wlr_scene.c:785
#10 0x00000000004b2833 in wlr_scene_buffer_set_buffer_with_options
    (scene_buffer=0x15963a0, buffer=0x1a24720, options=0x7ffc10db3cc0)
    at ../subprojects/wlroots/types/scene/wlr_scene.c:900
#11 0x00000000004b2f87 in wlr_scene_buffer_set_buffer
    (scene_buffer=0x15963a0, buffer=0x1a24720)
    at ../subprojects/wlroots/types/scene/wlr_scene.c:1012
#12 0x0000000000467faa in _update_buffer (self=0x18cd4d0, scale=1)
    at ../src/scaled-buffer/scaled-buffer.c:141
#13 0x0000000000468494 in scaled_buffer_request_update
    (self=0x18cd4d0, width=0, height=26) at ../src/scaled-buffer/scaled-buffer.c:252
#14 0x0000000000466bbc in scaled_font_buffer_update
    (self=0x772c20, text=0x1a5f870 "‎", max_width=696, font=0x5e8280 <rc+192>, color=0x7ffc10db443c, bg_color=0x7ffc10db3e50) at ../src/scaled-buffer/scaled-font-buffer.c:139
#15 0x000000000046b129 in ssd_update_title (ssd=0x13a7860)
    at ../src/ssd/ssd-titlebar.c:475
#16 0x0000000000437b2e in view_set_title (view=0x138bfa0, title=0x12e14d0 "‎")
    at ../src/view.c:2301
#17 0x000000000043da8f in handle_set_title (listener=0x138c1f8, data=0x0)
    at ../src/xwayland.c:514
#18 0x00007fb40c13383c in wl_signal_emit_mutable () at /usr/lib64/libwayland-server.so.0
#19 0x000000000050a753 in read_surface_title
    (xwm=0x135c4f0, xsurface=0x17f1020, property=255, reply=0x134d240)
    at ../subprojects/wlroots/xwayland/xwm.c:780
#20 0x000000000050b462 in read_surface_property
    (xwm=0x135c4f0, xsurface=0x17f1020, property=255, reply=0x134d240)
    at ../subprojects/wlroots/xwayland/xwm.c:1078
#21 0x000000000050c32a in xwm_handle_property_notify (xwm=0x135c4f0, ev=0x18a7570)
    at ../subprojects/wlroots/xwayland/xwm.c:1414
#22 0x000000000050d9e0 in read_x11_events (xwm=0x135c4f0)
    at ../subprojects/wlroots/xwayland/xwm.c:1996
#23 0x000000000050dac6 in x11_event_handler (fd=74, mask=1, data=0x135c4f0)
    at ../subprojects/wlroots/xwayland/xwm.c:2028
#24 0x00007fb40c135802 in wl_event_loop_dispatch () at /usr/lib64/libwayland-server.so.0
#25 0x00007fb40c1332e5 in wl_display_run () at /usr/lib64/libwayland-server.so.0
#26 0x0000000000422aca in main (argc=1, argv=0x7ffc10db57c8) at ../src/main.c:274
(gdb) 
```

</details>

In short, ab60379c7f3e46655f1c81dc76f25adae0b6c6f9 (allow an empty title in ssd titlebar rather than keeping the stale title) somehow caused `lab_data_buffer.data` being unset which caused segfault in `data_buffer_begin_data_ptr_access()` when a client sets an empty title.

We haven't managed to reproduce the issue and we are not sure if the reporter comes back to provide more information before the end of Sunday.

I think possible options are:

1. Delay the 0.9.6 release date by one week or so to see if the reporter comes back, we can reproduce the issue, or someone else reports the same issue
2. Not include ab60379c7f3e46655f1c81dc76f25adae0b6c6f9 in 0.9.6.
3. Replace ab60379c7f3e46655f1c81dc76f25adae0b6c6f9 with a different approach (e.g. replace an emtpy string with an space in the SSD title)

My preference is (2.). This means the issue described in #3443 happens in 0.9.6, but I think it's acceptable as that bug has existed for years (I reproduced the issue in labwc 0.8.0) but no one other than @Flrian has reported it.

johanmalm:
I think `2` makes sense. It feels like the lowest risk option for `0.9.6`

Consolatis:
Agree, I also prefer `2`. Hard to trigger visual bug vs potential  trigger of an `assert()`.

Edit:
Removed the commit. Also added the opt-in for client side always-on-top which messed up a-tab out of fullscreen games.

johanmalm:
> Edit: Removed the commit. Also added the opt-in for client side always-on-top which messed up a-tab out of fullscreen games.

I'm confused by the `xwayland: make X11 always-on-top request opt-in` commit. It does not appear to be in `master`. Should we add it there first?

johanmalm:
Okay, I've got it now. Suggest we review/finish/merge #3445 on `master` before including here.

Consolatis:
I could also replace it with a plain revert of the client side always-on-top support commit of 0.9.5, up to you.

PR #3448, opened 2026-03-12
NEWS.md: update notes for 0.9.6

johanmalm:
In readiness for Sunday.

Appreciate early review.

Intend to merge against `master` rather than `0.9.5-maintenance`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog]
 
 | Date       | All Changes   | wlroots version | lines-of-code |
 |------------|---------------|-----------------|---------------|
+| 2026-03-15 | [unreleased]  | 0.19.2          | 29244         |
+| 2026-03-15 | [0.9.6]       | 0.19.2          | 29271         |
 | 2026-03-04 | [0.9.5]       | 0.19.2          | 29251         |
 | 2026-02-27 | [0.9.4]       | 0.19.2          | 29225         |
 | 2025-12-19 | [0.9.3]       | 0.19.2          | 28968         |
@@ -42,6 +44,7 @@ The format is based on [Keep a Changelog]
 | 2021-03-05 | [0.1.0]       | 0.12.0          | 4627          |
 
 [unreleased]: NEWS.md#unreleased
+[0.9.6]: NEWS.md#096---2026-03-15
 [0.9.5]: NEWS.md#095---2026-03-04
 [0.9.4]: NEWS.md#094---2026-02-27
 [0.9.3]: NEWS.md#093---2025-12-19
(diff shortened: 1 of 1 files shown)


johanmalm:
I did not mention the one below as I assume it just for people compiling rather than something that is visible to users in the more general sense.

#3442 `src/output.c: guard against wlroots not providing the X11 backend`

Consolatis:
#3442 is not part of the cherry-picks in #3447 because the thing it fixes (terminating when the last nested output is destroyed) is not part of it. Agree on not including it in the unreleased section on top either.

NEWS.md
```diff
@@ -112,6 +115,44 @@ There are some regression warnings worth noting for the switch to wlroots 0.19:
...
+- Update titlebar title when set to empty and fix an associated issue causing
+  the title to be misplaced outside of the titlebar when the window is resized.
+  [#3443] @tokyo4j
+- Handle outputs better when all modes fail because Wayland-backend nested
+  outputs can still be configured. [#3428] [#3429] @Consolatis @kode54
+- Fix regression in `0.9.4` that causes `NextWindow` action to segfault when
+  no outputs are connected. This fixes a window-switcher crash with some
+  Nvidia GPUs/drivers after suspend [#3425] [#3430] @Consolatis
+- Fix regression introduced in 3f223fe whereby output layout changes may
+  move windows outside of the output layout. [#3433] @tokyo4j
```
Consolatis: I don't think the change causing the issue is part of 0.9.5 so the fix is not part of the cherry-picks.

NEWS.md
```diff
@@ -112,6 +115,44 @@ There are some regression warnings worth noting for the switch to wlroots 0.19:
...
+0.9.4--------0.9.5--------  <-- master
+```
+
+### Fixed
+
+- Update titlebar title when set to empty and fix an associated issue causing
+  the title to be misplaced outside of the titlebar when the window is resized.
+  [#3443] @tokyo4j
+- Handle outputs better when all modes fail because Wayland-backend nested
+  outputs can still be configured. [#3428] [#3429] @Consolatis @kode54
```
Consolatis: "Disable outputs where all modes fail" might be enough here.

NEWS.md
johanmalm: Thanks. Have removed all together as it fixes a problem introduced after `0.9.5`

NEWS.md
johanmalm: Okay. Thanks. Had misunderstood that one slightly. Have taken your suggested words.

johanmalm:
Have squashed to it's ready to go from my perspective - awaiting only a conclusion of the always-on-top discussion.

Consolatis:
LGTM. Do you want me to cherry-pick this commit into #3447 ?

NEWS.md
```diff
@@ -112,6 +115,55 @@ There are some regression warnings worth noting for the switch to wlroots 0.19:
...
+               /
+0.9.4--------0.9.5--------  <-- master
+```
+
+### Fixed
+
+- Disable outputs where all modes fail [#3428] [#3429] @Consolatis @kode54
+- Fix regression in `0.9.4` that causes `NextWindow` action to segfault when
+  no outputs are connected. This fixes a window-switcher crash with some
+  Nvidia GPUs/drivers after suspend [#3425] [#3430] @Consolatis
```
tokyo4j, line 150: A retrospective correction: the second sentence is incorrect as #3425 depends on #3387 which hasn't released yet. Sorry for missing this.

johanmalm: I'm confused by this. #3430 is merged and claims to fix #3425.

Please could you help me understand which words need to move to the `unreleased` section.

Should it be like this?

```diff
diff --git a/NEWS.md b/NEWS.md
index 3699c61c..c1d66fb7 100644
--- a/NEWS.md
+++ b/NEWS.md
@@ -126,6 +126,8 @@ There are some regression warnings worth noting for the switch to wlroots 0.19:
   destroyed. [#3440] @marler8997
 - Allow policy-based placement to apply when an initially-maximized/fullscreen
   view is restored to floating geometry. [#3387] @jlindgren90
+- Fixes window-switcher crash with some Nvidia GPUs/drivers after suspend.
+  [#3425] [#3430] @Consolatis
 
 ## 0.9.6 - 2026-03-15
 
@@ -146,8 +148,7 @@ refactoring and new features.
 
 - Disable outputs where all modes fail [#3428] [#3429] @Consolatis @kode54
 - Fix regression in `0.9.4` that causes `NextWindow` action to segfault when
-  no outputs are connected. This fixes a window-switcher crash with some
-  Nvidia GPUs/drivers after suspend [#3425] [#3430] @Consolatis
+  no outputs are connected. [#3425] [#3430] @Consolatis
 - Fix typo to allow `xdg-dialog-v1` global [#3426] @xi
 
 ### Changed
```

tokyo4j: I think we don't need to add anything to the `unreleased` seciton. Removing the second sentence should be enough.

See https://github.com/labwc/labwc/pull/3430#discussion_r2911781637 for background. I put it into a table below:

| version     | crashes with no outputs? | crashes after suspend on Nvidia GPUs (#3425)? |
| ----------- | ------------------------------- | -------------------------------------------------------- |
| 0.9.5       | yes                             | no                                                       |
| after #3387 | yes                             | yes                                                      |
| after #3430 | no                              | no                                                       |
| 0.9.6       | no                              | no                                                       |

PR #3450, opened 2026-03-14
xwayland: fix missing title and icon with override-redirect toggle

jlindgren90:
When an X11 client clears override-redirect on a window (i.e. it becomes "managed") without an unmap/map cycle, wlroots doesn't re-emit the set_title/class/icon events. We need to synthesize them to re-read the values from the surface.

From what I can tell, only older/misbehaving X11 clients would actually do this (toggle override-redirect without unmap/map), so this is perhaps a bit pedantic, but an easy fix anyway.

Test case (AI-assisted): https://gist.github.com/jlindgren90/712931190aaa027c8a7d60a5f65ce60a

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -1083,6 +1083,16 @@ xwayland_view_create(struct server *server,
 
 	if (xsurface->surface) {
 		handle_associate(&xwayland_view->associate, NULL);
+		/*
+		 * If a surface is already associated, then we've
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM.

Is there a real case situation that you're aware of where this comes into play?

jlindgren90:
No, just testing the case for completeness.

PR #3451, opened 2026-03-15
labnag.c: initialize sigset_t mask

johanmalm:


Consolatis:
LGTM. Just curious, did you see any weird behavior without it?

johanmalm:
> LGTM.

Thanks.

> Just curious, did you see any weird behavior without it?

I did with https://github.com/johanmalm/t2play/ which is based on `labnag`.

PR #3454, opened 2026-03-17
NEWS.md: fix 0.9.6 release notes

johanmalm:
...by removing the comment about crashes after suspend on some Nvidia GPUs because this only happens after #3387 has is not yet in any release.

Ref:
- https://github.com/labwc/labwc/pull/3430#discussion_r2911781637
- https://github.com/labwc/labwc/pull/3448#discussion_r2948247530

Helped-by: @tokyo4j


tokyo4j:
Thanks!

PR #3457, opened 2026-03-22
tree-wide: use rc.theme instead of server.theme

jlindgren90:
Having two global pointers to the same struct is redundant.


Consolatis:
Thanks

PR #3463, opened 2026-03-27
[Also proof of concept] Beveled Borders

hak-foo:
This riffs on https://github.com/labwc/labwc/pull/3382/changes and with huge props to jlindgren90

It adds two new theme options:

border.beveled: yes
border.bevel_width:2

If border.beveled is set, it renders classic 3-D borders around the windows, menus, and OSD displays.

These are sized based on the relevant border size configuration variables.

The window borders have inner and outer bevels around a flat centre, and the width of these if border.bevel_width.

The overall effect is very much intended to be  "we have MWM/FVWM at home".

The nice thing about a beveled borders as a design concept  is that it scales effortlessly, rather than having to redesign an image-based theme if you change the border or titlebar size.
demo


Consolatis:
Gotta say this looks pretty amazing!

From the implementation point of view I wonder if we can create the individual pieces once when (re)loading the theme and then only wrap them via `wlr_scene_buffer_create()` + `wlr_scene_buffer_set_dest_size()` for all the users. That approach could then also potentially be used for other simple textured borders which can be repeated in a given direction.

It would also open up optionally loading those textures from disk so they could be customized further outside of labwc, similar to the button icons. Themes could ship with those textures and we wouldn't need any pixel / cairo logic for them in labwc itself.

CC @jlindgren90

Jackojc:
That looks really nice. Any chance you could give a run down of your setup here in the show & tell discussions channel?

hak-foo:
> From the implementation point of view I wonder if we can create the individual pieces once when (re)loading the theme and then only wrap them via `wlr_scene_buffer_create()` + `wlr_scene_buffer_set_dest_size()` for all the users. That approach could then also potentially be used for other simple textured borders which can be repeated in a given direction.

I've tried to extend it in that direction.  There's now a central function that generates the border buffers, and caches the actual buffer images, and a central function that will draw them all them into a wlr_scene_tree (everywhere except the workplace switcher OSD world, which is doing its own thing with Cairo, so I used Cairo to render the border buffers)
 
> It would also open up optionally loading those textures from disk so they could be customized further outside of labwc, similar to the button icons. Themes could ship with those textures and we wouldn't need any pixel / cairo logic for them in labwc itself.

I tried to build it as "load on demand/cache" model.  Right now, we cache on image ID (currently the base colour) + border size + border type (single/double/flat/inset bevel).

This should be amenable to adding a new border type, where the ID is, for example, a lookup on a table of filenames to load or paths to slice up.

Consolatis:
I like the centralization. The current state still creates new wlr_buffers per instance, rough sketch to re-use the buffers is here: https://github.com/hak-foo/labwc/compare/master...Consolatis:labwc:tmp/pr_3463_beveled_borders.

Reusing the wlr_buffers (rather than just the pixels themselves) has the benefit that wlroots could theoretically reuse their GPU texture rather than uploading it again on each creation of a wlr_scene_buffer. Wlroots does not currently do that but will eventually at some point so we should design our buffers in a way that can benefit from wlroots starting to do so.

I also added the `NEAREST` filter because otherwise the alpha would fade out from the center by wlroots:
<details><summary>without WLR_SCALE_FILTER_NEAREST</summary>
20260329_14h32m09s_grim
</details>
<details><summary>with WLR_SCALE_FILTER_NEAREST</summary>
20260329_14h32m48s_grim
</details>

johanmalm:
@hak-foo Welcome :smile:

This PR seems to have attracted significant attention.

My TODO list has for some time contained an entry to implement borders (incl. bevels) for titlebar buttons. I have not  yet got around to look at it, but therefore read this with extra interest.

I have not yet studied the code, but would like to just throw in one consideration at an early stage: Is it possible to use the Openbox theme syntax for this to increase compatibility with Openbox themes?

I suspect it would be possible to support the Openbox syntax for things like `menu.title.bg`, `menu.item.bg` and `window.(in)active.button.*.bg`.

Regarding OSD borders, Openbox configuration seems to have some semantic overlap with `osd.border.width`, `osd.border.color`, `osd.bg` - so beveled borders would require a decision here in terms of syntactic approach.

For window borders, it looks like Openbox only supports `window.(in)active.border.color` and `border.width` (i.e. no texture).

For clarify, all the options above with `.bg` refer to an Openbox texture (not to be confused with "GPU texture"). The Openbox textures determine the visual look of elements. They contain a `border` setting which can be any of `Flat`, `Raised` and `Sunken`. Weirdly, `Flat` means "no border". `Raised` and `Sunken` have two bevel options available to them.

Ref:
- https://openbox.org/help/Themes#Textures
- https://labwc.github.io/labwc-theme.5.html#entry_textures
- https://wiki.xfce.org/howto/xfwm4_theme
- https://wiki.xfce.org/howto/xfwm4_theme#titlebar_decorations

hak-foo:
> I like the centralization. The current state still creates new wlr_buffers per instance, rough sketch to re-use the buffers is here: [hak-foo/labwc@master...Consolatis:labwc:tmp/pr_3463_beveled_borders](https://github.com/hak-foo/labwc/compare/master...Consolatis:labwc:tmp/pr_3463_beveled_borders).

Pulled in those changes.  Part of the reason I was using the raw pixel arrays as the lowest common denominator is because the workplace switcher was using them in a different way (as Cairo fills), but it turns out we can extract that from the lab_data_buffer instead.

chocolatemintychip:
Looks awesome! Just curious, is it feasible to add another border under the titlebar? I feel like the abrupt transition between solid colors when there's beveled elements present could look a bit jarring with some themes.

hak-foo:
> @hak-foo Welcome 😄
> 
> This PR seems to have attracted significant attention.
> 
> My TODO list has for some time contained an entry to implement borders (incl. bevels) for titlebar buttons. I have not yet got around to look at it, but therefore read this with extra interest.
> 
> I have not yet studied the code, but would like to just throw in one consideration at an early stage: Is it possible to use the Openbox theme syntax for this to increase compatibility with Openbox themes?


I've tried to surface configurability where it currently supports the beveled frames:

Since there's no "texture" on Openbox borders, I figure we can add this green-field

window.inactive.border.type: raised | sunken | doubleraised | doublesunken | flat
window.inactive.border.bevel-width: 4
window.active.border.type: raised | sunken | doubleraised | doublesunken | flat
window.active.border.bevel-width: 2

I tried to use osd and menu item bg similar to Openbox

osd.bg.type: raised | sunken | doubleraised | doublesunken | flat
osd.bg.bevel-width:1

menu.title.bg:raised | sunken | doubleraised | doublesunken | flat
menu.title.bg.bevel-width:1
menu.items.bg:raised | sunken | doubleraised | doublesunken | flat
menu.items.bg.bevel-width:1
menu.items.active.bg:raised | sunken | doubleraised | doublesunken | flat
menu.items.active.bg.bevel-width:2

and since there isn't a control for the menu border itself, it becomes menu.bg

menu.bg:raised | sunken | doubleraised | doublesunken | flat
menu.bg.bevel-width:3

I probably went overkill with offering the "bevel-width" option everywhere-- it might be reasonable to just default it to, say, 1/3 or 1/2 of the border width for the "doubleraised" and "doublesunken" formats, but it gives a lot of configurability.

hak-foo:
Still adding some extensions-- the centre part of the title bar is now bevelable.

The "exclusive" attribute indicates whether the bevel should cover just the area in between the side buttons, or the whole title bar.  This gives you some interesting style options, especially if you're using image buttons.

window.active.title.bg: raised | sunken | doubleraised | doublesunken | flat  (seems to coexist with gradients)
window.active.title.bg.width:4
window.active.title.bg.bevel-width:2
window.active.title.bg.exclusive:yes

window.inactive.title.bg: raised | sunken | doubleraised | doublesunken | flat
window.inactive.title.bg.width:6
window.inactive.title.bg.bevel-width:3
window.inactive.title.bg.exclusive: no

hak-foo:
And finally, buttons can be beveled.  There's also a bevel type of "none" (skips rendering the beveled border at all) to differentiate from "flat".  This becomes relevant for the title bar and buttons, where there may not be a flat colour "underneath" if you don't add anything "on top".

window.inactive.button.bg: raised | sunken | doubleraised | doublesunken | flat | none
window.inactive.button.bg.width:4
window.inactive.button.bg.bevel-width:2
window.inactive.button.bg.border-color: red
window.inactive.button.bg.border-hover-color: blue

window.active.button.bg:  raised | sunken | doubleraised | doublesunken | flat | none
window.active.button.bg.width:4
window.active.button.bg.bevel-width:2
window.active.button.bg.border-color: red
window.active.button.bg.border-hover-color: blue

Consolatis:
I am currently playing around with a generalized textured rectangle approach inspired by your code here https://github.com/labwc/labwc/compare/master...Consolatis:labwc:feature/textured_corners. My initial basic idea was to have a single 1 row pattern of a few pixels and then use the wlroots scene graph to rotate and scale it for each edge + create a single corner and then also rotate that one for each corner. So the whole rect consists of only 2 buffers which can be re-used across different scene rects. It also allows splitting the pattern renderer from the rectangle implementation. However, after looking a bit further into your PR it seems we actually need differently colored edges in some of the styles so that would rule out my whole approach of re-using the same buffer for each edge.

I see you handled the codestyle warnings. This PR is described as proof  of concept. Do you want us to look into merging it in labwc? In general the idea of supporting textured borders seems fine to me, the question is just how we can sensibly split and encapsulate the different parts involved to reduce future maintenance workload and if its ensured that there is no performance and memory impact of the feature if it isn't used by the user.

hak-foo:
If you're comfortable with advancing it forward, go ahead!  I feel like I've now added bevel support to most things that could conceivably take it. :)

When I first submitted it with that title, obviously it was a lot less complete.  But I figure "proof of concept" still works as a mild warning. I'm new to the code base and may not have perfect alignment with the project's goals and norms, so approach the changes with a healthy level of skepticism.

Actually, let me try to add the highlight/shadw parameters... the drawing code can use them internally now, but they need to actually be plumbed into each call.

hak-foo:
Added the highlight/shadow options similar to as described in the Openbox docs:
```
window.active.title.bg.highlight:128
window.active.title.bg.shadow:64

window.active.border.highlight:128
window.active.border.shadow:64

window.inactive.title.bg.highlight:128
window.inactive.title.bg.shadow:64

window.inactive.border.highlight:128
window.inactive.border.shadow:64

window.inactive.button.bg.highlight:128
window.inactive.button.bg.shadow:64

menu.bg.highlight:128
menu.bg.shadow:64

menu.items.bg.highlight:128
menu.items.bg.shadow:64

menu.items.active.bg.highlight:128
menu.items.active.bg.shadow:64


menu.title.bg.highlight:128
menu.title.bg.shadow:64

osd.bg.highlight:128
osd.bg.shadow:64
```

NOW I'm out of bevel-related ideas :D

hak-foo:
Added some docs for the new config fields.

johanmalm:
Hi there. I just wanted to get back to you with a holding position on this one.

There seems to be a lot of interest in this PR and we want to carry on reviewing with a view to merge. However, there has been a lot of traffic recently and we are currently prioritising some other PRs (e.g. HDR support and menu accelerators) for the cool-down (8 May) for the next release. Therefore, it may be a couple of weeks or more until we have time to get stuck into this one again. Hope that's okay.

snow32a:
Would be nice if it supported Openbox's bevel properties

PR #3464, opened 2026-03-27
xwayland: shuffle map_request/commit events a bit

kode54:
Only register map_request event when surface is associated, and only register commit event when surface is mapped. Unregister both in their matched disassociate and unmap functions.

Fixes a crash launching Steam when combined with the chase/0.20 branch that does not occur on wlroots 0.19 for some reason.

Fixes behavior triggered by 9903331995b4de7c99fb2c40c90ea9980ae70ad7, but not that commit is not likely to be the actual cause, but rather the map_request events triggered by it somehow.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -407,7 +407,6 @@ handle_destroy(struct wl_listener *listener, void *data)
 	wl_list_remove(&xwayland_view->set_strut_partial.link);
 	wl_list_remove(&xwayland_view->set_window_type.link);
 	wl_list_remove(&xwayland_view->focus_in.link);
-	wl_list_remove(&xwayland_view->map_request.link);
 
 	wl_list_remove(&xwayland_view->on_view.always_on_top.link);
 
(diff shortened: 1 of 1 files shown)


Consolatis:
Thoughts @jlindgren90?

jlindgren90:
I'd like to know more details on the crash, and understand what changed on the wlroots side.

domo141:
FYI: I cherry-picked this change, and with this `urxvt -g -0-0` works (again) as expected

I explained the problem with `urxvt -g ...` in the comments of PR #3439 yesterday;
the commit 9903331995b4de (final commit in that PR). 
Interestingly `xterm -g -0-0` was not affected by that change in 990333...

EDIT: Fedora 43, wlroots-0.19.2-1.fc43.x86_64, rxvt-unicode-9.31-13.fc43.x86_64

jlindgren90:
I think the issue with `urxvt` is that wlroots doesn't read the X11 window properties until after associate. If map_request happens first, then we won't see e.g. xsurface->size_hints yet.

Handling map_request seemed to make sense at one point, but learning more of the nitty-gritty of how wlroots handles this stuff, I'm not sure it does any more. Maybe the current map_request logic should move to the associate handler instead. Need to think about this a little more.

jlindgren90:
Please try:
- #3466

jlindgren90:
(and if this still crashes with wlroots 0.20, please post backtrace)

kode54:
Works with wlroots 0.20 and Steam as well.

kode54:
I guess I can close this.

kode54:
Crap. I applied my own PR instead of yours. One second.

kode54:
Yeah, it works too. Closing this finally.

PR #3466, opened 2026-03-28
xwayland: set initial view state + geometry in handle_associate()

jlindgren90:
Empirically, the associate event is always seen just after map_request but before surface map. Window properties are also read by wlroots just before emitting associate. So after some trial and error, this seems to be the best place to set initial view states and compute initial placement.

Fixes initial positioning of: `urxvt -g -0-0`

urxvt placement regressed in: 9903331995b4de7c99fb2c40c90ea9980ae70ad7
("xwayland: rework setting initial geometry from surface")

map_request handler was added ~2 years ago in: c9018da4c4d9539341e4592caa48e94a5e47c6ea
("xwayland: set initial geometry in map_request handler")

I'm not sure if the map_request handler was incorrect from day one, or if something changed in wlroots and/or Xwayland since then.

### DIFF
--- a/include/xwayland.h
+++ b/include/xwayland.h
@@ -54,7 +54,6 @@ struct xwayland_view {
 	struct wl_listener set_strut_partial;
 	struct wl_listener set_window_type;
 	struct wl_listener focus_in;
-	struct wl_listener map_request;
 
 	/* Not (yet) implemented */
 /*	struct wl_listener set_role; */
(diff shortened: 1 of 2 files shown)


kode54:
Okay, this fixes my crash on wlroots 0.20 + steam as well.

johanmalm:
LGTM.

@Consolatis It feels to me like we ought to merge this before the `chase/0.20` branch given @kode54's comment. Thoughts/concerns?

Consolatis:
No concerns with any changes in the master branch in regards to chase/0.20. The chase PR is pretty minimal this time around and should be easy to rebase.

johanmalm:
Thanks.

Issue #3468, opened 2026-03-29
Android/bionic portability: shm_open and librt


GlassOnTin:
### Context

We're cross-compiling labwc for Android (bionic libc) as part of an effort to embed a Wayland compositor in [Haven](https://github.com/GlassOnTin/Haven), an Android SSH/VNC client. The compositor would run natively in the app process with GPU-accelerated rendering via AHardwareBuffer + SurfaceControl (building on the work in [wlroots-android-bridge](https://github.com/Xtr126/wlroots-android-bridge)).

labwc's direct dependencies (wlroots, wayland, xkbcommon, pixman, libdrm) all cross-compile for Android arm64-v8a with NDK r28. wlroots needs two small patches for bionic compatibility — these are in the wlroots layer, not labwc, but are worth documenting here since labwc is the compositor being targeted.

### wlroots patches needed for Android

**1. `meson.build`** — bionic doesn't have a separate `librt` (rt functions are in libc):
```diff
-rt = cc.find_library('rt')
+rt = cc.find_library('rt', required: host_machine.system() != 'android')
```

**2. `util/shm.c`** — bionic lacks `shm_open`/`shm_unlink`. Replacement uses `memfd_create` syscall (available on Android API 26+) with temp file fallback.

### labwc-specific question

labwc has hard dependencies on glib, cairo, pangocairo, libinput, libpng, and libxml2. For an Android headless compositor:

- **libinput**: the headless backend doesn't use physical input devices (input comes via the Android AIDL bridge). Could `libinput` be made optional for builds that only use headless/custom backends? Even a compile-time `#ifdef` guard around the libinput code paths would help.

- **glib/cairo/pango**: these are used throughout for config parsing and server-side decorations. We'll cross-compile the full stack, but is there any appetite for making SSDs optional at build time for embedded/headless use cases?

### Current status

We have a working cross-compilation pipeline producing static libraries for the full wlroots chain. Extending it to labwc's additional dependencies (glib, cairo, pango, freetype, fontconfig, harfbuzz, fribidi) is the next step. These all support Meson/CMake cross-compilation.

Tracking progress at GlassOnTin/Haven#47. Happy to share the build scripts and patches.

Consolatis:
Nice work!

[...]

PR #3470, opened 2026-03-29
Make wlroots backends optional (DRM, session, libinput)

Consolatis:
There are cases where not all wlroots backends are required or are not supported. An example is the termux environment on Android which patches an own Android backend into wlroots and #3468. Other use-cases include usage of labwc in nested-only settings like only with the X11 or Wayland backends or completely headless, e.g. as a wayvnc-only host.

Currently labwc just fails to link in this cases as can be seen when using the reproduction steps below on the master branch.

Making the backends optional also eases porting labwc to some of the BSDs (or termux) which do not necessarily support them.

I didn't try to build it without any libdrm headers installed so this might miss things like the legacy `wlr_drm`. Further testing in environments completely lacking any of the headers and libs appreciated.

Can be tested with something like
```sh
git fetch origin pull/3470/head && git checkout FETCH_HEAD
meson setup build_tmp -Dwlroots:backends= -Dwlroots:session=disabled --force-fallback-for=wlroots
meson compile -C build_tmp
```

Supersedes #2907

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,6 +210,19 @@ jobs:
             meson compile -C build-gcc-no-feature
           ' | $TARGET
 
+      # No backend build, run on Arch only
+      - name: Build with gcc - no-backends test
+        if: matrix.name == 'Arch'
+        run: |
+          echo '
+            cd "$GITHUB_WORKSPACE"
+            export CC=gcc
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM. Not tested.

johanmalm:
Looks like this one has a merge conflict now.

PR #3473, opened 2026-03-30
Make labnag a meson feature flag

GlassOnTin:
Add a `labnag` meson option (type: feature, default: auto) to allow disabling the labnag notification daemon at build time.

### Motivation

As discussed in #3468, labnag uses `shm_open()` in `pool-buffer.c` which isn't available on Android (bionic). Rather than patching the shm implementation, making labnag optional is cleaner — embedded/headless deployments don't need it.

This also benefits other use cases like wayvnc-only hosts or nested-only settings where labnag isn't relevant.

### Usage

```sh
meson setup build -Dlabnag=disabled
```

Default is `auto` so existing builds are unaffected.

### Changes

- `meson_options.txt`: add `labnag` feature option
- `clients/meson.build`: wrap labnag build in `if get_option('labnag').allowed()`

Ref: #3468

### DIFF
--- a/clients/meson.build
+++ b/clients/meson.build
@@ -1,59 +1,61 @@
-wayland_client = dependency('wayland-client')
-wayland_cursor = dependency('wayland-cursor')
+if get_option('labnag').allowed()
+  wayland_client = dependency('wayland-client')
+  wayland_cursor = dependency('wayland-cursor')
 
-nag_sources = files(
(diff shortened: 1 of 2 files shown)


Consolatis:
Thanks, LGTM. Up to @johanmalm to decide on feature itself.

johanmalm:
Thanks.

PR #3474, opened 2026-03-30
keyboard: use XKB_CONTEXT_NO_SECURE_GETENV on Android

GlassOnTin:
On Android (bionic libc), `secure_getenv()` always returns NULL because app processes don't have `AT_SECURE` set. This prevents xkbcommon from reading `XKB_DEFAULT_LAYOUT` and other environment variables when resolving keyboard layouts via `set_layout()`.

Use `XKB_CONTEXT_NO_SECURE_GETENV` on Android so xkbcommon falls back to regular `getenv()`, which works correctly.

Guarded with `#ifdef __ANDROID__` so non-Android builds are unaffected.

Ref: #3468

### DIFF
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -745,7 +745,18 @@ set_layout(struct wlr_keyboard *kb)
 	static bool fallback_mode;
 
 	struct xkb_rule_names rules = { 0 };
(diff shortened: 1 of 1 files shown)


Consolatis:
Thanks, LGTM.

PR #3480, opened 2026-03-30
clients/labnag.c: return zero on get_text_size() error

johanmalm:
Ref: https://github.com/swaywm/sway/pull/9070


Consolatis:
:)

PR #3481, opened 2026-03-30
server: allow external renderer and allocator on Android

GlassOnTin:
On Android, `wlr_renderer_autocreate()` and `wlr_allocator_autocreate()` fail because there is no DRM subsystem. Android apps create EGL contexts via `eglGetDisplay(EGL_DEFAULT_DISPLAY)` and provide GPU buffers via `AHardwareBuffer`.

This adds hooks guarded by `#ifdef __ANDROID__` that check for externally provided renderer/allocator globals before falling back to autocreate. The JNI bridge can then:

1. Create an EGL context via Android's native EGL
2. Wrap it via `wlr_egl_create_with_context()`
3. Create a GLES2 renderer via `wlr_gles2_renderer_create()`
4. Provide a custom allocator (e.g. AHardwareBuffer-based)

The globals (`android_renderer`, `android_allocator`) are declared `extern` and only referenced when `__ANDROID__` is defined — no impact on non-Android builds.

Tested on Pixel 8 Pro: EGL 1.5, Mali-G715, OpenGL ES 3.2. The GLES2 renderer creates successfully through this path.

Context: https://github.com/GlassOnTin/Haven/issues/47

Follows on from #3473 and #3474 (both merged).

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -502,7 +502,17 @@ server_init(void)
 	 * The renderer is responsible for defining the various pixel formats it
 	 * supports for shared memory, this configures that for clients.
 	 */
-	server.renderer = wlr_renderer_autocreate(server.backend);
+#ifdef __ANDROID__
+	/* Android: accept externally created renderer (e.g. GLES2 via
+	 * eglGetDisplay + wlr_egl_create_with_context) */
+	extern struct wlr_renderer *android_renderer;
+	if (android_renderer) {
+		server.renderer = android_renderer;
+	} else
+#endif
+	{
+		server.renderer = wlr_renderer_autocreate(server.backend);
+	}
 	if (!server.renderer) {
 		wlr_log(WLR_ERROR, "unable to create renderer");
 		exit(EXIT_FAILURE);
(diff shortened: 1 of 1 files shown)


Consolatis:
I would prefer if this is instead implemented / patched into wlroots so `wlr_renderer_autocreate()` and `wlr_allocator_autocreate()` work as usual.

GlassOnTin:
@Consolatis Makes sense — the renderer and allocator creation is wlroots' domain. We've been carrying these patches in our wlroots fork ([GlassOnTin/wlroots@android-bionic](https://github.com/GlassOnTin/wlroots/tree/android-bionic)) and the labwc hooks were a pragmatic shortcut.

The wlroots changes needed for `wlr_renderer_autocreate()` to work on Android:

1. **EGL platform init** (`render/egl.c`) — skip `EGL_EXT_platform_base` check when `EGL_KHR_platform_android` is available, use `eglGetDisplay(EGL_DEFAULT_DISPLAY)`
2. **DMA-BUF import** (`render/gles2/renderer.c`) — allow `EGL_ANDROID_image_native_buffer` as alternative to `EGL_EXT_image_dma_buf_import`
3. **Format advertisement** (`render/egl.c`) — populate render format sets from Android-supported formats when DMA-BUF extension is absent
4. **GBM dependency** (`render/meson.build`) — optional on Android

For `wlr_allocator_autocreate()`, it currently fails because it checks the renderer's DRM FD (always -1 on Android). An AHardwareBuffer allocator would be the proper upstream solution.

I'll close this PR. The proper fix is upstream in wlroots — we can file those patches there once we have them stable. For now they live in our fork.

PR #3482, opened 2026-03-31
Drop wl_drm

johanmalm:
Xwayland + VA-API + amdvlk now all support linux-dmabuf-v1.

Ref:
- https://github.com/swaywm/sway/pull/8247

Note for @Consolatis: Possibly linked to #3470

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -10,7 +10,6 @@
 #include <wlr/types/wlr_alpha_modifier_v1.h>
 #include <wlr/types/wlr_data_control_v1.h>
 #include <wlr/types/wlr_data_device.h>
-#include <wlr/types/wlr_drm.h>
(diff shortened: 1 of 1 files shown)


xi:
`wl_drm` should probably also be removed from the list in `allow_for_sandbox()`, right?

Consolatis:
In general I am pretty hesitant to unnecessarily remove support for older protocols, especially if it doesn't come with a noticeable reduction in code complexity. Are there any reasons for removing support other than "sway did it"?

johanmalm:
> In general I am pretty hesitant to unnecessarily remove support for older protocols, especially if it doesn't come with a noticeable reduction in code complexity.

Fair.

> Are there any reasons for removing support other than "sway did it"?

No

Okay, I'll close this and cross it off the chase-discussion.

PR #3483, opened 2026-04-01
Bump the protocol versions

danielfrrrr:
I should mention that I’m not a programmer, so I don’t fully understand the technical implications of these changes. If it’s necessary to write C code to support these new versions, I’m afraid I won’t be able to do that. It would be easier for someone else to handle this PR.


johanmalm:
Thanks. I want to go through what each bump means and take reasonable steps to research that we've got it implemented. This might take a few days.

danielfikko:
> Thanks. I want to go through what each bump means and take reasonable steps to research that we've got it implemented. This might take a few days.

Ok! I should have imagined that it wouldn't be as simple as just bumping up the numbers 😅. Feel free to close this or take ownership of this PR.

PR #3484, opened 2026-04-01
config: migrate from amixer to pactl

danielfikko:
As suggested in https://github.com/labwc/labwc/discussions/3478. Another alternative to `amixer`/`pactl` would be `wpctl`, from WirePlumber, but `pactl` probably covers a wider range of users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,9 +217,9 @@ If you have not created an rc.xml config file, default bindings will be:
 | `super`-`mouse-right`    | resize window
 | `super`-`arrow`          | resize window to fill half the output
 | `alt`-`space`            | show the window menu
-| `XF86AudioLowerVolume`   | amixer sset Master 5%-
-| `XF86AudioRaiseVolume`   | amixer sset Master 5%+
-| `XF86AudioMute`          | amixer sset Master toggle
+| `XF86AudioLowerVolume`   | pactl set-sink-volume @DEFAULT_SINK@ -5%
+| `XF86AudioRaiseVolume`   | pactl set-sink-volume @DEFAULT_SINK@ +5%
+| `XF86AudioMute`          | pactl set-sink-mute @DEFAULT_SINK@ toggle
 | `XF86MonBrightnessUp`    | brightnessctl set +10%
 | `XF86MonBrightnessDown`  | brightnessctl set 10%-
 
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -845,7 +845,7 @@ References:
   A-Space - show window menu
 ```
 
-	Audio and MonBrightness keys are also bound to amixer and
+	Audio and MonBrightness keys are also bound to pactl and
 	brightnessctl, respectively.
 
 *<keyboard><repeatRate>*
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -292,13 +292,13 @@
       <action name="ShowMenu" menu="client-menu" atCursor="no" />
     </keybind>
     <keybind key="XF86AudioLowerVolume">
-      <action name="Execute" command="amixer sset Master 5%-" />
+      <action name="Execute" command="pactl set-sink-volume @DEFAULT_SINK@ -5%" />
     </keybind>
     <keybind key="XF86AudioRaiseVolume">
-      <action name="Execute" command="amixer sset Master 5%+" />
+      <action name="Execute" command="pactl set-sink-volume @DEFAULT_SINK@ +5%" />
     </keybind>
     <keybind key="XF86AudioMute">
-      <action name="Execute" command="amixer sset Master toggle" />
+      <action name="Execute" command="pactl set-sink-mute @DEFAULT_SINK@ toggle" />
     </keybind>
     <keybind key="XF86MonBrightnessUp">
       <action name="Execute" command="brightnessctl set +10%" />
--- a/include/config/default-bindings.h
+++ b/include/config/default-bindings.h
@@ -88,21 +88,21 @@ static struct key_combos {
 		.action = "Execute",
 		.attributes[0] = {
 			.name = "command",
-			.value = "amixer sset Master 5%-",
+			.value = "pactl set-sink-volume @DEFAULT_SINK@ -5%",
 		},
 	}, {
 		.binding = "XF86AudioRaiseVolume",
 		.action = "Execute",
 		.attributes[0] = {
 			.name = "command",
-			.value = "amixer sset Master 5%+",
+			.value = "pactl set-sink-volume @DEFAULT_SINK@ +5%",
 		},
 	}, {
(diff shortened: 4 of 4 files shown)


johanmalm:
LGTM. Would you mind verifying that you have successfully tested it?

danielfikko:
> LGTM. Would you mind verifying that you have successfully tested it?

My keyboard doesn't have those keys, but I tested replacing them with keys I do have (for example, replacing `XF86AudioMute` with `W-0`), and all the commands worked correctly.

Now, when I searched the `libxkbcommon` source code for the key names, I noticed that the names are different from how labwc currently uses them: there is no underline. I pushed a change that removes the underline to match [1](https://github.com/search?q=repo%3Axkbcommon%2Flibxkbcommon%20AudioLowerVolume%20AudioRaiseVolume%20AudioMute&type=code). Is this correct? ~~If necessary, I can open a PR for other keys as well (such as `XF86_MonBrightnessUp` and `XF86_MonBrightnessDown`)~~ Done in https://github.com/labwc/labwc/pull/3488.

davidphilipbarr:
I've been usuing pactl for a while on various thinkpads and surface devices and i've never had any issues, it's been much more reliable than amixer at changing the correct volume level.

Consolatis:
I am a bit skeptical about the change from `amixer` to `pactl` or `wpctl`. People might not necessarily use pulseaudio / wireplumber. `amixer` seems like a good common denominator. And users can change it to whatever they prefer / their system supports.

Regarding the keysym names.. nice catch! Does this mean these default keybinds never worked? I don't have any volume or brightness keys either so can't really test without dusting off my laptop.

danielfikko:
> Does this mean these default keybinds never worked? I don't have any volume or brightness keys either so can't really test without dusting off my laptop.

I can't say for sure either. According to https://github.com/labwc/labwc/pull/3488#issuecomment-4180310738, it used to work, so it was probably just a false positive on my part.

stefonarch:
> Regarding the keysym names.. nice catch! Does this mean these default keybinds never worked? I don't have any volume or brightness keys either so can't really test without dusting off my laptop.

They always worked (LXQt uses it's custom command)
```xml
...
    <keybind key="XF86_AudioLowerVolume">
      <action name="Execute" command="lxqt-qdbus volume down" />
    </keybind>
    
```

johanmalm:
I have tested this PR and it works well for me. Ironically the `master` branch `amixer` command do not currently work for me, so this actually fixes an issue (for me) and avoids having to over-write the default keybinds :smile:

> I am a bit skeptical about the change from `amixer` to `pactl` or `wpctl`. People might not necessarily use pulseaudio / wireplumber. `amixer` seems like a good common denominator. And users can change it to whatever they prefer / their system supports.

I favor swapping to `pactl`, but a unanimous position is more important to me.

Reasons for swapping - from my perspective:

1. `sway` swapped. I find their core devs to have good taste and staying in keeping with `sway` has proven useful in the past.
2. `amixer` doesn't always work for me depending on setup (USB-hub connected, etc), whereas `pactl` does. I think my ability to change volume up/down via amixer has got worse over the last few months (on Arch Linux) and my guess is that relates to conflict between alsa/pa/pipewire. That's not an area I know much about though, so it is just a complete guess.
3. I'm not sure this is a valid reason, but when asking AI to "compare amixer with pactl" I get the following which weigh heavily towards `pactl`

```
Amixer and pactl are command-line tools for Linux sound management.

Amixer directly controls the ALSA hardware layer (low-level), making it ideal for base system audio and limiting volume to 100%.

Pactl controls PulseAudio or PipeWire (high-level), allowing application-specific volume control, over-amplification, and hot-plugging management. 

Key Comparisons:

- Layer: Amixer interacts with hardware (ALSA), whereas pactl acts as a client for sound servers (PulseAudio/PipeWire).
- Capabilities: Pactl can manage per-application audio, sink/source volumes, and device switching. Amixer controls raw hardware channels (e.g., Master, PCM).
- Volume Limits: Amixer usually caps at 100% (safer), while pactl allows volume levels to exceed 100% which can cause distortion.
- Usage: For modern desktop systems with PulseAudio/PipeWire, pactl is generally preferred for desktop hotkeys and scripts to avoid direct hardware conflicts.
- Syntax:
    Amixer: amixer sset Master 5%+
    Pactl: pactl set-sink-volume @DEFAULT_SINK@ +5% 

When to use which:
- Use pactl/pamixer for daily desktop use, hotkeys, and per-app control.
- Use amixer for system-level scripts, minimal environments without a sound server, or if needing to directly unmute hardware channels.

On most modern Linux distributions (Ubuntu, Fedora, Arch), you should use the pactl man page as your primary reference, as using ALSA tools on a system with PulseAudio/PipeWire can sometimes lead to desynced audio channels or broken state.
```

Consolatis:
Moved `XF86_` comment to https://github.com/labwc/labwc/pull/3488#issuecomment-4183383962

Consolatis:
At least on my system (using pulseaudio) `amixer` seems to change volumes of pulse audio rather than the hardware:
```
$ amixer info
Card default 'pulse'/'PulseAudio'
  Mixer name	: 'PulseAudio'
  Components	: ''
  Controls      : 4
  Simple ctrls  : 2
```

I assume its the same for `amixer` on pipewire and `pacmd` on pipewire because from what I remember all of these handle the user side of alsa. And pipewire additionally also the user side of pulse so their respective configuration utils keep working. Similar for applications only supporting alsa output, it is transparently handled by either pulse or pipewire.

Edit:
So
- `amixer` should work on pure ALSA and likely `JACK` systems + pulseaudio + pipewire
- `pactl` should work on pulseaudio + pipewire (with pipewire-pulse installed)
- `wpctl` should work on pipewire with wireplumber

danielfikko:
One advantage of `pactl` is that, at least for Arch Linux users, it’s more probable to work out of the box, since the `libpulse` package is about 37% more popular than the `alsa-utils` package.  [1]

[1] https://pkgstats.archlinux.de/packages?limit=25&offset=0&compare=alsa-utils,libpulse,wireplumber

Consolatis:
My personal preference would still be `amixer` but fine with changing the default to `pactl` if that is what others prefer.

johanmalm:
> My personal preference would still be `amixer` but fine with changing the default to `pactl` if that is what others prefer.

@jlindgren90 @tokyo4j - would you mind offering a view on this?

I don't really feel strongly either way. I prefer moving to `pactl` for consistency with sway + the fact that I think it's more likely to work out of the boxes on more (probably almost all) systems. On the flip-side, there is something about not changing defaults for backward compatibility.

Thoughts?

jlindgren90:
I don't really have a preference. I use a custom script personally. pactl is probably best for mainstream distros, and maybe technically-inclined people who have a reason to use amixer instead would also have no problem with overriding the default.

PR #3488, opened 2026-04-02
config: remove the underline from the XF86 keys

danielfrrrr:
...to match the actual key names in https://github.com/xkbcommon/libxkbcommon/blob/48500df7160d0ec6278dfcec0665a7907cfdf302/include/xkbcommon/xkbcommon-keysyms.h#L2583

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,8 +220,8 @@ If you have not created an rc.xml config file, default bindings will be:
 | `XF86_AudioLowerVolume`  | amixer sset Master 5%-
 | `XF86_AudioRaiseVolume`  | amixer sset Master 5%+
 | `XF86_AudioMute`         | amixer sset Master toggle
-| `XF86_MonBrightnessUp`   | brightnessctl set +10%
-| `XF86_MonBrightnessDown` | brightnessctl set 10%-
+| `XF86MonBrightnessUp`    | brightnessctl set +10%
+| `XF86MonBrightnessDown`  | brightnessctl set 10%-
 
 A root-menu can be opened by clicking on the desktop.
 
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -300,10 +300,10 @@
     <keybind key="XF86_AudioMute">
       <action name="Execute" command="amixer sset Master toggle" />
     </keybind>
-    <keybind key="XF86_MonBrightnessUp">
+    <keybind key="XF86MonBrightnessUp">
       <action name="Execute" command="brightnessctl set +10%" />
     </keybind>
-    <keybind key="XF86_MonBrightnessDown">
+    <keybind key="XF86MonBrightnessDown">
       <action name="Execute" command="brightnessctl set 10%-" />
     </keybind>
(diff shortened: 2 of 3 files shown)


johanmalm:
I'll do some testing with this tomorrow. I use these keys for brightness up/down and have not had a problem.

Can you confirm that it works on your machine with this change - and that it doesn't without it?

danielfikko:
> I'll do some testing with this tomorrow. I use these keys for brightness up/down and have not had a problem.
> 
> Can you confirm that it works on your machine with this change - and that it doesn't without it?

I can't test it because my keyboard doesn't have any of the XF86 keys. I'll have to wait for you to test it tomorrow or for someone who has those keys to test it.

danielfikko:
Sorry for the bother. It was already working with (or without) the underline, it was just a kind of false positive on my part.

https://github.com/labwc/labwc/pull/3484#issuecomment-4183064015

Consolatis:
Requested the keysyms for some variants and at first  it didn't make any sense to me:
```
00:00:00.712 [../src/server.c:680] Got keysm for XF86AudioLowerVolume: 269025041
00:00:00.712 [../src/server.c:680] Got keysm for XF86_AudioLowerVolume: 269025041
00:00:00.713 [../src/server.c:680] Got keysm for XF86___________________AudioLowerVolume: 269025041
00:00:00.713 [../src/server.c:680] Got keysm for _XF86AudioLowerVolume: 0
00:00:00.713 [../src/server.c:680] Got keysm for XF8_6AudioLowerVolume: 0
00:00:00.713 [../src/server.c:680] Got keysm for XF86Audio_LowerVolume: 0
00:00:00.713 [../src/server.c:680] Got keysm for XF86AudioLower_Volume: 0
00:00:00.713 [../src/server.c:680] Got keysm for XF86AudioLowerVolume_: 0
```

But finally found this: https://github.com/xkbcommon/libxkbcommon/blob/master/src/keysym.c#L339
So libxkbcommon actually handles a specific edge case for the `XF86_` prefix.

As our default `XF86` keysym names are objectively different to the header and we don't need to make libxkbcommon run into the fallback my vote would be to changing them to the non-underscore variants. Existing setups should keep working because both result in the same sym which we use to detect duplicated keybinds.

johanmalm:
Thanks @Consolatis.

Interestingly, the defines in `/usr/include/X11/XF86keysym.h` contain the underscore.

johanmalm:
https://cgit.freedesktop.org/xorg/proto/x11proto/tree/XF86keysym.h

Consolatis:
That one seems to use a different prefix though, e.g. `XF86XK_AudioLowerVolume`

danielfikko:
So, what should I do? I made changes to https://github.com/danielfrrrr/labwc/tree/xf86 to include the audio keys, but GitHub won't let me reopen that PR.

Consolatis:
> but GitHub won't let me reopen that PR.

Same here, just says something about "branch was force-pushed or recreated" but no indicator as to why that is a problem. Feel free to just open a new PR.

Issue #3489, opened 2026-04-02
windowSwitcher/osd with thumbnail style causes labwc crash


elviosak:
### Problem description

It happened with multiple different games with wine/proton/linux native either wayland and xwayland and has been happening for a few months (not sure how long, but more than 3 months), currently i'm testing with [Beyond All Reason](https://www.beyondallreason.info/download#Download-Play), the AppImage version.

The crash happens when i'm in a match (game has a launcher, then a "Lobby/Menu" before starting a match) and Alt+Tab when there's a Browser window (Firefox/Floorp/Helium) open, doesn't seem to happen with Brave nor other applications. Also doesn't happen with OSD style "classic".

Might be relevant:
- Desktop with old CPU/GPU: i5-3450/GTX 1060 (nvidia-580xx drivers)
- 2 Monitors side-by-side
- Arch Linux

config: (i leave the "fields" there to easily switch between classic and thumbnail)
```
  <windowSwitcher preview="yes" outlines="yes" unshade="yes">
    <osd show="yes" style="thumbnail" output="all" thumbnailLabelFormat="%b %I | %T" />
    <fields>
      <field content="icon" width="3%" />
      <field content="custom" format="%b %I - %T" width="97%"/>
    </fields>
  </windowSwitcher>

...

    <keybind key="A-Tab">
      <action name="NextWindow"/>
    </keybind>
```

Got 2 logs with `labwc -d > lab.log 2>&1` and attached here, not sure if there's anything relevant aside from the last 2 lines: 
```
00:06:21.136 [ERROR] [render/allocator/gbm.c:118] gbm_bo_create failed: Invalid argument
(EE) failed to read Wayland events: Broken pipe
```

[lab1.log](https://github.com/user-attachments/files/26443735/lab1.log)
[lab2.log](https://github.com/user-attachments/files/26443736/lab2.log)

### Steps to reproduce

Sorry for the annoying steps here but i not sure which other games caused the crashes (and also they're not free), i tried to write a simple SDL window but it didn't trigger the crash.

1. Open firefox (load youtube, sometimes it seems to not happen when there's no page loaded, and i usually have a youtube page open).
[...]

PR #3490, opened 2026-04-02
osd-thumbnail: handle buffer allocation failure

Consolatis:
Fixes:
- #3489


Consolatis:
Confirmed fixing a crash and unlikely to introduce issues so self-merging.

johanmalm:
Thanks for this. Great to respond to quickly.

PR #3493, opened 2026-04-03
rcxml: allow to selectively enable privileged interfaces

xi:
This adds an allow list for privileged protocols to the settings.

For backwards compatibility (and ease of use), all protocols are allowed unless a `<privilegedInterfaces>` block exists.

> Having a user controllable block list for protocols is something we should definitely implement at some point. But that indeed raises the question of how some specific applications can then still use those protocols. -- https://github.com/labwc/labwc/pull/3089#issuecomment-3321933194

In comparison to #3089, this only implements the global allow list. There is no way to give additional privileges to applications that are launched either during autostart or by some action. Those features seemed to be too controversial.

On its own, this mechanism can already be used to reduce the attack surface by disabling interfaces that are not used by the specific user, e.g. `zwlr_virtual_pointer_manager_v1`, `zwp_virtual_keyboard_manager_v1`, or `zwlr_data_control_manager_v1`.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -1445,6 +1445,52 @@ situation.
 	Whether to apply a bilinear filter to the magnified image, or
 	just to use nearest-neighbour. Default is true - bilinear filtered.
 
+## PRIVILEGED INTERFACES
+
+Labwc supports a small set of privileged wayland interfaces. All of these
+interfaces are enabled by default for applications unless they are running
+via a sandbox environment supporting the security-context-v1 protocol.
+
+Security conscious users are suggested to use a sandbox framework to run
+potentially untrusted applications as it additionally limits access to the
+filesystem (including labwc configuration) and other services like dbus.
+
+In addition to that, privileged protocols can be restricted for non-sandboxed
+clients by defining a `<privilegedInterfaces>` block:
+
+```
+<privilegedInterfaces>
+  <allow>zwlr_layer_shell_v1</allow>
+  <allow>zwlr_virtual_pointer_manager_v1</allow>
+</privilegedInterfaces>
+```
+
+*<privilegedInterfaces><allow>*
+	Name of the interface that should be allowed.
+
+This is the full list of interfaces that can be controlled with this mechanism:
+
+- `wp_drm_lease_device_v1`
+- `zwlr_gamma_control_manager_v1`
+- `zwlr_output_manager_v1`
+- `zwlr_output_power_manager_v1`
+- `zwp_input_method_manager_v2`
+- `zwlr_virtual_pointer_manager_v1`
+- `zwp_virtual_keyboard_manager_v1`
+- `zwlr_export_dmabuf_manager_v1`
+- `zwlr_screencopy_manager_v1`
+- `ext_data_control_manager_v1`
+- `zwlr_data_control_manager_v1`
+- `wp_security_context_manager_v1`
+- `ext_idle_notifier_v1`
+- `zwlr_foreign_toplevel_manager_v1`
+- `ext_foreign_toplevel_list_v1`
+- `ext_session_lock_manager_v1`
+- `zwlr_layer_shell_v1`
+- `ext_workspace_manager_v1`
+- `ext_image_copy_capture_manager_v1`
+- `ext_output_image_capture_source_manager_v1`
+
 ## ENVIRONMENT VARIABLES
 
 *XCURSOR_PATH*
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -94,6 +94,43 @@ parse_window_type(const char *type)
 	}
 }
 
+uint32_t
+parse_privileged_interface(const char *name)
+{
+	static const char * const ifaces[] = {
+		"wp_drm_lease_device_v1",
+		"zwlr_gamma_control_manager_v1",
+		"zwlr_output_manager_v1",
+		"zwlr_output_power_manager_v1",
+		"zwp_input_method_manager_v2",
+		"zwlr_virtual_pointer_manager_v1",
+		"zwp_virtual_keyboard_manager_v1",
+		"zwlr_export_dmabuf_manager_v1",
+		"zwlr_screencopy_manager_v1",
+		"ext_data_control_manager_v1",
+		"zwlr_data_control_manager_v1",
+		"wp_security_context_manager_v1",
+		"ext_idle_notifier_v1",
+		"zwlr_foreign_toplevel_manager_v1",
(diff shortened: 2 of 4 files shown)


docs/labwc-config.5.scd
```diff
@@ -1445,6 +1445,41 @@ situation.
 	Whether to apply a bilinear filter to the magnified image, or
 	just to use nearest-neighbour. Default is true - bilinear filtered.
 
+## PRIVILEGED INTERFACES
+
+Labwc supports a small set of privileged wayland interfaces. All of these
+interfaces are enabled by default. Security conscious users may override this by
+defining a `<privilegedInterfaces>` block:
+
```
Consolatis, line 1460: This should mention that those privileges protocols are blocked when applications are run via the security-context protocol as used by some sandbox environments like flatpak. That is also the much more secure way to run potentially untrusted applications.

I suggest a wording similar to
```
Labwc supports a small set of privileged wayland interfaces. All of these
interfaces are enabled by default for applications unless they are running
via a sandbox environment supporting the security-context-v1 protocol.

Security conscious users are suggested to use a sandbox framework to run
potentially untrusted applications as it additionally limits access to the
filesystem (including labwc configuration) and other services like dbus.

Privileged protocols can additionally be selectively enabled for all clients
by defining a `<privilegedInterfaces>` block:
[..]

If no `privilegedInterfaces` exists all protocols are allowed for usual clients.
```

include/config/types.h
```diff
@@ -130,4 +130,25 @@ enum cycle_app_id_filter {
 	CYCLE_APP_ID_CURRENT,
 };
 
+enum lab_interface {
+	LAB_IFACE_INVALID          = 0,
+	LAB_IFACE_NOT_SET          = 1,
+	LAB_IFACE_DRM_LEASE        = 1 << 1,
```
Consolatis: Suggest to keep the ordering of `protocol_is_privileged()` in `src/server.c` to make it easier to sync the changes. It also seems to miss at least
- `ext_image_copy_capture_manager_v1`
- `ext_output_image_capture_source_manager_v1`

I wonder if we can make this a bit easier to sync automatically, maybe by using the array index from `protocol_is_privileged()` as left shift to parse the config values. E.g. moving the array into the file scope and then adding a small helper function to parse an incoming string returning either `0` or `1u << index`.

Consolatis:
In general I think this approach is fine. Maybe we can slightly improve the logic so newly added privileged protocols are automatically blocked if `rc.allowed_interfaces != LAB_IFACE_NOT_SET`.

include/config/types.h
xi: > maybe by using the array index from `protocol_is_privileged()` as left shift

I like that idea a lot! Will work on that when I have time.

Consolatis: Thinking about this some more, we don't actually need to move the array into the file scope. We could simply rename `protocol_is_privileged()` to something like `protocol_get_privileged_id()` and change it to return an `uint32_t` rather than `bool`. Can't come up with a better name right now. The current checks for `protocol_is_privileged()` would just keep working because it would return `0` if its not a privileged protocol.

We can additionally add a
```c
/* Privileged protocols are additionally mapped to rc.allowed_interfaces bitset */
static_assert(ARRAY_SIZE(rejected) <= (sizeof(rc.allowed_interfaces) * 8),
	"rc.allowed_interfaces too small for amount of privileged protocols");
```
so we don't end up overflowing when we have more than 32 privileged protocols. If it ever gets triggered in the future its easy to use `uint64_t` instead.

docs/labwc-config.5.scd
```diff
@@ -1445,6 +1445,52 @@ situation.
...
+Labwc supports a small set of privileged wayland interfaces. All of these
+interfaces are enabled by default for applications unless they are running
+via a sandbox environment supporting the security-context-v1 protocol.
+
+Security conscious users are suggested to use a sandbox framework to run
+potentially untrusted applications as it additionally limits access to the
+filesystem (including labwc configuration) and other services like dbus.
+
+Privileged protocols can additionally be selectively enabled for all clients
+by defining a `<privilegedInterfaces>` block:
```
Consolatis: This might sound like this mechanism allows specific protocols to also be exposed to sandboxed clients. Maybe something like this clears it up?
```
Privileged protocols can additionally be selectively enabled for all
non-sandboxed clients by defining a `<privilegedInterfaces>` block like:
```

docs/rc.xml.all
```diff
@@ -691,4 +691,8 @@
     <useFilter>yes</useFilter>
   </magnifier>
 
+  <privilegedInterfaces>
+    <allow>zwlr_layer_shell_v1</allow>
+  </privilegedInterfaces>
+
```
Consolatis: This must be removed because `rc.xml.all` represents the default internal configuration which allows all protocols.

src/config/rcxml.c
```diff
@@ -1459,6 +1507,7 @@ rcxml_init(void)
 	rc.allow_tearing = LAB_TEARING_DISABLED;
 	rc.auto_enable_outputs = true;
 	rc.reuse_output_mode = false;
+	rc.allowed_interfaces = LAB_IFACE_NOT_SET;
```
Consolatis: Maybe we could use `UINT32_MAX` instead? E.g. all protocols are allowed by default. Then we could also keep using `1u << 0` for the first element and don't need the `LAB_IFACE_NOT_SET` define.

src/server.c
```diff
@@ -338,18 +305,23 @@ server_global_filter(const struct wl_client *client, const struct wl_global *glo
...
 			wlr_log(WLR_DEBUG, "Blocking %s for security context %s->%s->%s",
 				iface->name, security_context->sandbox_engine,
 				security_context->app_id, security_context->instance_id);
 		}
 		return allow;
+	} else if (rc.allowed_interfaces != LAB_IFACE_NOT_SET) {
+		uint32_t i = parse_privileged_interface(iface->name);
+		if (i != 0 && (i & rc.allowed_interfaces) == 0) {
+			return false;
+		}
```
Consolatis: Can we use this instead?
```c
	uint32_t iface_id = parse_privileged_interface(iface->name);
	if (iface_id && !(iface_id & rc.allowed_interfaces)) {
		return false;
	}
```

Consolatis:
With the remaining nitpicks fixed this LGTM.

I want to give @johanmalm a chance to agree to the feature itself and config naming though.

johanmalm:
LGTM

Consolatis:
Thanks for your work and persistence on this!

xi:
Thanks for the thorough review! That was genuinely helpful.

Issue #1432, opened 2024-01-15
[Feature request] Menu accelerators


oredaze:
Openbox has the ability to activate a menu item by pressing the first letter of it's name or a letter prefixed with `_`.
This allows us to quickly execute key combos, for example I have `Meta+a` then `f` to launch firefox in fraction of a second.

Or is that not possible in wayland?

johanmalm:
Happy to merge this.
There is nothing about Wayland stopping that.
The underline can be achieved with `pango` markup.

Consolatis:
We have another request for this via IRC:

> \<yazi> in openbox client-menu has accelerators enabled. for maximize it's x. so I can quickly open client-menu, press x and the window will be maximized. same goes for other menu entries - they also have quick keyboard shortcuts. if needed i can quickly load my openbox setup and give all shortcuts
```

ch3rn1ka:
I've been piecing together the source code, and here's how I'd implement this:

1. menu.h: add a field to the `menuitem` struct to store menu accelerators.
2. menu.c: parse menu item text in `item_create()`, determine an accelerator for each item. 
    * If it's specified with '_', delete this character and set the accelerator to be the character after it. Wrap the target letter in \<u>\</u> so that it renders appropriately with an underline (pango supports this, tested with `pango_layout_set_markup()`).
    * Set it to the first char of the text otherwise.
3. keyboard.c & menu.c: in `handle_menu_keys()`, pass unhandled keys to a new helper function implemented in menu.c. This function will grab the active menu leaf and check for matching accelerators.
    * If it's the only match, execute the item associated with it (via `menu_execute_item()`, I presume).
    * Toggle selection between multiple items with the same accelerator otherwise.
Note: I didn't find any logic for selecting a particular item in menu.c, there are only `menu_item_select_previous()` and `menu_item_select_next()` for selecting the items directly above/lower the current one, would probably need to implement it as well.

I'd love to work on a PR. What do you think?

oredaze:
Sounds good to me.

Consolatis:
The general approach seems fine to me.

`include/common/string-helpers.h` and/or `include/common/buf.h` may help with building the new string. `font_buffer_create()` looks like it would need a new `bool` argument to use `pango_layout_set_markup()` instead of `_set_text()`. Along with a new argument for `scaled_font_buffer_update()` and `struct scaled_font_buffer` itself to do the same.

Although I would suggest to ignore the display part in the menu items for now and get the remaining logic working first.

[...]

PR #3505, opened 2026-04-10
menu: implement menu accelerators

ch3rn1ka:
This is a pull request following the discussion at #1432. I've implemented some basic functionality as described in the issue, but there are still things to be done:

- ~~Omit the underscore from the label after defining an accelerator~~ (done)
- ~~Render the accelerator with an underline by wrapping it in \<u>\</u>~~ (done)
- ~~Add a way to escape underscores~~ (done)

After launching a clean install of Openbox, however, I noticed that the accelerators aren't marked at all. Is this theme-specific?

### DIFF
--- a/src/common/font.c
+++ b/src/common/font.c
@@ -79,7 +79,7 @@ font_get_buffer_size(int max_width, const char *text, struct font *font,
 void
 font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	int height, const char *text, struct font *font, const float *color,
-	cairo_pattern_t *bg_pattern, double scale)
+	cairo_pattern_t *bg_pattern, double scale, bool use_markup)
 {
 	if (string_null_or_empty(text)) {
 		return;
@@ -123,7 +123,7 @@ font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	PangoLayout *layout = pango_cairo_create_layout(cairo);
 	pango_context_set_round_glyph_positions(pango_layout_get_context(layout), false);
 	pango_layout_set_width(layout, width * PANGO_SCALE);
-	pango_layout_set_text(layout, text, -1);
+	use_markup ? pango_layout_set_markup(layout, text, -1) : pango_layout_set_text(layout, text, -1);
 	pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
 
 	if (!opaque_bg) {
--- a/src/common/font.c
+++ b/src/common/font.c
@@ -123,7 +123,13 @@ font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	PangoLayout *layout = pango_cairo_create_layout(cairo);
 	pango_context_set_round_glyph_positions(pango_layout_get_context(layout), false);
 	pango_layout_set_width(layout, width * PANGO_SCALE);
-	use_markup ? pango_layout_set_markup(layout, text, -1) : pango_layout_set_text(layout, text, -1);
+
+	if (use_markup) {
+		pango_layout_set_markup(layout, text, -1);
+	} else {
+		pango_layout_set_text(layout, text, -1);
+	}
+
 	pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
 
 	if (!opaque_bg) {
--- a/src/common/font.c
+++ b/src/common/font.c
@@ -79,7 +79,7 @@ font_get_buffer_size(int max_width, const char *text, struct font *font,
 void
 font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	int height, const char *text, struct font *font, const float *color,
-	cairo_pattern_t *bg_pattern, double scale)
+	cairo_pattern_t *bg_pattern, double scale, bool use_markup)
 {
 	if (string_null_or_empty(text)) {
 		return;
@@ -123,7 +123,13 @@ font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	PangoLayout *layout = pango_cairo_create_layout(cairo);
 	pango_context_set_round_glyph_positions(pango_layout_get_context(layout), false);
 	pango_layout_set_width(layout, width * PANGO_SCALE);
-	pango_layout_set_text(layout, text, -1);
+
+	if (use_markup) {
+		pango_layout_set_markup(layout, text, -1);
+	} else {
+		pango_layout_set_text(layout, text, -1);
+	}
+
 	pango_layout_set_ellipsize(layout, PANGO_ELLIPSIZE_END);
 
 	if (!opaque_bg) {
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -139,6 +139,8 @@ handle_modifiers(struct wl_listener *listener, void *data)
 	struct seat *seat = keyboard->base.seat;
 	struct wlr_keyboard *wlr_keyboard = keyboard->wlr_keyboard;
 
+	key_state_indicator_update(seat);
+
 	if (server.input_mode == LAB_INPUT_STATE_MOVE) {
 		/* Any change to the modifier state re-enable region snap */
 		seat->region_prevent_snap = false;
@@ -201,8 +203,10 @@ match_keybinding_for_sym(uint32_t modifiers,
 		if (modifiers ^ keybind->modifiers) {
 			continue;
 		}
-		if (view_inhibits_actions(server.active_view, &keybind->actions)) {
-			continue;
+		if (!(keybind->override_inhibition)) {
+			if (view_inhibits_actions(server.active_view, &keybind->actions)) {
+				continue;
+			}
 		}
 		if (sym == XKB_KEY_NoSymbol) {
 			/* Use keycodes */
@@ -631,6 +635,9 @@ handle_key(struct wl_listener *listener, void *data)
 	struct seat *seat = keyboard->base.seat;
 	struct wlr_keyboard_key_event *event = data;
 	struct wlr_seat *wlr_seat = seat->wlr_seat;
+
+	key_state_indicator_update(seat);
+
 	idle_manager_notify_activity(seat->wlr_seat);
 
 	/* any new press/release cancels current keybind repeat */
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -460,7 +460,7 @@ handle_menu_keys(struct keysyms *syms)
 			menu_close_root();
 			cursor_update_focus();
 			break;
-		default:
+		default: {
 			char accelerator = keysym_to_char(syms->syms[i]);
 			if (accelerator == '\0') {
 				continue;
@@ -469,7 +469,7 @@ handle_menu_keys(struct keysyms *syms)
 				menu_call_selected_actions();
 			}
 			break;
-		}
+		}}
 		break;
 	}
 }
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -419,6 +419,19 @@ handle_change_vt_key(struct keyboard *keyboard,
 	return false;
 }
 
+static char
+keysym_to_char(uint32_t keysym) {
+	if (keysym >= 0x0020 && keysym <= 0x00FF) {
+		return (char)keysym;
+	}
+
+	if (keysym >= XKB_KEY_KP_0 && keysym <= XKB_KEY_KP_9) {
+		return (char)('0' + (keysym - XKB_KEY_KP_0));
+	}
+
+	return '\0';
+}
+
 static void
 handle_menu_keys(struct keysyms *syms)
 {
@@ -447,7 +460,14 @@ handle_menu_keys(struct keysyms *syms)
 			cursor_update_focus();
 			break;
 		default:
-			continue;
+			char accelerator = keysym_to_char(syms->syms[i]);
+			if (accelerator == '\0') {
+				continue;
+			}
+			if (menu_item_select_by_accelerator(accelerator)) {
+				menu_call_selected_actions();
+			}
+			break;
 		}
 		break;
 	}
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -444,15 +444,24 @@ handle_menu_keys(struct keysyms *syms)
 			break;
 		case XKB_KEY_Return:
 		case XKB_KEY_KP_Enter:
-			menu_call_selected_actions();
+			if (!menu_call_selected_actions()) {
+				menu_submenu_enter();
+			};
 			break;
 		case XKB_KEY_Escape:
 			menu_close_root();
 			cursor_update_focus();
 			break;
-		default:
-			continue;
-		}
+		default: {
+			uint32_t accelerator = xkb_keysym_to_utf32(syms->syms[i]);
+			if (accelerator == 0) {
+				continue;
+			}
+			if (menu_item_select_by_accelerator(accelerator)) {
+				menu_call_selected_actions();
+			}
+			break;
+		}}
 		break;
 	}
 }
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -419,20 +419,6 @@ handle_change_vt_key(struct keyboard *keyboard,
 	return false;
 }
 
-static char
-keysym_to_char(uint32_t keysym)
-{
-	if (keysym >= 0x0020 && keysym <= 0x00FF) {
-		return (char)keysym;
-	}
-
-	if (keysym >= XKB_KEY_KP_0 && keysym <= XKB_KEY_KP_9) {
-		return (char)('0' + (keysym - XKB_KEY_KP_0));
-	}
-
-	return '\0';
-}
-
 static void
 handle_menu_keys(struct keysyms *syms)
 {
@@ -454,15 +440,17 @@ handle_menu_keys(struct keysyms *syms)
 			break;
 		case XKB_KEY_Return:
 		case XKB_KEY_KP_Enter:
-			menu_call_selected_actions();
+			if (!menu_call_selected_actions()) {
+				menu_submenu_enter();
+			};
 			break;
 		case XKB_KEY_Escape:
 			menu_close_root();
 			cursor_update_focus();
 			break;
 		default: {
-			char accelerator = keysym_to_char(syms->syms[i]);
-			if (accelerator == '\0') {
+			uint32_t accelerator = xkb_keysym_to_utf32(syms->syms[i]);
+			if (accelerator == 0) {
 				continue;
 			}
 			if (menu_item_select_by_accelerator(accelerator)) {
--- a/src/input/keyboard.c
+++ b/src/input/keyboard.c
@@ -420,7 +420,8 @@ handle_change_vt_key(struct keyboard *keyboard,
 }
 
 static char
-keysym_to_char(uint32_t keysym) {
+keysym_to_char(uint32_t keysym)
+{
 	if (keysym >= 0x0020 && keysym <= 0x00FF) {
 		return (char)keysym;
 	}
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -148,7 +148,7 @@ item_create(struct menu *menu, const char *text, const char *icon_name, bool sho
 	menuitem->accelerator = tolower(*it);
 	while (*it != '\0') {
 		if (*it == '_' && *it != '\0') {
-			menuitem->accelerator = *(++it);
+			menuitem->accelerator = tolower(*(++it));
 			break;
 		}
 		it++;
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -1475,7 +1475,7 @@ bool
 menu_item_select_by_accelerator(char accelerator)
 {
 	struct menu *menu = get_selection_leaf();
-	if (!menu) {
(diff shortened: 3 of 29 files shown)


src/input/keyboard.c
```diff
@@ -447,7 +461,14 @@ handle_menu_keys(struct keysyms *syms)
...
-			continue;
+			char accelerator = keysym_to_char(syms->syms[i]);
+			if (accelerator == '\0') {
+				continue;
+			}
+			if (menu_item_select_by_accelerator(accelerator)) {
+				menu_call_selected_actions();
+			}
+			break;
 		}
```
Consolatis: Regarding the CI error
> label followed by a declaration is a C23 extension

You can use curly braces around the code block. It looks a bit weird for the default branch but its the best we've come up with so far. E.g. something like
```c
default: {
	char accelerator = ...
	[..]
	break;
}}
```

Consolatis:
Thanks for the PR.

Will look into it in more detail later. I don't quite understand the intended behavior when there are multiple menu items with the same accelerator. Should it never execute any but instead just jump from one to the next and thus require an additional `Return` to execute it (e.g. run the actions or enter a submenu)?

ch3rn1ka:
> Should it never execute any but instead just jump from one to the next and thus require an additional `Return` to execute it (e.g. run the actions or enter a submenu)?

Yes, it does exactly that! The `menu_select_item_by_accelerator()` function doesn't actually execute anything by itself, but rather selects an item and returns a boolean value that represents whether the item needs to be executed.

The logic of choosing an entry to select and deciding whether it needs to be executed is handled here:
```c
		if (item->accelerator == accelerator) {
			if (!matched) {
				/* Found first match */
				next_selection = item;
				needs_exec = true;
				matched = true;
			} else {
				/* Found another match,
				 * cycle selection instead of executing
				 */
				needs_exec = false;
				break;
			}
		}
```

P.S: `needs_exec` is the return value. If the only matched item is a submenu, then we just open it (since "executing" a submenu doesn't make sense), this is handled a bit later in the code.

P.P.S: if this approach seems bad to you, I can cram the execution logic into the function, should be a 3-4 lines change.

edit: grammar
edit2: changes in logic

ch3rn1ka:
On a side note, I've just found a case that I missed while testing: if an item is selected, and then a key not associated with any accelerator is pressed, the selected item is executed, which is unintended. Will push a quick fix in a couple of minutes.

johanmalm:
Welcome. And thanks for contributing with code :smile:

I just tried this PR and it seems to be working well so far.

I wasn't aware of the Openbox behaviour whereby the selection moves between items with the same "menu shortcut". I have just tried it though and it seems to work well.

Interestingly, Openbox does not require a '_' if you want the first character to be the menu shortcut. I think this is a good way for doing it. I quote from the Openbox website (https://openbox.org/help/Menus):

```
Specifying menu shortcuts

The first character in the label is used for a keyboard shortcut to directly use
the menu item. This can be overridden by using the "_" character in front of
the character, you wish to use for the shortcut. For example, "The _shortcut"
would make "s" the shortcut key for this menu item. If you want a "_" character
to appear in the string, you can use two underscores together, such as in
"An __ underscore". 
```

I noticed when test-driving the PR that with a label of `_Foo`, a capital `F` needs to be pressed to select it. Openbox requires a lower case `f` in the same situation. I don't really feel strongly about it, but suspect lower case is better.

> After launching a clean install of Openbox, however, I noticed that the accelerators aren't marked at all. Is this theme-specific?

I don't know. I've just tried the same. Maybe it's just Gtk/Qt clients that show the underscore (when opening menu with Alt+<letter> normally).

I feel that we ought to decide on when any underscore markup should be shown. For example, it could be:

1. Always
2. Never
3. Only while Alt is pressed
4. Only after Alt has been pressed/released
5. (4) + after any key other than arrow has been pressed (i.e. typically other letters)

I haven't looked at code yet, but can do if helpful.

ch3rn1ka:
@johanmalm, thanks so much for your feedback!

> Interestingly, Openbox does not require a '_' if you want the first character to be the menu shortcut.

Yes, I made sure to fall back to the first character of the label in case no accelerator is specified.

> I noticed when test-driving the PR that with a label of `_Foo`, a capital `F` needs to be pressed to select it. Openbox requires a lower case `f` in the same situation. I don't really feel strongly about it, but suspect lower case is better.

Good catch! I forgot to convert accelerators to lowercase when parsing labels. Fixed in the latest commit.

src/menu/menu.c
```diff
@@ -1460,6 +1471,53 @@ menu_item_select_previous(void)
...
+	bool matched = false;
+
+	struct menuitem *selection = menu->selection.item;
+	struct wl_list *start = selection ? &selection->link : &menu->menuitems;
+	struct wl_list *current = start;
+	struct menuitem *item = NULL;
+	struct menuitem *next_selection = NULL;
+	do {
+		current = current->next;
+		item = wl_container_of(current, item, link);
```
Consolatis, line 1531: This needs something like
```c
do {
	current = current->next;
	if (current == &menu->menuitems) {
		/* wrap around */
		continue;
	}
	item = wl_container_of(current, item, link);
	[..]
}  while (current != start);
```

Otherwise ASAN complains when accessing `item->accelerator` because it reads random memory. Can be seen with `meson configure build -Db_sanitize=address,undefined`.

We also need to check for an empty list at the start of the function, otherwise we'd have an endless loop when adding the `continue` or access random memory when not adding it (due to the condition being on the bottom of the loop).

So something like
```c
if (!menu || wl_list_empty(&menu->itemitems)) {
	return false;
}
```

Consolatis:
I do have some menu entries using unicode emoticons like
```xml
        <menu id="desktop-menu">
                <menu id="config" label="🔧  Config">
                        <item label="📜  rc.xml">
                                <action name="Execute" command="foot nano -w ~/.config/labwc/rc.xml" />
                        </item>
                        <item label="📜  menu.xml">
                                <action name="Execute" command="foot nano -w ~/.config/labwc/menu.xml" />
                        </item>
                </menu>
                <menu id="display" label="📺  Display">
                        <item label="💻  wlr-randr">
                            <action name="Execute" command="foot --hold wlr-randr" />
                        </item>
                        <item label="💻  xrandr">
                            <action name="Execute" command="foot --hold xrandr" />
                        </item>
                        <item label="📺  wdisplays">
                            <action name="Execute" command="wdisplays" />
                        </item>
                </menu>
                <item label="🔄  _Reconfigure">
                        <action name="Reconfigure" />
                </item>
                <item label="🛑  Exit">
                        <action name="Exit" />
                </item>
                <menu id="root-menu" />
        </menu>
```

Using `_` works as expected but I wonder if we should try to skip non ASCII characters (and maybe space) when assigning the default "first character" as accelerator. Thoughts?

src/menu/menu.c
ch3rn1ka: Addressed the random memory read issue in the latest commit, thanks!

ch3rn1ka:
> I wonder if we should try to skip non ASCII characters (and maybe space) when assigning the default "first character" as accelerator. Thoughts?

Openbox does actually ignore such characters. We probably should implement it, I'll think about it overnight.

ch3rn1ka:
From what I've gathered, all non-ASCII characters have a value larger than 127, assuming we use UTF-8. Would a simple check like this suffice? Seems to work fine in the latest commit.

```c
isspace(c) || c > 127
```

johanmalm:
> I wonder if we should try to skip non ASCII characters (and maybe space) when assigning the default "first character" as accelerator. Thoughts?

Yes, I think that's a good plan.

However, we ought to consider utf-8 characters when defining with `_`. I think it's reasonable to assume that some users will want to use this with languages that do not fit within ASCII characters. We already use some `g_utf8_` helpers for this (see `src/config/keybind.c`).

ch3rn1ka:
> I think it's reasonable to assume that some users will want to use this with languages that do not fit within ASCII characters

Do we want to stick to Openbox in this regard? I tested it, and cyrillic characters can't be used for accelerators.

Consolatis:
> Do we want to stick to Openbox in this regard? I tested it, and cyrillic characters are not used for accelerators, but rather the first ASCII character.

In general IMHO it would be good if labwc could support unicode accelerators as limiting it to ASCII feels a bit outdated.

If we try to support unicode characters as accelerators it might be a bit complicated to then also ignore certain ranges like those used for emoticons. If we want to support one or the other I think supporting unicode accelerators would be more important than ignoring icon replacements (especially as we are supporting proper icons per menuitem already).

I \*think\* we'd need a uint32_t for the codepoints rather than a char but its been a while since I looked into the libxkbcommon / glib APIs in regards to unicode parsing. Same for looping by unicode character through a utf-8 string.

ch3rn1ka:
I did some research, and I agree that implementing Unicode accelerators would make it unnecessarily hard to ignore emojis, since new ones get added every so often, and they are scattered across the table. 

@johanmalm proposed a compromise to allow for Unicode accelerators when defining them explicitly with an underscore, but still use ASCII for default bindings. Sounds very logical to me.

I can draft parsing logic for Unicode labels, but it'll take some time since I'm rather unfamiliar with it.

ch3rn1ka:
I think I got it working, it turned out to be easier than I expected, since C11 introduced `mbrtoc32()`. I tested it with various labels, but mainly with Cyrillic and Arabic. Couldn't decide what to do with the default keybinds, for now it takes the first character of a label if an accelerator wasn't specified. If everything's fine, I can move to the rendering part, should be as easy as modifying a string before rendering it with markdown (will implement it with boolean flags as @Consolatis suggested).

I also took the liberty to expand the functionality of the Return key when a submenu is selected: now pressing Return will enter the submenu.

Consolatis:
Had a quick test and it worked well for things like `ß` and `ö` or `€` and `ŧ`, both in the underscore variant and as first char of the label. Pretty cool!

> I also took the liberty to expand the functionality of the Return key when a submenu is selected: now pressing Return will enter the submenu.

Agree, that makes sense.

src/common/font.c
```diff
@@ -123,7 +123,7 @@ font_buffer_create(struct lab_data_buffer **buffer, int max_width,
 	PangoLayout *layout = pango_cairo_create_layout(cairo);
 	pango_context_set_round_glyph_positions(pango_layout_get_context(layout), false);
 	pango_layout_set_width(layout, width * PANGO_SCALE);
-	pango_layout_set_text(layout, text, -1);
+	use_markup ? pango_layout_set_markup(layout, text, -1) : pango_layout_set_text(layout, text, -1);
```
Consolatis: We don't usually execute functions like this in conditionals. Could you change it to something like
```c
if (use_markup) {
 ..
} else {
 ..
}
```

src/scaled-buffer/scaled-font-buffer.c
```diff
@@ -26,7 +26,7 @@ _create_buffer(struct scaled_buffer *scaled_buffer, double scale)
 
 	/* Buffer gets free'd automatically along the backing wlr_buffer */
 	font_buffer_create(&buffer, self->max_width, self->height, self->text,
-		&self->font, self->color, bg_pattern, scale);
+		&self->font, self->color, bg_pattern, scale, true);
```
Consolatis: Always using markup seems wrong and a change in behavior. E.g. titles could suddenly contain markup. This should be a boolean within `struct scaled_font_buffer` instead. Note that `_equal()` also has to have a check added for it, otherwise we could end up re-using a wrong buffer.

src/menu/menu.c
```diff
@@ -143,18 +143,31 @@ item_create(struct menu *menu, char *text, const char *icon_name, bool show_arro
...
+				accelerator =
+					(uint32_t)towlower((wint_t)codepoint);
+			}
+
+			if (*(it + 1) != '\0') {
+				int underscore_index = it - text;
+				new_text = malloc(strlen(text) + 8);
+				if (!new_text) {
+					break;
+				}
```
Consolatis: You can use `xmalloc()`, its basically the same but exits on allocation error.

Alternative is to use `struct buf` (from `common/buf.h`) and `buf_add()` +  `buf_add_fmt()` instead.
I prefer `struct buf` here because it makes the intention much clearer.

Edit:
I see we are missing some `buf_add_n(struct buf *buf, const char *data, size_t count)` though.. Fine with doing it manually for now, we can always change this one later.

Edit 2:
Technically I guess we could also do something like
```c
/* Mark accelerator character */
struct buf buf = BUF_INIT;
buf_add_fmt(&buf, "%.*s", underscore_index, text);
buf_add_fmt(&buf, "<u>%c</u>%s", *(it + 1), it + 2);
/* Result now in buf.data */
```

Another issue with both approach are unicode accelerators. So we likely need to use `%s` rather than `%c` and have another temporary 4 byte buffer to encode the codepoint into. Unless there is a `%c` variant which supports unicode. Or we use something like `buf_add_fmt(&buf, "<u>%.*s</u>%s", utf8_len, it + 1, it + 1 + utf8_len)`.

src/menu/menu.c
```diff
@@ -174,6 +187,11 @@ item_create(struct menu *menu, char *text, const char *icon_name, bool show_arro
 	}
 
 	menuitem->accelerator = accelerator;
+	if (new_text) {
+		menuitem->text = xstrdup(new_text);
```
Consolatis: `new_text` doesn't seem to be free'd. So either we assign it directly (without `xstrdup()`) or add a `free(new_text)`. In case of `struct buf` it would be a `xstrdup(buf.data)` and `buf_reset()` to get rid of slight potential over-allocation.

ch3rn1ka: Sorry for the commit noise! Is it possible for me to run the checks locally to not waste your time? :^)

Consolatis: Yes, via `scripts/check` or `scripts/check --file=src/menu/menu.c`

Edit:
And no worries about commit noise. I could also stop reviewing until you are ready if that is easier for your workflow. Really up to you.

ch3rn1ka: I appreciate that, please do keep reviewing, your feedback is very helpful! I was mostly just looking to catch the minor errors on my end before they reach you.

johanmalm:
@ch3rn1ka

I'm impressed by your methodical approach and commitment. Just a thought - would it be worth trying to merge what we have so far? I don't know if we're at a natural point to merge, but it could make sense before doing the pango-markup stuff.

The reason I think it could be a good idea is that:

1. It forces me/Consolatis/others to do some final testing/reviewing - which is always slightly different when we're "on the hook" to merge. 
2. It makes us tidy up the commit history. You might be a git-pro, but in case you're not, it's easier with a smaller scope.
3. We might start he cool-down period for the next release as early as first week of May, and it would be good to get it "in the bank" so to speak well before that :smile:

Ref:
- https://github.com/labwc/labwc/discussions/3510

ch3rn1ka:
I think [1c646af](https://github.com/labwc/labwc/pull/3505/commits/1c646af14448fa0d36ce56c0e47a76a752ab8595) is OK to merge, the only issue is that the underscore is not removed from the label if an accelerator is defined explicitly (should be an easy fix, I guess something like `memmove(it, it + 1, strlen(it + 1) + 1)` would work, where `it` points to the byte with the underscore). 

And now that I think of it, compared to what I'm trying to implement right now, it's closer to what Openbox does, due to the absence of markup.

> We might start he cool-down period for the next release as early as first week of May, and it would be good to get it "in the bank" so to speak well before that 😄

I think I'll need another two days to get the markup working. Feel free to merge any previous commits, and thanks for your patience!

src/menu/menu.c
ch3rn1ka: And as for freeing menuitems: I noticed that the `item_destroy()` function is not implemented. Is this intentional?

ch3rn1ka:
I've been daily-driving this PR, and everything seems to work fine. I made sure to approach adding the `use_markup` flag carefully, but I'd appreciate a second pair of eyes on the logic. :^)

Also, what are your thoughts on supporting escaping underscores in item labels with backslashes?

src/menu/menu.c
johanmalm: I think it auto-destroys when the nodes are freed

johanmalm:
> I've been daily-driving this PR, and everything seems to work fine. I made sure to approach adding the `use_markup` flag carefully, but I'd appreciate a second pair of eyes on the logic. :^)

I discussed with @Consolatis on IRC. They will review the PR. 

> Also, what are your thoughts on supporting escaping underscores in item labels with backslashes?

Yes, good idea. My advice is to focus on merging what we have first.

Consolatis:
> Also, what are your thoughts on supporting escaping underscores in item labels with backslashes?

Openbox seems to escape them with [doubling](https://openbox.org/help/Menus#Specifying_menu_shortcuts):
> If you want a "_" character to appear in the string, you can use two underscores together, such as in "An __ underscore". 

In general I really like the overall design here. Nice catch with `strdup_printf()`, completely forgot about that function.

There are some things I'd like to change but I could also do that in a follow-up PR depending on how much time you want to invest in the feature. Namely
- I wonder if we should add a `scaled_font_buffer_update_markup()` rather than having the additional `bool` argument in `scaled_font_buffer_update()`. That would allow us to drop the `false` argument from the other callers
- The parsing logic seems complicated enough that it deserves its own function in `menu.c`, that slightly cleans up `item_create()` so its easier to read
- We could use `strchr()` to find the next underscore
- `mbrtoc32()` seems to be locale dependent, so we should likely either set it to always use a utf-8 locale or use glib instead to get the codepoint

I do have a work-in-progress branch over [here](https://github.com/ch3rn1ka/labwc/compare/menu-accelerators...Consolatis:labwc:review/menu_unicode_accelerator) which creates a new parsing function and uses `strchr()` + ignores double underscores (but doesn't yet replace it with a single one).

src/menu/menu.c
Consolatis: It is implemented later in `menu.c` so there is a forward declaration.

ch3rn1ka:
I think I've reached the point where the implementation is mostly complete with only small fixes needed here and there. I took a close look at @Consolatis' review branch and tried to improve on it, adding logic for escaping doubled underscores and using the UTF-8 locale while reading. I didn't use glib for the latter, but rather the `uselocale()` function two times.

Is this a good way of doing this?
```c
	/* Temporarily set locale to UTF-8 */
	locale_t utf8_locale = newlocale(LC_CTYPE_MASK, "C.UTF-8", (locale_t)0);
	locale_t old_locale = (locale_t)0;
	if (utf8_locale != (locale_t)0) {
		old_locale = uselocale(utf8_locale);
	}

	/* Reading logic... /*

	/* Restore previous locale */
	if (utf8_locale != (locale_t)0) {
		uselocale(old_locale);
		freelocale(utf8_locale);
	}
```

johanmalm:
> I think I've reached the point where the implementation is mostly complete with only small fixes needed here and there. I took a close look at @Consolatis' review branch and tried to improve on it, adding logic for escaping doubled underscores and using the UTF-8 locale while reading. I didn't use glib for the latter, but rather the `uselocale()` function two times.
> 
> Is this a good way of doing this?
> 
> ```c
> 	/* Temporarily set locale to UTF-8 */
> 	locale_t utf8_locale = newlocale(LC_CTYPE_MASK, "C.UTF-8", (locale_t)0);
> 	locale_t old_locale = (locale_t)0;
> 	if (utf8_locale != (locale_t)0) {
> 		old_locale = uselocale(utf8_locale);
> 	}
> 
> 	/* Reading logic... /*
> 
> 	/* Restore previous locale */
> 	if (utf8_locale != (locale_t)0) {
> 		uselocale(old_locale);
> 		freelocale(utf8_locale);
> 	}
> ```

I haven't seen @Consolatis for a few days, so let me respond to this.

I haven't used `uselocale()` and `newlocale()` before, but they seem a good choice here and I have no issue with using this approach.

What is your rationale for not using the glib functions? As we link to glib anyway, they can provide useful functions for things like this. I have resorted to using `g_utf8_casefold()` before `strcmp()` where `strcasecmp()` has been inadequate.

ch3rn1ka:
> What is your rationale for not using the glib functions? 

[This part](https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#the-use-of-glib) of the contributing guidelines. The code didn't seem too verbose to use glib helper functions, so I used the simplest tool available. Plus, I like that the `get_unicode_char()` function's side effects are clearly visible when written like this.

Consolatis:
Sorry for the delayed response.
I've rebased the PR and split it into two commits and added a bit of documentation.

Also slightly changed `scaled-font-buffer.c` as it was setting `use_markup` after calling `scaled_font_buffer_update()` which may create a new buffer immediately due to the `scaled_buffer_request_update()` call (unlikely to happen in the menu case though as we create the font buffer before showing it on screen and thus the delayed rendering would use the overwritten `use_markup`).

The CI error for Arch is unrelated to this PR (seems like Arch got a new gcc version which made the static analyzer a bit smarter).

Thanks for your work on this and being so responsive to change requests!
This PR is ready to go from my side but I'd like a second pair of eyes on the documentation.

johanmalm:
> This PR is ready to go from my side but I'd like a second pair of eyes on the documentation.

LGTM

Happy to documentation. Will go for the merge...

johanmalm:
@ch3rn1ka Thanks for your work on this. Great contribution :smile:

ch3rn1ka:
I'm happy to help! Thank you for your guidance :^)

PR #3507, opened 2026-04-11
[Feature] Override view inhibition for individual keybinds

drougas:
When working with virtual machines, VNC clients, etc, all keybinds are delivered to the view (for obvious reasons). For certain keybinds, (i.e. `XF86AudioMute`, or even `W-Tab`) that might not make sense. Currently, the solution is to bind the `ToggleKeybinds` action and then use that to switch all keybinds between labwc and the view (VNC / VM / other).

This adds a new `overrideInhibition` flag to `keyboard.keybinds` to make the above scenario easier for certain keybinds: When set, the keybind is always delivered to the compositor, not the view.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -760,7 +760,8 @@ References:
 	Stores the keyboard layout either globally or per window and restores
 	it when switching back to the window. Default is global.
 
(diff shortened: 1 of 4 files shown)


johanmalm:
I support this in principle and think it's a useful feature/contribution.

I haven't yet reviewed code, but will do tomorrow if nobody else gets there first.

johanmalm:
LGTM. Thanks.

PR #3511, opened 2026-04-14
osd-thumbnail: fix failed to get client texture

yuiiio:
Use wlr_client_buffer->texture directly instead of wlr_texture_from_buffer. All buffers in content_tree are wlr_client_buffer.

wlr_texture_from_buffer calls client_buffer_begin_data_ptr_access which fails when client_buffer->source == NULL (client released buffer early, e.g. wl_shm foot-terminal).

### DIFF
--- a/src/cycle/osd-thumbnail.c
+++ b/src/cycle/osd-thumbnail.c
@@ -2,6 +2,7 @@
 #include <assert.h>
 #include <wlr/render/allocator.h>
 #include <wlr/render/swapchain.h>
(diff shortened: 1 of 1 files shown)


yuiiio:
may fix https://github.com/labwc/labwc/issues/3161

Consolatis:
Thanks for the PR.

I think this is the correct approach. CC @tokyo4j

tokyo4j:
LGTM. I can't remember why I didn't do this.

Issue #688, opened 2022-12-20
Implement Focus Delay


wesbluemarine:
If you activate focus follows mouse and raise on focus, but don't have the delay, the newly created subwindows lose focus as soon as you move the mouse. 

**The solution should be to implement scope 2.2.4 "focusDelay"**

[This](https://user-images.githubusercontent.com/14043560/208787661-a7585599-5cc9-4e13-aa4e-174f5b93c984.webm) is an example video of the issue.

Consolatis:
As pointed out by @Flrian via IRC: https://gitlab.freedesktop.org/wayland/weston/-/issues/231 is likely related.

So instead of implementing the focus delay workaround we could also try to prevent ever having the toplevel window above a subwindow. I think we are actually doing something similar for X11 applications already, e.g. raising all the sub windows when raising the toplevel. Regardless of us implementing one solution or the other (or both), this should really be fixed one way or the other because as of right now `followMouse` + `raiseOnFocus` is unusable in my opinion.

johanmalm:
Solved by #3513

PR #3513, opened 2026-04-15
focus: add raiseOnFocusDelay option

joske:
Adds `<focus><raiseOnFocusDelay> (ms)`. Every WM I've ever used had this feature, so it was quite surprising to me this feature was missing.

It defers raise-on-focus by that amount; zero or unset keeps the current behaviour. Use case is followMouse=yes + raiseOnFocus=yes, where a brief cursor pass over an intermediate window causes a raise that was never intended — a small delay (I run 500ms) lets the cursor settle on  the target first. 

Related to but not the same as #688, which is about focus delay. This PR is about the auto-raise.

The pending state is a single slot on `struct server` (view pointer + `wl_event_source_timer`). Any new focus change supersedes the previous one: raise=true reschedules, raise=false cancels. View destroy cancels to avoid a dangling pointer at timer fire, and config reload cancels in case the feature was just turned off. Timer is reused across reschedules via `wl_event_source_timer_update`, same pattern as the configure-timeout in `xdg.c`.

Split into six small commits for bisect-ability. Been running on my daily driver (0.9.6 + these patches) for a bit, works as expected.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -493,6 +493,13 @@ this is for compatibility with Openbox.
 *<focus><raiseOnFocus>* [yes|no]
 	Raise window to top when focused. Default is no.
 
+*<focus><raiseOnFocusDelay>* [milliseconds]
+	When raiseOnFocus is enabled, delay the actual raise by this many
+	milliseconds. Default is 0 (raise immediately). A subsequent focus
+	change before the timer elapses restarts or cancels the pending raise.
+	Useful together with followMouse to avoid brief passes of the cursor
+	stacking up z-order changes.
+
 ## WINDOW SNAPPING
 
 Windows may be "snapped" to an edge or user-defined region of an output when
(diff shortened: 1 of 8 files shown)


johanmalm:
Thanks for this contribution. It intuitively feels like a nice UX improvement.

At a glance the code looks tidy, but will try to take a thorough look + test Thu/Fri unless someone else gets there first :smile:

johanmalm:
Have tested and it all works well, except one thing: when window switching (using alt-tab to trigger `NextWindow` action), there is a delay before the selected window is raised.

Potentially this could be dealt with by passing `bool force/immediate` to `schedule_auto_raise()`.

Other than that, LGTM and happy to merge.

joske:
Thanks, will try to fix that.

joske:
~~Note: I was not able to reproduce this (labwc + dms shell), but I limited the delay to the sloppy focus code path anyway.~~

joske:
Oh, I noticed I CAN reproduce the issue, it's while you keep alt-tab pressed, the window raises immediately but when you release, it goes back to the back for `delay` ms before coming back up. The above commit fixes that.

johanmalm:
LGTM. Thanks.

PR #3514, opened 2026-04-15
[backport] osd-thumbnail: fix failed to get client texture

Consolatis:
Use wlr_client_buffer->texture directly instead of wlr_texture_from_buffer. All buffers in content_tree are wlr_client_buffer.

wlr_texture_from_buffer calls client_buffer_begin_data_ptr_access which fails when client_buffer->source == NULL (client released buffer early, e.g. wl_shm foot-terminal).

---

Backport of #3511


johanmalm:
Thanks

PR #3519, opened 2026-04-17
[Show-and-tell]: What-if LabWC had a Sway-Compatible IPC Interface 

st0rm-shad0w:
### Summary

This PR is a working demo that adds a Unix socket-based IPC interface to LabWC, providing Sway-compatible communication for scripting, automation, and custom panel/bar integration.

> Please note, this code implementation is 99% synthesized using AI (GLM/Claude). I'm an ML engineer, and C is not my primary language. So, please be kind during the inspection of these welds.

### Motivation

Currently, LabWC lacks a standardized interface for external programs to interact with the compositor. This limits users to:

- **No scripting**: Cannot programmatically control windows, workspaces, or compositor behavior
- **Limited panel/bar options**: Tools like Quickshell, AGS, or custom panels cannot read compositor state directly

This IPC implementation enables:

- **Shell scripting** for window management and automation
- **Custom panels/bars** that can read workspace, window, and output state

### Technical Design

| Aspect | Implementation |
|--------|----------------|
| **Transport** | Unix domain socket (`AF_UNIX`, `SOCK_STREAM`) |
| **Protocol** | Sway/i3-compatible wire format |
| **Socket Path** | `$XDG_RUNTIME_DIR/labwc-ipc.<pid>.sock` |
| **Data Format** | JSON payloads with 17-byte header |
| **Magic String** | `labwc-ipc` (distinguishes from sway) |

### Supported Operations

**Commands:**
- `exec`, `exit`, `reload`, `kill`
- `fullscreen`, `floating`, `sticky`, `border`
- `workspace` switching
- `move` (position, workspace, output, directional)
- `resize` (absolute, grow/shrink)
- Criteria-based targeting: `[app_id="..."]`, `[title="..."]`, `[class="..."]`

**Queries:**
- `get_workspaces`, `get_outputs`, `get_inputs`, `get_seats`
- `get_tree` (full scene hierarchy)
- `get_version` (labwc version + loaded config)
- `get_config` (current config file contents, this is a bit broken, doesn't print pretty) 

**Events (subscribeable):**
- `workspace` - focus changes
- `output` - creation/destruction
- `window` - new, close, focus, title, move, resize
- `shutdown`, `tick`

### Client Tool: `labmsg`

A reference client is provided for testing and scripting:

```bash
# Query state
labmsg -t get_workspaces
labmsg -t get_tree -p

# Run commands
labmsg -t run_command "fullscreen toggle"
labmsg -t run_command '[app_id="firefox"] move workspace 2'

# Monitor events
labmsg -m -t subscribe '["window","workspace"]'
```

### Acknowledgments

- **Consolatis** - For entertaining the concept and allowing PR submission. Also for multiple code reviews and assistance with fine-tuning/polishing.
- **Malm** - For encouraging to share the code here, for broader feedback from core project devs.

### DIFF
--- a/clients/labmsg.c
+++ b/clients/labmsg.c
@@ -1,0 +1,389 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * labmsg - IPC client for labwc (swaymsg-compatible interface)
+ */
+#define _POSIX_C_SOURCE 200809L
+#include <errno.h>
+#include <fcntl.h>
+#include <getopt.h>
+#include <json-c/json.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <unistd.h>
+
+#define IPC_MAGIC "labwc-ipc"
+#define IPC_MAGIC_LEN 9
+#define IPC_HEADER_SIZE (IPC_MAGIC_LEN + 4 + 4)
+
+/* Message types */
+enum ipc_msg_type {
+	IPC_RUN_COMMAND = 0,
+	IPC_GET_WORKSPACES = 1,
+	IPC_SUBSCRIBE = 2,
+	IPC_GET_OUTPUTS = 3,
+	IPC_GET_TREE = 4,
+	IPC_GET_BAR_CONFIG = 6,
+	IPC_GET_VERSION = 7,
+	IPC_GET_CONFIG = 9,
+	IPC_SEND_TICK = 10,
+	IPC_SYNC = 11,
+	IPC_GET_INPUTS = 100,
+	IPC_GET_SEATS = 101,
+};
+
+static const char *version_str = "labmsg 0.1";
+
+static const struct option long_options[] = {{"help", no_argument, NULL, 'h'},
+	{"monitor", no_argument, NULL, 'm'}, {"pretty", no_argument, NULL, 'p'},
+	{"quiet", no_argument, NULL, 'q'}, {"raw", no_argument, NULL, 'r'},
+	{"socket", required_argument, NULL, 's'},
+	{"type", required_argument, NULL, 't'},
+	{"version", no_argument, NULL, 'v'}, {0, 0, 0, 0}};
+
+static const char usage_str[] =
+	"Usage: labmsg [options] [message]\n"
+	"\n"
+	"Options:\n"
+	"  -h, --help           Show help and exit\n"
+	"  -m, --monitor        Monitor for events (subscribe mode)\n"
+	"  -p, --pretty         Force pretty-printed JSON output\n"
+	"  -q, --quiet          Suppress response output\n"
+	"  -r, --raw            Force raw JSON output\n"
+	"  -s, --socket <path>  Override socket path\n"
+	"  -t, --type <type>    Message type (default: run_command)\n"
+	"  -v, --version        Show version and exit\n"
+	"\n"
+	"Message types:\n"
+	"  run_command, get_workspaces, subscribe, get_outputs,\n"
+	"  get_tree, get_bar_config, get_version, get_config,\n"
+	"  send_tick, get_inputs, get_seats\n";
+
+static int
+parse_msg_type(const char *name)
+{
+	if (!name || !strcmp(name, "run_command")) {
+		return IPC_RUN_COMMAND;
+	} else if (!strcmp(name, "get_workspaces")) {
+		return IPC_GET_WORKSPACES;
+	} else if (!strcmp(name, "subscribe")) {
+		return IPC_SUBSCRIBE;
+	} else if (!strcmp(name, "get_outputs")) {
+		return IPC_GET_OUTPUTS;
+	} else if (!strcmp(name, "get_tree")) {
+		return IPC_GET_TREE;
+	} else if (!strcmp(name, "get_bar_config")) {
+		return IPC_GET_BAR_CONFIG;
+	} else if (!strcmp(name, "get_version")) {
+		return IPC_GET_VERSION;
+	} else if (!strcmp(name, "get_config")) {
+		return IPC_GET_CONFIG;
+	} else if (!strcmp(name, "send_tick")) {
+		return IPC_SEND_TICK;
+	} else if (!strcmp(name, "get_inputs")) {
+		return IPC_GET_INPUTS;
+	} else if (!strcmp(name, "get_seats")) {
+		return IPC_GET_SEATS;
+	}
+	fprintf(stderr, "Unknown message type: %s\n", name);
+	return -1;
+}
+
+static int
+ipc_connect(const char *socket_path)
+{
+	int fd = socket(AF_UNIX, SOCK_STREAM, 0);
+	if (fd < 0) {
+		perror("socket");
+		return -1;
+	}
+
+	struct sockaddr_un addr = {0};
+	addr.sun_family = AF_UNIX;
+	strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path) - 1);
+
+	if (connect(fd, (struct sockaddr *)&addr, sizeof(addr)) < 0) {
+		perror("connect");
+		close(fd);
+		return -1;
+	}
+	return fd;
+}
+
+static bool
+ipc_send(int fd, uint32_t type, const char *payload, uint32_t len)
+{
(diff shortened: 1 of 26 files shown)


Consolatis:
I assume the post was written with the help of an LLM. Some points I still like to clarify though:
> Currently, LabWC lacks a standardized interface for external programs to interact with the compositor.

That "standardized" interface is in fact wayland protocols. Its the whole point of a wayland compositor like labwc. Using the term "standardized interface" as selling point for a custom IPC system feels dishonest and its actually exactly the opposite.

> This limits users to:
>  * **No scripting**: Cannot programmatically control windows, workspaces, or compositor behavior

There are some wayland protocols and applications which allow for some of that, namely wlr-foreign-toplevel to minimize / maximize / fullscreen / focus / close windows (example client `wlrctl`) and ext-workspaces to select the active workspace (and technically add new workspaces or remove existing ones, although that is not implemented in labwc). Further wlr-output-management and wlr-output-power-management which allows for tools like `wlr-randr`, `kanshi` or `wlopm` to control outputs. Furthermore there are actions like `If` and `ForEach` which allow for some scripting within labwc itself.

>  * **Limited panel/bar options**: Tools like Quickshell, AGS, or custom panels cannot read compositor state directly

Its not clear what exactly is meant with "compositor state".

> This IPC implementation enables:
> Shell scripting for window management and automation

Depending on what exactly is needed the existing protocols may already allow for that.

> Custom panels/bars that can read workspace, window, and output state

There is still a missing link between windows and workspaces (which will change at some point) but outside of that panels do already have access to the most relevant information and actions as can be seen by multiple examples like `xfce4-panel`, `lxqt-panel`, `sfwbar`, `waybar` and various others.

johanmalm:
I find this very interesting. I think a good next step would be evaluate what the benefits of this would actually be. In other words:

1. What problems does it solve that could not sensible/credibly be achieved without this - and where there is no protocol in sight to solve it.
2. Same as (1) but where the benefit is to bridge a gap until a protocol is available in `$UNKNOWN_TIME_PERIOD`

Examples might include things that could be achieved `labmsg -t subscribe -m '["window"]'` and `labmsg -t get_tree` like

- pagers needing window x/y/width/height
- panel workspace functionality (sfwbar) and possibly lxqt-panel (which uses some non-standard IPC already - can't remember details) - will be solved by protocol eventually
- client based window-switching / expo layout needing width/height - e.g. #3526

@tokyo4j @jlindgren90 - It would be good to get your initial feel on this too. Any of a..e would be helpful.

a. No. Just feels like the wrong direction - let's just stick with standard stuff. No reason to justify further - just keen to gauge.
b. We ought to consider supporting some of this functionality that standard protocols are just unlikely to ever do (like window x/y/w/h), but sway-IPC is just the wrong route. Prefer custom-wayland-protocols.
c. Possibly, if net benefits can be demonstrated and it's optional at build time. sway-IPC is some sort of standard after all having existed for a long time.
d. Possibly, if optional via `rc.xml`
e. 100% yes. I'm already clear on use-cases with clear benefits that apply to large user communities.

LBCrion:
If I may pipe in with a few thoughts on this idea from the perspective of having implemented support for a number of compositor side channel IPCs.

The most immediate features brought in by a side channel IPC are the full window/workspace/output hierarchy availability. These are going to be addressed by the upcoming wayland protocols in the future, but there is a lot of demand from users to have support for these now (the question of toplevel-workspace mappings on labc comes up regularly in sfwbar issues).

Some functionality offered by i3/sway IPC is unlikely to make it into wayland protocols and I haven't seen proposals for standard protocols for these elsewhere (i.e. dbus): 
1. Keyboard layout switching.
2. Spawning processes via compositor.
3. Window geometry.

Of these, keyboard layout control is something users generally want to see. Labwc exposes current layout via wl_keyboard.keymap, and the layout can be changed by applying new configuration to the compositor, but exposing a list of available layouts would be nice and i3/sway IPC allows this reasonably elegantly. (albeit I don't think this PR implements it currently).

Spawning processes via compositor is also very useful. Application menu is a common feature for 3rd party programs (i.e. taskbars). But if the programs are launched by a taskabr and user restarts a taskbar, all programs it launched are terminated. Spawning via compositor provides a cleaner solution (`exec` command in the PR).

The applications of  the window geometry data are somewhat less obvious. Currently sfwbar uses this info to populate a preview of workspaces and provide semi-intelligent placement of new floating windows. The first will likely be superseded by workspace copy capture protocol and the later belongs in the compositor, not the taskbar. I've only implemented it because I needed this functionality and all compositors plopped floating window in the center of the display.

Having said all this, i3/sway IPC is designed for tiling compositors. If labwc were to adapt it, IPC would need to be extended at least to cover minimize functionality (both action and state variable). Currently there is no way to map between IPC toplevel IDs and wlr-foreign-toplevel handles, so clients would need to rely on the IPC as a wholesale replacement for the protocol.

Also, window geometry is currently exposed via get_tree request, but isn't updated when floating windows are moved. For clients to track the state of floating window geometry, a window event on geometry changes would be very useful (i3/sway only emits it on changes to tiling geometry).

johanmalm:
@LBCrion Thanks for the context there.

I think @st0rm-shad0w is working on a description of use-cases and net-benefits as a first step.

Step 2 will then be to consider which 'requirements' cannot be met with Wayland protocols. Step 3, is to determine if we live with the short-coming to develop/merge a solution. Sway-IPC could be a solution, but it could also be something like a cosmic-protocol.

I will try to draft a philosophy document over the next few days to give some structure to this type of conversation. I'd rather look after the project's long term coherence rather than rushing into something that is clearly outside of our explicit scope - as it's currently stated - just because it provides a solution to a a use-case.

jlindgren90:
I think we should try hard to design any IPC interfaces with an eye toward making them reusable under other compositors as well. This PR (at least from the initial description) seems to do the opposite, proposing a whole new set of labwc-specific operations. It would encourage development of new clients that work *only* with labwc.

As a concrete suggestion for this PR, I think that *if* we're going to clone sway's IPC feature, it should be compatible so that existing sway clients work, not "similar but incompatible" so that a whole new set of clients need to be written.

Another option could be an "x-labwc" wayland protocol defined where we prototype new stuff with an intent to later propose it to wlroots or wayland-protocols. But before adding anything new to it, I would always encourage some research to see if any other compositor has already implemented something similar and if we can make it compatible.

danielfikko:
This PR addresses two of the use cases mentioned in https://github.com/labwc/labwc/discussions/3533. It would be a great addition!

jlindgren90:
Getting the focused/active window is exactly a use-case that doesn't need a new interface/protocol. wlr-foreign-toplevel exists for that already. A CLI tool could be written (if it doesn't already exist) using that protocol, and it would work with any wlroots compositor.

danielfikko:
That's true. I hadn't thought of that. `lswt` can get the active/focused window using `lswt -c A --force-protocol zwlr-foreign-toplevel-management-unstable-v1`. However, `zwlr-foreign-toplevel-management-unstable-v1` doesn't support the window's unique identifier, only `ext-foreign-toplevel-list-v1` does. And `app_id` is too generic (there could be, for example, two `alacritty` windows). Furthermore, one of the use cases in https://github.com/labwc/labwc/discussions/3533 requires the window's `PID`.

jlindgren90:
> However, zwlr-foreign-toplevel-management-unstable-v1 doesn't support the window's unique identifier, only ext-foreign-toplevel-list-v1 does. And app_id is too generic (there could be, for example, two alacritty windows). Furthermore, one of the use cases in #3533 requires the window's PID.

Let's improve/extend the existing protocol then, to provide that information, rather than bypassing it and creating a new one.

It might possibly be more work up-front, but improves interoperability since then the tool can (in future) work with other compositors as well.

stefonarch:
Just to show which features can only be enabled with a IPC backend  ( here Wayfire) in lxqt-panel:

immagine

immagine

johanmalm:
@stefonarch Does `pcmanfm-qt --desktop` support custom non-standard Wayfire IPC?

Consolatis:
> Just to show which features can only be enabled with a IPC backend ( here Wayfire) in lxqt-panel:

- `Show only windows from panel's screen` is possible with `wlr-foreign-toplevel-management`, its has `output_enter` and `output_leave` events.
- `Show only minimized windows` is also possible
- `Scroll up/down to move to next/previous desktop/workspace` is possible with the `ext-workspace` protocol

What is missing is window <-> workspace interop, e.g. `show only windows from desktop x`, `raise minimized windows on current desktop` and `send window to desktop x`.

PR #3525, opened 2026-04-18
interactive: allow resize on fully maximized views

### LINKED ISSUE #3524: Modifier+right-drag resize is ignored on fully maximized windows
### Problem description

When a window is maximized in both axes, initiating interactive resize via the window-modifier+right-drag (Resize) mousebind does nothing - the drag is silently consumed without resizing. The similar bind W-Left Drag (Move) works on the same maximized window and un-maximizes it as expected.

I'm coming to labwc from Xfce and run into this issue on a daily basis when I try to reduce the size of a maximized window. Being forced to un-maximize it manually first seems annoying, so I hope this is just an oversight.

I'll open a PR to change this.

The issue is caused by the early-return at `src/interactive.c:126-132` (`view->maximized == VIEW_AXIS_BOTH`), which happens before the axis-specific un-maximize logic from #3043.

### Steps to reproduce

1. Launch labwc with default configuration.
2. Open any window (e.g. foot).
3. Maximize it (default W-Up or the maximize button on the titlebar).
4. Hold Super (the window modifier) and right-click-drag anywhere on the window.

Expected behavior: the window un-maximizes and resizes according to the drag, consistent with W-Left Drag (Move).

Actual behavior: nothing happens, the drag is ignored.

### labwc build source

Release

### labwc version

0.9.6 (also reproducible on master, last checked c5ea41e8)

### labwc environment

### Distribution

Fedora 43

bjorn:
Modifier+right-drag resize was silently ignored on fully maximized views because of an early-return guard in `interactive_begin()`. The axis-specific un-maximization logic introduced in #3043 already handles partial maximization correctly; extend that to the `VIEW_AXIS_BOTH` case so both axes are cleared while keeping the current geometry as the starting point of the resize.

Move already permits dragging maximized views, so this also removes an asymmetry between the move and resize paths.

Fixes: #3524

### DIFF
--- a/src/interactive.c
+++ b/src/interactive.c
@@ -123,11 +123,12 @@ interactive_begin(struct view *view, enum input_mode mode, enum lab_edge edges)
 		cursor_shape = LAB_CURSOR_GRAB;
 		break;
 	case LAB_INPUT_STATE_RESIZE: {
-		if (view->shaded || view->fullscreen ||
-				view->maximized == VIEW_AXIS_BOTH) {
+		if (view->shaded || view->fullscreen) {
 			/*
-			 * We don't allow resizing while shaded,
-			 * fullscreen or maximized in both directions.
+			 * We don't allow resizing while shaded or fullscreen.
+			 * Maximized views are handled below by un-maximizing
+			 * the axes being resized while keeping the current
+			 * geometry as the starting point.
 			 */
 			return;
 		}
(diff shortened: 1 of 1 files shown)


src/interactive.c
```diff
@@ -123,11 +123,12 @@ interactive_begin(struct view *view, enum input_mode mode, enum lab_edge edges)
...
-		if (view->shaded || view->fullscreen ||
-				view->maximized == VIEW_AXIS_BOTH) {
+		if (view->shaded || view->fullscreen) {
 			/*
-			 * We don't allow resizing while shaded,
-			 * fullscreen or maximized in both directions.
+			 * We don't allow resizing while shaded or fullscreen.
+			 * Maximized views are handled below by un-maximizing
+			 * the axes being resized while keeping the current
+			 * geometry as the starting point.
```
bjorn, line 131: Not sure if the extra information is useful here, we could just leave it at this:
```suggestion
			 * We don't allow resizing while shaded or fullscreen.
```

johanmalm:
@jlindgren90 Would you be okay with reviewing this one?

jlindgren90:
I can't think of any reason why this shouldn't work. I did not test it personally, but if it's working, then let's go for it.

johanmalm:
Thanks. Okay. I will give it a test tomorrow with a view to merge

johanmalm:
LGTM. Have tested. Let's go for the merge.

Thanks :smile: Appreciate the contribution.

PR #3526, opened 2026-04-19
add overview based on kwin's placement algorithm

yuiiio:
kwin/src/plugins/private/expolayout.cpp

kwin's overview placement algorithm first sorts by window height(2nd key center_y),

turning it into a one-dimensional row partitioning problem.

(limit binary search iteration to guarantee termination)

Then it sorts within the rows based on window center_x.

(first sort fallback rev creation_id
for stablility)

20260419_12h05m19s_grim


Consolatis:
Thanks for the PR. However, I think this is better suited for an external application now that the toplevel capture protocol has been merged. That has the benefit that it works on multiple compositors and decouples the desktop shell from the compositor.

johanmalm:
> Thanks for the PR. However, I think this is better suited for an external application now that the toplevel capture protocol has been merged. That has the benefit that it works on multiple compositors and decouples the desktop shell from the compositor.

Yes, I was thinking the same. 

Would an external client we need with width/height of windows though?

Consolatis:
> With an external client we need with width/height of windows though?

Client width and height (and contents of the window) can be retrieved via toplevel capture.

yuiiio:
Here, we want to position the thumbnails according to the window's position, 
so we need at least the sorted results for center_x and center_y separately.

Currently, it seems that only `cosmic-toplevel-info` can communicate the window position to the client.

I agree that it's desirable to keep labwc's functionality as minimal as possible, so having a separate client is preferable.

ffs-akl:
@yuiiio ,  How about refining the WindowSwitcher UI with a more modern thumbnail layout? We can make the thumbnails more visually appealing, which would be a great enhancement to the existing switcher.

PR #3527, opened 2026-04-19
Add more details customization options for labnag

st0rm-shad0w:
Description:
This PR adds environment variable support for appearance configuration in labnag, allowing users to avoid repeatedly redefining options. It also introduces border color and margin options for the details section.

Define ENV variables like:

```
export LABNAG_TEXT_COLOR=282828 \
LABNAG_BACKGROUND_COLOR=B7BDF7 \
LABNAG_BORDER_BOTTOM_SIZE=1 \
LABNAG_BORDER_BOTTOM_COLOR=000000 \
LABNAG_DETAILS_BACKGROUND_COLOR=9B8EC7 \
LABNAG_DETAILS_BORDER_SIZE=1 \
LABNAG_DETAILS_BORDER_COLOR=282828ff \
LABNAG_DETAILS_BORDER_MARGIN=20 \
LABNAG_BUTTON_BORDER_SIZE=0 \
LABNAG_BUTTON_BORDER_COLOR=B7BDF7 \
LABNAG_BUTTON_BACKGROUND_COLOR=7C719F \
LABNAG_BUTTON_TEXT_COLOR=FFFFFF \
LABNAG_BUTTON_PADDING=7 \
LABNAG_MESSAGE_PADDING=7 \
LABNAG_FONT=CozetteVector
```

**Image:** Custom border/ margin for detail message section

20260419_11h25m13s_grim

### DIFF
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -447,7 +447,7 @@ render_button(cairo_t *cairo, struct nag *nag, struct button *button,
 	}
 
 	button->x = *x - border - text_width - padding * 2 + 1;
-	button->y = (int)(ideal_height - text_height) / 2 - padding + 1;
+	button->y = (int)(ideal_height - text_height) / 2 - padding;
 	button->width = text_width + padding * 2;
 	button->height = text_height + padding * 2;
 
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -375,7 +375,7 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
 	bool show_buttons = nag->details.offset > 0;
 	int button_width = get_detailed_scroll_button_width(cairo, nag);
 	if (show_buttons) {
-		nag->details.width -= button_width;
+		nag->details.width += border - button_width;
 		pango_layout_set_width(layout,
 				(nag->details.width - padding * 2) * PANGO_SCALE);
 	}
@@ -388,7 +388,7 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
 
 			if (!show_buttons) {
 				show_buttons = true;
-				nag->details.width -= button_width;
+				nag->details.width += border - button_width;
 				pango_layout_set_width(layout,
 					(nag->details.width - padding * 2) * PANGO_SCALE);
 			}
@@ -408,16 +408,17 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
 
 	if (show_buttons) {
 		nag->details.button_up.x = nag->details.x + nag->details.width;
-		nag->details.button_up.y = nag->details.y;
+		nag->details.button_up.y = nag->details.y - border;
 		nag->details.button_up.width = button_width;
-		nag->details.button_up.height = nag->details.height / 2;
+		nag->details.button_up.height = (border_rect_height + border) / 2;
 		render_details_scroll_button(cairo, nag, &nag->details.button_up);
 
 		nag->details.button_down.x = nag->details.x + nag->details.width;
 		nag->details.button_down.y =
 			nag->details.button_up.y + nag->details.button_up.height;
 		nag->details.button_down.width = button_width;
-		nag->details.button_down.height = nag->details.height / 2;
+		nag->details.button_down.height =
+			border_rect_height - nag->details.button_up.height;
 		render_details_scroll_button(cairo, nag, &nag->details.button_down);
 	}
 
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -46,6 +46,7 @@ struct conf {
 	uint32_t button_text;
 	uint32_t button_background;
 	uint32_t details_background;
+	uint32_t details_border_color;
 	uint32_t background;
 	uint32_t text;
 	uint32_t button_border;
@@ -60,6 +61,7 @@ struct conf {
 	ssize_t button_gap_close;
 	ssize_t button_margin_right;
 	ssize_t button_padding;
+	ssize_t details_margin;
 };
 
 struct pointer {
@@ -343,8 +345,9 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
 	uint32_t width = nag->width;
 
 	int border = nag->conf->details_border_thickness;
+	int margin = nag->conf->details_margin;
 	int padding = nag->conf->message_padding;
-	int decor = padding + border;
+	int decor = margin + border;
 
 	nag->details.x = decor;
 	nag->details.y = y + decor;
@@ -401,6 +404,8 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
 
 	nag->details.visible_lines = pango_layout_get_line_count(layout);
 
+	int border_rect_height = nag->details.height + 2 * border;
+
 	if (show_buttons) {
 		nag->details.button_up.x = nag->details.x + nag->details.width;
 		nag->details.button_up.y = nag->details.y;
@@ -415,6 +420,11 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
 		nag->details.button_down.height = nag->details.height / 2;
 		render_details_scroll_button(cairo, nag, &nag->details.button_down);
 	}
+
+	cairo_set_source_u32(cairo, nag->conf->details_border_color);
+	cairo_rectangle(cairo, margin, nag->details.y - border,
+			nag->details.width + 2 * border, border_rect_height);
+	cairo_fill(cairo);
 
 	cairo_set_source_u32(cairo, nag->conf->details_background);
 	cairo_rectangle(cairo, nag->details.x, nag->details.y,
@@ -1464,14 +1474,16 @@ conf_init(struct conf *conf)
 	conf->keyboard_focus = ZWLR_LAYER_SURFACE_V1_KEYBOARD_INTERACTIVITY_NONE;
 	conf->bar_border_thickness = 2;
 	conf->message_padding = 8;
-	conf->details_border_thickness = 3;
 	conf->button_border_thickness = 3;
 	conf->button_gap = 20;
 	conf->button_gap_close = 15;
 	conf->button_margin_right = 2;
 	conf->button_padding = 3;
 	conf->button_background = 0x680A0AFF;
+	conf->details_margin = 11;
+	conf->details_border_thickness = 3;
 	conf->details_background = 0x680A0AFF;
+	conf->details_border_color = 0x680A0AFF;
 	conf->background = 0x900000FF;
 	conf->text = 0xFFFFFFFF;
 	conf->button_text = 0xFFFFFFFF;
@@ -1551,16 +1563,18 @@ nag_parse_options(int argc, char **argv, struct nag *nag,
 		TO_COLOR_BORDER_BOTTOM,
 		TO_COLOR_BUTTON_BG,
 		TO_COLOR_DETAILS,
+		TO_COLOR_DETAILS_BORDER,
 		TO_COLOR_TEXT,
 		TO_COLOR_BUTTON_TEXT,
 		TO_THICK_BAR_BORDER,
 		TO_PADDING_MESSAGE,
-		TO_THICK_DET_BORDER,
+		TO_THICK_DETAILS_BORDER,
 		TO_THICK_BTN_BORDER,
 		TO_GAP_BTN,
 		TO_GAP_BTN_DISMISS,
 		TO_MARGIN_BTN_RIGHT,
 		TO_PADDING_BTN,
+		TO_MARGIN_DETAILS,
 	};
 
 	static const struct option opts[] = {
@@ -1587,8 +1601,10 @@ nag_parse_options(int argc, char **argv, struct nag *nag,
 		{"button-text-color", required_argument, NULL, TO_COLOR_BUTTON_TEXT},
 		{"border-bottom-size", required_argument, NULL, TO_THICK_BAR_BORDER},
 		{"message-padding", required_argument, NULL, TO_PADDING_MESSAGE},
-		{"details-border-size", required_argument, NULL, TO_THICK_DET_BORDER},
+		{"details-border-size", required_argument, NULL, TO_THICK_DETAILS_BORDER},
 		{"details-background-color", required_argument, NULL, TO_COLOR_DETAILS},
+		{"details-border-color", required_argument, NULL, TO_COLOR_DETAILS_BORDER},
+		{"details-margin", required_argument, NULL, TO_MARGIN_DETAILS},
 		{"button-border-size", required_argument, NULL, TO_THICK_BTN_BORDER},
 		{"button-gap", required_argument, NULL, TO_GAP_BTN},
 		{"button-dismiss-gap", required_argument, NULL, TO_GAP_BTN_DISMISS},
@@ -1633,6 +1649,9 @@ nag_parse_options(int argc, char **argv, struct nag *nag,
 		"  --details-border-size size       Thickness for the details border.\n"
 		"  --details-background-color RRGGBB[AA]\n"
 		"                                   Details background color.\n"
+		"  --details-border-color RRGGBB[AA]\n"
+		"                                   Details border color.\n"
+		"  --details-margin margin          Margin for the details.\n"
 		"  --button-border-size size        Thickness for the button border.\n"
 		"  --button-gap gap                 Size of the gap between buttons\n"
 		"  --button-dismiss-gap gap         Size of the gap for dismiss button.\n"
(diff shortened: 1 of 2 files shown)


johanmalm:
I'm happy to add command-line options for detailed border color/thickness. I haven't fully reviewed it yet, but initial comments are:

1. `--details-border-margin` feels inconsistent general labwc language. Suggest `--details-border-thickness` instead. We try to use `margin` to refer to space outside an object; and padding for space inside.
2. The `-h` usage message has not been updated.

Have just tried it and I'm a bit confused now.

```
ls -1 | ./build/clients/labnag -l -m foo -Z 1 --details-border-size 100 --details-border-color 0000FFFF --details-border-margin 100
```
produces:

image

whereas

```
ls -1 | ./build/clients/labnag -l -m foo -Z 1 --details-border-color 0000FFFF --details-border-margin 100
```

...results in...

image

`details-border-color` seems to refer to the area between the border and the text.
`details-border-margin` is the thickness of the border.
The already existing `detailed-border-size` is the text margin (i.e. space between text and border).

I'm happy to add more configuration options, but I much prefer them to be well thought through and consistent. We have adopted the swaynag syntax, so there might be some stuff that could do with being re-designed - which I'm happy to do provided we stay backward compatible for a period. Would you be happy to give this some thought?

Regarding the environment variables, I prefer to not merge that because I think it adds complexity without significant wider benefit. `swaynag` supports a config file, but I removed that to purposefully keep `labnag` simple.

Regarding the environment variables, you could easily achieve something similar with a simple shell script like this:

```
#!/bin/sh
: ${LABNAG_BACKGROUND_COLOR=005500FF}
labnag --background-color ${LABNAG_BACKGROUND_COLOR} $*
```

With this type of approach you could of course also just parse a `key=value` style configuration file like this:

```
while IFS='=' read -r key val; do
    [ "${key##\#*}" ] || continue
    printf '%s: %s\n' "$key" "$val"
done < "file"
```

...so this approach gives users a lot of flexibility without `labnag` having to handle it within the .c code.

st0rm-shad0w:
Hi @johanmalm, Thanks for testing it out and the feedback. About your suggestion,  having the extra options would be really nice.

> -details-border-margin feels inconsistent general labwc language. Suggest --details-border-thickness instead. We try to use margin to refer to space outside an object; and padding for space inside. 

I actually followed the CSS box model for the layout nomenclature, where margin refers to the space outside the border line and details-border-size (the existing option) handles the thickness of the line itself. Skipped padding. Basically, I wanted to provide control over both the line thickness and the space outside. From my perspective, it’s working correctly, which seems corroborated by the screenshots you posted.

> The -h usage message has not been updated.

My bad! Totally missed expanding the new options for -h.

> Regarding the environment variables, I prefer to not merge that because I think it adds complexity without significant wider benefit. swaynag supports a config file, but I removed that to purposefully keep labnag simple.

This would have been nice-to-have! Goal here was just "define once, use everywhere". It keeps manual CLI commands short without the need to manage extra wrapper scripts. From an actual usage point of view, many users have felt the same way about wanting that convenience. However, I totally get  your perspective of keeping the codebase small as possible.

clients/labnag.c
```diff
@@ -294,29 +296,31 @@ render_message(cairo_t *cairo, struct nag *nag)
...
+		struct button *button, bool top_border)
 {
 	int text_width, text_height;
 	get_text_size(cairo, nag->conf->font_description, &text_width,
 		&text_height, NULL, 1, true, "%s", button->text);
 
-	int border = nag->conf->button_border_thickness;
-	int padding = nag->conf->button_padding;
+	int border = nag->conf->details_border_thickness;
+	int padding = nag->conf->button_padding * 0.3 + 2;
```
johanmalm, line 306: What does this line do? It is not described in the commit messages as a change.

clients/labnag.c
```diff
@@ -331,8 +335,8 @@ get_detailed_scroll_button_width(cairo_t *cairo, struct nag *nag)
 		NULL, 1, true, "%s", nag->details.button_down.text);
 
 	int text_width =  up_width > down_width ? up_width : down_width;
-	int border = nag->conf->button_border_thickness;
-	int padding = nag->conf->button_padding;
+	int border = nag->conf->details_border_thickness;
+	int padding = nag->conf->button_padding * 0.3 + 2;
```
johanmalm, line 339: I'm happy to review minor improvements like this, but I much prefer them in separate commits with an explanations of what they do and why.

johanmalm: A screenshot of before and after can be helpful to reduce review time.

clients/labnag.c
```diff
@@ -294,29 +296,31 @@ render_message(cairo_t *cairo, struct nag *nag)
 
 static void
 render_details_scroll_button(cairo_t *cairo, struct nag *nag,
-		struct button *button)
+		struct button *button, bool top_border)
```
johanmalm, line 299: It seems a weird API to have a `bool top_border` here. Suggest a cleaner way of approaching this.

clients/labnag.c
```diff
@@ -401,21 +408,29 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)
...
+		nag->details.button_up.height = (border_rect_height + border) / 2;
+		render_details_scroll_button(cairo, nag, &nag->details.button_up, true);
 
 		nag->details.button_down.x = nag->details.x + nag->details.width;
 		nag->details.button_down.y =
 			nag->details.button_up.y + nag->details.button_up.height;
 		nag->details.button_down.width = button_width;
-		nag->details.button_down.height = nag->details.height / 2;
-		render_details_scroll_button(cairo, nag, &nag->details.button_down);
+		nag->details.button_down.height =
```
johanmalm, line 424: There are lots of changes to the rendering of buttons. Could that be split out into a separate commit to make the logic changes easier to follow and review. Also, a description of what it does and why in the commit messages helps a lot.

clients/labnag.c
```diff
@@ -447,7 +462,7 @@ render_button(cairo_t *cairo, struct nag *nag, struct button *button,
 	}
 
 	button->x = *x - border - text_width - padding * 2 + 1;
-	button->y = (int)(ideal_height - text_height) / 2 - padding + 1;
+	button->y = (int)(ideal_height - text_height) / 2 - padding;
```
johanmalm, line 465: This change seems like a separate concern. Why is this being done?

clients/labnag.c
```diff
@@ -1561,6 +1597,7 @@ nag_parse_options(int argc, char **argv, struct nag *nag,
 		TO_GAP_BTN_DISMISS,
 		TO_MARGIN_BTN_RIGHT,
 		TO_PADDING_BTN,
+		TO_MARGIN_DET_BORDER,
```
johanmalm, line 1600: I can't see that `DETAILS` has been shortened to `DET` anywhere else. Would it be more consistent to stick with `DETAILS`?

clients/labnag.c
```diff
@@ -1589,6 +1626,8 @@ nag_parse_options(int argc, char **argv, struct nag *nag,
 		{"message-padding", required_argument, NULL, TO_PADDING_MESSAGE},
 		{"details-border-size", required_argument, NULL, TO_THICK_DET_BORDER},
 		{"details-background-color", required_argument, NULL, TO_COLOR_DETAILS},
+		{"details-border-color", required_argument, NULL, TO_COLOR_DETAILS_BORDER},
+		{"details-border-margin", required_argument, NULL, TO_MARGIN_DET_BORDER},
```
johanmalm, line 1630: I think `--details-border-margin` seems an odd thing to call this. I get the logic, but it feels inconsistent with everything else in `labwc` to say that the border has a margin.

Suggest just calling it `--details-margin` because it is the margin for the whole `details` area.

johanmalm (commented):
In principle, I'm happy with the new color and margin options.

johanmalm:
I wasn't sure if this was ready for final review, but I did one anyway.

Are you planning on dropping the env-var commit from this PR?

Screenshot below with CLI options marked up so I can remember it for next time I review this. 

labnag

clients/labnag.c
st0rm-shad0w: Referring to the attached image, applying a minimal patch for the details border color, size, and margin results in the appearance as shown in the topmost bar. However, setting the details border size to a very large value causes the stacked scroll button (on right side) to become hidden.

To resolve this issue and better integrate the scroll button into the details section, I adjusted its appearance/dimensions. Rather than controlling the details border and the scroll button thickness with two separate variable values, I used a single value, i.e. details border thickness  (Line 305). Additionally, I reduced the button's padding for a more compact appearance (Line 306). Basically, I intended to style these particular buttons a bit differently than the rest of the buttons in the UI. The result of this border change can be seen in the second bar. 

20260423_01h33m27s_grim

Minimal Patch - [minimal-change.patch](https://github.com/user-attachments/files/26985901/minimal-change.patch)

clients/labnag.c
st0rm-shad0w: Same comment as above, this change is just for the other details scroll  button.

clients/labnag.c
st0rm-shad0w: The `render_details_scroll_button` function helps render the individual detail scroll buttons. I introduced a new parameter, *top_border*, to its function signature that allows the button's top border to be hidden, which is specifically applied to the bottom scroll button. This ensures that the details border size remains consistent across all sides of the content, including the all sides of the buttons. You can visualize this change by comparing the second and third bars in the attached picture.

20260423_01h33m27s_grim

clients/labnag.c
st0rm-shad0w: The changes around line 424 ensure that the detail scroll buttons maintain consistent and even sizing, especially when large borders or margins are applied.

20260423_02h43m27s_grim

clients/labnag.c
st0rm-shad0w: You are right! This is probably the most unrelated change in this whole PR. It essentially removes an awkward top margin from the button on the main labnag bar. You can see the difference by comparing the first bar (without the +1) to the second bar (with the +1).

20260423_02h53m36s_grim

clients/labnag.c
st0rm-shad0w: Will update this in next iteration.

clients/labnag.c
st0rm-shad0w: Will update this in next iteration.

st0rm-shad0w:
Hi @johanmalm, thank you for the detailed review. I've provided more context to address your comments.

Because some of the UI rendering in my implementation is tightly coupled with the new options, I'm unsure of the best way to split the commits. How do you recommend dividing them up?

>  Are you planning on dropping the env-var commit from this PR?

I'd really like to keep it! :) But I'll go with whatever you think is best.

clients/labnag.c
johanmalm: Thanks for screenshot. Helps understand where the look is wrong. Using `* 0.3 + 2` for the padding feels like the wrong solution. Could we set a minimum size for the area that contains the arrow instead?

clients/labnag.c
johanmalm: > I introduced a new parameter, _top_border_, to its function signature that allows the button's top border to be hidden.

Understood - and I guessed as much. My challenge is that it's weird that this is needed. I suspect something is wrong in the pre-existing labnag code. It was probably written without extensibility in mind, so now that we're looking to extend it, I feel that we ought to fix it up first to avoid mortgaging technical debt.

clients/labnag.c
johanmalm: Okay. Understood. Fine.

clients/labnag.c
johanmalm: Okay. Suggest separate commit with what you've said here as the commit message.

johanmalm:
Have responded.

I don't want to merge the env-var commit.

Suggest doing fixup commits and then using `git rebase -i HEAD~n` where n is your total number of commits.

The more you structure the commits so that it makes sense to someone coming in cold, the better. In other words make them small; make one logical change per commit. Each commit should compile. It's a bit like students doing their Maths homework - show the work rather than giving the answer. The commit messages are important in understanding the reason why the change was made. There is rarely a need to explain the code - it's generally obvious to us as reviewers what it does. What is often not so obvious is why. This is also very useful when people later look back at what has changed.

clients/labnag.c
st0rm-shad0w: I updated the expression  to `int padding = (nag->conf->button_padding / 3) + 2;`. This keeps the default look identical, with the original minimum size of button_padding, i.e. 3. It would scale better to keep things compact for  larger padding  values. Also now, this is 100% integer math, and avoids any floating-point conversion overhead.

clients/labnag.c
st0rm-shad0w: Sorry, if I am missing something obvious, but I don't follow why the current approach is problematic. Could you elaborate on the drawbacks (for my learning)? Also, since I haven't  grasped the problem yet, tackling the fix is probably a bit beyond me right now."

clients/labnag.c
st0rm-shad0w: Updated.

st0rm-shad0w:
@johanmalm - Restructured all the commits as suggested above. Also dropped  env-var functionality.

johanmalm:
Thanks for all the work on this.

LGTM except one thing:

I don't like `bool draw_top_border` being passed to render_details_scroll_button(). Unless I completely misunderstand this function, it simply draws a rectangle with a border and then draws the text on top (i.e. the up or down arrow). Suggest that we instead simply position the top and bottom scroll buttons so that the adjacent borders overlap (i.e. the bottom border over up-button overlaps with the top border of the down-button.

Alternatively, just draw one big rectangle of color `--details-border-color` behind the up/down buttons - and then render the buttons on top without borders.

st0rm-shad0w:
If I undo the last commit (the one passing the `bool draw_top_border` parameter) and instead apply your first suggestion for overlapping adjacent borders, does the code below look correct?


```
[$] git diff HEAD --staged                                                                         
diff --git a/clients/labnag.c b/clients/labnag.c
index a3a526c5..8d47ee29 100644
--- a/clients/labnag.c
+++ b/clients/labnag.c
@@ -318,7 +318,7 @@ render_details_scroll_button(cairo_t *cairo, struct nag *nag,

        cairo_set_source_u32(cairo, nag->conf->button_text);
        cairo_move_to(cairo, button->x + border + padding,
-                       button->y + border + (button->height - text_height) / 2);
+                       button->y + (button->height - text_height) / 2);
        render_text(cairo, nag->conf->font_description, 1, true,
                        "%s", button->text);
 }
@@ -415,10 +415,10 @@ render_detailed(cairo_t *cairo, struct nag *nag, uint32_t y)

                nag->details.button_down.x = nag->details.x + nag->details.width;
                nag->details.button_down.y =
-                       nag->details.button_up.y + nag->details.button_up.height;
+                       nag->details.button_up.y + nag->details.button_up.height - border;
                nag->details.button_down.width = button_width;
                nag->details.button_down.height =
-                       border_rect_height - nag->details.button_up.height;
+                       border_rect_height - nag->details.button_up.height + border;
                render_details_scroll_button(cairo, nag, &nag->details.button_down);
        }
```

johanmalm:
> `button->y + border + (button->height - text_height) / 2);`

I don't think we should remove `border` in `render_details_scroll_button()`. The border is rendered within the rect - so the text position needs to take it into account.

The rest looks good.

Top tip: use semi-transparent colors to check the overlap of borders, etc. E.g. `#FF000080`

st0rm-shad0w:
@johanmalm -  Tried keeping `border`, it causes button text misalignment for larger border/ margin values. Seems to be working correct without it. Can I request you to check once more?

johanmalm:
I have just tried with `--details-border-color 00FFFF80` which looks odd. Maybe we've been thinking about this incorrectly.

I think I'm right in saying that fundamentally you're trying to add ability to draw the blue border/space in this picture.



It probably makes more sense to just draw one big blue rectangle, and then the three yellow rectangles with respective text inside.

Trying to draw the three yellow areas with functions like `render_details_scroll_button()` - which draw a rectangles + a border + text would work if colors were always opaque, but they are not.

Does that make sense?

st0rm-shad0w:
@johanmalm - Per our discussion on IRC, I'm pushing the changes from the diff above for further testing. 

Using a similar color combination, the rendering for the border, button, and button text looks good on my end. I've attached a screenshot for reference.

```
ls -la | build/clients/labnag -y overlay -e top -x --text-color 282828 --background-color FFFFFF --border-bottom-size 1 --border-bottom-color 000000 --button-background-color FFFF00 --button-border-size  0 --button-text-color 282828 --details-background-color FFFF00 --details-border-size 20 --details-border-color 0000F6 --details-margin 50 --button-padding 7 --message-padding 7 -f CozetteVector -m "\$ls -la" -l -L " (ﾉ◕ヮ◕)ﾉ*:・ﾟ✧ " --timeout 3600 & disown
```

20260428_02h38m35s_grim

johanmalm:
LGTM. Thanks.

PR #3528, opened 2026-04-19
desktop: refresh workspace appearance for workspace changes

g7fya:
Fix workspace state updates so panels such as sfwbar can correctly track workspace-filtered taskbars under labwc.

- remove restriction that skips workspace updates when the source is OSD
- trigger output_update_workspace_appearance() after workspace change
- return early when no view is present to avoid inconsistent updates

This resolves incorrect behavior where sfwbar (filter = workspace) does not update properly under labwc, while it works correctly on sway.

Related: https://github.com/LBCrion/sfwbar/issues/412

### DIFF
--- a/include/foreign-toplevel/foreign.h
+++ b/include/foreign-toplevel/foreign.h
@@ -8,6 +8,7 @@ struct foreign_toplevel;
 struct foreign_toplevel *foreign_toplevel_create(struct view *view);
 void foreign_toplevel_set_parent(struct foreign_toplevel *toplevel,
 	struct foreign_toplevel *parent);
+void foreign_toplevel_refresh(struct foreign_toplevel *toplevel);
 void foreign_toplevel_destroy(struct foreign_toplevel *toplevel);
 
 #endif /* LABWC_FOREIGN_TOPLEVEL_H */
--- a/src/foreign-toplevel/foreign.c
+++ b/src/foreign-toplevel/foreign.c
@@ -1,18 +1,66 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #include "foreign-toplevel/foreign.h"
 #include <assert.h>
+#include <stdbool.h>
 #include "common/mem.h"
 #include "foreign-toplevel/ext-foreign.h"
 #include "foreign-toplevel/wlr-foreign.h"
+#include "labwc.h"
 #include "view.h"
 
 struct foreign_toplevel {
+	struct view *view;
+
 	/* *-toplevel implementations */
 	struct wlr_foreign_toplevel wlr_toplevel;
 	struct ext_foreign_toplevel ext_toplevel;
-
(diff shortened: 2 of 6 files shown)


johanmalm:
Welcome.

Please could you describe the context for this PR? It is easier for me to evaluate how sensible it is to merge it upstream if I understand this. Obvious questions that come to mind for include the following:

1. Is it meant to a work-around until https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196 is merged upstream?
2. Removing the toplevel-handle from any non-omnipresent window which is not on the current workspace would affect all clients using the wlr-foreign-toplevel protocol. For example if you ran `lswt` or `wlrctl` you would get a different output. To what extent has this been considered and evaluated? Is it protocol compliant? If not, do we want to break compliance for this benefit. 
3. Has any thought been given to implementing this in a way that does not break backward compatibility for clients that use the wlr-foreign-toplevel protocol. For example, consider this script here: https://labwc.github.io/faq.html#run-or-raise
4. I have just tried `sfwbar` on `sway` with default configuration, but observe that the panel does not update when workspaces are switched. Are we sure this actually works on `sway`? https://github.com/LBCrion/sfwbar/issues/412 is not that clear on this claim. If does work, how is it achieved? 
5. Are there other solutions that should be considered - other than to destroy toplevel-handles and waiting for the upstream protocol?

LBCrion:
@johanmalm, sfwbar uses sway side channel ipc to track (and manipulate) toplevel-workspace assignments. It doesn't filter the toplevels by workspace by default,  but can do so when a taskbar option `filter = workspace` is set. On compositors with supported side channel IPCs, the bar remains aware of all toplevels. Any filtering is done by the taskbar. The proposed approach seems to achieve a very specific behaviour,  I.e. taskbar filtered by workspace.  It won't support other cases,  such as tint2 style multidesktop taskbar or moving toplevels across workspaces by dragging.

Consolatis:
This breaks tools like `wlrctl` and the ability for panels to show all windows, regardless of workspaces. It also prevents toplevel capture for windows not on the current workspace. The proper solution is [this](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/474) protocol in combination with either [this](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196) or [this](https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/merge_requests/141) one.

g7fya:
Thanks for the feedback and the detailed explanations.

I understand the concern about changing wlr-foreign-toplevel behavior and the potential impact on other tools like wlrctl and different panel use cases. My intention was mainly to address the workspace-filtered taskbar behavior I was seeing with sfwbar, but I see how this approach is too narrow and could introduce regressions.

I’ll keep this as a local workaround for now and revisit it if there’s a more protocol-aligned way to achieve the same result in the future.

Appreciate you taking the time to review and explain the reasoning.

PR #3529, opened 2026-04-19
ssd buttons: add always-on-top button

chocolatemintychip:
This PR adds an always-on-top titlebar button. Should be useful for those who use sticky windows a lot. Currently there is no visual indicator for always-on-top windows, so this feature also doubles as that.


johanmalm:
Thanks for contributing. 

Out of interest, are there any other compositor that support this? What do mutter (Gnome) and kwin (KDE) do in this regard?

chocolatemintychip:
Hi!

Not really sure how it's done in those as it's been a while since I last used them. I mainly based this feature on how it's done in Awesome. I also felt like since since sticky and ontop states are now decoupled, it's now worth adding this button.

chocolatemintychip:
Just a thought: what if the buttons were user-definable in the config? Right now it's possible to change bindings of an existing button, but doing so forces you to sacrifice its original purpose and to effectively just *use it wrong*. Also, that doesn't allow for custom button states. I am not familiar with the codebase, so I'd like to ask how feasible that is and if it'd be out-of-scope.

johanmalm:
Sorry it took a while to get back on this one. 

I'm happy that it is in scope.

It is okay to do `ssd_update_geometry(view->ssd)` in `view_toggle_always_on_top()`. Suggest removing comment.

I think that we ought to add screenshots of the proposed built-in ontop icons here (in both states) so that people can offer views and alternatives. Would you be happy to do that?

> Just a thought: what if the buttons were user-definable in the config? Right now it's possible to change bindings of an existing button, but doing so forces you to sacrifice its original purpose and to effectively just use it wrong. Also, that doesn't allow for custom button states. I am not familiar with the codebase, so I'd like to ask how feasible that is and if it'd be out-of-scope.

I suspect it would add significant complexity, but I might be wrong on that and sometimes things turn out to significantly simpler than I think :smile: Suggest researching that as a separate thread.

PR #3530, opened 2026-04-20
Add SendKey action for global shortcuts

elviosak:
as discussed in #3517 
this would allow configuration of applications to receive keyboard shortcuts when not in focus, like triggering start/stop recording in OBS without having to switch to its window.
```
<keybind key="W-g">
  <action name="SendKey" key="W-r" identifier="com.obsproject.Studio"/>
</keybind>
```
the "identifier" is the app_id or WM_CLASS depending on wayland/xwayland like on `<windowRules><windowRule identifier="">`, but its not working on xwayland, i don't know why.

the "key" is optional and the same format as `<keyboard><keybind key="">`, if it's not provided it will forward the currently pressed keys to the application.

i'll mark this as draft since i'm not sure if it's something the maintainers want to add to labwc and  i need to fix the xwayland issue and add documentation.
Also i don't like the code duplication, i copied and modified functions `keybind_update_keycodes` and `update_keycodes_iter` from `keybind.c` to `action.c`, not sure how i could make them more "generic" to be able to reuse for both cases.

edit: this being an action should also work on mousebinds and menu even without providing "key" it will send the currently pressed keys to the application.

### DIFF
--- a/include/config/keybind.h
+++ b/include/config/keybind.h
@@ -45,6 +45,7 @@ bool keybind_the_same(struct keybind *a, struct keybind *b);
 
 bool keybind_contains_keycode(struct keybind *keybind, xkb_keycode_t keycode);
 bool keybind_contains_keysym(struct keybind *keybind, xkb_keysym_t keysym);
+bool keybind_contains_any_keysym(struct keybind *keybind, const xkb_keysym_t *syms, int nr_syms);
 
 void keybind_update_keycodes(void);
 #endif /* LABWC_KEYBIND_H */
--- a/src/action.c
+++ b/src/action.c
@@ -8,6 +8,7 @@
 #include <unistd.h>
 #include <wlr/types/wlr_cursor.h>
 #include <wlr/types/wlr_scene.h>
+#include <wlr/types/wlr_seat.h>
 #include <wlr/util/log.h>
 #include "action-prompt-codes.h"
 #include "common/buf.h"
@@ -137,6 +138,7 @@ struct action_arg_list {
 	X(TOGGLE_SHOW_DESKTOP, "ToggleShowDesktop") \
 	X(WARP_CURSOR, "WarpCursor") \
 	X(HIDE_CURSOR, "HideCursor") \
+	X(SEND_KEY, "SendKey") \
 	X(DEBUG_TOGGLE_KEY_STATE_INDICATOR, "DebugToggleKeyStateIndicator")
 
 /*
@@ -531,6 +533,12 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			goto cleanup;
 		}
 		break;
+	case ACTION_TYPE_SEND_KEY:
+		if (!strcasecmp(argument, "identifier")) {
+			action_arg_add_str(action, "identifier", content);
+			goto cleanup;
+		}
+		break;
 	case ACTION_TYPE_IF:
 		if (!strcmp(argument, "message.prompt")) {
 			action_arg_add_str(action, "message.prompt", content);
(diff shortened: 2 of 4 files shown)


elviosak:
If it's preferrable, i could also remove the "key" argument to keep the changes to a minimum and still cover the majority of use-cases.

johanmalm:
Haven't reviewed the code in detail, but like this feature in principle. It feels like it solves global-shortcuts without needing xdg-desktop-portal, etc.

> If it's preferrable, i could also remove the "key" argument to keep the changes to a minimum and still cover the majority of use-cases.

I think it's important to have the `key` option because I can see that the key / key-combo that a user needs to send to a client will often be different to the key-bind itself, for example sending `<space>` to a video player or browser.

danielfikko:
One question: what about programs that use the `GlobalShortcuts` portal? Will an `xdg-desktop-portal-labwc` need to be created using this PR?

johanmalm:
> One question: what about programs that use the `GlobalShortcuts` portal? Will an `xdg-desktop-portal-labwc` need to be created using this PR?

Don't know. Appreciate help finding that out. Does sway support global shortcuts?

danielfikko:
> Don't know. Appreciate help finding that out. Does sway support global shortcuts?

`sway` will likely wait for a protocol:
https://github.com/swaywm/sway/issues/8062

elviosak:
> One question: what about programs that use the `GlobalShortcuts` portal? Will an `xdg-desktop-portal-labwc` need to be created using this PR?

Not sure, from what i read, when using the GlobalShortcuts portal u give a name to the shortcut and it can be triggered via dbus? If so it could be triggered by a regular keybind with action "Execute" i think. Or if labwc supports it internally, the name could be added as a new argument to "SendKey" too.

elviosak:
@johanmalm should this also handle push-to-talk use cases? i just realized it doesn't work correctly because it needs to send separate keypress and keyrelease.
Currently the only way i can think is to change keybind handling to differentiate keybinds by the `onRelease` arg in `keybind_the_same()` in `config/keybind.c` and match the press/release in `match_keybinding()` in `input/keyboard.c` and then add another arg to `SendKey` like `type=default|press|release` so push-to-talk would work by creating 2 keybinds and actions, one with `onRelease=yes` and another without, but also would need to do something about repeating keybind.
What do you think? Is there an easier/better way?

johanmalm:
> @johanmalm should this also handle push-to-talk use cases? i just realized it doesn't work correctly because it needs to send separate keypress and keyrelease. Currently the only way i can think is to change keybind handling to differentiate keybinds by the `onRelease` arg in `keybind_the_same()` in `config/keybind.c` and match the press/release in `match_keybinding()` in `input/keyboard.c` and then add another arg to `SendKey` like `type=default|press|release` so push-to-talk would work by creating 2 keybinds and actions, one with `onRelease=yes` and another without, but also would need to do something about repeating keybind. What do you think? Is there an easier/better way?

Sorry it took at while to respond. Traffic has been high recently. I don't use push-to-talk or similar myself, but it does seem a very valid use-case that ought to try to handle.

I think you're right in your analysis; it does not appear to be trivial. I suggest trying to finish+merge what we have in this PR first and then deal with press/release as a separate initiative. Key press/release logic is quite finicky, so want to keep commits as small as possible and do plenty of testing. A step at a time is preferred.

elviosak:
still need to figure out why xwayland is not working, i'm looking at hyprland's code and it looks like it's handled here https://github.com/hyprwm/Hyprland/blob/56d7a43102b53f79a2311662783d5dd94cb2f1a5/src/config/shared/actions/ConfigActions.cpp#L1487, i'll try to mirror it.

also i was thinking of replacing keybind onRelease with `action=press|release|both` like the mousebind
`<mousebind button="" direction="" action=""><action>`
but should we change action to something else to differentiate from `<action>`? I was thinking mouseAction/keyAction or trigger.

PR #3534, opened 2026-04-21
session: ship and activate labwc-session.target

joske:
Ships a tiny `labwc-session.target` (same shape as `miracle-wm-session.target`). Lets `WantedBy=graphical-session.target` services actually come up under labwc without needing uwsm or a hand-rolled wrapper.

Rationale: I am using dms with labwc, and the `dms.service` would never start. With other compositors under dms, it did work. Turned out that the `graphical-session.target` is never reached.

### DIFF
--- a/data/labwc-session.target
+++ b/data/labwc-session.target
@@ -1,0 +1,6 @@
+[Unit]
+Description=labwc session
+Documentation=man:labwc(1) man:systemd.special(7)
+BindsTo=graphical-session.target
+Wants=graphical-session-pre.target
+After=graphical-session-pre.target
--- a/docs/autostart
+++ b/docs/autostart
@@ -1,5 +1,13 @@
 # Example autostart file
 
+# When running under systemd, uncomment the systemctl line below to pull in
+# graphical-session.target via labwc-session.target. This lets systemd user
+# services declaring WantedBy=graphical-session.target (panels, portals,
+# notification daemons, etc.) start in sync with the labwc session. Enable
+# individual services with: systemctl --user enable <unit>
+#
+# systemctl --user --no-block start labwc-session.target
+
 # Set background color.
 swaybg -c '#113344' >/dev/null 2>&1 &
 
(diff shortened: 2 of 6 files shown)


Consolatis:
Thanks for the PR.

What are the benefits of running `systemctl --user --no-block start/stop labwc-session.target` via labwc rather than via `~/.config/labwc/autostart` and `~/.config/labwc/shutdown`? In general I feel like things like this should be handled by distro packaging (or in desktop environments using labwc) as labwc does not mandate or even recommends any specific init system.

We could add the two commands (in commented out state with explanation) to `autostart` and `shutdown` in the `docs/` directory though to make users aware of the option.

joske:
Thanks @Consolatis for your comment. Ok, I understand and agree with your stance. Are you ok to ship the systemd unit file though? 

I'm on Arch, and it doesn't come with any systemd units for labwc (even though Arch uses systemd), that's why I had the issue of `dms.service` not starting on login.

If you'd rather not, I'll try to work with dms upstream to have this documented there.

Consolatis:
Shipping the unit file in the `data` directory sounds sensible to me. If we should also install it when we detect systemd is another question. Any thoughts about this @johanmalm @jlindgren90 @tokyo4j ?

johanmalm:
PR LGTM in current state.

I'm happy to install it too, although I should caveat that by stating that my knowledge of `systemd` is very limited.

Suggest we notify and take some advice from Distro maintainers nearer to release time. Could always make the install optional if anyone has an issue with it.

joske:
Added a configure option to disable systemd unit file installation (defaults to on). If build system doesn't have systemd, it will also not be installed. I guess distros that oppose systemd will not have systemd on their build machines.

joske:
FreeBSD build fails for unrelated transient reason. Can someone restart the check?

johanmalm:
Thanks LGTM.

joske:
Thanks! @johanmalm can you also have a look at https://github.com/labwc/labwc/pull/3513 ? :pray:

PR #3536, opened 2026-04-24
Skip switcher state when osd.show=no

### LINKED ISSUE #3535: Focus not following window navigation when using F9/F10 keys
### Problem description

I'm trying to map F9 and F10 to previous-window and next-window respectively. However, even though the navigation cycles through the windows, they aren't actually being selected. It results in a "ghost selection" behavior where I have to manually press a modifier key to actually gain focus on the desired window.

### Steps to reproduce

    Map F9 to previous-window and F10 to next-window.
    Press F10 to cycle through open windows.
    Notice the selection highlight moves, but the window remains in the background.
    Press a modifier key (like Super or Alt) to finally focus the window.

Expected behavior
The window should be immediately focused and brought to the foreground as soon as the key is pressed, without requiring an extra input.

### labwc build source

Release

### labwc version

0.9.3

### labwc environment

### Distribution

NixOS unstable

elviosak:
fixes #3535 
Should we just skip the state (the border around the selected window, waiting for the modifier release) when `osd.show = no`?
I think this is closer to what is expected when turning osd off, i don't use it, but i makes sense to me.
When switching forward i had to send the current view to back so it doesn't get stuck cycling the same 2 windows.

If this is to be merged, does it need an update to docs too?
currently:
`show [yes|no] Draw the OnScreenDisplay when switching between windows. Default is yes.`

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -1150,6 +1150,15 @@ run_action(struct view *view, struct action *action,
 			if (action_get_bool(action, "immediate", false)) {
 				cycle_finish(true);
 				if (view && dir == LAB_CYCLE_DIR_FORWARD) {
+					/*
+					 * With immediate=no, running NextWindow just
+					 * once places the previously selected view
+					 * second from the top, causing the next call
+					 * to select it again. For immediate=yes, to
+					 * prevent getting stuck between the topmost
+					 * two views, we move the previously selected
+					 * view to the bottom.
+					 */
 					view_move_to_back(view);
 				}
 			}
--- a/src/action.c
+++ b/src/action.c
@@ -372,6 +372,10 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 			}
 			goto cleanup;
 		}
+		if (!strcasecmp(argument, "immediate")) {
+			action_arg_add_bool(action, argument, parse_bool(content, false));
+			goto cleanup;
+		}
 		break;
 	case ACTION_TYPE_SHOW_MENU:
 		if (!strcmp(argument, "menu")) {
@@ -1143,7 +1147,7 @@ run_action(struct view *view, struct action *action,
 			cycle_step(dir);
 		} else {
 			cycle_begin(dir, filter);
-			if (!rc.window_switcher.osd.show) {
+			if (action_get_bool(action, "immediate", false)) {
 				cycle_finish(true);
 				if (view && dir == LAB_CYCLE_DIR_FORWARD) {
 					view_move_to_back(view);
--- a/src/action.c
+++ b/src/action.c
@@ -1143,6 +1143,12 @@ run_action(struct view *view, struct action *action,
 			cycle_step(dir);
 		} else {
 			cycle_begin(dir, filter);
+			if (!rc.window_switcher.osd.show) {
+				cycle_finish(true);
+				if (view && dir == LAB_CYCLE_DIR_FORWARD) {
+					view_move_to_back(view);
+				}
+			}
 		}
 		break;
 	}
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -145,6 +145,10 @@ Actions are used in menus and keyboard/mouse bindings.
 	This determines whether to cycle through all windows or only windows of the
 	same application as the currently focused window. Default is "all".
 
+	*immediate* [yes|no]
+	This determines whether the switch happens immediately, useful when bound to
+	a key without modifier. Default is "no".
+
 *<action name="Reconfigure" />*
 	Re-load configuration and theme files.
 


tokyo4j:
I think we should rather add a new option like `<windowSwitcher immediate="yes|no">` or `<action name="NextWindow" immediate="yes|no">`. `<osd show="yes|no">` should only show or hide OSD.

src/action.c
```diff
@@ -1143,6 +1147,12 @@ run_action(struct view *view, struct action *action,
 			cycle_step(dir);
 		} else {
 			cycle_begin(dir, filter);
+			if (action_get_bool(action, "immediate", false)) {
+				cycle_finish(true);
+				if (view && dir == LAB_CYCLE_DIR_FORWARD) {
+					view_move_to_back(view);
+				}
```
tokyo4j, line 1163: What is this for?

elviosak: To avoid getting stuck in 2 windows like when u alt+tab (without immediate) and release both keys before pressing them again. Since with `immediate` the cycle state is not preserved i need to send the window to back so it won't be the "next window" again and it will have the same sequence as when u hold alt and press tab multiple times (without immediate).

tokyo4j: I see.  Then I suggest a comment like this:

```c
				/*
				 * With immediate=no, running NextWindow just
				 * once places the previously selected view
				 * second from the top, causing the next call
				 * to select it again. For immediate=yes, to
				 * prevent getting stuck between the topmost
				 * two views, we move the previously selected
				 * view to the bottom.
				 */
```

tokyo4j:
LGTM other than my comment above, but let's wait for feedbacks from @johanmalm or @Consolatis to see if the direction of this PR is acceptable.

I have some slight concerns about this PR:

1. It creates and destroys the OSD even when `immediate=yes`
2. Moving the previously selected window to the bottom (see the thread above) might behave in an unexpected way when window filter is applied. For example, `<action name="NextWindow" immediate="yes" identifier="current">` switches between some e.g. Alacritty windows and puts one of them below e.g. Firefox window.

But they are a bit annoying to fix and I think they are tolerable for now.

elviosak:
thanks for reviewing.

> It creates and destroys the OSD even when immediate=yes

it would be easier if it was tied to `windowSwitcher, but then we wouldn't be able to have 2 shorcuts: one regular alt+tab and another for immediate switch, not sure if that's desirable but i'm using it this way for testing and it's not bad, but i could change it if needed.

for the second one i'm not sure, i think it would need a lot of changes.

Narrat:
> Should we just skip the state (the border around the selected window, waiting for the modifier release) when osd.show = no?
I think this is closer to what is expected when turning osd off, i don't use it, but i makes sense to me.

Dunno how much I would have missed the borders if that change was kept. But as someone who disables the OSD this border is still part of recognizing which window will get the focus on tabbing. Especially with almost non existing theme borders.

johanmalm:
Could this be done in a cleaner way?

How about introducing a new action like NextWindowImmediate to completely decouple this from cycle_begin/finish?

elviosak:
> How about introducing a new action like NextWindowImmediate

I'm fine with it, it's pretty simple too.

> to completely decouple this from cycle_begin/finish?

do you mean to not use the "cycle" functions? I could ignore all windowSwitcher/osd stuff and have the action do it's stuff without starting cycle/osd, maybe add the filter arguments to the action so it acts the same as the windowSwitcher.

johanmalm:
> > to completely decouple this from cycle_begin/finish?
> 
> do you mean to not use the "cycle" functions? I could ignore all windowSwitcher/osd stuff and have the action do it's stuff without starting cycle/osd, maybe add the filter arguments to the action so it acts the same as the windowSwitcher.

Yes.

I haven't tried it, but conceptually I think it fits. If I'm thinking about it correctly it would avoid the two points raised by @tokyo4j and also avoid the `if (view && dir == LAB_CYCLE_DIR_FORWARD) { view_move_to_back(view); }` construct.

I suppose there would be a small amount of code duplication, but it may give a cleaner overall solution.

How about trying it as a separate branch/PR and then we can compare the two.

elviosak:
> > > to completely decouple this from cycle_begin/finish?
> > 
> > 
> > do you mean to not use the "cycle" functions? I could ignore all windowSwitcher/osd stuff and have the action do it's stuff without starting cycle/osd, maybe add the filter arguments to the action so it acts the same as the windowSwitcher.
> 
> Yes.
> 
> I haven't tried it, but conceptually I think it fits.

Would the new action have the same arguments/filters as the regular Next/PreviousWindow?

> If I'm thinking about it correctly it would avoid the two points raised by @tokyo4j

On the second point, instead of sending to bottom, i could send it to the position after the last one (in case filters are applied), not sure if that's better.

> and also avoid the `if (view && dir == LAB_CYCLE_DIR_FORWARD) { view_move_to_back(view); }` construct.

I don't think this can be avoided if we use the stacking order for the cycling order. When you switch to the second topmost view, the previous one has to go to the bottom to keep the same sequence.

> I suppose there would be a small amount of code duplication, but it may give a cleaner overall solution.
> 
> How about trying it as a separate branch/PR and then we can compare the two.

Sure, i'll give it a try.

elviosak:
let's say i have 6 windows open, 3 terminal(T), 3 firefox(F), filter by identifier=current
i'm thinking it should work like this:
|-|top|-|-|-|-|bottom|
|-|-|-|-|-|-|-|
|initial|F1|T1|F2|T2|F3|T3|
|next |F2|T1|F3|T2|F1|T3|
|next |F3|T1|F1|T2|F2|T3|
|next |same|as|initial|

previous would be the same in reverse

for output=cursor i think it would cycle through the filtered views like that but then also send the topmost of the filtered views to the front since the currently focused view might not be part of the filter.

@tokyo4j @johanmalm  do you agree?

johanmalm:
I like the analysis. It's a great thing to do prior to writing code :smile:

From a user's perspective I think the way you have done it is fine. I equally think that it would be fine if the topmost window was pushed all the way to the bottom on each `NextWindowImmediate`. I'm not sure which behavior would feel most natural to most people (minimum surprise). From a behavior perspective I don't feel strongly either way, but would lean towards the latter.

Do we have a feel for the complexity of implementing either of these two? Would it be easier to just insert them at the end of the list?

elviosak:
> Do we have a feel for the complexity of implementing either of these two? Would it be easier to just insert them at the end of the list?

I did try to implement the table from the previous post but it didn't work (the unfiltered views do not stay above the unfocused filtered ones), i'm only switching positions in the list with `wl_list_remove` and `wl_list_insert`, do i need to do something with the nodes or something else?
it's about 60 LOC, maybe a little more if i can fix it, if i just send it to back i'll have to try but it should be easier.

edit: ~~when trying to do the easier way i actually implemented the harder one, i'm not sure why it works now, but~~ i'll test a bit more and send another PR if i don't find issues.

edit2: nope, it was just one specific case.

elviosak:
closing this one, continuing in #3542

Issue #1079, opened 2023-09-10
Implement more libinput settings


Consolatis:
- [ ] disable while typing    
- [ ] natural scroll  
- [ ] scroll factor   
- [ ] middle button emulation 
- [ ] tap button map  
- [ ] clickfinger behavior    
- [ ] tap-to-click    
- [ ] drag lock       
- [ ] tap-and-drag
- [ ] accel profile
- [ ] sensitivity
- [ ] scroll_method
- [ ] scroll_button

Edit by Consolatis: made it into a list.

Edit to add: Thanks, I didn't think of that. :)

Edit 2 by Consolatis: added `scroll_method` and `scroll_button`.

_Originally posted by @droc12345 in https://github.com/labwc/labwc/issues/1064#issuecomment-1704303217_

PR #3540, opened 2026-04-25
Add onbutton scrollMethod, scrollButton

diniamo:
Adds the missing on-button(-down) scroll method (https://wayland.freedesktop.org/libinput/doc/latest/scrolling.html#button-scrolling).

The scrollButton option currently takes a decimal equivalent to a button in `linux/input-event-codes.h`, which is far from ideal, but I couldn't find a function in the labwc codebase that is a universal string-to-button conversion, and I'm not sure what the best way would be to implement it, because of the massive scope. The closest ones are `mouse_button_from_str` and `tablet_button_from_str`, but using them would
1. require a branch on the device type when setting device-unspecific options (not a big deal)
2. not cover all cases, because there might be other device types that support this scroll method

Related:
#2767 (pull request adding scrollMethod)
#1079 (related tracking issue)
#3404 (feature request discussion)

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -1140,6 +1140,7 @@ Note: To rotate touch events with output rotation, use the libinput
     <disableWhileTyping></disableWhileTyping>
     <clickMethod></clickMethod>
     <scrollMethod></scrollMethod>
+    <scrollButton></scrollButton>
     <sendEventsMode></sendEventsMode>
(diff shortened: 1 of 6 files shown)


johanmalm:
Thanks. LGTM. 

Can't think simple/elegant way to deal with the decimal dilemma. Suggest we merge this for now.

diniamo:
Thank you!

Issue #3255, opened 2025-12-05
Alt+Tab window switcher blocks other Alt-based keybindings until Alt is released


mbroemme:
### Problem description

When using the window switcher (`Alt+Tab`), other keybindings that also use the Alt modifier (e.g. `Alt+F1`, `Alt+F2` for workspace switching) stop working as long as `Alt` is still held after a window cycle. It feels as if the `Alt` key is "locked" by the window switcher until the modifier is released.

### Expected Behavior
* While the window switcher is active, pressing another Alt-based keybinding (e.g. `Alt+F1`, `Alt+F2`) should:
  * Finish the current window cycle (commit the selected window), and then
  * Execute the new keybinding (e.g. switch to workspace 1 or 2).

In other words, this sequence:
* `Alt` (pressed) → `Tab` (pressed, then released) → `F1` (pressed, `Alt` still held)

should both:
1. End the window switcher, focusing the selected window, and
2. Trigger the `Alt+F1` workspace switch.

### Actual Behavior
* When `Alt` is still held after using `Alt+Tab`, pressing `F1` (i.e. `Alt+F1`) does nothing:
  * The window switcher internally still considers `LAB_INPUT_STATE_CYCLE` active.
  * Only Next/Previous window actions continue to work; other actions (workspace switching) are ignored.
* Once `Alt` is released and pressed again, `Alt+F1` / `Alt+F2` work normally.

Observed sequences:
* `Alt` (pressed) + `Tab` (pressed) → **works** (window switcher starts and cycles)
* `Alt` (pressed) + `Tab` (pressed) + `Tab` (released) + `F1` (pressed) → **does NOT work**
* `Alt` (pressed) + `Tab` (pressed) + `Tab` (released) + `Alt` (released), then `Alt` (pressed) + `F1` (pressed) → **works**

This gives the impression that Alt is "stuck" until it is fully released.

### Additional Notes
* When the window switcher is active, `server->input_mode == LAB_INPUT_STATE_CYCLE`.
* In `actions_run()` (in `src/action.c`), only `ACTION_TYPE_NEXT_WINDOW` and `ACTION_TYPE_PREVIOUS_WINDOW` are allowed while in `LAB_INPUT_STATE_CYCLE`; other actions (e.g. the ones bound to `Alt+F1` / `Alt+F2`) are skipped.
* As a result, any non-cycling action triggered while the switcher is active is ignored until Alt is released and cycling ends.
[...]

PR #3542, opened 2026-04-29
Add Next/PreviousWindowImmediate to switch windows without OSD

### LINKED ISSUE #3535: Focus not following window navigation when using F9/F10 keys

elviosak:
closes #3535 supersedes #3536 
add 2 new actions to switch windows immediately without waiting for modifier release, using the same filters as Next/PreviousWindow, but without using the existing `cycle_begin` and `cycle_finish` functions to avoid creating the osd.

@johanmalm in this one i'm just sending the window to the back, should be the same behavior as the other PR.

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -630,4 +630,6 @@ enum lab_placement_policy view_placement_parse(const char *policy);
 /* xdg.c */
 struct wlr_xdg_surface *xdg_surface_from_view(struct view *view);
 
+bool view_matches_criteria(struct view *view, enum lab_view_criteria criteria);
+
 #endif /* LABWC_VIEW_H */
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -322,6 +322,61 @@ handle_osd_tree_destroy(struct wl_listener *listener, void *data)
 	free(osd_output);
 }
 
+static struct wl_list *prev(struct wl_list *elm) { return elm->prev; }
+static struct wl_list *next(struct wl_list *elm) { return elm->next; }
+
+void
+cycle_immediate(enum lab_cycle_dir direction, struct cycle_filter filter)
+{
+	if (wl_list_empty(&server.views)) {
+		return;
+	}
+	enum lab_view_criteria criteria =
+		LAB_VIEW_CRITERIA_NO_SKIP_WINDOW_SWITCHER
+		| LAB_VIEW_CRITERIA_NO_DIALOG;
+	if (filter.workspace == CYCLE_WORKSPACE_CURRENT) {
+		criteria |= LAB_VIEW_CRITERIA_CURRENT_WORKSPACE;
+	}
+	uint64_t cycle_outputs = get_outputs_by_filter(filter.output);
+	const char *cycle_app_id = NULL;
+	if (filter.app_id == CYCLE_APP_ID_CURRENT && server.active_view) {
+		cycle_app_id = server.active_view->app_id;
+	}
+
+	struct wl_list *head = &server.views;
+	struct wl_list *(*iter)(struct wl_list *list);
+	iter = direction == LAB_CYCLE_DIR_FORWARD ? next : prev;
+
+	struct wl_list *from = (direction == LAB_CYCLE_DIR_FORWARD) && server.active_view
+		? &server.active_view->link : head;
+
+	for (struct wl_list *elm = iter(from); elm != head; elm = iter(elm)) {
+		struct view *view = wl_container_of(elm, view, link);
+		if (!view_matches_criteria(view, criteria)) {
+			continue;
+		}
+		if (filter.output != CYCLE_OUTPUT_ALL) {
+			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+				continue;
+			}
+		}
+		if (cycle_app_id && strcmp(view->app_id, cycle_app_id) != 0) {
+			continue;
+		}
+		if (server.active_view && direction == LAB_CYCLE_DIR_FORWARD) {
+			/*
+			 * When cycling forward, the current active view needs to be
+			 * sent to back to keep the same sequence and avoid getting
+			 * stuck in the 2 topmost views.
+			 */
+			view_move_to_back(server.active_view);
+		}
+		desktop_focus_view(view, true);
+		break;
+	}
+	cursor_update_focus();
+}
+
 /* Return false on failure */
 static bool
 init_cycle(struct cycle_filter filter)
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -126,12 +126,17 @@ Actions are used in menus and keyboard/mouse bindings.
 	position if it had been maximized or tiled to a direction or region.
 
 *<action name="NextWindow" workspace="current" output="all" identifier="all" />*++
-*<action name="PreviousWindow" workspace="current" output="all" identifier="all" />*
+*<action name="PreviousWindow" workspace="current" output="all" identifier="all" />*++
+*<action name="NextWindowImmediate" workspace="current" output="all" identifier="all" />*++
+*<action name="PreviousWindowImmediate" workspace="current" output="all" identifier="all" />*++
 	Cycle focus to next/previous window, respectively.
 
 	Default keybinds for NextWindow and PreviousWindow are Alt-Tab and
 	Shift-Alt-Tab. While cycling through windows, the arrow keys move the
 	selected window forwards/backwards and the escape key halts the cycling.
+
+	NextWindowImmediate and PreviousWindowImmediate skip the Window Switcher
+	and OSD, useful for binding to keys without modifiers.
 
 	*workspace* [all|current]
 	This determines whether to cycle through windows on all workspaces or the
--- a/include/cycle.h
+++ b/include/cycle.h
@@ -103,6 +103,10 @@ void cycle_finish(bool switch_focus);
 /* Re-initialize the window switcher */
 void cycle_reinitialize(void);
 
+/* Immediately cycle to next/previous window */
+void cycle_immediate(enum lab_cycle_dir direction,
+	struct cycle_filter filter);
+
 /* Focus the clicked window and close OSD */
 void cycle_on_cursor_release(struct wlr_scene_node *node);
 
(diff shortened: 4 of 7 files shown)


johanmalm:
I like this much better :smile:

There is some common code in `cycle_immediate()` and `cycle_init()` but we can consider sorting that in a follow-up PR if we want to. It's only a few lines, so might not be worth it.

I found the logic within the `view_for_each()` loop hard to follow.

I think the code below is better. What's your view? Do you agree?

```C
index da7aef02..66c127e5 100644
--- a/include/view.h
+++ b/include/view.h
@@ -630,4 +630,6 @@ enum lab_placement_policy view_placement_parse(const char *policy);
 /* xdg.c */
 struct wlr_xdg_surface *xdg_surface_from_view(struct view *view);
 
+bool matches_criteria(struct view *view, enum lab_view_criteria criteria);
+
 #endif /* LABWC_VIEW_H */
index 6c7d2921..f9808368 100644
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -323,6 +323,9 @@ handle_osd_tree_destroy(struct wl_listener *listener, void *data)
 	free(osd_output);
 }
 
+static struct wl_list *prev(struct wl_list *elm) { return elm->prev; }
+static struct wl_list *next(struct wl_list *elm) { return elm->next; }
+
 void
 cycle_immediate(enum lab_cycle_dir direction, struct cycle_filter filter)
 {
@@ -341,10 +344,18 @@ cycle_immediate(enum lab_cycle_dir direction, struct cycle_filter filter)
 		cycle_app_id = server.active_view->app_id;
 	}
 
-	struct view *view;
-	struct view *view_to_focus = NULL;
-	struct view *view_to_back = NULL;
-	for_each_view(view, &server.views, criteria) {
+	struct wl_list *head = &server.views;
+	struct wl_list *(*iter)(struct wl_list *list);
+	iter = direction == LAB_CYCLE_DIR_FORWARD ? next : prev;
+
+	struct wl_list *elm, *from = direction == LAB_CYCLE_DIR_FORWARD && server.active_view
+		? &server.active_view->link : head;
+
+	for (elm = iter(from); elm != head; elm = iter(elm)) {
+		struct view *view = wl_container_of(elm, view, link);
+		if (!matches_criteria(view, criteria)) {
+			continue;
+		}
 		if (filter.output != CYCLE_OUTPUT_ALL) {
 			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
 				continue;
@@ -353,31 +364,11 @@ cycle_immediate(enum lab_cycle_dir direction, struct cycle_filter filter)
 		if (cycle_app_id && strcmp(view->app_id, cycle_app_id)) {
 			continue;
 		}
-		/* view matches the filters, now figure out the order */
-		if (server.active_view == view && direction == LAB_CYCLE_DIR_FORWARD) {
-			/* When cycling forward, the current active view needs to be sent to back
-			 * to keep the same sequence and avoid getting stuck in the 2 topmost views.
-			 * But if the active view does not match the filter, there's no need to do
-			 * the same since it's not part of the cycle, and this probably only happens
-			 * with output="cursor".
-			 */
-			view_to_back = view;
-			view_to_focus = view;
-			continue;
-		}
-		view_to_focus = view;
 		if (direction == LAB_CYCLE_DIR_FORWARD) {
-			/* When cycling forward, keep the second match and break,
-			 * otherwise keep going and keep the last one.
-			 */
-			break;
+			view_move_to_back(server.active_view);
 		}
-	}
-	if (view_to_back && view_to_back != view_to_focus) {
-		view_move_to_back(view_to_back);
-	}
-	if (view_to_focus) {
-		desktop_focus_view(view_to_focus, true);
+		desktop_focus_view(view, true);
+		break;
 	}
 	cursor_update_focus();
 }
index 1d43fcfa..4f00b88b 100644
--- a/src/view.c
+++ b/src/view.c
@@ -263,7 +263,7 @@ view_get_root(struct view *view)
 	return view;
 }
 
-static bool
+bool
 matches_criteria(struct view *view, enum lab_view_criteria criteria)
 {
 	if (!view_is_focusable(view)) {
```

I can push it on your branch if you want, but prefer to compare thinking on whether it actually is an improvement first + also ask for permission to push on branch.

elviosak:
> I think the code below is better. What's your view? Do you agree?

 Yeah, i like it. Just one thing, does this need to check if `server.active_view` is not `NULL`?
 ```
 +			view_move_to_back(server.active_view);
 ```

> I can push it on your branch if you want, but prefer to compare thinking on whether it actually is an improvement first + also ask for permission to push on branch.

You can push it, haven't tested but it looks better.

Edit: also the previous version only moved to back if the active view matched the filters, i think it won't match only with `output=cursor`, it might be weird to someone who uses it, but it can also be weird in some of the other cases when moving to back, so i'm fine with it either way.

johanmalm:
I have pushed the patch above.

> Yeah, i like it. Just one thing, does this need to check if `server.active_view` is not `NULL`?

Good catch :smile: I added a check for that.

johanmalm:
@Consolatis @jlindgren90 @tokyo4j  - I'm good with this PR, but as I authored the second commit, could one of you do a final review?

src/cycle/cycle.c
```diff
@@ -322,6 +323,56 @@ handle_osd_tree_destroy(struct wl_listener *listener, void *data)
...
+		struct view *view = wl_container_of(elm, view, link);
+		if (!matches_criteria(view, criteria)) {
+			continue;
+		}
+		if (filter.output != CYCLE_OUTPUT_ALL) {
+			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+				continue;
+			}
+		}
+		if (cycle_app_id && strcmp(view->app_id, cycle_app_id)) {
```
tokyo4j: nit: I prefer `strcmp(view->app_id, cycle_app_id) != 0` as it gives me "negative" (=do *not* match) impression.

Maybe at some point we should consider adding `str_match()` or something that @Consolatis suggested on IRC a while ago.

src/cycle/cycle.c
```diff
@@ -322,6 +323,56 @@ handle_osd_tree_destroy(struct wl_listener *listener, void *data)
...
+	if (filter.app_id == CYCLE_APP_ID_CURRENT && server.active_view) {
+		cycle_app_id = server.active_view->app_id;
+	}
+
+	struct wl_list *head = &server.views;
+	struct wl_list *(*iter)(struct wl_list *list);
+	iter = direction == LAB_CYCLE_DIR_FORWARD ? next : prev;
+
+	struct wl_list *elm, *from = direction == LAB_CYCLE_DIR_FORWARD && server.active_view
+		? &server.active_view->link : head;
```
tokyo4j: nit: Let's move `elm` declaration inside `for(..)`

I also suggest wrapping `LAB_CYCLE_DIR_FORWARD && server.active_view` with parenthesis for readability.

include/view.h
```diff
@@ -630,4 +630,6 @@ enum lab_placement_policy view_placement_parse(const char *policy);
 /* xdg.c */
 struct wlr_xdg_surface *xdg_surface_from_view(struct view *view);
 
+bool matches_criteria(struct view *view, enum lab_view_criteria criteria);
```
tokyo4j: I suggest renaming this to `view_matches_criteria()`.

src/cycle/cycle.c
```diff
@@ -322,6 +323,56 @@ handle_osd_tree_destroy(struct wl_listener *listener, void *data)
...
+		if (filter.output != CYCLE_OUTPUT_ALL) {
+			if (!view->output || !(cycle_outputs & view->output->id_bit)) {
+				continue;
+			}
+		}
+		if (cycle_app_id && strcmp(view->app_id, cycle_app_id)) {
+			continue;
+		}
+		if (server.active_view && direction == LAB_CYCLE_DIR_FORWARD) {
+			view_move_to_back(server.active_view);
```
tokyo4j, line 372: I suggest adding a comment like https://github.com/labwc/labwc/pull/3536#discussion_r3143392160. You can shorten it if you feel it's too verbose.

tokyo4j:
LGTM other than my nitpicks above. The code duplication is a bit unfortunate, but I understand it's the best option in this case.

@elviosak Thank you for your contribution!
@johanmalm Thank you too for reviewing this PR. I'm sorry I can't dedicate as much time to this project since I recently started working full-time.

src/cycle/cycle.c
elviosak: ok, i added the same message that was in the first commit, is that text ok?

include/view.h
elviosak: done, had to change window_rules.c too because of conflicting names, is the new name ok?

src/cycle/cycle.c
elviosak: done

src/cycle/cycle.c
elviosak: misremembered reading it on the coding style: https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#linux-kernel-style-basics 
but it's only mentioned to avoid "== 0"
changed, and i prefer this too.

elviosak:
@tokyo4j Thanks for reviewing.

this PR also partially solves #3255 i think

jlindgren90:
Congrats on the new job @tokyo4j :)

Day job has also been very busy on my end, don't wait for me to review.

johanmalm:
Thanks all. LGTM.

I'll squash the three commits because the second two are just refactoring and fixups.

johanmalm:
@elviosak I re-flowed your multi-line comment for consistency. https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#linux-kernel-style-basics

No changes other than that.

PR #3543, opened 2026-04-29
session: run activation env update synchronously

joske:
dbus-update-activation-environment and systemctl --user import-environment were fired asynchronously via spawn_async_no_shell, racing with the autostart script. Any systemd user service started from autostart (e.g. via labwc-session.target #3534) could start before the import completed, leaving WAYLAND_DISPLAY and related variables absent from its environment and those of any apps it launches. This causes DMS shell to fail to launch many apps.

Run both commands synchronously via a new spawn_sync_no_shell helper so the import is guaranteed to complete before the autostart script executes.

### DIFF
--- a/src/common/spawn.c
+++ b/src/common/spawn.c
@@ -108,7 +108,7 @@ spawn_sync_no_shell(char const *command)
 	switch (child) {
 	case -1:
 		wlr_log(WLR_ERROR, "unable to fork()");
-		goto out;
+		break;
 	case 0:
 		reset_signals_and_limits();
 		execvp(argv[0], argv);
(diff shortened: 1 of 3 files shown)


src/common/spawn.c
```diff
@@ -89,6 +89,38 @@ spawn_async_no_shell(char const *command)
...
+		g_message("%s", err->message);
+		g_error_free(err);
+		return;
+	}
+
+	pid_t child = fork();
+	switch (child) {
+	case -1:
+		wlr_log(WLR_ERROR, "unable to fork()");
+		goto out;
```
ahesford: This seems to be a remnant of the original function. You can drop the label and use a simple break here.

joske: Good catch, fixed — replaced the `goto out` with `break` and removed the label.

joske:
CI failures appear to be transient GH errors.

johanmalm:
Thanks

PR #3547, opened 2026-05-01
Fix for chromium popup on maximized window

### LINKED ISSUE #3544: Chromium and chromium-based browsers show submenu in the wrong monitor when maximized on the left monitor
### Problem description

I use 2 monitors side-by-side, both 1920x1080 and no scaling, when the browser window is maximized on the left monitor, the submenus from the hamburguer menu show on the right monitor, tested 3 chromium-based browsers: chromium, brave and helium as native wayland windows, all of them have the same issue.

I took 2 screenshots in the middle of the 2 monitors to show it, the border of the window is also the border of the monitors. In the first the window is snapped to the right, in the second it is maximized.

when tiled/snapped:

Image

when maximized:

Image

tried it in kde plasma and it worked correctly.

### Steps to reproduce

1. use 2 monitors side-by-side.
2. open a chromium based browser and maximize it on the left monitor.
3. open the hamburger menu and hover over some submenu

expected behavior: submenu opens on the same monitor.

actual behavior: submenu opens on another monitor.

### labwc build source

Local build

### labwc version

labwc 0.9.5-80-gd058f89c-dirty (+xwayland +nls +rsvg +libsfdo)

### labwc environment

### Distribution

Arch

elviosak:
closes #3544

the issue is the popup x + parent x coordinates have the exact value as the output width + output x, so the popups ends up with a coordinate x equal to 0 of the next monitor.

can be solved in a few different ways, this one made sense to me.

for easy checking the bug and fix: `WLR_WL_OUTPUTS=2 ./build/labwc -s chromium` move the chromium window to WL-1 and maximize it, then open the menu and some submenu. Use other chromium-based browser if u don't have chromium.

Edit: Just thought of something else, couldn't we just remove the MAX(...) part? just keep `output_nearest_to(parent_lx, parent_ly)` shouldn't it always be the same output as the parent?

### DIFF
--- a/src/xdg-popup.c
+++ b/src/xdg-popup.c
@@ -43,10 +43,13 @@ popup_unconstrain(struct xdg_popup *popup)
 	 * than zero, typically with Qt apps. We therefore clamp it to avoid for
 	 * example the 'File' menu of a maximized window to end up on an another
 	 * output.
+	 * Also some apps open the menu exactly at the right border when maximized,
+	 * causing popup_box->x (or y?) to be in the next output. We subtract one
+	 * inside MAX to avoid the problem mentioned above.
 	 */
 	struct wlr_box *popup_box = &popup->wlr_popup->scheduled.geometry;
-	struct output *output = output_nearest_to(parent_lx + MAX(popup_box->x, 0),
-		parent_ly + MAX(popup_box->y, 0));
+	struct output *output = output_nearest_to(parent_lx + MAX(popup_box->x - 1, 0),
+		parent_ly + MAX(popup_box->y - 1, 0));
 	struct wlr_box usable = output_usable_area_in_layout_coords(output);
 
 	/* Get offset of toplevel window from its surface */


johanmalm:
Thanks. LGTM.

> Edit: Just thought of something else, couldn't we just remove the MAX(...) part? just keep output_nearest_to(parent_lx, parent_ly) shouldn't it always be the same output as the parent?

If a window straddles across two outputs it wouldn't be enough to just use `parent_{x,y}`.

Consider the following:

1. You have two outputs side-by-side.
2. The window you described in #3544 is in its natural state (i.e. not maximized). Some of the window is on the left hand screen, but most of it in on the right hand screen.
3. You open the popup (as in the screenshot in #3544) which you expect to appear near the right hand edge of the Chromium window.
4. In this situation `output_nearest_to(parent_lx, parent_ly)` would return the left hand output, but you want the right hand one.

Do you agree.

elviosak:
Yeah makes sense, i didn't think of windows in 2 monitors.

Issue #3435, opened 2026-03-09
I can't use labwc with kanshi in mirroring mode


brometeo:
### Problem description

I have a basic kanshi configuration for both displays, one permanent and one not. When I turn on second display (both have same size and position 0,0 for mirroring), the first display deactivates or both work but not in mirroring mode. If I leave graphic mode and reenter again, then configuration works as expected and the displays are showed in mirroring mode.

My kanshi config file:
```
profile {
  output DVI-D-1 enable
}

profile {  
  output HDMI-A-1 enable position 0,0
  output DVI-D-1 enable position 0,0
}
```

### Steps to reproduce

1. Launch labwc with one display and kanshi running from autostart config file
2. Turn on second display
3. Test if both displays are showed in mirror

### labwc build source

Release

### labwc version

0.9.4

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

Debian testing

johanmalm:
IIRC native mirroring is not supported by `wlroots` (and therefore `labwc`).

It is possible that it can be achieved using `wl-mirror` (https://github.com/Ferdi265/wl-mirror). If you have the time and ability to verify this, any feedback would be gratefully received.

brometeo:
El lunes 23 de marzo, Johan Malm escribió:
> IIRC native mirroring is not supported by `wlroots` (and therefore `labwc`).

kanshi was created for being compatible with Sway, so I though it should work
with any window composer wlroots based.

I can configure mirroring wdisplays and labwc, so I don't understand why a tool
like kanshi is not good enough for the task.

> It is possible that it can be achieved using `wl-mirror`
> (https://github.com/Ferdi265/wl-mirror). If you have the time and ability to
> verify this, any feedback would be gratefully received.

I have seen that wl-mirror makes a copy of first screen in second screeen,
instead of configuring both for showing the same image. Perhaps it works, but
I don't understand the problem for kanshi not working as explained in
documentation with labwc, and why it can't be fixed as any other problem.

Greets.

-- 
David

johanmalm:
I don't know enough about it to answer. Help appreciated.

Are you saying that your `kanshi` setup supports mirroring with `sway`, but not with `labwc`?

brometeo:
El martes 24 de marzo, Johan Malm escribió:
> Are you saying that your `kanshi` setup supports mirroring with `sway`, but
> not with `labwc`?

Really I don't know. Perhaps I should test it. But kanshi is developed with
[...]

PR #3551, opened 2026-05-03
When output is turned off and there is a mirror, assign it to views.

### LINKED ISSUE #3550: View has no output error when laptop display is off
### Problem description

Whenever I connect an external display to my laptop, I will often mirror the external and laptop displays and then turn off my laptop display. However, this will cause `view has no output, not snapping to edge` error whenever I press my keybinds for `SnapToEdge` or `SnapToRegion`.

### Steps to reproduce

1. With external monitor already connected, launch labwc from TTY1: `labwc`
2. Launch Alacritty
3. Mirror the 2 displays: `wlr-randr --output DP-3 --mode 2560x1080 --pos 0,0 && wlr-randr --output eDP-1 --custom-mode 2560x1080 --pos 0,0 --scale 1`
4. Turn off laptop display: `wlopm --off eDP-1`
5. With focus still on Alacritty, press keybinds for `SnapToEdge` or `SnapToRegion`, **which has no effects**
6. Turn laptop display back on: `wlopm --on eDP-1`, and the keybinds work again

Log snippet:
```
00:00:22.366 [INFO] [backend/drm/drm.c:948] connector eDP-1: Turning off
00:00:22.465 [DEBUG] [backend/drm/drm.c:1372] Reallocating CRTCs
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector eDP-1 (connected, needs CRTC): no CRTC → CRTC 363
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-1 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-2 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-3 (connected, needs CRTC): CRTC 367 (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-4 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-5 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-6 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-7 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [backend/drm/drm.c:1435]   Connector DP-8 (disconnected): no CRTC (no change)
00:00:22.465 [DEBUG] [render/swapchain.c:101] Allocating new swapchain buffer
00:00:22.466 [DEBUG] [render/allocator/gbm.c:146] Allocated 256x256 GBM buffer with format AR24 (0x34325241), modifier LINEAR (0x0000000000000000)
00:00:22.466 [DEBUG] [render/gles2/renderer.c:157] Created GL FBO for buffer 256x256
00:00:32.002 [DEBUG] [../labwc/src/input/keyboard.c:252] keycode matched
00:00:32.002 [DEBUG] [../labwc/src/action.c:1654] Handling action 9: SnapToEdge
00:00:32.002 [ERROR] [../labwc/src/view.c:2031] view has no output, not snapping to edge
```

Attached is the full log and sample rc.xml

[rc-debug.xml](https://github.com/user-attachments/files/27313256/rc-debug.xml)

[labwc.log](https://github.com/user-attachments/files/27313235/labwc.log)

### labwc build source

Release

### labwc version

labwc 0.9.6 (+xwayland +nls +rsvg +libsfdo)

### labwc environment

### Distribution

ArchLinux

elviosak:
fixes #3550

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -1080,17 +1080,43 @@ output_usable_area_in_layout_coords(struct output *output)
 	return box;
 }
 
+static struct output *
+output_get_usable_mirror(struct output *output)
+{
+	struct output *mirror;
+	struct wlr_box mirror_box, output_box;
+	wlr_output_layout_get_box(server.output_layout, output->wlr_output, &output_box);
+	wl_list_for_each(mirror, &server.outputs, link) {
+		if (!output_is_usable(mirror) || output->id_bit == mirror->id_bit) {
+			continue;
+		}
+		wlr_output_layout_get_box(server.output_layout, mirror->wlr_output, &mirror_box);
+		if (wlr_box_equal(&output_box, &mirror_box)) {
+			return mirror;
+		}
+	}
+	return NULL;
+}
+
 void
 handle_output_power_manager_set_mode(struct wl_listener *listener, void *data)
 {
 	struct wlr_output_power_v1_set_mode_event *event = data;
 	struct output *output = event->output->data;
 	assert(output);
-
+	struct output *mirror = output_get_usable_mirror(output);
 	switch (event->mode) {
 	case ZWLR_OUTPUT_POWER_V1_MODE_OFF:
 		if (!event->output->enabled) {
 			return;
+		}
+		if (mirror) {
+			struct view *view;
+			for_each_view(view, &server.views, LAB_VIEW_CRITERIA_NONE) {
+				if (view->output == output) {
+					view->output = mirror;
+				}
+			}
 		}
 		wlr_output_state_set_enabled(&output->pending, false);
 		output_state_commit(output);
--- a/src/output.c
+++ b/src/output.c
@@ -927,6 +927,24 @@ output_from_name(const char *name)
 	return NULL;
 }
 
+static struct output *
+output_get_usable_mirror(struct output *output)
+{
+	struct output *mirror;
+	struct wlr_box mirror_box, output_box;
+	wlr_output_layout_get_box(server.output_layout, output->wlr_output, &output_box);
+	wl_list_for_each(mirror, &server.outputs, link) {
+		if (!output_is_usable(mirror) || output->id_bit == mirror->id_bit) {
+			continue;
+		}
+		wlr_output_layout_get_box(server.output_layout, mirror->wlr_output, &mirror_box);
+		if (wlr_box_equal(&output_box, &mirror_box)) {
+			return mirror;
+		}
+	}
+	return NULL;
+}
+
 struct output *
 output_nearest_to(int lx, int ly)
 {
@@ -934,8 +952,15 @@ output_nearest_to(int lx, int ly)
 	wlr_output_layout_closest_point(server.output_layout, NULL, lx, ly,
 		&closest_x, &closest_y);
 
-	return output_from_wlr_output(wlr_output_layout_output_at(server.output_layout,
-			closest_x, closest_y));
+	struct output *output = output_from_wlr_output(wlr_output_layout_output_at(
+		server.output_layout, closest_x, closest_y));
+	if (output && !output_is_usable(output)) {
+		struct output *mirror = output_get_usable_mirror(output);
+		if (mirror) {
+			return mirror;
+		}
+	}
+	return output;
 }
 
 struct output *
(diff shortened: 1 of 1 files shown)


johanmalm:
Not tested, but code looks fine in terms for fixing the reported problem. Before I merge though I'm of a mind to take a slight step back to look at the context here and request thoughts from others.

1. Do we actually support mirroring? See #3435 - where mirroring was reported not to work with `labwc` whereas it does with `sway`
2. If we do support mirroring, does it still make sense to use `view->output`
3. Do we need to develop strategies for handling the situation where multiple outputs have the same (lx, ly) but different width/height? 
4. In this PR we do a check on `wlr_box_equal(&output_box, &mirror_box)'`. Should we allow output/mirror to have different sizes?

@jlindgren90 @tokyo4j @Consolatis  - any thoughts on this gratefully received.

andyc1234:
Hi, I am the author of the issue that this PR is attempting to fix, and I would like to chime in with what little knowledge I have on this topic.

- Mirroring actual works perfectly fine for me. Instead of `kanshi`, I am using 2 `wlr-randr` commands to do the mirroring: `wlr-randr --output DP-3 --mode 2560x1080 --pos 0,0 && wlr-randr --output eDP-1 --custom-mode 2560x1080 --pos 0,0 --scale 1`
- When mirrored with those 2 commands, everything works as expect with `labwc`. **Problems only start occurring when I turned off my laptop display**: `wlopm --off eDP-1`
- If I turn on my laptop display again with `wlopm --on eDP-1`, everything is back to normal
- Personally, I always set my external and laptop displays to have the same resolution because that's what I expect mirroring to mean. (Notice that I am using `--custom-mode` for my laptop display `eDP-1` because it has a different aspect ratio than my external display)
- While this PR fixes the originally reported issue, I [documented](https://github.com/labwc/labwc/issues/3550#issuecomment-4367258107) addition issues that I am facing when my laptop display is off

My use case for this is that after my 2 displays are mirrored, I can close my laptop lid and turn off its display.

Appreciate you all for looking into this!

elviosak:
the 2 last commits are different ways i thought of to fix this issue, but i think the first one would need to change a lot more functions since the `wlr_output_*` functions will not check if the output is disabled, so the same issue would happen in other situations.
The last one should be more resilient, and probably not cause more issues.

Changed to draft while we wait for more opinions if that is a good solution.

@andyc1234 can you check if this fixes the last issues you had?

elviosak:
> 1. Do we actually support mirroring? See [I can't use labwc with kanshi in mirroring mode #3435](https://github.com/labwc/labwc/issues/3435) - where mirroring was reported not to work with `labwc` whereas it does with `sway`

wlr-randr as described in the issue #3550 seems to work well, i'm only using in nested labwc for testing.

> 2. If we do support mirroring, does it still make sense to use `view->output`

i think we need "something" that we can get a box from, for snapping, moving when output is removed, etc.
and `struct output` is convenient because we can use `wlr_*` functions

> 3. Do we need to develop strategies for handling the situation where multiple outputs have the same (lx, ly) but different width/height?
> 4. In this PR we do a check on `wlr_box_equal(&output_box, &mirror_box)'`. Should we allow output/mirror to have different sizes?

There might be some cases where ppl use overlapping outputs, but i think the most common case would be mirrorring (IMO mirrorring is only when 2 or more outputs have the same x, y, width and height)

andyc1234:
> @andyc1234 can you check if this fixes the last issues you had?

Yes, I have pull down the latest commit from this PR and all the issues I have mentioned so far are fixed! However, if I try to take a screenshot with `grim -g '100,100 50x50' out.png`, `grim` will error with `no supported format found`. Probably not labwc-related, just want to mention it as a failed use-case.

Interesting, I noticed that I can use `wlr-randr` to turn off my laptop display as well: `wlr-randr --output eDP-1 --off` If I do so instead of `wlopm --off eDP-1`, then everything works as expected **even without this PR**. I have verified this using the tip of `master`, and `SnapToEdge`, `SnapToRegion`, placement of new windows and static menu, and `grim` all work as normal.

My uneducated guess is that `wlopm --off` turns off the display but the `output` object stays around, which is causing issues with functions that handle `output` objects. On the other hand, `wlr-randr --off` destroys the `output` object, it was as if I have physically disconnected my laptop display.

I will continue to play around with the 2 methods of `--off` and see if I can find further issues.

elviosak:
yea AFAIK it's like you said, wlopm is better for turning off all displays so the windows don't get moved around when output is removed. just tested and the alacritty new window and grim issue both happen without this PR, not sure what happens with the grim issue.

elviosak:
closing this, it's probably better to just use `wlopm` or `wlr-randr` according to the user's need, as reordering the list might cause some other issues.

PR #3552, opened 2026-05-03
view: clamp position to prevent windows moving fully off-screen

### LINKED ISSUE #2381: Restrict MoveRelative to output layout
Currently its possible to move outputs outside of the output layout.

st0rm-shad0w:
**Description:**
> Assisted by AI: Claude

An attempt to fix #2381 (Restrict MoveRelative to output layout).

Clamps window position in view_move_relative() so at least titlebar_height
pixels remain visible within the nearest output's usable area, preventing
windows from being moved entirely off-screen.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -12,6 +12,7 @@
 #include "buffer.h"
 #include "common/box.h"
 #include "common/list.h"
+#include "common/macros.h"
 #include "common/match.h"
 #include "common/mem.h"
 #include "common/string-helpers.h"
@@ -609,6 +610,10 @@ view_resize_relative(struct view *view, int left, int right, int top, int bottom
 	view_set_untiled(view);
 }
 
+struct point {
+	int x, y;
+};
+
 /*
  * Returns the total squared distance of the box's four corners from the
  * output layout. A return value of 0 means the box is fully inside.
@@ -618,19 +623,35 @@ box_layout_overshoot(struct wlr_box *box)
 {
 	struct wlr_output_layout *layout = server.output_layout;
 	double overshoot = 0;
-	double corners[4][2] = {
-		{box->x, box->y},
-		{box->x + box->width - 1, box->y},
-		{box->x, box->y + box->height - 1},
-		{box->x + box->width - 1, box->y + box->height - 1},
+	struct point corners[] = { {
+			.x = box->x,
+			.y = box->y,
+		}, {
+			.x = box->x + box->width - 1,
+			.y = box->y,
+		}, {
+			.x = box->x,
+			.y = box->y + box->height - 1,
+		}, {
+			.x = box->x + box->width - 1,
+			.y = box->y + box->height - 1,
+		},
 	};
-	for (int i = 0; i < 4; i++) {
+
+	for (size_t i = 0; i < ARRAY_SIZE(corners); i++) {
+		struct point corner = corners[i];
+
+		/* We only care about corners outside the output-layout */
+		if (wlr_output_layout_output_at(layout, corner.x, corner.y)) {
+			continue;
+		}
+
 		double closest_x, closest_y;
-		wlr_output_layout_closest_point(layout, NULL,
-			corners[i][0], corners[i][1],
+		wlr_output_layout_closest_point(layout, NULL, corner.x, corner.y,
 			&closest_x, &closest_y);
-		double dx = corners[i][0] - closest_x;
-		double dy = corners[i][1] - closest_y;
+		wlr_log(WLR_ERROR, "closest-x=%f; closest-y=%f", closest_x, closest_y);
+		double dx = corner.x - closest_x;
+		double dy = corner.y - closest_y;
 		overshoot += dx * dx + dy * dy;
 	}
 	return overshoot;
@@ -667,7 +688,7 @@ view_move_relative(struct view *view, int x, int y)
 	dest.y += y;
 	double current_overshoot = box_layout_overshoot(&current_box);
 	double dest_overshoot = box_layout_overshoot(&dest);
-	if (dest_overshoot > 0 && dest_overshoot >= current_overshoot) {
+	if (dest_overshoot > 0 && dest_overshoot > current_overshoot) {
 		if (abs(x) > 1 || abs(y) > 1) {
 			/*
 			 * It is reasonable to assume that users will specify
--- a/src/view.c
+++ b/src/view.c
@@ -609,6 +609,33 @@ view_resize_relative(struct view *view, int left, int right, int top, int bottom
 	view_set_untiled(view);
 }
 
+/*
+ * Returns the total squared distance of the box's four corners from the
+ * output layout. A return value of 0 means the box is fully inside.
+ */
+static double
+box_layout_overshoot(struct wlr_box *box)
+{
+	struct wlr_output_layout *layout = server.output_layout;
+	double overshoot = 0;
+	double corners[4][2] = {
+		{box->x, box->y},
+		{box->x + box->width - 1, box->y},
+		{box->x, box->y + box->height - 1},
+		{box->x + box->width - 1, box->y + box->height - 1},
+	};
+	for (int i = 0; i < 4; i++) {
+		double closest_x, closest_y;
+		wlr_output_layout_closest_point(layout, NULL,
+			corners[i][0], corners[i][1],
+			&closest_x, &closest_y);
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks for contributing.

I think this should consider the entire output-layout (all outputs) rather than a specific one.

What would your thoughts be on something like this?

```C
index bcbd366e..54238869 100644
--- a/src/view.c
+++ b/src/view.c
@@ -609,6 +609,17 @@ view_resize_relative(struct view *view, int left, int right, int top, int bottom
 	view_set_untiled(view);
 }
 
+static bool
+is_box_in_output_layout(struct wlr_box *box)
+{
+	struct wlr_output_layout *layout = server.output_layout;
+	return wlr_output_layout_output_at(layout, box->x, box->y)
+		&& wlr_output_layout_output_at(layout, box->x, box->y + box->height - 1)
+		&& wlr_output_layout_output_at(layout, box->x + box->width - 1, box->y)
+		&& wlr_output_layout_output_at(layout, box->x + box->width - 1,
+			box->y + box->height - 1);
+}
+
 void
 view_move_relative(struct view *view, int x, int y)
 {
@@ -621,6 +632,26 @@ view_move_relative(struct view *view, int x, int y)
 		view_set_untiled(view);
 		view_move_resize(view, view->natural_geometry);
 	}
+
+	/* Only proceed if window destination is within output layout */
+	struct wlr_box dest = ssd_max_extents(view);
+	dest.x += x;
+	dest.y += y;
+	if (!is_box_in_output_layout(&dest)) {
+		if (abs(x) > 1 || abs(y) > 1) {
+			/*
+			 * It is reasonable to assume that users will specify
+			 * big numbers (like 20) for the MoveRelative x or y
+			 * value. When moving towards a layout outer edge, we
+			 * would not want the window to stop a distance away
+			 * from the edge, so recursively nudge it nearer until
+			 * it is touching.
+			 */
+			view_move_relative(view, x / 2, y / 2);
+		}
+		return;
+	}
+
 	view_move(view, view->pending.x + x, view->pending.y + y);
 }
```

st0rm-shad0w:
I agree, covering all outputs should be the right approach. In my attempt, somehow I totally overlooked that.

Also, I tested the patch you shared, but noticed that MoveRelative fails to bring a partially off-screen window back inside the output layout.

It looks like the overshoot check in view_move_relative was using ssd_max_extents(view), which reads view->current geometry. But since the actual move is applied relative to view->pending, current lags behind pending during key repeats (or  manual window placement). Because of that stale reference, an inward move can actually look like it's creating an overshoot rather than reducing it, which causes the function to  block the move.

I fixed this by computing the box directly from view->pending, ensuring the overshoot comparison uses the exact same reference frame as the move itself.

Let me know if this updated code looks correct to you.


```diff
diff --git a/src/view.c b/src/view.c
index bcbd366e..6c8b6406 100644
--- a/src/view.c
+++ b/src/view.c
@@ -609,6 +609,33 @@ view_resize_relative(struct view *view, int left, int right, int top, int bottom
 	view_set_untiled(view);
 }
 
+/*
+ * Returns the total squared distance of the box's four corners from the
+ * output layout. A return value of 0 means the box is fully inside.
+ */
+static double
+box_layout_overshoot(struct wlr_box *box)
+{
+	struct wlr_output_layout *layout = server.output_layout;
+	double overshoot = 0;
+	double corners[4][2] = {
+		{box->x, box->y},
+		{box->x + box->width - 1, box->y},
+		{box->x, box->y + box->height - 1},
+		{box->x + box->width - 1, box->y + box->height - 1},
+	};
+	for (int i = 0; i < 4; i++) {
+		double closest_x, closest_y;
+		wlr_output_layout_closest_point(layout, NULL,
+			corners[i][0], corners[i][1],
+			&closest_x, &closest_y);
+		double dx = corners[i][0] - closest_x;
+		double dy = corners[i][1] - closest_y;
+		overshoot += dx * dx + dy * dy;
+	}
+	return overshoot;
+}
+
 void
 view_move_relative(struct view *view, int x, int y)
 {
@@ -621,6 +648,40 @@ view_move_relative(struct view *view, int x, int y)
 		view_set_untiled(view);
 		view_move_resize(view, view->natural_geometry);
 	}
+
+	/*
+	 * Only proceed if movement brings the window closer to being fully
+	 * inside the output layout. This allows a partially-outside window
+	 * to be moved back inside, while still preventing a fully-inside
+	 * window from being moved outside.
+	 */
+	struct border border = ssd_thickness(view);
+	struct wlr_box current_box = {
+		.x = view->pending.x - border.left,
+		.y = view->pending.y - border.top,
+		.width = view->pending.width + border.left + border.right,
+		.height = view->pending.height + border.top + border.bottom,
+	};
+	struct wlr_box dest = current_box;
+	dest.x += x;
+	dest.y += y;
+	double current_overshoot = box_layout_overshoot(&current_box);
+	double dest_overshoot = box_layout_overshoot(&dest);
+	if (dest_overshoot > 0 && dest_overshoot >= current_overshoot) {
+		if (abs(x) > 1 || abs(y) > 1) {
+			/*
+			 * It is reasonable to assume that users will specify
+			 * big numbers (like 20) for the MoveRelative x or y
+			 * value. When moving towards a layout outer edge, we
+			 * would not want the window to stop a distance away
+			 * from the edge, so recursively nudge it nearer until
+			 * it is touching.
+			 */
+			view_move_relative(view, x / 2, y / 2);
+		}
+		return;
+	}
+
 	view_move(view, view->pending.x + x, view->pending.y + y);
 }
 

```

st0rm-shad0w:
Also, I was wondering if we should allow windows to go partially off-screen, provided a minimum width stays visible? Currently, it halts right at the boundary. Letting it tuck out of the way would be super helpful for making room for other windows.

st0rm-shad0w:
@tokyo4j , @johanmalm  - I rebased and force-pushed after 0ff9af4, but the Arch build is failing with a new error now.

Edit: Fixed here - f42e189

johanmalm:
I like the one proposed here: https://github.com/labwc/labwc/pull/3552#issuecomment-4374315561
It may need some fine-tuning, but I think it's close.
How do you want to proceed? I.e. you force-push this PR; you open a separate PR; or I open a separate PR?

I don't know if you rebased against master because of the failing CI, but generally it's easier if you don't unless there is a merge conflict. It's easier to see what's changed with each push/force-push that way and thus saves reviewers' time.

st0rm-shad0w:
I force pushed it again, thinking it'd be the quickest.

Noted, will keep in mind not to rebase unnecessarily against master in future.

johanmalm:
@st0rm-shad0w Is it okay if I push some fixup-commits on this branch for discussion? I'll try to have a go tonight if okay.

st0rm-shad0w:
@johanmalm - Certainly, please go ahead. Also, happy to chime in on any discussion, if needed.

johanmalm:
I pushed a commit with what I think is a set of improvements.

I feel we still haven't found quite the right solution. I'm thinking that calculating the area outside the output-layout (before and after the move) would be a better indicate of whether or not the move should be allowed. Just relying on square-euclidean distances is okay(ish) but doesn't always result in the right behaviour when there are multiple outputs of different sizes.

To calculate the area outside the output-layout, I imagined that we could just iterate over the outputs and subtract the intersection with the original area. The slight complication is that outputs can be mirroring each other resulting in a second bit at the cherry, so I'm not sure it's worth all the faff of keeping track of all that.

st0rm-shad0w:
Thanks @johanmalm,  for refining the logic.

Since I don't have a multiple-monitor setup and don't fully understand the output-layout, struggling to visualize the issue. Considering the complexity you described,  should we revisit this after the 0.20.0 launch, or just drop it entirely?

johanmalm:
> Since I don't have a multiple-monitor setup and don't fully understand the output-layout, struggling to visualize the issue. 

You can try by running `WLR_WL_OUTPUTS=2 ./build/labwc` and then just resize the windows a bit.

Recommend `ToggleKeybinds` actions too :smile: https://labwc.github.io/labwc-actions.5.html#entry_action_name=togglekeybinds

> Considering the complexity you described, should we revisit this after the 0.20.0 launch, or just drop it entirely?

I think it's definitely worth finishing this. I see it as maturing current scoping and ironing out small issues - which is important.

Unless `0.20.0` get significantly delayed, I think it's better merge after, but work should be able to progress in parallel.

PR #3549, opened 2026-05-02
cycle: accept Enter to finish window switcher 

tokyo4j:
Fixes #3535 (in another way) by accepting Enter key to finish window switcher, as suggested in https://github.com/labwc/labwc/issues/3535#issuecomment-4309659876.

This follows Openbox's behavior.

PR #3558, opened 2026-05-04
 keyboard: simplify key state handling 

tokyo4j:
We used to store both `pressed` keys and `bound` keys and derive `pressed_sent` keys from them, which are notified to the client when switching focus. But I think this was overengineered and we can remove `pressed` keys and `bound` keys. Instead we now directly modify `pressed_sent` keys just before calling `wlr_seat_keyboard_notify_key()`.

Technically we could even remove `pressed_sent` keys as they are duplicated in wlr_seat->keyboard_state.keyboard->keycodes, but we keep `pressed_sent` keys as it is easier to access.

I confirmed this fixes the potential bug in https://github.com/labwc/labwc/pull/3549#issuecomment-4365826749.

The only place which required `bound` keys was `handle_modifiers()`. This function checked `bound` keys to keep widnow switcher alive when Alt-Tab is pressed and only the Alt modifier is released. The window switcher is then finished when Tab key is released. I replaced the check with `(seat->keyboard_group->keyboard.num_keycodes > 0)`. This slightly changes the behavior; when Alt-Tab and any other key e.g. X are pressed and Tab key and Alt modifier are released, the window switcher is now kept alive. But I don't think this breaks any workflows for users.

This PR also removes key state debugging by `LABWC_DEBUG_KEY_STATE` environment variable and renames `key-state.c` to `key-state-indicator.c`.


tokyo4j:
I'll need to see if this PR correctly handles key forwarding to IME.

PR #3564, opened 2026-05-05
Add "next"/"prev" output navigation to MoveToOutput action

roomcays:
## Summary

Add support for `output="next"` and `output="prev"` in the `MoveToOutput` action. These values iterate through `server.outputs` with wraparound.

## Rationale

Quick move window to other display (possibly "out of the way") without having to think about the direction.
Occupies single keybinding (whereas all directions may occupy up to four of them).

## Changes

 - `src/action.c` — extended `get_target_output()` to handle `output="next"` and `output="prev"` as cyclic navigation through the output list
 - `docs/labwc-actions.5.scd` — documented the new values for the `output` attribute

## Usage

```xml
<action name="MoveToOutput" output="next" />
<action name="MoveToOutput" output="prev" />
```

Created with help of Qwen3.6 (35b), Ollama and OpenCode.

### DIFF
--- a/docs/labwc-actions.5.scd
+++ b/docs/labwc-actions.5.scd
@@ -294,9 +294,11 @@ Actions are used in menus and keyboard/mouse bindings.
 	fullscreen.
 
 	If *output* is specified, the window will be sent directly to the
-	specified output and *direction* is ignored. If *output* is omitted,
+	specified output and *direction* is ignored. *output* may also be
+	"next" or "prev" to move to the next or previous output in the
+	server.outputs list, with wraparound. If *output* is omitted,
 	*direction* may be one of "left", "right", "up" or "down" to indicate
-	that the window should be moved to the next output in that direction
+	that the window should be moved to the next output in the direction
 	(if one exists).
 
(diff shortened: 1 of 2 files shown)


Consolatis:
What is the rationale for wanting this feature? We do already have the directional `left/right/up/down` + `wrap` arguments. `next` and `prev` would iterate in basically any random screen position order which might even change during the lifetime of the labwc process due to monitor re-connects / tty switches.

roomcays:
Well, you've got the point. When I was starting this I just wanted to be able to quickly move window to "any" **second** monitor (I use two), with single keybinding, not having to think about the direction. In this case, even single `next` would be sufficient. But then I thought about three+ monitors and I added `prev`.

Not specifying the direction is the actual feature.

I've added "Rationale" section to the PR's body.

Consolatis:
Doesn't `<action name="MoveToOutput" direction="right" wrap="yes" />` do the same in that case? Assuming monitors are ordered from left to right.

roomcays:
I must admit, that I've not thought about that... :sweat_smile:
In my scenario it should work just fine (with "wrap", of course). Thanks!

PR #3566, opened 2026-05-06
Deduplicate code in `cycle.c` functions `cycle_immediate()` and `init_cycle()`

johanmalm:
Cc @elviosak 

No functional change.


elviosak:
Tested a bit in a nested session, no issues here.

Consolatis:
Thanks :)

PR #3567, opened 2026-05-07
Include wlroots runtime version in --version string

Consolatis:
Also change `LAB_WLR_VERSION_AT_LEAST()` macro to use runtime information.


johanmalm:
That's great. Thanks

PR #3570, opened 2026-05-07
menu: finalize menu accelerators

ch3rn1ka:
This is a follow-up PR to #3505. The changes include:

- Separating the unescaping logic from `item_parse_accelerator()` to a new function `unescape_underscore()` to avoid the side effect of modifying the string
- Renaming `get_unicode_char()` to `get_unicode_char_lowercase()` (since it converts the char to lowercase)
- Using GLib in `get_unicode_char_lowercase()` for parsing UTF-8 codepoints instead of `mbrtoc32()` (in the previous PR I argued against it, but after some consideration I think that it *is* the simplest tool to use in this case, since we link with it anyway).

I also added some comments for better readability.

### DIFF
--- a/src/menu/menu.c
+++ b/src/menu/menu.c
@@ -2,12 +2,16 @@
 #define _POSIX_C_SOURCE 200809L
 #include "menu/menu.h"
 #include <assert.h>
+#include <ctype.h>
 #include <libxml/parser.h>
+#include <locale.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <string.h>
 #include <strings.h>
+#include <uchar.h>
 #include <unistd.h>
+#include <wctype.h>
 #include <wlr/types/wlr_scene.h>
 #include <wlr/types/wlr_xdg_shell.h>
 #include <wlr/util/log.h>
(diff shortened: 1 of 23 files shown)


src/menu/menu.c
```diff
@@ -133,42 +133,30 @@ validate(void)
...
 
-	uint32_t result = 0;
-	char32_t codepoint = 0;
-	mbstate_t state = {0};
-	size_t bytes = mbrtoc32(&codepoint, first_byte, 4, &state);
-	if (bytes > 0 && bytes <= 4) {
-		*out_bytes = bytes;
-		result = (uint32_t)towlower((wint_t)codepoint);
-	} else {
+	if (codepoint == (gunichar)-1 || codepoint == (gunichar)-2) {
```
Consolatis: Are there any constants for these two? If no please add a comment above which explains what `-1` and `-2` mean.

src/menu/menu.c
```diff
@@ -133,42 +133,30 @@ validate(void)
...
 static uint32_t
-get_unicode_char(const char *first_byte, size_t *out_bytes)
+get_unicode_char_lowercase(const char *first_byte, size_t *out_bytes)
 {
-	if (!first_byte || first_byte[0] == '\0') {
-		*out_bytes = 0;
+	if (!first_byte || first_byte[0] == '\0' || !out_bytes) {
+		if (out_bytes) {
+			*out_bytes = 0;
+		}
```
Consolatis: I don't think the `if (out_bytes)`is required here. It could be replaced with an `assert(out_bytes)` on top of the function though.

We could also rename the argument to something like `bytes_consumed` to make its purpose more clear.

src/menu/menu.c
```diff
@@ -133,42 +133,30 @@ validate(void)
...
 }
 
+/* Extract a single Unicode codepoint and convert it to lowercase */
 static uint32_t
-get_unicode_char(const char *first_byte, size_t *out_bytes)
+get_unicode_char_lowercase(const char *first_byte, size_t *out_bytes)
 {
-	if (!first_byte || first_byte[0] == '\0') {
-		*out_bytes = 0;
+	if (!first_byte || first_byte[0] == '\0' || !out_bytes) {
```
Consolatis: If we have the `assert(out_bytes)` on top this could be replaced by
```c
if (string_null_or_empty(first_byte)) {
	*out_bytes = 0;
	return 0;
}
```

Consolatis (commented):
Thanks for the PR. Some notes below:

Some of the added headers in connection with locales, wide characters, `mbrtoc32` and `towlower` could likely be removed as well.

Overall: nice cleanup!

src/menu/menu.c
ch3rn1ka: > Are there any constants for these two?

No, not that I know of. Handled this better in the recent commit.

Consolatis: Intermediate bools are fine as well. New commit accidentally uses -1 two times.

ch3rn1ka: Damn, I did overlook that. Force-pushed a quick fix to avoid commit noise.

Consolatis:
Thanks!

Issue #1686, opened 2024-04-06
Enhancement: config error reporting


droc12345:
I sometimes use hyprland, and there is a feature that I like.
It has to do with reporting config file errors. 

A sample
![](https://github.com/droc12345/images/blob/main/data/hypr-msg.jpg?raw=true)

Not sure how they implement it, but should be something similar to the osd window.
the line is a the very top of the screen, and doesn't go away till the error is fixed.

stefonarch:
In labwc you can tell to have an issue in `rc.xml` when window decoration defaults after reload, but you have to find the error in the xml file by yourself. 

I remember sway having a similar message on top which indicates the faulty line. So kind of xml syntax checker tool would be needed probably.

Consolatis:
Labwc does log various config errors but only to stderr. That usually doesn't really help when within the labwc session.

droc12345:
I knew it output to stderr, but I typically run with logging >/dev/null, unless I have a problem.
That's why it would be nice to have an on screen or popup for errors, while in session. 

Anyway, just something to think about, no immediate need for it.

Consolatis:
Implemented on the master branch by @elviosak. Not all errors do currently show up but that will be improved later on. Thus closing this issue. A related issue specific to XML syntax errors is still open at #1250.

PR #3575, opened 2026-05-09
add dialog for config errors

### LINKED ISSUE #1250: Notify user with a GUI popup about invalid config file
We only log an error currently whenever the xml from the config file  is invalid (e.g. missing closing tags, misspelled start/end tag and similar).

This is really difficult for users to detect, it looks like we revert to the default settings for unknown reasons.
Example: #1249

We could additionally show a simple popup in those cases (and maybe even lay the groundwork for future `Prompt` actions while at it).

elviosak:
closes #1250 and #1686
Added a config dialog that shows up when there are errors in the config file.

screenshot:
config_dialog

the dialog tree stays on top windows and below the popups, maybe its better to put it below, but then if u have windows open u might not see the dialog (on reconfigure).

currently only logs the checks done in `rcxml.c` and `actions.c`, i'll set it to draft for now to get some input from the maintainers if this is a good way to do it or if it needs some changes, and add more logging later.

### DIFF
--- a/include/common/node-type.h
+++ b/include/common/node-type.h
@@ -50,6 +50,7 @@ enum lab_node_type {
 	LAB_NODE_CYCLE_OSD_ITEM,
 	LAB_NODE_LAYER_SURFACE,
 	LAB_NODE_UNMANAGED,
+	LAB_NODE_CONFIG_DIALOG,
 	LAB_NODE_ALL,
 
 	/* translated to LAB_NODE_CLIENT by get_cursor_context() */
--- a/include/config/dialog.h
+++ b/include/config/dialog.h
@@ -1,0 +1,23 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LAB_DIALOG_H
+#define LAB_DIALOG_H
+
+struct wlr_scene_node;
+
+/**
+ * dialog_create_callback - creates the config dialog,
+ * used with wl_event_loop_add_idle.
+ *
+ * @data: Ignored.
+ */
+void dialog_create_callback(void *data);
+
+/**
+ * dialog_destroy - destroys a dialog that is a child of
+ * server.dialog_tree, destroys all children if NULL is provided.
+ *
+ * @dialog: Dialog to be destroyed
+ */
+void dialog_destroy(struct wlr_scene_node *dialog);
+
+#endif /* LAB_DIALOG_H */
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -17,6 +17,28 @@
 /* max of one button of each type (no repeats) */
 #define TITLE_BUTTONS_MAX ((LAB_NODE_BUTTON_LAST + 1) - LAB_NODE_BUTTON_FIRST)
 
+struct log_item {
+	char *text;
+	struct wl_list link; /* struct rcxml.error_logs */
+};
+
+/**
+ * Calls wlr_log(verb, fmt, ...) and also saves it in
+ * rc.error_logs for displaying it in a dialog later.
+ * Only sets rc.has_error when WLR_ERROR is used to log
+ * other relevant info like the config file.
+ */
+#define lab_wlr_log(verb, fmt, ...)	\
+do { \
+	wlr_log(verb, fmt, ##__VA_ARGS__); \
+	if (verb == WLR_ERROR) { \
+		rc.has_error = true; \
+	} \
+	struct log_item *log_item = znew(*log_item); \
+	log_item->text = strdup_printf(fmt, ##__VA_ARGS__);	\
+	wl_list_append(&rc.error_logs, &log_item->link); \
+} while (0)
+
 enum adaptive_sync_mode {
 	LAB_ADAPTIVE_SYNC_DISABLED,
 	LAB_ADAPTIVE_SYNC_ENABLED,
@@ -213,6 +235,10 @@ struct rcxml {
 	float mag_scale;
 	float mag_increment;
 	bool mag_filter;
+
+	/* Error log */
+	bool has_error;
+	struct wl_list error_logs; /* struct log_item.link */
 };
 
 /* defined in main.c */
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -230,6 +230,9 @@ struct server {
 	/* Tree for all non-layer xdg/xwayland-shell surfaces */
 	struct wlr_scene_tree *workspace_tree;
 
+	/* Tree for error dialog */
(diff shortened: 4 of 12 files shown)


johanmalm:
I definitely want this feature. I had in mind doing it with `labnag` though. Was there a reason you chose not to implement it with `labnag`?

Haven't thought too much about it, but maybe something like:

```C
static struct buf naglog = BUF_INIT;

void
nag_add_fmt(const char *fmt, ...)
{
	va_list ap;
	va_start(ap, fmt);
	buf_add_fmt(&naglog, fmt, ap);
	va_end(ap);
}

void
nag_show(void)
{
	char template[1024];
	snprintf(template, sizeof(template) - 1, "/tmp/labnag.XXXXXX");
	int fd = mkstemp(template);
        write(fd, naglog.data, naglog.len);
        close(fd);

	printf("%s\n", naglog.data);

	char *command = strdup_printf("labnag -m \"Errors\" -l < %s", template);
	if (!fork()) {
		execlp("sh", "sh", "-c", command, (char *)NULL);
		unlink(template);
	}
	free(command);
	buf_reset(&naglog);
}
```

I think @Consolatis had a go at implementing it somehow - by redirecting `stderr` IIRC.

elviosak:
> I definitely want this feature. I had in mind doing it with labnag though. Was there a reason you chose not to implement it with labnag?

i tried it first for testing, but couldn't get it to put one error per line (i didn't use it before, is there a way to add linebreaks?), also i prefer the "dialog" layout than the bar on top/bottom.

johanmalm:
> i tried it first for testing, but couldn't get it to put one error per line (i didn't use it before, is there a way to add linebreaks?)

Use `labnag -m Errors: -l` and then write the message to `stdin`. The message can contain `\n` for new lines. You get a `toggle details` button where you can see all the errors. This is how sway does it.

Consolatis:
My old branch is here https://github.com/labwc/labwc/compare/master...Consolatis:labwc:work_in_progress/labnag_config_errors

It is interleaved with the initial implementation of labnag though.

Edit:
Not saying that the way its implemented there is necessarily the best one. But I do think this feature should if possible not use a labwc owned interface but write to stdin of an external client. And allow the user to change the command executed so it can be further customized by the user (e.g. could be shown by `dialog`, `labnag`, written to a logfile or whatever else floats their boat). Similar to `<prompt message="foo" />`.

elviosak:
> Haven't thought too much about it, but maybe something like:

Tested it but changed a bit, it works well, writing to a file is a lot better than whatever i tried before
screenshot:
config_nag

Should i go ahead and implement it with labnag?
If so i have a few more questions:
1. Should i save labnag's pid and kill it on reconfigure?
2. Should i keep variables static in the file instead of globals or using `rc`?
3. Is there a way to get the line numbers from the xml parser?
4. Other suggestions?

johanmalm:
Looks nice.

> Should i go ahead and implement it with labnag? 

I think so. Might be worth keep a copy of your old branch just in case I'm giving bad advice.

Offloading this stuff to the `labnag` client feels like the right direction because it gives us/users freedom to play around with the UI/UX without touching the compositor too much. This might include, timeout, theme-colors, size, position, buttons and even other clients like `zenity`.

>     1. Should i save labnag's pid and kill it on reconfigure?

I think labnag is short-lived (i.e. dies after each time it's show a message following startup or reconfigure).

>     2. Should i keep variables static in the file instead of globals or using `rc`?

Unless we need to access them from somewhere else, I think it's nice encapsulation to just keep them static within that translation unit. It's the only language supported way of doing things "private" in C :smile:

>     3. Is there a way to get the line numbers from the xml parser?

For the actual libxml2 errors, I think there is a way of specifying the libxml2 error handling function. I haven't never tried it, but it might be possible to get the line numbers that way. I.e. rather than letting libxml2 just print errors to `stderr`, let us handle it.

For our own messages, it looks like we could get the line number with `xmlGetLineNo()`

https://gnome.pages.gitlab.gnome.org/libxml2/html/tree_8h.html#a44b0a9b87c1ac0ee5f9393e18faded91

>     4. Other suggestions?

For the exit yes/no prompt, we implement a config option. We ought to consider something similar for this. Can consider that as a separate PR, but thought I ought to mention it.

https://labwc.github.io/labwc-config.5.html#entry_core_promptcommand

st0rm-shad0w:
> 4. Other suggestions?

Hey @elviosak,  I was thinking about  a few points:
1. When we call labnag to show the errors, could we add the -y,  -e and -x flags as shown below so it opens above the top bar (i.e if one present)? This makes it similar to Sway's behavior.

`labnag -y overlay -e top -x`

2. Can we extend it to menu.xml as well?

3. Also, should the labnag auto close when you have fixed the issue?

elviosak:
> I think labnag is short-lived (i.e. dies after each time it's show a message following startup or reconfigure).

It's about 5 seconds by default, but it stays around if u click it. If we need to store pid/kill it should be simple to add.
Also should we increase the timer? I think about 30 seconds or so would be good. I plan to have the Close button by default too.

> For our own messages, it looks like we could get the line number with xmlGetLineNo()

Thanks that would be a good addition to the messages.

> For the exit yes/no prompt, we implement a config option. We ought to consider something similar for this. Can consider that as a separate PR, but thought I ought to mention it.

I didn't know this existed, but it should be simple to add it (might not be able to report errors in it), ~~i think the only check i'd do is if the '%s' is present.~~ changed to launch it without shell and write to pipe to get the nag's pid, so no %s.
something like `<core><errorCommand>`?

> 1. When we call labnag to show the errors, could we add the -e and -y flags as shown below so it opens above the top bar (i.e if one present)? This makes it similar to Sway's behavior.
> 
> `labnag -y overlay -e top -x`

Didn't notice any difference here in those options other than `-x`, but i think i'll add the command to config so you could just change it anyway, and the defaults we'll see after it is done.

> 2. Also, should the labnag auto close when you have fixed the issue?

That's more or less the idea on my 1st question, but it would close every time it is reconfigured, then recheck it and show again if there are errors.

st0rm-shad0w:
> It's about 5 seconds by default, but it stays around if u click it. If we need to store pid/kill it should be simple to add. Also should we increase the timer? I think about 30 seconds or so would be good. I plan to have the Close button by default too.

The behavior I have seen with other WMs is that it is persistent until fixed, with no close buttons!  Personally I feel close button should be good and time can 5 mins/300s.

elviosak:
> The behavior I have seen with other WMs is that it is persistent until fixed

Yeah i prefer having the close button, not being able to close it always annoyed me when i tried hyprland.

closing this, continuing in #3578

PR #3576, opened 2026-05-10
fix(labnag): add missing --exclusive-zone to getopt options

st0rm-shad0w:
## Summary

The long option entry for exclusive-zone `--exclusive-zone` was never added to the opts[] array, preventing getopt_long from recognizing it.


johanmalm:
Thanks

PR #3577, opened 2026-05-11
window-title: add command line argument --title

### LINKED ISSUE #1850: Is it possible to customize nested labwc window title or app_id?
Suppose that I have 10 nested labwc windows all running at once on my desktop. Right now, they are all named the same thing: `wlroots - WL-1`
If one of them needs my attention, I cannot tell them apart. I would either have to remember which order I placed them in on the taskbar, or click through each one at a time before finding the right one.
More importantly, I cannot use `wlrctl` to manage one window at a time, as all labwc windows appear identical in its output.
```
$ wlrctl toplevel list
wlroots: wlroots - WL-1
wlroots: wlroots - WL-1
lxterminal: pi@raspberrypi: ~
chromium-browser: New Tab - Chromium
```
Would it be possible to configure labwc to pick a custom window title or app_id so that multiple windows can be differentiated programmatically?

mdsib:
Introduces a new command line argument, `-t`/`--title`, that can be used to set the title of nested instances of labwc via a format string. `%o` is replaced by the output name wayland gives us.

Fixes #1850

### DIFF
--- a/docs/labwc.1.scd
+++ b/docs/labwc.1.scd
@@ -70,13 +70,13 @@ the `--exit` and `--reconfigure` options use.
 	session clients support both X11 and Wayland, this command line option
 	avoids re-writes and fragmentation.
 
-*-t, --window-title* <fmtstr>
-	Set the window title for labwc to use when nested in a compositor.
-	<fmtstr> is a format string to be used as the window title, replacing
-	`%o` with the name of the output region. This is useful when simulating
-	multiple screens, such as with running labwc with the enviornment
-	variable `WLR_WL_OUTPUTS=2`. In this case, `%o` will be unique per
-	simulated screen.
+*-t, --title* <fmtstr>
+	Set the window title for labwc to use when it is running in a window
+	(i.e. nested in a compositor). <fmtstr> is a format string to be used as
+	the window title, replacing `%o` with the name of the output
+	region. This is useful when simulating multiple screens, such as with
+	running labwc with the enviornment variable `WLR_WL_OUTPUTS=2`. In this
+	case, `%o` will be unique per simulated screen.
 
 *-v, --version*
 	Show the version number and quit
--- a/docs/labwc.1.scd
+++ b/docs/labwc.1.scd
@@ -70,6 +70,14 @@ the `--exit` and `--reconfigure` options use.
 	session clients support both X11 and Wayland, this command line option
 	avoids re-writes and fragmentation.
 
+*-t, --window-title* <fmtstr>
+	Set the window title for labwc to use when nested in a compositor.
+	<fmtstr> is a format string to be used as the window title, replacing
+	`%o` with the name of the output region. This is useful when simulating
+	multiple screens, such as with running labwc with the enviornment
+	variable `WLR_WL_OUTPUTS=2`. In this case, `%o` will be unique per
+	simulated screen.
+
 *-v, --version*
 	Show the version number and quit
 
--- a/src/main.c
+++ b/src/main.c
@@ -9,6 +9,7 @@
 #include "common/font.h"
 #include "common/macros.h"
 #include "common/spawn.h"
+#include "common/string-helpers.h"
 #include "config/rcxml.h"
 #include "config/session.h"
 #include "labwc.h"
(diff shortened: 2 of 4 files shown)


Consolatis:
In general this seems sensible. As for command line arg vs env var: I am not sure what the best approach is either. But if we go the env var route we should definitely store it once after startup and then unset the env var so we don't keep it around for client processes.

I also wonder what the best option would be to make a custom title with `WLR_WL_OUTPUTS=2` work so one can distinguish the two windows. Always appending the output name doesn't seem that great, neither does setting the two titles to the same static string.

mdsib:
> I also wonder what the best option would be to make a custom title with WLR_WL_OUTPUTS=2 work so one can distinguish the two windows. Always appending the output name doesn't seem that great, neither does setting the two titles to the same static string.

One option is to omit appending the output name when `WLR_WL_OUTPUTS` is unset or `1`, at the cost of making it more difficult to target the output programmatically. On the other hand, it might be useful to be able to target the _only_ output separately from the _first_ output.

johanmalm:
I suggest command line option `--window-title` with a `%o` specifier for the output name.

docs/labwc.1.scd
```diff
@@ -70,6 +70,14 @@ the `--exit` and `--reconfigure` options use.
 	session clients support both X11 and Wayland, this command line option
 	avoids re-writes and fragmentation.
 
+*-t, --window-title* <fmtstr>
```
johanmalm: Seeing it like this, I think `--title` would be cleaner and more consistent with the  short option (`-t`) being the first letter of the long option.

src/main.c
```diff
@@ -37,24 +38,26 @@ static const struct option long_options[] = {
...
+"  -c, --config <file>         Specify config file (with path)\n"
+"  -C, --config-dir <dir>      Specify config directory\n"
+"  -d, --debug                 Enable full logging, including debug information\n"
+"  -e, --exit                  Exit the compositor\n"
+"  -h, --help                  Show help message and quit\n"
+"  -m, --merge-config          Merge user config files/theme in all XDG Base Dirs\n"
+"  -r, --reconfigure           Reload the compositor configuration\n"
+"  -s, --startup <command>     Run command on startup\n"
+"  -S, --session <command>     Run command on startup and terminate on exit\n"
+"  -t, --window-title <fmtstr> Specify title to use when the compositor is nested\n"
```
johanmalm: Consider replacing `nested` with `running in a window` to make it more understandable for end-users.

src/output.c
```diff
@@ -605,9 +607,11 @@ handle_new_output(struct wl_listener *listener, void *data)
 	}
 
 	if (wlr_output_is_wl(wlr_output)) {
-		char title[64];
-		snprintf(title, sizeof(title), "%s - %s", "labwc", wlr_output->name);
-		wlr_wl_output_set_title(wlr_output, title);
+		gchar** parts = g_strsplit(server.window_title_fmt, "%o", -1);
```
johanmalm: It might look cleaner with something like this (not tested):

```C
GString *title = g_string_new(server.window_title_fmt);
g_string_replace(title, "%o", wlr_output->name, 0);
wlr_wl_output_set_title(wlr_output, title->str);
g_string_free(title, TRUE);
```

We have not used `GString` before though. @Consolatis @jlindgren90 @tokyo4j Any views?

johanmalm (commented):
LGTM except for a few minor comments

johanmalm:
The CI currently errors on coding style. 

```
src/output.c:610: ERROR: "foo** bar" should be "foo **bar"
src/output.c:611: ERROR: "foo* bar" should be "foo *bar"
```

You can check this locally with `./scripts/check --file=src/output.c`

src/output.c
mdsib: I agree that that's cleaner, and we aren't using `g_strjoinv` either so both add some amount of new glib surface area.

johanmalm: Yeah, let's go for it :smile:

johanmalm:
Trust you will set this to "ready for review" when ready?

src/output.c
mdsib: Ok! I was waiting for another opinion since I'm unfamiliar with the codebase and glib in general. But this seems like a pretty safe change to me.

mdsib:
Yes, it'll be ready once I address that last piece and test it out a bit.

johanmalm:
LGTM

PR #3578, opened 2026-05-11
add logger for config errors and <core><errorCommand> to display it.

### LINKED ISSUE #1250: Notify user with a GUI popup about invalid config file

elviosak:
supersedes #3575
closes #1250 and #1686

Added `nag_log` macro to replace `wlr_log` when the error message should be saved for displaying later in addition to printing it.
Added `<core><errorCommand>` to display errors on startup and reconfigure.
Added `spawn_piped_async_no_shell` function to launch `errorCommand` program, get the pid and pipe fd to write to it.
When a reconfigure is triggered, sends a SIGTERM signal to the program so we don't get multiple "nags".

Default errorCommand is : `labnag --message 'Config Error' --button-dismiss 'Close' --layer overlay --timeout 0 --detailed-message` (no timeout)

video:

https://github.com/user-attachments/assets/efbe9fb6-a6b4-4a43-9938-36928ebd606a

@johanmalm @Consolatis ~~Still missing docs and more logging (currently onliy `actions.c` and filename/parse error in `rcxml.c`)~~ Added logging in `rcxml.c` in another branch since it's a lot of changes and might be annoying to review, i'll add later, wanted someone to check it first, especially the `spawn_piped_async_no_shell` function as i had no idea how to work with forks and pipes before this, i referenced the other functions in the `spawn.c` file and i think it should be ok.

Note: While searching, i found some mentions of a problem of using `##__VA_ARGS__` twice as i did in the `nag_log` macro because it will be evaluated twice, so if `i++` is passed in the args, it would increase it twice, but i don't see this issue anywhere `wlr_log` is used, should we worry about this?
I think to "fix" it it would need a function receiving `(verbosity, _WLR_FILENAME, __LINE__, fmt, ...)` use `va_start` and `va_copy` then calling `_wlr_vlog` and a variant of `buf_add_fmt` that accepts `va_list` (needs to be created).

### DIFF
--- a/include/common/nag.h
+++ b/include/common/nag.h
@@ -1,0 +1,20 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+#ifndef LABWC_NAG_H
+#define LABWC_NAG_H
+
+#include <wlr/util/log.h>
+#include "common/buf.h"
+
+struct buf *nag_get_buf(void);
+void nag_set_error(enum wlr_log_importance importance);
+void nag_reset(void);
+void nag_show_callback(void *data);
+
+#define nag_log(verb, fmt, ...) \
+do { \
+	wlr_log(verb, fmt, ##__VA_ARGS__); \
+	nag_set_error(verb); \
+	buf_add_fmt(nag_get_buf(), fmt "\n", ##__VA_ARGS__); \
+} while (0)
+
+#endif /* LABWC_NAG_H */
--- a/include/common/nag.h
+++ b/include/common/nag.h
@@ -10,10 +10,10 @@ void nag_set_error(enum wlr_log_importance importance);
 void nag_reset(void);
 void nag_show_callback(void *data);
 
-#define nag_log(verb, fmt, ...) \
+#define nag_log(verbosity, fmt, ...) \
 do { \
-	wlr_log(verb, fmt, ##__VA_ARGS__); \
-	nag_set_error(verb); \
+	wlr_log(verbosity, fmt, ##__VA_ARGS__); \
+	nag_set_error(verbosity); \
 	buf_add_fmt(nag_get_buf(), fmt "\n", ##__VA_ARGS__); \
 } while (0)
 
--- a/include/common/nag.h
+++ b/include/common/nag.h
@@ -2,11 +2,13 @@
 #ifndef LABWC_NAG_H
 #define LABWC_NAG_H
 
+#include <sys/wait.h>
 #include <wlr/util/log.h>
 #include "common/buf.h"
 
 struct buf *nag_get_buf(void);
 void nag_set_error(enum wlr_log_importance importance);
+void nag_check_pid(pid_t exited_pid);
 void nag_reset(void);
 void nag_show_callback(void *data);
 
--- a/src/common/nag.c
+++ b/src/common/nag.c
@@ -2,7 +2,6 @@
 #define _POSIX_C_SOURCE 200809L
 #include "common/nag.h"
 #include <stdarg.h>
-#include <sys/wait.h>
 #include <unistd.h>
 #include "common/buf.h"
 #include "config/rcxml.h"
@@ -27,15 +26,21 @@ nag_set_error(enum wlr_log_importance importance)
 }
 
 void
+nag_check_pid(pid_t exited_pid)
+{
+	if (pid && pid == exited_pid) {
+		pid = 0;
+	}
+}
+
+void
 nag_reset(void)
 {
 	has_error = false;
 	buf_reset(&log_buf);
 	if (pid > 0) {
-		if (!waitpid(pid, NULL, WNOHANG)) {
-			kill(pid, SIGTERM);
-			/* waitpid() is done in a generic SIGCHLD handler in src/server.c */
-		}
+		kill(pid, SIGTERM);
+		/* waitpid() is done in a generic SIGCHLD handler in src/server.c */
 	}
 	pid = 0;
 }
--- a/src/common/nag.c
+++ b/src/common/nag.c
@@ -1,0 +1,67 @@
+// SPDX-License-Identifier: GPL-2.0-only
+#define _POSIX_C_SOURCE 200809L
+#include "common/nag.h"
+#include <stdarg.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#include "common/buf.h"
+#include "config/rcxml.h"
+#include "common/spawn.h"
+
+static struct buf log_buf = BUF_INIT;
+static bool has_error = false;
+static pid_t pid = 0;
+
+struct buf *
+nag_get_buf(void)
+{
+	return &log_buf;
+}
+
+void
+nag_set_error(enum wlr_log_importance importance)
+{
+	if (!has_error && importance == WLR_ERROR) {
+		has_error = true;
+	}
+}
+
+void
+nag_reset(void)
+{
+	has_error = false;
+	buf_reset(&log_buf);
+	if (pid > 0) {
+		if (!waitpid(pid, NULL, WNOHANG)) {
+			kill(pid, SIGTERM);
+			/* waitpid() is done in a generic SIGCHLD handler in src/server.c */
+		}
+	}
+	pid = 0;
+}
+
+static void
+nag_show(void)
+{
+	if (!has_error) {
+		return;
+	}
+	int pipe_w;
+	pid = spawn_piped_async_no_shell(rc.error_command, &pipe_w);
+	if (pid > 0) {
+		ssize_t bytes = write(pipe_w, log_buf.data, log_buf.len);
+		if (bytes < 0) {
+			wlr_log(WLR_ERROR, "Failed to write errors to process: %s",
+				rc.error_command);
+		}
+		spawn_piped_close(pid, pipe_w);
+	} else if (pid < 0) {
+		wlr_log(WLR_ERROR, "Failed to launch process: %s", rc.error_command);
+	}
+}
+
+void
+nag_show_callback(void *data)
+{
+	nag_show();
+}
--- a/src/common/nag.c
+++ b/src/common/nag.c
@@ -1,15 +1,20 @@
 // SPDX-License-Identifier: GPL-2.0-only
 #define _POSIX_C_SOURCE 200809L
 #include "common/nag.h"
+#include <assert.h>
 #include <stdarg.h>
 #include <unistd.h>
 #include "common/buf.h"
 #include "config/rcxml.h"
 #include "common/spawn.h"
+#include "labwc.h"
 
 static struct buf log_buf = BUF_INIT;
 static bool has_error = false;
 static pid_t pid = 0;
+static int pipe_w;
+static struct wl_event_source *write_notifier = NULL;
+static size_t remaining = 0;
 
 struct buf *
 nag_get_buf(void)
@@ -45,21 +50,41 @@ nag_reset(void)
 	pid = 0;
 }
 
+static int
+handle_writable(int fd, uint32_t mask, void *data)
+{
+	assert(remaining > 0);
+	wlr_log(WLR_ERROR, "Writable");
+	ssize_t bytes = write(pipe_w, log_buf.data + (log_buf.len - remaining), remaining);
+	if (bytes < 0) {
+		wlr_log(WLR_ERROR, "Failed to write errors to process: %s", rc.error_command);
+	} else {
+		remaining -= bytes;
+		if (remaining > 0) {
+			goto keep_waiting;
+		}
+	}
+
+	wl_event_source_remove(write_notifier);
+	write_notifier = NULL;
+	spawn_piped_close(pid, pipe_w);
+
+keep_waiting:
+	return 0;
+}
+
 static void
 nag_show(void)
 {
 	if (!has_error) {
 		return;
 	}
-	int pipe_w;
 	pid = spawn_piped_async_no_shell(rc.error_command, &pipe_w);
 	if (pid > 0) {
-		ssize_t bytes = write(pipe_w, log_buf.data, log_buf.len);
-		if (bytes < 0) {
-			wlr_log(WLR_ERROR, "Failed to write errors to process: %s",
-				rc.error_command);
-		}
-		spawn_piped_close(pid, pipe_w);
+		assert(!write_notifier);
+		remaining = log_buf.len;
+		write_notifier = wl_event_loop_add_fd(server.wl_event_loop,
+			pipe_w, WL_EVENT_WRITABLE, handle_writable, NULL);
 	} else if (pid < 0) {
 		wlr_log(WLR_ERROR, "Failed to launch process: %s", rc.error_command);
 	}
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -17,6 +17,7 @@
 #include "common/list.h"
 #include "common/macros.h"
 #include "common/mem.h"
+#include "common/nag.h"
 #include "common/nodename.h"
 #include "common/parse-bool.h"
 #include "common/parse-double.h"
@@ -1167,6 +1168,8 @@ entry(xmlNode *node, char *nodename, char *content)
 
 	} else if (!strcasecmp(nodename, "promptCommand.core")) {
 		xstrdup_replace(rc.prompt_command, content);
+	} else if (!strcasecmp(nodename, "errorCommand.core")) {
+		xstrdup_replace(rc.error_command, content);
 
 	} else if (!strcmp(nodename, "policy.placement")) {
 		enum lab_placement_policy policy = view_placement_parse(content);
@@ -1476,7 +1479,7 @@ rcxml_parse_xml(struct buf *b)
 	int options = 0;
 	xmlDoc *d = xmlReadMemory(b->data, b->len, NULL, NULL, options);
 	if (!d) {
-		wlr_log(WLR_ERROR, "error parsing config file");
+		nag_log(WLR_ERROR, "error parsing config file");
 		return;
 	}
 	xmlNode *root = xmlDocGetRootElement(d);
@@ -1802,6 +1805,15 @@ post_processing(void)
 				"--layer overlay "
 				"--timeout 0");
 	}
+	if (!rc.error_command) {
+		rc.error_command =
+			xstrdup("labnag "
+				"--message 'Config Error' "
+				"--button-dismiss 'Close' "
+				"--layer overlay "
+				"--timeout 0 "
+				"--detailed-message");
+	}
 	if (!rc.fallback_app_icon_name) {
 		rc.fallback_app_icon_name = xstrdup("labwc");
 	}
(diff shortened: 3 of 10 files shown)


elviosak:
sorry, analyzer caught some errors, force pushed the fix.

domo141:
I'd consider this killing a process based on a pid - as the labnag process is run async -- and probably
been short-lived, after a while the stored pid no longer refers any live process. There is (theorical)
chance that the pid is reused and SIGTERM is sent to wrong process. Probability increases when pids
wraps to 2 at 32768 and system is heavy starting new (somewhat "long")  running processes
(recent 64-bit linux system wraps pids at 4194304 (0x400000, see /proc/sys/kernel/pid_max) but it is still possible.
That said, there are these cases where process writes "pidfile" and removes those when process exits (and that removal may fail), in which case one may send signal to wrong process in that case, too.

In any case I'd consider just leaving storing pid and killing process out (and see if system floods w/ labnag processes (which would probably be bad in any case)

elviosak:
Oh i didn't know that, but i still want that feature if i can implement it, if i don't orphan the process  (`fork`ing twice) and do this before calling `kill`:
``` C
pid_t r = waitpid(pid, NULL, WNOHANG);
if (r == 0) {
    kill(pid, SIGTERM);
}
```
It would work as expected right? I did that way because i was worried about having to `wait` and blocking but it seems it's already handled in `server.c` so it should be fine?

johanmalm:
Thanks.

I tried running this nested with an error in my rc.xml (`foo` instead of a `yes` for a bool value).

I get no labnag message. Do I need to do anything else for this to work?

elviosak:
> Thanks.
> 
> I tried running this nested with an error in my rc.xml (`foo` instead of a `yes` for a bool value).
> 
> I get no labnag message. Do I need to do anything else for this to work?

i just added the log to `actions.c`, left `rcxml.c` for later to not pollute the changes since there are changes due to addition of a new config option.

johanmalm:
Haven't looked a code in detail yet, but it works well :smile:

I like sending SIGTERM to the program so we don't get multiple nags. Hadn't thought of that before, but it seems a better UX.

As we're in "soft" cooldown and this is the sort of thing that has the potential to generate side effects + debates about default `<core><errorCommand>` settings, are you happy that we aim to merge it after `labwc-0.20.0` in ca two weeks?

elviosak:
> As we're in "soft" cooldown and this is the sort of thing that has the potential to generate side effects + debates about default <core><errorCommand> settings, are you happy that we aim to merge it after labwc-0.20.0 in ca two weeks?

Sure, i don't mind. I'll keep adding the missing logs and checking issues/suggestions here.

johanmalm:
> Sure, i don't mind. I'll keep adding the missing logs and checking issues/suggestions here.

Depending on your level of git competency, we could leave that `nag_log()`s till the last minute to minimize risk of having to rebase if we get merge conflicts.

elviosak:
> Depending on your level of git competency, we could leave that nag_log()s till the last minute to minimize risk of having to rebase if we get merge conflicts.

It's pretty low, but i'll try to do it in another branch then merge back when needed, it might work.

stefonarch:
Nice work and really a missing feature, as atm you can tell from shortcts not working and default window decoration that there is an error.
As I'm used to this  feature from niri (and Hyprland): displaying the line number for the error would be handy.

domo141:
For that
```
pid_t r = waitpid(pid, NULL, WNOHANG);
if (r == 0) {
    kill(pid, SIGTERM);
}
```
that waitpid() may return 0 if the process that labwc started is still running (as zombies
gets cleaned up by `handle_sigchld()`) -- but if that process had exited it is (theoretically)
possible than the some other, later-started process got the same pid and waitpid()
returns 0 for that other process -- and then SIGTERM is sent to this other process.

That check is unnecessary, one could just `kill(pid, SIGTERM);` - it just exits -1 if there
is no process to be killed (or there is no permission to kill that process...)

elviosak:
> that waitpid() may return 0 if the process that labwc started is still running (as zombies
gets cleaned up by handle_sigchld()) -- but if that process had exited it is (theoretically)
possible than the some other, later-started process got the same pid and waitpid()
returns 0 for that other process -- and then SIGTERM is sent to this other process.

Only if the other process is also a child, that's why i'm checking it.

domo141:
> 
> Only if the other process is also a child, that's why i'm checking it.

Oh yes -- I was dumb enough not to understand that. I tried w/ arbitrary live process id and got `-1`

By checking the (linux!) manual page of `waitpid(2)` it seems that if it returns zero then there is 
child process w/ that pid running (and it is extreme unlikely (if not impossible) it exited between
waitpid() and kill() and new process w/ that pid was started in-between...)

I would not say no to this `waitpid()` - `kill()` solution...

elviosak:
> it exited between
waitpid() and kill() and new process w/ that pid was started in-between...

Yeah there's `pidfd` too to eliminate this technically possible race condition, i tried and it works well, but it requires linux 5.3 i think, not sure if anyone is running labwc in something older than that, but BSD's might have issues so i think the current check is enough.

And considering that closing the nag and triggering the reconfigure (and the check/kill) are user-driven actions i'd say it's pretty much impossible for it to kill the wrong process unintentionally.

domo141:
This change looks pretty good to me and one that seems to work ok;
it was pleasure to browse through (no bad-looking code to my taste).

----

Some comments which may be something to consider.

First whether the naming could be more specific and more "consistent"
for the purpose as it is now. For example `log_reset()` -- if one sees
just that in `server.c` one may wonder what it means...

I'd suggest to prefix all names in this context with `nag_log`
first `*.ch`, and then all symbols created. the macro `nag_log()` 
would be the only one to stay... 

I'm keeping the current names in rest of this comment (e.g. not 
naming `nag_show_callback()` as `nag_log_show_callback()` and so on...)

The call to `buf_reset()` could be moved to be called after
`nag_show()` in `nag_show_callback()` so that the memory allocated
would be freed... (and `nag_show()` could be made static in any case).

----

Then something more to think about.

After this change `log_reset()` does not necessarily need to be where
it currently is in `srv/server.c`...

... since in such a case `nag_log()` (ok, `log_set_error()`) could set that 
idle handler, and after that `nag_log()` could be used in more purposes
than nagging about configuration parsing problems...

If I did such a thing, I'd probably add one tristate (enum) variable;
init 0 and if so, set callback, value 1 if only info-level messages
seen, and 2 after error-level message seen. `nag_show()` would work pretty
much as it works now. `nag_show_callback()` would then set this tristate
back to zero.

If this were done, when to kill possible previous error_command - either
in the `log_set_error()` call which sets the callback, or in the 
`nag_show()` function just before starting a new one (probably even also
when new command is not to be run (no error-level messages)).

elviosak:
domo141 thanks for looking at it, it might be better to change to `nag.c` or `nag_log.c`, but i have to fix other stuff first.
For some reason this stopped working after a system update, i have no idea why, but haven't looked too much into it yet.

johanmalm:
> For some reason this stopped working after a system update, i have no idea why, but haven't looked too much into it yet.

After a system upgrade you sometimes have to delete the build directory and re-build (typically because of changes to libraries that the labwc binaries are linked to)

johanmalm:
Is this ready for a thorough review?

elviosak:
> Is this ready for a thorough review?

I'll try to clean build and rebase it and get it working again first, maybe change the filename/prefix to nag or nag_log, what do you think?

johanmalm:
Yes filename `nag.c` and prefix `nag_` seem good.

elviosak:
well it seems nothing was wrong in the code, the config was wrong, starting with a non escaped newline.
but i messed up the git stuff, trying to fix it now.

elviosak:
alright it's fixed now, changing it to ready to review.

include/common/nag.h
```diff
@@ -0,0 +1,20 @@
...
+void nag_set_error(enum wlr_log_importance importance);
+void nag_reset(void);
+void nag_show_callback(void *data);
+
+#define nag_log(verb, fmt, ...) \
+do { \
+	wlr_log(verb, fmt, ##__VA_ARGS__); \
+	nag_set_error(verb); \
+	buf_add_fmt(nag_get_buf(), fmt "\n", ##__VA_ARGS__); \
+} while (0)
```
Consolatis: This seems to always append messages to the buffer, including debug or info ones. I think we could use a `if (verbosity == WLR_ERROR) { buf_add_fmt() }` here instead (and potentially allow users to define the loglevel for labnag individually at a later point). And then also get rid of the `has_error` in nag.c and instead check for `log_buf.len > 0`.

include/common/nag.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <wlr/util/log.h>
+#include "common/buf.h"
+
+struct buf *nag_get_buf(void);
+void nag_set_error(enum wlr_log_importance importance);
+void nag_reset(void);
+void nag_show_callback(void *data);
+
+#define nag_log(verb, fmt, ...) \
```
Consolatis: Suggest naming it `verbosity` rather then `verb`.

src/common/nag.c
```diff
@@ -0,0 +1,67 @@
...
+		return;
+	}
+	int pipe_w;
+	pid = spawn_piped_async_no_shell(rc.error_command, &pipe_w);
+	if (pid > 0) {
+		ssize_t bytes = write(pipe_w, log_buf.data, log_buf.len);
+		if (bytes < 0) {
+			wlr_log(WLR_ERROR, "Failed to write errors to process: %s",
+				rc.error_command);
+		}
```
Consolatis: Technically this should likely compare `bytes` to `log_buf.len` and if smaller store the difference and register a write notifier for the fd via the wayland event loop with `wl_event_loop_add_fd()`.

src/config/rcxml.c
```diff
@@ -1476,7 +1479,7 @@ rcxml_parse_xml(struct buf *b)
 	int options = 0;
 	xmlDoc *d = xmlReadMemory(b->data, b->len, NULL, NULL, options);
 	if (!d) {
-		wlr_log(WLR_ERROR, "error parsing config file");
+		nag_log(WLR_ERROR, "error parsing config file");
```
Consolatis, line 1482: @johanmalm I don't remember but wasn't there some way to get the error logs of libxml? That might be useful for XML syntax errors. I think libxml spits those out to stderr but that obviously isn't all that helpful for custom log targets like labnag.

src/common/nag.c
```diff
@@ -0,0 +1,67 @@
...
+void
+nag_reset(void)
+{
+	has_error = false;
+	buf_reset(&log_buf);
+	if (pid > 0) {
+		if (!waitpid(pid, NULL, WNOHANG)) {
+			kill(pid, SIGTERM);
+			/* waitpid() is done in a generic SIGCHLD handler in src/server.c */
+		}
```
Consolatis: We could eliminate the risk of pid reuse by having the generic SIGCHLD handler do a callback to nag_log and if it detects the pid terminating reset the pid to 0. Similar to `action_check_prompt_result()`. That way when the pid is != 0 it is either still alive and we can kill it or it has just died but the generic handler hasn't yet got triggered by the wayland event loop (but AFAIR the process would then still exist as zombie and thus not allow another process to take its pid).

Consolatis (commented):
Didn't yet test but the general approach looks good. Some notes below.
The various log messages in `rcxml.c` should likely be added as well at some point but it could also be done later in a separate PR.

src/common/nag.c
domo141: Actually I forgot that zombie process thing when I speculated that there could be (very brief) moment of time where kill(2) would kill wrong process - if there is no other reaper nor signal(SIGCHLD, SIG_IGN), then there should be no way that kill(pid) would kill wrong process (as the process with pid is there but zombie).
But if in labwc sources there is already way to reap these processes (I did not (yet) look
`action_check_prompt_result()`) then IMO it would be nice to use the same way to handle
these processes too)

Consolatis: There is another reaper in `src/server.c::handle_sigchld()`. Thus that one would need to supply the pid of the reaped process to the logger which then (if the pid matches) resets its `pid` variable to `0`.

include/common/nag.h
elviosak: I did it like this only to log the filename here: https://github.com/labwc/labwc/blob/9ba3303246b1b1bf4dbf13793faa62ec74872e3b/src/config/rcxml.c#L2033
and similar for other files
so the message would say (after adding the other files):
```
read config file path/to/rc.xml
some error 
another error 
read theme path/to/theme
read menu path/to/menu
more error
```
and skip when there are no errors.

maybe if we get file/line numbers from xml it could be removed, but i think it's useful to keep the file path (and it also serves as a kind of separation from other files too).

src/common/nag.c
elviosak: Is it something like this: 

```
void
nag_check_pid(pid_t exited_pid)
{
	if (pid == exited_pid) {
		has_error = false;
		buf_reset(&log_buf);
		pid = 0;
	}
}
```
and then call it from both `CLD_EXITED` and `CLD_DUMPED` in `src/server.c::handle_sigchld()` ?

src/common/nag.c
elviosak: i'm not sure how to do that, how do i use `wl_event_loop_add_fd()`?

include/common/nag.h
elviosak: done

src/common/nag.c
Consolatis: I  don't think it should reset the state other than the pid itself. So a simple `if (pid == exited_pid) { pid = 0; }` should be enough.

And the `kill(pid)` call could then just check for `pid > 0` and doesn't need the `waitpid()` anymore.

I think the general approach to reset the state (e.g. buffer and has_error) based on a reconfigure happening rather than on the user closing labnag makes sense.

include/common/nag.h
Consolatis: Hm.. that makes sense. I wonder if we should add some `nag_set_topic()` (or a better name) instead which then gets prepended to all following log messages until a new "topic" has been set. I guess at some point it would be good to let users decide on the log level they want to see in external tools which the current design doesn't allow.

src/common/nag.c
Consolatis: An example is in `src/menu/menu.c::open_pipemenu_async()`. That one uses it for reading from a pipe but writing would be the same. So approximately something like this (completely untested):

```c
static struct wl_event_source *write_notifier = NULL;
static size_t remaining = 0;

static int
handle_writable(int fd, uint32_t mask, void *data)
{
    assert(remaining > 0);
    ssize_t bytes = write(pipe_w, log_buf.data + (log_buf.len - remaining), remaining);
    if (bytes < 0) {
        wlr_log(WLR_ERROR, ..);
    } else {
        remaining -= bytes;
        if (remaining > 0) {
            goto keep_waiting;
        }
    }

    wl_event_source_remove(write_notifier);
    write_notifier = NULL;        

keep_waiting:
    return 0;
}

[..]

static void
nag_show(void)
{
    remaining = log.buf.len;
    ssize_t bytes = write(..);
    if (bytes < 0) {
        wlr_log(WLR_ERROR, ..);
    } else {
        remaining -= bytes;
        if (remaining > 0) {
            assert(!write_notifier);
            write_notifier = wl_event_loop_add_fd(server.wl_event_loop,
                pipe_w, WL_EVENT_WRITABLE, handle_writable, NULL);
        }
    }
}
```

pid check and `spawn_piped_close(pid, pipe_w)` are missing.

elviosak: so `handle_writable` will keep being called until `wl_event_source_remove(write_notifier);` is called?

could `nag_show()` not write and only call `wl_event_loop_add_fd()` so `handle_writable()` writes everything and handles the error and pipe closing?

Consolatis: > so `handle_writable` will keep being called until `wl_event_source_remove(write_notifier);` is called?

Correct.

> could `nag_show()` not write and only call `wl_event_loop_add_fd()` so `handle_writable()` writes everything and handles the error and pipe closing?

Yes, that could also be done. Usually it is preferable to just write and only register the fd for notifications if there is actually more to write but the pipe buffer is full. In this case however it doesn't really matter as long as we can write more than a pipe can hold without blocking.

include/common/nag.h
johanmalm: The log-handler in `src/desktop-entry.c` may be a source of inspiration here.

https://github.com/labwc/labwc/blob/9ba3303246b1b1bf4dbf13793faa62ec74872e3b/src/desktop-entry.c#L27

src/config/rcxml.c
johanmalm: Yes. Catching the libxml2 errors would definitely be of value. I can implement this in a follow-up PR if helpful.

Have done a quick proof-of-concept here if we want to incorporate/test/evaluate now.

```C
#include <libxml/parser.h>
#include <libxml/xmlerror.h>
#include <stdarg.h>
#include <stdio.h>
#include <string.h>

#include "common/buf.h"

static void
error_handler(void *data, const char *fmt, ...)
{
	struct buf *buf = data;

	char message[4096] = {0};
	va_list ap;
	va_start(ap, fmt);
	int n = vsnprintf(message, sizeof(message), fmt, ap);
	va_end(ap);
	if (n < 0) {
		return;
	}

	buf_add(buf, message);
}

int
main(void)
{
	struct buf buf = BUF_INIT;

	const char xml[] = "<foo>\n<bar>baz</bad>\n</foo>\n";
	xmlSetGenericErrorFunc(&buf, error_handler);
	xmlDocPtr doc = xmlReadMemory(xml, strlen(xml), "test.xml", NULL, 0);

	printf("%s", buf.data);

	if (doc) {
		xmlFreeDoc(doc);
	}
	xmlCleanupParser();
	buf_reset(&buf);
	return 0;
}
```

```Makefile
CFLAGS = -g -Wall

CFLAGS += `pkg-config --cflags libxml-2.0 wlroots-0.20`
LDFLAGS += `pkg-config --libs libxml-2.0 wlroots-0.20`

all:
	$(CC) $(CFLAGS) -o main main.c buf.c mem.c string-helpers.c $(LDFLAGS)
```

Have used some of the `labwc` helpers in `include/common/` to make it easier to just slot in.

src/common/nag.c
elviosak: done, do i also need to do something like this in `nag_reset()`?
```
	if (write_notifier) {
		wl_event_source_remove(write_notifier);
		write_notifier = NULL;
	}
```

could consecutive config reloads or something else trigger the `assert(!write_notifier)` before `handle_writable()` is called? i tried for a bit and it didn't.

src/common/nag.c
elviosak: done

Consolatis:
Merged via #3697. Thanks!

PR #3585, opened 2026-05-15
Use camelcase also in examples

stefonarch:
Minor nitpick, and wondering about if it's case sensitive at all. Working on 
https://github.com/labwc/labwc-tweaks/compare/master...stefonarch:labwc-tweaks:add_new_settings?expand=1 I wonder if it should be also `onButton`.


johanmalm:
I'm happy to go camelcase for `onButton` and `twoFinger` across the codebase. It uses `strcasecmp()` anyway, but I think it improves readability.

Suggest changing it everywhere in one hit.

```sh
rg -i 'twofinger|onbutton' --count
NEWS.md:2
docs/rc.xml.all:2
docs/labwc-config.5.scd:5
src/config/rcxml.c:2
```

(`rg` for ripgrep)

PR #3601, opened 2026-05-27
Check formats against primary formats list

kode54:
Rather than just testing commits with a blind shotgun approach.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -85,12 +85,16 @@ uint32_t output_formats_10bit[] = {
 };
 
 static bool
-output_set_render_format(struct output *output, uint32_t candidates[], size_t count)
+output_set_render_format(struct output *output, uint32_t candidates[], size_t count,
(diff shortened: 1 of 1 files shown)


kode54:
Never mind this, it breaks headless mode.

kode54:
All fixed.

Consolatis:
Does this work with `WLR_RENDERER="pixman"`? It seems to only requests gbm buffer formats.

Also, what is the practical benefit of this PR?

Consolatis:
In general I am fine with doing our own supported format checks. But then we should likely have an own function which intersects the supported render formats with the supported scanout formats for the given output and then check if the requested pixel format is in the remaining set. `wlr_output_test_state()` does that automatically but I agree that we don't necessarily need to do all the rest of it.

I don't assume that its actually an issue in practice though because chances are high that the first candidate is already supported.

PR #3623, opened 2026-06-08
action: teach MoveTo more sophisticated x/y usage

### LINKED ISSUE #3611: labwc: Feature request - support symbolic positions like "center" / "bottom" in MoveTo action
I would like <action name="MoveTo"> to support symbolic string values for x and y attributes, for example:

    x="center" y="center" → place the window at the center of the current monitor.

    x="center" y="bottom" → place the window at the bottom center of the current monitor (above the bar if there is one).

    x="left" y="center", x="right" y="top", etc.

These tokens should align with the existing AutoPlace policies (like center, top, bottom, left, right, top-left, center-right, etc.), but within the more explicit and rule-based MoveTo action without requiring AutoPlace policy.

johanmalm:
...including:

1. Negative values (changing the frame of reference to the opposite edge)
2. Percentage of output
3. "center" window within output

Ref:
- https://openbox.org/help/Actions#MoveResizeTo

Fixes: #3611

@linux36524 - appreciate any testing you can do

### DIFF
--- a/src/action.c
+++ b/src/action.c
@@ -447,6 +447,11 @@ action_arg_from_xml_node(struct action *action, const char *nodename, const char
 		}
 		break;
 	case ACTION_TYPE_MOVETO:
+		if (!strcmp(argument, "x") || !strcmp(argument, "y")) {
+			action_arg_add_str(action, argument, content);
+			goto cleanup;
+		}
+		break;
 	case ACTION_TYPE_MOVE_RELATIVE:
 		if (!strcmp(argument, "x") || !strcmp(argument, "y")) {
 			action_arg_add_int(action, argument, atoi(content));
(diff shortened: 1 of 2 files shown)


johanmalm:
Note to self: test with multiple outputs

elviosak:
should this also allow negative percentages?

src/action.c
```diff
@@ -1054,6 +1059,54 @@ warp_cursor(struct view *view, const char *to, const char *x, const char *y)
...
+
+static struct point
+get_moveto_point(struct view *view, const char *pos_x, const char *pos_y)
+{
+	struct point point = { 0 };
+	struct border margin = ssd_thickness(view);
+	struct wlr_box max_extents = ssd_max_extents(view);
+	struct wlr_box usable = output_usable_area_in_layout_coords(view->output);
+	point.x += usable.x;
+	point.y += usable.y;
```
Consolatis, line 1075: Could be
```c
struct point point = {
	.x = usable.x,
	.y = usable.y,
};
```
(or in a single line like `struct point point = { usable.x, usable.y };`)

src/action.c
```diff
@@ -1054,6 +1059,54 @@ warp_cursor(struct view *view, const char *to, const char *x, const char *y)
...
+	int y;
+};
+
+static struct point
+get_moveto_point(struct view *view, const char *pos_x, const char *pos_y)
+{
+	struct point point = { 0 };
+	struct border margin = ssd_thickness(view);
+	struct wlr_box max_extents = ssd_max_extents(view);
+	struct wlr_box usable = output_usable_area_in_layout_coords(view->output);
```
Consolatis, line 1073: This might return an empty 0,0@0x0 box if `view->output == NULL`. We might want to catch that to prevent moving windows out of the layout due to negative coordinates later on. Maybe just a simple
```c
if (!output_is_usable(view->output)) {
	return;
}
```
at the start of the function is enough.

Consolatis (commented):
With those two addressed this LGTM. Not tested.

PR #3627, opened 2026-06-10
[RFC] CONTRIBUTING.md: add section on LLM/AI Tools

johanmalm:
For comment. Suspect we need a debate on this, but wanted to just get something down on paper as a starter for ten.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -482,7 +482,7 @@ or documentation. Humans retain full legal and technical responsibility for the
 resulting work, including any defects.
 
 We have received Pull Requests where the submitter clearly did not understand
-the code, could not reason about it, and would be unable to maintain it.  This
+the code, could not reason about it, and would be unable to maintain it. This
 is not acceptable.
 
 Any AI-generated code contribution must be compatible with the GPL-2.0-only
@@ -493,7 +493,7 @@ similar tasks, we strongly discourage AI slop. Code review effort is usually the
 project's primary bottleneck, and submitting unreviewed, AI-generated garbage
 wastes reviewer time, slows development, and harms the overall health of the
 project. Likewise, AI-generated commit messages and AI-written text in issues
-and discussions are strongly discourage and may be treated as sabotage on this
+and discussions are strongly discouraged and may be treated as sabotage on this
 basis.
 
 # Upversion
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,8 @@
 - [6. Unit Tests](#unit-tests)
 - [7. Submitting Patches](#submitting-patches)
 - [8. Native Language Support](#native-language-support)
-- [9. Upversion](#upversion)
+- [9. Tools](#tools)
+- [10. Upversion](#upversion)
 
 # How to Contribute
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -472,6 +473,29 @@ This generates a new pot file at `po/labwc.pot`
...
+
+Any AI-generated code contribution must be compatible with the GPL-2.0-only
+license.
+
+While LLMs can be useful for triage, review, clean-up, API refactoring and
+similar tasks, we strongly discourage AI slop. Code review effort is usually the
+project's primary bottleneck, and submitting unreviewed, AI-generated garbage
+wastes reviewer time, slows development, and harms the overall health of the
+project. Likewise, AI-generated commit messages and AI-written text in issues
+and discussions are strongly discourage and may be treated as sabotage on this
```
Consolatis: `s/discourage/discouraged/`

CONTRIBUTING.md
```diff
@@ -472,6 +473,29 @@ This generates a new pot file at `po/labwc.pot`
...
+
+We permit LLM and AI-assisted tooling, but only on the basis that humans remain
+fully accountable. Contributors must use an `Assisted-by:` tag in the commit
+logs if AI tools generated or significantly assisted with any part of the code
+or documentation. Humans retain full legal and technical responsibility for the
+resulting work, including any defects.
+
+We have received Pull Requests where the submitter clearly did not understand
+the code, could not reason about it, and would be unable to maintain it.  This
+is not acceptable.
```
Consolatis: Double space before the final `This`.

Consolatis:
In general I like it. One thing we could add are constraints for the PR message itself, even if the code change is "assisted" by an LLM the message and reasoning should always be written by a human in their own words. I don't feel like parsing through walls of text with a signal to noise ratio of 1:100.

We also could go further and require a contributor being able to give reason for every single changed line (and thus \*must\* fully understand \*all\* changes) but I am not sure how to formulate that correctly. It is already partly there in the negative "This is not acceptable" example but I feel like it could also be added as last sentence in the first paragraph.

Also interested in the opinions of @tokyo4j @jlindgren90 and @ahesford.

ahesford:
My personal view is this: if skilled contributors use LLMs as "force multipliers" to improve throughout or clarify their own understanding, that is perfectly acceptable. If people substitute machine correlations for their own lack of understanding, those contributions should be summarily rejected. Presenting work one doesn't understand violates the covenant inherent to the intellectual endeavor that is open-source contribution, where an essential part of the process is to debate and defend the quality of submissions. One cannot defend what one fails to understand.

I also think that the types of contributions that would be acceptable would come from contributors skilled enough to sanitize the LLM output to the point that their use would be undetectable. Essentially, the people who can make acceptable LLM-assisted contributions should be skilled enough to know when restrictions on their use do not apply or cannot be enforced.

CONTRIBUTING.md
johanmalm: Done.

CONTRIBUTING.md
johanmalm: Thanks.

johanmalm:
Thanks @ahesford. You make good points and I agree with them. Do you think the proposed patch should be amended to reflect any of this - or shall we leave it supporting reference in the PR only?

ahesford:
I think the text you have now looks good as is.

dmnks:
FWIW, if you're looking for some more inspiration, here's an AI policy that we have recently adopted in the RPM project: https://github.com/rpm-software-management/rpm/blob/master/CONTRIBUTING.md#ai-policy

It focuses on the *human* ownership aspect, and basically puts AI tooling in the same category as any code assisting technology that has been around for decades (like static analysis and refactoring tools, IDE autocompletion and so on). Feel free to reuse (any part of) it if it resonates :)

jlindgren90:
I like this. Seems fair and reasonable. (For the record: while I use LLMs heavily at my day job, I haven't used any for labwc development up to this point. I guess I treat it as a hobby and find it more enjoyable to write the code the old-fashioned way.)

johanmalm:
Thanks all for feedback. I'll go for the merge. 

@dmnks Thanks for the link; interesting read.

PR #3630, opened 2026-06-11
src/buffer.c: prevent creating invalid buffers

Consolatis:
Testcase: set title of foot to an invisible string (here LTR)

First one works and correctly sets the title, 2nd one crashes
- `printf "\x1b]0;\xe2\x80\x8e%s\x07" "some title"`
- `printf "\x1b]0;\xe2\x80\x8e%s\x07" ""`

Reported-by: Domo via IRC


domo141:
works for me -- and change does not look bad to me (someone more knowledgeable of
the code around the change could review too)

johanmalm:
LGTM

PR #3631, opened 2026-06-11
[v0.9] src/buffer.c: prevent creating invalid buffers

Consolatis:
Testcase: set title of foot to an invisible string (here LTR)

First one works and correctly sets the title, 2nd one crashes
- `printf "\x1b]0;\xe2\x80\x8e%s\x07" "some title"`
- `printf "\x1b]0;\xe2\x80\x8e%s\x07" ""`

Reported-by: Domo via IRC

-----
Simplified version of #3630 to reduce chances of causing side effects.


johanmalm:
LGTM

PR #3635, opened 2026-06-12
cursor: add scroll wheel support for window cycle

iamcheyan:
## Summary

When cycle mode is active (Alt+Tab held), scrolling the mouse wheel cycles through windows. Scroll down advances to the next window, scroll up goes to the previous window.

## Usage

1. Press and hold **Alt+Tab** to enter cycle mode
2. **Scroll wheel down** → cycle to the next window
3. **Scroll wheel up** → cycle to the previous window
4. Release **Alt** to focus the selected window

No configuration needed — always active during cycle mode.

## Implementation

The patch intercepts `WL_POINTER_AXIS_VERTICAL_SCROLL` events in `handle_axis()` when `server.input_mode == LAB_INPUT_STATE_CYCLE`. It maps:

- Negative delta (scroll up) → `LAB_CYCLE_DIR_BACKWARD`
- Positive delta (scroll down) → `LAB_CYCLE_DIR_FORWARD`

These are the same direction constants used by the keyboard handler (`handle_cycle_view_key()` in `keyboard.c`), ensuring consistent behavior between keyboard and mouse navigation.

The event is consumed (not forwarded to clients or processed as mouse bindings) when cycle mode is active.

## Files Changed

- `src/input/cursor.c` — Added scroll wheel interception in `handle_axis()`
- `docs/scroll-cycle.md` — Added feature documentation


docs/scroll-cycle.md
```diff
@@ -0,0 +1,29 @@
...
+- Negative delta (scroll up) → `LAB_CYCLE_DIR_BACKWARD`
+- Positive delta (scroll down) → `LAB_CYCLE_DIR_FORWARD`
+
+These are the same direction constants used by the keyboard handler (`handle_cycle_view_key()` in `keyboard.c`), ensuring consistent behavior between keyboard and mouse navigation.
+
+The event is consumed (not forwarded to clients or processed as mouse bindings) when cycle mode is active.
+
+## Files Changed
+
+- `src/input/cursor.c` — Added scroll wheel interception in `handle_axis()`
```
Consolatis: This is useless as a file in the doc directory. Please remove. Also please be transparent about LLM usage.

src/input/cursor.c
```diff
@@ -1441,6 +1441,17 @@ handle_axis(struct wl_listener *listener, void *data)
...
+	/* Cycle mode: scroll wheel cycles through windows */
+	if (server.input_mode == LAB_INPUT_STATE_CYCLE) {
+		if (event->orientation == WL_POINTER_AXIS_VERTICAL_SCROLL
+				&& event->delta != 0) {
+			cycle_step(event->delta < 0
+				? LAB_CYCLE_DIR_FORWARD
+				: LAB_CYCLE_DIR_BACKWARD);
+		}
+		return;
+	}
```
Consolatis, line 1498: This prevents other actions to run, for example the magnifier if bound to `logo-scroll` in the `All` context. We might want to add another consuming mouse context like "window-switcher" instead to keep the user in control rather than hardcoding the behavior.

docs/scroll-cycle.md
iamcheyan: > This is useless as a file in the doc directory. Please remove. Also please be transparent about LLM usage.

Removed. Also added Assisted-by: Claude Code <noreply@anthropic.com> in the commit message as required by CONTRIBUTING.md.

src/input/cursor.c
iamcheyan: > This prevents other actions to run, for example the magnifier if bound to `logo-scroll` in the `All` context. We might want to add another consuming mouse context like "window-switcher" instead to keep the user in control rather than hardcoding the behavior.

Hi, thanks for the review. I want to be transparent — I'm not very familiar with C. I'm mainly a Python user and I'm just interested in this compositor and wanted to implement some features I find useful. This code was written with the help of an AI assistant (Claude Code), so I may not be able to reason about every design decision as well as a core contributor would.

That said, I've reworked the patch based on your feedback: instead of hardcoding the behavior, I added a LAB_NODE_WINDOW_SWITCHER node type and a corresponding WindowSwitcher mouse context. Scroll handling during cycle mode now goes through the standard mousebind system with process_cycle_axis(), so users can override the defaults in their rc.xml. Default bindings map Scroll Up → PreviousWindow and Scroll Down → NextWindow.

Please let me know if there are further issues — happy to make any changes needed.

PR #3636, opened 2026-06-13
readme: Use `meson setup` in example to fix meson warning

nikolas:
Using meson without an explicit command emits a warning.


johanmalm:
Thanks

PR #3641, opened 2026-06-19
xwayland: fix missing panel icon with Firefox running under X11

jlindgren90:
WM_CLASS usage is not totally consistent between X11 applications. Previous behavior was to use the 1st string (instance) which worked for:

    WM_CLASS(STRING) = "geany", "Geany"
    WM_CLASS(STRING) = "xterm", "XTerm"

But then we have:

    WM_CLASS(STRING) = "Navigator", "firefox"

which gives "Navigator" as the app_id and causes Wayland panels (using wlr_foreign_toplevel) to not find any icon.

### DIFF
--- a/src/xwayland.c
+++ b/src/xwayland.c
@@ -590,15 +590,13 @@ handle_set_class(struct wl_listener *listener, void *data)
 	struct view *view = &xwayland_view->base;
 
 	/*
-	 * Use the WM_CLASS 'instance' (1st string) for the app_id. Per
-	 * ICCCM, this is usually "the trailing part of the name used to
-	 * invoke the program (argv[0] stripped of any directory names)".
-	 *
-	 * In most cases, the 'class' (2nd string) is the same as the
-	 * 'instance' except for being capitalized. We want lowercase
-	 * here since we use the app_id for icon lookups.
+	 * Use the WM_CLASS 'class' (2nd string) for the app_id. (Note
+	 * that this is often capitalized, which must be accounted for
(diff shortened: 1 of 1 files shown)


jlindgren90:
Some background here:
- #2218

But I can't recall what specific app/case benefited from using the "instance" string.

zeusgoose:
this patch fixes `sfwbar` + Pale Moon.  `StartupWMClass=Pale moon`.
https://repo.palemoon.org/MoonchildProductions/Pale-Moon/src/branch/master/palemoon/branding/official/palemoon.desktop

PM is a little odd with the class sting:
main window:
`WM_CLASS(STRING) = "Navigator", "Pale moon"`
submenus (Preferences,About):
`WM_CLASS(STRING) = "Browser", "Pale moon"`

the workaround i had was change the .desktop file `StartupWMClass=Navigator`, but submenus were borked. this patch works for submenus too

edit:
it also fixes the steam icon showing in the taskbar! even tho `StartupWMClass=` isn't defined in the .desktop file. hooray!

johanmalm:
LGTM

@spl237 I'm of a mind to merge this, but remember that you had significant input into the original icon-finding code. Any thoughts on this? Would you want to do any testing before we merge?

spl237:
My icon loading changes were all on the Wayland side and look to be orthogonal to this, so I'm probably not in a position to comment!

johanmalm:
> My icon loading changes were all on the Wayland side and look to be orthogonal to this, so I'm probably not in a position to comment!

Okay. Thanks. Just wanted to check.

PR #3647, opened 2026-06-25
support on-screen keyboard by allow IME to be on top layer

Boteium:
**TL;DR for reviewers:** "show the on-screen keyboard (OSK) above the lock screen" may seem simple but whether it actually works depends on a fragile mix of *three independent stacks* the compositor doesn't control. This PR implement the compositor part (detect the OSK, render it above the lock) and is working. 

## Why do we need this PR

To support touchscreen users. Currently touch screen unlock can be done via a custom locker (e.g. https://codeberg.org/slatian/swaylock-mobile ) or we have make compositor support OSK

This PR is attempting to address https://github.com/labwc/labwc/discussions/3640

See this PR to for more discussion about why this should not be done in swaylock
https://github.com/swaywm/swaylock/pull/462

## what this PR does 

Detect a mapped layer-shell surface whose client owns the active `zwp_input_method_v2` (`client_match`) **or** a `zwp_virtual_keyboard_v1` (`vkbd_match`), then reparent it into the per-output `session_lock_tree` so it renders and takes touch/pointer input above the lock. The compositor **cannot** keep a surface mapped after the client destroys it — that single fact causes every limitation below.

## Why it's fragile (the three stacks)

1. **swaylock speaks only raw keyboard.** It reads `wl_keyboard` and ignores   `zwp_text_input_v3` / input-method. So keys from a **virtual keyboard** reach it , but text via input-method **`commit_string`** does not .

2. **squeekboard's visibility is tied to input-method activation.** It shows  when `activate`d (or forced via DBus `SetVisible`), and hides, which destroys its surface on `deactivate`.

3. **Only one client owns the input-method.** With fcitx present, squeekboard is a pure OSK (virtual keyboard). Without it, squeekboard owns the input-method itself.

And locking *must* clear keyboard focus for security, which sends `deactivate`.

## Compatibility matrix

✅ works · ⚠️ workaround · ❌ unsupported

| IME (fcitx/ibus)? | OSK | Result | Why |
|---|---|---|---|
| Yes | squeekboard | ✅ | `deactivate` hits fcitx, not squeekboard; it stays up and types via virtual keyboard → swaylock. `vkbd_match=1`. **Detection by IME-client alone (`client_match`) fails here — the virtual-keyboard auto-detect is what saves it.** Confirmed on desktop. |
| No | squeekboard (auto-show only) | ❌ | squeekboard owns the IME; lock's focus-clear `deactivate`s it and it destroys its surface right after we elevate it. Confirmed nested. |
| No | squeekboard + DBus `SetVisible` script | ✅ | Forcing visibility keeps it mapped while deactivated, so virtual-keyboard keys reach swaylock and it unlocks. Confirmed in nested (no fcitx). This is the documented workaround. |
| Any | onboard | ❌ | XWayland, not a layer-shell surface, uses XTEST/AT-SPI — never seen by detection. |
| Any | wvkbd / other layer-shell OSK with virtual keyboard | ✅ (expected, untested) | Auto-detected via `vkbd_match` and doesn't self-hide on `deactivate`. |

## Why we are here
Our ecosystem thrives on diversity, but fragmentation is often the cost. Touchscreen users feel that cost the most.

Today, unlocking a touchscreen device requires the compositor, locker, IME, OSK, and visibility/autohide behavior to all align perfectly. A mismatch anywhere can break the experience.

This interoperability issue is long overdue for a fix. Whether the solution belongs in the locker or the compositor, it's time we addressed it.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -179,6 +179,7 @@ this is for compatibility with Openbox.
   <reuseOutputMode>no</reuseOutputMode>
   <xwaylandPersistence>no</xwaylandPersistence>
   <primarySelection>yes</primarySelection>
+  <allowIMEOnLockScreen>no</allowIMEOnLockScreen>
   <promptCommand>[see details below]</promptCommand>
 </core>
 ```
@@ -266,6 +267,22 @@ this is for compatibility with Openbox.
 	configured at launch. This enables autoscroll (middle-click to scroll
 	up/down) in Chromium and electron based clients without inadvertently
 	pasting the primary clipboard. Default is yes.
+
+*<core><allowIMEOnLockScreen>* [yes|no]
+	Allow the on-screen keyboard (OSK) belonging to the active
+	input-method client (the holder of the zwp_input_method_v2 global) to
+	be shown, and to receive pointer/touch input, above an
+	ext-session-lock-v1 lock screen (e.g. swaylock). This is an opt-in
+	exception to the lock screen normally hiding all other surfaces, and
+	exists so touch-screen users can enter their password with an external
+	OSK. labwc does not show the OSK automatically; bring it up together
+	with the locker, for example:
+
+	```
+	busctl call sm.puri.OSK0 /sm/puri/OSK0 sm.puri.OSK0 SetVisible b true && swaylock
+	```
+
+	Default is no.
 
 *<core><promptCommand>*
 	Set command to be invoked for an action prompt (*<action><prompt>*)
--- a/docs/rc.xml.all
+++ b/docs/rc.xml.all
@@ -17,6 +17,7 @@
     <reuseOutputMode>no</reuseOutputMode>
     <xwaylandPersistence>no</xwaylandPersistence>
     <primarySelection>yes</primarySelection>
+    <allowIMEOnLockScreen>no</allowIMEOnLockScreen>
     <!--
       # See labwc-config(5) for details
       <promptCommand></promptCommand>
--- a/include/config/rcxml.h
+++ b/include/config/rcxml.h
@@ -86,6 +86,12 @@ struct rcxml {
 	uint32_t allowed_interfaces;
 	bool xwayland_persistence;
 	bool primary_selection;
+	/*
+	 * Opt-in exception that lets the on-screen keyboard belonging to the
+	 * active input-method client render (and receive pointer/touch input)
+	 * above an ext-session-lock-v1 lock screen.
+	 */
+	bool allow_ime_on_lock_screen;
 	char *prompt_command;
 
 	/* placement */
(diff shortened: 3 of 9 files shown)


Boteium:
tested on an actual touchscreen.

20260626_00h05m15s_grim

ManuLinares:
Have you read the [CONTRIBUTING.md](https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md) guidelines?

https://github.com/labwc/labwc?tab=contributing-ov-file#tools

Boteium:
thanks for the comment. If I missed something so you think i didn't read it, just let me know what would you like me to do.

i'm glad we are looking into this complex issue #3640, especially since it didn't get any attention and past attempt to solve it (both here and swaylock) have stalled. i don't  expect this PR to be merged as-is since there are still plenty of rooms to improve. But this could be served as a very good starting point as my specific setup (touch screen with OSK with IME on wlroots) is basically an edge case. This is much complex than other users' as seen on swaylock issues. 

Please fell free to ask me for testing regardles if we refine this PR, draft a better spec, or go with something complete different or even tackle this in a different project. I am all very happy to assist.

ManuLinares:
I invite you to read the [CONTRIBUTING.md](https://github.com/labwc/labwc?tab=contributing-ov-file#tools) guidelines again. I'm sorry, I won't be reviewing this.

Boteium:
since you didn't give me any detail. i can only assume you want me to flag this work as AI generated. I might as well just make this clear, this commit, this PR discussion, and this very comment I am writing are all asisted by LLM especially since I am not a native speaker. The AI using is not limited to IDE but also include web search and auto correct for typing.

Now, because there is any reviewer yet. This PR is looking for reviewers, testers, or just any users with a touch screen device. Any comment is welcome.

johanmalm:
@tokyo4j Any high-levels views on this in terms of direction? 

There are a few things I don't like about the code, but I think that's secondary at this stage.

Boteium:
lets keep brainstorming for better ways to solve this. This PR is only a POC for the idea mentioned in https://github.com/swaywm/swaylock/pull/462 as suggested by swaylock community. But I think its unreliable and insecure. My test proved that it only work in some environment and I think it will be easy for a malicious app to spoof as an OSK or IME. So, yes, we are now at the conceptual phase only. I don't think any of the code will be eventually used. Some ideas and comparison:

|Method |Descrption | Pros Cons |
| :--- | :---: | ---: |
| A <br> (https://github.com/swaywm/swaylock/pull/462) | swaylock built-in keyboard | work reliably, potentially most secure, but hard-coded keyboard layout |
| B (this PR)| compositor IME/OSK auto-detect | fragile, security issue |
| C | compositor with user-defined allowed OSK | much reliable than B but also have the same security issue |
| D | compositor start an OSK subprocess | potentially both reliable and secure. but some OSK (squeekboard) assert single instance |
| E | let's keep brainstorming | ??? |

PR #3656, opened 2026-07-06
src/output.c: Force the initial modeset commit to actually happen.

KiritakeKumi:
In a QEMU/virtio-gpu environment, the existing logic causes the "needs_frame" flag to fail to trigger when using WLR_NO_HARDWARE_CURSORS=1 (software cursors), resulting in a continuously black screen.

With hardware cursors, this is masked because uploading the cursor image raises needs_frame as a side effect.

This commit ensures that the enable operation is successfully committed regardless of the cursor mode.

This issue was originally discovered in the openRuyi distribution: https://github.com/openRuyi-Project/openRuyi/issues/867

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -530,6 +530,30 @@ configure_new_output(struct output *output)
 	server.pending_output_layout_change--;
 
 	/*
+	 * Force the initial modeset commit to actually happen.
+	 *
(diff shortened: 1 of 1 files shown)


johanmalm:
LGTM. Thanks.

Consolatis:
Nice catch. Thanks!

PR #3667, opened 2026-07-15
When ending cycle, restore focus when not switching focus.

### LINKED ISSUE #3650: Escape in window switcher causes window to lose keyboard focus
### Problem description

When pressing `Escape` in the window switcher, labwc raises the original window but doesn't restore keyboard focus. This happens even when there is only one window. I sometimes used this to preview window contents without switching.

### Steps to reproduce

1. Open one or more windows (eg. foot).
2. Press `A-Tab` and hold `Alt` to open the window switcher.
3. Press `Escape`.

Expected behavior: The window that was active before cycling should have keyboard focus again.
Actual behavior: No windows have keyboard focus.

### labwc build source

Release

### labwc version

0.20.1

### labwc environment

### Distribution

Void Linux

elviosak:
fixes #3650

found 3 places where `cycle_finish(false)` is called with `grep -rn "cycle_finish("` (where this PR would cause any change):

```
void
cycle_reinitialize(void)
...
src/cycle/cycle.c:118:		cycle_finish(/*switch_focus*/ false);
```

this is the one from the issue:
```
static bool
handle_cycle_view_key(struct keyinfo *keyinfo)
...
if (keyinfo->translated.syms[i] == XKB_KEY_Escape) {
...
src/input/keyboard.c:481:			cycle_finish(/*switch_focus*/ false);
```

```
static void
reload_config_and_theme(void)
...
src/server.c:115:	cycle_finish(/*switch_focus*/ false);
```

Tested a bit and it fixes the problem, i don't think these other calls will cause any problems, and IMO it makes sense that you restore focus when not switching it, is there a reason it wasn't like that?

### DIFF
--- a/src/cycle/cycle.c
+++ b/src/cycle/cycle.c
@@ -204,7 +204,7 @@ cycle_finish(bool switch_focus)
 	struct view *selected_view = server.cycle.selected_view;
 	destroy_cycle();
 
-	seat_focus_override_end(&server.seat, /*restore_focus*/ false);
+	seat_focus_override_end(&server.seat, /*restore_focus*/ !switch_focus);
 
 	/* Hiding OSD may need a cursor change */
 	cursor_update_focus();


johanmalm:
Thanks. LGTM

Issue #1799, opened 2024-05-12
Feature request: dynamic SnapToEdge sizing


Vladimir-csp:
Idea: resizing window in snapped-to-edge state does not make it drop snapped state, but instead resizes current workspace's two edge snap regions: current and the opposite.

Consolatis:
This goes quite a bit into the "tiled window manager" direction. The issue I see with this is that it would only work for 2 windows per direction. It is also not clear how this would work if you have multiple stacked windows on top of each other that are all snapped to the same output edge.

tsujan:
Describing my opposite point of view:

When the feature requested here was implemented in KWin (recently), it was annoying to me because I relied on the window size after snapping, and the only way to restore it was logging out and in and avoiding resizing of snapped windows (which I did a lot before).

Vladimir-csp:
> This goes quite a bit into the "tiled window manager" direction. The issue I see with this is that it would only work for 2 windows per direction. It is also not clear how this would work if you have multiple stacked windows on top of each other that are all snapped to the same output edge.

Perhaps it would set size for two snapped regions on workspace (left+right edge or top+bottom edge), so any window that is snapped to that edges on workspace would be affected.

> the only way to restore it was logging out and in

An action to restore horizontal or vertical edge snap border position?

tsujan:
> An action to restore horizontal or vertical edge snap border position?

Possible.

But, frankly, I'm getting a little worried. LabWC is very stable and lightweight. For keeping it so,  adding of new futures may need to be limited to the most useful or popular ones. Not setting a limit for new features might result in a chaos.

Of course, I'm not a developer of LabWC — I'm only a user of it — but I said that based on my experience as a developer in in other places.

Consolatis:
> For keeping it so, adding of new futures may need to be limited to the most useful or popular ones. Not setting a limit for new features might result in a chaos.

That sums it up pretty good. I'd add to that that some features can be provided in a way that allow for multiple usage patterns but  others simply can not be provided this way or introduce massive complexity in the code base which makes it harder to maintain / enhance in the future.

Then there is also the user expectation aspect, if we were to implement something like requested here users may then end up choosing labwc for this feature because it fits their expectations but related expectations are not fulfilled (e.g. more than 2 windows in each direction, separate sizes per workspace, a set of "layouts" which can be cycled through, "dynamic tiling" and so on).

At the current point I'd say that labwc adding this feature is unlikely.

johanmalm:
[...]

PR #3679, opened 2026-07-31
view: adjust neighboring tiled windows on interactive resize (opt-in)

Blackseafisher1:
This PR adds opt-in, mouse-driven neighbor adjustment for edge-snapped windows: when a tiled window is resized, adjacent tiled windows adjust to keep the layout gap-free — no full tiling engine required.
Controlled by a single option, off by default (stock behavior unchanged):
```xml
<resize>
  <adjustTiledNeighbors>true</adjustTiledNeighbors>
</resize>
``` 
## What it does
Maintains a list of edge-snapped windows; after an interactive resize, same-row/column neighbors stay aligned and adjacent rows/columns are resized to preserve the layout. Edge/corner snapping becomes layout-aware, filling the remaining space around existing tiles. The tiled decoration is preserved through resize, only dropped when dragged away. Supports A|B, A|B/C, A/B|C, A/B|C/D (where | separates columns and / separates vertically stacked windows) — deliberately capped at 2x2.

### Limitations
The layout is deliberately capped at 2 columns x 2 rows. It does not support arrangements like four windows side by side in a single row (4 columns x 1 row), which is a common desire on ultrawide monitors. That case is intentionally out of scope: users who want that kind of layout usually have the hardware to run a full desktop environment instead — KDE Plasma supports custom window layouts — while anyone wanting truly arbitrary  tiling arrangements should use a dedicated tiling WM. This feature stays a mouse-driven enhancement of edge snapping, not a tiling engine.

### Mitigations
The https://github.com/labwc/labwc/pull/433 SnapToRegion feature largely mitigates the limitations, since if you want say 4 columns and 1 row, you can just specify regions for that yes it's not interactive but still it's enough for most cases.

## Why compositor-side
On Wayland, clients can't control other windows' geometry the way wmctrl/xdotool can on X11 — the compositor is the only entity that moves/resizes toplevels. So this behavior is compositor-side or not at all; it's essentially unique among lightweight Wayland compositors (KDE/GNOME ship equivalents as part of full tiling systems).

Demo (outdated, 2 behavior changes since https://github.com/labwc/labwc/commit/d30eca9265a411a8749daadd7b7f477650d0af0d#comments) : 
https://www.youtube.com/watch?v=YAArSl0uoCs
 
## Design
One unified option gates snap box, tile list, and neighbor adjustment. A state invariant (view_untile_managed()) ensures no stale tile entries. Client minimum sizes take priority over perfect grid alignment.

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -726,6 +726,15 @@ extending outward from the snapped edge.
 	invisible zones just beyond the window that serve as click targets for
 	mouse actions. Default is 8.
 
+*<resize><adjustTiledNeighbors>* [yes|no]
+	When a tiled window is interactively resized, automatically adjust
+	neighboring tiled windows to keep the layout gap-free. Also makes
+	edge-snapping fill the remaining space around existing tiled
+	windows instead of splitting the output in half. The layout
+	is intentionally limited to 2 columns x 2 rows. Client minimum size
+	hints are enforced — grid alignment may fall out of sync in edge
+	cases. Default is no.
+
 # INPUT CONFIGURATION
 
 This section describes configuration of input devices including:
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -206,6 +206,9 @@ struct server {
 	uint64_t next_view_creation_id;
 	struct wl_list unmanaged_surfaces;
 
+	/* tiled-views list (snap-to-edge managed views) */
+	struct wl_list tiled_views;
+
 	struct seat seat;
 	struct wlr_scene *scene;
 	struct wlr_scene_output_layout *scene_layout;
--- a/include/labwc.h
+++ b/include/labwc.h
@@ -206,9 +206,6 @@ struct server {
 	uint64_t next_view_creation_id;
 	struct wl_list unmanaged_surfaces;
 
-	/* tiled-views list (snap-to-edge managed views) */
-	struct wl_list tiled_views;
-
 	struct seat seat;
 	struct wlr_scene *scene;
 	struct wlr_scene_output_layout *scene_layout;
--- a/include/view.h
+++ b/include/view.h
@@ -29,6 +29,43 @@
 #define VIEW_FALLBACK_HEIGHT 480
 
 /*
+ * Tile neighbor management intentionally limits layouts to a maximum
+ * of 2 columns x 2 rows, e.g.:
+ *
+ *   A | B      A | B/C      A/B | C      A/B | C/D
+ *
+ * where | separates columns and / separates vertically stacked
+ * windows. More complex arrangements are left to dedicated tiling WMs.
+ *
+ * A window's tile_col/tile_row use:
+ *   TILE_COL_NONE / TILE_ROW_NONE    : not tiled on this axis
+ *   TILE_COL_LEFT / TILE_ROW_TOP     : left column / top row
+ *   TILE_COL_RIGHT / TILE_ROW_BOTTOM : right column / bottom row
+ *
+ * NONE = -1 sorts below the valid positions so ordinal comparisons
+ * (e.g. neighbor->tile_col > view->tile_col) work without special-
+ * casing; 0/1 are the two halves of each axis, which is what caps
+ * layouts at 2 columns x 2 rows.
+ *
+ * NOTE: Client minimum size hints are enforced — neighbor adjustment
+ * will not shrink a window below its declared minimum. In edge cases
+ * this means grid-aligned edges may fall out of sync. This is
+ * intentional: preserving client size constraints takes priority over
+ * perfect layout alignment.
+ */
+enum tile_col_pos {
+	TILE_COL_NONE = -1,
+	TILE_COL_LEFT = 0,
+	TILE_COL_RIGHT = 1,
+};
+
+enum tile_row_pos {
+	TILE_ROW_NONE = -1,
+	TILE_ROW_TOP = 0,
+	TILE_ROW_BOTTOM = 1,
+};
+
+/*
  * In labwc, a view is a container for surfaces which can be moved around by
  * the user. In practice this means XDG toplevel and XWayland windows.
  */
@@ -141,6 +178,11 @@ struct view {
 
 	/* This is cleared when the view is not in the cycle list */
 	struct wl_list cycle_link;
+
+	/* Link in tiled-views list (managed by snap-to-edge) */
+	struct wl_list tile_link;
+	enum tile_col_pos tile_col;
+	enum tile_row_pos tile_row;
 
 	/*
 	 * The primary output that the view is displayed on. Specifically:
@@ -584,6 +626,8 @@ void view_adjust_for_layout_change(struct view *view);
 void view_move_to_edge(struct view *view, enum lab_edge direction, bool snap_to_windows);
 void view_grow_to_edge(struct view *view, enum lab_edge direction);
 void view_shrink_to_edge(struct view *view, enum lab_edge direction);
+void view_adjust_neighbors(struct view *view);
+void view_untile_managed(struct view *view);
 void view_snap_to_edge(struct view *view, enum lab_edge direction,
 	bool across_outputs, bool combine);
 void view_snap_to_region(struct view *view, struct region *region);
--- a/include/view.h
+++ b/include/view.h
@@ -29,43 +29,6 @@
 #define VIEW_FALLBACK_HEIGHT 480
 
 /*
- * Tile neighbor management intentionally limits layouts to a maximum
- * of 2 columns x 2 rows, e.g.:
- *
- *   A | B      A | B/C      A/B | C      A/B | C/D
- *
- * where | separates columns and / separates vertically stacked
- * windows. More complex arrangements are left to dedicated tiling WMs.
- *
- * A window's tile_col/tile_row use:
- *   TILE_COL_NONE / TILE_ROW_NONE    : not tiled on this axis
- *   TILE_COL_LEFT / TILE_ROW_TOP     : left column / top row
- *   TILE_COL_RIGHT / TILE_ROW_BOTTOM : right column / bottom row
- *
- * NONE = -1 sorts below the valid positions so ordinal comparisons
- * (e.g. neighbor->tile_col > view->tile_col) work without special-
- * casing; 0/1 are the two halves of each axis, which is what caps
- * layouts at 2 columns x 2 rows.
- *
- * NOTE: Client minimum size hints are enforced — neighbor adjustment
- * will not shrink a window below its declared minimum. In edge cases
- * this means grid-aligned edges may fall out of sync. This is
- * intentional: preserving client size constraints takes priority over
- * perfect layout alignment.
- */
-enum tile_col_pos {
-	TILE_COL_NONE = -1,
-	TILE_COL_LEFT = 0,
-	TILE_COL_RIGHT = 1,
-};
-
-enum tile_row_pos {
-	TILE_ROW_NONE = -1,
-	TILE_ROW_TOP = 0,
-	TILE_ROW_BOTTOM = 1,
-};
-
-/*
  * In labwc, a view is a container for surfaces which can be moved around by
  * the user. In practice this means XDG toplevel and XWayland windows.
  */
@@ -178,11 +141,6 @@ struct view {
 
 	/* This is cleared when the view is not in the cycle list */
 	struct wl_list cycle_link;
-
-	/* Link in tiled-views list (managed by snap-to-edge) */
-	struct wl_list tile_link;
-	enum tile_col_pos tile_col;
-	enum tile_row_pos tile_row;
 
 	/*
 	 * The primary output that the view is displayed on. Specifically:
@@ -627,7 +585,6 @@ void view_move_to_edge(struct view *view, enum lab_edge direction, bool snap_to_
 void view_grow_to_edge(struct view *view, enum lab_edge direction);
 void view_shrink_to_edge(struct view *view, enum lab_edge direction);
 void view_adjust_neighbors(struct view *view);
-void view_untile_managed(struct view *view);
 void view_snap_to_edge(struct view *view, enum lab_edge direction,
 	bool across_outputs, bool combine);
 void view_snap_to_region(struct view *view, struct region *region);
--- a/src/config/rcxml.c
+++ b/src/config/rcxml.c
@@ -1399,6 +1399,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.resize_corner_range = atoi(content);
 	} else if (!strcasecmp(nodename, "minimumArea.resize")) {
 		rc.resize_minimum_area = MAX(0, atoi(content));
+	} else if (!strcasecmp(nodename, "adjustTiledNeighbors.resize")) {
+		set_bool(content, &rc.resize_adjust_tiled_neighbors);
 	} else if (!strcasecmp(nodename, "mouseEmulation.tablet")) {
 		set_bool(content, &rc.tablet.force_mouse_emulation);
 	} else if (!strcasecmp(nodename, "mapToOutput.tablet")) {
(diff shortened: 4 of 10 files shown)


docs/labwc-config.5.scd
```diff
@@ -724,7 +724,16 @@ extending outward from the snapped edge.
 	indicated thickness, regardless of their visual size. Borders visually
 	narrower than the minimum effective thickness will be augmented with
 	invisible zones just beyond the window that serve as click targets for
-	mouse actions. Default is 8.
+	mouse actions. 	Default is 8.
+
+*<resize><adjustNeighbors>* [yes|no]
```
johanmalm: Suggest `<resize><adjustTiledNeighbors>`

docs/labwc-config.5.scd
```diff
@@ -724,7 +724,16 @@ extending outward from the snapped edge.
 	indicated thickness, regardless of their visual size. Borders visually
 	narrower than the minimum effective thickness will be augmented with
 	invisible zones just beyond the window that serve as click targets for
-	mouse actions. Default is 8.
+	mouse actions. 	Default is 8.
```
johanmalm: I don't think this line could be included in commit.

include/labwc.h
```diff
@@ -206,6 +206,9 @@ struct server {
 	uint64_t next_view_creation_id;
 	struct wl_list unmanaged_surfaces;
 
+	/* tiled-window list (snap-to-edge managed windows) */
+	struct wl_list tile_windows;
```
johanmalm: Consider renaming this to `tiled_views` for consistency with surrounding code.

include/view.h
```diff
@@ -28,6 +28,27 @@
 #define VIEW_FALLBACK_WIDTH  640
 #define VIEW_FALLBACK_HEIGHT 480
 
+/*
+ * Tile neighbor management intentionally limits layouts to a maximum
+ * of 2 columns x 2 rows (A|B, A|B+C, A+B|C, A-B|C-D).
```
johanmalm: What do the `+` and `-` indicate here?

Also, this is slightly different compared with the comment in `rc.xml.all`

include/view.h
```diff
@@ -28,6 +28,27 @@
...
+ * More complex arrangements are left to dedicated tiling WMs.
+ *
+ * NOTE: Client minimum size hints are enforced — neighbor adjustment
+ * will not shrink a window below its declared minimum. In edge cases
+ * this means grid-aligned edges may fall out of sync. This is
+ * intentional: preserving client size constraints takes priority over
+ * perfect layout alignment.
+ */
+enum tile_position {
+	TILE_POS_NONE = -1,
```
johanmalm: Is there a reason for setting `-1`, `0`, `1` here?

include/view.h
```diff
@@ -28,6 +28,27 @@
...
+	TILE_POS_LEFT_OR_TOP = 0,
+	TILE_POS_RIGHT_OR_BOTTOM = 1,
```
johanmalm: Consider using `TILE_POS_RIGHT` and `TILE_POS_BOTTOM`. Is there a reason for combining them?

src/interactive.c
```diff
@@ -199,7 +202,7 @@ edge_from_cursor(struct seat *seat, struct output **dest_output,
 	*edge1 = LAB_EDGE_NONE;
 	*edge2 = LAB_EDGE_NONE;
 
-	if (!view_is_floating(server.grabbed_view)) {
+	if (!server.grabbed_view || !view_is_floating(server.grabbed_view)) {
```
johanmalm, line 205: Why this change? Should that be a separate commit?

src/view.c
```diff
@@ -1708,9 +1879,21 @@ view_set_fullscreen(struct view *view, bool fullscreen)
 		 */
 		interactive_cancel(view);
 		view_store_natural_geometry(view);
+		/* Remove from tile list — view takes fullscreen */
+		if (!wl_list_empty(&view->tile_link)) {
```
johanmalm: This code is repeated a lot. Consider a function for this - like `tile_list_remove()`

src/interactive.c
```diff
@@ -154,7 +154,10 @@ interactive_begin(struct view *view, enum input_mode mode, enum lab_edge edges)
 			maximized &= ~VIEW_AXIS_VERTICAL;
 		}
 		view_set_maximized(view, maximized);
-		view_set_untiled(view);
+		if (view->tile_col == TILE_POS_NONE
+				&& view->tile_row == TILE_POS_NONE) {
+			view_set_untiled(view);
```
johanmalm, line 159: Why do we use `view_set_untiled()` here rather than the new `view_untile_managed()`?

johanmalm (commented):
Thanks for contributing. I like this feature.

Ref:
- https://github.com/labwc/labwc/discussions/3596
- https://github.com/labwc/labwc/issues/1799

I have done a small amount of testing and it seems to work well.

I have not yet reviewed fully - particularly the logic of the lengthier functions, but seems fine at a glance.

Have left a few comments as a starter for ten.

docs/labwc-config.5.scd
Blackseafisher1: Fair, will rename

docs/labwc-config.5.scd
Blackseafisher1: yes sry.

include/labwc.h
Blackseafisher1: yes will do

include/view.h
Blackseafisher1: My bad, it all means the same. Will unify.

include/view.h
Blackseafisher1: -1 for NONE keeps it sorting below 0/1, so ordinal comparisons like neighbor->tile_col > view->tile_col work without special-casing and 0/1 are the two binary halves per axis, which is what caps layouts at 2x2. The ordering stays the same after the enum split.

include/view.h
Blackseafisher1: Honestly, I used one enum because it makes no functional difference. Both axes share the same ordering and the adjustment logic only does > and < comparisons, so a single type was the path of least resistance.
But you're right, it's harder to read. I'll split it. The thing is, I wouldn't go to four flat values in one enum, because then tile_col could hold TILE_POS_BOTTOM and nothing would complain. Instead I'll use two axis specific enums: TILE_COL_NONE/LEFT/RIGHT and TILE_ROW_NONE/TOP/BOTTOM. Each field then documents its own axis, the ordering stays -1 < 0 < 1, so the comparison logic doesn't change a line. Only downside is NONE has to exist once per enum, since two enums can't share an enumerator name.

src/interactive.c
Blackseafisher1: Agreed, that's an unrelated fix that came out of testing. I'll split it into its own commit.

src/view.c
Blackseafisher1: Will extract to func and use it on all 3 occurrences

src/interactive.c
Blackseafisher1: No technical reason, in that branch the view isn't a managed tile, so both do exactly the same thing. I'll switch it to view_untile_managed() for consistency.

Blackseafisher1:
After the first round of changes, I retested on a 2-monitor setup again and could not observe any regressions.

src/config/rcxml.c
```diff
@@ -1399,6 +1399,8 @@ entry(xmlNode *node, char *nodename, char *content)
 		rc.resize_corner_range = atoi(content);
 	} else if (!strcasecmp(nodename, "minimumArea.resize")) {
 		rc.resize_minimum_area = MAX(0, atoi(content));
+	} else if (!strcasecmp(nodename, "adjustTiledNeighbors.resize")) {
```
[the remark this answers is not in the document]
Blackseafisher1: its renamed to 
```xml
<resize><adjustTiledNeighbors> [yes] [no]
``` 
 There are/were no users of this so no other changes needed

johanmalm:
@Blackseafisher1

1. Could you indicate if all my original review comments have been addressed.
2. We now have 10 commits, but probably only want 2 or 3. Are you happy to do an interactive rebase to sort?

On confirmation of (1) and (2), I'll do a final, full review + some testing with multiple outputs with different scales - with a view to merge.

Blackseafisher1:
@johanmalm
- Yes, all of your original points have been addressed.
- The feature is feature-complete in my view. A number of edge cases and behavioral nitpicks were found and fixed along the way.
- I tested on two monitors, including with different output scales, and it works fine.
- Yes, happy to rebase to 3 commits: the feature, the edge_from_cursor NULL-guard fix you asked me to keep separate, and the docs.

Blackseafisher1:
@johanmalm 
Rebase done - ready for review. Also retested the rebased codes binary.

johanmalm:
Thanks. I'll try to review + test tomorrow.

johanmalm:
I have tested quite a bit and it seems to work really well.

Testing included:

- multiple outputs at different scales
- setting `<core><gap>` to a value greater than one
- running without `<resize><adjustTiledNeighbors>` to check for regressions (none found)

Having reviewed _all_ the code, there are a few more points that I'm keen to explore:

1. Could we do this without `server->tiled_views`? Having another list of views complicates the code quite a bit. There is the construct `for_each_view(view, &server.views, LAB_VIEW_CRITERIA_CURRENT_WORKSPACE) { ... }`. Presumably we could just iterate with that one and do a `if (!view->tiled) { continue }` or something like that?

2. There is already `view->tiled` which is an `enum lab_edge`. Give this state variable, do we really also need `enum tile_col_pos tile_col;` and `enum tile_row_pos tile_row;`? 

3. Regarding the API of `find_opposite_tiled_view(struct view *view, struct output *output, const struct wlr_box *usable, bool horizontal, bool find_in_first_half)`, I think we should explore if `bool horizontal, bool find_in_first_half` could be simplified to `enum lab_edge target`.

If I am thinking straight about the above three, these could potentially serve to simplify the logic and increase readability. 

Ref:
- https://github.com/labwc/labwc/blob/b58dc05d7f5c0a3e4ab075e03c9d663149e16630/include/common/edge.h#L33-L62
- https://github.com/labwc/labwc/blob/b58dc05d7f5c0a3e4ab075e03c9d663149e16630/include/common/edge.h#L64

Blackseafisher1:
@johanmalm

All your points were addressed but before i push:

1. separate commit on top for the diff or rebase into the feature commit directly? (my plan was to first normal commit and after your review rebase)

2. one deliberate deviation: kept the explicit workspace check instead of LAB_VIEW_CRITERIA_CURRENT_WORKSPACE. only difference: on output-layout changes (e.g. unplugging a monitor), tiled windows on an inactive workspace would have their geometry recomputed to a plain 50/50 split instead of keeping their layout.  If you  want it the other way I'll change it. 

3. bugfix: maximized views are no longer used as reference windows when computing snap boxes, since their geometry no longer reflects the tile layout (before keyboard maximize action while it's tilled would cause a stale state). minimized/fullscreen are now filtered explicitly instead of by list membership.

4. found while testing: snapping to the same edge as an existing column/row didn't take its width/height when the opposite column/row was empty - only corner snaps did. fixed that, so it now matches the width/height of what's already there instead of a fresh 50/50. reduces weird behavior: before, the new window would just take 50%, and if there was already a view tiled underneath it, resizing it would leave them out of sync, which isn't really desirable i think.

5. 1 other thing.  In your Book, is it worth guarding resizing the edge of a tiled window that's touching the outputs edge  e. g. users snaps to top right corner and now resizes top border to be smaller or bigger. Should we proactively prevent this in view_adjust_neighbors().

anything else before I push?

johanmalm:
> anything else before I push?

Oh, I don't think I've ever been asked that before :smile: It gives me the feeling of 'point of no return'.

Consider pushing to a separate branch so that we can go back - unless you're convinced that the end-result is better. The three things I suggested were just intuitive areas I'd explore - if I was doing the coding - to try to simplify the code to increase readability. There might be code-cleanliness side-effects that aren't yet obvious to me.

This feature will add a level of complexity which I'm fine with, but some of the functions in `view.c` are quite lengthy, and not easy to read at a glance, which normally in an indication to simplification, refactoring and/or identifier renaming are worth exploring. Hopefully you see where I'm aiming.

>     1. separate commit on top for the diff or rebase into the feature commit directly? (my plan was to first normal commit and after your review rebase)

Yes, we normally do "[fixup]" commits during this type of discussion, and then do an interactive rebase at the end. Happy to go that route, but the reason I suggested a separate branch were

a. The changes I suggested could be quite deep and intrusive
b. Unless you're quite comfortable with interactive-rebase, it can get quite messy with this level of change.

>     2. one deliberate deviation: kept the explicit workspace check instead of LAB_VIEW_CRITERIA_CURRENT_WORKSPACE. only difference: on output-layout changes (e.g. unplugging a monitor), tiled windows on an inactive workspace would have their geometry recomputed to a plain 50/50 split instead of keeping their layout.  If you  want it the other way I'll change it.

I'm fine with this. 

It's not obvious to me why avoiding the `LAB_VIEW_CRITERIA_CURRENT_WORKSPACE` approach helps, so would be worth adding a comment in the code to explain that and avoid someone refactoring it the future.

>     3. bugfix: maximized views are no longer used as reference windows when computing snap boxes, since their geometry no longer reflects the tile layout (before keyboard maximize action while it's tilled would cause a stale state). minimized/fullscreen are now filtered explicitly instead of by list membership.

Great.

>     4. found while testing: snapping to the same edge as an existing column/row didn't take its width/height when the opposite column/row was empty - only corner snaps did. fixed that, so it now matches the width/height of what's already there instead of a fresh 50/50. reduces weird behavior: before, the new window would just take 50%, and if there was already a view tiled underneath it, resizing it would leave them out of sync, which isn't really desirable i think.

Good catch. I don't think I tested that! 

>     5. 1 other thing.  In your Book, is it worth guarding resizing the edge of a tiled window that's touching the outputs edge  e. g. users snaps to top right corner and now resizes top border to be smaller or bigger. Should we proactively prevent this in view_adjust_neighbors().

Suggest keeping it as simple as possible, and not getting stuck into that at this point.

Blackseafisher1:
@johanmalm 

I opted for the commit on top. I am pretty sure it's way better now, because of your suggestions. 

Also I reconsidered avoiding LAB_VIEW_CRITERIA_CURRENT_WORKSPACE - it was wrong. Introducing confusion is not worth handling the minimal, non-critical edge case of resetting tile dimensions on output actions like unplugging a monitor on non-active work spaces.

I had no disagreements so everything was changed as you suggested.

johanmalm:
Thanks. Can we leave this one till after `0.20.2` in ca two weeks' time?

ManuLinares:
nice, it doesn't seem to work when using `<keybind key="W-KP_9"><action name="SnapToRegion" region="top-right" /></keybind>`, and the other corners

maybe this is because I have:
```
	<regions>
		<region name="bottom"       x="0%"  y="50%" height="50%"  width="100%" />
		<region name="bottom-left"  x="0%"  y="50%" height="50%"  width="50%"  />
		<region name="bottom-right" x="50%" y="50%" height="50%"  width="50%"  />
		<region name="left"         x="0%"  y="0%"  height="100%" width="50%"  />
		<region name="center"       x="0%" y="0%" height="100%"  width="100%"  />
		<region name="right"        x="50%" y="0%"  height="100%" width="50%"  />
		<region name="top"          x="0%"  y="0%"  height="50%"  width="100%" />
		<region name="top-left"     x="0%"  y="0%"  height="50%"  width="50%"  />
		<region name="top-right"    x="50%" y="0%"  height="50%"  width="50%"  />
	</regions>
```

Blackseafisher1:
@ManuLinares

That is expected and in fact it would be a bug if it worked. This feature does not change how regions behave (`SnapToRegion`).

You should simply use keybinds with `SnapToEdge` like this and change keys if you want:
```rc.xml
     <keybind key="W-Left"><action name="SnapToEdge" direction="left" combine="yes" /></keybind> 
     <keybind key="W-Right"><action name="SnapToEdge" direction="right" combine="yes" /></keybind>
     <keybind key="W-Up"><action name="SnapToEdge" direction="up" combine="yes" /></keybind>
     <keybind key="W-Down"><action name="SnapToEdge" direction="down" combine="yes" /></keybind>
 ```

Issue #2394, opened 2024-12-01
With 150% scale, the sizes of server-side components looks inconsistent


tokyo4j:
### Problem description

When I set output scale to 150%, the width of window borders is either 1 pixel or 2 pixels, without aliasing:

![20241201_15h25m44s_grim](https://github.com/user-attachments/assets/74871c46-4c38-46b6-a2c4-d294cab50321)

Also, there is a 1-pixel gap in menu items and menu borders:

![20241201_15h26m36s_grim](https://github.com/user-attachments/assets/2ccfdec7-13d7-4a34-8c23-1ec195a9d8e0)

### Steps to reproduce

Run labwc, set output scale to 150% with kanshi, and open windows or menus

### labwc build source

Local build

### labwc version

0.8.1

### labwc environment

Nested on Wayland

tokyo4j:
Fixed by #2809

PR #3681, opened 2026-07-31
Round menu corners using rc.corner_radius

saeedtahmtan:
Add a shared helper, create_rounded_rect_bg(), that renders a filled rounded rectangle via Cairo into a lab_data_buffer. Uses rc.corner_radius from the theme config (no new config options added).

- include/common/graphic-helpers.h: forward-declare wlr_scene types, declare create_rounded_rect_bg()
- src/common/graphic-helpers.c: implement create_rounded_rect_bg() with fallback to plain scene rect on buffer creation failure
- src/menu/menu.c: use create_rounded_rect_bg() for item and title backgrounds; rewrite menu container background to draw rounded border + fill in a single Cairo buffer
- Separators remain rectangular

All of this pr is created with AI I need a review for that because I'm not familiar with codebase but it does have now rouned corners on the menu.
image

I don't have any problem with not accepting this because readme clearly said that no rounded corner.

### DIFF
--- a/include/common/graphic-helpers.h
+++ b/include/common/graphic-helpers.h
@@ -7,6 +7,9 @@
 #include <cairo.h>
 
 struct wlr_fbox;
+struct wlr_scene_buffer;
+struct wlr_scene_node;
+struct wlr_scene_tree;
 
 /**
  * Sets the cairo color.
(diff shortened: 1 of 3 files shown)


johanmalm:
Thanks for contributing. That does look nice.

It has been tried before in https://github.com/labwc/labwc/pull/2352

The conclusion was that https://github.com/labwc/labwc/issues/2394 is very hard to avoid with rounded corners.

Appreciate your views on that. If we can't find a way through the 150% scale issue, then it's a no.

Will convert to draft, so I don't merge my mistake. It would need to be made configurable/optional as a minimum anyway.

Cc: @tokyo4j

saeedtahmtan:
I have tested it with 1.5 scalling it didn't have the issue.
But I have another issue. the border should render on top of the items I will figure out that and update the pr.(it is not very important but make it much cleaner)
In my own case it didn't happened is it possible to happen in other resolotions or other environment?

tested with this
```
wlr-randr --output eDP-1 --scale 1.5
```

tokyo4j:
I'm not objecting to this feature (it looks good!), but let me note some challenges I had when I last tried to implement this feature:

1. Complexity

For example, I assumed separators with a label (e.g. `<separator title="Application">`) should not have rounded corners or padding. But it made the code more complicated because we should then round the upper/lower corners only when the separator should appear at the top/bottom in the menu. And of course, adding drawing logic with cairo inherently increases the code complexity.

2. Auto-scaling

I assumed we should always render rounded borders in the current output scale (which this PR currently doesn't), like we are doing for font rendering with `scaled_font_buffer`. Therefore I introduced auto-scaling `scaled_rect_buffer` with rounded corners in #2352. But it introduced non-negligible complexity and a minor regression (#2394). Perhaps #2394 can be fixed if we render both menus and items in cairo buffers like this PR currently does, though. Another way to address #2394 is to render the entire menu in one cairo buffer, but it would increase the implementation complexity, too.

Note that we already have the same problem with titlebar borders (#593) and I believe this should be fixed.

saeedtahmtan:
Thank you for lightening me up. sorry if I'm not responding that is because of internet situation which iran have right now.
I'm looking for a way to get scale from the output but I couldn't figure out it. if someone give me a hint that would be great.

PR #3683, opened 2026-08-01
overlay: <snapping><overlay><placement>[above|below]

Blackseafisher1:
Goes well with the tiling feature: https://github.com/labwc/labwc/pull/3679

Draw  snap-overlay above instead of below the dragged window:

```xml
<snapping>
  <overlay>
    <placement>above</placement>
  </overlay>
</snapping>
```
default: below


docs/labwc-config.5.scd
```diff
@@ -529,6 +529,10 @@ extending outward from the snapped edge.
 	Show an overlay when snapping a window to an output edge.
 	Default is yes.
 
+*<snapping><overlay><alwaysOnTop>* [yes|no]
```
johanmalm: `alwaysOnTop` doesn't sounds quite right. Suggest merely `onTop`

docs/rc.xml.all
```diff
@@ -169,6 +169,11 @@
     <cornerRange>50</cornerRange>
     <overlay enabled="yes">
       <delay inner="500" outer="500" />
+      <!--
+        Draw the snap overlay above all windows instead of below the
+        window being dragged.
+        <alwaysOnTop>yes</alwaysOnTop>
```
johanmalm: In `rc.xml.all` we show the default values - not commented out.

Happy to include a comment above the element, but suggest element should just be `<onTop>no</onTop>` outside of the comment block.

johanmalm (commented):
I think this is a good idea.

I have also noticed - at times - that the overlay is hard to see. Maybe we should have called it underlay :smile:

Have not yet tested, but have left a couple of minor comments.

Can test with a view to merge tomorrow.

Blackseafisher1:
ahh man i have typo in my commit msg but don't want to trigger the CI with another  force push is it fine if it just stays ? or shall i force push with corrected typo 

> underneth -> underneath

johanmalm:
Yes, now that we've noticed it, I prefer a force-push.

johanmalm:
Further comments:
1. For consistency with the rest of the `rc.xml` syntax, it should be `onTop` (lower case `o`) rather than `OnTop`
2. I prefer to name the variable as closely as possible to the config option, `snap_overlay_on_top` rather than `snap_overlay_always_on_top`.
3. `if (!strcasecmp(nodename, "alwaysOnTop.overlay.snapping")) {` still has `always` in it. Which made the patch not work when I first tested it.

src/overlay.c
```diff
@@ -40,11 +40,24 @@ show_overlay(struct seat *seat, struct theme_snapping_overlay *overlay_theme,
 		opts.border_width = overlay_theme->border_width;
 	}
 
+	/*
+	 * By default the overlay is drawn below the dragged view so it
+	 * stays under other windows. With <snapping><overlay
+	 * alwaysOnTop="yes"> it is placed in the preview tree above all
+	 * windows instead.
+	 */
```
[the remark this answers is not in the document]
Consolatis: What is this Copilot crap and why does it show up here?

> AI review requested due to automatic review settings

Did we enable this?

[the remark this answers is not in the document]
Blackseafisher1: I had a setting on apparently in my GitHub settings leading to this automatic Copilot review disabled it now. Seems like it was open by default. I had like redeemed this free GitHub student pack a while ago and there it must have enabled this thing so i disabled it now should not do this again 

image

> Did we enable this?

So, it was my fault not anything you guys enabled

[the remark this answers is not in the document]
Consolatis: Got it, thanks for checking!

johanmalm:
Is this ready for another review?

Blackseafisher1:
@johanmalm 

Ahh, yes

docs/labwc-config.5.scd
Consolatis: What about `above` instead? Or some key which takes the values `above|below` rather than `yes|no`?

`onTop` reminds me of window stacking rather than UI elements.

Blackseafisher1: Maybe with the key `<drawRelativeToDragged>` ? It's a bit long but the user only has to write under or above so and this makes it clear with no ambiguity

Consolatis: We could also go with a simple `placing="above|below"` / `<placing>above|below</placing>`. Or maybe `placement="above|below"`.

`layer` is kind of overloaded, `z-direction` feels a bit too technical. `drawRelativeToDragged` a bit too complicated.

I'd vote for `placement="above|below"` unless someone comes up with a better option.

Blackseafisher1: Ok, I'm fine with this: `<placing>above|below</placing>`

johanmalm: In the interest of consistence, I suggest either `position` or `placement` rather than `placing`. Most config keys are nouns rather than verbs with `ing` at the end.

My personal preference is `<position>above|below</position>`

Blackseafisher1: I'd like to just do the changes to: `<placement>above|below</placement>` now, if you guys don't have any objections.

johanmalm: Fine with me

Consolatis: Sounds fine. Thanks for your flexibility. User-facing config options are very annoying to change after the fact. This one allows to add more states in the future if they come up and is self-explaining.

johanmalm:
We are currently in a cool-down period for the release of `0.20.2` in ca two weeks' time, so if you're okay with it, I'd prefer to merge after that.

Blackseafisher1:
Yeah yeah, i saw the announcement already. Ofc bug fixes and such have priority.

johanmalm:
Looks good apart from the following:

In `rcxml.c`, should we still have this in there?

```
} else if (!strcasecmp(nodename, "onTop.overlay.snapping")) {
		set_bool(content, &rc.snap_overlay_on_top);
```



In `rc.xml.all`, this comment could be improved:

```
     <!--
        Draw the snap overlay above instead of below the
        window being dragged.   -->
```

Suggest a simpler

```
<!-- Specify whether to draw the snap overlay above or below the window being dragged -->
```

johanmalm:
Thanks

PR #3685, opened 2026-08-03
output: keep the working render format when every candidate fails

Andy1210:
### What

output_set_render_format() tries each candidate format on output->pending and
returns false if none of them passes. The last one it tried stays on the state
and nothing takes it off, so from then on no swapchain can be created for that
output: every frame fails with "Failed to pick primary buffer format" and
nothing is drawn.

### How

Clear the WLR_OUTPUT_STATE_RENDER_FORMAT bit when every candidate fails, so the
state doesn't change the format at all. Plus a debug line, since a probe that
rejects every candidate is otherwise invisible. It's gated on the existing
`silent` flag, because output_state_setup_hdr() runs twice per output on a
config apply and the second pass is meant to be quiet.

### Testing

Raspberry Pi 5 (vc4 + v3d), labwc 0.20.0 on wlroots 0.20.2.

I ran into this with a locally patched wlroots that hands the surface above a
fullscreen video to the backend as a wlr_output_layer. That made a render format
probe fail on a box where one never had before, and the screen went black:
"Failed to pick primary buffer format" into "Failed to create swapchain" into
"Failed to build output state", hundreds a second.

The bug isn't layer specific though. Anything that makes a probe fail ends up
there, layers are just how I got to it.

With the fix, same box: a 4K film plays with the app UI over it, both on their
own hardware planes, 0 dropped and 0 delayed frames, no format errors.

An earlier version of this PR also added layer bookkeeping to labwc's own output
states. That turned out to be papering over something my own wlroots patch did,
so it's gone. The branch name is left over from it.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -85,7 +85,8 @@ uint32_t output_formats_10bit[] = {
 };
 
 static bool
-output_set_render_format(struct output *output, uint32_t candidates[], size_t count)
+output_set_render_format(struct output *output, uint32_t candidates[],
+		size_t count, bool silent)
 {
 	for (size_t i = 0; i < count; i++) {
 		wlr_output_state_set_render_format(&output->pending, candidates[i]);
@@ -93,6 +94,16 @@ output_set_render_format(struct output *output, uint32_t candidates[], size_t co
 			return true;
 		}
 	}
+	if (!silent) {
+		wlr_log(WLR_DEBUG, "output %s supports none of the %zu candidate"
+			" render formats", output->wlr_output->name, count);
+	}
+	/*
+	 * Leaving the last candidate tried on the pending state means no
+	 * swapchain can be created for this output at all, so clear the flag
+	 * and leave the format alone instead.
+	 */
+	output->pending.committed &= ~WLR_OUTPUT_STATE_RENDER_FORMAT;
 	return false;
 }
 
@@ -174,14 +185,14 @@ output_state_setup_hdr(struct output *output, bool silent)
 		hdr_succeeded = true;
 	} else if (render_bit_depth == LAB_RENDER_BIT_DEPTH_10) {
 		hdr_succeeded = output_set_render_format(output, output_formats_10bit,
-			ARRAY_SIZE(output_formats_10bit));
+			ARRAY_SIZE(output_formats_10bit), silent);
 		if (!hdr_succeeded) {
 			if (!silent) {
 				wlr_log(WLR_INFO, "No 10 bit color formats"
 					" supported, HDR disabled.");
 			}
 			if (!output_set_render_format(output, output_formats_8bit,
-					ARRAY_SIZE(output_formats_8bit))) {
+					ARRAY_SIZE(output_formats_8bit), silent)) {
 				if (!silent) {
 					wlr_log(WLR_ERROR, "No 8 bit color formats"
 						" supported either!");
@@ -190,7 +201,7 @@ output_state_setup_hdr(struct output *output, bool silent)
 		}
 	} else {
 		if (!output_set_render_format(output, output_formats_8bit,
-				ARRAY_SIZE(output_formats_8bit)) && !silent) {
+				ARRAY_SIZE(output_formats_8bit), silent) && !silent) {
 			wlr_log(WLR_ERROR, "No 8 bit color formats supported!");
 		}
 	}


johanmalm:
It looks reasonable, but this is not logic that I am particularly familiar with, so I guess we've got three options:

1. You convince me that you understand it well.
2. I try to get help from someone more competent than me to review this.
3. I spend time studying it until I can comfortably review it.

In the meantime, just two small questions:

a. Is sway working as it should?
b. Looking at sway's code, they do not appear to add `struct wlr_output_layer_state *layer_states;` and `size_t layer_states_len;` to their `sway_output` struct. Do we understand what is different about their implementation to avoid this?

johanmalm:
Should have said thank you for working on this. So thanks you :smile:

Also, Cc @spl237 @cillian64 in case you're able to help review - although I don't suspect this is quite generic and not related to RPi.

spl237:
We've not actually released 0.20 for Raspberry Pi yet - I only packaged it last week, and it is still undergoing internal testing at the moment. We do apply some of our own patches to both wlroots and labwc, so those being missing might perhaps be an issue - @cillian64 is better placed to advise than me on that point.

Andy1210:
Thank you both for looking at it — and @johanmalm, your question (b) turned out to
be the important one. Chasing it down shrank the patch.

## (b) What sway does differently

sway builds a **fresh** `wlr_output_state` for every frame — `wlr_output_state_init(&pending)`
in `output_repaint()` (`sway/desktop/output.c`), finished before the function returns —
and its render-format probing happens on a *separate* config state in
`sway/config/output.c`, one that has never carried scene layers.

labwc uses **one long-lived `output->pending` for both**: `wlr_scene` builds a frame
into it through `lab_wlr_scene_output_commit()`, and `output_set_render_format()`
probes on that same struct. So in labwc a state can still be carrying
`WLR_OUTPUT_STATE_LAYERS` — with the scene's layer array — when a probe re-tests it
later, and `output_basic_test()` then compares that array's length against the
output's actual layer count. In sway no single state is ever in both roles, which is
why there is nothing for a `layer_states` array to solve there.

## And the consequence: I have dropped that half of the patch

Having understood the difference, I went back and tested labwc with **only** the
render-format restore and no layer handling at all — same wlroots build, same
hardware, a 4K film with a fullscreen translucent UI over it:

* 0 `Failed to pick primary buffer format`, 0 failed commits, 0 `All output layers
  must be specified`
* the plane offload still engages: three DRM planes in use (video on the primary,
  the UI on an overlay, cursor on its own)
* 0 dropped and 0 delayed frames over the measurement, compositor GPU time 0%

So the layer half was not needed. It was compensating for something an earlier
iteration of my own wlroots patch did — it kept its output layer alive between
offload episodes, where it now destroys it as soon as it stops offloading. My
mistake, and I would rather find it than have you review it.

**The PR is now the render-format fix alone**: `output_set_render_format()` remembers
the format that was working before it started asking, and puts it back when every
candidate fails. Six lines and a comment. That should also answer your (1)/(2)/(3) —
there is no longer much to be an expert in.

Worth saying plainly: this half is **not layer-specific**. When every candidate
fails, `output->pending` is left carrying the last one tried, and from then on no
swapchain can be created for that output — every message afterwards blames a colour
format the renderer never supported, which is a long way from whatever actually made
the probe fail. Any cause of a failed probe reaches it. Output layers are just how I
got there.

## (a) Is sway working as it should?

I have not run sway, so I will not claim either way — only what its code shows above.
Everything I measured is on labwc 0.20.0.

## The local patches, since the PR mentions them

They are public, with the reasoning in each commit message:

**https://github.com/Andy1210/tvbox/tree/main/scripts/patches**

| Patch | What it is |
| --- | --- |
| `wlroots-0001` | `pixel_format_has_alpha()` answers from a list naming four of the 58 YCbCr formats `pixel_format_is_ycbcr()` knows. `P030` — what a Pi 5 decodes 10-bit video into — is absent while `P010` and `NV12` are present, so the video buffer reads as possibly-translucent and every opacity-gated path refuses it. |
| `wlroots-0002` | Image descriptions are refused on the libliftoff interface: the check is spelled `iface != &atomic_iface`, and `liftoff_iface` is not that even though libliftoff drives the same atomic API. Every commit carrying one is rejected — and labwc attaches one to all of them, so nothing is presented at all once `WLR_DRM_FORCE_LIBLIFTOFF=1` is set. |
| `wlroots-0003` | Publishes whether a backend can display output layers, next to the existing `features.timeline`. |
| `wlroots-0004` | The feature itself: `wlr_scene` hands the surface above a fullscreen video to the backend as a `wlr_output_layer`. |
| `wlroots-0005` | libliftoff searches plane assignments against a 1 ms deadline and reports none when it expires. |
| `wlroots-0006` | The composition fallback is armed even when nothing needs it, and where the primary is the only plane at zpos 0 it takes the plane the video needs. |
| `labwc-0001` | This PR. |

**None of them is required for this PR.** `labwc-0001` stands on its own — the
render-format bug is there with or without any of the rest. They are listed because
the PR mentions local patches and it seems fair to say exactly which.

The two wlroots bugs above (`0001`, `0002`) are filed as issues upstream. I could not
open merge requests for them: this account has no fork rights on
gitlab.freedesktop.org yet, so the diffs are in the issue bodies instead. The scene
half belongs in the discussion on
[wlroots#3794](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3794) and
[!4348](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4348) rather
than in a competing MR of mine.

## @spl237, @cillian64 — one Raspberry Pi specific finding

Since you mentioned your own patches: `libliftoff 0.5.0-1.1+rpt6` calls
`drmCrtcGetSequence`, which upstream 0.5.0 does not. On a CRTC being switched off
there is no sequence to read, so the call fails and `liftoff_output_apply()` returns
`EPERM` for the whole commit. It only triggers when wlroots passes
`liftoff_output_apply_options` — which nothing upstream does today, and which my
`wlroots-0005` above started doing. The symptom was an output that could be turned
off and never back on, with a session restart not recovering it:

```
[ERROR] [libliftoff] drmCrtcGetSequence: Invalid argument
[DEBUG] connector HDMI-A-1: liftoff_output_apply failed: Operation not permitted
[ERROR] Found 0 GPUs, cannot create backend
```

I worked around it on my side by passing the options only for a connector that is
meant to show something — there is nothing to allocate for a commit that disables an
output. Flagging it because anyone else who starts using that API on a Pi will hit
it, and the fix may belong in the rpt patch rather than in every caller.

Happy to test whatever you would like on a Pi 5 with a 4K set — that is what this was
developed on.

cillian64:
As Johan said

> You convince me that you understand it well.
> I try to get help from someone more competent than me to review this.
> I spend time studying it until I can comfortably review it.

So far the initial PR and all responses appear to be AI generated.  Unless we get some input from a real human I'm not inclined to spend time and effort reviewing this.  Please explain what you're doing, why this patch is needed, and what you've done to confirm it is correct, in your own words without pasting a response from an AI.

It's not that I'm against AI assistance for coding, but when reviewers are just sent the output from AI without any context or explanation it places an unreasonable burden on us to figure out what is going on.  PR descriptions and comments should be for humans to write, if I wanted to know what AI thinks about your patch I can just go and ask it myself.

If it's the case that you don't understand the code well enough to explain/evaluate yourself (not saying this is necessarily the case) then that's fine but we'd much rather you submitted a bug report, in your own words, rather than an AI-generated pull request.

Andy1210:
Fair, and sorry.

Let me try to summarize it then: I run labwc on a Pi 5 as a TV box (my own chromium based app launcher). I was messing with wlroots locally to get a 4K video onto a hardware plane because of Plex, because PI's hardware is not powerful enough to render both the video and the semi transparent electron window as an overlay together on the same plane. Trying this made labwc's render format probe fail on my box. After that the screen stayed black and the log
filled up with "Failed to pick primary buffer format" forever.

And I believe the reason is that output_set_render_format() tries the candidate formats one
by one on output->pending, and when none of them works it gives up but leaves
the last one it tried sitting there. So the output is stuck with a format the
renderer can't do.

The patch saves the format that was there before and puts it back if all
candidates fail. Failed probe = nothing changed, instead of a dead output.

I tested it by toggling only this change on the same build: 
with it, video plays and no errors; without it, black screen every time.

I am not sure it can be triggered  at all on other hardware or in a different setup, and you would also need my wlroots patches as well, so if you don't think it's worth carrying, that's also fine.

cillian64:
Thanks for the explanation, that helps a lot.

This change makes sense to me, but I think it would be more technically correct to do `output->pending.committed &= ~WLR_OUTPUT_STATE_RENDER_FORMAT;` instead of re-setting the old working format.  The pending state here is like a diff with flags saying which aspects of the output state it affects, so we can just clear the flag to say "actually don't change the format at all".  Assuming that still fixes your problem I think it would then be fine to merge (though an actual labwc maintainer will have to okay it).

I'd be interested to see the wlroots patches if you could put them on a branch somewhere.  I've done a bunch of work in wlroots to enable scanout of fullscreen videos, and scanout of windowed video is something I want to ship one day.

Edit: Oh, also I reckon we should print a debug-level log message if this happens, it might help avoid some confusing debugging when unsupported output formats are being silently ignored.

Andy1210:
Pushed, thanks. 

I gated the debug line on the existing `silent` flag, because output_state_setup_hdr() runs twice per output on a config apply and the second pass is quiet, so otherwise it would print the same line twice.

Retested on my Pi5, everything is still fine.

the debug line itself has never fired for real. The probe doesn't fail on my tvbox any more, So that part is untested.

The wlroots side is here if you want a look the patches on top of 0.20.2: https://github.com/Andy1210/tvbox/tree/main/scripts/patches

Two of them are plain bugs with nothing to do with the feature (P030 missing from the alpha check, and image descriptions refused on the liftoff interface). 

The [scene one](https://github.com/Andy1210/tvbox/blob/main/scripts/patches/wlroots-0004-scene-offload-the-top-surface-to-an-output-layer.patch) is the same machinery as the windowed video scanout you mentioned: it teaches wlr_scene to hand a surface to the backend as an output layer, which is the piece that doesn't exist today. 
Mine only does the case I needed though, a fullscreen video with one surface above it, so the video keeps the primary plane and the UI goes on an overlay.

johanmalm:
Okay, let's go for it. Thanks.

PR #3687, opened 2026-08-06
output: update comment about wlr_output_layout_remove()

qaqland:
wlroots no longer destroys the wlr_scene_output when its corresponding wlr_output_layout_output is removed.

Refs: https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/e1e911d425fde6ce85475de95b44aa1340725224


Consolatis:
Thanks!

PR #3688, opened 2026-08-06
Hide cursor while idle for a period of time.

seamannnn:
Hide cursor while idle for a period of time.

### DIFF
--- a/src/input/cursor.c
+++ b/src/input/cursor.c
@@ -89,6 +89,14 @@ static_assert(
 	ARRAY_SIZE(cursors_x11) == LAB_CURSOR_COUNT,
 	"X11 cursor names are out of sync");
 
+static int
+hide_cursor(void *data)
+{
+	struct seat *seat = data;
+	cursor_set_visible(seat, false);
+	return 0;
+}
+
 enum lab_cursors
 cursor_get_from_edge(enum lab_edge resize_edges)
 {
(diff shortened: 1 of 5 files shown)


cillian64:
You may be able to do this without a code change https://github.com/labwc/labwc/discussions/3052#discussioncomment-14300031 (assuming that works, I haven't tried it).

seamannnn:
Not really, `swayidle` looks for idle status which could be inhibited from several things (keyboard, video playback, etc...), so it will not give the expected results.
Also note that `wtype` would inhibit the power manager and will keep your computer always on.

src/input/cursor.c
```diff
@@ -1645,6 +1673,15 @@ cursor_init(struct seat *seat)
 	CONNECT_SIGNAL(cursor_shape_manager, seat, request_set_shape);
 	CONNECT_SIGNAL(seat->wlr_seat, seat, request_set_selection);
 	CONNECT_SIGNAL(seat->wlr_seat, seat, request_set_primary_selection);
+
+	if (rc.hide_cursor) {
+		server.hide_cursor_timer = wl_event_loop_add_timer(
+			server.wl_event_loop,
```
johanmalm: Prefer putting multiple arguments on one line - and break at 80 characters.

johanmalm (commented):
I think I'm fine with this. Not tested yet though. Can we merge after `0.20.2` in ca two weeks?

seamannnn:
> I think I'm fine with this. Not tested yet though. Can we merge after `0.20.2` in ca two weeks?

Yeah, sure.

PR #3692, opened 2026-08-07
News for 0 20 2

johanmalm:


NEWS.md
```diff
@@ -93,13 +95,32 @@ The format is based on [Keep a Changelog]
 There are some regression warnings worth noting for the switch to wlroots 0.20:
 
 - lxqt-panel auto-hiding does not work with a panel size greater than 40 [#3600]
-  [wlroots-5392]
+  [wlroots-5392] [wlroots-5407]
```
Consolatis: Isn't this fixed with wlroots 0.20.2? If so suggest recommending that wlroots version. Or something like `With wlroots < 0.20.2 lxqt-panel auto-hiding does not work [..]`.

johanmalm: Thanks. Fixed.

Consolatis:
With commits squashed this LGTM.

PR #3697, opened 2026-08-10
Use labnag for config error display

Consolatis:
Based on
- #3578

20260810_15h06m13s_grim

Changes:
- the write fd is now non-blocking so we don't ever end up in a situation where a log client may block the compositor
- our generic SIGCHLD handler now loops through all terminated child processes instead of only handling the first one found
- the logic in `nag.c` is now (hopefully) more robust against future API usage errors by only allowing to modify the buffer when we are not currently writing to a client
- the log buffer doesn't free its heap allocations  between invocations and on labwc shutdown we reset the buffer and thus free everything. Mysteriously, ASAN doesn't detect the leak but we fix it anyway
- Original commits squashed into one
- Default title for labnag changed from `Config Error` to `Config errors detected`
- Unrelated random fix when compiling with `-Doptimization=1` on an older gcc / libc version

Thanks @elviosak for the initial work!

Testing with something like this seems to work fine and I didn't detect fd leaks nor was triggering any asserts but more testing appreciated:
```sh
#!/bin/bash
while true; do labwc --reconfigure; done
```

### DIFF
--- a/docs/labwc-config.5.scd
+++ b/docs/labwc-config.5.scd
@@ -180,6 +180,7 @@ this is for compatibility with Openbox.
   <xwaylandPersistence>no</xwaylandPersistence>
   <primarySelection>yes</primarySelection>
   <promptCommand>[see details below]</promptCommand>
+  <errorCommand>[see details below]</errorCommand>
 </core>
 ```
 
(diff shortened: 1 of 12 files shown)


johanmalm:
Thank you both. LGTM. Have done some testing and it seems fine.

PR #3699, opened 2026-08-11
nag_log: log xml syntax errors

Consolatis:
20260811_15h13m54s_grim

A slight annoyance is that this will also log pipe menu syntax errors. However, they won't show up because we don't set up a `nag_show()` idle callback and when reconfiguring we clear the buffer first. I am also not sure if we really want to show pipe menu syntax errors via labnag.

Another potential issue I didn't test yet is if we now also get errors from the SVG importers as libxml2 seems to only have a generic callback without referring to some parser instance.

### DIFF
--- a/include/common/buf.h
+++ b/include/common/buf.h
@@ -8,6 +8,8 @@
 #ifndef LABWC_BUF_H
 #define LABWC_BUF_H
 
+#include <stdarg.h>
+
 struct buf {
 	/**
 	 * Pointer to underlying string buffer. If alloc != 0, then
@@ -49,6 +51,16 @@ void buf_expand_shell_variables(struct buf *s);
  * @fmt: format string to be added
  */
 void buf_add_fmt(struct buf *s, const char *fmt, ...);
+
+/**
+ * buf_add_vfmt - add format string to C string buffer
+ * @s: buffer
+ * @fmt: format string to be added
+ * @args: variable arguments created by va_start() by the caller
+ *
+ * The caller is responsible to match its va_start() with a va_end()
+ */
+void buf_add_vfmt(struct buf *s, const char *fmt, va_list args);
 
 /**
  * buf_add_hex_color - add rgb color as hex string to C string buffer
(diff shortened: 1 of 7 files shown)


johanmalm:
@elviosak Fancy reviewing this one (as ~~an alternative~~ complementary to #3698)?

Consolatis:
I think the two are orthogonal to each other. This one just dumps the libxml2 error messages into the buffer while the other adds line numbers to each message. Both features feel good to have on their own.

johanmalm:
> I think the two are orthogonal to each other. This one just dumps the libxml2 error messages into the buffer while the other adds line numbers to each message. Both features feel good to have on their own.

Sorry - need to read with more care :smile:

elviosak:
@johanmalm i'm not sure i can do a good review, but i can take a look and point out some things i notice.

looking at https://gnome.pages.gitlab.gnome.org/libxml2/html/xmlerror_8h.html#ab1b127a183bcaaf7de7f08928dd5994a
it says `xmlSetGenericErrorFunc()` is deprecated, suggests using `xmlSetStructuredErrorFunc()` which is also deprecated in favor of various more specific callbacks for different errors, not sure which one is the one for syntax errors.

> A slight annoyance is that this will also log pipe menu syntax errors. However, they won't show up because we don't set up a nag_show() idle callback and when reconfiguring we clear the buffer first. I am also not sure if we really want to show pipe menu syntax errors via labnag.

for this you can set error handler to NULL after parsing to reset it to default, but would need to set it again on reconfigure, it should also prevent it from running in the SVG parser if it doesn't run during the config/menu parsing.

`buf_add_vfmt` is a good addition, it might be good to change `nag_log` to use it too to avoid using `##__VA_ARGS__` twice.

johanmalm:
@elviosak I think you already do good review :smile:

elviosak:
just tested here and it looks like it doesn't work on reconfigure:

https://github.com/user-attachments/assets/4ca61289-b300-47f4-90de-a1a8b07e53ea

PR #3700, opened 2026-08-13
menu: wait for pipemenus to render before entering via accelerator

### LINKED ISSUE #3696: menu accelerator keys for pipemenus?
Ever since menu accelerators were implemented, they're the primary way i drive menus now. Thank you @ch3rn1ka !
As the title suggests; is there any chance accelerators could be implemented for pipemenus; ie labwc-menu-generator?

ch3rn1ka:
This fix allows to enter pipemenus with accelerators (previously they only got selected because of the race condition).

Fixes #3696.

---

While figuring out the fix, I noticed that the creation of a pipemenu is blocking, i.e. `menu_process_item_selection()` immediately returns if a pipemenu is getting created. I tried to follow this rule, making `menu_submenu_enter()` return in the same case. But generally, as mentioned by @Consolatis in the related issue, it would be nice to allow to cancel pipemenu creation by moving the selection, for example.

### DIFF
--- a/include/menu/menu.h
+++ b/include/menu/menu.h
@@ -70,7 +70,7 @@ void menu_item_select_next(void);
 void menu_item_select_previous(void);
 
 /**
- * menu_item_select_by_accelerator - selects the next menu item with
+ * menu_process_accelerator - selects the next menu item with
  * a matching accelerator, starting after the current selection
  *
  * @accelerator a shortcut to quickly select/open an item, defined in menu.xml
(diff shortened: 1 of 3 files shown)


ch3rn1ka:
Whoops, accidentally used the branch from one of the previous PRs. Should I squash the unrelated commits? Sorry for the noise.

~The fix starts with 9374d1f.~

EDIT: fixed it

johanmalm:
> Whoops, accidentally used the branch from one of the previous PRs. Should I squash the unrelated commits? Sorry for the noise.


Yes please.

zeusgoose:
quickly tested; feels good. again thanks for the patience with my idiot arse. many many thanks

johanmalm:
Thanks

Issue #3643, opened 2026-06-22
SIGABRT in wlr_output_finish during session-active handling when locking via ext-session-lock-v1


ThatDawgMeta:
### Problem description

Environment:

labwc 0.20.0-2.1
wlroots0.20 0.20.1-1.1
CachyOS (Arch-based), AMD integrated GPU (AMDI0010)
Lock client: swaylock-fprintd-git (r336.536d9df-1), invoked via ext-session-lock-v1

Description:

Compositor crashes with SIGABRT when locking the session (triggered via a keybind calling swaylock). Crash is reproducible across multiple separate lock attempts.
```
Backtrace (symbolized via debuginfod):
#0 pthread_kill
#1 raise
#2 abort
#3 __assert_fail
#4 wlr_output_finish (cold path) — libwlroots-0.20.so
#5 drm_connector_destroy_output — backend/drm/drm.c:1249
#6 handle_session_active — backend/drm/backend.c:118 (discriminator 1)
#7 wl_signal_emit_mutable
#8 handle_disable_seat — backend/session/session.c:34
#9 libseat_event — backend/session/session.c:48 (discriminator 1)
#10 (libseat) set_active — libseat/backend/logind.c:359
#11 (libseat) dispatch_and_execute — libseat/backend/logind.c:267
```
Sequence: logind sends seat-active notification → libseat dispatches → wlroots session active-signal handler fires → DRM backend's handle_session_active attempts to destroy/recreate a DRM connector's output → assertion inside wlr_output_finish fails → abort.

### Steps to reproduce

Lock the session via a client using ext-session-lock-v1 (tested with a custom swaylock build with native fprintd support). Crash occurred on two separate lock attempts during the same session.

### labwc build source

Release

### labwc version

0.20.0-2.1

### labwc environment

From a TTY or some display manager like lightdm

### Distribution

CachyOS

johanmalm:
Sorry it took a while to answer this one. Am on summer break still.

It looks like the frames leading up to the abort() are mostly in wlroots. Would you mind raising an issue there to get some more eyes on it. It may well be something we have done on the labwc side, but I cannot tell from the trace.

I use swaylock with a keybind all the time, so suggest also doing some detective work if you are able to. For example:

1. Does it happen with sway (which has a lot in common)
2. Does it happy with plain swaylock (without modification)
3. At a guess, the abort may relate to a disappearing output, so it would be good to understand if the keybind does anything other than calling swaylock. If so, try without that.
[...]

PR #3701, opened 2026-08-14
output: destroy scene output before scene cleanup

### LINKED ISSUE #3689: Crash labwc when suspend with swaylock
### Problem description

labwc crashes with an assertion error (`labwc: types/output/output.c:401: wlr_output_finish: Assertion 'wl_list_empty(&output->events.bind.listener_list)' failed.`) when the system attemps to go to sleep/suspend while `swaylock` is active.

Intriguingly, this crash **only occurs if the monitor layout has been adjusted using `wlr-randr` or `kanshi`** beforehead. If `wlr-randr` is not used, `swaylock` displays correctly and the system suspends/wakes up without any issues.

Here is the log file generated with labwc --debug 2> /tmp/labwc.log

[labwc.log](https://github.com/user-attachments/files/30825997/labwc.log)

### Steps to reproduce

1. launch labwc with any DM or tty (I've tested on `Ly`, `SDDM`, and from tty)
2. adjust monitor layout using `kanshi` or `wlr-randr` like
```
$ wlr-randr \
--output DP-1 --mode 1920x1080@280.019012Hz --pos 0,440 \
--output DP-2 --mode 1920x1080@144.001007Hz --pos 1920,0 \
--output HDMI-A-1 --mode 1920x1080@60.042000Hz --pos 0,1520
```
3. run `swayidle -w before-sleep 'swaylock -f' &`
4. suspend manually with `systemctl suspend`
5. resume from suspend then labwc is already crashed

### labwc build source

Release

### labwc version

0.20.1

### labwc environment

### Distribution

Arch Linux

jlvihv:
## Problem

When an output is destroyed, existing `wlr_surface_output` bind listeners are removed while wlroots emits the output destroy signal. Labwc then destroys per-output scene nodes while the corresponding `wlr_scene_output` is still active.

Those scene changes can send a new surface-enter for the output being destroyed. The newly created bind listener misses the destroy signal already in progress, leaving `output->events.bind` non-empty and making `wlr_output_finish()` abort.

This is the failure mode described in [wlroots#4096](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4096).

## Change

Destroy the associated scene output before changing the per-output scene graph, and clear the pointer to prevent the later automatic cleanup from being relied upon.

## Testing

- `./scripts/check`
- Built with `-Dtest=enabled -Dwerror=true`
- `meson test --verbose`: 3/3 passed
- Reproduced the assertion with Labwc 0.20.1 and wlroots 0.20.2 on Arch Linux during suspend/resume with swaylock active. With this change applied, the DRM output was destroyed and recreated, the same Labwc process survived resume, and no assertion or Labwc core dump was produced.

Fixes #3689.
Related to #3643.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -291,6 +291,21 @@ handle_output_destroy(struct wl_listener *listener, void *data)
 	if (seat->overlay.active.output == output) {
 		overlay_finish(seat);
 	}
+
+	/*
+	 * Destroy the scene output before changing the scene graph below.
+	 * Otherwise those changes can send a surface-enter for this output while
+	 * its destroy signal is already being emitted. The new listener then
+	 * misses that signal and makes wlr_output_finish() abort because
+	 * output->events.bind is not empty.
+	 *
+	 * See https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/4096
+	 */
+	if (output->scene_output) {
+		wlr_scene_output_destroy(output->scene_output);
+		output->scene_output = NULL;
+	}
+
 	wl_list_remove(&output->link);
 	wl_list_remove(&output->frame.link);
 	wl_list_remove(&output->destroy.link);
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks.

PR #3710, opened 2026-08-31
action: allow Close and Kill while window switching

typebrook:
In `src/action.c`, `actions_run()` drops every action other than `NextWindow` and `PreviousWindow` while `server.input_mode == LAB_INPUT_STATE_CYCLE`.

Since default binding for cycling through windows use <kbd>Alt</kbd>+<kbd>Tab</kbd>, this makes it impossible to bind **any key with <kbd>Alt</kbd> to Close and delete windows** without first releasing <kbd>Alt</kbd> to dismiss the window switcher. Which is inconvenient to use combo on managing windows.  

This commit replaces the hard-coded pair of allowed actions with a small whitelist and add Close and Kill to it. Neither changes focus, opens a menu nor starts an interactive move/resize, so neither can fight with the switcher's focus override.

Actions run while the switcher is up now act on the view highlighted in the OSD rather than on the still-active view, which is the window the user is looking at.

Also name the rejected action in the log message, which is more useful now that the set of accepted actions is no longer a fixed pair.

--
This PR is requested by @johanmalm in https://github.com/labwc/labwc/issues/3255#issuecomment-5454772429, my purpose is to make behaviors of switching windows more close to Openbox.

#3255 originally consider adding new option `<cycleExclusive>` into `<windowSwitcher>` to allow more actions while switching windows. But **the possibility of unexpected behaviors** is the major reason of not doing it right now. Will glad to see more feedback about it.


johanmalm:
Thanks. LGTM. Have tested and seems to work fine.

Issue #270, opened 2022-03-10
Focus issues when minimizing window


acoolstraw:
I have 2 windows open: Chromium on top, and VS Code behind it. When I minimize Chromium and click on VS Code it just brings up Chromium and focuses on it. Alt+Tab works as expected, but sometimes I'd rather minimize a window than cycling between them

Consolatis:
What branch / version of labwc are you using?

acoolstraw:
Using `labwc` from the AUR, version 0.5.0. It does happen with `-c /dev/null` arguments

Consolatis:
After some more discussion in IRC the issue seems to be related to minimizing any X11 application.
It seems to fail giving keyboard / mouse focus to the new top most client, thus further keyboard / mouse input ends up on the minimized application.

In the case of Chromium that in turn seems to cause it to send a request to the xwayland server to be focused / unminimized.

Consolatis:
@acoolstraw would be great if you could test https://github.com/Consolatis/labwc/tree/fix/issue-270

acoolstraw:
Works!

PR #3712, opened 2026-09-04
docs: point the gettext tutorial at an archived copy

rawsun007:
The translations section of `CONTRIBUTING.md` ends with:

> [See this tutorial for further guidance](https://www.labri.fr/perso/fleury/posts/programming/a-quick-gettext-tutorial.html)

That URL returns 404. The page is gone rather than moved: the author's site root still resolves, but the post is absent from his blog index, so there is nothing live to repoint to. This uses the Wayback snapshot from 2026-03-25, which preserves the tutorial the sentence promises.

If you would rather not link an archive, the natural live substitute is the [GNU gettext manual](https://www.gnu.org/software/gettext/manual/gettext.html) — authoritative and stable, but a reference manual rather than a quick tutorial, so it changes what a translator gets. Happy to switch to that instead; I went with the archive because it keeps the original intent.

One other dead link I left alone: `NEWS.md` line 2893 has `[#270]: .../pull/270`, and neither PR nor issue 270 exists any more. That is the historical changelog, so rewriting entries would falsify the record.

AI disclosure: found and prepared with Claude Opus 5 in Claude Code, which swept every link in the repo markdown; I verified the dead URL, the author's blog index and the archive snapshot myself. The commit carries the `Assisted-by:` tag your contributing guide asks for. The change is a single URL, so there is no licensing question about generated code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -446,7 +446,7 @@ generate their `MY_LOCALE.po` file in a few steps:
 contact and locale details in the header of the file. Then, add the
 translation strings under each English string.
 
-[See this tutorial for further guidance](https://www.labri.fr/perso/fleury/posts/programming/a-quick-gettext-tutorial.html)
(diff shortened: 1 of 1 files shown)


johanmalm:
Thanks

PR #3715, opened 2026-09-05
input: support opt-in libinput custom acceleration

kgagne:
## Summary

Add opt-in native libinput custom acceleration through `<accelProfile>custom</accelProfile>`, with separate Motion, Scroll and Fallback curves. This allows precise low-speed scrolling while keeping faster scrolling responsive. Existing flat/adaptive configurations retain their behavior.

- Delegate velocity tracking, interpolation and extrapolation to libinput; remove the compositor-side scroll-curve machinery.
- Accept `motionCurve`, `scrollCurve` and `fallbackCurve` with named `step` and `points` fields, as attributes or child elements. Points represent output speeds, not multipliers.
- Reject malformed curves without partially applying a custom configuration. Build a fresh native configuration on reload so omitted curves return to their defaults.
- Document that custom acceleration is device-wide: unspecified Motion/Scroll uses Fallback, unspecified Fallback is unaccelerated, and `pointerSpeed` has no effect in custom mode. `scrollFactor` continues to scale scroll events.

`step` is retained because it specifies the spacing of libinput's input-speed samples; point count alone does not define that range. Named fields fix the XML attribute/content concatenation problem. Parsing uses GLib splitting, initialized locals and `nr_points`.

This follows the use case discussed in #3714.

## Validation

- Optimized build with warnings as errors and manual-page generation.
- All four normal and all four ASan/UBSan test suites pass.
- Full configuration-parser checks cover opt-in selection, malformed curves and clearing state on reload.
- Final binary passes isolated headless startup checks.
- Tested on a Dell XPS 15 9520 with libinput 1.31.1. After restarting TTY1 into the finalized binary, scrolling retained the tested low-speed precision; the running binary and preserved configuration were verified.

Blender 5.2.1 testing exposed a separate fractional-scroll precision loss in its Wayland handling. Incoming events and matching source were recorded; no Blender-specific workaround is included here.


johanmalm:
Thanks for contributing. This looks like a good feature to add.

I have tested it but cannot get it work though.

I get this error:

```
00:00:00.018 [ERROR] [../src/common/parse-double.c:77] value '0.100.10' contains multiple decimal markers
00:00:00.018 [ERROR] [../src/config/rcxml.c:744] <scrollCurve> factors must be finite and non-negative
00:00:00.018 [ERROR] [../src/config/rcxml.c:753] <scrollCurve> supports at most 32 points
00:00:00.018 [ERROR] [../src/config/rcxml.c:760] <scrollCurve> requires at least two factors
```

...with the following section in `rc.xml`:


```
  <libinput>
    <device>
      <scrollCurve step="0.10">0.10 0.15 0.20 0.25 0.30 0.42 0.58 0.78 1.00 1.10</scrollCurve>
    </device>
  </libinput>
```

Have only had a quick glance through code. Keen to get it working before doing full review. Quick feedback:

- Is the step attribute needed? Could it be deduced from the number of elements in the curve (1 / nr_elements)? Or am I misunderstanding what `step` means.
- For consistency with the rest of the code base I prefer `nr_` as a prefix indicating "number of", so `s/npoints/nr_points/g`
- I prefer to initialise variables most of the time if it makes it easier for the reader to verify correctness and prevent future refactoring breaking things. For example in `parse_scroll_curve()`, `step_string` and `step` are not initialized - which is fine, but I had to check the logic of both `lab_xml_get_string()` and `set_double()` to satisfy myself that it's not a ticking time bomb.
- I normally prefer `g_strsplit()` to `strtok_r()` for readability, but happy to leave as is.

johanmalm:
Could this be achieved with the `custom` option of libinput's acceleration profile via some libinput-configuration call rather than applying it directly in `cursor.c` `handle_axis()`?

https://wayland.freedesktop.org/libinput/doc/latest/pointer-acceleration.html#ptraccel-profile-custom

kgagne:
Thanks for the suggestion. libinput does provide LIBINPUT_ACCEL_TYPE_SCROLL through its custom acceleration profile, and using that would let us avoid the velocity tracking and interpolation in cursor.c.

The complication is that applying a custom configuration also switches the device’s pointer acceleration profile to CUSTOM. According to the [custom-profile documentation](https://wayland.freedesktop.org/libinput/doc/latest/pointer-acceleration.html#ptraccel-profile-custom), an unspecified Motion function uses Fallback, which defaults to no acceleration. So configuring only Scroll would also replace the touchpad’s existing adaptive pointer acceleration.

My intention with this PR was to adjust finger scrolling while preserving pointer motion and the existing <accelProfile> behavior. From the documented API, I don’t see a way to combine custom Scroll with the built-in adaptive Motion profile.

Would you prefer keeping the scroll adjustment in Labwc for that reason, or broadening this into support for libinput’s custom profile with explicit motion and scroll configuration? I’m happy to follow either direction; I’d like to settle that before revising the parser and curve configuration.

kgagne:
Having thought about this further, I think opt-in support for libinput’s custom acceleration profile would be a more complete solution.

Existing configurations would retain their current behavior. Users explicitly selecting the custom profile could configure Motion and Scroll curves, plus Fallback for other movement types, including pinch zoom.

Opting in would change the acceleration profile for the whole device, so we’d need to document how unspecified curves behave and how this interacts with `<pointerSpeed>` and `<accelProfile>`.

Would that direction fit what you had in mind? I’d be happy to rework this PR around native libinput configuration and remove the velocity tracking and interpolation from `cursor.c`.

johanmalm:
Yes I prefer to use libinput functionality where possible, and only resort to upstream overrides where necessary and there is a clear net benefit.

I think this also has benefits like:
- Consistency with eco-system
- Least surprise to users
- Avoidance of future potential regressions relating to having overlapping control knobs

kgagne:
Thanks for clarifying. I've reworked the PR around libinput's native custom acceleration profile and removed the velocity tracking and interpolation from `cursor.c`.

Users explicitly opt in through `<accelProfile>custom</accelProfile>`, with separate `motionCurve`, `scrollCurve` and `fallbackCurve` settings. Existing flat/adaptive configurations retain their behavior. The documentation covers the device-wide profile change, unspecified curves and the interaction with `pointerSpeed`.

I've also addressed the parser feedback: named `step` and `points` fields avoid the XML concatenation issue, and parsing now uses GLib splitting, initialized locals and `nr_points`. `step` remains because it defines the spacing of libinput's input-speed samples; the number of points alone doesn't determine that range.

Testing on my XPS gives excellent low-speed precision, including after restarting into the finalized build. The build, parser and sanitizer tests pass, along with all five GitHub CI checks.

Testing also exposed fractional-scroll precision loss in Blender's Wayland handling. I've documented that separately; no application-specific workaround is included in this PR.

johanmalm:
Thanks. Could you provide some sample curves that I can test with?

johanmalm:
So is the values of the curve the actual output rather than some factor?

And the range is 0.0 to `step * (nr_points - 1)` ?

kgagne:
Yes. The curve values are actual output speeds, not factors.

The input points are spaced by step, so with nr_points values the explicitly defined points cover:

0.0 through step * (nr_points - 1)

Speeds between those points are linearly interpolated. Speeds above the final defined point aren’t clamped to that range; libinput linearly extrapolates using the final two points.

The effective factor at any speed is therefore:

output_speed / input_speed

For example, an input speed of 0.20 mapped to an output speed of 0.04 gives an effective factor of 0.20x.

For testing, this is the Scroll curve I’ve been using on the XPS 15 9520:
```
<accelProfile>custom</accelProfile>
<motionCurve step="1" points="0 1" />
<fallbackCurve step="1" points="0 1" />
<scrollCurve step="0.10" points="0 0.015 0.040 0.075 0.120 0.210 0.348 0.546 0.800 0.990 1.100" />
```

I only tuned the Scroll curve. I wasn’t trying to tune Motion or Fallback, so those are identity curves here.

The goal was very precise low-speed two-finger scrolling. On this touchpad I wanted to be able to deliberately move content by roughly 2–3 px without the beginning of the gesture jumping farther than intended, while still retaining useful normal and fast scrolling.

Some points illustrate the intended progression:

* 0.10 -> 0.015 = 0.15x
* 0.20 -> 0.040 = 0.20x
* 0.30 -> 0.075 = 0.25x
* 0.40 -> 0.120 = 0.30x
* 0.50 -> 0.210 = 0.42x
* 0.60 -> 0.348 = 0.58x
* 0.70 -> 0.546 = 0.78x
* 0.80 -> 0.800 = 1.00x
* 0.90 -> 0.990 = 1.10x
* 1.00 -> 1.100 = 1.10x

Lowering the early output values gives finer low-speed control. Raising them makes small movements more responsive but also easier to overshoot.

The middle values control how quickly scrolling transitions out of that precision region, while the final values control faster gestures.

Changing step shifts where all of those samples occur. A smaller step compresses the curve into lower input speeds; a larger step stretches it across a wider input-speed range.

These values are tuned for this particular touchpad rather than intended as universal defaults.

johanmalm:
LGTM and happy to merge.

Would you want to enhance the documentation before we merge? For example:
- Describe the data series a bit more including that the values are outputs speeds and the range goes from `0.0` to `step * (nr_points - 1)`. Your sentence on `{inter,exra}polation` is also very helpful. I think that with this information, users should be able to understand.
- Provide some examples beyond `step=1` and `points="0 1"`. Your XPS example is good and should make the whole thing very understandable to users.

kgagne:
Thanks. I've expanded the documentation to explain the explicitly defined input-speed sample range, output-speed semantics, interpolation/extrapolation, and effective gain. I've also added the XPS scroll curve as a practical device-specific example, with identity Motion and Fallback functions. CI is passing.

johanmalm:
Thanks

PR #3716, opened 2026-09-06
buf.c: limit tilde expansion

domo141:
Expand tilde if:
it is at the beginning of a string, or following space, tab or ':'
…and…
it is at the end of string, or followed by space, tab, ':' or '/'.

Now it somewhat closer to how POSIX shells do it.

--8<-- --8<--

Now if HOME of user is /home/user, ~user no longer expands to /home/useruser (just one example :)

For comparison, openbox uses this regexp when expanding tildes in (potentially multiline) strings:

`(?:^|(?<=[ \\t]))~(?=[/ \\t$])`

That does not handle `VAR=part1:~:part2` but maybe there more to it than just one thing in
`opt_paths_expand_tilde()`.


johanmalm:
Thanks.

PR #3719, opened 2026-09-08
libinput: add clickfingerButtonMap option

RyoheiHashimoto:
This covers the `clickfinger behavior` item of #1079. libinput's click
configuration has exactly two settables — the click method and the clickfinger
button map — and the method landed in #1416, so this is the remaining one.

When the clickfinger click method is in use, libinput maps one-, two-, and
three-finger clicks to left/right/middle on most devices. Some users prefer
left/middle/right, which libinput supports but labwc had no way to select.

```xml
<libinput>
  <device category="touchpad">
    <clickMethod>clickfinger</clickMethod>
    <clickfingerButtonMap>lmr</clickfingerButtonMap>
  </device>
</libinput>
```

This touches the same six files as #1416 and follows the same shape.

**Implementation notes**

- The category field uses the `-1` sentinel used by `clickMethod` and its
  neighbours rather than the enum default used by `tapButtonMap`, so the device
  default is left untouched unless the option is set. The reset-to-default call
  is included, so removing the option and reloading restores the device default.
- Skipped when the device does not advertise
  `LIBINPUT_CONFIG_CLICK_METHOD_CLICKFINGER`.
- Values are matched case-insensitively, as `clickMethod` and most other options
  in rcxml.c are.
- libinput can still return `LIBINPUT_CONFIG_STATUS_UNSUPPORTED` for a device
  that allows clickfinger but rejects a specific map. The return value is not
  checked, matching 29 of the 30 `libinput_device_config_*_set_*` calls in
  seat.c.
- The API was added in libinput 1.26, already the minimum in meson.build, so no
  `has_header_symbol` guard is needed — unlike
  `LIBINPUT_CONFIG_DRAG_LOCK_ENABLED_STICKY` /
  `LIBINPUT_CONFIG_3FG_DRAG_ENABLED_3FG`, which are newer.

**Testing**

- gcc and clang, with and without xwayland, and with `-Dtest=enabled`, all with
  `--werror`: no warnings; `meson test` passes; `scripts/check` and
  `scripts/find-banned.sh` are clean. CI is green on Arch, Debian, FreeBSD and
  Void-musl on my fork:
  https://github.com/RyoheiHashimoto/labwc/actions/runs/34243277473
- Tested end-to-end with labwc against a uinput virtual clickpad (udev tags it
  `ID_INPUT_TOUCHPAD`; libinput reports `Click methods: *button-areas
  clickfinger`):

  ```
  [INFO] src/seat.c:166 matched category: touchpad
  [INFO] src/seat.c:319 click method configured (2)
  [INFO] src/seat.c:342 clickfinger button map configured (1)
  ```

  `lmr`, `LMR` and `Lmr` all give 1; `lrm` gives 0; an invalid value logs
  `invalid clickfingerButtonMap`. A virtual mouse on the same seat logs
  `clickfinger button map not configured`, as expected.
- Reading the value back from libinput on the same device: unset leaves the
  device default (LRM), `lmr` yields LMR, and removing the option again restores
  LRM — confirming the reset-to-default call.
- I do not have a physical clickfinger touchpad, so the above uses a virtual
  (uinput) device rather than real hardware.
- Built against libinput 1.31.3; the enum and both function declarations are
  identical in the 1.26.0 header.


johanmalm:
LGTM. Thanks.

PR #3720, opened 2026-09-09
libinput: add rotationAngle option

RyoheiHashimoto:
This covers the `rotation` item of #1079. Same six files and same shape as
#1416 and #3719.

```xml
<libinput>
  <device category="default">
    <rotationAngle>90</rotationAngle>
  </device>
</libinput>
```

**Implementation notes**

- The full libinput range (0-359) is accepted rather than the
  `[0|90|180|270]` vocabulary of `<tablet rotate>`. libinput does apply
  arbitrary angles: on a rotation-capable device `set_angle(45)` and
  `set_angle(135)` both return `SUCCESS` and read back unchanged, so
  restricting to multiples of 90 would drop working functionality.
  `<tablet rotate>` is a separate mechanism which maps to coordinate
  transforms rather than to libinput.
- Parsed with `strtol()` plus a terminator check, matching the
  `calibrationMatrix` parsing directly above it. `atoi()` is used elsewhere
  in this function, but 0 is a valid angle here, so it would silently accept
  a malformed value as "no rotation".
- Gated on `libinput_device_config_rotation_is_available()`. Note that
  libinput reports rotation as available on more than just pointer devices —
  a keyboard does too (`libinput list-devices` shows `Rotation: 0.0` rather
  than `n/a`) — so under `category="default"` the setting is applied there
  as well. That is harmless, since such devices emit no motion events, and it
  keeps the gate consistent with libinput's own answer rather than
  second-guessing it.
- The reset-to-default call is included so that removing the option and
  reloading restores the device default.

**Testing**

- gcc and clang, with and without xwayland, and with `-Dtest=enabled`, all
  with `--werror`: no warnings; `meson test` passes; `scripts/check` and
  `scripts/find-banned.sh` are clean. CI is green on Arch, Debian, FreeBSD
  and Void-musl on my fork: https://github.com/RyoheiHashimoto/labwc/actions/runs/34322736281
- End-to-end with labwc against uinput virtual devices (a clickpad, a mouse
  and a keyboard on one seat). With `<rotationAngle>90</rotationAngle>` under
  `category="default"`:

  ```
  configuring input device labwc-test-clickpad (event0)
  rotation angle not configured
  configuring input device labwc-test-mouse (event1)
  rotation angle configured (90)
  configuring input device labwc-test-keyboard (event2)
  rotation angle configured (90)
  ```

- Parsing: `0`, `45`, `90`, `270`, `359`, `+90` and `-0` are accepted;
  `360`, `-1`, `abc`, `9O`, `0x5A`, `90.5`, values which overflow `long`,
  and whitespace all log `invalid rotationAngle`; an empty element is caught
  by the existing "Empty string is not allowed" check.
- Reload: with the option set to 90, changing it to 270 and sending SIGHUP
  logs `rotation angle configured (270)`; removing the option and sending
  SIGHUP again logs `rotation angle not configured`, and reading the angle
  back from libinput confirms the device default is restored.
- I do not have a physically rotated device, so the above uses virtual
  (uinput) devices rather than real hardware.


johanmalm:
LGTM. Thanks

PR #3729, opened 2026-09-15
CONTRIBUTING.md: make AI policy stricter

johanmalm:


johanmalm:
@Consolatis @jlindgren90 - Appreciate views on this one.

Cc: @ell1e

jlindgren90:
It's fair and I have no objection. I am a little skeptical of the correlation with decreasing active contributors as I suspect that has more to do with unrelated real-life events and (perversely) the maturing of the project, i.e. fewer itches left to scratch. I saw the same happen with Audacious well before AI was on the scene. But that's only speculation on my part.

johanmalm:
> It's fair and I have no objection. I am a little skeptical of the correlation with decreasing active contributors as I suspect that has more to do with unrelated real-life events and (perversely) the maturing of the project, i.e. fewer itches left to scratch. I saw the same happen with Audacious well before AI was on the scene. But that's only speculation on my part.

Thanks. Yes, fair :smile:

ell1e:
I think this would be a great policy, personally! :exploding_head:

johanmalm:
I removed the sentence on "decreasing active contributors".

Added a note to make clear that this includes text in issues, discussion and PR descriptions.

ell1e:
It is awesome that you banned LLM use with today's problematic LLMs! :tada: 

However, I keep coming back ot this in my head, so while it seems like a nitpick I feel like I should point it out:

> The points above are not derived from any political position relating to ethical
concerns like energy consumption and similar. No doubt individual maintainers
and contributors have thoughts on this, but it has not knowingly shaped the
policy.

This paragraph is a bit strange. Some of the ethical concerns are really serious, e.g. [who these companies align with](https://hachyderm.io/@davidculley/117281012524335637) and [the copying could also be considered an ethical problem](https://www.theatlantic.com/technology/2026/01/ai-memorization-research/685552/) rather than just a legal risk.

You even point out "community concerns" in the next paragraph right after. It seems to me like deeply unethical practices might be something that is at risk of affecting the community.

I would therefore suggest you consider simply removing that paragraph!

PR #3736, opened 2026-09-21
Add optional Systemd User Daemon integration

Shua24:
## The Problem

Labwc currently relies on traditional session startup methods without other options. This makes Labwc behave differently from compositors with integrated graphical session lifecycle with Systemd user services, especially those with the standard `graphical-session.target` and `xdg-desktop-autostart.target` infrastructure as their integration dependencies.

Because of this applications that may require the Systemd XDG graphical session lifecycle may not be started or stopped properly as part of the Labwc session in the same way as they are under a systemd-integrated compositor. This detaches Labwc completely from other system services that may be useful for accessibility, such as Gammastep with the assistance of Geoclue for easier latitude/longitude searching. 

## Proposed Solution

Add an optional Systemd user-session integration mode for Labwc. The option to integrate is available via the XML configuration. The configuration defaults to "false" if not specified so that existing labwc session startup behavior remains unchanged for users who do not opt into systemd integration.

When enabled Labwc will be started and managed through a Systemd user service whose service files and dependencies are at `~/.config/systemd/user/`. The service participates in the standard graphical-session lifecycle associated with `graphical-session.target`, `graphical-session-pre.target` and `xdg-desktop-autostart.target`. This activates the three of them upon restart if the user specifies to turn the integration on in `rc,xml`. This allows applications and services which use these targets to follow the labwc session lifecycle.

## Implementation Summary

- Add an XML configuration item `<daemonize></daemonize>` as a configuration option in `rc.xml` to enable/disable systemd session integration.
- Generate or remove the corresponding systemd user service according to the configuration depending on the user's needs and preferences to integrate
- Define the service as part of the graphical session lifecycle by integrating the Systemd service API.
- Start labwc through the Systemd managed session path when integration is enabled.
- Add the corresponding session wrapper/launcher called `labwc-session` to handle Systemd managed startup.
- Handle session configuration changes upon restart and cleanup so the graphical session is not left with stale environment/state.
- Keep the feature opt-in rather than changing the default labwc startup model.

Hope this helps. This is my first time contributing to this project, and I would like some feedback regarding the contributions I've made. Thanks


johanmalm:
This seems to contravene https://github.com/labwc/labwc/blob/master/CONTRIBUTING.md#use-of-ai-generated-content

Would you be able to evaluate if the described issue is solved by #3534

I think it requires this line to be un-commented.
https://github.com/labwc/labwc/blob/7cc6459918ed28033151f33b16330524fb6ab0b4/docs/autostart#L3-L9

